PR #2553, opened 2025-11-05
Fix a lot of bugs in the procfs implementation

lrh2000:
Unfortunately, there are just too many bugs in our procfs implementation.

 1. Inconsistent lock orders:
     - process table -> cached entries
https://github.com/asterinas/asterinas/blob/ff021e0b89811c92069950a5a96ef41f9860de0b/kernel/src/fs/procfs/mod.rs#L142
     - cached entries -> process table
https://github.com/asterinas/asterinas/blob/ff021e0b89811c92069950a5a96ef41f9860de0b/kernel/src/fs/procfs/mod.rs#L204
 2. Atomic mode break:
     - file table (spin lock) -> cached entries (mutex)
https://github.com/asterinas/asterinas/blob/ff021e0b89811c92069950a5a96ef41f9860de0b/kernel/src/fs/procfs/pid/task/fd.rs#L54
 3. This causes wrong results if we execute a new program in a non-main thread:
https://github.com/asterinas/asterinas/blob/ff021e0b89811c92069950a5a96ef41f9860de0b/kernel/src/fs/procfs/pid/mod.rs#L32
 4. This causes wrong results if we unshare the file table:
https://github.com/asterinas/asterinas/blob/ff021e0b89811c92069950a5a96ef41f9860de0b/kernel/src/fs/procfs/pid/task/fd.rs#L39
 5. The task directory will still exist even after the thread exists:
https://github.com/asterinas/asterinas/blob/ff021e0b89811c92069950a5a96ef41f9860de0b/kernel/src/fs/procfs/pid/task/mod.rs#L202

---

This PR aims to fix the bugs. Most changes should be straightforward. The only exception is how to organize `/proc/[pid]/fd/[n]`. This leads to several challenges that make our previous `FdEvents` observer mechanism unlikely to work.
 - The file table is protected by a spin lock, so notifying the `FdEevents` to the procfs will cause atomic break because the cached entries are protected by a mutex.
 - The file table can be shared by multiple threads or processes. Each thread or process can choose to unshare at any time. Then how to deal with the previously registered observers? Cloning them will lead to redundant observers but discarding them will lead to inconsistent procfs states.

[...]

PR #2556, opened 2025-11-06
Revise the content of root field for the mountinfo file

cchanging:
Fix the bug mentioned in https://github.com/asterinas/asterinas/pull/2553#issuecomment-3495489473, that the root field should show the path name of the mount root within the filesystem instead of its `abs_path`.

I also find that our `abs_path` method and `effective_name` method does not be aware of the current process's root path. That is, the `effective_name` of the current process's root path should be `/` and the `abs_path` should be the path name relative to the process's root path.

Our current API doesn't seem to be able to pass the process root path into this method very elegantly (we can't pass `ctx` for all abs_path usages, we might have to use `current_thread!`). Maybe I need to find a better approach to solve this problem.

### DIFF
--- a/kernel/src/fs/path/dentry.rs
+++ b/kernel/src/fs/path/dentry.rs
@@ -332,6 +332,27 @@ impl Dentry {
         }
         Ok(())
     }
+
+    /// Gets the absolute path name of this `Dentry` within the filesystem.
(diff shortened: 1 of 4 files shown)


cqs21 (approved):

PR #2558, opened 2025-11-06
Remove IntervalSet::find_one unreachable code

HangShuYH:
### Incorrect Logic in IntervalSet::find_one
The `cursor.peek_next().start` should be strictly greater than point due to the use of `Excluded(point)` in the `lower_bound` parameter. Therefore, the condition in the else branch appears to never be satisfied.
```rust
pub fn find_one(&self, point: &K) -> Option<&V> {
    let cursor = self.btree.lower_bound(core::ops::Bound::Excluded(point));
    // There's one previous element and one following element that may
    // contain the point. If they don't, there's no other chances.
    if let Some((_, v)) = cursor.peek_prev() {
        if v.range().end > *point {
            return Some(v);
        }
    } else if let Some((_, v)) = cursor.peek_next() {
        if v.range().start <= *point {
            return Some(v);
        }
    }
    None
}
```

### DIFF
--- a/kernel/src/vm/vmar/interval_set.rs
+++ b/kernel/src/vm/vmar/interval_set.rs
@@ -70,18 +70,11 @@ where
     /// that contains the point.
     pub fn find_one(&self, point: &K) -> Option<&V> {
         let cursor = self.btree.lower_bound(core::ops::Bound::Excluded(point));
-        // There's one previous element and one following element that may
-        // contain the point. If they don't, there's no other chances.
-        if let Some((_, v)) = cursor.peek_prev() {
-            if v.range().end > *point {
-                return Some(v);
-            }
-        } else if let Some((_, v)) = cursor.peek_next() {
-            if v.range().start <= *point {
-                return Some(v);
-            }
-        }
-        None
+        // There's only one previous element that may contain the point.
+        // If it doesn't, there's no other chances.
+        cursor
+            .peek_prev()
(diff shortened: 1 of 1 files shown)


kernel/src/vm/vmar/interval_set.rs
```diff
@@ -70,16 +70,12 @@ where
...
     pub fn find_one(&self, point: &K) -> Option<&V> {
         let cursor = self.btree.lower_bound(core::ops::Bound::Excluded(point));
-        // There's one previous element and one following element that may
-        // contain the point. If they don't, there's no other chances.
+        // There's only one previous element that may contain the point.
+        // If it doesn't, there's no other chances.
         if let Some((_, v)) = cursor.peek_prev() {
             if v.range().end > *point {
                 return Some(v);
             }
```
cchanging: Thank you for your contribution! This was indeed some hidden dead code. After removing them, I think we can implement this function with a more compact approach:
```rust
if let Some((_, v)) = cursor.peek_prev() {
    (v.range().end > *point).then_some(v)
} else {
    None
}
```
or
```rust
cursor
    .peek_prev()
    .and_then(|(_, v)| (v.range().end > *point).then_some(v))
```

HangShuYH: I prefer the latter version and I've already updated the pull request

cchanging (approved):

Issue #2356, opened 2025-08-21
Install a minimal NixOS on Asterinas


cqs21:
# Overview

This tracking issue outlines the requirements for supporting the NixOS distribution. The goal is to enable full NixOS functionality on top of the Asterinas kernel.

# Motivation

Choosing NixOS as the first distribution supported for Asterinas is driven by several compelling factors:

**Reproducibility**

- Fully declarative system configuration ensures consistency across different machines
- Configuration as code, facilitating version control and collaboration

**Cross-platform and Customization**

- Support for multiple hardware architectures
- High degree of customization

**Package Management Benefits**

- Multiple package versions coexist without conflicts
- Purely functional package management with clear dependency relationships

**Atomic Upgrades and Rollbacks**

- System upgrades either succeed completely or fail completely
- Automatic rollback to previous working state on failure
- No risk of system corruption from interrupted upgrades

# PoC

A basic Proof of Concept has been implemented in this branch: https://github.com/cqs21/asterinas/tree/dev-iso-installer. This PoC demonstrates the fundamental feasibility of running NixOS on the Asterinas kernel by completing the entire installation and boot cycle.

The current PoC achieves the following key goals:
- **Customized NixOS ISO Installer**: Created a modified NixOS installation ISO that includes all necessary Asterinas kernel components and dependencies.
- **Asterinas Installation**: Successfully installs the Asterinas kernel and NixOS system to a target disk.
- **System Boot**: Boots the installed system and reaches a functional NixOS shell environment.

To reproduce the current PoC:

1. **Build Asterinas kernel**
```bash
# make build BOOT_PROTOCOL=linux-efi-handover64
```

2. **Build ISO Installer Image**
```
# nix-build test/nix -A iso-image
...
Writing to 'stdio:/nix/store/nc2723j4gqkj6vc1v2nvflkq93qn5d00-nixos-minimal-25.05pre-git-x86_64-linux.iso/iso/nixos-minimal-25.05pre-git-x86_64-linux.iso' completed successfully.

/nix/store/nc2723j4gqkj6vc1v2nvflkq93qn5d00-nixos-minimal-25.05pre-git-x86_64-linux.iso
# cp stdio:/nix/store/nc2723j4gqkj6vc1v2nvflkq93qn5d00-nixos-minimal-25.05pre-git-x86_64-linux.iso/iso/nixos-minimal-25.05pre-git-x86_64-linux.iso ./
```

3. **Boot ISO Installer Image**
```bash
[...]

PR #2560, opened 2025-11-06
Add partition support for block device

cqs21:
This PR adds basic disk partition to support NixOS ISO installer image (#2356 ).
- Supports dynamic allocation of block device IDs;
- Adds block devices (including partition nodes) to the `/dev` directory;
- Moves block device mounting from the kernel to `/etc/profile.d/init.sh`;

If you want to test the partition functionality, you can use tools like `fdisk` to partition `test/build/ext2.img`, then `make run` to observe the partition nodes as follows::
```
~ # ls /dev/
console  null     pts      shm      tty0     vda      vda5     vdb
full     ptmx     random   tty      urandom  vda1     vda7     zero
```

### DIFF
--- a/kernel/comps/virtio/src/device/block/device.rs
+++ b/kernel/comps/virtio/src/device/block/device.rs
@@ -20,7 +20,6 @@ use aster_block::{
 };
 use aster_util::mem_obj_slice::Slice;
 use device_id::{DeviceId, MinorId};
-use id_alloc::IdAlloc;
 use log::{debug, info};
 use ostd::{
     arch::trap::TrapFrame,
@@ -35,6 +34,7 @@ use crate::{
         block::{ReqType, RespStatus},
         VirtioDeviceError,
     },
+    id_alloc::SyncIdAlloc,
     queue::VirtQueue,
     transport::{ConfigManager, VirtioTransport},
     VIRTIO_BLOCK_MAJOR_ID,
@@ -206,7 +206,7 @@ struct DeviceInner {
     transport: SpinLock<Box<dyn VirtioTransport>>,
     block_requests: Arc<DmaStream>,
     block_responses: Arc<DmaStream>,
-    id_allocator: SpinLock<IdAlloc>,
+    id_allocator: SyncIdAlloc,
     submitted_requests: SpinLock<BTreeMap<u16, SubmittedRequest>>,
 }
 
@@ -254,7 +254,7 @@ impl DeviceInner {
             transport: SpinLock::new(transport),
             block_requests,
             block_responses,
-            id_allocator: SpinLock::new(IdAlloc::with_capacity(Self::QUEUE_SIZE as usize)),
+            id_allocator: SyncIdAlloc::with_capacity(Self::QUEUE_SIZE as usize),
             submitted_requests: SpinLock::new(BTreeMap::new()),
         });
 
@@ -304,7 +304,7 @@ impl DeviceInner {
                 Slice::new(&self.block_responses, id * RESP_SIZE..(id + 1) * RESP_SIZE);
             resp_slice.sync().unwrap();
             let resp: BlockResp = resp_slice.read_val(0).unwrap();
-            self.id_allocator.lock().free(id);
+            self.id_allocator.dealloc(id);
             match RespStatus::try_from(resp.status).unwrap() {
                 RespStatus::Ok => {}
                 // FIXME: Return an error instead of triggering a kernel panic
@@ -337,7 +337,7 @@ impl DeviceInner {
 
     /// Reads data from the device, this function is non-blocking.
     fn read(&self, bio_request: BioRequest) {
-        let id = self.id_allocator.disable_irq().lock().alloc().unwrap();
+        let id = self.id_allocator.alloc();
         let req_slice = {
             let req_slice = Slice::new(
                 self.block_requests.clone(),
@@ -404,7 +404,7 @@ impl DeviceInner {
 
     /// Writes data to the device, this function is non-blocking.
     fn write(&self, bio_request: BioRequest) {
-        let id = self.id_allocator.disable_irq().lock().alloc().unwrap();
+        let id = self.id_allocator.alloc();
         let req_slice = {
             let req_slice = Slice::new(
                 self.block_requests.clone(),
@@ -478,7 +478,7 @@ impl DeviceInner {
             return;
         }
 
-        let id = self.id_allocator.disable_irq().lock().alloc().unwrap();
+        let id = self.id_allocator.alloc();
         let req_slice = {
             let req_slice = Slice::new(&self.block_requests, id * REQ_SIZE..(id + 1) * REQ_SIZE);
             let req = BlockReq {
--- a/kernel/comps/virtio/src/id_alloc.rs
+++ b/kernel/comps/virtio/src/id_alloc.rs
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use id_alloc::IdAlloc;
+use ostd::sync::{SpinLock, WaitQueue};
+
+/// A synchronized ID allocator that manages a fixed-size pool of IDs.
+///
+/// We say this ID allocator "synchronized" because
+/// its `alloc` method may be used by multiple threads to allocate IDs concurrently
+/// and it would block if no free IDs are available at the moment.
+/// Once an ID is no longer in use, the `dealloc` method may be called
+/// to return the ID to the ID allocator.
+///
+/// This ID allocator is designed for use by device drivers.
+/// The `alloc` method should only be called in the task context,
+/// whereas the `dealloc` method should only be used in the IRQ handler of a device driver.
+/// Failing to conform with the above requirement may result in deadlock.
+pub struct SyncIdAlloc {
+    wait_queue: WaitQueue,
+    id_allocator: SpinLock<IdAlloc>,
+}
+
+impl SyncIdAlloc {
+    /// Creates an allocator that may return IDs in the range of `0..capacity`.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            wait_queue: WaitQueue::new(),
+            id_allocator: SpinLock::new(IdAlloc::with_capacity(capacity)),
+        }
+    }
+
+    /// Allocates a new ID.
+    ///
+    /// This method must only be called in the task context.
+    /// It will block until an ID is free to allocate.
+    pub fn alloc(&self) -> usize {
+        self.wait_queue
+            .wait_until(|| self.id_allocator.disable_irq().lock().alloc())
+    }
+
+    /// Deallocates an ID.
+    ///
+    /// This method assumes that the caller is in the context of an IRQ handler.
+    ///
+    /// # Panics
+    ///
+    /// This method would panic if `id` is greater than or equal to the capacity of this ID allocator.
+    pub fn dealloc(&self, id: usize) {
+        self.id_allocator.disable_irq().lock().free(id);
+        self.wait_queue.wake_all();
+    }
+}
+
+impl core::fmt::Debug for SyncIdAlloc {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        f.debug_struct("SyncIdAlloc")
+            .field("id_allocator", &self.id_allocator.lock())
+            .finish_non_exhaustive()
+    }
+}
--- a/kernel/comps/virtio/src/lib.rs
+++ b/kernel/comps/virtio/src/lib.rs
@@ -30,6 +30,7 @@ use crate::transport::VirtioTransport;
 
 pub mod device;
 mod dma_buf;
+mod id_alloc;
 pub mod queue;
 mod transport;
 


OSDK.toml
```diff
@@ -64,6 +64,7 @@ qemu.args = """\
     -chardev stdio,id=mux,mux=on,signal=off,logfile=qemu.log \
     -drive if=none,format=raw,id=x0,file=./test/build/ext2.img \
     -drive if=none,format=raw,id=x1,file=./test/build/exfat.img \
+    -device virtio-blk-device,drive=x1 \
     -device virtio-blk-device,drive=x0 \
```
tatetian, line 71: Switch the two lines to be consistent with the ordering of the `-drive if=none,format=raw,id=xN` lines (where N is 0 or 1).

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    Err(Error::NotEnoughResources)
+}
+
+/// An allocator for block device IDs.
+#[derive(Debug)]
+pub struct DeviceIdAllocator {
+    major: u32,
+    allocated_minors: RwLock<BTreeSet<u32>>,
```
tatetian: Why a read-write lock is used when all user code calls `self.allocated_minors.write()`...

kernel/src/device/disk.rs
```diff
@@ -0,0 +1,96 @@
...
+        let total = reader.remain();
+        self.device.write(0, reader)?;
+        let remain = reader.remain();
+        Ok(total - remain)
+    }
+}
+
+impl Pollable for BlockFile {
+    fn poll(&self, _: IoEvents, _: Option<&mut PollHandle>) -> IoEvents {
+        IoEvents::empty()
```
tatetian: A block device should always be ready to read and write.

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,67 @@
...
+use alloc::collections::btree_set::BTreeSet;
+use core::sync::atomic::{AtomicU32, Ordering};
+
+use device_id::DeviceId;
+use ostd::sync::Mutex;
+
+use crate::Error;
+
+/// The upper limit for block device major IDs (not included).
+const MAJOR_MAX: u32 = 512;
```
tatetian: We need to define a type for major IDs because `0..=u32::max()` is not the valid range for major IDs. The special value of 0 means _null_. In Rust, we represent the concept of null with `None`, not a special value like 0. Without a proper type for major IDs, some semantics are unclear at the type level.

```rust
/// Registers a major ID.
///
/// If the requested ID is 0, a dynamic ID will be registered.
pub fn register_major(major: u32) -> Result<u32, Error> {
```

For example, one cannot tell from the type that `major == 0` means null, and the returned `u32` is always non-zero.

Another reason that we need to define a type for major IDs is that it is unclear the upper bound of major IDs. Here, an internal, const is defined. But how could any device drivers know the upper bound? This knowledge should be centralized into the type for major IDs.

```rust
/// The major component of a device ID.
///
/// A major ID is a non-zero, 12-bit integer, thus falling in the range of `1..(1u16 >> 12)`.
pub struct MajorId(NonZeroU16);

impl MajorId {
    pub fn new(raw: u16) -> Self { ... }

    pub fn to_raw(self) -> u16 { ... }

    pub const fn min() -> Self { ... }

    pub const fn max() -> Self { ... }
}
```

Then, we will split the `register_major` into two two APIs:

```rust
/// Acquires a major ID.
///
/// The returned `MajorIdOwner` object represents the ownership to the major ID.
/// Until the object is dropped, this major ID cannot be acquired via `acquire_major` or `allocate_major` again.
pub fn acquire_major(major: MajorId) -> Result<MajorIdOwner, Error> {
    ...
}

/// An owned major ID.
///
/// Each instances of 
pub struct MajorIdOwner(MajorId);

impl Drop for MajorIdOwner {
    // Note that this drop method replaces the current unregister method
    fn drop(&mut self) {
        todo!("release the major")
    }
}
```

and

```rust
/// Allocates a major ID.
///
/// The returned `MajorIdOwner` object represents the ownership to the major ID.
/// Until the object is dropped, this major ID cannot be acquired via `acquire_major` or `allocate_major` again.
pub fn allocate_major() -> Result<MajorIdOwner, Error> {
    ...
}
```

The current code seems to suggest that block devices have an even lower upper bound than the one imposed by the in-kernel representation.

```rust
/// The upper limit for block device major IDs (not included).
const MAJOR_MAX: u32 = 512;
```

Could you provide references to support this upper limit?

Similarly, we should consider defining a type for minor IDs.

```rust
/// The minor component of a device ID.
///
/// A minor ID is a 20-bit integer, thus falling in the range of `0..(1u32 >> 20)`.
pub struct MinorId(u32);

impl MinorId {
    pub fn new(raw: u32) -> Self { ... }

    pub fn to_raw(self) -> u32 { ... }

    pub const fn min() -> Self { ... }

    pub const fn max() -> Self { ... }
}
```

Introducing both `MajorId` and `MinorId` may sound a bit of heavyweight. But I want to point out that this is actually a convention that we already follow. The kernel code already includes a [`define_ranged_integer`](https://github.com/asterinas/asterinas/blob/bb48b3814f17f16633c9d9967fe7aa20022883f6/kernel/src/sched/nice.rs#L59) macro to define a new wrapper integer type to constraint the range of an integer type. It is used to accurately represent nice values. If this macro is extracted into `aster-util`, then the `device-id` crate can simply define the two types with several lines of code.

```rust
// In aster-util
define_ranged_integer!(pub, RangedI8, i8);
define_ranged_integer!(pub, RangedU8, u8);
define_ranged_integer!(pub, RangedI16, i16);
define_ranged_integer!(pub, RangedU16, u16);
define_ranged_integer!(pub, RangedI32, i32);
define_ranged_integer!(pub, RangedU32, u32);

// In device-id
pub type MajorId = RangedU16<1, MAX_MAJOR_ID>;
pub type Minor = RangedU32<0, MAX_MINOR_ID>;
```

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,67 @@
...
+
+const LAST_DYNAMIC_MAJOR: u32 = 254;
+
+static MAJORS: Mutex<BTreeSet<u32>> = Mutex::new(BTreeSet::new());
+
+/// Registers a major ID.
+///
+/// If the requested ID is 0, a dynamic ID will be registered.
+pub fn register_major(major: u32) -> Result<u32, Error> {
+    if major >= MAJOR_MAX {
```
tatetian: Move the definition of `MAJOR_MAX` into `register_major` because this is the only place that the const is used.

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// The major ID used for extended partitions when the number of disk partitions exceeds the standard limit.
+const EXTENDED_MAJOR: u32 = 259;
+
+/// The next available minor ID for extended partitions.
+static NEXT_EXTENDED_MINOR: AtomicU32 = AtomicU32::new(0);
+
+/// Allocates an extended device ID.
+pub fn allocate_extended_device_id() -> DeviceId {
```
tatetian: I suppose that this major ID `259` is reserved for virtio drivers specifically. It would be quite a surprise to me if _all_ block devices share this major ID when they have too many disk partitions. If my understanding is correct, then this virtio-specific API should not appear in the `block` component.

In addition, this allocation logic does not handle the possibility that extended device IDs are exhausted.

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -33,37 +38,76 @@ use crate::{
     },
     queue::VirtQueue,
     transport::{ConfigManager, VirtioTransport},
+    VIRTIO_BLOCK_MAJOR_ID,
 };
 
+/// The number of minor device numbers allocated for each virtio disk,
+/// including the whole disk and its partitions. If a disk has more than
+/// 16 partitions, the extended major:minor numbers will be assigned.
```
tatetian: Need to define this concept of extended device IDs properly. Saying "the extended major:minor numbers will be assigned" is not sufficient.

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,67 @@
...
+const MAJOR_MAX: u32 = 512;
+
+const LAST_DYNAMIC_MAJOR: u32 = 254;
+
+static MAJORS: Mutex<BTreeSet<u32>> = Mutex::new(BTreeSet::new());
+
+/// Registers a major ID.
+///
+/// If the requested ID is 0, a dynamic ID will be registered.
+pub fn register_major(major: u32) -> Result<u32, Error> {
```
tatetian: I have questions about our current `DeviceId` type.

In Linux, [`dev_t` is a 32-bit integer with the most significant 12 bits representing the major ID , the rest 20 bits doing the minor ID](https://github.com/torvalds/linux/blob/6fa9041b7177f6771817b95e83f6df17b147c8c6/include/linux/kdev_t.h#L7-L12). Why does Asterinas need to two 32-bit values rather two one 32-bit values?

kernel/comps/block/src/lib.rs
```diff
@@ -75,51 +98,69 @@ impl dyn BlockDevice {
...
 
-pub fn get_device(str: &str) -> Option<Arc<dyn BlockDevice>> {
-    COMPONENT
-        .get()
-        .unwrap()
-        .block_device_table
+/// Registers a new block device.
+pub fn register(device: Arc<dyn BlockDevice>) -> Result<(), Error> {
+    let mut registry = DEVICE_REGISTRY.lock();
+    let id = device.id().as_encoded_u64();
```
tatetian: The `as_encoded_u64` API is used for the wrong reason. This special encoding is only needed for the compatibility purpose when the kernel code is interacting with the userspace or persistent data structures of a FS. These users already assume an encoding for the `dev_t` type. But here, the user is another kernel consumer. So simply converting `DeviceId` to an integer (without the special encoding) is sufficient.

kernel/comps/block/src/lib.rs
```diff
@@ -75,51 +98,69 @@ impl dyn BlockDevice {
...
-    block_devs
-        .iter()
-        .map(|(name, device)| (name.clone(), device.clone()))
-        .collect()
+/// Looks up a block device of a given device ID.
+pub fn lookup(id: DeviceId) -> Option<Arc<dyn BlockDevice>> {
+    DEVICE_REGISTRY
+        .lock()
+        .values()
+        .find(|dev| dev.id() == id)
```
tatetian: Use the more efficient lookup, `DEVICE_REGISTRY.lock().get(&dev.id())`, rather than iterating all devices.

kernel/comps/block/src/lib.rs
```diff
@@ -75,51 +98,69 @@ impl dyn BlockDevice {
...
-impl Component {
-    pub fn init() -> Result<Self, ComponentInitError> {
-        Ok(Self {
-            block_device_table: SpinLock::new(BTreeMap::new()),
-        })
+#[init_component(process)]
+fn init_in_first_process() -> Result<(), component::ComponentInitError> {
+    let devices = collect_all();
+    for device in devices {
+        let partition_info = partition::parse(&device);
```
tatetian: What if a block device is not partitioned?

kernel/comps/block/src/partition.rs
```diff
@@ -0,0 +1,305 @@
...
+    name: [u8; 72],
+}
+
+impl GptEntry {
+    fn is_valid(&self) -> bool {
+        self.type_guid != [0; 16]
+    }
+}
+
+pub(super) fn parse(device: &Arc<dyn BlockDevice>) -> Vec<Option<PartitionInfo>> {
```
tatetian: What if a block device is not partitioned at all? Is it desirable to return an empty `Vec` and let the caller to set an empty `Vec` as the partitions?

kernel/comps/block/src/partition.rs
```diff
@@ -0,0 +1,305 @@
...
+        let Some(device) = self.device.upgrade() else {
+            bio.complete(BioStatus::IoError);
+            return Ok(());
+        };
+
+        let start_sid = bio.sid_range().start + self.info.start_sector();
+        let segments = Vec::from_iter(bio.segments().iter().cloned());
+        let new_bio = Bio::new(bio.type_(), start_sid, segments, None);
+
+        let Ok(status) = new_bio.submit_and_wait(device.as_ref()) else {
```
tatetian: It is unacceptable to block in the non-blocking `enqueue` method. And cloning the entire block I/O request just to offset the block addresses is too inefficient. Need to figure out a clever way to do this.

kernel/src/syscall/mount.rs
```diff
@@ -227,10 +229,16 @@ fn get_fs(
...
-        )
+        let path = devname.to_string_lossy();
+        let fs_path = FsPath::from_fd_and_path(AT_FDCWD, path.as_ref())?;
+        let path = ctx
+            .thread_local
+            .borrow_fs()
+            .resolver()
+            .read()
+            .lookup_no_follow(&fs_path)?;
+        let id = DeviceId::from_encoded_u64(path.metadata().rdev);
```
tatetian, line 244: Check `metadata.type_` to see if it is indeed a block device.

kernel/src/syscall/mount.rs
```diff
@@ -227,10 +229,16 @@ fn get_fs(
...
+        aster_block::lookup(id)
```
tatetian: If the FS type needs a backing block device but the lookup fails, then the mount operation will definitely fail. So I suppose that a proper error should be reported here.

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -101,6 +145,71 @@ impl aster_block::BlockDevice for BlockDevice {
...
+
+            let partition = Arc::new(PartitionNode::new(id, name, device, *info));
+            new_partitions.push(partition);
+        }
+
+        for partition in new_partitions.iter() {
+            let _ = aster_block::register(partition.clone());
+        }
+
+        *self.partitions.lock() = Some(new_partitions);
```
tatetian: To be thread safe, this lock needs to be held during the entire function.

kernel/comps/block/src/partition.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    partitions
+}
+
+#[derive(Debug)]
+pub struct PartitionNode {
+    id: DeviceId,
+    name: String,
+    device: Weak<dyn BlockDevice>,
```
tatetian: A partition should hold a strong reference to its parent disk because a partition would not function properly if the parent disk is dropped. On the other hand, a parent disk does not need its partition to be alive to work properly.

In addition, a partition needs to refer to its parent disk frequently, as shown by `self.device.upgrade()` in `enqueue` and `metadata`, while the parent disk seldomly needs to visit the partition objects. So holding a strong reference is good for performance.

tatetian (commented):
Thanks for adding this important feature!

OSDK.toml
cqs21: This ordering ensures that the kernel will find `ext2.img` first, so that `mount -t ext2 /dev/vda /ext2` won't fail. Otherwise, CI will have issues. For x86_64 architecture, `-device virtio-blk-pci,bus=pcie.0,addr=0x6,drive=x0` seems to be able to control the order by specifying `addr`, but riscv64 cannot specify these parameters.

kernel/comps/block/src/device_id.rs
cqs21: In glibc `dev_t` is a 64-bit quantity, with 32-bit major and minor numbers.

kernel/comps/block/src/device_id.rs
cqs21: > I suppose that this major ID 259 is reserved for virtio drivers specifically.

Actually, NVMe disks also use 259 as their major ID.

kernel/comps/block/src/device_id.rs
cqs21: > The current code seems to suggest that block devices have an even lower upper bound than the one imposed by the in-kernel representation.

See https://github.com/torvalds/linux/blob/6fa9041b7177f6771817b95e83f6df17b147c8c6/include/linux/blkdev.h#L1730

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use device_id::{DeviceId, MajorId, MinorId};
+use id_alloc::IdAlloc;
+use ostd::sync::Mutex;
+use spin::Once;
+
+use crate::Error;
+
+/// The upper limit for block device major IDs (not included).
+const MAJOR_MAX: u16 = 512;
```
tatetian: Add a reference to the Linux source code or documentation for this magic value.

In addition, usually a maximum value is inclusive. For example, `u32::max()` returns the maximum value that is within the valid range. So reduce the value by 1.

In addition, this upper bound may be interesting to the user. So it should be public.

```
/// The maximum value of the major device ID of a block device.
pub const MAX_MAJOR: u16 = 511;
```

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// Block devices that request a dynamic allocation of major ID will
+/// take numbers starting from 254 and downward.
+const LAST_DYNAMIC_MAJOR: u16 = 254;
```
tatetian, line 21: Provide a Linux reference to justify this number.

kernel/libs/device-id/src/lib.rs
```diff
@@ -60,11 +64,25 @@ impl DeviceId {
...
 }
+
+const MAX_MAJOR_ID: u16 = 0x0fff;
+
+const MAX_MINOR_ID: u32 = 0x000f_ffff;
+
+/// The major component of a device ID.
+///
+/// A major ID is a non-zero, 12-bit integer, thus falling in the range of `0..(1u16 << 12)`.
+pub type MajorId = RangedU16<0, MAX_MAJOR_ID>;
```
tatetian: > A major ID is a non-zero, 12-bit integer

But the range starts from 0, why?

```rust
pub fn acquire_major(major: MajorId) -> Result<MajorIdOwner, Error> {
    if major.get() == 0 || major.get() >= MAJOR_MAX {
        return Err(Error::InvalidArgs);
```

And why does the `acquire_major` function still treat 0 as a _null_ major ID?

kernel/libs/device-id/src/lib.rs
```diff
@@ -60,11 +64,25 @@ impl DeviceId {
...
     }
 }
+
+const MAX_MAJOR_ID: u16 = 0x0fff;
+
+const MAX_MINOR_ID: u32 = 0x000f_ffff;
+
+/// The major component of a device ID.
+///
+/// A major ID is a non-zero, 12-bit integer, thus falling in the range of `0..(1u16 << 12)`.
```
tatetian: Add a reference to Linux source code to justify the value range.

kernel/libs/device-id/src/lib.rs
```diff
@@ -60,11 +64,25 @@ impl DeviceId {
...
+pub type MajorId = RangedU16<0, MAX_MAJOR_ID>;
+
+/// The minor component of a device ID.
+///
+/// A minor ID is a 20-bit integer, thus falling in the range of `0..(1u32 << 20)`.
```
tatetian, line 100: Add a reference to Linux source code to justify the value range.

OSDK.toml
tatetian: So you want `ext2` to be the _first_ virtual disk, vda. But in the QEMU argument,  `-device virtio-blk-device,drive=x0` (ext2) is the _second_ block device. This is counter-intuitive.  Which exactly is the relationship between the ordering of QEMU arguments and that of block devices?

The bottom line is that since the ordering matters, a comment needs to be added to explain the requirement on the ordering. Otherwise, developers may mess up with the ordering in the future.

kernel/comps/block/src/partition.rs
```diff
@@ -0,0 +1,298 @@
...
+    name: String,
+    device: Arc<dyn BlockDevice>,
+    info: PartitionInfo,
+}
+
+impl BlockDevice for PartitionNode {
+    fn enqueue(&self, bio: SubmittedBio) -> Result<(), BioEnqueueError> {
+        // FIXME: This function should not be blocked by the new bio,
+        // and cloning the entire block I/O request is too inefficient.
+        let start_sid = bio.sid_range().start + self.info.start_sector();
```
tatetian: This is an important limitation and it really deserves to be fixed---in this PR. What solutions have you explored? And why all of them are not working?

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -101,6 +144,72 @@ impl aster_block::BlockDevice for BlockDevice {
...
+    fn drop(&mut self) {
+        use aster_block::BlockDevice;
+
+        if let Some(partitions) = self.partitions.lock().take() {
+            for partition in partitions {
+                let _ = aster_block::unregister(partition.id());
+            }
+        }
+
+        let _ = aster_block::unregister(self.id());
```
tatetian: This `drop` implementation will never work as expected.

```rust
static DEVICE_REGISTRY: Mutex<BTreeMap<u32, Arc<dyn BlockDevice>>> = Mutex::new(BTreeMap::new());
```

First, the block component holds a strong reference for every registered block device. Without unregistering a disk first, the drop method can never be triggered.

```rust
let _ = aster_block::unregister(partition.id());
```

Second, `PartitionNode` holds a strong reference to its parent disk. Similarly, without unregistering the partition first, this parent block device will have chance to get dropped.

In short, the entire logic in the drop method is pointless.

tatetian (commented):
Thanks for the revision. I am glad with the improvements.

kernel/libs/device-id/src/lib.rs
cqs21: > And why does the acquire_major function still treat 0 as a null major ID?

When I tried to define `MajorId = RangedU16<1, MAX_MAJOR_ID>`, the LTP `statx02` test encountered a panic. This happens because file system inodes make use of `major=0, minor=0`, and furthermore, the Linux kernel codebase contains many instances of `MKDEV(0, 0)`. Consequently, I believe `major=0` represents a valid and meaningful `MajorId` with legitimate uses, and should not be excluded from the range. But, real devices should not actively acquire this ID.

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -290,21 +290,22 @@ impl DeviceInner {
...
+            let (complete_request, id) = {
                 let mut queue = self.queue.lock();
                 let Ok((token, _)) = queue.pop_used() else {
                     return;
                 };
-                self.submitted_requests.lock().remove(&token).unwrap()
+                let complete_request = self.submitted_requests.lock().remove(&token).unwrap();
+                let id = complete_request.id as usize;
+                self.id_allocator.lock().free(id);
+                (complete_request, id)
```
tatetian: What exactly is the race condition regarding ID management that this commit is trying to fix?

It is the revised version that is vulnerable to race conditions in my opinion. After freeing `id`, a concurrent thread may reuse the `id`-th request and response slots in `self.block_requests` and `self.block_responses`, respectively. The block device may override the content in the `id`-th block response slot even _before_ the current thread copies back the content in the response slot:

```rust
let resp_slice =
                Slice::new(&self.block_responses, id * RESP_SIZE..(id + 1) * RESP_SIZE);
...
let resp: BlockResp = resp_slice.read_val(0).unwrap();
```

kernel/comps/block/src/device_id.rs
```diff
@@ -0,0 +1,117 @@
...
+
+static MAJORS: Mutex<BTreeSet<u16>> = Mutex::new(BTreeSet::new());
+
+/// Acquires a major ID.
+///
+/// The returned `MajorIdOwner` object represents the ownership to the major ID.
+/// Until the object is dropped, this major ID cannot be acquired via `acquire_major` or `allocate_major` again.
+pub fn acquire_major(major: MajorId) -> Result<MajorIdOwner, Error> {
+    if major.get() == 0 || major.get() > MAX_MAJOR {
```
tatetian: `major.get() == 0` is impossible

kernel/comps/block/src/partition.rs
```diff
@@ -0,0 +1,287 @@
...
+impl BlockDevice for PartitionNode {
+    fn enqueue(&self, bio: SubmittedBio) -> Result<(), BioEnqueueError> {
+        bio.set_sid_offset(self.info.start_sector());
+        self.device.enqueue(bio)
+    }
+
+    fn metadata(&self) -> BlockDeviceMeta {
+        let mut metadata = BlockDeviceMeta::default();
+        metadata.max_nr_segments_per_bio = self.device.metadata().max_nr_segments_per_bio;
+        metadata.nr_sectors = self.info.total_sectors() as usize;
```
tatetian, line 259: Simplify the logic to the following form:

```rust
let mut metadata = self.device.metadata();
metadata.nr_sectors = self.info.total_sectors() as usize;
```

The code practically says that a partition should inherit all the information of the parent metadata except for `total_sectors`.

kernel/comps/block/src/request_queue.rs
```diff
@@ -211,9 +211,13 @@ impl BioRequest {
 
 impl From<SubmittedBio> for BioRequest {
     fn from(bio: SubmittedBio) -> Self {
+        let mut sid_range = bio.sid_range().clone();
+        sid_range.start = sid_range.start + bio.sid_offset();
+        sid_range.end = sid_range.start + bio.sid_offset();
```
tatetian, line 227: Take into consideration `can_merge` and `merge_bio` logic when adding the support for `sid_offset`. The current code breaks the `can_merge` and `merge_bio` methods.

kernel/comps/virtio/src/device/block/device.rs
cqs21: The previous implementation of `PartitionNode::enqueue` would create new bio, and when there were too many requests that couldn't be processed in time, it would exhaust the id pool and cause a panic: 
```
let id = self.id_allocator.disable_irq().lock().alloc().unwrap();
```
Therefore, this fix attempts to move the allocation to the moment when there are actually free slots available. The current implementation doesn't seem to trigger this issue, so the proper implementation of this fix can be considered for introduction at a later time.

tatetian: > when there were too many requests that couldn't be processed in time, it would exhaust the id pool and cause a panic:

I think this is because the BIO merge logic was broken.

> so the proper implementation of this fix can be considered for introduction at a later time.

Yes, introducing disk partitions should have nothing to do with ID exhaustion. So it should be fixed in a separate PR.

In the meantime, you can simply add a TODO to increase the awareness of this limitation.

kernel/comps/block/src/request_queue.rs
```diff
@@ -139,7 +139,7 @@ impl Debug for BioRequestSingleQueue {
 pub struct BioRequest {
     /// The type of the I/O
     type_: BioType,
-    /// The range of target sectors on the device
+    /// The physical range of target sectors on the device
     sid_range: Range<Sid>,
```
tatetian, line 151: Need to update the documentation of `BioRequest` to reflect the new functionality, that is, the conversion from logical block addresses to physical ones in `BioRequest`.

```rust
/// A block I/O request dequeued from [`BioRequestSingleQueue`].
///
/// This `BioRequest` type is more friendly to storage medium than `SubmittedBio` for two reasons.
/// 
/// First, a `BioRequest` can represent a merged request over multiple `SubmittedBio`s
/// that (1) are of the same request type and (2) are contiguous in terms of target sectors.
/// This helps reduce the number of I/O requests submitted to the underlying storage medium.
///
/// Second, a `BioRequest` provides the physical sector addresses suitable for storage medium.
/// The sector addresses returned from `SubmittedBio::sid_range()` are logical ones:
/// they need to be adjsuted with `SubmittedBio::sid_offset()` to calculate the physical ones.
/// This calculation is handled internally by `BioRequest`.
/// One can simply call `BioRequest::sid_range()` to obtain the physical sector addresses.
#[derive(Debug)]
pub struct BioRequest {
    /// The type of the I/O
    type_: BioType,
    /// The physical range of target sectors on the device
    sid_range: Range<Sid>,
    // ..
}
```

tatetian (approved):

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -335,9 +345,19 @@ impl DeviceInner {
         info!("Virtio block device config space change");
     }
 
+    fn alloc_request_id(&self) -> usize {
+        self.wait_queue
+            .wait_until(|| self.id_allocator.disable_irq().lock().alloc())
```
tatetian: Bravos! I really like this solution. It looks quite neat.

To make it even better, I hope to extract the request ID allocator logic into a new abstraction so that `DeviceInner` does not have to care about the implementation details of ID allocation and synchronization at all. The new abstract is called `SyncIdAlloc`.

```rust
mod id_alloc {
    use id_alloc::IdAlloc;

    /// A synchronized ID allocator that manages a fixed-size pool of IDs.
    ///
    /// We say this ID allocator "synchronized" because
    /// its `alloc` method may be used by multiple threads to allocate IDs concurrently
    /// and it would block if no free IDs are available at the moment.
    /// Once an ID is no longer in use, the `dealloc` method may be called
    /// to return the ID to the ID allocator.
    ///
    /// This ID allocator is designed for use by device drivers.
    /// The `alloc` method should only be called in the task context,
    /// whereas the `dealloc` method should only be used in the IRQ handler of a device driver.
    /// Failing to conform with the above requirement may result in deadlock.
    pub struct SyncIdAlloc {
        wait_queue: WaitQueue,
        id_allocator: SpinLock<IdAlloc>,
    }

    impl SyncIdAlloc {
        /// Creates an allocator that may return IDs in the range of `0..capacity`.
        pub fn with_capacity(capacity: usize) -> Self {
            todo!()
        }

        /// Allocates a new ID.
        ///
        /// This method must only be called in the task context.
        /// It will block until an ID is free to allocate.
        pub fn alloc(&self) -> usize {
            todo!()
        }

        /// Deallocates an ID.
        ///
        /// This method assumes that the caller is in the context of an IRQ handler.
        ///
        /// # Panics
        ///
        /// This method would panic if `id` is greater than or equal to the capacity of this ID allocator.
        pub fn dealloc(&self, id: usize) {
            todo!()
        }
    }    
}
```

This `SyncIdAlloc` should be a separate module in the virtio component as it could be used by other virtio drivers to solve the common problem of request ID allocation. In the future, if more drivers can utilize this type, then we may migrate it to an even more visible location.

PR #2475, opened 2025-09-28
Add the device component

cqs21:
## Motivation
In the current system, Device ID (`major:minor`) assignment is often hardcoded, leading to inflexibility and potential conflicts. Additionally, there's no unified mechanism for exposing device-related information in sysfs (`/sys/devices`, `/sys/dev`) or linking devices to their `/dev` nodes.
The device component aims to solve the following problems:
- **Centralized Device ID Management**: Device major/minor numbers should be allocated dynamically rather than hardcoded.
- **Sysfs Integration**: Provide a standardized way to register devices under `/sys/devices` and create symlinks in `/sys/dev` that point to device entries.

## Overview

The design introduces several core components:

### Device ID Allocation

Device IDs are represented by a `DeviceId` struct containing major and minor numbers. Allocation is managed by `DeviceIdAllocator`, which handles ranges of minor numbers and tracks used IDs.

```rust
pub enum DeviceType {
    Block,
    Char,
    Other,
}
pub struct DeviceId {
    major: u32,
    minor: u32,
}
pub struct DeviceIdAllocator {
    pub type_: DeviceType,
    pub major: u32,
    pub minors: Range<u32>,
    used: RwLock<BTreeSet<u32>>,
}

impl DeviceIdAllocator {
    fn new(type_: DeviceType, major: u32, minors: Range<u32>) -> Self;
    pub fn allocate(&self, minor: u32) -> Option<DeviceId>;
    pub fn release(&self, minor: u32) -> bool;
}

pub fn register_device_ids(
    type_: DeviceType,
    major: u32,
    minors: Range<u32>,
) -> Result<DeviceIdAllocator, ostd::Error>
```

### Device Trait

Each device implements the Device trait:
```rust
pub trait Device: Any + Debug + Send + Sync {
    fn device_type(&self) -> DeviceType;
    fn device_id(&self) -> Option<DeviceId>;
    fn sysnode(&self) -> Arc<dyn SysBranchNode>;
}

pub fn add_device(device: Arc<dyn Device>);
pub fn del_device(device: Arc<dyn Device>);
pub fn get_device(type_: DeviceType, id: DeviceId) -> Option<Arc<dyn Device>>;
pub fn all_devices() -> impl Iterator<Item = Arc<dyn Device>>;
```

### Device File Interface
[...]

PR #2479, opened 2025-09-29
Add i8042 mouse

Cautreoxit:
This PR introduces an i8042 mouse driver for Asterinas. Building on the existing i8042 chip and keyboard support, it enables handling mouse events and forwarding them to the input subsystem.

Since this work depends on the input subsystem PR (https://github.com/asterinas/asterinas/pull/2364), which has not yet been merged, this PR is currently marked as a draft. I will make it ready for review once the dependency is merged.

One open question is naming: after adding both the i8042 mouse and keyboard, the current component name “keyboard” is no longer appropriate. We would appreciate any advice from the community on a better naming for this component.

PR #2561, opened 2025-11-06
Add evdev module

Cautreoxit:
This PR integrates the evdev module into Asterinas as an necessary component of the graphics stack. It introduces an input handler that receives input events from the input core and forwards them to the corresponding device node at `/dev/input/eventX`.

One issue is that the method used to create device nodes under `/dev` is not ideal. I’ve noticed that PR [#2475](https://github.com/asterinas/asterinas/pull/2475) aims to improve the device node creation process, but it has not yet been merged. As a result, I have adopted a temporary approach for now and have kept the rest of the code under review.

### DIFF
--- a/test/src/apps/devfs/evdev.c
+++ b/test/src/apps/devfs/evdev.c
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "../test.h"
+
+#define INPUT_DIR "/dev/input"
+#define MAX_EVDEV_DEVICES 16
+
+static int evdev_fds[MAX_EVDEV_DEVICES];
+static size_t evdev_count = 0;
+
+static void cleanup_open_fds(DIR *dir)
+{
+	if (dir) {
+		closedir(dir);
+	}
+	for (size_t i = 0; i < evdev_count; ++i) {
+		close(evdev_fds[i]);
+	}
+}
+
+FN_SETUP(open_evdev)
+{
+	DIR *dir = opendir(INPUT_DIR);
+	if (!dir) {
+		if (errno == ENOENT || errno == ENODEV || errno == ENXIO) {
+			fprintf(stderr, "evdev tests skipped: %s (%s)\n",
+				INPUT_DIR, strerror(errno));
+			exit(EXIT_SUCCESS);
+		}
+		fprintf(stderr,
+			"fatal error: setup_open_evdev: opendir('%s') failed: %s\n",
+			INPUT_DIR, strerror(errno));
+		exit(EXIT_FAILURE);
+	}
+
+	struct dirent *entry;
+	while ((entry = readdir(dir)) != NULL) {
+		if (strncmp(entry->d_name, "event", 5) != 0) {
+			continue;
+		}
+
+		if (evdev_count >= MAX_EVDEV_DEVICES) {
+			cleanup_open_fds(dir);
+			fprintf(stderr,
+				"fatal error: setup_open_evdev: too many event devices (max %d)\n",
+				MAX_EVDEV_DEVICES);
+			exit(EXIT_FAILURE);
+		}
+
+		char path[PATH_MAX];
+		int len = snprintf(path, sizeof(path), "%s/%s", INPUT_DIR,
+				   entry->d_name);
+		if (len < 0 || len >= (int)sizeof(path)) {
+			cleanup_open_fds(dir);
+			fprintf(stderr,
+				"fatal error: setup_open_evdev: path too long\n");
+			exit(EXIT_FAILURE);
+		}
+
+		int fd = open(path, O_RDONLY);
+		if (fd < 0) {
+			int saved = errno;
+			cleanup_open_fds(dir);
+			fprintf(stderr,
+				"fatal error: setup_open_evdev: open('%s') failed: %s\n",
+				path, strerror(saved));
+			exit(EXIT_FAILURE);
+		}
+
+		evdev_fds[evdev_count++] = fd;
+	}
+
+	closedir(dir);
+
+	if (evdev_count == 0) {
+		fprintf(stderr,
+			"evdev tests skipped: no event devices found in %s\n",
+			INPUT_DIR);
+		exit(EXIT_SUCCESS);
+	}
+}
+END_SETUP()
+
+// TODO: Currently this test is meaningless.
+// It is a placeholder for future tests (e.g., for ioctl commands).
+FN_TEST(open_close)
+{
+	TEST_RES(evdev_count > 0, _ret == 1);
+	for (size_t i = 0; i < evdev_count; ++i) {
+		TEST_RES(evdev_fds[i] >= 0, _ret == 1);
+	}
+}
+END_TEST()
+
+FN_SETUP(close_evdev)
+{
+	for (size_t i = 0; i < evdev_count; ++i) {
+		CHECK(close(evdev_fds[i]));
+	}
+}
+END_SETUP()
--- a/test/src/apps/scripts/fs.sh
+++ b/test/src/apps/scripts/fs.sh
@@ -106,3 +106,4 @@ file_io/access_err
 file_io/iovec_err
 devfs/full
 devfs/random
+devfs/evdev


Cautreoxit:
To better verify the correctness of the **evdev** and **mouse** (https://github.com/asterinas/asterinas/pull/2479) implementations, I found a classic benchmark tool: evtest (https://github.com/freedesktop-unofficial-mirror/evtest). It runs in user space to query device information under `/dev/input`, and then repeatedly polls the events from each `eventX` device. Therefore, if we operate the keyboard or mouse and immediately see the corresponding events printed by evtest, it proves that the entire path—from the driver, through the input core and evdev handler, to evdev—is functioning correctly.

After the test, I eventually found the whole input system could function correctly. Here is the output in my test:

> ~ # ls /dev/input/
event0  event1
~ # ls /test/evtest/
evtest
~ # ./test/evtest/evtest
No device specified, trying to scan all of /dev/input/event*
Available devices:
/dev/input/event0:      i8042_keyboard
/dev/input/event1:      i8042_mouse
Select the device event number [0-1]: 0
Input driver version is 1.0.1
Input device ID: bus 0x11 vendor 0x1 product 0x1 version 0x1
Input device name: "i8042_keyboard"
Supported events:
Properties:
Testing ... (interrupt to exit)
\***********************************************
  This device is grabbed by another process.
  No events are available to evtest while the
  other grab is active.
  In most cases, this is caused by an X driver,
  try VT-switching and re-run evtest again.
  Run the following command to see processes with
  an open fd on this device
 "fuser -v /dev/input/event0"
\***********************************************
Event: time 117.021132, type 1 (EV_KEY), code 30 (KEY_A), value 1
Event: time 117.021222, -------------- SYN_REPORT ------------
Event: time 117.112374, type 1 (EV_KEY), code 30 (KEY_A), value 0
Event: time 117.112386, -------------- SYN_REPORT ------------
Event: time 118.679418, type 1 (EV_KEY), code 31 (KEY_S), value 1
Event: time 118.679435, -------------- SYN_REPORT ------------
Event: time 118.774203, type 1 (EV_KEY), code 31 (KEY_S), value 0
Event: time 118.774214, -------------- SYN_REPORT ------------
Event: time 119.189590, type 1 (EV_KEY), code 32 (KEY_D), value 1
Event: time 119.189601, -------------- SYN_REPORT ------------
Event: time 119.285206, type 1 (EV_KEY), code 32 (KEY_D), value 0
Event: time 119.285217, -------------- SYN_REPORT ------------
Event: time 119.862053, type 1 (EV_KEY), code 33 (KEY_F), value 1
Event: time 119.862064, -------------- SYN_REPORT ------------
Event: time 119.958152, type 1 (EV_KEY), code 33 (KEY_F), value 0
Event: time 119.958162, -------------- SYN_REPORT ------------
Event: time 120.343289, type 1 (EV_KEY), code 34 (KEY_G), value 1
Event: time 120.343300, -------------- SYN_REPORT ------------
Event: time 120.438942, type 1 (EV_KEY), code 34 (KEY_G), value 0
Event: time 120.438953, -------------- SYN_REPORT ------------
Event: time 120.757863, type 1 (EV_KEY), code 35 (KEY_H), value 1
Event: time 120.757874, -------------- SYN_REPORT ------------
Event: time 120.886257, type 1 (EV_KEY), code 35 (KEY_H), value 0
Event: time 120.886269, -------------- SYN_REPORT ------------
Event: time 121.142067, type 1 (EV_KEY), code 36 (KEY_J), value 1
Event: time 121.142078, -------------- SYN_REPORT ------------
Event: time 121.270385, type 1 (EV_KEY), code 36 (KEY_J), value 0
Event: time 121.270396, -------------- SYN_REPORT ------------
Event: time 121.558708, type 1 (EV_KEY), code 37 (KEY_K), value 1
Event: time 121.558719, -------------- SYN_REPORT ------------
Event: time 121.655338, type 1 (EV_KEY), code 37 (KEY_K), value 0
Event: time 121.655349, -------------- SYN_REPORT ------------
Event: time 121.987802, type 1 (EV_KEY), code 38 (KEY_L), value 1
Event: time 121.987813, -------------- SYN_REPORT ------------
Event: time 122.084289, type 1 (EV_KEY), code 38 (KEY_L), value 0
Event: time 122.084300, -------------- SYN_REPORT ------------
~ # ./test/evtest/evtest
No device specified, trying to scan all of /dev/input/event*
Available devices:
/dev/input/event0:      i8042_keyboard
/dev/input/event1:      i8042_mouse
Select the device event number [0-1]: 1
Input driver version is 1.0.1
Input device ID: bus 0x11 vendor 0x2 product 0x1 version 0x1
Input device name: "i8042_mouse"
Supported events:
Properties:
Testing ... (interrupt to exit)
\***********************************************
  This device is grabbed by another process.
  No events are available to evtest while the
  other grab is active.
  In most cases, this is caused by an X driver,
  try VT-switching and re-run evtest again.
  Run the following command to see processes with
  an open fd on this device
 "fuser -v /dev/input/event1"
\***********************************************
Event: time 136.194489, type 2 (EV_REL), code 0 (REL_X), value 127
Event: time 136.194506, type 2 (EV_REL), code 1 (REL_Y), value 127
Event: time 136.194509, -------------- SYN_REPORT ------------
Event: time 136.194573, type 2 (EV_REL), code 0 (REL_X), value 127
Event: time 136.194575, type 2 (EV_REL), code 1 (REL_Y), value 127
Event: time 136.194577, -------------- SYN_REPORT ------------
Event: time 136.194664, type 2 (EV_REL), code 0 (REL_X), value 53
Event: time 136.194667, type 2 (EV_REL), code 1 (REL_Y), value 127
Event: time 136.194669, -------------- SYN_REPORT ------------
Event: time 136.194730, type 2 (EV_REL), code 1 (REL_Y), value 127
Event: time 136.194732, -------------- SYN_REPORT ------------
Event: time 136.251998, type 2 (EV_REL), code 0 (REL_X), value 1
Event: time 136.252010, type 2 (EV_REL), code 1 (REL_Y), value 75
Event: time 136.252012, -------------- SYN_REPORT ------------
Event: time 136.252074, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.252076, -------------- SYN_REPORT ------------
Event: time 136.269724, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.269736, -------------- SYN_REPORT ------------
Event: time 136.269801, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.269804, -------------- SYN_REPORT ------------
Event: time 136.269864, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.269866, -------------- SYN_REPORT ------------
Event: time 136.295180, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.295191, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.295193, -------------- SYN_REPORT ------------
Event: time 136.295252, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.295254, -------------- SYN_REPORT ------------
Event: time 136.295313, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.295315, type 2 (EV_REL), code 1 (REL_Y), value -1
Event: time 136.295317, -------------- SYN_REPORT ------------
Event: time 136.295374, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.295376, type 2 (EV_REL), code 1 (REL_Y), value -2
Event: time 136.295378, -------------- SYN_REPORT ------------
Event: time 136.300310, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.300322, type 2 (EV_REL), code 1 (REL_Y), value -6
Event: time 136.300324, -------------- SYN_REPORT ------------
Event: time 136.317888, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.317899, type 2 (EV_REL), code 1 (REL_Y), value -4
Event: time 136.317901, -------------- SYN_REPORT ------------
Event: time 136.317959, type 2 (EV_REL), code 1 (REL_Y), value -2
Event: time 136.317962, -------------- SYN_REPORT ------------
Event: time 136.318020, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.318022, type 2 (EV_REL), code 1 (REL_Y), value -2
Event: time 136.318024, -------------- SYN_REPORT ------------
Event: time 136.318079, type 2 (EV_REL), code 1 (REL_Y), value -4
Event: time 136.318081, -------------- SYN_REPORT ------------
Event: time 136.339334, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.339345, type 2 (EV_REL), code 1 (REL_Y), value -3
Event: time 136.339348, -------------- SYN_REPORT ------------
Event: time 136.339412, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.339414, type 2 (EV_REL), code 1 (REL_Y), value -5
Event: time 136.339416, -------------- SYN_REPORT ------------
Event: time 136.339479, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.339481, type 2 (EV_REL), code 1 (REL_Y), value -4
Event: time 136.339483, -------------- SYN_REPORT ------------
Event: time 136.339544, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.339547, type 2 (EV_REL), code 1 (REL_Y), value -4
Event: time 136.339549, -------------- SYN_REPORT ------------
Event: time 136.343186, type 2 (EV_REL), code 0 (REL_X), value -3
Event: time 136.343196, type 2 (EV_REL), code 1 (REL_Y), value -18
Event: time 136.343198, -------------- SYN_REPORT ------------
Event: time 136.365750, type 2 (EV_REL), code 0 (REL_X), value -3
Event: time 136.365762, type 2 (EV_REL), code 1 (REL_Y), value -5
Event: time 136.365764, -------------- SYN_REPORT ------------
Event: time 136.365829, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.365832, type 2 (EV_REL), code 1 (REL_Y), value -9
Event: time 136.365834, -------------- SYN_REPORT ------------
Event: time 136.365897, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.365900, type 2 (EV_REL), code 1 (REL_Y), value -7
Event: time 136.365902, -------------- SYN_REPORT ------------
Event: time 136.365962, type 2 (EV_REL), code 0 (REL_X), value -1
Event: time 136.365965, type 2 (EV_REL), code 1 (REL_Y), value -9
Event: time 136.365967, -------------- SYN_REPORT ------------
Event: time 136.387456, type 2 (EV_REL), code 0 (REL_X), value -5
Event: time 136.387468, type 2 (EV_REL), code 1 (REL_Y), value -15
Event: time 136.387470, -------------- SYN_REPORT ------------
Event: time 136.387528, type 2 (EV_REL), code 0 (REL_X), value -3
Event: time 136.387531, type 2 (EV_REL), code 1 (REL_Y), value -9
Event: time 136.387533, -------------- SYN_REPORT ------------
Event: time 136.387590, type 2 (EV_REL), code 0 (REL_X), value -3
Event: time 136.387593, type 2 (EV_REL), code 1 (REL_Y), value -8
Event: time 136.387594, -------------- SYN_REPORT ------------
Event: time 136.387673, type 2 (EV_REL), code 0 (REL_X), value -4
Event: time 136.387677, type 2 (EV_REL), code 1 (REL_Y), value -9
Event: time 136.387679, -------------- SYN_REPORT ------------
Event: time 136.391268, type 2 (EV_REL), code 0 (REL_X), value -11
Event: time 136.391280, type 2 (EV_REL), code 1 (REL_Y), value -22
Event: time 136.391282, -------------- SYN_REPORT ------------

Cautreoxit:
Now this PR includes another commit, to make it convenient to test with evtest. Feel free to give a try.

test/nix/initramfs.nix
```diff
@@ -10,6 +10,8 @@ let
     path = "/lib/x86_64-linux-gnu";
   };
   resolv_conf = pkgs.callPackage ./resolv_conf.nix { dnsServer = dnsServer; };
+  # If you want to use evtest, replace line 15 with line 14 and uncomment line 30.
+  # all_pkgs = [ busybox pkgs.evtest etc resolv_conf ]
```
junyang-zh: If it can work in a headless environment and runs fast I think we can make it into the CI?

junyang-zh: Oh I see you need a mouse and a keyboard to work.

TinaZhangZW: Although evtest is not suitable to run in CI, it is useful for development and debugging, and similar tools may appear in the future. We should design a management mechanism for these extra test cases—e.g. add an extraTestPkgs flag (or list) to host development-oriented test packages. That way they can be excluded from builds/releases to keep CI lean, while remaining easy to enable and maintain locally or on development branches. 

For example:
{ lib, ..., pkgs,  extraTestPkgs ? []}:
    ...
   all_pkgs = base_pkgs
    ++ lib.optionals (apps != null) [ apps.package ]
    ++ lib.optionals (benchmark != null) [ benchmark.package ]
    ++ lib.optionals (extraTestPkgs !=null) [
        # Tools for development and debugging
        pkgs.evtest
    ]

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,585 @@
...
+                Errno::ENODEV,
+                "evdev device not found in registry",
+            )))
+        }
+    }
+}
+
+impl Pollable for Evdev {
+    fn poll(&self, _mask: IoEvents, _poller: Option<&mut PollHandle>) -> IoEvents {
+        // This shouldn't be called directly.
```
TinaZhangZW: That’s incorrect. User-space relies on evdev’s poll() interface to receive events—see libevdev_has_event_pending() in libevdev for details.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,585 @@
...
+        // Evdev devices are read-only, so never indicate writable.
```
TinaZhangZW: The annotation sounds ambiguous: is poll truly never expected to be called, or does it intentionally always return empty? If poll ever gets called expecting readable readiness, returning empty will hide bugs and cause callers to wait forever. So, make intent explicit. If truly unreachable, assert in debug builds so misuse is visible.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,585 @@
...
+            .copy_from_slice(&self.type_.to_le_bytes());
+        bytes[2 * core::mem::size_of::<u64>() + core::mem::size_of::<u16>()
+            ..2 * core::mem::size_of::<u64>() + 2 * core::mem::size_of::<u16>()]
+            .copy_from_slice(&self.code.to_le_bytes());
+        bytes[2 * core::mem::size_of::<u64>() + 2 * core::mem::size_of::<u16>()..]
+            .copy_from_slice(&self.value.to_le_bytes());
+        bytes
+    }
+}
+
```
TinaZhangZW: How about using a small offset variable to make the code easy for reading:
```
pub fn to_bytes(self) -> [u8; core::mem::size_of::<EvdevEvent>()] {
    const SZ: usize = core::mem::size_of::<EvdevEvent>();
    let mut bytes = [0u8; SZ];
    let mut off = 0;

    let sec_bytes = self.sec.to_le_bytes();
    bytes[off..off + sec_bytes.len()].copy_from_slice(&sec_bytes);
    off += sec_bytes.len();

    let usec_bytes = self.usec.to_le_bytes();
    bytes[off..off + usec_bytes.len()].copy_from_slice(&usec_bytes);
    off += usec_bytes.len();

    let type_bytes = self.type_.to_le_bytes();
    bytes[off..off + type_bytes.len()].copy_from_slice(&type_bytes);
    off += type_bytes.len();

    let code_bytes = self.code.to_le_bytes();
    bytes[off..off + code_bytes.len()].copy_from_slice(&code_bytes);
    off += code_bytes.len();

    let value_bytes = self.value.to_le_bytes();
    bytes[off..off + value_bytes.len()].copy_from_slice(&value_bytes);

    bytes
}
```

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,585 @@
...
+                        // Try to push the original event.
+                        if let Some(()) = producer.push(timed_event) {
+                            client.increment_event_count();
+
+                            // Check if this is a SYN_REPORT event.
+                            if self.is_syn_report_event(event) {
+                                client.increment_packet_count();
+                            }
+                        }
+                    }
```
TinaZhangZW: If pushing a SYN_DROPPED event fails, do not ignore it silently. Emit a warning (e.g. log::warn!).

TinaZhangZW (commented):
Thank you for the PR. Please see my comments inline.

test/nix/initramfs.nix
```diff
@@ -10,10 +10,13 @@ let
     path = "/lib/x86_64-linux-gnu";
   };
   resolv_conf = pkgs.callPackage ./resolv_conf.nix { dnsServer = dnsServer; };
+  # Toggle to include evtest binaries in initramfs.
+  evtest_involved = false;
```
tatetian: "involved" is a little bit vague. I suggest renaming the variable to `is_evtest_included`.

```nix
  # Whether the initramfs should include evtest, a common tool to debug input devices (`/dev/input/eventX`)
  is_evtest_included = false;
```

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,608 @@
...
+}
+
+impl InputHandler for EvdevHandler {
+    fn handle_events(&self, events: &[InputEvent]) {
+        self.evdev.pass_events(events);
+    }
+}
+
+/// Initializes evdev support in the kernel device system.
+pub fn init(fs_resolver: &FsResolver) -> Result<()> {
```
tatetian: Narrow down the visibility to `pub(super)`.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    fn id(&self) -> DeviceId {
+        // The same value as Linux.
+        DeviceId::new(13, self.minor)
+    }
+
+    fn open(&self) -> Option<Result<Arc<dyn FileIo>>> {
+        let devices = EVDEV_DEVICES.lock();
+        if let Some((_, evdev)) = devices.iter().find(|(minor, _)| *minor == self.minor) {
```
tatetian: To improve readability, revise the if statement to the following form:

```rust
        let device_entry = devices.iter().find(|(minor, _)| *minor == self.minor);
        let Some((_, evdev)) = device_entry else {
            // Create a new client for this evdev device.
            return Some(Err(Error::with_message(
                Errno::ENODEV,
                "evdev device not found in registry",
            )));
        };

        let evdev_client = evdev.create_client(EVDEV_BUFFER_SIZE);
        Some(evdev_client)
```

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,608 @@
...
+            minor,
+            open: SpinLock::new(0),
+            device,
+            client_list: SpinLock::new(Vec::new()),
+            device_name,
+        }
+    }
+
+    /// Checks if this evdev device is associated with the given input device.
+    pub fn matches_input_device(&self, input_device: &Arc<dyn InputDevice>) -> bool {
```
tatetian: All (or almost all) the methods of `Evdev` should have been given narrow visibilities. These methods should only be used within this module.  Use `pub(self)` to mark the methods (like this one) that are used by other code in this module.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,608 @@
...
+        Arc::ptr_eq(&self.device, input_device)
+    }
+
+    /// Adds a client to this evdev device.
+    pub fn attach_client(&self, client: Arc<EvdevClient>, producer: RbProducer<EvdevEvent>) {
```
tatetian: This method should have been marked as a private method as it is only used by the `Evdev` type itself, in the `create_client` method.

test/src/apps/devfs/evdev.c
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	struct dirent *entry;
+	while ((entry = readdir(dir)) != NULL) {
+		if (strncmp(entry->d_name, "event", 5) != 0) {
+			continue;
+		}
+
+		if (evdev_count >= MAX_EVDEV_DEVICES) {
+			closedir(dir);
```
TinaZhangZW: This cleanup code only closes the directory file descriptor. It may leak event event file descriptors which have been got. 

Consider the following reference code::
```
static void cleanup_open_fds(DIR *dir) {
    if (dir) {
        closedir(dir);
    }
    for (size_t i = 0; i < evdev_count; ++i) {
        close(evdev_fds[i]);
    }
}

FN_SETUP(open_evdev)
{
    ...
    if (evdev_count >= MAX_EVDEV_DEVICES) {
            cleanup_open_fds(dir);
            fprintf(stderr,
                "fatal error: setup_open_evdev: too many event devices (max %d)\n",
                MAX_EVDEV_DEVICES);
            exit(EXIT_FAILURE);
    }
    ...
            if (len < 0 || len >= (int)sizeof(path)) {
            cleanup_open_fds(dir);
            fprintf(stderr,
                "fatal error: setup_open_evdev: path too long\n");
            exit(EXIT_FAILURE);
        }
    ...
            int fd = open(path, O_RDONLY);
        if (fd < 0) {
            int saved = errno;
            cleanup_open_fds(dir);
            fprintf(stderr,
                "fatal error: setup_open_evdev: open('%s') failed: %s\n",
                path, strerror(saved));
            exit(EXIT_FAILURE);
        }
    ...
}
```

test/src/apps/devfs/evdev.c
```diff
@@ -0,0 +1,98 @@
...
+	if (evdev_count == 0) {
+		fprintf(stderr,
+			"evdev tests skipped: no event devices found in %s\n",
+			INPUT_DIR);
+		exit(EXIT_SUCCESS);
+	}
+}
+END_SETUP()
+
+FN_TEST(open_close)
```
TinaZhangZW, line 95: This test case appears redundant. The open_evdev setup function already guarantees that evdev_count > 0 and that all file descriptors are valid, since it would exit() on any failure.

If this is a placeholder for future tests (e.g., for ioctl commands), I'd suggest replacing the current implementation with a TODO comment to make that intent clear. Otherwise, this test function can be removed.

TinaZhangZW (commented):
Thanks for the update. Overall, this is good. I've left inline comments, mostly on the CI test."

Cautreoxit:
Thanks for your comments! I have addressed these two issues. For the second issue, it indeed is meaningless and works as a placeholder for further tests. I would add those tests in ioctl PR (https://github.com/asterinas/asterinas/pull/2582).

kernel/src/device/mod.rs
```diff
@@ -2,6 +2,7 @@
 
 mod char;
 mod disk;
+mod evdev;
 mod mem;
```
tatetian, line 6: Check out the [new version of the memory device](https://github.com/asterinas/asterinas/blob/main/kernel/src/device/mem/mod.rs) and implement evdev devices in the same pattern as the memory devices.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+impl Drop for EvdevClient {
+    fn drop(&mut self) {
+        if let Some(evdev) = self.evdev.upgrade() {
+            evdev.detach_client();
+            evdev.close_device();
+        }
+    }
+}
+
+pub struct Evdev {
```
tatetian: To follow the convention established in the new memory devices, the type to represent device files at `/dev/input/eventX` should be named `EvdevDevice` (rather than `Evdev`) while an open file of an evdev device should be called `EvdevFile` (rather than `EvdevClient`). The two types should implement a different set of traits.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// Minor device number.
+    minor: u32,
+    /// Reference count of open clients.
+    open: SpinLock<u32>,
```
tatetian: Using `AtomicU32` is sufficient, isn't it?

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+        // Create evdev device.
+        let evdev = Arc::new(Evdev::new(minor, dev.clone()));
+        let device_path = format!("input/event{}", minor);
+
+        // Create the device node.
+        let fs_resolver = FS_RESOLVER
+            .lock()
+            .clone()
+            .ok_or(ConnectError::InternalError)?;
+        match add_node(evdev.clone(), &device_path, &fs_resolver) {
```
tatetian: With the new char device infra, you only need to register the new `Evdev` instance as a char device at this point of time. The char device infra can take care of creating device inodes under `/dev` for you. This also means that `FS_RESOLVER` is no longer needed.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+
+pub struct Evdev {
+    /// Minor device number.
+    minor: u32,
+    /// Reference count of open clients.
+    open: SpinLock<u32>,
+    /// Input device associated with this evdev.
+    device: Arc<dyn InputDevice>,
+    /// List of clients with their producers.
+    client_list: SpinLock<Vec<(Weak<EvdevClient>, RbProducer<EvdevEvent>)>>,
```
tatetian: You need to be very careful with a spinlock that is used in both the interrupt and task contexts. When opening/closing an evdev device, this client list will be updated in the task context by acquiring this lock. But when this happen, an input event may occur and thus trigger the the interrupt handler. The handler eventually calls `Evdev::pass_events`, which takes the spinlock--again. This leads to deadlock.

To fix the deadlock, call `disable_irq` in the `attach_client` and `detach_client` methods:

```rust
    /// Adds a client to this evdev device.
    pub fn attach_client(&self, client: Weak<EvdevClient>, producer: RbProducer<EvdevEvent>) {
        let mut client_list = self.client_list.disable_irq().lock();
        client_list.push((client, producer));
    }

    /// Removes closed clients from this evdev device.
    pub fn detach_client(&self) {
        let mut client_list = self.client_list.disable_irq().lock();
        client_list.retain(|(client, _)| client.strong_count() > 0);
    }
```

Furthermore, add some comments to this field as a reminder to prevent future modifications from introducing such bugs.

```rust
    /// List of clients with their producers.
    ///
    /// # Deadlock
    ///
    /// This lock is acquired in both the task and interrupt contexts.
    /// We must make sure that this lock is taken with the local IRQs disabled.
    /// Otherwise, we would be vulnerable to deadlock.
    client_list: SpinLock<Vec<(Weak<EvdevClient>, RbProducer<EvdevEvent>)>>,
```

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+        let mut client_list = self.client_list.lock();
+        client_list.retain(|(client, _)| client.strong_count() > 0);
+    }
+
+    /// Distributes events to all clients.
+    pub fn pass_events(&self, events: &[InputEvent]) {
+        let mut client_list = self.client_list.lock();
+
+        // Send events to all clients using their producers.
+        client_list.retain_mut(|(client_weak, producer)| {
```
tatetian: I can't see why we want to try removing dropped clients during the time-critical, highly-frequent operation of interrupt handling. When a client is dropped, it can remove itself from the client list. This is a much better timing to update the list.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+        Self {
+            sec: time.as_secs(),
+            usec: time.subsec_micros() as u64,
+            type_,
+            code,
+            value,
+        }
+    }
+
+    pub fn to_bytes(self) -> [u8; core::mem::size_of::<EvdevEvent>()] {
```
tatetian: I have some reservations about this `to_bytes` method.

First of all, it introduces one extra data copy per event.

```rust
        for event in events {
            let event_bytes = event.to_bytes();
            writer.write_fallible(&mut event_bytes.as_slice().into())?;
            ...
        }
```

The above code illustrates how the caller uses it: first copying the event to the stack, then copying it into the writer. Why not copy the event directly to the writer?

Second, the serialization code in this method specifies the endianness.

```rust
        let sec_bytes = self.sec.to_le_bytes();
        bytes[off..off + sec_bytes.len()].copy_from_slice(&sec_bytes);
        off += sec_bytes.len();
```

This is completely unnecessary because both the producer (the kernel) and consumer (the userspace) of these serialized event records run on the same machine, using the same endianness. In fact, it is wrong on big-endian CPUs because the userspace assume big-endianness but the kernel serializes the data in little-endianness.

Luckily, the correct implementation is simple: just remove the `to_bytes` method. `EvdevEvent` implements `Pod` and `VmWriter` provides a [`write_val`](https://asterinas.github.io/api-docs-nightly/ostd/mm/io/struct.VmWriter.html#method.write_val-1) method that takes any `Pod` value. Thus, the event dumping loop can be simplified to the following form:

```rust
        for event in events {
            writer.write_val(event)?;
            ...
        }
```

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+
+        bytes
+    }
+}
+
+pub struct EvdevClient {
+    /// Consumer for reading events.
+    consumer: Mutex<RbConsumer<EvdevEvent>>,
+    /// Client-specific clock type.
+    clock_type: AtomicU32,
```
tatetian: Representing a clock ID as an (atomic) integer is unreadable and cumbersome. Use the [`define_atomic_version_of_integer_like_type`](https://github.com/asterinas/asterinas/blob/76c7752130316e9ce605aada13d49a09463ac0f7/kernel/libs/atomic-integer-wrapper/src/lib.rs#L3) macro to define an atomic version of `ClockId`, `AtomicClockId`.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+                // Try to push event to the buffer.
+                if let Some(()) = producer.push(timed_event) {
+                    client.increment_event_count();
+
+                    // Check if this is a SYN_REPORT event
+                    if self.is_syn_report_event(event) {
+                        client.increment_packet_count();
+                    }
+                } else {
+                    // TODO: In Linux's implementation, when the buffer is full, evdev will pop the two
```
tatetian: This is a very good writing 👍

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+            self.pollee.invalidate();
+        }
+    }
+
+    /// Processes events and write them to the writer.
+    /// Returns the total number of bytes written, or EAGAIN if no events available.
+    fn process_events(&self, max_events: usize, writer: &mut VmWriter) -> Result<usize> {
+        const EVENT_SIZE: usize = core::mem::size_of::<EvdevEvent>();
+
+        let events = self.read_events(max_events);
```
tatetian: Why not dump events directly into `writer`? This is more efficient: no heap allocation (due to `Vec`) and no extra data copies.

kernel/src/device/evdev.rs
```diff
@@ -0,0 +1,623 @@
...
+        for _ in 0..count {
+            if let Some(event) = consumer.pop() {
+                // Check if this is a SYN_REPORT event.
+                let is_syn_report = self.is_syn_report_event(&event);
+
+                events.push(event);
+                self.decrement_event_count();
+
+                if is_syn_report {
+                    self.decrement_packet_count();
```
tatetian: I suppose that you also want to  check `SYN_DROPPED` here. The `pass_events` method has counted `SYN_DROPPED` as a packet.

```rust
                    let dropped_event = EvdevEvent::from_event_and_time(
                        &InputEvent::from_sync_event(SynEvent::Dropped),
                        time,
                    );

                    // Try to push SYN_DROPPED event.
                    if let Some(()) = producer.push(dropped_event) {
                        client.increment_event_count();
                        client.increment_packet_count();
```

So we obviously want to decrease the counter if seeing a `SYN_DROPPED` event. Otherwise, the packet count will always be positive once a `SYN_DROPPED` is generated.

tatetian (commented):
This PR is in a great shape. Overall, I like it. Good job 👍

Cautreoxit:
Thanks for your reasonable advice! I have fixed these issues, and I would appreciate for further feedbacks. 

There is only one issue I would like to mention: the function `init_in_first_process` in `kernel/src/device/char.rs` appears to overlook cases where nodes are created under specific classes. For example, `eventX` nodes are created in `/dev/input/`. I have addressed this issue by adding a fix for this function in this PR.

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Event device (evdev) support.
+//!
+//! Character device with major number 13. The minor numbers are dynamically allocated.
+//! Devices appear as `/dev/input/eventX` where X is the minor number.
```
tatetian, line 6: Good writing 👍

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// List of clients with their producers.
+    ///
+    /// # Deadlock Prevention
+    ///
+    /// This lock is acquired in both the task and interrupt contexts.
+    /// We must make sure that this lock is taken with the local IRQs disabled.
+    /// Otherwise, we would be vulnerable to deadlock.
+    client_list: SpinLock<Vec<(Weak<EvdevFile>, RbProducer<EvdevEvent>)>>,
+    /// Device name.
+    device_name: String,
```
tatetian: This information can be cheaply queried from the `device` field, right? So we don't need to maintain a copy here.

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+};
+
+/// Major device number for evdev devices.
+const EVDEV_MAJOR_ID: u16 = 13;
+
+/// Global minor number allocator for evdev devices.
+static EVDEV_MINOR_COUNTER: AtomicU32 = AtomicU32::new(0);
+
+/// Global registry of evdev devices.
+static EVDEV_DEVICES: SpinLock<Vec<(u32, Arc<EvdevDevice>)>> = SpinLock::new(Vec::new());
```
tatetian: This registry should have been modeled with a key-value map data structure such as `BTreeMap` for efficiency.

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+
+pub struct EvdevDevice {
+    /// Minor device number.
+    minor: u32,
```
tatetian: After this PR was submitted, the main branch has included device ID types of `DeviceId`, `MajorId`, and `MinorId`.  Replace  `u32` with the appropriate device ID types to make the semantics more clear.

---

This struct also includes an `id: DeviceId` field, whose `DeviceId::minor` method provide a getter method that is as efficient as accessing a `u32` directly. So this field can be removed.

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Reference count of open clients.
```
tatetian: The original client type has been renamed to `EvdevFile`. So there is no "clients" any more. Refer to them as opened evdev file.

kernel/src/device/evdev/file.rs
```diff
@@ -0,0 +1,270 @@
...
+            sec: time.as_secs(),
+            usec: time.subsec_micros() as u64,
+            type_,
+            code,
+            value,
+        }
+    }
+}
+
+pub struct EvdevFile {
```
tatetian, line 72:
```rust
/// An opened file from an evdev device (`EvdevDevice`).
```

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Input device associated with this evdev.
+    device: Arc<dyn InputDevice>,
+    /// List of clients with their producers.
+    ///
+    /// # Deadlock Prevention
+    ///
+    /// This lock is acquired in both the task and interrupt contexts.
+    /// We must make sure that this lock is taken with the local IRQs disabled.
+    /// Otherwise, we would be vulnerable to deadlock.
+    client_list: SpinLock<Vec<(Weak<EvdevFile>, RbProducer<EvdevEvent>)>>,
```
tatetian: Again, the concept of "client" no longer exists. Rename all references to "clients" as "opened evdev files". For example, this field should be named to "opened_files" or "open_table".

kernel/src/device/evdev/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+static EVDEV_MINOR_COUNTER: AtomicU32 = AtomicU32::new(0);
+
+/// Global registry of evdev devices.
+static EVDEV_DEVICES: SpinLock<Vec<(u32, Arc<EvdevDevice>)>> = SpinLock::new(Vec::new());
+
+pub struct EvdevDevice {
+    /// Minor device number.
+    minor: u32,
+    /// Reference count of open clients.
+    open: AtomicU32,
```
tatetian: Is this field really needed? It seems that this counter is only useful for debug purpose. If so, the same information can be obtained by counting the length of `client_list`. So this field can be removed.

kernel/src/device/evdev/file.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// Consumer for reading events.
+    consumer: Mutex<RbConsumer<EvdevEvent>>,
+    /// Client-specific clock type.
+    pub(super) clock_type: AtomicClockId,
```
tatetian: Do not break object encapsulation unless it is absolutely necessary. Make this field private and add a simple getter method can avoid breaking object encapsulation:

```rust
impl EvdevFile {
    pub(super) fn clock_id(&self) -> ClockId { ... }
}
```

Rename this field to `clock_id`. Introducing a new term ("clock type") that basically means the same concept as another term ("clock ID") just causes confusions.

kernel/src/device/evdev/file.rs
```diff
@@ -0,0 +1,270 @@
...
+        })
+    }
+}
+
+impl InodeIo for EvdevFile {
+    fn read_at(
+        &self,
+        _offset: usize,
+        writer: &mut VmWriter,
+        _status_flags: StatusFlags,
```
tatetian: I don't know if any userspace program would call `fcntl()` to set the `O_NONBLOCK` flag on an opened evdev file. But respecting the status flags here is cheap and thus seem to be a good idea.  The `InodeHandle` type allows any file opened at a FS path to [set its status flags](https://github.com/asterinas/asterinas/blob/2034055f90607f67f04de044ee88dd52a488d612/kernel/src/fs/inode_handle/mod.rs#L158C18-L158C34), including the bit indicating whether the file I/O on the file should be blocking or nonblocking, and these status flags are [passed to](https://github.com/asterinas/asterinas/blob/2034055f90607f67f04de044ee88dd52a488d612/kernel/src/fs/inode_handle/mod.rs#L91) the read/write operations of the underlying file. If the evdev file does its part correctly, then it will also support non-blocking I/O. The following example shows how tty devices respect this flag:

https://github.com/asterinas/asterinas/blob/2034055f90607f67f04de044ee88dd52a488d612/kernel/src/device/tty/mod.rs#L263-L268

The evdev file can do it in a similar way.

tatetian (commented):
Thanks for addressing my preview comments! I found a few more minor issues (see below). I expect this to be the last round of revision.

Cautreoxit:
Thanks for your comments and patience! I have addressed these issues.

PR #2408, opened 2025-09-05
Support sealing memfd files

vvvvsv:
This PR is required for https://github.com/asterinas/asterinas/issues/2214.

This PR introduces the mechanism for sealing memfd files. See the *File Sealing* sections in [fcntl](https://man7.org/linux/man-pages/man2/fcntl.2.html) and [memfd_create](https://man7.org/linux/man-pages/man2/memfd_create.2.html).

## Discussions about `F_SEAL_WRITE`

According to [fcntl](https://man7.org/linux/man-pages/man2/fcntl.2.html),
> **F_SEAL_WRITE**
> ...if this seal is set.  Furthermore, trying to create new shared, writable memory-mappings via [mmap(2)](https://man7.org/linux/man-pages/man2/mmap.2.html) will also fail with EPERM.
> Using the F_ADD_SEALS operation to set the F_SEAL_WRITE seal fails with EBUSY if any writable, shared mapping exists.  Such mappings must be unmapped before you can add this seal...

Here, *"shared, writable memory-mappings"* refer to shared VMAs with the `VM_MAYWRITE` flag set. Since we currently do not support this flag, all shared mappings may be upgraded to `PROT_WRITE` via `mprotect`, and are therefore *"shared, writable memory-mappings"*. For more details about `VM_MAYWRITE`, please see https://github.com/asterinas/asterinas/issues/2407.

To implement `F_SEAL_WRITE`, we need to track the number of shared, writable memory-mappings for each file. `WritableMappingStatus` in the first commit introduces this functionality, following [the approach](https://github.com/torvalds/linux/blob/d69eb204c255c35abd9e8cb621484e8074c75eaa/include/linux/fs.h#L593-L627) used in Linux. Counting this number is not only useful for memfd files. in Linux, this counter is also used by [many other features](https://elixir.bootlin.com/linux/v6.14.7/A/ident/mapping_writably_mapped), mainly about I/O operations.

PR #2555, opened 2025-11-05
Ensure that every `FileLike` is associated with a `dyn Inode`

vvvvsv:
This PR is required for #2214.

As discussed in https://github.com/asterinas/asterinas/pull/2521#issuecomment-3458168421 and https://github.com/asterinas/asterinas/pull/2521#issuecomment-3460571218, this PR introduces `PseudoInode`, and ensures that each pseudo `FileLike` has an inode.

In Linux behavior, both ends of a pipe point to the same inode, each socket has its own inode, and most members of anon_inodefs (such as epollfile, eventfile, timerfdfile, signalfile, pidfd, inotifyfd, etc.) [share the same inode](https://elixir.bootlin.com/linux/v6.16.5/source/fs/anon_inodes.c#L208-L238). These inodes are not readable or writable, but their metadata is accessible (e.g., they can be `chmod`ed, `chown`ed, `fstat`ed, etc.). This PR adds `PseudoInode` with only metadata and filesystem info to match Linux behavior.

```
PipeReaderFile ┐
               ├── PseudoInode
PipeWriterFile ┘

Socket ── PseudoInode

EpollFile ──┐
EventFile ──┤
TimerfdFile ┼─ PseudoInode
SignalFile ─┤
Pidfd ──────┤
Inotifyfd ──┘

MemfdFile ── MemfdInode

InodeHandle ── dyn Inode
```

This PR uses `ReadLinkResult` and `PathOrInode` to support the `FsResolver` to lookup the inodes of pseudo files. Then, supports `chmod`, `chown`, `fstat` for pseudo files, and supports opening memfd files from `/proc/[pid]/fd/[fd]`. After that, removes redundant inode-specific methods (such as `set_mode`) from `FileLike`.

```rust
// The result of `inode.read_link()`
enum ReadLinkResult {
    Symbolic(String),
    Path(Path), // NOTE: For `InodeHandle` in `/proc/[pid]/fd/[fd]`, we should use this variant.
    // FIXME: This variant exists because not all `Arc<dyn FileLike>`s are associated with paths.
    // We should introduce pipefs, socketfs, etc., and then remove this variant.
    Inode(Arc<dyn Inode>),
}
impl FsResolver {
    fn lookup_inode(&self) -> Result<PathOrInode>;
}

impl PathOrInode {
    fn inode(&self) -> &Arc<dyn Inode>;
    fn open(&self) -> Arc<dyn FileLike>;
}
```

PR #2521, opened 2025-10-23
Support executing memfd files and then `open("/proc/self/exe")`

vvvvsv:
This PR is required for https://github.com/asterinas/asterinas/issues/2214.

Briefly, the goal of this PR is to ensure the following code runs successfully on Asterinas:

```c
    memfd = memfd_create("name", MFD_CLOEXEC);
    // Write an ELF file to memfd
    execveat(memfd, "", argv, envp, AT_EMPTY_PATH);

    // After `execveat`, in the new ELF.
    fd = open("/proc/self/exe");
```

# Discussion

In the following text, "pseudo files" refer to files that do not have a concrete file path, such as anon pipes, sockets, eventfds, etc. Thus, the goal of this PR is to `execveat(pseudo file)` and `open(pseudo file)`. To improve Linux compatibility and reduce future refactoring, let's first take a look at how Linux's VFS handles pseudo files.

## Does Linux support opening pseudo files with the syscall `open`?

In Linux, after generating pseudo files and obtaining their `fd`s, is it possible to use `open("/proc/self/fd/<fd>")` on them? 

**Only pipes can be opened this way.** Sockets will encounter `ENXIO`, and members of the `anoninodefs` (such as eventfd, epollfd, etc.) will encounter `EACCES`. Note that in Linux, memfd is not a pseudo inode. When Linux performs `memfd_create`, it first creates a RAM inode in tmpfs, then immediately [unlinks](https://elixir.bootlin.com/linux/v6.16.5/source/mm/shmem.c#L5850) it, and finally returns only the file descriptor.

## Linux VFS structures of pseudo files

image

In Linux, each `open`ed file handle corresponds to a `struct file*`. The `struct file` contains a `struct path*` (which may be a pseudo path). The `struct path` contains a `struct dentry*`, and the `struct dentry` contains a `struct inode*`. This part is similar to `InodeHandle` in Asterinas. 

The `struct file` also has a field `struct file_operations*`, which includes a set of function pointers such as read, write, seek, and open. Abstractly, this is analogous to the `trait FileLike` in Asterinas. What seems rather unusual is that **for pseudo files that can be re`open`ed (i.e., pipes), their `struct inode` contains a field `struct file_operations*`**, which points to the `file_operations` of the corresponding `struct file`.

Additionally, `struct file` has a field `void *private_data`, which points to different structs depending on the type of pseudo file.

[...]

PR #2563, opened 2025-11-07
Refine some access mode checks

lrh2000:
This PR simplifies some code added in https://github.com/asterinas/asterinas/pull/2408. For instance, `writable_mapping_status` does not need to be an `Option` in `Vmo`; `MemfdFile` can directly hold an `Arc<MemfdInode>` (instead of an `Arc<dyn Inode>`). Sorry for not catching these issues during the review process.

More importantly, this PR adds missing access mode checks for `MemfdFile` and `InodeHandle`.
 - The access mode check for `MemfdFile` is completely missing, which will be problematic once we reopen it (`memfd_create` always generates a FD with `O_RDWR`, so it may not be a problem now).
 - For `InodeHandle`, I found that we can map `O_PATH` to `Rights::empty()` to resolve this TODO.

https://github.com/asterinas/asterinas/blob/cdeb9ec5fd68603e48ec543c4f8ea88cfc97b2f4/kernel/src/fs/inode_handle/dyn_cap.rs#L26-L27

### DIFF
--- a/kernel/src/fs/inode_handle/dyn_cap.rs
+++ b/kernel/src/fs/inode_handle/dyn_cap.rs
@@ -40,13 +40,11 @@ impl InodeHandle {
 
     pub fn new_unchecked_access(
         path: Path,
-        mut access_mode: AccessMode,
+        access_mode: AccessMode,
         status_flags: StatusFlags,
     ) -> Result<Self> {
         let inode = path.inode();
         let (file_io, rights) = if status_flags.contains(StatusFlags::O_PATH) {
-            // We follow Linux to report O_RDONLY later (e.g., in `/proc/[pid]/fdinfo/[n]`).
-            access_mode = AccessMode::O_RDONLY;
             (None, Rights::empty())
         } else if inode.type_() == InodeType::Dir && access_mode.is_writable() {
             return_errno_with_message!(Errno::EISDIR, "a directory cannot be opened writable");
@@ -60,7 +58,6 @@ impl InodeHandle {
             path,
             file_io,
             offset: Mutex::new(0),
-            access_mode,
             status_flags: AtomicU32::new(status_flags.bits()),
         };
         Ok(Self(inner, rights))
@@ -133,7 +130,6 @@ impl Pollable for InodeHandle {
 #[inherit_methods(from = "self.0")]
 impl FileLike for InodeHandle {
     fn status_flags(&self) -> StatusFlags;
-    fn access_mode(&self) -> AccessMode;
     fn metadata(&self) -> Metadata;
     fn mode(&self) -> Result<InodeMode>;
     fn set_mode(&self, mode: InodeMode) -> Result<()>;
@@ -197,6 +193,10 @@ impl FileLike for InodeHandle {
     fn set_status_flags(&self, new_status_flags: StatusFlags) -> Result<()> {
         self.0.set_status_flags(new_status_flags);
         Ok(())
+    }
+
+    fn access_mode(&self) -> AccessMode {
+        self.1.into()
     }
 
     fn seek(&self, seek_from: SeekFrom) -> Result<usize> {
--- a/kernel/src/fs/inode_handle/mod.rs
+++ b/kernel/src/fs/inode_handle/mod.rs
@@ -15,9 +15,9 @@ use crate::{
         file_handle::Mappable,
         path::Path,
         utils::{
-            AccessMode, DirentVisitor, FallocMode, FileRange, FlockItem, FlockList, Inode,
-            InodeMode, InodeType, IoctlCmd, Metadata, RangeLockItem, RangeLockList, RangeLockType,
-            SeekFrom, StatusFlags, OFFSET_MAX,
+            DirentVisitor, FallocMode, FileRange, FlockItem, FlockList, Inode, InodeMode,
+            InodeType, IoctlCmd, Metadata, RangeLockItem, RangeLockList, RangeLockType, SeekFrom,
+            StatusFlags, OFFSET_MAX,
         },
     },
     prelude::*,
@@ -29,12 +29,11 @@ use crate::{
 
 struct HandleInner {
     path: Path,
-    /// `file_io` is Similar to `file_private` field in `file` structure in linux. If
-    /// `file_io` is Some, typical file operations including `read`, `write`, `poll`,
-    /// `ioctl` will be provided by `file_io`, instead of `path`.
+    /// `file_io` is similar to the `file_private` field in Linux's `file` structure. If `file_io`
+    /// is `Some(_)`, typical file operations including `read`, `write`, `poll`, and `ioctl` will
+    /// be provided by `file_io`, instead of `path`.
     file_io: Option<Arc<dyn FileIo>>,
     offset: Mutex<usize>,
-    access_mode: AccessMode,
     status_flags: AtomicU32,
 }
 
@@ -118,10 +117,6 @@ impl HandleInner {
 
     pub(self) fn resize(&self, new_size: usize) -> Result<()> {
         do_resize_util(self.path.inode().as_ref(), self.status_flags(), new_size)
-    }
-
-    pub(self) fn access_mode(&self) -> AccessMode {
-        self.access_mode
     }
 
     pub(self) fn status_flags(&self) -> StatusFlags {
@@ -278,7 +273,6 @@ impl Debug for HandleInner {
         f.debug_struct("HandleInner")
             .field("path", &self.path)
             .field("offset", &self.offset())
-            .field("access_mode", &self.access_mode())
             .field("status_flags", &self.status_flags())
             .finish_non_exhaustive()
     }
--- a/kernel/src/fs/utils/access_mode.rs
+++ b/kernel/src/fs/utils/access_mode.rs
@@ -38,6 +38,22 @@ impl AccessMode {
     }
 }
 
+impl From<Rights> for AccessMode {
+    fn from(rights: Rights) -> Self {
+        match (
+            rights.contains(Rights::READ),
+            rights.contains(Rights::WRITE),
+        ) {
+            (true, true) => AccessMode::O_RDWR,
+            (true, false) => AccessMode::O_RDONLY,
+            (false, true) => AccessMode::O_WRONLY,
+            // The file is opened with `O_PATH`. We follow Linux to report `O_RDONLY` here (e.g.,
+            // in `/proc/[pid]/fdinfo/[n]`).
+            (false, false) => AccessMode::O_RDONLY,
+        }
+    }
+}
+
 impl From<AccessMode> for Rights {
     fn from(access_mode: AccessMode) -> Rights {
         match access_mode {


kernel/src/fs/ramfs/memfd.rs
```diff
@@ -210,7 +206,7 @@ impl Inode for MemfdInode {
 }
 
 pub struct MemfdFile {
-    memfd_inode: Arc<dyn Inode>,
+    memfd_inode: Arc<MemfdInode>,
```
vvvvsv: As discussed in https://github.com/asterinas/asterinas/pull/2521#issuecomment-3458168421, we may need to add a method `inode` in `FileLike`:

```rust
trait FileLike {
    fn inode(&self) -> &Arc<dyn Inode>;
}
```

Therefore, this `memfd_inode` field in `MemfdFile` needs to be an `Arc<dyn Inode>`, I think.

If we still want this field to be an `Arc<MemfdInode>`, we can modify this method's signature to:

```rust
trait FileLike {
    fn inode(&self) -> Arc<dyn Inode>;
}
```

and then

```rust
impl FileLike for MemfdFile {
    fn inode(&self) -> Arc<dyn Inode> {
        self.memfd_inode.clone()
    }
}
```

but this will introduce some redundant ref count cloning.

kernel/src/vm/vmo/mod.rs
```diff
@@ -74,8 +74,7 @@ pub struct Vmo {
     /// `pages` will be the latest size.
     size: AtomicUsize,
     /// The status of writable mappings of the VMO.
-    /// This field is `None` if the VMO is not backed by a pager.
-    writable_mapping_status: Option<WritableMappingStatus>,
+    writable_mapping_status: WritableMappingStatus,
```
vvvvsv, line 81: > writable_mapping_status does not need to be an Option in Vmo

This was discussed in https://github.com/asterinas/asterinas/pull/2408#issuecomment-3269836352.

To summarize, a `Vmo` can serve as a pagecache or a backend of shared anonymous mappings, and the latter case never needs to count writable mappings. In the former case, besides the page cache of memfd, we currently don’t count writable mappings, but we will need to so in the future.

The advantage of using `Option<WritableMappingStatus>` instead of `WritableMappingStatus` is that if a careless user tries to access the `writable_mapping_status().as_ref().unwrap()` of a shared anonymous mapping backend, the `unwrap` will panic.

kernel/src/fs/ramfs/memfd.rs
lrh2000: >     fn inode(&self) -> &Arc<dyn Inode>;

Sorry, I totally forgot about this, I thought all the existing `do_xxx_util` functions could easily be fixed.

`do_xxx_util` should be fixed anyway as they shouldn't take an `&Arc<dyn Inode>` for no reason, it's a reference of another reference.

Well. So I will consider to keep the `Arc<dyn Inode>`. It's not ideal, but it's acceptable if there's a good reason for it.

kernel/src/vm/vmo/mod.rs
lrh2000: > This was discussed in https://github.com/asterinas/asterinas/pull/2408#issuecomment-3269836352.

At that time, I believe we had an `Arc<WritableMappingStatus>`. If so, then we can avoid an `Arc` if it's a `None`. But now, we don't have an `Arc` anyway.

I found it odd to have an `Option` here. Exactly as you said, I agree that the writable mapping status should be a property of the VMO. Then, *this field should only be set to `Some()` if and only if the VMO is a memfd's VMO*. However, setting this field to `Some()` for all VMOs with a page cache backend does not make sense.

In other words, the following two conditions should be equivalent. Otherwise, I think we should just remove this `Option` because it will cause confusion:
 - A VMO has a `WritableMappingStatus`.
 - The writable VM mapping of that VMO needs to be tracked.

In https://github.com/asterinas/asterinas/pull/2408#discussion_r2480011862, I suggest to make an VMO to have the `WritableMappingStatus` if and only if the writable mappings need to be tracked, but I don't think we really do so now. So I'll prefer to remove this `Option`.

> The advantage of using `Option<WritableMappingStatus>` instead of `WritableMappingStatus` is that if a careless user tries to access the `writable_mapping_status().as_ref().unwrap()` of a shared anonymous mapping backend, the `unwrap` will panic.

This check is incomplete because it cannot prevent non-memfd VMOs with page cache backend to be tracked, which should not happen, either.

> In the former case, besides the page cache of memfd, we currently don’t count writable mappings, but we will need to so in the future.

We can discuss this in the future. But I don't think it's necessary to consider this now. Also note that Linux is retiring the related mechanism, e.g., https://lwn.net/Articles/866493/.

kernel/src/fs/ramfs/memfd.rs
vvvvsv: > `do_xxx_util` should be fixed anyway as they shouldn't take an `&Arc<dyn Inode>` for no reason, it's a reference of another reference.

Agree!

kernel/src/vm/vmo/mod.rs
vvvvsv: Thanks for your explanation! Since we already have a field `is_writable_tracked` in `MappedVmo`, this change sounds reasonable to me.

> Also note that Linux is retiring the related mechanism

The related mechanism is also used in archs like ARM. This was discussed in https://github.com/asterinas/asterinas/pull/2408#issuecomment-3269173064:

>   ```c
>   if (mapping_writably_mapped(mapping)) {
>       flush_dcache_folio(folio);
>   }
>   ```
>     
>   The function `flush_dcache_folio` flushes the data cache and ensures that any changes in the CPU’s data cache are written back to main memory. On x86, `flush_dcache_folio` is a no-op. However, on architectures that do not guarantee cache and memory coherence (like ARM), `flush_dcache_folio` needs to be performed.

I think this usage can not be retired in the future.

Reference: https://elixir.bootlin.com/linux/v6.14.7/A/ident/mapping_writably_mapped

kernel/src/fs/file_handle.rs
```diff
@@ -48,15 +48,15 @@ pub trait FileLike: Pollable + Send + Sync + Any {
     }
 
     fn ioctl(&self, cmd: IoctlCmd, arg: usize) -> Result<i32> {
-        return_errno_with_message!(Errno::EINVAL, "ioctl is not supported");
+        return_errno_with_message!(Errno::ENOTTY, "ioctl is not supported");
```
tatetian, line 54: Taken at the surface value, this default error code looks quite surprising. But from the man page, this seems to be a more sensible choice. Could you add a comment that quotes the following description from the man page?

```
       ENOTTY The specified operation does not apply to the kind of
              object that the file descriptor fd references.
```

kernel/src/fs/file_handle.rs
```diff
@@ -48,15 +48,15 @@ pub trait FileLike: Pollable + Send + Sync + Any {
...
     }
 
     /// Obtains the mappable object to map this file into the user address space.
     ///
     /// If this file has a corresponding mappable object of [`Mappable`],
     /// then it can be either an inode or an MMIO region.
     fn mappable(&self) -> Result<Mappable> {
-        return_errno_with_message!(Errno::EINVAL, "the file is not mappable");
+        return_errno_with_message!(Errno::ENODEV, "the file is not mappable");
```
tatetian, line 65: Same request as the comment for `ioctl`.

kernel/src/fs/ramfs/memfd.rs
lrh2000: This change is reverted. Thanks for catching it!

kernel/src/vm/vmo/mod.rs
lrh2000: Thanks for your additional information.

I ended up with this. I think it is (slightly) more accurate:
```patch
diff --git a/kernel/src/vm/vmo/mod.rs b/kernel/src/vm/vmo/mod.rs
index 4d3a5f15f..50bb6d414 100644
--- a/kernel/src/vm/vmo/mod.rs
+++ b/kernel/src/vm/vmo/mod.rs
@@ -74,6 +74,10 @@ pub struct Vmo {
     /// `pages` will be the latest size.
     size: AtomicUsize,
     /// The status of writable mappings of the VMO.
+    //
+    // TODO: This field is used only by VMOs belonging to memfd (i.e., `MemfdInode`). But VMOs do
+    // not have the knowledge to determine if they belong to memfd. We may want to enhance
+    // `VmoOptions` to make VMOs aware of whether its writable mappings should be tracked.
     writable_mapping_status: WritableMappingStatus,
 }
 
@@ -453,6 +457,8 @@ impl Vmo {
 
     /// Returns the status of writable mappings of the VMO.
     pub fn writable_mapping_status(&self) -> &WritableMappingStatus {
+        // Only writable file-backed mappings may need to be tracked.
+        debug_assert!(self.pager.is_some());
         &self.writable_mapping_status
     }
 }
 
```

kernel/src/fs/file_handle.rs
lrh2000: > But from the man page, this seems to be a more sensible choice.

In fact, I'm changing the code to make my regression tests pass (the tests were included in the same commit).
So regardless of what the man pages say, we'll have to fix the code because the man pages can sometimes be outdated and inaccurate. 

> Could you add a comment that quotes the following description from the man page?

Of course. I've added the man pages as a reference.

lrh2000:
As (privately) discussed with @vvvvsv,
 - This PR needs to address https://github.com/asterinas/asterinas/pull/2555#discussion_r2509396676
 - I'll revert the memfd-related access mode checks so that #2555 can address it without causing unnecessary conflicts.

lrh2000:
In the same private discussion, @vvvvsv points out that our file lock implementation (`flock()` and `fcntl(F_SETLK)`) does not properly check the access mode, particularly for the `O_PATH` case. This issue should have been resolved in the last revision.

As an additional note, dropping the memfd changes related to access mode is perfectly fine because without the ability to reopen a memfd (https://github.com/asterinas/asterinas/pull/2555 will add that capability), all memfds created by `memfd_create()` have the `O_RDWR` access mode. Therefore, theoretically, there is nothing to check.

Read [the complete diff of the last force-push here](https://github.com/asterinas/asterinas/compare/ace814eebf8f527ac35b49b1dc2e8f2c0e739f7d..86286d87d186cdbfc5bf0005418e3d839e3b4234).

---

One more thing: the interaction between `O_PATH` and other flags (at least `O_CREAT` and `O_TRUNC` ) is also incorrect. Hopefully, the newly added commit will resolve this issue. See the commit ["Fix how O_PATH interacts with other flags"](https://github.com/asterinas/asterinas/pull/2563/commits/55dd0078cbd5b6daf8cee17e6221da7f93c90605).

vvvvsv:
> As an additional note, dropping the memfd changes related to access mode is perfectly fine

Thanks! That would be added in #2555. 

By the way, could you please check your WeChat messages?

kernel/src/fs/inode_handle/dyn_cap.rs
```diff
@@ -9,26 +9,32 @@ use crate::{fs::file_handle::Mappable, prelude::*, process::signal::Pollable};
...
+            // We follow Linux to report O_RDONLY later (e.g., in `/proc/[pid]/fdinfo/[n]`).
+            access_mode = AccessMode::O_RDONLY;
+            (None, Rights::empty())
+        } else if inode.type_() == InodeType::Dir && access_mode.is_writable() {
+            return_errno_with_message!(Errno::EISDIR, "a directory cannot be opened writable");
         } else {
-            inode.open(access_mode, status_flags).transpose()?
+            let file_io = inode.open(access_mode, status_flags).transpose()?;
+            let rights = Rights::from(access_mode);
+            (file_io, rights)
```
StevenJiang1110, line 54: This PR encodes the access mode in the `rights` field, this makes great sense to me. There are two minor issues:

1. Can we remove the `access_mode` field from `InodeHandle_`? It looks like maintain both `access_mode` and `rights` is redudant?
2. The changes only affect the dynamic-capability version of InodeHandle. In principle the static-capability implementation should be updated too, but the `static_cap` module appears to be unused since it was introduced and currently only contains dead code. My suggestion is to remove the `static_cap` module in this PR.


In the long term, however, using the static capability mechanism for InodeHandle might be the better design: permission bits are immutable after a file is opened, which is exactly what static capabilities are intended to model. That said, switching to static capabilities raises unresolved questions (for example, we currently downcast `Arc<dyn FileLike>` to `Arc<InodeHandle>`, and supporting an equivalent downcast for the static-capability variant may not be straightforward). So I propose leaving that as future work.

kernel/src/fs/inode_handle/mod.rs
```diff
@@ -114,7 +114,7 @@ impl InodeHandle_ {
     }
 
     pub fn seek(&self, pos: SeekFrom) -> Result<usize> {
-        do_seek_util(self.path.inode(), &self.offset, pos)
```
StevenJiang1110, line 117: All methods on `InodeHandle_` should not have the `pub` visibility. Could you also cleanup the `pub` keyword in this PR?

kernel/src/fs/inode_handle/dyn_cap.rs
vvvvsv: > In the long term, however, using the static capability mechanism for InodeHandle might be the better design: permission bits are immutable after a file is opened, which is exactly what static capabilities are intended to model.

The meaning of "static" here refers to permissions that can be checked at compile time, rather than that the permission bits are immutable. I think having a field `rights: Rights` without interior mutability in `InodeHandle` is already sufficient to express that the permission bits are immutable.


> Can we remove the `access_mode` field from `InodeHandle_`?
> My suggestion is to remove the `static_cap` module in this PR.

I vote on this. Please take a look at the third paragraph of https://github.com/asterinas/asterinas/pull/2555#issuecomment-3516482891. Removing `InodeHandle<TRights>` also brings some other benefits.

StevenJiang1110: > The meaning of "static" here refers to permissions that can be checked at compile time.

Exactly. 

> I think having a field rights: Rights without interior mutability in InodeHandle is already sufficient to express that the permission bits are immutable.

Static capabilities are not primarily intended to represent immutability. However, when a property is immutable it can be expressed naturally with a static capability. By definition, capabilities are monotonic and cannot be expanded, so immutability is a sufficient, though not necessary, condition for representing a property as a capability.

The main advantage, as you noted, is moving access mode checks to compile time: access mode can be validated without runtime overhead. This is the motivation behind introducing the static-capability machinery. If we can use a static capability in this case, we should prefer it.

However, there are unresolved issues if we switch to static capabilities now — for example, the downcast problem and how to convert an `access_mode` into a static capability during open. If we can design clean APIs to address those problems, we should migrate to static capabilities. For the moment, though, I suggest leaving this as future work.

> Please take a look at the third paragraph of https://github.com/asterinas/asterinas/pull/2555#issuecomment-3516482891

I'm glad to see this direction — I think it’s the right one. Then It’s up to you and @lrh2000 to decide who will take on the implementation.

lrh2000: > 1. Can we remove the `access_mode` field from `InodeHandle_`? It looks like maintain both `access_mode` and `rights` is redudant?
> 2. The changes only affect the dynamic-capability version of InodeHandle. In principle the static-capability implementation should be updated too, but the `static_cap` module appears to be unused since it was introduced and currently only contains dead code. My suggestion is to remove the `static_cap` module in this PR.

I wanted to do this as well. But I was afraid that it will cause some merge conflicts. But now I have a try.

I tried to avoid moving type definitions, so `InodeHandle` still locates at `inode_handle/dyn_cap.rs` and `HandleInner` (previously `InodeHandle_`) locates at `inode_handle/mod.rs`. It is a bit odd, but I want to stop here because fixing the files that define these types may cause a lot of conflicts to other PRs.

tatetian: This arrangement is ok. I am counting on you to refactor the code organization of `inode_handle/` in the near future.

PR #2216, opened 2025-06-26
Add support for user-space framebuffer(`/dev/fb0`)

Tangzh33:
This PR introduces user-space framebuffer access through `/dev/fb0`, allowing applications to interact directly with the display framebuffer. With this implementation, key capabilities are enabled for user apps:

- Basic framebuffer operations (read/write pixels)
- Screen resolution and color depth information
- Memory-mapped access for high-performance graphics
- Simple display control and manipulation

Note: For simplicity, we currently only support the GOP (Graphics Output Protocol) Framebuffer. In other words, the GOP Framebuffer is initialized by the EFI, and the device capabilities are limited. Currently, it doesn't support some advanced `ioctls`, such as FBIO_BLANK. The implementation focuses on core functionality first, making it suitable for basic graphics applications. Future enhancements can add more advanced features as needed.

Lastly, this PR is built upon PR #2099 and provides vital support for Issue #2008. Here are some materials for quick reference: 

1. https://wiki.osdev.org/GOP
2. https://docs.kernel.org/fb/efifb.html
3. https://github.com/torvalds/linux/blob/master/drivers/video/fbdev/efifb.c

PR #2550, opened 2025-11-03
Fix the behavior about closing pty

StevenJiang1110:
This PR fixes the behavior when closing a PTY master or slave, which is required by #2214.

When reading from a PTY master whose corresponding slave has been closed, the read should immediately return EIO if no data remains. In the current implementation the read can block instead. Generally, when one end of a PTY is closed the other end receives POLLHUP event and subsequent read/write calls return EIO. The semantics for master and slave differ in several respects; the included regression test should test those differences. 

A closed slave can be reopened by opening `/dev/pts/[num]`, which restores normal master–slave operation. The master, however, cannot be reopened after it is closed because there is no way to locate it via the filesystem.

When a PTY master is open, attempting to unlink the corresponding device node (`/dev/pts/[index]`) returns EPERM. After the master is closed, the slave node will be removed automatically, trying to unlink that node returns ENOENT on Linux. Our current implementation about devpts has a small problem about unlink, which is fixed in the fourth commit.

The third commit fixes another issue: when running Podman, PtyMaster's Drop implementation could panic because devpts had already been dropped. I couldn't determine the root cause, so as a workaround the code now skips the unwrapping step when devpts has been dropped.

PR #1284, opened 2024-09-03
Make special devices go to the `FileIo` fast path

lucassong-mh:

PR #2564, opened 2025-11-07
Add `InodeIo` to simplify `FileIo` and `Inode`

lrh2000:
**This PR is based on https://github.com/asterinas/asterinas/pull/2563 to avoid merge conflicts. Please consider reviewing it first.**

https://github.com/asterinas/asterinas/pull/2216 needs to support reading from and writing to a `FileIo` (i.e., the framebuffer) with an offset. However, we cannot duplicate the offset in both the `InodeHandle` and the `FileIo` because doing so would introduce race conditions. Meanwhile, I found that the following methods in `FileIo` and `Inode` are duplicates.
https://github.com/asterinas/asterinas/blob/cda06613f027d25d47a7670b7c8aa7d4cb118ae9/kernel/src/fs/inode_handle/mod.rs#L357-L361
https://github.com/asterinas/asterinas/blob/cda06613f027d25d47a7670b7c8aa7d4cb118ae9/kernel/src/fs/utils/inode.rs#L302-L316

This PR proposes adding a new trait, `InodeIo`, which provides offset-based `read`/`write` methods with the status flag.
```rust
pub trait InodeIo {
    /// Reads data from the file into the given `VmWriter`.
    fn read_at(
        &self,
        offset: usize,
        writer: &mut VmWriter,
        status_flags: StatusFlags,
    ) -> Result<usize>;

    /// Writes data from the given `VmReader` into the file.
    fn write_at(
        &self,
        offset: usize,
        reader: &mut VmReader,
        status_flags: StatusFlags,
    ) -> Result<usize>;
}
```

Then, `FileIo` and `Inode` will have `InodeIo` as a super-trait.
```rust
pub trait Inode: Any + InodeIo + Send + Sync {
    // ...
}

pub trait FileIo: Pollable + InodeIo + Send + Sync + 'static {
    // ...
}
```

This will enable us to naturally support `read_at`/`write_at` with `FileIo`.
```rust
impl InodeHandle_ {
    pub fn read(&self, writer: &mut VmWriter) -> Result<usize> {
        let (inode_io, is_seekable) = self.inode_io_and_is_seekable();
        let status_flags = self.status_flags();

        if !is_seekable {
            return inode_io.read_at(0, writer, status_flags);
        }

        let mut offset = self.offset.lock();

        let len = inode_io.read_at(*offset, writer, status_flags)?;
        *offset += len;

        Ok(len)
    }
```

Additionally, I noticed that it is unnecessary to have `FileIo` living in an `Arc`. I replaced `Arc<dyn FileIo>` with `Box<dyn FileIo>`, and I believe the code is improved as a result.

### DIFF
--- a/kernel/src/fs/inode_handle/mod.rs
+++ b/kernel/src/fs/inode_handle/mod.rs
@@ -123,7 +123,22 @@ impl HandleInner {
     }
 
     pub(self) fn seek(&self, pos: SeekFrom) -> Result<usize> {
-        do_seek_util(self.path.inode().as_ref(), &self.offset, pos)
+        if let Some(ref file_io) = self.file_io {
+            file_io.check_seekable()?;
+            if file_io.is_offset_aware() {
+                // TODO: Figure out whether we need to add support for seeking from the end of
+                // special files.
+                return do_seek_util(&self.offset, pos, None);
+            } else {
+                return Ok(0);
+            }
+        }
+
+        let inode = self.path.inode();
+        if !inode.type_().is_seekable() {
+            return_errno_with_message!(Errno::ESPIPE, "seek is not supported");
+        }
+        do_seek_util(&self.offset, pos, inode.seek_end())
     }
 
     pub(self) fn offset(&self) -> usize {
@@ -292,6 +307,10 @@ pub trait FileIo: Pollable + InodeIo + Send + Sync + 'static {
     fn check_seekable(&self) -> Result<()>;
 
     /// Returns whether the `read()`/`write()` operation should use and advance the offset.
+    ///
+    /// If [`is_seekable`] succeeds but this method returns `false`, the offset in the `seek()`
+    /// operation will be ignored. In that case, the `seek()` operation will do nothing but
+    /// succeed.
     fn is_offset_aware(&self) -> bool;
 
     // See `FileLike::mappable`.
@@ -305,34 +324,33 @@ pub trait FileIo: Pollable + InodeIo + Send + Sync + 'static {
 }
 
 pub(super) fn do_seek_util(
-    inode: &dyn Inode,
     offset: &Mutex<usize>,
     pos: SeekFrom,
+    end: Option<usize>,
 ) -> Result<usize> {
     let mut offset = offset.lock();
-    let new_offset: isize = match pos {
-        SeekFrom::Start(off /* as usize */) => {
-            if off > isize::MAX as usize {
-                return_errno_with_message!(Errno::EINVAL, "file offset is too large");
+
+    let new_offset = match pos {
+        SeekFrom::Start(off) => off,
+        SeekFrom::End(diff) => {
+            if let Some(end) = end {
+                end.wrapping_add_signed(diff)
+            } else {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "seeking the file from the end is not supported"
+                );
             }
-            off as isize
         }
-        SeekFrom::End(off /* as isize */) => {
-            let file_size = inode.size() as isize;
-            assert!(file_size >= 0);
-            file_size
-                .checked_add(off)
-                .ok_or_else(|| Error::with_message(Errno::EOVERFLOW, "file offset overflow"))?
-        }
-        SeekFrom::Current(off /* as isize */) => (*offset as isize)
-            .checked_add(off)
-            .ok_or_else(|| Error::with_message(Errno::EOVERFLOW, "file offset overflow"))?,
+        SeekFrom::Current(diff) => offset.wrapping_add_signed(diff),
     };
-    if new_offset < 0 {
-        return_errno_with_message!(Errno::EINVAL, "file offset must not be negative");
+
+    // Invariant: `*offset <= isize::MAX as usize`.
+    // TODO: Investigate whether `read`/`write` can break this invariant.
+    if new_offset.cast_signed() < 0 {
+        return_errno_with_message!(Errno::EINVAL, "the file offset cannot be negative");
     }
-    // Invariant: 0 <= new_offset <= isize::MAX
-    let new_offset = new_offset as usize;
+
     *offset = new_offset;
     Ok(new_offset)
 }
--- a/kernel/src/fs/utils/inode.rs
+++ b/kernel/src/fs/utils/inode.rs
@@ -389,7 +389,23 @@ pub trait Inode: Any + InodeIo + Send + Sync {
         true
     }
 
-    /// Get the extension of this inode
+    /// Returns the end position for [`SeekFrom::End`].
+    ///
+    /// [`SeekFrom::End`]: super::SeekFrom::End
+    fn seek_end(&self) -> Option<usize> {
+        if self.type_() == InodeType::File {
+            Some(self.size())
+        } else {
+            // This depends on the file system. For example, seeking directories from the end
+            // succeeds under procfs and btrfs but fails under tmpfs. Here, we just choose a
+            // safe default to reject it.
+            // TODO: Carefully check the Linux behavior of each file system and adjust ours
+            // accordingly.
+            None
+        }
+    }
+
+    /// Gets the extension of this inode.
     fn extension(&self) -> Option<&Extension> {
         None
     }
--- a/kernel/src/fs/procfs/template/dir.rs
+++ b/kernel/src/fs/procfs/template/dir.rs
@@ -204,6 +204,11 @@ impl<D: DirOps + 'static> Inode for ProcDir<D> {
     fn is_dentry_cacheable(&self) -> bool {
         !self.common.is_volatile()
     }
+
+    fn seek_end(&self) -> Option<usize> {
+        // Seeking directories under `/proc` with `SEEK_END` will start from zero.
+        Some(0)
+    }
 }
 
 pub trait DirOps: Sync + Send + Sized {
--- a/kernel/src/fs/procfs/template/file.rs
+++ b/kernel/src/fs/procfs/template/file.rs
@@ -78,6 +78,7 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
     fn fs(&self) -> Arc<dyn FileSystem>;
 
     fn resize(&self, _new_size: usize) -> Result<()> {
+        // Resizing files under `/proc` will succeed, but will do nothing.
         Ok(())
     }
 
@@ -95,6 +96,11 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
 
     fn is_dentry_cacheable(&self) -> bool {
         !self.common.is_volatile()
+    }
+
+    fn seek_end(&self) -> Option<usize> {
+        // Seeking regular files under `/proc` with `SEEK_END` will fail.
+        None
     }
 }
 
--- a/kernel/src/fs/ramfs/memfd.rs
+++ b/kernel/src/fs/ramfs/memfd.rs
@@ -433,7 +433,7 @@ impl FileLike for MemfdFile {
             return_errno_with_message!(Errno::EBADF, "the file is opened as a path");
         }
 
-        do_seek_util(self.memfd_inode.as_ref(), &self.offset, pos)
+        do_seek_util(&self.offset, pos, Some(self.memfd_inode.size()))
     }
 
     fn fallocate(&self, mode: FallocMode, offset: usize, len: usize) -> Result<()> {
--- a/kernel/src/syscall/lseek.rs
+++ b/kernel/src/syscall/lseek.rs
@@ -12,20 +12,25 @@ use crate::{
 pub fn sys_lseek(fd: FileDesc, offset: isize, whence: u32, ctx: &Context) -> Result<SyscallReturn> {
     debug!("fd = {}, offset = {}, whence = {}", fd, offset, whence);
 
-    let seek_from = match whence {
-        0 => {
-            if offset < 0 {
-                return_errno!(Errno::EINVAL);
-            }
-            SeekFrom::Start(offset as usize)
-        }
-        1 => SeekFrom::Current(offset),
-        2 => SeekFrom::End(offset),
-        _ => return_errno!(Errno::EINVAL),
+    let seek_from = match SeekType::try_from(whence)? {
+        SeekType::SEEK_SET => SeekFrom::Start(offset.cast_unsigned()),
+        SeekType::SEEK_CUR => SeekFrom::Current(offset),
+        SeekType::SEEK_END => SeekFrom::End(offset),
     };
+
     let mut file_table = ctx.thread_local.borrow_file_table_mut();
     let file = get_file_fast!(&mut file_table, fd);
 
     let offset = file.seek(seek_from)?;
     Ok(SyscallReturn::Return(offset as _))
+}
+
+// Reference: <https://elixir.bootlin.com/linux/v6.17.7/source/include/uapi/linux/fs.h#L52>
+#[derive(Clone, Copy, Debug, TryFromInt)]
+#[repr(u32)]
+#[expect(non_camel_case_types)]
+enum SeekType {
+    SEEK_SET = 0,
+    SEEK_CUR = 1,
+    SEEK_END = 2,
 }
--- a/test/src/apps/file_io/access_err.c
+++ b/test/src/apps/file_io/access_err.c
@@ -52,6 +52,7 @@ FN_TEST(readable)
 	TEST_SUCC(read(fd, buf, sizeof(buf)));
 	TEST_ERRNO(write(fd, buf, sizeof(buf)), EBADF);
 	TEST_SUCC(lseek(fd, 0, SEEK_SET));
+	TEST_SUCC(lseek(fd, 0, SEEK_END));
 	TEST_ERRNO(ioctl(fd, TCGETS), ENOTTY);
 	TEST_ERRNO(ftruncate(fd, 1), EINVAL);
 	TEST_ERRNO(fallocate(fd, FALLOC_FL_KEEP_SIZE, 0, 1), EBADF);
@@ -86,6 +87,7 @@ FN_TEST(readable)
 	TEST_ERRNO(read(fd, buf, sizeof(buf)), EISDIR);
 	TEST_ERRNO(write(fd, buf, sizeof(buf)), EBADF);
 	TEST_SUCC(lseek(fd, 0, SEEK_SET));
+	TEST_ERRNO(lseek(fd, 0, SEEK_END), EINVAL);
 	TEST_ERRNO(ioctl(fd, TCGETS), ENOTTY);
 	TEST_ERRNO(ftruncate(fd, 1), EINVAL);
 	TEST_ERRNO(fallocate(fd, FALLOC_FL_KEEP_SIZE, 0, 1), EBADF);
@@ -127,6 +129,7 @@ FN_TEST(writeable)
 	TEST_ERRNO(read(fd, buf, sizeof(buf)), EBADF);
 	TEST_SUCC(write(fd, buf, sizeof(buf)));
 	TEST_SUCC(lseek(fd, 0, SEEK_SET));
+	TEST_SUCC(lseek(fd, 0, SEEK_END));
 	TEST_ERRNO(ioctl(fd, TCGETS), ENOTTY);
 	TEST_SUCC(ftruncate(fd, 1));
 	TEST_SUCC(fallocate(fd, FALLOC_FL_KEEP_SIZE, 0, 1));
@@ -173,6 +176,7 @@ FN_TEST(path)
 	TEST_ERRNO(read(fd, buf, sizeof(buf)), EBADF);
 	TEST_ERRNO(write(fd, buf, sizeof(buf)), EBADF);
 	TEST_ERRNO(lseek(fd, 0, SEEK_SET), EBADF);
+	TEST_ERRNO(lseek(fd, 0, SEEK_END), EBADF);
 	TEST_ERRNO(ioctl(fd, TCGETS), EBADF);
 	TEST_ERRNO(ftruncate(fd, 1), EBADF);
 	TEST_ERRNO(fallocate(fd, FALLOC_FL_KEEP_SIZE, 0, 1), EBADF);
@@ -205,6 +209,7 @@ FN_TEST(path)
 	TEST_ERRNO(read(fd, buf, sizeof(buf)), EBADF);
 	TEST_ERRNO(write(fd, buf, sizeof(buf)), EBADF);
 	TEST_ERRNO(lseek(fd, 0, SEEK_SET), EBADF);
+	TEST_ERRNO(lseek(fd, 0, SEEK_END), EBADF);
 	TEST_ERRNO(ioctl(fd, TCGETS), EBADF);
 	TEST_ERRNO(ftruncate(fd, 1), EBADF);
 	TEST_ERRNO(fallocate(fd, FALLOC_FL_KEEP_SIZE, 0, 1), EBADF);
--- a/test/src/syscall/gvisor/Makefile
+++ b/test/src/syscall/gvisor/Makefile
@@ -9,6 +9,7 @@ TESTS ?= \
 	access_test \
 	chown_test \
 	creat_test \
+	dev_test \
 	dup_test \
 	epoll_test \
 	eventfd_test \
--- a/test/src/syscall/gvisor/blocklists/dev_test
+++ b/test/src/syscall/gvisor/blocklists/dev_test
@@ -1,0 +1,2 @@
+# TODO: Support `/dev/fuse`
+DevTest.ReadDevFuseWithoutMount
--- a/test/src/syscall/gvisor/blocklists/lseek_test
+++ b/test/src/syscall/gvisor/blocklists/lseek_test
@@ -1,6 +1,2 @@
-LseekTest.Overflow
-LseekTest.ProcFile
+# TODO: Add `/sys/devices` and support `lseek(SEEK_END)`.
 LseekTest.SysDir
-LseekTest.SeekCurrentDir
-LseekTest.ProcStatTwice
-LseekTest.EtcPasswdDup


kernel/src/fs/inode_handle/mod.rs
```diff
@@ -38,7 +38,7 @@ struct InodeHandle_ {
     /// `file_io` is Similar to `file_private` field in `file` structure in linux. If
     /// `file_io` is Some, typical file operations including `read`, `write`, `poll`,
     /// `ioctl` will be provided by `file_io`, instead of `path`.
-    file_io: Option<Arc<dyn FileIo>>,
+    file_io: Option<Box<dyn FileIo>>,
```
StevenJiang1110, line 31: I agree the `Arc<dyn FileIo>` is problematic.

While working on #2550 I ran into the same issue. It’s awkward that `PtySlaveFile` already contains an `Arc<PtySlave>`, yet we have to return an `Arc<PtySlaveFile>` because open must return `Arc<dyn FileIo>`. We need to perform cleanup when an `InodeHandle` is dropped, but the `Arc<dyn FileIo>` (i.e. `Arc<PtySlave>`) is also referenced by `PtyMaster` and `PtySlaveInode`, so we can’t rely on `PtySlave`'s `Drop` for that cleanup. That’s why we introduced `PtySlaveFile` (and similarly `NamedPipeHandle`) to handle this.

I see two possible approaches:
1. Return `Box<dyn FileIo>` instead of `Arc<dyn FileIo>` just like this PR suggests. That guarantees the `Box<dyn FileIo>` is dropped when the `InodeHandle` is dropped. The downside is an extra indirection: for example, a `Box` would still contain an inner `Arc`.  The `Box<PtySlaveFile>` is actually `Box<Arc<PtySlave>>`
2. Keep `Arc<dyn FileIo>` but add a `close_file` method to the `FileIo` trait and invoke it from `InodeHandle::drop`. This lets us explicitly run the `close_file` method in `InodeHandle::Drop`. The downside is that `Arc<dyn FileIo>` can be unnecessarily heavy for trivial devices (e.g. /dev/null), where a Box would be simpler and more efficient.

For me, both should be accepable but is not very satisfactory.  I’d like to hear your thoughts about this.

kernel/src/device/full.rs
```diff
@@ -23,8 +23,8 @@ impl Device for Full {
         DeviceId::new(1, 7)
     }
 
-    fn open(&self) -> Option<Result<Arc<dyn FileIo>>> {
-        Some(Ok(Arc::new(Full)))
+    fn open(&self) -> Result<Box<dyn FileIo>> {
+        Ok(Box::new(Self))
```
StevenJiang1110, line 29: I recently noticed that `Device::open` now returns `Some` for all devices (e.g. `/dev/full`, `/dev/null`). When the `file_io` field in `InodeHandle` was first introduced, `Device::open` returned Some only for tty devices (i.e. `/dev/tty`); for other devices it returned `None`. I check the commit history, the change in behavior was introduced by #1284 to speed up I/O on special devices.

As the comment on the `file_io` field in InodeHandle indicates, that field corresponds to the `private_data` field in Linux’s `struct file`. In [older Linux versions (e.g. 6.0)](https://elixir.bootlin.com/linux/v6.0.9/source/include/linux/fs.h#L968) that field is documented as being used for tty drivers and possibly some other drivers (the specific tty comment has been removed in [newer kernels](https://elixir.bootlin.com/linux/v6.17/source/include/linux/fs.h#L1104)). Since Asterinas doesn’t have those "other drivers", we didn't expect private_data used by code at that time. That explains why our `FileIo` trait is very limited (no seek, pread/pwrite, etc.).

To support the usage pattern introduced by #1284, the current `FileIo` trait is insufficient. This PR appears to introduce `InodeIo` to cover pread/pwrite cases. In Linux, opening a device provides a struct file_operations ([the /dev/zero case](https://elixir.bootlin.com/linux/v6.17/source/drivers/char/mem.c#L650)), which is conceptually closer to our `FileLike` trait. So perhaps Device::open should return `Box<dyn FileLike>` instead of `Box<dyn FileIo>`. What do you think?

I also agree that the direction in this PR to split file operations into separate trait groups (e.g. FileIo, Seekable, Pollable, InodeIo) to enable reuse is a good approach.

kernel/src/fs/inode_handle/mod.rs
lrh2000: Thanks for your review!

> While working on https://github.com/asterinas/asterinas/pull/2550

Sorry, I didn't rebase this PR on top of #2550 yet. Consequently, the diff is slightly outdated. Hopefully, this won't affect much.

> That guarantees the `Box<dyn FileIo>` is dropped when the `InodeHandle` is dropped.

Yes, this is one benefit. But it's not the only benefit.

Reference-counting here is not really necessary for some other cases, e.g., `/dev/null`:

```rust
pub struct Null;

impl Device for Null {
    fn open(&self) -> Result<Box<dyn FileIo>> {
        Ok(Box::new(Self))
    }
}
```
Note that `Null` is a unit type. So creating a `Box<Null>` does not need any heap allocation. Then this code should be as efficient as returning a `None`, but with simpler APIs (i.e., no need to return an `Option`, which makes the semantic more complex). `Arc<Null>` will need heap allocation for the reference counter and will also be slower because of reference counting.

Conceptually, a `FileIo` instance is always associated with an `InodeHandle`, so from the perspective of `InodeHandle`, I'd prefer to remove this `Arc`.

> The `Box<PtySlaveFile>` is actually `Box<Arc<PtySlave>>`

We should be able to avoid this in the future by playing more tricks with the pointer metadata. It should be possible to introduce a new type in OSTD, `Lane<dyn FileIo>`, that can represent either a `Box<X>` or an `Arc<Y>` and implement `FileIo` on them (`impl FileIo for Box<X>`/`impl FileIo for Arc<Y>`). It will not introduce any additional cost (especially no redundant vtable dispatch) but can avoid the unnecessary indirection.

> a `close_file` method to the `FileIo` trait

Then I guess the first approach is clearly better?

kernel/src/device/full.rs
lrh2000: > I check the commit history, the change in behavior was introduced by #1284 to speed up I/O on special devices.

Thanks for the background!

As I've explained in the comment above, using `Box` on a unit type does not introduce any cost. Then I believe there are no performance issues.

> So perhaps Device::open should return `Box<dyn FileLike>` instead of `Box<dyn FileIo>`.

This will make other things more complicated, such as reopening the file via `/proc/[pid]/fd/[n]`. I don't think we can do that easily now. Being a `FileLike` also means that it needs to manage the status flag, the offset, and perhaps other things. I'm not sure if this is the right approach.

> That explains why our `FileIo` trait is very limited

It seems that in Linux, a socket is a `struct inode` (instead of a `struct file`). Our design isn't really aligned with it well. But I am also not sure whether fixing it will provide any real benefit.

kernel/src/fs/inode_handle/mod.rs
StevenJiang1110: Thanks for your detailed reply! 

> Note that Null is a unit type. So creating a `Box<Null>` does not need any heap allocation.

I haven't noticed this benefit. This really makes sense!

> We should be able to avoid this in the future by playing more tricks with the pointer metadata. It should be possible to introduce a new type in OSTD, `Lane<dyn FileIo>`

This seems to an awesome pattern if we could have it!

> Then I guess the first approach is clearly better?

Yes, now I agree the first approach is absolutely better.

kernel/src/device/full.rs
StevenJiang1110: Thanks for your reply!

> Being a FileLike also means that it needs to manage the status flag, the offset, and perhaps other things. I'm not sure if this is the right approach.

Oh yes — I realized our `FileLike` doesn't really match Linux's `struct file_operations`, because it still carries state/data. So returning a `FileLike` would also cause problems.

> It seems that in Linux, a socket is a struct inode (instead of a struct file). Our design isn't really aligned with it well. But I am also not sure whether fixing it will provide any real benefit.

Makes sense. We can fix such problems until we encounter the real case.

kernel/src/device/pty/master.rs
```diff
@@ -99,7 +104,12 @@ impl FileIo for PtyMaster {
         Ok(read_len)
     }
 
-    fn write(&self, reader: &mut VmReader, status_flags: StatusFlags) -> Result<usize> {
+    fn write_at(
+        &self,
+        _offset: usize,
+        reader: &mut VmReader,
+        status_flags: StatusFlags,
+    ) -> Result<usize> {
```
StevenJiang1110, line 120: Using pread and pwrite for I/O triggers calls to these `read_at` and `write_at` methods. However, when applied to pty master and slaves, both pread and pwrite consistently fail with ESPIPE. I expect pipes and sockets would behave similarly. My mental model suggests that unseekable files cannot support pread/pwrite. Unseekable means this file cannot support read/write at random offset. Although I'm unsure if this applies to all files, it seems advisable to prioritize checking for seekability before calling read/write in read_at/write_at.

Another option is to check seekable in InodeHandle or HandleInner before calling InodeIo::read_at/write_at. However, that would make the InodeIo implementations for types like PtyMaster and PtySlave effectively dead code. We should make our intent explicit—for example, returning ESPIPE with an explanatory error message to indicate that those code paths are not meant to be reached.

Below is a test case for pty, it can be run by putting it in the last of `test/src/apps/pty/close_pty.c`.

```C
FN_TEST(pread)
{
	open_and_set_slave_raw_mode();

	char buf[3] = { 'a', 'b', 'c' };
	TEST_SUCC(write(master, buf, sizeof(buf)));
	TEST_ERRNO(pwrite64(master, buf, sizeof(buf), 0), ESPIPE);

	TEST_ERRNO(pread64(slave, buf, sizeof(buf), 0), ESPIPE);
	TEST_RES(read(slave, buf, sizeof(buf)),
		 _ret == 3 && buf[0] == 'a');

	TEST_ERRNO(pwrite64(slave, buf, sizeof(buf), 0), ESPIPE);
	TEST_SUCC(write(slave, buf, sizeof(buf)));
}
END_TEST()
```

kernel/src/fs/inode_handle/mod.rs
```diff
@@ -286,6 +300,15 @@ impl Debug for HandleInner {
...
+    /// Returns whether the file-like object is seekable.
+    ///
+    /// The concrete meanings of the return values are as follows:
+    ///  - `Ok(true)`: The `read()`/`write()` operation will use and advance the offset. The
+    ///    `seek()` operation is supported.
+    ///  - `Ok(false)`: The `read()`/`write()` operation will use a zero as the offset. The
+    ///    `seek()` operation is not supported but it will always succeed with zero.
+    ///  - `Err(_)`: The `read()`/`write()` operation will use a zero as the offset. The `seek()`
+    ///    operation is not supported and it will fail with the returned error.
     fn is_seekable(&self) -> Result<bool>;
```
StevenJiang1110: Should this method belong to `InodeIo`? Seekability seems like an Inode property.

We already have an `is_seekable` method on `Inode` that returns a bool; its meaning is clear — it simply indicates whether the file can be seeked. The newly added method encodes additional, more complex semantics, that is to say, whether we should update the offset. Since those are orthogonal concerns, could we split it into two methods: one for the pure seekable check, and a separate one (both on `InodeIo`) that decides whether the offset should be changed?

```rust
pub trait InodeIo {
    // Checks whether the Inode supports seek operation.
    fn check_seekable(&self) -> Result<()>;
    // Returns whether the `read()`/`write()` operation will use and advance the offset.
    fn is_offset_aware(&self) -> bool;
}
```

Further, I'm not sure whether `mappable` and `ioctl` methods should be moved from `FileIo` to `InodeIo`?

kernel/src/fs/inode_handle/mod.rs
```diff
@@ -35,70 +36,81 @@ struct HandleInner {
...
-        if self.status_flags().contains(StatusFlags::O_DIRECT) {
-            self.path.inode().read_direct_at(offset, writer)
+    fn inode_io_and_is_seekable(&self) -> (&dyn InodeIo, bool) {
+        if let Some(ref file_io) = self.file_io {
+            let is_seekable = file_io.is_seekable().is_ok_and(|is_seekable| is_seekable);
+            (file_io.as_ref(), is_seekable)
         } else {
-            self.path.inode().read_at(offset, writer)
+            let inode = self.path.inode();
+            let is_seekable = inode.type_().is_seekable();
```
StevenJiang1110: We can use `inode.is_seekable` here if is_seekable is moved from FileIo to InodeIo? I suppose this would simplify the code.

kernel/src/device/tdxguest/mod.rs
```diff
@@ -103,14 +103,30 @@ impl Pollable for TdxGuest {
...
+        _status_flags: StatusFlags,
+    ) -> Result<usize> {
         return_errno_with_message!(Errno::EPERM, "Write operation not supported")
     }
+}
+
+impl FileIo for TdxGuest {
+    fn is_seekable(&self) -> Result<bool> {
+        Ok(false)
+    }
```
StevenJiang1110, line 129: Perhaps we should conservatively return EPERM or ESPIPE here. The TdxGuest cannot be read from or written to at all. I don't know the exact errno used on Linux, though.

kernel/src/fs/inode_handle/mod.rs
```diff
@@ -35,70 +36,81 @@ struct HandleInner {
...
+    pub(self) fn write_at(&self, mut offset: usize, reader: &mut VmReader) -> Result<usize> {
+        let inode_io = self.inode_io();
         let status_flags = self.status_flags();
+
         if status_flags.contains(StatusFlags::O_APPEND) {
-            // If the file has the O_APPEND flag, the offset is ignored
+            // If the file has the `O_APPEND` flag, the offset is ignored.
+            // FIXME: `O_APPEND` should ensure that new content is appended even if another process
+            // is writing to the file concurrently.
             offset = self.path.size();
```
StevenJiang1110, line 103: If `inode_io` is `self.file_io`, the code will end up using the size of self.path. Because file_io doesn't always correspond to the same inode as `self.path`, we should obtain the size from `inode_io` instead. 

Adding a temporary FIXME here is acceptable, since `InodeIo` does not currently provide a `size()` method.

Or you can add another path to directly call FileIo::read for this special case. But I'm not sure whether it deserves.
```rust
if let Some(ref file_io) = self.file_io && status_flags.contains(StatusFlags::O_APPEND) {
    // Should we check seekable here?
    return file_io.read(reader, status_flags)?;
}
```

kernel/src/device/pty/master.rs
lrh2000: > Another option is to check seekable in InodeHandle or HandleInner before calling InodeIo::read_at/write_at.

I think we can do this.

> However, that would make the InodeIo implementations for types like PtyMaster and PtySlave effectively dead code.

It will not be dead code. `read`/`write` will call `InodeIo::read_at(0, ...)`/`InodeIo::write_at(0, ...)`. So this is not a concern?

kernel/src/fs/inode_handle/mod.rs
lrh2000: > We already have an `is_seekable` method on `Inode` that returns a bool

I want to remove `is_seekable` from `Inode` because `is_seekable` is a property of an `InodeType`, instead of a property of an `Inode`. But it's a property of `FileIo`, so it only belongs to `FileIo` instead of `InodeIo`.

If we have it in `InodeIo`, then we'll have to repeat its implementation for all inode implementations, which is not very ideal.

> could we split it into two methods: one for the pure seekable check, and a separate one (both on `InodeIo`) that decides whether the offset should be changed?

I'll give this a try, but I still think these methods should be in the `FileIo` trait. I don't see how moving them to `InodeIo` can help the code here:
```rust
    pub(self) fn seek(&self, pos: SeekFrom) -> Result<usize> {
        if let Some(ref file_io) = self.file_io {
            file_io.is_seekable();
            if file_io.is_offset_aware() {
                // TODO: Figure out whether we need to add support for seeking from the end of
                // special files.
                return do_seek_util(&self.offset, pos, None);
            } else {
                return Ok(0);
            }
        }

        let inode = self.path.inode();
        if !inode.type_().is_seekable() {
            return_errno_with_message!(Errno::ESPIPE, "seek is not supported");
        }
        do_seek_util(&self.offset, pos, inode.seek_end())
    }
```

kernel/src/fs/inode_handle/mod.rs
lrh2000: But the logic here is not inode-specific. Methods on `InodeIo` cannot have a default implementation because they cannot access `Inode::type_`. Therefore, the same implementation must be repeated for all inode implementations.

Besides, I don't see how removing `type_()` simplifies the code?

kernel/src/device/tdxguest/mod.rs
lrh2000: It will fail with `ESPIPE` in Linux:
https://github.com/torvalds/linux/blob/868941b14441282ba08761b770fc6cad69d5bdb7/fs/read_write.c#L287-L288

I guess the `read`/`write`'s error code is also wrong, it may be `EINVAL`:
https://github.com/torvalds/linux/blob/868941b14441282ba08761b770fc6cad69d5bdb7/fs/read_write.c#L457-L458

kernel/src/device/pty/master.rs
StevenJiang1110: > So this is not a concern?

Makes sense. Thanks for your explanation!

kernel/src/fs/inode_handle/mod.rs
StevenJiang1110: > If we have it in InodeIo, then we'll have to repeat its implementation for all inode implementations, which is not very ideal.

Thanks for the explanation — I can see the problem now. Since `InodeIo` is a supertrait of `Inode`, it can't access Inode-specific properties.

That means `is_seekable` can only reasonably be a method on the `Inode` trait (or on `FileIo`). The Inode trait already exposes a `type_()` method, so `is_seekable` could be provided as a default method on `Inode`. However, that alone doesn't eliminate the two code paths we currently have for `FileIo` and `Path`.

> I don't see how moving them to InodeIo can help the code here:

My original hope was to handle `FileIo` and `Path` with the same code, e.g.:

```rust
let inode_io = self.inode_io();
inde_io.check_seekable()?
if inode_io.is_offset_aware() {
    do_seek_util(&self.offset, pos, None)?
} else {
    Ok(0)
}
```

But if `InodeIo` can't provide seek-related methods, we can't write code like that.

In summary, it looks like we still need two branches for `FileIo` and `Path`. Given that, adding `is_seekable` to FileIo seems acceptable.

kernel/src/fs/inode_handle/mod.rs
StevenJiang1110: Same as https://github.com/asterinas/asterinas/pull/2564#discussion_r2532502415

kernel/src/device/tdxguest/mod.rs
StevenJiang1110: Thanks for your investigation! Then we'd better fix all the error numbers here.

kernel/src/fs/inode_handle/mod.rs
```diff
@@ -272,16 +299,17 @@ impl Debug for HandleInner {
...
-// TODO: The `status_flags` parameter in `read` and `write` may need to be stored directly
-// in the `FileIo`. We need further refactoring to find an appropriate way to enable `FileIo`
-// to utilize the information in the `HandleInner`.
-pub trait FileIo: Pollable + Send + Sync + 'static {
-    /// Reads data from the file into the given `VmWriter`.
-    fn read(&self, writer: &mut VmWriter, status_flags: StatusFlags) -> Result<usize>;
-
-    /// Writes data from the given `VmReader` into the file.
-    fn write(&self, reader: &mut VmReader, status_flags: StatusFlags) -> Result<usize>;
+pub trait FileIo: Pollable + InodeIo + Send + Sync + 'static {
```
StevenJiang1110, line 305: Another point for discussion is whether we should remove the read/write methods from FileIo. It appears FileIo::read/write consistently delegate to `InodeIo::read_at(0, ...) / write_at(0, ...)` respectively. Are there any exceptions to this behavior, and what potential issues could arise if we make this change?

lrh2000:
This is a rebase. See the next force-push for new changes.

kernel/src/fs/inode_handle/mod.rs
lrh2000: Now it's:
```c
        // FIXME: How can we deal with the `O_APPEND` flag if `file_io` is set?
        if status_flags.contains(StatusFlags::O_APPEND) && self.file_io.is_none() {
```

kernel/src/fs/inode_handle/mod.rs
lrh2000:
```rust
        do_seek_util(&self.offset, pos, inode.seek_end())
```

There is also a `Inode::seek_end`. Actually I admit it's a bit messy but I cannot figure out a better design.

kernel/src/fs/inode_handle/mod.rs
lrh2000: > what potential issues could arise if we make this change?

For example, if a `FileIo`'s `read`/`write` behavior is completely different from its `read_at`/`write_at` behavior? I'm not sure if there is such a case.

kernel/src/fs/inode_handle/mod.rs
StevenJiang1110: Makes sense to me!

kernel/src/fs/inode_handle/mod.rs
StevenJiang1110: I'm not sure about that either. So, we might not need to address this problem in this PR. Let's investigate this question later.

kernel/src/fs/inode_handle/mod.rs
lrh2000: > Further, I'm not sure whether `mappable` and `ioctl` methods should be moved from `FileIo` to `InodeIo`?

Sorry, I forgot to respond to this. It's because all the inodes currently have the same implementation (`mappable` always returns `None` and `ioctl` always fails with `ENOTTY`). So, moving it to `InodeIo` does not seem to offer any benefits.

StevenJiang1110: > Actually I admit it's a bit messy but I cannot figure out a better design.

Yes, I have the same feeling that the code could be refactored more elegantly, but I haven't come up with a better solution yet. For now, the current implementation is acceptable since it resolves many existing issues.

> It's because all the inodes currently have the same implementation (mappable always returns None and ioctl always fails with ENOTTY).

This makes sense.

kernel/src/fs/utils/inode.rs
```diff
@@ -242,7 +243,25 @@ impl From<Arc<dyn Device>> for MknodType {
     }
 }
 
-pub trait Inode: Any + Sync + Send {
+pub trait InodeIo {
```
cchanging, line 250: This refactoring looks good to me! However, like @StevenJiang1110 , I also feel that the seekable-related implementation is a bit unusual and somewhat logically redundant, but after reading your discussion, I understand the reasons for doing it this way.

I'm wondering if the following implementation approach would work? I haven't actually tried it, but if it could work, it might simplify the seekable-related logic somewhat. If not, I think the current implementation is also OK.

In addition, I feel that some documentation could be added to InodeIo to clarify its semantics.

```rust
pub trait Inode {
    /// Reads data from the file into the given `VmWriter`.
    fn read_at(
        &self,
        offset: usize,
        writer: &mut VmWriter,
        status_flags: StatusFlags,
    ) -> Result<usize>;

    /// Writes data from the given `VmReader` into the file.
    fn write_at(
        &self,
        offset: usize,
        reader: &mut VmReader,
        status_flags: StatusFlags,
    ) -> Result<usize>;
    ...
}

pub trait InodeIo {
    /// Reads data from the file into the given `VmWriter`.
    fn read_at(
        &self,
        offset: usize,
        writer: &mut VmWriter,
        status_flags: StatusFlags,
    ) -> Result<usize>;

    /// Writes data from the given `VmReader` into the file.
    fn write_at(
        &self,
        offset: usize,
        reader: &mut VmReader,
        status_flags: StatusFlags,
    ) -> Result<usize>;

    // Checks whether the Inode supports seek operation.
    fn check_seekable(&self) -> Result<()>;

    // Returns whether the `read()`/`write()` operation will use and advance the offset.
    fn is_offset_aware(&self) -> bool;
}

impl InodeIo for dyn Inode {
    fn read_at(
        &self,
        offset: usize,
        writer: &mut VmWriter,
        status_flags: StatusFlags,
    ) -> Result<usize> {
        self.read_at(...)
    }

    fn write_at(
        &self,
        offset: usize,
        reader: &mut VmReader,
        status_flags: StatusFlags,
    ) -> Result<usize> {
        self.write_at(...)
    }

    fn check_seekable(&self) -> Result<()> {
        Ok(())
    }

    fn is_offset_aware(&self) -> bool {
        self.type_().is_seekable()
    }
}

pub trait FileIo: InodeIo {
    ...
}
```

lrh2000: > ```rust
> impl InodeIo for dyn Inode {
> ```

Do you want to `impl InodeIo for dyn Inode` so that you can get a `dyn InodeIo` from a `dyn Inode`? It won't work ([Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9555afe736ced21195d5e811d512d82e)):
```rust
trait X {}

trait Y {}

impl Y for dyn X {}

fn cast(x: &dyn X) -> &dyn Y {
    x as _ // <- error: non-primitive cast: `&dyn X` as `&dyn Y`
}
```

There is no such feature in the Rust language. It cannot even be trivially implemented: `&dyn X` and `&dyn Y` are two fat pointers. How can you generate a `&dyn Y` from a `&dyn X`?

> In addition, I feel that some documentation could be added to InodeIo to clarify its semantics.

Yes. We should perhaps explain that it's the common part of both the `Inode` type and the `FileIo` type.

PR #2566, opened 2025-11-07
Make the semantics of tty-related devices correct

cchanging:
The current semantics of `/dev/ttyx` in Asterinas differ from those in Linux. This PR introduces `VtManager` to manage virtual terminals (tty1, tty2...), while ensuring that `/dev/tty0` carries more accurate semantics. 

### Some Semantics of TTY in Linux

| Device File     | Description                                                                                                                              |
|-----------------|------------------------------------------------------------------------------------------------------------------------------------------|
| `/dev/tty`      | The controlling terminal of the current process (e.g., a program running in tty3 will have `/dev/tty` pointing to `/dev/tty3`).           |
| `/dev/tty0`     | The globally current active virtual console (e.g., if you press Ctrl+Alt+F3, then `/dev/tty0` is equivalent to `/dev/tty3`).              |
| `/dev/tty1–63`  | Fixed-numbered virtual console devices (they exist even when inactive).                                                                 In modern systems, once activated, they default to using the same physical framebuffer (fb) device.                                      |
| `/dev/console`  | The target console for kernel message output (usually a soft link to `tty0`, but can be configured to a serial port like `ttyS0`).        |

### Current Implementation in Asterinas  
The virtual console device correspond to `tty0`, and if a framebuffer device exists, it corresponds to `tty1`.

### Is this necessary?

In addition, after enabling the framebuffer device, `VtManager`'s backend device becomes the framebuffer device, leaving the original virtual console device without a corresponding node. Therefore, this PR declares the `ttyS0` node to correspond to the virtual console device, ensuring that the virtual console has a corresponding node in the kernel even after the framebuffer device is enabled. 

Otherwise, if we don't preserve the virtual console device, where will the kernel logs output after enabling the desktop system?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,13 @@ FEATURES ?=
 NO_DEFAULT_FEATURES ?= 0
 COVERAGE ?= 0
 ENABLE_BASIC_TEST ?= false
+# Specify the primary system console (supported: hvc0, tty0).
+# - hvc0: The virtio-console terminal.
+# - tty0: The active virtual terminal (VT).
+# Asterinas will automatically fall back to tty0 if hvc0 is not available.
+# Note that currently the virtual terminal (tty0) can only work with
+# linux-efi-handover64 and linux-efi-pe64 boot protocol.
+CONSOLE ?= hvc0
 # End of global build options.
 
 # GDB debugging and profiling options.
@@ -57,6 +64,7 @@ CARGO_OSDK := ~/.cargo/bin/cargo-osdk
 CARGO_OSDK_COMMON_ARGS := --target-arch=$(OSDK_TARGET_ARCH)
 # The build arguments also apply to the `cargo osdk run` command.
 CARGO_OSDK_BUILD_ARGS := --kcmd-args="ostd.log_level=$(LOG_LEVEL)"
+CARGO_OSDK_BUILD_ARGS += --kcmd-args="console=$(CONSOLE)"
 CARGO_OSDK_TEST_ARGS :=
 
 ifeq ($(AUTO_TEST), syscall)
--- a/kernel/comps/framebuffer/src/dummy_console.rs
+++ b/kernel/comps/framebuffer/src/dummy_console.rs
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use aster_console::{
+    font::BitmapFont,
+    mode::{ConsoleMode, KeyboardMode},
+    AnyConsoleDevice, ConsoleCallback, ConsoleSetFontError,
+};
+
+/// A dummy console device.
+///
+/// This is used when no framebuffer is available. All operations are no-ops.
+#[derive(Debug)]
+pub struct DummyFramebufferConsole;
+
+impl AnyConsoleDevice for DummyFramebufferConsole {
+    fn send(&self, _buf: &[u8]) {}
+
+    fn register_callback(&self, _callback: &'static ConsoleCallback) {}
+
+    fn set_font(&self, _font: BitmapFont) -> Result<(), ConsoleSetFontError> {
+        Err(ConsoleSetFontError::InappropriateDevice)
+    }
+
+    fn set_mode(&self, _mode: ConsoleMode) -> bool {
+        false
+    }
+
+    fn mode(&self) -> Option<ConsoleMode> {
+        None
+    }
+
+    fn set_keyboard_mode(&self, _mode: KeyboardMode) -> bool {
+        false
+    }
+
+    fn keyboard_mode(&self) -> Option<KeyboardMode> {
+        None
+    }
+}
--- a/kernel/src/device/mod.rs
+++ b/kernel/src/device/mod.rs
@@ -10,8 +10,6 @@ mod pty;
 mod shm;
 pub mod tty;
 
-use alloc::format;
-
 use device_id::DeviceId;
 pub use mem::{getrandom, geturandom};
 pub use pty::{new_pty_pair, PtyMaster, PtySlave};
@@ -43,16 +41,22 @@ pub fn init_in_first_process(ctx: &Context) -> Result<()> {
     let dev_path = fs_resolver.lookup(&FsPath::try_from("/dev")?)?;
     dev_path.mount(RamFs::new(), PerMountFlags::default(), ctx)?;
 
-    tty::init();
+    tty::init_in_first_process();
+
+    let tty0 = Arc::new(tty::Tty0Device);
+    add_node(tty0, "tty0", &fs_resolver)?;
+
+    let tty1 = tty::tty1_device().clone();
+    add_node(tty1, "tty1", &fs_resolver)?;
 
     let tty = Arc::new(tty::TtyDevice);
     add_node(tty, "tty", &fs_resolver)?;
 
-    let console = tty::system_console().clone();
+    let console = tty::SystemConsole::singleton().clone();
     add_node(console, "console", &fs_resolver)?;
 
-    for (index, tty) in tty::iter_n_tty().enumerate() {
-        add_node(tty.clone(), &format!("tty{}", index), &fs_resolver)?;
+    if let Some(hvc0) = tty::hvc0_device() {
+        add_node(hvc0.clone(), "hvc0", &fs_resolver)?;
     }
 
     pty::init_in_first_process(&fs_resolver, ctx)?;
@@ -75,18 +79,18 @@ pub fn get_device(devid: DeviceId) -> Result<Arc<dyn Device>> {
     let minor = devid.minor().get();
 
     match (major, minor) {
-        (4, minor) => {
-            let Some(tty) = tty::iter_n_tty().nth(minor as usize) else {
-                return_errno_with_message!(Errno::EINVAL, "the TTY minor ID is invalid");
-            };
-            Ok(tty.clone())
-        }
+        (4, 0) => Ok(Arc::new(tty::Tty0Device)),
+        (4, 1) => Ok(tty::tty1_device().clone()),
         (5, 0) => Ok(Arc::new(tty::TtyDevice)),
+        (5, 1) => Ok(tty::SystemConsole::singleton().clone()),
+        (229, 0) => tty::hvc0_device()
+            .cloned()
+            .map(|device| device as _)
+            .ok_or_else(|| Error::with_message(Errno::ENODEV, "the hvc0 device is not available")),
         _ => char::lookup(devid)
             .map(|device| Arc::new(char::CharFile::new(device)) as Arc<dyn Device>)
-            .ok_or(Error::with_message(
-                Errno::EINVAL,
-                "the device ID is invalid or unsupported",
-            )),
+            .ok_or_else(|| {
+                Error::with_message(Errno::EINVAL, "the device ID is invalid or unsupported")
+            }),
     }
 }
--- a/kernel/src/device/tty/device.rs
+++ b/kernel/src/device/tty/device.rs
@@ -1,20 +1,69 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! TTY devices.
+//!
+//! This module implements TTY devices such as `/dev/tty0`, `/dev/tty`, and `/dev/console`.
+//!
+//! Reference: <https://www.kernel.org/doc/html/latest/admin-guide/devices.html>
+
 use device_id::{DeviceId, MajorId, MinorId};
+use spin::Once;
 
 use crate::{
+    device::tty::{hvc0_device, n_tty::VtDriver, tty1_device, Tty},
     fs::{
         device::{Device, DeviceType},
         inode_handle::FileIo,
     },
+    kcmdline::KCmdlineArg,
     prelude::*,
+    process::{JobControl, Terminal},
 };
 
+/// Corresponds to `/dev/tty0` in the file system. This device represents the active virtual
+/// terminal.
+pub struct Tty0Device;
+
+impl Tty0Device {
+    fn active_vt(&self) -> &Arc<Tty<VtDriver>> {
+        // Currently there is only one virtual terminal `tty1`.
+        tty1_device()
+    }
+}
+
+impl Device for Tty0Device {
+    fn type_(&self) -> DeviceType {
+        DeviceType::Char
+    }
+
+    fn id(&self) -> DeviceId {
+        DeviceId::new(MajorId::new(4), MinorId::new(0))
+    }
+
+    fn open(&self) -> Result<Box<dyn FileIo>> {
+        self.active_vt().open()
+    }
+}
+
+impl Terminal for Tty0Device {
+    fn job_control(&self) -> &JobControl {
+        self.active_vt().job_control()
+    }
+}
+
 /// Corresponds to `/dev/tty` in the file system. This device represents the controlling terminal
-/// of the session of current process.
+/// of the session of the current process.
 pub struct TtyDevice;
 
 impl Device for TtyDevice {
+    fn type_(&self) -> DeviceType {
+        DeviceType::Char
+    }
+
+    fn id(&self) -> DeviceId {
+        DeviceId::new(MajorId::new(5), MinorId::new(0))
+    }
+
     fn open(&self) -> Result<Box<dyn FileIo>> {
         let Some(terminal) = current!().terminal() else {
             return_errno_with_message!(
@@ -25,12 +74,60 @@ impl Device for TtyDevice {
 
         terminal.open()
     }
+}
 
+/// Corresponds to `/dev/console` in the file system. This device represents a console to which
+/// system messages will be sent.
+pub struct SystemConsole {
+    inner: Arc<dyn Terminal>,
+}
+
+impl SystemConsole {
+    /// Returns the singleton instance of the console device.
+    pub fn singleton() -> &'static Arc<SystemConsole> {
+        static INSTANCE: Once<Arc<SystemConsole>> = Once::new();
+
+        INSTANCE.call_once(|| {
+            // TODO: Support specifying multiple TTY devices, e.g., "console=hvc0 console=tty0".
+            let console_name = KCmdlineArg::singleton()
+                .get_console_names()
+                .first()
+                .map(String::as_str)
+                .unwrap_or("tty0");
+
+            let device = match console_name {
+                "tty0" => Some(Arc::new(Tty0Device) as _),
+                "hvc0" => hvc0_device().cloned().map(|device| device as _),
+                _ => None,
+            };
+            let inner = device.unwrap_or_else(|| {
+                warn!(
+                    "'{}' console not found, falling back to 'tty0'",
+                    console_name
+                );
+                Arc::new(Tty0Device) as _
+            });
+
+            Arc::new(Self { inner })
+        })
+    }
+
+    /// Returns the terminal associated with the console device.
+    pub fn terminal(&self) -> &Arc<dyn Terminal> {
+        &self.inner
+    }
+}
+
+impl Device for SystemConsole {
     fn type_(&self) -> DeviceType {
         DeviceType::Char
     }
 
     fn id(&self) -> DeviceId {
-        DeviceId::new(MajorId::new(5), MinorId::new(0))
+        DeviceId::new(MajorId::new(5), MinorId::new(1))
+    }
+
+    fn open(&self) -> Result<Box<dyn FileIo>> {
+        self.inner.open()
     }
 }
--- a/kernel/src/device/tty/n_tty.rs
+++ b/kernel/src/device/tty/n_tty.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use alloc::{boxed::Box, sync::Arc};
+
 use aster_console::AnyConsoleDevice;
+use aster_framebuffer::DummyFramebufferConsole;
 use inherit_methods_macro::inherit_methods;
 use ostd::mm::{Infallible, VmReader, VmWriter};
 use spin::Once;
@@ -16,16 +19,20 @@ use crate::{
     process::signal::{PollHandle, Pollable},
 };
 
-pub struct ConsoleDriver {
+/// The driver for VT (virtual terminal) devices.
+//
+// TODO: This driver needs to support more features for future VT management.
+#[derive(Clone)]
+pub struct VtDriver {
     console: Arc<dyn AnyConsoleDevice>,
 }
 
-impl TtyDriver for ConsoleDriver {
+impl TtyDriver for VtDriver {
     // Reference: <https://elixir.bootlin.com/linux/v6.17/source/include/uapi/linux/major.h#L18>.
     const DEVICE_MAJOR_ID: u32 = 4;
 
     fn open(tty: Arc<Tty<Self>>) -> Result<Box<dyn FileIo>> {
-        Ok(Box::new(ConsoleFile(tty)))
+        Ok(Box::new(TtyFile(tty)))
     }
 
     fn push_output(&self, chs: &[u8]) -> Result<usize> {
@@ -50,14 +57,50 @@ impl TtyDriver for ConsoleDriver {
     }
 }
 
-struct ConsoleFile(Arc<Tty<ConsoleDriver>>);
+/// The driver for hypervisor console devices.
+#[derive(Clone)]
+pub struct HvcDriver {
+    console: Arc<dyn AnyConsoleDevice>,
+}
+
+impl TtyDriver for HvcDriver {
+    // Reference: <https://elixir.bootlin.com/linux/v6.17/source/Documentation/admin-guide/devices.txt#L2936>.
+    const DEVICE_MAJOR_ID: u32 = 229;
+
+    fn open(tty: Arc<Tty<Self>>) -> Result<Box<dyn FileIo>> {
+        Ok(Box::new(TtyFile(tty)))
+    }
+
+    fn push_output(&self, chs: &[u8]) -> Result<usize> {
+        self.console.send(chs);
+        Ok(chs.len())
+    }
+
+    fn drain_output(&self) {}
+
+    fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
+        |chs| self.console.send(chs)
+    }
+
+    fn can_push(&self) -> bool {
+        true
+    }
+
+    fn notify_input(&self) {}
+
+    fn console(&self) -> Option<&dyn AnyConsoleDevice> {
+        Some(&*self.console)
+    }
+}
+
+struct TtyFile<D>(Arc<Tty<D>>);
 
 #[inherit_methods(from = "self.0")]
-impl Pollable for ConsoleFile {
+impl<D: TtyDriver> Pollable for TtyFile<D> {
     fn poll(&self, mask: IoEvents, poller: Option<&mut PollHandle>) -> IoEvents;
 }
 
-impl InodeIo for ConsoleFile {
+impl<D: TtyDriver> InodeIo for TtyFile<D> {
     fn read_at(
         &self,
         _offset: usize,
@@ -78,7 +121,7 @@ impl InodeIo for ConsoleFile {
 }
 
 #[inherit_methods(from = "self.0")]
-impl FileIo for ConsoleFile {
+impl<D: TtyDriver> FileIo for TtyFile<D> {
     fn ioctl(&self, cmd: IoctlCmd, arg: usize) -> Result<i32>;
 
     fn check_seekable(&self) -> Result<()> {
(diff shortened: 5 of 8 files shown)


TinaZhangZW:
This draft patch retrieves the console device specified in the boot arguments. It works for me.

```
diff --git a/kernel/src/fs/mod.rs b/kernel/src/fs/mod.rs
index 3149a4279..0229ba646 100644
--- a/kernel/src/fs/mod.rs
+++ b/kernel/src/fs/mod.rs
@@ -24,9 +24,10 @@ pub mod thread_info;
 pub mod tmpfs;
 pub mod utils;
 
+use ostd::boot::boot_info;
 use aster_block::BlockDevice;
 use aster_virtio::device::block::device::BlockDevice as VirtIoBlockDevice;
-
+use alloc::format;
 use crate::{
     fs::{
         exfat::{ExfatFs, ExfatMountOptions},
@@ -106,7 +107,13 @@ pub fn init_in_first_process(ctx: &Context) {
     }
 
     // Initialize the file table for the first process.
-    let tty_path = FsPath::try_from("/dev/console").unwrap();
+    let console_path = boot_info().kernel_cmdline
+        .split(' ')
+        .find(|arg| arg.starts_with("console="))
+        .map(|arg| arg.split('=').next_back().unwrap_or("tty0"));
+    let dev_path = console_path.map(|s| format!("/dev/{}", s)).unwrap_or("/dev/tty0".to_string());
+    let tty_path = FsPath::try_from(dev_path.as_str()).unwrap();
+
     let stdin = {
         let open_args = OpenArgs::from_modes(AccessMode::O_RDONLY, mkmod!(u+r));
         fs_resolver
```

cchanging:
> However, I found one discrepancy when using the busybox tty tool over virtio-console
> It sounds that we have hardcoded "/dev/console" for applications stdin/stdout/stderr, which isn't right.

You're right, currently using hard code to open `/dev/console` in the init process (PID 1) is incorrect. Actually, after systemd is implemented, these three file descriptors should be maintained by the user-space systemd. 

I have done some experiments, and in cases with systemd, even if I delete this entire section of code that initializes the standard file descriptors, it still works. The file descriptors 0~2 of the init process point to `/dev/null`. If the current shell logs in via tty1, the `tty` command returns `tty1`; if logging in via `hvc0`, it returns `hvc0`. 

Meanwhile, I checked the file descriptors 0\~2 of the init process on the host machine, and they all point to `/dev/null`, which matches the behavior of file descriptors 0\~2 after having systemd and deleting this code section. 

So I feel we should ultimately delete this logic. However, since systemd isn't yet available on the mainline, I think we should temporarily keep this code for now? But I'm unsure why currently when running `tty` in Linux, it returns `hvc0`.

lrh2000:
> I think we should temporarily keep this code for now?

As far as I know, running a BusyBox init script (place a script at `/init` starting with `#!/bin/busybox sh`) also works under Linux. In this case, who opens stdin/stdout/stderr?

> why currently when running `tty` in Linux, it returns `hvc0`

```
> strace -e readlink tty
readlink("/proc/self/fd/0", "/dev/pts/2", 4095) = 10
/dev/pts/2
+++ exited with 0 +++
```

StevenJiang1110:
> As far as I know, running a BusyBox init script (place a script at /init starting with #!/bin/busybox sh) also works under Linux. In this case, who opens stdin/stdout/stderr?

It seems Linux will aslo [open `/dev/console` for the init process](https://elixir.bootlin.com/linux/v6.17/source/init/main.c#L1584) as stdin/stdout/stderr. So I guess our current implementation to open console for the init process should be unavoidable? Though I don't know why `tty` command returns `/dev/hvc0` on Linux.

lrh2000:
> It seems Linux will aslo [open `/dev/console` for the init process](https://elixir.bootlin.com/linux/v6.17/source/init/main.c#L1584) as stdin/stdout/stderr.

This is strange. If devtmpfs is not automatically mounted, then `/dev/console` may not even exist. How does the Linux code manage to run? 

> Though I don't know why `tty` command returns `/dev/hvc0` on Linux.

Doesn't it show up in `ls -al /proc/self/fd/0`?

cchanging:
> Doesn't it show up in ls -al /proc/self/fd/0?

The reason we feel confused is that when we open a `/dev/console`, the link read from `/proc/self/fd/x` should be exactly `/dev/console`; if Linux opens `/dev/console` as fd0 in the init process, then I think what tty returns should be `/dev/console`.

lrh2000:
> if Linux opens `/dev/console` as fd0 in the init process, then I think what tty returns should be `/dev/console`.

You can execute `tty < /dev/console`. It will return `/dev/console`.
```
# tty < /dev/console
/dev/console
```

StevenJiang1110:
> This is strange. If devtmpfs is not automatically mounted, then /dev/console may not even exist. How does the Linux code manage to run?

Before calling [console_on_rootfs](https://elixir.bootlin.com/linux/v6.17/C/ident/console_on_rootfs), [do_basic_setup](https://elixir.bootlin.com/linux/v6.17/C/ident/do_basic_setup) will be called, will internally calls [driver_init](https://elixir.bootlin.com/linux/v6.17/C/ident/driver_init) -> [devtmpfs_init](https://elixir.bootlin.com/linux/v6.17/C/ident/devtmpfs_init), and will create [a vfs mount](https://elixir.bootlin.com/linux/v6.17/source/drivers/base/devtmpfs.c#L484) for devtmpfs, will this make the devtmpfs visible to user space?

lrh2000:
> Before calling [console_on_rootfs](https://elixir.bootlin.com/linux/v6.17/C/ident/console_on_rootfs), [do_basic_setup](https://elixir.bootlin.com/linux/v6.17/C/ident/do_basic_setup) will be called, will internally calls [driver_init](https://elixir.bootlin.com/linux/v6.17/C/ident/driver_init) -> [devtmpfs_init](https://elixir.bootlin.com/linux/v6.17/C/ident/devtmpfs_init)

It may be mounted in [`prepare_namespace`](https://elixir.bootlin.com/linux/v6.18-rc6/source/init/do_mounts.c#L464), [`devtmpfs_mount`](https://elixir.bootlin.com/linux/v6.18-rc6/source/drivers/base/devtmpfs.c#L361). The mount happens conditionally under [`CONFIG_DEVTMPFS_MOUNT`](https://elixir.bootlin.com/linux/v6.18-rc6/source/drivers/base/devtmpfs.c#L42).

However, [`prepare_namespace` happens after `console_on_rootfs`](https://elixir.bootlin.com/linux/v6.18-rc6/source/init/main.c#L1598-L1610)?

> , and will create [a vfs mount](https://elixir.bootlin.com/linux/v6.17/source/drivers/base/devtmpfs.c#L484) for devtmpfs, will this make the devtmpfs visible to user space?

It appears to be an internal kernel mount. I don't think it's visible. The mount does not accept a path, i.e., no parameters specifying `/dev`?

StevenJiang1110:
> It appears to be an internal kernel mount. I don't think it's visible. The mount does not accept a path, i.e., no parameters specifying /dev?

Oh yes, this step won't mount devtmpfs.

But I found devtmpfs will be mounted in a [late step](https://elixir.bootlin.com/linux/v6.17/source/drivers/base/devtmpfs.c#L502) in [devtmpfs_init](https://elixir.bootlin.com/linux/v6.17/C/ident/devtmpfs_init), it will spawn a kernel thread to run [devtmpfsd](https://elixir.bootlin.com/linux/v6.17/C/ident/devtmpfsd), which will call [devtmpfs_setup](https://elixir.bootlin.com/linux/v6.17/C/ident/devtmpfs_setup), and will create an init mount for devtmpfs under `/`. Though I can't stant why the path is not `/dev`.

lrh2000:
> will create an init mount for devtmpfs under `/`.

I suppose it's still an internal mount because https://elixir.bootlin.com/linux/v6.17/source/drivers/base/devtmpfs.c#L421:
```
  err = ksys_unshare(CLONE_NEWNS);
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -92,6 +92,38 @@ impl AnyConsoleDevice for FramebufferConsole {
     }
 }
 
+/// A dummy console device.
+///
+/// This is used when no framebuffer is available. All operations are no-ops.
+#[derive(Debug)]
+pub struct DummyFramebufferConsole;
```
lrh2000: Could you please explain why you define this type here?

By "here", I mean it is defined between `impl AnyConsoleDevice for FramebufferConsole` and `impl FramebufferConsole`. This is a very odd place to define a new type that is not relevant to the `FramebufferConsole` type at all.

In fact, I think this type might be worth its own file because it has little to do with anything in `console.rs`.

kernel/comps/framebuffer/src/console.rs
```diff
@@ -92,6 +92,38 @@ impl AnyConsoleDevice for FramebufferConsole {
...
+        None
+    }
+
+    fn set_keyboard_mode(&self, _mode: KeyboardMode) -> bool {
+        true
+    }
+
+    fn keyboard_mode(&self) -> Option<KeyboardMode> {
+        None
+    }
```
lrh2000: If you return `true` in `set_*_mode`, then you can't do nothing and return `None` in `*_mode`. You may want to save the mode somewhere?

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
+pub struct DevConsole {
+    // TODO: Support specifying multiple tty devices.
+    inner: Arc<dyn Terminal>,
+}
+
+impl DevConsole {
+    /// Returns the singleton instance of the console device.
+    pub fn singleton() -> &'static Arc<DevConsole> {
+        static INSTANCE: Once<Arc<DevConsole>> = Once::new();
+        INSTANCE.call_once(|| {
```
lrh2000, line 90: Insert a blank line because there are multiple blank lines inside `call_once`?

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
+                    }
+                }
+                _ => {
+                    warn!(
+                        "unsupported console device '{}', falling back to 'tty0'",
+                        console_name
+                    );
+                    Arc::new(Tty0Device) as _
+                }
+            };
```
lrh2000, line 102:
```rust
let device = match console_name {
    "tty0" => Some(Arc::new(Tty0Device) as _),
    "hcv0" => hvc0_device().cloned(),
    _ => None,
};
let inner = device.unwrap_or_else(|| {
    warn!("'{}' console not found, falling back to 'tty0'", console_name);
    Arc::new(Tty0Device) as _
});
```
Is this clearer?

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
+        return_errno_with_message!(Errno::EINVAL, "cannot read tty0 device");
+    }
+
+    fn write(&self, _reader: &mut VmReader, _status_flags: StatusFlags) -> Result<usize> {
+        return_errno_with_message!(Errno::EINVAL, "cannot write tty0 device");
+    }
+}
+
+/// Corresponds to `/dev/console` in the file system.
+pub struct DevConsole {
```
lrh2000: I think `SystemConsole` is a better name.
```rust
/// Corresponds to `/dev/console` in the file system. This device represents a console to which
/// system messages will be sent.
pub struct SystemConsole {
```

Search `/dev/console` in https://www.kernel.org/doc/html/latest/admin-guide/devices.html.

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
+    }
+}
+
+impl Pollable for Tty0Device {
+    fn poll(&self, _mask: IoEvents, _poller: Option<&mut PollHandle>) -> IoEvents {
+        IoEvents::empty()
+    }
+}
+
+impl FileIo for Tty0Device {
```
lrh2000: Please remove `impl {Pollable, FileIo} for {Tty0Device, DevConsole}` during rebase. They're no longer needed now.

Makefile
```diff
@@ -22,6 +22,13 @@ FEATURES ?=
 NO_DEFAULT_FEATURES ?= 0
 COVERAGE ?= 0
 ENABLE_BASIC_TEST ?= false
+# Specify the primary system console (supported: hvc0, tty0).
+# - hvc0: For QEMU virtual environments.
+# - tty0: Current virtual terminal.
+# System will automatically fallback to tty0 if hvc0 is not available.
```
lrh2000: - "System will" -> "Asterinas will"
 - "fallback" -> "fall back"

Makefile
```diff
@@ -22,6 +22,13 @@ FEATURES ?=
 NO_DEFAULT_FEATURES ?= 0
 COVERAGE ?= 0
 ENABLE_BASIC_TEST ?= false
+# Specify the primary system console (supported: hvc0, tty0).
+# - hvc0: For QEMU virtual environments.
+# - tty0: Current virtual terminal.
```
lrh2000:
```makefile
# - hvc0: The virtio-console terminal.
# - tty0: The active virtual terminal (VT).
```

Makefile
```diff
@@ -22,6 +22,13 @@ FEATURES ?=
...
+# System will automatically fallback to tty0 if hvc0 is not available.
+# Note that currently virtual terminal (tty0) can only work with
```
lrh2000:
```suggestion
# Note that currently the virtual terminal (tty0) can only work with
```

kernel/src/device/tty/n_tty.rs
```diff
@@ -41,54 +48,109 @@ impl TtyDriver for ConsoleDriver {
...
+        .find(|(name, _)| name.as_str() == aster_framebuffer::CONSOLE_NAME)
+        .map(|(_, device)| device.clone())
+        .unwrap_or(Arc::new(DummyFramebufferConsole));
+
+    let tty1 = {
+        let driver = VtDriver {
+            console: fb_device.clone(),
+        };
+        Tty::new(1, driver)
+    };
```
lrh2000: Be consistent to the `hcv0` code:
```
        let driver = VtDriver {
            console: fb_device.clone(),
        };
        let tty1 = Tty::new(1, driver);
```

kernel/src/device/tty/n_tty.rs
```diff
@@ -41,54 +48,109 @@ impl TtyDriver for ConsoleDriver {
...
+///
+/// Returns `None` if the device is not found nor initialized.
+pub fn hvc0_device() -> Option<&'static Arc<Tty<HvcDriver>>> {
+    HVC0.get()
+}
+
+pub(in crate::device) fn init_in_first_process() {
+    let devices = aster_console::all_devices();
+
+    // Initialize the `hvc0` device if virtio console is available.
```
lrh2000:
```suggestion
    // Initialize the `hvc0` device if the virtio console is available.
```

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
+    }
+
+    fn write(&self, _reader: &mut VmReader, _status_flags: StatusFlags) -> Result<usize> {
+        return_errno_with_message!(Errno::EINVAL, "cannot write tty0 device");
+    }
+}
+
+/// Corresponds to `/dev/console` in the file system.
+pub struct DevConsole {
+    // TODO: Support specifying multiple tty devices.
```
lrh2000:
```rust
    // TODO: Support specifying multiple TTY devices.
```
What does this TODO want to mean? I don't really understand.

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
     }
 }
+
+/// Corresponds to `/dev/tty0` in the file system. This device represents the active virtual
+/// terminal.
+pub struct Tty0Device;
+
+impl Tty0Device {
+    fn active_vt(&self) -> &Arc<Tty<VtDriver>> {
+        // Currently there is only one virtual terminal tty1.
```
lrh2000:
```suggestion
        // Currently there is only one virtual terminal `tty1`.
```

kernel/src/device/tty/device.rs
```diff
@@ -53,3 +59,127 @@ impl FileIo for TtyDevice {
...
+    }
+
+    /// Returns the terminal associated with the console device.
+    pub fn terminal(&self) -> &Arc<dyn Terminal> {
+        &self.inner
+    }
+}
+
+impl Device for DevConsole {
+    fn open(&self) -> Option<Result<Arc<dyn FileIo>>> {
```
lrh2000: Please reorder the methods:
```rust
/// The abstraction of a device.
pub trait Device: Send + Sync + 'static {
    /// Returns the device type.
    fn type_(&self) -> DeviceType;

    /// Returns the device ID.
    fn id(&self) -> DeviceId;

    /// Opens a device.
    fn open(&self) -> Result<Box<dyn FileIo>>;
}
```

Oh, this seems to be an existing issue in this file. Maybe we can fix the order of the methods for the existing type `TtyDevice` as well.

kernel/src/device/tty/device.rs
```diff
@@ -1,16 +1,22 @@
...
     },
     prelude::*,
-    process::signal::{PollHandle, Pollable},
+    process::{
+        signal::{PollHandle, Pollable},
+        JobControl, Terminal,
+    },
 };
 
 /// Corresponds to `/dev/tty` in the file system. This device represents the controlling terminal
```
lrh2000: Although it is not introduced in this PR, can you update line 23:
```diff
-/// of the session of current process.
+/// of the session of the current process.
```
?

kernel/src/device/tty/n_tty.rs
```diff
@@ -41,54 +48,109 @@ impl TtyDriver for ConsoleDriver {
...
-        devices
-    };
+/// The driver for hypervisor console devices.
+#[derive(Clone)]
+pub struct HvcDriver {
+    console: Arc<dyn AnyConsoleDevice>,
+}
+
+impl TtyDriver for HvcDriver {
+    // Reference: <https://github.com/torvalds/linux/blob/24172e0d79900908cf5ebf366600616d29c9b417/Documentation/admin-guide/devices.txt#L2936>.
```
lrh2000: Use a link starting with `https://elixir.bootlin.com/linux/v6.17/source/`.

lrh2000 (commented):
I think the current implementation is good according to the Linux documentation at https://www.kernel.org/doc/html/latest/admin-guide/devices.html, as far as I can tell. The comments are mostly minor, addressing some style issues.

I think the link to the Linux document should be in a more visible location. Maybe at the top of `tty/device.rs`?

kernel/src/device/tty/device.rs
cchanging: The meaning of this TODO is that we should later support specifying multiple consoles in the boot parameters, similar to Linux. For example, using `console=tty0, console=hvc0` would indicate that kernel messages should be sent to multiple consoles. Maybe put this TODO to the `Makefile`?

lrh2000: > For example, using `console=tty0, console=hvc0` would indicate that kernel messages should be sent to multiple consoles.

You can move this TODO comment to the location where you parse the kernel cmdline and include the specific example there. I wasn't aware that Linux accepts multiple `console=` parameters. Thanks for sharing your knowledge!

kernel/comps/framebuffer/src/console.rs
cchanging: > If you return true in set_*_mode, then you can't do nothing and return None in *_mode. 

So, wouldn't a more appropriate approach be to have the `set_*` method return `false`? Because a dummy console probably shouldn't have any mode that can be set.

lrh2000: I don't know which one is more appropriate, though. I think both approaches are correct (but they may or may not satisfy userspace applications, which is another story), but the current approach is incorrect. So generally speaking, I'm okay with your proposal as well.

But I guess Linux won't return errors in these cases. However, it needs additional experiments to confirm.

kernel/src/device/tty/device.rs
```diff
@@ -25,12 +43,92 @@ impl Device for TtyDevice {
...
+}
+
+impl SystemConsole {
+    /// Returns the singleton instance of the console device.
+    pub fn singleton() -> &'static Arc<SystemConsole> {
+        static INSTANCE: Once<Arc<SystemConsole>> = Once::new();
+
+        INSTANCE.call_once(|| {
+            // TODO: Support specifying multiple TTY devices.
+            // e.g. "console=hvc0, console=tty0"
```
lrh2000: > console=hvc0,

Is this an accurate example? I'm not sure if the Linux kernel cmdline can include a comma as a separator. I thought it was a typo, but maybe it isn't.
```rust
            // TODO: Support specifying multiple TTY devices, e.g., "console=hvc0 console=tty0".
```

kernel/src/device/tty/n_tty.rs
```diff
@@ -90,54 +97,147 @@ impl FileIo for ConsoleFile {
...
+                let mut chs = vec![0u8; reader.remain()];
+                reader.read(&mut VmWriter::from(chs.as_mut_slice()));
+                let _ = hvc0.push_input(chs.as_slice());
+            },
+        )));
+    }
+
+    // Initialize the `tty1` device.
+
+    let fb_device = devices
```
lrh2000: `fb_device` -> `fb_console`. The name is more consistent with `virtio_console`.

kernel/src/device/tty/n_tty.rs
```diff
@@ -90,54 +97,147 @@ impl FileIo for ConsoleFile {
...
+    fn console(&self) -> Option<&dyn AnyConsoleDevice> {
+        Some(&*self.console)
+    }
 }
 
-fn create_n_tty(index: u32, device: Arc<dyn AnyConsoleDevice>) -> Arc<Tty<ConsoleDriver>> {
-    let driver = ConsoleDriver {
-        console: device.clone(),
-    };
+struct HvcFile(Arc<Tty<HvcDriver>>);
```
lrh2000: You may want to have a `struct TtyFile<D>(Arc<Tty<D>>)` type to avoid repeating the code twice?

kernel/src/device/tty/device.rs
```diff
@@ -25,12 +43,92 @@ impl Device for TtyDevice {
...
+            let console_name = boot_info()
+                .kernel_cmdline
+                .split_whitespace()
+                .find(|item| item.starts_with("console="))
+                .and_then(|value| value.split('=').nth(1))
```
lrh2000: Use [`str::split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once) here?

It also seems that we have `kernel/src/kcmdline.rs`, but there is no reasonable way to access the parsed cmdline here?

kernel/src/device/mod.rs
```diff
@@ -75,13 +79,20 @@ pub fn get_device(devid: DeviceId) -> Result<Arc<dyn Device>> {
...
-        }
+        (4, 0) => Ok(Arc::new(tty::Tty0Device)),
+        (4, 1) => Ok(tty::tty1_device().clone()),
         (5, 0) => Ok(Arc::new(tty::TtyDevice)),
+        (5, 1) => Ok(tty::SystemConsole::singleton().clone()),
+        (229, 0) => {
+            if let Some(hvc0) = tty::hvc0_device() {
+                Ok(hvc0.clone())
+            } else {
+                Err(Error::with_message(
```
lrh2000: I think this is simpler:
```
        (229, 0) => tty::hvc0_device().cloned().ok_or_else(|_| {
            Err(Error::with_message(
                Errno::ENODEV,
                "the hvc0 device is not available",
            ))
        }),
```

kernel/comps/framebuffer/src/dummy_console.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Debug)]
+pub struct DummyFramebufferConsole;
+
+impl AnyConsoleDevice for DummyFramebufferConsole {
+    fn send(&self, _buf: &[u8]) {}
+
+    fn register_callback(&self, _callback: &'static ConsoleCallback) {}
+
+    fn set_font(&self, _font: BitmapFont) -> Result<(), ConsoleSetFontError> {
+        Ok(())
```
lrh2000: Given that `set*mode` returns `false`, you may want to return `Err(...)` here?

Again, I'm not sure if this is the correct behaviour for Linux. However, we need to be consistent unless there are special reasons, such as application requirements. For example, systemd may try to change the VT mode and fail, but I don't know, so you can decide for yourself.

kernel/src/kcmdline.rs
```diff
@@ -43,18 +46,39 @@ pub struct KCmdlineArg {
...
         &self.initproc.argv
     }
+
     /// Gets the environment vector(envp) of the initprocess.
     pub fn get_initproc_envp(&self) -> &Vec<CString> {
         &self.initproc.envp
     }
+
+    /// Gets the values of a given key from the initprocess environment.
+    pub fn get_initproc_env_values<'a>(&'a self, key: &'a str) -> impl Iterator<Item = &'a str> {
```
lrh2000: Should `console=` live in the initproc's envp? I don't think so. We should be able to modify the code below:
```rust
            if let Some(value) = value {
                // The option has a value.
                match option {
                    "init" => { /* ... */ }
                    "console" => todo!()
```
Then, we should have something like:
```rust
/// Gets the device name of the system console.
fn get_console_name(&self) {
   todo!()
}
```

Would this be more correct?

kernel/src/device/mod.rs
```diff
@@ -75,13 +79,14 @@ pub fn get_device(devid: DeviceId) -> Result<Arc<dyn Device>> {
...
-            Ok(tty.clone())
-        }
+        (4, 0) => Ok(Arc::new(tty::Tty0Device)),
+        (4, 1) => Ok(tty::tty1_device().clone()),
         (5, 0) => Ok(Arc::new(tty::TtyDevice)),
+        (5, 1) => Ok(tty::SystemConsole::singleton().clone()),
+        (229, 0) => tty::hvc0_device()
+            .cloned()
+            .map(|device| device as _)
+            .ok_or_else(|| Error::with_message(Errno::ENODEV, "the hvc0 device is not available")),
```
lrh2000, line 89: Oh well, since line 92 uses `ok_or`, we should either use `ok_or` here or change line 92. I didn't notice it in my previous comment.

kernel/src/device/tty/device.rs
```diff
@@ -25,12 +43,87 @@ impl Device for TtyDevice {
...
+}
+
+impl SystemConsole {
+    /// Returns the singleton instance of the console device.
+    pub fn singleton() -> &'static Arc<SystemConsole> {
+        static INSTANCE: Once<Arc<SystemConsole>> = Once::new();
+
+        INSTANCE.call_once(|| {
+            // TODO: Support specifying multiple TTY devices, e.g., "console=hvc0 console=tty0".
+            let console_name: &'static str = &KCmdlineArg::singleton().get_console_names()[0];
```
lrh2000: This will panic if no console is specified? I don't think it is the correct behavior.

kernel/src/device/tty/n_tty.rs
```diff
@@ -90,54 +133,71 @@ impl FileIo for ConsoleFile {
...
 
-    let tty = Tty::new(index, driver);
-    let tty_cloned = tty.clone();
+    // Initialize the `tty1` device.
+
+    let fb_console = devices
+        .iter()
+        .find(|(name, _)| name.as_str() == aster_framebuffer::CONSOLE_NAME)
+        .map(|(_, device)| device.clone())
+        .unwrap_or(Arc::new(DummyFramebufferConsole));
```
lrh2000:
```rust
        .unwrap_or_else(|| Arc::new(DummyFramebufferConsole));
```
There is no need to create an `Arc` if we can find the device.

kernel/src/device/mod.rs
```diff
@@ -43,18 +41,24 @@ pub fn init_in_first_process(ctx: &Context) -> Result<()> {
...
+    let console = tty::SystemConsole::singleton().clone();
     add_node(console, "console", &fs_resolver)?;
 
-    for (index, tty) in tty::iter_n_tty().enumerate() {
-        add_node(tty.clone(), &format!("tty{}", index), &fs_resolver)?;
+    let tty0 = Arc::new(tty::Tty0Device);
+    add_node(tty0, "tty0", &fs_resolver)?;
+
+    if let Some(hvc0) = tty::hvc0_device() {
+        add_node(hvc0.clone(), "hvc0", &fs_resolver)?;
```
lrh2000, line 59: `hvc0` is defined after `tty1`. Let's make the order consistent here.
```rust
   let tty1 = tty::tty1_device().clone();
   add_node(tty1, "tty1", &fs_resolver)?;

   if let Some(hvc0) = tty::hvc0_device() {
       add_node(hvc0.clone(), "hvc0", &fs_resolver)?;
   }
```

kernel/src/kcmdline.rs
```diff
@@ -38,23 +41,39 @@ pub enum ModuleArg {
...
     pub fn get_initproc_argv(&self) -> &Vec<CString> {
         &self.initproc.argv
     }
+
     /// Gets the environment vector(envp) of the initprocess.
     pub fn get_initproc_envp(&self) -> &Vec<CString> {
         &self.initproc.envp
     }
+
+    /// Gets the list of system console names specified in the command line.
```
lrh2000: `list` -> `vector`. It is slightly more correct and consistent with other doc comments here.

kernel/src/device/tty/device.rs
```diff
@@ -1,20 +1,38 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! TTY devices.
+//!
+//! This module implements TTY devices such as `/dev/tty`, `/dev/tty0`, and `/dev/console`.
```
lrh2000: Honestly, I think the current implementation still has a lot of inconsistent ways of arranging the various TTY devices. That's boring but annoying.

Let's sort everything according to their device ID, like what we've already done in the [`get_device`](https://github.com/cchanging/asterinas/blob/01a7f3e8459110c823f6af4b15acee193ede4f27/kernel/src/device/mod.rs#L77) method:
```patch
diff --git a/Makefile b/Makefile
index 3f13c8e4e..40d95cf2f 100644
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ ENABLE_BASIC_TEST ?= false
 # - tty0: The active virtual terminal (VT).
 # Asterinas will automatically fall back to tty0 if hvc0 is not available.
 # Note that currently the virtual terminal (tty0) can only work with
-# linux-efi-handover64 boot protocol.
+# linux-efi-handover64 and linux-efi-pe64 boot protocol.
 CONSOLE ?= hvc0
 # End of global build options.
 
diff --git a/kernel/src/device/mod.rs b/kernel/src/device/mod.rs
index 3405b0fe9..850536ff2 100644
--- a/kernel/src/device/mod.rs
+++ b/kernel/src/device/mod.rs
@@ -43,18 +43,18 @@ pub fn init_in_first_process(ctx: &Context) -> Result<()> {
 
     tty::init_in_first_process();
 
-    let tty = Arc::new(tty::TtyDevice);
-    add_node(tty, "tty", &fs_resolver)?;
-
-    let console = tty::SystemConsole::singleton().clone();
-    add_node(console, "console", &fs_resolver)?;
-
     let tty0 = Arc::new(tty::Tty0Device);
     add_node(tty0, "tty0", &fs_resolver)?;
 
     let tty1 = tty::tty1_device().clone();
     add_node(tty1, "tty1", &fs_resolver)?;
 
+    let tty = Arc::new(tty::TtyDevice);
+    add_node(tty, "tty", &fs_resolver)?;
+
+    let console = tty::SystemConsole::singleton().clone();
+    add_node(console, "console", &fs_resolver)?;
+
     if let Some(hvc0) = tty::hvc0_device() {
         add_node(hvc0.clone(), "hvc0", &fs_resolver)?;
     }
diff --git a/kernel/src/device/tty/device.rs b/kernel/src/device/tty/device.rs
index 86f7d07a5..a31f07af8 100644
--- a/kernel/src/device/tty/device.rs
+++ b/kernel/src/device/tty/device.rs
@@ -2,7 +2,7 @@
 
 //! TTY devices.
 //!
-//! This module implements TTY devices such as `/dev/tty`, `/dev/tty0`, and `/dev/console`.
+//! This module implements TTY devices such as `/dev/tty0`, `/dev/tty`, and `/dev/console`.
 //!
 //! Reference: <https://www.kernel.org/doc/html/latest/admin-guide/devices.html>
 
@@ -20,31 +20,6 @@ use crate::{
     process::{JobControl, Terminal},
 };
 
-/// Corresponds to `/dev/tty` in the file system. This device represents the controlling terminal
-/// of the session of the current process.
-pub struct TtyDevice;
-
-impl Device for TtyDevice {
-    fn type_(&self) -> DeviceType {
-        DeviceType::Char
-    }
-
-    fn id(&self) -> DeviceId {
-        DeviceId::new(MajorId::new(5), MinorId::new(0))
-    }
-
-    fn open(&self) -> Result<Box<dyn FileIo>> {
-        let Some(terminal) = current!().terminal() else {
-            return_errno_with_message!(
-                Errno::ENOTTY,
-                "the process does not have a controlling terminal"
-            );
-        };
-
-        terminal.open()
-    }
-}
-
 /// Corresponds to `/dev/tty0` in the file system. This device represents the active virtual
 /// terminal.
 pub struct Tty0Device;
@@ -76,6 +51,31 @@ impl Terminal for Tty0Device {
     }
 }
 
+/// Corresponds to `/dev/tty` in the file system. This device represents the controlling terminal
+/// of the session of the current process.
+pub struct TtyDevice;
+
+impl Device for TtyDevice {
+    fn type_(&self) -> DeviceType {
+        DeviceType::Char
+    }
+
+        DeviceId::new(MajorId::new(5), MinorId::new(0))
+    }
+
+        let Some(terminal) = current!().terminal() else {
+                Errno::ENOTTY,
+                "the process does not have a controlling terminal"
+            );
+        };
+
+        terminal.open()
+    }
+}
+
 /// Corresponds to `/dev/console` in the file system. This device represents a console to which
 /// system messages will be sent.
 pub struct SystemConsole {
diff --git a/kernel/src/device/tty/n_tty.rs b/kernel/src/device/tty/n_tty.rs
index c89a467db..c41aa6aa7 100644
--- a/kernel/src/device/tty/n_tty.rs
+++ b/kernel/src/device/tty/n_tty.rs
@@ -156,6 +156,28 @@ pub fn hvc0_device() -> Option<&'static Arc<Tty<HvcDriver>>> {
 pub(in crate::device) fn init_in_first_process() {
     let devices = aster_console::all_devices();
 
+    // Initialize the `tty1` device.
+
+    let fb_console = devices
+        .iter()
+        .find(|(name, _)| name.as_str() == aster_framebuffer::CONSOLE_NAME)
+        .map(|(_, device)| device.clone())
+        .unwrap_or_else(|| Arc::new(DummyFramebufferConsole));
+
+    let driver = VtDriver {
+        console: fb_console.clone(),
+    };
+    let tty1 = Tty::new(1, driver);
+    TTY1.call_once(|| tty1.clone());
+
+    fb_console.register_callback(Box::leak(Box::new(
+        move |mut reader: VmReader<Infallible>| {
+            let mut chs = vec![0u8; reader.remain()];
+            reader.read(&mut VmWriter::from(chs.as_mut_slice()));
+            let _ = tty1.push_input(chs.as_slice());
+        },
+    )));
+
     // Initialize the `hvc0` device if the virtio console is available.
 
     let virtio_console = devices
@@ -178,26 +200,4 @@ pub(in crate::device) fn init_in_first_process() {
             },
         )));
     }
-
-    // Initialize the `tty1` device.
-
-    let fb_console = devices
-        .iter()
-        .find(|(name, _)| name.as_str() == aster_framebuffer::CONSOLE_NAME)
-        .map(|(_, device)| device.clone())
-        .unwrap_or_else(|| Arc::new(DummyFramebufferConsole));
-
-    let driver = VtDriver {
-        console: fb_console.clone(),
-    };
-    let tty1 = Tty::new(1, driver);
-    TTY1.call_once(|| tty1.clone());
-
-    fb_console.register_callback(Box::leak(Box::new(
-        move |mut reader: VmReader<Infallible>| {
-            let mut chs = vec![0u8; reader.remain()];
-            reader.read(&mut VmWriter::from(chs.as_mut_slice()));
-            let _ = tty1.push_input(chs.as_slice());
-        },
-    )));
 }
```

Note that Linux-efi-pe64 boot protocol also supports the framebuffer console. I used it to test on real hardware, i.e., my laptop.

Other than that, it just changes the location of things without making any functional changes.

lrh2000 (approved):
Thanks! The code looks good to me.

I tested this PR on my laptop and confirmed that everything works smoothly.

A final comment to adjust some definitions' location. Hopefully it can make sense.

PR #2567, opened 2025-11-07
Use private memory for TDX `get_report` TDCALL to enhance security

taosue:
# Problem
The current `get_report` implementation uses shared memory to expose sensitive report data to the host, which introduces security risks. As a `TDCALL` (Trusted Domain Call), `get_report` should adhere to stricter isolation standards by utilizing private memory instead of shared memory.

# Changes
- Use private memory via `USegment` (single page) for get_report, aligning with Linux's approach. This also removes an extra data copy.
https://github.com/torvalds/linux/blob/4a0c9b3391999818e2c5b93719699b255be1f682/drivers/virt/coco/tdx-guest/tdx-guest.c#L53-L65
- Apply consistent size/alignment rules and compile-time checks to untyped metadata, matching typed metadata.

### DIFF
--- a/ostd/src/mm/frame/meta.rs
+++ b/ostd/src/mm/frame/meta.rs
@@ -82,7 +82,7 @@ pub(in crate::mm) struct MetaSlot {
     ///  - the implementation can simply cast a `*const MetaSlot`
     ///    to a `*const AnyFrameMeta` for manipulation;
     ///  - if the metadata need special alignment, we can provide
-    ///    at most `PAGE_METADATA_ALIGN` bytes of alignment;
+    ///    at most [`FRAME_METADATA_MAX_ALIGN`] bytes of alignment;
     ///  - the subsequent fields can utilize the padding of the
     ///    reference count to save space.
     ///
@@ -161,6 +161,17 @@ pub unsafe trait AnyFrameMeta: Any + Send + Sync {
     }
 }
 
+/// Checks that a frame metadata type has valid size and alignment.
+#[macro_export]
+macro_rules! check_frame_meta_layout {
+    ($t:ty) => {
+        $crate::const_assert!(size_of::<$t>() <= $crate::mm::frame::meta::FRAME_METADATA_MAX_SIZE);
+        $crate::const_assert!(
+            $crate::mm::frame::meta::FRAME_METADATA_MAX_ALIGN % align_of::<$t>() == 0
+        );
+    };
+}
+
 /// Makes a structure usable as a frame metadata.
 #[macro_export]
 macro_rules! impl_frame_meta_for {
@@ -169,10 +180,7 @@ macro_rules! impl_frame_meta_for {
         // SAFETY: `on_drop` won't read the page.
         unsafe impl $crate::mm::frame::meta::AnyFrameMeta for $t {}
 
-        $crate::const_assert!(size_of::<$t>() <= $crate::mm::frame::meta::FRAME_METADATA_MAX_SIZE);
-        $crate::const_assert!(
-            $crate::mm::frame::meta::FRAME_METADATA_MAX_ALIGN % align_of::<$t>() == 0
-        );
+        $crate::check_frame_meta_layout!($t);
     };
 }
 
--- a/ostd/src/mm/frame/untyped.rs
+++ b/ostd/src/mm/frame/untyped.rs
@@ -46,6 +46,8 @@ macro_rules! impl_untyped_frame_meta_for {
             }
         }
         impl $crate::mm::frame::untyped::AnyUFrameMeta for $t {}
+
+        $crate::check_frame_meta_layout!($t);
     };
     // Implement with a customized drop function.
     ($t:ty, $body:expr) => {
@@ -60,6 +62,8 @@ macro_rules! impl_untyped_frame_meta_for {
             }
         }
         impl $crate::mm::frame::untyped::AnyUFrameMeta for $t {}
+
+        $crate::check_frame_meta_layout!($t);
     };
 }
 


kernel/src/device/tdxguest/mod.rs
```diff
@@ -175,36 +181,34 @@ fn handle_get_report(arg: usize) -> Result<i32> {
     let report = tdx_get_report(&user_request.report_data)?;
 
     let tdx_report_vaddr = arg + TDX_REPORTDATA_LEN;
-    user_space.write_bytes(tdx_report_vaddr, &mut VmReader::from(report.as_ref()))?;
+    user_space.write_bytes(tdx_report_vaddr, &mut report.reader().limit(TDX_REPORT_LEN))?;
     Ok(0)
 }
 
-fn tdx_get_report(inblob: &[u8]) -> Result<Box<[u8]>> {
+const TDREPORT_REPORTDATA_OFFSET: usize = 128;
```
tatetian: First off, the name of this magic number fails to do a decent job on explaining and justifying the validity and rationale of choosing this magic value. The following comment seems to provide the necessary justification for this magic number:

```rust
// `TDREPORT_REPORTDATA_OFFSET` which is the offset of `REPORTDATA` in the report structure
```

But I am even more confused by the comment. Does `report: USegment` follow the report structure? What exactly is the format? Is it defined by this module, or `tdx-guest`, or the TDX spec? What `REPORTDATA` refers to exactly? This is the first time that the name appears.

Second, move this const definition into `tdx_get_report` because that is the only place that the const is ever used. In general, if a type/const/function has only one user, then moving the code close to this only user would make the code more clear. In addition, narrowing the scope or visibility of a type/const/function also helps with clarity.

In conclusion, based on my best guess, the following code expresses your original intention but in a more clear way:

```rust
fn tdx_get_report(inblob: &[u8]) -> Result<USegment> {
    if inblob.len() != TDX_REPORTDATA_LEN {
        return_errno_with_message!(Errno::EINVAL, "Invalid inblob length");
    }

    let report: USegment = FrameAllocOptions::new().alloc_segment(1)?.into();

    // Use `inblob` as the data associated with the report.
    let report_data_paddr = {
        // The `get_report` function requires report data to be 64-byte aligned.
        const REPORT_DATA_OFFSET: usize = align_up(TDX_REPORT_LEN, 64);
        report
            .write_bytes(REPORT_DATA_OFFSET, &inblob)
            .unwrap();
        report.paddr() + REPORT_DATA_OFFSET
    };

    // FIXME: The `get_report` API from the `tdx_guest` crate should have been marked `unsafe`
    // because it has no way to determine if the input physical address is safe or not.
    get_report(report.paddr() as u64, report_data_paddr);
    Ok(report)
}
```

kernel/src/device/tdxguest/mod.rs
```diff
@@ -175,36 +181,34 @@ fn handle_get_report(arg: usize) -> Result<i32> {
...
+
+fn tdx_get_report(inblob: &[u8]) -> Result<USegment> {
```
tatetian, line 249: Changing the return type from `Box<[u8]>` to `USegment` makes less clear the length of the returned report. Clarify this point by adding some Rust doc.

```rust
/// Gets the TDX report given the specified data in `inblob`.
///
/// The first `TDX_REPORT_LEN` bytes of data in the returned `USegment` is the report.
/// The rest in `USegment` should be ignored.
fn tdx_get_report(inblob: &[u8]) -> Result<USegment> {
```

kernel/src/device/tdxguest/mod.rs
```diff
@@ -127,7 +130,10 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
     field_ptr!(&report_ptr, TdxQuoteHdr, status).write(&0u64)?;
     field_ptr!(&report_ptr, TdxQuoteHdr, in_len).write(&(TDX_REPORT_LEN as u32))?;
     field_ptr!(&report_ptr, TdxQuoteHdr, out_len).write(&0u32)?;
-    buf.write_bytes(size_of::<TdxQuoteHdr>(), &report)?;
+    buf.write(
+        size_of::<TdxQuoteHdr>(),
+        &mut report.reader().limit(TDX_REPORT_LEN).clone().to_fallible(),
```
tatetian: Are you sure that `clone` is necessary?

tatetian (commented):
Good job on fixing this bug!

kernel/src/device/tdxguest/mod.rs
taosue: > Does `report: USegment` follow the report structure? What exactly is the format? Is it defined by this module, or `tdx-guest`, or the TDX spec? What `REPORTDATA` refers to exactly? This is the first time that the name appears.
> 

`report: USegment` is the report structure which is written by TDX Module. From TDX spec, report structure is `TDREPORT_STRUCT` and the offset 128 is `REPORTDATA` which stores the report data provided by user, i.e. `inblob`.

> In conclusion, based on my best guess, the following code expresses your original intention but in a more clear way:
> 
> ```rust
> fn tdx_get_report(inblob: &[u8]) -> Result<USegment> {
>     if inblob.len() != TDX_REPORTDATA_LEN {
>         return_errno_with_message!(Errno::EINVAL, "Invalid inblob length");
>     }
> 
>     let report: USegment = FrameAllocOptions::new().alloc_segment(1)?.into();
> 
>     // Use `inblob` as the data associated with the report.
>     let report_data_paddr = {
>         // The `get_report` function requires report data to be 64-byte aligned.
>         const REPORT_DATA_OFFSET: usize = align_up(TDX_REPORT_LEN, 64);

Sorry, I didn't describe it clearly here. Is it better to use the offset of `REPORTDATA`(128)? Because
1. The Linux TDX driver also uses offset 128 for report data placement, maintaining consistency with the kernel behavior.
2. The report structure returned by TDX Module places the report data at offset 128, so using the same offset keeps our memory layout consistent with the TDX Module's output format.
3. When we implement report refresh (e.g., when extending RTMRs), we can directly call `get_report` on the existing report structure without needing to rewrite the report data, as it's already at the correct offset.

kernel/src/device/tdxguest/mod.rs
taosue: `clone` can be removed using
```rust
    buf.write(
        size_of::<TdxQuoteHdr>(),
        report.reader().to_fallible().limit(TDX_REPORT_LEN),
    )?;
```
Thanks for the reminder!

kernel/src/device/tdxguest/mod.rs
tatetian:
```
struct TdxReport {
    ...
    report_data: [u8; REPORT_DATA_SIZE],
    ...
}

let report_data_paddr = core::mem::offset_of!(TdxReport, report_data);
```

kernel/src/device/tdxguest/mod.rs
```diff
@@ -168,6 +177,42 @@ struct TdxQuoteHdr {
...
+    rtmr1: [u8; 48],
+    rtmr2: [u8; 48],
+    rtmr3: [u8; 48],
+    rtmr4: [u8; 48],
+    servtd_hash: [u8; 48],
+}
+
+impl TdReport {
+    fn report_data_offset() -> usize {
+        offset_of!(ReportMac, report_data) + offset_of!(TdReport, report_mac)
```
tatetian: First, make this function a const so that the value can be determined at the compile time.

Second, reverse the components in the sum expression to make it more natural in logic:

```rust
        offset_of!(TdReport, report_mac) + offset_of!(ReportMac, report_data)
```

kernel/src/device/tdxguest/mod.rs
```diff
@@ -176,36 +221,40 @@ fn handle_get_report(arg: usize) -> Result<i32> {
     let report = tdx_get_report(&user_request.report_data)?;
 
     let tdx_report_vaddr = arg + TDX_REPORTDATA_LEN;
-    user_space.write_bytes(tdx_report_vaddr, &mut VmReader::from(report.as_ref()))?;
+    user_space.write_bytes(tdx_report_vaddr, report.reader().limit(TDX_REPORT_LEN))?;
```
tatetian: I am wondering if `TDX_REPORT_LEN == size_of::<TdReport>()`. If so, use the latter one to draw a connection between the struct definition and the copy length. If not, this could be surprising and need some clarifications.

In fact, my bigger question is that what is the connection between `TdReport` and `TdxReportRequest`?

```rust
const TDX_REPORTDATA_LEN: usize = 64;
const TDX_REPORT_LEN: usize = 1024;

#[derive(Debug, Clone, Copy, Pod)]
#[repr(C)]
pub struct TdxReportRequest {
    report_data: [u8; TDX_REPORTDATA_LEN],
    tdx_report: [u8; TDX_REPORT_LEN],
}
```

Is `TdxReportRequest::tdx_report` actually a `TdReport`? Is `TdxReportRequest::report_data` same as `ReportMac::report_data`? I believe so. If this is indeed the case, then we need to make the common parts of `TdReport` and `TdxReportRequest` use the same struct.

kernel/src/device/tdxguest/mod.rs
```diff
@@ -176,36 +221,40 @@ fn handle_get_report(arg: usize) -> Result<i32> {
...
+/// The rest in `USegment` should be ignored.
+fn tdx_get_report(inblob: &[u8]) -> Result<USegment> {
     if inblob.len() != TDX_REPORTDATA_LEN {
         return_errno_with_message!(Errno::EINVAL, "Invalid inblob length");
     }
 
-    let segment = FrameAllocOptions::new().alloc_segment(2)?;
-    let dma_coherent = DmaCoherent::map(segment.into(), false).unwrap();
-    dma_coherent.write_bytes(0, &inblob).unwrap();
+    let report: USegment = FrameAllocOptions::new().alloc_segment(1)?.into();
```
tatetian: When I read the code for the first time, it got me wonder whether one page is large enough. The following code would make this point more clear.

```rust
let report: USegment = {
    const REPORT_SIZE_IN_PAGES: usize = size_of::<TdReport>().div_ceil(PAGE_SIZE);
    FrameAllocOptions::new().alloc_segment(REPORT_SIZE_IN_PAGES)?.into()
};
```

Note that the above code is as efficient as the current one because `REPORT_NR_PAGES` can be calculated at the runtime.

kernel/src/device/tdxguest/mod.rs
```diff
@@ -176,36 +221,40 @@ fn handle_get_report(arg: usize) -> Result<i32> {
...
+
+    // Use `inblob` as the data associated with the report.
+    let report_data_paddr = {
+        // The `get_report` function requires report data to be 64-byte aligned.
```
tatetian: Now that you have defined `TdReport`, you don't need to justify the paddr. If `TdReport` is defined correctly and the starting address of `TdReport` is well aligned, then the addresses of _all_ its fields must have aligned well.

ostd/src/mm/frame/untyped.rs
```diff
@@ -60,6 +61,14 @@ macro_rules! impl_untyped_frame_meta_for {
             }
         }
         impl $crate::mm::frame::untyped::AnyUFrameMeta for $t {}
+        $crate::impl_untyped_frame_meta_for!(@check_constraints $t);
+    };
+    // Check the size and alignment constraints of the metadata type.
+    (@check_constraints $t:ty) => {
```
tatetian: @taosue Sharp eye! I believe that you have found a soundness hole.

@junyang-zh Could you confirm it?

kernel/src/device/tdxguest/mod.rs
```diff
@@ -168,6 +177,42 @@ struct TdxQuoteHdr {
     out_len: u32,
 }
 
+/// TDX Report structure (`TDREPORT_STRUCT`) as defined in the Intel TDX Module Specification.
+#[repr(C)]
+struct TdReport {
+    report_mac: ReportMac,
+    _reserved: [u8; 256],
+    td_info: TdInfo,
+}
```
tatetian, line 190: Add an assertion to verify the error-prone struct definition is actually correct.

```rust
const EXPECTED_SIZE_OF_TD_REPORT: usize = XXX;
const_assert!(size_of::<TdReport> == EXPECTED_SIZE_OF_TD_REPORT);
```

kernel/src/device/tdxguest/mod.rs
taosue: Yes, they are same. Thanks for the great suggestion!

kernel/src/device/tdxguest/mod.rs
taosue: I just use
```rust
const_assert!(size_of::<TdReport>() == 1024);
```

Because the const definition will trigger a warning:
```
warning: constant `EXPECTED_SIZE_OF_TD_REPORT` is never used
```

ostd/src/mm/frame/meta.rs
```diff
@@ -82,7 +82,7 @@ pub(in crate::mm) struct MetaSlot {
     ///  - the implementation can simply cast a `*const MetaSlot`
     ///    to a `*const AnyFrameMeta` for manipulation;
     ///  - if the metadata need special alignment, we can provide
-    ///    at most `PAGE_METADATA_ALIGN` bytes of alignment;
+    ///    at most `FRAME_METADATA_MAX_ALIGN` bytes of alignment;
```
junyang-zh: We can wrap it in square brackets so that rustdoc and analyzer can navigate to the definition.

```suggestion
   ///    at most [`FRAME_METADATA_MAX_ALIGN`] bytes of alignment;
```

ostd/src/mm/frame/untyped.rs
junyang-zh: Yes, the following code passes the check in the main branch.

```
#[expect(dead_code)]
pub struct TestOversizedMeta([u8; 512]);
crate::impl_untyped_frame_meta_for!(TestOversizedMeta);
```

The bug was introduced when we switched from dynamic checking to static checking. And only `impl_frame_meta_for` was modified.

ostd/src/mm/frame/untyped.rs
```diff
@@ -60,6 +61,14 @@ macro_rules! impl_untyped_frame_meta_for {
...
+        $crate::const_assert!(size_of::<$t>() <= $crate::mm::frame::meta::FRAME_METADATA_MAX_SIZE);
+        $crate::const_assert!(
+            $crate::mm::frame::meta::FRAME_METADATA_MAX_ALIGN % align_of::<$t>() == 0
+        );
```
junyang-zh: I suggest reusing the code from `impl_frame_meta_for` and separating the checking logic from `impl_frame_meta_for` into a new macro `check_frame_meta_layout` to improve conciseness.

junyang-zh (approved):

ostd/src/mm/frame/untyped.rs
lrh2000: > The bug was introduced when we switched from dynamic checking to static checking. And only `impl_frame_meta_for` was modified.

Yes, you can define it, but it should fail here when you try to use it:
https://github.com/asterinas/asterinas/blob/9c8a8f8df972d171e655093437c071ac93a72cf3/ostd/src/mm/frame/meta.rs#L360-L363

So this is not a soundness bug, though having `check_frame_meta_layout` may not be a bad idea (for clarity?).

taosue: Yes, the second commit is mainly for compile-time checks and alignment with `impl_frame_meta_for`. Maybe we can remove the runtime checks in the future?

lrh2000: > Yes, the second commit is mainly for compile-time checks and alignment with `impl_frame_meta_for`. Maybe we can remove the runtime checks in the future?

`const {}` is evaluated at compile time. It will fail to compile if an oversized meta is used. There is no runtime check.

taosue: OK, so the previous behavior would only trigger a compilation error when there was a call such as `alloc_segment_with(nframes, |_| TestOversizedMeta)`, while the change I made is to bring it forward to the declaration, i.e., `impl_untyped_frame_meta_for!(TestOversizedMeta)`.

lrh2000: Yes, I'm not against it as the current behavior might be clearer. I just want to clarify it.

PR #2494, opened 2025-10-08
Ensure the idle tasks do not sleep or block

tatetian:
The motivation of this PR is well explained in the comments below.

```rust
//-----------------------------------------------------------------------------
// Per-CPU idle threads
//-----------------------------------------------------------------------------

// Note: Keep the code in the idle loop to the bare minimum.
//
// We do not want the idle loop to
// rely on the APIs of other kernel subsystems for two reasons.
// First, the idle task must never sleep or block.
// This property is relied upon by the scheduler.
// Second, the idle task is spawned before the kernel is fully initialized.
// So other subsystems may not be ready, yet.
//
// In addition,
// doing more work in the idle task may have negative impact on
// the latency to switching from the idle task to a useful, runnable one.
```

This PR refactors the kernel initialization logic to ensure that the "main" function of the idle task contains only the bare minimum amount of code.

PR #2568, opened 2025-11-08
Refix idle tasks

wyt8:
Relevant details are in #2494.

### DIFF
--- a/kernel/src/init.rs
+++ b/kernel/src/init.rs
@@ -174,7 +174,7 @@ fn print_banner() {
     println!("{}", logo_ascii_art::get_gradient_color_version());
 }
 
-pub(super) fn init_in_first_process(ctx: &Context) {
+pub(crate) fn on_first_process_startup(ctx: &Context) {
     component::init_all(InitStage::Process, component::parse_metadata!()).unwrap();
     crate::device::init_in_first_process(ctx).unwrap();
     crate::fs::init_in_first_process(ctx);
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -60,10 +60,6 @@ mod util;
 mod vdso;
 mod vm;
 
-use crate::{
-    init::init_in_first_process, prelude::*, process::Process, thread::kernel_thread::ThreadOptions,
-};
-
 #[ostd::main]
 #[controlled]
 fn main() {
--- a/kernel/src/net/socket/util/datagram_common.rs
+++ b/kernel/src/net/socket/util/datagram_common.rs
@@ -5,10 +5,10 @@ use ostd::sync::RwMutex;
 use super::SendRecvFlags;
 use crate::{
     events::IoEvents,
+    prelude::*,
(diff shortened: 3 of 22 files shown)


kernel/src/lib.rs
```diff
@@ -77,164 +60,12 @@ mod util;
 mod vdso;
 mod vm;
 
+use crate::{
+    init::init_in_first_process, prelude::*, process::Process, thread::kernel_thread::ThreadOptions,
```
tatetian: Removes these imports from `lib.rs` and fix compiler errors by adding a third commit.

Currently, some modules (accidentally) use these imports in `lib.rs`, but these imports are only meant to be used by the code in `lib.rs` itself. For example, a module should explicitly import from `thread::kernel_thread::ThreadOptions` rather from `crate::ThreadOptions`.

kernel/src/init.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[cfg(any(target_arch = "x86_64", target_arch = "riscv64"))]
+    crate::vdso::init_in_first_kthread();
+}
+
+fn print_banner() {
+    println!("");
+    println!("{}", logo_ascii_art::get_gradient_color_version());
+}
+
+pub(super) fn init_in_first_process(ctx: &Context) {
```
tatetian: Rename this method to `on_first_process_startup` to convey the intention that this function is a callback.

The visibility needs to be changed to `pub(crate)` as `lib.rs` won't import this symbol.

kernel/src/lib.rs
wyt8: Should this import `use crate::prelude::*;` be kept?

tatetian: No.

kernel/src/time/mod.rs
```diff
@@ -63,7 +63,7 @@ impl From<timeval_t> for timespec_t {
 }
 
 impl TryFrom<timespec_t> for Duration {
-    type Error = crate::Error;
+    type Error = crate::prelude::Error;
```
tatetian: Everything exported by the `prelude` module is supposed to be widely used. So widely that each and every module should simply do `use crate::prelude::*` so that these common types are available by default. For example, this `time` module already does `use crate::prelude::*`. So we can use the `Error` type directly.

So use `Xxx` directly for every type `Xxx` implicitly imported from `use crate::prelude::*` (unless using `Xxx` directly causes ambiguity).

Other locations that refer to `crate::prelude::Xxx` explicitly should also be simplified to `Xxx`.

kernel/src/net/socket/util/datagram_common.rs
```diff
@@ -5,10 +5,10 @@ use ostd::sync::RwMutex;
 use super::SendRecvFlags;
 use crate::{
     events::IoEvents,
+    prelude::{Errno, Error, Result},
```
tatetian: Change this line to `prelude::*`.

PR #2570, opened 2025-11-11
Extend Cache Policies for the x86 Architecture

Tangzh33:
This PR aims to improve the display efficiency of the EFI framebuffer. Previously, Asterinas supported only 4 cache policies: Writeback, Writethrough, Uncacheable minus, and Uncacheable. Built on these basic cache policies, the GUI has catastrophic display efficiency: it updates the screen line-by-line when a large area is refreshed.

Therefore, this PR supports advanced cache policies like Write Combining and Write Protected. Write Combining is suitable for the framebuffer. Here is some reference from the Intel x86 architecture manual.

> Write Combining (WC) — System memory locations are not cached (as with uncacheable memory) and coherency is not enforced by the processor’s bus coherency protocol. Speculative reads are allowed. Writes may be delayed and combined in the write combining buffer (WC buffer) to reduce memory accesses.
> 

> This type of cache-control is appropriate for video frame buffers, where the order of writes is unimportant as long as the writes update memory so they can be seen on the graphics display.
> 

Additionally, most mainstream Intel x86 CPUs support the feature, so I think it is appropriate to include the advanced cache policy in the main branch rather than keep it as an experimental implementation.

> This memory type is available in the Pentium Pro and Pentium II processors by programming the MTRRs; or in processor families starting from the Pentium III processors by programming the MTRRs or by selecting it through the PAT.
> 

x86PageTableLayout

The implementation has already been tested on the experimental branch with framebuffer GUI support. And the display efficiency is much better. However, I have two open questions and do not have specific answers yet:

1. How should we incorporate the customized cache policy in our device memory framework? It is reasonable that our current device memory is populated when `mmap` is used, since device memory does not need to support on-demand paging. However, the cache policy for mapping `IoMem` is not customizable; by default, it is Uncachable (`VmMapping::populate_device`). We should support `IoMem` cache policy customization by providing more context for the device to use when customizing, and for the `mmap` logic to use.
2. How should we test the new cache policy with CI? The current implementation has no compilation errors and has proven effective on Xfce. But efficiency can only be validated by the eye. Can we incorporate a formal test to validate?

With questions to be decided, this PR is temporarily a draft PR, and will be ready when the questions are solved :-)

cc @junyang-zh @lrh2000

### DIFF
--- a/kernel/comps/framebuffer/src/framebuffer.rs
+++ b/kernel/comps/framebuffer/src/framebuffer.rs
@@ -5,7 +5,7 @@ use alloc::{sync::Arc, vec::Vec};
 use ostd::{
     boot::boot_info,
     io::IoMem,
-    mm::{HasSize, VmIo},
+    mm::{CachePolicy, HasSize, VmIo},
     sync::Mutex,
     Error, Result,
 };
@@ -96,7 +96,14 @@ pub(crate) fn init() {
         let fb_size = framebuffer_arg.height.checked_mul(line_size).unwrap();
 
         let fb_base = framebuffer_arg.address;
-        let io_mem = IoMem::acquire(fb_base..fb_base.checked_add(fb_size).unwrap()).unwrap();
+        // Use write-combining for framebuffer to enable faster write operations.
+        // Write-combining allows the CPU to combine multiple writes into fewer bus transactions,
+        // which is ideal for framebuffer access patterns (sequential writes).
+        let io_mem = IoMem::acquire_with_cache_policy(
+            fb_base..fb_base.checked_add(fb_size).unwrap(),
+            CachePolicy::WriteCombining,
+        )
+        .unwrap();
 
         let default_cmap = FbCmap {
             entries: Vec::new(),
--- a/kernel/comps/pci/src/cfg_space.rs
+++ b/kernel/comps/pci/src/cfg_space.rs
@@ -229,6 +229,8 @@ impl MemoryBar {
     /// Grants I/O memory access
     pub fn io_mem(&self) -> &IoMem {
         self.io_memory.call_once(|| {
+            // Use the `Uncacheable` cache policy for PCI device BARs by default.
+            // Device-specific drivers may remap with different cache policies if needed.
             IoMem::acquire((self.base as usize)..((self.base + self.size) as usize)).unwrap()
         })
     }
--- a/ostd/src/arch/riscv/irq/chip/plic.rs
+++ b/ostd/src/arch/riscv/irq/chip/plic.rs
@@ -319,7 +319,7 @@ impl Plic {
 
                 Self {
                     phandle,
-                    io_mem: io_mem_builder.reserve(range),
+                    io_mem: io_mem_builder.reserve(range, crate::mm::CachePolicy::Uncacheable),
                     hart_to_target_mapping,
                     interrupt_number_mappings: (0..num_interrupt_sources)
                         .map(|_| None)
--- a/ostd/src/arch/x86/irq/chip/ioapic.rs
+++ b/ostd/src/arch/x86/irq/chip/ioapic.rs
@@ -178,7 +178,10 @@ impl IoApicAccess {
     ///
     /// The caller must ensure that the base address is a valid I/O APIC base address.
     pub(self) unsafe fn new(base_address: usize, io_mem_builder: &IoMemAllocatorBuilder) -> Self {
-        let io_mem = io_mem_builder.reserve(base_address..(base_address + Self::MMIO_SIZE));
+        let io_mem = io_mem_builder.reserve(
+            base_address..(base_address + Self::MMIO_SIZE),
+            crate::mm::CachePolicy::Uncacheable,
+        );
 
         if_tdx_enabled!({
             assert_eq!(
--- a/ostd/src/arch/x86/kernel/apic/mod.rs
+++ b/ostd/src/arch/x86/kernel/apic/mod.rs
@@ -359,7 +359,10 @@ pub fn init(io_mem_builder: &IoMemAllocatorBuilder) -> Result<(), ApicInitError>
         log::info!("xAPIC found!");
         // SAFETY: xAPIC is present.
         let base_address = unsafe { xapic::read_xapic_base_address() };
-        let io_mem = io_mem_builder.reserve(base_address..(base_address + xapic::XAPIC_MMIO_SIZE));
+        let io_mem = io_mem_builder.reserve(
+            base_address..(base_address + xapic::XAPIC_MMIO_SIZE),
+            crate::mm::CachePolicy::Uncacheable,
+        );
         APIC_TYPE.call_once(|| ApicType::XApic(io_mem));
         Ok(())
     } else {
--- a/ostd/src/io/io_mem/allocator.rs
+++ b/ostd/src/io/io_mem/allocator.rs
@@ -179,7 +179,10 @@ mod test {
     use alloc::vec;
 
     use super::{IoMemAllocator, IoMemAllocatorBuilder};
-    use crate::{mm::PAGE_SIZE, prelude::ktest};
+    use crate::{
+        mm::{CachePolicy, PAGE_SIZE},
+        prelude::ktest,
+    };
 
     #[expect(clippy::reversed_empty_ranges)]
     #[expect(clippy::single_range_in_vec_init)]
@@ -189,18 +192,36 @@ mod test {
         let allocator =
             unsafe { IoMemAllocator::new(IoMemAllocatorBuilder::new(range).allocators) };
 
-        assert!(allocator.acquire(0..0).is_none());
-        assert!(allocator.acquire(usize::MAX..0).is_none());
+        assert!(allocator.acquire(0..0, CachePolicy::Uncacheable).is_none());
+        assert!(allocator
+            .acquire(usize::MAX..0, CachePolicy::Uncacheable)
+            .is_none());
 
-        assert!(allocator.acquire(0x4000_0000..0x4000_0000).is_none());
-        assert!(allocator.acquire(0x4000_1000..0x4000_1000).is_none());
-        assert!(allocator.acquire(0x41ff_0000..0x41ff_0000).is_none());
-        assert!(allocator.acquire(0x4200_0000..0x4200_0000).is_none());
+        assert!(allocator
+            .acquire(0x4000_0000..0x4000_0000, CachePolicy::Uncacheable)
+            .is_none());
+        assert!(allocator
+            .acquire(0x4000_1000..0x4000_1000, CachePolicy::Uncacheable)
+            .is_none());
+        assert!(allocator
+            .acquire(0x41ff_0000..0x41ff_0000, CachePolicy::Uncacheable)
+            .is_none());
+        assert!(allocator
+            .acquire(0x4200_0000..0x4200_0000, CachePolicy::Uncacheable)
+            .is_none());
 
-        assert!(allocator.acquire(0x4000_1000..0x4000_0000).is_none());
-        assert!(allocator.acquire(0x4000_2000..0x4000_1000).is_none());
-        assert!(allocator.acquire(0x41ff_f000..0x41ff_e000).is_none());
-        assert!(allocator.acquire(0x4200_0000..0x41ff_f000).is_none());
+        assert!(allocator
+            .acquire(0x4000_1000..0x4000_0000, CachePolicy::Uncacheable)
+            .is_none());
+        assert!(allocator
+            .acquire(0x4000_2000..0x4000_1000, CachePolicy::Uncacheable)
+            .is_none());
+        assert!(allocator
+            .acquire(0x41ff_f000..0x41ff_e000, CachePolicy::Uncacheable)
+            .is_none());
+        assert!(allocator
+            .acquire(0x4200_0000..0x41ff_f000, CachePolicy::Uncacheable)
+            .is_none());
     }
 
     #[ktest]
@@ -216,24 +237,42 @@ mod test {
             unsafe { IoMemAllocator::new(IoMemAllocatorBuilder::new(range).allocators) };
 
         assert!(allocator
-            .acquire((io_mem_region_a.start - 1)..io_mem_region_a.start)
+            .acquire(
+                (io_mem_region_a.start - 1)..io_mem_region_a.start,
+                CachePolicy::Uncacheable
+            )
             .is_none());
         assert!(allocator
-            .acquire(io_mem_region_a.start..(io_mem_region_a.start + 1))
+            .acquire(
+                io_mem_region_a.start..(io_mem_region_a.start + 1),
+                CachePolicy::Uncacheable
+            )
             .is_some());
 
         assert!(allocator
-            .acquire((io_mem_region_a.end + 1)..(io_mem_region_b.start - 1))
+            .acquire(
+                (io_mem_region_a.end + 1)..(io_mem_region_b.start - 1),
+                CachePolicy::Uncacheable
+            )
             .is_none());
         assert!(allocator
-            .acquire((io_mem_region_a.end - 1)..(io_mem_region_b.start + 1))
+            .acquire(
+                (io_mem_region_a.end - 1)..(io_mem_region_b.start + 1),
+                CachePolicy::Uncacheable
+            )
             .is_none());
 
         assert!(allocator
-            .acquire((io_mem_region_a.end - 1)..io_mem_region_a.end)
+            .acquire(
+                (io_mem_region_a.end - 1)..io_mem_region_a.end,
+                CachePolicy::Uncacheable
+            )
             .is_some());
         assert!(allocator
-            .acquire(io_mem_region_a.end..(io_mem_region_a.end + 1))
+            .acquire(
+                io_mem_region_a.end..(io_mem_region_a.end + 1),
+                CachePolicy::Uncacheable
+            )
             .is_none());
     }
 }


ostd/src/arch/x86/mm/mod.rs
```diff
@@ -69,6 +68,12 @@ bitflags::bitflags! {
         /// Ignored by the hardware. Free to use.
         const HIGH_IGN2 =       1 << 53;
 
+
+        /// Valid page marker. It indicates a valid leaf page entry at level 1.
+        /// For levels 2 and 3, the `HUGE` bit indicates a valid leaf entry. For
+        /// levels above 3, `PRESENT` is used for "valid".
+        const VALID_PAGE =      1 << 55;
```
junyang-zh: The comments could have mentioned that it is ignored by the hardware. We use it for validity tracking.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -174,13 +182,21 @@ impl PodOnce for PageTableEntry {}
...
+        self.0 & PageTableFlags::PRESENT.bits() != 0
+            || self.0 & PageTableFlags::HUGE.bits() != 0
+            || self.0 & PageTableFlags::VALID_PAGE.bits() != 0
     }
 
-    fn new_page(paddr: Paddr, _level: PagingLevel, prop: PageProperty) -> Self {
-        let flags = PageTableFlags::HUGE.bits();
+    fn new_page(paddr: Paddr, level: PagingLevel, prop: PageProperty) -> Self {
+        // For x86_64:
+        // - Level 1 (4KB pages): set VALID_PAGE bit (bit 55), bit 7 can be PAT
```
junyang-zh:
```suggestion
        // - Level 1 (4KB pages): set VALID_PAGE bit (bit 55), bit 7 is PAT
```

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -214,13 +230,35 @@ impl PageTableEntryTrait for PageTableEntry {
...
+        // PAT index encoding programmed by `configure_pat`:
+        // PAT PCD PWT -> Memory Type
+        //  0   0   0  -> WB (Writeback)
+        //  0   0   1  -> WT (Writethrough)
+        //  0   1   0  -> UC- (Uncacheable minus)
+        //  0   1   1  -> UC (Uncacheable)
+        //  1   0   0  -> WC (Write Combining)
+        //  1   0   1  -> WP (Write Protected)
+        //  1   1   0  -> UC (Uncacheable)
+        //  1   1   1  -> UC (Uncacheable)
```
junyang-zh: This looks ugly and prone to error if we will change the table later. And, the programmer's guide suggests clearly that you should combine the `PAT PCD PWT` bits as an 3-bit integer index to the array. So we should do so.

Firstly, map cache policies to PAT entry values instead of using your `const PAT_ENTRIES: [u64; 8]` array.

```rust
impl CachePolicy {
    /// Reference: Table 11-10 in Intel SDM Vol. 3A.
    fn to_ia32_pat_memory_type_entry(&self) -> u8 { match self { /* ... */ } }
}
```

Then, define our software-defined policy mappings.

```
const IA32_PAT_MAPPINGS: [CachePolicy; 8] = [/*...*/];
```

To query the policy, this array could be indexed by the 3-bit integer here.

Finally, when doing `configure_pat`, map each values in this array.

ostd/src/arch/x86/mod.rs
```diff
@@ -186,6 +188,14 @@ pub(crate) fn enable_cpu_features() {
 
     cpu::context::enable_essential_features();
 
+    let features = CpuId::new()
+        .get_feature_info()
+        .expect("failed to query CPUID feature information");
+    debug_assert!(features.has_pat(), "CPU must support PAT");
+    if features.has_pat() {
+        mm::configure_pat();
+    }
```
junyang-zh: We can proceed if there's no PAT instead of panicking. Like, maybe it should work on processors older than Pentium-III or some weird home-brew licensed x86 chips.

Read SDM 11.12.5 to see how the rest of the code should handle this.

> To allow compatibility for code written to run on earlier IA-32 processor that do not support the PAT, the PAT mechanism has been designed to allow backward compatibility to earlier processors. This compatibility is provided through the ordering of the PAT, PCD, and PWT bits in the 3-bit PAT entry index. For processors that do not implement the PAT, the PAT index bit (bit 7 in the page-table entries and bit 12 in the page-directory entries) is reserved and set to 0. With the PAT bit reserved, only the first four entries of the PAT can be selected with the PCD and PWT bits. At power-up or reset (see Table 11-12), these first four entries are encoded to select the same memory types as the PCD and PWT bits would normally select directly in an IA-32 processor that does not implement the PAT. So, if encodings of the first four entries in the PAT are left unchanged following a power-up or reset, code written to run on earlier IA-32 processors that do not implement the PAT will run correctly on IA-32 processors that do implement the PAT.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -302,3 +379,33 @@ impl fmt::Debug for PageTableEntry {
...
+    //   0: Write-back (WB)
+    //   1: Write-through (WT)
+    //   2: Uncacheable minus (UC-)
+    //   3: Uncacheable (UC)
+    //   4: Write-combining (WC)
+    //   5: Write-protected (WP)
+    //   6: Uncacheable (UC)
+    //   7: Uncacheable (UC)
+
+    unsafe {
```
junyang-zh, line 313: Always write safety reasoning. Just explain what it means to read/write these MSRs.

ostd/src/arch/x86/mod.rs
lrh2000: > Pentium-III 

It does not even have x86-64 support. We sometimes assume a feature exists because we're running in 64-bit mode, e.g., all 64-bit CPUs should have FPU.
https://github.com/asterinas/asterinas/blob/0615e6f6424ead8e12d73670428149fa7270571f/ostd/src/arch/x86/cpu/context/mod.rs#L662-L663

If we want to check it, we should use `arch::cpu::extensions` to check it. We should avoid using the external `x86::cpuid::CpuId` code in all of our code.

Given that I cannot find a 64-bit CPU that does not support PAT, I think it's reasonable to leave a comment and assume that it exists (following the FPU case: should we waste an extension bit for an extension that must exist?). But I'm not sure how to formally determine whether a feature must exist on all 64-bit CPUs...

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -302,3 +379,33 @@ impl fmt::Debug for PageTableEntry {
...
+        let current = rdmsr(IA32_PAT);
```
lrh2000: No need to read and compare. Just use `wrmsr` is okay?

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -302,3 +379,33 @@ impl fmt::Debug for PageTableEntry {
...
+    | (PAT_ENTRIES[6] << 48)
+    | (PAT_ENTRIES[7] << 56);
+
+/// Programs the PAT MSR so that write-combining mappings use the correct memory type.
+pub(crate) fn configure_pat() {
+    // Desired PAT entries (index 0..7):
+    //   0: Write-back (WB)
+    //   1: Write-through (WT)
+    //   2: Uncacheable minus (UC-)
+    //   3: Uncacheable (UC)
```
lrh2000: Actually a simpler alternative is to use replace `Uncacheable minus` with `Write Combine`.

Yes, it may not a long-term solution, but given that we do not have `UncacheableMinus` variant in `CachePolicy`, it's just simpler. So what's the reason for not choosing a simpler solution?

---

If we really want to use the complex solution (i.e., the solution proposed in this PR), I think it's better to correct the method signatures instead of playing some even more complex bit tricks, e.g.,
```patch
diff --git a/ostd/src/mm/page_table/node/entry.rs b/ostd/src/mm/page_table/node/entry.rs
index 0d0aef6f4..b9af1513f 100644
--- a/ostd/src/mm/page_table/node/entry.rs
+++ b/ostd/src/mm/page_table/node/entry.rs
@@ -72,7 +72,8 @@ impl<'a, 'rcu, C: PageTableConfig> Entry<'a, 'rcu, C> {
             return;
         }
 
-        self.pte.set_prop(new_prop);
+        let level = self.node.meta().level;
+        self.pte.set_prop(new_prop, level);
 
         // SAFETY:
         //  1. The index is within the bounds.
```

Meanwhile, `is_last` already accepts a `level` parameter. The current implementation is very odd (see the comment below).

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -270,21 +308,60 @@ impl PageTableEntryTrait for PageTableEntry {
...
         }
         self.0 = self.0 & !Self::PROP_MASK | flags;
     }
 
     fn is_last(&self, _level: PagingLevel) -> bool {
-        self.0 & PageTableFlags::HUGE.bits() != 0
+        // A PTE is "last" (leaf mapping) if:
+        // 1. It has HUGE bit set (huge page at level 2/3), OR
+        // 2. It has VALID_PAGE bit set (4KB page at level 1)
+        self.0 & PageTableFlags::HUGE.bits() != 0 || self.0 & PageTableFlags::VALID_PAGE.bits() != 0
```
lrh2000: Why do we want to do this instead of checking `level == 1 || self.0.contains(PageTableFlags::HUGE)`?

https://github.com/asterinas/asterinas/blob/0615e6f6424ead8e12d73670428149fa7270571f/ostd/src/arch/loongarch/mm/mod.rs#L289-L291

This can simplify the related logics a lot.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -270,21 +308,60 @@ impl PageTableEntryTrait for PageTableEntry {
...
+                    panic!("WriteCombining only supported for 4KB pages, not huge pages");
+                }
+            }
+            CachePolicy::WriteProtected => {
+                // PAT=1, PCD=0, PWT=1
+                // Only valid for 4KB pages
+                if is_4kb_page {
+                    flags |= PageTableFlags::HUGE.bits() | PageTableFlags::WRITE_THROUGH.bits();
+                } else {
+                    panic!("WriteProtected only supported for 4KB pages, not huge pages");
```
lrh2000:
```suggestion
                    assert!(is_4kb_page, "`CachePolicy::WriteProtected` is not supported for huge pages");
```

lrh2000:
> How should we incorporate the customized cache policy in our device memory framework?

We can easily extend `IoMem::acquire` to accept a `CachePolicy` parameter. The framebuffer will require `WriteCombine` I/O memory. So why we don't do this in this PR?

> I see no problem in the approach of using another unused bit instead of PAT to test validity.

The problem may be that it will cause confusion. So I think it needs to be justified why it's better than the other alternatives (see some random comments above).

tatetian:
> > How should we incorporate the customized cache policy in our device memory framework?
> 
> We can easily extend `IoMem::acquire` to accept a `CachePolicy` parameter. The framebuffer will require `WriteCombine` I/O memory. So why we don't do this in this PR?

I talked with @Tangzh33 this afternoon. This is the direction that we agreed upon. The desired cache policy will be an attribute of an `IoMem`. This way, whether `IoMem` is mapped into the kernel space or later the user address space, the memory management code can the cache policy appropriately for the memory mapping backed by this `IoMem`.

junyang-zh:
> > I see no problem in the approach of using another unused bit instead of PAT to test validity.
>
> The problem may be that it will cause confusion. So I think it needs to be justified why it's better than the other alternatives (see some random comments above).

The thing is we still need a bit to mark valid a page with no protection permissions (not readable, so in x86 no `PRESENT` bit). I agree that `is_last` needn't consult this bit.

junyang-zh:
OK now I understand what @lrh2000 proposes after a (private) discussion with @Tangzh33 . I agree and I think we could enable the PAT feature, but encode the attribute table with 4  policies, leaving 4 to 7 the same as 0 to 3 so that the PAT bit can have arbitrary value. Then, we could use PAT as the valid bit as we originally did.

Tangzh33:
@tatetian @junyang-zh @lrh2000 Thanks for your insightful discussions, and I have refactored the code accordingly.

In this iteration, I adopted @lrh2000's suggestion and replaced the `Uncacheable Minus` with `Write Combining`. The changes look clean and straightforward, as expected. Additionally, I enhanced `IoMem` to support customizable cache policies and updated the related code accordingly. Since the main open questions have been thoroughly discussed and reached consensus, this PR is no longer a draft and is ready for review. Please leave your comments at your convenience~

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -160,6 +181,10 @@ impl PageTableEntry {
         }
     }
     const PROP_MASK: usize = !Self::PHYS_ADDR_MASK & !PageTableFlags::HUGE.bits();
+
+    fn is_huge(&self) -> bool {
+        self.0 & PageTableFlags::HUGE.bits() != 0
+    }
```
junyang-zh: No I don't think it can exist because it is way too ambiguous. We encode entries in different levels with the same struct, and in different levels the huge bit doesn't always mean the huge bit.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -214,13 +239,16 @@ impl PageTableEntryTrait for PageTableEntry {
...
-        } else if self.0 & PageTableFlags::WRITE_THROUGH.bits() != 0 {
-            CachePolicy::Writethrough
-        } else {
-            CachePolicy::Writeback
-        };
+
+        // Determine cache policy from PAT, PCD, PWT bits.
+        // For leaf pages (level 1 4KB pages and level 2/3 huge pages), bit 7 (HUGE) is always 1,
+        // so they all use PAT indices 4-7, which mirror indices 0-3 in our configuration.
+        let pat_bit = self.is_huge();
```
junyang-zh: This is simply wrong. On level-2/level-3 huge page mappings, the PAT bit is the 12th bit, not the huge bit. You can add a `PAT_FOR_HUGR` bit, and use the `level` parameter to tell which to use (`PAT_FOR_HUGE` or `HUGE`).

ostd/src/mm/page_prop.rs
```diff
@@ -96,6 +96,22 @@ pub enum CachePolicy {
     Writeback,
 }
 
+#[cfg(target_arch = "x86_64")]
+impl CachePolicy {
+    /// Returns the IA32_PAT memory type encoding for this cache policy.
```
junyang-zh: Be consistent on formatting. Since you wrap `IA32_PAT` with backticks erlier, do it here as well.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -137,7 +154,11 @@ pub(crate) unsafe fn activate_page_table(root_paddr: Paddr, root_pt_cache: Cache
         CachePolicy::Writeback => x86_64::registers::control::Cr3Flags::empty(),
         CachePolicy::Writethrough => x86_64::registers::control::Cr3Flags::PAGE_LEVEL_WRITETHROUGH,
         CachePolicy::Uncacheable => x86_64::registers::control::Cr3Flags::PAGE_LEVEL_CACHE_DISABLE,
-        _ => panic!("unsupported cache policy for the root page table"),
+        // WriteCombining and WriteProtected are not supported for root page table (CR3)
```
junyang-zh: Use plain English in comments. Wrap them with backticks or do

```suggestion
        // Write-combining and write-protected are not supported for root page table (CR3)
```

instead.

kernel/comps/framebuffer/src/framebuffer.rs
```diff
@@ -68,7 +68,14 @@ pub(crate) fn init() {
         let fb_size = framebuffer_arg.height.checked_mul(line_size).unwrap();
 
         let fb_base = framebuffer_arg.address;
-        let io_mem = IoMem::acquire(fb_base..fb_base.checked_add(fb_size).unwrap()).unwrap();
+        // Use WriteCombining for framebuffer to enable faster write operations.
```
junyang-zh: Use plain English here.

ostd/src/io/io_mem/mod.rs
```diff
@@ -172,12 +180,12 @@ impl IoMem<Sensitive> {
 }
 
 impl IoMem<Insensitive> {
-    /// Acquires an `IoMem` instance for the given range.
-    pub fn acquire(range: Range<Paddr>) -> Result<IoMem<Insensitive>> {
+    /// Acquires an `IoMem` instance for the given range with the specified cache policy.
+    pub fn acquire(range: Range<Paddr>, cache_policy: CachePolicy) -> Result<IoMem<Insensitive>> {
```
tatetian: `IoMem` is mainly introduced to represent MMIO, which normally use `Uncacheable`. Only graphics display buffers use `WriteCombining`. So I would say that in 90% situations, the users of `IoMem` choose `Uncacheable`. So a more user-friendly design is to have two separate APIs:

```rust
impl IoMem {
    pub fn acquire(range: Range<Paddr>) -> Result<IoMem<Insensitive>> {
        Self::acquire_with_policy(range, CachePolicy::Uncachable)
    }

    pub fn acquire_with_policy(range: Range<Paddr>, cache_policy: CachePolicy) -> Result<IoMem<Insensitive>> { ... }
}
```

tatetian (commented):
Great work. The changes look mostly good.

Tangzh33:
@tatetian @junyang-zh Thank you for your attention! Your comments have been addressed in this iteration. This PR is ready for review again; please let me know if you have any further comments :-)

lrh2000:
> The thing is we still need a bit to mark valid a page with no protection permissions (not readable, so in x86 no `PRESENT` bit).

@junyang-zh So what do you think is the expected behavior if the underlying architecture-specific page table cannot support the required flags?

For example, what would happen if we tried to protect pages to be write-only? If we deny it (or ignore it), why not deny (or ignore) "not readable" as well?

A similar question is what will happen if we specify the cache policy that the underlying page table cannot support (e.g., RISC-V without `SVPBMT`).

Besides, we should not make use of `RWX=000` in RISC-V as well because it's reserved for future use per the spec?

ostd/src/mm/page_prop.rs
```diff
@@ -96,6 +96,22 @@ pub enum CachePolicy {
     Writeback,
 }
 
+#[cfg(target_arch = "x86_64")]
+impl CachePolicy {
+    /// Returns the `IA32_PAT` memory type encoding for this cache policy.
+    ///
+    /// Reference: Intel 64 and IA-32 Architectures SDM Vol. 3A, Table 11-10.
+    pub(crate) const fn to_ia32_pat_memory_type_entry(self) -> u8 {
```
lrh2000: This should be moved to the x86 module? I suspect it should live inside `configure_pat`...
```rust
pub(crate) fn configure_pat() {
    fn cache_policy_to_pat_entry(cache_policy: CachePolicy) -> u8 { ... }
}
```

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -302,3 +330,20 @@ impl fmt::Debug for PageTableEntry {
             .finish()
     }
 }
+
+/// Programs the PAT MSR so that write-combining mappings use the correct memory type.
+pub(crate) fn configure_pat() {
```
lrh2000: This is `pub(super)`.

Given that we have `cpu::context::enable_essential_features();`, can we have `mm::enable_essential_features()` which include both this and
https://github.com/asterinas/asterinas/blob/ee3488cbc586be775612400a3f7ee5611cdc04cb/ostd/src/arch/x86/mod.rs#L190-L195
?

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -302,3 +330,20 @@ impl fmt::Debug for PageTableEntry {
...
+    let mut programmed_pat = 0u64;
+    for (idx, policy) in IA32_PAT_MAPPINGS.iter().copied().enumerate() {
+        programmed_pat |= (policy.to_ia32_pat_memory_type_entry() as u64) << (idx * 8);
+    }
+
+    // SAFETY: Writing `IA32_PAT` merely programs the page attribute table MSR.
+    // This code runs on the bootstrap processor before other CPUs depend on
```
lrh2000: This is not true. The code will also run in APs.

---

To make things simpler, I think most operations in `enable_cpu_features` is safe just because it builds up our global invariant of the CPU state. We'd better figure out a better way to write safety comments.

ostd/src/arch/x86/mod.rs
```diff
@@ -186,6 +186,12 @@ pub(crate) fn enable_cpu_features() {
 
     cpu::context::enable_essential_features();
 
+    // PAT has been available since Pentium III (1999) and is ubiquitous in
+    // modern 64-bit CPUs. Therefore, we assume that all x86-64 CPUs should
+    // have Page Attribute Table (PAT) support. Otherwise, we should check
+    // `x86::cpuid::FeatureInfo::has_pat()` here.
```
lrh2000: Again, please do not mention `x86::cpuid` in our code. We should never use it. Mention the `cpu::extension` module instead. See the FPU example I've presented earlier.

ostd/src/arch/x86/mod.rs
```diff
@@ -186,6 +186,12 @@ pub(crate) fn enable_cpu_features() {
 
     cpu::context::enable_essential_features();
 
+    // PAT has been available since Pentium III (1999) and is ubiquitous in
+    // modern 64-bit CPUs. Therefore, we assume that all x86-64 CPUs should
+    // have Page Attribute Table (PAT) support. Otherwise, we should check
```
lrh2000: We should mention the full name "Page Attribute Table" when the name appears in its first time, i.e., line 189 instead of line 191.

kernel/comps/pci/src/cfg_space.rs
```diff
@@ -229,6 +229,8 @@ impl MemoryBar {
     /// Grants I/O memory access
     pub fn io_mem(&self) -> &IoMem {
         self.io_memory.call_once(|| {
+            // Use Uncacheable for PCI device BARs by default.
```
lrh2000:
```suggestion
            // Use the `Uncacheable` cache policy for PCI device BARs by default.
```

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -275,16 +300,19 @@ impl PageTableEntryTrait for PageTableEntry {
...
+            }
             _ => panic!("unsupported cache policy"),
         }
         self.0 = self.0 & !Self::PROP_MASK | flags;
     }
 
-    fn is_last(&self, _level: PagingLevel) -> bool {
-        self.0 & PageTableFlags::HUGE.bits() != 0
+    fn is_last(&self, level: PagingLevel) -> bool {
+        level == 1 || self.0 & PageTableFlags::HUGE.bits() != 0
```
lrh2000: Let's not change this in this PR. I think we should address this issue, but it is not related to this PR. Also, I don't think we've decided on everything here. For example, see https://github.com/asterinas/asterinas/pull/2570#issuecomment-3530728083. Maybe we should create a separate PR to address these questions, but it should happen after some consensus...

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -20,6 +21,22 @@ use crate::{
 
 mod util;
 
+/// Software-defined mapping from PAT bit combinations to cache policies.
+///
+/// Index encoding: `(PAT << 2) | (PCD << 1) | PWT`.
```
lrh2000: Insert a blank line below this line.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -20,6 +21,22 @@ use crate::{
...
+const IA32_PAT_MAPPINGS: [CachePolicy; 8] = [
+    CachePolicy::Writeback,      // Index 0: PAT=0, PCD=0, PWT=0
+    CachePolicy::Writethrough,   // Index 1: PAT=0, PCD=0, PWT=1
+    CachePolicy::WriteCombining, // Index 2: PAT=0, PCD=1, PWT=0 (replaces UC-)
+    CachePolicy::Uncacheable,    // Index 3: PAT=0, PCD=1, PWT=1
+    CachePolicy::Writeback,      // Index 4: PAT=1, PCD=0, PWT=0 (same as 0)
+    CachePolicy::Writethrough,   // Index 5: PAT=1, PCD=0, PWT=1 (same as 1)
+    CachePolicy::WriteCombining, // Index 6: PAT=1, PCD=1, PWT=0 (same as 2)
+    CachePolicy::Uncacheable,    // Index 7: PAT=1, PCD=1, PWT=1 (same as 3)
+];
```
lrh2000:
```rust
const IA32_PAT_MAPPINGS: [CachePolicy; 8] = [
    CachePolicy::Writeback,      // Index 0: PAT=0, PCD=0, PWT=0
    CachePolicy::Writethrough,   // Index 1: PAT=0, PCD=0, PWT=1
    CachePolicy::WriteCombining, // Index 2: PAT=0, PCD=1, PWT=0 (replaces UC-)
    CachePolicy::Uncacheable,    // Index 3: PAT=0, PCD=1, PWT=1
    CachePolicy::Writeback,      // Index 4: PAT=1, PCD=0, PWT=0 (same as 0)
    CachePolicy::Writethrough,   // Index 5: PAT=1, PCD=0, PWT=1 (same as 1)
    CachePolicy::WriteCombining, // Index 6: PAT=1, PCD=1, PWT=0 (same as 2)
    CachePolicy::Uncacheable,    // Index 7: PAT=1, PCD=1, PWT=1 (same as 3)
];

const fn pat_to_cache_policy(page_flags: PageTableFlags) -> CachePolicy {
    let mut index = 0usize;
    if page_flags.contains(PageTableFlags::NO_CACHE) {
        index |= 2;
    }
    if page_flags.contains(PageTableFlags::WRITE_THROUGH) {
        index |= 1;
    }
    IA32_PAT_MAPPINGS[index]
}

const fn cache_policy_to_pat(cache_policy: CachePolicy) -> PageTableFlags {
    match cache_policy {
        CachePolicy::Writeback => PageTableFlags::empty(),
        CachePolicy::Writethrough => PageTableFlags::WRITE_THROUGH,
        CachePolicy::Uncacheable => PageTableFlags::NO_CACHE.union(PageTableFlags::WRITE_THROUGH),
        CachePolicy::WriteCombining => PageTableFlags::NO_CACHE,
        _ => panic!("unsupported cache policy"),
    }
}

const_assert!(matches!(
    pat_to_cache_policy(cache_policy_to_pat(CachePolicy::Writeback)),
    CachePolicy::Writeback
));
const_assert!(matches!(
    pat_to_cache_policy(cache_policy_to_pat(CachePolicy::Writethrough)),
    CachePolicy::Writethrough
));
const_assert!(matches!(
    pat_to_cache_policy(cache_policy_to_pat(CachePolicy::Uncacheable)),
    CachePolicy::Uncacheable
));
const_assert!(matches!(
    pat_to_cache_policy(cache_policy_to_pat(CachePolicy::WriteCombining)),
    CachePolicy::WriteCombining
));
```

 1. Define methods that perform the translation work, and make the methods live near to `IA32_PAT_MAPPINGS`.
 2. Use constant assertions to ensure the correctness.
 3. These methods appear too early in this file. Try to find a better position. Well, I think creating a new `pat.rs` may be a better idea. So they can live near to `configure_pat`.

junyang-zh:
> > The thing is we still need a bit to mark valid a page with no protection permissions (not readable, so in x86 no `PRESENT` bit).
>
> @junyang-zh So what do you think is the expected behavior if the underlying architecture-specific page table cannot support the required flags?
>
> For example, what would happen if we tried to protect pages to be write-only? If we deny it (or ignore it), why not deny (or ignore) "not readable" as well?
>
> A similar question is what will happen if we specify the cache policy that the underlying page table cannot support (e.g., RISC-V without `SVPBMT`).
>
> Besides, we should not make use of `RWX=000` in RISC-V as well because it's reserved for future use per the spec?

The problem is, OSTD provides an abstraction that everything mentioned above can be done. We either fix it, or continue to do our "silently piggybacking with something else" approach. For example, in x86 fallback to RW for write-only; use another bit to keep track of "not readable" mapping's ownership of the frame; ignore the cache policy for RISC-V without SVPBMT...

I haven't got a clear answer on this matter. I am somewhat inclined to modify the abstraction and let the upper kernel decide what to do before doing protections (e.g. unmap the frame if `RWX=000`). OSTD could give an `UnsupportedPageProperty` error if the kernel doesn't do it well.

I would address it in the upcoming CortenMM refactors. Meanwhile, for flexibility, I think we could add a variant of mapping state, that MMU will think it's unmapped but the software can store something in the page table, e.g., the frame to be mapped. This is orthogonal to the abovementioned unsupported error handling because if the upper kernel still wants to have a valid but `RWX=000` mapping we still want to give an error.

But in this PR let's keep the piggybacking way for now...

junyang-zh:
> For example, in x86 fallback to RW for write-only

Acutally the status quo is, we will set `PRESENT=0 PAT/HUGE=1 WRITABLE=1` for write-only, and the MMU will fault on such writes. The actual fallback is done by the upper kernel...

It's kind of a mess right now.

ostd/src/io/io_mem/mod.rs
lrh2000: I was thinking that making `uncacheable` explicit, like with `acquire_uncachable`, might help (to clarify the caller's intent). Perhaps the name is considered too long? I don't have a strong opinion either way, but I'd like to know if it's been considered.


Another problem is that it's unclear what kind of policy `acquire_with_policy` wants to mean. So I considered:
```rust
impl IoMem {
    pub fn acquire_uncached(range: Range<Paddr>) -> Result<IoMem<Insensitive>> {
        Self::acquire_cached(range, CachePolicy::Uncachable)
    }

    pub fn acquire_cached(range: Range<Paddr>, cache_policy: CachePolicy) -> Result<IoMem<Insensitive>> {
       // ..
    }
}
```

But will `acquire_cached(range, CachePolicy::Uncacheable)` be considered too unusual to exist, i.e., the method name says it will be cached but the policy is uncacheable? So I'm also not sure whether these APIs are better...

junyang-zh: IMHO `pub fn acquire(range: Range<Paddr>, cache: CachePolicy)` is fine.

lrh2000:
> in x86 fallback to RW for write-only; use another bit to keep track of "not readable" mapping's ownership of the frame

So I think that we may want to follow a simpler, more consistent rule. For example, if the architecture page table cannot support a "not readable" mapping and the OSTD user specifies one,
 1. we should ignore the request: This means that "RW=01" (write-only) will become "RW=11" (read-write), and "RW=00" (neither readable nor writable) will become "RW=10" (read-only).
 2. we should reject the request: panicking, reporting errors, e.t.c.

For simpler cases, if we can make it into compile-time checks, I think it's even better; but I'm not sure about the complex cases like the RISC-V extensions:
```rust
struct PageFlags {
    #[cfg(not(target_arch = "x86_64"))] // x86 page tables do not have the "readable" bit.
    const R: u32 = 0x1;
}
```

> MMU will think it's unmapped but the software can store something in the page table [..] This is orthogonal to the abovementioned unsupported error handling

Maybe. This needs to be separate proposal anyway.

> But in this PR let's keep the piggybacking way for now...

Yes. In this PR, I want to avoid modifying the related code at all, such as https://github.com/asterinas/asterinas/pull/2570#discussion_r2525948230. So we don't need to discuss the related stuffs since they're totally existing issues.

ostd/src/io/io_mem/mod.rs
Tangzh33: It seems we haven't reached a consensus yet, and I'm still uncertain about the final decision. From a functional perspective, the current implementation works well. I will refactor the interface once we finalize the decision :-)

Tangzh33:
@lrh2000 @junyang-zh Thanks for your insightful discussions and thorough reviews. I have refined the code accordingly, and this PR is ready for review again. Please let me know if there are any problems.

ostd/src/io/io_mem/mod.rs
tatetian: **A user-friendly API design makes the common cases easy, the rare cases possible.** So I stand with my current position that there should be two constructor methods: one with a short name for the common case where the cache policy is `CachePolicy::Uncacheable`, the other with a longer name for the rare case where the cache policy needs to be given.

> Another problem is that it's unclear what kind of policy `acquire_with_policy` wants to mean.

To enhance clarity (on the price of a longer name), we can rename it to `acquire_with_cache_policy`. Since it is rarely used, the longer name is still acceptable.

So how about this one?

```rust
impl IoMem {
    pub fn acquire(range: Range<Paddr>) -> Result<IoMem<Insensitive>> {
        Self::acquire_with_cache_policy(range, CachePolicy::Uncachable)
    }

    pub fn acquire_with_cache_policy(range: Range<Paddr>, cache_policy: CachePolicy) -> Result<IoMem<Insensitive>> {
       // ..
    }
}
```

lrh2000: I don't have a problem with having two constructors. What I mean is that `uncacheable` seems to be an important property, which is the basis for the correctness of operating the hardware registers. So including it in the function name makes the code easier for the caller to understand, showing that it's the intended method to use without worrying about the implementation. So that's why I feel a bit better when seeing `acquire_uncached`/`acquire_cached` (`acquire_cached` accepts another parameter so it's already a bit awkward to use, or maybe we can choose a longer name for it so it will become even more awkward to use).

Of course, if you feel that it's I/O memory, then it's uncacheable by default. In that case, it's obvious that the default constructor should behave this way. I think it's also okay. I have no problem with it as long as you say it explicitly.

tatetian: > Of course, if you feel that it's I/O memory, then it's uncacheable by default.

Yes, this is my understanding. I think it is sufficient to point this out in the documentation, rather than in the function name.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -303,3 +304,46 @@ impl fmt::Debug for PageTableEntry {
...
+        // Enable non-executable page protection.
+        Efer::update(|efer| {
+            *efer |= EferFlags::NO_EXECUTE_ENABLE;
+        });
+    }
+}
+
+/// Programs the PAT MSR so that write-combining mappings use the
+/// correct memory type.
+fn configure_pat() {
```
lrh2000: 1. Move this to `pat.rs`.
 2. Remove `pub(super)` from `IA32_PAT_MAPPINGS`.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -215,13 +228,10 @@ impl PageTableEntryTrait for PageTableEntry {
...
-            CachePolicy::Uncacheable
-        } else if self.0 & PageTableFlags::WRITE_THROUGH.bits() != 0 {
-            CachePolicy::Writethrough
-        } else {
-            CachePolicy::Writeback
-        };
+
+        // Determine cache policy from PCD, PWT bits.
+        let cache = flags_to_cache_policy(PageTableFlags::from_bits_truncate(self.0));
+
```
lrh2000: Remove blank line 231 and 234 to be consistent to the `set_prop` function.

ostd/src/arch/x86/cpu/extension.rs
```diff
@@ -112,6 +112,7 @@ define_isa_extensions! {
         RDRAND,       Ecx(30), "The processor supports RDRAND instruction.";
 
         XAPIC,        Edx( 9), "APIC On-Chip.";
+        PAT,          Edx(16), "Supports the Page Attribute Table (PAT).";
```
lrh2000: There is no need to define this, nor do I intend to.

At least, we don't define "FPU" because it's available on all x86-64 CPUs.

ostd/src/io/io_mem/mod.rs
```diff
@@ -177,7 +185,19 @@ impl IoMem<Insensitive> {
         allocator::IO_MEM_ALLOCATOR
```
lrh2000: This should call `acquire_with_policy`.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -303,3 +304,46 @@ impl fmt::Debug for PageTableEntry {
...
+    let mut programmed_pat = 0u64;
+    for (idx, policy) in IA32_PAT_MAPPINGS.iter().copied().enumerate() {
+        programmed_pat |= (cache_policy_to_pat_entry(policy) as u64) << (idx * 8);
+    }
+
+    // SAFETY: Writing `IA32_PAT` only programs the PAT MSR of the current CPU.
+    // Updating PAT merely redefines how hardware interprets future cache
+    // policy encodings. No system components rely on the contents of the PAT
+    // being modified before this point.
```
lrh2000:
```rust
    // policy encodings. The programmed value is the global invariant, which
    // is set up before the kernel page table is activated.
```

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -303,3 +304,46 @@ impl fmt::Debug for PageTableEntry {
             .finish()
     }
 }
+
+/// Enables memory-management essential features for the x86 MMU.
+pub(super) fn enable_essential_features() {
+    // PAT has been available since Pentium III (1999) and is ubiquitous in
+    // modern 64-bit CPUs. Therefore, we assume that all x86-64 CPUs should
+    // have Page Attribute Table (PAT) support. Otherwise, we should check
```
lrh2000: https://github.com/asterinas/asterinas/pull/2570#discussion_r2525915985 means:
```rust
    // Page Attribute Table (PAT) has been available since Pentium III (1999)
    // and is ubiquitous in modern 64-bit CPUs. Therefore, we assume that all
    // x86-64 CPUs should have PAT support. Otherwise, we should check
```

ostd/src/arch/x86/mm/pat.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+    if bits & PageTableFlags::WRITE_THROUGH.bits() != 0 {
+        index |= 1;
+    }
+    IA32_PAT_MAPPINGS[index]
+}
+
+pub(super) const fn cache_policy_to_flags(cache_policy: CachePolicy) -> PageTableFlags {
+    let bits = match cache_policy {
+        CachePolicy::Writeback => 0,
```
lrh2000, line 40: Keep the reference in the previous version.

ostd/src/io/io_mem/mod.rs
lrh2000: Okay. I still think `acquire_uncached` could be clearer for the aforementioned reason, but I'm also fine with the current version.

lrh2000 (commented):
Thanks! Just some minor comments.

ostd/src/arch/x86/mm/pat.rs
Tangzh33: I probably miss some context and do not understand your meaning. Can you be more specific about what the "reference" is here?

tatetian:
@lrh2000 You can merge this PR once you are satisfied ❤️

ostd/src/arch/x86/mm/pat.rs
lrh2000: Oh I'm sorry. I posted the comment on the wrong line, so it's confusing.

I mean the comment should be added back in line 71:
```rust
    // Reference: Intel(R) 64 and IA-32 Architectures Software Developer's Manual, Table 12-10,
    // "Memory Types That Can Be Encoded With PAT".
    fn cache_policy_to_pat_entry(cache_policy: CachePolicy) -> u8 {
        match cache_policy {
            CachePolicy::Uncacheable => 0x00,
            CachePolicy::WriteCombining => 0x01,
            CachePolicy::WriteProtected => 0x05,
            CachePolicy::Writethrough => 0x04,
            CachePolicy::Writeback => 0x06,
        }
    }
```

Tangzh33: Thanks for your reply. I have added the requested comments. This PR is now ready for review again. Please leave your comments at your convenience~

lrh2000:
@tatetian It looks like this PR adds comments to the PCI crate, so GitHub requires some additional approvals. I guess I still need your help merging it :(

tatetian:
@Tangzh33 Need to fix the conflict. Thanks!

Tangzh33:
@tatetian Thanks for the reminder, and the conflicts have been resolved :-)

PR #2508, opened 2025-10-15
Adjust symbol names & Fix an off-by-one bug in `ostd::arch::trap`

lrh2000:
I doubt anyone will notice, since no code relies on the correctness of the RSP value in `TrapFrame`, but there is an off-by-one error:
https://github.com/asterinas/asterinas/blob/f69d39e9c16db9295f89c084376244d7265612de/ostd/src/arch/x86/trap/trap.S#L117-L118

There are 14 quad words that have been pushed onto the stack so far:
https://github.com/asterinas/asterinas/blob/f69d39e9c16db9295f89c084376244d7265612de/ostd/src/arch/x86/trap/trap.S#L108-L116
https://github.com/asterinas/asterinas/blob/f69d39e9c16db9295f89c084376244d7265612de/ostd/src/arch/x86/trap/trap.S#L100-L104

So the RSP value in `TrapFrame` is incorrect. But it will be ignored, so it will not cause any problems:
https://github.com/asterinas/asterinas/blob/f69d39e9c16db9295f89c084376244d7265612de/ostd/src/arch/x86/trap/trap.S#L139-L140

---

I also changed some symbol names. Since the symbols appear to be globally visible, even without the `.global` qualifier (e.g., they can be used to set breakpoints in GDB), names like `sysret` and `iret` might not be ideal.
https://github.com/asterinas/asterinas/blob/f69d39e9c16db9295f89c084376244d7265612de/ostd/src/arch/x86/trap/syscall.S#L84
https://github.com/asterinas/asterinas/blob/f69d39e9c16db9295f89c084376244d7265612de/ostd/src/arch/x86/trap/syscall.S#L94

This PR renames them to `_syscall_iret` and `_syscall_sysret`. I've considered several other alternatives:
 - `.Liret:` and `.Lsysret:`, which would make the symbols no longer globally visible. However, Rustc seems to compile all assembly files together, so conflicts will still occur if the same name appears in different assembly files (i.e., in different global_asm blocks). The resulting errors are too confusing.
 - `0:`, `1:`, e.t.c. This should work because numeric labels can appear multiple times, even in the same assembly file. However, we can no longer set break points on them.

PR #2571, opened 2025-11-11
Unify and fix RISC-V symbol names

junyang-zh:
The problem is as stated in the added section in the guide:

> Note that a Rust crate is a single translation unit, so `global_asm!` definitions in different modules within the same crate share the same global namespace. In RISC-V, we have experienced that defining two labels with the same name in different modules can nondeterministically cause [assembler errors](https://github.com/asterinas/asterinas/actions/runs/19164567169/job/54782072968?pr=2491) due to duplicate symbols. The `.L` prefix does not help. Therefore, add custom prefixes to your labels to avoid name clashes. Nevertheless, use `.global` to export symbols that need to be visible within the same crate.

Given that a crate is one TU, this behavior does not really count as a compiler bug. The Rust community seems like going to improve the situation by [renaming labels](https://github.com/rust-lang/rust/issues/81088). Before we can have that feature, let's name the labels with prefixes.


junyang-zh:
I've noticed https://github.com/asterinas/asterinas/pull/2508 , which has the same concerns. However uniformly using `_` as the symbol prefix seems don't help the issues here.

ostd/src/arch/riscv/boot/bsp_boot.S
```diff
@@ -19,6 +19,7 @@ KERNEL_VMA_OFFSET      = 0xffffffff00000000
 
 .section ".boot", "awx", @progbits
 
+# `_start` is also the entrypoint specified in the linker script.
```
lrh2000, line 22: Add this comment to the LoongArch's entry point as well?

ostd/src/arch/riscv/boot/bsp_boot.S
```diff
@@ -46,7 +47,7 @@ _start:
     # Check if the write to satp succeeds. If not, try Sv39.
     # Reference: <https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#satp>.
     csrr   t1, satp
-    beq    t0, t1, flush_tlb_bsp
+    beq    t0, t1, bsp_flush_tlb
```
lrh2000, line 50: Should we rename `riscv/boot/boot.S` to `bsp_boot.S`? This is more consistent with the x86 version (`x86/boot/bsp_boot.S`). Also, given that the symbols are prefixed with `bsp_`, it should make more sense.

book/src/to-contribute/style-guidelines/asm-guidelines.md
```diff
@@ -49,6 +49,15 @@ foo:
 The assembler treats the directives `.global` and `.globl` as the same.
 For clarity, however, `.global` is preferred.
 
+Note that a Rust crate is a single translation unit, so `global_asm!` definitions
+in different modules within the same crate share the same global namespace. In RISC-V,
+we have experienced that defining two labels with the same name in different modules
+can nondeterministically cause
+[assembler errors](https://github.com/asterinas/asterinas/actions/runs/19164567169/job/54782072968?pr=2491)
```
lrh2000: This link will become invalid after six months. Besides, I don't really know why the errors are nondeterministic?

junyang-zh: > nondeterministic

From observation, it depends on the number of [code-gen units](https://github.com/asterinas/asterinas/blob/0615e6f6424ead8e12d73670428149fa7270571f/Cargo.toml#L81) and the separation of code-gen units.

It must fail if LTO is enabled in that PR. But locally with debug mode I can compile it. In the main branch neither fails. So I think there's some nondeterminism.

junyang-zh:
@tatetian modified the book so need your approval.

StevenJiang1110:
It looks like the `asm-guidelines.md` is not rendered in the [Asterinas Book website](https://asterinas.github.io/book/to-contribute/style-guidelines/rust-guidelines.html). To make it be rendered, we should add this file to the book summary. Maybe this PR can also fix this problem?


```diff
diff --git a/book/src/SUMMARY.md b/book/src/SUMMARY.md
index deb2d3897..b4dbe4cf1 100644
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -59,6 +59,7 @@
     * [General Guidelines]() 
     * [Rust Guidelines](to-contribute/style-guidelines/rust-guidelines.md) 
     * [Git Guidelines]() 
+    * [Assembly Guidelines](to-contribute/style-guidelines/asm-guidelines.md)
 * [Boterinas](to-contribute/boterinas.md)
 * [Version Bump](to-contribute/version-bump.md)
 * [Community]()
```

Issue #2214, opened 2025-06-26
Tracking issue for supporting Podman on Asterinas


StevenJiang1110:
This issue tracks the lacking features for support Podman on Asterinas.

# What is Podman 

Podman is a container management tool used to create, manage, and run containers. Many of you are likely already familiar with Docker, and Podman provides a user interface that is nearly identical to Docker’s. According to the Podman documentation, in many cases, you can simply set `alias docker=podman` and run Docker commands without any issues. Commands such as podman pull, push, run, exec, and create share the same syntax and semantics as their Docker counterparts, making the transition straightforward.

### Podman Pods

One of the core concepts in Podman is the introduction of the "Pod" abstraction. Simply put, a Pod is a group of containers that share namespaces, cgroups, and filesystem volumes. This concept of a Pod is very similar to that in Kubernetes (k8s).

Kubernetes is a tool for container deployment and orchestration, where a Pod represents the smallest deployable unit. Typically, a Pod contains one or more containers that are tightly related from a business perspective. For example, a web server container and a database storage container might be deployed within the same Pod. In a distributed cluster, all resources within a Pod are guaranteed to run on the same physical node and share storage and network resources.

In Podman, each Pod contains a special "infra" container, as illustrated in the diagram below. This container doesn't perform any tasks; it simply sleeps. The purpose of the infra container is to ensure that the namespaces and other resources associated with the Pod are not released. On Linux systems, if all processes within a namespace exit, that namespace is destroyed. The infra container keeps the namespace alive as long as the Pod exists.

![Image](https://github.com/user-attachments/assets/887f7c00-1cda-4405-9c24-4368ac8d65eb)

### Podman vs Docker

Both Podman and Docker are OCI-compliant, meaning that they do not create containers themselves but instead invoke other container runtimes that adhere to the OCI standard to create containers. For example, Podman supports multiple runtimes such as runc, crun, krun, and runv. As a result, container images are generally interoperable between different tools.

Of course, there are also several differences between Podman and Docker:

1. Podman is daemonless, whereas Docker relies on a daemon (containerd). It is generally considered preferable to operate without a central daemon. In Docker’s architecture, containerd is responsible for managing all containers. If containerd crashes, all running containers become unmanageable.

[...]

PR #2572, opened 2025-11-11
Support socket options SO_BROADCAST and IP_RECVERR

StevenJiang1110:
This PR adds two socket options SO_BROADCAST and IP_RECVERR, which is required by #2214.

This PR now contains six commits:

1. The first commit is a cleanup commit. Previously the socket options were not listed in the same order as their definitions in `CSocketOptionName`; this commit makes the ordering consistent.
2. The second commit adds the IP level socket option support for UDP socket. 
3. The third commit is another cleanup. All `CHECK` macros are replaced with `TEST_SUCC` macros FN_TEST blocks in the sockoption.c.
4. The fourth commit adds the dummy support for IP_RECVERR.  According to [the manual](https://man7.org/linux/man-pages/man7/ip.7.html), this option queues generated errors in a per-socket error queue for UDP sockets, but implementing that behavior is complex, so this PR leaves it as a no-op for now.
5. The fifth and sixth commint adds support for SO_BROADCAST. Enabling this option allows UDP sockets to send to broadcast addresses. I added essential checks for connect and sendmsg on UDP sockets and a regression test that verifies a broadcast message can be received by a socket bound to the broadcast address. However, that regression test cannot be easily enabled on Asterinas because we do not yet support binding to addresses other than the interface address. For example, a socket can bind to 127.0.0.1 but not the broadcast address 127.255.255.255. Although the loopback device should permit binding to most addresses in the 127.0.0.0/8 subnet, this PR does not address that and leaves it as future work.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/iface/iface.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/iface.rs
@@ -2,7 +2,7 @@
 
 use alloc::sync::Arc;
 
-use smoltcp::wire::Ipv4Address;
+use smoltcp::wire::{Ipv4Address, Ipv4Cidr};
 
 use super::{port::BindPortConfig, BoundPort, InterfaceFlags, InterfaceType};
 use crate::{errors::BindError, ext::Ext};
@@ -76,6 +76,17 @@ impl<E: Ext> dyn Iface<E> {
     /// or both will return `None`.
     pub fn prefix_len(&self) -> Option<u8> {
         self.common().prefix_len()
+    }
+
+    /// Gets the broadcast address of the iface, if any.
+    pub fn broadcast_addr(&self) -> Option<Ipv4Address> {
+        let cidr = {
+            let common = self.common();
+            let ipv4_addr = common.ipv4_addr()?;
+            let prefix_len = common.prefix_len()?;
+            Ipv4Cidr::new(ipv4_addr, prefix_len)
+        };
+        cidr.broadcast()
     }
 
     /// Returns a reference to the associated [`ScheduleNextPoll`].
--- a/kernel/src/net/socket/ip/datagram/mod.rs
+++ b/kernel/src/net/socket/ip/datagram/mod.rs
@@ -11,17 +11,20 @@ use crate::{
     events::IoEvents,
     fs::utils::Inode,
     match_sock_option_mut,
-    net::socket::{
-        ip::options::{IpOptionSet, SetIpLevelOption},
-        new_pseudo_inode,
-        options::{Error as SocketError, SocketOption},
-        private::SocketPrivate,
-        util::{
-            datagram_common::{select_remote_and_bind, Bound, Inner},
-            options::{GetSocketLevelOption, SetSocketLevelOption, SocketOptionSet},
-            MessageHeader, SendRecvFlags, SocketAddr,
+    net::{
+        iface::is_broadcast_endpoint,
+        socket::{
+            ip::options::{IpOptionSet, SetIpLevelOption},
+            new_pseudo_inode,
+            options::{Error as SocketError, SocketOption},
+            private::SocketPrivate,
+            util::{
+                datagram_common::{select_remote_and_bind, Bound, Inner},
+                options::{GetSocketLevelOption, SetSocketLevelOption, SocketOptionSet},
+                MessageHeader, SendRecvFlags, SocketAddr,
+            },
+            Socket,
         },
-        Socket,
     },
     prelude::*,
     process::signal::{PollHandle, Pollable, Pollee},
@@ -147,6 +150,13 @@ impl Socket for DatagramSocket {
 
     fn connect(&self, socket_addr: SocketAddr) -> Result<()> {
         let endpoint = socket_addr.try_into()?;
+        let can_broadcast = self.options.read().socket.broadcast();
+        if !can_broadcast && is_broadcast_endpoint(&endpoint) {
+            return_errno_with_message!(
+                Errno::EACCES,
+                "connecting to a broadcast address without SO_BROADCAST is not allowed"
+            );
+        }
 
         self.inner.write().connect(&endpoint, &self.pollee)
     }
@@ -190,6 +200,16 @@ impl Socket for DatagramSocket {
             Some(addr) => Some(addr.try_into()?),
             None => None,
         };
+
+        if let Some(endpoint) = endpoint.as_ref() {
+            let can_broadcast = self.options.read().socket.broadcast();
+            if !can_broadcast && is_broadcast_endpoint(endpoint) {
+                return_errno_with_message!(
+                    Errno::EACCES,
+                    "sending to a broadcast address without SO_BROADCAST is not allowed"
+                );
+            }
+        }
 
         if !control_messages.is_empty() {
             // TODO: Support sending control message
--- a/kernel/src/net/iface/broadcast.rs
+++ b/kernel/src/net/iface/broadcast.rs
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use core::net::Ipv4Addr;
+
+use aster_bigtcp::wire::{IpAddress, IpEndpoint};
+use spin::Once;
+
+use crate::{net::iface::iter_all_ifaces, prelude::*};
+
+/// All known broadcast addresses.
+// FIXME: This information should be maintained in the routing table,
+// since a broadcast address might change if an interface's IP
+// or netmask changes, or if an interface is added/removed.
+static BROADCAST_ADDRS: Once<BTreeSet<Ipv4Addr>> = Once::new();
+
+pub(super) fn init() {
+    BROADCAST_ADDRS.call_once(|| {
+        let mut broadcast_addrs = BTreeSet::new();
+        // 255.255.255.255 is always included.
+        broadcast_addrs.insert(Ipv4Addr::BROADCAST);
+
+        for iface in iter_all_ifaces() {
+            let Some(broadcast_addr) = iface.broadcast_addr() else {
+                continue;
+            };
+
+            broadcast_addrs.insert(broadcast_addr);
+        }
+        broadcast_addrs
+    });
+}
+
+/// Determines if a given IP endpoint's address is a known broadcast address.
+pub fn is_broadcast_endpoint(endpoint: &IpEndpoint) -> bool {
+    let IpAddress::Ipv4(ipv4_addr) = &endpoint.addr;
+    BROADCAST_ADDRS.get().unwrap().contains(ipv4_addr)
+}
--- a/kernel/src/net/iface/init.rs
+++ b/kernel/src/net/iface/init.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use alloc::{borrow::ToOwned, sync::Arc};
+use alloc::borrow::ToOwned;
 use core::slice::Iter;
 
 use aster_bigtcp::{
@@ -11,7 +11,10 @@ use aster_softirq::BottomHalfDisabled;
 use spin::Once;
 
 use super::{poll::poll_ifaces, Iface};
-use crate::{net::iface::sched::PollScheduler, prelude::*};
+use crate::{
+    net::iface::{broadcast, sched::PollScheduler},
+    prelude::*,
+};
 
 static IFACES: Once<Vec<Arc<Iface>>> = Once::new();
 
@@ -50,6 +53,8 @@ pub fn init() {
         aster_network::register_recv_callback(VIRTIO_DEVICE_NAME, callback);
         aster_network::register_send_callback(VIRTIO_DEVICE_NAME, callback);
     }
+
+    broadcast::init();
 
     poll_ifaces();
 }
--- a/kernel/src/net/iface/mod.rs
+++ b/kernel/src/net/iface/mod.rs
@@ -1,10 +1,12 @@
 // SPDX-License-Identifier: MPL-2.0
 
+mod broadcast;
 mod ext;
 mod init;
 mod poll;
 mod sched;
 
+pub use broadcast::is_broadcast_endpoint;
 pub use init::{init, iter_all_ifaces, loopback_iface, virtio_iface};
 pub(super) use poll::init_in_first_kthread;
 
--- a/kernel/src/net/socket/options/mod.rs
+++ b/kernel/src/net/socket/options/mod.rs
@@ -15,6 +15,7 @@ pub trait SocketOption: Any + Send + Sync + Debug {
 impl_socket_options!(
     pub struct ReuseAddr(bool);
     pub struct Error(Option<crate::error::Error>);
+    pub struct Broadcast(bool);
     pub struct SendBuf(u32);
     pub struct RecvBuf(u32);
     pub struct KeepAlive(bool);
--- a/kernel/src/net/socket/util/options.rs
+++ b/kernel/src/net/socket/util/options.rs
@@ -11,8 +11,8 @@ use crate::{
     match_sock_option_mut, match_sock_option_ref,
     net::socket::{
         options::{
-            AcceptConn, KeepAlive, Linger, PassCred, PeerCred, PeerGroups, Priority, RecvBuf,
-            RecvBufForce, ReuseAddr, ReusePort, SendBuf, SendBufForce, SocketOption,
+            AcceptConn, Broadcast, KeepAlive, Linger, PassCred, PeerCred, PeerGroups, Priority,
+            RecvBuf, RecvBufForce, ReuseAddr, ReusePort, SendBuf, SendBufForce, SocketOption,
         },
         unix::{CUserCred, UNIX_DATAGRAM_DEFAULT_BUF_SIZE, UNIX_STREAM_DEFAULT_BUF_SIZE},
     },
@@ -25,6 +25,7 @@ use crate::{
 #[set = "pub"]
 pub struct SocketOptionSet {
     reuse_addr: bool,
+    broadcast: bool,
     send_buf: u32,
     recv_buf: u32,
     keep_alive: bool,
@@ -38,6 +39,7 @@ impl Default for SocketOptionSet {
     fn default() -> Self {
         Self {
             reuse_addr: false,
+            broadcast: false,
             send_buf: MIN_SENDBUF,
             recv_buf: MIN_RECVBUF,
             keep_alive: false,
@@ -100,6 +102,10 @@ impl SocketOptionSet {
             socket_reuse_addr: ReuseAddr => {
                 let reuse_addr = self.reuse_addr();
                 socket_reuse_addr.set(reuse_addr);
+            },
+            socket_broadcast: Broadcast => {
+                let broadcast = self.broadcast();
+                socket_broadcast.set(broadcast);
             },
             socket_send_buf: SendBuf => {
                 let send_buf = self.send_buf();
@@ -168,6 +174,10 @@ impl SocketOptionSet {
                 let reuse_addr = socket_reuse_addr.get().unwrap();
                 self.set_reuse_addr(*reuse_addr);
                 socket.set_reuse_addr(*reuse_addr);
+            },
+            socket_broadcast: Broadcast => {
+                let broadcast = socket_broadcast.get().unwrap();
+                self.set_broadcast(*broadcast);
             },
             socket_send_buf: SendBuf => {
                 let send_buf = socket_send_buf.get().unwrap();
--- a/kernel/src/util/net/options/socket.rs
+++ b/kernel/src/util/net/options/socket.rs
@@ -4,8 +4,8 @@ use super::RawSocketOption;
 use crate::{
     current_userspace, impl_raw_sock_option_get_only, impl_raw_socket_option,
     net::socket::options::{
-        AcceptConn, Error, KeepAlive, Linger, PassCred, PeerCred, PeerGroups, Priority, RecvBuf,
-        RecvBufForce, ReuseAddr, ReusePort, SendBuf, SendBufForce, SocketOption,
+        AcceptConn, Broadcast, Error, KeepAlive, Linger, PassCred, PeerCred, PeerGroups, Priority,
+        RecvBuf, RecvBufForce, ReuseAddr, ReusePort, SendBuf, SendBufForce, SocketOption,
     },
     prelude::*,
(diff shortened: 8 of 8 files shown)


lrh2000:
> 5\. Although the loopback device should permit binding to most addresses in the 127.0.0.0/8 subnet

It seems that the loopback interface is special. With all other interfaces, you can only bind to the interface address or the broadcast address.
```
    inet 10.4.227.181/22 brd 10.4.227.255 scope global dynamic noprefixroute wlan0
       valid_lft 12194sec preferred_lft 10394sec
```
```
>>> s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, 0)
>>> s.bind(('10.4.227.182', 2333))
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    s.bind(('10.4.227.182', 2333))
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 99] Cannot assign requested address
>>> s.bind(('10.4.227.255', 2333))
```

What would happen if we allowed binding to broadcast addresses?
https://github.com/asterinas/asterinas/blob/21365dd0bda83001737607428d51aed038dcecc3/kernel/src/net/socket/ip/common.rs#L14-L25

Will it cause issues and can it make the tests pass?

StevenJiang1110:
> What would happen if we allowed binding to broadcast addresses? 

I think broadcast can be supported and broadcast messages can be received. I ran local experiments with the unmodified smoltcp (v0.12.0), and they worked, so I guess our modified bigtcp will work as well.

<details><summary>A small smoltcp broadcast test case</summary>
<p>

```rust
use smoltcp::iface::{SocketHandle, SocketSet};
use smoltcp::socket::udp::Socket;
use smoltcp::wire::{IpAddress, IpEndpoint, Ipv4Address};
use smoltcp::{
    iface::{Config, Interface},
    phy::{Loopback, Medium},
    time::Instant,
    wire::{HardwareAddress, IpCidr, Ipv4Cidr},
};
const LOOPBACK_ADDRESS: Ipv4Address = Ipv4Address::new(127, 0, 0, 1);
const LOOPBACK_ADDRESS_PREFIX_LEN: u8 = 8;

fn main() {
    let mut loopback = Loopback::new(Medium::Ip);

    let mut interface = Interface::new(
        Config::new(HardwareAddress::Ip),
        &mut loopback,
        Instant::now(),
    );
    interface.update_ip_addrs(|ip_addrs| {
        debug_assert!(ip_addrs.is_empty());
        ip_addrs
            .push(IpCidr::Ipv4(Ipv4Cidr::new(
                LOOPBACK_ADDRESS,
                LOOPBACK_ADDRESS_PREFIX_LEN,
            )))
            .unwrap();
    });

    let mut sockets = SocketSet::new(Vec::new());

    let receive_broadcast_handle = {
        let mut udp_socket = new_udp_socket();
        udp_socket.bind(BORADCAST_ENDPOINT).unwrap();
        sockets.add(udp_socket)
    };

    let receive_unicast_handle = {
        let mut udp_socket = new_udp_socket();
        udp_socket.bind(UNICAST_ENDPOINT).unwrap();
        sockets.add(udp_socket)
    };

    let send_handle = {
        let mut udp_socket = new_udp_socket();
        udp_socket.bind(SENDER_ENDPOINT).unwrap();
         sockets.add(udp_socket)
    };

    test_broadcast(&mut sockets, &mut loopback, &mut interface, send_handle, receive_broadcast_handle);
    test_unicast(&mut sockets, &mut loopback, &mut interface, send_handle, receive_unicast_handle);

}

const SEND_DATA: [u8; 5] = [1u8, 2, 3, 4, 5];
const DATE_SIZE: usize = SEND_DATA.len();

fn test_broadcast(sockets: &mut SocketSet, loopback: &mut Loopback, interface: &mut Interface, send_handle: SocketHandle, receive_handle: SocketHandle) { 
    let send_socket: &mut Socket = sockets.get_mut(send_handle);

    let buffer = send_socket.send(DATE_SIZE, BORADCAST_ENDPOINT).unwrap();
    buffer[..DATE_SIZE].clone_from_slice(SEND_DATA.as_slice());

    for _ in 0..3 {
        let res = interface.poll(Instant::now(), loopback, sockets);
        println!("broadcast: poll res = {:?}", res);
    }

    let receive_socket: &mut Socket = sockets.get_mut(receive_handle);
    let (buffer, metadata) = receive_socket.recv().unwrap();
    assert_eq!(SEND_DATA.as_slice(), buffer);
    assert_eq!(metadata.endpoint, SENDER_ENDPOINT);
}

fn test_unicast(sockets: &mut SocketSet, loopback: &mut Loopback, interface: &mut Interface, send_handle: SocketHandle, receive_handle: SocketHandle) {
    let send_socket: &mut Socket = sockets.get_mut(send_handle);

    let buffer = send_socket.send(DATE_SIZE, UNICAST_ENDPOINT).unwrap();
    buffer[..DATE_SIZE].clone_from_slice(SEND_DATA.as_slice());

    for _ in 0..3 {
        let res = interface.poll(Instant::now(), loopback, sockets);
        println!("unicast: poll res = {:?}", res);
    }

    let receive_socket: &mut Socket = sockets.get_mut(receive_handle);
    let (buffer, metadata) = receive_socket.recv().unwrap();
    assert_eq!(SEND_DATA.as_slice(), buffer);
    assert_eq!(metadata.endpoint, SENDER_ENDPOINT);
}

const UNICAST_ENDPOINT: IpEndpoint =
    IpEndpoint::new(IpAddress::Ipv4(Ipv4Address::new(127, 0, 0, 1)), 2345);
const BORADCAST_ENDPOINT: IpEndpoint =
    IpEndpoint::new(IpAddress::Ipv4(Ipv4Address::new(127, 255, 255, 255)), 1234);
const SENDER_ENDPOINT: IpEndpoint =
    IpEndpoint::new(IpAddress::Ipv4(Ipv4Address::new(127, 0, 0, 1)), 3456);

fn new_udp_socket() -> Socket<'static> {
    const UDP_METADATA_LEN: usize = 256;
    const UDP_PAYLOAD_LEN: usize = 65536;

    let metadata = smoltcp::socket::udp::PacketMetadata::EMPTY;
    let rx_buffer = smoltcp::socket::udp::PacketBuffer::new(
        vec![metadata; UDP_METADATA_LEN],
        vec![0u8; UDP_PAYLOAD_LEN],
    );
    let tx_buffer = smoltcp::socket::udp::PacketBuffer::new(
        vec![metadata; UDP_METADATA_LEN],
        vec![0u8; UDP_PAYLOAD_LEN],
    );

    Socket::new(rx_buffer, tx_buffer)
}
```
</p>
</details>

> Will it cause issues and can it make the tests pass?

Most test cases should pass; I don't see any obvious failures. The problem is that this change touches a lot of code. For example, we can no longer keep bound_ports inside the interface because an interface may now have multiple IP addresses. We will likely need a Linux‑style bind table that records both IP address and port. In addition, many parts of the code currently construct a socket address from the interface address and port number; those sites will also need to be updated. I'd prefer to postpone these refactors if possible, although we will probably have to address them eventually.

lrh2000:
> In addition, many parts of the code currently construct a socket address from the interface address and port number

Actually I think we can do something like:
<details>
<summary> allow to bind to broadcast addresses </summary>

```patch
diff --git a/kernel/libs/aster-bigtcp/src/iface/common.rs b/kernel/libs/aster-bigtcp/src/iface/common.rs
index 028166b0e..1b56d3625 100644
--- a/kernel/libs/aster-bigtcp/src/iface/common.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/common.rs
@@ -122,9 +122,10 @@ impl<E: Ext> IfaceCommon<E> {
         iface: Arc<dyn Iface<E>>,
         config: BindPortConfig,
     ) -> core::result::Result<BoundPort<E>, BindError> {
-        let (port, can_reuse) = self.bind_port(config)?;
+        let (port, can_reuse) = self.bind_port(&config)?;
         Ok(BoundPort {
             iface,
+            is_boardcast: config.is_boardcast(),
             port,
             can_reuse: AtomicBool::new(can_reuse),
         })
@@ -151,7 +152,7 @@ impl<E: Ext> IfaceCommon<E> {
         None
     }
 
-    fn bind_port(&self, config: BindPortConfig) -> Result<(u16, bool), BindError> {
+    fn bind_port(&self, config: &BindPortConfig) -> Result<(u16, bool), BindError> {
         let mut used_ports = self.used_ports.lock();
         let config_can_reuse = config.can_reuse();
 
@@ -168,8 +169,7 @@ impl<E: Ext> IfaceCommon<E> {
             // FIXME: If the socket is not a backlog socket,
             // we should check whether there is a listening socket on the port.
             // If there is, the socket cannot be bound to that port.
-            let can_reuse = matches!(config, BindPortConfig::Backlog(_))
-                || (port_state.can_reuse() & config_can_reuse);
+            let can_reuse = config.is_backlog() || (port_state.can_reuse() & config_can_reuse);
             if can_reuse {
                 port_state.nsocket += 1;
                 if config_can_reuse {
@@ -272,8 +272,10 @@ impl<E: Ext> IfaceCommon<E> {
 /// When dropped, the port is automatically released.
 //
 // FIXME: TCP and UDP ports are independent. Find a way to track the protocol here.
+// FIXME: Ports on the interface address and the broadcast address should be independent.
 pub struct BoundPort<E: Ext> {
     iface: Arc<dyn Iface<E>>,
+    is_boardcast: bool,
     port: u16,
     can_reuse: AtomicBool,
 }
@@ -289,11 +291,19 @@ impl<E: Ext> BoundPort<E> {
         self.port
     }
 
+    /// Returns whether the bound address is the broadcast address.
+    pub(crate) fn is_boardcast(&self) -> bool {
+        self.is_boardcast
+    }
+
     /// Returns the bound endpoint.
     pub fn endpoint(&self) -> Option<IpEndpoint> {
-        let ip_addr = {
+        let ip_addr = if !self.is_boardcast {
             let ipv4_addr = self.iface().ipv4_addr()?;
             IpAddress::Ipv4(ipv4_addr)
+        } else {
+            let ipv4_addr = self.iface().broadcast_addr()?;
+            IpAddress::Ipv4(ipv4_addr)
         };
         Some(IpEndpoint::new(ip_addr, self.port))
     }
diff --git a/kernel/libs/aster-bigtcp/src/iface/port.rs b/kernel/libs/aster-bigtcp/src/iface/port.rs
index 8e38a3b5a..39f1c85d3 100644
--- a/kernel/libs/aster-bigtcp/src/iface/port.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/port.rs
@@ -1,7 +1,12 @@
 
 /// The configuration using for bind to a TCP/UDP port.
-pub enum BindPortConfig {
+pub struct BindPortConfig {
+    is_boardcast: bool,
+    inner: Inner,
+}
+
+enum Inner {
     /// Binds to the specified non-reusable port.
     CanReuse(u16),
     /// Binds to the specified reusable port.
@@ -14,22 +19,42 @@ pub enum BindPortConfig {
 
 impl BindPortConfig {
     /// Creates new configuration using for bind to a TCP/UDP port.
-    pub fn new(port: u16, can_reuse: bool) -> Self {
-        match (port, can_reuse) {
-            (0, can_reuse) => Self::Ephemeral(can_reuse),
-            (_, true) => Self::CanReuse(port),
-            (_, false) => Self::Specified(port),
+    pub fn new(port: u16, can_reuse: bool, is_boardcast: bool) -> Self {
+        let inner = match (port, can_reuse) {
+            (0, can_reuse) => Inner::Ephemeral(can_reuse),
+            (_, true) => Inner::CanReuse(port),
+            (_, false) => Inner::Specified(port),
+        };
+        Self {
+            is_boardcast,
+            inner,
+        }
+    }
+
+    pub(crate) fn new_backlog(port: u16, is_boardcast: bool) -> Self {
+        let inner = Inner::Backlog(port);
+        Self {
+            is_boardcast,
+            inner,
         }
     }
 
     pub(super) fn can_reuse(&self) -> bool {
-        matches!(self, Self::CanReuse(_)) || matches!(self, Self::Ephemeral(true))
+        matches!(self.inner, Inner::CanReuse(_)) || matches!(self.inner, Inner::Ephemeral(true))
+    }
+
+    pub(super) fn is_boardcast(&self) -> bool {
+        self.is_boardcast
+    }
+
+    pub(super) fn is_backlog(&self) -> bool {
+        matches!(self.inner, Inner::Backlog(_))
     }
 
     pub(super) fn port(&self) -> Option<u16> {
-        match self {
-            Self::CanReuse(port) | Self::Specified(port) | Self::Backlog(port) => Some(*port),
-            Self::Ephemeral(_) => None,
+        match &self.inner {
+            Inner::CanReuse(port) | Inner::Specified(port) | Inner::Backlog(port) => Some(*port),
+            Inner::Ephemeral(_) => None,
         }
     }
 }
diff --git a/kernel/libs/aster-bigtcp/src/socket/bound/tcp_listen.rs b/kernel/libs/aster-bigtcp/src/socket/bound/tcp_listen.rs
index 20ed84838..e07a70794 100644
--- a/kernel/libs/aster-bigtcp/src/socket/bound/tcp_listen.rs
+++ b/kernel/libs/aster-bigtcp/src/socket/bound/tcp_listen.rs
@@ -236,7 +236,10 @@ impl<E: Ext> TcpListenerBg<E> {
         let conn = TcpConnection::new_cyclic(
             self.bound
                 .iface()
-                .bind(BindPortConfig::Backlog(self.bound.port()))
+                .bind(BindPortConfig::new_backlog(
+                    self.bound.port(),
+                    self.bound.is_boardcast(),
+                ))
                 .unwrap(),
             |weak| {
                 TcpConnectionInner::new(
diff --git a/kernel/src/net/socket/ip/common.rs b/kernel/src/net/socket/ip/common.rs
index 91970be24..5d6be8987 100644
--- a/kernel/src/net/socket/ip/common.rs
+++ b/kernel/src/net/socket/ip/common.rs
@@ -11,17 +11,21 @@ use crate::{
     prelude::*,
 };
 
-pub(super) fn get_iface_to_bind(ip_addr: &IpAddress) -> Option<Arc<Iface>> {
+fn get_iface_to_bind(ip_addr: &IpAddress) -> Option<(Arc<Iface>, bool)> {
     let IpAddress::Ipv4(ipv4_addr) = ip_addr;
-    iter_all_ifaces()
-        .find(|iface| {
-            if let Some(iface_ipv4_addr) = iface.ipv4_addr() {
-                iface_ipv4_addr == *ipv4_addr
-            } else {
-                false
-            }
-        })
-        .map(Clone::clone)
+    for iface in iter_all_ifaces() {
+        if let Some(iface_ipv4_addr) = iface.ipv4_addr()
+            && iface_ipv4_addr == *ipv4_addr
+        {
+            return Some((iface.clone(), false));
+        }
+        if let Some(iface_ipv4_addr) = iface.broadcast_addr()
+            && iface_ipv4_addr == *ipv4_addr
+        {
+            return Some((iface.clone(), true));
+        }
+    }
+    None
 }
 
 /// Get a suitable iface to deal with sendto/connect request if the socket is not bound to an iface.
@@ -49,7 +53,7 @@ fn get_ephemeral_iface(remote_ip_addr: &IpAddress) -> Arc<Iface> {
 }
 
 pub(super) fn bind_port(endpoint: &IpEndpoint, can_reuse: bool) -> Result<BoundPort> {
-    let iface = match get_iface_to_bind(&endpoint.addr) {
+    let (iface, is_broadcast) = match get_iface_to_bind(&endpoint.addr) {
         Some(iface) => iface,
         None => {
             return_errno_with_message!(
@@ -59,7 +63,7 @@ pub(super) fn bind_port(endpoint: &IpEndpoint, can_reuse: bool) -> Result<BoundP
         }
     };
 
-    let bind_port_config = BindPortConfig::new(endpoint.port, can_reuse);
+    let bind_port_config = BindPortConfig::new(endpoint.port, can_reuse, is_broadcast);
 
     Ok(iface.bind(bind_port_config)?)
 }
```
</details>

I've tested this, but the test still fails. I think it's due to other bugs related to the broadcast address:
https://github.com/asterinas/asterinas/blob/76c7752130316e9ce605aada13d49a09463ac0f7/kernel/libs/aster-bigtcp/src/iface/poll.rs#L507

We should check [`iface.context().is_broadcast{,_v4}`](https://github.com/smoltcp-rs/smoltcp/blob/39cd44e585aae109f37a2bc9f31376414968456d/src/iface/interface/ipv4.rs#L64-L79) instead of `dst_addr().is_broadcast()`.

Maybe creating a follow-up PR for this one is okay. So this PR won't be blocked.

StevenJiang1110:
> Maybe creating a follow-up PR for this one is okay. So this PR won't be blocked.

Thanks for testing! Then we can enable UDP broadcast in a follow-up PR for fixing these problems.

kernel/src/net/socket/ip/datagram/mod.rs
```diff
@@ -147,6 +150,10 @@ impl Socket for DatagramSocket {
 
     fn connect(&self, socket_addr: SocketAddr) -> Result<()> {
         let endpoint = socket_addr.try_into()?;
+        let can_broadcast = self.options.read().socket.broadcast();
+        if !can_broadcast && is_broadcast_endpoint(&endpoint) {
+            return_errno_with_message!(Errno::EACCES, "trying to connect to a broadcast address");
```
lrh2000:
```suggestion
            return_errno_with_message!(Errno::EACCES, "connecting to a broadcast address without SO_BROADCAST is not allowed");
```

kernel/src/net/socket/ip/datagram/mod.rs
```diff
@@ -191,6 +198,16 @@ impl Socket for DatagramSocket {
             None => None,
         };
 
+        if let Some(endpoint) = endpoint.as_ref() {
+            let can_broadcast = self.options.read().socket.broadcast();
+            if !can_broadcast && is_broadcast_endpoint(endpoint) {
+                return_errno_with_message!(
+                    Errno::EACCES,
+                    "trying to send message to a broadcast address"
```
lrh2000:
```suggestion
                    "sending to a broadcast address without SO_BROADCAST is not allowed"
```

test/src/apps/network/udp_broadcast.c
```diff
@@ -0,0 +1,161 @@
...
+	CHECK(inet_aton(BROADCAST_ADDR, &broadcast_addr.sin_addr));
+	broadcast_addr.sin_port = htons(RECEIVE_PORT);
+	broadcast_addr.sin_family = AF_INET;
+
+	// Bind sender to SENDER_BIND_ADDR:SENDER_PORT
+	struct sockaddr_in sk_addr;
+	sk_addr.sin_family = AF_INET;
+	CHECK(inet_aton(SENDER_BIND_ADDR, &sk_addr.sin_addr));
+	sender = CHECK(socket(AF_INET, SOCK_DGRAM, 0));
+	sk_addr.sin_port = htons(SENDER_PORT);
```
lrh2000: Reorder the two statements:
```rust
	sk_addr.sin_port = htons(SENDER_PORT);
	sender = CHECK(socket(AF_INET, SOCK_DGRAM, 0));
```

test/src/apps/network/udp_broadcast.c
```diff
@@ -0,0 +1,161 @@
...
+	CHECK(bind(sender, (struct sockaddr *)&sk_addr, addr_len));
+
+	// Bind receiver to BROADCAST_ADDR:RECEIVE_PORT
+	receiver = CHECK(socket(AF_INET, SOCK_DGRAM, 0));
+	// FIXME: Asterinas cannot support binding to broadcast addresses now.
+	// So all below code related to receiver is commented out.
+	// CHECK(inet_aton(BROADCAST_ADDR, &sk_addr.sin_addr));
+	// sk_addr.sin_port = htons(RECEIVE_PORT);
```
lrh2000: Why not use `broadcast_addr`?

test/src/apps/network/udp_broadcast.c
```diff
@@ -0,0 +1,161 @@
...
+#define MESSAGE "Hello from broadcast"
+
+int sender;
+int receiver;
+char buf[128] = { 0 };
+struct sockaddr_in broadcast_addr;
+struct sockaddr_in received_addr;
+socklen_t addr_len = sizeof(struct sockaddr_in);
+int broadcast_opt;
+socklen_t broadcast_opt_len = sizeof(broadcast_opt);
```
lrh2000, line 29: Add some blank lines?
```c
int sender;
int receiver;
char buf[128] = { 0 };

struct sockaddr_in broadcast_addr;
struct sockaddr_in received_addr;
socklen_t addr_len = sizeof(struct sockaddr_in);

int broadcast_opt;
socklen_t broadcast_opt_len = sizeof(broadcast_opt);
```

test/src/apps/network/udp_broadcast.c
```diff
@@ -0,0 +1,161 @@
...
+
+FN_SETUP(create_and_bind)
+{
+	// Initialize the broadcast_addr
```
lrh2000: `Initialize the broadcast address` or `Initialize broadcast_addr`

lrh2000 (commented):
Thanks for cleaning up the code and adding the new socket options. It looks good and I only have a few minor comments about the last two commits.

test/src/apps/network/udp_broadcast.c
```diff
@@ -0,0 +1,161 @@
...
+#include <unistd.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include "../test.h"
+
+#define SENDER_PORT 12345
+#define RECEIVE_PORT 12346
+#define BROADCAST_ADDR "127.255.255.255"
+#define SENDER_BIND_ADDR "127.0.0.1"
+#define MESSAGE "Hello from broadcast"
```
lrh2000, line 17:
```c
#define SENDER_BIND_ADDR "127.0.0.1"
#define SENDER_PORT 12345

#define BROADCAST_ADDR "127.255.255.255"
#define RECEIVE_PORT 12346

#define MESSAGE "Hello from broadcast"
#define MESSAGE_LEN sizeof(MESSAGE)
```

This is clearer and I think we can avoid `strlen(MESSAGE)` as well.

kernel/libs/aster-bigtcp/src/iface/iface.rs
```diff
@@ -78,6 +78,17 @@ impl<E: Ext> dyn Iface<E> {
         self.common().prefix_len()
     }
 
+    /// Gets the broadcast of the iface, if any.
```
lrh2000:
```suggestion
    /// Gets the broadcast address of the iface, if any.
```

PR #2573, opened 2025-11-11
Revise some names and comments in `ostd::arch`

lrh2000:
- Fix some minor comment issues, like:
```patch
-/// Activates the given level 4 page table.
+/// Activates the given root-level page table.
```

 - https://github.com/asterinas/asterinas/pull/2508 adjusts some symbol names under `arch/x86/trap/`, leaving other architectures' symbol names in an inconsistent state. This PR fixes them.

### DIFF
--- a/ostd/src/arch/loongarch/mm/mod.rs
+++ b/ostd/src/arch/loongarch/mm/mod.rs
@@ -114,14 +114,14 @@ pub(crate) unsafe fn sync_dma_range(_range: Range<Vaddr>, _direction: DmaDirecti
 #[repr(C)]
 pub(crate) struct PageTableEntry(usize);
 
-/// Activates the given level 4 page table.
+/// Activates the given root-level page table.
 ///
 /// "pgdl" or "pgdh" register doesn't have a field that encodes the cache policy,
 /// so `_root_pt_cache` is ignored.
 ///
 /// # Safety
 ///
-/// Changing the level 4 page table is unsafe, because it's possible to violate memory safety by
+/// Changing the root-level page table is unsafe, because it's possible to violate memory safety by
 /// changing the page mapping.
 pub(crate) unsafe fn activate_page_table(root_paddr: Paddr, _root_pt_cache: CachePolicy) {
     assert!(root_paddr % PagingConsts::BASE_PAGE_SIZE == 0);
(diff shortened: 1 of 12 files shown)


junyang-zh:
So, should we name all labels starting with `_`? or just `.global` ones? or just under `ostd::arch::trap`? I haven't found the rules in the guide.

I am also fixing problems with name collisions in #2571 , perhaps we can form a holistic approach now.

lrh2000:
> So, should we name all labels starting

This is a good question and I have the same question, but:

> just under `ostd::arch::trap`

We don't have a global rule, so... I'm following this.

https://github.com/asterinas/asterinas/pull/2508 said:
>> This PR renames them to `_syscall_iret` and `_syscall_sysret`. I've considered several other alternatives:
>>  - `.Liret:` and `.Lsysret:`, which would make the symbols no longer globally visible. However, Rustc seems to compile all assembly files together, so conflicts will still occur if the same name appears in different assembly files (i.e., in different global_asm blocks). The resulting errors are too confusing.
>> - `0:`, `1:`, e.t.c. This should work because numeric labels can appear multiple times, even in the same assembly file. However, we can no longer set break points on them.

Currently, we sometimes prefix global names with `__`, as in:
https://github.com/asterinas/asterinas/blob/0615e6f6424ead8e12d73670428149fa7270571f/ostd/src/arch/x86/mm/util.rs#L9-L27

I really think the names are a mess, but I don't know how to clean them up properly. Cleaning up names within a single module should be simpler, though.

lrh2000:
> So, should we name all labels starting with `_`?

https://github.com/asterinas/asterinas/pull/2508 adds the underscore prefix for _local_ labels because I feel that `_syscall_iret` is different from `syscall_return`. `_syscall_iret` is a local label that belongs to the global function `syscall_return`. `trap.S` can contain several trampolines, e.g., syscall/exception entry/return, so I think doing so will clarify things.

For the code under `boot/`, I generally feel that https://github.com/asterinas/asterinas/pull/2571 is okay because it seems `boot.S` contains only one boot trampoline (so a symbol does not need to belong to something; it can only belong to the boot entrypoint.)

junyang-zh (approved):
LGTM!

PR #2575, opened 2025-11-12
Fix error behavior in various `sched_*` syscalls

lrh2000:
- Things like this are clearly wrong:
https://github.com/asterinas/asterinas/blob/0615e6f6424ead8e12d73670428149fa7270571f/kernel/src/syscall/sched_setattr.rs#L20

It is unreasonable to map `Errno::EFAULT` to `Errno::EINVAL`. In fact, the Linux behavior is to [return `EINVAL` if the provided address is NULL](https://elixir.bootlin.com/linux/v6.17.7/source/kernel/sched/syscalls.c#L983-L984), instead of returning `EINVAL` if there are page faults.

 - Meanwhile, this is unbounded allocation, which can cause the kernel to panic. We should also never perform unbounded allocation:
https://github.com/asterinas/asterinas/blob/0615e6f6424ead8e12d73670428149fa7270571f/kernel/src/syscall/sched_getattr.rs#L157

 - `read_linux_sched_attr_from_user` also has several behavior that is different from the Linux implementation. For example, we should [report the correct size to the user if we return `E2BIG`](https://elixir.bootlin.com/linux/v6.17.7/source/kernel/sched/syscalls.c#L926-L928).

### DIFF
--- a/kernel/src/syscall/sched_getattr.rs
+++ b/kernel/src/syscall/sched_getattr.rs
@@ -11,6 +11,7 @@ use crate::{
     process::posix_thread::thread_table,
     sched::{Nice, RealTimePolicy, SchedAttr, SchedPolicy},
     thread::Tid,
+    util::CopyCompat,
 };
 
 pub(super) const SCHED_NORMAL: u32 = 0;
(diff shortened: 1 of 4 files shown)


kernel/src/syscall/sched_getattr.rs
```diff
@@ -139,29 +140,48 @@ pub(super) fn read_linux_sched_attr_from_user(
...
+    while reader.remain() > size_of::<u64>() {
+        if reader.read_val::<u64>()? != 0 {
+            let _ = user_space.write_val(addr, &(size_of::<LinuxSchedAttr>() as u32));
+            return_errno_with_message!(Errno::E2BIG, "incompatible scheduling attributes");
+        }
+    }
+    while reader.has_remain() {
+        if reader.read_val::<u8>()? != 0 {
+            let _ = user_space.write_val(addr, &(size_of::<LinuxSchedAttr>() as u32));
+            return_errno_with_message!(Errno::E2BIG, "incompatible scheduling attributes");
```
StevenJiang1110: It seems these two while loop performs the check documented `copy_struct_from_user` to ensure all remaining bytes are zero. I guess such a behavior will also be used in other code. So I suggest moving such code to `kernel/util/copy_compat.rs`.

```rust
pub trait ReadCompat {
    fn read_val_compat<T: Pod>(&mut self, dst: &mut T) -> Result<()>;
}

impl ReadCompat for VmReader {
    // TODO
}

// The write version if required..
```

kernel/src/syscall/sched_getparam.rs
```diff
@@ -4,16 +4,18 @@ use super::{sched_getattr::access_sched_attr_with, SyscallReturn};
...
         SchedPolicy::RealTime { rt_prio, .. } => rt_prio.get().into(),
         _ => 0,
     };
 
-    let space = ctx.user_space();
-    space
-        .write_val(addr, &rt_prio)
-        .map_err(|_| Error::new(Errno::EINVAL))?;
+    let user_space = ctx.user_space();
+    user_space.write_val(addr, &rt_prio)?;
```
StevenJiang1110: Simplified as `ctx.user_space().write_val(addr, &rt_prio)?` since user_space is not used anywhere else?

kernel/src/syscall/sched_getattr.rs
```diff
@@ -47,6 +49,14 @@ pub(super) struct LinuxSchedAttr {
     pub(super) sched_util_max: u32,
 }
 
+// Reference: <https://elixir.bootlin.com/linux/v6.17.7/source/include/uapi/linux/sched/types.h#L7>
+const SCHED_ATTR_SIZE_VER0: u32 = 48;
+// Reference: <https://elixir.bootlin.com/linux/v6.17.7/source/include/uapi/linux/sched/types.h#L8>
+#[cfg_attr(target_arch = "x86_64", expect(dead_code))]
+const SCHED_ATTR_SIZE_VER1: u32 = 56;
```
StevenJiang1110, line 57: Why is `SCHED_ATTR_SIZE_VER1` dead code only on x86_64? It looks like other architectures don’t reference it either.

StevenJiang1110 (commented):
Thanks for your contribution! Just two small nits.

kernel/src/syscall/sched_getattr.rs
lrh2000: I have the same question. But I guess it's because:
https://github.com/asterinas/asterinas/blob/121da78aa2576da02294662ee1ca279aa1316c0f/kernel/src/syscall/mod.rs#L5-L8

StevenJiang1110: Sounds reasonable! Thanks for your investigation!

PR #2547, opened 2025-10-31
Full (<=32) RISC-V SMP support

junyang-zh:
This PR allows Asterinas to boot and run on <= 48 core RISC-V QEMU VMs. Under 64 cores, there seems to be either an SBI or a QEMU bug that prevents some harts being kicked up. Nevertheless, this is a good progress.

~~The first commit is #2368 . I'd suggest reviewing and merging that PR first so there could be iterative progresses.~~ (already merged!)

The second commit implements IPI with RISC-V software interrupt. And the following commits fix some orignally SMP-agnostic code.

PR #2576, opened 2025-11-12
Rename `trap::init` to `trap::init_on_cpu`

junyang-zh:
Following https://github.com/asterinas/asterinas/pull/2547#discussion_r2514892431 , but I think it is orthogonal to #2547 so this is made a separate PR.

### DIFF
--- a/ostd/src/arch/loongarch/trap/mod.rs
+++ b/ostd/src/arch/loongarch/trap/mod.rs
@@ -18,9 +18,20 @@ use crate::{
 };
 
 /// Initializes trap handling on LoongArch.
-pub(crate) unsafe fn init() {
+///
+/// This function will:
+/// - Set `ecfg`'s VS bit to zero.
+/// - Set `eentry` to the trap entry.
+///
+/// # Safety
+///
+/// On the current CPU, this function must be called
+/// - only once and
+/// - before any trap can occur.
+pub(crate) unsafe fn init_on_cpu() {
+    // SAFETY: The caller ensures the safety conditions.
     unsafe {
-        self::trap::init();
+        self::trap::init_on_cpu();
     }
 }
 
--- a/ostd/src/arch/loongarch/trap/trap.rs
+++ b/ostd/src/arch/loongarch/trap/trap.rs
@@ -10,11 +10,10 @@ global_asm!(include_str!("trap.S"));
 ///
 /// # Safety
 ///
-/// This function will:
-/// - Set `eentry` to `trap_entry`
-///
-/// You **MUST NOT** modify these registers later.
-pub(super) unsafe fn init() {
+/// On the current CPU, this function must be called
+/// - only once and
+/// - before any trap can occur.
+pub(super) unsafe fn init_on_cpu() {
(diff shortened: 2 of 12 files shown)


lrh2000:
- Could you also please rename `gdt::init`, `idt::init`, and `syscall::init`?
- Also, the safety comment should be consistent between architectures unless there is a special reason not to.
https://github.com/asterinas/asterinas/blob/d8e719a953d9451d43ae379f824dfb72651d371c/ostd/src/arch/x86/trap/mod.rs#L116-L118
https://github.com/asterinas/asterinas/blob/d8e719a953d9451d43ae379f824dfb72651d371c/ostd/src/arch/loongarch/trap/mod.rs#L24-L26
- Also check and update all the callers' safety comments.
- In addition, the commit message is broken.

ostd/src/boot/smp.rs
```diff
@@ -133,8 +133,8 @@ fn ap_early_entry(cpu_id: u32) -> ! {
 
     crate::arch::enable_cpu_features();
 
-    // SAFETY: This function is called in the boot context of the AP.
-    unsafe { crate::arch::trap::init() };
+    // SAFETY: This is called only once in the boot context of this AP.
```
lrh2000:
```suggestion
    // SAFETY: This is called only once on this AP in the boot context.
```

This aligns to the `late_init_on_bsp` version.

ostd/src/arch/loongarch/trap/trap.rs
```diff
@@ -10,11 +10,10 @@ global_asm!(include_str!("trap.S"));
 ///
 /// # Safety
 ///
-/// This function will:
-/// - Set `eentry` to `trap_entry`
```
lrh2000, line 14: It seems that this comment is kept in RISC-V version. We previously also kept the x86's comment. Then I wander why remove it in the LoongArch version?
https://github.com/asterinas/asterinas/blob/121da78aa2576da02294662ee1ca279aa1316c0f/ostd/src/arch/x86/trap/mod.rs#L105-L114

Maybe it's incomplete?
```rust
/// This function will:
/// - Set `ecfg`'s VS bit to zero.
/// - Set `eentry` to the trap entry.
```

ostd/src/arch/loongarch/trap/mod.rs
```diff
@@ -18,9 +18,16 @@ use crate::{
 };
 
 /// Initializes trap handling on LoongArch.
-pub(crate) unsafe fn init() {
+///
+/// # Safety
+///
+/// This function should be called:
+///  - only once; and
+///  - before any trap can occur on the current CPU.
```
lrh2000: Actually I think this writing is problematic. So it is:
 1. This function should be called only once;
 2. This function should be called before any trap can occur on the current CPU.

However, this function is called in both the BSP and the APs. Therefore, this function will actually be called more than once?

We should move "on the current CPU" to a correct position? Like this:
```rust
/// On the current CPU, this function must be called
/// - only once and
/// - before any trap can occur.
```
(Also try to replace `should` with `must` in the safety comments.)

PR #2581, opened 2025-11-13
Read pty master should block if no pty slave is opened

StevenJiang1110:
This PR fixes a problem introduced by #2550 that caused nix-env package downloads to fail.

After #2550, reading from the pty master would return EIO immediately when there were no open slaves (and no data). Returning EIO is correct if slaves were previously opened and then closed, but it is incorrect when only the master has ever been opened: in that case reads should block instead of returning EIO.

This PR restores the expected behavior by introducing the same tty flags as Linux, including a flag bit that represents the other side has been closed. The bit is cleared when pty master is opened and be set when all slaves are closed.

This PR also fixes the pty lock handling: a pty master is locked when it is created. While locked, the pty slave cannot be opened either via open(/dev/pts/index) or via ioctl(master, TIOCGPTPEER). Slaves may only be opened after the pty lock is released.

### DIFF
--- a/kernel/src/device/pty/file.rs
+++ b/kernel/src/device/pty/file.rs
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use core::sync::atomic::Ordering;
-
 use inherit_methods_macro::inherit_methods;
 
 use crate::{
@@ -19,21 +17,40 @@ use crate::{
 pub(super) struct PtySlaveFile(Arc<PtySlave>);
 
 impl PtySlaveFile {
-    pub(super) fn new(slave: Arc<PtySlave>) -> PtySlaveFile {
-        slave
-            .driver()
-            .opened_slaves()
-            .fetch_add(1, Ordering::Relaxed);
+    pub(super) fn new(slave: Arc<PtySlave>) -> Result<PtySlaveFile> {
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L220>.
+
+        // Hold the lock at first to avoid race condition.
+        let mut opened_slaves = slave.driver().opened_slaves().lock();
+
+        let master_flags = slave.driver().tty_flags();
+        if master_flags.is_pty_locked() || slave.tty_flags().is_other_closed() {
+            return_errno_with_message!(
+                Errno::EIO,
+                "a pty slave cannot be opened when the pty master is locked or closed"
+            );
+        }
+        master_flags.clear_other_closed();
+
+        *opened_slaves += 1;
+        drop(opened_slaves);
+
         slave.driver().pollee().invalidate();
-        PtySlaveFile(slave)
+        Ok(PtySlaveFile(slave))
     }
 }
 
 impl Drop for PtySlaveFile {
     fn drop(&mut self) {
         let driver = self.0.driver();
-        driver.opened_slaves().fetch_sub(1, Ordering::Relaxed);
-        driver.pollee().notify(IoEvents::HUP);
+
+        let mut opened_slaves = driver.opened_slaves().lock();
+        *opened_slaves -= 1;
+
+        if *opened_slaves == 0 {
+            driver.tty_flags().set_other_closed();
+            driver.pollee().notify(IoEvents::HUP);
+        }
     }
 }
 
--- a/kernel/src/device/pty/master.rs
+++ b/kernel/src/device/pty/master.rs
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use alloc::format;
-use core::sync::atomic::Ordering;
 
 use ostd::task::Task;
 
 use super::{driver::PtyDriver, PtySlave};
 use crate::{
     current_userspace,
+    device::tty::TtyFlags,
     events::IoEvents,
     fs::{
         devpts::Ptmx,
@@ -51,6 +51,14 @@ impl PtyMaster {
         &self.slave
     }
 
+    fn master_flags(&self) -> &TtyFlags {
+        self.slave.driver().tty_flags()
+    }
+
+    fn slave_flags(&self) -> &TtyFlags {
+        self.slave.tty_flags()
+    }
+
     fn check_io_events(&self) -> IoEvents {
         let mut events = IoEvents::empty();
 
@@ -62,7 +70,7 @@ impl PtyMaster {
             events |= IoEvents::OUT;
         }
 
-        if self.slave.driver().opened_slaves().load(Ordering::Relaxed) == 0 {
+        if self.master_flags().is_other_closed() {
             events |= IoEvents::HUP;
         }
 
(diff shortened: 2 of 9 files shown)


kernel/src/device/tty/mod.rs
```diff
@@ -106,6 +110,11 @@ impl<D> Tty<D> {
     pub fn notify_hup(&self) {
         self.pollee.notify(IoEvents::HUP);
     }
+
+    /// Returns the TTY flags.
+    pub fn tty_flags(&self) -> &TtyFlags {
```
lrh2000: This should be `pub(super)`. In fact, all methods in this file, except `fn index()`, should be `pub(super)`.

kernel/src/device/tty/mod.rs
```diff
@@ -266,6 +275,10 @@ impl<D: TtyDriver> FileIo for Tty<D> {
     }
 
     fn write(&self, reader: &mut VmReader, status_flags: StatusFlags) -> Result<usize> {
+        if self.tty_flags.is_other_closed() {
+            return_errno_with_message!(Errno::EIO, "the pty master has been closed");
```
lrh2000: This error message should be the same as line 344.
```rust

            IoctlCmd::FIONREAD => {
                if self.tty_flags.is_other_closed() {
                    return_errno_with_message!(Errno::EIO, "the TTY is closed");
                }
```

kernel/src/device/pty/master.rs
```diff
@@ -126,9 +134,31 @@ impl FileIo for PtyMaster {
...
+                } else {
+                    0
+                };
+                current_userspace!().write_val(arg, &val)?;
             }
             IoctlCmd::TIOCGPTPEER => {
+                if self.master_flags().is_pty_locked() {
+                    return_errno_with_message!(
+                        Errno::EIO,
+                        "pty slave cannot be opened because pty master is locked"
```
lrh2000: This will be checked in `PtySlaveFile::new`, so there is no need to check it here?

kernel/src/device/pty/file.rs
```diff
@@ -19,20 +17,40 @@ use crate::{
...
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L220>.
+
+        // Holds the lock at first to avoid race condition.
+        let mut opened_slaves = slave.driver().opened_slaves().lock();
+
+        let master_flags = slave.driver().tty_flags();
+        if master_flags.is_pty_locked() || slave.tty_flags().is_other_closed() {
+            return_errno_with_message!(
+                Errno::EIO,
+                "pty slave cannot be opened because pty master is locked or closed"
```
lrh2000:
```suggestion
                "a pty slave cannot be opened when the pty master is locked or closed"
```

kernel/src/device/pty/file.rs
```diff
@@ -19,20 +17,40 @@ use crate::{
...
 impl PtySlaveFile {
-    pub(super) fn new(slave: Arc<PtySlave>) -> PtySlaveFile {
-        slave
-            .driver()
-            .opened_slaves()
-            .fetch_add(1, Ordering::Relaxed);
+    pub(super) fn new(slave: Arc<PtySlave>) -> Result<PtySlaveFile> {
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L220>.
+
+        // Holds the lock at first to avoid race condition.
```
lrh2000:
```suggestion
        // Hold the lock at first to avoid race condition.
```

kernel/src/device/pty/file.rs
```diff
@@ -19,20 +17,40 @@ use crate::{
...
-        driver.opened_slaves().fetch_sub(1, Ordering::Relaxed);
+
+        let mut opened_slaves = driver.opened_slaves().lock();
+        *opened_slaves -= 1;
+
+        if *opened_slaves == 0 {
+            driver.tty_flags().set_other_closed();
+        }
+
         driver.pollee().notify(IoEvents::HUP);
```
lrh2000: This can be:
```rust

        if *opened_slaves == 0 {
            driver.tty_flags().set_other_closed();
            driver.pollee().notify(IoEvents::HUP);
        }
```
?

kernel/src/device/pty/master.rs
```diff
@@ -126,9 +134,31 @@ impl FileIo for PtyMaster {
...
+                let val = current_userspace!().read_val::<i32>(arg)?;
+                let flags = self.master_flags();
+                if val == 0 {
+                    flags.clear_pty_locked();
+                } else {
+                    flags.set_pty_locked();
+                }
+            }
+            IoctlCmd::TIOCGPTLCK => {
+                // Get pty lock state.
```
lrh2000: No need to have this comment here?

test/src/apps/pty/open_ptmx.c
```diff
@@ -0,0 +1,97 @@
...
+		exit(EXIT_SUCCESS);
+	}
+
+	// Set slave raw mode
+	struct termios term;
+	CHECK(tcgetattr(slave, &term));
+	term.c_lflag &= ~(ICANON | ECHO);
+	term.c_cc[VMIN] = 1;
+	term.c_cc[VTIME] = 0;
+	CHECK(tcsetattr(slave, TCSANOW, &term));
```
lrh2000: These should use `TEST_SUCC`?

lrh2000 (commented):
Thanks for the fix! The changes look good. Just some minor comments.

kernel/src/device/tty/line_discipline.rs
```diff
@@ -239,6 +239,14 @@ impl LineDiscipline {
 
     pub fn set_termios(&mut self, termios: CTermios) {
         self.termios = termios;
+
+        // When switching to raw mode, any pending input bytes should become immediately available.
+        // TODO: Define the correct behavior for pending bytes when switching back to canonical mode,
+        // and handle the case where the read buffer is full.
```
lrh2000: The read buffer cannot be full because of:
https://github.com/asterinas/asterinas/blob/c2d100f902a405ec055491703679d5f60d09bb2c/kernel/src/device/tty/line_discipline.rs#L232-L234

Actually we should define a `flush_line` method that performs:
https://github.com/asterinas/asterinas/blob/c2d100f902a405ec055491703679d5f60d09bb2c/kernel/src/device/tty/line_discipline.rs#L145-L149
?

lrh2000 (approved):

kernel/src/device/tty/line_discipline.rs
```diff
@@ -239,6 +237,19 @@ impl LineDiscipline {
...
+
+        // When switching to raw mode, any pending input bytes should become immediately available.
+        // Note that `unwrap()` below won't fail because we checked `is_full()` in `push_char`.
+        // TODO: Define the correct behavior for pending bytes when switching back to canonical mode.
+        if !self.termios.is_canonical_mode() {
+            self.flush_line().unwrap();
+        }
+    }
+
+    /// Flushes the bytes in current line into read buffer.
```
lrh2000: Honestly, I don't know if it's worth another push, but:
```suggestion
    /// Flushes the bytes in the current line into the read buffer.
```

PR #2582, opened 2025-11-13
Add some IOCTLs for evdev

Cautreoxit:
This PR is to add some IOCTLs for `evdev`, which are necessary for `xorg server`.

Actually the first commit is from evdev (PR https://github.com/asterinas/asterinas/pull/2561/), and the second commit is for IOCTLs. Since evdev has not been merged yet, I mark this PR as _draft_. After evdev has final version, I would modify this PR and mark it _open_.

### DIFF
--- a/kernel/comps/input/src/input_dev.rs
+++ b/kernel/comps/input/src/input_dev.rs
@@ -225,6 +225,21 @@ impl InputCapability {
     pub fn clear_supported_relative_axis(&mut self, rel_code: RelCode) {
         self.supported_relative_axes.clear(rel_code);
     }
+
+    /// Returns the supported event types as a bitmap.
+    pub fn event_types_bits(&self) -> u32 {
+        self.supported_event_types.bits()
+    }
+
+    /// Returns the supported key code bitmap as bytes.
+    pub fn supported_keys_bitmap(&self) -> &[u8] {
+        self.supported_keys.as_raw_slice()
+    }
+
+    /// Returns the supported relative axes bitmap as bytes.
+    pub fn supported_relative_axes_bitmap(&self) -> &[u8] {
+        self.supported_relative_axes.as_raw_slice()
+    }
 }
 
 pub trait InputDevice: Send + Sync + Any + Debug {
--- a/kernel/src/device/evdev/file.rs
+++ b/kernel/src/device/evdev/file.rs
@@ -27,13 +27,35 @@ use crate::{
     },
     prelude::*,
     process::signal::{PollHandle, Pollable, Pollee},
+    syscall::ClockId,
     util::{
-        ioctl::RawIoctl,
+        ioctl::{dispatch_ioctl, RawIoctl},
         ring_buffer::{RbConsumer, RbProducer, RingBuffer},
     },
 };
 
 pub(super) const EVDEV_BUFFER_SIZE: usize = 64;
+
+mod ioctl_defs {
+    use aster_input::input_dev::InputId;
+
+    use crate::util::ioctl::{ioc, InData, IoctlEnum, OutData};
+
+    // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/input.h>
+
+    pub(super) type GetDriverVer  = ioc!(EVIOCGVERSION, b'E', 0x01, OutData<i32>);
+    pub(super) type GetInputId    = ioc!(EVIOCGID,      b'E', 0x02, OutData<InputId>);
+    pub(super) type GetDeviceName = ioc!(EVIOCGNAME,    b'E', 0x06, OutData<[u8]>);
+    pub(super) type GetDevicePhys = ioc!(EVIOCGPHYS,    b'E', 0x07, OutData<[u8]>);
+    pub(super) type GetDeviceUniq = ioc!(EVIOCGUNIQ,    b'E', 0x08, OutData<[u8]>);
+    pub(super) type GetKeyState   = ioc!(EVIOCGKEY,     b'E', 0x18, OutData<[u8]>);
+    pub(super) type GetLedState   = ioc!(EVIOCGLED,     b'E', 0x19, OutData<[u8]>);
+    pub(super) type GetSwState    = ioc!(EVIOCGSW,      b'E', 0x1B, OutData<[u8]>);
+    pub(super) type SetClockId    = ioc!(EVIOCSCLOCKID, b'E', 0xA0, InData<i32>);
+
+    /// The `EVIOCGBIT` ioctl enum.
+    pub(super) type GetEventBits = IoctlEnum<b'E', 0x20, 0x1F, OutData<[u8]>>;
+}
 
 // Reference: <https://elixir.bootlin.com/linux/v6.17.9/source/include/uapi/linux/input.h#L28>
 #[repr(C)]
@@ -61,7 +83,7 @@ impl EvdevEvent {
 
 // Reference: <https://elixir.bootlin.com/linux/v6.17.9/source/drivers/input/evdev.c#L181-L189>
 #[repr(u8)]
-#[derive(Debug, Clone, Copy, TryFromInt)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
 enum EvdevClock {
     Realtime = 0,
     Monotonic = 1,
@@ -78,6 +100,22 @@ define_atomic_version_of_integer_like_type!(EvdevClock, try_from = true, {
     #[derive(Debug)]
     struct AtomicEvdevClock(AtomicU8);
 });
+
+impl TryFrom<ClockId> for EvdevClock {
+    type Error = Error;
+
+    fn try_from(clock_id: ClockId) -> Result<Self> {
+        match clock_id {
+            ClockId::CLOCK_REALTIME => Ok(EvdevClock::Realtime),
+            ClockId::CLOCK_MONOTONIC => Ok(EvdevClock::Monotonic),
+            ClockId::CLOCK_BOOTTIME => Ok(EvdevClock::Boottime),
+            _ => return_errno_with_message!(
+                Errno::EINVAL,
+                "the clock is not a valid clock for evdev files"
+            ),
+        }
+    }
+}
 
 /// An opened file from an evdev device ([`EvdevDevice`]).
 pub(super) struct EvdevFile {
@@ -129,6 +167,29 @@ impl EvdevFileInner {
         consumer.clear();
         self.packet_count.store(0, Ordering::Relaxed);
         self.pollee.invalidate();
+    }
+
+    pub(self) fn clear_with_consumer_producer_locked(
+        &self,
+        consumer: &mut RbConsumer<EvdevEvent>,
+        producer: &mut RbProducer<EvdevEvent>,
+    ) {
+        // Note that the following operations are race-free because both the consumer and the
+        // producer locks have been held.
+
+        if consumer.is_empty() {
+            return;
+        }
+
+        consumer.clear();
+
+        let time = self.read_clock();
+        let dropped_event =
+            EvdevEvent::from_event_and_time(&InputEvent::from_sync_event(SynEvent::Dropped), time);
+        producer.push(dropped_event).unwrap();
+
+        self.packet_count.store(1, Ordering::Relaxed);
+        self.pollee.notify(IoEvents::IN);
     }
 
     /// Checks if buffer has complete event packets.
@@ -214,6 +275,30 @@ impl EvdevFile {
             IoEvents::empty()
         }
     }
+
+    fn upgrade_device(&self) -> Result<Arc<EvdevDevice>> {
+        self.evdev.upgrade().ok_or_else(|| {
+            Error::with_message(Errno::ENODEV, "the evdev device has been disconnected")
+        })
+    }
+
+    fn set_clock(&self, clock: EvdevClock) -> Result<()> {
+        // Lock the mutex in advance to avoid race conditions.
+        let mut consumer = self.inner.consumer.lock();
+
+        if self.inner.clock_id.load(Ordering::Relaxed) == clock {
+            return Ok(());
+        }
+
+        let device = self.upgrade_device()?;
+        device.with_producer_locked(&self.inner, |producer| {
+            self.inner.clock_id.store(clock, Ordering::Relaxed);
+            self.inner
+                .clear_with_consumer_producer_locked(&mut consumer, producer);
+        });
+
+        Ok(())
+    }
 }
 
 /// Checks if the event is a `SYN_REPORT` event.
@@ -224,6 +309,36 @@ pub(super) fn is_syn_report_event(event: &EvdevEvent) -> bool {
 /// Checks if the event is a `SYN_DROPPED` event.
 pub(super) fn is_syn_dropped_event(event: &EvdevEvent) -> bool {
     event.type_ == EventTypes::SYN.as_index() && event.code == SynEvent::Dropped as u16
+}
+
+fn write_bytes_and_zeros_to_userspace(writer: &mut VmWriter, bytes: &[u8]) -> Result<()> {
+    let mut reader = VmReader::from(bytes);
+    writer.write_fallible(&mut reader)?;
+    writer.fill_zeros(writer.avail())?;
+    Ok(())
+}
+
+fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWriter) -> Result<()> {
+    let capability = evdev.device.capability();
+
+    match event_type as u16 {
+        0 => {
+            let event_types_bytes = capability.event_types_bits().to_ne_bytes();
+            write_bytes_and_zeros_to_userspace(writer, &event_types_bytes)?;
+        }
+        t if t == EventTypes::KEY.as_index() => {
+            let bitmap = capability.supported_keys_bitmap();
+            write_bytes_and_zeros_to_userspace(writer, bitmap)?;
+        }
+        t if t == EventTypes::REL.as_index() => {
+            let bitmap = capability.supported_relative_axes_bitmap();
+            write_bytes_and_zeros_to_userspace(writer, bitmap)?;
+        }
+        _ => {
+            return_errno_with_message!(Errno::EINVAL, "the event type is not supported yet");
+        }
+    }
+    Ok(())
 }
 
 impl Pollable for EvdevFile {
@@ -292,17 +407,88 @@ impl FileIo for EvdevFile {
         false
     }
 
-    fn ioctl(&self, _raw_ioctl: RawIoctl) -> Result<i32> {
-        // TODO: Support ioctl operations for evdev files.
+    fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
+        use ioctl_defs::*;
 
-        // Most ioctl implementations return `ENOTTY` for invalid ioctl commands, representing "The
-        // specified operation does not apply". However, according to the Linux implementation,
-        // evdev files return `EINVAL` in this case.
-        // Reference: <https://elixir.bootlin.com/linux/v6.17.8/source/drivers/input/evdev.c#L1251>
-        return_errno_with_message!(
-            Errno::EINVAL,
-            "the ioctl command is not supported by evdev files"
-        );
+        dispatch_ioctl!(match raw_ioctl {
+            cmd @ GetDriverVer => {
+                const EVDEV_DRIVER_VERSION: i32 = 0x010001;
+                cmd.write(&EVDEV_DRIVER_VERSION)?;
+            }
+            cmd @ GetInputId => {
+                let evdev = self.upgrade_device()?;
+                let input_id = evdev.device.id();
+                cmd.write(&input_id)?;
+            }
+            cmd @ SetClockId => {
+                let clock_id_raw: i32 = cmd.read()?;
+                let clock_id = ClockId::try_from(clock_id_raw)?;
+                let evdev_clock = EvdevClock::try_from(clock_id)?;
+                self.set_clock(evdev_clock)?;
+            }
+            cmd @ GetDeviceName => {
+                let evdev = self.upgrade_device()?;
+                cmd.with_writer(|mut writer| {
+                    write_bytes_and_zeros_to_userspace(&mut writer, evdev.device.name().as_bytes())
+                })?;
+            }
+            cmd @ GetDevicePhys => {
+                let evdev = self.upgrade_device()?;
+                cmd.with_writer(|mut writer| {
+                    write_bytes_and_zeros_to_userspace(&mut writer, evdev.device.phys().as_bytes())
+                })?;
+            }
+            cmd @ GetDeviceUniq => {
+                let evdev = self.upgrade_device()?;
+                cmd.with_writer(|mut writer| {
+                    write_bytes_and_zeros_to_userspace(&mut writer, evdev.device.uniq().as_bytes())
+                })?;
+            }
+            cmd @ GetEventBits => {
+                let evdev = self.upgrade_device()?;
+                let event_type = cmd.discriminant();
+                cmd.base_ioctl()
+                    .with_writer(|mut writer| handle_get_bit(&evdev, event_type, &mut writer))?;
+            }
+            cmd @ GetKeyState => {
+                // TODO: We need to track whether the key is currently pressed and report that state
+                // here. If we report states, we need to flush the queue to avoid interfering with the
+                // user space's state tracking. See the Linux implementation at:
+                // <https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L872-L876>.
+                //
+                // Currently, no key state tracking is supported. So we report zeros as key states here.
+                cmd.with_writer(|mut writer| {
+                    writer.fill_zeros(writer.avail())?;
+                    Ok(())
+                })?;
+            }
+            cmd @ GetLedState => {
+                // No LED events are supported. So we can report zeros as LED states here.
+                cmd.with_writer(|mut writer| {
+                    writer.fill_zeros(writer.avail())?;
+                    Ok(())
+                })?;
+            }
+            cmd @ GetSwState => {
+                // No switch events are supported. So we can report zeros as switch states here.
+                cmd.with_writer(|mut writer| {
+                    writer.fill_zeros(writer.avail())?;
+                    Ok(())
+                })?;
+            }
+            _ => {
+                // Most ioctl implementations return `ENOTTY` for invalid ioctl commands, representing "The
+                // specified operation does not apply". However, according to the Linux implementation,
+                // evdev files return `EINVAL` in this case.
+                // Reference: <https://elixir.bootlin.com/linux/v6.17.8/source/drivers/input/evdev.c#L1251>
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "the ioctl command is not supported by evdev files"
+                );
+            }
+        });
+
+        Ok(0)
     }
 }
 
--- a/kernel/comps/input/src/event_type_codes.rs
+++ b/kernel/comps/input/src/event_type_codes.rs
@@ -151,6 +151,11 @@ impl RelCodeSet {
         let index = rel_code as usize;
         self.0.get(index).map(|b| *b).unwrap()
     }
+
+    /// Returns the bitmap as a byte slice.
+    pub fn as_raw_slice(&self) -> &[u8] {
+        self.0.as_raw_slice()
+    }
 }
 
 /// A set of [`KeyCode`] represented as a bitmap.
@@ -195,6 +200,11 @@ impl KeyCodeSet {
     pub fn contain_any(&self, range: core::ops::Range<usize>) -> bool {
         assert!(range.is_empty() || range.end <= KEY_COUNT);
         range.into_iter().any(|i| *self.0.get(i).unwrap())
+    }
+
+    /// Returns the bitmap as a byte slice.
+    pub fn as_raw_slice(&self) -> &[u8] {
+        self.0.as_raw_slice()
     }
 }
 
--- a/kernel/src/device/evdev/mod.rs
+++ b/kernel/src/device/evdev/mod.rs
@@ -108,6 +108,18 @@ impl EvdevDevice {
         opened_files.swap_remove(pos);
     }
 
+    pub(self) fn with_producer_locked<F>(&self, file: &Arc<EvdevFileInner>, f: F)
+    where
+        F: FnOnce(&mut RbProducer<EvdevEvent>),
+    {
+        let mut opened_files = self.opened_files.disable_irq().lock();
+        let pos = opened_files
+            .iter()
+            .position(|(f, _)| Arc::ptr_eq(f, file))
+            .unwrap();
+        f(&mut opened_files[pos].1)
+    }
+
     /// Distributes events to all opened evdev files.
     fn pass_events(&self, events: &[InputEvent]) {
         // No need to disable IRQs because this method can only be called in the interrupt context.
--- a/kernel/src/util/ioctl/mod.rs
+++ b/kernel/src/util/ioctl/mod.rs
@@ -364,7 +364,6 @@ impl<const MAGIC: u8, const NR: u8, const IS_MODERN: bool>
     /// Obtains a [`VmWriter`] that can write the dynamically-sized ioctl argument to userspace.
     ///
     /// The size of the ioctl argument is specified in [`VmWriter::avail`].
-    #[expect(dead_code)]
     pub fn with_writer<F, R>(&self, f: F) -> Result<R>
     where
         F: for<'a> FnOnce(VmWriter<'a>) -> Result<R>,


Cautreoxit:
The reason for the CI test failure is that the kernel API used by the evdev module has changed in the past few days. Simply updating the evdev module will resolve the issue, which has already been addressed in the evdev PR. These errors are unrelated to ioctl itself.

Cautreoxit:
Since evdev PR (https://github.com/asterinas/asterinas/pull/2561/) has been merged, I modified this PR and made it ready for review.

kernel/src/device/evdev/file.rs
```diff
@@ -33,6 +35,33 @@ use crate::{
 
 pub(super) const EVDEV_BUFFER_SIZE: usize = 64;
 
+/// Linux evdev driver version returned by `EVIOCGVERSION`.
+const EVDEV_DRIVER_VERSION: i32 = 0x010001;
```
tatetian: Move this const definition to its only user:

```rust
            Some(EvdevIoctl::GetVersion) => {
                const EVDEV_DRIVER_VERSION: i32 = 0x010001;

                current_userspace!().write_val(arg, &EVDEV_DRIVER_VERSION)?;
            }
```

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+            },
+            _ => None,
+        }
+    }
+
+    fn handle_evdev_ioctl(&self, raw: u32, arg: usize) -> Result<()> {
+        match Self::parse_evdev(raw) {
+            Some(EvdevIoctl::GetName { len }) => {
+                let evdev = self.upgrade_evdev_device()?;
+                self.write_string_to_userspace(evdev.device.name(), len as usize, arg)?;
```
tatetian: The `write_string_to_userspace` and `write_bitmap_to_userspace` methods should NOT have been made as methods of `EvdevFile`. A method accesses `self`. If it does not need `self`, then it should not be a method. So turn the two methods into functions.

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+            .upgrade()
+            .ok_or_else(|| Error::with_message(Errno::ENODEV, "evdev device is unavailable"))
+    }
+
+    fn write_string_to_userspace(&self, value: &str, len: usize, user_ptr: usize) -> Result<()> {
+        if len == 0 {
+            return Ok(());
+        }
+
+        let mut buffer = vec![0u8; len];
```
tatetian: Optimize this function to avoid heap allocation and an extra data copy.

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+        current_userspace!().write_bytes(user_ptr, &mut reader)?;
+        Ok(())
+    }
+
+    fn write_bitmap_to_userspace(&self, bitmap: &[u8], len: usize, user_ptr: usize) -> Result<()> {
+        if len == 0 {
+            return Ok(());
+        }
+
+        let mut buffer = vec![0u8; len];
```
tatetian: Optimize this function to avoid heap allocation and an extra data copy.

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+        const EVIOCGBIT_BASE_NR: u32 = 0x20;
+        const EVIOCGKEY_NR: u32 = 0x18;
+        const EVIOCGLED_NR: u32 = 0x19;
+        const EVIOCGSW_NR: u32 = 0x1b;
+        const EVIOCSCLOCKID_NR: u32 = 0xa0;
+
+        let dir = (raw >> IOC_DIRSHIFT) & IOC_DIRMASK;
+        let type_ = (raw >> IOC_TYPESHIFT) & IOC_TYPEMASK;
+        let nr = (raw >> IOC_NRSHIFT) & IOC_NRMASK;
+        let len = (raw >> IOC_SIZESHIFT) & IOC_SIZEMASK;
```
tatetian: This is Linux's general encoding for ioctl command, right? To improve readability, separate the general encoding/decoding scheme from the decoding logic specific to evdev. In practice, this means that every const values whose names start with `IOC_` should not be entangled with those whose names start with `EVIOC`.

I instructed LLM to implement such a helper type. Take it as a reference. Revise it as you see fit.

```rust
/// Raw 32-bit ioctl command word (Linux-style encoding).
///
/// Layout (from LSB to MSB):
/// - bits 0..=7   : command number (nr)
/// - bits 8..=15  : type / magic
/// - bits 16..=29 : size (in bytes)
/// - bits 30..=31 : direction
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub struct RawIoctl(u32);

/// Direction of data transfer for an ioctl command.
///
/// The meaning is from the *user-space* perspective,
/// following Linux's `_IOC_*` macros:
///
/// - `None`  -> `_IOC_NONE`
/// - `Write` -> `_IOC_WRITE` (user → kernel)
/// - `Read`  -> `_IOC_READ`  (kernel → user)
/// - `ReadWrite` -> `_IOC_READ | _IOC_WRITE`
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub enum IoctlDir {
    None,
    Write,
    Read,
    ReadWrite,
    /// Any unrecognized (non-standard) bit pattern.
    Other(u8),
}

impl IoctlDir {
    fn from_raw(bits: u8) -> Self {
        match bits {
            0 => IoctlDir::None,
            1 => IoctlDir::Write,
            2 => IoctlDir::Read,
            3 => IoctlDir::ReadWrite,
            other => IoctlDir::Other(other),
        }
    }

    fn to_raw(self) -> u8 {
        match self {
            IoctlDir::None => 0,
            IoctlDir::Write => 1,
            IoctlDir::Read => 2,
            IoctlDir::ReadWrite => 3,
            IoctlDir::Other(v) => v,
        }
    }
}

// Bit widths (match asm-generic/ioctl.h on Linux)
const IOC_NRBITS: u32 = 8;
const IOC_TYPEBITS: u32 = 8;
const IOC_SIZEBITS: u32 = 14;
const IOC_DIRBITS: u32 = 2;

// Shifts
const IOC_NRSHIFT: u32 = 0;
const IOC_TYPESHIFT: u32 = IOC_NRSHIFT + IOC_NRBITS;
const IOC_SIZESHIFT: u32 = IOC_TYPESHIFT + IOC_TYPEBITS;
const IOC_DIRSHIFT: u32 = IOC_SIZESHIFT + IOC_SIZEBITS;

// Masks for each field (before shifting)
const IOC_NRMASK: u32 = (1 << IOC_NRBITS) - 1;
const IOC_TYPEMASK: u32 = (1 << IOC_TYPEBITS) - 1;
const IOC_SIZEMASK: u32 = (1 << IOC_SIZEBITS) - 1;
const IOC_DIRMASK: u32 = (1 << IOC_DIRBITS) - 1;

impl TryFrom<u32> for RawIoctl {
    type Error = Infallible;

    #[inline]
    fn try_from(value: u32) -> Result<Self, Self::Error> {
        // Any u32 can be interpreted as a raw ioctl command,
        // so this conversion is infallible.
        Ok(RawIoctl(value))
    }
}

impl From<RawIoctl> for u32 {
    #[inline]
    fn from(cmd: RawIoctl) -> u32 {
        cmd.0
    }
}

impl RawIoctl {
    /// Create a new raw ioctl command from its components,
    /// following Linux's `_IOC` layout.
    ///
    /// # Panics
    ///
    /// Panics if `size` does not fit into 14 bits (i.e., >= 16,384).
    pub fn new(dir: IoctlDir, ty: u8, nr: u8, size: u16) -> Self {
        let size_u32 = size as u32;
        assert!(
            size_u32 <= IOC_SIZEMASK,
            "ioctl size too large to encode in 14 bits"
        );

        let mut raw: u32 = 0;
        raw |= (u32::from(nr)       & IOC_NRMASK)   << IOC_NRSHIFT;
        raw |= (u32::from(ty)       & IOC_TYPEMASK) << IOC_TYPESHIFT;
        raw |= (size_u32            & IOC_SIZEMASK) << IOC_SIZESHIFT;
        raw |= (u32::from(dir.to_raw()) & IOC_DIRMASK) << IOC_DIRSHIFT;

        RawIoctl(raw)
    }

    /// Return the underlying 32-bit value.
    #[inline]
    pub fn raw(&self) -> u32 {
        self.0
    }

    /// Direction part (`_IOC_DIR(cmd)`).
    #[inline]
    pub fn dir(&self) -> IoctlDir {
        let bits = ((self.0 >> IOC_DIRSHIFT) & IOC_DIRMASK) as u8;
        IoctlDir::from_raw(bits)
    }

    /// Size part in bytes (`_IOC_SIZE(cmd)`).
    ///
    /// Note: this is just the encoded size field; some legacy ioctl
    /// commands may have incorrect size encodings, just like in C.
    #[inline]
    pub fn size(&self) -> u16 {
        ((self.0 >> IOC_SIZESHIFT) & IOC_SIZEMASK) as u16
    }

    /// Type / magic part (`_IOC_TYPE(cmd)`).
    ///
    /// Often this is an ASCII character like `b'r'` or `b'M'`.
    #[inline]
    pub fn ty(&self) -> u8 {
        ((self.0 >> IOC_TYPESHIFT) & IOC_TYPEMASK) as u8
    }

    /// Command number part (`_IOC_NR(cmd)`).
    #[inline]
    pub fn nr(&self) -> u8 {
        ((self.0 >> IOC_NRSHIFT) & IOC_NRMASK) as u8
    }
}
```

This helper struct should be extracted out of the evdev module, into a common location.

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+                let evdev = self.upgrade_evdev_device()?;
+                self.write_string_to_userspace(evdev.device.phys(), len as usize, arg)?;
+            }
+            Some(EvdevIoctl::GetUniq { len }) => {
+                let evdev = self.upgrade_evdev_device()?;
+                self.write_string_to_userspace(evdev.device.uniq(), len as usize, arg)?;
+            }
+            Some(EvdevIoctl::GetId) => {
+                let evdev = self.upgrade_evdev_device()?;
+                let id = evdev.device.id();
```
tatetian: Please rename this variable to `input_id: InputId` to distinguish it from `DeviceId`. I thought that `device.id()` returns `DeviceId` at the first glance.

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+            }
+            Some(EvdevIoctl::GetVersion) => {
+                current_userspace!().write_val(arg, &EVDEV_DRIVER_VERSION)?;
+            }
+            Some(EvdevIoctl::GetBit { event_type, len }) => {
+                let evdev = self.upgrade_evdev_device()?;
+                let capability = evdev.device.capability();
+                let event_types_bytes = capability.event_types_bits().to_le_bytes();
+                let bitmap = match event_type as u16 {
+                    0 => Some(&event_types_bytes[..]),
```
tatetian: The line `let event_types_bytes = capability.event_types_bits().to_le_bytes();` should be moved into the first branch because only this branch makes use of the variable.

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+                    t if t == EventTypes::KEY.as_index() => {
+                        Some(capability.supported_keys_bitmap())
+                    }
+                    t if t == EventTypes::REL.as_index() => {
+                        Some(capability.supported_relative_axes_bitmap())
+                    }
+                    _ => None,
```
tatetian: If `t` is not a recognized type, shouldn't we return an error?

kernel/src/device/evdev/file.rs
```diff
@@ -185,6 +214,192 @@ impl EvdevFile {
...
+                EVIOCSCLOCKID_NR => Some(EvdevIoctl::SetClockId),
+                _ => None,
+            },
+            _ => None,
+        }
+    }
+
+    fn handle_evdev_ioctl(&self, raw: u32, arg: usize) -> Result<()> {
+        match Self::parse_evdev(raw) {
+            Some(EvdevIoctl::GetName { len }) => {
```
tatetian: A more systematic approach is needed to define and handle ioctls.

Linux provides a standard approach to define the properties of ioctl commands. Here is an example.

```c
#define EVIOCGVERSION   _IOR('E', 0x01, int)			/* get driver version */
#define EVIOCGID		_IOR('E', 0x02, struct input_id)	/* get device ID */
#define EVIOCSCLOCKID   _IOW('E', 0xa0, int)            /* Set clockid to be used for timestamps */
#define EVIOCGNAME(len) _IOC(_IOC_READ, 'E', 0x06, len)		/* get device name */
```

I propose a counterpart in Asterinas. A magical `define_ioctl_cmd` macro can define ioctl commands of various characteristics as new types.

```rust
/// Get API version.
define_ioctl_cmd!(IoctlGetVersion, 	    'E', 0x01, IoctlDirection::Read, i32);
/// Get input device ID.
define_ioctl_cmd!(IoctlGetId, 		    'E', 0x02, IoctlDirection::Read, InputId);
/// Set clock ID.
define_ioctl_cmd!(IoctlSetClockId,      'E', 0xa0, IoctlDirection::Write, u32 /* ClockId */);
/// Get device name.
define_ioctl_cmd!(IoctlGetDeviceName,   'E', 0x06, IoctlDirection::Write, [u8]);
```

An ioctl type include information about the number, direction, data size (including fixed or variable length), and data type about the ioctl at the compile time. The generated ioctl types provide convenient methods to handle the ioctl correctly. With these ioctl types, the ioctl handling logic can be much simplified:

```rust
impl EvdevFile
    fn handle_evdev_ioctl(&self, raw: u32, arg: usize) -> Result<()> {
        if let Ok(ioctl) = IoctlGetVersion::try_from(raw, arg) {
            let version = todo!("get version");
            ioctl.write_val_to_userspace(version)?
        } else if Ok(ioctl) = IoctlGetId::try_from(raw, arg) {
            let input_id = todo!("get input device ID");
            ioctl.write_val_to_userspace(input_id);
        } else if Ok(ioctl) = IoctlSetClockId::try_from(raw, arg) {
            let raw_clock_id = ioctl.read_val_from_userspace()?;
            let clock_id = ClockId::from_raw(raw_clock_id)?;
            todo!("update clock ID")
        } else if Ok(ioctl) = IoctlGetDeviceName::try_from(raw, arg) {
            let mut vm_reader = todo!("make a vm reader out of the internal device name representation");
            ioctl.write_bytes_to_userspace(&mut vm_reader)?;
        } else {
            todo!("return error")
        }
    }
}
```

I believe you can figure out the details about the generated ioctl types. These ioctl types can make use of the `RawIoctlCmd` proposed before.

Cautreoxit:
Thanks for your comments! It is a really good idea to make a formal paradigm for ioctls. I am not quite sure if I understood correctly, please check my modifies. I would appreciate further comments, thank you very much!

Some details:
1. I have also addressed those small issues. 
2. I think so many `try_from(raw, arg)` looks bad:
```
impl EvdevFile
    fn handle_evdev_ioctl(&self, raw: u32, arg: usize) -> Result<()> {
        if let Ok(ioctl) = IoctlGetVersion::try_from(raw, arg) {
            let version = todo!("get version");
            ioctl.write_val_to_userspace(version)?
        } else if Ok(ioctl) = IoctlGetId::try_from(raw, arg) {
            let input_id = todo!("get input device ID");
            ioctl.write_val_to_userspace(input_id);
        } else if Ok(ioctl) = IoctlSetClockId::try_from(raw, arg) {
            let raw_clock_id = ioctl.read_val_from_userspace()?;
            let clock_id = ClockId::from_raw(raw_clock_id)?;
            todo!("update clock ID")
        } else if Ok(ioctl) = IoctlGetDeviceName::try_from(raw, arg) {
            let mut vm_reader = todo!("make a vm reader out of the internal device name representation");
            ioctl.write_bytes_to_userspace(&mut vm_reader)?;
        } else {
            todo!("return error")
        }
    }
}
```
so I add `struct IoctlRequest` to wrap them.
3. Honestly, I am not quite familiar with Rust's macro, so I implement `macro_rules! define_ioctl_cmd` with LLM's help. Please check if it has any problems. Thanks!

tatetian:
@Cautreoxit The revised ioctl is still not quite satisfying. Check out #2649 for a complete design. It addresses some limitations of the ioctl infra in this PR.

kernel/comps/input/src/input_dev.rs
```diff
@@ -225,6 +225,21 @@ impl InputCapability {
     pub fn clear_supported_relative_axis(&mut self, rel_code: RelCode) {
         self.supported_relative_axes.clear(rel_code);
     }
+
+    /// Returns the supported event types as a bitmap.
+    pub fn event_types_bits(&self) -> u32 {
+        self.supported_event_types.bits()
+    }
+
+    /// Returns the supported key codes bitmap.
```
lrh2000:
```suggestion
    /// Returns the supported key code bitmap.
```

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+            Ok(())
+        } else if let Ok(ioctl) = GetId::try_from(request) {
+            let evdev = self.upgrade_evdev_device()?;
+            let input_id = evdev.device.id();
+            write_val_to_userspace(ioctl.user_ptr(), input_id)?;
+            Ok(())
+        } else if let Ok(ioctl) = SetClockId::try_from(request) {
+            let clock_id_raw: i32 = read_val_from_userspace(ioctl.user_ptr())?;
+            let clock_id = ClockId::try_from(clock_id_raw)
+                .map_err(|_| Error::with_message(Errno::EINVAL, "invalid clock id"))?;
```
lrh2000: `ClockId::try_from(clock_id_raw)?` should be okay.

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+            let evdev_clock = match clock_id {
+                ClockId::CLOCK_REALTIME => EvdevClock::Realtime,
+                ClockId::CLOCK_MONOTONIC => EvdevClock::Monotonic,
+                ClockId::CLOCK_BOOTTIME => EvdevClock::Boottime,
+                _ => {
+                    return_errno_with_message!(Errno::EINVAL, "unsupported clock id");
```
lrh2000:
```suggestion
                    return_errno_with_message!(Errno::EINVAL, "the clock is not a valid clock for evdev files");
```

Also, it is better to add `impl TryFrom<ClockId> for EvdevClock`.

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+                }
+            };
+            self.inner.clock_id.store(evdev_clock, Ordering::Relaxed);
```
lrh2000: A logic to flush the queue is missed here:
https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L194-L209

Honestly, I don't really know what's the best way to implement this... Something like this should work but it is quite inelegant:
```patch
diff --git a/kernel/src/device/evdev/file.rs b/kernel/src/device/evdev/file.rs
index 534680f2c..6511909d7 100644
--- a/kernel/src/device/evdev/file.rs
+++ b/kernel/src/device/evdev/file.rs
@@ -131,6 +131,28 @@ impl EvdevFileInner {
         self.pollee.invalidate();
     }
 
+    pub(self) fn clear_with_consumer_producer_locked(
+        &self,
+        consumer: &mut RbConsumer<EvdevEvent>,
+        producer: &mut RbProducer<EvdevEvent>,
+    ) {
+        // Note that the following operations are race-free because both the consumer and the
+        // producer locks have been held.
+
+        if consumer.len() == 0 {
+            return;
+        }
+
+        consumer.clear();
+
+        let dropped_event =
+            EvdevEvent::from_event_and_time(&InputEvent::from_sync_event(SynEvent::Dropped), time);
+        producer.push(dropped_event);
+
+        self.packet_count.store(1, Ordering::Relaxed);
+        self.pollee.notify(IoEvents::IN);
+    }
+
     /// Checks if buffer has complete event packets.
     pub(self) fn has_complete_packets(&self) -> bool {
         self.packet_count.load(Ordering::Relaxed) > 0
@@ -214,6 +236,30 @@ impl EvdevFile {
             IoEvents::empty()
         }
     }
+
+    fn upgrade_device(&self) -> Result<Arc<EvdevDevice>> {
+        self.evdev.upgrade().ok_or_else(|| {
+            Error::with_message(Errno::ENODEV, "the evdev device has been disconnected")
+        })
+    }
+
+    fn set_clock(&self, clock: EvdevClock) -> Result<()> {
+        // Lock the mutex in advance to avoid race conditions.
+        let mut consumer = self.inner.consumer.lock();
+
+        if self.inner.clock_id.load(Ordering::Relaxed) == clock {
+            return Ok(());
+        }
+
+        let device = self.upgrade_device()?;
+        device.with_producer_locked(&self.inner, |producer| {
+            self.inner.clock_id.store(clock, Ordering::Relaxed);
+            self.inner
+                .clear_with_consumer_producer_locked(&mut consumer, producer);
+        });
+
+        Ok(())
+    }
 }
 
 /// Checks if the event is a `SYN_REPORT` event.
diff --git a/kernel/src/device/evdev/mod.rs b/kernel/src/device/evdev/mod.rs
index cd2319cfb..6ef8e53a5 100644
--- a/kernel/src/device/evdev/mod.rs
+++ b/kernel/src/device/evdev/mod.rs
@@ -108,6 +108,18 @@ impl EvdevDevice {
         opened_files.swap_remove(pos);
     }
 
+    pub(self) fn with_producer_locked<F>(&self, file: &Arc<EvdevFileInner>, f: F)
+    where
+        F: FnOnce(&mut RbProducer<EvdevEvent>),
+    {
+        let mut opened_files = self.opened_files.disable_irq().lock();
+        let pos = opened_files
+            .iter()
+            .position(|(f, _)| Arc::ptr_eq(f, file))
+            .unwrap();
+        f(&mut opened_files[pos].1)
+    }
+
     /// Distributes events to all opened evdev files.
     fn pass_events(&self, events: &[InputEvent]) {
         // No need to disable IRQs because this method can only be called in the interrupt context.
```

I am curious if there is a better way...

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+                        ioctl.user_ptr(),
+                        &event_types_bytes,
+                        ioctl.buffer_len(),
+                    )?;
+                    return Ok(());
+                }
+                t if t == EventTypes::KEY.as_index() => capability.supported_keys_bitmap(),
+                t if t == EventTypes::REL.as_index() => capability.supported_relative_axes_bitmap(),
+                _ => {
+                    return_errno_with_message!(Errno::EINVAL, "unsupported event type");
```
lrh2000:
```rust
                    return_errno_with_message!(Errno::EINVAL, "the event type is not supported yet");
```

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+                }
+            };
+            write_bitmap_to_userspace(ioctl.user_ptr(), bitmap, ioctl.buffer_len())?;
+            Ok(())
+        } else if let Ok(ioctl) = GetKeyState::try_from(request) {
+            let zero = vec![0u8; ioctl.buffer_len()];
+            write_bitmap_to_userspace(ioctl.user_ptr(), &zero[..], ioctl.buffer_len())?;
+            Ok(())
```
lrh2000: This is really just incorrect. This needs to return whether the key is currently pressed (and deal with the queue stuffs: https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L872-L876).

At least, this needs a detailed TODO comment to explain the Linux behavior:
```rust
            // TODO: We need to track whether the key is currently pressed and report that state
            // here. If we report states, we need to flush the queue to avoid interfering with the
            // user space's state tracking. See the Linux implementation at:
            // <https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L872-L876>.
```

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+        } else if let Ok(ioctl) = GetLedState::try_from(request) {
+            let zero = vec![0u8; ioctl.buffer_len()];
```
lrh2000: Use `fill_zeros` here should be fine?

kernel/src/device/evdev/file.rs
```diff
@@ -223,6 +423,57 @@ pub(super) fn is_syn_dropped_event(event: &EvdevEvent) -> bool {
...
+        let mut reader = VmReader::from(&bytes[..copy_len]);
+        current_userspace!().write_bytes(user_ptr, &mut reader)?;
+    }
+
+    let remaining = len - copy_len;
+    if remaining > 0 {
+        current_userspace!()
+            .writer(user_ptr + copy_len, remaining)?
+            .fill_zeros(remaining)?;
+    }
```
lrh2000: This is unnecessarily complex. You can construct a writer (the new ioctl API should also provide you with one) and do something like this:
```rust
let writer = todo!();
writer.write_fallible(&mut VmReader::from(str.as_bytes()))?;
writer.fill_zeros(writer.remain())?;
```

There is no need to compute the length (i.e., `copy_len`/`remaining`) manually.

kernel/src/device/evdev/file.rs
```diff
@@ -289,17 +540,15 @@ impl FileIo for EvdevFile {
...
-        // TODO: Support ioctl operations for evdev files.
-
-        // Most ioctl implementations return `ENOTTY` for invalid ioctl commands, representing "The
-        // specified operation does not apply". However, according to the Linux implementation,
-        // evdev files return `EINVAL` in this case.
-        // Reference: <https://elixir.bootlin.com/linux/v6.17.8/source/drivers/input/evdev.c#L1251>
-        return_errno_with_message!(
-            Errno::EINVAL,
-            "the ioctl command is not supported by evdev files"
-        );
```
lrh2000, line 305: Please keep the comments (and the error message).

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
             IoEvents::empty()
         }
     }
+
+    fn upgrade_evdev_device(&self) -> Result<Arc<EvdevDevice>> {
+        self.evdev
+            .upgrade()
+            .ok_or_else(|| Error::with_message(Errno::ENODEV, "evdev device is unavailable"))
```
lrh2000:
```suggestion
            .ok_or_else(|| Error::with_message(Errno::ENODEV, "the evdev device has been disconnected"))
```

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
             IoEvents::empty()
         }
     }
+
+    fn upgrade_evdev_device(&self) -> Result<Arc<EvdevDevice>> {
```
lrh2000: No need to repeat the `evdev` word in method names. Use `upgrade_device` & `handle_ioctl` instead.

kernel/src/device/evdev/file.rs
```diff
@@ -211,6 +326,91 @@ impl EvdevFile {
...
+            };
+            write_bitmap_to_userspace(ioctl.user_ptr(), bitmap, ioctl.buffer_len())?;
+            Ok(())
+        } else if let Ok(ioctl) = GetKeyState::try_from(request) {
+            let zero = vec![0u8; ioctl.buffer_len()];
+            write_bitmap_to_userspace(ioctl.user_ptr(), &zero[..], ioctl.buffer_len())?;
+            Ok(())
+        } else if let Ok(ioctl) = GetLedState::try_from(request) {
+            let zero = vec![0u8; ioctl.buffer_len()];
+            write_bitmap_to_userspace(ioctl.user_ptr(), &zero[..], ioctl.buffer_len())?;
```
lrh2000:
```rust
            // No LED events are supported. So we can report zeros as LED states here.
```

Cautreoxit:
Thanks for the thoughtful review! I’ve rebased this PR and addressed the comments above. I also responded to the two issues individually. If anything still looks off, please let me know. Thanks!

kernel/src/device/evdev/file.rs
Cautreoxit: Yes. I haven’t actually implemented `GetKeyState` yet because it would require tracking the state of all keys, which we don’t currently support. Our target application, `Xorg` (via `libevdev`), does not use the information returned by `GetKeyState` (or `GetLedState` and `GetSwState`). So I left a dummy implementation in place and I have added a TODO comment here.

kernel/src/device/evdev/file.rs
Cautreoxit: Yes, you're right — the current implementation indeed misses updating the event timestamp. The fix is admittedly a little inelegant, but I think it is the best approach for this situation. I’m using this solution for now.

kernel/comps/input/src/input_dev.rs
```diff
@@ -225,6 +225,21 @@ impl InputCapability {
...
+        self.supported_event_types.bits()
+    }
+
+    /// Returns the supported key code bitmap.
+    pub fn supported_keys_bitmap(&self) -> &[u8] {
+        self.supported_keys.as_raw_slice()
+    }
+
+    /// Returns the supported relative axes bitmap as bytes.
+    pub fn supported_relative_axes_bitmap(&self) -> &[u8] {
```
lrh2000: So this is more consistent?
```patch
-    /// Returns the supported key code bitmap.
+    /// Returns the supported key code bitmap as bytes.
```

kernel/src/device/evdev/file.rs
```diff
@@ -27,14 +28,35 @@ use crate::{
...
+    pub(super) type GetVersion    = ioc!(EVIOCGVERSION, b'E', 0x01, OutData<i32>);
+    pub(super) type GetId         = ioc!(EVIOCGID,      b'E', 0x02, OutData<InputId>);
+    pub(super) type GetDeviceName = ioc!(EVIOCGNAME,    b'E', 0x06, OutData<[u8]>);
+    pub(super) type GetPhys       = ioc!(EVIOCGPHYS,    b'E', 0x07, OutData<[u8]>);
+    pub(super) type GetUniq       = ioc!(EVIOCGUNIQ,    b'E', 0x08, OutData<[u8]>);
+    pub(super) type GetKeyState   = ioc!(EVIOCGKEY,     b'E', 0x18, OutData<[u8]>);
+    pub(super) type GetLedState   = ioc!(EVIOCGLED,     b'E', 0x19, OutData<[u8]>);
+    pub(super) type GetSwState    = ioc!(EVIOCGSW,      b'E', 0x1B, OutData<[u8]>);
+
+    pub(super) type GetBit = IoctlEnum<b'E', 0x20, 0x1F, OutData<[u8]>>;
```
lrh2000: Please mention the Linux name otherwise it's hard to look it up from the reference:
```rust
/// The `EVIOCGBIT` ioctl enum.
type GetEventBits = IoctlEnum<b'E', 0x20, 0x1F, OutData<[u8]>>;
```

Besides, the example in `ioctl.rs` uses `GetEventBits`, I think it's clearer so let's use it here as well?

https://github.com/asterinas/asterinas/blob/a176e4c3c6f40a3c54ea3ed4d471ba41cf736daf/kernel/src/util/ioctl/mod.rs#L504-L507

kernel/src/device/evdev/file.rs
```diff
@@ -27,14 +28,35 @@ use crate::{
...
+    use crate::util::ioctl::{ioc, InData, IoctlEnum, OutData};
+
+    // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/input.h>
+
+    pub(super) type SetClockId    = ioc!(EVIOCSCLOCKID, b'E', 0xa0, InData<i32>);
+    pub(super) type GetVersion    = ioc!(EVIOCGVERSION, b'E', 0x01, OutData<i32>);
+    pub(super) type GetId         = ioc!(EVIOCGID,      b'E', 0x02, OutData<InputId>);
+    pub(super) type GetDeviceName = ioc!(EVIOCGNAME,    b'E', 0x06, OutData<[u8]>);
+    pub(super) type GetPhys       = ioc!(EVIOCGPHYS,    b'E', 0x07, OutData<[u8]>);
+    pub(super) type GetUniq       = ioc!(EVIOCGUNIQ,    b'E', 0x08, OutData<[u8]>);
```
lrh2000: Isn't it a bit odd because `GetDeviceName` contains device but `GetPhys`/`GetUniq` does not?

I think the following version is better.
```suggestion
    pub(super) type GetDriverVer  = ioc!(EVIOCGVERSION, b'E', 0x01, OutData<i32>);
    pub(super) type GetInputId    = ioc!(EVIOCGID,      b'E', 0x02, OutData<InputId>);
    pub(super) type GetDeviceName = ioc!(EVIOCGNAME,    b'E', 0x06, OutData<[u8]>);
    pub(super) type GetDevicePhys = ioc!(EVIOCGPHYS,    b'E', 0x07, OutData<[u8]>);
    pub(super) type GetDeviceUniq = ioc!(EVIOCGUNIQ,    b'E', 0x08, OutData<[u8]>);
```

kernel/src/device/evdev/file.rs
```diff
@@ -27,14 +28,35 @@ use crate::{
...
 pub(super) const EVDEV_BUFFER_SIZE: usize = 64;
 
+mod ioctl_defs {
+    use aster_input::input_dev::InputId;
+
+    use crate::util::ioctl::{ioc, InData, IoctlEnum, OutData};
+
+    // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/input.h>
+
+    pub(super) type SetClockId    = ioc!(EVIOCSCLOCKID, b'E', 0xa0, InData<i32>);
```
lrh2000:
```suggestion
    pub(super) type SetClockId    = ioc!(EVIOCSCLOCKID, b'E', 0xA0, InData<i32>);
```

Keep the list sorted. This should be moved to the end because `0xA0` > `0x1B` (`GetSwState`).

kernel/src/device/evdev/file.rs
```diff
@@ -131,6 +169,29 @@ impl EvdevFileInner {
...
+        if consumer.is_empty() {
+            return;
+        }
+
+        consumer.clear();
+
+        let time = self.read_clock();
+        let dropped_event =
+            EvdevEvent::from_event_and_time(&InputEvent::from_sync_event(SynEvent::Dropped), time);
+        producer.push(dropped_event);
```
lrh2000:
```suggestion
        producer.push(dropped_event).unwrap();
```

kernel/src/device/evdev/file.rs
```diff
@@ -226,6 +311,56 @@ pub(super) fn is_syn_dropped_event(event: &EvdevEvent) -> bool {
...
+        let mut reader = VmReader::from(&bitmap[..copy_len]);
+        writer.write_fallible(&mut reader)?;
+    }
+
+    let remaining = len - copy_len;
+    if remaining > 0 {
+        writer.fill_zeros(remaining)?;
+    }
+
+    Ok(())
```
lrh2000: This should be implemented similarly as `write_string`:
```rust
    let mut reader = VmReader::from(bitmap);
    writer.write_fallible(&mut reader)?;
    writer.fill_zeros(writer.avail())?;
    Ok(())
}
```

So I think you will only need to provide a `write_bytes_and_zeros_to_userspace` instead of both `write_string_to_userspace` and `write_bitmap_to_userspace`?

kernel/src/device/evdev/file.rs
```diff
@@ -292,17 +427,88 @@ impl FileIo for EvdevFile {
...
+            }
+            cmd @ GetLedState => {
+                // No LED events are supported. So we can report zeros as LED states here.
+                cmd.with_writer(|mut writer| {
+                    writer.fill_zeros(writer.avail())?;
+                    Ok(())
+                })?;
+            }
+            cmd @ GetSwState => {
+                // No SW events are supported. So we can report zeros as switch states here.
```
lrh2000:
```suggestion
                // No switch events are supported. So we can report zeros as switch states here.
```

kernel/src/device/evdev/file.rs
```diff
@@ -292,17 +427,88 @@ impl FileIo for EvdevFile {
...
+                cmd.base_ioctl()
+                    .with_writer(|mut writer| handle_get_bit(&evdev, event_type, &mut writer))?;
+            }
+            cmd @ GetKeyState => {
+                // TODO: We need to track whether the key is currently pressed and report that state
+                // here. If we report states, we need to flush the queue to avoid interfering with the
+                // user space's state tracking. See the Linux implementation at:
+                // <https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L872-L876>.
+
+                // Currently, no key state tracking is supported. So we report zeros as key states here.
```
lrh2000, line 459:
```suggestion
                // <https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L872-L876>.
                //
                // Currently, no key state tracking is supported. So we report zeros as key states here.
```

kernel/src/device/evdev/file.rs
lrh2000: I think you got my point. Just a small thing to clarify.

> Our target application, `Xorg` (via `libevdev`), does not use the information returned by `GetKeyState` (or `GetLedState` and `GetSwState`).

Here, what you mean is that it uses these ioctls but never checks the results, right? So we cannot make these ioctls fail.

kernel/src/device/evdev/file.rs
```diff
@@ -226,6 +311,56 @@ pub(super) fn is_syn_dropped_event(event: &EvdevEvent) -> bool {
...
+
+    Ok(())
+}
+
+fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWriter) -> Result<()> {
+    let capability = evdev.device.capability();
+
+    match event_type as u16 {
+        0 => {
+            let event_types_bytes = capability.event_types_bits().to_le_bytes();
```
lrh2000: I think it is `to_ne_bytes`:
https://elixir.bootlin.com/linux/v4.3/source/drivers/input/evdev.c#L877-L878
https://elixir.bootlin.com/linux/v4.3/source/drivers/input/evdev.c#L620-L625

lrh2000 (approved):
Thanks for the revision. I took another look and have a few more comments, but they're minor style issues. I'm approving this, and then @tatetian can perform the final checks.

Cautreoxit:
Thanks for your comments and approval! I have addressed these issues now.

Regarding the question in this comment https://github.com/asterinas/asterinas/pull/2582#discussion_r2596137771, yes, `libevdev` calls the `GetKeyState` ioctl during initialization, although our application doesn't use the information returned. However, it does check the result code, and if there's an error, the initialization would fail. Therefore, we cannot allow these ioctls to fail. Here's how `evdevlib` uses `GetKeyState`:
https://github.com/whot/libevdev/blob/c3953e1bb8f813f3e248047d23fb0775b147da9f/libevdev/libevdev.c#L428

PR #2414, opened 2025-09-08
Introduce `VmPrinter` to write formatted output to the user space

cchanging:
A implemenation for the design of @lrh2000 in the [comment](https://github.com/asterinas/asterinas/pull/2160#discussion_r2322555452).

This PR can give the code that generates output to user-space data and writes to the writer a better shape. It can also benefit the code in `ProcFs`.

One thing I'm not entirely sure about is the location of the implementation. Currently, since this writer serves kernel-generated data and is somewhat aligned with `SysTree`, and the primary users are methods related to `SysTree`, I've placed it in the corresponding crate. However, I'm not certain if this would still be appropriate if there are future users, such as those related to `ProcFs`.

PR #2583, opened 2025-11-14
Refactor `procfs` with `VmPrinter`

vvvvsv:
Thanks to https://github.com/asterinas/asterinas/pull/2414 that introduced `VmPrinter`, we can use it to refactor `procfs` to avoid unnecessary heap allocations.

After the refactoring, I think we now have too much code following this pattern:

```rust
impl FileOps for SomeFileOps {
    fn data(&self) -> Result<Vec<u8>> {
        unreachable!()
    }

    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
        // Some writes
        Ok(printer.bytes_written())
    }
}
```

This is a bit ugly. We can enforce that **exactly** one of `fn data` and `fn read_at` is implemented at type level. The second commit achieves this by:

https://github.com/asterinas/asterinas/blob/bfec8061fa843c763786ac4e33d6f6301fe22e4e/kernel/src/fs/procfs/template/file.rs#L99-L130

### DIFF
--- a/kernel/src/fs/procfs/template/file.rs
+++ b/kernel/src/fs/procfs/template/file.rs
@@ -97,17 +97,7 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
 }
 
 pub trait FileOps: Sync + Send {
-    fn data(&self) -> Result<Vec<u8>>;
-
-    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
-        let data = self.data()?;
-        if offset >= data.len() {
-            return Ok(0);
-        }
-
-        let written_len = writer.write_fallible(&mut (&data[offset..]).into())?;
-        Ok(written_len)
-    }
+    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize>;
 
     fn write_at(&self, _offset: usize, _reader: &mut VmReader) -> Result<usize> {
         return_errno_with_message!(Errno::EPERM, "the file is not writable");
--- a/kernel/src/fs/procfs/cmdline.rs
+++ b/kernel/src/fs/procfs/cmdline.rs
@@ -5,8 +5,7 @@
 //!
 //! Reference: <https://man7.org/linux/man-pages/man5/proc_cmdline.5.html>
 
-use alloc::format;
-
+use aster_util::printer::VmPrinter;
 use ostd::boot::boot_info;
 
 use crate::{
@@ -33,10 +32,13 @@ impl CmdLineFileOps {
 }
 
 impl FileOps for CmdLineFileOps {
-    fn data(&self) -> Result<Vec<u8>> {
+    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        let mut printer = VmPrinter::new_skip(writer, offset);
+
         // TODO: Parse additional kernel command line information with `bootconfig`.
         // See <https://docs.kernel.org/admin-guide/bootconfig.html> for details.
-        let cmdline = format!("{}\n", boot_info().kernel_cmdline);
-        Ok(cmdline.into_bytes())
+        writeln!(printer, "{}", boot_info().kernel_cmdline)?;
+
+        Ok(printer.bytes_written())
     }
 }
--- a/kernel/src/fs/procfs/cpuinfo.rs
+++ b/kernel/src/fs/procfs/cpuinfo.rs
@@ -5,6 +5,7 @@
 //!
 //! Reference: <https://man7.org/linux/man-pages/man5/proc_cpuinfo.5.html>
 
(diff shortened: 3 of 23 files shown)


vvvvsv:
@boterinas rerun

kernel/src/fs/procfs/template/file.rs
```diff
@@ -96,9 +96,15 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
     }
 }
 
-pub trait FileOps: Sync + Send {
+pub trait FileOpsData {
     fn data(&self) -> Result<Vec<u8>>;
```
tatetian: Could you enlighten me as to why the `data` method is not deprecated completely in favor of `read_at`? I mean `read_at` can output both text and non-text data. So it seems to me that `read_at` is always superior than `data`.

tatetian (commented):
Thanks for cleaning up ProcFS. Using the printer API should be the new norm for ProcFS.

kernel/src/fs/procfs/template/file.rs
vvvvsv: > `read_at` can output both text and non-text data.

Every `FileOps` implementation that writes non-text data to a writer needs to follow the pattern:

```rust
        let data: Vec<u8> or [u8];
        if offset >= data.len() {
            return Ok(0);
        }

        let written_len = writer.write_fallible(&mut (&data[offset..]).into())?;
        Ok(written_len)
```

I think this pattern a bit complex, so it probably needs to be a function?

However, I think it is better to add this function and deprecate the `data` method. This would make the code clearer.

cchanging: The currently only usage of this pattern is the `comm` file. For this file, I feel a more appropriate approach would be something like this: 
```rust
let exe_path = self.0.executable_path();
let last_component = exe_path.rsplit('/').next().unwrap_or(&exe_path);
writeln!(printer, "{:.TASK_COMM_LEN}", last_component)?;

...
```

I'm not sure if there will be other procfs files that need this pattern in the future, but it seems there aren't any at the moment.

vvvvsv: > ```rust
> writeln!(printer, "{:.TASK_COMM_LEN}", last_component)?;
> ```

This truncation is based on the number of UTF-8 characters. [Playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b87df9270110683289dc3a2eba17606a)

In Linux, we can set `comm` to a UTF-8 string. When Linux truncates this string using `TASK_COMM_LEN`, it counts bytes rather than characters.

```c
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <string.h>

void print_comm() {
    char buf[64] = {0};
    int fd = open("/proc/self/comm", O_RDONLY);
    if (fd == -1) {
        perror("open /proc/self/comm");
        return;
    }
    ssize_t n = read(fd, buf, sizeof(buf));
    if (n == -1) {
        perror("read /proc/self/comm");
        close(fd);
        return;
    }
    buf[n] = '\0';
    printf("Read %zd bytes from /proc/self/comm\n", n);
    printf("content: '%s'\n", buf);
    close(fd);
}

void write_comm(const char *name) {
    int fd = open("/proc/self/comm", O_WRONLY);
    if (fd == -1) {
        perror("open /proc/self/comm for write");
        return;
    }
    ssize_t n = write(fd, name, strlen(name));
    if (n == -1) {
        perror("write /proc/self/comm");
    }
    close(fd);
}

int main() {
    printf("Before change:\n");
    print_comm();

    write_comm("测试测试测试");
    printf("After change 1:\n");
    print_comm();

    write_comm("测试测试αα");
    printf("After change 1:\n");
    print_comm();


    return 0;
}
```

The code above produces the following output on Linux:

```
Before change:
Read 5 bytes from /proc/self/comm
content: 'tmp1
'
After change 1:
Read 16 bytes from /proc/self/comm
content: '测试测试测
' // "测"=3 bytes, "试" = 3 bytes, '\n' = 1 byte 
After change 1:
Read 16 bytes from /proc/self/comm
content: '测试测试α�
' // "测"=3 bytes, "α" = 2 bytes, '\n' = 1 byte 
```

So we still need to generate a byte slice and truncate it to `TASK_COMM_LEN`.

tatetian: I suggest removing this helper function because 
1. It is only used once;
2. It is quite short;
3. It is not discoverable. An implementator of the `read_at` method is likely to re-invent this logic in a subtly different way rather than calling this helper method.

In short, it is basically useless as a helper method.

In addition, I think a slightly better way to write it is the following form:

```rust
let mut vm_reader  = VmReader::from(&data[offset.min(data.len())..]);
writer.write_fallible(&mut vm_reader);
```

PR #2584, opened 2025-11-14
Add a new RFC: Asterinas NixOS

tatetian:
This RFC formally proposes the establishment of an Asterinas distribution as a new, top-level sub-project of Asterinas. We intend for this distribution to leverage [NixOS](https://nixos.org/) due to its unparalleled customizability and rich package ecosystem. Accordingly, we propose naming this new sub-project **NixOS on Asterinas (NOSA)**.

Edit: The sub-project name is changed to "Asterinas NixOS".

### DIFF
--- a/book/src/rfcs/0002-asterinas-nixos.md
+++ b/book/src/rfcs/0002-asterinas-nixos.md
@@ -1,0 +1,139 @@
+# RFC-0002: Asterinas NixOS
+
+* Status: Approved
+* Pull request: https://github.com/asterinas/asterinas/pull/2584
+* Date submitted: 2025-11-14
+* Date approved: 2025-12-01
+
+## Summary
+
+This RFC formally proposes the establishment of an Asterinas distribution as a new, top-level sub-project of Asterinas. We intend for this distribution to leverage [NixOS](https://nixos.org/) due to its unparalleled customizability and rich package ecosystem. Accordingly, we propose naming this new sub-project **Asterinas NixOS (AsterNixOS)**.
+
+## Motivations
+
+### What is a "distro"?
+
+In the context of operating systems (OSes), a "distribution" (or "distro") typically refers to a complete OS built around a kernel, complemented by userspace tools, libraries, and applications managed via a package management system. A Linux distro, for example, combines the Linux kernel with a rich userspace. Similarly, an Asterinas distro will pair the Asterinas kernel with a comprehensive userspace environment. For the purpose of this discussion, "distro" will refer broadly to either a Linux or Asterinas-based distro.
+
+### Short-term strategic advantages
+
+Achieving a Minimum Viable Product (MVP) milestone is crucial for Asterinas's maturation. Reaching MVP means that Asterinas is ready for evaluation by early adopters, who expect a seamless experience comparable to mainstream OSes. This includes easy application installation and out-of-the-box functionality. Simply providing a kernel is insufficient; we must deliver a user-friendly experience, which necessitates a full-fledged distro complete with an intuitive package manager.
+
+Furthermore, direct control over an Asterinas distro offers a pragmatic approach to addressing Linux compatibility challenges. While Asterinas is committed to a high degree of Linux compatibility, achieving perfect parity in the near term is impractical. A dedicated distro allows us to configure or patch specific packages to circumvent reliance on advanced or bleeding-edge Linux features that do not have high priority in Asterinas. This significantly reduces pressure on Asterinas kernel developers to implement complex features prematurely, enabling a more focused and stable development roadmap.
+
+Beyond external users, Asterinas developers themselves will greatly benefit from a dedicated distro. Testing complex applications with intricate dependencies is a significant hurdle. While "Hello World" or even medium-sized projects like Nginx or Redis can be manually built and integrated into a disk image, this approach does not scale for complex applications such as Spark, PyTorch, or Chromium. Identifying, understanding, and building every dependency for such projects would divert critical kernel development resources. A distro, by its very nature, abstracts this complexity through its package manager, streamlining the testing and development workflow for all Asterinas developers.
+
+### Long-term ecosystem vision
+
+Looking ahead, we envision a thriving ecosystem of Asterinas distros. This initial, kernel-developer-maintained distro will serve as a vital reference implementation, fostering the creation of diverse Asterinas-based distros. While the Linux world boasts numerous distros, Asterinas-specific distros will be essential for two primary reasons:
+
+Firstly, Asterinas prioritizes Linux compatibility at the external ABI level, not the internal kernel module interface. This means Asterinas cannot load Linux kernel modules (`.ko` files). An Asterinas distro will ensure that all system programs exclusively load Asterinas kernel modules.
+
+Secondly, Asterinas will eventually provide features and value propositions that Linux simply does not have—for example, new system calls, file systems, and device drivers. These benefits only become real when applications can detect and leverage them.
+
+An Asterinas distro is the natural vehicle for this. It can ship new packages written specifically for Asterinas and carry patches to existing packages to make them "Asterinas-aware", ensuring that userspace can meaningfully take advantage of the kernel’s unique capabilities.
+
+In essence, establishing Asterinas distros is paramount for Asterinas's long-term success and differentiation. This RFC lays the groundwork for that future by proposing our foundational distro.
+
+## Design
+
+### Why NixOS as the base distro?
+
+Building the first Asterinas distro from scratch would be an immense undertaking, diverting focus from core kernel development. Therefore, basing it on an existing Linux distro is the most rational path forward. This leads to the crucial question: which Linux distro should serve as our foundation?
+
+The landscape of Linux distros is vast, featuring prominent names such as [Arch](https://archlinux.org/), [CentOS](https://www.centos.org/), [Debian](https://www.debian.org/), and [Gentoo](https://www.gentoo.org/). While theoretically, any of these could serve, we have identified [NixOS](https://nixos.org/) as a particularly attractive and uniquely suited option.
+
+In most distros, package recipes are centered around shell scripts (e.g., [`PKGBUILD`](https://man.archlinux.org/man/PKGBUILD.5#EXAMPLE) in Arch Linux or the [`rules`](https://www.debian.org/doc/manuals/maint-guide/dreq.en.html#rules) file in a Debian package). NixOS, however, utilizes a purpose-built, purely functional language called [Nix](https://nixos-and-flakes.thiscute.world/the-nix-language/), which offers unparalleled expressiveness and flexibility for defining package recipes.
+
+A core requirement for our base distro is the ability to easily tweak existing package recipes to work around Asterinas's limitations or customize the new distro's look or behavior. The Nix language excels in this regard.
+
+For instance, consider overriding specific attributes of an existing Nix package, such as `xfdesktop`, with minimal code:
+
+```nix
+{ pkgs }:
+{
+  xfdesktop = pkgs.xfce.xfdesktop.overrideAttrs (oldAttrs: {
+    version = "4.16.0";
+    patches = (oldAttrs.patches or []) ++ [
+      ./patches/xfdesktop4/0001-Fix-not-using-consistent-monitor-identifiers.patch
+    ];
+  });
+}
+```
+
+This Nix file creates a customized `xfdesktop` package without requiring intrusive modifications to its original recipe. This capability, known as [the override design pattern](https://nixos.org/guides/nix-pills/14-override-design-pattern.html) in Nix, is uniquely feasible due to Nix's first-class functions and lazy evaluation.
+
+Another powerful example of NixOS's flexibility is how easily we can create an Asterinas distro ISO installer by customizing an existing [NixOS installer ISO](https://nixos.org/download/#nixos-iso). Consider the following `iso-image.nix` file:
+
+```nix
+{ lib, pkgs, ... }:
+let
+  asterinas_kernel = builtins.path { path = "asterinas_bin"; };
+  auto_install_script = pkgs.replaceVarsWith {
+    src = "./auto_install.sh";
+    isExecutable = true;
+    replacements = {
+      shell = "${pkgs.bash}/bin/sh";
+      inherit asterinas_kernel;
+    };
+  };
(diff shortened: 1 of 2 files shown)


junyang-zh (approved):
LGTM!

tatetian:
> one of Nix's biggest strengths is its reproducibility:

Yes, reproducibility is one of the main strengths of NixOS, but a lesser consideration when it comes to picking our base distro. If NixOS offered reproducibility but was hard to customize, then I won't promote it. So I left this point in the RFC.

arthurpaulino:
NixOS is a great pick, from a pragmatic technical standpoint!

May I just suggest other names? This is of course subjective, but I don't think "NOSA" is quite punchy.

Asking some AI, it came up with the following names:
1. Asternix
2. AsterixOS
3. AstraNix
4. AsterNixOS
5. Astera

I think "Asternix" captures the essence of the project while still sounding great. Isn't there a crate that's already named `aster_nix`? Why wouldn't that name be picked?

tatetian:
> May I just suggest other names? This is of course subjective, but I don't think "NOSA" is quite punchy.

Yeah, I wouldn't say NOSA is a punchy name. But I wanted to have a descriptive name and avoid inventing unique names such as Asternix. Unique names should be considered a new brand; but the distro is a sub-project of Asterinas, not an independent project. So it is not significant enough to introduce a new brand name and bring all the troubles associated with trademark registration around the globe.

Since we mentioned trademark registration, I am wondering if we are allowed to have a project name starting with NixOS, a trademark owned by the NixOS Foundation. Starting the project name with "NixOS" could be interpreted as an official NixOS project. So I guess that we will need to have permissions from the NixOS foundation to use the "NixOS on Asterinas" name. I will send an email to the Nix Foundation.

In the meantime, feel free to propose other names for this sub-project.

book/src/rfcs/0002-asterinas-nixos.md
```diff
@@ -0,0 +1,151 @@
...
+
+Firstly, Asterinas prioritizes Linux compatibility at the external ABI level, not the internal kernel module interface. This means Asterinas cannot load Linux kernel modules (`.ko` files). An Asterinas distro will ensure that all system programs exclusively load Asterinas kernel modules, maintaining integrity and security within our native environment.
+
+Secondly, Asterinas will eventually offer unique features and differentiated value propositions compared to Linux. These may include novel system calls, file systems, or device drivers. Realizing the full potential and value of these unique features will inherently require cooperation from the userspace—that is, the distro itself.
+
+In essence, establishing Asterinas distros is paramount for Asterinas's long-term success and differentiation. This RFC lays the groundwork for that future by proposing our foundational distro.
+
+## Design
```
lrh2000, line 38: There is no design in this section, is there?

First of all, from a higher level, it does not say what we want to deliver.

How is a Linux distro delivered? Typically, there are three steps:
 1. There is an installable ISO (which contains many pre-built utilities for installing the OS, e.g., https://archlinux.org/download).
 2. Using the installer, one can install the kernel and a set of basic userspace programs (e.g., https://wiki.archlinux.org/title/Installation_guide#Install_essential_packages).
 3. Then, one can install more necessary programs via the package manager (e.g., `pacman` for ArchLinux, `apt-get` for Ubuntu).

Then, what do we want to deliver?
 - For (1), we have a script to generate an ISO?
 - For (2), we maintain a list of tested and supported programs to install?
 - What about (3)?
   - A reasonable distro will support thousands of Linux programs (https://archlinux.org/packages/: 15635 matching packages found), so will they appear under `distro/`?
   - What is the policy for accepting (or not accepting) a userspace program into `distro/`?
   - We should distribute prebuilt packages via a server so that users do not have to build them locally?
   - How does it interact with our CI workflow and the publish workflow?

book/src/rfcs/0002-nixos-on-asterinas.md
```diff
@@ -0,0 +1,151 @@
...
+
+A further significant advantage of NixOS is the portability of [Nix](https://nixos.org/download/#nix-install-linux), its package manager (not Nix, the programming language), across various Linux distros and even macOS.
+
+Unlike most package managers, which are tightly coupled to their parent distros (e.g., Pacman for Arch, `dpkg` for Debian, `rpm` for Red Hat-based systems), Nix operates independently. Nix packages are installed in the `/nix/store` directory, ensuring they do not conflict with native packages. This portability is invaluable for debugging: if a package in our NixOS-based distro encounters an issue, we can replicate the exact same package environment on any Linux development machine, greatly simplifying troubleshooting.
+
+### The New Top-Level Sub-Project
+
+Given these compelling rationales, we propose the creation of a new top-level sub-project named _NixOS on Asterinas (NOSA)_. This sub-project will reside in a new `distro/` directory at the project root.
```
lrh2000: This section also does not explain how files are (planned to be) organized under the new `distro/` sub-project at the implementation level.

 - ArchLinux creates one GitLab repository for each package it supported, e.g, https://gitlab.archlinux.org/archlinux/packaging/packages/gcc.
 - Gentoo creates a directory under the category's directory (e.g., `app-admin`, `dev-lang`, see https://github.com/gentoo/gentoo/), e.g., https://github.com/gentoo/gentoo/tree/master/dev-lang/rust.

Perhaps there are not only packages, but also other items, such as the script that builds the ISO.

We cannot decide everything here but we should say something?

book/src/rfcs/0002-nixos-on-asterinas.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Long-term ecosystem vision
+
+Looking ahead, we envision a thriving ecosystem of Asterinas distros. This initial, kernel-developer-maintained distro will serve as a vital reference implementation, fostering the creation of diverse Asterinas-based distros. While the Linux world boasts numerous distros, Asterinas-specific distros will be essential for two primary reasons:
+
+Firstly, Asterinas prioritizes Linux compatibility at the external ABI level, not the internal kernel module interface. This means Asterinas cannot load Linux kernel modules (`.ko` files). An Asterinas distro will ensure that all system programs exclusively load Asterinas kernel modules, maintaining integrity and security within our native environment.
+
+Secondly, Asterinas will eventually offer unique features and differentiated value propositions compared to Linux. These may include novel system calls, file systems, or device drivers. Realizing the full potential and value of these unique features will inherently require cooperation from the userspace—that is, the distro itself.
```
lrh2000: Per the existing practice, Linux distros only package existing user programs and almost never create new ones. So if there is a new system call, it's not the distro's responsibility to make use of it. It's almost always the upstream maintainer's responsibility.

It also sounds odd to me that we will write a user program and put its source code in the `distro/` directory. Well, there are examples like Gentoo maintains [OpenRC](https://wiki.gentoo.org/wiki/OpenRC), but is that even intended by this RFC?

For Linux,
 - A few userspace programs are maintained in tree, and the distro may choose to package them, e.g., [`perf`](https://github.com/torvalds/linux/tree/master/tools/perf).
 - Some userspace programs are maintained out of tree, but they share the same maintainer as their kernel subsystem's maintainer, and the distro may choose to package them, e.g., x86 KVM and x86 QEMU.
 - Most userspace programs are developed independently. They will use the new system calls if they feel appropriate.

For "Long-term ecosystem vision", I think distros will not help much. But maybe my long-term vision is too long?

book/src/rfcs/0002-asterinas-nixos.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Drawbacks, Alternatives, and Unknowns
+
+### Drawbacks
+
+* **Learning curve for Nix/NixOS:** This is arguably the most significant hurdle. While exceptionally powerful, the Nix ecosystem, including the Nix language and its functional paradigm, presents a steeper learning curve compared to conventional package managers and build systems. This could pose an initial barrier for new contributors and existing Asterinas developers unfamiliar with Nix. Investing in clear documentation and onboarding resources will be critical.
+* **Maintenance overhead of Nixpkgs overlay:** While the override pattern offers flexibility, maintaining a dedicated overlay for Asterinas-specific patches and configurations within the vast `nixpkgs` repository will still require continuous effort. Keeping pace with upstream `nixpkgs` changes, resolving potential conflicts, and ensuring compatibility will be an ongoing challenge requiring dedicated resources.
```
lrh2000, line 128: We'll have to handle userspace security reports: https://lists.ubuntu.com/archives/ubuntu-security-announce/2025-October/subject.html. But do we really want to handle them ourselves for thousands of packages?

We'll definitely maintain the kernel and kernel modules ourselves. But what about everything else? Assuming we won't deviate too much from NixOS, we can say that we'll rely on upstream resources. Am I understanding this correctly?

I think I've asked similar question several times from different perspectives, but this RFC still leaves me wondering which parts we'll have to rely on NixOS's upstream for and which parts we won't.

book/src/rfcs/0002-asterinas-nixos.md
tatetian: > There is no design in this section, is there?

The biggest design decision for the first Asterinas distro is which distro to choose as the base distro. The RFC take  a great length to answer that. Since no one is challenging this point, I think that it is fair to say that the design section does a decent job on this point.

> Then, what do we want to deliver?
> 
> * For (1), we have a script to generate an ISO?
> * For (2), we maintain a list of tested and supported programs to install?
> * What about (3)?
>   
>   * A reasonable distro will support thousands of Linux programs (https://archlinux.org/packages/: 15635 matching packages found), so will they appear under `distro/`?
>   * What is the policy for accepting (or not accepting) a userspace program into `distro/`?
>   * We should distribute prebuilt packages via a server so that users do not have to build them locally?
>   * How does it interact with our CI workflow and the publish workflow?

The questions that you posed are good ones, but most of them can follow the established practices of NixOS (e.g., "we have a script to generate an ISO?") or are considered implementation details. (e.g., "How does it interact with our CI workflow and the publish workflow").

>   * A reasonable distro will support thousands of Linux programs (https://archlinux.org/packages/: 15635 matching packages found), so will they appear under `distro/`?
>   * What is the policy for accepting (or not accepting) a userspace program into `distro/`?

These two questions are particularly interesting because they cannot be derived from NixOS directly and need to be answered by the RFC author. Although they won't affect the feasibility of using NixOS, addressing them directly in the RFC will strength the argument made in the RFC. So I will expand the "The New Top-Level Sub-Project" section.

book/src/rfcs/0002-nixos-on-asterinas.md
tatetian: I want to emphasize that this RFC should provide the right level of details. Implementation details are subject to change and should be considered out of scope for an RFC, which is an immutable, historic record. For example, a Rust RFC is almost always focused on describing the proposed language feature, not how the language feature is implemented.

Again, I will slightly expand this section to give the reader a more concrete idea on how the "NixOS on Asterinas" distro may be built upon the vanilla NixOS.

book/src/rfcs/0002-nixos-on-asterinas.md
tatetian: Assume that we have created a tool that can leverage Asterinas unique features. Will a Linux distro include the tool? I don't think so. By definition, a Linux distro is built for the Linux kernel. It only makes sense to include the Asterinas-specific tool in an Asterinas disro. So we need to foster Asterinas distros.

book/src/rfcs/0002-asterinas-nixos.md
tatetian: > I think I've asked similar question several times from different perspectives, but this RFC still leaves me wondering which parts we'll have to rely on NixOS's upstream for and which parts we won't.

I got your confusion. No worry. Without some prior knowledge about Nixpkgs, a reader won't be able to imagine how "NixOS on Asterinas" can easily reuse tens of thousands of existing package from Nixpkgs without including them explicitly in our repo. I will explain how this can be done.

tatetian:
> I will send an email to the Nix Foundation.
> 
> In the meantime, feel free to propose other names for this sub-project.

I thought of three more descriptive names:
* "Asterinas-Powered NixOS"
* "Asterinas NixOS Edition"
* "Asterinas NixOS"

These three names start with "Asterinas", rather than "NixOS", clarifying the primary ownership and focus. I am in favor of "Asterinas NixOS". It is clear, short, direct, and thus sounds powerful.

---

Edit: I have sent the email to the NixOS foundation. It has been over a week and I haven't received any responses. I will proceed with the name "Asterinas NixOS". Its abbreviation is "AsterNixOS".

tatetian:
@lrh2000 I didn't expand the "The New Top-Level Sub-Project" section, but did bring up two key NixOS features: `config.system.build.isoImage` and `configuration.nix` (and its overlays).

https://github.com/asterinas/asterinas/blob/d0a5d16b10ca502ae7429fd9094314d6eb1d4427/book/src/rfcs/0002-nixos-on-asterinas.md?plain=1#L64-L109

The discussions about these features shed some light on how AsterNixOS may be created on top of the vanilla NixOS.

book/src/rfcs/0002-nixos-on-asterinas.md
lrh2000: A new system call for existing applications differs from a new system call for new applications. We cannot patch every application ourselves for the former; it is not the distro's job either.

I can get your point and I don't really require any changes here, though I think it's still difficult to discuss "long-term ecosystem vision" at the project's current stage.

book/src/rfcs/0002-asterinas-nixos.md
lrh2000, line 1: The filename is outdated:
 - `0002-nixos-on-asterinas.md`

The commit message is outdated:
 - `Add a RFC: NixOS on Asterinas`

book/src/rfcs/0002-nixos-on-asterinas.md
tatetian: To address your concern, I rewrote the paragraph to make the roles of packages and distros more clear:

> Secondly, Asterinas will eventually provide features and value propositions that Linux simply does not have—for example, new system calls, file systems, and device drivers. These benefits only become real when applications can detect and leverage them.
>
> An Asterinas distro is the natural vehicle for this. It can ship new packages written specifically for Asterinas and carry patches to existing packages to make them "Asterinas-aware", ensuring that userspace can meaningfully take advantage of the kernel’s unique capabilities.

Since this is not a required change and I am quite confident with the revised paragraph, I will merge the PR. Thanks for the feedback!

PR #2587, opened 2025-11-15
Remove outdated safety comments and correct `init_on_xx` names

lrh2000:
This is totally outdated now because the arguments do not even exist?
https://github.com/asterinas/asterinas/blob/35ab40057a50c587fcfca9a13110e2db55178b45/ostd/src/arch/x86/irq/ipi.rs#L36-L41

I think we should just remove them. There seem to be no unsafety here. So I mark the method as a safe method.

---

Additionally, let's follow this rule when naming OSTD's initialization methods:
 - If a module has an initialization method that only needs to be called on the BSP, name it as `fn init()`.
 - If a module has an initialization method that needs to be called on each CPU, name it as `fn init_on_cpu()`.
 - If a module has two initialization methods, one that needs to be called on the BSP and one that needs to be called on the APs, name them `fn init_on_bsp()` and `fn init_on_ap()`, respectively.

See the second commit for details. I think this rule is very clear, and following it won't be too burdensome.

### DIFF
--- a/ostd/src/arch/loongarch/mod.rs
+++ b/ostd/src/arch/loongarch/mod.rs
@@ -51,6 +51,14 @@ pub(crate) unsafe fn late_init_on_bsp() {
     unsafe { crate::io::init(io_mem_builder) };
 }
 
+/// Initializes application-processor-specific state.
+///
+/// # Safety
+///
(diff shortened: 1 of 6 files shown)


junyang-zh (approved):
LGTM. Thanks for correcting them.

PR #2282, opened 2025-07-25
Implement controller framework for cgroup subsystem

cchanging:
Based on #2160.

This PR implements a controller framework for cgroup subsystem and add the dummy implementations of "memory", "cpuset" and "pids" controller.

With this PR, one can view the active controllers in current cgroup through "cgroup.controllers", and can delegate control functionality to the children of the cgroup through "cgroup.subtree_control".

Serves for #2214.

PR #2592, opened 2025-11-17
Support some cgroup attributes required by Podman

cchanging:
Based on #2282 .

This PR supports some attributes of cgroup pids controller and cpuset controller, allowing reading/writing behavior for them (these attributes actually have no effect). In addition, This PR supports cgroup.freeze attribute, which can stop or resume the processes in the corresponding cgroup.

### DIFF
--- a/kernel/src/fs/cgroupfs/controller/pids.rs
+++ b/kernel/src/fs/cgroupfs/controller/pids.rs
@@ -1,15 +1,19 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use alloc::sync::Arc;
+use core::sync::atomic::{AtomicU32, Ordering};
 
-use aster_systree::{Error, Result, SysAttrSetBuilder, SysPerms, SysStr};
+use aster_systree::{Error, MAX_ATTR_SIZE, Result, SysAttrSetBuilder, SysPerms, SysStr};
+use aster_util::printer::VmPrinter;
 use ostd::mm::{VmReader, VmWriter};
+
+use crate::{process::posix_thread::PID_MAX, util::ReadCString};
 
 /// A sub-controller responsible for PID resource management in the cgroup subsystem.
 ///
 /// This controller will only provide interfaces in non-root cgroup nodes.
 pub struct PidsController {
-    _private: (),
+    max_pid: AtomicU32,
 }
 
 impl PidsController {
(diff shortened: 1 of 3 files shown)


kernel/src/fs/cgroupfs/controller/pids.rs
```diff
@@ -30,20 +39,55 @@ impl super::SubControl for PidsController {
...
+                let value = content
+                    .to_str()
+                    .map_err(|_| Error::InvalidOperation)?
+                    .trim();
+                let value = if value == "max" {
+                    usize::MAX
+                } else {
+                    value
+                        .parse::<usize>()
+                        .map_err(|_| Error::InvalidOperation)?
```
lrh2000: Can we reject this case since we don't really support it?

kernel/src/fs/cgroupfs/systree_node.rs
```diff
@@ -185,6 +210,49 @@ impl Debug for CgroupNode {
...
+    fn do_freeze(&mut self, freeze_op: FreezeOp) -> bool {
+        match freeze_op {
+            FreezeOp::Freeze => {
+                if self.is_frozen {
+                    return false;
+                }
+
+                for weak_process in self.processes.values() {
+                    if let Some(process) = weak_process.upgrade() {
+                        process.stop(SIGSTOP);
```
lrh2000: I don't believe this is correct. SIGSTOP notifies the parent process if the parent uses `wait()`. A process can resume after receiving `SIGSTOP` if someone sends a `SIGCONT` to it. However, it is unlikely that `cgroup.freeze` is the case.

lrh2000: What will happen if someone tries to move a process into a frozen cgroup. I'm not seeing the related code at all.

kernel/src/fs/cgroupfs/systree_node.rs
```diff
@@ -125,6 +127,29 @@ impl CgroupMembership {
...
+                continue;
+            };
+
+            if has_changed {
+                for child in node.fields.children_ref().read().values() {
+                    worklist.push(child.clone());
+                }
+            }
+
+            // TODO: Add the logics of upward propagation.
```
lrh2000: So, what is the behavior? The current logic stops if the value of is_frozen remains unchanged, but that doesn't make much sense because someone could unfreeze a cgroup in the middle.

kernel/src/fs/cgroupfs/systree_node.rs
```diff
@@ -185,6 +210,49 @@ impl Debug for CgroupNode {
...
+                if self.is_frozen {
+                    return false;
+                }
+
+                for weak_process in self.processes.values() {
+                    if let Some(process) = weak_process.upgrade() {
+                        process.stop(SIGSTOP);
+                    }
+                }
+
```
lrh2000: Remove this blank line.

kernel/src/fs/cgroupfs/systree_node.rs
cchanging: Sorry for late. You are correct — freeze and stop are actually two parallel concepts. Linux processes have a dedicated freeze state, and implementing it correctly involves certain complexities. 

In recent days, we conducted additional experiments and found that even without supporting write operations on the freeze attribute, Podman usage remains unaffected. Considering that introducing a proper freeze attribute implementation would require substantial effort, we have now changed it to only support read operations for the time being.

cchanging: It will be frozen. I did not implement this logic previously.

kernel/src/fs/cgroupfs/systree_node.rs
cchanging: The TODO here refers to that when all child cgroups under a parent cgroup enter the frozen state, this parent cgroup should also be marked as `frozen`.

kernel/src/fs/cgroupfs/controller/pids.rs
cchanging: I'm not sure what you mean. Are you suggesting that we should not implement dummy read/write operations for this attribute?

lrh2000: I mean, we could return an error if a value other than "max" is written. Currently, we only support the "max" value.
```rust
                log::warn!("The PIDs controller does not enforce PID limits yet.");
```

Of course, this may not be a big issue. But if Podman won't really write a non-"max" value (or we can configure it not to do so), I guess just rejecting the case is better and more correct.

kernel/src/fs/cgroupfs/systree_node.rs
lrh2000: Thanks for the explanation! Of course, simplifying the PR is a good thing.

kernel/src/fs/cgroupfs/controller/pids.rs
cchanging: We have done some tests today and found that Podman indeed requires to write a non-"max" value to this attribute. If writing is not allowed, it will cause runtime failures. 

There are quite a few locations involving the write behavior, hence patching would be rather challenging. Therefore, we currently lean towards implementing such a dummy solution in the kernel😂.

kernel/src/fs/cgroupfs/controller/pids.rs
```diff
@@ -21,18 +25,54 @@ impl PidsController {
...
+                    .map_err(|_| Error::PageFault)?;
+                let value = content
+                    .to_str()
+                    .map_err(|_| Error::InvalidOperation)?
+                    .trim();
+                let value = if value == "max" {
+                    u32::MAX
+                } else {
+                    value.parse::<u32>().map_err(|_| Error::InvalidOperation)?
+                };
```
lrh2000, line 65:
```rust
} else if Ok(value) = value.parse::<u32>() {
    if value >= PID_MAX {
        return Err(Error::InvalidOperation);
    }
    log::warn!("The cgroup's PID limit '{}' will not be enforced", value)
} else {
    return Err(Error::InvalidOperation);
}
```

 1. We should reject invalid PIDs like `u32::MAX` (which represents `max` instead), see https://elixir.bootlin.com/linux/v6.6-rc4/source/kernel/cgroup/pids.c#L300-L301. (We've defined `PID_MAX` [here](https://github.com/asterinas/asterinas/blob/0e3ef19ada8d6e833093a47a389d6390a54a4e01/kernel/src/process/posix_thread/mod.rs#L365)).
 2. We should include more information in warning messages. Ideally, we would print the cgroup path and the limit value. However, it is not possible to obtain the cgroup path here. Therefore, I only print the limit value in the suggested version.
    - Meanwhile, we should not print the warning messages if `max` is written.

kernel/src/fs/cgroupfs/controller/pids.rs
lrh2000: Thanks for the experiments! As we discussed (privately), I think we'll have to accept the dummy implementation due to the design difficulties in enforcing the limit correctly and the tight schedule before the release deadline.

PR #685, opened 2024-03-15
Format code for regression tests

lrh2000:
This PR uses `clang-format` to standardize the code formats of regression tests to the Linux kernel code style.

The choice of Linux kernel code style has already been [discussed on Zulip](https://asterinas.zulipchat.com/#narrow/stream/414044-general/topic/Code.20formatting.20for.20regression.20tests/near/422954697).

To pass the CI, the version of Asterinas container needs to be updated, as `clang-format` needs to be installed inside the docker.

PR #2593, opened 2025-11-18
Enforce `-Werror` for all tests (again) 

lrh2000:
This really helps. I suppose the first commit will show why. Our pty test is buggy, which makes our code buggy, too.

I originally added the `-Werror` flag to https://github.com/asterinas/asterinas/pull/685, but it no longer works. This PR resolves the warnings and enforces the warnings-as-errors rule again.

### DIFF
--- a/test/src/apps/devfs/full.c
+++ b/test/src/apps/devfs/full.c
@@ -37,8 +37,12 @@ FN_TEST(read)
 		 _ret == READ_SIZE &&
 			 memcmp(buffer, all_zeros, READ_SIZE) == 0);
 	TEST_RES(read(fd, buffer, 0), _ret == 0);
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnonnull"
 	TEST_ERRNO(read(fd, NULL, 1), EFAULT);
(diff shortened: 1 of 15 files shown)


test/src/apps/pty/close_pty.c
```diff
@@ -108,15 +107,16 @@ FN_TEST(close_master)
...
-	TEST_RES(poll(&pfd, 1, -1), pfd.revents == POLLIN | POLLOUT);
+	TEST_RES(poll(&pfd, 1, -1), pfd.revents == (POLLIN | POLLOUT));
 	TEST_RES(ioctl(slave, FIONREAD, &bytes), bytes == 1);
 
 	TEST_SUCC(close(master));
 	TEST_ERRNO(unlink(slave_name), ENOENT);
 
-	TEST_RES(poll(&pfd, 1, -1), pfd.revents == POLLOUT | POLLHUP);
+	TEST_RES(poll(&pfd, 1, -1),
+		 pfd.revents == (POLLIN | POLLOUT | POLLERR | POLLHUP));
```
StevenJiang1110, line 119: Thanks for fixing this problem! I haven't realized `==` will have the higher priority than `|`.

PR #2594, opened 2025-11-18
Support pty packet mode

StevenJiang1110:
This PR adds the basic support for pty packet mode, which is required by the terminal user program.

Packet mode can exclusively be enabled via the PTY master. Once activated, the PTY master gains the ability to read control messages in addition data.

1. In the absence of control messages, the master will read normal data, which will be prefixed with an additional leading 0 byte; 
2. Control messages are generated when the PTY slave modifies certain termios settings or performs specific operations (e.g., flushing buffers). This PR's implementation currently supports only PKT_NOSTOP, PKT_DOSTOP, and PKT_IOCTL control message types; other types are not yet supported. Each read operation from the PTY master consumes any pending control messages, returning them to the user before processing further data.

The current implementation in this PR diverges slightly from Linux's approach: Linux stores control messages within the PTY slave and packet mode settings within the PTY master. In contrast, this PR stores both entities within the PTY master. However, I think this difference will not lead to any user-observable variations in behavior.

More information abount packet mode can be seen at https://man7.org/linux/man-pages/man2/TIOCPKT.2const.html.

### DIFF
--- a/kernel/src/device/pty/driver.rs
+++ b/kernel/src/device/pty/driver.rs
@@ -5,7 +5,13 @@ use ostd::sync::SpinLock;
 
 use super::file::PtySlaveFile;
 use crate::{
-    device::tty::{Tty, TtyDriver, TtyFlags},
+    device::{
+        pty::packet::{PacketCtrl, PacketStatus},
+        tty::{
+            termio::{CCtrlCharId, CInputFlags, CLocalFlags, CTermios},
+            Tty, TtyDriver, TtyFlags,
+        },
+    },
     events::IoEvents,
     fs::inode_handle::FileIo,
     prelude::*,
@@ -25,6 +31,7 @@ pub struct PtyDriver {
     pollee: Pollee,
     opened_slaves: SpinLock<usize>,
     tty_flags: TtyFlags,
+    packet_ctrl: PacketCtrl,
 }
 
 /// A pseudoterminal slave.
@@ -39,6 +46,7 @@ impl PtyDriver {
             pollee: Pollee::new(),
             opened_slaves: SpinLock::new(0),
             tty_flags,
+            packet_ctrl: PacketCtrl::new(),
         }
     }
 
@@ -47,6 +55,27 @@ impl PtyDriver {
             return Ok(0);
         }
 
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/n_tty.c#L2245>.
+        match self.packet_ctrl.take_status() {
+            None => {
+                // Packet mode is disabled.
+                self.read_output(buf)
+            }
+            Some(packet_status) if !packet_status.is_empty() => {
+                // Some packet status is pending.
+                buf[0] = packet_status.bits();
+                Ok(1)
+            }
+            Some(_) => {
+                // There's no pending packet status.
+                let data_len = self.read_output(&mut buf[1..])?;
+                buf[0] = 0;
+                Ok(data_len + 1)
+            }
+        }
+    }
+
+    fn read_output(&self, buf: &mut [u8]) -> Result<usize> {
         let mut output = self.output.lock();
         if output.is_empty() {
             if self.tty_flags.is_other_closed() {
@@ -57,7 +86,6 @@ impl PtyDriver {
 
         let read_len = output.len().min(buf.len());
         output.pop_slice(&mut buf[..read_len]).unwrap();
-
         Ok(read_len)
     }
 
@@ -75,6 +103,10 @@ impl PtyDriver {
 
     pub(super) fn tty_flags(&self) -> &TtyFlags {
         &self.tty_flags
+    }
+
+    pub(super) fn packet_ctrl(&self) -> &PacketCtrl {
+        &self.packet_ctrl
     }
 }
 
@@ -110,7 +142,7 @@ impl TtyDriver for PtyDriver {
             len += 1;
         }
 
-        self.pollee.notify(IoEvents::IN);
+        self.pollee.notify(IoEvents::IN | IoEvents::RDNORM);
         Ok(len)
     }
 
@@ -129,7 +161,7 @@ impl TtyDriver for PtyDriver {
             }
 
             if !has_notified {
-                self.pollee.notify(IoEvents::IN);
+                self.pollee.notify(IoEvents::IN | IoEvents::RDNORM);
                 has_notified = true;
             }
         }
@@ -146,5 +178,41 @@ impl TtyDriver for PtyDriver {
 
     fn console(&self) -> Option<&dyn AnyConsoleDevice> {
         None
+    }
+
+    fn on_termios_change(&self, old_termios: &CTermios, new_termios: &CTermios) {
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L246>.
+        let extproc = old_termios.local_flags().contains(CLocalFlags::EXTPROC)
+            || new_termios.local_flags().contains(CLocalFlags::EXTPROC);
+        let old_flow = old_termios.input_flags().contains(CInputFlags::IXON)
+            && old_termios.special_char(CCtrlCharId::VSTOP) == 0o23
+            && old_termios.special_char(CCtrlCharId::VSTART) == 0o21;
+        let new_flow = new_termios.input_flags().contains(CInputFlags::IXON)
+            && new_termios.special_char(CCtrlCharId::VSTOP) == 0o23
+            && new_termios.special_char(CCtrlCharId::VSTART) == 0o21;
+
+        if (old_flow == new_flow) && !extproc {
+            return;
+        }
+
+        let has_set = self.packet_ctrl.set_status(|packet_status| {
+            if old_flow != new_flow {
+                *packet_status &= !(PacketStatus::DOSTOP | PacketStatus::NOSTOP);
+                if new_flow {
+                    *packet_status |= PacketStatus::DOSTOP;
+                } else {
+                    *packet_status |= PacketStatus::NOSTOP;
+                }
+            }
+
+            if extproc {
+                *packet_status |= PacketStatus::IOCTL;
+            }
+        });
+
+        if has_set {
+            self.pollee
+                .notify(IoEvents::PRI | IoEvents::IN | IoEvents::RDNORM);
+        }
     }
 }
--- a/kernel/src/device/pty/master.rs
+++ b/kernel/src/device/pty/master.rs
@@ -63,7 +63,7 @@ impl PtyMaster {
         let mut events = IoEvents::empty();
 
         if self.slave().driver().buffer_len() > 0 {
-            events |= IoEvents::IN;
+            events |= IoEvents::IN | IoEvents::RDNORM;
         }
 
         if self.slave().can_push() {
@@ -72,6 +72,12 @@ impl PtyMaster {
 
         if self.master_flags().is_other_closed() {
             events |= IoEvents::HUP;
+        }
+
+        // Deal with packet mode.
+        let packet_ctrl = self.slave.driver().packet_ctrl();
+        if packet_ctrl.has_status() {
+            events |= IoEvents::PRI | IoEvents::IN | IoEvents::RDNORM;
         }
 
         events
@@ -144,6 +150,9 @@ mod ioctl_defs {
     pub(super) type GetPtyLock   = ioc!(TIOCGPTLCK,  b'T', 0x39, OutData<i32>);
 
     pub(super) type OpenPtySlave = ioc!(TIOCGPTPEER, b'T', 0x41, NoData);
+
+    pub(super) type SetPktMode   = ioc!(TIOCPKT,     0x5420,     InData<i32>);
+    pub(super) type GetPktMode   = ioc!(TIOCGPKT,    b'T', 0x38, OutData<i32>);
 }
 
 impl FileIo for PtyMaster {
@@ -222,6 +231,20 @@ impl FileIo for PtyMaster {
                 let len = self.slave.driver().buffer_len() as i32;
 
                 cmd.write(&len)?;
+            }
+            cmd @ SetPktMode => {
+                let new_mode = cmd.read()? != 0;
+
+                self.slave.driver().packet_ctrl().set_mode(new_mode);
+            }
+            cmd @ GetPktMode => {
+                let packet_mode = if self.slave.driver().packet_ctrl().mode() {
+                    1
+                } else {
+                    0
+                };
+
+                cmd.write(&packet_mode)?;
             }
 
             _ => (self.slave.clone() as Arc<dyn Terminal>).job_ioctl(raw_ioctl, true)?,
--- a/kernel/src/device/pty/packet.rs
+++ b/kernel/src/device/pty/packet.rs
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use core::sync::atomic::{AtomicBool, Ordering};
+
+use crate::prelude::*;
+
+/// Control states related to packet mode.
+///
+/// Reference: <https://man7.org/linux/man-pages/man2/TIOCPKT.2const.html>.
+pub(super) struct PacketCtrl {
+    mode: AtomicBool,
+    status: SpinLock<PacketStatus>,
+}
+
+impl PacketCtrl {
+    // Creates a new `PacketCtrl`.
+    pub(super) fn new() -> Self {
+        Self {
+            mode: AtomicBool::new(false),
+            status: SpinLock::new(PacketStatus::empty()),
+        }
+    }
+
+    /// Returns whether packet mode is enabled.
+    pub(super) fn mode(&self) -> bool {
+        self.mode.load(Ordering::Relaxed)
+    }
+
+    /// Sets whether packet mode is enabled.
+    pub(super) fn set_mode(&self, mode: bool) {
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L158>.
+        let mut status = self.status.lock();
+
+        let old_mode = self.mode.load(Ordering::Relaxed);
+        self.mode.store(mode, Ordering::Relaxed);
+
+        if mode && !old_mode {
+            *status = PacketStatus::empty();
+        }
+    }
+
+    /// Sets packet status if packet mode is enabled.
+    pub(super) fn set_status(&self, set_packet_status: impl FnOnce(&mut PacketStatus)) -> bool {
+        // Fast path: Packet mode is disabled.
+        if !self.mode.load(Ordering::Relaxed) {
+            return false;
+        }
+
+        // Packet mode is enabled.
+        let mut packet_status = self.status.lock();
+        set_packet_status(&mut packet_status);
+        true
+    }
+
+    /// Checks if packet mode is enabled and there is pending packet status.
+    pub(super) fn has_status(&self) -> bool {
+        // Fast path: Packet mode is disabled.
+        if !self.mode.load(Ordering::Relaxed) {
+            return false;
+        }
+
+        // Packet mode is enabled.
+        !self.status.lock().is_empty()
+    }
+
+    /// Takes out packet status if packet mode is enabled.
+    pub(super) fn take_status(&self) -> Option<PacketStatus> {
+        // Fast path: Packed mode is disabled.
+        if !self.mode.load(Ordering::Relaxed) {
+            return None;
+        }
+
+        // Packet mode is enabled.
+        let mut status = self.status.lock();
+        let old_status = *status;
+        *status = PacketStatus::empty();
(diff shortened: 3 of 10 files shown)


kernel/src/device/tty/driver.rs
```diff
@@ -58,4 +63,13 @@ pub trait TtyDriver: Send + Sync + 'static {
...
     fn console(&self) -> Option<&dyn AnyConsoleDevice>;
+
+    /// Returns the packet mode control information.
+    ///
+    /// This method will only return `Some` for pty;
+    /// other TTY types will return `None`.
+    fn packet_ctrl(&self) -> Option<&PacketCtrl>;
+
+    /// Notifies that some events has happened.
+    fn notify_events(&self, events: IoEvents);
```
lrh2000: I do not really find these methods make sense. I think provide a `on_termious_change` callback is enough?

Then the pty logic (`pty_set_termios`) can be moved to the `pty` module.

Besides, if `packet_ctrl` belongs to the `pty` module, I don't think it needs to return an `Option`.

kernel/src/device/tty/line_discipline.rs
```diff
@@ -235,7 +236,13 @@ impl LineDiscipline {
         &self.termios
     }
 
-    pub fn set_termios(&mut self, termios: CTermios) {
+    pub fn set_termios(
+        &mut self,
+        mut termios: CTermios,
+        packet_mode_set_termios: impl Fn(&CTermios, &mut CTermios),
+    ) {
+        packet_mode_set_termios(&self.termios, &mut termios);
```
lrh2000: Why should it be called here? You can call it outside `set_termious` as well, so there is no need to pass a closure to `set_termios`?

kernel/src/device/pty/driver.rs
```diff
@@ -47,6 +52,29 @@ impl PtyDriver {
             return Ok(0);
         }
 
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/n_tty.c#L2245>
+        if self.packet_ctrl.mode() {
+            let mut packet_status = self.packet_ctrl.status().lock();
```
lrh2000: Do we really need to hold the lock during `read_output`? I think we just need to have `packet_ctrl.take_status()`.

kernel/src/device/pty/driver.rs
```diff
@@ -47,6 +52,29 @@ impl PtyDriver {
...
+                    self.read_output(&mut buf[1..])?
+                } else {
+                    0
+                };
+                buf[0] = 0;
+                return Ok(read_len + 1);
+            }
+        }
+
+        self.read_output(buf)
```
lrh2000: I think we can return earlier. The non-packet mode should be in the `if` block.

test/src/apps/pty/pty_packet_mode.c
```diff
@@ -0,0 +1,179 @@
...
+	int packet_mode;
+	TEST_RES(ioctl(master, TIOCGPKT, &packet_mode), packet_mode == 0);
+	TEST_ERRNO(ioctl(slave, TIOCGPKT, &packet_mode), ENOTTY);
+
+	packet_mode = 1;
+	TEST_SUCC(ioctl(master, TIOCPKT, &packet_mode));
+	TEST_ERRNO(ioctl(slave, TIOCPKT, &packet_mode), ENOTTY);
+
+	TEST_RES(ioctl(master, TIOCGPKT, &packet_mode), packet_mode == 1);
+	TEST_ERRNO(ioctl(slave, TIOCPKT, &packet_mode), ENOTTY);
```
lrh2000: This wants to mean `TIOCGPKT`?

kernel/src/device/pty/master.rs
```diff
@@ -74,6 +74,18 @@ impl PtyMaster {
...
+        // Deal with packet mode
+        let Some(packet_ctrl) = self.slave.driver().packet_ctrl() else {
+            return events;
+        };
+        if !packet_ctrl.mode() {
+            return events;
+        }
+        let status = packet_ctrl.status().lock();
+        if !status.is_empty() {
+            events |= IoEvents::PRI | IoEvents::IN | IoEvents::RDNORM;
```
lrh2000, line 80: Could you please add `PRI`/`RDNORM` to existing pty tests, like:
https://github.com/asterinas/asterinas/blob/a176e4c3c6f40a3c54ea3ed4d471ba41cf736daf/test/src/apps/pty/close_pty.c#L15

Then confirm whether these events will appear in other cases?

(To me, `PRI` sounds related to `packet_ctrl`, but `RDNORM` sounds unrelated.)

lrh2000 (commented):
Thanks for adding support for pty packet mode!

kernel/src/device/pty/master.rs
StevenJiang1110: Oh, yes, thanks for pointing that out! It appears that for TTYs (and consequently PTYs), POLLRDNORM is consistently included whenever a POLLIN event is present. I've updated the test accordingly to reflect this behavior.

https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/n_tty.c#L2452
https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/n_tty.c#L2459

Furthermore, it seems we might also be missing the POLLWRNORM event when POLLOUT events exist. While this particular issue doesn't affect PTYs, it does present a similar problem for sockets. This broader problem could be addressed in a future PR.

kernel/src/device/pty/master.rs
```diff
@@ -144,6 +154,9 @@ mod ioctl_defs {
     pub(super) type GetPtyLock   = ioc!(TIOCGPTLCK,  b'T', 0x39, OutData<i32>);
 
     pub(super) type OpenPtySlave = ioc!(TIOCGPTPEER, b'T', 0x41, NoData);
+
+    pub(super) type SetPktMode = ioc!(TIOCPKT, 0x5420, InData<i32>);
+    pub(super) type GetPktMode = ioc!(TIOCGPKT, b'T', 0x38, OutData<i32>);
```
lrh2000: Align vertically:
```suggestion
    pub(super) type SetPktMode   = ioc!(TIOCPKT,     0x5420,     InData<i32>);
    pub(super) type GetPktMode   = ioc!(TIOCGPKT,    b'T', 0x38, OutData<i32>);
```

kernel/src/device/pty/master.rs
```diff
@@ -223,7 +236,18 @@ impl FileIo for PtyMaster {
...
+            }
+            cmd @ GetPktMode => {
+                let packet_mode = if self.slave.driver().packet_ctrl().mode() {
+                    1
+                } else {
+                    0
+                };
+                cmd.write(&packet_mode)?;
+            }
             _ => (self.slave.clone() as Arc<dyn Terminal>).job_ioctl(raw_ioctl, true)?,
```
lrh2000, line 250: Keep the original blank line. Add blank lines after `cmd.read`/before `cmd.write`, like other methods.
```rust
            cmd @ SetPktMode => {
                let new_mode = cmd.read()? != 0;

                self.slave.driver().packet_ctrl().set_mode(new_mode);
            }
            cmd @ GetPktMode => {
                let packet_mode = if self.slave.driver().packet_ctrl().mode() {
                    1
                } else {
                    0
                };

                cmd.write(&packet_mode)?;
            }

            _ => (self.slave.clone() as Arc<dyn Terminal>).job_ioctl(raw_ioctl, true)?,
```

kernel/src/device/tty/driver.rs
```diff
@@ -58,4 +62,7 @@ pub trait TtyDriver: Send + Sync + 'static {
     ///
     /// If the TTY is not associated with any console device, this method will return `None`.
     fn console(&self) -> Option<&dyn AnyConsoleDevice>;
+
+    /// The method will be called when the TTY termios is changed.
```
lrh2000:
```suggestion
    /// Notifies that the TTY termios is changed.
    ///
    /// This method will be called with a spin lock held, so it cannot break atomic mode.
```

kernel/src/device/tty/mod.rs
```diff
@@ -37,6 +37,7 @@ mod termio;
 pub use device::SystemConsole;
 pub use driver::TtyDriver;
 pub(super) use flags::TtyFlags;
+pub(in crate::device) use termio::{CCtrlCharId, CInputFlags, CLocalFlags, CTermios};
```
lrh2000: `pub(in crate::device)` is `pub(super)`?

Besides, we can make the `termio` module `pub(super)`?

kernel/src/device/pty/driver.rs
```diff
@@ -47,6 +52,28 @@ impl PtyDriver {
             return Ok(0);
         }
 
+        if !self.packet_ctrl.mode() {
+            return self.read_output(buf);
+        }
+
+        // Deal with packet mode.
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/n_tty.c#L2245>
+        let packet_status = self.packet_ctrl.take_status();
```
lrh2000: Is it possible to do something like the following?
```rust
fn take_status(&self) -> Option<PacketStatus> {
    // Fast path: We're not in the packet mode.
    if !self.mode.load(Ordering::Relaxed) {
        return None;
    }

    let status = self.status.lock();

    // Check again to avoid race conditions.
    if self.mode.load(Ordering::Relaxed) {
        return None;
    }

    // ..
}
```

If the race condition won't cause any problems, we can remove the second check. It doesn't seem like there are any obvious problems caused by removing that check. Do you know if it's necessary?

But even without that check, doing so would encapsulate the logic in the `PacketCtrl` type. I think we should remove the `mode()` method.

kernel/src/device/pty/driver.rs
```diff
@@ -147,4 +177,44 @@ impl TtyDriver for PtyDriver {
     fn console(&self) -> Option<&dyn AnyConsoleDevice> {
         None
     }
+
+    fn on_termios_change(&self, old_termios: &CTermios, new_termios: &CTermios) {
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L246>.
+
+        let packet_ctrl = &self.packet_ctrl;
+
+        if !packet_ctrl.mode() {
```
lrh2000: Similar to what I said above, I think `status()`/`mode()` method should also be removed. Could we move this logic to `packet_ctrl`?

kernel/src/device/pty/master.rs
```diff
@@ -74,6 +74,16 @@ impl PtyMaster {
             events |= IoEvents::HUP;
         }
 
+        // Deal with packet mode
```
lrh2000:
```suggestion
        // Deal with packet mode.
```

kernel/src/device/pty/master.rs
```diff
@@ -74,6 +74,16 @@ impl PtyMaster {
...
+        let packet_ctrl = self.slave.driver().packet_ctrl();
+        if !packet_ctrl.mode() {
+            return events;
+        }
+        let status = packet_ctrl.status().lock();
+        if !status.is_empty() {
```
lrh2000:
```suggestion
        if packet_ctrl.has_status() {
```

kernel/src/device/pty/packet.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    pub(super) fn mode(&self) -> bool {
+        self.mode.load(Ordering::Relaxed)
+    }
+
+    pub(super) fn set_mode(&self, mode: bool) {
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L158>.
+        let mut status = self.status.lock();
+
+        let old_mode = self.mode.swap(mode, Ordering::Relaxed);
```
lrh2000: `mode` will always be changed with `status` lock held. So I guess we can do this (it's faster because `swap` is slow):
```rust
let old_mode = self.mode.load(Ordering::Relaxed);
self.mode.store(mode, Ordering::Relaxed);

if mode && !old_mode {
    *status = PacketStatus::empty();
}
```

test/src/apps/pty/pty_packet_mode.c
```diff
@@ -0,0 +1,180 @@
...
+
+FN_TEST(close_and_reopen_slave)
+{
+	// Closing pty slave will not reset the packet status.
+
+	struct termios term;
+	TEST_SUCC(tcgetattr(slave, &term));
+	term.c_lflag |= EXTPROC;
+	TEST_SUCC(tcsetattr(slave, TCSANOW, &term));
+	TEST_RES(poll(&pfd, 1, -1),
```
lrh2000, line 159: Insert a blank line above to be consistent with other tests.

test/src/apps/pty/pty_packet_mode.c
```diff
@@ -0,0 +1,180 @@
...
+
+	char read_buf[128] = { 0 };
+	TEST_RES(read(master, read_buf, sizeof(read_buf)),
+		 _ret == 1 && read_buf[0] == TIOCPKT_DOSTOP);
+}
+END_TEST()
+
+FN_TEST(close_and_reopen_slave)
+{
+	// Closing pty slave will not reset the packet status.
```
lrh2000:
```suggestion
	// Closing the pty slave will not reset the packet status.
```

kernel/src/device/pty/driver.rs
```diff
@@ -47,6 +52,28 @@ impl PtyDriver {
...
+        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/n_tty.c#L2245>
+        let packet_status = self.packet_ctrl.take_status();
+        if !packet_status.is_empty() {
+            buf[0] = packet_status.bits();
+            Ok(1)
+        } else {
+            let data_len = if buf.len() > 1 {
+                self.read_output(&mut buf[1..])?
+            } else {
+                0
```
lrh2000: Should we return `EIO`/`EAGAIN` in this case? If so, should we use `self.read_output(&mut buf[1..])` even if `buf.len() == 1`?

kernel/src/device/pty/master.rs
lrh2000: Thanks for the investigation!

kernel/src/device/pty/packet.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use core::sync::atomic::{AtomicBool, Ordering};
+
+use ostd::sync::LocalIrqDisabled;
+
+use crate::prelude::*;
+
+pub(super) struct PacketCtrl {
+    mode: AtomicBool,
+    status: SpinLock<PacketStatus, LocalIrqDisabled>,
```
lrh2000: Is it necessary to disable IRQs here? I don't think pty has anything to do with interrupts. TTY needs to disable interrupts because input comes from keyboard interrupts.

kernel/src/device/pty/driver.rs
StevenJiang1110: > If the race condition won't cause any problems, we can remove the second check. It doesn't seem like there are any obvious problems caused by removing that check. Do you know if it's necessary?

I also think the race will cause any real problems. So I prefer removing the second check.

After checking the mode and before lock status, other threads may concurrently change the packet mode to false, in this case, the packet status to read will be losted. But I think kernel is not responsible to handle such cases.

kernel/src/device/pty/driver.rs
```diff
@@ -147,4 +179,37 @@ impl TtyDriver for PtyDriver {
...
+                if extproc {
+                    *packet_status |= PacketStatus::IOCTL;
+                }
+            }
+        });
+
+        self.pollee
+            .notify(IoEvents::PRI | IoEvents::IN | IoEvents::RDNORM);
+    }
 }
```
lrh2000, line 218: I think this version is better?
```rust
        let extproc = old_termios.local_flags().contains(CLocalFlags::EXTPROC)
            || new_termios.local_flags().contains(CLocalFlags::EXTPROC);
        let old_flow = old_termios.input_flags().contains(CInputFlags::IXON)
            && old_termios.special_char(CCtrlCharId::VSTOP) == 0o23
            && old_termios.special_char(CCtrlCharId::VSTART) == 0o21;
        let new_flow = new_termios.input_flags().contains(CInputFlags::IXON)
            && new_termios.special_char(CCtrlCharId::VSTOP) == 0o23
            && new_termios.special_char(CCtrlCharId::VSTART) == 0o21;

        if (old_flow == new_flow) && !extproc {
            return;
        }

        let has_set = self.packet_ctrl.set_status(|packet_status| {
            if old_flow != new_flow {
                *packet_status &= !(PacketStatus::DOSTOP | PacketStatus::NOSTOP);
                if new_flow {
                    *packet_status |= PacketStatus::DOSTOP;
                } else {
                    *packet_status |= PacketStatus::NOSTOP;
                }
            }

            if extproc {
                *packet_status |= PacketStatus::IOCTL;
            }
        });

        if has_set {
            self.pollee
                .notify(IoEvents::PRI | IoEvents::IN | IoEvents::RDNORM);
        }

File: kernel/src/device/pty/packet.rs [hunk 21]
```diff
@@ -0,0 +1,82 @@
...
+        if !self.mode.load(Ordering::Relaxed) {
+            return;
+        }
+
+        // Packet mode is enabled.
+        let mut packet_status = self.status.lock();
+        set_packet_status(&mut packet_status);
+    }
+
+    pub fn has_status(&self) -> bool {
```
lrh2000:
```suggestion
    pub(super) fn has_status(&self) -> bool {
```

kernel/src/device/pty/packet.rs
```diff
@@ -0,0 +1,82 @@
...
+        !self.status.lock().is_empty()
+    }
+
+    pub(super) fn take_status(&self) -> Option<PacketStatus> {
+        // Fast path: Packed mode is disabled.
+        if !self.mode.load(Ordering::Relaxed) {
+            return None;
+        }
+
+        let mut status = self.status.lock();
```
lrh2000, line 74:
```rust
        // Packet mode is enabled.
        let mut status = self.status.lock();
```

kernel/src/device/pty/packet.rs
```diff
@@ -0,0 +1,82 @@
...
+        let old_status = *status;
+        *status = PacketStatus::empty();
+        Some(old_status)
+    }
```
lrh2000, line 78: Add some doc comments here?
```patch
diff --git a/kernel/src/device/pty/packet.rs b/kernel/src/device/pty/packet.rs
index f235aec51..749a39d62 100644
--- a/kernel/src/device/pty/packet.rs
+++ b/kernel/src/device/pty/packet.rs
@@ -4,6 +4,9 @@ use core::sync::atomic::{AtomicBool, Ordering};
 
 use crate::prelude::*;
 
+/// Control states related to packet mode.
+///
+/// Reference: <https://man7.org/linux/man-pages/man2/TIOCPKT.2const.html>.
 pub(super) struct PacketCtrl {
     mode: AtomicBool,
     status: SpinLock<PacketStatus>,
@@ -17,10 +20,12 @@ impl PacketCtrl {
         }
     }
 
+    /// Returns whether packet mode is enabled.
     pub(super) fn mode(&self) -> bool {
         self.mode.load(Ordering::Relaxed)
     }
 
+    /// Sets whether packet mode is enabled.
     pub(super) fn set_mode(&self, mode: bool) {
         // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L158>.
         let mut status = self.status.lock();
@@ -33,6 +38,7 @@ impl PacketCtrl {
         }
     }
 
+    /// Sets packet status if packet mode is enabled.
     pub(super) fn set_status(&self, set_packet_status: impl FnOnce(&mut PacketStatus)) {
         // Fast path: Packet mode is disabled.
         if !self.mode.load(Ordering::Relaxed) {
@@ -44,6 +50,7 @@ impl PacketCtrl {
         set_packet_status(&mut packet_status);
     }
 
+    /// Checks if packet mode is enabled and there is pending packet status.
     pub fn has_status(&self) -> bool {
         // Fast path: Packet mode is disabled.
         if !self.mode.load(Ordering::Relaxed) {
@@ -54,6 +61,7 @@ impl PacketCtrl {
         !self.status.lock().is_empty()
     }
 
+    /// Takes out packet status if packet mode is enabled.
     pub(super) fn take_status(&self) -> Option<PacketStatus> {
         // Fast path: Packed mode is disabled.
         if !self.mode.load(Ordering::Relaxed) {
@@ -68,6 +76,8 @@ impl PacketCtrl {
 }
 
 bitflags! {
+    /// Control status information in packet mode.
+    ///
     /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/ioctls.h#L110>.
     pub(super) struct PacketStatus: u8 {
         const DATA = 0;
```

kernel/src/device/pty/driver.rs
lrh2000: Given that re-enabling packet mode always clears the packet status, I agree that the current code seems fine.

Issue #2224, opened 2025-06-30
Add the support for LoongArch


wyt8:
根据之前的[建议](https://github.com/asterinas/asterinas/pull/2202)，我将提交多个commit来完成对LoongArch的支持。每个提交都会记录到这个Issues中。

- [ ] OSDK中添加对LoongArch的支持。
- [ ] OSTD中添加对LoongArch的支持。
- [ ] Kernel中添加对LoongArch的支持。

根据[仓库](https://github.com/oscomp/asterinas)的要求，我的提交时间有点晚。且个人时间不是很充裕，望见谅，但我会尽可能加强代码规范，争取可以合并到主仓库中。

lrh2000:
@wyt8 Could you please tell me the correct steps for successfully booting Asterinas and enabling it to handle user input in LoongArch, as you mentioned in https://github.com/asterinas/asterinas/pull/2260#discussion_r2230303438?

As far as I know, the command `make run ARCH=loongarch64` won't work without https://github.com/asterinas/asterinas/pull/2308. Did you use `make run` when performing tests in #2260? If not, could you share your test steps?

With #2308, `make run ARCH=loongarch64 LOG_LEVEL=debug` stops at:
```text
INFO: Component initializing:ComponentInfo { name: "aster-network", path: "/root/asterinas/kernel/comps/network", priority: 1 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-logger", path: "/root/asterinas/kernel/comps/logger", priority: 1 }
```

I think the VirtIO console isn't working properly so there is no output after initializing `aster-logger` (following that, logs will be printed via the VirtIO console).

I haven't debugged further, but given the claim in https://github.com/asterinas/asterinas/pull/2260#discussion_r2230303438, I think you should have some ideas here?

wyt8:
1. 启动命令为：`make run ARCH=loongarch64`
2. 注释掉这行[代码](https://github.com/asterinas/asterinas/blob/main/ostd/src/mm/dma/dma_stream.rs#L157)，否则在输出时就会陷入到死锁中：这个需要后续编写相应的代码
3. 需要编写一个loongarch版的initramfs.cpio，可以临时使用这个[文件](https://github.com/wyt8/BITOS/blob/loongarch-new/ostd/src/arch/loongarch/boot/initramfs.cpio)，它包含了一个busybox：好像nix不支持loongarch架构

lrh2000:
Thanks for the information!

[...]

PR #2595, opened 2025-11-18
Revise `PageTableEntryTrait`

junyang-zh:
## The problem

The former `PageTableEntryTrait` contains the following methods:

```rust
    fn new_absent() -> Self;
    fn is_present(&self) -> bool;
    fn new_page(paddr: Paddr, level: PagingLevel, prop: PageProperty) -> Self;
    fn new_pt(paddr: Paddr) -> Self;
    fn paddr(&self) -> Paddr;
    fn prop(&self) -> PageProperty;
    fn set_prop(&mut self, prop: PageProperty);
    fn is_last(&self, level: PagingLevel) -> bool;
```

This is problematic in many ways. The implementor must ensure the following properties that are hard to check:
 - for PTEs created by `new_absent`, `is_present` returns false; for PTEs created by `new_page` or `new_pt` (and possibly modified with `set_prop` later), `is_present` returns true;
 - for PTEs created with `new_page`, `is_last` should return true, for PTEs created with `new_pt` it is the contrary; also it should not be affected by later `set_prop`s;
 - `prop` should return the last value used by `set_prop` or that value used to create the PTE;
 - ...

Actually, `PageTableEntryTrait` should be `unsafe` since the implementer should uphold the above properties.

Furthermore, there are unexplained questions:
 - What will happen if we call `is_last`/`paddr` on absent entries?
 - What will happen if we call `set_prop`/`prop` on PTEs that point to child page tables?
 - ...
 
## The solution

Actually, what we want from the architectural-specific PTE definition is something like this:

```rust
/// The scalar representation of a page table entry (PTE).
///
/// This is an architecture-agnostic representation that can be converted
/// to/from architecture-specific PTEs via the [`PteTrait`]. This is a scalar
/// value that can be cloned or compared, and does not own the underlying page
/// table node or mapped item if present.
#[derive(Debug, Clone, PartialEq, Eq)]
pub(crate) enum PteScalar {
    /// A PTE that is considered absent by the MMU.
    Absent,
    /// A PTE that points to the next-level page table.
    ///
    /// The level is the level of the current page table entry.
    PageTable(Paddr, PagingLevel, PageTableFlags),
    /// A PTE that establishes the mapping to a physical frame.
    ///
    /// The level is the level of the current page table entry.
    Mapped(Paddr, PagingLevel, PageProperty),
}
```

So we can avoid reasoning about incompatibilities, such as calling `prop` on absent PTEs. Also, the "preserving" properties (`fn x` should return `y` for PTEs created with `fn z`) can be unified into one rule that is easy to check.

The trait now only contains two methods that cast PTEs to and from the representation above, plus the additional `from_usize` and `into_usize` utilities.

### Comparing with `Child` and `PageTableConfig::Item`

`Child` and `PageTableConfig::Item` are similar to `PageTableEntryRepr`. We now have many duplicate abstractions over a PTE. However, I think there are valid reasons:
 - Comparing with `PageTableEntryRepr`, `Child` makes the child page table managed by the Rust typing system rather than managed manually with raw physical addresses. This is convenient for `Cursor` operations.
 - Comparing with `Child`, `PageTableConfig::Item` makes the mapped item managed by the Rust typing system rather than (...same as above).

I have made `Child` own the mapped item in #2605 .

### Relation to CortenMM

What CortenMM needs is to add another variant to all `PageTableEntryRepr`, `Child`, and `PageTableConfig::Item`, that is `AbsentWith(Metadata)` (maybe an awkward name). This is similar to `PageTableEntryRepr::Absent`, which is not recognized by the MMU, but stores software-defined metadata in other bits rather than `PRESENT` or `VALID` bits. Without this PR, the CortenMM implementation looks awful...

However, this PR is motivated enough with or without CortenMM. It just makes things better.

## Unresolved questions

 1. Naming of the `Repr` enum, its variants, and other related functions. Could be improved. Could Copilot give some ideas?
 2. https://github.com/asterinas/asterinas/pull/2570#issuecomment-3530728083, https://github.com/asterinas/asterinas/pull/2570#issuecomment-3531261066, and https://github.com/asterinas/asterinas/pull/2570#issuecomment-3531444235. However, resolving them would be easier with this PR. I would do it after this PR.
 3. What about the weird `level` parameters? The level cannot be encoded in the entry, and the level determines the PTE format. This PR hasn't neatly sorted it out.

--

Edits:
 - changed the naming of enum variants to unify the naming of `Child` and `Repr`.

### DIFF
--- a/ostd/src/mm/page_table/boot_pt.rs
+++ b/ostd/src/mm/page_table/boot_pt.rs
@@ -168,11 +168,14 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
         // Walk to the last level of the page table.
         while level > 1 {
             let index = pte_index::<C>(from, level);
+            // SAFETY: The result pointer is within the PT frame.
             let pte_ptr = unsafe { (paddr_to_vaddr(pt) as *mut E).add(index) };
+            // SAFETY: The pointer to the entry is valid to read.
             let pte = unsafe { pte_ptr.read() };
             match pte.to_repr(level) {
                 PteScalar::Absent => {
                     let (pte, child_pt) = self.alloc_child(level);
+                    // SAFETY: The pointer to the entry is valid to write.
                     unsafe { pte_ptr.write(pte) };
                     pt = child_pt;
                 }
@@ -187,11 +190,14 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
         }
         // Map the page in the last level page table.
         let index = pte_index::<C>(from, 1);
+        // SAFETY: The result pointer is within the PT frame.
         let pte_ptr = unsafe { (paddr_to_vaddr(pt) as *mut E).add(index) };
+        // SAFETY: The pointer to the entry is valid to read.
         let pte = unsafe { pte_ptr.read() };
         if matches!(pte.to_repr(1), PteScalar::Mapped(_, _)) {
             panic!("mapping an already mapped page in the boot page table");
         }
+        // SAFETY: The pointer to the entry is valid to write.
         unsafe { pte_ptr.write(E::from_repr(&PteScalar::Mapped(to, prop), 1)) };
     }
 
@@ -219,7 +225,9 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
         // Walk to the last level of the page table.
         while level > 1 {
             let index = pte_index::<C>(virt_addr, level);
+            // SAFETY: The result pointer is within the PT frame.
             let pte_ptr = unsafe { (paddr_to_vaddr(pt) as *mut E).add(index) };
+            // SAFETY: The pointer to the entry is valid to read.
             let pte = unsafe { pte_ptr.read() };
             match pte.to_repr(level) {
                 PteScalar::Absent => {
@@ -232,7 +240,9 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
                     // Split the huge page.
                     let (child_pte, child_frame_pa) = self.alloc_child(level);
                     for i in 0..nr_subpage_per_huge::<C>() {
+                        // SAFETY: The result pointer is within the new PT frame.
                         let nxt_ptr = unsafe { (paddr_to_vaddr(child_frame_pa) as *mut E).add(i) };
+                        // SAFETY: The pointer to the entry is valid to write.
                         unsafe {
                             nxt_ptr.write(E::from_repr(
                                 &PteScalar::Mapped(huge_pa + i * C::BASE_PAGE_SIZE, prop),
@@ -240,6 +250,7 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
                             ))
                         };
                     }
+                    // SAFETY: The pointer to the entry is valid to write.
                     unsafe { pte_ptr.write(child_pte) };
                     pt = child_frame_pa;
                 }
@@ -248,12 +259,15 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
         }
         // Do protection in the last level page table.
         let index = pte_index::<C>(virt_addr, 1);
+        // SAFETY: The result pointer is within the PT frame.
         let pte_ptr = unsafe { (paddr_to_vaddr(pt) as *mut E).add(index) };
+        // SAFETY: The pointer to the entry is valid to read.
         let pte = unsafe { pte_ptr.read() };
         let PteScalar::Mapped(pa, mut prop) = pte.to_repr(1) else {
             panic!("protecting an unmapped page in the boot page table");
         };
         op(&mut prop);
+        // SAFETY: The pointer to the entry is valid to write.
         unsafe { pte_ptr.write(E::from_repr(&PteScalar::Mapped(pa, prop), 1)) };
     }
 
@@ -273,6 +287,7 @@ impl<E: PteTrait, C: PagingConstsTrait> BootPageTable<E, C> {
 
         // Zero it out.
         let vaddr = paddr_to_vaddr(frame_paddr) as *mut u8;
+        // SAFETY: The allocated frame is valid to write.
         unsafe { core::ptr::write_bytes(vaddr, 0, PAGE_SIZE) };
 
         (
@@ -304,11 +319,17 @@ fn dfs_walk_on_leave<E: PteTrait, C: PagingConstsTrait>(
 ) {
     if level >= 2 {
         let pt_vaddr = paddr_to_vaddr(pt) as *mut E;
-        let pt = unsafe { core::slice::from_raw_parts_mut(pt_vaddr, nr_subpage_per_huge::<C>()) };
-        for pte in pt {
+        for offset in 0..nr_subpage_per_huge::<C>() {
+            // SAFETY: The result pointer is within the PT frame.
+            let pte_ptr = unsafe { pt_vaddr.add(offset) };
+            // SAFETY: The pointer to the entry is valid to read.
+            let mut pte = unsafe { pte_ptr.read() };
+
             if let PteScalar::PageTable(child_pt, flags) = pte.to_repr(level) {
                 dfs_walk_on_leave::<E, C>(child_pt, level - 1, op);
-                op(pte, child_pt, level, flags);
+                op(&mut pte, child_pt, level, flags);
+                // SAFETY: The pointer to the entry is valid to write.
+                unsafe { pte_ptr.write(pte) };
             }
         }
     }


ostd/src/mm/page_table/mod.rs
```diff
@@ -491,71 +496,70 @@ pub(super) unsafe fn page_walk<C: PageTableConfig>(
...
-    ///
-    /// The physical address recorded in the PTE is either:
-    /// - the physical address of the next-level page table, or
-    /// - the physical address of the page that the PTE maps to.
-    fn paddr(&self) -> Paddr;
-
-    /// Returns the page property of the PTE.
-    fn prop(&self) -> PageProperty;
+    /// The level is the level of the current page table entry.
+    Mapped(Paddr, PagingLevel, PageProperty),
```
lrh2000: I think the API will look better if we can remove `PagingLevel` from `PteScalar`.
```rust
    fn from_repr(repr: &PteScalar, paging_level: PagingLevel) -> Self;
    fn to_repr(&self, paging_level: PagingLevel) -> PteScalar;
```

Otherwise, `to_repr` duplicates `paging_level` in its input arguments and output result, which sounds odd to me.

I checked all the call sites of `to_repr`, and it seems that the paging level in the result is never used.

ostd/src/arch/loongarch/mm/mod.rs
```diff
@@ -110,10 +111,6 @@ pub(crate) unsafe fn sync_dma_range(_range: Range<Vaddr>, _direction: DmaDirecti
     unimplemented!("DMA synchronization is unimplemented in LoongArch64")
 }
 
-#[derive(Clone, Copy, Pod, Default)]
-#[repr(C)]
-pub(crate) struct PageTableEntry(usize);
-
 /// Activates the given root-level page table.
```
lrh2000, line 123: Consistency: If you decide to move `PageTableEntry` below `activate_page_table`/`current_page_table_paddr`, please do so for all architectures. Otherwise, revert this change for `LoongArch`.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -166,141 +167,152 @@ pub(crate) fn current_page_table_paddr() -> Paddr {
...
+    fn to_repr(&self, level: PagingLevel) -> PteScalar {
+        if !self.is_present() {
+            return PteScalar::Absent;
+        }
+
+        if level == 1 || (self.0 & PteFlags::HUGE.bits() != 0) {
+            let pa_mask = Self::pa_mask_at_level(level);
+            let paddr = self.0 & pa_mask;
+            let prop = self.prop();
+            PteScalar::Mapped(paddr, level, prop)
```
lrh2000:
```suggestion
            let paddr = self.0 & Self::pa_mask_at_level(level);
            PteScalar::Mapped(paddr, level, self.prop());
```

I think this is more clearer and also more consistent with the `else` branch.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -166,141 +167,152 @@ pub(crate) fn current_page_table_paddr() -> Paddr {
...
+
+        let pa_mask = Self::pa_mask_at_level(level);
+
+        assert_eq!(
+            paddr & !pa_mask,
+            0,
+            "physical address contains invalid bits"
+        );
+
+        Self(paddr & pa_mask | flags)
```
lrh2000: I'm not sure why you need to use `& pa_mask` after an assertion. Would you like this version?
```patch
diff --git a/ostd/src/arch/x86/mm/mod.rs b/ostd/src/arch/x86/mm/mod.rs
index dbb751082..a7ce77a3f 100644
--- a/ostd/src/arch/x86/mm/mod.rs
+++ b/ostd/src/arch/x86/mm/mod.rs
@@ -260,15 +260,12 @@ impl PageTableEntry {
 
         flags |= cache_policy_to_flags(prop.cache).bits();
 
-        let pa_mask = Self::pa_mask_at_level(level);
-
         assert_eq!(
-            paddr & !pa_mask,
+            paddr & !Self::pa_mask_at_level(level),
             0,
             "physical address contains invalid bits"
         );
-
-        Self(paddr & pa_mask | flags)
+        Self(paddr | flags)
     }
 
     fn new_pt(paddr: Paddr, flags: PageTableFlags) -> Self {
@@ -281,7 +278,13 @@ impl PageTableEntry {
             | PteFlags::USER.bits()
             | parse_flags!(flags.bits(), PageTableFlags::AVAIL1, PteFlags::HIGH_IGN1)
             | parse_flags!(flags.bits(), PageTableFlags::AVAIL2, PteFlags::HIGH_IGN2);
-        Self(paddr & Self::CHILD_PT_ADDR_MASK | flags)
+
+        assert_eq!(
+            paddr & !Self::CHILD_PT_ADDR_MASK,
+            0,
+            "physical address contains invalid bits"
+        );
+        Self(paddr | flags)
     }
 }
 
```

ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
```diff
@@ -82,83 +81,104 @@ bitflags::bitflags! {
...
+
+    fn new_pt(paddr: Paddr, flags: PageTableFlags) -> Self {
+        let flags = PteFlags::READABLE.bits()
+            | PteFlags::WRITABLE.bits()
+            | parse_flags!(flags.bits(), PageTableFlags::AVAIL1, PteFlags::IGN1)
+            | parse_flags!(flags.bits(), PageTableFlags::AVAIL2, PteFlags::IGN2);
+        Self(paddr & Self::PHYS_MASK | flags)
     }
 
     fn is_last(&self, level: PagingLevel) -> bool {
```
lrh2000: This method should live near `is_present()`?

ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
```diff
@@ -82,83 +81,104 @@ bitflags::bitflags! {
...
+        }
+    }
+
+    fn to_repr(&self, level: PagingLevel) -> PteScalar {
+        if !self.is_present() {
+            return PteScalar::Absent;
+        }
+
+        let paddr = self.0 & Self::PHYS_MASK;
+
```
lrh2000: Remove this blank line to be consistent to other page table code.

ostd/src/mm/page_table/boot_pt.rs
```diff
@@ -212,45 +224,61 @@ impl<E: PageTableEntryTrait, C: PagingConstsTrait> BootPageTable<E, C> {
...
-        }
-        let mut prop = pte.prop();
-        op(&mut prop);
-        unsafe { pte_ptr.write(E::new_page(pte.paddr(), 1, prop)) };
+        match pte.to_repr(1) {
+            PteScalar::Absent => {
+                panic!("protecting an unmapped page in the boot page table");
+            }
+            PteScalar::PageTable(_, _, _) => {
+                unreachable!();
```
lrh2000: This seems simpler:
```rust
let PteScalar::Mapped(pa, _, mut prop) = pte.to_repr(1) else {
        panic!("protecting an unmapped page in the boot page table");
};
```

ostd/src/mm/page_table/boot_pt.rs
```diff
@@ -312,6 +327,8 @@ fn dfs_walk_on_leave<E: PteTrait, C: PagingConstsTrait>(
 ) {
     if level >= 2 {
         let pt_vaddr = paddr_to_vaddr(pt) as *mut E;
+        // SAFETY: The pointer points to a valid page table frame with
+        // `nr_subpage_per_huge` entries.
```
lrh2000: Better to avoid Rust slices here?

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -166,141 +167,152 @@ pub(crate) fn current_page_table_paddr() -> Paddr {
...
+            PteScalar::Mapped(paddr, level, prop) => Self::new_page(*paddr, *level, *prop),
+        }
+    }
+
+    fn to_repr(&self, level: PagingLevel) -> PteScalar {
+        if !self.is_present() {
+            return PteScalar::Absent;
+        }
+
+        if level == 1 || (self.0 & PteFlags::HUGE.bits() != 0) {
```
lrh2000: It seems that this PR does not address problems related to HUGE bit usage. Then, keep the original `is_last` implementation that checks the HUGE bit?
```c
        if self.0 & PteFlags::HUGE.bits() != 0 {
```

ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
```diff
@@ -82,83 +81,104 @@ bitflags::bitflags! {
     }
 }
 
+/// Parse a bit-flag bits `val` in the representation of `from` to `to` in bits.
+macro_rules! parse_flags {
```
lrh2000, line 89: This should live before `impl PageTableEntry`?

lrh2000 (commented):
The previous trait had the advantage of being able to
 - query `is_present()` without querying `paddr()` or `property()`, or
 - query `paddr()` without also querying `property()`. 

But the new trait will force the caller to query all of them.

Additionally, if we only want to determine whether a page is readable, perhaps we don't need to parse other unrelated flags. I was once thought how to properly implement that, but I can't come up with any ideas. I'm not sure if anyone else has any brilliant ideas...

---

However, I agree that this PR does make the code clearer in several ways. Below are some consistency issues that I noticed.

junyang-zh:
> The previous trait had the advantage of being able to
>     query is_present() without querying paddr() or property(), or
>     query paddr() without also querying property().


Good point. Maybe we can count on the compiler optimizing them away? Although I am not sure if it will always work. I may do some experiments later.

ostd/src/arch/loongarch/mm/mod.rs
```diff
@@ -205,24 +185,24 @@ impl PageTableEntryTrait for PageTableEntry {
...
+            | parse_flags!(!(self.0), PteFlags::NOT_EXECUTABLE, PageFlags::X)
             // TODO: How to get the accessed bit in loongarch?
-            | (parse_flags!(self.0, PageTableFlags::PRESENT, PageFlags::ACCESSED))
-            | (parse_flags!(self.0, PageTableFlags::DIRTY, PageFlags::DIRTY))
-            | (parse_flags!(self.0, PageTableFlags::RSV2, PageFlags::AVAIL2));
-        let mut priv_flags = parse_flags!(self.0, PageTableFlags::RSV1, PrivFlags::AVAIL1);
+            | parse_flags!(self.0, PteFlags::PRESENT, PageFlags::ACCESSED)
+            | parse_flags!(self.0, PteFlags::DIRTY, PageFlags::DIRTY)
+            | parse_flags!(self.0, PteFlags::RSV2, PageFlags::AVAIL2);
+        let mut priv_flags = parse_flags!(self.0, PteFlags::RSV1, PrivFlags::AVAIL1);
```
lrh2000, line 205: Insert a blank line before `let priv_flags` since we should now separate `flags`, `priv_flags`, and `cache` into different code blocks (as done in other architectures), and `cache` is already in a separate code block below.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -166,141 +163,158 @@ pub(crate) fn current_page_table_paddr() -> Paddr {
...
+
         flags |= cache_policy_to_flags(prop.cache).bits();
-        self.0 = self.0 & !Self::PROP_MASK | flags;
+
+        assert_eq!(
+            paddr & !Self::pa_mask_at_level(level),
+            0,
+            "page physical address contains invalid bits"
+        );
+
```
lrh2000: Remove this blank line, as `new_pt` does now.

Also, line 288:
```patch
-        Self(paddr & Self::CHILD_PT_ADDR_MASK | flags)
+        Self(paddr | flags)
```

ostd/src/mm/page_table/mod.rs
```diff
@@ -491,71 +495,70 @@ pub(super) unsafe fn page_walk<C: PageTableConfig>(
...
+    /// A PTE that establishes the mapping to a physical frame.
     ///
-    /// The physical address recorded in the PTE is either:
-    /// - the physical address of the next-level page table, or
-    /// - the physical address of the page that the PTE maps to.
-    fn paddr(&self) -> Paddr;
-
-    /// Returns the page property of the PTE.
-    fn prop(&self) -> PageProperty;
+    /// The level is the level of the current page table entry.
```
lrh2000: Outdated comments in line 533 and line 529.

ostd/src/mm/page_table/boot_pt.rs
```diff
@@ -160,33 +162,37 @@ impl<E: PageTableEntryTrait, C: PagingConstsTrait> BootPageTable<E, C> {
...
+                PteScalar::Absent => {
+                    let (pte, ch_pt_pa) = self.alloc_child(level);
+                    unsafe { pte_ptr.write(pte) };
+                    pt = ch_pt_pa;
+                }
+                PteScalar::Mapped(_, _) => {
+                    panic!("mapping an already mapped huge page in the boot page table");
+                }
+                PteScalar::PageTable(ch_pt_pa, _) => {
+                    pt = ch_pt_pa;
```
lrh2000: `ch_pt_pa` -> `child_pt`?

It seems that only this place uses the prefix `ch_` (which is a bit odd) and that other places tend to use something like `child_pt` or `child_pt_addr`.

ostd/src/mm/mod.rs
```diff
@@ -39,7 +39,7 @@ pub use self::{
     },
     kspace::{KERNEL_VADDR_RANGE, MAX_USERSPACE_VADDR},
     mem_obj::{HasDaddr, HasPaddr, HasPaddrRange, HasSize},
-    page_prop::{CachePolicy, PageFlags, PageProperty},
+    page_prop::{CachePolicy, PageFlags, PageProperty, PageTableFlags},
```
lrh2000: This does not seem useful for OSTD users (at least now?). So move it to `pub(crate)` below?

Also, maybe fix the visibility mark for `PageTableFlags`.

ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
```diff
@@ -84,81 +83,101 @@ bitflags::bitflags! {
...
+
         // TODO: The determination cache policy is not rigorous. We should revise it.
-        let cache = if self.0 & PageTableFlags::SNOOP.bits() != 0 {
+        let cache = if self.0 & PteFlags::SNOOP.bits() != 0 {
             CachePolicy::Writeback
         } else {
             CachePolicy::Uncacheable
         };
 
+        let priv_flags = parse_flags!(self.0, PteFlags::IGN2, PageFlags::AVAIL2);
```
lrh2000: I think you mean `IGN1` and `AVAIL1` here. Besides, please move this statement before `let cache`?

ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
```diff
@@ -84,81 +83,101 @@ bitflags::bitflags! {
...
-                | PageTableFlags::WRITABLE.bits(),
-        )
-    }
+impl PageTableEntry {
+    const PHYS_MASK: usize = 0xffff_ffff_f000;
 
-    fn paddr(&self) -> Paddr {
-        (self.0 & Self::PHYS_MASK) as usize
+    fn is_present(&self) -> bool {
+        self.0 & (PteFlags::READABLE | PteFlags::WRITABLE).bits() != 0
```
lrh2000: `PteFlags::LAST_PAGE | PteFlags::READABLE` because `new_page` will set `LAST_PAGE` and `new_pt` will set `READABLE`?

Otherwise, `is_present` will return false if a page is protected as non-readable and non-writable.

lrh2000 (commented):
Additionally, please test the LoongArch64 code? I guess #2825 will enable this (please read its PR description), however, the initramfs is still missing.

To test the userspace, we need initramfs from [here](https://github.com/asterinas/asterinas/issues/2224#issuecomment-3163676007) and patch the FPU:
```patch
diff --git a/ostd/src/arch/loongarch/cpu/context.rs b/ostd/src/arch/loongarch/cpu/context.rs
index fb8b0c593..17eaefa03 100644
--- a/ostd/src/arch/loongarch/cpu/context.rs
+++ b/ostd/src/arch/loongarch/cpu/context.rs
@@ -204,7 +204,7 @@ impl UserContextApiInternal for UserContext {
                             "Floating point unit is not available, badv: {badv:#x?}, badi: {badi:#x?}, era: {era:#x?}"
                         );
                         // TODO: Add FPU support and enable it when this exception occurs.
-                        break ReturnReason::UserException;
+                        self.enable_fpu();
                     }
                     Exception::TLBRFill => unreachable!(),
                 },
```

lrh2000:
> Maybe we can count on the compiler optimizing them away?

I think `prop()` is too long to be inlined, so it's unlikely to be optimized away. Besides, even if it is inlined, the compiler must be smart enough to eliminate all side effects, such as panics due to `PageFlags::from().unwrap()`. I doubt that's the case, though I'm not quite sure.

ostd/src/mm/page_table/boot_pt.rs
```diff
@@ -304,11 +319,18 @@ fn dfs_walk_on_leave<E: PteTrait, C: PagingConstsTrait>(
 ) {
     if level >= 2 {
         let pt_vaddr = paddr_to_vaddr(pt) as *mut E;
-        let pt = unsafe { core::slice::from_raw_parts_mut(pt_vaddr, nr_subpage_per_huge::<C>()) };
-        for pte in pt {
+        for offset in 0..nr_subpage_per_huge::<C>() {
+            // SAFETY: The pointer points to a valid page table frame with
+            // `nr_subpage_per_huge` entries.
+            let pte_ptr = unsafe { pt_vaddr.add(offset) };
```
lrh2000, line 324: This is same to:
https://github.com/asterinas/asterinas/blob/a95c3c04d476478291fedf4df0184428cddb658f/ostd/src/mm/page_table/boot_pt.rs#L240-L244

Use the same safety comment?

lrh2000:
Also I think next time you can split the rename of `PageTableFlags` to `PteFlags` into a separate commit. The current diff is really difficult to review... I'm not sure if you tried reading the diff yourself.

Perhaps there's no need to do that this time, since this PR is already being reviewed.

ostd/src/arch/loongarch/mm/mod.rs
```diff
@@ -244,73 +225,95 @@ impl PageTableEntryTrait for PageTableEntry {
...
-            .field("present", &self.is_present())
-            .field(
-                "flags",
-                &PageTableFlags::from_bits_truncate(self.0 & !Self::PHYS_ADDR_MASK),
-            )
-            .field("prop", &self.prop())
-            .finish()
+impl PodOnce for PageTableEntry {}
+
+/// SAFETY: The implementation is safe because:
```
lrh2000, line 295: The second point of the trait's safety requirement seems missing:
> a zeroed PTE represents an absent entry;

Same for all other `unsafe impl PteTrait for`.

ostd/src/arch/loongarch/mm/mod.rs
```diff
@@ -148,64 +145,47 @@ pub(crate) fn current_page_table_paddr() -> Paddr {
     pgdl
 }
 
+#[derive(Debug, Clone, Copy, Pod, Default)]
+#[repr(C)]
+pub(crate) struct PageTableEntry(usize);
+
+/// Parse a bit-flag bits `val` in the representation of `from` to `to` in bits.
```
lrh2000: Since this PR affects all files that define a similar macro, would you mind fixing it:
```suggestion
/// Parses a bit-flag bits `val` in the representation of `from` to `to` in bits.
```

Additionally, I'm unsure whether we should move it to the `utils` module as a `pub(crate)` macro. This would allow us to avoid duplicating the code.

junyang-zh:
> So let me ask again, did you test the LoongArch64 code?

I ran with this:

```
make clean
mkdir test/initramfs/build
curl -L "https://github.com/wyt8/BITOS/raw/eec47ca984d556275612152238e10d85d657cae3/ostd/src/arch/loongarch/boot/initramfs.cpio" -o test/initramfs/build/initramfs.cpio
gzip test/initramfs/build/initramfs.cpio
make run_kernel OSDK_TARGET_ARCH=loongarch64
```

And I got:

```
qemu-system-loongarch64: Some ROM regions are overlapping
These ROM regions might have been loaded by direct user request or by default.
They could be BIOS/firmware images, a guest kernel, initrd or some other file loaded into guest memory.
Check whether you intended to load all this guest code, and whether it has been built to load to the correct addresses.

The following two regions overlap (in the memory address space):
  /root/asterinas/target/osdk/aster-kernel/aster-kernel-osdk-bin.qemu_elf ELF program header segment 0 (addresses 0x0000000080000000 - 0x0000000080042000)
  /root/asterinas/test/initramfs/build/initramfs.cpio.gz (addresses 0x0000000080000000 - 0x000000008001e070)
```

Something may have changed.

If do without initramfs:

```
[kernel] Spawn the first kernel thread
[     0.000] ERROR: Uncaught panic:
        No initramfs found!
        at /root/asterinas/kernel/src/fs/rootfs.rs:31
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(0), real_time: RealTimeAttr { prio: 99, time_slice: 1500000 }, fair: FairAttr { weight: 1024, pending_weight: 0, vruntime: 600000 } } })
```

By this far we already get to the page table. So I think it works?

ostd/src/mm/page_table/mod.rs
```diff
@@ -467,71 +471,66 @@ pub(super) unsafe fn page_walk<C: PageTableConfig>(
...
-    Clone + Copy + Debug + Default + Pod + PodOnce + Sized + Send + Sync + 'static
+/// A PTE refers to an entry in any level of a page table. This trait requires
+/// that any architecture-specific PTEs are scalar types that essentially
+/// encodes [`PteScalar`].
+///
+/// # Safety
+///
+/// An implementor must ensure that:
+///  - the methods `as_usize` and `from_usize` are not overridden;
+///  - a zeroed PTE represents an absent entry;
```
lrh2000: So.. Move this to the last safety requirement? So it's consistent that we list this as the third safety requirement:
https://github.com/asterinas/asterinas/blob/affcb5ccff00ecfc6495c68325f389d25f6a46f3/ostd/src/arch/loongarch/mm/mod.rs#L295-L298

lrh2000:
> ```
> qemu-system-loongarch64: Some ROM regions are overlapping
> These ROM regions might have been loaded by direct user request or by default.
> They could be BIOS/firmware images, a guest kernel, initrd or some other file loaded into guest memory.
> Check whether you intended to load all this guest code, and whether it has been built to load to the correct addresses.
> 
> The following two regions overlap (in the memory address space):
>   /root/asterinas/target/osdk/aster-kernel/aster-kernel-osdk-bin.qemu_elf ELF program header segment 0 (addresses 0x0000000080000000 - 0x0000000080042000)
>   /root/asterinas/test/initramfs/build/initramfs.cpio.gz (addresses 0x0000000080000000 - 0x000000008001e070)
> ```

It sounds like a QEMU bug:
https://github.com/qemu/qemu/blob/38879a667fbb4ef54c70de71494882615f600a64/hw/loongarch/boot.c#L295
https://github.com/qemu/qemu/blob/38879a667fbb4ef54c70de71494882615f600a64/hw/loongarch/boot.c#L333-L335

QEMU does not respect the suggested address `ROUND_UP(kernel_high + 4 * kernel_size, 64 * KiB)` if we try to load the kernel and initramfs in the highmem (which [starts at 0x80000000](https://github.com/qemu/qemu/blob/38879a667fbb4ef54c70de71494882615f600a64/include/hw/loongarch/virt.h#L72)).

Something like the following patch might work around this problem. However, I just realize that https://github.com/wyt8/BITOS/raw/eec47ca984d556275612152238e10d85d657cae3/ostd/src/arch/loongarch/boot/initramfs.cpio has 404 now, so testing the kernel page table in this PR alone may be sufficient.
```patch
diff --git a/osdk/src/base_crate/loongarch64.ld.template b/osdk/src/base_crate/loongarch64.ld.template
index 40796e9d2..5826bc784 100644
--- a/osdk/src/base_crate/loongarch64.ld.template
+++ b/osdk/src/base_crate/loongarch64.ld.template
@@ -2,10 +2,10 @@ ENTRY(_start)
 OUTPUT_ARCH(loongarch)
 
 # The physical address where the kernel will start to be loaded.
-KERNEL_LMA = 0x80000000;
+KERNEL_LMA = 0x800000;
 
 # The virtual memory of the kernel mapping.
-KERNEL_VMA = 0x9000000080000000;
+KERNEL_VMA = 0x9000000000800000;
 
 KERNEL_VMA_OFFSET = KERNEL_VMA - KERNEL_LMA;
 
```

lrh2000:
I tested the code with LoongArch64 userspace as well (see the instructions at https://github.com/asterinas/asterinas/issues/2224#issuecomment-3767169878). No problems were found.

PR #2597, opened 2025-11-18
Lint all features

junyang-zh:
This PR lets `make check` check with all features enabled. Also it fixes some problems previously missed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -368,14 +368,14 @@ check: initramfs $(CARGO_OSDK)
 	@# Check compilation of the Rust code
 	@for dir in $(NON_OSDK_CRATES); do \
 		echo "Checking $$dir"; \
-		(cd $$dir && cargo clippy -- -D warnings) || exit 1; \
+		(cd $$dir && cargo clippy --all-features -- -D warnings) || exit 1; \
 	done
 	@for dir in $(OSDK_CRATES); do \
 		echo "Checking $$dir"; \
 		# Exclude linux-bzimage-setup since it only supports x86-64 currently and will panic \
 		# in other architectures. \
 		[ "$$dir" = "ostd/libs/linux-bzimage/setup" ] && [ "$(OSDK_TARGET_ARCH)" != "x86_64" ] && continue; \
-		(cd $$dir && cargo osdk clippy -- -- -D warnings) || exit 1; \
+		(cd $$dir && cargo osdk clippy -- --all-features -- -D warnings) || exit 1; \
 	done
 	@
 	@# Check formatting issues of the C code (regression tests)
--- a/kernel/src/device/tdxguest/mod.rs
+++ b/kernel/src/device/tdxguest/mod.rs
@@ -232,7 +232,7 @@ fn handle_get_report(arg: usize) -> Result<i32> {
     let user_space = CurrentUserSpace::new(current_task.as_thread_local().unwrap());
     let user_request: TdxReportRequest = user_space.read_val(arg)?;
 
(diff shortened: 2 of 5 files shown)


lrh2000:
Can't we use [`--all-features`](https://doc.rust-lang.org/cargo/reference/features.html#command-line-feature-options)?

junyang-zh:
> Can't we use [`--all-features`](https://doc.rust-lang.org/cargo/reference/features.html#command-line-feature-options)?

The `coverage` feature is broken on loongarch but OK for other archs. I think we should be smarter. Let me investigate how to deal with it.

lrh2000:
[Another problem](https://github.com/asterinas/asterinas/actions/runs/19488633643/job/55776254393?pr=2597) seems to be:
```
    Checking ostd v0.16.1 (/__w/asterinas/asterinas/ostd)
error: feature "riscv_sv39_mode" cannot be specified for architectures other than RISC-V
   --> ostd/src/lib.rs:160:5
    |
160 | /     compile_error!(
161 | |         "feature \"riscv_sv39_mode\" cannot be specified for architectures other than RISC-V"
162 | |     );
    | |_____^
```

I'm wandering if we can/should solve it by something like:
```patch
diff --git a/ostd/src/arch/riscv/mm/mod.rs b/ostd/src/arch/riscv/mm/mod.rs
index 8aabb3ec7..a796a05f3 100644
--- a/ostd/src/arch/riscv/mm/mod.rs
+++ b/ostd/src/arch/riscv/mm/mod.rs
@@ -26,7 +26,7 @@ mod util;
 #[derive(Clone, Debug, Default)]
 pub(crate) struct PagingConsts {}
 
-#[cfg(not(feature = "riscv_sv39_mode"))]
+#[cfg(target_pt_level = "4")]
 impl PagingConstsTrait for PagingConsts {
     const BASE_PAGE_SIZE: usize = 4096;
     const NR_LEVELS: PagingLevel = 4;
@@ -36,7 +36,7 @@ impl PagingConstsTrait for PagingConsts {
     const PTE_SIZE: usize = size_of::<PageTableEntry>();
 }
 
-#[cfg(feature = "riscv_sv39_mode")]
+#[cfg(target_pt_level = "3")]
 impl PagingConstsTrait for PagingConsts {
     const BASE_PAGE_SIZE: usize = 4096;
     const NR_LEVELS: PagingLevel = 3;
```


It should make it easier for us to support  [Intel's 5-level paging](https://en.wikipedia.org/wiki/Intel_5-level_paging).

But I'm not saying that `target_pt_level` is a good name. Maybe we can come up with a better one if this approach is on the right track.

junyang-zh:
> I'm wandering if we can/should solve it by something like:

My concern is that this solution introduces an [unexpected cfg](https://blog.rust-lang.org/2024/05/06/check-cfg/) that generates warnings. We need to bypass this check by passing `--check-cfg` arguments in OSDK, like what we do to `cfg(ktest)`, which is not easy to maintain compared to maintaining all the `all` features.

lrh2000:
> My concern is that this solution introduces an [unexpected cfg](https://blog.rust-lang.org/2024/05/06/check-cfg/) that generates warnings. We need to bypass this check by passing `--check-cfg` arguments in OSDK, like what we do to `cfg(ktest)`, which is not easy to maintain compared to maintaining all the `all` features.

Yes. I think it's acceptable because it's a configuration used in OSTD. However, what you said is also valid. I posted it as an alternative worth considering, but I don't have a preference either way anymore.

Issue #2562, opened 2025-11-07
Manage char devices separately


tatetian:
This issue proposes to manage char devices separately from block or other devices.

## Background

#2475 attempts to manage _all_ devices by 
1. Providing a `Device` trait that abstract any device, which has a device ID and a device type.
2. Providing a centralized place to register any devices, including both char and block devices.

This approach, unfortunately, does not make much sense. This is because---perhaps surprising to some common belief---**not all devices are block or char devices** and **not all devices have device IDs**. But the abstractions proposed by #2475 seems to be based on these misbelief.  

Any device appear under `/dev` is either a block or char device. But the problem is that not all devices have corresponding device files under `/dev`. A prominent example is network cards, which are manipulated through dedicated APIs such as netlink rather than device files under `/dev`. 

Similarly, any device appear under `/dev` has a device ID. In other words, all block and char devices have device IDs. This point is obvious from the organization of sysfs  (`/sys/dev/char/MAJOR:MINOR` and `/sys/dev/block/MAJOR:MINOR`). But not all devices have device IDs. If you really think about it, this is unsurprising: the kernel must provide a way for the userspace to identify a device and create its device file under `/dev` or other FSes via `mknod(path, S_IFCHR|mode, devid)` or `mknod(path, S_IFBLK|mode, devid)` . This is why block and char devices must have IDs. But if a device does not need to have a device file, then assigning it a device ID would be a waste of IDs.

For the reasons above, the device abstraction in #2475 is deemed too devtmpfs-centric and not general or useful enough to serve as a proper abstraction for all devices.

```rust
pub trait Device: Send + Sync + Debug {
    fn type_(&self) -> DeviceType;
    fn id(&self) -> Option<DeviceId>;
}

pub enum DeviceType {
    Block,
    Char,
    Other,
}

pub struct DeviceId {
    major: u32,
    minor: u32,
}
```

In addition, **the registration API is duplicated with the class-specific registration API**.

```rust
/// Adds a device to the system.
///
/// This function adds a device to the system's device management infrastructure.
/// It performs the following actions:
///
/// 1. If the device doesn't have a parent, it adds the device to the global devices node.
/// 2. It creates a symlink in the appropriate device type directory (block or char)
///    under `sys/dev` using the device's major and minor numbers.
pub fn add_device(device: Arc<dyn Device>) { ... }
```

For example, the block component already has a registration API,

```rust
pub fn register_device(name: String, device: Arc<dyn BlockDevice>) { ... }
[...]

PR #2598, opened 2025-11-19
Refactor mem/misc char devices

cqs21:
This PR is based on #2560 and refactors the `mem/misc` character devices according to the basic design in #2562. Some systree-related features have not been implemented yet.

### DIFF
--- a/kernel/src/device/disk.rs
+++ b/kernel/src/device/disk.rs
@@ -45,8 +45,11 @@ pub(super) fn init_in_first_process(fs_resolver: &FsResolver) -> Result<()> {
     Ok(())
 }
 
+/// Represents a block device inode in the filesystem.
+///
+/// Only implements the `Device` trait.
 #[derive(Debug)]
-struct BlockFile(Arc<dyn BlockDevice>);
+pub struct BlockFile(Arc<dyn BlockDevice>);
 
 impl BlockFile {
     fn new(device: Arc<dyn BlockDevice>) -> Self {
@@ -54,7 +57,27 @@ impl BlockFile {
     }
 }
 
-impl InodeIo for BlockFile {
+impl Device for BlockFile {
+    fn type_(&self) -> DeviceType {
+        DeviceType::Block
+    }
+
+    fn id(&self) -> DeviceId {
+        self.0.id()
+    }
+
+    fn open(&self) -> Result<Box<dyn FileIo>> {
+        Ok(Box::new(OpenBlockFile(self.0.clone())))
+    }
+}
+
+/// Represents an opened block device file ready for I/O operations.
+///
+/// Does not implement the `Device` trait but provides full implementations
+/// for I/O related traits.
+pub struct OpenBlockFile(Arc<dyn BlockDevice>);
+
+impl InodeIo for OpenBlockFile {
     fn read_at(
         &self,
         offset: usize,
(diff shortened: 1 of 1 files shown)


kernel/src/device/char.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// The bottom of the second segment of free char majors.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/linux/fs.h#L2843>.
+const CHAR_SECOND_DYNAMIC_MAJOR_START: u16 = 384;
+
+/// The top of the second segment of free char majors (included).
+///
+/// Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/linux/fs.h#L2842>.
+const CHAR_SECOND_DYNAMIC_MAJOR_END: u16 = 511;
```
tatetian: Define the two dynamic major ranges as two `Range<u16>`s rather than two integer values to improve readability:

```rust
/// TODO: Add Rust doc.
const DYNAMIC_MAJOR_ID_RANGES: [Range<u16>; 2] = todo!();
```

An addition benefit of using `Range` is that you don't need to clarify whether the bound is inclusive or exclusive.

kernel/src/device/char.rs
```diff
@@ -0,0 +1,212 @@
...
+        utils::{InodeIo, StatusFlags},
+    },
+    prelude::*,
+    process::signal::{PollHandle, Pollable},
+};
+
+/// A character device.
+pub trait CharDevice: Send + Sync + Debug {
+    /// Returns the name of the char device.
+    fn name(&self) -> &str;
```
tatetian: Return `DevtmpfsName` as suggested in #2562. 

[The Evdev PR](https://github.com/asterinas/asterinas/pull/2561) needs to add inodes whose paths are `/dev/input/eventX`. Simply returning a device name of `&str` does not provide the sufficient information for the full path like `/dev/input/eventX`.

kernel/src/device/char.rs
```diff
@@ -0,0 +1,212 @@
...
+const CHAR_SECOND_DYNAMIC_MAJOR_END: u16 = 511;
+
+static MAJORS: Mutex<BTreeSet<u16>> = Mutex::new(BTreeSet::new());
+
+/// Acquires a major ID.
+///
+/// The returned `MajorIdOwner` object represents the ownership to the major ID.
+/// Until the object is dropped, this major ID cannot be acquired via `acquire_major` or `allocate_major` again.
+pub fn acquire_major(major: MajorId) -> Result<MajorIdOwner> {
+    if major.get() == 0 || major.get() > MAX_MAJOR {
```
tatetian: Remove `major.get() == 0`

kernel/src/device/char.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn type_(&self) -> DeviceType {
+        DeviceType::Char
+    }
+
+    fn id(&self) -> DeviceId {
+        self.0.id()
+    }
+
+    fn open(&self) -> Result<Box<dyn FileIo>> {
+        Ok(Box::new(CharFile(self.0.clone())))
```
tatetian: Need to use two separate types to represent a char device inode and an _open_ char device file. Let's say one is `CharFile` and the other is `OpenCharFile`. `CharFile` need implement the `Device` trait, but only provide dummy or error-returning implementations for `InodeIo`, `Pollable`, and `FileIo`. On the other hand, `OpenCharFile` does not implement the `Device` trait, but implement the `InodeIo`, `Pollable`, and `FileIo` traits correctly.

The same `BlockFile` type merged in #2560 also suffers this problem. Fix it in a separate commit.

tatetian (commented):
Thanks for adding the initial support for char devices. This PR is in a good shape, although some minor issues need to be addressed.

tatetian (approved):

PR #2601, opened 2025-11-20
Fix some mount-related bugs

cchanging:
During the recursive subtree clone operation in `Mount`, if the child mount of the current mount is mounted on the root dentry of the current mount, it also needs to be cloned. In addition, the original input parameters for the `is_descendant_of` were incorrect, which would cause errors once multiple layers of mounts exist.

### DIFF
--- a/kernel/src/fs/path/dentry.rs
+++ b/kernel/src/fs/path/dentry.rs
@@ -101,16 +101,18 @@ impl Dentry {
         DentryFlags::from_bits(flags).unwrap()
     }
 
-    /// Checks if this dentry is a descendant (child, grandchild, or
-    /// great-grandchild, etc.) of another dentry.
-    pub(super) fn is_descendant_of(&self, ancestor: &Arc<Self>) -> bool {
-        let mut parent = self.parent();
-        while let Some(p) = parent {
-            if Arc::ptr_eq(&p, ancestor) {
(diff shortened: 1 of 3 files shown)


kernel/src/fs/path/dentry.rs
```diff
@@ -101,16 +101,20 @@ impl Dentry {
...
-    /// great-grandchild, etc.) of another dentry.
-    pub(super) fn is_descendant_of(&self, ancestor: &Arc<Self>) -> bool {
-        let mut parent = self.parent();
-        while let Some(p) = parent {
-            if Arc::ptr_eq(&p, ancestor) {
+    /// Checks whether the `Dentry` is in the subtree of the given `root` dentry.
+    ///
+    /// If the `Dentry` is the same as the `root` or is a descendant of the `root`,
+    /// returns true.
+    pub(super) fn is_in_subtree_of(&self, root: &Arc<Self>) -> bool {
```
tatetian: I still think the original interface is more clear. People may think that a `Dentry` called `root` is one that has no parent. In addition, throughout this module, no documentation or comments have called a `Dentry` _subtree_. Up til now, the mental model for a `Dentry` has been a node in a tree, not a subtree.

tatetian (commented):
Good job on fixing the bug!

tatetian (approved):

PR #2083, opened 2025-05-15
Add framework of fsnotify and syscall about inotify

Plucky923:

PR #2603, opened 2025-11-21
Refactor inode extensions and ensure that all inodes have extensions

### LINKED ISSUE #2579: `generate_tdx_quote` test case fails due to `flock` with `configfs` inode
### Describe the bug

The [TDX integration-test (general)](https://github.com/asterinas/asterinas/actions/runs/19313198664/job/55238451072#logs) pipeline is failing because the `generate_tdx_quote` test case fails when attempting to `flock` a `configfs` inode:
```c
    // Lock `inblob` to avoid other processes accessing it using libtdx_attest
    // Will unlock it via close()
    int fd_lock = open(inblob_path, O_WRONLY | O_CLOEXEC);
    if (-1 == fd_lock) {
        TDX_TRACE;
        syslog(LOG_ERR, "libtdx_attest: failed to open configFS inblob.");
        return TDX_ATTEST_ERROR_UNEXPECTED;
    }
    if (flock(fd_lock, LOCK_EX)) {
        TDX_TRACE;
        close(fd_lock);
        syslog(LOG_ERR, "libtdx_attest: failed to lock configFS inblob.");
        return TDX_ATTEST_ERROR_UNEXPECTED;
    }
```
However, we recently made it return error directly under `sysfs`.  The PR is #2563.
https://github.com/asterinas/asterinas/blob/bb48b3814f17f16633c9d9967fe7aa20022883f6/kernel/src/fs/inode_handle/mod.rs#L233-L237

### To Reproduce

This issue requires TDX-capable hardware to reproduce. I'm available for direct communication to coordinate debugging sessions or provide additional environment details.

lrh2000:
This fixes #2579. See https://github.com/asterinas/asterinas/issues/2579#issuecomment-3532507461.

As I said in https://github.com/asterinas/asterinas/issues/2579#issuecomment-3532507461, maybe I should add inode extensions for other sysfs/procfs inodes as well, but I'm not sure if there is a use case?

### DIFF
--- a/kernel/src/fs/notify/mod.rs
+++ b/kernel/src/fs/notify/mod.rs
@@ -59,37 +59,46 @@ impl FsEventPublisher {
     /// Adds a subscriber to this publisher.
     pub fn add_subscriber(&self, subscriber: Arc<dyn FsEventSubscriber>) -> bool {
         let mut subscribers = self.subscribers.write();
+
         // This check must be done after locking `self.subscribers.write()` to avoid race
         // conditions.
         if !self.accepts_new_subscribers.load(Ordering::Relaxed) {
             return false;
         }
+
         let current = self.all_interesting_events.load(Ordering::Relaxed);
         self.all_interesting_events
             .store(current | subscriber.interesting_events(), Ordering::Relaxed);
+
         subscribers.push(subscriber);
+
         true
     }
 
     /// Removes a subscriber from this publisher.
     pub fn remove_subscriber(&self, subscriber: &Arc<dyn FsEventSubscriber>) -> bool {
         let mut subscribers = self.subscribers.write();
+
         let orig_len = subscribers.len();
         subscribers.retain(|m| !Arc::ptr_eq(m, subscriber));
+
         let removed = subscribers.len() != orig_len;
         if removed {
             subscriber.deliver_event(FsEvents::IN_IGNORED, None);
             self.recalc_interesting_events(&subscribers);
         }
+
         removed
     }
 
     /// Removes all subscribers from this publisher.
     pub fn remove_all_subscribers(&self) -> usize {
         let mut subscribers = self.subscribers.write();
+
         for subscriber in subscribers.iter() {
             subscriber.deliver_event(FsEvents::IN_IGNORED, None);
         }
+
         let num_subscribers = subscribers.len();
         subscribers.clear();
 
@@ -122,19 +131,14 @@ impl FsEventPublisher {
         }
     }
 
-    /// Recalculates the aggregated interesting events from all subscribers.
-    fn recalc_interesting_events(&self, subscribers: &[Arc<dyn FsEventSubscriber>]) {
-        let mut new_events = FsEvents::empty();
-        for subscriber in subscribers.iter() {
-            new_events |= subscriber.interesting_events();
-        }
-        self.all_interesting_events
-            .store(new_events, Ordering::Relaxed);
-    }
-
     /// Updates the aggregated events when a subscriber's interesting events change.
     pub fn update_subscriber_events(&self) {
         let subscribers = self.subscribers.read();
+        self.recalc_interesting_events(&subscribers);
+    }
+
+    /// Recalculates the aggregated interesting events from all subscribers.
+    fn recalc_interesting_events(&self, subscribers: &[Arc<dyn FsEventSubscriber>]) {
         let mut new_events = FsEvents::empty();
         for subscriber in subscribers.iter() {
             new_events |= subscriber.interesting_events();


StevenJiang1110:
> As I said in https://github.com/asterinas/asterinas/issues/2579#issuecomment-3532507461, maybe I should add inode extensions for other sysfs/procfs inodes as well, but I'm not sure if there is a use case?

Theoretically, sysfs and procfs should support flock, although that functionality may not be used at the moment. We may run into related issues eventually, so it would be reasonable to fix flock support for sysfs/procfs. Would implementing that extension require significant effort? If not, we could include it in this PR; otherwise, addressing only the configfs problem is also acceptable.

tatetian:
We do not want a rarely-used feature to put a toll on every inode by default. I propose to make `Extension` (almost) zero cost by wrapping its fields into a `Once<Box<_>>`.

```rust
pub struct Extension {
    data: Once<Box<RwLock<BTreeMap<TypeId, Arc<dyn Any + Send + Sync>>>>>,
}
```

With this change, I would be much more comfortable with adding `Extension` to all inodes. In fact, I think making this object a mandatory property of an inode is acceptable.

lrh2000:
> We do not want a rarely-used feature to put a toll on every inode by default. I propose to make `Extension` (almost) zero cost by wrapping its fields into a `Once<Box<_>>`.

I don't think `RwLock<BTreeMap<>>` is heavy at all. Its size is constant and reasonably small (regardless of how many extensions we have in the future). Wrapping it in an `Once` doesn't add much value.

Maybe you want to see:
```rust
struct Extension {
    inner: Once<Box<Inner>>,
}

struct Inner {
    flock_list: FlockList,
    range_lock_list: RangeLockList,
    // Other extensions in the future, if any.
}
```

Also I don't think it is reasonable to use `RwLock`, `BTreeMap<TypeId, ...>`, and `Arc<dyn Any>` here. The new design should eliminate all of them and I don't see any problems.

lrh2000:
> Would implementing that extension require significant effort? If not, we could include it in this PR

If so, we may want to just make the extension a mandatory property of an inode because inodes of all FSes will have its `Extension` type. (As @tatetian said above.)

tatetian:
> I don't think `RwLock<BTreeMap<>>` is heavy at all. Its size is constant and reasonably small (regardless of how many extensions we have in the future).

The struct size is 32 bytes. This is not very heavy-weight, but it is far from zero cost. Remember this is an overhead that adds to _every_ inode struct. There could be many of them and that most inodes will never use the file lock-related features. On the other hand, wrapping it inside a `Once<Box<_>>` (or something similar) should be able to reduce the memory footprint to 8 bytes, the size of a pointer--in theory. Although in practice, the `Once` struct attaches the pointer with an atomic flag. So using `Once<Box<_>>` directly would be 16 bytes, reducing the memory footprint by 50% . But obviously, we could invent an `OnceArc` type in OSTD to achieve the minimal memory footprint.

> ```rust
> struct Extension {
>     inner: Once<Box<Inner>>,
> }
> 
> struct Inner {
>     flock_list: FlockList,
>     range_lock_list: RangeLockList,
>     // Other extensions in the future, if any.
> }
> ```

No, I am not enthusiastic about this proposal. Why do we want `Extension` in the first place? Because in my opinion, FS implementations should not be coupled with kernel structs such as file locks. If our FS interface include kernel-only concepts, we will never be able to extract FS implementations out of the kernel crate as OS components.

In addition, each extension should be initialized _on demand_. The new proposal will initialize all or nothing, which is acceptable although not ideal. This is a minor point.

lrh2000:
> Because in my opinion, FS implementations should not be coupled with kernel structs such as file locks. If our FS interface include kernel-only concepts, we will never be able to extract FS implementations out of the kernel crate as OS components.

There are at least three ways to address this issue.
 1. Use generic parameters.
```rust
// In FS crates:
struct FsInode<E> {
    ext: E,
}

impl Inode<E> for FsInode<E> {
    fn extension(&self) -> &E {
        &self.ext
    }
}

// In `aster-nix`:
struct Extension { /* ... */ }

type Inode = fs::Inode<Extension>;
```
   - This is currently done in `aster-bigtcp`. The downside is that adding generic parameters to every inode type/trait is very verbose.
https://github.com/asterinas/asterinas/blob/a616682fe3974e0b1ca04a8e009e31570593e1eb/kernel/src/net/iface/mod.rs#L11
 2. Use existential types: https://github.com/rust-lang/rfcs/pull/2492. Of course, the RFC has been postponed so we have no way to use it in a near feature.
 3. Use `Box<dyn Any>`:
```rust
// In FS crates:
struct Extension {
    inner: Once<Box<dyn Any>>,
}

// In `aster-nix`:
struct Inner {
    flock_list: FlockList,
    range_lock_list: RangeLockList,
}
```

If we have `Once<Box<>>`, `BTreeMap<TypeId, ...>`, and `Arc<dyn Any>`, then there are three levels plus one map and one lock to reach a concrete extension. While the efficiency may not matter, the design is still quite complex and not very elegant. I think there are quite some reasonable ways to eliminate one or two levels of indirection without harming anything. For example, by having extension groups:
```rust
struct Extension {
     // Use `dyn Any` or `ExtensionGroup` depending on whether the type can be named.
     group1: OnceBox<dyn Any>,
     group2: OnceBox<dyn Any>,
}

struct ExtensionGroup1 {
    flock_list: FlockList,
    range_lock_list: RangeLockList,
}
```
Or:
```rust
struct Extension {
    inner: OnceBox<dyn Any>,
}

struct Inner {
    flock_list: OnceBox<FlockList>,
    range_lock_list: OnceBox<RangeLockList>,
}
```

Neither of them has maps or locks. The first has one level of indirection, while the second has two, which are better than the original design's three levels of indirection: `Once<Box<>>`/`HashMap<TypeId,>`/`Arc<dyn Any>`. In addition, it avoids `Arc`'s reference counting.

tatetian:
I am worried that the first approach would introduce some ergonomic issues. Too many locations may need to add generic parameters. The second approach is infeasible in the short term.

The third approach sounds good to me, especially option 1.

I also want to mention that there is one emerging user of the new `Extension`, which is fsnotify/inotify. The [pending inotify PR](https://github.com/asterinas/asterinas/pull/2083) (to be merged soon) will add a new per-inode property, `FsEventPublisher`.

```rust
    /// Returns a reference to the `FsEventPublisher` structure associated with this inode.
    ///
    /// This method provides access to the notification framework for this inode,
    /// allowing the addition and removal of notification subscribers and the delivery
    /// of filesystem events to interested parties.
    fn fs_event_publisher(&self) -> &FsEventPublisher;
```

This property will be accessed on every file or FS operation (if the FS currently has at one inode being monitored by inotify files). So getting the object efficiently is a key requirement. This is why the property is not added to `Extension`. (Another reason is that `Extension` is optional, but the support for fsnotify/inotify is not). If `Extension` can be made zero cost, then we can migrate `FsEventPublisher` to `Extension`. This will free FS implementations from worrying about VFS concepts such as `FsEventPublisher`.

taosue:
Hi @lrh2000, could you prioritize this PR? TDX CI has been failing for a long time...

lrh2000:
> Hi @lrh2000, could you prioritize this PR? TDX CI has been failing for a long time...

Sorry for the delay and thanks for the reminder. I'll address the PR within this week.

lrh2000:
Now I rewrite the inode extension to something like:
```rust
/// An extension is a set of object groups that is attached to an inode.
///
/// In this structure, we do not specify the exact type, but instead use [`Any`], which makes the
/// FS types (e.g., [`Inode`]) independent of the kernel types. This allows the file system
/// implementation to exist outside the kernel.
#[derive(Debug)]
pub struct Extension  { /* private */ }

impl Extension {
    /// Gets the first extension group.
    pub fn group1(&self) -> &Once<ThinBox<dyn Any + Send + Sync>>;
    /// Gets the second extension group.
    pub fn group2(&self) -> &Once<ThinBox<dyn Any + Send + Sync>>;
}
```

Then I provide methods via `InodeExt`:
```rust
/// A trait that instantiates kernel types for the inode [`Extension`].
///
/// [`Extension`]: super::Extension
pub trait InodeExt {
    /// Gets or initializes the FS event publisher.
    ///
    /// If the publisher does not exist for this inode, it will be created.
    fn fs_event_publisher_or_init(&self) -> &FsEventPublisher;

    /// Returns a reference to the FS event publisher.
    ///
    /// If the publisher does not exist for this inode, a [`None`] will be returned.
    fn fs_event_publisher(&self) -> Option<&FsEventPublisher>;

    /// Gets or initializes the FS lock context.
    ///
    /// If the context does not exist for this inode, it will be created.
    fn fs_lock_context_or_init(&self) -> &FsLockContext;

    /// Returns a reference to the FS lock context.
    ///
    /// If the context does not exist for this inode, a [`None`] will be returned.
    fn fs_lock_context(&self) -> Option<&FsLockContext>;

}

impl InodeExt for dyn Inode { /* impls */ }
```

---

Additionally, this PR modifies some inotify code. I didn't examine all of the code carefully, but I noticed and fixed some problems as I went along.

#### Remove `VecDeque::try_with_capacity`

For large buffers, they need to be (1) allocated dynamically and (2) accounted to some limitation. Unfortunately, we currently don't have a strategy to account large kernel buffers to the user process (Linux's `__GFP_ACCOUNT` flag, please read [this documentations](docs.kernel.org/core-api/memory-allocation.html#:~:text=Untrusted,accounted) and see [the implementations](https://elixir.bootlin.com/linux/v6.18.2/source/fs/notify/fanotify/fanotify.c#L753)), but using `try_with_capacity` for this purpose does not make any sense.

#### Fix `accepts_new_subscribers` race

This is obviously racy. Also, the `Acquire`/`Release` pair does not make sense; synchronization is done via the `self.subscribes.write()` lock.

kernel/src/fs/notify/inotify.rs
```diff
@@ -98,16 +98,12 @@ impl InotifyFile {
     /// Watch Description starts from 1.
     /// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/notify/inotify/inotify_user.c#L402>
     pub fn new(is_nonblocking: bool) -> Result<Arc<Self>> {
-        let event_queue = VecDeque::try_with_capacity(DEFAULT_MAX_QUEUED_EVENTS).map_err(|_| {
```
tatetian, line 97: Actually, the motivation of using `VecDeque::try_with_capacity` is to handle potential kernel out-of-memory (OOM) due to large buffer allocation. Of course. There are too many locations that may cause OOM; so fixing one place is futile. But it does removes one such location. So it should be kept. In the run long, we will need to replace almost `Box::new`, `Arc::new`, `Vec::with_capacity`, and more with their `try_` version.

tatetian (commented):
The code looks perfect. Just one point for discussion before the PR gets merged.

kernel/src/fs/notify/inotify.rs
lrh2000: Yes, we need to handle OOM, but this is really too far from being the correct way, as I said in https://github.com/asterinas/asterinas/pull/2603#issuecomment-3692043830:
> For large buffers, they need to be (1) allocated dynamically and (2) accounted to some limitation. Unfortunately, we currently don't have a strategy to account large kernel buffers to the user process (Linux's `__GFP_ACCOUNT` flag, please read [this documentations](docs.kernel.org/core-api/memory-allocation.html#:~:text=Untrusted,accounted) and see [the implementations](https://elixir.bootlin.com/linux/v6.18.2/source/fs/notify/fanotify/fanotify.c#L753)), but using `try_with_capacity` for this purpose does not make any sense.

There are quite some points here:
 - `DEFAULT_MAX_QUEUED_EVENTS` is a large value representing the upper limit. Assuming the user space functions properly, this limit won't be reached. Therefore, allocating such a large buffer in advance would waste too much memory.
 - To properly handle OOMs, it makes no sense to only handle them for large allocations. A program can use large allocations to _almost_ exhaust kernel memory; then, small kernel allocations can also fail. Therefore, we need to handle failures for all allocations.
 - For simple collections (perhaps only `Vec` at the time of writing), the standard library provides APIs (like [`Vec::push_within_capacity`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push_within_capacity)) to avoid unexpected allocations. However, complex collections like `VecDeque` provide no similar functions. For `VecDeque`, there are at least methods like [`VecDeque::try_with_capacity`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.try_with_capacity). But those methods do not exist for more complex ones like [`BTreeMap`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html).
 - This is not kernel memory; it should be attributed to the user-space program. Even if there is enough memory, this allocation should fail once the user space reaches certain limits (e.g., the cgroup memory controller). 

It's important to investigate how to properly handle OOMs. However, due to the complexity of the task, I would like to see clear roadmaps for doing so correctly. Performing `try_with_capacity` in an ad hoc (and presumably inappropriate) way will not help at all.

For example, we can try to enable `cfg(no_global_oom_handling)`. But this means that we **cannot use**:
 - [`vec!`](https://github.com/rust-lang/rust/blob/4c62aa446803c8f31eafdc0c0119209ff129b3e0/library/alloc/src/macros.rs#L37), [`Vec::push`](https://github.com/rust-lang/rust/blob/4c62aa446803c8f31eafdc0c0119209ff129b3e0/library/alloc/src/vec/mod.rs#L2574), [`Vec::clone`](https://github.com/rust-lang/rust/blob/4c62aa446803c8f31eafdc0c0119209ff129b3e0/library/alloc/src/vec/mod.rs#L3656), ...
 - [`VecDeque`](https://github.com/rust-lang/rust/blob/4c62aa446803c8f31eafdc0c0119209ff129b3e0/library/alloc/src/collections/mod.rs#L14), [`BTreeMap`](https://github.com/rust-lang/rust/blob/4c62aa446803c8f31eafdc0c0119209ff129b3e0/library/alloc/src/collections/mod.rs#L17), [`BTreeSet`](https://github.com/rust-lang/rust/blob/4c62aa446803c8f31eafdc0c0119209ff129b3e0/library/alloc/src/collections/mod.rs#L25), ...
 - ...

Converting our entire codebase at once isn't possible. But I think we can fix it module by module. For modules that have been fixed, we can mark them as OOM-free modules. It's important to note that it's just _too_ easy to call methods that aren't marked as `no_global_oom_handling` and trigger OOM panics. Therefore, we should develop tools that check for these calls at compile time and reject them within OOM-free modules. Once all the modules are marked as OOM-free, we can pass the `--cfg no_global_oom_handling` flag to the Rust compiler.

tatetian: You are asking too many questions at once. Let's re-focus on our question at hand, which is quite simple: for this specific case, is `VecDeque::try_with_capacity` an improvement over `VecDeque::with_capacity`? My answer is YES.

In any universe that we are able to solve OOM or emulate `__GFP_ACCOUNT`, we have to use a _fallible_ constructor for a target collection type. In this specific case, it is `VecDeque::try_with_capacity`. I don't think a complete roadmap is needed to justify the use of fallible constructor.

Even before we find a systematic solution to handle OOM or emulate `__GFP_ACCOUNT`, it is still good to call `VecDeque::try_with_capacity` as it reduces the odds of kernel panicking due to kernel OOM.

We can't deny the merit of a small net-positive effort simply because much more efforts are needed to address the whole problem.

lrh2000: > for this specific case, is VecDeque::try_with_capacity an improvement over VecDeque::with_capacity?

But `VecDeque::try_with_capacity` is not necessarily an improvement over `VecDeque::new`, as it wastes memory to a large extent.
>> * `DEFAULT_MAX_QUEUED_EVENTS` is a large value representing the upper limit. Assuming the user space functions properly, this limit won't be reached. Therefore, allocating such a large buffer in advance would waste too much memory.

So why should we allocate so much memory here? This pre-allocation is not present in the Linux kernel implementation either. In fact, it uses a linked list, so it won't pre-allocate anything. Even for our `VecDeque` case, however, it's not well justified.

lrh2000: > I don't think a complete roadmap is needed to justify the use of fallible constructor.

We don't need to resolve all OOM problems in a single PR, but the first step in the roadmap can be: Go through the list of our `Vec.*::with_capacity` calls and, if applicable, convert them to `try_with_capacity`.

The number of `with_capacity` calls is not significant. I think it's okay to do it manually.
```
> grep -R "Vec.*::with_capacity" kernel | wc -l
52
> grep -R "Vec.*::with_capacity" ostd | wc -l
9
```

When reviewing that PR, we can establish some rules, such as whether to use `Vec::push_within_capacity` for `Vec`s created with `Vec::try_with_capacity` (there are concerns either way, but I don't think we need to continue that discussion now). The PR is part of the process of establishing new coding styles in our codebase, and we can follow it when creating or reviewing new PRs.

Even better, we should be able to add `Vec::with_capacity` to the [`disallowed_methods` list](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods) provided by Clippy. This would allow us to write `#[expect(disallowed_methods)] Vec::with_capacity` (with a TODO or an explanation of why the infallible version should be used), which would make the code clearer and reduce the maintenance burden.

What I don't like is that: We only have one `try_with_capacity` in our entire codebase. The rules are unclear, so it's impossible to create or review PRs.

tatetian: > But `VecDeque::try_with_capacity` is not necessarily an improvement over `VecDeque::new`, as it wastes memory to a large extent.

That's a good counter-argument. If we want to keep using `VecDeque` but not to preallocate a big chunk of memory, then we need to use `try_reserve` properly within the event enqueueing logic.

I don't believe real workloads would create many inotify files. So preallocation is not an emergent problem.

Anyway, now I see your point. I am fine with replacing `VecDeque::try_with_capacity` with `VecDeque::new`.

> When reviewing that PR, we can establish some rules, such as whether to use `Vec::push_within_capacity` for `Vec`s created with `Vec::try_with_capacity` (there are concerns either way, but I don't think we need to continue that discussion now). The PR is part of the process of establishing new coding styles in our codebase, and we can follow it when creating or reviewing new PRs.

It would take a whole documentation to address the problem of OOM. So let's pause on this topic for now.

tatetian (approved):

PR #2604, opened 2025-11-21
bugfix: Correctly extract the crate name from the path argument in execute_new_command()

jondo2010:
While trying to create a new project with osdk in a directory other than CWD, I discovered that `execute_new_command` was incorrectly assuming a single "crate_name". This PR fixes that, so now `cargo osdk new --kernel ~/path/to/my_crate` works correctly.

### DIFF
--- a/osdk/src/commands/new/mod.rs
+++ b/osdk/src/commands/new/mod.rs
@@ -12,7 +12,7 @@ use crate::{
 
 pub fn execute_new_command(args: &NewArgs) {
     cargo_new_lib(&args.crate_name);
-    
+
     // Extract the actual crate name from the path if a full path was provided.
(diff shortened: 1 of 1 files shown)


junyang-zh:
Thanks for your contribution! To align with the behavior of `cargo new`, this PR provides the right fix.

To fix the linter errors, you can run `make format`; amend the changes into your commit; and then do a force push.

junyang-zh:
Sorry. It seems that we can only rebase merge. @tatetian Do we have a "Squash and rebase merge" option? If there is, I think we can enable it.

tatetian:
> Sorry. It seems that we can only rebase merge. @tatetian Do we have a "Squash and rebase merge" option? If there is, I think we can enable it.

Done 😄

PR #2052, opened 2025-04-30
Don't race between enabling IRQs and halting CPU

lrh2000:
> I wonder if this is caused by the race condition between checking the preemption and halting the CPU. If the interrupts come in between, we halt the CPU and miss an opportunity to schedule to a worker task.
> 
> We can try to fix this race condition instead. Will https://github.com/lrh2000/asterinas/commit/31033ea89330ba430d647c26794f010f697df32b help this problem?  

 _Originally posted by @lrh2000 in [#2023](https://github.com/asterinas/asterinas/issues/2023#issuecomment-2809534654)_

~I don't know if this PR helps #2023 or not, since the comment above didn't get a response.~ This PR is not related to #2023 per https://github.com/asterinas/asterinas/pull/2052#issuecomment-2854850965.

But regardless of whether it helps, fixing the race condition should not be controversial.

PR #2605, opened 2025-11-22
Add a page table item reference type

junyang-zh:
Following #2595 , This PR renames the `Child` type into `PteState`, and replaces the raw representation in the type with `PageTableConfig::Item`, so that `PteState` fully tracks the ownership of the items pointed to by the PTE. Furthermore, `ChildRef` should contain a new `PageTableConfig::ItemRef` type, which is introduced in this PR.

By introducing `PageTableConfig::ItemRef` and allows the `query` API to return the reference, we also avoid some `<Frame as Clone>::clone` calls.

This PR is also based on #2744 though not relevant.

### DIFF
--- a/ostd/src/mm/page_table/cursor/mod.rs
+++ b/ostd/src/mm/page_table/cursor/mod.rs
@@ -233,7 +233,7 @@ impl<'rcu, C: PageTableConfig> Cursor<'rcu, C> {
 
                     let split_child = cur_entry
                         .split_if_mapped_huge(rcu_guard)
-                        .expect("The entry must be a huge page");
+                        .expect("the entry must be a huge page");
                     self.push_level(split_child);
                     continue;
                 }
@@ -296,7 +296,7 @@ impl<'rcu, C: PageTableConfig> Cursor<'rcu, C> {
     fn pop_level(&mut self) {
         let taken = self.path[self.level as usize - 1]
             .take()
-            .expect("Popping a level without a lock");
+            .expect("popping a level without a lock");
         let _ = ManuallyDrop::new(taken);
 
         debug_assert!(self.level < self.guard_level);
@@ -413,7 +413,11 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
         let (_, level, _) = C::item_raw_info(&item);
         assert!(level <= C::HIGHEST_TRANSLATION_LEVEL);
         let size = page_size::<C>(level);
-        assert_eq!(self.0.va % size, 0);
+        assert_eq!(
+            self.0.va % size,
+            0,
+            "cursor virtual address not aligned for mapping"
+        );
         let end = self.0.va + size;
         assert!(end <= self.0.barrier_va.end);
 
@@ -445,7 +449,7 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
         }
 
         if !matches!(self.0.cur_entry().to_ref(), PteStateRef::Absent) {
-            panic!("Mapping over an already mapped page at {:#x}", self.0.va);
+            panic!("mapping over an already mapped page");
         }
 
         let _ = self.replace_cur_entry(PteState::Mapped(RcuDrop::new(item)));
@@ -550,7 +554,7 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
 
                 if !C::TOP_LEVEL_CAN_UNMAP && level == C::NR_LEVELS {
                     let _ = ManuallyDrop::new(pt); // leak it to make shared PTs stay `'static`.
-                    panic!("Unmapping shared kernel page table nodes");
+                    panic!("unmapping shared kernel page table nodes");
                 }
 
                 // SAFETY: We must have locked this node.
--- a/ostd/src/mm/page_table/test.rs
+++ b/ostd/src/mm/page_table/test.rs
@@ -33,7 +33,7 @@ mod test_utils {
         unsafe {
             page_table
                 .cursor_mut(&preempt_guard, &virt_range)
-                .unwrap()
+                .expect("failed to create the cursor")
                 .map(VmItem::new_tracked(frame.into(), page_property))
         };
 
@@ -220,12 +220,12 @@ mod create_page_table {
             let user_entry = user_root.entry(i);
 
             let PteStateRef::PageTable(kernel_node) = kernel_entry.to_ref() else {
-                panic!("Expected a node reference at {} of kernel root PT", i);
+                panic!("expected a node reference at {} of kernel root PT", i);
             };
             assert_eq!(kernel_node.level(), PagingConsts::NR_LEVELS - 1);
 
             let PteStateRef::PageTable(user_node) = user_entry.to_ref() else {
-                panic!("Expected a node reference at {} of user root PT", i);
+                panic!("expected a node reference at {} of user root PT", i);
             };
             assert_eq!(user_node.level(), PagingConsts::NR_LEVELS - 1);
 
@@ -322,7 +322,7 @@ mod range_checks {
     }
 
     #[ktest]
-    #[should_panic]
+    #[should_panic(expected = "failed to create the cursor")]
     fn overflow_boundary_mapping() {
         let virt_range =
             (MAX_USERSPACE_VADDR - (PAGE_SIZE / 2))..(MAX_USERSPACE_VADDR + (PAGE_SIZE / 2));
@@ -439,7 +439,7 @@ mod overlapping_mappings {
     use super::{test_utils::*, *};
 
     #[ktest]
-    #[should_panic]
+    #[should_panic(expected = "mapping over an already mapped page")]
     fn overlapping_mappings() {
         let page_table = PageTable::<TestPtConfig>::empty();
         let vrange1 = PAGE_SIZE..(PAGE_SIZE * 2);
@@ -466,7 +466,7 @@ mod overlapping_mappings {
     }
 
     #[ktest]
-    #[should_panic]
+    #[should_panic(expected = "cursor virtual address not aligned for mapping")]
     fn unaligned_map() {
         let page_table = PageTable::<TestPtConfig>::empty();
         let virt_range = (PAGE_SIZE + 512)..(PAGE_SIZE * 2 + 512);
@@ -536,7 +536,7 @@ mod navigation {
         cursor.jump(FIRST_MAP_ADDR).unwrap();
         assert_eq!(cursor.virt_addr(), FIRST_MAP_ADDR);
         let (queried_va, Some(queried_item)) = cursor.query().unwrap() else {
-            panic!("Expected a mapped item at the first address");
+            panic!("expected a mapped item at the first address");
         };
         assert_eq!(queried_va, FIRST_MAP_ADDR..FIRST_MAP_ADDR + PAGE_SIZE);
         let TestPtItemRef((pa, _, prop), _) = queried_item;
@@ -617,7 +617,7 @@ mod navigation {
         assert_eq!(cursor.virt_addr(), 0);
 
         let Some(va) = cursor.find_next(FIRST_MAP_ADDR + PAGE_SIZE) else {
-            panic!("Expected to find the next mapping");
+            panic!("expected to find the next mapping");
         };
         assert_eq!(va, FIRST_MAP_ADDR);
         assert_eq!(cursor.virt_addr(), FIRST_MAP_ADDR);
@@ -625,7 +625,7 @@ mod navigation {
         cursor.jump(FIRST_MAP_ADDR + PAGE_SIZE).unwrap();
 
         let Some(va) = cursor.find_next(SECOND_MAP_ADDR - FIRST_MAP_ADDR) else {
-            panic!("Expected to find the next mapping");
+            panic!("expected to find the next mapping");
         };
         assert_eq!(va, SECOND_MAP_ADDR);
         assert_eq!(cursor.virt_addr(), SECOND_MAP_ADDR);
@@ -656,7 +656,7 @@ mod unmap {
         let Some(PageTableFrag::Mapped { va, item }) =
             (unsafe { cursor.take_next(virt_range.len()) })
         else {
-            panic!("Expected to take a mapped item");
+            panic!("expected to take a mapped item");
         };
 
         assert_eq!(va, virt_range.start);
@@ -690,7 +690,7 @@ mod unmap {
             num_frames,
         }) = (unsafe { cursor.take_next(large_range.len()) })
         else {
-            panic!("Expected to take a stray page table");
+            panic!("expected to take a stray page table");
         };
 
         // Should take a level-2 page table with 512 entries.
@@ -743,7 +743,7 @@ mod mapping {
             let Some(PageTableFrag::Mapped { va: frag_va, item }) =
                 (unsafe { cursor.take_next(unmap_len) })
             else {
-                panic!("Expected to unmap a page, but got `None`");
+                panic!("expected to unmap a page, but got `None`");
             };
 
             // Calculates the expected PA for the unmapped item.
@@ -801,7 +801,7 @@ mod mapping {
                 let (va, item) = cursor.query().unwrap();
 
                 let Some(TestPtItemRef((pa, level, prop), _)) = item else {
-                    panic!("Expected mapped untracked physical address, got `None`");
+                    panic!("expected mapped untracked physical address, got `None`");
                 };
 
                 assert_eq!(pa, mapped_pa_of_va(va.start));
@@ -843,7 +843,7 @@ mod mapping {
             let va_low = protect_va_range.start - PAGE_SIZE;
             let (va_low_pa, prop_low) = pt
                 .page_walk(va_low)
-                .expect("Page should be mapped before protection");
+                .expect("page should be mapped before protection");
             assert_eq!(va_low_pa, mapped_pa_of_va(va_low));
             assert_eq!(
                 prop_low,
@@ -858,7 +858,7 @@ mod mapping {
                 let (va, item) = cursor.query().unwrap();
 
                 let Some(TestPtItemRef((pa, level, prop), _)) = item else {
-                    panic!("Expected mapped untracked physical address, got `None`");
+                    panic!("expected mapped untracked physical address, got `None`");
                 };
 
                 assert_eq!(pa, mapped_pa_of_va(va.start));
@@ -877,7 +877,7 @@ mod mapping {
             let va_high = protect_va_range.end;
             let (va_high_pa, prop_high) = pt
                 .page_walk(va_high)
-                .expect("Page should be mapped after protection");
+                .expect("page should be mapped after protection");
             assert_eq!(va_high_pa, mapped_pa_of_va(va_high));
             assert_eq!(
                 prop_high,
@@ -912,7 +912,7 @@ mod protection_and_query {
             let va_to_check = PAGE_SIZE * i;
             let (_, prop) = page_table
                 .page_walk(va_to_check)
-                .expect("Mapping should exist");
+                .expect("mapping should exist");
             assert_eq!(prop.flags, PageFlags::RW);
             assert_eq!(prop.cache, CachePolicy::Writeback);
         }
@@ -928,7 +928,7 @@ mod protection_and_query {
             let va_to_check = PAGE_SIZE * i;
             let (_, prop) = page_table
                 .page_walk(va_to_check)
-                .expect("Mapping should exist");
+                .expect("mapping should exist");
             assert_eq!(prop.flags, PageFlags::R);
             assert_eq!(prop.cache, CachePolicy::Writeback);
         }
@@ -1013,7 +1013,7 @@ mod boot_pt {
     }
 
     #[ktest]
-    #[should_panic]
+    #[should_panic(expected = "mapping an already mapped huge page in the boot page table")]
     fn map_base_page_already_mapped() {
         let root_frame = FrameAllocOptions::new().alloc_frame().unwrap();
         let root_paddr = root_frame.paddr();
@@ -1031,7 +1031,7 @@ mod boot_pt {
     }
 
     #[ktest]
-    #[should_panic]
+    #[should_panic(expected = "protecting an unmapped page in the boot page table")]
     fn protect_base_page_unmapped() {
         let root_frame = FrameAllocOptions::new().alloc_frame().unwrap();
         let root_paddr = root_frame.paddr();
--- a/ostd/src/mm/test.rs
+++ b/ostd/src/mm/test.rs
@@ -523,7 +523,7 @@ mod vmspace {
         let preempt_guard = disable_preempt();
         let mut cursor = vmspace
             .cursor(&preempt_guard, &range)
-            .expect("Failed to create cursor");
+            .expect("failed to create the cursor");
         assert!(matches!(
             cursor.query(),
             Ok((r, None)) if r == range
@@ -542,7 +542,7 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             // Initially, the page should not be mapped.
             assert!(matches!(
                 cursor_mut.query().unwrap(),
@@ -556,7 +556,7 @@ mod vmspace {
         {
             let mut cursor = vmspace
                 .cursor(&preempt_guard, &range)
-                .expect("Failed to create cursor");
+                .expect("failed to create the cursor");
             assert_eq!(cursor.virt_addr(), range.start);
             assert_matches_mapped!(cursor, range.clone(), frame, prop);
         }
@@ -564,7 +564,7 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             // Unmaps the frame.
             cursor_mut.unmap(range.start);
         }
@@ -572,7 +572,7 @@ mod vmspace {
         // Queries to ensure it's unmapped.
         let mut cursor = vmspace
             .cursor(&preempt_guard, &range)
-            .expect("Failed to create cursor");
+            .expect("failed to create the cursor");
         assert!(matches!(
             cursor.query().unwrap(),
             (r, None) if r == range
@@ -581,7 +581,7 @@ mod vmspace {
 
     /// Maps a page twice and unmaps twice using `CursorMut`.
     #[ktest]
-    #[should_panic = "Mapping over an already mapped page at 0x1000"]
+    #[should_panic(expected = "mapping over an already mapped page")]
     fn vmspace_map_twice() {
         let vmspace = VmSpace::default();
         let range = 0x1000..0x2000;
@@ -592,21 +592,21 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.map(frame.clone(), prop);
         }
 
         {
             let mut cursor = vmspace
                 .cursor(&preempt_guard, &range)
-                .expect("Failed to create cursor");
+                .expect("failed to create the cursor");
             assert_matches_mapped!(cursor, range.clone(), frame, prop);
         }
 
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.map(frame.clone(), prop);
         }
     }
@@ -623,27 +623,27 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.map(frame.clone(), prop);
         }
 
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.unmap(range.start);
         }
 
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.unmap(range.start);
         }
 
         let mut cursor = vmspace
             .cursor(&preempt_guard, &range)
-            .expect("Failed to create cursor");
+            .expect("failed to create the cursor");
         assert!(matches!(
             cursor.query().unwrap(),
             (r, None) if r == range
@@ -677,7 +677,7 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.map(frame.clone(), prop);
         }
 
@@ -685,7 +685,7 @@ mod vmspace {
             // Verifies that the mapping exists.
             let mut cursor = vmspace
                 .cursor(&preempt_guard, &range)
-                .expect("Failed to create cursor");
+                .expect("failed to create the cursor");
             assert_matches_mapped!(cursor, range.clone(), frame, prop);
         }
 
@@ -693,7 +693,7 @@ mod vmspace {
             // Flushes the TLB using a mutable cursor.
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             cursor_mut.flusher().issue_tlb_flush(TlbFlushOp::for_all());
             cursor_mut.flusher().dispatch_tlb_flush();
         }
@@ -702,7 +702,7 @@ mod vmspace {
             // Verifies that the mapping still exists.
             let mut cursor = vmspace
                 .cursor(&preempt_guard, &range)
-                .expect("Failed to create cursor");
+                .expect("failed to create the cursor");
             assert_matches_mapped!(cursor, range.clone(), frame, prop);
         }
     }
@@ -716,7 +716,7 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             let frame = create_dummy_frame();
             let prop = PageProperty::new_user(PageFlags::R, CachePolicy::Writeback);
             cursor_mut.map(frame, prop);
@@ -763,7 +763,7 @@ mod vmspace {
         // Creates the first cursor.
         let _cursor1 = vmspace
             .cursor(&preempt_guard, &range1)
-            .expect("Failed to create first cursor");
+            .expect("failed to create first cursor");
 
         // Attempts to create the second overlapping cursor.
         let cursor2_result = vmspace.cursor(&preempt_guard, &range2);
@@ -780,10 +780,10 @@ mod vmspace {
         {
             let mut cursor_mut = vmspace
                 .cursor_mut(&preempt_guard, &range)
-                .expect("Failed to create mutable cursor");
+                .expect("failed to create the mutable cursor");
             let prop = PageProperty::new_user(PageFlags::RW, CachePolicy::Writeback);
             cursor_mut.map(frame.clone(), prop);
-            cursor_mut.jump(range.start).expect("Failed to jump cursor");
+            cursor_mut.jump(range.start).expect("failed to jump cursor");
             let protected_range = cursor_mut.protect_next(0x1000, |flags, _cache| {
                 *flags = PageFlags::R;
             });
@@ -793,7 +793,7 @@ mod vmspace {
         // Confirms that the property was updated.
         let mut cursor = vmspace
             .cursor(&preempt_guard, &range)
-            .expect("Failed to create cursor");
+            .expect("failed to create the cursor");
         assert_matches_mapped!(
             cursor,
             range.clone(),
@@ -811,7 +811,7 @@ mod vmspace {
         let preempt_guard = disable_preempt();
         let mut cursor_mut = vmspace
             .cursor_mut(&preempt_guard, &range)
-            .expect("Failed to create mutable cursor");
+            .expect("failed to create the mutable cursor");
         cursor_mut.unmap(0x800); // Not page-aligned.
     }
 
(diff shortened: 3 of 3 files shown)


ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
```diff
@@ -26,17 +28,29 @@ unsafe impl PageTableConfig for IommuPtConfig {
...
+        paddr: Paddr,
+        level: PagingLevel,
+        prop: PageProperty,
+    ) -> Self::ItemRef<'a> {
+        PtItemRef((paddr, level, prop), PhantomData)
+    }
 }
 
+pub(crate) type PtItem = (Paddr, PagingLevel, PageProperty);
+pub(crate) struct PtItemRef<'a>(#[expect(dead_code)] pub PtItem, PhantomData<&'a ()>);
```
lrh2000: So by `#[expect(dead_code)]`, no one actually uses the reference type?

Another case is:
```rust
#[derive(Debug)]
pub(crate) enum MappedItemRef<'a> {
    #[expect(dead_code)]
    Tracked(FrameRef<'a, dyn AnyFrameMeta>, PageProperty),
    #[expect(dead_code)]
    Untracked(Paddr, PagingLevel, PageProperty),
}
```

For the two cases, what about doing something simpler:
```rust
    type Item = (Paddr, PagingLevel, PageProperty);
    type ItemRef<'a> = PhantomData<'a>;

    unsafe fn item_ref_from_raw<'a>(
        _paddr: Paddr,
        _level: PagingLevel,
        _prop: PageProperty,
    ) -> Self::ItemRef<'a> {
        PhantomData
    }
```

So we can avoid these dead code warnings?

ostd/src/mm/page_table/mod.rs
```diff
@@ -24,6 +24,7 @@ use crate::{
 };
 
 mod node;
+
 use node::*;
 mod cursor;
```
lrh2000: Why inserting a blank line here? I suppose the following is more reasonable:
```rust
mod node;
use node::*;

mod cursor;
```

Besides, we should avoid wildcard imports. However, we can address the issue later, since it's not introduced in this PR.

ostd/src/mm/page_table/mod.rs
```diff
@@ -55,12 +56,16 @@ pub enum PageTableError {
...
+/// The implementor must ensure that `item_raw_info`, `item_into_raw`,
+/// `item_from_raw` and `item_ref_from_raw` are implemented correctly so that:
+///  - `item_into_raw` is not overridden;
+///  - `item_raw_info` returns the exact same values as if called with
+///    `item_into_raw`;
 ///  - if the provided raw form matches the item that was consumed by
 ///    `item_into_raw`, `item_from_raw` restores the exact item that was
 ///    consumed by `item_into_raw`.
+///  - `Item` and `ItemRef` can be converted to each other (by either borrowing
+///    or cloning) without changing the raw form.
```
lrh2000: Rewrite this sentence: "If ... can be converted, ...", as `PhantomData` cannot be converted to a page table item.

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -523,8 +495,10 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     ///
     /// # Safety
     ///
-    /// The caller should ensure that the range being unmapped does not affect
-    /// kernel's memory safety.
+    /// The caller should ensure that:
+    ///  - the range being unmapped does not affect kernel's memory safety.
+    ///  - the items mapped in `PageTableFrag` must outlive `'rcu` and any TLB
+    ///    entries that caches the mappings in it.
```
lrh2000: For the doc comments of `map` and `take_next`:
```suggestion
    ///    entries that cache the mappings.
````

I think you want to say "TLB entries that cache the mappings", right? However, "entries" is plural, so "caches" and "in it" are perhaps incorrect.

File: ostd/src/mm/page_table/node/pte_state.rs [hunk 5]
```diff
@@ -21,11 +20,9 @@ use crate::{
...
     /// A child page table node.
     PageTable(RcuDrop<PageTableNode<C>>),
-    /// Physical address of a mapped physical frame.
-    ///
-    /// It is associated with the virtual page property and the level of the
-    /// mapping node, which decides the size of the frame.
-    Mapped(Paddr, PagingLevel, PageProperty),
+    /// A mapped item, often a mapped physical frame. The actual type is
+    /// defined by the page table configuration.
+    Mapped(C::Item),
```
lrh2000: To avoid dropping the item too early, it should be changed to `RcuDrop(C::Item)`?

This allows us to remove the `'rcu` lifetime safety requirement at `map()`/`take_next()`.

We should still be able to group multiple frames to one `RcuDrop` in `TlbFlusher`, by adding a (unsafe) `RcuDrop::into_inner`.

ostd/src/mm/vm_space.rs
```diff
@@ -691,28 +694,31 @@ impl From<VmItem> for VmQueriedItem {
 #[derive(Clone, Debug)]
 pub(crate) struct UserPtConfig {}
 
-// SAFETY: `item_into_raw` and `item_from_raw` are implemented correctly,
+// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw` and
```
lrh2000:
```
sed -i 's|`item_from_raw` and|`item_from_raw`, and|g' ostd/**/*.rs
```
gives:
```patch
diff --git a/ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs b/ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
index 10f32ed21..34c379cb1 100644
--- a/ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
+++ b/ostd/src/arch/x86/iommu/dma_remapping/second_stage.rs
@@ -16,7 +16,7 @@ use crate::{
 #[derive(Clone, Debug)]
 pub struct IommuPtConfig {}
 
-// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw` and
+// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw`, and
 // `item_ref_from_raw` are correctly implemented with respect to the `Item` and
 // `ItemRef` types.
 unsafe impl PageTableConfig for IommuPtConfig {
diff --git a/ostd/src/mm/kspace/mod.rs b/ostd/src/mm/kspace/mod.rs
index f2fdc85c8..3ffc6872d 100644
--- a/ostd/src/mm/kspace/mod.rs
+++ b/ostd/src/mm/kspace/mod.rs
@@ -141,7 +141,7 @@ pub(super) static KERNEL_PAGE_TABLE: Once<PageTable<KernelPtConfig>> = Once::new
 pub(super) struct KernelPtConfig {}
 
 // We use the first available PTE bit to mark the frame as tracked.
-// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw` and
+// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw`, and
 // `item_ref_from_raw` are correctly implemented with respect to the `Item` and
 // `ItemRef` types.
 unsafe impl PageTableConfig for KernelPtConfig {
diff --git a/ostd/src/mm/page_table/mod.rs b/ostd/src/mm/page_table/mod.rs
index 764b8bbee..eb6fdc926 100644
--- a/ostd/src/mm/page_table/mod.rs
+++ b/ostd/src/mm/page_table/mod.rs
@@ -57,7 +57,7 @@ pub enum PageTableError {
 /// # Safety
 ///
 /// The implementor must ensure that `item_raw_info`, `item_into_raw`,
-/// `item_from_raw` and `item_ref_from_raw` are implemented correctly so that:
+/// `item_from_raw`, and `item_ref_from_raw` are implemented correctly so that:
 ///  - `item_into_raw` is not overridden;
 ///  - `item_raw_info` returns the exact same values as if called with
 ///    `item_into_raw`;
diff --git a/ostd/src/mm/page_table/test.rs b/ostd/src/mm/page_table/test.rs
index df2d50e3d..42ef92de4 100644
--- a/ostd/src/mm/page_table/test.rs
+++ b/ostd/src/mm/page_table/test.rs
@@ -84,7 +84,7 @@ mod test_utils {
     #[derive(Clone, Debug)]
     pub struct TestPtConfig;
 
-    // SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw` and
+    // SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw`, and
     // `item_ref_from_raw` are correctly implemented with respect to the `Item`
     // and `ItemRef` types.
     unsafe impl PageTableConfig for TestPtConfig {
diff --git a/ostd/src/mm/vm_space.rs b/ostd/src/mm/vm_space.rs
index 624d59e7d..6cf6930ef 100644
--- a/ostd/src/mm/vm_space.rs
+++ b/ostd/src/mm/vm_space.rs
@@ -652,7 +652,7 @@ impl<'a> From<VmItemRef<'a>> for VmQueriedItem<'a> {
 #[derive(Clone, Debug)]
 pub(crate) struct UserPtConfig {}
 
-// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw` and
+// SAFETY: `item_raw_info`, `item_into_raw`, `item_from_raw`, and
 // `item_ref_from_raw` are correctly implemented with respect to the `Item` and
 // `ItemRef` types.
 unsafe impl PageTableConfig for UserPtConfig {
```

ostd/src/mm/page_table/node/entry.rs
```diff
@@ -178,7 +179,10 @@ impl<'a, 'rcu, C: PageTableConfig> Entry<'a, 'rcu, C> {
         for i in 0..nr_subpage_per_huge::<C>() {
             let small_pa = pa + i * page_size::<C>(level - 1);
             let mut entry = pt_lock_guard.entry(i);
-            let old = entry.replace(PteState::Mapped(small_pa, level - 1, prop));
+            // SAFETY: It's a part of the mapped item, and the ownership is
+            // properly transferred to the new sub-entry.
+            let small_item = unsafe { C::item_from_raw(small_pa, level - 1, prop) };
```
lrh2000, line 188: This may not be a problem introduced by this PR, but it doesn't seem sound to panic in the middle of the `for` loop. We will drop `small_item` twice because it has not been taken out from the page table?

ostd/src/mm/vm_space.rs
```diff
@@ -729,4 +735,25 @@ unsafe impl PageTableConfig for UserPtConfig {
...
+        prop: PageProperty,
+    ) -> Self::ItemRef<'a> {
+        debug_assert_eq!(level, 1);
+        if prop.priv_flags.contains(PrivilegedPageFlags::AVAIL1) {
+            VmItemRef {
+                prop,
+                mapped_item: MappedItemRef::UntrackedIoMem { paddr, level },
+            }
+        } else {
+            // SAFETY: The caller ensures that the lifetime is valid.
```
lrh2000:
```suggestion
            // SAFETY: The caller ensures that the frame outlives `'a` and that
            // the type matches the frame.
```

Besides, also add the following comments to be consistent with `item_from_raw`?
```
            // AVAIL1 is set, this is I/O memory.
```
```
            // AVAIL1 is clear, this is tracked memory.
```

ostd/src/mm/vm_space.rs
```diff
@@ -646,12 +638,23 @@ pub(crate) struct VmItem {
...
+    mapped_item: MappedItemRef<'a>,
+}
+
 #[derive(Debug, Clone, PartialEq)]
 enum MappedItem {
     TrackedFrame(UFrame),
     UntrackedIoMem { paddr: Paddr, level: PagingLevel },
 }
 
+enum MappedItemRef<'a> {
```
lrh2000, line 626: Derive some traits like `VmItem` and `MappedItem`?

I think `Debug`/`Clone` is feasible. Not sure about `PartialEq`.

ostd/src/mm/tlb.rs
```diff
@@ -381,7 +388,7 @@ impl OpsStack {
     fn clear_without_flush(&mut self) {
         self.num_pages_to_flush = 0;
         self.num_ops = 0;
-        self.page_keeper.clear();
+        let _ = RcuDrop::new(core::mem::take(&mut self.frame_keeper));
```
lrh2000: Note that we also need to do this in `Drop` if we change `TlbFlusher` to accept `RcuDrop`. However, we may want to check `self.frame_keeper.is_empty()` to avoid unnecessary `RcuDrop` instance creation.

ostd/src/mm/page_table/test.rs
```diff
@@ -438,6 +437,7 @@ mod overlapping_mappings {
     use super::{test_utils::*, *};
 
     #[ktest]
+    #[should_panic]
```
lrh2000: Add a panic message to test here? (If it is feasible, I am not sure).

Same to another `#[should_panic]` in `mm/test.rs`.

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -406,29 +393,22 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
...
     /// Maps the item starting from the current address to a physical address range.
     ///
-    /// If the current address has already mapped pages, it will do a re-map,
-    /// taking out the old physical address and replacing it with the new one.
-    /// This function will return [`Err`] with a [`PageTableFrag`], the not
-    /// mapped item. The caller should drop it after TLB coherence.
-    ///
-    /// If there is no mapped pages in the specified virtual address range,
-    /// the function will return [`None`].
+    /// The current virtual address must not be mapped.
```
lrh2000:
```suggestion
    /// The current virtual address should not be mapped.
```

I think we should use `must` in the `# Safety` section. However, since this file uses `should` in that section, I doubt we should use `must` outside of it.

ostd/src/mm/vm_space.rs
junyang-zh: `FrameRef` can be `Clone`, but we chose not to implement it. So I add `Debug` here.

ostd/src/mm/page_table/node/entry.rs
junyang-zh: `debug_assert!(old.is_absent());` may panic. `Entry::replace` also checks the level and may panic. But they should never panic since the page table is newly allocated and the level is correct. I wonder if we should remove the `debug_assert`.

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -523,8 +493,10 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     ///
     /// # Safety
     ///
-    /// The caller should ensure that the range being unmapped does not affect
-    /// kernel's memory safety.
+    /// The caller should ensure that:
+    ///  - the range being unmapped does not affect kernel's memory safety.
+    ///  - the items mapped in `PageTableFrag` must outlive `'rcu` and any TLB
```
lrh2000: No need to mention `'rcu` here as the item is already in `RcuDrop`?

ostd/src/sync/rcu/mod.rs
```diff
@@ -465,6 +465,19 @@ impl<T: Send + 'static> RcuDrop<T> {
             value: ManuallyDrop::new(value),
         }
     }
+
+    /// Extracts the value from the `RcuDrop` container.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the returned value will be moved or dropped
```
lrh2000: > be moved

No idea why it cannot be moved? We certainly move `Frame` to the TLB flusher in practice.

ostd/src/mm/page_table/node/mod.rs
```diff
@@ -262,7 +262,7 @@ impl<C: PageTableConfig> Drop for PageTableGuard<'_, C> {
 /// The metadata of any kinds of page table pages.
 /// Make sure the the generic parameters don't effect the memory layout.
 #[derive(Debug)]
-pub(in crate::mm) struct PageTablePageMeta<C: PageTableConfig> {
+pub(crate) struct PageTablePageMeta<C: PageTableConfig> {
```
lrh2000, line 265: I am not sure why your lastest commits change visibility here? The reason is not obvious to me. Not sure if I'm missing anything:
```patch
-pub(super) type PageTableNode<C> = Frame<PageTablePageMeta<C>>;
+pub(crate) type PageTableNode<C> = Frame<PageTablePageMeta<C>>;
```
```patch
-pub(in crate::mm) struct PageTablePageMeta<C: PageTableConfig> {
+pub(crate) struct PageTablePageMeta<C: PageTableConfig> {
```

ostd/src/mm/kspace/mod.rs
```diff
@@ -178,6 +181,21 @@ unsafe impl PageTableConfig for KernelPtConfig {
...
     }
+
+    unsafe fn item_ref_from_raw<'a>(
+        paddr: Paddr,
+        level: PagingLevel,
+        prop: PageProperty,
+    ) -> Self::ItemRef<'a> {
+        if prop.priv_flags.contains(PrivilegedPageFlags::AVAIL1) {
+            debug_assert_eq!(level, 1);
+            // SAFETY: The caller ensures that the lifetime is valid.
```
lrh2000:
```rust
            // SAFETY: The caller ensures that the frame outlives `'a` and that
            // the type matches the frame.
```

ostd/src/mm/vm_space.rs
```diff
@@ -646,12 +648,24 @@ pub(crate) struct VmItem {
     mapped_item: MappedItem,
 }
 
+/// A reference to a VM item.
+pub(crate) struct VmItemRef<'a> {
```
lrh2000, line 614: Also add `#[dervie(Debug)]` to this type?

ostd/src/mm/page_table/node/entry.rs
lrh2000: > But they should never panic

It seems to be the case. I can't think of an elegant solution to the problem. I'm fine with the current form, especially since this PR didn't introduce the problem.

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -445,7 +449,7 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
         }
 
         if !matches!(self.0.cur_entry().to_ref(), PteStateRef::Absent) {
-            panic!("Mapping over an already mapped page at {:#x}", self.0.va);
+            panic!("mapping over an already mapped page");
```
lrh2000, line 452: Update these messages as well:
```
> grep -R "\"[A-Z]" "ostd/src/mm/page_table/cursor/mod.rs"
                        .expect("The entry must be a huge page");
            .expect("Popping a level without a lock");
                    panic!("Unmapping shared kernel page table nodes");
```

ostd/src/mm/page_table/node/mod.rs
lrh2000: Oh, I just noticed that it's because the type is used in `PageTableFrag`.

Perhaps you want to restrict the methods/the fields' visibility:
https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/ostd/src/mm/page_table/node/mod.rs#L94
https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/ostd/src/mm/page_table/node/mod.rs#L283
https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/ostd/src/mm/page_table/node/mod.rs#L266-L278

ostd/src/mm/page_table/node/entry.rs
lrh2000: Perhaps we could add a panic guard here.
```rust
struct PanicGuard;
impl Drop for PanicGuard {
   fn drop(&mut self) {
       // This prevents page table items from being dropped twice during the
       // transfer of ownership.
       crate::panic::abort();
   }
}
```

An infinite loop would also help if `abort` is considered too aggressive.

It's similar to what was done in the Rust standard library.
https://github.com/rust-lang/rust/blob/9415853279dda1394c1f7d4114e1fe1e3ced76ec/library/alloc/src/collections/btree/mem.rs#L19-L24

ostd/src/mm/page_table/node/mod.rs
lrh2000: Perhaps you don't really address https://github.com/asterinas/asterinas/pull/2605#discussion_r2721337624, I guess.

ostd/src/mm/vm_space.rs
```diff
@@ -507,15 +501,20 @@ impl<'a> CursorMut<'a> {
         let end_va = self.virt_addr() + len;
         let mut num_unmapped: usize = 0;
         loop {
-            // SAFETY: It is safe to un-map memory in the userspace.
+            // SAFETY: It is safe to un-map memory in the userspace. And the
+            // un-mapped items are dropped after TLB flushes. So they outlive
+            // the current RCU lifetime `'a`.
```
lrh2000: No need to mention the RCU grace period for `take_next` anymore.
```rust
            // SAFETY: It is safe to un-map memory in the userspace. The
            // un-mapped items are dropped after TLB flushes.
```

ostd/src/mm/tlb.rs
```diff
@@ -324,7 +331,7 @@ impl OpsStack {
 
     fn push(&mut self, op: TlbFlushOp, drop_after_flush: Option<Frame<dyn AnyFrameMeta>>) {
         if let Some(frame) = drop_after_flush {
-            self.page_keeper.push(frame);
+            self.frame_keeper.push(frame);
```
lrh2000, line 337: If this `push` pancis (e.g., because there is no memory?), `frame` will be dropped immediately.

I think we may want to change `issue_tlb_flush_with` and `push` to accept `RcuDrop<>`. Then we want the something like:
```rust
let panic_guard = PanicGuard;
self.frame_keeper.push(RcuDrop::into_inner(...));
core::mem::forget(panic_guard);
```

ostd/src/sync/rcu/mod.rs
```diff
@@ -465,6 +465,19 @@ impl<T: Send + 'static> RcuDrop<T> {
...
         }
     }
+
+    /// Extracts the value from the `RcuDrop` container.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the returned value will be dropped after
+    /// all the threads cannot access it anymore. Specifically, dropping it
+    /// after the RCU grace period must be safe.
```
lrh2000: Not sure why we cannot simplify this to:
```rust
    /// The caller must ensure that the returned value will be dropped after
    /// after the RCU grace period.
```

If the current form is preferred, I think "must be safe" should be replaced with "is guaranteed to be safe." It tells something that the caller can do instead of requiring the caller to do something.

In addition, add a note about the panics:
```rust
    /// Note that panic unwinding may cause the returned value to be dropped
    /// immediately, which is not sound. Therefore, the caller must pay extra
    /// attention to panics when using this method.
```

ostd/src/sync/rcu/mod.rs
```diff
@@ -465,6 +465,19 @@ impl<T: Send + 'static> RcuDrop<T> {
...
+    pub unsafe fn into_inner(slot: RcuDrop<T>) -> T {
```
lrh2000: `pub(crate)` since it's unsafe?

ostd/src/mm/page_table/test.rs
lrh2000, line 1: Missed updates in the last commit for the two `test.rs` files:
```patch
diff --git a/ostd/src/mm/page_table/test.rs b/ostd/src/mm/page_table/test.rs
index 1b2930cae..6ef5e9356 100644
--- a/ostd/src/mm/page_table/test.rs
+++ b/ostd/src/mm/page_table/test.rs
@@ -218,12 +218,12 @@ mod create_page_table {
             let user_entry = user_root.entry(i);
 
             let PteStateRef::PageTable(kernel_node) = kernel_entry.to_ref() else {
-                panic!("Expected a node reference at {} of kernel root PT", i);
+                panic!("expected a node reference at {} of kernel root PT", i);
             };
             assert_eq!(kernel_node.level(), PagingConsts::NR_LEVELS - 1);
 
             let PteStateRef::PageTable(user_node) = user_entry.to_ref() else {
-                panic!("Expected a node reference at {} of user root PT", i);
+                panic!("expected a node reference at {} of user root PT", i);
             };
             assert_eq!(user_node.level(), PagingConsts::NR_LEVELS - 1);
 
@@ -534,7 +534,7 @@ mod navigation {
         cursor.jump(FIRST_MAP_ADDR).unwrap();
         assert_eq!(cursor.virt_addr(), FIRST_MAP_ADDR);
         let (queried_va, Some(queried_item)) = cursor.query().unwrap() else {
-            panic!("Expected a mapped item at the first address");
+            panic!("expected a mapped item at the first address");
         };
         assert_eq!(queried_va, FIRST_MAP_ADDR..FIRST_MAP_ADDR + PAGE_SIZE);
         let TestPtItemRef((pa, _, prop), _) = queried_item;
@@ -615,7 +615,7 @@ mod navigation {
         assert_eq!(cursor.virt_addr(), 0);
 
         let Some(va) = cursor.find_next(FIRST_MAP_ADDR + PAGE_SIZE) else {
-            panic!("Expected to find the next mapping");
+            panic!("expected to find the next mapping");
         };
         assert_eq!(va, FIRST_MAP_ADDR);
         assert_eq!(cursor.virt_addr(), FIRST_MAP_ADDR);
@@ -623,7 +623,7 @@ mod navigation {
         cursor.jump(FIRST_MAP_ADDR + PAGE_SIZE).unwrap();
 
         let Some(va) = cursor.find_next(SECOND_MAP_ADDR - FIRST_MAP_ADDR) else {
-            panic!("Expected to find the next mapping");
+            panic!("expected to find the next mapping");
         };
         assert_eq!(va, SECOND_MAP_ADDR);
         assert_eq!(cursor.virt_addr(), SECOND_MAP_ADDR);
@@ -654,7 +654,7 @@ mod unmap {
         let Some(PageTableFrag::Mapped { va, item }) =
             (unsafe { cursor.take_next(virt_range.len()) })
         else {
-            panic!("Expected to take a mapped item");
+            panic!("expected to take a mapped item");
         };
 
         assert_eq!(va, virt_range.start);
@@ -688,7 +688,7 @@ mod unmap {
             num_frames,
         }) = (unsafe { cursor.take_next(large_range.len()) })
         else {
-            panic!("Expected to take a stray page table");
+            panic!("expected to take a stray page table");
         };
 
         // Should take a level-2 page table with 512 entries.
@@ -741,7 +741,7 @@ mod mapping {
             let Some(PageTableFrag::Mapped { va: frag_va, item }) =
                 (unsafe { cursor.take_next(unmap_len) })
             else {
-                panic!("Expected to unmap a page, but got `None`");
+                panic!("expected to unmap a page, but got `None`");
             };
 
             // Calculates the expected PA for the unmapped item.
@@ -799,7 +799,7 @@ mod mapping {
                 let (va, item) = cursor.query().unwrap();
 
                 let Some(TestPtItemRef((pa, level, prop), _)) = item else {
-                    panic!("Expected mapped untracked physical address, got `None`");
+                    panic!("expected mapped untracked physical address, got `None`");
                 };
 
                 assert_eq!(pa, mapped_pa_of_va(va.start));
@@ -841,7 +841,7 @@ mod mapping {
             let va_low = protect_va_range.start - PAGE_SIZE;
             let (va_low_pa, prop_low) = pt
                 .page_walk(va_low)
-                .expect("Page should be mapped before protection");
+                .expect("page should be mapped before protection");
             assert_eq!(va_low_pa, mapped_pa_of_va(va_low));
             assert_eq!(
                 prop_low,
@@ -856,7 +856,7 @@ mod mapping {
                 let (va, item) = cursor.query().unwrap();
 
                 let Some(TestPtItemRef((pa, level, prop), _)) = item else {
-                    panic!("Expected mapped untracked physical address, got `None`");
+                    panic!("expected mapped untracked physical address, got `None`");
                 };
 
                 assert_eq!(pa, mapped_pa_of_va(va.start));
@@ -875,7 +875,7 @@ mod mapping {
             let va_high = protect_va_range.end;
             let (va_high_pa, prop_high) = pt
                 .page_walk(va_high)
-                .expect("Page should be mapped after protection");
+                .expect("page should be mapped after protection");
             assert_eq!(va_high_pa, mapped_pa_of_va(va_high));
             assert_eq!(
                 prop_high,
@@ -910,7 +910,7 @@ mod protection_and_query {
             let va_to_check = PAGE_SIZE * i;
             let (_, prop) = page_table
                 .page_walk(va_to_check)
-                .expect("Mapping should exist");
+                .expect("mapping should exist");
             assert_eq!(prop.flags, PageFlags::RW);
             assert_eq!(prop.cache, CachePolicy::Writeback);
         }
@@ -926,7 +926,7 @@ mod protection_and_query {
             let va_to_check = PAGE_SIZE * i;
             let (_, prop) = page_table
                 .page_walk(va_to_check)
-                .expect("Mapping should exist");
+                .expect("mapping should exist");
             assert_eq!(prop.flags, PageFlags::R);
             assert_eq!(prop.cache, CachePolicy::Writeback);
         }
diff --git a/ostd/src/mm/test.rs b/ostd/src/mm/test.rs
index 58d9bd007..d9577bcd8 100644
--- a/ostd/src/mm/test.rs
+++ b/ostd/src/mm/test.rs
@@ -763,7 +763,7 @@ mod vmspace {
         // Creates the first cursor.
         let _cursor1 = vmspace
             .cursor(&preempt_guard, &range1)
-            .expect("Failed to create first cursor");
+            .expect("failed to create first cursor");
 
         // Attempts to create the second overlapping cursor.
         let cursor2_result = vmspace.cursor(&preempt_guard, &range2);
@@ -783,7 +783,7 @@ mod vmspace {
                 .expect("failed to create the mutable cursor");
             let prop = PageProperty::new_user(PageFlags::RW, CachePolicy::Writeback);
             cursor_mut.map(frame.clone(), prop);
-            cursor_mut.jump(range.start).expect("Failed to jump cursor");
+            cursor_mut.jump(range.start).expect("failed to jump cursor");
             let protected_range = cursor_mut.protect_next(0x1000, |flags, _cache| {
                 *flags = PageFlags::R;
             });
```

ostd/src/mm/page_table/node/mod.rs
junyang-zh: Sorry. I missed that.

ostd/src/mm/tlb.rs
junyang-zh: Accepting `RcuDrop<>` would lead to some extra complexity since the user may have `RcuDrop<Frame<()>>` and the function accepts `RcuDrop<Frame<dyn AnyFrameMeta>>`. We should write some casting utilities for that.

Another solution is just to use [fallible push](https://doc.rust-lang.org/beta/alloc/vec/struct.Vec.html#method.push_within_capacity). But it requires a nightly feature.

lrh2000: There seem to be two problems:
 1. Whether to accept a `RcuDrop<Frame>` instead of a `Frame`.
 2. How to resolve the panicking problem.

If "1" is not realistic, we can add a comment to indicate that we need to be careful about panicking when modifying the code around.

For "2", we can use a panic guard instead of the fallible push since it's unlikely to panic anyway.

> Another solution is just to use

I think we should address the two problems separately and not mix them together.

ostd/src/mm/page_table/node/entry.rs
```diff
@@ -177,13 +179,25 @@ impl<'a, 'rcu, C: PageTableConfig> Entry<'a, 'rcu, C> {
...
+
             let small_pa = pa + i * page_size::<C>(level - 1);
             let mut entry = pt_lock_guard.entry(i);
             // SAFETY: It's a part of the mapped item, and the ownership is
             // properly transferred to the new sub-entry.
             let small_item = unsafe { C::item_from_raw(small_pa, level - 1, prop) };
             let old = entry.replace(PteState::Mapped(RcuDrop::new(small_item)));
             debug_assert!(old.is_absent());
+
+            let _ = ManuallyDrop::new(panic_guard);
```
lrh2000: I think we can only forget the guard after:
```rust
        unsafe { self.node.write_pte(self.idx, self.pte) };
```

I'm not sure if we're on the same page about the problem. I think the double `Drop` will happen because we will drop
 1. small items and the new page (which contains small items), and
 2. the old page table, which contains the huge item.

Correct me if I'm wrong?

ostd/src/mm/tlb.rs
lrh2000: Honestly, given this is really error-prone, I was even thinking about:
```rust
impl RcuDrop<T> {
    /// Note that panic unwinding may cause the returned value to be dropped
    /// immediately, which is not sound. Therefore, the caller must forget the
    /// [`PanicGuard`] after it ensures that the value will be dropped at the
    /// correct time.
    unsafe fn into_inner(self) -> (T, PanicGuard) { todo!() }
}
```

Then, `TlbFlusher` can accept a frame and a panic guard. `TlbFlusher` will forget the guard at the correct time. 
What do you think of this solution?

(The downside is that we'll need to make `PanicGuard` a formal type, which requires documentation and perhaps more associated methods, e.g., `PanicGuard::new`/`PanicGuard::forget`, for flexibility reasons? But it should worth it?)

ostd/src/mm/vm_space.rs
```diff
@@ -460,11 +452,17 @@ impl<'a> CursorMut<'a> {
...
             PageTableFrag::Mapped { va, item } => {
                 debug_assert_eq!(va, start_va);
+                // SAFETY: If the item is not a scalar (e.g., a frame
+                // pointer), we will drop it after the RCU grace period
+                // (see `issue_tlb_flush_with`).
+                let (item, panic_guard) = unsafe { RcuDrop::into_inner(item) };
+
                 match item.mapped_item {
                     MappedItem::TrackedFrame(old_frame) => {
+                        let drop_after_flush = Frame::rcu_from_unsized(RcuDrop::new(old_frame));
```
lrh2000: Should this be converted to:
```rust
                        let drop_after_flush =  RcuDrop::new(old_frame);
                        panic_guard.forget();
                        self.flusher.issue_tlb_flush_with(
                            TlbFlushOp::for_single(start_va),
                            Frame::rcu_from_unsized(drop_after_flush),
                        );
```

Since we should forget the panic guard once we create a new `RcuDrop`?

ostd/src/mm/vm_space.rs
```diff
@@ -507,27 +507,37 @@ impl<'a> CursorMut<'a> {
...
                         VmItem {
                             mapped_item: MappedItem::TrackedFrame(old_frame),
                             ..
                         } => {
                             num_unmapped += 1;
+
+                            let drop_after_flush = Frame::rcu_from_unsized(RcuDrop::new(old_frame));
                             self.flusher
-                                .issue_tlb_flush_with(TlbFlushOp::for_single(va), old_frame.into());
+                                .issue_tlb_flush_with(TlbFlushOp::for_single(va), drop_after_flush);
```
lrh2000: Same.

ostd/src/mm/tlb.rs
```diff
@@ -322,9 +329,13 @@ impl OpsStack {
         self.num_pages_to_flush == u32::MAX
     }
 
-    fn push(&mut self, op: TlbFlushOp, drop_after_flush: Option<Frame<dyn AnyFrameMeta>>) {
+    fn push(&mut self, op: TlbFlushOp, drop_after_flush: Option<RcuDrop<Frame<dyn AnyFrameMeta>>>) {
         if let Some(frame) = drop_after_flush {
-            self.page_keeper.push(frame);
+            // SAFETY: by pushing into the `frame_keeper`, the frame will be
```
lrh2000:
```suggestion
            // SAFETY: By pushing into the `frame_keeper`, the frame will be
```

ostd/src/mm/vm_space.rs
```diff
@@ -507,27 +507,37 @@ impl<'a> CursorMut<'a> {
...
                         }
                         VmItem {
-                            mapped_item: MappedItem::UntrackedIoMem { .. },
+                            mapped_item: MappedItem::UntrackedIoMem { level, .. },
                             ..
                         } => {
+                            num_unmapped += nr_base_per_page::<UserPtConfig>(level);
```
lrh2000: I don't think we currently count `IoMem` as frames in the RSS?
https://github.com/asterinas/asterinas/blob/a599a2c563b857c8c69ab512d3405c8fa6a12932/kernel/src/vm/vmar/vm_mapping.rs#L197-L213

ostd/src/panic.rs
```diff
@@ -44,6 +46,34 @@ pub fn abort() -> ! {
...
+
+impl Drop for PanicGuard {
+    fn drop(&mut self) {
+        crate::panic::abort();
+    }
+}
+
+impl PanicGuard {
+    /// Creates a panic guard that aborts the system if dropped.
+    pub fn new() -> Self {
```
lrh2000: I think the type and the methods should be `pub(crate)`. It does not seem very useful for the kernel part especially given that the kernel cannot write unsafe code.

ostd/src/panic.rs
```diff
@@ -44,6 +46,34 @@ pub fn abort() -> ! {
...
+        PanicGuard(())
+    }
+
+    /// Finishes panic guarding by forgetting the guard.
+    ///
+    /// After the panic guarding finishes, it no longer aborts the system
+    /// when panic happens.
+    pub fn forget(self) {
+        let _ = ManuallyDrop::new(self);
```
lrh2000: As discussed in https://github.com/asterinas/asterinas/pull/2052#discussion_r2118827714, I think should just use `core::mem::forget` here.

ostd/src/panic.rs
```diff
@@ -44,6 +46,34 @@ pub fn abort() -> ! {
...
+/// A guard that aborts the system if dropped.
+///
+/// This is useful to ensure that certain objects will not be dropped during
+/// panic handling.
+#[derive(Debug, Default)]
+pub struct PanicGuard(());
+
+impl Drop for PanicGuard {
+    fn drop(&mut self) {
+        crate::panic::abort();
```
lrh2000: Add a message here? Something like:
```rust
early_println!("Panicked in `PanicGuard`, aborting the system")
```

Also this `Drop` impl is now in `panic.rs`, so `abort()` can be used directly without the full path `crate::panic`.

ostd/src/panic.rs
```diff
@@ -50,27 +48,28 @@ pub fn abort() -> ! {
 ///
 /// This is useful to ensure that certain objects will not be dropped during
 /// panic handling.
-#[derive(Debug, Default)]
-pub struct PanicGuard(());
+#[derive(Debug)]
+pub(crate) struct PanicGuard(());
```
lrh2000: Changes in `Query the page table item reference` made to `ostd/src/panic.rs` should be moved to commit `Add a panic guard at huge mapping splitting`.

ostd/src/panic.rs
```diff
@@ -44,6 +44,35 @@ pub fn abort() -> ! {
     crate::power::poweroff(crate::power::ExitCode::Failure);
 }
 
+/// A guard that aborts the system if dropped.
+///
+/// This is useful to ensure that certain objects will not be dropped during
+/// panic handling.
+#[derive(Debug)]
+pub(crate) struct PanicGuard(());
```
lrh2000: Given:
```
> grep -R "_private: ()" ostd
ostd/src/arch/loongarch/irq/remapping.rs:    _private: (),
ostd/src/arch/loongarch/irq/remapping.rs:        Self { _private: () }
ostd/src/arch/riscv/irq/remapping.rs:    _private: (),
ostd/src/arch/riscv/irq/remapping.rs:        Self { _private: () }
ostd/src/arch/riscv/serial.rs:    _private: (),
ostd/src/arch/riscv/serial.rs:        Self { _private: () }
ostd/src/arch/x86/kernel/apic/x2apic.rs:    _private: (),
ostd/src/arch/x86/kernel/apic/x2apic.rs:        Some(Self { _private: () })
ostd/src/task/preempt/guard.rs:    _private: (),
ostd/src/task/preempt/guard.rs:        Self { _private: () }
```
And:
```
> grep -R "struct.*(());" ostd
ostd/src/panic.rs:pub(crate) struct PanicGuard(()););
```

Use the `_private: ()` format instead.

junyang-zh:
@cchanging could you please review the code. Thanks!

Issue #2495, opened 2025-10-09
An efficient and flexible way to structure device and sysfs nodes


lrh2000:
# Summary

This proposal aims to create a Rust design that achieves the same *effect*, the same *efficiency*, and the same *flexibility* as Linux's `struct koject`, `struct device`, and `struct {net,input,...}_device` structures.

Note that these concrete structures are meant to serve as examples. The underlying concept is widely used across the Linux kernel implementation. If the proposed design is accepted, it may also be useful in other places, such as code related to `struct file`.

# Motivation

In Linux, [`struct kobject`](https://elixir.bootlin.com/linux/v6.17.1/source/include/linux/kobject.h#L64) contains necessary information for a kernel object to appear as a node in sysfs.

It is not intended for standalone allocation. Instead, a `kobject` is embedded in a larger structure, e.g, [`struct device`](https://elixir.bootlin.com/linux/v6.17.1/source/include/linux/device.h#L582-L583).
```c
struct device {
	struct kobject kobj;
	// Many other fields..
}
```

If a function accepts a `struct device *` argument, it can directly convert it to a `kobject` and invoke methods associated with `kobject`. [For example:](https://elixir.bootlin.com/linux/v6.17.1/source/drivers/base/core.c#L3436)
```c
static int device_add_class_symlinks(struct device *dev)
{
	// ...
	error = sysfs_create_link(&dev->kobj, &sp->subsys.kobj, "subsystem");
	// ...
}
```

But callbacks from the `kobject` accept a `struct kobject *` argument. In Linux, it is possible to use `container_of` to restore the `struct device *` object. [For example:](https://elixir.bootlin.com/linux/v6.17.1/source/drivers/base/core.c#L2417)
```c
static ssize_t dev_attr_show(struct kobject *kobj, struct attribute *attr,
			     char *buf)
{
	// ...
	struct device *dev = kobj_to_dev(kobj);
	// ...
}
```

The device object can be further embedded in more complex structures, e.g., [`struct net_device`](https://elixir.bootlin.com/linux/v6.17.1/source/include/linux/netdevice.h#L2088):
```c
struct net_device {
	// Many other fields...
	struct device		dev;
	// Many other fields...
}
```

or [`struct input_dev`](https://elixir.bootlin.com/linux/v6.17.1/source/include/linux/input.h#L137):
```c
struct input_dev {
	// Many other fields...
	struct device dev;
	// Many other fields...
}
```

[...]

PR #2606, opened 2025-11-23
Remove unnecessary `CharFile` and `OpenCharFile` types

lrh2000:
I don't understand why these two types exist at all. Therefore, I propose eliminating them.
 - `CharFile` is not necessary if we can have `pub trait CharDevice: Device {`.
 - `OpenCharFile` is unnecessary if `CharDevice::open` can return a `Box<dyn FileIo>`.


cqs21:
> * `CharFile` is not necessary if we can have `pub trait CharDevice: Device {`.

We want to extract `char` from the kernel crate into separate component, similar to `block`, so we need `CharFile`.

tatetian:
> I don't understand why these two types exist at all. Therefore, I propose eliminating them.
> 
> * `CharFile` is not necessary if we can have `pub trait CharDevice: Device {`.
> * `OpenCharFile` is unnecessary if `CharDevice::open` can return a `Box<dyn FileIo>`.

I agree. I didn't catch the redundancy on my review. In a hindsight, both types do not provide much values.

> We want to extract `char` from the kernel crate into separate component, similar to `block`, so we need `CharFile`.

We don't have a clear blueprint to do this, yet. The char device abstraction relies on some kernel types that won't be available in an OS component. So this future goal can't really justify having redundant code at the present time.

----

@lrh2000 Since this is a cleanup/refactoring PR, I have to give some additional cleanup suggestions and hope that you can include them in this PR.
* Remove the `DevtmpfsName` type. I proposed `DevtmpfsName` because I thought a device path on devtmpfs is in the form of either `/dev/<device_name>` or `/dev/<class_name>/<device_name>`. But further investigation shows that device drivers may create devices on paths of multiple nested levels. For example, the USB bus creates inodes of path `/dev/bus/usb/XXX/YYY`. So `DevtmpfsName` should be replaced with `String`.
* Remove the `CharDevice` trait. Now that most methods of `CharDevice` have been removed, the only method left is `devtmpfs_name`. This makes the merits of having the single-method `CharDevice` trait questionable. And conceptually, block devices should also have paths on devtmpfs. So it appears that we should move `devtmpfs_name`(or rename it to `devtmpfs_path`/`full_name`) method from `CharDevice` to the `Device` trait.
* Group `kernel/src/device/disk.rs` and `kernel/src/device/char.rs` into a new module called `registry` (`kernel/src/device/registry/`). This new organization reflects the fact these modules are alike, yet quite different from the rest of the modules under `device`.
* Rename `kernel/src/device/disk.rs` to `block.rs` because the code in `block.rs` uses the term "block devices" and never does the term "disk".

lrh2000:
> We want to extract `char` from the kernel crate into separate component, similar to `block`, so we need `CharFile`.

Thanks for your explanation. However, `CharFile` is wrapped in an `Arc`, which internally contains an `Arc<dyn CharDevice>`. There are too many `Arc` objects, and I don't think this is the correct design, even for extraction. I would prefer a different approach, such as #2495.

Since `char` currently lives in `aster-nix`, I'd prefer not to have an imperfect design for some future reasons. Even for the `block` case, I think it's better to write TODO comments like this:
```patch
diff --git a/kernel/src/device/disk.rs b/kernel/src/device/disk.rs
index f361af460..961274f78 100644
--- a/kernel/src/device/disk.rs
+++ b/kernel/src/device/disk.rs
@@ -47,7 +47,9 @@ pub(super) fn init_in_first_process(fs_resolver: &FsResolver) -> Result<()> {
 
 /// Represents a block device inode in the filesystem.
 ///
-/// Only implements the `Device` trait.
+// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Arc` just to implement the `Device`
+// trait. It leads to unnecessary reference counting and heap allocation. We should devise a better
+// strategy to eliminate the unnecessary intermediate `Arc`.
 #[derive(Debug)]
 pub struct BlockFile(Arc<dyn BlockDevice>);
 
@@ -72,9 +74,10 @@ impl Device for BlockFile {
 }
 
 /// Represents an opened block device file ready for I/O operations.
-///
-/// Does not implement the `Device` trait but provides full implementations
-/// for I/O related traits.
+//
+// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Box` just to implement the `FileIo`
+// trait. It leads to unnecessary heap allocation. We should devise a better strategy to eliminate
+// the unnecessary intermediate `Box`.
 pub struct OpenBlockFile(Arc<dyn BlockDevice>);
 
 impl InodeIo for OpenBlockFile {
```

> Since this is a cleanup/refactoring PR, I have to give some additional cleanup suggestions and hope that you can include them in this PR.

In fact, this PR now needs to touch something in evdev. I just have a quick look at the evdev code and unfortunately I think I'll have to clean up some buggy code in evdev first. It has some more severe problems like atomic mode breaks, wrong error codes, (and perhaps some race conditions, but which need to be checked more carefully).

lrh2000:
> Remove the `CharDevice` trait. Now that most methods of `CharDevice` have been removed, the only method left is `devtmpfs_name`. This makes the merits of having the single-method `CharDevice` trait questionable. And conceptually, block devices should also have paths on devtmpfs. So it appears that we should move `devtmpfs_name`(or rename it to `devtmpfs_path`/`full_name`) method from `CharDevice` to the `Device` trait.

I tried this. This statement is mostly correct and I almost succeeded. However, ptmx and pty slave also need to implement the `Device` trait. But I'm afarid that they do not have a devtmpfs path:
 - ptmx:
https://github.com/asterinas/asterinas/blob/44c3ffa4e290a2d47eba767b9c8f400053af8fe3/kernel/src/fs/devpts/ptmx.rs#L159
 - pty slave (it relies on the `Tty` type):
https://github.com/asterinas/asterinas/blob/44c3ffa4e290a2d47eba767b9c8f400053af8fe3/kernel/src/device/tty/mod.rs#L398

They live under `/dev/pts` but it is a separate FS.

---

> I have to give some additional cleanup suggestions and hope that you can include them in this PR.

I think I have included all the other suggestions. Please let me know if I missed something. Thanks for your suggestions!

tatetian (commented):
Great job. The changes are almost perfect.

I just have just one concern: the Git history shows that the original `char.rs` file is removed, but in fact it is just renamed. Could you fix the Git commit?

tatetian:
> However, ptmx and pty slave also need to implement the Device trait. But I'm afarid that they do not have a devtmpfs path:

The problem can be solved by letting the `devtmpfs_path` method to return an `Option<String>` to indicate if this char device should be created an inode on devtmpfs or not.

lrh2000:
> The problem can be solved by letting the `devtmpfs_path` method to return an `Option<String>` to indicate if this char device should be created an inode on devtmpfs or not.

Oh, this is a good idea! After doing that, I found out that we no longer need to repeat the device IDs for TTY:
https://github.com/asterinas/asterinas/blob/2fc94a0d060bc674b576b3ba7789062de0d9f035/kernel/src/device/mod.rs#L82-L86

So the code should look much better. See the [force-push diff](https://github.com/asterinas/asterinas/compare/d9b348e92e5c5907e39db2a66fe76cf9fc0d4fa9..0fb9f70f4f5562f4d24898880ea0b753d3d77e8c).

> I just have just one concern: the Git history shows that the original `char.rs` file is removed, but in fact it is just renamed. Could you fix the Git commit?

Perhaps you're checking the diff of the whole PR? I think you need to check the commit diff, e.g., [Group modules into `device/registry`](https://github.com/asterinas/asterinas/pull/2606/commits/0fb9f70f4f5562f4d24898880ea0b753d3d77e8c). It renames `char.rs` to `registry/char.rs`.

tatetian (approved):

PR #2607, opened 2025-11-24
Fix osdk Docker image build failures

taosue:
This PR fixes two issues in the osdk Docker image build:

1. [**GRUB build failure**](https://github.com/asterinas/asterinas/actions/runs/19570491196/job/56106388289#step:4:4764): The `gnulib` repository at `https://git.savannah.gnu.org/git/gnulib` is unstable. Replaced it with the GitHub mirror for better reliability.

2. **Cargo tools installation failure**: The current Rust toolchain (1.86) is incompatible with the latest versions of `mdbook` and `typos-cli` (which require 1.88+). Pinned both tools to compatible versions:
   - `mdbook@0.4.52` (supports Rust 1.82+)
   - `typos-cli@1.39.0` (supports Rust 1.80+)

### DIFF
--- a/osdk/tools/docker/Dockerfile
+++ b/osdk/tools/docker/Dockerfile
@@ -157,10 +157,10 @@ RUN mkdir -p /github/home \
     && ln -s /root/.rustup /github/home/.rustup
 
 # Install cargo tools
-RUN cargo install \
+RUN cargo install --locked \
     cargo-binutils \
-    mdbook \
-    typos-cli
+    mdbook@0.4.52 \
+    typos-cli@1.39.0
 
 # Install QEMU built from the previous stages
 COPY --from=qemu /usr/local/qemu /usr/local/qemu


osdk/tools/docker/Dockerfile
```diff
@@ -106,6 +106,7 @@ RUN wget -O unifont.pcf.gz https://ftp.gnu.org/gnu/unifont/unifont-15.1.04/unifo
 WORKDIR /root/grub
 
 RUN echo depends bli part_gpt > grub-core/extra_deps.lst \
+    && git config --global url."https://github.com/coreutils/gnulib".insteadOf "https://git.savannah.gnu.org/git/gnulib" \
```
tatetian: >  The gnulib repository at https://git.savannah.gnu.org/git/gnulib is unstable.

I don't understand the argument here. First of all, the build process used to work, why "unstable" makes it stop working? If it no longer works, what about the upstream repo? Does it mean that our forked repo at `https://github.com/asterinas/grub.git` should be synchronized with the upstream version? I assume the upstream version should work.

taosue: Sorry, my description was not clear. When publishing the OSDK image, it is necessary to pull `gnulib`, but the link being pulled seems to have an unstable network, causing [CI errors](https://github.com/asterinas/asterinas/actions/runs/19570491196/job/56106388289#step:4:4764). The GitHub link should have a more stable network connection, so a redirect was implemented. The code pulled from both links is same.

tatetian: Ok. Add a comment here to explain why replacing the URL.

osdk/tools/docker/Dockerfile
```diff
@@ -105,7 +105,10 @@ RUN wget -O unifont.pcf.gz https://ftp.gnu.org/gnu/unifont/unifont-15.1.04/unifo
     && rm unifont.pcf.gz
 WORKDIR /root/grub
 
+# Use GitHub mirror for `gnulib` to improve connection stability because `git.savannah.gnu.org`
+# repository may have unstable connections.
```
tatetian, line 109: Attaching the comment to `RUN echo depends ...` is confusing because the comment only targets the `git config` line. Extract the `git config ` line as a separate command.

```bash
WORKDIR /root/grub

# Use GitHub mirror for `gnulib` to improve connection stability because `git.savannah.gnu.org`
# repository may have unstable connections.
git config --global url."https://github.com/coreutils/gnulib".insteadOf "https://git.savannah.gnu.org/git/gnulib"

RUN echo depends bli part_gpt > grub-core/extra_deps.lst \
```

PR #2609, opened 2025-11-24
Make the mount-related fields in statx effective

cchanging:
When debugging systemd in nixos, I found that after entering the shell, the `/run/current-system` symlink we prepared in the initialization script could no longer be found, which prevented the `$PATH` environment variable from correctly locating `/run/current-system/sw/bin`. The underlying reason is that systemd automatically mounts a tmpfs at `/run`, overwriting the original contents. In fact, before mounting, systemd will use the `statx` syscall to check if the current path is already a mount root. If it is, it skips the corresponding operation.

In our current `statx` implementation, the `stx_attributes_mask` includes `STATX_ATTR_MOUNT_ROOT`, which implies that the current `stx_attributes` supports the `STATX_ATTR_MOUNT_ROOT` flag. But in reality, we don't actually support this feature, which caused systemd to mistakenly believe that the current path was not a mount root. This PR fixes that issue.

### DIFF
--- a/kernel/src/syscall/statx.rs
+++ b/kernel/src/syscall/statx.rs
@@ -4,7 +4,7 @@ use core::time::Duration;
 
 use super::SyscallReturn;
 use crate::{
-    fs::{file_table::FileDesc, fs_resolver::FsPath, utils::Metadata},
+    fs::{file_table::FileDesc, fs_resolver::FsPath, path::Path},
     prelude::*,
     syscall::constants::MAX_FILENAME_LEN,
 };
@@ -60,7 +60,7 @@ pub fn sys_statx(
         }
     };
 
-    let statx = Statx::from(path.metadata());
+    let statx = Statx::new(&path);
 
     user_space.write_val(statx_buf_ptr, &statx)?;
     Ok(SyscallReturn::Return(0))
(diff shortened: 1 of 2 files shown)


kernel/src/syscall/statx.rs
```diff
@@ -164,7 +169,7 @@ impl From<Metadata> for Statx {
             stx_rdev_minor,
             stx_dev_major,
             stx_dev_minor,
-            stx_mnt_id: 0,
+            stx_mnt_id: path.mount_node().id() as u64,
```
TinaZhangZW, line 163: Before updating the stx_mnt_id field, according to statx system call definition (https://www.man7.org/linux/man-pages/man2/statx.2.html), we need to check whether the STATX_MNT_ID bit is set in stx_mask to make sure that this field is required by caller.

TinaZhangZW (commented):
Thanks for the fix. I've tested this PR and confirmed it resolves the issue where the PATH environment variable loses its settings after systemd starts.

kernel/src/syscall/statx.rs
cchanging: > It should be noted that the kernel may return fields that weren't requested and may fail to return fields that were requested (from manual page).

Since obtaining the mount ID is straightforward, I now provide it by default along with other existing fields to avoid additional branch judgments. Following your suggestion, I have added `STATX_MNT_ID` to the returned mask, allowing users to be aware that we provide this information.

PR #2611, opened 2025-11-25
Fix the behavior of syncing `BlockGroup` metadata in ext2

cchanging:
During debugging NixOS in Asterinas, when creating a file in the ext2 filesystem and then shutting down the system, the file cannot be read correctly upon reboot. The root cause is an issue with the sync operation of ext2's `BlockGroup`. The current implementation directly writes the `IdAlloc` as bytes, but the length of the slice after `as_bytes` does not meet the alignment requirement with `SECTOR_SIZE` specified in `write_bytes_async`, preventing the write operation from proceeding. 

This PR fixes the problem by ensuring the size of `IdAlloc` is aligned with `SECTOR_SIZE` when loading.

### DIFF
--- a/ostd/libs/id-alloc/src/lib.rs
+++ b/ostd/libs/id-alloc/src/lib.rs
@@ -26,30 +26,6 @@ impl IdAlloc {
         }
     }
 
-    /// Constructs a new id allocator from a slice of `u8` bytes and a maximum capacity.
-    ///
-    /// The slice of `u8` bytes is the raw data of a bitmap.
-    pub fn from_bytes_with_capacity(slice: &[u8], capacity: usize) -> Self {
-        let bitset = if capacity > slice.len() * 8 {
-            let mut bitset = BitVec::from_slice(slice);
-            bitset.resize(capacity, false);
-            bitset
-        } else {
-            let mut bitset = BitVec::from_slice(&slice[..capacity.div_ceil(8)]);
-            bitset.truncate(capacity);
-            bitset
-        };
-
-        let first_available_id = (0..bitset.len())
-            .find(|&i| !bitset[i])
-            .map_or(bitset.len(), |i| i);
-
-        Self {
-            bitset,
-            first_available_id,
-        }
-    }
-
     /// Allocates and returns a new `id`.
     ///
     /// If allocation is not possible, it returns `None`.
@@ -176,11 +152,6 @@ impl IdAlloc {
     /// If the `id` is out of bounds, this method will panic.
     pub fn is_allocated(&self, id: usize) -> bool {
         self.bitset[id]
-    }
-
-    /// Views the id allocator as a slice of `u8` bytes.
-    pub fn as_bytes(&self) -> &[u8] {
-        self.bitset.as_raw_slice()
     }
 }
 


kernel/src/fs/ext2/block_group.rs
```diff
@@ -55,7 +56,10 @@ impl BlockGroup {
                     }
                     let mut buf = vec![0u8; BLOCK_SIZE];
                     block_device.read_bytes(bid as usize * BLOCK_SIZE, &mut buf)?;
-                    Ok(IdAlloc::from_bytes_with_capacity(&buf, capacity))
+                    Ok(IdAlloc::from_bytes_with_capacity(
+                        &buf,
+                        capacity.align_up(SECTOR_SIZE),
```
tatetian: This is incorrect. The units of `capacity` and `SECTOR_SIZE` do not match: `capacity` is in bits, `SECTOR_SIZE` is in bytes.

kernel/src/fs/ext2/block_group.rs
```diff
@@ -55,7 +56,10 @@ impl BlockGroup {
                     }
                     let mut buf = vec![0u8; BLOCK_SIZE];
                     block_device.read_bytes(bid as usize * BLOCK_SIZE, &mut buf)?;
-                    Ok(IdAlloc::from_bytes_with_capacity(&buf, capacity))
+                    Ok(IdAlloc::from_bytes_with_capacity(
+                        &buf,
+                        capacity.align_up(SECTOR_SIZE * 8),
+                    ))
```
cqs21: The `capacity` comes from`super_block.blocks_per_group()` or `super_block.inodes_per_group()`. After the change, this would cause allocation of invalid blocks and inodes.

tatetian: Precisely 👍

kernel/src/fs/ext2/block_group.rs
```diff
@@ -49,16 +49,28 @@ impl BlockGroup {
                     GroupDescriptor::from(raw_descriptor)
                 };
 
-                let get_bitmap = |bid: Ext2Bid, capacity: usize| -> Result<IdAlloc> {
-                    if capacity > BLOCK_SIZE * 8 {
-                        return_errno_with_message!(Errno::EINVAL, "bad bitmap");
-                    }
-                    let mut buf = vec![0u8; BLOCK_SIZE];
-                    block_device.read_bytes(bid as usize * BLOCK_SIZE, &mut buf)?;
-                    Ok(IdAlloc::from_bytes_with_capacity(&buf, capacity))
```
tatetian, line 58: Here is a simpler, more efficient, and more elegant solution to this problem.

First off, a piece of background knowledge about Ext2:

> Two blocks near the start of each (block) group are reserved for the block usage bitmap and the inode usage bitmap which show which blocks and inodes are in use. Since each bitmap is limited to a single block, this means that the maximum size of a block group is 8 times the size of a block.

(From https://www.nongnu.org/ext2-doc/ext2.html#block-bitmap)

This means that usually, the capacity of a bitmap is `BLOCK_SIZE * 8`; but the last block group may have fewer available blocks. For the purpose of unification and convenience, we should use a buffer of `BLOCK_SIZE` bytes for each bitmap--regardless of whether it belongs to the last block group or not. This way, the failure of writing a bitmap back with `write_bytes_async` can be trivially fixed.

But what about `IdAlloc::capacity` for the last, partial block group? `IdAlloc` is backed by `BitVec` internally and lucky for us, `BitVec` allows us to back the bitmap with an underlying buffer much larger than the user-visible capacity.

So we only need to change the closure to get `bitmap` to the form below:

```rust
                let get_bitmap = |bid: Ext2Bid, capacity: usize| -> Result<IdAlloc> {
                    // A bitmap is stored as one block on disk.
                    debug_assert!(capacity <= BLOCK_SIZE * 8);

                    let mut bitmap = IdAlloc::with_capacity(BLOCK_SIZE * 8);
                    let bitmap_buf = bitmap.as_bytes_mut();
                    block_device.read_bytes(bid as usize * BLOCK_SIZE, &mut bitmap_buf)?;

                    if capacity < BLOCK_SIZE * 8 {
                        // Set only the first `capacity` bits as valid.
                        bitmap.resize(capacity);

                        // The resize operation guarantees that the size of the underlying buffer is still `BLOCK_SIZE`.
                        // This is important as we will reuse this buffer when flushing the bitmap to disk.
                        debug_assert!(bitmap.as_bytes().len() == BLOCK_SIZE * 8);
                    }

                    bitmap
                };
```

In `IdAlloc`, add the following Rust doc to `as_bytes`:

```rust
    /// Views the ID allocator as a slice of `u8` bytes.
    ///
    /// The IDs are kept in a bitmap,
    /// where bit 0 means a free ID and bit 1 an allocated ID.
    /// Bits are stored in the least-significant-first order.
    pub fn as_bytes(&self) -> &[u8] {
        self.bitset.as_raw_slice()
    }
```

The above guarantee should have been upheld by the current implementation.

Also, a new method called `as_bytes_mut` needs to be added.

cchanging:
> But what about IdAlloc::capacity for the last, partial block group? IdAlloc is backed by BitVec internally and lucky for us, BitVec allows us to back the bitmap with an underlying buffer much larger than the user-visible capacity.

Thank you for your suggestion! I had actually considered using `resize` to achieve this goal before, but `BitVec`'s `as_raw_slice` returns a slice whose length is based on its `len` property rather than its `capacity` property. That means after resizing, `BitVec` cannot provide bytes beyond `len` through `as_raw_slice` without `unsafe`. I think this might be because `BitVec` cannot guarantee that the data in the capacity is initialized. 

I currently have a similar idea, and the general implementation looks like this—by using additional parameters to impose limits for `IdAlloc`:

```rust
/// An id allocator implemented by the bitmap.
/// The true bit implies that the id is allocated, and vice versa.
#[derive(Clone)]
pub struct IdAlloc {
    bitset: BitVec<u8>,
    first_available_id: usize,
    /// The maximum number of ids that can be allocated.
    limit: usize,
}

impl IdAlloc {
    fn view(&self) -> &BitSlice<u8> {
        &self.bitset[..self.limit]
    }

    fn view_mut(&mut self) -> &mut BitSlice<u8> {
        &mut self.bitset[..self.limit]
    }

    /// Constructs a new id allocator with a maximum capacity.
    pub fn with_capacity(capacity: usize) -> Self {
        let mut bitset = BitVec::with_capacity(capacity);
        bitset.resize(capacity, false);
        Self {
            bitset,
            first_available_id: 0,
            limit: capacity,
        }
    }

    /// Constructs a new id allocator from a slice of `u8` bytes and a maximum capacity.
    ///
    /// The slice of `u8` bytes is the raw data of a bitmap.
    pub fn from_bytes_with_capacity(slice: &[u8], capacity: usize) -> Self {
        let mut bitset = BitVec::from_slice(slice);
        if capacity > slice.len() * 8 {
            bitset.resize(capacity, false);
        }

        let first_available_id = (0..bitset.len())
            .find(|&i| !bitset[i])
            .map_or(bitset.len(), |i| i);

        Self {
            bitset,
            first_available_id,
            limit: capacity,
        }        
    }

    /// Views the id allocator as a slice of `u8` bytes.
    /// 
    /// This will view the entire underlying bitset, including the bits beyond the limit.
    pub fn as_bytes(&self) -> &[u8] {
        self.bitset.as_raw_slice()
    }
}
```

@tatetian What do you think? This should work, but I'm not sure if it might seem a bit semantically odd.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -5,14 +5,16 @@
...
+use bitvec::{prelude::BitVec, slice::BitSlice};
 
 /// An id allocator implemented by the bitmap.
 /// The true bit implies that the id is allocated, and vice versa.
 #[derive(Clone)]
 pub struct IdAlloc {
     bitset: BitVec<u8>,
     first_available_id: usize,
+    /// The maximum number of ids that can be allocated.
+    limit: usize,
```
tatetian: These changes to `IdAlloc` significantly increase the complexity of this type and the burden of maintenance. In addition, almost all other users of `IdAlloc` use it as an in-memory data structure; only Ext2 needs to persist it on the disk and calls `alloc_consecutive` to allocate IDs in batch.

In light of the observations above, I suggest adding an `IdAlloc` equivalence for Ext2, `IdBitmap`. Having two separate data structures can better satisfy the needs of two different kinds of users. The draft code of `IdBitmap` is shown below:

```rust

/// A disk I/O-friendly bitmap for ID management (e.g., block or inode IDs).
///
/// An ID bitmap has the same size as a block, i.e., `BLOCK_SIZE`.
/// Each bit in the bitmap represents one ID:
/// bit 0 at the i-th position of the bitmap means the i-th ID is free
/// and bit 1 means the ID is in use.
/// As such, the bitmap can contain contain at most `BLOCK_SIZE` * 8 bits/IDs.
pub struct IdBitmap {
    buf: Box<[u8]>,
    first_available_id: u16,
    len: u16,
}

impl IdBitmap {
    /// Creates a new ID bitmap with all IDs free to use.
    ///
    /// The length of the returned ID bitmap is equal to `IdBitmap::max_len()`.
    pub fn new() -> Self { ... }

    /// Creates a new ID bitmap out of a given buffer, whose first `len`-bits represent valid IDs.
    ///
    /// # Panics
    ///
    /// This method panics if `len` is greater than `IdBitmap::max_len()`.
    pub fn from_buf(buf: Box<[u8]>, len: u16) -> Self { ... }

    /// The length of the ID bitmap, i.e., the maximum number of IDs.
    pub const fn len(&self) -> u16 { ... }

    /// The maximum length of an ID bitmap.
    pub const fn max_len() -> u16 {
        BLOCK_SIZE * 8
    }

    /// Returns a reference to the underlying buffer of `BLOCK_SIZE` bytes.
    pub fn as_bytes(&self) -> &[u8] { ... }

    // Create a `bitvec::slice::BitSlice` out of the `u8` slice to manipulate the bits easily...
    fn bit_slice(&self) -> bitvec::slice::BitSlice<u8, _> { ... }

    // The following methods have the same semantics as the old ones.

    pub fn alloc(&mut self) -> Option<u16> {
        // Implement with `bit_slice`...
    }

    pub fn alloc_consecutive(&mut self, count: u16) -> Option<Range<u16>> {
        // Implement with `bit_slice`...
    }

    pub fn free(&mut self) -> u16 {
        // Implement with `bit_slice`...
    }

    pub fn free_consecutive(&mut self, range: Range<u16>) -> Option<Range<u16>> {
        // Implement with `bit_slice`...
    }
}
```

kernel/src/fs/utils/id_bitmap.rs
```diff
@@ -0,0 +1,184 @@
...
+            if curr_range.len() < count as usize {
+                return None;
+            }
+
+            curr_range
+        };
+
+        // Set every bit to 1 within the allocated range
+        for id in allocated_range.clone() {
+            self.bit_slice_mut().set(id as usize, true);
```
tatetian: Do not call `bit_slice_mut` in a loop. There could be potentially an overhead associated with the conversion. Review all methods for this flaw.

kernel/src/fs/utils/id_bitmap.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /// Returns the length of the ID bitmap, i.e., the maximum number of IDs.
+    #[expect(unused)]
+    pub const fn len(&self) -> u16 {
+        self.len
+    }
+
+    /// Returns the maximum length of an ID bitmap.
+    pub const fn max_len() -> u16 {
```
tatetian: On a second, calling this method `capacity` would be even more appropriate. Change all occurrences of `max_len`.

tatetian (commented):
Good job! The new revision looks almost perfect.

PR #2612, opened 2025-11-25
Add directory search permission check during path traversal on fs_res…

zzj-5341:
This PR introduces directory execute/search permission (MAY_EXEC) checking during path lookup, aligning Asterinas with the Linux VFS behavior. 
Added `FsResolver::check_dir_search_permission()`
Inserted per-directory `MAY_EXEC` check in `lookup_from_parent()` 
Path traversal now fails with `EACCES` when execute/search permission is missing

### DIFF
--- a/kernel/src/fs/fs_resolver.rs
+++ b/kernel/src/fs/fs_resolver.rs
@@ -2,6 +2,8 @@
 
 use alloc::str;
 
+use crate::fs::utils::Permission;
+
 use ostd::task::Task;
 
 use super::{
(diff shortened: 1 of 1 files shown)


kernel/src/fs/fs_resolver.rs
```diff
@@ -176,6 +178,13 @@ impl FsResolver {
 
         Ok(lookup_res)
     }
+        fn check_dir_search_permission(path: &Path) -> Result<()> {
+        let inode = path.inode();
+        if inode.type_() == InodeType::Dir {
+            inode.check_permission(Permission::MAY_EXEC)?;
```
tatetian, line 184: This check is already included in the `Path::lookup` method:

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/fs/path/mod.rs#L73-L82

PR #2613, opened 2025-11-25
Fix page table cursor invariants by exposing level details

junyang-zh:
This PR is based on #2595 and #2605 . Please review them first.

The problem is that for page table cursor's operations (`map`, `take_next`, and `protect_next`), they traverse the cursor forward to the next virtual address when the operations are finished. Assuming a cursor holding a range `0x1000..0x2000`, before `map` the cursor is at virtual address `0x1000`, and after mapping a page, the cursor is at `0x2000`. So the cursor can point to the end of the range.

But we want to maintain an invariant that the current virtual address must be within the range of each node in the path, so that every operation (like `query` or `map`) always succeeds. After the fix, the return values are drastically simplified.

The way to fix the problem is to expose level details to the user, and define `map`/`unmap`/`protect` to operate on the current PTE only, without a compulsory `find_next`.

---

This PR is also based on https://github.com/asterinas/asterinas/pull/2744 though not relevant.

### DIFF
--- a/kernel/core/src/vm/vmar/vm_mapping.rs
+++ b/kernel/core/src/vm/vmar/vm_mapping.rs
@@ -20,8 +20,8 @@ use ostd::{
 };
 
 use super::{
-    PageFaultInfo, Rmap, RssType, Vmar, interval_set::Interval, util::is_intersected,
-    vmar_impls::RssDelta,
+    PageFaultInfo, Rmap, RssType, Vmar, cursor::CursorMutExt, interval_set::Interval,
+    util::is_intersected, vmar_impls::RssDelta,
 };
 use crate::{
     fs::{
@@ -420,7 +420,7 @@ impl VmMapping {
                     &preempt_guard,
                     &(page_aligned_addr..page_aligned_addr + PAGE_SIZE),
                 )?;
-                if let (_, Some(_)) = cursor.query().unwrap() {
+                if cursor.query().is_some() {
                     return Ok(());
                 }
             }
@@ -473,21 +473,20 @@ impl VmMapping {
         required_perms: VmPerms,
         rss_delta: &mut RssDelta,
     ) -> Result<()> {
+        let is_write = required_perms.contains(VmPerms::WRITE);
+        let va_range = page_aligned_addr..page_aligned_addr + PAGE_SIZE;
+
         'retry: loop {
             let preempt_guard = disable_preempt();
-            let mut cursor = vm_space.cursor_mut(
-                &preempt_guard,
-                &(page_aligned_addr..page_aligned_addr + PAGE_SIZE),
-            )?;
+            let mut cursor = vm_space.cursor_mut(&preempt_guard, &va_range)?;
 
-            let (va, item) = cursor.query().unwrap();
-            let is_write = required_perms.contains(VmPerms::WRITE);
+            let item = cursor.query();
             match item {
                 Some(VmQueriedItem::MappedRam { frame, mut prop }) => {
                     if VmPerms::from(prop.flags).contains(required_perms) {
                         // The page fault is already handled maybe by other threads.
                         // Just flush the TLB and return.
-                        TlbFlushOp::for_range(va).perform_on_current();
+                        TlbFlushOp::for_range(va_range).perform_on_current();
                         return Ok(());
                     }
                     assert!(is_write);
@@ -515,23 +514,24 @@ impl VmMapping {
                     let new_flags = PageFlags::W | PageFlags::ACCESSED | PageFlags::DIRTY;
 
                     if self.is_shared || only_reference {
-                        cursor.protect_next(va.end, |flags, _cache| {
+                        cursor.protect(|flags, _cache| {
                             *flags |= new_flags;
                         });
-                        cursor.flusher().issue_tlb_flush(TlbFlushOp::for_range(va));
-                        cursor.flusher().dispatch_tlb_flush();
+                        cursor
+                            .flusher()
+                            .issue_tlb_flush(TlbFlushOp::for_range(va_range));
                     } else {
                         // Perform COW if it is a write access to a private mapping.
                         let new_frame = duplicate_frame(&frame)?;
                         prop.flags |= new_flags;
-                        cursor.unmap(va.end);
-                        cursor.jump(va.start).unwrap();
+                        cursor.unmap();
                         cursor.map(new_frame.into(), prop);
                         // FIXME: Linux re-classifies the page from `File` to `Anon` in RSS,
                         // when a COW on a file-backed mapping happens.
                         // We currently do not support this re-classification,
                         // since it will introduce some complexity when unmapping this page.
                     }
+                    cursor.flusher().dispatch_tlb_flush();
                     cursor.flusher().sync_tlb_flush();
                 }
                 Some(VmQueriedItem::MappedIoMem { .. }) => {
@@ -680,29 +680,30 @@ impl VmMapping {
             let mut cursor = vm_space.cursor_mut(&preempt_guard, &(start_addr..end_addr))?;
 
             let rss_delta_ref = &mut rss_delta;
-            let operate =
-                move |commit_fn: &mut dyn FnMut() -> Result<(usize, CachePage), VmoCommitError>| {
-                    if let (_, None) = cursor.query().unwrap() {
-                        // We regard all the surrounding pages as accessed, no matter
-                        // if it is really so. Then the hardware won't bother to update
-                        // the accessed bit of the page table on following accesses.
-                        let page_flags = PageFlags::from(vm_perms) | PageFlags::ACCESSED;
-                        let page_prop = PageProperty::new_user(page_flags, CachePolicy::Writeback);
-                        let (_, frame) = commit_fn()?;
-                        cursor.map(frame.into(), page_prop);
-                        rss_delta_ref.add(self.rss_type(), 1);
-                    } else {
-                        let next_addr = cursor.virt_addr() + PAGE_SIZE;
-                        if next_addr < end_addr {
-                            let _ = cursor.jump(next_addr);
-                        }
-                    }
-                    Ok(())
-                };
+
+            let mut cur_va = start_addr;
+            let operate_fn = move |commit_fn: &mut dyn FnMut() -> core::result::Result<
+                (usize, CachePage),
+                VmoCommitError,
+            >| {
+                cursor.jump(cur_va).unwrap();
+                if cursor.query().is_none() {
+                    // We regard all the surrounding pages as accessed, no matter
+                    // if it is really so. Then the hardware won't bother to update
+                    // the accessed bit of the page table on following accesses.
+                    let page_flags = PageFlags::from(vm_perms) | PageFlags::ACCESSED;
+                    let page_prop = PageProperty::new_user(page_flags, CachePolicy::Writeback);
+                    let (_, frame) = commit_fn()?;
+                    cursor.map(frame.into(), page_prop);
+                    rss_delta_ref.add(self.rss_type(), 1);
+                }
+                cur_va += PAGE_SIZE;
+                Ok(())
+            };
 
             let start_offset = start_addr - self.map_to_addr;
             let end_offset = end_addr - self.map_to_addr;
-            match vmo.try_operate_on_range(&(start_offset..end_offset), operate, mode) {
+            match vmo.try_operate_on_range(&(start_offset..end_offset), operate_fn, mode) {
                 Ok(_) => return Ok(()),
                 Err(err) => {
                     let index = err.pending_index()?;
@@ -840,8 +841,13 @@ impl VmMapping {
         let preempt_guard = disable_preempt();
         let range = self.range();
         let mut cursor = vm_space.cursor_mut(&preempt_guard, &range).unwrap();
+        let mut num_unmapped = 0;
 
-        let num_unmapped = cursor.unmap(range.end);
+        while cursor.find_next_unmappable_subtree(range.end).is_some() {
+            cursor.split_if_map_exceeds_range(&range);
+            num_unmapped += cursor.unmap();
+        }
+
         cursor.flusher().dispatch_tlb_flush();
         cursor.flusher().sync_tlb_flush();
 
@@ -862,13 +868,19 @@ impl VmMapping {
         let mut cursor = vm_space.cursor_mut(&preempt_guard, &range).unwrap();
 
         let op = |flags: &mut PageFlags, _cache: &mut CachePolicy| *flags = new_flags;
-        while cursor.virt_addr() < range.end {
-            if let Some(va) = cursor.protect_next(range.end, op) {
-                cursor.flusher().issue_tlb_flush(TlbFlushOp::for_range(va));
-            } else {
+
+        while cursor.find_next(range.end).is_some() {
+            cursor.split_if_map_exceeds_range(&range);
+            cursor.protect(op);
+            let va = cursor.cur_va_range();
+            cursor
+                .flusher()
+                .issue_tlb_flush(TlbFlushOp::for_range(va.clone()));
+            if cursor.jump(va.end).is_err() {
                 break;
             }
         }
+
         cursor.flusher().dispatch_tlb_flush();
         cursor.flusher().sync_tlb_flush();
 
--- a/kernel/core/src/vm/vmar/vm_mapping.rs
+++ b/kernel/core/src/vm/vmar/vm_mapping.rs
@@ -515,7 +515,7 @@ impl VmMapping {
                     let new_flags = PageFlags::W | PageFlags::ACCESSED | PageFlags::DIRTY;
 
                     if self.is_shared || only_reference {
-                        cursor.protect_next(PAGE_SIZE, |flags, _cache| {
+                        cursor.protect_next(va.end, |flags, _cache| {
                             *flags |= new_flags;
                         });
                         cursor.flusher().issue_tlb_flush(TlbFlushOp::for_range(va));
@@ -524,7 +524,7 @@ impl VmMapping {
                         // Perform COW if it is a write access to a private mapping.
                         let new_frame = duplicate_frame(&frame)?;
                         prop.flags |= new_flags;
-                        cursor.unmap(PAGE_SIZE);
+                        cursor.unmap(va.end);
                         cursor.jump(va.start).unwrap();
                         cursor.map(new_frame.into(), prop);
                         // FIXME: Linux re-classifies the page from `File` to `Anon` in RSS,
@@ -841,7 +841,7 @@ impl VmMapping {
         let range = self.range();
         let mut cursor = vm_space.cursor_mut(&preempt_guard, &range).unwrap();
 
-        let num_unmapped = cursor.unmap(range.len());
+        let num_unmapped = cursor.unmap(range.end);
         cursor.flusher().dispatch_tlb_flush();
         cursor.flusher().sync_tlb_flush();
 
@@ -863,7 +863,7 @@ impl VmMapping {
 
         let op = |flags: &mut PageFlags, _cache: &mut CachePolicy| *flags = new_flags;
         while cursor.virt_addr() < range.end {
-            if let Some(va) = cursor.protect_next(range.end - cursor.virt_addr(), op) {
+            if let Some(va) = cursor.protect_next(range.end, op) {
                 cursor.flusher().issue_tlb_flush(TlbFlushOp::for_range(va));
             } else {
                 break;
--- a/kernel/core/src/vm/vmar/vm_mapping.rs
+++ b/kernel/core/src/vm/vmar/vm_mapping.rs
@@ -20,8 +20,11 @@ use ostd::{
 };
 
 use super::{
-    PageFaultInfo, Rmap, RssType, Vmar, cursor::CursorMutExt, interval_set::Interval,
-    util::is_intersected, vmar_impls::RssDelta,
+    PageFaultInfo, Rmap, RssType, Vmar,
+    cursor::{CursorExt, CursorMutExt},
+    interval_set::Interval,
+    util::is_intersected,
+    vmar_impls::RssDelta,
 };
 use crate::{
     fs::{
@@ -420,7 +423,7 @@ impl VmMapping {
                     &preempt_guard,
                     &(page_aligned_addr..page_aligned_addr + PAGE_SIZE),
                 )?;
-                if cursor.query().is_some() {
+                if cursor.to_leaf().query().is_some() {
                     return Ok(());
                 }
             }
@@ -480,9 +483,9 @@ impl VmMapping {
             let preempt_guard = disable_preempt();
             let mut cursor = vm_space.cursor_mut(&preempt_guard, &va_range)?;
 
-            let item = cursor.query();
+            let item = cursor.to_leaf().query();
             match item {
-                Some(VmQueriedItem::MappedRam { frame, mut prop }) => {
+                VmQueriedItem::MappedRam { frame, mut prop } => {
                     if VmPerms::from(prop.flags).contains(required_perms) {
                         // The page fault is already handled maybe by other threads.
                         // Just flush the TLB and return.
@@ -534,7 +537,7 @@ impl VmMapping {
                     cursor.flusher().dispatch_tlb_flush();
                     cursor.flusher().sync_tlb_flush();
                 }
-                Some(VmQueriedItem::MappedIoMem { .. }) => {
+                VmQueriedItem::MappedIoMem { .. } => {
                     // The page of I/O memory is populated when the memory
                     // mapping is created.
                     return_errno_with_message!(
@@ -542,7 +545,7 @@ impl VmMapping {
                         "device memory page faults cannot be resolved"
                     );
                 }
-                None => {
+                VmQueriedItem::None => {
                     // Map a new frame to the page fault address.
                     let (frame, is_readonly) = match self.prepare_page(page_aligned_addr, is_write)
                     {
@@ -576,6 +579,9 @@ impl VmMapping {
 
                     cursor.map(frame, map_prop);
                     rss_delta.add(self.rss_type(), 1);
+                }
+                VmQueriedItem::PageTable => {
+                    unreachable!("pushed but still queried a page table")
                 }
             }
             break 'retry;
@@ -687,13 +693,14 @@ impl VmMapping {
                 VmoCommitError,
             >| {
                 cursor.jump(cur_va).unwrap();
-                if cursor.query().is_none() {
+                if cursor.to_leaf().query().is_none() {
                     // We regard all the surrounding pages as accessed, no matter
                     // if it is really so. Then the hardware won't bother to update
                     // the accessed bit of the page table on following accesses.
                     let page_flags = PageFlags::from(vm_perms) | PageFlags::ACCESSED;
                     let page_prop = PageProperty::new_user(page_flags, CachePolicy::Writeback);
                     let (_, frame) = commit_fn()?;
+                    cursor.adjust_level(1);
                     cursor.map(frame.into(), page_prop);
                     rss_delta_ref.add(self.rss_type(), 1);
                 }
--- a/kernel/src/vm/vmar/vmar_impls/remap.rs
+++ b/kernel/src/vm/vmar/vmar_impls/remap.rs
@@ -196,7 +196,7 @@ impl Vmar {
         let mut current_offset = 0;
         while current_offset < old_size {
             cursor.jump(old_range.start + current_offset).unwrap();
-            let Some(mapped_va) = cursor.find_next(old_size - current_offset) else {
+            let Some(mapped_va) = cursor.find_next(old_range.end) else {
                 break;
             };
             let (va, Some(item)) = cursor.query().unwrap() else {
@@ -211,13 +211,13 @@ impl Vmar {
                 VmQueriedItem::MappedRam { frame, prop } => {
                     let frame = (*frame).clone();
 
-                    cursor.unmap(PAGE_SIZE);
+                    cursor.unmap(mapped_va + PAGE_SIZE);
                     cursor.jump(new_map_va).unwrap();
 
                     cursor.map(frame, prop);
                 }
                 VmQueriedItem::MappedIoMem { paddr, prop } => {
-                    cursor.unmap(PAGE_SIZE);
+                    cursor.unmap(mapped_va + PAGE_SIZE);
                     cursor.jump(new_map_va).unwrap();
 
                     // For MMIO pages, find the corresponding `IoMem` and map it
--- a/kernel/src/vm/vmar/vmar_impls/remap.rs
+++ b/kernel/src/vm/vmar/vmar_impls/remap.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use ostd::{mm::vm_space::VmQueriedItem, task::disable_preempt};
+use ostd::{
+    mm::{page_size_at, vm_space::VmQueriedItem},
+    task::disable_preempt,
+};
 
 use super::{RssDelta, Vmar, util::is_intersected};
 use crate::{prelude::*, vm::vmar::is_userspace_vaddr_range};
@@ -199,30 +202,32 @@ impl Vmar {
             let Some(mapped_va) = cursor.find_next(old_range.end) else {
                 break;
             };
-            let Some(item) = cursor.query() else {
-                panic!("Found mapped page but query failed");
-            };
 
             let offset = mapped_va - old_range.start;
             let new_map_va = new_range.start + offset;
 
-            match item {
+            match cursor.query() {
                 VmQueriedItem::MappedRam { frame, prop } => {
                     let frame = (*frame).clone();
+                    let level = frame.map_level();
 
                     cursor.unmap();
                     cursor.jump(new_map_va).unwrap();
+                    cursor.adjust_level(level);
 
                     cursor.map(frame, prop);
                 }
-                VmQueriedItem::MappedIoMem { paddr, prop } => {
+                VmQueriedItem::MappedIoMem { paddr, prop, level } => {
                     cursor.unmap();
                     cursor.jump(new_map_va).unwrap();
 
                     // For MMIO pages, find the corresponding `IoMem` and map it
                     // at the new location
                     let (iomem, offset) = cursor.find_iomem_by_paddr(paddr).unwrap();
-                    cursor.map_iomem(iomem, prop, PAGE_SIZE, offset);
+                    cursor.map_iomem(iomem, prop, page_size_at(level), offset);
+                }
+                _ => {
+                    unreachable!("mapped item found but query failed")
                 }
             }
 
--- a/kernel/src/vm/vmar/vmar_impls/remap.rs
+++ b/kernel/src/vm/vmar/vmar_impls/remap.rs
@@ -199,10 +199,9 @@ impl Vmar {
             let Some(mapped_va) = cursor.find_next(old_range.end) else {
                 break;
             };
-            let (va, Some(item)) = cursor.query().unwrap() else {
+            let Some(item) = cursor.query() else {
                 panic!("Found mapped page but query failed");
             };
-            debug_assert_eq!(mapped_va, va.start);
 
             let offset = mapped_va - old_range.start;
             let new_map_va = new_range.start + offset;
@@ -211,13 +210,13 @@ impl Vmar {
                 VmQueriedItem::MappedRam { frame, prop } => {
                     let frame = (*frame).clone();
 
-                    cursor.unmap(mapped_va + PAGE_SIZE);
+                    cursor.unmap();
                     cursor.jump(new_map_va).unwrap();
 
                     cursor.map(frame, prop);
                 }
                 VmQueriedItem::MappedIoMem { paddr, prop } => {
-                    cursor.unmap(mapped_va + PAGE_SIZE);
+                    cursor.unmap();
                     cursor.jump(new_map_va).unwrap();
 
                     // For MMIO pages, find the corresponding `IoMem` and map it
--- a/ostd/src/mm/page_table/cursor/mod.rs
+++ b/ostd/src/mm/page_table/cursor/mod.rs
@@ -154,20 +154,21 @@ impl<'rcu, C: PageTableConfig> Cursor<'rcu, C> {
 
     /// Moves the cursor forward to the next mapped virtual address.
     ///
-    /// If there is mapped virtual address following the current address within
-    /// next `len` bytes, it will return that mapped address. In this case, the
+    /// If there is a mapped virtual address at or after the current address
+    /// and before `end`, it will return that mapped address. In this case, the
     /// cursor will stop at the mapped address.
     ///
-    /// Otherwise, it will return `None`. And the cursor may stop at any
-    /// address after `len` bytes.
+    /// Otherwise, it will return `None`. The cursor may stop at any address
+    /// before `end`.
     ///
     /// # Panics
     ///
     /// Panics if:
-    ///  - the length is longer than the remaining range of the cursor;
-    ///  - the length is not page-aligned.
-    pub fn find_next(&mut self, len: usize) -> Option<Vaddr> {
-        self.find_next_impl(len, false, false)
+    ///  - `end` is before the current virtual address;
+    ///  - `end` exceeds the cursor's range;
+    ///  - `end` is not page-aligned.
+    pub fn find_next(&mut self, end: Vaddr) -> Option<Vaddr> {
+        self.find_next_impl(end, false, false)
     }
 
     /// Moves the cursor forward to the next fragment in the range.
@@ -181,21 +182,17 @@ impl<'rcu, C: PageTableConfig> Cursor<'rcu, C> {
     /// stop at leaf entries.
     ///
     /// `split_huge` specifies whether the cursor should split huge pages when
-    /// it finds a huge page that is mapped over the required range (`len`).
+    /// it finds a huge page that is mapped over the required range ending at
+    /// `end`.
     fn find_next_impl(
         &mut self,
-        len: usize,
+        end: Vaddr,
         find_unmap_subtree: bool,
         split_huge: bool,
     ) -> Option<Vaddr> {
-        assert_eq!(len % C::BASE_PAGE_SIZE, 0);
-        assert!(
-            len <= self.barrier_va.end - self.va,
-            "len exceeds remaining cursor range"
-        );
-
-        let end = self.va + len;
-        debug_assert_eq!(end % C::BASE_PAGE_SIZE, 0);
+        assert_eq!(end % C::BASE_PAGE_SIZE, 0);
+        assert!(end >= self.va, "end precedes current cursor position");
+        assert!(end <= self.barrier_va.end, "end exceeds cursor range");
 
         let rcu_guard = self.rcu_guard;
 
@@ -374,8 +371,8 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     /// Moves the cursor forward to the next mapped virtual address.
     ///
     /// This is the same as [`Cursor::find_next`].
-    pub fn find_next(&mut self, len: usize) -> Option<Vaddr> {
-        self.0.find_next(len)
+    pub fn find_next(&mut self, end: Vaddr) -> Option<Vaddr> {
+        self.0.find_next(end)
     }
 
     /// Jumps to the given virtual address.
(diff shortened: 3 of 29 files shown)


ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -48,13 +48,29 @@ use crate::{
...
+pub(crate) type CursorMut<'rcu, C> = Cursor_<'rcu, C, true>;
+
+/// The cursor for traversing the page table.
+///
+/// The `MUTABLE` generic parameter indicates whether the cursor has mutable
+/// access to the page table entries. Prefer to use [`CursorMut`] and [`Cursor`]
+/// types aliases.
 #[derive(Debug)]
-pub(crate) struct Cursor<'rcu, C: PageTableConfig> {
+pub(crate) struct Cursor_<'rcu, C: PageTableConfig, const MUTABLE: bool = false> {
```
lrh2000: Could you investigate whether this will cause the same function (e.g., `query`) to be [monomorphized](https://rustc-dev-guide.rust-lang.org/backend/monomorph.html) twice?

ostd/src/mm/vm_space.rs
```diff
@@ -434,120 +472,91 @@ impl<'a> CursorMut<'a> {
...
-            let Some(frag) = (unsafe { self.pt_cursor.take_next(end_va - self.virt_addr()) })
-            else {
-                break; // No more mappings in the range.
-            };
+    ///
+    /// Panics if the current level is at the top level.
+    pub fn unmap(&mut self) -> usize {
+        // SAFETY: It is safe to un-map memory in the userspace. And the
+        // un-mapped items are dropped after TLB flushes. So they outlive
+        // the current RCU lifetime `'a`.
```
lrh2000: I think you should try to resolve the merge conflicts correctly.

ostd/src/mm/vm_space.rs
```diff
@@ -434,120 +472,91 @@ impl<'a> CursorMut<'a> {
...
+                    } => {
+                        panic_guard.forget();
+                        // Flush the TLB entry for the current address, but
+                        // in the current design, we cannot drop the
+                        // corresponding `IoMem`. This is because we manage
+                        // the range of I/O as a whole, but the frames
+                        // handled here might be one segment of it.
+                        self.flusher.issue_tlb_flush(TlbFlushOp::for_single(va));
+
+                        nr_base_per_page::<UserPtConfig>(level)
```
lrh2000: Same?

ostd/src/mm/vm_space.rs
```diff
@@ -434,120 +472,91 @@ impl<'a> CursorMut<'a> {
...
+                num_frames,
+            } => {
+                // SAFETY: By `issue_tlb_flush_with`, the page table will
+                // be dropped after the RCU grace period.
+                let (pt, panic_guard) = unsafe { RcuDrop::into_inner(pt) };
+                let rcu_pt = RcuDrop::new(pt);
+                panic_guard.forget();
+                let rcu_frame = Frame::rcu_from_unsized(rcu_pt);
+                self.flusher
+                    .issue_tlb_flush_with(TlbFlushOp::for_range(va..va + len), rcu_frame);
```
lrh2000: Same? I'm not sure why you don't call `Frame::rcu_from_unsized`?

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -377,56 +409,20 @@ impl<C: PageTableConfig> CursorMut<'_, C> {
         let end = self.va + size;
         assert!(end <= self.barrier_va.end);
```
lrh2000: I suppose you should update something here:

- Some comments are outdated:
```rust
    /// # Panics
    ///
    /// This function will panic if
    ///  - the virtual address range to be mapped is out of the locked range;
```
 - `assert!(self.va < self.barrier_va.end);` and `assert!(end <= self.barrier_va.end);` shouldn't be necessary. If we still want them, they can be debug assertions like:
```rust
        let (_, level, _) = C::item_raw_info(&item);
        assert!(
            level <= C::HIGHEST_TRANSLATION_LEVEL,
            "cursor level not suitable for mapping"
        );
        let size = page_size::<C>(level);
        assert_eq!(
            self.va % size,
            0,
            "cursor virtual address not aligned for mapping"
        );

        let end = self.va + size;
        debug_assert!(self.va < self.barrier_va.end);
        debug_assert!(end <= self.barrier_va.end);
```

**EDIT:** Maybe `end <= self.barrier_va.end` still needs to be checked after this PR? If so, we can have something like this?
```rust
        debug_assert!(self.0.va < self.0.barrier_va.end);

        let (pa, level, prop) = C::item_into_raw(item);
        assert!(
            level <= C::HIGHEST_TRANSLATION_LEVEL,
            "cursor level not suitable for mapping"
        );
        let size = page_size::<C>(level);
        assert_eq!(
            self.0.va % size,
            0,
            "cursor virtual address not aligned for mapping"
        );
        let end = self.0.va + size;
        assert!(
            end <= self.0.barrier_va.end,
            "cursor virtual address out-of-bound for mapping"
        );
```

lrh2000 (commented):
I just checked a few random places in this PR. It's not a complete review. It's not even close to complete because I really only checked a few random places.

I'm submitting them anyway, since you should have an updated version of this PR after merging #2605.

ostd/src/mm/page_table/cursor/mod.rs
junyang-zh: In debug build `query` is monomorphized twice:

```
root@avalon:~/asterinas# objdump -t target/x86_64-unknown-none/debug/aster-kernel-osdk-bin | grep "page_table.*query"
ffffffff88990ab0 l     F .text  0000000000000113 .hidden _RNvMNtNtNtCs5YUapEsGKdy_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigE5queryB8_
ffffffff88991650 l     F .text  0000000000000113 .hidden _RNvMNtNtNtCs5YUapEsGKdy_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigKb1_E5queryB8_
```

In release build it is optimized away:

```
root@avalon:~/asterinas# objdump -t target/x86_64-unknown-none/release/aster-kernel-osdk-bin | grep "page_table.*query"
```

For another example, `push_level_if_exists`, we can see that it is monomorphized twice in release mode but the code is the same:

```
root@avalon:~/asterinas# objdump -t target/x86_64-unknown-none/debug/aster-kernel-osdk-bin | grep "page_table.*cursor.*push_level_if_exists"
ffffffff889be3a0 l     F .text  00000000000000b9 .hidden _RNvMNtNtNtCs2PGBv0vzYSM_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigE20push_level_if_existsB8_
ffffffff889becf0 l     F .text  00000000000000b9 .hidden _RNvMNtNtNtCs2PGBv0vzYSM_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigKb1_E20push_level_if_existsB8_
root@avalon:~/asterinas# objdump -t target/x86_64-unknown-none/release/aster-kernel-osdk-bin | grep "page_table.*cursor.*push_level_if_exists"
ffffffff883657f0 l     F .text  000000000000011f .hidden _RNvMNtNtNtCs8LA18vg2m3n_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigE20push_level_if_existsB8_
ffffffff883657f0 l     F .text  000000000000011f .hidden _RNvMNtNtNtCs8LA18vg2m3n_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigKb1_E20push_level_if_existsB8_
```

Another longer example, `unlock_range`, is monomorphized twice in debug mode but optimized away in release mode:

```
root@avalon:~/asterinas# objdump -t target/x86_64-unknown-none/debug/aster-kernel-osdk-bin | grep "page_table.*cursor.*unlock_range"
ffffffff886a9320 g     F .text  0000000000000241 _RINvNtNtNtNtCs2PGBv0vzYSM_4ostd2mm10page_table6cursor7locking12unlock_rangeNtNtB8_8vm_space12UserPtConfigKb0_ECsiG5DnrsjjPm_12aster_kernel
ffffffff886a9570 g     F .text  0000000000000241 _RINvNtNtNtNtCs2PGBv0vzYSM_4ostd2mm10page_table6cursor7locking12unlock_rangeNtNtB8_8vm_space12UserPtConfigKb1_ECsiG5DnrsjjPm_12aster_kernel
ffffffff88a00980 g     F .text  0000000000000241 _RINvNtNtNtNtCs2PGBv0vzYSM_4ostd2mm10page_table6cursor7locking12unlock_rangeNtNtB8_6kspace14KernelPtConfigKb1_EBa_
ffffffff88a00bd0 g     F .text  0000000000000241 _RINvNtNtNtNtCs2PGBv0vzYSM_4ostd2mm10page_table6cursor7locking12unlock_rangeNtNtNtNtNtBa_4arch5iommu13dma_remapping12second_stage13IommuPtConfigKb1_EBa_
root@avalon:~/asterinas# objdump -t target/x86_64-unknown-none/release/aster-kernel-osdk-bin | grep "page_table.*cursor.*unlock_range"
ffffffff88361cf0 l     F .text  000000000000013d .hidden _RINvNtNtNtNtCs8LA18vg2m3n_4ostd2mm10page_table6cursor7locking12unlock_rangeNtNtNtNtNtBa_4arch5iommu13dma_remapping12second_stage13IommuPtConfigKb1_EBa_
ffffffff88361bb0 l     F .text  000000000000013d .hidden _RINvNtNtNtNtCs8LA18vg2m3n_4ostd2mm10page_table6cursor7locking12unlock_rangeNtNtB8_6kspace14KernelPtConfigKb1_EBa_
```

So I guess the conclusion is, it will always result in duplicate code in debug mode. In release mode, I only find cases either both are optimized away or both exist.

ostd/src/mm/page_table/cursor/mod.rs
junyang-zh: > EDIT: Maybe end <= self.barrier_va.end still needs to be checked after this PR? If so, we can have something like this?

Yes this needs to be checked. I will add this to the `# Panics` section.

tatetian:
@boterinas codex

junyang-zh:
@boterinas codex

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -459,31 +458,23 @@ impl<C: PageTableConfig> CursorMut<'_, C> {
...
-
-        frag
+    /// Panics if the current level is at the top level and the corresponding
+    /// [`PageTableConfig::TOP_LEVEL_CAN_UNMAP`] is false.
+    pub unsafe fn unmap(&mut self) -> Option<PageTableFrag<C>> {
+        if !C::TOP_LEVEL_CAN_UNMAP && self.level == C::NR_LEVELS {
+            panic!("Unmapping top-level page table nodes");
+        }
+        self.replace_cur_entry(PteState::Absent)
     }
```
lrh2000, line 525: ### Make the cursor level externally observable and controllable.

Before this PR, the only externally observable and mutable state of a cursor was its current virtual address. I believe that all previous APIs will behave identically, regardless of the internal state of the cursor (e.g., the current level), as long as the virtual addresses are the same.

One big problem of the changes in this PR is that they will allow quite a few invalid states like:
```
barrier_va = HUGE_PAGE_SIZE - PAGE_SIZE..HUGE_PAGE_SIZE + PAGE_SIZE;
va = HUGE_PAGE_SIZE - PAGE_SIZE; // Or, `va == HUGE_PAGE_SIZE`.
level = 2;
// `protect()` and `unmap()` will succeed?
```

To me, those states appear to be much "more" invalid than the `va == barrier_va.end` case. But they are allowed in this PR. Also I don't think there is a way to forbid them.

So my current feelings are that making the cursor level externally observable is not ideal.

Additionally, it will make the code uglier than before. For example,
```rust
while cursor.push_if_exists().is_some() {}
match cursor.query() { /* .. */ }
```

But, on the other hand, the code changes in the PR make me wonder if we can have something like:
```rust
impl Cursor {
    fn query_leaf(&mut self) -> Entry<Leaf>;
    fn query_subtree(&mut self, level: PagingLevel) -> Entry<Subtree>;

    fn find_leaf(&mut self) -> Option<Leaf>;
    fn find_subtree(&mut self) -> Option<Subtree>;
}

enum Entry<T> {
    Value(T),
    Vacant(VacantEntry),
}

/// A cursor that stops at a PTE that containing an item.
struct Leaf {
    cursor: &mut Cursor,
    item: C::ItemRef,
}

impl Leaf {
    fn item(&self) -> C::ItemRef;

    fn protect(self, op: impl FnOnce());

    fn unmap(self);
}

/// A cursor that stops at a PTE that containing an item or points to a subtree.
struct Subtree {
    cursor: &mut Cursor,
    item: Option<C::ItemRef>, // Is it necessary?
}

impl Subtree {
    fn unmap(self);
}

/// A cursor that stops at a PTE that is empty.
struct VacantEntry {
    cursor: &mut Cursor,
}

impl VacantEntry {
    fn map(self, item: C::Item);
}
```

I think the above APIs are more reasonable, but I don't yet know how they interact with `VmSpace`'s public APIs.

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -275,30 +280,26 @@ impl<'rcu, C: PageTableConfig, const MUTABLE: bool> Cursor_<'rcu, C, MUTABLE> {
...
+    /// base-page-aligned, this method will return `Err`.
     pub fn jump(&mut self, va: Vaddr) -> Result<(), PageTableError> {
-        assert!(va.is_multiple_of(C::BASE_PAGE_SIZE));
-        if !self.barrier_va.contains(&va) {
+        if !va.is_multiple_of(C::BASE_PAGE_SIZE) || !self.barrier_va.contains(&va) {
             return Err(PageTableError::InvalidVaddr(va));
         }
 
-        // FIXME: Maintain the `self.barrier_va.contains(self.va)` invariant:
-        // <https://github.com/asterinas/asterinas/pull/2613>.
```
lrh2000, line 271: ### Forbid `self.va == barrier_va`.

What is the exact benefit that this type invariant provides?

Before:
```rust
while cursor.find_next().is_some() {
    cursor.unmap_next();
}
```
After:
```rust
while cursor.find_next().is_some() {
    cursor.unmap_next();
    if cursor.jump_next().is_err() {
        break;
    }
}
```

I think the code looks uglier this way. Using `self.va == barrier_va.end` to indicate that the cursor has been exhausted does _not_ seem like a bad idea.

From the PR description:
> But we want to maintain an invariant that the current virtual address must be within the range of each node in the path, so that every operation (like query or map) always succeeds.

It is impossible to ensure that `map()` always succeeds because mapping a page that was already mapped (maybe even at a different level) or mapping a too-large page must fail. We cannot know if or when this will happen at compile time.

For `protect()`/`unmap()`, refer to my comment above. The current approach to make it always succeed is not correct.

I agree that there is a benefit to having `query()` always succeed, but it seems minimal. It does not seem to outweigh the disadvantages of the ugly code related to the use cases of `find_next()`.

ostd/src/mm/vm_space.rs
```diff
@@ -306,6 +302,16 @@ impl Cursor<'_> {
...
+
+    /// Get the current level of the cursor.
+    pub fn level(&self) -> PagingLevel {
+        self.0.level()
+    }
+
+    /// Get the current virtual address range of the cursor.
+    pub fn cur_va_range(&self) -> Range<Vaddr> {
+        self.0.cur_va_range()
+    }
```
lrh2000: `Get` -> `Gets`

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -275,30 +280,26 @@ impl<'rcu, C: PageTableConfig, const MUTABLE: bool> Cursor_<'rcu, C, MUTABLE> {
         None
```
lrh2000: You should remove the code and convert the above `while` loop to `loop { ... return None; }`.

`while self.va < end` is always true.

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -459,31 +458,23 @@ impl<C: PageTableConfig> CursorMut<'_, C> {
...
-        let frag = self.replace_cur_entry(PteState::Absent);
-
-        self.move_forward();
-
-        frag
+    /// Panics if the current level is at the top level and the corresponding
+    /// [`PageTableConfig::TOP_LEVEL_CAN_UNMAP`] is false.
+    pub unsafe fn unmap(&mut self) -> Option<PageTableFrag<C>> {
+        if !C::TOP_LEVEL_CAN_UNMAP && self.level == C::NR_LEVELS {
+            panic!("Unmapping top-level page table nodes");
```
lrh2000:
```suggestion
            panic!("unmapping top-level page table nodes");
```

ostd/src/mm/page_table/cursor/mod.rs
```diff
@@ -349,35 +350,68 @@ impl<'rcu, C: PageTableConfig, const MUTABLE: bool> Cursor_<'rcu, C, MUTABLE> {
...
+        }
+    }
+
     /// Maps the item starting from the current address to a physical address range.
     ///
     /// The current virtual address should not be mapped.
     ///
     /// # Panics
     ///
     /// This function will panic if
```
lrh2000:
```suggestion
    /// Panics if
```

This seems the only exception in the whole file. All other doc comments use "Panics if"?

ostd/src/mm/page_table/node/entry.rs
```diff
@@ -41,9 +41,10 @@ impl<'a, 'rcu, C: PageTableConfig> Entry<'a, 'rcu, C> {
     /// Gets a reference to the child.
     pub(in crate::mm) fn to_ref(&self) -> PteStateRef<'rcu, C> {
         // SAFETY:
-        //  - The child pointed to by the PTE outlives the reference, since
-        //    either PTs and mapped items outlive `'rcu`.
-        //  - The level matches the current node.
+        //  - The PTE was read from the node, which contains valid PTEs;
+        //  - the child pointed to by the PTE outlives the reference, since
+        //    either PTs and mapped items outlive `'rcu`;
+        //  - the level matches the current node.
```
lrh2000: I think we use:
```rust
// The caller must ensure that:   // <- This is an incomplete sentence.
// - aaa;
// - aaa;
// - aaa.
```
And:
```rust
// SAFETY:  // <- This is not a sentence.
// - Aaa.
// - Aaa.
// - Aaa.
```

However, it seems that you use a style that combines the two above. I think we should use the second style instead.

ostd/src/mm/page_table/node/mod.rs
```diff
@@ -187,6 +187,17 @@ impl<'rcu, C: PageTableConfig> PageTableGuard<'rcu, C> {
...
     }
 
+    /// Gets the state of the entry at a given index.
+    pub(super) fn entry_state(&self, idx: usize) -> PteStateRef<'rcu, C> {
+        let pte = self.read_pte(idx);
+        // SAFETY:
+        //  - The PTE must be the result of `PteState::into_pte` since it was
+        //    written into the node;
+        //  - the child outlives `'rcu`;
+        //  - the level matches the residing node.
```
lrh2000: Same here.

ostd/src/mm/page_table/node/entry.rs
```diff
@@ -209,8 +210,7 @@ impl<'a, 'rcu, C: PageTableConfig> Entry<'a, 'rcu, C> {
     ///
     /// The caller must ensure that the index is within the bounds of the node.
     pub(super) unsafe fn new_at(guard: &'a mut PageTableGuard<'rcu, C>, idx: usize) -> Self {
-        // SAFETY: The index is within the bound.
-        let pte = unsafe { guard.read_pte(idx) };
+        let pte = guard.read_pte(idx);
         Self {
```
lrh2000: The safety comment is outdated.

ostd/src/mm/page_table/cursor/mod.rs
lrh2000: I was thinking about to use:
```rust
impl Deref for CursorMut {
    fn deref(&self) -> &Cursor;
}
```
which is guaranteed not to monomorphized the same implementation twice.

---

> ffffffff883657f0 l     F .text  000000000000011f .hidden _RNvMNtNtNtCs8LA18vg2m3n_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigE20push_level_if_existsB8_
> ffffffff883657f0 l     F .text  000000000000011f .hidden _RNvMNtNtNtCs8LA18vg2m3n_4ostd2mm10page_table6cursorINtB2_7Cursor_NtNtB6_8vm_space12UserPtConfigKb1_E20push_level_if_existsB8_

Perhaps the const generic approach is also fine if the release build will merge two identical implementations.There are two symbols but they're both referring to the code at `ffffffff883657f0`.

I also tried to figure out a better name than `Cursor_`. But I failed.

lrh2000 (commented):
Sorry for the extremely long delay...

So there are two major changes in this PR;
 - Make the cursor level externally observable and controllable.
 - Forbid `self.va == barrier_va`.

To be honest, checking the changes in this PR, I am not entirely sure either of them look the right direction. For details, see the comments below.

Note that maybe your part of motivation is to enable CortenMM, however, I did not (and cannot, due to lack of knowledge) take that into consideration. So if it's part of your motivation you should elaborate more.

kernel/src/vm/vmar/vmar_impls/remap.rs
```diff
@@ -199,31 +202,30 @@ impl Vmar {
...
             let offset = mapped_va - old_range.start;
             let new_map_va = new_range.start + offset;
 
-            match item {
+            match cursor.query() {
                 VmQueriedItem::MappedRam { frame, prop } => {
                     let frame = (*frame).clone();
 
-                    cursor.unmap(PAGE_SIZE);
+                    cursor.unmap();
```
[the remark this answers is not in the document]
junyang-zh: No need. Currently userspace cannot map huge pages so it is so far good.

junyang-zh:
> Note that maybe your part of motivation is to enable CortenMM, however, I did not (and cannot, due to lack of knowledge) take that into consideration. So if it's part of your motivation you should elaborate more.

Thank you. 

> Make the cursor level externally observable and controllable.

The motivation to make the cursor level externally visible is to enable CortenMM. Since when storing extra state (VMA pieces actually) in the page table, the caller need to put it in the correct level. Therefore the cursor must allow the user to navigate to specific levels and do operations.

> Forbid `self.va == barrier_va`.

This is not actually related with CortenMM. It's because I find the by-default advance behavior of `map`, `unmap` and `protect` not very tidy at some places, e.g., mapping or skipping already present pages:

https://github.com/asterinas/asterinas/blob/6c314ef6da150a9ce394642bbb0d89c81b045da6/kernel/src/vm/vmar/vm_mapping.rs#L621-L628

, some inconsistencies:

https://github.com/asterinas/asterinas/blob/6c314ef6da150a9ce394642bbb0d89c81b045da6/kernel/src/vm/vmar/vmar_impls/fork.rs#L105-L108

, and jumping back after `unmap` to `map`:

https://github.com/asterinas/asterinas/blob/6c314ef6da150a9ce394642bbb0d89c81b045da6/kernel/src/vm/vmar/vm_mapping.rs#L474-L476

If we do not let `map`, `unmap` and `protect` advance after operation by default, there seems no need to have `self.va == barrier_va`. Therefore I disallowed it. Another possible approach would be not advance by default and allow `self.va == barrier_va`. So the only case that will exhaust the cursor is `find_next`. It seems to resolve your concerns about the tidiness of `find_next` cases.

lrh2000:
> The motivation to make the cursor level externally visible is to enable CortenMM. Since when storing extra state (VMA pieces actually) in the page table, the caller need to put it in the correct level. Therefore the cursor must allow the user to navigate to specific levels and do operations.

OK. Then will the APIs listed in https://github.com/asterinas/asterinas/pull/2613#discussion_r3649609756 work for you?

One problem I am aware of is that: There will be a disaster if we have all the structures and methods duplicate for the `VmSpace` cursor APIs.

If it's the only problem, I hope we can find a way to resolve it. I'm not sure how to do that at the moment, though.

> mapping or skipping already present pages:
> , some inconsistencies:
> , and jumping back after `unmap` to `map`:

Thanks for the explanation! Avoiding the automatic advance of the cursor can simplify some code. That makes sense.

It seems that both have their use cases, in an ideal world, maybe we can end up with something like:
```
cursor.find().map().next();
```
This allows each caller to decide whether to move forward.

But perhaps this is a fairly minor point.... and we don't really need to do it if implementing so is difficult.

> So the only case that will exhaust the cursor is `find_next`. It seems to resolve your concerns about the tidiness of `find_next` cases.

I guess this is not very related...
```patch
 while cursor.find_next().is_some() {
     cursor.unmap_next();
+    if cursor.jump_next().is_err() {
+        break;
+    }
 }
```

I want to use this use case to argue that `unmap` should advance the cursor. However, you also provided me with some counterexamples that suggest advancing might be redundant. What you said also makes sense to me.

junyang-zh:
> Then will the APIs listed in https://github.com/asterinas/asterinas/pull/2613#discussion_r3649609756 work for you?

It is indeed a good suggestion and I can understand. I perhaps need some time to see if it requires too much effort.

kernel/core/src/vm/vmar/vm_mapping.rs
```diff
@@ -481,15 +483,15 @@ impl VmMapping {
                     }
                     cursor.flusher().sync_tlb_flush();
                 }
```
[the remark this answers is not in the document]
junyang-zh: Yes but not related with this branch. Fix in https://github.com/asterinas/asterinas/pull/3651

junyang-zh:
I prototyped the proposal of @lrh2000 at https://github.com/asterinas/asterinas/commit/93352fd7009c8754eb5dff8762ba3aef4ca0aebb, with some help from AI tools. It can make some call site clearer and make invalid state explicit.

My concern of this approach is that it may introduce much complexities into OSTD (around 400 diff lines), especially when we need to duplicate it into VmSpace cursor. I currently do not have an idea to solve it. The benefited caller sites are not many. Therefore I still prefer the current solution.

I think this state machine style API can be good if we can make it a safe wrapper for current raw OSTD APIs. Then we can balance OSTD complexity and the caller's ease.

Further thoughts @lrh2000 ? Do the proposal at https://github.com/asterinas/asterinas/commit/93352fd7009c8754eb5dff8762ba3aef4ca0aebb implement what you think?

lrh2000:
What is your response to the "problem" part? My main motivation is to forbid invalid states at the type level. Assuming we keep the current implementation, how can we best resolve this issue? Just add more assertions?

The "problem" part is:
> One big problem of the changes in this PR is that they will allow quite a few invalid states like:
> 
> ```
> barrier_va = HUGE_PAGE_SIZE - PAGE_SIZE..HUGE_PAGE_SIZE + PAGE_SIZE;
> va = HUGE_PAGE_SIZE - PAGE_SIZE; // Or, `va == HUGE_PAGE_SIZE`.
> level = 2;
> // `protect()` and `unmap()` will succeed?
> ```
> 
> To me, those states appear to be much "more" invalid than the `va == barrier_va.end` case. But they are allowed in this PR. Also I don't think there is a way to forbid them.
> 
> So my current feelings are that making the cursor level externally observable is not ideal.

---

> My concern of this approach is that it may introduce much complexities into OSTD (around 400 diff lines), especially when we need to duplicate it into VmSpace cursor. I currently do not have an idea to solve it.

Could you analyze it further to determine where it comes from?

From what I can see:
 1. I don't understand why the `VmSpace` types do not wrap the `Subtree`/`Leaf` types. Instead, they wrap `CursorMut`. This leads to duplicate checks at runtime and bloats the code with quite a few `unreachable!()`.
 2. For some reason, AI is generally "good" at defensive programming, so it inserts code like this:
    ```
                        let level = self
                            .0
                            .level
                            .checked_sub(1)
                            .expect("a base-page mapping cannot cross the cursor boundary");
    ```
    Code like this looks awful and makes it longer.
 3. We can avoid defining and implementing `unmap()` twice (once for `Subtree` and once for `Leaf`) by introducing `Leaf::to_subtree() -> Subtree`.
 4. Part of the code is boilerplate code, such as `impl Deref{,Mut}`. If we can have `#[derive(Deref, DerefMut)]`, they may go away.
 5. ...

> The benefited caller sites are not many.

It seems that there are quite `unreachable!()` which comes from `entry.unmap()` + `cursor.query()`. I think `entry.unmap()` should probably return a `VacantEntry`. So the code will look much better.

junyang-zh:
> What is your response to the "problem" part?

Sorry I missed that. However just add more assertions seems OK. Nearly no additional code since `map` already checks them, so we just need to add the same checks to `protect` and `unmap`. `query` can still have the current range exceeding the barrier.

<details><summary>Diff</summary>
<p>

```
diff --git a/ostd/src/mm/page_table/cursor/mod.rs b/ostd/src/mm/page_table/cursor/mod.rs
index e01146506..bac316066 100644
--- a/ostd/src/mm/page_table/cursor/mod.rs
+++ b/ostd/src/mm/page_table/cursor/mod.rs
@@ -429,7 +429,8 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     ///  - the cursor level does not match the level of the item to be mapped;
     ///  - the current virtual address is not aligned to the page size of the
     ///    item to be mapped;
-    ///  - the end of the current virtual address range exceeds the locked range;
+    ///  - the current virtual address range is not fully contained in the
+    ///    cursor range;
     ///  - the current virtual address range contains mappings.
     ///
     /// # Safety
@@ -438,8 +439,6 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     ///  - the range being mapped does not affect kernel's memory safety;
     ///  - the physical address to be mapped is valid and safe to use.
     pub unsafe fn map(&mut self, item: C::Item) {
-        debug_assert!(self.va < self.barrier_va.end);
-
         let (_, level, _) = C::item_raw_info(&item);
         assert!(
             level <= C::HIGHEST_TRANSLATION_LEVEL,
@@ -449,17 +448,7 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
             self.level, level,
             "cursor level do not match the item mapping level"
         );
-        let size = page_size::<C>(level);
-        assert_eq!(
-            self.va % size,
-            0,
-            "cursor virtual address not aligned for mapping"
-        );
-        let end = self.va + size;
-        assert!(
-            end <= self.barrier_va.end,
-            "cursor virtual address out-of-bound for mapping"
-        );
+        self.assert_cur_va_range_valid();
 
         if !matches!(self.cur_entry().to_ref(), PteStateRef::Absent) {
             panic!("mapping over an already mapped page");
@@ -488,9 +477,15 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     ///
     /// # Panics
     ///
-    /// Panics if the current level is at the top level and the corresponding
-    /// [`PageTableConfig::TOP_LEVEL_CAN_UNMAP`] is false.
+    /// Panics if:
+    ///  - the current virtual address is not aligned to the current entry
+    ///    range;
+    ///  - the current virtual address range is not fully contained in the
+    ///    cursor range;
+    ///  - the current level is at the top level and the corresponding
+    ///    [`PageTableConfig::TOP_LEVEL_CAN_UNMAP`] is false.
     pub unsafe fn unmap(&mut self) -> Option<PageTableFrag<C>> {
+        self.assert_cur_va_range_valid();
         if !C::TOP_LEVEL_CAN_UNMAP && self.level == C::NR_LEVELS {
             panic!("unmapping top-level page table nodes");
         }
@@ -507,6 +502,14 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     /// It only modifies the page properties if the current entry state is
     /// [`PteState::Mapped`]. Otherwise, it does nothing.
     ///
+    /// # Panics
+    ///
+    /// Panics if:
+    ///  - the current virtual address is not aligned to the current entry
+    ///    range;
+    ///  - the current virtual address range is not fully contained in the
+    ///    cursor range.
+    ///
     /// # Safety
     ///
     /// The caller should ensure that:
@@ -515,9 +518,22 @@ impl<'rcu, C: PageTableConfig> CursorMut<'rcu, C> {
     ///  - the privileged flag `AVAIL1` should not be altered, since this flag
     ///    is reserved for all page tables.
     pub unsafe fn protect(&mut self, op: &mut impl FnMut(&mut PageProperty)) {
+        self.assert_cur_va_range_valid();
         self.cur_entry().protect(op);
     }
 
+    fn assert_cur_va_range_valid(&self) {
+        let cur_va_range = self.cur_va_range();
+        assert_eq!(
+            self.va, cur_va_range.start,
+            "current virtual address is not aligned to the current entry range"
+        );
+        assert!(
+            self.barrier_va.start <= cur_va_range.start && cur_va_range.end <= self.barrier_va.end,
+            "current virtual address range exceeds cursor range"
+        );
+    }
+
     fn replace_cur_entry(&mut self, new_child: PteState<C>) -> Option<PageTableFrag<C>> {
         let rcu_guard = self.rcu_guard;
 
diff --git a/ostd/src/mm/page_table/test.rs b/ostd/src/mm/page_table/test.rs
index c0877fb99..79bd05cfa 100644
--- a/ostd/src/mm/page_table/test.rs
+++ b/ostd/src/mm/page_table/test.rs
@@ -471,7 +471,7 @@ mod overlapping_mappings {
     }
 
     #[ktest]
-    #[should_panic(expected = "cursor virtual address not aligned for mapping")]
+    #[should_panic(expected = "current virtual address is not aligned to the current entry range")]
     fn unaligned_map() {
         const HUGE_PAGE_SIZE: usize = PAGE_SIZE * 512;
 
@@ -508,6 +508,85 @@ mod overlapping_mappings {
     }
 }
 
+mod cursor_range_checks {
+    use super::{test_utils::*, *};
+
+    const HUGE_PAGE_SIZE: usize = PAGE_SIZE * 512;
+    const ALIGNED_CURSOR_RANGE: Range<Vaddr> = 0..PAGE_SIZE;
+    const UNALIGNED_CURSOR_RANGE: Range<Vaddr> = PAGE_SIZE..PAGE_SIZE * 2;
+
+    fn page_table_with_huge_mapping() -> PageTable<TestPtConfig> {
+        let page_table = PageTable::<TestPtConfig>::empty();
+        let prop = PageProperty::new_user(PageFlags::RW, CachePolicy::Writeback);
+        map_untracked(&page_table, 0..HUGE_PAGE_SIZE * 2, 0, prop);
+        page_table
+    }
+
+    #[ktest]
+    fn query_allows_range_outside_cursor() {
+        let page_table = page_table_with_huge_mapping();
+        let preempt_guard = disable_preempt();
+        let cursor = page_table
+            .cursor(&preempt_guard, &UNALIGNED_CURSOR_RANGE)
+            .unwrap();
+
+        assert_eq!(cursor.level(), 2);
+        assert!(matches!(cursor.query(), PteStateRef::Mapped(_)));
+    }
+
+    #[ktest]
+    #[should_panic(expected = "current virtual address range exceeds cursor range")]
+    fn protect_rejects_range_outside_cursor() {
+        let page_table = page_table_with_huge_mapping();
+        let preempt_guard = disable_preempt();
+        let mut cursor = page_table
+            .cursor_mut(&preempt_guard, &ALIGNED_CURSOR_RANGE)
+            .unwrap();
+
+        assert_eq!(cursor.level(), 2);
+        unsafe { cursor.protect(&mut |_| {}) };
+    }
+
+    #[ktest]
+    #[should_panic(expected = "current virtual address range exceeds cursor range")]
+    fn unmap_rejects_range_outside_cursor() {
+        let page_table = page_table_with_huge_mapping();
+        let preempt_guard = disable_preempt();
+        let mut cursor = page_table
+            .cursor_mut(&preempt_guard, &ALIGNED_CURSOR_RANGE)
+            .unwrap();
+
+        assert_eq!(cursor.level(), 2);
+        let _ = unsafe { cursor.unmap() };
+    }
+
+    #[ktest]
+    #[should_panic(expected = "current virtual address is not aligned to the current entry range")]
+    fn protect_rejects_unaligned_current_address() {
+        let page_table = page_table_with_huge_mapping();
+        let preempt_guard = disable_preempt();
+        let mut cursor = page_table
+            .cursor_mut(&preempt_guard, &UNALIGNED_CURSOR_RANGE)
+            .unwrap();
+
+        assert_eq!(cursor.level(), 2);
+        unsafe { cursor.protect(&mut |_| {}) };
+    }
+
+    #[ktest]
+    #[should_panic(expected = "current virtual address is not aligned to the current entry range")]
+    fn unmap_rejects_unaligned_current_address() {
+        let page_table = page_table_with_huge_mapping();
+        let preempt_guard = disable_preempt();
+        let mut cursor = page_table
+            .cursor_mut(&preempt_guard, &UNALIGNED_CURSOR_RANGE)
+            .unwrap();
+
+        assert_eq!(cursor.level(), 2);
+        let _ = unsafe { cursor.unmap() };
+    }
+}
+
 mod navigation {
     use super::{test_utils::*, *};
 
diff --git a/ostd/src/mm/vm_space.rs b/ostd/src/mm/vm_space.rs
index 053d015ab..9832ad217 100644
--- a/ostd/src/mm/vm_space.rs
+++ b/ostd/src/mm/vm_space.rs
@@ -421,7 +421,13 @@ impl<'a> CursorMut<'a> {
     ///
     /// # Panics
     ///
-    /// Panics if the current virtual address is already mapped.
+    /// Panics if:
+    ///  - the current virtual address is not aligned to the current entry
+    ///    range;
+    ///  - the current virtual address range is not fully contained in the
+    ///    cursor range;
+    ///  - the cursor level does not match the frame's mapping level;
+    ///  - the current virtual address is already mapped.
     pub fn map(&mut self, frame: UFrame, prop: PageProperty) {
         let item = VmItem::new_tracked(frame, prop);
 
@@ -508,6 +514,14 @@ impl<'a> CursorMut<'a> {
     /// unmapped range.
     ///
     /// The number of unmapped frames is returned.
+    ///
+    /// # Panics
+    ///
+    /// Panics if:
+    ///  - the current virtual address is not aligned to the current entry
+    ///    range;
+    ///  - the current virtual address range is not fully contained in the
+    ///    cursor range.
     pub fn unmap(&mut self) -> usize {
         // SAFETY:
         // 1. It is safe to unmap memory in the userspace.
@@ -575,6 +589,14 @@ impl<'a> CursorMut<'a> {
     /// cursor, and can be queried via [`Self::cur_va_range`]. Adjust the
     /// level via [`Self::adjust_level`] before protecting to change the
     /// protected range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if:
+    ///  - the current virtual address is not aligned to the current entry
+    ///    range;
+    ///  - the current virtual address range is not fully contained in the
+    ///    cursor range.
     pub fn protect(&mut self, mut op: impl FnMut(&mut PageFlags, &mut CachePolicy)) {
         // SAFETY: It is safe to set `PageFlags` and `CachePolicy` of memory
         // in the userspace.
```

</p>
</details> 

> Could you analyze it further to determine where it comes from?

Yes I admit that these AI generated code is not ideal. Yet if we do the best I think it still adds complexity. I am doubtful if the effort worth if the only problem is more panic cases.

> I don't understand why the `VmSpace` types do not wrap the `Subtree`/`Leaf` types.

Because operations (e.g., `unmap`) of `page_table::Subtree`/`page_table::Leaf` types are unsafe. And `VmSpace` types still need to wrap the `vm_space::Cursor` to access the TLB flusher.

lrh2000:
> I am doubtful if the effort worth if the only problem is more panic cases.

I think the problem is that exposing the cursor level directly leads to rather surprising API surface. I think it's very odd to require the caller to adjust the level manually.

In particular. I think following code is not ideal:
```rust
            while cursor.push_level_if_exists().is_some() {}
            let item = cursor.query();
```
```rust
        while cursor.find_next_unmappable_subtree(range.end).is_some() {
            while cursor.cur_va_range().start < range.start || cursor.cur_va_range().end > range.end
            {
                cursor.adjust_level(cursor.level() - 1);
            }
            num_unmapped += cursor.unmap();
        }
```
```rust
                VmQueriedItem::PageTable => {
                    unreachable!("pushed but still queried a page table")
                }
```
(There may be more, but I just selected some.)

Furthermore, I think we should remove the `VmQueriedItem::None` and `VmQueriedItem::PageTable` variants. `Leaf::item()` shouldn't return them, after all. This is also not done in your prototype. However, I think we should do it so that the benefit is clearer.

> Because operations (e.g., `unmap`) of `page_table::Subtree`/`page_table::Leaf` types are unsafe.

Yes. So why `vm_space::Subtree` wraps `page_table::CursorMut` instead of `page_table::Subtree`? This does not answer my question.

junyang-zh:
@lrh2000 

> I think it's very odd to require the caller to adjust the level manually.

Yes I admit that the caller's additional code seems ugly. However in CortenMM the caller indeed need more control and they are wrapped in macros to make it concise. This PR is part of the effort to move forward.

> I think we should remove the VmQueriedItem::None and VmQueriedItem::PageTable variants.

Also this is needed in CortenMM.

I agree that we cannot include CortenMM as the motivation of this standalone PR, I think I can move the last commit to #2615 .

> Yes. So why `vm_space::Subtree` wraps `page_table::CursorMut` instead of `page_table::Subtree`?

In https://github.com/asterinas/asterinas/commit/93352fd7009c8754eb5dff8762ba3aef4ca0aebb `vm_space::Subtree` wraps `vm_space::CursorMut` instead of `page_table::CursorMut`. I believe "And `VmSpace` types still need to wrap the `vm_space::Cursor` to access the TLB flusher." answers your question.

lrh2000:
> Also this is needed in CortenMM.

Why doesn't it work with the `Subtree` APIs? These also allow you to select the level without exposing it directly in the cursor.
```rust
fn query_subtree(&mut self, level: PagingLevel) -> Subtree<'_, 'rcu>;
```

> I agree that we cannot include CortenMM as the motivation of this standalone PR

The real problem is that you never well explained what you need, except from some obscure words like "more control".

For now, the most detailed response from you about what CortenMM needs is the following statement:
>> The motivation to make the cursor level externally visible is to enable CortenMM. Since when storing extra state (VMA pieces actually) in the page table, the caller need to put it in the correct level. Therefore the cursor must allow the user to navigate to specific levels and do operations.

I think `Subtree` APIs work much better for exactly the same purpose. Avoiding `VmQueriedItem::None` and `VmQueriedItem::PageTable` is a real benefit when we're knowing at a non-empty, leaf PTE.

> I believe "And `VmSpace` types still need to wrap the `vm_space::Cursor` to access the TLB flusher." answers your question.

Shouldn't `vm_space::Subtree` wraps `page_table::Subtree`, plus a mutable reference to the TLB flusher of `vm_space::Cursor`? I'm still not sure what I missed.

junyang-zh:
> Shouldn't vm_space::Subtree wraps page_table::Subtree, plus a mutable reference to the TLB flusher of vm_space::Cursor?

That, would be two mutable references to two fields in one object, which would be rejected by the compiler?

lrh2000:
> > Shouldn't vm_space::Subtree wraps page_table::Subtree, plus a mutable reference to the TLB flusher of vm_space::Cursor?
> 
> That, would be two mutable references to two fields in one object, which would be rejected by the compiler?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1b379df1f1c334dcbe5b3f69f7461185

junyang-zh:
Sorry for the delay.

> I think the problem is that exposing the cursor level directly leads to rather surprising API surface. I think it's very odd to require the caller to adjust the level manually.
>
> In particular. I think following code is not ideal:

Let me clarify my position first. The ideal Cursor API I want is actually a minimal, self-contained set of operations for arbitrary complicated operations. This PR is still not ideal in this sense. Actually, `find_next` should be removed from OSTD, and we should let the caller implement such logic. Complexity on the caller side is a minor goal. We can always add a layer of helpers for it. However, ambiguity on the caller side is a bigger problem.

The most problematic ambiguity so far is whether the cursor will automatically modify the page table upon queries. E.g., if `query_leaf` is called on a huge page or an empty upper-level entry, would it split the page/add new child page tables? What if `find_next_unmappable_subtree` is called on a huge page with the end exceeding the `end` but the start ahead of `end`? In your ideal case, it seems that automatic modifications are needed to make the caller site simple. But I want such a caller site to explicitly say what modifications are needed. I don't think the subtree/leaf API helps in this case and adds more complexity if we do not want automatic modifications.

Furthermore, I don't think there is any way to reduce the net added lines to less than 100 to implement the subtree/leaf API, even if your suggestions (avoiding `VmQueriedItem::None`, adding `Leaf::to_subtree()`, merging definitions, ...) are applied.

junyang-zh:
I added some wrappers in the kernel. Now it looks like:

```
            let item = cursor.to_leaf().query();
```

```
        while cursor.find_next_unmappable_subtree(range.end).is_some() {
            cursor.split_if_map_exceeds_range(&range);
            num_unmapped += cursor.unmap();
        }
```

PR #1374, opened 2024-09-24
Introduce kernel virtual memory allocation

vvvvsv:
https://github.com/asterinas/asterinas/issues/929

PR #2614, opened 2025-11-25
Remove the x86 kernel page fault handler

junyang-zh:
`IoMem` is already implemented by `KVirtArea` in #1374 . It seems that the kernel page fault handler is no longer needed. So I removed it in this PR. Need confirmation that linear mapping is not used for I/O in TDX @Hsy-Intel @taosue .

### DIFF
--- a/ostd/src/arch/x86/trap/mod.rs
+++ b/ostd/src/arch/x86/trap/mod.rs
@@ -20,27 +20,19 @@ pub(super) mod gdt;
 mod idt;
 mod syscall;
 
-use align_ext::AlignExt;
 use cfg_if::cfg_if;
-use log::debug;
 use spin::Once;
 
 use super::cpu::context::GeneralRegs;
 use crate::{
     arch::{
-        cpu::context::{CpuException, PageFaultErrorCode, RawPageFaultInfo},
-        if_tdx_enabled,
+        cpu::context::CpuException,
         irq::{disable_local, enable_local, HwIrqLine},
     },
     cpu::PrivilegeLevel,
     ex_table::ExTable,
     irq::call_irq_callback_functions,
-    mm::{
-        kspace::{KERNEL_PAGE_TABLE, LINEAR_MAPPING_BASE_VADDR, LINEAR_MAPPING_VADDR_RANGE},
-        page_prop::{CachePolicy, PageProperty},
-        PageFlags, PrivilegedPageFlags as PrivFlags, MAX_USERSPACE_VADDR, PAGE_SIZE,
-    },
-    task::disable_preempt,
+    mm::MAX_USERSPACE_VADDR,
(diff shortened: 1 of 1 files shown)


taosue:
Good cleanup! Based on my understanding, I/O memory should not be accessed through linear mapping, but let's wait for @Hsy-Intel to double confirm.

However, I noticed that `unprotect_gpa_range()` still flushes TLB for an incorrect virtual address - i.e., the linear mapping of I/O memory. While this doesn't trigger an error currently, should we ensure we're flushing the correct virtual address range?

junyang-zh:
> However, I noticed that unprotect_gpa_range() still flushes TLB for an incorrect virtual address - i.e., the linear mapping of I/O memory. While this doesn't trigger an error currently, should we ensure we're flushing the correct virtual address range?

Fixing it would be simple. Would you like to open up a PR for this?

taosue:
> Fixing it would be simple. Would you like to open up a PR for this?

Sure, I will try.

Hsy-Intel:
No questions about this PR. As @taosue's comment, we may need to modify the `unprotect_gpa_range()` and `protect_gpa_range()` functions.

ostd/src/arch/x86/trap/mod.rs
```diff
@@ -177,7 +169,10 @@ extern "sysv64" fn trap_handler(f: &mut TrapFrame) {
             if (0..MAX_USERSPACE_VADDR).contains(&raw_page_fault_info.addr) {
                 handle_user_page_fault(f, cpu_exception.as_ref().unwrap());
             } else {
-                handle_kernel_page_fault(raw_page_fault_info);
+                panic!(
+                    "Kernel page fault: {:#x?}, trapframe: {:#x?}",
+                    raw_page_fault_info, f
+                );
```
lrh2000, line 175: So maybe you can update line 172-175 & line 181-184 to use a consistent style here?

For example,
```rust
                panic!(
                    "Cannot handle kernel page fault: {:#x?}, trapframe: {:#x?}",
                    raw_page_fault_info, f
                );
```

```rust
            panic!(
                "Cannot handle kernel CPU exception: {:#x?}, trapframe: {:#x?}",
                exception, f
            );
```

lrh2000:
Thanks for the cleanup. So I think the above discussion means that this PR can be merged without affecting TDX usage? `unprotect_gpa_range` seems to be a separate issue anyway and not really related to the kernel page fault handler?

tatetian:
> Thanks for the cleanup. So I think the above discussion means that this PR can be merged without affecting TDX usage? `unprotect_gpa_range` seems to be a separate issue anyway and not really related to the kernel page fault handler?

I think so.

tatetian (approved):

PR #2616, opened 2025-11-25
Fix many evdev bugs

lrh2000:
cc @Cautreoxit Could you please check if the changes in this PR are correct? Thanks!

---

This is a commit-by-commit explanation of this PR:

#### Use `size_of` in preludes 

Follows https://github.com/asterinas/asterinas/pull/2403.

#### Clarify error codes and messages

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/file.rs#L250-L252

This is wrong because evdev files fail with `ESPIPE` at `lseek()`.

This commit also adds some additional comments for several other error codes.

#### Remove the unused event count

`event_count` is useless and we can always use `consumer.lock().len()` instead.

#### Simplify the logic related to I/O events

This aligns the style of I/O event handling:
https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/pty/master.rs#L99-L106
https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/pty/master.rs#L83-L86

#### Align the read behavior to Linux's

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/file.rs#L167-L179
This is not correct because it forgets to update the counters if `write_val` fails.

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/file.rs#L215-L217
This is not correct because some checks are necessary even if `max_events` is zero.

#### Define a specific clock type

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/mod.rs#L199-L202

This is confusing. For evdev files, it can only support a limited set of clock IDs and we should define a specific type for it.

#### Fix the handling when the buffer is full

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/mod.rs#L146-L147

AFAICK, this is not correct. See a detailed explanation below.

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/mod.rs#L157-L158

This will almost always fail so no `SYN_DROPPED` events can be pushed. I think we can have a better strategy to push the `SYN_DROPPED` event when the buffer is almost full.

```rust
                // When the buffer is full and a new event arrives, Linux drops all unconsumed
                // events and queues a `SYN_DROPPED` event with the new one [1].
                //
                // Since we cannot remove events from the producer, we queue a `SYN_DROPPED` event
                // when the buffer is almost full to indicate that events are about to be dropped.
                // This should match the correct semantics of the `SYN_DROPPED` event [2].
                //
                // [1]: https://elixir.bootlin.com/linux/v6.17.9/source/drivers/input/evdev.c#L221-L225
                // [2]: https://elixir.bootlin.com/linux/v6.17.9/source/Documentation/input/event-codes.rst#L113-L118
```

#### Fix potential deadlocks

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/mod.rs#L128-L130
https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/file.rs#L276-L282

This is not correct. If `pass_events` holds `EvdevFile`'s last strong reference, `Drop::drop` will be called with `opened_files` locked. Then we'll deadlock.

#### Fix potential atomic mode breaks

https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/evdev/mod.rs#L266

This can break the atomic mode because `InputHandlerClass::connect` is called under:
https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/comps/input/src/lib.rs#L126
but `register` will lock:
https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/device/char.rs#L40

#### Perform some remaining cleanups

 - Use early returns.
 - Restrict some visibility marks.
 - Remove the unnecessary `EvdevHandler` wrapper.

### DIFF
--- a/kernel/src/device/evdev/mod.rs
+++ b/kernel/src/device/evdev/mod.rs
@@ -42,7 +42,7 @@ static EVDEV_MINOR_COUNTER: AtomicU32 = AtomicU32::new(0);
 /// Global registry of evdev devices.
 static EVDEV_DEVICES: Mutex<BTreeMap<MinorId, Arc<EvdevDevice>>> = Mutex::new(BTreeMap::new());
 
-pub struct EvdevDevice {
+struct EvdevDevice {
     /// Input device associated with this evdev.
     device: Arc<dyn InputDevice>,
     /// List of opened evdev files with their producers.
@@ -75,7 +75,7 @@ impl Debug for EvdevDevice {
 impl EvdevDevice {
     pub(self) fn new(minor: u32, device: Arc<dyn InputDevice>) -> Self {
         let node_name = format!("event{}", minor);
-        let major = EVDEV_MAJOR.get().unwrap().get();
+        let major = MajorId::new(EVDEV_MAJOR_ID);
         let minor_id = MinorId::new(minor);
 
         Self {
@@ -108,7 +108,7 @@ impl EvdevDevice {
     }
 
     /// Distributes events to all opened evdev files.
-    pub(self) fn pass_events(&self, events: &[InputEvent]) {
+    fn pass_events(&self, events: &[InputEvent]) {
         // No need to disable IRQs because this method can only be called in the interrupt context.
         let mut opened_files = self.opened_files.lock();
 
@@ -162,7 +162,7 @@ impl EvdevDevice {
     }
 
     /// Creates a new opened evdev file for this evdev device.
-    pub(self) fn create_file(self: &Arc<Self>, buffer_size: usize) -> Result<Arc<EvdevFile>> {
+    fn create_file(self: &Arc<Self>, buffer_size: usize) -> Result<Arc<EvdevFile>> {
         // Create the evdev file and get the producer.
         let (file, producer) = EvdevFile::new(buffer_size, Arc::downgrade(self));
 
@@ -171,6 +171,12 @@ impl EvdevDevice {
 
         // Note that this can and should be a `Box` after fixing the char device subsystem.
         Ok(Arc::new(file))
+    }
+}
+
+impl InputHandler for EvdevDevice {
+    fn handle_events(&self, events: &[InputEvent]) {
+        self.pass_events(events);
     }
 }
 
@@ -199,7 +205,7 @@ impl CharDevice for EvdevDevice {
     }
 }
 
-/// Evdev handler class that creates device nodes for input devices.
+/// The evdev handler class that creates device nodes for input devices.
 #[derive(Debug)]
 struct EvdevHandlerClass;
 
@@ -216,18 +222,17 @@ impl InputHandlerClass for EvdevHandlerClass {
         let minor = EVDEV_MINOR_COUNTER.fetch_add(1, Ordering::Relaxed);
         let minor_id = MinorId::new(minor);
 
-        // Create evdev device.
+        // Create an evdev device.
         let evdev = Arc::new(EvdevDevice::new(minor, dev.clone()));
 
         // Register the device with the char device subsystem.
         register(evdev.clone()).map_err(|_| ConnectError::InternalError)?;
 
-        // Add to our registry for lookup during disconnect.
+        // Add to our registry for looking up during disconnection.
         EVDEV_DEVICES.lock().insert(minor_id, evdev.clone());
 
-        // Create handler instance for this device.
-        let handler = EvdevHandler::new(evdev);
-        Ok(Arc::new(handler))
+        // Return the device as a handler instance.
+        Ok(evdev as _)
     }
 
     fn disconnect(&self, dev: &Arc<dyn InputDevice>) {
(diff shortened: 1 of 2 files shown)


kernel/src/device/evdev/mod.rs
```diff
@@ -122,54 +122,47 @@ impl EvdevDevice {
...
                 // Read the current time according to the opened evdev file's clock type.
                 let time = file.read_clock();
-                let timed_event = EvdevEvent::from_event_and_time(event, time);
 
-                // Try to push event to the buffer.
-                if let Some(()) = producer.push(timed_event) {
-                    // Check if this is a SYN_REPORT event
-                    if self.is_syn_report_event(event) {
+                // When the buffer is full and a new event arrives, Linux drops all unconsumed
+                // events and queues a `SYN_DROPPED` event with the new one [1].
```
tatetian, line 122: The new implementation is better the current one. But still, it is inferior to Linux's strategy, I suppose. Because if we have to lose some events, then losing old inputs events is more tolerable than new events.

> Since we cannot remove events from the producer

In fact, we can--if we keep a reference to the consumer side of the ring buffer in `EvdevDevice`. To do this, we first wrap the consumer side of the ring buffer with an `Arc`

```rust
pub struct EvdevFile {
    consumer: Arc<Mutex<RbConsumer<EvdevEvent>>>,
```

Then, we change `EvdevDevice` to keep a reference to the consumer side.

```rust
pub struct EvdevDevice {
    opened_files: SpinLock<Vec<OpenFileEntry>>,
    ...
}

struct OpenFileEntry {
    weak_file: Weak<EvdevFile>,
    producer: RbProducer<EvdevEvent>,
    consumer: Arc<Mutex<RbConsumer<EvdevEvent>>>
}
```

This way, the `pass_events` method can use `Arc<Mutex<RbConsumer<EvdevEvent>>>` to evict old events and free space for new events.

One problem with this solution is that the `pass_events` method is called in the interrupt context, so it cannot use `Mutex`--at least directly. Changing this `Mutex` to `SpinLock` looks fine with `pass_events`, but it would cause some problems with the `process_events` method:

```rust
    fn process_events(&self, max_events: usize, writer: &mut VmWriter) -> Result<usize> {
        ...

        let mut consumer = self.consumer.lock();

        for _ in 0..max_events {
            let Some(event) = consumer.pop() else {
                break;
            };

            writer.write_val(&event)?;

            ...
        }
     }
```

We cannot call `Writer<'_, Fallible>::write_val` when holding a spinlock. So we have to rewrite `process_events` significantly to satisfy the constraint of no `write_val` invocation within a spinlock-protected critical section. This is doable. We can allocate a small kernel buffer so that event data are copied first to the kernel buffer while holding the spinlock. After releasing the spinlock, we then copy data from the kernel buffer to the target user buffer.

Another possible solution is to keep `Mutex` but only use `Mutex::try_lock` in the atomic-mode `pass_events` method. If the `pass_events` method successfully acquire the `Mutex` with `Mutex::try_lock`, then it can clear up the ring buffer. Failing to acquire the `Mutex` is very unlikely because this means that a concurrent consumer is executing, but haven't made the progress of evicting at least one events from the ring buffer. If this very unlike case occurs, we can fallback to your current strategy.

Both options should work. The first option is simpler.

kernel/comps/input/src/lib.rs
```diff
@@ -117,13 +117,13 @@ fn component_init() -> Result<(), ComponentInitError> {
 
 #[derive(Debug)]
 struct Component {
-    input_core: SpinLock<InputCore>,
+    input_core: Mutex<InputCore>,
```
tatetian, line 120: This is a good catch!

Note that `input_core` will be locked when an input device registers itself.

```rust
/// Registers an input device.
pub fn register_device(device: Arc<dyn InputDevice>) -> RegisteredInputDevice {
    let component = COMPONENT.get().unwrap();
    component.input_core.lock().register_device(device)
}
```

This means that the init function of an input device should be called in a task context. 

So an input component like `keyboard` (which will be renamed to `i8042_chip`) should have marked its init function as `init_component(kthread)`:

```rust
#[init_component(kthread)]
fn init() -> Result<(), ComponentInitError> {
    i8042_chip::init();
    Ok(())
}
```

In fact, I think the default mode for initializing a driver component should be `kthread`. Components like `virtio`, `net`, and `mlsdisk` should all have been fixed. Of course. This is out of the scope of this PR.

kernel/src/device/evdev/mod.rs
lrh2000: > The new implementation is better the current one. But still, it is inferior to Linux's strategy, I suppose. Because if we have to lose some events, then losing old inputs events is more tolerable than new events.

Yes. While it is inferior to Linux's strategy, it is at least correct. In particular, its behavior matches the correct semantics of the `SYN_DROPPED` event. The previous implementation is incorrect.

But another point is that Linux can clear the ring buffer in `O(1)` time complexity but we'll have to pop the events one by one. To resolve this problem, I have added a `RbConsumer::clear` method.

> Both options should work. The first option is simpler.

I would prefer not to remove the mutex in case it is needed for other synchronization purposes.
 - Ideally, the evdev event would not be consumed if `write_val` fails. Therefore, `write_val` must be called with a mutex held. However, [even Linux does not address this issue](https://elixir.bootlin.com/linux/v6.17.9/source/drivers/input/evdev.c#L586-L587), so I'm not addressing it either.
 - But Linux does [hold a mutex when doing ioctl](https://elixir.bootlin.com/linux/v6.17.9/source/drivers/input/evdev.c#L1261), so there may be some other reasons to justify the mutex.

---

See the diff here: https://github.com/asterinas/asterinas/compare/329a7541ef8e458d9e6a4a55fc4d078e6335d8a4..ed3416a2530ce9fc1a2bfad9f87786f644584853

kernel/comps/input/src/lib.rs
lrh2000: I agree, but it might not cause any problems since `Mutex::lock` cannot theoretically fail before the first kernel thread is spawned... (Unless we try to lock the same mutex twice and cause a deadlock?)

tatetian: > it might not cause any problems since `Mutex::lock` cannot theoretically fail before the first kernel thread is spawned...

Sure. So this is not an urgent matter.

tatetian:
@lrh2000 Since evdev is not tested by the CI, I want to confirm that the improved implementation has been tested manually, right?

tatetian:
@Cautreoxit Need your blessing before merging this PR. And your pending evdev ioctl PR should be rebased on this PR. Thanks!

Cautreoxit:
Thanks for your improments on evdev! I reviewed these changes and did not see any bugs. I would use evtest to verify this PR later.

tatetian:
> nd did not see any bugs. I would use evtest to verify this PR later.

Ok. Then I will merge this PR as it is now and leave further testing to you.

lrh2000:
> Thanks for your improments on evdev! I reviewed these changes and did not see any bugs. I would use evtest to verify this PR later.

Thank you!

PR #2617, opened 2025-11-26
Enable some security configurations in TDX environment

Hsy-Intel:
This PR enable three security configurations in TDX environment:

1. Reduce #VE. When `VE_REDUCTION` is enabled, the TDX module obtains the static value of MSR/CPUID from the VMM when the TD starts and it significantly reduce the occurrence of #VE.
2. Disable SEPT #VE. Disable EPT violation conversion to `#VE(PENDING)` on guest TD access of PENDING pages. There is a known attack path could be leveraged for the MMIO emulation.
3. Enable notification for zero step attack detection in TDX environment.


tatetian (approved):

PR #2618, opened 2025-11-26
Add TDX hardware randomness requirement note

Hsy-Intel:
In Intel TDX environment, always use the `RDRAND`/`RDSEED` instructions, which are not controlled by the host/VMM, as the entropy source.

Ref: [Randomness inside TDX guest](https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest)

### DIFF
--- a/kernel/src/util/random.rs
+++ b/kernel/src/util/random.rs
@@ -27,6 +27,10 @@ fn get_random_seed() -> <StdRng as SeedableRng>::Seed {
 
     let mut seed = <StdRng as SeedableRng>::Seed::default();
 
+    // Notes for future refactorings: If hardware randomness cannot be generated (i.e., if
+    // `read_random` fails), we can usually continue with pseudorandomness. However, we should stop
+    // if we are TD guests. For more details, see
+    // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
     let mut chunks = seed.as_mut().chunks_exact_mut(size_of::<u64>());
     for chunk in chunks.by_ref() {
         let src = read_random().expect("`read_random` failed").to_ne_bytes();


lrh2000:
```patch
diff --git a/ostd/src/arch/x86/mod.rs b/ostd/src/arch/x86/mod.rs
index 300b661ac..4a15ee053 100644
--- a/ostd/src/arch/x86/mod.rs
+++ b/ostd/src/arch/x86/mod.rs
@@ -120,16 +120,24 @@ pub fn read_random() -> Option<u64> {
 
     use cpu::extension::{has_extensions, IsaExtensions};
 
-    if !has_extensions(IsaExtensions::RDRAND) {
-        return None;
-    }
-
-    // Recommendation from "Intel(R) Digital Random Number Generator (DRNG) Software
-    // Implementation Guide" - Section 5.2.1 and "Intel(R) 64 and IA-32 Architectures
-    // Software Developer's Manual" - Volume 1 - Section 7.3.17.1.
-    const RETRY_LIMIT: usize = 10;
+    let has_rdrand = has_extensions(IsaExtensions::RDRAND);
+    let num_retries = if_tdx_enabled!({
+        assert!(
+            has_rdrand,
+            "RDRAND instruction not available in TDX environment - hardware random number generation is mandatory"
+        );
+        u64::MAX
+    } else {
+        if !has_rdrand {
+            return None;
+        }
+        // Recommendation from "Intel(R) Digital Random Number Generator (DRNG) Software
+        // Implementation Guide" - Section 5.2.1 and "Intel(R) 64 and IA-32 Architectures
+        // Software Developer's Manual" - Volume 1 - Section 7.3.17.1.
+        10
+    });
 
-    for _ in 0..RETRY_LIMIT {
+    for _ in 0..num_retries {
         let mut val = 0;
         let generated = unsafe { _rdrand64_step(&mut val) };
         if generated == 1 {
```

Will this satisfy your needs? I think the code looks better, and, in practice, it is equivalent unless you're concerned that `u64::MAX` is not infinite. However, I'm not sure if humans will still exist after `u64::MAX` instructions.

---

Also note that, in fact, the boot will already fail even without this patch if the hardware randomness cannot be generated due to:
https://github.com/asterinas/asterinas/blob/70eda539dfade526c28d44e6cb60c056c56cc06c/kernel/src/util/random.rs#L32

According to the [Linux patch](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/arch/x86/coco/core.c?h=v6.9-rc5&id=99485c4c026f024e7cb82da84c7951dbe3deb584), is it more intended to move the panicking to `kernel/src/util/random.rs`? But it already panics anyway...
```patch
+	for (i = 0; i < ARRAY_SIZE(rng_seed); i += longs) {
+		longs = arch_get_random_longs(&rng_seed[i], ARRAY_SIZE(rng_seed) - i);
+
+		/*
+		 * A zero return value means that the guest doesn't have RDRAND
+		 * or the CPU is physically broken, and in both cases that
+		 * means most crypto inside of the CoCo instance will be
+		 * broken, defeating the purpose of CoCo in the first place. So
+		 * just panic here because it's absolutely unsafe to continue
+		 * executing.
+		 */
+		if (longs == 0)
+			panic("RDRAND is defective.");
+	}
+	add_device_randomness(rng_seed, sizeof(rng_seed));
+	memzero_explicit(rng_seed, sizeof(rng_seed));
```

I think `arch_get_random_longs` is our `ostd::arch::read_random`, but `panic("RDRAND is defective.");` lives outside it.

lrh2000:
So do you have a response to the second question?


> According to the [Linux patch](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/arch/x86/coco/core.c?h=v6.9-rc5&id=99485c4c026f024e7cb82da84c7951dbe3deb584), is it more intended to move the panicking to `kernel/src/util/random.rs`? But it already panics anyway...
> 
> ```diff
> +	for (i = 0; i < ARRAY_SIZE(rng_seed); i += longs) {
> +		longs = arch_get_random_longs(&rng_seed[i], ARRAY_SIZE(rng_seed) - i);
> +
> +		/*
> +		 * A zero return value means that the guest doesn't have RDRAND
> +		 * or the CPU is physically broken, and in both cases that
> +		 * means most crypto inside of the CoCo instance will be
> +		 * broken, defeating the purpose of CoCo in the first place. So
> +		 * just panic here because it's absolutely unsafe to continue
> +		 * executing.
> +		 */
> +		if (longs == 0)
> +			panic("RDRAND is defective.");
> +	}
> +	add_device_randomness(rng_seed, sizeof(rng_seed));
> +	memzero_explicit(rng_seed, sizeof(rng_seed));
> ```
> 
> I think `arch_get_random_longs` is our `ostd::arch::read_random`, but `panic("RDRAND is defective.");` lives outside it.

e.g.,
```patch
diff --git a/kernel/src/util/random.rs b/kernel/src/util/random.rs
index e829e165d..97edd6f40 100644
--- a/kernel/src/util/random.rs
+++ b/kernel/src/util/random.rs
@@ -27,6 +27,10 @@ fn get_random_seed() -> <StdRng as SeedableRng>::Seed {
 
     let mut seed = <StdRng as SeedableRng>::Seed::default();
 
+    // Notes for future refactorings: If hardware randomness cannot be generated (i.e., if
+    // `read_random` fails), we can usually continue with pseudorandomness. However, we should stop
+    // if we are TD guests. For more details, see
+    // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
     let mut chunks = seed.as_mut().chunks_exact_mut(size_of::<u64>());
     for chunk in chunks.by_ref() {
         let src = read_random().expect("`read_random` failed").to_ne_bytes();
```

If we do this, I don't think we need to fix `read_random`. We may want to investigate further to see if it aligns with the Linux implementation, but it seems to align with https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest:
>> The commit [x86/coco: Require seeding RNG with RDRAND on CoCo systems](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/arch/x86/coco/core.c?h=v6.9-rc5&id=99485c4c026f024e7cb82da84c7951dbe3deb584) ensures that a TDX guest cannot boot unless 256 bits of RDRAND output is mixed into the entropy pool early during the boot process.

Hsy-Intel:
I understand now; I misunderstood your first comment. It seems the `read_random()` function indeed doesn't need modification.

lrh2000 (approved):

PR #2621, opened 2025-11-26
Support NixOS distro and add CI to test

StevenJiang1110:
This PR introduces fundamental support for installing NixOS on Asterinas.

~The initial commit includes several temporary fixes specifically for the ext2 file system. These fixes are crucial for enabling NixOS installation and the actual fix is included in #2624.~

~Furthermore, this PR depends on #2619, which ensures the inclusion of essential tools within our development container.~

The scripts and Nix configurations are largely derived from @cqs21's ISO image support, found in [this branch](https://github.com/cqs21/asterinas/tree/dev-iso-image).

# How to use

Before building NixOS, please ensure that the tools included in #2619 are installed, and that the `/dev` directory is mapped into the container using the `-v` option.

Building NixOS can then be achieved with a single command:

```bash
make build NIXOS=1
```

To run it, use:
```bash
make run NIXOS=1
```

Some Makefile options, such as `LOG_LEVEL` and `ENABLE_KVM`, are compatible with `NIXOS=1`. However, others like `ARCH`, `AUTO_TEST`, and `BOOT_PROTOCOL` (those typically used with OSDK) are not currently compatible with NIXOS=1. Additionally, multi-architecture support is not available, as the binaries are compiled directly on the host and thus are specific to the host's architecture.

# The distro directory

Currently, the distro directory has the following structure, with the build scripts located under `tools/nixos`:

```plain
distro
|-- configuration.nix
|-- modules
|   `-- core.nix
|-- overlays
|   |-- default.nix
|   `-- hello-asterinas
|       `-- default.nix
`-- splash.png
```

- configuration.nix: This is the primary configuration file for NixOS.
- modules: Each file contains a part of the system configuration and can be imported by `configuration.nix`.
- overlays: This directory contains all modified or added applications.
- splash.png: This image serves as the boot-up logo.

# The build process

Largely follow the [official manual](https://nixos.org/manual/nixos/stable/#sec-installation-manual).

# Unresolved questions

What information should we present to users in `/etc/os-release`? The exact specification is available [here](https://man7.org/linux/man-pages/man5/os-release.5.html). Currently this PR uses the default NixOS values, with the distribution name set to "Asterinas NixOS". The current contents are shown below.

```plain
ANSI_COLOR="0;38;2;126;186;228"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="25.05.813095.1c8ba8d3f763"
CPE_NAME="cpe:/o:nixos:nixos:25.05"
DEFAULT_HOSTNAME=nixos
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
ID_LIKE=""
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME="Asterinas NixOS"
PRETTY_NAME="Asterinas NixOS 25.05 (Warbler)"
SUPPORT_END="2025-12-31"
SUPPORT_URL="https://nixos.org/community.html"
VARIANT=""
VARIANT_ID=""
VENDOR_NAME=NixOS
VENDOR_URL="https://nixos.org/"
VERSION="25.05 (Warbler)"
VERSION_CODENAME=warbler
VERSION_ID="25.05"
```


tools/nixos/install_asterinas.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+trap cleanup EXIT INT TERM ERR
+
+cp ${DISTRO_DIR}/configuration.nix ${BUILD_DIR}/etc/nixos
+cp -r ${DISTRO_DIR}/modules ${BUILD_DIR}/etc/nixos
+cp -r ${DISTRO_DIR}/overlays ${BUILD_DIR}/etc/nixos
+
+nixos-install --root ${BUILD_DIR} --no-root-passwd
+
+echo "Congratulations! NixOS has been installed successfully!"
```
tatetian: NixOS --> Asterinas NixOS

distro/modules/core.nix
```diff
@@ -0,0 +1,65 @@
+{ config, lib, pkgs, ... }:
+let
+  splash = builtins.path { path = builtins.getEnv "NIXOS_SPLASH_PATH"; };
```
tatetian: This `getEnv` function is impure. Using impure functions make the Nix expression indeterministic and unreproducible. I won't block this PR for this problem. But I think this is a point that should be fixed in the future.

distro/configuration.nix
```diff
@@ -0,0 +1,38 @@
+# Edit this configuration file to define what should be installed on
+# your system. Help is available in the configuration.nix(5) man page, on
+# https://search.nixos.org/options and in the NixOS manual (`nixos-help`).
+
+# The content of this file largely mirrors or aligns with the output
+# generated by `nixos-generate-config`.
+
+{ config, lib, pkgs, ... }: {
+  # Include other configuration files.
+  imports = [ ./overlays/default.nix ./modules/core.nix ];
```
tatetian: Rewrite the comments:

```nix
  # Do not change these imports, which describe system-wide settings for AsterNixOS
  imports = [ ./overlays/default.nix ./modules/core.nix ];
```

---

What are "modules"? Any more modules that you plan to add?

distro/modules/core.nix
StevenJiang1110: Thanks for pointing this out! This is currently a temporary solution to allow nixos-install and ISO image builds to use different splash/kernel paths. We haven't found an elegant way to pass command-line options yet, which is a real problem. We'll need to figure out a proper solution for this later so `nixos-rebuild` conld work as expected.

distro/configuration.nix
StevenJiang1110: > What are "modules"? Any more modules that you plan to add?

My thought is that every Nix file located under the `modules` directory should be considered a part of `configuration.nix`. Then, `configuration.nix` can import each of these individual files. For instance, configurations for `containers.nix` or `desktop_env.nix` could reside within the `modules` directory, while `configuration.nix` would simply contain an enable/disable switch for them. This approach ensures that the content of `configuration.nix` remains as modular as possible, preventing it from becoming a single monolithic file.

```plain
distro/modules
|-- containers.nix
|-- core.nix
`-- desktop_env.nix
```

For example, we could define all container-related configurations within `containers.nix`.

```nix
# distro/modules/containers.conf
{ config, lib, pkgs, ... }: {
  virtualisation.containers.storage.settings = {
    storage = {
      driver = "vfs";
      runroot = "/run/containers/storage";
      graphroot = "/var/lib/containers/storage";
    };
  };

  virtualisation.containers.policy = {
    default = [{ type = "insecureAcceptAnything"; }];
  };

  virtualisation.containers.containersConf.settings = {
    containers = {
      cgroupns = "host";
      cgroups = "enabled";
      default_sysctls = [ ];
      ipcns = "host";
      keyring = false;
      netns = "host";
      pidns = "host";
      privileged = true;
      seccomp_profile = "unconfined";
      userns = "host";
    };

    secrets = { };
    network = { };
    engine = {
      cgroup_manager = "cgroupfs";
      events_logger = "none";
      no_pivot_root = true;
      runtime = "runc";
    };
  };
}
```

Then, in `configuration.nix`, we would only keep a single enable/disable switch.

```nix
# distro/configuration.nix
{ config, lib, pkgs, ... }: {
   # Do not change these imports, which describe system-wide settings for AsterNixOS
  imports = [ ./overlays/default.nix ./modules/core.nix ./modules/containers.nix ];

  # Uncomment the following two lines to enable podman
  # virtualisation.podman.enable = true;
  # virtualisation.containers.enable = true;
}

File: Makefile [hunk 4]
```diff
@@ -54,6 +54,14 @@ VHOST ?= off
 DNS_SERVER ?= none
 # End of network settings
 
+# NixOS settings
+NIXOS ?= 0
+NIXOS_DISK_SIZE_IN_MB ?= 8196
+NIXOS_DISABLE_SYSTEMD ?= true
```
TinaZhangZW, line 60: Do we want to disable system by default? If so, why?

If NIXOS_DISABLE_SYSTEMD is true, how do we plan to run a desktop environment such as Xfce? Will a DE/session be started automatically (for example via a non-systemd display/session manager or wrapper init), or do we expect users to start it manually (e.g., by running a script)?

distro/configuration.nix
tatetian: Ok. I got your point. But don't you think that adding these Nix files to the `imports` list exposes too much implementation details? This list is likely to grow. And how we organize the Nix expression into modules should not be a concern to the user.

I think a better approach is to import just one file:

```nix
# distro/configuration.nix
{ config, lib, pkgs, ... }: {
   # Do not change these imports, which describe system-wide settings for AsterNixOS
  imports = [ ./aster_configuration.nix ];

  # Uncomment the following two lines to enable podman
  # virtualisation.podman.enable = true;
  # virtualisation.containers.enable = true;
}
```

In the `aster_configuration.nix` file, you can merge `./overlays/default.nix`, `./modules/core.nix`, `./modules/containers.nix`, and more into a single file. Because `aster_configuration.nix` is not supposed to be read by AsterNixOS users, you can add comments to explain the concepts of "modules" and "overlays".

Makefile
StevenJiang1110: > Do we want to disable system by default? If so, why?

Currently, systemd cannot be run on the main branch. It will trigger a kernel panic. Therefore, this PR disables systemd by default. Once systemd support is merged, we can enable systemd by default. In the long term, the option to disable systemd is primarily intended for running simple test cases, similar to its usage in the GitHub Actions workflow for this PR.

> If NIXOS_DISABLE_SYSTEMD is true, how do we plan to run a desktop environment such as Xfce? Will a DE/session be started automatically (for example via a non-systemd display/session manager or wrapper init), or do we expect users to start it manually (e.g., by running a script)?

I think we may not want to support XFCE without systemd? If so, then adding the following logic to the `Makefile` should suffice:

```Makefile
ifeq ($(XFCE), true)
    NIXOS_DISABLE_SYSTEMD := false
endif
```

if a user attempts to run a command like `make run NIXOS=1 XFCE=true NIXOS_DISABLE_SYSTEMD=true, the NIXOS_DISABLE_SYSTEMD=true` option will be effectively overridden and ignored, ensuring that systemd is enabled when XFCE is requested.

TinaZhangZW: As discussed, the plan remains unchanged: Xfce will automatically start when systemd boots to the graphical target. Thank you for the clarification.

tatetian (approved):
LGTM. Thanks for contribution!

kuraga:
Why did you call Nix/NixPkg as NixOS? Only the package manager has been used, isn't it?

tatetian:
> Why did you call Nix/NixPkg as NixOS? Only the package manager has been used, isn't it?

We [use `nixos-install`](https://github.com/asterinas/asterinas/blob/77f0b0e746ce2df0066b007fb050fc5be0e86e47/distro/aster_nixos_installer/templates/install_nixos.sh#L135) to install an entire NixOS system. This is by defintion a NixOS, not only the Nix package manager.

PR #2624, opened 2025-11-26
Fix ext2 directory entry iteration

cqs21:
This PR fixes some issues in ext2 directory entry iteration logic:
- skip logic for unused entries (ino=0) and unknown file types
- uses record_len=0 as the termination condition

### DIFF
--- a/kernel/src/fs/ext2/dir.rs
+++ b/kernel/src/fs/ext2/dir.rs
@@ -58,7 +58,7 @@ impl DirEntry {
 
     /// Returns the inode type of the entry.
     pub fn type_(&self) -> InodeType {
-        InodeType::from(DirEntryFileType::try_from(self.header.inode_type).unwrap())
+        InodeType::from(DirEntryFileType::from(self.header.inode_type))
     }
 
     /// Returns the distance to the next entry.
@@ -108,7 +108,7 @@ impl DirEntryHeader {
 
 /// The type indicator in the `DirEntry`.
 #[repr(u8)]
-#[derive(Copy, Clone, Debug, Eq, PartialEq, TryFromInt)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
 enum DirEntryFileType {
     Unknown = 0,
     File = 1,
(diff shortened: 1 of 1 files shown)


kernel/comps/block/src/request_queue.rs
```diff
@@ -224,7 +224,7 @@ impl From<SubmittedBio> for BioRequest {
     fn from(bio: SubmittedBio) -> Self {
         let mut sid_range = bio.sid_range().clone();
         sid_range.start = sid_range.start + bio.sid_offset();
-        sid_range.end = sid_range.start + bio.sid_offset();
+        sid_range.end = sid_range.end + bio.sid_offset();
```
tatetian, line 227: Separate the bugfixes into their own commits. Describe the bug in the commit title and message

kernel/src/fs/ext2/dir.rs
```diff
@@ -248,11 +265,25 @@ impl<'a> DirEntryReader<'a> {
...
+                || DirEntryFileType::from(header.inode_type) == DirEntryFileType::Unknown
+            {
+                self.offset += header.record_len as usize;
+                continue;
+            }
+
+            break header;
+        };
 
-        let header = self.read_header()?;
```
tatetian, line 255: To improve readability, extract all logic that updates the cursor (`self.offset`) into a single method. The draft code is provided:

```rust
impl DirEntryIter<'_> {
    fn read_next_dir_entry(&mut self) -> Option<DirEntryItem> {
        let max_size = self.page_cache.pages().size();
        if self.offset >= max_size {
            return None;
        };

        let header = self.page_cache
            .pages()
            .read_val::<DirEntryHeader>(self.offset)
            .ok()?;

        if header.record_len == 0 {
            return None;
        }

        let next_dir_entry = DirEntryItem {
            header,
            offset: self.offset,
        };

        self.offset += next_dir_entry.header.record_len;
        Some(next_dir_entry)
    }
}

impl Iterator for DirEntryIter<'_> {
    type Item = DirEntryItem;

    fn next(&mut self) -> Option<Self::Item> {
        loop {
            let next_item = self.read_next_dir_entry()?;

            // Skip an entry if it is not used (ino == 0, see [1]) or its file type is unknown
            //
            // [1]: <https://www.nongnu.org/ext2-doc/ext2.html#linked-directory-entry-structure>.
            if next_item.header.ino == 0
                || DirEntryFileType::from(header.inode_type) == DirEntryFileType::Unknown
            {
                continue;
            }

            return Some(next_item);
        }
    }
}
```

tatetian (commented):
Good job on identifying these bugs!

PR #2351, opened 2025-08-18
Refactor DMA APIs

junyang-zh:
This PR is mainly for working around #2269 

Meanwhile, it makes progress towards fixing this:

```rust
pub trait HasVmReaderWriter {
    /// A marker type that denotes the return types of [`Self::reader`] and [`Self::writer`].
    ///
    /// This can be either [`VmReaderWriterIdentity`] or [`VmReaderWriterResult`].
    //
    // TODO: This exists because `DmaStream` and related types track the DMA direction at runtime.
    // The goal is to achieve this at compile time, which would eliminate the need for
    // `VmReaderWriterTypes`. See the discussion at
    // <https://github.com/asterinas/asterinas/pull/2289#discussion_r2261801694>.
    type Types: VmReaderWriterTypes;
```

cc @lrh2000 

<details><summary>Here are earlier changes that was merged separately</summary>
<p>

# Modification of each commit

## Commit 1: Unify address and size APIs for memory objects

Currently in OSTD, while we have `HasPaddr`/`HasDaddr` traits, the types implementing them also have similar duplicated methods, such as `start_paddr`, `paddr`, etc. This commit **forces all users to use the trait method, not the type method**. Therefore, the usages can be unified. Furthermore, the traits enable us to perform advanced code reuse, such as the slicing work in the next commit.

## Commit 2: Unify memory object slicing

In the previous `DmaStream` API, there was `DmaStreamSlice`. But no similar things were provided for `DmaCoherent`. It also reminds me of the `SegmentSlice`, which we had used (but are not using now). Such slicing implementation 1. should not be implemented everywhere, and 2. should not be in OSTD. This commit moves it out of OSTD and makes it generic, replacing the original `SegmentSlice` and `DmaStreamSlice`.

## Commit 3: Refactor implicit `Arc` APIs for DMA

It prevents `DmaStream`/`DmaCoherent` from implicitly using an `Arc`. The benefits are apparent. We can avoid allocation and cloning in some places. Also, the developers will become more aware of such sharing, which is detrimental to scalability.

## Commit 5

EDIT: Added a commit to fix the virtio driver that delivers physical addresses to the devices instead of device addresses. It worked in the past because the daddr to paddr mapping was identical.

</p>
</details> 

This PR, in brief:
 - Unify the implementations of `DmaCoherent` and `DmaStream`. It becomes clear why these variants exist (actually, I get to know why after I refactored it...).
[...]

PR #2629, opened 2025-11-27
Pass virtual address to TDX `unprotect_gpa_range`

taosue:
Since `IoMem` ranges are not in linear mapping and should be shared, this PR passes actual mapped virtual addresses to `unprotect_gpa_range`, instead of assuming linear mapping.

Key changes:
- Add `gva` parameter to `unprotect_gpa_range()` to accept mapped virtual addresses
- Remove boot page table operations as they're no longer used when invoking `protect_gpa_range` and `unprotect_gpa_range`
- Update all callers (IoApic, DMA coherent, DMA stream, IO memory) to pass 
  the correct virtual addresses
- For linear-mapped memory (USegment), use `paddr_to_vaddr()`
- For IO memory, use the actual `KVirtArea` mapping

### DIFF
--- a/ostd/src/arch/x86/tdx_guest.rs
+++ b/ostd/src/arch/x86/tdx_guest.rs
@@ -1,76 +1,40 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use log::warn;
 use tdx_guest::{tdcall::accept_page, tdvmcall::map_gpa, TdxTrapFrame};
 
 use super::trap::TrapFrame;
-use crate::{
-    mm::{
-        kspace::KERNEL_PAGE_TABLE,
-        paddr_to_vaddr,
-        page_prop::{PageProperty, PrivilegedPageFlags as PrivFlags},
-        page_table::boot_pt,
-        PAGE_SIZE,
-    },
-    prelude::Paddr,
-};
+use crate::{mm::PAGE_SIZE, prelude::Paddr};
 
 const SHARED_BIT: u8 = 51;
 const SHARED_MASK: u64 = 1u64 << SHARED_BIT;
 
 #[derive(Debug)]
 pub enum PageConvertError {
-    PageTable,
+    InvalidAddress,
     TdCall,
     TdVmcall,
 }
 
 /// Converts physical pages to Intel TDX shared pages.
 ///
-/// This function sets the [`PrivFlags::SHARED`] bit in the linear mapping of physical pages. Then,
-/// it invokes the [`map_gpa`] TDVMCALL to convert those pages into Intel TDX shared pages. Due to
+/// Invokes the [`map_gpa`] TDVMCALL to convert those pages into Intel TDX shared pages. Due to
 /// the conversion, any existing data on the pages will be erased.
 ///
 /// # Safety
 ///
 /// The caller must ensure that:
-///  - The provided physical address is page aligned.
 ///  - The provided physical address range is in bounds, i.e., it should fall within the maximum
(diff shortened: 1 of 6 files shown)


ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -54,21 +58,11 @@ pub unsafe fn unprotect_gpa_range(gpa: Paddr, page_num: usize) -> Result<(), Pag
...
-    });
 
     // Protect the page in the kernel page table.
     let pt = KERNEL_PAGE_TABLE.get().unwrap();
-    let vaddr = paddr_to_vaddr(gpa);
-    // SAFETY: The caller ensures that the address range exists in the linear mapping and can be
-    // mapped as shared pages.
-    unsafe { pt.protect_flush_tlb(&(vaddr..vaddr + page_num * PAGE_SIZE), protect_op) }
+    // SAFETY: The caller ensures that the address range can be mapped as shared pages.
+    unsafe { pt.protect_flush_tlb(&(gva..gva + page_num * PAGE_SIZE), protect_op) }
```
tatetian: I think both `unprotect_gpa_range` and `protect_gpa_range` methods should remove the call to `protect_flush_tlb`. There are two users of the two methods: `IoMem` and `DmaXxx`. In both cases, before calling `unprotect_gpa_range`, they manipulate the page table, creating new mappings or updating protection bits. Either way, they already have the chance to add the SHARED bit. Calling `protect_flush_tlb` here, simply to set the SHARED bit, doubles the work.

To fix this inefficiency, remove the call to `protect_flush_tlb` from the two functions and let their callers to set the SHARE bit properly when they manipulate the page table.

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -100,6 +94,14 @@ pub unsafe fn protect_gpa_range(gpa: Paddr, page_num: usize) -> Result<(), PageC
         warn!("Misaligned address: {:x}", gpa);
     }
 
+    if gpa + page_num * PAGE_SIZE > crate::mm::frame::max_paddr() {
+        warn!(
+            "GPA range [{:#x}, {:#x}) is not within the linear mapping",
```
tatetian: I suppose that you want to print the warning in case that this function is misused by `IoMem` in the future when it implements the `Drop` trait. But why not make the function right so that it can't be misused? The protect function is symmetric to the unprotect function. So I suppose the protect function should be refactored in the same way as the unprotect function.

tatetian:
@taosue By the way, did you test the changes in TDX VMs?

taosue:
> @taosue By the way, did you test the changes in TDX VMs?

Yes, I tested it using `make run INTEL_TDX=1 AUTO_TEST=test` and `make run INTEL_TDX=1 AUTO_TEST=syscall` in TDX VMs, all passed.

ostd/src/arch/x86/tdx_guest.rs
junyang-zh: I agree. But currently implementing it for DMA without #2351 would look like a wrong-to-wrong fix, since it would force the DMA implementation to protect the linear mapping when it does not have to.

tatetian: > I agree. But currently implementing it for DMA without #2351 would look like a wrong-to-wrong fix, since it would force the DMA implementation to protect the linear mapping when it does not have to.

Thanks for the insight. I will need to squeeze the time to review #2351 ...

taosue:
Close it since #2351 exactly covers what this PR did.

PR #2632, opened 2025-11-28
Install qemu with NUMA support in container

vvvvsv:
This PR installs a QEMU with NUMA binding support in the development container, enabling VM memory to be bound to a specific NUMA node on the host. This PR is intended to support #2633.

After these two PRs are approved, I’ll bump the container version.

Additionally, the first commit performs some cleanup in the frame allocator.


vvvvsv:
@boterinas rerun

tatetian:
Need @junyang-zh 's approval for the changes to the frame allocator.

PR #2633, opened 2025-11-28
Introduce NUMA-Aware Frame Allocator

vvvvsv:
This PR introduces a NUMA-aware frame allocator that follows a first-touch allocation strategy. Under this approach, physical pages are allocated from the NUMA node of the CPU that first accesses the memory.

---

# Non-NUMA-aware frame allocator

Our old frame allocator follows the following strategy:

 - Data Structure: Each CPU has a local cache and a local pool. All other memory resides in the global pool.

 - `alloc`: Allocation first tries the CPU’s local cache. If that fails, it tries the local pool, and if that also fails, it falls back to the global pool. After allocation, balance the sizes of the local pool and global pool.

 - `dealloc`: Deallocation first tries the CPU’s local cache. If that fails, it tries the local pool, and if that also fails, it falls back to the global pool. After deallocation, balance the sizes of the local pool and global pool.

 - `add_free_memory` (at boot stage): All free memory is added to the global pool.

# NUMA-aware frame allocator

The NUMA-aware frame allocator in this PR follows the following strategy:

 - Data Structure: Each CPU has a local cache and a local pool. Each NUMA node has its own pool.

 - `alloc`: Allocation first tries the CPU’s local cache. If that fails, it tries the local pool, and if that also fails, it falls back to the pool of the NUMA node that the current CPU belongs to. After allocation, balance the sizes of the local pool and node pool.

 - `dealloc`: 
   - If the current CPU and the memory deallocated belong to the same NUMA node, deallocation first tries the local cache. If that fails, it tries the local pool, and if that also fails, it falls back to the NUMA node’s pool. After deallocation, balance the sizes of the local pool and node pool.
   -  If the current CPU and the memory deallocated do not belong to the same NUMA node, the memory is directly deallocated to the pool of the NUMA node it belongs to.

 - `add_free_memory` (at boot stage): Free memory is added to the corresponding NUMA node’s pool based on its physical address.

# Performance

In short, the results are evaluated on a 2-socket AMD EPYC 9965 machine, each socket has a 192-core CPU and 512GB DRAM, showing a **7.88%** speedup.

For details, you can refer to [this branch](https://github.com/vvvvsv/asterinas/tree/parse_numa) in my own repository, and read the file `test_numa_readme.md`.

---

NOTE: @jellllly420 submitted two PRs to the `acpi` crate to add support for [SRAT](https://github.com/rust-osdev/acpi/pull/241) and [SLIT](https://github.com/rust-osdev/acpi/pull/243). This PR needs them to parse the NUMA topology at boot stage. However, it seems unlikely that these PRs will be merged in the near term, as the maintainer hasn’t responded for a long time.

Therefore, I directly used the code from those two PRs in the commit "Parse NUMA topology at boot stage" and co-authored it with @jellllly420. This approach may require further discussion. Perhaps we need to consider maintaining our own `acpi` crate?

UPDATE: Will go after https://github.com/rust-osdev/acpi/pull/264.

---

~NOTE: After https://github.com/asterinas/asterinas/pull/2643 is merged and the container version is bumped, I will update tools/qemu_args.sh according to [the comments](https://github.com/asterinas/asterinas/blob/1dd9a33f9823019c86acbecb90164bbd9a7da889/tools/qemu_args.sh#L126-L131) to fully support NUMA-aware. For now, NUMA binding is disabled so that the CI tests can pass.~

---

TODO:
 - [ ] Add dummy riscv and loongarch NUMA topology.
 - [ ] Add fall back lists.
 - [ ] Add syscalls `mbind`, `set_mempolicy`, `get_mempolicy`, `move_pages`.
 - [ ] Implement policies other than first-touch.

### DIFF
--- a/osdk/deps/frame-allocator/src/pools/mod.rs
+++ b/osdk/deps/frame-allocator/src/pools/mod.rs
@@ -5,14 +5,17 @@ mod balancing;
 use core::{
     alloc::Layout,
     cell::RefCell,
-    ops::DerefMut,
+    ops::{DerefMut, Drop, Range},
     sync::atomic::{AtomicUsize, Ordering},
 };
 
 use ostd::{
+    cpu::{CpuId, PinCurrentCpu},
     cpu_local,
     irq::DisabledLocalIrqGuard,
+    leader_cpu_local,
     mm::Paddr,
+    numa::{leader_cpu_of, leader_cpu_of_node, memory_ranges, num_cpus_in_node, NodeId},
     sync::{LocalIrqDisabled, SpinLock, SpinLockGuard},
 };
 
@@ -20,14 +23,16 @@ use crate::chunk::{greater_order_of, lesser_order_of, size_of_order, split_to_ch
 
 use super::set::BuddySet;
 
-/// The global free buddies.
-static GLOBAL_POOL: SpinLock<BuddySet<MAX_BUDDY_ORDER>, LocalIrqDisabled> =
-    SpinLock::new(BuddySet::new_empty());
-/// A snapshot of the total size of the global free buddies, not precise.
-static GLOBAL_POOL_SIZE: AtomicUsize = AtomicUsize::new(0);
+// The NUMA node pools.
+leader_cpu_local! {
+    /// Free buddies in the NUMA node of the leader CPU.
+    static NODE_POOL: SpinLock<BuddySet<MAX_BUDDY_ORDER>, LocalIrqDisabled> = SpinLock::new(BuddySet::new_empty());
+    /// A snapshot of the total size of the free buddies in the NUMA node of the leader CPU, not precise.
+    static NODE_POOL_SIZE: AtomicUsize = AtomicUsize::new(0);
+}
 
-// CPU-local free buddies.
 cpu_local! {
+    /// CPU-local free buddies.
     static LOCAL_POOL: RefCell<BuddySet<MAX_LOCAL_BUDDY_ORDER>> = RefCell::new(BuddySet::new_empty());
 }
 
@@ -51,10 +56,11 @@ const MAX_BUDDY_ORDER: BuddyOrder = 32;
 /// chunks.
 const MAX_LOCAL_BUDDY_ORDER: BuddyOrder = 18;
 
+/// Allocates frames from the pools of the local NUMA node.
 pub(super) fn alloc(guard: &DisabledLocalIrqGuard, layout: Layout) -> Option<Paddr> {
     let local_pool_cell = LOCAL_POOL.get_with(guard);
     let mut local_pool = local_pool_cell.borrow_mut();
-    let mut global_pool = OnDemandGlobalLock::new();
+    let mut node_pool = OnDemandNodeLock::new(leader_cpu_of(guard.current_cpu()));
 
     let size_order = greater_order_of(layout.size());
     let align_order = greater_order_of(layout.align());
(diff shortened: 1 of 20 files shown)


ostd/src/numa.rs
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! NUMA(Non Uniform Memory Access) support.
```
junyang-zh, line 3:
```suggestion
//! NUMA (Non Uniform Memory Access) support.
```

ostd/src/numa.rs
```diff
@@ -52,8 +55,87 @@ pub fn node_id_of_cpu(cpu_id: CpuId) -> NodeId {
...
+/// # Panics
+///
+/// This method panics if the NUMA topology is not initialized.
+pub fn leader_cpu_of(cpu_id: CpuId) -> CpuId {
+    debug_assert!(LEADER_CPU.get_on_cpu(cpu_id).is_completed());
+
+    // SAFETY: As far as the safe APIs are concerned, `LEADER_CPU` has been initialized.
+    unsafe { *LEADER_CPU.get_on_cpu(cpu_id).get_unchecked() }
+}
+/// Defines a statically-allocated CPU-local variable, which is only meaningful
```
junyang-zh, line 69: Add a line break.

```suggestion
}

/// Defines a statically-allocated CPU-local variable, which is only meaningful
```

ostd/src/numa.rs
```diff
@@ -52,8 +55,87 @@ pub fn node_id_of_cpu(cpu_id: CpuId) -> NodeId {
...
+///
+/// pub fn foo(leader_cpu: CpuId) -> &'static AtomicU32 {
+///   debug_assert!(is_leader_cpu(leader_cpu));
+///   FOO.get_on_cpu(leader_cpu)
+/// }
+/// ```
+///
+/// # Panics
+///
+/// The accessor functions panic if the given CPU is not a leader CPU.
```
junyang-zh, line 95: Move this `# Panics` section to the expanded function definition? Specifically, it can appear both before line 87 and line 109.

ostd/src/numa.rs
```diff
@@ -52,8 +55,87 @@ pub fn node_id_of_cpu(cpu_id: CpuId) -> NodeId {
...
+///
+/// ```rust
+/// cpu_local! {
+///     static FOO: AtomicU32 = AtomicU32::new(1);
+/// }
+///
+/// pub fn foo(leader_cpu: CpuId) -> &'static AtomicU32 {
+///   debug_assert!(is_leader_cpu(leader_cpu));
+///   FOO.get_on_cpu(leader_cpu)
+/// }
```
junyang-zh, line 90: I think this API is weird. Why not just call the macro `numa_node_local!` and use `foo(node: NodeId)` as the accessor function signature? Internally you can still use `cpu_local!` to implement it and optimize later.

ostd/src/arch/x86/boot/numa.rs
```diff
@@ -0,0 +1,618 @@
...
+
+/// Parses the SRAT table and returns the number of NUMA nodes, along with
+/// initialized processor affinity and memory range arrays.
+///
+/// Returns `None` if the SRAT table is missing.
+fn parse_srat_table() -> Option<(
+    usize,
+    &'static mut [ProcessorAffinity],
+    &'static mut [MemoryRange],
+)> {
```
junyang-zh, line 59: This signature looks problematic since calling it twice will lead to two mutable references to the same memory location. Better to make it `unsafe`.

osdk/deps/frame-allocator/src/pools/mod.rs
```diff
@@ -5,29 +5,34 @@ mod balancing;
...
     sync::atomic::{AtomicUsize, Ordering},
 };
 
 use ostd::{
+    cpu::{CpuId, PinCurrentCpu},
     cpu_local,
     irq::DisabledLocalIrqGuard,
+    leader_cpu_local,
     mm::Paddr,
+    numa::{leader_cpu_of, leader_cpu_of_node, memory_ranges, num_cpus_in_node, NodeId},
```
junyang-zh, line 18: The NUMA functionalities should also be behind a feature gate in `osdk-frame-allocator`.

junyang-zh (commented):
Here's some initial comment. I haven't gone through the ACPI parsing yet.

ostd/src/arch/x86/boot/numa.rs
vvvvsv: > calling it twice will lead to two mutable references to the same memory location

Calling it twice will `early_alloc` twice and return references to two different regions, containing the same contents. This is meaningless but safe?

However, I realized this method can return `&` instead of `&mut`, then it is fine.

PR #2552, opened 2025-11-04
Support reboot syscall

cchanging:
When using systemd as the init process, the original system `exit` only terminates the current shell and cannot shut down the entire system. Therefore, it is necessary to implement the `reboot` syscall to allow the system to shut down properly. 

With this PR, when systemd is not used, the `poweroff -f` command can be employed to invoke reboot for shutdown (without the `-f` option, the system may fail to shut down properly due to pre-shutdown service stopping operations and exit with an error like `ake: *** [Makefile:275: run] Error 1`, though it still exits...). When systemd is used, the `poweroff` command allows systemd to handle pre-shutdown cleanup and stopping operations before invoking `reboot` to exit.

PR #2636, opened 2025-11-28
Make `reboot -f` work on real hardware

lrh2000:
Following https://github.com/asterinas/asterinas/pull/2552, this PR makes `reboot -f` work on my laptop. `reboot` itself won't work because it tries to communicate with the init process, but our init process is currently a busybox shell.

This PR creates a new top-level module, `power`, in OSTD for power management. Currently, it provides interfaces for `restart()` and `poweroff()`.
```rust
/// Restarts the system.
///
/// On success, this function won't return. However, it may return if a restart handler is missing
/// or the handler does not work.
pub fn restart(code: ExitCode) { /* .. */ }

/// Powers off the system.
///
/// On success, this function won't return. However, it may return if a poweroff handler is missing
/// or the handler does not work.
pub fn poweroff(code: ExitCode) { /* .. */ }
```

But the problem is:
 - We cannot implement them all in OSTD. For example, implementing ACPI power-off is very complex. It requires parsing the AML table to determine the register and value. The AML is a Turing-complete language that requires a very complex interpreter. So it should live outside OSTD.
 - OSTD should provide a basic implementation for simple cases. For example, it should support QEMU power-off through the ISA-debug-exit device. Otherwise, we have problems when performing unit tests (i.e., `ktest`s).

I follow our injection strategy. Depending on the specific architecture, OSTD may or may not have a built-in restart or power-off mechanism. If it does not have, then it's the kernel's (i.e., `aster-nix`'s) job to provide one and inject it into OSTD.
```rust
/// Injects a handler that can restart the system.
///
/// The function may be called only once; subsequent calls take no effect.
///
/// Note that, depending on the specific architecture, OSTD may already have a built-in handler. If
/// so, calling this function outside of OSTD will never take effect. Currently, it happens in
///  - x86_64: Never;
///  - riscv64: Always;
///  - loongarch64: Never.
pub fn inject_restart_handler(handler: fn(ExitCode)) { /* .. */ }

/// Injects a handler that can power off the system.
///
/// The function may be called only once; subsequent calls take no effect.
///
/// Note that, depending on the specific architecture, OSTD may already have a built-in handler. If
/// so, calling this function outside of OSTD will never take effect. Currently, it happens in
///  - x86_64: If a QEMU hypervisor is detected;
///  - riscv64: Always;
///  - loongarch64: Never.
pub fn inject_poweroff_handler(handler: fn(ExitCode)) { /* .. */ }
```

Based on this framework, I implemented a restart mechanism via the ACPI table in `aster-nix`. It works on my laptop.

With this PR, we can achieve:
|               | Restart                      | Power off                                   |
|---------------|------------------------------|---------------------------------------------|
| x86_64 (QEMU) | Yes (via ACPI, in aster-nix) | Partial (if isa-debug-exit exists, in OSTD) |
| x86_64 (Real) | Yes (via ACPI, in aster-nix) | No (AML is too complex)                     |
| riscv64       | Yes (via SBI, in OSTD)       | Yes (via SBI, in OSTD)                      |
| loongarch64   | No                           | Yes (via syscon, in aster-nix) [^1]         |

[^1]: If we want to run `ktest`s in CI, we should have a power-off mechanism in OSTD. But it can be a naive implementation that hardcodes constants (i.e., register addresses) under the `#[cfg(ktest)]` gate.

### DIFF
--- a/kernel/comps/pci/src/arch/x86/mod.rs
+++ b/kernel/comps/pci/src/arch/x86/mod.rs
@@ -50,7 +50,11 @@ pub(crate) fn has_pci_bus() -> bool {
 }
 
 pub(crate) fn init() {
-    PCI_ADDRESS_PORT.call_once(|| IoPort::acquire(0xCF8).unwrap());
+    // We use `acquire_overlapping` to acquire the port at 0xCF8 because 0xCF9 may be used as a
+    // reset control register in the PIIX4. Although the two ports overlap in their I/O range, they
+    // serve completely different purposes. See
+    // <https://www.intel.com/Assets/PDF/datasheet/290562.pdf>.
+    PCI_ADDRESS_PORT.call_once(|| IoPort::acquire_overlapping(0xCF8).unwrap());
     PCI_DATA_PORT.call_once(|| IoPort::acquire(0xCFC).unwrap());
 }
 
--- a/ostd/src/io/io_port/allocator.rs
+++ b/ostd/src/io/io_port/allocator.rs
@@ -23,9 +23,17 @@ pub(super) struct IoPortAllocator {
 }
 
 impl IoPortAllocator {
-    /// Acquires the `IoPort`. Return None if any region in `port` cannot be allocated.
-    pub(super) fn acquire<T, A>(&self, port: u16) -> Option<IoPort<T, A>> {
-        let range = port..port.checked_add(size_of::<T>().try_into().ok()?)?;
+    /// Acquires an `IoPort`. Returns `None` if the PIO range is unavailable.
+    ///
+    /// `is_overlapping` indicates whether another `IoPort` can have a PIO range that overlaps with
+    /// this one. If it is true, only the first port in the PIO range will be marked as occupied;
+    /// otherwise, all ports in the PIO range will be marked as occupied.
+    pub(super) fn acquire<T, A>(&self, port: u16, is_overlapping: bool) -> Option<IoPort<T, A>> {
+        let range = if !is_overlapping {
+            port..port.checked_add(size_of::<T>().try_into().ok()?)?
+        } else {
+            port..port.checked_add(1)?
+        };
         debug!("Try to acquire PIO range: {:#x?}", range);
 
         let mut allocator = self.allocator.lock();
@@ -38,7 +46,7 @@ impl IoPortAllocator {
         }
 
         // SAFETY: The created `IoPort` is guaranteed not to access system device I/O.
-        unsafe { Some(IoPort::new(port)) }
+        unsafe { Some(IoPort::new_overlapping(port, is_overlapping)) }
     }
 
     /// Recycles an PIO range.
@@ -105,6 +113,7 @@ mod test {
     use crate::{arch::device::io_port::ReadWriteAccess, prelude::*};
 
     type IoPort = crate::io::IoPort<u32, ReadWriteAccess>;
+    type ByteIoPort = crate::io::IoPort<u8, ReadWriteAccess>;
 
     #[ktest]
     fn illegal_region() {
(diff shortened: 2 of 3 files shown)


lrh2000:
This PR originally included a commit ["Allocate I/O ports instead of ranges"](https://github.com/asterinas/asterinas/pull/2636/commits/ee5ded85eace48e33fe109c51d2a8fa6a24977b8). However, it appears that this commit is incorrect, as Intel documentation states:
> Any two consecutive 8-bit ports can be treated as a 16-bit port, and any four consecutive ports can be a 32-bit port. In this manner, the processor can transfer 8, 16, or 32 bits to or from a device in the I/O address space. Like words in memory, 16-bit ports should be aligned to even addresses (0, 2, 4, ...) so that all 16 bits can be transferred in a single bus cycle. Likewise, 32-bit ports should be aligned to addresses that are multiples of four (0, 4, 8, ...). The processor supports data transfers to unaligned ports, but there is a performance penalty because one or more extra bus cycle must be used.

I have removed the commit. But without the commit, the ACPI reset will not work in QEMU or LENOVO 21EX002GCD due to I/O port conflicts. It can work on LENOVO LNVNB161216 (my personal laptop), though.
 - For QEMU and LENOVO 21EX002GCD, there is a [Reset Control Register at I/O port 0xCF9](https://edc.intel.com/content/www/jp/ja/design/products-and-solutions/processors-and-chipsets/comet-lake-u/intel-400-series-chipset-on-package-platform-controller-hub-register-database/reset-control-register-rst-cnt-offset-cf9/). However, the well-known 4-byte PCI address register is located at 0xCF8, so they conflict: https://github.com/asterinas/asterinas/blob/2fc94a0d060bc674b576b3ba7789062de0d9f035/kernel/comps/pci/src/arch/x86/mod.rs#L53
 - For LENOVO LNVNB161216, the ACPI reset goes through I/O port 0xB2 (which seems to be [APMC—Advanced Power Management Control Port (IO)](https://www.intel.com/Assets/PDF/datasheet/290562.pdf), but I'm not sure). So there are no conflicts.

ostd/src/arch/loongarch/qemu.rs
```diff
@@ -15,7 +15,7 @@ pub enum QemuExitCode {
 
 /// Exits QEMU with the given exit code.
 //  FIXME: Support the transfer of the exit code to QEMU and multiple platforms.
-pub fn exit_qemu(_exit_code: QemuExitCode) -> ! {
+pub fn exit_qemu(_exit_code: QemuExitCode) {
```
tatetian: By changing the return type, the semantics of the function is changed. The previous form provides the guarantee that the kernel will exit. But the new form is saying that the action of QEMU exit may not succeed. To avoid confusion, I suppose that it makes sense to change the function name to `try_exit_qemu`. And you might want to extend the function's Rust doc to explain in what circumstances that this function would return.

ostd/src/power.rs
```diff
@@ -0,0 +1,74 @@
...
+///  - loongarch64: Never.
+pub fn inject_restart_handler(handler: fn(ExitCode)) {
+    RESTART_HANDLER.call_once(|| handler);
+}
+
+/// Restarts the system.
+///
+/// On success, this function won't return. However, it may return if a restart handler is missing
+/// or the handler does not work.
+pub fn restart(code: ExitCode) {
```
tatetian: What do you expect the caller to do if this function does return? I don't think this caller can do anything meaningful. So I suppose that making this function guarantee it will never return would be better.

The same argument applies to `poweroff`.

You might also want to explain the meaning of the exit code. On QEMU, the exit code will be the one QEMU process; on hardware platforms, the exit code does not have a real impact.

lrh2000:
I'm still wondering what I should do to resolve https://github.com/asterinas/asterinas/pull/2636#issuecomment-3592169038.

A possible stragety is to add a new `IoPort::acquire_overlapping` and use it to allocate the PCI port:
```patch
diff --git a/kernel/comps/pci/src/arch/x86/mod.rs b/kernel/comps/pci/src/arch/x86/mod.rs
index 3885bfbf7..3e923b745 100644
--- a/kernel/comps/pci/src/arch/x86/mod.rs
+++ b/kernel/comps/pci/src/arch/x86/mod.rs
@@ -50,7 +50,11 @@ pub(crate) fn has_pci_bus() -> bool {
 }
 
 pub(crate) fn init() {
-    PCI_ADDRESS_PORT.call_once(|| IoPort::acquire(0xCF8).unwrap());
+    // We use `acquire_overlapping` to acquire the port at 0xCF8 because 0xCF9 may be used as a
+    // reset control register in the PIIX4. Although the two ports overlap in their I/O range, they
+    // serve completely different purposes. See
+    // <https://www.intel.com/Assets/PDF/datasheet/290562.pdf>.
+    PCI_ADDRESS_PORT.call_once(|| IoPort::acquire_overlapping(0xCF8).unwrap());
     PCI_DATA_PORT.call_once(|| IoPort::acquire(0xCFC).unwrap());
 }
``` 

`IoPort::acquire_overlapping` will occupy only the port (e.g., `0xCF8`) instead of the port range (`0xCF8..0xCFC`). Then the reset port (`0xCF9`) can be allocated without problems.

Then, both QEMU and our two types of laptops will have reboot functionality.

I don't think this approach is perfect, but it's at least acceptable. I'll add a commit for this. Is there a better strategy of which I'm unaware?

PR #2640, opened 2025-11-29
Fix a bug of caching negative dentry

cchanging:
This bug fix PR is extracted from PR #2282 because it is relatively independent. 

PR #1784 previously introduced a mechanism for caching negative dentries to accelerate filesystem lookups. Unfortunately, due to the misuse of `is_dentry_cachable`, the current implementation of this mechanism contains a bug, as detailed in the [comment](https://github.com/asterinas/asterinas/pull/2282#discussion_r2567808151). 

To fix this bug while preserving the negative dentry caching mechanism would require significant modifications to our current implementation of cachable logic. This step is reserved as future work and will be addressed in subsequent PRs.


tatetian (approved):

PR #2645, opened 2025-11-30
Invalidate IRE cache to resolve IOMMU faults on real hardware

lrh2000:
On LENOVO 21EX002GCD, an IOMMU fault with fault reason 0x22 occurs when I press a key. The fault explanation is:
> The Present (P) field in the IRTE entry corresponding to the interrupt_index of the interrupt request is Clear.

It seems that we need to invalidate the interrupt remapping cache after modifying the interrupt remapping table. Then, the IOMMU fault will go away.

### DIFF
--- a/ostd/src/arch/x86/iommu/registers/mod.rs
+++ b/ostd/src/arch/x86/iommu/registers/mod.rs
@@ -140,32 +140,6 @@ impl IommuRegisters {
         self.write_global_command(GlobalCommand::IRE, true);
         while !self.read_global_status().contains(GlobalStatus::IRES) {}
 
-        // Invalidate interrupt cache
-        if self.read_global_status().contains(GlobalStatus::QIES) {
-            let mut queue = QUEUE.get().unwrap().lock();
-
-            // Construct global invalidation of interrupt cache and invalidation wait.
-            queue.append_descriptor(InterruptEntryCache::global_invalidation().0);
-            let tail = queue.tail();
-            self.invalidate
-                .queue_tail
-                .as_mut_ptr()
-                .write((tail << 4) as u64);
-            while (self.invalidate.queue_head.as_ptr().read() >> 4) + 1 == tail as u64 {}
-
-            // We need to set the interrupt flag so that the `Invalidation Completion Status Register` can report the completion status.
-            queue.append_descriptor(InvalidationWait::with_interrupt_flag().0);
-            self.invalidate
-                .queue_tail
-                .as_mut_ptr()
-                .write((queue.tail() << 4) as u64);
-
-            // Wait for completion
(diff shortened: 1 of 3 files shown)


ostd/src/arch/x86/iommu/registers/mod.rs
```diff
@@ -216,6 +190,35 @@ impl IommuRegisters {
...
+    pub(super) fn invalidate_interrupt_cache(&mut self) {
+        if !self.read_global_status().contains(GlobalStatus::QIES) {
+            self.global_invalidation();
+            return;
+        }
+
+        let mut queue = QUEUE.get().unwrap().lock();
+
+        // Clear the Invalidation Completion Status Register.
+        self.invalidate.completion_status.as_mut_ptr().write(1);
```
sdww0: I checked the iommu specification, but didn't find a reference to clearing this register by setting it to 1. The specification seems to suggest unsetting the register instead.

Additionally, I think it would be better to clear the register after waiting for completion (line 219).

lrh2000: > didn't find a reference to clearing this register by setting it to 1

I think it's:

image

RW1CS: “Sticky Read-only status, Write-1-to-clear status” field. Software can read this bit to find the value of status. Software can write a value of '1' to Clear this bit. Writing a '0' to the bit has no effect. This bit is only re-
initialized to its default value by a “Power Good Reset”.

image

> Additionally, I think it would be better to clear the register after waiting for completion (line 219).

Does this mean clearing the register twice, once at the beginning and once at the end?

sdww0: > RW1CS: “Sticky Read-only status, Write-1-to-clear status” field. Software can read this bit to find the value of status. Software can write a value of '1' to Clear this bit. Writing a '0' to the bit has no effect. This bit is only re- initialized to its default value by a “Power Good Reset”.

Thanks for the information. That should make sense.

> Does this mean clearing the register twice, once at the beginning and once at the end?

Yes, in case we use the queue to do other operations in the future.

lrh2000: > in case we use the queue to do other operations in the future

I think we should establish a rule to clear the flag before or after use. Clearing it twice sounds a bit odd to me.

How do you like [this version](https://github.com/asterinas/asterinas/compare/637762bec38c69dbcfd5d4b139f42d21f21b0a63..207eeee3f068508d66029582ee399fb33958aa82)? We can clear the register at initialization and again each time the queue is used.

sdww0: I think it is ok. We can encapsulate the operations when we need to use the queue for other operations.

PR #2650, opened 2025-12-01
Correct `{get,set}_option` for `NetlinkSocket`

lrh2000:
https://github.com/asterinas/asterinas/blob/0b664a012e689b3f730cc4faa39bd03613f5b995/kernel/src/net/socket/netlink/common/mod.rs#L176-L180

This should be fixed so that we don't return an `Ok()` for all unrecognized socket options.

It seems that the test only needs to change the buffer size, which is already handled in the socket-level option set, although it does not take effect. So this PR adds `SocketOptionSet` for `NetlinkSocket`.

### DIFF
--- a/kernel/src/net/socket/netlink/common/mod.rs
+++ b/kernel/src/net/socket/netlink/common/mod.rs
@@ -9,14 +9,15 @@ use super::{GroupIdSet, NetlinkSocketAddr};
 use crate::{
     events::IoEvents,
     fs::utils::Inode,
-    match_sock_option_ref,
+    match_sock_option_mut, match_sock_option_ref,
     net::socket::{
         netlink::{table::SupportedNetlinkProtocol, AddMembership, DropMembership},
         new_pseudo_inode,
-        options::SocketOption,
+        options::{Error as SocketError, SocketOption},
         private::SocketPrivate,
         util::{
             datagram_common::{select_remote_and_bind, Bound, Inner},
+            options::{GetSocketLevelOption, SetSocketLevelOption, SocketOptionSet},
             MessageHeader, SendRecvFlags, SocketAddr,
         },
         Socket,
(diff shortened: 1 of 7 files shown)


kernel/src/net/socket/netlink/common/mod.rs
```diff
@@ -169,17 +185,38 @@ where
...
-                    e
-                );
-                Ok(())
+    fn get_option(&self, option: &mut dyn SocketOption) -> Result<()> {
+        // All methods in `GetSocketLevelOption` have nothing to do with netlink sockets. So this
+        // is a correct implementation as far as `GetSocketLevelOption` is concerned.
+        struct DummyState;
+        impl GetSocketLevelOption for DummyState {
+            fn is_listening(&self) -> bool {
+                false
```
StevenJiang1110: Should we add a branch for `SO_ERROR` just like UDP socket? `SocketOptionSet::get_option` will not handle `SO_ERROR` actually.

kernel/src/net/socket/netlink/common/mod.rs
```diff
@@ -169,17 +185,38 @@ where
...
-            Err(e) => {
-                warn!(
-                    "We currently ignore set option errors to pass libnl test: {:?}",
-                    e
-                );
-                Ok(())
+    fn get_option(&self, option: &mut dyn SocketOption) -> Result<()> {
+        // All methods in `GetSocketLevelOption` have nothing to do with netlink sockets. So this
+        // is a correct implementation as far as `GetSocketLevelOption` is concerned.
+        struct DummyState;
```
StevenJiang1110: Introducing a `DummyState` for the current socket options should be sufficient. However, if we add more options in the future — for example, `SO_PROTOCOL` to obtain the socket protocol — won't `DummyState` be unable to provide that information? How do you like implementing `GetSocketLevelOption` and `SetSocketLevelOption` for `Inner` struct instead?

lrh2000: Oh well, sorry. I misinterpreted another problem that was preventing us from implementing `GetSocketLevelOption` for Inner. But it seems that the problem doesn't really exist. My mistake for the noise.

kernel/src/net/socket/netlink/common/mod.rs
lrh2000: This problem exists for UNIX sockets as well. Thanks for noticing this!

StevenJiang1110 (approved):
LGTM! Thank for your contribution!

PR #2652, opened 2025-12-01
Build the Asterinas NixOS ISO installer image

cqs21:
This PR adds support for ISO-based installation of the Asterinas NixOS distribution. It depends on #2621.

## How to use

Build and run the ISO installer image with automatic installation:
```
make run_iso AUTO_INSTALL=1
```
Run the installed system:
```
make run_nixos ISO=1
```


test/nix/iso-image.nix
```diff
@@ -0,0 +1,24 @@
...
+  configuration = {
+    imports = [
+      "${pkgs.path}/nixos/modules/installer/cd-dvd/installation-cd-minimal.nix"
+      "${pkgs.path}/nixos/modules/installer/cd-dvd/channel.nix"
+    ];
+
+    services.getty.autologinUser = lib.mkForce "root";
+    environment.loginShellInit = "${installer}";
+  };
+in (pkgs.nixos configuration).config.system.build.isoImage
```
tatetian: The Nix expression to build the ISO image should be part of the distro. Move this file to `distro/`.

tools/nixos/iso_installer.sh
```diff
@@ -0,0 +1,126 @@
...
+echo "nameserver $dns_server_ip" > "$NIXOS_RESOLV_CONF"
+
+get_user_input "Set log level (error/warn/info/debug/trace)" "error" LOG_LEVEL
+get_user_input "Set console (hvc0/tty0)" "hvc0" CONSOLE
+get_user_input "Disable systemd (true/false)" "true" NIXOS_DISABLE_SYSTEMD
+get_user_input "Stage 2 init command (fallback when systemd disabled)" "/bin/sh" NIXOS_STAGE_2_INIT
+export LOG_LEVEL
+export CONSOLE
+export NIXOS_STAGE_2_INIT
+export NIXOS_DISABLE_SYSTEMD
```
tatetian: Do not ask the user for these advanced env vars.

First of all, they should NOT have been set as env vars in the first place. Settings like `LOG_LEVEL`, `CONSOLE`, `NIXOS_STAGE_2_INIT` should have been set in `configuration.nix`. If they are Nix options, then they can be set in the editor step.

Second, they are advanced settings. Advanced settings should not be present to beginner users. Yes, you are kindly enough to provide default values for them. But the user is still bothered by the prompt to set it or take the default value. The user may be wondering what "log level" is set, what is "hvc0", why "Disable systemd"? Advanced users can directly edit this script or set the corresponding env vars.

In short, set these variables with sensible default values in this script. Advanced users may override them either by executing `export` or edit this file.

.github/workflows/test_iso_image.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - main
+
+jobs:
+  iso-test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: endersonmenezes/free-disk-space@v3
```
tatetian: Add a comment to explain the necessity of calling this action.

Makefile
```diff
@@ -288,13 +299,23 @@ initramfs: check_vdso
...
+ifeq ($(ISO),1)
+	@nix-build test/nix -A iso-image --out-link target/nixos/iso_image
+else ifeq ($(NIXOS),1)
 	@./tools/nixos/install_asterinas.sh target/nixos
 endif
 
 .PHONY: run
 run: initramfs $(CARGO_OSDK)
-ifeq ($(NIXOS),1)
+ifeq ($(ISO), 1)
```
tatetian: This `run` Makefile target looks terrible. It is completely unreadable. It has four branches, in which the command `cargo osdk build $(CARGO_OSDK_BUILD_ARGS)` is duplicated twice and `./tools/nixos/run_iso.sh target/nixos` is duplicated triple. The command `nix-build test/nix -A iso-image --out-link target/nixos/iso_image` is duplicated in both the `run` and `build` targets...

Could you try reorganize the Makefile to simplify the Makefile? Maybe now is the time to add new Makefile targets:

```bash
# Make the kernel with an initramfs and then run it
make kernel
make run_kernel

# Make the Asterinas NixOS ISO installer
make iso
make run_iso

# Create an Asterinas NixOS installation
make nixos

# After creating a Asterinas NixOS installation (via either the `run_iso` or `nixos` target),
# run the NixOS
make run_nixos
```

tatetian (commented):
Thanks for adding ISO installer!

.github/workflows/test_iso_image.yml
cqs21: The larger runner (`ubuntu-2-cores-75GB-ssd`) actually only has 8.8GB of available disk space (the same as the standard runner), see https://github.com/asterinas/asterinas/actions/runs/19881237165/job/56979321494#step:4:10. Perhaps we could create runners with larger disk (e.g., 150GB), so we wouldn't need to manually free disk space.

Makefile
```diff
@@ -438,6 +450,9 @@ check: initramfs $(CARGO_OSDK)
 	@
 	@# Check typos
 	@typos
+	@
+	@# Check if the Nix expressions are formatted correctly
+	@nixfmt --check distro test/nix
```
tatetian: Remove these lines. Another PR has introduced the same change.

Makefile
```diff
@@ -343,6 +313,48 @@ else ifeq ($(AUTO_TEST), vsock)
 		|| (echo "Vsock test failed" && exit 1)
 endif
 
+ASTERINAS_DISK ?= $(abspath target/nixos/asterinas.img)
+
+# Build the Asterinas NixOS ISO installer image
+.PHONY: iso
+iso: BOOT_PROTOCOL = linux-efi-handover64
+iso:
+	@make kernel
```
tatetian, line 316: Any reason not to treat the `kernel` target as a dependency of the `iso` target?

```Makefile
iso: kernel
	@nix-build \
        ...
```

Makefile
```diff
@@ -343,6 +313,48 @@ else ifeq ($(AUTO_TEST), vsock)
 		|| (echo "Vsock test failed" && exit 1)
 endif
 
+ASTERINAS_DISK ?= $(abspath target/nixos/asterinas.img)
```
tatetian: Let's name Asterinas NixOS stuff consistently. The Makefile targets are called `nixos` and `run_nixos`. We do not call them `aster_nixos` or `run_aster_nixos` because in the Asterinas repo, we only build, run, and test AsterNixOS, not the vanilla NixOS. So in all places that refer to Asterinas NixOS, let's stick with "nixos". This means the following changes:

* `ASTERINAS_DISK` --> `NIXOS_DISK`
* `nixos/asterinas.img` --> `nixos/nixos.img`
* `tools/nixos/install_asterinas.sh` --> `tools/nixos/install_nixos.sh`

Makefile
```diff
@@ -54,9 +54,17 @@ VHOST ?= off
 DNS_SERVER ?= none
 # End of network settings
 
+# ISO installer settings
+ISO ?= 0
```
tatetian: I don't think this variable is or should be useful anymore as the command for building ISO is `make iso`, not `make build ISO=1`.

tools/nixos/iso_installer.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -e
+
+echo "=== Asterinas NixOS Installer ==="
+echo "Default paths:"
+echo "  distro: @distroPath@"
+echo "  kernel: @kernelPath@"
+echo "  tools: @toolsPath@"
```
tatetian: How does printing these paths help the user of this ISO installer? They always point to folders under Nix stores. The hash values do not mean anything to the user.

In addition, these paths should NOT be called "default paths" because the script provide no means to override them.

Makefile
```diff
@@ -317,6 +313,48 @@ else ifeq ($(AUTO_TEST), vsock)
...
+.PHONY: run_iso
+run_iso: OVMF = off
+run_iso: iso
+	@./tools/nixos/run_iso.sh target/nixos
+
+# Create an Asterinas NixOS installation on host
+.PHONY: nixos
+nixos: BOOT_PROTOCOL = linux-efi-handover64
+nixos:
+	@make kernel
```
tatetian, line 327: Likewise, the `kernel` Makefile target should be added to the `nixos` target as a dependency.

distro/install_script/aster_configuration_template.nix
```diff
@@ -23,9 +23,63 @@
   # Overlays provide patches to 'nixpkgs' that enable these packages to run effectively on AsterNixOS.
   # For details on the overlay file definition format, 
   # please refer to https://nixos.org/manual/nixpkgs/stable/#sec-overlays-definition.
-  nixpkgs.overlays = [
+  config.nixpkgs.overlays = [
```
tatetian, line 30: What is the difference? Is the original line wrong?

distro/aster_configuration.nix
```diff
@@ -23,9 +23,63 @@
...
-  nixpkgs.overlays = [
+  config.nixpkgs.overlays = [
     (import ./overlays/hello-asterinas/default.nix)
     (import ./overlays/desktop/default.nix)
     (import ./overlays/podman/default.nix)
   ];
+
+  # The Asterinas NixOS special options.
+  options = {
+    asterinas = {
```
tatetian: Rename this to `aster_nixos`

tools/nixos/iso_installer.sh
```diff
@@ -0,0 +1,118 @@
...
+    echo "Error: Configuration file not found in @distroPath@"
+    exit 1
+fi
+
+get_user_input "Proceed with installation? (y/n)" "y" confirm
+if [ "$confirm" = "n" ] || [ "$confirm" = "N" ]; then
+    echo "Installation cancelled!"
+    exit 0
+fi
+@toolsPath@/install_asterinas.sh ${NIXOS_DIR} ${DEVICE}
```
tatetian: Could you try combine the `install_asterinas.sh` and `iso_installer.sh` files into one?

The first hint that having two separate scripts are kind of redundant is that their names are very close. How can one tell the differences between `install_asterinas.sh` and `iso_installer.sh`?

Second, `iso_installer.sh` does a few more things than `install_asterinas.sh`, including
1. Erasing the partitions off the target disk;
2. Opening an editor
3. Powering off

I think these extra steps can be added to the `install_asterinas.sh`.

Third, `iso_installer.sh` has used the template technique (e.g., `@distroPath@`) to avoid relying on env vars. But I suppose that `install_asterinas.sh` can benefit from the same technique, right?

So I suggest rethinking the CLI interface of `install_asterinas.sh` (e.g., adding more arguments or adjusting existing arguments) so that it can cover the functionality of both use cases: installing AsterNixOS via ISO and on the host.

tools/nixos/install_asterinas.sh
```diff
@@ -1,28 +1,49 @@
...
 
 NIXOS_DIR=$(realpath $1)
-SCRIPT_DIR=$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)
+SCRIPT_DIR=$(cd "$( dirname "$0" )" && pwd)
 ASTER_IMAGE_PATH=${NIXOS_DIR}/asterinas.img
 BUILD_DIR=$(mktemp -d -p /mnt)
 ASTERINAS_DIR=$(realpath ${SCRIPT_DIR}/../..)
-DISTRO_DIR=$(realpath ${ASTERINAS_DIR}/distro)
+DISTRO_DIR=${DISTRO_DIR:-"$(realpath ${ASTERINAS_DIR}/distro)"}
 NIXOS_DISK_SIZE_IN_MB=${NIXOS_DISK_SIZE_IN_MB:-"8196"}
```
tatetian: I think this input can be removed from the script. We can ask the Makefile to prepare a memory disk of the right size and let this script simply take a target disk (an mandatory argument) rather than preparing the memory disk in this script.

distro/install_script/aster_configuration_template.nix
cqs21: We cannot use the attribute `nixpkgs` directly when introduces a top-level `config` or `options` attribute.

Makefile
cqs21: If use `nixos: kernel` as a dependency, the `kernel` target runs in a separate subprocess that doesn't inherit the `BOOT_PROTOCOL` variable set within the `nixos` target. Alternatively, we would keep this condition:
```
ifeq ($(NIXOS), 1)
BOOT_PROTOCOL = linux-efi-handover64
endif
```

distro/iso_image/default.nix
```diff
@@ -0,0 +1,27 @@
...
+    imports = [
+      "${pkgs.path}/nixos/modules/installer/cd-dvd/installation-cd-minimal.nix"
+      "${pkgs.path}/nixos/modules/installer/cd-dvd/channel.nix"
+    ];
+
+    services.getty.autologinUser = pkgs.lib.mkForce "root";
+    environment.systemPackages = [ installer ];
+    environment.loginShellInit = ''
+      if [ ! -e "$HOME/configuration.nix" ]; then
+        cp -L ${installer}/etc_nixos/configuration.nix $HOME
```
tatetian, line 16: Add a comment to explain why do this copy:

```nix
      # Present the user with an editable copy of configuration.nix at her home
      if [ ! -e "$HOME/configuration.nix" ]; then
        cp -L ${installer}/etc_nixos/configuration.nix $HOME
```

.github/workflows/test_iso_image.yml
```diff
@@ -0,0 +1,32 @@
...
+      image: asterinas/asterinas:0.16.2-20251209
+      options: -v /dev:/dev --privileged
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build Asterinas NixOS ISO image and do installation
+        run: |
+          make iso || true
+          make run_iso || true
+          tail --lines 10 qemu.log | grep -q "^Installation finished" || (echo "Test ISO failed" && exit 1)
```
tatetian, line 33: After running the ISO installer, can we do `make run_nixos` to verify the installation is actually working?

tatetian (commented):
Great job! This code looks mostly good.

PR #2653, opened 2025-12-02
frame: Add Reclaimable memory region to the `max_paddr` filter

sdww0:
Sometimes, the reclaimable memory region (such as the initramfs) may exceed the usable memory region, causing the initramfs region to be ignored when establishing linear mappings. 

This issue had previously gone undetected because the usable memory in the QEMU virtual machine was set to a relatively large value, with the highest usable memory region exceeding 4GB. This allowed all lower regions to establish linear mappings, but it can cause problems in low-memory environments, such as microVMs.

In fact, I only discovered this issue in Firecracker, and here is the screenshot of the memory regions:

image

I think adding the Reclaimable region to the filter should make sense since these regions also bind with physical frames.

### DIFF
--- a/ostd/src/mm/frame/meta.rs
+++ b/ostd/src/mm/frame/meta.rs
@@ -458,7 +458,12 @@ pub(crate) unsafe fn init() -> Segment<MetaPageMeta> {
         let regions = &crate::boot::EARLY_INFO.get().unwrap().memory_regions;
         regions
             .iter()
-            .filter(|r| r.typ() == MemoryRegionType::Usable)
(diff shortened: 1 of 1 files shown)


lrh2000:
Forgets to update `mark_unusable_ranges`?
https://github.com/asterinas/asterinas/blob/e648528c46b32bb9d78a0f395563501172caad21/ostd/src/mm/frame/meta.rs#L598

Well, let's add `MemoryRegionType::is_physical` instead. It should avoid similar issues in the future. See #2656.

PR #2658, opened 2025-12-02
Support ext2 to deal with fifo and devices

### LINKED ISSUE #2565: Device files should behave like pointers
**Think of device files (usually under `/dev`) as pointers to devices.** Thus, `mknod(path, S_IFCHR|mode, devid)` or `mknod(path, S_IFBLK|mode, devid)`   are “just” creating a special inode that points at a `(type, devid)`.  But this mental model is not followed by our current implementation.

## Asterinas's status quo

Device files, usually under `/dev`, may be spawned automatically the kernel thank to devtmpfs or manually by the userspace with the `mknod` system call.

https://github.com/asterinas/asterinas/blob/cdeb9ec5fd68603e48ec543c4f8ea88cfc97b2f4/kernel/src/syscall/mknod.rs#L49-L52

As shown by the code snippet above, our current `mknod` system call immediately transform a device ID, `dev`, into the corresponding device file, `device_inode`.

https://github.com/asterinas/asterinas/blob/cdeb9ec5fd68603e48ec543c4f8ea88cfc97b2f4/kernel/src/fs/utils/inode.rs#L322-L324

https://github.com/asterinas/asterinas/blob/cdeb9ec5fd68603e48ec543c4f8ea88cfc97b2f4/kernel/src/fs/utils/inode.rs#L235-L241

And the device file, represented by a `Arc<dyn Device>`, is passed to the underlying FS.
## Linux behaviors

In Asterinas, device files are "real": they are implemented by the devices themselves. But in Linux, device files are just "pointers" to devices, in a similar sense as symlinks are pointers to files.

This "pointer" conception has some direct implications:

* **Device files with nonexistent devices IDs may be created** with `mknod`. At creation time, the given device IDs are not validated; at open time, an error will be reported if the device ID is invalid. 
* **Device files with valid device IDs may become stale** if the corresponding device unregisters. Devtmpfs can remove stale devices automatically. But if you manually create such a device file, the kernel won't remove it for you and you end up with a stale device file.
* **File systems persist device files by easily** storing their device types and IDs.

This understanding is supported by Linux source code. For example, the [`chrdev_open`](https://github.com/torvalds/linux/blob/4a0c9b3391999818e2c5b93719699b255be1f682/fs/char_dev.c#L373-L424) function is called every time a character special file needs to be opened.

```c
static int chrdev_open(
    struct inode *inode, // The device inode
    struct file *filp // The open file struct to be initialized
)
```

This function looks up the char device using the device ID stored inside the inode.

```c
kobj = kobj_lookup(cdev_map, inode->i_rdev, &idx);
```

Then, using the file operations from the char device to set the target file struct:

```c
fops = fops_get(p->ops);
...
replace_fops(filp, fops);
```

Also, one can validate this understanding quickly by using the `mknod` command to create a device file for an non-existing device:

```
TYPE=c # char device
MAJOR=666
MINOR=888
mknod /dev/foo $TYPE $MAJOR $MINOR
```

In conclusion, Asterinas implementation needs to be fixed to comply with Linux's behaviors.

StevenJiang1110:
Fixes #2565.

When porting Podman to NixOS, I encountered several issues related to the ext2 file system. This PR addresses one of those issues.

The core problem is that Podman creates device nodes and FIFO nodes for containers, typically within the `/var/containers/storage/......`. On our current setup, this path is backed by an ext2 file system. However, creating FIFO and device inodes on an ext2 file system presently does not function correctly and leads to a kernel panic. Therefore, this PR implements a fix for this specific problem.

I also add a test for testing creating UNIX socket on ext2 and it shows ext2 works correctly for this case.

### DIFF
--- a/kernel/src/fs/ext2/impl_for_vfs/inode.rs
+++ b/kernel/src/fs/ext2/impl_for_vfs/inode.rs
@@ -133,16 +133,14 @@ impl Inode for Ext2Inode {
         match self.inode_type() {
             inode_type @ (InodeType::BlockDevice | InodeType::CharDevice) => {
                 let device_id = self.device_id();
-                if device_id == 0 {
+                let Some(device_id) = DeviceId::from_encoded_u64(device_id) else {
                     return Some(Err(Error::with_message(
                         Errno::ENODEV,
-                        "the device of ID 0 does not exist",
+                        "the device ID is invalid",
                     )));
-                }
+                };
                 let device_type = inode_type.device_type().unwrap();
-                let Some(device) =
-                    device::lookup(device_type, DeviceId::from_encoded_u64(device_id))
-                else {
+                let Some(device) = device::lookup(device_type, device_id) else {
                     return Some(Err(Error::with_message(
                         Errno::ENODEV,
                         "the required device ID does not exist",
--- a/kernel/libs/device-id/src/lib.rs
+++ b/kernel/libs/device-id/src/lib.rs
@@ -43,10 +43,12 @@ impl DeviceId {
 impl DeviceId {
     /// Creates a device ID from the encoded `u64` value.
     ///
-    /// Panics if the major or minor device number is not falling in the valid range.
-    pub fn from_encoded_u64(raw: u64) -> Self {
+    /// Returns `None` if the major or minor device number is not falling in the valid range.
+    pub fn from_encoded_u64(raw: u64) -> Option<Self> {
         let (major, minor) = decode_device_numbers(raw);
-        Self::new(MajorId::new(major as u16), MinorId::new(minor))
+        let major = MajorId::try_from(major as u16).ok()?;
+        let minor = MinorId::try_from(minor).ok()?;
+        Some(Self::new(major, minor))
     }
 
     /// Encodes the device ID as a `u64` value.
--- a/kernel/src/fs/ramfs/fs.rs
+++ b/kernel/src/fs/ramfs/fs.rs
@@ -211,17 +211,15 @@ impl Inner {
     ) -> Option<Result<Box<dyn FileIo>>> {
         match self {
             Self::BlockDevice(device_id) | Self::CharDevice(device_id) => {
-                if *device_id == 0 {
+                let Some(device_id) = DeviceId::from_encoded_u64(*device_id) else {
                     return Some(Err(Error::with_message(
                         Errno::ENODEV,
-                        "the device of ID 0 does not exist",
+                        "the device ID is invalid",
                     )));
-                }
+                };
 
                 let device_type = self.device_type().unwrap();
-                let Some(device) =
-                    device::lookup(device_type, DeviceId::from_encoded_u64(*device_id))
-                else {
+                let Some(device) = device::lookup(device_type, device_id) else {
                     return Some(Err(Error::with_message(
                         Errno::ENODEV,
                         "the required device ID does not exist",
--- a/kernel/src/syscall/mount.rs
+++ b/kernel/src/syscall/mount.rs
@@ -242,7 +242,7 @@ fn get_fs(
         }
 
         let id = DeviceId::from_encoded_u64(path.metadata().rdev);
-        let device = aster_block::lookup(id);
+        let device = id.and_then(aster_block::lookup);
         if device.is_none() {
             return_errno_with_message!(Errno::ENODEV, "the device is not found");
         }


kernel/src/fs/ext2/impl_for_vfs/inode.rs
```diff
@@ -121,6 +125,25 @@ impl Inode for Ext2Inode {
...
 
+    fn open(
+        &self,
+        access_mode: AccessMode,
+        status_flags: StatusFlags,
+    ) -> Option<Result<Box<dyn FileIo>>> {
+        match self.inode_type() {
+            InodeType::BlockDevice | InodeType::CharDevice => {
+                let device_id = self.device_id();
+                let device = match get_device(DeviceId::from_encoded_u64(device_id)) {
```
tatetian: The implementation of `get_device` actually only returns char devices.

https://github.com/asterinas/asterinas/blob/2168b32d88c6d4d423aefbb7212e2502c7acc15f/kernel/src/device/mod.rs#L47-L52

So this implementation of the `open` method fails to support block devices.

I suggest making the following changes.

First, add a `lookup` API for `device/registry/block`. This API should have been added before.

```rust
/// Looks up a char device of a given device ID.
pub fn lookup(id: DeviceId) -> Option<Arc<dyn Device>> {
    let block_device = aster_block::lookup(id)?;

    let mut registry = DEVICE_REGISTRY.lock();
    let block_device_file = registry
        .entry(&id)
        .or_insert_with(move || Arc::new(BlockFile::new(block_device)))
        .clone();
    Some(block_device_file)
}

// TODO: Merge the two mapping tables, one is here and the other is in the block component.
// Maintaining two mapping tables is undesirable due to duplication and (potential) inconsistency.
static DEVICE_REGISTRY: Mutex<BTreeMap<u32, Arc<dyn Device>>> = Mutex::new(BTreeMap::new());
```

Second, remove the old `get_device` API and add a new `lookup` API.

```rust
// file: src/device/registry/mod.rs

pub fn lookup(device_type: DeviceType, device_id: DeviceId) -> Option<Arc<dyn Device>> {
    match device_type {
        DeviceType::Char => char::lookup(device_id),
        DeviceType::Block => block::lookup(device_id)
    }
}

// file: src/device/mod.rs

pub use registry::lookup;
``` 

Last, in this method, call the new `device::lookup` function in two branches of the `match` expression, one for `InodeType::BlockDevice` and the other `InodeType::CharDevice`.

kernel/src/fs/ext2/impl_for_vfs/inode.rs
```diff
@@ -121,6 +125,25 @@ impl Inode for Ext2Inode {
         Some(self.page_cache())
     }
 
+    fn open(
+        &self,
+        access_mode: AccessMode,
+        status_flags: StatusFlags,
+    ) -> Option<Result<Box<dyn FileIo>>> {
+        match self.inode_type() {
+            InodeType::BlockDevice | InodeType::CharDevice => {
```
tatetian: Independent from this PR, could you be a dear and remove the `Misc` variant by adding another commit?

https://github.com/asterinas/asterinas/blob/2168b32d88c6d4d423aefbb7212e2502c7acc15f/kernel/src/fs/device.rs#L41-L48

kernel/src/fs/ext2/impl_for_vfs/inode.rs
```diff
@@ -121,6 +125,25 @@ impl Inode for Ext2Inode {
         Some(self.page_cache())
     }
 
+    fn open(
```
tatetian, line 128: To be consistent, the `open` method of `RamFs` should be changed to look up the device on open.

https://github.com/asterinas/asterinas/blob/2168b32d88c6d4d423aefbb7212e2502c7acc15f/kernel/src/fs/ramfs/fs.rs#L192-L203

As explained in #2565, an inode itself should only record the device ID. When the inode is opened, the open logic looks up the `Arc<dyn Device>` object using the `device` module.

---

Now that `Arc<dyn Device>` are looked up on open, inode objects such as `RamInode` 

https://github.com/asterinas/asterinas/blob/2168b32d88c6d4d423aefbb7212e2502c7acc15f/kernel/src/fs/ramfs/fs.rs#L126-L133

do not need to keep `Arc<dyn Device>` inside. They only need to keep the device type and ID.

This also means that `MknodType` does not need to store `Arc<dyn Device>`.

https://github.com/asterinas/asterinas/blob/2168b32d88c6d4d423aefbb7212e2502c7acc15f/kernel/src/fs/utils/inode.rs#L219-L223

Please incorporate the above changes as well in this PR.

kernel/src/fs/ext2/impl_for_vfs/inode.rs
```diff
@@ -131,12 +131,24 @@ impl Inode for Ext2Inode {
         status_flags: StatusFlags,
     ) -> Option<Result<Box<dyn FileIo>>> {
         match self.inode_type() {
-            InodeType::BlockDevice | InodeType::CharDevice => {
+            inode_type @ (InodeType::BlockDevice | InodeType::CharDevice) => {
                 let device_id = self.device_id();
-                let device = match get_device(DeviceId::from_encoded_u64(device_id)) {
-                    Ok(device) => device,
-                    Err(e) => return Some(Err(e)),
+                if device_id == 0 {
```
tatetian: There are other device ID values that are invalid in addition to the special value of 0. 

https://github.com/asterinas/asterinas/blob/5ef4cd820973f0dcab80957bd4a8aa3223cafa69/kernel/libs/device-id/src/lib.rs#L47-L50

The return type of `from_encoded_u64` should have been changed to `Option<Self>`. `MajorId` and `MinorId` implements `TryFrom<u16>` and `TryFrom<u32>`, respectively. Use them in the revised `from_encoded_u64` method.

tatetian (commented):
Thanks for the revision! The code is almost perfect. Just one minor issue.

PR #2661, opened 2025-12-03
Add workflow to test large runner

StevenJiang1110:


vvvvsv:
Is this PR intended to fix CI failures like [this](https://github.com/asterinas/asterinas/actions/runs/19885324824/job/56993691902?pr=2666)?

StevenJiang1110:
> Is this PR intended to fix CI failures like [this](https://github.com/asterinas/asterinas/actions/runs/19885324824/job/56993691902?pr=2666)?

This PR aims to determine if a larger runner can resolve the CI failures you point out. The current GitHub CI default storage seems inadequate for building a 3GB image. While the repository has enabled an ubuntu-2-cores-75GB-ssd runner, it hasn't helped so far. Since the standard runner also offers 75GB of storage, this test might not yield a different outcome. So this PR may not help.....

PR #2662, opened 2025-12-03
Add `i8042.exist` as a kernel cmdline option to override ACPI flags

lrh2000:
LENOVO 21EX002GCD hits this logic:
https://github.com/asterinas/asterinas/blob/6a87dad817595190c7e8e345fdd997630a227560/kernel/comps/i8042/src/controller.rs#L132-L141

The flag from the ACPI FADT does not seem to be very reliable. Even if the flag indicates that the i8042 controller is not present, the controller may actually be present and enumerable from other source, such as PnP devices (see [the Linux implementation](https://elixir.bootlin.com/linux/v6.18/source/drivers/input/serio/i8042-acpipnpio.h#L1578)).

But we don't currently support enumerating PnP devices. So if we want to make the i8042 controller work on LENOVO 21EX002GCD, we have the following four options:

 - **Option 1 (this PR):** Add a kernel command line option `i8042.exist` to override the ACPI flag. Linux has [a few similar options](https://elixir.bootlin.com/linux/v6.12.6/source/drivers/input/serio/i8042.c#L32-L50) like `i8042.nokbd`. But it does not have the exactly same option.
 - **Option 2:** Just disable the check, add a FIXME, and print a warning instead. It should be very simple and definitely work:
```patch
diff --git a/kernel/comps/i8042/src/controller.rs b/kernel/comps/i8042/src/controller.rs
index 4f4853239..9e01d5ca3 100644
--- a/kernel/comps/i8042/src/controller.rs
+++ b/kernel/comps/i8042/src/controller.rs
@@ -137,7 +137,9 @@ impl I8042Controller {
         {
             // The PS/2 controller does not exist. See:
             // <https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#ia-pc-boot-architecture-flags>.
-            return Err(I8042ControllerError::NotPresent);
+            //
+            // FIXME: (to be written...)
+            log::warn!("ACPI says i8042 controller does not exist")
         }
 
         let controller = Self {
```
 - **Option 3:** Do nothing. The user can patch our kernel to override the ACPI flag if needed.
 - **Option 4:** Implement PnP device enumeration similar to what Linux did. But due to the complexity, it cannot be done in a near future.

I'm also open to other options if they're preferable. I can switch to options 2 or 3 if they are considered better solutions. Of course, I cannot implement option 4 because it is too complex.

### DIFF
--- a/kernel/comps/i8042/src/controller.rs
+++ b/kernel/comps/i8042/src/controller.rs
@@ -5,6 +5,7 @@
 //! Reference: <https://wiki.osdev.org/I8042_PS/2_Controller>
 //!
 
+use aster_cmdline::{ModuleArg, KCMDLINE};
 use bitflags::bitflags;
 use ostd::{
     arch::{device::io_port::ReadWriteAccess, kernel::ACPI_INFO},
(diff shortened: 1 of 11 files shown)


kernel/comps/i8042/src/controller.rs
```diff
@@ -129,15 +130,43 @@ impl I8042Controller {
...
+            //
+            // However, it may actually be present and enumerable from other sources, such as PnP
+            // devices. See:
+            // <https://elixir.bootlin.com/linux/v6.18/source/drivers/input/serio/i8042-acpipnpio.h#L1578>.
+            //
+            // Currently, we lack the necessary support, so we allow the user to manually override
+            // the ACPI flag by appending "i8042.exist" to the kernel command line.
+            let is_present_cmdline = !KCMDLINE
+                .get()
+                .unwrap()
```
tatetian: Could you extract the kernel CLI argument parsing logic for the `is_present_cmdline` expression into a separate function? The new change has made the `new` method too complex.

kernel/comps/i8042/src/controller.rs
```diff
@@ -129,15 +130,43 @@ impl I8042Controller {
...
             // The PS/2 controller does not exist. See:
             // <https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#ia-pc-boot-architecture-flags>.
-            return Err(I8042ControllerError::NotPresent);
+            //
+            // However, it may actually be present and enumerable from other sources, such as PnP
+            // devices. See:
+            // <https://elixir.bootlin.com/linux/v6.18/source/drivers/input/serio/i8042-acpipnpio.h#L1578>.
+            //
+            // Currently, we lack the necessary support, so we allow the user to manually override
+            // the ACPI flag by appending "i8042.exist" to the kernel command line.
```
tatetian, line 142: Since this is a workaround, let's add a TODO here as a reminder.

tatetian (commented):
Adding this CLI argument is ok to me. Just two minor comments.

Issue #2649, opened 2025-11-30
Introducing a new ioctl infrastructure


tatetian:
This issue introduces a new ioctl infrastructure for Asterinas.

### Status quo

Currently, Asterinas has defined over 30 ioctl in a centralized, enum type called `IoctlCmd`:

https://github.com/asterinas/asterinas/blob/4040f2c77c3b020f672cb66032f18bd3b1329a9b/kernel/src/fs/utils/ioctl.rs#L8-L75

This approach is quite rudimentary and its shortcomings become increasingly obvious and intolerable as we add more module-specific ioctls. We have added the ioctl commands unique to `/dev/fb0` and `/dev/tdx_guest`. And there is a [pending PR](https://github.com/asterinas/asterinas/pull/2582) to add ioctl commands to `/dev/input/eventX`. In doing so, we have found that the current approach is not well suited for the following reasons.

**First, the status quo is unscalable and unsustainable.** Every module would have to edit this centralized enum `IoctlCmd` type to add its own ioctl commands. These ioctl commands are only interesting to this module but made visible to the entire kernel. This is obviously a bad thing from the perspective of abstraction and information hiding.

Having to edit this `IoctlCmd` type for every new ioctl command is also against the well-known ["open–closed principle"](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle), which states that "software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification". In other words, it promotes writing code that is so stable that you never have to change the existing source files to add new features.

Ideally, we want to establish some sort of ioctl infrastructure, based on which new ioctl commands can be defined by each module itself, without touching the common infrastructure.

**Second, the status quo does not support Linux's modern encoding.**  In Linux's old days, ioctl numbers were arbitrary constants. This led to collisions where the same number triggered different behaviors in different drivers. To solve this, Linux adopted a **standard 32-bit encoding scheme**. This structure embeds a driver (or module) magic number, the command sequence number, the data direction, the size of the data into the ioctl command integer. The scheme is summarized in the table below:

| **Bit Position** | **Field Name** | **Width (Standard)** | **Description**                                   |
| ---------------- | -------------- | -------------------- | ------------------------------------------------- |
| **30 - 31**      | **Direction**  | 2 bits               | Is data being read, written, or neither?          |
| **16 - 29**      | **Size**       | 14 bits              | The size of the argument being passed (in bytes). |
[...]

PR #2664, opened 2025-12-03
Rustfmt socket option macros

lrh2000:
See the second point in https://github.com/asterinas/asterinas/issues/2649#issuecomment-3604934414 (quoted below):
> One major downside is that the macro does not work with rustfmt. However, I just conducted some experiments and discovered a solution.
> 
> This version does not work with rustfmt:
> ```rust
>         match_sock_option_mut!(option, {
>             socket_errors: SocketError => {
>                 let a = Some(1i32);
>                 a.map(|val| val + 1).map(|val| val + 11231).map(|val| val + 23233).filter(|val| *val > 2333).expect("abcd");
>                 socket_errors.set(self.test_and_clear_error());
>                 return Ok(());
>             },
>             _ => ()
>         });
> ```
> But this version works:
> ```rust
>     sock_option_mut!(match option {
>         socket_errors @ SocketError => {
>             let a = Some(1i32);
>             a.map(|val| val + 1)
>                 .map(|val| val + 11231)
>                 .map(|val| val + 23233)
>                 .filter(|val| *val > 2333)
>                 .expect("abcd");
>             socket_errors.set(self.test_and_clear_error());
>             return Ok(());
>         }
>         _ => (),
>     });
> ```
> It seems that rustfmt only formats code that is written in valid Rust syntax. Otherwise it won't format anything inside the macro.

Note that this follows the [exact Rust syntax](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4b1cd07f50768c2228a89c6d64726efe):
```rust
let val = 3;
match val {
    n @ 1..=9 => n + 1,
    () => 0,
}
```

The downside might be:
```
        match $option:ident {
```
 - This can only be an `ident`. If we use `$option:expr`, there is an error:
```
error: `$option:expr` is followed by `{`, which is not allowed for `expr` fragments
  --> kernel/src/net/socket/options/macros.rs:69:28
   |
69 |         match $option:expr {
   |                            ^ not allowed after `expr` fragments
   |
   = note: allowed there are: `=>`, `,` or `;`
```

And:
```
            $( $bind:ident @ $ty:ty => $arm:block $(,)? )*
```
This can only be a `block`. If we use `$arm:expr`, there is an error:
```
error: `$arm:expr` may be followed by `_`, which is not allowed for `expr` fragments
  --> kernel/src/net/socket/options/macros.rs:71:13
   |
71 |             _ => $default:expr $(,)?
   |             ^ not allowed after `expr` fragments
   |
   = note: allowed there are: `=>`, `,` or `;`
```

For now, I don't think these downsides are having a negative impact.

### DIFF
--- a/kernel/src/net/socket/ip/datagram/mod.rs
+++ b/kernel/src/net/socket/ip/datagram/mod.rs
@@ -10,13 +10,12 @@ use super::addr::UNSPECIFIED_LOCAL_ENDPOINT;
 use crate::{
     events::IoEvents,
     fs::utils::Inode,
-    match_sock_option_mut,
     net::{
         iface::is_broadcast_endpoint,
         socket::{
             ip::options::{IpOptionSet, SetIpLevelOption},
             new_pseudo_inode,
-            options::{Error as SocketError, SocketOption},
+            options::{macros::sock_option_mut, Error as SocketError, SocketOption},
             private::SocketPrivate,
             util::{
                 datagram_common::{select_remote_and_bind, Bound, Inner},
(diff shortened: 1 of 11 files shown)


StevenJiang1110 (approved):
Formatting macros still appears to be [an open problem](https://github.com/rust-lang/rustfmt/issues/8) for rustfmt. If we can find a way to format them, it would certainly be an improvement. I can't see any drawbacks to this PR.

PR #2668, opened 2025-12-03
Free disk spaces before running nixos test

StevenJiang1110:
This PR addresses the NixOS CI test failures(such as [this run](https://github.com/asterinas/asterinas/actions/runs/19885910373/job/56995124627?pr=2665)) by implementing a disk space cleanup before test execution. This approach is similar to the method introduced in #2652.

It appears that creating a 3GB image within GitHub CI frequently encounters failures due to insufficient disk space. However, it's possible to free up significant space before the build process.


.github/workflows/test_nixos.yml
```diff
@@ -10,21 +10,37 @@ on:
...
+            cp qemu.log hello-asterinas.log
       - name: Basic nix-env test
+        uses: addnab/docker-run-action@v3
+        with:
+          image: asterinas/asterinas:0.16.1-20251130
+          options: --privileged -v /dev:/dev -v ${{ github.workspace }}:/root/asterinas
+          run: |
+            make run NIXOS=1 NIXOS_DISK_SIZE_IN_MB=4096 NIXOS_STAGE_2_INIT="nix-env -iA nixos.hello && hello"
+            cp qemu.log nix-env.log
+            make clean
```
vvvvsv: Why do we need to `make clean` here?

.github/workflows/test_nixos.yml
```diff
@@ -10,21 +10,37 @@ on:
 jobs:
   nixos-install-test:
     runs-on: ubuntu-latest
-    container:
-      image: asterinas/asterinas:0.16.1-20251130
-      options: --device=/dev/kvm --privileged -v /dev:/dev
     strategy:
       fail-fast: false
     steps:
+      - uses: endersonmenezes/free-disk-space@v3
```
tatetian: I am wondering if the current approach to NixOS CI tests is sustainable and an overall good deal. 

By "good deal", I mean that the workflow is written in an unconventional--even cumbersome--way compared to other workflows that we maintain just to save disk space. We have already enabled Github Action runners with larger disk space (`ubuntu-2-cores-75GB-ssd`), but this PR chose not to use it. So in essence, this PR trades less readable code and more development cost for reduced costs for paid runners. But I believe our development time is more valuable.

By "sustainable", I mean that although we can squeeze some more space (using `endersonmenezes/free-disk-space`), the disk space is still quite limited. Some day in a near future will hit the disk space limitation. If it is inevitable to switch to the larger, paid runner (`ubuntu-2-cores-75GB-ssd`) someday in the future, why not use it from now?

I am thinking about a hybrid approach to running NixOS CI tests. We write two workflows, one is named "Test NixOS distro (minimal)", the other "Test NixOS distro (full)". The first one only runs the most basic NixOS test, consumes the minimal disk space, and thus can run on free Github runners. This workflow will be triggered on all PRs. The other one includes most advanced NixOS tests, which install more applications and thus consume much more disk space. It is assigned to paid Github runners with larger disk space.

.github/workflows/test_nixos.yml
StevenJiang1110: > Why do we need to `make clean` here?

My initial concern was that `nix-env -iA nixos.hello` would modify the root filesystem content without subsequently synchronizing those changes to disk. Consequently, running `make run NIXOS=1` again might lead to unexpected issues, such as being unable to re-execute `nix-env -iA nixos.hello`. However, upon reflection, instead of using `make clean` here, a sync operation might be more appropriate, allowing subsequent tests to leverage cached results. I will change it accordingly.

.github/workflows/test_nixos.yml
StevenJiang1110: >  If it is inevitable to switch to the larger, paid runner (ubuntu-2-cores-75GB-ssd) someday in the future, why not use it from now?

Yes, the problem is that `ubuntu-2-cores-75GB-ssd` exactly has the same size of storage as the standard runner. Maybe I can use the newly-created 150GB runner.

> I am thinking about a hybrid approach to running NixOS CI tests. We write two workflows, one is named "Test NixOS distro (minimal)", the other "Test NixOS distro (full)".

I agree with this approach. The application test will consume too much disk space, and we'd better adopt different runners.

> The first one only runs the most basic NixOS test, consumes the minimal disk space, and thus can run on free Github runners.

The problem is that even the simplest tests fail to run on free GitHub runners without first freeing up disk space. Attempting to allocate a 3GB image on a free GitHub runner results in random failures, though it often succeeds. Should these tests be moved to larger runners, and should they be run on every PR?

I admit the current workflow implementation is somewhat cumbersome. This is primarily due to the peculiar behavior of the docker run action currently in use: if a command executed within the container fails, the job itself does not reliably report a failure. Consequently, I'm forced to copy `qemu.log` out of the container and check its results externally. If it's acceptable to proceed with the 'free disk space first' approach, I can also investigate alternative Docker actions that provide more robust error reporting.

tatetian: > I admit the current workflow implementation is somewhat cumbersome.

What if only one most basic NixOS test runs on the standard runner? Will this restriction simplify the workflow?

> Should these tests be moved to larger runners, and should they be run on every PR?

The basic NixOS CI test on free runners can be triggered on all PRs.

The advanced NixOS CI tests on paid runners should be triggered selectively when relevant files are changed.

.github/workflows/test_nixos_minimal.yml
```diff
@@ -0,0 +1,33 @@
+name: Test NixOS distro (minimal)
```
tatetian: Add a comment to explain the differences between the minimal and full versions. This is probably the first question that pops up in one's head when reading the code.

```yaml
# A basic CI test for NixOS.
#
# This workflow is intended to include the most basic test for AsterNixOS.
# It requires only a minimal amount of disk space
# and thus can be run on standard Github runners _for free_.
name: Test AsterNixOS (minimal)
```

In addition, change "NixOS distro" to "AsterNixOS". The latter one is more specific and accurate.

.github/workflows/test_nixos_minimal.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: endersonmenezes/free-disk-space@v3
+        with:
+          rm_cmd: "rmz"
+          remove_android: true
+          remove_dotnet: true
+          remove_haskell: true
+          remove_tool_cache: true
+      - uses: actions/checkout@v4
+      - name: Basic NixOS test
```
tatetian: Rename this to "Run a Hello World program on AsterNixOS"

.github/workflows/test_nixos_full.yml
```diff
@@ -0,0 +1,34 @@
+name: Test NixOS distro (full)
```
tatetian: Update this name field accordingly

.github/workflows/test_nixos_minimal.yml
```diff
@@ -0,0 +1,33 @@
...
+    branches:
+      - main
+
+jobs:
+  minimal-nixos-test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: endersonmenezes/free-disk-space@v3
```
tatetian, line 27: Add a comment to explain why this needs to be done and how much space it can free. This is some knowledge that you learned in the hard way. It should be passed on to other developers.

.github/workflows/test_nixos_full.yml
```diff
@@ -0,0 +1,34 @@
...
+      - tools/nixos/**
+
+jobs:
+  full-nixos-test:
+    runs-on: ubuntu-4-cores-150GB-ssd
+    container:
+      image: asterinas/asterinas:0.16.1-20251130
+      options: -v /dev:/dev --privileged
+    strategy:
+      fail-fast: false
```
tatetian: Roll back to take the default value (false) to save runner minutes

.github/workflows/test_nixos.yml
StevenJiang1110: Fixed as suggested.

.github/workflows/test_nixos_full.yml
```diff
@@ -0,0 +1,34 @@
...
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+      - name: Basic nix-env test
```
tatetian: These are names for _steps_, which are actions. So name them by starting with verbs.

```
- name: Use nix-env to install and run the hello package
```

tatetian (commented):
Mostly good. Just a few minor issues.

PR #2671, opened 2025-12-04
Rename `syscall-feature-coverage` to `syscall-flag-coverage`

taosue:
Rename `Syscall Feature Coverage` to `Syscall Flag Coverage` to better reflect that we're documenting flag coverage rather than feature.


tatetian (approved):

PR #2673, opened 2025-12-04
Add safety requirements for `read` and `write` in `IoApicAccess`

Unparalleled-Calvin:
This PR adds missing safety comments to the `read` and `write` unsafe functions in `IoApicAccess`. The added safety comments focus only on the current implemented functionality and do not attempt to mention:
- The IOAPICARB register (not relevant to current implementation)
- The IOAPICID register in writable register handling (edge case)

### DIFF
--- a/ostd/src/arch/x86/irq/chip/ioapic.rs
+++ b/ostd/src/arch/x86/irq/chip/ioapic.rs
@@ -206,6 +206,16 @@ impl IoApicAccess {
         Self { io_mem }
     }
 
+    /// # Safety
+    ///
+    /// 1. ValidReg: The value of `register` should be a valid IOAPIC readable register index.
+    ///    Valid registers include:
+    ///    - ID register (0x00)
+    ///    - Version register (0x01)
+    ///    - Redirection table entries:
+    ///      * Low 32 bits:  0x10 + 2*n
+    ///      * High 32 bits: 0x10 + 2*n + 1
+    ///    Where 0 <= n <= max redirection entry (obtained from the version register)
     pub(self) unsafe fn read(&mut self, register: u8) -> u32 {
         // SAFETY: This reads data from an I/O APIC register. The safety is upheld by the caller.
         unsafe {
@@ -215,6 +225,19 @@ impl IoApicAccess {
         }
     }
 
+    /// # Safety
+    ///
+    /// 1. ValidReg: The value of `register` should be a valid IOAPIC writable register index.
+    ///    Valid registers include:
+    ///    - Redirection table entries:
+    ///      * Low 32 bits:  0x10 + 2*n
+    ///      * High 32 bits: 0x10 + 2*n + 1
+    ///    Where 0 <= n <= max redirection entry (obtained from the version register)
+    /// 
+    /// 2. ValidRegValue: data must be a valid value for register to ensure correct APIC operation and system safety.
+    ///    See the following links for valid bit configurations:
+    ///    - IOREDTBL structure: https://wiki.osdev.org/IOAPIC#IOREDTBL
+    ///    - Full specification: https://pdos.csail.mit.edu/6.828/2016/readings/ia32/ioapic.pdf
     pub(self) unsafe fn write(&mut self, register: u8, data: u32) {
         // SAFETY: This writes data to an I/O APIC register. The safety is upheld by the caller.
         unsafe {


ostd/src/arch/x86/irq/chip/ioapic.rs
```diff
@@ -206,6 +206,16 @@ impl IoApicAccess {
         Self { io_mem }
     }
 
+    /// # Safety
+    ///
+    /// 1. ValidReg: The value of `register` should be a valid IOAPIC readable register index.
```
lrh2000, line 211:
```suggestion
    /// 1. The value of `register` must be a valid IOAPIC readable register index.
```
 - Safety tags require a separate discussion. Could we please not squash it here? Even when checking the changes in this PR, I'm confused. `ValidReg` in `read` and `ValidReg` in `write` obviously mean different things, yet they use the same identifier. Won't that cause confusion? Also, other devices have different requirements. A tag can't simplify this because they all have different semantics.
 - Use `must` instead of `should` when writing safety comments. Also, `data must be a valid value` appears below, which uses `must`. We should at least be consistent.

ostd/src/arch/x86/irq/chip/ioapic.rs
```diff
@@ -206,6 +206,16 @@ impl IoApicAccess {
...
+    ///    Valid registers include:
+    ///    - ID register (0x00)
+    ///    - Version register (0x01)
```
lrh2000, line 214:
```rust
    ///    - ID register ([`Self::IOAPICID`])
    ///    - Version register ([`Self::IOAPICVER`])
```

ostd/src/arch/x86/irq/chip/ioapic.rs
```diff
@@ -206,6 +206,16 @@ impl IoApicAccess {
...
+    ///    - Redirection table entries:
+    ///      * Low 32 bits:  0x10 + 2*n
+    ///      * High 32 bits: 0x10 + 2*n + 1
+    ///    Where 0 <= n <= max redirection entry (obtained from the version register)
```
lrh2000, line 218:
```rust
    ///    - Redirection table entries (where `index` is in `0..=self.max_redirection_entry`):
    ///      * Low 32 bits:  `Self::IOREDTBL + 2 * index`
    ///      * High 32 bits: `Self::IOREDTBL + 2 * index + 1`
```

ostd/src/arch/x86/irq/chip/ioapic.rs
```diff
@@ -215,6 +225,19 @@ impl IoApicAccess {
...
+    /// # Safety
+    ///
+    /// 1. ValidReg: The value of `register` should be a valid IOAPIC writable register index.
+    ///    Valid registers include:
+    ///    - Redirection table entries:
+    ///      * Low 32 bits:  0x10 + 2*n
+    ///      * High 32 bits: 0x10 + 2*n + 1
+    ///    Where 0 <= n <= max redirection entry (obtained from the version register)
+    /// 
+    /// 2. ValidRegValue: data must be a valid value for register to ensure correct APIC operation and system safety.
```
lrh2000, line 237:
```suggestion
    /// 2. `data` must be a valid value for `register` to ensure correct IOAPIC operation and system safety.
```

Also this line seems to long. So it should be:
```rust
    /// 2. `data` must be a valid value for `register` to ensure correct IOAPIC operation and
    ///    system safety. See the following links for valid bit configurations:
```

ostd/src/arch/x86/irq/chip/ioapic.rs
```diff
@@ -215,6 +225,19 @@ impl IoApicAccess {
...
+    ///    See the following links for valid bit configurations:
+    ///    - IOREDTBL structure: https://wiki.osdev.org/IOAPIC#IOREDTBL
+    ///    - Full specification: https://pdos.csail.mit.edu/6.828/2016/readings/ia32/ioapic.pdf
```
lrh2000, line 240:
```rust
    ///    - IOREDTBL structure: <https://wiki.osdev.org/IOAPIC#IOREDTBL>
    ///    - Full specification: <https://pdos.csail.mit.edu/6.828/2016/readings/ia32/ioapic.pdf>
```

lrh2000 (commented):
Adding safety comments for the I/O APIC in this way is okay, although I am generally skeptical of this approach.

It's fine for I/O APIC because the device is very simple. Therefore, repeating the register definition three times seems good. Note that we've already defined the registers once in our current codebase using the `IOAPICID`, `IOAPICID`, and `IOREDTBL` constants. This PR adds two more reparations to the safety requirements of the `read` and `write` functions.

When it comes to more complex interrupt controllers, such as the [GICv3 for ARM](https://www.scs.stanford.edu/~zyedidia/docs/arm/gic_v3.pdf), there are hundreds of registers. I don't think that repeating the register list three times is a good idea.

tatetian:
> Note that we've already defined the registers once in our current codebase using the `IOAPICID`, `IOAPICID`, and `IOREDTBL` constants. This PR adds two more reparations to the safety requirements of the `read` and `write` functions.
> 
> When it comes to more complex interrupt controllers, such as the [GICv3 for ARM](https://www.scs.stanford.edu/~zyedidia/docs/arm/gic_v3.pdf), there are hundreds of registers. I don't think that repeating the register list three times is a good idea.

One solution is to reorganize the code so that a concept is represented in the code as a single language construct, which could be a const, an enum variant, or a type. This way, we can attach the safety comment for the concept to the single, corresponding language construct and thus avoiding duplicating safety comments.

tatetian:
I get an increasingly stronger feeling that the safety-tag team needs more help from the development team to incorporate the safety tags in a satisfying way...

ostd/src/arch/x86/irq/chip/ioapic.rs
hxuhack: These are not the safety tags we proposed. Instead, they are merely abbreviations of the respective safety properties, added to make the comments more structured and easier for developers to read. Do you find them intrusive? Please note that the actual safety tags we proposed are intended to be implemented as Rust attributes, which may affect the current development practice. The abbreviations can help pave the way for introducing full safety tags.

hxuhack:
> > Note that we've already defined the registers once in our current codebase using the `IOAPICID`, `IOAPICID`, and `IOREDTBL` constants. This PR adds two more reparations to the safety requirements of the `read` and `write` functions.
> > When it comes to more complex interrupt controllers, such as the [GICv3 for ARM](https://www.scs.stanford.edu/~zyedidia/docs/arm/gic_v3.pdf), there are hundreds of registers. I don't think that repeating the register list three times is a good idea.
> 
> One solution is to reorganize the code so that a concept is represented in the code as a single language construct, which could be a const, an enum variant, or a type. This way, we can attach the safety comment for the concept to the single, corresponding language construct and thus avoiding duplicating safety comments.

Yes, I think this is a problem that can be addressed by using safety-tag attributes. Developers can specify the safety tags and related concepts in a TOML file, and using a safety tag would then be similar to invoking a function. Our tool can automatically look up the TOML file and generate the corresponding Rust documentation. Before safety tags are fully adopted, we can create an introductory page and link each concept to that page, following the current practice of the Rust standard library.

ostd/src/arch/x86/irq/chip/ioapic.rs
zjp-CN: > ValidReg in read and ValidReg in write obviously mean different things, yet they use the same identifier.

Yeah, they should be ReadableReg and WritableReg respectively.

> other devices have different requirements

It's necessary to document what IO devices have been supported for this function to accept. We're documenting such one here. If more devices are added in the future, they should be documented too.

lrh2000: > Do you find them intrusive?

I found it odd because it is not an existing practice in either our codebase or the Rust standard library.

However, I admit it's a chicken-and-egg problem. Proposing a new syntax to document things always seems to be the case.

> It's necessary to document what IO devices have been supported for this function to accept. We're documenting such one here. If more devices are added in the future, they should be documented too.

This is not what I mean. If we introduce identifiers, we should have more rules to manage the identifiers. It is not clear to me whether we should use the same identifier for other devices and how these identifiers can be managed properly.

tatetian: I discussed with @hxuhack offline about introducing abbreviations/names (“安全缩略词”) for safe comments. We agreed on the following strategy in my own words:

> 这个我觉得过程大概是这样，第一阶段就专注于完善safety comments，当写了足够多的safety comments之后，某些模式开始清晰的浮现时，进入第二阶段，引入“安全缩略词”。这一定是两阶段的，因为“安全缩略词”说白了就是一个归纳法的思考过程。只写了几个safety comments是没有足够的样本支撑提炼出“安全缩略词”的。

PR #2678, opened 2025-12-04
Correct zero padding behavior in ELF loading

junyang-zh:
In Linux, only the `.bss` segment needs zero padding (no zero padding if the beginning of the segment is not aligned, or at the other end of other segments).

Reference: https://elixir.bootlin.com/linux/v6.17/source/fs/binfmt_elf.c#L410-L422

Originally, filling zero paddings on ELF loading was implemented by manually mapping the frames with cursors. This is a surprise for the `Vmar` implementation because it bypasses all the RSS counting, PF/COW handling, etc. This PR uses `fill_zero_remote` (which is similar to `write_remote`) and does the CoW work correctly.

In Linux, it seems that the VM is activated before the program gets loaded, while in our case, it's not guaranteed. So this PR doesn't use fallible writes but uses remote access. Therefore, the error handling is a little bit different. Linux always returns `EFAULT`, but we can return `ENOMEM`. I believe no one cares about it.


junyang-zh:
@StevenJiang1110 @cchanging, ping for reviews.

PR #2679, opened 2025-12-05
Add `/proc/version` support

enkerewpo:
image

### DIFF
--- a/kernel/src/fs/procfs/version.rs
+++ b/kernel/src/fs/procfs/version.rs
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! This module offers `/proc/version` file support, which provides
+//! information about the kernel version.
+//!
+//! Reference: <https://man7.org/linux/man-pages/man5/proc_version.5.html>
+
+use aster_util::printer::VmPrinter;
+
+use crate::{
+    fs::{
+        procfs::template::{FileOps, ProcFileBuilder},
+        utils::{mkmod, Inode},
+    },
+    net::UtsNamespace,
+    prelude::*,
+};
+
+pub struct VersionFileOps;
+
+impl VersionFileOps {
+    pub fn new_inode(parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        ProcFileBuilder::new(Self, mkmod!(a+r))
+            .parent(parent)
+            .build()
+            .unwrap()
+    }
+}
+
+impl FileOps for VersionFileOps {
+    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        let mut printer = VmPrinter::new_skip(writer, offset);
+
+        // Get UTS namespace information from the init namespace
+        let uts_name = UtsNamespace::get_init_singleton().uts_name();
+
+        let sysname = uts_name.sysname()?;
+        let release = uts_name.release()?;
+        let version = uts_name.version()?;
+
+        // Get info from compile environment variables
+        let compile_by = option_env!("OSDK_BUILD_USERNAME").unwrap_or("unknown");
+        let compile_host = option_env!("OSDK_BUILD_HOSTNAME").unwrap_or("unknown");
+        let compiler = option_env!("OSDK_BUILD_RUSTC").unwrap_or("unknown");
+
+        // Reference:
+        // <https://elixir.bootlin.com/linux/v6.17/source/init/version.c>
+        // <https://elixir.bootlin.com/linux/v6.17/source/fs/proc/version.c>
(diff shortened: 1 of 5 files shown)


kernel/src/fs/procfs/version.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn new_inode(parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        ProcFileBuilder::new(Self, mkmod!(a+r))
+            .parent(parent)
+            .build()
+            .unwrap()
+    }
+
+    fn collect_version() -> Result<String> {
+        // Get UTS namespace information from the init namespace
+        let uts_name = UtsNamespace::get_init_singleton().uts_name();
```
StevenJiang1110: Instead of using information from the init UTS namespace, I think you should get information from the UTS namespace of current thread. The code may look like: 

```rust
let current_thread = current_thread!();
let posix_thread = current_thread.as_posix_thread().unwrap();
let ns_proxy = posix_thread.ns_proxy().lock();
let Some(ns_proxy) = ns_proxy.as_ref() else {
    // Return some error here like ESRCH?
};
let uts_ns = ns_proxy.uts_ns();
......
```
You can refer to the Linux implementation of `uts_name`.
https://elixir.bootlin.com/linux/v6.0.9/source/fs/proc/version.c#L12

osdk/src/commands/build/mod.rs
```diff
@@ -231,6 +231,36 @@ fn build_kernel_elf(
 
     let mut command = cargo();
     command.env_remove("RUSTUP_TOOLCHAIN");
+    command.env(
+        "ASTER_COMPILE_BY",
```
tatetian: OSTD and OSDK are positioned as a crate and tool for Rust OS development in general, not specific to Asterinas, although they origin from Asterinas. So these environment variable names should not be prefixed with "ASTER_". I suggest renaming them to:
* `ASTER_COMPILE_BY` --> `OSDK_BUILD_USERNAME`
* `ASTER_COMPILE_HOST ` --> `OSDK_BUILD_HOSTNAME`
* `ASTER_COMPILER ` --> `OSDK_BUILD_RUSTC`
* `ASTER_COMPILE_VERSION ` --> `OSDK_BUILD_TIMESTAMP`

kernel/src/fs/procfs/version.rs
```diff
@@ -0,0 +1,61 @@
...
+        let compile_by = option_env!("ASTER_COMPILE_BY").unwrap_or("unknown");
+        let compile_host = option_env!("ASTER_COMPILE_HOST").unwrap_or("unknown");
+        let compiler = option_env!("ASTER_COMPILER").unwrap_or("unknown");
+
+        // Reference:
+        // <https://elixir.bootlin.com/linux/v6.17/source/init/version.c>
+        // <https://elixir.bootlin.com/linux/v6.17/source/fs/proc/version.c>
+        Ok(format!(
+            "{} version {} ({}@{}) ({}) {}\n",
+            sysname, release, compile_by, compile_host, compiler, version
```
tatetian, line 52: Do not create the `String`. Write directly to the `printer`. The latter approach is more efficient.

kernel/src/net/uts_ns.rs
```diff
@@ -100,6 +120,18 @@ impl UtsNamespace {
 }
 
 const UTS_FIELD_LEN: usize = 65;
+macro_rules! uts_field_getter {
```
tatetian: I am afraid that the macro is a bit of an overkill. Rust macros are powerful, but they also have some pitfalls, e.g., incompatibility with Rust format and poor readability. So we want to use them wisely, when the benefits of using Rust macros clearly outweigh the drawbacks.

```rust
impl UtsName {
    pub fn sysname(&self) -> Result<&str> {
        Self::cstr_bytes_to_str(&self.sysname)
    }

    pub fn release(&self) -> Result<&str> {
        Self::cstr_bytes_to_str(&self.release)
    }

    pub fn version(&self) -> Result<&str> {
        Self::cstr_bytes_to_str(&self.version)
    }

    fn cstr_bytes_to_str(cstr_bytes: &[u8]) -> &str {
        CStr::from_bytes_until_nul(cstr_bytes)
            .map_err(|_| Error::with_message(Errno::EINVAL, "not a null-terminated C string"))?
            .to_str()
            .map_err(|_| Error::with_message(Errno::EINVAL, "not a UTF-8 string"))?
    }
}
```

The resulted code is longer. But it is actually easier for a human to read (and potentially edit) than the original macro-based version.

enkerewpo:
I've updated the related part, thanks for your comments!

tatetian (approved):

PR #2680, opened 2025-12-05
pci: Improve PCI device enumeration and add detection of different PCI device

sdww0:
This PR aims to optimize PCI device enumeration and improve system boot time.

Recently, I attempted to boot Asterinas on Firecracker and measured the startup time — from launching the Firecracker binary until the kernel’s user-space program printed “hello_world” — was about `459ms` (with `INITRAMFS_SKIP_GZIP`=1). After removing PCI bus traversal, boot time dropped to around `88ms`. Therefore, I plan to optimize the current brute-force PCI device scanning approach and refactor the existing PCI code.

This PR is a first step. It introduces the detection of multi‑function capability in PCI devices to decide whether to scan subsequent functions. After this PR, boot time will drop to `133ms`.

Additional changes include:
1. Split the configuration space offset into generic and device-specific offsets
2. Added `PciDeviceType` to declare different PCI devices (General devices, PCI-to-PCI bridges, PCI-to-Cardbus bridges)
3. Refactored portions of the code, including `BarManager`

### DIFF
--- a/kernel/comps/pci/src/common_device.rs
+++ b/kernel/comps/pci/src/common_device.rs
@@ -52,6 +52,11 @@ impl PciCommonDevice {
         self.header_type.device_type()
     }
 
+    /// Checks whether the device is a multi-function device
+    pub fn has_multi_funcs(&self) -> bool {
+        self.header_type.has_multi_funcs()
+    }
+
     /// Gets the PCI Command
     pub fn command(&self) -> Command {
         Command::from_bits_truncate(self.location.read16(PciCommonCfgOffset::Command as u16))
--- a/kernel/comps/pci/src/device_info.rs
+++ b/kernel/comps/pci/src/device_info.rs
@@ -19,33 +19,17 @@ impl PciDeviceLocation {
     // TODO: Find a proper way to obtain the bus range. For example, if the PCI bus is identified
     // from a device tree, this information can be obtained from the `bus-range` field (e.g.,
     // `bus-range = <0x00 0x7f>`).
-    const MIN_BUS: u8 = 0;
+    pub const MIN_BUS: u8 = 0;
     #[cfg(not(target_arch = "loongarch64"))]
-    const MAX_BUS: u8 = 255;
+    pub const MAX_BUS: u8 = 255;
     #[cfg(target_arch = "loongarch64")]
-    const MAX_BUS: u8 = 127;
+    pub const MAX_BUS: u8 = 127;
 
-    const MIN_DEVICE: u8 = 0;
-    const MAX_DEVICE: u8 = 31;
+    pub const MIN_DEVICE: u8 = 0;
+    pub const MAX_DEVICE: u8 = 31;
 
-    const MIN_FUNCTION: u8 = 0;
-    const MAX_FUNCTION: u8 = 7;
-
-    /// Returns an iterator that enumerates all possible PCI device locations.
-    pub fn all() -> impl Iterator<Item = PciDeviceLocation> {
-        let all_bus = Self::MIN_BUS..=Self::MAX_BUS;
-        let all_dev = Self::MIN_DEVICE..=Self::MAX_DEVICE;
-        let all_func = Self::MIN_FUNCTION..=Self::MAX_FUNCTION;
-
-        all_bus
-            .flat_map(move |bus| all_dev.clone().map(move |dev| (bus, dev)))
-            .flat_map(move |(bus, dev)| all_func.clone().map(move |func| (bus, dev, func)))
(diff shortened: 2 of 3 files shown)


kernel/comps/pci/src/cfg_space.rs
```diff
@@ -30,8 +30,12 @@ pub enum PciDeviceCommonCfgOffset {
     Status = 0x06,
     /// Revision ID
     RevisionId = 0x08,
+    /// Prog IF
+    ProgIf = 0x09,
+    /// Subclass
+    Subclass = 0x0A,
     /// Class code
-    ClassCode = 0x09,
+    ClassCode = 0x0B,
```
junyang-zh: https://en.wikipedia.org/wiki/PCI_configuration_space#/media/File:Pci-config-space.svg

The class code should be 0x9 0xA 0xB, 3 bytes. We are using 0xA as a subclass and 0xB as a class code. Could you please elaborate on why we do so?

sdww0: [The oswiki](https://wiki.osdev.org/PCI#Common_Header_Fields) further splits the class code into three types: `class`, `subclass`, and `prog_if`. 

I further checked the introduction of this field in the PCI specification:

image

Update: I think we can change the `ClassCode` to `BaseClassCode` to distinguish them.

kernel/comps/pci/src/common_device.rs
```diff
@@ -8,61 +8,107 @@ use alloc::vec::Vec;
...
+    /// PCI-to-PCI bridge (header type 0x01).
+    /// Contains the primary, secondary, and subordinate bus numbers.
+    PciToPciBridge(u8, u8, u8),
+    /// PCI-to-Cardbus bridge (header type 0x02).
+    PciToCardbusBridge,
+}
+
+impl PciDeviceType {
+    /// Creates a new `PciDeviceType` based on the header type read from the PCI device's configuration space.
+    pub fn new(location: &PciDeviceLocation) -> Option<Self> {
```
tatetian: It does not make sense to expose this method as a public constructor. Refactor the signature as follows:

```rust
impl PciDeviceType {
    pub fn try_from_raw(raw: u8) -> Option<Self> { .. }
}
```

Move the register read operation from `PciDeviceType::new` to `PciCommonDevice::new`.

kernel/comps/pci/src/common_device.rs
```diff
@@ -8,61 +8,107 @@ use alloc::vec::Vec;
...
         &self.capabilities
     }
 
+    /// Gets the PCI device type
+    pub fn device_type(&self) -> PciDeviceType {
+        self.device_type
+    }
+
+    /// Checks whether the device is a multi-function device
+    pub fn has_multi_function(&self) -> bool {
```
tatetian: Simplify the name to `has_multi_funcs`.

kernel/comps/pci/src/common_device.rs
```diff
@@ -8,61 +8,107 @@ use alloc::vec::Vec;
...
+        (self.location.read8(PciCommonCfgOffset::HeaderType as u16) & 0x80) != 0
```
tatetian: We have read the header type before to determine the device type. It does not make sense to read the information again. To do so, we can replace the `device_type` field of `PciCommonDevice` with the header type:

```rust
pub struct PciCommonDevice {
    device_id: PciDeviceId,
    location: PciDeviceLocation,
    header_type: PciHeaderType,
    bar_manager: BarManager,
    capabilities: Vec<Capability>,
}

impl PciCommonDevice {
	pub fn device_type(&self) -> PciDeviceType {
        self.header_type.device_type
    }

	pub fn has_multi_funcs(&self) -> bool {
        self.header_type.has_multi_funcs
    }
}

/// The header type field of a PCI device struct in the PCI configuration space.
///
/// A header type is comprised of two pieces of information:
/// 1. The device type ([`PciDeviceType`]);
/// 2. Whether the device has multiple functions (`bool`).
#[derive(Copy, Clone, Eq, PartialEq)]
struct PciHeaderType {
	pub device_type: PciDeviceType,
	pub has_multi_funcs: bool,
}

impl PciHeaderType {
    /// Converts a byte into a header type.
    ///
    /// According to the PCI spec, the encoding of a header type is as follows:
    /// * Bit 0-6 encodes the raw value of `PciDeviceType`;
    /// * Bit 7 indicates whether the PCI device has multiple functions.
	pub fn try_from_raw(raw: u8) -> Option<Self> {
		todo!()
	}

	pub fn device_type(self) -> PciDeviceType { todo!() }

	pub fn has_multi_funcs(self) -> bool { todo!() }
}

pub enum PciDeviceType {
	// As you implemented
}
```

Issue #2672, opened 2025-12-04
Tracking issue for nix commands


vvvvsv:
This issue tracks the remaining gaps needed to run `nix-env`, `nix-shell`, `nix-build`, and `nix-rebuild` in Asterinas NixOS.

# nix-env

`nix-env` is now running correctly. We need to avoid the warning:
```
GC Warning: pthread_getattr_np or pthread_attr_getstack failed for main thread
```

This requires:
 - [x] Adding `/proc/<pid>/maps`. Reference: [pthread_getattr_np](https://man7.org/linux/man-pages/man3/pthread_getattr_np.3.html) @vvvvsv  https://github.com/asterinas/asterinas/pull/2725

# nix-shell

This requires:
 - [x] Use log in shell to correctly set environment variables. @vvvvsv  https://github.com/asterinas/asterinas/pull/2665

# nix-build

This requires:
 - [x] Support the `fchmodat2(AT_SYMLINK_NOFOLLOW)` syscall. @vvvvsv https://github.com/asterinas/asterinas/pull/2666
 - Then, there are two options:
   - Skip some build phases
     ```
      pkgs.stdenv.mkDerivation {
        # ...
        dontPatchELF = 1;
        dontPatchShebangs = 1;
        dontStrip = 1;
        dontCheckReferences = 1;
        dontCheckForBrokenSymlinks = 1;
        dontRewriteSymlinks = 1;
      }
     ```
   - **OR** Full support
     - [x] Add symlink `/dev/fd -> /proc/self/fd`. Can dummy support this by adding `ln -s /proc/self/fd /dev/fd
`  to `postBootCommands`. @vvvvsv https://github.com/asterinas/asterinas/pull/2716
     - [x] **Bypassed.** Files under `/proc/<pid>` should have the owner set to the corresponding process owner, not root. Can bypass this by `nix.settings.build-users-group = [];` @vvvvsv https://github.com/asterinas/asterinas/pull/2710
     - [x] If `script.sh` has no shebang, `execve(script.sh)` should return `ENOEXEC`. @vvvvsv https://github.com/asterinas/asterinas/pull/2681
     - [x] Support reopening anonymous pipes via `/proc/<pid>/fd/<fd>`. @vvvvsv https://github.com/asterinas/asterinas/pull/2694
     - [x] Running some simple scripts with Bash ends up accessing address 0x8. I don't know why. @wyt8  https://github.com/asterinas/asterinas/pull/2700
 
# nixos-rebuild test

[...]

PR #2682, opened 2025-12-05
Support syscall `syncfs`

vvvvsv:
This is required for supporting `nixos-rebuild` in Asterinas NixOS. https://github.com/asterinas/asterinas/issues/2672

Reference: https://man7.org/linux/man-pages/man2/syncfs.2.html


tatetian (approved):

PR #2685, opened 2025-12-05
Add cachix binary cache

cqs21:


.github/workflows/push_cachix_pkgs.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build Asterinas NixOS patched packages (dry run)
+        if: github.event_name == 'pull_request'
+        run: |
+          make kernel BOOT_PROTOCOL=linux-efi-handover64 || true
+          make cachix
+
+      - name: Build Asterinas NixOS patched packages and push to devlopment cache
```
tatetian: devlopment --> development

.github/workflows/push_cachix_dev.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          make kernel BOOT_PROTOCOL=linux-efi-handover64 || true
+          make cachix
+
+      - name: Build Asterinas NixOS patched packages and push to devlopment cache
+        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+        run: |
+          make kernel BOOT_PROTOCOL=linux-efi-handover64 || true
+          export CACHIX_AUTH_TOKEN=${{ secrets.CACHIX_AUTH_TOKEN_FOR_DEV_CACHE }}
+          make push_cachix 2>&1 | tee cachix.log || true
```
tatetian, line 37: We never use the token for the release cache. Do you plan to push the release artifacts in another workflow? If so, this workflow should be named `push_cachix_dev`.

```
name: Push packages to AsterNixOS's development cache on Cachix
```

tatetian:
I don't know why your PR does not receive Copilot reviews automatically. So I manually ask Copilot for review comments. It seems that Copilot is more professional than I am on Nix-related stuff. Below are comments from Copilot.

----

Quick summary
- Overall this adds support for building a closure of important Nix outputs and pushing them to Cachix (workflow + Makefile + new Nix derivation) and wires the release/dev Cachix substituters + trusted keys through the installer/ISO build.
- Good direction: copying files into the installer output (cp -L / cp -r) makes the installer derivation self-contained; the Makefile target and push logic are sensible.
- There are a few correctness, robustness, and style issues (some of them likely to cause build/runtime failures). Below I list problems, why they matter, and concrete fixes/suggestions.

Blocking / high-priority issues
1) Invalid attribute access (hyphenated attribute) in distro/etc_nixos/modules/core.nix
- Problem: you use config.aster_nixos.trusted-public-keys (unquoted) which is invalid Nix syntax (attribute names with hyphens must be quoted).
- Why it matters: Nix parser will error and the configuration derivation will fail.
- Fix (one of):
  - Quote the attribute: config.aster_nixos."trusted-public-keys"
  - Or better: rename the option everywhere to a hyphen-free name such as trustedPublicKeys (recommended for simpler dot-access).
Example change:
  substituters = if config.aster_nixos.substituters == "" then [] else lib.strings.split " " config.aster_nixos.substituters;
  trusted-public-keys -> config.aster_nixos."trusted-public-keys"  (or rename as above)

2) substituters / trusted-public-keys are passed as a single space-separated string but used as a single list element
- Problem: you pass "--argstr extra-substituters '${RELEASE_SUBSTITUTER} ${DEV_SUBSTITUTER}'" and later do
  nix.settings.substituters = [ "${config.aster_nixos.substituters}" ];
  That produces a list with one element equal to the whole space-separated string. Nix wants a list of substituter URLs (each element a URL) in the configuration.
- Why it matters: Nix may try to use the single string as one substituter which will be invalid; the substituter list will not be parsed correctly and binary cache access may fail.
- Fix: convert the space-separated string into a proper list inside Nix (or pass a list from the caller).
Option A (convert inside Nix, robust):
  let
    substitutersList = if config.aster_nixos.substituters == "" then [] else lib.strings.split " " config.aster_nixos.substituters;
    trustedKeysList = if config.aster_nixos."trusted-public-keys" == "" then [] else lib.strings.split " " config.aster_nixos."trusted-public-keys";
  in
  nix.settings = {
    ...
    substituters = substitutersList;
    trusted-public-keys = trustedKeysList;
  };
Notes:
  - Use lib.strings.split (or builtins.split if you prefer builtin). I recommend lib.strings.split for clarity in nixpkgs code.
  - If you rename "trusted-public-keys" to trustedPublicKeys, update the code accordingly.

Medium / important fixes and checks
3) distro/cachix/default.nix: verify pkgs.writeClosure usage and list contents
- Concerns:
  - I don't recall pkgs.writeClosure being a standard helper available everywhere in nixpkgs. Confirm that function exists in your pkgs set and does what you expect.
  - The intent appears to be producing a derivation whose output is a file containing store paths to push; make sure the produced file is a newline-separated list of store paths suitable for `cachix push < cachix.list`.
  - Several package names you include may not exist exactly as referenced (e.g. xfce.xfdesktop/xfce.xfwm4, runc.man, podman.man — confirm these attributes exist in the nixpkgs you target).
- Suggestions:
  - Explicit, simple approach that is easier to review/test: create a derivation that writes a text file containing the store paths:
    pkgs.writeTextFile {
      name = "cachix.list";
      text = builtins.concatStringsSep "\n" (map (p: toString p) cachixPkgs);
    }
  - Or use a runCommand that builds the closure and writes the list. Test locally `nix-build distro/cachix` and inspect the resulting file.
  - Confirm every referenced attr (hello-asterinas, runc.man, podman.man, systemd.package.dev/man) actually exists for your chosen nixpkgs.

4) Workflow: runs-on and container options
- runs-on: you use runs-on: ubuntu-4-cores-150GB-ssd
  - That is not a standard GitHub-hosted runner label (ubuntu-latest, etc.). If this is a self-hosted runner label, that's OK — otherwise it will fail.
- container.options: you pass options: -v /dev:/dev --privileged
  - If this workflow runs on GitHub-hosted runners with a docker container job, --privileged is rarely permitted. Only self-hosted runners will be able to start privileged containers and mount /dev. Make sure the runner environment supports these options.
- Suggestion: document in the workflow or README whether this workflow expects a self-hosted runner. If it must be GitHub-hosted, remove privileged options or restructure.

5) Workflow: ensure cachix client is available in container image
- You use `cachix push` in Makefile; the container image should have the cachix binary installed (or the workflow runner must install it). Verify asterinas/asterinas:0.16.2-20251209 contains cachix and nix installed.
- Suggestion: add a quick step in the workflow or Dockerfile to verify cachix/nix are present (or install them in a small step) to help catch failures early.

6) Minor: typo in workflow step name and messages
- "devlopment" -> "development". Not critical, but nicer to fix.

7) Cachix push log parsing
- You grep the last 10 lines for "^(All done|Nothing to push)"
  - That may be brittle across cachix versions/locales. Consider:
    - check for exit status of `cachix push` directly (it returns non-zero on failure)
    - or grep for "All done" OR success pattern that you observed. The current approach allows non-zero exit (you pipe to tee and || true), then check tail; that approach is okay if the output reliably contains those strings, but be ready for false negatives.

8) Makefile: USE_RELEASE_CACHE variable handling OK, but document usage
- You added USE_RELEASE_CACHE ?= 0 and logic to push either release or dev cache. Consider documenting how to call `make push_cachix USE_RELEASE_CACHE=1` in README.

9) Replacing symlinks with cp -L / cp -r in installer derivation
- Good for creating a self-contained artifact. Note it increases the output size (embedding kernel and overlays). That's probably intended for the ISO/installer.

Testing recommendations
- Locally run the nix-builds:
  - nix-build distro/cachix - test output file (cachix.list) contents
  - nix-build distro/aster_nixos_installer - test that install_aster_nixos template replacements are correct and the installer outputs contain correct files
- Test `make cachix` then `make push_cachix` locally with a small set of packages or a dry-run to validate `cachix push` invocation and the input format.
- Run the GH Actions workflow in a test branch on a runner that mirrors the intended environment (self-hosted vs GitHub-hosted) to validate container options.

Suggested concrete small patches
- core.nix: use safe quoting and splitting
  Replace the two lines you added with something like:
    let
      substitutersList = if config.aster_nixos.substituters == "" then [] else lib.strings.split " " config.aster_nixos.substituters;
      trustedKeysList = if config.aster_nixos."trusted-public-keys" == "" then [] else lib.strings.split " " config.aster_nixos."trusted-public-keys";
    in
    nix.settings = {
      ...
      substituters = substitutersList;
      "trusted-public-keys" = trustedKeysList;
    };
  (Or rename the option to trustedPublicKeys and avoid quotes.)

- distro/cachix/default.nix: produce a cachix.list via writeTextFile
  Example:
    in pkgs.writeTextFile {
      name = "cachix-list";
      text = builtins.concatStringsSep "\n" (map (p: toString p) cachixPkgs);
    }

- Workflow: fix runner label and typo, and add a check step that cachix is present (or document required self-hosted runner):
  - If this is self-hosted, add a comment: # requires self-hosted runner with Docker and privileged mode
  - Else change runs-on to ubuntu-latest and remove container.options --privileged (or add steps to install necessary tools).

Summary of recommended actions (priority order)
1. Fix the hyphen attribute access (config.aster_nixos."trusted-public-keys" or rename).
2. Convert space-separated substituters/trusted-keys strings into list form when setting nix.settings.substituters and nix.settings.trusted-public-keys.
3. Verify pkgs.writeClosure usage or replace with writeTextFile/runCommand to produce a proper cachix.list file; confirm package attribute names.
4. Ensure the workflow runner (label) and container options are compatible with the intended runner and that the container image contains cachix.
5. Minor cleanup: fix typos and consider more robust cachix push failure detection.

If you want, I can:
- produce exact patch snippets for each file (core.nix, distro/cachix/default.nix, workflow) using the code patterns above, or
- run a semantic search across the repo to find all occurrences of the new options so we can rename safely, or
- suggest a test matrix and sample local commands to validate the change set. Which would you prefer next?

PR #2686, opened 2025-12-05
Add new ioctl infrastructure

### LINKED ISSUE #2649: Introducing a new ioctl infrastructure

lrh2000:
Mostly follows #2649, so closes #2649.

Some noticeable differences:
  - Introduce the `dispatch_ioctl` macro for flexibility.
```rust
            GetTermios | SetTermios | SetTermiosDrain | SetTermiosFlush | GetWinSize
            | SetWinSize | GetPtyNumber => {
                return self.slave.ioctl(raw_ioctl);
            }

            cmd @ SetPtyLock => {
                let should_lock = cmd.read()? != 0;

                let flags = self.master_flags();
                if should_lock {
                    flags.set_pty_locked();
                } else {
                    flags.clear_pty_locked();
                }
            }
            // ...

            _ => (self.slave.clone() as Arc<dyn Terminal>).job_ioctl(raw_ioctl, true)?,
	}
```
  - Introduce `SafePtr::copy_from` for type safety.
```rust
impl<T: Pod, M: VmIo, R: TRights> SafePtr<T, M, TRightSet<R>> {
    #[require(R > Write)]
    #[require(R1 > Read)]
    pub fn copy_from<M1: HasVmReaderWriter, R1: TRights>(
        &self,
        ptr: &SafePtr<T, M1, TRightSet<R1>>,
    ) -> Result<()>;
}
```
  - Work around a Rust compiler bug.
```rust
// FIXME: With `impl VmIo for &CurrentUserSpace<'_>`, the Rust compiler seems to think that
// `CurrentUserSpace` is a publicly exposed type, despite the fact that it is contained in a
// private module and is never actually exposed. Consequently, it incorrectly suppresses many dead
// code lints (for *lots of* types that are recursively reached via `CurrentUserSpace`'s APIs). As
// a workaround, we mark the type as `pub(crate)`. We can restore it to `pub` once the compiler bug
// is resolved.
pub(crate) struct CurrentUserSpace<'a>(Ref<'a, Option<Arc<Vmar>>>);
```

### DIFF
--- a/kernel/src/util/ioctl/mod.rs
+++ b/kernel/src/util/ioctl/mod.rs
@@ -92,6 +92,9 @@
 //!    [`Ioctl::with_reader`]), which allows for the writing (or reading) of variable-length data.
 //!    `EVIOCGNAME` is such an example.
 //!
+//!  - For a group of ioctls that serve similar purposes, the [`IoctlEnum`] type allows to
+//!    represent them all with a single type (just like an enum). `EVIOCGBIT` is such an example.
+//!
 
 use core::marker::PhantomData;
 
@@ -471,3 +474,250 @@ macro_rules! dispatch_ioctl {
     };
 }
 pub(crate) use dispatch_ioctl;
+
+/// An ioctl enum that represents one of a possible set of almost identical [`Ioctl`]s.
+///
+/// # Motivation
+///
+/// Sometimes, we want to define a group of ioctls that serve similar purposes.
+/// Their magic numbers, data directions, and data types are exactly the same.
+/// The only difference is the NR byte (see [`IoctlCmd::nr`]).
+///
+/// Let's take `EVIOCGBIT` as an example.
+/// It is an ioctl of input devices (`/dev/input/eventX`),
+/// whose definition in C is shown below:
+///
+/// ```c
+/// #define EVIOCGBIT(ev,len)   _IOC(_IOC_READ, 'E', 0x20 + (ev), len)  /* get event bits */
+/// ```
+///
+/// The valid range of `ev` is `0..0x1F`.
+/// Each possible value of `ev` results in an ioctl command with a different NR.
+/// This is like an ioctl number having an enum embedded in its NR byte!
+///
+/// # Example
+///
+/// To handle ioctl enums,
+/// we introduce `IoctlEnum<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D>`.
+/// Using this type, we can represent the `EVIOCGBIT` ioctl enum readily in Rust:
+///
+/// ```
+/// /// The `EVIOCGBIT` ioctl enum.
+/// type GetEventBits = IoctlEnum<b'E', 0x20, 0x1F, OutData<[u8]>>;
+/// ```
+///
+/// The first (`{b'E'}`) and last (`OutData<[u8]>`) generic parameters
+/// have the same semantics of the ones in `Ioctl`.
+/// The second (`{0x20}`) and third (`{0x1F}`) ones together
+/// dictate the bit patterns for the NR byte.
+/// Assuming the NR byte has a value of `nr`,
+/// `nr` is valid if and only if `(nr & !0x1F) == 0x20`.
+///
+/// Given an ioctl enum `IoctlEnum<_, BASE_NR: u8, ENUM_MASK: u8, _>`,
+/// we say an ioctl _belongs_ to the ioctl enum
+/// if the NR byte of the ioctl, `nr`, satisfies `(nr & !ENUM_MASK) == BASE_NR`.
+/// A `RawIoctl` instance can be fed to `IoctlEnum::try_from_raw`
+/// to check if it belongs to the ioctl enum.
+///
+/// ```
+/// let raw_ioctl = RawIoctl::new(0x80004520, 0);
+/// let ioctl_enum = GetEventBits::try_from_raw(raw_ioctl);
+/// assert!(ioctl_enum.is_some());
+/// assert_eq!(ioctl_enum.unwrap().discriminant(), 0);
+/// ```
+///
+/// Regardless of the NR byte of `raw_ioctl` (say, `nr`),
+/// the `try_from_raw` constructor will mask out the enum bits (`nr & !ENUM_MASK`)
+/// to create a _base_ ioctl whose NR byte equals to `BASE_NR & !ENUM_MASK`.
+/// For simplicity, we require that `(BASE_NR & !ENUM_MASK) == BASE_NR`;
+/// otherwise, a compile-time constant assertion will fail.
+/// The [`IoctlEnum::base_ioctl`] method returns this base ioctl represented as an `Ioctl` instance,
+/// which is useful to handle the ioctl.
+///
+/// The enum variant of the ioctl can be obtained by calling [`IoctlEnum::discriminant`],
+/// whose value is equal to `nr & ENUM_MASK`.
+pub struct IoctlEnum<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D> {
+    base_ioctl: Ioctl<MAGIC, BASE_NR, /* IS_MODERN = */ true, D>,
+    discriminant: u8,
+}
+
+impl<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D: DataSpec>
+    IoctlEnum<MAGIC, BASE_NR, ENUM_MASK, D>
+{
+    /// Tries to interpret a [`RawIoctl`] as an ioctl command that belongs to the ioctl enum.
+    ///
+    /// This method succeeds only if the ioctl command belongs to the ioctl enum. Otherwise, this
+    /// method returns `None`.
+    pub fn try_from_raw(raw_ioctl: RawIoctl) -> Option<Self> {
+        const {
+            assert!(
+                (BASE_NR & ENUM_MASK) == 0,
+                "the base NR must not have the mask bits set"
+            )
+        };
+
+        let (discriminant, raw_base_ioctl) = {
+            let mut cmd = IoctlCmd::new(raw_ioctl.cmd);
+
+            let discriminant = cmd.nr() & ENUM_MASK;
+            cmd.set_nr(cmd.nr() & !ENUM_MASK);
+
+            (discriminant, RawIoctl::new(cmd.as_u32(), raw_ioctl.arg()))
+        };
+
+        let base_ioctl = Ioctl::try_from_raw(raw_base_ioctl)?;
+
+        Some(Self {
+            base_ioctl,
+            discriminant,
+        })
+    }
+}
+
+impl<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D>
+    IoctlEnum<MAGIC, BASE_NR, ENUM_MASK, D>
+{
+    /// Returns a reference to the base ioctl.
+    ///
+    /// See the type-level documentations for the definition of the base ioctl.
+    pub const fn base_ioctl(&self) -> &Ioctl<MAGIC, BASE_NR, /* IS_MODERN = */ true, D> {
+        &self.base_ioctl
+    }
+
+    /// Returns the discriminant of the ioctl enum variant.
+    ///
+    /// See the type-level documentations for the definition of the enum discriminant.
+    pub const fn discriminant(&self) -> u8 {
+        self.discriminant
+    }
+}
+
+#[cfg(ktest)]
+mod test {
+    use ostd::prelude::ktest;
+
+    use super::*;
+
+    type SetControlTty = ioc!(TIOCSCTTY, 0x540E, InData<i32, PassByVal>);
+
+    /// The `EVIOCGBIT` ioctl enum.
+    type GetEventBits = IoctlEnum<b'E', 0x20, 0x1F, OutData<[u8]>>;
+
+    #[derive(Debug, PartialEq, Eq)]
+    enum Action {
+        SetControlTty { should_steal_tty: bool },
+        GetEventBits { event_type: u8, buffer_length: u16 },
+    }
+
+    impl Action {
+        fn parse(cmd: u32, arg: usize) -> Option<Self> {
+            let raw_ioctl = RawIoctl::new(cmd, arg);
+
+            dispatch_ioctl!(match raw_ioctl {
+                cmd @ SetControlTty => {
+                    let should_steal_tty = cmd.get() == 1;
+                    Some(Self::SetControlTty { should_steal_tty })
+                }
+                cmd @ GetEventBits => {
+                    let event_type = cmd.discriminant();
+                    // Ideally, we can use
+                    // `cmd.base_ioctl().with_writer(|writer| Ok(writer.avail()))`.
+                    // But it panics because there is no userspace when performing unit tests.
+                    let buffer_length = cmd.base_ioctl().cmd.size();
+                    Some(Self::GetEventBits {
+                        event_type,
+                        buffer_length,
+                    })
+                }
+                _ => None,
+            })
+        }
+    }
+
+    #[ktest]
+    fn invalid() {
+        let action = Action::parse(0xFFFF, 1234);
+        assert!(action.is_none());
+
+        // Similar to `TIOCSCTTY`, but not quite.
+        let action = Action::parse(0x540F, 4567);
+        assert!(action.is_none());
+
+        // `TIOCSCTTY` with an unexpected data direction.
+        let action = Action::parse(0x8000540E, 1234);
+        assert!(action.is_none());
+
+        // `TIOCSCTTY` with an unexpected data size.
+        let action = Action::parse(0x0001540E, 4567);
+        assert!(action.is_none());
+
+        // `EVIOCGBIT` with an unexpected data direction.
+        let action = Action::parse(0x40004520, 1234);
+        assert!(action.is_none());
+
+        // `EVIOCGBIT` with an underflowed event type.
+        let action = Action::parse(0x4000451F, 4567);
+        assert!(action.is_none());
+
+        // `EVIOCGBIT` with an overflowed event type.
+        let action = Action::parse(0x40004540, 4567);
+        assert!(action.is_none());
+    }
+
+    #[ktest]
+    fn tiocsctty() {
+        let action = Action::parse(0x540E, 0);
+        assert_eq!(
+            action,
+            Some(Action::SetControlTty {
+                should_steal_tty: false
+            })
+        );
+
+        let action = Action::parse(0x540E, 1);
+        assert_eq!(
+            action,
+            Some(Action::SetControlTty {
+                should_steal_tty: true
+            })
+        );
+
+        let action = Action::parse(0x540E, 1 << 63);
+        assert_eq!(
+            action,
+            Some(Action::SetControlTty {
+                should_steal_tty: false
+            })
+        );
+    }
+
+    #[ktest]
+    fn eviocgbit() {
+        let action = Action::parse(0x80004520, 0);
+        assert_eq!(
+            action,
+            Some(Action::GetEventBits {
+                event_type: 0,
+                buffer_length: 0
+            })
+        );
+
+        let action = Action::parse(0x80104528, 1234);
+        assert_eq!(
+            action,
+            Some(Action::GetEventBits {
+                event_type: 8,
+                buffer_length: 16
+            })
+        );
+
+        let action = Action::parse(0xBFFF453F, 4567);
+        assert_eq!(
+            action,
+            Some(Action::GetEventBits {
+                event_type: 0x1F,
+                buffer_length: 0x3FFF
+            })
+        );
+    }
+}
--- a/kernel/src/util/ioctl/sealed.rs
+++ b/kernel/src/util/ioctl/sealed.rs
@@ -38,9 +38,18 @@ impl IoctlCmd {
         Self(cmd)
     }
 
+    pub(super) const fn as_u32(self) -> u32 {
+        self.0
+    }
+
     pub(super) const fn nr(self) -> u8 {
         // Bits 0..=7
         self.0 as u8
+    }
+
+    pub(super) const fn set_nr(&mut self, nr: u8) {
+        // Bits 0..=7
+        self.0 = (self.0 & !0xFF) | (nr as u32);
     }
 
     pub(super) const fn magic(self) -> u8 {


kernel/src/util/ioctl/sealed.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// Layout (from LSB to MSB):
+/// - bits 0..=7   : command number (nr)
+/// - bits 8..=15  : type / magic
+/// - bits 16..=29 : size (in bytes)
+/// - bits 30..=31 : direction
+///
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/ioctl.h#L69-L73>
+#[derive(Clone, Copy, Debug)]
+pub(super) struct IoctlCmd(u32);
```
tatetian, line 15: `IoctlCmd` will be useful in cases (e.g., for ioctl families) where the user of the `ioctl` module has to implement its own ioctl types. So it should be exposed to outside the `ioctl` module. Since there are already two similarly-named structs, `RawIoctl` and `Ioctl`, and both include the `arg` part. To distinguish this type from `RawIoctl` and `Ioctl`, I suggest renaming it to `StdIoctlNumber`.

kernel/src/util/ioctl/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+
+use crate::{current_userspace, prelude::*};
+
+mod sealed;
+
+/// An ioctl command and its argument in raw form.
+#[derive(Clone, Copy, Debug)]
+pub struct RawIoctl {
+    cmd: u32,
+    arg: usize,
```
tatetian, line 112: This falls into the rare cases that I would recommend breaking encapsulation and making fields public. Without exposing its fields, `RawIoctl` would be completely inaccessible and useless except getting fed to `Ioctl::try_from_raw`.

kernel/src/util/ioctl/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+//!             cmd @ GetPtyLock => {
+//!                 cmd.write(if self.is_locked { &1 } else { &0 })?;
+//!             }
+//!             _ => return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown"),
+//!         });
+//!         Ok(())
+//!     }
+//! }
+//!
+//! fn ioctl_dispatch() {
```
tatetian: I suggest renaming to a more specific and accurate name, `do_tty_ioctl` or `fake_tty_ioctl`, for improved readability.

kernel/src/util/ioctl/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+//! Additionally, we support the following advanced usage:
+//!
+//!  - For [`InOutData`], it is possible to obtain a [`SafePtr`] instance to the underlying data
+//!    using [`Ioctl::with_data_ptr`]. This allows some structure fields to be treated as input and
+//!    others as output.
+//!
+//!  - For `OutData<[u8]>` (or `InData<[u8]>`), the argument size is encoded in the ioctl command.
+//!    A [`VmWriter`] (or [`VmReader`]) can be obtained from [`Ioctl::with_writer`] (or
+//!    [`Ioctl::with_reader`]), which allows for the writing (or reading) of variable-length data.
+//!
```
tatetian, line 97: I suggest adding the following paragraph:

```markdown
- As a last resort, you may define your own custom ioctl types. For example, you may write a newtype
  wrapper around `Ioctl<const MAGIC: u8, const NR: u8, const IS_MODERN: bool, D>` to 
  tweak the behavior of the original `Ioctl::try_from_raw` constructor or add additional methods 
  that are specific to a class of ioctl commands.
```

kernel/src/util/ioctl/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+///
+/// For legacy encoding, a 16-bit raw number is specified as the ioctl command.
+///
+/// For example,
+/// ```
+/// type NoTty = ioc!(TIOCNOTTY, 0x5422, NoData);
+/// ```
+/// It is equivalent to:
+/// ```
+/// type NoTty = Ioctl<{0x54}, {0x0E}, NoData>;
```
tatetian:
```rust
type NoTty = Ioctl<{0x54}, {0x0E}, {/* IS_MODERN = */false}, NoData>;
```

kernel/src/util/ioctl/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+/// For modern encoding, a magic (type number) and a command ID (sequence number) are specified.
+/// These, along with the data direction and size, compose the 32-bit ioctl command.
+///
+/// For example,
+/// ```
+/// type SetPtyLock = ioc!(TIOCSPTLCK, b'T', 0x31, InData<i32>);
+/// ```
+/// It is equivalent to:
+/// ```
+/// type SetPtyLock = Ioctl<{b'T'}, {0x31}, InData<i32>>;
```
tatetian:
```rust
type SetPtyLock = Ioctl<{b'T'}, {0x31}, {/* IS_MODERN = */true}, InData<i32>>;
```

kernel/src/util/ioctl/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+/// ```
+macro_rules! ioc {
+    // Legacy encoding.
+    ($linux_name:ident, $raw:literal, $data:ty) => {
+        $crate::util::ioctl::Ioctl::<
+            {
+                // TYPE: bits 8..=15
+                ($raw >> 8) as u8
```
tatetian: Is it possible to write arbitrary const expressions in the location of a const generic parameter?

```rust
                IoctlCmd::new($raw).magic()
```

If so, we can hide the knowledge about the bit offsets and marks into `sealed::IoctlCmd`. If not, then I have no opposition to the current form.

kernel/src/util/ioctl/sealed.rs
```diff
@@ -0,0 +1,69 @@
...
+#[repr(u8)]
+pub enum IoctlDir {
+    None = 0,
+    Write = 1,
+    Read = 2,
+    ReadWrite = 3,
+}
+
+impl IoctlCmd {
+    pub(super) fn new(cmd: u32) -> Self {
```
tatetian: Mark all the methods const.

kernel/src/device/tty/ioctl_defs.rs
```diff
@@ -0,0 +1,28 @@
...
+    termio::{CTermios, CWinSize},
+    CFontOp,
+};
+use crate::util::ioctl::{ioc, InData, OutData, PassByVal};
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/ioctls.h>
+
+pub type GetTermios = ioc!(TCGETS, 0x5401, OutData<CTermios>);
+pub type SetTermios = ioc!(TCSETS, 0x5402, InData<CTermios>);
+pub type SetTermiosDrain = ioc!(TCSETSW, 0x5403, InData<CTermios>);
```
tatetian: It would be really nice if these ioctl definitions can align vertically:

```rust
pub type GetTermios          = ioc!(TCGETS,   0x5401, OutData<CTermios>);
pub type SetTermios          = ioc!(TCSETS,   0x5402, InData<CTermios>);
pub type SetTermiosDrain     = ioc!(TCSETSW,  0x5403, InData<CTermios>);
```

This will greatly improve the readability of these definitions. But rust-fmt will be helping in the wrong way by erasing all the manually-inserted whitespace. I can think of two solutions using `#[rustfmt::skip]`:

```rust
#![rustfmt::skip::macros(ioc)]

pub type GetTermios/*******/= ioc!(TCGETS,   0x5401, OutData<CTermios>);
pub type SetTermios/*******/= ioc!(TCSETS,   0x5402, InData<CTermios>);
pub type SetTermiosDrain/**/= ioc!(TCSETSW,  0x5403, InData<CTermios>);
```

and

```rust
#[rustfmt::skip] pub type GetTermios      = ioc!(TCGETS,   0x5401, OutData<CTermios>);
#[rustfmt::skip] pub type SetTermios      = ioc!(TCSETS,   0x5402, InData<CTermios>);
#[rustfmt::skip] pub type SetTermiosDrain = ioc!(TCSETSW,  0x5403, InData<CTermios>);
```

Both look better than the unaligned form. I am slightly leaning towards the former one.

If you agree, choose either one of them, update the examples given by the `ioctl` module, and format all occurrences of `ioc`.

tatetian (commented):
The new ioctl definition and handling code look beautiful! You have executed the plan perfectly. The new `dispatch_ioctl` macro is a great addition to the original design. I am immensely satisfied.

I raised a few minor points. See below.

kernel/src/fs/utils/mod.rs
```diff
@@ -40,7 +39,6 @@ mod fs;
 mod id_bitmap;
 mod inode;
 mod inode_mode;
-mod ioctl;
```
lrh2000: a

kernel/src/util/ioctl/sealed.rs
lrh2000: I'm confused because I don't understand how you plan to address the ioctl family issue. Creating a wrapper for the `Ioctl` type might not work because `try_from_raw` will still check everything. Therefore, we'll have to add a `from_raw_unchecked` method, but that seems inelegant.

I'll propose something different, as shown in the newly added "Add ioctl family support" commit. It isn't perfect because the Rust compiler rejects `NR: RangeInclusive<u8>`, so I have to write `NR_RANGE_INCLUSIVE: (u8, u8)`. Even the latter requires another unstable feature, `adt_const_params`. However, I believe this approach will (1) work and (2) be more elegant than the `from_raw_unchecked` method. Am I correct?

kernel/src/util/ioctl/mod.rs
lrh2000: I've added some getter methods. I think doing so is enough?

kernel/src/util/ioctl/mod.rs
lrh2000: See my previous comment.

kernel/src/device/tty/ioctl_defs.rs
lrh2000: > #![rustfmt::skip::macros(ioc)]

It seems that if we instruct rustfmt not to format the `ioc` macro, it will not modify any part of the line containing the macro.

So, this will work (I'm surprised!):
```rust
pub type GetTermios         = ioc!(TCGETS,   0x5401, OutData<CTermios>);
```

Actually this behavior is quite necessary because we don't want rustfmt to format the line like below (even if the `ioc!(...)` part is long):
```patch
-pub type GetTermios/*******/= ioc!(TCGETS,   0x5401, OutData<CTermios>);
+pub type GetTermios/*******/=
+    ioc!(TCGETS,   0x5401, OutData<CTermios>);
```

---

But I don't know where to insert `#![rustfmt::skip::macros(ioc)]`. The compiler seems to complain:
```

error[E0658]: custom inner attributes are unstable
 --> src/lib.rs:2:8
  |
2 |     #![rustfmt::skip::macros(ioc)]
  |        ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: see issue #54726 <https://github.com/rust-lang/rust/issues/54726> for more information
  = help: add `#![feature(custom_inner_attributes)]` to the crate attributes to enable
  = note: this compiler was built on 2025-12-05; consider upgrading it if it is out of date
```

We can enable the nightly feature, but it seems that modifying `/rustfmt.toml` is simpler?
```patch
-skip_macro_invocations = ["chmod", "mkmod"]
+skip_macro_invocations = ["chmod", "mkmod", "ioc"]
```

tatetian: > So, this will work (I'm surprised!):
> 
> ```rust
> pub type GetTermios         = ioc!(TCGETS,   0x5401, OutData<CTermios>);
> ```

Maybe the rustfmt's behaviors vary from version to version. Tried [this live example](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=24debcb63a60d658ccd49c57d50dee0a
). On both the latest stable and nightly toolchains, rustfmt erases the whitespaces.

Let's take the current approach and if the toolchain changes on next upgrade, then we will take care of the problem.

lrh2000: This is odd enough.. But [`type` is different from `let`](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=51494250f9b0ab0e40108c576c3e5f1f).. I have no idea why it is the case.

kernel/src/util/ioctl/sealed.rs
tatetian: > I'm confused because I don't understand how you plan to address the ioctl family issue.

> I'll propose something different, as shown in the newly added "Add ioctl family support" commit. It isn't perfect because the Rust compiler rejects `NR: RangeInclusive<u8>`, so I have to write `NR_RANGE_INCLUSIVE: (u8, u8)`.

I saw the last commit. It is a viable solution. But I am afraid that piggybacking the "ioctl family" concept onto the `Ioctl` type makes it mental model much more complex. After all, most people would think an ioctl has a fixed value (at least its NR byte should be fixed) and I suppose that the "ioctl family" is rarely used.

So I propose to dedicate a new type to the "ioctl family" concept. If a developer does not implement ioctl family, then he only needs to read `Ioctl` and do not need to care about `IoctlEnum` (see below). And having a new type is also good for the purpose of educating developers about a new concept.

```rust
/// An ioctl enum represents one of a possible set of almost identical ioctl commands.
///
/// # Motivation
///
/// Sometimes, we want to define a group of ioctls that serve similar purposes.
/// Their magic numbers, data directions, and data types are exactly the same.
/// The only difference is the NR byte (see [`IoctlCmd::nr`]).
///
/// Let's take `EVIOCGBIT` as an example.
/// It is an ioctl of input devices (`/dev/input/eventX`),
/// whose definition in C is shown below:
///
/// ```c
/// #define EVIOCGBIT(ev,len)   _IOC(_IOC_READ, 'E', 0x20 + (ev), len)  /* get event bits */
/// ```
///
/// The valid range of `ev` is `0..0x1F`.
/// Each possible value of `ev` results in an ioctl command with a different NR.
/// This is like an ioctl number having an enum embedded in its NR byte!
///
/// # Example
///
/// To handle ioctl enums,
/// we introduce `IoctlEnum<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D>`.
/// Using this type, we can represent the `EVIOCGBIT` ioctl enum readily in Rust:
///
/// ```rust
/// /// The `EVIOCGBIT` ioctl enum.
/// pub type GetEventBits = IoctlEnum<{b'E'}, {0x20}, {0x1F}, OutData<[u8]>>;
/// ```
///
/// The first (`{b'E'}`) and last (`OutData<[u8]>`) generic parameters
/// have the same semantics of the ones in `Ioctl`.
/// The second (`{0x20}`) and third (`{0x1F}`) ones together 
/// dictate the bit patterns for the NR byte.
/// Assuming the NR byte has a value of `nr`,
/// `nr` is valid if and only if `(nr & !0x1F) == 0x20`.
/// 
/// Given an ioctl enum `IoctlEnum<_, BASE_NR: u8, ENUM_MASK: u8, _>`,
/// we say an ioctl _belongs_ to the ioctl enum 
/// if the NR byte of the ioctl, `nr`, satisfies `(nr & !ENUM_MASK) == BASE_NR`.
/// A `RawIoctl` instance can be fed to `IoctlEnum::try_from_raw`
/// to check if it belongs to the ioctl enum.
///
/// ```
/// let raw_ioctl: RawIoctl = todo!();
/// let ioctl_enum = GetEventBits::try_from_raw(raw_ioctl)?;
/// ```
///
/// Regardless of the `NR` byte of `raw_ioctl` (say, `nr`),
/// the `try_from_raw` constructor will mask out the enum bits (`nr & !ENUM_MASK`)
/// to create a _base_ ioctl whose `NR` byte equals to `BASE_NR & !ENUM_MASK`.
/// The [`IoctlEnum::base_ioctl`] method returns this base ioctl represented as an `Ioctl` instance,
/// which is useful to handle the ioctl.
/// 
/// The enum part of the ioctl can be obtained by calling [`IoctlEnum::enum`],
/// whose value is equal to `nr & ENUM_MASK`.
pub struct IoctlEnum<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D> {
    base_ioctl: Ioctl<MAGIC, {BASE_NR & !ENUM_MASK}, {/* IS_MODERN =*/true}, D>,
    enum_val: u8,
}

impl<const MAGIC: u8, const BASE_NR: u8, const ENUM_MASK: u8, D> IoctlEnum<MAGIC, BASE_NR, ENUM_MASK, D> {
    /// Creates an ioctl enum if the given raw ioctl belongs to it.
    pub fn try_from_raw(raw_ioctl: RawIoctl) -> Option<Self> {
        todo!("try create the base ioctl to see if the raw ioctl belongs to the ioctl enum")
    }

    pub fn base_ioctl(&sel) -> &Ioctl<Self::MAGIC, {Self::BASE_NR & !Self::ENUM_MASK}, {/* IS_MODERN =*/true}, D> { ... }

    pub fn enum_val(&sel) -> u8 { ... }
}
```

lrh2000: > I saw the last commit. It is a viable solution. But I am afraid that piggybacking the "ioctl family" concept onto the `Ioctl` type makes it mental model much more complex. After all, most people would think an ioctl has a fixed value (at least its NR byte should be fixed) and I suppose that the "ioctl family" is rarely used.

The `ioc!` macro should be used instead, so the `Ioctl` type won't appear explicitly (which means that we don't need to supply the inclusive range manually). I don't really think it's a big problem.

> `IoctlEnum`

I think it's okay, and it's an improvement. However, the improvement does not seem significant (given that we have the `ioc!` macro for the common case). But since you wrote the suggested code, which is almost complete, I will follow it.

> base_ioctl: Ioctl<MAGIC, {BASE_NR & !ENUM_MASK}, {/* IS_MODERN =*/true}, D>,

This isn't possible:
```
error: generic parameters may not be used in const operations
 --> src/lib.rs:5:12
  |
5 |     g2: G<{X+1}>,
  |            ^ cannot perform const operation using `X`
  |
  = help: const parameters may only be used as standalone arguments here, i.e. `X`
  = help: add `#![feature(generic_const_exprs)]` to allow generic const expressions
```

`generic_const_exprs` is incomplete and has caused problems in our codebase in the past. So we cannot enable it.

But using `BASE_NR` (instead of `BASE_NR & !ENUM_MASK`) should be fine.

tatetian (approved):

PR #2665, opened 2025-12-03
Support `nix-shell` by logging in as root

vvvvsv:
With this patch, we can correctly set `PATH` and `NIX_PATH` in Asterinas NixOS to support `nix-shell`. https://github.com/asterinas/asterinas/issues/2672

`nix-shell` is a nix command that starts a temporary shell with all the specified packages, without installing them.

```
sh-5.2# nix-shell -p hello
GC Warning: pthread_getattr_np or pthread_attr_getstack failed for main thread
warning: error: unable to download 'https://cache.nixos.org/nix-cache-info': Could not resolve hostname (6) Could not resolve host: cache.nixos.org; retrying in -2147483648 ms
this path will be fetched (0.05 MiB download, 0.22 MiB unpacked):
  /nix/store/na5dnb0l6rjkzrbld3xn4188dd8nknhh-hello-2.12.1
copying path '/nix/store/na5dnb0l6rjkzrbld3xn4188dd8nknhh-hello-2.12.1' from 'https://cache.nixos.org'...

[nix-shell:/]# hello
Hello, world!

[nix-shell:/]# exit
exit
sh-5.2# hello
The program 'hello' is not in your PATH. It is provided by several packages.
You can make it available in an ephemeral shell by typing one of the following:
  nix-shell -p fltk
  nix-shell -p fltk14
  nix-shell -p fyne
  nix-shell -p haskellPackages.hello
  nix-shell -p hello
  nix-shell -p mbedtls
  nix-shell -p tests.haskell.cabalSdist.helloFromCabalSdist
  nix-shell -p tests.rust-hooks.cargoBuildHook
  nix-shell -p tests.rust-hooks.cargoCheckHook
  nix-shell -p tests.rust-hooks.cargoInstallHook
  nix-shell -p tests.rust-hooks.cargoNextestHook
  nix-shell -p tests.rust-hooks.cargoSetupHook
sh-5.2# 
```

Please review https://github.com/asterinas/asterinas/pull/2668 before reviewing this PR.

PR #2687, opened 2025-12-05
Default to systemd boot

cchanging:
Based on #2652 .

This PR enables systemd by default in the stage2 of the Asterinas NixOS distribution, adjusts the corresponding CI tests accordingly, and allows users to enable xfce with systemd.

When booting with systemd, it automatically invokes the login program for user login. When users need to exit the system, they need to execute the `poweroff` command in the shell, which will sync the current file system states and shut down the machine.

With this PR, users can uncomment the xfce-related sections in `configuration.nix` to enable the graphical interface. During boot, systemd will automatically start the services required for the graphical interface.


distro/modules/xfce/start_xfce.sh
```diff
@@ -2,6 +2,8 @@
 
 # SPDX-License-Identifier: MPL-2.0
 
+source /etc/profile
```
tatetian, line 5: Is this still needed after #2665?

.github/workflows/test_nixos_minimal.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         with:
           image: asterinas/asterinas:0.16.1-20251130
           options: --privileged -v /dev:/dev -v ${{ github.workspace }}:/root/asterinas
-          run: make run_nixos NIXOS_DISK_SIZE_IN_MB=4096 NIXOS_STAGE_2_INIT='/bin/sh -l -c "hello-asterinas"' || true
+          run: make run_nixos NIXOS_DISK_SIZE_IN_MB=4096 NIXOS_RUN_TEST=true || true
```
tatetian: `NIXOS_RUN_TEST` and `NIXOS_TEST_COMMAND` can be combined into one. If `NIXOS_TEST_COMMAND` is non-empty, then it implies running a test.

distro/modules/core.nix
```diff
@@ -36,6 +36,22 @@ in {
...
+    ${lib.optionalString config.asterinas.run-test ''
+      cp /etc/profile /etc/profile.bak
+      echo '
+      # --- Run tests in the hvc0 terminal ---
+      if [ "$(tty)" = "/dev/hvc0" ]; then
+        # Resume /etc/profile
+        cp /etc/profile.bak /etc/profile
+        rm /etc/profile.bak
+        # Execute specified tests
+        ${config.asterinas.test-command}
```
tatetian: This looks very hacky. Can't think of a more "correct" approach?

After booting, the system presents the user with an interactive shell. Can't we emulate typing by writing to the QEMU console?

distro/modules/core.nix
```diff
@@ -36,6 +36,22 @@ in {
...
+        # Wait and poweroff 
+        sleep 2
+        poweroff
```
tatetian, line 44: Why sleep?

tatetian (commented):
Good job. The PR looks mostly fine with me.

distro/modules/core.nix
tatetian: To provide an extra hint, the ISO installer has an `environment.loginShellInit` option:

https://github.com/asterinas/asterinas/blob/386e174f0c686dd3b6b4821202426c5a0442c10d/distro/iso-image.nix#L20-L28

Does NixOS has a similar option that allows us to control the login shell easily?

distro/modules/xfce/start_xfce.sh
cchanging: Yes, #2665 only enables root login when systemd is not present, and with systemd it already login as root by default. The issue here is that the script to launch xfce is not invoked by a logined shell, but rather by a custom systemd service we created. Therefore, when it executes, it lacks certain environment variable information that is only available after a proper login.

distro/modules/core.nix
cchanging: Previously, the `sleep` command was added because during my testing, I noticed that after running it twice, it suddenly failed to start on the third attempt. I suspected it might be due to some background service not finishing its initialization. However, after conducting several more tests, I haven’t been able to reproduce the issue... I also checked and confirmed there are no corresponding background services involved. 

I suspect it might still be an ext2-related issue... but in reality, the `sleep` should indeed be unnecessary.

tatetian:
I am satisfied with the current code. After #2652 is merged, rebase this PR so that it can be merged.

tatetian (approved):

Issue #2689, opened 2025-12-07
Support running TCG-accelerated QEMU


LordaeronESZ:
### Target
Compile a qemu-system-x86_64 statically and prepare the necessary pc-bios, rootfs, and Linux kernel, then attempt to run the emulator. If it fails to run, try introducing new features for Asterinas or fixing existing bugs.

### Current issues
- [ ] The prctl parameter `PR_SET_TIMERSLACK` is not implemented.

LordaeronESZ:
After solving the issues aformentioned, we can use TCG-accelerated qemu-system-x86_64 to boot a Linux kernel with a simple rootfs built by buildroot. Here is my screenshot when I test it.

<img width="2524" height="964" alt="Image" src="https://github.com/user-attachments/assets/c7dfe186-64dc-4d51-bd56-2c2708a1cd3e" />

PR #2690, opened 2025-12-07
Add `timer_slack` field in `PosixThread` and implement related methods.

LordaeronESZ:
As #2689 explained, I found the issue that the prctl parameter `PR_SET_TIMERSLACK` is not implemented in Asterinas. 

``` bash
/ext2/qemu-test # ls
bzImage             pc-bios             rootfs.ext4
output.log          qemu-system-x86_64  run_static_tcg.sh
/ext2/qemu-test # ./run_static_tcg.sh
[    71.426] ERROR: Uncaught panic:
        not yet implemented
        at /root/asterinas/kernel/src/syscall/prctl.rs:163
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(0), real_time: RealTimeAttr { prio: 99, time_slice: 59486260 }, fair: FairAttr { weight: 1024, pending_weight: 0, vruntime: 2055674966 } } })
```

In this pr, I add support for it and some other related fields and methods. However, I do not use `timer_slack_ns` for any usage, it is just a placeholder for now.


kernel/src/process/posix_thread/mod.rs
```diff
@@ -312,6 +317,32 @@ impl PosixThread {
...
+    }
+
+    /// Returns the default timer slack value in nanoseconds.
+    pub fn default_timer_slack_ns(&self) -> u64 {
+        self.default_timer_slack_ns.load(Ordering::Relaxed)
+    }
+
+    /// Sets the default timer slack value in nanoseconds.
+    pub fn set_default_timer_slack_ns(&self, slack_ns: u64) {
+        self.default_timer_slack_ns.store(slack_ns, Ordering::Relaxed);
```
tatetian: Need to implement the correct behavior on how the default timer slack is determined. The man page of fork says:

> The default timer slack value is set to the parent's current timer slack value.

So when a child process / thread is forked or cloned, its current and default timer slack value are set to the value of its parent's current timer slack. After the child is up and running, its default timer slack cannot be changed. This means that `PosixThread` should not have a `set_default_timer_slack_ns` method.

To do so, you need to add a new method, `default_timer_slack`, to `PosixThreadBuilder` (no need for a `timer_slack` method) and call this new method [when `clone`ing a thread](https://github.com/asterinas/asterinas/blob/85ca6b8934417635ea6ec3cd1b6150b02c329cf7/kernel/src/process/clone.rs#L325).

kernel/src/syscall/prctl.rs
```diff
@@ -159,8 +175,8 @@ impl PrctlCmd {
             PR_SET_DUMPABLE => Ok(PrctlCmd::PR_SET_DUMPABLE(Dumpable::try_from(arg2)?)),
             PR_SET_NAME => Ok(PrctlCmd::PR_SET_NAME(arg2 as _)),
             PR_GET_NAME => Ok(PrctlCmd::PR_GET_NAME(arg2 as _)),
-            PR_GET_TIMERSLACK => todo!(),
-            PR_SET_TIMERSLACK => todo!(),
+            PR_GET_TIMERSLACK => Ok(PrctlCmd::PR_GET_TIMERSLACK),
+            PR_SET_TIMERSLACK => Ok(PrctlCmd::PR_SET_TIMERSLACK(arg2)),
```
tatetian, line 177: Update [the `prctl` section](https://github.com/asterinas/asterinas/tree/85ca6b8934417635ea6ec3cd1b6150b02c329cf7/book/src/kernel/linux-compatibility/syscall-flag-coverage/namespaces-cgroups-and-security#prctl) of the Book and its SCML file.

kernel/src/process/posix_thread/mod.rs
LordaeronESZ: Based on your suggestions, I have modified my code.

In addition, I used LLM to write a simple test program for me:

``` c
#include <stdio.h>
#include <sys/prctl.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/wait.h>

int main() {
    // 测试1：获取当前 timer slack
    long current_slack = prctl(PR_GET_TIMERSLACK);
    printf("Current timer slack: %ld ns\n", current_slack);
    
    // 测试2：设置新的 timer slack
    long new_slack = 100000; // 100 us
    if (prctl(PR_SET_TIMERSLACK, new_slack) == 0) {
        printf("Set timer slack to %ld ns\n", new_slack);
        
        // 验证设置是否成功
        long verify_slack = prctl(PR_GET_TIMERSLACK);
        printf("Verified timer slack: %ld ns\n", verify_slack);
    } else {
        perror("PR_SET_TIMERSLACK failed");
    }
    
    // 测试3：测试继承（fork）
    pid_t pid = fork();
    if (pid == 0) {
        // 子进程
        long child_slack = prctl(PR_GET_TIMERSLACK);
        printf("Child timer slack (inherited): %ld ns\n", child_slack);
        
        // 修改子进程的 slack
        prctl(PR_SET_TIMERSLACK, 200000);
        printf("Child set timer slack to 200000 ns\n");
    } else if (pid > 0) {
        // 父进程
        wait(NULL);
        long parent_slack = prctl(PR_GET_TIMERSLACK);
        printf("Parent timer slack (should still be %ld): %ld ns\n", new_slack, parent_slack);
    }
    
    // 测试4：重置为默认值
    if (prctl(PR_SET_TIMERSLACK, 0) == 0) {
        long default_slack = prctl(PR_GET_TIMERSLACK);
        printf("Reset to default slack: %ld ns\n", default_slack);
    }
    
    return 0;
}
```

And it has been verified that the program behaves in a consistent manner with my local Linux environment.

slack

PR #2691, opened 2025-12-07
Add Docker development environment on ARM (aarch64)

junyang-zh:
Depends on #2995 

As discussed in https://github.com/asterinas/asterinas/discussions/2574, it would be great to improve the development experience on ARM platforms such as Apple Silicon Macs. This PR releases `linux/arm64` Docker images to reduce the Rosetta 2 translator overhead.

Nevertheless, Docker on Darwin is far from mature. This PR doesn't address how to use HVF to accelerate AArch64 QEMU guests.

I have built a preview image using this PR: [`junyangzhang/asterinas:0.16.1-20251207-arm-preview`](https://hub.docker.com/repository/docker/junyangzhang/asterinas/tags/0.17.0-20260114-arm-preview/sha256:bc8eae5af4b267aef74460e2a5cfe38ebf6f48374d6cce736303a989fec15dea).

Currently, x86 & RISC-V kernel development works, except for Gvisor system call tests. Gvisor system call tests work only in x86-64 images. It is tested on macOS with OrbStack.


tatetian:
This is quite an upgrade to our Docker image, involving  a lot of changes. Have you tested it?

Could you update README or the Book to provide the instructions to run the Asterinas docker image on macOS?

Also, Github has [free runners on macOS](https://docs.github.com/en/actions/reference/runners/github-hosted-runners). In subsequent PRs, we should add a Github workflow to run some tests on MacOS.

Edit: Instructions should be provided after having the new Docker image. I am worried about the uncertainty brought by the major change to the Docker image. How about merging this PR after we release AsterNixOS?

junyang-zh:
> involving a lot of changes.

Not quite a lot. Just fixing some cross-compilation things that used not to work, and enabling another Docker platform.

> Have you tested it?

Yes. I have pushed a preview image and have been working on it for a while. See the updated PR description.

> In subsequent PRs, we should add a GitHub workflow to run some tests on macOS.

Another option is to test on `linux/arm64` hosts, since what we release is actually for `linux/arm64` platforms, not `darwin/arm64`.

junyang-zh:
> I am worried about the uncertainty brought by the major change to the Docker image.

Again, the change is not significant. The x86 workflow has not changed. The change is not related to AsternixOS. However about the merging pace, you can proceed at your own judgement.

tatetian:
Thanks for the contribution! This PR bundles two separate tasks: (1) upgrading Ubuntu from 22.04 to 24.04, and (2) adding ARM64 Docker images for the development environment. Both are valuable changes. However, I'd prefer splitting them into two separate PRs, even though this means generating two consecutive versions of Docker images. The extra effort is worth the more focused review and smoother, steadier progress.

For the second PR (the ARM Docker image), we'll also need to discuss some documentation work. Specifically, when people see "ARM" in our documentation, they may get the wrong impression that Asterinas supports ARM as a *deployment* target. To avoid this confusion, I'd like to introduce the concepts of "development platform" vs. "deployment platform" in our documentation. In particular, I'm considering adding something like the following to the `README.md`:

```md
## Getting Started

### Supported CPU Architectures

Asterinas distinguishes between _development platforms_ and _deployment platforms_:
- A **development platform** is where you build and test Asterinas
  (i.e., the host machine running the Docker-based development environment).
- A **deployment platform** is a CPU architecture that Asterinas can run on as an OS kernel.

| CPU Architecture | As Development Platform | As Deployment Platform |
| ---------------- | ----------------------- | ---------------------- |
| x86-64           | Yes                     | Yes (Tier-1)           |
| ARM64            | Yes                     | No                     |
| RISC-V           | No                      | Yes (Tier-2)           |
| LoongArch        | No                      | Yes (Tier-3)           |

The deployment platform tiers are defined as follows:
- **Tier-1**: Fully supported and tested.
  CI pipelines run the full test suite on every PR.
  This is the recommended platform for production use.
- **Tier-2**: Actively developed with basic functionality working.
  CI pipelines run build checks and basic tests on every PR,
  but the full test suite is not yet covered.
- **Tier-3**: Early-stage or experimental support.
  The kernel can boot and perform basic operations,
  but the architecture is not regularly tested in CI.
```

tatetian:
> I'm considering adding something like the following to the `README.md`:

The suggested doc improvement has been done in #2991.

tatetian:
I edited the PR description with the following line:

> Depends on #2995 

to prevent myself from merging this PR prematurely.

osdk/tools/docker/Dockerfile
```diff
@@ -3,6 +3,7 @@
 FROM ubuntu:24.04 AS build-base
 
 ARG DEBIAN_FRONTEND=noninteractive
+ARG TARGETARCH
```
tatetian: Is the argument name `TARGETARCH` inherited from somewhere or coined by you? `TARGET_ARCH` looks better to me.

tatetian:
@junyang-zh The PR looks fine with me. After having linux/arm64 Docker images, could you also add new CI or modify existing ones to run tests on the new linux/arm64 Docker images?

chen1plus:
I noticed that `make format` doesn't format `.h` correctly in this image. Could you reproduce this issue?

junyang-zh:
> I noticed that `make format` doesn't format `.h` correctly in this image. Could you reproduce this issue?

https://github.com/asterinas/asterinas/pull/3008

This is a known problem of upgrading to Ubuntu 24.04, not ARM. So the behavior will be aligned when we merge the ref bump PR.

.github/workflows/benchmark_x86.yml
```diff
@@ -109,7 +109,7 @@ jobs:
     container: 
       image: asterinas/asterinas:0.17.1-20260317
       options: --device=/dev/kvm --privileged
-
+-v /dev:/dev
```
tatetian: **Broken YAML from garbled substitution**: The blank line separating `container:` from `steps:` was replaced with `-v /dev:/dev` at zero indentation. This is not a valid YAML key or value and will cause a parse error, preventing the workflow from loading. It looks like a find-and-replace gone wrong — this pattern recurs across many workflow files.

.github/workflows/benchmark_x86_tdx.yml
```diff
@@ -110,7 +110,7 @@ jobs:
     container: 
       image: asterinas/asterinas:0.17.1-20260317
       options: --device=/dev/kvm --privileged
-
+-v /dev:/dev
```
tatetian: **Same broken YAML as `benchmark_x86.yml`**: Stray `-v /dev:/dev` at zero indentation will cause a YAML parse error.

.github/workflows/test_asterinas_vsock.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         run: |
             docker run \
               --privileged --network=host --device=/dev/kvm \
-              -v ./:/root/asterinas asterinas/asterinas:0.17.1-20260317 \
+              -v ./:/root/asterinas asteri-v /dev:/dev7.1-20260317 \
```
tatetian: **Docker image name destroyed**: The substitution mangled the image reference `asterinas/asterinas:0.17.1-20260317` into the nonsensical string `asteri-v /dev:/dev7.1-20260317`. This `docker run` command will fail because the image name is invalid. The `-v /dev:/dev` flag was inserted mid-string, destroying both the image name and the volume flag.

.github/workflows/test_loongarch.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.1-20260317
       options: --device=/dev/kvm --privileged
-    strategy:
+    strategy:-v /dev:/dev
```
tatetian: **YAML `strategy` key corrupted**: `-v /dev:/dev` was appended directly to the `strategy:` key, making it `strategy:-v /dev:/dev`. This turns the key into a scalar value instead of a mapping, so the `matrix:` block beneath is no longer parsed as part of `strategy`. The workflow will fail to load.

.github/workflows/test_riscv.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.1-20260317
       options: --device=/dev/kvm --privileged
-      volumes:
+      volumes:-v /dev:/dev
```
tatetian: **YAML `volumes` key corrupted**: Same pattern — `-v /dev:/dev` appended directly to `volumes:`, destroying the mapping. The volume mount list beneath becomes unparseable.

.github/workflows/test_riscv.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         # the container can remove them to save some precious disk space (~16GB),
         # which is quite limited on free-tier Github runners.
         - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
-        - /usr/local/lib/android:/usr/local/lib/android
+        - /usr/-v /dev:/dev:/usr/local/lib/android
```
tatetian: **Volume mount path corrupted**: The path `/usr/local/lib/android:/usr/local/lib/android` was mangled into `/usr/-v /dev:/dev:/usr/local/lib/android`. This is an invalid volume mount that will prevent the container from starting.

.github/workflows/test_x86_tdx.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.1-20260317
       options: --device=/dev/kvm --privileged
-    strategy:
+    strategy:-v /dev:/dev
```
tatetian: **YAML `strategy` key corrupted**: Same pattern as `test_loongarch.yml`. The `strategy` key is destroyed and the workflow will fail to parse.

book/src/kernel/intel-tdx.md
```diff
@@ -67,7 +67,7 @@ The following result is an example:
 
     ```bash
     docker run -it --privileged --network=host --device=/dev/kvm -v $(pwd)/asterinas:/root/asterinas asterinas/asterinas:0.17.1-20260317
-    ```
+    ```-v /dev:/dev
```
tatetian: **Markdown code fence corrupted**: The closing ` ``` ` was replaced with `` ```-v /dev:/dev ``, so the code block is never properly closed. The rest of the document will render as code in the Asterinas Book.

book/src/osdk/guide/intel-tdx.md
```diff
@@ -33,7 +33,7 @@ Run a TDX Docker container:
 
 ```bash
 docker run -it --privileged --network=host --device=/dev/kvm asterinas/osdk:0.17.1-20260317
-```
+```-v /dev:/dev
```
tatetian: **Markdown code fence corrupted**: Same issue — the closing fence becomes `` ```-v /dev:/dev ``, breaking the code block termination and corrupting the rendered documentation.

osdk/tools/docker/Dockerfile
```diff
@@ -169,7 +214,15 @@ RUN cargo install --locked \
...
-ENV LD_LIBRARY_PATH="/usr/local/qemu/lib/x86_64-linux-gnu:${LD_LIBRARY_PATH}"
+# Buildx `TARGETARCH` (amd64 and arm64) is different from QEMU's target (x86_64
+# and aarch64). Create aliases so the lib path always exists.
+RUN if [ -d /usr/local/qemu/lib/aarch64-linux-gnu ] && [ ! -e /usr/local/qemu/lib/arm64-linux-gnu ]; then \
+        ln -s /usr/local/qemu/lib/aarch64-linux-gnu /usr/local/qemu/lib/arm64-linux-gnu; \
+    fi \
+    && if [ -d /usr/local/qemu/lib/x86_64-linux-gnu ] && [ ! -e /usr/local/qemu/lib/amd64-linux-gnu ]; then \
+        ln -s /usr/local/qemu/lib/x86_64-linux-gnu /usr/local/qemu/lib/amd64-linux-gnu; \
+    fi
+ENV LD_LIBRARY_PATH="/usr/local/qemu/lib/${TARGETARCH}-linux-gnu:${LD_LIBRARY_PATH:-''}"
```
tatetian: **Literal single quotes baked into `LD_LIBRARY_PATH`**: The `:-''` default-value syntax will inject literal `''` characters into the path when `LD_LIBRARY_PATH` is unset, resulting in `/usr/local/qemu/lib/amd64-linux-gnu:''`. The dynamic linker may interpret these literal quotes as part of a directory name. Use `${LD_LIBRARY_PATH:-}` or simply `${LD_LIBRARY_PATH}` to avoid injecting quote characters.

osdk/tools/docker/Dockerfile
tatetian, line 1: **`TARGETARCH` ARG may need re-declaration in each stage**: Per Docker BuildKit docs, `TARGETARCH` is an automatic platform ARG, but to use it in a build stage you must declare `ARG TARGETARCH` (without a value) after the `FROM` line. The `ARG TARGETARCH` on line 6 is scoped to `build-base` only. Stages `build-cross-tools`, `ovmf`, `grub`, and `osdk-final` reference `${TARGETARCH}` but may not have it in scope unless BuildKit implicitly provides it. In practice, BuildKit's automatic ARGs are available globally without re-declaration, so this may work — but the Docker documentation is ambiguous on this point, and adding `ARG TARGETARCH` in each stage that uses it would be more robust and explicit.

README.md
```diff
@@ -96,36 +96,28 @@ in data centers, autonomous vehicles, and embodied AI.
...
+- A **development platform** is where you build and test Asterinas
+  (i.e., the host machine running the Docker-based development environment).
+- A **deployment platform** is a CPU architecture that Asterinas can run on as an OS kernel.
+
+| CPU Architecture | As Development Platform | As Deployment Platform |
+| ---------------- | ----------------------- | ---------------------- |
+| x86-64           | Yes                     | Yes (Tier-1)           |
+| ARM64            | Yes                     | No                     |
+| RISC-V           | No                      | Yes (Tier-2)           |
+| LoongArch        | No                      | Yes (Tier-3)           |
```
tatetian: **Outdated tables**: This PR modifes the tables for supported CPU architectures based on an outdated version. Please rebase this PR and update the table based on the latest version.

README.md
```diff
@@ -96,36 +96,28 @@ in data centers, autonomous vehicles, and embodied AI.
...
-  (per PR for RISC-V and nightly for Intel TDX),
+Asterinas distinguishes between _development platforms_ and _deployment platforms_:
+- A **development platform** is where you build and test Asterinas
+  (i.e., the host machine running the Docker-based development environment).
+- A **deployment platform** is a CPU architecture that Asterinas can run on as an OS kernel.
+
+| CPU Architecture | As Development Platform | As Deployment Platform |
+| ---------------- | ----------------------- | ---------------------- |
+| x86-64           | Yes                     | Yes (Tier-1)           |
+| ARM64            | Yes                     | No                     |
```
tatetian: **Naming inconsistency across the PR**: The table uses "ARM64" (title-case), the PR title uses "ARM (aarch64)", the Dockerfiles use `arm64` (lowercase from `TARGETARCH`), and the official ARM architecture name is "AArch64". The other table entries use canonical names: "x86-64", "RISC-V", "LoongArch". Consider settling on one name consistently — "AArch64" would be most parallel with the formal naming of the other entries.

In general, "AArch64" is better than `arm64`, which looks almost identical to `amd64`. So whenever possible, use "AArch64".

tatetian:
By the way, this PR can also bump the Docker image version to trigger the generation of new Docker images as soon as this PR is merged.

README.md
junyang-zh: Disagree. From a beginner's perspective, I think `ARM64` is better known than `AArch64`. This is also the case for `x86-64` v.s. `amd64`. We pick the former.

tatetian: I agree that ARM64 is more obvious than AARCH64. I am fine with ARM64. Just be consistent.

book/src/kernel/README.md
```diff
@@ -28,39 +28,33 @@ we are steadfastly progressing towards our goal.
...
-  CI runs build checks and basic tests on a regular basis
-  (per PR for RISC-V and nightly for Intel TDX),
+Asterinas distinguishes between _development platforms_ and _deployment platforms_:
+- A **development platform** is where you build and test Asterinas
+  (i.e., the host machine running the Docker-based development environment).
+- A **deployment platform** is a CPU architecture that Asterinas can run on as an OS kernel.
+
+|   CPU Architecture  | As Development Platform | As Deployment Platform |
+| ------------------- | ----------------------- | ---------------------- |
+| x86-64              | Yes                     | Yes (Tier-1)           |
```
tatetian: This PR changes this section from two-table form to one-table one. The one-table form was the original one that I proposed, but I switched to the two-table form later because it would make which development platforms (only one or two) are supported more obvious. So please revise the supported CPU architecture info base on the current two-table form.

README.md
```diff
@@ -96,36 +96,29 @@ in data centers, autonomous vehicles, and embodied AI.
...
-- **Tier 2:** Actively developed with basic functionality working.
-  CI runs build checks and basic tests on a regular basis
-  (per PR for RISC-V and nightly for Intel TDX),
+Asterinas distinguishes between _development platforms_ and _deployment platforms_:
+- A **development platform** is where you build and test Asterinas
+  (i.e., the host machine running the Docker-based development environment).
+- A **deployment platform** is a CPU architecture that Asterinas can run on as an OS kernel.
+
+|   CPU Architecture  | As Development Platform | As Deployment Platform |
+| ------------------- | ----------------------- | ---------------------- |
```
tatetian: Same comment as above

tools/docker/README.md
```diff
@@ -11,6 +11,7 @@ cd <asterinas dir>
 # Build Docker image
 docker buildx build \
     -f tools/docker/Dockerfile \
+    --platform linux/amd64,linux/arm64 \
```
tatetian, line 14: > I have built a preview image using this PR: [`junyangzhang/asterinas:0.16.1-20251207-arm-preview`](https://hub.docker.com/repository/docker/junyangzhang/asterinas/tags/0.17.0-20260114-arm-preview/sha256:bc8eae5af4b267aef74460e2a5cfe38ebf6f48374d6cce736303a989fec15dea).
> 
> Currently, x86 & RISC-V kernel development works, except for Gvisor system call tests. Gvisor system call tests work only in x86-64 images. It is tested on macOS with OrbStack.

You acknowledged that "Gvisor system call tests work only in x86-64 images". This is fine. But I expect all three Asterinas development Docker image made ARM compatible, including the `asterinas:asterinas` Docker images. And the `--platform linux/amd64,linux/arm64 ` argument shows that you also intend to make the Docker image compatible for both x86-64 and ARM64. 

However, `tools/docker/Dockerfile` still unconditionally installed x86-specific packages such as `libc6-dev-i386`, `grub-efi-amd64-bin`, and `grub-efi-amd64-dbg`. So I guess this ARM64-compatible `junyangzhang/asterinas:0.16.1-20251207-arm-preview` is only the OSDK Docker image?

I am afraid that the `asterinas:nix` Docker image is also not made ARM compatible. This PR does not touch `tools/docker/nix/Dockerfile` at all. It should ensure that `make install_host_pkgs` on ARM64 and add a `arm64_pkgs` Makefile target to pre-build ARM64 Nix packages.

README.md
tatetian: This line should mention ARM64.

tatetian (commented):
Thanks for the revision. Unfortunately, I found a major gap that I missed in the last round of review.

tools/docker/README.md
junyang-zh: > However, tools/docker/Dockerfile still unconditionally installed x86-specific packages such as libc6-dev-i386, grub-efi-amd64-bin, and grub-efi-amd64-dbg.

These are for building and running x86 Apps on x86 Asterinas on an ARM host.

> So I guess this ARM64-compatible junyangzhang/asterinas:0.16.1-20251207-arm-preview is only the OSDK Docker image?

No. It is the kernel development image.

> This PR does not touch tools/docker/nix/Dockerfile at all.

There's no need to touch it.

> It should ensure that make install_host_pkgs on ARM64 and add a arm64_pkgs Makefile target to pre-build ARM64 Nix packages.

Nope. `install_host_pkgs` installs the tools on the host for benchmarks (like an HTTP client that sends requests to the server in Asterinas and measures time). They SHOULD follow the host arch.

tatetian: Thanks for the explanation!

PR #2692, opened 2025-12-08
Add some stubs for `madvise`.

LordaeronESZ:
As #2689 mentioned, I found that madvise parameter `MADV_HUGEPAGE`, `MADV_MERGEABLE` and `MADV_DONTFORK` are not implemented. So I add some related stubs for them.


tatetian:
Need to update [the `madvise` section](https://github.com/asterinas/asterinas/tree/main/book/src/kernel/linux-compatibility/syscall-flag-coverage/memory-management#madvise) of the book. Moving the three flags from the "unsupported" list to the "silently-ignored" list. And [the SCML file](https://github.com/asterinas/asterinas/blob/main/book/src/kernel/linux-compatibility/syscall-flag-coverage/memory-management/madvise.scml) of `madvise` should also be updated.

By the way, please revise the code to ensure all CI tests pass.

LordaeronESZ:
I have modified the code according to your suggestions. In addition, I found that the madvise section of the book has inconsistencies with the code, which I have also modified.

tatetian (approved):

PR #2693, opened 2025-12-08
Update README for FAST'26 news

tatetian:


tatetian:
Bypass the merge rules for this trivial PR

PR #2694, opened 2025-12-08
Support re`open`ing anonymous pipes from `/proc`

vvvvsv:
This is required for https://github.com/asterinas/asterinas/issues/2672.

This PR supports re`open`ing anonymous pipes by making an anonymous pipe as a wrapper around a named pipe.


kernel/src/fs/pipe/named_pipe.rs
```diff
@@ -157,10 +173,21 @@ impl NamedPipe {
         access_mode: AccessMode,
         status_flags: StatusFlags,
     ) -> Result<Box<dyn FileIo>> {
+        self.open_handle(access_mode, status_flags, true)
+            .map(|handle| handle as Box<dyn FileIo>)
```
StevenJiang1110: This can be slightly shorter

```rust
Ok(self.open_handle(access_mode, status_flags, true)?)
```

kernel/src/fs/pipe/named_pipe.rs
```diff
@@ -157,10 +173,21 @@ impl NamedPipe {
...
+    }
+
+    /// Opens the pipe and returns a `NamedPipeHandle`.
+    pub(super) fn open_handle(
+        &self,
+        access_mode: AccessMode,
+        status_flags: StatusFlags,
+        is_named_pipe: bool,
+    ) -> Result<Box<NamedPipeHandle>> {
```
StevenJiang1110: Now that the namedpipe-related structs is used for both named pipe and anonymous pipe, perhaps we should rename `NamedPipeHandle` as `PipeHandle`. And similarly, `NamedPipe` should be named as `Pipe`.

vvvvsv: > rename `NamedPipeHandle` as `PipeHandle`. And similarly, `NamedPipe` should be named as `Pipe`.

I agree. 

They and their inner structs are the entire contents of `named_pipe.rs`. My suggestion is to move everything from `namedpipe.rs` into `common.rs` after the rename, since that fits the semantics better. Doing so would also allow us to make `common::{PipeReader, PipeWriter}` private to `common` to reduce the exposed API.

StevenJiang1110: > My suggestion is to move everything from namedpipe.rs into common.rs after the rename, since that fits the semantics better.

Makes sense to me. You can proceed with the changes.

kernel/src/fs/pipe/anon_pipe.rs
```diff
@@ -22,93 +29,85 @@ use crate::{
...
-    Ok((
-        PipeReaderFile::new(reader, StatusFlags::empty(), pseudo_inode.clone())?,
-        PipeWriterFile::new(writer, StatusFlags::empty(), pseudo_inode)?,
-    ))
-}
-
-/// A file handle for reading from a pipe.
-pub struct PipeReaderFile {
-    reader: PipeReader,
+/// An anonymous pipe file.
```
StevenJiang1110, line 45: It's also a little better better to rename the file from `anony_pipe.rs` to `anon_pipe.rs`, this better matches the struct name.

vvvvsv:
@boterinas rerun

PR #2695, opened 2025-12-08
Revise safety comments of entry points

lrh2000:
https://github.com/asterinas/asterinas/blob/4f71f4bbe6456849929d7035ec5f23bb04200279/ostd/src/boot/smp.rs#L129-L134

I don't think this is enough. It needs to guarantee the following:
https://github.com/asterinas/asterinas/blob/4f71f4bbe6456849929d7035ec5f23bb04200279/ostd/src/arch/riscv/mod.rs#L74-L77
More specifically:
https://github.com/asterinas/asterinas/blob/4f71f4bbe6456849929d7035ec5f23bb04200279/ostd/src/arch/riscv/irq/chip/mod.rs#L57-L59

However, the current safety requirements cannot. `riscv_ap_early_entry` can call many (safe) methods without violating the safety condition of `ap_early_entry`. However, it can violate the safety condition of `arch::riscv::init_on_ap`/`arch::riscv::irq::chip::init`.

The safety comments for the `ap_early_entry` method need to be more specific than just stating that the method can only be called once. I propose:
```rust
/// - This function must be called only once on each AP at a proper timing in the AP's boot
///   assembly code, or via a thin Rust wrapper that does not access uninitialized AP states.
```

Then,
 - If it is called from the AP's boot assembly code, we know that no Rust methods have been called yet.
 - If it is called via a thin Rust wrapper, we must also state that the Rust wrapper will not access uninitialized AP states. This means it cannot call methods inside `arch::riscv::irq::chip`, which would violate the safety conditions of `arch::riscv::irq::chip::init`.

In addition, I use "at a proper timing" to mean that this method must be called after the AP's boot assembly code prepares all the necessary resources. It is not safe to call it in the middle of the boot assembly code, for example. It is similar to:
https://github.com/asterinas/asterinas/blob/4f71f4bbe6456849929d7035ec5f23bb04200279/ostd/src/arch/riscv/timer/mod.rs#L28

### DIFF
--- a/ostd/src/arch/loongarch/boot/mod.rs
+++ b/ostd/src/arch/loongarch/boot/mod.rs
@@ -102,7 +102,7 @@ fn check_cpu_config() {
 ///
 /// # Safety
 ///
-/// - This function must be called only once through assembly code.
+/// - This function must be called only once at a proper timing in the BSP's boot assembly code.
 /// - The caller must follow C calling conventions and put the right arguments in registers.
 #[no_mangle]
 unsafe extern "C" fn loongarch_boot(
--- a/ostd/src/arch/riscv/boot/mod.rs
+++ b/ostd/src/arch/riscv/boot/mod.rs
@@ -112,14 +112,15 @@ static mut BOOTSTRAP_HART_ID: u32 = u32::MAX;
 ///
 /// # Safety
 ///
-/// - This function must be called only once through assembly code.
+/// - This function must be called only once at a proper timing in the BSP's boot assembly code.
 /// - The caller must follow C calling conventions and put the right arguments in registers.
 #[no_mangle]
 unsafe extern "C" fn riscv_boot(hart_id: usize, device_tree_paddr: usize) -> ! {
     early_println!("Enter riscv_boot");
 
-    // SAFETY: We only write it once this time. Other processors will only read
-    // it. And other processors are not booted yet so there's no races.
+    // We will only write it once. Other processors will only read it.
+    // SAFETY: We don't create Rust references, so there are no aliasing problems. Other processors
+    // have not been booted yet, so there are no data races.
     unsafe { BOOTSTRAP_HART_ID = hart_id as u32 };
 
     let device_tree_ptr = paddr_to_vaddr(device_tree_paddr) as *const u8;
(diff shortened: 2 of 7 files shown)


zjp-CN:
Totally agree.

And for this requirement:

> This function must be called only once on each AP at a proper timing in the AP's boot assembly code, or via a thin Rust wrapper that does not access uninitialized AP states.

With the draft tag annotations implemented in [tag-asterinas](https://github.com/Artisan-Lab/tag-asterinas/tree/2025-11), I constructed an unsafe call tree on x86-64 boot since https://github.com/Artisan-Lab/tag-std/pull/84 like this

```
boot::call_ostd_main
├── init { CallOnce, Context }
│   ├── mm::frame::allocator::init_early_allocator { CallOnce, PostToFunc }
│   ├── cpu::init_on_bsp { Context, PostToFunc, Unaccessed }
│   │   ├── cpu::local::copy_bsp_for_ap { Context, Unaccessed, Valid }
│   │   │   ├── core::intrinsics::write_bytes
│   │   │   ├── core::slice::from_raw_parts_mut
│   │   │   ├── cpu::local::__cpu_local_start
│   │   │   ├── cpu::local::__cpu_local_end
│   │   │   └── core::intrinsics::copy_nonoverlapping
│   │   ├── cpu::set_this_cpu_id { Valid, CallOnce }
│   │   └── cpu::init_num_cpus { Context, Valid }
│   ├── mm::frame::meta::init { CallOnce, Context }
│   │   └── mm::page_table::boot_pt::BootPageTable::<E, C>::map_base_page { KernelMemorySafe }
│   │       └─── ...
│   ├── mm::frame::allocator::init { CallOnce, PostToFunc, PostToFunc }
│   ├── arch::late_init_on_bsp { CallOnce, Context }
│   └── mm::kspace::activate_kernel_page_table { CallOnce }
│       ├── mm::page_table::PageTable::<C>::first_activate_unchecked { CallOnce }
│       │   └── mm::page_table::node::<impl mm::frame::Frame<mm::page_table::node::PageTablePageMeta<C>>>::first_activate { Valid, CallOnce }
│       │       └── arch::mm::activate_page_table { ValidInstanceAddr }
│       └── mm::page_table::boot_pt::dismiss { NotPriorToFunc, NotPostToFunc, CallOnce }
└── boot::call_ostd_main::__ostd_main
```

`PostToFunc` is one of the missing requirements for entry functions on other platform targets.

PR #2698, opened 2025-12-08
Workaround unrecognized input event types by writing all zeros

Cautreoxit:
This PR serves as a debug patch for PR [#2582](https://github.com/asterinas/asterinas/pull/2582). While that PR successfully passed `evtest`, we discovered that it would crash when running XFCE due to additional calls of  other discriminants of `GetEventBits`. Although it doesn't utilize these uncommon event types, the ioctl would crash if it returned an error. This patch ensures that those calls return Ok instead of errors, preventing the crash.

### DIFF
--- a/kernel/src/device/evdev/file.rs
+++ b/kernel/src/device/evdev/file.rs
@@ -334,6 +334,19 @@ fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWrite
             let bitmap = capability.supported_relative_axes_bitmap();
             write_bytes_and_zeros_to_userspace(writer, bitmap)?;
         }
+        t if t == EventTypes::ABS.as_index()
+            || t == EventTypes::LED.as_index()
+            || t == EventTypes::SW.as_index()
+            || t == EventTypes::MSC.as_index()
+            || t == EventTypes::FF.as_index()
+            || t == EventTypes::SND.as_index() =>
+        {
+            // TODO: We do not support these uncommon event types yet, but
+            // returning an error would cause `libevdev` to crash. So we report
+            // a zeroed bitmap here. See
+            // <https://cgit.freedesktop.org/libevdev/tree/libevdev/libevdev.c?h=libevdev-1.13.6#n459>.
+            writer.fill_zeros(writer.avail())?;
+        }
         _ => {
             return_errno_with_message!(Errno::EINVAL, "the event type is not supported yet");
         }


TinaZhangZW:
To make the intent clearer, can we rename PR to: "Return success and zeroed bitmap for unsupported event types"?

kernel/src/device/evdev/file.rs
```diff
@@ -335,7 +335,10 @@ fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWrite
             write_bytes_and_zeros_to_userspace(writer, bitmap)?;
         }
         _ => {
-            return_errno_with_message!(Errno::EINVAL, "the event type is not supported yet");
+            // TODO: We do not support other uncommon event types yet, but
+            // returning error would cause `libevdev` to crash. So we return
+            // Ok(()) here.
+            log::warn!("event type is not supported yet: {}", event_type);
```
tatetian: We should at least write all zeros. I don't think the user is required to zeroize the output buffer before calling the system call. If the user does not clear the buffer, then he would take whatever garbage in the buffer as the bitmap.

tatetian:
> Return success and zeroed bitmap for unsupported event types"?

I would suggest "Workaround unrecognized input event types by writing all zeros" to emphasize that this commit is a dummy implementation.

kernel/src/device/evdev/file.rs
```diff
@@ -335,7 +335,10 @@ fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWrite
             write_bytes_and_zeros_to_userspace(writer, bitmap)?;
         }
         _ => {
```
lrh2000, line 350: Could you please be more specific and explicitly list the event types that we don't support and that will be needed by Xorg/libevdev? This should be the job of this PR, not a future PR that tries to resolve this TODO.

If we complete the aforementioned task, I don't think we'll need to print warnings here. However, we still need to return errors for types outside the known set.

kernel/src/device/evdev/file.rs
```diff
@@ -335,7 +335,10 @@ fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWrite
...
-            return_errno_with_message!(Errno::EINVAL, "the event type is not supported yet");
+            // TODO: We do not support other uncommon event types yet, but
+            // returning error would cause `libevdev` to crash. So we return
+            // Ok(()) here.
```
lrh2000:
```rust
	    // TODO: We do not support these uncommon event types yet, but
	    // returning an error would cause `libevdev` to crash. So we report
	    // a zeroed bitmap here.
```

kernel/src/device/evdev/file.rs
Cautreoxit: Thanks for your advice! Here is how `libevdev` calls for these discriminants of `GetEventBits`, including six uncommon event types in total. I am not sure if we really need to add this into source code as Reference.

https://github.com/whot/libevdev/blob/master/libevdev/libevdev.c#L400

kernel/src/device/evdev/file.rs
```diff
@@ -334,6 +334,18 @@ fn handle_get_bit(evdev: &Arc<EvdevDevice>, event_type: u8, writer: &mut VmWrite
...
+        // a zeroed bitmap here.
+        t if t == EventTypes::ABS.as_index()
+            || t == EventTypes::LED.as_index()
+            || t == EventTypes::SW.as_index()
+            || t == EventTypes::MSC.as_index()
+            || t == EventTypes::FF.as_index()
+            || t == EventTypes::SND.as_index() =>
+        {
+            writer.fill_zeros(writer.avail())?;
+        }
```
lrh2000: This isn't a big issue, but can't we be more consistent with the existing code?
https://github.com/asterinas/asterinas/blob/2919f43c654cea630160bf24e7033356fb818c26/kernel/src/device/evdev/file.rs#L453-L457

```rust
        t if t == EventTypes::ABS.as_index()
            || t == EventTypes::LED.as_index()
            || t == EventTypes::SW.as_index()
            || t == EventTypes::MSC.as_index()
            || t == EventTypes::FF.as_index()
            || t == EventTypes::SND.as_index() =>
        {
            // TODO: We do not support these uncommon event types yet, but
            // returning an error would cause `libevdev` to crash. So we report
            // a zeroed bitmap here.
            writer.fill_zeros(writer.avail())?;
        }
```

kernel/src/device/evdev/file.rs
lrh2000: Either way is fine, but I'd prefer a different link because the GitHub repository appears to be outdated.
```rust
            // TODO: We do not support these uncommon event types yet, but
            // returning an error would cause `libevdev` to crash. So we report
            // a zeroed bitmap here. See
            // <https://cgit.freedesktop.org/libevdev/tree/libevdev/libevdev.c?h=libevdev-1.13.6#n459>.
```

Cautreoxit:
@boterinas rerun

PR #2700, opened 2025-12-08
Fix the implementation of prlimit64 syscall and the init process value of ResourceLimits

wyt8:
### Short description

Fix incorrect handling of the `pid` argument in the `prlimit64` syscall implementation. This patch updates the implementation to follow the correct behavior as documented in the [Linux manual page](https://man7.org/linux/man-pages/man2/prlimit.2.html).

According to the [Linux code](https://github.com/torvalds/linux/blob/cb015814f8b6eebcbb8e46e111d108892c5e6821/kernel/fork.c#L869C50-L870), this pr also correct the value of ResourceLimits for the init process.

### Review notes

When reviewing and merging this PR, please pay extra attention to the capability-related logic to make sure it remains correct and consistent with the existing design.

### DIFF
--- a/kernel/src/syscall/prlimit64.rs
+++ b/kernel/src/syscall/prlimit64.rs
@@ -5,7 +5,13 @@ use ostd::mm::VmIo;
 use super::SyscallReturn;
 use crate::{
     prelude::*,
-    process::{Pid, ResourceType, rlimit::RawRLimit64},
+    process::{
+        Pid, Process, ResourceType,
+        credentials::capabilities::CapSet,
+        posix_thread::{AsPosixThread, PosixThread},
+        process_table,
+        rlimit::RawRLimit64,
+    },
 };
 
 pub fn sys_getrlimit(resource: u32, rlim_addr: Vaddr, ctx: &Context) -> Result<SyscallReturn> {
@@ -45,19 +51,74 @@ pub fn sys_prlimit64(
         "pid = {}, resource = {:?}, new_rlim_addr = 0x{:x}, old_rlim_addr = 0x{:x}",
         pid, resource, new_rlim_addr, old_rlim_addr
     );
-    let resource_limits = ctx.process.resource_limits();
-    if old_rlim_addr != 0 {
-        let rlimit = resource_limits.get_rlimit(resource);
-        let (cur, max) = rlimit.get_cur_and_max();
-        let rlimit_raw = RawRLimit64 { cur, max };
-        ctx.user_space().write_val(old_rlim_addr, &rlimit_raw)?;
-    }
-    if new_rlim_addr != 0 {
-        let new_raw: RawRLimit64 = ctx.user_space().read_val(new_rlim_addr)?;
-        debug!("new_rlimit = {:?}", new_raw);
-        resource_limits
-            .get_rlimit(resource)
-            .set_cur_and_max(new_raw.cur, new_raw.max)?;
+
+    let get_and_set_rlimit = |process: &Process| -> Result<()> {
+        let resource_limits = process.resource_limits();
+
+        if old_rlim_addr != 0 {
+            let rlimit = resource_limits.get_rlimit(resource);
+            let (cur, max) = rlimit.get_cur_and_max();
+            let rlimit_raw = RawRLimit64 { cur, max };
+            ctx.user_space().write_val(old_rlim_addr, &rlimit_raw)?;
+        }
+        if new_rlim_addr != 0 {
+            let new_raw: RawRLimit64 = ctx.user_space().read_val(new_rlim_addr)?;
+            debug!("new_rlimit = {:?}", new_raw);
+            resource_limits
+                .get_rlimit(resource)
+                .set_cur_and_max(new_raw.cur, new_raw.max)?;
+        }
+
+        Ok(())
+    };
+
+    if pid == 0 || pid == ctx.process.pid() {
+        get_and_set_rlimit(ctx.process.as_ref())?;
+    } else {
+        let target_process = process_table::get_process(pid).ok_or_else(|| {
+            Error::with_message(Errno::ESRCH, "the target process does not exist")
+        })?;
+
+        // Check permissions
+        check_rlimit_perm(target_process.main_thread().as_posix_thread().unwrap(), ctx)?;
+        get_and_set_rlimit(target_process.as_ref())?;
     }
     Ok(SyscallReturn::Return(0))
+}
+
+/// Checks whether the current process has permission to access
+/// the resource limits of the target process.
+// Reference: <https://man7.org/linux/man-pages/man2/prlimit.2.html>
+fn check_rlimit_perm(target: &PosixThread, ctx: &Context) -> Result<()> {
+    let target_process = target.process();
+
+    let current_cred = ctx.posix_thread.credentials();
+    let target_cred = target.credentials();
+
+    if target_process
+        .user_ns()
+        .lock()
+        .check_cap(CapSet::SYS_RESOURCE, ctx.posix_thread)
+        .is_ok()
+    {
+        return Ok(());
+    }
+
+    let current_ruid = current_cred.ruid();
+    let current_rgid = current_cred.rgid();
+
+    if current_ruid == target_cred.ruid()
+        && current_ruid == target_cred.euid()
+        && current_ruid == target_cred.suid()
+        && current_rgid == target_cred.rgid()
+        && current_rgid == target_cred.egid()
+        && current_rgid == target_cred.sgid()
+    {
+        return Ok(());
+    }
+
+    return_errno_with_message!(
+        Errno::EPERM,
+        "accessing the resource limits of the target process is not allowed"
+    )
 }
--- a/kernel/src/process/process/init_proc.rs
+++ b/kernel/src/process/process/init_proc.rs
@@ -17,7 +17,7 @@ use crate::{
         posix_thread::{PosixThreadBuilder, ThreadName, allocate_posix_tid},
         process_table,
         process_vm::new_vmar_and_map,
-        rlimit::ResourceLimits,
+        rlimit::new_resource_limits_for_init,
         signal::sig_disposition::SigDispositions,
     },
     sched::Nice,
@@ -56,7 +56,7 @@ fn create_init_process(
 
     let pid = allocate_posix_tid();
     let process_vm = new_vmar_and_map(PathOrInode::Path(elf_path.clone()));
-    let resource_limits = ResourceLimits::default();
+    let resource_limits = new_resource_limits_for_init();
     let nice = Nice::default();
     let oom_score_adj = 0;
     let sig_dispositions = Arc::new(Mutex::new(SigDispositions::default()));
--- a/kernel/src/process/rlimit.rs
+++ b/kernel/src/process/rlimit.rs
@@ -149,7 +149,8 @@ impl RLimit64 {
 
     /// Sets the rlimit with synchronization.
     ///
-    /// Only called when handling the `setrlimit` or `prlimit` syscall.
+    /// Only called when handling the `setrlimit` or `prlimit` syscall
+    /// or during init process creation.
     pub fn set_cur_and_max(&self, new_cur: u64, new_max: u64) -> Result<()> {
         if new_cur > new_max {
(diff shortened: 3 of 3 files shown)


distro/modules/core.nix
```diff
@@ -34,6 +34,7 @@ in {
     if [ "${builtins.getEnv "NIXOS_DISABLE_SYSTEMD"}" = "true" ]; then
       ${builtins.getEnv "NIXOS_STAGE_2_INIT"}
     fi
+    prlimit --nproc 50000
```
vvvvsv: `prlimit` should only be run when `NIXOS_DISABLE_SYSTEMD` is set to `true`, and it should be executed before `/bin/sh -l`?

```bash
    if [ "${builtins.getEnv "NIXOS_DISABLE_SYSTEMD"}" = "true" ]; then
      prlimit --pid 1 --nproc=50000
      ${builtins.getEnv "NIXOS_STAGE_2_INIT"}
    fi
```

Please provide a detailed explanation in the PR about why we need this `prlimit`, just as we discussed privately. Also include a brief comment before this line.

kernel/src/syscall/prlimit64.rs
```diff
@@ -43,19 +49,80 @@ pub fn sys_prlimit64(
...
+    if target_process
+        .user_ns()
+        .lock()
+        .check_cap(CapSet::SYS_RESOURCE, ctx.posix_thread)
+        .is_ok()
+    {
+        return Ok(());
+    }
+
+    if current_cred.ruid() == target_cred.ruid()
```
tatetian: This `current_cred.ruid()` call is duplicated many times. Assign it to a variable to make it more clear that it is the real user ID that is repeatedly compared. Same argument applies to the real group ID.

kernel/src/syscall/prlimit64.rs
```diff
@@ -43,19 +49,80 @@ pub fn sys_prlimit64(
...
+        && current_cred.ruid() == target_cred.euid()
+        && current_cred.ruid() == target_cred.suid()
+        && current_cred.rgid() == target_cred.rgid()
+        && current_cred.rgid() == target_cred.egid()
+        && current_cred.rgid() == target_cred.sgid()
+    {
```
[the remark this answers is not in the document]
wyt8: > The pid argument specifies the ID of the process on which the call
> is to operate.  If pid is 0, then the call applies to the calling
> process.  To set or **get** the resources of a process other than
> itself, the caller must have the CAP_SYS_RESOURCE capability in
> the user namespace of the process whose resource limits are being
> changed, or the real, effective, **and** saved set user IDs of the
> target process must match the real user ID of the caller **and** the
> real, effective, **and** saved set group IDs of the target process
> must match the real group ID of the caller.

请问你的原文是这个吗

kernel/src/syscall/prlimit64.rs
```diff
@@ -43,19 +49,80 @@ pub fn sys_prlimit64(
...
+                .get_rlimit(resource)
+                .set_cur_and_max(new_raw.cur, new_raw.max)?;
+        }
+    } else {
+        let target_process = process_table::get_process(pid).ok_or_else(|| {
+            Error::with_message(Errno::ESRCH, "the target process does not exist")
+        })?;
+
+        // Check permissions
+        check_rlimit_perm(target_process.main_thread().as_posix_thread().unwrap(), ctx)?;
```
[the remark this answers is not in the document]
wyt8: > The pid argument specifies the ID of the process on which the call
> is to operate.  If pid is 0, then the call applies to the calling
> process.  To set or **get** the resources of a process other than
> itself, the caller must have the CAP_SYS_RESOURCE capability in
> the user namespace of the process whose resource limits are being
> changed, or the real, effective, **and** saved set user IDs of the
> target process must match the real user ID of the caller **and** the
> real, effective, **and** saved set group IDs of the target process
> must match the real group ID of the caller.

请问你的原文是这个吗

kernel/src/syscall/prlimit64.rs
```diff
@@ -43,19 +49,80 @@ pub fn sys_prlimit64(
...
+            ctx.user_space().write_val(old_rlim_addr, &rlimit_raw)?;
+        }
+        if new_rlim_addr != 0 {
+            let new_raw: RawRLimit64 = ctx.user_space().read_val(new_rlim_addr)?;
+            debug!("new_rlimit = {:?}", new_raw);
+            resource_limits
+                .get_rlimit(resource)
+                .set_cur_and_max(new_raw.cur, new_raw.max)?;
+        }
     }
```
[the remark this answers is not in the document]
wyt8: 由于借用的作用域影响，你的建议更改方式有点问题，请你重新想一个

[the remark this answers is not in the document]
vvvvsv: The Copilot reviewer does not respond automatically. You need to push a new commit, then click the 🔄 button next to "Copilot" under "Reviewers" to re-request a review.

You can use this approach to resolve the borrow scope issue here:

```rust
pub fn sys_prlimit64(
    pid: Pid,
    resource: u32,
    new_rlim_addr: Vaddr,
    old_rlim_addr: Vaddr,
    ctx: &Context,
) -> Result<SyscallReturn> {
    let resource = ResourceType::try_from(resource)?;
    debug!(
        "pid = {}, resource = {:?}, new_rlim_addr = 0x{:x}, old_rlim_addr = 0x{:x}",
        pid, resource, new_rlim_addr, old_rlim_addr
    );

    let get_and_set_rlimit = |process: &Process| -> Result<()> {
        let resource_limits = process.resource_limits();

        if old_rlim_addr != 0 {
            let rlimit = resource_limits.get_rlimit(resource);
            let (cur, max) = rlimit.get_cur_and_max();
            let rlimit_raw = RawRLimit64 { cur, max };
            ctx.user_space().write_val(old_rlim_addr, &rlimit_raw)?;
        }
        if new_rlim_addr != 0 {
            let new_raw: RawRLimit64 = ctx.user_space().read_val(new_rlim_addr)?;
            debug!("new_rlimit = {:?}", new_raw);
            resource_limits
                .get_rlimit(resource)
                .set_cur_and_max(new_raw.cur, new_raw.max)?;
        }

        Ok(())
    };

    if pid == 0 || pid == ctx.process.pid() {
        get_and_set_rlimit(&ctx.process)?;
    } else {
        let target_process = process_table::get_process(pid).ok_or_else(|| {
            Error::with_message(Errno::ESRCH, "the target process does not exist")
        })?;
        // check_rlimit_perm(...)?;
        get_and_set_rlimit(&target_process)?;
    }

    Ok(SyscallReturn::Return(0))
}
```

distro/modules/core.nix
```diff
@@ -32,6 +32,8 @@ in {
     rm -rf /etc/resolv.conf
     ln -s ${resolv-conf} /etc/resolv.conf
     if [ "${builtins.getEnv "NIXOS_DISABLE_SYSTEMD"}" = "true" ]; then
+      # We use this command to simulate the behaviour of `systemd` temporarily.
+      prlimit --pid 1 --nproc=50000
```
tatetian: The comment is still insufficient on explaining the reason to set this limit. Saying "simulate the behaviour of `systemd`" is missing the point. The more fundamental reason needs to be given. What is the default value? What problem would be caused if not setting this limit? And why 50000? If the `nproc` limit is to be raised, why raise limits for other resources as well?

It seems that your intention is to relax the resource limits on the subsequent `NIXOS_STAGE_2_INIT` command. If so, a more direct and robust approach would be:

```bash
prlimit --nproc=50000 ${builtins.getEnv "NIXOS_STAGE_2_INIT"}
```

Doing "prlimit --pid 1" seems to imply that the init process (pid = 1) is a shell and it is this shell that will execute `${builtins.getEnv "NIXOS_STAGE_2_INIT"}`. Otherwise, setting the init process's limit wouldn't affect `${builtins.getEnv "NIXOS_STAGE_2_INIT"}`. But this is a fragile assumption: there is no telling if a simple change to a NixOS option or the NixOS implementation would cause `${builtins.getEnv "NIXOS_STAGE_2_INIT"}` to become the child process of another process other than the init process.

distro/modules/core.nix
wyt8: 之前在运行bash时候产生的0x8错误，根本原因是因为启动了job control功能的bash在初始化的时候会根据prlimit64返回的nproc值，用来设置它的某个结构体中的值。如果这个值为0，就会导致0x8错误。

[启动时nproc值为0](https://github.com/asterinas/asterinas/blob/main/kernel/src/process/rlimit.rs#L17)，需要systemd等类似的进程设置rlimit的nproc值，这样后续子进程会继承该值。

这里的50000值是参考在ubuntu运行时返回的值，这个值大概是5万多。

关于bash，具体可以看jobs.c中的`set_maxchild`、`initialize_job_control`等函数。

distro/modules/core.nix
wyt8: 之前在运行bash时候产生的0x8错误，根本原因是因为启动了job control功能的bash在初始化的时候会根据prlimit64返回的nproc值，用来设置它的某个结构体中的值。如果这个值为0，就会导致0x8错误。

[启动时nproc值为0](https://github.com/asterinas/asterinas/blob/main/kernel/src/process/rlimit.rs#L17)，需要systemd等类似的进程设置rlimit的nproc值，这样后续子进程会继承该值。

这里的50000值是参考在ubuntu运行时返回的值，这个值大概是5万多。

关于bash，具体可以看jobs.c中的set_maxchild、initialize_job_control等函数。

distro/modules/core.nix
```diff
@@ -32,7 +32,13 @@ in {
...
     ln -s ${resolv-conf} /etc/resolv.conf
     if [ "${builtins.getEnv "NIXOS_DISABLE_SYSTEMD"}" = "true" ]; then
-      ${builtins.getEnv "NIXOS_STAGE_2_INIT"}
+      # We use this command to simulate the behaviour of `systemd` temporarily.
+      # When system boots, the nproc of rlimit is 0, which can lead to `bash` error.
+      # So we need to set the nproc to a non-zero value, which is usually done by `systemd`.
+      # The value we set now is adopted from the true environment.
+      # I think the specific value is not very important, and the only matter is
+      # non-zero and not too small.
+      prlimit --nproc=50000 ${builtins.getEnv "NIXOS_STAGE_2_INIT"}
```
tatetian: It turned out that we are fixing the wrong place.

Let's first ask this question: what is the default value of `NPROC` for the init process on Linux? It is set to `max_threads / 2`.

```c
void __init fork_init(void)
{
    ...
    set_max_threads(MAX_THREADS);
	init_task.signal->rlim[RLIMIT_NPROC].rlim_cur = max_threads/2;
	init_task.signal->rlim[RLIMIT_NPROC].rlim_max = max_threads/2;
    ...
}
```
(https://github.com/torvalds/linux/blob/cb015814f8b6eebcbb8e46e111d108892c5e6821/kernel/fork.c#L869C50-L870)

The `max_threads` static variable is initialized by `set_max_threads`, which sets the maximum number of threads to a large value, although not not too large so that thread structs do not consume a significant portion of the system memory.

Then, take a guess: what is the default `NPROC` value on Asterinas?

Unfortunately, it is 0.

https://github.com/asterinas/asterinas/blob/8ce2f9373dc9a81bdae3af3027f8b367c7044fdf/kernel/src/process/rlimit.rs#L17

So instead of setting this value here, this PR should fix the default value in Asterinas--ideally, using a strategy inspired by Linux.

Could you also examine the default values of other resource limits and see if they are initialized properly?

tatetian:
> Thanks for the revision. Now the code looks good. But the commit message is outdated.

@wyt8 I meant the _commit_ message, not the PR title.

kernel/src/syscall/prlimit64.rs
```diff
@@ -45,19 +51,74 @@ pub fn sys_prlimit64(
...
+        })?;
+
+        // Check permissions
+        check_rlimit_perm(target_process.main_thread().as_posix_thread().unwrap(), ctx)?;
+        get_and_set_rlimit(target_process.as_ref())?;
     }
     Ok(SyscallReturn::Return(0))
 }
+
+/// Check whether the current process has permission to access
```
lrh2000: Style:
```suggestion
/// Checks whether the current process has permission to access
```

PR #2702, opened 2025-12-09
Bump Docker image version to 0.16.2

junyang-zh:
For #2701

PR #2703, opened 2025-12-09
Bump version to 0.16.2

junyang-zh:
Following #2702


tatetian:
Need to split this PR into two: the first commit is one, the next two are the other. Doing so allows us to ensure the code is correct before bumping the project version.

junyang-zh:
> Need to split this PR into two: the first commit is one, the next two are the other. Doing so allows us to ensure the code is correct before bumping the project version.

Do you mean #2702 ?

tatetian:
> > Need to split this PR into two: the first commit is one, the next two are the other. Doing so allows us to ensure the code is correct before bumping the project version.
> 
> Do you mean #2702 ?

Yes. Sorry, I was wrong.

junyang-zh:
@tatetian CI passed. The exfat failure is known. The publishing dry-run failure is expected. The xorriso failure is still caused by insufficient disk space.

tatetian (approved):

PR #2705, opened 2025-12-09
Support `rt_sigtimedwait` syscall

cchanging:
`rt_sigtimedwait` syscall will be used by systemd when calling `poweroff`. systemd uses this syscall to wait for related processes to complete `sync` operations. Without this syscall, the system might sometimes shut down before `sync` is fully completed, leading to data corruption.

### DIFF
--- a/test/src/syscall/gvisor/Makefile
+++ b/test/src/syscall/gvisor/Makefile
@@ -54,6 +54,7 @@ TESTS ?= \
 	sigaction_test \
 	sigaltstack_test \
 	signalfd_test \
+	sigtimedwait_test \
 	socket_netlink_route_test \
 	socket_unix_dgram_local_test \
 	socket_unix_dgram_non_blocking_test \
--- a/test/src/syscall/gvisor/blocklists/sigtimedwait_test
+++ b/test/src/syscall/gvisor/blocklists/sigtimedwait_test
@@ -1,0 +1,2 @@
+# TODO: Support `tkill` syscall.
+SigtimedwaitTest.SIGKILLUncaught


book/src/kernel/linux-compatibility/README.md
```diff
@@ -15,7 +15,7 @@ support the loading of Linux kernel modules.
 ## System Calls
 
 At the time of writing,
-Asterinas supports over 230 Linux system calls for the x86-64 architecture,
+Asterinas supports over 233 Linux system calls for the x86-64 architecture,
```
tatetian: Do not update the count until it exceeds 240.

kernel/src/syscall/rt_sigtimedwait.rs
```diff
@@ -0,0 +1,126 @@
...
+pub fn sys_rt_sigtimedwait(
+    set_ptr: Vaddr,
+    info_ptr: Vaddr,
+    timeout_ptr: Vaddr,
+    sigset_size: usize,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    debug!(
+        "set_ptr = 0x{:x}, info_ptr = 0x{:x}, timeout_ptr = 0x{:x}, sigset_size = {}",
+        set_ptr, info_ptr, timeout_ptr, sigset_size
```
StevenJiang1110, line 29: Using `{:#x}` instead of `0x{:x}` if possible. `{:#x}` will add `0x` prefix automatically.

kernel/src/syscall/rt_sigtimedwait.rs
```diff
@@ -0,0 +1,126 @@
...
+                if !block_list.contains(signal.num())
+                    && sig_disposition.get(signal.num()) == SigAction::Ign
+                {
+                    continue;
+                }
+
+                return Ok(SyscallReturn::Return(signal.num().as_u8() as _));
+            }
+            drop(sig_disposition);
+            drop(sig_dispositions);
```
StevenJiang1110: The fast path should be added before `with_sigmask_changed`. It we can dequeue some signal, we can avoid changing the sigmask.

kernel/src/syscall/rt_sigtimedwait.rs
```diff
@@ -0,0 +1,126 @@
...
+                    let siginfo = signal.to_info();
+                    ctx.user_space().write_val(info_ptr, &siginfo)?;
+                }
+
+                // If the signal is ignored and not blocked, the signal will be directly dropped.
+                if !block_list.contains(signal.num())
+                    && sig_disposition.get(signal.num()) == SigAction::Ign
+                {
+                    continue;
+                }
```
StevenJiang1110: If the signal is going to be dropped, we shouldn't write to siginfo？ Therefore, lines 77–81 should be moved above line 71.

kernel/src/syscall/rt_sigtimedwait.rs
```diff
@@ -0,0 +1,126 @@
...
+            let sig_disposition = sig_dispositions.lock();
+            while let Some(signal) = ctx.posix_thread.dequeue_signal(&mask) {
+                if info_ptr != 0 {
+                    let siginfo = signal.to_info();
+                    ctx.user_space().write_val(info_ptr, &siginfo)?;
+                }
+
+                // If the signal is ignored and not blocked, the signal will be directly dropped.
+                if !block_list.contains(signal.num())
+                    && sig_disposition.get(signal.num()) == SigAction::Ign
```
StevenJiang1110: You can use `sig_disposition.will_ignore(&signal)` instead of `sig_disposition.get(signal.num()) == SigAction::Ign` here. There are two cases when a signal will be ignored, the action is `SigAction::Ign`, or the action is `SigAction::Dfl`, and the default action is to ignore the signal.

kernel/src/syscall/rt_sigtimedwait.rs
```diff
@@ -0,0 +1,126 @@
...
+                        while let Some(signal) = ctx.posix_thread.dequeue_signal(&mask) {
+                            // If the signal is ignored and not blocked, the signal will be directly dropped.
+                            if !block_list.contains(signal.num())
+                                && sig_disposition.get(signal.num()) == SigAction::Ign
+                            {
+                                continue;
+                            }
+
+                            return Some(signal);
+                        }
```
StevenJiang1110: It looks like dequeuing a signal and checking the signal whether it is ignored and blocked is repeated twice in this file. Can we define a universal function for them?

StevenJiang1110 (commented):
Thanks for your contribution!

PR #2706, opened 2025-12-09
Pin Nix channels and install cachix

cqs21:


tools/docker/nix/Dockerfile
```diff
@@ -11,11 +11,12 @@ ARG DEBIAN_FRONTEND=noninteractive
 # FIXME: Installing Nix as root is not supported in single-user mode.
```
tatetian, line 13: Add a comment here:

```
# Install Nix package manager and other Nix tools
#
# The nixpkgs and nixos channels are pinned to a particular commit (NixOS 25.05, 2025-07-01) for reproducibility.
```

tools/nixos/install_asterinas.sh
```diff
@@ -85,6 +85,7 @@ cp ${DISTRO_DIR}/aster_configuration.nix ${BUILD_DIR}/etc/nixos
 cp -r ${DISTRO_DIR}/modules ${BUILD_DIR}/etc/nixos
 cp -r ${DISTRO_DIR}/overlays ${BUILD_DIR}/etc/nixos
 
+export PATH=${PATH}:/run/current-system/sw/bin
```
tatetian, line 88: Does this change has anything to do with Docker image or with pinning Nix channels or installing cachix?

cqs21: With the pinned channel, `make build NIXOS=1` fails if nixos-install does not add `/run/current-system/sw/bin` to `PATH`.

tatetian (approved):

PR #2708, opened 2025-12-09
Fix a small case of brk syscall

wyt8:
A small change that not need much time to review.

### DIFF
--- a/kernel/src/process/process_vm/heap.rs
+++ b/kernel/src/process/process_vm/heap.rs
@@ -21,7 +21,7 @@ pub struct Heap {
     /// The heap size limit
     limit: usize,
     /// The current heap highest address
-    current_heap_end: AtomicUsize,
+    current_program_break: AtomicUsize,
 }
 
 impl Heap {
@@ -29,7 +29,7 @@ impl Heap {
         Heap {
             base: USER_HEAP_BASE,
             limit: USER_HEAP_SIZE_LIMIT,
-            current_heap_end: AtomicUsize::new(USER_HEAP_BASE),
+            current_program_break: AtomicUsize::new(USER_HEAP_BASE),
         }
     }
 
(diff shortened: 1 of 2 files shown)


kernel/src/process/process_vm/heap.rs
```diff
@@ -66,12 +66,12 @@ impl Heap {
...
             Some(new_heap_end) => {
-                if new_heap_end > self.base + self.limit {
-                    return_errno_with_message!(Errno::ENOMEM, "heap size limit was met.");
-                }
                 let current_heap_end = self.current_heap_end.load(Ordering::Acquire);
 
-                if new_heap_end <= current_heap_end {
+                // According to the Linux source code, when the `brk` value is more than the
+                // rlimit, the `brk` syscall returns the current `brk` value.
+                // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/mm/mmap.c#L152>
```
tatetian: No, we should not modify the `brk` method this way. Do you realize that after this change, the function always returns `Ok(_)`? If this method always returns `Ok(_)`, then why not change the return type to `Vaddr`?

In the Rust style, whether the API succeeds or fails should reflect on the return value. Returning `Vaddr` or `Ok(_)` even when it actually fails is misleading. In addition, letting a method to be both a getter and setter is also very unconventional in Rust style. So I suggest refactoring the method this way:

1. Break the original `brk` method into two methods, `program_break` and `set_program_break`.
2. Change the return type of `set_program_break` to `core::result::Result<Vaddr, Vaddr>`.
3. Add Rust documentation for the two methods. In particular, explain the meaning of `Err(Vaddr)`.

The implementation of the `brk` system call should utilize the new `program_break` and `set_program_break` methods and conform with Linux's behavior. In particular, when `set_program_break` returns an error, it should return the current break.

----

Since the method names include the term "program break" (as defined by the [man page](https://man7.org/linux/man-pages/man2/brk.2.html)), rename `current_heap_end` to `current_program_break` and `new_heap_end` to `new_program_break` for consistency.

tatetian:
I found that `brk` is missing [its gvisor test](https://github.com/google/gvisor/blob/561ff6d235e64ed97ee1c3fd1d0f543f159d17bb/test/syscalls/linux/brk.cc#L24). Add it to [the test list](https://github.com/asterinas/asterinas/blob/8ce2f9373dc9a81bdae3af3027f8b367c7044fdf/test/src/syscall/gvisor/Makefile#L8).

kernel/src/process/process_vm/heap.rs
```diff
@@ -60,38 +60,61 @@ impl Heap {
         Ok(())
     }
 
-    pub fn brk(&self, new_heap_end: Option<Vaddr>, ctx: &Context) -> Result<Vaddr> {
+    /// Returns the current program break.
+    pub fn program_break(&self) -> Vaddr {
+        self.current_heap_end.load(Ordering::Relaxed)
```
tatetian: Use the term `program_break` (or `brk` for short) consistently rather than `heap_end` throughout this file.

This principle also applies to the `brk.rs` file.

kernel/src/process/process_vm/heap.rs
```diff
@@ -60,53 +60,76 @@ impl Heap {
...
+    pub fn program_break(&self) -> Vaddr {
+        self.current_program_break.load(Ordering::Relaxed)
+    }
+
+    /// Sets the program break to `new_heap_end`.
+    /// Returns the new program break on success, or the current program break on failure.
+    ///
+    /// # Errors
+    /// Returns `Err(current_program_break)` if setting the program break fails. This behavior
+    /// is consistent with the Linux `brk` syscall.
```
lrh2000: Style:
```suggestion
    /// Sets the program break to `new_heap_end`.
    ///
    /// Returns the new program break on success, or the current program break on failure.
    /// This behavior is consistent with the Linux `brk` syscall.
```

 - The first paragraph in the doc comments should contain a summary sentence. Keep it short.
 - If you need to write a `# Errors` section, insert a blank line below:
```rust
/// # Errors
///
/// The error section...
```
 - But the error section just repeats information that we've already covered. I don't think it adds much value.

Issue #2709, opened 2025-12-09
Unexpected panic at `Cursor::pop_level`


vvvvsv:
### Describe the bug

In our previous `mremap` implementation, when the user called `mremap(old_addr, old_size, new_size, MREMAP_MAYMOVE)` with `new_size > old_size`, and there is a free region of length `new_size - old_size` immediately after the old mapping, we would allocate a new free region of length `new_size` and then move the old mapping into it. In fact, we should simply extend the old mapping to length `new_size`.

Our previous implementation caused the memory region to repeatedly "jump" here and there in the address space when the user tried to `mremap` it multiple times, increasing its size by a few pages each time. In this situation, `Cursor::pop_level` will panic at the `debug_assert!(self.level < self.guard_level);`.

### To Reproduce

Due to limited time, I can not give a simple script that can trigger this panic.

1. Checkout to [this branch](https://github.com/vvvvsv/asterinas/tree/nixos_rebuild_remap_panic).
2. Run `make run NIXOS=1 LOG_LEVEL=warn`
3. In qemu, run `source /root/for_rebuild/exports.sh`
4. Edit `/etc/nixos/configuration.nix` and add some packages in `environment.systemPackages`.
5. `nixos-rebuild test`
6. Repeat 4-5.

After a few repeats, this kernel panic will be reproduced. I will first fix the `mremap` behavior to support `nixos-rebuild`, but this looks like a bug in our `Cursor` that will also need to be addressed in the long term.

### Expected behavior

Asterinas is not expected to panic.

### Environment

docker image asterinas/asterinas:0.16.1-20251130
branch [vvvvsv/nixos_rebuild_remap_panic](https://github.com/vvvvsv/asterinas/tree/nixos_rebuild_remap_panic)

### Additional context

Maybe @junyang-zh can help?

### Logs

```
sh-5.2# nixos-rebuild test
[    44.243] WARN : CLONE_SYSVSEM is not supported now
GC Warning: pthread_getattr_np or pthread_attr_getstack failed for main thread
[    57.926] WARN : PID 92: terminating on signal SIGKILL
building Nix...
[    58.149] WARN : CLONE_SYSVSEM is not supported now
GC Warning: pthread_getattr_np or pthread_attr_getstack failed for main thread
[    63.835] WARN : PID 97: terminating on signal SIGKILL
[    63.948] WARN : CLONE_SYSVSEM is not supported now
GC Warning: pthread_getattr_np or pthread_attr_getstack failed for main thread
[...]

PR #2712, opened 2025-12-09
Add a fast path for `Vmar::remap`

vvvvsv:
This PR adds a fast path for `Vmar::remap` when we can allocate a free region right after the old range.

Related to https://github.com/asterinas/asterinas/issues/2709.

### DIFF
--- a/kernel/src/vm/vmar/mod.rs
+++ b/kernel/src/vm/vmar/mod.rs
@@ -333,6 +333,11 @@ impl Vmar {
     ///   they will be unmapped before the move.
     /// - If `new_addr` is `None`, a new range of size `new_size` will be
     ///   allocated, and the original mapping will be moved there.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if `new_addr` is `None` and `new_size <= old_size`.
+    /// Use `resize_mapping` instead in this case.
     pub fn remap(
         &self,
         old_addr: Vaddr,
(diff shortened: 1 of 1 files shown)


vvvvsv:
@boterinas rerun

kernel/src/vm/vmar/mod.rs
```diff
@@ -393,6 +404,20 @@ impl Vmar {
                 &mut rss_delta,
             )?
         } else {
+            debug_assert!(new_size > old_size);
+
+            // Fast path if we can allocate a free region right after the old range.
+            if inner
+                .alloc_free_region_exact(old_range.end, new_size - old_size)
```
tatetian, line 411: Can we rename this method to `can_alloc_free_region_exact`? The method name is really confusing. I thought its return already means memory allocation.

kernel/src/vm/vmar/mod.rs
```diff
@@ -393,6 +404,20 @@ impl Vmar {
                 &mut rss_delta,
             )?
         } else {
+            debug_assert!(new_size > old_size);
+
+            // Fast path if we can allocate a free region right after the old range.
```
tatetian:
```rust
// Fast path: expand the old mapping in place to the new size
```

kernel/src/vm/vmar/mod.rs
vvvvsv: The advantage of `fn alloc_free_region_exact(...) -> Result<Range<Vaddr>>` is that we don't have to repeatedly write `return_errno_with_message!(Errno::EACCES, ...)`, and that the functions `alloc_free_region_exact`, `alloc_free_region_exact_truncate` and `alloc_free_region` have the same return value type. I think it's better to keep this function, at least keep `fn can_alloc_free_region_exact(...) -> Result<()>`.

tatetian: Fair enough. Let's keep it that way--for now.

tatetian (approved):

PR #2716, opened 2025-12-10
Add symlink `/dev/fd -> /proc/self/fd` if systemd is disabled

vvvvsv:
Required for https://github.com/asterinas/asterinas/issues/2672.

### DIFF
--- a/distro/etc_nixos/modules/core.nix
+++ b/distro/etc_nixos/modules/core.nix
@@ -71,6 +71,7 @@ in {
   boot.postBootCommands = ''
     echo "Executing postBootCommands..."
     if [ "${config.aster_nixos.disable-systemd}" = "true" ]; then
+      ln -s /proc/self/fd /dev/fd
       ${config.aster_nixos.stage-2-hook}
     fi
   '';
--- a/kernel/src/fs/path/mod.rs
+++ b/kernel/src/fs/path/mod.rs
@@ -246,13 +246,16 @@ impl Path {
     ///
     /// Returns the newly created child mount on success.
     ///
+    /// This method consumes the `Path`. Use `Path::new_fs_root` to create a new one
+    /// if needed after mounting.
+    ///
     /// # Errors
     ///
     /// Returns `ENOTDIR` if the path is not a directory.
     /// Returns `EINVAL` if attempting to mount on root or if the path is not
     /// in the current mount namespace.
     pub fn mount(
-        &self,
+        self,
         fs: Arc<dyn FileSystem>,
         flags: PerMountFlags,
         ctx: &Context,


kernel/src/device/mod.rs
```diff
@@ -35,6 +40,11 @@ pub fn init_in_first_process(ctx: &Context) -> Result<()> {
     let dev_path = fs_resolver.lookup(&FsPath::try_from("/dev")?)?;
     dev_path.mount(RamFs::new(), PerMountFlags::default(), ctx)?;
 
+    // Add symlink `/dev/fd -> /proc/self/fd`
+    let dev_path = fs_resolver.lookup(&FsPath::try_from("/dev")?)?;
```
cchanging: Here I think you can get the root path of the new fs directly:
```rust
let mount = dev_path.mount(RamFs::new(), PerMountFlags::default(), ctx)?;

let dev_root = Path::new_fs_root(mount);
let devfd_path = dev_root.new_fs_child("fd", InodeType::SymLink, mkmod!(a+rwx))?;
```

kernel/src/device/mod.rs
```diff
@@ -33,7 +38,12 @@ pub fn init_in_first_process(ctx: &Context) -> Result<()> {
 
     // Mount devtmpfs.
     let dev_path = fs_resolver.lookup(&FsPath::try_from("/dev")?)?;
-    dev_path.mount(RamFs::new(), PerMountFlags::default(), ctx)?;
+    let mount = dev_path.mount(RamFs::new(), PerMountFlags::default(), ctx)?;
+
+    // Add symlink `/dev/fd -> /proc/self/fd`
+    let dev_root = Path::new_fs_root(mount);
+    let devfd_path = dev_root.new_fs_child("fd", InodeType::SymLink, mkmod!(a+rwx))?;
```
tatetian:
```
mkmod!(a+rwx)
```

Are you sure that the permissions are so relaxed?

vvvvsv: Sorry, I just checked the code. Linux kernel does not create this symlink, it is created by systemd. 

https://github.com/systemd/systemd/blob/a07b184e8ccbecafc1fce5a0cfffe87c6a497134/src/shared/dev-setup.c#L19-L25

I will only add this symlink if `NIXOS_DISABLE_SYSTEMD` is true.

vvvvsv:
Do we need to support `nixos-rebuild` when `NIXOS_DISABLE_SYSTEMD=true`? If not, this PR can be closed.

tatetian:
> Do we need to support `nixos-rebuild` when `NIXOS_DISABLE_SYSTEMD=true`? If not, this PR can be closed.

I don't think so. Let's close it.

PR #2719, opened 2025-12-10
Remove env variables from distro configuration

StevenJiang1110:
This PR refactors distro-related scripts and Nix configuration files. The primary goal is to remove environment variable dependencies from the Nix files so that the same configuration can be reused by both nixos-install and nixos-rebuild.

To simplify the problem, now nixos-install will not rely on any files on the host except the kernel. The resolv.conf and stage-1-init script has been prepared by some Nix expressions.

In addition, it changes the installation script (install_nixos.sh) to be generated from a Nix expression. This makes the script suitable for use in both the ISO environment and the development container.


distro/etc_nixos/modules/core.nix
```diff
@@ -1,7 +1,49 @@
...
+    fi
+
+    if [ -z "$NEW_ROOT" ] || [ -z "$NEW_INIT" ]; then
+      echo "Error: 'root=' and 'init=' parameters are required."
+      exit 1
+    fi
+
+    mkdir /sysroot
+    mount -t ext2 "$NEW_ROOT" /sysroot
+    mount -t proc none /sysroot/proc
```
tatetian, line 45: I am a bit of confused here: `/sysroot/proc` is non-existential. Doesn't `mount` require the target directory to be created first?

distro/configuration.nix
```diff
@@ -10,6 +10,10 @@
   imports = [ ./aster_configuration.nix ];
```
tatetian: I don't understand. The original `aster_configuration.nix` file has been removed. So how can Nix find the file referred to by this line? In addition, it seems that the original file is moved to `install_script/`. But the Git history shows that the original file is removed. You should have used `git mv` to tell Git that the file is moved, not deleted, so that Git can show the differences in `install_script/aster_configuration.nix`. Now this entire file appears to be new...

distro/install_script/default.nix
```diff
@@ -0,0 +1,38 @@
+{ disable-systemd ? "false", stage-2-hook ? "/bin/sh -l", log-level ? "error"
+, console ? "hvc0", test-command ? "", pkgs ? import <nixpkgs> { } }:
+let
+  aster-kernel = builtins.path {
+    path = ../../target/osdk/iso_root/boot/aster-nix-osdk-bin;
+  };
+  distro = builtins.path { path = ../../distro; };
+
+  aster_configuration = pkgs.replaceVarsWith {
+    src = ./aster_configuration.nix;
```
tatetian: Rename this file to make it more apparent that this file is a template. Some candidate names are `./aster_configuration.nix.template` (will this affect syntax highlights?) and `./aster_configuration_template.nix`.

distro/install_script/install_nixos.sh.template
```diff
@@ -0,0 +1,136 @@
...
+
+# Default values
+CONFIG_PATH=""
+DISK=""
+FORCE_FORMAT_DISK=false
+
+# Function to display help
+show_help() {
+    cat << EOF
+Usage: $0 [OPTIONS]
```
tatetian: The following form is more accurate:

```
Usage:
$0 --config <CONFIG_PATH> --disk <DISK> [--force-format-disk]
$0 [-h | --help]
```

distro/install_script/default.nix
```diff
@@ -0,0 +1,38 @@
...
+    replacements = {
+      aster-configuration = aster_configuration;
+      aster-distro = distro;
+      aster-kernel = aster-kernel;
+    };
+    isExecutable = true;
+  };
+
+in pkgs.stdenv.mkDerivation {
+  name = "nixos_install_script";
```
tatetian: The package will be visible to the entire to a greater audience than the Asterinas developers. So the name should include Asterinas. For example, you can name it `install_aster_nixos` or `aster_nixos_installer`.

distro/install_script/default.nix
```diff
@@ -0,0 +1,38 @@
...
+  buildCommand = ''
+    mkdir -p $out/{bin,nixos}
+    cp ${install_nixos} $out/bin/install_nixos.sh
```
tatetian: Name this script `install_aster_nixos.sh`. It will be visible and used by the user of the ISO installer, right? So let's emphasize that it is installing AsterNixOS, rather than NixOS.

distro/install_script/default.nix
```diff
@@ -0,0 +1,37 @@
+{ disable-systemd ? "false", stage-2-hook ? "/bin/sh -l", log-level ? "error"
+, console ? "hvc0", test-command ? "", pkgs ? import <nixpkgs> { } }:
+let
+  aster-kernel = builtins.path {
+    path = ../../target/osdk/iso_root/boot/aster-nix-osdk-bin;
+  };
+  distro = builtins.path { path = ../../distro; };
```
tatetian: This looks strange to me. First, `install_script/default.nix` itself is under the `distro/` directory. So this script is relying on itself... Second, it is unclear what is the relationships between the Nix files under the `distro/` directory. The situation would be worse when ISO code is added. I suggest reorganizing the directory this way:

```
distro/
    etc_nixos/
        configuration.nix
        modules/
            ...
        overlays/
            ...
    install_script/
```

distro/install_script/default.nix
StevenJiang1110: If naming it as `./aster_configuration.nix.template`, nixfmt will not format this file since the filename does not end with `.nix`. Renaming it as `./aster_configuration_template.nix ` now.

distro/install_script/default.nix
```diff
@@ -0,0 +1,41 @@
...
+      aster-kernel = aster-kernel;
+      aster-disable-systemd = disable-systemd;
+      aster-stage-2-hook = stage-2-hook;
+      aster-log-level = log-level;
+      aster-console = console;
+      aster-test-command = test-command;
+    };
+  };
+  install_aster_nixos = pkgs.replaceVarsWith {
+    src = ./install_nixos_template.sh;
```
tatetian: Create a new `templates/` directory under `install_script/` and move the two templates, one script and one Nix file, to it. As the parent directory is called `templates/`, the two file names do not need to include the `_template` suffixes.

distro/aster_nixos_installer/default.nix
```diff
@@ -0,0 +1,41 @@
...
+    src = ./install_nixos_template.sh;
+    replacements = {
+      aster-configuration = aster_configuration;
+      aster-etc-nixos = etc-nixos;
+    };
+    isExecutable = true;
+  };
+
+in pkgs.stdenv.mkDerivation {
+  name = "aster_nixos_installer";
```
tatetian, line 31: For consistency, rename the `install_script` directory to `aster_nixos_installer`.

PR #2720, opened 2025-12-10
Fix the wrong behavior of TCSETSW flag

cchanging:
We currently call flag TCSETSW `SetTermiosDrain` in Asterinas, but this makes no sense. Flag TCSETSW in Linux has a wait semantic, not a drain semantic. This misuse is the root cause of the current pty output issues. 

Since we currently lack related mechanisms to support wait, this fix does not introduce additional behavior for this flag.

### DIFF
--- a/kernel/src/device/pty/driver.rs
+++ b/kernel/src/device/pty/driver.rs
@@ -146,11 +146,6 @@ impl TtyDriver for PtyDriver {
         Ok(len)
     }
 
-    fn drain_output(&self) {
-        self.output.lock().clear();
-        self.pollee.invalidate();
-    }
-
     fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
         let mut output = self.output.lock();
         let mut has_notified = false;
--- a/kernel/src/device/pty/master.rs
+++ b/kernel/src/device/pty/master.rs
@@ -170,7 +170,7 @@ impl FileIo for PtyMaster {
         use crate::{device::tty::ioctl_defs::*, fs::utils::ioctl_defs::GetNumBytesToRead};
 
         dispatch_ioctl!(match raw_ioctl {
-            GetTermios | SetTermios | SetTermiosDrain | SetTermiosFlush | GetWinSize
+            GetTermios | SetTermios | SetTermiosWait | SetTermiosFlush | GetWinSize
             | SetWinSize | GetPtyNumber => {
                 return self.slave.ioctl(raw_ioctl);
             }
--- a/kernel/src/device/tty/driver.rs
+++ b/kernel/src/device/tty/driver.rs
@@ -36,9 +36,6 @@ pub trait TtyDriver: Send + Sync + 'static {
     /// pushed.
     fn push_output(&self, chs: &[u8]) -> Result<usize>;
 
-    /// Drains the output buffer.
-    fn drain_output(&self);
-
     /// Returns a callback function that echoes input characters to the output buffer.
     ///
     /// Note that the implementation may choose to hold a lock during the life of the callback.
--- a/kernel/src/device/tty/ioctl_defs.rs
+++ b/kernel/src/device/tty/ioctl_defs.rs
@@ -10,7 +10,7 @@ use crate::util::ioctl::{InData, OutData, PassByVal, ioc};
 
 pub type GetTermios      = ioc!(TCGETS,     0x5401,     OutData<CTermios>);
 pub type SetTermios      = ioc!(TCSETS,     0x5402,     InData<CTermios>);
-pub type SetTermiosDrain = ioc!(TCSETSW,    0x5403,     InData<CTermios>);
+pub type SetTermiosWait  = ioc!(TCSETSW,    0x5403,     InData<CTermios>);
 pub type SetTermiosFlush = ioc!(TCSETSF,    0x5404,     InData<CTermios>);
 
 pub type GetWinSize      = ioc!(TIOCGWINSZ, 0x5413,     OutData<CWinSize>);
--- a/kernel/src/device/tty/mod.rs
+++ b/kernel/src/device/tty/mod.rs
@@ -320,24 +320,29 @@ impl<D: TtyDriver> Tty<D> {
                 self.driver().on_termios_change(old_termios, &termios);
                 ldisc.set_termios(termios);
             }
-            cmd @ SetTermiosDrain => {
+            cmd @ SetTermiosWait => {
                 let termios = cmd.read()?;
 
+                // TODO: If applicable, wait for the output buffer to drain. For now, we don't need
+                // to do anything here because:
+                //  - Linux does not consider a pty to have an output buffer, so it does not drain
+                //    it. See
+                //    <https://elixir.bootlin.com/linux/v5.10.247/source/drivers/tty/pty.c#L137-L148>.
+                //  - We don't currently have an output buffer for other TTYs.
                 let mut ldisc = self.ldisc.lock();
                 let old_termios = ldisc.termios();
                 self.driver().on_termios_change(old_termios, &termios);
                 ldisc.set_termios(termios);
-                self.driver.drain_output();
             }
             cmd @ SetTermiosFlush => {
                 let termios = cmd.read()?;
 
+                // TODO: If applicable, wait for the output buffer to drain. (See comments above.)
                 let mut ldisc = self.ldisc.lock();
                 let old_termios = ldisc.termios();
                 self.driver().on_termios_change(old_termios, &termios);
                 ldisc.set_termios(termios);
                 ldisc.drain_input();
-                self.driver.drain_output();
 
                 self.pollee.invalidate();
             }
--- a/kernel/src/device/tty/n_tty.rs
+++ b/kernel/src/device/tty/n_tty.rs
@@ -46,8 +46,6 @@ impl TtyDriver for VtDriver {
         Ok(chs.len())
     }
 
-    fn drain_output(&self) {}
-
     fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
         |chs| self.console.send(chs)
     }
@@ -87,8 +85,6 @@ impl TtyDriver for HvcDriver {
         self.console.send(chs);
         Ok(chs.len())
     }
-
-    fn drain_output(&self) {}
 
     fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
         |chs| self.console.send(chs)


lrh2000:
This issue exists for `SetTermiosFlush` as well. We should to fix `TtyDriver::drain_output` instead to correctly implement the waiting behavior.

> Since we currently lack related mechanisms to support wait,

Why? For pty, using the pollee is sufficient. For a TTY, we don't need to do anything because there is no output buffer.
https://github.com/asterinas/asterinas/blob/c292ccdccaf0c558ecad4f7b34b5dd15aeab6d15/kernel/src/device/pty/driver.rs#L145
https://github.com/asterinas/asterinas/blob/c292ccdccaf0c558ecad4f7b34b5dd15aeab6d15/kernel/src/device/tty/n_tty.rs#L91

---


We currently do lack a mechanism to avoid race conditions between the termios changes and the wait. However, this is not a problem since we currently do not support output flags in termios:
 https://github.com/asterinas/asterinas/blob/c292ccdccaf0c558ecad4f7b34b5dd15aeab6d15/kernel/src/device/pty/driver.rs#L130-L132

kernel/src/device/pty/driver.rs
```diff
@@ -216,3 +224,10 @@ impl TtyDriver for PtyDriver {
         }
     }
 }
+
+impl Pollable for PtyDriver {
+    fn poll(&self, mask: IoEvents, poller: Option<&mut PollHandle>) -> IoEvents {
+        self.pollee
+            .poll_with(mask, poller, || self.check_io_events())
```
lrh2000: The I/O events cannot be defined for the same pollee (with different meanings!) twice. The I/O events are already specified here:
https://github.com/asterinas/asterinas/blob/c292ccdccaf0c558ecad4f7b34b5dd15aeab6d15/kernel/src/device/pty/master.rs#L87-L94
https://github.com/asterinas/asterinas/blob/c292ccdccaf0c558ecad4f7b34b5dd15aeab6d15/kernel/src/device/pty/master.rs#L62-L84

```rust
     if self.slave().can_push() { 
         events |= IoEvents::OUT; 
     } 
```
```rust
        if self.output.lock().is_empty() {
            events |= IoEvents::OUT;
        }
```

The two code snippets will conflict with either other and cannot make sense.

lrh2000:
I'm sorry, but based on the changes you made here, it seems that you don't understand how our event notification mechanism works.

Here is my version that implements the waiting behavior:
<details>

<summary> Waiting to drain the ouput buffer </summary>

```patch
diff --git a/kernel/src/device/pty/driver.rs b/kernel/src/device/pty/driver.rs
index 39b430ba8..7499be7eb 100644
--- a/kernel/src/device/pty/driver.rs
+++ b/kernel/src/device/pty/driver.rs
@@ -114,9 +114,8 @@ impl TtyDriver for PtyDriver {
         Ok(len)
     }
 
-    fn drain_output(&self) {
+    fn has_output(&self) -> bool {
         self.output.lock().clear();
-        self.pollee.invalidate();
     }
 
     fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
diff --git a/kernel/src/device/pty/master.rs b/kernel/src/device/pty/master.rs
index 4d6e2a421..37d50b4a7 100644
--- a/kernel/src/device/pty/master.rs
+++ b/kernel/src/device/pty/master.rs
@@ -161,7 +161,7 @@ impl FileIo for PtyMaster {
         use crate::{device::tty::ioctl_defs::*, fs::utils::ioctl_defs::GetNumBytesToRead};
 
-            GetTermios | SetTermios | SetTermiosDrain | SetTermiosFlush | GetWinSize
+            GetTermios | SetTermios | SetTermiosWait | SetTermiosFlush | GetWinSize
             | SetWinSize | GetPtyNumber => {
                 return self.slave.ioctl(raw_ioctl);
             }
diff --git a/kernel/src/device/tty/driver.rs b/kernel/src/device/tty/driver.rs
index d62c59dfc..bdaba59d7 100644
--- a/kernel/src/device/tty/driver.rs
+++ b/kernel/src/device/tty/driver.rs
@@ -32,8 +32,8 @@ pub trait TtyDriver: Send + Sync + 'static {
     /// pushed.
     fn push_output(&self, chs: &[u8]) -> Result<usize>;
 
-    /// Drains the output buffer.
-    fn drain_output(&self);
+    /// Returns whether the output buffer is empty.
+    fn has_output(&self) -> bool;
 
     /// Returns a callback function that echoes input characters to the output buffer.
     ///
diff --git a/kernel/src/device/tty/ioctl_defs.rs b/kernel/src/device/tty/ioctl_defs.rs
index 364a03eb1..2bc96dd01 100644
--- a/kernel/src/device/tty/ioctl_defs.rs
+++ b/kernel/src/device/tty/ioctl_defs.rs
@@ -10,7 +10,7 @@ use crate::util::ioctl::{ioc, InData, OutData, PassByVal};
 
 pub type GetTermios      = ioc!(TCGETS,     0x5401,     OutData<CTermios>);
 pub type SetTermios      = ioc!(TCSETS,     0x5402,     InData<CTermios>);
-pub type SetTermiosDrain = ioc!(TCSETSW,    0x5403,     InData<CTermios>);
+pub type SetTermiosWait  = ioc!(TCSETSW,    0x5403,     InData<CTermios>);
 pub type SetTermiosFlush = ioc!(TCSETSF,    0x5404,     InData<CTermios>);
 
 pub type GetWinSize      = ioc!(TIOCGWINSZ, 0x5413,     OutData<CWinSize>);
diff --git a/kernel/src/device/tty/mod.rs b/kernel/src/device/tty/mod.rs
index a04deb559..ce062154f 100644
--- a/kernel/src/device/tty/mod.rs
+++ b/kernel/src/device/tty/mod.rs
@@ -108,7 +108,12 @@ impl<D> Tty<D> {
     /// Notifies that the output buffer now has room for new characters.
     ///
     /// This method should be called when the state of [`TtyDriver::can_push`] changes from `false`
-    /// to `true`.
+    /// to `true` or when the state of [`TtyDriver::has_output`] changes from `true` to `false`.
+    ///
+    /// The easiest way to meet the above requirements is to call the method each time a new room
+    /// appears in the output buffer.
+    //
+    // Note: The second requirement is necessary for `wait_output_to_drain` to work.
     pub(super) fn notify_output(&self) {
         self.pollee.notify(IoEvents::OUT);
     }
@@ -157,6 +162,18 @@ impl<D: TtyDriver> Tty<D> {
         Ok(len)
     }
 
+    fn wait_output_to_drain(&self) -> Result<()> {
+        // I/O events are edge-triggered and `notify_input()` is required to be called whenever
+        // there is a new room in the output buffer.
+        self.wait_events(IoEvents::OUT, None, || {
+            if self.driver().has_output() {
+                return_errno_with_message!(Errno::EAGAIN, "the output buffer is not empty");
+            } else {
+                Ok(())
+            }
+        })
+    }
+
     fn check_io_events(&self) -> IoEvents {
         let mut events = IoEvents::empty();
 
@@ -318,21 +335,29 @@ impl<D: TtyDriver> Tty<D> {
 
                 self.ldisc.lock().set_termios(termios);
             }
-            cmd @ SetTermiosDrain => {
+            cmd @ SetTermiosWait => {
                 let termios = cmd.read()?;
 
+                // FIXME: There are races between waiting and termios changes. This doesn't matter
+                // now because we never care about the output termios when writing to the output
+                // buffer. However, this race should be fixed once we support output termios flags.
+                self.wait_output_to_drain()?;
+
                 let mut ldisc = self.ldisc.lock();
                 ldisc.set_termios(termios);
-                self.driver.drain_output();
             }
             cmd @ SetTermiosFlush => {
                 let termios = cmd.read()?;
 
+                // FIXME: There are races between waiting and termios changes. This doesn't matter
+                // now because we never care about the output termios when writing to the output
+                // buffer. However, this race should be fixed once we support output termios flags.
+                self.wait_output_to_drain()?;
+
                 let mut ldisc = self.ldisc.lock();
                 ldisc.set_termios(termios);
-                ldisc.drain_input();
-                self.driver.drain_output();
 
+                ldisc.drain_input();
                 self.pollee.invalidate();
             }
             cmd @ GetWinSize => {
diff --git a/kernel/src/device/tty/n_tty.rs b/kernel/src/device/tty/n_tty.rs
index 374fedcb9..bb6e73bdc 100644
--- a/kernel/src/device/tty/n_tty.rs
+++ b/kernel/src/device/tty/n_tty.rs
@@ -46,7 +46,9 @@ impl TtyDriver for VtDriver {
         Ok(chs.len())
     }
 
-    fn drain_output(&self) {}
+    fn has_output(&self) -> bool {
+        false
+    }
 
     fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
         |chs| self.console.send(chs)
@@ -86,7 +88,9 @@ impl TtyDriver for HvcDriver {
         Ok(chs.len())
     }
 
-    fn drain_output(&self) {}
+    fn has_output(&self) -> bool {
+        false
+    }
 
     fn echo_callback(&self) -> impl FnMut(&[u8]) + '_ {
         |chs| self.console.send(chs)
diff --git a/test/src/apps/pty/pty_blocking.c b/test/src/apps/pty/pty_blocking.c
index a0ee670ff..eeca587cb 100644
--- a/test/src/apps/pty/pty_blocking.c
+++ b/test/src/apps/pty/pty_blocking.c
@@ -127,3 +127,15 @@ FN_TEST(pty_blocking)
 	// constructing that state is difficult without first knowing the size of the input buffer.
 }
 END_TEST()
+
+FN_TEST(drain)
+{
+	struct termios term;
+
+	TEST_SUCC(tcgetattr(slave, &term));
+	TEST_SUCC(tcsetattr(slave, TCSADRAIN, &term));
+
+	TEST_SUCC(write_repeat(slave, 'x', 1024));
+	TEST_SUCC(tcsetattr(master, TCSADRAIN, &term));
+}
+END_TEST()
```

</details>

However, after implementing it, I found that the pty ioctl does not actually block in Linux (regression tests are important!). Upon checking the behavior of Linux, I discovered that it always reports the pty's output buffer as empty regardless of whether the master has read the data: https://elixir.bootlin.com/linux/v5.10.247/source/drivers/tty/pty.c#L137-L148

So we should just do nothing and say something like:
```patch
diff --git a/kernel/src/device/tty/mod.rs b/kernel/src/device/tty/mod.rs
index a04deb559..04fbde7df 100644
--- a/kernel/src/device/tty/mod.rs
+++ b/kernel/src/device/tty/mod.rs
@@ -321,17 +321,22 @@ impl<D: TtyDriver> Tty<D> {
             cmd @ SetTermiosDrain => {
                 let termios = cmd.read()?;
 
+                // TODO: If applicable, wait for the output buffer to drain. For now, we don't need
+                // to do anything here because:
+                //  - Linux does not consider a pty to have an output buffer, so it does not drain
+                //    it. See
+                //    <https://elixir.bootlin.com/linux/v5.10.247/source/drivers/tty/pty.c#L137-L148>.
+                //  - We don't currently have an output buffer for other TTYs.
                 let mut ldisc = self.ldisc.lock();
                 ldisc.set_termios(termios);
-                self.driver.drain_output();
             }
             cmd @ SetTermiosFlush => {
                 let termios = cmd.read()?;
 
+                // TODO: If applicable, wait for the output buffer to drain. (See comments above.)
                 let mut ldisc = self.ldisc.lock();
                 ldisc.set_termios(termios);
                 ldisc.drain_input();
-                self.driver.drain_output();
 
                 self.pollee.invalidate();
             }
```

cchanging:
> However, after implementing it, I found that the pty ioctl does not actually block in Linux (regression tests are important!). Upon checking the behavior of Linux, I discovered that it always reports the pty's output buffer as empty regardless of whether the master has read the data: https://elixir.bootlin.com/linux/v5.10.247/source/drivers/tty/pty.c#L137-L148

Thanks for your explanations!

lrh2000 (approved):

PR #2723, opened 2025-12-10
Rewrite the README of `sctrace`

tatetian:
This PR completely rewrites the README of the `sctrace` tool to enhance its quality to the level expected by a wider audience.

See the [rendered version](https://github.com/tatetian/asterinas/tree/rewrite_sctrace_readme/tools/sctrace).

### DIFF
--- a/tools/sctrace/README.md
+++ b/tools/sctrace/README.md
@@ -1,189 +1,268 @@
 # Syscall Compatibility Tracer
 
-Syscall Compatibility Tracer (`sctrace`) is a powerful system call compatibility
-verification tool that analyzes and validates system call against user-defined
-patterns. Written in
-[SCML (System Call Matching Language)](https://asterinas.github.io/book/kernel/linux-compatibility/syscall-flag-coverage/system-call-matching-language.html),
-these patterns describe supported functionality of system calls.
-`sctrace` supports both real-time monitoring of running programs and post-analysis of
-existing trace logs, providing comprehensive insights into system call compatibility
-with intuitive pattern matching and visual feedback.
+Syscall Compatibility Tracer (`sctrace`) is an oracle
+to answer the following questions
+for developers and users of a Linux ABI-compatible OS
+(such as [Asterinas](https://github.com/asterinas/asterinas)):
+**Is a target Linux application supported by the OS?
+If not, where are the gaps?**
 
-## Features
+## Motivation
 
-- **Pattern-based filtering**: Define system call patterns using SCML syntax
-- **Dual mode operation**:
-  - Online mode: Real-time tracing of running programs
-  - Offline mode: Analysis of existing strace log files
-- **Multi-threaded support**: Automatic handling of multi-threaded program traces with syscall reconstruction.
-When tracing multi-threaded programs, strace may split system calls across multiple lines due to thread interleaving.
-`sctrace` automatically handles this reconstruction.
-- **Multiple SCML files support**: Specify multiple `.scml` files as arguments to load all of them.
-Each file maintains its own scope for bitflags and struct definitions, preventing cross-file pollution.
+There are tons of Linux ABI-compatible OSes out there:
+some of them have been deployed at scale
+(e.g., [HongMeng Kernel](https://www.usenix.org/conference/osdi24/presentation/chen-haibo) and [gVisor](https://gvisor.dev/)),
+some are in rapid development
+(e.g., [Asterinas](https://github.com/asterinas/asterinas)),
+some target niche markets
+(e.g., [Occlum](https://github.com/occlum/occlum)),
+some are research prototypes
+(e.g., [Graphene](https://grapheneproject.io/)),
+and some are just hobby projects
+(e.g., [Maestro](https://github.com/maestro-os/maestro)).
+They all provide a subset of Linux system calls and features
+so that at least some Linux applications can run on them _unmodified_.
 
-## How to build and install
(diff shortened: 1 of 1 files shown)


tools/sctrace/README.md
```diff
@@ -1,189 +1,254 @@
...
+sctrace open.scml -- cat open.scml
 ```
 
-**Note**: When generating strace logs for offline analysis, use `-yy` and `-f` flags:
+The output would look like below:
 
-```bash
-strace -yy -f -o trace.log ls -la
+```
+// TODO: copy a portion of the sctrace output here
```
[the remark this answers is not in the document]
tatetian: @taosue Could you copy (a select portion of) the real output of `sctrace open.scml -- cat open.scml` here?

tools/sctrace/README.md
```diff
@@ -1,189 +1,250 @@
...
+we now use it to track the execution of a simple command
+that prints out the content of `open.scml`:
 
 ```bash
-sctrace pattern1.scml pattern2.scml --input trace.log
-sctrace file_ops.scml network.scml --input trace.log --quiet
+sctrace open.scml -- cat open.scml
 ```
 
-**Note**: When generating strace logs for offline analysis, use `-yy` and `-f` flags:
```
taosue, line 91: When generating strace logs for offline mode, we should add `-yy` and `-f` flags to be compatible with online analysis. Should we keep this note?

tools/sctrace/README.md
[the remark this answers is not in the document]
taosue: Sure, I just select the first part of the output. Please let me know if it is not enough.

```
1045884 execve("/usr/bin/cat", ["cat", "open.scml"], 0x5d08ad413588 /* 25 vars */)           = 0 (unsupported)
1045884 brk(NULL)                       = 0x5a3f59f86000 (unsupported)
1045884 mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7d321cf1f000 (unsupported)
1045884 access("/etc/ld.so.preload", R_OK) = -1 ENOENT (No such file or directory) (unsupported)
1045884 openat(AT_FDCWD</home/sutao>, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 5</etc/ld.so.cache>
1045884 fstat(5</etc/ld.so.cache>, {st_mode=S_IFREG|0644, st_size=39847, ...}) = 0 (unsupported)
1045884 mmap(NULL, 39847, PROT_READ, MAP_PRIVATE, 5</etc/ld.so.cache>, 0) = 0x7d321cf15000 (unsupported)
1045884 close(5</etc/ld.so.cache>)      = 0 (unsupported)
1045884 openat(AT_FDCWD</home/sutao>, "/lib/x86_64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 5</usr/lib/x86_64-linux-gnu/libc.so.6>
1045884 read(5</usr/lib/x86_64-linux-gnu/libc.so.6>, "\177ELF\2\1\1\3\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\220\243\2\0\0\0\0\0"..., 832) = 832 (unsupported)
```

tools/sctrace/README.md
```diff
@@ -1,189 +1,250 @@
...
+// TODO: copy a portion of the sctrace output here
 ```
 
-- `-yy`: Print paths associated with file descriptor arguments
-- `-f`: Trace child processes created by fork/vfork/clone
+According to the SCML file,
+only specific patterns of `open` are supported.
+All other system calls, and the `read`/`write` calls used by `cat`,
+are not defined in `open.scml`,
+so `sctrace` appends an `[unsupported]` tag to almost every system call entry.
```
taosue: Currently `sctrace` appends `(unsupported)`, should I change to `[unsupported]` in the `sctrace`?

tools/sctrace/README.md
```diff
@@ -1,189 +1,250 @@
...
+In the Asterinas development Docker image,
+we have pre-installed the `sctrace` tool
+along with a convenient wrapper script, `sctrace.sh`,
+that automatically finds the SCML files for you.
+So its syntax is made even simpler:
 
-`sctrace` provides colored output to distinguish between supported and unsupported system calls:
+```bash
+sctrace.sh <prog> <arg>...
+```
```
taosue, line 236: The behavior of `strace.sh` is mostly the same as `sctrace`. But `sctrace.sh` first adds all SCML files in the Asterinas and makes `scml_file` be optional, so the command should be like:
```
sctrace.sh [scml_file]... [--quiet] -- <prog> <arg>...
sctrace [scml_file]... [--quiet] --input <strace_log>
```

tools/sctrace/README.md
tatetian: Sure.

tools/sctrace/README.md
tatetian: No, I will fix the README.

tools/sctrace/README.md
tatetian: Yes, your description is more accurate. But I found using `sctrace.sh` weird for the most common cases where the user does not want to provide any extra SCML files.

```bash
sctrace.sh -- <prog> <arg>
```

The user would wonder: "why do I need to add `--`?". And this is inconsistent with `sctrace` because it does not allow zero SCML files.

PR #2725, opened 2025-12-10
Add `/proc/[pid]/maps`

vvvvsv:
Required for https://github.com/asterinas/asterinas/issues/2672.

This PR only adds the `... [stack]` line in `/proc/[pid]/maps`. I will support printing the status of non-stack mappings in the future.

### DIFF
--- a/kernel/src/vm/vmar/vm_mapping.rs
+++ b/kernel/src/vm/vmar/vm_mapping.rs
@@ -7,6 +7,7 @@ use core::{
 };
 
 use align_ext::AlignExt;
+use aster_util::printer::VmPrinter;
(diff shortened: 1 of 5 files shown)


kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -200,6 +201,54 @@ impl VmMapping {
         Ok(())
     }
 
+    /// Prints the mapping information to `/proc/[pid]/maps`.
```
tatetian: A slightly better writing:

```rust
/// Prints the mapping information in the format of `/proc/[pid]/maps`.
```

PR #2726, opened 2025-12-11
Clarify the UTS names in `/proc/version`

lrh2000:
https://github.com/asterinas/asterinas/blob/4013104c12a0d6b0f28524ea5ba5cb9659fe39e3/kernel/src/fs/procfs/version.rs#L34-L39

The code is confusing because it leads readers to believe that the values come from the initial UTS namespace, even though the values in the current UTS namespace may differ.

However, this is not true. The value is immutable and the same in all UTS namespaces because no system calls can change it.

To clarify, let's define the constants as `UtsName::SYSNAME`, etc.


StevenJiang1110 (approved):

PR #2727, opened 2025-12-11
Fix atomic mode break and sort cgroup attributes

lrh2000:
https://github.com/asterinas/asterinas/blob/4013104c12a0d6b0f28524ea5ba5cb9659fe39e3/kernel/src/process/clone.rs#L294-L300

This will break atomic mode because `move_process_to_node` (that will lock mutexes) cannot be called with preemption disabled (under the RCU guard `cgroup().get()`).

So we'll have to clone the cgroup node here.

---

Also, I don't understand the order of the attributes in `kernel/src/fs/cgroupfs/systree_node.rs`. Are they ordered arbitrarily, or according to rules of which I am unaware? This PR proposes ordering them alphabetically.


cchanging (approved):

PR #2729, opened 2025-12-11
Pre-install `sctrace` in Docker image

taosue:
This PR pre-installs `sctrace` and sets `ASTER_SCML` environment variable to save the paths of all Asterinas SCML files in the Docker image.


.github/workflows/publish_docker_images.yml
```diff
@@ -62,6 +62,12 @@ jobs:
         run: |
           ./tools/github_workflows/prepare_for_docker_build_and_push.sh ${{ secrets.DOCKERHUB_USERNAME }} ${{ secrets.DOCKERHUB_TOKEN }} asterinas
 
+      - name: Find Asterinas SCML files
+        id: find-aster-scml
+        run: |
+          ASTER_SCML=$(find ./book/src/kernel/linux-compatibility/ -name "*.scml" -type f | sed 's|^\./|/root/asterinas/|' | tr '\n' ' ')
```
tatetian: Do not evaluate the `ASTER_SCML` value at the creation of the Docker image. If more SCML files are added (which is very likely to happen) after the Docker image is released, then these newly-added SCML files won't appear in `ASTER_SCML`. This would be a big surprise to the user.

I think a better approach is to evaluate the SCML file in `/etc/profile` or `/root/.bashrc`:

```
ASTER_SCML=$(find /root/asterinas/book/src/kernel/linux-compatibility/ -name "*.scml" -type f | sed 's|^\./|/root/asterinas/|' | tr '\n' ' ')
```

tools/docker/Dockerfile
```diff
@@ -75,7 +75,10 @@ RUN apt update && apt-get install -y --no-install-recommends \
 RUN apt clean && rm -rf /var/lib/apt/lists/*
 
 # Install sctrace
-RUN ln -s /root/asterinas/tools/sctrace.sh /usr/local/bin/sctrace
+RUN cargo install sctrace
```
tatetian: On a second thought, installing `sctrace` in the Dockerfile won't work. Our release process is comprised of two steps: (1) generating the new Docker image and (2) launching the new release, including uploading crates to crates.io. So if we do as proposed by this PR, a Docker image for `0.(MINOR + 1).0` version would include the sctrace of version `0.MINOR.LAST_PATCH`.

Do I guess that the acceptable solution is to
1. Create a wrapper called `sctrace`, which calls `cargo run --manifest ... `. This is almost equivalent to installing sctrace.
2. Install the wrapper in the Dockerfile.
3. Export the `ASTER_SCML` env var in the Dockerfile

taosue: Totally agree, thanks for pointing out!

tools/docker/Dockerfile
```diff
@@ -77,6 +77,9 @@ RUN apt clean && rm -rf /var/lib/apt/lists/*
 # Install sctrace
 RUN ln -s /root/asterinas/tools/sctrace.sh /usr/local/bin/sctrace
 
+# Dynamically export `ASTER_SCML` with all SCML file paths at shell startup
+RUN echo 'export ASTER_SCML=$(find /root/asterinas/book/src/kernel/linux-compatibility/ -name "*.scml")' >> /root/.bashrc
```
tatetian, line 82:
```Dockerfile
# Dynamically export `ASTER_SCML` with all SCML file paths at shell startup
```

This is a worthless comment. Instead of explaining what the command does (which is obvious), explain the why.

```Dockerfile
# Set the `ASTER_SCML` env var to make sctrace easier to use for Asterinas developers.
# For more information, check out the README of sctrace.
RUN echo 'export ASTER_SCML=$(find /root/asterinas/book/src/kernel/linux-compatibility/ -name "*.scml")' >> /root/.bashrc
```

PR #2731, opened 2025-12-11
Disable systemd network services and fix the conflict between xfce and getty

cchanging:


distro/etc_nixos/modules/systemd.nix
```diff
@@ -4,11 +4,13 @@
   systemd.package = pkgs.aster_systemd;
 
   systemd.coredump.enable = false;
```
tatetian, line 9: Could you add a TODO before this huge paragraph of disabled features/services of systemd?

PR #2732, opened 2025-12-11
Test more podman commands in CI

StevenJiang1110:
Except `podman run`, this PR also add `podman image ls`, `podman ps -a` and `podman rm` commands into CI.

### DIFF
--- a/.github/workflows/test_nixos_full.yml
+++ b/.github/workflows/test_nixos_full.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Test nix commands
         run: |
-          make nixos NIXOS_DISK_SIZE_IN_MB=6144 NIXOS_TEST_COMMAND='test-nix-commands' || true
+          make nixos NIXOS_DISK_SIZE_IN_MB=6144 NIXOS_TEST_COMMAND='test-nix-commands'
           make run_nixos || true
 
       - name: Check results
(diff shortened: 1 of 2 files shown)


.github/workflows/test_nixos_full.yml
```diff
@@ -56,6 +56,13 @@ jobs:
           distro/etc_nixos/configuration.nix
       - name: Run podman commands
         run: |
-          make run NIXOS=1 NIXOS_DISK_SIZE_IN_MB=8192 NIXOS_TEST_COMMAND='podman run docker.io/library/alpine:latest ls /etc' || true
-          tail --lines 50 qemu.log | grep -q "alpine-release" || (echo "Test podman failed" && exit 1)
+          make run \
+            NIXOS=1 \
+            NIXOS_DISK_SIZE_IN_MB=8192 \
+            NIXOS_TEST_COMMAND="podman run --name=c1 docker.io/library/alpine:latest ls /etc && podman image ls && podman ps -a && podman rm c1 && echo 'rm succeeds'" \
```
tatetian: Break it into multiple lines to make the test command more readable

.github/workflows/test_nixos_full.yml
```diff
@@ -56,6 +56,13 @@ jobs:
...
+            || true
+          tail --lines 100 qemu.log | grep -q "alpine-release" || (echo "Test `podman run` failed" && exit 1)
```
tatetian:
```bash
echo "begin `cmd args` end"
```

The above command makes Bash to execute the embedded `cmd args` command.

PR #2736, opened 2025-12-11
Place the heap after bss segment

wyt8:
## Short description

This PR mainly focuses on the problem that origin heap is fixed and unaligned with Linux behavior.

The main changes are as follows:
1. Acquire the data segment and break info when loading the elf.
2. Add a new method of allocaling free region which starts from high address.
3. Reimplement the heap code.

### DIFF
--- a/kernel/src/process/process/init_proc.rs
+++ b/kernel/src/process/process/init_proc.rs
@@ -13,16 +13,16 @@ use crate::{
     },
     prelude::*,
     process::{
-        Credentials, UserNamespace,
+        Credentials, ProcessVm, UserNamespace,
         posix_thread::{PosixThreadBuilder, ThreadName, allocate_posix_tid},
         process_table,
-        process_vm::new_vmar_and_map,
         program_loader::ProgramToLoad,
         rlimit::new_resource_limits_for_init,
         signal::sig_disposition::SigDispositions,
     },
     sched::Nice,
     thread::Tid,
+    vm::vmar::Vmar,
 };
 
 /// Creates and schedules the init process to run.
@@ -56,7 +56,7 @@ fn create_init_process(
     let elf_path = fs.resolver().read().lookup(&fs_path)?;
 
     let pid = allocate_posix_tid();
-    let process_vm = new_vmar_and_map(elf_path.clone());
+    let vmar = Vmar::new(ProcessVm::new(elf_path.clone()));
     let resource_limits = new_resource_limits_for_init();
     let nice = Nice::default();
     let oom_score_adj = 0;
@@ -65,7 +65,7 @@ fn create_init_process(
 
     let init_proc = Process::new(
         pid,
-        process_vm,
+        vmar,
         resource_limits,
         nice,
         oom_score_adj,
--- a/kernel/src/process/process_vm/heap.rs
+++ b/kernel/src/process/process_vm/heap.rs
@@ -1,139 +1,157 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use core::sync::atomic::{AtomicUsize, Ordering};
+use core::ops::Range;
 
 use align_ext::AlignExt;
 
 use crate::{
     prelude::*,
+    process::ResourceType,
+    util::random::getrandom,
     vm::{perms::VmPerms, vmar::Vmar},
 };
 
-/// The base address of user heap
-pub const USER_HEAP_BASE: Vaddr = 0x0000_0000_1000_0000;
-/// The max allowed size of user heap
-pub const USER_HEAP_SIZE_LIMIT: usize = 16 * 1024 * PAGE_SIZE; // 16 * 4MB
-
 #[derive(Debug)]
 pub struct Heap {
-    /// The lowest address of the heap
-    base: Vaddr,
-    /// The heap size limit
-    limit: usize,
-    /// The current heap highest address
-    current_program_break: AtomicUsize,
+    inner: Mutex<Option<HeapInner>>,
+}
+
+#[derive(Clone, Debug)]
+struct HeapInner {
+    /// The size of the data segment, used for rlimit checking.
+    data_segment_size: usize,
+    /// The heap range.
+    // NOTE: `heap_range.end` is decided by user input and may not be page-aligned.
+    heap_range: Range<Vaddr>,
 }
 
 impl Heap {
-    pub const fn new() -> Self {
-        Heap {
-            base: USER_HEAP_BASE,
-            limit: USER_HEAP_SIZE_LIMIT,
-            current_program_break: AtomicUsize::new(USER_HEAP_BASE),
+    pub(super) const fn new_uninitialized() -> Self {
+        Self {
+            inner: Mutex::new(None),
         }
     }
 
     /// Initializes and maps the heap virtual memory.
-    pub(super) fn alloc_and_map(&self, vmar: &Vmar) -> Result<()> {
+    pub(super) fn map_and_init_heap(
+        &self,
+        vmar: &Vmar,
+        data_segment_size: usize,
+        heap_base: Vaddr,
+    ) -> Result<()> {
+        let mut inner = self.inner.lock();
+
+        let nr_pages_padding = {
+            // Some random padding pages are added as a simple measure to
+            // make the heap values of a buggy user program harder
+            // to be exploited by attackers.
+            let mut nr_random_padding_pages: u8 = 0;
+            getrandom(nr_random_padding_pages.as_bytes_mut());
+
+            nr_random_padding_pages as usize
+        };
+
+        let heap_start = heap_base.align_up(PAGE_SIZE) + nr_pages_padding * PAGE_SIZE;
+
         let vmar_map_options = {
             let perms = VmPerms::READ | VmPerms::WRITE;
-            vmar.new_map(PAGE_SIZE, perms).unwrap().offset(self.base)
+            vmar.new_map(PAGE_SIZE, perms).unwrap().offset(heap_start)
         };
         vmar_map_options.build()?;
 
-        // If we touch another mapped range when we are trying to expand the
-        // heap, we fail.
-        //
-        // So a simple solution is to reserve enough space for the heap by
-        // mapping without any permissions and allow it to be overwritten
-        // later by `brk`. New mappings from `mmap` that overlaps this range
-        // may be moved to another place.
-        let vmar_reserve_options = {
-            let perms = VmPerms::empty();
-            vmar.new_map(USER_HEAP_SIZE_LIMIT - PAGE_SIZE, perms)
-                .unwrap()
-                .offset(self.base + PAGE_SIZE)
-        };
-        vmar_reserve_options.build()?;
+        debug_assert!(inner.is_none());
+        *inner = Some(HeapInner {
+            data_segment_size,
+            heap_range: heap_start..heap_start + PAGE_SIZE,
+        });
 
-        self.set_uninitialized();
         Ok(())
     }
 
-    /// Returns the current program break.
-    pub fn program_break(&self) -> Vaddr {
-        self.current_program_break.load(Ordering::Relaxed)
+    /// Returns the current heap end.
+    pub fn heap_end(&self) -> Vaddr {
+        let inner = self.inner.lock();
+        let inner = inner.as_ref().expect("Heap is not initialized");
+        inner.heap_range.end
     }
 
-    /// Sets the program break to `new_heap_end`.
+    /// Modifies the end address of the heap.
     ///
-    /// Returns the new program break on success, or the current program break on failure.
+    /// Returns the new heap end on success, or the current heap end on failure.
     /// This behavior is consistent with the Linux `brk` syscall.
-    pub fn set_program_break(
+    //
+    // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/mm/mmap.c#L115>
+    pub fn modify_heap_end(
         &self,
-        new_program_break: Vaddr,
+        new_heap_end: Vaddr,
         ctx: &Context,
     ) -> core::result::Result<Vaddr, Vaddr> {
         let user_space = ctx.user_space();
         let vmar = user_space.vmar();
 
-        let current_program_break = self.current_program_break.load(Ordering::Acquire);
+        let mut inner = self.inner.lock();
+        let inner = inner.as_mut().expect("Heap is not initialized");
 
-        // According to the Linux source code, when the `brk` value is more than the
-        // rlimit, the `brk` syscall returns the current `brk` value.
-        // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/mm/mmap.c#L152>
-        if new_program_break > self.base + self.limit {
-            return Err(current_program_break);
-        }
-        if new_program_break < current_program_break {
-            // FIXME: should we allow shrink current user heap?
-            return Ok(current_program_break);
-        }
+        let heap_start = inner.heap_range.start;
+        let current_heap_end = inner.heap_range.end;
+        let new_heap_range = heap_start..new_heap_end;
 
-        let current_program_break_aligned = current_program_break.align_up(PAGE_SIZE);
-        let new_program_break_aligned = new_program_break.align_up(PAGE_SIZE);
-
-        // No need to expand the heap.
-        if new_program_break_aligned == current_program_break_aligned {
-            self.current_program_break
-                .store(new_program_break, Ordering::Release);
-            return Ok(new_program_break);
+        // Check if the new heap end is valid.
+        if new_heap_end < heap_start
+            || check_data_rlimit(inner.data_segment_size, &new_heap_range, ctx).is_err()
+            || new_heap_end.checked_add(PAGE_SIZE).is_none()
+        {
+            return Err(current_heap_end);
         }
 
-        // Remove the reserved space.
-        vmar.remove_mapping(current_program_break_aligned..new_program_break_aligned)
-            .map_err(|_| current_program_break)?;
+        let current_heap_end_aligned = current_heap_end.align_up(PAGE_SIZE);
+        let new_heap_end_aligned = new_heap_end.align_up(PAGE_SIZE);
 
-        let old_size = current_program_break_aligned - self.base;
-        let new_size = new_program_break_aligned - self.base;
-        // Expand the heap.
-        vmar.resize_mapping(self.base, old_size, new_size, false)
-            .map_err(|_| current_program_break)?;
+        let old_size = current_heap_end_aligned - heap_start;
+        let new_size = new_heap_end_aligned - heap_start;
 
-        self.current_program_break
-            .store(new_program_break, Ordering::Release);
-        Ok(new_program_break)
-    }
+        // No change in the heap mapping.
+        if old_size == new_size {
+            inner.heap_range = new_heap_range;
+            return Ok(new_heap_end);
+        }
 
-    pub(super) fn set_uninitialized(&self) {
-        self.current_program_break
-            .store(self.base + PAGE_SIZE, Ordering::Relaxed);
+        // Because the mapped heap region may contain multiple mappings, which can be
+        // done by `mmap` syscall or other ways, we need to be careful when modifying
+        // the heap mapping.
+        // For simplicity, we set `check_single_mapping` to `true` to ensure that the
+        // heap region contains only a single mapping.
+        vmar.resize_mapping(heap_start, old_size, new_size, true)
+            .map_err(|_| current_heap_end)?;
+
+        inner.heap_range = new_heap_range;
+        Ok(new_heap_end)
     }
 }
 
 impl Clone for Heap {
     fn clone(&self) -> Self {
-        let current_program_break = self.current_program_break.load(Ordering::Relaxed);
         Self {
-            base: self.base,
-            limit: self.limit,
-            current_program_break: AtomicUsize::new(current_program_break),
+            inner: Mutex::new(self.inner.lock().clone()),
         }
     }
 }
 
-impl Default for Heap {
-    fn default() -> Self {
-        Self::new()
+/// Checks whether the new heap range exceeds the data segment size limit.
+// Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/include/linux/mm.h#L3287>
+fn check_data_rlimit(
+    data_segment_size: usize,
+    new_heap_range: &Range<Vaddr>,
+    ctx: &Context,
+) -> Result<()> {
+    let rlimit_data = ctx
+        .process
+        .resource_limits()
+        .get_rlimit(ResourceType::RLIMIT_DATA)
+        .get_cur();
+
+    if rlimit_data.saturating_sub(data_segment_size as u64) < new_heap_range.len() as u64 {
+        return_errno_with_message!(Errno::ENOSPC, "the data segment size limit is reached");
     }
+    Ok(())
 }
--- a/kernel/src/process/program_loader/elf/elf_file.rs
+++ b/kernel/src/process/program_loader/elf/elf_file.rs
@@ -168,6 +168,14 @@ impl ElfHeaders {
             .reduce(|r1, r2| r1.start.min(r2.start)..r1.end.max(r2.end))
             .unwrap()
     }
+
+    /// Finds the last loadable segment and returns its virtual address bounds.
+    pub(super) fn find_last_vaddr_bound(&self) -> Option<Range<Vaddr>> {
+        self.loadable_phdrs
+            .iter()
+            .max_by_key(|phdr| phdr.virt_range().end)
+            .map(|phdr| phdr.virt_range().clone())
+    }
 }
 
 struct ElfHeader {
--- a/kernel/src/process/program_loader/elf/load_elf.rs
+++ b/kernel/src/process/program_loader/elf/load_elf.rs
@@ -64,7 +64,7 @@ pub fn load_elf_to_vmar(
         not(any(target_arch = "x86_64", target_arch = "riscv64")),
         expect(unused_mut)
     )]
-    let (_range, entry_point, mut aux_vec) =
+    let (elf_mapped_info, entry_point, mut aux_vec) =
         map_vmos_and_build_aux_vec(vmar, ldso, &elf_headers, elf_inode)?;
 
     // Map the vDSO and set the entry.
@@ -81,6 +81,11 @@ pub fn load_elf_to_vmar(
 
     vmar.process_vm()
         .map_and_write_init_stack(vmar, argv, envp, aux_vec)?;
+    vmar.process_vm().map_and_init_heap(
+        vmar,
+        elf_mapped_info.data_segment_size,
+        elf_mapped_info.heap_base,
+    )?;
 
     let user_stack_top = vmar.process_vm().init_stack().user_stack_top();
     Ok(ElfLoadInfo {
@@ -132,26 +137,26 @@ fn lookup_and_parse_ldso(
 
 /// Maps the VMOs to the corresponding virtual memory addresses and builds the auxiliary vector.
 ///
-/// Returns the mapped range, the entry point, and the auxiliary vector.
+/// Returns the mapped information, the entry point, and the auxiliary vector.
 fn map_vmos_and_build_aux_vec(
     vmar: &Vmar,
     ldso: Option<(Path, ElfHeaders)>,
     parsed_elf: &ElfHeaders,
     elf_inode: &Arc<dyn Inode>,
-) -> Result<(RelocatedRange, Vaddr, AuxVec)> {
+) -> Result<(ElfMappedInfo, Vaddr, AuxVec)> {
     let ldso_load_info = if let Some((ldso_file, ldso_elf)) = ldso {
         Some(load_ldso(vmar, &ldso_file, &ldso_elf)?)
     } else {
         None
     };
 
-    let elf_map_range = map_segment_vmos(parsed_elf, vmar, elf_inode, ldso_load_info.is_some())?;
+    let elf_mapped_info = map_segment_vmos(parsed_elf, vmar, elf_inode, ldso_load_info.is_some())?;
 
     let mut aux_vec = {
         let ldso_base = ldso_load_info
             .as_ref()
             .map(|load_info| load_info.range.relocated_start());
-        init_aux_vec(parsed_elf, &elf_map_range, ldso_base)?
+        init_aux_vec(parsed_elf, &elf_mapped_info.full_range, ldso_base)?
     };
 
     // Set AT_SECURE based on setuid/setgid bits of the executable file.
@@ -166,7 +171,8 @@ fn map_vmos_and_build_aux_vec(
     let entry_point = if let Some(ldso_load_info) = ldso_load_info {
         ldso_load_info.entry_point
     } else {
-        elf_map_range
+        elf_mapped_info
+            .full_range
             .relocated_addr_of(parsed_elf.entry_point())
             .ok_or_else(|| {
                 Error::with_message(
@@ -176,7 +182,7 @@ fn map_vmos_and_build_aux_vec(
             })?
     };
 
-    Ok((elf_map_range, entry_point, aux_vec))
+    Ok((elf_mapped_info, entry_point, aux_vec))
 }
 
 struct LdsoLoadInfo {
@@ -189,7 +195,8 @@ struct LdsoLoadInfo {
 }
 
 fn load_ldso(vmar: &Vmar, ldso_file: &Path, ldso_elf: &ElfHeaders) -> Result<LdsoLoadInfo> {
-    let range = map_segment_vmos(ldso_elf, vmar, ldso_file.inode(), false)?;
+    let elf_mapped_info = map_segment_vmos(ldso_elf, vmar, ldso_file.inode(), false)?;
+    let range = elf_mapped_info.full_range;
     let entry_point = range
         .relocated_addr_of(ldso_elf.entry_point())
         .ok_or_else(|| {
@@ -201,11 +208,22 @@ fn load_ldso(vmar: &Vmar, ldso_file: &Path, ldso_elf: &ElfHeaders) -> Result<Lds
     Ok(LdsoLoadInfo { entry_point, range })
 }
 
+/// The information of mapped ELF segments.
+struct ElfMappedInfo {
+    /// The range covering all the mapped segments.
+    full_range: RelocatedRange,
+    /// The size of the data segment.
+    data_segment_size: usize,
+    /// The base address for the heap start.
+    heap_base: Vaddr,
+}
+
 /// Initializes a [`Vmo`] for each segment and then map to the [`Vmar`].
 ///
-/// This function will return the mapped range that covers all segments. The
-/// range will be tight, i.e., will not include any padding bytes. So the
-/// boundaries may not be page-aligned.
+/// This function will return the mapped information, which contains the
+/// mapped range that covers all segments. The range will be tight,
+/// i.e., will not include any padding bytes. So the boundaries may not
+/// be page-aligned.
 ///
 /// [`Vmo`]: crate::vm::vmo::Vmo
 fn map_segment_vmos(
@@ -213,8 +231,11 @@ fn map_segment_vmos(
     vmar: &Vmar,
     elf_inode: &Arc<dyn Inode>,
     has_interpreter: bool,
-) -> Result<RelocatedRange> {
+) -> Result<ElfMappedInfo> {
     let elf_va_range = elf.calc_total_vaddr_bounds();
+
+    // The base address for the heap start. If it's `None`, we will use the end of ELF segments.
+    let mut heap_base = None;
 
     let map_range = if elf.is_shared_object() {
         // Relocatable object.
@@ -261,6 +282,16 @@ fn map_segment_vmos(
                 .offset(offset)
         } else {
             // Static PIE program: pick an aligned address from the mmap region.
+
+            // When executing static PIE programs, place the heap area away from the
+            // general mmap region and into the unused `PIE_BASE_ADDR` space.
+            // This helps avoid early collisions, since the heap grows upward while
+            // the stack grows downward, and other mappings (such as the vDSO) may
+            // also be placed in the mmap region.
+            //
+            // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/fs/binfmt_elf.c#L1293>
+            heap_base = Some(PIE_BASE_ADDR);
+
             vmar.new_map(map_size, VmPerms::empty())?.align(align)
         };
         let aligned_range = vmar_map_options.build().map(|addr| addr..addr + map_size)?;
@@ -312,7 +343,16 @@ fn map_segment_vmos(
         map_segment_vmo(loadable_phdr, &elf_vmo, vmar, map_at)?;
     }
 
-    Ok(relocated_range)
+    // Calculate the data segment size.
+    // According to Linux behavior, the data segment only includes the last loadable segment.
+    // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/fs/binfmt_elf.c#L1200-L1227>
+    let data_segment_size = elf.find_last_vaddr_bound().map_or(0, |range| range.len());
+
+    Ok(ElfMappedInfo {
+        full_range: relocated_range,
+        data_segment_size,
+        heap_base: heap_base.unwrap_or(map_range.end),
+    })
 }
 
 /// Creates and maps the segment VMO to the VMAR.
--- a/kernel/src/syscall/brk.rs
+++ b/kernel/src/syscall/brk.rs
@@ -10,15 +10,16 @@ pub fn sys_brk(heap_end: u64, ctx: &Context) -> Result<SyscallReturn> {
         Some(heap_end as usize)
     };
     debug!("new heap end = {:x?}", heap_end);
+
     let user_space = ctx.user_space();
     let user_heap = user_space.vmar().process_vm().heap();
 
-    let syscall_ret = match new_heap_end {
+    let current_heap_end = match new_heap_end {
         Some(addr) => user_heap
-            .set_program_break(addr, ctx)
-            .unwrap_or_else(|current_break| current_break),
-        None => user_heap.program_break(),
+            .modify_heap_end(addr, ctx)
+            .unwrap_or_else(|cur_heap_end| cur_heap_end),
+        None => user_heap.heap_end(),
     };
 
-    Ok(SyscallReturn::Return(syscall_ret as _))
+    Ok(SyscallReturn::Return(current_heap_end as _))
 }
--- a/kernel/src/process/execve.rs
+++ b/kernel/src/process/execve.rs
@@ -15,7 +15,7 @@ use crate::{
     process::{
         ContextUnshareAdminApi, Credentials, Process,
         posix_thread::{PosixThread, ThreadLocal, ThreadName, sigkill_other_threads, thread_table},
-        process_vm::{MAX_LEN_STRING_ARG, MAX_NR_STRING_ARGS, new_vmar_and_map},
+        process_vm::{MAX_LEN_STRING_ARG, MAX_NR_STRING_ARGS, ProcessVm},
         program_loader::{ProgramToLoad, elf::ElfLoadInfo},
         signal::{
             HandlePendingSignal, PauseReason, SigStack,
@@ -54,7 +54,7 @@ pub fn do_execve(
     let program_to_load =
         ProgramToLoad::build_from_inode(elf_inode.clone(), &fs_resolver, argv, envp)?;
 
-    let new_vmar = new_vmar_and_map(elf_file.clone());
+    let new_vmar = Vmar::new(ProcessVm::new(elf_file.clone()));
     let elf_load_info = program_to_load.load_to_vmar(new_vmar.as_ref(), &fs_resolver)?;
 
     // Ensure no other thread is concurrently performing exit_group or execve.
--- a/kernel/src/process/mod.rs
+++ b/kernel/src/process/mod.rs
@@ -38,7 +38,7 @@ pub use process::{
     Terminal, broadcast_signal_async, enqueue_signal_async, spawn_init_process,
 };
 pub use process_filter::ProcessFilter;
-pub use process_vm::ProcessVm;
+pub use process_vm::{INIT_STACK_SIZE, ProcessVm};
 pub use rlimit::ResourceType;
 pub use stats::collect_process_creation_count;
 pub use term_status::TermStatus;
--- a/kernel/src/process/process_vm/mod.rs
+++ b/kernel/src/process/process_vm/mod.rs
@@ -18,7 +18,7 @@ use core::sync::atomic::{AtomicUsize, Ordering};
 use ostd::{sync::MutexGuard, task::disable_preempt};
 
 pub use self::{
-    heap::{Heap, USER_HEAP_SIZE_LIMIT},
+    heap::Heap,
     init_stack::{
         INIT_STACK_SIZE, InitStack, InitStackReader, MAX_LEN_STRING_ARG, MAX_NR_STRING_ARGS,
         aux_vec::{AuxKey, AuxVec},
@@ -28,8 +28,6 @@ use crate::{fs::path::Path, prelude::*, vm::vmar::Vmar};
 
 /*
  * The user's virtual memory space layout looks like below.
- * TODO: The layout of the userheap does not match the current implementation,
- * And currently the initial program break is a fixed value.
  *
  *  (high address)
  *  +---------------------+ <------+ The top of Vmar, which is the highest address usable
@@ -77,10 +75,10 @@ pub struct ProcessVm {
 
 impl ProcessVm {
     /// Creates a new `ProcessVm` without mapping anything.
-    fn new(executable_file: Path) -> Self {
+    pub(super) fn new(executable_file: Path) -> Self {
         Self {
             init_stack: InitStack::new(),
-            heap: Heap::new(),
+            heap: Heap::new_uninitialized(),
             executable_file,
             #[cfg(target_arch = "riscv64")]
             vdso_base: AtomicUsize::new(0),
@@ -122,6 +120,17 @@ impl ProcessVm {
         aux_vec: AuxVec,
     ) -> Result<()> {
         self.init_stack().map_and_write(vmar, argv, envp, aux_vec)
+    }
+
+    /// Maps and initializes the heap virtual memory.
+    pub(super) fn map_and_init_heap(
+        &self,
+        vmar: &Vmar,
+        data_segment_size: usize,
+        heap_base: Vaddr,
+    ) -> Result<()> {
+        self.heap()
+            .map_and_init_heap(vmar, data_segment_size, heap_base)
     }
 
     /// Returns the base address for vDSO segment.
@@ -198,19 +207,6 @@ impl<'a> ProcessVmarGuard<'a> {
         self.as_ref()
             .map(|vmar| vmar.process_vm().init_stack.reader(vmar))
     }
-}
-
-/// Creates a new VMAR and map the heap.
-///
-/// This method should only be used to create a VMAR for the init process.
-pub(super) fn new_vmar_and_map(executable_file: Path) -> Arc<Vmar> {
-    let new_vmar = Vmar::new(ProcessVm::new(executable_file));
-    new_vmar
-        .process_vm()
-        .heap()
-        .alloc_and_map(new_vmar.as_ref())
-        .unwrap();
-    new_vmar
 }
 
 /// Activates the [`Vmar`] in the current process's context.
--- a/kernel/src/process/rlimit.rs
+++ b/kernel/src/process/rlimit.rs
@@ -6,7 +6,7 @@ use core::{
     sync::atomic::{AtomicU64, Ordering},
 };
 
-use super::process_vm::{INIT_STACK_SIZE, USER_HEAP_SIZE_LIMIT};
+use super::process_vm::INIT_STACK_SIZE;
 use crate::{
     prelude::*,
     process::{UserNamespace, credentials::capabilities::CapSet},
@@ -48,8 +48,7 @@ impl Default for ResourceLimits {
         // Sets the resource limits with predefined values
         rlimits[ResourceType::RLIMIT_CPU as usize] = RLimit64::new(RLIM_INFINITY, RLIM_INFINITY);
         rlimits[ResourceType::RLIMIT_FSIZE as usize] = RLimit64::new(RLIM_INFINITY, RLIM_INFINITY);
-        rlimits[ResourceType::RLIMIT_DATA as usize] =
-            RLimit64::new(USER_HEAP_SIZE_LIMIT as u64, RLIM_INFINITY);
+        rlimits[ResourceType::RLIMIT_DATA as usize] = RLimit64::new(RLIM_INFINITY, RLIM_INFINITY);
         rlimits[ResourceType::RLIMIT_STACK as usize] =
             RLimit64::new(INIT_STACK_SIZE as u64, RLIM_INFINITY);
         rlimits[ResourceType::RLIMIT_CORE as usize] = RLimit64::new(0, RLIM_INFINITY);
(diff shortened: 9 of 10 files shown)


kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
-
 #[derive(Debug)]
 pub struct Heap {
-    /// The lowest address of the heap
-    base: Vaddr,
-    /// The heap size limit
-    limit: usize,
-    /// The current heap highest address
-    current_program_break: AtomicUsize,
+    inner: RwMutex<Option<HeapInner>>,
```
lrh2000: I don't think using `RwMutex` is necessary here. Generally, only the main thread uses `brk`, so there is no contention for this lock. It is unlikely that `heap_end()`/`modify_heap_end` will be called in a way that makes an `RwMutex` behave better than a `Mutex`. So I'll prefer to use a simpler lock, i.e., `Mutex`.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
+}
+
+#[derive(Clone, Debug)]
+struct HeapInner {
+    /// The data segment range of the process.
+    data_range: Range<Vaddr>,
```
lrh2000: I don't think it's necessary to use a range here? We only need to know the size of the data segment.
```rust
    /// The size of all loaded data segments.
    data_size: usize,
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
+    // NOTE: heap_range.end is decided by user input and may not be page aligned.
+    heap_range: Range<Vaddr>,
```
lrh2000, line 25: - `heap_range.end` -> `` `heap_range.end` ``
 - `page aligned` -> `page aligned`
 - Add doc comments for consistency with other fields.
```rust
    /// The heap range.
    // NOTE: `heap_range.end` is decided by user input and may not be page-aligned.
    heap_range: Range<Vaddr>,
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
-        self.set_uninitialized();
         Ok(())
     }
 
-    /// Returns the current program break.
-    pub fn program_break(&self) -> Vaddr {
-        self.current_program_break.load(Ordering::Relaxed)
+    /// Returns the current heap end.
+    pub fn heap_end(&self) -> Vaddr {
+        assert!(self.is_initialized());
```
lrh2000: Remove all instances of `assert!(self.is_initialized());` as well as the `is_initialized()` method.

If it is not initialized, we will panic due to `inner.as_ref().unwrap()` below. There is no need for this assertion here.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
+            // Some random padding pages are added as a simple measure to
+            // make the heap values of a buggy user program harder
+            // to be exploited by attackers.
+            let mut nr_random_padding_pages: u8 = 0;
+            getrandom(nr_random_padding_pages.as_bytes_mut());
+
+            nr_random_padding_pages as usize
+        };
+
+        let heap_start = elf_brk.align_up(PAGE_SIZE) + nr_pages_padding * PAGE_SIZE;
```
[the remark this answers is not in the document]
lrh2000: It is really hard to review changes if the existing code is too buggy...

I don't think there's any need to check for overflows here. `elf_brk` should be a guaranteed userspace address, so adding some padding pages after it cannot overflow.

The problem is that the existing code does not verify the validity of the address or size in the program header. Instead, we should fix the existing problems. Adding overflow checks here would still be unnecessary.

Additionally, `VmarMapOptions::offset` should perform more checks to ensure that `heap_start` is still a userspace address. (The check exists in `sys_mmap` but it should really live in `VmarMapOptions`.) This is just another existing problem outside the scope of this PR.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
-        // later by `brk`. New mappings from `mmap` that overlaps this range
-        // may be moved to another place.
-        let vmar_reserve_options = {
-            let perms = VmPerms::empty();
-            vmar.new_map(USER_HEAP_SIZE_LIMIT - PAGE_SIZE, perms)
-                .unwrap()
-                .offset(self.base + PAGE_SIZE)
-        };
-        vmar_reserve_options.build()?;
+        *inner = Some(HeapInner {
```
lrh2000, line 63:
```rust
debug_assert!(inner.is_none());
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
     }
 
-    /// Sets the program break to `new_heap_end`.
+    /// Modifies the end address of the heap.
     ///
-    /// Returns the new program break on success, or the current program break on failure.
+    /// Returns the new heap end on success, or the current heap end on failure.
     /// This behavior is consistent with the Linux `brk` syscall.
-    pub fn set_program_break(
+    // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/mm/mmap.c#L115>
```
lrh2000, line 83: Insert a blank line above, like this:
```rust
    /// doc comments (first paragraph)
    ///
    /// doc comments (second paragraph)
    //
    // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/mm/mmap.c#L115>
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
-        let new_size = new_program_break_aligned - self.base;
-        // Expand the heap.
-        vmar.resize_mapping(self.base, old_size, new_size, false)
-            .map_err(|_| current_program_break)?;
+        // Because the mapped heap region may contain multiple mappings, which can be
+        // done by `mmap` syscall or other ways, we need to be careful when modifying
+        // the heap mapping.
+        // There we consider the two cases: expanding and shrinking the heap.
+        if old_size < new_size {
+            // Expanding the heap.
```
lrh2000:
```rust
    if old_size < new_size {
        // Expand the heap.
        ...
    } else {
        // Shrink the heap.
        ...
    }
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,166 @@
...
+    data_range: &Range<Vaddr>,
+    new_heap_range: &Range<Vaddr>,
+    ctx: &Context,
+) -> Result<()> {
+    let rlimit_data = ctx
+        .process
+        .resource_limits()
+        .get_rlimit(ResourceType::RLIMIT_DATA);
+    if data_range.len() as u64 + new_heap_range.len() as u64 > rlimit_data.get_cur() {
+        return_errno_with_message!(Errno::ENOSPC, "data segment size limit was met.");
```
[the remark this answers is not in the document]
lrh2000:
```suggestion
        return_errno_with_message!(Errno::ENOMEM, "the data segment size limit is reached");
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -222,31 +225,79 @@ pub fn map_segment_vmos(
...
+            // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/fs/binfmt_elf.c#L1200>
+            if program_header.flags.is_execute() {
+                // Update code segment range.
+                if program_header.virtual_addr < start_code {
+                    start_code = program_header.virtual_addr;
+                }
+                if end_code < program_header.virtual_addr + program_header.file_size {
+                    end_code = program_header.virtual_addr + program_header.file_size;
+                }
+            }
```
lrh2000: No need to compute the code range because no one uses it?

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -222,31 +225,79 @@ pub fn map_segment_vmos(
...
+            // Update data segment range.
+            if start_data < program_header.virtual_addr {
+                start_data = program_header.virtual_addr;
+            }
+            if end_data < program_header.virtual_addr + program_header.file_size {
+                end_data = program_header.virtual_addr + program_header.file_size;
+            }
```
lrh2000: The computed data range will be the same as `elf_range`? No need to do duplication computation?

I think you implemented this because Linux does something similar. However, our goal is not to make the code implementation the same as Linux. Instead, our code must be self-explanatory.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -222,31 +225,79 @@ pub fn map_segment_vmos(
...
+
+            let map_at = if elf_va_range.contains(&(program_header.virtual_addr as Vaddr)) {
+                map_range.start + (program_header.virtual_addr as Vaddr - elf_va_range.start)
+            } else {
+                panic!("Address not covered by `get_range_for_all_segments`");
```
lrh2000: This should be a debug assertion.
```rust
debug_assert!(elf_va_range.contains(&(program_header.virtual_addr as Vaddr)));
let map_at = map_range.start + (program_header.virtual_addr as Vaddr - elf_va_range.start);
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -222,31 +225,79 @@ pub fn map_segment_vmos(
...
     ///
     /// If the relocated address overflows, it will return `None`.
-    pub fn new(original_range: Range<Vaddr>, relocated_start: Vaddr) -> Option<Self> {
+    pub fn new(
+        original_range: Range<Vaddr>,
+        original_code_range: Range<Vaddr>,
+        original_data_range: Range<Vaddr>,
+        relocated_start: Vaddr,
+    ) -> Option<Self> {
         relocated_start.checked_add(original_range.len())?;
```
lrh2000: This overflow check makes no sense. It happens after:
https://github.com/asterinas/asterinas/blob/95e971870097f903dbdb22c2146ed7d28a36cc7c/kernel/src/process/program_loader/elf/load_elf.rs#L190-L213

If it really overflows, it must already be causing problems (e.g., `map_range` will overflow).

As I said before, the existing code has a lot of bugs related to the overflow checks. I think we can remove this check and add a FIXME to note that this needs to be considered in a future PR:
```patch
 pub fn map_segment_vmos(
     elf: &ElfHeaders,
     vmar: &Vmar,
     elf_inode: &Arc<dyn Inode>,
 ) -> Result<RelocatedRange> {
+    // FIXME: Add checks to ensure that all the address ranges in the program headers
+    // are valid, i.e., belong to userspace.
     let elf_va_range = get_range_for_all_segments(elf)?;
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -222,31 +225,79 @@ pub fn map_segment_vmos(
...
+        if !original_range.contains(&original_code_range.start)
+            || !original_range.contains(&(original_code_range.end - 1))
+        {
+            return None;
+        }
+        if !original_range.contains(&original_data_range.start)
+            || !original_range.contains(&(original_data_range.end - 1))
+        {
+            return None;
+        }
```
lrh2000: These checks also don't make sense. If you think they're prone to errors, there should be debug assertions. As I said before, though, we should remove the computation of `code_range` and `data_range` because I don't think they make a difference.

kernel/src/vm/vmar/mod.rs
```diff
@@ -872,6 +872,67 @@ impl VmarInner {
...
+        // This value represents the highest possible address for a new mapping.
+        // For simplicity, we use a fixed value `512` here. The value contains the following considerations:
+        // - The stack fixed padding size.
+        // - The stack random padding size.
+        // - The future growth of the stack.
+        // FIXME: This value should consider the process's actual stack configuration, which may
+        // exist in ResourceLimits.
+        let high_limit = VMAR_CAP_ADDR - INIT_STACK_SIZE - PAGE_SIZE * 512;
+
+        let low_limit = VMAR_LOWEST_ADDR;
```
lrh2000: `low_limit` has nothing to do with `try_alloc_in_hole_topdown`. Move it near to `high_limit` instead.
```rust
let high_limit = ...;
let low_limit = ...;

let try_alloc_in_hole_topdown = ...;
```

kernel/src/vm/vmar/mod.rs
```diff
@@ -872,6 +872,67 @@ impl VmarInner {
...
+        let try_alloc_in_hole_topdown = |hole_start: Vaddr,
```
lrh2000: It is possible to define a method here (a `fn` item can exist in another `fn`):
```rust
fn try_alloc_in_hole_topdown(...) { ... }
```

This does not need to be a closure because it does not capture anything.

kernel/src/vm/vmar/mod.rs
```diff
@@ -872,6 +872,67 @@ impl VmarInner {
...
+                                         hole_end: Vaddr,
+                                         size: usize,
+                                         align: usize|
+         -> Option<Range<Vaddr>> {
+            if hole_start < hole_end {
+                let max_start = hole_end.checked_sub(size)?;
+                let max_start_aligned = max_start.align_down(align);
+                let max_end = max_start_aligned.checked_add(size)?;
```
lrh2000: This cannot overflow.

kernel/src/vm/vmar/mod.rs
```diff
@@ -872,6 +872,67 @@ impl VmarInner {
...
+
+                if max_start_aligned >= hole_start && max_end <= hole_end {
+                    return Some(max_start_aligned..max_end);
+                }
+            }
+            None
```
lrh2000: I believe this method can be simplified to something like:
```rust
let start = hole_end.checked_sub(size)?.align_down(align);
if start >= hole_start {
    Some(start..start + size)
} else {
    None
}
```

kernel/src/vm/vmar/mod.rs
```diff
@@ -1171,6 +1235,15 @@ impl<'a> VmarMapOptions<'a> {
         self
     }
 
+    /// Sets whether to allocate the mapping from high address to low address.
+    ///
+    /// The default value is true.
+    #[expect(dead_code)]
```
lrh2000: Then `alloc_free_region` is useless. I think you should just remove `alloc_free_region`, rename `alloc_free_region_topdown` to `alloc_free_region`, and remove all `topdown`-related variables/methods.

kernel/src/vm/vmar/mod.rs
```diff
@@ -872,6 +872,67 @@ impl VmarInner {
...
+    ///
+    /// If no such region is found, return an error.
+    fn alloc_free_region_topdown(&mut self, size: usize, align: usize) -> Result<Range<Vaddr>> {
+        // This value represents the highest possible address for a new mapping.
+        // For simplicity, we use a fixed value `512` here. The value contains the following considerations:
+        // - The stack fixed padding size.
+        // - The stack random padding size.
+        // - The future growth of the stack.
+        // FIXME: This value should consider the process's actual stack configuration, which may
+        // exist in ResourceLimits.
```
lrh2000:
```suggestion
        // exist in `ResourceLimits`.
```

kernel/src/vm/vmar/mod.rs
```diff
@@ -872,6 +872,67 @@ impl VmarInner {
...
+        let high_limit = VMAR_CAP_ADDR - INIT_STACK_SIZE - PAGE_SIZE * 512;
```
lrh2000: We can query `ProcessVm` to get the stack bottom?

lrh2000 (commented):
The most exciting thing is that this resolves the awful race conditions in `Heap`. Our previous `brk` implementation is completely buggy when called concurrently, which is problematic.

kernel/src/vm/vmar/mod.rs
wyt8: Sometimes we may allocate region before stack init, such as ldso, vdso. So here the constant value use.

kernel/src/process/process_vm/heap.rs
wyt8: We should not assert the inner is always None when initializing, and when fork or exec syscalls are called, the heap may be initialized by old process.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -222,31 +225,79 @@ pub fn map_segment_vmos(
...
     Ok(relocated_range)
 }
 
 /// A virtual range and its relocated address.
 pub struct RelocatedRange {
     original_range: Range<Vaddr>,
+    original_code_range: Range<Vaddr>,
+    original_data_range: Range<Vaddr>,
     relocated_start: Vaddr,
 }
```
junyang-zh: The `original_range` represents the entire range occupied by the loaded elf (maybe holes in it), and the newly added `code`/`data` range specifies the `code`/`data` ranges.

This is not what `struct RelocatedRange` was introduced for. The name indicates that the struct represents only one range. Now with the modification, it seems that a new name `ElfRelocatedRanges` looks more appropriate. And `original_range` should be renamed to `original_full_range`.

Another option is, since you need only to know where the bss ends, just return

```
struct ElfMappedInfo {
    full_range: RelocatedRange,
    relocated_bss_end: Vaddr,
}
```

on `map_segment_vmos`.

kernel/src/vm/vmar/mod.rs
```diff
@@ -1171,6 +1235,15 @@ impl<'a> VmarMapOptions<'a> {
         self
     }
 
+    /// Sets whether to allocate the mapping from high address to low address.
+    ///
+    /// The default value is true.
+    #[expect(dead_code)]
+    pub fn topdown(mut self, topdown: bool) -> Self {
```
junyang-zh: The function should be removed too. I don't see https://github.com/asterinas/asterinas/pull/2736/changes#r2611426295 fully resolved.

junyang-zh (commented):
It's good to see you fixing the heap placement! I also wanted it and drafted it in #1999 , but that PR has problems and goes unmaintained quite a while ago. Without this fix, some really aging applications in the PARSEC benchmarks cannot run correctly. Thanks for your contributions. I have some minor comments.

I see some of @lrh2000 's suggestions are marked as resolved but I cannot see the revised code. I will revisit later when the changes are pushed.

tatetian:
The upcoming version 0.17.0 is feature frozen. I will try release it today (or tomorrow). Please do not merge any substantial changes (including this PR).

lrh2000:
> Please do not merge any substantial changes (including this PR).

That's good to know. If this PR is not required for the 0.17.0 release, I will post another PR to refactor the ELF-related code and then prefer to merge this PR after my PR (there should not be any significant conflicts; I just need to clarify some behaviors, especially regarding the overflow checks, e.g, https://github.com/asterinas/asterinas/pull/2736#discussion_r2611393507).

kernel/src/process/process_vm/heap.rs
lrh2000: Oh yes. Thanks for your explanation!

kernel/src/process/program_loader/elf/load_elf.rs
lrh2000: > This is not what `struct RelocatedRange` was introduced for.

I was also confused by this when reviewing the code.

As I mentioned in my previous comments, we don't really need the data and code ranges, at least for now.

If we don't modify `RelocatedRange`, comments such as https://github.com/asterinas/asterinas/pull/2736#discussion_r2611404211 can be ignored. If `RelocatedRange` only contains `original_range`/`relocated_start`, then the original APIs, such as `new` and `relocated_addr_of`, make sense from a utility type perspective (`RelocatedRange` is a pure utility type that has nothing to do with ELF logic, so it is unaware of whether the overflow has been checked beforehand).

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -263,6 +287,21 @@ impl RelocatedRange {
...
+    /// The range covering all the mapped segments.
+    full_range: RelocatedRange,
+    /// The size of data segment.
+    data_segment_size: usize,
+}
+
+impl ElfMappedInfo {
+    /// Returns the relocated break address.
+    fn relocated_brk(&self) -> Vaddr {
+        self.full_range.relocated_start + self.full_range.original_range.len()
```
junyang-zh: How about `self.full_range.relocated_addr_of(self.full_range.original_range.end).unwrap()`? Looks slightly more understandable.

junyang-zh (approved):
~~LGTM!~~

Please ignore the approval. Sorry.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,161 @@
...
+            || check_data_rlimit(inner.data_segment_size, &new_heap_range, ctx).is_err()
+        {
+            return Err(current_heap_end);
         }
 
-        // Remove the reserved space.
-        vmar.remove_mapping(current_program_break_aligned..new_program_break_aligned)
-            .map_err(|_| current_program_break)?;
+        let current_heap_end_aligned = current_heap_end.align_up(PAGE_SIZE);
+        let new_heap_end_aligned = new_heap_end.align_up(PAGE_SIZE);
```
junyang-zh, line 108: These up-aligns can overflow. Maybe we should avoid using it here and fix `AlignExt` in the future.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,161 @@
...
+fn check_data_rlimit(
+    data_segment_size: usize,
+    new_heap_range: &Range<Vaddr>,
+    ctx: &Context,
+) -> Result<()> {
+    let rlimit_data = ctx
+        .process
+        .resource_limits()
+        .get_rlimit(ResourceType::RLIMIT_DATA);
+    if data_segment_size as u64 + new_heap_range.len() as u64 > rlimit_data.get_cur() {
```
junyang-zh: Although unlikely if the parameters are correct, please also consider checking the overflow here.

kernel/src/process/program_loader/elf/load_elf.rs
wyt8: Because the end of range is not included in the range, `self.full_range.relocated_addr_of(self.full_range.original_range.end).unwrap()` is not proper. We can use `self.full_range.relocated_addr_of(self.full_range.original_range.end - 1).unwrap() + 1` instead.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,169 @@
...
+    heap_range: Range<Vaddr>,
 }
 
 impl Heap {
-    pub const fn new() -> Self {
-        Heap {
-            base: USER_HEAP_BASE,
-            limit: USER_HEAP_SIZE_LIMIT,
-            current_program_break: AtomicUsize::new(USER_HEAP_BASE),
+    pub const fn new_uninitialized() -> Self {
```
lrh2000: `pub(super)`

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -55,6 +55,11 @@ pub fn load_elf_to_vmar(
 
     vmar.process_vm()
         .map_and_write_init_stack(vmar, argv, envp, aux_vec)?;
+    vmar.process_vm().initialize_heap(
```
lrh2000: Should we rename this to `map_and_initialize_heap`? I think it aligns better with `map_and_write_init_stack`.

Similar for `Heap::initialize`.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -263,6 +287,21 @@ impl RelocatedRange {
...
+/// The information of mapped ELF segments.
+struct ElfMappedInfo {
+    /// The range covering all the mapped segments.
+    full_range: RelocatedRange,
+    /// The size of data segment.
+    data_segment_size: usize,
+}
+
+impl ElfMappedInfo {
+    /// Returns the relocated break address.
```
lrh2000: I assume the name of `elf_brk` and all the other similar methods and documentation come from the Linux source code, which contains a variable named `elf_brk`.

But it really makes no sense for me. It should be named as something like `elf_end`.
```rust
impl RelocatedRange {
    pub fn reloacted_end(&self) -> Vaddr { todo!() }
}
```

It's not the exact program break value, either, because there are random padding pages in between.

This applies to all occurrences of the variable named `elf_brk`.

kernel/src/process/process_vm/heap.rs
lrh2000: https://github.com/asterinas/asterinas/pull/2736#discussion_r2612642668

> We should not assert the inner is always None when initializing, and when fork or exec syscalls are called, the heap may be initialized by old process.

Actually, I realized that it may not be true because...
1. In `execve()`, we'll create a new `Vmar`/`ProcessVm` from scratch.
2. In `fork()`, we should not call `map_and_xxx` methods of `Heap`/`InitStack`.

Am I still missing anything?

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,169 @@
...
         Ok(())
     }
 
-    /// Returns the current program break.
-    pub fn program_break(&self) -> Vaddr {
-        self.current_program_break.load(Ordering::Relaxed)
+    /// Returns the current heap end.
+    pub fn heap_end(&self) -> Vaddr {
+        let inner = self.inner.lock();
+        let inner = inner.as_ref().expect("Heap is not initialized.");
```
lrh2000:
```suggestion
        let inner = inner.as_ref().expect("Heap is not initialized");
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,169 @@
...
+        // the heap mapping.
+        // There we consider the two cases: expanding and shrinking the heap.
+        if old_size < new_size {
+            // Expanding the heap.
+            vmar.resize_mapping(heap_start, old_size, new_size, false)
+                .map_err(|_| current_heap_end)?;
+        } else {
+            // Shrinking the heap.
+            // FIXME: Here we don't consider the case that the mapped heap region contains
+            // multiple mappings. We set the `check_single_mapping` to `true` for simplicity.
```
lrh2000: So why `check_single_mapping` can be `false` when expanding the heap? I assume that `check_single_mapping = true` is always a safer option and it is also quite reasonable.

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,169 @@
...
+    let rlimit_data = ctx
+        .process
+        .resource_limits()
+        .get_rlimit(ResourceType::RLIMIT_DATA);
+
+    if rlimit_data.get_cur() == u64::MAX {
+        return Ok(());
     }
+    if let Some(sum) = (data_segment_size as u64).checked_add(new_heap_range.len() as u64)
+        && sum <= rlimit_data.get_cur()
```
lrh2000: Calling `get_cur()` twice may see different values due to race conditions.

Besides, this can be simplified to something like:
```rust
    let rlimit_data = ctx
        .process
        .resource_limits()
        .get_rlimit(ResourceType::RLIMIT_DATA)
        .get_cur();

    if rlimit_data.saturating_sub(data_segment_size) < new_heap_range.len() {
        return_errno_with_message!(Errno::ENOMEM, "the data segment size limit is reached");
    }
    Ok(())
```

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,169 @@
...
+    {
+        return Ok(());
+    }
+
+    return_errno_with_message!(Errno::ENOMEM, "the data segment size limit is reached.");
```
lrh2000:
```suggestion
    return_errno_with_message!(Errno::ENOMEM, "the data segment size limit is reached");
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -215,13 +223,26 @@ pub fn map_segment_vmos(
...
+
     for program_header in &elf.program_headers {
         let type_ = program_header.get_type().map_err(|_| {
             Error::with_message(Errno::ENOEXEC, "Failed to parse the program header")
         })?;
         if type_ == program::Type::Load {
             check_segment_align(program_header)?;
 
+            // Update data segment range, not including .bss part.
+            if program_header.flags.is_write() {
```
lrh2000: https://github.com/torvalds/linux/blob/9094662f6707d1d4b53d18baba459604e8bb0783/fs/binfmt_elf.c#L1238-L1239

It does not seem to check `flags.is_writable()`?

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -215,13 +223,26 @@ pub fn map_segment_vmos(
...
+    let mut end_data = 0;
+
     for program_header in &elf.program_headers {
         let type_ = program_header.get_type().map_err(|_| {
             Error::with_message(Errno::ENOEXEC, "Failed to parse the program header")
         })?;
         if type_ == program::Type::Load {
             check_segment_align(program_header)?;
 
+            // Update data segment range, not including .bss part.
```
lrh2000:
```rust
            // Update the data segment range. According to Linux behavior,
            // it will only cover `file_size` instead of `mem_size`.
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -263,6 +287,21 @@ impl RelocatedRange {
     }
 }
 
+/// The information of mapped ELF segments.
+struct ElfMappedInfo {
+    /// The range covering all the mapped segments.
+    full_range: RelocatedRange,
+    /// The size of data segment.
```
lrh2000:
```suggestion
    /// The size of data segments.
```

kernel/src/vm/vmar/mod.rs
```diff
@@ -829,47 +829,62 @@ impl VmarInner {
...
+        // For simplicity, we use a fixed value `512` here. The value contains the following considerations:
+        // - The stack fixed padding size.
+        // - The stack random padding size.
+        // - The future growth of the stack.
+        // FIXME: This value should consider the process's actual stack configuration, which may
+        // exist in `ResourceLimits`.
+        let high_limit = VMAR_CAP_ADDR - INIT_STACK_SIZE - PAGE_SIZE * 512;
+        let low_limit = VMAR_LOWEST_ADDR;
+
+        fn try_alloc_in_hole_topdown(
```
lrh2000:
```suggestion
        fn try_alloc_in_hole(
```

kernel/src/vm/vmar/mod.rs
```diff
@@ -829,47 +829,62 @@ impl VmarInner {
...
+            let hole_end = prev_vm_mapping_start.min(high_limit);
+            if let Some(region) = try_alloc_in_hole_topdown(hole_start, hole_end, size, align) {
+                return Ok(region);
+            }
         }
 
-        return_errno_with_message!(Errno::ENOMEM, "Cannot find free region for mapping");
+        return_errno_with_message!(
+            Errno::ENOMEM,
+            "Cannot find free region for mapping (topdown)"
```
lrh2000:
```suggestion
            "no free region for mapping can be found"
```

kernel/src/syscall/brk.rs
```diff
@@ -15,9 +15,9 @@ pub fn sys_brk(heap_end: u64, ctx: &Context) -> Result<SyscallReturn> {
 
     let syscall_ret = match new_heap_end {
         Some(addr) => user_heap
-            .set_program_break(addr, ctx)
+            .modify_heap_end(addr, ctx)
             .unwrap_or_else(|current_break| current_break),
```
lrh2000: - Insert a blank line before line 13.
 - Rename `current_break` to `current_heap_end`.
 - Rename `syscall_ret` to `heap_end` (or `current_heap_end`?).

lrh2000:
Hi @wyt8, now that #2777 has been merged and I think we can continue with this PR, consider rebasing? Thank you!

wyt8:
> Hi @wyt8, now that #2777 has been merged and I think we can continue with this PR, consider rebasing? Thank you!

OK, I will refactor this code soon. Thanks for your serious review. I really learn a lot from you.

kernel/src/process/process/init_proc.rs
```diff
@@ -13,11 +13,10 @@ use crate::{
     },
     prelude::*,
     process::{
-        Credentials, UserNamespace,
+        Credentials, ProgramToLoad, UserNamespace,
         posix_thread::{PosixThreadBuilder, ThreadName, allocate_posix_tid},
         process_table,
-        process_vm::new_vmar_and_map,
-        program_loader::ProgramToLoad,
+        process_vm::new_vmar_for_init,
```
lrh2000: It's not `for_init` in the latest code anymore. Remove this method and use `Vmar::new(ProcessVm::new(...))` instead.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -225,66 +217,172 @@ fn map_segment_vmos(
...
+            Some(addr - self.original_range.start + self.relocated_start)
+        } else {
+            None
+        }
     }
+}
 
-    Ok(relocated_range)
+/// The information of mapped ELF segments.
+struct ElfMappedInfo {
```
lrh2000: In the latest version, this type should be defined before `map_segment_vmos`. You should be able to understand it after checking the latest code in `load_elf.rs`.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -225,66 +217,172 @@ fn map_segment_vmos(
...
     } else {
         // Not relocatable object. Map as-is.
+        elf_va_range.clone()
+    };
+
+    let relocated_range =
+        RelocatedRange::new(elf_va_range, map_range.start).expect("Mapped range overflows");
+
+    let mut start_data = !0;
+    let mut end_data = 0;
```
lrh2000: I propose to add a new method to `ElfHeaders` as follows:
```rust
    /// Calculates the virtual address bounds of data segments as a range.
    ///
    /// Note that this method only considers the data portion. In other words, it accounts for the
    /// file size of the segment instead of the memory size.
    pub(super) fn calc_data_vaddr_bounds(&self) -> Option<Range<Vaddr>> {
```

Check `ElfHeaders::calc_total_vaddr_bounds` for an example. I think it can be clearer than this form.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -82,63 +75,67 @@ fn lookup_and_parse_ldso(
...
-/// Returns the mapped range, the entry point, and the auxiliary vector.
-fn map_vmos_and_build_aux_vec(
+/// Returns the mapped range, the entry point and the auxiliary vector.
+fn init_and_map_vmos(
     vmar: &Vmar,
     ldso: Option<(Path, ElfHeaders)>,
     parsed_elf: &ElfHeaders,
     elf_inode: &Arc<dyn Inode>,
-) -> Result<(RelocatedRange, Vaddr, AuxVec)> {
+) -> Result<(ElfMappedInfo, Vaddr, AuxVec)> {
```
lrh2000, line 146: Remember to update the doc comments of `init_and_map_vmos`.

```rust
/// Returns the mapped range, the entry point and the auxiliary vector.
```

lrh2000 (commented):
> OK, I will refactor this code soon.

Thanks! Meanwhile, I think this may help with the rebasing process and reduce a round of reviews afterwards.

kernel/src/process/process_vm/heap.rs
wyt8: You are right. I try adding the assert and find the kernel doesn't panic.

kernel/src/vm/vmar/mod.rs
wyt8: > We can query `ProcessVm` to get the stack bottom?

I have a question. How to query `ProcessVm` in `VmarInner`?

lrh2000: Emmm... Perhaps the only way is to require the caller to pass `ProcessVm` as an argument.

kernel/src/process/process_vm/heap.rs
wyt8: > So why `check_single_mapping` can be `false` when expanding the heap? I assume that `check_single_mapping = true` is always a safer option and it is also quite reasonable.

I use Chinese here: 当时我想的是在expanding时，通过其他方式映射的区域不会被收到影响，因为它是扩大的，不会导致已有映射丢失；但是在shrinking时，我们可能把通过其他方式映射的区域unmapped掉。

这里确实直接设为true更简单一些。

wyt8: To merge faster, I will use `true` directly.

kernel/src/process/program_loader/elf/load_elf.rs
wyt8: The linux source code seem only calculate the last loaded segment and use `start_data` and `end_data` to represent it. As you mentioned before, our code should be self-explanatory. So I reserved the semantic of `data_segment` here.

wyt8:
There are some bugs need me to find and fix. So more time is needed.😭

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -19,13 +21,26 @@ use crate::{
...
         vmar::{VMAR_CAP_ADDR, VMAR_LOWEST_ADDR, Vmar},
         vmo::Vmo,
     },
 };
 
+/// Base address for PIE (ET_DYN with INTERP) loading.
+///
+/// Linux calls this `ELF_ET_DYN_BASE`. It has some intentions:
+/// - The base load address for PIE programs (ET_DYN with INTERP).
+/// - The heap start address for Loader or Interpreter (ET_DYN without INTERP).
```
lrh2000: - I am not sure what you want to mean by saying "Loader or Interpreter". Are loaders and interpreters the same thing, or are they two different things? I assume they're the same, so I'll call them interpreters and not mention loaders.
 - Static PIE programs are not necessarily interpreters. For example, you can also compile a "Hello world" program using `-static-pie`.

```suggestion
/// The base address for PIE (ET_DYN with INTERP) loading.
///
/// Linux calls this `ELF_ET_DYN_BASE`. It has some intentions:
/// - The base load address for PIE programs (ET_DYN with INTERP).
/// - The heap start address for static PIE programs (ET_DYN without INTERP).
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -19,13 +21,26 @@ use crate::{
...
+///
+/// References: <https://elixir.bootlin.com/linux/v6.16.9/source/arch/x86/include/asm/elf.h#L235>
+/// - riscv:    ELF_ET_DYN_BASE = (DEFAULT_MAP_WINDOW / 3) * 2
+/// - loongarch ELF_ET_DYN_BASE = TASK_SIZE / 3 * 2
+/// - x86_64:   ELF_ET_DYN_BASE = DEFAULT_MAP_WINDOW / 3 * 2
```
lrh2000:
```suggestion
/// - x86_64:      ELF_ET_DYN_BASE = DEFAULT_MAP_WINDOW / 3 * 2
/// - riscv64:     ELF_ET_DYN_BASE = (DEFAULT_MAP_WINDOW / 3) * 2
/// - loongarch64: ELF_ET_DYN_BASE = TASK_SIZE / 3 * 2
```

 - Use the exact target name that can work in `cfg(target_arch = ...)`.
 - Follow the suggested order to maintain consistency.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -199,9 +214,13 @@ fn map_segment_vmos(
     elf: &ElfHeaders,
     vmar: &Vmar,
     elf_inode: &Arc<dyn Inode>,
+    is_interpreter: bool,
 ) -> Result<RelocatedRange> {
     let elf_va_range = elf.calc_total_vaddr_bounds();
 
+    // Keep track of the full reserved range, so we can unmap unused padding after all PT_LOAD segments are mapped.
```
lrh2000:
```suggestion
    // Keep track of the full reserved range, so we can unmap unused padding after all loadable
    // segments are mapped.
```

- Break this line. It's too long.
- Use `loadable segments` instead of `PT_LOAD segments`.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -220,12 +239,35 @@ fn map_segment_vmos(
             elf_va_range.start.align_down(align)..elf_va_range.end.align_up(align);
         let map_size = elf_va_range_aligned.len();
 
-        let vmar_map_options = vmar.new_map(map_size, VmPerms::empty())?.align(align);
+        // There are effectively two types of ET_DYN ELF binaries:
+        // - programs (i.e. PIE: ET_DYN with PT_INTERP)
+        // - loaders or interpreters (i.e. static PIE: ET_DYN without PT_INTERP, usually the ELF interpreter itself)
```
lrh2000:
```suggestion
        // - PIE programs (ET_DYN with PT_INTERP) and
        // - static PIE programs (ET_DYN without PT_INTERP, usually the ELF interpreter itself).
```

Similar to my previous comments. Please also fix the comments below:
```rust
            // Static PIE program: pick an aligned address from the mmap region.
```
```rust
            // PIE program: map near a dedicated base.
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -220,12 +239,35 @@ fn map_segment_vmos(
...
+        //
+        // Reference: <https://elixir.bootlin.com/linux/v6.19-rc2/source/fs/binfmt_elf.c#L1109>
+        let vmar_map_options = if is_interpreter {
```
lrh2000: This check does not perform as claimed in the comment. I believe that you should rename `is_interpreter` to `has_interpreter` and:
```rust
    // map_vmos_and_build_aux_vec:
    let elf_map_range = map_segment_vmos(parsed_elf, vmar, elf_inode, ldso_load_info.is_some())?;
    // load_ldso:
    let elf_map_range = map_segment_vmos(parsed_elf, vmar, elf_inode, false)?;
```

Otherwise, you can check the following example to see if it works correctly. I don't believe it works as intended:
```
> cat a.c
#include <unistd.h>
#include <stdio.h>

int main(void)
{
    printf("%p\n", sbrk(0));
}
> gcc a.c -static-pie -o a
> ./a
0x55558ea9b000
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -220,12 +239,35 @@ fn map_segment_vmos(
...
+            // Add some random padding.
+            let nr_pages_padding = {
+                let mut nr_random_padding_pages: u8 = 0;
+                getrandom(nr_random_padding_pages.as_bytes_mut());
+                nr_random_padding_pages as usize
+            };
+            let offset = (PIE_BASE_ADDR + nr_pages_padding * PAGE_SIZE).align_down(align);
+            vmar.new_map(map_size, VmPerms::empty())?
+                .align(align)
+                .offset(offset)
```
lrh2000, line 282: `VmarMapOptions::offset` requires the `[offset, offset + map_size)` locates within the VMAR. It's arguably very error prone, but you can check its docs:
https://github.com/asterinas/asterinas/blob/1e77c0a938d0312e6198bc95bc0862000b9f152b/kernel/src/vm/vmar/vmar_impls/map.rs#L152-L153

So we need to check manually:
```rust
    if VMAR_CAP_ADDR - offset <= map_size {
        return_errno_with_message!(Errno::ENOMEM, "the mapping address is too large");
    }
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -262,6 +306,21 @@ fn map_segment_vmos(
         map_segment_vmo(loadable_phdr, &elf_vmo, vmar, map_at)?;
     }
 
+    // After all segments are mapped, we may safely unmap unused padding at the
```
lrh2000: I'm still unsure why we can't just unmap the entire region.
```patch
-        reserved_range = aligned_range.clone();
+        vmar.remove_mapping(aligned_range.clone());
```

> After all segments are mapped

I'm confused by this comment because I don't think the reserved mapping needs to exist during `map_segment_vmo` either. Am I missing something?

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,157 @@
...
-        // Reference: <https://elixir.bootlin.com/linux/v6.16.9/source/mm/mmap.c#L152>
-        if new_program_break > self.base + self.limit {
-            return Err(current_program_break);
-        }
-        if new_program_break < current_program_break {
-            // FIXME: should we allow shrink current user heap?
-            return Ok(current_program_break);
-        }
+        let mut inner = self.inner.lock();
+        let inner = inner.as_mut().expect("Heap is not initialized.");
```
lrh2000:
```suggestion
        let inner = inner.as_mut().expect("Heap is not initialized");
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -203,23 +210,36 @@ fn load_ldso(vmar: &Vmar, ldso_file: &Path, ldso_elf: &ElfHeaders) -> Result<Lds
     Ok(LdsoLoadInfo { entry_point, range })
 }
 
+/// The information of mapped ELF segments.
+struct ElfMappedInfo {
+    /// The range covering all the mapped segments.
+    full_range: RelocatedRange,
+    /// The size of data segments.
+    data_segment_size: usize,
+    /// The base address for heap starting.
```
lrh2000:
```suggestion
    /// The base address for the heap start.
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -246,6 +266,15 @@ fn map_segment_vmos(
         // Reference: <https://elixir.bootlin.com/linux/v6.19-rc2/source/fs/binfmt_elf.c#L1109>
         let vmar_map_options = if is_interpreter {
             // Loader or interpreter: pick an aligned address from the mmap region.
+
+            // When executing a loader directly (i.e. static PIE: ET_DYN without PT_INTERP),
+            // move the brk area out of the mmap region and into the unused ELF_ET_DYN_BASE region.
+            // Since heap grows up it may collide early with the stack growing down or other regions
+            // being put into the mmap region by the kernel (e.g. vdso).
```
lrh2000: Try not to copy the Linux comments due to License issues.

> When executing a loader directly (i.e. static PIE: ET_DYN without PT_INTERP),

"executing static PIE programs"

Also I don't think we need to mention `ET_DYN without PT_INTERP` again since it has just been mentioned earlier.

> brk area

"heap area"

> vdso

"vDSO"

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -321,7 +350,18 @@ fn map_segment_vmos(
         vmar.remove_mapping(map_range_end_aligned..reserved_range.end)?;
     }
 
-    Ok(relocated_range)
+    // Calculate the data segment size.
+    // According to Linux behavior, the data segment only includes the last loadable segment.
+    // Reference: <https://github.com/torvalds/linux/blob/9094662f6707d1d4b53d18baba459604e8bb0783/fs/binfmt_elf.c#L1218>
```
lrh2000: Unless you have to use a GitHub link for some reason (e,g., a specific Linux commit or to be consistent with existing code), _always_ prefer a link under `elixir.bootlin.com/` because
 1. the link is shorter and 
 2. the UI is more friendly to use.

kernel/src/process/program_loader/elf/elf_file.rs
```diff
@@ -168,6 +168,14 @@ impl ElfHeaders {
             .reduce(|r1, r2| r1.start.min(r2.start)..r1.end.max(r2.end))
             .unwrap()
     }
+
+    /// Gets the virtual address bounds of the last loadable segment.
+    pub(super) fn last_loadable_vaddr_bound(&self) -> Option<Range<Vaddr>> {
```
lrh2000:
```suggestion
    /// Finds the last loadable segment and returns its virtual address bounds.
    pub(super) fn find_last_vaddr_bound(&self) -> Option<Range<Vaddr>> {
```

Prefer to have a verb (i.e., "find") here since it's not a (zero-cost) getter method.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -19,13 +19,26 @@ use crate::{
...
+/// The base address for PIE (ET_DYN with INTERP) loading.
+///
+/// Linux calls this `ELF_ET_DYN_BASE`. It has some intentions:
+/// - The base load address for PIE programs (ET_DYN with INTERP).
+/// - The heap start address for static PIE programs (ET_DYN without INTERP).
+///
+/// References: <https://elixir.bootlin.com/linux/v6.16.9/source/arch/x86/include/asm/elf.h#L235>
+/// - x86_64:       ELF_ET_DYN_BASE = DEFAULT_MAP_WINDOW / 3 * 2
+/// - riscv64:      ELF_ET_DYN_BASE = (DEFAULT_MAP_WINDOW / 3) * 2
+/// - loongarch64   ELF_ET_DYN_BASE = TASK_SIZE / 3 * 2
```
lrh2000:
```suggestion
/// - loongarch64:  ELF_ET_DYN_BASE = TASK_SIZE / 3 * 2
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -220,9 +234,41 @@ fn map_segment_vmos(
...
+                .align(align)
+                .offset(offset)
+        } else {
+            // Static PIE program: pick an aligned address from the mmap region.
+            vmar.new_map(map_size, VmPerms::empty())?.align(align)
+        };
         let aligned_range = vmar_map_options.build().map(|addr| addr..addr + map_size)?;
 
+        // After acquiring a suitable range, we can remove the mapping and then
+        // remap each segment at the desired address.
```
lrh2000:
```suggestion
        // map each segment at the desired address.
```

I don't think we remap anything....

Besides, let's do the same thing in the `else` branch? In general, I think it's the right thing to do because we don't want to have mappings for the holes in the ELF range.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -220,9 +234,41 @@ fn map_segment_vmos(
...
+                nr_random_padding_pages as usize
+            };
+            let offset = (PIE_BASE_ADDR + nr_pages_padding * PAGE_SIZE).align_down(align);
+
+            if VMAR_CAP_ADDR - offset <= map_size || offset < VMAR_LOWEST_ADDR {
+                return_errno_with_message!(
+                    Errno::ENOMEM,
+                    "failed to find a suitable address for PIE mapping"
+                );
+            }
```
lrh2000: `offset < VMAR_LOWEST_ADDR` requires a different error code:
```suggestion
        if offset < VMAR_LOWEST_ADDR {
            return_errno_with_message!(Errno::EPERM, "the mapping address is too small");
        }
        if VMAR_CAP_ADDR - offset < map_size {
            return_errno_with_message!(Errno::ENOMEM, "the mapping address is too large");
        }
```

Also, should it be `< map_size` instead of `<= map_size`? Let's use the same check as in the existing code:
https://github.com/asterinas/asterinas/blob/15d7b3532a408c575203d4f81bc392df5d2a4f9f/kernel/src/process/program_loader/elf/elf_file.rs#L289-L295

It's exactly what the docs say:
https://github.com/asterinas/asterinas/blob/4f62f2db8e355b67da71e6ac164e41fc727fc1ec/kernel/src/vm/vmar/vmar_impls/map.rs#L152-L153

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -201,21 +208,35 @@ fn load_ldso(vmar: &Vmar, ldso_file: &Path, ldso_elf: &ElfHeaders) -> Result<Lds
     Ok(LdsoLoadInfo { entry_point, range })
 }
 
+/// The information of mapped ELF segments.
+struct ElfMappedInfo {
+    /// The range covering all the mapped segments.
+    full_range: RelocatedRange,
+    /// The size of data segments.
```
lrh2000: According to the latest implementation, it appears that there can only be one data segment. This is a bit odd, but it appears to be Linux behavior. So, to avoid confusion, let's avoid plural here.
```suggestion
    /// The size of the data segment.
```

lrh2000 (commented):
Thanks for the revision! I only have several minor comments to follow up on.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -292,6 +323,10 @@ fn map_segment_vmos(
             .offset(elf_va_range_aligned.start)
             .build()?;
 
+        // After acquiring a suitable range, we can remove the mapping and then
+        // map each segment at the desired address.
+        vmar.remove_mapping(elf_va_range_aligned.clone())?;
```
lrh2000, line 328: Please keep things in the correct commit. This should be moved to the first commit. The same goes for the "remap" -> "map" typo fix.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -201,21 +208,35 @@ fn load_ldso(vmar: &Vmar, ldso_file: &Path, ldso_elf: &ElfHeaders) -> Result<Lds
...
     elf: &ElfHeaders,
     vmar: &Vmar,
     elf_inode: &Arc<dyn Inode>,
     has_interpreter: bool,
-) -> Result<RelocatedRange> {
+) -> Result<ElfMappedInfo> {
     let elf_va_range = elf.calc_total_vaddr_bounds();
 
+    // The base address for heap starting, which is initially set to the end of the ELF segments.
+    let mut heap_base = elf_va_range.end;
```
lrh2000: This is wrong. We need the relocated address here, not the original address.

Try this yourself:
```patch
diff --git a/test/src/apps/hello_c/Makefile b/test/src/apps/hello_c/Makefile
index 37e7cb4c0..09d364b9e 100644
--- a/test/src/apps/hello_c/Makefile
+++ b/test/src/apps/hello_c/Makefile
@@ -2,4 +2,4 @@
 
 include ../test_common.mk
 
-EXTRA_C_FLAGS := -static -mno-sse
+EXTRA_C_FLAGS := -pie -mno-sse
diff --git a/test/src/apps/hello_c/hello.c b/test/src/apps/hello_c/hello.c
index c3ec64e05..cff99e024 100644
--- a/test/src/apps/hello_c/hello.c
+++ b/test/src/apps/hello_c/hello.c
@@ -1,9 +1,10 @@
 
 #include <stdio.h>
+#include <unistd.h>
 
 int main()
 {
-       printf("hello world from hello_c!\n");
+       printf("%p %p\n", main, sbrk(0));
        return 0;
 }
```

In Asterinas (with this changes in this PR):
```
root@localhost:~/asterinas# make run_kernel RELEASE=1 ENABLE_BASIC_TEST=true
[ .. ]
~ # ./test/hello_c/hello 
0x5555555f5060 0xbe000
```

In Linux:
```
root@localhost:~/asterinas# ./test/build/initramfs/test/hello_c/hello 
0x5604fdcc8060 0x560513194000
```

Perhaps you want:
```suggestion
    // The base address for the heap start. If it's `None`, we will use the end of ELF segments.
    let mut heap_base = None;
```

Then use `heap_base.unwrap_or(map_range.end)` later?

kernel/src/process/process_vm/heap.rs
```diff
@@ -1,139 +1,157 @@
...
-        vmar.resize_mapping(self.base, old_size, new_size, false)
-            .map_err(|_| current_program_break)?;
+        let old_size = current_heap_end_aligned - heap_start;
+        let new_size = new_heap_end_aligned - heap_start;
 
-        self.current_program_break
-            .store(new_program_break, Ordering::Release);
-        Ok(new_program_break)
-    }
+        // No change in heap mapping.
```
lrh2000:
```suggestion
        // No change in the heap mapping.
```

kernel/src/process/program_loader/elf/load_elf.rs
wyt8: Oh, it is really a big bug. Thanks for your serious check.

wyt8: Now its output is like this:
```plain
~ # ./test/hello_c/hello
hello world from hello_c!
0x55555555c070 0x5555555c4000
~ # ./test/hello_c/hello
hello world from hello_c!
0x5555555ff070 0x5555556c1000
~ # ./test/hello_c/hello
hello world from hello_c!
0x555555572070 0x555555640000
~ # ./test/hello_c/hello
hello world from hello_c!
0x555555593070 0x55555568b000
```

PR #2738, opened 2025-12-11
Switch to a new Docker image 0.16.2-20251211

taosue:


tatetian:
@taosue Need to fix the conflict

taosue:
@tatetian Done.

PR #2739, opened 2025-12-11
Add CI for nix commands

vvvvsv:


vvvvsv:
The Nix formatter suggests using `"@aster-test-command@"` instead of `''@aster-test-command@''`, but since my test command contains `""`, I have to use the `''` form.

Moreover, this way of writing makes it hard for me to add comments for `NIXOS_TEST_COMMAND`. I think these test commands  should be placed in a separate `test_script.sh` file. Then, use an environment variable `NIXOS_TEST_SCRIPT=true or false`, if it is set to true, the script would be copied to `/root` during installation and executed after login (`/root/test_script.sh`)."

tatetian (approved):
We definitely need a better test infrastructure for AsterNixOS. This PR has quite some room for improvements. But since the schedule to release version 0.17.0 is tight, I will merge this PR as it is now. So thanks for the contribution!

Issue #2740, opened 2025-12-11
Typing poweroff in Busybox's shell to shut down the system will cause the Makefile to display an error message.


LordaeronESZ:
### Issue
When I type `poweroff` in Busybox's shell to shut down the system, it does shut down normally, but the Makefile displays an error.

<img width="878" height="131" alt="Image" src="https://github.com/user-attachments/assets/970594ed-488b-440d-8aab-fa76e3b19388" />

### Cause
The reason is that Busybox's poweroff shuts down the system by sending a `SIGUSR2` signal to the init process, which causes it to exit using the signal number (for `SIGUSR2`, it is 12) as the exit code. Then it will call poweroff to shut down the system with an error `exit_code`, as shown in the code below:

``` rust
// kernel/src/init.rs:110
...
    let raw_exit_code = init_process.status().exit_code();
    let exit_code = if raw_exit_code == 0 {
        ostd::power::ExitCode::Success
    } else {
        ostd::power::ExitCode::Failure
    };
    ostd::power::poweroff(exit_code);
...
```

PR #2741, opened 2025-12-11
Modified the handling of the init process exit code in `bsp_idle_loop`.

LordaeronESZ:
As the #2740 mentioned, we are currently unable to gracefully shut down the system using `poweroff` because it results in a Makefile error.

This PR modifies the handling of init process exit codes in `bsp_idle_loop`. For some non-zero exit codes, we now consider them normal and allow QEMU to exit with a successful code. Here is the screenshot when I test it:

### DIFF
--- a/kernel/src/init.rs
+++ b/kernel/src/init.rs
@@ -108,10 +108,27 @@ fn bsp_idle_loop() {
     // According to the Linux implementation, we should panic once the init process exits.
     // Currently, we choose to power off the machine for more flexibility in testing with QEMU.
     let raw_exit_code = init_process.status().exit_code();
+
+    // Determine if this is a normal shutdown
+    // Exit code 0 means normal exit
+    // Exit codes 1-31 are signal numbers
     let exit_code = if raw_exit_code == 0 {
         ostd::power::ExitCode::Success
     } else {
-        ostd::power::ExitCode::Failure
+        // Some signals indicate normal shutdown requests:
+        // - SIGUSR2 (12): Used by busybox poweroff
+        // - SIGTERM (15): Standard termination signal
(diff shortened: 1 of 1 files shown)


lrh2000:
This is because we run a busybox shell as the init process, but the `poweroff` command expects the init process to be [busybox `init`](https://git.busybox.net/busybox/tree/init/init.c?h=1_37_0&id=be7d1b7b1701d225379bc1665487ed0871b592a5#n831).

The kernel is not the correct place to work around a userspace bug. I don't think this PR's approach is the right way to go. To fix a bug in userspace, fix userspace instead (e.g., the scripts under `test/`).

For example, the kernel should know nothing about this:
https://github.com/asterinas/asterinas/blob/648b7f9107d36ef909ce6eaf5185da1bb6c05cf7/kernel/src/init.rs#L118-L123

Please also read this:
https://github.com/asterinas/asterinas/blob/648b7f9107d36ef909ce6eaf5185da1bb6c05cf7/kernel/src/init.rs#L108-L110

We should just panic here. The existing code to parse the init process's exit code is **already incorrect and should be removed in the future**. #2740 is not a bug severe enough to make the logic even worse.

> for more flexibility in testing with QEMU

Note that this only means we need a workaround to make our CI infrastructure function properly. Nothing else. #2740 does not apply here. #2740 is not significant enough to warrant a workaround in the kernel.

---

In addition, as quoted from https://man7.org/linux/man-pages/man2/kill.2.html,
>       The only signals that can be sent to process ID 1, the init
>       process, are those for which init has explicitly installed signal
>       handlers.  This is done to assure the system is not brought down
>       accidentally.

Therefore, the init process should never be terminated by SIGUSR2 or SIGTERM in this way.

LordaeronESZ:
Understood, thank you for your patient explanation!

PR #2743, opened 2025-12-12
Add GitHub workflow to publish ISO image and package cache

StevenJiang1110:
This PR adds a workflow to publish release for ISO image.

The created release can be viewed in my own forked repo at [here](https://github.com/StevenJiang1110/asterinas/releases/tag/0.16.2).

This workflow uses the GitHub CLI (gh command) to manage releases, and largely follows the approach used by [Ruby](https://github.com/ruby/ruby-dev-builder/blob/b0bf59a17c17985d4692243d4689c273f6348fa5/.github/workflows/build.yml). The GitHub release is managed in three steps:
1. Create a draft release in the create-release job.
2. Upload artifacts (the ISO installer in our case) in the build-iso job.
3. Mark the release as ready (in the publish-release job).

I don't have permissions to create release in this repository, see [this run](https://github.com/asterinas/asterinas/actions/runs/20131914425/job/57775230087?pr=2735#step:4:12). So it's better to manually trigger this workflow once to check whether this repository can create release.

### DIFF
--- a/.github/workflows/publish_aster_nixos.yml
+++ b/.github/workflows/publish_aster_nixos.yml
@@ -1,0 +1,105 @@
+name: Publish AsterNixOS
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  push-cachix:
+    runs-on: ubuntu-4-cores-150GB-ssd
+    container:
+      image: asterinas/asterinas:0.16.2-20251209
+      options: -v /dev:/dev --privileged
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build Asterinas NixOS patched packages and push to release cache
+        run: |
+          make kernel BOOT_PROTOCOL=linux-efi-handover64
+          export CACHIX_AUTH_TOKEN=${{ secrets.CACHIX_AUTH_TOKEN_FOR_RELEASE_CACHE }}
+          make push_cachix USE_RELEASE_CACHE=1
+          echo "Push cachix succeeds!"
+
+  create-github-release:
+    runs-on: ubuntu-latest
+    needs: [ push-cachix ]
+    outputs:
+      tag: ${{ steps.tag.outputs.tag }}
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Get Asterinas version
+        id: tag
+        run: |
+          ASTER_VERSION=$(cat ./VERSION)
+          echo "ASTER_VERSION=${ASTER_VERSION}"
+          echo "tag=${ASTER_VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Create release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          tag="${{ steps.tag.outputs.tag }}"
+          body="TODO: Write the release notes for version $tag"
+          echo "tag=$tag"
+          echo "body=$body"
+          if gh release view "$tag" > /dev/null 2>&1; then
+            echo "Release with tag $tag already exists. Skipping creation."
+          else
+            gh release create --draft "$tag" --title "$tag" --notes "$body"
+          fi
+
+  build-iso:
+    runs-on: ubuntu-latest
+    needs: [ create-github-release ]
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ 'x86_64' ]
+    timeout-minutes: 60
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,103 @@
+name: Publish release
```
tatetian: We can simply name it "Publish AsterNixOS". Releasing the ISO and pushing to Cachix serves the ultimate purpose of publishing AsterNixOS. Rename this workflow file accordingly.

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,104 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+    - '*'
```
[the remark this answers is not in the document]
tatetian: The pattern should be changed to match version tags:

```
    tags:
        - v*
```

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,104 @@
...
+    container:
+      image: asterinas/asterinas:0.16.2-20251209
+      options: -v /dev:/dev --privileged
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build Asterinas NixOS patched packages and push to release cache
+        run: |
+          make kernel BOOT_PROTOCOL=linux-efi-handover64 || true
```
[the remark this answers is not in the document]
tatetian: Yes, I have the same question. Why ignore potential build failures when doing `make kernel`?

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,103 @@
...
+          export CACHIX_AUTH_TOKEN=${{ secrets.CACHIX_AUTH_TOKEN_FOR_RELEASE_CACHE }}
+          make push_cachix USE_RELEASE_CACHE=1 2>&1 | tee cachix.log || true
+          tail --lines 10 cachix.log | grep -q -E "^(All done|Nothing to push)" || (echo "Push cachix failed" && exit 1)
```
tatetian: Why do we want to check the log? This is not a test. Shouldn't the exit code of `make push_cachix` already imply if the push succeeds or fails?

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,103 @@
+name: Publish release
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+    - '*'
+
+jobs:
+  publish-release-package-cache:
```
tatetian: Simplify this name to `push-cachix`, which is more concise but equally clear.

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Build Asterinas NixOS patched packages and push to release cache
+        run: |
+          make kernel BOOT_PROTOCOL=linux-efi-handover64 || true
+          export CACHIX_AUTH_TOKEN=${{ secrets.CACHIX_AUTH_TOKEN_FOR_RELEASE_CACHE }}
+          make push_cachix USE_RELEASE_CACHE=1 2>&1 | tee cachix.log || true
+          tail --lines 10 cachix.log | grep -q -E "^(All done|Nothing to push)" || (echo "Push cachix failed" && exit 1)
+          echo "Push cachix succeeds!"
+
+  create-release:
```
tatetian: To be more specific, rename this to `create-github-release`

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo "ASTER_VERSION=${ASTER_VERSION}"
+          echo "tag=${ASTER_VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Create release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          tag="${{ steps.tag.outputs.tag }}"
+          body="asterinas-nixos/asterinas@$tag"
```
tatetian: This field is the body of the Github Release, right? If so, we can make it more human readable:

```
body="TODO: Write the release notes for version $tag"
```

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,104 @@
...
+          echo "tag=$tag"
+          echo "body=$body"
+          gh release create --draft "$tag" --title "$tag" --notes "$body"
```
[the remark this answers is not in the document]
tatetian: The suggestion makes sense. In case of failure, this workflow may be triggered repeatedly. So we need to ensure that if an intermediate job or step has succeeded, then rerunning it would not report any error.

PR #2744, opened 2025-12-12
Split `vmar/mod.rs` into several files

junyang-zh:
`vmar/mod.rs` has ~1.7 kLoC. It should be separated into multiple files according the functionality it implements.

### DIFF
--- a/kernel/src/vm/mod.rs
+++ b/kernel/src/vm/mod.rs
@@ -3,18 +3,11 @@
 //! Virtual memory (VM).
 //!
 //! There are two primary VM abstractions:
-//!  * Virtual Memory Address Regions (VMARs) a type of capability that manages
-//!    user address spaces.
-//!  * Virtual Memory Objects (VMOs) are are a type of capability that
-//!    represents a set of memory pages.
-//!
-//! The concepts of VMARs and VMOs are originally introduced by
-//! [Zircon](https://fuchsia.dev/fuchsia-src/reference/kernel_objects/vm_object).
-//! As capabilities, the two abstractions are aligned with our goal
-//! of everything-is-a-capability, although their specifications and
-//! implementations in C/C++ cannot apply directly to Asterinas.
-//! In Asterinas, VMARs and VMOs, as well as other capabilities, are implemented
-//! as zero-cost capabilities.
+//!  * The VMAR (used to be Virtual Memory Address Region, now an orphan
+//!    initialism) represents the entire virtual address space of a process;
+//!  * The VMO (Virtual Memory Object) is a set of logically contiguous memory
+//!    frames that can be mapped into one virtual address range. Frames in a
+//!    VMO can be non-contiguous in physical memory.
 
 use osdk_frame_allocator::FrameAllocator;
 use osdk_heap_allocator::{HeapAllocator, type_from_layout};


lrh2000:
If we want to avoid mixing the VMAR impls with other modules like `vm_mapping` and `interval_set`, how do you like this?

```
vmar/
 |- util.rs
 |- vm_mapping.rs
 |- interval_set.rs
 |- vmar_def.rs
 |- vmar_impls/
    |- query.rs
    |- fork.rs
    |- ...
```

`util.rs` is for storing things like `is_intersected`, which currently lives in `vmar/mod.rs`, but I don't think it belongs in `vmar_def.rs` as it is also used in `vm_mapping.rs`...?

---

**Edit:** VMAR impls should be able to access VMAR's private fields. This means that `vmar_impls/*.rs` should be moved to `vmar_def/*.rs`. However, I'm hesitant about the module name. Is `vmar_def` (or `vmar_impl`?) good enough, or is there a better name?

lrh2000:
I expect this to go before any other PRs that touch the VMAR module because rebasing changes like this is (1) unnecessary, (2) error-prone, (3) difficult to review, and (4) tedious. What do you think?

I may fix MM syscalls' error codes and/or perform some trivial code simplifications that touch the VMAR mode (slightly). I'm inclined to wait for this PR to be merged first.

junyang-zh:
> I expect this to go before any other PRs that touch the VMAR module because rebasing changes like this is (1) unnecessary, (2) error-prone, (3) difficult to review, and (4) tedious. What do you think?
> 
> I may fix MM syscalls' error codes and/or perform some trivial code simplifications that touch the VMAR mode (slightly). I'm inclined to wait for this PR to be merged first.

OK. Sorry for the late reply.

> how do you like this?

Yes I thought about this layout earlier but I have no inclinations. I will address your concerns about the new file layout. One thing is that `vmar_def.rs` should be replaced with `mod.rs`. Otherwise we should mark all field defs with `pub(super)`? A little bit verbose and useless.

lrh2000:
> One thing is that `vmar_def.rs` should be replaced with `mod.rs`. Otherwise we should mark all field defs with `pub(super)`? A little bit verbose and useless.

Yeah. I'm not sure why it isn't `vmar_impls/mod.rs` though. I suppose that's more accurate, since `vm_mapping.rs`/`util.rs`/`interval_set.rs`/... shouldn't access VMAR's private fields. Also, we should be able to avoid `use super::*` in `vmar_impls/mod.rs` if we do so.

kernel/src/vm/vmar/mod.rs
```diff
@@ -1,928 +1,18 @@
...
-
-        self.check_extra_size_fits_rlimit(new_map_end - old_map_end)?;
-        let last_mapping = self.remove(&last_mapping_addr).unwrap();
-        let last_mapping = last_mapping.enlarge(new_map_end - old_map_end);
-        self.insert_try_merge(last_mapping);
-        Ok(())
-    }
-}
+use ostd::mm::{MAX_USERSPACE_VADDR, Vaddr};
+use vmar_impls::RssDelta;
```
lrh2000: Please remove this? I don't think it's useful.

kernel/src/vm/vmar/mod.rs
```diff
@@ -1,928 +1,18 @@
...
+pub use vmar_impls::{RssType, Vmar};
 
 pub const VMAR_LOWEST_ADDR: Vaddr = 0x001_0000; // 64 KiB is the Linux configurable default
 const VMAR_CAP_ADDR: Vaddr = MAX_USERSPACE_VADDR;
```
lrh2000: Since `MAX_USERSPACE_VADDR` is only used once, consider using `ostd::mm::MAX_USERSPACE_VADDR` instead.

Mixing `pub` imports, imports for the current file, and imports for sub-modules in `mod.rs` is always painful to read and maintain. Therefore, I'd prefer to
1. avoid any imports for sub-modules (like the comment above) and
2. minimize imports for the current file if possible (like this comment).

junyang-zh:
@cchanging take a review?

cchanging (approved):
LGTM! Thanks for the contribution! Let's leave it to @lrh2000 to merge this PR.

PR #2745, opened 2025-12-12
xfdesktop4: Use consistent monitor identifiers for wallpaper

TinaZhangZW:
The settings dialog used gdk_monitor_get_model() which often returns NULL, generating invalid property paths. The main daemon uses xfw_monitor_get_connector(), causing a mismatch that prevented wallpaper changes.

Fix by using GDK APIs to obtain a valid monitor model name, aligning with the daemon's identification logic.


tatetian (approved):

PR #2748, opened 2025-12-13
Make configuration.nix in ISO editable

tatetian:


tatetian:
Merging a trivial change passing CI without other code owners' approvals

PR #2753, opened 2025-12-13
Define `ASTER_SCML` in `validate_scmls` workflow

taosue:
GitHub Actions doesn't load `~/.bashrc`, so `ASTER_SCML` was undefined. This PR explicitly sets it by finding all `.scml` files before running `sctrace`.


tatetian (approved):

PR #2756, opened 2025-12-14
Define `__asterinas__` macro in regression tests

lrh2000:
This PR proposes to replace:
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/test/src/apps/file_io/access_err.c#L183-L186
with:
```c
	// FIXME: Asterinas reports `EACCES` because it performs the permission check first.
#ifdef __asterinas__
	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
			0),
		   EACCES);
#else
	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
			0),
		   EBADF);
#endif
```
In this way:
 - We can test the Linux behavior more easily (by default, gcc won't define `__asterinas__` for us).
 - We can test the Asterinas behavior in CI (by defining the `__asterinas__` macro manually in the script that builds the initramfs).

`__asterinas__` is chose to match the macro name used in Linux ([godbolt link](https://godbolt.org/z/joK55d4vb)):
```c
#include <stdio.h>

int main(void)
{
#ifdef __linux__
	puts("Linux!");
#else
	puts("Not Linux!");
#endif
	return 0;
}
```

---

Additionally, `TEST_SUCC` should work perfectly for `mmap` because it checks `errno`:
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/test/src/apps/test.h#L105
There should not be any difference between `mmap` and other system calls in that respect. Therefore, let's remove the unnecessary `CHECK_MM` and use `TEST_SUCC` instead.
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/test/src/apps/mmap/mmap_and_mremap.c#L64

### DIFF
--- a/test/src/apps/mmap/mmap_and_mremap.c
+++ b/test/src/apps/mmap/mmap_and_mremap.c
@@ -65,23 +65,21 @@ END_TEST()
 
 const char *content = "kjfkljk*wigo&h";
 
-#define CHECK_MM(func) CHECK_WITH(func, _ret != MAP_FAILED)
-
 FN_TEST(mmap_and_mremap)
 {
-	char *addr = CHECK_MM(mmap(NULL, 3 * PAGE_SIZE, PROT_READ | PROT_WRITE,
-				   MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+	char *addr = TEST_SUCC(mmap(NULL, 3 * PAGE_SIZE, PROT_READ | PROT_WRITE,
+				    MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
 	TEST_SUCC(munmap(addr, 3 * PAGE_SIZE));
 
-	addr = CHECK_MM(mmap(addr, PAGE_SIZE, PROT_READ | PROT_WRITE,
-			     MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0));
+	addr = TEST_SUCC(mmap(addr, PAGE_SIZE, PROT_READ | PROT_WRITE,
+			      MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0));
 	strcpy(addr, content);
 
-	char *addr2 = CHECK_MM(
+	char *addr2 = TEST_SUCC(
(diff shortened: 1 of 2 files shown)


StevenJiang1110:
Thanks for your contribution! This change makes a lot of sense. Could we perhaps add a make option to control whether the macro is enabled or disabled?

For example
```bash
# Enable the basic test for Asterinas platform
make run_kernel ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=asterinas
# Enable the basic test for Linux platform
make run_kernel ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=linux
```

lrh2000:
> ```shell
> 
> make run_kernel ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=linux
> ```
This command will certainly fail, so I'm not sure what the point of adding it is or how it could be useful. Do you have a specific use case in mind?

How do you typically test Linux behavior by running regression tests? I usually compile the tests manually:
```
root@localhost:~/asterinas# gcc test/src/apps/file_io/access_err.c -o /tmp/x
root@localhost:~/asterinas# /tmp/x
```
Then this will use the Linux version (unless someone specifies `-D__asterinas__` manually).

StevenJiang1110:
> This command will certainly fail, so I'm not sure what the point of adding it is or how it could be useful. Do you have a specific use case in mind?

Oh, sorry. It's my fault. I should mean `make kernel ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=linux`, not `make run_kernel`. 

> How do you typically test Linux behavior by running regression tests? 

Yes, this is because I usually run regression test to test Linux behavior by `make kernel ENABLE_BASIC_TEST=true` to build all regression test and then go to the target directory to run the test.

lrh2000:
Emmm, I pushed my changes (along with https://github.com/asterinas/asterinas/pull/2757#issuecomment-3653696116) to the wrong remote. Sorry about the mess...

This is not the correct practice, so let's close this PR and open a new one.

PR #2757, opened 2025-12-14
Clarify the code for a long shebang

lrh2000:
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/kernel/src/process/program_loader/shebang.rs#L15-L22

I'm confused about this. `!file_first_page.contains("\n")` is checked twice in different ways (lines 15 and 20) and corresponds to two different actions (lines 16 and 21). We should remove one check of the two?

As far as Linux is concerned, it [returns `ENOEXEC` if the interpreter is  truncated](https://elixir.bootlin.com/linux/v6.18.1/source/fs/binfmt_script.c#L64-L69), but it [allows the argument to be truncated](https://elixir.bootlin.com/linux/v6.18.1/source/fs/binfmt_script.c#L54-L56). I think we should return `ENOEXEC` for both cases because we don't support the Linux behavior of second one (yet).


lrh2000:
Oh, why I push my changes to `origin`... This is a mistake, sorry...

PR #2759, opened 2025-12-15
Remove push_tags workflow and let publish_aster_nixos workflow create git tags

StevenJiang1110:
This PR updates the distribution version to "25.12", aligning with NixOS versioning conventions, and modifies the `publish_aster_nixos` workflow accordingly.

This PR consists of four commits:
1. The first commit changes the trigger condition for the `publish_aster_nixos` workflow. It is now activated by modifications to the VERSION file.
2. The second commit removes the `push_tag` workflow. `publish_aster_nixos` will be the sole workflow responsible for generating Git tags.
3. The third commit lets publish_aster_nixos workflow create git tags with prefix 'v'

Below is original scope of this PR, which has been outdated.

~The publish_aster_nixos workflow, which is configured to trigger on tag pushes, is currently not being executed successfully.~

~This is because our Git tags are added by a GitHub workflow. However, the previous tag push condition (on: push: tags: ...) is typically only activated by tags that are pushed manually from a local Git client, for example, using commands like:~

~```bash~
~git tag v1.0.0~
~git push origin v1.0.0~
~```~

~To resolve this, this PR reverts to using a path-based condition for triggering the workflow.~

### DIFF
--- a/.github/workflows/publish_aster_nixos.yml
+++ b/.github/workflows/publish_aster_nixos.yml
@@ -54,6 +54,12 @@ jobs:
           echo "tag=$tag"
           echo "tag=$tag" >> $GITHUB_OUTPUT
 
+      - name: Add Git tag
+        uses: pxpm/github-tag-action@1.0.1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          tag: ${{ steps.tag.outputs.tag }}
+
       - name: Create release
         env:
           GH_TOKEN: ${{ github.token }}
@@ -66,7 +72,7 @@ jobs:
           if gh release view "$tag" > /dev/null 2>&1; then
             echo "Release with tag $tag already exists. Skipping creation."
           else
-            gh release create --draft "$tag" --title "$tag" --notes "$body"
+            gh release create --draft "$tag" --title "$tag" --notes "$body" --verify-tag
           fi
 
   build-iso:


.github/workflows/publish_aster_nixos.yml
```diff
@@ -3,8 +3,10 @@ name: Publish AsterNixOS
 on:
   workflow_dispatch:
   push:
-    tags:
-      - 'v*'
+    branches:
+      - main
+    paths: 
+      - VERSION
```
tatetian, line 9: I see. So Github releases have nothing to dow with Git tags. In our arrangement, they happens to have a 1:1 mapping. But from Github's perspective, it is perfectly fine to create a release out of an non-existential tag. Am I right?

StevenJiang1110: GitHub releases are fundamentally tied to Git tags.

Our current release process utilizes the `gh release create` command. According to its [manual](https://cli.github.com/manual/gh_release_create), it states:

```plain
If a matching git tag does not yet exist, 
one will automatically get created from the latest state of the default branch.
```

This implies that we can create a GitHub release even if the corresponding Git tag hasn't been explicitly pushed beforehand.

Currently, on our [tags page](https://github.com/asterinas/asterinas/tags), we observe two distinct tags for seemingly the same version:`v0.17.0` and `0.17.0`. The `v0.17.0` tag is generated by the `push_tags` workflow, while the `0.17.0` tag is created by the `publish_aster_nixos` workflow (which this PR modifies).

Considering this, maintaining both v0.17.0 and 0.17.0 tags seems redundant. 

- Should we only leave one workflow for our tag creation? Perhaps we should remove the `push_tags` workflow and allow `publish_aster_nixos` to be the sole generator of Git tags.

- What is our preferred tagging convention going forward? Should we standardize on tags with a 'v' prefix (e.g., v0.17.0) or without (e.g., 0.17.0)?"

tatetian: > maintaining both v0.17.0 and 0.17.0 tags seems redundant.

They are definitely redundant. And 0.17.0 should be removed.

> * Should we only leave one workflow for our tag creation? Perhaps we should remove the `push_tags` workflow and allow `publish_aster_nixos` to be the sole generator of Git tags.

Yes, remove `push_tags`.

> * What is our preferred tagging convention going forward? Should we standardize on tags with a 'v' prefix (e.g., v0.17.0) or without (e.g., 0.17.0)?"

Following the established convention (such as `v0.17.0`).

definfo:
> This PR updates the distribution version to "25.12", aligning with NixOS versioning conventions

AFAIK, NixOS (and nixpkgs channel) versions always end with "05" or "11" (instead of 12), which stand for half-year release months, and unstable version simply concatenate the next release version with commit ID.
Whatever this may not be an issue if Asterinas differs on release management.

StevenJiang1110:
> AFAIK, NixOS (and nixpkgs channel) versions always end with "05" or "11" (instead of 12), which stand for half-year release months, and unstable version simply concatenate the next release version with commit ID. Whatever this may not be an issue if Asterinas differs on release management.

Thanks for your information at first! 

We think that the versioning for Asterinax NixOS should not be strongly coupled with NixOS's release cycles. Given Asterinax NixOS's rapid development and potentially faster update cadence, a bi-annual release schedule is unlikely to be suitable. For this reason, we initially adopted a YY.MM format, but  MM could be any month, not just the 05 and 11 typically seen in NixOS.

However, following some internal discussions yesterday, we concluded that we are still quite far from being production-ready. Adopting a YY.MM version format at this stage might convey misleading information to users.

Therefore, we've decided, for the time being, to align the distribution's version number with the kernel's x.y.z format (i.e., 0.17.0 at this moment). Once Asterinax NixOS approaches production readiness, we will revisit and implement a YY.MM versioning scheme for the distribution. At that point, we anticipate an RFC process to formally discuss version management, including considerations about its potential alignment with upstream NixOS versions.

I will revert the portion of this PR that pertains to the distro version.

tatetian:
> AFAIK, NixOS (and nixpkgs channel) versions always end with "05" or "11" (instead of 12), which stand for half-year release months, and unstable version simply concatenate the next release version with commit ID.

Well said

.github/workflows/publish_aster_nixos.yml
```diff
@@ -96,11 +101,12 @@ jobs:
...
         run: |
+          kernel_version="${{ needs.create-github-release.outputs.kernel_version }}"
           tag="${{ needs.create-github-release.outputs.tag }}"
-          echo "upload asterinas-nixos-$tag-${{ matrix.arch }}.iso"
+          echo "upload asterinas-nixos-$kernel_version-${{ matrix.arch }}.iso"
           cd ${{ github.workspace }}
-          mv ./asterinas-nixos.iso asterinas-nixos-$tag-${{ matrix.arch }}.iso
-          gh release upload "$tag" "asterinas-nixos-$tag-${{ matrix.arch }}.iso"
+          mv ./asterinas-nixos.iso asterinas-nixos-$kernel_version-${{ matrix.arch }}.iso
+          gh release upload "$tag" "asterinas-nixos-$kernel_version-${{ matrix.arch }}.iso"
```
tatetian, line 116: As I understand it, this command will create a new tag named `$tag` if this tag does not exist already. What about the message associated with the tag?

StevenJiang1110: > As I understand it, this command will create a new tag named $tag if this tag does not exist already.

Exactly

> What about the message associated with the tag?

Emmm, actually, I can't see any message associated with the tags on my own GitHub repository. This makes it confusing to me why tags in _this_ repository are displaying a commit message.

After some online research, it appears there are two primary types of [Git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging): **annotated tags** and **lightweight tags**.

My understanding is that GitHub typically displays the commit message for lightweight tags when you view them, but for annotated tags, it will not display.

However, the `gh release` command (i.e., GitHub CLI), when used to create a release for a tag that doesn't pre-exist, seems to generate a lightweight tag by default. This behavior appears to be [a known issue](https://github.com/orgs/community/discussions/4924) within GitHub CLI. This likely explains why the tags were showing the associated commit message (though my own repository's behavior remains puzzling).

To address this, in the latest commit, I've reverted to using our previously established GitHub Action for pushing Git tags. This method ensures that the tags created are annotated tags, which should resolve the issue of commit messages being displayed.

tatetian (approved):

PR #2761, opened 2025-12-15
Define `__asterinas__` macro in regression tests

lrh2000:
Reopen https://github.com/asterinas/asterinas/pull/2756 due to https://github.com/asterinas/asterinas/pull/2756#issuecomment-3653728481.

---

This PR proposes to replace:
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/test/src/apps/file_io/access_err.c#L183-L186
with:
```c
	// FIXME: Asterinas reports `EACCES` because it performs the permission check first.
#ifdef __asterinas__
	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
			0),
		   EACCES);
#else
	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
			0),
		   EBADF);
#endif
```
In this way:
 - We can test the Linux behavior more easily (by default, gcc won't define `__asterinas__` for us).
 - We can test the Asterinas behavior in CI (by defining the `__asterinas__` macro manually in the script that builds the initramfs).

`__asterinas__` is chose to match the macro name used in Linux ([godbolt link](https://godbolt.org/z/joK55d4vb)):
```c
#include <stdio.h>

int main(void)
{
#ifdef __linux__
	puts("Linux!");
#else
	puts("Not Linux!");
#endif
	return 0;
}
```

---

Additionally, `TEST_SUCC` should work perfectly for `mmap` because it checks `errno`:
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/test/src/apps/test.h#L105
There should not be any difference between `mmap` and other system calls in that respect. Therefore, let's remove the unnecessary `CHECK_MM` and use `TEST_SUCC` instead.
https://github.com/asterinas/asterinas/blob/ecd0ac96626dc2ab7c9251a61b45fcb0b4f44f0c/test/src/apps/mmap/mmap_and_mremap.c#L64

### DIFF
--- a/test/src/apps/file_io/access_err.c
+++ b/test/src/apps/file_io/access_err.c
@@ -181,9 +181,15 @@ FN_TEST(path)
 
 	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_SHARED, fd, 0), EBADF);
 	// FIXME: Asterinas reports `EACCES` because it performs the permission check first.
-	// TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
-	// 		0),
-	// 	   EBADF);
+#ifdef __asterinas__
+	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
+			0),
+		   EACCES);
+#else
+	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
+			0),
+		   EBADF);
+#endif
 
 	TEST_ERRNO(flock(fd, LOCK_SH), EBADF);
 	TEST_ERRNO(flock(fd, LOCK_UN), EBADF);
@@ -214,9 +220,15 @@ FN_TEST(path)
 
 	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_SHARED, fd, 0), EBADF);
 	// FIXME: Asterinas reports `EACCES` because it performs the permission check first.
-	// TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
-	// 		0),
-	// 	   EBADF);
+#ifdef __asterinas__
+	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
+			0),
+		   EACCES);
+#else
+	TEST_ERRNO(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
+			0),
+		   EBADF);
+#endif
 
 	TEST_ERRNO(flock(fd, LOCK_SH), EBADF);
 	TEST_ERRNO(flock(fd, LOCK_UN), EBADF);
--- a/test/src/apps/mmap/mmap_and_mremap.c
+++ b/test/src/apps/mmap/mmap_and_mremap.c
@@ -42,8 +42,12 @@ FN_TEST(mremap)
 	TEST_ERRNO(mremap(addr, 0, PAGE_SIZE, 0), EINVAL);
 
 	// There is no enough room to expand the mapping.
-	// FIXME: Asterinas returns EACCESS here, which is not a correct error code.
-	// TEST_ERRNO(mremap(addr, PAGE_SIZE, 2 * PAGE_SIZE, 0), ENOMEM);
+	// FIXME: Asterinas returns EACCES here, which is not a correct error code.
+#ifdef __asterinas__
+	TEST_ERRNO(mremap(addr, PAGE_SIZE, 2 * PAGE_SIZE, 0), EACCES);
+#else
+	TEST_ERRNO(mremap(addr, PAGE_SIZE, 2 * PAGE_SIZE, 0), ENOMEM);
+#endif
 
 	char *hole = addr + 2 * PAGE_SIZE;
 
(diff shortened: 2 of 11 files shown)


lrh2000:
Reply to @StevenJiang1110's comment https://github.com/asterinas/asterinas/pull/2756#issuecomment-3652646728 here:

I think you mean something like https://github.com/asterinas/asterinas/pull/2761/commits/e7c80b0a76e23325a077831dd6fef09a1b38f877?

I'm too infamilar with Nix scripts, so I don't know if I'm doing something silly. Please point out if there are any mistakes. Thanks!

At least I test `BASIC_TEST_PLATFORM=linux` and it seems to work...

Also, should we retrieve Makefile options through environment variables or via argument passing? I see that INTEL_TDX is passed through environment variables, which are different from the other configuration options. It seems confusing?
https://github.com/asterinas/asterinas/blob/49a2cecc8126f0de7ef637e4be5c763f2db78699/test/nix/apps.nix#L4-L5
https://github.com/asterinas/asterinas/blob/49a2cecc8126f0de7ef637e4be5c763f2db78699/test/nix/apps.nix#L24-L26

Makefile
```diff
@@ -21,7 +21,12 @@ OSTD_TASK_STACK_SIZE_IN_PAGES ?= 64
 FEATURES ?=
 NO_DEFAULT_FEATURES ?= 0
 COVERAGE ?= 0
+# Specify whether to build regression tests under `test/src/apps`.
 ENABLE_BASIC_TEST ?= false
+# Specify platform macros when building regression tests (supported: asterinas, linux).
+# - asterinas: Define both `__asterinas__` and `__linux__`. Tests may fail in Linux.
+# - linux: Define only `__linux__`. Tests may fail in Asterinas.
+BASIC_TEST_PLATFORM ?= asterinas
```
tatetian: I suggest removing this option from the root `Makefile` and only keep it in the `Makefile` under `test/apps`. There are two reasons:
1. There are already too many knobs in the root `Makefile` and `apps.nix`. So I think we ought to be more refrained when it comes to adding new options to the root `Makefile`.
2. If we want to test a program under Linux, we can directly go to `test/apps` and do `make`. I can't see the pointing of doing this using the root `Makefile`.
3. Testing Linux-targeting apps under Asterinas could be useful, but I don't think it would be done often:
    ```
    make run_kernel AUTO_TEST=test BASIC_TEST_PLATFORM=linux
    ```
    In the rare occasions that one does want to do this, he can manually change the environment variable in `tests/Makefile` instead.

lrh2000: It seems that, even if I define `BASIC_TEST_PLATFORM` only under `test` (like [the latest changes](https://github.com/asterinas/asterinas/compare/26fe5ad0f2ed466f28e1941254d90799804228f8..2e22aacd7db083a88ffe1f8cfc623c9faa3196b5)), we can still use it when running `make` at the top level. The environment variables will be passed regardless of whether the option is listed in the `Makefile`.

> make run_kernel AUTO_TEST=test BASIC_TEST_PLATFORM=linux

This will definitely **never** be useful. The tests will fail. Meanwhile, the tests will stop after the first failed test. Therefore, it is unreliable to see all failed tests.

This option was added at the request of @StevenJiang1110 because it can be useful for building tests for the Linux platform without running them in Asterinas (see https://github.com/asterinas/asterinas/pull/2756#issuecomment-3652646728):

> make kernel ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=linux

StevenJiang1110: > This option was added at the request of @StevenJiang1110 because it can be useful for building tests for the Linux platform without running them in Asterinas (see https://github.com/asterinas/asterinas/pull/2756#issuecomment-3652646728):

Thanks for your work. I've tested locally, and the following commands function as expected for testing these apps on Linux or Asterinas.
```bash
make initramfs ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=linux
```
or 
```bash
make initramfs ENABLE_BASIC_TEST=true BASIC_TEST_PLATFORM=asterinas
```

StevenJiang1110:
> I'm too infamilar with Nix scripts, so I don't know if I'm doing something silly. Please point out if there are any mistakes. Thanks!

I think the current changes to Nix scripts is in good shape. 

> Also, should we retrieve Makefile options through environment variables or via argument passing? I see that INTEL_TDX is passed through environment variables, which are different from the other configuration options. It seems confusing?

Ideally, we should retrieve Makefile options via argument passing. Indeed, we should avoid using environment variables in Nix scripts. 

The primary reason to avoid environment variables in Nix scripts stems from Nix's fundamental design philosophy, which means purity.

<details><summary>Some explanation about purity generated by LLM</summary>
<p>

In the context of Nix, a pure function (or a pure Nix expression) is one whose output depends solely on its declared inputs and nothing else. This means:
- Given the same inputs, it will always produce the exact same output.
- It has no side effects on the outside world, and its output is not influenced by any external state.
How Environment Variables Break Purity:

If a Nix script reads an environment variable (e.g., CC, HOME, MY_CUSTOM_FLAG), it immediately becomes impure. The script's behavior (and thus its output) now depends on something outside its explicit inputs.

- If user A runs the script with export MY_VAR="value1", they get Output A.
- If user B runs the exact same script with export MY_VAR="value2", they get Output B.

Even if user A runs it again later, but their MY_VAR has changed, they get a different output.
This directly violates the first rule of purity: "Given the same inputs, it will always produce the exact same output." The script itself is the same, but an external factor (the environment variable) changed the result.

</p>
</details>

While direct use of environment variables in Nix scripts is generally discouraged, they might occasionally be employed for convenience in specific contexts. I'm uncertain if INTEX_TDX falls into that category.

lrh2000:
> The primary reason to avoid environment variables in Nix scripts stems from Nix's fundamental design philosophy, which means purity.

Thanks for the explanation! It sounds reasonable to me and should make sense.

PR #2763, opened 2025-12-16
Add `Sized` trait bound to `NonNullPtr`

rikosellic:
This PR adds the `Self: Sized` trait bound to `NonNullPtr::from_raw` and `NonNullPtr::into_raw` as their arguments require a size known at compile-time. Adding this trait bound does not change the code behavior, as the Rust compiler will reject attempts to implement it with an unsized type. 

The main reason for this change is that the formal verification tool [Verus](https://github.com/asterinas/vostd/pull/253) will reject it earlier, when checking the trait definition. I also believe that this modification makes the trait bound requirement a bit more explicit.

### DIFF
--- a/ostd/src/sync/rcu/non_null/mod.rs
+++ b/ostd/src/sync/rcu/non_null/mod.rs
@@ -23,7 +23,7 @@ use crate::prelude::*;
 /// raw pointers.
 ///
 /// [`Rcu`]: super::Rcu
-pub unsafe trait NonNullPtr: 'static {
+pub unsafe trait NonNullPtr: Sized + 'static {
     /// The target type that this pointer refers to.
     // TODO: Support `Target: ?Sized`.
     type Target;


ostd/src/sync/rcu/non_null/mod.rs
junyang-zh: It's probably better to just:

```
pub unsafe trait NonNullPtr: Sized + 'static {
```

Since doing `raw_as_ref`/`ref_as_raw` on `?Sized` pointer doesn't make sense anyway.

rikosellic: Good point, I've updated the code, thanks!

tatetian:
> The main reason for this change is that the formal verification tool [Verus](https://github.com/asterinas/vostd/pull/253) will reject it earlier, when checking the trait definition. I also believe that this modification makes the trait bound requirement a bit more explicit.

I think you need to justify the change from the system code itself, not the verification code. From the system code perspective, having the `Sized` trait bound is redundant, I suppose.

Verus and by extension, the vostd project, has many restraints on the look of verifiable Rust code. If adding this trait bound alone is sufficient to make the entire OSTD codebase verifiable, then I would be very eager to merge it. But this is one of the hundreds of changes required to make the code verifiable. So why you singled out this particular change to the upstream? And do you plan to submit more changes like this?

rikosellic:
In the vostd project, our guiding principle is that translating source code into verifiable code should not change the behavior of the original code.

In practice, due to limitations of Verus, we often need to make adjustments such as adding ghost parameters to function signatures, or wrapping third-party library functions.

However, these changes are purely a transcription of the original code into a form that Verus can reason about. They do not alter the structure or semantics of the original code, and therefore should not be reflected in the upstream system code. For this reason, I have not submitted such changes in the past, and I do not plan to do so in the future.

Introducing this bound, however, is creating a new thing that does not exist in the original source code, and it creates the possibility of changing the program’s behavior.  That said, if we all agree that adding or omitting this bound in the original source code would not change its behavior, then it would indeed be unnecessary to modify the source code itself.

tatetian:
> Introducing this bound, however, is creating a new thing that does not exist in the original source code, and it creates the possibility of changing the program’s behavior. That said, if we all agree that adding or omitting this bound in the original source code would not change its behavior, then it would indeed be unnecessary to modify the source code itself.

Type generic parameters are `Sized` by default. If Verus is not aware of this, isn't this a bug or limitation of Verus? I don't think it makes sense to workaround a Verus limitation in the system code.

Not altering the behaviors of the system code behavior is not a strong argument. In theory, any transformation to make the system code verifiable should make the resulted code equivalent to the origin one, right?

You need to make an argument as to why the proposed change _improves_ the system code, including its impact to _readability_, _conciseness_, and _maintainability_. Adding this `Sized` trait would make people start wondering why it is added and whether there are any deeper reasons.

PR #2764, opened 2025-12-17
Handle the QEMU exit code

cchanging:
Currently, our QEMU exit logic is implemented through the isa-debug-exit device, which undergoes certain encoding processing by QEMU. While OSDK contains handling logic for this mechanism, it was not being executed in the distribution version, resulting in persistent errors when exiting QEMU previously.  

This PR fixes that issue. Additionally, we now support the ACPI reset operation. With this operation, QEMU will return exit code 0, hence this PR also adds the check for the corresponding condition in OSDK.


tools/nixos/run_iso.sh
```diff
@@ -37,4 +37,4 @@ QEMU_ARGS="
 	${COMMON_QEMU_ARGS} \
 "
 
-${QEMU_ARGS}
+${SCRIPT_DIR}/run_common.sh ${QEMU_ARGS}
```
tatetian: The logic in `run_common.sh` is not the only code that is duplicated between `run_iso.sh` and `run_nixos.sh`. I suggest merging the three scripts into one script, `run.sh`, to reduce duplicated code to the bare minimum. The usage of the new script is shown below:

```
Run an NixOS installation or NixOS ISO installer image built by the root Makefile inside a VM

Usage: ./run.sh [nixos | iso]
```

tools/nixos/run_common.sh
```diff
@@ -0,0 +1,24 @@
...
+# isa-debug-exit device.
+KERNEL_SUCCESS_EXIT_CODE=16 # 0x10 in hexadecimal
+# QEMU translates the value written to the isa-debug-exit device into a final
+# process exit code using following formula.
+QEMU_SUCCESS_EXIT_CODE=$(((KERNEL_SUCCESS_EXIT_CODE << 1) | 1))
+
+"$@" || exit_code=$?
+exit_code=${exit_code:-0}
+
+if [ ${exit_code} -eq 0 ] || [ ${exit_code} -eq ${QEMU_SUCCESS_EXIT_CODE} ]; then
```
tatetian: For maximal clarity, it would be better to explain why 0 also means success in addition to `$QEMU_SUCCESS_EXIT_CODE`

tatetian (commented):
Great to see the exit code of `make run_nixos` being fixed!

PR #2771, opened 2025-12-17
Refactor the framework for Asterinas NixOS tests

### LINKED ISSUE #2822: Support interactive NixOS test
Currently, our NixOS test suite only supports non-interactive testing. However, certain scenarios require interaction—for instance, when testing Podman, we may need the `-it` flags to access a container's internal shell.

There are mature solutions available to address this. The core approach involves intercepting the QEMU process's stdio, allowing us to provide direct input and verify output in real-time. This method enables interactive testing and eliminates our current dependency on the `qemu.log` file.

Below is an example using the Rust [rexpect library](https://crates.io/crates/rexpect). This example demonstrates how to test an interactive session such as `podman run -it docker.1ms.run/alpine`. While this example is in Rust, we could also consider using Python's [pexpect library](https://pexpect.readthedocs.io/en/stable/overview.html) as an alternative.

```rust
use std::{env::set_current_dir, process::Command};

use rexpect::process::wait::WaitStatus;

fn main() {
    set_current_dir("/root/asterinas").unwrap();

    // Edit `configuration.nix` to include Podman support
    let edit_status = Command::new("nix-editor")
        .args([
            "distro/etc_nixos/configuration.nix",
            "virtualisation.podman.enable",
            "--val",
            "true",
            "--inplace",
        ])
        .status()
        .unwrap();
    assert!(edit_status.success());

    let build_status = Command::new("make").arg("nixos").status().unwrap();
    assert!(build_status.success());
    
    // Start QEMU and wait for the root prompt
    println!("run `make run_nixos`");
    let mut qemu_process = rexpect::spawn("make run_nixos", Some(60_000)).unwrap();
    qemu_process.exp_string("root@asterinas:").unwrap();

    // Test `podman -v`
    println!("run `podman -v`");
    qemu_process.send_line("podman -v").unwrap();
    qemu_process.exp_string("podman version 5.4.1").unwrap();

    // Test `podman run -it`
    println!("run `podman run -it docker.1ms.run/alpine`");
    qemu_process.send_line("podman run -it docker.1ms.run/alpine").unwrap();
    qemu_process.exp_string("/ #").unwrap();

    // Test alpine version
    println!("run `ls /etc/alpine-release`");
    qemu_process.send_line("ls /etc/alpine-release").unwrap();
    qemu_process.exp_string("/etc/alpine-release").unwrap();

    // Exit the container
    println!("run `exit`");
    qemu_process.send_line("exit").unwrap();
    qemu_process.exp_string("root@asterinas:").unwrap();

    // Poweroff
    println!("run `poweroff`");
    qemu_process.send_line("poweroff").unwrap();

    qemu_process.flush().unwrap();

    let status = qemu_process.process.wait().unwrap();
    let pid = status.pid().unwrap();
    // FIXME: Currently, the exit code of Makefile is incorrect, and maybe a non-zero value
    assert_eq!(status, WaitStatus::Exited(pid, 0));
}
```

cchanging:
Fixes #2822.

This PR attempts to provide a more convenient method for adding NixOS tests.

### DIFF
--- a/tools/github_workflows/config_for_publish_cachix.nix
+++ b/tools/github_workflows/config_for_publish_cachix.nix
@@ -1,0 +1,12 @@
+# This configuration file enables all packages and services that are modified
+# by Asterinas NixOS. By enabling them, we ensure they are built during the
+# evaluation and can subsequently be uploaded to our Cachix binary cache.
+
+{ config, lib, pkgs, ... }:
+
+{
+  services.xserver.enable = true;
+  services.xserver.desktopManager.xfce.enable = true;
+
+  virtualisation.podman.enable = true;
+}
--- a/.github/workflows/publish_aster_nixos.yml
+++ b/.github/workflows/publish_aster_nixos.yml
@@ -20,11 +20,9 @@ jobs:
 
       - name: Modify configuration.nix to include xfce and podman
         run: |
-          sed -i \
-            -e 's/^[[:space:]]*# services\.xserver\.enable = true;/  services.xserver.enable = true;/' \
-            -e 's/^[[:space:]]*# services\.xserver\.desktopManager\.xfce\.enable = true;/  services.xserver.desktopManager.xfce.enable = true;/' \
-            -e 's/^[[:space:]]*# virtualisation\.podman\.enable = true;/  virtualisation.podman.enable = true;/' \
-            distro/etc_nixos/configuration.nix
+          config_path="distro/etc_nixos/configuration.nix"
+          extra_config_path="tools/github_workflows/config_for_publish_cachix.nix"
+          ./test/nixos/common/merge_nixos_config.sh ${config_path} ${extra_config_path} ${config_path}
 
       - name: Build Asterinas NixOS patched packages and push to release cache
         run: |
--- a/.github/workflows/push_cachix_dev.yml
+++ b/.github/workflows/push_cachix_dev.yml
@@ -25,11 +25,9 @@ jobs:
 
       - name: Modify configuration.nix to include xfce and podman
         run: |
-          sed -i \
-            -e 's/^[[:space:]]*# services\.xserver\.enable = true;/  services.xserver.enable = true;/' \
-            -e 's/^[[:space:]]*# services\.xserver\.desktopManager\.xfce\.enable = true;/  services.xserver.desktopManager.xfce.enable = true;/' \
-            -e 's/^[[:space:]]*# virtualisation\.podman\.enable = true;/  virtualisation.podman.enable = true;/' \
-            distro/etc_nixos/configuration.nix
+          config_path="distro/etc_nixos/configuration.nix"
+          extra_config_path="tools/github_workflows/config_for_publish_cachix.nix"
+          ./test/nixos/common/merge_nixos_config.sh ${config_path} ${extra_config_path} ${config_path}
 
       - name: Build Asterinas NixOS patched packages (dry run)
         if: github.event_name == 'pull_request'
--- a/.github/workflows/test_iso_image.yml
+++ b/.github/workflows/test_iso_image.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build Asterinas NixOS ISO installer image
         run: |
-          make iso NIXOS_TEST_COMMAND='hello-asterinas'
+          make iso NIXOS_TEST_SUITE=hello
       - name: Run ISO image and install Asterinas NixOS
         run: |
           make run_iso || true
@@ -34,6 +34,4 @@ jobs:
           echo "Test ISO succeeds!"
       - name : Run Asterinas NixOS
         run: |
-          make run_nixos || true
-          tail --lines 10 qemu.log | grep -q "^Hello Asterinas!" || (echo "Test NixOS failed" && exit 1)
-          echo "Test NixOS succeeds!"
+          make run_nixos NIXOS_TEST_SUITE=hello
--- a/.github/workflows/test_nixos_full.yml
+++ b/.github/workflows/test_nixos_full.yml
@@ -28,52 +28,21 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.0-20251228
       options: -v /dev:/dev --privileged
-    timeout-minutes: 20
-    steps:
-      - uses: actions/checkout@v4
-      - name: Modify configuration.nix to include test-asterinas
-        run: |
-          sed -i 's/environment.systemPackages = with pkgs; \[ hello-asterinas \];/environment.systemPackages = with pkgs; \[ test-asterinas \];/' \
-            distro/etc_nixos/configuration.nix
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        test:
+          - name: nix
+            disk_size: 6144
+          - name: podman
+            disk_size: 8192
 
-      - name: Test nix commands
-        run: |
-          make nixos NIXOS_DISK_SIZE_IN_MB=6144 NIXOS_TEST_COMMAND='test-nix-commands'
-          make run_nixos
-
-      - name: Check results
-        run: |
-          tail --lines 50 qemu.log | awk '
-          BEGIN { hello_count=0 } {
-            if ($0 ~ /^Hello, world!/) hello_count++
-          }
-          END {
-            if (hello_count == 4) exit 0
-            else exit 1
-          }' || (echo "Test nix commands failed" && exit 1)
-          echo "Test nix commands succeeds!"
-
-  podman-test:
-    runs-on: ubuntu-4-cores-150GB-ssd
-    container:
-      image: asterinas/asterinas:0.17.0-20251228
-      options: -v /dev:/dev --privileged
+    name: Test ${{ matrix.test.name }}
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
-      - name: Modify configuration.nix to include test-asterinas and podman
+      - name: Run the ${{ matrix.test.name }} test on AsterNixOS 
         run: |
-          sed -i \
-            -e 's/environment.systemPackages = with pkgs; \[ hello-asterinas \];/environment.systemPackages = with pkgs; \[ test-asterinas \];/' \
-            -e 's/^[[:space:]]*# virtualisation\.podman\.enable = true;/  virtualisation.podman.enable = true;/' \
-            distro/etc_nixos/configuration.nix
-
-      - name: Run podman commands
-        run: |
-          make nixos NIXOS_DISK_SIZE_IN_MB=8192 NIXOS_TEST_COMMAND='test-podman'
-          make run_nixos
-
-      - name: Check results
-        run: |
-          tail --lines 50 qemu.log | grep -q "^Test podman succeeds" || (echo "Test podman failed" && exit 1)
-          echo "Test podman succeeds!"
+          make nixos NIXOS_TEST_SUITE=${{ matrix.test.name }} NIXOS_DISK_SIZE_IN_MB=${{ matrix.test.disk_size }}
+          make run_nixos NIXOS_TEST_SUITE=${{ matrix.test.name }}
--- a/.github/workflows/test_nixos_minimal.yml
+++ b/.github/workflows/test_nixos_minimal.yml
@@ -32,15 +32,12 @@ jobs:
           remove_haskell: true
           remove_tool_cache: true
       - uses: actions/checkout@v4
-      - name: Run a Hello World program on AsterNixOS
+      - name: Run the hello test on AsterNixOS
+        timeout-minutes: 30
         uses: addnab/docker-run-action@v3
         with:
           image: asterinas/asterinas:0.17.0-20251228
           options: --privileged -v /dev:/dev -v ${{ github.workspace }}:/root/asterinas
           run: |
-            make nixos NIXOS_DISK_SIZE_IN_MB=6144 NIXOS_TEST_COMMAND='hello-asterinas'
-            make run_nixos
-      - name: Check results
-        run: |
-          tail --lines 10 ${{ github.workspace }}/qemu.log | grep -q "^Hello Asterinas!" || (echo "Test NixOS failed" && exit 1)
-          echo "Test NixOS succeeds!"
+            make nixos NIXOS_TEST_SUITE=hello NIXOS_DISK_SIZE_IN_MB=6144
+            make run_nixos NIXOS_TEST_SUITE=hello


tools/nixos/modify_nix_config.sh
```diff
@@ -0,0 +1,79 @@
...
+CONFIG_FILE=""
+CONFIG=""
+PACKAGES=""
+
+# Usage function
+usage() {
+    echo "Usage: $0 -f <file> [-c <config>] [-p <packages>]"
+    echo "  -f <file>      : Path to configuration.nix"
+    echo "  -c <config>    : Multi-line string of Nix options (e.g., 'vm.enable = true;')"
+    echo "  -p <packages>  : Space-separated string of packages to overwrite the existing list"
```
tatetian: Having a dedicated script is definitely an improvement over manually writing `sed` commands in Github Workflow files. But I am afraid that we are still not doing it in the Nix-native way. Our `test-xxx.sh` scripts should have been made Nix packages. As Nix packages, they can express their configurations and dependency packages in the Nix files. Reading and maintaining these Nix files is easier and more straightforward than doing these homebrew CLI arguments.

distro/etc_nixos/configs/enable-all.nix
```diff
@@ -0,0 +1,9 @@
+# configs/enable-all.nix
```
tatetian: What constitutes this "all" is unclear. The users of `enable-all.nix` take "all" as all packages that are modified by Asterinas NixOS and thus must be uploaded to Cachix. On the other hand, we intend to establish a framework for testing Nix packages. Not all of them are patched by us. So you see. The file name `enable-all.nix` is confusing. And it is not clear when `enable-all.nix` should be updated.

distro/tests/test-framework.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/sh
```
tatetian: This script is definitely a great help to write test programs. But this script alone is not a test framework. What we want is to set up the convention/framework to add more tests for Nix packages. Think of Python, Go, QEMU, and more. We will need to have a proper README.md to describe how to add new tests. This is missing from this PR.

If this documentation was written, then one would find that at least three locations must be modified separately to add a new test:
* `distro/etc_nixos/overlays/test-asterinas/`
* `distro/etc_nixos/configs/`
* `.github/workflows/test_nixos_full.yml`

Don't you think that the information about a Nix package test is spread over too many locations?

distro/etc_nixos/generate_config.nix
```diff
@@ -0,0 +1,42 @@
...
+          ;
+
+          base = baseModule { inherit config lib pkgs; };
+          extra = extraModule { inherit config lib pkgs; };
+        in
+          # Deeply merge the base, extra, and common configurations.
+          # The module system will combine all `environment.systemPackages` lists.
+          lib.recursiveUpdate base extra
+  '';
+in pkgs.writeText "configuration.nix" finalConfigText
```
tatetian: I can't see why this should be written as a Nix module and built with `nix-build`. A Bash script can do the job equally well. Surely, you are not using Nix for `concatMapStrings`.

Instead of calling this script `generate_config.sh` and inventing an ad-hoc interface and semantics for it, I think a simpler mental model is to think of it as _merging_ two `configuration.nix` files. So I propose to refactor this way:

```
Usage: merge_nixos_config.sh <base_file> <extra_file> <merged_file>

This script takes two NixOS configuration files, <base_file> and <extra_file>,  as inputs 
and produces a new NixOS configuration file whose content is the combination of the two inputs.
If the same key is set by both <base_file> and <extra_file>,
then the value provided by <extra_file> takes precedence.

A NixOS configuration file, usually named `configuration.nix`, is written in the following form:

{ config, lib, pkgs, ... }: {
    key1 = value1;
    key2 = value2;
}
```

distro/tests/nix-commands/configuration.nix
```diff
@@ -0,0 +1,5 @@
+{ config, lib, pkgs, ... }:
```
tatetian: Rename test-specific configuration files to `extra_config.nix` to highlight the fact that it is not a complete `configuration.nix`.

distro/tests/add_test.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+# Usage: ./add_test.sh <test_name>
```
tatetian: As a potential user of this test framework, I hate to read or learn any new script or tool. Why not provide a template directory so that the user can simply `cp -r template <my_test>`? This removes the magic script and everything will be straightforward.

distro/tests/add_test.sh
```diff
@@ -0,0 +1,63 @@
...
+# Add your test steps here.
+
+finish_test
+EOF
+
+chmod +x "$ACTUAL_SCRIPT_FILE"
+
+# --- Create Symlink to the Test Script ---
+echo "Creating symlink: ${SYMLINK_PATH} -> ${SYMLINK_TARGET}"
+ln -sf "$SYMLINK_TARGET" "$SYMLINK_PATH"
```
tatetian: Please do not use symbolic links into our repo---unless you got a very, very strong argument. Symbolic links are unprecedented in our repo. It makes harder for people to know where the real files reside.

How about having a directory under `test-asterinas/` for each test?

```
overlays/test-asterinas/
    test-nix-commands/
        default.nix
        extra_config.nix
        run_test.sh
    test-podman/
        default.nix
        extra_config.nix
        run_test.sh
    common/
        test-framework.sh
        template/
            default.nix
            extra_config.nix
            run_test.sh
```

It brings several benefits:
1. Every test is a Nix package and can be included in the image independent, rather than one-or-nothing as part of the `test-asterinas` package.
2. Everything about a test is centralized under a directory. No symbolic links. Easy to read and easy to find.
3. To create a new test, simply `cp -r common/template my_test`.

test/nixos/common/merge_nixos_config.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+
+# SPDX-License-Identifier: MPL-2.0
+
+# Usage: merge_nixos_config.sh <base_file> <extra_file> <merged_file>
```
tatetian, line 5: The `etc_nixos` directory is intended to reflect the structure of `/etc/nixos` on an installed Asterinas NixOS system. If you put `merge_nixos_config.sh` under this directory, it implies that you want it to be copied to an installed Asterinas NixOS system. But I don't think doing so would be useful to anyone. At least judging from the current PR, it is only used by CI. So `tools/github_workflows/` seems to be a more appropriate home.

tools/github_workflows/config_for_publish_cachix.nix
```diff
@@ -0,0 +1,12 @@
+# This configuration file enables all packages and services that are modified
```
tatetian, line 1: Again, if a file is not useful in an installed Asterinas NixOS system, then it should not be put under `etc_nixos`. Find it a better home.

distro/etc_nixos/overlays/test-asterinas/test-framework.sh
```diff
@@ -0,0 +1,74 @@
...
+        echo "$output"
+        echo "----------------------"
+        exit 1
+    fi
+    echo "    Check PASSED"
+}
+
+# Runs a command where only the exit code matters.
+# Usage: run_command "my_command"
+run_command() {
```
tatetian: Let's rename this to `run_cmd` and do `run_and_expect` to `run_cmd_and_expect`. This way, people immediately get the relationship between `run_cmd` and `run_cmd_and_expect`.

distro/tests/README.md
```diff
@@ -0,0 +1,87 @@
...
+Run the `add_test.sh` script with the name of your new test. The name should be descriptive and use hyphens for separators.
+
+```sh
+./add_test.sh <my-test-name>
+```
+
+This command will perform the following actions:
+-   Create a new directory named `<my-test-name>/`.
+-   Create `my-test-name/configuration.nix`: The extra NixOS configuration for the test.
+-   Create `../etc_nixos/overlays/test-asterinas/test-my-test-name.sh`: The actual shell script that performs the test.
```
tatetian: Rename `test-asterinas` to `tests/` because it is no longer a package (see my previous comments), but only a directory.

distro/tests/add_test.sh
cchanging: Thanks for the suggestion! This makes lots of sense. My only concern is that `distro/etc_nixos/overlays/test-asterinas` is a relatively deep path, and users might be surprised that the distribution's tests are placed here, especially since they need to locate them under the `overlays` directory. 

Could this be a major issue? If so, aside from using symlinks, another solution is to place the `tests` in the `distro` directory and modify the install script to additionally track the `distro/tests` directory during installation. Alternatively, we could find an appropriate place in the book to explain where we've added the tests. What do you think?

cchanging: I have discussed with @StevenJiang1110 and we think the approach in #2822 might be a better solution. That is, all our tests could use this interactive method, allowing us to discard the `test-asterinas` package. 

Each test would essentially be an additional function within the interactive framework. The test framework directory could be placed in the upper layer, with each individual NixOS test only needing to generate an extra `extra_config.nix` and the interactive functions for testing. 

In addition, we also wouldn't need to rely on `qemu.log` for correctness verification, and in the future this framework can be used for other initramfs-based tests since currently they are also required to check `qemu.log`.

tatetian: `nix-editor` seems to be a new addition to our testing framework. It may very well remove the need for introducing the `merge_nix_config.sh` utility script.

In #2822, the test program is written in Rust to utilize the `rexpect` crate. But I doubt that anyone would want to write test programs in Rust... I suppose `pexpect` is more desirable. Although this means that we are introducing Python as another major programming language of Asterinas.

> Each test would essentially be an additional function within the interactive framework.

Do you mean that tests for Podman and tests for Nix commands should be written in the same test crate? I don't think so. They should be separate test crates, assuming we are writing tests in Rust. If `pexpect` is used, they should be different Python scripts.

tatetian: > My only concern is that `distro/etc_nixos/overlays/test-asterinas` is a relatively deep path, and users might be surprised that the distribution's tests are placed here, especially since they need to locate them under the `overlays` directory.

Maybe you can find some way to put test source code under `distro/tests` and let a `distro/etc_nixos/overlays/tests.nix` file to reference those Nix files under `distro/tests`.

cchanging: > nix-editor seems to be a new addition to our testing framework. It may very well remove the need for introducing the merge_nix_config.sh utility script.

`nix-editor` is an optional tool, but I'm not inclined to use it because: 1. It has poor documentation and hasn't been updated for a long time. 2. Using nix-editor requires users to modify each required configuration item one by one with its commands, which I don't think makes things easier for users. In contrast, although the merge approach introduces an additional script, it feels more natural for users as they only need to edit the `.nix` file.

> Do you mean that tests for Podman and tests for Nix commands should be written in the same test crate? I don't think so. They should be separate test crates, assuming we are writing tests in Rust. If pexpect is used, they should be different Python scripts.

For rust version, my current thought is organizing them in the following way (If using `merge_nixos_config.sh`):
```
aster-test
    |-- Cargo.toml
    `-- src
        |-- framework.rs // define `run_cmd` and `run_cmd_and_expect`, can also be used by initramfs-based tests
        |-- main.rs
        `-- nixos_tests
            |-- merge_nixos_config.sh
            |-- mod.rs // prepare the NixOS setup 
            |-- podman
            |   |-- extra_config.nix
            |   `-- mod.rs
            `-- template
                |-- extra_config.nix
                `-- mod.rs
```
For the python version, I think we can also organize the scripts like the above tree. I think both approaches are acceptable. If we use Rust, the framework implementation might be a bit more complex, but the content users need to add should be the same as with Python. If we go with Python, we might need to additionally install the `pexpect` package in the Docker container.

___

I gave a quick try to the test framework based on Rust's rexpect (honestly, writing in Rust feels simpler for me compared to Python). It follows the organizational structure as mentioned above, and this framework basically works. It can unify previous NixOS tests and interactive tests that require `-it` flags. Users adding new tests just need to copy the `template` and use the `run_cmd` and `run_cmd_and_expect` methods in their new test modules—not much different from writing tests in `.sh` scripts before. The overall usage of this test framework is simply `cargo run -- <testname>` to invoke the corresponding test.

However, a possible limitation discovered during quick testing with this new framework is that QEMU's output is no longer directly passed to the current terminal, so we can't see the full QEMU log like before. But we can still print the current command's output to the terminal when an `expect` fails, allowing users to compare and check, so I think this is acceptable.

In summary, I'm currently leaning towards using this type of interactive test framework to replace the current `test-asterinas` package. What do you think? If an interactive test framework is more desirable, regarding whether to use Rust or Python, I personally prefer the Rust version. First, the `rexpect` library doesn't seem to have any practical functional limitations compared to `pexpect`, and the user experience appears similar. Using the Rust version avoids introducing the Python language into our codebase, which would mean we might need to prepare code standards and formatting scripts for Python, which I think are unnecessary noise.

tatetian: > In contrast, although the merge approach introduces an additional script, it feels more natural for users as they only need to edit the `.nix` file.

Sure. We can keep the script, if you say so.

> aster-test
>     |-- Cargo.toml
>     `-- src
>         |-- framework.rs // define `run_cmd` and `run_cmd_and_expect`, can also be used by initramfs-based tests
>         |-- main.rs
>         `-- nixos_tests
>             |-- merge_nixos_config.sh
>             |-- mod.rs // prepare the NixOS setup 
>             |-- podman
>             |   |-- extra_config.nix
>             |   `-- mod.rs
>             `-- template
>                 |-- extra_config.nix
>                 `-- mod.rs

I have advised against the idea of having a single `test-asterinas` package.

> 1. Every test is a Nix package and can be included in the image independent, rather than one-or-nothing as part of the `test-asterinas` package.

Tests for Podman have nothing to do with those for Nix commands. We should be able to build and run tests for different applications _independently_. So I strongly feel that they should not be put under the same Nix package or Rust crate.

> cargo run -- <testname>

You may manage to run tests individually, but not to compile individually. Remember the technical debt that we have discussed about treating the entire set of tests under `test/apps/` as a single Nix package is undesirable. If they are a Nix package or Rust crate, a change to any test would cause the entire artifact to be changed.

The other aspects of the proposed code organization look alright.

> QEMU's output is no longer directly passed to the current terminal, so we can't see the full QEMU log like before

Yes, I think it is acceptable.

> I personally prefer the Rust version. First, the `rexpect` library doesn't seem to have any practical functional limitations compared to `pexpect`, and the user experience appears similar. Using the Rust version avoids introducing the Python language into our codebase, which would mean we might need to prepare code standards and formatting scripts for Python, which I think are unnecessary noise.

Ok. Let's start with Rust. If people complained about writing theses in Rust, we can reevaluate the pros and cons of Rust vs Python.

Last, since the new test programs run on the host, we don't need to build them as Nix packages. You might consider writing a Makefile.

tools/aster_test/nixos_tests/template/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "test-<test_name>"
+version = "0.1.0"
+edition = "2021"
```
tatetian: Why use Edition 2021 given the main project is using Edition 2024?

For all configurations in the manifist, use `{ workspace = true }` when possible. Apparently, this applies to `version` and `edition`. Even the local dependencies can do this.

tools/aster_test/nixos_tests/base/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+    process::Command,
+};
+
+use aster_test_framework::*;
+
+/// Runs a full NixOS test, including setup, build, and execution.
+///
+/// Users can specify a timeout for each command executed within the VM.
+/// If no timeout is provided, a default of 300,000 milliseconds (5 minutes) is used.
+pub fn run_nixos_test(
```
tatetian: The existence of this crate is questionable as it is using the wrong tool for the job. What can be done in a few lines of Bash scripts is implemented in hundreds line of Rust code--without clear benefits. There is no point of preparing config files and constructing the Makefile command in Rust. A script or Makefile can do the job and feed them to a test runner written in Rust with `rexpect`.

In addition, calling the root Makefile inside these test programs makes the two of them completely coupled, blurring the boundaries and responsibilities between them.

To address the shortcomings above, I suggest re-designing the interface of these test programs in Rust. Take the `podman` test program as an example. Its help message is shown below:

```
Drive a set of test cases for Podman by interacting with a QEMU VM running Asterinas NixOS.

Usage: ./drive-podman-tests [--test <test_name>] <qemu_cmd>

The <qemu_cmd> argument specifies the command to start the QEMU VM.
The user must ensure that
1. A Bash shell of the root user is ready to use after the OS finishes booting;
2. podman is pre-installed in Asterinas NixOS.

The <test_name> argument is optional.
If given, the test runner will only run the selected test case.
```

tools/aster_test/framework/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+        let res = self.run_cmd(command);
+        if res.is_err() {
+            self.runtime_sessions.pop_back();
+        }
+
+        res
+    }
+
+    /// Exits the current runtime session by running its exit command.
+    pub fn exit_current_session(&mut self) -> Result<(), Error> {
```
tatetian: Having `exit_current_session` and `exit_all` methods are unnecessary. Check out the following design, shown by this example:

```rust
fn podman_test(session: &mut Session) -> Result<()> {
    session.run_cmd_and_expect(
        "podman run --name=c1 docker.io/library/alpine ls /etc",
        "alpine-release",
    )?;

    session
        .new("podman run -it docker.io/library/alpine")
        .expected_prompt("/ #")
        .cmd_to_exit("exit");
        .enter(apline_test)?;
}

fn apline_test(session: &mut Session) -> Result<()> {
    session.run_cmd_and_expect("ls /etc/alpine-release", "/etc/alpine-release")?;
}
```

Thanks to the new API design, the test code is greatly simplified.

The `Session` struct implements the `Drop` trait to do cleanup automatically.

tools/aster_test/nixos_tests/nix-commands/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+    nixos_test_base::run_nixos_test("nix-commands", run_test, None)
+}
+
+fn run_test(runner: &mut Runner) -> Result<(), Box<dyn std::error::Error>> {
+    // Test nix-env install and run
+    runner.run_cmd("nix-env -iA nixos.hello")?;
+    runner.run_cmd_and_expect("hello", "Hello, world!")?;
+
+    // Test nix-env uninstall
+    runner.run_cmd("nix-env -e hello")?;
```
tatetian: A test framework should promote good practices for writing test cases. Think about Rust test harness (`#test`), gVisor test infra ([`gtest.h`](https://github.com/google/googletest/blob/7d7e750850c65099e49cc1a1aac94a79a914bba7/googletest/include/gtest/gtest.h#L219)), and even our own `test/src/apps/test.h` file. A good test framework should make it the purposes and structures of test cases obvious. In fact, the current framework does not even have the concept of _test cases_.

With that said, I may be asking too much for a single PR. We can address this issue in a follow-up PR.

tools/aster_test/README.md
```diff
@@ -0,0 +1,208 @@
+# Asterinas Test Framework
```
tatetian: This title is inappropriate. Not all test cases (those under `test/`) are governed by this test framework. In addition, it is unclear why the tile, the crate name, and the directory name wants to emphasize Asterinas. Isn't everything under the repo for Asterinas?

This brings me to another topic: the relationship between the `test/` directory and `tools/aster_test/`. Both directories are for tests. Since `test/` is a root-level directory, people naturally expect that they can find all test-related stuff there. So putting the NixOS tests under `tools/aster_test/` would be a surprise.

To address this issue, I propose re-organizing all tests in the following structure:

```
test/
    README.md
    initramfs/              # Tests run on initramfs (the original test/)
        Makefile            # Used by the root Makefile
        README.md
        src/
        nix/
    nixos/                  # Tests run on NixOS (the new tests)
        Makefile            # Used by the root Makefile
        README.md
        Cargo.toml          # Workspace manifest
        common/
            framework/      # Test framework crate
            template/       # A template directory for new tests
            merge_nixos_config.sh
        tests/
            podman/
            nix/
            ...
```

tatetian (commented):
Introducing the new `rexpect`-base test framework crate is a great step forward.

tools/aster_test/framework/src/lib.rs
cchanging: `exit_all` can indeed be removed. However, `exit_current_session` represents an exit behavior, which is actually a step in the testing process. For example, when debugging the `-it` flag before, we also encountered situations where it couldn't exit properly. If we perform the exit in `drop`, I feel that on one hand, users cannot explicitly perceive this as a test statement, and on the other hand, `drop` cannot return error to the upper layer. I'm not sure if this approach is worthwhile.

tatetian: > users cannot explicitly perceive this as a test statement

Yes, sometimes we might want to test if the "exit" command for a session works or not. But you must agree that our tests will spawn many, many sessions of a particular kind to run different commands or functionalities. The ability to start and end a session is the most basic one and implicitly tested by all these test cases. And do you think our developers will be happy with writing out the exit command explicitly for every session, even though he does not care about testing exit?

cchanging: > writing out the exit command explicitly for every session

In fact, if users do not need to exit the session and test additional operations after completing the internal operations within the session, they do not need to be concerned about `exit` or explicitly call `exit_current_session()`. The framework will call `exit_all()` to exit all current sessions one by one. In the `-it` test, I explicitly called `exit_current_session()` because I knew that `-it` might fail to exit due to missing functionalities, so I made that test command explicit.  

For cases where users need to exit the session after completing internal operations and then test additional operations—and there are many such test cases—users can also encapsulate them with functions to avoid repeatedly calling `exit_current_session()`.  

As for the situation below, I am not sure if the new API can work smoothly because the `rexpect::PtySession` relied upon in `Session` cannot be cloned. I am uncertain about how to more elegantly preserve the old session after entering a new one.

> does not care about testing exit?

If it is implemented in `Drop`, as I mentioned above, `Drop` cannot return errors to the higher level when the test fails. In that case, how can the framework know that the `exit` failed? Using global variables or `catch_unwind`? I can't think of an elegant solution at the moment. I believe if `exit` may fail, it should be verified. Just like in our app tests, even for every operation like "close(fd)", we explicitly use `TEST_SUCC` to verify it.

tatetian: ## Design concerns

First of all, the very existence of `exit_current_session` makes the user wonder why it is needed and when to use it. **The best code is no code; the best API is no API.** Removing the API makes the interface more clear.

> In the `-it` test, I explicitly called `exit_current_session()` because I knew that `-it` might fail to exit due to missing functionalities, so I made that test command explicit.

Second, this argument is only partial true. The design of this API requires the user to call it explicitly. Consider the follow scenario:

```rust
    // Test run podman with `-it`
    runner.run_cmd_and_enter_new_session(
        "podman run -it docker.io/library/alpine",
        "/ #".to_string(),
        "exit".to_string(),
    )?;

    // Run command inside the alpine container session
    runner.run_cmd_and_expect("ls /etc/alpine-release", "/etc/alpine-release")?;

    // Exit the alpine container session
    runner.exit_current_session()?;

    // More commands after exiting the Docker session
    runner.run_cmd("podmam ls")?;
```

Because more Bash commands are needed after the container session, the user has no choice but to invoke `exit_current_session`.

Third, the relationship between sessions are _nested_, but the current test code using the current API is _flat_. A human needs to pay special attention to `run_cmd_and_enter_new_session` and `exit_current_session` to identify the scope of a session.

In contrast, **the new API design uses the scopes of functions or closures to mark those of sessions _naturally_**, improving code readability.

Compare this

```rust
fn podman_test(bash_session: &mut Session) -> Result<()> {
    bash_session.run_cmd_and_expect(
        "podman run --name=c1 docker.io/library/alpine ls /etc",
        "alpine-release",
    )?;

    bash_session
        .new_session_cmd("podman run -it docker.io/library/alpine")
        .expected_prompt("/ #")
        .cmd_to_exit("exit");
        .enter(|podman_session| {
            podman_session.run_cmd_and_expect("ls /etc/alpine-release", "/etc/alpine-release")?;    
        })?;
}
```

with that

```rust
fn podman_test(session: &mut Session) -> Result<()> {
    session.run_cmd_and_expect(
        "podman run --name=c1 docker.io/library/alpine ls /etc",
        "alpine-release",
    )?;

    runner.run_cmd_and_enter_new_session(
        "podman run -it docker.io/library/alpine",
        "/ #".to_string(),
        "exit".to_string(),
    )?;

    session.run_cmd_and_expect("ls /etc/alpine-release", "/etc/alpine-release")?;
}
```

It is apparent that the former one using the new API does a better job on distinguishing between different sessions. In other words, **the new API provides a structure to the test code, a huge win for readability**. As to the old API, if I am new to the framework and the test code, then I could easily miss the important fact that a new session is started.

## Implementation concerns

> I am not sure if the new API can work smoothly because the `rexpect::PtySession` relied upon in `Session` cannot be cloned.

> If it is implemented in `Drop`, as I mentioned above, `Drop` cannot return errors to the higher level when the test fails. In that case, how can the framework know that the `exit` failed?

These are some superficial arguments.

None of these implementation-level obstacles are intractable to me. With your ability and expertise, I don't think that they are really a problem to you. And I don't think that you have tried hard enough to work them out.

As engineers, we leave difficulties to ourselves, convenience to users. Our company value is **customer first**. Now our customers here are the API users. So we should not use implementation difficulties or inconvenience as excuses. In the other day, you shared your personal growth in 2025, saying that you learned a hard lesson that **when making a design, one should drive the design by studying the true needs of users, rather than considering technical constraints or ability limitations of developers.** Now is the good time to put this principle in practice.

test/nixos/tests/podman/src/main.rs
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use nixos_test_framework::*;
+use nixos_test_macro::nixos_test;
+
+nixos_test_main!();
+
+#[nixos_test]
+fn alpine_container_basic(session: &mut Session) -> Result<(), Error> {
```
tatetian: To further improve readability, I propose renaming `session: &mut Session` to `nixos_shell: &mut Session`. The type of this variable has been shown in the type location. So using name to pass extra info would be good. Intuitively, the developer can view this variable as a shell, rather than a more abstract concept of session. I pick the word "shell" rather than "bash" because I think bash is too specific. It just happens that Bash is chosen as the default shell by NixOS. We do not want to encourage the test developers to use Bash-specific features.

This way, the test code would all look like:

```rust
    nixos_shell.run_cmd_and_expect(cmd1, expected_output1);

    nixos_shell.run_cmd_and_expect(cmd2, expected_output2);
```

The code is extremely readable as one can read it as plain English and get its intention.

The alpine test code would look like:

```rust
#[nixos_test]
fn alpine_interactive_session(nixos_shell: &mut Session) -> Result<(), Error> {
    let container_session_desc = SessionDesc::new()
        .expect_prompt("/ #")
        .cmd_to_enter("podman run -it docker.1ms.run/alpine")
        .cmd_to_exit("exit");

    nixos_shell.enter_session_and_run(container_session_desc, |alpine_shell| {
        alpine_shell.run_cmd_and_expect("ls /etc/alpine-release", "/etc/alpine-release")?;
        Ok(())
    })?;

    Ok(())
}
```

test/nixos/common/template/src/main.rs
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use nixos_test_framework::*;
+use nixos_test_macro::nixos_test;
```
tatetian: Let `nixos_test_framework` re-export `nixos_test_macro::nixos_test` so that the user crate only needs to do an one-line import, `use nixos_test_framework::*;`

test/nixos/common/template/src/main.rs
```diff
@@ -0,0 +1,13 @@
...
+
+nixos_test_main!();
+
+#[nixos_test]
+fn my_test(session: &mut Session) -> Result<(), Error> {
+    // Add your test logic here
```
tatetian: Make the example test case a "Hello World":

```rust
    session.run_cmd("echo 'Hello, World!' > out.txt")?;
    session.run_cmd_and_expect("ls out.txt", "out.txt")?;
```

which is more useful than an empty one.

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
...
+    environment.systemPackages = with pkgs; [
+        # Add required packages here
+        vim
+        git
+    ];
+    
+    # Configure system services
+    virtualisation.podman.enable = true;
+}
+```
```
tatetian, line 84: Add the following paragraph below the code snippet:

```markdown
This content of this file will be merged with the [default configuration file](../../distro/etc_nixos/configuration.nix) to generate the final configuration file, i.e., `configuration.nix`, for the testing Asterinas NixOS system.
```

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
...
+```bash
+# Build the image once
+make build_nixos_test NIXOS_TEST_NAME=my-test
+
+# Run tests multiple times without rebuilding
+make run_nixos_test NIXOS_TEST_NAME=my-test
+make run_nixos_test NIXOS_TEST_NAME=my-test TEST_CASE_NAME=specific_test
+```
+
+## Session API Reference
```
tatetian: This is the job of the rendered Rust doc. I don't think repeating it manually in the README is a good idea. Let's focus on improving the Rust doc of the framework. The crate is quite small; people can skim the code as well as its doc in the source code. One day, we may render the Rust docs of all kernel crates and upload it to a website so that people can easily read those Rust docs.

test/initramfs/README.md
```diff
@@ -0,0 +1,84 @@
...
+Most tests in this directory are compiled and packaged using [Nix](https://nixos.org/), a powerful package manager. This ensures consistency and reproducibility across environments.
+
+> **Note**: If you are adding a new test to the `apps` directory, ensure that it supports multiple architectures. Some of the existing apps lack proper architecture-specific handling.
+
+### Syscall Test Suite - Gvisor Exception
+
+While most tests rely on `Nix` for compilation, the `gvisor` syscall test suite currently cannot be built with `Nix`. Instead, the `gvisor` tests are compiled in the Docker image. For details, refer to `tools/docker/Dockerfile`.
+
+### Multi-Architecture Support
+The test suite supports building for multiple architectures, including `x86_64` and `riscv64`. You can specify the desired architecture by running:
```
tatetian, line 37: Insert a blank line before the paragraph.

test/initramfs/README.md
```diff
@@ -0,0 +1,84 @@
...
+├── nix/
+│   ├── benchmark/     # Nix expressions for `benchmark`
+│   ├── syscall/       # Nix expressions for `syscall`
+│   ├── apps.nix       # Nix expression for `apps`
+│   └── initramfs.nix  # Nix expression for packaging initramfs
+├── Makefile
+└── README.md
+```
+
+---
```
tatetian: Remove the inappropriate page break here

test/initramfs/README.md
```diff
@@ -0,0 +1,84 @@
+# Initramfs Test Suite Overview
```
tatetian:
```markdown
# Initramfs-based Test Suites
```

Next time, use `git mv` to move the original file into a new location. Moving content from an old file to a new one by copying would mislead Git, which is exactly the case here. What is changed in this file in comparison to the old one is not clear to me.

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
...
+│   ├── podman/                 # Example: Podman container test
+│   │   ├── Cargo.toml
+│   │   ├── extra_config.nix    # (Optional) Additional NixOS configuration
+│   │   └── src/
+│   │       └── main.rs
+│   └── ...                     # Other tests
+└── Makefile                    # Test orchestration
+```
+
+## Creating a New Test
```
tatetian, line 22: Good job! This section is well written. A future test developer will highly appreciate your guidance.

test/nixos/common/framework/src/lib.rs
```diff
@@ -0,0 +1,141 @@
+// SPDX-License-Identifier: MPL-2.0
+
```
tatetian, line 2: Add the crate-level Rust doc.

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
...
+│   │   └── src/
+│   │       └── lib.rs
+│   ├── template/               # Template for creating new tests
+│   │   ├── Cargo.toml
+│   │   ├── extra_config.nix
+│   │   └── src/
+│   │       └── main.rs
+│   └── merge_nixos_config.sh   # Helper script for merging NixOS configurations
+├── tests/
+│   ├── podman/                 # Example: Podman container test
```
tatetian: This is not an "example". This is a real test.

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
...
+## Directory Structure
+
+```
+test/nixos/
+├── common/
+│   ├── framework/              # Core testing library
+│   │   ├── Cargo.toml
+│   │   └── src/
+│   │       ├── lib.rs          # Main framework entry
+│   │       └── session.rs      # Session management
```
tatetian: Too many details are disclosed. You are not doing a reader a favor by overwhelming the reader's mind with all the whole structure of this directory. Focus on _where_ to find _what_ at a high level.

test/nixos/Makefile
```diff
@@ -0,0 +1,101 @@
...
+	@if [ "$(TEST_ISO)" = "true" ]; then \
+        echo "    Building ISO installer..."; \
+        $(MAKE) --no-print-directory -C "$(ASTERINAS_ROOT)" iso; \
+        if [ $$? -ne 0 ]; then \
+            echo "Error: ISO build failed"; \
+            $(MAKE) restore; \
+            exit 1; \
+        fi; \
+        echo "    Running ISO installation..."; \
+        $(MAKE) --no-print-directory -C "$(ASTERINAS_ROOT)" run_iso; \
```
tatetian: The root Makefile calls this Makefile and then this Makefile calls the root Makefile... Can we break the cyclic dependency?

Makefile
```diff
@@ -401,6 +399,25 @@ ktest: initramfs $(CARGO_OSDK)
 		rm -r target/osdk/*; \
 	done
 
+.PHONY: build_nixos_test
+build_nixos_test:
```
tatetian: This name breaks the current convention. Note the established convention in this Makefile to name all targets for building "xxx" as "xxx". For example, the build target for "kernel" is "kernel", "nixos" is "nixos". Their names do not start with "build".

In addition, I am wondering if it is possible to combine the three newly-added targets with the existing targets of `nixos` and `run_nixos`. Note that we do not add a `test_initramfs` or "test_kernel" target to run the old initramfs-based tests.

tatetian (commented):
Thank for the revision! The new version looks much, much better. I am excited to see this upcoming NixOS test framework.

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
+# NixOS Test Suite Overview
```
tatetian:
```markdown
# NixOS-Based Test Suites
```

test/nixos/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+This directory contains a comprehensive test suite for Asterinas running in NixOS environments within QEMU virtual machines. Built on top of [rexpect](https://github.com/rust-cli/rexpect), it provides a declarative testing framework using procedural macros for writing system-level integration tests.
```
tatetian: This paragraph needs rework. Think again: what you _really_ want to tell the reader in the very first paragraph. Why we bother the reader with technical detail such as "rexpect" and "procedural macros". Think in the shoes of the reader, not yourself, the author.

test/nixos/Makefile
```diff
@@ -0,0 +1,101 @@
+# SPDX-License-Identifier: MPL-2.0
+
+NIXOS_TEST_NAME ?=
+TEST_CASE_NAME ?=
```
tatetian: First of all, give them unified prefixes to make the scope of influence apparent.

Second, I don't think "test name" and "test case name" are very distinguishing. Use "test suite" and "test case" instead. Use the term "test suite" and "test case" consistently in the README.

Third, is it possible to put time unit in the timeout variable? So that people can set `TIMEOUT=1min`, `TIMEOUT=5s`, or `TIMEOUT=100ms`.

So the variables should be:

```Makefile
NIXOS_TEST_SUITE ?=
NIXOS_TEST_CASE ?= 
NIXOS_TEST_TIMEOUT ?= 5min
```

test/nixos/common/template/src/main.rs
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MPL-2.0
+
```
tatetian, line 2: Add the following Rust doc to the template crate and all test crates.

```rust
//! The test suite for <TargetAppName> on Asterinas NixOS.
//!
//! # Document maintenance
//!
//! An application's test suite and its "Verified Usage" section in Asterinas Book
//! should always be kept in sync.
//! So whenever you modify the test suite,
//! review the documentation and see if should be updated accordingly.
```

test/nixos/common/framework/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+macro_rules! nixos_test_main {
+    () => {
+        fn main() -> Result<(), Box<dyn std::error::Error>> {
+            $crate::__nixos_test_main()
+        }
+    };
+}
+
+#[doc(hidden)]
+pub fn __nixos_test_main() -> Result<(), Box<dyn std::error::Error>> {
```
tatetian, line 52: Display the help message with `--help`.

test/README.md
```diff
@@ -1,84 +1,20 @@
...
-│       └── gvisor/    # Gvisor syscall test suite
-├── nix/
-│   ├── benchmark/     # Nix expressions for `benchmark`
-│   ├── syscall/       # Nix expressions for `syscall`
-│   ├── apps.nix       # Nix expression for `apps`
-│   └── initramfs.nix  # Nix expression for packaging initramfs
-├── Makefile
-└── README.md
-```
+### Initramfs Tests ([`initramfs/`](initramfs/))
```
tatetian:
```md
### Initramfs-Based Tests
```

test/README.md
```diff
@@ -1,84 +1,20 @@
...
+Tests running in a minimal initramfs environment. Best for:
+- System call validation
+- Core functionality testing
+- Performance benchmarks
 
-## Building and Packaging Tests
+See [`initramfs/README.md`](initramfs/README.md) for details.
 
-Most tests in this directory are compiled and packaged using [Nix](https://nixos.org/), a powerful package manager. This ensures consistency and reproducibility across environments.
+### NixOS Tests ([`nixos/`](nixos/))
```
tatetian:
```md
### NixOS-Based Tests
```

test/initramfs/README.md
```diff
@@ -0,0 +1,83 @@
+# Initramfs-based Test Suites
```
tatetian: base --> Based

test/nixos/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+repository = "https://github.com/asterinas/asterinas"
+license = "MPL-2.0"
+edition = "2024"
+
+[workspace.dependencies]
+inventory = "0.3"
+rexpect = "0.6"
+strip-ansi-escapes = "0.2"
+nixos-test-framework = {path = "common/framework"}
+nixos-test-macro = {path = "common/test_macro"}
```
tatetian: Missing whitespaces in the braces.

test/nixos/Makefile
```diff
@@ -0,0 +1,108 @@
...
+            TEST_CASE_ARG="--test $(NIXOS_TEST_CASE)"; \
+        fi; \
+        QEMU_CMD="bash $(RUN_SCRIPT) nixos"; \
+        cargo run -- --qemu-cmd "$$QEMU_CMD" $$TEST_CASE_ARG || \
+        exit 1
+
+.PHONY: restore
+restore:
+	@if [ -f "$(BACKUP_CONFIG)" ]; then \
+        mv "$(BACKUP_CONFIG)" "$(BASE_CONFIG)"; \
```
tatetian: I think this is very unreliable and ugly. What if something goes wrong during the Makefile, which  causes `make` to end abruptly, failing to run the `restore` target? Could you think of a way that does not modify the version-controlled `configuration.nix` file?

Note that for the `nixo` target, updating the `configuration.nix` in-place can be easily avoided:

```
${DISTRO_DIR}/result/bin/install_aster_nixos.sh --config ${CONFIG_PATH} --disk ${DISK}
```

The `install_aster_nixos.sh` tool allows specifying the path for the `configuration.nix` file. So you only need to solve the other half of the puzzle, the `iso` target. And **please do try hard to figure out a _clean_ solution**.

test/nixos/README.md
```diff
@@ -0,0 +1,150 @@
+# NixOS-Based Test Suites
+
+This directory contains NixOS-based tests and a framework for writing and running them. The framework executes tests by interacting with a live instance of the operating system in a virtual environment. Thanks to this interactive design, the framework can test virtually any behavior that a real user could trigger through a terminal. It also offers a simple, declarative API, making it easy to write and maintain these interactive test scenarios.
```
tatetian: This is not what one normally would call a _declarative_ approach. Think of SQL (a SQL query describes what the records you want, not how the DB retrieves them), Kubernetes (a YAML file is used to describe what state you want a cluster to be, not how K8S achieves it), and Nix (the `configuration.nix` file describes what the desired system state is, not how to Nix gets there). Writing a series of `run_cmd_xxx` statements in our test framework should still be considered an _imperative_ approach, just like other common test frameworks.

test/nixos/README.md
tatetian: This one is not resolved, yet.

Although some comments are removed, the tree structure still shows _all_ files under `common/`. This means that you are disclosing implementation detail to the users of the test framework, who are most interested in how to use it, not how it is implemented. As a user of the test framework, why do I need to be aware of the existance of `merge_nixos_config.sh`?

Disclose select files under `common/`. Do not burden the mind of a first-time reader with implementation detail. The documentation helps _people use_ it, not educates people how to _you implement_ it.

test/nixos/README.md
```diff
@@ -0,0 +1,150 @@
...
+    
+    # Configure system services
+    virtualisation.podman.enable = true;
+}
+```
+This content of this file will be merged with the [default configuration file](../../distro/etc_nixos/configuration.nix) to generate the final configuration file for the testing Asterinas NixOS system.
+
+## Running Tests
+
+It is recommended to run commands from the **Asterinas root directory**.
```
tatetian: Do not write in vague terms. What exactly does "recommended" means here? Can one run these commands in another directory or not? If not, the word should be "required". If yes, what are the other allowed directories?

If they can only work under the root, then write:

> The following commands should be run under the project root.

If they can also work under `test/nixos`, rewrite it:

> The following commands should be run under the project root or the `test/nixos/` directory.

tatetian (commented):
This should be the last round of revision. Cheers.

test/nixos/README.md
cchanging: > The following commands should be run under the project root.

I adopted this modification. 

Users could actually run the tests in the `/test/nixos` directory, but it might require slightly more complex commands because the Makefile in `/test/nixos` lacks the `run_iso` target compared to the root directory's Makefile, and the corresponding targets are missing declarations for `OVMF` environment variables. 

I think it makes more sense to directly require users to run commands under the root. First, I don't think it's necessary to further complicate the `/test/nixos` Makefile to allow users to run tests with the same commands in `/test/nixos`, as I can't think of any reason why users would prefer running tests in the `/test/nixos` directory. Second, I also don't think it's necessary to inform users about how they could run tests in `/test/nixos` directory with "slightly more complex" operations, as that would just add unnecessary burden on users.

tatetian: Yes, I totally agree with your reasoning.

PR #2775, opened 2025-12-20
Fix atomic violation on caps_lock

zjp-CN:
Updating the caps_lock state through a combination of load and store is an atomic violation, because individual atomic operations don't form an atomic sequence when executed consecutively.

The bug is detected by lockbud. cc https://github.com/os-checker/bugs-found/issues/3

Also see the Gemini's [explanation](https://aistudio.google.com/app/prompts?state=%7B%22ids%22:%5B%221JkqMQJccTq7rlYNusmovLl9krACM9aZd%22%5D,%22action%22:%22open%22,%22userId%22:%22107770117539303321065%22,%22resourceKeys%22:%7B%7D%7D&usp=sharing).

### DIFF
--- a/kernel/comps/framebuffer/src/console_input.rs
+++ b/kernel/comps/framebuffer/src/console_input.rs
@@ -213,8 +213,7 @@ impl FbConsoleHandler {
             }
             KeyCode::CapsLock => {
                 if key_status == KeyStatus::Pressed {
-                    let new_caps = !self.caps_lock.load(Ordering::Relaxed);
-                    self.caps_lock.store(new_caps, Ordering::Relaxed);
+                    self.caps_lock.fetch_xor(true, Ordering::Relaxed);
                 }
                 return;
             }


zjp-CN:
(Irrelevant to the PR.)

There is a failing [workflow](https://github.com/asterinas/asterinas/actions/runs/20390267607/job/58598423087?pr=2775) saying

```
Creating image at /root/asterinas/target/nixos/asterinas.img of size 6144MB......
dd: error writing '/root/asterinas/target/nixos/asterinas.img': No space left on device
```

After checking some histories, it rarely happens. But if the disk shortage is a problem, it can be solved by storing the img to a larger disk like `/mnt` on Github action runners.

```
$ df -alh
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   54G   19G  75% /
...
/dev/sda1        74G  4.1G   66G   6% /mnt
```

lrh2000:
> Updating the caps_lock state through a combination of load and store is an atomic violation, because individual atomic operations don't form an atomic sequence when executed consecutively.

Yes, but do we really need an atomic sequence here? Currently, we assume that the second key cannot be processed until the first one is finished processing. Specifically, the PS/2 controller (or the IRQ controller) will not raise the second IRQ until the first one has been processed.

This code was originally in the PS/2 driver, so I believe it is correct (under our current assumption). However, now that it has been moved to the framebuffer console, the above reasoning is not quite obvious.

> cc https://github.com/os-checker/bugs-found/issues/3

I assume many of the atomic "bugs" are similar. If they're under some locks, using `load + store` is fine and preferred because it's faster. For example, if you use `fetch_add`, it requires `lock; xadd` in x86; the `lock` prefix and fetch-and-modify operation are very slow.

For example,
```
{
  "bug_kind": "AtomicityViolation",
  "possibility": "Possibly",
  "diagnosis": {
    "fn_name": "process::status::StopStatus::stop",
    "atomic_reader": "kernel/src/process/status.rs:116:12: 116:51 (#0)",
    "atomic_writer": "kernel/src/process/status.rs:119:13: 119:59 (#0)",
    "dep_kind": "Control"
  },
  "explanation": "atomic::store is data/control dependent on atomic::load"
}
```
```
{
  "bug_kind": "AtomicityViolation",
  "possibility": "Possibly",
  "diagnosis": {
    "fn_name": "aster_bigtcp::iface::BoundPort::<E>::set_can_reuse",
    "atomic_reader": "/mnt/check/batch_1/asterinas/asterinas/kernel/libs/aster-bigtcp/src/iface/common.rs:306:12: 306:50 (#0)",
    "atomic_writer": "/mnt/check/batch_1/asterinas/asterinas/kernel/libs/aster-bigtcp/src/iface/common.rs:318:9: 318:59 (#0)",
    "dep_kind": "Control"
  },
  "explanation": "atomic::store is data/control dependent on atomic::load"
}
```

---

Implementation-wise:
 - Please use [`fetch_xor`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#method.fetch_xor) instead of `fetch_update`.
 - For FIXMEs, please write `FIXME: Use ...` instead of `FIXME: use ...`.
 - For commit messages, please write `Fix atomic ...` instead of `fix atomic ...`.

zjp-CN:
> do we really need an atomic sequence here?

Or are these bool states on keystrokes necessary to be atomic? Plain bool operation is faster than that of atomic bools.

> I assume many of the atomic "bugs" are similar. If they're under some locks, using `load + store` is fine and preferred because it's faster.

Yeah, they are false postives.

lrh2000:
> Or are these bool states on keystrokes necessary to be atomic?

How can you safely mutate a plain bool in Rust? Rust requires either an atomic variable or a lock.

Even in unsafe Rust, data races are UB, so we'll have to recreate atomics or locks.

> Plain bool operation is faster than that of atomic bools.

In general, it may be true. But for x86, they're equally fast (from the instruction level, i.e., without considering compiler optimizations).

zjp-CN:
Is fetch_xor accepatble here? If not, I can close this PR.

lrh2000:
> Is fetch_xor accepatble here? If not, I can close this PR.

Acceptable. I'm okay with merging the `fetch_xor` version because:

(1) As mentioned earlier,
>> However, now that it has been moved to the framebuffer console, **the above reasoning is not quite obvious.**

(2) Performance is not an issue here. Besides, we're going to take a lock for non-special keys anyway, and it's heavier than the `fetch_xor` operation for special keys.
https://github.com/asterinas/asterinas/blob/98c94969c5d889d17b9c2be47da0d05c6d7e066f/kernel/comps/framebuffer/src/console_input.rs#L233-L237

zjp-CN:
Thank you for the clarification.

lrh2000:
> it has been moved to the framebuffer console

I just realized that there may be ways to trigger this "race condition" in the future. For instance, if we support [input event injection via evdev](https://elixir.bootlin.com/linux/v6.15/source/drivers/input/evdev.c#L493-L536), concurrent keystrokes could originate from real devices and event injection. The same would be true if we supported two keyboards, such as USB and PS/2.

Currently, we don't support either of these options. However, these possibilities are still a good justification for the change. In fact, it would be stronger than all previous discussions.

kernel/comps/framebuffer/src/console_input.rs
```diff
@@ -213,8 +213,7 @@ impl FbConsoleHandler {
             }
             KeyCode::CapsLock => {
                 if key_status == KeyStatus::Pressed {
-                    let new_caps = !self.caps_lock.load(Ordering::Relaxed);
-                    self.caps_lock.store(new_caps, Ordering::Relaxed);
+                    _ = self.caps_lock.fetch_xor(true, Ordering::Relaxed);
```
lrh2000: This is actually the first time I've learned that `_ = ...` will compile. Thanks for letting me know. However, I prefer the syntax `let _ = ` since it's more widely adopted in our codebase and the Rust ecosystem.

Furthermore, I don't think it's necessary or helpful to explicitly ignore the results of the `fetch_xxx` methods. I can think of two cases where we would want to explicitly ignore the results:
 1) To suppress compiler warnings caused by the `#[must_use]` attribute.
 2) The result will almost always be used, so not using the result is the exception. If we can fix the function signature, it would be better to fall back to the first case by adding `#[must_use]`.

However, I feel it's normal for the results to be ignored for `fetch_xxx`. Based on [the search results from our codebase](https://github.com/search?q=repo%3Aasterinas%2Fasterinas%20fetch_add&type=code), I estimate that about half of the use cases simply ignore the results. Then I feel that there is no need to use `let _ =` or `_ = ` here.

zjp-CN: Thanks! Done.

zjp-CN:
Grateful for the potential case! We have been thinking about such two strokes from keybord in the past weeks. 😃

tatetian:
> After checking some histories, it rarely happens. But if the disk shortage is a problem, it can be solved by storing the img to a larger disk like `/mnt` on Github action runners.
> 
> ```
> $ df -alh
> Filesystem      Size  Used Avail Use% Mounted on
> /dev/root        72G   54G   19G  75% /
> ...
> /dev/sda1        74G  4.1G   66G   6% /mnt
> ```

@cqs21 Could you confirm this? The disk space problem of Github runners is troubling us recently. Have we overlooked the fact that `/mnt` is usable on Github runners? If so, how much disk space is there under `/mnt` for free and paid Github runners, respectively?

PR #2766, opened 2025-12-17
Fix many wrong error codes and other buggy behavior in various MM syscalls

lrh2000:
~~I'd like to wait until https://github.com/asterinas/asterinas/pull/2744 is merged and rebase this PR (see https://github.com/asterinas/asterinas/pull/2744#issuecomment-3656190723).~~

---

As with all my previous PRs that fix error codes, this one fixes them for MM-related system calls, i.e., `mmap`, `mremap` `msync`, `madvise`, `mprotect`, and `munmap`.

Note that incorrect error codes are the most typical issue, but they won't be the only one. I also fix many other problems, see the detailed explanation below.

#### Fix error codes in mmap

 - An address below `VMAR_LOWEST_ADDR` should result in `EPERM`.
 - An unrecognized flag with `MAP_SHARED_VALIDATED` should result in `EOPNOTSUPP`.
 - `MAP_SHARED_VALIDATED` implies `MAP_SHARED`.
 - `MAP_FIXED_NOREPLACE` implies `can_overwrite(false)`.
 - ...

#### Drop `MAP_SYNC` support

There is no need to support this flag because https://man7.org/linux/man-pages/man2/mmap.2.html says:
>  This flag is supported only for files supporting DAX (direct mapping of persistent memory).  For other files, creating a mapping with this flag results in an EOPNOTSUPP error.

#### Add `is_userspace_vaddr_range`

It seems that we repeat the following logics several times. Let's define a method instead.

https://github.com/asterinas/asterinas/blob/f30bd698d1b3d91452b3dd4311c8d15ebb4caeb8/kernel/src/vm/vmar/mod.rs#L941-L946

#### Fix error codes in MM syscalls

`msync`, `madvise`, `mprotect`, and `munmap` all accept a `vaddr` and a `len`. They all need to align `len` to a page boundary and add it to `vaddr` to get the end address. These steps are prone to integer overflow, and the implementation is inconsistent (some approaches are incorrect and can lead to panic). Fix #2779.

This commit unifies them to a consistent style:
```rust
    if !addr.is_multiple_of(PAGE_SIZE) {
        return_errno_with_message!(Errno::EINVAL, "the mapping address is not aligned");
    }
    if len == 0 {
        return Ok(SyscallReturn::Return(0));
    }
    if VMAR_CAP_ADDR.checked_sub(addr).is_none_or(|gap| gap < len) {
        // FIXME: Linux returns `EINVAL` if `(addr + len).align_up(PAGE_SIZE)` overflows. Here, we
[...]

PR #2777, opened 2025-12-21
Fix buggy behavior when trying to load corrupted ELF files

lrh2000:
This PR resolves several issues related to loading corrupted ELF files.

For example,
 - We should not fail an assertion simply because the ELF header is not valid.
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/process/program_loader/elf/elf_file.rs#L215
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/process/program_loader/elf/elf_file.rs#L220
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/process/program_loader/elf/elf_file.rs#L33-L35
 - The memory and file ranges are untrusted. Therefore, it may overflow.
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/process/program_loader/elf/elf_file.rs#L77
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/process/program_loader/elf/elf_file.rs#L80
 - Some error codes are wrong.
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/process/program_loader/elf/elf_file.rs#L106-L109
 - ...

### Changes in `kernel/src/vm/vmar/mod.rs`

Changes to `kernel/src/vm/vmar/mod.rs` are better explained by https://github.com/asterinas/asterinas/pull/2766. This PR only exposes `VMAR_CAP_ADDR` to make the code compile. #2766 also exposes `VMAR_CAP_ADDR` for other use cases and removes the duplication of the `userspace_addr()` method.

### About the last commit: "Separate VMAR load and activation"

We can roughly divide the `execve()` system call into three stages:
 1. Load and verify the ELF headers of the executable
 2. Load and verify the ELF headers of the interpreter
 3. Load the ELF segments as virtual address maps

Meanwhile,
 - In Linux, a point of no return (see https://elixir.bootlin.com/linux/v6.18/source/fs/binfmt_elf.c#L1010 and https://elixir.bootlin.com/linux/v6.18/source/fs/exec.c#L1117-L1120) occurs between *Step 2* and *Step 3*. So if virtual address maps cannot be created (which happens in *Step 3*), the process will be terminated by SIGSEGV and the error code can only be retrieved through `ptrace()`.
 - In Asterinas, the point of no return occurs between *Step 1* and *Step 2* before this PR. This means that error codes in *Step 2* and *Step 3* won't be visible to the user because the process has been terminated.

Having the same point of no return as Linux would require additional refactoring. However, I found that it is quite easy for us to defer VMAR activation, allowing us to have the point of no return after *Step 3*. This means that all error codes will be visible to the user and can be properly tested.

### DIFF
--- a/kernel/src/process/execve.rs
+++ b/kernel/src/process/execve.rs
@@ -8,16 +8,14 @@ use ostd::{
     user::UserContextApi,
 };
 
+use super::process_vm::activate_vmar;
 use crate::{
-    fs::{
-        fs_resolver::{FsResolver, PathOrInode},
-        utils::Inode,
-    },
+    fs::{fs_resolver::PathOrInode, utils::Inode},
     prelude::*,
     process::{
         ContextUnshareAdminApi, Credentials, Process,
         posix_thread::{PosixThread, ThreadLocal, ThreadName, sigkill_other_threads, thread_table},
-        process_vm::{MAX_LEN_STRING_ARG, MAX_NR_STRING_ARGS, unshare_and_renew_vmar},
+        process_vm::{MAX_LEN_STRING_ARG, MAX_NR_STRING_ARGS, new_vmar_and_map},
         program_loader::{ProgramToLoad, elf::ElfLoadInfo},
         signal::{
             HandlePendingSignal, PauseReason, SigStack,
@@ -25,6 +23,7 @@ use crate::{
             signals::kernel::KernelSignal,
         },
     },
+    vm::vmar::Vmar,
 };
 
 pub fn do_execve(
@@ -55,6 +54,9 @@ pub fn do_execve(
     let program_to_load =
         ProgramToLoad::build_from_inode(elf_inode.clone(), &fs_resolver, argv, envp)?;
 
+    let new_vmar = new_vmar_and_map(elf_file.clone());
+    let elf_load_info = program_to_load.load_to_vmar(new_vmar.as_ref(), &fs_resolver)?;
+
     // Ensure no other thread is concurrently performing exit_group or execve.
     // If such an operation is in progress, return EAGAIN.
     let mut task_set = ctx.process.tasks().lock();
@@ -73,7 +75,7 @@ pub fn do_execve(
     // After this point, failures in subsequent operations are fatal: the process
     // state may be left inconsistent and it can never return to user mode.
 
-    let res = do_execve_no_return(ctx, user_context, elf_file, &fs_resolver, program_to_load);
+    let res = do_execve_no_return(ctx, user_context, elf_file, new_vmar, &elf_load_info);
 
     if res.is_err() {
         ctx.posix_thread
@@ -127,8 +129,8 @@ fn do_execve_no_return(
     ctx: &Context,
     user_context: &mut UserContext,
     elf_file: PathOrInode,
-    fs_resolver: &FsResolver,
-    program_to_load: ProgramToLoad,
+    new_vmar: Arc<Vmar>,
+    elf_load_info: &ElfLoadInfo,
 ) -> Result<()> {
     let Context {
         process,
@@ -142,20 +144,16 @@ fn do_execve_no_return(
     wait_other_threads_exit(ctx)?;
     thread_table::make_current_main_thread(ctx);
 
-    let elf_load_info = {
-        let mut vmar = ctx.process.lock_vmar();
-        // Reset the virtual memory state.
-        unshare_and_renew_vmar(ctx, &mut vmar, elf_file.clone());
-        // Load the binary into the process's address space
-        program_to_load.load_to_vmar(vmar.unwrap(), fs_resolver)?
-    };
+    // Activate the new VMAR, where the ELF has been loaded, in the current context.
+    activate_vmar(ctx, new_vmar);
+
     // After the program has been successfully loaded, the virtual memory of the current process
     // is initialized. Hence, it is necessary to clear the previously recorded robust list.
     *thread_local.robust_list().borrow_mut() = None;
     thread_local.clear_child_tid().set(0);
 
     // Set up the CPU context.
-    set_cpu_context(thread_local, user_context, &elf_load_info);
+    set_cpu_context(thread_local, user_context, elf_load_info);
 
     // Apply file-capability changes.
     apply_caps_from_exec(process, posix_thread, elf_file.inode())?;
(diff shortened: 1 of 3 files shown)


kernel/src/process/program_loader/elf/elf_file.rs
```diff
@@ -9,136 +10,174 @@ use xmas_elf::{
...
+    vm::{perms::VmPerms, vmar::VMAR_CAP_ADDR},
 };
+
+/// A wrapper for the [`xmas_elf`] ELF parser.
 pub struct ElfHeaders {
-    pub elf_header: ElfHeader,
-    pub program_headers: Vec<ProgramHeader64>,
+    elf_header: ElfHeader,
+    load_phdrs: Vec<ElfLoadPhdr>,
+    load_align_ilog2: u32,
```
StevenJiang1110: Can we directly store `load_align` (without ilog2 here) for simplicity?

kernel/src/process/program_loader/elf/elf_file.rs
```diff
@@ -9,136 +10,174 @@ use xmas_elf::{
...
+            "the ELF program headers are not located in any segments"
+        );
     }
 
-    // An offset to be subtracted from ELF vaddr for PIE
-    pub fn base_load_address_offset(&self) -> u64 {
-        let phdr = self.program_headers.first().unwrap();
-        phdr.virtual_addr - phdr.offset
+    /// Sums (i.e., unions) the virtual address ranges of the program headers.
+    pub(super) fn sum_vaddrs_in_phdrs(&self) -> Range<Vaddr> {
```
StevenJiang1110: `sum` may not be very accurate here. You want a range including all program headers. How about `total_phdrs_virt_range` or `phdrs_virt_bounds`?

kernel/src/process/program_loader/elf/elf_file.rs
```diff
@@ -9,136 +10,174 @@ use xmas_elf::{
...
     prelude::*,
+    vm::{perms::VmPerms, vmar::VMAR_CAP_ADDR},
 };
+
+/// A wrapper for the [`xmas_elf`] ELF parser.
 pub struct ElfHeaders {
-    pub elf_header: ElfHeader,
-    pub program_headers: Vec<ProgramHeader64>,
+    elf_header: ElfHeader,
+    load_phdrs: Vec<ElfLoadPhdr>,
```
StevenJiang1110: Would it be better to rename this field to `loadable_phdrs` and the type to `ElfLoadablePhdr`? If `ElfLoadablePhdr` feels too verbose, we could simply drop the `Elf` prefix.

StevenJiang1110 (approved):
LGTM! Just a few minor comments.

kernel/src/process/execve.rs
```diff
@@ -142,20 +144,16 @@ fn do_execve_no_return(
...
 
-    let elf_load_info = {
-        let mut vmar = ctx.process.lock_vmar();
-        // Reset the virtual memory state.
-        unshare_and_renew_vmar(ctx, &mut vmar, elf_file.clone());
-        // Load the binary into the process's address space
-        program_to_load.load_to_vmar(vmar.unwrap(), fs_resolver)?
-    };
+    // Activate the new VMAR, where the ELF has been loaded, in the current context.
+    activate_vmar(ctx, new_vmar);
```
junyang-zh, line 148: Another option is to activate a new VMAR before loading the elf program, just like how Linux does. On `execve`, Linux does some minimal preparation, activate a new address space, [do a schedule here since the context is effectively minimal](https://elixir.bootlin.com/linux/v6.17/source/kernel/sched/core.c#L5468), then loads the program.

In this way we may do some auditing before loading the new VMAR to ensure the failure cases are not hard errors. But I am not sure how much changes are needed. Doing it in the future is acceptable.

junyang-zh (approved):
LGTM. I like it and want it to be merged soon.

kernel/src/process/program_loader/elf/elf_file.rs
lrh2000: Rename it to:
```rust
    /// Calculates the virtual address bounds of all segments as a range.
    pub(super) fn calc_total_vaddr_bounds(&self) -> Range<Vaddr> {
```

Since it is not a zero-cost method, should we prefer to have a verb here?

https://github.com/asterinas/asterinas/pull/2736 may want:
```rust
    /// Calculates the virtual address bounds of data segments as a range.
    ///
    /// Note that this method only considers the data portion. In other words, it accounts for the
    /// file size of the segment instead of the memory size.
    pub(super) fn calc_data_vaddr_bounds(&self) -> Option<Range<Vaddr>> {
```

kernel/src/process/program_loader/elf/elf_file.rs
lrh2000: My initial thought is that using `ilog2` will make it clearer that it's a power of two. However, it is already mentioned in the doc comments, so I will make the change.

kernel/src/process/program_loader/elf/elf_file.rs
StevenJiang1110: > Rename it to:
> 
> ```rust
>     /// Calculates the virtual address bounds of all segments as a range.
>     pub(super) fn calc_total_vaddr_bounds(&self) -> Range<Vaddr> {
> ```
> 
> Since it is not a zero-cost method, should we prefer to have a verb here?

`calc_total_vaddr_bounds` is acceptable to me.

kernel/src/process/execve.rs
lrh2000: Yes. However, I think you missed my PR description of the `About the last commit: "Separate VMAR load and activation"` section.

Specially,
>> Having the same point of no return as Linux would require additional refactoring. However, I found that it is quite easy for us to defer VMAR activation, allowing us to have the point of no return after Step 3. This means that all error codes will be visible to the user and can be properly tested.

"Additional refactoring" refers to moving the `ld.so` processing to the `ElfHeaders::parse` function. It's feasible, but the current implementation is simpler and sufficient, so I'm treating it as future work.

> [do a schedule here since the context is effectively minimal](https://elixir.bootlin.com/linux/v6.17/source/kernel/sched/core.c#L5468),

It's interesting that I hadn't noticed it before. Thanks for the information!

Issue #2630, opened 2025-11-28
Tracking issue: Add SCML documentations for all syscalls


taosue:
### Description

Currently, 60 syscalls require analysis to determine feature coverage and subsequent SCML documentation. I have utilized `git blame` to statistically analyze the contribution proportions and have assigned owners to document the related syscalls. The assignment rule designates the primary contributor (≥40% of code lines) as the owner if they are an active contributor; otherwise, ownership is assigned to me. Please review these assignments and share any concerns.

### Current Status

Here is the table about the syscalls and owners.

| System Call | Contributors | Owner | Status |
|------------|-------------|-------|-------|
| ioctl | File System Coverage related | Tao Su | Done |
| select | LI Qing (63.2%), Jianfeng Jiang (12.6%), Zhang Junyang (8.9%) | Tao Su | Done |
| nanosleep | Jianfeng Jiang (52.6%), Chen Chengjun (35.0%), Zhang Junyang (7.3%) | Jianfeng Jiang | Done |
| getitimer | Chen Chengjun (88.1%), Zhang Junyang (7.3%), Ruihan Li (2.8%) | Chen Chengjun | Done |
| setitimer | Chen Chengjun (88.1%), Zhang Junyang (7.3%), Ruihan Li (2.8%) | Chen Chengjun | Done |
| accept | Jianfeng Jiang (75.0%), Ruihan Li (8.7%), Zhang Junyang (5.4%) | Jianfeng Jiang | Done |
| shutdown | Ruihan Li (52.4%), Jianfeng Jiang (42.9%), YanWQ-monad (4.8%) | Ruihan Li | Done |
| listen | Ruihan Li (47.4%), Jianfeng Jiang (47.4%), YanWQ-monad (5.3%) | Ruihan Li | Done |
| getpeername | Ruihan Li (44.4%), Jianfeng Jiang (33.3%), Zhang Junyang (18.5%) | Ruihan Li | Done |
| flock | Fabing Li (86.4%), Zejun Zhao (7.4%), Ruihan Li (4.9%) | Tao Su | Done |
| getrlimit | Jianfeng Jiang (34.4%), Wang Siyuan (31.1%), Fabing Li (26.2%) | Tao Su | Done |
| sigaltstack | jiangjianfeng (63.0%), Jianfeng Jiang (27.5%), Zhang Junyang (3.6%) | jiangjianfeng | Done |
[...]

PR #2778, opened 2025-12-21
Add SCMLs for `shutdown()`, `listen()`, and `getpeername()`

lrh2000:
It seems that we have full support for `shutdown()`, `listen()`, and `getpeername()`. So I add SCMLs under `book/src/kernel/linux-compatibility/syscall-flag-coverage/networking-and-sockets/fully_covered.scml` and update the entries under `book/src/kernel/linux-compatibility/README.md` (cc https://github.com/asterinas/asterinas/issues/2630).

---

In addition, I caught some outdated information. For example, we support `vfork()` system call, and I don't know why it is marked as unsupported. 
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/syscall/arch/x86.rs#L233

`LINUX_REBOOT_CMD_RESTART` is also a supported `reboot()` flag now.

I included some corrections for them in the second commit.


book/src/kernel/linux-compatibility/syscall-flag-coverage/system-information-and-misc/reboot.scml
```diff
@@ -4,6 +4,6 @@ reboot_magic2 = LINUX_REBOOT_MAGIC2 | LINUX_REBOOT_MAGIC2A | LINUX_REBOOT_MAGIC2
 reboot(
     magic  = LINUX_REBOOT_MAGIC1,
     magic2 = <reboot_magic2>,
-    op     = LINUX_REBOOT_CMD_HALT | LINUX_REBOOT_CMD_POWER_OFF,
+    op     = LINUX_REBOOT_CMD_HALT | LINUX_REBOOT_CMD_POWER_OFF | LINUX_REBOOT_CMD_RESTART,
```
taosue, line 7: @cchanging could you confirm if this flag is supported? If yes, we should remove this flag at the same time:https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/book/src/kernel/linux-compatibility/syscall-flag-coverage/system-information-and-misc/README.md?plain=1#L66

lrh2000: > we should remove this flag at the same time

Thanks for the reminder! It's a bit harder to keep everything in sync...

> confirm if this flag is supported

I assume that we can mark it as supported because https://github.com/asterinas/asterinas/commit/35d70fca712b80d55021be6528eda76d5bb41dcc#diff-130a8a4600e68a4dd64976732e1fdc45cc69e8287f31ff0293c6102da4bd9a53L58-R64.

cchanging: > @cchanging could you confirm if this flag is supported?

I think we can mark it as supported, just as @lrh2000 mentioned, its basic behavior is already supported.

tatetian (approved):

Issue #2749, opened 2025-12-13
Syscall `setrlimit` not work as expected (inconsistent with Linux)


Marsman1996:
### Describe the bug
Syscall `setrlimit` not work as expected, stacks cannot grow past 8 MiB despite higher `RLIMIT_STACK`.

### To Reproduce

Compile the program and run:
```c
#define _GNU_SOURCE
#include <err.h>
#include <stdio.h>
#include <string.h>
#include <sys/resource.h>

__attribute__((noinline))
void eat_stack(size_t bytes)
{
    volatile char padding[4096];
    memset((void *)padding, 0xA5, sizeof(padding));
    if (bytes > sizeof(padding))
        eat_stack(bytes - sizeof(padding));
}

int main(void)
{
    struct rlimit rl = { 64UL * 1024 * 1024, 64UL * 1024 * 1024 };
    if (setrlimit(RLIMIT_STACK, &rl) == -1)
        err(1, "setrlimit");

    puts("Growing stack…");
    eat_stack(32UL * 1024 * 1024);   // walk 32 MiB below the current SP
    puts("Success");
    return 0;
}
```

### Expected behavior
The program should not SEGV.

### Environment
Official docker asterinas/asterinas:0.16.2-20251211
Intel(R) Xeon(R) Gold 6230R CPU @ 2.10GHz
Asterinas version: main 8b219c077e6cd1c179075bb13b77806554a3af72

### Logs
- In Asterinas
```
Growing stack…
Segmentation fault
```

- In Linux
```
Growing stack…
Success
```

lrh2000:
This is to be expected since I don't believe we support dynamically resized stack mappings. In Linux, the stack mapping automatically grows down when the guard page is accessed, assuming the limit isn't reached. However, we simply don't have this mechanism.

So this is mostly a memory-management issue rather than a `setrlimit` issue.

PR #2783, opened 2025-12-22
Fix OOM by adding SYSCTL_NR_OPEN limit for `setrlimit` and `prlimit64`

### LINKED ISSUE #2782: System exits when calling `poll` with large `nfds`
### Describe the bug
`Vec::with_capacity(nfds as _)` in `kernel/src/syscall/poll.rs` allocates `nfds` entries without bounding the size beyond `RLIMIT_NOFILE`.
https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/src/syscall/poll.rs#L34-L51

### To Reproduce
1. Raise `RLIMIT_NOFILE` to a huge value (e.g., `INT_MAX`) and call `poll` with an equally huge `nfds` (e.g., `0x40000000`). The allocation overflows/OOMs and panics before processing descriptors.
```C
#define _GNU_SOURCE
#include <errno.h>
#include <limits.h>
#include <poll.h>
#include <stdio.h>
#include <sys/resource.h>
#include <unistd.h>

int main(void) {
    struct rlimit rl = {INT_MAX, INT_MAX};
    if (setrlimit(RLIMIT_NOFILE, &rl) != 0) {
        perror("setrlimit");
    }

    nfds_t nfds = 0x40000000; // large but <= INT_MAX
    struct pollfd dummy = {.fd = -1, .events = 0};

    printf("poll with nfds=%u\n", (unsigned)nfds);
    int ret = poll(&dummy, nfds, 0);
    printf("poll ret=%d errno=%d\n", ret, errno);
    return ret;
}
```
2. Run the compiled program in Asterinas.

### Environment
- Asterinas version: main 9c4f644bd

### Logs
```
~ # /etc/poll-nfds-oom.c 
poll with nfds=2147483646
[   487.205] ERROR: Failed to allocate a large slot
[   487.206] ERROR: Heap allocation error, layout = Layout {
    size: 0x7ffffffe0,
    align: 4 (1 << 2),
}
ake: *** [Makefile:308: run_kernel] Error 1
```

Marsman1996:
Fix #2782

### DIFF
--- a/kernel/src/process/rlimit.rs
+++ b/kernel/src/process/rlimit.rs
@@ -25,6 +25,8 @@ const INIT_RLIMIT_NOFILE_MAX: u64 = 4096;
 const INIT_RLIMIT_MEMLOCK: u64 = 8 * 1024 * 1024;
 // https://github.com/torvalds/linux/blob/fac04efc5c793dccbd07e2d59af9f90b7fc0dca4/include/uapi/linux/mqueue.h#L26
 const INIT_RLIMIT_MSGQUEUE: u64 = 819200;
+// https://github.com/torvalds/linux/blob/fac04efc5c793dccbd07e2d59af9f90b7fc0dca4/fs/file.c#L90
+pub const SYSCTL_NR_OPEN: u64 = 1024 * 1024;
 
 #[derive(Clone)]
 pub struct ResourceLimits {
@@ -80,7 +82,7 @@ impl Default for ResourceLimits {
 }
 
 #[repr(u32)]
-#[derive(Debug, Clone, Copy, TryFromInt)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
 pub enum ResourceType {
     RLIMIT_CPU = 0,
     RLIMIT_FSIZE = 1,
--- a/kernel/src/syscall/prlimit64.rs
+++ b/kernel/src/syscall/prlimit64.rs
@@ -10,7 +10,7 @@ use crate::{
         credentials::capabilities::CapSet,
         posix_thread::{AsPosixThread, PosixThread},
         process_table,
-        rlimit::RawRLimit64,
+        rlimit::{RawRLimit64, SYSCTL_NR_OPEN},
     },
 };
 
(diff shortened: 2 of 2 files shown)


lrh2000:
Technically, this does not resolve issue https://github.com/asterinas/asterinas/issues/2782; it merely makes exploitation more challenging.

A malicious user could still exploit issue https://github.com/asterinas/asterinas/issues/2782 in an attempt to exhaust **almost** all of the kernel's memory. This would trigger the same problem when other system calls perform infallible allocation.

Marsman1996:
Well, is this because of the implementation of the `setrlimit`? 
Since system shall reject the unreasonable rlimit extend, and the following check will work:
```
 if nfds as u64 
     > ctx 
         .process 
         .resource_limits() 
         .get_rlimit(ResourceType::RLIMIT_NOFILE) 
         .get_cur() 
 { 
     return_errno_with_message!( 
         Errno::EINVAL, 
         "the `nfds` value exceeds the `RLIMIT_NOFILE` value" 
     ) 
 } 
```

Marsman1996:
And #2749 shows that the implementation of `setrlimit` is inconsistent with Linux.

In Linux, the `setrlimit` will fail for 
```
struct rlimit rl = {INT_MAX, INT_MAX};
    if (setrlimit(RLIMIT_NOFILE, &rl) != 0) {
        perror("setrlimit");
    }
```

```
setrlimit: Operation not permitted
```

StevenJiang1110:
> And https://github.com/asterinas/asterinas/issues/2749 shows that the implementation of setrlimit is inconsistent with Linux.

The Linux kernel validates constraints when configuring RLIMIT_NOFILE (see [here](https://elixir.bootlin.com/linux/v6.18/source/kernel/sys.c#L1513)). The relevant `sysctl_nr_open` definition is located [here](https://elixir.bootlin.com/linux/v6.18/source/fs/file.c#L96). Do you think fixing this behavior would solve the problem? If so, we should target this fix instead.

Marsman1996:
> > And #2749 shows that the implementation of setrlimit is inconsistent with Linux.
> 
> The Linux kernel validates constraints when configuring RLIMIT_NOFILE (see [here](https://elixir.bootlin.com/linux/v6.18/source/kernel/sys.c#L1513)). The relevant `sysctl_nr_open` definition is located [here](https://elixir.bootlin.com/linux/v6.18/source/fs/file.c#L96). Do you think fixing this behavior would solve the problem? If so, we should target this fix instead.

Well, I think this will definitely solve the problem due to the current constraint about `nfds` and `RLIMIT_NOFILE` in `sys_poll`.

I will push a new commit later.

Marsman1996:
@StevenJiang1110 
Add checking about`SYSCTL_NR_OPEN` in `setrlimit` and `prlimit64`.

With this commit, the PoC runs:
```
    .:-. Presented by the Asterinas developers
    .. Released under Mozilla Public License v2.0

~ # /etc/poll-nfds-oom.c 
setrlimit: Operation not permitted
poll with nfds=2147483646
poll ret=-1 errno=22
```

Marsman1996:
Same CI error as https://github.com/asterinas/asterinas/actions/runs/20424143342/job/58681115875

StevenJiang1110 (approved):

It seems a bit redundant to have the same check logic duplicated. Ideally, getrlimit and setrlimit should call do_prlimit instead to avoid this duplication. However, that refactor might be out of scope for this particular PR.

Issue #1052, opened 2024-07-12
MMIO access generates unexpected instruction


Hsy-Intel:
### Describe the bug

The [two lines in MMIO init file](https://github.com/asterinas/asterinas/blob/main/ostd/src/bus/mmio/mod.rs#L62-L63):

```rust
let value = unsafe { *(paddr_to_vaddr(current) as *const u32) };
if value == VIRTIO_MMIO_MAGIC {
```
will generate a `Cmp_rm32_imm32` instruction if we set `RELEASE=1`.

However, we expect to access the MMIO region using only read or write instructions. Only these cases are handled in TDX MMIO emulation.

Therefore, we need to prevent the compiler from generating similar instructions when accessing MMIO. (Currently only found this one)

sdww0:
Maybe we can use `Volatile<&' static u32, ReadOnly>` or `core::ptr::read_volatile` to read the magic value. I think it will fix the problem. Can you try the following code and see whether the issue will fixed?

```rust
let magic = Volatile::new_read_only(unsafe { &*(paddr_to_vaddr(current) as *const u32) });
if magic.read() == VIRTIO_MMIO_MAGIC {
```

or

```rust
let magic = unsafe { core::ptr::read_volatile(paddr_to_vaddr(current) as *const u32) };
if magic == VIRTIO_MMIO_MAGIC {
```

Hsy-Intel:
> Maybe we can use `Volatile<&' static u32, ReadOnly>` or `core::ptr::read_volatile` to read the magic value. I think it will fix the problem. Could you please try running the following code to see if it fixes the issue?
> 
> ```rust
> let magic = Volatile::new_read_only(unsafe { &*(paddr_to_vaddr(current) as *const u32) });
> if magic.read() == VIRTIO_MMIO_MAGIC {
> ```
> 
> or
> 
> ```rust
> let magic = unsafe { core::ptr::read_volatile(paddr_to_vaddr(current) as *const u32) };
> if magic == VIRTIO_MMIO_MAGIC {
> ```

Both work fine, thanks!

PR #2701, opened 2025-12-08
Rust 2024 & 20251208 toolchain

junyang-zh:
This PR upgrades the toolchain to 20251208 and adopts Rust 2024 for all crates. Numerous fixes are addressed, and most changes resulting from the same language change are grouped into separate commits.

A Docker version bump is not required for merging this PR. But if merging it without new Docker images, users and CI would have to spend extra time downloading the new toolchain.

PR #1546, opened 2024-11-02
Add TDX compilation test

Hsy-Intel:

PR #2785, opened 2025-12-22
Fix issues in TDX release mode

### LINKED ISSUE #2790: Failed to boot as TDX guest due to VE after logger initialization
### Describe the bug

Asterinas failed to boot as TDX guest due to virtualization exception after logger initialization.

### To Reproduce

On a TDX-ready machine:

```bash
make run_kernel NETDEV=tap AUTO_TEST=syscall INTEL_TDX=1 RELEASE=1 ENABLE_KVM=1 LOG_LEVEL=info
```

See also https://github.com/asterinas/asterinas/actions/runs/20445069325/job/58746679579

It is weird that if `RELEASE=0` or `LOG_LEVEL=debug`, the VE will not happen. Only when `RELEASE=1` and `LOG_LEVEL=info`/`LOG_LEVEL=error` will trigger this VE.

### Logs

```
INFO: QEMU hypervisor detected, assuming that the isa-debug-exit device exists
INFO: Exists components that are not initialized
INFO: Components initializing in Bootstrap stage...
INFO: Component initializing:ComponentInfo { name: "aster-time", path: "/root/asterinas/kernel/comps/time", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-systree", path: "/root/asterinas/kernel/comps/systree", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-softirq", path: "/root/asterinas/kernel/comps/softirq", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-pci", path: "/root/asterinas/kernel/comps/pci", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-input", path: "/root/asterinas/kernel/comps/input", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-console", path: "/root/asterinas/kernel/comps/console", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-cmdline", path: "/root/asterinas/kernel/comps/cmdline", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-block", path: "/root/asterinas/kernel/comps/block", priority: 0 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-network", path: "/root/asterinas/kernel/comps/network", priority: 1 }
INFO: Component initialize complete
INFO: Component initializing:ComponentInfo { name: "aster-logger", path: "/root/asterinas/kernel/comps/logger", priority: 1 }
Printing stack trace:
   1: fn 0xffffffff88223d80 - pc 0xffffffff88224059 / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx              0x6b0;
     rsi                0x2; rdi 0xffffffff880490c0; rbp                0x2; rsp 0xffffffff880490f0;

   2: fn 0xffffffff8823d320 - pc 0xffffffff8823d326 / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx 0xffffffff88049338;
     rsi                0x2; rdi 0xffffffff880490c0; rbp                0x2; rsp 0xffffffff88049200;

   3: fn 0xffffffff882a4550 - pc 0xffffffff882a4556 / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx 0xffffffff88049338;
     rsi                0x2; rdi 0xffffffff880490c0; rbp                0x2; rsp 0xffffffff88049210;

   4: fn 0xffffffff882f9020 - pc 0xffffffff882f904b / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx 0xffffffff88049338;
     rsi                0x2; rdi 0xffffffff880490c0; rbp                0x2; rsp 0xffffffff88049220;

   5: fn 0xffffffff882f1dc0 - pc 0xffffffff882f1e12 / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx 0xffffffff88049338;
     rsi                0x2; rdi 0xffffffff880490c0; rbp                0x2; rsp 0xffffffff88049250;

   6: fn 0xffffffff883691b0 - pc 0xffffffff883695a9 / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx 0xffffffff88049338;
     rsi                0x2; rdi 0xffffffff880490c0; rbp                0x2; rsp 0xffffffff880492a0;

   7: fn 0xffffffff883475e0 - pc 0xffffffff88347833 / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx           0x210246;
     rsi                0x2; rdi 0xffffffff880490c0; rbp 0xffff8002608b1d00; rsp 0xffffffff88049330;

   8: fn                0x0 - pc 0xffffffff8834419a / registers:

     rax             0xfc00; rdx                0x1; rcx               0x2d; rbx 0xffff800260151fe0;
     rsi                0x2; rdi 0xffffffff880490c0; rbp 0xffff8002608b1d00; rsp 0xffffffff880493e0;

[OSDK] The kernel seems panicked. Parsing stack trace for source lines:
(  1) _RNvNtNtCs2yKuD6yrEUj_9aster_nix6thread4oops13panic_handler.llvm.16892048459120571493
          at aster_nix.1dd2d24a4ff03c8d-cgu.12:?
(  2) __ostd_panic_handler
          at ??:?
(  3) __rustc::rust_begin_unwind
          at aster_nix_osdk_bin.61e25f6fc658d493-cgu.0:?
(  4) core::panicking::panic_fmt
          at ??:?
(  5) core::result::unwrap_failed
          at ??:?
(  6) tdx_guest::handle_virtual_exception
          at ??:?
(  7) trap_handler
          at ??:?
(  8) _trap_from_kernel
          at ostd.bbf9d39bf62ce35d-cgu.08:?
```

Hsy-Intel:
In TDX release mode:
1. Using `LOGGER` in Asterinas results in the generation of `CMP` instructions
2. And calling the `fn num_queues()` function in the `kernel/comps/virtio/src/transport/pci/device.rs` file may also generate `CMP` instructions.

e.g.

https://github.com/asterinas/asterinas/blob/9c4f644bd9287da1815a13115fbdfa914d8426f0/kernel/comps/virtio/src/transport/pci/device.rs#L82-L84

However, we expect to access the MMIO region using only read or write instructions. Similar to #1052.

Fixes #2790

### DIFF
--- a/ostd/src/mm/io.rs
+++ b/ostd/src/mm/io.rs
@@ -42,6 +42,8 @@
 
 use core::{marker::PhantomData, mem::MaybeUninit};
 
+#[cfg(target_arch = "x86_64")]
+use crate::if_tdx_enabled;
 use crate::{
     Error, Pod,
     arch::mm::{
@@ -509,6 +511,52 @@ impl<'a> VmReader<'a, Infallible> {
     /// This method will panic if the current position of the reader does not meet the alignment
     /// requirements of type `T`.
     pub fn read_once<T: PodOnce>(&mut self) -> Result<T> {
+        #[cfg(target_arch = "x86_64")]
+        {
+            if_tdx_enabled!({
+                self.do_read_once_noinline::<T>()
+            } else {
+                self.do_read_once::<T>()
+            })
+        }
+        #[cfg(not(target_arch = "x86_64"))]
+        {
+            self.do_read_once::<T>()
+        }
+    }
+
+    // Accessing MMIO within a CVM guest triggers a hardware exception
+    // (#VE on Intel TDX or #VC on AMD SEV-SNP)
+    // which requires the guest kernel to manually decode the faulting instruction.
+    // Due to the inherent complexity of x86 instruction decoding,
+    // the kernel exception handler only supports a subset of simple memory-access instructions,
+    // such as the standard `mov`.
+    // Consequently, the kernel must be compiled such that
+    // all MMIO-triggering operations use only these supported instructions.
+    // To achieve this, we rely on `ptr::write_volatile` and `ptr::read_volatile`,
+    // assuming these APIs are sufficient to
+    // force the compiler to emit a straightforward `mov` instruction.
+    //
+    // This assumption held true until a recent upgrade of the Rust toolchain.
+    // We discovered that the newer LLVM backend now optimizes
+    // a volatile read followed by an integer comparison
+    // into a single `cmp` instruction that operates directly on memory.
+    // Decoding this fused instruction is beyond the capabilities of the CVM exception handler.
+    //
+    // We consider this aggressive optimization of a volatile access to be an LLVM bug;
+    // therefore, it should not be permanently "fixed" within Asterinas.
+    // Instead, we introduce a workaround using `#[inline(never)]`,
+    // which effectively prevents LLVM from performing this problematic instruction fusion.
+    //
+    // For discussion detail, see
+    // <https://github.com/asterinas/asterinas/pull/2785#issuecomment-3717554895>.
+    #[cfg(all(target_arch = "x86_64", feature = "cvm_guest"))]
+    #[inline(never)]
+    fn do_read_once_noinline<T: PodOnce>(&mut self) -> Result<T> {
+        self.do_read_once::<T>()
+    }
+
+    fn do_read_once<T: PodOnce>(&mut self) -> Result<T> {
         if self.remain() < size_of::<T>() {
             return Err(Error::InvalidArgs);
         }
@@ -760,6 +808,28 @@ impl<'a> VmWriter<'a, Infallible> {
     /// This method will panic if the current position of the writer does not meet the alignment
     /// requirements of type `T`.
     pub fn write_once<T: PodOnce>(&mut self, new_val: &T) -> Result<()> {
+        #[cfg(target_arch = "x86_64")]
+        {
+            if_tdx_enabled!({
+                self.do_write_once_noinline::<T>(new_val)
+            } else {
+                self.do_write_once::<T>(new_val)
+            })
+        }
+        #[cfg(not(target_arch = "x86_64"))]
+        {
+            self.do_write_once::<T>(new_val)
+        }
+    }
+
+    // See the doc of `do_read_once_noinline` for why this noinline version is needed.
+    #[cfg(all(target_arch = "x86_64", feature = "cvm_guest"))]
+    #[inline(never)]
+    fn do_write_once_noinline<T: PodOnce>(&mut self, new_val: &T) -> Result<()> {
+        self.do_write_once::<T>(new_val)
+    }
+
+    fn do_write_once<T: PodOnce>(&mut self, new_val: &T) -> Result<()> {
         if self.avail() < size_of::<T>() {
             return Err(Error::InvalidArgs);
         }


StevenJiang1110:
> And calling the fn num_queues() function in the kernel/comps/virtio/src/transport/pci/device.rs file may also generate CMP instructions.

I don't quite see how that behavior is possible. Could you please share more information?

The `num_queues` method follows a call chain from `SafePtr::read_once` to `IoMem<Insensitive>::read_once`, and ultimately to `VmReader::read_once`, which invokes `ptr::read_volatile`. Because of this, the compiler cannot optimize away the read operation to perform a direct CMP on the MMIO region.

I conducted a simple experiment to verify this:

```rust
fn main() {
    let mut s: u64 = 0;
    let ptr = &mut s as *mut u64;
    unsafe {
        ptr.write_volatile(101);
    }

    let new_value = unsafe { ptr.read_volatile() };

    if new_value < 100 {
        println!("new value less than 100");
    } else {
        println!("new value greater than 100");
    }
}
```

Compiling with `cargo rustc --release -- --emit asm` (output located in `target/release/deps`) confirms that the volatile read is preserved in the generated assembly and is not optimized away.

```s
movq	$101, (%rsp)
movq	(%rsp), %rax
cmpq	$100, %rax
```

junyang-zh:
fix #2790

Hsy-Intel:
> I don't quite see how that behavior is possible. Could you please share more information?

This is somewhat strange; previously, similar commands were not generated here, but now they are, and the code hasn't been changed. I will try to provide more information.

Hsy-Intel:
I found something interesting when examining the actual disassembly that explains the MMIO exceptions we're seeing.

Disassembled code of this line:
   https://github.com/asterinas/asterinas/blob/2c80e0c3a4bfcd3b877790e50b4f27d52613629b/kernel/comps/virtio/src/transport/pci/device.rs#L82
   
   Debug mode:
   ```txt
  ffffffff88842710 <_RNvXs1_NtNtNtCs4A0tpGdnXYE_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9set_queue>:
  ffffffff88842710:       48 81 ec f8 01 00 00    sub    $0x1f8,%rsp
  ffffffff88842717:       4c 89 4c 24 08          mov    %r9,0x8(%rsp)
  ffffffff8884271c:       4c 89 44 24 10          mov    %r8,0x10(%rsp)
  ffffffff88842721:       48 89 4c 24 18          mov    %rcx,0x18(%rsp)
  ffffffff88842726:       48 89 7c 24 20          mov    %rdi,0x20(%rsp)
  ffffffff8884272b:       66 89 d0                mov    %dx,%ax
  ffffffff8884272e:       66 89 f2                mov    %si,%dx
  ffffffff88842731:       66 89 54 24 2a          mov    %dx,0x2a(%rsp)
  ffffffff88842736:       66 89 44 24 2c          mov    %ax,0x2c(%rsp)
  ffffffff8884273b:       48 89 bc 24 c0 00 00    mov    %rdi,0xc0(%rsp)
  ffffffff88842742:       00
  ffffffff88842743:       48 89 8c 24 c8 00 00    mov    %rcx,0xc8(%rsp)
  ffffffff8884274a:       00
  ffffffff8884274b:       4c 89 84 24 d0 00 00    mov    %r8,0xd0(%rsp)
  ffffffff88842752:       00
  ffffffff88842753:       4c 89 8c 24 d8 00 00    mov    %r9,0xd8(%rsp)
  ffffffff8884275a:       00
  ffffffff8884275b:       48 c7 84 24 e0 00 00    movq   $0x16,0xe0(%rsp)
  ffffffff88842762:       00 16 00 00 00
  ffffffff88842767:       48 c7 84 24 e8 00 00    movq   $0x16,0xe8(%rsp)
  ffffffff8884276e:       00 16 00 00 00
  ffffffff88842773:       48 c7 84 24 f0 00 00    movq   $0x18,0xf0(%rsp)
  ffffffff8884277a:       00 18 00 00 00
  ffffffff8884277f:       48 c7 84 24 f8 00 00    movq   $0x20,0xf8(%rsp)
  ffffffff88842786:       00 20 00 00 00
  ffffffff8884278b:       48 c7 84 24 00 01 00    movq   $0x28,0x100(%rsp)
  ffffffff88842792:       00 28 00 00 00
  ffffffff88842797:       48 c7 84 24 08 01 00    movq   $0x30,0x108(%rsp)
  ffffffff8884279e:       00 30 00 00 00
  ffffffff888427a3:       48 c7 84 24 10 01 00    movq   $0x1c,0x110(%rsp)
  ffffffff888427aa:       00 1c 00 00 00

  # function call preserved
  ffffffff888427af:       e8 0c ea ff ff          call   ffffffff888411c0 <_RNvXs1_NtNtNtCs4A0tpGdnXYE_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport10num_queues>
  ffffffff888427b4:       66 39 44 24 2a          cmp    %ax,0x2a(%rsp)
  ffffffff888427b9:       0f 83 c4 01 00 00       jae    ffffffff88842983 <_RNvXs1_NtNtNtCs4A0tpGdnXYE_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9set_queue+0x273>

   ```
   Release mode:
   ```txt
   ffffffff882e3000    <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9s   et_queue>:
   ffffffff882e3000:       41 57                   push   %r15
   ffffffff882e3002:       41 56                   push   %r14
   ffffffff882e3004:       53                      push   %rbx
   ffffffff882e3005:       48 83 ec 10             sub    $0x10,%rsp

   # Calculate the MMIO address
   ffffffff882e3009:       4c 8b 97 50 01 00 00    mov    0x150(%rdi),%r10
   ffffffff882e3010:       4d 8d 7a 12             lea    0x12(%r10),%r15

   # Bounds checking (SafePtr's Safety mechanism is inlined)
   ffffffff882e3014:       4c 8b 9f 68 01 00 00    mov    0x168(%rdi),%r11
   ffffffff882e301b:       4d 39 df                cmp    %r11,%r15
   ffffffff882e301e:       0f 87 b6 01 00 00       ja     ffffffff882e31da    <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9s   et_queue+0x1da>
   ffffffff882e3024:       4c 89 d8                mov    %r11,%rax
   ffffffff882e3027:       4c 29 f8                sub    %r15,%rax
   ffffffff882e302a:       48 83 f8 02             cmp    $0x2,%rax
   ffffffff882e302e:       0f 82 d6 01 00 00       jb     ffffffff882e320a    <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9s   et_queue+0x20a>

   # Address alignment check
   ffffffff882e3034:       48 8b 9f 58 01 00 00    mov    0x158(%rdi),%rbx
   ffffffff882e303b:       48 8b bf 60 01 00 00    mov    0x160(%rdi),%rdi
   ffffffff882e3042:       4c 8b 73 10             mov    0x10(%rbx),%r14
   ffffffff882e3046:       49 01 fe                add    %rdi,%r14
   ffffffff882e3049:       4d 01 f7                add    %r14,%r15
   ffffffff882e304c:       41 f6 c7 01             test   $0x1,%r15b
   ffffffff882e3050:       0f 85 9c 01 00 00       jne    ffffffff882e31f2    <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9s   et_queue+0x1f2>

   # Key direct MMIO comparisons
   ffffffff882e3056:       b0 01                   mov    $0x1,%al
   ffffffff882e3058:       66 41 3b 37             cmp    (%r15),%si
   ffffffff882e305c:       0f 83 3e 01 00 00       jae    ffffffff882e31a0    <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport9s   et_queue+0x1a0>
   ...
   ```

Disassembled code of `VirtioPciModernTransport::num_queues()`:

```txt
ffffffff882e24f0 <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport10num_queues>:
ffffffff882e24f0:       50                      push   %rax
ffffffff882e24f1:       48 8b 87 50 01 00 00    mov    0x150(%rdi),%rax
ffffffff882e24f8:       48 83 c0 12             add    $0x12,%rax
ffffffff882e24fc:       48 8b 8f 68 01 00 00    mov    0x168(%rdi),%rcx
ffffffff882e2503:       48 39 c8                cmp    %rcx,%rax
ffffffff882e2506:       77 28                   ja     ffffffff882e2530 <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport10num_queues+0x40>
ffffffff882e2508:       48 29 c1                sub    %rax,%rcx
ffffffff882e250b:       48 83 f9 02             cmp    $0x2,%rcx
ffffffff882e250f:       72 4f                   jb     ffffffff882e2560 <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport10num_queues+0x70>
ffffffff882e2511:       48 8b 97 58 01 00 00    mov    0x158(%rdi),%rdx
ffffffff882e2518:       48 8b 8f 60 01 00 00    mov    0x160(%rdi),%rcx
ffffffff882e251f:       48 03 4a 10             add    0x10(%rdx),%rcx
ffffffff882e2523:       48 01 c1                add    %rax,%rcx
ffffffff882e2526:       f6 c1 01                test   $0x1,%cl
ffffffff882e2529:       75 1d                   jne    ffffffff882e2548 <_RNvXs1_NtNtNtCseH8niFamGuW_12aster_virtio9transport3pci6deviceNtB5_24VirtioPciModernTransportNtB9_15VirtioTransport10num_queues+0x58>
# read_volatile() equivalent
ffffffff882e252b:       0f b7 01                movzwl (%rcx),%eax
ffffffff882e252e:       59                      pop    %rcx
ffffffff882e252f:       c3                      ret
```

Therefore, we can see that:
1. `num_queues()` does use `read_volatile()` and should prevent optimization.
2. However, the compiler is performing aggressive function inlining in release mode.
3. Instead of calling `num_queues()` and using its return value, the compiler generates code that directly accesses the MMIO address during the comparison.

The release mode comparison `cmp (%r15),%si` is directly reading from the MMIO-mapped num_queues register address that was calculated inline, bypassing the `read_volatile()` protection that exists within the `num_queues()` function.

To fix this issue, maybe we can use `#[inline(never)]` attribute for the `VirtioPciModernTransport::num_queues()` method.

Hsy-Intel:
@boterinas rerun

StevenJiang1110:
FYI: A similar problem has been reported on aarch64, the `read_volatile` is optimized away. https://github.com/rust-lang/rust/issues/131894. But I haven't found any problems related to x86_64-unknown-none target yet.

lrh2000:
> * Instead of calling `num_queues()` and using its return value, the compiler generates code that directly accesses the MMIO address during the comparison.

Shouldn't @StevenJiang1110's code be the "inlined" case, since there is only one method, i.e., `main`? However, the compiler still won't generate `cmp` instructions there. I don't think whether it's inline or not can be a problem here.

It would be better to construct a minimal reproducible example...

Hsy-Intel:
@StevenJiang1110 ‘s example failed to reproduce the problem because his code used constants during the comparison:

```RUST
// test1.rs
#![no_std]

#[no_mangle]
pub unsafe extern "C" fn compare_const() -> bool {
    let ptr = 0x1000 as *const u16;
    ptr.read_volatile() < 100
}

#[no_mangle]
pub unsafe extern "C" fn compare_var(index: u16) -> bool {
    let ptr = 0x1000 as *const u16;
    index < ptr.read_volatile()
}

#[no_mangle]
pub unsafe extern "C" fn compare_ptr(ptr: *const u16, index: u16) -> bool {
    index < ptr.read_volatile()
}

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    loop {}
}
```

Compile:

```BASH
rustc --target=x86_64-unknown-none --crate-type=staticlib --emit=asm -O test1.rs -o test1.s
```

Results:

```nasm
// ...
compare_const:
        movzwl  4096, %eax
        cmpw    $100, %ax
        setb    %al
        retq
// ...
compare_ptr:
        cmpw    (%rdi), %si
        setb    %al
        retq
// ...
compare_var:
        cmpw    4096, %di
        setb    %al
        retq

```

It can be observed that both `compare_ptr` and `compare_var` are optimized into a single `cmp` instruction. Why? The x86_64 instruction set does not support `cmp [mem], 100`; the memory value must first be read into a register.

The compiler considers `cmp [mem], reg` to be equivalent to `mov reg, [mem]; cmp reg, reg` under volatile semantics. However, for certain MMIO registers, we need to ensure separate `mov` instructions.

Hsy-Intel:
The following code reproduces the `CMP` instruction issue related to MMIO access in Asterinas:

```RUST
// test2.rs
#![no_std]

#[no_mangle]
pub unsafe extern "C" fn original_num_queues(addr: usize) -> u16 {
    use core::ptr;
    ptr::read_volatile(addr as *const u16)
}

#[no_mangle]
#[inline(never)]
pub unsafe extern "C" fn original_num_queues_noinline(addr: usize) -> u16 {
    use core::ptr;
    ptr::read_volatile(addr as *const u16)
}

#[no_mangle]
pub unsafe extern "C" fn original_set_queue(addr: usize, index: u16) -> bool {
    index < original_num_queues(addr)
}

#[no_mangle]
pub unsafe extern "C" fn original_set_queue_noinline(addr: usize, index: u16) -> bool {
    index < original_num_queues_noinline(addr)
}

#[no_mangle]
pub unsafe extern "C" fn asm_num_queues(addr: usize) -> u16 {
    let value: u16;
    core::arch::asm!(
        "mov {0:x}, [{1}]",
        out(reg) value,
        in(reg) addr,
        options(preserves_flags, nostack)
    );
    value
}

#[no_mangle]
pub unsafe extern "C" fn asm_set_queue(addr: usize, index: u16) -> bool {
    index < asm_num_queues(addr)
}

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    loop {}
}
```

Compile:

```BASH
rustc --target=x86_64-unknown-none --crate-type=staticlib --emit=asm -O test2.rs -o test2.s
```

Results:

```nasm
asm_num_queues:
        #APP

        movw    (%rdi), %ax

        #NO_APP
        retq
// ...
asm_set_queue:
        #APP

        movw    (%rdi), %ax

        #NO_APP
        cmpw    %ax, %si
        setb    %al
        retq
// ...
original_num_queues:
        movzwl  (%rdi), %eax
        retq
// ...
original_num_queues_noinline:
        movzwl  (%rdi), %eax
        retq
// ...
original_set_queue:
        cmpw    (%rdi), %si
        setb    %al
        retq
// ...
original_set_queue_noinline:
        pushq   %rbx
        movl    %esi, %ebx
        callq   *original_num_queues_noinline@GOTPCREL(%rip)
        cmpw    %ax, %bx
        setb    %al
        popq    %rbx
        retq
// ...
```

By comparing `original_set_queue` and `asm_set_queue`, it can be seen that, based on the same logic, the assembly instructions and the disassembled code generated using `read_volatile` are different.

By comparing `original_set_queue` and `original_set_queue_noinline`, it can be seen that `#[inline(never)]` can mitigate this problem.

Therefore, we need to fully consider the side effects of using `volatile`.

> FYI: A similar problem has been reported on aarch64, the read_volatile is optimized away. https://github.com/rust-lang/rust/issues/131894. But I haven't found any problems related to x86_64-unknown-none target yet.

Furthermore, based on the information in the links above, this phenomenon doesn't seem to be a bug... The root cause seems to lie in the LLVM implementation.

lrh2000:
> @StevenJiang1110 ‘s example failed to reproduce the problem because his code used constants during the comparison:

[Godbolt link](https://godbolt.org/z/EzP9xqzn9)

rustc 1.87:
```
compare_ptr:
        cmp     si, word ptr [rdi]
        setb    al
        ret
```

rustc 1.86:
```
compare_ptr:
        movzx   eax, word ptr [rdi]
        cmp     ax, si
        seta    al
        ret
```

From [the release notes of Rust 1.87](https://github.com/rust-lang/rust/releases/tag/1.87.0), it's likely caused by [Update to LLVM 20](https://github.com/rust-lang/rust/pull/135763), which [was merged](https://github.com/rust-lang/rust/commit/ce36a966c79e109dabeef7a47fe68e5294c6d71e) on Feb 18, 2025.

We switch the toolchain version from nightly-2025-02-01 to nightly-2025-12-06 [only recently](https://github.com/asterinas/asterinas/pull/2701) on Dec 9, 2025. So we encountered the problem almost a year after it was introduced to the Rust compiler...

StevenJiang1110:
> Furthermore, based on the information in the links above, this phenomenon doesn't seem to be a bug... The root cause seems to lie in the LLVM implementation.

> From [the release notes of Rust 1.87](https://github.com/rust-lang/rust/releases/tag/1.87.0), it's likely caused by https://github.com/rust-lang/rust/pull/135763, which [was merged](https://github.com/rust-lang/rust/commit/ce36a966c79e109dabeef7a47fe68e5294c6d71e) on Feb 18, 2025.

I also think this should be a LLVM bug, the generated LLVM IR seems correct, which contains a volatile load. But it is finally optimized away in the generated assembly.

```llvm
define noundef zeroext i1 @compare_ptr(ptr noundef %ptr, i16 noundef zeroext %index) unnamed_addr {
start:
  %0 = load volatile i16, ptr %ptr, align 2
  %_0 = icmp ult i16 %index, %0
  ret i1 %_0
}
```

From the [LLVM language reference](https://llvm.org/docs/LangRef.html#volatile-memory-accesses),
```plain
IR-level volatile loads and stores cannot safely be optimized into llvm.memcpy or llvm.memmove intrinsics 
even when those intrinsics are flagged volatile. 
Likewise, the backend should never split or merge target-legal volatile load/store instructions.
```

LLVM backend should never merge volatile load/store instructions. But it seems it does so......

ostd/src/mm/io.rs
```diff
@@ -760,6 +783,27 @@ impl<'a> VmWriter<'a, Infallible> {
     /// This method will panic if the current position of the writer does not meet the alignment
     /// requirements of type `T`.
     pub fn write_once<T: PodOnce>(&mut self, new_val: &T) -> Result<()> {
```
lrh2000, line 810:
```suggestion
//
// CVM guests typically require plain loads and stores for MMIO access.
// However, LLVM may optimize these accesses into more complex instructions
// (e.g., `cmp` in x86), even when volatile semantics are used. This is likely
// due to LLVM bugs. Currently, we work around this issue by preventing this
// function from being inlined, which should prevent the compiler from
// performing these improper optimizations. For more details, see
// <https://github.com/asterinas/asterinas/pull/2785#issuecomment-3717554895>.
#[cfg_attr(feature = "cvm", inline(never))]
pub fn write_once<T: PodOnce>(&mut self, new_val: &T) -> Result<()> {
```

Note that if LLVM refuses to treat this as a bug, a more appropriate solution would be to use inline assembly. For now, however, `inline(never)` sounds like a reasonable workaround, but:
 1. It needs to be well-documented.
 2. Compiler-time feature detection is intended for this case, as the inline semantics need to be carried out at compile time. `if_tdx_enabled` is a mixed compile-time and runtime detection. I doubt the runtime detection is worth it.

Hsy-Intel: The `cvm_guest` feature is enabled by default, even in non-TDX scenarios. This means it will also be compiled as `inline(never)`. Is this behavior consistent with our expectations?

lrh2000: > The `cvm_guest` feature is enabled by default, even in non-TDX scenarios.

Why is it enabled by default? Several times, I have encountered compiler errors that were not reported until I opened a PR on GitHub, which has a workflow that specifically enables the CVM feature. I don't believe the command `make run_kernel` enables this feature.

For example, I applied the following patch, yet `make run_kernel` still succeeded.
```patch
diff --git a/kernel/src/device/misc/tdxguest.rs b/kernel/src/device/misc/tdxguest.rs
index 32a46b446..9ea3fdb12 100644
--- a/kernel/src/device/misc/tdxguest.rs
+++ b/kernel/src/device/misc/tdxguest.rs
@@ -5,7 +5,7 @@ use core::{
     mem::{offset_of, size_of},
     time::Duration,
 };
-
+dadsadas
 use aster_util::{field_ptr, safe_ptr::SafePtr};
 use device_id::{DeviceId, MinorId};
 use ostd::{
```

Hsy-Intel: It's not enabled by default in the `aster-kernel` crate, but it is by default in the `ostd` crate.

lrh2000: Oh, thanks for the explanation! However, it's not clear to me why OSTD should enable this feature by default. Do you have an idea? To make Asterinas boot in TDX under the default settings?

I feel it a bit odd to dispatch a compile-time feature, such as `inline(never)`, via a runtime check. But I am not sure if there is a better approach _if_ we do want to enable `cvm` by default....

Hsy-Intel: My understanding is that OSTD is designed to natively support TEE (Trusted Execution Environment) and checks for different TEEs (For example, Intel TDX) at runtime. If no TEE environment is detected at runtime, OSTD will run in a non-TEE environment.

ostd/src/mm/io.rs
```diff
@@ -509,6 +511,34 @@ impl<'a> VmReader<'a, Infallible> {
...
+    }
+
+    // CVM guests typically require plain loads and stores for MMIO access.
+    // However, LLVM may optimize these accesses into more complex instructions
+    // (e.g., `cmp` in x86), even when volatile semantics are used. This is likely
+    // due to LLVM bugs. Currently, we work around this issue by preventing this
+    // function from being inlined, which should prevent the compiler from
+    // performing these improper optimizations. For more details, see
+    // <https://github.com/asterinas/asterinas/pull/2785#issuecomment-3717554895>.
+    #[cfg(target_arch = "x86_64")]
```
tatetian: Add the `cvm_guest` feature gate to the compile-time condition

ostd/src/mm/io.rs
```diff
@@ -509,6 +511,34 @@ impl<'a> VmReader<'a, Infallible> {
...
+                self.read_once_impl::<T>()
+            })
+        }
+        #[cfg(not(target_arch = "x86_64"))]
+        {
+            self.read_once_impl::<T>()
+        }
+    }
+
+    // CVM guests typically require plain loads and stores for MMIO access.
```
tatetian: I expand the paragraph to provide enough more background and explanation.

```rust
// Accessing MMIO within a CVM guest triggers a hardware exception
// (#VE on Intel TDX or #VC on AMD SEV-SNP)
// which requires the guest kernel to manually decode the faulting instruction.
// Due to the inherent complexity of x86 instruction decoding,
// the kernel exception handler only supports a subset of simple memory-access instructions,
// such as the standard `mov`.
// Consequently, the kernel must be compiled such that 
// all MMIO-triggering operations use only these supported instructions.
// To achieve this, we rely on `ptr::write_volatile` and `ptr::read_volatile`,
// assuming these APIs are sufficient to
// force the compiler to emit a straightforward `mov` instruction.
//
// This assumption held true until a recent upgrade of the Rust toolchain.
// We discovered that the newer LLVM backend now optimizes 
// a volatile read followed by an integer comparison 
// into a single `cmp` instruction that operates directly on memory.
// Decoding this fused instruction is beyond the capabilities of the CVM exception handler.
//
// We consider this aggressive optimization of a volatile access to be an LLVM bug;
// therefore, it should not be permanently "fixed" within Asterinas.
// Instead, we introduce a workaround using `#[inline(never)]`,
// which effectively prevents LLVM from performing this problematic instruction fusion.
//
// For discussion detail, see
// <https://github.com/asterinas/asterinas/pull/2785#issuecomment-3717554895>.
```

ostd/src/mm/io.rs
```diff
@@ -509,6 +511,34 @@ impl<'a> VmReader<'a, Infallible> {
     /// This method will panic if the current position of the reader does not meet the alignment
     /// requirements of type `T`.
     pub fn read_once<T: PodOnce>(&mut self) -> Result<T> {
+        #[cfg(target_arch = "x86_64")]
+        {
+            if_tdx_enabled!({
+                self.read_once_noinline::<T>()
+            } else {
+                self.read_once_impl::<T>()
```
tatetian: Rename `read_once_noinline` to `do_read_once_noinline` and `read_once_impl` to `do_read_once`. This makes the code real more symmetric.

```rust
    pub fn read_once<T: PodOnce>(&mut self) -> Result<T> {
        {
            if_tdx_enabled!({
                self.do_read_once_noinline::<T>()
            } else {
                self.do_read_once::<T>()
            }
        }
        #[cfg(not(target_arch = "x86_64"))]
        {
            self.do_read_once::<T>()
        }        
    }
```

The `write_once` methods should also be updated this way.

tatetian (commented):
Thanks everyone for diagnosing this bug. I have some minor comments for the workaround.

ostd/src/mm/io.rs
```diff
@@ -760,6 +808,52 @@ impl<'a> VmWriter<'a, Infallible> {
...
+    // into a single `cmp` instruction that operates directly on memory.
+    // Decoding this fused instruction is beyond the capabilities of the CVM exception handler.
+    //
+    // We consider this aggressive optimization of a volatile access to be an LLVM bug;
+    // therefore, it should not be permanently "fixed" within Asterinas.
+    // Instead, we introduce a workaround using `#[inline(never)]`,
+    // which effectively prevents LLVM from performing this problematic instruction fusion.
+    //
+    // For discussion detail, see
+    // <https://github.com/asterinas/asterinas/pull/2785#issuecomment-3717554895>.
```
tatetian: The text is too long to be repeated. Simply say:

```rust
// See the doc of `do_read_once_noinline` for why this noinline version is needed.
```

tatetian (commented):
Thanks! The code is almost perfect. Just fix the last tiny issue.

lrh2000:
While I don't object to merging this workaround as it is, the more I investigate this issue, the more convinced I am that we should use inline assembly for all MMIO accesses.

 - The first thing I discovered is that Linux uses inline assembly for MMIO accesses in common architectures including [x86](https://elixir.bootlin.com/linux/v6.18.6/source/arch/x86/include/asm/io.h#L57-L71), [RISC-V](https://elixir.bootlin.com/linux/v6.18.6/source/arch/riscv/include/asm/mmio.h#L18-L81), and [ARM](https://elixir.bootlin.com/linux/v6.18.6/source/arch/arm64/include/asm/io.h#L22-L96). This at least suggests that volatile semantics do not always work well with MMIO access.

 - Another thing I discovered is that my conclusion in https://github.com/asterinas/asterinas/pull/2785#issuecomment-3717619422 was incorrect. This is not a bug introduced in LLVM 20.

For the following code snippet,
```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn compare_ptr(ptr: *const u16, index: u16) -> bool {
    index < ptr.read_volatile()
}
```

rustc 1.86 [generates](https://godbolt.org/z/h1Wa4jfa8) the following LLVM IR:
```
define noundef zeroext i1 @compare_ptr(ptr noundef %ptr, i16 noundef zeroext %index) unnamed_addr {
start:
  %0 = load volatile i16, ptr %ptr, align 2
  %_0 = icmp ugt i16 %0, %index
  ret i1 %_0
}
```

rustc 1.87 [generates](https://godbolt.org/z/sorKq86Y8) a slightly different version:
```
define noundef zeroext i1 @compare_ptr(ptr noundef %ptr, i16 noundef zeroext %index) unnamed_addr {
start:
  %0 = load volatile i16, ptr %ptr, align 2
  %_0 = icmp ult i16 %index, %0
  ret i1 %_0
}
```

The small difference is:
```patch
-  %_0 = icmp ugt i16 %0, %index
+  %_0 = icmp ult i16 %index, %0
```

This should be equivalent. LLVM will [translate](https://godbolt.org/z/1heGaf9q7) the first one to:
```
compare_ptr:                            # @compare_ptr
        movzx   eax, word ptr [rdi]
        cmp     ax, si
        seta    al
        ret
```

However, LLVM will [translate](https://godbolt.org/z/jrfPx6s5W) the second one to:
```
compare_ptr:                            # @compare_ptr
        cmp     si, word ptr [rdi]
        setb    al
        ret
```

If we treat it as an LLVM bug, then it's one that has existed for years. I can [reproduce](https://godbolt.org/z/cEb5Wbazs) it in LLVM 15, which was released in 2022. I doubt it will be fixed anytime soon.

 - Finally, I run some experiments to investigate the behavior of C compilers using the following code:
```c
int test(volatile short *ptr, short index)
{
	return index < *ptr;
}
```

It appears that Clang versions 20 and higher [generate](https://godbolt.org/z/drafsW6sx) the problematic assembly:
```
test:
        xor     eax, eax
        cmp     si, word ptr [rdi]
        setl    al
        ret
```
Even more surprisingly, the latest GCC (which has not yet been released) [does](https://godbolt.org/z/c35oeoTcs) the same thing:
```
"test":
        xor     eax, eax
        cmp     WORD PTR [rdi], si
        setg    al
        ret
```
But GCC 15.2 still [behaves](https://godbolt.org/z/8zofKox1T) as expected:
```
test:
        movzx   eax, WORD PTR [rdi]
        cmp     ax, si
        setg    al
        movzx   eax, al
        ret
```

If neither GCC nor Clang can behave as expected, using inline assembly seems like the right choice. Linux uses inline assembly for MMIO accesses, so it seems to have learned from this.

---

By the way, @tatetian, could you please explain why we want to enable the CVM feature by default in OSTD (https://github.com/asterinas/asterinas/pull/2785#discussion_r2707645164)? It still feels quite odd to me that we don't have a way to disable the CVM feature of OSTD.

tatetian:
@lrh2000 In light of your new finding, I wrote a new issue, #2948, which would be the right fix to our target problem. We should pursue #2948 and abandon this PR.

@Hsy-Intel Do you agree? If so, could you take #2948?

Hsy-Intel:
> @lrh2000 In light of your new finding, I wrote a new issue, #2948, which would be the right fix to our target problem. We should pursue #2948 and abandon this PR.
> 
> @Hsy-Intel Do you agree? If so, could you take #2948?

OK, I will take this new issue.

tatetian:
> By the way, @tatetian, could you please explain why we want to enable the CVM feature by default in OSTD ([#2785 (comment)](https://github.com/asterinas/asterinas/pull/2785#discussion_r2707645164))? It still feels quite odd to me that we don't have a way to disable the CVM feature of OSTD.

This is doable in theory. A third-party `ostd`-based kernel could choose to disable the default feature of `ostd` with its `Cargo.toml` and add only the features that it wants.

Since Asterinas is positioned as a general-purpose kernel, it should be built to cover as many CPU features and support as many devices as possible. So why does include the `cvm_guest` feature by default surprise you?

lrh2000:
> > @Hsy-Intel Do you agree? If so, could you take #2948?
> 
> OK, I will take this new issue.

If we prefer a complete solution, I still have another thing to say about the design in #2948. Let's wait until we reach a consensus before starting the implementation. I will reply to #2948 later.

> Since Asterinas is positioned as a general-purpose kernel, it should be built to cover as many CPU features and support as many devices as possible. So why does include the `cvm_guest` feature by default surprise you?

But the CVM feature in `aster-kernel` isn't enabled by default, right?
 - So by default, we enable the `cvm_guest` in OSTD but disable the `cvm_guest` in the kernel.

Is this behavior really intuitive? Had @HSY-Intel not pointed it out to me, I would have had no idea that the two features with the same name worked this way. This oddity can only be discovered by reading the two `Cargo.toml` files of the OSTD and the kernel **very** carefully.

StevenJiang1110:
> But the CVM feature in aster-kernel isn't enabled by default, right?

I suspect this is a typo/misconfiguration. We currently have `all = ["cvm_guest"]`, but that's clearly not correct: we obviously have more than just the `cvm_guest` feature. Semantically, `all` should include every feature. So I think what we intended here was `default = ["cvm_guest"]`. Needs confirmation from @Hsy-Intel .

https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/kernel/Cargo.toml#L76-L81

Hsy-Intel:
> > But the CVM feature in aster-kernel isn't enabled by default, right?
> 
> I suspect this is a typo/misconfiguration. We currently have `all = ["cvm_guest"]`, but that's clearly not correct: we obviously have more than just the `cvm_guest` feature. Semantically, `all` should include every feature. So I think what we intended here was `default = ["cvm_guest"]`. Needs confirmation from @Hsy-Intel .
> 
> https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/kernel/Cargo.toml#L76-L81

I have checked the commit history, seems that we intentionally set it to ` all = ["cvm_guest"] `:

https://github.com/asterinas/asterinas/pull/1546#issuecomment-2452832242

https://github.com/asterinas/asterinas/blob/main/.github/actions/test/action.yml#L70

Should we change it to the `default = ["cvm_guest"]`?

lrh2000:
> Should we change it to the `default = ["cvm_guest"]`?

I don't know. However, even if we want to maintain the current behavior, I would argue that we should choose two different names for the feature names to improve clarity.

StevenJiang1110:
> I have checked the commit history, seems that we intentionally set it to all = ["cvm_guest"]:

> https://github.com/asterinas/asterinas/pull/1546#issuecomment-2452832242

> https://github.com/asterinas/asterinas/blob/main/.github/actions/test/action.yml#L70

> Should we change it to the default = ["cvm_guest"]?

Thanks for the information. When the `all` feature was first introduced, `cvm_guest` was the only feature, so having `all = ["cvm_guest"]` made sense at the time. However, now that we have additional features, the current feature table looks a bit misleading.

Semantically, the `all` feature should enable all available features, while the `default` feature set should stay consistent with OSTD.

Could we switch to something like:

```toml
[features]
# Please ensure every newly added feature is included in this list.
all = ["coverage", "cvm_guest", "riscv_sv39_mode"]
default = ["cvm_guest"]
```

tatetian:
> Semantically, the `all` feature should enable all available features, while the `default` feature set should stay consistent with OSTD.
> 
> Could we switch to something like:
> 
> ```toml
> [features]
> # Please ensure every newly added feature is included in this list.
> all = ["coverage", "cvm_guest", "riscv_sv39_mode"]
> default = ["cvm_guest"]
> ```

I don't think we have any good reason to keep the `all` feature. Just including `cvm_guest` to the `default` feature is suffcient.

> Should we change it to the `default = ["cvm_guest"]`?

Agree.

tatetian:
> I would argue that we should choose two different names for the feature names to improve clarity.

This feature opts in the support for Asterinas as the guest OS of a CVM. That's why it is called `cvm_guest`. So I think the name is quite precise. Asterinas does not support hardware-based virtualization, not to mention as the host OS for CVMs. If Asterinas has such support in the future, then this feature would be called `cvm_host`, if this support is to be feature gated.

@lrh2000 Sounds reasonable?

lrh2000:
> This feature opts in the support for Asterinas as the guest OS of a CVM. That's why it is called `cvm_guest`. So I think the name is quite precise. Asterinas does not support hardware-based virtualization, not to mention as the host OS for CVMs. If Asterinas has such support in the future, then this feature would be called `cvm_host`, if this support is to be feature gated.
> 
> @lrh2000 Sounds reasonable?

Oh, I don't mean to "rename" the feature. What I mean is: If OSTD enables `cvm_guest` by default but the kernel disables `cvm_guest` by default, then we should choose a different name for the kernel's feature to avoid confusion.

However, if you say that we should enable the kernel's `cvm_guest` feature by default, then I'm fine with that because there would be no confusion. In that case, there's no need to come up with another name.

What you explained above makes total sense to me. I am not questioning the precision of the name. Sorry for my misleading words, if any.

tatetian:
> > Semantically, the `all` feature should enable all available features, while the `default` feature set should stay consistent with OSTD.
> > Could we switch to something like:
> > ```toml
> > [features]
> > # Please ensure every newly added feature is included in this list.
> > all = ["coverage", "cvm_guest", "riscv_sv39_mode"]
> > default = ["cvm_guest"]
> > ```
> 
> I don't think we have any good reason to keep the `all` feature. Just including `cvm_guest` to the `default` feature is suffcient.
> 
> > Should we change it to the `default = ["cvm_guest"]`?
> 
> Agree.

@Hsy-Intel Could you make this change as well? ❤️

Hsy-Intel:
> @Hsy-Intel Could you make this change as well? ❤️

Sure.

zjp-CN:
> FYI: A similar problem has been reported on aarch64, the `read_volatile` is optimized away. [rust-lang/rust#131894](https://github.com/rust-lang/rust/issues/131894). But I haven't found any problems related to x86_64-unknown-none yet.

I reported such one in that issue. Rust-lang thinks volatile be [the recommended way](https://github.com/rust-lang/rust/issues/131894#issuecomment-3719922651) to do MMIO. 

Not sure if https://github.com/rust-lang/unsafe-code-guidelines/issues/321 will be accepted though.

tatetian:
> MMIO.
> 
> Not sure if [rust-lang/unsafe-code-guidelines#321](https://github.com/rust-lang/unsafe-code-guidelines/issues/321) will be accepted though.

@zjp-CN Thanks for sharing the discussion in rust-lang on the same problem.

Since kernel developers are used to write assembly code and this problem can be resolved by replacing `write_volatile` and `read_volatile` with assembly code, I don't think Asterinas can and should count on Rust or LLVM to fix this issue properly in the near future. We will fix this problem as outlined in #2948 and implemented in #2958.

PR #2791, opened 2025-12-24
Fix OOM in `getdents(64)` by user space writer

### LINKED ISSUE #1326: Capacity overflow in `sys_getdents()`
### Describe the bug
There is a capacity overflow in `sys_getdents()` at kernel/src/syscall/getdents64.rs:36 when call `sys_getdents` with large `buf_len`.

https://github.com/asterinas/asterinas/blob/f01772ca853e76d3076a561da3281034e3a46196/kernel/src/syscall/getdents64.rs#L36

### To Reproduce
1. Compile a program which calls `sys_getdents`: 
```C
#include <stdio.h>
#include <string.h>
#include <sys/socket.h>
#include <sys/mman.h>
#include <sys/stat.h>
#include <arpa/inet.h>

int main() {
    int fd = socket(1, 0x5, 0);
    char ubuf[8];
    ssize_t size = -7;
    int flags = 0x0;
    struct sockaddr_in addr;
    socklen_t addr_len = 18;

    recvfrom(fd, ubuf, size, flags, (struct sockaddr *)&addr, &addr_len);

    return 0;
}
```
2. Run the compiled program in Asterinas.

### Expected behavior

Asterinas reports capacity overflow and is terminated.

### Environment
- Official docker asterinas/asterinas:0.8.1
- 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
- Asterinas version: main 8bfbdf6

### Logs
```
~ # /root/getdents.c
panicked at /root/.rustup/toolchains/nightly-2024-06-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/raw_vec.rs:25:5:
capacity overflow
Printing stack trace:
   1: fn 0xffffffff887e5490 - pc 0xffffffff887e54fa / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a9296c0;

   2: fn 0xffffffff887e52a0 - pc 0xffffffff887e53f0 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929720;

   3: fn 0xffffffff88049000 - pc 0xffffffff8804900a / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929870;

   4: fn 0xffffffff889c2af0 - pc 0xffffffff889c2b72 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929880;

   5: fn 0xffffffff88987fe0 - pc 0xffffffff88988009 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929910;

   6: fn 0xffffffff88989710 - pc 0xffffffff8898974e / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929950;

   7: fn 0xffffffff8874d1b0 - pc 0xffffffff8874d20f / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929990;

   8: fn 0xffffffff88742380 - pc 0xffffffff887423b4 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a9299f0;

   9: fn 0xffffffff88749e10 - pc 0xffffffff88749e2f / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929a90;

  10: fn 0xffffffff885af2a0 - pc 0xffffffff885af899 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929ab0;

  11: fn 0xffffffff884ac9d0 - pc 0xffffffff884c30aa / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a929f30;

  12: fn 0xffffffff884ac1b0 - pc 0xffffffff884ac23e / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a9403d0;

  13: fn 0xffffffff884a15f0 - pc 0xffffffff884a216f / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a940570;

  14: fn 0xffffffff883a06f0 - pc 0xffffffff883a06fe / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a940f90;

  15: fn 0xffffffff88857ed0 - pc 0xffffffff88857ee6 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a940fb0;

  16: fn 0xffffffff88832c80 - pc 0xffffffff88832ce9 / registers:

     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a940fd0;


     rax                0x1; rdx                0x8; rcx                0x0; rbx                0x0;
     rsi                0x1; rdi 0xffffffff8883b687; rbp                0x0; rsp 0xffff80000a941000;

[OSDK] The kernel seems panicked. Parsing stack trace for source lines:
(  1) /root/asterinas/ostd/src/panicking.rs:113
(  2) /root/asterinas/ostd/src/panicking.rs:59
(  3) 73ctk84we8vz1kw7hv2sr00er:?
(  4) ??:?
(  5) alloc.4436a8b0cd505e33-cgu.0:?
(  6) /root/.rustup/toolchains/nightly-2024-06-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/raw_vec.rs:594
(  7) ??:?
(  8) /root/.rustup/toolchains/nightly-2024-06-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/spec_from_elem.rs:52
(  9) /root/.rustup/toolchains/nightly-2024-06-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:2737
( 10) /root/asterinas/kernel/src/syscall/getdents64.rs:36
( 11) /root/asterinas/kernel/src/syscall/mod.rs:171
( 12) /root/asterinas/kernel/src/syscall/mod.rs:328
( 13) /root/asterinas/kernel/src/thread/task.rs:69
( 14) /root/.rustup/toolchains/nightly-2024-06-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:79
( 15) /root/.rustup/toolchains/nightly-2024-06-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/boxed.rs:2077
( 16) /root/asterinas/ostd/src/task/mod.rs:175
```

Marsman1996:
Fix #1326

### DIFF
--- a/kernel/comps/systree/src/utils.rs
+++ b/kernel/comps/systree/src/utils.rs
@@ -11,7 +11,7 @@ use alloc::{
 use inherit_methods_macro::inherit_methods;
 use ostd::{
     mm::{VmReader, VmWriter},
-    sync::RwLock,
+    sync::RwMutex,
 };
 use spin::Once;
 
@@ -108,7 +108,7 @@ impl<T: SysNode> NormalNodeFields<T> {
 #[derive(Debug)]
 pub struct AttrLessBranchNodeFields<C: SysObj + ?Sized, T: SysBranchNode> {
     base: ObjFields<T>,
-    pub children: RwLock<BTreeMap<SysStr, Arc<C>>>,
+    pub children: RwMutex<BTreeMap<SysStr, Arc<C>>>,
 }
 
 #[inherit_methods(from = "self.base")]
@@ -116,7 +116,7 @@ impl<C: SysObj + ?Sized, T: SysBranchNode> AttrLessBranchNodeFields<C, T> {
     pub fn new(name: SysStr, weak_self: Weak<T>) -> Self {
         Self {
             base: ObjFields::new(name, weak_self),
-            children: RwLock::new(BTreeMap::new()),
+            children: RwMutex::new(BTreeMap::new()),
         }
     }
 
@@ -178,7 +178,7 @@ impl<C: SysObj + ?Sized, T: SysBranchNode> AttrLessBranchNodeFields<C, T> {
         children.get(name).cloned()
     }
 
-    pub fn children_ref(&self) -> &RwLock<BTreeMap<SysStr, Arc<C>>> {
+    pub fn children_ref(&self) -> &RwMutex<BTreeMap<SysStr, Arc<C>>> {
         &self.children
     }
 
@@ -228,7 +228,7 @@ impl<C: SysObj + ?Sized, T: SysBranchNode> BranchNodeFields<C, T> {
 
     pub fn child(&self, name: &str) -> Option<Arc<C>>;
 
-    pub fn children_ref(&self) -> &RwLock<BTreeMap<SysStr, Arc<C>>>;
+    pub fn children_ref(&self) -> &RwMutex<BTreeMap<SysStr, Arc<C>>>;
 
     pub fn attr_set(&self) -> &SysAttrSet {
         &self.attr_set
--- a/kernel/src/fs/cgroupfs/systree_node.rs
+++ b/kernel/src/fs/cgroupfs/systree_node.rs
@@ -22,7 +22,7 @@
 //!    `inner` data.
 //!
 //! 3. **Children Lock**: Each cgroup node ([`CgroupNode`] and [`CgroupSystem`]) inherits
-//!    an `RwLock` from `BranchNodeFields`. This lock protects access to and
+//!    an `RwMutex` from `BranchNodeFields`. This lock protects access to and
 //!    modification of the list of child cgroup nodes.
 //!
 //! 4. **Cgroup Membership Lock**: A global `Mutex` managed by [`CgroupMembership`] that


kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +100,26 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
+        let mut buffer = [0u8; MAX_BUFFER_SIZE];
+        if len > buffer.len() {
+            return_errno_with_message!(Errno::EINVAL, "buffer is too small");
+        }
+        dirent_serializer.serialize(&mut buffer[..len])?;
+
+        let mut reader = VmReader::from(&buffer[..len]);
+        self.writer
+            .write_fallible(&mut reader)
+            .map_err(|_| Errno::EFAULT)?;
```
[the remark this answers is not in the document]
Marsman1996: `return_errno_with_message ` cannot format output, maybe I shall return the `err` directly?

kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +98,32 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
 
 impl<T: DirentSerializer> DirentVisitor for DirentBufferReader<'_, T> {
     fn visit(&mut self, name: &str, ino: u64, type_: InodeType, offset: usize) -> Result<()> {
+        /// The maximum bytes in a file name is 255 in fs/utils/mod.rs
+        const MAX_BUFFER_SIZE: usize = 512;
```
StevenJiang1110: Could we use the `NAME_MAX` constant from fs/utils/mod.rs directly? If you're concerned about a potential null terminator, we could define `MAX_BUFFER_SIZE` as `NAME_MAX + 1`.

kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +98,32 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
+                "user buffer has insufficient space for directory entry"
+            );
+        }
+
+        let mut buffer = [0u8; MAX_BUFFER_SIZE];
+        if len > buffer.len() {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "directory entry exceeds maximum supported size"
+            );
```
StevenJiang1110, line 109: Move the validation logic from lines 113–118 to before line 112. This ensures we don't allocate the buffer if the check fails.

```rust
if len > MAX_BUFFER_LEN {
    // ......
}
```

kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +98,32 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
-            return_errno_with_message!(Errno::EINVAL, "buffer is too small");
+        let len = dirent_serializer.len();
+        if self.writer.avail() < len {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "user buffer has insufficient space for directory entry"
+            );
+        }
+
+        let mut buffer = [0u8; MAX_BUFFER_SIZE];
```
StevenJiang1110: The buffer's length should be `len` instead of `MAX_BUFFER_LEN`?

kernel/src/syscall/getdents64.rs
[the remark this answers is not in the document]
StevenJiang1110: The current `map_err` logic is fine as is.

kernel/src/syscall/getdents64.rs
Marsman1996: `NAME_MAX + 1` is not enough for the buffer, see `serialize` in `DirentSerializer for Dirent64`, it holds not only the file name, but also the struct.

kernel/src/syscall/getdents64.rs
Marsman1996: Only const value is allowed here.

kernel/src/syscall/getdents64.rs
Marsman1996: fixed

kernel/src/syscall/getdents64.rs
Marsman1996: How about `NAME_MAX + 1 + size_of::<Dirent64>()` 🤔

kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +98,32 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
         }
-        dirent_serializer.serialize(&mut self.buffer[self.read_len..])?;
-        self.read_len += dirent_serializer.len();
+        let mut buffer = [0u8; MAX_BUFFER_SIZE];
+        dirent_serializer.serialize(&mut buffer[..len])?;
+
+        let mut reader = VmReader::from(&buffer[..len]);
+        self.writer
+            .write_fallible(&mut reader)
+            .map_err(|_| Errno::EFAULT)?;
```
lrh2000:
```suggestion
            .write_fallible(&mut reader)?;
```

kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +98,32 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "user buffer has insufficient space for directory entry"
+            );
+        }
+
+        if len > MAX_BUFFER_SIZE {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "directory entry exceeds maximum supported size"
```
lrh2000: This is a kernel problem?

Instead of doing this, I think we should fix the `serialize` method to accept `VmWriter`. Why can't we do that?

Additionally, it appears that this file was written in the early stages of our project. I suggest the following cleanup:
 - Remove the `align_up` method in favor of the `AlignUp` trait.
https://github.com/asterinas/asterinas/blob/8fbf34f8ae24e5f8715fcc939f99ad62fb253d23/kernel/src/syscall/getdents64.rs#L282C4-L284
 - Use the `Pod` trait whenever possible instead of doing the following.
https://github.com/asterinas/asterinas/blob/8fbf34f8ae24e5f8715fcc939f99ad62fb253d23/kernel/src/syscall/getdents64.rs#L235-L239

lrh2000:
This will cause atomic mode break because:
https://github.com/asterinas/asterinas/blob/bcf8f5e93c5898f38cd8c8dee2c35eab6abff49b/kernel/comps/systree/src/utils.rs#L111

We cannot write to user space while holding a spin lock (`RwLock`).

Marsman1996:
> This will cause atomic mode break because:
> 
> https://github.com/asterinas/asterinas/blob/bcf8f5e93c5898f38cd8c8dee2c35eab6abff49b/kernel/comps/systree/src/utils.rs#L111
> 
> We cannot write to user space while holding a spin lock (`RwLock`).

How about just using `Vec::try_with_capacity()` to replace the `vec![0u8; buf_len];`

kernel/src/syscall/getdents64.rs
StevenJiang1110: Emmm, so `dirent_serializer.len()` cannot exceed `NAME_MAX + 1 + size_of::<Dirent64>()`? If so, I guess the code can be simplified. We don't need to compare `len` vs `MAX_BUFFER_LEN`. 

```rust
fn visit(&mut self, name: &str, ino: u64, type_: InodeType, offset: usize) -> Result<()> {
        let dirent_serializer = T::new(ino, offset as u64, type_, CString::new(name)?);
        let len = dirent_serializer.len();
        if self.writer.avail() < len {
            return_errno_with_message!(
                Errno::EINVAL,
                "user buffer has insufficient space for directory entry"
            );
        }

        // The allocation must be bounded.
        debug_assert!(len <= NAME_MAX + 1 + size_of::<Dirent64>());
        let mut buffer = vec![0u8; len];
        dirent_serializer.serialize(&mut buffer)?;
        // .......
}
```

But you may want to deal with the problem mentioned in https://github.com/asterinas/asterinas/pull/2791#issuecomment-3690978705 at first. It may affect your design choice.

kernel/src/syscall/getdents64.rs
```diff
@@ -102,12 +98,32 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
+        /// The maximum bytes in a file name is 255 in fs/utils/mod.rs
+        const MAX_BUFFER_SIZE: usize = 512;
         let dirent_serializer = T::new(ino, offset as u64, type_, CString::new(name)?);
-        if self.read_len >= self.buffer.len() {
-            return_errno_with_message!(Errno::EINVAL, "buffer is too small");
+        let len = dirent_serializer.len();
+        if self.writer.avail() < len {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "user buffer has insufficient space for directory entry"
```
lrh2000:
```suggestion
                "the buffer is too small for the directory entry"
```

lrh2000:
> How about just using `Vec::try_with_capacity()` to replace the `vec![0u8; buf_len];`

As I mentioned in your previous PR, it doesn't solve the actual problem; it just makes it harder to exploit. I am hesitant to introduce the `try_with_capacity` pattern before we determine the correct way to handle out-of-memory errors in Rust data structures (e.g., `BTreeMap`...).

Marsman1996:
> > How about just using `Vec::try_with_capacity()` to replace the `vec![0u8; buf_len];`
> 
> As I mentioned in your previous PR, it doesn't solve the actual problem; it just makes it harder to exploit. I am hesitant to introduce the `try_with_capacity` pattern before we determine the correct way to handle out-of-memory errors in Rust data structures (e.g., `BTreeMap`...).

In my previous PR you mentioned that _A malicious user could still exploit issue in an attempt to exhaust almost all of the kernel's memory._ 
But what if a malicious user attempts to allocate small memory regions multiple times, with the goal of exhausting nearly all of the kernel's memory?

lrh2000:
> But what if a malicious user attempts to allocate small memory regions multiple times, with the goal of exhausting nearly all of the kernel's memory?

It is limited to the "maximum number of threads" $\times$ "maximum amount of kernel memory that a system call can allocate" because you cannot invoke two system calls simultaneously within one thread.

The maximum number of threads is limited by a system restriction [^1]. On top of that, we must ensure that no system call can allocate unbound memory.

[^1]: Otherwise, if an user could create an infinite number of processes or threads, they could exhaust all system resources by using `fork()` or `clone()`.

kernel/src/syscall/getdents64.rs
Marsman1996: Updated

kernel/src/syscall/getdents64.rs
Marsman1996: With `VmWriter`, no more `MAX_BUFFER_LEN ` here

kernel/src/syscall/getdents64.rs
```diff
@@ -223,26 +205,18 @@ impl DirentSerializer for Dirent64 {
...
+
+        let name_bytes = self.name.as_c_str().to_bytes_with_nul();
+        let mut reader = VmReader::from(name_bytes);
+        writer.write_fallible(&mut reader)?;
+        let written_len = size_of::<Dirent64Inner>() + name_bytes.len();
+        let padding_len = self.len() - written_len;
+        if padding_len > 0 {
+            let zeros = [0u8; 8];
+            let mut reader = VmReader::from(&zeros[..padding_len]);
+            writer.write_fallible(&mut reader)?;
```
lrh2000: Use `write.fill_zeros(padding_len)`?

kernel/src/syscall/getdents64.rs
```diff
@@ -223,26 +205,18 @@ impl DirentSerializer for Dirent64 {
...
-        for item in items {
-            buf[offset..offset + item.len()].copy_from_slice(item);
-            offset += item.len();
+    fn serialize(&self, writer: &mut VmWriter) -> Result<()> {
+        writer.write_val(&self.inner)?;
+
+        let name_bytes = self.name.as_c_str().to_bytes_with_nul();
+        let mut reader = VmReader::from(name_bytes);
+        writer.write_fallible(&mut reader)?;
+        let written_len = size_of::<Dirent64Inner>() + name_bytes.len();
```
lrh2000, line 212: Insert a blank line before line 214.

It is also preferable to insert a blank line before the final `Ok(())` in the two `serialize()` methods.

lrh2000 (commented):
@cchanging I think the spin lock mentioned in https://github.com/asterinas/asterinas/pull/2791#issuecomment-3690978705 has caused too many problems for us.... Should we change it to a mutex? Would that cause problems? Are you aware of any similar problems in other FS implementations?

---

I also expect this PR to contain implementation details about Linux that justify the change, since it affects how the FS code is written. I've had a quick look and I think it's the right approach:
https://elixir.bootlin.com/linux/v6.18.2/source/fs/readdir.c#L108
https://elixir.bootlin.com/linux/v6.18.2/source/fs/readdir.c#L181

lrh2000:
If we cannot change the lock to a mutex, then we can fix the sysfs `readdir` implementation so that it releases the lock before invoking the callback? I think it's possible, though it's somewhat complex (and unnecessary?)..

kernel/src/syscall/getdents64.rs
Marsman1996: Updated

cchanging:
> If we cannot change the lock to a mutex

I checked the current usages of `children` fileds, and I think changing it to `mutex` will not cause problems, at least for now...I think this lock shouldn't be fundamentally much different from other locks in fs-related structs, and there's no issue with using a `mutex`.

kernel/src/syscall/getdents64.rs
```diff
@@ -103,11 +101,17 @@ impl<'a, T: DirentSerializer> DirentBufferReader<'a, T> {
...
+                "the buffer is too small for the directory entry"
+            );
         }
-        dirent_serializer.serialize(&mut self.buffer[self.read_len..])?;
-        self.read_len += dirent_serializer.len();
+
+        dirent_serializer.serialize(&mut self.writer)?;
+
+        self.read_len += len;
         Ok(())
```
lrh2000, line 115:
```suggestion
        dirent_serializer.serialize(&mut self.writer)?;
        self.read_len += len;

        Ok(())
```

kernel/src/syscall/getdents64.rs
```diff
@@ -146,26 +150,19 @@ impl DirentSerializer for Dirent {
...
-            d_reclen.as_bytes(),
-            self.name.as_c_str().to_bytes_with_nul(),
-        ];
-        let mut offset = 0;
-        for item in items {
-            buf[offset..offset + item.len()].copy_from_slice(item);
-            offset += item.len();
+        let written_len = size_of::<DirentInner>() + name_bytes.len();
+        let padding_len = self.len() - written_len;
+        if padding_len > 0 {
```
lrh2000: It seems that `fill_zeros` will check if the length is zero anyway. Do we still need to check it again?
https://github.com/asterinas/asterinas/blob/1aff7d03d19cdc921d4c95d86f8b7958635323e3/ostd/src/mm/io.rs#L790-L793

lrh2000:
> I checked the current usages of `children` fileds, and I think changing it to `mutex` will not cause problems, at least for now...I think this lock shouldn't be fundamentally much different from other locks in fs-related structs, and there's no issue with using a `mutex`.

Then maybe it's the time to change it..

Before invoking `iterate_shared`, Linux [acquires a read semaphore on the inode](https://elixir.bootlin.com/linux/v6.18.2/source/fs/readdir.c#L101-L103). Then, for most file systems, `iterate_shared` does not take additional locks internally. For kernfs, however, there is [an additional read semaphore inside](https://elixir.bootlin.com/linux/v6.18.2/source/fs/kernfs/dir.c#L1893).

Therefore, it seems reasonable to us to hold a mutex when writing directory entries back to user space.

kernel/src/syscall/getdents64.rs
Marsman1996: You are absolutely right! No need for this duplicate check.

Marsman1996:
@boterinas rerun

lrh2000:
The first commit LGTM. In the second commit, use a `Mutex` or an `RwMutex`?

Maybe the problem is that I didn't use the correct terminology earlier (sorry if it was the case)? However, the entire discussion that occurred in this PR should not be related to `Rw` at all? Especially, we found that the Linux implementation uses RW semaphore in https://github.com/asterinas/asterinas/pull/2791#issuecomment-3691322781.

---

In addition, please start the commit message with "Replace RwLock with..." to align with current practices.

Marsman1996:
@boterinas rerun

PR #2615, opened 2025-11-25
CortenMM

junyang-zh:
Implements #2792

---

Dependencies:
 - #2351
 - #2744
 - #2595
 - #2605
 - #2613
 
 Conflicts with this PR and can be resolved in either way:
 - #2766
 - #2736
 - #2157

Status:

 - [x] fork
 - [x] map
 - [x] page fault
 - [x] protect
 - [x] query
 - [x] remap
 - [x] unmap
 - [x] access_remote
 - [x] ~~Fix mmap races by synchronizing in the allocator~~
 - [x] Fix for tests in #2766
 - [x] Remove the allocator to avoid races
 - [x] make correct the rlimit
 - [x] optimize unfixed map allocation

PR #2792, opened 2025-12-24
Add RFC-0003 CortenMM

junyang-zh:
# RFC-0003: CortenMM

* Status: Draft
* Pull request: https://github.com/asterinas/asterinas/pull/2792
* Date submitted: 2025-12-24
* Date approved: YYYY-MM-DD

## Summary

This RFC introduces CortenMM[^1]’s single-level, page table–centric virtual memory management in Asterinas. Instead of maintaining a separate [`VmMapping`](https://github.com/asterinas/asterinas/blob/459a634e4acc865b4bad99116a416910835083e5/kernel/src/vm/vmar/vm_mapping.rs#L51) tree (which is equivalent to [Linux's Virtual Memory Area](https://github.com/torvalds/linux/blob/fcb70a56f4d81450114034b2c61f48ce7444a0e2/include/linux/mm_types.h#L904)) plus page tables, we split mappings on page table boundaries so that mapping metadata can be attached to page tables. This yields better scalability, simpler invariants, and stronger correctness properties while preserving Linux-compatible semantics.

## Motivation

Two-level VM designs (e.g., VMA trees and page tables) hurt scalability and are a recurring source of subtle concurrency bugs; Linux continues to exhibit `mmap_lock`/VMA contention and correctness CVEs. Our paper, CortenMM, addresses these issues and has gained substantial positive feedback from the OS research community.

The CortenMM paper shows that mainstream ISAs all use radix page tables, so we can collapse to a single level without losing portability. By colocating mapping metadata with page tables, we eliminate cross-structure synchronization and reduce cache/lock traffic. The paper reports 1.2×–26× speedups on real workloads; we expect similar gains for Asterinas user space and kernel services.

A single-level abstraction simplifies reasoning and enables stronger correctness: invariants are local to page table subtrees; our merged locking protocol already enforces serialized updates per covered range. This reduces risk as we continue to develop for production and broaden hardware coverage.

## Design

This RFC proposes a full replacement of the existing two-level VM system with CortenMM’s single-level design. The design largely follows the CortenMM paper but uses a slightly different metadata storage strategy for compatibility and memory-efficiency reasons.

### High-level structure

The page table is the source of truth; per-page table auxiliary metadata stores `VmMapping`s that cover that page table frame's range. No global VMA tree is required.

`VmMapping`s are split at page table boundaries when inserted and each PT node only carries mappings aligned to its range. This keeps lookups and updates local and mergeable.

For example, this is the page table after `mmap(offset=0x3fe000, size=0x604000)`. A new $2 \text{MiB} \times 3 + 4 \text{KiB} \times 4$ `VmMapping` is split into three pieces (`[x]`, `[y]`, and `[z]`) and stored across two levels:

```
                   .----------------.
L4                 |      #4-1      |
                   '.---------------'
                   /
                  v
                 .----------------.
L3               |      #3-1      |
                 '.---------------'
                 /
                v
               .----------------.
L2             |      #2-1      |
               '-.---------.----'
      .----------'   [y]   |
      v                    v
      .----------------.   .----------------.
L1    |      #1-2      |   |      #1-6      |
      '----------------'   '----------------'
                     [x]   [z]

Pieces stored:
    0x3fe000       0x400000               0xa00000      0xa02000
    [      x      ][          y          ][     z      ]
      at #1-2 PT         at #2-1 PT         at #1-6 PT
```

In the figure, page table frame `#3-1` means the first page table frame (the page table frame that manages the lowest virtual address) in the third level.

If a page fault happens at `0x600000`, the handler locks the `#2-1` PT node, splits `[y]` into `[a]`, `[b]`, and `[c]` at the `#1-4` PT boundary, and maps the page.

```
                   .----------------.
L4                 |      #4-1      |
                   '.---------------'
                   /
                  v
                 .----------------.
L3               |      #3-1      |
                 '.---------------'
                 /
                v
               .----------------.
L2             |      #2-1      |
               '-.----.----.----'
     .-----------'[a] | [c]'-----------------.
     v                '--v                   v
     .----------------.  .----------------.  .----------------.
L1   |      #1-2      |  |      #1-4      |  |      #1-6      |
     '----------------'  '----------------'  '----------------'
                    [x]  [        b       ]  [z]

Pieces stored:
   0x3fe000    0x400000        0x600000        0x800000        0xa00000    0xa02000
   [     x    ][       a      ][       b      ][      c       ][     z    ]
    at #1-2 PT    at #2-1 PT      at #1-4 PT      at #2-1 PT    at #1-6 PT
```

Within each PT node, the `VmMapping`s are stored in an interval set (e.g., a `BTreeMap` keyed by range) to support efficient lookups, splits, merges, and removals.

The page table locking protocol described in the paper is already present in Asterinas (cursor-based per-range locking over PT pages). This RFC focuses on the data-structure shift and metadata placement; the locking protocol remains unchanged.

Upper kernel components access the page table via the same OSTD cursor APIs as before, with minor adjustments to define and access metadata in PT-local storage.

Other optimizations introduced in the paper for full scalability (LATR[^4], MCS locks) are out of scope for this RFC but can be pursued later.

The following sections describe the key implementations of existing memory management system calls.

### Address space allocation for `mmap`/`mremap`

There are two kinds of `mmap`/`mremap` calls, depending on the `MAP_FIXED`/`MREMAP_FIXED` flag. Without this flag, non-fixed `mmap`/`mremap` allocates a new address space range. With this flag, fixed `mmap`/`mremap` overwrites an existing address space range.

According to the CortenMM paper, an operation starts with a cursor that locks the given range. However, non-fixed `mmap`/`mremap` does not know the range yet, so it must find a free range first. To do this, non-fixed `mmap`/`mremap` first locks a large range of size $\dfrac{1}{N}$ (where $N$ is the number of CPUs) and tries to find a free range inside it according to the auxiliary metadata stored in the page tables. If it fails, it unlocks the range and retries with a different range (or the entire address space) until it finds a free range. On the other hand, fixed `mmap`/`mremap` directly locks the given range.

With this design, both fixed and non-fixed `mmap`/`mremap` can be implemented scalably with the same cursor-based APIs.

### Dealing with `RLIMIT_AS`

After scaling `mmap`/`mremap`, the resource limit `RLIMIT_AS` (address space size limit) becomes a scalability bottleneck if the total size of all mappings is tracked in a single atomic variable. To address this, we track CPU-local quotas plus a global limit instead of a single global counter. If a CPU-local quota is exhausted, we borrow from other CPUs' quotas. When the limit is unlikely to be reached, most `mmap`/`mremap` calls proceed without global synchronization.

### Traversing the page table

With auxiliary metadata, page table traversal is more efficient. We can provide APIs that consult the `VmMapping` set rather than the page table entries, enabling faster `fork`s.

### New OSTD APIs

OSTD must not be aware of the auxiliary metadata type. We introduce a new `AuxPageTableMeta` trait that allows users to define and access auxiliary metadata in page table nodes. The `VmMapping` set is defined as the auxiliary metadata type. The trait is:

```rust
/// Auxiliary metadata for user page tables.
pub trait AuxPageTableMeta: AuxPtMetaLayoutChecked + Debug + Send + Sync + 'static {}
```

The user-defined metadata type will be provided to OSTD via `VmSpace`'s new generic parameter `VmSpace<A: AuxPageTableMeta>`. The `()` type also implements the `AuxPageTableMeta` trait as a no-op implementation for backward compatibility. The `VmSpace` will instantiate the page table with the `Aux` type in the `PageTableConfig` trait specified as the user-defined type.

Furthermore, the cursor user must be aware of the level of the cursor, rather than only the virtual address. Therefore new APIs to get and adjust the current page table level are added to the cursor.

## Sleeping page table locks

Currently the kernel cannot sleep while a cursor is holding page table locks. This is because the OSTD implementation currently uses RCU to reason about the safety of accessing page tables. When operations (e.g., updating reverse mappings, doing `madvise`, or invoking the driver) need to break atomic mode, we have to drop the locks and retry, which does not guarantee stable mappings and may lead to inconsistent behavior. We are doing such retries upon faulting the file-backed mappings.

Furthermore, after merging the page table locks and the address space locks, the page table operations are heavier than before, which also violates the "use spin locks for short critical sections" principle.

Nevertheless, the non-sleeping requirement for page table locks is not fundamental: we can change the locks to sleepable mutexes. Only the lock protocol's traversal phase is RCU-critical. After locking the root of the child page table, we ensure that the subtree is alive and the RCU critical section should end. Scheduling can happen afterward.

## Reverse mappings

Although current Asterinas lacks reverse mappings, CortenMM must not preclude reverse mapping support. File-backed reverse mappings are needed for protecting mappings as read-only when writing back pages; anonymous reverse mappings are needed for swapping. Some device drivers (e.g., GPU) also demand reverse mappings for device-backed memory pages.

### File/device-backed reverse mappings

Like Linux's `i_mmap` tree in each `inode`, Asterinas can keep all the mapping address spaces in the `Vmo` object (protected by a per-`Vmo` reverse mapping lock). Each `Vmo` maintains a `BTreeMap` that stores the virtual offset and the range in each mapping memory space.

The challenge of adopting this reverse mapping design in CortenMM is the lock ordering: operations like `mremap` need to traverse the page table to get all the reverse mapping locks; operations like `fsync` need to traverse the reverse mapping set to get all the page tables.

The solution is to retry. We keep the forward lock ordering: to acquire the reverse mapping lock, the thread must first acquire the page table lock. When doing the reverse operations, if the page table lock cannot be acquired with `try_lock`, the thread must drop the reverse mapping lock and retry the traversal of reverse mappings.

File-backed and device-backed mappings share similar solutions. If there's a driver that needs to acquire its locks, it does after the page table locks, before the reverse-mappings locks.

### Anonymous reverse mappings

Anonymous reverse mappings, unlike file-backed reverse mappings, have different problems and need different solutions. Each physical page in the anonymous mapping can be easily shared by `fork` and unshared by copy-on-write independently. Therefore, each physical page can have a distinct set of virtual memory spaces that map to it. To address this, Linux links each physical page to one `anon_vma`, and all the VMAs in the system have a bipartite graph relationship with all the `anon_vma`s. Reverse mapping is done by traversing this bipartite graph.

Instead, CortenMM's page table metadata is owned solely by a page table. Sharing it would cause complications. As a trade-off, CortenMM links each `anon_vma`-like object with all the memory spaces instead of page table metadata ranges. This approach still maintains `O(1)` page fault and `fork` overhead. The downside is that when doing reverse mapping traversal, there is a chance of iterating all the irrelevant memory spaces before finding the actual ones. This is because, unlike Linux, upon `munmap` of an entire VMA, the affected `anon_vma` cannot be unlinked from parent memory spaces (it can only be unlinked in address-space-wide unmapping like `exec`), resulting in a larger number of memory spaces to search.

## Drawbacks, Alternatives, and Unknown

**Drawbacks:**
 - Slightly higher metadata footprint when mappings straddle many PT boundaries (mitigated by merging to upper-level page table frames that manage larger ranges).
 - Portability to non-radix MMUs would require a different scheme; this RFC targets x86_64/RISC-V/ARM.

**Alternatives:**
 - Keep the existing two-level design and place CortenMM under a feature gate `#[cfg(feature = "cortenmm")]`. This would allow opt-in testing but complicates maintenance and feature development.
 - Keep two-level design but optimize VMA management (e.g., via RCU-safe maple trees[^3] or concurrent skip lists[^2]). This would improve scalability but not eliminate cross-structure synchronization or correctness risks. These scalable data structures are also complex and require `unsafe` code, which bloats the TCB and increases maintenance burden.

**Unresolved Questions / follow-ups:**
 - With only 512 entries per page table, we can optimize the auxiliary metadata layout to avoid deep BTreeMap levels.
 - Implementing efficient anonymous reverse mappings without sacrificing scalability is difficult. Since Asterinas does not support swapping at the moment, we leave efficient anonymous reverse mappings in CortenMM as future work.

## Prior Art and References

[RadixVM: scalable address spaces for multithreaded applications](https://dl.acm.org/doi/10.1145/2465351.2465373) provides another way to unify abstractions. It makes the software-level abstraction a radix tree, similar to hardware page tables, and makes the page tables cache the radix-tree nodes. This design is effective in terms of scalability, but it is not suitable for Asterinas because the software-level abstraction cannot be implemented in safe code, and its Linux compatibility is limited.

[^1]: [CortenMM: Efficient Memory Management with Strong Correctness Guarantees](https://doi.org/10.1145/3731569.3764836)

[^2]: [Scalable Address Spaces using Concurrent Interval Skiplist](https://dl.acm.org/doi/10.1145/3731569.3764807)

[^3]: [Concurrent page-fault handling with per-VMA locks](https://lore.kernel.org/all/20220901173516.702122-1-surenb@google.com/)

[^4]: [LATR: Lazy Translation Coherence](https://dl.acm.org/doi/10.1145/3296957.3173198)

### DIFF
--- a/book/src/rfcs/0003-cortenmm.md
+++ b/book/src/rfcs/0003-cortenmm.md
@@ -1,0 +1,181 @@
+# RFC-0003: CortenMM
+
+* Status: Draft
+* Pull request: https://github.com/asterinas/asterinas/pull/2792
+* Date submitted: 2025-12-24
+* Date approved: YYYY-MM-DD
+
+## Summary
+
+This RFC introduces CortenMM[^1]’s single-level, page table–centric virtual memory management in Asterinas. Instead of maintaining a separate [`VmMapping`](https://github.com/asterinas/asterinas/blob/459a634e4acc865b4bad99116a416910835083e5/kernel/src/vm/vmar/vm_mapping.rs#L51) tree (which is equivalent to [Linux's Virtual Memory Area](https://github.com/torvalds/linux/blob/fcb70a56f4d81450114034b2c61f48ce7444a0e2/include/linux/mm_types.h#L904)) plus page tables, we split mappings on page table boundaries so that mapping metadata can be attached to page tables. This yields better scalability, simpler invariants, and stronger correctness properties while preserving Linux-compatible semantics.
+
+## Motivation
+
+Two-level VM designs (e.g., VMA trees and page tables) hurt scalability and are a recurring source of subtle concurrency bugs; Linux continues to exhibit `mmap_lock`/VMA contention and correctness CVEs. Our paper, CortenMM, addresses these issues and has gained substantial positive feedback from the OS research community.
+
+The CortenMM paper shows that mainstream ISAs all use radix page tables, so we can collapse to a single level without losing portability. By colocating mapping metadata with page tables, we eliminate cross-structure synchronization and reduce cache/lock traffic. The paper reports 1.2×–26× speedups on real workloads; we expect similar gains for Asterinas user space and kernel services.
+
+A single-level abstraction simplifies reasoning and enables stronger correctness: invariants are local to page table subtrees; our merged locking protocol already enforces serialized updates per covered range. This reduces risk as we continue to develop for production and broaden hardware coverage.
+
+## Design
+
+This RFC proposes a full replacement of the existing two-level VM system with CortenMM’s single-level design. The design largely follows the CortenMM paper but uses a slightly different metadata storage strategy for compatibility and memory-efficiency reasons.
+
+### High-level structure
+
+The page table is the source of truth; per-page table auxiliary metadata stores `VmMapping`s that cover that page table frame's range. No global VMA tree is required.
+
+`VmMapping`s are split at page table boundaries when inserted and each PT node only carries mappings aligned to its range. This keeps lookups and updates local and mergeable.
+
+For example, this is the page table after `mmap(offset=0x3fe000, size=0x604000)`. A new $2 \text{MiB} \times 3 + 4 \text{KiB} \times 4$ `VmMapping` is split into three pieces (`[x]`, `[y]`, and `[z]`) and stored across two levels:
+
+```
+                   .----------------.
+L4                 |      #4-1      |
+                   '.---------------'
+                   /
+                  v
+                 .----------------.
+L3               |      #3-1      |
+                 '.---------------'
+                 /
+                v
+               .----------------.
+L2             |      #2-1      |
+               '-.---------.----'
+      .----------'   [y]   |
+      v                    v
+      .----------------.   .----------------.
+L1    |      #1-2      |   |      #1-6      |
+      '----------------'   '----------------'
+                     [x]   [z]
+
+Pieces stored:
+    0x3fe000       0x400000               0xa00000      0xa02000
+    [      x      ][          y          ][     z      ]
+      at #1-2 PT         at #2-1 PT         at #1-6 PT
+```
+
+In the figure, page table frame `#3-1` means the first page table frame (the page table frame that manages the lowest virtual address) in the third level.
+
+If a page fault happens at `0x600000`, the handler locks the `#2-1` PT node, splits `[y]` into `[a]`, `[b]`, and `[c]` at the `#1-4` PT boundary, and maps the page.
+
+```
+                   .----------------.
+L4                 |      #4-1      |
+                   '.---------------'
+                   /
+                  v
+                 .----------------.
+L3               |      #3-1      |
+                 '.---------------'
+                 /
+                v
+               .----------------.
+L2             |      #2-1      |
+               '-.----.----.----'
+     .-----------'[a] | [c]'-----------------.
+     v                '--v                   v
+     .----------------.  .----------------.  .----------------.
+L1   |      #1-2      |  |      #1-4      |  |      #1-6      |
+     '----------------'  '----------------'  '----------------'
+                    [x]  [        b       ]  [z]
+
+Pieces stored:
+   0x3fe000    0x400000        0x600000        0x800000        0xa00000    0xa02000
+   [     x    ][       a      ][       b      ][      c       ][     z    ]
+    at #1-2 PT    at #2-1 PT      at #1-4 PT      at #2-1 PT    at #1-6 PT
+```
+
+Within each PT node, the `VmMapping`s are stored in an interval set (e.g., a `BTreeMap` keyed by range) to support efficient lookups, splits, merges, and removals.
+
+The page table locking protocol described in the paper is already present in Asterinas (cursor-based per-range locking over PT pages). This RFC focuses on the data-structure shift and metadata placement; the locking protocol remains unchanged.
+
+Upper kernel components access the page table via the same OSTD cursor APIs as before, with minor adjustments to define and access metadata in PT-local storage.
+
+Other optimizations introduced in the paper for full scalability (LATR[^4], MCS locks) are out of scope for this RFC but can be pursued later.
+
+The following sections describe the key implementations of existing memory management system calls.
+
+### Address space allocation for `mmap`/`mremap`
+
+There are two kinds of `mmap`/`mremap` calls, depending on the `MAP_FIXED`/`MREMAP_FIXED` flag. Without this flag, non-fixed `mmap`/`mremap` allocates a new address space range. With this flag, fixed `mmap`/`mremap` overwrites an existing address space range.
+
+According to the CortenMM paper, an operation starts with a cursor that locks the given range. However, non-fixed `mmap`/`mremap` does not know the range yet, so it must find a free range first. To do this, non-fixed `mmap`/`mremap` first locks a large range of size $\dfrac{1}{N}$ (where $N$ is the number of CPUs) and tries to find a free range inside it according to the auxiliary metadata stored in the page tables. If it fails, it unlocks the range and retries with a different range (or the entire address space) until it finds a free range. On the other hand, fixed `mmap`/`mremap` directly locks the given range.
+
+With this design, both fixed and non-fixed `mmap`/`mremap` can be implemented scalably with the same cursor-based APIs.
+
+### Dealing with `RLIMIT_AS`
+
+After scaling `mmap`/`mremap`, the resource limit `RLIMIT_AS` (address space size limit) becomes a scalability bottleneck if the total size of all mappings is tracked in a single atomic variable. To address this, we track CPU-local quotas plus a global limit instead of a single global counter. If a CPU-local quota is exhausted, we borrow from other CPUs' quotas. When the limit is unlikely to be reached, most `mmap`/`mremap` calls proceed without global synchronization.
+
+### Traversing the page table
+
+With auxiliary metadata, page table traversal is more efficient. We can provide APIs that consult the `VmMapping` set rather than the page table entries, enabling faster `fork`s.
+
+### New OSTD APIs
+
+OSTD must not be aware of the auxiliary metadata type. We introduce a new `AuxPageTableMeta` trait that allows users to define and access auxiliary metadata in page table nodes. The `VmMapping` set is defined as the auxiliary metadata type. The trait is:
+
+```rust
+/// Auxiliary metadata for user page tables.
+pub trait AuxPageTableMeta: AuxPtMetaLayoutChecked + Debug + Send + Sync + 'static {}
+```
+
+The user-defined metadata type will be provided to OSTD via `VmSpace`'s new generic parameter `VmSpace<A: AuxPageTableMeta>`. The `()` type also implements the `AuxPageTableMeta` trait as a no-op implementation for backward compatibility. The `VmSpace` will instantiate the page table with the `Aux` type in the `PageTableConfig` trait specified as the user-defined type.
+
+Furthermore, the cursor user must be aware of the level of the cursor, rather than only the virtual address. Therefore new APIs to get and adjust the current page table level are added to the cursor.
+
+## Sleeping page table locks
+
+Currently the kernel cannot sleep while a cursor is holding page table locks. This is because the OSTD implementation currently uses RCU to reason about the safety of accessing page tables. When operations (e.g., updating reverse mappings, doing `madvise`, or invoking the driver) need to break atomic mode, we have to drop the locks and retry, which does not guarantee stable mappings and may lead to inconsistent behavior. We are doing such retries upon faulting the file-backed mappings.
+
+Furthermore, after merging the page table locks and the address space locks, the page table operations are heavier than before, which also violates the "use spin locks for short critical sections" principle.
+
+Nevertheless, the non-sleeping requirement for page table locks is not fundamental: we can change the locks to sleepable mutexes. Only the lock protocol's traversal phase is RCU-critical. After locking the root of the child page table, we ensure that the subtree is alive and the RCU critical section should end. Scheduling can happen afterward.
+
+## Reverse mappings
+
+Although current Asterinas lacks reverse mappings, CortenMM must not preclude reverse mapping support. File-backed reverse mappings are needed for protecting mappings as read-only when writing back pages; anonymous reverse mappings are needed for swapping. Some device drivers (e.g., GPU) also demand reverse mappings for device-backed memory pages.
+
+### File/device-backed reverse mappings
+
+Like Linux's `i_mmap` tree in each `inode`, Asterinas can keep all the mapping address spaces in the `Vmo` object (protected by a per-`Vmo` reverse mapping lock). Each `Vmo` maintains a `BTreeMap` that stores the virtual offset and the range in each mapping memory space.
+
+The challenge of adopting this reverse mapping design in CortenMM is the lock ordering: operations like `mremap` need to traverse the page table to get all the reverse mapping locks; operations like `fsync` need to traverse the reverse mapping set to get all the page tables.
+
+The solution is to retry. We keep the forward lock ordering: to acquire the reverse mapping lock, the thread must first acquire the page table lock. When doing the reverse operations, if the page table lock cannot be acquired with `try_lock`, the thread must drop the reverse mapping lock and retry the traversal of reverse mappings.
+
+File-backed and device-backed mappings share similar solutions. If there's a driver that needs to acquire its locks, it does after the page table locks, before the reverse-mappings locks.
+
+### Anonymous reverse mappings
+
+Anonymous reverse mappings, unlike file-backed reverse mappings, have different problems and need different solutions. Each physical page in the anonymous mapping can be easily shared by `fork` and unshared by copy-on-write independently. Therefore, each physical page can have a distinct set of virtual memory spaces that map to it. To address this, Linux links each physical page to one `anon_vma`, and all the VMAs in the system have a bipartite graph relationship with all the `anon_vma`s. Reverse mapping is done by traversing this bipartite graph.
+
+Instead, CortenMM's page table metadata is owned solely by a page table. Sharing it would cause complications. As a trade-off, CortenMM links each `anon_vma`-like object with all the memory spaces instead of page table metadata ranges. This approach still maintains `O(1)` page fault and `fork` overhead. The downside is that when doing reverse mapping traversal, there is a chance of iterating all the irrelevant memory spaces before finding the actual ones. This is because, unlike Linux, upon `munmap` of an entire VMA, the affected `anon_vma` cannot be unlinked from parent memory spaces (it can only be unlinked in address-space-wide unmapping like `exec`), resulting in a larger number of memory spaces to search.
+
+## Drawbacks, Alternatives, and Unknown
+
+**Drawbacks:**
+ - Slightly higher metadata footprint when mappings straddle many PT boundaries (mitigated by merging to upper-level page table frames that manage larger ranges).
+ - Portability to non-radix MMUs would require a different scheme; this RFC targets x86_64/RISC-V/ARM.
+
+**Alternatives:**
+ - Keep the existing two-level design and place CortenMM under a feature gate `#[cfg(feature = "cortenmm")]`. This would allow opt-in testing but complicates maintenance and feature development.
+ - Keep two-level design but optimize VMA management (e.g., via RCU-safe maple trees[^3] or concurrent skip lists[^2]). This would improve scalability but not eliminate cross-structure synchronization or correctness risks. These scalable data structures are also complex and require `unsafe` code, which bloats the TCB and increases maintenance burden.
+
+**Unresolved Questions / follow-ups:**
+ - With only 512 entries per page table, we can optimize the auxiliary metadata layout to avoid deep BTreeMap levels.
+ - Implementing efficient anonymous reverse mappings without sacrificing scalability is difficult. Since Asterinas does not support swapping at the moment, we leave efficient anonymous reverse mappings in CortenMM as future work.
+
+## Prior Art and References
+
+[RadixVM: scalable address spaces for multithreaded applications](https://dl.acm.org/doi/10.1145/2465351.2465373) provides another way to unify abstractions. It makes the software-level abstraction a radix tree, similar to hardware page tables, and makes the page tables cache the radix-tree nodes. This design is effective in terms of scalability, but it is not suitable for Asterinas because the software-level abstraction cannot be implemented in safe code, and its Linux compatibility is limited.
+
+[^1]: [CortenMM: Efficient Memory Management with Strong Correctness Guarantees](https://doi.org/10.1145/3731569.3764836)
+
+[^2]: [Scalable Address Spaces using Concurrent Interval Skiplist](https://dl.acm.org/doi/10.1145/3731569.3764807)
+
+[^3]: [Concurrent page-fault handling with per-VMA locks](https://lore.kernel.org/all/20220901173516.702122-1-surenb@google.com/)
+
+[^4]: [LATR: Lazy Translation Coherence](https://dl.acm.org/doi/10.1145/3296957.3173198)
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -120,4 +120,5 @@
 * [RFC Overview](rfcs/README.md)
   * [RFC-0001: RFC Process](rfcs/0001-rfc-process.md)
   * [RFC-0002: Asterinas NixOS](rfcs/0002-asterinas-nixos.md)
+  * [RFC-0003: CortenMM](rfcs/0003-cortenmm.md)
 * [RFC Template](rfcs/rfc-template.md)


book/src/rfcs/0003-cortenmm.md
lrh2000, line 1: I found that we should discuss the lock usage, and it seems to be entirely missing here.

In Linux, there is a big RW semaphore for the MM structure, i.e., [`mm->mmap_lock`](https://elixir.bootlin.com/linux/v6.18.4/source/include/linux/mm_types.h#L1052).

On the one hand, it causes pain in terms of scalability. On the other hand, it provides a consistent view of all the mappings once the lock is held (i.e., no one can create, edit, or delete memory mappings once the MM semaphore is held). More importantly, it can outlive the atomic context (e.g., we can sleep while holding the MM semaphore).

It seems that we will lose the ability forever with our new design. To have a consistent view of all the mappings, we can lock the entire page table. While locking the entire page table may not be as efficient as locking a single semaphore, it is an option. However, since all of the page table locks are spin locks, we cannot sleep.

We could use an approach similar to the page fault handler with per-VMA locks to resolve this issue. I'm not sure if you have already adopted this approach, as the relevant information seems to be completely missing from this RFC:
 1. Lock the page table (spin locks), find the VM maping, and release the page table lock.
 2. Lock the VM mapping (mutexes), read the disk from the hard drive, and release the VM mapping lock.
 3. Lock the page table again, check if the VM mapping is still there, put the page in the page table if yes or retry if no.

I am not aware of any problem that the above approach cannot solve. However, I'm unsure if it aligns with the semantics of various MM system calls. There are quite some callers of [these `mmap_read_lock*` and `mmap_write_lock*` methods](https://elixir.bootlin.com/linux/v6.18.4/source/include/linux/mmap_lock.h#L309-L410), e.g., [`madvise`](https://elixir.bootlin.com/linux/v6.18.4/source/mm/madvise.c#L1699-L1722). I don't know if it's feasible to assess all of them, but it's certainly a drawback.

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,173 @@
...
+Pieces stored:
+    0x3fe000       0x400000               0xa00000      0xa02000
+    [      x      ][          y          ][     z      ]
+      at #1-2 PT         at #2-1 PT         at #1-6 PT
+
+[        I        ]                       [        J        ]
+    at #2-1 PT                                at #2-1 PT
+```
+
+If a page fault happens at `0x600000`, the handler locks the `#2-1` PT node, splits `[y]` into `[a]`, `[b]`, and `[c]` at the `#1-4` PT boundary, maps the page, and inserts `[L]` as a placeholder indicating the presence of a child page table:
```
lrh2000: So, we can split the VMA with the page table locks held? Do we need to acquire additional locks on the VMA? How does that interact with accessing the VMA without the page table lock?
> 1. Lock the page table (spin locks), find the VM mapping, and release the page table lock.
> 2. Lock the VM mapping (mutexes), read the disk from the hard drive, and release the VM mapping lock.

I don't have a clear picture of how the lock is being used here. Have you carefully considered this in your implementation? Maybe I should check your code if you think you've done it correctly...

**EDIT:** I checked the implementation. It appears that VMA locks are unnecessary because the VMA is entirely owned by the page table, which prevents concurrent access. The mutex mentioned above is the VMO mutex, so the current implementation should be quite fine. Am I right?

I had some wrong impression from the Linux implementation, where VMA operations (e.g., split, dup, e.t.c.) may even break the atomic mode ([`__split_vma`](https://elixir.bootlin.com/linux/v6.18.4/source/mm/vma.c#L499) -> [`vm_ops->open`](https://elixir.bootlin.com/linux/v6.18.4/source/mm/vma.c#L543) -> [`drm_gem_shmem_vm_open`](https://elixir.bootlin.com/linux/v6.18.4/source/drivers/gpu/drm/drm_gem_shmem_helper.c#L604) -> [`dma_resv_lock`](https://elixir.bootlin.com/linux/v6.18.4/source/drivers/gpu/drm/drm_gem_shmem_helper.c#L575), which is a mutex). However, this seems to be a rare case. The common case is to [increase a reference count](https://elixir.bootlin.com/linux/v6.18.4/source/drivers/gpu/drm/drm_gem.c#L1096-L1109), similar to our approach that clones a `Arc<Vmo>`/`Arc<SomeHandleToDeviceMappings>`.

lrh2000 (commented):
Thanks for writing the RFC!

Today I realized that the discussion about lock usage seems to be missing here. I think this is now the main difference from the Linux design, since we decided to keep the VMA. So, is it worth discussing here?

book/src/rfcs/0003-cortenmm.md
junyang-zh: > We could use an approach similar to the page fault handler with per-VMA locks to resolve this issue. ...

Indeed the page fault handler in the current implementation #2615 does the similar thing. It locks the page table and find the VM mapping. If need to read from the disk, it releases the page table lock, loads the page and retry locking the page table. If the VM mapping still exists, map. Otherwise `EFAULT`.

> However, I'm unsure if it aligns with the semantics of various MM system calls. There are quite some callers of [these mmap_read_lock* and mmap_write_lock* methods](https://elixir.bootlin.com/linux/v6.18.4/source/include/linux/mmap_lock.h#L309-L410), e.g., [madvise](https://elixir.bootlin.com/linux/v6.18.4/source/mm/madvise.c#L1699-L1722). I don't know if it's feasible to assess all of them, but it's certainly a drawback.

I don't know any system calls that offers stability of all the mappings as a specification (i.e. all mappings in the address space will stay stable when the operation is being executed). I bet there's no applications depending on such behaviors either. I guess by holding the `mmap_lock` in system calls (e.g. `madvice`) , the developers can be freed from reasoning about complex race conditions. But this is a less elegant solution. On the other hand, by holding page table locks on a range, CortenMM guarantees stability of mappings in the range.

Therefore I don't see the change of semantics as a drawback, it actually make things clearer.

book/src/rfcs/0003-cortenmm.md
junyang-zh: > So, we can split the VMA with the page table locks held? ...

As you said, truly there's no VMA locks, so the question does not hold.

> where VMA operations (e.g., split, dup, e.t.c.) may even break the atomic mode

This is what I am not aware of. Thanks for suggesting it. I don't know if we should solve this problem right now by introducing some retry mechanisms. Perhaps no.

book/src/rfcs/0003-cortenmm.md
lrh2000: > Indeed the page fault handler in the current implementation https://github.com/asterinas/asterinas/pull/2615 does the similar thing.

Thanks for the explanation!

> Therefore I don't see the change of semantics as a drawback, it actually make things clearer.

Sorry, maybe I wasn't clear in my comment. The drawback is that we cannot break the atomic mode while holding the page table locks. So VM mappings' stability guarantee cannot break atomic mode, either.
 - For example, if we need to sleep in a system call like `madvise`, we must first release the page table lock. However, releasing the lock can lead to various types of race conditions and requires more analysis to ensure correctness.
 - In addition, releasing the lock can also affect the semantics of the system call, meaning that it may no longer be an atomic operation (with respect to the stability of the VM mappings).
 - In Linux, though, we can sleep while holding the `mmap_lock` and we know all the mappings are still stable.

Do you think it's a drawback?

book/src/rfcs/0003-cortenmm.md
lrh2000: > This is what I am not aware of. Thanks for suggesting it.

For context, I recently worked on NVIDIA drivers, which brought this problem to my attention.
 - https://github.com/NVIDIA/open-gpu-kernel-modules/blob/2ccbad25e1af6a6ee6f38cf569f89f8b65d658ab/kernel-open/nvidia-uvm/uvm.c#L571
 - https://github.com/NVIDIA/open-gpu-kernel-modules/blob/2ccbad25e1af6a6ee6f38cf569f89f8b65d658ab/kernel-open/nvidia-uvm/uvm.c#L468

https://github.com/asterinas/asterinas/pull/2792#discussion_r2679779577 provides an in-tree version in which the `vm_ops_operation::open` function can break the atomic mode.

> I don't know if we should solve this problem right now by introducing some retry mechanisms. Perhaps no.

Of course, I'm not suggesting that we introduce such mechanisms now since there are no users.

I want to identify typical use cases of Linux's sleep behavior under the `mmap_lock` and determine whether it will cause real/fundamental problems in our design.

junyang-zh:
As discussed in https://github.com/asterinas/asterinas/pull/2792#discussion_r2681291018 and https://github.com/asterinas/asterinas/pull/2792#discussion_r2681249674 , the problem now is that our page table locks are not sleepable. We aren't able to make the page table locks sleepable at the moment because we use RCU in the locking phase. This is a drawback that will lead to problems in the short future.

I can think of naive solutions, such as add a wait-list in every `VmMapping`. When modifying `VmMappings`, the caller need to check if anyone is on the list. If so, it needs to add itself to the list, drop the page table locks and waits for others to complete. Then retry getting the page table lock again.

lrh2000:
> As discussed in [#2792 (comment)](https://github.com/asterinas/asterinas/pull/2792#discussion_r2681291018) and [#2792 (comment)](https://github.com/asterinas/asterinas/pull/2792#discussion_r2681249674) , the problem now is that our page table locks are not sleepable. We aren't able to make the page table locks sleepable at the moment because we use RCU in the locking phase. This is a drawback that will lead to problems in the short future.

Yes, I have another use case that is more relevant to the present (I may start working on it this or next week). The goal is to allow device-specific logic to populate device mappings.

The logic is clear in the Linux implementation.
 - [First, take the mmap write lock](https://elixir.bootlin.com/linux/v6.18.4/source/mm/util.c#L579).
 - [Then, find an unmapped area and create the VM mapping](https://elixir.bootlin.com/linux/v6.18.4/source/mm/mmap.c#L406).
 - [Finally, let the device-specific logic populate the VM mapping](https://elixir.bootlin.com/linux/v6.18.4/source/include/linux/fs.h#L2405).

Device-specific logic can sleep for various reasons. It is unreasonable to expect the device to provide pages in the atomic context.

With CortenMM, it's still doable, but I guess we have to make the logic more complex:
 - We can save the pages to map in a `Vec` before acquiring the page table lock. Then, after taking the page table lock, we can atomically map the pages.
   - This will require extra heap allocations and addition checks to avoid unbounded allocations caused the `Vec`.
 - We can release the page table lock before calling the device-specific logic that populates the pages.
   - But what should happen, if the user manages to access/unmap/remap the incomplete VM mappings concurrently?

> I can think of naive solutions, such as add a wait-list in every `VmMapping`. When modifying `VmMappings`, the caller need to check if anyone is on the list. If so, it needs to add itself to the list, drop the page table locks and waits for others to complete. Then retry getting the page table lock again.

Even if no one else is operating the VMA, it needs to drop the page table lock. This is because it needs to call the device-specific logic, which may lock device-specific mutexes.

I think we can find a way to make it work. But the drawbacks are:
 - This seems like an error-prone logic to me, as the VMA may die (e.g., `munmap`) or be modified (e.g., `mprotect`/`remap`/...) in the middle if we don't carefully write some synchronization primitives.
 - It may change the semantics of the system call.
   - For example, if we release the page table lock while copying multiple VM mappings (e.g., in the middle of `fork()`), and we don't restart the copy from scratch the next time, we will end up with inconsistent VM mappings, which can never occur in the Linux implementation.
   - Perhaps no program relies on this. But it's still something to note.

Though you don't call them wait lists instead of VM mapping locks, I believe we will essentially introduce VM mapping locks (at least for some device mappings) if we do so. It isn't obvious how it interacts with the page table lock. That's where the complexity comes from.

By the way, if you are curious why the device needs to care about the VM mapping operations, you can try to have a look at NVIDIA Unified Virtual Memory (UVM). For example, [Unified Memory for CUDA Beginners](https://developer.nvidia.com/blog/unified-memory-cuda-beginners/):
>> When code running on a CPU or GPU accesses data allocated this way (often called CUDA managed data), the CUDA system software and/or the hardware takes care of migrating memory pages to the memory of the accessing processor.

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
+# RFC-0003: CortenMM
+
+* Status: Draft
+* Pull request: https://github.com/asterinas/asterinas/pull/2792
+* Date submitted: 2025-12-24
+* Date approved: YYYY-MM-DD
+
+## Summary
+
+This RFC introduces CortenMM[^1]’s single-level, page-table–centric virtual memory management in Asterinas. Instead of maintaining a separate `VmMapping` tree plus page tables, we store mapping metadata alongside page tables and split mappings on page-table boundaries. This yields better scalability, simpler invariants, and stronger correctness properties while preserving Linux-compatible semantics.
```
tatetian: Change 

> Instead of maintaining a separate `VmMapping` tree

to

> Instead of maintaining a separate [`VmMapping`](https://github.com/asterinas/asterinas/blob/459a634e4acc865b4bad99116a416910835083e5/kernel/src/vm/vmar/vm_mapping.rs#L51) tree (which is equivalent to Linux's [Virtual Memory Area](https://github.com/torvalds/linux/blob/fcb70a56f4d81450114034b2c61f48ce7444a0e2/include/linux/mm_types.h#L904))

to provide more context.

tatetian: Change 

> we store mapping metadata alongside page tables and split mappings on page-table boundaries

to 

> we split mappings on page-table boundaries so that mapping metadata can be attached to page tables.

The latter one is more accurate.

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
...
+
+The page table is the source of truth; per-page-table auxiliary metadata stores `VmMapping`s that cover that page table frame's range. No global VMA tree is required.
+
+`VmMapping`s are split at page-table boundaries when inserted and each PT node only carries mappings aligned to its range. This keeps lookups and updates local and mergeable.
+
+For example, this is the page table after `mmap(offset=0x3fe000, size=0x604000)`. A new $2 \text{MiB} \times 3 + 4 \text{KiB} \times 4$ `VmMapping` is split into 3 pieces (`[x]`, `[y]` and `[z]`) and stored across 2 levels. There are also two placeholders (`[I]` and `[J]`) indicating the presence of child page tables:
+
+```
+                   .----------------.
+L4                 |      #4-1      |
```
tatetian, line 34: What is the meaning of this notation, `#X-Y`? This is not explained and the `Y` part is not even used, I suppose. So this notation brings unnecessarily confusion. I think you can simply name `#4-1` to `4A`, `#1-2` to `1B`, etc.

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
...
+Pieces stored:
+    0x3fe000       0x400000               0xa00000      0xa02000
+    [      x      ][          y          ][     z      ]
+      at #1-2 PT         at #2-1 PT         at #1-6 PT
+
+[        I        ]                       [        J        ]
+    at #2-1 PT                                at #2-1 PT
+```
+
+If a page fault happens at `0x600000`, the handler locks the `#2-1` PT node, splits `[y]` into `[a]`, `[b]`, and `[c]` at the `#1-4` PT boundary, maps the page, and inserts `[L]` as a placeholder indicating the presence of a child page table:
```
tatetian: The notations in lower letters such as `[x]` and `[y]` actually refer to `VmMapping`? If so, you can point this out explicitly.

> inserts `[L]` as a placeholder

What is `[L]` anyway? Is it a page entry entry since it points to the child page table? What is the name of this concept in code?

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
...
+    /// It is called when a new page table is allocated under the page table
+    /// entry at virtual address `va` and at level `level`. A new page table
+    /// will be allocated when:
+    ///  - preparing a new page table for mapping in lower levels;
+    ///  - splitting a huge mapping into smaller mappings.
+    ///
+    /// The receiver [`AuxPageTableMeta`] is the metadata of the parent page
+    /// table and the returned [`AuxPageTableMeta`] will be the metadata of
+    /// the allocated child page table.
+    fn alloc_child_page_table(&mut self, va: Vaddr, level: PagingLevel) -> Self;
```
tatetian: This API design is a bit of weird: metadata (`&mut self`) allocates metadata (`-> Self`)? Does the API implicitly require the parent metadata to allocate its child metadata? If yes, then what if given arguments violates this requirement? The API is not fallible. If not, why does it matter to take a `self`?

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+The user-defined metadata type will be provided to OSTD via `VmSpace`'s new generic parameter `VmSpace<A: AuxPageTableMeta>`. The `()` type also implements the `AuxPageTableMeta` trait as a no-op implementation for backward compatibility. The `VmSpace` will instantiate page table with the `Aux` type in the `PageTableConfig` trait specified as the user-defined type.
+
+The existing OSTD cursor APIs need to support the auxiliary metadata. Therefore, the cursor user must be aware of the level of the cursor, rather than only the virtual address.
```
tatetian: The RFC does not show the API. If you do not list the new API, at least you should give an example. It would be very helpful to show a snippet of simplified sample code that _uses_ the new metadata-enhanced API to implement mmap or handle page fault.

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
...
+* Date submitted: 2025-12-24
+* Date approved: YYYY-MM-DD
+
+## Summary
+
+This RFC introduces CortenMM[^1]’s single-level, page-table–centric virtual memory management in Asterinas. Instead of maintaining a separate `VmMapping` tree plus page tables, we store mapping metadata alongside page tables and split mappings on page-table boundaries. This yields better scalability, simpler invariants, and stronger correctness properties while preserving Linux-compatible semantics.
+
+## Motivation
+
+Two-level VM designs (e.g., VMA trees and page tables) hurt scalability and are a recurring source of subtle concurrency bugs; Linux continues to carry `mmap_lock`/VMA contention and correctness CVEs. Our paper, CortenMM, addresses these issues and has gained substantial positive feedback from the OS research community.
```
tatetian: > carry `mmap_lock`/VMA contention and correctness CVEs.

The verb "carry" is inappropriate here.

book/src/rfcs/0003-cortenmm.md
```diff
@@ -0,0 +1,174 @@
...
+
+The CortenMM paper shows that mainstream ISAs all use radix page tables, so we can collapse to a single level without losing portability. By colocating mapping metadata with page tables, we eliminate cross-structure synchronization and reduce cache/lock traffic. The paper reports 1.2×–26× speedups on real workloads; we expect similar gains for Asterinas user space and kernel services.
+
+A single-level abstraction simplifies reasoning and enables stronger correctness: invariants are local to page-table subtrees; our merged locking protocol already enforces serialized updates per covered range. This reduces risk as we continue to develop for production and broaden hardware coverage.
+
+## Design
+
+This RFC proposes a full replacement of the existing two-level VM system with CortenMM’s single-level design. The design largely follows the CortenMM paper but uses a slightly different metadata storage strategy for compatibility and memory-efficiency reasons.
+
+### High-level shape
```
tatetian: The word "structure" seems to be better than "shape"?

tatetian (commented):
Overall, I am quite positive with the RFC. And I don't see a structural change is required. But there are some minor improvements to be made.

tatetian:
> Though you don't call them wait lists instead of VM mapping locks, I believe we will essentially introduce VM mapping locks (at least for some device mappings) if we do so. It isn't obvious how it interacts with the page table lock. That's where the complexity comes from.

@junyang-zh The handling of spinlocks and mutexes could be a real challenge. It appears to demand a non-trivial extension to the design of CortenMM. Could you try figure out a satisfying solution as part of this RFC?

A related but separate case is acquiring mutexes for I/O. The current proposed implementation (#2615) would [temporarily drop](https://github.com/junyang-zh/asterinas/blob/0ebe29cd45213ccaf5d2aa09e2182862291bf58a/kernel/src/vm/vmar/vmar_impls/page_fault.rs#L56) the VmSpace cursor to commit a missing, faulted page backed by a file. Dropping the cursor introduces the possibility of races, although I analyzed the code and can't see any real harm is done. But I think this case weakens the promises of CortenMM as simplified concurrency reasoning is one of the key benefits.

If CortenMM has built-in support for locking with mutexes, then maybe the "acquiring mutexes for I/O" case can also be handled without dropping the cursor.

book/src/rfcs/0003-cortenmm.md
junyang-zh: `#X-Y` is the `Y`th page table of level `X`. This is not a very useful information but I think its better than `A`/`B`/`C`/... I will add a short explanation.

book/src/rfcs/0003-cortenmm.md
junyang-zh: https://github.com/junyang-zh/asterinas/blob/0ebe29cd45213ccaf5d2aa09e2182862291bf58a/kernel/src/vm/vmar/vmar_impls/mod.rs#L110-L114

Each item in the metadata has this enum type. the placeholder is that `ChildPt` variant. We can remove all these if it is confusing.

book/src/rfcs/0003-cortenmm.md
junyang-zh: Metadata of one virtual page should only exist in one level. So when the page table is allocated, metadata that used to exist in the parent level must be moved to the lower level. That's why the change is needed in the parent page table.

> If yes, then what if given arguments violates this requirement?

Given arguments cannot violate this requirement (require the parent metadata to allocate its child metadata). OSTD ensures this. But we cannot write this as an unsafe function.

junyang-zh:
> If CortenMM has built-in support for locking with mutexes, then maybe the "acquiring mutexes for I/O" case can also be handled without dropping the cursor.

On second thought I think it is possible. Because after getting the lock and checking the `stale` bit in the page table frame metadata, the page table frame will be alive regardless of RCU grace periods. So the reason why current page table locks cannot sleep is not fundamental. We could change it to mutexes with moderate coding efforts. I reflected this in the RFC's unresolved questions part.

tatetian:
> the reason why current page table locks cannot sleep is not fundamental. We could change it to mutexes with moderate coding efforts.

I guess so. I can't think of a reason that page tables must be manipulated in the atomic mode.

book/src/rfcs/0003-cortenmm.md
tatetian: Sure. I can speculate your intention--and you have confirmed it--which makes sense. It's just that the documentation needs more work to reflect your rationale. Could you mention in the method's Rust doc the fact that some metadata in the parent page table would be migrated into the child? This information would be very helpful to justify the interface.

In addition, you should point out the fact that this caller of this method is OSTD, which is responsible for giving some sensible values for the input arguments. That's why the interface is infallible. Clearly, the arguments must satisfy some (unwritten) constraints (e.g., `va: Vaddr` should within the scope of the parent node and page-aligned; also, if a page table child should not be created twice) and you should try do a better job on writing out these constraints.

In addition, I am wondering if a `dealloc` API is needed. When creating the metadata for a child, some metadata in the parent might be migrated to the child. So when the child is destroyed, some metadata needs to be returned back to the parent?

lrh2000:
> I guess so. I can't think of a reason that page tables must be manipulated in the atomic mode.

For performance? We need to use quite a few locks with CortenMM. Mutexes can be slower and may not scale?

tatetian:
> > I guess so. I can't think of a reason that page tables must be manipulated in the atomic mode.
> 
> For performance? We need to use quite a few locks with CortenMM. Mutexes can be slower and may not scale?

If different CPU cores access different mutexes on different page table pages, it should scale, I guess. Of course, it won't perform as well as using RCU.

By the way, Linux has a thing called sleepable RCUs. I am not sure if it is what we want.

lrh2000:
> If different CPU cores access different mutexes on different page table pages, it should scale, I guess. Of course, it won't perform as well as using RCU.

I mean the efficiency and scalability of the lock. Mutexes are slower as spin locks and are not as scalable as qspinlocks. If I recall correctly, qspinlocks seem to play an important role in the benchmark results, though I'm not sure about the details.

tatetian:
> > If different CPU cores access different mutexes on different page table pages, it should scale, I guess. Of course, it won't perform as well as using RCU.
> 
> I mean the efficiency and scalability of the lock. Mutexes are slower as spin locks and are not as scalable as qspinlocks. If I recall correctly, qspinlocks seem to play an important role in the benchmark results, though I'm not sure about the details.

CortenMM shines when different CPU cores manipulate non-overlapping memory regions, which means access to different page table pages and their locks. Mutexes is slower than spinlocks only when there is contention. But CortenMM maximizes the performance for no contention cases, where spinlocks and mutexes do not make much a difference.

lrh2000:
> Mutexes is slower than spinlocks only when there is contention.

Mutexes are slower even if there is no contention. This is because when unlocking a mutex, we need to try to wake up the queue:
https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/ostd/src/sync/mutex.rs#L57-L59

Then we will hit this logic:
https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/ostd/src/sync/wait.rs#L89-L93
https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/ostd/src/sync/wait.rs#L136-L140

Therefore, a synchronization operation (i.e., `mfence`) is still emitted. This overhead may not be negligible.

> But CortenMM maximizes the performance for no contention cases, where spinlocks and mutexes do not make much a difference.

This needs explanation from @junyang-zh. But [this commit switches page table locks to qspinlocks](https://github.com/asterinas/asterinas/commit/28c64b6d050a0df47d9356e6d5e9a7ccdac1034c) (taken from [TELOS-syslab/asterinas](https://github.com/TELOS-syslab/asterinas/tree/corten-2) repo). So for some reason I am not aware of, I believe it must make a difference for some benchmark results.

This is also [mentioned in this RFC](https://github.com/junyang-zh/asterinas/blob/42242f69f3b2d3b3486ba0aa21e1749013f72cd2/book/src/rfcs/0003-cortenmm.md?plain=1#L96).

book/src/rfcs/0003-cortenmm.md
junyang-zh: > So, when the child is destroyed, some metadata needs to be returned to the parent?

Currently, the OSTD will not destroy the child. But the kernel controls when to destroy the child via the `unmap` API. So the kernel handles the lifting.

As for the `alloc` API, ideally, there seems to be no reason why we cannot let the kernel fully decide when to propagate and allocate child page tables. It exists at this moment because it will mandate another OSTD API that creates new empty page tables and makes the kernel logic slightly more complicated. But I believe it's the right direction. I may change the RFC to exclude the `alloc` API if you agree with me about the previous point.

junyang-zh:
> Mutexes are slower as spin locks and are not as scalable as qspinlocks.

This is true.

> If I recall correctly, qspinlocks seem to play an important role in the benchmark results, though I'm not sure about the details.

> But CortenMM maximizes the performance for no contention cases, where spinlocks and mutexes do not make much a difference.

This is true, `qspinlocks` play an important role in high-contention benchmarks, where operations are close to each other (in terms of virtual addresses). In low contention benchmarks, the operations need to only lock different page tables, and the operations on the higher level page tables are synchronized by RCU rather than locks. So In low contention benchmarks, Mutexes will not lead to much degradation of scalability.

> Mutexes are slower even if there is no contention.

This is also true; address-space-wide operations, such as fork, which takes all the locks, will experience a small performance degradation due to the many mutexes.

---

Still, using Mutexes as the page table locks is a better solution than per-VMA locks to resolve the sleeping problem, as per-VMA locks will introduce **more** mutexes to acquire for all the operations we discussed above.

tatetian:
@junyang-zh So I think we can we agree upon two points:
1. Sooner or later, we will have to introduce mutexes as the locks for page table to accommodate the use cases of mapping device memory (the major reason) and file-backed memory (a secondary reason).
2. Introducing mutexes will affect the performance, presumably to a small extent.

But I am not comfortable with the second point without experimental results. Do you think it is feasible to have some experimental results to back up the projection?

lrh2000:
> A related but separate case is acquiring mutexes for I/O. The current proposed implementation (#2615) would [temporarily drop](https://github.com/junyang-zh/asterinas/blob/0ebe29cd45213ccaf5d2aa09e2182862291bf58a/kernel/src/vm/vmar/vmar_impls/page_fault.rs#L56) the VmSpace cursor to commit a missing, faulted page backed by a file. Dropping the cursor introduces the possibility of races, although I analyzed the code and can't see any real harm is done. But I think this case weakens the promises of CortenMM as simplified concurrency reasoning is one of the key benefits.

@tatetian In fact, it's not as simple as you thought. With reverse mappings, CortenMM will worsen the lock order problem, and we'll still have to retry to ensure a consistent lock order.

In the Linux's design or Asterinas' current design, the lock order is:
 - mmap lock -> rmap lock -> page table lock

Locking either the mmap lock when performing an `mmap` system call or the rmap lock when writing back file-backed pages can ensure that the VMA cannot be modified, e.g., remapped or unmapped. Then, we can insert a new page into the page table or remove an existing page from it.

However, CortenMM proposes to merge the mmap lock into the page table lock. None of the following lock orders will work out of the box:
 - rmap lock -> page table lock: This won't work for many MM system calls (e.g., `mremap`/`munmap`). First, we have to lock the page table lock to get the VMA, and then we have to lock the rmap lock in the VMO to update the reverse mapping information.
 - page table lock -> rmap lock: This won't work when removing pages from the reserve mapping because we have to lock the RMAP lock first in order to find the page tables that map the pages.

So, we'll have to introduce some retry mechanisms somewhere. I think it's feasible to choose the second lock order (page table lock -> rmap lock). Then, we can retry if the page tables cannot be locked when removing pages according to the reverse mappings.
 - I don't want to retry when handling MM system calls because it will disrupt the system call's atomicity. If we release the page table lock in the middle, users can create, change, or delete mappings.

Considering device mappings & page cache, this situation will only get worse. I believe the lock order used in the Linux is:
 - mmap lock -> driver lock / page lock -> rmap lock -> page table lock

For example, the driver (e.g., NVIDIA drivers) can take its lock and trigger an unmap from the reverse mappings. Meanwhile, the driver lock will be acquired after the mmap lock when processing the mmap system call. This ensures that the memory region is filled before the VMA can be modified.

The "page lock" ensures that the page being mapped to user space is neither an uninitialized page nor a page being written back to disk (see https://github.com/asterinas/asterinas/issues/2937).

Again, CortenMM proposes to merge the mmap lock to the page table lock. Then, the problem is where to put the driver lock and the page lock? In order to make it work, we will need to retry on both the page fault (for the page lock) and the mmap system call paths (for the driver lock).

However, retrying the mmap system call is problematic because the user may try to modify the mapping concurrently from user space. I cannot think of a solution other than caching all the pages the device wants to map in a standalone data structure (e.g., `Vec<UFrameOrIoMem>`).

(Actually, implementing reverse mappings in the current main branch is quite easy. I can create a draft very soon. It may help understand the problem.)

junyang-zh:
The RFC has been updated based on the above discussions. Welcome to further questions!

PR #2793, opened 2025-12-24
Fix the issue of notifying parent in fsnotify

Plucky923:


kernel/src/fs/path/mod.rs
```diff
@@ -177,6 +177,18 @@ impl Path {
         mount_parent.effective_parent()
     }
 
+    /// Gets the parent `Path` within the same mount.
+    ///
+    /// This method returns the parent directory within the same filesystem/mount.
```
cchanging: "returns the parent directory" -> "returns the parent path"

Plucky923: Done

PR #2794, opened 2025-12-25
Make `effective_parent` be aware of the current thread's root path

cchanging:
After chroot, when looking up parent directories along a path, the traversal should not go beyond the root path.

### DIFF
--- a/test/initramfs/src/apps/chroot/chroot_jail.c
+++ b/test/initramfs/src/apps/chroot/chroot_jail.c
@@ -1,0 +1,160 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+#include <dirent.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <sched.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mount.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "../test.h"
+
+FN_SETUP(create_chroot_env)
+{
+	// Create chroot target structure
+	CHECK_WITH(mkdir("/foo", 0755), _ret >= 0 || errno == EEXIST);
+	CHECK_WITH(mkdir("/foo/proc", 0755), _ret >= 0 || errno == EEXIST);
+	CHECK_WITH(mkdir("/foo/nix", 0755), _ret >= 0 || errno == EEXIST);
+
+	// Perform bind mounts
+	CHECK(mount("proc", "/foo/proc", "proc", 0, ""));
+	CHECK(mount("/nix", "/foo/nix", NULL, MS_BIND, NULL));
+}
+END_SETUP()
+
+// Helper function to check if a directory does NOT contain a specific entry
+static int dir_not_contains(const char *path, const char *entry_name)
+{
+	DIR *dir = CHECK(opendir(path));
+	struct dirent *entry;
+	while ((entry = readdir(dir)) != NULL) {
+		if (strcmp(entry->d_name, entry_name) == 0) {
+			CHECK(closedir(dir));
+			return -1; // Found the entry, which means failure
+		}
+	}
+	CHECK(closedir(dir));
+	return 0; // Entry not found, success
+}
+
+// Helper function to read a file and check for a substring
+static int file_contains(const char *filepath, const char *substring)
+{
+	int fd = CHECK(open(filepath, O_RDONLY));
+	char buf[4096] = { 0 };
+	CHECK(read(fd, buf, sizeof(buf) - 1));
+	CHECK(close(fd));
+	return strstr(buf, substring) != NULL ? 0 : -1;
+}
+
+// Macro to wait for a child process and check its exit status
+#define WAIT_AND_CHECK_CHILD(pid)                            \
+	do {                                                 \
+		int status;                                  \
+		TEST_RES(waitpid(pid, &status, 0),           \
+			 _ret == pid && WIFEXITED(status) && \
+				 WEXITSTATUS(status) == 0);  \
+	} while (0)
+
+FN_TEST(chroot_getcwd)
+{
+	pid_t pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		CHECK(chroot("/foo"));
+		CHECK(chdir("/"));
+
+		char cwd[PATH_MAX];
+		CHECK_WITH(getcwd(cwd, sizeof(cwd)), strcmp(cwd, "/") == 0);
+
+		exit(EXIT_SUCCESS);
+	} else {
+		WAIT_AND_CHECK_CHILD(pid);
+	}
+}
+END_TEST()
+
+FN_TEST(chroot_chdir)
+{
+	pid_t pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		CHECK(chroot("/foo"));
+		CHECK(chdir("/"));
+		CHECK(chdir(".."));
+
+		// Verify we can't see 'foo' directory
+		CHECK(dir_not_contains(".", "foo"));
+
+		// Verify we are still at root
+		char cwd[PATH_MAX];
+		CHECK_WITH(getcwd(cwd, sizeof(cwd)), strcmp(cwd, "/") == 0);
+
+		exit(EXIT_SUCCESS);
+	} else {
+		WAIT_AND_CHECK_CHILD(pid);
+	}
+}
+END_TEST()
+
+FN_TEST(chroot_mountinfo)
+{
+	pid_t pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		CHECK(chroot("/foo"));
+		CHECK(chdir("/"));
+
+		CHECK(file_contains("/proc/self/mountinfo", "/nix /nix"));
+
+		exit(EXIT_SUCCESS);
+	} else {
+		WAIT_AND_CHECK_CHILD(pid);
+	}
+}
+END_TEST()
+
+FN_TEST(chroot_fd_escape)
+{
+	pid_t pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		int pre_chroot_fd =
+			CHECK(open("/test", O_RDONLY | O_DIRECTORY));
+
+		CHECK(chroot("/foo"));
+		CHECK(chdir("/"));
+
+		// Use fchdir to go to the pre-chroot /test directory
+		CHECK(syscall(SYS_fchdir, pre_chroot_fd));
+
+		// Now getcwd should add "(unreachable)" prefix because we're outside the chroot jail
+		char cwd[PATH_MAX];
+		CHECK_WITH(syscall(SYS_getcwd, cwd, sizeof(cwd)),
+			   strcmp(cwd, "(unreachable)/test") == 0);
+
+		// But we should be able to see 'foo' directory by listing parent
+		CHECK(chdir(".."));
+		CHECK_WITH(dir_not_contains(".", "foo"), _ret == -1);
+
+		CHECK(close(pre_chroot_fd));
+		exit(EXIT_SUCCESS);
+	} else {
+		WAIT_AND_CHECK_CHILD(pid);
+	}
+}
+END_TEST()
+
+FN_SETUP(cleanup_chroot_env)
+{
+	CHECK(umount("/foo/proc"));
+	CHECK(umount("/foo/nix"));
+	CHECK(rmdir("/foo/proc"));
+	CHECK(rmdir("/foo/nix"));
+	CHECK(rmdir("/foo"));
+}
+END_SETUP()
--- a/test/initramfs/src/apps/scripts/process.sh
+++ b/test/initramfs/src/apps/scripts/process.sh
@@ -10,6 +10,7 @@ cd ${SCRIPT_DIR}/..
 echo "Start process test......"
 # These test programs are sorted by name.
 tests="
+chroot/chroot_jail
 clone3/clone_exit_signal
 clone3/clone_files
 clone3/clone_no_exit_signal
--- a/test/initramfs/src/apps/Makefile
+++ b/test/initramfs/src/apps/Makefile
@@ -15,6 +15,7 @@ TEST_BUILD_DIR ?= $(INITRAMFS)/test
 TEST_APPS := \
 	alarm \
 	capability \
+	chroot \
 	clone3 \
 	cpu_affinity \
 	devfs \
--- a/test/initramfs/src/apps/chroot/Makefile
+++ b/test/initramfs/src/apps/chroot/Makefile
@@ -1,0 +1,5 @@
+# SPDX-License-Identifier: MPL-2.0
+
+include ../test_common.mk
+
+EXTRA_C_FLAGS :=


kernel/src/fs/path/mod.rs
```diff
@@ -170,11 +181,32 @@ impl Path {
             return Some(Self::new(self.mount.clone(), self.dentry.parent().unwrap()));
```
tatetian: What if `self` is already the root?

This shows the importance of adding regression tests. If regression tests were written properly, all corner cases would have been considered in the implementation. Please add regression tests to show after `chroot`, one cannot `cd ..` his way out of the new root.

tatetian: What if `self` is the current root that is made out of a non-mount-root inode via `chroot`?

Consider the following scenario:

```
$ mkdir -p /foo
$ cd /foo
$ chroot .
$ cd ..
```

I believe--not confirmed by experiments--that the current implementation would allow `cd ..` because the first two lines do not check if `self` may be the current root and this root may not be a mount root due to the use of `chroot`.

This shows the importance of adding regression tests. If regression tests were written properly, all corner cases would have been considered in the implementation. Please add regression tests to show after `chroot`, one cannot `cd ..` his way out of the new root.

kernel/src/fs/path/mod.rs
```diff
@@ -146,19 +152,24 @@ impl Path {
...
 
-        let Some(parent) = self.mount.parent() else {
-            return self.dentry.name();
-        };
-        let Some(mountpoint) = self.mount.mountpoint() else {
-            return self.dentry.name();
-        };
+        loop {
+            if !current.is_mount_root() {
+                return current.dentry.name();
```
tatetian: Similarly, the implementation fails to consider the corner case where `current` is the current root that is made out of a non-mount-root inode via `chroot`. In this case, the returned effective name should be `/`, rather than the root inode's original name.

A regression like the following should be added:

```
$ mkdir /foo
$ cd /foo
$ chroot .
$ pwd
```

I believe--not confirmed by experiments--that the current implementation would print in `/foo`, but the correct result should be `/`.

kernel/src/fs/path/mod.rs
```diff
@@ -170,11 +181,32 @@ impl Path {
             return Some(Self::new(self.mount.clone(), self.dentry.parent().unwrap()));
         }
 
-        let parent = self.mount.parent()?;
-        let mountpoint = self.mount.mountpoint()?;
+        let current_thread = current_thread!();
+        let current_fsinfo = current_thread.as_posix_thread()?.read_fs();
+        let current_resolver = current_fsinfo.resolver().read();
+        let root = current_resolver.root();
```
tatetian: I suggest rethinking the relationship between the `path` and `fs_resolver` modules. I have two observations:

First, after this change, `Path::abs_path` will repeatedly get the current FS resolver when traversing up to the root. This is highly undesirable in terms of efficiency.

Second, after this change, `Path::effective_parent`, `Path::abs_path`, and `Path::effective_name` now rely on `FsResolver` _implicitly_. But this is non-trivial dependency. From the API level, this dependency would be quite a surprise.

Base on these two observations, I suggest migrating the three methods to `FsResolver`. Doing so allows the user to immediately see that it needs to two sources of information, a `Path` and the current `FsResolver`, to determine the absolute path, the effective name, the effective parent of a path. In addition, since they are now methods of `FsResolver`, the root inode is readily accessible.

```rust
impl FsResolver {
    /// Constructs the absolute path of a `Path`.
    ///
    /// This method internally uses `FsResolver::resolve_name` and `FsResolver::resolve_parent` repeatedly to wind up a path to the root.
    ///
    /// This methods guarantee that the returned string is non-empty
    /// and starts with `"/"`.
    pub fn make_abs_path(&self, path: &Path) -> String { ... }

    /// Resolves the name of a `Path`.
    ///
    /// The method resolves the name by the following rules:
    /// 1. If the path is the root of the FS resolver, then the name is `"/"`.
    /// 2. If the path is not the root of a mount, then the name is the same as that of the underlying dentry;
    /// 3. If the path is the root of a mount and
    ///     a. If the mount has a parent mount, then the name is that of the corresponding mountpoint in the parent mount;
    ///     b. If the mount has no parent, then the name is `"/"`.
    pub fn resolve_name(&self, path: &Path) -> String { ... }

    /// Resolves the parent of a `Path`.
    ///
    /// The method resolves the parent by the following rules:
    /// 1. If the path is the root of the FS resolver, then the parent is none.
    /// 2. If the path is not the root of a mount, then the parent is the same as that of the underlying dentry;
    /// 3. If the path is the root of a mount and
    ///     a. If the mount has a parent mount, then the parent is that of the corresponding mountpoint in the parent mount;
    ///     b. If the mount has no parent, then the parent is none.
    pub fn resolve_parent(&self, path: &Path) -> Option<Path> { ... }
}
```

And if you really think about it, this feels natural. Because the three methods are indeed quite relevant to path resolution/lookup. The existing `lookup_`-family methods are forward lookup, where the new methods are more like reverse lookup.

Now that we consider `fs_resolver` and `path` having closer relationship, I suggest moving `fs_resolver` into the `path` module, renaming `fs_resolver` to `resolver`, `FsResolver` to `PathResolver`.

---

One blocker for migrating `Path::effective_name` to `PathResolver::resolve_name` is the user of fsnotify. But I have explained in #2844 that fsnotify should not have used `Path::effective_name`. After #2844 is resolved, I think this PR can be continued.

tatetian (commented):
Thanks for addressing this major limitation due to the introduction of the `chroot` syscall. I am afraid that the implementation needs quite some rework to address correctness, efficiency, and maintainability concerns.

kernel/src/syscall/getcwd.rs
```diff
@@ -6,11 +6,25 @@ use super::SyscallReturn;
...
+        path_resolver.make_abs_path(path_resolver.cwd())
+    };
+
     debug!("getcwd: {:?}", name);
 
-    let cwd = CString::new(name)?;
+    // Linux will add a "(unreachable)" suffix to the path if the cwd is not reachable.
+    // However, according to POSIX, `getcwd()` should fail with ENOENT in this case.
+    // `glibc` treats the Linux's behavior as a bug and handles the Linux-specific suffix
+    // to conform to POSIX. Here follows the POSIX behavior directly.
```
tatetian: These are some really good comments 👍 

> Here follows the POSIX behavior directly.

But I don't understand why you chose to follow POSIX. Being Linux compatible means  copying Linux's bugs sometime...

kernel/src/fs/path/mod.rs
```diff
@@ -206,7 +206,7 @@ impl Path {
             return_errno_with_message!(Errno::ENOTDIR, "the path is not a directory");
         }
 
-        if self.effective_parent().is_none() {
+        if ctx.thread_local.borrow_fs().resolver().read().root() == self {
             return_errno_with_message!(Errno::EINVAL, "the root cannot be mounted on");
```
tatetian, line 213: Keep the resolver lock during the mount operation. Otherwise, this invariant may not be held due to concurrent `cwd` syscall.

kernel/src/fs/procfs/pid/task/comm.rs
```diff
@@ -33,7 +35,19 @@ impl FileOps for CommFileOps {
...
+            current_task
+                .as_thread_local()
+                .unwrap()
+                .borrow_fs()
+                .resolver()
+                .read()
+                .make_abs_path(executable_file)
+                .into_string()
+        };
         let last_component = exe_path.rsplit('/').next().unwrap_or(&exe_path);
```
tatetian: If all the needed is the last component of the path, why bother concat the absolute path? Isn't it simply to call `Path::name`?

kernel/src/process/execve.rs
```diff
@@ -165,7 +166,12 @@ fn do_execve_no_return(
     unshare_and_close_files(ctx);
 
     // Update the process's executable path and set the thread name
-    let executable_path = elf_file.abs_path();
+    let executable_path = thread_local
+        .borrow_fs()
+        .resolver()
+        .read()
```
tatetian: The current function, `do_execve_no_return`, has only one user, `do_execve`, which already holds the read lock of the path resolver. So a reference to the path resolver should be passed from `do_execve` to `do_execve_no_return`, rather than locking it again.

kernel/src/process/process/init_proc.rs
```diff
@@ -106,19 +106,24 @@ fn create_init_task(
 ) -> Result<Arc<Task>> {
     let credentials = Credentials::new_root();
 
-    let elf_load_info = {
+    let (elf_load_info, thread_name) = {
```
tatetian: This code block is about loading the ELF. So returning ELF-related stuff is more natural  than doing thread name. So I suggest rewriting it this way:

```rust
let (elf_load_info, elf_abs_path) = {
    ...
};

...

let thread_name = ThreadName::new_from_executable_path(&elf_abs_path);

let thread_builder = PosixThreadBuilder::new(tid, thread_name, Box::new(user_ctx), credentials);

```

kernel/src/syscall/rename.rs
```diff
@@ -60,8 +60,8 @@ pub fn sys_renameat2(
     };
 
     // Check the absolute path
-    let old_abs_path = old_path.abs_path();
-    let new_abs_path = new_dir_path.abs_path() + "/" + new_name;
+    let old_abs_path = path_resolver.make_abs_path(&old_path).into_string();
+    let new_abs_path = path_resolver.make_abs_path(&new_dir_path).into_string() + "/" + new_name;
     if new_abs_path.starts_with(&old_abs_path) {
```
tatetian, line 67: It is wrong to check whether the old path is a prefix of the new path by concat'ing the path in string. This is because `new_name` may include special characters such as `.` and `..` or it may end up with `/`. In addition, concat'ing the path string is also inefficient.

Instead, the `lookup_unresolved`-faimily API should be used to obtain the `Path` object for the new parent directory (say, `new_dir_path`) and then check if `new_dir_path.is_equal_or_descendant_of(old_dir_path)` returns true. This `is_equal_or_descendant_of` method needs to be added to `Path`.

This is not a problem introduced by this PR. So I am open with adding a TODO here and fix it later in a follow-up PR.

test/initramfs/src/apps/scripts/process.sh
```diff
@@ -45,6 +45,7 @@ mmap/mmap_vmrss
 namespace/mnt_ns
 namespace/setns
 namespace/unshare
+path_resolver/resolve_after_chroot
```
tatetian: Path resolver is a kernel-space implementation detail. How can the userspace program test it directly? And from the angle of consistency, all other test programs under `apps` are named after userspace-visible concepts. But this test program is not. This should give you the hint that there is something is wrong with the test...

I found that I recently made several comments for your PRs that repeat the same underlying principle: **DO NOT EXPOSE IMPLEMENTATION DETAIL**. I think this principle is straightforward and its benefits are obvious. If you have any doubt about this principle, let's discuss it. If not, I hope you can adhere to this principle from now on.

We should test the interactions between `chroot` and other FS operations. So I find it perfectly fine to name the test directory `chroot/`.

test/initramfs/src/apps/path_resolver/resolve_after_chroot.c
```diff
@@ -0,0 +1,138 @@
...
+		CHECK(chdir("/"));
+
+		CHECK(file_contains("/proc/self/mountinfo", "/nix /nix"));
+
+		exit(EXIT_SUCCESS);
+	} else {
+		int status;
+		TEST_RES(waitpid(pid, &status, 0),
+			 _ret == pid && WIFEXITED(status) &&
+				 WEXITSTATUS(status) == 0);
```
tatetian: This branch is repeated four times. Rewrite the code to avoid duplication.

test/initramfs/src/apps/path_resolver/resolve_after_chroot.c
```diff
@@ -0,0 +1,138 @@
...
+	if (pid == 0) {
+		int pre_chroot_fd =
+			CHECK(open("/test", O_RDONLY | O_DIRECTORY));
+
+		CHECK(chroot("/foo"));
+		CHECK(chdir("/"));
+
+		// After chroot, getcwd should work fine
+		char cwd[PATH_MAX];
+		CHECK_WITH(getcwd(cwd, sizeof(cwd)), strcmp(cwd, "/") == 0);
```
tatetian: This assertion duplicates the one in `chroot_getcwd`. Having duplicated assertions dilute the true purpose of this test.

test/initramfs/src/apps/chroot/chroot_jail.c
```diff
@@ -0,0 +1,138 @@
...
+
+		// Use fchdir to go to the pre-chroot /test directory
+		CHECK(syscall(SYS_fchdir, pre_chroot_fd));
+
+		// Now getcwd should fail with ENOENT because we're outside the chroot jail
+		CHECK_WITH(getcwd(cwd, sizeof(cwd)), errno == ENOENT);
+
+		// But we should be able to see 'foo' directory by listing parent
+		CHECK(chdir(".."));
+		CHECK_WITH(dir_not_contains(".", "foo"), _ret == -1);
```
tatetian, line 142: I am not sure about the purpose of this test. Would this regression test fail on the mainline? I don't think so. This is all about `cwd`, not about `chroot`. Our treatment for the root dir does not seem to matter.

kernel/src/fs/path/resolver.rs
```diff
@@ -84,6 +224,185 @@ impl FsResolver {
...
         Ok(())
     }
+}
+
+/// The result of resolving an absolute path name.
+///
+/// If the path can be traced back to the root of the resolver, it is `Reachable`.
+/// Otherwise, it is `Unreachable`.
+#[derive(Debug, Clone)]
+pub enum AbsolutePathName {
```
tatetian: Calling it `AbsolutePathResult` or `AbsPathResult` (the latter is consistent with the method name) would be better.

kernel/src/fs/path/resolver.rs
```diff
@@ -84,6 +224,185 @@ impl FsResolver {
...
+    }
+
+    /// Returns whether the path is reachable from the root of the resolver.
+    pub fn is_reachable(&self) -> bool {
+        matches!(self, AbsolutePathName::Reachable(_))
+    }
+}
+
+// Mount info reading implementation
+impl PathResolver {
```
tatetian, line 255: Breaking the methods into multiple implementation blocks with distinct purposes is a great idea 👍

kernel/src/syscall/getcwd.rs
cchanging: > Being Linux compatible means copying Linux's bugs sometime...

Sorry, I did not realize this before...

test/initramfs/src/apps/chroot/chroot_jail.c
cchanging: > I am not sure about the purpose of this test.

Do you mean the purpose of `chroot_fd_escape` test case or 124-125 lines? 

The `chroot_fd_escape` test case is used to test the behaviors when the resolver is outside the chroot jail. It will fail on the mainline in `CHECK_WITH(getcwd(cwd, sizeof(cwd)), errno == ENOENT)`. 

The checks on lines 124-125 verify that `cd`/`lookup` operations function correctly in this situation. They are designed to detect overly restrictive logic that could inadvertently fail actions which would otherwise succeed. These checks will pass on the mainline because the mainline contains no such restrictive logic.

kernel/src/syscall/rename.rs
cchanging: I've added a FIXME comment; I'll correct this and add appropriate tests in a later PR.

test/initramfs/src/apps/chroot/chroot_jail.c
tatetian: Ok. Testing "(unreachable)" prefix makes sense to me.

kernel/src/syscall/getcwd.rs
```diff
@@ -3,14 +3,29 @@
...
+
+    // Linux will add a "(unreachable)" prefix to the path if the cwd is not reachable.
+    // However, according to POSIX, `getcwd()` should fail with ENOENT in this case.
+    // `glibc` treats the Linux's behavior as a bug and handles the Linux-specific prefix
+    // to conform to POSIX. Here follows Linux's behavior to keep compatibility.
+    //
+    // Reference: <https://man7.org/linux/man-pages/man3/getcwd.3.html>
+    let abs_path = match abs_path {
+        AbsPathResult::Reachable(s) => s,
+        AbsPathResult::Unreachable(s) => alloc::format!("(unreachable){}", s),
```
tatetian, line 25: I am wondering if the `"(unreachable)"` prefix should be added here or in `AbsPathResult::to_string`. In other words, will this prefix be shown to the userspace by other means? For example, what about `/proc/<pid>/fd/<fd>`? Will it show the prefix if the corresponding file is unreachable.

cchanging: >  if the "(unreachable)" prefix should be added here or in AbsPathResult::to_string

In Linux, this logic is implemented in [getcwd](https://github.com/torvalds/linux/blob/b54345928fa1dbde534e32ecaa138678fd5d2135/fs/d_path.c#L433) syscall, so I think this is a specific operation in `getcwd`. 

>  For example, what about /proc/<pid>/fd/<fd>? Will it show the prefix if the corresponding file is unreachable.

I have attempted these cases and found that there is no "unreachable" prefix in such scenarios. For example, if the `/test` directory is opened and then chroot to `/foo`, the link under fd remains as `/test`. 

I also tried execve with an fd that is out of jail, and its `comm` attribute similarly does not have the "unreachable" prefix. In addition, in this test I discovered a bug in our current system: when `execveat` is passed an fd (assuming its value is 3), the executed path seems to be `/proc/self/fd/3` and the `comm` attribute displays the value as 3 instead of the actual dentry name of the file. This bug will also be addressed in a subsequent PR.

tatetian:
Ok. I have no further problems. I rerun all failed CI tests. Let's see if they can pass.

cchanging:
> I rerun all failed CI tests. Let's see if they can pass.

Currently, the CI failure seems to be due to the new image occupying too much space, which results in a relatively high probability of failure. @cqs21 is currently fixing this issue.

kernel/src/fs/path/resolver.rs
```diff
@@ -51,6 +59,143 @@ impl PathResolver {
...
+    /// - [`AbsPathResult::Reachable`]: The path can be traced back to the resolver's
+    ///   root. The returned string is guaranteed to be non-empty and starts with `/`.
+    /// - [`AbsPathResult::Unreachable`]: The path cannot be traced back to the resolver's
+    ///   root. This can happen for:
+    ///   - Pseudo paths that have no parent in the filesystem tree.
+    ///   - Paths where the parent chain terminates before reaching the resolver's root.
+    ///
+    ///   For pseudo paths, the returned string is simply the path's name. For other
+    ///   unreachable paths, the returned string starts with `/` but represents a partial
+    ///   path that could not reach the root.
```
lrh2000:
```suggestion
    ///  For pseudo paths, the returned string is simply the path's name. For other
    ///  unreachable paths, the returned string starts with `/` but represents a partial
    ///  path that could not reach the root.
```

kernel/src/fs/path/resolver.rs
```diff
@@ -421,6 +665,80 @@ impl LookupParentResult {
...
+    }
+}
+
+impl<'a> TryFrom<&'a str> for FsPath<'a> {
+    type Error = crate::error::Error;
+
+    fn try_from(path: &'a str) -> Result<FsPath<'a>> {
+        FsPath::from_fd_and_path(AT_FDCWD, path)
+    }
+}
```
lrh2000, line 741: Please insert a blank line below this line.

Besides, I'm aware that this is not introduced in this PR, but it shows up in the diff. Would you mind revising the comment style as well? Thanks.

Something like:
```patch
diff --git a/kernel/src/fs/path/resolver.rs b/kernel/src/fs/path/resolver.rs
index b7c3663f4..93b8fbc41 100644
--- a/kernel/src/fs/path/resolver.rs
+++ b/kernel/src/fs/path/resolver.rs
@@ -665,7 +665,7 @@ impl LookupParentResult {
     }
 }
 
-/// Path in the file system.
+/// A path in the file system.
 #[derive(Debug)]
 pub struct FsPath<'a> {
     inner: FsPathInner<'a>,
@@ -673,15 +673,15 @@ pub struct FsPath<'a> {
 
 #[derive(Debug)]
 enum FsPathInner<'a> {
-    // Absolute path
+    /// An absolute path.
     Absolute(&'a str),
-    // Path is relative to `Cwd`
+    /// A relative path from the current working directory.
     CwdRelative(&'a str),
-    // Current working directory
+    /// The path of the current working directory.
     Cwd,
-    // Path is relative to the directory fd
+    /// A relative path from the directory FD (dirfd).
     FdRelative(FileDesc, &'a str),
-    // Fd
+    /// The path of the FD.
     Fd(FileDesc),
 }
 
@@ -739,6 +739,7 @@ impl<'a> TryFrom<&'a str> for FsPath<'a> {
         FsPath::from_fd_and_path(AT_FDCWD, path)
     }
 }
+
 /// Utilities to split a string into its path components.
 pub trait SplitPath {
     /// Splits a path into the parent directory name and the final component name, which is
```

kernel/src/fs/path/mod.rs
```diff
@@ -260,6 +172,11 @@ impl Path {
         Some(corresponding_path)
     }
 
+    /// Returns true if the `Path` represents a pseudo file.
+    pub(super) fn is_pseudo(&self) -> bool {
```
lrh2000: Actually I don't think it needs to be `pub(super)` anymore since `resolver.rs` lives under the `path` module.

kernel/src/syscall/getcwd.rs
```diff
@@ -3,14 +3,29 @@
...
+    let abs_path = {
+        let fs_ref = ctx.thread_local.borrow_fs();
+        let path_resolver = fs_ref.resolver().read();
+        path_resolver.make_abs_path(path_resolver.cwd())
+    };
 
-    let cwd = CString::new(name)?;
+    debug!("getcwd: {:?}", abs_path);
+
+    // Linux will add a "(unreachable)" prefix to the path if the cwd is not reachable.
```
lrh2000:
```suggestion
    // Linux will add a "(unreachable)" prefix to the path if the CWD is not reachable.
```

test/initramfs/src/apps/chroot/setup.h
```diff
@@ -0,0 +1,29 @@
...
+#include <sys/mount.h>
+#include <fcntl.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "../test.h"
+
+int test_fd;
+
+FN_SETUP(create_chroot_env)
```
lrh2000: I have no idea why you created the `setup.h` and `cleanup.h` files. They're only meaningful if multiple tests share the same environment, but we only have one use case here. Do you plan to add multiple tests in the near future?

test/initramfs/src/apps/chroot/chroot_jail.c
```diff
@@ -0,0 +1,131 @@
...
+#include <limits.h>
+#include <sys/wait.h>
+
+#include "setup.h"
+
+// Helper function to check if a directory does NOT contain a specific entry
+static int dir_not_contains(const char *path, const char *entry_name)
+{
+	DIR *dir = CHECK(opendir(path));
+
```
lrh2000: Remove this blank line.

lrh2000 (commented):
Only some style comments...

tatetian (approved):

PR #2795, opened 2025-12-25
Support QOM format parsing and remove the TDX object hardcoding

### LINKED ISSUE #2467: QOM Format Not Supported in QEMU Command-Line Parsing
When enabling TDX attestation, we need to add a QEMU command like:
```bash
-object '{ "qom-type": "tdx-guest", "id": "tdx0", "sept-ve-disable": true, "quote-generation-socket": { "type": "vsock", "cid": "2", "port": "4050" } }'
```
However, our current QEMU argument parsing does not support this format due to the following reasons:
- The parser cannot extract keys from arguments containing multiple separators (e.g., space).
https://github.com/asterinas/asterinas/blob/fc114019d6caf575e77f60d271bef3403b15b6cd/osdk/src/config/unix_args.rs#L133-L135
- `shlex::split` removes quotes from arguments. It is a bit ugly to reconstruct the original quoted string.
https://github.com/asterinas/asterinas/blob/fc114019d6caf575e77f60d271bef3403b15b6cd/osdk/src/config/unix_args.rs#L13-L20

This limitation prevents us from passing QOM objects directly in the command line, which is required for TDX attestation.

taosue:
This PR fixes #2467.

### DIFF
--- a/book/src/osdk/guide/intel-tdx.md
+++ b/book/src/osdk/guide/intel-tdx.md
@@ -51,7 +51,7 @@ supported_archs = ["x86_64"]
 boot.method = "grub-qcow2"
 grub.mkrescue_path = "~/tdx-tools/grub"
 grub.boot_protocol = "linux"
-qemu.args = """\
+qemu.args = '''\
     -accel kvm \
     -m 8G \
     -vga none \
@@ -60,11 +60,12 @@ qemu.args = """\
     -drive file=target/osdk/asterinas/asterinas.qcow2,if=virtio,format=qcow2 \
     -monitor telnet:127.0.0.1:9001,server,nowait \
     -bios /root/ovmf/release/OVMF.fd \
+    -object '{ \"qom-type\": \"tdx-guest\", \"id\": \"tdx0\", \"sept-ve-disable\": true, \"quote-generation-socket\": { \"type\": \"vsock\", \"cid\": \"2\", \"port\": \"4050\" } }' \
     -cpu host,-kvm-steal-time,pmu=off \
     -machine q35,kernel-irqchip=split,confidential-guest-support=tdx0 \
     -smp 1 \
     -nographic \
-"""
+'''
 ```
 
 To choose the configurations specified by the TDX scheme over the default ones,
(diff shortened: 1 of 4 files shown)


osdk/src/config/unix_args.rs
```diff
@@ -2,22 +2,15 @@
...
 /// Positional arguments are not supported.
 pub fn split_to_kv_array(args: &str) -> Vec<String> {
-    let target = match shlex::split(args) {
-        Some(v) => v,
-        None => {
-            error_msg!("Failed to parse unix args: {:#?}", args);
-            process::exit(Errno::ParseMetadata as _);
-        }
-    };
+    let target = split_preserving_quotes(args);
```
StevenJiang1110: I think we should stick with `shlex::split` if possible, as implementing a custom shell parser is inherently complex and error-prone.

That said, I noticed the `shlex` crate provides a `Quoter` struct to help with string quoting. Could you try integrating it into the `split_to_kv_array` function as shown in the patch below? It would allow us to keep using `shlex::split` while fixing the current problem.

```patch
diff --git a/osdk/src/config/unix_args.rs b/osdk/src/config/unix_args.rs
index 3a63276d5..b46e503f9 100644
--- a/osdk/src/config/unix_args.rs
+++ b/osdk/src/config/unix_args.rs
@@ -28,7 +28,9 @@ pub fn split_to_kv_array(args: &str) -> Vec<String> {
             && let Some(last) = joined.last_mut()
         {
             last.push(' ');
-            last.push_str(&elem);
+            let quoter = shlex::Quoter::new();
+            let quoted = quoter.quote(elem.as_str()).unwrap();
+            last.push_str(&quoted);
             last_has_value = true;
             continue;
         }
```


I’ve also written a test case confirming that `split_to_kv_array` preserves quotes correctly with this approach. It would also be beneficial to include a unit test like this in your PR to clearly demonstrate the problem you are addressing.

```rust
    #[test]
    fn test_split_kv_array() {
        let args = "-enable-kvm -object '{ \"qom-type\": \"tdx-guest\", \"id\": \"tdx0\", \"sept-ve-disable\": true, \"quote-generation-socket\": { \"type\": \"vsock\", \"cid\": \"2\", \"port\": \"4050\" } }'";

        let expected = &[
            "-enable-kvm",
            "-object '{ \"qom-type\": \"tdx-guest\", \"id\": \"tdx0\", \"sept-ve-disable\": true, \"quote-generation-socket\": { \"type\": \"vsock\", \"cid\": \"2\", \"port\": \"4050\" } }'",
        ];

        let array = split_to_kv_array(args);
        let expected: Vec<_> = expected.iter().map(ToString::to_string).collect();
        assert_eq!(expected, array);
    }
```

Issue #2776, opened 2025-12-20
Remove a formatted disk image file in ext2 fs will cause kernel panic


LordaeronESZ:
### Describe the bug

When I try to remove a 256MB-large ext4 disk image file used to boot a QEMU-TCG emulator in ext2 mount point, it will cause kernel panic.

### To Reproduce

1. store a 256MB-large ext4 disk image file into `test/build/ext2.img`, eg:
    1. mount ext2.img tmp/
    2. cp rootfs.ext4 tmp/
    3. umount tmp
2. `make run_kernel` to boot Asterinas
3. `cd ext2`
4. `rm rootfs.ext4` and `sync` -> trigger kernel panic!

### Screenshots

<img width="2559" height="908" alt="Image" src="https://github.com/user-attachments/assets/465a132d-ea8d-45d3-a215-a27f7fcfa0c1" />

### Environment

CPU: AMD Ryzen 8845HS
OS: Ubuntu 24.04.3 LTS with 5.15.167.4-microsoft-standard-WSL2 kernel

### Logs

``` bash
~ # ls
benchmark  etc        lib        opt        sbin       tmp
bin        exfat      lib64      proc       sys        usr
dev        ext2       nix        root       test       var
~ # cd ext2/
/ext2 # ls
lost+found   rootfs.ext4  static-tcg
/ext2 # rm rootfs.ext4
/ext2 # sync
[   116.696] ERROR: Uncaught panic:
        assertion failed: inner.metadata.is_block_allocated(idx)
        at /root/asterinas/kernel/src/fs/ext2/block_group.rs:212
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom:)
Printing stack trace:
   1: fn 0xffffffff88953b50 - pc 0xffffffff88953b9b / registers:

     rax 0xffffdfffffd6fb38; rdx 0xffffffff889b6352; rcx  0x10000020100d1cf; rbx                0x0;
     rsi                0x0; rdi 0xffffffff88bd6fd8; rbp                0x0; rsp 0xffffdfffffd6fc20;

   2: fn 0xffffffff8804c3d0 - pc 0xffffffff8804cbcd / registers:

     rax 0xffffdfffffd6fb38; rdx 0xffffffff889b6352; rcx  0x10000020100d1cf; rbx                0x0;
[...]

PR #2796, opened 2025-12-25
Fix the conversion of `blocks_count` field between `InodeDesc` and `RawInode`.

LordaeronESZ:
In ext2, the granule of the `blocks_count` field in disk-inode is fixed at 512B. In the current implementation of Asterinas, the `blocks_count` granule of memory-inode is BLOCK_SIZE. However, when we read/write inode between memory and disk, no conversion is performed.

### DIFF
--- a/kernel/src/fs/ext2/inode.rs
+++ b/kernel/src/fs/ext2/inode.rs
@@ -1488,7 +1488,7 @@ impl InodeImpl {
                 self.fs().free_blocks(device_range).unwrap();
                 return Err(e);
             }
-            self.desc.blocks_count = range.start + device_range.len() as Ext2Bid;
+            self.desc.sector_count = blocks_to_sectors(range.start + device_range.len() as u32);
             self.last_alloc_device_bid = Some(device_range.end - 1);
             return Ok(device_range.len() as Ext2Bid);
         }
@@ -1534,7 +1534,7 @@ impl InodeImpl {
             return Err(e);
         }
 
-        self.desc.blocks_count = range.start + device_range.len() as Ext2Bid;
+        self.desc.sector_count = blocks_to_sectors(range.start + device_range.len() as u32);
         self.last_alloc_device_bid = Some(device_range.end - 1);
         Ok(device_range.len() as Ext2Bid)
     }
@@ -1696,7 +1696,7 @@ impl InodeImpl {
             current_range.end -= free_cnt;
         }
 
-        self.desc.blocks_count = range.start;
+        self.desc.sector_count = blocks_to_sectors(range.start);
         self.last_alloc_device_bid = if range.start == 0 {
             None
         } else {
@@ -2143,8 +2143,8 @@ pub(super) struct InodeDesc {
     dtime: Duration,
     /// Hard links count.
     hard_links: u16,
-    /// Number of blocks.
-    blocks_count: Ext2Bid,
+    /// Number of sectors.
+    sector_count: u32,
     /// File flags.
     flags: FileFlags,
     /// Pointers to blocks.
@@ -2173,7 +2173,7 @@ impl TryFrom<RawInode> for InodeDesc {
             mtime: Duration::from(inode.mtime),
             dtime: Duration::from(inode.dtime),
             hard_links: inode.hard_links,
-            blocks_count: inode.blocks_count,
+            sector_count: inode.sector_count,
             flags: FileFlags::from_bits(inode.flags)
                 .ok_or(Error::with_message(Errno::EINVAL, "invalid file flags"))?,
             block_ptrs: inode.block_ptrs,
@@ -2201,7 +2201,7 @@ impl InodeDesc {
             mtime: now,
             dtime: Duration::ZERO,
             hard_links: 1,
-            blocks_count: 0,
+            sector_count: 0,
             flags: FileFlags::empty(),
             block_ptrs: BlockPtrs::default(),
             acl: match type_ {
@@ -2220,7 +2220,7 @@ impl InodeDesc {
     /// Ext2 allows the `block_count` to exceed the actual number of blocks utilized.
     pub fn blocks_count(&self) -> Ext2Bid {
         let blocks = self.size_to_blocks(self.size);
-        assert!(blocks <= self.blocks_count);
+        debug_assert!(blocks <= sectors_to_blocks(self.sector_count));
         blocks
     }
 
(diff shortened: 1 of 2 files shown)


kernel/src/fs/ext2/inode.rs
```diff
@@ -2173,7 +2175,7 @@ impl TryFrom<RawInode> for InodeDesc {
             mtime: Duration::from(inode.mtime),
             dtime: Duration::from(inode.dtime),
             hard_links: inode.hard_links,
-            blocks_count: inode.blocks_count,
+            blocks_count,
```
tatetian: Let's rename this field to `sector_count` (not in the plural form) to avoid making the same mistake in the future. And for consistency, could you also rename `blocks_count` to `block_count` (in a separate commit)?

kernel/src/fs/ext2/inode.rs
```diff
@@ -2359,6 +2361,8 @@ pub(super) struct RawInode {
 
 impl From<&InodeDesc> for RawInode {
     fn from(inode: &InodeDesc) -> Self {
+        // Convert from filesystem blocks to 512-byte sectors
+        let blocks_count = inode.blocks_count * (BLOCK_SIZE / 512) as u32;
```
tatetian: Use `SECTOR_SIZE` (defined in the block component) rather than 512.

kernel/src/fs/ext2/inode.rs
```diff
@@ -2359,6 +2361,8 @@ pub(super) struct RawInode {
 
 impl From<&InodeDesc> for RawInode {
     fn from(inode: &InodeDesc) -> Self {
+        // Convert from filesystem blocks to 512-byte sectors
```
tatetian: **Principle: do not write good comments to compensate bad code.** So remove this line of worthless comment. If the variables and const values are named properly, the intention of the line should become crystal clear, making any explicit explanation looks redundant.

```rust
let sector_count = inode.block_count * (BLOCK_SIZE / SECTOR_SIZE) as u32;
```

To push this idea even further, wrap it into a utility function:

```rust
fn sectors_to_blocks(sector_count: u32) -> Option<u32> { todo!("trivial") }

fn blocks_to_sectors(block_count: u32) -> u32 { todo!("trivial") }
```

This way, the code explains for itself, without having a comment.

tatetian (commented):
Great catch! This shows that how little Ext2 has been tested for re-opening a Ext2 instance that is previously sync'ed to disk by our implementation.

kernel/src/fs/ext2/inode.rs
tatetian: I am wondering if using two units in the in-memory and on-disk representations, respectively, is a good idea. One consequence of having two different units is the risk of integer overflow at this line.

We should either (1) change `Inode::block_count` to `Inode::sector_count` or (2) guard against `Inode::block_count` exceeding a new const `INODE_MAX_BLOCKS`, which is equal to `u32::max() / (BLOCK_SIZE / SECTOR_SIZE)`.

LordaeronESZ: I think the former might be more suitable, which is also the approach taken by Linux.
``` c
/* 
 * Linux 6.18 source code
 * fs/ext2/inode.c:1387
 */
struct inode *ext2_iget (struct super_block *sb, unsigned long ino)
{
    ...
    inode->i_blocks = le32_to_cpu(raw_inode->i_blocks);
    ...
}
```

LordaeronESZ:
Thanks, I have modified my code based on your detailed suggestions.

kernel/src/fs/ext2/inode.rs
```diff
@@ -2230,6 +2232,21 @@ impl InodeDesc {
         }
         size.div_ceil(BLOCK_SIZE) as Ext2Bid
     }
+
+    fn sectors_to_blocks(sector_count: u32) -> Ext2Bid {
```
tatetian: Both `sectors_to_blocks` and `blocks_to_sectors` have no inherent connection with `InodeDesc`. They should be made static utility functions, rather than methods.

kernel/src/fs/ext2/inode.rs
```diff
@@ -2220,7 +2222,7 @@ impl InodeDesc {
     /// Ext2 allows the `block_count` to exceed the actual number of blocks utilized.
     pub fn blocks_count(&self) -> Ext2Bid {
         let blocks = self.size_to_blocks(self.size);
-        assert!(blocks <= self.blocks_count);
+        assert!(blocks <= InodeDesc::sectors_to_blocks(self.sector_count));
```
tatetian: The original code should have used `debug_assert`. Could you fix it?

kernel/src/fs/ext2/inode.rs
```diff
@@ -2230,6 +2232,21 @@ impl InodeDesc {
...
+
+    fn sectors_to_blocks(sector_count: u32) -> Ext2Bid {
+        sector_count.div_ceil((BLOCK_SIZE / SECTOR_SIZE) as u32)
+    }
+
+    /// Converts block count to sector count with overflow protection.
+    ///
+    /// # Panics
+    /// Panics if the conversion would cause u32 overflow.
+    fn blocks_to_sectors(block_count: u32) -> u32 {
```
tatetian: For trivial internal functions, you don't have to add Rust doc. As I said before, do not write worthless comments. Using four lines of comments to explain four lines of code, without providing an extra information, is not a good bargain.

tatetian (commented):
Thanks for the revision. The PR looks almost perfect, only two minor issues left.

By the way, after fixing this bug, do you witness any Ext2 failures go away? I meant did you find this bug by experiments or by reviewing?

LordaeronESZ:
Thanks, I have modified my code again based on your suggestions.

---
> By the way, after fixing this bug, do you witness any Ext2 failures go away? I meant did you find this bug by experiments or by reviewing?

I found this bug when I was debugging #2776 using GDB, and following strange case was discovered:



Well, it certainly won't fail the assertion, but shouldn't the two sides be equal under normal circumstances? So I read the relevant code in depth and finally found it.

tatetian:
> Thanks, I have modified my code again based on your suggestions.
> 
> > By the way, after fixing this bug, do you witness any Ext2 failures go away? I meant did you find this bug by experiments or by reviewing?
> 
> I found this bug when I was debugging #2776 using GDB, and following strange case was discovered:
> 
> 
> Well, it certainly won't fail the assertion, but shouldn't the two sides be equal under normal circumstances? So I read the relevant code in depth and finally found it.

Ok. Great. Thanks for the input. You can add the following line to the description of this PR so that after this PR is merged, the issue will be closed automatically.

```
Closes #2776
```

LordaeronESZ:
> You can add the following line to the description of this PR so that after this PR is merged, the issue will be closed automatically.
```
Closes #2776 
```

Thanks, I might need to explain that the issue this PR fixes is not the root cause of the #2776, but just a byproduct of debugging it. I'm currently working on a PR to resolve #2776 (and I created a related feature request in #2807) and plan to push it soon.

tatetian (approved):

Issue #2625, opened 2025-11-27
Add pseudo `Path`s


vvvvsv:
TODO:

1. Add pseudo `Dentry` and `Path` for each `FileLike` object. https://github.com/asterinas/asterinas/pull/2521#issuecomment-3460571218
2. Remove `MemfdFile` and keep only `InodeHandle(PseudoPath(MemfdInode))`. https://github.com/asterinas/asterinas/pull/2555#issuecomment-3516482891

PR #2798, opened 2025-12-25
Add pseudo `Path`

vvvvsv:
~Update: Should go after https://github.com/asterinas/asterinas/pull/2850.~

---

This PR adds pseudo `Path`s for anonymous pipes, sockets, memfd files, and `anon_inodefs` members to improve Linux compatibility, and includes many cleanups.

~The commit "Remove `inode_handle/dyn_cap.rs` and `HandleInner`" is a cleanup for https://github.com/asterinas/asterinas/pull/2563#discussion_r2517648267.~

~The commit "Polling a `fd` opened with `O_PATH` should return `POLLNVAL`" is a bug-fix.~

The rest of this PR adds pseudo `Dentry`, `Mount` and `Path` for each `FileLike` object, which completes the "final plan" in https://github.com/asterinas/asterinas/pull/2521#issuecomment-3460571218. 

whiteboard_exported_image

After this, all `FileLike` implementations can be migrated to the `InodeHandle & FileIo` method, and the `trait FileLike` can be deprecated. To reduce code complexity, this PR applies this change only to `FileLike` objects that can be reopened from a `Path` (anonymous pipes and memfd files), and removes `AnonPipeFile` and `MemfdFile` (which was discussed in https://github.com/asterinas/asterinas/pull/2555#issuecomment-3516482891).

Closes https://github.com/asterinas/asterinas/issues/2625.

### DIFF
--- a/kernel/src/fs/inode_handle.rs
+++ b/kernel/src/fs/inode_handle.rs
@@ -2,10 +2,7 @@
 
 //! Opened Inode-backed File Handle
 
-use core::{
-    fmt::Display,
-    sync::atomic::{AtomicU32, Ordering},
-};
+use core::{fmt::Display, sync::atomic::Ordering};
 
 use aster_rights::Rights;
 
@@ -18,8 +15,8 @@ use crate::{
         path::Path,
         pipe::PipeHandle,
         utils::{
-            AccessMode, CreationFlags, DirentVisitor, FallocMode, FileRange, FlockItem, InodeType,
-            OFFSET_MAX, RangeLockItem, RangeLockType, SeekFrom, StatusFlags,
+            AccessMode, AtomicStatusFlags, CreationFlags, DirentVisitor, FallocMode, FileRange,
+            FlockItem, InodeType, OFFSET_MAX, RangeLockItem, RangeLockType, SeekFrom, StatusFlags,
         },
     },
     prelude::*,
@@ -34,7 +31,7 @@ pub struct InodeHandle {
     /// be provided by `file_io`, instead of `path`.
     file_io: Option<Box<dyn FileIo>>,
     offset: Mutex<usize>,
-    status_flags: AtomicU32,
+    status_flags: AtomicStatusFlags,
     rights: Rights,
 }
 
@@ -71,7 +68,7 @@ impl InodeHandle {
             path,
             file_io,
             offset: Mutex::new(0),
-            status_flags: AtomicU32::new(status_flags.bits()),
+            status_flags: AtomicStatusFlags::new(status_flags),
             rights,
         })
     }
@@ -360,8 +357,7 @@ impl FileLike for InodeHandle {
     }
 
     fn status_flags(&self) -> StatusFlags {
-        let bits = self.status_flags.load(Ordering::Relaxed);
-        StatusFlags::from_bits(bits).unwrap()
+        self.status_flags.load(Ordering::Relaxed)
     }
 
     fn set_status_flags(&self, new_status_flags: StatusFlags) -> Result<()> {
@@ -377,8 +373,7 @@ impl FileLike for InodeHandle {
             crate::fs::pipe::check_status_flags(new_status_flags)?;
         }
 
-        self.status_flags
-            .store(new_status_flags.bits(), Ordering::Relaxed);
+        self.status_flags.store(new_status_flags, Ordering::Relaxed);
 
         Ok(())
     }
--- a/kernel/src/fs/utils/mod.rs
+++ b/kernel/src/fs/utils/mod.rs
@@ -23,7 +23,7 @@ pub use page_cache::{CachePage, PageCache, PageCacheBackend};
 #[cfg(ktest)]
 pub use random_test::{generate_random_operation, new_fs_in_memory};
 pub use range_lock::{FileRange, OFFSET_MAX, RangeLockItem, RangeLockList, RangeLockType};
-pub use status_flags::StatusFlags;
+pub use status_flags::{AtomicStatusFlags, StatusFlags};
 pub use xattr::{
     XATTR_LIST_MAX_LEN, XATTR_NAME_MAX_LEN, XATTR_VALUE_MAX_LEN, XattrName, XattrNamespace,
     XattrSetFlags,
--- a/kernel/src/fs/utils/status_flags.rs
+++ b/kernel/src/fs/utils/status_flags.rs
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use core::sync::atomic::AtomicU32;
+
+use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
 use bitflags::bitflags;
 
 bitflags! {
@@ -23,3 +26,21 @@ bitflags! {
         const O_PATH = 1 << 21;
     }
 }
+
+impl From<u32> for StatusFlags {
+    fn from(value: u32) -> Self {
+        Self::from_bits_truncate(value)
+    }
+}
+
+impl From<StatusFlags> for u32 {
+    fn from(value: StatusFlags) -> Self {
+        value.bits()
+    }
+}
+
+define_atomic_version_of_integer_like_type!(StatusFlags, {
+    /// An atomic version of `StatusFlags`.
+    #[derive(Debug)]
+    pub struct AtomicStatusFlags(AtomicU32);
+});


kernel/src/fs/pseudofs.rs
```diff
@@ -57,23 +61,38 @@ impl PseudoFs {
...
+        type_: InodeType,
+        mode: InodeMode,
+        uid: Uid,
+        gid: Gid,
+    ) -> PseudoInode {
+        PseudoInode::new(self.alloc_id(), type_, mode, uid, gid, Arc::downgrade(self))
+    }
+
+    fn alloc_id(&self) -> u64 {
+        self.inode_allocator.fetch_add(1, Ordering::Relaxed)
```
cchanging, line 97: I remember that `ino` should all be recyclable? Maybe using `IdAlloc` here would be better.

kernel/src/fs/pseudofs.rs
```diff
@@ -112,7 +131,18 @@ fn anon_inodefs_singleton() -> &'static Arc<PseudoFs> {
 //
 // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/fs/anon_inodes.c#L153-L164>
 pub fn anon_inodefs_shared_inode() -> &'static Arc<dyn Inode> {
-    &anon_inodefs_singleton().root
+    static SHARED_INODE: Once<Arc<dyn Inode>> = Once::new();
```
cchanging: So the root inode of `anon_inodefs_singleton` will be unused? What's the reason for not using it?

kernel/src/fs/path/mount.rs
```diff
@@ -217,6 +220,31 @@ impl Mount {
...
+    pub(in crate::fs) fn new_pseudo(fs: Arc<dyn FileSystem>) -> Arc<Self> {
+        let id = ID_ALLOCATOR.get().unwrap().lock().alloc().unwrap();
+        let root_dentry = Dentry::new_root(fs.root_inode());
+        root_dentry.inc_mount_count();
+
+        Arc::new_cyclic(|weak_self| Self {
+            id,
+            root_dentry: root_dentry.clone(),
+            mountpoint: RwLock::new(Some(root_dentry)),
+            parent: RwLock::new(Some(weak_self.clone())),
```
cchanging: > has no parent

Why the `parent` points to the `Mount` itself? In addtion, the `mountpoint` should be `None` since the `Mount` has no parent.

I think the input parameter can be `pseudo_fs: Arc<PseudoFs>`, and the implementation can reuse `Mount::new` method.

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -233,12 +233,23 @@ impl Inode for MemfdInode {
...
+fn memfd_tmpfs_singleton() -> &'static Arc<TmpFs> {
+    static MEMFD_TMPFS: Once<Arc<TmpFs>> = Once::new();
+
+    MEMFD_TMPFS.call_once(TmpFs::new)
+}
+
+fn memfd_tmpfs_mount() -> &'static Arc<Mount> {
+    static MEMFD_TMPFS_MOUNT: Once<Arc<Mount>> = Once::new();
+
+    MEMFD_TMPFS_MOUNT.call_once(|| Mount::new_pseudo(memfd_tmpfs_singleton().clone()))
```
cchanging: The `Mount` for memfd should be more special.  Through fdinfo, the mount ID we get for memfd would be 0, which represents an invalid mount.

kernel/src/fs/path/dentry.rs
```diff
@@ -16,13 +16,26 @@ use crate::{
...
+    Pseudo(PseudoDentryInner),
+}
+
+struct RealDentryInner {
+    name_and_parent: RwLock<Option<(String, Arc<Dentry>)>>,
+    children: RwMutex<DentryChildren>,
+}
+
+struct PseudoDentryInner {
+    name_fn: fn(&dyn Inode) -> String,
```
cchanging: It is kind of weird to store such a function here. Why not calculate and store the name directly? Is it to save memory space? I don't know if this is worth it.

kernel/src/fs/path/dentry.rs
```diff
@@ -56,27 +91,40 @@ impl Dentry {
...
+                .name_and_parent
+                .read()
+                .as_ref()
+                .map(|name_and_parent| name_and_parent.1.clone())
+        })
     }
 
     fn set_name_and_parent(&self, name: &str, parent: Arc<Self>) {
-        let mut name_and_parent = self.name_and_parent.write();
+        let mut name_and_parent = self.real_inner().unwrap().name_and_parent.write();
```
cchanging: Please double check the `unwrap()` usages in this file. Maybe adding some documents to clarify why they will not fail?

kernel/src/fs/pseudofs.rs
vvvvsv: `IdAlloc` uses a bitmap to track whether an ID is available. However, `ino` is unbounded (can go up to `u64::MAX`), which is impossible to manage with a bitmap. As far as I remember, `ino`s are not required to be recyclable, and they only need to be unique within a filesystem?

kernel/src/fs/pseudofs.rs
vvvvsv: > So the root inode of `anon_inodefs_singleton` will be unused?

Yes. Actually all the roots of pseudo fses will only be used in the pseudo mounts.

> What's the reason for not using it?

Please see the next reply first. 

This is to align with Linux: https://elixir.bootlin.com/linux/v6.16.5/source/fs/anon_inodes.c#L370

kernel/src/fs/path/mount.rs
vvvvsv: This part of code is just copying the logic of `kern_mount` in Linux.

> Root Dentry and Root Inode
https://elixir.bootlin.com/linux/v6.16.5/source/fs/namespace.c#L1310
https://elixir.bootlin.com/linux/v6.16.5/source/fs/fs_context.c#L318
https://elixir.bootlin.com/linux/v6.16.5/source/fs/anon_inodes.c#L84-L99
https://elixir.bootlin.com/linux/v6.16.5/source/fs/libfs.c#L719 (This sets `fc->ops`)
https://elixir.bootlin.com/linux/v6.16.5/source/fs/namespace.c#L1290
https://elixir.bootlin.com/linux/v6.16.5/source/fs/super.c#L1804 (This calls `fc->ops->get_tree`, i.e. https://elixir.bootlin.com/linux/v6.16.5/source/fs/libfs.c#L691)
https://elixir.bootlin.com/linux/v6.16.5/source/fs/libfs.c#L672-L684 (Sets the root inode and root dentry)

> Pseudo Mounts
https://elixir.bootlin.com/linux/v6.16.5/source/fs/namespace.c#L6301-L6314
https://elixir.bootlin.com/linux/v6.16.5/source/fs/namespace.c#L1252-L1327 (This points `parent` to `mnt` itself, and `mountpoint` to the pseudo root dentry)

At the moment I don’t know what these fields are used for, but in the future we may find that Linux exposes or relies on this information in some places. Given that the cost of adding a few pseudo roots and mounts (only one for a pseudo fs) is very low, I simply copied the Linux logic.

kernel/src/fs/ramfs/memfd.rs
vvvvsv: The mount for memfd files is also get via `kern_mount`.

https://elixir.bootlin.com/linux/v6.16.5/source/mm/shmem.c#L5438

As for why their `mnt_id` is 0, I will double-check that, maybe 0 is also a minimal valid mount ID?

kernel/src/fs/path/dentry.rs
vvvvsv: > Is it to save memory space?

Yes. If we have 100 pipes, I want to store only 100 function pointers pointing to the same function, rather than 100 strings. This is also Linux approach: https://elixir.bootlin.com/linux/v6.16.5/source/fs/pipe.c#L869-L880h

kernel/src/fs/path/dentry.rs
vvvvsv: All `real_inner().unwrap()` calls are made on a `Dentry` of type `InodeType::Dir` or on a `Dentry` that is a child of another `Dentry`. If either type of `Dentry` is a pseudo `Dentry`, it can only be the root `Dentry` of a pseudo filesystem, and we never expose these roots to users.

To be honest, I have no idea how to write this document... Maybe it is better to return a static empty `RwMutex<DentryChildren>` for `children` calls on pseudo dentries, and log a warning?

kernel/src/fs/ramfs/memfd.rs
vvvvsv: > maybe 0 is also a minimal valid mount ID?

Yes. Mount IDs are allocated from an [Xarray](https://elixir.bootlin.com/linux/v6.16.5/source/fs/namespace.c#L69) via [mnt_alloc_id](https://elixir.bootlin.com/linux/v6.16.5/source/fs/namespace.c#L276), thus can be 0. This method is first called when creating the `shm_mnt` associated with memfds.

```
(gdb) c
Continuing.

Breakpoint 1, mnt_alloc_id (mnt=0xffff88800393c000) at fs/namespace.c:280
280             xa_lock(&mnt_id_xa);
(gdb) bt
#0  mnt_alloc_id (mnt=0xffff88800393c000) at fs/namespace.c:280
#1  alloc_vfsmnt (name=0xffff888003846070 "tmpfs") at fs/namespace.c:354
#2  0xffffffff814ed73b in vfs_create_mount (fc=fc@entry=0xffff88800385d300) at fs/namespace.c:1267
#3  0xffffffff814ed900 in fc_mount (fc=0xffff88800385d300) at fs/namespace.c:1292
#4  fc_mount (fc=0xffff88800385d300) at fs/namespace.c:1287
#5  vfs_kern_mount (type=type@entry=0xffffffff82b68040 <shmem_fs_type>, flags=flags@entry=4194304, name=<optimized out>, data=data@entry=0x0) at fs/namespace.c:1319
#6  0xffffffff814ed9df in vfs_kern_mount (data=0x0, name=<optimized out>, flags=4194304, type=0xffffffff82b68040 <shmem_fs_type>) at fs/namespace.c:1306
#7  kern_mount (type=type@entry=0xffffffff82b68040 <shmem_fs_type>) at fs/namespace.c:6249
#8  0xffffffff8329f77a in shmem_init () at mm/shmem.c:5415
#9  0xffffffff832aff40 in mnt_init () at fs/namespace.c:6233
#10 0xffffffff832af96a in vfs_caches_init () at fs/dcache.c:3233
#11 0xffffffff83263dc2 in start_kernel () at init/main.c:1082
#12 0xffffffff8326f6b8 in x86_64_start_reservations (real_mode_data=real_mode_data@entry=0x14770 <error: Cannot access memory at address 0x14770>) at arch/x86/kernel/head64.c:513
#13 0xffffffff8326f7f6 in x86_64_start_kernel (real_mode_data=0x14770 <error: Cannot access memory at address 0x14770>) at arch/x86/kernel/head64.c:494
#14 0xffffffff8122f6b6 in secondary_startup_64 () at arch/x86/kernel/head_64.S:418
```

kernel/src/fs/pseudofs.rs
cchanging: >  As far as I remember, inos are not required to be recyclable, and they only need to be unique within a filesystem?

Ok, it seems that the ino for Linux's anonymous filesystems are recyclable but not mandatory. The current implementation is acceptable, as our implementations like `Ramfs` operate similarly. However, I suggest adding a `FIXME` note to highlight the potential for overflow.

kernel/src/fs/path/mount.rs
cchanging: > This part of code is just copying the logic of kern_mount in Linux.

This is indeed a bit odd. 

However, I feel like Linux might be arbitrarily assigning values to fields such as `parent` first, and then modifying them when actually attaching the mount? After all, Linux doesn’t have a `None`-like value, and simply setting them to `NULL` doesn’t seem appropriate. 

So, I think we can still handle these fields with more accurate semantics? If Linux actually has special ways of utilizing these fields, we can make modifications later.

kernel/src/fs/ramfs/memfd.rs
cchanging:
```c
static int mnt_alloc_id(struct mount *mnt)
{
	int res;

	xa_lock(&mnt_id_xa);
	res = __xa_alloc(&mnt_id_xa, &mnt->mnt_id, mnt, XA_LIMIT(1, INT_MAX), GFP_KERNEL);
	if (!res)
		mnt->mnt_id_unique = ++mnt_id_ctr;
	xa_unlock(&mnt_id_xa);
	return res;
}
```

Here Linux uses `XA_LIMIT(1, INT_MAX)` to limit the minimal valid id to 1, so I think 0 is still an invalid ID?

kernel/src/fs/pseudofs.rs
vvvvsv: > it seems that the ino for Linux's anonymous filesystems are recyclable

I just double-checked this. Linux's anonymous fses do not recycle `ino`s. They use a [incrementing per-CPU variable](https://elixir.bootlin.com/linux/v6.16.5/source/fs/inode.c#L1087-L1127) to alloc `ino`s.

kernel/src/fs/path/dentry.rs
cchanging: Thanks for the explaination! I think some documents can be added to the usages of `real_inner` method. This way, unwrapping doesn't seem as intimidating😂.

kernel/src/fs/ramfs/memfd.rs
vvvvsv: > I think 0 is still an invalid ID

Thanks, I think so. The weird thing is, when I run the Linux kernel directly in QEMU, the `mnt_id` of the `memfd` I got is 1:

```
~ # ./memfd_mnt_id
memfd fd = 3
mnt_id: 1
```

What's even weirder is that this `mnt_id` value is different in different distributions:

Debian/Ubuntu
```
wsy@avalon:~$ ./test
memfd fd = 3
mnt_id: 0
```

CentOS
```
[yuhs@localhost wsy]$ ./test
memfd fd = 3
mnt_id: 3
```

How about just following Linux kernel behavior?

kernel/src/fs/path/dentry.rs
vvvvsv: > some documents can be added to the usages of `real_inner` method

Okay. I plan to use `Dentry::new_root` for roots of pseudo fses, since they do not cost much memory (only one for a pseudo fs). This way we can do `real_inner().unwrap()` more safely.

kernel/src/fs/ramfs/memfd.rs
cchanging: Thanks for the experiments! Can you provide some additional information? For example, what is the Linux kernel version in your experiment, and could you also show the output of `cat /proc/self/mountinfo` to see if there is a mount entry corresponding to that ID?

kernel/src/fs/pseudofs.rs
cchanging: Thanks for the explanation!

kernel/src/fs/ramfs/memfd.rs
vvvvsv: > Linux kernel version

```
~ # cat /proc/version
Linux version 6.15.0-gcdfc8592b328 (vvsv@vvsv-XiaoXin) (gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #1 SMP PREEMPT_DYNAMIC Mon Dec 29 16:41:00 CST 2025
```

> cat /proc/self/mountinfo

```
~ # cat /proc/self/mountinfo
2 2 0:2 / / rw - rootfs rootfs rw,size=43424k,nr_inodes=10856
21 2 0:19 / /proc rw,relatime - proc proc rw
22 2 0:20 / /sys rw,relatime - sysfs sysfs rw
23 2 0:6 / /dev rw,relatime - devtmpfs devtmpfs rw,size=43424k,nr_inodes=10856,mode=755
```

It seems that `mountinfo` is printed by traversing the mount tree, while mounts created with `kern_mount` do not belong to the mount tree.

cchanging: > It seems that mountinfo is printed by traversing the mount tree, while mounts created with kern_mount do not belong to the mount tree.

I see. I'm actually mainly interested in knowing the corresponding Linux version and mountinfo for the cases where the mount ID is displayed as 1 or 3. I'm curious why differences between Linux distributions would affect such a low-level implementation.

vvvvsv: > mount ID is displayed as 1


```
# Directly run the Linux kernel
~ # cat /proc/version
Linux version 6.15.0-gcdfc8592b328 (vvsv@vvsv-XiaoXin) (gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #1 SMP PREEMPT_DYNAMIC Mon Dec 29 16:41:00 CST 2025
```
> displayed as 0

```
# Debian
wsy@avalon:~$ ./test
memfd fd = 3
mnt_id: 0
wsy@avalon:~$ cat /proc/version
Linux version 6.13.8 (hk@avalon) (gcc (Debian 12.2.0-14) 12.2.0, GNU ld (GNU Binutils for Debian) 2.40) #1 SMP PREEMPT_DYNAMIC Fri Mar 28 11:41:46 HKT 2025
```

```
# Ubuntu
vvsv@vvsv-XiaoXin:~/TMP$ ./tmp1
memfd fd = 3
mnt_id: 0
vvsv@vvsv-XiaoXin:~/TMP$ cat /proc/version
Linux version 6.6.87.2-microsoft-standard-WSL2 (root@439a258ad544) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025
```

> displayed as 3

```
# CentOS 7
[yuhs@localhost wsy]$ ./test
memfd fd = 3
mnt_id: 3
[yuhs@localhost wsy]$ cat /proc/version
Linux version 3.10.0-1160.42.2.el7.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) ) #1 SMP Tue Sep 7 14:49:57 UTC 2021
```

---

Sorry, I just compiled a Linux kernel of version 6.13.8 (the same as the Debian example above). It shows:

```
~ # cat /proc/version
Linux version 6.13.8 (vvsv@vvsv-XiaoXin) (gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #1 SMP PREEMPT_DYNAMIC Tue Dec 30 15:38:28 CST 2025
~ # ./memfd_mnt_id
memfd fd = 3
mnt_id: 0
```

In this version, 0 is a valid mnt_id. https://elixir.bootlin.com/linux/v6.13.8/source/fs/namespace.c#L237-L246

Let's just follow the latest version, where 0 is not a valid mnt_id: https://elixir.bootlin.com/linux/v6.19-rc3/source/fs/namespace.c#L224

kernel/src/fs/inode_handle.rs
```diff
@@ -413,13 +421,41 @@ impl FileLike for InodeHandle {
...
 
-        do_fallocate_util(
-            self.path.inode().as_ref(),
-            self.status_flags(),
-            mode,
-            offset,
-            len,
-        )
+        let inode = self.path.inode().as_ref();
+        let status_flags = self.status_flags();
```
cchanging: Move this line to 439. This way we load `status_flags` only when really using it.

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -237,280 +223,91 @@ impl Inode for MemfdInode {
...
-
-        do_resize_util(self.memfd_inode.as_ref(), self.status_flags(), new_size)
-    }
+    memfd_inode_or_err(file)?.add_seals(new_seals)
+}
 
-    fn status_flags(&self) -> StatusFlags {
-        let bits = self.status_flags.load(Ordering::Relaxed);
-        StatusFlags::from_bits(bits).unwrap()
+pub fn do_get_seals(file: &InodeHandle) -> Result<FileSeals> {
```
cchanging: I think naming these function `add_seals` and `get_seals` is better. We usually name internal implementation functions as `do_xx` only when implementing internal functionality for `sys_xx`.

kernel/src/fs/pseudofs.rs
```diff
@@ -172,6 +202,8 @@ impl FsType for SockFsType {
     }
 }
 
+/// Root Inode ID.
+const ROOT_INO: u64 = 1;
```
cchanging, line 278: Move this line to `PseudoFs::new` since this const is only used there.

kernel/src/fs/path/mod.rs
```diff
@@ -55,6 +55,16 @@ impl Path {
         Ok(Self::new(self.mount.clone(), new_child_dentry))
     }
 
+    /// Creates a new pseudo `Path`.
+    pub fn new_pseudo(
+        mount: Arc<Mount>,
+        inode: Arc<dyn Inode>,
+        name_fn: fn(&dyn Inode) -> String,
+    ) -> Self {
```
cchanging, line 63: There might be some usage issues with this API. While it's public, it cannot guarantee that the passed `mount` and `inode` correspond to each other. This allows users to create a path using a `memfd` inode with a `pipefs` mount anywhere in the kernel. 

I'm wondering if we should manage these pseudo filesystems with separate abstractions, like this:
```rust
pub struct PipeFs {
    _private: (),
}

impl PipeFs {
    pub fn singleton() -> &'static Arc<PseudoFs>;

    pub fn mount_node() -> &'static Arc<Mount>;

    pub fn new_path(inode: Arc<AnonPipeInode>) -> Path;
}

pub struct AnonInodeFs {
    _private: (),
}

impl AnonInodeFs {
    pub fn singleton() -> &'static Arc<PseudoFs>;

    pub fn mount_node() -> &'static Arc<Mount>;
    
    pub fn shared_inode() -> &'static Arc<PseudoInode>;
    
    pub fn shared_path() -> Path;
}
...
```

In this way, we only need to import a specific `xxFs` to use `singleton`, `mount_node` ... and we can turn the `new_pseudo` method to `pub(super)`.

cchanging (commented):
Thanks for the revision! I have only a few minor comments.

vvvvsv:
This force-push includes the changes discussed above. I will rebase to the main branch and force-push again later.

vvvvsv:
**Additional information:**

1. This PR now use `InodeHandle` to manage anonymous pipes to reduce code duplication. `InodeHandle` is too heavy for `ktests` (e.g., `RealTimeCoarseClock` and `mount::ID_ALLOCATOR` need to be initialized, and `RangeLockItem::new` requires a current process to be running). See this [run](https://github.com/asterinas/asterinas/actions/runs/20689286838/job/59394962792?pr=2798). Therefore, I moved the pipe ktests from `pipe::anon_pipe` to `pipe::common` and directly uses `PipeHandle` to test pipe behaviors.

> this PR applies this change only to `FileLike` objects that can be reopened from a `Path` (anonymous pipes and memfd files)

2. Note that in Linux, users cannot reopen sockets or anon inodes with read or write permissions, but can still reopen them with `O_PATH`. This is also supported in this PR, where the behavior is to open an `InodeHandle { path: pseudo_path, rights: Rights::empty() }`.

vvvvsv:
Resolved some conflicts.

vvvvsv:
@boterinas rerun

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -235,286 +221,101 @@ impl Inode for MemfdInode {
...
-
-impl MemfdTmpFs {
-    // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/mm/shmem.c#L3828-L3850>
-    fn singleton() -> &'static Arc<TmpFs> {
-        static MEMFD_TMPFS: Once<Arc<TmpFs>> = Once::new();
-
-        MEMFD_TMPFS.call_once(TmpFs::new)
+pub fn do_memfd_create(name: String, memfd_flags: MemfdFlags) -> Result<InodeHandle> {
+    if name.len() > MAX_MEMFD_NAME_LEN {
+        return_errno_with_message!(Errno::EINVAL, "MemfdInode: `name` is too long.");
```
cchanging: Our practice is to use plain phrases without ending punctuation for error messages.

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -235,286 +221,101 @@ impl Inode for MemfdInode {
...
 
-pub struct MemfdFile {
-    memfd_inode: Arc<dyn Inode>,
-    offset: Mutex<usize>,
-    status_flags: AtomicU32,
-    rights: Rights,
-}
+    let memfd_inode = Arc::new_cyclic(|weak_self| {
+        let ram_inode =
+            RamInode::new_file_detached_in_memfd(weak_self, mode, Uid::new_root(), Gid::new_root());
```
cchanging: Move `let mode = if executable {...` to this block? It seems that `mode` is only used here.

kernel/src/net/socket/mod.rs
```diff
@@ -208,14 +207,14 @@ impl<T: Socket + 'static> FileLike for T {
...
-    let pseudo_inode = SockFs::singleton().alloc_inode(
+/// Creates a pseudo `Path` for a socket.
+fn new_pseudo_path() -> Path {
+    let pseudo_inode = Arc::new(SockFs::singleton().alloc_inode(
         InodeType::Socket,
         mkmod!(a+rwx),
         Uid::new_root(),
         Gid::new_root(),
-    );
+    ));
```
cchanging: We can move this logics to `SockFs::new_path` and remove the helper function `new_pseudo_path`

kernel/src/fs/pipe/common.rs
```diff
@@ -272,6 +270,21 @@ impl Pipe {
     }
 }
 
+pub(in crate::fs) fn check_status_flags(status_flags: StatusFlags) -> Result<()> {
```
cchanging, line 273: To be honest, this function leaves me somewhat confused, especially regarding its special handling in `InodeHandle::set_status_flags`:
```rust
    fn set_status_flags(&self, new_status_flags: StatusFlags) -> Result<()> {
        if self
            .file_io
            .as_ref()
            .and_then(|file_io| (file_io.as_ref() as &dyn Any).downcast_ref::<PipeHandle>())
            .is_some()
        {
            // TODO: Support "packet" mode for pipes.
            crate::fs::pipe::check_status_flags(new_status_flags)?;
        }

        self.status_flags.store(new_status_flags, Ordering::Relaxed);

        Ok(())
    }
```

Is the `status_flag` checking specific to pipes? If not, I think this function should rather be a method of `FileIo`. Moreover, the newer Linux versions mentioned in the manual actually allow setting `O_DIRECT`, so I feel we need to add a corresponding `TODO`.

cchanging (commented):
Thanks for the revision!

kernel/src/fs/pipe/common.rs
vvvvsv: > Is the `status_flag` checking specific to pipes?

Yes. This function is currently only called when setting status flags for `AnonPipeFile`s. After switching to managing pipes with `InodeHandle`, I did not find an appropriate place to put this logic.

https://github.com/asterinas/asterinas/blob/main/kernel/src/fs/pipe/anon_pipe.rs#L162-L175



> the newer Linux versions mentioned in the manual actually allow setting `O_DIRECT`, so I feel we need to add a corresponding `TODO`.

The `O_DIRECT` flag indicates that the pipe should operate in ["packet" mode](https://man7.org/linux/man-pages/man2/pipe.2.html). Luckily, if we support "packet" mode in the future, we can remove this `downcast_ref` logic. Is this "TODO: Support "packet" mode for pipes." not enough?

cchanging: > The O_DIRECT flag indicates that the pipe should operate in ["packet" mode](https://man7.org/linux/man-pages/man2/pipe.2.html). Luckily, if we support "packet" mode in the future, we can remove this downcast_ref logic. Is this "TODO: Support "packet" mode for pipes." not enough?

Thank you for your explanation! Now I understand the meaning of this TODO here. I think we can move the "packet" TODO into the `check_status_flags` function and explain their relationship.

> Yes. This function is currently only called when setting status flags for AnonPipeFiles. After switching to managing pipes with InodeHandle, I did not find an appropriate place to put this logic.

I originally thought that different devices might have their own specific requirements for `status_flags`, so we could make `check_status_flags` a method of `FileIo`. However, after checking the Linux source code, I didn't find any related logic. Considering that after completing packet mode, we can remove the special case here, I think the current approach is acceptable.

kernel/src/fs/path/dentry.rs
```diff
@@ -18,13 +18,26 @@ use crate::{
 pub(super) struct Dentry {
     inode: Arc<dyn Inode>,
     type_: InodeType,
-    name_and_parent: RwLock<Option<(String, Arc<Dentry>)>>,
-    children: RwMutex<DentryChildren>,
+    inner: DentryInner,
```
lrh2000: I think a dentry's ability to have children is not determined by whether it is real or pseudo. Rather, it is determined by the inode type. For example, a dentry of the inode type `File` can have no children.

I also agree that [there seem to be too many `unwrap()`s related to children](https://github.com/asterinas/asterinas/pull/2798#discussion_r2650138089), which is difficult to understand, especially given that sometimes `Inode::Dir` is checked in `dentry.rs` and sometimes in `path.rs`.

So is it better to keep the `children` field as it is? I think it's something independent of this PR given that a file dentry can also have no children but it does have the `children` field.

kernel/src/fs/path/dentry.rs
```diff
@@ -18,13 +18,26 @@ use crate::{
...
     this: Weak<Dentry>,
 }
 
+enum DentryInner {
+    Real(RealDentryInner),
+    Pseudo(PseudoDentryInner),
+}
+
+struct RealDentryInner {
+    name_and_parent: RwLock<Option<(String, Arc<Dentry>)>>,
```
lrh2000: Given that a root dentry can never have a name or a parent, this seems to be an `Option<RwLock<(String, Arc<Dentry))>>`?

kernel/src/fs/path/dentry.rs
```diff
@@ -58,27 +93,45 @@ impl Dentry {
...
+                .name_and_parent
+                .read()
+                .as_ref()
+                .map(|name_and_parent| name_and_parent.1.clone())
+        })
     }
 
     fn set_name_and_parent(&self, name: &str, parent: Arc<Self>) {
-        let mut name_and_parent = self.name_and_parent.write();
+        let mut name_and_parent = self.real_inner().unwrap().name_and_parent.write();
```
lrh2000: I think we can `unwrap()` at the caller. Something like (please add some comments to explain when the `SetNameAndParentError` will occur):
```rust
struct SetNameAndParentError;

impl NameAndParent {
    fn set(&self, name: &str, parent: Arc<Self>) -> Result<(), SetNameAndParentError>
}
```

I skimmed through the call sites, where the caller always takes a child from the `children` field. Therefore, it's reasonable to assume that `set` won't fail.

kernel/src/syscall/chmod.rs
```diff
@@ -16,10 +16,10 @@ pub fn sys_fchmod(fd: FileDesc, mode: u16, ctx: &Context) -> Result<SyscallRetur
 
     let mut file_table = ctx.thread_local.borrow_file_table_mut();
     let file = get_file_fast!(&mut file_table, fd);
-    file.inode().set_mode(InodeMode::from_bits_truncate(mode))?;
-    if let Some(path) = file.path() {
-        fs::notify::on_attr_change(path);
-    }
+    file.path()
+        .inode()
+        .set_mode(InodeMode::from_bits_truncate(mode))?;
```
lrh2000: If a method already exists in `Path`, I think we we should call methods in `Path` instead of doing `path.inode().some_method()`
https://github.com/asterinas/asterinas/blob/922b204280440203d0e5f9d6a06040df64d1e7e5/kernel/src/fs/path/mod.rs#L491-L499

Please go through the commit `` Remove `PathOrInode` `` and resolve all similar issues.

test/src/apps/pseudofs/pseudo_dentry.c
```diff
@@ -0,0 +1,54 @@
...
+		if (fstat(fd, &st) < 0 || st.st_ino != ino)
+			return -1;
+	}
+
+	return 0;
+}
+
+FN_TEST(pseudo_dentry)
+{
+	TEST_SUCC(readlink_check(pipe_1[0], "pipe:[", 1));
```
lrh2000: In fact, I think you'll have to write `TEST_RES(..., _ret == 0)` because `TEST_SUCC` does not care about the return value. It only checks to see if `errno` has been set. However, I don't believe `readlink_check` ever sets `errno`.

kernel/src/fs/inode_handle.rs
```diff
@@ -364,6 +365,15 @@ impl FileLike for InodeHandle {
     }
 
     fn set_status_flags(&self, new_status_flags: StatusFlags) -> Result<()> {
+        if self
+            .file_io
+            .as_ref()
+            .and_then(|file_io| (file_io.as_ref() as &dyn Any).downcast_ref::<PipeHandle>())
```
lrh2000, line 370: This is a bit ugly. Have you checked whether other inodes have similar restrictions on status flags? Perhaps we could add a TODO comment here?

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -281,240 +239,83 @@ impl MemfdFile {
...
-            return_errno_with_message!(Errno::EBADF, "the file is not opened writable");
-        }
-
-        do_fallocate_util(
-            self.memfd_inode.as_ref(),
-            self.status_flags(),
-            mode,
-            offset,
-            len,
+fn memfd_inode_or_err(file: &InodeHandle) -> Result<&MemfdInode> {
```
lrh2000: This method should be defined before `MemfdTmpFs`?

Also I'm not sure if you found the code better if writing it as:
```rust
trait MemfdInodeHandle {
    fn new_memfd() -> InodeHandle;
    fn add_seals(&self) -> Result<()>;
    // ...
}

impl MemfdInodeHandle for InodeHandle {
    // ...
}
```

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -231,27 +216,16 @@ impl Inode for MemfdInode {
...
-impl MemfdFile {
-    pub fn new(name: &str, memfd_flags: MemfdFlags) -> Result<Self> {
-        if name.len() > MAX_MEMFD_NAME_LEN {
-            return_errno_with_message!(Errno::EINVAL, "MemfdManager: `name` is too long.");
-        }
-
-        let name = format!("/memfd:{}", name);
+pub fn do_memfd_create(name: String, memfd_flags: MemfdFlags) -> Result<InodeHandle> {
+    if name.len() > MAX_MEMFD_NAME_LEN {
+        return_errno_with_message!(Errno::EINVAL, "the name of memfd is too long");
```
lrh2000:
```suggestion
        return_errno_with_message!(Errno::EINVAL, "the memfd name is too long");
```

lrh2000 (commented):
Thank you for your wonderful work! This really makes the code much better in various ways. I only have a few minor points.

kernel/src/fs/inode_handle.rs
vvvvsv: Please see this [comment](https://github.com/asterinas/asterinas/pull/2798#discussion_r2666978190).

I will add TODOs both here and in `check_status_flags`.

kernel/src/fs/path/dentry.rs
vvvvsv: In the last commit, I removed `children` from all non-directory dentries, please take a look. What do you think about this approach?

kernel/src/fs/path/dentry.rs
```diff
@@ -415,7 +479,6 @@ impl Dentry {
             current_dir = parent_dir;
         }
 
-        debug_assert!(path_name.starts_with('/'));
```
lrh2000, line 418: Consider:
```rust
        debug_assert!(self.is_pesudo() || path_name.starts_with('/'));
```
?

kernel/src/fs/path/dentry.rs
```diff
@@ -18,13 +18,56 @@ use crate::{
...
+                None => String::from("/"),
+            },
+            NameAndParent::Pseudo(name_fn) => (name_fn)(inode),
+        }
+    }
+
+    fn parent(&self) -> Option<Arc<Dentry>> {
+        match self {
+            NameAndParent::Real(Some(name_and_parent)) => Some(name_and_parent.read().1.clone()),
+            _ => None,
```
lrh2000: Use `NameAndParent::Pseudo(_)` here. The same applies for `fn set`.

Otherwise, it should be written as a `let ... else` block.

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -45,8 +34,8 @@ use crate::{
 pub const MAX_MEMFD_NAME_LEN: usize = 249;
 
 pub struct MemfdInode {
+    pub(self) name: String,
     inode: RamInode,
-    name: String,
```
lrh2000, line 49: Keep the original order. Otherwise, you can also update the order in the constructor below:
```rust
            MemfdInode {
                inode: ram_inode,
                name,
                seals: Mutex::new(seals),
            }
```

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -45,8 +34,8 @@ use crate::{
 pub const MAX_MEMFD_NAME_LEN: usize = 249;
 
 pub struct MemfdInode {
+    pub(self) name: String,
```
lrh2000: What's wrong with the getter method that exists before this PR? I think it's better than the current approach because the current approach gives me the impression that someone wants to modify the field, which is not true.
```rust
    pub(self) fn name(&self) -> &str {
        &self.name
    }
```

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -235,53 +220,31 @@ impl Inode for MemfdInode {
...
+pub trait MemfdInodeHandle {
+    fn new_memfd(name: String, memfd_flags: MemfdFlags) -> Result<InodeHandle>;
+    fn add_seals(&self, new_seals: FileSeals) -> Result<()>;
+    fn get_seals(&self) -> Result<FileSeals>;
 }
 
-impl MemfdFile {
-    pub fn new(name: &str, memfd_flags: MemfdFlags) -> Result<Self> {
+impl MemfdInodeHandle for InodeHandle {
+    fn new_memfd(name: String, memfd_flags: MemfdFlags) -> Result<InodeHandle> {
```
lrh2000: `Result<Self>`. Also for the trait definition.

kernel/src/fs/ramfs/memfd.rs
```diff
@@ -304,216 +267,70 @@ impl MemfdFile {
...
+        Ok(self.memfd_inode_or_err()?.get_seals())
     }
 }
 
-impl Pollable for MemfdFile {
-    fn poll(&self, mask: IoEvents, _poller: Option<&mut PollHandle>) -> IoEvents {
-        let events = IoEvents::IN | IoEvents::OUT;
-        events & mask
+impl InodeHandle {
+    fn memfd_inode_or_err(&self) -> Result<&MemfdInode> {
```
lrh2000: This is not the correct file in which this `impl` should exist. I think for local helper methods, keeping it as a normal method that does not accept `self` is fine.

kernel/src/fs/inode_handle.rs
```diff
@@ -84,6 +84,10 @@ impl InodeHandle {
         *offset
     }
 
+    pub fn rights(&self) -> Rights {
```
lrh2000: `pub(super)`

kernel/src/fs/inode_handle.rs
lrh2000: Thanks for the explanation!

kernel/src/fs/path/mod.rs
```diff
@@ -483,7 +486,15 @@ impl Path {
             return_errno_with_message!(Errno::EXDEV, "the operation cannot cross mounts");
         }
 
-        self.dentry.link(&old.dentry, name)
+        self.dentry.as_dir_dentry_or_err()?.link(old.inode(), name)
+    }
+
+    pub fn unlink(&self, name: &str) -> Result<()> {
```
lrh2000: Be consistent. Add doc comments like the methods near to it.

kernel/src/fs/path/dentry.rs
```diff
@@ -94,11 +93,14 @@ impl Dentry {
             DentryOptions::Pseudo(name_fn) => NameAndParent::Pseudo(name_fn),
         };
 
+        let children =
+            matches!(inode.type_(), InodeType::Dir).then_some(RwMutex::new(DentryChildren::new()));
```
lrh2000: Use `then(|| RwMutex::new(...))`.

kernel/src/fs/path/dentry.rs
```diff
@@ -203,25 +205,58 @@ impl Dentry {
...
+            return_errno_with_message!(
+                Errno::ENOTDIR,
+                "this dentry is not related to a directory inode"
+            );
+        }
+        Ok(DirDentry(self))
+    }
+}
+
+pub(super) struct DirDentry<'a>(&'a Dentry);
```
lrh2000: Add doc comments to the type.

kernel/src/fs/path/dentry.rs
```diff
@@ -203,25 +205,58 @@ impl Dentry {
...
+                }
+            };
+            current_dir = parent_dir;
+        }
+
+        path_name
+    }
+
+    pub(super) fn as_dir_dentry_or_err(&self) -> Result<DirDentry<'_>> {
+        if self.type_() != InodeType::Dir {
```
lrh2000: How about:
```rust
        let Some(children) = self.children else {
           return_errno_with_message!(...)
        };
        DirDentry {
            inner: self,
            children: children,
        }
```

If using `self.inner` is too annoying, maybe we can try to `impl Deref`.

kernel/src/fs/path/dentry.rs
```diff
@@ -274,23 +310,18 @@ impl Dentry {
...
     }
 
-    /// Links a new name for the `Dentry` by `link()` the inner inode.
-    pub(super) fn link(&self, old: &Arc<Self>, name: &str) -> Result<()> {
-        if self.type_() != InodeType::Dir {
-            return_errno!(Errno::ENOTDIR);
-        }
-
-        let children = self.children.upread();
+    /// Links a new `Dentry` for the old inode via `link()`.
```
lrh2000: Be consistent with other methods' doc comments:
```rust
    /// Links a new `Dentry` by `link()` the old inode.
```

kernel/src/fs/path/dentry.rs
```diff
@@ -403,31 +428,33 @@ impl Dentry {
     }
 
     /// Renames a `Dentry` to the new `Dentry` by `rename()` the inner inode.
-    pub(super) fn rename(&self, old_name: &str, new_dir: &Arc<Self>, new_name: &str) -> Result<()> {
+    pub(super) fn rename(&self, old_name: &str, new_dir: &Self, new_name: &str) -> Result<()> {
```
lrh2000: It's fine to accept an `Arc<Dentry>` because it can avoid some unnecessary atomic operations. Arguably, `self` should also be an `Arc`.

In addition, there are some readability problems caused by `self` and `new_dir`, which makes me to think about this:

```rust
    pub(super) fn rename(old_dir_arc: &Arc<Dentry>, old_name: &str, new_dir_arc: &Arc<Dentry>, new_name: &str) -> Result<()> {
        let old_dir = old_dir.as_dir_dentry_or_err()?;
        let new_dir = new_dir.as_dir_dentry_or_err()?;

        if Arc::ptr_eq(&old_dir_arc, &new_dir_arc) {
```

But then the caller has to write `DirDentry::rename(...)`.

kernel/src/fs/path/dentry.rs
```diff
@@ -203,25 +205,58 @@ impl Dentry {
...
+            );
+        }
+        Ok(DirDentry(self))
+    }
+}
+
+pub(super) struct DirDentry<'a>(&'a Dentry);
+
+impl DirDentry<'_> {
     /// Creates a `Dentry_` by creating a new inode of the `type_` with the `mode`.
```
lrh2000, line 200: `Dentry_` -> `Dentry`

kernel/src/fs/path/dentry.rs
lrh2000: I thought it should be in separate PRs because this PR is already too large. But...

vvvvsv: > separate PRs

Sorry, I was just trying this out. I’ll update the commit based on your suggestion and open a new PR to include this commit once this PR is merged.

kernel/src/fs/path/mount.rs
```diff
@@ -188,9 +190,17 @@ impl Mount {
...
+    pub(in crate::fs) fn new_pseudo(fs: Arc<dyn FileSystem>) -> Arc<Self> {
+        Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new())
+    }
+
     /// The internal constructor.
     ///
-    /// Root mount node has no mountpoint which other mount nodes must have mountpoint.
+    /// Root mount node has no mountpoint, while other mount nodes must have one.
     ///
     /// Here, a Mount is instantiated without an initial mountpoint,
```
lrh2000: Given that you've already fixed a grammar issue here...
```suggestion
    /// A root mount node has no mountpoint, while other mount nodes must have one.
    ///
    /// Here, a `Mount` is instantiated without an initial mountpoint,
```

kernel/src/fs/path/dentry.rs
```diff
@@ -18,13 +18,56 @@ use crate::{
...
+            *name_and_parent = (String::from(name), parent);
+            Ok(())
+        } else {
+            Err(SetNameAndParentError)
+        }
+    }
+}
+
+#[derive(Debug)]
+struct SetNameAndParentError;
```
lrh2000: Move this before the `impl` block.

---

Also I think the following version, **without `Dentry::set_name_and_parent`**, is clearer:
```rust
/// The name and parent of a `Dentry`.
enum NameAndParent {}

/// An error returned by [`NameAndParent::set`].
#[derive(Debug)]
struct SetNameAndParentError;

impl NameAndParent {
    /// Sets the name and parent of the `Dentry`.
    ///
    /// # Errors
    ///
    /// Returns `SetNameAndParentError` if the `Dentry` is a root or pseudo `Dentry`.
    fn set(
        &self,
        name: &str,
        parent: Arc<Dentry>,
    ) -> core::result::Result<(), SetNameAndParentError> {
}
```

The problem with the current version is that: When people see `NameAndParent::set` (which occurs first in the file), they can see no documentations. They have to scroll down to find the `set_name_and_parent` method to read the documentation, which is annoying. Additionally, `self.name_and_parent.set` is equivalent to `self.set_name_and_parent`, so the latter seems unnecessary (unless it needs to be `pub(super)`, but that's not the case).

vvvvsv:
@cqs21 Could you please help to review this PR?

kernel/src/fs/path/dentry.rs
```diff
@@ -34,21 +84,38 @@ impl Dentry {
         Self::new(inode, DentryOptions::Root)
     }
 
+    /// Creates a new pseudo `Dentry` with the given inode and name function.
+    pub(super) fn new_pseudo(
+        inode: Arc<dyn Inode>,
+        name_fn: fn(&dyn Inode) -> String,
```
tatetian: This closure type appears multiple times. Define a type alias so that (1) we don't need to repeat ourselves multiple times and (2) we can attach some proper documentation to the type alias.

PR #2799, opened 2025-12-25
Manage package metadata and dependencies at workspace level

StevenJiang1110:
This PR centralizes package metadata and dependencies at the workspace level. By allowing each crate to reference definitions in the workspace-root `Cargo.toml`, we ensure consistency for metadata and dependency versions across the entire project.

This is primarily a refactoring effort; no functional changes are introduced.

Commit 1: Centralizes package metadata (currently `version`, `edition`, `repository`, and `license`) at the workspace level. A full list of supported keys can be found [here](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table).
Commit 2: Manages dependencies at the workspace level to enforce version consistency project-wide. This helps reduce binary size and improve compilation speed. Note that optional and  dependencies are not yet handled this way due to [certain limitations](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table). Architecture-specific dependencies are not handled yet. All third-party dependencies have been unified except for bitflags, which still requires versions 1.3 and 2.5.
Commit 3: Relocates `int-to-c-enum` from `kernel/libs` to `ostd/libs`. As `ostd` depends on this crate, it should be part of the TCB.
Commit 4: Updates mlsdisk to use `Pod` from `ostd` instead of `ostd-pod`, aligning it with the usage in other crates.
Commit 5: Updates `bump_version.sh` to accommodate these workspace-level changes.
Commit 6: Let OSDK get dependent crate version via `cargo metadata`, instead of directly reading the `Cargo.toml`. The `package.version` in Cargo.toml now becomes `version.workspace = true`.

### DIFF
--- a/osdk/src/base_crate/mod.rs
+++ b/osdk/src/base_crate/mod.rs
@@ -127,24 +127,25 @@ fn do_new_base_crate(
         std::fs::remove_dir_all(&base_crate_path).unwrap();
     }
 
+    let dep_meta = get_cargo_metadata(Some(dep_crate_path.as_ref()), None::<&[&str]>).unwrap();
     let (dep_crate_version, dep_crate_features) = {
-        let cargo_toml = dep_crate_path.as_ref().join("Cargo.toml");
-        let cargo_toml = fs::read_to_string(cargo_toml).unwrap();
-        let cargo_toml: toml::Value = toml::from_str(&cargo_toml).unwrap();
-        let dep_version = cargo_toml
-            .get("package")
-            .unwrap()
-            .as_table()
-            .unwrap()
-            .get("version")
-            .unwrap()
-            .as_str()
-            .unwrap()
-            .to_string();
-        let dep_features = cargo_toml
-            .get("features")
-            .map(|f| f.as_table().unwrap().clone())
-            .unwrap_or_default();
+        let dep_meta = dep_meta.as_object().unwrap();
+
+        let dep_package = {
+            let packages = dep_meta.get("packages").unwrap().as_array().unwrap();
+            packages
+                .iter()
+                .find(|package| {
+                    let package = package.as_object().unwrap();
+                    let name = package.get("name").unwrap().as_str().unwrap();
+                    name == dep_crate_name
+                })
+                .unwrap()
+                .as_object()
+                .unwrap()
+        };
+        let dep_version = dep_package.get("version").unwrap().as_str().unwrap();
+        let dep_features = dep_package.get("features").unwrap().as_object().unwrap();
         (dep_version, dep_features)
     };
 
@@ -156,7 +157,7 @@ fn do_new_base_crate(
     // Write Cargo.toml
     let cargo_toml = include_str!("Cargo.toml.template");
     let cargo_toml = cargo_toml.replace("#NAME#", &(dep_crate_name.to_string() + "-osdk-bin"));
-    let cargo_toml = cargo_toml.replace("#VERSION#", &dep_crate_version);
+    let cargo_toml = cargo_toml.replace("#VERSION#", dep_crate_version);
     fs::write(base_crate_path.as_ref().join("Cargo.toml"), cargo_toml).unwrap();
 
     // Set the current directory to the target osdk directory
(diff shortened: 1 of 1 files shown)


osdk/deps/heap-allocator/Cargo.toml
```diff
@@ -1,15 +1,15 @@
...
-version = "0.17.0"
-edition = "2024"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
 description = "The default slab-based heap allocator shipped with OSDK."
-license = "MPL-2.0"
-repository ="https://github.com/asterinas/asterinas"
 readme = "README.md"
```
tatetian, line 4: Is it desirable to sort the fields under `[package]` in a specific order? For example, crate-provided fields first, then workspace-inherited ones. Or simply in the alphabetical order. The current items do not seem to be organized in any specific order. Since the PR is changing most, if not all, `Cargo.toml` files in the repo, I think we can use it as a good chance to think about it and work on it.

tools/sctrace/Cargo.toml
```diff
@@ -16,6 +16,3 @@ clap = { version = "4.5.47", features = ["derive"] }
 regex = "1.11.2"
 nom = "7"
 nix = { version = "0.27", features = ["process"] }
-
-[lints]
-workspace = true
```
tatetian: Why removing the `[lints]` section?

kernel/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "aster-nix"
```
tatetian: Could you rename this crate to `aster-kernel` as a new commit in this PR? Now that we have the NixOS-based distribution and use Nix in our build system. Having `nix` in the `aster-nix` crate name would cause confusion.

Cargo.toml
```diff
@@ -65,6 +65,117 @@ edition = "2024"
...
+minicov = "0.3"
+num-traits = { version = "0.2", default-features = false }
+ostd-pod = { git = "https://github.com/asterinas/ostd-pod", rev = "c4644be", version = "0.1.1" }
+serde = { version = "1.0.192", default-features = false, features = ["alloc", "derive"] }
+smallvec = "1.13.2"
+uart_16550 = "0.3.0"
+volatile = "0.6.1"
+
+# External dependencies only for safe crates (i.e., crates under kernel or osdk/deps directories)
+# Enable `force-soft` feature to disable `AES-NI` and `CLMUL` intrinsics, ensuring that the implementation
```
tatetian, line 145: Insert a blank line before this line

```toml
# External dependencies only for safe crates (i.e., crates under kernel or osdk/deps directories)
#
# Enable `force-soft` feature to disable `AES-NI` and `CLMUL` intrinsics, ensuring that the implementation
```

Cargo.toml
```diff
@@ -8,6 +8,8 @@ members = [
     "ostd/libs/align_ext",
     "ostd/libs/ostd-macros",
     "ostd/libs/id-alloc",
```
tatetian, line 17: I think it would be very helpful to add some comments at the top of this file to explain the set of crates (or the criterion for crates) to be governed by the Cargo workspace. I write a draft. Feel free to revise it as you see fit.

```toml
# The Cargo workspace configurations shared by all in-kernel crates.
#
# As a kernel project, most crates in Asterinas are written and built for the kernel space.
# But there are also a few userspace crates
# whose crate-level `Cargo.toml` files are not affected by this workspace-level `Cargo.toml`.
# The complete list can be found in the `exclude` field.
```

This should help developers to decide whether a new crate should be a workspace member or not in the future.

osdk/deps/heap-allocator/Cargo.toml
StevenJiang1110: > Is it desirable to sort the fields under [package] in a specific order?

Yes, I think these fields should be organized in a consistent order. However, the community has yet to reach a consensus on what that order should be.

The current [official guidelines](https://doc.rust-lang.org/nightly/style-guide/cargo.html#formatting-conventions) specify that `name` and `version` should be at the top and `description` at the bottom, but they offer no guidance for the keys in between. These guidelines are somewhat outdated and do not account for modern features like workspace inheritance. Furthermore, many prominent projects do not strictly follow them; for instance, [clap](https://github.com/clap-rs/clap/blob/master/Cargo.toml#L113) places the `description` immediately after the `version` rather than at the end.

According to recent [discussions on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/.60Cargo.2Etoml.60.20style.20guide/with/564500780), there is an ongoing effort to develop a tool called `cargo-cargofmt` to automate `Cargo.toml` formatting. A few key takeaways from those discussions include:
1. Inherited keys (`.workspace = true`) should generally follow keys defined locally within the crate.
2. Keys should be sorted by priority and relevance (e.g., `name` and `version` should be at the top; grouping `edition` and `rust-version`) rather than alphabetically.

For this PR, I have organized the keys based on the [preliminary order proposed in cargo-cargofmt](https://github.com/crate-ci/cargo-cargofmt/issues/29#issuecomment-3647516745). While this order is not yet finalized, it provides a basic standard. Once the tool matures, we can adopt it to automatically enforce the key order and format all `Cargo.toml` files across the project.

PR #2803, opened 2025-12-26
Checks documentation in all crates

StevenJiang1110:
This PR introduces documentation checks for all crates. Previously, checks were only performed on the specific crates intended for publishing.

The kernel crate is a special case because most of its items are intentionally private. Since Rustdoc ignores private items by default, I have added the `--document-private-items` flag for this crate to ensure its internal documentation is also verified.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/lib.rs
+++ b/kernel/libs/aster-bigtcp/src/lib.rs
@@ -12,6 +12,10 @@
 
 #![no_std]
 #![deny(unsafe_code)]
+// Rustdoc does not render methods of public type aliases that reference private types,
+// so links pointing to them cannot be resolved.
+// See <https://github.com/rust-lang/rust/issues/132952>.
+#![expect(rustdoc::private_intra_doc_links)]
 
 pub mod boolean_value;
 pub mod device;
--- a/kernel/src/fs/inode_handle/mod.rs
+++ b/kernel/src/fs/inode_handle/mod.rs
@@ -309,9 +309,9 @@ pub trait FileIo: Pollable + InodeIo + Send + Sync + 'static {
 
     /// Returns whether the `read()`/`write()` operation should use and advance the offset.
     ///
-    /// If [`is_seekable`] succeeds but this method returns `false`, the offset in the `seek()`
-    /// operation will be ignored. In that case, the `seek()` operation will do nothing but
-    /// succeed.
+    /// If [`FileIo::check_seekable`] succeeds but this method returns `false`,
+    /// the offset in the `seek()` operation will be ignored.
+    /// In that case, the `seek()` operation will do nothing but succeed.
     fn is_offset_aware(&self) -> bool;
 
     // See `FileLike::mappable`.
--- a/kernel/comps/console/src/font.rs
+++ b/kernel/comps/console/src/font.rs
@@ -8,10 +8,7 @@ use font8x8::UnicodeFonts;
 
 /// A bitmap font.
 ///
-/// Currently it's mainly used to draw texts on the framebuffer console. See
-/// [`FramebufferConsole::set_font`].
-///
-/// [`FramebufferConsole::set_font`]: crate::FramebufferConsole::set_font
+/// Currently it's mainly used to draw texts on the framebuffer console.
 #[derive(Debug)]
 pub struct BitmapFont {
     width: usize,
--- a/kernel/comps/input/src/input_handler.rs
+++ b/kernel/comps/input/src/input_handler.rs
@@ -27,7 +27,6 @@ pub enum ConnectError {
 ///
 /// [`register_handler_class`]: crate::register_handler_class
 /// [`connect`]: Self::connect
-/// [`InputHandler`]: crate::InputHandler
 pub trait InputHandlerClass: Send + Sync + Any + Debug {
     /// Returns the class name of the handler class.
     fn name(&self) -> &str;
--- a/kernel/comps/virtio/src/device/input/mod.rs
+++ b/kernel/comps/virtio/src/device/input/mod.rs
@@ -34,7 +34,7 @@ use crate::transport::VirtioTransport;
 
 pub const DEVICE_NAME: &str = "Virtio-Input";
 
-/// Select value used for [`device::InputDevice::query_config_select()`].
+/// Select value used for `device::InputDevice::select_config`.
 #[repr(u8)]
 #[derive(Debug, Clone, Copy)]
 pub enum InputConfigSelect {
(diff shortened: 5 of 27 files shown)


kernel/src/fs/inode_handle/mod.rs
```diff
@@ -309,7 +309,7 @@ pub trait FileIo: Pollable + InodeIo + Send + Sync + 'static {
 
     /// Returns whether the `read()`/`write()` operation should use and advance the offset.
     ///
-    /// If [`is_seekable`] succeeds but this method returns `false`, the offset in the `seek()`
+    /// If [`FileIo::check_seekable`] succeeds but this method returns `false`, the offset in the `seek()`
     /// operation will be ignored. In that case, the `seek()` operation will do nothing but
     /// succeed.
```
lrh2000:
```suggestion
    /// If [`FileIo::check_seekable`] succeeds but this method returns `false`, the offset in the
    /// `seek()` operation will be ignored. In that case, the `seek()` operation will do nothing
    /// but succeed.
```

kernel/src/lib.rs
```diff
@@ -19,6 +19,7 @@
 #![feature(associated_type_defaults)]
 #![feature(try_with_capacity)]
 #![register_tool(component_access_control)]
+#![allow(rustdoc::private_intra_doc_links)]
```
lrh2000: I was wondering if we could enable this in the command line together with the `--document-private-items` flag. It seems that `rustdoc` has a `--allow` option.
```
> rustdoc --help
[ .. ]
    -A, --allow LINT    Set lint allowed
```

lrh2000 (commented):
It's exciting to see this checked in CI!

kernel/src/lib.rs
StevenJiang1110: For the kernel crate, I have moved -A rustdoc::private_intra_doc_links to RUSTDOCFLAGS.

However, for aster-bigtcp, I am retaining `#![allow(rustdoc::private_intra_doc_links)]` at the crate level. Otherwise, the following error occurs:

```plain
error: public documentation for `is_rst_closed` links to private item `TcpConnection::recv`
   --> kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs:105:76
    |
105 |     /// [`TcpConnection::clear_rst_closed`], [`TcpConnection::send`], or [`TcpConnection::recv`].
    |                                                                            ^^^^^^^^^^^^^^^^^^^ this item is private
    |
    = note: this link will resolve properly if you pass `--document-private-items`
```

Interestingly, both `TcpConnection` and `TcpConnection::clear_rst_closed` are public items. the LLM tolds me this is a known limitation in Rustdoc: when methods are defined on a type alias, Rustdoc often fails to resolve the visibility correctly. I haven't found a cleaner workaround for this yet.

lrh2000: This is something that needs to be fixed, either on the Rust side or on our side. Of course, it won't be urgent to fix it. However, let's clarify this:
```rust
// Rustdoc does not render methods of public type aliases that reference private types, so links
// pointing to them cannot be resolved. See <https://github.com/rust-lang/rust/issues/132952>.
#![expect(rustdoc::private_intra_doc_links)]
```

Makefile
```diff
@@ -403,10 +403,15 @@ ktest: initramfs $(CARGO_OSDK)
...
-		(cd $$dir && cargo doc --no-deps) || exit 1; \
+		(cd $$dir && RUSTDOCFLAGS="-Dwarnings" cargo doc --no-deps) || exit 1; \
 	done
 	@for dir in $(OSDK_CRATES); do \
-		(cd $$dir && cargo osdk doc --no-deps) || exit 1; \
+		EXTRA_DOC_FLAGS=""; \
+		# The kernel crate is primarily composed of private items. \
+		# We include the --document-private-items flag \
+		# to ensure documentation of the internal items is fully checked. \
+		if [ "$$dir" = "kernel" ]; then EXTRA_DOC_FLAGS="--document-private-items -Arustdoc::private_intra_doc_links"; fi; \
```
lrh2000: This line is too long. In fact, it's the longest line in this file. Break it up?

Makefile
```diff
@@ -403,10 +403,17 @@ ktest: initramfs $(CARGO_OSDK)
...
+		(cd $$dir && RUSTDOCFLAGS="-Dwarnings" cargo doc --no-deps) || exit 1; \
 	done
 	@for dir in $(OSDK_CRATES); do \
-		(cd $$dir && cargo osdk doc --no-deps) || exit 1; \
+		EXTRA_DOC_FLAGS=""; \
+		# The kernel crate is primarily composed of private items. \
+		# We include the --document-private-items flag \
+		# to ensure documentation of the internal items is fully checked. \
+		if [ "$$dir" = "kernel" ]; \
+			then EXTRA_DOC_FLAGS="--document-private-items -Arustdoc::private_intra_doc_links"; \
```
lrh2000:
```suggestion
		if [ "$$dir" = "kernel" ]; then \
			EXTRA_DOC_FLAGS="--document-private-items -Arustdoc::private_intra_doc_links"; \
```

This does not seem to be the correct style... You can find some examples in the same file:
https://github.com/asterinas/asterinas/blob/ebfa7977b19107967703f6f97e11fe3c73cecdc9/Makefile#L285-L289

kernel/libs/aster-bigtcp/src/lib.rs
```diff
@@ -12,6 +12,10 @@
 
 #![no_std]
 #![deny(unsafe_code)]
+// Rustdoc does not render methods of public type aliases that reference private types,
```
lrh2000:
```suggestion
// Rustdoc does not render methods of public type aliases that reference sealed types,
```

Sorry, my previous wording was not precise. "Sealed" is a more accurate term than "private."

lrh2000 (approved):

lrh2000:
It seems that the doc tests fail due to some concurrently merged PRs...
```
error: unresolved link to `ostd::mm::DmaStream`
  --> kernel/libs/aster-util/src/mem_obj_slice.rs:11:20
   |
11 | //! [`DmaStream`]: ostd::mm::DmaStream
   |                    ^^^^^^^^^^^^^^^^^^^ no item named `DmaStream` in module `mm`
   |
   = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`
```

StevenJiang1110:
> It seems that the doc tests fail due to some concurrently merged PRs...

I will open an PR to solve the problem soon.

PR #2804, opened 2025-12-26
Fix invalid environment name escaping in Makefile

jakobhellermann:
Previously this would print `Error: this \ environment variable must be given`.

If you do want to include the `$`, you'd need `\$$` to escape through both Makefile and bash.


tatetian (approved):

PR #2810, opened 2025-12-28
Fix the broken link to "Contriubting to Asterinas" in Book

tatetian:


tatetian:
Merge a trivial change to the Book without review

PR #2811, opened 2025-12-28
Update mdBook in the Docker image to 0.5.2

tatetian:
The latest mdBook has some new good features that are present in our currently-installed mdBook. This PR triggers the generation of a new Docker image with the latest mdBook.


tatetian:
Merged a trivial change without other's reviews

PR #2816, opened 2025-12-29
Bump docker ref to 0.17.0-20251228

StevenJiang1110:
Bumps the docker ref to the latest.

It looks like `books.toml` fails to build now. See [this run](https://github.com/asterinas/asterinas/actions/runs/20564129418/job/59059277312). Fix this problem in the second commit.


tatetian (approved):

PR #2818, opened 2025-12-29
Fix mmap address allocation to stay below stack region

cqs21:
This PR could resolve the sqlite benchmark failure (https://github.com/asterinas/asterinas/actions/runs/20541891630/job/59007314580) 
 
The root cause was that `mmap` could allocate addresses higher than the stack region, causing user programs to potentially access addresses beyond `MAX_USERSPACE_VADDR`, which resulted in a general protection fault.

### DIFF
--- a/kernel/src/vm/vmar/vmar_impls/mod.rs
+++ b/kernel/src/vm/vmar/vmar_impls/mod.rs
@@ -309,15 +309,15 @@ impl VmarInner {
     /// If no such region is found, return an error.
     fn alloc_free_region(&mut self, size: usize, align: usize) -> Result<Range<Vaddr>> {
         // Fast path that there's still room to the end.
-        let highest_occupied = self
-            .vm_mappings
-            .iter()
+        let mut vm_mappings_iter = self.vm_mappings.iter();
+        let stack_mapping = vm_mappings_iter.next_back().unwrap();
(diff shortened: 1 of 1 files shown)


lrh2000:
> access addresses beyond `MAX_USERSPACE_VADDR`

How can this happen? Given:
https://github.com/asterinas/asterinas/blob/ebfa7977b19107967703f6f97e11fe3c73cecdc9/kernel/src/vm/vmar/vmar_impls/mod.rs#L319-L320

---

In addition, this method (i.e., `alloc_free_region`) will be completely rewritten in https://github.com/asterinas/asterinas/pull/2736...

cqs21:
> How can this happen? Given:
> 
> https://github.com/asterinas/asterinas/blob/ebfa7977b19107967703f6f97e11fe3c73cecdc9/kernel/src/vm/vmar/vmar_impls/mod.rs#L319-L320
> 

You're right, this `mmap` won't exceed the range. However, the issue might occur later with `mremap` operations on this mapping. This is a quick fix, and I haven't figured out a more proper solution yet.

cqs21:
> In addition, this method (i.e., `alloc_free_region`) will be completely rewritten in #2736...

It looks like that #2736 could resolve sqlite problem, so I'll close this PR.

PR #2819, opened 2025-12-29
Add generic syscall table

### LINKED ISSUE #2294: Support the generic syscall table for Asterinas
Recently, Asterinas has added basic support for the LoongArch architecture. If we compare the contents of the syscall tables for [RISC-V](https://github.com/asterinas/asterinas/blob/main/kernel/src/syscall/arch/riscv.rs) and [LoongArch](https://github.com/asterinas/asterinas/blob/main/kernel/src/syscall/arch/loongarch.rs), we can see that they are almost identical.

This similarity is no coincidence. In Linux, there is a [generic syscall table](https://elixir.bootlin.com/linux/v6.0.9/source/include/uapi/asm-generic/unistd.h) that can be used by most architectures. While x86-64 does not use this generic table, RISC-V, LoongArch, and many other archtectures do. In addition to the syscalls defined in the generic table, each architecture may also implement a few architecture-specific syscalls. For example, RISC-V has a special syscall called [riscv_flush_icache](https://elixir.bootlin.com/linux/v6.0.9/source/arch/riscv/include/uapi/asm/unistd.h#L45).

I think it is now time to consider introducing a generic syscall table for Asterinas. This approach would help reduce code duplication significantly. Furthermore, if we plan to support additional architectures in the future, such as Arm64, the generic syscall table could be easily reused.

The generic syscall table API should allow architectures to reuse all common syscalls while also enabling them to define their own architecture-specific syscalls.

StevenJiang1110:
Fixes #2294.

This PR introduces the generic syscall table, and thus the syscall table of LoongArch and RISC-V are replaced with the generic syscall table. 

# Handling Architecture-Generic Code

This is a follow-up to the discussion in https://github.com/asterinas/asterinas/issues/2294#issuecomment-3142020111.

We can organize shared code within a generic directory and have arch-specific modules explicitly include them using the `#[path]` attribute:

```plain
arch/
    generic/
        io.rs
        pci.rs
    loongarch/
        mod.rs
    riscv/
        mod.rs
```
in `loongarch/mod.rs` and `riscv/mod.rs`

```rust
// In `loongarch/mod.rs` and `riscv/mod.rs`
#[path = "../generic/io.rs"]
mod io;
#[path = "../generic/pci.rs"]
mod pci;
```

> There can be multiple generic implementations for one module. Every architecture can choose one of them to use. How should this be dealt with?

To handle cases where different architectures require different versions of a "generic" module (e.g., different IO implementations), we can provide multiple files within the generic folder and let each architecture opt-in to the desired version:

```
arch/
    generic/
        io/
          impl1.rs
          impl2.rs
        pci.rs
    loongarch/
        mod.rs
    riscv/
        mod.rs
```

```rust
// In `riscv/mod.rs`
#[path = "../generic/io/impl1.rs"]
mod io;
```

```rust
// In `loongarch/mod.rs`
#[path = "../generic/io/impl2.rs"]
mod io;
```

> What about mostly reusing the generic implementation while still keeping a small part of architecture-specific mechanisms?

For implementations that are mostly identical but require minor architecture-specific adjustments, we can utilize conditional compilation directly within the generic source files:

```rust
// generic/io.rs
#[cfg(target_arch = "riscv")]
fn example(riscv_params: RiscvParams);

#[cfg(target_arch = "loongarch")]
fn example(loongarch_params: LoongarchParams);
```


lrh2000:
> We can organize shared code within a generic directory and have arch-specific modules explicitly include them using the `#[path]` attribute:

It appears that the Rust standard library uses this same approach to organize code that is common but arch-specific.
https://github.com/search?q=repo%3Arust-lang%2Frust+%22%23%5Bpath%22+path%3A%2Flibrary%2F&type=code

kernel/src/syscall/arch/generic.rs
```diff
@@ -0,0 +1,404 @@
...
+/// # Example
+///
+/// Defining architecture-specific syscalls:
+///
+/// ```rust
+/// define_syscalls_with_generic_syscall_table! {
+///     SYS_EXAMPLE1 = 500 => sys_example1(args[..0]);
+///     SYS_EXAMPLE2 = 501 => sys_example2(args[..1], &user_ctx);
+/// }
+/// ```
```
lrh2000:
```suggestion
/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/unistd.h>.
///
/// # Examples
///
/// Define architecture-specific syscalls `example1` and `example2` with generic syscalls:
///
/// ```rust
/// define_syscalls_with_generic_syscall_table! {
///     SYS_EXAMPLE1 = 500 => sys_example1(args[..0]);
///     SYS_EXAMPLE2 = 501 => sys_example2(args[..1], &user_ctx);
/// }
/// ```
```

tatetian (approved):
Very impressive. This solution is quite elegant!

tatetian:
@lrh2000 I leave this PR for you to merge in case that you have any reservations.

lrh2000 (approved):

PR #2821, opened 2025-12-30
Fix `offset: *mut Self` by `offset: usize` in SingleInstruction traits

### LINKED ISSUE #2813: Weird for SingleInstruction traits to take `offset: *mut Self`
`offset: *mut Self` looks weird, because offset should be an integer, but all functions in SingleInstruction traits take a pointer as an offset. For example, `bitor_assign` is defined as follows

https://github.com/asterinas/asterinas/blob/dfe19e4f3cb1a4886f420d55605e7ee10bb405a4/ostd/src/cpu/local/single_instr.rs#L79-L85

And the `offset` is cast back and forth from `usize` to `*mut Self` and back to `usize`:

https://github.com/asterinas/asterinas/blob/dfe19e4f3cb1a4886f420d55605e7ee10bb405a4/ostd/src/cpu/local/cell.rs#L205-L210

https://github.com/asterinas/asterinas/blob/dfe19e4f3cb1a4886f420d55605e7ee10bb405a4/ostd/src/cpu/local/single_instr.rs#L88-L92

BTW I noticed the difference from [miri_asterinas](https://github.com/asterinas/atc25-artifact-evaluation/blob/59dca48f8b2d9d3e5edd8ef89443417b57749682/ostd/src/cpu/local/single_instr.rs#L41), and didn't find any discussion around it in https://github.com/asterinas/asterinas/pull/1125 which initially brought them in.

zjp-CN:
close https://github.com/asterinas/asterinas/issues/2813

### DIFF
--- a/ostd/src/arch/x86/cpu/local.rs
+++ b/ostd/src/arch/x86/cpu/local.rs
@@ -19,7 +19,7 @@ macro_rules! impl_numeric_single_instruction_for {
     ($([$typ: ty, $inout_type: ident, $register_format: expr])*) => {$(
 
         impl SingleInstructionAddAssign<$typ> for $typ {
-            unsafe fn add_assign(offset: *mut Self, val: Self) {
+            unsafe fn add_assign(offset: usize, val: Self) {
                 // SAFETY:
                 // 1. `gs` points to the CPU-local region (global invariant).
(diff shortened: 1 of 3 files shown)


junyang-zh (approved):
LGTM! Thanks for contribution.

PR #2824, opened 2025-12-30
Refactor the implementation of prlimit64

StevenJiang1110:
This PR refactors the implementation of setrlimit, getrlimit, and prlimit64.

The primary goal of this refactoring is to reuse common logic across these three syscalls. Additionally, this PR introduces a missing permission check: if the new `rlim_max` exceeds the current maximum, the calling thread must have CAP_SYS_RESOURCE in the initial user namespace ([referencing](https://elixir.bootlin.com/linux/v6.18/source/kernel/sys.c#L1526)).

With this change, the setrlimit02 LTP test now passes. (setrlimit03 was already passing prior to this PR).

### DIFF
--- a/kernel/src/process/rlimit.rs
+++ b/kernel/src/process/rlimit.rs
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: MPL-2.0
 // FIXME: The resource limits should be respected by the corresponding subsystems of the kernel.
 
-#![expect(non_camel_case_types)]
-
 use core::{
     array,
     sync::atomic::{AtomicU64, Ordering},
 };
 
 use super::process_vm::{INIT_STACK_SIZE, USER_HEAP_SIZE_LIMIT};
-use crate::prelude::*;
+use crate::{
+    prelude::*,
+    process::{UserNamespace, credentials::capabilities::CapSet},
+};
 
 // Constants for the boot-time rlimit defaults
 // See https://github.com/torvalds/linux/blob/fac04efc5c793dccbd07e2d59af9f90b7fc0dca4/include/asm-generic/resource.h#L11
@@ -34,7 +35,7 @@ pub struct ResourceLimits {
 }
 
 impl ResourceLimits {
-    // Get a reference to a specific resource limit
+    /// Returns a reference to a specific resource limit.
     pub fn get_rlimit(&self, resource: ResourceType) -> &RLimit64 {
         &self.rlimits[resource as usize]
     }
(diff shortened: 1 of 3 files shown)


Marsman1996:
Can this fix the #2749 ?

lrh2000:
> Can this fix the #2749 ?

I think it's unrelated. Please see https://github.com/asterinas/asterinas/issues/2749#issuecomment-3699594815.

kernel/src/syscall/prlimit64.rs
```diff
@@ -51,59 +33,75 @@ pub fn sys_prlimit64(
...
+        resource_limits.get_rlimit(resource)
+    };
+
+    let old_raw = {
+        let (cur, max) = rlimit.get_cur_and_max();
+        RawRLimit64 { cur, max }
+    };
+
+    if let Some(new_raw) = new_raw {
+        debug!("new_rlimit = {:?}", new_raw);
```
lrh2000: This seems to be a duplicate of the previous `debug!()`?

kernel/src/syscall/prlimit64.rs
```diff
@@ -51,59 +33,75 @@ pub fn sys_prlimit64(
...
+        if resource == ResourceType::RLIMIT_NOFILE && new_raw.max > SYSCTL_NR_OPEN {
+            return_errno_with_message!(Errno::EPERM, "the new limit exceeds the system limit");
+        }
+        if new_raw.max > old_raw.max {
+            let init_user_ns = UserNamespace::get_init_singleton();
+            init_user_ns.check_cap(CapSet::SYS_RESOURCE, ctx.posix_thread)?;
+        }
+        rlimit.set_cur_and_max(new_raw.cur, new_raw.max)?;
```
lrh2000: `get_cur_and_max` & `set_cur_and_max` should be an atomic operation? I guess it's an existing problem, though.
https://elixir.bootlin.com/linux/v6.18.2/source/kernel/sys.c#L1519-L1538

kernel/src/syscall/prlimit64.rs
```diff
@@ -51,59 +33,71 @@ pub fn sys_prlimit64(
...
+    };
+
+    let old_raw = if let Some(new_raw) = new_raw {
+        if resource == ResourceType::RLIMIT_NOFILE && new_raw.max > SYSCTL_NR_OPEN {
+            return_errno_with_message!(Errno::EPERM, "the new limit exceeds the system limit");
+        }
+        if new_raw.max > rlimit.get_max() {
+            let init_user_ns = UserNamespace::get_init_singleton();
+            init_user_ns.check_cap(CapSet::SYS_RESOURCE, ctx.posix_thread)?;
+        }
```
lrh2000: This is still not an atomic operation. The old `rlimit` may be outdated after this check. We need to perform `new_raw.max > rlimit.get_max()` with the rlimit lock held.

kernel/src/process/rlimit.rs
lrh2000, line 1: Additionally cleanup suggestions in this file:
 - Move `#![expect(non_camel_case_types)]` to the concrete type.
 - Remove `#[repr(C)]` from `RLimit64`.
 - Doc style:
```patch
-    // Get a reference to a specific resource limit
+    // Returns a reference to a specific resource limit.
     pub fn get_rlimit(&self, resource: ResourceType) -> &RLimit64 {
```
 - Variable name style:
```patch
-    pub fn new(cur_: u64, max_: u64) -> Result<Self> {
+    pub fn new(cur: u64, max: u64) -> Result<Self> {
```
 - Error message style:
```patch
-            return_errno_with_message!(Errno::EINVAL, "invalid rlimit");
+            return_errno_with_message!(Errno::EINVAL, "the current rlimit exceeds the max rlimit");
```

lrh2000, line 1: Also, `RLIMIT_COUNT` is not used elsewhere. It is better to remove `pub`.

kernel/src/process/rlimit.rs
```diff
@@ -112,55 +108,70 @@ pub struct RawRLimit64 {
...
+        };
+        self.set_raw_rlimit_unchecked(new);
+        Ok(old)
+    }
+
+    /// Sets the rlimit _without_ synchronization and permission check.
+    ///
+    /// Only called during init process creation.
+    pub(self) fn set_raw_rlimit_unchecked(&self, new: RawRLimit64) {
+        assert!(new.cur < new.max);
```
lrh2000: `cur < max`? Will it work?
```suggestion
        assert!(cur <= max, "the current rlimit exceeds the max rlimit");
```

lrh2000: Also the assertion should be reported at the caller. So `set_raw_rlimit_unchecked`/`RLimit64::new` should be marked as `#[track_caller]`?

kernel/src/process/rlimit.rs
```diff
@@ -112,55 +108,75 @@ pub struct RawRLimit64 {
...
 
-    /// Gets the max rlimit without synchronization.
-    pub fn get_max(&self) -> u64 {
+    /// Returns the max rlimit without synchronization.
+    fn get_max(&self) -> u64 {
         self.max.load(Ordering::Relaxed)
     }
 
-    /// Gets the rlimit with synchronization.
+    /// Returns the rlimit with synchronization.
```
lrh2000: Use "Returns" for cheap operations (just like in most cases we name getter methods as `foo()` instead of `get_foo()`).

Here, `get_raw_limit` takes a lock, so I think it should be "Gets".

PR #2827, opened 2025-12-31
Run clippy test with and without ktest configuration

### LINKED ISSUE #2808: There is dead code when ktest cfg is not enabled by default
Image

You can find clippy  (rustc) diagnostics in os-checker UI: https://os-checker.github.io/file-tree?user=asterinas&repo=asterinas&target=x86_64-unknown-none&checker=clippy&lock=true

StevenJiang1110:
Fixes #2808.

Currently, our `make check` command invokes `cargo osdk clippy`, which by default enables the `ktest` configuration. However, this causes an issue: any code used only within the ktest scope is flagged as dead_code during standard builds (e.g., `make kernel`). This also results in numerous warning messages in our CI, as seen in [this action log](https://github.com/asterinas/asterinas/actions/runs/20615127617/job/59206321917#step:5:393).

To address this, this PR modifies the behavior of OSDK:

- `cargo osdk clippy` and `cargo osdk check` no longer automatically enable `cfg(ktest)`.
- `make check` has been updated to verify the codebase twice: once with `cfg(ktest)` enabled and once with it disabled.

Several pre-existing errors/warnings caused by this discrepancy have also been fixed.

### Alignment with `cargo clippy`

This change better aligns OSDK with standard cargo behavior. For example, consider the following code where a function is only used within a test module:

```rust
//! src/lib.rs

fn add(left: u64, right: u64) -> u64 {
    left + right
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn it_works() {
        let result = add(2, 2);
        assert_eq!(result, 4);
    }
}
```

- Running `cargo clippy` triggers a dead_code warning because add is unused in a non-test build.

```plain
warning: function `add` is never used
 --> src/lib.rs:1:4
  |
1 | fn add(left: u64, right: u64) -> u64 {
  |    ^^^
  |
  = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: `cfg_test` (lib) generated 1 warning
```
- Running `cargo clippy --tests` (enabling `cfg(test)`) clears the warning as the function is now used.

### OSDK interface

To align with `cargo clippy --tests`, this PR introduce a dedicated `--ktests` flag for OSDK (clippy and check commands, e.g., `cargo osdk clippy --ktests`).

### DIFF
--- a/kernel/src/fs/configfs/mod.rs
+++ b/kernel/src/fs/configfs/mod.rs
@@ -28,6 +28,7 @@ pub(super) fn init() {
 /// this function returns an error.
 ///
 /// [`ConfigFs`]: fs::ConfigFs
+#[cfg_attr(not(ktest), expect(dead_code))]
 pub fn register_subsystem(subsystem: Arc<dyn SysBranchNode>) -> Result<()> {
     ConfigRootNode::singleton().add_child(subsystem)?;
 
--- a/Makefile
+++ b/Makefile
@@ -453,14 +453,16 @@ check: initramfs $(CARGO_OSDK)
 	@# Check compilation of the Rust code
 	@for dir in $(NON_OSDK_CRATES); do \
 		echo "Checking $$dir"; \
-		(cd $$dir && cargo clippy -- -D warnings) || exit 1; \
+		(cd $$dir && cargo clippy --no-deps -- -D warnings) || exit 1; \
 	done
 	@for dir in $(OSDK_CRATES); do \
 		echo "Checking $$dir"; \
 		# Exclude linux-bzimage-setup since it only supports x86-64 currently and will panic \
 		# in other architectures. \
 		[ "$$dir" = "ostd/libs/linux-bzimage/setup" ] && [ "$(OSDK_TARGET_ARCH)" != "x86_64" ] && continue; \
-		(cd $$dir && cargo osdk clippy -- -- -D warnings) || exit 1; \
+		# Run clippy on each crate with and without the ktest configuration. \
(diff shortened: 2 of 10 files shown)


kernel/src/fs/configfs/mod.rs
```diff
@@ -26,6 +26,7 @@ pub(super) fn init() {
 ///
 /// If a subsystem with the same name has already been registered,
 /// this function returns an error.
+#[cfg_attr(not(ktest), expect(dead_code))]
```
zjp-CN, line 31: Why not just `#[cfg(ktest)]` ?

StevenJiang1110: There are subtle differences between these two approaches. Consider the following case: if we use `#[cfg(ktest)]` on the function `f1`, the `use core::iter::zip;` statement at the top level will trigger an unused_imports warning whenever ktest is disabled, as its only caller is removed from the AST.

In contrast, using `#[cfg_attr(not(ktest), expect(dead_code))]` is a less intrusive approach. It keeps the function in the AST—thereby satisfying the use statement—while suppressing the dead_code warning when the function is not in use. While `#[cfg(ktest)]` is still appropriate in some scenarios, the current approach helps minimize ripple-effect changes across the codebase.

```rust
use core::iter::zip;

#[cfg(ktest)]
fn f1() {
    zip(iter1, iter2);
}
```

zjp-CN: Thanks for the explanation!

StevenJiang1110:
> @StevenJiang1110 Could you fix the conflict?

The conflict is resolved now.

PR #2829, opened 2025-12-31
Upgrade syscall test suite versions

### LINKED ISSUE #2115: gVisor's version might be out of date
### Current Feature

We use gVisor to test new system calls. However, the version is **20200921.0** specified in `test/syscall_test/Makefile`, which makes it absent for some syscalls added in Linux later. When we need to add these new syscalls, we could not take advantage of current gVisor.

### Enhancement Description

Just update the version specified in `test/syscall_test/Makefile`.

### Motivation

To provide tests for new syscalls.

cqs21:
This PR upgrades the syscall test suites to the latest version and updates corresponding blocklists.
- Upgrade gvisor syscall test version to release-20251215.0
- Upgrade LTP test version to 20250930

Fixes #2115


tatetian (approved):

lrh2000:
I'm not sure if I missed something here, but I think you should only update the block lists after the Docker image has been generated, i.e., in the PR that bumps our Docker version reference.

Otherwise, the current main branch will become odd because no one will be able to update the block list until the Docker image is available. For example, the conflicts in the #2859 are unresolvable because I don't know how to update `inotify_test` without building the image with the latest gvisor myself...

cqs21:
> I think you should only update the block lists after the Docker image has been generated

You're right, the gvisor blocklists should be updated after switching to the new Docker image. However, updating the blocklist early won't cause issues since non-existent test cases in the old gvisor tests will simply be skipped. For LTP, the blocklists need to be updated together because `ltp.nix` is updated, and `nix-build` is triggered every time.
Sorry for the trouble. For now, maybe you can update the blocklist based on the old gvisor, and after updating the Docker image, rerun the CI to see if it works with the new one.

lrh2000:
> For LTP, the blocklists need to be updated together because `ltp.nix` is updated, and `nix-build` is triggered every time.

OK. Thanks for explaining the complexity...

PR #2830, opened 2025-12-31
Fix ftruncate hang on large sparse files

### LINKED ISSUE #2817: System hangs when shrinking a file using `ftruncate` (inconsist with Linux)
### Describe the bug
The issue is located in `kernel/src/vm/vmo/mod.rs`, specifically in the `decommit_pages` method of the `Vmo` struct.

When `ftruncate` shrinks a file, it eventually calls `Vmo::resize`, which calls `decommit_pages` to remove pages in the truncated range.

The `decommit_pages` function iterates over every page index in the range to be decommitted.

However, the Linux won't hang.

### To Reproduce
Compile and run the program:
```C
#define _GNU_SOURCE
#include <sys/mman.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>
#include <fcntl.h>
#include <errno.h>

int main() {
    int fd = memfd_create("test_memfd", 0);
    if (fd == -1) {
        perror("memfd_create");
        return 1;
    }

    // Large size: 1 PB (1 << 50)
    // Note: off_t is signed 64-bit, so max is 2^63 - 1.
    // 1 PB is safe.
    off_t large_size = (off_t)1 << 50; 

    printf("Truncating to %ld bytes...\n", large_size);
    if (ftruncate(fd, large_size) == -1) {
        perror("ftruncate large");
        return 1;
    }
    printf("Truncated to large size.\n");

    printf("Truncating to 0 bytes...\n");
    if (ftruncate(fd, 0) == -1) {
        perror("ftruncate zero");
        return 1;
    }
    printf("Truncated to 0 bytes.\n");

    close(fd);
    return 0;
}
```

### Environment
- Asterinas version: main ebfa7977b1

### Logs
- In Asterinas
```
~ # /etc/ftruncate_hang.c 
Truncating to 1125899906842624 bytes...
Truncated to large size.
Truncating to 0 bytes...
```
Hangs here, and crtl-C won't work

- In Linux
```
root@ps:~/asterinas# ./test/src/etc/ftruncate_hang.c 
Truncating to 1125899906842624 bytes...
Truncated to large size.
Truncating to 0 bytes...
Truncated to 0 bytes.
root@ps:~/asterinas# 
```

Marsman1996:
Fix #2817 

This addresses a system hang issue occurring when `ftruncate` is called on very large sparse files (e.g., truncating a 1 PB file to 0). The previous implementation of `Vmo::decommit_pages` performed a linear scan over the entire virtual address range, resulting in $O(N)$ complexity where $N$ is the size of the range. For large ranges, this caused the kernel to spin for an extended period, effectively hanging the system.

The new method `next_present` efficiently searches for the next present item in the XArray. It leverages the tree structure to skip empty slots and entire empty subtrees, providing a "sparse iteration" capability.

Modified the decommit logic to use `cursor.next_present()`. Instead of iterating over every single page index in the target range, the loop now jumps directly to the next allocated page.

### DIFF
--- a/kernel/libs/xarray/src/cursor.rs
+++ b/kernel/libs/xarray/src/cursor.rs
@@ -10,7 +10,7 @@ use ostd::{
 };
 
 use crate::{
-    SLOT_SIZE, XArray, XLockGuard,
+    BITS_PER_LAYER, SLOT_SIZE, XArray, XLockGuard,
     entry::NodeEntryRef,
     mark::{NoneMark, XMark},
     node::{Height, XNode},
@@ -225,6 +225,114 @@ impl<'a, P: NonNullPtr + Send + Sync, M> Cursor<'a, P, M> {
 
         self.state.move_to(current_node, self.index);
         self.continue_traverse_to_target();
+    }
+
+    /// Moves the cursor to the first present item at or after the current index.
+    /// If found, updates the cursor's index and state, and returns the index.
+    /// If not found, returns None.
+    pub fn next_present(&mut self) -> Option<u64> {
+        loop {
+            self.traverse_to_target();
+
+            let state = core::mem::take(&mut self.state);
+            if let CursorState::AtNode {
+                node,
+                operation_offset,
+            } = state
+            {
+                if node.is_leaf() {
+                    // Check current slot
+                    if node.entry_with(self.guard, operation_offset).is_some() {
+                        self.state = CursorState::AtNode {
+                            node,
+                            operation_offset,
+                        };
+                        return Some(self.index);
+                    }
+
+                    // Check subsequent slots in this leaf
+                    let mut off = operation_offset + 1;
+                    while off < SLOT_SIZE as u8 {
+                        if node.entry_with(self.guard, off).is_some() {
+                            self.index += (off - operation_offset) as u64;
+                            self.state = CursorState::AtNode {
+                                node,
+                                operation_offset: off,
+                            };
+                            return Some(self.index);
+                        }
+                        off += 1;
+                    }
+
+                    // Move to next leaf
+                    let remaining_in_leaf = SLOT_SIZE as u64 - operation_offset as u64;
+                    self.index = self.index.checked_add(remaining_in_leaf)?;
+                    self.reset();
+                } else {
+                    // Should not happen if traverse_to_target works as expected (it stops at leaf or inactive).
+                    // But if it stops at internal node, it means missing child.
+                    // We should treat it as Inactive logic.
+                    self.reset();
+                }
+            } else {
+                // Inactive. Current index is empty.
+                // Find next present from root.
+                let head = self.xa.head.read_with(self.guard)?;
+                let max_index = head.height().max_index();
+                if self.index > max_index {
+                    return None;
+                }
+
+                if let Some(next_idx) = self.find_next_from_root(self.index) {
+                    self.index = next_idx;
+                    // Loop will continue and traverse_to_target will succeed
+                } else {
+                    return None;
+                }
+            }
+        }
+    }
+
+    fn find_next_from_root(&self, target: u64) -> Option<u64> {
+        let head = self.xa.head.read_with(self.guard)?;
+        self.find_next_in_node(&head, 0, target)
+    }
+
+    fn find_next_in_node(&self, node: &XNode<P>, node_base: u64, target: u64) -> Option<u64> {
+        let height = node.height();
+        let shift = (*height - 1) * BITS_PER_LAYER as u8;
+
+        let start_offset = if target > node_base {
+            ((target - node_base) >> shift) as u8
+        } else {
+            0
+        };
+
+        for off in start_offset..SLOT_SIZE as u8 {
+            let child_base = node_base + ((off as u64) << shift);
+
+            if let Some(entry) = node.entry_with(self.guard, off) {
+                if *height == 1 {
+                    // Leaf node.
+                    if child_base >= target {
+                        return Some(child_base);
+                    }
+                } else {
+                    // Internal node.
+                    let child = entry.left().unwrap();
+                    let search_start = if off == start_offset {
+                        target
+                    } else {
+                        child_base
+                    };
+
+                    if let Some(found) = self.find_next_in_node(&child, child_base, search_start) {
+                        return Some(found);
+                    }
+                }
+            }
+        }
+        None
     }
 }
 
--- a/kernel/src/vm/vmo/mod.rs
+++ b/kernel/src/vm/vmo/mod.rs
@@ -415,26 +415,38 @@ impl Vmo {
         let page_idx_range = get_page_idx_range(&range);
         let mut cursor = locked_pages.cursor_mut(page_idx_range.start as u64);
 
-        let Some(pager) = &self.pager else {
-            for _ in page_idx_range {
-                cursor.remove();
-                cursor.next();
+        // Ensure the cursor points to the first present item in the range
+        if cursor.load().is_none() {
+            if let Some(idx) = cursor.next_present() {
+                cursor.reset_to(idx);
+            } else {
+                return Ok(());
             }
-            return Ok(());
-        };
+        }
 
         let mut removed_page_idx = Vec::new();
-        for page_idx in page_idx_range {
+        while cursor.index() < page_idx_range.end as u64 {
             if cursor.remove().is_some() {
-                removed_page_idx.push(page_idx);
+                removed_page_idx.push(cursor.index() as usize);
             }
+
+            // Advance to the next index, then skip forward to the next present item if this slot is empty
             cursor.next();
+            if cursor.load().is_none() {
+                if let Some(idx) = cursor.next_present() {
+                    cursor.reset_to(idx);
+                } else {
+                    break;
+                }
+            }
         }
 
         drop(locked_pages);
 
-        for page_idx in removed_page_idx {
-            pager.decommit_page(page_idx)?;
+        if let Some(pager) = &self.pager {
+            for page_idx in removed_page_idx {
+                pager.decommit_page(page_idx)?;
+            }
         }
 
         Ok(())


kernel/libs/xarray/src/cursor.rs
```diff
@@ -226,6 +226,114 @@ impl<'a, P: NonNullPtr + Send + Sync, M> Cursor<'a, P, M> {
...
+                            operation_offset,
+                        };
+                        return Some(self.index);
+                    }
+
+                    // Check subsequent slots in this leaf
+                    let mut off = operation_offset + 1;
+                    while off < SLOT_SIZE as u8 {
+                        if node.entry_with(self.guard, off).is_some() {
+                            self.index += (off - operation_offset) as u64;
```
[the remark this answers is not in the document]
Marsman1996: 🤔

kernel/libs/xarray/src/cursor.rs
```diff
@@ -226,6 +226,114 @@ impl<'a, P: NonNullPtr + Send + Sync, M> Cursor<'a, P, M> {
         self.state.move_to(current_node, self.index);
```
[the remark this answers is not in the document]
Marsman1996: The `max_index` is `((SLOT_SIZE as u64) << self.height_shift()) - 1`, so the unwrap cannot actually happen

kernel/libs/xarray/src/cursor.rs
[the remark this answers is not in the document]
Marsman1996: The `max_index` is `((SLOT_SIZE as u64) << self.height_shift()) - 1`, so the unwrap cannot actually happen

tatetian:

lrh2000:
The changes to `kernel/libs/xarray/src/cursor.rs` (i.e., the implementation of `next_present`) are too complex, and I don't think they're necessary. I should have a much simpler version...

kernel/libs/xarray/src/cursor.rs
```diff
@@ -226,6 +226,114 @@ impl<'a, P: NonNullPtr + Send + Sync, M> Cursor<'a, P, M> {
...
+                            node,
+                            operation_offset,
+                        };
+                        return Some(self.index);
+                    }
+
+                    // Check subsequent slots in this leaf
+                    let mut off = operation_offset + 1;
+                    while off < SLOT_SIZE as u8 {
+                        if node.entry_with(self.guard, off).is_some() {
```
cchanging, line 256: Thanks for pointing out and trying to address this issue! This is likely an issue caused by the incomplete functionality and API support in our current version of `XArray`. In the Linux kernel, `XArray`'s marking system has a fixed mark bit called `XA_PRESENT`, which indicates valid child nodes within a node. 
```c
#define XA_MARK_0		((__force xa_mark_t)0U)
#define XA_MARK_1		((__force xa_mark_t)1U)
#define XA_MARK_2		((__force xa_mark_t)2U)
#define XA_PRESENT		((__force xa_mark_t)8U)
```
Additionally, the `XArray` marking system provides APIs like `xas_next_marked` to retrieve the next slot marked with a specific flag. Therefore, the `next_present` mentioned here would essentially be `xas_next_marked` with `XA_PRESENT`.

I think it might be time to enhance `XArray`'s marking system and its related APIs, though this would significantly enlarge the scope of this PR... What do you think? Personally, I feel this issue could be resolved once and for all with a more general approach, especially since it doesn't seem to be a very urgent problem.

Marsman1996: I prefer a complete solution.

So, what shall I do in this PR? rename the API and use const? maybe a C2Rust work could help?

cchanging: > I prefer a complete solution.

If you feel you can handle the final complete implementation, you can add the `XA_PRESENT` mark in this PR and include APIs similar to `xas_next_marked`, ultimately fixing this issue. During this process, you may need to familiarize yourself with the current XArray mark mechanism. 

I'm not sure whether C2Rust can effectively port Linux's mechanism to our codebase, as I think there are still significant differences in our current abstraction and organization...

If you feel that implementing the complete solution is a bit too troublesome for you right now and you can't dedicate much time to it in the short term, you can continue working on it later when you have time, or others can help fix this issue.

lrh2000:
> I should have a much simpler version...

I have a trial. I think this version is much better than the implementation in this PR:
<details>
<summary> [PATCH] Implement `next_present` in xarray </summary>

```patch
6655cb0142cfaf32e88edaeeec63e2545d27 Mon Sep 17 00:00:00 2001
From: Ruihan Li <lrh2000@pku.edu.cn>
Date: Sat, 3 Jan 2026 23:20:06 +0800
Subject: [PATCH] Implement `next_present` in xarray

---
 kernel/libs/xarray/src/cursor.rs | 66 ++++++++++++++++++++++++++++++++
 kernel/libs/xarray/src/node.rs   |  5 +++
 kernel/libs/xarray/src/test.rs   | 44 +++++++++++++++++++++
 3 files changed, 115 insertions(+)

diff --git a/kernel/libs/xarray/src/cursor.rs b/kernel/libs/xarray/src/cursor.rs
index 523304a14..f9076a1de 100644
--- a/kernel/libs/xarray/src/cursor.rs
+++ b/kernel/libs/xarray/src/cursor.rs
@@ -226,6 +226,72 @@ impl<'a, P: NonNullPtr + Send + Sync, M> Cursor<'a, P, M> {
         self.state.move_to(current_node, self.index);
         self.continue_traverse_to_target();
     }
+
+    /// Moves the cursor to the next present item.
+    ///
+    /// If an item is present after the cursor's current index, the cursor will be
+    /// positioned on the corresponding leaf node and the index of the item will be
+    /// returned.
+    ///
+    /// Otherwise, the cursor will stay where it is and a [`None`] will be returned.
+    pub fn next_present(&mut self) -> Option<u64> {
+        self.traverse_to_target();
+
+        let mut index = self.index.checked_add(1)?;
+        let (mut current_node, mut operation_offset) =
+            if let Some((node, offset)) = core::mem::take(&mut self.state).into_node() {
+                (node, offset + 1)
+            } else if let Some(node) = self.xa.head.read_with(self.guard) {
+                let offset = node.entry_offset(index);
+                (node, offset)
+            } else {
+                return None;
+            };
+
+        loop {
+            // If we reach the end of the current node, go to its parent node.
+            if operation_offset == SLOT_SIZE as u8 {
+                let Some(parent_node) = current_node.deref_target().parent(self.guard) else {
+                    self.reset();
+                    return None;
+                };
+
+                operation_offset = current_node.offset_in_parent() + 1;
+                current_node = parent_node;
+                continue;
+            }
+
+            // Otherwise, iterate through the remaining children to see if they contain any items.
+
+            let Some(child_node) = current_node
+                .deref_target()
+                .entry_with(self.guard, operation_offset)
+            else {
+                let index_step = current_node.height().index_step();
+                let Some(new_index) = (index & !(index_step - 1)).checked_add(index_step) else {
+                    self.reset();
+                    return None;
+                };
+
+                index = new_index;
+                operation_offset += 1;
+                continue;
+            };
+
+            if !current_node.is_leaf() {
+                current_node = child_node.left().unwrap();
+                operation_offset = current_node.entry_offset(index);
+                continue;
+            }
+
+            self.index = index;
+            self.state = CursorState::AtNode {
+                node: current_node,
+                operation_offset,
+            };
+            return Some(index);
+        }
+    }
 }
 
 impl<P: NonNullPtr + Send + Sync, M: Into<XMark>> Cursor<'_, P, M> {
diff --git a/kernel/libs/xarray/src/node.rs b/kernel/libs/xarray/src/node.rs
index d167f005d..69e969b86 100644
--- a/kernel/libs/xarray/src/node.rs
+++ b/kernel/libs/xarray/src/node.rs
@@ -95,6 +95,11 @@ impl Height {
     pub(super) fn max_index(&self) -> u64 {
         ((SLOT_SIZE as u64) << self.height_shift()) - 1
     }
+
+    /// Calculates the index step representing one offset at the current height.
+    pub(super) fn index_step(&self) -> u64 {
+        1 << self.height_shift()
+    }
 }
 
 /// The `XNode` is the intermediate node in the tree-like structure of the `XArray`.
diff --git a/kernel/libs/xarray/src/test.rs b/kernel/libs/xarray/src/test.rs
index 9c92c52c9..9d42a2928 100644
--- a/kernel/libs/xarray/src/test.rs
+++ b/kernel/libs/xarray/src/test.rs
@@ -163,6 +163,50 @@ fn cursor_store_sparse() {
     }
 }
 
+#[ktest]
+fn cursor_next_present_sparse() {
+    let xarray_arc: XArray<Arc<u32>> = XArray::new();
+    let mut locked_xarray = xarray_arc.lock();
+    locked_xarray.store(0, Arc::new(1));
+    locked_xarray.store(n!(10), Arc::new(2));
+    locked_xarray.store(n!(100), Arc::new(3));
+
+    let mut cursor = locked_xarray.cursor(0);
+    for i in 0..n!(10) {
+        cursor.reset_to(i);
+        let _ = cursor.load();
+        assert_eq!(cursor.next_present(), Some(n!(10)));
+        assert_eq!(*cursor.load().unwrap().as_ref(), 2);
+    }
+    for i in n!(10)..n!(100) {
+        cursor.reset_to(i);
+        let _ = cursor.load();
+        assert_eq!(cursor.next_present(), Some(n!(100)));
+        assert_eq!(*cursor.load().unwrap().as_ref(), 3);
+    }
+}
+
+#[ktest]
+fn cursor_next_present_continuous() {
+    let xarray_arc: XArray<Arc<u32>> = XArray::new();
+    let mut locked_xarray = xarray_arc.lock();
+
+    let mut cursor = locked_xarray.cursor_mut(0);
+    for i in 0..n!(100) {
+        let value = Arc::new(i);
+        cursor.store(value);
+        cursor.next();
+    }
+
+    cursor.reset_to(0);
+    for i in 0..(n!(100) - 1) {
+        assert_eq!(cursor.next_present(), Some(i as u64 + 1));
+        assert_eq!(*cursor.load().unwrap().as_ref(), i + 1);
+    }
+    assert_eq!(cursor.next_present(), None);
+    assert_eq!(*cursor.load().unwrap().as_ref(), n!(100) - 1);
+}
+
 #[ktest]
 fn set_mark() {
     let xarray_arc: XArray<Arc<u32>, XMark> = XArray::new();
-- 
2.52.0
```
</details>

kernel/libs/xarray/src/cursor.rs
lrh2000: > XA_PRESENT

Won't Linux recycle dead interior nodes that contain no items? If so, I don't feel it's necessary to have this mark?

This is a known limitation of our current implementation. However, Linux shouldn't be affected by this limitation.
https://github.com/asterinas/asterinas/blob/4f62f2db8e355b67da71e6ac164e41fc727fc1ec/kernel/libs/xarray/src/cursor.rs#L360

Honestly, I think support for finding the next marked item is also not quite relevant here. If it's really necessary, I guess it could be implemented based on finding the next present item without significant refactoring.

Marsman1996:
> > I should have a much simpler version...
> 
> I have a trial. I think this version is much better than the implementation in this PR:
> 
> [PATCH] Implement `next_present` in xarray

This looks much better.

I think you can open a new PR and I will close this one.

lrh2000:
> I think you can open a new PR and I will close this one.

OK. That's reasonable. So #2842 can replace this PR. Thanks for your efforts here!

PR #2837, opened 2026-01-04
Support NS16650A UART console, `/dev/ttyS0`, and `console=ttyS0`

lrh2000:
The NS16650A UART is a serial console commonly found in x86. It is also found in some RISC-V/LoongArch platforms, and QEMU uses it in the "virt" machine (`-machine virt`).

Both x86 and LoongArch use it as the early console in OSTD. However, RISC-V does not need it because its early console can be implemented via the SBI.

https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/ostd/src/arch/x86/serial.rs#L15-L16

In any case, we should have an architecture-independent implementation of it. It must live inside OSTD; otherwise, it cannot be used as an early console. I put it in `ostd/src/console/uart_ns16650a.rs`. See the first commit for details.

In the following commits, I added a `uart` component, which registers the UART device as an `AnyConsoleDevice`. After the kernel boots, this `AnyConsoleDevice` will be used as `/dev/ttyS0`.

#### The following things will work after this PR:

 - We can run Asterinas on a real x86 PC that does not have a PS/2 keyboard or mouse. We can communicate with it via the serial console by passing the kernel parameter `console=ttyS0`.
 - We can run Asterinas on x86 QEMU without a VirtIO or framebuffer console.
```
make kernel BOOT_PROTOCOL=linux-legacy32 RELEASE=1
qemu-system-x86_64 \
        -kernel target/osdk/aster-nix-osdk-bin.qemu_elf \
        -append 'init=/bin/busybox console=ttyS0 -- sh' \
        -initrd target/osdk/iso_root/boot/initramfs.cpio.gz \
        -cpu host -enable-kvm -machine microvm -m 4G \
        -serial stdio -monitor none -nographic \
        -device isa-debug-exit,iobase=0xf4,iosize=4 -no-reboot
```
 - We can run Asterinas on riscv QEMU without a VirtIO or framebuffer console.
```
make kernel OSDK_TARGET_ARCH=riscv64
qemu-system-riscv64 \
        -kernel target/osdk/aster-nix-osdk-bin.qemu_elf \
        -append 'init=/bin/busybox console=ttyS0 -- sh' \
        -initrd target/osdk/iso_root/boot/initramfs.cpio.gz \
        -machine virt -m 4G \
        -serial stdio -monitor none -nographic
```

---

https://github.com/asterinas/asterinas/pull/2529 adds sifive UART support, which is not covered by this PR and can ideally go after this one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,10 @@ NO_DEFAULT_FEATURES ?= 0
 COVERAGE ?= 0
 # Specify whether to build regression tests under `test/initramfs/src/apps`.
 ENABLE_BASIC_TEST ?= false
-# Specify the primary system console (supported: hvc0, tty0).
-# - hvc0: The virtio-console terminal.
+# Specify the primary system console (supported: tty0, ttyS0, hvc0).
 # - tty0: The active virtual terminal (VT).
+# - ttyS0: The serial (UART) terminal.
+# - hvc0: The virtio-console terminal.
 # Asterinas will automatically fall back to tty0 if hvc0 is not available.
 # Note that currently the virtual terminal (tty0) can only work with
 # linux-efi-handover64 and linux-efi-pe64 boot protocol.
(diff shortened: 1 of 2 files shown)


junyang-zh (approved):
LGTM!

PR #2838, opened 2026-01-04
Improve the inotify implementation

lrh2000:
Resolve some bugs and perform some optimization:

#### Refactor `add_watch` and `remove_watch`

 - We can rewrite this to avoid using `inode().fs_event_publisher()` and instead use `inode().fs_event_publisher_or_init()`. Since `create_new_subscriber()` will create the extension anyway, this is preferable.
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L140-L144
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L177-L184
 - I don't see the value of this lock. Why not rely on the `watch_map` lock instead?
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L152
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L53
 - This does not seem necessary. Why not find the watch from the `watch_map`?
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L188

#### Refactor `try_read`

 - This may cause the events to be reordered if a user tries to call `read()` concurrently. I think we should take a big lock here:
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L302-L309
```rust
        // This ensures that we report continuous events even when the user program attempts to
        // call `read()` concurrently.
        let _guard = self.read_mutex.lock();
```

####  Optimize `copy_to_user`

 - We can write the byte slices as a whole:
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L632-L634
 - We can make the type `Pod` and write it as a whole:
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L624-L627
 - We can merge this to `fill_zeros`:
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L635-L637
 - We don't need to compute these twice:
https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/src/fs/notify/inotify.rs#L618-L621

#### Remove some `fdinfo` attributes

According to Linux implementations, these attributes only appear if exportfs is enabled. Since we don't support it, let's just remove these attributes instead of reporting fake values.
https://elixir.bootlin.com/linux/v6.18.3/source/fs/notify/fdinfo.c#L42-L72

### DIFF
--- a/kernel/src/fs/notify/inotify.rs
+++ b/kernel/src/fs/notify/inotify.rs
@@ -388,7 +388,7 @@ impl FileLike for InotifyFile {
                     let sdev = inode.fs().sb().fsid;
                     writeln!(
                         f,
-                        "inotify wd:{} ino:{:x} sdev:{:x} mask:{:x} ignored_mask:0 fhandle-bytes:0 fhandle-type:0 f_handle:0",
+                        "inotify wd:{} ino:{:x} sdev:{:x} mask:{:x} ignored_mask:0",
                         wd,
                         inode.ino(),
                         sdev,


lrh2000:
https://github.com/asterinas/asterinas/blob/6eb69687164853a35a52e1b012546f1dd908c735/kernel/src/fs/path/dentry.rs#L281-L282

This logic is also just wrong, as shown in the following tests:
```c
#include <sys/inotify.h>
#include <sys/fcntl.h>
#include <unistd.h>
#include "../test.h"

#define TEST_FILE "/tmp/test1"

FN_TEST(unlink_remove)
{
	int inotify_fd, fd, wd;

	inotify_fd = TEST_SUCC(inotify_init());

	fd = TEST_SUCC(open(TEST_FILE, O_CREAT | O_WRONLY, 0644));
	wd = TEST_SUCC(inotify_add_watch(inotify_fd, TEST_FILE, IN_DELETE_SELF));

	TEST_SUCC(unlink(TEST_FILE));

	// This should fail. `fd` is still alive, so the watch has not been removed yet.
	// The watch will only be removed after the last strong reference of the inode is dropped.
	TEST_ERRNO(inotify_rm_watch(inotify_fd, wd), EINVAL);

	TEST_SUCC(close(fd));
	TEST_SUCC(close(inotify_fd));
}
END_TEST()
```

I'm not going to fix it, as doing so may require a certain amount of refactoring...

tatetian:
@Plucky923 Could you review this PR?

kernel/src/fs/notify/mod.rs
```diff
@@ -146,23 +148,6 @@ impl FsEventPublisher {
...
-        F: FnMut(&Arc<dyn FsEventSubscriber>) -> Option<T>,
-    {
-        let subscribers = self.subscribers.read();
-        for subscriber in subscribers.iter() {
-            if let Some(result) = matcher(subscriber) {
-                return Some(result);
-            }
-        }
-        None
-    }
```
Plucky923, line 165: This function is used to find a subscriber in a general situation, because in the implementation of the notify mechanism, a subscriber can subscribe to the file system events of an inode and also to file system events related to mount... This function provides a method to find the corresponding subscriber from the Publisher, as the subscriber can be implemented based on different notify mechanisms, such as fanotify and inotify....

So in order to be compatible with subsequent implementations, this function can well support these functions. Therefore, even though inotify can conveniently find the corresponding subscriber in the watch map, this method is still necessary.

- https://elixir.bootlin.com/linux/v6.18/source/fs/mount.h#L79
- https://man7.org/linux/man-pages/man7/fanotify.7.html

lrh2000: OK. Thanks for the explanation. I have no clear idea how this relates to fanotify, since we haven't implemented it at the moment... I thought that we could add the method back when we implement fanotify, at which point things will become clear.

That being said, treating this method as a small utility method, I don't have a strong opinion about removing it. I'll revert the change and mark it as `#[expect(dead_code)]`.

Plucky923 (approved):

lrh2000:
https://github.com/asterinas/asterinas/pull/2838#issuecomment-3707584484:

According to the Linux behavior, I think the most reasonable implementation is to trigger `IN_DELETE_SELF` when the `Inode`/`FsEventPublisher` is dropped:
```c
#include <sys/inotify.h>
#include <sys/fcntl.h>
#include <unistd.h>
#include "../test.h"

#define TEST_FILE "/tmp/test1"

FN_TEST(unlink_remove)
{
	int inotify_fd, fd, wd;
	union {
		struct inotify_event ev;
		char raw[sizeof(struct inotify_event) * 2];
	} buf;

	inotify_fd = TEST_SUCC(inotify_init1(O_NONBLOCK));

	fd = TEST_SUCC(open(TEST_FILE, O_CREAT | O_WRONLY, 0644));
	wd = TEST_SUCC(
		inotify_add_watch(inotify_fd, TEST_FILE, IN_DELETE_SELF));

	// `IN_DELETE_SELF` won't be generated because `fd` is still alive.
	TEST_SUCC(unlink(TEST_FILE));
	TEST_ERRNO(read(inotify_fd, buf.raw, sizeof(buf.raw)), EAGAIN);

	// `IN_DELETE_SELF` will be generated once we close `fd`.
	TEST_SUCC(close(fd));
	TEST_RES(read(inotify_fd, buf.raw, sizeof(buf.raw)),
		 _ret == sizeof(buf.raw) && buf.ev.mask == IN_DELETE_SELF);

	TEST_SUCC(close(inotify_fd));
}
END_TEST()
```

I think it's feasible and it should simplify the related logic. I don't think the `accepts_new_subscribers` variable is necessary afterwards. I also checked the Linux implementation, but I didn't find anything similar to `accepts_new_subscribers` (if I'm not missing anything). We may want to hold a FS weak reference in `FsEventPublisher` to update the subscriber statistics in `FsEventPublisher::Drop`.

It's worth a separate PR anyway. I'm not going to add more content to this one.

PR #2842, opened 2026-01-04
Find present pages in xarray

### LINKED ISSUE #2817: System hangs when shrinking a file using `ftruncate` (inconsist with Linux)

lrh2000:
Alternative to https://github.com/asterinas/asterinas/pull/2830. I suppose it's better because the `next_present` logic is much clearer: https://github.com/asterinas/asterinas/pull/2830#issuecomment-3707757852.

Fixes #2817

### DIFF
--- a/kernel/libs/xarray/src/cursor.rs
+++ b/kernel/libs/xarray/src/cursor.rs
@@ -12,7 +12,7 @@ use ostd::{
 use crate::{
     SLOT_SIZE, XArray, XLockGuard,
     entry::NodeEntryRef,
-    mark::{NoneMark, XMark},
+    mark::{NoneMark, PRESENT_MARK, XMark},
     node::{Height, XNode},
 };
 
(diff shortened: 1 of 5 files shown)


kernel/libs/xarray/src/cursor.rs
```diff
@@ -226,6 +226,73 @@ impl<'a, P: NonNullPtr + Send + Sync, M> Cursor<'a, P, M> {
...
+    /// returned.
+    ///
+    /// Otherwise, the cursor will stay where it is and a [`None`] will be returned.
+    pub fn next_present(&mut self) -> Option<u64> {
+        let mut index = self.index.checked_add(1)?;
+        let (mut current_node, mut operation_offset) =
+            if let Some((node, offset)) = core::mem::take(&mut self.state).into_node() {
+                (node, offset + 1)
+            } else if let Some(node) = self.xa.head.read_with(self.guard) {
+                let offset = node.entry_offset(index);
```
cchanging: According to the handling logic here, if the current `Cursor` is invalid and the index corresponds to the `SLOT_SIZE`-th slot of the root node, then in the loop below, it would directly return None due to the absence of a parent. However, after the current index maps to a leaf node, could there still be valid slots after it?

cchanging (commented):
Thanks for clearing the logics! The code is in great shape! 

In my previous comment about `XA_PRESENT`, my original thought was that using the mark system to locate the next valid slot in the current node would be more efficient, so implementing it that way might be the final form. However, I agree with your point that the overall logic would be similar to the current approach, so I think it's acceptable to introduce the `next_present` API first.

kernel/libs/xarray/src/cursor.rs
lrh2000: > the current Cursor is invalid and the index corresponds to the SLOT_SIZE-th slot of the root node

How can `entry_offset()` return `SLOT_SIZE`? `entry_offset()` return something like:
https://github.com/asterinas/asterinas/blob/4f62f2db8e355b67da71e6ac164e41fc727fc1ec/kernel/libs/xarray/src/node.rs#L89-L90

This is guaranteed to be in `0..SLOT_SIZE` because `SLOT_SIZE.is_power_of_two() && SLOT_MASK == SLOT_SIZE - 1`.

> after the current index maps to a leaf node, could there still be valid slots after it?

I don't understand how there can be nodes after the root node.

---

But the logic here is indeed problematic in another way. If the `index` is beyond the maximum index, we should return `None`. However, I forget to check this condition:
https://github.com/asterinas/asterinas/blob/4f62f2db8e355b67da71e6ac164e41fc727fc1ec/kernel/libs/xarray/src/cursor.rs#L139-L142

I fixed the problem and added a new test to confirm the behavior.

lrh2000:
> using the mark system to locate the next valid slot in the current node would be more efficient

Assuming we can remove the empty interior nodes, I'm not sure how it could be more efficient. In other words, if we don't have empty interior nodes, finding a marked item should be as efficient as finding a present item. Am I missing anything?
https://github.com/asterinas/asterinas/blob/4f62f2db8e355b67da71e6ac164e41fc727fc1ec/kernel/libs/xarray/src/cursor.rs#L360

cchanging:
> finding a marked item should be as efficient as finding a present item

My understanding is that with the present mark, we can use bit operations within a node to quickly find the next present child. Of course, this may only offer a constant‑level efficiency improvement.

lrh2000:
> My understanding is that with the present mark, we can use bit operations within a node to quickly find the next present child. Of course, this may only offer a constant‑level efficiency improvement.

Thanks for your explanation. I didn't understand this point before, but now I see what you meant. I think it's okay to find the item according to the mark, just like Linux's implementation. The latest commit should address it. [force-push diff](https://github.com/asterinas/asterinas/compare/a24e515b890c8cc1a7a6de1159aac467c8106f34..9b778d18e27be7ff61d48038a6156c7e128fd15c)

cchanging (approved):

PR #2845, opened 2026-01-05
Improve fsnotify functions to pass name information more effectively

### LINKED ISSUE #2844: Improvement ideas for fsnotify to pass the name info to an inode's parent
## Background

The FS event notification subsystem ([`kernel/src/fs/notify/`](), Linux's counterpart is called `fsnotify`) is responsible for delivering FS events generated by FS event publishers (inodes) to subscribers (e.g., inotify files).  A FS event may be attached with an optional _name_ field.
### Using `Path::effective_name` is inappropriate

When an FS event occurs on an inode (referred to by a `Path`), the current code informs the path's parent the name of the inode. This name is obtained via `Path::effective_name`:

https://github.com/asterinas/asterinas/blob/a42b4cd078cb24b0f4c49e1a05a86a997a32a79e/kernel/src/fs/notify/mod.rs#L406-L416

But actually, it is the  _real name_ of the inode (`path.dentry.name()`, rather than the _effective name_ (`path.effective_name()`), that should be passed. This is because we are telling an directory inode which _child_  has interesting events. So we must use the child's real name.

From another angel, FS events do not spread across the boundaries of mounts. At least, this is my understanding and how the current implementation works:

https://github.com/asterinas/asterinas/blob/a42b4cd078cb24b0f4c49e1a05a86a997a32a79e/kernel/src/fs/notify/mod.rs#L424-L434

(Notice `parent_within_mount` is invoked)

An effective name of a `Path` differs from the real name when the path is the root of a mount. In other words, effective names are only useful when referencing inodes across mount boundaries. So using `Path::effective_name` is inappropriate.
### Optimizing away unnecessary string allocation for `notify_parent`

As shown above, `notify_parent` requires the caller to pass the name of child. This means the caller has to construct the name _unconditioanlly_:

https://github.com/asterinas/asterinas/blob/a42b4cd078cb24b0f4c49e1a05a86a997a32a79e/kernel/src/fs/notify/mod.rs#L405-L416

But both `notify_parent` and the follow-up `notify_inode` have some short-circuit logic to avoid calling `publish_event` when possible. This means the name constructed by the caller might be useless. So we should delay the construction of the child inode's name to the last minute. This means passing a closure that returns a string. This techniques has been used in circumstances when it can help:

https://github.com/asterinas/asterinas/blob/a42b4cd078cb24b0f4c49e1a05a86a997a32a79e/kernel/src/fs/notify/mod.rs#L287-L291

The `notify_inode` API should also adopt this technique:

```rust
fn notify_inode(
    path: &Path,
    mut events: FsEvents,
    name: Option<impl FnOnce() -> String>
) { ... }
```

This allows us to simplify the signature of `notify_parent` to the follows:

```rust
fn notify_parent(path: &Path, mut events: FsEvents) { ... }
```

And we can construct a closure that generates the name tag inside `notify_parent`.

Plucky923:
Fix the issue where fsnotify uses `effective name()` #2794 , and optimize performance by delaying the construction of strings through closures.

- Add the `Path::name()` method, change `effective_name()` to private 
- `notify_parent/notify_inode` use closure delay string construction
- update all related call points

Fixes #2844.

### DIFF
--- a/kernel/src/fs/notify/mod.rs
+++ b/kernel/src/fs/notify/mod.rs
@@ -249,7 +249,7 @@ pub fn on_access(file: &Arc<dyn FileLike>) {
     {
         return;
     }
-    notify_parent(path, FsEvents::ACCESS, path.effective_name());
+    notify_parent(path, FsEvents::ACCESS);
 }
 
 /// Notifies that a file was modified.
@@ -267,7 +267,7 @@ pub fn on_modify(file: &Arc<dyn FileLike>) {
     {
         return;
     }
-    notify_parent(path, FsEvents::MODIFY, path.effective_name());
+    notify_parent(path, FsEvents::MODIFY);
 }
 
 /// Notifies that a path's content was changed.
@@ -280,7 +280,7 @@ pub fn on_change(path: &Path) {
     {
         return;
     }
-    notify_parent(path, FsEvents::MODIFY, path.effective_name());
+    notify_parent(path, FsEvents::MODIFY);
 }
 
 /// Notifies that a file was deleted from a directory.
@@ -297,11 +297,10 @@ pub fn on_delete(
         return;
     }
 
-    let name = name();
     if inode.type_() == InodeType::Dir {
-        notify_inode(dir_inode, FsEvents::DELETE | FsEvents::ISDIR, Some(name))
+        notify_inode_with_name(dir_inode, FsEvents::DELETE | FsEvents::ISDIR, name)
     } else {
-        notify_inode(dir_inode, FsEvents::DELETE, Some(name))
+        notify_inode_with_name(dir_inode, FsEvents::DELETE, name)
     }
 }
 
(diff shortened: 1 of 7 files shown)


kernel/src/fs/notify/mod.rs
```diff
@@ -310,32 +309,32 @@ pub fn on_link_count(inode: &Arc<dyn Inode>) {
     if !inode.fs().fs_event_subscriber_stats().has_any_subscribers() {
         return;
     }
-    notify_inode(inode, FsEvents::ATTRIB, None);
+    notify_inode(inode, FsEvents::ATTRIB, None::<fn() -> String>);
```
tatetian: This looks silly. If it has to be written this way, then a more user-friendly API design is to split `notify_inode` into two:

```rust
fn notify_inode(inode: &Arc<dyn Inode>, events: FsEvents) { ... }

fn notify_inode_with_name(inode: &Arc<dyn Inode>, events: FsEvents, name: impl FnOnce() -> String) { ... }
```

kernel/src/fs/path/mod.rs
```diff
@@ -141,11 +141,20 @@ impl Path {
         path_name
     }
 
+    /// Gets the real name of the `Path` from its dentry.
+    ///
+    /// This returns the actual name stored in the dentry, without
+    /// crossing mount boundaries. For cross-mount name resolution,
+    /// use `effective_name()` instead.
```
tatetian: Do not mention the name of a private method (`effective_name`) in the doc of a public method.

tatetian (approved):

Issue #2832, opened 2026-01-03
`OSTD initialized. Preparing components` process was slowed down drastically


LordaeronESZ:
### Describe the bug

(https://github.com/asterinas/asterinas/commit/71681dd947b7cfc46ef28b381d16484c740be347) seems to have slowed down the `[kernel] OSTD initialized. Preparing components.` process drastically. On my local machine, it took about 55 seconds to finish, compared to the previous commit (https://github.com/asterinas/asterinas/commit/0c30edb5da67c21caa8613573086856c58da4cfd) which only took about 5 seconds.

### To Reproduce

1. `git pull upstream main --rebase` to synchronize with the latest repository version.
2. `git reset --hard 71681dd947b7cfc46ef28b381d16484c740be347` and `make run_kernel`, then calculate the time taken to start up.
3. `git reset --hard 0c30edb5da67c21caa8613573086856c58da4cfd` and `make run_kernel`, then calculate the time taken to start up.

### Expected behavior

The startup times of the two versions are not significantly different.

### Environment

- Docker image: official asterinas:0.17.0-20251228
- CPU: Intel Ultra5 225H

StevenJiang1110:
I can reproduce the issue on my machine (dual AMD EPYC 9335 32-Core CPUs). I haven't identified the root cause yet, however. Could @junyang-zh take a look at this?

junyang-zh:
The regression comes from the debug assertions in `DmaStream::map`. And `aster_virtio::device::block` creates large DMA segments. The performance is not affected when debug assertions are off. We can try optimize it.

lrh2000:
It also seems that the CI is quite slow now, taking more than half an hour, which is painful to wait for. Is this related?

junyang-zh:
> It also seems that the CI is quite slow now, taking more than half an hour, which is painful to wait for. Is this related?

RISC-V ktest has debug assertions and is slowed down to 30 min. I think it's related.

Optimizing it may require implementing a segment tree in OSTD. At a cost of complexity?

tatetian:
> > It also seems that the CI is quite slow now, taking more than half an hour, which is painful to wait for. Is this related?
> 
> RISC-V ktest has debug assertions and is slowed down to 30 min. I think it's related.
> 
> Optimizing it may require implementing a segment tree in OSTD. At a cost of complexity?

This is quite a well-defined problem. Just ask LLMs to do the heavy-lifting and you can sit back and relax, just reviewing the code.

PR #2846, opened 2026-01-05
Rename aster-nix to aster-kernel

### LINKED ISSUE #2836: Rename `aster-nix` to `asterinas`
Now that we have the NixOS-based distribution and use Nix in our build system, still calling the [kernel crate](https://github.com/asterinas/asterinas/blob/eb83a38d127960c067b5b3db93ca048e0e9f18ee/kernel/Cargo.toml#L2) `aster-nix` would cause confusion. So I suppose that we should  rename it to `asterinas` (or `aster-kernel`). In addition to the `name` field in `Cargo.toml`, all references to `aster-nix` should be updated.

cqs21:
Fix #2836

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,9 +26,9 @@ virtio-net.pcap
 /result_*.json
 
 # profile results
-aster-nix-profile-*.json
-aster-nix-profile-*.folded
-aster-nix-profile-*.svg
+aster-kernel-profile-*.json
+aster-kernel-profile-*.folded
+aster-kernel-profile-*.svg
 
 # distro results
 distro/result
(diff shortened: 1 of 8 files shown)


tatetian:
@cqs21 The CI failures seemed to be caused by timeout. I am wondering why this happens after changing the crate name and if they can be solved by raising timeout thresholds.

cqs21:
> I am wondering why this happens after changing the crate name and if they can be solved by raising timeout thresholds.

After the DMA APIs were refactored, many `debug_assert` statements were added, causing AsterNixOS tests to timeout. This issue should be resolvable by raising the thresholds.

junyang-zh:
#2832 should have fixed it. Have you rebased?

tatetian (approved):

PR #2847, opened 2026-01-05
Fix the DMA debug assertion regression

### LINKED ISSUE #2832: `OSTD initialized. Preparing components` process was slowed down drastically

junyang-zh:
It makes the debug assertion 4096 times faster?

I found that the DMA assertion is per-byte. But per-page is suffice. That's why the performance degrades so much. I found this easy fix after implementing a proper segment tree with 330 LoC... But I think we can still keep the < 100 LoC `RangeCounter` implementation since we want simple and stupid things in OSTD.

Fix #2832

### DIFF
--- a/ostd/src/mm/dma/util.rs
+++ b/ostd/src/mm/dma/util.rs
@@ -161,29 +161,33 @@ pub(super) unsafe fn unprepare_dma(pa_range: &Range<Paddr>, daddr: Option<Daddr>
 unsafe fn alloc_unprotect_physical_range(pa_range: &Range<Paddr>) {
     use alloc::{vec, vec::Vec};
 
+    debug_assert!(pa_range.start.is_multiple_of(PAGE_SIZE));
+    debug_assert!(pa_range.end.is_multiple_of(PAGE_SIZE));
+    let pfn_range = pa_range.start / PAGE_SIZE..pa_range.end / PAGE_SIZE;
+
     let mut refcnts = PADDR_REF_CNTS.lock();
-    let ranges = refcnts.add(pa_range);
+    let ranges = refcnts.add(&pfn_range);
     #[cfg(target_arch = "x86_64")]
     crate::arch::if_tdx_enabled!({
         for partial in ranges {
-            debug_assert_eq!(partial, pa_range.clone());
+            debug_assert_eq!(partial, pfn_range.clone());
             // SAFETY:
             //  - The provided physical address is page aligned.
             //  - The provided physical address range is in bounds.
             //  - All of the physical pages are untyped memory.
             unsafe {
                 crate::arch::tdx_guest::unprotect_gpa_tdvm_call(
-                    partial.start,
-                    partial.end - partial.start,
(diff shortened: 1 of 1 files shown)


ostd/src/mm/dma/util.rs
```diff
@@ -194,21 +198,25 @@ unsafe fn alloc_unprotect_physical_range(pa_range: &Range<Paddr>) {
...
             //  - All of the physical pages are untyped memory.
             unsafe {
                 crate::arch::tdx_guest::protect_gpa_tdvm_call(
-                    pa_range.start,
-                    pa_range.end - pa_range.start,
+                    removed.start * PAGE_SIZE,
+                    removed.len() * PAGE_SIZE,
                 )
-                .expect("Failed to protect the DMA segment in TDX guest");
+                .expect("failed to unprotect the DMA segment in TDX guest");
```
lrh2000: If you want to change `Failed` -> `failed`, please change line 183 as well?
https://github.com/asterinas/asterinas/blob/48dd714d685f3b0caca1c9a8bdb6de91005b720c/ostd/src/mm/dma/util.rs#L183

I'm okay either way, but I really want to be consistent at least within one file...

ostd/src/mm/dma/util.rs
```diff
@@ -161,29 +161,33 @@ pub(super) unsafe fn unprepare_dma(pa_range: &Range<Paddr>, daddr: Option<Daddr>
...
             //  - All of the physical pages are untyped memory.
             unsafe {
                 crate::arch::tdx_guest::unprotect_gpa_tdvm_call(
-                    partial.start,
-                    partial.end - partial.start,
+                    partial.start * PAGE_SIZE,
+                    partial.len() * PAGE_SIZE,
                 )
-                .expect("Failed to protect the DMA segment in TDX guest");
+                .expect("failed to protect the DMA segment in TDX guest");
```
lrh2000: I think line 183 should use "unprotect" and line 219 should use "protect"... Right?

Seems to address the copilot comment on the wrong line.

PR #2848, opened 2026-01-05
Fix `FsEvent::{DELETE_SELF, IN_IGNORED}` generation

lrh2000:
There are some minor conflicts with https://github.com/asterinas/asterinas/pull/2845. I think this PR can wait for #2845.

---

This PR addresses the problem mentioned in https://github.com/asterinas/asterinas/pull/2838#issuecomment-3708113761. `DELETE_SELF` and `IGNORED` should be published when dropping the `FsEventPublisher` (assuming that the link count has already reached zero). It should not happen in `Dentry::unlink`. This PR adds a regression test to confirm the behavior of the Linux system.

In addition, this PR adds support for `InotifyControls::ONESHOT`. I don't think `InotifyControls::ISDIR` makes sense so I remove it.

TODOs in the gvisor tests have also been made clearer:
```
# TODO: Report `MOVE_*` events at the `rename()` system call.
Inotify.MoveGeneratesEvents
Inotify.MoveWatchedTargetGeneratesEvents

# TODO: Support the `splice()` system call.
Inotify.SpliceOnInotifyFD
Inotify.SpliceOnWatchTarget

# Symbolic links to the busybox binary does not work:
# "gvisor_test_temp_1103_1767603858010282988: applet not found"
Inotify.Exec

# TODO: Support the `EXCL_UNLINK` flag.
Inotify.ExcludeUnlink_NoRandomSave
Inotify.ExcludeUnlinkDirectory_NoRandomSave
Inotify.ExcludeUnlinkMultipleChildren_NoRandomSave
Inotify.ExcludeUnlinkInodeEvents_NoRandomSave
```

### DIFF
--- a/kernel/src/fs/notify/inotify.rs
+++ b/kernel/src/fs/notify/inotify.rs
@@ -169,19 +169,9 @@ impl InotifyFile {
         let subscriber = inotify_subscriber.clone() as Arc<dyn FsEventSubscriber>;
 
         let inode = path.inode();
-        if !inode
+        inode
             .fs_event_publisher_or_init()
-            .add_subscriber(subscriber)
-        {
-            // FIXME: This can be triggered by `unlink()` race conditions or
-            // `inotify_add_watch("/proc/self/fd/[fd]")`, where `[fd]` is a file descriptor
-            // pointing to a deleted file. Although Linux allows this, we don't support it, so an
-            // error is returned here.
-            return_errno_with_message!(
-                Errno::ENOENT,
-                "adding an inotify watch to a deleted inode is not supported yet"
-            );
-        }
+            .add_subscriber(subscriber);
 
         let wd = inotify_subscriber.wd();
         let entry = SubscriberEntry {
--- a/kernel/src/fs/notify/mod.rs
+++ b/kernel/src/fs/notify/mod.rs
@@ -3,7 +3,7 @@
 use alloc::{sync::Arc, vec::Vec};
 use core::{
     any::Any,
-    sync::atomic::{AtomicBool, AtomicU32, Ordering},
+    sync::atomic::{AtomicU32, Ordering},
 };
 
 use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
@@ -29,8 +29,8 @@ pub struct FsEventPublisher {
     subscribers: RwLock<Vec<Arc<dyn FsEventSubscriber>>>,
     /// All interesting FS event types (aggregated from all subscribers).
     all_interesting_events: AtomicFsEvents,
-    /// Whether this publisher still accepts new subscribers.
-    accepts_new_subscribers: AtomicBool,
+    /// FS events to be published when dropped.
+    events_on_drop: AtomicFsEvents,
     /// A weak reference to the file system.
     fs: Weak<dyn FileSystem>,
 }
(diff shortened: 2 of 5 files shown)


lrh2000:
This does not work for ext2 in our current design because ext2 caches dead inodes indefinitely. Is this the correct design?
https://github.com/asterinas/asterinas/blob/d7cefdb761de7ec55929fbb733104315a48b6b9d/kernel/src/fs/ext2/block_group.rs#L363

tatetian:
> This does not work for ext2 in our current design because ext2 caches dead inodes indefinitely. Is this the correct design?
> 
> https://github.com/asterinas/asterinas/blob/d7cefdb761de7ec55929fbb733104315a48b6b9d/kernel/src/fs/ext2/block_group.rs#L363

The `BlockGroup` does have inode deallocation logic.

https://github.com/asterinas/asterinas/blob/d7cefdb761de7ec55929fbb733104315a48b6b9d/kernel/src/fs/ext2/block_group.rs#L175C12-L181

What makes you say that "ext2 caches dead inodes indefinitely"? Maybe this phenomenon (I assume that you observed) is caused by other code?

lrh2000:
> The `BlockGroup` does have inode deallocation logic.
> 
> https://github.com/asterinas/asterinas/blob/d7cefdb761de7ec55929fbb733104315a48b6b9d/kernel/src/fs/ext2/block_group.rs#L175C12-L181
> 
> What makes you say that "ext2 caches dead inodes indefinitely"? Maybe this phenomenon (I assume that you observed) is caused by other code?

AFAICK, this logic can only be triggered in `Inode::sync_metadata`:
https://github.com/asterinas/asterinas/blob/3dd881dec5153ba05596db906f72fce3fff0fcaf/kernel/src/fs/ext2/inode.rs#L1362-L1374

First, someone has to invoke the sync manually, perhaps via `fsync`. However, we cannot rely on this because the inotify behavior should not be related to `fsync`...

The sync behavior is also incorrect because `fsync` should not resize a dead inode to zero size.

Another way to trigger the logic that frees inodes is to sync the entire file system. However, this also needs to be done manually:
https://github.com/asterinas/asterinas/blob/3dd881dec5153ba05596db906f72fce3fff0fcaf/kernel/src/fs/ext2/fs.rs#L411-L415

lrh2000:
> The `BlockGroup` does have inode deallocation logic.

Setting aside the current design, when do you think the inode/block deallocation logic should be invoked?

It should be invoked once the hard link count reaches zero **and all the file handles using the inode disappear** (since writing to the inode's blocks is possible via those inode handles).

It seems, then, that we should do this in `Inode::drop`. However, it's not possible in our current design because the inode cache holds strong references.

Perhaps the inode cache should hold weak references. We would then need an additional wait queue to synchronize race conditions caused by `Inode::drop` and concurrent inode lookups.

**EDIT:** Something like:
```rust
// OSTD provides the `Destruct` type to allow ownership to be taken when dropping.
type ArcInode = Destruct<Arc<dyn Inode>>;

impl Destructure for Arc<dyn Inode> {
    fn on_drop(self: Self) {
        let Some(unique) = Arc::into_unique(self) else {
            return;
        };
        let fs = unique.fs();
        // `release_inode` should either drop the inode or
        // reset the extension.
        //
        // Note that it can have a mutable reference to the inode,
        // so it's quite easy for it to reset the extension,
        // which will drop the inotify watches.
        fs.release_inode(unique);
    }
}

struct BlockGroup {
    inode_cache: BTreeMap<u32, CachedInode>,

    // Wait on this queue if `CachedInode` is `Active`
    // but the weak inode cannot be upgraded
    // (race conditions between `on_drop` and `release_inode`).
    wait_queue: WaitQueue,
}

enum CachedInode {
    // The inode is active in the VFS layer.
    // We'll be notified via `release_inode`
    // when dropping the inode's last strong reference count.
    Active(WeakInode),
    // The inode is not active.
    // We can remove this after syncing the metadata and the data.
    Inactive(ArcUnique<dyn Inode>),
}
```

lrh2000:
I've extracted the part independent of the ext2/exfact implementation to #2859.

lrh2000:
Closed this because we need to fix the lifetime of on-disk inodes first. Feel free to take on the task if someone wants to. Perhaps I should open an issue...

PR #2850, opened 2026-01-06
Cleanups for pseudo `Path`

vvvvsv:
This PR contains some independent prerequisite changes for https://github.com/asterinas/asterinas/pull/2798.

The commit "Remove `inode_handle/dyn_cap.rs` and `HandleInner`" is a cleanup for https://github.com/asterinas/asterinas/pull/2563#discussion_r2517648267.

The commit "Polling a `fd` opened with `O_PATH` should return `POLLNVAL`" is a bug-fix.

The commit "Support allocating inodes from `PseudoFs`" ensures that the `ino`s of `PseudoInode`s are allocated from their corresponding `PseudoFs`.

### DIFF
--- a/kernel/src/fs/epoll/file.rs
+++ b/kernel/src/fs/epoll/file.rs
@@ -16,7 +16,7 @@ use crate::{
         file_handle::FileLike,
         file_table::{FdFlags, FileDesc, get_file_fast},
         path::RESERVED_MOUNT_ID,
-        pseudofs::anon_inodefs_shared_inode,
+        pseudofs::AnonInodeFs,
         utils::{CreationFlags, Inode},
     },
     prelude::*,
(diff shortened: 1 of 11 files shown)


test/src/apps/file_io/access_err.c
```diff
@@ -153,6 +159,11 @@ FN_TEST(writeable)
 	f_rdlck.l_type = F_RDLCK;
 	f_wrlck.l_type = F_WRLCK;
 
+	struct pollfd pfd;
```
lrh2000: Following this file's current style, please move this to the beginning of the method.

More context: The code style used in the Linux kernel requires that all variables be defined at the beginning. I don't intend to enforce this in all regression tests, since it's an outdated rule from an ancient version of the C standard. However, I want to maintain the rule if it is already used in a file.

vvvvsv:
@cchanging @cqs21 Could you please help to review this PR?

cchanging (approved):

PR #2852, opened 2026-01-06
Introduce `DmaStream::alloc` to reduce memory copy in TDX

taosue:
As discussion in https://github.com/asterinas/asterinas/pull/2351#discussion_r2650916943, bounce buffer is introduced in TDX and leads to one extra data copy. This PR is for reducing the memory copy in TDX by introducing `DmaStream::alloc`.

To better demonstrate the effectiveness of this optimization, I ran three relatively stable benchmarks related to DMA streams, and observed a significant performance improvement. Below are the details of these three benchmarks. I ran each benchmark 5 times before and after introducing optimizations.

Based on the average of the 5 runs, `[Ext2] The throughput of creating/deleting small files (10KB)` improved by 156.9%, `[Ext2] The bandwidth of copying data between files` improved by 27.4%, `[Ext2] The bandwidth of sequential writes (IOMMU disabled on Asterinas)` improved by 23.8%.

### DIFF
--- a/ostd/src/mm/dma/test.rs
+++ b/ostd/src/mm/dma/test.rs
@@ -16,7 +16,7 @@ mod dma_coherent {
     use super::*;
 
     #[ktest]
-    fn map_with_coherent_device() {
+    fn alloc_with_coherent_device() {
         let dma_coherent = DmaCoherent::alloc(1, true).unwrap();
         assert_eq!(dma_coherent.size(), PAGE_SIZE);
     }
@@ -89,6 +89,19 @@ mod dma_coherent {
         assert_eq!(&read_buf[2 * PAGE_SIZE..3 * PAGE_SIZE], &pattern1[..]);
         assert_eq!(&read_buf[3 * PAGE_SIZE..4 * PAGE_SIZE], &pattern2[..]);
     }
+
+    #[ktest]
+    fn alloc_uninit_dma_coherent() {
+        let dma_coherent = DmaCoherent::alloc_uninit(1, false).unwrap();
+        assert_eq!(dma_coherent.size(), PAGE_SIZE);
+
+        let buf_write = vec![0xCDu8; PAGE_SIZE];
+        dma_coherent.write_bytes(0, &buf_write).unwrap();
+
+        let mut buf_read = vec![0u8; PAGE_SIZE];
+        dma_coherent.read_bytes(0, &mut buf_read).unwrap();
+        assert_eq!(buf_write, buf_read);
+    }
 }
 
 mod dma_stream {
@@ -115,6 +128,7 @@ mod dma_stream {
         dma_stream.write_bytes(0, &buf_write).unwrap();
         dma_stream.sync_to_device(0..2 * PAGE_SIZE).unwrap();
         let mut buf_read = vec![0u8; 2 * PAGE_SIZE];
+        dma_stream.sync_from_device(0..2 * PAGE_SIZE).unwrap();
         dma_stream.read_bytes(0, &mut buf_read).unwrap();
         assert_eq!(buf_write, buf_read);
     }
@@ -134,6 +148,7 @@ mod dma_stream {
         let mut buf_read = vec![0u8; 2 * PAGE_SIZE];
         let buf_write = vec![1u8; 2 * PAGE_SIZE];
         let mut reader = dma_stream.reader().unwrap();
+        dma_stream.sync_from_device(0..2 * PAGE_SIZE).unwrap();
         reader.read(&mut buf_read.as_mut_slice().into());
         assert_eq!(buf_read, buf_write);
     }
@@ -199,8 +214,47 @@ mod dma_stream {
             .unwrap();
         let mut read_buf = [0u8; 128];
         dma_stream
+            .sync_from_device(PAGE_SIZE - 64..PAGE_SIZE + 64)
+            .unwrap();
+        dma_stream
             .read_bytes(PAGE_SIZE - 64, &mut read_buf)
             .unwrap();
         assert_eq!(read_buf, small_buf);
+    }
+
+    #[ktest]
+    fn alloc_dma_stream() {
+        let dma_stream = DmaStream::<FromAndToDevice>::alloc(2, false).unwrap();
+        assert_eq!(dma_stream.size(), 2 * PAGE_SIZE);
+
+        // Verify allocated memory is zeroed
+        let mut buf_read = vec![1u8; 2 * PAGE_SIZE];
+        dma_stream.sync_from_device(0..2 * PAGE_SIZE).unwrap();
+        dma_stream.read_bytes(0, &mut buf_read).unwrap();
+        assert_eq!(buf_read, vec![0u8; 2 * PAGE_SIZE]);
+
+        let buf_write = vec![0xABu8; 2 * PAGE_SIZE];
+        dma_stream.write_bytes(0, &buf_write).unwrap();
+        dma_stream.sync_to_device(0..2 * PAGE_SIZE).unwrap();
+
+        let mut buf_read = vec![0u8; 2 * PAGE_SIZE];
+        dma_stream.sync_from_device(0..2 * PAGE_SIZE).unwrap();
+        dma_stream.read_bytes(0, &mut buf_read).unwrap();
+        assert_eq!(buf_write, buf_read);
+    }
+
+    #[ktest]
+    fn alloc_uninit_dma_stream() {
+        let dma_stream = DmaStream::<FromAndToDevice>::alloc_uninit(1, false).unwrap();
+        assert_eq!(dma_stream.size(), PAGE_SIZE);
+
+        let buf_write = vec![0xCDu8; PAGE_SIZE];
+        dma_stream.write_bytes(0, &buf_write).unwrap();
+        dma_stream.sync_to_device(0..PAGE_SIZE).unwrap();
+
+        let mut buf_read = vec![0u8; PAGE_SIZE];
+        dma_stream.sync_from_device(0..PAGE_SIZE).unwrap();
+        dma_stream.read_bytes(0, &mut buf_read).unwrap();
+        assert_eq!(buf_write, buf_read);
     }
 }


ostd/src/mm/dma/dma_stream.rs
```diff
@@ -149,6 +169,36 @@ impl<D: DmaDirection> DmaStream<D> {
...
+            let paddr_range = segment.paddr_range();
+
+            (Inner::Segment(segment), paddr_range)
+        } else {
+            let (kva, paddr) = alloc_kva(nframes, is_cache_coherent, ZEROED)?;
+
+            (Inner::Kva(kva, paddr), paddr..paddr + nframes * PAGE_SIZE)
+        };
+
+        let map_daddr = unsafe { prepare_dma(&paddr_range) };
```
junyang-zh: Add safety comments to it?

```
// SAFETY: The physical address range is untyped DMA memory before `drop`.
```

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,30 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
 
 impl<D: DmaDirection> DmaStream<D> {
+    /// Allocates a region of physical memory and zeroes all bytes for streaming
+    /// DMA access.
+    ///
+    /// The `is_cache_coherent` argument specifies whether the target device
+    /// that the DMA mapping is prepared for can access the main memory in a
+    /// CPU cache coherent way or not.
+    pub fn alloc(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
+        Self::alloc_impl::<true>(nframes, is_cache_coherent)
```
tatetian: I don't think taking the `zeroed` argument as a type parameter can be justified. Make it a normal value argument.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,30 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
 }
 
 impl<D: DmaDirection> DmaStream<D> {
+    /// Allocates a region of physical memory and zeroes all bytes for streaming
+    /// DMA access.
+    ///
+    /// The `is_cache_coherent` argument specifies whether the target device
+    /// that the DMA mapping is prepared for can access the main memory in a
+    /// CPU cache coherent way or not.
+    pub fn alloc(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
```
tatetian, line 111: Add a section in the Rust doc to compare `alloc` with segment allocation plus `map`.

Add unit tests for the two new APIs.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,30 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+        Self::alloc_impl::<true>(nframes, is_cache_coherent)
+    }
+
+    /// Allocates a region of physical memory for streaming DMA access.
+    ///
+    /// The memory content is uninitialized. The `is_cache_coherent` argument
+    /// specifies whether the target device that the DMA mapping is prepared for
+    /// can access the main memory in a CPU cache coherent way or not.
+    pub fn alloc_uninitialized(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
```
tatetian: Simplify the name to `alloc_uninit`. Rust std include many methods with the `_uninit` suffix.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,30 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
 #[derive(Debug)]
 enum Inner {
     Segment(USegment),
+    Kva(KVirtArea, Paddr),
     Both(KVirtArea, Paddr, USegment),
 }
 
 impl<D: DmaDirection> DmaStream<D> {
+    /// Allocates a region of physical memory and zeroes all bytes for streaming
+    /// DMA access.
```
tatetian: The summary sentences of `DmaStream::alloc` and `DmaStream::alloc_uninit` are asymmetric and should be re-written.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,30 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+    /// The `is_cache_coherent` argument specifies whether the target device
+    /// that the DMA mapping is prepared for can access the main memory in a
+    /// CPU cache coherent way or not.
+    pub fn alloc(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
+        Self::alloc_impl::<true>(nframes, is_cache_coherent)
+    }
+
+    /// Allocates a region of physical memory for streaming DMA access.
+    ///
+    /// The memory content is uninitialized. The `is_cache_coherent` argument
```
tatetian: Do not repeat yourself. Rewrite the doc this way.

```rust
/// This method is same as [`DmaStream::alloc`]
/// except that it skips zeroing the memory of newly-allocated DMA region.
```

lrh2000:
TDX replaces the physical page when marking it as shared (`unprotect_gpa_tdvm_call`)... So it doesn't make sense to zero the page beforehand (If implemented as such, `alloc` cannot guarantee that the page is initialized)?

We may want to reconsider the DMA allocation APIs based on that (including `DmaCoherent::alloc`, which is not touched by this PR but it suffers from the same problem).

Perhaps we can do this:
```rust
fn alloc() {
    let result = Self::alloc_uninitialized()?;
    result.writer().fill_zeros();
    Ok(result)
}
```
However, we could also move the logic completely outside of OSTD. In this case, OSTD only needs to provide uninitialized APIs? I'm not sure if that will cause problems.

---

In addition, it seems that the current implementation has some problems interacting with `can_sync_dma`. I suppose you want this:
```rust
        // alloc:
        let (inner, paddr) = if (can_sync_dma() || is_cache_coherent) && !cvm {
            // ...
        } else {
            let (kva, paddr) = alloc_kva(nframes, can_sync_dma() || is_cache_coherent, ZEROED)?;
            // ...
        }

        // sync:
        if self.is_cache_coherent {
            return Ok(());
        }

        let va_range = match &self.inner {
            Inner::Kva(kva, _) => {
                if !can_sync_dma() {
                    // The KVA is mapped as uncachable.
                    return Ok(());
                }
                // ...
            }
```

If `can_sync_dma()` evaluates to `true`, we can use `Writeback` as the cache policy and sync with the device using `sync_dma()`. If `can_sync_dma()` evaluates to `false`, however, we cannot call `sync_dma()` because it will trigger undefined instructions or other ill-defined behaviors.

Hsy-Intel:
Supplementary information for understanding the TDX mechanism:

image

image

ACT: A table in memory which controls memory access. Each bit in the table represents the state of a 4K memory page: shared or private.

taosue:
> TDX replaces the physical page when marking it as shared (`unprotect_gpa_tdvm_call`)... So it doesn't make sense to zero the page beforehand (If implemented as such, `alloc` cannot guarantee that the page is initialized)?
> 

Why can't `alloc` guarantee that the page is initialized? I understand that the page will be zeroed twice (once when allocating segment, and once when invoking `unprotect_gpa_tdvm_call`).

> In addition, it seems that the current implementation has some problems interacting with `can_sync_dma`. I suppose you want this:
...
> If `can_sync_dma()` evaluates to `true`, we can use `Writeback` as the cache policy and sync with the device using `sync_dma()`. If `can_sync_dma()` evaluates to `false`, however, we cannot call `sync_dma()` because it will trigger undefined instructions or other ill-defined behaviors.

Yes, I just realize this issue, thanks for pointing out!!

lrh2000:
> Why can't `alloc` guarantee that the page is initialized? I understand that the page will be zeroed twice (once when allocating segment, and once when invoking `unprotect_gpa_tdvm_call`).

[The spec](https://cdrdv2-public.intel.com/726790/TDX%20Guest-Hypervisor%20Communication%20Interface_1.0_344426_004.pdf) (section 3.2 `TDG.VP.VMCALL<MapGPA>`) does not guarantee that MapGPA will provide us with a zeroed page?

In practice, I guess Linux will almost certainly provide a zeroed page, since that's the easiest way to avoid unintended information leakage. However, I'm not sure if it's really required by the spec.

For example, as a performance optimization, VMM may provide us with non-zeroed, previously used shared pages, which are more cache-friendly and won't leak host information.

taosue:
> [The spec](https://cdrdv2-public.intel.com/726790/TDX%20Guest-Hypervisor%20Communication%20Interface_1.0_344426_004.pdf) (section 3.2 `TDG.VP.VMCALL<MapGPA>`) does not guarantee that MapGPA will provide us with a zeroed page?
> 

Yes, I can't find any description in the spec stating that `TDG.VP.VMCALL<MapGPA>` will erase the page content. However, the TDX module spec does describe that `TDG.MEM.PAGE.ACCEPT` will zero out all content. I.e. converting shared memory to private memory will zero all content, but converting private to shared will not according to the spec.

I quickly checked the KVM/QEMU code and didn't find that the VMM zeros the content either. Perhaps the comment https://github.com/asterinas/asterinas/blob/922b204280440203d0e5f9d6a06040df64d1e7e5/ostd/src/arch/x86/tdx_guest.rs#L17-L21 is not correct?  @Hsy-Intel, could you help check?

lrh2000:
> erase the page content

To me, erasing old data does not mean providing zeros as the new data. `MapGPA` provides us with a completely new physical page. The old data disappears, but the new data may not be zeros. However, the new data is certainly unrelated to the old data (otherwise, our data will leak to the host), so the old data is still "erased".

It's mostly just a text trick, though.

**EDIT:** Note that `MapGPA` cannot guarantee that the new page is a zeroed page. There is a real-world use case that contradicts the guarantee: the new page may be an MMIO page. For example,
https://github.com/asterinas/asterinas/blob/922b204280440203d0e5f9d6a06040df64d1e7e5/ostd/src/io/io_mem/mod.rs#L106

lrh2000:
> We may want to reconsider the DMA allocation APIs based on that (including `DmaCoherent::alloc`, which is not touched by this PR but it suffers from the same problem).
>
> Perhaps we can do this:
> 
> ```rust
> fn alloc() {
>     let result = Self::alloc_uninitialized()?;
>     result.writer().fill_zeros();
>     Ok(result)
> }
> ```

I also realized that there is another argument for filling zeros after `alloc_uninitialized()`. This example is completely unrelated to the TDX scenario.

Suppose someone writes code like this:
```rust
let dma_stream = DmaStream::alloc();
dma_stream.sync_for_device();
tell_device_to_start_read_at!(dma_stream.daddr());
```

Then, the device must see zeros. However, consider the uncacheable KVA case (`Inner::Kva(..)`) with `can_sync_dma() == false`. `sync_for_device()` is a no-op. If we write zeros via the linear mapping (instead of via the KVA), the zeros may remain in the cache. Therefore, the device will see incorrect data. It's wrong.

taosue:
> Then, the device must see zeros. However, consider the uncacheable KVA case (`Inner::Kva(..)`) with `can_sync_dma() == false`. `sync_for_device()` is a no-op. If we write zeros via the linear mapping (instead of via the KVA), the zeros may remain in the cache. Therefore, the device will see incorrect data. It's wrong.

Ok, so we only can zero content via KVA. Thanks for the detailed argument and I will fill zeros after invoking `alloc_uninit()`.

Hsy-Intel:
> EDIT: Note that MapGPA cannot guarantee that the new page is a zeroed page. There is a real-world use case that contradicts the guarantee: the new page may be an MMIO page.

@lrh2000 In fact, MMIO differs from regular memory in TD. From TDVF spec:

image


From TDX module base spec:

> The VMM may assign devices directly to the guest TD. The addresses mapping the MMIO resources of such devices must
be mapped in the shared memory space of the TD. When submitting data buffers to these devices, the guest TD must
locate the data buffers in shared memory such that the directly assigned device can move data in/out of such buffers
using DMA. The data buffers placed in shared memory should be programmed in IOMMU page tables.
The SRIOV virtual function devices assigned to guest TD also follow the DDA guidelines stated above with respect to
MMIO and data buffers. The control plane of the virtual function would use the soft or hard mechanism to configure the
virtual functions:
• The soft mechanism would use para-virtualization to configure the virtual function.
• The hard mechanism would use hardware mailboxes accessed using MMIO in the shared memory region.

> I quickly checked the KVM/QEMU code and didn't find that the VMM zeros the content either. Perhaps the comment is not correct?

@taosue Based on my understanding, the page erase operation in the private to shared conversion is performed by the host-side `TDH.MEM.PAGE.REMOVE` function: please refer to step 6 in the image I included in my previous reply. Therefore, this comment might not be entirely accurate, but I think valid content was indeed removed during the conversion.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,61 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+        let (inner, paddr_range) = if (can_sync_dma() || is_cache_coherent) && !cvm {
+            let segment: USegment = FrameAllocOptions::new()
+                .zeroed(false)
+                .alloc_segment(nframes)?
+                .into();
+            let paddr_range = segment.paddr_range();
+
+            (Inner::Segment(segment), paddr_range)
+        } else {
+            let (kva, paddr) = alloc_kva(nframes, can_sync_dma() || is_cache_coherent)?;
```
tatetian, line 137: `alloc_kva` internally allocates a zeroed segment. In this case, `DmaStream::alloc` will zero out the DMA segment _twice_. To resolve this redundancy, extends `alloc_kva` with an extra argument to specify if a newly-allocated KVA is zeroed.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,61 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+    /// # Comparison with [`DmaStream::map`]
+    ///
+    /// This method is semantically equivalent to allocating a [`USegment`] via
+    /// [`FrameAllocOptions::alloc_segment`] and then mapping it with
+    /// [`DmaStream::map`]. However, [`DmaStream::alloc`] combines these two
+    /// operations and can be more efficient in certain scenarios, particularly
+    /// on TDX which avoids the overhead of bounce buffers.
+    pub fn alloc(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
+        Self::alloc_uninit(nframes, is_cache_coherent).and_then(|dma| {
+            dma.writer()?.fill_zeros(dma.size());
```
tatetian, line 115: Note that the `writer` method will return an error if `D` is `FromDevice`. This is because we do not allow `DmaStream<FromDevice>` to obtain a writer:

https://github.com/asterinas/asterinas/blob/922b204280440203d0e5f9d6a06040df64d1e7e5/ostd/src/mm/dma/dma_stream.rs#L311-L314

This means that `DmaStream::<FromDevice>::alloc()` always fails.

This compels me to search the codebase and see if there are indeed no users of `DmaStream::<FromDevice>::alloc()`. From this PR, it appears that there must be no such users, otherwise the CI would fail.

https://github.com/asterinas/asterinas/blob/922b204280440203d0e5f9d6a06040df64d1e7e5/kernel/comps/network/src/dma_pool.rs#L145-L149

But I found out that the network stack is using `DmaStream<FrameDevice>` for `RxBuffer`.  And the use pattern in `DmaPool` does allow us to replace `DmaStream::map` with `DmaStream::alloc`.

@lrh2000 Does `RxBuffer` rely on buffers being initialized to zeros beforehand?

If not, we should replace the use of `DmaStream::<FromDevice>::map` in `DmaPool<FromDevice>` with `DmaStream::<FromDevice>::alloc_uninit`, with comments to stress why non-zeroed buffers are ok.

In addition, we need to update the Rust doc of `DmaStream::alloc` to stress two points:
1. The memory of a newly-allocated DMA buffer from this method is initialized to zeros. Currently, this point is missing in the Rust doc.
2. The zeroing `DmaStream::alloc` method is permitted only when `D` is `ToDevice` or `FromAndToDevice` (not `FromDevice`). This restriction is aligned with the fact that `DmaStream::writer` is permitted only when `D` is `ToDevice` or `FromAndToDevice`.

I am not sure if `DmaStream::<FromDevice>::alloc` has any legit use case. If so, we will have to make it work.

ostd/src/mm/dma/dma_coherent.rs
```diff
@@ -41,10 +41,20 @@ impl DmaCoherent {
     /// that the DMA mapping is prepared for can access the main memory in a
     /// CPU cache coherent way or not.
     pub fn alloc(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
+        Self::alloc_uninit(nframes, is_cache_coherent).inspect(|dma| {
```
tatetian, line 46: My comments for `DmaStream` also apply to `DmaCoherent`.

ostd/src/mm/dma/dma_stream.rs
taosue: I assume `alloc_kva` doesn't need to zero the segment, so I added `zeroed(false)` to `alloc_kva`. Is it necessary to add the `zeroed` argument?

ostd/src/mm/dma/dma_stream.rs
lrh2000: > And the use pattern in `DmaPool` does allow us to replace `DmaStream::map` with `DmaStream::alloc`.

I tend not to replace it. Direct access to the segment enables the network stack to parse and generate network packets more flexibly. This unified approach works for virtual interfaces, which should not have a DMA stream. This is similar to what was previously mentioned at https://github.com/asterinas/asterinas/pull/2351#issuecomment-3538858324.

Otherwise, we'll have to create an `enum` between `Segment` and `DmaStream`, and every access to the packet will need to be dispatched to a concrete enum variant. Even worse, under the current design, we cannot modify the packet because `DmaStream` disallows writing if `!D::CAN_WRITE_TO_DEVICE`. However, in some scenarios (e.g., eBPF programs), an incoming packet can be modified before being sent to the network stack.

It's a more complex story. I'd prefer to leave the code as it is. There are still too many TODOs in our network stack.

> I am not sure if `DmaStream::<FromDevice>::alloc` has any legit use case. If so, we will have to make it work.

If a device fills a buffer only partially, we want the remaining part to be zero-padded. In that case, we could use `DmaStream::<FromDevice>::alloc` to zero the entire buffer in advance. Does this seem like a use case?

ostd/src/mm/dma/dma_coherent.rs
```diff
@@ -37,14 +37,26 @@ enum Inner {
...
     /// The `is_cache_coherent` argument specifies whether the target device
     /// that the DMA mapping is prepared for can access the main memory in a
     /// CPU cache coherent way or not.
     pub fn alloc(nframes: usize, is_cache_coherent: bool) -> Result<Self, Error> {
+        Self::alloc_uninit(nframes, is_cache_coherent).inspect(|dma| {
+            dma.writer().fill_zeros(dma.size());
+        })
+    }
+
+    /// This method is same as [`DmaCoherent::alloc`]
```
lrh2000: Please keep the [summary sentence](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#summary-sentence).

```rust
    /// Allocates a region of physical memory for coherent DMA access
    /// without initialization.
    ///
    /// This method is same as [`DmaCoherent::alloc`] ......
```

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,65 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+            Ok(dma)
+        })
+    }
+
+    /// This method is same as [`DmaStream::alloc`]
```
lrh2000: Same here.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,65 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+    Kva(KVirtArea, Paddr),
     Both(KVirtArea, Paddr, USegment),
 }
 
 impl<D: DmaDirection> DmaStream<D> {
+    /// Allocates a region of physical memory for streaming DMA access.
+    ///
+    /// The memory of the newly-allocated DMA buffer is initialized to zeros.
+    /// This method is only available when `D` is [`ToDevice`] or
+    /// [`FromAndToDevice`], as zeroing requires write access to the buffer.
```
lrh2000, line 97: Like `sync_to_device`, you can add a const assertion here?
```rust
        const { assert!(D::CAN_WRITE_TO_DEVICE) };
```

ostd/src/mm/dma/dma_coherent.rs
lrh2000: Also, it seems that we'll have to write `is the same as` to be correct in English grammar?

ostd/src/mm/dma/dma_stream.rs
tatetian: > Direct access to the segment enables the network stack to parse and generate network packets more flexibly.

Got it. Thanks for explaining.

> Does this seem like a use case?

In theory, yes. But it sounds a bit hyperthetical.

I am thinking introducing `force_writer` and `force_reader` methods, which allows the user to bypass the type-level constraints, providing flexibility when it is needed. But I guess that this can be done in a separate PR. So for now, I think it is sufficient to improve the Rust doc.

ostd/src/mm/dma/dma_stream.rs
```diff
@@ -85,10 +85,70 @@ pub struct DmaStream<D: DmaDirection = FromAndToDevice> {
...
+    /// that the DMA mapping is prepared for can access the main memory in a
+    /// CPU cache coherent way or not.
+    ///
+    /// # Comparison with [`DmaStream::map`]
+    ///
+    /// This method is semantically equivalent to allocating a [`USegment`] via
+    /// [`FrameAllocOptions::alloc_segment`] and then mapping it with
+    /// [`DmaStream::map`]. However, [`DmaStream::alloc`] combines these two
+    /// operations and can be more efficient in certain scenarios, particularly
+    /// on TDX which avoids the overhead of bounce buffers.
```
lrh2000: According to https://github.com/asterinas/asterinas/pull/2351#discussion_r2649734164, maybe we don't want to mention TDX here?
```suggestion
    /// operations and can be more efficient in certain scenarios, particularly
    /// in confidential VMs, where the overhead of bounce buffers can be
    /// avoided.
```

ostd/src/mm/dma/dma_stream.rs
lrh2000: > I am thinking introducing `force_writer` and `force_reader` methods, which allows the user to bypass the type-level constraints, providing flexibility when it is needed. But I guess that this can be done in a separate PR. So for now, I think it is sufficient to improve the Rust doc.

Yes. Agree.

lrh2000 (approved):
Otherwise the changes seem good to me. Thanks for your contribution!

Issue #2807, opened 2025-12-27
Support sparse files in Ext2


LordaeronESZ:
### Feature Description

The current Ext2 file system does not appear to support sparse files, which are files whose logical address space contains holes with unallocated physical disk blocks.

### Problem or Need

Specifically, many current Ext2 kernel codes do not account for sparse files, which can cause unexpected panics when Ext2 processes such files. Here is an example:
https://github.com/asterinas/asterinas/blob/c4e8bf2c21aa8ad4b272dbc1d68461fa3b42004e/kernel/src/fs/ext2/inode.rs#L2218-L2233
In sparse files case, `size_to_blocks` is not actual number of blocks utilized. Well, it is the root cause of panic in #2776 .

### Suggested Implementation

Here are some code logic changes I currently think of that need to be made, this list may not be exhaustive.

- [ ] Remove unreasonable assertions.
- [ ] Fixes issues caused by deleting sparse files.
- [ ] Fixes issues caused by copying sparse files.
- [ ] Reading sparse file holes will return 0 directly.
- [ ] Writing to sparse file holes will dynamically allocate disk blocks.
- [ ] Supports `fallocate`'s `PunchHoleKeepSize`(Current implementation just fill zeros, which will occupy disk spaces).
- [ ] Supports sparse hole creation with `lseek`.
- [ ] ...

tatetian:
Good job on identifying the root cause.

Is the term "sparse file" invented by you or coined in existing literature?

This term sounds weird to me because by your definition, a 10TB file having a 4KB hole would be called "sparse". If no existing literature has coined a term, then calling them "holed files" sounds good to me.

LordaeronESZ:
> Is the term "sparse file" invented by you or coined in existing literature?

This term is mentioned in many entries in the man page, such as `man 1 tar`:
``` man
...
--hole-detection=METHOD
      Use METHOD to detect holes in sparse files.  This option implies --sparse.  Valid values for METHOD are seek and raw.  De‐
      fault is seek with fallback to raw when not applicable.
...
```
In addition, the term can be found on [Wikipedia](https://en.wikipedia.org/wiki/Sparse_file) and other sources.

> This term sounds weird to me because by your definition, a 10TB file having a 4KB hole would be called "sparse". If no existing literature has coined a term, then calling them "holed files" sounds good to me.

By definition, as in your example, a file containing even a single, very small hole can indeed be called a "sparse file," even if its sparsity is not high.

[...]

PR #2853, opened 2026-01-06
Fix the issue of `InodeBlockManager::read_blocks`

### LINKED ISSUE #2835: Ext2 O_DIRECT read cannot return data from disk
Ext2's [`Inode::read_direct_at`](https://github.com/asterinas/asterinas/blob/eb76ded5fc4f56b6ebc469728acbc87bf56aec35/kernel/src/fs/ext2/inode.rs#L709) is implemented with [`InodeInner::read_direct_at`](https://github.com/asterinas/asterinas/blob/eb76ded5fc4f56b6ebc469728acbc87bf56aec35/kernel/src/fs/ext2/inode.rs#L1011), which in turns calls [`InodeBlockManager::read_blocks`](https://github.com/asterinas/asterinas/blob/eb76ded5fc4f56b6ebc469728acbc87bf56aec35/kernel/src/fs/ext2/inode.rs#L1884). And this `read_blocks` method is a simple wrapper around the async version:

https://github.com/asterinas/asterinas/blob/eb76ded5fc4f56b6ebc469728acbc87bf56aec35/kernel/src/fs/ext2/inode.rs#L1859-L1882

But there is a silly mistake in this `read_blocks_async` method. This line

```rust
bio_segment.reader().unwrap().read_fallible(writer)?; 
```

should be executed _after_ the bio request is complete (rather than before).

LordaeronESZ:
Closes #2835 

This PR resolves the issue aforementioned, and I wrote a simple C program with the help of LLM to test it:
``` c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <fcntl.h>
#include <errno.h>
#include <sys/stat.h>
#include <sys/types.h>

#define BLOCK_SIZE 4096
#define TEST_PATTERN 0xAA

/* Fill buffer with known pattern */
static void fill_pattern(void *buf, size_t size, unsigned char pattern)
{
    memset(buf, pattern, size);
}

/* Verify buffer contains the expected pattern */
static int verify_pattern(const void *buf, size_t size, unsigned char pattern)
{
    const unsigned char *ptr = buf;
    for (size_t i = 0; i < size; i++) {
        if (ptr[i] != pattern) {
            fprintf(stderr, "Pattern mismatch at offset %zu: expected 0x%02X, got 0x%02X\n",
                    i, pattern, ptr[i]);
            return -1;
        }
    }
    return 0;
}

/* Print hex dump of buffer for debugging */
static void hexdump(const void *buf, size_t size, const char *prefix)
{
    const unsigned char *ptr = buf;
    size_t i;

    printf("%s (first %zu bytes):\n", prefix, size < 64 ? size : 64);
    for (i = 0; i < size && i < 64; i++) {
        if (i % 16 == 0) {
            printf("%s%04zx: ", prefix ? "  " : "", i);
        }
        printf("%02x ", ptr[i]);
        if (i % 16 == 15) {
            printf("\n");
        }
    }
    if (i % 16 != 0) {
        printf("\n");
    }
}

static int test_basic_oadirect(const char *filename)
{
    int fd;
    void *write_buf = NULL;
    void *read_buf = NULL;
    int ret = -1;

    printf("=== Test 1: Basic O_DIRECT Read ===\n");

    /* Allocate aligned buffers for O_DIRECT */
    if (posix_memalign(&write_buf, BLOCK_SIZE, BLOCK_SIZE) != 0) {
        perror("posix_memalign for write_buf");
        goto out;
    }
    if (posix_memalign(&read_buf, BLOCK_SIZE, BLOCK_SIZE) != 0) {
        perror("posix_memalign for read_buf");
        goto out;
    }

    /* Open file with O_DIRECT */
    fd = open(filename, O_RDWR | O_DIRECT | O_CREAT, 0644);
    if (fd < 0) {
        /* O_DIRECT might not be supported, try without it */
        printf("Warning: O_DIRECT not supported, testing without it\n");
        fd = open(filename, O_RDWR | O_CREAT, 0644);
        if (fd < 0) {
            perror("open");
            goto out;
        }
    }

    /* Write known pattern */
    fill_pattern(write_buf, BLOCK_SIZE, TEST_PATTERN);
    if (write(fd, write_buf, BLOCK_SIZE) != BLOCK_SIZE) {
        perror("write");
        close(fd);
        goto out;
    }
    fsync(fd);

    /* Reset file position */
    lseek(fd, 0, SEEK_SET);

    /* Clear read buffer to detect uninitialized reads */
    memset(read_buf, 0x00, BLOCK_SIZE);

    /* Read back the data */
    ssize_t nread = read(fd, read_buf, BLOCK_SIZE);
    if (nread != BLOCK_SIZE) {
        perror("read");
        close(fd);
        goto out;
    }

    /* Verify the pattern */
    if (verify_pattern(read_buf, BLOCK_SIZE, TEST_PATTERN) != 0) {
        fprintf(stderr, "FAIL: Read data does not match written pattern!\n");
        hexdump(read_buf, BLOCK_SIZE, "Read data");
        hexdump(write_buf, BLOCK_SIZE, "Expected data");
        close(fd);
        goto out;
    }

    printf("PASS: O_DIRECT read test succeeded\n");
    close(fd);
    ret = 0;

out:
    free(write_buf);
    free(read_buf);
    return ret;
}

int main(int argc, char *argv[])
{
    const char *filename;
    int failed = 0;

    if (argc < 2) {
        fprintf(stderr, "Usage: %s <test_file_path>\n", argv[0]);
        return 1;
    }

    filename = argv[1];

    /* Run all tests */
    if (test_basic_oadirect(filename) != 0) {
        failed++;
    }

    if (failed == 0) {
        printf("ALL TESTS PASSED\n");
    } else {
        printf("%d TEST(S) FAILED\n", failed);
    }

    /* Clean up test file */
    unlink(filename);

    return failed > 0 ? 1 : 0;
}

```

- Before fix:
issue

- After fix: 
fix

### DIFF
--- a/kernel/comps/block/src/bio.rs
+++ b/kernel/comps/block/src/bio.rs
@@ -204,6 +204,11 @@ impl BioWaiter {
         self.bios.append(&mut other.bios);
     }
 
+    /// Returns an iterator for the `Bio` requests associated with `self`.
+    pub fn reqs(&self) -> impl Iterator<Item = Bio> {
+        self.bios.iter().map(|bio_inner| Bio(bio_inner.clone()))
+    }
+
     /// Waits for the completion of all `Bio` requests.
     ///
     /// This method iterates through each `Bio` in the list, waiting for their
--- a/kernel/src/fs/ext2/inode.rs
+++ b/kernel/src/fs/ext2/inode.rs
@@ -1817,12 +1817,6 @@ impl InodeImpl {
 
 #[inherit_methods(from = "self.block_manager")]
 impl InodeImpl {
-    pub fn read_blocks_async(
-        &self,
-        bid: Ext2Bid,
-        nblocks: usize,
-        writer: &mut VmWriter,
-    ) -> Result<BioWaiter>;
     pub fn read_blocks(&self, bid: Ext2Bid, nblocks: usize, writer: &mut VmWriter) -> Result<()>;
     pub fn read_block_async(&self, bid: Ext2Bid, frame: &CachePage) -> Result<BioWaiter>;
     pub fn write_blocks_async(
@@ -1850,13 +1844,7 @@ struct InodeBlockManager {
 
 impl InodeBlockManager {
     /// Reads one or multiple blocks to the segment start from `bid` asynchronously.
-    pub fn read_blocks_async(
-        &self,
-        bid: Ext2Bid,
-        nblocks: usize,
-        writer: &mut VmWriter,
-    ) -> Result<BioWaiter> {
-        debug_assert!(nblocks * BLOCK_SIZE <= writer.avail());
+    fn read_blocks_async(&self, bid: Ext2Bid, nblocks: usize) -> Result<BioWaiter> {
         let mut bio_waiter = BioWaiter::new();
 
         for dev_range in DeviceRangeReader::new(self, bid..bid + nblocks as Ext2Bid)? {
@@ -1864,8 +1852,6 @@ impl InodeBlockManager {
             let range_nblocks = dev_range.len();
 
             let bio_segment = BioSegment::alloc(range_nblocks, BioDirection::FromDevice);
-            bio_segment.reader().unwrap().read_fallible(writer)?;
-
             let waiter = self.fs().read_blocks_async(start_bid, bio_segment)?;
             bio_waiter.concat(waiter);
         }
@@ -1874,10 +1860,22 @@ impl InodeBlockManager {
     }
 
     pub fn read_blocks(&self, bid: Ext2Bid, nblocks: usize, writer: &mut VmWriter) -> Result<()> {
-        match self.read_blocks_async(bid, nblocks, writer)?.wait() {
-            Some(BioStatus::Complete) => Ok(()),
-            _ => return_errno!(Errno::EIO),
+        debug_assert!(nblocks * BLOCK_SIZE <= writer.avail());
+        let bio_waiter = self.read_blocks_async(bid, nblocks)?;
+        if Some(BioStatus::Complete) != bio_waiter.wait() {
+            return_errno!(Errno::EIO);
         }
+
+        for bio in bio_waiter.reqs() {
+            for segment in bio.segments() {
+                segment
+                    .reader()?
+                    .read_fallible(writer)
+                    .map_err(|(e, _)| Error::from(e))?;
+            }
+        }
+
+        Ok(())
     }
 
     pub fn read_block_async(&self, bid: Ext2Bid, frame: &CachePage) -> Result<BioWaiter> {


kernel/src/fs/ext2/inode.rs
```diff
@@ -1857,33 +1857,53 @@ struct InodeBlockManager {
...
-            let waiter = self.fs().read_blocks_async(start_bid, bio_segment)?;
+            segments.push(bio_segment);
             bio_waiter.concat(waiter);
         }
 
+        Ok((bio_waiter, segments))
+    }
+
+    /// Reads one or multiple blocks to the segment start from `bid` asynchronously.
+    pub fn read_blocks_async(
```
tatetian: When fixing this bug, you must have also realized that the current signature of `read_blocks_async` is impossible. That's why you have added a new method called `read_blocks_internal`, which returns `Vec<BioSegment>` in addition to `BioWaiter`.

For this reason, we should either 
1. Remove the `read_blocks_async` API, or
2. Expose a revised `read_blocks_async` API (which is actually the so-called `read_blocks_internal`), or
3. Extend `BioWaiter` so that `BioWaiter::wait` can automatically do the data copy.

We should analyze the pros and cons of these options (any more?), pick the best option, and execute it. Keeping an insane API is not an option.

tatetian (commented):
Thanks for fixing the bug!

LordaeronESZ:
I have revised my code based on 3rd options you have given.

In this revised version, I retained the original `BioWaiter::wait` method and introduced a `BioWaiter::wait_and_copy_to` method, which can conveniently complete the copying of `BioSegment`. This allows the caller to choose to complete other tasks after calling `read_blocks_async`, and then call `wait_and_copy_to` to perform the copy.

kernel/comps/block/src/bio.rs
```diff
@@ -232,6 +232,28 @@ impl BioWaiter {
         ret
     }
 
+    /// Waits for completion and copies data to the specified writer.
+    ///
+    /// This is a convenience method for the common case of copying data from
+    /// completed I/O operations to a `VmWriter`.
+    pub fn wait_and_copy_to(&self, writer: &mut VmWriter) -> Result<(), Error> {
```
tatetian: This method is sufficient to implement the required functionality, but its coupling with the target use case (copying back the data of a completed `BioWaiter`) is too strong. `BioWaiter::concat` allows one to merge multiple `BioWaiter` into one. This way, a single invocation of `BioWaiter::wait` can synchronize the processing of multiple BIO requests. These requests could be _unrelated_, _noncontiguous_, and _heterogeneous_.

On the other hand, `wait_and_copy_to` has a strong assumption about the BIO requests monitored by this `BioWaiter`: they must be read requests (otherwise, copyback woudn't make sense) and their target physical blocks must be contiguous. So this `wait_and_copy_to` method would not make a great API.

Inspired by your solution, I think a better API design is to add a `bios` method:

```rust
impl BioWaiter {
    /// Returns an iterator for the `Bio` requests associated with `self`.
    pub fn bios(&self) -> impl Iterator<Item = &'a Bio> { ... }
}
```

This API is more self-contained, composable, and flexible. The user code would then looks like the following:


```rust
    pub fn read_blocks(&self, bid: Ext2Bid, nblocks: usize, writer: &mut VmWriter) -> Result<()> {
        let bio_waiter = self.read_blocks_async(bid, nblocks, writer)?;
        if Some(BioStatus::Complete) != bio_waiter.wait() {
            return_errno!(Errno::EIO);
        }

        let bios = bio_waiter.bios();
        todo!("copy back the data by using bios");
     }
```

LordaeronESZ:
`wait_and_copy_to` is indeed not a good interface; I have revised the code according to your suggestion.

LordaeronESZ:
Support for sparse files (#2807 ) is related to DirectIO, so set this PR as a draft for now.

kernel/src/fs/ext2/inode.rs
tatetian: This fix raises a different question: the `writer` parameter is still in the signature of `read_blocks_async` yet is now completely unused. In other words, the API is now nonsensical (it takes a `writer` it never writes to), and should be cleaned up.

I am wondering if `read_blocks_async` has any users outside this module. If not, you can simply make it private because the API is not easy to use correctly. If yes, add Rust doc for it to describe how to use the API correctly.

kernel/src/fs/ext2/inode.rs
```diff
@@ -1874,10 +1872,21 @@ impl InodeBlockManager {
...
+        if Some(BioStatus::Complete) != bio_waiter.wait() {
+            return_errno!(Errno::EIO);
+        }
+
+        for bio in bio_waiter.bios() {
+            for segment in bio.segments() {
+                segment
+                    .reader()?
+                    .read_fallible(writer)
+                    .map_err(|(e, _)| Error::new(Errno::EIO))?;
```
tatetian: The `read_fallible` may return `PageFault`.

https://github.com/asterinas/asterinas/blob/c9032ad97cbc002afcc5eab41c837a850fb70aa0/ostd/src/mm/io.rs#L400-L404

 In this case, returning `EIO` would be misleading.

kernel/comps/block/src/bio.rs
```diff
@@ -204,6 +204,11 @@ impl BioWaiter {
         self.bios.append(&mut other.bios);
     }
 
+    /// Returns an iterator for the `Bio` requests associated with `self`.
+    pub fn bios(&self) -> impl Iterator<Item = Bio> {
+        self.bios.iter().map(|bio_inner| Bio(bio_inner.clone()))
```
tatetian, line 209: For consistency, rename this method to `reqs` since we already have `req` and `nreqs` method.

tatetian (commented):
There are still a few minor issues.

tatetian (approved):

PR #2449, opened 2025-09-22
Support accessing user spaces of other processes

vvvvsv:
This PR enables accessing the user spaces of other processes, i.e., reading from or writing to a `Vmar` whose corresponding `VmSpace` is not activated on the current CPU.

This memory access interface behaves the same as directly activating the corresponding `VmSpace` and accessing the memory in it. For example, if the target page has been mapped but not yet loaded into memory, a page fault will be triggered and handled to map the page. If a COW page needs to be written, the COW operation will be correctly performed before writing.

The first commit introduces the ability to access user spaces of other processes.

The second commit supports reading a process's initial stack across multiple pages, which is required by `/proc/[pid]/cmdline` and `/proc/[pid]/environ`. Fixes https://github.com/asterinas/asterinas/issues/2249. This is required for https://github.com/asterinas/asterinas/issues/2112.

The third commit adds `/proc/[pid]/mem` and `/proc/[pid]/task/[tid]/mem` without force-write, which is required to support GDB https://github.com/asterinas/asterinas/issues/2323. 

The fourth commit adds a regression test for `/proc/[pid]/mem`, since there is no such test in ltp or gvisor.

---

The third commit (`/proc/[pid]/mem`) was once discussed in https://github.com/asterinas/asterinas/pull/2324#discussion_r2292603292, and some unresolved points from the discussion are addressed in this commit. 

> There are quite some corner cases when implementing this. Please carefully consider what will happen if:
> - short read happens (e.g., the first page exists but the second page does not exist);

If short read happens, the `read` syscall should return the number of bytes already read.

> - the page has been mapped but the page hasn't been loaded into the memory;

A page fault should be triggered and handled to map the page.

> - the page cannot be written (e.g., it is mapped as read-only)

On Linux, it is possible to write to a read-only VMA with `/proc/[pid]/mem` ("force-write"). Reference Linux implementation: [1](https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/base.c#L919-L921) and [2](https://elixir.bootlin.com/linux/v6.16.5/source/mm/gup.c#L1282-L1301). In fact, this is equivalent to using `ptrace` with the `PTRACE_POKEDATA` request. If this read-only VMA is a copy-on-write VMA (i.e., !VM_SHARED and VM_MAYWRITE), a COW operation will be performed and the write will go to the newly copied page; otherwise, the write will fail.

[...]

PR #2855, opened 2026-01-07
Support force-write via `/proc/pid/mem`

vvvvsv:
Follow-up work of https://github.com/asterinas/asterinas/pull/2449#issuecomment-3322079238.

"Force-write" means that a write to `/proc/pid/mem` will trigger COW and then succeed even if the corresponding memory region does not have `PROT_WRITE`.

About access checking: Linux performs special access checks for `/proc/pid/mem` in two places.
1. On [open](https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/base.c#L830-L857). This is more strict than `rw-------`. This check is closely related to ptrace. Seems better introduce ptrace access checks first and then add ptrace? 
2.  On [read and write](https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/base.c#L877-L897). Linux has a configuration option that controls when force read/write is allowed. By default, it is `CONFIG_PROC_MEM_ALWAYS_FORCE`. This PR simply implements the default behavior.
```
vvsv@vvsv-XiaoXin:~/LINUX/linux-6.15$ grep PROC_MEM .config
CONFIG_PROC_MEM_ALWAYS_FORCE=y
# CONFIG_PROC_MEM_FORCE_PTRACE is not set
# CONFIG_PROC_MEM_NO_FORCE is not set
```

Minor: The RSS counter should not be incremented when a COW occurs. My mistake.

### DIFF
--- a/kernel/src/fs/procfs/pid/task/mem.rs
+++ b/kernel/src/fs/procfs/pid/task/mem.rs
@@ -81,7 +81,7 @@ impl InodeIo for MemFileHandle {
             return Ok(0);
         };
 
-        match vmar.read_remote(offset, writer) {
+        match vmar.read_alien(offset, writer) {
             Ok(bytes) => Ok(bytes),
             Err((_, 0)) => Err(Error::new(Errno::EIO)),
             Err((_, bytes)) => Ok(bytes),
@@ -104,7 +104,7 @@ impl InodeIo for MemFileHandle {
             return Ok(0);
         };
 
-        match vmar.write_remote(offset, reader) {
+        match vmar.write_alien(offset, reader) {
             Ok(bytes) => Ok(bytes),
             Err((_, 0)) => Err(Error::new(Errno::EIO)),
             Err((_, bytes)) => Ok(bytes),
--- a/kernel/src/vm/vmar/vmar_impls/access_alien.rs
+++ b/kernel/src/vm/vmar/vmar_impls/access_alien.rs
@@ -1,5 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! This module enables access to a VMAR
+//! from an [_alien thread_] (crate::process::posix_thread::alien_access),
+//! which is a non-member of the owner process of the VMAR.
+
 use align_ext::AlignExt;
 use ostd::{
     mm::{PAGE_SIZE, PageFlags, UFrame, io_util::HasVmReaderWriter, vm_space::VmQueriedItem},
@@ -13,7 +17,7 @@ use crate::{
 };
 
 impl Vmar {
-    /// Reads memory from the process user space.
+    /// Reads memory in the context of an alien thread.
     ///
     /// This method reads until one of the conditions is met:
     /// 1. The writer has no available space.
@@ -23,7 +27,7 @@ impl Vmar {
     /// On error, both the error and the number of bytes read so far are returned.
     ///
     /// The `VmSpace` of the process is not required to be activated on the current CPU.
-    pub fn read_remote(
+    pub fn read_alien(
         &self,
         vaddr: Vaddr,
         writer: &mut VmWriter,
@@ -35,10 +39,10 @@ impl Vmar {
             reader.read_fallible(writer)
         };
 
-        self.access_remote(vaddr, len, PageFlags::R, read)
+        self.access_alien(vaddr, len, PageFlags::R, read)
     }
 
-    /// Writes memory to the process user space.
+    /// Writes memory in the context of an alien thread.
     ///
     /// This method writes until one of the conditions is met:
     /// 1. The reader has no remaining data.
@@ -48,7 +52,7 @@ impl Vmar {
     /// On error, both the error and the number of bytes written so far are returned.
     ///
     /// The `VmSpace` of the process is not required to be activated on the current CPU.
-    pub fn write_remote(
+    pub fn write_alien(
         &self,
         vaddr: Vaddr,
         reader: &mut VmReader,
@@ -60,17 +64,17 @@ impl Vmar {
             writer.write_fallible(reader)
         };
 
-        self.access_remote(vaddr, len, PageFlags::W, write)
+        self.access_alien(vaddr, len, PageFlags::W, write)
     }
 
-    /// Writes zeros to the process user space.
+    /// Writes zeros in the context of an alien thread.
     ///
     /// This method writes at most `len` bytes of zeros to the process user space.
     /// On success, the number of bytes written is returned; on error, both the
     /// error and the number of bytes written so far are returned.
     ///
     /// The `VmSpace` of the process is not required to be activated on the current CPU.
-    pub fn fill_zeros_remote(
+    pub fn fill_zeros_alien(
         &self,
         vaddr: Vaddr,
         len: usize,
@@ -84,15 +88,15 @@ impl Vmar {
             Ok(res)
         };
 
-        self.access_remote(vaddr, len, PageFlags::W, write)
+        self.access_alien(vaddr, len, PageFlags::W, write)
     }
 
-    /// Accesses memory at `vaddr..vaddr+len` within the process user space using `op`.
+    /// Accesses memory at `vaddr..vaddr+len` in the context of an alien thread using `op`.
     ///
     /// The `VmSpace` of the process is not required to be activated on the current CPU.
     /// If any page in the range is not mapped or does not have the required page
     /// flags, a page fault will be handled to try to make the page accessible.
-    fn access_remote<F>(
+    fn access_alien<F>(
         &self,
         vaddr: Vaddr,
         len: usize,
@@ -158,7 +162,7 @@ impl Vmar {
                         VmQueriedItem::MappedIoMem { .. } => {
                             return_errno_with_message!(
                                 Errno::EOPNOTSUPP,
-                                "accessing remote MMIO memory is not supported currently"
+                                "accessing alien MMIO memory is not supported currently"
                             );
                         }
                     }
--- a/kernel/src/vm/vmar/vmar_impls/mod.rs
+++ b/kernel/src/vm/vmar/vmar_impls/mod.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-mod access_remote;
+mod access_alien;
 mod fork;
 pub(super) mod map;
 pub(super) mod page_fault;
--- a/kernel/src/process/process_vm/init_stack/mod.rs
+++ b/kernel/src/process/process_vm/init_stack/mod.rs
@@ -415,7 +415,7 @@ impl InitStackReader<'_> {
 
         let read_at = self.argv_range.start + offset;
         writer.limit(self.argv_range.end - read_at);
-        let bytes_read = self.vmar.read_remote(read_at, writer)?;
+        let bytes_read = self.vmar.read_alien(read_at, writer)?;
 
         Ok(bytes_read)
     }
@@ -428,7 +428,7 @@ impl InitStackReader<'_> {
 
         let read_at = self.envp_range.start + offset;
         writer.limit(self.envp_range.end - read_at);
-        let bytes_read = self.vmar.read_remote(read_at, writer)?;
+        let bytes_read = self.vmar.read_alien(read_at, writer)?;
 
         Ok(bytes_read)
     }
--- a/kernel/src/process/program_loader/elf/load_elf.rs
+++ b/kernel/src/process/program_loader/elf/load_elf.rs
@@ -399,7 +399,7 @@ fn map_segment_vmo(
         let vaddr_to_zero = map_addr + (file_range.end - segment_offset);
         let size_to_zero = map_addr + segment_size - vaddr_to_zero;
         if size_to_zero != 0 {
-            let res = vmar.fill_zeros_remote(vaddr_to_zero, size_to_zero);
+            let res = vmar.fill_zeros_alien(vaddr_to_zero, size_to_zero);
             if let Err((err, _)) = res
                 && perms.contains(VmPerms::WRITE)
             {


kernel/src/process/posix_thread/ptrace.rs
```diff
@@ -0,0 +1,76 @@
...
+// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/kernel/ptrace.c#L276>
+pub fn check_may_access(
+    current_pthread: &PosixThread,
+    target_pthread: &PosixThread,
+    mode: PtraceMode,
+) -> Result<()> {
+    if mode.contains(PtraceMode::FSCREDS) == mode.contains(PtraceMode::REALCREDS) {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "should specify exactly one of FSCREDS and REALCREDS"
```
lrh2000: This is a kernel-space problem instead of a system call error, right? It would be better to enforce this at the type level. I don't understand why we should follow Linux's kernel warning.

kernel/src/fs/procfs/pid/task/mem.rs
```diff
@@ -25,10 +34,49 @@ impl MemFileOps {
...
+
+    fn open(
+        &self,
+        _access_mode: AccessMode,
+        _status_flags: StatusFlags,
+    ) -> Option<Result<Box<dyn FileIo>>> {
+        if self.0.lock_vmar().as_ref().is_none() {
+            return Some(Err(Error::with_message(
+                Errno::EACCES,
+                "the process has exited",
```
lrh2000: I do not believe this is a correct error code..

kernel/src/fs/procfs/pid/task/mem.rs
```diff
@@ -25,10 +34,49 @@ impl MemFileOps {
...
+            )));
+        }
+
+        let handle = check_may_access(
+            current_thread!().as_posix_thread().unwrap(),
+            self.0.main_thread().as_posix_thread().unwrap(),
+            PtraceMode::ATTACH_FSCREDS,
+        )
+        .map(|_| Box::new(MemFileHandle(self.0.clone())) as Box<dyn FileIo>);
```
lrh2000: This cannot be the correct way to implement a permission check. If it is implemented this way, one can escalate privileges by:
 - Execute a program and open its `/proc/pid/mem`.
 - Let the program execute `sudo`.
 - Write to the program's `/proc/pid/mem`, which has been opened before.

You'll become root if you can write to the memory of the sudo program, because sudo is a suid program and always runs as root.

The Linux implementation appears to record the pointer of the MM struct. For the same purpose, we can hold a `Weak<Vmar>` in the `MemFileHandle`. Please note that cloning an `Arc<Vmar>` or upgrading a `Weak<Vmar>` is not permitted, as this will cause the `Vmar` to be dropped in the wrong context and break the reference count:
https://github.com/asterinas/asterinas/blob/97bc17fa58ff2d3be555c48be8973e4d1c924b94/kernel/src/context.rs#L89-L94

So we should do some pointer arithmetic to ensure the `Vmar` is still up-to-date in `read_at` and `write_at` instead of relying on `Weak::upgrade`.

kernel/src/fs/procfs/pid/task/maps.rs
```diff
@@ -50,4 +91,30 @@ impl FileOps for MapsFileOps {
...
+        &self,
+        _offset: usize,
+        _reader: &mut VmReader,
+        _status_flags: StatusFlags,
+    ) -> Result<usize> {
+        return_errno_with_message!(Errno::EACCES, "`/proc/[pid]/maps` is read-only");
+    }
+}
+
+impl Pollable for MapsFileHandle {
```
lrh2000: Reorder the trait impls. Follow `Pollable`, `InodeIo`, `FileIo` instead.

kernel/src/process/posix_thread/ptrace.rs
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use bitflags::bitflags;
+
+use crate::{
+    prelude::*,
+    process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
+};
+
+/// Checks whether the current `PosixThread` may access the given target `PosixThread`
```
lrh2000:
```suggestion
/// Checks whether the current `PosixThread` may access the given target `PosixThread`.
```

Should we also mention ptrace in the documentation comments (or the method name?)? I'm not sure if the current one is clear enough. What does `access` mean?

kernel/src/fs/procfs/pid/task/maps.rs
```diff
@@ -50,4 +91,30 @@ impl FileOps for MapsFileOps {
...
         Ok(printer.bytes_written())
     }
+
+    fn write_at(
+        &self,
+        _offset: usize,
+        _reader: &mut VmReader,
+        _status_flags: StatusFlags,
+    ) -> Result<usize> {
+        return_errno_with_message!(Errno::EACCES, "`/proc/[pid]/maps` is read-only");
```
lrh2000: https://github.com/asterinas/asterinas/blob/97bc17fa58ff2d3be555c48be8973e4d1c924b94/kernel/src/fs/procfs/template/file.rs#L116-L117

Should the error code be `EACCESS` or `EPERM`?

Align the error message, "is read-only" => "is not writable".

lrh2000 (commented):
Thanks for adding force-write support and ptrace permission checks.

kernel/src/fs/procfs/pid/task/maps.rs
```diff
@@ -28,12 +37,44 @@ impl MapsFileOps {
 }
 
 impl FileOps for MapsFileOps {
-    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+    fn read_at(&self, _offset: usize, _writer: &mut VmWriter) -> Result<usize> {
+        unreachable!("should read via opened `MapsFileHandle`")
```
lrh2000: Is it possible to avoid repeating this by introducing a `FileOpsViaHandle` trait?
```rust
trait FileOpsByHandle {
    fn open() -> Result<Box<dyn FileIo>>;
}

impl<T: FileOpsByHandle> FileOps for T {
    // TODO
}
```

Is there a better name than `FileOpsByHandle`?

test/initramfs/src/syscall/gvisor/blocklists/proc_test
```diff
@@ -11,7 +11,7 @@ ProcPid.AccessDeny
 ProcPidCwd.Subprocess
 ProcPidRoot.Subprocess
 ProcPidEnviron.MatchesEnviron
-ProcPidMem.AfterExit
+# TODO: support `ptrace`.
```
lrh2000:
```suggestion
# TODO: Support `ptrace` syscall.
```

kernel/src/fs/procfs/pid/task/mem.rs
vvvvsv: Thanks! I just checked that in Linux 6.15, its behavior is returning 0.

https://elixir.bootlin.com/linux/v6.15/source/fs/proc/base.c#L843

tatetian:
@boterinas codex

kernel/src/process/posix_thread/ptrace.rs
```diff
@@ -0,0 +1,81 @@
...
+    if !caller_is_same && !caller_has_cap {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "the calling process does not have the required permissions"
+        );
+    }
+
+    // TODO: Add further security checks (e.g., YAMA LSM).
+
+    Ok(())
```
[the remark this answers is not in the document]
vvvvsv: > notably dumpability/commoncap-style checks and other security hooks

In practice, we can not add all security hooks in one go. At least these changes are already stricter than having no access mode checks at all.

kernel/src/thread/exception.rs
```diff
@@ -28,6 +28,30 @@ pub struct PageFaultInfo {
     /// The [`VmPerms`] required by the memory operation that causes page fault.
     /// For example, a "store" operation may require `VmPerms::WRITE`.
     pub required_perms: VmPerms,
+
+    /// Whether this page fault is forced (e.g., manually triggered by `ptrace`).
+    /// A forced page fault may bypass some permission checks.
+    pub is_forced: bool,
```
tatetian: This field is security relevant. Because it is a public field, every caller/user can set this field and thus may potentially authorize himself to force page fault handling.

To address this problem, move `PageFaultInfo` to the `vmar` module and make this field private. This way only the `vmar` module itself can set this security-sensitive field, making the `handle_page_fault` API robust against misuse.

kernel/src/process/posix_thread/ptrace.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[expect(dead_code)]
+pub struct PtraceMode(PtraceFlags, PtraceCredsMode);
+
+impl PtraceMode {
+    #[expect(dead_code)]
+    pub const READ_REALCREDS: Self = Self(PtraceFlags::READ, PtraceCredsMode::RealCreds);
+    pub const ATTACH_REALCREDS: Self = Self(PtraceFlags::ATTACH, PtraceCredsMode::RealCreds);
+    pub const READ_FSCREDS: Self = Self(PtraceFlags::READ, PtraceCredsMode::FsCreds);
+    pub const ATTACH_FSCREDS: Self = Self(PtraceFlags::ATTACH, PtraceCredsMode::FsCreds);
```
tatetian: > Note that all of the PTRACE_MODE_* constants described in this
       subsection are kernel-internal, and not visible to user space.

These names are internal to Linux. As such, we have less reasons to follow Linux's naming conventions, which are not self-explaining.

```rust
/// Checks whether the current `PosixThread` may access the given target
/// `PosixThread` via ptrace operations.
pub fn check_may_access(..) { ... }
```

As I understand it, this permission check is not coupled with ptrace. This permission check is intended for any access from a thread to another: ptrace is just one (prominent) use case of such remote access. For example, seeing `check_may_access(.., .., ptrace_mode)` in `pidfd_getfd.rs` is a surprise to me in the first glance because at the surface, the module has nothing to do with ptrace. So I would suggest repositioning everything in this module as _foreign access_, rather than ptrace. 

Because I only see a few use cases of `check_may_access` and `PtraceMode`, I can't make a verdict with a very high confidence. So I won't insist on the renaming suggestion. You can make the final call.

Either renaming or not, every types `PtraceXxx` should be specified its semantics by adding some comments.

tatetian (commented):
Thanks for the contribution! The PR looks for the most part.

kernel/src/fs/procfs/template/file.rs
```diff
@@ -116,4 +127,38 @@ pub trait FileOps: Sync + Send {
...
+        None
+    }
+}
+
+pub trait FileOpsByHandle: Sync + Send {
+    fn open(
+        &self,
+        access_mode: AccessMode,
+        status_flags: StatusFlags,
+    ) -> Option<Result<Box<dyn FileIo>>>;
```
lrh2000: Why `FileOpsByHandle` still returns a `Option`? If it returns `None`, then `unreachable!`s in `read_at` and `write_at` will be reachable...

Perhaps it can be emphasized by:
```rust
    fn read_at(&self, _offset: usize, _writer: &mut VmWriter) -> Result<usize> {
        unreachable!("`read_at` is never called when `open` returns `Some(_)`")
    }
```
So it's clearly true _only_ if:
```rust
    fn open(
        &self,
        access_mode: AccessMode,
        status_flags: StatusFlags,
    ) -> Option<Result<Box<dyn FileIo>>> {
        Some(self.open(access_mode, status_flags))
    }
```

kernel/src/fs/procfs/pid/task/mem.rs
```diff
@@ -24,28 +34,84 @@ impl MemFileOps {
...
+        _status_flags: StatusFlags,
+    ) -> Option<Result<Box<dyn FileIo>>> {
+        let vmar = self.0.lock_vmar().as_weak();
+
+        let handle = check_may_access(
+            current_thread!().as_posix_thread().unwrap(),
+            self.0.main_thread().as_posix_thread().unwrap(),
+            PtraceMode::ATTACH_FSCREDS,
+        )
+        .map(|_| Box::new(MemFileHandle(self.0.clone(), vmar)) as Box<dyn FileIo>);
```
lrh2000: We currently call `activate_vmar` before `apply_caps_from_exec` in `do_execve_no_return`, so this is still incorrect due to possible race conditions (i.e., if `check_may_access` is called after `activate_vmar` but before `apply_caps_from_exec`).

kernel/src/process/process_vm/mod.rs
```diff
@@ -183,6 +183,13 @@ impl<'a> ProcessVmarGuard<'a> {
         self.inner.as_ref().map(|v| &**v)
     }
 
+    /// Returns a weak reference to the process VMAR if it exists.
+    ///
+    /// Returns an empty `Weak` if the process has exited and its VMAR has been dropped.
+    pub fn as_weak(&self) -> Weak<Vmar> {
```
lrh2000: As I explained before, `Weak<Vmar>` cannot be upgraded. This must be documented.

Even better, it can be enforced at the type level:
```rust
impl ProcessVmarGuard {
    fn clone_weak(&self) -> VmarWeak;

    fn is_same_weak(&self, weak: &VmarWeak) {
        // (No need to call `Arc::downgrade` here..)
        core::ptr::eq(...)
    }
}

struct VmarWeak(Weak<Vmar>);
```

Document the types and methods. I'd appreciate it if you could come up with better method names than `clone_weak`/`is_same_weak`, but the original proposal, `as_weak`, sounds too cheap to me. It is not a cheap cast method.

vvvvsv: Thanks! Regarding the naming, how about `VmarWeak` -> `VmarSnapshot`, `clone_weak` -> `snapshot`, `is_same_weak` -> `is_same`?

kernel/src/process/posix_thread/ptrace.rs
vvvvsv: Thanks! Regarding the naming, how about `check_remote_access` and `RemoteAccessMode`? I think this name is better than "foreign access", since we already use `Vmar::access_remote` to refer to accessing another process's user space.

kernel/src/process/process_vm/mod.rs
lrh2000: I am fine with them. If we don't want to repeat the type name in the method name of is_same (the type appears in the argument list anyway), can it be renamed `is_same_as`? After all, `foo.is_same_as(bar)` looks better than `foo.is_same(bar)`.

kernel/src/process/posix_thread/ptrace.rs
tatetian: > how about `check_remote_access` and `RemoteAccessMode`

They look great. Even better, I think we should make `check_remote_access` a method:

```rust
// file: posix_thread/remote_access.rs

impl PosixThread {
    pub fn check_remote_access_from(&self, thread: &PosixThread, mode: RemoteAccessMode) -> Result<()> { ... }
}
```

This makes it more clear which is the local thread, which is the remote one.

---

Second thought. My concern with the term remote access is that it sounds like through network. "Remote" emphasizes distance. How about "alien access"? "alien" means one outside a group. An alien thread is one outside the current thread's thread group (the process).

vvvvsv: > "alien" means one outside a group. An alien thread is one outside the current thread's thread group (the process).

That makes sense to me. I will also rename `Vmar::access_remote` to `Vmar::access_alien`.

kernel/src/vm/vmar/vmar_impls/page_fault.rs
```diff
@@ -21,3 +28,47 @@ impl Vmar {
...
+    pub fn new(address: Vaddr, required_perms: VmPerms) -> Self {
+        Self {
+            address,
+            required_perms,
+            is_forced: false,
+            _private: (),
+        }
+    }
+
+    /// Whether this page fault is forced.
```
lrh2000:
```suggestion
    /// Returns whether this page fault is forced.
```

kernel/src/vm/vmar/vmar_impls/page_fault.rs
```diff
@@ -21,3 +28,47 @@ impl Vmar {
...
+
+    /// The [`VmPerms`] required by the memory operation that causes page fault.
+    /// For example, a "store" operation may require `VmPerms::WRITE`.
+    pub(in crate::vm::vmar) required_perms: VmPerms,
+
+    /// Whether this page fault is forced (e.g., manually triggered by `ptrace`).
+    /// A forced page fault may bypass some permission checks.
+    is_forced: bool,
+
+    _private: (),
```
lrh2000: So we don't need this field since `is_forced` is already private...Right?

kernel/src/vm/vmar/vmar_impls/page_fault.rs
```diff
@@ -1,7 +1,14 @@
 // SPDX-License-Identifier: MPL-2.0
 
+#[expect(unused_imports)]
+#[cfg(any(target_arch = "x86_64", target_arch = "riscv64"))]
+use ostd::arch::cpu::context::CpuException;
+#[expect(unused_imports)]
+#[cfg(target_arch = "loongarch64")]
+use ostd::arch::cpu::context::CpuExceptionInfo as CpuException;
```
lrh2000: This unused import is just to fix the doc comments for LoongArch64, right? I doubt it worth...

I think the following is fine. Unless someone wants to worry about LoongArch64's documentation, which is not a major issue compared to the other problems LoongArch64 has.
```rust
/// [`CpuException`]: ostd::arch::cpu::context::CpuException
```

kernel/src/vm/vmar/vmar_impls/mod.rs
```diff
@@ -14,6 +14,7 @@ use core::{array, ops::Range};
 use align_ext::AlignExt;
 use aster_util::per_cpu_counter::PerCpuCounter;
 use ostd::{cpu::CpuId, mm::VmSpace};
+pub use page_fault::PageFaultInfo;
```
lrh2000: So should we re-export the type twice or just mark the impl module as `pub(super)`?
```rust
pub(super) mod page_fault;
```

I slightly prefer the latter because it reduces the number of compiled imports from three to two:
```rust
 // This exists for code outside the module.
use crate::vm::vmar::PageFaultInfo;

// This shouldn't compile. This exists for no reason.
use crate::vm::vmar::vmar_impls::PageFaultInfo;

// This exists for code inside the module.
use crate::vm::vmar::vmar_impls::page_fault::PageFaultInfo;
```
I will also need to export something from the `map` module in another PR. So let's reach a consensus first.

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -417,7 +436,6 @@ impl VmMapping {
                         cursor.unmap(PAGE_SIZE);
                         cursor.jump(va.start).unwrap();
                         cursor.map(new_frame.into(), prop);
-                        rss_delta.add(self.rss_type(), 1);
```
lrh2000, line 420: Could this be made into a separate commit with a regression test added to it?

Adding something to `test/initramfs/src/apps/memory/mmap/mmap_vmrss.c` will work?

kernel/src/process/process_vm/mod.rs
```diff
@@ -183,6 +188,20 @@ impl<'a> ProcessVmarGuard<'a> {
...
+    pub fn snapshot(&self) -> VmarSnapshot {
+        VmarSnapshot(self.inner.as_ref().map(Arc::downgrade).unwrap_or_default())
+    }
+
+    /// Returns whether the current VMAR has the same identity as the `snapshot`.
+    pub fn is_same_as(&self, snapshot: &VmarSnapshot) -> bool {
+        // Upgrading the `Weak<Vmar>` in the `snapshot` is not permitted,
+        // as this will cause the `Vmar` to be dropped in the wrong context,
+        // and break the reference count used in [`CurrentUserSpace::is_vmar_shared`].
+        let current = self.inner.as_ref().map(Arc::downgrade).unwrap_or_default();
```
lrh2000: Why don't you use `core::ptr::eq` here? I still don't see a reason to pay the cost of `Arc::downgrade`.

kernel/src/process/process_vm/mod.rs
```diff
@@ -210,7 +229,17 @@ impl<'a> ProcessVmarGuard<'a> {
...
+///
+/// The `after_activate` hook is called after switching to `new_vmar`,
+/// while holding the process VMAR lock.
+pub(super) fn activate_vmar_and<F>(
+    ctx: &Context,
+    new_vmar: Arc<Vmar>,
+    after_activate: F,
+) -> Result<()>
+where
+    F: FnOnce() -> Result<()>,
```
lrh2000: Why not just returning the VMAR guard? So we don't need to pass the closure.

kernel/src/process/execve.rs
```diff
@@ -155,8 +155,13 @@ fn do_execve_no_return(
...
     thread_table::make_current_main_thread(ctx);
 
-    // Activate the new VMAR, where the ELF has been loaded, in the current context.
-    activate_vmar(ctx, new_vmar);
+    // Activate the new VMAR in the current context and apply file-capability changes,
+    // while holding the process VMAR lock.
+    // This prevents race conditions when checking access permissions while opening
+    // `/proc/[pid]/mem` or `/proc/[pid]/maps`.
+    activate_vmar_and(ctx, new_vmar, || {
+        apply_caps_from_exec(process, posix_thread, elf_file.inode())
```
lrh2000: I thought it was just a matter of fixing the order, but I agree with you that we need to take the VMAR lock while calling `apply_caps_from_exec` to completely rule out race conditions when gaining and dropping privileges on SUID (though the common use of SUID is not for dropping privileges).

kernel/src/process/posix_thread/mod.rs
```diff
@@ -25,6 +25,7 @@ use crate::{
     time::{Timer, TimerManager, clocks::ProfClock, timer::TimerGuard},
 };
 
+pub mod alien_access;
```
lrh2000, line 28: I think the practice with this file is to re-export the types, especially since it seems that there is only one type to expose in this case: `AlienAccessMode`. Meanwhile, it's `posix_thread.check_alien_access_from` so `posix_thread::AlienAccessMode` sounds more reasonable than `posix_thread::alien_access::AlienAccessMode`... Or do you plan on adding more types to this module?

kernel/src/vm/vmar/vmar_impls/mod.rs
vvvvsv: I agree with this.

kernel/src/vm/vmar/vm_mapping.rs
vvvvsv: I did some experiments on Linux, and surprisingly, when a COW occurs on file-backed pages, they are reclassified from `RssFile` to `RssAnon`. Also, the RSS accounting on Linux appears to be imprecise. I will open another PR to resolve this issue.

kernel/src/process/process_vm/mod.rs
```diff
@@ -183,6 +188,21 @@ impl<'a> ProcessVmarGuard<'a> {
...
+    /// Takes a snapshot of the current VMAR identity.
+    pub fn snapshot(&self) -> VmarSnapshot {
+        VmarSnapshot(self.inner.as_ref().map(Arc::downgrade).unwrap_or_default())
+    }
+
+    /// Returns whether the current VMAR has the same identity as the `snapshot`.
+    pub fn is_same_as(&self, snapshot: &VmarSnapshot) -> bool {
+        // Upgrading the `Weak<Vmar>` in the `snapshot` is not permitted,
+        // as this will cause the `Vmar` to be dropped in the wrong context,
+        // and break the reference count used in [`CurrentUserSpace::is_vmar_shared`].
```
lrh2000: Move this near `VmarSnapshot`. Also using type hyperlinks in non-doc (normal) comments is meaningless.

```rust
/// A snapshot of the process VMAR identity.
///
/// This type is used only for identity comparison.
//
// NOTE: Upgrading the `Weak<Vmar>` in the snapshot is not permitted,
// as this will cause the `Vmar` to be dropped in the wrong context,
// and break the reference count used in `CurrentUserSpace::is_vmar_shared`.
pub struct VmarSnapshot(Weak<Vmar>);

File: kernel/src/fs/procfs/pid/task/mem.rs [hunk 25]
```diff
@@ -24,28 +30,89 @@ impl MemFileOps {
...
+        reader: &mut VmReader,
+        _status_flags: StatusFlags,
+    ) -> Result<usize> {
         let vmar_guard = self.0.lock_vmar();
+        if !vmar_guard.is_same_as(&self.1) {
+            return Ok(0);
+        }
         let Some(vmar) = vmar_guard.as_ref() else {
-            return_errno_with_message!(Errno::ESRCH, "the process has exited");
+            return Ok(0);
```
lrh2000, line 104: Keep some comments in `read_at`/`write_at` and also `maps.rs`:
```rust
        if !vmar_guard.is_same_as(&self.1) {
            // The process has executed a new program.
            return Ok(0);
        }
        let Some(vmar) = vmar_guard.as_ref() else {
            // The process has exited.
            return Ok(0);
        };
```

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -366,6 +358,33 @@ impl VmMapping {
...
+            if perms.contains(VmPerms::MAY_READ) {
+                perms.insert(VmPerms::READ);
+            }
+            if self.is_cow() {
+                perms.insert(VmPerms::WRITE);
+            }
+        }
+
+        if !perms.contains(page_fault_info.required_perms) {
+            return_errno_with_message!(Errno::EACCES, "perm check fails");
```
lrh2000:
```suggestion
            return_errno_with_message!(Errno::EACCES, "the mapping does not have the required permissions");
```

kernel/src/fs/procfs/pid/task/mem.rs
```diff
@@ -24,28 +30,89 @@ impl MemFileOps {
...
+                current_thread!().as_posix_thread().unwrap(),
+                AlienAccessMode::ATTACH_WITH_FS_CREDS,
+            )?;
+
+        let vmar = vmar_guard.snapshot();
+        Ok(Box::new(MemFileHandle(self.0.clone(), vmar)))
+    }
+}
+
+struct MemFileHandle(Arc<Process>, VmarSnapshot);
```
lrh2000, line 58:
```rust
/// A file handle opened from `/proc/[pid]/task/[tid]/mem` (and also `/proc/[pid]/mem`).
struct MemFileHandle(Arc<Process>, VmarSnapshot);
```

Similar for `maps.rs`:
```rust
/// A file handle opened from `/proc/[pid]/task/[tid]/maps` (and also `/proc/[pid]/maps`).
struct MapsFileHandle(Arc<Process>, VmarSnapshot);
```

vvvvsv:
@boterinas rerun

kernel/src/vm/vmar/vmar_impls/access_alien.rs
```diff
@@ -7,7 +7,10 @@ use ostd::{
 };
 
 use super::{Vmar, is_userspace_vaddr};
-use crate::{prelude::*, thread::exception::PageFaultInfo, vm::vmar::is_userspace_vaddr_range};
+use crate::{
+    prelude::*,
+    vm::vmar::{PageFaultInfo, is_userspace_vaddr_range},
+};
 
 impl Vmar {
```
tatetian, line 19: The term `alien` is not defined throughout this module.

```rust
//! This module enables access to a VMAR from an [_alien thread](crate::process::posix_thread::alien_access),
//! which is a non-member of the owner process of the VMAR.
```

kernel/src/vm/vmar/vmar_impls/access_alien.rs
```diff
@@ -7,7 +7,10 @@ use ostd::{
...
     /// Reads memory from the process user space.
```
tatetian: Update all doc comments of the `xxx_alien` methods to mention the term alien.

```rust
    /// Reads memory in the context of an alien thread.
```

kernel/src/vm/vmar/vmar_impls/access_alien.rs
vvvvsv: This way we need to make the module `posix_thread::alien_access` pubic to make the docs. I’m OK with this, since I found that in a new PR I will need to add some types and methods to `alien_access` to support `/proc/sys/kernel/yama/ptrace_scope`, which controls the scope of allowed “alien access”.

@lrh2000 What do you think?

kernel/src/vm/vmar/vmar_impls/access_alien.rs
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! This module enables access to a VMAR from an [_alien thread](crate::process::posix_thread::alien_access),
```
lrh2000: This line is too long. Also, what is `_alien thread`? You should use either `alien thread` or `_alien thread_`.

```suggestion
//! This module enables access to a VMAR
//! from an [_alien thread_](crate::process::posix_thread::alien_access),
```

kernel/src/vm/vmar/vmar_impls/access_alien.rs
lrh2000: The previous discussion is at https://github.com/asterinas/asterinas/pull/2855#discussion_r2890965819.
> I think the practice with this file is to re-export the types, especially since it seems that there is only one type to expose in this case: `AlienAccessMode`. Meanwhile, it's `posix_thread.check_alien_access_from` so `posix_thread::AlienAccessMode` sounds more reasonable than `posix_thread::alien_access::AlienAccessMode`...

That being said, I have no strong opinion.

tatetian: > since I found that in a new PR I will need to add some types and methods to `alien_access` to support `/proc/sys/kernel/yama/ptrace_scope`

Then I think we should make the sub-module public and reference it in doc comments.

vvvvsv: Already done in the latest push :)

PR #2859, opened 2026-01-08
Add FIXMEs for `FsEvent::{DELETE_SELF, IN_IGNORED}` and support `InotifyControls::ONESHOT`

lrh2000:
Extracted from #2848.

---

This PR does _not_ attempt to resolve the issues with the `DELETE_SELF` and `IN_IGNORED` generations because doing so would necessitate further refactoring of the ext2/exfat file system implementation (see https://github.com/asterinas/asterinas/pull/2848#issuecomment-3712780420). However, this PR adds FIXMEs and regression tests to clarify the behavior.

In addition, this PR adds support for `InotifyControls::ONESHOT`. I don't think `InotifyControls::ISDIR` makes sense so I remove it.

TODOs in the gvisor tests have also been made clearer:
```bash
# TODO: Report `MOVE_*` events at the `rename()` system call.
Inotify.MoveGeneratesEvents
Inotify.MoveWatchedTargetGeneratesEvents

# TODO: Support the `splice()` system call.
Inotify.SpliceOnInotifyFD
Inotify.SpliceOnWatchTarget

# Symbolic links to the busybox binary does not work:
# "gvisor_test_temp_1103_1767603858010282988: applet not found"
Inotify.Exec

# TODO: Generate `DELETE_SELF`/`IGNORED` after closing the last FD.
Inotify.IncludeUnlinkedFile_NoRandomSave

# TODO: Support the `EXCL_UNLINK` flag.
Inotify.ExcludeUnlink_NoRandomSave
Inotify.ExcludeUnlinkDirectory_NoRandomSave
Inotify.ExcludeUnlinkMultipleChildren_NoRandomSave
Inotify.ExcludeUnlinkInodeEvents_NoRandomSave
```

### DIFF
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/inotify_add_watch.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/inotify_add_watch.scml
@@ -3,7 +3,7 @@ inotify_events = IN_ACCESS | IN_MODIFY | IN_ATTRIB | IN_CLOSE_WRITE |
                  IN_DELETE_SELF | IN_CLOSE;
 
 inotify_controls = IN_ONLYDIR | IN_DONT_FOLLOW | IN_MASK_CREATE |
-                   IN_MASK_ADD | IN_ISDIR | IN_ONESHOT;
+                   IN_MASK_ADD | IN_ONESHOT;
 
 // Add a watch to an initialized inotify instance
 inotify_add_watch(fd, pathname, mask = <inotify_events> | <inotify_controls>);
--- a/kernel/src/fs/notify/inotify.rs
+++ b/kernel/src/fs/notify/inotify.rs
@@ -239,8 +239,12 @@ impl InotifyFile {
             }
 
             // If the queue is full, drop the event.
-            // We do not return an error to the caller.
-            if event_queue.len() >= self.queue_capacity {
+            // Try to queue an overflow event in advance to alert the user.
+            if event_queue.len() == self.queue_capacity - 1 {
+                let overflow_event = InotifyEvent::new(u32::MAX, FsEvents::Q_OVERFLOW, 0, None);
+                event_queue.push_back(overflow_event);
+                return;
+            } else if event_queue.len() >= self.queue_capacity {
                 return;
             }
 
@@ -672,7 +676,6 @@ bitflags! {
         const EXCL_UNLINK   = 1 << 26; // Exclude events on unlinked objects
         const MASK_CREATE   = 1 << 28; // Only create watches
         const MASK_ADD      = 1 << 29; // Add to existing watch mask
-        const ISDIR         = 1 << 30; // Event occurred on a directory
         const ONESHOT       = 1 << 31; // Send event once
     }
 }


tatetian:
@Plucky923 Could you review this PR?

kernel/src/fs/notify/mod.rs
```diff
@@ -126,8 +126,18 @@ impl FsEventPublisher {
...
+            has_oneshot |= subscriber.deliver_event(events, name.clone());
+        }
+        drop(subscribers);
+
+        if has_oneshot {
+            let mut subscribers = self.subscribers.write();
+            // The `deliver_event()` method should already deliver the `FsEvents::IN_IGNORED`
+            // events for one-shot subscribers. Here, we simply remove them.
+            subscribers.retain(|m| !m.is_dead());
+            self.recalc_interesting_events(&subscribers);
```
Plucky923: In fsnotify mechanism, not all subscribers need oneshot, such as dnotify (although obsolete), fanotify.
At present, only inotify needs to be marked as dead, and other subscribers are monitored persistently.
So, propagating inotify's unique flags to the common interface layer might not be a relatively good choice.

- https://man7.org/linux/man-pages/man7/fanotify.7.html

lrh2000:
Note: The Docker image isn't available when I try to fix the gvisor block lists (https://github.com/asterinas/asterinas/pull/2829#issuecomment-3748605978), so I manually build the latest gvisor and update the gvisor block list according to the latest gvisor.  I hope I'm doing things correctly. However, we should rerun the CI after bumping the Docker image version anyway (https://github.com/asterinas/asterinas/pull/2890).

kernel/src/fs/notify/mod.rs
```diff
@@ -126,8 +126,18 @@ impl FsEventPublisher {
...
         }
     }
```
Plucky923, line 140: When a one-shot event triggers, we iterate the list 3 times (1 read, 1 retain, 1 recalc). At least the retain of this traversal should be avoidable, right

lrh2000: Seems to be the case. Thanks for catching this. I think we should just replace `recalc_interesting_events` with `retain_and_recalc_events`.

kernel/src/fs/notify/mod.rs
lrh2000: Do you have an alternative design? This is the best I can come up with.

I considered some alternative designs, such as maintaining a special `Vec<OneshotSubscriber>` list. However, I found that it would complicate the code too much.

> the common interface layer

The "common" layer needs to know something about the one-shot behavior in order to remove subscribers because subscribers cannot remove themselves (which would cause deadlocks). It's a fundamental thing, and I don't see how to avoid it.

> not all subscribers need oneshot, such as dnotify (although obsolete), fanotify.

So we can provide a default implementation of `is_dead` that returns `false`.
```rust
fn is_dead(&self) -> bool {
    false
}
```

Then, other subscribers won't need to worry about it.

---

[In the last revision](https://github.com/asterinas/asterinas/compare/4b3dacd79daa2a6f2917d21727c5077b9f7f4167..63d7642d3d9599ebc831969d96786622959ad9f0), I renamed `is_dead` to `is_oneshot_and_dead` and provided a default implementation that returns `false`. Hopefully, this is clearer.

Plucky923: I think, as you said, since the subscriber cannot delete itself, I believe the current implementation should be the most concise approach.

lrh2000:
[Block `InotifyAndTargetDestructionDoNotDeadlock` in ext2](https://github.com/asterinas/asterinas/compare/63d7642d3d9599ebc831969d96786622959ad9f0..8bbda9b902a105be56e558e2250e1db367e5caa0) to fix the CI. I think `PosixError(errno=28 No space left on device)` is caused by:
https://github.com/google/gvisor/blob/eca1f2af7f09d6ac062816a8bf448f899ede9523/test/syscalls/linux/inotify.cc#L2344-L2350

tatetian (approved):

PR #2860, opened 2026-01-08
Use accurate build time only when publishing

StevenJiang1110:
While refactoring the Makefile, I discovered that OSDK's cache build mechanism is currently broken. The issue stems from the fact that every `cargo osdk build` command passes the current system time as an environment variable to the kernel, which is then used to generate the contents of `/proc/version`.

While the impact on `make kernel` is negligible due to its short build time, it becomes highly problematic for `make nixos` or `make iso`. Since the kernel's binary content changes with every build, it triggers a full, redundant rebuild of the NixOS system or the ISO image, which is quite frustrating.

To resolve this, this PR introduces a new environment variable to toggle the use of the current timestamp. In most scenarios, the build will now default to a fixed timestamp (1970-01-01 00:00:00). The actual system time will only be used for official release builds, effectively minimizing unnecessary rebuilds.

### DIFF
--- a/osdk/src/commands/build/mod.rs
+++ b/osdk/src/commands/build/mod.rs
@@ -250,19 +250,7 @@ fn build_kernel_elf(
                 .unwrap_or_else(|_| "unknown".to_string())
         ),
     );
-    // Use system date command to get the formatted time string
-    let date_output = std::process::Command::new("date")
-        // Reference: <https://man7.org/linux/man-pages/man1/date.1.html>
-        .arg("+%a %b %e %H:%M:%S %Z %Y")
-        .output()
-        .ok()
-        .and_then(|output| {
-            String::from_utf8(output.stdout)
-                .ok()
-                .map(|s| s.trim().to_string())
-        })
-        .unwrap_or_else(|| "Thu Jan  1 00:00:00 UTC 1970".to_string());
-    command.env("OSDK_BUILD_TIMESTAMP", format!("#1 {}", date_output));
+
     command.env("RUSTFLAGS", rustflags.join(" "));
     command.arg("build");
     command.arg("--features").arg(features.join(" "));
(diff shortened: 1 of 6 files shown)


osdk/src/commands/build/mod.rs
```diff
@@ -250,18 +250,28 @@ fn build_kernel_elf(
...
+                    String::from_utf8(output.stdout)
+                        .ok()
+                        .map(|s| s.trim().to_string())
+                })
+                .unwrap_or_else(|| DEFAULT_TIME.to_string())
+        } else {
+            DEFAULT_TIME.to_string()
+        }
+    };
     command.env("OSDK_BUILD_TIMESTAMP", format!("#1 {}", date_output));
```
tatetian: I am thinking if introducing `OSDK_BUILD_TIMESTAMP` was a mistake. If we remove `OSDK_BUILD_TIMESTAMP` from OSDK and let the kernel take another env var, say `ASTER_BUILD_TIMESTAMP`, the the solution would be simpler:

1. In `uts_ns.rs`: 

```rust
    pub const VERSION: &str = {
        if let Some(version) = option_env!("ASTER_BUILD_TIMESTAMP") {
            version
        } else {
            const UNIX_EPOCH: &'static str = "Thu Jan  1 00:00:00 UTC 1970";
            UNIX_EPOCH
        }
    };
```

2. In `publish_aster_nixos.yml`:

```
export ASTER_BUILD_TIMESTAMP=`date +%a %b %e %H:%M:%S %Z %Y`
make iso RELEASE=1 AUTO_INSTALL=false ARCH=${{ matrix.arch }}
```

If this solution works, then it means that baking the build timestamp into OSDK is unnecessary and we should remove this feature from OSDK.

tatetian (commented):
Kudos for identifying this bug!

tatetian (approved):

PR #2863, opened 2026-01-09
Refactor apps test to support fine-grained build

cqs21:
This PR refactors the apps test structure to support more granular build. The main changes include:
- ~~Rename apps to custom: The `apps` directory has been renamed to `custom` for better semantic clarity.~~
- Fine-grained test compilation: Previously, all tests under `apps` were compiled into a single Nix package. This has been restructured so that test cases are categorized and each category is compiled into separate Nix packages.
- Architecture-specific filtering: Added filter mechanisms in the common Makefile to exclude certain test cases based on target architecture. For example, x86_64-specific tests are now filtered out when building for other architectures:
```
ifneq ($(HOST_PLATFORM), x86_64-linux)
C_OBJS_FILTER += \
    sigaltstack \
    signal_fpu \
    signal_rflags_df \
    signal_test
endif
```


test/initramfs/nix/custom/mongoose.nix
```diff
@@ -0,0 +1,22 @@
...
+      url = "${prefix}/mongoose.h";
+      sha256 = "sha256-JMqSTzO5c9qlptzN8mE84eB9DWiq6Fb3DlLVyPcQ498";
+    };
+  };
+in stdenv.mkDerivation {
+  pname = "mongoose";
+  version = "0.1.0";
+  buildCommand = ''
+    mkdir -p $out
+    cp ${files.mongoose_c} $out/mongoose.c
```
tatetian: I suggest removing Mongoose. The motivation of including this program in the "custom" test suite was to demonstrate that Asterinas could support running a Web framework in its early days. But given the maturity of Asterinas now, we don't need to show this point with this _out-liner_ test program: all other test programs are written to test one specific aspect of the kernel, but the Mongoose test involves a group of network-related system calls. It is more akin to those under our current NixOS-based test suites.

test/initramfs/nix/default.nix
```diff
@@ -1,7 +1,7 @@
-{ target ? "x86_64", enableBasicTest ? false, basicTestPlatform ? "asterinas"
+{ target ? "x86_64", testPlatform ? "asterinas", enableCustomTest ? false
 , enableBenchmark ? false, enableSyscallTest ? false, syscallTestSuite ? "ltp"
 , syscallTestWorkDir ? "/tmp", dnsServer ? "none", smp ? 1
-, initramfsCompressed ? true, }:
+, initramfsCompressed ? true, intelTdx ? 0, }:
```
tatetian: This is Nix code, not Makefile or Bash scripts. Use a boolean type to express a true or false value.

test/initramfs/nix/custom/default.nix
```diff
@@ -0,0 +1,37 @@
...
+    root = ./../../src/custom/scripts;
+    fileset = lib.fileset.fileFilter (file: file.hasExt "sh")
+      ./../../src/custom/scripts;
+  };
+
+  buildDir = dir:
+    callPackage ./common.nix { inherit dir testPlatform intelTdx; };
+  subDirs = [
+    "device"
+    "examples"
```
tatetian: The `examples` directory is NOT in the same league as other directories in `custom`. Every thing under the `custom` directory should be tests, not examples. So let's abandon this `examples` directory and move every thing beneath it to a proper location.

First, we can add a `hello_world` directory (not `hello`) under the `custom` directory, which includes all versions of Hello World.

```
custom/
    hello_world/
        in_c/
        in_c_pie/
        in_assembly/
```

Why keep `hello_world` as a test? Why not just remove it since the maturity of Asterinas has passed the stage of testing Hello World programs?

The rationale is that we will port Asterinas to new CPU architectures or platforms. And when doing so, the developer can target the Hello World tests as the first step.

The `generate_tdx_quote` program can be seen as a test case for the TDX subsystem. So we can add a TDX directory and move the program into this new directory.

```rust
custom/
    intel_tdx/
        gen_quote/
```

The Mongoose-based Web server can be removed all together. The server is added in a very early stage of Asterinas, when running a Web server is still something to show off. But now running a full program should be demonstrated as a NixOS test. All tests under `custom` should aim at testing a specific aspect of the kernel.

test/initramfs/nix/custom/common.nix
```diff
@@ -0,0 +1,37 @@
+{ lib, pkgs, stdenv, dir, testPlatform ? "asterinas", intelTdx ? 0, }:
```
tatetian: This file needs an overhaul.

It is named `common`; so it should contain the "common" part. But it actually contains a lot "differences", i.e., the directory-specific logic:
* `if dir == "examples" then pkgs.callPackage ./mongoose.nix `
* `if intelTdx != 0 then pkgs.callPackage ./tdx-attest.nix { }`
* `MONGOOSE_DIR = lib.optionalString (dir == "examples") "${mongoose}";`
* `lib.optionalString (dir == "network") " -I${pkgs.libnl.dev}/include/libnl3";`
* more...

This file smells "bad" because it is against the open–closed principle:

> The open–closed principle (OCP) states "software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification".

A scalable test build system should allow for new tests to be added by _extension_ (e.g., adding new files or inserting new items to a list), not by _modification_ (e.g., to this file).

A good example of a scalable build approach that satisfies the OCP principle is how the Makefile files under `apps/` (now you call it `custom/`) is organized:

```rust
apps/
    my_test/
        Makefile     % Most Makefile only needs a single line:
                     %    include ../test_common.mk
                     % But may also optionally set more flags as it wish
    test_common.mk   % A test framework shared by all test programs
    Makefile         % Adding a new test only requires editing the test list
```

Think of a better approach to organize Nix files.

test/initramfs/Makefile
```diff
@@ -65,15 +65,16 @@ $(INITRAMFS_IMAGE): $(INITRAMFS)
...
+		--argstr testPlatform $(TEST_PLATFORM) \
+		--arg enableCustomTest $(ENABLE_CUSTOM_TEST) \
 		--arg enableBenchmark $(ENABLE_BENCHMARK) \
 		--arg enableSyscallTest $(ENABLE_SYSCALL_TEST) \
 		--argstr syscallTestSuite $(SYSCALL_TEST_SUITE) \
 		--argstr syscallTestWorkDir $(SYSCALL_TEST_WORKDIR) \
 		--argstr dnsServer ${DNS_SERVER} \
 		--arg initramfsCompressed $(INITRAMFS_COMPRESSED) \
 		--arg smp $(SMP) \
+		--arg intelTdx $(INTEL_TDX) \
```
tatetian: I don't think this test should be turned on or off. There would be too many knobs if each CPU features is to be given one. Instead, the test should be refactored to test if the TDX is available on the current platform (e.g., by checking the existence of certain file or directory).

.github/actions/test/action.yml
```diff
@@ -6,7 +6,7 @@ branding:
 inputs:
   # Test Configuration
   auto_test:
-    description: 'Test type (general, osdk, boot, syscall, test)'
+    description: 'Test type (general, osdk, boot, syscall, custom)'
```
tatetian: The new name is probably better than `apps/`, but it is still unsatisfactory to me. Let me explain why. The stuff under `test/initramfs/src/` are below:

```
benchmark/
custom/
etc/
syscall/
```

#2945 introduces the xfstest test suite. It proposes to add a `fs/xfstest` directory. As such, the combination of this PR and #2945 would lead to the following structure:

```
benchmark/
custom/
etc/
syscall/
fs/
```

(Ignore `etc/`; we will deal with it later)

Is this structure (and the names) good or bad? What is the criterion to make such a judgement?

My criterion is that the sub-directories (and their names) should be **mutually exclusive and collectively exhaustive**. What we are really doing here is not to choose names, but to find a way to **categorize all sorts of tests that run on initramfs**. In other words, we try to figure out a set of names

```
dir1/
dir2/
dir3/
..
dirN/
```

so that the test categories represented by `dirX` do not overlap and cover all foreseeable test suites.

Are the currently proposed directories (`benchmark/`, `custom/`, `syscall/`, and `fs/`) mutually exclusive and collectively exhaustive? No, they are not. The categories `custom` and `syscall` overlap: in fact, the `custom` directory contains syscall test programs, which seem to belong to `syscall/`. The categories `benchmark` and `fs` overlap. Where would a FS benchmark go: `benchmark` or `fs`?

You see. Poorly-chosen categories are confusing and unmaintainable. Our problem is not about names, but categorization.

I have two proposals for good categorization.

The first one is:

```
in_house/
    syscall/    % the old "custom" or "apps" directory
    examples/   % If there are really "examples"
third_party/
    gvisor/
    ltp/
    xfstests/
    benchmarks/
```

It is obvious that `in_house/` and `third_party/` are mutually exclusive and collectively exhaustive because you either write the tests yourself or borrow existing ones from somewhere else.

The second proposal is shown below:

```
regression/   % In-house syscall tests to catch regressions
conformance/  % Third-party test suites to verify Linux compatibility
    gvisor/
    ltp/
    xfstests/
benchmark/    % Third-party benchmark suites to compare Asterinas against Linux
```

The second option also satisfies our criterion because it categorizes test programs according to its purposes (catch regressions, verify conformance, or evaluate performance), which are mutually exclusive and collectively exhaustive.

I am inclined to the second option because it keeps the `benchmark` directory at the first level.

What about `etc/`? My answer is that it should never have appeared in `src/` at all. It should be moved to the upper directory:

```rust
initramfs/
    src/
    nix/
    etc/
```

Don't you think? This looks much better.

tatetian (commented):
Thanks for making this long-awaited refactoring a reality.

The PR description says that the PR has three main changes:

> * Rename apps to custom: The `apps` directory has been renamed to `custom` for better semantic clarity.
> * Fine-grained test compilation: Previously, all tests under `apps` were compiled into a single Nix package. This has been restructured so that test cases are categorized and each category is compiled into separate Nix packages.
> * Architecture-specific filtering: Added filter mechanisms in the common Makefile to exclude certain test cases based on target architecture. For example, x86_64-specific tests are now filtered out when building for other architectures

But these changes are squeezed into a single commit. This is unfortunate, making it hard to do code review. They should have been split into three commits.

In my last review comment, I suggested a new structure for the `test/initramfs/src/` directory. If you agree with it, do it in a separate PR so that this PR is less likely to conflict with #2945.

tatetian:
I believe it is more logical to do all the renaming work together in a separate PR:
* apps --> regression
* syscall --> conformance
* etc --> moved to the upper directory
* test/initramfs/README.md --> updated to reflect the new structure

Because as I said, renaming is about a new way to categorize the tests. Simply changing `apps` to `regression` would not the directory structure less sensible.

After the renaming, you can split the regression test into small build units in this PR.

tools/docker/nix/Dockerfile
```diff
@@ -36,9 +36,8 @@ RUN make riscv64_pkgs VERBOSE=0 \
 RUN make x86_64_pkgs VERBOSE=0 \
     && make nix_gc
 
-# Build general packages and install host required packages
-RUN make general_pkgs VERBOSE=0 \
-    && make install_host_pkgs \
+# Install host required packages
+RUN make install_host_pkgs \
     && make nix_gc
```
tatetian, line 41: The Docker image is affected by this PR. Did you test the changes by creating a new Docker image and running tests on it? If you didn't test, are you highly confident that this PR wouldn't break the Docker image?

test/initramfs/src/apps/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+# Prevent the implicit rules from compiling ".c" or ".s" files automatically.
+MAKEFLAGS += --no-builtin-rules
+
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+CUR_DIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+BUILD_DIR ?= $(CUR_DIR)/../../build
+OUTPUT_DIR ?= $(BUILD_DIR)/initramfs/test
+
+SUBDIRS := \
```
tatetian, line 14: Add a reminder:

```Makefile
# REMINDER: Whenever you change this list, update the corresponding `default.nix` file.
SUBDIRS := \
```

test/initramfs/nix/regression/default.nix
```diff
@@ -0,0 +1,42 @@
+{ lib, pkgs, stdenv, callPackage, testPlatform ? "asterinas", }: rec {
```
tatetian: Using `rec` here is error-prone. A better style would be:

```nix
  { lib, pkgs, stdenv, callPackage, testPlatform ? "asterinas" }:
  let
    scripts = ...;
    allPkgs = ...;
  in {
    inherit scripts allPkgs;
    package = ...;
  }
```

tatetian:
@cqs21 I don't why your PRs cannot request Copilot for review. Other people's PRs have the "Ask Copilot for Review" button.

image

I recommend you to ask a coding agent to improve the `regression/default.nix` file. It does not look quite clean or idiomatic.

tatetian:
> I believe it is more logical to do all the renaming work together in a separate PR:
> 
> * apps --> regression
> * syscall --> conformance
> * etc --> moved to the upper directory
> * test/initramfs/README.md --> updated to reflect the new structure

You can do the renaming and reorganizing job in a separate PR _after_ this PR.

tools/docker/nix/Dockerfile
cqs21: >  If you didn't test, are you highly confident that this PR wouldn't break the Docker image?

I don't think building one less package will damage the docker image.

PR #2871, opened 2026-01-11
Rename `blocks_count` to `logical_block_count` and fix the related code.

LordaeronESZ:
Related with #2807, #2776

This PR fixes: 
- Contrary to its comment, current `InodeDesc::blocks_count` does not return the actual number of blocks occupied by the file. Besides, in the case of sparse files, `debug_assert!(blocks <= sectors_to_blocks(self.sector_count));` will be triggered incorrectly.
- When `fstat` a file, the `Blocks` field should be its sector count, not its logical block count.

### DIFF
--- a/kernel/src/fs/ext2/inode.rs
+++ b/kernel/src/fs/ext2/inode.rs
@@ -107,7 +107,11 @@ impl Inode {
             ino: self.ino() as _,
             size: inner.file_size() as _,
             blk_size: BLOCK_SIZE,
-            blocks: inner.blocks_count() as _,
+            // We should return the number of blocks occupied by this file rather than the file size.
+            // The former value would be smaller than the latter one if the file is sparse.
+            //
+            // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/stat.h#L36>
+            blocks: sectors_to_blocks(inner.occupied_sectors_count()) as usize,
             atime: inner.atime(),
             mtime: inner.mtime(),
             ctime: inner.ctime(),
@@ -884,7 +888,7 @@ impl Inode {
     pub fn gid(&self) -> u32;
     pub fn file_flags(&self) -> FileFlags;
     pub fn hard_links(&self) -> u16;
-    pub fn blocks_count(&self) -> Ext2Bid;
+    pub fn size_in_blocks(&self) -> Ext2Bid;
     pub fn acl(&self) -> Option<Bid>;
     pub fn atime(&self) -> Duration;
     pub fn mtime(&self) -> Duration;
@@ -1188,7 +1192,8 @@ impl InodeInner {
     pub fn hard_links(&self) -> u16;
     pub fn inc_hard_links(&mut self);
     pub fn dec_hard_links(&mut self);
-    pub fn blocks_count(&self) -> Ext2Bid;
+    pub fn size_in_blocks(&self) -> Ext2Bid;
+    pub fn occupied_sectors_count(&self) -> u32;
     pub fn acl(&self) -> Option<Bid>;
     pub fn set_acl(&mut self, bid: Bid);
     pub fn atime(&self) -> Duration;
@@ -1213,7 +1218,7 @@ struct InodeImpl {
 impl InodeImpl {
     pub fn new(desc: Dirty<InodeDesc>, weak_self: Weak<Inode>, fs: Weak<Ext2>) -> Self {
         let block_manager = InodeBlockManager {
-            nblocks: AtomicUsize::new(desc.blocks_count() as _),
+            nblocks: AtomicUsize::new(desc.size_in_blocks() as _),
             block_ptrs: RwMutex::new(desc.block_ptrs),
             indirect_blocks: RwMutex::new(IndirectBlockCache::new(fs.clone())),
             fs,
@@ -1284,8 +1289,12 @@ impl InodeImpl {
         self.desc.hard_links -= 1;
     }
 
-    pub fn blocks_count(&self) -> Ext2Bid {
-        self.desc.blocks_count()
+    pub fn size_in_blocks(&self) -> Ext2Bid {
+        self.desc.size_in_blocks()
+    }
+
+    pub fn occupied_sectors_count(&self) -> u32 {
+        self.desc.occupied_sectors_count
     }
 
     pub fn acl(&self) -> Option<Bid> {
@@ -1396,7 +1405,7 @@ impl InodeImpl {
     /// which may result in an increased block count.
     fn expand(&mut self, new_size: usize) -> Result<()> {
         let new_blocks = self.desc.size_to_blocks(new_size);
-        let old_blocks = self.desc.blocks_count();
+        let old_blocks = self.desc.size_in_blocks();
 
         // Expands block count if necessary
         if new_blocks > old_blocks {
@@ -1473,7 +1482,8 @@ impl InodeImpl {
                 self.fs().free_blocks(device_range).unwrap();
                 return Err(e);
             }
-            self.desc.sector_count = blocks_to_sectors(range.start + device_range.len() as u32);
+            self.desc.occupied_sectors_count =
+                blocks_to_sectors(range.start + device_range.len() as u32);
             self.last_alloc_device_bid = Some(device_range.end - 1);
             return Ok(device_range.len() as Ext2Bid);
         }
@@ -1519,7 +1529,8 @@ impl InodeImpl {
             return Err(e);
         }
 
-        self.desc.sector_count = blocks_to_sectors(range.start + device_range.len() as u32);
+        self.desc.occupied_sectors_count =
+            blocks_to_sectors(range.start + device_range.len() as u32);
         self.last_alloc_device_bid = Some(device_range.end - 1);
         Ok(device_range.len() as Ext2Bid)
     }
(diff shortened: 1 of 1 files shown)


kernel/src/fs/ext2/inode.rs
```diff
@@ -1188,7 +1188,8 @@ impl InodeInner {
     pub fn hard_links(&self) -> u16;
     pub fn inc_hard_links(&mut self);
     pub fn dec_hard_links(&mut self);
-    pub fn blocks_count(&self) -> Ext2Bid;
+    pub fn logical_block_count(&self) -> Ext2Bid;
+    pub fn sector_count(&self) -> u32;
```
tatetian: Since these methods are adjacent and looks similar, people would easily get confused. I suppose that `logical_block_count` is more like `size_in_blocks` and `sector_count` is more like `total_allocated_sectors`. Am I right?

kernel/src/fs/ext2/inode.rs
```diff
@@ -2197,16 +2202,11 @@ impl InodeDesc {
...
     }
 
-    /// Returns the actual number of blocks utilized.
-    ///
-    /// Ext2 allows the `block_count` to exceed the actual number of blocks utilized.
-    pub fn blocks_count(&self) -> Ext2Bid {
-        let blocks = self.size_to_blocks(self.size);
-        debug_assert!(blocks <= sectors_to_blocks(self.sector_count));
-        blocks
+    pub fn logical_block_count(&self) -> Ext2Bid {
```
tatetian: Is this "virtual block / count" or "virtual / block count"? The "/" mark means a breath break. Either way, calling it "virtual" is confusing to me. And this concept does not appear anywhere else in the codebase.

kernel/src/fs/ext2/inode.rs
LordaeronESZ: You're right.

kernel/src/fs/ext2/inode.rs
LordaeronESZ: It's "logical block / count".

In my consideration, I used "logical" to emphasize that it refers to the file's `size_to_blocks` rather than the actual number of blocks occupied; I feel the original `blocks_count` might be confusing.

Or, is there any better naming suggestion?

tatetian: `size_in_blocks` sounds good to me. There is already a `size` field, which is in bytes. So having this method feels natural.

Since the Ext2 spec and the current code prefer to name things as `xxx_count`, I suppose renaming `sector_count` to `used_sectors_count` or `inuse_sectors_count` or `occupied_sectors_count` or `allocated_sectors_count` would improve clarity and also keep consistency.

kernel/src/fs/ext2/inode.rs
```diff
@@ -107,7 +107,7 @@ impl Inode {
             ino: self.ino() as _,
             size: inner.file_size() as _,
             blk_size: BLOCK_SIZE,
-            blocks: inner.blocks_count() as _,
+            blocks: sectors_to_blocks(inner.sector_count()) as usize,
```
tatetian: Some explanations would benefit future readers:

```rust
            // We should return the number of blocks occupied by this file rather than the file size.
            // The former value would be smaller than the latter one if the file is sparse.
            //
            // Reference: <https://give.the.reference>
            blocks: sectors_to_blocks(inner.sector_count()) as usize,
```

kernel/src/fs/ext2/inode.rs
```diff
@@ -107,7 +107,11 @@ impl Inode {
             ino: self.ino() as _,
             size: inner.file_size() as _,
             blk_size: BLOCK_SIZE,
-            blocks: inner.blocks_count() as _,
+            // We should return the number of blocks occupied by this file rather than the file size.
+            // The former value would be smaller than the latter one if the file is sparse.
+            //
+            // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/stat.h#L36>
+            blocks: sectors_to_blocks(inner.occupied_sectors_count()) as usize,
```
tatetian, line 114:
```c
	long		st_blocks;	/* Number 512-byte blocks allocated. */
```

Wait. If the comment is accurate, our code is wrong in converting the number to 4KB blocks.

LordaeronESZ: This does look rather wired.

The calling chain for `fstat` system call to statistic a file's metadata in Ext2 is as follows:

``` rust
sys_fstat /* kernel/src/syscall/stat.rs */
	-> Ext2Inode::metadata /* kernel/src/fs/ext2/impl_for_vfs/inode.rs */
		-> Inode::metadata /* kernel/src/fs/ext2/inode.rs */
```

https://github.com/asterinas/asterinas/blob/738d14f3c033e8dc26b8d374872e5bd515e7a1fc/kernel/src/syscall/stat.rs#L23

The `fstat` will eventually call a `Stat::from` method to further parse the metadata obtained from the inode. The definition of this method is as follows:

https://github.com/asterinas/asterinas/blob/738d14f3c033e8dc26b8d374872e5bd515e7a1fc/kernel/src/syscall/stat.rs#L175-L194

In the line 187, it performes a conversion similar to `blocks_to_sectors`. Therefore, for the sake of consistancy, I need to do a `sectors_to_blocks` in Ext2's `Inode::metadata` in advance.

tatetian: Ok. Great. I think we have identified an area of improvement.

```rust
#[derive(Debug, Clone, Copy)]
pub struct Metadata {
    pub dev: u64,
    pub ino: u64,
    pub size: usize,
    pub blk_size: usize,
    pub blocks: usize,
    pub atime: Duration,
    pub mtime: Duration,
    pub ctime: Duration,
    pub type_: InodeType,
    pub mode: InodeMode,
    pub nlinks: usize,
    pub uid: Uid,
    pub gid: Gid,
    pub rdev: u64,
}
```

The current semantics of `Metadata` is wrong. The `blocks` field should have been named `nr_sectors_allocated` for accuracy. Throughout the codebase, we have used the term "block size" for 4KB data chunks and "sector size" for 512-byte data chunks. So the `blk_size` field should have been named `optimal_block_size`.

Since the `Metadata` is widely used, it deserves to be absolute clear and well documented. Here is a revised version:

```rust
/// File metadata, providing detailed information about an inode.
///
/// Asterinas's type-safe counterparts for Linux's `struct stat`.
#[derive(Debug, Clone, Copy)]
pub struct Metadata {
    /// The inode number, which uniquely identifies the file within the filesystem.
    ///
    /// Corresponds to `st_ino`.
    pub ino: u64,

    /// The size of the inode.
    ///
    /// The interpretation depends on `inode_type`:
    /// - **Regular File**: The total length of the file content.
    /// - **Directory**: The size of the directory's internal table (usually a multiple of block size).
    /// - **Symbolic Link**: The length of the target pathname.
    /// - **Device/Socket/FIFO**: Usually zero.
    ///
    /// Corresponds to `st_size`.
    pub size: usize,

    /// The optimal block size for filesystem I/O operations.
    ///
    /// Corresponds to `st_blksize`.
    pub optimal_block_size: usize,

    /// Number of 512-byte sectors allocated for the inode on disk.
    ///
    /// Corresponds to `st_blocks`.
    ///
    /// This represents physical usage.
    /// For sparse files (those having holes), `size` is greater than this field.
    /// For files with preallocated blocks (`FALLOC_FL_KEEP_SIZE`),
    /// `size` is smaller than this field.
    pub nr_sectors_allocated: usize,

    /// The timestamp of the last access to the inode's data.
    ///
    /// Corresponds to `st_atime`.
    pub last_access_at: Duration,

    /// The timestamp of the last modification to the inode's content.
    ///
    /// Corresponds to `st_mtime`.
    pub last_modify_at: Duration,

    /// The timestamp of the last change to the inode's metadata.
    ///
    /// This is updated when permissions, ownership, or link count change,
    /// not just when the inode content is modified.
    ///
    /// Corresponds to `st_ctime`.
    pub last_meta_change_at: Duration,

    /// The type of the inode (e.g., regular file, directory, symlink).
    ///
    /// Derived from the file type bits of `st_mode` (using the `S_IFMT` mask).
    pub type_: InodeType,

    /// The inode mode, representing access permissions.
    ///
    /// Derived from the permission bits of `st_mode`.
    pub mode: InodeMode,

    /// The number of hard links pointing to this inode.
    ///
    /// Corresponds to `st_nlink`.
    pub nr_hard_links: usize,

    /// The User ID (UID) of the inode's owner.
    ///
    /// Corresponds to `st_uid`.
    pub uid: Uid,

    /// The Group ID (GID) of the inode's owner.
    ///
    /// Corresponds to `st_gid`.
    pub gid: Gid,

    /// The ID of the device containing the inode.
    ///
    /// For persisted files, this device could be a on-disk partition or a logical volume (with RAID).
    /// For pseudo files (e.g., those on sockfs), this device is also "pseudo".
    ///
    /// Corresponds to `st_dev`.
    pub container_dev_id: DeviceId,

    /// The device ID of the inode itself, if this inode represents a special device file (character or block).
    ///
    /// Corresponds to `st_rdev`.
    pub self_dev_id: Option<DeviceId>,
}
```

Could you submit a separate PR for this refactoring? It could affect quite some existing users of `Metadata`.

LordaeronESZ: The PR has been created: #2887

Issue #2866, opened 2026-01-09
Support pidfd_send_signal and pidfd_getfd to ensure pidfd feature parity


StevenJiang1110:
We currently support the `pidfd_open` syscall, but lack support for other related syscalls such as `pidfd_send_signal` and `pidfd_getfd`. This partial implementation causes issues for userspace applications that rely on feature detection.

Specifically, some applications use `pidfd_open` to determine if the kernel supports the pidfd framework. If `pidfd_open` succeeds, they proceed to use `pidfd_send_signal`. Because our implementation is incomplete, these applications pass the initial check but fail during subsequent operations, leading to unexpected behavior.

I think implementing these two syscalls would be relatively straightforward:

- [ ] pidfd_send_signal
- [ ] pidfd_getfd

PR #2877, opened 2026-01-13
Refactor `PidFile` and add `pidfd_getfd` syscall

li041:
- Correct PidFile's reference to Process: Change from `Arc<Process>` to `Weak<Process>` to ensure PidFile does not prevent the process from being properly destroyed, and fix associated syscall impl
- implement `sys_pidfd_getfd` 
- Note:  **Permission management via `ptrace`** is currently missing. This should be added to ensure that only authorized processes can access the `pidfd` file descriptor.

cc https://github.com/asterinas/asterinas/issues/2866

### DIFF
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -346,6 +346,7 @@ which are summarized in the table below.
 | 434     | pidfd_open             | ✅             | 💯 |
 | 435     | clone3                 | ✅             | [⚠️](syscall-flag-coverage/process-and-thread-management/#clone-and-clone3) |
 | 436     | close_range            | ✅             | 💯 |
+| 438     | pidfd_getfd            | ✅             | 💯 |
 | 439     | faccessat2             | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#faccessat2) |
 | 441     | epoll_pwait2           | ✅             | 💯 |
 | 452     | fchmodat2              | ✅             | 💯 |
--- a/kernel/src/syscall/pidfd_getfd.rs
+++ b/kernel/src/syscall/pidfd_getfd.rs
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use crate::{
+    fs::file_table::{FdFlags, FileDesc, get_file_fast},
+    prelude::*,
+    process::{PidFile, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    syscall::SyscallReturn,
+};
+
+pub fn sys_pidfd_getfd(
+    pidfd: FileDesc,
+    targetfd: FileDesc,
+    flags: u32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    // The `flags` argument is reserved for future use. Currently, it must be specified as 0.
+    if flags != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
+    debug!(
+        "pidfd_getfd: pidfd={}, targetfd={}, flags={}",
+        pidfd, targetfd, flags
+    );
+
+    let mut file_table = ctx.thread_local.borrow_file_table_mut();
+    let file = get_file_fast!(&mut file_table, pidfd);
+    let Some(pid_file) = file.downcast_ref::<PidFile>() else {
+        return_errno_with_message!(Errno::EINVAL, "the file is not a PID file");
+    };
+
+    let process = pid_file
+        .process_opt()
+        .ok_or_else(|| Error::with_message(Errno::ESRCH, "the target process has been reaped"))?;
+
+    // The calling process should have PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
+    // over the process referred to by `pidfd`.
+    // Currently, this is implemented as requiring the calling process to have the
+    // CAP_SYS_PTRACE capability, which is stricter.
+    // TODO: Implement appropriate PTRACE_MODE_ATTACH_REALCREDS permission check.
+    if process
+        .user_ns()
+        .lock()
+        .check_cap(CapSet::SYS_PTRACE, ctx.posix_thread)
+        .is_err()
+    {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "the calling process does not have the required permissions"
+        );
+    }
+
+    let main_thread = process.main_thread();
+
+    // Get the file description corresponding to to the file descriptor `targetfd` in the process
+    // referred to by the PID file.
+    let target_file_table = main_thread.as_posix_thread().unwrap().file_table();
+    let target_file = target_file_table
+        .lock()
+        .as_ref()
+        .ok_or_else(|| Error::with_message(Errno::ESRCH, "the target process has exited"))?
+        .read()
+        .get_file(targetfd)?
+        .clone();
+
+    // Duplicate the file descriptor into the caller's file descriptor table.
+    let new_fd = {
+        let mut file_table_locked = file_table.unwrap().write();
+        file_table_locked.insert(target_file, FdFlags::CLOEXEC)
+    };
+
+    Ok(SyscallReturn::Return(new_fd as _))
+}
--- a/test/initramfs/src/apps/process/pidfd_getfd.c
+++ b/test/initramfs/src/apps/process/pidfd_getfd.c
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include <unistd.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <pthread.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include "../test.h"
+
+static int pidfd;
+static pid_t pid;
+static int fd;
+static int target_fd;
+static const char *TESTFILE = "/tmp/pidfd_getfd_testfile";
+static int invalid_pidfd = -1;
+
+static int pidfd_open(pid_t pid, unsigned int flags)
+{
+	return syscall(SYS_pidfd_open, pid, flags);
+}
+
+static int pidfd_getfd(int pidfd, int targetfd, unsigned int flags)
+{
+	return syscall(SYS_pidfd_getfd, pidfd, targetfd, flags);
+}
+
+FN_TEST(pidfd_getfd_valid)
+{
+	size_t size = sizeof(int);
+	volatile int *shared_fd =
+		TEST_SUCC(mmap(NULL, size, PROT_READ | PROT_WRITE,
+			       MAP_SHARED | MAP_ANONYMOUS, -1, 0));
+
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		fd = CHECK(open(TESTFILE, O_CREAT | O_RDWR | O_TRUNC, 0644));
+
+		CHECK_WITH(write(fd, "Test content\n", 13), _ret == 13);
+		*shared_fd = fd;
+
+		pause();
+		exit(0);
+	}
+
+	while (*shared_fd == 0) {
+		usleep(100);
+	}
+	fd = *shared_fd;
+
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+	target_fd = TEST_SUCC(pidfd_getfd(pidfd, fd, 0));
+
+	char buffer[128] = { 0 };
+	TEST_RES(pread(target_fd, buffer, sizeof(buffer), 0),
+		 strcmp(buffer, "Test content\n") == 0);
+
+	TEST_SUCC(munmap((void *)shared_fd, size));
+	TEST_SUCC(close(target_fd));
+	TEST_SUCC(close(pidfd));
+	TEST_SUCC(unlink(TESTFILE));
+
+	TEST_SUCC(kill(pid, SIGKILL));
+	TEST_SUCC(waitpid(pid, NULL, 0));
+}
+END_TEST()
+
+FN_TEST(pidfd_getfd_after_child_exits)
+{
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		exit(0);
+	}
+	TEST_SUCC(waitid(P_PID, pid, NULL,
+			 WNOWAIT | WEXITED)); // Ensure the child has exited
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+
+	TEST_ERRNO(pidfd_getfd(pidfd, fd, 0), ESRCH);
+	TEST_SUCC(waitpid(pid, NULL, 0));
+	TEST_ERRNO(pidfd_getfd(pidfd, fd, 0), ESRCH);
+
+	TEST_SUCC(close(pidfd));
+}
+END_TEST()
+
+FN_TEST(pidfd_getfd_errnos)
+{
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		exit(0);
+	}
+	TEST_SUCC(waitid(P_PID, pid, NULL,
+			 WNOWAIT | WEXITED)); // Ensure the child has exited
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+
+	TEST_ERRNO(pidfd_getfd(invalid_pidfd, fd, 0), EBADF);
+	TEST_ERRNO(pidfd_getfd(pidfd, -1, 0), ESRCH);
+	TEST_ERRNO(pidfd_getfd(pidfd, fd, 1), EINVAL);
+
+	TEST_SUCC(waitpid(pid, NULL, 0));
+	TEST_SUCC(close(pidfd));
+}
+END_TEST()
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/fully_covered.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/fully_covered.scml
@@ -32,5 +32,8 @@ select(nfds, readfds, writefds, exceptfds, timeout);
 // Obtain a file descriptor that refers to a process
 pidfd_open(pid, flags = PIDFD_NONBLOCK);
 
+// Obtain a duplicate of another process's file descriptor
+pidfd_getfd(pid, targetfd, flags = 0);
+
 // Close all file descriptors in the inclusive range [first, last]
 close_range(first, last, flags = CLOSE_RANGE_UNSHARE | CLOSE_RANGE_CLOEXEC);
--- a/kernel/src/syscall/arch/generic.rs
+++ b/kernel/src/syscall/arch/generic.rs
@@ -80,6 +80,7 @@ macro_rules! import_generic_syscall_entries {
             munmap::sys_munmap,
             nanosleep::{sys_clock_nanosleep, sys_nanosleep},
             open::sys_openat,
+            pidfd_getfd::sys_pidfd_getfd,
             pidfd_open::sys_pidfd_open,
             pipe::sys_pipe2,
             ppoll::sys_ppoll,
@@ -392,6 +393,7 @@ macro_rules! define_syscalls_with_generic_syscall_table {
             SYS_PIDFD_OPEN = 434             => sys_pidfd_open(args[..2]);
             SYS_CLONE3 = 435                 => sys_clone3(args[..2], &user_ctx);
             SYS_CLOSE_RANGE = 436            => sys_close_range(args[..3]);
+            SYS_PIDFD_GETFD = 438            => sys_pidfd_getfd(args[..3]);
             SYS_FACCESSAT2 = 439             => sys_faccessat2(args[..4]);
             SYS_EPOLL_PWAIT2 = 441           => sys_epoll_pwait2(args[..5]);
             SYS_FCHMODAT2 = 452              => sys_fchmodat2(args[..4]);
--- a/kernel/src/syscall/arch/x86.rs
+++ b/kernel/src/syscall/arch/x86.rs
@@ -81,6 +81,7 @@ use super::{
     nanosleep::{sys_clock_nanosleep, sys_nanosleep},
     open::{sys_creat, sys_open, sys_openat},
     pause::sys_pause,
+    pidfd_getfd::sys_pidfd_getfd,
     pidfd_open::sys_pidfd_open,
     pipe::{sys_pipe, sys_pipe2},
     poll::sys_poll,
@@ -407,6 +408,7 @@ impl_syscall_nums_and_dispatch_fn! {
     SYS_PIDFD_OPEN = 434       => sys_pidfd_open(args[..2]);
     SYS_CLONE3 = 435           => sys_clone3(args[..2], &user_ctx);
     SYS_CLOSE_RANGE = 436      => sys_close_range(args[..3]);
+    SYS_PIDFD_GETFD = 438      => sys_pidfd_getfd(args[..3]);
     SYS_FACCESSAT2 = 439       => sys_faccessat2(args[..4]);
     SYS_EPOLL_PWAIT2 = 441     => sys_epoll_pwait2(args[..5]);
     SYS_FCHMODAT2 = 452        => sys_fchmodat2(args[..4]);
--- a/kernel/src/syscall/mod.rs
+++ b/kernel/src/syscall/mod.rs
@@ -93,6 +93,7 @@ mod munmap;
 mod nanosleep;
 mod open;
 mod pause;
+mod pidfd_getfd;
 mod pidfd_open;
 mod pipe;
 mod poll;
--- a/test/initramfs/src/apps/scripts/process.sh
+++ b/test/initramfs/src/apps/scripts/process.sh
@@ -48,6 +48,7 @@ namespace/unshare
 process/group_session
 process/job_control
 process/pidfd
+process/pidfd_getfd
 process/wait4
 procfs/dentry_cache
 procfs/pid_mem
--- a/test/initramfs/src/syscall/ltp/testcases/all.txt
+++ b/test/initramfs/src/syscall/ltp/testcases/all.txt
@@ -1042,7 +1042,7 @@ pathconf01
 # personality02
 
 # pidfd_getfd01
-# pidfd_getfd02
+pidfd_getfd02
 
 pidfd_open01
 pidfd_open02


kernel/src/process/pid_file.rs
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use alloc::sync::Weak;
```
StevenJiang1110: This import is useless. `Weak` is already included by `crate::prelude::*`.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use crate::{
+    fs::file_table::{FdFlags, FileDesc, get_file_fast},
+    prelude::*,
+    process::{PidFile, posix_thread::AsPosixThread},
+    syscall::SyscallReturn,
+};
+
+pub fn sys_pidfd_getfd(
```
StevenJiang1110, line 10: Usually, when you adding a new syscall, you should also enable some test about it. Currently we have LTP and gVisor as two syscall test suites. For example, LTP has some test cases about `pidfd_getfd`, then you can enable them by uncommeting the corresponding test cases:

https://github.com/asterinas/asterinas/blob/3d27b376d7e9780aa2f8c75b0c472409b4cf529d/test/src/syscall/ltp/testcases/all.txt#L1014-L1015

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+) -> Result<SyscallReturn> {
+    // The flags argument is reserved for future use. Currently, it must be specified as 0.
+    if flags != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
+    // Todo: check permissions
+    // The calling process should PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
+    // over the process referred to by pidfd.
+    let mut file_table = ctx.thread_local.borrow_file_table_mut();
+    if let Some(pid_file) = get_file_fast!(&mut file_table, pidfd).downcast_ref::<PidFile>() {
```
StevenJiang1110: Excessive indentation for large code segments makes them hard to read. Use `let-else` instead to reduce nesting and keep the flow clear.

```rust
let Some(pid_file) = get_file_fast!(&mut file_table, pidfd).downcast_ref::<PidFile>() else {
    return_errno_with_message!(Errno::EINVAL, "not a pid file");
};

let target_file = {
       let process = pid_file.process_opt().ok_or_else(|| {
            Error::with_message(Errno::ESRCH, "the target process has been reaped")
        })?;
        // ......
```

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+        // get an existing file descriptor, targetfd,
+        // in the process referred to by the PID file descriptor pidfd.
+        let target_file = {
+            let process = pid_file.process_opt().ok_or_else(|| {
+                Error::with_message(Errno::ESRCH, "the target process has been reaped")
+            })?;
+            let main_thread = process.main_thread();
+            let file_table = main_thread
+                .as_posix_thread()
+                .ok_or(Errno::EBADF)?
```
StevenJiang1110: Using `unwrap` here. The main thread is always posix thread.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+    pidfd: FileDesc,
+    targetfd: FileDesc,
+    flags: u32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    // The flags argument is reserved for future use. Currently, it must be specified as 0.
+    if flags != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
+    // Todo: check permissions
```
StevenJiang1110:
```suggestion
    // TODO: Check permissions
```

Both TODO and FIXME must be uppercase, and the first letter following them should also be capitalized.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+    // The calling process should PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
```
StevenJiang1110:
```suggestion
    // The calling process should have PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
```

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+            })?;
+            let main_thread = process.main_thread();
+            let file_table = main_thread
+                .as_posix_thread()
+                .ok_or(Errno::EBADF)?
+                .file_table();
+            file_table
+                .lock()
+                .as_ref()
+                .ok_or(Errno::EBADF)?
```
StevenJiang1110:
```suggestion
                .ok_or_else(|| return_errno_with_message!(Errno::EBADF, "the target process has exited"))?
```

We'd prefer to include a descriptive error message instead of only return an error numer.

Have you verified the error code here? I'm not sure whether this value is correct if the process referenced by the pidfd has exited but hasn't been waited for (reaped) yet. Could you please write a test case to double-check this case?

kernel/src/process/pid_file.rs
```diff
@@ -125,18 +135,22 @@ impl FileLike for PidFile {
...
     fn poll(&self, mask: IoEvents, poller: Option<&mut PollHandle>) -> IoEvents {
-        self.process
-            .pidfile_pollee
-            .poll_with(mask, poller, || self.check_io_events())
+        match self.process.upgrade() {
+            Some(process) => process
+                .pidfile_pollee
+                .poll_with(mask, poller, || self.check_io_events()),
+            None => {
+                // target process has been reaped
```
StevenJiang1110:
```suggestion
                // The target process has been reaped.
```

All comments should be written in grammatically correct English and start with a capital letter. Please review other comments as well.

kernel/src/process/pid_file.rs
```diff
@@ -125,18 +135,22 @@ impl FileLike for PidFile {
         if fd_flags.contains(FdFlags::CLOEXEC) {
             flags |= CreationFlags::O_CLOEXEC.bits();
         }
+        let pid = self.process.upgrade().map_or(u32::MAX, |p| p.pid());
```
StevenJiang1110, line 133: Could you confirm if `u32::MAX` is indeed the correct default when the process has already been reaped? You can verify this by either referring to the Linux implementation or writing a test case.

StevenJiang1110 (commented):
Thanks for your contribution! The PR is in great shape. I've left a few comments, most of which are minor style issues that should be easy to fix.

Please also update the Linux capability documentation for pidfd_getfd under `book/src/kernel/linux-capability`.

Finally, please squash the last two commits into one. We prefer each commit to be atomic, so the license addition should be part of the commit that adds the pidfd_getfd syscall. And the first letter of commit message should also be capitalized.

kernel/src/process/pid_file.rs
li041: - linux use -1 to dump info for pidfd whose process has been reaped 
image

StevenJiang1110: Thanks for your investigation! Then it will also be better to include this hand-writing test under `test/src/apps`(perhaps `test/src/apps/process/pidfd.c`), if possible.

kernel/src/syscall/pidfd_getfd.rs
li041: The file table exists as long as the process isn't reaped, so `unwarp()` here is okk.

book/src/kernel/linux-compatibility/README.md
```diff
@@ -346,6 +346,7 @@ which are summarized in the table below.
 | 434     | pidfd_open             | ✅             | 💯 |
 | 435     | clone3                 | ✅             | [⚠️](syscall-flag-coverage/process-and-thread-management/#clone-and-clone3) |
 | 436     | close_range            | ✅             | 💯 |
+| 438     | pidfd_getfd            | ✅             | 💯 |
```
StevenJiang1110, line 349: You also need to update `book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/fully_covered.scml` to include `pidfd_getfd`.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // Get an existing file descriptor, targetfd, in the process referred to by the PID file descriptor pidfd.
+    let process = pid_file
+        .process_opt()
+        .ok_or_else(|| Error::with_message(Errno::ESRCH, "the target process has been reaped"))?;
+    let main_thread = process.main_thread();
+
+    let target_file_table = main_thread
+        .as_posix_thread()
+        .ok_or(Errno::EBADF)?
```
StevenJiang1110: Using `unwrap` here. `as_posix_thread` can never fail in this syscall.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,55 @@
...
+        .file_table();
+
+    let target_file = target_file_table
+        .lock()
+        .as_ref()
+        .ok_or(Errno::EBADF)?
```
StevenJiang1110: As you explained in https://github.com/asterinas/asterinas/pull/2877#discussion_r2689262133, here we should use unwrap? 

But I think you may want to double check the bahavior here, it may fail the test case I wrote below.

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,69 @@
...
+
+static int pidfd_getfd(int pidfd, int targetfd, unsigned int flags)
+{
+	return syscall(SYS_pidfd_getfd, pidfd, targetfd, flags);
+}
+
+FN_TEST(pidfd_getfd_valid)
+{
+	size_t size = sizeof(int);
+	int *shared_fd = CHECK(mmap(NULL, size, PROT_READ | PROT_WRITE,
```
StevenJiang1110: Use TEST_SUCC instead of CHECK. TEST_SUCC is always preferred over CHECK.
However, in certain cases—such as within a child process or inside FN_SETUP—TEST_SUCC may not correctly report results. In those situations only should CHECK be used.

Please fix other CHECK usages as well.

li041:
@boterinas rerun

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,69 @@
...
+	target_fd = TEST_SUCC(pidfd_getfd(pidfd, fd, 0));
+
+	char buffer[128] = { 0 };
+
+	TEST_RES(pread(target_fd, buffer, sizeof(buffer), 0),
+		 strcmp(buffer, "Test content\n") == 0);
+
+	CHECK(munmap(shared_fd, size));
+}
+END_TEST()
```
StevenJiang1110, line 72: Add more test here to check the result after child exits.

```C
    // Kill the child process and wait until it exits.
	TEST_SUCC(kill(pid, SIGKILL));
	usleep(100);

	TEST_ERRNO(pidfd_getfd(pidfd, fd, 0), ESRCH);
#define P_PIDFD 3
	TEST_SUCC(waitid(P_PIDFD, pidfd, NULL, WEXITED));
	TEST_ERRNO(pidfd_getfd(pidfd, fd, 0), ESRCH);
```

li041: In Linux, the fd_table is released when `do_exit` calls `exit_files`. When calling `pidfd_getfd` for a process that has exited but has not yet called that function, the operation can still succeed.

```c
// Ref: linux/kernel/pid.c::__pidfd_fget()
		 * It is possible that the target thread is exiting; it can be
		 * either:
		 * 1. before exit_signals(), which gives a real fd
		 * 2. before exit_files() takes the task_lock() gives a real fd
		 * 3. after exit_files() releases task_lock(), ->files is NULL;
		 *    this has PF_EXITING, since it was set in exit_signals(),
		 *    __pidfd_fget() returns EBADF.
		 * In case 3 we get EBADF, but that really means ESRCH, since
		 * the task is currently exiting and has freed its files
		 * struct, so we fix it up.
		 */
```

> ESRCH The process referred to by pidfd does not exist (i.e., it has terminated and has been waited on).

The second condition in the parentheses must also be met for ESRCH to be returned.

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,97 @@
...
+
+FN_TEST(pidfd_getfd_after_child_exits)
+{
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		exit(0);
+	}
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
```
StevenJiang1110, line 82: Add a `usleep(100)` or `sleep(1)` to wait for the child process to exit. The current implementation is racy—on some runs, line 78 may even execute before the child process has started. The test case below suffers from the same issue.

StevenJiang1110 (commented):
Looks fine to me! Just a small issue left. 

Could you also check why the [flock gVisor test](https://github.com/asterinas/asterinas/actions/runs/21018249190/job/60427608524?pr=2877) fails? Is it related to this PR? I haven't seen this error recently.

li041:
> Could you also check why the [flock gVisor test](https://github.com/asterinas/asterinas/actions/runs/21018249190/job/60427608524?pr=2877) fails? Is it related to this PR? I haven't seen this error recently.

I feel it sorry that I conclude from simple testcase that `pidfd_getfd` succeed before the target process is reaped. I referred to Linux code, and found it's just a special case: the target process has exited, but before `exit_files` is called. After `exit_files()`, it should return `EBADF`. After the process was reaped, it should return `ESRCH` finally.

li041:
@boterinas rerun

kernel/src/process/pid_file.rs
```diff
@@ -56,19 +57,27 @@ impl PidFile {
...
+        match self.process.upgrade() {
+            Some(process) => {
+                if process.status().is_zombie() {
+                    IoEvents::IN
+                } else {
+                    IoEvents::empty()
+                }
+            }
+            None => {
+                // The target process has been reaped.
```
lrh2000: In `pid_file.rs`, I don't think we have a process as a "target". There is no target process when someone calls `self.check_io_events()`, it's just the process associated with the PID file (i.e., `self`).

It perhaps makes some sense to say "the target process" in the system call files because a system call is a concrete action that has a target (i.e., its arguments).

---

Also prefer to early return. I believe it's clear for two reasons:
 1. The error (`None`) case is closer to the operation (`upgrade()`).
 2. We can avoid nested indentation.
```rust
let Some(process) = self.process.upgrade() else {
    // The process has been reaped.
    return IoEvents::IN;
};
if process.status().is_zombie() {
    IoEvents::IN
} else {
    IoEvents::empty()
}
```

kernel/src/process/pid_file.rs
```diff
@@ -125,18 +134,22 @@ impl FileLike for PidFile {
...
     fn poll(&self, mask: IoEvents, poller: Option<&mut PollHandle>) -> IoEvents {
-        self.process
-            .pidfile_pollee
-            .poll_with(mask, poller, || self.check_io_events())
+        match self.process.upgrade() {
+            Some(process) => process
+                .pidfile_pollee
+                .poll_with(mask, poller, || self.check_io_events()),
+            None => {
+                // The target process has been reaped.
```
lrh2000: Same. And also use early return here, mostly for consistency with the previous code.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,57 @@
...
+    flags: u32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    // The flags argument is reserved for future use. Currently, it must be specified as 0.
+    if flags != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
+    // TODO: Check permissions
+    // The calling process should have PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
+    // over the process referred to by pidfd.
```
lrh2000: Don't group unrelated items together. Blank lines are a good tool to group related items together.

Besides. please quote variable names inside `` ` ` ``, like `` `flags` ``, `` `targetfd` ``, e.t.c.

I hope this helps to understand:
```patch
diff --git a/kernel/src/syscall/pidfd_getfd.rs b/kernel/src/syscall/pidfd_getfd.rs
index bd61a148b..2ea698113 100644
--- a/kernel/src/syscall/pidfd_getfd.rs
+++ b/kernel/src/syscall/pidfd_getfd.rs
@@ -13,27 +13,29 @@ pub fn sys_pidfd_getfd(
     flags: u32,
     ctx: &Context,
 ) -> Result<SyscallReturn> {
-    // The flags argument is reserved for future use. Currently, it must be specified as 0.
+    // The `flags` argument is reserved for future use. Currently, it must be specified as 0.
     if flags != 0 {
         return_errno_with_message!(Errno::EINVAL, "invalid flags");
     }
-    // TODO: Check permissions
+
+    // TODO: Check permissions.
     // The calling process should have PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
-    // over the process referred to by pidfd.
+    // over the process referred to by `pidfd`.
+
     let file = get_file_fast!(&mut file_table, pidfd);
     let Some(pid_file) = file.downcast_ref::<PidFile>() else {
         return_errno_with_message!(Errno::EINVAL, "not a pid file");
     };
 
-    // Get an existing file descriptor, targetfd, in the process referred to by the PID file descriptor pidfd.
     let process = pid_file
         .process_opt()
         .ok_or_else(|| Error::with_message(Errno::ESRCH, "the target process has been reaped"))?;
     let main_thread = process.main_thread();
 
+    // Get the file description corresponding to the file descriptor `targetfd` in the process
+    // referred to by the PID file.
     let target_file_table = main_thread.as_posix_thread().unwrap().file_table();
-
     let target_file = target_file_table
         .lock()
         .as_ref()
```

Also note `target_file` is a file **description**. It is not a file **descriptor**. So I rewrite some comments to correct the concept.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut file_table = ctx.thread_local.borrow_file_table_mut();
+    let file = get_file_fast!(&mut file_table, pidfd);
+    let Some(pid_file) = file.downcast_ref::<PidFile>() else {
+        return_errno_with_message!(Errno::EINVAL, "not a pid file");
```
lrh2000:
```suggestion
        return_errno_with_message!(Errno::EINVAL, "the file is not a PID file");
```

Following the existing wording:
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/process/process_filter.rs#L40-L42

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let target_file_table = main_thread.as_posix_thread().unwrap().file_table();
+
+    let target_file = target_file_table
+        .lock()
+        .as_ref()
+        .ok_or_else(|| {
+            Error::with_message(
+                Errno::EBADF,
+                "the target process has no file descriptor table",
```
lrh2000:
```suggestion
                "the target process has exited",
```

Following the existing wording:
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/fs/procfs/pid/task/fd.rs#L226-L228

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,99 @@
...
+}
+END_TEST()
+
+FN_TEST(pidfd_getfd_errnos)
+{
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		exit(0);
+	}
+	sleep(1); // Ensure child has exited
```
lrh2000:
```suggestion
	TEST_SUCC(waitpid(pid, NULL, WNOWAIT)); // Ensure the child has exited
```

I think we support `WNOWAIT` now so it's more reliable, efficient, and elegant than `sleep()`?

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,99 @@
...
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+	TEST_ERRNO(pidfd_getfd(invalid_pidfd, fd, 0), EBADF);
```
lrh2000, line 102: Insert a blank line after line 94.

There is a blank line after line 79 in `pidfd_getfd_after_child_exits`. Why is there no such blank line in `pidfd_getfd_errnos`? We should be consistent, which would be helpful since the first code blocks in the two functions are identical.

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,99 @@
...
+	TEST_ERRNO(pidfd_getfd(pidfd, -1, 0), EBADF);
+	TEST_ERRNO(pidfd_getfd(pidfd, fd, 1), EINVAL);
+}
```
lrh2000, line 108: Always perform the cleanup. In a separate code block:
```c
    TEST_SUCC(waitpid(pid, NULL, 0));
    TEST_SUCC(close(pidfd));
```

The same applies for `pidfd_getfd_after_child_exits`:
```c
    TEST_SUCC(close(pidfd));
```

And also for `pidfd_getfd_valid`.

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,99 @@
...
+		*shared_fd = fd;
+
+		pause();
+		exit(0);
+	}
+
+	while (*shared_fd == 0) {
+		usleep(100);
+	}
+
```
lrh2000: - Group related items to one code block.
 - Use `volatile int *` as `shared_fd`.
 - Clean up after everything is done.
```patch
diff --git a/test/initramfs/src/apps/process/pidfd_getfd.c b/test/initramfs/src/apps/process/pidfd_getfd.c
index 95015cb5b..0f1233135 100644
--- a/test/initramfs/src/apps/process/pidfd_getfd.c
+++ b/test/initramfs/src/apps/process/pidfd_getfd.c
@@ -34,16 +34,14 @@ static int pidfd_getfd(int pidfd, int targetfd, unsigned int flags)
 FN_TEST(pidfd_getfd_valid)
 {
 	size_t size = sizeof(int);
-	int *shared_fd = TEST_SUCC(mmap(NULL, size, PROT_READ | PROT_WRITE,
+	volatile int *shared_fd = TEST_SUCC(mmap(NULL, size, PROT_READ | PROT_WRITE,
 					MAP_SHARED | MAP_ANONYMOUS, -1, 0));
 
 	pid = TEST_SUCC(fork());
-
 	if (pid == 0) {
 		fd = CHECK(open(TESTFILE, O_CREAT | O_RDWR | O_TRUNC, 0644));
 
 		CHECK_WITH(write(fd, "Test content\n", 13), _ret == 13);
-
 		*shared_fd = fd;
 
 		pause();
@@ -51,21 +49,24 @@ FN_TEST(pidfd_getfd_valid)
 	}
 
 	while (*shared_fd == 0) {
-		usleep(100);
+		sched_yield();
 	}
-
 	fd = *shared_fd;
 
 	pidfd = TEST_SUCC(pidfd_open(pid, 0));
-
 	target_fd = TEST_SUCC(pidfd_getfd(pidfd, fd, 0));
 
 	char buffer[128] = { 0 };
-
 	TEST_RES(pread(target_fd, buffer, sizeof(buffer), 0),
 		 strcmp(buffer, "Test content\n") == 0);
 
 	TEST_SUCC(munmap(shared_fd, size));
+	TEST_SUCC(close(pidfd));
+	TEST_SUCC(close(target_fd));
+
+	TEST_SUCC(kill(pid, SIGKILL));
+	TEST_SUCC(waitpid(pid, NULL, 0));
+
 }
 END_TEST()
 
@@ -96,4 +97,4 @@ FN_TEST(pidfd_getfd_errnos)
 	TEST_ERRNO(pidfd_getfd(pidfd, -1, 0), EBADF);
 	TEST_ERRNO(pidfd_getfd(pidfd, fd, 1), EINVAL);
 }
-END_TEST()
\ No newline at end of file
+END_TEST()
```

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,57 @@
...
+    targetfd: FileDesc,
+    flags: u32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    // The flags argument is reserved for future use. Currently, it must be specified as 0.
+    if flags != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
+    // TODO: Check permissions
+    // The calling process should have PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
```
lrh2000, line 35: For the permission check, I think a better strategy than "leaving a TODO and doing nothing" is to "perform the strictest check and leave a TODO that says it can be relaxed in the future".

Please read the "/proc/sys/kernel/yama/ptrace_scope" section in "https://man7.org/linux/man-pages/man2/ptrace.2.html".

The easiest (but still reasonable) option for us to implement is the following:
>       2 ("admin-only attach")
>              Only processes with the CAP_SYS_PTRACE capability in the
>              user namespace of the target process may perform
>              PTRACE_MODE_ATTACH operations or trace children that employ
>              PTRACE_TRACEME.

An example of performing the capability check within the target process's user namespace can be found here:
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/process/kill.rs#L163-L170

test/initramfs/src/apps/process/pidfd_getfd.c
li041: See https://github.com/asterinas/asterinas/pull/2877#discussion_r2692659199 for detalis. In this testcase, we should ensure that the process has exited and not yet reaped.

kernel/src/syscall/pidfd_getfd.rs
li041: See https://github.com/asterinas/asterinas/pull/2877#discussion_r2692659199 for detalis. Maybe `has no file descritor table` is more precise.

test/initramfs/src/apps/process/pidfd_getfd.c
lrh2000: https://man7.org/linux/man-pages/man2/wait.2.html

>       WNOWAIT
>              Leave the child in a waitable state; a later wait call can
>              be used to again retrieve the child status information.

I think WNOHWAIT means to wait for the child process but not to reap it. It seems to do what you want here.

However, it appears that it can only be used with `waitid` (instead of `waitpid`). Therefore, it's `waitid(P_PID, pid, NULL, WNOWAIT)`.

See also the implementation in our kernel:
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/process/wait.rs#L106-L107

kernel/src/syscall/pidfd_getfd.rs
lrh2000: https://github.com/asterinas/asterinas/pull/2877#discussion_r2692659199 also states that we encounter the "no file table" case for an exiting or exited process?

Currently, we need to access the remote file table or the remote VMAR (i.e., Linux's MM structure) in various procfs files. If we find that the struct does not exist, we conclude that the process has exited.
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/fs/procfs/pid/task/maps.rs#L35-L37
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/fs/procfs/pid/task/fd.rs#L226-L228
https://github.com/asterinas/asterinas/blob/815cbf061b6d8ef0871f47f6237dda4eb32fe0d6/kernel/src/syscall/setns.rs#L74-L76

I think "has no file descriptor file" refers to the kernel's implementation details. However, the real cause of a process having no file table is the user exiting the process. The latter seems more meaningful here and is consistent with the existing code. As long as no other cause can make a POSIX process have no file table, they are equally precise.

(For the MM case, "the process has no MM structure" or "the process has no VMAR" sound even more odd to me because "MM structure" and "VMAR" are kernel-internal concepts.)

kernel/src/syscall/pidfd_getfd.rs
StevenJiang1110: > For the permission check, I think a better strategy than "leaving a TODO and doing nothing" is to "perform the strictest check and leave a TODO that says it can be relaxed in the future".

This is reasonable. I think we can enforce this rule—especially if a stricter check is feasible.

test/initramfs/src/apps/process/pidfd_getfd.c
lrh2000: Sorry, ignore the `usleep(100)` -> `sched_yield()` change here. There are some known issues with CFS schedulers where `sched_yield()` may not actually release the CPU. I forgot about this issue when writing this...

kernel/src/process/pid_file.rs
lrh2000: Please fix https://github.com/asterinas/asterinas/pull/2877#discussion_r2693978913 and https://github.com/asterinas/asterinas/pull/2877#discussion_r2693984955 in the correct commit, i.e, `` Refactor `PidFile` to use `Weak<Process>` ``. We should use `git rebase` to edit the right commit instead of squashing all new changes to the latest commit, which will keep the commit logs clean.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,68 @@
...
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
+
+    // The calling process should have PTRACE_MODE_ATTACH_REALCREDS permissions (see ptrace(2))
+    // over the process referred to by `pidfd`.
+    // Currently, this is implemented as requiring the calling process to have the
+    // CAP_SYS_PTRACE capability, which is stricter.
+    // TODO: Implement appropriate PTRACE_MODE_ATTACH_REALCREDS permission check.
+    if ctx
+        .process
```
lrh2000: `process.user_ns().check_cap(..., posix_thread)` means `posix_thread` to have the required capability in the `process`'s user namespace.

But read this again:
> Only processes with the CAP_SYS_PTRACE capability in the
> **user namespace of the target process** may perform
> PTRACE_MODE_ATTACH operations or trace children that employ
> PTRACE_TRACEME.


This needs to be moved below. Before line 50:
```c
process.user_ns().lock().check_cap(CapSet::SYS_PTRACE, ctx.posix_thread)?;
```

"process" is the PID file's associated process.

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,105 @@
...
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+	target_fd = TEST_SUCC(pidfd_getfd(pidfd, fd, 0));
+
+	char buffer[128] = { 0 };
+	TEST_RES(pread(target_fd, buffer, sizeof(buffer), 0),
+		 strcmp(buffer, "Test content\n") == 0);
+
+	TEST_SUCC(munmap((void *)shared_fd, size));
+	TEST_SUCC(close(target_fd));
```
lrh2000, line 65:
```suggestion
	TEST_SUCC(close(pidfd));
	TEST_SUCC(close(target_fd));
	TEST_SUCC(unlink(TESTFILE));
```

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,105 @@
...
+}
+END_TEST()
+
+FN_TEST(pidfd_getfd_after_child_exits)
+{
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		exit(0);
+	}
+	TEST_SUCC(waitpid(pid, NULL, WNOWAIT)); // Ensure the child has exited
```
lrh2000: https://github.com/asterinas/asterinas/pull/2877#discussion_r2696570152:
> However, it appears that it can only be used with `waitid` (instead of `waitpid`). Therefore, it's `waitid(P_PID, pid, NULL, WNOWAIT)`.

I'm surprised that this succeeds in our kernel. It seems to be a bug. In Linux, this won't succeed:
```
test_pidfd_getfd_after_child_exits: `waitpid(pid, ((void *)0), 0x01000000)` failed [got Invalid argument, but expected Success]
```

Something like this is needed instead:
```c
TEST_SUCC(waitid(P_PID, pid, NULL, WEXITED | WNOWAIT)); // Ensure the child has exited
```

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,105 @@
...
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+
+	TEST_ERRNO(pidfd_getfd(pidfd, fd, 0), EBADF);
```
lrh2000: Did you really try to run the test in Linux yourself? This won't pass:
```
test_pidfd_getfd_after_child_exits: `pidfd_getfd(pidfd, fd, 0)` failed [got No such process, but expected Bad file descriptor]
```

It seems to be related to the case where the file table is missing. Following the code you quoted below, we should return `ESRCH`.
```c
		 /*
		 * [ .. ]
		 * In case 3 we get EBADF, but that really means ESRCH, since
		 * the task is currently exiting and has freed its files
		 * struct, so we fix it up.
		 */
		if (task->flags & PF_EXITING)
			file = ERR_PTR(-ESRCH);
		else
			file = ERR_PTR(-EBADF);
```

The "else" branch (the `EBADF` case) is likely for kernel threads. Currently, all our system calls do not work well with kernel threads (because we manage a table for all POSIX processes, and kernel threads do not exist in the table at all), so it's a separate issue anyway.

kernel/src/syscall/pidfd_getfd.rs
```diff
@@ -0,0 +1,68 @@
...
+pub fn sys_pidfd_getfd(
+    pidfd: FileDesc,
+    targetfd: FileDesc,
+    flags: u32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    // The `flags` argument is reserved for future use. Currently, it must be specified as 0.
+    if flags != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+    }
```
lrh2000, line 19: Also most system calls have a `debug!` statement after some basic flag parse/check. Add it here as well.

`pidfd_open`:
https://github.com/asterinas/asterinas/blob/8e6406ae352c402ea4f65517bdb7d290d6c09b2f/kernel/src/syscall/pidfd_open.rs#L12-L15

test/initramfs/src/apps/process/pidfd_getfd.c
```diff
@@ -0,0 +1,108 @@
...
+	pidfd = TEST_SUCC(pidfd_open(pid, 0));
+	target_fd = TEST_SUCC(pidfd_getfd(pidfd, fd, 0));
+
+	char buffer[128] = { 0 };
+	TEST_RES(pread(target_fd, buffer, sizeof(buffer), 0),
+		 strcmp(buffer, "Test content\n") == 0);
+
+	TEST_SUCC(munmap((void *)shared_fd, size));
+	TEST_SUCC(close(target_fd));
+	TEST_SUCC(close(pidfd));
```
lrh2000, line 66: https://github.com/asterinas/asterinas/pull/2877#discussion_r2696737889:
```c
	TEST_SUCC(unlink(TESTFILE));
```

Delete the test file.

PR #2882, opened 2026-01-14
Remove `init_stack` dead code and clean it up

lrh2000:
1. The first commit removes `init_stack` dead code (e.g., `InitStackReader::argc()`) and cleans up the comment style.

https://github.com/asterinas/asterinas/blob/738d14f3c033e8dc26b8d374872e5bd515e7a1fc/kernel/src/process/process_vm/init_stack/mod.rs#L414-L415

 2. The second commit makes `AuxVec::set` infallible by removing `AT_NULL` from `AuxKey`. `AT_IGNORE` should not have any unexpected side effects, so it can be allowed. However, setting it may not have a real effect. If we really want to avoid it, we can do something similar to what we did with `AT_NULL`.

https://github.com/asterinas/asterinas/blob/738d14f3c033e8dc26b8d374872e5bd515e7a1fc/kernel/src/process/process_vm/init_stack/aux_vec.rs#L74-L76

 3. The third commit fixes a small bug that I did not notice when merging https://github.com/asterinas/asterinas/pull/2736. We need to check `heap_end` is a user space address to satisfy the `VmarMapOptions::offset` requirement. `heap_start` is a user space address, but `heap_end` may no longer be a user space address because we added several `PAGE_SIZE`.

https://github.com/asterinas/asterinas/blob/738d14f3c033e8dc26b8d374872e5bd515e7a1fc/kernel/src/vm/vmar/vmar_impls/map.rs#L152-L153

### DIFF
--- a/kernel/src/process/process_vm/heap.rs
+++ b/kernel/src/process/process_vm/heap.rs
@@ -8,7 +8,10 @@ use crate::{
     prelude::*,
     process::ResourceType,
     util::random::getrandom,
-    vm::{perms::VmPerms, vmar::Vmar},
+    vm::{
+        perms::VmPerms,
+        vmar::{VMAR_CAP_ADDR, Vmar},
+    },
 };
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


StevenJiang1110 (approved):

PR #2887, opened 2026-01-14
Refactor `Metadata`'s fields and fix pseudo-filesystems' Device ID

LordaeronESZ:
Required by #2871 

This PR includes:
- Refactor `Metadata`'s fields and document them.
- Change `MajorId`'s lower bound to 0, which is used to represents a invalid or absent device (pseudo filesystems).

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/template/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/template/mod.rs
@@ -34,7 +34,7 @@ struct Common {
 }
 
 impl Common {
-    pub fn new(metadata: Metadata, fs: Weak<dyn FileSystem>, is_volatile: bool) -> Self {
+    fn new(metadata: Metadata, fs: Weak<dyn FileSystem>, is_volatile: bool) -> Self {
         Self {
             metadata: RwLock::new(metadata),
             extension: Extension::new(),
@@ -43,82 +43,82 @@ impl Common {
         }
     }
 
-    pub fn fs(&self) -> Arc<dyn FileSystem> {
+    fn fs(&self) -> Arc<dyn FileSystem> {
         self.fs.upgrade().unwrap()
     }
 
-    pub fn metadata(&self) -> Metadata {
+    fn metadata(&self) -> Metadata {
         *self.metadata.read()
     }
 
-    pub fn ino(&self) -> u64 {
+    fn ino(&self) -> u64 {
         self.metadata.read().ino
     }
 
-    pub fn type_(&self) -> InodeType {
+    fn type_(&self) -> InodeType {
         self.metadata.read().type_
     }
 
-    pub fn size(&self) -> usize {
+    fn size(&self) -> usize {
         self.metadata.read().size
     }
 
-    pub fn atime(&self) -> Duration {
+    fn atime(&self) -> Duration {
         self.metadata.read().last_access_at
     }
 
-    pub fn set_atime(&self, time: Duration) {
+    fn set_atime(&self, time: Duration) {
         self.metadata.write().last_access_at = time;
     }
 
-    pub fn mtime(&self) -> Duration {
+    fn mtime(&self) -> Duration {
         self.metadata.read().last_modify_at
     }
 
-    pub fn set_mtime(&self, time: Duration) {
+    fn set_mtime(&self, time: Duration) {
         self.metadata.write().last_modify_at = time;
     }
 
-    pub fn ctime(&self) -> Duration {
+    fn ctime(&self) -> Duration {
         self.metadata.read().last_meta_change_at
     }
 
-    pub fn set_ctime(&self, time: Duration) {
+    fn set_ctime(&self, time: Duration) {
         self.metadata.write().last_meta_change_at = time;
     }
 
-    pub fn mode(&self) -> Result<InodeMode> {
+    fn mode(&self) -> Result<InodeMode> {
         Ok(self.metadata.read().mode)
     }
 
-    pub fn set_mode(&self, mode: InodeMode) -> Result<()> {
+    fn set_mode(&self, mode: InodeMode) -> Result<()> {
         self.metadata.write().mode = mode;
         Ok(())
     }
 
-    pub fn owner(&self) -> Result<Uid> {
+    fn owner(&self) -> Result<Uid> {
         Ok(self.metadata.read().uid)
     }
 
-    pub fn set_owner(&self, uid: Uid) -> Result<()> {
+    fn set_owner(&self, uid: Uid) -> Result<()> {
         self.metadata.write().uid = uid;
         Ok(())
     }
 
-    pub fn group(&self) -> Result<Gid> {
+    fn group(&self) -> Result<Gid> {
         Ok(self.metadata.read().gid)
     }
 
-    pub fn set_group(&self, gid: Gid) -> Result<()> {
+    fn set_group(&self, gid: Gid) -> Result<()> {
         self.metadata.write().gid = gid;
         Ok(())
     }
 
-    pub fn is_volatile(&self) -> bool {
+    fn is_volatile(&self) -> bool {
         self.is_volatile
     }
 
-    pub fn extension(&self) -> &Extension {
+    fn extension(&self) -> &Extension {
         &self.extension
     }
 }


kernel/libs/device-id/src/lib.rs
```diff
@@ -92,10 +103,13 @@ const MAX_MINOR_ID: u32 = 0x000f_ffff;
...
 ///
-/// A major ID is a non-zero, 12-bit integer, thus falling in the range of `1..(1u16 << 12)`.
+/// A major ID is a 12-bit integer, thus falling in the range of `0..(1u16 << 12)`.
+///
+/// - **0**: Represents an invalid or absent device (used for pseudo filesystems)
+/// - **1-4095**: Valid major device numbers
 ///
 /// Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/linux/kdev_t.h#L10>.
-pub type MajorId = RangedU16<1, MAX_MAJOR_ID>;
+pub type MajorId = RangedU16<0, MAX_MAJOR_ID>;
```
tatetian, line 120: The current code of `DeviceId` has made it quite clear that a design decision is made to forbid the "null" major ID. This allows kernel developers to express at the type level whether a value is guaranteed to refer to a valid device (`DeviceId`) or may refer to a valid or null device (`Option<DeviceId>`). Please do not rollback a design decision unless you really understand what you are doing.

Your code suggests that you believe a null device ID is assigned to pseudo files. Is this by speculation or experiments or studying the source code? To show that your understanding matches Linux's behavior, new [tests for pseudo files](https://github.com/asterinas/asterinas/tree/main/test/initramfs/src/apps/pseudofs) should be added to ensure that the `dev` values of different pseudo files on Asterinas match those on Linux.

If Linux does assign a null device ID to pseudo files, then the type of the `Metadata::container_dev_id` field should be changed to `Option<DeviceId>`, rather than allowing `DeviceId` to have a value of 0. The document should be updated accordingly.

If Linux assigns a valid device ID to pseudo files, study Linux source code to understand the strategy for assigning these pseudo device IDs and implement the same strategy in Asterinas.

tatetian (commented):
Thanks for the refactoring.

kernel/libs/device-id/src/lib.rs
LordaeronESZ: Ok, I'll dive into it.

In current PR, I just performed a simple "translation" work. For instance, when `dev` value in original code is zero (maybe not accurate), I translate it to `DeviceID(0)`.

LordaeronESZ:
In Linux, when mounting a file system, the `st_dev` field is retrieved depending on the file system type.

- For disk file systems (such as Ext4), it locates the VFS inode structure of the corresponding device file, reads the corresponding `st_rdev` field from it, and copies it to the `st_dev` field of the VFS superblock of current mounting file system.

- For pseudo file-systems (such as ramfs), the major number of `st_dev` is fixed at 0, while the minor number is dynamically allocated using the ID Allocator in the Linux kernel, and then stored in the `st_dev` field of the VFS superblock.

In current revised version of PR. I implement a simple allocator named `PseudoFSDeviceIdAllocator` for dynamically allocating device ID of pseudo file systems.

---

However, none of the current Asterinas file system would do "create and store `st_dev`" during mounting process.

- For disk file systems (such as Ext2), they dynamically access the `id` method of their correspoding block device data structure when metadata is needed.

https://github.com/asterinas/asterinas/blob/2219e83f8e13e3f139b1120e8f46e76da61bcec1/kernel/src/fs/ext2/inode.rs#L102-L121

- For pseudo file-systems (such as ramfs), they simply set `dev` field to 0 when metadata is needed. This causes the device ID to always be `0,0` when `stat`ing files under these file systems, which is inconsistent with the behavior in Linux, where the device ID is `0,x` (where x is an integer greater than 0).

https://github.com/asterinas/asterinas/blob/2219e83f8e13e3f139b1120e8f46e76da61bcec1/kernel/src/fs/ramfs/fs.rs#L1169-L1188

---

If such a mechanism of creation and storage of `dev` is required, then expanding the superblock might be necessary.

LordaeronESZ:
I'm not sure what the exact definition of "pseudo-filesystem" is in Asterinas. Based on the `kernel/src/fs/pseudofs.rs` code, it appears to represent **a specific file system** used to manage pseudo inodes that cannot exist in the directory tree. If that's the case, then in this PR, I called **file systems without backing storage devices** (such as procfs and ramfs) pseudo-filesystem seems inappropriate.

---

Even in Linux, this term seems to be context-dependent.

Sometimes it's considered a file system where data resides in memory ([reference](https://elixir.bootlin.com/linux/v6.18/source/Documentation/filesystems/vfs.rst#L48-L49)), as in this PR.
Other times it's considered one that cannot be mounted by the user ([reference](https://elixir.bootlin.com/linux/v6.18/source/fs/libfs.c#L717-L720)), similar to `kernel/src/fs/pseudofs.rs` code does.

LordaeronESZ:
I have no idea why ktest in CI/CD was failed, and I ran it in my local environment and checked the logs:

```
...
[caught panic] Panicked at /root/asterinas/kernel/src/fs/ramfs/fs.rs:56:14
PSEUDO_FS_DEVICE_ID_ALLOCATOR not initialized

---- /root/asterinas/kernel/src/fs/overlayfs/fs.rs:1423:4 - opaque_dir ----

[caught panic] Panicked at /root/asterinas/kernel/src/fs/ramfs/fs.rs:56:14
PSEUDO_FS_DEVICE_ID_ALLOCATOR not initialized

---- /root/asterinas/kernel/src/fs/overlayfs/fs.rs:1408:4 - whiteout_file ----

[caught panic] Panicked at /root/asterinas/kernel/src/fs/ramfs/fs.rs:56:14
PSEUDO_FS_DEVICE_ID_ALLOCATOR not initialized
...
```

It showed that `PSEUDO_FS_DEVICE_ID_ALLOCATOR` was not initialized, is that because ktest is done before the component is initialized (`kernel/src/fs/mod.rs:init`)?

kernel/libs/device-id/src/lib.rs
```diff
@@ -13,6 +13,61 @@
 #![deny(unsafe_code)]
 
 use aster_util::ranged_integer::{RangedU16, RangedU32};
+use id_alloc::IdAlloc;
+use spin::{Mutex, Once};
```
tatetian: Asterinas only uses `Once` from `spin`. All lock primitives are provided by OSTD.

kernel/libs/device-id/src/lib.rs
```diff
@@ -13,6 +13,61 @@
...
+
+pub struct PseudoFSDeviceIdAllocator {
```
tatetian: According to Asterinas's established convention, the name should use `Fs` rather than `FS`.

kernel/libs/device-id/src/lib.rs
tatetian: Ok. I see. Linux reserves the major ID 0 for unnamed devices. So we should adjust the lower bound to 0.

kernel/libs/device-id/src/lib.rs
```diff
@@ -92,10 +152,13 @@ const MAX_MINOR_ID: u32 = 0x000f_ffff;
 
 /// The major component of a device ID.
 ///
-/// A major ID is a non-zero, 12-bit integer, thus falling in the range of `1..(1u16 << 12)`.
+/// A major ID is a 12-bit integer, thus falling in the range of `0..(1u16 << 12)`.
+///
+/// - **0**: Represents an invalid or absent device (used for pseudo filesystems)
+/// - **1-4095**: Valid major device numbers
```
tatetian: A more accurate description:

```diff
- /// - **0**: Represents an invalid or absent device (used for pseudo filesystems) 
- /// - **1-4095**: Valid major device numbers
+ /// The major ID 0 is reserved to represent an invalid device (when paired with a minor ID of 0) or an anonymous device (used for pseudo filesystems).
```

kernel/libs/device-id/src/lib.rs
```diff
@@ -38,6 +93,11 @@ impl DeviceId {
     pub fn minor(&self) -> MinorId {
         MinorId::new(self.0 & 0xf_ffff)
     }
+
+    /// Checks if the container device is valid (major != 0).
+    pub fn has_valid_container(&self) -> bool {
```
tatetian: Given the addition of this method and `PseudoFSDeviceIdAllocator` to this file, it looks that your mental model about how Asterinas manages `DeviceId`s is misguided.

First, you need to realize that **device IDs of char and block devices are two separate namespaces**. For example, both memory char devices and memory disks are assigned a major device ID of 1. You can check out [this table](https://www.kernel.org/doc/Documentation/admin-guide/devices.txt) for more information about the assignment of device IDs.

For this reason, Asterinas has three different modules regarding device ID management:
* The `device-id` crate, which defines the `DeviceId`, `MajorId`, `MinorId` types.
* The `device_id` module of the `block` component, which provides the `acquire_major` and `EXTENDED_DEVICE_ID_ALLOCATOR` API for allocating device IDs of _block_ devices.
* The `device::registry::char` module in the kernel, which provides the `acquire_major` API for allocating device IDs of _char_ devices.

According to this arrangement on separating the responsibilities of device ID management, we can easily conclude that **any code related to unnamed devices for pseudo FSes should NOT appear in the `device-id` crate.**

The name `has_valid_container` clearly refers to a concept that is related to block devices and FSes. But the ability to check whether device ID is zero could be handy if such a method is provided on `DeviceId`. So we should add this ability without referring to block device or FS-related concepts.

Here is my suggestion:

```rust
impl DeviceId {
    /// Returns whether the device ID refers to a null device.
    ///
    /// A null device takes the value of 0 for both the major and minor IDs.
    pub fn is_null(self) -> bool { todo!() }

    /// Returns whether the device ID refers to an anonymous or unnamed device.
    ///
    /// An anonymous device has the value of 0 as the major ID but a non-zero value for the minor ID.
    pub fn is_anonymous(self) -> bool { todo!() }
}
```

As to `PseudoFSDeviceIdAllocator`, move it to the `pseudofs` module in the kernel crate.

kernel/libs/device-id/src/lib.rs
```diff
@@ -13,6 +13,61 @@
...
+
+    /// Allocate a device ID for pseudo filesystems.
+    /// Returns `None` if minor number allocation fails (exhausted).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the allocator was not initialized.
+    pub fn allocate(&self) -> DeviceId {
+        let major = MajorId::new(0);
+        let minor = self.minor_allocator.lock().alloc().unwrap() as u32;
```
tatetian: Do not `unwrap` here. Change the return type to `Option<DeviceId>` and check if the `alloc` method returns `None`.

kernel/src/fs/utils/inode.rs
```diff
@@ -124,101 +125,177 @@ impl From<AccessMode> for Permission {
...
     pub gid: Gid,
-    pub rdev: u64,
+
+    /// The ID of the device containing the inode.
+    ///
+    /// For persisted files, this device could be a on-disk partition or a logical volume (with RAID).
+    /// For pseudo files (e.g., those on sockfs), this device is also "pseudo".
+    ///
+    /// Corresponds to `st_dev`.
+    pub container_dev_id: Option<DeviceId>,
```
tatetian: This field should be changed to `DeviceId`

tatetian:
> I'm not sure what the exact definition of "pseudo-filesystem" is in Asterinas

This is a good question. Broadly speaking, pseudo FSes means FSes without physical storage. This includes ramfs, procfs, sysfs, anonyfs, sockfs, etc. My understanding aligns with yours.

> Based on the `kernel/src/fs/pseudofs.rs` code, it appears to represent **a specific file system**

Yes. This module contains a few specific implementations of pseudo FSes. I think we can clarify the concept by adding some module-level documentation and reorganize the file structure.

Edit the `pseudofs.rs` file this way:

```rust
//! The infrastructure of pseudo FSes along with some naive implementations of pseudo FSes.
//!
//! Pseudo FSes are those that have no physical storage.
//! Some pseudo FSes are trivial,
//! e.g., [`PipeFs`], [`SockFs`], [`PidfdFs`], and [`AnonInodeFs`].
//! Such trivial pseudo FSes are directly implemented in this module
//! using a generic pseudo FS implementation called [`NaivePseudoFs`].
//!
//! On the other hand, some pseudo FSes are quite complex,
//! including [`RamFs`], [`ProcFs`], and [`SysFs`]. Just to name a few.
//! They are implemented in their own modules.
//!
//! Pseudo FSes are not backed by physical storage devices,
//! but they still need to be assigned container device IDs
//! (see [`Metadata::container_dev_id`]).
//! [Linux's behavior](https://github.com/torvalds/linux/blob/dc855b77719fe452d670cae2cf64da1eb51f16cc/fs/super.c#L1257)
//! is to assign _anonymous device IDs_ (whose major IDs are 0).
//! As such, this module provides a device ID allocator called [`DEVICE_ID_ALLOCATOR`]
//! to allocate and free anonymous device IDs.
//!
//! [`RamFs`]: crate::fs::ramfs::RamFs
//! [`ProcFs`]: crate::fs::procfs::ProcFs
//! [`SysFs`]: crate::fs::sysfs::SysFs
```

According to this documentation, you need to migrate `PseudoFSDeviceIdAllocator` to this module and rename it. Add a sub-module for it so that the code is clearly separated from the rest of this module.

kernel/src/fs/devpts/mod.rs
```diff
@@ -196,7 +202,16 @@ impl Inode for RootInode {
...
 
     fn metadata(&self) -> Metadata {
-        *self.metadata.read()
+        let metadata = *self.metadata.read();
+        if metadata.container_dev_id.is_null()
+            && let Some(devpts) = self.fs.upgrade()
+        {
+            let dev_id = devpts.sb().dev_id;
+            let mut metadata_lock = self.metadata.write();
+            metadata_lock.container_dev_id = dev_id;
```
tatetian: Why set `container_dev_id` lazily? Why not assign the right value for `container_dev_id` when creating `RootInode`?

kernel/src/fs/exfat/fs.rs
```diff
@@ -38,6 +39,7 @@ use crate::{
 pub struct ExfatFs {
     block_device: Arc<dyn BlockDevice>,
     super_block: ExfatSuperBlock,
+    dev_id: DeviceId,
```
tatetian: This field is unnecessary: we can already retrieve this information cheaply from the `block_device` field.

kernel/src/fs/ext2/fs.rs
```diff
@@ -22,6 +24,7 @@ const ROOT_INO: u32 = 2;
 pub struct Ext2 {
     block_device: Arc<dyn BlockDevice>,
     super_block: RwMutex<Dirty<SuperBlock>>,
+    dev_id: DeviceId,
```
tatetian: Again, adding this field is unnecessary.

kernel/src/fs/ext2/fs.rs
```diff
@@ -110,6 +114,11 @@ impl Ext2 {
         self.block_device.as_ref()
     }
 
+    /// Returns the device ID containing this filesystem.
+    pub fn dev_id(&self) -> DeviceId {
```
tatetian: For consistency, rename this method to `container_device_id`.

kernel/src/fs/utils/fs.rs
```diff
@@ -20,10 +21,11 @@ pub struct SuperBlock {
     pub namelen: usize,
     pub frsize: usize,
     pub flags: u64,
+    pub dev_id: DeviceId,
```
tatetian: For consistency, rename this to `container_dev_id`.

kernel/src/fs/utils/inode.rs
```diff
@@ -234,7 +234,7 @@ impl Metadata {
             nr_hard_links: 2,
             uid: Uid::new_root(),
             gid: Gid::new_root(),
-            container_dev_id: DeviceId::none(), // FIXME: placeholder
+            container_dev_id: DeviceId::null(),
```
tatetian: Like all other fields in `Metadata`, the right value should be set upon creation. The `container_dev_id` of an inode almost always inherits from its parent. I think we should be able to fill the correct value into `Metadata::container_dev_id` upfront. This simplifies the logic to use this field.

kernel/src/fs/overlayfs/fs.rs
```diff
@@ -123,12 +126,18 @@ impl OverlayFs {
         Self::validate_work_and_upper(&work, &upper)?;
         Self::validate_work_empty(&work)?;
 
+        let dev_id = pseudofs::allocator::DEVICE_ID_ALLOCATOR
```
tatetian: Are you sure that OverlayFS should be treated as a pseudo FS? Usually, the layers of an OverlayFS are indeed backed by physical storage.

kernel/src/fs/utils/systree_inode.rs
```diff
@@ -336,6 +336,9 @@ impl<KInode: SysTreeInodeTy + Send + Sync + 'static> Inode for KInode {
     default fn metadata(&self) -> Metadata {
         let mut metadata = *self.metadata();
         metadata.mode = self.mode().unwrap();
+        if metadata.container_dev_id.is_null() {
+            metadata.container_dev_id = self.fs().sb().dev_id;
```
tatetian: What is modified here is the _default_ implementation. A concrete type `T: SysTreeInodeTy` is allowed to override the default implementation. So it is impossible that this single location of modification is sufficient to guarantee that _all_ inodes of `SysFs` or `ConfigFS` returns the correct value of `container_dev_id`.

I think you should modify the implementators of `SysTreeInodeTy` (the concrete structs) so that their `metadata` methods return the correct value of `container_dev_id`.

tatetian (commented):
Thanks for the revision. This looks much better.

(Remember to resolve issues when you think your revision has address them.)

kernel/src/fs/devpts/mod.rs
LordaeronESZ: In the revised version, I implemented the logic for setting the `container_dev_id` when creating an inode in a separate commit, which you can check out.

In my implementation, it causes many "new_xxx" methods to have an additional `dev_id` parameter added, which looks ugly compared to lazy methods. I haven't thought of a more elegant solution yet.

kernel/src/fs/overlayfs/fs.rs
LordaeronESZ: I tested it on my local Linux environment, and its DeviceID meets the standard of `major = 0 && minor >= 1`.

Furthermore, this can also be seen in the Linux kernel code:

``` c
/* 
 * Linux 6.18 kernel source code
 * fs/overlayfs/super.c:1576
 */
struct file_system_type ovl_fs_type = {
	.owner			= THIS_MODULE,
	.name			= "overlay",
	.init_fs_context	= ovl_init_fs_context,
	.parameters		= ovl_parameter_spec,
	.fs_flags		= FS_USERNS_MOUNT,
	.kill_sb		= kill_anon_super,  /* Typical characteristics of a pseudo file system */
};
MODULE_ALIAS_FS("overlay");
```

kernel/src/fs/cgroupfs/fs.rs
```diff
@@ -41,8 +42,13 @@ impl CgroupFs {
     }
 
     fn new(root_node: Arc<CgroupSystem>) -> Arc<Self> {
-        let sb = SuperBlock::new(MAGIC_NUMBER, BLOCK_SIZE, NAME_MAX);
-        let root_inode = CgroupInode::new_root(root_node);
+        let dev_id = pseudofs::allocator::DEVICE_ID_ALLOCATOR
```
lrh2000: This can be simplified to `pseudofs::DEVICE_ID_ALLOCATOR` by marking the module as private. Then, re-export the type instead:
```rust
pub use allocator::DEVICE_ID_ALLOCATOR;
```

So we don't need to repeat the "allocator" keyword twice.

kernel/src/fs/cgroupfs/fs.rs
```diff
@@ -41,8 +42,13 @@ impl CgroupFs {
...
+            .get()
+            .expect("pseudofs's DEVICE_ID_ALLOCATOR is not initialized")
```
lrh2000: I don't think it's beneficial to repeat this error message across files. Failures caused by `Once` unwraps mean that the data is not initialized. It's common knowledge, so I think calling `unwrap` without repeating the error message is fine.

If we don't want to have this `unwrap`, we can also just follow the pseudofs's initialization pattern:
https://github.com/asterinas/asterinas/blob/e94a000b1193238b8e8e0ee237451c0fcf13150b/kernel/src/fs/configfs/fs.rs#L41-L46

```rust
impl DeviceIdAllocator {
    pub fn singleton() -> &'static DeviceIdAllocator {
        todo!()
    }
}
```

kernel/src/fs/cgroupfs/fs.rs
```diff
@@ -41,8 +42,13 @@ impl CgroupFs {
...
+            .allocate()
+            .expect("failed to allocate device ID for CgroupFs: minor numbers exhausted");
```
lrh2000:
```suggestion
            .expect("no device ID is available for cgroupfs");
```
 - "minor numbers exhausted" sounds like an implementation detail of the `allocate()` function. This detail should not be exposed/repeated here.
 - Use cgroupfs, configfs, devpts, procfs, e.t.c. when writing plain English.

kernel/src/fs/pseudofs.rs
```diff
@@ -1,11 +1,34 @@
...
+//! using a generic pseudo FS implementation called `NaivePseudoFs`.
+//!
+//! On the other hand, some pseudo FSes are quite complex,
+//! including [`RamFs`], `ProcFs`, and `SysFs`. Just to name a few.
+//! They are implemented in their own modules.
+//!
+//! Pseudo FSes are not backed by physical storage devices,
+//! but they still need to be assigned container device IDs
+//! (see [`Metadata::container_dev_id`]).
+//! [Linux's behavior](https://github.com/torvalds/linux/blob/dc855b77719fe452d670cae2cf64da1eb51f16cc/fs/super.c#L1257)
```
lrh2000:
```suggestion
//! [Linux's behavior](https://elixir.bootlin.com/linux/v6.19.3/source/fs/super.c#L1254)
```

Use LXR link because it's shorter and more friendly to use.

kernel/src/fs/utils/inode.rs
```diff
@@ -124,101 +125,193 @@ impl From<AccessMode> for Permission {
...
-    pub atime: Duration,
-    pub mtime: Duration,
-    pub ctime: Duration,
+
+    /// The optimal block size for filesystem I/O operations.
+    ///
+    /// Corresponds to `st_blksize`.
+    pub optimal_block_size: usize,
+
+    /// Number of 512-byte sectors allocated for the inode on disk.
```
lrh2000:
```suggestion
    /// The number of 512-byte sectors allocated for the inode on disk.
```

kernel/src/fs/utils/inode.rs
```diff
@@ -124,101 +125,193 @@ impl From<AccessMode> for Permission {
...
+    ///
+    /// Corresponds to `st_blocks`.
```
lrh2000: Be consistent with other fields. This should be moved below?
```rust
    /// This represents physical usage.
    /// For sparse files (those having holes), `size` is greater than this field.
    /// For files with preallocated blocks (`FALLOC_FL_KEEP_SIZE`),
    /// `size` is smaller than this field.
    ///
    /// Corresponds to `st_blocks`.
```

kernel/src/fs/pseudofs.rs
```diff
@@ -25,6 +48,60 @@ use crate::{
...
+        minor_allocator: Mutex<IdAlloc>,
+    }
+
+    /// An allocator for pseudo filesystems (no backing block device) device ID.
+    ///
+    /// This follows the Linux convention where pseudo filesystems use major=0
+    /// and dynamically allocate minor numbers (starting from 1) to distinguish different
+    /// pseudo filesystem instances.
+    ///
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/super.c#L1242-L1271>
```
lrh2000: It seems that the doc comments are put at the wrong place. Move it before `pub struct DeviceIdAllocator`?

kernel/src/fs/pseudofs.rs
```diff
@@ -25,6 +48,60 @@ use crate::{
...
+    impl DeviceIdAllocator {
+        pub fn new() -> Self {
```
lrh2000: This method does not need to be `pub`. Code outside this module should not be allowed to construct a new `DeviceIdAllocator`.

kernel/src/fs/pseudofs.rs
```diff
@@ -25,6 +48,60 @@ use crate::{
...
+            let mut minor_allocator = IdAlloc::with_capacity(MinorId::MAX.get() as usize + 1);
+            // Mark 0 as allocated to ensure minor numbers start from 1.
+            let _ = minor_allocator.alloc_specific(0);
```
lrh2000:
```suggestion
            let _ = minor_allocator.alloc_specific(0).unwrap();
```

kernel/src/fs/pseudofs.rs
```diff
@@ -25,6 +48,60 @@ use crate::{
...
+
+            Self {
+                minor_allocator: Mutex::new(minor_allocator),
+            }
+        }
+
+        /// Allocate a device ID for pseudo filesystems.
```
lrh2000:
```suggestion
        /// Allocates a device ID for pseudo filesystems.
```

Same for the `release` method.

test/initramfs/src/apps/fs/pseudofs/pseudo_dev_id.c
```diff
@@ -0,0 +1,118 @@
...
+	TEST_SUCC(get_dev_id("/sys", &dev_ids[1]));
+	TEST_SUCC(get_dev_id("/dev/pts", &dev_ids[2]));
+	TEST_SUCC(get_dev_id("/sys/fs/cgroup", &dev_ids[3]));
+	TEST_SUCC(get_dev_id("/sys/kernel/config", &dev_ids[4]));
+	TEST_SUCC(get_dev_id("/dev", &dev_ids[5]));
+
+	// Test that each pseudo filesystem has a unique device ID
+	for (int i = 0; i < 6; i++) {
+		for (int j = i + 1; j < 6; j++) {
+			TEST_RES(0, !dev_id_equal(&dev_ids[i], &dev_ids[j]));
```
lrh2000:
```suggestion
			TEST_RES(!dev_id_equal(&dev_ids[i], &dev_ids[j]), _ret);
```

This should work better because it will print the thing being tested?

Same for the following `pseudo_fs_files_share_same_container_dev_id` test.

test/initramfs/src/apps/fs/run_test.sh
```diff
@@ -112,3 +112,4 @@ echo "All mount bind file test passed."
 ./pseudofs/pseudo_dentry
 ./pseudofs/pseudo_inode
 ./pseudofs/pseudo_mount
+./pseudofs/pseudo_dev_id
```
lrh2000: The list should be sorted. Move it below `./pseudofs/pseudo_dentry`.

kernel/src/fs/utils/inode.rs
```diff
@@ -124,101 +125,193 @@ impl From<AccessMode> for Permission {
...
+
+    /// The ID of the device containing the inode.
+    ///
+    /// For persisted files, this device could be a on-disk partition or a logical volume (with RAID).
+    /// For pseudo files (e.g., those on sockfs), this device is also "pseudo".
+    ///
+    /// Corresponds to `st_dev`.
+    pub container_dev_id: DeviceId,
+
+    /// The device ID of the inode itself, if this inode represents a special device file (character or block).
```
lrh2000: This line is too long. Find a way to break this line, e.g.,
```rust
    /// The device ID of the inode itself, if this inode represents a special device file
    /// (character or block).
```

Same for line 209, which is also too long.

test/initramfs/src/apps/fs/pseudofs/pseudo_dev_id.c
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+END_TEST()
+
+FN_TEST(pseudo_fs_files_share_same_container_dev_id)
+{
+	struct dev_id dev_id_proc_self, dev_id_proc_cpuinfo;
+	struct dev_id dev_id_sys_kernel, dev_id_sys_fs;
+
+	// Files in the same pseudo fs should share the same container dev_id
```
lrh2000:
```suggestion
	// Files in the same pseudo fs should share the same container device ID
```

lrh2000:
In addition, to keep the PR comments readable, you can press the "Resolve conversation" button after addressing the comment. Thanks!

kernel/libs/device-id/src/lib.rs
```diff
@@ -38,6 +38,25 @@ impl DeviceId {
...
+
+    /// Returns whether the device ID refers to an anonymous or unnamed device.
+    ///
+    /// An anonymous device has the value of 0 as the major ID but a non-zero value for the minor ID.
+    pub fn is_anonymous(self) -> bool {
+        self.major().get() == 0 && self.minor().get() >= 1
+    }
+
+    /// Creates a null device ID (both major and minor are 0).
+    pub const fn null() -> Self {
```
lrh2000: This method should be moved just below `new` because it's a constructor. Constructors can be grouped together for better readability.

kernel/src/fs/pseudofs/mod.rs
```diff
@@ -27,10 +50,12 @@ use crate::{
     time::clocks::RealTimeCoarseClock,
 };
 
+mod allocator;
 mod anon_inodefs;
 mod pidfdfs;
 mod pipefs;
 mod sockfs;
+pub use allocator::DEVICE_ID_ALLOCATOR;
```
lrh2000: Move this to a correct place. We already have something like:
```
pub use pidfdfs::PidfdFs;
pub use sockfs::SockFs;
```

Move this item near it, then rustfmt will sort the imports automatically when you format the code.

kernel/src/fs/pseudofs/mod.rs
```diff
@@ -110,6 +149,7 @@ pub(super) fn init() {
     super::registry::register(&SockFsType).unwrap();
     // Note: `AnonInodeFs` does not need to be registered in the FS registry.
     // Reference: <https://elixir.bootlin.com/linux/v6.16.5/A/ident/anon_inode_fs_type>
+    allocator::DEVICE_ID_ALLOCATOR.call_once(allocator::DeviceIdAllocator::new);
```
lrh2000: Add a blank line before this line. This is not related to the registration of FS types.

test/initramfs/src/apps/fs/pseudofs/pseudo_dev_id.c
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
+#include <sys/types.h>
+
+#include "../../common/test.h"
+
+struct dev_id {
```
tatetian: Remove `struct dev_id` and simply use `dev_t`.

This `dev_id` struct adds extra complexity without giving us any benefit. All the information packed inside `struct dev_id` is already inside `dev_t`. And retrieving major and minor device IDs from a `dev_t` value only takes a macro call based on bit mask operations.

Using `dev_t` rather than `struct dev_id` simplifies the code. For example, we wouldn't need `dev_id_equal` any more.

test/initramfs/src/apps/fs/pseudofs/pseudo_dev_id.c
```diff
@@ -0,0 +1,118 @@
...
+{
+	return a->st_dev == b->st_dev;
+}
+
+FN_TEST(pseudo_fs_dev_id_major_is_zero)
+{
+	struct dev_id dev_id;
+
+	TEST_SUCC(get_dev_id("/proc", &dev_id));
+	TEST_RES(dev_id.major, _ret == 0);
```
tatetian: The following lines

```c
	TEST_SUCC(get_dev_id("/dev/pts", &dev_id));
	TEST_RES(dev_id.major, _ret == 0);
```

should be simplified to an one-liner:

```c
	TEST(get_dev_id("/dev/pts", &dev_id), 0, dev_id.major == 0));
```

To make it even more readable, you could define a macro called `OK`:

```
#define OK (0)
```

which allows you to write

```c
	TEST(get_dev_id("/dev/pts", &dev_id), OK, dev_id.major == 0));
```

kernel/src/fs/pseudofs/mod.rs
```diff
@@ -1,11 +1,34 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! The infrastructure of pseudo FSes along with some naive implementations of pseudo FSes.
+//!
+//! Pseudo FSes are those that have no physical storage.
+//! Some pseudo FSes are trivial,
+//! e.g., [`PipeFs`], [`SockFs`], [`PidfdFs`], and [`AnonInodeFs`].
+//! Such trivial pseudo FSes are directly implemented in this module
+//! using a generic pseudo FS implementation called `NaivePseudoFs`.
```
tatetian: You expressed the confusion about `PseudoFs`, which is called a pseudo FS, but there are more pseudo FSes other than the ones defined through `PseudoFs`. Let's why I proposed writing this module-level doc comments and renaming `PseudoFs` to `NaivePseudoFs`. But you adopted the new name in this doc comment without actually conducting the renaming. But do this renaming in a separate commit.

kernel/src/fs/pseudofs/mod.rs
```diff
@@ -28,11 +51,13 @@ use crate::{
     time::clocks::RealTimeCoarseClock,
 };
 
+mod allocator;
 mod anon_inodefs;
 mod nsfs;
 mod pidfdfs;
 mod pipefs;
 mod sockfs;
+pub use allocator::{DeviceIdAllocator, release_container_dev_id};
```
tatetian: Do not mix `mod` statements with `use` statements.

kernel/src/syscall/stat.rs
```diff
@@ -175,19 +175,19 @@ struct Stat {
...
-            st_rdev: info.rdev,
+            st_rdev: info.self_dev_id.map_or(0, |id| id.as_encoded_u64()),
             st_size: info.size as i64,
-            st_blksize: info.blk_size as _,
-            st_blocks: (info.blocks * (info.blk_size / 512)) as i64,
-            st_atime: info.atime.into(),
-            st_mtime: info.mtime.into(),
-            st_ctime: info.ctime.into(),
+            st_blksize: info.optimal_block_size as _,
+            st_blocks: info.nr_sectors_allocated as _,
```
tatetian, line 189: The old code converted from filesystem-block-sized counts to 512-byte sector counts via `blocks * (blk_size / 512)`. The new code passes `nr_sectors_allocated` through directly, assuming all producers now store values in 512-byte sector units.

However, multiple filesystem implementations still populate `nr_sectors_allocated` in filesystem-block units:

- **ext2** (`kernel/src/fs/ext2/inode.rs:113`): `nr_sectors_allocated: inner.blocks_count()` where `blocks_count()` computes `self.size.div_ceil(BLOCK_SIZE)` with `BLOCK_SIZE = 4096`. This is 4096-byte blocks, not 512-byte sectors.
- **ramfs** (`kernel/src/fs/ramfs/fs.rs:1202`): `nr_sectors_allocated: inode_metadata.blocks` where `blocks` is maintained as `new_size.align_up(BLOCK_SIZE) / BLOCK_SIZE` with `BLOCK_SIZE = 4096`.
- **exfat** (`kernel/src/fs/exfat/inode.rs:1364`): `nr_sectors_allocated: inner.size.div_ceil(blk_size)` where `blk_size` is the exfat sector size (512–4096 bytes).

This means `st_blocks` will be under-reported by a factor of `blk_size / 512` (e.g., 8x for 4096-byte block filesystems). The producers should be updated to store 512-byte sector counts internally or do the conversion in the their `metadata` methods.

kernel/libs/device-id/Cargo.toml
```diff
@@ -7,6 +7,8 @@ edition.workspace = true
 
 [dependencies]
 aster-util.workspace = true
+id-alloc.workspace = true
```
tatetian: These dependencies are added to the `device-id` library crate but are not used anywhere in `kernel/libs/device-id/src/lib.rs` (the only source file in the crate). The `DeviceIdAllocator` that uses `id-alloc` and `spin` lives in `kernel/src/fs/pseudofs/allocator.rs`, which is part of the kernel crate. These dependencies should be removed from the `device-id` crate.

kernel/src/fs/pseudofs/allocator.rs
```diff
@@ -0,0 +1,62 @@
...
+        Some(DeviceId::new(major, MinorId::new(minor as u32)))
+    }
+
+    /// Frees a dynamically allocated pseudo filesystem device ID.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the device ID's major is not 0.
+    pub fn release(&self, id: DeviceId) {
+        debug_assert!(id.major().get() == 0);
```
tatetian: The doc comment says "# Panics: Panics if the device ID's major is not 0" but `debug_assert!` is compiled out in release builds. If a non-pseudo device ID is accidentally released, this would silently corrupt the allocator state. Consider using `assert!` here, since this is a low-frequency operation (filesystem unmount) where the performance cost is negligible.

kernel/src/fs/ramfs/fs.rs
```diff
@@ -64,6 +68,7 @@ impl RamFs {
                 typ: InodeType::Dir,
                 this: weak_root.clone(),
                 fs: weak_fs.clone(),
+                container_dev_id: DeviceId::null(),
```
tatetian: The root inode (and all child inodes) of `RamFs` use `DeviceId::null()` as a sentinel, with the `metadata()` method at line 1193 falling back to `self.fs().sb().container_dev_id` when `is_null()` is true. This works because the allocator reserves minor=0, but it adds an extra branch on every `metadata()` call. Consider passing the actual `dev_id` directly instead — at this point in `RamFs::new()`, it's already available.

The `dev_id` is computed at line 54 and the root inode is created at line 60 within the same `Arc::new_cyclic` closure, so it could be captured directly.

test/initramfs/src/apps/fs/run_test.sh
```diff
@@ -113,5 +113,6 @@ echo "All mount bind file test passed."
 
 ./pseudofs/memfd_access_err
 ./pseudofs/pseudo_dentry
+./pseudofs/pseudo_dev_id
```
tatetian, line 116: Missing trailing newline at end of file.

tatetian:
## Missing `Drop` impls for device ID release

The manual `release_container_dev_id` pattern is easy to forget. In fact, three filesystems in this PR allocate anonymous device IDs but never release them:

| Filesystem | Allocates? | Has `Drop` release? |
|---|---|---|
| `RamFs` | Yes | Yes |
| `ProcFs` | Yes | Yes |
| `PseudoFs` | Yes | Yes |
| `DevPts` | Yes | Yes |
| `OverlayFs` | Yes | Yes |
| **`CgroupFs`** | **Yes** | **No** |
| **`ConfigFs`** | **Yes** | **No** |
| **`SysFs`** | **Yes** | **No** |

### Suggested fix: use RAII

The root cause is that `DeviceId` is a plain `Copy` type with no destructor. Consider introducing a guard type that owns the allocated ID and releases it on drop:

```rust
/// An RAII guard for an anonymous device ID allocated from `DeviceIdAllocator`.
///
/// The device ID is automatically released back to the allocator when this
/// guard is dropped. This prevents leaks from missing manual `release` calls.
pub struct AnonDeviceId {
    id: DeviceId,
}

impl AnonDeviceId {
    /// Returns the underlying `DeviceId`.
    pub fn id(&self) -> DeviceId {
        self.id
    }
}

impl Drop for AnonDeviceId {
    fn drop(&mut self) {
        DeviceIdAllocator::singleton().release(self.id);
    }
}
```

Then change `DeviceIdAllocator::allocate()` to return `Option<AnonDeviceId>` instead of `Option<DeviceId>`. Each filesystem stores an `AnonDeviceId` field (e.g., alongside or inside its `SuperBlock`), and the release happens automatically — no manual `Drop` impl needed on every filesystem. The free function `release_container_dev_id` and all the hand-written `Drop` impls (`impl Drop for RamFs`, etc.) can be deleted entirely.

This makes it **impossible to forget the release** — the compiler enforces it through ownership. Since `AnonDeviceId` is not `Copy` or `Clone`, it can't be silently duplicated either, preventing double-free bugs.

kernel/src/fs/pseudofs/allocator.rs
```diff
@@ -17,6 +19,12 @@ pub struct DeviceIdAllocator {
 }
 
 impl DeviceIdAllocator {
+    pub fn singleton() -> &'static Arc<Self> {
+        static SINGLETON: Once<Arc<DeviceIdAllocator>> = Once::new();
+
+        SINGLETON.call_once(|| Arc::new(Self::new()))
+    }
+
     pub(super) fn new() -> Self {
```
tatetian: Exposing `new` to the parent module is useless; make it private. See the [`narrow-visibility`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#narrow-visibility) guideline.

test/initramfs/src/apps/fs/pseudofs/pseudo_dev_id.c
lrh2000: TEST_RES already ensures TEST_SUCC, so the following is good enough?
```
TEST_RES(get_dev_id("/dev/pts", &dev_id), dev_id.major == 0);
```

tatetian:
@LordaeronESZ Mark this PR "Ready for Review" when you are done

LordaeronESZ:
> However, multiple filesystem implementations still populate nr_sectors_allocated in filesystem-block units:

In this revised PR, I fixed the filling issue of `nr_sectors_allocated` in these filesystems. However, there's a risk of overflow when converting from `nr_blocks` to `nr_sectors`. My current approach is to have the kernel panic immediately upon overflow, which is clearly unreasonable.

In my opinion, the correct approach would be to calculate the maximum supported single-file size when the filesystem is mounted and then store it to a metadata structure such as the superblock (e.g., `s_maxbytes` in Linux's VFS superblock). This would limit the single-file size before subsequent file extending writes (by returning an error code to user space), preventing metadata field overflow. Could this be implemented in a separate PR in the future?

tatetian:
> In this revised PR, I fixed the filling issue of `nr_sectors_allocated` in these filesystems. However, there's a risk of overflow when converting from `nr_blocks` to `nr_sectors`.

The overflow only exists in theory. `nr_sectors_allocated` is a 64-bit `usize` (Asterinas only targets 64-bit platforms), whose maximum value is far beyond any sectors that a single file could possibly occupy.

LordaeronESZ:
> The overflow only exists in theory. nr_sectors_allocated is a 64-bit usize (Asterinas only targets 64-bit platforms), whose maximum value is far beyond any sectors that a single file could possibly occupy.

If we only focus on the `Metadata::nr_sectors_allocated` field, that's indeed the case. However, this field ultimately originates from the `sector_count` field in the disk inode, and this field can be much smaller than 64 bits, such as in Ext2.

https://github.com/asterinas/asterinas/blob/d4c092f9ad08489b06b0cf236e711b1137fa92ef/kernel/src/fs/ext2/inode.rs#L2351-L2369

tatetian:
> However, this field ultimately originates from the `sector_count` field in the disk inode, and this field can be much smaller than 64 bits, such as in Ext2.

Sorry, I can't get your point: when the size of the on-disk value is smaller than 64 bits, the possibility of `Metadata:: nr_sectors_allocated` being overflowed is eliminated. So what's the problem?

LordaeronESZ:
> Sorry, I can't get your point: when the size of the on-disk value is smaller than 64 bits, the possibility of Metadata:: nr_sectors_allocated being overflowed is eliminated. So what's the problem?

I am sorry for not being clear enough. My concern is that Asterinas' current file system does not have a size limit based on the length of the metadata field when performing extended writes; such as Ext2, it only has a limit on the number of remaining blocks.

https://github.com/asterinas/asterinas/blob/49946619939b8962277ba558f4cfaa031a6b9b6f/kernel/src/fs/ext2/inode.rs#L1398-L1413

It may be necessary to introduce a "maximum file size" upperbound (similar to `super_block::s_maxbytes` in Linux) that is determined by multiple factors such as the size of metadata fields and the maximum number of data block pointers?

tatetian:
> It may be necessary to introduce a "maximum file size" upperbound (similar to `super_block::s_maxbytes` in Linux) that is determined by multiple factors such as the size of metadata fields and the maximum number of data block pointers?

This is a good concern, but it is out of the scope of this PR. You may raise an issue or a follow-up PR.

kernel/src/fs/utils/fs.rs
```diff
@@ -36,6 +42,7 @@ impl SuperBlock {
             namelen: name_max_len,
             frsize: block_size,
             flags: 0,
+            container_dev_id: dev_id.into(),
         }
```
tatetian: **Misplaced RAII responsibility**: `SuperBlock` stores an `AnonDeviceId`, which carries `Arc`-based automatic deallocation semantics. But `SuperBlock` is a data-only struct that gets cloned, returned by `FileSystem::sb()`, and passed around freely. This means every `SuperBlock::clone()` bumps the `Arc` refcount, and callers that only need the raw `DeviceId` value (e.g., `sb.container_dev_id.id()`) pay for unnecessary atomic reference counting.

Consider having the filesystem itself own the `AnonDeviceId` (for RAII cleanup) and storing a plain `DeviceId` in `SuperBlock`. The `Ext2` implementation already does something like this — it calls `self.block_device.id()` to get a plain `DeviceId` and stores it in `SuperBlock`. This pattern would be cleaner for pseudo filesystems too: the FS holds the `AnonDeviceId`, and `SuperBlock` just holds the extracted `DeviceId`.

kernel/src/fs/utils/inode.rs
```diff
@@ -124,101 +125,195 @@ impl From<AccessMode> for Permission {
...
-            dev: 0,
             ino,
             size: 2,
-            blk_size,
-            blocks: 1,
-            atime: now,
-            mtime: now,
-            ctime: now,
+            optimal_block_size: blk_size,
+            nr_sectors_allocated: 1,
```
tatetian: **Incorrect `nr_sectors_allocated` for pseudo FS directories**: `new_dir` hardcodes `nr_sectors_allocated: 1`, meaning 1 sector (512 bytes). This field maps to `st_blocks`, which represents physical disk usage. For in-memory pseudo filesystem directories, no sectors are allocated — Linux reports `st_blocks = 0` for procfs/sysfs/devpts directories.

In Linux, [`inode_init_always`](https://elixir.bootlin.com/linux/v6.13/source/fs/inode.c#L253) initializes `i_blocks = 0` for all newly allocated inodes. Pseudo filesystem code never overrides this for directories:
- **procfs**: [`proc_get_inode`](https://elixir.bootlin.com/linux/v6.13/source/fs/proc/inode.c#L657) sets mode, uid, gid, size, and nlink but never touches `i_blocks`.
- **sysfs/kernfs**: [`kernfs_refresh_inode`](https://elixir.bootlin.com/linux/v6.13/source/fs/kernfs/inode.c#L187) updates mode and nlink but not `i_blocks`.
- **tmpfs**: [`__shmem_get_inode`](https://elixir.bootlin.com/linux/v6.13/source/mm/shmem.c#L2579) explicitly sets `inode->i_blocks = 0` and the `S_IFDIR` case never modifies it.
- **libfs**: [`pseudo_fs_fill_super`](https://elixir.bootlin.com/linux/v6.13/source/fs/libfs.c#L671) and [`simple_fill_super`](https://elixir.bootlin.com/linux/v6.13/source/fs/libfs.c#L687) also leave `i_blocks` at the default 0.

In this codebase, `Metadata::new_dir` is only called by pseudo filesystems (devpts, procfs), so the value should be `0`, not `1`.

kernel/src/fs/pseudofs/allocator.rs
```diff
@@ -0,0 +1,99 @@
...
+use id_alloc::IdAlloc;
+use ostd::sync::Mutex;
+use spin::Once;
+
+/// A device ID wrapper that automatically recycles anonymous IDs on drop.
+///
+/// Clones share the same ownership so the wrapped ID is recycled only after the
+/// last reference is dropped.
+#[derive(Clone, Debug)]
+pub struct AnonDeviceId(Arc<AnonDeviceIdInner>);
```
tatetian: **Unnecessary `Arc` inside `AnonDeviceId`**: Not every type needs to support cloning or multiple owners. `AnonDeviceId` should be defined as:

```rust
pub struct AnonDeviceId(DeviceId);
```

If a use case really needs `AnonDeviceId` to be owned by multiple users, then it can wrap `AnonDeviceId` within an `Arc<_>`.

kernel/src/fs/pseudofs/allocator.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+impl Drop for AnonDeviceIdInner {
+    fn drop(&mut self) {
+        if self.id.is_anonymous() {
+            DeviceIdAllocator::singleton().release(self.id);
+        }
+    }
+}
```
tatetian: **Broken type invariant**: The `AnonDeviceId` type is given its name for a reason: it is supposed to only represent an anonymous device ID, which has the invariant that major ID is zero and minor ID is non-zero. But `From<DeviceId>` accepts *any* `DeviceId` — including non-anonymous ones (e.g., real block devices from ext2, which passes `self.container_device_id().into()` in `impl_for_vfs/fs.rs` line 42). This breaks the type-level invariant.

As a consequence, the `Drop` impl has to defensively check `is_anonymous()` before releasing, which is a code smell — if the type truly enforced its invariant at construction time, the drop could unconditionally release. (Note that `is_null()` and `is_anonymous()` partition the "major=0" space: a null device ID `(0, 0)` is not anonymous. The defensive check exists precisely because the constructor doesn't enforce this boundary.)

---

Furthermore, `From<DeviceId>` (or `TryFrom<DeviceId>`) should never have been implemented for `AnonDeviceId` because doing so breaks the type-level invariant of `DeviceIdAllocator`:

> The anonymous device ID of every `AnonDeviceId` origins from `DeviceIdAllocator`.

Consider the following code snippet:

```rust
let anony_device_id = DeviceId::new(MajerId::new(0), MinorId::new(1234));
let owned_anony_device_id = AnonDeviceId::From(anony_device_id);
drop(owned_anony_device_id);
// Error: The drop triggers the internal allocator to release an ID that is not allocated by it!
```

kernel/src/fs/pseudofs/allocator.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// An allocator for pseudo filesystems (no backing block device) device ID.
+///
+/// This follows the Linux convention where pseudo filesystems use major=0
+/// and dynamically allocate minor numbers (starting from 1) to distinguish different
+/// pseudo filesystem instances.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/super.c#L1242-L1271>
+pub struct DeviceIdAllocator {
```
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): With the addition of `AnonDeviceId`, exposing `DeviceIdAllocator` publicly is unnecessary. The allocator is an implementation detail. Add a constructor method on `AnonDeviceId` directly and hide `DeviceIdAllocator` from the user:

```rust
pub struct AnonDeviceId(DeviceId);

impl AnonDeviceId {
    pub fn acquire() -> Option<Self> {
        todo!("using the internal allocator")
    }
}
```

Callers would simply write `AnonDeviceId::acquire()` instead of `DeviceIdAllocator::singleton().allocate_anon_device_id()`.

This pattern is adopted by Asterinas. For example, the frame allocator and the heap allocator are not exposed to the end users; they can only see the `Frame<_>` and `Box<_>`.

kernel/src/fs/pseudofs/allocator.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        Self {
+            minor_allocator: Mutex::new(minor_allocator),
+        }
+    }
+
+    /// Allocates a wrapped anonymous device ID for pseudo filesystems.
+    ///
+    /// Returns `None` if minor number allocation fails (exhausted).
+    pub fn allocate_anon_device_id(&self) -> Option<AnonDeviceId> {
```
tatetian: **Redundant name**: Why is this method called `allocate_anon_device_id` rather than just `allocate`? The type is `DeviceIdAllocator` — do you plan to allow it to allocate any IDs other than anonymous device IDs? If not, the name is unnecessarily verbose. (This becomes moot if `DeviceIdAllocator` is hidden behind `AnonDeviceId::acquire()` as suggested above.)

kernel/src/fs/utils/inode.rs
```diff
@@ -124,101 +125,195 @@ impl From<AccessMode> for Permission {
...
+
+    /// The Group ID (GID) of the inode's owner.
+    ///
+    /// Corresponds to `st_gid`.
     pub gid: Gid,
-    pub rdev: u64,
+
+    /// The ID of the device containing the inode.
+    ///
+    /// For persisted files, this device could be a on-disk partition
```
tatetian: **Grammar**: "a on-disk" should be "an on-disk".

kernel/src/fs/ramfs/fs.rs
```diff
@@ -64,6 +69,7 @@ impl RamFs {
                 typ: InodeType::Dir,
```
tatetian: **Concrete example of misplaced RAII responsibility**: This is a tangible consequence of storing `AnonDeviceId` in `SuperBlock` (see the first comment). The `dev_id` is moved into `SuperBlock::new()`, then every subsequent `RamInode` creation (e.g., lines 465-568) recovers the raw `DeviceId` via `fs.sb.container_dev_id.id()`, traversing `fs → sb → AnonDeviceId → Arc → DeviceId`. If `RamFs` itself owned the `AnonDeviceId` and `SuperBlock` stored a plain `DeviceId`, this indirection would disappear and the code would be simpler.

kernel/src/fs/procfs/template/mod.rs
```diff
@@ -37,6 +37,47 @@ impl Common {
...
     }
 
+    fn new_dir(fs: Weak<dyn FileSystem>, mode: InodeMode, is_volatile: bool) -> Self {
+        let metadata = Self::with_procfs(&fs, |procfs| {
+            let ino = procfs.alloc_id();
+            Metadata::new_dir(ino, mode, BLOCK_SIZE, procfs.sb().container_dev_id.id())
+        });
+        Self::new(metadata, fs, is_volatile)
+    }
+
```
tatetian: **Repeated `procfs.sb().container_dev_id.id()` pattern**: The `new_dir`, `new_file`, and `new_symlink` constructors all call `procfs.sb().container_dev_id.id()` inside `with_procfs`. This is another instance of the "traverse through SuperBlock to get raw DeviceId" pattern. Caching the raw `DeviceId` on `ProcFs` directly would make the intent clearer. (This becomes a non-issue if `SuperBlock` stores a plain `DeviceId` as suggested in the first comment.)

Also, each of `Common::new_dir`, `Common::new_file`, and `Common::new_symlink` has exactly one caller in the corresponding `template/dir.rs`, `template/file.rs`, and `template/sym.rs` respectively. Consider moving them to those files (perhaps as free functions) rather than grouping them on `Common` in `template/mod.rs`.

kernel/src/syscall/statx.rs
```diff
@@ -146,21 +148,21 @@ impl Statx {
...
+            stx_blocks: info.nr_sectors_allocated as u64,
             stx_attributes_mask,
-            stx_atime: StatxTimestamp::from(info.atime),
-            stx_btime: StatxTimestamp::from(info.atime),
-            stx_ctime: StatxTimestamp::from(info.ctime),
-            stx_mtime: StatxTimestamp::from(info.ctime),
+            stx_atime: StatxTimestamp::from(info.last_access_at),
+            stx_btime: StatxTimestamp::from(info.last_access_at),
+            stx_ctime: StatxTimestamp::from(info.last_meta_change_at),
+            stx_mtime: StatxTimestamp::from(info.last_meta_change_at),
```
tatetian: **Pre-existing bug preserved: `stx_btime` and `stx_mtime` use wrong sources**: `stx_btime` (birth/creation time) uses `last_access_at` (atime), and `stx_mtime` (modification time) uses `last_meta_change_at` (ctime). These mappings are incorrect:
- `stx_btime` should use a birth time field (not currently tracked), or be zeroed out with a TODO.
- `stx_mtime` should use `last_modify_at`, not `last_meta_change_at`.

This bug predated this PR (the old code had `stx_btime: atime` and `stx_mtime: ctime` too), but since this PR touches these exact lines for the rename, it's a good opportunity to fix it. At minimum, `stx_mtime` should be `info.last_modify_at`.

tatetian:
#2946 reorganizes the FS code, thus causing a lot of conflicts. You have to fix the conflicts first. Sorry for the trouble caused.

PR #2889, opened 2026-01-15
Fix a typo in i8042 `is_present_cmdline`

lrh2000:
https://github.com/asterinas/asterinas/blob/f1a144b87658e856d82b5945a58f9461d9c33981/kernel/comps/i8042/src/controller.rs#L176-L186

This method does the opposite of what it is intended to do. The `!` operation should be removed.


tatetian (approved):

PR #2891, opened 2026-01-15
Add TSM-MR (measurement register) sysfs support

taosue:
This PR introduces TSM-MR (Trusted Security Module - Measurement Register). It provides sysfs interface to read measurement registers from the TD report and to extend measurement registers. The interface is documented at 
[here](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-virtual-misc-tdx_guest).

The implementation largely follows Linux behavior. In particular, TSM and TSM-MR are treated as mostly independent subsystems. E.g.,
1. The TD report is cached so that RTMR values can be retrieved without generating a new report every time.
2. TSM-MR generates a report during initialization and before user-provided `inblob` is written, i.e. user can read RTMR without requesting a quote and the report is generated with an all-zero `inblob`.
3. Although the report is updated by `extend_rtmr`, the cached quote continues to reflect the previous report. I.e. the quote is regenerated only when new `inblob` is provided.

### DIFF
--- a/kernel/src/device/misc/tdxguest.rs
+++ b/kernel/src/device/misc/tdxguest.rs
@@ -1,5 +1,43 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! Intel TDX guest device and measurement register API.
+//!
+//! This module provides two things:
+//!
+//! 1. **`/dev/tdx_guest` character device** — exposes the `TDX_CMD_GET_REPORT0`
+//!    ioctl, which lets userspace request a `TDREPORT_STRUCT` from the TDX module
+//!    (see [`TdxGuest`]).
+//!
+//! 2. **Measurement register API** — a set of public functions used by other
+//!    kernel subsystems (e.g. the TSM-MR sysfs layer) to interact with TDX
+//!    measurement registers.
+//!
+//! # Cached-report model
+//!
+//! A **TDX report** (`TDREPORT_STRUCT`) is a hardware-signed snapshot of the
+//! TD's current measurement state, produced by the `TDG.MR.REPORT` TDCALL.
+//! Throughout this module the word **"refresh"** always means issuing that
+//! TDCALL to regenerate the snapshot; it never refers to individual register
+//! reads.
+//!
+//! Because `TDG.MR.REPORT` is relatively expensive, the module keeps a single
+//! **global cached report** in `TDX_REPORT` and satisfies measurement-register
+//! reads from that cache.  The cache becomes **stale** after any
+//! [`extend_tdx_mr`] call, because extending an RTMR changes hardware state
+//! that is not reflected in the snapshot until the next refresh.
+//!
+//! # Measurement register API quick reference
+//!
+//! | Function | TDCALL? | Use when… |
+//! |---|---|---|
+//! | [`get_tdx_mr`] | No | Reading a register whose value is not expected to have changed since the last refresh (cheap). |
+//! | [`get_tdx_mr_refresh`] | Yes | Reading a register whose current hardware value is needed — e.g. immediately after [`extend_tdx_mr`]. The refresh and the register read are performed atomically under the write lock. |
+//! | [`refresh_tdx_report`] | Yes | Regenerating the report with a custom `report_data` blob (used by the `TDX_CMD_GET_REPORT0` ioctl and the quote path) without reading a specific register. |
+//! | [`extend_tdx_mr`] | Yes | Extending an RTMR. After extending, the cached report is stale; use [`get_tdx_mr_refresh`] (or [`refresh_tdx_report`] followed by [`get_tdx_mr`]) to observe the updated value. |
+//!
+//! For the TDX architecture specification see Intel's
+//! [TDX Module Specification](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-trust-domain-extensions.html).
+
 use alloc::sync::Arc;
 use core::{
     mem::{offset_of, size_of},
@@ -11,16 +49,17 @@ use device_id::{DeviceId, MinorId};
 use ostd::{
     const_assert,
     mm::{FrameAllocOptions, HasPaddr, HasSize, PAGE_SIZE, USegment, VmIo, dma::DmaCoherent},
-    sync::WaitQueue,
+    sync::{RwMutexWriteGuard, WaitQueue},
 };
+use spin::Once;
 use tdx_guest::{
     SHARED_MASK,
-    tdcall::{TdCallError, get_report},
-    tdvmcall::{TdVmcallError, get_quote},
+    tdcall::{self, TdCallError},
+    tdvmcall::{self, TdVmcallError},
 };
 
 use crate::{
-    device::{Device, DeviceType},
+    device::{Device, DeviceType, registry::char::register},
     events::IoEvents,
     fs::{
         file::{FileIo, StatusFlags},
@@ -160,9 +199,15 @@ impl FileIo for TdxGuestFile {
                         inblob_ptr.read()?
                     };
 
-                    let outblob = tdx_get_report(inblob.as_bytes())?;
+                    let report = TDX_REPORT
+                        .get()
+                        .ok_or_else(|| {
+                            Error::with_message(Errno::ENODEV, "TDX report not initialized")
+                        })?
+                        .write();
+                    refresh_tdx_report_locked(&report, Some(inblob.as_bytes()))?;
                     let outblob_ptr = field_ptr!(&data_ptr, TdxReportRequest, tdx_report);
-                    outblob_ptr.copy_from(&outblob)?;
+                    outblob_ptr.copy_from(&SafePtr::new(&*report, 0))?;
 
                     Ok(0)
                 })
@@ -172,28 +217,58 @@ impl FileIo for TdxGuestFile {
     }
 }
 
-pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
+static TDX_REPORT: Once<RwMutex<USegment>> = Once::new();
+
+/// Runtime Measurement Register (RTMR) index.
+///
+/// RTMRs are the only measurement registers that can be extended at runtime
+/// via [`extend_tdx_mr`].
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum Rtmr {
+    Rtmr0,
+    Rtmr1,
+    Rtmr2,
+    Rtmr3,
+}
+
+/// TDX measurement register.
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum MeasurementReg {
+    MrConfigId,
+    MrOwner,
+    MrOwnerConfig,
+    MrTd,
+    Rtmr(Rtmr),
+}
+
+/// Gets the TDX quote given the specified data in `inblob`.
+pub(crate) fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
     const GET_QUOTE_IN_FLIGHT: u64 = 0xFFFF_FFFF_FFFF_FFFF;
     const GET_QUOTE_BUF_SIZE: usize = 8 * 1024;
 
-    let report = tdx_get_report(inblob)?;
-
-    let buf = alloc_dma_buf(GET_QUOTE_BUF_SIZE)?;
-    let report_ptr: SafePtr<TdxQuoteHdr, _, _> = SafePtr::new(&buf, 0);
+    let buf = DmaCoherent::alloc(GET_QUOTE_BUF_SIZE.div_ceil(PAGE_SIZE), false)?;
+    let header_ptr: SafePtr<TdxQuoteHdr, _, _> = SafePtr::new(&buf, 0);
     let payload_ptr: SafePtr<TdReport, _, _> = SafePtr::new(&buf, size_of::<TdxQuoteHdr>());
 
-    field_ptr!(&report_ptr, TdxQuoteHdr, version).write(&1u64)?;
-    field_ptr!(&report_ptr, TdxQuoteHdr, status).write(&0u64)?;
-    field_ptr!(&report_ptr, TdxQuoteHdr, in_len).write(&(size_of::<TdReport>() as u32))?;
-    field_ptr!(&report_ptr, TdxQuoteHdr, out_len).write(&0u32)?;
-    payload_ptr.copy_from(&report)?;
+    field_ptr!(&header_ptr, TdxQuoteHdr, version).write(&1u64)?;
+    field_ptr!(&header_ptr, TdxQuoteHdr, status).write(&0u64)?;
+    field_ptr!(&header_ptr, TdxQuoteHdr, in_len).write(&(size_of::<TdReport>() as u32))?;
+    field_ptr!(&header_ptr, TdxQuoteHdr, out_len).write(&0u32)?;
+
+    let report = TDX_REPORT
+        .get()
+        .ok_or_else(|| Error::with_message(Errno::ENODEV, "TDX report not initialized"))?
+        .write();
+    refresh_tdx_report_locked(&report, Some(inblob))?;
+    payload_ptr.copy_from(&SafePtr::new(&*report, 0))?;
+    drop(report);
 
     // FIXME: The `get_quote` API from the `tdx_guest` crate should have been marked `unsafe`
     // because it has no way to determine if the input physical address is safe or not.
-    get_quote((buf.paddr() as u64) | SHARED_MASK, buf.size() as u64)?;
+    tdvmcall::get_quote((buf.paddr() as u64) | SHARED_MASK, buf.size() as u64)?;
 
     // Poll for the quote to be ready.
-    let status_ptr = field_ptr!(&report_ptr, TdxQuoteHdr, status);
+    let status_ptr = field_ptr!(&header_ptr, TdxQuoteHdr, status);
     let sleep_queue = WaitQueue::new();
     let sleep_duration = Duration::from_millis(100);
     loop {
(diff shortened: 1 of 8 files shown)


kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
...
+
+pub(super) fn init() {
+    let node = {
+        let tdx_guest_node = TdxGuestSysNodeRoot::new();
+        let measurement = Measurement::new(SysStr::from("measurements"));
+        tdx_guest_node.add_child(measurement).unwrap();
+
+        // FIXME: Temporary folder node until we have a proper sysfs devices
+        // implementation.
+        let misc_node = FolderNode::new("misc");
```
tatetian, line 42: @cchanging We need to figure out a way to add common device categories such as `misc/virtual/`. It is unacceptable for a leaf module such as tsm-mr to add the parent nodes.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
...
+inherit_sys_leaf_node!(Measurement, fields, {
+    fn perms(&self) -> SysPerms {
+        SysPerms::DEFAULT_RW_PERMS
+    }
+
+    fn read_attr_at(&self, name: &str, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        let attr = MEASUREMENT_ATTRS
+            .iter()
+            .find(|attr| attr.name == name)
+            .unwrap();
```
tatetian: You can't assume the user-provided `name` exists.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -211,12 +226,91 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
...
+    let mut blob = [0u8; SHA384_DIGEST_SIZE];
+    report
+        .read_bytes(TdReport::mr_offset(reg), blob.as_mut())
+        .unwrap();
+    Ok(blob)
+}
+
+/// Extends the measurement register in the TDX report.
+pub fn tdx_extend_mr(reg: MeasurementReg, data: &[u8]) -> Result<()> {
+    if data.len() != SHA384_DIGEST_SIZE {
```
tatetian: This is a public API. The constraint on the data length should be stated in the Rust doc.

kernel/src/security/tsm_mr.rs
cchanging: > add common device categories

Maybe we can do something like initializing `sys/kernel`. We can create a `devices` module under `sysfs`, which provides initialization for `sys/devices` and `sys/devices/[categories]`, and then exposes a register function externally:

```rust
pub fn register_device(device: Arc<dyn SysNode>, category_path: &[SysStr])  -> Result<()>;
```

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
...
+            if !*in_sync_write {
+                tdx_get_report(None).map_err(|_| Error::AttributeError)?;
+                *in_sync_write = true;
+            }
+            in_sync = in_sync_write.downgrade();
+        }
+
+        let mr = tdx_get_mr(attr.reg).map_err(|_| Error::AttributeError)?;
+
+        let mut reader = VmReader::from(&mr[offset..]);
```
tatetian: **Out-of-bounds panic on large `offset`**: The `offset` parameter originates from a userspace file read position. If `offset >= SHA384_DIGEST_SIZE` (48), the slice expression `&mr[offset..]` panics. This should return 0 bytes (EOF) instead. Either clamp `offset` or check it explicitly:

```rust
if offset >= mr.len() {
    return Ok(0);
}
let mut reader = VmReader::from(&mr[offset..]);
```

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
...
+        if attr.refresh_on_read() && !*in_sync {
+            let mut in_sync_write = in_sync.upgrade();
+            if !*in_sync_write {
+                tdx_get_report(None).map_err(|_| Error::AttributeError)?;
+                *in_sync_write = true;
+            }
+            in_sync = in_sync_write.downgrade();
+        }
+
+        let mr = tdx_get_mr(attr.reg).map_err(|_| Error::AttributeError)?;
```
tatetian: **Non-RTMR attributes read all-zeros**: For `mrconfigid`, `mrowner`, `mrownerconfig`, and `mrtd`, `refresh_on_read` is `false`, so `tdx_get_report` is never called before `tdx_get_mr`. The `TDX_REPORT` segment is zero-initialized in `tdxguest::init()` and never populated with an actual report until something triggers `tdx_get_report`. If a user reads any of these sysfs attributes before any RTMR read (which triggers the refresh), `tdx_get_mr` returns all zeros.

The PR description says "TSM-MR generates a report during initialization" but the code doesn't actually do this. Either generate an initial report in `init()` (either `tdxguest::init()` or `tsm_mr::init()`), or remove the `refresh_on_read` guard and always refresh when `!*in_sync` regardless of which register is being read.

**TOCTOU gap between refresh and read**: On line 177, `tdx_get_report(None)` acquires the `TDX_REPORT` write lock, regenerates the report, and returns a downgraded guard -- which is immediately dropped (it is not bound to a variable). On line 183, `tdx_get_mr` acquires a *fresh* read lock. Between these two operations, a concurrent `tdx_get_report(Some(inblob))` (via the `GetTdxReport` ioctl or `tdx_get_quote`) can overwrite the report data and regenerate the report with a different `inblob`. The `in_sync` flag does not protect against this because the ioctl path does not acquire `in_sync`.

Consider holding the `tdx_get_report` return guard until after reading the MR values, or restructuring so that the refresh + read happens atomically.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -208,12 +225,94 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
     // to private memory in TDX, `TDG.MEM.PAGE.ACCEPT` will zero out all content.
     // TDX Module Specification - `TDG.MEM.PAGE.ACCEPT` Leaf:
     // "Accept a pending private page and initialize it to all-0 using the TD ephemeral private key."
-    let out_len = field_ptr!(&report_ptr, TdxQuoteHdr, out_len).read()?;
+    let out_len = field_ptr!(&header_ptr, TdxQuoteHdr, out_len).read()?;
```
tatetian: **Lock held during unbounded polling**: The `report` variable (an `RwMutexUpgradeableGuard`) is alive from line 195 until end of function. After line 206 (`payload_ptr.copy_from(&report_ptr)`), the report data has been fully copied into the DMA buffer and the guard is no longer needed. Yet it continues to be held through the entire polling loop, which sleeps in 100ms increments waiting for the VMM to produce a quote -- an operation with no upper time bound.

While the upgradeable guard is held, `tdx_get_report(...)` (which calls `.write()`) will block. This means the `GetTdxReport` ioctl and any sysfs RTMR read that needs a refresh are blocked for the entire quote generation duration.

Fix: drop the guard explicitly after the data is copied to the DMA buffer:

```rust
payload_ptr.copy_from(&report_ptr)?;
drop(report_ptr);
drop(report);
```

kernel/src/device/misc/tdxguest.rs
```diff
@@ -208,12 +225,94 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
...
     Ok(outblob)
 }
 
+/// Gets the TDX report given the specified data in `inblob`.
+///
+/// The first `size_of::<TdReport>()` bytes of data in the returned `USegment` is the report.
+/// The rest in `USegment` should be ignored.
+/// If `inblob` is `None`, use the existing report data to get the report, i.e. refresh the
+/// report.
+pub fn tdx_get_report(inblob: Option<&[u8]>) -> Result<RwMutexUpgradeableGuard<'_, USegment>> {
```
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): Returning `RwMutexUpgradeableGuard<'_, USegment>` leaks the internal locking strategy to every caller. The lock hold duration is entirely at the caller's mercy, leading to the issue in `tdx_get_quote` above. The two current call patterns are:

1. "Refresh and immediately drop the guard" (`tsm_mr.rs` line 177) -- doesn't need a guard at all.
2. "Refresh and copy data out" (`tdx_get_quote`) -- only needs a brief read.

Consider restructuring so that the lock is managed internally. For example, expose two operations: `tdx_refresh_report(inblob)` that refreshes and releases the lock, and let `tdx_get_mr` (or a variant) read data under the lock atomically.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -172,28 +173,44 @@ impl FileIo for TdxGuestFile {
     }
 }
 
+static TDX_REPORT: Once<RwMutex<USegment>> = Once::new();
+
+#[derive(Debug, Clone, Copy)]
```
tatetian, line 226: [**`rust-type-invariants`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/types-and-traits.md#rust-type-invariants): `tdx_extend_mr` accepts any `MeasurementReg` but rejects the non-RTMR variants at runtime. This pushes a compile-time check into a runtime error. Consider splitting into two types:

```rust
pub(crate) enum Rtmr { Rtmr0, Rtmr1, Rtmr2, Rtmr3 }

pub(crate) enum MeasurementReg {
    MrConfigId,
    MrOwner,
    MrOwnerConfig,
    MrTd,
    Rtmr(Rtmr),
}

pub(crate) fn tdx_extend_mr(reg: Rtmr, data: &[u8; SHA384_DIGEST_SIZE]) -> Result<()>;
```

This makes `tdx_extend_mr` impossible to misuse with a read-only register.

Additionally, `MeasurementReg` has no doc comment. Per [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.md#rfc1574-summary), public types should have a doc comment.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -208,12 +225,94 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
...
 }
 
+/// Gets the TDX report given the specified data in `inblob`.
+///
+/// The first `size_of::<TdReport>()` bytes of data in the returned `USegment` is the report.
+/// The rest in `USegment` should be ignored.
+/// If `inblob` is `None`, use the existing report data to get the report, i.e. refresh the
+/// report.
+pub fn tdx_get_report(inblob: Option<&[u8]>) -> Result<RwMutexUpgradeableGuard<'_, USegment>> {
+    let report = TDX_REPORT.get().unwrap().write();
```
tatetian: [**`propagate-errors`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/error-handling.md#propagate-errors): `TDX_REPORT.get()` returns `None` if `init()` was never called. While initialization ordering makes this unlikely at runtime, the function signature already returns `Result`, so this could propagate an error instead of panicking. The same applies to `tdx_get_mr` (line 271). Consider:

```rust
let report = TDX_REPORT.get().ok_or_else(|| {
    Error::with_message(Errno::ENODEV, "TDX report not initialized")
})?.write();
```

kernel/src/device/misc/tdxguest.rs
```diff
@@ -208,12 +225,94 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
...
+        .read_bytes(TdReport::mr_offset(reg), blob.as_mut())
+        .unwrap();
+    Ok(blob)
+}
+
+/// Extends the measurement register in the TDX report.
+///
+/// `reg` must be one of the RTMR registers, and `data` must be exactly
+/// [`SHA384_DIGEST_SIZE`] (48) bytes long.
+pub fn tdx_extend_mr(reg: MeasurementReg, data: &[u8]) -> Result<()> {
```
tatetian: [**`rust-type-invariants`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/types-and-traits.md#rust-type-invariants): The `data` parameter is `&[u8]` but must be exactly 48 bytes. Since the sole caller (`tsm_mr.rs` `write_attr`) already validates the length, this internal API should accept `&[u8; SHA384_DIGEST_SIZE]` to enforce the constraint at compile time instead of runtime.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
...
+    BranchNodeFields, Error, NormalNodeFields, Result, SysAttrSetBuilder, SysObj, SysPerms, SysStr,
+    inherit_sys_branch_node, inherit_sys_leaf_node,
+};
+use inherit_methods_macro::inherit_methods;
+use ostd::{
+    mm::{FallibleVmRead, FallibleVmWrite, VmReader, VmWriter},
+    sync::RwMutex,
+};
+
+use crate::device::misc::tdxguest::{MeasurementReg, tdx_extend_mr, tdx_get_mr, tdx_get_report};
```
tatetian: [**`qualified-fn-imports`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.md#qualified-fn-imports): Free functions should be imported via the parent module and called qualified. Change to:

```rust
use crate::device::misc::tdxguest::{self, MeasurementReg};
```

Then use `tdxguest::tdx_get_report(...)`, `tdxguest::tdx_get_mr(...)`, etc. at call sites.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
...
+            .map_err(|_| Error::AttributeError)
+    }
+
+    fn write_attr(&self, name: &str, reader: &mut VmReader) -> Result<usize> {
+        match name {
+            "rtmr0:sha384" | "rtmr1:sha384" | "rtmr2:sha384" | "rtmr3:sha384" => {
+                let attr = MEASUREMENT_ATTRS
+                    .iter()
+                    .find(|attr| attr.name == name)
+                    .unwrap();
```
tatetian: [**`propagate-errors`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/error-handling.md#propagate-errors): This `.unwrap()` is in the runtime `write_attr` path. Although the preceding `match` arm guarantees `name` is one of the four `rtmr*` strings, the invariant is implicit. If the constant table or match arm ever diverges, this panics the kernel. Use `.ok_or(Error::NotFound)?` for consistency (as `read_attr_at` already does on line 170).

Additionally, the writable attribute names are hardcoded as string literals duplicating the names from `MEASUREMENT_ATTRS`. If an attribute name in the table changes, this match arm silently stops matching.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,256 @@
+// SPDX-License-Identifier: MPL-2.0
```
tatetian, line 1: [**`module-docs`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.md#module-docs): This new module has no `//!` module-level doc comment. The sibling `tsm.rs` has thorough module-level documentation. Add a brief description of what TSM-MR provides, the sysfs paths it creates, and a reference link (similar to the Linux kernel documentation link in the PR description).

The same applies to `kernel/src/device/misc/tdxguest.rs`, which now serves as the kernel's TDX report and measurement register API layer but lacks module-level documentation.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -208,12 +225,94 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
...
+
+    let index = match reg {
+        MeasurementReg::Rtmr0 => 0,
+        MeasurementReg::Rtmr1 => 1,
+        MeasurementReg::Rtmr2 => 2,
+        MeasurementReg::Rtmr3 => 3,
+        _ => return_errno_with_message!(Errno::EINVAL, "Only RTMR can be extended"),
+    };
+
+    let segment: USegment = FrameAllocOptions::new().alloc_segment(1)?.into();
```
tatetian: **Per-call page allocation**: `tdx_extend_mr` allocates and frees a 4 KiB physical page frame on every call just to pass 48 bytes via physical address. Since the `extend_rtmr` TDCALL only reads from the buffer and doesn't retain a reference, consider pre-allocating a dedicated `USegment` at `init()` time (protected by a mutex) and reusing it, similar to how `TDX_REPORT` is pre-allocated.

test/initramfs/src/apps/intel_tdx/tsm_mr/tsm_mr.sh
tatetian, line 1: **Incomplete test coverage**: The test only exercises write-then-read for RTMRs 0-3. Missing coverage:

1. **Read-only registers**: `mrconfigid`, `mrowner`, `mrownerconfig`, and `mrtd:sha384` are never read. A smoke test should verify these return non-zero data (or at least are readable without error).
2. **Before/after assertion**: The test never captures values before and after extend to verify the operation had an effect. With `set -e`, a silent no-op would still pass.
3. **Error paths**: No tests for wrong data size (e.g., 32 or 64 bytes) or writing to read-only registers. These exercise `write_attr`'s error handling.

tatetian (commented):
This PR adds a well-structured sysfs interface for TDX measurement registers, following the Linux TSM-MR model. The overall design -- caching the TD report and exposing per-register attributes through systree -- is sound.

However, there are a few correctness issues that should be addressed before merging: (1) reading non-RTMR attributes (e.g., `mrconfigid`) returns all-zeros because no initial report is generated; (2) `read_attr_at` panics if the userspace-supplied `offset` exceeds the 48-byte register size; (3) `tdx_get_quote` holds the `TDX_REPORT` lock guard through the entire quote-polling loop, which can block all report generation for an unbounded duration. There is also a TOCTOU gap in the sysfs read path where the report is refreshed and the lock released before the MR value is actually read.

On the API design side, the `MeasurementReg` enum conflates read-only and extendable registers, pushing what could be a compile-time check into a runtime error. The `tdx_get_report` function leaks a lock guard to callers, creating a fragile cross-module locking contract. Several items are `pub` where `pub(crate)` would suffice.

The test script covers the basic extend-and-read path for RTMRs but misses read-only registers, before/after value assertions, and error-path tests.

Last but not least, please add module-level doc comments. TDX-related features and concepts are lessor known. Use module-level doc comments as an opportunity to educate the readers, share your knowledge, and write down the spec.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -208,12 +261,125 @@ pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
...
     payload_ptr.cast().read_slice(outblob.as_mut())?;
     Ok(outblob)
 }
 
+/// Refreshes the TDX report and releases the lock immediately.
+///
+/// The first `size_of::<TdReport>()` bytes of data in the returned `USegment` is the report.
+/// The rest in `USegment` should be ignored. If `inblob` is `None`, use the existing report
+/// data to get the report.
+pub(crate) fn tdx_refresh_report(inblob: Option<&[u8]>) -> Result<()> {
```
tatetian: **Naming, documentation, and API clarity**: Three issues with this public API surface:

**1. Stale doc comment on `tdx_refresh_report`** (lines 270-274): The function now returns `Result<()>`, but the doc comment still references "the returned `USegment`" and its byte layout. The first line mentioning "releases the lock immediately" leaks an implementation detail.

**2. Naming**: In Rust, function names conventionally start with a verb. The `tdx_` prefix is redundant — the functions live in the `tdxguest` module and callers use qualified paths, so `tdxguest::tdx_get_mr(...)` stutters. Drop the prefix:

| Current | Suggested |
|---------|-----------|
| `tdx_refresh_report` | `refresh_tdx_report` |
| `tdx_get_mr` | `get_tdx_mr` |
| `tdx_get_mr_refresh` | `get_tdx_mr_refresh` |
| `tdx_extend_mr` | `extend_tdx_mr` |

The internal helper `tdx_refresh_report_locked` (line 355) should follow the same pattern: `refresh_tdx_report_locked`.

**3. API relationship is unclear**: These four functions form a coherent API around a cached TDX report, but their doc comments don't explain how they relate to each other or when to use which. A reader seeing both `get_mr` and `get_mr_refresh` has to read the implementations to understand the difference. Consider:

- Adding a section to the module-level doc (`//!`) that describes the cached-report model and lists the API with a brief usage guide, e.g.: "`get_mr` reads from the cached report (cheap, no TDCALL); `get_mr_refresh` regenerates the report first (needed after an `extend_mr`); `refresh_report` regenerates the report without reading a register (used by the ioctl path with a custom `inblob`)."
- Each function's doc comment should cross-reference the others where relevant. For example, `extend_mr`'s doc should note that subsequent reads require `get_mr_refresh` (or a `refresh_report` followed by `get_mr`) to see the updated value.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,274 @@
...
+inherit_sys_branch_node!(TdxGuestSysNodeRoot, fields, {
+    fn perms(&self) -> SysPerms {
+        SysPerms::DEFAULT_RW_PERMS
+    }
+});
+
+#[derive(Debug)]
+struct Measurement {
+    fields: NormalNodeFields<Self>,
+    in_sync: RwMutex<bool>,
```
tatetian, line 103: **Key design concept is undocumented**: The "refresh" model is central to this module's correctness but is never explained. A reader encountering `refresh_on_read`, `in_sync`, `refresh_report`, and `get_mr_refresh` has to reverse-engineer the design from the code.

The underlying idea is: the TDX report is a cached snapshot that must be regenerated via TDCALL to reflect current RTMR values. Static registers (`mrconfigid`, `mrowner`, `mrownerconfig`, `mrtd`) are fixed at TD creation time and never change, so they can always be read from the cached report. RTMRs, however, can be extended at runtime, which invalidates the cached snapshot — so a read must first "refresh" (regenerate) the report to get up-to-date values. The `in_sync` flag tracks whether an extend has happened since the last refresh.

Add doc comments on:
- `in_sync`: what it tracks, who sets it to `false` (extend), who sets it to `true` (refresh), and the invariant it maintains.
- `refresh_on_read`: why static registers don't need it and RTMRs do.
- In `tdxguest.rs`, the module-level doc or a comment near `refresh_report` should explain what "refresh" means in TDX terms (i.e., calling `TDG.MR.REPORT` to regenerate the cached `TDREPORT_STRUCT`).

kernel/src/device/misc/tdxguest.rs
```diff
@@ -172,28 +194,59 @@ impl FileIo for TdxGuestFile {
     }
 }
 
-pub fn tdx_get_quote(inblob: &[u8]) -> Result<Box<[u8]>> {
+static TDX_REPORT: Once<RwMutex<USegment>> = Once::new();
+static TDX_EXTEND_BUF: Once<Mutex<USegment>> = Once::new();
```
tatetian: **Unnecessary static for a cold-path buffer**: `TDX_EXTEND_BUF` pre-allocates a page-sized buffer that is only used inside `tdx_extend_mr` to pass 48 bytes to the `extend_rtmr` TDCALL. The TDCALL reads the buffer synchronously and does not retain a reference. RTMR extension is an infrequent operation, so the per-call cost of allocating and freeing a single page frame is negligible. Meanwhile the static adds a `Once`, a `Mutex`, init-time setup, and an "not initialized" error path.

Consider replacing with a local allocation inside `tdx_extend_mr`:

```rust
pub(crate) fn extend_mr(reg: Rtmr, data: &[u8; SHA384_DIGEST_SIZE]) -> Result<()> {
    let index = match reg { ... };
    let buf: USegment = FrameAllocOptions::new().alloc_segment(1)?.into();
    buf.write_bytes(0, data).unwrap();
    extend_rtmr(buf.paddr() as u64, index)?;
    Ok(())
}
```

This eliminates the global state and the `Mutex` serialization, and makes the function self-contained.

kernel/src/device/misc/tdxguest.rs
```diff
@@ -11,16 +26,17 @@ use device_id::{DeviceId, MinorId};
...
     const_assert,
     mm::{FrameAllocOptions, HasPaddr, HasSize, PAGE_SIZE, USegment, VmIo, dma::DmaCoherent},
-    sync::WaitQueue,
+    sync::{Mutex, RwMutexWriteGuard, WaitQueue},
 };
+use spin::Once;
 use tdx_guest::{
     SHARED_MASK,
-    tdcall::{TdCallError, get_report},
+    tdcall::{TdCallError, extend_rtmr, get_report},
```
tatetian: [**`qualified-fn-imports`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#qualified-fn-imports): The free functions `extend_rtmr`, `get_report`, and `get_quote` are imported directly. The PR already fixed this pattern in `tsm_mr.rs` (using `tdxguest::` qualified calls); the same style should be applied here for the external crate:

```rust
use tdx_guest::{
    SHARED_MASK,
    tdcall::{self, TdCallError},
    tdvmcall::{self, TdVmcallError},
};
```

Then call `tdcall::get_report(...)`, `tdcall::extend_rtmr(...)`, `tdvmcall::get_quote(...)` at call sites.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,274 @@
...
+    crate::fs::sysfs::systree_singleton()
+        .root()
+        .add_child(node.clone())
+        .unwrap();
+}
+
+/// A systree node representing the `/sys/devices/virtual/misc/tdx_guest`
+/// directory.
+#[derive(Debug)]
+pub struct TdxGuestSysNodeRoot {
```
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): `TdxGuestSysNodeRoot` is `pub` but is only constructed and used within `tsm_mr.rs` (line 36). Per guidelines, default to the narrowest visibility — `pub(super)` or even private would suffice here, unless the `inherit_sys_branch_node!` macro requires `pub`. If the macro does require it, add a brief comment explaining the constraint.

kernel/src/security/tsm_mr.rs
```diff
@@ -0,0 +1,274 @@
...
+            .iter()
+            .find(|attr| attr.name == name)
+            .ok_or(Error::NotFound)?;
+
+        let MeasurementReg::Rtmr(rtmr) = attr.reg else {
+            return Err(Error::PermissionDenied);
+        };
+
+        let data = {
+            let mut buf = [0u8; 48];
```
tatetian: **Magic number**: This `48` is the SHA-384 digest size, already defined as `SHA384_DIGEST_SIZE` in `tdxguest.rs` (line 283). Reuse the constant here to keep the two modules in sync:

```rust
use crate::device::misc::tdxguest::SHA384_DIGEST_SIZE;
// ...
let mut buf = [0u8; SHA384_DIGEST_SIZE];
```

This requires making `SHA384_DIGEST_SIZE` `pub(crate)`.

test/initramfs/src/regression/intel_tdx/tsm_mr/tsm_mr.sh
```diff
@@ -0,0 +1,112 @@
...
+    name="$1"
+    path="$SYSFS_DIR/$name"
+    data=$(hd "$path")
+    if [ -z "$data" ]; then
+        echo "FAIL: $name is empty" >&2
+        exit 1
+    fi
+    # hd prints "00" for every zero byte; if the output contains any non-space
+    # hex digit other than '0', at least one byte is non-zero.
+    if echo "$data" | grep -qE '[1-9a-f]'; then
```
tatetian: (Suggested by my review skill; I didn't confirm it independently)

**Non-zero assertion always passes**: `hd` (hexdump -C) output includes hex address offsets at the start of each line. For 48 bytes of all-zeros, the final line is `00000030` — the `3` matches `[1-9a-f]`, so the grep succeeds even when every data byte is zero. This makes the `mrtd:sha384` non-zero check ineffective.

Fix by using `od` with no address column, or by stripping the address prefix:

```sh
check_readable_nonzero() {
    name="$1"
    path="$SYSFS_DIR/$name"
    # od -A n suppresses the address column, printing only hex data bytes.
    nz=$(od -A n -t x1 < "$path" | tr -d ' \n')
    if [ -z "$nz" ]; then
        echo "FAIL: $name is empty" >&2
        exit 1
    fi
    cleaned=$(echo "$nz" | tr -d '0')
    if [ -z "$cleaned" ]; then
        echo "FAIL: $name read back as all-zeros" >&2
        exit 1
    fi
    echo "PASS: $name is readable and non-zero"
}
```

tatetian (commented):
Excellent follow-up iteration. All 13 comments from the previous review have been addressed: the out-of-bounds panic is fixed, the all-zeros bug is resolved by generating an initial report in `init()`, the TOCTOU gap is closed with `tdx_get_mr_refresh`, the lock-holding-during-polling bug is fixed with an explicit `drop`, the API no longer leaks lock guards, the type system now enforces RTMR-only extends, `TDX_EXTEND_BUF` pre-allocates the extend buffer, and the test script has comprehensive coverage of read-only registers, before/after assertions, and error paths.

The remaining comments are mostly about documentation and naming. The measurement-register API in `tdxguest.rs` needs clearer documentation: the relationship between the four public functions (`refresh_report`, `get_mr`, `get_mr_refresh`, `extend_mr`) is not obvious without reading the implementations, and the "refresh" concept that spans both modules is undocumented. The `tdx_` function name prefix stutters at call sites. `TDX_EXTEND_BUF` adds unnecessary global state for a cold path. There is also a real bug in the test script's non-zero assertion.

test/initramfs/src/regression/intel_tdx/tsm_mr/tsm_mr.sh
taosue: > (Suggested by my review skill; I didn't confirm it independently)
> 
Indeed, that's a very clever review skill.

tatetian (approved):

PR #2895, opened 2026-01-15
Add disk space cleanup step for test_riscv/x86 workflows

cqs21:
The `basic-test` and `integration-test` workflows are experiencing disk space issues, causing build or test failures. This PR removes large packages of the host system, including Google Cloud SDK, Android SDK, and .NET SDK, which frees up approximately 16GB of disk space to ensure reliable workflow execution.

### DIFF
--- a/.github/workflows/test_riscv.yml
+++ b/.github/workflows/test_riscv.yml
@@ -13,11 +13,21 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.0-20260114
       options: --device=/dev/kvm --privileged
+      volumes:
+        # Map the following directories on the host into the container so that
+        # the container can remove them to save some precious disk space (~16GB),
+        # which is quite limited on free-tier Github runners.
+        - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
+        - /usr/local/lib/android:/usr/local/lib/android
+        - /usr/share/dotnet:/usr/share/dotnet
     strategy:
       matrix:
         id: ['lint', 'compile', 'usermode_test', 'ktest']
       fail-fast: false
     steps:
+      - name: Free disk space
+        # Remove unused packages from the host to save disk space.
+        run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
       - uses: actions/checkout@v4
       - name: Run basic tests
         uses: ./.github/actions/test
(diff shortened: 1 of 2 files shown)


.github/workflows/test_riscv.yml
```diff
@@ -13,11 +13,19 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.0-20260114
       options: --device=/dev/kvm --privileged
+      volumes:
+        # Map the following volumes to the container in order to free disk space.
```
tatetian:
```yml
# Map the following directories on the host into the container so that
# the container can remove them to save some precious disk space (~16GB),
# which is quite limited on free-tier Github runners.
```

.github/workflows/test_riscv.yml
```diff
@@ -13,11 +13,19 @@ jobs:
...
+        - /usr/local/lib/android:/usr/local/lib/android
+        - /usr/share/dotnet:/usr/share/dotnet
     strategy:
       matrix:
         id: ['lint', 'compile', 'usermode_test', 'ktest']
       fail-fast: false
     steps:
+      - name: Free disk space
+        # Clean up the mapped volumes from host.
+        run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
```
tatetian, line 30:
```yml
# Remove unused packages from the host to save disk space.
```

.github/workflows/test_x86.yml
```diff
@@ -30,6 +38,11 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.0-20260114
       options: --device=/dev/kvm --privileged
+      volumes:
+        # Map the following volumes to the container in order to free disk space.
+        - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
+        - /usr/local/lib/android:/usr/local/lib/android
+        - /usr/share/dotnet:/usr/share/dotnet
```
tatetian, line 49: The code has been repeated multiple times. This adds burden for maintainability. I am wondering if this is truly unavoidable.

> Use jobs.<job_id>.container to create a container to run any steps in a job that don't already specify a container. If you have steps that use both script and container actions, the container actions will run as sibling containers on the same network with the same volume mounts.
>
> If you do not set a container, all steps will run directly on the host specified by runs-on unless a step refers to an action configured to run in a container.

[Github documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idcontainer) says: "If you have steps that use both script and container actions". Does it mean a job can run both commands on the host (so-called "script actions") and in the container (so-called "container actions")? If so, we don't need to map volumes.

Another angle to address this duplication problem is the [`defaults`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaults) setting. If I understand it correctly, you should be able to move the `volumes` key into `defaults`, so that you don't have to repeatedly specify `volumes` for each job. There are other shared keys (e.g., `runs-on`). They can all be moved into `defaults.`

tatetian (approved):

PR #2900, opened 2026-01-19
Add Linux kernel selftest (kselftest) support

### LINKED ISSUE #2857: Add Linux kernel selftests test suite
Linux itself maintains an in-tree userspace test suite called [selftests](https://docs.kernel.org/dev-tools/kselftest.html), which can be found under [`<linux>/tools/testing/selftests`](https://github.com/torvalds/linux/tree/master/tools/testing/selftests). This test suite should be introduced as a complement to LTP and gVisor test suites. A separate directory for Linux selftests can be added in adjacent to LTP and gVisor:

```
test/
    src/
        syscall/
            ltp/
            gvisor/
            kselftests/   # <-- Linux kselftests
```

Linux selftests include a large number of tests that target different subsystems. We need to identify test programs that are useful to Asterinas at the current stage and enable them as the initial step. In the future, we will add more of them.

cqs21:
This PR enables initial support for Linux Kernel Selftest (kselftest) on x86_64.

Fixes #2857

### DIFF
--- a/.github/workflows/test_x86.yml
+++ b/.github/workflows/test_x86.yml
@@ -21,9 +21,9 @@ jobs:
         - /usr/local/lib/android:/usr/local/lib/android
         - /usr/share/dotnet:/usr/share/dotnet
     strategy:
+      fail-fast: false
       matrix:
         id: ['lint', 'compile', 'usermode_test', 'ktest']
-      fail-fast: false
     steps:
       - name: Free disk space
         # Remove unused packages from the host to save disk space.
@@ -35,7 +35,44 @@ jobs:
           auto_test: 'general'
           runs_on: 'ubuntu-latest'
 
-  integration-test:
+  osdk-test:
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.image }}
+      options: -v /dev:/dev --privileged
+    strategy:
+      fail-fast: false
+      matrix:
+        image: ['asterinas/asterinas:0.17.1-20260319', 'asterinas/osdk:0.17.1-20260319']
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run OSDK tests
+        uses: ./.github/actions/test
+        with:
+          auto_test: 'osdk'
+          intel_tdx: false
+
+  boot-test:
+    runs-on: ubuntu-latest
+    container:
+      image: asterinas/asterinas:0.17.1-20260319
+      options: -v /dev:/dev --privileged
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { variant: 'legacy32', boot_protocol: 'linux-legacy32', smp: 4 }
+          - { variant: 'multiboot2-smp4', boot_protocol: 'multiboot', smp: 4 }
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run boot test (${{ matrix.variant }})
+        uses: ./.github/actions/test
+        with:
+          auto_test: 'boot'
+          smp: ${{ matrix.smp }}
+          boot_protocol: ${{ matrix.boot_protocol }}
+
+  regression-test:
     runs-on: ubuntu-latest
     container:
       image: asterinas/asterinas:0.17.1-20260319
@@ -48,73 +85,58 @@ jobs:
         - /usr/local/lib/android:/usr/local/lib/android
         - /usr/share/dotnet:/usr/share/dotnet
     strategy:
+      fail-fast: false
       matrix:
         include:
-          # SMP Boot Test (Linux Legacy 32-bit Boot Protocol)
-          - test_id: 'boot-legacy32'
-            boot_protocol: 'linux-legacy32'
-            smp: 4
-          # SMP Boot Test (Multiboot)
-          - test_id: 'boot-multiboot2-smp4'
-            boot_protocol: 'multiboot'
-            smp: 4
-
-          # Conformance Test (Linux EFI PE/COFF Boot Protocol) (Debug Build)
-          - test_id: 'conformance-debug'
-            release: false
-            boot_protocol: 'linux-efi-pe64'
-          # Conformance Test at Ext2 (MicroVM)
-          - test_id: 'conformance-ext2-microvm'
-            scheme: 'microvm'
-            extra_blocklists: 'blocklists.ext2'
-            conformance_test_workdir: '/ext2'
-          # Conformance Test at Ext2 (IOMMU) (Debug Build)
-          - test_id: 'conformance-ext2-iommu-debug'
-            scheme: 'iommu'
-            extra_blocklists: 'blocklists.ext2'
-            conformance_test_workdir: '/ext2'
-            release: false
-          # Conformance Test at Exfat (Multiboot2) (without KVM enabled)
-          - test_id: 'conformance-exfat-multiboot2-nokvm'
-            boot_protocol: 'multiboot2'
-            enable_kvm: false
-            extra_blocklists: 'blocklists.exfat'
-            conformance_test_workdir: '/exfat'
-          # SMP Conformance Test (Multiboot2)
-          - test_id: 'conformance-multiboot2-smp4'
-            boot_protocol: 'multiboot2'
-            smp: 4
-
-          # Regression Test (Linux EFI Handover) (Debug Build)
-          - test_id: 'regression-handover64-debug'
-            release: false
-            boot_protocol: 'linux-efi-handover64'
-          # SMP Regression Test (Multiboot2)
-          - test_id: 'regression-multiboot2-smp4'
-            boot_protocol: 'multiboot2'
-            smp: 4
-      fail-fast: false
+          - { variant: 'handover64-debug', release: false, boot_protocol: 'linux-efi-handover64' }
+          - { variant: 'multiboot2-smp4',  boot_protocol: 'multiboot2', smp: 4 }
     steps:
       - name: Free disk space
         # Remove unused packages from the host to save disk space.
         run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
       - uses: actions/checkout@v4
-      - name: Run integration tests ${{ (startsWith(matrix.test_id, 'conformance') && 'with LTP') || '' }}
+      - name: Run regression test (${{ matrix.variant }})
         uses: ./.github/actions/test
         with:
-          auto_test: ${{ (startsWith(matrix.test_id, 'boot') && 'boot') ||
-              (startsWith(matrix.test_id, 'conformance') && 'conformance') || 'regression' }}
+          auto_test: 'regression'
           release: ${{ !contains(matrix.release, 'false') }}
-          enable_kvm: ${{ !contains(matrix.enable_kvm, 'false') }}
           smp: ${{ matrix.smp }}
-          netdev: ${{ matrix.netdev || 'tap' }}
-          scheme: ${{ matrix.scheme }}
-          extra_blocklists: ${{ matrix.extra_blocklists }}
-          conformance_test_suite: 'ltp'
-          conformance_test_workdir: ${{ matrix.conformance_test_workdir }}
           boot_protocol: ${{ matrix.boot_protocol || 'linux-efi-handover64' }}
-      - name: Run integration tests ${{ (startsWith(matrix.test_id, 'conformance') && 'with gVisor') || '' }}
-        if: ${{ startsWith(matrix.test_id, 'conformance') }}
+
+  conformance-test:
+    runs-on: ubuntu-latest
+    container:
+      image: asterinas/asterinas:0.17.1-20260319
+      options: -v /dev:/dev --privileged
+      volumes:
+        # Map the following directories on the host into the container so that
+        # the container can remove them to save some precious disk space (~16GB),
+        # which is quite limited on free-tier Github runners.
+        - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
+        - /usr/local/lib/android:/usr/local/lib/android
+        - /usr/share/dotnet:/usr/share/dotnet
+    strategy:
+      fail-fast: false
+      matrix:
+        suite: ['ltp', 'gvisor', 'kselftest']
+        variant:
+          - 'debug'
+          - 'ext2-microvm'
+          - 'ext2-iommu-debug'
+          - 'exfat-multiboot2-nokvm'
+          - 'multiboot2-smp4'
+        include:
+          - { variant: 'debug', release: false, boot_protocol: 'linux-efi-pe64' }
+          - { variant: 'ext2-microvm', scheme: 'microvm', extra_blocklists: 'blocklists.ext2', conformance_test_workdir: '/ext2' }
+          - { variant: 'ext2-iommu-debug', scheme: 'iommu', extra_blocklists: 'blocklists.ext2', conformance_test_workdir: '/ext2', release: false }
+          - { variant: 'exfat-multiboot2-nokvm', boot_protocol: 'multiboot2', enable_kvm: false, extra_blocklists: 'blocklists.exfat', conformance_test_workdir: '/exfat' }
+          - { variant: 'multiboot2-smp4', boot_protocol: 'multiboot2', smp: 4 }
+    steps:
+      - name: Free disk space
+        # Remove unused packages from the host to save disk space.
+        run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
+      - uses: actions/checkout@v4
+      - name: Run conformance test (${{ matrix.variant }}, ${{ matrix.suite }})
         uses: ./.github/actions/test
         with:
           auto_test: 'conformance'
@@ -124,23 +146,6 @@ jobs:
           netdev: ${{ matrix.netdev || 'tap' }}
           scheme: ${{ matrix.scheme }}
           extra_blocklists: ${{ matrix.extra_blocklists }}
-          conformance_test_suite: 'gvisor'
+          conformance_test_suite: ${{ matrix.suite }}
           conformance_test_workdir: ${{ matrix.conformance_test_workdir }}
           boot_protocol: ${{ matrix.boot_protocol || 'linux-efi-handover64' }}
-
-  osdk-test:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        image: ['asterinas/asterinas:0.17.1-20260319', 'asterinas/osdk:0.17.1-20260319']
-      fail-fast: false
-    container:
-      image: ${{ matrix.image }}
-      options: -v /dev:/dev --privileged
-    steps:
-      - uses: actions/checkout@v4
-      - name: Run OSDK tests
-        uses: ./.github/actions/test
-        with:
-          auto_test: 'osdk'
-          intel_tdx: false
--- a/test/initramfs/Makefile
+++ b/test/initramfs/Makefile
@@ -109,6 +109,7 @@ x86_64_pkgs:
 		-A benchmark.sqlite-speedtest1 \
 		-A benchmark.sysbench \
 		-A conformance.ltp \
+		-A conformance.kselftest \
 		-A regression.package
 
 # Prebuild riscv64 packages
@@ -159,7 +160,7 @@ $(EXT2_IMAGE):
 
 $(EXFAT_IMAGE):
 	@mkdir -p $(BUILD_DIR)
-	@fallocate -l 64M $(EXFAT_IMAGE)
+	@fallocate -l 512M $(EXFAT_IMAGE)
 	@mkfs.exfat $(EXFAT_IMAGE)
 
 .PHONY: format
--- a/test/initramfs/nix/conformance/default.nix
+++ b/test/initramfs/nix/conformance/default.nix
@@ -7,6 +7,7 @@
     name = "gvisor-prebuilt";
     path = builtins.getEnv "GVISOR_PREBUILT_DIR";
   };
+  kselftest = callPackage ./kselftest.nix { };
 
   package = stdenvNoCC.mkDerivation {
     pname = "conformance";
@@ -19,11 +20,15 @@
       cd $src
       mkdir -p $out
       export INITRAMFS=$out
-      export LTP_PREBUILT_DIR=${ltp}
-      export GVISOR_PREBUILT_DIR=${gvisor}
       export CONFORMANCE_TEST_SUITE=${testSuite}
       export CONFORMANCE_TEST_WORKDIR=${workDir}
       export SMP=${toString smp}
+      ${lib.optionalString (testSuite == "ltp")
+      "export LTP_PREBUILT_DIR=${ltp}"}
+      ${lib.optionalString (testSuite == "gvisor")
+      "export GVISOR_PREBUILT_DIR=${gvisor}"}
+      ${lib.optionalString (testSuite == "kselftest")
+      "export KSELFTEST_PREBUILT_DIR=${kselftest}"}
       make
     '';
   };
--- a/test/initramfs/nix/conformance/kselftest.nix
+++ b/test/initramfs/nix/conformance/kselftest.nix
@@ -1,0 +1,69 @@
+{ lib, stdenv, fetchgit, pkgsBuildBuild, pkgsBuildHost, }:
+stdenv.mkDerivation rec {
+  pname = "kselftest";
+  version = "6.18";
+
+  # Sparse-checkout the kselftest subtree instead of the full ~1 GB kernel tree.
+  # To bump: update `version`, temporarily set `hash = lib.fakeHash;`, and run
+  # the Nix build once to let `fetchgit` report the real hash for this sparse
+  # checkout set.
+  src = fetchgit {
+    url = "https://github.com/torvalds/linux.git";
+    rev = "v${version}";
+    sparseCheckout = [
+      "tools/testing/selftests"
+      "tools/arch"
+      "tools/include"
+      "tools/scripts"
+      "arch"
+      "include"
+      "scripts"
+      "samples/check-exec"
+    ];
+    hash = "sha256-7Ie6f1aNvBo6ipoS/no+IkVhQ/C9bedzQCUoYwM4ED4=";
+  };
+
+  # Explicit allowlist of selftest subsystems Asterinas exercises today.
+  # Keeping this small (rather than building "all") means
+  # a regression in an unlisted subsystem upstream cannot silently change our conformance surface.
+  # To enable a new subsystem,
+  # add it here and triage any resulting build/run failures before landing.
+  baseKselftestTargets =
+    [ "exec" "lsm" "proc" "signal" "splice" "timers" "vDSO" ];
+
+  kselftestTargets = lib.concatStringsSep " " (baseKselftestTargets
+    ++ lib.optionals stdenv.hostPlatform.isx86_64 [ "x86" ]);
+
+  enableParallelBuilding = true;
+
+  nativeBuildInputs = with pkgsBuildBuild; [ gcc_multi rsync ];
+
+  buildInputs = with pkgsBuildHost; [
+    glibc_multi
+    glibc_multi.static
+    libcap.dev
+  ];
+
+  buildPhase = ''
+    runHook preBuild
+    make ARCH=${stdenv.hostPlatform.linuxArch} -j$NIX_BUILD_CORES headers
+    make -C tools/testing/selftests ARCH=${stdenv.hostPlatform.linuxArch} \
+         -j$NIX_BUILD_CORES TARGETS="$kselftestTargets" all
+    runHook postBuild
+  '';
+
+  installPhase = ''
+    runHook preInstall
+    make -C tools/testing/selftests ARCH=${stdenv.hostPlatform.linuxArch} \
+         -j$NIX_BUILD_CORES TARGETS="$kselftestTargets" KSFT_INSTALL_PATH=$out install
+    runHook postInstall
+  '';
+
+  meta = with lib; {
+    description =
+      "Linux in-kernel selftests (kselftest) for Asterinas conformance testing";
+    homepage = "https://docs.kernel.org/dev-tools/kselftest.html";
+    license = licenses.gpl2Only;
+    platforms = platforms.linux;
+  };
+}
--- a/test/initramfs/src/conformance/Makefile
+++ b/test/initramfs/src/conformance/Makefile
@@ -6,25 +6,21 @@ INITRAMFS ?= $(CUR_DIR)/../../build/initramfs
 RUN_BASH := $(CUR_DIR)/run_conformance_test.sh
 TARGET_DIR := $(INITRAMFS)/opt
 CONFORMANCE_TEST_SUITE ?= ltp
-GVISOR_DIR := $(INITRAMFS)/opt/gvisor
-LTP_DIR := $(INITRAMFS)/opt/ltp
+VALID_SUITES := ltp gvisor kselftest
+SUITE_DIR := $(INITRAMFS)/opt/$(CONFORMANCE_TEST_SUITE)
+
+$(if $(filter $(CONFORMANCE_TEST_SUITE),$(VALID_SUITES)),,\
+  $(error CONFORMANCE_TEST_SUITE='$(CONFORMANCE_TEST_SUITE)' not one of: $(VALID_SUITES)))
 
 .PHONY: all
 all: $(TARGET_DIR)
 
-$(TARGET_DIR): $(RUN_BASH) $(GVISOR_DIR) $(LTP_DIR)
+$(TARGET_DIR): $(RUN_BASH) $(SUITE_DIR)
 	@# Copy bash script
 	@cp -f $(RUN_BASH) $@
 
-$(GVISOR_DIR):
-ifeq ($(CONFORMANCE_TEST_SUITE), gvisor)
-	@$(MAKE) --no-print-directory -C gvisor
-endif
-
-$(LTP_DIR):
-ifeq ($(CONFORMANCE_TEST_SUITE), ltp)
-	@$(MAKE) --no-print-directory -C ltp
-endif
+$(SUITE_DIR):
+	@$(MAKE) --no-print-directory -C $(CONFORMANCE_TEST_SUITE)
 
 .PHONY: clean
 clean:
--- a/test/initramfs/src/conformance/kselftest/Makefile
+++ b/test/initramfs/src/conformance/kselftest/Makefile
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: MPL-2.0
+
+KSELFTEST_PREBUILT_DIR ?= /opt/kselftest
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+CUR_DIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+INITRAMFS ?= $(CUR_DIR)/../../../build/initramfs
+TARGET_DIR := $(INITRAMFS)/opt/kselftest
+RUN_BASH := $(CUR_DIR)/run_kselftest.sh
+BLOCK_LIST := $(CUR_DIR)/blocklists
+EXT2_BLOCK_LIST := $(CUR_DIR)/blocklists.ext2
+EXFAT_BLOCK_LIST := $(CUR_DIR)/blocklists.exfat
+
+$(if $(wildcard $(BLOCK_LIST)),,$(error missing default blocklist: $(BLOCK_LIST)))
+
+.PHONY: all
+all: $(TARGET_DIR)
+
+$(TARGET_DIR): $(RUN_BASH) $(BLOCK_LIST) $(EXT2_BLOCK_LIST) $(EXFAT_BLOCK_LIST)
+	@rm -rf $@ && mkdir -p $@
+	@# Copy testcases
+	@cp -rf $(KSELFTEST_PREBUILT_DIR)/* $@
+	@# Copy blocklists
+	@cp -f $(BLOCK_LIST) $@/blocklists
+	@cp -f $(EXT2_BLOCK_LIST) $@/blocklists.ext2
+	@cp -f $(EXFAT_BLOCK_LIST) $@/blocklists.exfat
+	@# Copy bash script
+	@cp -f $(RUN_BASH) $@
+
+.PHONY: clean
+clean:
+	@rm -rf $(TARGET_DIR)
--- a/test/initramfs/src/conformance/kselftest/run_kselftest.sh
+++ b/test/initramfs/src/conformance/kselftest/run_kselftest.sh
@@ -1,0 +1,201 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+echo "=== Kselftest Runner Started ==="
+
+KSELFTEST_DIR=$(dirname "$0")
+DEFAULT_BLOCKLISTS="$KSELFTEST_DIR/blocklists"
+COMBINED_BLOCKLISTS=$(mktemp)
+RESOLVED_BLOCKLISTS=$(mktemp)
+AVAILABLE_TESTS=$(mktemp)
+FAILED_TESTS=$(mktemp)
+
+trap 'rm -f "$COMBINED_BLOCKLISTS" "$RESOLVED_BLOCKLISTS" "$AVAILABLE_TESTS" "$FAILED_TESTS"' EXIT
+
+# The `kselftest-list.txt` is a generated file produced by the kselftest build.
+#
+# Each line describes a single kselftest case in the form:
+#     <test-dir>:<test-command>
+#
+# Example lines (matching enabled TARGETS in kselftest.nix):
+#   exec:binfmt_script.py
+#   timers:posix_timers
+#   vDSO:vdso_test_abi
+TESTS="$KSELFTEST_DIR/kselftest-list.txt"
+if [ ! -r "$TESTS" ]; then
+    echo "$0: missing $TESTS; kselftest build is broken" >&2
+    exit 1
+else
+    available_count=$(grep -cve '^$' "$TESTS")
+    echo "Found $available_count available test cases"
+fi
+
+if [ ! -r "$DEFAULT_BLOCKLISTS" ]; then
+    echo "$0: missing $DEFAULT_BLOCKLISTS; kselftest blocklist is broken" >&2
+    exit 1
+fi
+
+cat "$DEFAULT_BLOCKLISTS" > "$COMBINED_BLOCKLISTS"
+for extra_file in $EXTRA_BLOCKLISTS ; do
+    extra_blocklists="$KSELFTEST_DIR/$extra_file"
+    if [ -r "$extra_blocklists" ]; then
+        printf '\n' >> "$COMBINED_BLOCKLISTS"
+        cat "$extra_blocklists" >> "$COMBINED_BLOCKLISTS"
+    else
+        echo "Warning: extra blocklist not found: $extra_blocklists" >&2
+    fi
+done
+
+echo "Processing blocklists..."
+while IFS= read -r line || [ -n "$line" ]; do
+    line=${line#"${line%%[![:space:]]*}"}
+    line=${line%"${line##*[![:space:]]}"}
+    case "$line" in
+        "" | "#"*)
+            continue ;;
+        *:*)
+            dir="${line%%:*}"
+            command="${line#*:}"
+            if [ "$command" = "*" ]; then
+                awk -F: -v d="$dir" '$1 == d { print }' "$TESTS" >> "$RESOLVED_BLOCKLISTS"
+            else
+                printf '%s\n' "$line" >> "$RESOLVED_BLOCKLISTS"
+            fi
+            ;;
+        *)
+            echo "Error: Invalid format in blocklist: $line" >&2
+            exit 1
+            ;;
+    esac
+done < "$COMBINED_BLOCKLISTS"
+
+sort -u -o "$RESOLVED_BLOCKLISTS" "$RESOLVED_BLOCKLISTS"
+if [ -s "$RESOLVED_BLOCKLISTS" ]; then
+    blocked_count=$(wc -l < "$RESOLVED_BLOCKLISTS")
+    grep -vxFf "$RESOLVED_BLOCKLISTS" "$TESTS" | grep -v '^$' > "$AVAILABLE_TESTS"
+else
+    blocked_count=0
+    grep -v '^$' "$TESTS" > "$AVAILABLE_TESTS"
+fi
+echo "Total blocklist entries processed: $blocked_count"
+
+run_count=$(wc -l < "$AVAILABLE_TESTS")
+if [ "$run_count" -eq 0 ]; then
+    echo "$0: no test cases survived blocklist filtering" >&2
+    exit 1
+fi
+echo "Test cases to be executed: $run_count"
+
+echo "================================"
+
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+NC='\033[0m'
+total_tests=$run_count
+passed_tests=0
+failed_tests=0
+
+if ! command -v timeout >/dev/null 2>&1; then
+    echo "$0: missing timeout command" >&2
+    exit 1
+fi
+
+KSELFTEST_TIMEOUT=${KSELFTEST_TIMEOUT:-300}
+run_test() {
+    test_dir=$1
+    shift
+
+    (
+        cd "$test_dir" &&
+        timeout "$KSELFTEST_TIMEOUT" "$@"
+    )
+}
+
+# Use a dedicated subdirectory under the selected workdir.
+TEST_WORKDIR="${CONFORMANCE_TEST_WORKDIR:-/tmp}/kselftest"
+rm -rf "$TEST_WORKDIR"
+mkdir -p "$TEST_WORKDIR"
+if [ ! -d "$TEST_WORKDIR" ]; then
+    echo "$0: failed to create TEST_WORKDIR: $TEST_WORKDIR" >&2
+    exit 1
+fi
+
+dirs=$(cut -d: -f1 "$AVAILABLE_TESTS" | sort -u)
+for dir in $dirs ; do
+    TEST_DIR="$TEST_WORKDIR/$dir"
+    echo "Running tests in dir: $TEST_DIR"
+    if ! cp -rL "$KSELFTEST_DIR/$dir" "$TEST_DIR"; then
+        echo "$0: failed to copy kselftest dir: $KSELFTEST_DIR/$dir" >&2
+        exit 1
+    fi
+
+    commands=$(awk -F: -v d="$dir" '$1 == d { sub(/^[^:]*:/, ""); print }' "$AVAILABLE_TESTS")
+    for command in $commands ; do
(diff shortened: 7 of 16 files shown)


test/initramfs/Makefile
```diff
@@ -69,6 +70,7 @@ $(INITRAMFS_IMAGE): $(INITRAMFS)
 		--argstr basicTestPlatform $(BASIC_TEST_PLATFORM) \
 		--arg enableBenchmark $(ENABLE_BENCHMARK) \
 		--arg enableSyscallTest $(ENABLE_SYSCALL_TEST) \
+		--arg enableKselfTest $(ENABLE_KSELFTEST_TEST) \
```
tatetian: I think kselftest should join the league of gvisor and LTP test suites and share the same option `enableSyscallTest` with them, rather than adding a new option. Because kselftest are very much alike to gvisor and LTP tests: they are all smoke/conformance tests that should be run for every PR.

test/initramfs/src/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+echo "=== Kselftest Runner Started ==="
+
+BASE_DIR=@kselftest@
+
+TESTS="$BASE_DIR"/kselftest-list.txt
```
tatetian: Where is this file?

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
```
tatetian, line 1: Both gVisor and LTP directories have `Makefile`. I am wondering if the Makefile is still any good. If yes, then `kselftest` should have one? If not, we should consider delete those `Makefile`s.

test/initramfs/src/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+echo "=== Kselftest Runner Started ==="
+
+BASE_DIR=@kselftest@
```
tatetian: Either mark this file as a template explicitly (e.g., `run_kselftest.sh.template`) or make this file executable even without having `@kselftest@` replaced.

In addition, the name `BASE_DIR` is kind of vague: the project root could be the base, the current directory could also be the base. Rename this variable to `KSELFTEST_INSTALL_DIR`.

test/initramfs/nix/kselftest.nix
```diff
@@ -0,0 +1,90 @@
...
+      popt
+      zlib.dev
+    ];
+
+    configurePhase = ''
+      runHook preConfigure
+
+      patchShebangs tools/net/ynl/pyynl
+      sed -i '206d' tools/testing/selftests/net/Makefile # FIXME: bpf build fails
+      sed -i '17d' tools/testing/selftests/cgroup/Makefile # FIXME: test_memcontrol build fails
```
tatetian: Make the change with a patch file. Having a patch file to make it more apparent what we change. And you can explain more about why you make the changes in the patch file.

test/initramfs/src/kselftest/run_kselftest.sh
tatetian: Could you add a comment to describe the content or format of this file? A reader of this script cannot make sense of this script unless he/she knows what is inside this file. I assume that this file is generated by the build process of kselftest. So this file cannot be found either in Asterinas or Linux repo.

test/initramfs/Makefile
tatetian, line 91: Similar to gvisor and LTP tests, kselftest is also a smoke test that is quick to run and intended to capture potential bugs in every change to the project. So it is important that kselftest is not rebuilt for every CI run.

Gvisor tests are prebuilt in the `asterinas/asterinas` Docker image manually, whereas the LTP test artifacts are cached in the Nix store when building the `asterinas/nix` image.

https://github.com/asterinas/asterinas/blob/6882eb73f4d0d5a934c8c89cb7897dc8213e1e37/tools/docker/nix/Dockerfile#L31-L48

As I understand it, you need to ensure that the `x86_64_pkgs` target enables kselftest so that the kselftest artifacts are cached in the `asterinas/nix` image.

tatetian (commented):
Good job on adding kselftest.

.github/workflows/test_x86.yml
```diff
@@ -127,6 +127,20 @@ jobs:
...
+        with:
+          auto_test: 'conformance'
+          release: ${{ !contains(matrix.release, 'false') }}
+          enable_kvm: ${{ !contains(matrix.enable_kvm, 'false') }}
+          smp: ${{ matrix.smp }}
+          netdev: ${{ matrix.netdev || 'tap' }}
+          scheme: ${{ matrix.scheme }}
+          extra_blocklists: ${{ matrix.extra_blocklists }}
+          syscall_test_suite: 'kselftest'
+          syscall_test_workdir: ${{ matrix.syscall_test_workdir }}
```
tatetian: **Kselftest is never actually run — unknown composite-action inputs** (blocker): `action.yml` declares `conformance_test_suite` / `conformance_test_workdir`; the names used here are silently dropped. The downstream `make run_kernel AUTO_TEST=conformance` then falls back to `CONFORMANCE_TEST_SUITE ?= ltp` and re-runs LTP.

Fix:

```diff
-          syscall_test_suite: 'kselftest'
-          syscall_test_workdir: ${{ matrix.syscall_test_workdir }}
+          conformance_test_suite: 'kselftest'
+          conformance_test_workdir: ${{ matrix.conformance_test_workdir }}
```

Verify the fix by confirming the runner log contains `CONFORMANCE_TEST_SUITE=kselftest`. (If you adopt the DRY restructure further down, this typo becomes structurally impossible.)

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,111 @@
...
```
tatetian, line 1: **File committed as non-executable** (blocker): `git ls-tree` shows mode `100644` vs `100755` on sibling `run_*.sh`. `run_conformance_test.sh:26` invokes it directly (no `sh` prefix), and `kselftest/Makefile:21`'s `cp -f` preserves the bit, so the copy in the initramfs also fails to exec. Currently masked by the CI-input bug above. Fix: `chmod +x` and re-commit.

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,111 @@
...
+#   cgroup:test_stress.sh
+#   exec:binfmt_script.py
+TESTS="$KSELFTEST_DIR"/kselftest-list.txt
+if [ ! -r "$TESTS" ] ; then
+	echo "$0: Could not find list of tests to run ($TESTS)" >&2
+	available=""
+else
+	available="$(cat "$TESTS")"
+	echo "Found $(echo "$available" | grep -v '^$' | wc -l) available test cases"
+fi
```
tatetian: **Silent success when the test list is missing or empty** (significant): When `kselftest-list.txt` is missing, `available=""` and `testcases=""`, but `run_count=$(echo "" | wc -l)` evaluates to `1` (not 0) — so the guard at line 59 doesn't fire. Control falls through to the `for dir in $dirs` loop, which iterates zero times on empty input, then reaches the final `exit 0` at line 110. CI reports pass on a broken build. Both paths should hard-fail:

```sh
if [ ! -r "$TESTS" ]; then
    echo "$0: missing $TESTS; kselftest build is broken" >&2
    exit 1
fi
# ... later ...
if [ "$run_count" -eq 0 ]; then
    echo "$0: no test cases survived blocklist filtering" >&2
    exit 1
fi
```

Paired with the `kselftest-all` tolerance comment below, this is the other half of the "zero-test green build" problem.

test/initramfs/nix/conformance/kselftest/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    runHook postConfigure
+  '';
+
+  buildPhase = ''
+    runHook preBuild
+
+    make kselftest-all
+
+    runHook postBuild
+  '';
```
tatetian: **`make kselftest-all` tolerates per-subdir failures and is single-threaded** (significant): Upstream make continues past subdir compile failures and silently drops those binaries; combined with the silent-zero-tests runner bug, a broken kselftest build still passes CI.

Fix — pass explicit `TARGETS=` and `enableParallelBuilding`:

```nix
stdenv.mkDerivation rec {
  pname = "kselftest";
  version = "6.18";

  # Only these subsystems are exercised by Asterinas today. Adding a
  # new one here makes the build fail-hard if the subsystem breaks —
  # which is what we want.
  kselftestTargets = "acct cgroup exec fpu ipc pidfd proc ...";

  enableParallelBuilding = true;

  buildPhase = ''
    runHook preBuild
    make -C tools/testing/selftests TARGETS="$kselftestTargets" kselftest-all
    runHook postBuild
  '';

  installPhase = ''
    runHook preInstall
    make -C tools/testing/selftests TARGETS="$kselftestTargets" \
         KSFT_INSTALL_PATH=$out kselftest-install
    runHook postInstall
  '';
}
```

`make` stops on the first target failure (no `-k`), so a regression in any listed subsystem fails the nix build loudly. This also shrinks the install set by gigabytes, which further motivates the blocklist-consolidation fix below.

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,111 @@
...
+				else
+					blocklists="$blocklists $line"
+				fi
+                ;;
+            *)
+                echo "Warning: Invalid format in blocklist: $line" >&2
+                continue ;;
+		esac
+	done < "$blocklist_file"
+done
```
tatetian: **`EXTRA_BLOCKLISTS_DIRS` and `CONFORMANCE_TEST_WORKDIR` are ignored** (significant): `run_gvisor_test.sh:25-29` and `run_ltp_test.sh:6` both read these; this runner doesn't. So matrix entries for `extra_blocklists: 'blocklists.ext2'`/`blocklists.exfat` and `conformance_test_workdir: '/ext2'` have no effect for kselftest.

Fix — load per-scheme blocklists the same way gVisor does, and use the workdir:

```sh
KSELFTEST_DIR=$(dirname "$0")
TEST_TMP_DIR=${CONFORMANCE_TEST_WORKDIR:-/tmp}
BLOCKLISTS_DIR="$KSELFTEST_DIR/blocklists"

# ... load primary blocklists from $BLOCKLISTS_DIR ...

# Merge extra per-scheme blocklists (e.g. blocklists.ext2, blocklists.exfat)
for extra_dir in $EXTRA_BLOCKLISTS_DIRS ; do
    [ -d "$KSELFTEST_DIR/$extra_dir" ] || continue
    for f in "$KSELFTEST_DIR/$extra_dir"/*; do
        [ -e "$f" ] || continue
        # ... same parsing as the primary loop ...
    done
done

# Use $TEST_TMP_DIR as CWD / temp for any test that writes files.
```

Also fix the stray unquoted `$(dirname $0)` on line 29 by reusing line 7's value: `BLOCKLISTS_DIR="$KSELFTEST_DIR/blocklists"`.

test/initramfs/src/conformance/kselftest/blocklists/net
tatetian, line 1: **Collapse 90 blocklist files into one** (significant): Every entry already carries its directory prefix (`fpu:test_fpu`, `filesystems/fuse:fusectl_test`), so the split across 90 per-suite files buys nothing. Most files carry a single entry (often a lone `dir:*` wildcard), and the script just iterates all of them in order. A single flat `blocklist` file (matching the LTP/gVisor convention) removes that noise, eliminates the missing-trailing-newline problem, and makes `grep`/`wc` trivial for reviewers.

Fix (from the repo root):

```sh
cd test/initramfs/src/conformance/kselftest/blocklists
{ echo "# kselftest blocklist. Format: <dir>:<test-name>."
  echo "# Use '<dir>:*' to skip every test under <dir>."
  echo
  for f in *; do
      printf '# --- %s ---\n' "$f"
      cat "$f"
      echo    # ensure a trailing newline between sections
  done
} > ../blocklist
cd .. && rm -rf blocklists && mv blocklist blocklists
```

Then update `run_kselftest.sh` to read a single file rather than loop over a directory, and move the `dir:*` wildcard comment to the top of the file.

Orthogonally, for suites a guest OS can't exercise at all (`kvm`, `vfio`, `tpm2`, `sgx`, `tdx`, `amd-pstate`, `intel_pstate`, `breakpoints`, `lkdtm`, `livepatch`, `kexec`, ...), filter them at build time via the `TARGETS=` approach in the nix-derivation fix above — they won't even be compiled, so the blocklist doesn't need to track them at all.

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,111 @@
...
+BLOCKLISTS_DIR="$(dirname $0)/blocklists"
+for blocklist_file in "$BLOCKLISTS_DIR"/*; do
+    while IFS= read -r line || [ -n "$line" ]; do
+        line=$(echo "$line" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
+		case "$line" in
+			"#"*)
+				continue ;;
+			*:*)
+				dir=$(echo "$line" | cut -d: -f1)
+                command=$(echo "$line" | cut -d: -f2)
```
tatetian: **Subshell-per-line is expensive** (minor): Inside the `while read` loop, each iteration forks two `cut`s (and an `echo`). Parameter expansion avoids all of them:

```sh
*:*)
    dir="${line%%:*}"
    command="${line#*:}"
```

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,111 @@
...
+					blocklists="$blocklists $line"
+				fi
+                ;;
+            *)
+                echo "Warning: Invalid format in blocklist: $line" >&2
+                continue ;;
+		esac
+	done < "$blocklist_file"
+done
+blocklists="$(echo "$blocklists" | tr ' ' '\n' | grep -v '^$' | sort | uniq)"
```
tatetian: **`blocked_count=1` on empty blocklist; `grep -vxF` relies on `-x`** (minor):
- `echo "" | wc -l` returns 1. Guard: `[ -z "$blocklists" ] && blocked_count=0 || blocked_count=$(printf '%s\n' "$blocklists" | wc -l)`.
- `grep -vxF "$blocklists"` with an empty pattern list is correct today (with `-x`, an empty pattern matches only empty lines, which the trailing `grep -v '^$'` would strip anyway), but it relies on keeping the `-x` flag — if `-x` is ever dropped, an empty pattern would match *every* line and filter out all tests. Safer: write patterns to a file and use `grep -vxFf`, which errors loudly on an empty file.

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,111 @@
...
+failed_tests=0
+failed_list=""
+
+dirs=$(echo "$testcases" | cut -d: -f1 | sort | uniq)
+for dir in $dirs ; do
+	echo "Running tests in dir: $dir"
+	commands=$(echo "$testcases" | grep "^$dir:" | cut -d: -f2)
+	for command in $commands ; do
+		echo "[ PROCESS  ]: $dir:$command"
+		if [ -d "$KSELFTEST_DIR/$dir" ] && [ -x "$KSELFTEST_DIR/$dir/$command" ]; then
```
tatetian: **Regex injection on `$dir`; `.py`/`.sh` entries falsely reported as failures** (minor):

1. `grep "^$dir:"` interpolates into a regex. Swap to fixed-string with an explicit colon separator:

   ```sh
   commands=$(printf '%s\n' "$testcases" | awk -F: -v d="$dir" '$1==d {print $2}')
   ```

2. `kselftest-list.txt` contains `.py` / `.sh` / wrapper entries that aren't `+x`. They fail the `-x` check and get counted as "executable not found". Distinguish "missing" from "not executable", and invoke via the interpreter when needed:

   ```sh
   bin="$KSELFTEST_DIR/$dir/$command"
   if   [ ! -e "$bin" ];                 then status=missing
   elif [ -x "$bin" ];                   then (cd "$KSELFTEST_DIR/$dir" && ./"$command"); status=$?
   else  # Present but not executable — run via interpreter based on suffix.
         case "$command" in
             *.py) (cd "$KSELFTEST_DIR/$dir" && python3 "./$command"); status=$? ;;
             *.sh) (cd "$KSELFTEST_DIR/$dir" && sh        "./$command"); status=$? ;;
             *)    status=noexec ;;
         esac
   fi
   ```

test/initramfs/src/conformance/kselftest/run_kselftest.sh
tatetian, line 1: **`echo -e` is not POSIX `sh`** (minor): Under dash, `echo -e "foo"` prints `-e foo`. Busybox ash happens to accept `-e`, but either switch the shebang or use `printf`.

Fix — stay on `/bin/sh` and use `printf '%b\n'`:

```diff
-			echo -e "[  ${GREEN}PASSED${NC}  ]: $dir:$command"
+			printf '[  %bPASSED%b  ]: %s:%s\n' "$GREEN" "$NC" "$dir" "$command"
...
-			failed_list="$failed_list$dir:$command\n"
+			failed_list="$failed_list$dir:$command
+"
...
-echo -e "$GREEN$passed_tests$NC of $GREEN$total_tests$NC test cases passed."
+printf '%b%s%b of %b%s%b test cases passed.\n' \
+    "$GREEN" "$passed_tests" "$NC" "$GREEN" "$total_tests" "$NC"
```

Alternatively, switch the shebang to `#!/bin/bash` and leave `echo -e` as-is. Also run `shfmt` over the file — lines 32-49 mix tabs with 12- and 16-space indentation on adjacent lines (e.g., the `case` arms are tab-indented but `command=$(...)` and the `*)` default arm are space-indented).

test/initramfs/nix/conformance/kselftest/default.nix
tatetian, line 1: **Nix derivation nits** (minor, grouped). Fixes below, to be folded on top of the `kselftestTargets` / `enableParallelBuilding` change from the earlier comment:

```nix
{ lib, stdenv, fetchgit, pkgsBuildBuild, pkgsBuildHost, }:
stdenv.mkDerivation rec {
  pname = "kselftest";
  version = "6.18";

  # Sparse-checkout the kselftest subtree instead of the full ~1 GB kernel tree.
  # To bump: update `version` and refresh `hash` with
  #   nix-prefetch-git --url https://github.com/torvalds/linux \
  #     --rev v<new> --sparse-checkout "tools/testing/selftests" --leave-dotGit
  src = fetchgit {
    url = "https://github.com/torvalds/linux.git";
    rev = "v${version}";
    sparseCheckout = [ "tools/testing/selftests" "tools/include" "scripts" ];
    hash = "sha256-...";  # regenerate after switching fetcher
  };

  patches = [ ./0001-kselftest-drop-memcontrol-and-bpf-test-gen.patch ];

  enableParallelBuilding = true;

  # ... nativeBuildInputs / buildInputs unchanged ...

  meta = with lib; {
    description = "Linux in-kernel selftests (kselftest) for Asterinas conformance testing";
    homepage = "https://docs.kernel.org/dev-tools/kselftest.html";
    license = licenses.gpl2Only;
    platforms = platforms.linux;
  };
}
```

Addresses:
1. ~1 GB Linux tarball → sparse checkout of `tools/testing/selftests` + its transitive deps (`tools/include`, `scripts`).
2. `lib` now used by `meta`.
3. `meta` block added.
4. `enableParallelBuilding` covered by the earlier patch.
5. `hash` must be regenerated after switching fetcher.
6. Version-bump procedure documented inline.

test/initramfs/nix/conformance/kselftest/0001-Skip-build-targets-that-would-fail.patch
```diff
@@ -0,0 +1,37 @@
+From 5335f115936c018c2670d6c76689a365bb7e624e Mon Sep 17 00:00:00 2001
+From: Qingsong Chen <changxian.cqs@antgroup.com>
+Date: Thu, 5 Mar 2026 07:28:58 +0000
+Subject: [PATCH] Skip build targets that would fail
```
tatetian: **Patch has no rationale and is rebase-fragile** (minor): "Would fail" doesn't say why. The next maintainer bumping the pin has to rediscover. Two-line deletions don't really warrant a standalone patch — inline them as `postPatch` with a comment, which also anchors on surrounding text rather than exact line numbers (survives upstream reshuffles).

Fix — delete the `.patch` file and add to `default.nix`:

```nix
# The following test targets fail to build under the nix sandbox and are not
# relevant to Asterinas; drop them rather than carrying skip-flags at runtime.
#   - cgroup/test_memcontrol: needs runtime cgroup v2 memory controller
#     which the nix builder does not expose.
#   - net/*.bpf.o:            needs libbpf + clang BPF backend; the cross
#     toolchain we use for selftests does not ship one.
postPatch = ''
  substituteInPlace tools/testing/selftests/cgroup/Makefile \
      --replace-fail "TEST_GEN_PROGS += test_memcontrol" ""
  substituteInPlace tools/testing/selftests/net/Makefile \
      --replace-fail 'TEST_GEN_FILES += $(patsubst %.c,%.o,$(wildcard *.bpf.c))' ""
'';
```

`--replace-fail` hard-errors if the anchor string isn't found, so an upstream rename is surfaced immediately instead of silently dropping the workaround.

test/initramfs/src/conformance/Makefile
```diff
@@ -26,6 +27,11 @@ ifeq ($(CONFORMANCE_TEST_SUITE), ltp)
 	@$(MAKE) --no-print-directory -C ltp
 endif
 
+$(KSELFTEST_DIR):
+ifeq ($(CONFORMANCE_TEST_SUITE), kselftest)
+	@$(MAKE) --no-print-directory -C kselftest
+endif
```
tatetian: **Empty recipe when suite not selected makes the dep graph meaningless** (minor): Pre-existing pattern (LTP/gvisor too), but worth fixing while kselftest is being added. `$(TARGET_DIR)` lists all three dirs as prereqs, two of which are no-ops; make sees the directory "doesn't exist", runs no recipe, undefined mtime behavior. Cleaner:

```make
SUITE_DIR := $(INITRAMFS)/opt/$(CONFORMANCE_TEST_SUITE)
$(TARGET_DIR): $(RUN_BASH) $(SUITE_DIR)
$(SUITE_DIR):
	@$(MAKE) --no-print-directory -C $(CONFORMANCE_TEST_SUITE)
```

test/initramfs/src/conformance/kselftest/Makefile
```diff
@@ -0,0 +1,25 @@
...
+CUR_DIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+INITRAMFS ?= $(CUR_DIR)/../../../build/initramfs
+TARGET_DIR := $(INITRAMFS)/opt/kselftest
+RUN_BASH := $(CUR_DIR)/run_kselftest.sh
+BLOCK_LIST := $(CUR_DIR)/blocklists
+
+.PHONY: all
+all: $(TARGET_DIR)
+
+$(TARGET_DIR): $(RUN_BASH) $(BLOCK_LIST)
```
tatetian: **`BLOCK_LIST` is a directory; editing a blocklist won't trigger rebuild** (minor): `BLOCK_LIST := $(CUR_DIR)/blocklists` tracks the directory's mtime (files added/removed), not content edits — a big miss given how much of this PR's value is in the blocklists. Use `BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists/*)`.

Also, `KSELFTEST_PREBUILT_DIR ?= /opt/kselftest` collides with the install target. If someone runs this with `INITRAMFS=/`, the recipe wipes the source before copying. Same footgun as LTP, but worth a guard: `$(if $(filter $(TARGET_DIR),$(KSELFTEST_PREBUILT_DIR)),$(error ...))`.

test/initramfs/src/conformance/kselftest/blocklists/acct
tatetian, line 1: **Missing trailing newlines** (minor): Every new blocklist file ends mid-line. Parser tolerates it via `|| [ -n "$line" ]`, but it's a POSIX violation and noisy in `git diff`. Resolved automatically if the files are collapsed into one (see the blocklist-design comment); otherwise: `sed -i -e '$a\' blocklists/*`. Either way, document the `dir:*` wildcard convention inline rather than burying it in a run-script comment.

tatetian (commented):
Thanks for the revision, although there are some major issues that need your attention.

**kselftest is not actually exercised by CI today.** It was LTP test that run.

image

This was caused by two compounding bugs:

1. `test_x86.yml` passes `syscall_test_suite` / `syscall_test_workdir`, but
`action.yml` only declares `conformance_test_suite` / `conformance_test_workdir`.
Unknown inputs are silently dropped, so the step runs with the default
`CONFORMANCE_TEST_SUITE=ltp` — it just re-runs LTP a third time.
2. `run_kselftest.sh` is committed as `100644` (siblings are `100755`), so
`run_conformance_test.sh`'s direct exec would fail. (1) masks (2).

Both must be fixed together. On top of that, the runner exits `0` when
`kselftest-list.txt` is missing, and `make kselftest-all` tolerates
per-subdir build failures — three paths to a "green build that ran zero tests."

Secondary themes:

- **Blocklist design.** 90 files, most carrying a single entry and ~11 a
single `dir:*` line disabling the whole suite. Since every entry already
encodes the directory (`fpu:test_fpu`), collapse all blocklists into one
flat file — consistent with LTP/gVisor, eliminates the trailing-newline
noise, and makes `grep`/`wc` trivial. Orthogonally: for suites a guest
OS can't exercise at all (`kvm`, `vfio`, `tpm2`, `sgx`, `tdx`, `*_pstate`),
filter them at build time via `TARGETS=` so they aren't compiled or
shipped.
- **Runner ergonomics.** `EXTRA_BLOCKLISTS_DIRS` and `CONFORMANCE_TEST_WORKDIR`
(honored by LTP/gVisor) are ignored, so matrix `extra_blocklists` /
`conformance_test_workdir` are ineffective here.
- **Reproducibility.** The nix derivation pulls the full ~1 GB torvalds/linux
tarball, builds single-threaded, and patches two failures without
explanation.
- **Docs.** Commit/PR body is one sentence — please state which suites run,
how many pass, and the update procedure for the Linux pin.

test/initramfs/src/conformance/kselftest/run_kselftest.sh
cqs21: > `run_gvisor_test.sh:25-29` and `run_ltp_test.sh:6` both read these; this runner doesn't.

The runner ignores `EXTRA_BLOCKLISTS_DIRS` and `CONFORMANCE_TEST_WORKDIR` because `kselftest` cases do not use an external temp/work directory mechanism. Passing these knobs is misleading and has no effect. Accordingly, `extra_blocklists` and `conformance_test_workdir` are dropped from the workflow invocation.

test/initramfs/nix/conformance/default.nix
```diff
@@ -7,6 +7,7 @@
     name = "gvisor-prebuilt";
     path = builtins.getEnv "GVISOR_PREBUILT_DIR";
   };
+  kselftest = callPackage ./kselftest.nix { };
```
tatetian, line 10: **Eager kselftest build for every test suite** (significant): `export KSELFTEST_PREBUILT_DIR=${kselftest}` in `buildCommand` is an unconditional string interpolation, so Nix realizes the `kselftest` derivation (a full Linux kernel fetch + selftest build) on every `conformance` instantiation — including `testSuite = "ltp"` and `testSuite = "gvisor"`. LTP and gVisor have the same smell with their own prebuilt paths; kselftest amplifies it because the selftest build dominates wall time.

Fix — gate each `*_PREBUILT_DIR` export on `testSuite`:

```nix
buildCommand = ''
  cd $src
  mkdir -p $out
  export INITRAMFS=$out
  export CONFORMANCE_TEST_SUITE=${testSuite}
  export CONFORMANCE_TEST_WORKDIR=${workDir}
  export SMP=${toString smp}
  ${lib.optionalString (testSuite == "ltp")       "export LTP_PREBUILT_DIR=${ltp}"}
  ${lib.optionalString (testSuite == "gvisor")    "export GVISOR_PREBUILT_DIR=${gvisor}"}
  ${lib.optionalString (testSuite == "kselftest") "export KSELFTEST_PREBUILT_DIR=${kselftest}"}
  make
'';
```

test/initramfs/nix/conformance/kselftest.nix
```diff
@@ -0,0 +1,66 @@
...
+    glibc_multi.static
+    libcap.dev
+  ];
+
+  buildPhase = ''
+    runHook preBuild
+    make headers
+    make -C tools/testing/selftests TARGETS="$kselftestTargets" all
+    runHook postBuild
+  '';
```
tatetian, line 53: **`enableParallelBuilding` and `ARCH` not reaching the inner `make`** (minor): `enableParallelBuilding = true` causes Nix to pass `-jN` to the default `buildPhase`'s outer `make` — but this derivation overrides `buildPhase` with its own explicit `make headers` / `make -C tools/testing/selftests ... all` invocations, neither of which inherits `$NIX_BUILD_CORES`. `make headers` also infers `ARCH` from `uname -m`, which is wrong for cross-builds.

Fix — pass `-j` and `ARCH` explicitly (and mirror in `installPhase`):

```nix
buildPhase = ''
  runHook preBuild
  make ARCH=${stdenv.hostPlatform.linuxArch} -j$NIX_BUILD_CORES headers
  make -C tools/testing/selftests ARCH=${stdenv.hostPlatform.linuxArch} \
       -j$NIX_BUILD_CORES TARGETS="$kselftestTargets" all
  runHook postBuild
'';
```

test/initramfs/nix/conformance/kselftest.nix
```diff
@@ -0,0 +1,66 @@
...
+      "include"
+      "scripts"
+      "samples/check-exec"
+    ];
+    hash = "sha256-7Ie6f1aNvBo6ipoS/no+IkVhQ/C9bedzQCUoYwM4ED4=";
+  };
+
+  # Only these subsystems are exercised by Asterinas today. Adding a
+  # new one here makes the build fail-hard if the subsystem breaks —
+  # which is what we want.
```
tatetian: **Comment reads as "we want the build to be fragile"** (minor): The wording says adding a subsystem makes the build fail-hard, which inverts the intent. What's actually desired is: keeping the allowlist small means upstream churn in unexercised subsystems can't silently alter Asterinas's conformance surface, and a regression in a listed subsystem must be triaged rather than auto-skipped.

Fix:

```nix
# Explicit allowlist of selftest subsystems Asterinas exercises today.
# Keeping this small (rather than building "all") means
# a regression in an unlisted subsystem upstream cannot silently change our conformance surface.
# To enable a new subsystem,
# add it here and triage any resulting build/run failures before landing.
baseKselftestTargets =
  [ "exec" "lsm" "proc" "signal" "splice" "timers" "vDSO" ];
```

test/initramfs/src/conformance/Makefile
```diff
@@ -6,25 +6,17 @@ INITRAMFS ?= $(CUR_DIR)/../../build/initramfs
...
-ifeq ($(CONFORMANCE_TEST_SUITE), gvisor)
-	@$(MAKE) --no-print-directory -C gvisor
-endif
-
-$(LTP_DIR):
-ifeq ($(CONFORMANCE_TEST_SUITE), ltp)
-	@$(MAKE) --no-print-directory -C ltp
-endif
+$(SUITE_DIR):
+	@$(MAKE) --no-print-directory -C $(CONFORMANCE_TEST_SUITE)
```
tatetian, line 23: **Unknown `CONFORMANCE_TEST_SUITE` produces a confusing `make: *** No such file or directory`** (minor): A typo like `CONFORMANCE_TEST_SUITE=kseftest` falls into `make -C kseftest`, and the user gets a generic directory-not-found. An explicit `$(error)` names the variable so the user sees the real problem.

Fix:

```make
VALID_SUITES := ltp gvisor kselftest
$(if $(filter $(CONFORMANCE_TEST_SUITE),$(VALID_SUITES)),,\
  $(error CONFORMANCE_TEST_SUITE='$(CONFORMANCE_TEST_SUITE)' not one of: $(VALID_SUITES)))
```

test/initramfs/src/conformance/kselftest/Makefile
```diff
@@ -0,0 +1,34 @@
...
+INITRAMFS ?= $(CUR_DIR)/../../../build/initramfs
+TARGET_DIR := $(INITRAMFS)/opt/kselftest
+RUN_BASH := $(CUR_DIR)/run_kselftest.sh
+BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists/*)
+EXT2_BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists.ext2/*)
+EXFAT_BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists.exfat/*)
+
+NORMALIZED_KSELFTEST_PREBUILT_DIR := $(patsubst %/,%,$(abspath $(KSELFTEST_PREBUILT_DIR)))
+NORMALIZED_TARGET_DIR := $(patsubst %/,%,$(abspath $(TARGET_DIR)))
+$(if $(filter $(NORMALIZED_TARGET_DIR),$(NORMALIZED_KSELFTEST_PREBUILT_DIR)),$(error KSELFTEST_PREBUILT_DIR must not equal TARGET_DIR: $(NORMALIZED_TARGET_DIR)))
```
tatetian: **Drop the `KSELFTEST_PREBUILT_DIR == TARGET_DIR` guard — over-engineered for the scenario it protects against** (minor): This guard was added in response to the previous review flagging that `make INITRAMFS=/ ...` would collide `/opt/kselftest` (prebuilt default) with `$(INITRAMFS)/opt/kselftest` (target) and `rm -rf` the prebuilt tree. On reflection the scenario is narrow: the nix-driven CI path always passes both as distinct store paths and can't collide, and a developer hand-running `make INITRAMFS=/` is unusual enough that three lines of defensive code + a duplicated copy in `ltp/Makefile:15` aren't worth their footprint. Delete the guard in both Makefiles. My previous-review note that suggested adding it was over-cautious.

Fix — remove lines 13-15 (and the matching block in `ltp/Makefile:13-15`):

```diff
-NORMALIZED_KSELFTEST_PREBUILT_DIR := $(patsubst %/,%,$(abspath $(KSELFTEST_PREBUILT_DIR)))
-NORMALIZED_TARGET_DIR := $(patsubst %/,%,$(abspath $(TARGET_DIR)))
-$(if $(filter $(NORMALIZED_TARGET_DIR),$(NORMALIZED_KSELFTEST_PREBUILT_DIR)),$(error KSELFTEST_PREBUILT_DIR must not equal TARGET_DIR: $(NORMALIZED_TARGET_DIR)))
```

test/initramfs/src/conformance/kselftest/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+KSELFTEST_PREBUILT_DIR ?= /opt/kselftest
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+CUR_DIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+INITRAMFS ?= $(CUR_DIR)/../../../build/initramfs
+TARGET_DIR := $(INITRAMFS)/opt/kselftest
+RUN_BASH := $(CUR_DIR)/run_kselftest.sh
+BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists/*)
+EXT2_BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists.ext2/*)
+EXFAT_BLOCK_LIST := $(wildcard $(CUR_DIR)/blocklists.exfat/*)
```
tatetian: **The `blocklists.*/` directory layer is vestigial — each dir holds exactly one file** (minor): After consolidation `blocklists/` contains only `blocklists`, `blocklists.ext2/` contains only `blocklists`, and `blocklists.exfat/` contains only `blocklists`. The doubled path (`blocklists/blocklists`) is a smell, and forcing a `$(wildcard ...)` around a single known file just so the Makefile can support "multiple files later" is YAGNI — and it also leaks: `$(wildcard)` happily picks up `.swp`/`~`/`.bak` editor droppings, and the symmetry with gVisor that motivated the dir shape doesn't apply here (gVisor's `EXTRA_BLOCKLISTS_DIRS` genuinely iterates one-file-per-test-binary at `run_gvisor_test.sh:25-29`; kselftest has a single consolidated file).

Flatten to plain files. This cascades through the Makefile, the runner, and the `extra_blocklists` matrix inputs.

Fix — rename on disk:

```sh
git mv test/initramfs/src/conformance/kselftest/blocklists/blocklists       test/initramfs/src/conformance/kselftest/blocklists
git mv test/initramfs/src/conformance/kselftest/blocklists.ext2/blocklists  test/initramfs/src/conformance/kselftest/blocklists.ext2
git mv test/initramfs/src/conformance/kselftest/blocklists.exfat/blocklists test/initramfs/src/conformance/kselftest/blocklists.exfat
rmdir  test/initramfs/src/conformance/kselftest/blocklists/ \
       test/initramfs/src/conformance/kselftest/blocklists.ext2/ \
       test/initramfs/src/conformance/kselftest/blocklists.exfat/
```

Then in `kselftest/Makefile`, each variable is a concrete file path with no wildcard:

```make
BLOCK_LIST       := $(CUR_DIR)/blocklists
EXT2_BLOCK_LIST  := $(CUR_DIR)/blocklists.ext2
EXFAT_BLOCK_LIST := $(CUR_DIR)/blocklists.exfat
$(if $(wildcard $(BLOCK_LIST)),,$(error missing default blocklist: $(BLOCK_LIST)))

$(TARGET_DIR): $(RUN_BASH) $(BLOCK_LIST) $(EXT2_BLOCK_LIST) $(EXFAT_BLOCK_LIST)
	@rm -rf $@ && mkdir -p $@
	@cp -rf $(KSELFTEST_PREBUILT_DIR)/. $@/
	@cp -f $(BLOCK_LIST)       $@/blocklists
	@cp -f $(EXT2_BLOCK_LIST)  $@/blocklists.ext2
	@cp -f $(EXFAT_BLOCK_LIST) $@/blocklists.exfat
	@cp -f $(RUN_BASH) $@
```

Then in `run_kselftest.sh`, replace `DEFAULT_BLOCKLISTS="$KSELFTEST_DIR/blocklists/blocklists"` with `DEFAULT_BLOCKLISTS="$KSELFTEST_DIR/blocklists"`, and rename the `EXTRA_BLOCKLISTS_DIRS` loop to take filenames:

```sh
DEFAULT_BLOCKLISTS="$KSELFTEST_DIR/blocklists"
...
for extra_file in $EXTRA_BLOCKLISTS ; do
    extra_path="$KSELFTEST_DIR/$extra_file"
    if [ -r "$extra_path" ]; then
        printf '\n' >> "$COMBINED_BLOCKLISTS"
        cat "$extra_path" >> "$COMBINED_BLOCKLISTS"
    else
        echo "Warning: extra blocklist not found: $extra_path" >&2
    fi
done
```

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,200 @@
...
+            dir="${line%%:*}"
+            command="${line#*:}"
+            if [ "$command" = "*" ]; then
+                awk -F: -v d="$dir" '$1 == d { print }' "$TESTS" >> "$RESOLVED_BLOCKLISTS"
+            else
+                printf '%s\n' "$line" >> "$RESOLVED_BLOCKLISTS"
+            fi
+            ;;
+        *)
+            echo "Warning: Invalid format in blocklist: $line" >&2
```
tatetian: **"Invalid format" is a warning only — typos silently disable entries** (significant): Now that the 90 files are consolidated into a single source of truth, a typo like `exec execveat` (space instead of colon) skips the intended entry for the lifetime of the PR and nothing in CI notices — the runner continues, the malformed line never matches any test, and the blocklist becomes smaller than intended. This matches the spirit of the `baseKselftestTargets` "fail-hard" comment but the code doesn't.

Fix:

```sh
*)
    echo "Error: Invalid format in blocklist: $line" >&2
    exit 1
    ;;
```

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -0,0 +1,200 @@
...
+
+# The `kselftest-list.txt` is a generated file produced by the kselftest build.
+#
+# Each line describes a single kselftest case in the form:
+#     <test-dir>:<test-command>
+#
+# Example lines (see blocklists):
+#   acct:acct_syscall
+#   cgroup:test_stress.sh
+#   exec:binfmt_script.py
```
tatetian: **Examples reference subsystems that are not built** (minor): `baseKselftestTargets` in `kselftest.nix` enables `exec lsm proc signal splice timers vDSO` (+ `x86` on x86_64), so `acct` and `cgroup` entries never appear in `kselftest-list.txt` at runtime. The "(see blocklists)" aside is also misleading for the `acct` / `cgroup` examples — those subsystems aren't in `blocklists/blocklists` either. Looks copy-pasted from upstream kernel docs.

Fix — replace with examples that actually match the enabled targets:

```sh
# Example lines (matching enabled TARGETS in kselftest.nix):
#   exec:binfmt_script.py
#   timers:posix_timers
#   vDSO:vdso_test_abi
```

test/initramfs/src/conformance/kselftest/blocklists/blocklists
```diff
@@ -0,0 +1,109 @@
+# kselftest blocklist. Format: <dir>:<test-name>.
+# Use '<dir>:*' to skip every test under <dir>.
```
tatetian: **Under-documentation of what is accepted in the file** (minor): Consider the following version:

```text
# kselftest blocklist. One entry per line:
#   <dir>:<test-name>   skip a single test case
#   <dir>:*             skip every test case under <dir>
# Blank lines and `#...` comments are ignored; leading/trailing whitespace is trimmed.
```

.github/workflows/test_x86.yml
```diff
@@ -127,6 +127,20 @@ jobs:
...
+          auto_test: 'conformance'
+          release: ${{ !contains(matrix.release, 'false') }}
+          enable_kvm: ${{ !contains(matrix.enable_kvm, 'false') }}
+          smp: ${{ matrix.smp }}
+          netdev: ${{ matrix.netdev || 'tap' }}
+          scheme: ${{ matrix.scheme }}
+          extra_blocklists: ${{ matrix.extra_blocklists }}
+          conformance_test_suite: 'kselftest'
+          conformance_test_workdir: ${{ matrix.conformance_test_workdir }}
+          boot_protocol: ${{ matrix.boot_protocol || 'linux-efi-handover64' }}
```
tatetian: **Three near-identical 14-line step blocks — duplication remains** (minor): The original `syscall_test_suite` typo is gone, but the underlying copy-paste that enabled it was not refactored: three ~14-line `uses: ./.github/actions/test` blocks differ only in `conformance_test_suite` and the name. Each future suite needs another copy, and any change to the call (e.g., a new composite-action input) must be applied three times in lock-step. Structurally, today's shape also forces gVisor and kselftest to wait for LTP to finish on the same runner.

Fix — replace `integration-test` with three sibling jobs, one per test kind: `boot-test`, `regression-test`, `conformance-test`. Each matrix then holds only variations of that kind, `auto_test` becomes a literal instead of a `startsWith`-based ternary, and the conformance job uses a 2D matrix (`suite × test_id`) so all 5 × 3 cells run in parallel.

```yaml
  boot-test:
    runs-on: ubuntu-latest
    container:
      image: asterinas/asterinas:0.17.1-20260319
      options: -v /dev:/dev --privileged
      volumes:
        - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
        - /usr/local/lib/android:/usr/local/lib/android
        - /usr/share/dotnet:/usr/share/dotnet
    strategy:
      fail-fast: false
      matrix:
        include:
          - { variant: 'legacy32',        boot_protocol: 'linux-legacy32', smp: 4 }
          - { variant: 'multiboot2-smp4', boot_protocol: 'multiboot',      smp: 4 }
    steps:
      - name: Free disk space
        run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
      - uses: actions/checkout@v4
      - name: Run boot test (${{ matrix.variant }})
        uses: ./.github/actions/test
        with:
          auto_test: 'boot'
          smp: ${{ matrix.smp }}
          boot_protocol: ${{ matrix.boot_protocol }}

  regression-test:
    runs-on: ubuntu-latest
    container:
      image: asterinas/asterinas:0.17.1-20260319
      options: -v /dev:/dev --privileged
      volumes:
        - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
        - /usr/local/lib/android:/usr/local/lib/android
        - /usr/share/dotnet:/usr/share/dotnet
    strategy:
      fail-fast: false
      matrix:
        include:
          - { variant: 'handover64-debug', release: false, boot_protocol: 'linux-efi-handover64' }
          - { variant: 'multiboot2-smp4',  boot_protocol: 'multiboot2', smp: 4 }
    steps:
      - name: Free disk space
        run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
      - uses: actions/checkout@v4
      - name: Run regression test (${{ matrix.variant }})
        uses: ./.github/actions/test
        with:
          auto_test: 'regression'
          release: ${{ !contains(matrix.release, 'false') }}
          smp: ${{ matrix.smp }}
          boot_protocol: ${{ matrix.boot_protocol || 'linux-efi-handover64' }}

  conformance-test:
    runs-on: ubuntu-latest
    container:
      image: asterinas/asterinas:0.17.1-20260319
      options: -v /dev:/dev --privileged
      volumes:
        - /usr/lib/google-cloud-sdk:/usr/lib/google-cloud-sdk
        - /usr/local/lib/android:/usr/local/lib/android
        - /usr/share/dotnet:/usr/share/dotnet
    strategy:
      fail-fast: false
      matrix:
        suite: ['ltp', 'gvisor', 'kselftest']
        include:
          - { variant: 'debug',                  release: false, boot_protocol: 'linux-efi-pe64' }
          - { variant: 'ext2-microvm',           scheme: 'microvm', extra_blocklists: 'blocklists.ext2',  conformance_test_workdir: '/ext2' }
          - { variant: 'ext2-iommu-debug',       scheme: 'iommu',   extra_blocklists: 'blocklists.ext2',  conformance_test_workdir: '/ext2', release: false }
          - { variant: 'exfat-multiboot2-nokvm', boot_protocol: 'multiboot2', enable_kvm: false, extra_blocklists: 'blocklists.exfat', conformance_test_workdir: '/exfat' }
          - { variant: 'multiboot2-smp4',        boot_protocol: 'multiboot2', smp: 4 }
    steps:
      - name: Free disk space
        run: rm -rf /usr/lib/google-cloud-sdk/* /usr/local/lib/android/* /usr/share/dotnet/*
      - uses: actions/checkout@v4
      - name: Run conformance test (${{ matrix.variant }}, ${{ matrix.suite }})
        uses: ./.github/actions/test
        with:
          auto_test: 'conformance'
          release: ${{ !contains(matrix.release, 'false') }}
          enable_kvm: ${{ !contains(matrix.enable_kvm, 'false') }}
          smp: ${{ matrix.smp }}
          netdev: ${{ matrix.netdev || 'tap' }}
          scheme: ${{ matrix.scheme }}
          extra_blocklists: ${{ matrix.extra_blocklists }}
          conformance_test_suite: ${{ matrix.suite }}
          conformance_test_workdir: ${{ matrix.conformance_test_workdir }}
          boot_protocol: ${{ matrix.boot_protocol || 'linux-efi-handover64' }}
```

This expands to 2 + 2 + 5 × 3 = 19 parallel jobs (vs today's 8 jobs each running 3 sequential steps), cutting wall-clock meaningfully. The `container:` block duplicates across the three jobs — ~24 lines of YAML — which is the acceptable tradeoff for one-kind-per-job, literal `auto_test`, and matrix shape matching semantics. (GitHub Actions honors YAML anchors if the duplication bothers you.) Apply the same split to `test_x86_tdx.yml`.

test/initramfs/nix/conformance/kselftest.nix
tatetian, line 1: **`test/initramfs/README.md` still lists only gvisor and ltp** (significant): Filing this as a file-level comment on `kselftest.nix` because `README.md` itself isn't touched by this PR. Previous review flagged it as minor; with kselftest now live in CI the gap matters more — contributors landing on `test/initramfs/README.md` see no entry point for kselftest (tree diagram missing a `kselftest/` node, no dedicated subsection like the existing "Conformance Test Suite - gVisor Exception" at line 32), so the pinned Linux version (currently v6.18), the invocation (`make run AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest`), and the version-bump procedure documented inside this `kselftest.nix` are only discoverable by reading this file.

Fix — add a `kselftest/` node to the tree and a `### Linux Kernel Selftest (kselftest)` subsection in `test/initramfs/README.md`:

````markdown
│   ├── conformance/   # Third-party test suites to verify Linux compatibility
│   │   ├── gvisor/    # Gvisor syscall test suite
│   │   ├── kselftest/ # Linux kernel in-tree selftests
│   │   └── ltp/       # Linux Test Project syscall test suite
````

````markdown
### Linux Kernel Selftest (kselftest)

The `kselftest` suite builds a subset of Linux's in-tree selftests
(`tools/testing/selftests`) from the Linux version pinned in
`nix/conformance/kselftest.nix` (currently **v6.18**). Only the
subsystems listed in `baseKselftestTargets` are compiled, plus `x86`
on x86_64 hosts.

Invoke via:

```bash
make run AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest
```

**Bumping the Linux version.** Edit `version` in
`nix/conformance/kselftest.nix`, temporarily set `hash = lib.fakeHash;`,
run the Nix build once, and copy the real hash reported by `fetchgit`
back into `hash`.

**Blocklist layout.** `src/conformance/kselftest/blocklists` is the
default blocklist; `blocklists.ext2` and `blocklists.exfat` hold
filesystem-specific extras. See the header of `blocklists` for the
entry format.
````

tatetian (commented):
Two prior comments remain open: the three near-identical step blocks in `test_x86.yml` were not collapsed, and `README.md` still lists only gvisor/ltp. With kselftest now live in CI the README gap is worth treating as significant.

Three issues found this round rise above nits:

1. **Eager kselftest build for every `testSuite`.** `nix/conformance/default.nix` realizes `kselftest` unconditionally via `export KSELFTEST_PREBUILT_DIR=${kselftest}`, so every `ltp`/`gvisor` CI invocation also builds the selftest derivation. LTP/gVisor had the same smell; kselftest amplifies it materially.
2. **Toothless "Invalid format" warning in the consolidated blocklist.** The runner prints to stderr and continues on a malformed line, so a typo like `exec execveat` (space instead of colon) silently disables the entry; nothing in CI notices. Since the file is the single source of truth, the right response is `exit 1`.
3. **Vestigial `blocklists.*/` directory layer.** Each `blocklists/`, `blocklists.ext2/`, `blocklists.exfat/` directory now holds exactly one file (`blocklists`), so the doubled path `blocklists/blocklists` and the surrounding `$(wildcard)` machinery are leftover from the 90-file layout. Flatten to plain files at the suite root.

tatetian (approved):

PR #2901, opened 2026-01-19
Allow empty UNIX control messages

lrh2000:
So we can fix:
https://github.com/asterinas/asterinas/blob/bdc1c797701a677f3f77e2d1ed73a7f1133a9b4a/test/initramfs/src/syscall/gvisor/blocklists/socket_unix_pair_test#L29-L31

---

The following tests are also relatively easy to fix and should be fixed. I will do them later in separate PRs.
https://github.com/asterinas/asterinas/blob/bdc1c797701a677f3f77e2d1ed73a7f1133a9b4a/test/initramfs/src/syscall/gvisor/blocklists/socket_unix_pair_test#L27
https://github.com/asterinas/asterinas/blob/bdc1c797701a677f3f77e2d1ed73a7f1133a9b4a/test/initramfs/src/syscall/gvisor/blocklists/socket_unix_pair_test#L33-L36

### DIFF
--- a/kernel/src/net/socket/util/message_header.rs
+++ b/kernel/src/net/socket/util/message_header.rs
@@ -49,7 +49,7 @@ impl ControlMessage {
 
         while reader.has_remain() && msgs.len() < MAX_NR_MSGS {
             let header = reader.read_val::<CControlHeader>()?;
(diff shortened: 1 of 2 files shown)


StevenJiang1110 (approved):

PR #2903, opened 2026-01-20
Fix `SO_PEERCRED` behavior regarding datagrams and inheritance

lrh2000:
From [man pages](https://man7.org/linux/man-pages/man7/unix.7.html):
> SO_PEERCRED
>               This read-only socket option returns the credentials of the
>               peer process connected to this socket.  The returned
>               credentials are those that were in effect at the time of
>               the call to [connect(2)](https://man7.org/linux/man-pages/man2/connect.2.html), [listen(2)](https://man7.org/linux/man-pages/man2/listen.2.html), or [socketpair(2)](https://man7.org/linux/man-pages/man2/socketpair.2.html).

It was thought that datagram sockets should not have peer credentials because they do not have a well-defined peer (since they are not connection-oriented). However, I was wrong because I missed the case about `socketpair`.

It seems that `socketpair` should record peer credentials even if the socket is not connection-oriented. The first commit resolves this issue and some gvisor tests now pass.

---

In addition, according to Linux implementation, we should inherit some socket flags (i.e., [`SO_PASSCRED`, `SO_PASSSEC`, `SO_PASSPIDFD`, `SO_PASSRIGHTS`](https://elixir.bootlin.com/linux/v6.18.6/source/include/net/sock.h#L543-L550)) for the new socket to be accepted later when [connecting UNIX stream sockets](https://elixir.bootlin.com/linux/v6.18.6/source/net/unix/af_unix.c#L1765). We already have a TODO for this:
https://github.com/asterinas/asterinas/blob/77f0b0e746ce2df0066b007fb050fc5be0e86e47/kernel/src/net/socket/unix/stream/listener.rs#L61-L62

The second commit fixes this to pass another gvisor test.

### DIFF
--- a/kernel/src/net/socket/unix/stream/connected.rs
+++ b/kernel/src/net/socket/unix/stream/connected.rs
@@ -14,7 +14,7 @@ use crate::{
         unix::{
             UnixSocketAddr, addr::UnixSocketAddrBound, cred::SocketCred, ctrl_msg::AuxiliaryData,
         },
-        util::{ControlMessage, SockShutdownCmd, options::SocketOptionSet},
+        util::{ControlMessage, SockShutdownCmd},
     },
     prelude::*,
     process::signal::Pollee,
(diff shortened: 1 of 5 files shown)


StevenJiang1110 (approved):

PR #2908, opened 2026-01-20
Fix ext2 fast symlink inode detection logic

### LINKED ISSUE #2883: The symlink to Go is broken on NixOS
When attempting to install Go on Asterinas NixOS, I discovered that the symlink to the Go binary is consistently broken. As a result, Go cannot be executed on Asterinas NixOS.

### Steps to Reproduce

1. Add Go as a system package in `distro/etc_nixos/configuration.nix`:

```nix
environment.systemPackages = with pkgs; [ hello-asterinas go ];
```

2. Build and run:

```bash
make nixos
make run_nixos
```

3. Check the Go symlink under `/run/current-system/sw/bin`. You'll find it is broken:

```bash
[root@asterinas:~]# ls -l /run/current-system/sw/bin/go
lrwxrwxrwx 1 root root 60 Jan 14 03:07 /run/current-system/sw/bin/go -> 'E'$'\017\003'
```

This appears to be an Asterinas-specific bug. Notably, when using nixos-enter to access the same NixOS environment from the host system, the symlink is correct and fully functional.

cqs21:
Fixes #2883

### DIFF
--- a/kernel/src/fs/ext2/inode.rs
+++ b/kernel/src/fs/ext2/inode.rs
@@ -66,9 +66,12 @@ impl Inode {
                 None
             },
             block_group_idx,
-            xattr: desc
-                .acl
-                .map(|acl| Xattr::new(acl, weak_self.clone(), fs.clone())),
+            xattr: match desc.type_ {
+                InodeType::Dir | InodeType::File => {
+                    Some(Xattr::new(desc.acl, weak_self.clone(), fs.clone()))
+                }
+                _ => None,
+            },
             inner: RwMutex::new(InodeInner::new(desc, weak_self.clone(), fs.clone())),
             fs,
             extension: Extension::new(),
@@ -885,7 +888,7 @@ impl Inode {
     pub fn file_flags(&self) -> FileFlags;
     pub fn hard_links(&self) -> u16;
     pub fn blocks_count(&self) -> Ext2Bid;
-    pub fn acl(&self) -> Option<Bid>;
+    pub fn acl(&self) -> Bid;
     pub fn atime(&self) -> Duration;
     pub fn mtime(&self) -> Duration;
     pub fn ctime(&self) -> Duration;
@@ -1189,7 +1192,7 @@ impl InodeInner {
     pub fn inc_hard_links(&mut self);
     pub fn dec_hard_links(&mut self);
     pub fn blocks_count(&self) -> Ext2Bid;
-    pub fn acl(&self) -> Option<Bid>;
+    pub fn acl(&self) -> Bid;
     pub fn set_acl(&mut self, bid: Bid);
     pub fn atime(&self) -> Duration;
     pub fn set_atime(&mut self, time: Duration);
@@ -1288,7 +1291,7 @@ impl InodeImpl {
         self.desc.blocks_count()
     }
 
-    pub fn acl(&self) -> Option<Bid> {
+    pub fn acl(&self) -> Bid {
         self.desc.acl
     }
 
@@ -2136,7 +2139,7 @@ pub(super) struct InodeDesc {
     /// Pointers to blocks.
     block_ptrs: BlockPtrs,
     /// File or directory acl block.
-    acl: Option<Bid>,
+    acl: Bid,
 }
 
 impl TryFrom<RawInode> for InodeDesc {
@@ -2164,9 +2167,9 @@ impl TryFrom<RawInode> for InodeDesc {
                 .ok_or(Error::with_message(Errno::EINVAL, "invalid file flags"))?,
             block_ptrs: inode.block_ptrs,
             acl: match inode_type {
-                InodeType::File => Some(Bid::new(inode.file_acl as _)),
-                InodeType::Dir => Some(Bid::new(inode.size_high as _)),
-                _ => None,
+                InodeType::File => Bid::new(inode.file_acl as _),
+                InodeType::Dir => Bid::new(inode.size_high as _),
+                _ => Bid::new(0),
             },
         })
     }
@@ -2190,10 +2193,7 @@ impl InodeDesc {
             sector_count: 0,
             flags: FileFlags::empty(),
             block_ptrs: BlockPtrs::default(),
-            acl: match type_ {
-                InodeType::File | InodeType::Dir => Some(Bid::new(0)),
-                _ => None,
-            },
+            acl: Bid::new(0),
         })
     }
 
@@ -2221,11 +2221,7 @@ impl InodeDesc {
 
     /// Returns the number of sectors used for ACL.
     fn acl_sectors(&self) -> u32 {
-        let Some(bid) = self.acl else {
-            return 0;
-        };
-
-        if bid.to_raw() == 0 {
+        if self.acl.to_raw() == 0 {
             return 0;
         }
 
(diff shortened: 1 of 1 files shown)


kernel/src/fs/ext2/inode.rs
```diff
@@ -2210,11 +2210,23 @@ impl InodeDesc {
...
+
+    /// Returns whether the inode is a fast symlink.
+    ///
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/ext2/inode.c#L48-L55>.
+    fn is_fast_symlink(&self) -> bool {
+        if self.type_ != InodeType::SymLink {
+            return false;
+        }
+
+        let extra_sectors = self.acl.map_or(0, |_| BLOCK_SIZE / SECTOR_SIZE) as u32;
```
tatetian: Just a few lines below is a `sectors_to_blocks` utility function for exactly this purpose...

kernel/src/fs/ext2/inode.rs
```diff
@@ -2210,11 +2210,23 @@ impl InodeDesc {
...
+        self.sector_count == extra_sectors
```
tatetian: On what the basis should the ACL field be used to check whether a symbolic link is fast or not...

tatetian (commented):
Good job on identifying the bug so quickly.

While I was reading the code, the PR turned to draft. But I submitted the reviews anyway.

The logic of this bugfix does not feel right to me. And you might want to take a look at the `write_link` method, which is also problematic:

https://github.com/asterinas/asterinas/blob/cab259e0287e6e83ffe1f234ac362f7451814175/kernel/src/fs/ext2/inode.rs#L1343-L1352

Calling `InodeImpl::resize` does not make any sense. As I understand it, a symbolic link whose link string is stored inline should have a block count of 0.

---

Edit: when writing a fast symbolic link, a correct implementation needs to take into account that the inode may currently store a normal symbolic link. In this case, the block count needs to be updated to zero. Otherwise, the `write_link` method should not touch the block count (which must already be zero).

kernel/src/fs/ext2/inode.rs
cqs21: My understanding is that `self.sector_count` stores the total number of disk sectors used by this inode, including both the file content itself and the extended ACL file (which occupies one block). Therefore, if we subtract the ACL file sectors and sector_count is still non-zero, it indicates that this symlink's content is stored in a separate data block.

tatetian: > My understanding is that `self.sector_count` stores the total number of disk sectors used by this inode

Then, what is the basis of this understanding? 

* From the perspective of our implementation: the [`try_expand_blocks`](https://github.com/asterinas/asterinas/blob/cab259e0287e6e83ffe1f234ac362f7451814175/kernel/src/fs/ext2/inode.rs#L1436) method sets `sector_count` without taking into account of `acl`.
* From the perspective of [the Ext2 spec](https://www.nongnu.org/ext2-doc/ext2.html): 
  > i_blocks
  > 32-bit value representing the total number of 512-bytes blocks reserved to contain the data of this inode. The block numbers of these reserved blocks are contained in the i_block array.
  >
  > h_blocks
  > In Linux a value of h_blocks higher than 1 is considered invalid. This effectively restrict the amount of extended attributes to what can be fit in a single block.

  `i_blocks` only accounts for the blocks for _data_, not ACL. And the length of ACL is restricted to one block.

cqs21: > Then, what is the basis of this understanding?

See reference: https://elixir.bootlin.com/linux/v6.18/source/fs/ext2/inode.c#L48-L55

> i_blocks
32-bit value representing the total number of 512-bytes blocks reserved to contain the data of this inode.

The `i_blocks` field represents "the data of this inode" rather than "the data of this file". Based on Linux kernel implementation, ACL is counted as part of `i_blocks`.

tatetian: Then the up-to-date Linux implementation is inconsistent with the written spec. We have to respect the "executing spec" (the code itself) over the written one.

However, the changes of this PR can't be right because the semantics of `i_blocks` is inconsistent across our Ext2 implementation. In the other part of the code (`try_expand_blocks`), `i_blocks` is interpreted as a counter that only accounts the data blocks, not ACL ones.

kernel/src/fs/ext2/inode.rs
```diff
@@ -1519,7 +1518,8 @@ impl InodeImpl {
             return Err(e);
         }
 
-        self.desc.sector_count = blocks_to_sectors(range.start + device_range.len() as u32);
+        self.desc.sector_count = blocks_to_sectors(range.start + device_range.len() as u32)
+            + self.desc.file_acl_sectors();
```
tatetian: This pattern has appeared triple. Wrap this pattern into a reusable method.

kernel/src/fs/ext2/inode.rs
```diff
@@ -2215,6 +2213,30 @@ impl InodeDesc {
...
+
+    /// Returns whether the inode is a fast symlink.
+    ///
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/ext2/inode.c#L48-L55>.
+    fn is_fast_symlink(&self) -> bool {
+        if self.type_ != InodeType::SymLink {
+            return false;
+        }
+
+        self.sector_count == self.file_acl_sectors()
```
tatetian: I prefer a more revealing way to write the same logic:

```rust
let nr_data_sectors = self.sector_count - self.file_acl_sectors();
nr_data_sectors == 0
```

kernel/src/fs/ext2/inode.rs
```diff
@@ -2215,6 +2218,44 @@ impl InodeDesc {
...
         size.div_ceil(BLOCK_SIZE) as Ext2Bid
     }
+
+    /// Returns the number of sectors used for ACL.
+    fn acl_sectors(&self) -> u32 {
+        let Some(bid) = self.acl else {
+            return 0;
+        };
+
+        if bid.to_raw() == 0 {
```
tatetian: If `bid` is allowed to contain zero, then using `Option<_>` for the `acl` field becomes pointless. So I suggest changing its type to `Bid`. And this method will also be simplified.

tatetian (commented):
Great. The code is correct... finally.

kernel/src/fs/ext2/inode.rs
```diff
@@ -66,9 +66,12 @@ impl Inode {
                 None
             },
             block_group_idx,
-            xattr: desc
-                .acl
-                .map(|acl| Xattr::new(acl, weak_self.clone(), fs.clone())),
+            xattr: match desc.type_ {
+                InodeType::Dir | InodeType::File => {
+                    Some(Xattr::new(desc.acl, weak_self.clone(), fs.clone()))
```
tatetian, line 71: This is NOT equivalent to the original code, which skips the creation of `Xattr` if `desc.acl` is none (or 0 in the new code).

cqs21: From the old code, `acl` would not be `None` for `File` and `Dir`:
https://github.com/asterinas/asterinas/blob/306cc714ae9fe1dada55499d7ecab0853b068a5a/kernel/src/fs/ext2/inode.rs#L2165-L2169

From the new code, acl can indeed be `0`, but this is not a problem. The lazy initialization in `xattr::lazy_init` will check the `Bid` value; if an `Xattr` is not created there, it can still be created later when functions like `set_xattr` are called, based on the inode type:
https://github.com/asterinas/asterinas/blob/306cc714ae9fe1dada55499d7ecab0853b068a5a/kernel/src/fs/ext2/xattr.rs#L110-L116

tatetian: Yes, you are right. Thanks for the explanation.

tatetian (approved):

PR #2909, opened 2026-01-20
Handle the listener in `State::set_pass_cred`

lrh2000:
I am sorry that https://github.com/asterinas/asterinas/pull/2903 forgets to update the code in `State::set_pass_cred`.... 

Hopefully, this won't cause too much noise. It's a one-line fix PR with an 83-line regression test.


StevenJiang1110 (approved):

PR #2912, opened 2026-01-21
Add `pidfd_send_signal` syscall

li041:
## Changes in this PR
- Generalize signal-sending support with `Signal` trait.
- Add `RawSignal` struct to represent `siginfo_t` which user provide.
- Add `pidfd_send_signal` syscall.
## Why `FIXME` in the code
- It seems that `/proc/<pid>` represents specific `task` in linux kernel, which is the same semantic as `thread` in asterinas. But when we open `<proc/<pid>` in asterinas, it is always a process.
- Besides, `pidfd_open` can create a file desciptor that refers to a specific thread, while we doen't support it yet.

Closes https://github.com/asterinas/asterinas/issues/2866

### DIFF
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/signals-and-timers/fully_covered.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/signals-and-timers/fully_covered.scml
@@ -16,6 +16,9 @@ kill(pid, sig);
 // Send signal to a thread
 tgkill(tgid, tid, sig);
 
+// Send signal to the target process referred to by pidfd
+pidfd_send_signal(pidfd, sig, info, flags = PIDFD_SIGNAL_THREAD | PIDFD_SIGNAL_THREAD_GROUP | PIDFD_SIGNAL_PROCESS_GROUP);
+
 // Wait for signal
 pause();
 
--- a/kernel/src/process/signal/signals/raw.rs
+++ b/kernel/src/process/signal/signals/raw.rs
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use core::fmt::Debug;
+
+use crate::process::signal::{c_types::siginfo_t, sig_num::SigNum, signals::Signal};
+
+/// A signal that carries raw [`siginfo_t`] information.
+#[derive(Clone, Copy)]
+pub struct RawSignal {
+    info: siginfo_t,
+}
+
+impl Debug for RawSignal {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("RawSignal")
+            .field("signo", &self.info.si_signo)
+            .field("errno", &self.info.si_errno)
+            .field("code", &self.info.si_code)
+            .finish_non_exhaustive()
+    }
+}
+
+impl RawSignal {
+    /// Creates a signal that carries raw [`siginfo_t`] information.
+    ///
+    /// The caller must ensure that the `info.si_signo` is a valid signal number.
+    pub fn new(info: siginfo_t) -> Self {
+        Self { info }
+    }
+}
+
+impl Signal for RawSignal {
+    fn num(&self) -> SigNum {
+        SigNum::from_u8(self.info.si_signo as u8)
+    }
+
+    fn to_info(&self) -> siginfo_t {
+        self.info
+    }
+}
--- a/kernel/src/syscall/pidfd_send_signal.rs
+++ b/kernel/src/syscall/pidfd_send_signal.rs
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use ostd::mm::VmIo;
+
+use crate::{
+    fs::file_table::{FileDesc, get_file_fast},
+    prelude::*,
+    process::{
+        Pgid, Pid, PidFile, kill, kill_group,
+        posix_thread::AsPosixThread,
+        signal::{
+            c_types::siginfo_t,
+            constants::SI_TKILL,
+            sig_num::SigNum,
+            signals::{Signal, raw::RawSignal, user::UserSignal},
+        },
+        tgkill,
+    },
+    syscall::SyscallReturn,
+    thread::Tid,
+};
+
+pub fn sys_pidfd_send_signal(
+    pidfd: FileDesc,
+    sig_num: u64,
+    info_ptr: Vaddr,
+    flags: u32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    let flags = PidfdSendSignalFlags::try_from(flags)?;
+    let sig_num = SigNum::try_from(sig_num as u8)?;
+    debug!(
+        "pidfd={}, info_ptr={:#x}, flags={:?}",
+        pidfd, info_ptr, flags
+    );
+
+    let siginfo = read_siginfo_from_user(info_ptr, sig_num, ctx)?;
+    let signal = RawSignal::new(siginfo);
+
+    let target = get_target_from_pidfd(pidfd, flags, ctx)?;
+
+    let is_self = match &target {
+        SignalTarget::Thread { tid, tgid: _ } => *tid == ctx.posix_thread.tid(),
+        SignalTarget::Process { pid } => *pid == ctx.posix_thread.tid(),
+        SignalTarget::ProcessGroup { pgid: _ } => false,
+    };
+
+    if !is_self && (siginfo.si_code >= 0 || siginfo.si_code == SI_TKILL) {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "signals with custom code can only be sent to the current thread/process"
+        );
+    }
+
+    match target {
+        SignalTarget::Thread { tid, tgid } => {
+            tgkill(tid, tgid, Some(Box::new(signal) as Box<dyn Signal>), ctx)?;
+        }
+        SignalTarget::Process { pid } => {
+            kill(pid, Some(Box::new(signal) as Box<dyn Signal>), ctx)?;
+        }
+        SignalTarget::ProcessGroup { pgid } => {
+            kill_group(pgid, Some(signal), ctx)?;
+        }
+    }
+
+    Ok(SyscallReturn::Return(0))
+}
+
+fn read_siginfo_from_user(info_ptr: Vaddr, sig_num: SigNum, ctx: &Context) -> Result<siginfo_t> {
+    if info_ptr != 0 {
+        let si = ctx.user_space().read_val::<siginfo_t>(info_ptr)?;
+        if si.si_signo != sig_num.as_u8() as i32 {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "`siginfo.si_signo` does not match the specified signal number"
+            );
+        }
+        Ok(si)
+    } else {
+        // If `info_ptr` is NULL, the kernel constructs a default `siginfo_t` structure
+        // whose fields match the values that are implicitly supplied when a signal is sent using the kill(2).
+        Ok(UserSignal::new_kill(sig_num, ctx).to_info())
+    }
+}
+
+fn get_target_from_pidfd(
+    pidfd: FileDesc,
+    flags: PidfdSendSignalFlags,
+    ctx: &Context,
+) -> Result<SignalTarget> {
+    // Helper closures to create signal targets.
+    let thread_target = |tid: Tid, tgid: Pid| SignalTarget::Thread { tid, tgid };
+    let process_target = |pid: Pid| SignalTarget::Process { pid };
+    let group_target = |pgid: Pgid| SignalTarget::ProcessGroup { pgid };
+
+    let target = match pidfd {
+        PIDFD_SELF_THREAD => match flags {
+            PidfdSendSignalFlags::Default | PidfdSendSignalFlags::Thread => {
+                thread_target(ctx.posix_thread.tid(), ctx.process.pid())
+            }
+            PidfdSendSignalFlags::ThreadGroup => process_target(ctx.process.pid()),
+            PidfdSendSignalFlags::ProcessGroup => group_target(ctx.posix_thread.tid()),
+        },
+        PIDFD_SELF_THREAD_GROUP => match flags {
+            PidfdSendSignalFlags::Default | PidfdSendSignalFlags::ThreadGroup => {
+                process_target(ctx.process.pid())
+            }
+            PidfdSendSignalFlags::Thread => thread_target(
+                ctx.process.main_thread().as_posix_thread().unwrap().tid(),
+                ctx.process.pid(),
+            ),
+            PidfdSendSignalFlags::ProcessGroup => group_target(ctx.process.pid()),
+        },
+        _ => {
+            let mut file_table = ctx.thread_local.borrow_file_table_mut();
+            let file = get_file_fast!(&mut file_table, pidfd);
+
+            // FIXME: On Linux, a pidfd can be also obtained by opening a `/proc/pid` directory.
+            // Reference: <https://man7.org/linux/man-pages/man2/pidfd_send_signal.2.html>
+            let Some(pid_file) = file.downcast_ref::<PidFile>() else {
+                return_errno_with_message!(Errno::EBADF, "the file is not a PID file");
+            };
+
+            let Some(process) = pid_file.process_opt() else {
+                return_errno_with_message!(Errno::ESRCH, "the target process has been reaped");
+            };
+
+            match flags {
+                PidfdSendSignalFlags::Default => {
+                    // FIXME: On Linux, a pidfd can refer to either a process or a thread.
+                    // We currently only support pidfds that refer to processes.
+                    process_target(process.pid())
+                }
+                PidfdSendSignalFlags::Thread => {
+                    // FIXME: On Linux, the signal can be sent to any thread.
+                    // We currently only support the main thread.
+                    thread_target(
+                        process.main_thread().as_posix_thread().unwrap().tid(),
+                        process.pid(),
+                    )
+                }
+                PidfdSendSignalFlags::ThreadGroup => process_target(process.pid()),
+                PidfdSendSignalFlags::ProcessGroup => group_target(process.pid()),
+            }
+        }
+    };
+
+    Ok(target)
+}
+
+enum SignalTarget {
+    Thread { tid: Tid, tgid: Pid },
+    Process { pid: Pid },
+    ProcessGroup { pgid: Pgid },
+}
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/pidfd.h#L19>.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, TryFromInt)]
+#[repr(u32)]
+enum PidfdSendSignalFlags {
+    Default = 0x0,
+    Thread = 0x1,
+    ThreadGroup = 0x2,
+    ProcessGroup = 0x4,
+}
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fcntl.h#L110>.
+const PIDFD_SELF_THREAD: i32 = -10000;
+const PIDFD_SELF_THREAD_GROUP: i32 = -10001;
--- a/test/initramfs/src/apps/scripts/process.sh
+++ b/test/initramfs/src/apps/scripts/process.sh
@@ -70,6 +70,7 @@ sched/sched_param_idle
 shm/posix_shm
 signal_c/kill
 signal_c/parent_death_signal
+signal_c/pidfd_send_signal
 signal_c/sigaltstack
 signal_c/signal_fd
 signal_c/signal_fpu
--- a/test/initramfs/src/apps/signal_c/pidfd_send_signal.c
+++ b/test/initramfs/src/apps/signal_c/pidfd_send_signal.c
@@ -1,0 +1,260 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+
+#include <pthread.h>
+#include <signal.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#include <sys/wait.h>
+#include <fcntl.h>
+
+#include "../test.h"
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fcntl.h#L110>.
+#define PIDFD_SELF_THREAD -10000
+#define PIDFD_SELF_THREAD_GROUP -10001
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/pidfd.h#L20>
+#define PIDFD_SIGNAL_THREAD_GROUP (1UL << 1)
+#define PIDFD_SIGNAL_PROCESS_GROUP (1UL << 2)
+
+const int sig = SIGUSR1;
+siginfo_t siginfo;
+
+static int pidfd_open(pid_t pid, unsigned int flags)
+{
+	return syscall(SYS_pidfd_open, pid, flags);
+}
+
+static int pidfd_send_signal(int pidfd, int sig, siginfo_t *info,
+			     unsigned int flags)
+{
+	return syscall(SYS_pidfd_send_signal, pidfd, sig, info, flags);
+}
+
+void setup_test_siginfo(siginfo_t *info, int sig, int si_code)
+{
+	memset(info, 0, sizeof(*info));
+	info->si_signo = sig;
+	info->si_code = si_code;
+	info->si_pid = getpid();
+	info->si_uid = getuid();
+}
+
+/* ==========================
+ *    Tests for processes
+ * ========================== */
+
+int process_pid;
+int process_pidfd;
+
+FN_SETUP(create_process)
+{
+	process_pid = CHECK(fork());
+	if (process_pid == 0) {
+		while (1) {
+			usleep(100);
+		}
+	}
+
+	process_pidfd = CHECK(pidfd_open(process_pid, 0));
+}
+END_SETUP()
+
+FN_TEST(pidfd_send_signal_errnos)
+{
+	setup_test_siginfo(&siginfo, sig, SI_USER);
+	TEST_ERRNO(pidfd_send_signal(process_pidfd, sig, &siginfo,
+				     PIDFD_SIGNAL_PROCESS_GROUP),
+		   EPERM);
+
+	setup_test_siginfo(&siginfo, sig, -666);
+	TEST_ERRNO(pidfd_send_signal(process_pidfd, sig + 1, &siginfo, 0),
+		   EINVAL);
+}
+END_TEST()
+
+FN_TEST(pidfd_send_signal_process)
+{
+	TEST_SUCC(pidfd_send_signal(process_pidfd, sig, &siginfo, 0));
+	TEST_SUCC(waitid(P_PID, process_pid, NULL, WNOWAIT | WEXITED));
+}
+END_TEST()
+
+FN_SETUP(cleanup_process)
+{
+	CHECK(close(process_pidfd));
+}
+END_SETUP()
+
+/* ==========================
+ *  Tests for `PIDFD_SELF_*`
+ * ========================== */
+
+// PIDFD_SELF_THREAD/PIDFD_SELF_THREAD_GROUP won't work with
+// PIDFD_SIGNAL_PROCESS_GROUP unless the current process is
+// the process group leader.
+FN_TEST(pidfd_send_signal_self_process_group)
+{
+	pid_t pid;
+	int stat;
+
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		setup_test_siginfo(&siginfo, SIGTERM, -666);
+
+		CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD, SIGTERM,
+					     &siginfo,
+					     PIDFD_SIGNAL_PROCESS_GROUP),
+			   _ret == -1 && errno == ESRCH);
+		CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD_GROUP, SIGTERM,
+					     &siginfo,
+					     PIDFD_SIGNAL_PROCESS_GROUP),
+			   _ret == -1 && errno == ESRCH);
+
+		exit(0);
+	}
+
+	TEST_RES(waitpid(pid, &stat, 0),
+		 WIFEXITED(stat) && WEXITSTATUS(stat) == 0);
+}
+END_TEST()
+
+void *pidfd_send_signal_self_child_thread(void *arg)
+{
+	setup_test_siginfo(&siginfo, SIGTERM, -666);
+
+	CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD, SIGTERM, &siginfo,
+				     PIDFD_SIGNAL_THREAD_GROUP),
+		   _ret == 0);
+
+	return NULL;
+}
+
+// PIDFD_SELF_THREAD will work with PIDFD_SIGNAL_THREAD_GROUP
+// even if the current thread is not the thread group leader.
+FN_TEST(pidfd_send_signal_self_thread_group)
+{
+	pid_t pid;
+	int stat;
+	pthread_t thread;
+
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		CHECK(pthread_create(&thread, NULL,
+				     &pidfd_send_signal_self_child_thread,
+				     NULL));
+		CHECK(pthread_join(thread, NULL));
+
+		exit(0);
+	}
+
+	TEST_RES(waitpid(pid, &stat, 0),
+		 WIFSIGNALED(stat) && WTERMSIG(stat) == SIGTERM);
+}
+END_TEST()
+
+void *pidfd_send_signal_self_process_group_child_thread(void *arg)
+{
+	setup_test_siginfo(&siginfo, SIGTERM, -666);
+
+	CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD, SIGTERM, &siginfo,
+				     PIDFD_SIGNAL_PROCESS_GROUP),
+		   _ret == -1 && errno == ESRCH);
+
+	return NULL;
+}
+
+// PIDFD_SELF_THREAD won't work with PIDFD_SIGNAL_PROCESS_GROUP
+// unless the current process is the process group leader and
+// the current thread is the main thread.
+FN_TEST(pidfd_send_signal_self_process_group_non_main_thread)
+{
+	pid_t pid;
+	int stat;
+	pthread_t thread;
+
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		CHECK(setpgid(0, 0));
+
+		CHECK(pthread_create(
+			&thread, NULL,
+			pidfd_send_signal_self_process_group_child_thread,
+			NULL));
+		CHECK(pthread_join(thread, NULL));
+
+		exit(0);
+	}
+
+	TEST_RES(waitpid(pid, &stat, 0),
+		 WIFEXITED(stat) && WEXITSTATUS(stat) == 0);
+}
+END_TEST()
+
+// FIXME: Enable thread tests once pidfd for threads is supported
+#ifndef __asterinas__
+/* ==========================
+ *     Tests for threads
+ * ========================== */
+
+volatile sig_atomic_t signal_received = 0;
+
+pthread_t thread;
+volatile pid_t thread_tid;
+
+int proc_fd;
+int proc_task_fd;
+
+void signal_handler(int signo)
+{
+	signal_received = 1;
+}
+
+void *thread_func(void *arg)
+{
+	thread_tid = syscall(SYS_gettid);
+	signal(sig, signal_handler);
+
+	while (!signal_received) {
+		usleep(100);
+	}
+
+	return NULL;
+}
+
+FN_SETUP(create_thread)
+{
+	static char path[256];
+
+	CHECK(pthread_create(&thread, NULL, thread_func, NULL));
+
+	while (thread_tid == 0) {
+		usleep(100);
+	}
+
+	snprintf(path, sizeof(path), "/proc/%d/", thread_tid);
+	proc_fd = CHECK(open(path, O_DIRECTORY | O_CLOEXEC));
+
+	snprintf(path, sizeof(path), "/proc/%d/task/%d", getpid(), thread_tid);
+	proc_task_fd = CHECK(open(path, O_DIRECTORY | O_CLOEXEC));
+}
+END_SETUP()
+
+FN_TEST(pidfd_send_signal_thread)
+{
+	TEST_ERRNO(pidfd_send_signal(proc_task_fd, sig, &siginfo, 0), EBADF);
+
+	TEST_SUCC(pidfd_send_signal(proc_fd, sig, &siginfo, 0));
+	TEST_SUCC(pthread_join(thread, NULL));
+}
+END_TEST()
+
+FN_SETUP(cleanup_thread)
+{
+	CHECK(close(proc_fd));
+	CHECK(close(proc_task_fd));
+}
+END_SETUP()
+#endif
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -343,6 +343,7 @@ which are summarized in the table below.
 | 327     | preadv2                | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#preadv2-and-pwritev2) |
 | 328     | pwritev2               | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#preadv2-and-pwritev2) |
 | 332     | statx                  | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#statx) |
+| 424     | pidfd_send_signal      | ✅             | 💯 |
 | 434     | pidfd_open             | ✅             | 💯 |
 | 435     | clone3                 | ✅             | [⚠️](syscall-flag-coverage/process-and-thread-management/#clone-and-clone3) |
 | 436     | close_range            | ✅             | 💯 |
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/fully_covered.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/fully_covered.scml
@@ -33,7 +33,7 @@ select(nfds, readfds, writefds, exceptfds, timeout);
 pidfd_open(pid, flags = PIDFD_NONBLOCK);
 
 // Obtain a duplicate of another process's file descriptor
-pidfd_getfd(pid, targetfd, flags = 0);
+pidfd_getfd(pidfd, targetfd, flags = 0);
 
 // Close all file descriptors in the inclusive range [first, last]
 close_range(first, last, flags = CLOSE_RANGE_UNSHARE | CLOSE_RANGE_CLOEXEC);
--- a/kernel/src/process/signal/signals/mod.rs
+++ b/kernel/src/process/signal/signals/mod.rs
@@ -2,6 +2,7 @@
 
 pub mod fault;
 pub mod kernel;
+pub mod raw;
 pub mod user;
 
 use core::{any::Any, fmt::Debug};
--- a/kernel/src/process/signal/signals/user.rs
+++ b/kernel/src/process/signal/signals/user.rs
@@ -3,12 +3,15 @@
 #![expect(dead_code)]
 
 use super::Signal;
-use crate::process::{
-    Pid, Uid,
-    signal::{
-        c_types::siginfo_t,
-        constants::{SI_QUEUE, SI_TKILL, SI_USER},
-        sig_num::SigNum,
+use crate::{
+    context::Context,
+    process::{
+        Pid, Uid,
+        signal::{
+            c_types::siginfo_t,
+            constants::{SI_QUEUE, SI_TKILL, SI_USER},
+            sig_num::SigNum,
+        },
     },
 };
 
@@ -37,6 +40,15 @@ impl UserSignal {
         }
     }
 
+    pub fn new_kill(num: SigNum, ctx: &Context) -> Self {
+        Self {
+            num,
+            kind: UserSignalKind::Kill,
+            pid: ctx.process.pid(),
+            uid: ctx.posix_thread.credentials().ruid(),
+        }
+    }
+
     pub fn pid(&self) -> Pid {
         self.pid
     }
@@ -58,11 +70,9 @@ impl Signal for UserSignal {
             UserSignalKind::Sigqueue => SI_QUEUE,
         };
 
-        siginfo_t::new(self.num, code)
-        // info.set_si_pid(self.pid);
-        // info.set_si_uid(self.uid);
-        // if let UserSignalKind::Sigqueue(val) = self.kind {
-        //     info.set_si_value(val);
-        // }
+        let mut info = siginfo_t::new(self.num, code);
+        info.set_pid_uid(self.pid, self.uid);
+
+        info
     }
 }
--- a/kernel/src/syscall/arch/generic.rs
+++ b/kernel/src/syscall/arch/generic.rs
@@ -82,6 +82,7 @@ macro_rules! import_generic_syscall_entries {
             open::sys_openat,
             pidfd_getfd::sys_pidfd_getfd,
             pidfd_open::sys_pidfd_open,
+            pidfd_send_signal::sys_pidfd_send_signal,
             pipe::sys_pipe2,
             ppoll::sys_ppoll,
             prctl::sys_prctl,
@@ -390,6 +391,7 @@ macro_rules! define_syscalls_with_generic_syscall_table {
             SYS_PREADV2 = 286                => sys_preadv2(args[..6]);
             SYS_PWRITEV2 = 287               => sys_pwritev2(args[..6]);
             SYS_STATX = 291                  => sys_statx(args[..5]);
+            SYS_PIDFD_SEND_SIGNAL = 424      => sys_pidfd_send_signal(args[..4]);
             SYS_PIDFD_OPEN = 434             => sys_pidfd_open(args[..2]);
             SYS_CLONE3 = 435                 => sys_clone3(args[..2], &user_ctx);
             SYS_CLOSE_RANGE = 436            => sys_close_range(args[..3]);
--- a/kernel/src/syscall/arch/x86.rs
+++ b/kernel/src/syscall/arch/x86.rs
@@ -83,6 +83,7 @@ use super::{
     pause::sys_pause,
     pidfd_getfd::sys_pidfd_getfd,
     pidfd_open::sys_pidfd_open,
+    pidfd_send_signal::sys_pidfd_send_signal,
     pipe::{sys_pipe, sys_pipe2},
     poll::sys_poll,
     ppoll::sys_ppoll,
@@ -405,6 +406,7 @@ impl_syscall_nums_and_dispatch_fn! {
     SYS_PREADV2 = 327          => sys_preadv2(args[..6]);
     SYS_PWRITEV2 = 328         => sys_pwritev2(args[..6]);
     SYS_STATX = 332            => sys_statx(args[..5]);
+    SYS_PIDFD_SEND_SIGNAL = 424 => sys_pidfd_send_signal(args[..4]);
     SYS_PIDFD_OPEN = 434       => sys_pidfd_open(args[..2]);
     SYS_CLONE3 = 435           => sys_clone3(args[..2], &user_ctx);
     SYS_CLOSE_RANGE = 436      => sys_close_range(args[..3]);
--- a/kernel/src/syscall/kill.rs
+++ b/kernel/src/syscall/kill.rs
@@ -7,10 +7,7 @@ use crate::{
         ProcessFilter, kill, kill_all, kill_group,
         signal::{
             sig_num::SigNum,
-            signals::{
-                Signal,
-                user::{UserSignal, UserSignalKind},
-            },
+            signals::{Signal, user::UserSignal},
         },
     },
 };
@@ -31,11 +28,7 @@ pub fn sys_kill(process_filter: u64, sig_num: u64, ctx: &Context) -> Result<Sysc
 }
 
 pub fn do_sys_kill(filter: ProcessFilter, sig_num: Option<SigNum>, ctx: &Context) -> Result<()> {
-    let signal = sig_num.map(|sig_num| {
-        let pid = ctx.process.pid();
-        let uid = ctx.posix_thread.credentials().ruid();
-        UserSignal::new(sig_num, UserSignalKind::Kill, pid, uid)
-    });
+    let signal = sig_num.map(|sig_num| UserSignal::new_kill(sig_num, ctx));
 
     match filter {
         ProcessFilter::Any => kill_all(signal, ctx)?,
--- a/kernel/src/syscall/mod.rs
+++ b/kernel/src/syscall/mod.rs
@@ -95,6 +95,7 @@ mod open;
 mod pause;
 mod pidfd_getfd;
 mod pidfd_open;
+mod pidfd_send_signal;
 mod pipe;
 mod poll;
 mod ppoll;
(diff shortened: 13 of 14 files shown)


kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,121 @@
...
+        si
+    };
+    let signal = RawSignal::new(siginfo);
+
+    let mut file_table = ctx.thread_local.borrow_file_table_mut();
+    let file = get_file_fast!(&mut file_table, pidfd);
+
+    let process = file.process()?;
+
+    if (flags == PidfdSendSignalFlags::ProcessGroup || process.pid() == ctx.posix_thread.tid())
```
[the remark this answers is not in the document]
li041: - Refer: `linux/kernel/signal.c`
```c
/* Only allow sending arbitrary signals to yourself. */
if ((task_pid(current) != pid || type > PIDTYPE_TGID) &&
		  (kinfo.si_code >= 0 || kinfo.si_code == SI_TKILL))
	return -EPERM;
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,124 @@
...
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum PidfdSendSignalFlags {
+    Default = 0x0,
+    Thread = 0x1,
+    ThreadGroup = 0x2,
+    ProcessGroup = 0x4,
+}
+
+impl TryFrom<u32> for PidfdSendSignalFlags {
```
StevenJiang1110: Just using `derive(TryFromInt)` for `PidfdSendSignalFlags` instead of manually implementing `TryFrom<u32>`.

book/src/kernel/linux-compatibility/syscall-flag-coverage/signals-and-timers/fully_covered.scml
```diff
@@ -16,6 +16,9 @@ kill(pid, sig);
 // Send signal to a thread
 tgkill(tgid, tid, sig);
 
+// Send signal to the target process referred to by pidfd
+pidfd_send_signal(pidfd, sig, info, flags);
```
StevenJiang1110: In SCML, for each syscall flag, we should list all supported flags. You can follow the `close_range` case as an example.

```
close_range(first, last, flags = CLOSE_RANGE_UNSHARE | CLOSE_RANGE_CLOEXEC);
```

kernel/src/fs/file_handle.rs
```diff
@@ -111,6 +111,13 @@ pub trait FileLike: Pollable + Send + Sync + Any {
     ///    collected later in its `Display::display()` method, after dropping the file table's spin
     ///    lock.
     fn dump_proc_fdinfo(self: Arc<Self>, fd_flags: FdFlags) -> Box<dyn Display>;
+
+    /// Retrieves the `Process` associated with the file.
+    ///
+    /// The return value should be the `Arc<Process>` associated with the file, or an error if no such association exists.
+    fn process(&self) -> Result<Arc<Process>> {
+        return_errno_with_message!(Errno::EBADF, "the file is not associated with a process");
+    }
```
StevenJiang1110: Adding such a method to the `FileLike` trait is almost always the wrong approach. `FileLike` should only define operations that are common to **all** files, not APIs that exist solely to support a particular file type. If we add methods to `FileLike` just for `PidFile`, the trait will quickly become bloated and will lose its value as a clean abstraction.

I understand you’re proposing this API to support opening the `/proc/<pid>` directory and returning a type that can handle `pidfd_send_signal` (i.e., a PID file). However, we can implement this in other ways. For example, we could implement a custom `open` method on the inode corresponding to `proc/<pid>` (it is currently a `ProcDir<T>`, so we may need to wrap it with another layer of abstraction).

It would be good to investigate why Linux allows open on this directory to yield a PID file, and then implement the behavior at an appropriate abstraction level. Also, I don’t think this feature is strictly necessary, and it isn’t required by the syscall implementation itself. If the investigation shows it is difficult or too invasive, we can defer it for now. If you only need it for testing, you can write tests that obtain a pidfd via `pidfd_open` instead.

kernel/src/process/kill.rs
```diff
@@ -23,8 +19,8 @@ use crate::{
 ///
 /// If `signal` is `None`, this method will only check permission without sending
 /// any signal.
-pub fn kill(pid: Pid, signal: Option<UserSignal>, ctx: &Context) -> Result<()> {
-    // Fast path: If the signal is sent to self, we can skip most check.
+pub fn kill<S: Signal + Clone + Copy>(pid: Pid, signal: Option<S>, ctx: &Context) -> Result<()> {
```
StevenJiang1110: Why change the API like this? Does this mean sending a kernel/timer signal also requires permission check?

test/initramfs/src/syscall/ltp/testcases/all.txt
```diff
@@ -1049,8 +1049,8 @@ pidfd_open02
 pidfd_open03
 pidfd_open04
 
-# pidfd_send_signal01
-# pidfd_send_signal02
+pidfd_send_signal01
+pidfd_send_signal02
 # pidfd_send_signal03
```
StevenJiang1110, line 1054: Why `pidfd_send_signal03` still fails? Add some comment about it?

kernel/src/fs/file_handle.rs
li041: Linux allows opening on this directory to yield a PID file which represents thread(`pidfd_open_file` only yields fd which represents process), and use `pidfd_send_signal` to send to the specific thread instead of process. It seems to have a good point that this feature is not necessary, because we doesn't support that funtionality. I preserved this interface for future support of operations on thread level.

kernel/src/process/kill.rs
li041: `pidfd_send_signal`  use the same function to send signal as `kill`  in Linux. And permission check is required as well. Besides, `pidfd_send_signal` only allow thread sending arbitrary signals to itself. 
```
/* Only allow sending arbitrary signals to yourself. */
if ((task_pid(current) != pid || type > PIDTYPE_TGID) &&
		  (kinfo.si_code >= 0 || kinfo.si_code == SI_TKILL))
	return -EPERM;
```

test/initramfs/src/syscall/ltp/testcases/all.txt
li041: Something in procfs is still missing.
image

kernel/src/fs/inode_handle.rs
```diff
@@ -77,6 +77,10 @@ impl InodeHandle {
         &self.path
     }
 
+    pub fn file_io(&self) -> Option<&dyn FileIo> {
+        self.file_io.as_deref()
```
StevenJiang1110: In general, we should not expose `FileIo` from `InodeHandle`. `InodeHandle` encapsulates the permission bits, and we must ensure that all reads and writes cannot bypass permission checks. If we expose `file_io`, callers could invoke methods on it directly to read/write, which would break `InodeHandle`’s abstraction and undermine its responsibility for permission enforcement.

For this PR's specific use case, I think we don't need to add such a general-purpose API. Adding an `as_pid_file` method should be sufficient. However, it must validate the permission bits and only expose a `PidFile` when the `InodeHandle` has the required permissions. (The current implementation returns `PidDirOps`, which is very strange.)

```rust
pub fn pid_file(&self) -> Option<&PidFile> {
        // check permissions and downcast self.file_io
}
```

kernel/src/fs/procfs/pid/mod.rs
```diff
@@ -87,4 +97,48 @@ impl DirOps for PidDirOps {
...
+        _status_flags: StatusFlags,
+    ) -> Result<usize> {
+        Err(Error::new(Errno::EISDIR))
+    }
+}
+
+impl FileIo for PidDirOps {
+    fn check_seekable(&self) -> Result<()> {
+        Ok(())
+    }
```
StevenJiang1110: This looks strange. `PidDirOps` supports `lseek`? Better adding a test case to demonstrate it.

kernel/src/fs/procfs/pid/mod.rs
```diff
@@ -87,4 +97,48 @@ impl DirOps for PidDirOps {
...
         cached_children.downgrade()
     }
+
+    fn open(
+        &self,
+        _access_mode: crate::fs::utils::AccessMode,
+        _status_flags: crate::fs::utils::StatusFlags,
+    ) -> Option<Result<Box<dyn crate::fs::inode_handle::FileIo>>> {
+        Some(Ok(Box::new(self.clone())))
+    }
```
StevenJiang1110: So opening `/proc/<pid>` does not require any permission checks? That looks very strange to me.

Also, based on my understanding, opening a `PidDirOps` should return a `PidFile` rather than the directory itself. Of course, I could be mistaken, but I think you should run more experiments to validate the current behavior. For example, can the file returned by open `/proc/<pid>` be used with wait? Can it be polled?

It would be great if you could add the experiments you run as tests under `test/apps`, so it’s easier for me to understand the rationale behind this approach.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,135 @@
...
+        return_errno_with_message!(
+            Errno::EPERM,
+            "only allowed to send arbitrary signals to yourself"
+        );
+    }
+
+    match flags {
+        PidfdSendSignalFlags::Default => {
+            // FIXME: Infer scope from the type of pidfd.
+            // For now, we always treat it as process scope.
```
StevenJiang1110: The issue is that when we implemented this syscall, the Linux man pages hadn’t been updated yet, so we assumed a pidfd could only represent a process. Starting with Linux 6.9, however, a pidfd can also represent a thread.

Reword the comment here:
```rust
// FIXME: On Linux, a pidfd can refer to either a process or a thread.
// We currently only support pidfds that refer to processes.
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,135 @@
...
+        PidfdSendSignalFlags::Thread => {
+            // FIXME: Send signal to specific thread.
+            // For now, we just send to the main thread.
+            tgkill(
+                process.main_thread().as_posix_thread().unwrap().tid(),
+                process.pid(),
+                Some(signal),
+                ctx,
+            )?;
+        }
```
StevenJiang1110, line 58: Does Linux validate whether the `PidfdSendSignalFlags` type matches the kind of object the pidfd refers to (process vs. thread)? 

You could also consider fixing the PidFile implementation so that a PidFile can represent either a thread or a process, e.g.:

```rust
struct PidFile {
    inner: Inner,
    // other fields
}

enum Inner {
    Process(Weak<Process>),
    Thread(Weak<Thread>),
}
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut si = siginfo_t::new(sig_num, SI_USER);
+        let pid = ctx.process.pid();
+        let uid = ctx.posix_thread.credentials().ruid();
+        si.set_pid_uid(pid, uid);
+        si
+    };
+    let signal = RawSignal::new(siginfo);
+
+    let mut file_table = ctx.thread_local.borrow_file_table_mut();
+    let file = get_file_fast!(&mut file_table, pidfd);
```
StevenJiang1110: pidfd can also have other special cases, like PIDFD_SELF_THREAD?

https://elixir.bootlin.com/linux/v6.18/source/kernel/signal.c#L4081

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+        si
+    } else {
+        // If `info_ptr` is NULL, the kernel constructs a default `siginfo_t` structure
+        // whose fields match the values that are implicitly supplied when a signal is sent using the kill(2).
+        let mut si = siginfo_t::new(sig_num, SI_USER);
+        let pid = ctx.process.pid();
+        let uid = ctx.posix_thread.credentials().ruid();
+        si.set_pid_uid(pid, uid);
+        si
```
StevenJiang1110: If `siginfo` is null, we should reuse the logic that kill syscall use (If kill misses some infomation, you can fix the kill implementation). Something like `UserSignal::new(sig_num).to_siginfo()`.

kernel/src/syscall/pidfd_send_signal.rs
li041: I'm Working on Linux kernel 6.11, and I implemented `PIDFD_SELF_THREAD` and `PIDFD_SELF_THREAD_GROUP` at first and tested them on my envioriment which returned `EINVAL`, and I saw these flags don't exist in [man7]()https://man7.org/linux/man-pages/man2/pidfd_send_signal.2.html), so I dropped relative code.

kernel/src/fs/inode_handle.rs
li041: I see. Check is carried out during the opening of file and before we send signal . And now I only use this fd to get coressponding process.

kernel/src/syscall/pidfd_send_signal.rs
li041: To be honest, I don't find a good way to do so.
`kill` syscall populate siginfo in `do_sys_kill`, and then pass it to `kill_*`. 
```
pub fn do_sys_kill(filter: ProcessFilter, sig_num: Option<SigNum>, ctx: &Context) -> Result<()> {
    let signal = sig_num.map(|sig_num| {
        let pid = ctx.process.pid();
        let uid = ctx.posix_thread.credentials().ruid();
        UserSignal::new(sig_num, UserSignalKind::Kill, pid, uid)
    });
```
It may be a good choice that if `signal` argument is null, then use default one. However, it means only check perm but do not enqueue signal in asterinas.
```
fn kill_process<S: Signal + Clone + Copy>(
    process: &Process,
    signal: Option<S>,
    ctx: &Context,
) -> Result<()> {
	...
    check_signal_perm(target_main_thread.as_posix_thread().unwrap(), ctx, signum)?;
    ...
    if let Some(signal) = signal {
        process.enqueue_signal(signal);
    }
}
```
So if I still want to use `kill_*` to check perm and enqueue signal, I should prepare siginfo before I call them.

StevenJiang1110: I mean you could reuse the logic for constructing a UserSignal in both kill and pidfd_send_signal, rather than duplicating it.

For example, introduce a method like this for both syscalls:

```rust
impl UserSignal {
    pub fn new_kill(sig_num: SigNum, ctx: &Context) -> Self { 
         let pid = ctx.process.pid();
        let uid = ctx.posix_thread.credentials().ruid();
        UserSignal::new(sig_num, UserSignalKind::Kill, pid, uid)
    }
}
```

kernel/src/syscall/pidfd_send_signal.rs
StevenJiang1110: Yes, these are fairly new flags. If they’re not difficult to implement, we can include them in this PR as well. Their semantics seem straightforward, so it should be acceptable even if we don’t test them locally.

kernel/src/fs/procfs/pid/mod.rs
li041: It should return `Ok(())`, or it won't pass `getdents` test in gVisor.
image

kernel/src/process/kill.rs
```diff
@@ -130,7 +134,11 @@ pub fn kill_all(signal: Option<UserSignal>, ctx: &Context) -> Result<()> {
     result
 }
 
-fn kill_process(process: &Process, signal: Option<UserSignal>, ctx: &Context) -> Result<()> {
+fn kill_process<S: Signal + Clone + Copy>(
+    process: &Process,
+    signal: Option<S>,
+    ctx: &Context,
+) -> Result<()> {
```
StevenJiang1110: I think `S` does not needs to be `Copy`. `Clone` should be enough. 

You can change the line 142 as

```rust
let signum = signal.as_ref().map(|signal| signal.num());
```

to make it compile.

Please change other functions as well.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+fn get_siginfo_from_user(info_ptr: Vaddr, sig_num: SigNum, ctx: &Context) -> Result<siginfo_t> {
+    if info_ptr != 0 {
+        let si = ctx
+            .user_space()
+            .read_val::<siginfo_t>(info_ptr)
+            .map_err(|_| {
+                Error::with_message(Errno::EFAULT, "failed to read siginfo from user space")
+            })?;
```
StevenJiang1110: Here `map_err` seems to be useless, you can directly use `?` to return the `read_val` result.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,153 @@
...
+        Ok(UserSignal::new_kill(sig_num, ctx).to_info())
+    }
+}
+
+fn get_process_from_pidfd(pidfd: FileDesc, ctx: &Context) -> Result<Arc<Process>> {
+    match pidfd {
+        PIDFD_SELF_THREAD => {
+            // FIXME: On Linux, the signal can be sent to any thread.
+            // We currently only support pidfds that refer to processes.
+            Ok(ctx.posix_thread.process())
```
StevenJiang1110: Even we don't support pidfds that refer to threads. I think we can support `PIDFD_SELF_THREAD` correctly. It does not related to pidfd at all. We don't need to leave a FIXME here.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+        PIDFD_SELF_THREAD_GROUP => Ok(ctx.process.clone()),
```
StevenJiang1110: The `clone` can be avoided here. You can let this method return `Result<Cow<'_, Arc<Process>>>`, then return `Cow::Borrow(ctx.process)` here.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,137 @@
...
+        PidfdSendSignalFlags::ProcessGroup => {
+            let pgid = process.pgid();
+            kill_group(pgid, Some(signal), ctx)?;
+        }
+    }
+
+    Ok(SyscallReturn::Return(0))
+}
+
+fn get_siginfo_from_user(info_ptr: Vaddr, sig_num: SigNum, ctx: &Context) -> Result<siginfo_t> {
```
StevenJiang1110: Raname it as `read_siginfo_from_user`.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,113 @@
...
+// }
+// END_SETUP()
+
+// FN_TEST(pidfd_send_signal_thread)
+// {
+// 	TEST_ERRNO(pidfd_send_signal(thread_pidfd, sig, &siginfo, 0), EBADF);
+// 	TEST_SUCC(pidfd_send_signal(pidfd, sig, &siginfo, 0));
+// 	TEST_SUCC(pthread_join(thread, NULL));
+// }
+// END_TEST()
```
StevenJiang1110: We don't need to comment out these code. Just using the __asterinas__ macro. 
```C
#ifndef __asterinas__
	......
#endif
```

StevenJiang1110 (commented):
Thanks for your work. Just a few minor comments. Please also mark those outdated comments as resolved.

kernel/src/process/kill.rs
li041: `Copy` is necessary for Some of them which send signal to multiple object.

StevenJiang1110: `Clone` is enough even when you try to send signal to multiple targets. Why must the signal be `Copy`?

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,130 @@
...
+    } else {
+        // If `info_ptr` is NULL, the kernel constructs a default `siginfo_t` structure
+        // whose fields match the values that are implicitly supplied when a signal is sent using the kill(2).
+        Ok(UserSignal::new_kill(sig_num, ctx).to_info())
+    }
+}
+
+fn get_process_from_pidfd<'a>(pidfd: FileDesc, ctx: &'a Context) -> Result<Cow<'a, Arc<Process>>> {
+    match pidfd {
+        PIDFD_SELF_THREAD => Ok(Cow::Owned(ctx.posix_thread.process())),
```
StevenJiang1110: The current implementation doesn’t handle `PIDFD_SELF_THREAD` when the pidfd flags are `PidfdSendSignalFlags::Default`, right? The expected behavior is to send the signal to the current thread, but the current implementation sends it to the current process instead. We could refactor the code to support the correct behavior.

For example, you could introduce an enum like this:

```rust
enum Target<'a> {
     Thread(Cow<'a, Arc<Thread>>),
     Process(Cow<'a, Arc<Process>>),
}
```

and let `get_process_from_pidfd` return this enum.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+        Ok(si)
+    } else {
+        // If `info_ptr` is NULL, the kernel constructs a default `siginfo_t` structure
+        // whose fields match the values that are implicitly supplied when a signal is sent using the kill(2).
+        Ok(UserSignal::new_kill(sig_num, ctx).to_info())
+    }
+}
+
+fn get_process_from_pidfd<'a>(pidfd: FileDesc, ctx: &'a Context) -> Result<Cow<'a, Arc<Process>>> {
```
StevenJiang1110: Remove redundnt lifetime here 
```suggestion
fn get_process_from_pidfd(pidfd: FileDesc, ctx: &Context) -> Result<Cow<'_, Arc<Process>>> {
```

kernel/src/process/kill.rs
```diff
@@ -23,8 +19,8 @@ use crate::{
 ///
 /// If `signal` is `None`, this method will only check permission without sending
 /// any signal.
-pub fn kill(pid: Pid, signal: Option<UserSignal>, ctx: &Context) -> Result<()> {
-    // Fast path: If the signal is sent to self, we can skip most check.
+pub fn kill<S: Signal + Clone>(pid: Pid, signal: Option<S>, ctx: &Context) -> Result<()> {
```
lrh2000: Using generics in this way will cause all the methods here (`kill`/`kill_group`/`tgkill`/`kill_process`) to be [monomorphized](https://rustc-dev-guide.rust-lang.org/backend/monomorph.html). This does not really seem necessary to me. Is it possible to accept `Box<dyn Signal>` instead?

StevenJiang1110 (approved):

kernel/src/process/signal/signals/raw.rs
```diff
@@ -0,0 +1,37 @@
...
+    info: siginfo_t,
+}
+
+impl Debug for RawSignal {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("RawSignal")
+            .field("signo", &self.info.si_signo)
+            .field("errno", &self.info.si_errno)
+            .field("code", &self.info.si_code)
+            .finish()
```
lrh2000:
```suggestion
            .finish_non_exhaustive()
```

kernel/src/process/signal/signals/raw.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+impl RawSignal {
+    // Caller must ensure the si_signo is valid.
```
lrh2000:
```suggestion
    // The caller should ensure that `info.si_signo` is a valid signal number.
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+    Ok(SyscallReturn::Return(0))
+}
+
+fn read_siginfo_from_user(info_ptr: Vaddr, sig_num: SigNum, ctx: &Context) -> Result<siginfo_t> {
+    if info_ptr != 0 {
+        let si = ctx.user_space().read_val::<siginfo_t>(info_ptr)?;
+        if si.si_signo != sig_num.as_u8() as i32 {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "siginfo.si_signo does not match the specified signal number"
```
lrh2000:
```suggestion
                "`siginfo.si_signo` does not match the specified signal number"
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+    let is_self = match &target {
+        SignalTarget::Thread { tid, tgid: _ } => *tid == ctx.posix_thread.tid(),
+        SignalTarget::Process { pid } => *pid == ctx.process.pid(),
+        SignalTarget::ProcessGroup { pgid: _ } => false,
+    };
+
+    if is_self && (siginfo.si_code >= 0 || siginfo.si_code == SI_TKILL) {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "pidfd is only allowed to send arbitrary signals to itself"
```
lrh2000:
```suggestion
            "signals with custom code can only be sent to the current process"
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+            pid: ctx.process.pid(),
+        }),
+        _ => {
+            let mut file_table = ctx.thread_local.borrow_file_table_mut();
+            let file = get_file_fast!(&mut file_table, pidfd);
+
+            // FIXME: On Linux, a pidfd can be also obtained by opening a `/proc/pid` directory.
+            // Reference: <https://man7.org/linux/man-pages/man2/pidfd_send_signal.2.html>
+            let Some(pid_file) = file.downcast_ref::<PidFile>() else {
+                return_errno_with_message!(Errno::EBADF, "not a pidfd");
```
lrh2000:
```suggestion
                return_errno_with_message!(Errno::EBADF, "the file is not a PID file");
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+
+            let target = match flags {
+                PidfdSendSignalFlags::Default => {
+                    // FIXME: On Linux, a pidfd can refer to either a process or a thread.
+                    // We currently only support pidfds that refer to processes.
+                    SignalTarget::Process { pid: process.pid() }
+                }
+                PidfdSendSignalFlags::Thread => {
+                    // FIXME: On Linux, the signal can be sent to any thread.
+                    // We currently only support main_thread.
```
lrh2000:
```suggestion
                    // We currently only support the main thread.
```

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,115 @@
...
+#include <sys/syscall.h>
+#include <sys/wait.h>
+#include <fcntl.h>
+
+#include "../test.h"
+
+const int sig = SIGUSR1;
+volatile sig_atomic_t signal_received = 0;
+
+pid_t pid;
```
lrh2000: This file's code is hard to follow for various reasons. For instance, some variable names are misleading. Additionally, the thread code and process code are mixed together.

I almost completely rewrote it, but I didn't test it at all. I hope the following diff helps you understand my approach:
```diff
diff --git a/test/initramfs/src/apps/signal_c/pidfd_send_signal.c b/test/initramfs/src/apps/signal_c/pidfd_send_signal.c
index 858ecb51a..c30097e32 100644
--- a/test/initramfs/src/apps/signal_c/pidfd_send_signal.c
+++ b/test/initramfs/src/apps/signal_c/pidfd_send_signal.c
@@ -12,16 +12,6 @@
 #include "../test.h"
 
 const int sig = SIGUSR1;
-volatile sig_atomic_t signal_received = 0;
-
-pid_t pid;
-pid_t thread_tid;
-pthread_t thread;
-
-char path[256];
-int process_pidfd;
-int thread_pidfd;
-int pidfd;
 
 siginfo_t siginfo;
 
@@ -36,39 +26,33 @@ static int pidfd_send_signal(int pidfd, int sig, siginfo_t *info,
 	return syscall(SYS_pidfd_send_signal, pidfd, sig, info, flags);
 }
 
-void signal_handler(int sig)
+FN_SETUP(siginfo)
 {
-	signal_received = 1;
+	CHECK(memset(&siginfo, 0, sizeof(siginfo)));
+	siginfo.si_signo = sig;
+	siginfo.si_code = -666;
+	siginfo.si_pid = getpid();
+	siginfo.si_uid = getuid();
 }
+END_SETUP()
 
-void *thread_func(void *arg)
-{
-	thread_tid = syscall(SYS_gettid);
-	signal(sig, signal_handler);
+// ==========================
+//    Tests for processes
+// ==========================
 
-	while (!signal_received) {
-		usleep(100);
-	}
-
-	return NULL;
-}
+int process_pid;
+int process_pidfd;
 
 FN_SETUP(create_process)
 {
-	pid = CHECK(fork());
-	if (pid == 0) {
+	process_pid = CHECK(fork());
+	if (process_pid == 0) {
 		while (1) {
 			usleep(100);
 		}
 	}
 
-	process_pidfd = CHECK(pidfd_open(pid, 0));
-
-	CHECK(memset(&siginfo, 0, sizeof(siginfo)));
-	siginfo.si_signo = sig;
-	siginfo.si_code = -666;
-	siginfo.si_pid = pid;
-	siginfo.si_uid = getuid();
+	process_pidfd = CHECK(pidfd_open(process_pid, 0));
 }
 END_SETUP()
 
@@ -76,41 +60,78 @@ FN_TEST(pidfd_send_signal_process)
 {
 	TEST_SUCC(pidfd_send_signal(process_pidfd, sig, &siginfo, 0));
 
-	TEST_SUCC(waitid(P_PID, pid, NULL, WNOWAIT | WEXITED));
+	TEST_SUCC(waitid(P_PID, process_pid, NULL, WNOWAIT | WEXITED));
 }
 END_TEST()
 
+FN_SETUP(cleanup_process)
+{
+	CHECK(close(process_pidfd));
+}
+END_SETUP()
+
 // FIXME: Enable thread tests once pidfd for threads is supported
 #ifndef __asterinas__
+// ==========================
+//     Tests for threads
+// ==========================
+
+volatile sig_atomic_t signal_received = 0;
+
+pthread_t thread;
+pid_t thread_tid;
+
+int proc_fd;
+int proc_task_fd;
+
+void signal_handler(int sig)
+{
+	signal_received = 1;
+}
+
+void *thread_func(void *arg)
+{
+	thread_tid = syscall(SYS_gettid);
+	signal(sig, signal_handler);
+
+	while (!signal_received) {
+		usleep(100);
+	}
+
+	return NULL;
+}
+
 FN_SETUP(create_thread)
 {
+	static char path[256];
+
 	CHECK(pthread_create(&thread, NULL, thread_func, NULL));
-	usleep(100);
+
+	while (thread_tid != 0) {
+		usleep(100);
+	}
 
 	snprintf(path, sizeof(path), "/proc/%d/", thread_tid);
-	pidfd = CHECK(open(path, O_RDONLY | O_CLOEXEC));
+	proc_fd = CHECK(open(path, O_RDONLY | O_CLOEXEC));
 
 	snprintf(path, sizeof(path), "/proc/%d/task/%d", getpid(), thread_tid);
-	thread_pidfd = CHECK(open(path, O_DIRECTORY | O_CLOEXEC));
+	proc_task_fd = CHECK(open(path, O_DIRECTORY | O_CLOEXEC));
 }
 END_SETUP()
 
 FN_TEST(pidfd_send_signal_thread)
 {
-	TEST_ERRNO(pidfd_send_signal(thread_pidfd, sig, &siginfo, 0), EBADF);
-	TEST_SUCC(pidfd_send_signal(pidfd, sig, &siginfo, 0));
+	TEST_ERRNO(pidfd_send_signal(proc_task_fd, sig, &siginfo, 0), EBADF);
+
+	TEST_SUCC(pidfd_send_signal(proc_fd, sig, &siginfo, 0));
 	TEST_SUCC(pthread_join(thread, NULL));
 }
 END_TEST()
-#endif
 
-FN_SETUP(cleanup)
+FN_TEST(cleanup_thread)
 {
-	CHECK(close(process_pidfd));
-#ifndef __asterinas__
-	CHECK(close(pidfd));
-	CHECK(close(thread_pidfd));
-#endif
+	CHECK(close(proc_task_fd));
+	CHECK(close(proc_fd));
 }
 END_SETUP()
-
+#endif
```

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,115 @@
...
+FN_SETUP(create_thread)
+{
+	CHECK(pthread_create(&thread, NULL, thread_func, NULL));
+	usleep(100);
+
+	snprintf(path, sizeof(path), "/proc/%d/", thread_tid);
+	pidfd = CHECK(open(path, O_RDONLY | O_CLOEXEC));
+
+	snprintf(path, sizeof(path), "/proc/%d/task/%d", getpid(), thread_tid);
+	thread_pidfd = CHECK(open(path, O_DIRECTORY | O_CLOEXEC));
```
lrh2000: I'm not sure why you're using different flags for `pidfd` and `thread_pidfd`. This makes it unclear whether the result is related to the flags.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,115 @@
...
+	siginfo.si_signo = sig;
+	siginfo.si_code = -666;
+	siginfo.si_pid = pid;
+	siginfo.si_uid = getuid();
+}
+END_SETUP()
+
+FN_TEST(pidfd_send_signal_process)
+{
+	TEST_SUCC(pidfd_send_signal(process_pidfd, sig, &siginfo, 0));
```
lrh2000, line 80: Add some tests for the error cases? For example, I think these two cases worth testing:
```rust
        if si.si_signo != sig_num.as_u8() as i32 {
            return_errno_with_message!(
                Errno::EINVAL,
```
```rust
    if is_self && (siginfo.si_code >= 0 || siginfo.si_code == SI_TKILL) {
        return_errno_with_message!(
            Errno::EPERM,
```

lrh2000 (commented):
Thanks for implementing `pidfd_send_signal` system call!

lrh2000:
> * It seems that `/proc/<pid>` represents specific `task` in linux kernel, which is the same semantic as `thread` in asterinas. But when we open `<proc/<pid>` in asterinas, it is always a process.

This seems quite surprising to me. I was really shocked by this.

Although it seems that `/proc/[tid]` does not exist (in `ls /proc`), it also does (by opening it directly). It is in a mysterious mixed state.
```
> ls /proc/611/task
611  628  629  631
> ls /proc/ -al | grep ' 611$'
dr-xr-xr-x   9 root             root                           0 Jan 27 15:13 611
> ls /proc/ -al | grep ' 628$'
> ls /proc/ -al | grep ' 629$'
> ls /proc/628/
arch_status  clear_refs          cwd      gid_map            loginuid   mounts      oom_adj        projid_map  setgroups     statm           timers
attr         cmdline             environ  io                 maps       mountstats  oom_score      root        smaps         status          timerslack_ns
autogroup    comm                exe      ksm_merging_pages  map_files  net         oom_score_adj  sched       smaps_rollup  syscall         uid_map
auxv         coredump_filter     fd       ksm_stat           mem        ns          pagemap        schedstat   stack         task            wchan
cgroup       cpu_resctrl_groups  fdinfo   limits             mountinfo  numa_maps   personality    sessionid   stat          timens_offsets
> ls /proc/629/
arch_status  clear_refs          cwd      gid_map            loginuid   mounts      oom_adj        projid_map  setgroups     statm           timers
attr         cmdline             environ  io                 maps       mountstats  oom_score      root        smaps         status          timerslack_ns
autogroup    comm                exe      ksm_merging_pages  map_files  net         oom_score_adj  sched       smaps_rollup  syscall         uid_map
auxv         coredump_filter     fd       ksm_stat           mem        ns          pagemap        schedstat   stack         task            wchan
cgroup       cpu_resctrl_groups  fdinfo   limits             mountinfo  numa_maps   personality    sessionid   stat          timens_offsets
```

So:
 1. Threads do not show up when listing the directory under `/proc`.
 2. But opening the directory succeeds, and there are indeed files under the thread path.

kernel/src/process/kill.rs
li041: Maybe `clone_box` method is needed to use `Box<dyn Signal>`? I didn't really find a good way to solve the distribution of signal in `kill_all` with a few changes.

lrh2000: Since there are legal reasons for it and the function bodies are relatively small, `kill_group` and `kill_all` can keep the current form.

What about `tgkill` and `kill`? Additionally, `Process::enqueue_signal` should be fixed to accept only a `Box<dyn Signal>` (though not introduced in this PR)?

kernel/src/syscall/timer_create.rs
```diff
@@ -45,9 +45,9 @@ pub fn sys_timer_create(
         // If `sigevent_addr` is NULL, use the default method (like `sys_alarm`) to send signal.
         if sigevent_addr == 0 {
             let process = current_process.clone();
-            let signal = KernelSignal::new(SIGALRM);
+            let signal = Box::new(KernelSignal::new(SIGALRM));
             Box::new(move || {
-                process.enqueue_signal(signal);
+                process.enqueue_signal(signal.clone());
```
lrh2000: No need to clone the signal here?

kernel/src/syscall/timer_create.rs
```diff
@@ -60,9 +60,9 @@ pub fn sys_timer_create(
                 // Send a signal to the current process when the timer is expired.
                 SigNotify::SIGEV_SIGNAL => {
                     let process = current_process.clone();
-                    let signal = KernelSignal::new(SigNum::try_from(signo as u8)?);
+                    let signal = Box::new(KernelSignal::new(SigNum::try_from(signo as u8)?));
                     Box::new(move || {
-                        process.enqueue_signal(signal);
+                        process.enqueue_signal(signal.clone());
```
lrh2000: Same?

kernel/src/process/signal/signals/raw.rs
```diff
@@ -0,0 +1,37 @@
...
+        f.debug_struct("RawSignal")
+            .field("signo", &self.info.si_signo)
+            .field("errno", &self.info.si_errno)
+            .field("code", &self.info.si_code)
+            .finish_non_exhaustive()
+    }
+}
+
+impl RawSignal {
+    /// The caller must ensure that `info.si_signo` is a valid signal number.
```
lrh2000: I don't mean to make this a doc comment (your previous version uses normal comments here).

However, if it's a doc comment, let's add a proper summary sentence anyway.
```rust
/// A signal that carries raw [`siginfo_t`] information.
pub struct RawSignal;

/// Creates a signal that carries raw [`siginfo_t`] information.
///
/// The caller must ensure that `info.si_signo` is a valid signal number.
fn new(...) -> Self {}
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+
+            let target = match flags {
+                PidfdSendSignalFlags::Default => {
+                    // FIXME: On Linux, a pidfd can refer to either a process or a thread.
+                    // We currently only support pidfds that refer to processes.
+                    SignalTarget::Process { pid: process.pid() }
+                }
+                PidfdSendSignalFlags::Thread => {
+                    // FIXME: On Linux, the signal can be sent to any thread.
+                    // We currently only support the main_thread.
```
lrh2000:
```suggestion
                    // We currently only support the main thread.
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+enum SignalTarget {
+    Thread { tid: Tid, tgid: Pid },
+    Process { pid: Pid },
+    ProcessGroup { pgid: Pgid },
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, TryFromInt)]
```
lrh2000, line 159:
```
// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/pidfd.h#L19>.
```

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+fn get_target_from_pidfd(
+    pidfd: FileDesc,
+    flags: PidfdSendSignalFlags,
+    ctx: &Context,
+) -> Result<SignalTarget> {
+    match pidfd {
+        PIDFD_SELF_THREAD => Ok(SignalTarget::Thread {
```
lrh2000: What happens if someone uses `pidfd_send_signal(PIDFD_SELF_THREAD, PIDFD_SIGNAL_THREAD)`.

The current implementation does not quite align well to the Linux behavior? It seems the flag always takes priority in the Linux implementation:
https://elixir.bootlin.com/linux/v6.18/source/kernel/signal.c#L4112

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,150 @@
...
+
+	while (thread_tid == 0) {
+		usleep(100);
+	}
+
+	snprintf(path, sizeof(path), "/proc/%d/", thread_tid);
+	proc_fd = CHECK(open(path, O_RDONLY | O_CLOEXEC));
+
+	snprintf(path, sizeof(path), "/proc/%d/task/%d", getpid(), thread_tid);
+	proc_task_fd = CHECK(open(path, O_DIRECTORY | O_CLOEXEC));
```
lrh2000, line 241: Response to https://github.com/asterinas/asterinas/pull/2912#discussion_r2739755593? Is it possible to use the same flag here?

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,150 @@
...
+// FIXME: Enable thread tests once pidfd for threads is supported
+#ifndef __asterinas__
+/* ==========================
+ *     Tests for threads
+ * ========================== */
+
+volatile sig_atomic_t signal_received = 0;
+
+pthread_t thread;
+pid_t thread_tid;
```
lrh2000: Mark this as `volatile` as well, since it's written in a different thread.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,155 @@
...
+        },
+        tgkill,
+    },
+    syscall::SyscallReturn,
+    thread::Tid,
+};
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fcntl.h#L110>.
+const PIDFD_SELF_THREAD: i32 = -10000;
+const PIDFD_SELF_THREAD_GROUP: i32 = -10001;
```
lrh2000: The two constants can be moved above or below `PidfdSendSignalFlags`? (Since they're similar constants/flags and used in similar places).

kernel/src/syscall/timer_create.rs
li041: Error message: `Fn` and `FnMut` closures require captured values to be able to be consumed multiple times, but `FnOnce` closures may consume them only once. And `Box<KernelSignal>` is not `Copy`.

kernel/src/process/signal/signals/raw.rs
li041: okk, maybe normal comments is enough.

kernel/src/syscall/timer_create.rs
lrh2000: Oh sorry. Got it. Thanks for the explanation.

kernel/src/process/signal/signals/raw.rs
lrh2000: But since there is already a (better? I assume) doc comment version, I guess we can use the version above :(

kernel/src/syscall/timer_create.rs
lrh2000: Just took another look here, given `KernelSignal` is a tiny struct that impls `Copy`:
```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub struct KernelSignal {
    num: SigNum,
}
```

I think the following code snippet should work better?
```rust
            let signal = KernelSignal::new(SIGALRM);
            Box::new(move || {
                process.enqueue_signal(Box::new(signal));
            )
```
It will avoid the heap allocation for the initial `KernelSignal`, which does not sound necessary to me. Will it work?

Same to https://github.com/asterinas/asterinas/pull/2912#discussion_r2744466910.

li041: Yeah, it's better and works well.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,170 @@
...
+        },
+        PIDFD_SELF_THREAD_GROUP => match flags {
+            PidfdSendSignalFlags::Default | PidfdSendSignalFlags::ThreadGroup => {
+                Ok(process_target(ctx.process.pid()))
+            }
+            PidfdSendSignalFlags::Thread => Ok(thread_target(
+                ctx.process.main_thread().as_posix_thread().unwrap().tid(),
+                ctx.process.pid(),
+            )),
+            PidfdSendSignalFlags::ProcessGroup => Ok(group_target(ctx.process.pgid())),
```
lrh2000: I have no idea how to review this without some tests for it... "take priority" in my previous comment is a vague word that really needs more information to confirm the Linux behavior.

I am afraid that the case for the process group needs to be `pid()`. The following seems to pass:
```c
// PIDFD_SELF_THREAD/PIDFD_SELF_THREAD_GROUP won't work with
// PIDFD_SIGNAL_PROCESS_GROUP unless the current process is
// the process group leader.
FN_TEST(pidfd_send_signal_self_process_group)
{
	pid_t pid;
	int stat;

	pid = TEST_SUCC(fork());
	if (pid == 0) {
		setup_test_siginfo(&siginfo, SIGTERM, -666);

		CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD, SIGTERM,
					     &siginfo,
					     PIDFD_SIGNAL_PROCESS_GROUP),
			   _ret == -1 && errno == ESRCH);
		CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD_GROUP, SIGTERM,
					     &siginfo,
					     PIDFD_SIGNAL_PROCESS_GROUP),
			   _ret == -1 && errno == ESRCH);

		exit(0);
	}

	TEST_RES(wait(&stat), WIFEXITED(stat) && WEXITSTATUS(stat) == 0);
}
END_TEST()
```
But the remaining code should be correct given the following test passes:
```c
void *pidfd_send_signal_self_child_thread(void *arg)
{
	setup_test_siginfo(&siginfo, SIGTERM, -666);

	CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD, SIGTERM, &siginfo,
				     PIDFD_SIGNAL_THREAD_GROUP),
		   _ret == 0);

	return NULL;
}

// PIDFD_SELF_THREAD will work with PIDFD_SIGNAL_THREAD_GROUP
// even if the current thread is not the thread group leader.
FN_TEST(pidfd_send_signal_self_thread_group)
{
	pid_t pid;
	pthread_t thread;
	int stat;

	pid = TEST_SUCC(fork());
	if (pid == 0) {
		CHECK(pthread_create(&thread, NULL,
				 &pidfd_send_signal_self_child_thread, NULL));
		CHECK(pthread_join(thread, NULL));
		exit(0);
	}

	TEST_RES(wait(&stat), WIFSIGNALED(stat) && WTERMSIG(stat) == SIGTERM);
}
END_TEST()
```

Check out the regression tests and commit them to ensure our behavior is correct.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: MPL-2.0
```
lrh2000, line 1: In addition, you need to update `test/initramfs/src/apps/scripts/process.sh` so that this test can be checked in the CI.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,170 @@
...
+                    thread_target(
+                        process.main_thread().as_posix_thread().unwrap().tid(),
+                        process.pid(),
+                    )
+                }
+                PidfdSendSignalFlags::ThreadGroup => process_target(process.pid()),
+                PidfdSendSignalFlags::ProcessGroup => group_target(process.pgid()),
+            };
+
+            Ok(target)
```
lrh2000: The `let target = ...; Ok(target)` pattern should now be moved to the outside layer:
```rust
let target = match pidfd {
    PIDFD_SELF_THREAD => ...,
    ...
};

Ok(target)
```

So you don't need to repeat `Ok()` too many times in the `PIDFD_SELF_*` branch.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,170 @@
...
+    let process_target = |pid: Pid| SignalTarget::Process { pid };
+    let group_target = |pgid: Pgid| SignalTarget::ProcessGroup { pgid };
+
+    match pidfd {
+        PIDFD_SELF_THREAD => match flags {
+            PidfdSendSignalFlags::Default | PidfdSendSignalFlags::Thread => {
+                Ok(thread_target(ctx.posix_thread.tid(), ctx.process.pid()))
+            }
+            PidfdSendSignalFlags::ThreadGroup => Ok(process_target(ctx.process.pid())),
+            PidfdSendSignalFlags::ProcessGroup => Ok(group_target(ctx.process.pid())),
```
lrh2000: Actually, I think we still miss a test to check what will happen if we send a signal to SELF_THREAD and PROCESS_GROUP in a _non-main_ thread of a process that _is_ the process group leader. The result of that test will determine whether we should use `tid()` in:
```rust
    match pidfd {
        PIDFD_SELF_THREAD => match flags {
            // ...
            PidfdSendSignalFlags::ProcessGroup => Ok(group_target(/* pid() or tid() ?? */)),
        },
```

We should be able to confirm this in `pidfd_send_signal_self_process_group` (or another test near it), I think.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,224 @@
...
+END_TEST()
+
+FN_SETUP(cleanup_process)
+{
+	CHECK(close(process_pidfd));
+}
+END_SETUP()
+
+/* ==========================
+ *    Tests for `PIDFD_SELF_THREAD` and `PIDFD_SELF_THREAD_GROUP`
```
lrh2000:
```c
/* ==========================
 *  Tests for `PIDFD_SELF_*`
 * ========================== */
```

Try to make this shorter than the "=" markers; otherwise, it looks a bit ugly.

kernel/src/syscall/pidfd_send_signal.rs
```diff
@@ -0,0 +1,170 @@
...
+    );
+
+    let siginfo = read_siginfo_from_user(info_ptr, sig_num, ctx)?;
+    let signal = RawSignal::new(siginfo);
+
+    let target = get_target_from_pidfd(pidfd, flags, ctx)?;
+
+    let is_self = match &target {
+        SignalTarget::Thread { tid: _, tgid } => *tgid == ctx.process.pid(),
+        SignalTarget::Process { pid } => *pid == ctx.process.pid(),
```
lrh2000: If I read the Linux implementation correctly, it checks TIDs for both `SignalTarget::Thread` and `SignalTarget::Process` cases.

Then if we send a signal from the thread other than the main thread to the current process, `is_self` is also `false`. Possible to confirm (or disconfirm) this?
```rust
        SignalTarget::Process { pid } => *pid == ctx.thread.tid(),
```

Note that `task_pid` actually returns the thread ID: https://elixir.bootlin.com/linux/v6.18.6/source/kernel/signal.c#L4038

To be honest, I think the exact Linux behavior is a bit odd. I am not sure if it is the intended behavior or it just occasionally works for some historical reasons.

(I'm not sure why you changed the TID check to the TGID check in your last revision. But thanks to it, I noticed this issue; otherwise, it would have been pretty likely to be missed...)

test/initramfs/src/apps/scripts/process.sh
```diff
@@ -75,6 +75,7 @@ signal_c/signal_fd
 signal_c/signal_fpu
 signal_c/signal_test
 signal_c/signal_test2
+signal_c/pidfd_send_signal
```
lrh2000: Keep the list sorted. This needs to be placed after `signal_c/parent_death_signal`.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,224 @@
...
+	pthread_t thread;
+
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		CHECK(pthread_create(&thread, NULL,
+				     &pidfd_send_signal_self_child_thread,
+				     NULL));
+		CHECK(pthread_join(thread, NULL));
+
+		pause();
```
lrh2000: No need to have this `pause()`?

lrh2000 (commented):
Thank you for quickly revising the code. I appreciate a lot.

I still have two behaviors to confirm. Hopefully, these are the last things to confirm on my end and we can reach the final consensus soon after that.

kernel/src/syscall/pidfd_send_signal.rs
li041: Yeah, the argument `pid` can be thread, process, and even process group id according to its `PIDTYPE_*` while `task_pid(current)` returns thread ID. I misunderstood it. Maybe I should check `ctx.posix_thread.tid()` for both cases.

kernel/src/syscall/pidfd_send_signal.rs
li041: I used the testcase below: 
```
static void *pidfd_send_signal_self_process_group_child_thread(void *arg)
{
	setup_test_siginfo(&siginfo, SIGTERM, -666);

	CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD_GROUP, SIGTERM, &siginfo,
			       PIDFD_SIGNAL_PROCESS_GROUP), _ret == 0);

	return NULL;
}

// PIDFD_SELF_THREAD does not work with PIDFD_SIGNAL_PROCESS_GROUP
// even if the current process is the process group leader.
FN_TEST(pidfd_send_signal_self_process_group_non_main_thread)
{
	pid_t pid;
	int stat;
	pthread_t thread;

	pid = TEST_SUCC(fork());
	if (pid == 0) {
		CHECK(setpgid(0, 0));

		CHECK(pthread_create(&thread, NULL,
				     pidfd_send_signal_self_process_group_child_thread,
				     NULL));
		CHECK(pthread_join(thread, NULL));

		exit(0);
	}

	TEST_RES(waitpid(pid, &stat, 0),
		 WIFSIGNALED(stat) && WTERMSIG(stat) == SIGTERM);
}
END_TEST()
```
The only difference is the `pidfd` argument to use for `pidfd_send_signal`. I got successs for `PIDFD_SELF_THREAD_GROUP` after I set child process to be a process group leader. However, I got `ESRCH` for `PIDFD_SELF_THREAD`. It seems that `tid` is appropriate.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,260 @@
...
+	CHECK_WITH(pidfd_send_signal(PIDFD_SELF_THREAD, SIGTERM, &siginfo,
+				     PIDFD_SIGNAL_PROCESS_GROUP),
+		   _ret == -1 && errno == ESRCH);
+
+	return NULL;
+}
+
+// PIDFD_SELF_THREAD won't work with PIDFD_SIGNAL_PROCESS_GROUP
+// unless the current process is the process group leader i
+// and the current thread is the main thread.
```
lrh2000:
```suggestion
// unless the current process is the process group leader and
// the current thread is the main thread.
```

- Fix the typo "i"?
- Also if we put "and" in line 170, the line is still shorter than line 169. So maybe "and" should be in line 170.

test/initramfs/src/apps/signal_c/pidfd_send_signal.c
```diff
@@ -0,0 +1,260 @@
...
+
+		exit(0);
+	}
+
+	TEST_RES(waitpid(pid, &stat, 0),
+		 WIFSIGNALED(stat) && WTERMSIG(stat) == SIGTERM);
+}
+END_TEST()
+
+static void *pidfd_send_signal_self_process_group_child_thread(void *arg)
```
lrh2000: Remove this `static` as other thread funcs in this file are not `static`.

Either way is fine, just to fix consistency.

tatetian (approved):

PR #2914, opened 2026-01-21
Get PCI bus range from FDT/ACPI and add support for PCI ECAM in x86

lrh2000:
- This first commit resolves this TODO.
   - In x86, we can get the bus range from ACPI MCFG (which will be implemented in the third commit).
   - In RISC-V/LoongArch, we can get the bus range from FDT.
https://github.com/asterinas/asterinas/blob/5792b497224e383feff3e17fbaec619c028e27d6/kernel/comps/pci/src/device_info.rs#L19-L26
 - The second commit protects the PCI PIO ports with a spin lock.
   - I believe it's necessary because bad things will happen if two threads access `0xcf8`/`0cfc` concurrently.
https://github.com/asterinas/asterinas/blob/5792b497224e383feff3e17fbaec619c028e27d6/kernel/comps/pci/src/arch/x86/mod.rs#L57-L58
 - The third commit adds support for parsing PCI MCFG to find out the ECAM base address and the bus range in x86.
```rust
/// A memory region that is stolen for PCI configuration space.
#[derive(Debug)]
pub struct PciEcamRegion {
    /// The base address of the memory region.
    pub base_address: u64,
    /// The start of the bus number.
    pub bus_start: u8,
    /// The end of the bus number.
    pub bus_end: u8,
}
```

### DIFF
--- a/kernel/comps/pci/src/arch/riscv/mod.rs
+++ b/kernel/comps/pci/src/arch/riscv/mod.rs
@@ -13,18 +13,27 @@ use crate::PciDeviceLocation;
 static PCI_ECAM_CFG_SPACE: Once<IoMem> = Once::new();
 
 pub(crate) fn write32(location: &PciDeviceLocation, offset: u32, value: u32) -> Result<(), Error> {
+    if offset > PCI_ECAM_MAX_OFFSET {
+        return Err(Error::InvalidArgs);
+    }
     PCI_ECAM_CFG_SPACE.get().ok_or(Error::IoError)?.write_once(
-        (encode_as_address_offset(location) | (offset & 0xfc)) as usize,
+        (encode_as_address_offset(location) | offset) as usize,
         &value,
     )
 }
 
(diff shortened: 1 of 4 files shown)


kernel/comps/pci/src/arch/riscv/mod.rs
```diff
@@ -13,17 +13,23 @@ use crate::PciDeviceLocation;
 static PCI_ECAM_CFG_SPACE: Once<IoMem> = Once::new();
 
 pub(crate) fn write32(location: &PciDeviceLocation, offset: u32, value: u32) -> Result<(), Error> {
+    if offset > 0xffc {
```
tatetian: Define a const to make this bound more obvious

lrh2000:
Sorry there were some typos. The full force-push diff can be viewed [here](https://github.com/asterinas/asterinas/compare/8539eda0026fd756a29ec09516102700166157db..7c5450e3690d6c1d0897d536aaa6429314596751).

tatetian (approved):

PR #2915, opened 2026-01-21
x86/tdx: integrate fatal error reporting and improve initialization

### LINKED ISSUE #2902: TDX boot failure when `sept-ve-disable` is set to false in QEMU Command
### Describe the bug

The `sept-ve-disable` parameter is an option that controls SEPT (Secure Encrypted Page Table) virtualization exception handling. Currently, Asterinas enforces `sept-ve-disable=true` regardless of the QEMU command configuration. However, when attempting to remove `sept-ve-disable=true` or set `sept-ve-disable=false` in the QEMU command for TDX-enabled boot, the system fails to start.

### To Reproduce

1. Go to `tools/qemu_args.sh`
2. Modify the QEMU command for TDX by either:
- Removing the `"sept-ve-disable": true` entry, or
- Explicitly setting `"sept-ve-disable": false`
3. `make run_kernel INTEL_TDX=1`

### Expected behavior

Asterinas should successfully boot with TDX enabled when `sept-ve-disable` is set to false in the QEMU command, maintaining compatibility with different SEPT configuration options.

### Logs

[EFI stub] Processing 117 memory map entries                                   
[EFI stub] Accepting pending pages                                             
[EFI stub] Entering the Asterinas entry point at 0x8001200                                                                                                     
qemu-system-x86_64: cpus are not resettable, terminating

Hsy-Intel:
Fix #2902

### DIFF
--- a/ostd/src/arch/x86/mod.rs
+++ b/ostd/src/arch/x86/mod.rs
@@ -22,13 +22,30 @@ pub(crate) mod tdx_guest;
 #[cfg(feature = "cvm_guest")]
 pub(crate) fn init_cvm_guest() {
     use ::tdx_guest::{
-        disable_sept_ve, init_tdx, metadata, reduce_unnecessary_ve,
+        SeptVeError, disable_sept_ve, init_tdx, metadata, reduce_unnecessary_ve,
         tdcall::{InitError, write_td_metadata},
+        tdvmcall::report_fatal_error_simple,
     };
     match init_tdx() {
         Ok(td_info) => {
             reduce_unnecessary_ve().unwrap();
-            disable_sept_ve(td_info.attributes).unwrap();
+            match disable_sept_ve(td_info.attributes) {
+                Ok(_) => {}
+                Err(SeptVeError::Misconfiguration) => {
+                    crate::early_println!(
+                        "[kernel] Error: TD misconfiguration: \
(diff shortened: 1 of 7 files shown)


ostd/src/arch/x86/mod.rs
```diff
@@ -22,13 +22,28 @@ pub(crate) mod tdx_guest;
...
         Ok(td_info) => {
             reduce_unnecessary_ve().unwrap();
-            disable_sept_ve(td_info.attributes).unwrap();
+            match disable_sept_ve(td_info.attributes) {
+                Ok(_) => {}
+                Err(SeptVeError::Misconfiguration) => {
+                    crate::early_println!(
+                        "[kernel] Error: TD misconfiguration: SEPT #VE has to be disabled"
+                    );
+                    report_fatal_error_simple("TD misconfiguration: SEPT #VE has to be disabled");
```
tatetian, line 40: This error message is confusing and not helpful. It should explain why TD is misconfigured and how it should be configured.

Hsy-Intel: The `REPORT_FETAL_ERROR` TDVMCALL only support for 64 bytes messages. And this message is consistent with that in Linux. I modified the print messages in kernel.

tatetian (approved):

PR #2920, opened 2026-01-23
Disable BAR decoding when sizing PCI BARs

lrh2000:
According to the PCIe v5.0 specification, we should disable address space decoding during the size calculation of PCI BARs:
> Decode (I/O or memory) of the appropriate address space is disabled via the Command Register before sizing a
Base Address register. [ .. ] The original value in the Base Address register is restored before re-enabling decode in the Command Register of the Function.

But we do the opposite:
https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/kernel/comps/pci/src/common_device.rs#L106-L109

This will cause problem when passing through PCI devices via VFIO:
```
qemu-system-x86_64: kvm_set_user_memory_region: KVM_SET_USER_MEMORY_REGION failed, slot=11, start=0xffffffff00000000, size=0x10000000: Invalid argument
kvm_set_phys_mem: error registering slot: Invalid argument
[1]    5474 IOT instruction  sudo ~/qemu/build/qemu-system-x86_64 -drive  -cpu  -m 8G --no-reboot  -displa
````
in
https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/kernel/comps/pci/src/cfg_space.rs#L359-L368

This was introduced in commit 0370f8fdf3b366e5510c19678645bc97e26f4313 (#2260) without reasonable justification. LoongArch still works after this PR, as tested via the steps listed at https://github.com/asterinas/asterinas/issues/2224#issuecomment-3767169878, so I don't believe this breaks anything. (cc @wyt8)

In addition, thanks for @TinaZhangZW's initial diagnostics on this issue.

### DIFF
--- a/kernel/comps/pci/src/common_device.rs
+++ b/kernel/comps/pci/src/common_device.rs
@@ -103,10 +103,14 @@ impl PciCommonDevice {
             capabilities,
         };
 
-        device.write_command(
-            device.read_command() | Command::MEMORY_SPACE | Command::BUS_MASTER | Command::IO_SPACE,
-        );
+        // While setting up the BARs, we need to ensure that
+        // "Decode (I/O or memory) of the appropriate address space is disabled via the Command
+        // Register before sizing a Base Address register."
+        let command_val = device.read_command() | Command::BUS_MASTER;
+        device.write_command(command_val - (Command::MEMORY_SPACE | Command::IO_SPACE));
         device.bar_manager = BarManager::new(device.header_type.device_type(), location);
+        device.write_command(command_val | (Command::MEMORY_SPACE | Command::IO_SPACE));
+
         device.capabilities = Capability::device_capabilities(&mut device);
 
         Some(device)

### CODE REVIEW

File: kernel/comps/pci/src/common_device.rs [hunk 1]
```diff
@@ -103,10 +103,14 @@ impl PciCommonDevice {
             capabilities,
         };
 
+        // "Decode (I/O or memory) of the appropriate address space is disabled via the Command
+        // Register before sizing a Base Address register."
+        device.set_command(device.command() - (Command::MEMORY_SPACE | Command::IO_SPACE));
```
tatetian: I think all MMIO-based reader methods should be named with `read_xxx` or `load_xxx` to signal the significance of the hardware-related side effect and of course, the cost. A call like `device.command()` reads cheap and immutable, but it is not. Renaming such methods would be helpful to prevent mistakes like this in the future. Let's start from this file:

```rust
impl PciCommonDevice {
    /// Reads the PCI Command
    pub fn read_command(&self) -> Command {
        Command::from_bits_truncate(self.location.read16(PciCommonCfgOffset::Command as u16))
    }

    /// Writes the PCI Command
    pub fn write_command(&self, command: Command) {
        self.location
            .write16(PciCommonCfgOffset::Command as u16, command.bits())
    }

    /// Reads the PCI status
    pub fn read_status(&self) -> Status {
        Status::from_bits_truncate(self.location.read16(PciCommonCfgOffset::Status as u16))
    }    

    pub(super) fn new(location: PciDeviceLocation) -> Option<Self> {
        ...


        // While setting up the BARs, we need to ensure that 
        // "Decode (I/O or memory) of the appropriate address space is disabled via the Command
        // Register before sizing a Base Address register."        
        let base_command_val = device.read_command();
        device.write_command(base_command_val - (Command::MEMORY_SPACE | Command::IO_SPACE));
        device.bar_manager = BarManager::new(device.header_type.device_type(), location);
        device.write_command(base_command_val | (Command::MEMORY_SPACE | Command::IO_SPACE));

        ...
    }
}
```

lrh2000: I agree. In fact, this has already done as a cleanup commit in https://github.com/asterinas/asterinas/pull/2892...

https://github.com/asterinas/asterinas/pull/2892/changes/22500d9ec3d395dc1a8afda7843de91f0774502b#diff-fabf5657775ec052cfc62c30d7a5ca31661199e03175001562869b85c8856c39

lrh2000: There are two cleanup commits in #2892:
 - "[Remove unused code in PCI](https://github.com/asterinas/asterinas/pull/2892/commits/52ae7b85bb49affaeea2fe7177a342377bcc8258)"
 - "[Fix minor problems in PCI](https://github.com/asterinas/asterinas/pull/2892/commits/22500d9ec3d395dc1a8afda7843de91f0774502b)"

If you're happy with them, I can move them to this PR. Then, we can make #2892 a bit smaller.

Otherwise, we can accept this PR as it is. I will rebase #2892 later.

Of course, if really necessary, I can add another commit in this PR to just rename the `command`/`status` methods.

tatetian: Sure. You can move the two commits to this PR.

kernel/comps/pci/src/common_device.rs
```diff
@@ -103,22 +103,19 @@ impl PciCommonDevice {
...
-        device.set_command(
-            device.command() | Command::MEMORY_SPACE | Command::BUS_MASTER | Command::IO_SPACE,
-        );
+        // "Decode (I/O or memory) of the appropriate address space is disabled via the Command
+        // Register before sizing a Base Address register."
+        device.write_command(device.read_command() - (Command::MEMORY_SPACE | Command::IO_SPACE));
         device.bar_manager = BarManager::new(device.header_type.device_type(), location);
+
+        device.write_command(
+            device.read_command()
```
tatetian: I thought reading the command again was unnecessary. Do I miss anything?

lrh2000: Sorry. I was too careless to notice this point in your previous code snippet... I think the new version should resolve this. [diff](https://github.com/asterinas/asterinas/compare/8fb9c7a79e4cd8ccb9face11807f8f65f4940be7..65003354c98bf5896046cb9d88fa141c1c26a7a3)

To make the code symmetric as you wrote in your code snippet, I have to change it so that it enables `BUS_MASTER` before sizing the BARs. I tested this, and it seems to work properly. No relevant sentence in the spec could be found, and the behavior was the same before this PR. Therefore, I believe this is a valid change.
```rust
        let command_val = device.read_command() | Command::BUS_MASTER;
        device.write_command(command_val - (Command::MEMORY_SPACE | Command::IO_SPACE));
        device.bar_manager = BarManager::new(device.header_type.device_type(), location);
        device.write_command(command_val | (Command::MEMORY_SPACE | Command::IO_SPACE));
```

PR #2511, opened 2025-10-17
Correct the errno of waitid when the input pidfd is negative

cchanging:
While debugging systemd on the latest main branch, I discovered that our current implementation of the `waitid` syscall returns an inconsistent errno compared to Linux when handling negative pidfd values. 

Our existing implementation returns `EBADF` in this case because it cannot find the file, but on Linux, this situation returns `EINVAL`. This inconsistency is causing some operational failures in systemd. 

I tried some other syscalls like `read`. They will return `EBADF` if the fd is negative. I'm not sure if there are other syscalls similar to waitid in that regard, but I haven't found any ones yet.

PR #2922, opened 2026-01-23
Align `PidFile` semantics with `POSIX` standards

li041:
## Changes in this PR
- Fix `pseudo_mount.c` and `pidfd.c` testcases to pass under linux.
- Fix `pidfd` part in `pseudo_inode.c` to pass under linux by adding `PidfdFs` and `PseudoInodeType`
- Fix `epoll` for `PidFile` to fit description in [man7](https://man7.org/linux/man-pages/man2/pidfd_open.2.html).
>  When the task that it refers to terminates and becomes a zombie, these interfaces indicate the file descriptor as readable (EPOLLIN).  When the task is reaped, these interfaces produce a hangup event (EPOLLHUP).
- When read or write at pidfd file with offset, return EINVAL instead of ESPIPE.

### DIFF
--- a/kernel/src/fs/pseudofs.rs
+++ b/kernel/src/fs/pseudofs.rs
@@ -71,7 +71,7 @@ impl PseudoFs {
                 sb: SuperBlock::new(magic, aster_block::BLOCK_SIZE, NAME_MAX),
                 root: Arc::new(PseudoInode::new(
                     ROOT_INO,
-                    InodeType::Dir,
+                    PseudoInodeType::Root,
                     mkmod!(u+rw),
                     Uid::new_root(),
                     Gid::new_root(),
@@ -85,7 +85,7 @@ impl PseudoFs {
 
     pub fn alloc_inode(
         self: &Arc<Self>,
-        type_: InodeType,
+        type_: PseudoInodeType,
         mode: InodeMode,
         uid: Uid,
         gid: Gid,
@@ -140,7 +140,7 @@ impl SockFs {
     /// Creates a pseudo `Path` for a socket.
     pub fn new_path() -> Path {
         let socket_inode = Arc::new(Self::singleton().alloc_inode(
-            InodeType::Socket,
+            PseudoInodeType::Socket,
             mkmod!(a+rwx),
             Uid::new_root(),
             Gid::new_root(),
@@ -206,13 +206,58 @@ impl AnonInodeFs {
 
         SHARED_INODE.call_once(|| {
             let shared_inode = Self::singleton().alloc_inode(
-                InodeType::Unknown,
+                PseudoInodeType::AnonInode,
                 mkmod!(u+rw),
                 Uid::new_root(),
                 Gid::new_root(),
             );
 
             Arc::new(shared_inode)
+        })
+    }
+}
+
+pub struct PidfdFs {
+    _private: (),
+}
+
+impl PidfdFs {
+    /// Returns the singleton instance of the pidfd file system.
+    pub fn singleton() -> &'static Arc<PseudoFs> {
+        static PIDFDFS: Once<Arc<PseudoFs>> = Once::new();
+
+        PseudoFs::singleton(&PIDFDFS, "pidfdfs", PIDFDFS_MAGIC)
+    }
+
+    /// Creates a pseudo `Path` for a pidfd.
+    pub fn new_path(name_fn: fn(&dyn Inode) -> String) -> Path {
+        Path::new_pseudo(
+            Self::mount_node().clone(),
+            Self::shared_inode().clone(),
+            name_fn,
+        )
+    }
+
+    /// Returns the pseudo mount node of the pidfd file system.
+    pub fn mount_node() -> &'static Arc<Mount> {
+        static PIDFDFS_MOUNT: Once<Arc<Mount>> = Once::new();
+
+        PIDFDFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()))
+    }
+
+    /// Returns the shared inode of the pidfd file system.
+    pub fn shared_inode() -> &'static Arc<dyn Inode> {
+        static SHARED_INODE: Once<Arc<dyn Inode>> = Once::new();
+
+        SHARED_INODE.call_once(|| {
+            let pidfd_inode = Self::singleton().alloc_inode(
+                PseudoInodeType::Pidfd,
+                mkmod!(u+rwx),
+                Uid::new_root(),
+                Gid::new_root(),
+            );
+
+            Arc::new(pidfd_inode)
         })
     }
 }
@@ -282,24 +327,50 @@ const PIPEFS_MAGIC: u64 = 0x50495045;
 const SOCKFS_MAGIC: u64 = 0x534F434B;
 // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/include/uapi/linux/magic.h#L93>
 const ANON_INODEFS_MAGIC: u64 = 0x09041934;
+// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/include/uapi/linux/magic.h#L105>
+const PIDFDFS_MAGIC: u64 = 0x50494446;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PseudoInodeType {
+    Root,
+    Pipe,
+    Socket,
+    AnonInode,
+    Pidfd,
+}
+
+impl From<PseudoInodeType> for InodeType {
+    fn from(pseudo_type: PseudoInodeType) -> Self {
+        match pseudo_type {
+            PseudoInodeType::Root => InodeType::Dir,
+            PseudoInodeType::Pipe => InodeType::NamedPipe,
+            PseudoInodeType::Socket => InodeType::Socket,
+            PseudoInodeType::AnonInode => InodeType::Unknown,
+            PseudoInodeType::Pidfd => InodeType::Unknown,
+        }
+    }
+}
 
 /// A pseudo inode that does not correspond to any real path in the file system.
 pub struct PseudoInode {
     metadata: SpinLock<Metadata>,
     extension: Extension,
     fs: Weak<PseudoFs>,
+    is_anon: bool,
 }
 
 impl PseudoInode {
     fn new(
         ino: u64,
-        type_: InodeType,
+        type_: PseudoInodeType,
         mode: InodeMode,
         uid: Uid,
         gid: Gid,
         fs: Weak<PseudoFs>,
     ) -> Self {
         let now = now();
+        let type_ = InodeType::from(type_);
+
         let metadata = Metadata {
             dev: 0,
             ino,
@@ -316,10 +387,12 @@ impl PseudoInode {
             gid,
             rdev: 0,
         };
+
         PseudoInode {
             metadata: SpinLock::new(metadata),
             extension: Extension::new(),
             fs,
+            is_anon: type_ == InodeType::Unknown,
         }
     }
 }
@@ -380,6 +453,13 @@ impl Inode for PseudoInode {
     }
 
     fn set_mode(&self, mode: InodeMode) -> Result<()> {
+        if self.is_anon {
+            return_errno_with_message!(
+                Errno::EOPNOTSUPP,
+                "the mode of anonymous inodes cannot be changed"
+            );
+        }
+
         let mut meta = self.metadata.lock();
         meta.mode = mode;
         meta.ctime = now();
--- a/kernel/src/process/pid_file.rs
+++ b/kernel/src/process/pid_file.rs
@@ -11,7 +11,7 @@ use crate::{
         file_handle::FileLike,
         file_table::FdFlags,
         path::Path,
-        pseudofs::AnonInodeFs,
+        pseudofs::PidfdFs,
         utils::{CreationFlags, StatusFlags},
     },
     prelude::*,
@@ -43,7 +43,7 @@ impl Debug for PidFile {
 
 impl PidFile {
     pub fn new(process: Arc<Process>, is_nonblocking: bool) -> Self {
-        let pseudo_path = AnonInodeFs::new_path(|_| "anon_inode:[pidfd]".to_string());
+        let pseudo_path = PidfdFs::new_path(|_| "anon_inode:[pidfd]".to_string());
 
         Self {
             process: Arc::downgrade(&process),
@@ -59,7 +59,7 @@ impl PidFile {
         // Reference: <https://man7.org/linux/man-pages/man2/pidfd_open.2.html>.
         let Some(process) = self.process.upgrade() else {
             // The process has been reaped.
-            return IoEvents::IN;
+            return IoEvents::IN | IoEvents::HUP;
         };
         if process.status().is_zombie() {
             IoEvents::IN
@@ -84,6 +84,20 @@ impl FileLike for PidFile {
 
     fn write(&self, _reader: &mut VmReader) -> Result<usize> {
         return_errno_with_message!(Errno::EINVAL, "PID file cannot be written");
+    }
+
+    fn read_at(&self, _offset: usize, _writer: &mut VmWriter) -> Result<usize> {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "PID file cannot be read at a specific offset"
+        );
+    }
+
+    fn write_at(&self, _offset: usize, _reader: &mut VmReader) -> Result<usize> {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "PID file cannot be written at a specific offset"
+        );
     }
 
     fn set_status_flags(&self, new_flags: StatusFlags) -> Result<()> {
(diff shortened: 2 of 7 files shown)


li041:
@boterinas rerun

lrh2000:
> * Fix `pseudo_mount.c` and `pidfd.c` testcases to pass under linux.
> * Fix `epoll` for `PidFile` to fit description in [man7](https://man7.org/linux/man-pages/man2/pidfd_open.2.html).

FYI, I assume that most regression test fixes are due to Linux version issues. For example, the EPOLLHUP issue was previously noticed [here](https://github.com/asterinas/asterinas/pull/2511#:~:text=I%20tried%20this%20patch%20in%20Linux).

Of course, updating our behavior to align with the latest Linux version is still a good thing.

> * Besides, linux has this virtual fs.

The virtual FS for pidfds is also [a relatively new feature that only exists in recent Linux versions](https://lwn.net/Articles/963749/). I'm not even sure if the problem that Linux is trying to solve exists in our implementation.

The need for a separate `PidfdInode` type definition should be justified. It seems to contain a lot of duplicate code with the `PseudoInode`. Currently, the only real reason I can see is:
> * return `EPERM` when chmod at pid file

However, Linux 6.18 seems to give (running as the root user):
```
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` passed [got Success]
test_anon_inodefs_share_inode: `set_mode(fds[i].fd, fds[i].modes[1])` failed [got Operation not supported, but expected Success]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[1]` is false]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[0]` is false]
test_anon_inodefs_share_inode: `set_mode(fds[i].fd, fds[i].modes[1])` failed [got Operation not supported, but expected Success]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[1]` is false]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[0]` is false]
test_anon_inodefs_share_inode: `set_mode(fds[i].fd, fds[i].modes[1])` failed [got Operation not supported, but expected Success]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[1]` is false]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[0]` is false]
test_anon_inodefs_share_inode: `set_mode(fds[i].fd, fds[i].modes[1])` failed [got Operation not supported, but expected Success]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[1]` is false]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[0]` is false]
test_anon_inodefs_share_inode: `set_mode(fds[i].fd, fds[i].modes[1])` failed [got Operation not supported, but expected Success]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[1]` is false]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[0]` is false]
test_anon_inodefs_share_inode: `set_mode(fds[i].fd, fds[i].modes[1])` failed [got Operation not supported, but expected Success]
test_anon_inodefs_share_inode: `get_mode(fds[i].fd)` failed [got Success, but `_ret == fds[i].modes[1]` is false]
test_anon_inodefs_share_inode summary: 1 tests passed, 17 tests failed
```

So this looks like a Linux version issue again? Any input from @vvvvsv?

vvvvsv:
Thanks for tracking the latest Linux behavior!

> So this looks like a Linux version issue again?

Yes. In Linux 6.18, these anonymous files still share the same inode, but `chmod` and `chown` are rejected now. See this [patch](https://github.com/torvalds/linux/commit/22bdf3d6581af6d06ed8a46c6835648421cca0ea).

To fix this, I suggest:
1. Add `struct AnonInode(PseudoInode)` and inherit all methods except for `set_*` from `PseudoInode` to `AnonInode`. The `set_*` methods of `AnonInode` should return `EPERM`.
2. For `pidfd`s, if , if a `PidfdFs` is needed, we can link an `AnonInode` to `PidfdFs` when creating a `pidfd`.

---

By the way, should we add a CI that runs all tests on the latest Linux kernel?

lrh2000:
> 1. Add `struct AnonInode(PseudoInode)` and inherit all methods except for `set_*` from `PseudoInode` to `AnonInode`. The `set_*` methods of `AnonInode` should return `EPERM`.

Indeed, I was thinking that we could add a flag (a boolean field) to the `PseudoInode` to distinguish the fact. I think it's simpler because I already felt that there were too many methods in `Inode`, and even inheriting them is annoying because the list of methods is long.

> By the way, should we add a CI that runs all tests on the latest Linux kernel?

I'm not sure if all of our core developers have access to the latest Linux kernel. It's not a problem for me because I use ArchLinux, but I guess others may use older versions, so I don't know if this could be a problem.

vvvvsv:
> there were too many methods in `Inode`, and even inheriting them is annoying because the list of methods is long.

Agree.

> we could add a flag (a boolean field) to the `PseudoInode` to distinguish the fact

If we are to introduce a new flag, I think it would be better to use an enum flag. It is clearer than a `boolean`, and would also simplify things. For example, we will not need to specify both the `InodeType` and `InodeMode` in `PseudoInode::new`.

```rust
enum PseudoInodeType {
    Pipe,
    Socket,
    Anon, 
}
```

> I'm not sure if all of our core developers have access to the latest Linux kernel.

We can provide a VM with the latest Linux kernel? Just like [bench_linux_and_aster.sh](https://github.com/asterinas/asterinas/blob/main/test/initramfs/src/benchmark/bench_linux_and_aster.sh).

lrh2000:
> If we are to introduce a new flag, I think it would be better to use an enum flag. It is clearer than a `boolean`, and would also simplify things. For example, we will not need to specify both the `InodeType` and `InodeMode` in `PseudoInode::new`.

I'm not sure if we should distinguish them in the future. If so, using an enum would be better. Do you have a specific use case in mind?

For now, adding two constructors `PseudoInode::new` and `PseudoInode::new_anon` with an internal flag called `is_anon` also seems acceptable.

> We can provide a VM with the latest Linux kernel? Just like [bench_linux_and_aster.sh](https://github.com/asterinas/asterinas/blob/main/test/initramfs/src/benchmark/bench_linux_and_aster.sh).

That sounds like a good idea to me!

li041:
In 6.11.0-rc3, pidfds have no longer shared the same anon_inode implementation, causing chmod operations on pidfd to return `EPERM` instead of success. And it seems that anon_inode return `EOPNOSUPPORT` in the latest Linux. 

`PseduInode` in Asterinas implements the same semantics as Linux's anon_inode. Should I modify `set_mode` method for PseduInode  to return `EOPNOSUPPORT`? And if a flag is added to distingsuish between anon_inode and pidfd inode. Maybe branchs may exist at least  in `set_mode` and `get_mode` function.

What's more, files may share the same anon_inode with different inode mode.

lrh2000:
The concept of a `PseduInode` is larger than that of a Linux `anno_inode`. It is also used in `sockfs` and `pipefs`. Anonymous inodes are a specific type of pseudo-inode, as are pipefs and sockfs inodes.

> causing chmod operations on pidfd to return `EPERM` instead of success

I think this is incorrect. [It's `EOPNOTSUPP`](https://lore.kernel.org/all/20250407-work-anon_inode-v1-4-53a44c20d44e@kernel.org/), same to anonymous inode.

> if a flag is added to distingsuish between anon_inode and pidfd inode

We need to distinguish between (anonymous inodes & pidfs inodes) and (sockfs & pipefs inodes). The first class forbids `set_mode` with `EOPNOTSUPP`, but sockfs and pipefs inodes do not seem to be affected.

li041:
> I think this is incorrect. [It's `EOPNOTSUPP`](https://lore.kernel.org/all/20250407-work-anon_inode-v1-4-53a44c20d44e@kernel.org/), same to anonymous inode.

okk, I see. Pidfd use the generic version after this patch. How about `get_mode`? In 6.11-rc.3, epoll file returns 0400 while pidfd returns 0700.

lrh2000:
> okk, I see. Pidfd use the generic version after this patch. How about `get_mode`? In 6.11-rc.3, epoll file returns 0400 while pidfd returns 0700.

@vvvvsv's approach of introducing a `PseudoInodeType` enum may solve this problem?

```rust
impl PseudoInode {
   fn new(type_: PseudoInodeType) -> Self {
       let mode = match type_ {
           PseudoInodeType::Pidfd => mkmod!(...),
           // ...
       }
   }
}
```

To be honest, I don't fully like this approach because I feel the mode belongs in a wrong file. For example, the pidfd mode should be near the pidfd file definition, not in `pseudo_inode.rs`. For now, however, I have failed to come up with a better idea. Is there any better solution?

li041:
This may not work for pidfd. `AnonInode` init once, and share the same inode.
```
pub fn shared_inode() -> &'static Arc<dyn Inode> {
    static SHARED_INODE: Once<Arc<dyn Inode>> = Once::new();
    SHARED_INODE.call_once(|| {
	    ...
    })
}
```

lrh2000:
> This may not work for pidfd. `AnonInode` init once, and share the same inode.

Don't you plan to build a `PidfdFs` in this PR? `PidfdFs::shared_inode()` will certainly return a different inode than `AnonInodeFs::shared_inode()`, won't it?

li041:
I feel sorry that I misunderstood previous commits. I thought introducing `PidfdFs` was also unnecessary, so I'm confused about how to resolve related issues. Then `is_anon` is enough to solve this problem, no need to use `enum` for now(maybe `enum` for future scalibility).

kernel/src/process/pid_file.rs
```diff
@@ -140,7 +148,7 @@ impl Pollable for PidFile {
     fn poll(&self, mask: IoEvents, poller: Option<&mut PollHandle>) -> IoEvents {
         let Some(process) = self.process.upgrade() else {
             // The process has been reaped.
-            return mask & IoEvents::IN;
+            return mask & (IoEvents::IN | IoEvents::HUP);
```
lrh2000, line 157: This means that we need to call `pollee.notify(IoEvents::HUP)` when the strong reference disappears. Otherwise, we cannot wake up programs waiting for HUP events.

The only correct place to do this is in a `Drop` impl. Other locations, such as in `wait`, will likely incur race conditions.

kernel/src/process/pid_file.rs
```diff
@@ -86,6 +86,14 @@ impl FileLike for PidFile {
         return_errno_with_message!(Errno::EINVAL, "PID file cannot be written");
     }
 
+    fn read_at(&self, _offset: usize, _writer: &mut VmWriter) -> Result<usize> {
+        return_errno_with_message!(Errno::EINVAL, "read_at is not supported");
+    }
+
+    fn write_at(&self, _offset: usize, _reader: &mut VmReader) -> Result<usize> {
+        return_errno_with_message!(Errno::EINVAL, "write_at is not supported");
+    }
```
lrh2000, line 101: Use the format near it:
```rust
        return_errno_with_message!(Errno::EINVAL, "PID file cannot be read at a specific offset");
```

kernel/src/fs/pseudofs.rs
```diff
@@ -380,6 +441,10 @@ impl Inode for PseudoInode {
     }
 
     fn set_mode(&self, mode: InodeMode) -> Result<()> {
+        if self.is_anon {
+            return_errno_with_message!(Errno::EOPNOTSUPP, "anon inode's mode can not be changed");
```
lrh2000:
```suggestion
            return_errno_with_message!(Errno::EOPNOTSUPP, "the mode of anonymous inodes cannot be changed");
```

kernel/src/fs/pseudofs.rs
```diff
@@ -125,6 +135,50 @@ impl PipeFs {
     }
 }
 
+pub struct PidfdFs {
```
lrh2000: Move this below `AnonInodeFs`. This file system is more similar to an `AnonInodeFs` than to a `SockFs` or `PipeFs`.

kernel/src/fs/pseudofs.rs
```diff
@@ -125,6 +135,50 @@ impl PipeFs {
...
+}
+
+impl PidfdFs {
+    /// Returns the singleton instance of the anonymous pidfd file system.
+    pub fn singleton() -> &'static Arc<PseudoFs> {
+        static PIDFDFS: Once<Arc<PseudoFs>> = Once::new();
+
+        PseudoFs::singleton(&PIDFDFS, "pidfdfs", PIDFDFS_MAGIC)
+    }
+    /// Creates a pseudo `Path` for an anonymous pidfd.
```
lrh2000: - Insert a blank line between these two methods.
 - "the anonymous pidfd file system" -> "the pidfd file system"
 - "an anonymous pidfd" -> "a pidfd"

I don't think there is a non-anonymous (named?) pidfd... The comment seems to be copied from `PipeFs`, but we do have named pipes there.

kernel/src/fs/pseudofs.rs
```diff
@@ -125,6 +135,50 @@ impl PipeFs {
...
+    }
+
+    /// Returns the pseudo mount node of the pidfd file system.
+    pub fn mount_node() -> &'static Arc<Mount> {
+        static PIDFDFS_MOUNT: Once<Arc<Mount>> = Once::new();
+
+        PIDFDFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()))
+    }
+
+    pub fn shared_inode() -> &'static Arc<dyn Inode> {
```
lrh2000:
```rust
    /// Returns the pseudo mount node of the pidfd file system.
```

kernel/src/fs/pseudofs.rs
```diff
@@ -144,6 +198,7 @@ impl SockFs {
             mkmod!(a+rwx),
             Uid::new_root(),
             Gid::new_root(),
+            false,
```
lrh2000: I found this argument is completely determined by `InodeType`. So now I think we do want to have `PseudoInodeType`, but in this way:
```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq)
enum PseudoInodeType {
    Pipe,
    Socket,
    Anon,
}

impl From<PseudoInodeType> for InodeType { /* ... */ }

impl PeudoInode {
    fn new(type_: PseudoInodeType) -> Self {
        let metadata = Metadata {
            type_: type_.into(),
        };
        Self {
            is_anon: type_ == PseudoInodeType::Anon,
        }
    }
}
```

li041: Maybe the root inode of `PseudoFs` is an exception. It is `InodeType::Dir`. Is it okk to add `Dir` to PseudoInodeType?
```
    /// Returns a reference to the singleton pseudo file system.
    fn singleton(
        ...
    ) -> &'static Arc<Self> {
        // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/fs/libfs.c#L659-L689>
        fs.call_once(|| {
            Arc::new_cyclic(|weak_fs: &Weak<Self>| Self {
                ...
                InodeType::Dir,
                ...
```

lrh2000: > Is it okk to add `Dir` to PseudoInodeType?

I think it makes sense to add a `Root` variant to it.

tatetian:
> By the way, should we add a CI that runs all tests on the latest Linux kernel?

This is desirable. We should run all `app/` tests under Linux. Maybe not the latest one, because the "latest" is a moving target. We can target a relatively-recent Linux version, advertise this version in the doc, and enforce the version by CI tests.

kernel/src/process/pid_file.rs
lrh2000, line 161: `check_io_events` needs to be updated as well for `IoEvents::HUP`.

kernel/src/fs/pseudofs.rs
```diff
@@ -282,24 +327,50 @@ const PIPEFS_MAGIC: u64 = 0x50495045;
...
+// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/include/uapi/linux/magic.h#L105>
+const PIDFDFS_MAGIC: u64 = 0x50494446;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PseudoInodeType {
+    Root,
+    Pipe,
+    Socket,
+    Pidfd,
+    AnonInode,
```
lrh2000, line 338:
```rust
    AnonInode,
    Pidfd,
```

Reorder to match the FS definition and the FS magic definition.

kernel/src/fs/pseudofs.rs
```diff
@@ -282,24 +327,50 @@ const PIPEFS_MAGIC: u64 = 0x50495045;
...
+}
+
+impl From<PseudoInodeType> for InodeType {
+    fn from(pseudo_type: PseudoInodeType) -> Self {
+        match pseudo_type {
+            PseudoInodeType::Root => InodeType::Dir,
+            PseudoInodeType::Pipe => InodeType::NamedPipe,
+            PseudoInodeType::Socket => InodeType::Socket,
+            PseudoInodeType::Pidfd => InodeType::Unknown,
+            PseudoInodeType::AnonInode => InodeType::Unknown,
```
lrh2000, line 348: Same.

kernel/src/fs/pseudofs.rs
```diff
@@ -380,6 +453,13 @@ impl Inode for PseudoInode {
     }
 
     fn set_mode(&self, mode: InodeMode) -> Result<()> {
+        if self.is_anon {
+            return_errno_with_message!(
+                Errno::EOPNOTSUPP,
+                "the mode of anonymous inode cannot be changed"
```
lrh2000:
```suggestion
                "the mode of anonymous inodes cannot be changed"
```

cqs21 (approved):

PR #2926, opened 2026-01-25
Fix permission validation in mprotect and mmap

Marsman1996:
Add permission check in `mprotect` according to https://github.com/asterinas/asterinas/pull/2918#discussion_r2721470506

Refer:
https://github.com/torvalds/linux/blob/d91a46d6805af41e7f2286e0fc22d498f45a682b/include/linux/mman.h#L106-L109
```C
static inline bool arch_validate_prot(unsigned long prot, unsigned long addr)
{
	return (prot & ~(PROT_READ | PROT_WRITE | PROT_EXEC | PROT_SEM)) == 0;
}
```
https://github.com/torvalds/linux/blob/d91a46d6805af41e7f2286e0fc22d498f45a682b/mm/mprotect.c#L827-L828
```C
	if (!arch_validate_prot(prot, start))
		return -EINVAL;
```

### DIFF
--- a/kernel/src/vm/perms.rs
+++ b/kernel/src/vm/perms.rs
@@ -42,6 +42,24 @@ impl VmPerms {
 
         Ok(())
     }
+
+    /// Parses `bits` as requested permissions from user programs and returns errors
+    /// if there are unknown permissions.
+    pub fn from_user_bits(bits: u32) -> Result<Self> {
+        if let Some(vm_perms) = VmPerms::from_bits(bits)
+            && Self::ALL_PERMS.contains(vm_perms)
+        {
+            Ok(vm_perms)
+        } else {
+            return_errno_with_message!(Errno::EINVAL, "invalid permissions");
+        }
+    }
+
+    /// Parses `bits` as requested permissions from user programs and ignores any
+    /// unknown permissions.
+    pub fn from_user_bits_truncate(bits: u32) -> Self {
+        VmPerms::from_bits_truncate(bits) & Self::ALL_PERMS
+    }
 }
 
 impl From<Rights> for VmPerms {
--- a/test/initramfs/src/apps/mmap/mmap_and_mprotect.c
+++ b/test/initramfs/src/apps/mmap/mmap_and_mprotect.c
@@ -22,10 +22,10 @@ FN_TEST(mprotect_shared_writable_mapping_on_read_only_file)
 	fd = TEST_SUCC(open(filename, O_RDONLY));
 
 	char *addr =
-		CHECK_WITH(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_SHARED, fd, 0),
-			   _ret != MAP_FAILED);
+		TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_SHARED, fd, 0));
 	TEST_ERRNO(mprotect(addr, PAGE_SIZE, PROT_READ | PROT_WRITE), EACCES);
 
+	TEST_SUCC(munmap(addr, PAGE_SIZE));
 	TEST_SUCC(close(fd));
 	TEST_SUCC(unlink(filename));
 }
@@ -38,11 +38,9 @@ FN_TEST(mprotect_private_writable_mapping_copy_on_write)
 	TEST_SUCC(write(fd, "AAAA", 5));
 
 	char *addr1 =
-		CHECK_WITH(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_PRIVATE, fd, 0),
-			   _ret != MAP_FAILED);
+		TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_PRIVATE, fd, 0));
 	char *addr2 =
-		CHECK_WITH(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_PRIVATE, fd, 0),
-			   _ret != MAP_FAILED);
+		TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ, MAP_PRIVATE, fd, 0));
 	TEST_RES(strcmp(addr1, "AAAA"), _ret == 0);
 	TEST_RES(strcmp(addr2, "AAAA"), _ret == 0);
 	TEST_SUCC(mprotect(addr1, PAGE_SIZE, PROT_READ | PROT_WRITE));
@@ -50,7 +48,24 @@ FN_TEST(mprotect_private_writable_mapping_copy_on_write)
 	TEST_RES(strcmp(addr1, "BBBB"), _ret == 0);
 	TEST_RES(strcmp(addr2, "AAAA"), _ret == 0);
 
+	TEST_SUCC(munmap(addr1, PAGE_SIZE));
+	TEST_SUCC(munmap(addr2, PAGE_SIZE));
 	TEST_SUCC(close(fd));
 	TEST_SUCC(unlink(filename));
+}
+END_TEST()
+
+FN_TEST(mprotect_invalid_permission)
+{
+#define PROT_INVALID 0x20
+
+	void *addr = TEST_SUCC(mmap(NULL, PAGE_SIZE,
+				    PROT_READ | PROT_WRITE | PROT_INVALID,
+				    MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+
+	TEST_SUCC(mprotect(addr, PAGE_SIZE, PROT_READ | PROT_EXEC));
+	TEST_ERRNO(mprotect(addr, PAGE_SIZE, PROT_READ | PROT_INVALID), EINVAL);
+
+	TEST_SUCC(munmap(addr, PAGE_SIZE));
 }
 END_TEST()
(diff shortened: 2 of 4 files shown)


lrh2000:
```
kernel/src/syscall/mmap.rs:    let perms = VmPerms::from_bits_truncate(perms as u32);
kernel/src/syscall/mprotect.rs:    let vm_perms = VmPerms::from_bits_truncate(perms as u32);
```

This two need to be fixed together. Let's add a new `VmPerms::from_user_bits` method to perform the check.

---

Also, use `from_bits` instead of `from_bits_truncated`, such as:
```rust
let vm_perms = if let Some(vm_perms) = VmPerms::from_bits(...) && ALL_PERMS.contain(vm_perms) {
   vm_perms
} else {
   return_errno_and_error_message!(...);
}
```

---

In addition, please add some regression tests for this, like the one in https://github.com/asterinas/asterinas/pull/2918 for both `mmap` and `mprotect` (but in `mmap_err.c`).

Since you're submitting a large number of PRs to fix various issues in our codebase (thanks for the efforts), maybe it's time to start writing some regression tests. It's acceptable to omit some tests if they're difficult to write and the fix is obvious, but when feasible and appropriate, let's write them.
 - I added some regression tests for the recent ones that were missed. See #2927. That can serve as an example to get started.

Marsman1996:
> ```
> kernel/src/syscall/mmap.rs:    let perms = VmPerms::from_bits_truncate(perms as u32);
> kernel/src/syscall/mprotect.rs:    let vm_perms = VmPerms::from_bits_truncate(perms as u32);
> ```
> 
> This two need to be fixed together. Let's add a new `VmPerms::from_user_bits` method to perform the check.

But in Linux, `mmap` won't validate the `perms` like `mprotect` (except for powerpc arch). Shall we keep consist as Linux?

lrh2000:
> But in Linux, `mmap` won't validate the `perms` like `mprotect` (except for powerpc arch). Shall we keep consist as Linux?

OK I don't know that. But perhaps we need a `VmPerms::from_user_bits_truncate` to ignore such bits. Are you sure our current implementation correctly ignores the `MAY_*` flags? After quickly skimming the code, I found that we may return an error here.

https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/kernel/src/vm/vmar/vmar_impls/map.rs#L372-L376

The code is far away from the system call entry so I'm not 100% sure about it. So having a regression test to confirm is always a good idea, no matter whether our current implementation is correct or not.

---

As a side note, I've continuously been wondering if we should create separate types for the `MAY_*` permissions and the normal permissions. For now, I'm treating it as a separate issue because it may require more effort in the design, change the MM code, and conflict with ongoing refactorings.

Marsman1996:
> Are you sure our current implementation correctly ignores the `MAY_*` flags?

No, Asterinas keeps the `MAY_*` flags in `perms`.

https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/kernel/src/syscall/mmap.rs#L25

But currently Asterinas maintain a seperate `may_perms`, and cannot directly modified by the user.

lrh2000:
> But currently Asterinas maintain a seperate `may_perms`, and cannot directly modified by the user.

This has nothing to do with my previous reply. You need to re-read the previous reply more carefully and check if the aforementioned code snippet can be triggered by user input.

Marsman1996:
> > But currently Asterinas maintain a seperate `may_perms`, and cannot directly modified by the user.
> 
> This has nothing to do with my previous reply. You need to re-read the previous reply more carefully and check if the aforementioned code snippet can be triggered by user input.

It can be triggered by the user input through the call chain:
`sys_mmap` -> `do_sys_mmap` -> `VmarMapOptions::build` -> `VmarMapOptions::check_options` -> `VmarMapOptions::check_perms`

~But it does not relate to the `MAY_*` filter. `self.perms` could still contain the `MAY_*` flags here.~

Marsman1996:
`mmap(NULL, len, PROT_READ | PROT_WRITE | 0x10 | 0x20, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);`, the `self.perms` here is `READ | WRITE | MAY_WRITE | MAY_EXEC`, and will return a `Permission denied`, which is inconsist with Linux.

lrh2000:
> `mmap(NULL, len, PROT_READ | PROT_WRITE | 0x10 | 0x20, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);`, the `self.perms` here is `READ | WRITE | MAY_WRITE | MAY_EXEC`, and will return a `Permission denied`, which is inconsist with Linux.

Yes. Thanks for the confirmation. That's exactly my point.

kernel/src/vm/perms.rs
```diff
@@ -42,6 +42,20 @@ impl VmPerms {
 
         Ok(())
     }
+
+    pub fn from_user_bits(bits: u32) -> Result<Self> {
```
lrh2000, line 48:
```rust
    /// Parses `bits` as requested permissions from user programs and returns errors
    /// if there are unknown permissions.
```

kernel/src/vm/perms.rs
```diff
@@ -42,6 +42,20 @@ impl VmPerms {
...
+        if let Some(vm_perms) = VmPerms::from_bits(bits)
+            && Self::ALL_PERMS.contains(vm_perms)
+        {
+            Ok(vm_perms)
+        } else {
+            return_errno_with_message!(Errno::EINVAL, "invalid permissions");
+        }
+    }
+
+    pub fn from_user_bits_truncate(bits: u32) -> Self {
```
lrh2000, line 60:
```rust
    /// Parses `bits` as requested permissions from user programs and ignores any
    /// unknown permissions.
```

test/initramfs/src/apps/mmap/mmap_and_mprotect.c
```diff
@@ -54,3 +54,13 @@ FN_TEST(mprotect_private_writable_mapping_copy_on_write)
...
 }
 END_TEST()
+
+FN_TEST(mprotect_permission)
+{
+	void *addr =
+		TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE | 0x20,
+			       MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+	TEST_SUCC(mprotect(addr, PAGE_SIZE, PROT_READ | PROT_EXEC));
+	TEST_ERRNO(mprotect(addr, PAGE_SIZE, PROT_READ | 0x20), EINVAL);
```
lrh2000:
```c
FN_TEST(mprotect_invalid_permission)
{
#define PROT_INVALID 0x20

	void *addr =
		TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE | PROT_INVALID,
			       MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));

	TEST_SUCC(mprotect(addr, PAGE_SIZE, PROT_READ | PROT_EXEC));
	TEST_ERRNO(mprotect(addr, PAGE_SIZE, PROT_READ | PROT_INVALID), EINVAL);
}
END_TEST()
```

 - `mprotect_permission` -> `mprotect_invalid_permission`
 - Define a `PROT_INVALID` for clarity.
 - Add blank lines in between.

test/initramfs/src/apps/mmap/mmap_and_mprotect.c
```diff
@@ -54,3 +54,13 @@ FN_TEST(mprotect_private_writable_mapping_copy_on_write)
 	TEST_SUCC(unlink(filename));
```
lrh2000, line 54: Also if you want, do a few more minor cleanups in this file together:
 - Replace `CHECK(mmap(...), ret != MAP_FAILED)` with `TEST_SUCC(mmap(...))`.
   - The latter is generally preferred because it is shorter and non-fatal. It also prints the test command to the screen. Long ago, `mmap()` did not work in `TEST_SUCC()` because it returned a pointer. However, that problem has been resolved.
 - Add the cleanup `munmap()`s. (Also in the newly added test `mprotect_permission`):
```c
TEST_SUCC(munmap(addr1, PAGE_SIZE));
TEST_SUCC(munmap(addr2, PAGE_SIZE));
```

lrh2000 (commented):
Thank you for fixing this important security vulnerability!

test/initramfs/src/apps/mmap/mmap_and_mprotect.c
```diff
@@ -38,19 +38,35 @@ FN_TEST(mprotect_private_writable_mapping_copy_on_write)
...
+	TEST_SUCC(munmap(addr1, PAGE_SIZE));
+	TEST_SUCC(munmap(addr2, PAGE_SIZE));
 	TEST_SUCC(close(fd));
 	TEST_SUCC(unlink(filename));
 }
 END_TEST()
+
+FN_TEST(mprotect_invalid_permission)
+{
+#define PROT_MAY_EXEC 0x20
```
lrh2000: Prefer to avoid the kernel's internal name in regression tests. We may change the exact value in the future (e.g., if we add `PROT_SEM`), and we don't want to update the regression tests, which are user-space programs, when we change the kernel implementation details.

This is (just) a known invalid permission in regression tests. Use `PROT_INVALID` instead.

lrh2000 (approved):
LGTM except the minor comment above (https://github.com/asterinas/asterinas/pull/2926#discussion_r2726749544). Thanks for your contribution!

Issue #2856, opened 2026-01-07
Add the xfstests file system test suite


tatetian:
The [xfstests](https://github.com/kdave/xfstests) is perhaps the most comprehensive test suite out there for file systems. Recently, a number of bugs have been found in Asterinas's Ext2 implementation. This clearly show the urgency to improve its maturity and use a comprehensive FS test suite as a measure for the maturity improvement.

This issue proposes to introduce a **FS test infra** as a complement to our current syscall test infra (under `test/initramfs/syscall`), benchmark test infra (under `test/initramfs/benchmark`), and NixOS application test infra (under `test/nixos`).

(The test organization mentioned above is different from the current one and will be put into place soon. See this [comment](https://github.com/asterinas/asterinas/pull/2771#discussion_r2667127935)).

Under this test organization, the most natural location to add the FS test infra is `test/initramfs`:

```
test/
    initramfs/
        src/
            fs/
                xfstests/
        nix/
            fs/
                default.nix
                xfstests.nix
```

Adding the FS test infra requires changes to the root Makefile, the `test` directory, and the Github workflow files. The FS test infra should support testing multiple FS types, including but not limited to Ext2, RamFS, Exfat, and OverlayFS.

Initially, a great portion of test cases in xfstests is expected to fail. So we need to control the subset to run with a whitelist or a blocklist. Similar to how we control the LTP and gVisor test suites.

Another consideration is the timing to run FS tests. It may take an substantial amount of time to run all tests. So it would be impractical to run them for every PR. We may identify a subset of them as _essential_, which are light-weight enough to run as "smoke test" for every PR. The full set of tests may be triggered on a nightly basis, similar to the TDX CI test and the benchmark CI tests.

PR #2929, opened 2026-01-26
Add `/proc/mounts` support and expose mount sources in mountinfo

Fischer0522:
This PR implements `/proc/[pid]/mounts` and establishes `/proc/mounts` as a symlink, enabling support for standard tools like `findmnt` and `umount` by device.

In `xfstests` (#2856 ), test scripts rely on `findmnt` to check mount status and often perform unmount operations using the device path (e.g., `umount /dev/vdb`) rather than the mount point. Both actions require parsing `/proc/mounts` to retrieve the mapping between devices and mount points. Previously, the lack of mount source tracking and the missing `/proc/mounts` interface caused these tests to fail.

**Changes**
1.  Track Mount Source: Added a `Source` field to the `Mounts` struct to store the device path or filesystem name.
2.  Update `mountinfo`: The source string is now correctly exposed in `/proc/self/mountinfo`.
3.  Implement `mounts`:
    - Implemented `/proc/[pid]/mounts` (formatted according to Linux standards).
    - Created `/proc/mounts` as a symlink pointing to `/proc/self/mounts`.

With these changes, Asterinas can now:
- Correctly display source devices in `/proc/[pid]/mountinfo`.
- Support `findmnt` and `umount <device_path>` commands that are used in `xfstests` (#2856 ).

### DIFF
--- a/kernel/src/fs/path/mount.rs
+++ b/kernel/src/fs/path/mount.rs
@@ -160,6 +160,15 @@ pub struct Mount {
     mountpoint: RwLock<Option<Arc<Dentry>>>,
     /// The associated FS.
     fs: Arc<dyn FileSystem>,
+    /// The mount source (e.g., a device path like "/dev/vda" or a filesystem name like "proc").
+    ///
+    /// The source is stored in `Mount` instead of requiring each filesystem to provide it.
+    /// If a filesystem does not provide a source, it falls back to the value stored in `Mount`.
+    /// This behavior aligns with that of Linux. Concrete examples can be found here:
+    /// <https://github.com/asterinas/asterinas/pull/2929#discussion_r2729739818>.
+    ///
+    /// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/mount.h#L68>
+    source: Option<String>,
     /// The parent mount node.
     parent: RwLock<Option<Weak<Mount>>>,
     /// Child mount nodes which are mounted on one dentry of self.
@@ -186,7 +195,8 @@ impl Mount {
         fs: Arc<dyn FileSystem>,
         mnt_ns: Weak<MountNamespace>,
     ) -> Arc<Self> {
-        Self::new(fs, PerMountFlags::default(), None, mnt_ns)
+        let source = fs.name().to_string();
+        Self::new(fs, PerMountFlags::default(), None, mnt_ns, Some(source))
     }
 
     /// Creates a pseudo mount node with an associated FS.
@@ -194,7 +204,7 @@ impl Mount {
     /// This pseudo mount is not mounted on other mount nodes, has no parent, and does not
     /// belong to any mount namespace.
     pub(in crate::fs) fn new_pseudo(fs: Arc<dyn FileSystem>) -> Arc<Self> {
-        Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new())
+        Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new(), None)
     }
 
     /// The internal constructor.
@@ -210,6 +220,7 @@ impl Mount {
         flags: PerMountFlags,
         parent_mount: Option<Weak<Mount>>,
         mnt_ns: Weak<MountNamespace>,
+        source: Option<String>,
     ) -> Arc<Self> {
         let id = ID_ALLOCATOR.get().unwrap().lock().alloc().unwrap();
 
@@ -221,6 +232,7 @@ impl Mount {
             children: RwLock::new(HashMap::new()),
             propagation: RwLock::new(MountPropType::default()),
             fs,
+            source,
             mnt_ns,
             flags: AtomicPerMountFlags::new(flags),
             this: weak_self.clone(),
@@ -230,6 +242,11 @@ impl Mount {
     /// Gets the mount ID.
     pub fn id(&self) -> usize {
         self.id
+    }
+
+    /// Returns the mount source.
+    pub(in crate::fs) fn source(&self) -> Option<&str> {
+        self.fs.source().or(self.source.as_deref())
     }
 
     /// Mounts a fs on the mountpoint, it will create a new child mount node.
@@ -242,19 +259,28 @@ impl Mount {
     /// It is allowed to mount a fs even if the fs has been provided to another
     /// mountpoint. It is the fs's responsibility to ensure the data consistency.
     ///
+    /// If the source is provided by user, it will be recorded in the new mount.
+    ///
     /// Return the mounted child mount.
     pub(super) fn do_mount(
         self: &Arc<Self>,
         fs: Arc<dyn FileSystem>,
         flags: PerMountFlags,
         mountpoint: &Arc<Dentry>,
+        source: Option<String>,
     ) -> Result<Arc<Self>> {
         if mountpoint.type_() != InodeType::Dir {
             return_errno!(Errno::ENOTDIR);
         }
 
         let key = mountpoint.key();
-        let child_mount = Self::new(fs, flags, Some(Arc::downgrade(self)), self.mnt_ns.clone());
+        let child_mount = Self::new(
+            fs,
+            flags,
+            Some(Arc::downgrade(self)),
+            self.mnt_ns.clone(),
+            source,
+        );
         self.children.write().insert(key, child_mount.clone());
         child_mount.set_mountpoint(mountpoint);
 
@@ -296,6 +322,7 @@ impl Mount {
             children: RwLock::new(HashMap::new()),
             propagation: RwLock::new(MountPropType::default()),
             fs: self.fs.clone(),
+            source: self.source.clone(),
             mnt_ns: new_ns.cloned().unwrap_or_else(|| self.mnt_ns.clone()),
             flags: AtomicPerMountFlags::new(self.flags.load(Ordering::Relaxed)),
             this: weak_self.clone(),
@@ -405,12 +432,12 @@ impl Mount {
     }
 
     /// Gets the root `Dentry` of this mount node.
-    pub(super) fn root_dentry(&self) -> &Arc<Dentry> {
+    pub(in crate::fs) fn root_dentry(&self) -> &Arc<Dentry> {
         &self.root_dentry
     }
 
     /// Gets the mountpoint `Dentry` of this mount node if any.
-    pub(super) fn mountpoint(&self) -> Option<Arc<Dentry>> {
+    pub(in crate::fs) fn mountpoint(&self) -> Option<Arc<Dentry>> {
         self.mountpoint.read().clone()
     }
 
@@ -481,7 +508,7 @@ impl Mount {
     }
 
     /// Gets the parent mount node if any.
-    pub(super) fn parent(&self) -> Option<Weak<Self>> {
+    pub(in crate::fs) fn parent(&self) -> Option<Weak<Self>> {
         self.parent.read().as_ref().cloned()
     }
 
@@ -491,11 +518,11 @@ impl Mount {
     }
 
     /// Gets the associated FS.
-    pub(super) fn fs(&self) -> &Arc<dyn FileSystem> {
+    pub(in crate::fs) fn fs(&self) -> &Arc<dyn FileSystem> {
         &self.fs
     }
 
-    pub(super) fn flags(&self) -> PerMountFlags {
+    pub(in crate::fs) fn flags(&self) -> PerMountFlags {
         self.flags.load(Ordering::Relaxed)
     }
 
--- a/kernel/src/fs/path/resolver.rs
+++ b/kernel/src/fs/path/resolver.rs
@@ -2,15 +2,14 @@
 
 use alloc::str;
 
-use aster_util::printer::VmPrinter;
 use ostd::task::Task;
 
-use super::Path;
+use super::{Mount, Path};
 use crate::{
     fs::{
         file_table::{FileDesc, get_file_fast},
-        path::{MountNamespace, PerMountFlags},
-        utils::{FsFlags, InodeType, NAME_MAX, PATH_MAX, Permission, SYMLINKS_MAX, SymbolicLink},
+        path::MountNamespace,
+        utils::{InodeType, NAME_MAX, PATH_MAX, Permission, SYMLINKS_MAX, SymbolicLink},
     },
     prelude::*,
     process::posix_thread::AsThreadLocal,
@@ -253,122 +252,43 @@ impl AbsPathResult {
 
 // Mount info reading implementation
 impl PathResolver {
-    /// Reads the information of the mounts visible to this resolver.
+    /// Collects the mounts visible to this resolver.
     ///
     /// Here, the visible mounts are defined as follows:
     /// 1. If the resolver's root is a mount point, the visible mounts are the mount of the
     ///    resolver's root directory and all of its descendant mounts in the mount tree.
     /// 2. If the resolver's root is not a mount point, the visible mounts are all descendant
     ///    mounts that are mounted under the resolver's root directory.
-    pub fn read_mount_info(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
-        let mut printer = VmPrinter::new_skip(writer, offset);
-
-        let mut stack = Vec::new();
-        if self.root.is_mount_root() {
-            stack.push(self.root.mount.clone());
-        } else {
-            // The root is not a mount root, so we need to find the visible child mounts.
-            let children = self.root.mount.children.read();
-            for child_mount in children.values() {
-                if child_mount
-                    .mountpoint()
-                    .is_some_and(|dentry| dentry.is_equal_or_descendant_of(&self.root.dentry))
-                {
-                    stack.push(child_mount.clone());
-                }
-            }
-        }
+    ///
+    /// The mounts are collected in depth-first order.
+    pub(in crate::fs) fn collect_visible_mounts(&self) -> Vec<Arc<Mount>> {
+        let mut visible = Vec::new();
+        let mut stack = vec![self.root.mount.clone()];
+        let is_root_mount_root = self.root.is_mount_root();
 
         while let Some(mount) = stack.pop() {
-            let mount_id = mount.id();
-            let parent = mount.parent().and_then(|parent| parent.upgrade());
-            let parent_id = parent.as_ref().map_or(mount_id, |p| p.id());
-            let root = mount.root_dentry().path_name();
-            let mount_point = if let Some(parent) = parent {
-                if let Some(mount_point_dentry) = mount.mountpoint() {
-                    self.make_abs_path(&Path::new(parent, mount_point_dentry))
-                        .into_string()
-                } else {
-                    "".to_string()
-                }
-            } else {
-                // No parent means it's the root of the namespace.
-                "/".to_string()
-            };
-            let mount_flags = mount.flags();
-            let fs_type = mount.fs().name();
-            let fs_flags = mount.fs().flags();
+            let is_root_mount = Arc::ptr_eq(&mount, &self.root.mount);
 
-            // The following fields are dummy for now.
-            let major = 0;
-            let minor = 0;
-            let source = "none";
-
-            let entry = MountInfoEntry {
-                mount_id,
-                parent_id,
-                major,
-                minor,
-                root: &root,
-                mount_point: &mount_point,
-                mount_flags,
-                fs_type,
-                source,
-                fs_flags,
-            };
-
-            writeln!(printer, "{}", entry)?;
+            // Add the root mount only if `self` is at the mount root.
+            if !is_root_mount || is_root_mount_root {
+                visible.push(mount.clone());
+            }
 
             let children = mount.children.read();
             for child_mount in children.values() {
+                if is_root_mount && !is_root_mount_root {
+                    let Some(mountpoint) = child_mount.mountpoint() else {
+                        continue;
+                    };
+                    if !mountpoint.is_equal_or_descendant_of(&self.root.dentry) {
+                        continue;
+                    }
+                }
                 stack.push(child_mount.clone());
             }
         }
 
-        Ok(printer.bytes_written())
-    }
-}
-
-/// A single entry in the mountinfo file.
-struct MountInfoEntry<'a> {
-    /// A unique ID for the mount (but not guaranteed to be unique across reboots).
-    mount_id: usize,
-    /// The ID of the parent mount (or self if it has no parent).
-    parent_id: usize,
-    /// The major device ID of the filesystem.
-    major: u32,
-    /// The minor device ID of the filesystem.
-    minor: u32,
-    /// The root of the mount within the filesystem.
-    root: &'a str,
-    /// The mount point relative to the process's root directory.
-    mount_point: &'a str,
-    /// Per-mount flags.
-    mount_flags: PerMountFlags,
-    /// The type of the filesystem in the form "type[.subtype]".
-    fs_type: &'a str,
-    /// Filesystem-specific information or "none".
-    source: &'a str,
-    /// Per-filesystem flags.
-    fs_flags: FsFlags,
-}
-
-impl core::fmt::Display for MountInfoEntry<'_> {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(
-            f,
-            "{} {} {}:{} {} {} {} - {} {} {}",
-            self.mount_id,
-            self.parent_id,
-            self.major,
-            self.minor,
-            &self.root,
-            &self.mount_point,
-            &self.mount_flags,
-            &self.fs_type,
-            &self.source,
-            &self.fs_flags,
-        )
+        visible
     }
 }
 
--- a/kernel/src/fs/procfs/mounts.rs
+++ b/kernel/src/fs/procfs/mounts.rs
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use crate::{
+    fs::{
+        procfs::{ProcSymBuilder, SymOps},
+        utils::{Inode, SymbolicLink, mkmod},
+    },
+    prelude::*,
+};
+
+/// Represents the inode at `/proc/mounts`.
+pub struct MountsSymOps;
+
+impl MountsSymOps {
+    pub fn new_inode(parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        // Reference:
+        // <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/root.c#L291>
+        // <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/generic.c#L466>
+        ProcSymBuilder::new(Self, mkmod!(a+rwx))
+            .parent(parent)
+            .build()
+            .unwrap()
+    }
+}
+
+impl SymOps for MountsSymOps {
+    fn read_link(&self) -> Result<SymbolicLink> {
+        Ok(SymbolicLink::Plain("self/mounts".to_string()))
+    }
+}
--- a/kernel/src/fs/procfs/pid/task/mod.rs
+++ b/kernel/src/fs/procfs/pid/task/mod.rs
@@ -12,8 +12,8 @@ use crate::{
                 cgroup::CgroupFileOps, cmdline::CmdlineFileOps, comm::CommFileOps,
                 environ::EnvironFileOps, exe::ExeSymOps, fd::FdDirOps, gid_map::GidMapFileOps,
                 maps::MapsFileOps, mem::MemFileOps, mountinfo::MountInfoFileOps,
-                oom_score_adj::OomScoreAdjFileOps, stat::StatFileOps, status::StatusFileOps,
-                uid_map::UidMapFileOps,
+                mounts::MountsFileOps, oom_score_adj::OomScoreAdjFileOps, stat::StatFileOps,
+                status::StatusFileOps, uid_map::UidMapFileOps,
             },
             template::{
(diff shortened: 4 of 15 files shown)


Fischer0522:
Additionally, since `PathResolver::read_mount_info` and `PathResolver::read_mounts` share the same logic for collecting visible mounts, maybe we can abstract this logic into a shared helper method？

kernel/src/fs/procfs/mounts.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::{
+    fs::{
+        procfs::{ProcSymBuilder, SymOps},
+        utils::{Inode, SymbolicLink, mkmod},
+    },
+    prelude::*,
+};
+
+/// Represents the inode at `/proc/mounts`.
```
cchanging, line 11: Insert a blank line here.

kernel/src/syscall/mount.rs
```diff
@@ -198,8 +198,29 @@ fn do_new_mount(
...
+
+    let source = if fs_type_obj.properties().contains(FsProperties::NEED_DISK) {
+        // Block device filesystem: use device path
+        ctx.user_space()
+            .read_cstring(src_name_addr, MAX_FILENAME_LEN)?
+            .to_string_lossy()
+            .into_owned()
+    } else {
+        // Virtual filesystem: use filesystem type name
+        fs_type_str.to_string()
```
cchanging: I tested this logic in Linux 6.11.0. It seems that when the source parameter is NULL, the source string in the `mounts` and `mountinfo` will be `none`.

kernel/src/fs/path/mount.rs
```diff
@@ -186,15 +188,17 @@ impl Mount {
...
+        Self::new(fs, PerMountFlags::default(), None, mnt_ns, source)
     }
 
     /// Creates a pseudo mount node with an associated FS.
     ///
     /// This pseudo mount is not mounted on other mount nodes, has no parent, and does not
     /// belong to any mount namespace.
     pub(in crate::fs) fn new_pseudo(fs: Arc<dyn FileSystem>) -> Arc<Self> {
-        Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new())
+        let source = fs.name().to_string();
```
cchanging: I think the `source` field in `Mount` should be `Option<String>`. On one hand, for pseudo filesystems, this field is completely unnecessary. On the other hand, when the mount syscall does not pass a source parameter, it is natural to pass `None` as the source argument.

kernel/src/fs/path/mount.rs
```diff
@@ -232,6 +238,11 @@ impl Mount {
         self.id
     }
 
+    /// Returns the mount source (device path or filesystem name).
+    pub fn source(&self) -> &str {
```
cchanging: `pub(super)` is enough. We should control the visibility of the API to the greatest extent possible.

kernel/src/fs/path/resolver.rs
```diff
@@ -318,11 +339,45 @@ impl PathResolver {
...
+        }
 
-            let children = mount.children.read();
-            for child_mount in children.values() {
-                stack.push(child_mount.clone());
-            }
+        Ok(printer.bytes_written())
+    }
+
+    /// Reads the information of the mounts visible to this resolver.
```
cchanging: Make the documentation for `read_mounts` and `read_mountinfo` somewhat distinct, clarifying which file each typically corresponds to.

kernel/src/fs/path/resolver.rs
```diff
@@ -372,6 +427,37 @@ impl core::fmt::Display for MountInfoEntry<'_> {
     }
 }
 
+struct MountsEntry<'a> {
```
cchanging: To be consistent, add documents to `MountsEntry` like `MountInfoEntry`

kernel/src/fs/path/resolver.rs
```diff
@@ -260,31 +260,52 @@ impl PathResolver {
...
         let mut stack = Vec::new();
-        if self.root.is_mount_root() {
-            stack.push(self.root.mount.clone());
-        } else {
-            // The root is not a mount root, so we need to find the visible child mounts.
-            let children = self.root.mount.children.read();
+        let mut visible = Vec::new();
+        let filter_root_children = !self.root.is_mount_root();
+
+        stack.push((self.root.mount.clone(), filter_root_children));
```
cchanging: I think this approach is a bit odd. 

To handle the special case where root isn't the mount root, every element in the stack now has an extra bool, which doesn't seem very necessary. 

Additionally, maintaining two vectors here introduces some efficiency concerns. 

I think we could stick with just one working vector and avoid popping elements, instead using something like `while index < visible.len()` for iteration.

Fischer0522: The `while index < len` approach effectively implements BFS, which diverges from the DFS strategy used by Linux. To maintain compatibility with the Linux implementation, perhaps we should switch to recursion?

kernel/src/syscall/mount.rs
```diff
@@ -198,8 +198,29 @@ fn do_new_mount(
...
+    };
+
     let fs = get_fs(src_name_addr, flags, fs_type, data_addr, ctx)?;
-    target_path.mount(fs, flags.into(), ctx)?;
+    target_path.mount(fs, flags.into(), source, ctx)?;
```
tatetian, line 224: First, I really doubt that the correct implementation would record this `source` literally as a string. Consider four lightly different `mount` system call.

```c
mount("/dev/vda",          "/mnt", "ext4", null)
mount("vda",               "/mnt", "ext4", null) // CWD = "/dev/"
mount("../vda",            "/mnt", "ext4", null) // CWD = "/dev/xxx"
mount("/dev/../dev/vda",   "/mnt", "ext4", null)
```

If `source` is recorded literally, then the four `mount` system call would have four different "source" displayed in procfs, while in fact they refer to the same source.

Second, at the conceptual level, the source information is already captured by a FS instance. A pseudo FS does not have a source. A persistent FS such as Ext2 must include a field that refers to the underlying disk, which is the source.

https://github.com/asterinas/asterinas/blob/459a634e4acc865b4bad99116a416910835083e5/kernel/src/fs/ext2/fs.rs#L22-L33

I think we should find an elegant way to extract the source information already kept by a FS. I emphasize "elegant" because there are some technical obstacles to do so. Although `BlockDevice` is this source, this trait does not provide a way to get its path (I am not saying that the trait should do so!). In addition, the `FileSystem` trait does not provide a method to expose the underlying block device. So the trait should be extended in one of the two options:

```rust
pub trait FileSystem {
    fn source(&self) -> Option<Path>;
}

pub trait FileSystem {
    fn source(&self) -> Option<String>;
}
```

I think you need to check out Linux's implementation in order to decide what to do.

kernel/src/fs/path/resolver.rs
```diff
@@ -260,31 +260,52 @@ impl PathResolver {
     ///    resolver's root directory and all of its descendant mounts in the mount tree.
     /// 2. If the resolver's root is not a mount point, the visible mounts are all descendant
     ///    mounts that are mounted under the resolver's root directory.
-    pub fn read_mount_info(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
-        let mut printer = VmPrinter::new_skip(writer, offset);
-
+    fn collect_visible_mounts(&self) -> Vec<Arc<Mount>> {
```
tatetian: This is a good refactoring. We can push in this direction one step further. It is actually not good to have `read_mount_info` and `read_mounts` as methods of `PathResolver` because they are highly coupled with procfs. Now that you add a `collect_visible_mounts` method, we can make it public and then migrate the logic of `read_mount_info` and `read_mounts` to ProcFS.

kernel/src/syscall/mount.rs
cchanging: > Second, at the conceptual level, the source information is already captured by a FS instance.

This makes sense. We can create many `Mount` objects for a file system through bind mounts, but they should correspond to the same source. There is no need for us to store multiple copies.  

> So the trait should be extended in one of the two options:

For filesystems that do not require a disk, I recall that in Linux, the source can be specified as any arbitrary string. If true, we might need to consider the latter API:

```rust
pub trait FileSystem {
    fn source(&self) -> Option<String>;
}
```

kernel/src/fs/path/resolver.rs
cchanging: > which diverges from the DFS strategy used by Linux. 

I'm not sure whether the order in which mounts are displayed matters, but it does make sense to align with Linux's behavior. I agree with maintaining DFS order.

> perhaps we should switch to recursion

In operating systems, thread stack resources are relatively limited, so we should avoid recursion as much as possible. 

After removing the bool values from the vector, the current implementation feels quite clean. If we want to further eliminate the extra vector, we could consider refactoring the function like this:

```rust
pub fn for_each_visible_mount<F>(&self, mut f: F) -> Result<()>
where
    F: FnMut(&Arc<Mount>) -> Result<()>,
{
    let mut stack = Vec::new();
    ...

    while let Some(mount) = stack.pop() {
        f(&mount)?;
        ...
    }
}
```

Yet this function is unlikely to be a performance bottleneck, so keeping the current implementation is also acceptable.

kernel/src/syscall/mount.rs
Fischer0522: > If source is recorded literally, then the four mount system call would have four different "source" displayed in procfs, while in fact they refer to the same source.

I did a quick experiment on Linux regarding how mount sources are recorded, and the results were quite different from what I expected. I initially thought Linux would normalize the source path, but it actually records the literal string provided in the syscall.

For example, if I mount the same device using different relative paths, `cat /proc/mounts` shows them exactly as I typed them:

```bash
# cat /proc/mounts
/dev/loop6 /tmp/mnt1 ext4 rw,relatime 0 0
./loop6 /tmp/mnt3 ext4 rw,relatime 0 0
../dev/loop6 /tmp/mnt2 ext4 rw,relatime 0 0
```

I did some digging into the kernel source, and here is how Linux handles it:
1. Linux stores the raw string in `fs_context->source` during the parsing stage.
2. Even though it validates the source (via `lookup_bdev`), it never actually updates that original string.
3. The `mount` struct just copies this raw string into `mnt_devname` (if the source is  `NULL`, `mnt_devname` will be set as "none").
4. When reading /proc/mounts, the kernel tries a `show_devname` callback that is provided by file systems. However, most filesystems like ext2/ext4 do not implement this, so it just falls back to printing the raw `mnt_devname`. (Only a few like Btrfs or NFS actually register the callback).

With the standard mount command, paths typically appear absolute because the utility resolves them (e.g., via `getcwd`) before issuing the syscall.

Since Linux keeps it this way, should we align with their behavior and record the raw `source` in Asterinas' `Mount` struct? Or would we prefer to let the filesystem provide a normalized source?

tatetian: @Fischer0522 Thanks for digging into Linux's source code! This is quite a surprise to me because recording and displaying the raw source string leads to ambiguity. But if this is the Linux behavior, then let's follow it in the name of compatibility. Please add some comments (referring to the discussion) to justify recoding the raw source string.

Fischer0522: One last thing: I’ve introduced a `source()` method in trait `FileSystem` to align with Linux’s `get_devname`.
``` Rust
pub trait FileSystem {
    fn source(&self) -> Option<String>;
}
```
But since most file systems in Linux don't register this callback, should we implement this method for our file systems (e.g., ext2 or exFAT)?

kernel/src/syscall/mount.rs
```diff
@@ -198,16 +198,37 @@ fn do_new_mount(
...
+        Some(s)
+    };
+
+    let fs = get_fs(source.as_deref(), flags, fs_type, data_addr, ctx)?;
+    target_path.mount(fs, flags.into(), source, ctx)?;
     Ok(())
 }
 
 /// Gets the filesystem by fs_type and devname.
 fn get_fs(
```
tatetian: Let's rename this function to `open_fs` because the verb `get` sounds too light-weight.

kernel/src/syscall/mount.rs
```diff
@@ -198,16 +198,37 @@ fn do_new_mount(
...
-    src_name_addr: Vaddr,
+    devname: Option<&str>,
```
tatetian: `devname` --> `dev_name`

kernel/src/syscall/mount.rs
```diff
@@ -198,16 +198,37 @@ fn do_new_mount(
...
         return_errno_with_message!(Errno::EINVAL, "fs_type is empty");
     }
-    let fs = get_fs(src_name_addr, flags, fs_type, data_addr, ctx)?;
-    target_path.mount(fs, flags.into(), ctx)?;
+
+    // Determine the source string based on filesystem type
+    let fs_type_str = fs_type
+        .to_str()
+        .map_err(|_| Error::with_message(Errno::ENODEV, "invalid file system type"))?;
+    let fs_type = crate::fs::registry::look_up(fs_type_str).ok_or(Error::with_message(
```
tatetian: Group the entire code paragraph about `fs_type` into a code block for improved clarity.

```rust
    let fs_type = {
        let fs_type_cstr = ctx
            .user_space()
            .read_cstring(fs_type_addr, MAX_FILENAME_LEN)?;
        if fs_type_cstr.is_empty() {
            return_errno_with_message!(Errno::EINVAL, "fs_type is empty");
        }

        // Determine the source string based on filesystem type
        let fs_type_str = fs_type_cstr
            .to_str()
            .map_err(|_| Error::with_message(Errno::ENODEV, "invalid file system type"))?;

        crate::fs::registry::look_up(fs_type_str).ok_or(Error::with_message(
            Errno::ENODEV,
            "the filesystem is not configured in the kernel",
        ))?
    };
```

kernel/src/fs/procfs/pid/task/mounts.rs
```diff
@@ -0,0 +1,116 @@
...
+        path::{Path, PathResolver, PerMountFlags},
+        procfs::template::{FileOps, ProcFileBuilder},
+        utils::{Inode, mkmod},
+    },
+    prelude::*,
+    process::posix_thread::AsPosixThread,
+};
+
+/// A single entry in the mounts file.
+struct MountsEntry<'a> {
```
tatetian: A single entry represents _a_ mount. So this struct should be named `MountEntry`.

tatetian (commented):
The code looks great, with only a few minor issues to get fixed.

To ensure that our implementation is correct, it would be great to have some tests added for `/proc/self/mounts` or `/proc/self/mountinfo`. There are at least three relevant test cases from the Gvisor [procfs test suites](https://github.com/google/gvisor/blob/9dd9b9b1b6c4456ad37a904aee09169ca9ff6c01/test/syscalls/linux/proc.cc) that should be removed from [the block list](https://github.com/asterinas/asterinas/blob/main/test/initramfs/src/syscall/gvisor/blocklists/proc_test):

```
ProcMounts.IsSymlink
ProcSelfMounts.ContainsProcfsEntry
ProcSelfMounts.RequiredFieldsArePresent
```

kernel/src/fs/path/mount.rs
```diff
@@ -160,6 +160,12 @@ pub struct Mount {
     mountpoint: RwLock<Option<Arc<Dentry>>>,
     /// The associated FS.
     fs: Arc<dyn FileSystem>,
+    /// The mount source (device path like "/dev/vda" or filesystem name like "proc").
+    ///
+    /// Aligned with Linux behavior for consistency.
+    /// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/mount.h#L68>
+    // Discussion: <https://github.com/asterinas/asterinas/pull/2929#discussion_r2729739818>
```
lrh2000: Without delving into the discussion, it's difficult for the reader to understand what "Linux behavior" is trying to convey. Let's add a few words to summarize why we want to leave the comments. It can be as simple as:
```rust
    /// The source is a string, not a [`Path`]. This behavior aligns with that of Linux.
    /// Concrete examples can be found here:
    /// <https://github.com/asterinas/asterinas/pull/2929#discussion_r2729739818>.
    ///
    /// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/mount.h#L68>
```

kernel/src/fs/path/resolver.rs
```diff
@@ -253,122 +252,43 @@ impl AbsPathResult {
...
-                mount_point: &mount_point,
-                mount_flags,
-                fs_type,
-                source,
-                fs_flags,
-            };
+            let is_root_mount = Arc::ptr_eq(&mount, &self.root.mount);
 
-            writeln!(printer, "{}", entry)?;
+            // Only add root mount if it's a mount root
```
lrh2000:
```suggestion
            // Add the root mount only if `self` is at the mount root.
```

kernel/src/fs/procfs/pid/task/mounts.rs
```diff
@@ -0,0 +1,116 @@
...
+            &self.source,
+            &self.mount_point,
+            &self.fs_type,
+            &self.mount_flags,
+            &self.dump,
+            &self.pass,
+        )
+    }
+}
+pub struct MountsFileOps(TidDirOps);
```
lrh2000, line 52: - Add a blank line.
 - Also the comments:
```rust
/// Represents the inode at `/proc/[pid]/task/[tid]/mounts` (and also `/proc/[pid]/mounts`).
pub struct MountInfoFileOps(TidDirOps);
```

kernel/src/fs/procfs/pid/task/mounts.rs
```diff
@@ -0,0 +1,116 @@
...
+            .parent(parent)
+            .build()
+            .unwrap()
+    }
+
+    /// Reads mount information for `/proc/[pid]/mounts` and `/proc/mounts`.
+    ///
+    /// Provides a simplified view of mounted filesystems in the traditional
+    /// `/etc/fstab` format.
+    pub fn read_mounts(
```
lrh2000:
```suggestion
    fn read_mounts(
```

kernel/src/syscall/mount.rs
```diff
@@ -218,18 +241,10 @@ fn get_fs(
...
-        Errno::ENODEV,
-        "the filesystem is not configured in the kernel",
-    ))?;
-
     let disk = if fs_type.properties().contains(FsProperties::NEED_DISK) {
-        let devname = user_space.read_cstring(src_name_addr, MAX_FILENAME_LEN)?;
-        let path = devname.to_string_lossy();
-        let fs_path = FsPath::from_fd_and_path(AT_FDCWD, path.as_ref())?;
+        let devname =
+            dev_name.ok_or_else(|| Error::with_message(Errno::EINVAL, "No source specified"))?;
```
lrh2000:
```suggestion
            dev_name.ok_or_else(|| Error::with_message(Errno::EINVAL, "the source is not specified"))?;
```

Be consistent with the following "the path is no the device file"/"the device is not found" messages.

kernel/src/syscall/mount.rs
```diff
@@ -192,22 +192,45 @@ fn do_new_mount(
...
-        return_errno_with_message!(Errno::EINVAL, "fs_type is empty");
-    }
-    let fs = get_fs(src_name_addr, flags, fs_type, data_addr, ctx)?;
-    target_path.mount(fs, flags.into(), ctx)?;
+    let fs_type = {
+        let fs_type_cstr = ctx
+            .user_space()
+            .read_cstring(fs_type_addr, MAX_FILENAME_LEN)?;
+        if fs_type_cstr.is_empty() {
+            return_errno_with_message!(Errno::EINVAL, "fs_type is empty");
```
lrh2000: To be consistent with the following `invalid file system type` message, change this to `empty file system type`. (This was not introduced in this PR, but it appears in the diff because the code was moved around. )

```suggestion
            return_errno_with_message!(Errno::EINVAL, "empty file system type");
```

kernel/src/fs/procfs/mounts.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct MountsSymOps;
+
+impl MountsSymOps {
+    pub fn new_inode(parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        ProcSymBuilder::new(Self, mkmod!(a+rwx))
```
lrh2000, line 19: For the permission (symbolic links are almost always `a+rwx`, but let's be consistent with other files anyway):
```rust
        // Reference:
        // <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/root.c#L291>
        // <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/generic.c#L466>
```

kernel/src/syscall/mount.rs
```diff
@@ -192,22 +192,45 @@ fn do_new_mount(
...
+        crate::fs::registry::look_up(fs_type_str).ok_or(Error::with_message(
+            Errno::ENODEV,
+            "the filesystem is not configured in the kernel",
+        ))?
+    };
+
+    let source = if src_name_addr == 0 {
+        None
+    } else {
+        let s = ctx
```
lrh2000:
```suggestion
        let source = ctx
```

kernel/src/syscall/mount.rs
lrh2000: https://github.com/asterinas/asterinas/pull/2929#discussion_r2744431600

> devname --> dev_name


I suppose the following `devname` should be changed together:
```rust
        let devname =
            dev_name.ok_or_else(...)
```

tatetian:
@Fischer0522 Could you resolve or reply all issues after you update the PR?

Fischer0522:
> @Fischer0522 Could you resolve or reply all issues after you update the PR?

All done.

kernel/src/fs/path/mount.rs
```diff
@@ -248,13 +265,20 @@ impl Mount {
         fs: Arc<dyn FileSystem>,
         flags: PerMountFlags,
         mountpoint: &Arc<Dentry>,
+        source: Option<String>,
```
tatetian, line 270: The function's Rust doc should add a description about this `source` argument.

tatetian (approved):
LGTM. Only a minor issue.

tatetian:
Need @cchanging and @lrh2000 's approvals.

kernel/src/syscall/mount.rs
```diff
@@ -192,22 +192,45 @@ fn do_new_mount(
...
+            .into_owned();
+        Some(source)
+    };
+
+    let fs = open_fs(source.as_deref(), flags, fs_type, data_addr, ctx)?;
+    target_path.mount(fs, flags.into(), source, ctx)?;
     Ok(())
 }
 
 /// Gets the filesystem by fs_type and devname.
```
lrh2000:
```suggestion
/// Gets the filesystem by fs_type and dev_name.
```

kernel/src/syscall/mount.rs
```diff
@@ -192,22 +192,45 @@ fn do_new_mount(
...
-    target_path.mount(fs, flags.into(), ctx)?;
+    let fs_type = {
+        let fs_type_cstr = ctx
+            .user_space()
+            .read_cstring(fs_type_addr, MAX_FILENAME_LEN)?;
+        if fs_type_cstr.is_empty() {
+            return_errno_with_message!(Errno::EINVAL, "empty file system type");
+        }
+
+        // Determine the source string based on filesystem type
```
lrh2000: What does this comment mean? The following code looks up the FS and has nothing to do with the source string. In any case, I don't think this comment is necessary. Maybe just remove it.
```suggestion
```

kernel/src/fs/procfs/pid/task/mountinfo.rs
```diff
@@ -21,6 +67,71 @@ impl MountInfoFileOps {
...
+                }
+            } else {
+                // No parent means it's the root of the namespace.
+                "/".to_string()
+            };
+            let mount_flags = mount.flags();
+            let fs_type = mount.fs().name();
+            let fs_flags = mount.fs().flags();
+
+            let source = mount.source().unwrap_or("none");
```
lrh2000, line 124: The existing fields are listed in order. Keep the newly added field as well:
```rust
            let mount_point = ...;
            let mount_flags = mount.flags();
            let fs_type = mount.fs().name();
            let source = mount.source().unwrap_or("none");
            let fs_flags = mount.fs().flags();
```

kernel/src/fs/procfs/pid/task/mounts.rs
```diff
@@ -0,0 +1,118 @@
...
+                    "".to_string()
+                }
+            } else {
+                // No parent means it's the root of the namespace.
+                "/".to_string()
+            };
+            let mount_flags = mount.flags();
+            let fs_type = mount.fs().name();
+            let source = mount.source().unwrap_or("none");
+
```
lrh2000, line 87: Similar to `mount_info`. Let's make the dummy fields explicit by writing:
```rust
            // The following fields are dummy for now.
            let dump = 0;
            let pass = 0;

            let entry = MountEntry {
                source,
                mount_point: &mount_point,
                fs_type,
                mount_flags,
                dump,
                pass,
            };
```

kernel/src/fs/procfs/pid/task/mountinfo.rs
```diff
@@ -21,6 +67,71 @@ impl MountInfoFileOps {
...
+            let mount_id = mount.id();
+            let parent = mount.parent().and_then(|parent| parent.upgrade());
+            let parent_id = parent.as_ref().map_or(mount_id, |p| p.id());
+            let is_resolver_root_mount = Arc::ptr_eq(&mount, path_resolver.root().mount_node());
+            let root = if is_resolver_root_mount {
+                path_resolver.root().dentry().path_name()
+            } else {
+                mount.root_dentry().path_name()
+            };
+            let mount_point = if is_resolver_root_mount {
```
lrh2000: To avoid duplicate code, this can be extracted to a function. Given that it is only used in `mountinfo.rs` and `mount.rs`, perhaps the simplest solution is to define a local helper?
```rust
pub(super) fn make_mount_point_path(
    is_resolver_root_mount: bool,
    parent: Option<&Mount>,
    mount: &Mount,
    path_resolver: &PathResolver
) -> String {
    todo!();
}
```

kernel/src/fs/procfs/pid/task/mounts.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl FileOps for MountsFileOps {
+    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        let thread = self.0.thread();
+        let posix_thread = thread.as_posix_thread().unwrap();
+
+        let fs = posix_thread.read_fs();
+        let path_resolver = fs.resolver().read();
+        self.read_mounts(&path_resolver, offset, writer)
```
lrh2000, line 119: To be consistent, add a blank line before `self.read_mounts` in this file:
https://github.com/asterinas/asterinas/blob/66a0f7d2e8ae4e037ccf8a02009798d9d52eb06c/kernel/src/fs/procfs/pid/task/mounts.rs#L114-L116

**Or** remove the blank line before `self.read_mount_info` in `mountinfo.rs`.
https://github.com/asterinas/asterinas/blob/66a0f7d2e8ae4e037ccf8a02009798d9d52eb06c/kernel/src/fs/procfs/pid/task/mountinfo.rs#L142-L145

kernel/src/fs/procfs/pid/task/mounts.rs
Fischer0522: Unlike the `major` and `minor` fields in `mountinfo`, `dump` and `pass` are currently functional fields instead of  dummy fields in Asterinas. On Linux, `dump` and `pass` are hardcoded to 0. But I agree we should add a comment to clarify this.

kernel/src/fs/procfs/pid/task/mod.rs
```diff
@@ -37,11 +38,35 @@ mod gid_map;
...
 mod mem;
 mod mountinfo;
+mod mounts;
 mod oom_score_adj;
 mod stat;
 mod status;
 mod uid_map;
 
+/// A helper function to create the mount point path for a given mount (used by `mounts` and `mountinfo`).
+pub(super) fn make_mount_point_path(
```
lrh2000: Actually we should put the helper method in `mountinfo.rs` [^1].

[^1]: Only this way, it needs to be `pub(super)`, otherwise if it lives in the `mod.rs` it does not need to be `pub(super)`.

It's odd to see it in `task/mod.rs` since those who read `task/mod.rs` likely won't care about mount-related stuff.

kernel/src/fs/procfs/pid/task/mounts.rs
```diff
@@ -0,0 +1,118 @@
...
+                parent.as_ref(),
+                mount.as_ref(),
+                path_resolver,
+            );
+            let mount_flags = mount.flags();
+            let fs_type = mount.fs().name();
+            let source = mount.source().unwrap_or("none");
+
+            // The dump and pass fields are hardcoded to 0, because the kernel considers them userspace policy
+            // (managed by /etc/fstab) and does not store them in the VFS layer.
```
lrh2000:
```rust
            // The dump and pass fields are hardcoded to 0, because the kernel considers them
            // userspace policy (managed by /etc/fstab) and does not store them in the VFS layer.
```

The first line is too long. Break it.

kernel/src/fs/procfs/pid/task/mounts.rs
lrh2000: Yes. Thanks for the explanation and for adding the comments!

lrh2000:
Thank you for your contribution. I have no further comments here :(

tatetian:
> :(

Why unhappy face? 😆

lrh2000:
> Why unhappy face? 😆

I accidentally replaced the period with random sequences. I didn't even know it was a unhappy face??

PR #2933, opened 2026-01-28
Remove outdated gvisor workaround

lrh2000:
Remove this workaround since it should not make any sense now:
https://github.com/asterinas/asterinas/blob/22dd288b070b40a6a62232644a30145477c4916c/kernel/src/process/process/mod.rs#L383-L387

---

The second commit performs minor cleanups, such as re-unifying the variable name of `CurrentUserSpace`. Across our codebase, it's always `user_space`, except for two exceptions that use `userspace` and `uspace`.


StevenJiang1110 (approved):

PR #2934, opened 2026-01-28
Improve input parsing in `OomScoreAdjFileOps` by trimming whitespace

Fischer0522:
This PR fixes Asterinas failing to write to `/proc/[pid]/oom_score_adj` via `echo`. Currently, running:

```bash
~ # echo 1 > /proc/self/oom_score_adj
sh: write error: Invalid argument
```
fails because `echo` appends a trailing `\n`, which Asterinas does not parse correctly. This PR trims the input before parsing, aligning the behavior with Linux.


tatetian (approved):

Issue #2916, opened 2026-01-22
Some heap management problems


vvvvsv:
### Describe the bug

In Linux, if a page in the middle of the heap region is unmapped, causing the heap to be split into multiple VMAs, the heap can still be extended by `brk`.

### To Reproduce

Compile and run:

```c
#define _GNU_SOURCE
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mman.h>

static void dump_maps(const char *tag)
{
    printf("\n==== %s ====\n", tag);
    FILE *f = fopen("/proc/self/maps", "r");
    if (!f) {
        perror("fopen");
        exit(1);
    }

    char buf[4096];
    while (fgets(buf, sizeof(buf), f)) {
        fputs(buf, stdout);
    }
    fclose(f);
}

int main(void)
{
    printf("\n");
    long ps = sysconf(_SC_PAGESIZE);
    if (ps <= 0) {
        perror("sysconf");
        return 1;
    }

    void *brk0 = sbrk(0);
    printf("initial brk = %p\n", brk0);

    if (sbrk(5 * ps) == (void *)-1) {
        perror("sbrk #1");
        return 1;
    }
    void *brk1 = sbrk(0);
    printf("after sbrk #1: brk = %p\n", brk1);

    dump_maps("maps after sbrk(5 pages)");

    void *mid = (char *)brk0 + 2 * ps;
    printf("\nmunmap middle page: %p - %p\n",
           mid, (char *)mid + ps);

    if (munmap(mid, ps) != 0) {
        perror("munmap");
        return 1;
    }

    dump_maps("maps after munmap middle page");

    if (sbrk(5 * ps) == (void *)-1) {
        perror("sbrk #2");
        return 1;
    }
    void *brk2 = sbrk(0);
    printf("\nafter sbrk #2: brk = %p\n", brk2);

    dump_maps("maps after second sbrk(5 pages)");

    return 0;
}
```

### Expected behavior

```
initial brk = 0x62601fe76000
after sbrk #1: brk = 0x62601fe7b000

==== maps after sbrk(5 pages) ====
62600c18e000-62600c18f000 r--p 00000000 08:50 14997                      /home/vvsv/TMP/tmp1
62600c18f000-62600c190000 r-xp 00001000 08:50 14997                      /home/vvsv/TMP/tmp1
62600c190000-62600c191000 r--p 00002000 08:50 14997                      /home/vvsv/TMP/tmp1
[...]

PR #2942, opened 2026-01-30
Fix dead locks when generating `/proc/[pid]/maps`

lrh2000:
### Fix heap lock issues

The first commit resolves an issue with the lock order when generating `/proc/[pid]/maps`.
 - When modifying the heap end, the heap lock is taken before the MM lock. However, when printing `/proc/[pid]/maps`, the heap lock is taken after the MM lock. This inconsistent lock order will cause deadlock issues.
   https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/kernel/src/vm/vmar/vm_mapping.rs#L271
 - This commit also fixes _Prob2_ in https://github.com/asterinas/asterinas/issues/2916#issuecomment-3783075905.
> - **Prob2**: There is a race condition in `Vmar::fork_from()`. If the user program attempts to perform `fork()` and `brk()` concurrently, the forked program may end up with an incorrect heap region. Depending on how the race occurs, `brk()` in the child process may fail or succeed with unexpected results.
> https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/kernel/src/vm/vmar/vmar_impls/fork.rs#L26-L28

### Remove `VmMapping::inode`

The second commit removes the `inode` field from `VmMapping` because it contains duplicate information from the `path` field, which is also added when implementing `/proc/[pid]/maps`.
https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/kernel/src/vm/vmar/vm_mapping.rs#L70-L77

### Make `populate_device` infallible 

The third commit makes `VmMapping::populate_device` infallible. This is because it makes more sense to simply `unwrap` when creating the page table cursor. For example, we simply unwrap it in `VmMapping::unmap`:
https://github.com/asterinas/asterinas/blob/6d2ff13a639b9fff72842869cfd78f78ff21bbf2/kernel/src/vm/vmar/vm_mapping.rs#L738

### DIFF
--- a/kernel/src/vm/vmar/vmar_impls/map.rs
+++ b/kernel/src/vm/vmar/vmar_impls/map.rs
@@ -350,7 +350,7 @@ impl<'a> VmarMapOptions<'a> {
         // Exchange the operation is ok since we hold the write lock on the
         // VMAR.
         if let Some(io_mem) = io_mem {
-            vm_mapping.populate_device(parent.vm_space(), io_mem, vmo_offset)?;
+            vm_mapping.populate_device(parent.vm_space(), io_mem, vmo_offset);
         }
 
         // Add the mapping to the VMAR.
--- a/kernel/src/vm/vmar/vm_mapping.rs
+++ b/kernel/src/vm/vmar/vm_mapping.rs
@@ -188,22 +188,15 @@ impl VmMapping {
     ///
     /// In debug builds, this method panics if the mapping is not a device
     /// memory mapping.
-    pub(super) fn populate_device(
-        &self,
-        vm_space: &VmSpace,
-        io_mem: IoMem,
-        vmo_offset: usize,
-    ) -> Result<()> {
+    pub(super) fn populate_device(&self, vm_space: &VmSpace, io_mem: IoMem, vmo_offset: usize) {
         debug_assert!(matches!(self.mapped_mem, MappedMemory::Device));
 
         let preempt_guard = disable_preempt();
         let map_range = self.map_to_addr..self.map_to_addr + self.map_size.get();
-        let mut cursor = vm_space.cursor_mut(&preempt_guard, &map_range)?;
+        let mut cursor = vm_space.cursor_mut(&preempt_guard, &map_range).unwrap();
         let io_page_prop =
             PageProperty::new_user(PageFlags::from(self.perms), io_mem.cache_policy());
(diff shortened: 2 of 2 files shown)


tatetian:
@vvvvsv Since you are familiar with the affected code, could you review this PR?

kernel/src/vm/vmar/vmar_impls/map.rs
vvvvsv:
```rust
    /// Binds memory to the mapping based on the [`Mappable`] object of the file, and sets the [`Path`] of the mapping according to its [`Path`].
```

kernel/src/vm/vmar/vmar_impls/map.rs
```diff
@@ -310,48 +306,32 @@ impl<'a> VmarMapOptions<'a> {
...
-                        vmo.unwrap(),
-                        vmo_offset,
-                        is_writable_tracked,
-                    )?);
-                    (mapped_mem, Some(inode), None)
-                }
-                Mappable::IoMem(iomem) => (MappedMemory::Device, None, Some(iomem)),
+        let (mapped_mem, io_mem) = match mappable {
+            Some(Mappable::Vmo(vmo)) => {
+                let is_writable_tracked = if let Some(ref path) = path
```
vvvvsv, line 316: Maybe add a `debug_assert` here to ensure that the path matches the vmo?

```rust
if let Some(ref path) = path {
    debug_assert!(Arc::ptr_eq(&path.inode().page_cache().unwrap(), &vmo));
}
```

kernel/src/vm/vmar/vmar_impls/map.rs
lrh2000: This is a bit too long for the summary line. I removed ``` according to its [`Path`] ``` since it should be obvious.

The `Binds memory to ...` part is slightly reworded to align better with the existing `vmo` method:
```rust
    /// Binds a [`Vmo`] to the mapping.
    pub fn vmo(mut self, vmo: Arc<Vmo>) -> Self;
    /// Sets the [`Path`] of the mapping.
    pub fn path(mut self, path: Path) -> Self;

    /// Binds the file's [`Mappable`] object to the mapping and sets the
    /// [`Path`] of the mapping.
    pub fn mappable(mut self, file: &dyn FileLike) -> Result<Self>;
```

tatetian (approved):

PR #2944, opened 2026-01-30
Add Python regression tests on Asterinas NixOS

cqs21:


.github/workflows/test_nixos_minimal.yml
```diff
@@ -17,6 +17,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+      matrix:
+        test:
+          - name: hello
+            disk_size: 6144
+          - name: python
+            disk_size: 8192
```
StevenJiang1110: Why adding python test to test_nixos_minimal? This workflow intends to only run the most basic test(i.e., the hello test), it seems python test should be added to test_nixos_full?

cqs21: > Why adding python test to test_nixos_minimal?

These tests primarily focus on kernel functionality rather than NixOS compatibility, so they should be triggered for every PR that involves kernel changes. That's why I added them to test_nixos_minimal. Another reason is that it doesn't require additional payment.

test/nixos/tests/python/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    let testcases = std::fs::read_to_string("src/testcases.txt")?;
+    for line in testcases.lines() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with("#") {
+            continue;
+        }
+
+        let cmd = format!("python -m test -u all {}", line);
+        nixos_shell.run_cmd_and_expect(&cmd, "Result: SUCCESS")?;
```
tatetian: **Document the expected regrtest output format next to the check.** Neither the bare literal `"Result: SUCCESS"` nor the `python -m test` invocation explains *why* those two strings work together, and the next maintainer to touch this code (Python bump, regrtest rewrite, switch to `pytest`) has no in-file signal for what to verify. Please add a comment citing the exact CPython source that emits the string, and the state machine behind it, e.g.:

```rust
// Printed once by regrtest at the end of every run as `Result: {state}`:
//   https://github.com/python/cpython/blob/v3.12.12/Lib/test/libregrtest/main.py#L455-L456
// `state` is the literal "SUCCESS" iff no failure/env-change/interrupt/etc.
// flag is set; any failure produces a disjoint string ("FAILURE",
// "NO TESTS RAN, INTERRUPTED", ...), so this substring match is sufficient:
//   https://github.com/python/cpython/blob/v3.12.12/Lib/test/libregrtest/results.py#L55-L69
const RESULT_SUCCESS: &str = "Result: SUCCESS";

for testcase in TESTCASES.lines().map(str::trim)
    .filter(|l| !l.is_empty() && !l.starts_with('#'))
{
    let cmd = format!("python -m test -u all {testcase}");
    nixos_shell.run_cmd_and_expect(&cmd, RESULT_SUCCESS)?;
}
```

Two bonus hardenings worth considering while you're there:

1. **Anchor the match** so an unrelated CPython stdout line that happens to contain `Result: SUCCESS` in passing can't satisfy the check: `\nResult: SUCCESS\n` (the framework's substring matcher is literal, so embedding newlines is enough).
2. **Cross-check the exit code** so a crash after a partial "SUCCESS" print can't sneak through:

   ```rust
   nixos_shell.run_cmd_and_expect(
       &format!("python -m test -u all {testcase}; echo EXIT=$?"),
       "EXIT=0",
   )?;
   ```

The documentation is the important part — the hardenings are optional.

test/nixos/tests/python/src/testcases.txt
```diff
@@ -0,0 +1,499 @@
...
+# curl https://www.python.org/ftp/python/3.12.12/Python-3.12.12.tar.xz -o Python-3.12.12.tar.xz
+# tar xf Python-3.12.12.tar.xz
+# export PYTHONPATH=$PWD/Python-3.12.12/Lib
+# python -m test --list-tests
+
+test.test_asyncio.test_base_events -i test_create_server_stream_bittype
+test.test_asyncio.test_buffered_proto
+test.test_asyncio.test_context
+test.test_asyncio.test_eager_task_factory
+test.test_asyncio.test_events -i python -m test -u all -v test.test_asyncio.test_events -i test_ssl_connect_accepted_socket -i test_server_close -i test_get_event_loop_new_process -i test_get_loop -i test_create_server* -i test_create_ssl*
```
tatetian: **Almost certainly a copy-paste error; passes today by accident.** CI run 21575168043 shows this line completing in ~16 s and emitting `Result: SUCCESS`. That's because regrtest happily parses the splice: `-m test` is the match-pattern flag (matches every `test_*` name), `-v` is verbose, `-i python` ignores a non-existent pattern, and the duplicated `test.test_asyncio.test_events` positional is tolerated. So the command runs *something* and passes — but the effective match/verbose/ignore set is clearly not what the author meant. Collapse to the intended form:

```
test.test_asyncio.test_events -i test_ssl_connect_accepted_socket -i test_server_close -i test_get_event_loop_new_process -i test_get_loop -i test_create_server* -i test_create_ssl*
```

test/nixos/tests/python/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+//! review the documentation and see if should be updated accordingly.
+
+use nixos_test_framework::*;
+
+nixos_test_main!();
+
+#[nixos_test]
+fn python(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("curl https://www.python.org/ftp/python/3.12.12/Python-3.12.12.tar.xz -o Python-3.12.12.tar.xz")?;
+    nixos_shell.run_cmd("tar xf Python-3.12.12.tar.xz")?;
```
tatetian: **`curl` not CI-hardened; no integrity check.** Failure modes:

- No `-f`/`--fail`: 4xx/5xx writes the HTML error page as the tarball; `tar xf` then fails cryptically after a VM boot.
- No retries/timeout: one transient python.org blip fails the run.
- No checksum: a compromised mirror silently taints every run.

Minimal hardening:

```rust
nixos_shell.run_cmd(
    "curl -fsSL --retry 5 --retry-all-errors --connect-timeout 30 --max-time 300 \
     -o Python-3.12.12.tar.xz \
     https://www.python.org/ftp/python/3.12.12/Python-3.12.12.tar.xz"
)?;
nixos_shell.run_cmd_and_expect(
    "echo '<sha256-from-python.org>  Python-3.12.12.tar.xz' | sha256sum -c -",
    "OK",
)?;
```

Strongly preferred: drop the download entirely and use the SHA-verified tarball already in the Nix store by adding to `extra_config.nix`:

```nix
environment.etc."python-src.tar.xz".source = pkgs.python312.src;
```

then `cp /etc/python-src.tar.xz ./Python-3.12.12.tar.xz` inside the VM — air-gap-safe, no python.org CI dependency, no tarball/binary version skew.

test/nixos/tests/python/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+    let testcases = std::fs::read_to_string("src/testcases.txt")?;
+    for line in testcases.lines() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with("#") {
+            continue;
+        }
+
+        let cmd = format!("python -m test -u all {}", line);
+        nixos_shell.run_cmd_and_expect(&cmd, "Result: SUCCESS")?;
+    }
```
tatetian: **Fail-fast is wrong for a 499-entry suite.** `?` aborts on the first failure — PR authors see "failed at module #173" with no signal on the remainder, and one flaky entry masks every real regression below it. Also `starts_with("#")` should be `starts_with('#')`, and the inner `line` shadows itself. Fix:

```rust
let mut failed: Vec<String> = Vec::new();
for raw in TESTCASES.lines() {
    let testcase = raw.trim();
    if testcase.is_empty() || testcase.starts_with('#') {
        continue;
    }
    let cmd = format!("python -m test -u all {testcase}");
    if let Err(e) = nixos_shell.run_cmd_and_expect(&cmd, "== Tests result: SUCCESS ==") {
        println!("FAIL: {testcase}: {e}");
        failed.push(testcase.to_string());
    }
}
if !failed.is_empty() {
    return Err(/* summarize failed.len() / failed */);
}
```

.github/workflows/test_nixos_minimal.yml
```diff
@@ -17,6 +17,15 @@ jobs:
...
+
+    name: Test ${{ matrix.test.name }}
+    timeout-minutes: 60
```
tatetian: **Set per-matrix-entry timeouts so `hello` isn't given `python`'s budget.** CI run 21575168043 shows the python leg finishing in ~31 min and `hello` in ~6 min, so the 60-min job budget is adequate today. However, `hello` currently inherits the same 60 min even though it only needs a fraction of that, so a hang in `hello` burns the full hour before surfacing. Per-entry timeouts also document the expected runtimes:

```yaml
matrix:
  test:
    - name: hello
      disk_size: 6144
      timeout_minutes: 15
    - name: python
      disk_size: 8192
      timeout_minutes: 60
timeout-minutes: ${{ matrix.test.timeout_minutes }}
```

Additionally keep a step-level cap on the docker-run step (say `timeout-minutes: ${{ matrix.test.timeout_minutes - 5 }}`) so a hung VM fails at the step level with a clearer error than a silent job timeout.

test/nixos/tests/python/extra_config.nix
```diff
@@ -0,0 +1,5 @@
+{ config, lib, pkgs, ... }:
+
+{
+  environment.systemPackages = with pkgs; [ python312 ];
+}
```
tatetian: **Potential version mismatches between tarball and nixpkgs binary.** `main.rs` pins 3.12.12 source, but `python312` is whatever the channel ships — a future bump to 3.12.13 silently mixes a stdlib from one patch level with a C binary at another. Fix — guarantee they match by construction:

```nix
{ config, lib, pkgs, ... }:

let
  py = pkgs.python312;
in
{
  environment.systemPackages = [ py ];
  # Make the exact matching source tree available without a download.
  environment.etc."python-src.tar.xz".source = py.src;
}
```

then inside the VM:

```rust
nixos_shell.run_cmd("cp /etc/python-src.tar.xz Python.tar.xz && tar xf Python.tar.xz")?;
```

(Unused `config` and `lib` in the lambda binding are stylistic and can stay or go.)

test/nixos/tests/python/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+//! So whenever you modify the test suite,
+//! review the documentation and see if should be updated accordingly.
+
+use nixos_test_framework::*;
+
+nixos_test_main!();
+
+#[nixos_test]
+fn python(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("curl https://www.python.org/ftp/python/3.12.12/Python-3.12.12.tar.xz -o Python-3.12.12.tar.xz")?;
```
tatetian: [**naming-constants**](https://asterinas.github.io/book/to-contribute/coding-guidelines/naming.html): `"3.12.12"` alone appears 4 times across URL/tarball/`PYTHONPATH`; `"Result: SUCCESS"` and `"python -m test -u all"` are also bare literals. Fix:

```rust
// `concat!` only accepts string literals, so wrap the version in a
// macro that expands to one. No extra crate needed.
macro_rules! python_version { () => { "3.12.12" }; }

const PYTHON_VERSION: &str = python_version!();
const PYTHON_SRC_DIR: &str = concat!("Python-", python_version!());
const PYTHON_TARBALL: &str = concat!("Python-", python_version!(), ".tar.xz");
const RESULT_MARKER: &str = "== Tests result: SUCCESS ==";
const RUN_PREFIX: &str = "python -m test -u all";

nixos_shell.run_cmd(&format!(
    "curl -fsSL --retry 5 -o {PYTHON_TARBALL} \
     https://www.python.org/ftp/python/{PYTHON_VERSION}/{PYTHON_TARBALL}"
))?;
```

Bumping the Python version now touches the single `python_version!` macro body; the derived consts update at compile time.

tatetian (commented):
Thanks for adding the Python test suite. Good to see that a large portion of Python tests can run well.

.github/workflows/test_nixos_full.yml
```diff
@@ -108,7 +108,8 @@ jobs:
               { "name": "multimedia", "disk_size": 8192 },
               { "name": "communication", "disk_size": 8192 },
               { "name": "file-management-and-terminal", "disk_size": 8192 },
-              { "name": "ai-and-machine-learning", "disk_size": 10240, "timeout_minutes": 45 }
+              { "name": "ai-and-machine-learning", "disk_size": 10240, "timeout_minutes": 45 },
+              { "name": "python", "disk_size": 8192, "timeout_minutes": 60 }
```
tatetian: Technically, Python belongs to the "Development Tools" category. So I suggest name this test suite `development-tools-python`.

test/nixos/tests/python/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    if !failed_tests.is_empty() {
+        println!("=== Failed Python 3 regression tests ===");
+        for test in &failed_tests {
+            println!("  - {test}");
+        }
+        println!("======================================");
+
+        return Err(Error::ExpectMismatch {
+            expected: "All selected Python 3 regression tests to pass".to_string(),
```
tatetian: The name of this enum is confusing. I think that your intention is: `run_cmd_and_expect_regex` fails because the actual string does not match the pattern. But in the test case, you manually construct this error value. Looking at it out of the `run_cmd_and_expect_regex` context, the enum variant name reads like "I expect here to encounter a mismatch", which has an opposite meaning. So I suggest that you think of a different name such as `UnexpectedOutput`.

tatetian (commented):
Thanks for the revision. The code looks mostly good. Just a few minor issues.

test/nixos/tests/python/src/main.rs
cqs21: > The name of this enum is confusing. 

I split the semantics: output matching failures are now reported as `UnexpectedOutput`, while suite-level failures use a separate `TestFailure` variant.

test/nixos/common/framework/src/lib.rs
```diff
@@ -38,11 +38,13 @@ pub enum Error {
     /// The PTY/process produced an underlying rexpect failure.
     Pty(rexpect::error::Error),
     /// The command output reached the prompt but did not contain the expected substring.
-    ExpectMismatch { expected: String, got: String },
+    UnexpectedOutput { expected: String, got: String },
```
tatetian, line 42: I am wondering if you want to save the entire output string in `got`, which could be huge.

test/nixos/common/framework/src/lib.rs
```diff
@@ -38,11 +38,13 @@ pub enum Error {
...
     /// The framing marker was missing or unparsable.
     Protocol { reason: String, got: String },
     /// A command exited with a non-zero status.
     NonZeroExit { exit_status: i32, output: String },
+    /// A test-specific expectation was not met.
+    TestFailure { reason: String },
```
tatetian: This name is too vague; so is the description. I think one way to make it more useful and specific is:

```rust
pub enum Error {
    //..
    /// Polling for an expected state timed out.
    Timeout {
        expected: String,
        got: String,
        timeout: Duration,
    },
    /// An error collection that aggregates multiple errors.
    ///
    /// For each collected error, it keeps a name (in `String`)
    /// and a detailed error info (in `Error`).
    Aggregated {
        summary: String,
        collections: Vec<(String, Error)>
    },
}
```

tatetian (approved):

PR #2945, opened 2026-01-31
Integrate xfstests framework for ext2

Fischer0522:
This PR integrates the **xfstests** framework into Asterinas and our CI pipeline (related to #2856). 

### Key Changes
1. Added and formatted two disk images as `ext2`. Updated the root `Makefile` and `tool/qemu_args.sh` to support these new devices.
2. Packaged `xfstests` using Nix to streamline its deployment within the `test/initramfs` environment.
3. Integrated `xfstests` execution into the GitHub Actions workflow.

### Configuration
The framework supports both **whitelist** and **blocklist** execution modes. Since the current `ext2` implementation passes approximately 50 out of several hundred tests, using the **whitelist** approach for stable CI runs is recommended.

Since some `xfstests` cases take a lot time, two whilelists are provided, The first one named `quick.list` for daily CI workflows, the second file named `full.list` is exclusively scheduled to run during nighttime hours, currently set at 02:00 CST every day, following the same operational arrangement as benchmark_x86 and benchmark_x86_tdx.

### TODO
- [x] Remove the temporary `fix.sh` once all upstream dependencies are merged.
- [ ] Extend support to additional file systems (e.g., `initramfs`, `exFAT`, etc.).


tatetian:
Remember to clear the commit history (e.g., no intermediate commits and well-written commit titles)

.github/workflows/test_x86.yml
```diff
@@ -123,11 +124,29 @@ jobs:
...
           netdev: ${{ matrix.netdev || 'tap' }}
           scheme: ${{ matrix.scheme }}
+          mem: ${{ matrix.mem }}
           extra_blocklists: ${{ matrix.extra_blocklists }}
           conformance_test_suite: 'gvisor'
           conformance_test_workdir: ${{ matrix.conformance_test_workdir }}
           boot_protocol: ${{ matrix.boot_protocol || 'linux-efi-handover64' }}
 
+  xfstests:
+    runs-on: ubuntu-4-cores-150GB-ssd
```
tatetian: **Try stick to the free-tier Github runner for CI tests triggered by every PR.** Every CI test would love to have a runner with a larger disk. But this brings larger payment bills. Could you try to use the default, free-tier runner?

For nightly CI tests, it is ok to use the paid Github runners.

.github/actions/test/action.yml
```diff
@@ -45,6 +45,12 @@ inputs:
   boot_protocol:
     description: 'Boot protocol (linux-efi-handover64/linux-efi-pe64/linux-legacy32/multiboot/multiboot2)'
     required: false
+  mem:
+    description: 'Guest memory size'
+    required: false
```
tatetian, line 55: This parameter belongs to the virtualization settings

.github/actions/test/action.yml
```diff
@@ -45,6 +45,12 @@ inputs:
   boot_protocol:
     description: 'Boot protocol (linux-efi-handover64/linux-efi-pe64/linux-legacy32/multiboot/multiboot2)'
     required: false
```
tatetian, line 50: [**Leave code cleaner than you found it:**](https://asterinas.github.io/book/to-contribute/coding-guidelines/how-guidelines-are-written.html#continuous-improvement). Although not part of the original goal of this PR, I suggest reordering and regrouping the test parameters:

```yml
inputs:
  # General Configurations
  auto_test:
    description: 'Test type (general, osdk, boot, regression, conformance)'
    required: true
  arch:
    description: 'Architecture (x86_64/riscv64/loongarch64)'
    required: false
  release:
    description: 'Whether to run in release mode'
    required: false
  scheme:
    description: 'OSDK scheme (e.g., default/microvm/iommu/tdx)'
    required: false
  
  # VM Setup
  boot_protocol:
    description: 'Boot protocol (linux-efi-handover64/linux-efi-pe64/linux-legacy32/multiboot/multiboot2)'
    required: false
  enable_kvm:
    description: 'Enable KVM acceleration'
    required: false
  smp:
    description: 'Number of CPUs'
    required: false
  netdev:
    description: 'Network device type (user/tap)'
    required: false
  intel_tdx:
    description: 'Enable Intel TDX support'
    required: false
  
  # Conformance Test
  conformance_test_suite:
    description: 'Selected conformance test suite'
    required: false
  conformance_test_workdir:
    description: 'Conformance test working directory'
    required: false
  extra_blocklists:
    description: 'Extra blocklists directories'
    required: false
```

test/initramfs/src/conformance/xfstests/short.list
```diff
@@ -0,0 +1,50 @@
+generic/001
```
tatetian, line 1: The opposite of "full" would be "short", "small", "core", or "essential", not "quick". So rename this file to `short.list`.

test/initramfs/src/conformance/run_conformance_test.sh
```diff
@@ -7,6 +7,7 @@ set -e
 CONFORMANCE_TEST_SUITE=${CONFORMANCE_TEST_SUITE:-ltp}
 LTP_DIR=/opt/ltp
 GVISOR_DIR=/opt/gvisor
+XFSTESTS_DIR=/xfstests
```
tatetian: Other tests are put under `opt`, why `xfstests` need to sit in the root?

.github/actions/test/action.yml
```diff
@@ -45,6 +45,12 @@ inputs:
...
+  mem:
+    description: 'Guest memory size'
+    required: false
+  xfstests_runlist:
+    description: 'Run list file for xfstests'
+    required: false
```
tatetian, line 58: **This parameter does not blend well with other parameters:** I suggest adding a new parameter group:

```yml
# Xfstest-specific parameters
xfstests_disk_size:
    description: 'The disk size for xfstests'
    required: false
xfstests_runlist:
    description: 'Run list file for xfstests'
    required: false
```

In addition to `xfstests_runlist`, I think you would need the disk size parameter so that a smaller disk can be used on the free-tier runner, while a larger disk used on paid runners.

Makefile
```diff
@@ -99,6 +100,8 @@ ENABLE_CONFORMANCE_TEST := true
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="CONFORMANCE_TEST_SUITE=$(CONFORMANCE_TEST_SUITE)"
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="CONFORMANCE_TEST_WORKDIR=$(CONFORMANCE_TEST_WORKDIR)"
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="EXTRA_BLOCKLISTS_DIRS=$(EXTRA_BLOCKLISTS_DIRS)"
+# XFSTESTS_RUNLIST is only consumed when CONFORMANCE_TEST_SUITE=xfstests.
+CARGO_OSDK_BUILD_ARGS += --kcmd-args="XFSTESTS_RUNLIST=$(XFSTESTS_RUNLIST)"
```
tatetian, line 115: Only add this parameter to the arguments when the conformance test suite is xfstest. It is confusing to see this environment variable even when the xfstest is not built into the initramfs.

test/initramfs/nix/conformance/xfstests/default.nix
```diff
@@ -0,0 +1,96 @@
...
+  ];
+
+  sbinDeps = with pkgs; [ util-linux kmod xfsprogs e2fsprogs ];
+
+  runtimePath = lib.makeBinPath runtimeDeps + ":"
+    + lib.concatMapStringsSep ":" (package: "${package}/sbin") sbinDeps
+    + ":/bin:/usr/bin:/sbin:/usr/sbin";
+
+  runScript = pkgs.writeTextFile {
+    name = "run_xfstests.sh";
```
tatetian: Other conformance tests put their runner scripts under `src/conformance/xxx/`. You got to have a reason to break this established convention. In fact, the entire `src/conformance/xfstest/` directory is missing. If you think the convention is sub-optimal, submit a PR to improve the status quo. Otherwise, respect the established convention for the sake of consistency.

tatetian (commented):
These comments are written by me manually.

test/initramfs/nix/conformance/xfstests/default.nix
```diff
@@ -0,0 +1,96 @@
...
+      #!${pkgs.bash}/bin/bash
+      export PATH=${runtimePath}
+      cd /xfstests
+
+      # Mount xfstests images if they exist
+      [ -b /dev/vdc ] && mount -t ext2 /dev/vdc /xfstests/test
+      [ -b /dev/vdd ] && mount -t ext2 /dev/vdd /xfstests/scratch
+
+      RUNLIST_FILE=""
+      args=()
```
tatetian: **Silent CI pass when the test setup is broken** (blocker): The generated `run_xfstests.sh` has no `set -e -o pipefail`. The two mounts use the `[ -b /dev/vdN ] && mount ...` idiom, which only short-circuits when the device is missing — when the device is present but `mount` itself fails (Asterinas's ext2 driver rejecting an option, image corruption, fs detection mismatch), the `&&` evaluates to false, the script keeps going, and `./check` runs against the unmounted initramfs ramdisk dirs at `/xfstests/test` and `/xfstests/scratch`. Because the project's pass detection is `tail qemu.log | grep -q "^All conformance tests passed."` (`Makefile:317`), and `run_conformance_test.sh` only checks `run_xfstests.sh`'s exit status, an `./check` that runs against unmounted dirs (or against an empty test set) can still print the success line — turning a broken setup into a green CI run.

Fix:

```bash
#!${pkgs.bash}/bin/bash
set -euo pipefail

export PATH=${runtimePath}
cd /xfstests

for entry in "vdc:/xfstests/test:TEST_DEV" "vdd:/xfstests/scratch:SCRATCH_DEV"; do
    dev="/dev/${entry%%:*}"; rest="${entry#*:}"; mnt="${rest%%:*}"; role="${rest##*:}"
    if [ ! -b "$dev" ]; then
        echo "Expected $dev to be a block device for xfstests $role" >&2
        exit 1
    fi
    if ! mount -t ext2 "$dev" "$mnt"; then
        echo "Failed to mount $dev on $mnt" >&2
        exit 1
    fi
    if ! mountpoint -q "$mnt"; then
        echo "$mnt is not a mountpoint after mount(8) succeeded" >&2
        exit 1
    fi
done
```

test/initramfs/nix/conformance/xfstests/local.config
```diff
@@ -0,0 +1,15 @@
+export FSTYP=ext2
+
+export TEST_DEV=/dev/vdc
+export TEST_DIR=/xfstests/test
+
+export SCRATCH_DEV=/dev/vdd
+export SCRATCH_MNT=/xfstests/scratch
+
+export MOUNT_OPTIONS="-o noload,noacl"
+export EXT_MOUNT_OPTIONS="-o noload,noacl"
```
tatetian: **`noload` is an ext3/ext4-only mount option** (significant): `noload` disables journal replay; ext2 has no journal, so the option is semantically meaningless on ext2. On Linux the kernel typically silently ignores unknown ext2 options, but Asterinas's ext2 driver may reject it — in which case every test fails at mount time and the `[ -b /dev/vdN ] && mount` chain in `run_xfstests.sh` swallows the failure (see the comment on `default.nix` above). The `mkfs.ext2` calls in `test/initramfs/Makefile:179,184` already create journal-less ext2, so there's nothing to suppress. Drop it.

Fix:

```ini
export MOUNT_OPTIONS="-o noacl"
export EXT_MOUNT_OPTIONS="-o noacl"
```

test/initramfs/nix/conformance/xfstests/local.config
```diff
@@ -0,0 +1,15 @@
+export FSTYP=ext2
+
+export TEST_DEV=/dev/vdc
+export TEST_DIR=/xfstests/test
+
+export SCRATCH_DEV=/dev/vdd
+export SCRATCH_MNT=/xfstests/scratch
```
tatetian: **Hard-coded `/dev/vdc`/`/dev/vdd` couples this file to `qemu_args.sh` drive ordering** (significant): On the q35 path the fixed PCI addresses (`0x6` ext2 → vda, `0x7` exfat → vdb, `0x8` xfstests test → vdc, `0x9` xfstests scratch → vdd) make the mapping deterministic, but on the `microvm` path device naming is purely attachment order and any future reorder of the `-device virtio-blk-device,...` lines silently shifts these names. The QEMU drives already carry stable serials (`serial=vxfstest`/`serial=vxfsscratch` — `tools/qemu_args.sh:162-163,167-168`); use them via `/dev/disk/by-id/virtio-…` so `local.config` is decoupled from the bus topology.

Fix in `run_xfstests.sh` (the embedded script in `default.nix`), and let `local.config` inherit:

```bash
TEST_DEV=$(readlink -f /dev/disk/by-id/virtio-vxfstest)
SCRATCH_DEV=$(readlink -f /dev/disk/by-id/virtio-vxfsscratch)
export TEST_DEV SCRATCH_DEV
mount -t ext2 "$TEST_DEV" /xfstests/test     || exit 1
mount -t ext2 "$SCRATCH_DEV" /xfstests/scratch || exit 1
```

…and remove the `TEST_DEV`/`SCRATCH_DEV` lines from `local.config` (xfstests `check` honors environment overrides).

test/initramfs/src/conformance/xfstests/block.list
tatetian, line 1: **`block.list` is shipped to the image but never read** (significant): The `configFiles` fileset in `default.nix` line 6-8 globs every `*.list`, so `block.list` ends up at `/xfstests/block.list` in the initramfs — but neither `run_xfstests.sh`, `run_conformance_test.sh`, nor `local.config` references it. The PR description claims a "blocklist execution mode" alongside the runlist mode, yet only the `-R` path is wired. From a testing standpoint this is a regression vector: a contributor adding a test to `quick.list`/`full.list` may unwittingly pick one already known to hang, and CI then burns the entire 30-minute timeout with no signal about which test wedged. Either wire it in via xfstests' native exclude flag, or delete the file until it's used.

Fix (wire it up) in `run_xfstests.sh`, before the `./check` invocation:

```bash
if [ -f /xfstests/block.list ]; then
    args=("-E" "/xfstests/block.list" "${args[@]}")
fi
```

…and add a one-line CI assertion (e.g. in the workflow or in `default.nix`) that `quick.list ∩ block.list == ∅`. Or remove `block.list` and the surrounding "blocklist mode" framing from the PR description.

test/initramfs/src/conformance/xfstests/full.list
tatetian, line 1: **`full.list` is `quick.list` + 4 tests; the nightly workflow earns very little extra coverage** (significant): `diff <(sort quick.list) <(sort full.list)` shows the only additions are `generic/074`, `generic/129`, `generic/269`, `generic/707`. The `test_xfstests_full.yml` job is timed at 90 min on the same paid 4-core runner as the per-PR job, but it's almost the same workload. Either expand `full.list` to actually exercise more coverage — the obvious candidates are xfstests' `auto` group filtered by `block.list`, or the broader generic/* set — so the extra runner time pays for itself, or drop the separate nightly workflow until there's enough delta to justify it.

Fix: regenerate `full.list` from `./check -n -g auto` minus `block.list`, capture the result, and prune to whatever fits in 90 min. Alternatively, delete `test_xfstests_full.yml` and `full.list` until coverage actually expands.

Makefile
tatetian: **`XFSTESTS_RUNLIST` is appended to kcmd-args even for ltp/gvisor** (significant): The default is set at top level and the `--kcmd-args` line is unconditional inside `ifeq ($(AUTO_TEST), conformance)`. So every conformance run — including ltp and gvisor — boots with `XFSTESTS_RUNLIST=/xfstests/quick.list` on the kernel command line. It's silently ignored by `run_conformance_test.sh` for non-xfstests suites, but it pollutes the cmdline and means a user who passes `XFSTESTS_RUNLIST=foo` while running ltp gets no error. The companion comment "only consumed when CONFORMANCE_TEST_SUITE=xfstests" admits the smell — gate the line instead of documenting around it.

Fix:

```make
ifeq ($(AUTO_TEST), conformance)
ENABLE_CONFORMANCE_TEST := true
CARGO_OSDK_BUILD_ARGS += --kcmd-args="CONFORMANCE_TEST_SUITE=$(CONFORMANCE_TEST_SUITE)"
CARGO_OSDK_BUILD_ARGS += --kcmd-args="CONFORMANCE_TEST_WORKDIR=$(CONFORMANCE_TEST_WORKDIR)"
CARGO_OSDK_BUILD_ARGS += --kcmd-args="EXTRA_BLOCKLISTS_DIRS=$(EXTRA_BLOCKLISTS_DIRS)"
ifeq ($(CONFORMANCE_TEST_SUITE), xfstests)
CARGO_OSDK_BUILD_ARGS += --kcmd-args="XFSTESTS_RUNLIST=$(XFSTESTS_RUNLIST)"
endif
CARGO_OSDK_BUILD_ARGS += --init-args="/opt/run_conformance_test.sh"
```

test/initramfs/Makefile
```diff
@@ -162,6 +173,16 @@ $(EXFAT_IMAGE):
...
 
+$(XFSTESTS_TEST_IMAGE):
+	@mkdir -p $(BUILD_DIR)
+	@fallocate -l 12G $(XFSTESTS_TEST_IMAGE)
+	@mkfs.ext2 $(XFSTESTS_TEST_IMAGE)
+
+$(XFSTESTS_SCRATCH_IMAGE):
+	@mkdir -p $(BUILD_DIR)
+	@fallocate -l 12G $(XFSTESTS_SCRATCH_IMAGE)
+	@mkfs.ext2 $(XFSTESTS_SCRATCH_IMAGE)
```
tatetian, line 186: **12 GiB × 2 reserves 24 GiB of physical disk; size is unjustified and rules are non-atomic** (significant): `fallocate -l 12G` reserves real extents (it is *not* sparse), so each CI run consumes 24 GiB of the runner's 150 GiB SSD even though `quick.list`'s 50 generic tests likely need ≪ 1 GiB of fs space each. There's no comment explaining why 12 G was chosen — please name the test that needs it, or drop both to 4 G (or smaller). Independently, the rules `fallocate` and `mkfs.ext2` directly into the final filename with no `-F`; if a build is interrupted between the two, the next `make` sees the file exists and considers the rule satisfied, leaving an unformatted image. Build to a `.tmp` and rename atomically.

Fix:

```make
$(XFSTESTS_TEST_IMAGE):
	@mkdir -p $(BUILD_DIR)
	@fallocate -l 4G $@.tmp
	@mkfs.ext2 -F $@.tmp
	@mv $@.tmp $@

$(XFSTESTS_SCRATCH_IMAGE):
	@mkdir -p $(BUILD_DIR)
	@fallocate -l 4G $@.tmp
	@mkfs.ext2 -F $@.tmp
	@mv $@.tmp $@
```

(Adjust the size after measuring the actual high-water mark of `quick.list`/`full.list`.)

tools/qemu_args.sh
```diff
@@ -15,14 +15,21 @@
...
 VHOST=${VHOST:-"off"}
 VSOCK=${VSOCK:-"off"}
 NETDEV=${NETDEV:-"user"}
 CONSOLE=${CONSOLE:-"hvc0"}
 
+ATTACH_XFSTESTS_IMAGES="false"
+if [ "${ENABLE_CONFORMANCE_TEST:-"false"}" = "true" ] && \
+   [ "${CONFORMANCE_TEST_SUITE:-"ltp"}" = "xfstests" ]; then
+    ATTACH_XFSTESTS_IMAGES="true"
+fi
```
tatetian: **`ATTACH_XFSTESTS_IMAGES` documented as a user-controllable env var but unconditionally overwritten** (minor): The header docblock lists it alongside `OVMF`, `BOOT_METHOD`, etc., as a configurable variable, but the script ignores any user-supplied value and derives it solely from `ENABLE_CONFORMANCE_TEST` + `CONFORMANCE_TEST_SUITE`. So setting `ATTACH_XFSTESTS_IMAGES=true` directly has no effect. The new entry is also the only one that ends in a period and runs ~190 columns while every neighbor is a short semicolon-terminated phrase — the inconsistency reads like the entry was meant to be a controllable flag but ended up as derived state.

Fix: drop the line from the header docblock (it's an internal computed variable now), or honor the user value with `ATTACH_XFSTESTS_IMAGES=${ATTACH_XFSTESTS_IMAGES:-false}` and only auto-flip to `true` when unset.

test/initramfs/src/conformance/run_conformance_test.sh
```diff
@@ -20,9 +21,20 @@ elif [ "${CONFORMANCE_TEST_SUITE}" == "gvisor" ]; then
...
+    fi
+    # shellcheck disable=SC2086
+    if ! "${XFSTESTS_DIR}/run_xfstests.sh" ${RUNLIST_ARG}; then
+        echo "Error: xfstests failed." >&2
+        exit 3
+    fi
 else
     echo "Error: Unknown test suite '${CONFORMANCE_TEST_SUITE}'." >&2
-    exit 3
+    exit 255
```
tatetian: **Word-split `RUNLIST_ARG` and silent exit-code shift** (minor): Two issues, same chunk: (1) `RUNLIST_ARG="-R ${XFSTESTS_RUNLIST}"` then unquoted expansion forces a `# shellcheck disable=SC2086` and breaks if the path ever contains whitespace — the value comes from a Make var threaded through `--kcmd-args` and a GitHub Action input, both user-settable surfaces; (2) the unknown-suite branch shifted from `exit 3` to `exit 255`, which now collides with the xfstests-failed branch's `exit 3` (so the unique sentinel for "unknown suite" was lost) and 255 is conventionally reserved for shell errors / signal-killed.

Fix (rewriting the elif and reverting the unknown-suite code):

```sh
elif [ "${CONFORMANCE_TEST_SUITE}" == "xfstests" ]; then
    echo "Running xfstests..."
    if [ -n "${XFSTESTS_RUNLIST}" ]; then
        set -- -R "${XFSTESTS_RUNLIST}"
    else
        set --
    fi
    if ! "${XFSTESTS_DIR}/run_xfstests.sh" "$@"; then
        echo "Error: xfstests failed." >&2
        exit 4
    fi
else
    echo "Error: Unknown test suite '${CONFORMANCE_TEST_SUITE}'." >&2
    exit 3
fi
```

(Pick whatever fresh code you like for "xfstests failed"; just keep "unknown suite" stable so any external scripts greping for `exit 3` still work.)

test/initramfs/src/conformance/run_conformance_test.sh
```diff
@@ -20,9 +21,20 @@ elif [ "${CONFORMANCE_TEST_SUITE}" == "gvisor" ]; then
         echo "Error: gVisor syscall tests failed." >&2
         exit 2
     fi
+elif [ "${CONFORMANCE_TEST_SUITE}" == "xfstests" ]; then
```
tatetian: **`==` in `[ ]` under `#!/bin/sh`** (minor): POSIX `test`/`[` only specifies `=`. `==` works in busybox `ash` and bash but is non-portable. The two pre-existing branches already had this bug, but the PR perpetuates it in a third place. Either fix all three (replace `==` with `=`) or change the shebang to `#!/bin/bash`.

Fix:

```diff
-if [ "${CONFORMANCE_TEST_SUITE}" == "ltp" ]; then
+if [ "${CONFORMANCE_TEST_SUITE}" = "ltp" ]; then
...
-elif [ "${CONFORMANCE_TEST_SUITE}" == "gvisor" ]; then
+elif [ "${CONFORMANCE_TEST_SUITE}" = "gvisor" ]; then
...
-elif [ "${CONFORMANCE_TEST_SUITE}" == "xfstests" ]; then
+elif [ "${CONFORMANCE_TEST_SUITE}" = "xfstests" ]; then
```

test/initramfs/nix/conformance/xfstests/default.nix
```diff
@@ -0,0 +1,96 @@
...
+          *)
+            args+=("$1")
+            shift
+            ;;
+        esac
+      done
+
+      for arg in "$@"; do
+        args+=("$arg")
+      done
```
tatetian: **Trailing `for arg in "$@"` is dead in the no-`--` path and over-clever in the `--` path** (minor): When `--` is never seen, the `while` loop exhausts `$@` via `shift`, so the trailing `for` is a no-op. When `--` is seen, the loop's `break` leaves the post-`--` args in `$@` and the `for` forwards them. Either case is better expressed inline by appending directly inside the `--` case, which makes the contract explicit.

Fix:

```bash
        --)
          shift
          args+=("$@")
          break
          ;;
```

…and delete the trailing `for arg in "$@"; do args+=("$arg"); done`.

test/initramfs/nix/conformance/xfstests/default.nix
tatetian, line 1: **xfstests files are copied through the package chain rather than referenced via the Nix store** (minor): The plumbing is `pkgs.xfstests/lib/xfstests/*` → this derivation's `$out/xfstests/` (cp #1) → `conformance/default.nix:32` `cp -r ${xfstests}/xfstests/* $out/xfstests/` into `conformance.package` (cp #2) → `initramfs.nix:57` `cp -r "${conformance.package}"/* $out/` into the initramfs (cp #3). Plus the closure walk at `initramfs.nix:77-82` materializes `pkgs.xfstests` into `$out/nix/store/` because the package isn't in the `pkg_path` dedup set. Each pass `cp -r`'s the same xfstests tree, inflating the cpio. For xfstests specifically you can short-circuit the chain by branching in `initramfs.nix` (`testSuite == "xfstests"`) and copying directly from `${conformance.xfstests}/xfstests/`, skipping the `conformance.package` indirection.

Fix sketch (in `initramfs.nix`, replacing the unconditional copy of conformance.package for xfstests):

```nix
${lib.optionalString (conformance != null && conformance.testSuite != "xfstests") ''
  cp -r "${conformance.package}"/opt/* $out/opt/
''}
${lib.optionalString (conformance != null && conformance.testSuite == "xfstests") ''
  cp -r "${conformance.package}"/opt/* $out/opt/
  mkdir -p $out/xfstests/{test,scratch}
  cp -r "${conformance.xfstests}/xfstests/"* $out/xfstests/
''}
```

(Then drop the xfstests block from `conformance/default.nix` since `initramfs.nix` does the work.)

.github/workflows/test_xfstests_full.yml
tatetian, line 1: **Nightly job duplicates the per-PR job and has no failure notification** (minor): Two issues: (1) the runner block (`runs-on`, container image, `--device=/dev/kvm --privileged`, the call to `./.github/actions/test`) is byte-for-byte identical to the new `xfstests:` job in `test_x86.yml:133-148` except for the runlist file and the timeout — when the container image bumps, one will lag behind the other; (2) on failure the workflow turns red on the Actions tab and that's it: no Slack/email/issue, no `if: failure()` step. For a nightly suite a regression can sit unnoticed for days. Add an `if: failure()` step that opens or updates a tracking issue, and consider extracting the shared bits into a `workflow_call` reusable workflow.

Fix sketch (notification only):

```yaml
      - name: Run xfstests full.list
        uses: ./.github/actions/test
        with:
          auto_test: 'conformance'
          conformance_test_suite: 'xfstests'
          release: true
          mem: '12G'
          xfstests_runlist: '/xfstests/full.list'
      - name: Open issue on failure
        if: failure()
        uses: actions/github-script@v7
        with:
          script: |
            const title = `xfstests-full failure on ${context.payload.repository.default_branch} (${context.runId})`;
            await github.rest.issues.create({
              owner: context.repo.owner,
              repo: context.repo.repo,
              title,
              body: `Run: ${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`,
              labels: ['xfstests-nightly']
            });
```

.github/workflows/test_x86.yml
```diff
@@ -123,11 +124,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run xfstests
+        uses: ./.github/actions/test
+        with:
+          auto_test: 'conformance'
+          conformance_test_suite: 'xfstests'
+          release: true
+          mem: '12G'
+          xfstests_runlist: '/xfstests/quick.list'
```
tatetian: **No per-test progress logging — a wedged test consumes the full 30-minute budget silently** (minor): xfstests `check` has no built-in per-test timeout and the wrapper doesn't add one. If a single test wedges (deadlock, livelock, busy-loop on a missing syscall), the runner is killed at `timeout-minutes: 30` with no indication of which test was stuck. Pipe `./check`'s output through a per-line wall-clock prefix so the qemu log makes the stuck test obvious — and consider adding a per-test watchdog.

Fix in `run_xfstests.sh`'s tail:

```bash
./check "${args[@]}" 2>&1 | awk '{ printf "[%s] %s\n", strftime("%H:%M:%S"), $0; fflush(); }'
```

(`gawk` is already in `runtimeDeps`. For a stricter watchdog, wrap each test in `timeout 120s` via xfstests' `TIMEOUT` env var if supported by your xfstests version, or add a `kill-after` parent process.)

tatetian (commented):
The following comments are generated by my coding agent using the `pr-review` skill. Take them with a grain of salt.

test/initramfs/src/conformance/xfstests/short.list
Fischer0522: The term "short" makes sense, but I feel it fails to clarify the reason behind why it is labeled short. Meanwhile, the `xfstests` also maintains a test group that is named `quick`. I'm ok with `short.list`, but `quick` aligns with `xfstests`.

.github/workflows/test_x86.yml
Fischer0522: I added a new configuration parameter named `XFSTESTS_DISK_SIZE`. However, if the actual disk size is smaller than 5GB, some simple test cases (related to `fallocate`/`truncate`) must be removed from the `short.list`.

.github/workflows/test_xfstests_full.yml
Fischer0522: Other nightly jobs (`benchmark_x86`、`benchmark_x86_tdx`、`test_x86_tdx`) also do not have any failure notification, so I think that this situation is actually outside the scope of the current PR.

test/initramfs/nix/conformance/xfstests/local.config
Fischer0522: Currently, `/sys/block/` or `/dev/disk/by-id/` are unsupported yet — sysfs exists but block devices
aren't wired in, and virtio-blk GetId is unimplemented. So I pass device paths via `kcmd-args` (`XFSTESTS_TEST_DEV`/`XFSTESTS_SCRATCH_DEV`) as a workaround.

test/initramfs/nix/conformance/xfstests/default.nix
Fischer0522: Same duplication exists for `ltp`/`gvisor`/`kselftest` — it's a general conformance build issue. I think it should be solved separately.

test/initramfs/src/conformance/xfstests/block.list
Fischer0522: `block.list` contains test cases that cause kernel panic or hang, breaking test progress. With the new feature, we can run `./run_xfstests -E block.list` once to identify newly passed tests without being blocked by kernel issues, so I keep it here.

I added it into `./run_xfstests.sh` to prevent a contributor adding a test to quick.list/full.list from unwittingly picking one already known to hang.

Makefile
```diff
@@ -42,6 +42,10 @@ GDB_PROFILE_INTERVAL ?= 0.1
 # mode using the kernel command line.
 # Here are the options for the auto test feature.
 AUTO_TEST ?= none
+XFSTESTS_RUNLIST ?= /opt/xfstests/short.list
+XFSTESTS_DISK_SIZE ?= 12G
```
tatetian: All `XFSTEST_*` variables should be moved to near the end of this code  block, after `EXTRA_BLOCKLISTS ?= ""`. And a comment line should be inserted before them to note the beginning of the set of parameters for xfstests.

tatetian (commented):
Thanks for the revision. The code looks mostly great. Just one nit and you need to rebase to the latest main branch to fix the conflict. Thanks!

Makefile
Fischer0522: Done.

PR #2950, opened 2026-02-03
Enable `cvm_guest` feature by default in aster-kernel

Hsy-Intel:


lrh2000:
Given that `make kernel` will enable `cvm_guest` for both the kernel and for the OSTD, could you please also ensure `make kernel NO_DEFAULT_FEATURES=1` disables `cvm_guest` for both the kernel and OSTD?

tatetian:
> Given that `make kernel` will enable `cvm_guest` for both the kernel and for the OSTD, could you please also ensure `make kernel NO_DEFAULT_FEATURES=1` disables `cvm_guest` for both the kernel and OSTD?

From the recent discussion, it looks that having the `cmv_guest` feature enabled is bothering you for some reason. May I ask why?

lrh2000:
> From the recent discussion, it looks that having the `cmv_guest` feature enabled is bothering you for some reason. May I ask why?

I just cannot figure out a valid reason why we should leave the inconsistency between OSTD and the kernel. For example, why would we want a build that does the following things simultaneously:
 - disables `cvm_guest` in the kernel,
 - enables `cvm_guest` in the OSTD.

Do you have a valid reason for wanting this? It looks like you're enabling or disabling a CONFIG_XXX option in the Linux kernel, but only for files under certain directories (instead of all source files). Without a reason to justify this, it seems quite odd to me.

I believe a more intuitive approach would be to enable or disable the feature simultaneously in all the crates involved. Do you have a reason why we shouldn't do that?

Additionally, the `cvm_guest` feature appears to exist in the `aster-virtio` crate as well. If multiple crates share the same feature gate, keeping the enable state in sync is the simplest way to maintain it. Otherwise, the behavior will become quite messy.

tatetian:
> I just cannot figure out a valid reason why we should leave the inconsistency between OSTD and the kernel. For example, why would we want a build that does the following things simultaneously:
> 
> * disables `cvm_guest` in the kernel,
> * enables `cvm_guest` in the OSTD.

Sure. This would be undesirable.

On the other hand, I don't want us to add/support a feature/functionality that does not have a clear user. This would be over-engineering.

In the case of `NO_DEFAULT_FEATURES `, I don't think there is any legit use case to disable the default feature (`NO_DEFAULT_FEATURES`) to enable select features. In other words, `NO_DEFAULT_FEATURES` can be removed from the root `Makefile`.

lrh2000:
> On the other hand, I don't want us to add/support a feature/functionality that does not have a clear user. This would be over-engineering.

Of course. But what I propose can be implemented as simple as:
```patch
diff --git a/Cargo.toml b/Cargo.toml
index 3fcc0779d..e93d2d891 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,7 +80,7 @@ license = "MPL-2.0"
 edition = "2024"
 
 [workspace.dependencies]
-ostd = { version = "0.17.0", path = "ostd" }
+ostd = { version = "0.17.0", path = "ostd", default-features = false }
 
 # Crates under ostd/libs
 align_ext = { path = "ostd/libs/align_ext", version = "0.1.0" }
```

Would you consider this to be "over-engineering"?

> In the case of `NO_DEFAULT_FEATURES `, I don't think there is any legit use case to disable the default feature (`NO_DEFAULT_FEATURES`) to enable select features. In other words, `NO_DEFAULT_FEATURES` can be removed from the root `Makefile`.

Then, there is no way to disable the `cvm_guest` feature. How can we ensure that the kernel (and the OSTD) still compiles with this feature disabled?

Actually, I'm not sure if the OSTD in the main branch can be built with the feature disabled since there is no existing way to disable the `cvm_guest` feature for the OSTD (AFAICK).

lrh2000:
> But what I propose can be implemented as simple as:

As far as I can know, the feature dependency has already been correctly implemented in:
https://github.com/asterinas/asterinas/blob/b0407dd5173552b59cf08f84a5e44914b60e67bd/kernel/Cargo.toml#L79

So the simple approach will work because the default feature set, `cvm_guest`, still depends on the OSTD's `cvm_guest` feature. Therefore, it won't break the default build. Please correct me if I'm missing anything.

tatetian:
> ```diff
> ostd = { version = "0.17.0", path = "ostd", default-features = false }
> ```

Sure. This is a simple change. I thought that `kernel/Cargo.toml` would need to change as well. But since this is a one-line change, I won't be against including this change as part of this PR.

With that said, I won't advertise for setting `default-features = false`, either. Here are the reasons:
1. Whenever people see `default-features = false` in `Cargo.toml`, they would wonder "why not use the default feature" and "what's wrong with default feature"?
2. Future OSTD features which are part of the `default` have to be opted in by the downstream crates _manually_. This could be a surprise to the author of this OSTD feature.
3. Asterinas is the primary user of OSTD. If this primary user prefers `default-features = false`, then I would highly doubt the wisdom of OSTD's `default` feature. I prefer fixing the `default` feature rather than shortcircuit-ing the default feature with `default-features = false`.

> Then, there is no way to disable the `cvm_guest` feature. How can we ensure that the kernel (and the OSTD) still compiles with this feature disabled?
> 
> Actually, I'm not sure if the OSTD in the main branch can be built with the feature disabled since there is no existing way to disable the `cvm_guest` feature for the OSTD (AFAICK).

Thanks for sharing your concern. I got your motive. Fundamentally, you are pursuing a (reasonable) goal but it is beyond the scope of this PR. So you want to ensure that all combinations of features of OSTD and the kernel compiles and run. But I think it should be done in a separate PR that includes new CI tests to ensure this goal is achieved. Setting `default-features = false` is easy but it takes more than this one-line change to ensure the goal is not only achieved but unbroken in the long run.

If our goal is to ensure all combinations of features work, then I think we should remove `cvm_guest` from `default`. It is much easier to add features than remove with Cargo.

In conclusion, I think we can keep this PR as it is now. No one is using `make NO_DEFAULT_FEATURES=1` and even if someone does, having `cvm_guest` enabled for OSTD but disabled for the kernel would not break any functionality. After this PR, we can pursue the higher goal in another PR.

lrh2000:
> With that said, I won't advertise for setting `default-features = false`, either. Here are the reasons:

Agree with this points.

> If our goal is to ensure all combinations of features work, then I think we should remove `cvm_guest` from `default`. It is much easier to add features than remove with Cargo.

This does seem to be a more reasonable approach. We can set the default features at the Makefile level:
```
FEATURES ?= cvm_guest
```

> In conclusion, I think we can keep this PR as it is now. No one is using `make NO_DEFAULT_FEATURES=1` and even if someone does, having `cvm_guest` enabled for OSTD but disabled for the kernel would not break any functionality. After this PR, we can pursue the higher goal in another PR.

Thanks for the detailed explanation! I'm fine with pursuing the goal in the future, taking your points into consideration separately.

tatetian:
> If our goal is to ensure all combinations of features work, then I think we should remove `cvm_guest` from `default`. It is much easier to add features than remove with Cargo.
> 
> In conclusion, I think we can keep this PR as it is now. No one is using `make NO_DEFAULT_FEATURES=1` and even if someone does, having `cvm_guest` enabled for OSTD but disabled for the kernel would not break any functionality. After this PR, we can pursue the higher goal in another PR.

@StevenJiang1110 May I count on you to work on this? I recall that you are working on refactoring the Makefile and its environment variables and have recently merged a PR that refactors all `Cargo.toml` files to be workspace aware.

StevenJiang1110:
> @StevenJiang1110 May I count on you to work on this? I recall that you are working on refactoring the Makefile and its environment variables and have recently merged a PR that refactors all `Cargo.toml` files to be workspace aware.

Surely. I will take it.

PR #2952, opened 2026-02-05
Fix `Credentials` system calls and clean it up

lrh2000:
There are some issues with our `Credentials` implementation.

 1. We need to check the `CAP_SETUID` and `CAP_SETGID` instead of checking whether the effective user is root.
https://github.com/asterinas/asterinas/blob/5d0779bd832cc6ac029e6d4131586635c4cd4779/kernel/src/process/credentials/credentials_.rs#L79-L81
 2. The zero UID is a valid UID, which represents the root user.
https://github.com/asterinas/asterinas/blob/5d0779bd832cc6ac029e6d4131586635c4cd4779/kernel/src/syscall/setresuid.rs#L7
 3. `setfsuid()` and `setfsgid()` never fails. It will always return the old filesystem IDs.
https://github.com/asterinas/asterinas/blob/5d0779bd832cc6ac029e6d4131586635c4cd4779/kernel/src/syscall/setfsuid.rs#L17
 4. `setuid()` and `setgid()` may fail.
https://github.com/asterinas/asterinas/blob/5d0779bd832cc6ac029e6d4131586635c4cd4779/kernel/src/syscall/setuid.rs#L16

Additionally, there are some style issues. For instance, functions that operate on user IDs are written differently than those that operate on group IDs (e.g., there is `set_fsuid_unchecked`, but not `set_fsgid_unchecked`). I don't know why that is, so I unified them.

The doc comments have been rewritten. The previous version contained things like "user ids", which do not make sense. We should write "user IDs" instead.

In the last commit, I stole a small change from https://github.com/asterinas/asterinas/pull/2912 because the `UidGidRootTest.SetgidNotFromThreadGroupLeader` test failed due to some signal checks.

### DIFF
--- a/kernel/src/fs/procfs/pid/task/status.rs
+++ b/kernel/src/fs/procfs/pid/task/status.rs
@@ -109,6 +109,25 @@ impl FileOps for StatusFileOps {
         writeln!(printer, "Pid:\t{}", posix_thread.tid())?;
         writeln!(printer, "PPid:\t{}", process.parent().pid())?;
         writeln!(printer, "TracerPid:\t{}", 0)?;
+
+        let credentials = posix_thread.credentials();
+        writeln!(
+            printer,
+            "Uid:\t{}\t{}\t{}\t{}",
+            u32::from(credentials.ruid()),
+            u32::from(credentials.euid()),
+            u32::from(credentials.suid()),
+            u32::from(credentials.fsuid()),
+        )?;
+        writeln!(
+            printer,
+            "Gid:\t{}\t{}\t{}\t{}",
+            u32::from(credentials.rgid()),
+            u32::from(credentials.egid()),
+            u32::from(credentials.sgid()),
+            u32::from(credentials.fsgid()),
+        )?;
+
         writeln!(
             printer,
             "FDSize:\t{}",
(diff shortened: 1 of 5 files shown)


lrh2000:
I just realized that properly implementing the `Uid:` and `Gid:` fields in `/proc/self/status` should be easy. It's certainly better than leaving several TODOs.
https://github.com/asterinas/asterinas/blob/389020862934420f770fd9390c66aa9ec9d1d915/test/initramfs/src/syscall/ltp/testcases/all.txt#L473
https://github.com/asterinas/asterinas/blob/389020862934420f770fd9390c66aa9ec9d1d915/test/initramfs/src/syscall/ltp/testcases/all.txt#L481

I'll try to add these (trivial) changes to the last commit and see if the tests can pass later.

lrh2000:
> I'll try to add these (trivial) changes to the last commit and see if the tests can pass later.

Done. As shown by [the force-push diff](https://github.com/asterinas/asterinas/compare/389020862934420f770fd9390c66aa9ec9d1d915..deef64f3332ac38f81a4a4fa6b6c685b4439762b), trivial changes can make these tests pass.

PR #2956, opened 2026-02-06
Add documentation for more verified applications

cqs21:


tatetian:
The chosen popular applications in this PR are not justified and the categorization won't scale to hundreds of packages. See #2968 for a more comprehensive plan.

book/src/distro/popular-applications/ai-machine-learning.md
```diff
@@ -0,0 +1,260 @@
...
+```python
+#!/usr/bin/env python3.12
+
+import unittest
+import math
+import tensorflow as tf
+
+
+class TestTFEnv(unittest.TestCase):
+    def test_import_and_version(self):
```
tatetian: The Python code is duplicated here and the CI Nix file. Extract the Python code into a file and include it in here and the Nix file. This comment applies to all testing Python code.

book/src/distro/popular-applications/nix-and-nixos-tools/README.md
```diff
@@ -1,11 +1,17 @@
-# Package Management
+# Package Management & Containerization
+
+This category covers Nix package management tools and container runtimes.
+
+## Package Management
+
+### Nix
```
tatetian, line 7: I think that Nix commands are fundamental to NixOS so it deserves its own category. So you are going to have two categories:
* "Nix and NixOS Tools"
* "Containerization and Virtualization"

For the virtualization category, we can later add QEMU.

book/src/distro/popular-applications/README.md
```diff
@@ -6,3 +6,26 @@ but it is impractical to test every package
 and document its support status.
 Instead, we group popular applications into categories
 and document the ones we have verified to work.
+
+## Categorization Overview
```
tatetian: "Categorization Overview" --> "Categories"

book/src/distro/popular-applications/README.md
```diff
@@ -6,3 +6,26 @@ but it is impractical to test every package
...
+
+The categorization is designed to scale to hundreds of applications while serving both server and desktop use cases.
+It follows a two-level hierarchy (Category > Sub-category) with the following top-level categories:
+
+| Category | Description | Packages |
+|----------|-------------|----------|
+| [System Core](system-core.md) | Shells, init systems, system monitoring, and essential utilities | `Bash`, `Fish`, `Zsh`, `BusyBox`, `Systemd`, `Fastfetch`, `Htop`, `Lsof`, `Ncdu`, `Procps`, `Coreutils`, `Diffutils`, `Findutils`, `Grep`, `Hostname`, `Less`, `Man-pages`, `More`, `Texinfo`, `Util-linux`, `Which` |
```
tatetian: Application names should not be wrapped inside escaping backticks. For example, write "Apache HTTP Server", not `Apache HTTP Server`.

tatetian (commented):
The following comments are manually written by me

book/src/distro/popular-applications/package-management-containerization.md
```diff
@@ -93,3 +99,82 @@ TODO: upgrade mdbook to enable admonition blocks like the one below:
...
+[Podman](https://docs.podman.io/en/stable/Introduction.html) is a modern, daemonless container engine
+that provides a Docker-compatible command-line interface,
+making it easy for users familiar with Docker to transition.
+
+#### Installation
+
+To install Podman, add the following line to `configuration.nix`:
+
+```nix
+virtualization.podman.enable = true;
```
tatetian: **Wrong NixOS option name**: The module is `virtualisation` (British spelling), so this snippet fails at `nixos-rebuild`. Fix to `virtualisation.podman.enable = true;` and audit other `virtualization.*` references.

book/src/distro/popular-applications/system-core.md
```diff
@@ -0,0 +1,593 @@
...
+man -P cat ls
+```
+
+### More
+
+[More](https://man7.org/linux/man-pages/man1/more.1.html) is a pager for viewing text files.
+
+#### Installation
+
+```nix
```
tatetian: **Prefer the canonical attribute**: `pkgs.more` is a thin wrapper that just re-exposes the binary from `pkgs.util-linux` (which is already documented later in the same file). The snippet works but is confusing when paired with the util-linux section; consider citing `pkgs.util-linux` directly.

book/src/distro/popular-applications/development-tools.md
```diff
@@ -0,0 +1,671 @@
...
+## Programming Language Runtimes
+
+### Clang
+
+[Clang](https://clang.llvm.org/) is a C language family frontend for the LLVM compiler.
+
+#### Installation
+
+```nix
+environment.systemPackages = [ pkgs.llvmPackages.clangUseLLVM ];
```
tatetian: **Unusual Clang attribute**: `pkgs.llvmPackages.clangUseLLVM` is a specialized variant that swaps libstdc++ for libc++ — likely overkill for the "Hello, world" examples that follow. The canonical install attribute is `pkgs.clang` (or `pkgs.llvmPackages.clang` to pin an LLVM version); please switch unless the libc++ variant is intentional.

book/src/distro/popular-applications/development-tools.md
```diff
@@ -0,0 +1,671 @@
...
+
+```nix
+environment.systemPackages = [ pkgs.hugo ];
+```
+
+#### Verified Usage
+
+```bash
+# Create a new site
+hugo new project my-site
```
tatetian: **Wrong scaffolding subcommand**: `hugo new site my-site` is the command for scaffolding a new site. As written, `hugo new project my-site` is interpreted as content creation (`hugo new <path>`), which silently creates a `project/my-site` content file rather than a site — not what the "Create a new site" comment implies.

book/src/distro/popular-applications/development-tools/README.md
```diff
@@ -0,0 +1,671 @@
...
+```
+
+## Build Tools
+
+### Cargo
+
+[Cargo](https://doc.rust-lang.org/cargo/) is the Rust package manager and build system.
+
+#### Installation
+
```
tatetian, line 316: **Rust coverage is thin and split**: `pkgs.rustc` ships only the compiler; `cargo` is a separate section further down. Recommend installing both together, add a `cargo run`/`cargo build` example, and cross-link the two sections.

book/src/distro/popular-applications/file-management-terminal.md
```diff
@@ -0,0 +1,637 @@
...
+
+```bash
+# Find files in current directory
+fd
+
+# Find files containing a pattern
+fd pattern
+
+# Find files with specific size
+fd -S 1M
```
tatetian: **Invalid `fd` size filter**: `-S` requires a sign (`-S +1M` or `-S -1M`); a bare `-S 1M` errors out. Fix to `fd -S +1M` and note the sign convention.

book/src/distro/popular-applications/networking.md
```diff
@@ -0,0 +1,234 @@
...
+# Connect to HTTP server
+lftp http://ftp.sjtu.edu.cn/ubuntu-cd/
+
+# Download single file
+lftp -c "open ftp.sjtu.edu.cn; cd /ubuntu-cd; get robots.txt"
+```
+
+### Netcat
+
+[Netcat](https://www.libressl.org/) is a networking utility for reading from and writing to network connections.
```
tatetian: **Link is accurate but unhelpful**: `pkgs.netcat` in nixpkgs is in fact the LibreSSL `nc`, so the link is not technically wrong — but a reader looking for netcat background will be confused by landing on the LibreSSL homepage. Prefer `https://man.openbsd.org/nc.1`, or add a clarifying sentence. Separately, the `Crunch` (line 510) and `Wipe` (line 621) links in `file-management-terminal.md` point to unrelated GitHub repositories (upstream is SourceForge in both cases).

book/src/distro/popular-applications/communication/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Join a channel
+/join #mychannel
+
+# Leave a channel
+/part #mychannel
+
+# Quit irssi
+/quit
+```
```
tatetian, line 28: **`bash` fence mixes shell and TUI commands**: `/connect` and `/join` are irssi-internal, not shell. Split into two blocks (or prefix the TUI lines with `# Inside irssi:`, matching the `# Inside htop:` pattern in `system-core.md`). WeeChat section below has the same issue.

book/src/distro/popular-applications/README.md
```diff
@@ -6,3 +6,26 @@ but it is impractical to test every package
...
+| [CI/CD & DevOps](cicd-devops.md) | CI/CD runners and infrastructure automation | `Just`, `Task`, `GoReleaser` |
+| [Monitoring & Observability](monitoring-observability.md) | Metrics, logging, and tracing tools | `Prometheus` |
+| [Desktop Environments & Display](desktop-environments-display.md) | Desktop environments, window managers, and display servers | `Xfce`, `Xorg`, `galculator`, `mousepad`, `fairymax`, `gnome-mines` |
+| [Web Browsers](web-browsers.md) | Web browsers | `Links2`, `W3m` |
+| [Office & Productivity](office-productivity.md) | Office suites, document viewers, and note-taking apps | `MuPDF`, `Pandoc` |
+| [Multimedia](multimedia.md) | Video/audio players, graphics tools, and streaming software | `SoX`, `ImageMagick`, `FFmpeg` |
+| [Communication](communication.md) | Email clients, instant messaging, and video conferencing | `Irssi`, `WeeChat` |
```
tatetian: **Descriptions overstate coverage**: Each of these advertises sub-categories that have no verified packages on the page (VPN, firewalls, search engines, message queues, note-taking, email, video). Trim descriptions to what is actually verified, or add a disclaimer.

book/src/distro/popular-applications/desktop-environments-and-display/README.md
tatetian, line 1: **Empty `Xorg` section and README/page mismatch**: The `Xorg` heading has no description or install snippet (every other app does) — add content or remove. Also, the README's package list for this category omits `mupdf`, `five-or-more`, `lbreakout2`, `gnome-chess`, `gnome-sudoku`, `tali`, `xboard`, `xgalaga` that the page verifies. Align the two.

.github/workflows/test_popular_apps.yml
```diff
@@ -0,0 +1,71 @@
...
+on:
+  schedule:
+    # Schedule to run on every day at 16:00 UTC (00:00 Beijing Time)
+    - cron: '0 16 * * *'
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - .github/workflows/test_popular_apps.yml
+      - book/src/distro/popular-applications/**
+      - test/nixos/tests/**
```
tatetian: **Coverage regression**: The deleted workflow triggered on `distro/**`, `tools/nixos/**`, and `push: main`. The new one drops all three. PRs touching `distro/**` or `tools/nixos/**` (used by `make nixos` via `tools/nixos/build_nixos.sh`) now skip NixOS validation, and post-merge regressions wait for the daily cron. Restore both paths and add a `push: { branches: [main] }` trigger.

.github/workflows/test_popular_apps.yml
```diff
@@ -0,0 +1,71 @@
...
+          - name: office-productivity
+            disk_size: 8192
+          - name: multimedia
+            disk_size: 8192
+          - name: communication
+            disk_size: 8192
+          - name: file-management-terminal
+            disk_size: 8192
+          - name: ai-machine-learning
+            disk_size: 8192
```
tatetian: **`nix` and `podman` suites silently dropped**: The old workflow ran both as dedicated suites. They appear subsumed by `package-management-containerization`, but the mapping is not documented and the old `main.rs` files (46 and 49 lines) are deleted without an audit. Either document the subsumption in the PR body after verifying every old assertion has an equivalent, or keep them as separate matrix entries.

.github/workflows/test_popular_apps.yml
```diff
@@ -0,0 +1,71 @@
...
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - .github/workflows/test_popular_apps.yml
+      - book/src/distro/popular-applications/**
+      - test/nixos/tests/**
+
+jobs:
+  popular-apps-test:
+    runs-on: ubuntu-latest
```
tatetian: **No `timeout-minutes`, and runner downgraded**: Old workflow had `timeout-minutes: 30` on `ubuntu-4-cores-150GB-ssd` (chosen for "extended storage"). New workflow inherits GitHub's 6-hour default on standard `ubuntu-latest`. A hang (e.g. the `wipe` issue below) burns up to 6×14 job-hours. Add `timeout-minutes: 45` and confirm the standard runner has enough disk/RAM for the heaviest suites.

.github/workflows/test_popular_apps.yml
tatetian: **`disk_size: 8192` likely too small for heavy suites**: Browsers, LibreOffice, multimedia codecs, and PyTorch/TensorFlow/Ollama routinely exceed 8 GiB under `/nix/store`. Lighter suites (`system-core`, `networking`, `development-tools`) got 10240 — the allocation is inverted. Bump the heavy suites to ≥10240 (12288 for AI/ML) or size from measured footprints.

.github/workflows/test_popular_apps.yml
```diff
@@ -0,0 +1,71 @@
...
+      - uses: endersonmenezes/free-disk-space@v3
+        with:
+          rm_cmd: "rmz"
+          remove_android: true
+          remove_dotnet: true
+          remove_haskell: true
+          remove_tool_cache: true
+      - uses: actions/checkout@v4
+      - name: Run the ${{ matrix.test.name }} test on AsterNixOS
+        uses: maus007/docker-run-action-fork@v1
```
tatetian: **Supply-chain risk: third-party fork pinned by floating tag**: A force-push to `v1` in that personal fork would execute immediately with `--privileged -v /dev:/dev`. Pin to an immutable SHA and document why the fork is needed. Same for `endersonmenezes/free-disk-space@v3` (line 56).

test/nixos/tests/databases-middleware/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+        "is healthy",
+    )?;
+
+    nixos_shell.run_cmd("etcdctl --endpoints=localhost:2379 put testkey testvalue")?;
+    nixos_shell.run_cmd_and_expect(
+        "etcdctl --endpoints=localhost:2379 get testkey",
+        "testvalue",
+    )?;
+    nixos_shell.run_cmd_and_expect("etcdctl --endpoints=localhost:2379 del testkey", "1")?;
+    nixos_shell.run_cmd_and_expect("etcdctl --endpoints=localhost:2379 get testkey", "")?;
```
tatetian: **Tautological expectation**: Every string contains `""`, so this assertion always passes regardless of whether the key was deleted. Use a sentinel (e.g. `etcdctl ... --print-value-only; echo __END__` and expect `__END__` without preceding `testvalue`) or an exit-status check.

test/nixos/tests/databases-middleware/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+
+// ============================================================================
+// Time Series Databases - InfluxDB
+// ============================================================================
+
+#[nixos_test]
+fn influxdb_server(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("influxd config > /tmp/influxdb.conf")?;
+    nixos_shell.run_cmd("sed -i '/bind-address/s/:8086/10.0.2.15:8086/' /tmp/influxdb.conf")?;
+    nixos_shell.run_cmd("influxd -config /tmp/influxdb.conf > /tmp/influxd.log 2>&1 &")?;
```
tatetian: [**`test-cleanup`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/testing-guidelines.html#test-cleanup): Redis, Valkey, etcd, and InfluxDB are spawned in background/daemon mode and never stopped. Since suites share a VM, leaks outlive each test. Add shutdowns before returning: `redis-cli -p 6380 shutdown nosave`, `valkey-cli -p 6381 shutdown nosave`, `pkill etcd`, `pkill influxd`.

test/nixos/tests/databases-middleware/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+        "influx -host 10.0.2.15 -port 8086 -execute 'USE testdb'",
+        "Using database testdb",
+    )?;
+    nixos_shell.run_cmd("influx -host 10.0.2.15 -port 8086 -database testdb -execute 'INSERT cpu,host=server1 value=0.64'")?;
+    nixos_shell.run_cmd_and_expect(
+        "influx -host 10.0.2.15 -port 8086 -database testdb -execute 'SELECT * FROM cpu'",
+        "server1",
+    )?;
+    Ok(())
+}
```
tatetian: **InfluxDB version pin is implicit**: `pkgs.influxdb` currently aliases to 1.x (which matches the v1 `influx -execute 'CREATE DATABASE ...'` flow used here), but that alias is not stable — nixpkgs periodically promotes InfluxDB 2.x under the bare name, and the v1 CLI does not work against v2. Consider pinning an explicit `influxdb_1_x` attribute (or the 2.x `influx write` / `influx bucket` flow) so the test does not silently break on a channel bump.

test/nixos/tests/file-management-terminal/src/main.rs
```diff
@@ -0,0 +1,601 @@
...
+}
+
+// ============================================================================
+// Security & Backup - Wipe
+// ============================================================================
+
+#[nixos_test]
+fn wipe_zero_pass(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("echo 'erase me' > /tmp/wipe-zero.txt")?;
+    nixos_shell.run_cmd("wipe -z /tmp/wipe-zero.txt")?;
```
tatetian: **`wipe` blocks on interactive prompt**: Without `-f`, `wipe` prompts `Okay to proceed (yes/no)?`, hanging the pty until timeout. Add `-f`: `wipe -zf ...`, `wipe -pf 8 ...` (applies to line 598 too).

test/nixos/tests/ai-machine-learning/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ============================================================================
+// Deep Learning Frameworks - TensorFlow
+// ============================================================================
+
+#[nixos_test]
+fn run_tensorflow(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd_and_expect("python /tmp/test_tensorflow.py", "OK")?;
+    Ok(())
```
tatetian: **Fragile `"OK"` substring match**: `unittest`'s final uppercase `OK` line is a reasonable success signal today, but any future test name or helper output that happens to contain the substring `OK` will mask a `FAILED` run (since `run_cmd_and_expect` uses substring-contains). A sentinel chained on exit status is more robust: `python /tmp/test_torch.py && echo __PYTORCH_OK__` and expect `__PYTORCH_OK__`. Same for `run_tensorflow` (line 32).

test/nixos/tests/web-servers-proxies/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+        .run_cmd("caddy file-server --listen 10.0.2.15:8001 --browse > /tmp/caddy.log 2>&1 &")?;
+    nixos_shell.run_cmd("sleep 3")?;
+    nixos_shell.run_cmd_and_expect("curl http://10.0.2.15:8001", "Hello from Caddy")?;
+    nixos_shell.run_cmd("pkill caddy")?;
+    Ok(())
+}
+
+#[nixos_test]
+fn nginx_server(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("cp -rL $(nginx -h 2>&1 | grep -o '/nix/store[^)]*') /tmp/nginx")?;
```
tatetian: **Fragile `/nix/store` parsing**: Today `nginx -h` prints a single store path, so the command works, but the help-text format is not a stable contract; if a future nginx prints another store path (for a module or referenced file) `cp -rL $(...)` will treat all but the last as sources. Prefer `nginx -V 2>&1 | grep -oP '\-\-prefix=\K[^ ]+'`, or pass the prefix via an env var set in `extra_config.nix`.

test/nixos/tests/package-management-containerization/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+    nixos_shell.run_cmd_and_expect("./result/bin/hello", "Hello, world!")?;
+    Ok(())
+}
+
+#[nixos_test]
+fn nixos_rebuild(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("echo '{ pkgs, ... }: { environment.systemPackages = [ pkgs.hello ]; }' > /tmp/add-hello.nix")?;
+    nixos_shell.run_cmd("echo '{ imports = [ /etc/nixos/configuration.nix /tmp/add-hello.nix ]; }' > /tmp/test-config.nix")?;
+    nixos_shell.run_cmd("nixos-rebuild test -I nixos-config=/tmp/test-config.nix")?;
+    nixos_shell.run_cmd("rm /tmp/*")?;
```
tatetian: **`rm /tmp/*` is too broad**: It wipes any fixture files from `extra_config.nix` and artifacts from earlier tests in the same binary. If test ordering or shared fixtures change, later tests break silently. Replace with explicit paths: `rm /tmp/add-hello.nix /tmp/test-config.nix`.

test/nixos/tests/communication/extra_config.nix
```diff
@@ -0,0 +1,44 @@
...
+      expect "irssi"
+      send "/connect irc.libera.chat\r"
+      expect {
+          "established" { puts "OK: connected" }
+          timeout       { puts "FAIL: timeout"; exit 1 }
+      }
+      send "/quit\r"
+      expect eof
+      puts "PASS"
+    '';
```
tatetian: **External dependency + missing timeout branch**: Reaching out to `irc.libera.chat` makes CI flaky on Libera outages or network issues — use a local `pkgs.ngircd`/`pkgs.inspircd` on `127.0.0.1` instead. Also, the initial `expect "irssi"` lacks a `timeout` branch, so on failure it silently falls through; add `timeout { puts "FAIL: irssi did not start"; exit 1 }`.

test/nixos/tests/monitoring-observability/extra_config.nix
```diff
@@ -0,0 +1,20 @@
...
+      global:
+        scrape_interval: 15s
+      scrape_configs:
+        - job_name: "prometheus"
+          static_configs:
+            - targets: [ "10.0.2.15:9090" ]
+    '';
+  };
+in {
+  environment.systemPackages = with pkgs; [ prometheus grafana ];
```
tatetian: **Installed but untested**: `main.rs` only tests Prometheus; claiming Grafana as "verified" without a smoke test is misleading. Either add one (start on a non-default port, curl `/api/health`) or drop it. Same concern for `htop` and `texinfoInteractive` in `test/nixos/tests/system-core/extra_config.nix` (lines 8, 20).

test/nixos/tests/databases-middleware/src/main.rs
tatetian, line 1: **`run_cmd` does not check exit status (framework gap)**: Several findings above (empty-string etcd, wipe hang, fragile nginx parsing) pass silently because `run_cmd` only waits for the next prompt. Add `run_cmd_and_check_status` to `test/nixos/common/framework/` — append `; echo __STATUS__=$?` and assert `__STATUS__=0`. Can be a follow-up PR.

test/nixos/tests/system-core/src/main.rs
```diff
@@ -0,0 +1,424 @@
...
+fn man(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("echo 'less' > /tmp/less_test.txt")?;
+    nixos_shell.run_cmd_and_expect("man -P cat ls", "list directory contents")?;
+    Ok(())
+}
+
+// ============================================================================
+// Less
+// ============================================================================
+
```
tatetian: [**`explain-why`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#explain-why): Banner says `// Less` but brackets `fn more` (real `Less` block is at lines 336-344). Change to `// More`.

test/nixos/tests/system-core/src/main.rs
```diff
@@ -0,0 +1,424 @@
...
+}
+
+#[nixos_test]
+fn coreutils_touch_wc(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("touch -t 202601011200 /tmp/touched.txt")?;
+    nixos_shell.run_cmd("echo 'one two three' > /tmp/wc_test.txt")?;
+    nixos_shell.run_cmd_and_expect("wc -l /tmp/wc_test.txt", "1")?;
+    nixos_shell.run_cmd_and_expect("wc -w /tmp/wc_test.txt", "3")?;
+    nixos_shell.run_cmd_and_expect("wc -c /tmp/wc_test.txt", "14")?;
+    Ok(())
```
tatetian: **Dead setup lines**: `coreutils_touch_wc` creates `/tmp/touched.txt` with a timestamp but never checks it — either assert the timestamp or drop the `touch`. `man` creates an unused `/tmp/less_test.txt` (likely copy-paste from the `less` test) — remove.

test/nixos/tests/web-servers-proxies/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    nixos_shell.run_cmd_and_expect("curl http://10.0.2.15:8000", "It works!")?;
+    nixos_shell.run_cmd("httpd -f /tmp/httpd/conf/httpd.conf -k stop")?;
+    Ok(())
+}
+
+#[nixos_test]
+fn caddy_server(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd("mkdir -p /tmp/caddy")?;
+    nixos_shell.run_cmd("echo 'Hello from Caddy' > /tmp/caddy/index.html")?;
+    nixos_shell.run_cmd("cd /tmp/caddy")?;
```
tatetian: **Stray `cd` between independent `run_cmd` calls (systemic)**: A standalone `cd` on its own line, followed by an unrelated command, appears here and in `ci-cd-devops/src/main.rs`. Other tests correctly chain with `&&` (e.g. `"cd /tmp/git_test && git status"`); the current ones happen to work only because later commands use absolute paths. Either chain with `&&` or add a `run_cmd_in(dir, cmd)` helper.

test/nixos/tests/system-core/Cargo.toml
tatetian, line 1: [**`workspace-deps`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#workspace-deps): No new `test/nixos/tests/*/Cargo.toml` includes `[lints] workspace = true`, and `test/nixos/Cargo.toml` lacks `[workspace.lints]`. AGENTS.md requires it. This silently disables Clippy policy across the whole `test/nixos` workspace (also affects pre-existing `hello`). Add both.

test/nixos/tests/system-core/src/main.rs
tatetian, line 1: **Doc boilerplate duplicated across 14 files**: The "Document maintenance" block is identical in every new `main.rs`; any update must touch 15 files in lockstep. Extract into `test/nixos/tests/README.md` (or a `common/framework` module doc) and replace per-file copies with a one-line reference. Not a blocker.

tatetian (commented):
The following comments are generated by my agents. I don't have the time to validate them one by one. Take some with a great of salt.

tatetian (approved):

Issue #1762, opened 2024-12-26
Inconsistent naming convention for test functions


grief8:
#### **Description**
Currently, the naming convention for test functions in the codebase is inconsistent. Some test functions are prefixed with `test_`, while others are not. For example:

```rust
test aster_nix::vm::vmar::interval_set::tests::test_take_one ... ok
test aster_nix::vm::vmar::interval_set::tests::test_take ... ok
test aster_nix::vm::vmar::interval_set::tests::test_clear ... ok
test aster_nix::vm::vmo::options::test::alloc_vmo ... ok
test aster_nix::vm::vmo::options::test::alloc_continuous_vmo ... ok
test aster_nix::vm::vmo::options::test::write_and_read ... ok
test aster_nix::vm::vmo::options::test::resize ... ok
```

This inconsistency can lead to confusion for developers and make it harder to maintain and navigate the test suite.

#### **Proposed Solution**
To improve consistency and readability, I propose adopting the common Rust convention of prefixing all test functions without `test_`. This aligns with the examples in the [Rust Book - Writing Tests](https://doc.rust-lang.org/book/ch11-01-writing-tests.html) and is widely used in the Rust ecosystem. Generally, we define the test in a module, and this module contains the meaning of test, so adding `test_` to the function in the module becomes redundant.

For example:
```rust
#[cfg(ktest)]
mod tests {
    // Before
    [ktest]
    fn test_clear() {
        // Test logic
    }
    
    // After
    #[ktest]
    fn clear() {
        // Test logic
    }
}
```

#### **Benefits**
1. **Consistency**: A uniform naming convention makes the codebase easier to read and maintain.
2. **Clarity**: The `mod *_test` clearly indicates that a function is a test module.
3. **Alignment with Community Standards**: This convention is widely used in the Rust community, as seen in the Rust Book.

PR #2959, opened 2026-02-08
Implement `clippy::redundant_test_prefix` for `#[ktest]`

lrh2000:
This is to enforce https://github.com/asterinas/asterinas/issues/1762. I have mentioned this many times in the review comments. It should be checked by a lint.

There is a lint called `clippy::redundant_test_prefix` that serves the same purpose. However, it does not work for `#[ktest]` items.

In this PR, I enabled the workspace lint `clippy::redundant_test_prefix` so that normal `#[test]` items could be checked. I extended the `#[ktest]` derive macro to implement similar functionality by (ab)using a deprecated item to emit compiler warnings.

The compiler will complain if the `#[ktest]` attribute is applied to a method whose name begins with `test_`:
https://github.com/asterinas/asterinas/blob/3d6d91ea18161646f71000fc0139af463a55dafd/ostd/src/mm/page_table/test.rs#L156-L157
```
warning: use of deprecated constant `mm::page_table::test::test_utils::_::REDUNDANT_TEST_PREFIX`: 
             redundant `test_` prefix in test function name
             help: consider removing the `test_` prefix: `subset_iter`
             help: consider allowing the lint by: `#[ktest(expect_redundant_test_prefix)]`
             help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_test_prefix
   --> ostd/src/mm/page_table/test.rs:156:5
    |
156 |     #[ktest]
    |     ^^^^^^^^
    |
    = note: this warning originates in the attribute macro `ktest` (in Nightly builds, run with -Z macro-backtrace for more info)
```

In this specific case, `test_subset_iter` is fine: It needs the `test_` prefix because it is not located within a `test` module. So it can be fixed by:
```patch
-#[ktest]
+#[ktest(expect_redundant_test_prefix)]
 fn test_subset_iter() {
```

On the other hand, if `expect_redundant_test_prefix` is specified but the method name does not start with `test_`, the compiler will also complain:
```
warning: use of deprecated constant `mm::frame::allocator::_::REDUNDANT_TEST_PREFIX`: 
             no redundant `test_` prefix in test function name
             help: remove the expectation from: `#[ktest(expect_redundant_test_prefix)]`
   --> ostd/src/mm/frame/allocator.rs:109:1
    |
109 | #[ktest(expect_redundant_test_prefix)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
    = note: this warning originates in the attribute macro `ktest` (in Nightly builds, run with -Z macro-backtrace for more info)
```

As a reference, this is the message that will be emitted by the `clippy::redundant_test_prefix` lint:
```
warning: redundant `test_` prefix in test function name
  --> src/lib.rs:27:8
   |
27 |     fn test_it_works() {
   |        ^^^^^^^^^^^^^ help: consider removing the `test_` prefix: `it_works`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_test_prefix
note: the lint level is defined here
  --> src/lib.rs:26:12
   |
26 |     #[warn(clippy::redundant_test_prefix)]
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/kernel/libs/cpio-decoder/src/test.rs
+++ b/kernel/libs/cpio-decoder/src/test.rs
@@ -5,7 +5,7 @@ use lending_iterator::LendingIterator;
 use super::{CpioDecoder, FileType, error::*};
 
 #[test]
-fn test_decoder() {
+fn decoder() {
     use std::process::{Command, Stdio};
 
     let manifest_path = std::env::var("CARGO_MANIFEST_DIR").unwrap();
@@ -63,7 +63,7 @@ fn test_decoder() {
 }
 
 #[test]
-fn test_short_buffer() {
+fn short_buffer() {
     let short_buffer: Vec<u8> = Vec::new();
     let mut decoder = CpioDecoder::new(short_buffer.as_slice());
     let entry_result = decoder.next().unwrap();
@@ -72,7 +72,7 @@ fn test_short_buffer() {
 }
 
 #[test]
-fn test_invalid_buffer() {
+fn invalid_buffer() {
     let buffer: &[u8] = b"invalidmagic.invalidmagic.invalidmagic.invalidmagic.invalidmagic.invalidmagic.invalidmagic.invalidmagic.invalidmagic.invalidmagic";
     let mut decoder = CpioDecoder::new(buffer);
     let entry_result = decoder.next().unwrap();
--- a/kernel/libs/jhash/src/lib.rs
+++ b/kernel/libs/jhash/src/lib.rs
@@ -260,7 +260,7 @@ mod test {
     const JHASH_INITVAL: u32 = 0;
 
     #[test]
-    fn test_jhash_3vals() {
+    fn jhash_3vals_works() {
         assert_eq!(jhash_3vals(1, 2, 3, JHASH_INITVAL), 2757843189);
         assert_eq!(jhash_3vals(4, 5, 6, JHASH_INITVAL), 3701334656);
 
@@ -275,7 +275,7 @@ mod test {
     }
 
     #[test]
-    fn test_jhash_2vals() {
+    fn jhash_2vals_works() {
         assert_eq!(jhash_2vals(1, 2, JHASH_INITVAL), 2337044857);
         assert_eq!(jhash_2vals(3, 4, JHASH_INITVAL), 3842257880);
         assert_eq!(jhash_2vals(0, 0, JHASH_INITVAL), 1489077439);
@@ -285,7 +285,7 @@ mod test {
     }
 
     #[test]
-    fn test_jhash_1vals() {
+    fn jhash_1vals_works() {
         assert_eq!(jhash_1vals(1, JHASH_INITVAL), 1923623579);
         assert_eq!(jhash_1vals(5, JHASH_INITVAL), 4121471414);
         assert_eq!(jhash_1vals(0, JHASH_INITVAL), 76781240);
@@ -295,7 +295,7 @@ mod test {
     }
 
     #[test]
-    fn test_jhash_u32_array() {
+    fn jhash_u32_array_works() {
         assert_eq!(jhash_u32_array(&[1, 2, 3], JHASH_INITVAL), 2757843189);
         assert_eq!(jhash_u32_array(&[4, 5, 6, 7, 8], JHASH_INITVAL), 581654130);
         assert_eq!(jhash_u32_array(&[], JHASH_INITVAL), 3735928559);
@@ -309,7 +309,7 @@ mod test {
     }
 
     #[test]
-    fn test_jhash_slice() {
+    fn jhash_slice_works() {
         assert_eq!(jhash_slice(b"hello world", JHASH_INITVAL), 1252609637);
         assert_eq!(jhash_slice(b"12345", JHASH_INITVAL), 729031446);
         assert_eq!(jhash_slice(b"\n\t\r", JHASH_INITVAL), 483925400);
--- a/kernel/src/fs/configfs/test.rs
+++ b/kernel/src/fs/configfs/test.rs
@@ -158,7 +158,7 @@ fn init_demo_subsystem() {
 }
 
 #[ktest]
-fn test_config_fs() {
+fn config_fs() {
     init_demo_subsystem();
     let config_fs = super::fs::ConfigFs::singleton();
 
--- a/kernel/src/fs/pipe/common.rs
+++ b/kernel/src/fs/pipe/common.rs
@@ -539,7 +539,7 @@ mod test {
     }
 
     #[ktest]
-    fn test_read_empty() {
+    fn read_empty() {
         test_blocking(
             |writer| {
                 assert_eq!(write(writer.as_ref(), &[1]).unwrap(), 1);
@@ -554,7 +554,7 @@ mod test {
     }
 
     #[ktest]
-    fn test_write_full() {
+    fn write_full() {
         test_blocking(
             |writer| {
                 assert_eq!(write(writer.as_ref(), &[1, 2, 3]).unwrap(), 2);
@@ -572,7 +572,7 @@ mod test {
     }
 
     #[ktest]
-    fn test_read_closed() {
+    fn read_closed() {
         test_blocking(
             drop,
             |reader| {
@@ -584,7 +584,7 @@ mod test {
     }
 
     #[ktest]
-    fn test_write_closed() {
+    fn write_closed() {
         test_blocking(
             |writer| {
                 assert_eq!(write(writer.as_ref(), &[1, 2, 3]).unwrap(), 2);
@@ -599,7 +599,7 @@ mod test {
     }
 
     #[ktest]
-    fn test_write_atomicity() {
+    fn write_atomicity() {
         test_blocking(
             |writer| {
                 assert_eq!(write(writer.as_ref(), &[1]).unwrap(), 1);
--- a/kernel/src/fs/sysfs/test.rs
+++ b/kernel/src/fs/sysfs/test.rs
@@ -245,7 +245,7 @@ fn init_sysfs_with_mock_tree() -> Arc<SysFs> {
 }
 
 #[ktest]
-fn test_sysfs_root_lookup() {
+fn root_lookup() {
     // Setup: Create SysFs instance backed by the mock systree
     let sysfs = init_sysfs_with_mock_tree();
     let root_inode = sysfs.root_inode(); // Get the sysfs root inode
@@ -280,7 +280,7 @@ fn test_sysfs_root_lookup() {
 }
 
 #[ktest]
-fn test_sysfs_branch_lookup() {
+fn branch_lookup() {
     let sysfs = init_sysfs_with_mock_tree();
     let root_inode = sysfs.root_inode();
     // Action: Lookup a branch node within sysfs
@@ -313,7 +313,7 @@ fn test_sysfs_branch_lookup() {
 }
 
 #[ktest]
-fn test_sysfs_leaf_lookup() {
+fn leaf_lookup() {
     let sysfs = init_sysfs_with_mock_tree();
     let root_inode = sysfs.root_inode();
     // Action: Lookup a leaf node (represented as a directory in sysfs)
@@ -350,7 +350,7 @@ fn test_sysfs_leaf_lookup() {
 }
 
 #[ktest]
-fn test_sysfs_read_attr() {
+fn read_attr() {
     let sysfs = init_sysfs_with_mock_tree();
     let root_inode = sysfs.root_inode();
     let leaf1_dir_inode = root_inode
@@ -381,7 +381,7 @@ fn test_sysfs_read_attr() {
 }
 
 #[ktest]
-fn test_sysfs_write_attr() {
+fn write_attr() {
     let sysfs = init_sysfs_with_mock_tree();
     let root_inode = sysfs.root_inode();
     let leaf1_dir_inode = root_inode
@@ -426,7 +426,7 @@ fn test_sysfs_write_attr() {
 }
 
 #[ktest]
-fn test_sysfs_read_link() {
+fn read_link() {
     use crate::fs::utils::SymbolicLink;
 
     let sysfs = init_sysfs_with_mock_tree();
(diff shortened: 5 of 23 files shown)


ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -299,18 +299,66 @@ pub fn ktest(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        && let Some(stripped_test_prefix) = stripped_test_prefix
+    {
+        let note = format!(
+            "\n\tredundant `test_` prefix in test function name\
+             \n\thelp: consider removing the `test_` prefix: `{}`\
+             \n\thelp: consider allowing the lint by: `#[ktest(expect_redundant_test_prefix)]`\
+             \n\thelp: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_test_prefix",
+            stripped_test_prefix
+        );
+        Some(note)
```
StevenJiang1110: Have you considered using [Diagnostic](https://doc.rust-lang.org/nightly/proc_macro/struct.Diagnostic.html) to emit compiler warnings? It seems to be the most "official" API for generate warnings in proc macros, especially when you want to attach richer information like `help` or `note`. The output also matches the compiler's native warning formatting—for example, labels like `help` and `note` are rendered in bold.

By contrast, the "deprecated item" trick produces extra generated code and it’s hard to make the output fully match the compiler’s warning format. Though I think the "deprecated item" trick should be acceptable if we do not have better solutions.

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -299,18 +299,66 @@ pub fn ktest(_attr: TokenStream, item: TokenStream) -> TokenStream {
         proc_macro2::Span::call_site(),
     );
 
-    let is_should_panic_attr = |attr: &&syn::Attribute| {
+    // Emit warnings similar to `clippy::redundant_test_prefix`.
+    let should_have_test_prefix = if !attr.is_empty() {
+        assert_eq!(
+            attr.to_string(),
+            "expect_redundant_test_prefix",
```
StevenJiang1110, line 368: I can't see why we should need such an attribute. I think even a ktest is not in a test(or ktest) module, the `test_` prefix is redundant. 

It seems that the clippy's corresponding lint also does not permit `test_` prefix if the test is not defined in a test module. This comment explains why.

https://github.com/rust-lang/rust-clippy/issues/8931#issuecomment-2080327078

StevenJiang1110 (commented):
Thanks for bringing this lint check! I think it's really helpful!

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -171,16 +176,25 @@ pub fn global_heap_allocator(_attr: TokenStream, item: TokenStream) -> TokenStre
...
-    // Reject if the input is not a `const fn`.
+
     let item = parse_macro_input!(item as syn::ItemFn);
+    let fn_name = &item.sig.ident;
+
+    // Reject if the input is not a `const fn`.
     assert!(
         item.sig.constness.is_some(),
         "the annotated function must be `const`"
     );
```
StevenJiang1110, line 188: It looks like the `assert` can be removed now? Because the function will be called in line 195 now, which is inside a const function. If `item` itself is not a const function, line 195 will generate an error?

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -171,16 +176,25 @@ pub fn global_heap_allocator(_attr: TokenStream, item: TokenStream) -> TokenStre
...
 
     quote!(
         /// SAFETY: The name does not collide with other symbols.
-        #[unsafe(export_name = "__GLOBAL_HEAP_SLOT_INFO_FROM_LAYOUT")]
+        #[unsafe(no_mangle)]
+        const extern "Rust" fn __GLOBAL_HEAP_SLOT_INFO_FROM_LAYOUT(
```
StevenJiang1110: Now that it’s a const fn, wouldn’t it be better to use a lower_snake_case name? It also feels a bit odd to me: if it’s uppercase, why don’t we need to add `#[expect(non_snake_case)]`  for it?

ostd/libs/ostd-macros/src/lib.rs
lrh2000: https://github.com/asterinas/asterinas/blob/3d6d91ea18161646f71000fc0139af463a55dafd/ostd/src/mm/page_table/test.rs#L156-L157

https://github.com/asterinas/asterinas/blob/3d6d91ea18161646f71000fc0139af463a55dafd/ostd/src/mm/page_table/test.rs#L15

Could it be confused for a utility method? Yes, there is the `#[test]` attribute, but if someone manages to call `test_util::subset_iter`, that's not obvious.

Using the `test_` prefix can prevent incorrect and non-sense auto-completions, at least.



> It seems that the clippy's corresponding lint also does not permit `test_` prefix if the test is not defined in a test module.

This is equivalent to `#[expect(clippy::redundant_test_prefix)]`. If we use clippy lints, we should be able to use it.

ostd/libs/ostd-macros/src/lib.rs
lrh2000: I'm not aware of it. I'll take a look at it later. Thanks for the information.

ostd/libs/ostd-macros/src/lib.rs
lrh2000: https://github.com/asterinas/asterinas/blob/3d6d91ea18161646f71000fc0139af463a55dafd/ostd/libs/ostd-macros/src/lib.rs#L265-L272

This is also redundant because:
https://github.com/asterinas/asterinas/blob/3d6d91ea18161646f71000fc0139af463a55dafd/ostd/libs/ostd-test/src/lib.rs#L132-L133

I am not sure whether I should remove the assertion. Maybe the assertion message could be clearer than the generated errors. So I just leave the code as it is (since the problem was not introduced in this PR, as we already have redundant assertions in existing code, i.e., the above example).

ostd/libs/ostd-macros/src/lib.rs
StevenJiang1110: Thanks for the explanation—that sounds reasonable. We’ll need something like `#[expect(clippy::redundant_test_prefix)]` to silence the warning, and the current implementation seems like the simplest option.

ostd/libs/ostd-macros/src/lib.rs
lrh2000: > why don’t we need to add `#[expect(non_snake_case)]` for it?

Many lints won't fire for macro-generated code. It depends on the lint.

Example: https://github.com/rust-lang/rust-clippy/blob/15e0ce9d69340dc83fd8a37ba2c1596354bea6bc/clippy_lints/src/redundant_test_prefix.rs#L77-L79



> Now that it’s a const fn, wouldn’t it be better to use a lower_snake_case name? It also feels a bit odd to me: if it’s uppercase, why don’t we need to add `#[expect(non_snake_case)]` for it?

I will fix this so that the function name follows the correct convention, as do the other methods in this file.

ostd/libs/ostd-macros/src/lib.rs
StevenJiang1110: Ok, we can just leave the code as it is since the problem is not introduced in this PR.

ostd/libs/ostd-macros/src/lib.rs
StevenJiang1110: Thanks for your explanation and efforts!

ostd/libs/ostd-macros/src/lib.rs
lrh2000: It does not seem to work with `make check` (or `cargo clippy -- -D warnings`. I haven't found a workaround...

You can try this patch:
```patch
diff --git a/ostd/libs/ostd-macros/src/lib.rs b/ostd/libs/ostd-macros/src/lib.rs
index 0e580da89..c7920c5e3 100644
--- a/ostd/libs/ostd-macros/src/lib.rs
+++ b/ostd/libs/ostd-macros/src/lib.rs
@@ -1,8 +1,9 @@
 
 #![feature(proc_macro_span)]
+#![feature(proc_macro_diagnostic)]
 
-use proc_macro::TokenStream;
+use proc_macro::{Diagnostic, TokenStream};
 use quote::quote;
 use rand::{Rng, distr::Alphanumeric};
 use syn::{Expr, Ident, ItemFn, parse_macro_input};
@@ -260,6 +261,8 @@ pub fn test_panic_handler(_attr: TokenStream, item: TokenStream) -> TokenStream
 /// ```
 #[proc_macro_attribute]
 pub fn ktest(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    Diagnostic::new(proc_macro::Level::Warning, "hello").emit();
+
     // Assuming that the item has type `fn() -> ()`, otherwise panics.
     let input = parse_macro_input!(item as ItemFn);
     assert!(
```

`make check` will print warnings but it does not convert them as errors:
```
    Checking ostd v0.17.0 (/root/asterinas/ostd)
warning: hello

warning: `ostd` (lib) generated 1 warning
    Checking xarray v0.1.0 (/root/asterinas/kernel/libs/xarray)
warning: `xarray` (lib) generated 1 warning (1 duplicate)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.76s
Checking code format for general tests...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
root@localhost:~/asterinas# echo $?
0
```

StevenJiang1110: > It does not seem to work with make check (or cargo clippy -- -D warnings. I haven't found a workaround...

I’ve tried several approaches as well, and it seems we can’t currently control these warnings—they’re unconditional. I also looked into the community discussions, and there’s [ongoing work](https://github.com/rust-lang/rust/pull/135432) to put emitted warnings behind a dedicated lint. So my understanding is that, for now, we can’t use lints to turn this warning into an error or to allow/suppress it.

That said, this PR can’t wait for upstream progress, so we’ll need a workaround anyway:

1. Keep the current approach and emit the warning via `#[deprecated]`.
2. Use `Diagnostic` but emit an error instead of a warning.

Either option looks fine for me—you can choose whichever you prefer.

lrh2000: Thanks for your suggestions!

> 2\. Use `Diagnostic` but emit an error instead of a warning.

If we can do this, maybe we can try to generate the proper `Level` dynamically, according to the environment variables.

It seems that checking the `RUSTFLAGS` environment variable is a possible workaround. However, it is not very robust because it does not work well with `clippy -- -D warnings`. However, it's the best solution I could come up with.

lrh2000:
As suggested by @StevenJiang1110 (https://github.com/asterinas/asterinas/pull/2959#discussion_r2779099031), this PR now uses [`proc_macro::Diagnostics`](https://doc.rust-lang.org/proc_macro/struct.Diagnostic.html) to generate warnings/errors.

Examples:

```
error: redundant `test_` prefix in test function name
   --> ostd/src/sync/mutex.rs:143:8
    |
143 |     fn test_mutex_try_lock_does_not_unlock() {
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: consider removing the `test_` prefix: `mutex_try_lock_does_not_unlock`
   --> ostd/src/sync/mutex.rs:143:8
    |
143 |     fn test_mutex_try_lock_does_not_unlock() {
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider allowing the lint: `#[ktest(expect_redundant_test_prefix)]`
   --> ostd/src/sync/mutex.rs:142:5
    |
142 |     #[ktest]
    |     ^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_test_prefix
    = note: this error originates in the attribute macro `ktest` (in Nightly builds, run with -Z macro-backtrace for more info)
```

```
error: no redundant `test_` prefix in test function name
   --> ostd/src/mm/frame/allocator.rs:110:4
    |
110 | fn alloc_dealloc() {
    |    ^^^^^^^^^^^^^
    |
help: consider removing the expectation: `#[ktest]`
   --> ostd/src/mm/frame/allocator.rs:109:1
    |
109 | #[ktest(expect_redundant_test_prefix)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the attribute macro `ktest` (in Nightly builds, run with -Z macro-backtrace for more info)
```

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -284,32 +300,99 @@ pub fn ktest(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            "unknown arguments"
+        );
+        true
+    } else {
+        false
+    };
+    // FIXME: `-Dwarnings` cannot automatically convert warnings generated from procedural macros
+    // into errors. So we do this manually. Remove this workaround when the upstream changes
+    // resolve the issue (see <https://github.com/rust-lang/rust/pull/135432>).
+    let warning_level = if let Ok(rustflags) = std::env::var("RUSTFLAGS")
```
StevenJiang1110: Ranaming `warning_level` as `lint_level`?

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -284,32 +300,99 @@ pub fn ktest(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        && rustflags
+            .split_whitespace()
+            .any(|flag| flag == "-Dwarnings")
```
StevenJiang1110: RUSTFLAGS can include several forms that are equivalent to `-Dwarnings`. Let’s also handle `-D warnings` (with a space after `-D`) and `--deny warnings`. While there could be multiple spaces after `-D` or `--deny`, covering these patterns should address the most common cases.

lrh2000: We could implement this. However, doing so would make the function too complex.

I extracted two standalone methods: `emit_redundant_test_prefix_warnings` and `generate_panic_expectation_tokens`. Hopefully, this will make the logic easier to follow.

tatetian (approved):

PR #2963, opened 2026-02-10
Fix some ktest CI issues

cchanging:
Currently, our ktest CI has several issues. 

One is that it outputs to `qemu-serial.log` by default, which invalidates the logic of determining test success by checking tail `qemu.log`. 

Another issue is that in the Makefile, when a test fails, the command `(echo "Test failed" && exit 1);` only exits the subshell within the parentheses and does not propagate the failure status outward, causing the CI to report success even when tests have actually failed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -399,6 +399,7 @@ test:
 	done
 
 .PHONY: ktest
+ktest: CONSOLE = ttyS0
 ktest: initramfs $(CARGO_OSDK)
 	@# linux-bzimage-setup is excluded from ktest since it's hard to be unit tested.
 	@CRATES=$$(echo $(filter-out ostd/libs/linux-bzimage/setup,$(OSDK_CRATES)) | tr ' ' ','); \


Makefile
```diff
@@ -402,19 +399,33 @@ test:
 	done
 
 .PHONY: ktest
+ktest: CONSOLE = ttyS0
```
tatetian, line 402: > One is that it outputs to qemu-serial.log by default, which invalidates the logic of determining test success by checking tail qemu.log.

From your problem description, it looks that another solution is to change the logic below to check `qemu-serial.log`. Which is better?

Makefile
```diff
@@ -402,19 +399,33 @@ test:
...
-		echo "[make] Testing $$dir"; \
-		(cd $$dir && cargo osdk test $(CARGO_OSDK_TEST_ARGS)) || exit 1; \
-		tail --lines 10 qemu.log | grep -q "^\\[ktest runner\\] All crates tested." \
-			|| (echo "Test failed" && exit 1); \
-		rm -r target/osdk/*; \
-	done
+	@# 3. Even if a crate test fails, we continue testing other crates but mark
+	@#    the overall result as failed.
+	@failed=0; \
+    for dir in $(OSDK_CRATES); do \
```
tatetian: After this change, the `ktest` target becomes too long to fit into the Makefile. Extract the logic as a script with well-defined CLI under the `tools/` directory.

Makefile
```diff
@@ -402,19 +399,33 @@ test:
...
 
 .PHONY: ktest
+ktest: CONSOLE = ttyS0
 ktest: initramfs $(CARGO_OSDK)
 	@# Notes:
 	@# 1. linux-bzimage-setup is excluded from ktest since it's hard to be unit tested;
 	@# 2. Artifacts are removed after testing each crate to save the limited disk space
 	@#    available to free-tier Github runners.
-	@for dir in $(OSDK_CRATES); do \
-		[ $$dir = "ostd/libs/linux-bzimage/setup" ] && continue; \
```
tatetian, line 407:
```
tail --lines 10 qemu.log
```

This line checks the `qemu.log` file in the project root.

```
(cd $$dir && cargo osdk test $(CARGO_OSDK_TEST_ARGS))
```

But the `cargo osdk test` command runs in `$$dir`. I am wondering if the command generates the `qemu.log` file in the project root or `$$dir`. Could you verify?

Makefile
cchanging: >  change the logic below to check qemu-serial.log

This method can correctly determine whether the test result is successful, but outputting to `qemu-serial.log` means that users cannot see the internal output during the test in the CI, only a final result. If the test fails, it is not easy to locate the specific point of failure. Hence I think outputting to `qemu.log` is better.

Makefile
cchanging: I have verified that the output from all modules during testing is written to `qemu.log` in the root directory.

Makefile
```diff
@@ -402,19 +399,13 @@ test:
 	done
 
 .PHONY: ktest
+ktest: CONSOLE = ttyS0
 ktest: initramfs $(CARGO_OSDK)
-	@# Notes:
-	@# 1. linux-bzimage-setup is excluded from ktest since it's hard to be unit tested;
-	@# 2. Artifacts are removed after testing each crate to save the limited disk space
-	@#    available to free-tier Github runners.
-	@for dir in $(OSDK_CRATES); do \
```
tatetian, line 406: > Another issue is that in the Makefile, when a test fails, the command (echo "Test failed" && exit 1); only exits the subshell within the parentheses and does not propagate the failure status outward,

But the pattern of `cmd || exit 1` is used all over this Makefile. You are saying that they are all wrong?

tools/run_ktest.sh
```diff
@@ -0,0 +1,92 @@
...
+
+        # Remove artifacts to save disk space (useful for CI runners).
+        rm -rf "${PROJECT_ROOT}/target/osdk/"*
+    done
+
+    if [[ $failed -ne 0 ]]; then
+        echo "SUMMARY: Some crate tests failed"
+        exit 1
+    else
+        echo "SUMMARY: All crate tests passed"
```
tatetian: How about using the word "kernel-mode unit tests"? It is more specific

```
    if [[ $failed -ne 0 ]]; then
        echo "SUMMARY: Some kernel-mode unit tests failed"
        exit 1
    else
        echo "SUMMARY: All kernel-mode unit tests passed"
    if
```

Makefile
cchanging: Sorry my explanation was not complete enough. The issue here also arises because `(echo "Test failed" && exit 1);` is executed in a subshell within a `for` block, followed by a semicolon (`;`), which means that regardless of the return value of this statement, the next statement will still be executed sequentially. As a result, the `exit 1` inside the parentheses only causes the return value of this particular statement to be 1, but it does not make the return value of the entire `for` block 1. 

Other `exit 1` statements in our Makefile within the `for` block are not executed in a subshell, so they directly cause the return value of the `for` statement to be 1, which is not an issue.

Issue #2928, opened 2026-01-26
Soundness of Task


hxuhack:
Hi, I’m reviewing the Task module and noticed that the design of `ForceSync` may be unsound.

Relevant code:

https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/ostd/src/task/utils.rs#L5-L6

https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/ostd/src/task/utils.rs#L14-L16

https://github.com/asterinas/asterinas/blob/b7ae7383102800529a079f00f92178511d35b8f3/ostd/src/task/utils.rs#L28-L32

**Main issue**: the method `get` is marked unsafe, but this safety requirement can be bypassed by directly accessing the inner field via `self.0`, which does not require any unsafe context.

As a result, `unsafe impl Sync for ForceSync<T>` effectively allows sharing `&T` across threads even when `T: !Sync`, leading to potential data races.

Here is a minimal proof of concept demonstrating the issue:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=57954e258a692ad2aa1d92964d2273ff.

lrh2000:
> the method `get` is marked unsafe, but this safety requirement can be bypassed by directly accessing the inner field via `self.0`

This is a private field, so you cannot argue as such.

If we have [unsafe fields](https://github.com/rust-lang/rfcs/pull/3458), the field should be marked as `unsafe`, e.g., `ForceSync<T>(unsafe T)`.

Otherwise, you can also argue that there are "soundness problems" in [`Vec::set_len`](https://github.com/rust-lang/rust/blob/873d4682c7d285540b8f28bfe637006cef8918a6/library/alloc/src/vec/mod.rs#L2099-L2107), e.g., `vec.len = usize::MAX;` is clearly unsound if you can access the private fields of `Vec`. But you cannot. This is part of the type's safety invariant.

> Here is a minimal proof of concept demonstrating the issue:

A PoC should demonstrate that UBs can be triggered using the type's public APIs. But:
```rust
        x2.0.set(1); // 💥 data race (UB)
```

I don't think it makes sense... No one can access the private field safely outside the module. This is why we can `unsafe impl Sync`.

hxuhack:
> If we have https://github.com/rust-lang/rfcs/pull/3458, the field should be marked as unsafe, e.g., ForceSync<T>(unsafe T).

[...]

PR #2964, opened 2026-02-10
Mark `call_ostd_main` as unsafe and rename it to `start_kernel`

lrh2000:
`call_ostd_main` needs to be marked as `unsafe` because it requires the caller to uphold some safety preconditions, ensuring that `ostd::init` can be called safely.
https://github.com/asterinas/asterinas/blob/c9032ad97cbc002afcc5eab41c837a850fb70aa0/ostd/src/boot/mod.rs#L129-L130

This change is already discussed in https://github.com/asterinas/asterinas/issues/2928#issuecomment-3872848397.

### DIFF
--- a/ostd/src/boot/mod.rs
+++ b/ostd/src/boot/mod.rs
@@ -113,10 +113,13 @@ pub(crate) fn init_after_heap() {
     });
 }
 
-/// Initializes OSTD and then jumps to the `#[ostd::main]` entry point.
+/// Starts the kernel.
 ///
-/// Any kernel that uses the `ostd` crate should define a function marked with
-/// `#[ostd::main]` as the kernel's entry function.
+/// The job of this function is to continue the early bootstrap (started in [`arch::boot`])
+/// and performs the initialization of OSTD.
+/// Eventually, it transfers control to the entrypoint function
+/// that the user of OSTD defines with `#[ostd::main]`,
+/// which completes the kernel initialization.
 ///
 /// # Safety
 ///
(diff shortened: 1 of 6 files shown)


ostd/src/boot/mod.rs
```diff
@@ -113,13 +113,18 @@ pub(crate) fn init_after_heap() {
...
 ///
-/// This function should be only called from the bootloader-specific module.
-pub(crate) fn call_ostd_main() -> ! {
+/// # Safety
+///
+/// This function must be called only once at a proper timing on the BSP by the
+/// [`arch::boot`] module.
+///
+/// [`arch::boot`]: crate::arch::boot
+pub(crate) unsafe fn call_ostd_main() -> ! {
```
tatetian: Although out of the original scope of this PR, I would like to change the name of this function. The name `call_ostd_main` overly simplifies the job of this function, whose main complexity lies with the entire initialization procedure, before jumping to the `#[ostd_main]` entrypoint. A more proper name would be `start_kernel`, `launch_kernel`, or `boot_into_kernel`.

```rust
/// Starts the kernel.
///
/// The job of this function is to continue the early bootstrap (started in [`arch::boot`])
/// and performs the initialization of OSTD.
/// Eventually, it transfers control to the entrypoint function
/// that the user of OSTD defines with `#[ostd::main]`,
/// which completes the kernel initialization.
///
/// # Safety
///
/// This function must be called only once at a proper timing on the BSP by the
/// [`arch::boot`] module.
///
/// [`arch::boot`]: crate::arch::boot
pub(crate) unsafe fn start_kernel() -> ! {
```

If you agree with this change, do it in a second commit.

lrh2000: I think `call_ostd_main` is inaccurate as well. I just didn't spend time figuring out a better name. You came up with a reasonable name, so I'll follow your suggestion.

PR #2965, opened 2026-02-11
Multi tty

wyt8:
# Multi TTY (Virtual Terminal) Support

This PR implements multi TTY (virtual terminal) support in the Asterinas kernel, extending the existing single-TTY console model to support multiple independent virtual terminals. It brings the TTY/VT behavior of Asterinas closer to Linux-like expectations, enables dynamic VT switching, and lays the foundation for further virtual console improvements.

## Overview

1. Virtual Terminal Manager

  - Introduced a central VT manager responsible for maintaining all created TTY instances.
  
  - Each VT holds its own screen state, cursor position, and console attributes.
  
  - Only the active VT is rendered to the framebuffer at any given time.

2. Runtime VT Switching

  - Supported runtime VT switching between virtual terminals.
  
  - Switching triggers a redraw of the framebuffer using the newly active VT’s stored state.
  
  - Switching behavior follows Linux VT semantics where practical.

3. ioctl & TTY Control Enhancements

  - Enhanced TTY control paths to support VT-related operations.
  
  - Improved handling of terminal control ioctls related to virtual terminal behavior.

## Additional Improvements

1. More complete ANSI escape sequences handling
2. More complete keyboard key handling
3. Remove some hardcode of xorg 

## How to verify it

1. Enable i8042 (this behavior is not caused by me): Add a line in `OSDK.toml`

  ```
  [run.boot]
  kcmd_args = [
      "SHELL=/bin/sh",
      "LOGNAME=root",
      "HOME=/",
      "USER=root",
      "PATH=/bin:/benchmark",
      "init=/usr/bin/busybox",
      "i8042.exist",
  ]
  ```

2. Run without `xorg`

  ```
  make BOOT_PROTOCOL=linux-efi-handover64 run_kernel
  ```

3. Run with `xorg`

  - Enable `xorg`: Modify the `distro/etc_nixos/configuration.nix`
  
  ```
    # Uncomment the two options below to enable the X11 (X.Org) desktop (XFCE).
    services.xserver.enable = true;
    services.xserver.desktopManager.xfce.enable = true;
  ```
  
  - Run nixos
  
  ```
  make run_nixos
  ```

### DIFF
--- a/distro/etc_nixos/modules/systemd.nix
+++ b/distro/etc_nixos/modules/systemd.nix
@@ -22,7 +22,14 @@
     shell = "${pkgs.bash}/bin/bash";
     hashedPassword = null;
   };
-  systemd.targets.getty.wants = [ "autovt@hvc0.service" ];
+  systemd.targets.getty.wants = [
+    "autovt@hvc0.service"
+    "getty@tty2.service"
+    "getty@tty3.service"
+    "getty@tty4.service"
+    "getty@tty5.service"
+    "getty@tty6.service"
+  ];
 
   systemd.extraConfig = ''
     LogLevel=crit      
--- a/test/initramfs/nix/initramfs.nix
+++ b/test/initramfs/nix/initramfs.nix
@@ -37,6 +37,9 @@ in stdenvNoCC.mkDerivation {
 
     cp -r ${etc}/* $out/etc/
 
+    cp ${./../src/init} $out/init
+    chmod +x $out/init
+
     cp ${resolv_conf}/resolv.conf $out/etc/
 
     ${lib.optionalString (apps != null) ''
--- a/test/initramfs/src/init
+++ b/test/initramfs/src/init
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# Start the init process
+exec /bin/init
--- a/OSDK.toml
+++ b/OSDK.toml
@@ -21,7 +21,7 @@ kcmd_args = [
     "PATH=/bin:/benchmark",
     "init=/usr/bin/busybox",
 ]
-init_args = ["sh", "-l"]
+init_args = ["/init"]
 initramfs = "test/initramfs/build/initramfs.cpio.gz"
 
 # Special options for testing
--- a/test/initramfs/src/etc/inittab
+++ b/test/initramfs/src/etc/inittab
@@ -1,0 +1,15 @@
+hvc0::respawn:/bin/sh
+tty1::askfirst:/bin/cttyhack /bin/sh
+tty2::askfirst:/bin/cttyhack /bin/sh
+tty3::askfirst:/bin/cttyhack /bin/sh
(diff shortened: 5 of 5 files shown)


tatetian:
This PR is too large. Break it down in digestible pieces to accommodate code review.

wyt8:
> This PR is too large. Break it down in digestible pieces to accommodate code review.

OK, I will try to break it down.

TinaZhangZW:
Thank you for this contribution.

I was testing the TTY functionality and tried using standard Linux tools like chvt to switch between virtual terminals. It appears to fail with the current changes.

Could you investigate this? Ensuring compatibility with tools like chvt would be a great way to validate the TTY device implementation.

test/initramfs/nix/initramfs.nix
```diff
@@ -37,6 +37,9 @@ in stdenvNoCC.mkDerivation {
 
     cp -r ${etc}/* $out/etc/
 
+    cp ${./../src/init} $out/init
+    chmod +x $out/init
+
```
TinaZhangZW, line 42: This step appears redundant. Whether or not we need TTY switching in the initramfs, init is already installed with BusyBox.

test/initramfs/src/init
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Start the init process
+exec /bin/init
```
TinaZhangZW, line 4: The current init implementation is too simplistic.

In general, full TTY switching functionality (e.g., Ctrl+Alt+F1/F2) cannot be expected to work with just a minimal initramfs and BusyBox—like the current setup in Asterinas. This is a known limitation in Linux. If we want to align with Linux's behavior, we have two options:

1) Enable full TTY switching only for NixOS and defer support for the minimal initramfs + BusyBox until it's actually needed.

2) In addition to adding full TTY switching to NixOS, also implement it in the initramfs by including all necessary configurations in the init script.

I think option 1 sounds more reasonable, at least until someone demonstrates a clear case for pursuing option 2.

distro/etc_nixos/modules/systemd.nix
```diff
@@ -22,7 +22,14 @@
...
   };
-  systemd.targets.getty.wants = [ "autovt@hvc0.service" ];
+  systemd.targets.getty.wants = [
+    "autovt@hvc0.service"
+    "getty@tty2.service"
+    "getty@tty3.service"
+    "getty@tty4.service"
+    "getty@tty5.service"
+    "getty@tty6.service"
+  ];
```
TinaZhangZW, line 32: Are these actually required for NixOS? Could you elaborate on how multi-TTY works with NixOS? Thanks.

lrh2000:
It seems that this PR should be closed because all commits have been submitted in separate PRs (e.g., https://github.com/asterinas/asterinas/pull/3049).

Correct me if I'm wrong.

Issue #2911, opened 2026-01-21
Tracking issue for supporting Asterinas as the guest kernel of kata container


StevenJiang1110:
# Overview

Kata Containers is a widely adopted secure container runtime, commonly used in security-sensitive cloud-native scenarios.

In public cloud environments, containers from different tenants may run on the same host. Traditional runtimes (such as runc) rely on namespaces and cgroups for isolation, meaning different tenants still share the same host kernel and host environment. This model exposes a broad attack surface, including the risk of namespace escape vulnerabilities. The key difference with Kata is that it is VM-based: each pod (often effectively a single container in typical deployments) runs inside its own lightweight virtual machine. Strictly speaking, Kata’s VM maps to the Kubernetes Pod concept—a group of containers that share resources such as volumes and network interfaces. In the rest of this document, unless otherwise specified, we will not distinguish between "pod" and "container".

This issue proposes supporting Asterinas as the guest kernel for Kata Containers, which brings two main benefits:

1. **Production adoption**: Kata is a production-grade runtime that integrates well with mainstream container ecosystems such as containerd and Kubernetes. If Asterinas can run as Kata’s guest kernel, it can reach production users directly.
2. **Reduced isolation requirements in the guest**: Because Kata’s isolation primarily comes from the VM boundary, the guest kernel can omit many container-level isolation features (e.g., user and network namespaces). For Asterinas at its current stage, this keeps the scope manageable and avoids requiring full resource-isolation support upfront.

Asterinas already supports running basic container workloads via Podman + runc, but there is still a significant gap before it can serve as a full guest OS/kernel for Kata Containers. This issue tracks the major gaps.

# The Architecture of Kata Containers

Let’s first review Kata's overall architecture.

### Kata 1.0

The following diagram is from [kata's repo](https://github.com/kata-containers/kata-containers/blob/main/docs/design/arch-images/CNM_detailed_diagram.png).

<img width="960" height="540" alt="Image" src="https://github.com/user-attachments/assets/5b6db514-8cd6-4404-8cf4-413750a952ea" />

To be candid, this diagram is somewhat outdated (it reflects the Kata 1.0 era), but it is still useful at a high level. For our purpose—running Asterinas as the guest kernel—most architectural changes in Kata do not fundamentally alter the big picture.

In the diagram:

- The central “OCI Runtime” corresponds to Kata (the kata-runtime binary). Its role is to launch the VM and communicate with processes inside the VM to create and manage containers.
[...]

PR #2050, opened 2025-04-29
Support PID namespace

StevenJiang1110:

PR #2966, opened 2026-02-11
Support nsfs (/proc/pid/ns)

StevenJiang1110:
This PR adds support for nsfs — the pseudo filesystem behind `/proc/[pid]/ns/` and the entries within it. nsfs exposes the namespaces a process belongs to: each namespace has a corresponding file under `/proc/[pid]/ns/` (e.g., `/proc/[pid]/ns/uts` for the UTS namespace). These files can be opened and then used to query namespace information via ioctl or to switch namespaces via setns.

nsfs is a pseudo filesystem, so the implementation builds on top of our existing `PseudoFs` framework. However, unlike existing `PseudoInode`s, nsfs inodes must also support the open operation so that the resulting file handle can be used for ioctl and setns. To achieve this, the PR introduces a new `NsInode` type that wraps `PseudoInode` with the additional capabilities.

This PR consists of four commits:
1. ~Refactor: split `pseudofs.rs` into per-filesystem modules. The original `pseudofs.rs` housed multiple pseudo filesystems (sockfs, pipefs, etc.) and had already grown quite large. Adding nsfs to the same file would have made it unwieldy, so each filesystem is now moved into its own module.~
2. Add regression tests for nsfs. 
3. Implement nsfs and its common operations. This includes the `/proc/[pid]/ns/` directory entries and file operations such as open and ioctl.
4. Implement setns via nsfs and enable the gVisor setns tests. 

Required by #2911.

### DIFF
--- a/kernel/src/fs/inode_handle.rs
+++ b/kernel/src/fs/inode_handle.rs
@@ -215,6 +215,18 @@ impl InodeHandle {
 
         Ok(())
     }
+
+    pub fn downcast_file_io<T: 'static>(&self) -> Result<Option<&T>> {
+        if self.rights.is_empty() {
+            return_errno_with_message!(Errno::EBADF, "the file is opened as a path");
+        }
+
+        let Some(file_io) = self.file_io.as_ref() else {
+            return Ok(None);
+        };
+
+        Ok((file_io.as_ref() as &dyn Any).downcast_ref::<T>())
+    }
 }
 
 impl Pollable for InodeHandle {
--- a/kernel/src/fs/pseudofs/mod.rs
+++ b/kernel/src/fs/pseudofs/mod.rs
@@ -6,7 +6,7 @@ use core::{
 };
 
 pub use anon_inodefs::AnonInodeFs;
-pub use nsfs::{NsCommonOps, NsType, StashedDentry};
+pub use nsfs::{NsCommonOps, NsFile, NsType, StashedDentry};
 pub use pidfdfs::PidfdFs;
 pub(super) use pipefs::PipeFs;
 use pipefs::PipeFsType;
--- a/kernel/src/syscall/setns.rs
+++ b/kernel/src/syscall/setns.rs
@@ -11,7 +11,13 @@
 //! 2. A `PidFile` opened by `pidfd_open` or by opening `/proc/[pid]` directory.
 
 use crate::{
-    fs::{file_table::FileDesc, path::MountNamespace},
+    fs::{
+        file_handle::FileLike,
+        file_table::FileDesc,
+        inode_handle::InodeHandle,
+        path::MountNamespace,
+        pseudofs::{NsCommonOps, NsFile},
+    },
     net::uts_ns::UtsNamespace,
     prelude::*,
     process::{
@@ -34,13 +40,8 @@ pub fn sys_setns(fd: FileDesc, flags: u32, ctx: &Context) -> Result<SyscallRetur
 
     let new_ns_proxy = if let Some(pid_file) = file.downcast_ref::<PidFile>() {
         build_proxy_from_pid_file(pid_file, ns_type_flags, ctx)?
-    }
-    // TODO: Support setting namespaces from `/proc/[pid]/ns`.
-    else {
-        return_errno_with_message!(
-            Errno::EINVAL,
-            "the FD does not refer to a supported namespace file"
-        );
+    } else {
+        build_proxy_from_ns_file(file.as_ref(), ns_type_flags, ctx)?
     };
 
     // Install the newly created `NsProxy`.
@@ -89,13 +90,6 @@ fn build_proxy_from_pid_file(
     }
 
     if flags.contains(CloneFlags::CLONE_NEWNS) {
-        if ctx.thread_local.is_fs_shared() {
-            return_errno_with_message!(
-                Errno::EINVAL,
-                "setting a mount namespace is not allowed with shared filesystem information"
-            );
-        }
-
         let target_ns = target_proxy.mnt_ns();
         set_mnt_ns(&mut builder, target_ns, ctx)?;
     }
@@ -103,6 +97,63 @@ fn build_proxy_from_pid_file(
     // TODO: Support setting other namespaces from the target process.
 
     Ok(builder.build())
+}
+
+fn build_proxy_from_ns_file(
+    file: &dyn FileLike,
+    flags: CloneFlags,
+    ctx: &Context,
+) -> Result<NsProxy> {
+    if flags.contains(CloneFlags::CLONE_NEWUSER) {
+        return_errno_with_message!(Errno::EINVAL, "setting a user namespace is not supported");
+    }
+
+    check_unsupported_ns_flags(flags)?;
+
+    let inode_handle = file
+        .downcast_ref::<InodeHandle>()
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "the file is not a ns file"))?;
+
+    let current_proxy = ctx.thread_local.borrow_ns_proxy();
+    let current_proxy = current_proxy.unwrap();
+
+    let mut builder = NsProxyBuilder::new(current_proxy);
+
+    #[expect(clippy::nonminimal_bool)]
+    let applied = false
+        || try_apply_ns_from_inode::<UtsNamespace>(inode_handle, flags, |ns| {
+            set_uts_ns(&mut builder, &ns, ctx)
+        })?
+        || try_apply_ns_from_inode::<MountNamespace>(inode_handle, flags, |ns| {
+            set_mnt_ns(&mut builder, &ns, ctx)
+        })?;
+    // TODO: Support setting other namespaces from the ns file.
+
+    if !applied {
+        return_errno_with_message!(Errno::EINVAL, "invalid flags are specified with a ns file");
+    }
+
+    Ok(builder.build())
+}
+
+fn try_apply_ns_from_inode<T: NsCommonOps>(
+    inode_handle: &InodeHandle,
+    flags: CloneFlags,
+    apply: impl FnOnce(Arc<T>) -> Result<()>,
+) -> Result<bool> {
+    let Some(ns_file) = inode_handle.downcast_file_io::<NsFile<T>>()? else {
+        return Ok(false);
+    };
+
+    if !flags.is_empty() && flags != T::TYPE.into() {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "the flags do not match the type of the ns file"
+        );
+    }
+
+    apply(ns_file.ns().clone())?;
+    Ok(true)
 }
 
 fn set_uts_ns(
@@ -141,6 +192,13 @@ fn set_mnt_ns(
     ctx.thread_local
         .borrow_user_ns()
         .check_cap(CapSet::SYS_ADMIN, ctx.posix_thread)?;
+
+    if ctx.thread_local.is_fs_shared() {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "setting a mount namespace is not allowed with shared filesystem information"
+        );
+    }
 
     // TODO: Are the checks above sufficient?
 
--- a/test/initramfs/src/syscall/gvisor/Makefile
+++ b/test/initramfs/src/syscall/gvisor/Makefile
@@ -52,6 +52,7 @@ TESTS ?= \
 	sched_yield_test \
 	semaphore_test \
 	sendfile_test \
+	setns_test \
 	sigaction_test \
 	sigaltstack_test \
 	signalfd_test \
--- a/test/initramfs/src/syscall/gvisor/blocklists/setns_test
+++ b/test/initramfs/src/syscall/gvisor/blocklists/setns_test
@@ -1,0 +1,2 @@
+SetnsTest.ChangeIPCNamespace
+SetnsTest.ChangePIDNamespace


kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,178 @@
...
+    dir: TidDirOps,
+}
+
+impl NsDirOps {
+    /// Creates a new directory inode for the `ns` directory.
+    pub fn new_inode(dir: &TidDirOps, parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        ProcDirBuilder::new(
+            Self { dir: dir.clone() },
+            // Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/proc/base.c#L3321>
+            mkmod!(u + r, a + x),
```
lrh2000: No need to have whitespaces...

Same for other `mkmod!` instances added in this PR.

kernel/src/fs/path/mount_namespace.rs
```diff
@@ -134,3 +139,26 @@ impl Drop for MountNamespace {
...
+            Errno::EINVAL,
+            "a mnt namespace does not have a parent namespace"
+        );
+    }
+
+    fn path(&self) -> &Path {
+        &self.path
+    }
+
+    fn as_any(&self) -> &dyn Any {
```
lrh2000: You should be able to require `Any` as a super trait:
```rust
trait FileIo: Any {}
trait NsCommonOps: Any {}
```

It isn't necessary to add the `as_any` method to `FileIo` and `NsCommonOps`.

kernel/src/fs/path/mount_namespace.rs
```diff
@@ -134,3 +139,26 @@ impl Drop for MountNamespace {
         }
     }
 }
+
+impl NsCommonOps for MountNamespace {
+    const TYPE: NsType = NsType::Mnt;
+
+    fn get_owner_user_ns(&self) -> Result<&Arc<UserNamespace>> {
```
lrh2000: But it seems that there is already a `owner` method?

kernel/src/process/namespace/user_ns.rs
```diff
@@ -37,4 +49,47 @@ impl UserNamespace {
...
+        return_errno_with_message!(
+            Errno::EPERM,
+            "a user namespace does not have an owner user namespace"
+        );
+    }
+
+    fn get_parent(&self) -> Result<Arc<Self>> {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "getting the parent of a user namespace is not supported"
```
lrh2000: So what's the parent namespace of the init user namespace? We do support the init user namespace now.

In addition, I am not sure whether we should return `Option` or `Result` for `owner` and `parent` methods. Do you have some justification?

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// Gets the type of the namespace (e.g., user, pid, mnt, etc.).
+    pub type GetType         = ioc!(NS_GET_NSTYPE, 0xb703, NoData);
+    /// Gets the user ID of the namespace owner.
+    ///
+    /// Only user namespace supports this operation.
+    pub type GetOwnerUid     = ioc!(NS_GET_OWNER_UID, 0xb704, OutData<u32>);
+
+    // Modern encoding ioctl commands
+
+    #[expect(unused)]
```
lrh2000: Move this below the doc comments.

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,178 @@
...
+        Ok(inode)
+    }
+
+    fn populate_children<'a>(
+        &self,
+        dir: &'a ProcDir<Self>,
+    ) -> RwMutexUpgradeableGuard<'a, SlotVec<(String, Arc<dyn Inode>)>> {
+        let mut cached_children = dir.cached_children().write();
+
+        // Refresh NsProxy-backed entries: remove stale ones and re-add
```
lrh2000:
```suggestion
        // Refresh `NsProxy`-backed entries: remove stale ones and re-add
```

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,178 @@
...
+            cached_children.remove_entry_by_name(name);
+            if let Some(ns_proxy) = ns_proxy.as_ref() {
+                let inode = constructor(ns_proxy, dir.this_weak().clone());
+                cached_children.put((name.to_string(), inode));
+            }
+        }
+
+        drop(ns_proxy);
+
+        // Refresh the user namespace entry unconditionally.
```
lrh2000:
```rust
        // Refresh NsProxy-backed entries: remove stale ones and re-add
        // if the proxy is still alive.
```

```rust
        // Refresh the user namespace entry unconditionally.
```

Please choose a consistent comment style and use it for both `NsProxy`-backed entries and the user namespace entry. I cannot see any substantial differences between them.

The current comment style is very confusing because it seems to imply we're dealing with them differently.

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,178 @@
...
+        NsSymOps::new_inode(&*user_ns, parent)
+    }
+}
+
+impl DirOps for NsDirOps {
+    fn lookup_child(&self, dir: &ProcDir<Self>, name: &str) -> Result<Arc<dyn Inode>> {
+        let inode = if name == "user" {
+            self.new_user_ns_inode(dir.this_weak().clone())
+        } else {
+            self.lookup_ns_proxy_child(name, dir.this_weak().clone())?
```
lrh2000: This is racy. If two threads invoke `lookup_child` concurrently, we may end up with an outdated cached entry.

Better to avoid this. See `fd.rs` for an example.

kernel/src/fs/procfs/pid/task/ns.rs
lrh2000: I don't understand why we want to refresh the inodes unconditionally, which means that `ls` will change the inode ID.

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,366 @@
...
+    pub type GetTidFromPidNs = ioc!(NS_GET_PID_FROM_PIDNS, 0xb7, 0x6, InData<i32>);
+    /// Translates process ID from the target PID namespace into the caller's PID namespace.
+    #[expect(unused)]
+    pub type GetPidFromPidNs = ioc!(NS_GET_TGID_FROM_PIDNS, 0xb7, 0x7, InData<i32>);
+    /// Translates thread ID from the caller's PID namespace into the target PID namespace.
+    #[expect(unused)]
+    pub type GetTidInPidNs   = ioc!(NS_GET_PID_IN_PIDNS, 0xb7, 0x8, InData<i32>);
+    /// Translates process ID from the caller's PID namespace into the target PID namespace.
+    #[expect(unused)]
+    pub type GetPidInPidNs   = ioc!(NS_GET_TGID_IN_PIDNS, 0xb7, 0x9, InData<i32>);
```
lrh2000: Try to align the arguments?
```rust
    pub type GetTidInPidNs   = ioc!(NS_GET_PID_IN_PIDNS,  0xb7, 0x8, InData<i32>);
    pub type GetPidInPidNs   = ioc!(NS_GET_TGID_IN_PIDNS, 0xb7, 0x9, InData<i32>);
```

Same for the legacy ioctl commands.

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,366 @@
...
+        }
+    }
+}
+
+// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/include/uapi/linux/magic.h#L95>
+const NSFS_MAGIC: u64 = 0x6e736673;
+
+mod ioctl_defs {
+    use crate::util::ioctl::{InData, NoData, OutData, ioc};
+
```
lrh2000, line 392: Is there a reference URL?

kernel/src/net/uts_ns.rs
```diff
@@ -195,3 +204,26 @@ impl UtsName {
...
+    const TYPE: NsType = NsType::Uts;
+
+    fn get_owner_user_ns(&self) -> Result<&Arc<UserNamespace>> {
+        Ok(&self.owner)
+    }
+
+    fn get_parent(&self) -> Result<Arc<Self>> {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "UTS namespace does not have parent namespace"
```
lrh2000:
```suggestion
            "a UTS namespace does not have a parent namespace"
```

kernel/src/fs/inode_handle.rs
```diff
@@ -215,6 +216,11 @@ impl InodeHandle {
 
         Ok(())
     }
+
+    pub fn as_ns_file<T: NsCommonOps>(&self) -> Option<&NsFile<T>> {
```
lrh2000: Should this succeed for the `O_PATH` case?
```rust
    pub fn unlock_flock(&self) -> Result<()> {
        if self.rights.is_empty() {
            return_errno_with_message!(Errno::EBADF, "the file is opened as a path");
        }
```

kernel/src/fs/inode_handle.rs
```diff
@@ -215,6 +216,11 @@ impl InodeHandle {
...
+        let file_io = self.file_io.as_ref()?;
+        file_io.as_any().downcast_ref::<NsFile<T>>()
```
lrh2000: How do you feel about expanding on this method a bit by:
```rust
pub fn downcast_file_io<T>(&self)
```

Then `T` be used for `NsFile` or other cases, such as PID files?

kernel/src/fs/path/mount_namespace.rs
```diff
@@ -134,3 +139,26 @@ impl Drop for MountNamespace {
...
+        Ok(&self.owner)
+    }
+
+    fn get_parent(&self) -> Result<Arc<Self>> {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "a mnt namespace does not have a parent namespace"
```
lrh2000:
```suggestion
            "a mount namespace does not have a parent namespace"
```

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,366 @@
...
+    pub fn mount_node() -> &'static Arc<Mount> {
+        static NSFS_MOUNT: Once<Arc<Mount>> = Once::new();
+        NSFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()))
+    }
+}
+
+/// An inode representing a namespace entry in [`NsFs`].
+struct NsInode<T: NsCommonOps> {
+    common: PseudoInode,
+    ns: Weak<T>,
```
lrh2000: Please double-check this behavior.

If I remember correctly, for at least some namespace types, we can bind mount a namespace to keep it alive, even if all processes exit. Then, we can re-enter the namespace through the bind-mounted file.

lrh2000: > If I remember correctly, for at least some namespace types, we can bind mount a namespace to keep it alive,

It is not allowed to bind a mount namespace because it will create loops (it fails with `ELOOP` in the following test, but I don't know if it fails unconditionally or only when the loop exists.). 
```
> sudo mount -o bind /proc/self/ns/mnt /tmp/x
mount: /tmp/x: move_mount() failed: Too many levels of symbolic links.
       dmesg(1) may have more information after failed mount system call.
> sudo mount -o bind /proc/self/ns/net /tmp/x
> ls /tmp/x -al
-r--r--r-- 1 root root 0 Feb 10 16:42 /tmp/x
> sudo umount /tmp/x
```

I think you should investigate other types as well.

kernel/src/process/namespace/user_ns.rs
StevenJiang1110: > So what's the parent namespace of the init user namespace? We do support the init user namespace now.

`get_parent` and `get_owner_user_ns` have special semantics for user namespaces. 

Specifically, `get_parent` is not permitted on user namespaces; and `get_owner_user_ns`, when called on a user namespace, returns its parent user namespace instead.

Since the init user namespace has no parent, this operation will return EPERM.

https://elixir.bootlin.com/linux/v6.19/source/kernel/user_namespace.c#L140
https://elixir.bootlin.com/linux/v6.19/source/kernel/user.c#L37

> In addition, I am not sure whether we should return Option or Result for owner and parent methods. Do you have some justification?

I don't have a strong preference either way. My general thinking is: if all failure cases share the same error code, `Option` is probably the cleaner choice; otherwise, `Result` is more appropriate. Based on the Linux implementation, I have updated `get_parent` to return `Result` (as it can return either `EINVAL` or `EPERM`), and `get_owner_user_ns` to return `Option` (as it always returns `EPERM` on failure).

https://elixir.bootlin.com/linux/v6.19/source/fs/nsfs.c#L207

kernel/src/fs/inode_handle.rs
StevenJiang1110: It should fail when O_PATH is specified. The behavior is correctet, and a new regression test is added.

kernel/src/fs/pseudofs/nsfs.rs
StevenJiang1110: Thanks for pointing out this! 

> If I remember correctly, for at least some namespace types, we can bind mount a namespace to keep it alive, even if all processes exit. Then, we can re-enter the namespace through the bind-mounted file.

Yes, that is correct. I have refactored the current implementation so that a namespace caches a `Weak<Dentry>`, while `NsInode` holds an `Arc<Namespace>`. This aligns more closely with the Linux implementation.

There was also an issue with bind mounting a namespace file. Since an NsFile (which has a pseudo mount) does not belong to any mount namespace, it was incompatible with our current implementation, which requires the source file to belong to the current mount namespace. The latest commit addresses this problem.

kernel/src/fs/pseudofs/anon_inode_fs.rs
```diff
@@ -0,0 +1,74 @@
...
+    // Some members of anon_inodefs (such as epollfd, eventfd, timerfd, etc.) share
+    // the same inode. The sharing is not only within the same category (e.g., two
+    // epollfds share the same inode) but also across different categories (e.g.,
+    // an epollfd and a timerfd share the same inode). Even across namespaces, this
+    // inode is still shared. Although this Linux behavior is a bit odd, we keep it
+    // for compatibility.
+    //
+    // A small subset of members in anon_inodefs (i.e., userfaultfd, io_uring, and
+    // kvm_guest_memfd) have their own dedicated inodes. We need to support creating
+    // independent inodes within anon_inodefs for them in the future.
```
lrh2000, line 54: There are several `anon_inodefs` in these paragraphs. Rename the file or edit the comments to ensure a consistent use of `anon_inodefs`/`anon_inode_fs`.

StevenJiang1110: I open #2985 for it. The module has been renamed as `anon_inodefs` to conform to the Linux convention.

test/initramfs/src/apps/security/namespace/proc_nsfs.c
```diff
@@ -0,0 +1,441 @@
...
+	 *    The bind mount keeps the namespace alive. */
+	nsfd = TEST_SUCC(open(BIND_MOUNT_PATH, O_RDONLY));
+	TEST_RES(ioctl(nsfd, NS_GET_NSTYPE), _ret == CLONE_NEWUTS);
+	userns_fd = TEST_SUCC(ioctl(nsfd, NS_GET_USERNS));
+	TEST_SUCC(close(userns_fd));
+	TEST_SUCC(close(nsfd));
+
+	/* 6. Clean up: unmount and remove the mount point. */
+	TEST_SUCC(umount(BIND_MOUNT_PATH));
+	unlink(BIND_MOUNT_PATH);
```
lrh2000:
```suggestion
	TEST_SUCC(unlink(BIND_MOUNT_PATH));
```

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,245 @@
...
+            mkmod!(u+r, a+x),
+        )
+        .parent(parent)
+        .build()
+        .unwrap()
+    }
+}
+
+/// Namespace entries backed by the thread's [`NsProxy`].
+const NS_PROXY_ENTRIES: &[&str] = &["uts", "mnt"];
```
lrh2000: This should be an enum that is parsed from raw strings? So we don't need to have `unreachable!`...

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+
+            let inode = NsSymOps::<UserNamespace>::new_inode(current_path, dir.this_weak().clone());
+            cached_children.remove_entry_by_name(name);
+            cached_children.put((name.to_string(), inode.clone()));
+            Ok(inode)
+        } else {
+            // Validate the name and get the current namespace path.
+            if !NS_PROXY_ENTRIES.contains(&name) {
+                return Err(Error::with_message(
```
lrh2000: `return_errno_with_message`

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,245 @@
...
+                    Errno::ENOENT,
+                    "the file does not exist",
+                ));
+            }
+            let thread = self.dir.thread();
+            let ns_proxy_guard = thread.as_posix_thread().unwrap().ns_proxy().lock();
+            let ns_proxy = ns_proxy_guard.as_ref().ok_or_else(|| {
+                Error::with_message(
+                    Errno::ENOENT,
+                    "the thread's namespace proxy no longer exists",
```
lrh2000:
```suggestion
                    "the thread has exited",
```

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+        return_errno_with_message!(Errno::ESPIPE, "ns files are not seekable");
+    }
+
+    fn is_offset_aware(&self) -> bool {
+        false
+    }
+
+    fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
+        use ioctl_defs::*;
+        dispatch_ioctl!(match raw_ioctl {
```
lrh2000, line 171: Insert a blank line after `use` lines.

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+impl<T: NsCommonOps> InodeIo for NsFile<T> {
+    fn read_at(
+        &self,
+        _offset: usize,
+        _writer: &mut VmWriter,
+        _status_flags: StatusFlags,
+    ) -> Result<usize> {
+        return_errno_with_message!(Errno::EINVAL, "ns files do not support read_at");
```
lrh2000:
```suggestion
        return_errno_with_message!(Errno::EINVAL, "ns files cannot be read at a specific offset");
```

Same for `write_at`.

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+    util::ioctl::{RawIoctl, dispatch_ioctl},
+};
+
+/// A pseudo filesystem for namespace files.
+struct NsFs {
+    _private: (),
+}
+
+impl NsFs {
+    /// Returns the singleton instance of the ns filesystem.
```
lrh2000:
```suggestion
    /// Returns the singleton instance of the ns file system.
```
Be consistent with other files under `kernel/src/fs/pseudofs/*.rs`.

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+    pub(self) fn singleton() -> &'static Arc<PseudoFs> {
+        static NSFS: Once<Arc<PseudoFs>> = Once::new();
+        PseudoFs::singleton(&NSFS, "nsfs", NSFS_MAGIC)
+    }
+
+    /// Creates a pseudo [`Path`] for a namespace file.
```
lrh2000:
```suggestion
    /// Creates a pseudo `Path` for a namespace file.
```
Be consistent with other files under `kernel/src/fs/pseudofs/*.rs`.

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+            Arc::new(NsInode::new(ino, Uid::new_root(), Gid::new_root(), ns, fs))
+        };
+
+        Path::new_pseudo(Self::mount_node().clone(), ns_inode, |inode| {
+            let inode = inode.downcast_ref::<NsInode<T>>().unwrap();
+            inode.name().to_string()
+        })
+    }
+
+    /// Returns the pseudo mount node of the ns filesystem.
```
lrh2000:
```suggestion
    /// Returns the pseudo mount node of the ns file system.
```

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
+        use ioctl_defs::*;
+        dispatch_ioctl!(match raw_ioctl {
+            _cmd @ GetUserNs => {
+                let user_ns = self.ns.get_owner_user_ns().ok_or_else(|| {
+                    Error::with_message(
+                        Errno::EPERM,
+                        "the namespace does not have owner user namespace",
```
lrh2000:
```suggestion
                        "the namespace does not have an owner user namespace",
```

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,408 @@
...
+                        Errno::EINVAL,
+                        "the ns file does not correspond to a user namespace",
+                    )
+                })?;
+                let uid = user_ns.get_owner_uid()?;
+                cmd.write(&uid.into())?;
+                Ok(0)
+            }
+            // TODO: Support additional ioctl commands
+            _ => return_errno_with_message!(Errno::ENOTTY, "unsupported ioctl command"),
```
lrh2000:
```suggestion
            _ => return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown"),
```

Because:
```
> grep -R "ENOTTY, .*command" kernel
kernel/src/device/misc/tdxguest.rs:            _ => return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown"),
kernel/src/device/fb.rs:                return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown");
kernel/src/fs/notify/inotify.rs:            _ => return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown"),
kernel/src/process/process/terminal.rs:                return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown")
kernel/src/util/ioctl/mod.rs://!             _ => return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown"),
```

kernel/src/process/namespace/user_ns.rs
```diff
@@ -37,4 +44,42 @@ impl UserNamespace {
             "the thread does not have the required capability"
         )
     }
+
+    /// Returns the owner UID of the user namespace.
+    pub fn get_owner_uid(&self) -> Result<Uid> {
```
lrh2000: Why do you want the `get_` prefix here? Do you anticipate that this will be an expensive operation?

kernel/src/process/namespace/user_ns.rs
```diff
@@ -37,4 +44,42 @@ impl UserNamespace {
...
+        // whether `self` is an ancestor of `other`.
+        Arc::ptr_eq(self, other)
+    }
+}
+
+impl NsCommonOps for UserNamespace {
+    const TYPE: NsType = NsType::User;
+
+    fn get_owner_user_ns(&self) -> Option<&Arc<UserNamespace>> {
+        // For user namespaces, `NS_GET_USERNS` returns the parent user namespace
```
lrh2000, line 68: Will user namespaces hold strong references to its parent? Should this method return `Option<Arc<..>>` or `Option<&Arc<..>>`?

Though I questioned the duplication between `get_owner_user_ns` and `owner_ns` in the previous version, I now think that maybe we want a cheaper version for non-user namespaces because it's infallible and easy to implement without `Arc::clone`?

kernel/src/fs/pseudofs/nsfs.rs
lrh2000: > The latest commit addresses this problem.

It allows us to bind-mount all pseudo-inodes, but I don't think that's the case. Most pseudo inodes cannot be bind-mounted, and I mentioned a special case regarding the mount ns inode (https://github.com/asterinas/asterinas/pull/2966#discussion_r2802711892). If we can bind-mount it arbitrarily, it will create cyclic `Arc<>`s in the FS tree and then we'll leak resources.

See also https://github.com/asterinas/asterinas/pull/2521#issuecomment-3461603205.

In fact, I think we can treat bind mounts as outside the scope of this PR. If it fails, that's fine; we can maintain the current failure behavior and add support in a future PR when necessary. Otherwise, further investigation is needed.

kernel/src/process/namespace/user_ns.rs
lrh2000: Thanks for the revision! The newly added comments are clear.

kernel/src/fs/pseudofs/nsfs.rs
StevenJiang1110: Thanks for your suggestion! I have removed the last commit from this PR and disabled the bind mount test. I still cannot figure out why attempting to bind mount `/proc/<pid>/ns/mnt` results in `ELOOP`. We can investigate this issue further in the future.

kernel/src/process/namespace/user_ns.rs
StevenJiang1110: > Will user namespaces hold strong references to its parent?

Yes, I plan to do so. According to [the namespaces manual](https://man7.org/linux/man-pages/man7/namespaces.7.html):

```plain
Namespace lifetime
       Absent any other factors, a namespace is automatically torn down
       when the last process in the namespace terminates or leaves the
       namespace.  However, there are a number of other factors that may
       pin a namespace into existence even though it has no member
       processes.  These factors include the following:

       •  ......

       •  The namespace is hierarchical (i.e., a PID or user namespace),
          and has a child namespace.
```

A PID/user namespace remains alive as long as any of its child namespaces are alive. I think the most straightforward way to implement this behavior is to have each child hold a strong reference to its parent, while the parent holds only a weak reference to its children(maybe the weak reference is redundant). (Although I implemented PID namespaces in the opposite manner in #2050, I think that should be corrected.)

I have also renamed `get_parent` to `parent` and `get_owner_user_ns` to `owner_user_ns`, since both operations are inexpensive.

lrh2000: Thanks for your explanation!

kernel/src/fs/pseudofs/mod.rs
```diff
@@ -132,6 +135,7 @@ impl From<PseudoInodeType> for InodeType {
             PseudoInodeType::Socket => InodeType::Socket,
             PseudoInodeType::AnonInode => InodeType::Unknown,
             PseudoInodeType::Pidfd => InodeType::Unknown,
+            PseudoInodeType::Ns => InodeType::File,
```
lrh2000, line 138: https://github.com/asterinas/asterinas/blob/2f549c521282a460a371e1d691ddf46ca021d1ea/kernel/src/fs/pseudofs/mod.rs#L245-L250
https://github.com/asterinas/asterinas/blob/2f549c521282a460a371e1d691ddf46ca021d1ea/kernel/src/fs/pseudofs/mod.rs#L184

Do namespace files support `chmod`?

Is it possible to extend `test/initramfs/src/apps/fs/pseudofs/pseudo_file_create.h` as well as the corresponding `pseudofs` tests to include namespace files?

test/initramfs/src/apps/security/namespace/proc_nsfs.c
```diff
@@ -0,0 +1,447 @@
...
+ * bind-mounts /proc/<child>/ns/uts to a path under /tmp, verifies that
+ * ioctl(2) works on the mounted path, kills the child, waits for it to
+ * exit, then verifies that ioctl(2) still works (the bind mount keeps
+ * the namespace alive). Finally the mount is cleaned up.
+ *
+ * TODO: Asterinas does not yet support bind-mounting namespace files,
+ * so this test is disabled for now. See the discussion at
+ * <https://github.com/asterinas/asterinas/pull/2966#discussion_r2870191714>.
+ */
+#ifndef __asterinas__
```
lrh2000: Can we have something like:
```c
#ifdef __asterinas__
	TEST_ERRNO(mount(proc_ns_path, BIND_MOUNT_PATH, NULL, MS_BIND, NULL), EINVAL);
#else
	TEST_SUCC(mount(proc_ns_path, BIND_MOUNT_PATH, NULL, MS_BIND, NULL));
#endif
```
to ensure that the operation fails?

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,415 @@
...
+    ///
+    /// Only user namespace supports this operation.
+    pub type GetOwnerUid     = ioc!(NS_GET_OWNER_UID,  0xb704, OutData<u32>);
+
+    // Modern encoding ioctl commands
+
+    /// Gets the ID of the mount namespace.
+    #[expect(unused)]
+    pub type GetMntNsId      = ioc!(NS_GET_MNTNS_ID,        0xb7, 0x5, OutData<u64>);
+    /// Translates thread ID from the target PID namespace into the caller's PID namespace.
```
lrh2000:
```suggestion
    /// Translates a thread ID from the target PID namespace into the caller's PID namespace.
```

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Returns a file descriptor of the owner user namespace.
+    pub type GetUserNs       = ioc!(NS_GET_USERNS,     0xb701, NoData);
+    /// Returns a file descriptor of the parent namespace.
+    pub type GetParent       = ioc!(NS_GET_PARENT,     0xb702, NoData);
+    /// Gets the type of the namespace (e.g., user, pid, mnt, etc.).
+    pub type GetType         = ioc!(NS_GET_NSTYPE,     0xb703, NoData);
+    /// Gets the user ID of the namespace owner.
+    ///
+    /// Only user namespace supports this operation.
```
lrh2000:
```suggestion
    /// Only user namespaces support this operation.
```

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,415 @@
...
+    const NAME: &str = Self::TYPE.as_str();
+
+    /// The [`NsType`] discriminant for this namespace kind.
+    const TYPE: NsType;
+
+    /// Returns the owner user namespace.
+    fn owner_user_ns(&self) -> Option<&Arc<UserNamespace>>;
+
+    /// Returns the parent namespace, if one exists.
+    fn parent(&self) -> Result<&Arc<Self>>;
```
lrh2000, line 286: Can we get more docs for `fn parent` and `owner_user_ns`? For example,
 - `fn parent` should say it only works for hierarchical namespaces (user namespaces and PID namespaces);
 - `fn owner_user_ns` should say it returns `None` only if the namespace is the initial user namespace.

kernel/src/fs/pseudofs/mod.rs
StevenJiang1110: > Do namespace files support chmod?

No, `chmod` should return `EPERM`. The behavior has been corrected and test has been added.

test/initramfs/src/apps/security/namespace/proc_nsfs.c
```diff
@@ -0,0 +1,455 @@
...
+
+		TEST_SUCC(close(nsfd));
+	}
+
+	/* An invalid fd must fail with EBADF. */
+	TEST_ERRNO(setns(-1, 0), EBADF);
+
+	/* A child process should be able to join its parent's UTS namespace. */
+	pid_t pid = fork();
+	TEST_RES(pid >= 0, 1);
```
lrh2000:
```suggestion
	pid_t pid = TEST_SUCC(fork());
```

test/initramfs/src/apps/security/namespace/proc_nsfs.c
```diff
@@ -0,0 +1,455 @@
...
+
+	if (pid == 0) {
+		snprintf(path, sizeof(path), "/proc/%d/ns/uts", getppid());
+		int parent_ns = CHECK(open(path, O_RDONLY));
+		CHECK(setns(parent_ns, 0));
+		close(parent_ns);
```
lrh2000:
```suggestion
		CHECK(close(parent_ns));
```

kernel/src/fs/procfs/pid/task/ns.rs
```diff
@@ -0,0 +1,264 @@
...
+
+        // TODO: Support additional namespace types.
+        false
+    }
+}
+
+type NsSymlink<T> = ProcSym<NsSymOps<T>>;
+
+/// Represents the inode at `/proc/[pid]/task/[tid]/ns/<type>` (and also `/proc/[pid]/ns/<type>`).
+pub struct NsSymOps<T: NsCommonOps> {
```
lrh2000:
```suggestion
struct NsSymOps<T: NsCommonOps> {
```
This does not need to be `pub`?

kernel/src/process/namespace/user_ns.rs
```diff
@@ -3,21 +3,28 @@
...
 use crate::{
+    fs::pseudofs::{NsCommonOps, NsType, StashedDentry},
     prelude::*,
-    process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
+    process::{Uid, credentials::capabilities::CapSet, posix_thread::PosixThread},
 };
 
 /// The user namespace.
 pub struct UserNamespace {
     _private: (),
```
lrh2000: `_private` should be removed when we add the first field?

test/initramfs/src/apps/security/namespace/proc_nsfs.c
lrh2000, line 1: Adjusted some minor points in the comment:
```patch
diff --git a/test/initramfs/src/apps/security/namespace/proc_nsfs.c b/test/initramfs/src/apps/security/namespace/proc_nsfs.c
index 74d1ed395..a26be7d35 100644
--- a/test/initramfs/src/apps/security/namespace/proc_nsfs.c
+++ b/test/initramfs/src/apps/security/namespace/proc_nsfs.c
@@ -134,7 +134,7 @@ FN_TEST(zombie_process)
        TEST_RES(pid >= 0, 1);
 
        if (pid == 0) {
-               /* Child exits immediately to become a zombie. */
+               /* Child: exit immediately to become a zombie. */
                exit(0);
        }
 
@@ -194,7 +194,7 @@ FN_TEST(ioctl)
                /*
                 * NS_GET_PARENT: returns the parent namespace fd.
                 * Non-hierarchical namespaces return EINVAL;
-                * the user namespace returns EPERM.
+                * the initial user namespace returns EPERM.
                 */
                if (!is_user_ns) {
                        TEST_ERRNO(ioctl(nsfd, NS_GET_PARENT), EINVAL);
@@ -233,7 +233,7 @@ FN_TEST(setns)
 {
        char path[PATH_MAX];
 
-       /* Joining our own namespace should succeed (except for user ns). */
+       /* Joining our own namespace should succeed (except for the user namespace). */
        for (size_t i = 0; i < ns_count; i++) {
                snprintf(path, sizeof(path), "%s/%s", NS_DIR, ns_files[i]);
                int nsfd = TEST_SUCC(open(path, O_RDONLY));
@@ -404,7 +404,7 @@ FN_TEST(bind_mount_ns_lifetime)
 #else
        TEST_SUCC(mount(proc_ns_path, BIND_MOUNT_PATH, NULL, MS_BIND, NULL));
 
-       /* 3. Verify ioctl works on the bind-mounted path while child is alive. */
+       /* 3. Verify ioctl works on the bind-mounted path while the child is alive. */
        int nsfd = TEST_SUCC(open(BIND_MOUNT_PATH, O_RDONLY));
        TEST_RES(ioctl(nsfd, NS_GET_NSTYPE), _ret == CLONE_NEWUTS);
        int userns_fd = TEST_SUCC(ioctl(nsfd, NS_GET_USERNS));
```

kernel/src/syscall/setns.rs
```diff
@@ -105,6 +99,60 @@ fn build_proxy_from_pid_file(
...
+    let current_proxy = ctx.thread_local.borrow_ns_proxy();
+    let current_proxy = current_proxy.unwrap();
+
+    let mut builder = NsProxyBuilder::new(current_proxy);
+
+    let applied =
+        try_apply_ns::<UtsNamespace>(inode_handle, flags, |ns| set_uts_ns(&mut builder, &ns, ctx))?
+            || try_apply_ns::<MountNamespace>(inode_handle, flags, |ns| {
+                set_mnt_ns(&mut builder, &ns, ctx)
+            })?;
```
lrh2000: If feasible (allowed by rustfmt or `#[rustfmt::skip]` if it worths), can we make this more symmetric:
```patch
diff --git a/kernel/src/syscall/setns.rs b/kernel/src/syscall/setns.rs
index cc977b04a..909cb5a53 100644
--- a/kernel/src/syscall/setns.rs
+++ b/kernel/src/syscall/setns.rs
@@ -119,9 +119,11 @@ fn build_proxy_from_ns_file(
 
     let mut builder = NsProxyBuilder::new(current_proxy);
 
-    let applied =
-        try_apply_ns::<UtsNamespace>(inode_handle, flags, |ns| set_uts_ns(&mut builder, &ns, ctx))?
-            || try_apply_ns::<MountNamespace>(inode_handle, flags, |ns| {
+    let applied = false
+        || try_apply_ns_from_inode::<UtsNamespace>(inode_handle, flags, |ns| {
+                set_uts_ns(&mut builder, &ns, ctx)
+            })?
+        || try_apply_ns_from_inode::<MountNamespace>(inode_handle, flags, |ns| {
                 set_mnt_ns(&mut builder, &ns, ctx)
             })?;
     // TODO: Support setting other namespaces from the ns file.
@@ -133,7 +135,7 @@ fn build_proxy_from_ns_file(
     Ok(builder.build())
 }
 
-fn try_apply_ns<T: NsCommonOps>(
+fn try_apply_ns_from_inode<T: NsCommonOps>(
     inode_handle: &InodeHandle,
     flags: CloneFlags,
     apply: impl FnOnce(Arc<T>) -> Result<()>,
```

We could also mention `inode` in the `try_apply_ns` method name to improve clarity.

tatetian:
@boterinas codex

kernel/src/fs/pseudofs/nsfs.rs
```diff
@@ -0,0 +1,425 @@
...
+                let current = current!();
+                let current_user_ns = current.user_ns().lock();
+                if !current_user_ns.is_same_or_ancestor_of(user_ns) {
+                    return_errno_with_message!(
+                        Errno::EPERM,
+                        "the owner user namespace is not an ancestor of the current namespace"
+                    );
+                }
+
+                open_ns_as_file(user_ns)
```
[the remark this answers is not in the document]
StevenJiang1110: This is incorrect. You can check the logic at https://elixir.bootlin.com/linux/v6.0.9/source/kernel/user_namespace.c#L1365, which does not require the caller has CAP_SYS_ADMIN in the relevant user namespace. If the caller will use the nsfd to setns, then the CAP_SYS_ADMIN will be checked in setns syscall.

test/initramfs/src/apps/security/namespace/proc_nsfs.c
```diff
@@ -0,0 +1,455 @@
...
+ * `ns_names` lists the names as they appear in readlink(2) output.
+ *   - For most namespaces these are identical to `ns_files`.
+ *   - For "pid_for_children" and "time_for_children", readlink(2) shows
+ *     "pid" and "time" respectively (not applicable to the entries below,
+ *     but noted here for future reference).
+ * `clone_flags` lists the corresponding CLONE_NEW* flag for each entry.
+ */
+static const char *ns_files[] = { "uts", "mnt", "user" };
+static const char *ns_names[] = { "uts", "mnt", "user" };
+static const int clone_flags[] = { CLONE_NEWUTS, CLONE_NEWNS, CLONE_NEWUSER };
```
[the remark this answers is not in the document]
StevenJiang1110: This is intented. Just as the document said, only pid and user namespaces are visible for zombie process. The code reflects the behavior.

kernel/src/fs/pseudofs/mod.rs
```diff
@@ -28,6 +29,7 @@ use crate::{
 };
 
 mod anon_inodefs;
+mod nsfs;
```
tatetian: In another concurrent PR, I clarified the position of `pseudofs` in a [comment](https://github.com/asterinas/asterinas/issues/2887#issuecomment-3881555824):

```rust
//! The infrastructure of pseudo FSes along with some naive implementations of pseudo FSes.
//!
//! Pseudo FSes are those that have no physical storage.
//! Some pseudo FSes are trivial,
//! e.g., [`PipeFs`], [`SockFs`], [`PidfdFs`], and [`AnonInodeFs`].
//! Such trivial pseudo FSes are directly implemented in this module
//! using a generic pseudo FS implementation called [`NaivePseudoFs`].
//!
//! On the other hand, some pseudo FSes are quite complex,
//! including [`RamFs`], [`ProcFs`], and [`SysFs`]. Just to name a few.
//! They are implemented in their own modules.
```

In other words, non-trivial implementations of pseudo FSes should not reside in the `pseudofs` module itself. I think you should either put it under the `fs` module or the `namespace` module.

PR #2969, opened 2026-02-13
Fix OSDK and ostd_macro panic problems

StevenJiang1110:
Fixes #2957.

# Root Cause

The panic reported in #2957 occurs because the `ostd::main` macro calls `yield_now` after invoking the user-provided main function:

https://github.com/asterinas/asterinas/blob/0f57097ae8251fe7d66c28d3b82e97d2501b7a2f/ostd/libs/ostd-macros/src/lib.rs#L33-L36

Inside `yield_now`, the global scheduler is accessed. However, if no scheduler has been initialized, the unwrap panics:

https://github.com/asterinas/asterinas/blob/0f57097ae8251fe7d66c28d3b82e97d2501b7a2f/ostd/src/task/scheduler/mod.rs#L557-L559

# Changes

This PR contains three commits that address the scheduler panic along with two related issues:

1. Fix `cargo osdk new` when `OSDK_LOCAL_DEV` is set. When `OSDK_LOCAL_DEV` is set, the newly created kernel crate should depend on OSTD from the local path instead of from crates.io.

2. Fix the scheduler panic by introducing a lazy singleton accessor. Instead of accessing and unwrapping the SCHEDULER static directly, all call sites now go through a `scheduler_singleton` method that lazily initializes the scheduler with a default FIFO implementation if no custom scheduler has been injected. This prevents similar panics in the future.

3. Replace` unreachable!` with `poweroff` in `ostd::main`. The code after `yield_now` in `ostd::main` was marked as unreachable, but it is reachable when the user-provided main function does not spawn any tasks. This commit replaces the `unreachable!` call with `poweroff`, which also fixes a failure in the OSDK tests.

PR #2971, opened 2026-02-14
Extend the doc of the `[ostd::main]` attribute macro

tatetian:
After reviewing #2969, I found that the purpose and constraints of the kernel entrypoint marked with `#[ostd::main]` are not properly documented. So I extend its doc extensively, providing a comprehensive account of the concept of the **bootstrap context**.

### DIFF
--- a/ostd/libs/ostd-macros/src/lib.rs
+++ b/ostd/libs/ostd-macros/src/lib.rs
@@ -10,14 +10,68 @@ use syn::{Expr, Ident, ItemFn, parse_macro_input};
 
 /// A macro attribute to mark the kernel entry point.
 ///
+/// # Bootstrap Context
+///
+/// After initializing itself on the BSP,
+/// OSTD transfers control to this kernel entrypoint function marked with `#[ostd::main]`.
+/// Symmetrically, each AP also has a kernel entrypoint function,
+/// which is specified with `ostd::boot::smp::register_ap_entry`.
+///
+/// Both BSP and AP kernel entrypoints execute code in the **bootstrap context**,
+/// which is intended to provide an opportunity for an OSTD-based kernel to
+/// perform all sorts of necessary initialization steps
+/// before spawning kernel/user tasks to execute code in the task context.
+///
+/// The bootstrap context has the following limitations:
+/// 1. No current task: Unlike the task context, the bootstrap context is not
+///    associated with a task. So calling `Task::current` in the bootstrap
+///    context always gets a `None`.
+/// 2. No turning back: The kernel code execution eventually transitions
+///    from the bootstrap context to the task context. This transition is
+///    triggered by doing `return` or by calling `Task::yield_now`
+///    in a kernel entrypoint function.
+///    Either way, there is no turning back once leaving the bootstrap context.
+///
+/// # Custom Task Scheduler
+///
+/// One key initialization step is to set up a custom task scheduler.
+/// To do so, the bootstrap code should initialize the task scheduler and then
+/// enable it by injecting it into OSTD (see `ostd::task::scheduler::inject_scheduler`).
+/// If no scheduler is injected until a scheduler-related operation (e.g., `Task::yield_now`) is triggered,
+/// then the default FIFO task scheduler will be used by OSTD.
+///
 /// # Examples
 ///
+/// The high-level workflow of a typical entrypoint function is demonstrated below:
+///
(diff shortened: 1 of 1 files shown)


ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -10,14 +10,74 @@ use syn::{Expr, Ident, ItemFn, parse_macro_input};
...
+/// fn idle_task_fn() {
+///     loop {
+///         Task::yield_now();
+///     }
+/// }
+///
+/// struct MyScheduler { /* fields */ };
+///
+/// impl Scheduler for MyScheduler {
+///     // Implement the methods such as `enqueue` and `local_rq_with`.
```
lrh2000: I hope we can at least compile the code. Even better would be if it could run. So:
 1. I don't think it's necessary to include `impl Scheduler for` in this code snippet. We can leave `inject_scheduler` in a comment. The reader should refer to `inject_scheduler` and `Scheduler`'s documentation for more information about how to write a scheduler.
    - If we rename the `Scheduler`'s method names, I don't think we should update the comments here. Let's avoid mentioning the scheduler's interfaces, as I don't believe they provide additional benefits.
 2. `todo!()` can be replaced with a comment.

I think this version is better:
```rust
/// use ostd::boot::smp;
/// use ostd::task::{Task, TaskOptions};
///
/// #[ostd::main]
/// fn bsp_entrypoint() {
///     // Call `ostd::task::scheduler::inject_scheduler` here
///     // to specify a custom scheduler.
/// 
///     // Perform initialization steps on BSP here.
/// 
///     smp::register_ap_entry(ap_entrypoint);
/// 
///     TaskOptions::new(idle_task_fn).spawn();
///     // Returning from here transitions to the task context.
/// }
///
/// fn ap_entrypoint() {
///     // Perform initialization steps on AP here.
/// 
///     TaskOptions::new(idle_task_fn).spawn();
///     // Returning from here transitions to the task context.
/// }
///
/// fn idle_task_fn() {
///     loop {
///         Task::yield_now();
///     }
/// }
```

I formatted the code and addressed the following warnings:
```
warning: unused import: `ostd::prelude::*`
 --> /tmp/tst/src/lib.rs:2:5
  |
2 | use ostd::prelude::*;
  |     ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default
```
```
warning: unused `Result` that must be used
  --> /tmp/tst/src/lib.rs:22:5
   |
22 |     TaskOptions::new(idle_task_fn).spawn();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this `Result` may be an `Err` variant, which should be handled
   = note: `#[warn(unused_must_use)]` (part of `#[warn(unused)]`) on by default
```

tatetian: I like the new simplified version of the example.

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -10,14 +10,68 @@ use syn::{Expr, Ident, ItemFn, parse_macro_input};
...
-///     println!("hello world");
+/// fn bsp_entrypoint() {
+///     // Call `ostd::task::scheduler::inject_scheduler` here
+///     // to specify a custom scheduler.
+///
+///     // Perform initialization steps on BSP here.
+///
+///     smp::register_ap_entry(ap_entrypoint);
+///
+///     TaskOptions::new(idle_task_fn).spawn();
```
lrh2000, line 60: Ohh... line 60 and line 67:
```suggestion
///     TaskOptions::new(idle_task_fn).spawn().unwrap();
```

Sorry, I missed something when I synchronized the code between the GitHub comment and the one in my editor. It's not really a big deal though. But, given that I have noticed this again...

```
warning: unused `Result` that must be used
  --> /tmp/tst/src/lib.rs:22:5
   |
22 |     TaskOptions::new(idle_task_fn).spawn();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this `Result` may be an `Err` variant, which should be handled
   = note: `#[warn(unused_must_use)]` (part of `#[warn(unused)]`) on by default
```

tatetian:
All issues are addressed. I am merging it.

PR #2975, opened 2026-02-23
tools: add openat syscall compatibility test generator

zzj-5341:
Add a C-based test generator that covers 17 O_* flags with 24
constraint rules, generating ~900k valid test cases for open/openat
syscall compatibility testing against Linux behavior.

### DIFF
--- a/kernel/src/fs/fs_resolver.rs
+++ b/kernel/src/fs/fs_resolver.rs
@@ -2,6 +2,8 @@
 
 use alloc::str;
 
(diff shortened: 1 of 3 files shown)


tatetian:
Thanks for the contribution!

The test generator tool itself is not ready for review and merge because it lacks of high-level documentation that describes the design of the tool. I will mark this PR as draft.

The first commit of this PR seems to fix a bug. You can submit it as a separate PR.

PR #2978, opened 2026-02-25
Implement "Capabilities and execution of programs by root"

lrh2000:
This PR implements the following Linux behavior:
```
~ # cat /proc/self/status | grep 'Cap*'
CapInh:	0000000000000000
CapPrm:	000001ffffffffff
CapEff:	000001ffffffffff
CapBnd:	000001ffffffffff
CapAmb:	0000000000000000
~ # su guest
/ $ cat /proc/self/status | grep 'Cap*'
CapInh:	0000000000000000
CapPrm:	0000000000000000
CapEff:	0000000000000000
CapBnd:	000001ffffffffff
CapAmb:	0000000000000000
/ $ su -c 'cat /proc/self/status | grep "Cap*"'
Password: 
CapInh:	0000000000000000
CapPrm:	000001ffffffffff
CapEff:	000001ffffffffff
CapBnd:	000001ffffffffff
CapAmb:	0000000000000000
```

Before this PR, Asterinas's `su guest` lost capabilities, and there was no way to regain root capabilities. This is because we currently do not support file capabilities, nor do we support capability updates when executing a new program with `execve()`.

This PR implements the behavior documented in the "Capabilities and execution of programs by root" section of [the man pages](https://man7.org/linux/man-pages/man7/capabilities.7.html). Note that file capabilities are still not implemented.

In addition, this PR corrects some behavior of the `capset` and `capget` system calls.

### DIFF
--- a/test/initramfs/src/apps/security/capability/capset.c
+++ b/test/initramfs/src/apps/security/capability/capset.c
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include <stdint.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#include <linux/capability.h>
+
+#include "../../common/test.h"
+
+static struct __user_cap_header_struct caphdr = {
+	.version = _LINUX_CAPABILITY_VERSION_3
+};
+static struct __user_cap_data_struct capdat[2];
+
+#define CAPS_ALL 0x000001ffffffffff
+#define CAPS_NONE 0x0000000000000000
+
+FN_TEST(capget)
+{
+	TEST_SUCC(syscall(SYS_capget, &caphdr, capdat));
+
+	TEST_RES(0, (capdat[0].effective |
+		     (((uint64_t)capdat[1].effective) << 32)) == CAPS_ALL);
+	TEST_RES(0, (capdat[0].permitted |
+		     (((uint64_t)capdat[1].permitted) << 32)) == CAPS_ALL);
+	TEST_RES(0, (capdat[0].inheritable |
+		     (((uint64_t)capdat[1].inheritable) << 32)) == CAPS_NONE);
+
+	TEST_SUCC(syscall(SYS_capset, &caphdr, &capdat));
+}
+END_TEST()
+
+FN_TEST(capset_permitted)
+{
+	// Effective capabilities must be permitted.
+	capdat[0].permitted -= 1 << CAP_SYS_ADMIN;
+	TEST_ERRNO(syscall(SYS_capset, &caphdr, &capdat), EPERM);
+
+	capdat[0].effective -= 1 << CAP_SYS_ADMIN;
+	TEST_SUCC(syscall(SYS_capset, &caphdr, &capdat));
+
+	// Adding permitted capabilities is not allowed.
+	capdat[0].permitted += 1 << CAP_SYS_ADMIN;
+	TEST_ERRNO(syscall(SYS_capset, &caphdr, &capdat), EPERM);
+	capdat[0].permitted -= 1 << CAP_SYS_ADMIN;
+}
+END_TEST()
+
+FN_TEST(capset_inheritable)
+{
+	// With CAP_SETPCAP, new inheritable capabilities may not be permitted.
+	capdat[0].inheritable += 1 << CAP_SYS_ADMIN;
+	TEST_SUCC(syscall(SYS_capset, &caphdr, &capdat));
+
+	capdat[0].effective -= 1 << CAP_SETPCAP;
+	TEST_SUCC(syscall(SYS_capset, &caphdr, &capdat));
+
+	// Without CAP_SETPCAP, old inheritable capabilities may not be permitted.
+	TEST_SUCC(syscall(SYS_capset, &caphdr, &capdat));
+
+	capdat[0].inheritable -= 1 << CAP_SYS_ADMIN;
+	TEST_SUCC(syscall(SYS_capset, &caphdr, &capdat));
+
+	// Without CAP_SETPCAP, new inheritable capabilities must be permitted.
+	capdat[0].inheritable += 1 << CAP_SYS_ADMIN;
+	TEST_ERRNO(syscall(SYS_capset, &caphdr, &capdat), EPERM);
+	capdat[0].inheritable -= 1 << CAP_SYS_ADMIN;
+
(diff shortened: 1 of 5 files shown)


kernel/src/process/credentials/credentials_.rs
```diff
@@ -178,6 +178,34 @@ impl Credentials_ {
...
+                // TODO: Get the file capabilities from the file system.
+                CapSet::empty()
+            };
+
+        let file_effective = if self.euid().is_root() && !self.securebits().no_root() {
+            CapSet::all()
+        } else {
+            // TODO: Get the file capabilities from the file system.
+            CapSet::empty()
+        };
```
StevenJiang1110, line 203: I don't really understand the `file_effective` part, from [the manual](https://man7.org/linux/man-pages/man7/capabilities.7.html),

```plain
Effective:
              This is not a set, but rather just a single bit.
```

So `file_effective` should be a boolean value. Am I missing something?

kernel/src/process/credentials/credentials_.rs
```diff
@@ -178,6 +178,34 @@ impl Credentials_ {
...
+
+        // TODO: Add support for the file's inheritable capability set.
+        // TODO: Add support for the bounding capability set and the ambient capability set.
+        let new_permitted = file_permitted;
+        let new_effective = new_permitted & file_effective;
```
StevenJiang1110: It's a bit hard to check the logic here against the manual calculation logic without ambient capabilities. Can we have ambient capabilities in this PR or if you plan to add ambient capabilities in a future PR?

kernel/src/process/credentials/credentials_.rs
```diff
@@ -178,6 +178,34 @@ impl Credentials_ {
 
     pub(super) fn set_suid(&self, suid: Uid) {
         self.set_resuid_unchecked(None, None, Some(suid));
+
+        // Begin to adjust capabilities.
```
StevenJiang1110, line 185: Since we now handle capability calculation during execve, the following comments can be removed.

https://github.com/asterinas/asterinas/blob/97bc17fa58ff2d3be555c48be8973e4d1c924b94/kernel/src/process/execve.rs#L255-L256

kernel/src/process/credentials/capabilities.rs
```diff
@@ -56,9 +56,15 @@ bitflags! {
...
-    pub fn as_u32(&self) -> u32 {
-        self.bits() as u32
+    /// Converts two `u32`s to the capability set. The high bits are truncated.
+    pub fn from_u32_pair(lo: u32, hi: u32) -> Self {
+        let bits = lo as u64 | ((hi as u64) << 32);
+        Self::from_bits_truncate(bits)
+    }
+
+    /// Converts the capability set to two `u32`s.
+    pub fn as_u32_pair(&self) -> (u32, u32) {
```
StevenJiang1110: Would `from_lo_hi/to_lo_hi` be clearer names here? The word pair doesn't tell the caller which u32 is the low half and which is the high half, but lo/hi makes the ordering self-evident at the call site.

kernel/src/syscall/capset.rs
```diff
@@ -6,57 +6,71 @@ use super::SyscallReturn;
...
+    if !credentials.permitted_capset().contains(permitted_capset) {
+        return_errno_with_message!(Errno::EPERM, "adding a permitted capability is not allowed");
+    }
+    if !permitted_capset.contains(effective_capset) {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "adding an effective capability that is not permitted is not allowed"
+        );
+    }
+    if !permitted_capset.contains(inheritable_capset)
```
StevenJiang1110: Based on [the Linux implementation](https://elixir.bootlin.com/linux/v6.19/source/security/commoncap.c#L278), the new inheritable capset should be a subset of `old_inheritable | old_permitted`, but I don't see that check here. Am I missing something?

kernel/src/process/credentials/credentials_.rs
lrh2000: I think it means that `file_effective` can be either `CapSet::all()` or `CapSet::empty()`. In other words, it cannot contain some capabilities instead of all or none.

kernel/src/process/credentials/credentials_.rs
lrh2000: > It's a bit hard to check the logic here against the manual calculation logic without ambient capabilities.

To resolve this problem, I introduced two global constants, BOUNDING_CAPSET and AMBIENT_CAPSET. I hope [the new version](https://github.com/asterinas/asterinas/compare/4f4c5c777e3bf6d89cd3c0dd3460b8b41b0643a3..a3c33ae4ced219513bc929cb0f4d8853f63eabaa) can look better.

> Can we have ambient capabilities in this PR or if you plan to add ambient capabilities in a future PR?

I don't plan on adding support because I'm not aware of any real applications that use it. As far as I'm concerned, it's not a high-priority task.

kernel/src/syscall/capset.rs
lrh2000: Thanks for pointing out the problem. The logic has been corrected.

I've added [some regression tests](https://github.com/asterinas/asterinas/compare/4f4c5c777e3bf6d89cd3c0dd3460b8b41b0643a3..a3c33ae4ced219513bc929cb0f4d8853f63eabaa) related to `capset` to ensure that their behavior conforms to that of the Linux version.

kernel/src/process/credentials/credentials_.rs
StevenJiang1110: Thanks for your explanation!

kernel/src/process/credentials/credentials_.rs
StevenJiang1110: > To resolve this problem, I introduced two global constants, BOUNDING_CAPSET and AMBIENT_CAPSET. I hope [the new version](https://github.com/asterinas/asterinas/compare/4f4c5c777e3bf6d89cd3c0dd3460b8b41b0643a3..a3c33ae4ced219513bc929cb0f4d8853f63eabaa) can look better.

It looks much better to me! Thanks for your efforts.

> I don't plan on adding support because I'm not aware of any real applications that use it. As far as I'm concerned, it's not a high-priority task.

Ok, we can leave it as future work.

StevenJiang1110 (approved):

PR #2979, opened 2026-02-25
Improve `aster_network::TxBuffer` and `aster_network::RxBuffer`

lrh2000:
This PR addresses the following issues relating to `aster_network::TxBuffer` and `aster_network::RxBuffer`:
 1. The `TxBuffer::writer` function cannot be used correctly as it does not call the `sync_to_device` function. This PR removes the unused API.
 2. The `TxBuffer::new` function accepts a packet as a `u8` slice, so the caller must provide the payload in this format. For future vsock refactorings, the intention is to construct the `TxBuffer` directly from the user-space buffer (i.e. the `VmReader`). Therefore, the API has been refactored to accept a `VmReader`.
 3. I do not understand why we want to use `&'static SpinLock<LinkedList<Arc<DmaStream<ToDevice>>, BottomHalfDisabled>>` as the TX pool rather than `DmaPool<ToDevice>`. Is there any valid explanation?

Additionally, this PR adjusts the visibility of comments and items under kernel/comps/virtio/src/device. Hopefully, the resulting code will be improved.

### DIFF
--- a/kernel/comps/network/src/dma_pool.rs
+++ b/kernel/comps/network/src/dma_pool.rs
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
 
-#![expect(unused)]
-
 use alloc::{
     collections::VecDeque,
     sync::{Arc, Weak},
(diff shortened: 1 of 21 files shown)


kernel/comps/network/src/buffer.rs
```diff
@@ -25,36 +27,40 @@ pub struct TxBuffer {
...
-        let nbytes = header.len() + packet.len();
+    ) -> Result<Self> {
+        Self::new_read(header.as_bytes(), packet, pool)
+    }
 
+    fn new_read(
+        header: &[u8],
+        packet: &mut VmReader,
+        pool: &'static SpinLock<LinkedList<Arc<DmaStream<ToDevice>>>, BottomHalfDisabled>,
+    ) -> Result<Self> {
```
StevenJiang1110, line 35: Merge the contents of `new_read` into `new`? The `new` method is just a simple wrapper of `new_read`.

lrh2000: > I checked the Git history, and this does appear to be a legacy issue.

Thanks for your investigation and explanation!

> Merge the contents of `new_read` into `new`? The `new` method is just a simple wrapper of `new_read`.

This is intentional to prevent the Rust compiler from generating multiple function implementations ([monomorphization](https://rustc-dev-guide.rust-lang.org/backend/monomorph.html#monomorphization)) simply because the header type is different.

It seems that `new_read` is not a good name, perhaps:
```rust
        // This wraps `new_impl` to prevent unnecessary monomorphization caused by `H`.
        Self::new_impl(header.as_bytes(), packet, pool)
```
?

StevenJiang1110: Thanks for your explanation! It makes sense to me.

> Self::new_impl(header.as_bytes(), packet, pool)

Yes, `new_impl` is a better name for this.

PR #2987, opened 2026-02-28
Implement cgroup pid sub-controller

cchanging:
This PR implements the cgroup PIDs sub-controller, supporting three attributes: `pids.current`, `pids.max`, and `pids.peak`, which are required by Kata.

This PR also replaces the original `Mutex` in `CgroupMembership` with an `RwMutex`, allowing concurrent `fork` operations by only acquiring a read lock during `fork`, while explicit moves require a write lock.

### DIFF
--- a/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
@@ -1,13 +1,18 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use alloc::{collections::vec_deque::VecDeque, sync::Arc};
-use core::{fmt::Display, str::FromStr};
+use core::{
+    fmt::Display,
+    str::FromStr,
+    sync::atomic::{AtomicU8, Ordering},
+};
 
 use aster_systree::{Error, Result, SysAttrSetBuilder, SysBranchNode, SysObj};
+use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
 use bitflags::bitflags;
 use ostd::{
     mm::{VmReader, VmWriter},
-    sync::{Mutex, MutexGuard, Rcu},
+    sync::Rcu,
 };
 
 use crate::fs::cgroupfs::{
@@ -123,6 +128,24 @@ impl From<SubCtrlType> for SubCtrlSet {
     }
 }
 
+impl From<u8> for SubCtrlSet {
+    fn from(value: u8) -> Self {
+        Self::from_bits_truncate(value)
+    }
+}
+
+impl From<SubCtrlSet> for u8 {
+    fn from(value: SubCtrlSet) -> Self {
+        value.bits()
+    }
+}
+
+define_atomic_version_of_integer_like_type!(SubCtrlSet, {
+    /// An atomic version of `SubCtrlSet`.
+    #[derive(Debug)]
+    struct AtomicSubCtrlSet(AtomicU8);
+});
+
 /// The sub-controller for a specific cgroup controller type.
 ///
 /// If the sub-controller is inactive, the `inner` field will be `None`.
@@ -135,9 +158,9 @@ struct SubController<T: SubControlStatic> {
 }
 
 impl<T: SubControlStatic> SubController<T> {
-    fn new(parent_controller: Option<&LockedController>) -> Self {
+    fn new(parent_controller: Option<&Controller>) -> Self {
         let is_active = if let Some(parent) = parent_controller {
-            parent.active_set.contains_type(T::type_())
+            parent.active_set().contains_type(T::type_())
         } else {
             true
         };
@@ -148,7 +171,7 @@ impl<T: SubControlStatic> SubController<T> {
             None
         };
 
-        let parent = parent_controller.map(|controller| T::read_from(controller.controller));
+        let parent = parent_controller.map(T::read_from);
 
         Self { inner, parent }
     }
@@ -179,7 +202,9 @@ impl<T: SubControlStatic> TryGetSubControl for SubController<T> {
 /// sub-control, its corresponding sub-controller will be activated.
 pub struct Controller {
     /// A set of types of active sub-controllers.
-    active_set: Mutex<SubCtrlSet>,
+    ///
+    /// Updates to this set are serialized by `CgroupMembership::write_lock()`.
+    active_set: AtomicSubCtrlSet,
 
     memory: Rcu<Arc<SubController<MemoryController>>>,
     cpuset: Rcu<Arc<SubController<CpuSetController>>>,
@@ -188,13 +213,13 @@ pub struct Controller {
 
 impl Controller {
     /// Creates a new controller manager for a cgroup.
-    pub(super) fn new(locked_parent_controller: Option<&LockedController>) -> Self {
-        let memory_controller = Arc::new(SubController::new(locked_parent_controller));
-        let cpuset_controller = Arc::new(SubController::new(locked_parent_controller));
-        let pids_controller = Arc::new(SubController::new(locked_parent_controller));
+    pub(super) fn new(parent_controller: Option<&Controller>) -> Self {
+        let memory_controller = Arc::new(SubController::new(parent_controller));
+        let cpuset_controller = Arc::new(SubController::new(parent_controller));
+        let pids_controller = Arc::new(SubController::new(parent_controller));
 
         Self {
-            active_set: Mutex::new(SubCtrlSet::empty()),
+            active_set: AtomicSubCtrlSet::new(SubCtrlSet::empty()),
             memory: Rcu::new(memory_controller),
             cpuset: Rcu::new(cpuset_controller),
             pids: Rcu::new(pids_controller),
@@ -205,13 +230,6 @@ impl Controller {
         MemoryController::init_attr_set(builder, is_root);
         CpuSetController::init_attr_set(builder, is_root);
         PidsController::init_attr_set(builder, is_root);
-    }
-
-    pub(super) fn lock(&self) -> LockedController<'_> {
-        LockedController {
-            active_set: self.active_set.lock(),
-            controller: self,
-        }
     }
 
     fn read_sub(&self, ctrl_type: SubCtrlType) -> Arc<dyn TryGetSubControl> {
@@ -271,6 +289,150 @@ impl Controller {
         };
 
         controller.write_attr(name, reader)
+    }
+
+    /// Activates a sub-control of the specified type.
+    pub(super) fn activate(
+        &self,
+        ctrl_type: SubCtrlType,
+        current_node: &dyn CgroupSysNode,
+        parent_controller: Option<&Controller>,
+        cgroup_membership: &mut CgroupMembership,
+    ) -> Result<()> {
+        let mut active_set = self.active_set();
+        if active_set.contains_type(ctrl_type) {
+            return Ok(());
+        }
+
+        // A cgroup can activate the sub-control only if this
+        // sub-control has been activated in its parent cgroup.
+        if parent_controller
+            .is_some_and(|controller| !controller.active_set().contains_type(ctrl_type))
+        {
+            return Err(Error::NotFound);
+        }
+
+        active_set.add_type(ctrl_type);
+        self.active_set.store(active_set, Ordering::Relaxed);
+        self.update_sub_controllers_for_descents(ctrl_type, current_node, cgroup_membership);
+
+        Ok(())
+    }
+
+    /// Deactivates a sub-control of the specified type.
+    pub(super) fn deactivate(
+        &self,
+        ctrl_type: SubCtrlType,
+        current_node: &dyn CgroupSysNode,
+        cgroup_membership: &mut CgroupMembership,
+    ) -> Result<()> {
+        let mut active_set = self.active_set();
+        if !active_set.contains_type(ctrl_type) {
+            return Ok(());
+        }
+
+        // If any child node has activated this sub-control,
+        // the deactivation operation will be rejected.
+        for child in current_node.children() {
+            let cgroup_child = Arc::downcast::<CgroupNode>(child).unwrap();
+            // This is race-free because if a child wants to activate a sub-controller, it must
+            // acquire the write lock of `CgroupMembership` which has already been held here.
+            if cgroup_child
+                .controller()
+                .active_set()
+                .contains_type(ctrl_type)
+            {
+                return Err(Error::InvalidOperation);
+            }
+        }
+
+        active_set.remove_type(ctrl_type);
+        self.active_set.store(active_set, Ordering::Relaxed);
+        self.update_sub_controllers_for_descents(ctrl_type, current_node, cgroup_membership);
+
+        Ok(())
+    }
+
+    fn update_sub_controllers_for_descents(
+        &self,
+        ctrl_type: SubCtrlType,
+        current_node: &dyn CgroupSysNode,
+        cgroup_membership: &mut CgroupMembership,
+    ) {
+        fn update_sub_controller_for_one_child(
+            child: &Arc<dyn SysObj>,
+            ctrl_type: SubCtrlType,
+            parent_controller: &Controller,
+            cgroup_membership: &mut CgroupMembership,
+        ) {
+            let child_node = child.as_any().downcast_ref::<CgroupNode>().unwrap();
+            match ctrl_type {
+                SubCtrlType::Memory => {
+                    let new_controller = Arc::new(SubController::new(Some(parent_controller)));
+                    child_node.controller().memory.update(new_controller);
+                }
+                SubCtrlType::CpuSet => {
+                    let new_controller = Arc::new(SubController::new(Some(parent_controller)));
+                    child_node.controller().cpuset.update(new_controller);
+                }
+                SubCtrlType::Pids => {
+                    let mut new_controller: SubController<PidsController> =
+                        SubController::new(Some(parent_controller));
+                    if let Some(inner) = new_controller.inner.as_mut() {
+                        // When the pids sub-controller is being activated, initialize
+                        // `pids.current` with the number of processes already present
+                        // in this cgroup's subtree. The parent's counter is already
+                        // correct because charges propagated through inactive levels.
+                        let count = cgroup_membership.count_subtree_processes(child_node);
+                        if count > 0 {
+                            inner.init_count(count);
+                        }
+                    }
+                    child_node
+                        .controller()
+                        .pids
+                        .update(Arc::new(new_controller));
+                }
+            }
+        }
+
+        let mut descents = VecDeque::new();
+
+        // Subtree-control writes hold `CgroupMembership::write_lock()`, so operations
+        // protected by `CgroupMembership` like controller activation and deactivation
+        // and child node creation are serialized while the subtree updates are applied.
+        //
+        // Concurrent child removal is still fine: If a child is removed, we may update
+        // a sub-controller that's about to be destroyed, which is harmless.
+
+        // Update the direct children first.
+        current_node.visit_children_with(0, &mut |child_node| {
+            descents.push_back(child_node.clone());
+            update_sub_controller_for_one_child(child_node, ctrl_type, self, cgroup_membership);
+
+            Some(())
+        });
+
+        // Then update all the other descendent nodes.
+        while let Some(node) = descents.pop_front() {
+            let current_node = Arc::downcast::<CgroupNode>(node).unwrap();
+            let current_controller = current_node.controller();
+            current_node.visit_children_with(0, &mut |child_node| {
+                descents.push_back(child_node.clone());
+                update_sub_controller_for_one_child(
+                    child_node,
+                    ctrl_type,
+                    current_controller,
+                    cgroup_membership,
+                );
+
+                Some(())
+            });
+        }
+    }
+
+    pub(super) fn active_set(&self) -> SubCtrlSet {
+        self.active_set.load(Ordering::Relaxed)
     }
 }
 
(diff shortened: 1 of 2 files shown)


kernel/src/fs/cgroupfs/systree_node.rs
```diff
@@ -70,18 +80,21 @@ use crate::{
...
 /// This struct encapsulates the logic for moving processes between cgroups.
-/// By calling `CgroupMembership::lock()`, a thread can attempt to acquire a lock
-/// on the global instance. Upon success, it returns a guard that provides mutable
-/// access, allowing for safe cgroup membership modifications.
+/// The global instance is protected by an [`RwMutex`]:
+/// - **Read lock** ([`CgroupMembership::read_lock`]): used during the fork
+///   path to allow concurrent forks while still preventing sub-controller
+///   activation/deactivation from observing an inconsistent process count.
+/// - **Write lock** ([`CgroupMembership::write_lock`]): used for explicit
+///   migration, process exit, and sub-controller activation/deactivation.
```
lrh2000: It seems that the same thing is repeated three times in this PR. Additionally, it may be difficult to properly maintain a list of all the usage examples since it's easy to forget to update them when updating the code.

It would be better to figure out a model that explains when to use the read lock and when to use the write lock. Do you actually have this model in mind?

In your PR description:
> This PR also replaces the original `Mutex` in `CgroupMembership` with an `RwMutex`, allowing concurrent `fork` operations by only acquiring a read lock during `fork`, while explicit moves require a write lock.

This just repeats the code and doesn't help at all. Explain **why** we can write the code this way instead of describing **what** the code does.

I don't understand the difference between fork and explicit migration, nor do I understand why they take different types of locks. Normal read-write locks require an exclusive lock for mutable operations. Both operations are mutable. But you don't seem to follow normal read-write locks' semantics here, it's actually more like (1) _shared locking for all operations_ and (2) _exclusive locking for a stable view_ as it prevents all concurrent operations.

kernel/src/fs/cgroupfs/controller/mod.rs
```diff
@@ -275,6 +274,43 @@ impl Controller {
     }
 }
 
+// For pid sub-controller
```
lrh2000:
```suggestion
// For pids sub-controller
```

kernel/src/fs/cgroupfs/controller/mod.rs
```diff
@@ -358,7 +394,18 @@ impl LockedController<'_> {
                     child_node.controller().cpuset.update(new_controller);
                 }
                 SubCtrlType::Pids => {
-                    let new_controller = SubController::new(Some(parent_controller));
+                    let new_controller: Arc<SubController<PidsController>> =
+                        SubController::new(Some(parent_controller));
+                    if let Some(inner) = new_controller.inner.as_ref() {
+                        // When the pids sub-controller is being activated, initialize
+                        // pids.current with the number of processes already present
```
lrh2000:
```suggestion
                        // `pids.current` with the number of processes already present
```

kernel/src/fs/cgroupfs/systree_node.rs
```diff
@@ -671,6 +768,7 @@ inherit_sys_branch_node!(CgroupNode, fields, {
             "cgroup.subtree_control" => {
                 let (activate_set, deactivate_set, len) = read_subtree_control_from_reader(reader)?;
 
+                let _cgroup_guard = CgroupMembership::write_lock();
```
lrh2000: Can we pass the guard as an argument and use `guard.count_subtree_processes` to avoid calling that function without holding the guard?

kernel/src/fs/fs_impls/cgroupfs/controller/pids.rs
```diff
@@ -13,30 +13,84 @@ use crate::{process::posix_thread::PID_MAX, util::ReadCString};
...
 
 impl PidsController {
     pub(super) fn init_attr_set(builder: &mut SysAttrSetBuilder, is_root: bool) {
         if !is_root {
             builder.add(SysStr::from("pids.max"), SysPerms::DEFAULT_RW_ATTR_PERMS);
+            builder.add(
+                SysStr::from("pids.current"),
+                SysPerms::DEFAULT_RO_ATTR_PERMS,
+            );
+            builder.add(SysStr::from("pids.peak"), SysPerms::DEFAULT_RO_ATTR_PERMS);
```
lrh2000, line 33: https://github.com/asterinas/asterinas/pull/2727 sorts the cgroupfs attributes. I'm not sure if it still makes sense to follow it now.

lrh2000 (commented):
Thanks for implementing pids cgroup limit!

kernel/src/fs/cgroupfs/systree_node.rs
cchanging: > I don't understand the difference between fork and explicit migration, nor do I understand why they take different types of locks. 

Sorry for not making it clear. The difference between fork and explicit migration operations is that during the fork operation, the cgroup of the existing process will not be changed. It only changes the cgroup of the newly created process. Therefore, when multiple fork processes are executed in parallel, as long as there are no other explicit migration operations, the cgroup of the parent process performing these forks will not change, thus we will still have a stable cgroup view.  

This PR performs the charge operation before creating the actual child process, hence it requires acquiring the global lock `CgroupMembership` in advance and holding it for a longer time. If this lock is still a `Mutex`, I think it might cause performance bottlenecks in scenarios with frequent forks.

I will modify the documentation to provide a better explanation.

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -149,6 +223,12 @@ impl CgroupMembership {
             .ok_or(Error::IsDead)?;
         drop(controller);
```
lrh2000: This seems to be the only part that is useful with `move_process_to_node_inner` if `pids_precharged == false`. Consider to extract it to a method and use it in `move_forked_process_to_node`. There is no need to introduce `move_process_to_node_inner` after that?

Also you seem to unwrap the result at `clone.rs`. Let's unwrap it here so it's easy to maintain when modifying the related code.

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -189,6 +272,9 @@ impl CgroupMembership {
                 }
             })
             .unwrap();
+
+        // Notify the pids sub-controller about the removed process.
+        old_cgroup.controller.uncharge_pids();
```
lrh2000, line 267: Use the same comment:
https://github.com/asterinas/asterinas/blob/a5b969092e5c9c7b5425a05c80766afea458c431/kernel/src/fs/fs_impls/cgroupfs/systree_node.rs#L246-L247

kernel/src/process/clone.rs
```diff
@@ -289,16 +289,43 @@ pub fn clone_child(
...
+        // Pre-charge the pids controller before creating the child process.
+        // This enforces `pids.max` at fork time per cgroupv2 semantics.
+        // The charge must happen before process creation so that on failure
+        // we can return EAGAIN without leaving an orphaned process.
+        let parent_cgroup = ctx.process.cgroup().get().map(|cgroup| cgroup.clone());
+        if let Some(ref cgroup) = parent_cgroup {
+            cgroup.controller().try_charge_pids().map_err(|_| {
+                Error::with_message(
+                    Errno::EAGAIN,
+                    "failed to fork process due to pids controller limit",
```
lrh2000:
```suggestion
                    "the pids sub-controller limit is reached",
```

kernel/src/process/clone.rs
```diff
@@ -289,16 +289,43 @@ pub fn clone_child(
...
+                )
+            })?;
+        }
+
+        let child_process = match clone_child_process(ctx, parent_context, clone_args) {
+            Ok(child) => child,
+            Err(e) => {
```
lrh2000:
```suggestion
            Err(err) => {
```

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -92,18 +97,65 @@ pub struct CgroupMembership {
...
+            .unwrap_or(0);
+        cgroup_node.visit_children_with(0, &mut |child| {
+            stack.push(child.clone());
+            Some(())
+        });
+
+        while let Some(node) = stack.pop() {
+            let Some(cgroup_node) = node.as_any().downcast_ref::<CgroupNode>() else {
+                continue;
+            };
```
lrh2000: https://github.com/asterinas/asterinas/blob/a5b969092e5c9c7b5425a05c80766afea458c431/kernel/src/fs/fs_impls/cgroupfs/systree_node.rs#L475

Use a consistent style. Also it cannot fail?

kernel/src/fs/fs_impls/cgroupfs/controller/pids.rs
```diff
@@ -91,3 +144,61 @@ impl super::SubControlStatic for PidsController {
...
+
+    /// Tries to charge one process across the hierarchy with limit checking.
+    ///
+    /// Walks from this cgroup upward to the root. At each level where the
+    /// pids sub-controller is active, the counter is incremented and checked
+    /// against `pids.max`. If any level exceeds its limit, all previously
+    /// charged levels are rolled back and `Err` is returned.
+    ///
+    /// This is used at fork time where `pids.max` must be enforced.
+    pub(super) fn try_charge_hierarchy(&self) -> Result<()> {
```
lrh2000: `Result<(), TryChargeError>` since it can only return `Error::ResourceUnavailable`?

Or a boolean with `#[must_use]`..?

test/initramfs/src/apps/process/cgroup.sh
```diff
@@ -8,124 +8,444 @@ CGROUP_ROOT="/sys/fs/cgroup"
...
+echo "pids.peak after children exit: $PEAK_AFTER"
+
+if [ "$PEAK_AFTER" -lt "$PEAK_DURING" ]; then
+    echo -e "Error: pids.peak decreased after exit (should be monotonically non-decreasing)!"
+    exit 1
+else
+    echo -e "Verified: pids.peak retained"
+fi
+
+# ─── Section 5: Teardown ──────────────────────────────────────────────────────
```
lrh2000: Use the same style:
```bash
# --- Section 4: pids controller -----------------------------------------------
```

kernel/src/process/clone.rs
```diff
@@ -289,16 +289,43 @@ pub fn clone_child(
         let child_tid = child_thread.as_posix_thread().unwrap().tid();
         Ok(child_tid)
     } else {
-        let child_process = clone_child_process(ctx, parent_context, clone_args)?;
+        // Hold the read lock before charge to ensure the cgroup of current process
+        // won't change during the charge and the subsequent move operation.
+        let cgroup_read_guard = CgroupMembership::read_lock();
+
+        // Pre-charge the pids controller before creating the child process.
```
lrh2000:
```suggestion
        // Pre-charge the pids sub-controller before creating the child process.
```

test/initramfs/src/apps/process/cgroup.sh
```diff
@@ -8,124 +8,444 @@ CGROUP_ROOT="/sys/fs/cgroup"
...
+echo "-pids" > cgroup.subtree_control
+echo $PROCESS_ID > cgroup.procs
+verify "Process 1 added to user hierarchy" \
+    "cat cgroup.procs | grep -w $PROCESS_ID" \
+    "$PROCESS_ID"
 
-log_step "10. Return to parent directory"
-cd ..
-echo "Current directory: $(pwd)"
+# --- Section 4: pids controller -----------------------------------------------
```
lrh2000: sub-controller

kernel/src/fs/fs_impls/cgroupfs/controller/pids.rs
```diff
@@ -13,30 +13,84 @@ use crate::{process::posix_thread::PID_MAX, util::ReadCString};
...
+                let max_pids = self.max_pids.load(Ordering::Relaxed);
+                if max_pids == u32::MAX {
                     writeln!(printer, "max")?;
                 } else {
-                    writeln!(printer, "{}", max_pid)?;
+                    writeln!(printer, "{}", max_pids)?;
                 }
             }
-            // TODO: Add support for reading other attributes.
+            "pids.current" => {
```
lrh2000: This list should also be sorted if you decide to address my previous comment.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -275,6 +274,43 @@ impl Controller {
     }
 }
 
+// For pids sub-controller
+impl Controller {
+    /// Charge a process in the pids sub-controller hierarchy.
```
lrh2000:
```suggestion
    /// Charges a process in the pids sub-controller hierarchy.
```

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -275,6 +274,43 @@ impl Controller {
...
+    /// Note that concurrent operations of moving processes to cgroups and
+    /// setting `pids.max` may cause the actual charge to exceed the peak at
+    /// charge time.
+    pub fn try_charge_pids(&self) -> Result<()> {
+        let guard = self.pids.read();
+        let sub = guard.get();
+        sub.try_charge_hierarchy()
+    }
+
+    /// Uncharges a process in the pids sub-controller hierarchy
```
lrh2000:
```suggestion
    /// Uncharges a process in the pids sub-controller hierarchy.
```

kernel/src/fs/cgroupfs/systree_node.rs
lrh2000: These are somehow odd rules at my first sight. But I cannot think of problems. So maybe they're right and reasonable rules. Thanks for your explanation.

kernel/src/fs/fs_impls/cgroupfs/controller/pids.rs
```diff
@@ -1,42 +1,97 @@
...
     }
+
+    /// Sets the initial process count during sub-controller activation.
+    ///
+    /// When the pids sub-controller is activated on a cgroup that already
+    /// has processes, this method initializes `current_count` and
+    /// `peak_count` to reflect the existing state.
+    pub(super) fn init_count(&self, count: u32) {
+        self.current_count.store(count, Ordering::Relaxed);
+        self.peak_count.store(count, Ordering::Relaxed);
```
lrh2000: As an initialization method, this method should require `&mut self`. Use [`AtomicU8::get_mut`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU8.html#method.get_mut) to modify the counters.

I believe that it isn't hard to make `SubController::new` return an `SubController` without `Arc`: 
```rust
                    let new_controller = SubController::new(Some(parent_controller));
                    child_node.controller().cpuset.update(Arc::new(new_controller));
```

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -124,8 +174,40 @@ impl CgroupMembership {
             None
         };
 
-        // Try to add the process to the new cgroup first.
```
lrh2000, line 127: Keep this comment?

We still have the following comment anyway:
```rust
        // Remove the process from the old cgroup second.
```

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -124,8 +174,40 @@ impl CgroupMembership {
...
+        Ok(())
+    }
+
+    /// Moves a newly forked process into its inherited cgroup node.
+    ///
+    /// This is the fork-path variant of [`CgroupMembership::move_process_to_node`].
+    /// It assumes the caller already holds the read side of the membership lock,
+    /// and that the `CgroupNode` has already reserved a pids slot during fork.
+    pub fn move_forked_process_to_node(&self, process: Arc<Process>, new_cgroup: &CgroupNode) {
+        self.add_process_to_node(&process, new_cgroup).unwrap();
```
lrh2000: While we cannot assert that it's a newly forked process, we can at least assert that it is not related to an existing cgroup.

```rust
    pub fn move_forked_process_to_node(&self, process: Arc<Process>, new_cgroup: &CgroupNode) {
        debug_assert!(process.cgroup().get().is_none());

        self.add_process_to_node(&process, new_cgroup).unwrap();
    }
```

test/initramfs/src/apps/process/cgroup.sh
```diff
@@ -8,124 +8,444 @@ CGROUP_ROOT="/sys/fs/cgroup"
...
+# Re-enable the pids controller at the root level.
+echo "+pids" > "$CGROUP_ROOT/cgroup.subtree_control"
+echo "pids controller re-enabled"
+
+# After the toggle, pids.current must equal the number of processes still
+# present in cgroup.procs.
+ACTUAL_AFTER=$(wc -l < "$CGROUP_ROOT/$CGROUP_NAME/cgroup.procs")
+REPORTED_AFTER=$(cat "$CGROUP_ROOT/$CGROUP_NAME/pids.current")
+echo "Processes in cgroup after toggle: $ACTUAL_AFTER"
+echo "pids.current after toggle: $REPORTED_AFTER"
```
lrh2000, line 369: This does not seem to check anything between `ACTUAL_AFTER` and `REPORTED_AFTER`?

test/initramfs/src/apps/process/cgroup.sh
```diff
@@ -8,124 +8,444 @@ CGROUP_ROOT="/sys/fs/cgroup"
...
+verify "Process 1 added to user hierarchy" \
+    "cat cgroup.procs | grep -w $PROCESS_ID" \
+    "$PROCESS_ID"
 
-log_step "10. Return to parent directory"
-cd ..
-echo "Current directory: $(pwd)"
+# --- Section 4: pids sub-controller -------------------------------------------
+
+log_section "Section 4: pids controller"
```
lrh2000:
```suggestion
log_section "Section 4: pids sub-controller"
```

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -358,7 +402,18 @@ impl LockedController<'_> {
                     child_node.controller().cpuset.update(new_controller);
                 }
                 SubCtrlType::Pids => {
-                    let new_controller = SubController::new(Some(parent_controller));
+                    let new_controller: Arc<SubController<PidsController>> =
```
lrh2000: Also, I don't see a reason why we should continue to wrap the `PidsController` in an `Arc`, since a `Box` should work just fine.

However, `MemoryController` should continue to be wrapped in an `Arc` because it exists for a reason; for example, resource handles will need to clone and hold that `Arc`. What about the `CpusetController`?

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -275,6 +274,47 @@ impl Controller {
...
 
+// For pids sub-controller
+impl Controller {
+    /// Charges a process in the pids sub-controller hierarchy.
+    ///
+    /// This operation is used for explicit migration and will not enforce
+    /// the `pids.max` limit.
+    ///
+    /// Reference: <https://docs.kernel.org/admin-guide/cgroup-v2.html#pid>
+    pub fn charge_pids(&self) {
```
lrh2000: This should be `pub(super)` since it should only be used during explicit migration?

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
lrh2000: Oh, perhaps it still needs to exist because `read_sub`..

cchanging: Yes, I remember that in the controller PR I didn't find a good way to avoid the `clone` in `read_sub`...😂

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -275,6 +274,47 @@ impl Controller {
...
+    /// Tries to charge a process in the pids sub-controller hierarchy,
+    /// enforcing `pids.max` at each level.
+    ///
+    /// This is used at fork time. Returns `Err` if any ancestor's limit
+    /// would be exceeded; all intermediate charges are rolled back.
+    ///
+    /// Note that concurrent operations of moving processes to cgroups and
+    /// setting `pids.max` may cause the actual charge to exceed the peak at
+    /// charge time.
+    pub fn try_charge_pids(&self) -> core::result::Result<(), TryChargeError> {
```
lrh2000: **(Free feel to revise the type names/method names if you can come up with better ones).**

How do you like this:
```rust
// For pids sub-controller
impl Controller {
    pub fn pre_charge_pids(&self)
        -> core::result::Result<PidsPreCharge<'_>, TryChargeError>;

    fn uncharge_pids(&self) {}
}
```

Then define `PidsPreCharge` as something like:
```rust
struct PidsPreCharge<'a> {
}

impl PidsPreCharge {
    pub(super) fn apply(self) {
        core::mem::forget(self);
    }
}

impl Drop for PidsPreCharge {
    fn drop(&mut self) {
        self.uncharge_pids();
    }
}
```

Finally, revise `move_forked_process_to_node`:
```rust
    pub fn move_forked_process_to_node(
        &self,
        process: Arc<Process>,
        new_cgroup: &CgroupNode
        pids_charge: PidsPreCharge,
    ) {
        pids_charge.apply();
        self.add_process_to_node(..).unwrap();
    }
```

We don't need to expose `uncharge_pids()` as a public method because it's not suitable for public use.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
lrh2000: OK. This issue was not introduced in this PR and can be considered out of scope.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -275,6 +274,67 @@ impl Controller {
...
+    /// Pre-charges a process in the pids sub-controller hierarchy,
+    /// enforcing `pids.max` at each level.
+    ///
+    /// This is used at fork time. Returns a guard that will roll back the
+    /// charge unless it is explicitly applied.
+    ///
+    /// Note that concurrent operations of moving processes to cgroups and
+    /// setting `pids.max` may cause the actual charge to exceed the peak at
+    /// charge time.
+    pub fn pre_charge_pids(&self) -> core::result::Result<PidsPreCharge<'_>, TryChargeError> {
```
lrh2000: Oh, I just realize that we must drop (or apply) `PidsPreCharge` before dropping `cgroup_read_guard`. Otherwise, a rollback will occur at an incorrect pids sub-controller.

This is quite tricky and may be accidentally overlooked if we refactor the code incorrectly. So I suggest enforcing this via a Rust lifetime. Also add a comment to explain this fact:
```rust
    pub fn pre_charge_pids<'a>(&'a self, _cgroup_membership: &'a CgroupMembership)
        -> core::result::Result<PidsPreCharge<'a>, TryChargeError>
    {
        let guard = self.pids.read();
        let sub = guard.get();
        sub.try_charge_hierarchy()?;

        // This will not outlive `_cgroup_membership`, so we will roll back
        // the charge for the right pids sub-controller.
        Ok(PidsPreCharge { controller: self })
    }
```

We should have a `else`-branch in `close.rs` to drop the pre-charge to make the compiler happy, but that should be expected:
```patch
diff --git a/kernel/src/process/clone.rs b/kernel/src/process/clone.rs
index e8d667a08..53c0d2066 100644
--- a/kernel/src/process/clone.rs
+++ b/kernel/src/process/clone.rs
@@ -299,9 +299,13 @@ pub fn clone_child(
         // we can return EAGAIN without leaving an orphaned process.
         let parent_cgroup = ctx.process.cgroup().get().map(|cgroup| cgroup.clone());
         let pids_charge = if let Some(ref cgroup) = parent_cgroup {
-            Some(cgroup.controller().pre_charge_pids().map_err(|_| {
-                Error::with_message(Errno::EAGAIN, "the pids sub-controller limit is reached")
-            })?)
+            let pids_charge = cgroup
+                .controller()
+                .pre_charge_pids(&cgroup_read_guard)
+                .map_err(|_| {
+                    Error::with_message(Errno::EAGAIN, "the pids sub-controller limit is reached")
+                })?;
+            Some(pids_charge)
         } else {
             None
         };
@@ -316,6 +320,8 @@ pub fn clone_child(
                 cgroup,
                 pids_charge.unwrap(),
             );
+        } else {
+            drop(pids_charge);
         }
         drop(cgroup_read_guard);
 
```

lrh2000:
These **new** changes look good.

We now require `CgroupMembership::write_lock()` for subtree controls. This means that the controller lock is redundant:
https://github.com/asterinas/asterinas/blob/a34449a325d27dfc4cae58080f63da4839b90623/kernel/src/fs/fs_impls/cgroupfs/systree_node.rs#L17-L19

This can be made to an atomic variable:
https://github.com/asterinas/asterinas/blob/a34449a325d27dfc4cae58080f63da4839b90623/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs#L181-L182

Comments can be simplified due to the new fact:
https://github.com/asterinas/asterinas/blob/a34449a325d27dfc4cae58080f63da4839b90623/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs#L394-L396
https://github.com/asterinas/asterinas/blob/a34449a325d27dfc4cae58080f63da4839b90623/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs#L452-L468
https://github.com/asterinas/asterinas/blob/a34449a325d27dfc4cae58080f63da4839b90623/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs#L481-L483

Due to the new changes, we should perform some cleanup. Do it in a separate commit or a separate PR if you want this PR to be merged first? I will approve this PR if you want to do it in a separate PR.

cchanging:
> We now require CgroupMembership::write_lock() for subtree controls. This means that the controller lock is redundant

Thanks for pointing this out! It does make sense.

> Do it in a separate commit or a separate PR if you want this PR to be merged first?

I'll add a separate commit to implement this change.

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -648,8 +644,13 @@ inherit_sys_branch_node!(CgroupSystem, fields, {
     }
 
     fn create_child(&self, name: &str) -> Result<Arc<dyn SysObj>> {
-        let controller = self.controller.lock();
-        let new_child = CgroupNode::new(name.to_string().into(), 1, &controller);
+        let mut cgroup_membership = CgroupMembership::write_lock();
```
lrh2000: I think it's fine to take a read lock here. It just needs to ensure that `active_set()` is stable. It's just like `fork()`, where we only need to ensure that the parent process's cgroup is stable.

(See also https://github.com/asterinas/asterinas/pull/2987#discussion_r2922421556).

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -648,8 +644,13 @@ inherit_sys_branch_node!(CgroupSystem, fields, {
...
+        let new_child = CgroupNode::new(
+            name.to_string().into(),
+            1,
+            &self.controller,
+            &mut cgroup_membership,
```
lrh2000: Not really sure why we want `CgroupNode::new` to take the membership guard. `CgroupNode::new` is always fine, but it seems that the race condition from the interaction between `CgroupNode::new` and `self.add_child`:
```rust
let new_child = CgroupNode::new(...);
// If `active_set()` changes here... things will go wrong!
self.add_child(new_child);
```

It would be more helpful to add a comment explaining this fact than to let the `new` method accept a membership guard, which is confusing and inaccurate.

(See also https://github.com/asterinas/asterinas/pull/2987#discussion_r2922421556).

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -14,24 +14,22 @@
...
-//!    existing membership, or when it needs an exclusive subtree snapshot. For
-//!    example, fork uses the read lock, while explicit migration, process exit,
-//!    and `cgroup.subtree_control` writes use the write lock.
+//! 3. **Cgroup Membership Lock**: A global `RwMutex` managed by [`CgroupMembership`]
+//!    that serializes topology-sensitive cgroup operations. It protects both the
+//!    process-to-cgroup associations and the cgroup hierarchy changes that must
+//!    observe a stable subtree snapshot. Use the read lock when an operation only
+//!    needs such state to remain stable while it works from a snapshot; use the
+//!    write lock when an operation may change an existing membership, create or
+//!    destroy cgroup nodes, or update subtree controls. For example, fork uses
```
lrh2000: > the cgroup hierarchy changes

> use the write lock when [..] create or destroy cgroup nodes

First, this does not seem correct. I don't think you take the lock when destroying cgroup nodes?

Second, we may not be necessary to protect cgroup hierarchy changes with this lock: We need to take this lock at `create_child` **not** because we want the cgroup hierarchy to be stable, but because we want `active_set()` to be stable.

This also applies to several doc changes below.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -179,7 +202,7 @@ impl<T: SubControlStatic> TryGetSubControl for SubController<T> {
 /// sub-control, its corresponding sub-controller will be activated.
 pub struct Controller {
     /// A set of types of active sub-controllers.
-    active_set: Mutex<SubCtrlSet>,
+    active_set: AtomicSubCtrlSet,
```
lrh2000, line 207: Add note explaining what lock protects this atomic variable.

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -648,8 +640,12 @@ inherit_sys_branch_node!(CgroupSystem, fields, {
     }
 
     fn create_child(&self, name: &str) -> Result<Arc<dyn SysObj>> {
-        let controller = self.controller.lock();
-        let new_child = CgroupNode::new(name.to_string().into(), 1, &controller);
+        let cgroup_membership = CgroupMembership::read_lock();
+        // Keep `active_set()` stable across both steps. If the active
+        // sub-controllers changed after `CgroupNode::new()` but before
+        // `add_child()`, the new child could inherit stale controller
+        // activation state.
```
lrh2000: - Why describe the wrong behavior? I don't mean it. Perhaps my previous statement wasn't clear (The previous pseudocode is only for illustrating the problem). Sorry.
 - `inherit` is probably a wrong word to use here. The child does not inherit the state:
https://github.com/asterinas/asterinas/blob/12654564c1e932042b5037add5d4b4fa5923a7be/kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs#L222
 - Also the dead variable lint does not trigger here? Rename `cgroup_membership` to `_cgroup_read_guard`.
```suggestion
        // The child node's content depends on our sub-controller activation
        // state. Take a read lock to prevent any state changes.
        let _cgroup_read_guard = CgroupMembership::read_lock();
```

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -798,9 +793,14 @@ inherit_sys_branch_node!(CgroupNode, fields, {
...
 
     fn create_child(&self, name: &str) -> Result<Arc<dyn SysObj>> {
-        let controller = self.controller.lock();
+        let cgroup_membership = CgroupMembership::read_lock();
         self.with_inner(|_| {
-            let new_child = CgroupNode::new(name.to_string().into(), self.depth + 1, &controller);
+            // Keep `active_set()` stable across both steps. If the active
+            // sub-controllers changed after `CgroupNode::new()` but before
+            // `add_child()`, the new child could inherit stale controller
+            // activation state.
```
lrh2000: Same.

kernel/src/fs/fs_impls/cgroupfs/systree_node.rs
```diff
@@ -14,24 +14,23 @@
...
-//!    works from a snapshot; use the write lock when an operation may change an
-//!    existing membership, or when it needs an exclusive subtree snapshot. For
-//!    example, fork uses the read lock, while explicit migration, process exit,
-//!    and `cgroup.subtree_control` writes use the write lock.
+//! 3. **Cgroup Membership Lock**: A global `RwMutex` managed by [`CgroupMembership`]
+//!    that serializes updates to process-to-cgroup associations and
+//!    sub-controller activation state. It also lets readers take a stable
+//!    snapshot of those states when they need to derive cgroup behavior from
+//!    them. Use the read lock when an operation only needs the current process
+//!    memberships or `active_set()` values to stay stable while it works from a
```
lrh2000: Also, `active_set()` should be avoided in the documentation, especially in the module-level documentation. This seems odd to me.

I tried to improve this a bit, hopefully it can be better:
```c
//! 3. **Cgroup Membership Lock**: A global `RwMutex` managed by [`CgroupMembership`]
//!    that serializes updates to process-to-cgroup associations and sub-controller
//!    activation states. Only a writer can change cgroup associations for _existing_
//!    processes or modify sub-controller activation, leaving readers with a stable
//!    view of those states. Note that because readers cannot block _new_ processes
//!    from joining, only a writer has a truly stable view of all the processes
//!    within a cgroup subtree.
//!
//!    For example, fork and cgroup child creation use the read lock, while explicit
//!    migration, process exit, and sub-controller activation use the write lock.
```

PR #2988, opened 2026-02-28
Support IPC namespace

StevenJiang1110:
Based on #2966.

This PR adds basic support for IPC namespaces. An IPC namespace is used to isolate certain IPC resources, including semaphores, message queues (`/dev/mqueue`), and shared memory segments. Since only semaphores are currently supported in our implementation, the scope of this PR is limited to semaphore isolation accordingly.

Required by #2911.

### DIFF
--- a/kernel/src/ipc/ipc_ns.rs
+++ b/kernel/src/ipc/ipc_ns.rs
@@ -1,0 +1,255 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Defines the IPC namespace abstraction.
+//!
+//! An IPC namespace isolates System V IPC resources from other namespaces.
+//! It currently manages semaphore sets only, while message queues and shared
+//! memory remain to be added.
+//!
+//! Each namespace stores its semaphore sets in a per-namespace map keyed by
+//! IPC key and uses a dedicated ID allocator to assign semaphore identifiers.
+
+use aster_rights::ReadOp;
+use spin::Once;
+
+use super::{
+    ipc_ids::IpcIds,
+    key_t,
+    semaphore::system_v::{
+        PermissionMode,
+        sem_set::{SEMMNI, SemaphoreSet},
+    },
+};
+use crate::{
+    fs::pseudofs::{NsCommonOps, NsType, StashedDentry},
+    ipc::IpcFlags,
+    prelude::*,
+    process::{
+        Credentials, UserNamespace, credentials::capabilities::CapSet, posix_thread::PosixThread,
+    },
+};
+
+/// The IPC namespace.
+///
+/// An IPC namespace isolates System V IPC objects
+/// (semaphores, message queues, shared memory).
+/// Each namespace maintains its own independent set
+/// of IPC resources and identifier allocator.
+///
+/// Lock ordering:
+/// `sem_ids` -> `SemaphoreSet::inner`.
+pub struct IpcNamespace {
+    /// Semaphore sets within this namespace.
+    sem_ids: IpcIds<SemaphoreSet>,
+    /// Owner user namespace.
+    owner: Arc<UserNamespace>,
+    /// Stashed dentry for nsfs.
+    stashed_dentry: StashedDentry,
+}
+
+impl IpcNamespace {
+    /// Returns a reference to the singleton initial IPC namespace.
+    pub fn get_init_singleton() -> &'static Arc<IpcNamespace> {
+        static INIT: Once<Arc<IpcNamespace>> = Once::new();
+
+        INIT.call_once(|| {
+            let owner = UserNamespace::get_init_singleton().clone();
+            Self::new(owner)
+        })
+    }
+
+    fn new(owner: Arc<UserNamespace>) -> Arc<Self> {
+        let stashed_dentry = StashedDentry::new();
+
+        Arc::new(Self {
+            sem_ids: IpcIds::new(SEMMNI),
+            owner,
+            stashed_dentry,
+        })
+    }
+
+    /// Clones a new IPC namespace from `self`.
+    ///
+    /// The new namespace starts with an empty set of IPC resources.
+    pub fn new_clone(
+        &self,
+        owner: Arc<UserNamespace>,
+        posix_thread: &PosixThread,
+    ) -> Result<Arc<Self>> {
+        owner.check_cap(CapSet::SYS_ADMIN, posix_thread)?;
+        Ok(Self::new(owner))
+    }
+
+    /// Calls `op` with the semaphore set identified by `semid`.
+    pub fn with_sem_set<T, F>(
+        &self,
+        semid: key_t,
+        num_sems: Option<usize>,
+        required_perm: PermissionMode,
+        op: F,
+    ) -> Result<T>
+    where
+        F: FnOnce(&SemaphoreSet) -> Result<T>,
+    {
+        if semid <= 0 {
+            return_errno_with_message!(Errno::EINVAL, "semaphore ID must be positive");
+        }
+
+        self.sem_ids
+            .with(semid, |sem_set| {
+                Self::validate_sem_set(semid, sem_set, num_sems, required_perm)?;
+                op(sem_set)
+            })
+            .map_err(Self::map_sem_id_error)
+    }
+
+    /// Removes the semaphore set identified by `semid`.
+    pub fn remove_sem_set<F>(&self, semid: key_t, may_remove: F) -> Result<()>
+    where
+        F: FnOnce(&SemaphoreSet) -> Result<()>,
+    {
+        if semid <= 0 {
+            return_errno_with_message!(Errno::EINVAL, "semaphore ID must be positive");
+        }
+
+        self.sem_ids
+            .remove(semid, may_remove)
+            .map_err(Self::map_sem_id_error)
+    }
+
+    /// Returns the existing semaphore set or creates a new one.
+    pub fn get_or_create_sem_set(
+        self: &Arc<Self>,
+        key: key_t,
+        num_sems: usize,
+        flags: IpcFlags,
+        mode: u16,
+        credentials: Credentials<ReadOp>,
+    ) -> Result<key_t> {
+        const IPC_NEW: key_t = 0;
+
+        if key < 0 {
+            return_errno_with_message!(Errno::EINVAL, "semaphore key must not be negative");
+        }
+
+        if key == IPC_NEW || (key as usize > SEMMNI && flags.contains(IpcFlags::IPC_CREAT)) {
+            if num_sems == 0 {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "num_sems must not be zero when creating"
+                );
+            }
+            return self.create_sem_set(num_sems, mode, credentials);
+        }
+
+        let mut credentials = Some(credentials);
+        loop {
+            match self.sem_ids.with(key, |sem_set| {
+                Self::validate_sem_set(
+                    key,
+                    sem_set,
+                    Some(num_sems),
+                    PermissionMode::ALTER | PermissionMode::READ,
+                )?;
+                if flags.contains(IpcFlags::IPC_CREAT | IpcFlags::IPC_EXCL) {
+                    return_errno_with_message!(
+                        Errno::EEXIST,
+                        "semaphore set already exists with IPC_EXCL"
+                    );
+                }
+
+                Ok(key)
+            }) {
+                Ok(key) => return Ok(key),
+                Err(err) if err.error() == Errno::ENOENT && flags.contains(IpcFlags::IPC_CREAT) => {
+                }
+                Err(err) => return Err(err),
+            }
+
+            if num_sems == 0 {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "num_sems must not be zero when creating"
+                );
+            }
+
+            match self.sem_ids.insert_at(key, |key| {
+                let Some(credentials) = credentials.take() else {
+                    return_errno_with_message!(Errno::EINVAL, "credentials were already consumed");
+                };
+                SemaphoreSet::new(key, num_sems, mode, credentials, self)
+            }) {
+                Ok(()) => return Ok(key),
+                Err(err) if err.error() == Errno::EEXIST => continue,
+                Err(err) => return Err(err),
+            }
+        }
+    }
+
+    fn validate_sem_set(
+        key: key_t,
+        sem_set: &SemaphoreSet,
+        num_sems: Option<usize>,
+        required_perm: PermissionMode,
+    ) -> Result<()> {
+        if key <= 0 {
+            return_errno_with_message!(Errno::EINVAL, "semaphore key must be positive");
+        }
+
+        if let Some(num_sems) = num_sems
+            && num_sems > sem_set.num_sems()
+        {
+            return_errno_with_message!(Errno::EINVAL, "num_sems exceeds the set size");
+        }
+
+        if !required_perm.is_empty() {
+            // TODO: Support permission check
+            warn!("Semaphore doesn't support permission check now");
+        }
+
+        Ok(())
+    }
+
+    /// Creates a new semaphore set and returns its key.
+    fn create_sem_set(
+        self: &Arc<Self>,
+        num_sems: usize,
+        mode: u16,
+        credentials: Credentials<ReadOp>,
+    ) -> Result<key_t> {
+        self.sem_ids
+            .insert_auto(|key| SemaphoreSet::new(key, num_sems, mode, credentials, self))
+    }
+
+    /// Frees a semaphore key back to the allocator.
+    pub(super) fn free_sem_key(&self, key: key_t) {
+        self.sem_ids.free_key(key);
+    }
+
+    fn map_sem_id_error(err: Error) -> Error {
+        if err.error() == Errno::ENOENT {
+            Error::new(Errno::EINVAL)
+        } else {
+            err
+        }
+    }
+}
+
+impl NsCommonOps for IpcNamespace {
+    const TYPE: NsType = NsType::Ipc;
+
+    fn owner_user_ns(&self) -> Option<&Arc<UserNamespace>> {
+        Some(&self.owner)
+    }
+
+    fn parent(&self) -> Result<&Arc<Self>> {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "an IPC namespace does not have a parent namespace"
+        );
+    }
+
+    fn stashed_dentry(&self) -> &StashedDentry {
+        &self.stashed_dentry
+    }
+}
--- a/kernel/src/ipc/semaphore/system_v/sem_set.rs
+++ b/kernel/src/ipc/semaphore/system_v/sem_set.rs
@@ -1,19 +1,14 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use alloc::{collections::btree_map::BTreeMap, vec::Vec};
+use alloc::vec::Vec;
 use core::sync::atomic::{AtomicU64, Ordering};
 
 use aster_rights::ReadOp;
-use id_alloc::IdAlloc;
-use ostd::sync::{PreemptDisabled, RwLockReadGuard, RwLockWriteGuard};
-use spin::Once;
+use ostd::sync::PreemptDisabled;
 
-use super::{
-    PermissionMode,
-    sem::{PendingOp, Status, update_pending_alter, wake_const_ops},
-};
+use super::sem::{PendingOp, Status, update_pending_alter, wake_const_ops};
 use crate::{
-    ipc::{IpcPermission, key_t, semaphore::system_v::sem::Semaphore},
+    ipc::{IpcNamespace, IpcPermission, key_t, semaphore::system_v::sem::Semaphore},
     prelude::*,
     process::{Credentials, Pid},
     time::clocks::RealTimeCoarseClock,
@@ -39,7 +34,7 @@ pub const SEMAEM: i32 = SEMVMX;
 #[derive(Debug)]
 pub struct SemaphoreSet {
     /// Number of semaphores in the set
-    nsems: usize,
+    num_sems: usize,
     /// Inner
     inner: SpinLock<SemSetInner>,
     /// Semaphore permission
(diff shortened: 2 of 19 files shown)


kernel/src/syscall/semctl.rs
```diff
@@ -106,12 +105,17 @@ pub fn sys_semctl(
...
 where
     F: FnOnce(&SemaphoreSet) -> Result<T>,
 {
-    check_sem(semid, None, permission)?;
-    let sem_sets = sem_sets();
+    ipc_ns.check_sem(semid, None, permission)?;
+    let sem_sets = ipc_ns.sem_sets();
     let sem_set = sem_sets.get(&semid).ok_or(Error::new(Errno::EINVAL))?;
     ctl_func(sem_set)
 }
```
tatetian: [**`atomic-critical-sections`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/concurrency-and-races.html#atomic-critical-sections): `check_sem` acquires and releases the `semaphore_sets` read lock, then `sem_sets()` acquires it again. Between the two acquisitions another thread can call `IPC_RMID` and remove the set. Today the permission check is a no-op (`TODO`), but once it is implemented the race becomes a correctness bug: the check could pass on set A, A is removed, a new set B is created with the same key but different permissions, and `ctl_func` operates on B without authorization. The check and get operation should occur under a single lock acquisition. Need to reconsider the API design.

kernel/src/syscall/semget.rs
```diff
@@ -58,10 +64,10 @@ pub fn sys_semget(key: i32, nsems: i32, semflags: i32, ctx: &Context) -> Result<
...
             }
             if nsems == 0 {
-                return_errno!(Errno::EINVAL);
+                return_errno_with_message!(Errno::EINVAL, "nsems must not be zero when creating");
             }
 
-            create_sem_set_with_id(key, nsems, mode, credentials)?
+            ipc_ns.create_sem_set_with_key(key, nsems, mode, credentials)?
         }
     };
```
tatetian: [**`atomic-critical-sections`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/concurrency-and-races.html#atomic-critical-sections): Same TOCTOU pattern. `check_sem` returns `ENOENT`, the lock is released, then `create_sem_set_with_key` re-acquires it. Between the two calls another thread could create the same key, causing `alloc_specific` to fail with `EEXIST`. Conversely, on the `Ok` path the semaphore could be removed before the syscall returns its key. For `semget` with `IPC_CREAT` (without `IPC_EXCL`), Linux would atomically return the existing set; here the caller gets a spurious `EEXIST`. The check-and-create logic should be performed under a single lock acquisition.

kernel/src/ipc/ipc_ns.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub fn sem_sets(&self) -> RwLockReadGuard<'_, BTreeMap<key_t, SemaphoreSet>, PreemptDisabled> {
+        self.semaphore_sets.read()
+    }
+
+    /// Acquires write access to the semaphore sets.
+    pub fn sem_sets_mut(
+        &self,
+    ) -> RwLockWriteGuard<'_, BTreeMap<key_t, SemaphoreSet>, PreemptDisabled> {
+        self.semaphore_sets.write()
+    }
```
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): These methods expose `RwLockReadGuard<BTreeMap<...>, PreemptDisabled>` in the public API, leaking three implementation details: lock type, collection type, and lock mode. Callers in `semctl.rs` and `sem.rs` use these to perform lookups and removals -- operations that should be encapsulated by `IpcNamespace`. Providing higher-level methods (e.g., `get_sem_set`, `remove_sem_set`) would also eliminate the TOCTOU races above, since the check-and-act logic could be performed under a single lock acquisition inside `IpcNamespace`.

kernel/src/ipc/semaphore/system_v/sem_set.rs
```diff
@@ -282,97 +288,10 @@ impl Drop for SemaphoreSet {
...
-    let sem_set = sem_sets.get(&id).ok_or(Error::new(Errno::ENOENT))?;
-
-    if let Some(nsems) = nsems {
-        debug_assert!(nsems <= SEMMSL);
-        if nsems > sem_set.nsems() {
-            return_errno!(Errno::EINVAL);
+        if let Some(ipc_ns) = self.ipc_ns.upgrade() {
+            ipc_ns.free_sem_key(self.permission.key());
         }
     }
```
tatetian, line 303: **`Weak::upgrade` returns `None` during namespace destruction -- document why this is safe**: When the `IpcNamespace` is dropped (last process in the namespace exits), the `BTreeMap` drop cascades to each `SemaphoreSet::Drop`. At that point `Weak::upgrade()` returns `None` because the `Arc` strong count is zero, so `free_sem_key` is never called. This is correct because the `id_allocator` is also being destroyed, but the silent `None` path is a code smell. A comment explaining why the `None` case is acceptable would prevent future maintainers from treating it as a bug. Alternatively, a `debug_assert!` that fires if the `Weak` fails outside of namespace destruction would catch regressions if `SemaphoreSet` is ever stored outside its namespace's map.

kernel/src/ipc/ipc_ns.rs
```diff
@@ -0,0 +1,205 @@
...
+            Err(e) => {
+                id_allocator.free(key as usize);
+                return Err(e);
+            }
+        };
+
+        sem_sets_mut.insert(key, sem_set);
+
+        Ok(())
+    }
```
tatetian: [**`small-functions`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/functions-and-methods.html#small-functions): These two methods share substantial logic: both acquire `sem_sets_mut` and `id_allocator`, create a `SemaphoreSet`, handle rollback on error, and insert into the map. The only difference is how the key is obtained (auto-allocated vs. caller-specified). Consider extracting the shared tail into a private helper to reduce duplication.

[**`lock-ordering`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/concurrency-and-races.html#lock-ordering): The `sem_sets -> id_allocator` lock ordering is repeated as an inline comment and manually followed in both methods. Encapsulate it in a single private helper that returns both guards:

```rust
fn lock_sem_sets_and_id_allocator(
    &self,
) -> (
    RwLockWriteGuard<'_, BTreeMap<key_t, SemaphoreSet>, PreemptDisabled>,
    SpinLockGuard<'_, IdAlloc, PreemptDisabled>,
) {
    let sem_sets = self.semaphore_sets.write();
    let id_allocator = self.id_allocator.lock();
    (sem_sets, id_allocator)
}
```

This makes the ordering enforced structurally rather than by convention, and eliminates the duplicated `// Lock order:` comments.

**Unnecessary `self: &Arc<Self>` receiver**: Both methods use the `&Arc<Self>` receiver only to pass `self` to `SemaphoreSet::new`, which calls `Arc::downgrade(ipc_ns)` to store a `Weak`. The methods themselves never clone the `Arc`. `SemaphoreSet::new` could accept `Weak<IpcNamespace>` (or `&Weak<IpcNamespace>`) directly, letting these methods use a plain `&self` receiver. This removes the constraint that callers must hold an `Arc` and makes the API lighter.

Per [**`accurate-names`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#accurate-names), `create_sem_set` auto-allocates a key while `create_sem_set_with_key` uses a caller-supplied one. A more precise pair might be `create_sem_set` / `create_sem_set_at` (or `_auto_key` / `_fixed_key`) to make the distinction immediately obvious from the name.

Per [**`descriptive-names`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#descriptive-names), the parameter `nsems` is a C-ism abbreviation. Rust style prefers full words -- rename to `num_sems`. The same applies throughout the file (`check_sem`'s `nsems` at line 157, `SemaphoreSet::new`'s `nsems`, etc.).

kernel/src/ipc/ipc_ns.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct IpcNamespace {
+    /// Semaphore sets within this namespace.
+    semaphore_sets: RwLock<BTreeMap<key_t, SemaphoreSet>>,
+    /// ID allocator for semaphore sets.
+    id_allocator: SpinLock<IdAlloc>,
+    /// Owner user namespace.
+    owner: Arc<UserNamespace>,
+    /// Stashed dentry for nsfs.
+    stashed_dentry: StashedDentry,
+}
```
tatetian, line 48: [**`lock-ordering`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/concurrency-and-races.html#lock-ordering): The lock order `semaphore_sets -> id_allocator` is documented inline in `create_sem_set` (line 97) and `create_sem_set_with_key` (line 132), but not at the struct level. The `IPC_RMID` + `SemaphoreSet::Drop` path also involves `SemSetInner::inner` SpinLock, making the full ordering `semaphore_sets -> inner -> id_allocator`. Documenting this on the struct definition makes it discoverable without reading every method. Also, `IpcNamespace` is `pub` but only used within the crate -- consider `pub(crate)`.

kernel/src/ipc/ipc_ns.rs
tatetian, line 1: [**`module-docs`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#module-docs): The new file `ipc_ns.rs` has no module-level `//!` documentation. For a major component like IPC namespace, a brief module doc explaining the purpose, scope (currently semaphores only), and design (per-namespace map + allocator) would be valuable.

kernel/src/ipc/ipc_ns.rs
```diff
@@ -0,0 +1,205 @@
...
+    process::{
+        Credentials, UserNamespace, credentials::capabilities::CapSet, posix_thread::PosixThread,
+    },
+};
+
+/// The IPC namespace.
+///
+/// An IPC namespace isolates System V IPC objects (semaphores, message queues,
+/// shared memory). Each namespace maintains its own independent set of IPC
+/// resources and identifier allocator.
```
tatetian: [**`semantic-line-breaks`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#semantic-line-breaks): The doc comment wraps at a fixed width rather than breaking at clause boundaries. This applies throughout the file (e.g., `new_clone` at line 66, `check_sem` at line 154, `NsCommonOps` at line 195). Suggested rewrite for the struct doc:

```rust
/// The IPC namespace.
///
/// An IPC namespace isolates System V IPC objects
/// (semaphores, message queues, shared memory).
/// Each namespace maintains its own independent set
/// of IPC resources and identifier allocator.
```

tatetian (commented):
Great job. This PR looks mostly good, although there are some minor issues to be addressed.

kernel/src/ipc/ipc_ns.rs
StevenJiang1110: > Unnecessary self: &Arc<Self> receiver

This is necessary. Otherwise how can a `Weak<Self>` pointer be created?

> A more precise pair might be create_sem_set / create_sem_set_at (or _auto_key / _fixed_key)

The problem should disappear since they now become internal APIs.

kernel/src/ipc/ipc_ns.rs
StevenJiang1110: >  making the full ordering semaphore_sets -> inner -> id_allocator

This is incorrect. the lock of inner and id_allocator won't be held together. So the lock order should be semaphore_sets -> inner and semaphore_sets -> id_allocator

> Also, IpcNamespace is pub but only used within the crate -- consider pub(crate).

kernel crate will not use `pub(crate)` since this crate has no users.

kernel/src/ipc/ipc_ns.rs
```diff
@@ -0,0 +1,322 @@
...
+        mode: u16,
+        credentials: Credentials<ReadOp>,
+    ) -> Result<key_t> {
+        const IPC_NEW: key_t = 0;
+
+        if key < 0 {
+            return_errno_with_message!(Errno::EINVAL, "semaphore key must not be negative");
+        }
+
+        if key == IPC_NEW || (key as usize > SEMMNI && flags.contains(IpcFlags::IPC_CREAT)) {
```
tatetian, line 135: **Keys exceeding `SEMMNI` silently get auto-key creation instead of an error**: When `key > SEMMNI` and `IPC_CREAT` is set, this path calls `create_sem_set` (auto-allocate), which returns a _different_ key than the one the caller requested. Since this implementation uses key == semid, the caller cannot use their original key to refer to the created set.

For `key > SEMMNI` _without_ `IPC_CREAT`, the code correctly falls through to `lookup_sem_set` which returns `ENOENT`. But with `IPC_CREAT`, the caller expects `semget(K, n, IPC_CREAT)` to associate the set with key K, not a random different key. Consider returning `EINVAL` for keys that exceed `SEMMNI` regardless of flags:

```rust
if key as usize > SEMMNI {
    return_errno_with_message!(Errno::EINVAL, "semaphore key exceeds SEMMNI");
}
```

tatetian (commented):
Thanks for address the previous issues

tatetian:
After seeing all minor issues get fixed, I realized a bigger problem, which requires a structural change to this PR.

## Problem

`IpcNamespace` currently violates the single-responsibility principle. It acts as both the namespace container (lifecycle, nsfs integration, owner tracking) and the semaphore subsystem manager (create, lookup, remove, ID allocation, permission checking). When message queues and shared memory are added, this struct would balloon with three copies of similar-but-different logic.

The root cause: there is no intermediate abstraction between the namespace and the IPC objects.

## Linux's design (for reference)

Linux solves this with `struct ipc_ids` -- a generic, mechanism-agnostic registry that manages IPC objects. `struct ipc_namespace` holds three `ipc_ids` instances (one per mechanism: sem, msg, shm). Each mechanism (`sem.c`, `shm.c`, `msg.c`) operates on its assigned `ipc_ids` slot via shared helpers in `ipc/util.c`. The namespace itself contains no mechanism-specific logic -- only the three registries and per-mechanism tunables.

## Proposed design

Introduce `IpcIds<T>` -- a generic IPC object registry, parameterized by the object type. `IpcNamespace` becomes a thin container holding one `IpcIds` per mechanism.

### New types

```
ipc/
  mod.rs            -- IpcNamespace, IpcFlags, IpcControlCmd, IpcPermission, key_t
  ipc_ids.rs        -- IpcIds<T> (generic registry)
  semaphore/
    system_v/
      sem_set.rs    -- SemaphoreSet (unchanged)
      sem.rs        -- Semaphore, sem_op (unchanged)
```

### `IpcIds<T>`: the generic registry

```rust
/// A per-mechanism registry of IPC objects within a namespace.
///
/// Manages the mapping from keys to objects and the ID allocator.
/// All operations are internally synchronized.
pub(crate) struct IpcIds<T> {
    // Lock ordering: `objects` -> `id_allocator`.
    objects: RwLock<BTreeMap<key_t, T>>,
    id_allocator: SpinLock<IdAlloc>,
}
```

Methods on `IpcIds<T>`:

```rust
impl<T> IpcIds<T> {
    /// Creates an empty registry with the given ID capacity.
    pub fn new(capacity: usize) -> Self;

    /// Looks up an object by key under a read lock, then calls `op` on it.
    pub fn with<R>(&self, key: key_t, op: impl FnOnce(&T) -> Result<R>) -> Result<R>;

    /// Looks up an object by key under a write lock, verifies `may_remove`, then removes it.
    pub fn remove(&self, key: key_t, may_remove: impl FnOnce(&T) -> Result<()>) -> Result<()>;

    /// Inserts a new object with an auto-allocated key.
    /// `make` receives the allocated key and returns the object.
    pub fn insert_auto(&self, make: impl FnOnce(key_t) -> Result<T>) -> Result<key_t>;

    /// Inserts a new object at a specific key.
    pub fn insert_at(&self, key: key_t, make: impl FnOnce(key_t) -> Result<T>) -> Result<()>;

    /// Frees a key back to the allocator. Called from `T::drop` via `Weak<IpcNamespace>`.
    pub fn free_key(&self, key: key_t);
}
```

The `with` method eliminates TOCTOU races structurally: the lock is held for the entire check-then-act sequence, and callers never see raw lock guards.

### Slimmed-down `IpcNamespace`

```rust
pub struct IpcNamespace {
    sem_ids: IpcIds<SemaphoreSet>,
    // Future:
    // msg_ids: IpcIds<MessageQueue>,
    // shm_ids: IpcIds<SharedMemorySegment>,
    owner: Arc<UserNamespace>,
}
```

`IpcNamespace` provides thin accessors and mechanism-specific policy:

```rust
impl IpcNamespace {
    pub fn get_init_singleton() -> &'static Arc<Self>;
    pub fn new_clone(&self, owner: Arc<UserNamespace>, pt: &PosixThread) -> Result<Arc<Self>>;

    // -- Semaphore-specific entry points --

    /// Delegates to `self.sem_ids.with(...)` after validating semid.
    pub fn with_sem_set<R, F: FnOnce(&SemaphoreSet) -> Result<R>>(
        &self, semid: key_t, num_sems: Option<usize>,
        perm: PermissionMode, op: F,
    ) -> Result<R>;

    /// Delegates to `self.sem_ids.remove(...)`.
    pub fn remove_sem_set(
        &self, semid: key_t,
        may_remove: impl FnOnce(&SemaphoreSet) -> Result<()>,
    ) -> Result<()>;

    /// Combines lookup-or-create logic for semget.
    pub fn get_or_create_sem_set(self: &Arc<Self>, ...) -> Result<key_t>;

    /// Frees a semaphore key (called from SemaphoreSet::Drop).
    pub(super) fn free_sem_key(&self, key: key_t) {
        self.sem_ids.free_key(key);
    }
}
```

The thin methods on `IpcNamespace` handle semaphore-specific validation (semid > 0, num_sems check, permission mode) and then delegate to `IpcIds`. The `get_or_create_sem_set` method stays on `IpcNamespace` because it needs to coordinate between lookup and insert (which is a policy decision, not a registry operation).

### What moves where

| Current location | Destination | What |
|---|---|---|
| `IpcNamespace::semaphore_sets` | `IpcIds<SemaphoreSet>::objects` | The `BTreeMap` |
| `IpcNamespace::id_allocator` | `IpcIds<SemaphoreSet>::id_allocator` | The `IdAlloc` |
| `IpcNamespace::lookup_sem_set` | `IpcIds::with` (generic) + `IpcNamespace::with_sem_set` (validation) | Split |
| `IpcNamespace::insert_sem_set` | `IpcIds::insert_at` / `insert_auto` | Generic insert |
| `IpcNamespace::create_sem_set` | `IpcNamespace::create_sem_set` (calls `sem_ids.insert_auto`) | Stays, thinner |
| `IpcNamespace::create_sem_set_with_key_locked` | `IpcNamespace::get_or_create_sem_set` (calls `sem_ids.insert_at`) | Inlined |
| `IpcNamespace::free_sem_key` | Delegates to `self.sem_ids.free_key` | One-liner |

### Adding message queues later

When message queues are added:

```rust
pub struct IpcNamespace {
    sem_ids: IpcIds<SemaphoreSet>,
    msg_ids: IpcIds<MessageQueue>,   // new
    owner: Arc<UserNamespace>,
}
```

`IpcNamespace` gets a parallel set of thin methods (`with_msg_queue`, `remove_msg_queue`, `get_or_create_msg_queue`) that validate message-queue-specific invariants and delegate to `msg_ids`. The `IpcIds<T>` code is fully reused.

### `SemaphoreSet`'s back-reference

`SemaphoreSet` currently stores `Weak<IpcNamespace>` to call `free_sem_key` from `Drop`. With this design, it would ideally store a `Weak` to `IpcIds<SemaphoreSet>` instead, but `IpcIds` is not `Arc`-wrapped. Two options:

1. **Keep `Weak<IpcNamespace>`** and route through `IpcNamespace::free_sem_key` -> `sem_ids.free_key`. Simple, one extra indirection. This is the recommended approach.

2. **Remove the back-reference** and have `IpcIds::remove` handle the `free_key` call. `Drop` only wakes pending ops. This is cleaner but requires ensuring removal always goes through `IpcIds`.

Option 1 requires no change to `SemaphoreSet`.

## Summary

| Concern | Before | After |
|---|---|---|
| Registry logic | Mixed into `IpcNamespace` | `IpcIds<T>` (reusable) |
| Lock management | Callers see `RwLockGuard` | Encapsulated in `IpcIds` |
| TOCTOU prevention | Achieved by `with_sem_set_by_id` | Structural via `IpcIds::with` |
| Adding new mechanisms | Copy-paste all fields + methods | Add one `IpcIds<NewType>` field + thin wrappers |
| `IpcNamespace` size | ~300 LOC, grows with each mechanism | ~100 LOC, stable |

lrh2000:
https://github.com/asterinas/asterinas/pull/3109 sorts the namespace list in alphabetic order. So there are some more conflicts to solve here.

In addition, for consistency with other namespaces:
```diff
-            "IPC namespaces do not have a parent namespace"
+            "an IPC namespace does not have a parent namespace"
```
```diff
         static INIT: Once<Arc<IpcNamespace>> = Once::new();
+
         INIT.call_once(|| {
             let owner = UserNamespace::get_init_singleton().clone();
             Self::new(owner)
         })
```

tatetian:
> `IpcNamespace` currently violates the single-responsibility principle. It acts as both the namespace container (lifecycle, nsfs integration, owner tracking) and the semaphore subsystem manager (create, lookup, remove, ID allocation, permission checking). When message queues and shared memory are added, this struct would balloon with three copies of similar-but-different logic.

@StevenJiang1110 The revised version did not follow the suggestion. Any concerns or difficulties?

StevenJiang1110:
> @StevenJiang1110 The revised version did not follow the suggestion. Any concerns or difficulties?

Sorry, this happened because I have two working directories on two different machines. While working on https://github.com/asterinas/asterinas/pull/2988#issuecomment-4234553030, I made changes on a branch that hadn't been updated and pushed them, which caused the branch in this PR to fall behind.

It should be up to date now.

tatetian (approved):

tatetian:
@StevenJiang1110 Need to fix the conflict

StevenJiang1110:
> @StevenJiang1110 Need to fix the conflict

The conflict has been resolved.

PR #2996, opened 2026-03-02
fs/path: fix symlink resolution with trailing slash

zzj-5341:
When a path ends with '/' (target_is_dir=true), a symlink in the final component must still be followed, because the trailing slash implies the target should be a directory.

Previously, the resolver fell into the else branch and returned ENOTDIR without following the symlink. Linux follows the symlink, and if it points to a non-directory, returns EISDIR.

Fix: add `|| target_is_dir` to the symlink-following condition.

### DIFF
--- a/test/initramfs/src/apps/fs/run_test.sh
+++ b/test/initramfs/src/apps/fs/run_test.sh
@@ -118,3 +118,5 @@ echo "All mount bind file test passed."
 ./pseudofs/pseudo_dev_id
 ./pseudofs/pseudo_inode
 ./pseudofs/pseudo_mount
+
+./symlink/symlink
--- a/test/initramfs/src/apps/fs/symlink/Makefile
+++ b/test/initramfs/src/apps/fs/symlink/Makefile
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: MPL-2.0
+
+include ../../common/Makefile
--- a/test/initramfs/src/apps/fs/symlink/symlink.c
+++ b/test/initramfs/src/apps/fs/symlink/symlink.c
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "../../common/test.h"
+
+#define TEST_DIR "/ext2/symlink_test"
+#define DIRECTORY_TARGET TEST_DIR "/dir"
+#define REGULAR_TARGET TEST_DIR "/file"
+#define DIRECTORY_SYMLINK TEST_DIR "/dir_link"
+#define REGULAR_SYMLINK TEST_DIR "/file_link"
+
+FN_SETUP(prepare)
+{
+	int fd = -1;
+
+	CHECK(mkdir(TEST_DIR, 0755));
+	CHECK(mkdir(DIRECTORY_TARGET, 0755));
+
+	fd = CHECK(open(REGULAR_TARGET, O_CREAT | O_RDWR | O_TRUNC, 0644));
+	CHECK(close(fd));
+
+	CHECK(symlink("dir", DIRECTORY_SYMLINK));
+	CHECK(symlink("file", REGULAR_SYMLINK));
+}
+END_SETUP()
+
+FN_TEST(symlink_trailing_slash_nofollow)
+{
+	// `dir_link` points to a directory, so the trailing slash forces directory
+	// semantics and the symlink must still be resolved with `O_NOFOLLOW`.
+	int fd = TEST_SUCC(open(DIRECTORY_SYMLINK "/", O_RDONLY | O_NOFOLLOW));
+	TEST_SUCC(close(fd));
+}
+END_TEST()
+
+FN_TEST(symlink_trailing_slash_nofollow_non_directory_returns_enotdir)
+{
+	// `file_link` points to a regular file, so the trailing slash still forces
+	// directory semantics but the post-resolution check must reject it.
+	TEST_ERRNO(open(REGULAR_SYMLINK "/", O_RDONLY | O_NOFOLLOW), ENOTDIR);
+}
+END_TEST()
+
+FN_TEST(open_creat_excl_nofollow_symlink)
+{
+	// `O_CREAT | O_EXCL | O_NOFOLLOW` on an existing symlink should return
+	// `EEXIST` before the symlink-specific `ELOOP`.
+	TEST_ERRNO(open(DIRECTORY_SYMLINK,
+			O_CREAT | O_EXCL | O_NOFOLLOW | O_RDWR, 0644),
+		   EEXIST);
+}
+END_TEST()
+
+FN_SETUP(cleanup)
+{
+	CHECK(unlink(REGULAR_SYMLINK));
+	CHECK(unlink(DIRECTORY_SYMLINK));
+	CHECK(unlink(REGULAR_TARGET));
+	CHECK(rmdir(DIRECTORY_TARGET));
+	CHECK(rmdir(TEST_DIR));
+}
+END_SETUP()
--- a/test/initramfs/src/apps/fs/Makefile
+++ b/test/initramfs/src/apps/fs/Makefile
@@ -9,5 +9,6 @@ SUBDIRS := \
 	overlayfs \
 	procfs \
 	pseudofs \
+	symlink \
 
 include ../common/Makefile


kernel/src/fs/path/resolver.rs
```diff
@@ -449,7 +449,7 @@ impl PathResolver {
 
             let next_type = next_path.type_();
             // If next inode is a symlink, follow symlinks at most `SYMLINKS_MAX` times.
-            if next_type == InodeType::SymLink && (follow_tail_link || !next_is_tail) {
+            if next_type == InodeType::SymLink && (follow_tail_link || !next_is_tail || target_is_dir) {
```
cchanging: Thanks for the contribution! Could you add a related regression test to confirm this behavior?

In addtion, you can run `make format` and `make check` at root to aovid lint errors.

tatetian:
@boterinas codex

kernel/src/fs/path/resolver.rs
zzj-5341: I've added a regression test in `test/initramfs/src/apps/fs/symlink_trailing_slash/` that verifies the correct behavior. Also ran `cargo fmt` for Rust formatting

tatetian:
@boterinas codex

tatetian:
@boterinas codex

test/initramfs/src/apps/fs/symlink_trailing_slash/symlink_trailing_slash.c
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: MPL-2.0
```
cchanging: Insert a blank line here.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -449,7 +449,9 @@ impl PathResolver {
 
             let next_type = next_path.type_();
             // If next inode is a symlink, follow symlinks at most `SYMLINKS_MAX` times.
-            if next_type == InodeType::SymLink && (follow_tail_link || !next_is_tail) {
+            if next_type == InodeType::SymLink
+                && (follow_tail_link || !next_is_tail || target_is_dir)
```
cchanging: I am not sure this modification can solve the bug that tested in regression test.

```c
// Without O_CREAT: symlink -> regular_file, trailing slash -> ENOTDIR
TEST_ERRNO(open(SYMLINK "/", O_RDONLY), ENOTDIR);

// With O_CREAT: target exists and is a directory (via symlink) -> EISDIR
TEST_ERRNO(open(SYMLINK "/", O_RDONLY | O_CREAT, 0644), EISDIR);
```

I feel that the two open operations above will still follow the same path here, so why would one be able to return a different error code?

cchanging (commented):
Thanks for adding a regression test! There are still a few issues that need to be pointed out in this PR:

> ran cargo fmt for Rust formatting

`cargo fmt` is not enough, you need to run `make format` to format all kinds of files including the c files. That is why the lint CI will fail in this PR.

the folder name "symlink_trailing_slash" is too specific. You could simply name it "symlink" so that other related tests can be added here in the future.

You also need to add the running command in `test/initramfs/.../fs/run_test.sh` so that the added regression test will be tested in CI.

kernel/src/fs/vfs/path/resolver.rs
zzj-5341: You're right to question this. After re-analysis:

The fix addresses this specific case: when a path has a trailing slash and 
O_NOFOLLOW is set (e.g. `open("lnk/", O_NOFOLLOW)` where `lnk -> dir`):

- Before fix: `follow_tail_link=false` (due to O_NOFOLLOW), so the resolver 
  skipped symlink resolution and returned ENOTDIR incorrectly.
- After fix: the trailing slash (`target_is_dir=true`) forces symlink resolution 
  regardless of O_NOFOLLOW, correctly reaching the target directory and returning 
  Success.

The semantics: a trailing slash explicitly requests directory resolution, which 
takes precedence over O_NOFOLLOW. This matches Linux/POSIX behavior.

I've updated the regression test to correctly reflect this fix.

test/initramfs/src/apps/fs/symlink/Makefile
```diff
@@ -0,0 +1,2 @@
+# SPDX-License-Identifier: MPL-2.0
```
cchanging, line 1: Insert a blank line here.

test/initramfs/src/apps/fs/run_test.sh
```diff
@@ -112,3 +112,4 @@ echo "All mount bind file test passed."
 ./pseudofs/pseudo_dentry
 ./pseudofs/pseudo_inode
 ./pseudofs/pseudo_mount
```
cchanging, line 120: Insert a blank line here to be consistent with other test suites.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -106,18 +106,18 @@ impl Path {
...
+        {
+            return_errno_with_message!(Errno::EEXIST, "the file already exists");
+        }
+
         if inode_type == InodeType::SymLink
             && creation_flags.contains(CreationFlags::O_NOFOLLOW)
             && !status_flags.contains(StatusFlags::O_PATH)
         {
             return_errno_with_message!(Errno::ELOOP, "the file is a symlink");
         }
```
cchanging, line 124: Insert a blank line here so that each `if` block is separated by a blank line.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -106,18 +106,18 @@ impl Path {
         let creation_flags = &open_args.creation_flags;
         let status_flags = &open_args.status_flags;
 
+        if creation_flags.contains(CreationFlags::O_CREAT)
```
cchanging, line 113: Moving this check forward seems to be for solving the issue where another error code is inconsistent with Linux? Can we also add a test for it?

zzj-5341:
Thanks for the review! Addressed all comments in the latest commit:

1. Added the requested blank line in `check_open_util`.
2. Added regression test `open_creat_excl_nofollow.c` to verify that
   `open(..., O_CREAT | O_EXCL | O_NOFOLLOW, ...)` on an existing symlink
   returns `EEXIST`.
3. Hooked the new test into `test/src/apps/scripts/fs.sh`.

Could you please take another look?

tatetian:
@zzj-5341 To make your future contribution more smoothly, I recommend reading the Asterinas [Coding Guidelines](https://asterinas.github.io/book/to-contribute/coding-guidelines/index.html).

test/initramfs/src/apps/fs/run_test.sh
```diff
@@ -112,3 +112,6 @@ echo "All mount bind file test passed."
...
+
+
```
cchanging: Here is an extra blank line.

tatetian:
@zzj-5341 Could you fix all the conflicts?

test/initramfs/src/apps/fs/symlink/Makefile
cchanging: There is still no blank line here.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -567,7 +567,9 @@ impl PathResolver {
...
```
tatetian: **Suggestion**: Consider adding a brief inline comment explaining the `target_is_dir` disjunct, e.g., `// A trailing '/' forces symlink resolution even when follow_tail_link is false`. The other two conditions have fairly obvious semantics, but the interplay between `target_is_dir` and `follow_tail_link` (O_NOFOLLOW) is the exact subtlety this PR fixes and could save a future reader from re-deriving the reasoning.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -110,18 +110,18 @@ impl Path {
...
+        {
+            return_errno_with_message!(Errno::EEXIST, "the file already exists");
+        }
+
         if inode_type == InodeType::SymLink
             && creation_flags.contains(CreationFlags::O_NOFOLLOW)
             && !status_flags.contains(StatusFlags::O_PATH)
         {
             return_errno_with_message!(Errno::ELOOP, "the file is a symlink");
         }
```
tatetian, line 124: **Commit message coverage**: The PR description only mentions the trailing-slash fix but this reordering is a separate behavioral change (EEXIST vs ELOOP for `O_CREAT|O_EXCL|O_NOFOLLOW` on symlinks). Consider putting this fix into another commit and mentioning both fixes in the PR description so reviewers and future `git log` readers understand both changes were intentional.

test/initramfs/src/apps/fs/symlink/symlink.c
```diff
@@ -0,0 +1,49 @@
...
+END_SETUP()
+
+FN_TEST(symlink_trailing_slash_nofollow)
+{
+	// lnk -> dir, O_NOFOLLOW + trailing slash should succeed
+	// (trailing slash implies directory, symlink must be followed)
+	int fd = TEST_SUCC(open(SYMLINK "/", O_RDONLY | O_NOFOLLOW));
+	TEST_SUCC(close(fd));
+}
+END_TEST()
```
tatetian, line 40: **Suggestion -- negative case**: Consider adding a test for the case where the symlink target is *not* a directory (e.g., `lnk -> regular_file`). Opening `lnk/` with `O_NOFOLLOW` should return `ENOTDIR` in that case. This would exercise the post-resolution validation at `resolver.rs:613` and guard against regressions where the symlink is followed but the directory check is accidentally dropped.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -567,7 +567,11 @@ impl PathResolver {
 
             let next_type = next_path.type_();
             // If next inode is a symlink, follow symlinks at most `SYMLINKS_MAX` times.
-            if next_type == InodeType::SymLink && (follow_tail_link || !next_is_tail) {
+            // A trailing `/` forces directory semantics, so the tail symlink must still be
+            // resolved even when `follow_tail_link` is false.
+            if next_type == InodeType::SymLink
+                && (follow_tail_link || !next_is_tail || target_is_dir)
```
tatetian: Attach new the comment to `target_is_dir` because that is what the comment is about:

```rust
            if next_type == InodeType::SymLink
                && (follow_tail_link || !next_is_tail
                    // A trailing `/` forces directory semantics,
                    // so the tail symlink must still be resolved
                    // even when `follow_tail_link` is false.
                    || target_is_dir)
```

PR #2998, opened 2026-03-02
Refactor ANSI escape sequence processing

wyt8:
# Refactor ANSI escape sequence processing

Improved the handling of ANSI escape sequences, providing more comprehensive support for CSI sequences, while also silently handling OSC sequences.

With this PR, the screen output is now cleaner, and running the `clear` command in the command line to clear the screen is fully supported.

## Previous Behavior

- Run with `make run_nixos`, and then open the vnc.

  image

- Run with `make BOOT_PROTOCOL=linux-efi-handover64 run_kernel`, and then execute the command `setsid sh -c 'exec sh -i < /dev/tty1 > /dev/tty1 2>&1'`. Open the vnc viewer and execute the `clear` command.

  image

---

**References**:

- [CSI](https://en.wikipedia.org/wiki/ANSI_escape_code#Control_Sequence_Introducer_commands)

- [OSC](https://en.wikipedia.org/wiki/ANSI_escape_code#Operating_System_Command_sequences)

### DIFF
--- a/kernel/comps/framebuffer/src/ansi_escape.rs
+++ b/kernel/comps/framebuffer/src/ansi_escape.rs
@@ -6,7 +6,18 @@ use crate::Pixel;
 #[derive(Debug)]
 pub(super) struct EscapeFsm {
     state: WaitFor,
-    params: [u32; MAX_PARAMS],
+    params: [Option<u32>; MAX_PARAMS],
+}
+
+/// The mode for "Erase in Display" (ED) commands.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v6.13/source/drivers/tty/vt/vt.c#L1488>.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) enum EraseInDisplay {
+    CursorToEnd,
+    CursorToBeginning,
+    EntireScreen,
+    EntireScreenAndScrollback,
 }
 
 /// A trait to execute operations from ANSI escape sequences.
@@ -18,15 +29,37 @@ pub(super) trait EscapeOp {
     fn set_fg_color(&mut self, val: Pixel);
     /// Sets the background color.
     fn set_bg_color(&mut self, val: Pixel);
+
+    /// Erases part or all of the display.
+    fn erase_in_display(&mut self, mode: EraseInDisplay);
 }
 
 const MAX_PARAMS: usize = 8;
+const MAX_OSC_LEN: u8 = 128;
 
+// FIXME: Currently we only support a few ANSI escape sequences, and we just swallow the
+// unsupported ones.
 #[derive(Clone, Copy, Debug)]
 enum WaitFor {
+    /// Waits for an ESC to start the ANSI escape sequence.
     Escape,
+    /// Waits for a bracket after the ESC.
+    ///
+    /// "ESC[" will start a Control Sequence Introducer (CSI) sequence and "ESC]" will start a
+    /// Operating System Command (OSC) sequence.
     Bracket,
-    Params(u8),
+    /// Waits for CSI parameters.
+    ///
+    /// This will be terminated by a byte in the range 0x40 through 0x7E.
+    Csi {
+        idx: u8,
+        is_private: bool,
+        in_intermediate: bool,
+    },
+    /// Waits for OSC payload.
+    ///
+    /// This will be terminated by a bell (BEL, 0x07) or a String Terminator (ST, "ESC\").
+    Osc { len: u8, is_last_esc: bool },
 }
 
 /// Foreground and background colors.
@@ -69,92 +102,250 @@ const COLORS: [Pixel; 16] = [
 ];
 
 impl EscapeFsm {
+    /// The ESC character.
+    const ESC: u8 = 0x1b;
+    /// The BEL character.
+    const BEL: u8 = 0x07;
+
     pub(super) fn new() -> Self {
         Self {
             state: WaitFor::Escape,
-            params: [0; MAX_PARAMS],
+            params: [None; MAX_PARAMS],
         }
     }
 
-    /// Tries to eat a character as part of the ANSI escape sequence.
+    /// Tries to eat a byte as part of the ANSI escape sequence.
     ///
-    /// This method returns a boolean value indicating whether the character is part of an ANSI
-    /// escape sequence. In other words, if the method returns true, then the character has been
-    /// eaten and should not be displayed in the console.
+    /// Returns `true` if the byte is consumed by the FSM and should not be rendered as text.
+    /// Returns `false` if the byte is not part of an escape sequence and should be rendered.
     pub(super) fn eat<T: EscapeOp>(&mut self, byte: u8, op: &mut T) -> bool {
-        let num_params = match (self.state, byte) {
-            // Handle '\033'.
-            (WaitFor::Escape, 0o33) => {
-                self.state = WaitFor::Bracket;
-                return true;
-            }
-            (WaitFor::Escape, _) => {
-                // This is not an ANSI escape sequence.
-                return false;
+        match self.state {
+            WaitFor::Escape => {
+                if byte == Self::ESC {
+                    self.state = WaitFor::Bracket;
+                    return true;
+                }
+                false
             }
 
-            // Handle '['.
-            (WaitFor::Bracket, b'[') => {
-                self.state = WaitFor::Params(0);
-                self.params[0] = 0;
-                return true;
-            }
-            (WaitFor::Bracket, _) => {
-                // The character is invalid. We cannot handle it, so we are aborting the ANSI
-                // escape sequence.
-                self.state = WaitFor::Escape;
-                return true;
+            WaitFor::Bracket => {
+                match byte {
+                    // CSI begins.
+                    b'[' => {
+                        self.params.fill(None);
+                        self.state = WaitFor::Csi {
+                            idx: 0,
+                            is_private: false,
+                            in_intermediate: false,
+                        };
+                    }
+
+                    // OSC begins.
+                    b']' => {
+                        self.state = WaitFor::Osc {
+                            len: 0,
+                            is_last_esc: false,
+                        }
+                    }
+
+                    // The character is invalid. We cannot handle it, so we are aborting the ANSI
+                    // escape sequence.
+                    _ => self.state = WaitFor::Escape,
+                }
+
+                true
             }
 
-            // Handle numeric parameters.
-            (WaitFor::Params(i), b'0'..=b'9') => {
-                let param = &mut self.params[i as usize];
-                *param = param.wrapping_mul(10).wrapping_add((byte - b'0') as u32);
-                return true;
-            }
-            (WaitFor::Params(i), b';') if (i as usize + 1) < MAX_PARAMS => {
-                self.state = WaitFor::Params(i + 1);
-                self.params[i as usize + 1] = 0;
-                return true;
-            }
-            (WaitFor::Params(_), b';') => {
-                // There are too many parameters. We cannot handle that many, so we are aborting
-                // the ANSI escape sequence.
-                self.state = WaitFor::Escape;
-                return true;
+            WaitFor::Csi {
+                idx,
+                is_private,
+                in_intermediate,
+            } => {
+                self.parse_csi(byte, idx, is_private, in_intermediate, op);
+                true
             }
 
-            // Break and handle the final action.
-            (WaitFor::Params(i), _) => {
-                self.state = WaitFor::Escape;
-                (i + 1) as usize
-            }
-        };
+            WaitFor::Osc { len, is_last_esc } => {
+                match byte {
+                    // Terminate the OSC sequence by a BEL or a ST.
+                    Self::BEL => self.state = WaitFor::Escape,
+                    b'\\' if is_last_esc => self.state = WaitFor::Escape,
 
+                    // Swallow the OSC payload.
+                    _ => {
+                        if len > MAX_OSC_LEN {
+                            self.state = WaitFor::Escape;
+                            return false;
+                        }
+                        self.state = WaitFor::Osc {
+                            len: len + 1,
+                            is_last_esc: byte == Self::ESC,
+                        }
+                    }
+                }
+
+                true
+            }
+        }
+    }
+
+    /// Parses CSI arguments.
+    fn parse_csi<T: EscapeOp>(
+        &mut self,
+        byte: u8,
+        idx: u8,
+        is_private: bool,
+        in_intermediate: bool,
+        op: &mut T,
+    ) {
         match byte {
-            // CUP - Cursor Position
-            b'H' if num_params == 2 => {
+            // Intermediate bytes (0x20..=0x2F).
+            // We transition to the intermediate section once we see any intermediate byte;
+            // later bytes must be either intermediate or final.
+            0x20..=0x2f => {
+                self.state = WaitFor::Csi {
+                    idx,
+                    is_private,
+                    in_intermediate: true,
+                };
+            }
+
+            // Parameter bytes (0x30..=0x3F).
+            0x30..=0x3f if !in_intermediate => {
+                match byte {
+                    // Digits contribute to numeric parameters.
+                    b'0'..=b'9' => {
+                        let i = idx as usize;
+                        if i < MAX_PARAMS {
+                            let p = &mut self.params[i];
+                            *p = Some(
+                                p.unwrap_or(0)
+                                    .saturating_mul(10)
+                                    .saturating_add((byte - b'0') as u32),
+                            );
+                        }
+                        self.state = WaitFor::Csi {
+                            idx,
+                            is_private,
+                            in_intermediate: false,
+                        };
+                    }
+
+                    // ';' separates numeric parameters.
+                    b';' => {
+                        let next = idx + 1;
+                        if (next as usize) < MAX_PARAMS {
+                            // If there are no digits for this parameter, it will remain `None`.
+                            self.state = WaitFor::Csi {
+                                idx: next,
+                                is_private,
+                                in_intermediate: false,
+                            };
+                        } else {
+                            // There are too many parameters. We cannot handle that many, so
+                            // we are aborting the ANSI escape sequence.
+                            self.state = WaitFor::Escape;
+                        }
+                    }
+
+                    // The behavior of ':' is not defined by the standard. We don't support it,
+                    // so we are aborting the ANSI escape sequence.
+                    b':' => self.state = WaitFor::Escape,
+
+                    // Sequences containing "<=>?" are private. We swallow them and mark
+                    // `is_private`.
+                    b'<' | b'=' | b'>' | b'?' => {
+                        self.state = WaitFor::Csi {
+                            idx,
+                            is_private: true,
+                            in_intermediate: false,
+                        };
+                    }
+
+                    _ => unreachable!(),
+                }
+            }
+
+            // Parameter bytes in the intermediate section are illegal by the formal grammar.
+            // We'll abort and swallow to avoid leaking garbage.
+            0x30..=0x3f if in_intermediate => {
+                self.state = WaitFor::Escape;
+            }
+
+            // Final byte (0x40..=0x7E): ends the CSI.
+            0x40..=0x7e => {
+                self.state = WaitFor::Escape;
+
+                let num_params = idx as usize + 1;
+                self.handle_csi(byte, num_params, is_private, op);
+            }
+
+            // Terminal behavior is undefined if a CSI contains bytes outside 0x20..=0x7E.
+            // We'll abort and swallow to avoid leaking garbage.
+            _ => {
+                self.state = WaitFor::Escape;
+            }
+        }
+    }
+
+    /// Handles the "Control Sequence Introducer" sequence.
+    fn handle_csi<T: EscapeOp>(
+        &self,
+        final_byte: u8,
+        num_params: usize,
+        is_private: bool,
+        op: &mut T,
+    ) {
+        if is_private {
+            // For now we don't handle any private sequences, so just swallow them.
+            return;
+        }
+
+        match final_byte {
+            // CUP - Cursor Position: CSI n ; m H
+            b'H' => {
+                // `n` and `m` are 1-based row and column numbers, respectively.
+                // They default to 1 if the parameter is missing.
+                let row_1b = self.param_or(0, 1);
+                let col_1b = self.param_or(1, 1);
+
                 op.set_cursor(
-                    self.params[1].saturating_sub(1) as usize,
-                    self.params[0].saturating_sub(1) as usize,
+                    col_1b.saturating_sub(1) as usize,
+                    row_1b.saturating_sub(1) as usize,
                 );
             }
 
-            // SGR - Select Graphic Rendition
-            b'm' => self.handle_srg(num_params, op),
+            // ED - Erase in Display: CSI n J
+            b'J' => {
+                // The default mode is `CursorToEnd` if the parameter is missing.
+                let n = self.param_or(0, 0);
+                let mode = match n {
+                    0 => EraseInDisplay::CursorToEnd,
+                    1 => EraseInDisplay::CursorToBeginning,
+                    2 => EraseInDisplay::EntireScreen,
+                    3 => EraseInDisplay::EntireScreenAndScrollback,
+                    _ => {
+                        // Invalid parameter.
+                        return;
+                    }
+                };
 
-            // Invalid or unsupported
+                op.erase_in_display(mode);
+            }
+
+            // SGR - Select Graphic Rendition
+            b'm' => self.handle_sgr(num_params, op),
+
+            // Unknown CSI: swallow silently.
             _ => {}
         }
-
-        true
     }
 
     /// Handles the "Select Graphic Rendition" sequence.
-    fn handle_srg<T: EscapeOp>(&self, num_params: usize, op: &mut T) {
+    fn handle_sgr<T: EscapeOp>(&self, num_params: usize, op: &mut T) {
         let mut cursor = 0;
         while cursor < num_params {
-            let op_code = self.params[cursor];
+            let op_code = self.param_or(cursor, 0) as u8;
             cursor += 1;
 
             match op_code {
@@ -167,15 +358,15 @@ impl EscapeFsm {
                 // Set foreground colors
                 // Reference: <https://en.wikipedia.org/wiki/ANSI_escape_code#Colors>
                 30..=37 => op.set_fg_color(COLORS[op_code as usize - 30]),
-                38 if num_params - cursor >= 2 && self.params[cursor] == 5 => {
-                    op.set_fg_color(Self::get_256_color(self.params[cursor + 1] as u8));
+                38 if num_params - cursor >= 2 && self.param_or(cursor, 0) == 5 => {
+                    op.set_fg_color(Self::get_256_color(self.param_or(cursor + 1, 0) as u8));
                     cursor += 2;
                 }
-                38 if num_params - cursor >= 4 && self.params[cursor] == 2 => {
+                38 if num_params - cursor >= 4 && self.param_or(cursor, 0) == 2 => {
                     op.set_fg_color(Pixel {
-                        red: self.params[cursor + 1] as u8,
-                        green: self.params[cursor + 2] as u8,
-                        blue: self.params[cursor + 3] as u8,
+                        red: self.param_or(cursor + 1, 0) as u8,
+                        green: self.param_or(cursor + 2, 0) as u8,
+                        blue: self.param_or(cursor + 3, 0) as u8,
                     });
                     cursor += 4;
                 }
@@ -186,15 +377,15 @@ impl EscapeFsm {
                 // Set background colors
                 // Reference: <https://en.wikipedia.org/wiki/ANSI_escape_code#Colors>
                 40..=47 => op.set_bg_color(COLORS[op_code as usize - 40]),
-                48 if num_params - cursor >= 2 && self.params[cursor] == 5 => {
-                    op.set_bg_color(Self::get_256_color(self.params[cursor + 1] as u8));
+                48 if num_params - cursor >= 2 && self.param_or(cursor, 0) == 5 => {
+                    op.set_bg_color(Self::get_256_color(self.param_or(cursor + 1, 0) as u8));
                     cursor += 2;
                 }
-                48 if num_params - cursor >= 4 && self.params[cursor] == 2 => {
+                48 if num_params - cursor >= 4 && self.param_or(cursor, 0) == 2 => {
                     op.set_bg_color(Pixel {
-                        red: self.params[cursor + 1] as u8,
-                        green: self.params[cursor + 2] as u8,
-                        blue: self.params[cursor + 3] as u8,
+                        red: self.param_or(cursor + 1, 0) as u8,
+                        green: self.param_or(cursor + 2, 0) as u8,
+                        blue: self.param_or(cursor + 3, 0) as u8,
                     });
                     cursor += 4;
                 }
@@ -247,6 +438,11 @@ impl EscapeFsm {
 
         Pixel { red, green, blue }
     }
+
+    /// Gets the parameter at the given index, or returns the default value if it is absent.
+    fn param_or(&self, i: usize, default: u32) -> u32 {
+        self.params.get(i).and_then(|p| *p).unwrap_or(default)
+    }
 }
 
 #[cfg(ktest)]
@@ -260,6 +456,7 @@ mod test {
         y: usize,
         fg: Pixel,
         bg: Pixel,
+        last_ed: Option<EraseInDisplay>,
     }
 
     impl Default for State {
@@ -269,6 +466,7 @@ mod test {
                 y: 0,
                 fg: Pixel::WHITE,
                 bg: Pixel::BLACK,
+                last_ed: None,
             }
         }
     }
@@ -285,6 +483,10 @@ mod test {
 
         fn set_bg_color(&mut self, val: Pixel) {
             self.bg = val;
+        }
+
+        fn erase_in_display(&mut self, mode: EraseInDisplay) {
+            self.last_ed = Some(mode);
         }
     }
 
@@ -311,6 +513,49 @@ mod test {
         eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[0;0H");
         assert_eq!(state.x, 0);
         assert_eq!(state.y, 0);
+
+        assert!(!esc_fsm.eat(b'a', &mut state));
+
+        // If parameters are missing, the cursor should move to the first row and column.
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[H");
+        assert_eq!(state.x, 0);
+        assert_eq!(state.y, 0);
+
+        assert!(!esc_fsm.eat(b'a', &mut state));
+    }
+
+    #[ktest]
+    fn erase_in_display() {
+        let mut esc_fsm = EscapeFsm::new();
+        let mut state = State::default();
+
+        // If parameters are missing, erase from the cursor to the end of the screen.
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[J");
+        assert_eq!(state.last_ed, Some(EraseInDisplay::CursorToEnd));
+        state.last_ed = None;
+
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[0J");
+        assert_eq!(state.last_ed, Some(EraseInDisplay::CursorToEnd));
+        state.last_ed = None;
+
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[1J");
+        assert_eq!(state.last_ed, Some(EraseInDisplay::CursorToBeginning));
+        state.last_ed = None;
+
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[2J");
+        assert_eq!(state.last_ed, Some(EraseInDisplay::EntireScreen));
+        state.last_ed = None;
+
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[3J");
+        assert_eq!(
+            state.last_ed,
+            Some(EraseInDisplay::EntireScreenAndScrollback)
+        );
+        state.last_ed = None;
+
+        // If the parameter is invalid, do nothing.
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[4J");
+        assert_eq!(state.last_ed, None);
 
         assert!(!esc_fsm.eat(b'a', &mut state));
     }
--- a/kernel/comps/framebuffer/src/console.rs
+++ b/kernel/comps/framebuffer/src/console.rs
@@ -15,7 +15,7 @@ use spin::Once;
 
 use crate::{
     FRAMEBUFFER, FrameBuffer, Pixel,
-    ansi_escape::{EscapeFsm, EscapeOp},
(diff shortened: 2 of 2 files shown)


tatetian:
@boterinas codex

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -18,15 +29,29 @@ pub(super) trait EscapeOp {
...
+    /// Just saw ESC (0x1B), expecting the next selector.
+    AfterEscape,
+    /// Currently parsing CSI parameters.
+    Csi {
+        idx: u8,
+        is_private: bool,
+        in_intermediate: bool,
+    },
+    /// OSC payload after ESC] . Swallow until BEL (0x07) or ST (ESC \).
+    Osc,
```
lrh2000: 1. Please try to be consistent when writing comments, e.g, the comments for `Osc` and `Csi`.
 2. Try explaining the shorthand.
 4. The type name is `WaitFor`, so `AfterEscape` may not make sense.
 5. `Osc` and `OscEscape` should be merged to handle the case `ESC ] <payload> ESC ESC \`.

```rust
enum WaitFor {
    /// Waits for an ESC to start the ANSI escape sequence.
    Escape,
    /// Waits for a bracket after the ESC.
    ///
    /// "ESC[" will start a Control Sequence Introducer (CSI) sequence and "ESC]" will start a
    /// Operating System Command (OSC) sequence.
    Bracket,
    /// Waits for CSI parameters.
    ///
    /// This will be terminated by a byte in the range 0x40 through 0x7E.
    Csi {
        idx: u8,
        is_private: bool,
        in_intermediate: bool,
    },
    /// Waits for OSC payload.
    ///
    /// This will be terminated by a bell (BEL, 0x07) or a String Terminator (ST, "ESC\").
    Osc {
        is_last_esc: bool,
    }
}
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
         }
     }
 
-    /// Tries to eat a character as part of the ANSI escape sequence.
+    /// Tries to eat a byte as part of the ANSI escape sequence.
     ///
-    /// This method returns a boolean value indicating whether the character is part of an ANSI
-    /// escape sequence. In other words, if the method returns true, then the character has been
-    /// eaten and should not be displayed in the console.
+    /// Returns `true` if the byte is consumed by the FSM and must not be rendered as text.
```
lrh2000:
```suggestion
    /// Returns `true` if the byte is consumed by the FSM and should not be rendered as text.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
-            (WaitFor::Escape, 0o33) => {
-                self.state = WaitFor::Bracket;
-                return true;
-            }
-            (WaitFor::Escape, _) => {
-                // This is not an ANSI escape sequence.
-                return false;
+        match self.state {
+            WaitFor::Escape => {
+                if byte == 0x1b {
```
lrh2000: Define constants:
```rust
/// The ESC character.
const ESC: u8 = 0x1b;
/// The BEL character.
const BEL: u8 = 0x07;
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                match byte {
+                    b'[' => {
+                        // CSI begins.
+                        self.params.fill(None);
+                        self.state = WaitFor::Csi {
+                            idx: 0,
+                            is_private: false,
+                            in_intermediate: false,
+                        };
+                        true
```
lrh2000: All match arms from here seem to return true. Try to avoid duplication by moving it out of the `match`.

The first match arm can use explicit `return true`/`return false`.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
-                return true;
-            }
-            (WaitFor::Params(i), b';') if (i as usize + 1) < MAX_PARAMS => {
-                self.state = WaitFor::Params(i + 1);
-                self.params[i as usize + 1] = 0;
-                return true;
+            WaitFor::Osc => {
+                match byte {
+                    0x07 => {
+                        // BEL terminator
```
lrh2000: Improve the comment by:
```rust
match byte {
    // Terminate the OSC sequence by a BEL or a ST.
    BEL => self.state = WaitFor::Escape,
    '\\' if is_last_esc => self.state = WaitFor::Escape,

    // Swallow the OSC payload.
    _ => self.state = WaitFor::Osc { is_last_esc: byte == ESC, },
}
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
-                (i + 1) as usize
+            WaitFor::Csi {
+                idx,
+                is_private,
+                in_intermediate,
+            } => {
+                match byte {
+                    // Intermediate bytes (0x20..=0x2F).
+                    // Once we see any intermediate, we are in the intermediate section;
+                    // later bytes must be intermediate or final.
```
lrh2000:
```suggestion
                    // We transition to the intermediate section once we see any intermediate byte;
                    // later bytes must be either intermediate or final.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                    0x20..=0x2f => {
+                        // If we already entered intermediate section, just keep swallowing them.
+                        // If we were still in parameter section, we now transition to intermediate.
```
lrh2000: These comments are duplicates of the previous ones, especially after I revised them. Consider removing them.
```suggestion
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                            is_private,
+                            in_intermediate: true,
+                        };
+                        true
+                    }
+
+                    // Parameter bytes (0x30..=0x3F).
+                    0x30..=0x3f if !in_intermediate => {
+                        match byte {
+                            // digits: contribute to numeric params
```
lrh2000:
```suggestion
                            // Digits contribute to numeric parameters.
```

Be consistent to the following version:
```
                            // ';' separates numeric parameters.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                                } else {
+                                    // There are too many parameters. We cannot handle that many, so we are aborting
+                                    // the ANSI escape sequence.
+                                    self.state = WaitFor::Escape;
+                                }
+                            }
+
+                            b':' => {
+                                // The behavior of ':' is not defined by the standard.
+                                log::warn!("EscapeFsm: unsupported ':' parameter separator in CSI");
```
lrh2000: 1. We cannot print logs in the console module. It may cause a deadlock.
 2. The `log::warn!` function should never be triggered from userspace. This will result in log spam.
 3. Choose a consistent way to deal with bad escape sequences. Why not abort the sequence as we do in other places?

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                            }
+
+                            // Sequences containing <=>? are "private". We swallow them and mark `is_private`.
```
lrh2000:
```suggestion
                            // Sequences containing "<=>?" are private. We swallow them and mark `is_private`.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                                    in_intermediate: false,
+                                };
+                            }
+
+                            _ => unreachable!(),
+                        }
+                        true
+                    }
+
+                    // Parameter bytes after intermediate section is illegal by the formal grammar.
```
lrh2000: Use either:
```suggestion
                    // Parameter bytes in the intermediate section is illegal by the formal grammar.
```
or
```suggestion
                    // Parameter bytes after intermediate bytes is illegal by the formal grammar.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
                 );
             }
 
+            // ED - Erase in Display: CSI n J
+            //
+            // n:
+            // - 0 (or missing): cursor to end of screen
+            // - 1: cursor to beginning of screen
+            // - 2: entire screen
+            // - 3: entire screen + scrollback
```
lrh2000: The following code is obvious. There's no need to write such comments, as I don't think they provide any value.

```suggestion
```

Perhaps the only value is "or missing", just use the following version:
```rust
                // The default mode is `CursorToEnd` if the parameter is missing.
                let n = self.param_or(0, 0);
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                    0 => EraseInDisplay::CursorToEnd,
+                    1 => EraseInDisplay::CursorToBeginning,
+                    2 => EraseInDisplay::EntireScreen,
+                    3 => EraseInDisplay::EntireScreenAndScrollback,
+                    _ => {
+                        // Invalid parameter.
+                        return;
+                    }
+                };
+                op.erase_in_display(mode);
```
lrh2000, line 333: Insert a blank line before `op.erase_in_display` or remove the blank line before `op.set_cursor`.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+            //
+            // - n=row, m=col
+            // - default to 1 if omitted
+            //
+            // Examples:
+            // - CSI H        -> 1;1
+            // - CSI ;5H      -> 1;5
+            // - CSI 17H      -> 17;1
+            // - CSI 17;H     -> 17;1
+            // - CSI 17;1H    -> 17;1
```
lrh2000: Similarly, they do not provide much value.
```suggestion
```
I prefer the following version:
```rust
                // `n` and `m` are 1-based row and column numbers, respectively.
                // They default to 1 if the parameter is missing.
                let row_1b = self.param_or(0, 1);
                let col_1b = self.param_or(1, 1);
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -306,15 +507,38 @@ mod test {
 
         assert!(!esc_fsm.eat(b'a', &mut state));
 
-        // There is invalid as there is no 0-th row or 0-th column. But in this case, let's move
-        // the cursor to the first row and the first column.
```
lrh2000, line 310: Why do you remove this test? I think it should be kept.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -306,15 +507,38 @@ mod test {
...
-        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[0;0H");
+        // CUP defaults to 1;1 when omitted.
```
lrh2000:
```suggestion
        // If parameters are missing, the cursor should move to the first row and column.
````

File: kernel/comps/framebuffer/src/ansi_escape.rs [hunk 17]
```diff
@@ -306,15 +507,38 @@ mod test {
...
     }
 
+    #[ktest]
+    fn erase_in_display() {
+        let mut esc_fsm = EscapeFsm::new();
+        let mut state = State::default();
+
+        // Default (or missing) is 0: cursor to end of screen.
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[J");
+        assert_eq!(state.last_ed, Some(EraseInDisplay::CursorToEnd));
```
lrh2000, line 534: Please test both the missing case and the `n=1` case:
```rust
        // If parameters are missing, erase from the cursor to the end of the screen.
        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[J");
        assert_eq!(state.last_ed, Some(EraseInDisplay::CursorToEnd));
        state.last_ed = None;

        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[0J");
        assert_eq!(state.last_ed, Some(EraseInDisplay::CursorToEnd));
        state.last_ed = None;
```

In addition, please add `state.last_ed = None` to the remaining tests for consistency.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -306,15 +507,38 @@ mod test {
...
+
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[2J");
+        assert_eq!(state.last_ed, Some(EraseInDisplay::EntireScreen));
+
+        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[3J");
+        assert_eq!(
+            state.last_ed,
+            Some(EraseInDisplay::EntireScreenAndScrollback)
+        );
+
```
lrh2000, line 559:
```rust
        // If the parameter is invalid, do nothing.
        eat_escape_sequence(&mut esc_fsm, &mut state, b"\x1B[4J");
        assert_eq!(state.last_ed, None);
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
             ConsoleMode::Graphics
         }
     }
+
+    /// Fill a rectangular pixel region [x0, x1) × [y0, y1) with the given color.
+    ///
+    /// This writes to the shadow buffer, and optionally flushes
+    /// the affected region to the real framebuffer.
```
lrh2000: This should be the behavior of all the methods in this file. There is no need to document this method separately.
```suggestion
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
...
+    fn fill_rect_pixels(&mut self, x0: usize, y0: usize, x1: usize, y1: usize, color: Pixel) {
+        if x0 >= x1 || y0 >= y1 {
+            return;
+        }
+
```
lrh2000: This check is duplicate, right? We will check it again after take the minimums.
```suggestion
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
...
+        let rendered_pixel = self.backend.render_pixel(color);
+        let rendered_pixel_size = rendered_pixel.nbytes();
+        let row_bytes = (x1 - x0) * rendered_pixel_size;
+
+        // Fill shadow buffer row by row.
+        for y in y0..y1 {
+            let off = self.backend.calc_offset(x0, y).as_usize();
+            let buf = &mut self.bytes[off..off + row_bytes];
+
+            // Write pixels.
```
lrh2000: Be consistent with the existing code:
```suggestion
            // Write pixels to the console buffer.
```
https://github.com/asterinas/asterinas/blob/f6257571c521e54a20a9f3147fb5c5c37c886af3/kernel/comps/framebuffer/src/console.rs#L238-L248

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
...
+            for chunk in buf.chunks_exact_mut(rendered_pixel_size) {
+                chunk.copy_from_slice(rendered_pixel.as_slice());
+            }
+
+            // Flush to framebuffer if needed.
```
lrh2000: Be consistent with the existing code:
```suggestion
            // Write pixels to the framebuffer.
```
https://github.com/asterinas/asterinas/blob/f6257571c521e54a20a9f3147fb5c5c37c886af3/kernel/comps/framebuffer/src/console.rs#L238-L248

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
...
+
+        if x0 >= x1 || y0 >= y1 {
+            return;
+        }
+
+        let rendered_pixel = self.backend.render_pixel(color);
+        let rendered_pixel_size = rendered_pixel.nbytes();
+        let row_bytes = (x1 - x0) * rendered_pixel_size;
+
+        // Fill shadow buffer row by row.
```
lrh2000: No need to have this comment.
```suggestion
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
             ConsoleMode::Graphics
         }
     }
+
+    /// Fill a rectangular pixel region [x0, x1) × [y0, y1) with the given color.
```
lrh2000:
```suggestion
    /// Fills a rectangular pixel region [x0, x1) × [y0, y1) with the given color.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
...
+            }
+
+            // Flush to framebuffer if needed.
+            if self.is_output_enabled {
+                self.backend.write_bytes_at(off, buf).unwrap();
+            }
+        }
+    }
+
+    /// Pixel coordinates for the cursor cell.
```
lrh2000:
```suggestion
    /// Calculates the pixel coordinates for the cursor cell.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,56 @@ impl ConsoleState {
...
+    fn cursor_cell_rect(&self) -> (usize, usize, usize, usize) {
+        let cx0 = self.x_pos.min(self.backend.width());
+        let cy0 = self.y_pos.min(self.backend.height());
+        let cx1 = (cx0 + self.font.width()).min(self.backend.width());
+        let cy1 = (cy0 + self.font.height()).min(self.backend.height());
```
lrh2000: All `min()`s are unnecessary. This type maintains the type invariant that the cursor cell is always within the screen.

Otherwise, `draw_char` may not work correctly:
https://github.com/asterinas/asterinas/blob/f6257571c521e54a20a9f3147fb5c5c37c886af3/kernel/comps/framebuffer/src/console.rs#L221

kernel/comps/framebuffer/src/console.rs
```diff
@@ -313,4 +363,36 @@ impl EscapeOp for ConsoleState {
...
+    fn erase_in_display(&mut self, mode: EraseInDisplay) {
+        let bg = self.bg_color;
+        let w = self.backend.width();
+        let h = self.backend.height();
+
+        let (cx0, cy0, cx1, cy1) = self.cursor_cell_rect();
+
+        match mode {
+            EraseInDisplay::CursorToEnd => {
+                // Clear from cursor x to end-of-line, within the cursor row.
```
lrh2000:
```suggestion
                // Clear from the cursor to the end of the line, within the cursor row.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -313,4 +363,36 @@ impl EscapeOp for ConsoleState {
...
+                self.fill_rect_pixels(cx0, cy0, w, cy1, bg);
+
+                // Clear all rows below cursor row.
```
lrh2000:
```suggestion
                // Clear all rows below the cursor row.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -313,4 +363,36 @@ impl EscapeOp for ConsoleState {
...
+                if cy1 < h {
+                    self.fill_rect_pixels(0, cy1, w, h, bg);
+                }
+            }
+            EraseInDisplay::CursorToBeginning => {
+                // Clear all rows above cursor row.
```
lrh2000:
```suggestion
                // Clear all rows above the cursor row.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -313,4 +363,36 @@ impl EscapeOp for ConsoleState {
...
+                if cy0 > 0 {
+                    self.fill_rect_pixels(0, 0, w, cy0, bg);
+                }
+
+                // Clear from start-of-line to cursor cell end, within cursor row.
```
lrh2000:
```suggestion
                // Clear from the start of the line to the cursor, within the cursor row.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -313,4 +363,36 @@ impl EscapeOp for ConsoleState {
...
+                self.fill_rect_pixels(0, cy0, cx1, cy1, bg);
+            }
+            EraseInDisplay::EntireScreen | EraseInDisplay::EntireScreenAndScrollback => {
+                self.fill_rect_pixels(0, 0, w, h, bg);
```
lrh2000, line 389:
```suggestion
                // Clear the entire screen.
                self.fill_rect_pixels(0, 0, w, h, bg);
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -72,89 +97,259 @@ impl EscapeFsm {
...
+                        match byte {
+                            // digits: contribute to numeric params
+                            b'0'..=b'9' => {
+                                let i = idx as usize;
+                                if i < MAX_PARAMS {
+                                    let p = &mut self.params[i];
+                                    *p = Some(
+                                        p.unwrap_or(0)
+                                            .saturating_mul(10)
+                                            .saturating_add((byte - b'0') as u32),
```
lrh2000: There are too many levels of indentation. Consider extracting some logics to a separate method.

lrh2000 (commented):
Thanks for your work.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+                }
+                true
             }
-            (WaitFor::Bracket, _) => {
-                // The character is invalid. We cannot handle it, so we are aborting the ANSI
-                // escape sequence.
-                self.state = WaitFor::Escape;
-                return true;
+
+            WaitFor::Csi {
```
lrh2000: Shouldn't `WaitFor::Csi` come before `WaitFor::Osc`? Be consistent in the order in which you define the enum variants when matching the enum, unless there is a special reason not to do so.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+    }
+
+    fn handle_csi<T: EscapeOp>(
+        &mut self,
+        byte: u8,
+        idx: u8,
+        is_private: bool,
+        in_intermediate: bool,
+        op: &mut T,
+    ) -> bool {
```
lrh2000: This method seems always return `true`. Or am I missing anything?

If it always return `true`, remove the return value from this method.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+                        if (next as usize) < MAX_PARAMS {
+                            // If there were no digits for this param, it already stays None.
+                            self.state = WaitFor::Csi {
+                                idx: next,
+                                is_private,
+                                in_intermediate: false,
+                            };
+                        } else {
+                            // There are too many parameters. We cannot handle that many, so we are aborting
+                            // the ANSI escape sequence.
```
lrh2000: The line is too long.
```suggestion
                            // There are too many parameters. We cannot handle that many, so we are
                            // aborting the ANSI escape sequence.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+                            is_private,
+                            in_intermediate: false,
+                        };
+                    }
+
+                    // ';' separates numeric parameters.
+                    b';' => {
+                        let next = idx.saturating_add(1);
+                        if (next as usize) < MAX_PARAMS {
+                            // If there were no digits for this param, it already stays None.
```
lrh2000:
```suggestion
                            // If there are no digits for this parameter, it will remain `None`.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+                            // the ANSI escape sequence.
+                            self.state = WaitFor::Escape;
+                        }
+                    }
+
+                    // The behavior of ':' is not defined by the standard. We don't support it,
+                    // so we are aborting the ANSI escape sequence.
+                    b':' => self.state = WaitFor::Escape,
+
+                    // Sequences containing "<=>?" are private. We swallow them and mark `is_private`.
```
lrh2000: This line also exceeds 100 characters..
```rust
                    // Sequences containing "<=>?" are private. We swallow them and mark
                    // `is_private`.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+            _ => {
                 self.state = WaitFor::Escape;
-                (i + 1) as usize
+                true
             }
-        };
+        }
+    }
+
+    /// Gets the parameter at the given index, or returns the default value if the parameter is not present.
```
lrh2000: This line also exceeds 100 characters...
```suggestion
    /// Gets the parameter at the given index, or returns the default value if it is absent.
```

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,48 @@ impl ConsoleState {
             ConsoleMode::Graphics
         }
     }
+
+    /// Fills a rectangular pixel region [x0, x1) × [y0, y1) with the given color.
```
lrh2000: This renders to:

Image

---

```rust
    /// Fills a rectangular pixel region `[x0, x1) × [y0, y1)` with the given color.
```
renders to:
Image

Will the latter version look better?

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,48 @@ impl ConsoleState {
...
+        let h = self.backend.height();
+        let x0 = x0.min(w);
+        let y0 = y0.min(h);
+        let x1 = x1.min(w);
+        let y1 = y1.min(h);
+
+        if x0 >= x1 || y0 >= y1 {
+            return;
+        }
+
```
lrh2000: Actually all these conditions cannot be hit? The caller always guarantee these conditions..

Replace them with a comment:
```rust
    /// Fills a rectangular pixel region `[x0, x1) × [y0, y1)` with the given color.
    ///
    /// The caller must pass arguments that form a valid rectangle within the framebuffer, i.e.,
    /// `x0 ≤ x1 ≤ w` and `y0 ≤ y1 ≤ h`, where `w` and `h` are the framebuffer width and height.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -18,15 +29,36 @@ pub(super) trait EscapeOp {
...
     fn set_bg_color(&mut self, val: Pixel);
+
+    /// Erases part or all of the display.
+    fn erase_in_display(&mut self, mode: EraseInDisplay);
 }
 
 const MAX_PARAMS: usize = 8;
+const MAX_OSC_LEN: usize = 256;
 
+// FIXME: Currently we only support a few ANSI escape sequences, and we just swallow the unsupported ones.
```
lrh2000: Again, this line is too long. Find a way to make it shorter or just wrap it:
```rust
// FIXME: Currently we only support a few ANSI escape sequences, and we just swallow the
// unsupported ones.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +101,253 @@ const COLORS: [Pixel; 16] = [
...
+                            len: 0,
+                            is_last_esc: false,
+                        }
+                    }
+
+                    // The character is invalid. We cannot handle it, so we are aborting the ANSI
+                    // escape sequence.
+                    _ => self.state = WaitFor::Escape,
+                }
+                true
```
lrh2000, line 165: Insert a blank line before line 154 and line 175.

If the `match` body contains blank lines, make sure there is a blank line before the final `true`. I believe this makes more sense because the final `true` works for all `match` arms.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -6,7 +6,18 @@ use crate::Pixel;
 #[derive(Debug)]
 pub(super) struct EscapeFsm {
     state: WaitFor,
-    params: [u32; MAX_PARAMS],
+    params: [Option<u32>; MAX_PARAMS],
+}
+
+/// The mode for "Erase in Display" (ED) command.
```
lrh2000:
```suggestion
/// The mode for "Erase in Display" (ED) commands.
```

wyt8:
> Thanks for the revision. I only have a few minor comments.

I really admire your rigorous attitude and thorough review.

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,47 @@ impl ConsoleState {
...
+    ///
+    /// The caller must pass arguments that form a valid rectangle within the framebuffer, i.e.,
+    /// `x0 ≤ x1 ≤ w` and `y0 ≤ y1 ≤ h`, where `w` and `h` are the framebuffer width and height.
+    fn fill_rect_pixels(&mut self, x0: usize, y0: usize, x1: usize, y1: usize, color: Pixel) {
+        let w = self.backend.width();
+        let h = self.backend.height();
+        let x0 = x0.min(w);
+        let y0 = y0.min(h);
+        let x1 = x1.min(w);
+        let y1 = y1.min(h);
```
lrh2000: https://github.com/asterinas/asterinas/pull/2998#discussion_r2893473062 also apply to the `min` operations? If you read the comments...

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +102,254 @@ const COLORS: [Pixel; 16] = [
...
+                    }
+
+                    _ => unreachable!(),
+                }
             }
-            (WaitFor::Params(_), b';') => {
-                // There are too many parameters. We cannot handle that many, so we are aborting
-                // the ANSI escape sequence.
+
+            // Parameter bytes in the intermediate section is illegal by the formal grammar.
```
lrh2000: Grammar:
```suggestion
            // Parameter bytes in the intermediate section are illegal by the formal grammar.
```

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +102,254 @@ const COLORS: [Pixel; 16] = [
...
             }
+        }
+    }
 
-            // Handle numeric parameters.
-            (WaitFor::Params(i), b'0'..=b'9') => {
-                let param = &mut self.params[i as usize];
-                *param = param.wrapping_mul(10).wrapping_add((byte - b'0') as u32);
-                return true;
+    fn handle_csi<T: EscapeOp>(
```
lrh2000: I think we should rename this to `parse_csi` and rename the `dispatch_csi` method to `handle_csi`.
The former does not handle anything; it just parses the arguments.

Also doc comments:
```rust
    /// Parses CSI arguments.
    fn parse_csi<T: EscapeOp>(

    /// Handles the "Control Sequence Introducer" sequence.
    fn handle_csi<T: EscapeOp>(
```

To be consistent with the doc comments of `handle_sgr`.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +102,254 @@ const COLORS: [Pixel; 16] = [
...
+                        self.state = WaitFor::Csi {
+                            idx,
+                            is_private,
+                            in_intermediate: false,
+                        };
+                    }
+
+                    // ';' separates numeric parameters.
+                    b';' => {
+                        let next = idx.saturating_add(1);
```
lrh2000:
```suggestion
                        let next = idx + 1;
```
This cannot overflow since we'll abort when we reach `MAX_PARAMS`?

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +102,254 @@ const COLORS: [Pixel; 16] = [
...
 
-            // Break and handle the final action.
-            (WaitFor::Params(i), _) => {
+            // Final byte (0x40..=0x7E): ends the CSI.
+            0x40..=0x7e => {
                 self.state = WaitFor::Escape;
-                (i + 1) as usize
+
+                let num_params = (idx as usize).saturating_add(1).min(MAX_PARAMS);
+
```
lrh2000:
```suggestion
                let num_params = idx + 1;
```

This cannot overflow and it cannot exceed `MAX_PARAMS`?

Additionally, the blank line (line 279) can be removed since the sole purpose of the `num_params` computation is to facilitate the `dispatch_csi` method call.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +102,254 @@ const COLORS: [Pixel; 16] = [
...
+            // Terminal behavior is undefined if a CSI contains bytes outside 0x20..=0x7E.
+            // We'll abort and swallow to avoid leaking garbage.
+            _ => {
+                self.state = WaitFor::Escape;
+            }
+        }
+    }
+
+    /// Gets the parameter at the given index, or returns the default value if it is absent.
+    fn param_or(&self, i: usize, default: u32) -> u32 {
```
lrh2000: Move this to the end of the `impl` block. Currently, `get_256_color` is a helper method for `handle_sgr`, so it is placed after `handle_sgr`. `param_or` is a helper method for `handle_csi` and `handle_sgr`, so it should be placed at the end.

lrh2000 (commented):
Thanks for the quick updates!

tatetian:
@boterinas codex

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -18,15 +29,37 @@ pub(super) trait EscapeOp {
...
+    /// This will be terminated by a byte in the range 0x40 through 0x7E.
+    Csi {
+        idx: u8,
+        is_private: bool,
+        in_intermediate: bool,
+    },
+    /// Waits for OSC payload.
+    ///
+    /// This will be terminated by a bell (BEL, 0x07) or a String Terminator (ST, "ESC\").
+    Osc { len: usize, is_last_esc: bool },
```
lrh2000: Since `idx` is a `u8`, let's choose a shorter type for this `len`. If we want to keep `MAX_OSC_LEN` equal to `256`, using `u16` here is fine? It can also be a `u8` if `MAX_OSC_LEN` can be reduced by a few bytes.

The type of `MAX_OSC_LEN` may also be changed since it is never used as an index. Being a `usize` does not provide any benefits.

kernel/comps/framebuffer/src/ansi_escape.rs
```diff
@@ -69,92 +102,250 @@ const COLORS: [Pixel; 16] = [
...
+                    // Terminate the OSC sequence by a BEL or a ST.
+                    Self::BEL => self.state = WaitFor::Escape,
+                    b'\\' if is_last_esc => self.state = WaitFor::Escape,
+
+                    // Swallow the OSC payload.
+                    _ => {
+                        if len > MAX_OSC_LEN {
+                            self.state = WaitFor::Escape;
+                            return false;
+                        }
```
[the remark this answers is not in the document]
lrh2000: Neither having this limit or not to having this limit is correct Linux behavior... The current form should be okay and we can leave the OSC sequence handling as future work.

kernel/comps/framebuffer/src/console.rs
```diff
@@ -293,6 +293,40 @@ impl ConsoleState {
...
+        }
+    }
+
+    /// Calculates the pixel coordinates for the cursor cell.
+    fn cursor_cell_rect(&self) -> (usize, usize, usize, usize) {
+        let cx0 = self.x_pos;
+        let cy0 = self.y_pos;
+        let cx1 = cx0 + self.font.width();
+        let cy1 = cy0 + self.font.height();
+        (cx0, cy0, cx1, cy1)
```
lrh2000, line 338: Sorry, this should be my last comment and it's really my mistake in my previous comment.

The type invariant is that `self.y_pos + self.font.height()` should never exceed the height of the framebuffer. However, the same does not hold true for `x`.
 - `self.newline()` will ensure `self.y_pos <= self.backend.height() - self.font.height()`, but we need to check `self.x_pos` before calling to `self.draw_char`.
https://github.com/asterinas/asterinas/blob/d4a04ea81b86d58cf53ee99dcf3e25a3f991c475/kernel/comps/framebuffer/src/console.rs#L177-L183
 - After changing `self.font`, we also ensures that `self.y_pos <= self.backend.height() - self.font.height()`.
https://github.com/asterinas/asterinas/blob/d4a04ea81b86d58cf53ee99dcf3e25a3f991c475/kernel/comps/framebuffer/src/console.rs#L262-L266

So... We should end up with something like:
```patch
diff --git a/kernel/comps/framebuffer/src/console.rs b/kernel/comps/framebuffer/src/console.rs
index 8b6a57e31..8fef2ff4b 100644
--- a/kernel/comps/framebuffer/src/console.rs
+++ b/kernel/comps/framebuffer/src/console.rs
@@ -299,6 +299,9 @@ impl ConsoleState {
     /// The caller must pass arguments that form a valid rectangle within the framebuffer, i.e.,
     /// `x0 ≤ x1 ≤ w` and `y0 ≤ y1 ≤ h`, where `w` and `h` are the framebuffer width and height.
     fn fill_rect_pixels(&mut self, x0: usize, y0: usize, x1: usize, y1: usize, color: Pixel) {
+        debug_assert!(x1 <= self.backend.width());
+        debug_assert!(y1 <= self.backend.height());
+
         let rendered_pixel = self.backend.render_pixel(color);
         let rendered_pixel_size = rendered_pixel.nbytes();
         let row_bytes = (x1 - x0) * rendered_pixel_size;
@@ -322,9 +325,16 @@ impl ConsoleState {
     /// Calculates the pixel coordinates for the cursor cell.
     fn cursor_cell_rect(&self) -> (usize, usize, usize, usize) {
         let cx0 = self.x_pos;
-        let cy0 = self.y_pos;
         let cx1 = cx0 + self.font.width();
+        // `min` is necessary if we are at the end of the line.
+        let cx1 = cx1.min(self.backend.width());
+
+        let cy0 = self.y_pos;
         let cy1 = cy0 + self.font.height();
+        // This will always hold because we will scroll if it is no longer true after a new line or
+        // font change.
+        debug_assert!(cy1 <= self.backend.height());
+
         (cx0, cy0, cx1, cy1)
     }
 }
 
```
I tested the code and found that the previous version may cause panics for some odd cursor positions (e.g., the last character on the screen and `echo -ne '\x1b[1J' > /dev/tty1` leads OOB writes to the framebuffer, and then panics due to the invalid offset.). However, the new code should be fine.

Issue #2939, opened 2026-01-29
Tracking issue for supporting Asterinas as the guest kernel of CoCo


taosue:
# Overview

## Background

As discussion in #2911, traditional Kata Containers inherently trust the host. In confidential computing scenarios, the host is considered untrusted - a fundamental shift in security assumptions.

Given that many Trusted Execution Environment (TEE) technologies rely on a strong isolation boundary between host and guest — often enforced through KVM — the VM-based architecture of Kata Containers provides a natural foundation for confidential computing. Building upon this, Confidential Containers (CoCo) extends Kata to deliver a framework for secure workload deployment.

This issue proposes adopting Asterinas as the guest kernel for CoCo, marking a strategic step toward enabling Asterinas in real-world confidential computing scenarios.

## Workflow Summary

The diagram below, sourced from the [CoCo repository](https://github.com/confidential-containers/confidential-containers/blob/main/images/CC_TEE_container.png), illustrates the architecture of CoCo:
<img width="1382" height="761" alt="Image" src="https://github.com/user-attachments/assets/48fce9bb-ad77-4040-a409-750ad8832286" />

The workflow is like:
- The CoCo runtime on the host initiates a VM-based TEE (e.g. Intel TDX).
- The enclave agent inside the guest performs remote attestation (Steps 1–2).
- Upon successful attestation, the enclave retrieves cryptographic keys needed to verify or decrypt container images (Steps 3–4).
- The image management component downloads the encrypted container image(s) from the registry (Step 5).
- The image is verified and decrypted within the secure guest environment (Step 6).
- Finally, the container runtime inside the enclave launches the workload securely.

## Key Differences Between CoCo and Kata

Although CoCo builds on the foundation of Kata Containers, several critical modifications are required to ensure data confidentiality.

### Secure Image Pulling

In standard Kata Containers deployments, container images are pulled on the host and then passed into the guest via mechanisms like `virtio-fs`. This approach exposes sensitive image data to the host — an unacceptable risk in confidential computing.

In contrast, CoCo ensures that images are pulled and unpacked entirely within the guest VM, preserving confidentiality. To achieve this, additional components `image-rs` is embedded into the guest's root filesystem — typically layered using `overlayfs`.

### Remote Attestation Components

Attestation is a cornerstone of confidential computing. It enables a remote party to verify the integrity and authenticity of the execution environment before releasing sensitive data.

Inside the guest, two key components manage attestation workflows:
[...]

PR #3005, opened 2026-03-04
Support for TDX on-demand lazy accept memory

Hsy-Intel:
# Background and Goal

Background from Linux implementation: 

> UEFI Specification version 2.9 introduces the concept of memory acceptance: some Virtual Machine platforms, such as Intel TDX or AMD SEV-SNP, requiring memory to be accepted before it can be used by the guest. Accepting happens via a protocol specific for the Virtual Machine platform.
> Accepting memory is costly and it makes VMM allocate memory for the accepted guest physical address range. It's better to postpone memory acceptance until memory is needed. It lowers boot time and reduces memory overhead.
> There are several ways kernel can deal with unaccepted memory:
> 1. Accept all the memory during the boot. It is easy to implement and it doesn't have runtime cost once the system is booted. The downside is very long boot time. Accept can be parallelized to multiple CPUs to keep it manageable (i.e. via DEFERRED_STRUCT_PAGE_INIT), but it tends to saturate memory bandwidth and does not scale beyond the point.
> 2. Accept a block of memory on the first use. It requires more infrastructure and changes in page allocator to make it work, but it provides good boot time. On-demand memory accept means latency spikes every time kernel steps onto a new memory block. The spikes will go away once workload data set size gets stabilized or all memory gets accepted.
> 3. Accept all memory in background. Introduce a thread (or multiple) that gets memory accepted proactively. It will minimize time the system experience latency spikes on memory allocation while keeping low boot time. This approach cannot function on its own. It is an extension of 2: background memory acceptance requires functional scheduler, but the page allocator may need to tap into unaccepted memory before that. The downside of the approach is that these threads also steal CPU cycles and memory bandwidth from the user's workload and may hurt user experience.
> Linux implement 1 and 2 for now. 2 is the default. Some workloads may want to use 1 with `accept_memory=eager` in kernel command line.

Currently, Asterinas handles unaccepted memory by calling TDCALL during the EFI stub phase to accept all unaccepted memory at once. This is similar to Solution 1 in Linux, which, while simple enough, significantly increases boot time, and not all pages may be used during the OS's lifetime.

To shorten boot time and optimize resource utilization, this PR, referencing the Linux implementation, adds on-demand accept functionality (Solution 2 in Linux) and accept all memory in background(Solution 3 not implemented in Linux) for dynamically accepting this type of memory.

# Implementation overview

The implementation follows a two-stage model:

EFI stub stage:
- Build an “unaccepted memory bitmap” from firmware memory map.
- Only accept memory required for boundary cases and for booting the kernel image.
- Publish bitmap header and metadata via EFI config table.

Kernel stage:
- Parse and import the bitmap table.
- In early frame allocators, accept physical memory through bitmap information.
- Create `UNACCEPTED_POOL` in the frame-allocator crate to manage unaccepted memory in physical memory.
- On each physical allocation path, accept memory only if bitmap says the range is still unaccepted.
- Provide the size of currently unaccepted mem in `proc/meminfo`.
- Run a background accept worker to accept all unaccepted memory if `lazy-background` mode is configured.

To support the above features, the `EfiUnacceptedMemory` structure and related abstractions (similar to `struct efi_unaccepted_memory` in Linux) and efficient bitmap management have been added to the `tdx-guest` crate.

Users can control the acceptance policy for unacceptable memory by adding `accept_memory=$(ACCEPT_MEMORY_MODE)` to the kernel command line or by adding the `ACCEPT_MEMORY_MODE` parameter (`eager` or `lazy` or `lazy-background`) to the `make run_kernel INTEL_TDX=1` command line. And the `lazy` is the default mode.

The current amount of unaccepted memory can be queried via `/proc/meminfo`.

# Limitation

- Better watermark control strategies.
- Better background acceptance strategies.
- SMP support and stability testing
- No KASLR(kernel address space layout randomization)-driven acceptance policy (Asterinas has no KASLR).
- No `kexec` acceptance policy (Asterinas does not currently support the kexec feature).
- Get unaccepted page num from `vmstat` info (Asterinas does not currently support this virtual file system).

### DIFF
--- a/kernel/comps/cmdline/src/unimplemented.rs
+++ b/kernel/comps/cmdline/src/unimplemented.rs
@@ -54,3 +54,27 @@ define_unimplemented_param!(
     "noreplace_smp",
     "initcall_debug"
 );
+
+/// Defines kernel command-line parameters that are handled by OSTD during early boot.
+///
+/// OSTD parses these parameters before the kernel command-line framework initializes.
+/// Registration here prevents them from being forwarded to the init process.
+#[macro_export]
+macro_rules! define_ostd_handled_param {
+    ($($name:expr),+ $(,)?) => {
+        $(
+            const _: () = {
+                fn __kparam_setup(_occurrences: &[Option<&str>]) {
+                    // Intentionally empty: OSTD handles this parameter during
+                    // early boot, before the kernel command-line framework runs.
+                }
+                $crate::submit! {
+                    $crate::KernelParam::new($name, __kparam_setup, false)
+                }
+            };
+        )+
+    };
+}
+
+// Parameters handled directly by OSTD during early boot.
+define_ostd_handled_param!("accept_memory");
--- a/ostd/libs/linux-bzimage/setup/src/x86/amd64_efi/efi.rs
+++ b/ostd/libs/linux-bzimage/setup/src/x86/amd64_efi/efi.rs
@@ -1,14 +1,43 @@
 // SPDX-License-Identifier: MPL-2.0
 
+extern crate alloc;
+
+use alloc::vec::Vec;
 use core::{ffi::CStr, mem::MaybeUninit};
 
 use boot::{AllocateType, open_protocol_exclusive};
+use cfg_if::cfg_if;
 use linux_boot_params::BootParams;
 use uefi::{mem::memory_map::MemoryMap, prelude::*, proto::BootPolicy};
 use uefi_raw::table::system::SystemTable;
 
 use super::decoder::decode_payload;
 use crate::x86::amd64_efi::alloc::alloc_pages;
+
+cfg_if! {
+    if #[cfg(feature = "cvm_guest")] {
+        use align_ext::AlignExt;
+        use tdx_guest::{
+            is_tdx_guest_early,
+            unaccepted_memory::EfiUnacceptedMemory,
+        };
+        use super::unaccepted_memory::{
+            UnacceptedRange, accept_bitmap_range, allocate_unaccepted_bitmap, find_unaccepted_table,
+            install_unaccepted_bitmap,
+        };
+
+        /// Safety margin added to the kernel footprint to account for
+        /// any additional memory that may be accessed during early boot.
+        const KERNEL_FOOTPRINT_SAFETY_MARGIN: u64 = 2 * 1024 * 1024;
+
+        /// AP execution page range required for AP startup.
+        /// In the linker script, AP_EXEC_MA = 0x8000.
+        const AP_EXEC_RANGE: (u64, u64) = (0x8000, 0x9000);
+
+        /// Kernel load ranges that must be accepted before entering the kernel.
+        type PreAcceptedRanges = Vec<(u64, u64)>;
+    }
+}
 
 pub(super) const PAGE_SIZE: u64 = 4096;
 
@@ -37,11 +66,11 @@ unsafe extern "sysv64" fn main_efi_common64(
         unsafe { &mut *boot_params_ptr }
     };
 
-    efi_phase_boot(boot_params);
+    let early_boot_info = efi_phase_boot(boot_params);
 
     // SAFETY: All previously opened boot service protocols have been closed. At this time, we have
     // no references to the code and data of the boot services.
-    unsafe { efi_phase_runtime(boot_params) };
+    unsafe { efi_phase_runtime(boot_params, early_boot_info) };
 }
 
 fn allocate_boot_params() -> &'static mut BootParams {
@@ -57,7 +86,7 @@ fn allocate_boot_params() -> &'static mut BootParams {
     boot_params
 }
 
-fn efi_phase_boot(boot_params: &mut BootParams) {
+fn efi_phase_boot_common(boot_params: &mut BootParams) -> Vec<u8> {
     uefi::println!(
         "[EFI stub] Loaded with offset {:#x}",
         crate::x86::image_load_offset(),
@@ -95,11 +124,28 @@ fn efi_phase_boot(boot_params: &mut BootParams) {
         fill_screen_info(&mut boot_params.screen_info);
     }
 
+    // Fill the EFI info in boot params.
+    fill_efi_info(boot_params);
+
     // Decode the payload and load it as an ELF file.
     uefi::println!("[EFI stub] Decoding the kernel payload");
     let kernel = decode_payload(crate::x86::payload());
     uefi::println!("[EFI stub] Loading the payload as an ELF file");
     crate::loader::load_elf(&kernel);
+
+    kernel
+}
+
+#[cfg(feature = "cvm_guest")]
+fn efi_phase_boot(boot_params: &mut BootParams) -> PreAcceptedRanges {
+    let kernel = efi_phase_boot_common(boot_params);
+
+    compute_kernel_load_ranges(&kernel)
+}
+
+#[cfg(not(feature = "cvm_guest"))]
+fn efi_phase_boot(boot_params: &mut BootParams) {
+    let _kernel = efi_phase_boot_common(boot_params);
 }
 
 fn load_cmdline() -> Option<&'static CStr> {
@@ -110,14 +156,14 @@ fn load_cmdline() -> Option<&'static CStr> {
             .unwrap();
 
     let Some(load_options) = loaded_image.load_options_as_bytes() else {
-        uefi::println!("[EFI stub] Warning: No cmdline is available!");
+        uefi::println!("[EFI stub] warning: command line is unavailable");
         return None;
     };
 
     if !load_options.len().is_multiple_of(2)
         || load_options.iter().skip(1).step_by(2).any(|c| *c != 0)
     {
-        uefi::println!("[EFI stub] Warning: The cmdline contains non-ASCII characters!");
+        uefi::println!("[EFI stub] warning: command line contains non-ASCII characters");
         return None;
     }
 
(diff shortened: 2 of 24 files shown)


Hsy-Intel:
Related to #2939

Hsy-Intel:
@boterinas codex

kernel/comps/cmdline/src/unimplemented.rs
```diff
@@ -54,3 +54,27 @@ define_unimplemented_param!(
...
+                $crate::submit! {
+                    $crate::KernelParam::new($name, __kparam_setup, false)
+                }
+            };
+        )+
+    };
+}
+
+// Parameters handled directly by OSTD during early boot.
+define_ostd_handled_param!("accept_memory");
```
taosue, line 80: I don't think `accept_memory` should be registered through `define_ostd_handled_param!` in `unimplemented.rs`.

This parameter is not unimplemented. It is a supported Linux-defined kernel parameter, but it has to be parsed by OSTD before the `aster-cmdline` component is initialized. The kernel-side registration is only needed to prevent it from being forwarded to the init process.

The `define_ostd_handled_param!` abstraction is also a bit confusing: native OSTD parameters such as `ostd.xxx` do not need this registration, since module-style parameters containing `.` are already not forwarded to `init`. So this is really a narrower case: Linux-defined parameters supported by Asterinas but consumed earlier by OSTD.

Can we register it explicitly near the x86/TDX code instead? A simple declaration in `kernel/src/arch/x86/mod.rs` (or `kernel/src/device/misc/tdxguest.rs`?) should be enough:

```rust
// OSTD parses `accept_memory` before the kernel cmdline component is initialized.
// Register it here so this Linux-defined kernel parameter is not forwarded to init.
static _ACCEPT_MEMORY: Once<String> = Once::new();
aster_cmdline::define_kv_param!("accept_memory", _ACCEPT_MEMORY);
```

This keeps the parameter typed as a normal key-value kernel parameter and makes its ownership clearer.

ostd/libs/linux-bzimage/setup/src/x86/amd64_efi/efi.rs
```diff
@@ -3,13 +3,44 @@
...
+        const AP_EXEC_RANGE: (u64, u64) = (0x8000, 0x9000);
+
+        /// Kernel load ranges computed from the ELF payload during the boot phase.
+        type EarlyBootInfo = Vec<(u64, u64)>;
+    }
+}
+
+#[cfg(not(feature = "cvm_guest"))]
+type EarlyBootInfo = ();
+
```
taosue: The `EarlyBootInfo` alias feels confusing to me. With `cvm_guest` enabled it is `Vec<(u64, u64)>`, but without `cvm_guest` it becomes `()`. The name also does not describe what the data actually represents.

Can we avoid this cross-cfg alias by extracting the common EFI boot logic and making the CVM-only return value explicit? For example:

```rust
fn efi_phase_boot_common(boot_params: &mut BootParams) -> KernelPayload {
  // load cmdline/initrd, fill ACPI/framebuffer/EFI info,
  // decode and load the kernel ELF, then return the decoded payload
}

#[cfg(feature = "cvm_guest")]
fn efi_phase_boot(boot_params: &mut BootParams) -> PreAcceptedRanges {
  let kernel = efi_phase_boot_common(boot_params);
  compute_kernel_load_ranges(&kernel)
}

#[cfg(not(feature = "cvm_guest"))]
fn efi_phase_boot(boot_params: &mut BootParams) {
  let _kernel = efi_phase_boot_common(boot_params);
}
```

This would make the non-CVM path return nothing, and make the CVM path explicitly return the ranges that must be accepted before entering the kernel. It also avoids giving one alias two very different meanings across feature configurations.

ostd/libs/linux-bzimage/setup/src/x86/amd64_efi/efi.rs
```diff
@@ -408,21 +472,136 @@ fn parse_memory_type(
...
+        .entries()
+        .filter(|e| e.ty == uefi::table::boot::MemoryType::UNACCEPTED)
+        .map(|e| UnacceptedRange {
+            start: e.phys_start,
+            end: e.phys_start + e.page_count * PAGE_SIZE,
+        })
+        .collect();
+
+    let unaccepted_table = if let Some(existing) = find_unaccepted_table() {
+        Some(core::ptr::NonNull::from(existing))
```
taosue: In our current TDX setup we generally run with OVMF. If OVMF already provides the unaccepted-memory table, then the fallback path that allocates and installs a new bitmap table in the EFI stub may be untested code.

Could you check whether OVMF provides this table? If so, maybe we can drop the installation for the initial lazy-accept? If not, we should document this explicitly.

ostd/libs/linux-bzimage/setup/src/x86/amd64_efi/efi.rs
```diff
@@ -408,21 +472,136 @@ fn parse_memory_type(
...
+        if !accept_bitmap_range(table, k_start, k_end) {
+            panic!("[EFI stub] failed to accept kernel mapped unaccepted memory via bitmap");
+        }
+    }
+}
+
+#[cfg(feature = "cvm_guest")]
+type LazyAcceptInfo = (
+    Vec<UnacceptedRange>,
+    Option<core::ptr::NonNull<EfiUnacceptedMemory>>,
```
taosue: Can we avoid returning `Option<NonNull<EfiUnacceptedMemory>>` from `prepare_lazy_accept_info()`?

The table is found or allocated in `prepare_lazy_accept_info()` and then immediately consumed by `apply_lazy_accept_info()`, where it is cast back to `&mut EfiUnacceptedMemory`. This makes the EFI-stub flow more C-like than necessary and exposes raw pointer handling.

A small typed context seems clearer here, for example:

```rust
struct LazyAcceptInfo {
    unaccepted_ranges: Vec<UnacceptedRange>,
    table: Option<&'static mut EfiUnacceptedMemory>,
}
```

Then `prepare_lazy_accept_info()` can return `LazyAcceptInfo`, and `apply_lazy_accept_info()` can consume it directly. This keeps the lifetime/ownership information in the type system and avoids an extra raw-pointer round trip in the main flow.

ostd/src/arch/x86/boot/linux_boot/mod.rs
```diff
@@ -210,6 +218,13 @@ unsafe extern "sysv64" fn __linux_boot(params_ptr: *const BootParams) -> ! {
     #[cfg(feature = "cvm_guest")]
     init_cvm_guest();
 
+    #[cfg(feature = "cvm_guest")]
+    if is_tdx_guest_early() {
```
taosue: After `init_cvm_guest()` has run, I don't think we should use `is_tdx_guest_early()` here. That helper is documented for very early environments such as the EFI stub, where full TDX initialization is not available yet. At this point `init_cvm_guest()` has already called `tdx_guest::init_tdx()`, so the initialized runtime state should be used instead.

ostd/src/mm/frame/unaccepted.rs
```diff
@@ -0,0 +1,1008 @@
...
+                    .map(|bytes| bytes.max(PAGE_SIZE).align_up(PAGE_SIZE))
+                    .unwrap_or(PAGE_SIZE),
+                Ordering::Release,
+            );
+
+            let unit_size = u64::from(t.unit_size_bytes());
+            // SAFETY: `t` points to valid table memory from boot info.
+            let total_bits = unsafe { t.pending_unit_count() }.unwrap_or(0);
+            let initial_bytes = total_bits.saturating_mul(unit_size);
+            TOTAL_UNACCEPTED_BYTES.store(initial_bytes as usize, Ordering::Release);
```
taosue, line 106: Can we add a `pending_bytes()` helper to `EfiUnacceptedMemory`. The table already exposes derived queries like `total_coverage_size()` and `bitmap_coverage_end()`. A `pending_bytes()` (= `pending_unit_count()` * `unit_size`) would fit the same pattern, eliminating the need for OSTD to manually extract raw fields and reconstruct this calculation.

Should `EfiUnacceptedMemory` itself maintain the pending-bytes counter? Currently OSTD keeps a shadow `TOTAL_UNACCEPTED_BYTES` atomic that's initialized here by scanning the bitmap (`O(n)` via `pending_unit_count()`) and then decremented on each successful accept. The reason is clear — re-scanning the bitmap on every query would be too expensive. But this means OSTD must carefully keep the counter in sync with the actual bitmap state (including rollback paths). If `EfiUnacceptedMemory` maintained its own atomic counter (updated in `accept_range` / `restore_pending_range` internally), the bookkeeping stays co-located with the bitmap mutation logic, eliminates the risk of OSTD's counter drifting from reality, and makes the table a more self-contained abstraction. It would simplify the OSTD side to just calling `table.pending_bytes()` without needing the separate `TOTAL_UNACCEPTED_BYTES` at all.

taosue (commented):
Thanks a lot for working on this PR. This is an important piece for making CoCo support more production-ready. The PR not only implements TDX lazy memory acceptance, but also includes several thoughtful optimizations. But this PR has been pending for quite a while. To help move it forward, I would like to give two high-level suggestions.
- **Split the PR.** The reservoir and shard-based accept protocol look like optimization layers rather than the minimal mechanism needed for lazy acceptance correctness. Since the primary motivation is reducing TDX boot time, I think the first PR should implement the simplest on-demand accept path and report boot-time improvement plus runtime benchmark impact. (I believe only the simplest lazy accept can obviously descrese the boot time because now only one thread accepts all pages one by one). Reservoir/refill and sharded concurrent accept should be split into follow-up PRs, **each with benchmark data showing the additional benefit**. This would make the correctness review smaller and make the performance tradeoff measurable.
- **Introduce the design in the community meeting.** Following the usual Asterinas workflow, a complex PR like this would benefit from a high-level design discussion in the community before going too deep into code review. This PR touches several areas, so discussion in the community would help the community quickly understand the motivation, the expected behavior, and the implementation boundaries. It would also make it easier for reviewers with different expertise to join the review.

PR #3007, opened 2026-03-05
Make `try_upgrade` private in `RwLock` and `RwMutex`

### LINKED ISSUE #3006: `RwLockUpgradeableGuard::try_upgrade` can never succeed if called directly
### Describe the bug
The public API `RwLockUpgradeableGuard::try_upgrade` can only succeed when the `AtomicUsize` value is `UPGRADEABLE_READER | BEING_UPGRADED`.
https://github.com/asterinas/asterinas/blob/b205cacbf5e5406172e5c7216179c2559c091885/ostd/src/sync/rwlock.rs#L379-L385

However, the only place where the `BEING_UPGRADED` is set is in `RwLockUpgradeableGuard::upgrade`:
https://github.com/asterinas/asterinas/blob/b205cacbf5e5406172e5c7216179c2559c091885/ostd/src/sync/rwlock.rs#L367-L375

As a result, `try_upgrade` can never succeed when called directly by an OSTD user, even if there are no readers or writers.

### Expected behavior
The `try_upgrade` method should succeed if there are no readers or writers.

I propose two possible fixes
- Make `try_upgrade` private, since it is currently unused.
- Change the condition in `try_upgrade.` from `UPGRADEABLE_READER | BEING_UPGRADED` to `UPGRADEABLE_READER`. 
 
The second option requires careful review to ensure that relaxing the condition does not introduce soundness issues.

rikosellic:
Fix #3006. The same pattern is also found in `RwMutex`.

### DIFF
--- a/ostd/src/sync/rwlock.rs
+++ b/ostd/src/sync/rwlock.rs
@@ -373,10 +373,14 @@ impl<'a, T: ?Sized, G: SpinGuardian> RwLockUpgradeableGuard<'a, T, G> {
             };
         }
     }
+
     /// Attempts to upgrade this upread guard to a write guard atomically.
     ///
     /// This function will never spin-wait and will return immediately.
-    pub fn try_upgrade(mut self) -> Result<RwLockWriteGuard<'a, T, G>, Self> {
+    ///
+    /// This function is not exposed publicly because the `BEING_UPGRADED` bit
+    /// is set only in [`Self::upgrade`].
+    fn try_upgrade(mut self) -> Result<RwLockWriteGuard<'a, T, G>, Self> {
         let res = self.inner.lock.compare_exchange(
             UPGRADEABLE_READER | BEING_UPGRADED,
             WRITER | UPGRADEABLE_READER,
--- a/ostd/src/sync/rwmutex.rs
+++ b/ostd/src/sync/rwmutex.rs
@@ -340,7 +340,10 @@ impl<'a, T: ?Sized> RwMutexUpgradeableGuard<'a, T> {
     /// Attempts to upgrade this upread guard to a write guard atomically.
     ///
     /// This function will return immediately.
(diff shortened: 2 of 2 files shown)


ostd/src/sync/rwlock.rs
```diff
@@ -376,7 +376,9 @@ impl<'a, T: ?Sized, G: SpinGuardian> RwLockUpgradeableGuard<'a, T, G> {
     /// Attempts to upgrade this upread guard to a write guard atomically.
     ///
     /// This function will never spin-wait and will return immediately.
-    pub fn try_upgrade(mut self) -> Result<RwLockWriteGuard<'a, T, G>, Self> {
+    /// It is not exposed publicly because the `BEING_UPGRADED` bit
+    /// is set only in the upgrade method.
+    fn try_upgrade(mut self) -> Result<RwLockWriteGuard<'a, T, G>, Self> {
```
lrh2000:
```suggestion
    ///
    /// This function is not exposed publicly because the `BEING_UPGRADED` bit
    /// is set only in [`Self::upgrade`].
    fn try_upgrade(mut self) -> Result<RwLockWriteGuard<'a, T, G>, Self> {
```

 1. Insert a blank line before. This paragraph is not related to the previous one.
 2. If you use doc comments here, reference the method with hyperlinks.

ostd/src/sync/rwlock.rs
```diff
@@ -376,7 +376,9 @@ impl<'a, T: ?Sized, G: SpinGuardian> RwLockUpgradeableGuard<'a, T, G> {
     /// Attempts to upgrade this upread guard to a write guard atomically.
```
lrh2000, line 377: By the way, could you please also add a blank line before this line? Thanks!

ostd/src/sync/rwlock.rs
rikosellic: Thanks! Fixed the doc.

PR #3009, opened 2026-03-05
Change `FileDesc` to `u32`

### LINKED ISSUE #2907: Make `FileDesc` unsigned integer(`u32`) instead of signed(`i32`)
Currently, `FileDesc` is defined as an `i32`. In practice, however, a file descriptor can never be negative. This was a design mistake introduced early in the project.

Following Rust’s principle of making invalid states unrepresentable, we should change this type to `u32`, just as we already do for `Pid`, `Pgid`, `Uid`, and `Gid`, which are all defined as `u32`.

chen1plus:
Closes #2907 

The syscall portion continues to use `RawFileDesc`, but all other sections now use `FileDesc`.
I noticed that some parts of the code use `fd` while others use `file_desc`. It might be a good idea to standardize these variable names.

### DIFF
--- a/test/initramfs/src/apps/fs/pseudofs/pseudo_file_create.h
+++ b/test/initramfs/src/apps/fs/pseudofs/pseudo_file_create.h
@@ -61,8 +61,8 @@ static void __attribute__((unused)) fd_path(int fd, char *buf, size_t buflen)
 		   _ret > 0 && _ret < buflen);
 }
 
-static void __attribute__((unused)) fdinfo_path(int fd, char *buf,
-						size_t buflen)
+static void __attribute__((unused))
+fdinfo_path(int fd, char *buf, size_t buflen)
 {
 	CHECK_WITH(snprintf(buf, buflen, "/proc/self/fdinfo/%d", fd),
 		   _ret > 0 && _ret < buflen);


chen1plus:
The current logic remains the same. Should I handle these error-handling issues in this PR, or would you prefer me to address them in a new one?

StevenJiang1110:
> The current logic remains the same. Should I handle these error-handling issues in this PR, or would you prefer me to address them in a new one?

What do you mean by "error-handling issues"? If the issues were introduced by this PR, they must be fixed within this PR. If the issues existed prior to this PR, you can choose to fix them here in a separate commit, or address them in a new PR — that is up to your own judgement. Our principle is that no new issues should be introduced.

Additionally, it looks like quite a few CI checks are currently failing. Some of the failures are formatting-related — you can run `make format` locally to fix trivial formatting issues, and `make check` to catch any remaining issues flagged by the other format checkers. There also appear to be some test failures, which may indicate logic issues introduced by this PR's changes, and those will need to be investigated and fixed as well.

chen1plus:
Fixed the formatting issues.

> What do you mean by "error-handling issues"?

Converting `i32` to `u32` without error handling is a pre-existing issue, not a newly introduced one. The test failed because the `fd` was outside the valid `0 to i32::MAX` range. Linux returns `EBADF` for this, but the original test incorrectly assumed it would succeed.

I think it makes more sense to open a PR to fix the tests that don't align with Linux behavior before handling this PR.

StevenJiang1110:
> I think it makes more sense to open a PR to fix the tests that don't align with Linux behavior before handling this PR.

Sure, feel free to go ahead with that.

chen1plus:
Closing this PR to move the changes to a dedicated feature branch for better management. I will resubmit a new PR shortly. Sorry for the noise!

PR #3010, opened 2026-03-05
Introduce Kernel Parameter Framework

### LINKED ISSUE #2982: Introduce Kernel Parameter Framework
# Motivation & Background
## Current State
Asterinas currently adheres strictly to the [Linux's parsing rule of command line](https://www.kernel.org/doc/html/v6.18/admin-guide/kernel-parameters.html):
- The kernel parses parameters from the command line up to the delimiter `--`.
- Everything after `--` is unconditionally passed to `init` as an argument.
- Undefined parameters without a `.` are passed to `init`. Parameters containing `=` are injected into `init`'s environment; others are passed as `init`'s arguments.

At present, Asterinas supports only a minimal set of Linux kernel parameters:
```
init
console
i8042.exist
```
When Asterinas is running as a guest OS in Kata Containers or Confidential Containers (CoCo), standard Linux command lines are often used. In such cases, numerous undefined kernel parameters are incorrectly forwarded to `init`, polluting its argument list and environment space. This can lead to **fatal boot failures**.
The root cause lies in **behavioral inconsistency**: although Asterinas follows Linux's syntactic parsing rules, it diverges semantically. When a parameter is defined in Linux but not in Asterinas, Linux consumes it internally and does not pass it to `init`. In contrast, Asterinas treats it as undefined and forwards it—violating expected compatibility.
To enable rapid convergence with Linux behavior (ensuring system bootability even if a parameter is not yet fully implemented), we cannot simply classify parameters as "defined" or "undefined." Instead, we should introduce a new category: **"unimplemented"**—a mechanism that acknowledges the existence of a parameter without providing an implementation. This prevents unintended propagation to `init`.
Furthermore, as Asterinas evolves rapidly, support for Linux parameters will expand significantly. **The number of defined parameters will grow quickly, making scalability and maintainability critical concerns.**
## Limitations of the Current Implementation
Currently, adding a new kernel parameter requires intrusive modifications in the `cmdline` component:
### Centralized Matching Logic
```rust
// KernelOptionName => /*literal string alternatives*/ | /*init environment*/
if let Some(value) = value {
    // The option has a value.
    match option {
        "init" => {
            if result.initproc.path.is_some() {
                panic!("[KCmdline] Init process specified twice");
            }
            result.initproc.path = Some(value.to_string());
        }
        "console" => {
            result.console_names.push(value.to_string());
        }
        _ => {
            // If the option is not recognized, it is passed to the init process.
            // Pattern 'option=value' is treated as the init environment.
            let envp_entry = CString::new(option.to_string() + "=" + value).unwrap();
            result.initproc.envp.push(envp_entry);
        }
    }
```
### Monolithic Data Structure
```rust
pub struct KCmdlineArg {
    initproc: InitprocArgs,
    console_names: Vec<String>,
    module_args: BTreeMap<String, Vec<ModuleArg>>,
}
```
This design violates the [Open-Closed Principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle): every new parameter requires changes to existing code. It lacks modularity and prevents components from declaring their own parameters locally. An ideal solution should allow distributed, decentralized parameter definitions within individual components.
### Issues with Module Parameters
Asterinas currently collects all module-style parameters (those containing `.`) using the following logic:
```rust
if let Some(modname) = node {
    let modarg = if let Some(v) = value {
        ModuleArg::KeyVal(
            CString::new(option.to_string()).unwrap(),
            CString::new(v).unwrap(),
        )
    } else {
        ModuleArg::Arg(CString::new(option).unwrap())
    };
    result
        .module_args
        .entry(modname.to_string())
        .and_modify(|v| v.push(modarg.clone()))
        .or_insert(vec![modarg.clone()]);
    continue;
}
```
This approach suffers from two major drawbacks:
- Inefficient lookup: Accessing module parameters requires scanning a global map.
- Information leakage: Components may access parameters not intended for them, breaking encapsulation.
## Linux's Parameter Handling Mechanism
Linux uses macro-based declarations to define kernel and module parameters, enabling loose coupling and automatic registration.
### Kernel Parameters
```c
__setup(str, fn);
early_param(str, fn);
```
For example, the `init= ` parameter is defined as:
```c
static char *execute_command;
static int __init init_setup(char *str)
{
	execute_command = str;
	return 1;
}
__setup("init=", init_setup);
```
The handler follows a uniform signature:
```c
int (*setup_func)(char *);
```
### Module Parameters
```c
module_param(name, type, perm);
core_param(name, var, type, perm);
```
Example:
```c
static bool i8042_dumbkbd;
module_param_named(dumbkbd, i8042_dumbkbd, bool, 0);
```
During boot (`start_kernel()`), the kernel scans the command line and matches against all registered parameters, invoking their respective handlers with the provided values. Parameter definitions are automatically collected via linker sections, eliminating the need for manual lists.
## Design Requirements
To address these issues, we propose a new kernel parameter framework for Asterinas that meets the following requirements:
1. **Adherence to the Open-Closed Principle**
Kernel parameter definitions should follow the open-closed principle—kernel parameters should be defined within the component that owns them, distributed across various files.
2. **Strong Typing via Rust's Type System**
Parameter definitions should bind their Rust types at compile time (e.g., `bool`, `u8`). Support for both single and repeated occurrences (e.g., multiple `console=` entries) is required.
3. **Initialization Order Control**
Components that register kernel parameters must be initialized after the command-line parsing subsystem, ensuring parameters are available before use.
4. **Support for Module Parameters (Including Those with `.`)**
Although Asterinas does not support loadable kernel modules, it should still support module-style parameters—those containing a dot (`.`) in their name, such as `i8042.exist` or `i8042.dumbkbd=1`. These are widely used in Linux to configure device behavior, even for built-in drivers, and are part of the standard boot interface.
5. **Compatibility with Linux: Support for "Unimplemented" Parameters**
To maintain compatibility with real-world workloads (e.g., Kata, OVMF), we should support marking parameters as unimplemented—defined but not fully supported—so they are not passed to `init`.

Examples of such parameters include:
```
tsc=reliable
no_timer_check
rcupdate.rcu_expedited=1
i8042.direct=1
i8042.dumbkbd=1
i8042.nopnp=1
i8042.noaux=1
noreplace-smp
reboot=k
cryptomgr.notests
pci=lastbus=0
debug
panic=1
nr_cpus=1
selinux=0
net.ifnames=0
scsi_mod.scan=none
initrd=initrd
```
# Design
## Overview
Example Usage:
```rust
static LOG_LEVEL: spin::Once<u8> = spin::Once::new();
define_kernel_param!("log_level=", LOG_LEVEL);

static CONSOLES: Mutex<Vec<Console>> = Mutex::new(Vec::new());
define_kernel_param_vec!("console=", CONSOLES);

static I8042_EXIST: Once<KernelFlag> = Once::new();
define_kernel_param!("i8042.exist", I8042_EXIST);
```
These declarations can appear in any component (e.g., `init`, `console`) and are automatically collected at link time. Each kernel parameter is represented as a global static variable, registered via a macro that binds:
- A name (e.g., `"console="`)
- A storage slot (where parsed values are stored)
- Metadata (e.g., whether it's early, implemented, takes a value)

Depending on semantics, storage slots differ:
- For single-assignment parameters: `spin::Once<T>`
- For repeating parameters (e.g., multiple `console=`): `Mutex<Vec<T>>`

Since Asterinas does not support loadable modules—and because our macro system already provides type-safe registration, there is no need for a separate mechanism to handle module-style parameters. The same powerful, type-driven abstraction can be used uniformly across all kernel configuration, whether for built-in subsystems or device-specific options like `i8042.exist`.

## Type Conversion
To ensure type safety and extensibility, all parameter types must implement the `FromKernelParam` trait, which governs how C-style string values are converted into Rust types.
```rust
/// A trait to convert a kernel parameter value to a Rust type.
pub trait FromKernelParam: Sized + 'static {
    /// Converts a kernel parameter value into a Rust type.
    ///
    /// # Input
    /// - `None`: the parameter was present without an explicit value
    ///   (flag-style), e.g. `debug`.
    /// - `Some(value)`: the parameter was provided with a value, e.g.
    ///   `log_level=info` where `value` is `"info"`.
    ///
    /// # Return value
    /// Implementations should return:
    /// - `Some(Self)` if the input is acceptable and can be constructed.
    /// - `None` if the input is invalid, or if a required value is missing.
    fn from_value(value: Option<&'static str>) -> Option<Self>;
}
```
For common types like `bool`, `u32`, `String`, etc., a blanket implementation is provided:
```rust
impl<T> FromKernelParam for T
where
    T: core::str::FromStr + 'static,
{
    fn from_value(value: Option<&'static str>) -> Option<Self> {
        let s = value?;
        s.trim().parse::<T>().ok()
    }
}
```
For more complex semantics, users can define custom types:
```rust
struct InitProcPath {
    pub path: &'static str,
}

impl FromKernelParam for InitProcPath {
    fn from_value(value: Option<&'static str>) -> Option<Self> {
        value.map(|path| Self { path })
    }
}

static INIT_PROC_PATH: Once<InitProcPath> = Once::new();
define_kernel_param!("init=", INIT_PROC_PATH);
```
For flag-style parameters (e.g., `debug`, `quiet`), we introduce a marker type `KernelFlag`:
```rust
/// A marker type for kernel parameters without value.
///
/// Recommends all kernel parameters without value to use this type as the slot
/// type.
pub struct KernelFlag;

impl FromKernelParam for KernelFlag {
    fn from_value(value: Option<&'static str>) -> Option<Self> {
        if value.is_none() { Some(Self) } else { None }
    }
}
```
## Uniform Registration
Since parameters are defined across different files and modules, they must be collected into a single array during linking. However, `spin::Once<T>` and `Mutex<Vec<T>>` have varying sizes and types.
To solve this, we define a low-level uniform structure `KernelParam` that abstracts away type-specific details:
```rust
pub struct KernelParam {
    name: &'static str,
    setup: fn(Option<&'static str>) -> (),
    early: bool,
    implemented: bool,
}

impl KernelParam {
    /// Creates a low-level kernel parameter.
    pub const fn new(
        name: &'static str,
        setup: fn(Option<&'static str>) -> (),
        early: bool,
        implemented: bool,
    ) -> Self {
        Self {
            name,
            setup,
            early,
            implemented,
        }
    }

    /// Calls the setup function.
    ///
    /// Invoke the callback which is the internal function defined in the
    /// `define_kernel_param!` and `define_kernel_param_vec!` macros. The
    /// callback will parse the kernel parameter value and store it in the
    /// corresponding slot, i.e. `spin::Once<T>` or `Mutex<Vec<T>>`.
    ///
    /// The `param` argument is the value of the kernel parameter, which is
    /// `None` for parameters without value and `Some(value)` for parameters
    /// with value.
    pub fn call_setup(&self, param: Option<&'static str>) {
        (self.setup)(param);
    }

    /// Gets the name of the kernel parameter.
    pub fn name(&self) -> &str {
        if self.name.ends_with('=') {
            &self.name[..self.name.len() - 1]
        } else {
            self.name
        }
    }

    /// Checks whether the kernel parameter has a value.
    pub fn has_value(&self) -> bool {
        self.name.ends_with('=')
    }

    /// Checks whether the kernel parameter is an early parameter.
    pub fn early(&self) -> bool {
        self.early
    }

    /// Checks whether the kernel parameter is implemented.
    pub fn implemented(&self) -> bool {
        self.implemented
    }
}
```
## Distributed Registration by Macros
We provide a set of declarative macros to simplify parameter definition and ensure correct registration. I.e. the helper macros for converting kernel parameters from high-level to low-level.

### Single-Value Parameters
```rust
#[macro_export]
macro_rules! define_kernel_param {
    ($name:literal, $slot:path, $early:expr) => {
        const _: () = {
            use $crate::boot::cmdline::FromKernelParam;

            fn __setup<T: FromKernelParam + 'static>(
                value: Option<&'static str>,
                slot: &'static spin::Once<T>,
            ) {
                if let Some(v) = T::from_value(value) {
                    let _ = slot.call_once(|| v);
                }
            }

            $crate::__define_kernel_param_common!($name, $early, __setup, $slot, true);
        };
    };
}
```
Handles parameters like `init=`, `log_level=`.

### Multi-Value Parameters
```rust
#[macro_export]
macro_rules! define_kernel_param_vec {
    ($name:literal, $slot:path, $early:expr) => {
        const _: () = {
            use $crate::boot::cmdline::FromKernelParam;

            fn __setup_vec<T: FromKernelParam + 'static>(
                value: Option<&'static str>,
                slot: &'static $crate::sync::Mutex<Vec<T>>,
            ) {
                if let Some(v) = T::from_value(value) {
                    slot.lock().push(v);
                }
            }

            $crate::__define_kernel_param_common!($name, $early, __setup_vec, $slot, true);
        };
    };
}
```
Used for repeatable parameters like `console=`.

### Unimplemented Parameters
```rust
#[macro_export]
macro_rules! define_kernel_param_unimpl {
    ($name:literal) => {
        const _: () = {
            fn __setup(_value: Option<&'static str>, _slot: &'static ()) {
                // Do nothing since this parameter is unimplemented.
            }

            $crate::__define_kernel_param_common!($name, false, __setup, (), false);
        };
    };
}
```
This ensures the parameter is defined (not passed to `init`) but marked as unimplemented (potential warning log).

### Internal Registration Primitive
```rust
#[macro_export]
#[doc(hidden)]
macro_rules! __define_kernel_param_common {
    ($name:literal, $early:expr, $setup_fn:ident, $slot:expr, $implemented:expr) => {
        use $crate::boot::cmdline::KernelParam;

        #[used]
        // SAFETY: This is properly handled in the linker script.
        #[unsafe(link_section = ".kernel_param")]
        static __KERNEL_PARAM: KernelParam = KernelParam::new(
            $name,
            |value| $setup_fn(value, &$slot),
            $early,
            $implemented,
        );
    };
}
```
The `#[used]` and `#[unsafe(link_section = ".kernel_param")]` attributes ensure the symbol is retained and placed in the designated section.

## Linker Integration
The linker script collects all registered parameters into a contiguous region:
```
.kernel_param           : AT(ADDR(.kernel_param) - KERNEL_VMA) {
    __kernel_param_start = .;
    KEEP(*(SORT(.kernel_param)))
    KEEP(*(.kernel_param .kernel_param.*))
    __kernel_param_end = .;
} : rodata
```
At runtime, the kernel iterates over the range `[__kernel_param_start, __kernel_param_end)` to build a lookup table or directly match parameters.

## Initialization Order
Because parameter parsing occurs in the `cmdline` component, all components that define kernel parameters must be initialized after `cmdline`.
To enforce this and avoid footguns, we provide wrapper macros in the `cmdline` component:
```rust
/// Convenience wrapper for single-value parameters.
#[macro_export]
macro_rules! kernel_param {
    ($name:literal, $slot:path) => {
        ostd::define_kernel_param!($name, $slot, false);
    };
}
 
/// Convenience wrapper for multi-value parameters.
#[macro_export]
macro_rules! kernel_param_vec {
    ($name:literal, $slot:path) => {
        ostd::define_kernel_param_vec!($name, $slot, false);
    };
}
```
By using these wrappers, developers express an explicit dependency on the `cmdline` subsystem, making initialization order clear and safe.

# Conclusion
The proposed design achieves all the design goals.

> 1. Adherence to the Open-Closed Principle

This design eliminates centralized coupling by enabling each component to declare its own parameters locally. There is no need to modify a global parsing function or shared data structure when adding new parameters.

> 2. Strong Typing via Rust's Type System

Unlike C-style string-based handling, our framework leverages Rust’s compile-time type system to enforce correctness at the definition site. Each parameter is bound to a concrete type (`bool`, `u32`, `String`, custom enums, etc.) via the `FromKernelParam` trait, ensuring that only valid values are accepted.
Moreover, support for both singular (`kernel_param!`) and repeated (`kernel_param_vec!`) parameters enables precise semantic modeling.

> 3. Initialization Order Control

By providing wrapper macros within the `cmdline` component, we establish a clear dependency hierarchy: any component using `kernel_param!` or `kernel_param_vec!` explicitly depends on the `cmdline` parser. This enforces correct initialization order at the architectural level.

> 4. Support for Module Parameters (Including Those with `.`)

Given that Asterinas does not support loadable kernel modules, maintaining a separate module parameter system introduces unnecessary complexity and abstraction overhead. Our design collapses this distinction, treating all parameters uniformly under a single, coherent model.

> 5. Compatibility with Linux: Support for "Unimplemented" Parameters

With `define_kernel_param_unimpl!`, we can declare the existence of parameters like `tsc=reliable`, `pci=lastbus=0`, preventing them from being forwarded to `init` while logging warnings for visibility.
This allows Asterinas to boot unmodified guest images from Kata, CoCo, and OVMF, even before full parameter support is implemented. It enables incremental development without sacrificing compatibility—a key enabler for production adoption.

# Misc

Previously, it was observed that OVMF injects the parameter `initrd=initrd` at the **end** of the `cmdline`—potentially after the `--` delimiter. According to Linux’s command-line parsing rules, everything after `--` is treated as an argument to `init`. This issue has been confirmed as a bug in OVMF, and was fixed in this [commit](https://github.com/tianocore/edk2/commit/a2ac0fea49996ab484c1a8761c234cc354f5a760). We can use an updated version of OVMF that includes the fix.

taosue:
Please see #2982 for details.

Closes #2982.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -30,6 +30,7 @@
         * [Namespaces, cgroups & security](kernel/linux-compatibility/syscall-flag-coverage/namespaces-cgroups-and-security/README.md)
         * [System information & misc](kernel/linux-compatibility/syscall-flag-coverage/system-information-and-misc/README.md)
     * [File System Coverage]()
+    * [Kernel Parameters](kernel/linux-compatibility/kernel-parameters.md)
 * [Roadmap](kernel/roadmap.md)
 
 # Asterinas OSTD
--- a/book/src/kernel/linux-compatibility/kernel-parameters.md
+++ b/book/src/kernel/linux-compatibility/kernel-parameters.md
@@ -1,0 +1,69 @@
+# Kernel Parameters
+
+This section documents kernel command-line parameters supported by Asterinas.
+
+## Inherited from Linux
+
+### `init`
+
+Run the specified binary as `init`.
+
+Example:
+```text
+init=/bin/busybox
+```
+
+Notes:
+- The value is the path to the executable.
+- If omitted, Asterinas will fail to boot.
+
+### `console`
+
+Select console devices for kernel messages.
+This parameter may be specified multiple times.
+Kernel messages are delivered to each listed console.
+
+Valid values:
+- `tty0`
+- `ttyS0`
+- `hvc0`
+
+Examples:
+```text
+console=ttyS0
+console=ttyS0 console=hvc0
+```
+
+## Asterinas-specific
+
+### `ostd.log_level`
+
+Set the verbosity level for Asterinas's logs.
+
+Valid values:
+- `off`
+- `error`
+- `warn`
+- `info`
+- `debug`
+- `trace`
+
+Example:
+```text
+ostd.log_level=error
+```
+
+### `i8042.exist`
+
+Override ACPI's indication of whether a PS/2 (i8042) controller exists.
+
+Valid values:
+- `1`, `on`, `yes`, `true` or no value — treat the i8042 controller as present (force probing)
+- `0`, `off`, `no`, `false` - treat the i8042 controller as absent (skip probing)
+
+Examples:
+```text
+i8042.exist
+i8042.exist=1
+i8042.exist=0
+```


kernel/comps/cmdline/src/lib.rs
```diff
@@ -16,65 +16,345 @@ use alloc::{
...
+#[macro_export]
+macro_rules! define_flag_param {
+    ($name:expr, $storage:expr) => {
+        $crate::__define_param!(@late, $name, $storage, $crate::setup_flag_param);
+    };
+}
+
+/// Defines an **early** `key=value` kernel command-line parameter.
+///
+/// Almost same as [`define_flag_param!`], but it is processed earlier in the boot sequence.
```
StevenJiang1110:
```suggestion
/// Almost same as [`define_flag_param`], but it is processed earlier in the boot sequence.
```

kernel/comps/cmdline/src/lib.rs
```diff
@@ -1,11 +1,11 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! The module to parse kernel command-line arguments.
+//! The module to define and parse kernel command-line arguments.
```
tatetian: The module --> A crate

kernel/comps/cmdline/src/types.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub fn expand_bounded(&self, max_elems: usize) -> Result<Vec<u32>, ParamError> {
+        if max_elems == 0 {
+            return Ok(Vec::new());
+        }
+
+        let mut out = Vec::new();
+
+        // Helper: push a cpu id; if we reached the bound, stop expansion.
+        let push_or_done = |out: &mut Vec<u32>, cpu: u32| -> bool {
+            if out.len() >= max_elems {
```
tatetian: [**`closure-fn-suffix`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/naming.html#closure-fn-suffix): This should be `push_or_done_fn`.

kernel/comps/cmdline/src/lib.rs
```diff
@@ -89,123 +369,115 @@ fn split_arg(input: &str) -> impl Iterator<Item = &str> {
...
+                    $crate::setup_unimplemented(occurrences, $name);
+                }
+                $crate::submit! {
+                    $crate::KernelParam { name: $name, setup: __kparam_setup, early: false }
+                }
+            };
+        )+
+    };
+}
+
```
tatetian: This macro is private (not `#[macro_export]`), so only the cmdline crate can define unimplemented params. The design doc in #2982 envisions components declaring their own unimplemented parameters locally (e.g., the i8042 component could mark `i8042.dumbkbd`, `i8042.nopnp` as unimplemented). Consider exporting this macro.

kernel/comps/cmdline/src/lib.rs
```diff
@@ -16,65 +16,345 @@ use alloc::{
...
+macro_rules! define_repeatable_kv_param {
+    ($name:expr, $storage:expr) => {
+        $crate::__define_param!(@late, $name, $storage, $crate::setup_repeatable_kv_param);
+    };
+}
+
+/// Defines an **early** `key=value` kernel command-line parameter.
+///
+/// Almost same as [`define_repeatable_kv_param!`], but it is processed earlier in the boot sequence.
+#[macro_export]
```
tatetian: Copy-paste error: the doc for `define_repeatable_kv_param_early!` says "Defines an **early** `key=value` kernel command-line parameter" which describes a plain kv param, not a repeatable one. Should be "Defines an **early repeatable** `key=value` kernel command-line parameter."

The same issue exists for `define_flag_param_early!` at line 134: it says "Defines an **early** `key=value` kernel command-line parameter" but should say "Defines an **early flag** kernel command-line parameter."

kernel/comps/cmdline/src/lib.rs
```diff
@@ -16,65 +16,345 @@ use alloc::{
...
+}
+
+/// Parse a flag parameter (e.g., "ro", "debug", "nokaslr").
+///
+/// Flags may optionally accept a value (e.g., "debug=1" in Linux).
+pub trait ParseFlag: Sized {
+    /// Called when the flag is present on the command line.
+    /// `value` is `Some(v)` if written as "flag=v", `None` if bare.
+    fn parse_flag(value: Option<&str>) -> Result<Self, ParamError>;
+}
```
tatetian: [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary): These trait doc comments use bare imperative "Parse" -- should be "Parses a single-value..." etc.

Same issue with the helper functions at lines 291, 312, 332 -- "Helper for single-value key-value parameters" is not a verb phrase. Consider "Sets up a single-value key-value parameter (last-wins)." or similar.

kernel/comps/cmdline/src/lib.rs
```diff
@@ -16,65 +16,345 @@ use alloc::{
...
+    /// Gets the argument vector (`argv`) of the init process.
+    pub fn argv(&self) -> &Vec<CString> {
+        &self.argv
+    }
+
+    /// Gets the environment vector (`envp`) of the init process.
+    pub fn envp(&self) -> &Vec<CString> {
+        &self.envp
+    }
+}
```
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): Returning `&Vec<CString>` exposes the storage type. Idiomatic Rust returns `&[CString]` from getters, which hides the fact that the backing store is a `Vec`. The callers in `init.rs` use `.to_vec()`, which works the same on slices.

Also, per [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary), "Gets" is acceptable but the standard Rust convention for getters is "Returns" (e.g., "Returns the argument vector...").

kernel/comps/cmdline/src/lib.rs
```diff
@@ -89,123 +369,115 @@ fn split_arg(input: &str) -> impl Iterator<Item = &str> {
...
+
+        let (key, value) = match arg.find('=') {
+            Some(pos) => (&arg[..pos], Some(&arg[pos + 1..])),
+            None => (arg, None),
         };
+        // Normalize hyphens to underscores (Linux compatibility)
+        let normalized = key.replace('-', "_");
+        // Group by normalized name, but keep the original key and value for forwarding.
+        grouped.entry(normalized).or_default().push((key, value));
+    }
```
tatetian: Grouping by normalized name before dispatching merges all unknown occurrences, forwarding only the last one to init. Linux processes unknown tokens individually: given `foo=1 foo foo=2`, Linux forwards both `foo=1` and `foo=2` to init's env and `foo` to argv. This code only forwards `foo=2`.

This deviates from both Linux and the previous Asterinas behavior. The boot test documents this as intentional, but it could silently drop values that callers depend on.

One possible fix: look up each token individually against the registry (normalizing for lookup only), and only group recognized params for their setup functions. Unknown tokens would be forwarded one-by-one in command-line order, matching Linux behavior.

kernel/comps/cmdline/Cargo.toml
```diff
@@ -10,6 +10,7 @@ component.workspace = true
 log.workspace = true
 ostd.workspace = true
 spin.workspace = true
+inventory = { git = "https://github.com/asterinas/inventory", rev = "9dce587" }
```
tatetian: [**`workspace-deps`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#workspace-deps): The `inventory` crate is already used by `kernel/libs/comp-sys/component/Cargo.toml` with the same URL and rev, but neither goes through the workspace table. This PR should add `inventory` to the workspace `[workspace.dependencies]` and use `inventory.workspace = true` in both crates.

kernel/comps/cmdline/src/types.rs
```diff
@@ -0,0 +1,432 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Common parameter value types for the cmdline framework.
+//!
+//! This module provides Linux-style parsers that are frequently used by kernel
+//! command lines so users of this framework don't need to rewrite them.
+
+#![allow(dead_code)]
+
```
tatetian: [**`narrow-lint-suppression`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/macros-and-attributes.html#narrow-lint-suppression) + [**`expect-dead-code`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/macros-and-attributes.html#expect-dead-code): A module-wide `#![allow(dead_code)]` is the broadest possible suppression. Apply `#[expect(dead_code)]` on each individual struct instead.

kernel/comps/cmdline/src/types.rs
```diff
@@ -0,0 +1,432 @@
...
+pub struct CpuListSegment {
+    pub start: u32,
+    pub end: u32,
+    /// Step within range, default 1.
+    pub stride: u32,
+    /// Optional group size (the `/N` part). When present, stride selection is
+    /// applied within each group window.
+    pub group: Option<u32>,
+}
+
```
tatetian: [**`getter-encapsulation`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/types-and-traits.html#getter-encapsulation): All fields of `CpuListSegment` are `pub`. Consider making them private with getters, consistent with how `CpuList` already keeps `segments` private.

kernel/comps/cmdline/src/types.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub fn segments(&self) -> &[CpuListSegment] {
+        &self.segments
+    }
+
+    pub fn contains(&self, cpu: u32) -> bool {
+        self.segments.iter().any(|s| segment_contains(s, cpu))
+    }
+
+    /// Expand to concrete CPU IDs, returning at most `max_elems` elements.
+    ///
```
tatetian: [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary): Should be "Expands to concrete CPU IDs..." (third-person singular present).

kernel/comps/cmdline/src/lib.rs
```diff
@@ -1,11 +1,11 @@
...
 
-//! The module to parse kernel command-line arguments.
+//! The module to define and parse kernel command-line arguments.
 //!
 //! The format of the Asterinas command line string conforms
 //! to the Linux kernel command line rules:
 //!
-//! <https://www.kernel.org/doc/html/v6.4/admin-guide/kernel-parameters.html>
+//! <https://docs.kernel.org/admin-guide/kernel-parameters.html>
 //!
```
tatetian, line 13: [**`module-docs`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#module-docs): This crate was completely rewritten from a centralized parser into a distributed registration framework, but the crate-level doc is still a single sentence carried over from the old code.

The doc should describe the macro-based registration model (`define_kv_param!`, `define_repeatable_kv_param!`, `define_flag_param!`), the key traits consumers implement (`ParseParamValue`, `ParseRepeatableParamValue`, `ParseFlag`, `ParamStorage`), how parameters are collected via `inventory` and dispatched during boot, and the relationship to the component initialization system (components depending on `aster-cmdline` get their parameters parsed before they initialize). A short usage example showing how a component registers a parameter would also help newcomers.

kernel/comps/cmdline/src/lib.rs
```diff
@@ -16,65 +16,345 @@ use alloc::{
...
     }
+}
+
+#[derive(Debug)]
+pub struct KernelParam {
+    pub name: &'static str,
+    pub setup: fn(occurrences: &[(&str, Option<&str>)]),
+    pub early: bool,
+}
+
```
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): `KernelParam` is purely an implementation detail of the registration macros -- end users never interact with it directly. It should be marked `#[doc(hidden)]` to keep it out of the public rustdoc surface. The same applies to the helper functions it depends on (`setup_kv_param`, `setup_repeatable_kv_param`, `setup_flag_param`, `setup_unimplemented`) and the re-exported `inventory::submit`, which are all macro internals.

[**`getter-encapsulation`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/types-and-traits.html#getter-encapsulation): All three fields are `pub`. Since the only reason is macro construction, consider providing a `const fn new(name, setup_fn, early)` constructor (as shown in the #2982 design) and making the fields private.

[**`closure-fn-suffix`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/naming.html#closure-fn-suffix): The `setup` field holds a function pointer, so it should end with `_fn` (e.g., `setup_fn`).

**Hyphen normalization pitfall**: The `name` field stores the registered name verbatim, but `dispatch_params` normalizes hyphens to underscores for lookup. If someone registers a param with hyphens (e.g., `"no-timer-check"`), the lookup would search for `"no_timer_check"` and fail silently. Current registrations all use underscores, so this works today. A `debug_assert!(!name.contains('-'))` in the constructor or a doc comment on the naming requirement would prevent future mistakes.

kernel/comps/cmdline/src/lib.rs
tatetian, line 133: Copy-paste error: the doc for `define_repeatable_kv_param_early!` says "Defines an **early** `key=value` kernel command-line parameter" which describes a plain kv param, not a repeatable one. Should be "Defines an **early repeatable** `key=value` kernel command-line parameter."

The same issue exists for `define_flag_param_early!` at line 134: it says "Defines an **early** `key=value` kernel command-line parameter" but should say "Defines an **early flag** kernel command-line parameter."

kernel/comps/cmdline/src/lib.rs
tatetian: [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary): These trait doc comments use bare imperative "Parse" -- should be "Parses a single-value..." etc.

Same issue with the helper functions at lines 291, 312, 332 -- "Helper for single-value key-value parameters" is not a verb phrase. Consider "Sets up a single-value key-value parameter (last-wins)." or similar.

kernel/comps/cmdline/src/lib.rs
tatetian: [**`hide-impl-details`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-impl-details): Returning `&Vec<CString>` exposes the storage type. Idiomatic Rust returns `&[CString]` from getters, which hides the fact that the backing store is a `Vec`. The callers in `init.rs` use `.to_vec()`, which works the same on slices.

Also, per [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary), "Gets" is acceptable but the standard Rust convention for getters is "Returns" (e.g., "Returns the argument vector...").

kernel/comps/cmdline/src/lib.rs
tatetian: Grouping by normalized name before dispatching merges all unknown occurrences, forwarding only the last one to init. Linux processes unknown tokens individually: given `foo=1 foo foo=2`, Linux forwards both `foo=1` and `foo=2` to init's env and `foo` to argv. This code only forwards `foo=2`.

This deviates from both Linux and the previous Asterinas behavior. The boot test documents this as intentional, but it could silently drop values that callers depend on.

One possible fix: look up each token individually against the registry (normalizing for lookup only), and only group recognized params for their setup functions. Unknown tokens would be forwarded one-by-one in command-line order, matching Linux behavior.

kernel/comps/cmdline/Cargo.toml
tatetian, line 14: [**`workspace-deps`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#workspace-deps): The `inventory` crate is already used by `kernel/libs/comp-sys/component/Cargo.toml` with the same URL and rev, but neither goes through the workspace table. This PR should add `inventory` to the workspace `[workspace.dependencies]` and use `inventory.workspace = true` in both crates.

kernel/comps/cmdline/src/types.rs
tatetian: [**`narrow-lint-suppression`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/macros-and-attributes.html#narrow-lint-suppression) + [**`expect-dead-code`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/macros-and-attributes.html#expect-dead-code): A module-wide `#![allow(dead_code)]` is the broadest possible suppression. Apply `#[expect(dead_code)]` on each individual struct instead.

kernel/comps/cmdline/src/types.rs
tatetian, line 35: [**`getter-encapsulation`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/types-and-traits.html#getter-encapsulation): All fields of `CpuListSegment` are `pub`. Consider making them private with getters, consistent with how `CpuList` already keeps `segments` private.

kernel/comps/cmdline/src/types.rs
tatetian, line 46: [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary): Should be "Expands to concrete CPU IDs..." (third-person singular present).

kernel/comps/cmdline/src/types.rs
tatetian, line 57: [**`closure-fn-suffix`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/naming.html#closure-fn-suffix): This should be `push_or_done_fn`.

kernel/comps/cmdline/src/lib.rs
tatetian: This macro is private (not `#[macro_export]`), so only the cmdline crate can define unimplemented params. The design doc in #2982 envisions components declaring their own unimplemented parameters locally (e.g., the i8042 component could mark `i8042.dumbkbd`, `i8042.nopnp` as unimplemented). Consider exporting this macro.

kernel/comps/cmdline/src/lib.rs
tatetian, line 1: [**`one-concept-per-file`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#one-concept-per-file): This 484-line file contains several distinct concepts interleaved:

1. **Public macro API** (lines 32-168): `define_kv_param!`, `define_flag_param!`, etc. -- the user-facing surface.
2. **Trait definitions and blanket impls** (lines 189-280): `ParamStorage`, `ParseParamValue`, `ParseRepeatableParamValue`, `ParseFlag`, `ParamError`, plus `FromStr`/`bool` blanket impls -- the type system for parameter parsing.
3. **Registration and setup helpers** (lines 282-347): `KernelParam`, `inventory::collect!`, `setup_kv_param`, `setup_repeatable_kv_param`, `setup_flag_param` -- internal machinery bridging macros to dispatch.
4. **Dispatch and init forwarding** (lines 349-447): `InitprocArgs`, `INIT_PROC_ARGS`, `split_arg`, `dispatch_params`, `init()` -- the actual command-line parsing engine.
5. **Unimplemented parameter stubs** (lines 449-483): `define_unimplemented_param!`, `setup_unimplemented`, and the stub list.

A possible decomposition:

| File | Contents |
|---|---|
| `lib.rs` | Crate-level docs, re-exports, and the public macros |
| `parse.rs` | `ParamStorage`, `ParseParamValue`, `ParseRepeatableParamValue`, `ParseFlag`, `ParamError`, and their blanket impls |
| `dispatch.rs` | `InitprocArgs`, `INIT_PROC_ARGS`, `KernelParam`, `split_arg`, `dispatch_params`, `init()` |
| `unimplemented.rs` | `define_unimplemented_param!`, `setup_unimplemented`, and the stub list |
| `types.rs` | (already exists) `CpuList`, `MetricU64` |

This also helps with [**`top-down-reading`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#top-down-reading): `lib.rs` would present the public API first, with implementation details pushed into sub-modules.

test/initramfs/src/boot_hello.sh
tatetian, line 1: Two concerns with how the test is structured:

**Tight coupling between `Makefile` and `boot_hello.sh`**: The Makefile injects specific magic strings via `--kcmd-args` and `boot_hello.sh` hardcodes assertions against those exact strings. Changing one requires updating the other, with no obvious link between the two files. This mixes mechanism (how to pass cmdline args) with policy (what forwarding behavior to verify).

**`boot_hello.sh` name no longer reflects its scope**: The script was originally a minimal boot smoke test ("did the system boot?"). It now also tests cmdline parameter forwarding semantics, which is a separate concern.

[**`test-visible-behavior`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/testing-guidelines.html#test-visible-behavior): The forwarding logic lives in `dispatch_params`, which is a pure function taking a `&str` and returning `InitprocArgs`. This is directly unit-testable. Consider:

1. **Add `#[ktest]` tests for `dispatch_params` in the cmdline crate**, testing the forwarding/ignoring logic at the right abstraction level. For example:

    ```rust
    #[ktest]
    fn unknown_kv_forwarded_to_init_env() {
        let args = dispatch_params("unknown_key=1");
        assert!(args.envp().iter().any(|e| e.to_bytes() == b"unknown_key=1"));
    }

    #[ktest]
    fn dotted_unknown_param_not_forwarded() {
        let args = dispatch_params("some_module.flag");
        assert!(args.argv().is_empty());
        assert!(args.envp().is_empty());
    }
    ```

    This eliminates the Makefile/shell coupling and follows the pattern already established by the `types.rs` tests.

2. **Keep `boot_hello.sh` as a minimal boot smoke test.** If an end-to-end integration test is still desired, put it in a dedicated script (e.g., `test_cmdline.sh`) with its own `AUTO_TEST` target, so the coupling is at least contained within one test concern.

kernel/comps/cmdline/src/dispatch.rs
```diff
@@ -0,0 +1,216 @@
...
+        envp: Vec::new(),
+    };
+
+    let mut kcmdline_end = false;
+
+    // Step 1: Build lookup from registered param name to handler.
+    let registry: BTreeMap<&str, &KernelParam> = inventory::iter::<KernelParam>
+        .into_iter()
+        .map(|p| (p.name, p))
+        .collect();
```
tatetian: **Duplicate parameter names silently dropped**: If two components accidentally register the same parameter name (e.g., two crates both call `define_kv_param!("debug", ...)`), the `BTreeMap` silently keeps only one entry. The other handler is lost with no warning, which could be very difficult to diagnose. Consider adding a duplicate-detection pass that logs a warning (or panics in debug builds) when a name collision is detected, e.g.:

```rust
let mut registry = BTreeMap::new();
for p in inventory::iter::<KernelParam> {
    if let Some(prev) = registry.insert(p.name, p) {
        log::warn!(
            "duplicate kernel parameter '{}' registered; keeping last",
            prev.name
        );
    }
}
```

kernel/comps/cmdline/src/dispatch.rs
```diff
@@ -0,0 +1,216 @@
...
+        static TEST_VARIABLE: Once<u32> = Once::new();
+        static TEST_VARIABLE_REPEATED: Once<Vec<String>> = Once::new();
+
+        crate::define_kv_param!("log_level", TEST_VARIABLE);
+        crate::define_repeatable_kv_param!("console", TEST_VARIABLE_REPEATED);
+
+        let args = dispatch_params("log_level=4 console=ttyS0 console=ttyS1");
+        assert!(args.argv().is_empty());
+        assert!(args.envp().is_empty());
+    }
```
tatetian, line 220: **Test param names may collide with production registrations**: The test registers `"log_level"` and `"console"`, but `"console"` is already registered in production code (`kernel/src/device/tty/device.rs`). Since `inventory` collects *all* submissions at link time, the registry will contain duplicate entries for `"console"`. The `BTreeMap::collect()` silently keeps one, making the test depend on which handler wins. Consider using test-specific names (e.g., `"__test_kv"`, `"__test_repeatable"`) to avoid collisions with real parameters.

kernel/comps/cmdline/src/dispatch.rs
```diff
@@ -0,0 +1,216 @@
...
+            }
+            i += 1;
+        }
+        false
+    }
+}
+
+inventory::collect!(KernelParam);
+
+pub static INIT_PROC_ARGS: Once<InitprocArgs> = Once::new();
```
tatetian, line 79: [**`narrow-visibility`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#narrow-visibility): Do not disclose the entire `dispatch` module. From its name, one can see that it is an implementation detail. Re-export select items from the `dispatch` module as the crate level.

Similarly, I don't think it is wise to diclose the `unimplemented` module to the users of this crate. Simply re-export the `define_unimplemented_param` macro at the crate level would do the job.

kernel/comps/cmdline/src/types.rs
```diff
@@ -0,0 +1,450 @@
...
+}
+
+fn parse_u32(s: &str) -> Result<u32, ParamError> {
+    if s.is_empty() {
+        return Err(ParamError::InvalidValue);
+    }
+    s.parse::<u32>().map_err(|_| ParamError::InvalidValue)
+}
+
+impl ParseParamValue for MetricU64 {
```
tatetian: [**`top-down-reading`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/README.html#top-down-reading): The `MetricU64` struct definition (line 47) is separated from its `impl ParseParamValue for MetricU64` block (line 270) by over 220 lines of `CpuList`-related code. Consider moving the `MetricU64` definition closer to its implementation -- either right above its `ParseParamValue` impl, or grouping all `MetricU64` code together after all `CpuList` code. This makes it easier to read the type and its behavior in one place.

kernel/comps/logger/src/console.rs
```diff
@@ -25,9 +25,13 @@ pub fn _print(args: fmt::Arguments) {
     );
     impl Write for Printer<'_> {
         fn write_str(&mut self, s: &str) -> fmt::Result {
-            self.0
-                .values()
-                .for_each(|console| console.send(s.as_bytes()));
+            if self.0.is_empty() {
+                ostd::early_print!("{}", s);
```
tatetian: **Separate concern**: This `early_print!` fallback addresses a real gap (log messages before any console device is registered), but it is orthogonal to the kernel parameter framework. Mixing it into this PR makes the diff harder to review. Consider splitting it into a separate commit with its own rationale so reviewers can evaluate each change independently.

PR #3011, opened 2026-03-05
Bump the version to 0.17.1

junyang-zh:
Goes after #3008


tatetian:
@cqs21 Could you figure out if the timeout failure of "Test AsterNixOS (full)" is expected and what is the right fix (e.g., increasing the timeout)?

PR #3012, opened 2026-03-06
Respect `mmap` address hints

lrh2000:
Some applications call `mmap` with an address hint but without MAP_FIXED. They simply fail if the kernel does not place the mapping at the specified address.

```
mmap(0x200000000, 4297064448, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x200000000
```

The userspace is a closed-source, proprietary blob that cannot be fixed. But we can fix the kernel. The [man pages](https://man7.org/linux/man-pages/man2/mmap.2.html) documented:
> If addr is not NULL, then the kernel takes it as a hint about where to place the mapping; on Linux, the kernel will pick a nearby page boundary (but always above or equal to the value specified by /proc/sys/vm/mmap_min_addr) and attempt to create the mapping there.  If another mapping already exists there, the kernel picks a new address that may or may not depend on the hint.

### DIFF
--- a/test/initramfs/src/apps/security/capability/capset.c
+++ b/test/initramfs/src/apps/security/capability/capset.c
@@ -19,12 +19,13 @@ FN_TEST(capget)
 {
 	TEST_SUCC(syscall(SYS_capget, &caphdr, capdat));
 
-	TEST_RES(0, (capdat[0].effective |
-		     (((uint64_t)capdat[1].effective) << 32)) == CAPS_ALL);
-	TEST_RES(0, (capdat[0].permitted |
-		     (((uint64_t)capdat[1].permitted) << 32)) == CAPS_ALL);
-	TEST_RES(0, (capdat[0].inheritable |
-		     (((uint64_t)capdat[1].inheritable) << 32)) == CAPS_NONE);
+	TEST_RES(capdat[0].effective | (((uint64_t)capdat[1].effective) << 32),
+		 _ret == CAPS_ALL);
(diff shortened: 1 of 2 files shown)


junyang-zh:
This is also done in #2615 . Ok if merge it first.

lrh2000:
> This is also done in #2615 . Ok if merge it first.

Okay. Thanks for the information. I suppose adding the new functionality separately would be better (although I wasn't aware of that when I wrote the code).

kernel/src/vm/vmar/vmar_impls/map.rs
cchanging: The usage documents should be also modified.

lrh2000: Emm.. This example looks broken because there is currently no public way to create a VMAR because the constructor of `ProcessVm` is not public...
https://github.com/asterinas/asterinas/blob/670ce782c2732b10da6d7d2e060f93a4fd6b29b0/kernel/src/vm/vmar/vmar_impls/mod.rs#L47-L48
https://github.com/asterinas/asterinas/blob/670ce782c2732b10da6d7d2e060f93a4fd6b29b0/kernel/src/process/process_vm/mod.rs#L76-L78
            
[In the last version, I tried fixing the code using `current_userspace`.](https://github.com/asterinas/asterinas/compare/eabc514fb4473c243dd8a18d4be80f42f8651429..2c57975808328d419cd899e3a5cad7410a0e136d) However, this is not a perfect solution. `current_userspace!()` will panic in the ktest environment because there is no userspace for the kernel thread. Unfortunately, I cannot figure out a better way.

PR #3013, opened 2026-03-06
fs/path: fix `EISDIR`/`EINVAL` errno precedence when opening directories

zzj-5341:
Note: This PR depends on #2996.

Problem:

When opening an existing directory, Asterinas returned incorrect error
codes because type-related checks were missing in `path.open()`:
--`O_CREAT | O_DIRECTORY` on an existing directory returned `EACCES` or `EEXIST` instead of `EINVAL`
--`O_CREAT` on an existing directory returned `EACCES` or `EEXIST` instead of `EISDIR`
--Write mode (`O_WRONLY`/`O_RDWR`) on a directory returned `EACCES` instead of `EISDIR`

Root Cause

Linux checks type errors (`EISDIR`, `EINVAL`) before permission errors
(`EACCES`). `path.open()` had no such type checks, so permission checks
fired first and masked the correct error code.

Fix

Add three early checks at the top of `path.open()`, before any
permission checks, matching Linux errno precedence:
1. `O_CREAT | O_DIRECTORY` → `EINVAL`
2. `O_CREAT` + target is directory → `EISDIR`
3. Write mode + target is directory → `EISDIR` (excluding `O_PATH`)

Testing

Five regression tests added in `test/initramfs/src/apps/fs/open_dir/`.
Linux syscall compatibility improved from **68.05% → 90.54%** over
~2.8 million `open`/`openat` test cases.

### DIFF
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -115,6 +115,12 @@ impl Path {
         {
             return_errno_with_message!(Errno::EEXIST, "the file already exists");
         }
+        if creation_flags.contains(CreationFlags::O_CREAT) && inode_type == InodeType::Dir {
+            return_errno_with_message!(
+                Errno::EISDIR,
+                "O_CREAT is specified but the file is a directory"
+            );
+        }
 
         if inode_type == InodeType::SymLink
             && creation_flags.contains(CreationFlags::O_NOFOLLOW)
--- a/test/initramfs/src/regression/fs/ext2/open_dir.c
+++ b/test/initramfs/src/regression/fs/ext2/open_dir.c
@@ -29,15 +29,55 @@ FN_TEST(open_creat_directory_on_existing_dir_returns_einval)
 }
 END_TEST()
 
+FN_TEST(open_creat_excl_directory_on_existing_dir_returns_einval)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_EXCL | O_DIRECTORY | O_WRONLY,
+			0644),
+		   EINVAL);
+}
+END_TEST()
+
+FN_TEST(open_creat_excl_directory_rdwr_on_existing_dir_returns_einval)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_EXCL | O_DIRECTORY | O_RDWR,
+			0644),
+		   EINVAL);
+}
+END_TEST()
+
+FN_TEST(open_creat_read_only_on_existing_dir_returns_eisdir)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_RDONLY, 0644), EISDIR);
+}
+END_TEST()
+
 FN_TEST(open_creat_on_existing_dir_returns_eisdir)
 {
 	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_WRONLY, 0644), EISDIR);
 }
 END_TEST()
 
+FN_TEST(open_creat_rdwr_on_existing_dir_returns_eisdir)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_RDWR, 0644), EISDIR);
+}
+END_TEST()
+
+FN_TEST(open_creat_excl_read_only_on_existing_dir_returns_eexist)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_EXCL | O_RDONLY, 0644), EEXIST);
+}
+END_TEST()
+
 FN_TEST(open_creat_excl_on_existing_dir_returns_eexist)
 {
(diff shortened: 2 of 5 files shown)


tatetian:
@zzj-5341 Thanks for the contribution! I set this PR to draft. After passing all CI tests, you can mark this PR as Ready for Review.

zzj-5341:
This PR includes the changes from #2996 (fix symlink resolution with trailing slash) in addition to the new errno precedence fixes. If this PR is merged, #2996 can be closed as it is already covered here.

tatetian:
> This PR includes the changes from #2996 (fix symlink resolution with trailing slash) in addition to the new errno precedence fixes. If this PR is merged, #2996 can be closed as it is already covered here.

When PR A depends on PR B, we still review and merge the two PRs separately. See the coding guideline [`focused-prs`](https://asterinas.github.io/book/to-contribute/coding-guidelines/git-guidelines.html#focused-prs). So you should re-open #2996. I am sorry for the trouble but this is how things work.

tatetian:
A PR may depend on a base PR (#2996), but the dependency must be based on the commits of the base PR. This PR directly contains the changes introduced by #2996. This is undesirable because this increases the difficulties of code review and causes conflicts after the base PR is merged. This PR needs to be split into atomic commits.

boterinas:
Some changes occurred in .github/ISSUE_TEMPLATE

cc @grief8

test/initramfs/src/apps/fs/open_dir/open_dir.c
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MPL-2.0
```
cchanging: Insert a blank line here, and I think this file can be put in `fs/ext2`.

test/initramfs/src/regression/fs/ext2/open_dir.c
```diff
@@ -0,0 +1,59 @@
...
+FN_TEST(open_creat_directory_on_existing_dir_returns_einval)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_DIRECTORY | O_RDONLY, 0644),
+		   EINVAL);
+}
+END_TEST()
+
+FN_TEST(open_creat_on_existing_dir_returns_eisdir)
+{
+	TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_WRONLY, 0644), EISDIR);
```
cchanging, line 56: Given that we now have many logics for returning errno, can you test some more complex permutations? e.g.

```c
TEST_ERRNO(open(TARGET_DIR, O_CREAT | O_EXCL | O_WRONLY, 0644), ???)
```

I think if possible, we should try to make the order of returned errno consistent with Linux.

cchanging (commented):
Thanks for adding some checks for the `open` operation!

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -110,6 +110,26 @@ impl Path {
         let creation_flags = &open_args.creation_flags;
         let status_flags = &open_args.status_flags;
 
+        if creation_flags.contains(CreationFlags::O_CREAT)
+            && creation_flags.contains(CreationFlags::O_DIRECTORY)
```
tatetian, line 114: This check should be moved into [`do_open`](https://github.com/asterinas/asterinas/blob/4872e5dc5aea959f94b16adb7fb383efa4279d20/kernel/src/syscall/open.rs#L77). This check has nothing to do with `self`. Checking it as early as possible can avoid some unnecessary work.

In fact, the same check already exists in `do_open`:

https://github.com/asterinas/asterinas/blob/4872e5dc5aea959f94b16adb7fb383efa4279d20/kernel/src/syscall/open.rs#L99-L107

In fact, since the semantics of `O_CREAT` and `O_DIRECTORY` is conflicting. You should move the check into `OpenArgs::from_flags_and_mode`.

tatetian (commented):
Thanks for the bug fixes.

test/initramfs/src/regression/fs/ext2/open_dir.c
zzj-5341: Updated. I added more flag permutations beyond the example case and checked them against Linux behavior.

In particular, the regression coverage now includes:
- `O_CREAT | O_EXCL | O_DIRECTORY | ...` -> `EINVAL`
- `O_CREAT | ...` on an existing directory -> `EISDIR`
- `O_CREAT | O_EXCL | ...` on an existing directory -> `EEXIST`

I also updated the open-path handling so the errno precedence is closer to Linux.

kernel/src/fs/vfs/path/mod.rs
zzj-5341: Updated. I moved the `O_CREAT | O_DIRECTORY` validation into `OpenArgs::from_flags_and_mode()` and removed the duplicated check from `do_open()`.

Since this is a conflict in the open arguments themselves rather than a property of a resolved path, validating it at argument-construction time seems more appropriate.

tatetian (approved):

Issue #3015, opened 2026-03-06
Inconsistent Dentry Cache State after Failed `rmdir` Operations


li041:
### Describe the bug

In the current implementation of `DirDentry::rmdir`, the child dentry is removed from the `children` cache before the actual filesystem operation is performed on the underlying inode.

If the inode operation (e.g., `dir_inode.rmdir(name)`) fails—for instance, when attempting to remove a non-empty directory—the dentry is already gone from the cache. This leads to an inconsistent state where the file or directory still exists on the disk, but the kernel's dentry cache no longer tracks it correctly, treats it as a negative entry. Subsequent lookups may fail.

### To Reproduce

1. Create a non-empty directory: `mkdir test_dir && touch test_dir/a.txt`
2. Attempt to remove the directory: `rmdir test_dir` (This will fail with `Directory not empty`).
3. Try to list the directory or access the file inside: `ls test_dir` or `ls test_dir/a.txt`.
4. The system may report that the file/directory does not exist because the dentry cache was cleared despite the operation failing.

### Expected behavior

If rmdir or unlink fails, the dentry cache should remain unchanged so that the file system state and the dentry cache remain synchronized.

### Screenshots

<img width="333" height="125" alt="Image" src="https://github.com/user-attachments/assets/a76f8218-4543-4ec0-9948-74b78cb2e35e" />

>

tatetian:
Sharp eye!

PR #3016, opened 2026-03-06
Fix dentry cache inconsistency on failed `rmdir`

li041:
Closes https://github.com/asterinas/asterinas/issues/3015

Currently, the DirDentry implementation removes a child entry from the dentry cache (via children.delete(name)) before attempting the actual filesystem operation (unlink or rmdir). If the underlying filesystem operation fails (for example, attempting to rmdir a non-empty directory), the child remains deleted from the cache.

This results in a cache inconsistency where subsequent lookups for the existing file/directory might result in a "Not Found" error or hit a negative dentry, even though the inode still exists on the storage.

This PR ensures that the children dentry remains populated if the rmdir (specifically for non-empty directories) or unlink returns an error.

### DIFF
--- a/kernel/src/fs/vfs/path/dentry.rs
+++ b/kernel/src/fs/vfs/path/dentry.rs
@@ -358,22 +358,24 @@ impl DirDentry<'_> {
         children.check_mountpoint(name)?;
 
         let mut children = children.upgrade();
-        let cached_child = children.delete(name);
-
         let dir_inode = &self.inode;
-        let child_inode = match cached_child {
-            Some(child) => {
-                // Cache hit: use the cached dentry
-                child.inode().clone()
-            }
-            None => {
-                // Cache miss: need to lookup from the underlying filesystem
-                drop(children);
-                dir_inode.lookup(name)?
-            }
+
+        let child_inode = if let Some(cached_dentry) = children.entry(name)? {
+            let child_inode = cached_dentry.inode().clone();
+
+            dir_inode.unlink(name)?;
+            children.delete(name);
+
+            child_inode
+        } else {
+            // Cache miss: need to lookup from the underlying filesystem
+            let child_inode = dir_inode.lookup(name)?;
+            dir_inode.unlink(name)?;
+
+            child_inode
         };
 
-        dir_inode.unlink(name)?;
+        drop(children);
 
         let nlinks = child_inode.metadata().nr_hard_links;
         fs::vfs::notify::on_link_count(&child_inode);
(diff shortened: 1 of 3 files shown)


kernel/src/fs/path/dentry.rs
```diff
@@ -357,19 +357,24 @@ impl DirDentry<'_> {
...
-            None => {
-                // Cache miss: need to lookup from the underlying filesystem
-                drop(children);
-                dir_inode.lookup(name)?
+        let child_inode = if let Some(child) = cached_child {
+            let child_inode = child.inode().clone();
+
+            if let Err(err) = dir_inode.unlink(name) {
+                children.insert(name.to_string(), child);
+                return Err(err);
```
tatetian: Consider adding an `entry`-style API to `DentryChildren` that would let you probe for the child, obtain a reference, and only remove it from the table if the FS operation succeeds. This would avoid the inefficient delete-then-reinsert pattern entirely, simplifies the logic (just delete, not re-inset), and eliminate a redundant allocation (`name.to_string()`).

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -357,19 +357,24 @@ impl DirDentry<'_> {
...
+            child_inode
+        } else {
+            // Cache miss: need to lookup from the underlying filesystem
+            drop(children);
+            let child_inode = dir_inode.lookup(name)?;
+
+            dir_inode.unlink(name)?;
+
+            child_inode
+        };
```
tatetian, line 377: Pre-existing issue (not introduced by this PR, but worth noting since the surrounding code is being modified): the cache-miss path drops the write lock before performing `lookup` + `unlink`/`rmdir` as two separate filesystem operations. This creates a TOCTOU window where a concurrent `rename` can change what the name refers to between the two calls.

**Example — two threads, same directory, `rename` racing with `unlink`:**

| Step | Thread A: `unlink("foo")` (cache miss) | Thread B: `rename("bar", "foo")` |
|------|----------------------------------------|----------------------------------|
| 1 | Acquires write lock | |
| 2 | `delete("foo")` → `None` (miss) | |
| 3 | **Drops write lock** | |
| 4 | `lookup("foo")` → inode X | |
| 5 | | Acquires write lock |
| 6 | | `rename("bar", "foo")` — atomically replaces inode X with inode Y under "foo" |
| 7 | | Releases write lock |
| 8 | `unlink("foo")` → **removes inode Y** | |

Thread A intended to unlink inode X but ended up removing inode Y (the file that was just renamed into place). The returned `child_inode` is still X, so all subsequent notifications reference the wrong inode.

This is a pre-existing design issue in the cache-miss path. The fix would be to keep the write lock held across both `lookup` and `unlink`/`rmdir`.

tatetian:
Good job. Thanks for identifying and fixing this bug.

This PR  does not include regression tests. Per the project's [coding guidelines](https://github.com/asterinas/asterinas/blob/main/book/src/to-contribute/coding-guidelines/):

> **Add regression tests for every bug fix**.

Please add a test that:
1. Creates a directory with a child entry.
2. Attempts `rmdir` on the non-empty directory (expecting `ENOTEMPTY`).
3. Verifies that the directory is still accessible via lookup after the failed `rmdir`.

Similarly, a test for the `unlink` path where the FS operation fails (although such a failure does not seem to be likely for a local FS) would strengthen confidence in the fix.

PR #3017, opened 2026-03-06
Add JDK test on Asterinas NixOS

cqs21:
This PR adds support for running JDK tests on Asterinas NixOS using the `jtreg` test framework. The implementation includes:

1. **Add jtreg package**: A complete nix overlay that builds jtreg and its dependencies (jtharness, asmtools, junit, testng) from source
2. **Add JDK testcases**: Initial set of JDK testcases selected for Asterinas NixOS

### DIFF
--- a/.github/workflows/test_nixos_full.yml
+++ b/.github/workflows/test_nixos_full.yml
@@ -103,6 +103,7 @@ jobs:
               { "name": "development-tools", "disk_size": 10240, "timeout_minutes": 45 },
               { "name": "development-tools-go", "disk_size": 8192, "timeout_minutes": 60 },
               { "name": "development-tools-python", "disk_size": 8192, "timeout_minutes": 60 },
+              { "name": "development-tools-openjdk", "disk_size": 8192, "timeout_minutes": 60 },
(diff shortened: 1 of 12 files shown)


distro/etc_nixos/overlays/jtreg/0001-Fix-tool-paths-and-build-flags.patch
```diff
@@ -0,0 +1,149 @@
+From 017b5d4f857e68dda1bf5594ffcac9d14b82c678 Mon Sep 17 00:00:00 2001
+From: Qingsong Chen <changxian.cqs@antgroup.com>
+Date: Mon, 2 Mar 2026 03:26:07 +0000
+Subject: [PATCH] Fix tool paths and build flags
```
tatetian: Could you explain in the commit message what are the specific problems that the patch is trying to address? A reader needs to understand whether the patch is a short-term workaround (e.g., for an Asterinas bug) or a long-term solution (e.g., for compatibility with NixOS). And if there are more than one problem that you are trying to solve, then I think each problem should be a separate patch for maximal clarity.

tatetian (approved):

PR #3018, opened 2026-03-07
Add Yama ptrace scope

vvvvsv:
This PR adds Yama ptrace scope (/proc/sys/kernel/yama/ptrace_scope), which controls the scope of allowed “alien access”.

https://github.com/asterinas/asterinas/blob/d4c092f9ad08489b06b0cf236e711b1137fa92ef/kernel/src/process/posix_thread/alien_access.rs#L3-L5

Also adds some regression tests via opening `proc/pid/mem` and checking the behavior.

Reference: https://elixir.bootlin.com/linux/v6.16.5/source/security/yama/yama_lsm.c
and the note `/proc/sys/kernel/yama/ptrace_scope` in https://man7.org/linux/man-pages/man2/ptrace.2.html

Note: In the long term, we should introduce a framework to register security hooks, and move the Yama-related code into `kernel/src/security/`. But for now, we only have a single LSM Yama, and it is partial (without `yama_task_prctl` and `yama_task_free`), so I think it is OK to put it in the mod `alien_access`.

### DIFF
--- a/test/initramfs/src/apps/fs/procfs/pid_mem.c
+++ b/test/initramfs/src/apps/fs/procfs/pid_mem.c
@@ -7,10 +7,12 @@
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <stdint.h>
 #include <string.h>
+#include <linux/capability.h>
 
 #include "../../common/test.h"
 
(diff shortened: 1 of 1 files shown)


kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,93 @@ enum CredsSource {
...
+        })?;
+
+    Ok(())
+}
+
+static YAMA_SCOPE: AtomicI32 = AtomicI32::new(YamaScope::Relational as i32);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
+#[repr(i32)]
+/// The Yama scope levels.
```
lrh2000: Move doc comments above `#[derive(...)]`.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,93 @@ enum CredsSource {
...
     RealCreds,
 }
+
+/// Returns the current Yama scope for alien access.
+pub fn get_yama_scope() -> i32 {
+    YAMA_SCOPE.load(Ordering::Relaxed)
+}
+
+/// Sets the Yama scope for alien access.
+pub fn set_yama_scope(val: i32) -> Result<()> {
```
lrh2000: Why should public methods accept raw integers? This does not seem reasonable cross-module interfaces to me.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,93 @@ enum CredsSource {
...
+    YAMA_SCOPE
+        .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current_scope| {
+            let is_downgrading_from_no_attach =
+                current_scope == YamaScope::NoAttach as i32 && new_scope != YamaScope::NoAttach;
+            (!is_downgrading_from_no_attach).then_some(new_scope as i32)
+        })
+        .map_err(|_| {
+            Error::with_message(
+                Errno::EINVAL,
+                "can not downgrade Yama scope from `NoAttach`",
```
lrh2000:
```suggestion
                "`YamaScope::NoAttach` cannot be changed once set",
```

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,93 @@ enum CredsSource {
...
+    if !mode.0.contains(AlienAccessFlags::ATTACH) {
+        return Ok(());
+    }
+
+    match YamaScope::try_from(get_yama_scope()).unwrap() {
+        YamaScope::Disabled => {}
+        YamaScope::Relational => {
+            let caller_is_ancestor = is_ancestor_of(accessor.process(), target.process());
+            if !caller_is_ancestor && !caller_has_cap {
+                return_errno_with_message!(Errno::EPERM, "yama alien access check failed");
```
lrh2000:
```suggestion
                return_errno_with_message!(Errno::EPERM, "alien access is denied due to Yama scope");

File: kernel/src/fs/fs_impls/procfs/pid/task/maps.rs [hunk 5]
```diff
@@ -50,7 +50,8 @@ impl FileOpsByHandle for MapsFileOps {
             .check_alien_access_from(
                 current_thread!().as_posix_thread().unwrap(),
                 AlienAccessMode::READ_WITH_FS_CREDS,
-            )?;
+            )
+            .map_err(|_| Error::with_message(Errno::EACCES, "alien access check denied"))?;
```
lrh2000:
```suggestion
            .map_err(|_| Error::with_message(Errno::EACCES, "alien access is denied"))?;

File: kernel/src/process/posix_thread/alien_access.rs [hunk 6]
```diff
@@ -90,3 +96,93 @@ enum CredsSource {
...
+        }
+        YamaScope::Capability => {
+            if !caller_has_cap {
+                return_errno_with_message!(Errno::EPERM, "yama alien access check failed");
+            }
+        }
+        YamaScope::NoAttach => {
+            return_errno_with_message!(Errno::EPERM, "yama alien access check failed");
+        }
+    }
```
lrh2000, line 126: The code can be simplified as:
```rust
let is_denied = match scope {
    YamaScope::Capability => !caller_has_cap,
};

if is_denied {
    return_errno_with_message!();
}
```

test/initramfs/src/apps/fs/procfs/pid_mem.c
```diff
@@ -130,3 +143,205 @@ FN_TEST(proc_mem_local)
...
+
+		// 2. From the parent without `CAP_SYS_PTRACE`: allowed at 0 and 1, denied at 2.
+		write_yama_scope(0);
+		TEST_RES(access_from_parent(1), _ret == 0);
+		write_yama_scope(1);
+		TEST_RES(access_from_parent(1), _ret == 0);
+		write_yama_scope(2);
+		TEST_RES(access_from_parent(1), _ret == EACCES);
+
+		// 3. From the parent with `CAP_SYS_PTRACE`: allow at 0, 1 and 2.
```
lrh2000:
```suggestion
		// 3. From the parent with `CAP_SYS_PTRACE`: allowed at 0, 1 and 2.
```

lrh2000 (commented):
Thanks for your contribution. Just some minor comments.

vvvvsv:
@boterinas rerun

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,86 @@ enum CredsSource {
     FsCreds,
     RealCreds,
 }
+
+/// Returns the current Yama scope for alien access.
+pub fn get_yama_scope() -> i32 {
+    YAMA_SCOPE.load(Ordering::Relaxed)
```
lrh2000: I assume this method should also return `YamaScope`.

test/initramfs/src/apps/fs/procfs/pid_mem.c
```diff
@@ -130,3 +143,205 @@ FN_TEST(proc_mem_local)
...
+	CHECK(close(fd));
+}
+
+static void drop_cap_sys_ptrace(void)
+{
+	struct __user_cap_header_struct hdr = {
+		.version = _LINUX_CAPABILITY_VERSION_3,
+	};
+	struct __user_cap_data_struct capdat[2] = { 0 };
+	CHECK(syscall(SYS_capget, &hdr, &capdat));
```
lrh2000, line 216: Add a blank line before this line.

lrh2000 (approved):
Two minor nits. Otherwise LGTM.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -54,14 +60,14 @@ impl PosixThread {
             );
         }
 
-        // TODO: Add further security checks (e.g., YAMA LSM).
+        // TODO: Add a framework to register security hooks.
```
tatetian: **TODO scope is too narrow**: The TODO says "a framework to register security hooks," but hooks are just one piece of the Linux Security Module (LSM) subsystem. What's ultimately needed is the full LSM subsystem (of which hooks are a part), and Yama should become a minor LSM module within it. Consider moving the Yama-specific code (`YamaScope`, `YAMA_SCOPE`, `get_yama_scope`, `set_yama_scope`, `yama_check_alien_access`, `is_ancestor_of`) into a `yama` submodule now, and updating the TODO to reflect the larger goal:

```rust
// TODO: Rewrite Yama as a minor LSM module when Asterinas introduces the LSM subsystem.
mod yama {
    ...
}
```

This keeps the `alien_access` module focused on the generic access check while isolating the Yama policy, and makes it clear what the future migration target is.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,86 @@ enum CredsSource {
...
+pub enum YamaScope {
+    /// No restrictions.
+    Disabled = 0,
+    /// Only allow alien access by ascendant processes, or processes with `CapSet::SYS_PTRACE`.
+    Relational = 1,
+    /// Only allow alien access by processes with `CapSet::SYS_PTRACE`.
+    Capability = 2,
+    /// Disallow any alien access.
+    NoAttach = 3,
+}
```
tatetian: **Inaccurate scope descriptions**: Two issues:

1. The variant doc comments say "alien access" but Yama only restricts **ATTACH-level** alien access. READ-level alien access (e.g., opening `/proc/pid/maps`) is unaffected at every scope -- `yama_check_alien_access` returns `Ok(())` immediately when the mode lacks the `ATTACH` flag. Linux's documentation is precise about this, consistently saying "PTRACE_ATTACH" rather than "ptrace" generically.

2. The `Relational` variant uses "ascendant" which is an uncommon synonym for "ancestor." Since the helper function is already named `is_ancestor_of`, using "ancestor" in the doc comment would be more consistent and immediately clear.

Suggested wording (using "alien attach" to mirror `AlienAccessFlags::ATTACH`, and "ancestor" to match `is_ancestor_of`):

```rust
/// No additional restrictions on alien attach.
Disabled = 0,
/// Only allow alien attach by ancestor processes, or processes with `CapSet::SYS_PTRACE`.
Relational = 1,
/// Only allow alien attach by processes with `CapSet::SYS_PTRACE`.
Capability = 2,
/// Disallow any alien attach.
NoAttach = 3,
```

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -90,3 +96,86 @@ enum CredsSource {
...
+    };
+
+    if is_denied {
+        return_errno_with_message!(Errno::EPERM, "alien access is denied due to Yama scope");
+    }
+
+    Ok(())
+}
+
+fn is_ancestor_of(ancestor: Arc<Process>, descendant: Arc<Process>) -> bool {
```
tatetian: [**`minimize-copies`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/performance.html#minimize-copies): The `ancestor` parameter is only used for `Arc::ptr_eq`, which takes `&Arc<T>`. Taking it by value triggers an unnecessary atomic ref-count increment/decrement pair. Consider changing the signature to `fn is_ancestor_of(ancestor: &Arc<Process>, descendant: Arc<Process>) -> bool` and adjusting the call site to pass `&accessor.process()`.

kernel/src/fs/fs_impls/procfs/template/file.rs
```diff
@@ -165,3 +165,7 @@ impl<T: FileOpsByHandle> FileOps for T {
         Some(self.open(access_mode, status_flags))
     }
 }
+
+/// Worst case buffer size needed for holding an integer.
+// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/internal.h#L163>.
+pub const BUF_SIZE_I32: usize = 13;
```
tatetian: [**`cite-sources`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#cite-sources): Instead of copying the magic number from Linux without justification, derive it independently so a reader can verify it at a glance. The longest `i32` string is `"-2147483648\n\0"` (11-char numeral + newline + NUL = 13). Suggested replacement:

```rust
/// Worst case buffer size needed for holding an integer.
///
/// The longest possible string is `"-2147483648\n\0"`,
/// whose length is 13 bytes.
pub const BUF_SIZE_I32: usize = 13;
```

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn write_at(&self, _offset: usize, reader: &mut VmReader) -> Result<usize> {
+        let (cstr, read_bytes) = reader.read_cstring_until_end(BUF_SIZE_I32 - 1)?;
+        let val = cstr
+            .to_str()
+            .ok()
+            .map(|str| str.trim())
+            .and_then(|str| str.parse::<i32>().ok())
+            .ok_or_else(|| {
+                Error::with_message(Errno::EINVAL, "the value is not a valid integer")
+            })?;
```
tatetian: **Duplicated integer-from-userspace parsing**: This block is identical to the one in `oom_score_adj.rs` (lines 42-52). Since `BUF_SIZE_I32` was already extracted into a shared constant for exactly this reason, consider also extracting a small helper (e.g., on the `ReadCString` trait or as a free function in the `procfs::template` module) that encapsulates the read-parse-trim-validate chain:

```rust
/// Reads a string from `reader` and parses it as an `i32`.
fn read_i32_from(reader: &mut VmReader) -> Result<(i32, usize)> { ... }
```

This would deduplicate the two existing call sites and provide a ready-made building block for future procfs/sysctl integer knobs.

Even better, now that we have helper function, `BUF_SIZE_I32` can be encapsulateed inside to hide it as an implementation detail.

kernel/src/process/posix_thread/alien_access.rs
vvvvsv: > Consider changing the signature to `fn is_ancestor_of(ancestor: &Arc<Process>, descendant: Arc<Process>) -> bool` and adjusting the call site to pass `&accessor.process()`.

Thanks! This cannot reduce unnecessary atomic ref-count increment/decrement pairs, because the `Weak::upgrade` happens inside `accessor.process()`. The following implementation can reduce this ref-count:

```rust
    fn is_ancestor_of(ancestor: &Weak<Process>, descendant: Arc<Process>) -> bool {
        let mut current = descendant;
        loop {
            let parent_guard = current.parent().lock();
            let parent = parent_guard.process();
            if Weak::ptr_eq(parent, ancestor) {
                return true;
            }
            let Some(parent) = parent.upgrade() else {
                return false;
            };
            drop(parent_guard);
            current = parent;
        }
    }
```

tatetian (approved):

PR #3019, opened 2026-03-08
Support `qemu-direct` boot mode for TDX-enabled Asterinas guests

taosue:
Since CoCo uses the `qemu-direct` boot method to launch guests, we should support this method as well. As an added benefit, we can directly extract the Asterinas bzImage without needing to extract it from a QCOW2 image.

Previously, we observed that OVMF injects the parameter `initrd=initrd` at the end of the kernel command line—potentially after the `--` delimiter. According to Linux kernel command-line parsing rules, everything after `--` is treated as an argument to `init`, causing boot failures. This issue has been confirmed as a bug in OVMF and was fixed in a upstream [commit](https://github.com/tianocore/edk2/commit/a2ac0fea49996ab484c1a8761c234cc354f5a760). Therefore, we are updating the OVMF version to include this fix.

Relies on #3011


lrh2000:
QEMU direct boot is tested by the CI, which passes before this PR.

However, this PR breaks the CI, e.g., [syscall-ext2-microvm](https://github.com/asterinas/asterinas/actions/runs/22815293515/job/66179214391). All new PRs are failing the CI, which is very annoying.

@taosue Do you know what's happening and do you have a fix? If not, I propose to just revert the commit ("Support all boot methods for TDX") and re-land the "fix" later in a proper way.

taosue:
> QEMU direct boot is tested by the CI, which passes before this PR.
> 
> However, this PR breaks the CI, e.g., [syscall-ext2-microvm](https://github.com/asterinas/asterinas/actions/runs/22815293515/job/66179214391). All new PRs are failing the CI, which is very annoying.
> 
> @taosue Do you know what's happening and do you have a fix? If not, I propose to just revert the commit ("Support all boot methods for TDX") and re-land the "fix" later in a proper way.

Previously, `qemu-direct` continued to use the `ELF` format even when the boot protocol was Linux-related. In essence, the Linux boot protocol was not actually enabled when using `qemu-direct` (which is also utilized by `microvm` scheme).

This PR addresses the issue by adding `bzImage` support to `qemu-direct`, thereby enabling the combination of `qemu-direct` with the Linux boot protocol. So we can simply use either `-bios OVMF.fd` or `-bios MICROVM.fd` when booting `bzImage`.

`linux-legacy32` previously encountered the same issue—it still used `ELF` format. However, simply adding OVMF would not resolve this problem. I am currently working on implementing a proper fix for this case.

**I'm not sure if I can provide a fix soon, so no objections to reverting this PR.**

lrh2000:
Thank you for your reply. I now understand your motivation, and it makes sense. Then, let's first accept https://github.com/asterinas/asterinas/pull/3039 so that you won't need to hurry to work out a proper fix for the CI issue.

PR #3020, opened 2026-03-09
Optimized the build & install workflow of nixos

wyt8:
# Overview

This PR optimizes the NixOS installation script by decoupling the system build from the installation process. By using `nix-build` separately, we ensure that the Nix store can effectively cache packages even across different build sessions.

# Motivation

Previously, running `make clean` followed by `make nixos` would often cause Nix to lose track of intermediate build states. This forced the script to re-download all packages defined in configuration.nix every time.

For large packages like `cudatoolkit`, these redundant downloads significantly increased the total build time and network overhead.

With this change, `nix-build` can better leverage the Nix store. Even if the local make environment is cleaned, the Nix store paths remain valid, allowing `nix-build` to detect that the packages (like CUDA) already exist locally, skipping the download phase entirely.

---

**Reference**:

* [nixos-install](https://www.mankier.com/8/nixos-install#--system)

### DIFF
--- a/distro/aster_nixos_installer/templates/install_nixos.sh
+++ b/distro/aster_nixos_installer/templates/install_nixos.sh
@@ -131,9 +131,25 @@ cp @aster-configuration@ ${BUILD_DIR}/etc/nixos/aster_configuration.nix
 cp -r @aster-etc-nixos@/modules ${BUILD_DIR}/etc/nixos
 cp -r @aster-etc-nixos@/overlays ${BUILD_DIR}/etc/nixos
 
+COMMON_NIX_ARGS=(
+    --option extra-substituters "@aster-substituters@"
+    --option extra-trusted-public-keys "@aster-trusted-public-keys@"
+)
+
+# Pre-build the NixOS system closure on the host
+# so that the result is cached in the host's Nix store.
+# Without this, nixos-install would build the system inside the (possibly empty) target image,
+# causing redundant downloads after every `make clean`.
+#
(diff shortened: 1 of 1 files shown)


distro/aster_nixos_installer/templates/install_nixos.sh
```diff
@@ -131,8 +131,15 @@ cp @aster-configuration@ ${BUILD_DIR}/etc/nixos/aster_configuration.nix
 cp -r @aster-etc-nixos@/modules ${BUILD_DIR}/etc/nixos
 cp -r @aster-etc-nixos@/overlays ${BUILD_DIR}/etc/nixos
 
+SYSTEM_CLOSURE=$(nix-build '<nixpkgs/nixos>' -A system \
```
tatetian, line 145: [**`explain-why`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#explain-why): Please add a comment above this block explaining *why* the system closure is pre-built separately. A future reader won't know that `make clean` removes the disk image (and thus the target's Nix store), or that without `--system` `nixos-install` would re-download every package into the fresh image. Something like:

```sh
# Pre-build the NixOS system closure on the host
# so that the result is cached in the host's Nix store.
# Without this, nixos-install would build the system inside the (possibly empty) target image,
# causing redundant downloads after every `make clean`.
#
# See: https://www.mankier.com/8/nixos-install#--system
```

tatetian:
@wyt8 Please also address Copilot comments, which sounds reasonable to me.

Issue #2801, opened 2025-12-26
Undefined behaviour in `<Weak as NonNullPtr>::into_raw`


rikosellic:
### Describe the bug
The `NonNullPtr::into_raw` implementation for `Weak<T>` constructs a `NonNull` from the pointer returned by `Weak::into_raw`. However, according to the [std documentation](https://github.com/rust-lang/rust/blob/1aa9bab4ecbce4859eaad53000f78158ebe2be2c/library/alloc/src/sync.rs#L3042-L3043), the safety contract of the pointer returned by `Weak::into_raw` is the same as that of `Weak::as_ptr`, which explicitly allows [returning a null pointer](https://github.com/rust-lang/rust/blob/1aa9bab4ecbce4859eaad53000f78158ebe2be2c/library/alloc/src/sync.rs#L3080-L3081).

https://github.com/asterinas/asterinas/blob/08be9252402cf14e14000035f2e9ad40fc2a6a1d/ostd/src/sync/rcu/non_null/mod.rs#L233-L238

lrh2000:
Thank you for reporting this soundness issue!

I don't think anyone is using `Weak` as `NonNullPtr` now. I assume we should just remove this `impl`.

That being said, I'm surprised that `Weak::as_ptr` can return a null pointer. I thought it could only return a dangling (but still non-null) pointer, as noted in the code:
https://github.com/asterinas/asterinas/blob/08be9252402cf14e14000035f2e9ad40fc2a6a1d/ostd/src/sync/rcu/non_null/mod.rs#L229-L231

rikosellic:
Indeed, I was also surprised when I read the doc. I agree that we should remove this `impl`, as `Weak` cannot soundly implement `NonNullPtr` under the existing API contract.

StevenJiang1110:
Following the discussion at https://github.com/asterinas/asterinas/pull/3021#discussion_r2959669997, I opened an issue on [Zulip about whether Weak::as_ptr can return a null pointer](https://rust-lang.zulipchat.com/#narrow/channel/327149-t-libs-api.2Fapi-changes/topic/Can.20the.20result.20of.20.60sync.3A.3AWeak.3A.3Aas_ptr.60.20be.20a.20null.20pointer.3F/with/582306780).

[...]

PR #3021, opened 2026-03-09
Refactor process and thread lookups into unified PID table

StevenJiang1110:
This PR introduces `PidTable`, which consolidates the thread table, process table, process group table, and session table into a single unified structure. This is a prerequisite for supporting PID namespaces.

`PidTable` stores `PidEntry`, which mirrors Linux's [struct pid](https://elixir.bootlin.com/linux/v6.18/source/include/linux/pid.h#L57). Since PID namespaces are hierarchical, maintaining separate thread tables, process tables, and so on at each namespace level would require recursively updating every ancestor namespace whenever a process or thread is added or removed. By introducing `PidTable`, adding a process in the future only requires updating a single `PidEntry`, rather than propagating changes across all PID namespaces in the hierarchy.

This PR contains four commits:

1. Introduces a unified PID table that stores threads, processes, process groups, and sessions in one place, replacing the previous split lookup paths.
2. Adjusts locking around PID-table-related operations to preserve a consistent lock order. That is to say, we need to lock PID table at first, then `process.tasks`. 
3. Removes thin public wrapper functions around the PID table and switches call sites to access PidTable directly through pid_table_mut().
4. Renames the module/file from process_table to pid_table so the name matches its role more directly.
5. Refactored process/session ownership so `Process` and `ProcessGroup` hold strong references to their `ProcessGroup`/`Session` while membership links and unified PID table entries use `Weak`, making lifetime management consistent.

Commit 3, 4 and 5 are solely refactor commits.

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/mod.rs
@@ -188,8 +188,8 @@ impl DirOps for RootDirOps {
         {
             let mut cached_children = dir.cached_children().write();
             return Ok(cached_children
-                .put_entry_if_not_found(name, || {
-                    PidDirOps::new_inode(process_ref.clone(), dir.this_weak().clone())
+                .put_entry_if_not_found(name, move || {
+                    PidDirOps::new_inode(process_ref, dir.this_weak().clone())
                 })
                 .clone());
         }
@@ -216,8 +216,8 @@ impl DirOps for RootDirOps {
 
         for process_ref in pid_table.iter_processes() {
             let pid = process_ref.pid().to_string();
-            cached_children.put_entry_if_not_found(&pid, || {
-                PidDirOps::new_inode(process_ref.clone(), dir.this_weak().clone())
+            cached_children.put_entry_if_not_found(&pid, move || {
+                PidDirOps::new_inode(process_ref, dir.this_weak().clone())
             });
         }
 
--- a/kernel/src/process/pid_table.rs
+++ b/kernel/src/process/pid_table.rs
@@ -48,11 +48,11 @@ impl PidTable {
     // ---- Thread operations ----
 
     /// Inserts a thread into the table.
-    pub(super) fn insert_thread(&mut self, tid: Tid, thread: Arc<Thread>) {
+    pub(super) fn insert_thread(&mut self, tid: Tid, thread: &Arc<Thread>) {
         debug_assert_eq!(tid, thread.as_posix_thread().unwrap().tid());
 
         let entry = self.get_or_create_entry(tid);
-        entry.lock().set_thread(&thread);
+        entry.lock().set_thread(thread);
     }
 
     /// Removes a thread from the table.
@@ -119,16 +119,17 @@ impl PidTable {
     // ---- Process operations ----
 
     /// Inserts a process into the table.
-    pub(super) fn insert_process(&mut self, pid: Pid, process: Arc<Process>) {
-        // Clone the entry handle so the borrow of `self` ends before we update
-        // `self.process_count`.
-        let entry = self.get_or_create_entry(pid).clone();
-        let mut entry = entry.lock();
-
-        debug_assert!(!entry.has_live_process());
+    pub(super) fn insert_process(&mut self, pid: Pid, process: &Arc<Process>) {
+        debug_assert!(
+            !self
+                .entries
+                .get(&pid)
+                .is_some_and(|entry| entry.lock().has_live_process())
+        );
         self.process_count += 1;
 
-        entry.set_process(&process);
+        let entry = self.get_or_create_entry(pid);
+        entry.lock().set_process(process);
     }
 
     /// Removes a process from the table and notifies observers.
@@ -177,9 +178,9 @@ impl PidTable {
     // ---- Process group operations ----
 
     /// Inserts a process group into the table.
-    pub(super) fn insert_process_group(&mut self, pgid: Pgid, group: Arc<ProcessGroup>) {
+    pub(super) fn insert_process_group(&mut self, pgid: Pgid, group: &Arc<ProcessGroup>) {
         let entry = self.get_or_create_entry(pgid);
-        entry.lock().set_process_group(&group);
+        entry.lock().set_process_group(group);
     }
 
     /// Removes a process group from the table.
@@ -217,9 +218,9 @@ impl PidTable {
     // ---- Session operations ----
 
     /// Inserts a session into the table.
-    pub(super) fn insert_session(&mut self, sid: Sid, session: Arc<Session>) {
+    pub(super) fn insert_session(&mut self, sid: Sid, session: &Arc<Session>) {
         let entry = self.get_or_create_entry(sid);
-        entry.lock().set_session(&session);
+        entry.lock().set_session(session);
     }
 
     /// Removes a session from the table.
--- a/kernel/src/fs/fs_impls/procfs/pid/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/mod.rs
@@ -33,7 +33,7 @@ impl PidDirOps {
             thread_ref: None,
         };
         // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/base.c#L3493>
-        ProcDirBuilder::new(Self(tid_dir_ops.clone()), mkmod!(a+rx))
+        ProcDirBuilder::new(Self(tid_dir_ops), mkmod!(a+rx))
             .parent(parent)
             // The PID directories must be volatile, because it is just associated with one process.
             .volatile()
--- a/kernel/src/process/posix_thread/builder.rs
+++ b/kernel/src/process/posix_thread/builder.rs
@@ -209,7 +209,7 @@ impl PosixThreadBuilder {
                 ns_proxy,
             );
 
-            pid_table::pid_table_mut().insert_thread(tid, thread.clone());
+            pid_table::pid_table_mut().insert_thread(tid, &thread);
             task::create_new_user_task(user_ctx, thread, thread_local, is_init_process)
         })
     }
--- a/kernel/src/process/process/init_proc.rs
+++ b/kernel/src/process/process/init_proc.rs
@@ -119,7 +119,7 @@ fn set_session_and_group(process: &Arc<Process>) {
     process.set_new_session(&mut process.process_group.lock(), &mut pid_table);
 
     // Add the new process to the global table
-    pid_table.insert_process(process.pid(), process.clone());
+    pid_table.insert_process(process.pid(), process);
 }
 
 /// Creates the init task from the given executable file.


kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+};
+
+static PID_TABLE: Mutex<PidTable> = Mutex::new(PidTable::new());
+
+/// An entry in the unified PID table.
+///
+/// Each entry stores references to the thread, process, process group, and
+/// session that share the same numeric identifier. Not all slots need to be
+/// occupied at the same time.
+struct PidEntry {
```
tatetian: [**`top-down-reading`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#top-down-reading): The lower-level `PidEntry` and `PidEntryInner` structs appear before the higher-level `PidTable` (line 123). The file should be reorganized so readers encounter `PidTable` (the public-facing struct) first, then `PidEntry` (its internal building block).

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+        self.inner.lock().process.strong_count() > 0
+    }
+
+    /// Returns the process group associated with the entry, if any.
+    fn process_group(&self) -> Option<Arc<ProcessGroup>> {
+        self.inner.lock().process_group.upgrade()
+    }
+
+    /// Sets the thread reference.
+    pub(super) fn set_thread(&self, thread: &Arc<Thread>) {
```
tatetian: [**`narrow-visibility`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#narrow-visibility): Two visibility issues:

**`pub(super)` methods on a private struct**: `PidEntry` itself has no visibility modifier (module-private), but its mutator methods (`set_thread`, `clear_thread`, `set_process`, etc.) are all marked `pub(super)`. This is contradictory -- the `pub(super)` is dead since no code outside the module can name the type. These should just be `fn` (private).

**`pub` methods on a `pub(crate)` struct:** `PidTable` is marked `pub(crate)`, but its methods are marked `pub`. Since `pub` and `pub(crate)` make no difference for the kernel crate (no downstream users), the shorter one (`pub`) should be preferred.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+
+static PID_TABLE: Mutex<PidTable> = Mutex::new(PidTable::new());
+
+/// An entry in the unified PID table.
+///
+/// Each entry stores references to the thread, process, process group, and
+/// session that share the same numeric identifier. Not all slots need to be
+/// occupied at the same time.
+struct PidEntry {
+    inner: Mutex<PidEntryInner>,
```
tatetian: **Move the `Mutex` from inside `PidEntry` to outside**: The inner `Mutex<PidEntryInner>` is likely intended for future PID namespace support where entries are shared across namespace levels. However, the `Mutex` should be on the outside -- i.e., `BTreeMap<u32, Arc<Mutex<PidEntry>>>` -- so that callers can hold the entry lock across multiple operations.

Currently, every `set_*`, `clear_*`, and getter method on `PidEntry` independently locks and unlocks `inner`. Consider `make_current_main_thread` (line 302), which touches two entries (`pid` and `old_tid`) with 4 separate inner-mutex acquisitions:

```rust
pid_table.remove_thread(pid);        // lock pid entry, clear, unlock
let thread = pid_table.get_thread(old_tid).unwrap(); // lock old_tid entry, upgrade, unlock
pid_table.remove_thread(old_tid);    // lock old_tid entry, clear, unlock
pid_table.insert_thread(pid, thread); // lock pid entry, set, unlock
```

That's 2 acquisitions on the `pid` entry and 2 on the `old_tid` entry. With the `Mutex` moved outside, `PidTable` can expose an `entry()` method and callers can hold each entry lock across its related operations:

```rust
pub(crate) struct PidTable {
    entries: BTreeMap<u32, Arc<Mutex<PidEntry>>>,
    ...
}

impl PidTable {
    pub fn entry(&self, id: u32) -> &Mutex<PidEntry> { ... }
}
```

```rust
// 1 acquisition on old_tid entry (was 2)
let mut old_entry = pid_table.entry(old_tid).lock();
let thread = old_entry.thread().unwrap();
old_entry.clear_thread();
drop(old_entry);

// 1 acquisition on pid entry (was 2)
let mut pid_entry = pid_table.entry(pid).lock();
pid_entry.clear_thread();
pid_entry.set_thread(&thread);
drop(pid_entry);
```

This doesn't increase API surface for misuse since everything available on `PidEntry` is already available through the corresponding `PidTable` methods. `PidTable` can keep its convenience methods (`insert_thread`, `get_thread`, etc.) for the common single-operation case.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub fn get_process_group(&self, pgid: &Pgid) -> Option<Arc<ProcessGroup>> {
+        self.entries.get(pgid).and_then(|e| e.process_group())
+    }
+
+    /// Returns whether a process group with the given pgid exists.
+    pub fn contains_process_group(&self, pgid: &Pgid) -> bool {
+        self.entries
+            .get(pgid)
+            .is_some_and(|entry| entry.inner.lock().process_group.strong_count() > 0)
+    }
```
tatetian, line 216: [**`hide-implementation`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#hide-implementation): This method directly accesses `entry.inner.lock()` instead of going through `PidEntry`'s accessor methods. Other methods (e.g., `has_live_process()` on line 60) correctly encapsulate this pattern. Add a corresponding `has_live_process_group()` method to `PidEntry` and call it here:

```rust
fn has_live_process_group(&self) -> bool {
    self.inner.lock().process_group.strong_count() > 0
}
```

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    /// Removes a session from the table.
+    pub fn remove_session(&mut self, sid: Sid) {
+        if let Some(entry) = self.entries.get(&sid) {
+            entry.clear_session();
+        }
+        self.try_remove_entry(sid);
+    }
+
+    /// Returns an iterator over threads that have a live thread reference.
```
tatetian: **Misplaced method**: `iter_threads` is placed in the "Session operations" section but it iterates over threads. Move it to the "Thread operations" section (after `get_thread` on line 174) for consistency with the other section-based groupings.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    assert!(tasks.has_exited_main());
+    assert!(tasks.in_execve());
+    assert_eq!(tasks.as_slice().len(), 2);
+    assert!(core::ptr::eq(ctx.task, tasks.as_slice()[1].as_ref()));
+
+    tasks.swap_main(pid, old_tid);
+    ctx.posix_thread.set_main(pid);
+
+    pid_table.remove_thread(pid);
```
tatetian: **Misplaced function:**  `make_current_main_thread` only uses `PidTable`'s public methods and has a single caller (`process/execve.rs`). It would be more natural to move this function to `process/execve.rs` rather than keeping it in `pid_table.rs`, where it looks like part of the table's internal machinery.

kernel/src/process/process/mod.rs
```diff
@@ -493,25 +477,18 @@ impl Process {
...
             return_errno_with_message!(
                 Errno::ESRCH,
                 "the process to set the PGID is neither the current process nor its child process"
             );
         };
 
-        if let Some(new_process_group) = group_table_mut.get(&pgid).cloned() {
-            process.to_existing_group(current_session, group_table_mut, new_process_group)
+        if let Some(new_process_group) = pid_table.get_process_group(&pgid) {
+            process.to_existing_group(current_session, &mut pid_table, new_process_group)
```
tatetian, line 493: **Unnecessary clone**: `self.process_group.lock()` returns a `MutexGuard<Option<Arc<ProcessGroup>>>`. Calling `.clone()` on the guard clones the `Option<Arc<ProcessGroup>>`. Then `.unwrap()` extracts the `Arc`. Instead, use `.as_ref()` to avoid cloning the `Arc` until needed:

```rust
Some(self.process_group.lock().as_ref().unwrap().session().clone())
```

This is consistent with the pattern used elsewhere in this PR (e.g., `process_group_mut.as_ref().cloned().unwrap()` on lines 164, 520, etc.).

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,315 @@
...
+/// session that share the same numeric identifier. Not all slots need to be
+/// occupied at the same time.
+struct PidEntry {
+    inner: Mutex<PidEntryInner>,
+}
+
+struct PidEntryInner {
+    thread: Weak<Thread>,
+    process: Weak<Process>,
+    process_group: Weak<ProcessGroup>,
```
tatetian: [`explain-why`](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#explain-why). Explain the rationale behind keeping weak references rather than strong ones.

tatetian:
@boterinas codex

kernel/src/syscall/pivot_root.rs
```diff
@@ -28,14 +27,12 @@ pub fn sys_pivot_root(
...
-    with_global_threads(move |table| {
-        let fs_ref = ctx.thread_local.borrow_fs();
-        let mut fs_resolver = fs_ref.resolver().write();
-
-        fs_resolver.pivot_root(new_root_path, put_old_path, table, ctx)
-    })?;
+    // TODO: Locking the global PID table here is a workaround. We need to use a
+    // more suitable lock (i.e. the global mount lock or the namespace lock) to avoid deadlock.
+    let pid_table = pid_table::pid_table_mut();
+    let fs_ref = ctx.thread_local.borrow_fs();
```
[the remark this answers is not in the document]
StevenJiang1110: This issue was not introduced by this PR. pivot_root should hold the per-PID namespace lock rather than the global PID table lock. Since PID namespace support is not yet implemented, holding the PID table lock is the only option for now. A follow-up PR will address this.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    /// Clears the session reference.
+    fn clear_session(&mut self) {
+        self.session = Weak::new();
+    }
+
+    /// Returns `true` if the entry no longer tracks any live object.
+    fn is_empty(&self) -> bool {
+        self.thread.strong_count() == 0
```
lrh2000: I think you meant to write `Weak::ptr_eq(&self.thread, &Weak::new())` instead of `strong_count() == 0`. The reference counts of `Arc<Thread>`, `Arc<Process>`, e.t.c. are unstable, unreliable, and unsuable because they can reach zero in many ways.

To clarify the purpose, how do you like:
```rust
trait WeakIsDanging {
    fn is_dangling(&self) -> bool;
}

impl<T> WeakIsDangling for Weak<T> {
    fn is_dangling(&self) -> bool {
        const DANGLING: Weak<T> = Weak::new();
        Weak::ptr_eq(self, &DANGLING)
    }
}
```

https://doc.rust-lang.org/std/rc/struct.Weak.html#method.ptr_eq:
> Since this compares pointers it means that Weak::new() will equal each other, even though they don’t point to any allocation.

Note that `strong_count()` will also have to check `is_dangling()` first to decide whether the allocation exists. So this suggested approach should also be more efficient.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Removes a thread from the table.
+    pub fn remove_thread(&mut self, tid: Tid) {
+        if let Some(entry) = self.entries.get(&tid) {
+            entry.lock().clear_thread();
+        }
+        self.try_remove_entry(tid);
+    }
+
+    /// Gets a thread by tid.
```
lrh2000:
```suggestion
    /// Gets a thread by a TID.
```

Same to other similar cases.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,326 @@
...
+};
+
+static PID_TABLE: Mutex<PidTable> = Mutex::new(PidTable::new());
+
+/// The unified PID table.
+///
+/// Combines the process, process-group, session, and thread tables into a
+/// single structure.
+pub struct PidTable {
+    entries: BTreeMap<u32, Arc<Mutex<PidEntry>>>,
```
lrh2000, line 27: Could you please explain why you discarded the original `Rcu` version (in your old PID namespace PR) in favour of a `Mutex`?

kernel/src/fs/fs_impls/procfs/mod.rs
```diff
@@ -176,15 +176,15 @@ impl Observer<PidEvent> for ProcDir<RootDirOps> {
 }
 
 impl DirOps for RootDirOps {
-    // Lock order: process table -> cached entries
+    // Lock order: PID table -> cached entries
```
lrh2000, line 179: For some reason, it seems you've decided to use "pid table" (instead of "PID table") in the "Lock order:" comments. This is an inconsistency that needs to be fixed. I am not sure if there are others.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Gets a thread by tid.
+    pub fn get_thread(&self, tid: Tid) -> Option<Arc<Thread>> {
+        self.entries
+            .get(&tid)
+            .and_then(|entry| entry.lock().thread())
+    }
+
+    /// Removes a thread from the table and returns it.
+    pub fn take_thread(&mut self, tid: Tid) -> Option<Arc<Thread>> {
```
lrh2000: I think all mutable operations should be `pub(super)`. I cannot think of a valid use case for users outside of the `process` module, can you?

kernel/src/process/pid_table.rs
StevenJiang1110: The suggestion makes sense to me! I will refactor it as suggested.

kernel/src/process/pid_table.rs
StevenJiang1110: This is because the contents of `PidEntry` was changed from `Arc<Process>` to `Weak<Process>` in this PR. However, our `Rcu` implementation does not support `Weak`. I recall this used to work, but it no longer does. If `Rcu<Weak<Process>>` could be made to work, it would be preferable to using a `Mutex`.

lrh2000: > it no longer does

It is because https://github.com/asterinas/asterinas/issues/2801.

In practice, I'd say that `Rcu<Weak>` works because the Rust standard library's `Weak::new()` uses a non-zero dangling pointer. This means that `Weak::as_ptr` never returns a null pointer.

But the Rust standard library's _documentation_ [says](https://github.com/rust-lang/rust/blob/1aa9bab4ecbce4859eaad53000f78158ebe2be2c/library/alloc/src/sync.rs#L3080-L3081) `Weak::as_ptr` may return a null pointer, though I believe the current _implementation_ will never...

We can ask the Rust [libs-api team](https://github.com/rust-lang/libs-team) (or on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/327149-t-libs-api.2Fapi-changes)) to find out if they are willing to document the current implementation. Even if they prefer not to, we can find out why. If they agree to document it, we can safely bring back our previous `Rcu<Weak>` implementation.

StevenJiang1110: Thanks for your information!

> We can ask the Rust [libs-api team](https://github.com/rust-lang/libs-team) (or on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/327149-t-libs-api.2Fapi-changes)) to find out if they are willing to document the current implementation.

I will take it. If there's any progress, I will share it under #2801.

kernel/src/process/process/session.rs
```diff
@@ -33,30 +33,22 @@ impl Session {
     /// The caller needs to ensure that the process does not belong to other process group or other
     /// session.
     pub(in crate::process) fn new_pair(process: Arc<Process>) -> (Arc<Self>, Arc<ProcessGroup>) {
```
lrh2000:
```suggestion
    pub(in crate::process) fn new_pair(process: &Arc<Process>) -> (Arc<Self>, Arc<ProcessGroup>) {
```

kernel/src/process/process/process_group.rs
```diff
@@ -26,12 +26,12 @@ impl ProcessGroup {
     /// process of the new process group.
     ///
     /// The caller needs to ensure that the process does not belong to other process group.
-    pub(super) fn new(process: Arc<Process>, session: Weak<Session>) -> Arc<Self> {
+    pub(super) fn new(process: Arc<Process>, session: Arc<Session>) -> Arc<Self> {
```
lrh2000:
```suggestion
    pub(super) fn new(process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
```

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    // ---- Process operations ----
+
+    /// Inserts a process into the table.
+    pub(super) fn insert_process(&mut self, pid: Pid, process: Arc<Process>) {
+        let entry = self.get_or_create_entry(pid);
+        let mut entry = entry.lock();
+        if !entry.has_live_process() {
+            self.process_count += 1;
+        }
```
lrh2000: We cannot overwrite a process?
```rust
debug_assert!(entry.has_live_process());
self.process_count += 1;
```

`remove_process` should be similar.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,340 @@
...
+        let entry = self.get_or_create_entry(tid);
+        entry.lock().set_thread(&thread);
+    }
+
+    /// Removes a thread from the table.
+    pub(super) fn remove_thread(&mut self, tid: Tid) {
+        if let Some(entry) = self.entries.get(&tid) {
+            entry.lock().clear_thread();
+        }
+        self.try_remove_entry(tid);
```
lrh2000: We should use `BTreeMap::entry_mut` for all `remove_*`/`take_*` methods to avoid looking up the B-Tree map twice. `try_remove_entry` should be removed because it is a bad pattern that always locks the entry twice.

Here's an example (which may not compile) .The variable names are placeholders. Maybe there are better ones, since I feel there are too many entries (B-Tree map entries, PID entries, occupied B-Tree map entries...?).
```rust
let map_entry = self.entries.entry_mut(&tid);

debug_assert!(matches!(map_entry, Entry::Occupied(_));
let Entry::Occupied(entry) = map_entry else {
    return;
};

let locked_entry = entry.value().lock();
locked_entry.clear_thread();
if locked_entry.is_empty() {
    drop(locked_entry);
    entry.remove();
}
```

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    // ---- Process operations ----
+
+    /// Inserts a process into the table.
+    pub(super) fn insert_process(&mut self, pid: Pid, process: Arc<Process>) {
+        let entry = self.get_or_create_entry(pid);
+        let mut entry = entry.lock();
+        debug_assert!(!entry.has_live_process());
+        self.process_count += 1;
+        entry.set_process(&process);
```
lrh2000: Make `debug_assert!` and `process_count += 1` to a separate code block:
```rust
        let entry = self.get_or_create_entry(pid);
        let mut entry = entry.lock();

        debug_assert!(!entry.has_live_process());
        self.process_count += 1;

        entry.set_process(&process);
```

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        let should_remove = {
+            let mut pid_entry = map_entry.get().lock();
+
+            debug_assert!(pid_entry.has_live_process());
+
+            self.process_count -= 1;
+            pid_entry.clear_process();
+            // Drop the locked PID entry before removing the B-tree entry.
+            pid_entry.is_empty()
```
lrh2000, line 149: Same:
```rust
            let mut pid_entry = map_entry.get().lock();

            debug_assert!(pid_entry.has_live_process());
            self.process_count -= 1;

            pid_entry.clear_process();
            // Drop the locked PID entry before removing the B-tree entry.
            pid_entry.is_empty()
```

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,370 @@
...
+            subject: Subject::new(),
+        }
+    }
+
+    /// Returns the entry for the given ID, or creates a new one if absent.
+    fn get_or_create_entry(&mut self, id: u32) -> Arc<Mutex<PidEntry>> {
+        self.entries
+            .entry(id)
+            .or_insert_with(|| Arc::new(Mutex::new(PidEntry::new())))
+            .clone()
```
lrh2000: Why clone it? Returning a reference should work?

StevenJiang1110:
The CI failure is not related to this PR. Block this test in #3060

kernel/src/fs/fs_impls/procfs/mod.rs
lrh2000: Example 1

kernel/src/process/process/mod.rs
```diff
@@ -395,59 +391,49 @@ impl Process {
...
-        session_table_mut: &mut MutexGuard<BTreeMap<Sid, Arc<Session>>>,
-        group_table_mut: &mut MutexGuard<BTreeMap<Pgid, Arc<ProcessGroup>>>,
+        process_group_mut: &mut MutexGuard<Option<Arc<ProcessGroup>>>,
+        pid_table: &mut PidTable,
     ) {
-        let process_group = process_group_mut.upgrade().unwrap();
+        let process_group = process_group_mut.as_ref().cloned().unwrap();
         let mut process_group_inner = process_group.lock();
-        let session = process_group.session().unwrap();
+        let session = process_group.session().clone();
```
lrh2000: Example 2

kernel/src/process/process/session.rs
```diff
@@ -32,31 +32,23 @@ impl Session {
...
+            }),
         });
 
-        (session, process_group.unwrap())
+        let process_group = ProcessGroup::new(process, session.clone());
+        session
+            .inner
+            .lock()
+            .process_groups
+            .insert(process_group.pgid(), Arc::downgrade(&process_group));
```
lrh2000: `session.lock().insert_process_group(&process_group)`?

kernel/src/process/process/mod.rs
```diff
@@ -579,36 +565,45 @@ impl Process {
...
+        if current_session
+            .as_ref()
+            .is_some_and(|current| !Arc::ptr_eq(current, session))
+        {
             return_errno_with_message!(Errno::EPERM, "the process belongs to a different session");
         }
         if process_group.pgid() == self.pid {
-            // We'll hit this if the process is a session leader. There is no need to check below.
+            // We'll hit this if the process is a session leader. There is no need to check
+            // below.
```
lrh2000: This line is less than 100 characters long and shorter than the line above it:
```rust
            return_errno_with_message!(Errno::EPERM, "the process belongs to a different session");
```

Not sure why we want to wrap the last word?

---

There are several other problems in `to_new_group` and `to_existing_group`:
 - It seems `session` is defined twice. I don't think it's necessary.
 - Some unnecessary `clone()`s can be removed.
```patch
diff --git a/kernel/src/process/process/mod.rs b/kernel/src/process/process/mod.rs
index 2889259df..23a4570a0 100644
--- a/kernel/src/process/process/mod.rs
+++ b/kernel/src/process/process/mod.rs
@@ -506,7 +506,7 @@ impl Process {
     ) -> Result<()> {
         let mut process_group_mut = self.process_group.lock();
 
-        let process_group = process_group_mut.as_ref().unwrap().clone();
+        let process_group = process_group_mut.as_ref().unwrap();
         let session = process_group.session();
 
         if session.sid() == self.pid {
@@ -543,15 +543,16 @@ impl Process {
                 }
                 core::cmp::Ordering::Equal => return Ok(()),
             };
-        let session = process_group.session();
-        let mut session_inner = session.lock();
 
         // Remove the process from the old process group
+        let mut session_inner = session.lock();
         process_group_inner.remove_process(&self.pid);
         if process_group_inner.is_empty() {
             pid_table.remove_process_group(process_group.pgid());
             session_inner.remove_process_group(&process_group.pgid());
         }
+        drop(session_inner);
+        drop(process_group_inner);
 
         // Insert the process to the new process group
         new_group_inner.insert_process(self);
@@ -569,7 +570,7 @@ impl Process {
     ) -> Result<()> {
         let mut process_group_mut = self.process_group.lock();
 
-        let process_group = process_group_mut.as_ref().unwrap().clone();
+        let process_group = process_group_mut.as_ref().unwrap();
         let session = process_group.session();
 
         if current_session
@@ -579,30 +580,25 @@ impl Process {
             return_errno_with_message!(Errno::EPERM, "the process belongs to a different session");
         }
         if process_group.pgid() == self.pid {
-            // We'll hit this if the process is a session leader. There is no need to check
-            // below.
+            // We'll hit this if the process is a session leader. There is no need to check below.
             return Ok(());
         }
 
+        // Remove the process from the old process group
         let mut process_group_inner = process_group.lock();
-        let session = process_group.session().clone();
-
         let mut session_inner = session.lock();
-        // Remove the process from the old process group
         process_group_inner.remove_process(&self.pid);
         if process_group_inner.is_empty() {
             pid_table.remove_process_group(process_group.pgid());
             session_inner.remove_process_group(&process_group.pgid());
         }
         drop(session_inner);
+        drop(process_group_inner);
 
         // Create a new process group and insert the process to it
-        let new_process_group = ProcessGroup::new(self, session);
+        let new_process_group = ProcessGroup::new(self, session.clone());
         pid_table.insert_process_group(new_process_group.pgid(), &new_process_group);
-        new_process_group
-            .session()
-            .lock()
-            .insert_process_group(&new_process_group);
+        session.lock().insert_process_group(&new_process_group);
         *process_group_mut = Some(new_process_group);
 
         Ok(())
```

kernel/src/process/process/mod.rs
```diff
@@ -381,73 +379,57 @@ impl Process {
...
             return_errno_with_message!(
                 Errno::EPERM,
                 "a process group leader cannot be moved to a new session"
             );
         }
 
         let mut process_group_mut = self.process_group.lock();
+        self.clear_old_group_and_session(&mut process_group_mut, &mut pid_table);
 
-        self.clear_old_group_and_session(
```
lrh2000, line 398: Why removing the blank line before `clear_old_group_and_session`? I think it should be kept as `process_group_mut` is used twice.

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub(super) fn insert_process(&mut self, pid: Pid, process: &Arc<Process>) {
+        debug_assert!(
+            !self
+                .entries
+                .get(&pid)
+                .is_some_and(|entry| entry.lock().has_live_process())
+        );
+        self.process_count += 1;
+
+        self.get_or_create_entry(pid).lock().set_process(process);
```
lrh2000: All other `insert_xxx` methods use:
```rust
    let entry = self.get_or_create_entry(pid);
    entry.lock().set_xxx();
```

kernel/src/process/wait.rs
```diff
@@ -211,27 +208,22 @@ fn reap_zombie_child(
     let child_process = children_lock.remove(&child_pid).unwrap();
     assert!(child_process.status().is_zombie());
 
+    // Lock order: children of process -> PID table -> tasks of process
```
lrh2000: This lock order comment may be better to place above `for task`?

kernel/src/process/execve.rs
```diff
@@ -228,6 +226,31 @@ fn wait_other_threads_exit(ctx: &Context) -> Result<()> {
...
 
+fn make_current_main_thread(ctx: &Context) {
+    let pid = ctx.process.pid();
+    let old_tid = ctx.posix_thread.tid();
+
+    if old_tid == pid {
+        return;
+    }
+
+    // Lock order: PID table -> tasks of process.
```
lrh2000:
```suggestion
    // Lock order: PID table -> tasks of process
```

kernel/src/process/clone.rs
```diff
@@ -802,25 +801,31 @@ fn set_parent_and_group(clone_flags: CloneFlags, parent: &Arc<Process>, child: &
...
+        let mut pid_table = pid_table::pid_table_mut();
 
-        let process_group_mut = parent.process_group.lock();
-
-        let process_group = process_group_mut.upgrade().unwrap();
-        let mut process_group_inner = process_group.lock();
+        let process_group = {
+            let process_group_mut = parent.process_group.lock();
+            let Some(process_group) = process_group_mut.as_ref() else {
+                // The parent is concurrently moving to another process group.
```
lrh2000: `process_group_mut` cannot be `None` _just_ because it is moving to another process group. It will be `None` only before inserting to the PID table (i.e., during process creation) or after removing from the PID table (after being reaped). I don't think the two cases are possible?

(If `parent` is exiting group, we should reject it above.)

kernel/src/process/pid_table.rs
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    /// Clears the thread reference.
+    fn clear_thread(&mut self) {
+        self.thread = Weak::new();
+    }
+
+    /// Sets the process reference.
+    fn set_process(&mut self, process: &Arc<Process>) {
+        self.process = Arc::downgrade(process);
```
lrh2000, line 303: To make debugging easier in case we mess up the process groups, sessions, threads, etc., can we have:
```rust
    fn set_thread(&mut self, thread: &Arc<Thread>) {
        debug_assert!(self.thread.is_dangling());
```
```rust
    fn clear_thread(&mut self) {
        debug_assert!(!self.thread.is_dangling());
```

For all `set_*`/`clear_*` methods? While reviewing this PR, I realize `Session`/`ProcessGroup` already have similar things.

**EDIT:** Oh actually `make_current_main_thread` will violate this. Perhaps we can have a separate `overwrite_thread` / `replace_thread` / .. so the intention is clear.

In addition, let's order the methods in a consistent way:
 - `insert_xxx`
 - `remove_xxx`
 - (other mutable methods)
 - `get_xxx`
 - `iter_xxx`
 - (other immutable methods)
```patch
diff --git a/kernel/src/process/pid_table.rs b/kernel/src/process/pid_table.rs
index 9489f6d9a..247702222 100644
--- a/kernel/src/process/pid_table.rs
+++ b/kernel/src/process/pid_table.rs
@@ -50,6 +50,7 @@ impl PidTable {
     /// Inserts a thread into the table.
     pub(super) fn insert_thread(&mut self, tid: Tid, thread: &Arc<Thread>) {
         debug_assert_eq!(tid, thread.as_posix_thread().unwrap().tid());
+
         let entry = self.get_or_create_entry(tid);
         entry.lock().set_thread(thread);
     }
@@ -72,13 +73,6 @@ impl PidTable {
         }
     }
 
-    /// Gets a thread by a TID.
-    pub fn get_thread(&self, tid: Tid) -> Option<Arc<Thread>> {
-        self.entries
-            .get(&tid)
-            .and_then(|entry| entry.lock().thread())
-    }
-
     /// Removes a thread from the table and returns it.
     pub(super) fn take_thread(&mut self, tid: Tid) -> Option<Arc<Thread>> {
         let Entry::Occupied(map_entry) = self.entries.entry(tid) else {
@@ -100,6 +94,13 @@ impl PidTable {
         Some(thread)
     }
 
+    /// Gets a thread by a TID.
+    pub fn get_thread(&self, tid: Tid) -> Option<Arc<Thread>> {
+        self.entries
+            .get(&tid)
+            .and_then(|entry| entry.lock().thread())
+    }
+
     /// Returns an iterator over threads that have a live thread reference.
     pub fn iter_threads(&self) -> impl Iterator<Item = Arc<Thread>> + '_ {
         self.entries
```

kernel/src/process/execve.rs
```diff
@@ -228,6 +226,31 @@ fn wait_other_threads_exit(ctx: &Context) -> Result<()> {
...
+    let pid = ctx.process.pid();
+    let old_tid = ctx.posix_thread.tid();
+
+    if old_tid == pid {
+        return;
+    }
+
+    // Lock order: PID table -> tasks of process
+    let mut pid_table = pid_table::pid_table_mut();
+    let mut tasks = ctx.process.tasks().lock();
```
lrh2000, line 242: Why do you remove the original comment?
```rust
    // The current thread is already the main thread.
    if old_tid == pid {
        return;
    }

    // The current thread is not the main thread.

    // Lock order: PID table -> tasks of process
    let mut pid_table = pid_table::pid_table_mut();
    let mut tasks = ctx.process.tasks().lock();
```

tatetian (approved):

PR #2958, opened 2026-02-07
Replace the use of `read_volatile` and `write_volatile` with assembly code

Hsy-Intel:
Related to #2948, #2785.

PR #3022, opened 2026-03-09
Add Memcpy/Memset trait framework for typed memory copies

tatetian:
## Summary

The low-level memory copy infrastructure in OSTD (`memcpy`, `memcpy_fallible`, `memset`, `memset_fallible`) currently lives as four independent private functions inside `mm/io.rs`. Each function hard-codes a single copy strategy, and callers must choose the right one by name. This works today because there are only two memory categories—kernel and user—but it does not scale.

PR #2958 (TDX-aware I/O memory) needs to introduce a third category: **I/O memory** (MMIO). Copies involving I/O memory require different underlying primitives (e.g., single-instruction `MOV` on TDX instead of `rep movsb`), and every combination of source × destination type needs its own dispatch. Extending four ad-hoc functions to handle a 3×3 matrix would produce a combinatorial mess of function names (`memcpy_io_to_kernel`, `memcpy_kernel_to_io`, …) with no compile-time enforcement that the caller picked the right variant.

This PR solves the problem by encoding the memory category in the type system:

```rust
// Marker types — no instances, just compile-time tags
pub(crate) enum KernelMem {}
pub(crate) enum UserMem {}
// pub(crate) enum IoMem {}   ← added by #2958

// One trait per operation, parameterized by Src/Dst
pub(crate) unsafe trait Memcpy<Src> {
    type Result;  // () for infallible, usize for fallible
    unsafe fn memcpy(dst: *mut u8, src: *const u8, len: usize) -> Self::Result;
}

// Callers state intent via type parameters
unsafe { memcpy::<KernelMem, UserMem>(dst, src, len) }
```

The compiler resolves the correct implementation at monomorphization time. Adding `IoMem` in #2958 amounts to defining the enum and writing `unsafe impl Memcpy<IoMem> for KernelMem { ... }`—no changes to existing callers, no new function-name conventions to remember, and the type checker rejects nonsensical combinations that lack an impl.

As a preparatory step, this PR also reorganizes `mm/io.rs` and `mm/io_util.rs` into an `mm/io/` directory so that the new `copy` module sits alongside related code without bloating the main `mod.rs`.

## Changes

**Commit 1 — Reorganize `mm/io.rs` into `mm/io/` directory**

- `mm/io.rs` → `mm/io/mod.rs` (unchanged content)
- `mm/io_util.rs` → `mm/io/util.rs` (import path update)
- All `io_util` references across the crate updated to `io::util`

**Commit 2 — Add `Memcpy`/`Memset` trait framework**

- New file `mm/io/copy.rs` with:
  - `KernelMem`, `UserMem` marker enums
  - `unsafe trait Memcpy<Src>` (4 impls: K→K, U→K, K→U, U→U)
  - `unsafe trait Memset` (2 impls: K, U)
  - `pub(crate)` free functions `memcpy::<Dst, Src>()` and `memset::<Dst>()`
- `mm/io/mod.rs`: removes 4 private functions, imports from `copy`, updates all call sites

### DIFF
--- a/ostd/src/mm/io/copy.rs
+++ b/ostd/src/mm/io/copy.rs
@@ -1,0 +1,213 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Generalized memory copy and fill operations for different memory types.
+//!
+//! This module provides [`Memcpy`] and [`Memset`] traits that generalize
+//! the classic C `memcpy` and `memset` for typed memory categories
+//! ([`Infallible`], [`Fallible`]).
+//!
+//! # Examples
+//!
+//! Infallible kernel-to-kernel copy:
+//!
+//! ```ignore
+//! // SAFETY: Both pointers are valid kernel memory for `len` bytes.
+//! unsafe { memcpy::<Infallible, Infallible>(dst, src, len) };
+//! ```
+//!
+//! Fallible copy from user space into kernel space:
+//!
+//! ```ignore
+//! // SAFETY: `dst` is valid kernel memory; `src` is in user space.
+//! let copied = unsafe { memcpy::<Infallible, Fallible>(dst, src, len) };
+//! if copied < len {
+//!     // A page fault occurred after `copied` bytes.
+//! }
+//! ```
+//!
+//! Fallible zero-fill of user memory:
+//!
+//! ```ignore
+//! // SAFETY: `dst` is in user space for `len` bytes.
+//! let filled = unsafe { memset::<Fallible>(dst, 0u8, len) };
+//! if filled < len {
+//!     // A page fault occurred after `filled` bytes.
+//! }
+//! ```
+
+use super::{Fallible, Infallible};
+use crate::arch::mm::{__memcpy_fallible, __memset_fallible};
+
+/// Copies `len` bytes from `src` to `dst`.
+///
+/// This is a generalized `memcpy` that dispatches
+/// to the appropriate implementation based on the memory types.
+/// Infallible-to-infallible copies return `()`.
+/// Copies involving fallible memory are fallible
+/// and return the number of bytes successfully copied.
+///
+/// # Safety
+///
+/// - `src` must point to a virtual memory region of type `Src`
+///   that allows memory reads.
+/// - `dst` must point to a virtual memory region of type `Dst`
+///   that allows memory writes.
+pub(crate) unsafe fn memcpy<Dst: Memcpy<Src>, Src>(
+    dst: *mut u8,
+    src: *const u8,
+    len: usize,
+) -> Dst::Result {
+    // SAFETY: The safety is upheld by the caller.
+    unsafe { Dst::memcpy(dst, src, len) }
+}
+
+/// Fills `len` bytes of memory at `dst` with the specified `value`.
+///
+/// This is a generalized `memset` that dispatches
+/// to the appropriate implementation based on the memory type.
+/// Infallible memory fills return `()`.
+/// Fallible memory fills return the number of bytes successfully set.
+///
+/// # Safety
+///
+/// - `dst` must point to a virtual memory region of type `Dst`
+///   that allows memory writes.
+pub(crate) unsafe fn memset<Dst: Memset>(dst: *mut u8, value: u8, len: usize) -> Dst::Result {
+    // SAFETY: The safety is upheld by the caller.
+    unsafe { Dst::memset(dst, value, len) }
+}
+
+/// Generalizes `memcpy` for different memory types.
+///
+/// The destination type is `Self` and the source type is `Src`,
+/// following the `memcpy(dst, src, len)` argument convention.
+///
+/// When the source and destination virtual/physical memory regions overlap,
+/// the copy may produce unexpected bytes in the destination.
+/// The exact behavior is up to the implementation.
+///
+/// # Safety
+///
+/// Assuming the caller upholds the preconditions of [`Memcpy::memcpy`],
+/// the memory copy implementation must not cause soundness problems.
+/// The exact safety conditions depend on the type of the source and destination memory.
+/// For example, when both sides are infallible memory,
+/// the implementation must copy exactly `len` bytes from `src` to `dst`.
+pub(crate) unsafe trait Memcpy<Src> {
+    /// The result type of the copy operation.
+    ///
+    /// Infallible copies return `()`.
+    /// Fallible copies return the number of bytes successfully copied
+    /// as `usize`.
+    type Result;
+
+    /// Copies `len` bytes from `src` to `dst`.
+    ///
+    /// # Safety
+    ///
+    /// - `src` must point to a virtual memory region of type `Src`
+    ///   that allows memory reads.
+    /// - `dst` must point to a virtual memory region of type `Self`
+    ///   that allows memory writes.
+    unsafe fn memcpy(dst: *mut u8, src: *const u8, len: usize) -> Self::Result;
+}
+
+/// Generalizes `memset` for different memory types.
+///
+/// # Safety
+///
+/// The implementation must correctly fill `len` bytes at `dst` with `value`
+/// when the caller upholds the preconditions of [`Memset::memset`].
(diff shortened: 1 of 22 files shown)


ostd/src/mm/io/copy.rs
```diff
@@ -50,6 +50,8 @@ use crate::arch::mm::{__memcpy_fallible, __memset_fallible};
 ///   according to the destination memory type `Dst`.
 /// - For [`KernelMem`], "valid" means meeting the [validity requirements].
 /// - For [`UserMem`], the pointer must be in user space for `len` bytes.
+/// - The source and destination ranges must not overlap
+///   in virtual address space.
```
lrh2000: I don't believe this is true. It is impossible to reliably ensure this via safe APIs. How can two `VmReader`/`VmWriter` overlaps be prevented? We can create arbitrary `VmReader`/`VmWrite` for user space pointers anyway.

ostd/src/mm/io/mod.rs
```diff
@@ -860,7 +860,7 @@ impl VmWriter<'_, Fallible> {
         }
 
         // SAFETY: The destination is a subset of the memory range specified by
-        // the current writer, so it is either valid for writing or in user space.
+        // the current writer, so it is in user space.
         let set_len = unsafe { memset::<UserMem>(self.cursor, 0u8, len_to_set) };
```
lrh2000: Since we can convert an infallible `VmReader`/`VmWriter` to a fallible one, I think the previous comment is correct.

However, I am not sure if it is worth repeating in the `UserMem` doc comments, as it seems quite redundant. But if we want to be strict, the following statement (and similar statements in `copy.rs`) is also incorrect:
```c
/// - For [`UserMem`], the pointer must be in user space for `len` bytes.
```

ostd/src/mm/io/copy.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+// SAFETY: Delegates to `volatile_set_memory`,
+// which correctly fills bytes in valid kernel memory.
+unsafe impl Memset for KernelMem {
+    type Result = ();
+
+    unsafe fn memset(dst: *mut u8, value: u8, len: usize) {
+        // SAFETY: The safety is guaranteed by the caller.
```
lrh2000: Though not introduced in this PR, this file has two ways to express the same thing:
```c
        // SAFETY: The safety is guaranteed by the caller.
```
```c
        // SAFETY: The safety is upheld by the caller.
```

Since the first sentence only occurs in two places, it can be unified with the second sentence?

ostd/src/mm/io/copy.rs
```diff
@@ -0,0 +1,225 @@
...
+//!     // A page fault occurred after `copied` bytes.
+//! }
+//! ```
+//!
+//! Fallible zero-fill of user memory:
+//!
+//! ```rust,ignore
+//! // SAFETY: `dst` is in user space for `len` bytes.
+//! let filled = unsafe { memset::<UserMem>(dst, 0u8, len) };
+//! ```
```
lrh2000, line 36: I guess it will not cause harm:
```rust
//! if filled < len {
//!     // A page fault occurred after `filled` bytes.
//! }
```

ostd/src/mm/io/copy.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// Copies `len` bytes from `src` to `dst`.
+    ///
+    /// # Safety
+    ///
+    /// - `src` must be valid for reads of `len` bytes
+    ///   according to the source memory type.
+    /// - `dst` must be valid for writes of `len` bytes
+    ///   according to the destination memory type.
+    /// - The source and destination ranges must not overlap
+    ///   in virtual address space.
```
lrh2000: Same to https://github.com/asterinas/asterinas/pull/3022#discussion_r2909196670.

ostd/src/mm/io/copy.rs
tatetian: I removed the mistaken statement about overlapped memory. The new safety requirements are stated this way:

https://github.com/asterinas/asterinas/blob/484dc1f1df4e8811bc01c1c6b5b7414206cdfda5/ostd/src/mm/io/copy.rs#L49-L55

The handling of overlapped memory is critical to safety. So I added the following paragraph in the safety section for the `Memcpy` trait:

https://github.com/asterinas/asterinas/blob/484dc1f1df4e8811bc01c1c6b5b7414206cdfda5/ostd/src/mm/io/copy.rs#L90-L98

tatetian:
The revised version removed `KernelMem` and `UserMem` because they basically mean the same semantics as `Infallible` and `Fallible`.

ostd/src/mm/io/mod.rs
```diff
@@ -50,10 +50,11 @@ use crate::{
         __atomic_cmpxchg_fallible, __atomic_load_fallible, __memcpy_fallible, __memset_fallible,
     },
     mm::{
-        MAX_USERSPACE_VADDR,
         kspace::{KERNEL_BASE_VADDR, KERNEL_END_VADDR},
+        MAX_USERSPACE_VADDR,
```
lrh2000, line 55: Run `make format` at the first commit to remove such changes.

kernel/src/vm/vmo/mod.rs
```diff
@@ -13,7 +13,7 @@ use core::{
 use align_ext::AlignExt;
 use ostd::{
     mm::{
-        FrameAllocOptions, UFrame, VmIo, VmIoFill, VmReader, VmWriter, io_util::HasVmReaderWriter,
+        FrameAllocOptions, UFrame, VmIo, VmIoFill, VmReader, VmWriter, io::util::HasVmReaderWriter,
```
lrh2000, line 16: `io_util` -> `io::util` should belong to the second commit, not the third. If you want to fix it... Not really a big problem though.

ostd/src/mm/io/copy.rs
```diff
@@ -0,0 +1,215 @@
...
+/// when the caller upholds the preconditions of [`Memcpy::memcpy`].
+///
+/// An implementation must handle the case
+/// where the source and destination virtual memory regions overlap.
+/// When they do,
+/// a naïve forward copy would clobber source bytes before they are read.
+/// The implementation may need to copy in the backward direction
+/// to preserve correctness.
+/// For reference, compare [`core::ptr::copy_nonoverlapping`] (assumes no overlap)
+/// with [`core::ptr::copy`] (handles overlap).
```
lrh2000: When memory overlaps, `__memcpy_fallible` does not guarantee correctness; it only ensures that there are no soundness problems:
https://github.com/asterinas/asterinas/blob/417d58dbac51a71cbe02b5615a7198184371ebd6/ostd/src/mm/io.rs#L281-L284

Does it meet the rule here? It does not seem so. I think we should either (1) add a FIXME for `__memcpy_fallible` or (2) change the rules and document the expected behavior.

ostd/src/mm/io/copy.rs
```diff
@@ -0,0 +1,215 @@
...
+    ///
+    /// # Safety
+    ///
+    /// - `dst` must point to a virtual memory region of type `Self`
+    ///   that allows memory writes.
+    unsafe fn memset(dst: *mut u8, value: u8, len: usize) -> Self::Result;
+}
+
+// SAFETY: Delegates to `volatile_copy_memory`,
+// which correctly copies bytes between valid kernel pointers.
```
lrh2000: For consistency:
```suggestion
// which correctly copies bytes between valid kernel memory regions.
```

https://github.com/asterinas/asterinas/blob/484dc1f1df4e8811bc01c1c6b5b7414206cdfda5/ostd/src/mm/io/copy.rs#L183-L184

ostd/src/mm/io/copy.rs
lrh2000: To make the code snippet render, you can convert "tatetian" to "asterinas." I edited your comment.

ostd/src/mm/io/copy.rs
tatetian: I changed the rules in the documentation.

Issue #2423, opened 2025-09-11
Basic implementation of `bpf` syscall


marcin-makulec:
_To preface the proposal - I want to implement it myself and become a contributor. I am obviously open for communication, cooperation and I want to consult subsytem architecture etc._

### Feature Description

A limited implementation of the `bpf` syscall, designed in such a way, that adding other features wouldn't require painful refactors in the future. My proposal is to implement loading [`BPF_PROG_TYPE_SOCKET_FILTER`](https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_SOCKET_FILTER/) BPF programs (these are the original packet filters), attaching them to sockets and running them via interpreter (no JIT at first - maybe no eBPF instructions, only cBPF?).

### Problem or Need

Implementing the most basic `bpf` functionality now will be very helpful when other kernel features are implemented, as there would be a steady infrastructure for extending the BPF subsystem with required features. For example, when cgroups are fully implemented, it will be sufficient to add BPF-cgroups-programs related features to the subsystem. Also, network filtering, tracing and orchestration would be very valuable if cloud-based VMs are the primary target - this base work would enable implementing these features in the future. 

### Suggested Implementation

I have tried for a long time to draw a diagram that would encompass all BPF related concepts in Linux - this is work in progress. However, although enumerating these concepts and how they relate is a necessary preliminary work (we don't want to make assumptions be stuck with a huge refactor in the future), `BPF_PROG_TYPE_SOCKET_FILTER` programs do not require many BPF features - I think it works in proposal's favor. I will try to check it as thoroughly as I can, but at this moment I think that `bpf()` [commands](https://elixir.bootlin.com/linux/v6.16/source/kernel/bpf/syscall.c#L5797) which are strictly necessary for this limited implentation are:
- `BPF_PROG_LOAD`
- `BPF_PROG_TEST_RUN`
- `BPF_PROG_GET_FD_BY_ID`
 
Obviously, there are many questions around the topic. What about maps, tokens, verifier, JIT, eBPF vs cBPF, ...? I am ready to provide my take on how a roadmap could look like, however, I am not sure if my ideas are compatible with community's plans and would like to discuss the basics first.

### Possible Alternatives

[...]

PR #3024, opened 2026-03-09
Seccomp syscall

marcin-makulec:
This is a prototype, code needs more cleanup, documentation and testing. Strict mode has been verified and works, but proper unit tests should be added. Filter mode is in infancy, there is only untested filter loading and verification. Work is in progress, but I publish it now, so I can get some directions whether I am doing things right.

There is inconsistency as some seccomp specific (in contrast to netfilter) functions are isolated, and some are not (mainly the loading process). Common responsibilities should be seperated into non-seccomp-specific functions and modules in order to not accumulate technical debt before we add netfilter cBPF - at this point it's a simple refactor.

cBPF opcodes may need to have a more readable representation + better naming convention for cBPF related structs.

`num_enum` crate has been added to make opcode verification easier (via derived `TryFromPrimitive` trait). Perhaps it should be removed, maybe we can just expand the resulting macro and remove the crate.

In strict mode, prohibited syscalls should kill only the calling thread, not the entire process (I was not sure how to achieve that).

Structs and the seccomp field in the Task struct should have documentation and not be `pub` - instead, proper setters and getters should be implemented.

The seccomp field is a SpinLock - if by chance it's the proper way to handle interior mutability (ctx.current is immutable in syscall context, but we need to update the seccomp state of a thread when handling seccomp) then great, but I mostly tried to "hack" it as I was not sure what is the best practice.

I may have missed some details or potential issues with my code - will try to verify it later and post a comment about it. I will do the same with relevant findings and plan of action from #2423.

@tatetian sorry for the long wait, but I am back!


marcin-makulec:
Pasting the plan of action from #2423 I mentioned before

# Roadmap (in progress?)

## `seccomp` operations
- [x] `SECCOMP_SET_MODE_STRICT` - allows `read()`, `write()`, `_exit()` and `sigreturn()`
- [ ]  add `prctl(PR_SET_SECCOMP, SECCOMP_MODE_STRICT)` - does the same thing

### `SECCOMP_SET_MODE_FILTER` - assign a BPF filter to the thread
- [ ] `prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, args)` - does the same thing but without flags
- [x] cBPF verifier
- [x] `seccomp` verifier
- [ ] eBPF transpiler
- [ ] eBPF verifier (for the time being, it should reject non-seccomp features, those will be implemented and allowed later)
- [ ] JIT (same as above)
- [ ] runtime
  - [ ] registers
  - [ ] stack frame
  - [ ] interpreter
  - [ ] machine code runtime
  - [ ] ... (this list will be much longer when implementing BPF proper, not sure if I missed something for `seccomp`)

#### `_MODE_FILTER` flags
- [ ] `SECCOMP_FILTER_FLAG_LOG` - log every filter return action
- [ ] `_NEW_LISTENER` - return a user-space notification file descriptor (see `SECCOMP_RET_USER_NOTIF`)
- [ ] `_SPEC_ALLOW` - disable Speculative Store Bypass mitigation
- [ ] `_TSYNC` - sync all process threads to the same `seccomp` filter tree

#### Filter return value handling
- [ ] `SECCOMP_RET_KILL_PROCESS` - do not execute syscall, terminate process and all threads in thread group, core dump)
- [ ] `SECCOMP_RET_KILL_THREAD` / `SECCOMP_RET_KILL` - terminate thread, core dump if single-threaded
- [ ] `SECCOMP_RET_TRAP` send SIGSYS to thread - syscall not executed, filter returns data via signal, continue execution
- [ ] `SECCOMP_RET_ERRNO` - do not execute syscall, filter return data in `errno`
- [ ] `SECCOMP_RET_USER_NOTIF` - handle syscall with user-space supervisor
  - [ ] `seccomp_unotify()`
- [ ] `SECCOMP_RET_TRACE` - notify tracer (executes syscall only if a tracer is present)
- [ ] `SECCOMP_RET_LOG` - log filter action, then execute syscall
- [ ] `SECCOMP_RET_ALLOW` - execute syscall

## other `seccomp` operations
- [ ] `SECCOMP_GET_ACTION_AVAIL` - check if action is available
- [ ] `SECCOMP_GET_NOTIF_SIZES` - get sizes of the seccomp user-space notification structures

## misc
- [ ] `prctl(PR_SET_NO_NEW_PRIVS, 1)`
- [ ] `/proc/sys/kernel/seccomp/actions_avail`
- [ ] `/proc/sys/kernel/seccomp/actions_logged`
- [ ] logging
- [ ] add missing error codes

chen1plus:
Hi @marcin-makulec,

I'm working on a university project to implement a Netfilter syscall using eBPF in Asterinas. I’d love to build on your BPF infrastructure and collaborate to ensure our work is complementary.

Regarding the eBPF VM, since it’s currently missing, I suggest we integrate the [rbpf](https://github.com/qmonnet/rbpf) crate instead of building one from scratch. This would allow us to focus entirely on system integration rather than distracted by VM implementation.

Let me know your thoughts!

marcin-makulec:
@chen1plus I am really happy to hear that you would like to cooperate on this. I propose that we both work on this branch of my fork and here is my idea how we go about this:
- we need to figure out common elements for netfilter and seccomp, refactor those I already implemented and missing ones,
- then we could split the branch into 2 if you'd like to do so,
- communicate with each other during development so our features are consistent with each other.

Would you like to arrange a call to talk through the features and set up a development plan? You can send me you Discord or whatever you prefer to m.makulec321@gmail.com and we can set up a date then (keep in mind I am 6 time zones "earlier").

Also, do you have any kind of list of required features? I did some work on preparing such a list for seccomp, but maybe there are netfilter elements that would influence what proper code structure would be.

PR #3026, opened 2026-03-10
Do not increase RSS when COW

vvvvsv:
The RSS of a `Vmar` should not be increased when a COW happens.

Linux re-classify the page from `RssFile` to `RssAnon` in the RSS counter when a COW on a file-backed mapping happens. This PR currently does not support this re-classification, since it will introduce extra complexity when unmapping the page.

The RSS count in Linux may differ slightly from the expected value, because its per-CPU counters are updated lazily. This PR relaxes the restriction in `mmap_vmrss` on Linux, allowing it to pass on Linux.

### DIFF
--- a/kernel/src/vm/vmar/vm_mapping.rs
+++ b/kernel/src/vm/vmar/vm_mapping.rs
@@ -166,8 +166,8 @@ impl VmMapping {
     /// Returns the mapping's RSS type.
     pub fn rss_type(&self) -> RssType {
         match &self.mapped_mem {
-            MappedMemory::Anonymous => RssType::RSS_ANONPAGES,
-            MappedMemory::Vmo(_) | MappedMemory::Device => RssType::RSS_FILEPAGES,
+            MappedMemory::Anonymous => RssType::Anon,
+            MappedMemory::Vmo(_) | MappedMemory::Device => RssType::File,
         }
     }
 
@@ -441,7 +441,10 @@ impl VmMapping {
                         cursor.unmap(PAGE_SIZE);
                         cursor.jump(va.start).unwrap();
                         cursor.map(new_frame.into(), prop);
-                        rss_delta.add(self.rss_type(), 1);
+                        // FIXME: Linux re-classifies the page from `File` to `Anon` in RSS,
+                        // when a COW on a file-backed mapping happens.
+                        // We currently do not support this re-classification,
+                        // since it will introduce some complexity when unmapping this page.
                     }
                     cursor.flusher().sync_tlb_flush();
                 }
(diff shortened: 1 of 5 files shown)


test/initramfs/src/apps/memory/mmap/mmap_vmrss.c
```diff
@@ -68,6 +76,10 @@ long get_vm_rss_kb(rss_type type)
 	return rss_kb;
 }
 
+// Tests whether the return value is within the expected range.
+#define TEST_RES_EXPECTED(expr, expected) \
+	TEST_RES(expr, (labs(_ret - (expected)) <= RSS_DEVIATION_KB))
```
lrh2000: Given that it hard-codes RSS_DEVIATION_KB, we may want to name this `TEST_RSS`, as it is unusable in other types of variables' expected ranges.

In addition, how is `RSS_DEVIATION_KB` selected? By experiments or by reasoning? Are there any experimental data or theoretical support from Linux codebase?

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -439,7 +439,10 @@ impl VmMapping {
                         cursor.unmap(PAGE_SIZE);
                         cursor.jump(va.start).unwrap();
                         cursor.map(new_frame.into(), prop);
-                        rss_delta.add(self.rss_type(), 1);
+                        // FIXME: Linux re-classify the page from `File` to `Anon` in RSS,
```
lrh2000: classify -> classifies?

This also applies to the test file:
```c
	// Linux re-classify a page from `RssFile` to `RssAnon`,
```

In addition, use one of Linux's name or Asterinas' name? I don't think the keyword `RssFile` appears in either our codebase or Linux's codebase.

test/initramfs/src/apps/memory/mmap/mmap_vmrss.c
vvvvsv: > how is `RSS_DEVIATION_KB` selected? By experiments or by reasoning?

I selected this value by experiments when opening this PR.

I just double-checked the Linux source. In the single-threaded case, the deviation of the per-CPU counter is `max(32, nr_online_cpus * 2)`.

https://elixir.bootlin.com/linux/v6.16.5/source/lib/percpu_counter.c#L258-L264
https://elixir.bootlin.com/linux/v6.16.5/source/lib/percpu_counter.c#L126-L134

Then `RSS_DEVIATION_KB` should be `max(32, nr_online_cpus * 2) * 2 * (PAGE_SIZE / 1024)`. This is because there are two RSS counters (anon and file), and the RSS counters track the number of pages.

vvvvsv:
@boterinas rerun

lrh2000 (approved):

PR #3027, opened 2026-03-10
Fix some bugs in the VFS layer

cchanging:
## Fix dup2 and dup3 error semantics

In Asterinas, `dup2`/`dup3` first close `new_fd`, then check/copy `old_fd`; if `old_fd` is invalid, Linux requires `new_fd` to remain unchanged, but currently it would be incorrectly closed. In addition, `FileTable::dup()` returns `ENOENT` for a bad fd, while Linux should return `EBADF`.

## Fix POSIX record lock close semantics

Currently there are some bugs in `fcntl` record lock: first, `set_end()` modifies a copy rather than the actual internal range; second, range locks are only released when `InodeHandle` is dropped for the last time, whereas Linux [releases such locks as soon as any fd referencing the inode is closed by the same process](https://man7.org/linux/man-pages/man2/fcntl_locking.2.html):

> If a process closes any file descriptor referring to a file, then all of the process's locks on that file are released 

## Fix a bug of move mount operation

`move_mount_to()` does not prevent grafting a mount into its own subtree, which could theoretically create a mount tree cycle

### DIFF
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -407,6 +407,17 @@ impl Path {
                 "the destination path is not in this mount namespace"
             );
         }
+        if dst_path
+            .mount_node()
+            .is_equal_or_descendant_of(self.mount_node())
+        {
+            // Reject moves that would place a mount beneath itself, because the mount tree
+            // must remain acyclic.
+            return_errno_with_message!(
+                Errno::ELOOP,
+                "the destination path is inside the mount subtree being moved"
+            );
+        }
 
         self.mount.graft_mount_tree(dst_path);
 
--- a/kernel/src/fs/vfs/path/mount.rs
+++ b/kernel/src/fs/vfs/path/mount.rs
@@ -514,6 +514,21 @@ impl Mount {
         self.parent.read().as_ref().cloned()
     }
 
+    /// Returns whether `self` is `ancestor` or a descendant of it in the mount tree.
+    pub(super) fn is_equal_or_descendant_of(&self, ancestor: &Arc<Self>) -> bool {
+        let mut current = self.this();
+        loop {
+            if Arc::ptr_eq(&current, ancestor) {
+                return true;
+            }
+
+            let Some(parent) = current.parent().and_then(|parent| parent.upgrade()) else {
+                return false;
+            };
+            current = parent;
+        }
+    }
+
     /// Gets the associated mount namespace.
     pub(super) fn mnt_ns(&self) -> &Weak<MountNamespace> {
         &self.mnt_ns
--- a/test/initramfs/src/apps/fs/Makefile
+++ b/test/initramfs/src/apps/fs/Makefile
@@ -5,6 +5,7 @@ SUBDIRS := \
 	fdatasync \
 	inotify \
 	isolation \
+	mount \
 	overlayfs \
 	procfs \
 	pseudofs \
--- a/test/initramfs/src/apps/fs/mount/Makefile
+++ b/test/initramfs/src/apps/fs/mount/Makefile
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: MPL-2.0
+
+include ../../common/Makefile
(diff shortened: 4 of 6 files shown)


kernel/src/fs/file/file_table.rs
```diff
@@ -56,18 +52,46 @@ impl FileTable {
...
+        &mut self,
+        fd: FileDesc,
+        new_fd: FileDesc,
+        flags: FdFlags,
+    ) -> Result<Option<Arc<dyn FileLike>>> {
+        let entry = self.duplicate_entry(fd, flags)?;
+        let closed_file = self.close_file(new_fd);
+        self.table.put_at(new_fd as usize, entry);
+        Ok(closed_file)
+    }
```
StevenJiang1110, line 77: In this PR, `dup_to` and `dup` select `new_fd` in different ways: `dup_to` picks the fd that is exactly `new_fd`, while dup picks the lowest available fd that is greater than or equal to new_fd. However, I don't see this behavior mentioned in the PR description — is this an intentional change?

If this does reflect the correct Linux behavior, I would suggest renaming the methods to make the semantics clearer: `dup_to` could be renamed to `dup_exact`, and `dup` could be renamed to `dup_ceil`, to better convey the distinction between the two. It would also be helpful to add appropriate documentation for both methods.

kernel/src/fs/file/file_table.rs
```diff
@@ -34,11 +34,7 @@ impl FileTable {
     }
 
     pub fn dup(&mut self, fd: FileDesc, new_fd: FileDesc, flags: FdFlags) -> Result<FileDesc> {
```
StevenJiang1110: This is not an issue introduced by this PR, but I'd like to flag it here: the signature of the `dup` method is incorrect — it should return a `FileLike` along with `new_fd`. The underlying problem is that we cannot drop a `FileLike` while holding the file table's lock, since the drop implementation of `FileLike` may invoke sleeping operations, and `FileTable` is protected by a `RwLock`. Dropping inside the lock would break atomic mode.

You can refer to the implementation of `sys_close` as a correct example of how this should be handled.

I'm not sure whether this PR is the right place to fix this issue, but since the semantics of `dup_to` are already being corrected here, it would be ideal to address this problem in the same PR.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -179,7 +179,7 @@ impl InodeHandle {
         range_lock_list.set_lock(lock, is_nonblocking)
     }
 
-    fn release_range_locks(&self) {
+    pub(crate) fn release_range_locks(&self) {
```
StevenJiang1110: `pub(crate)` does not make sense for kernel crate. Just use `pub` here.

kernel/src/syscall/dup.rs
```diff
@@ -68,8 +68,7 @@ fn do_dup3(
 
     let file_table = ctx.thread_local.borrow_file_table();
     let mut file_table_locked = file_table.unwrap().write();
-    let _ = file_table_locked.close_file(new_fd);
-    let new_fd = file_table_locked.dup(old_fd, new_fd, flags)?;
+    let _ = file_table_locked.dup_to(old_fd, new_fd, flags)?;
```
StevenJiang1110: Similar to what I said above, you should drop `FileLike` without holding the lock of `FileTable`.

kernel/src/fs/file/file_table.rs
```diff
@@ -93,8 +117,9 @@ impl FileTable {
             .collect();
 
         for fd in closed_fds {
-            let removed_entry = self.table.remove(fd as usize).unwrap();
-            closed_files.push(removed_entry.file);
+            if let Some(file) = self.close_file(fd) {
+                closed_files.push(file);
+            }
```
StevenJiang1110: `close_file` cannot return None here, right? If so, using `unwarp` may be the better choice.

test/initramfs/src/apps/io/file_io/fcntl_lock.c
```diff
@@ -0,0 +1,97 @@
...
+FN_SETUP(create)
+{
+	int fd = CHECK(open_test_file());
+	CHECK(close(fd));
+}
+END_SETUP()
+
+FN_TEST(unlock_middle_range)
+{
+	int fd = CHECK(open(TEST_FILE, O_RDWR));
```
StevenJiang1110: In `FN_TEST`, use `TEST_SUCC` instead of `CHECK` if possible.

PR #3028, opened 2026-03-10
Change `FileDesc` to `u32`

chen1plus:
Closes https://github.com/asterinas/asterinas/issues/2907

### DIFF
--- a/kernel/src/fs/file/file_table.rs
+++ b/kernel/src/fs/file/file_table.rs
@@ -160,7 +160,7 @@ impl FileTable {
             .table
             .get(fd.into())
             .map(|entry| entry.file.clone())
-            .ok_or(Error::with_message(Errno::EBADF, "fd does not exist"))?;
+            .ok_or(Error::with_message(Errno::EBADF, "the FD does not exist"))?;
         Ok(FileTableEntry::new(file, flags))
     }
 
@@ -215,19 +215,19 @@ impl FileTable {
         self.table
             .get(fd.into())
             .map(|entry| entry.file())
-            .ok_or(Error::with_message(Errno::EBADF, "fd not exits"))
+            .ok_or(Error::with_message(Errno::EBADF, "the FD does not exist"))
     }
 
     pub fn get_entry(&self, fd: FileDesc) -> Result<&FileTableEntry> {
         self.table
             .get(fd.into())
-            .ok_or(Error::with_message(Errno::EBADF, "fd not exits"))
+            .ok_or(Error::with_message(Errno::EBADF, "the FD does not exist"))
     }
 
     pub fn get_entry_mut(&mut self, fd: FileDesc) -> Result<&mut FileTableEntry> {
         self.table
             .get_mut(fd.into())
-            .ok_or(Error::with_message(Errno::EBADF, "fd not exits"))
+            .ok_or(Error::with_message(Errno::EBADF, "the FD does not exist"))
     }
 
     pub fn fds_and_files(&self) -> impl Iterator<Item = (FileDesc, &'_ Arc<dyn FileLike>)> {


tatetian:
The commit messages need to follow the [Git guideline](https://asterinas.github.io/book/to-contribute/coding-guidelines/git-guidelines.html#imperative-subject).

kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
```diff
@@ -106,7 +114,9 @@ impl<T: FdOps> DirOps for FdDirOps<T> {
             let child = child.downcast_ref::<T::NodeType>().unwrap();
             let child_ops = T::ref_from_inode(child);
 
-            let Ok(file) = file_table.get_file(child_ops.file_desc()) else {
+            let Ok(file) =
+                file_table.get_file(FileDesc::new(child_ops.file_desc().cast_unsigned()))
+            else {
```
StevenJiang1110: `FileDesc::new` will panic if `child_ops.file_desc().cast_unsigned()` is out of range. It would be safer to use `child_ops.file_desc().cast_unsigned().try_into()` here, consistent with how it is handled in other places.

kernel/src/fs/fs_impls/pseudofs/nsfs.rs
```diff
@@ -260,7 +260,7 @@ fn open_ns_as_file<T: NsCommonOps>(ns: &Arc<T>) -> Result<RawFileDesc> {
     let mut file_table = file_table_ref.unwrap().write();
     let fd = file_table.insert(Arc::new(inode_handle), FdFlags::CLOEXEC);
 
-    Ok(fd)
+    Ok(fd.get() as _)
```
StevenJiang1110: I think this method should return `FileDesc`, leaving it to the caller to convert it to `RawFileDesc` when necessary. My reasoning is that `FileDesc` should be preferred wherever possible, as it is more type-safe. The conversion to `RawFileDesc` should only be performed when strictly required, such as at the syscall boundary or when dictated by certain trait method signatures.

kernel/src/syscall/getdents64.rs
```diff
@@ -29,7 +29,10 @@ pub fn sys_getdents(
     );
 
     let mut file_table = ctx.thread_local.borrow_file_table_mut();
-    let file = get_file_fast!(&mut file_table, fd);
+    let file = get_file_fast!(
+        &mut file_table,
+        fd.cast_unsigned().try_into().map_err(|_| Errno::EBADF)?
+    );
```
StevenJiang1110: It looks like the code pattern is duplicated for too many times. I suggest introducing a new macro `get_file_fast_by_rawfd!` to simplify it.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -717,17 +723,17 @@ enum FsPathInner<'a> {
     /// The path of the current working directory.
     Cwd,
     /// A relative path from the directory FD (dirfd).
-    FdRelative(FileDesc, &'a str),
+    FdRelative(RawFileDesc, &'a str),
```
StevenJiang1110: Instead of using `RawFileDesc` here, I think you should introduce a new type safe `DirFd` here.

```rust
pub enum DirFd {
     Cwd, 
     Fd(FileDesc),
}

impl TryFrom<RawFileDesc> for DirFd {
     //.......
}
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -717,17 +723,17 @@ enum FsPathInner<'a> {
...
     /// The path of the FD.
-    Fd(FileDesc),
+    Fd(RawFileDesc),
```
StevenJiang1110: `FileDesc` should be used here?

lrh2000:
So, is it possible to avoid repeating the pattern `cast_unsigned().try_into().map_err(|| Errno::EBADF)`?

https://github.com/asterinas/asterinas/pull/3028#discussion_r2951793958 is a possible approach, but I think a better way is to define `struct FileDesc(RangedU32)`, so we can implement `TryFrom<RawFileDesc>` with a custom error type and implement something like `TryFrom<FileDescTryFromError> for Error` (see [`kernel/src/error.rs`](https://github.com/asterinas/asterinas/blob/f6c07e24a15ad1c92fcbe44765b2aa0d68805ef8/kernel/src/error.rs) for examples).
 - We can use `FileDesc::try_from(raw_fd)?` afterwards.

---

By the way, I think the panicking version of `RangedU32::new` is too risky. It's not good practice. The equivalent function in the Rust standard library, [`NonZeroU32::new`](https://doc.rust-lang.org/std/num/type.NonZeroU32.html#method.new), returns an `Option`, deferring the error handling to the caller. However, fixing this would require a separate PR, as it is unrelated to the changes in this one.

At least let's avoid repeating the same mistake when defining `FileDesc` by hiding the `new` method entirely (it doesn't seem useful to me; we want to use `try_from` instead) or defining a fallible version.
 - Then, hopefully, we won't suffer from similar comments to https://github.com/asterinas/asterinas/pull/3028#discussion_r2951711365 in future.

kernel/src/syscall/fcntl.rs
```diff
@@ -44,8 +45,8 @@ fn handle_dupfd(fd: FileDesc, arg: u64, flags: FdFlags, ctx: &Context) -> Result
     let new_fd = file_table
         .unwrap()
         .write()
-        .dup_ceil(fd, arg as FileDesc, flags)?;
-    Ok(SyscallReturn::Return(new_fd as _))
+        .dup_ceil(fd, FileDesc::new(arg as _), flags)?;
```
tatetian: **Kernel panic on user input**: `arg` is a `u64` from the syscall argument. `FileDesc::new()` panics when the value exceeds `i32::MAX`. A user can trigger a kernel panic with e.g. `fcntl(fd, F_DUPFD, 0x80000000)`. This should use the fallible `try_from` path, consistent with how other syscalls in this PR validate their fd arguments:

```rust
.dup_ceil(fd, FileDesc::try_from(arg as u32).map_err(|_| Errno::EINVAL)?, flags)?;
```

(Note: Linux returns `EINVAL` for invalid `F_DUPFD` arg, not `EBADF`.)

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +15,8 @@ use crate::{
     },
 };
 
-pub type FileDesc = i32;
+pub type FileDesc = RangedU32<0, { i32::MAX as _ }>;
+pub type RawFileDesc = i32;
```
tatetian: **Implementing `TryFrom<RawFileDesc>` for `FileDesc`**: The `RawFileDesc` → `FileDesc` conversion pattern `fd.cast_unsigned().try_into().map_err(|_| Errno::EBADF)?` is repeated ~40 times across the syscall layer. Implement a direct conversion to reduce boilerplate and centralize the validation:

```rust
impl TryFrom<RawFileDesc> for FileDesc {
    type Error = Error;

    fn try_from(raw: RawFileDesc) -> Result<Self> {
        Self::try_from(raw.cast_unsigned())
            .map_err(|_| Error::new(Errno::EBADF))
    }
}
```

This would turn each conversion site into `FileDesc::try_from(raw_fd)?` -- shorter, idiomatic, and harder to get wrong.

Similarly, implement `From<FileDesc> for i32` for the reverse direction. The `fd.get() as _` pattern appears ~23 times (e.g., `writer.write_val::<i32>(&(fd.get() as _))?` in `ctrl_msg.rs:145`), and `as _` silently performs a potentially lossy conversion. With an explicit `From` impl:

```rust
impl From<FileDesc> for i32 {
    fn from(fd: FileDesc) -> i32 {
        fd.get() as i32
    }
}
```

the conversion is guaranteed lossless (since `FileDesc` values are in `0..=i32::MAX`) and call sites become clearer:

```rust
writer.write_val::<i32>(&fd.into())?;
```

Also implement `From<FileDesc> for usize` for the same reason. Several `FileTable` methods use `fd.get() as _` to index into the underlying `SlotVec` (e.g., `self.table.get(fd.get() as _)` on lines 133, 121, 128). With `From<FileDesc> for usize` these become `self.table.get(fd.into())`, again eliminating the `as` cast.

kernel/src/fs/file/file_table.rs
```diff
@@ -60,7 +61,7 @@ impl FileTable {
 
         let min_free_fd = get_min_free_fd();
         self.table.put_at(min_free_fd, entry);
-        Ok(min_free_fd as FileDesc)
+        Ok(FileDesc::new(min_free_fd as _))
```
tatetian: [**`checked-arithmetic`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/variables-expressions-and-statements.html#checked-arithmetic): `min_free_fd` is a `usize` from `SlotVec` indexing. While resource limits make overflow practically impossible, `FileDesc::new()` will panic if the index exceeds `i32::MAX`. The same applies to `FileDesc::new(self.table.put(entry) as _)` on line 90 and similar sites in this file. Consider using `FileDesc::try_from(...).map_err(...)` to stay consistent with the fallible conversion pattern used elsewhere, or at minimum add a comment explaining why the panicking path is safe (e.g., resource limits guarantee the table never exceeds `i32::MAX` entries).

kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
```diff
@@ -106,7 +114,9 @@ impl<T: FdOps> DirOps for FdDirOps<T> {
             let child = child.downcast_ref::<T::NodeType>().unwrap();
             let child_ops = T::ref_from_inode(child);
 
-            let Ok(file) = file_table.get_file(child_ops.file_desc()) else {
+            let Ok(file) =
+                file_table.get_file(FileDesc::new(child_ops.file_desc().cast_unsigned()))
```
tatetian: **Roundtripping through `RawFileDesc` is fragile**: `file_desc()` returns `RawFileDesc` (i32) that was originally converted from `FileDesc` via `.get() as _` (line 201). This roundtrip is correct today but silently depends on the invariant that only valid fds are stored. **Since `FdOps` is a trait with `file_desc(&self) -> RawFileDesc`, consider changing the trait to return `FileDesc` directly** -- both implementors (`FileSymOps`, `FileInfoOps`) store values that originated as valid file descriptors, so they could store `FileDesc` instead of `RawFileDesc`, eliminating the roundtrip conversions and the panicking `FileDesc::new()` calls.

kernel/src/net/socket/unix/ctrl_msg.rs
```diff
@@ -101,7 +101,11 @@ impl FileMessage {
         let mut file_table = current.as_thread_local().unwrap().borrow_file_table_mut();
         for _ in 0..nfiles {
             let fd = reader.read_val::<i32>()?;
-            let file = get_file_fast!(&mut file_table, fd).into_owned();
+            let file = get_file_fast!(
+                &mut file_table,
+                fd.cast_unsigned().try_into().map_err(|_| Errno::EBADF)?
+            )
+            .into_owned();
```
tatetian: [**`explain-variables`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/variables-expressions-and-statements.html#explain-variables): Embedding the `RawFileDesc` → `FileDesc` conversion inside the macro call hurts readability and chains the `?` operator inside a macro argument. Pull the conversion into its own line (this applies to all ~40 similar sites in the PR, and pairs well with the `TryFrom<RawFileDesc>` impl suggested above).

[**`accurate-names`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#accurate-names): Throughout the PR, variables of type `RawFileDesc` are named `fd`, which is the same name used for validated `FileDesc` values. This makes it easy to confuse the two at a glance. Rename unvalidated parameters/variables to `raw_fd` so the distinction is visible at every use site:

```rust
let fd = FileDesc::try_from(raw_fd)?;
let file = get_file_fast!(&mut file_table, fd)
    .into_owned();
```

The same argument applies to function arguments. If a function argument is of type `RawFileDesc`, it should be named `raw_fd`, not `fd`.

kernel/src/process/process_filter.rs
```diff
@@ -35,7 +35,11 @@ impl ProcessFilter {
                 };
                 let file = {
                     let mut file_table = ctx.thread_local.borrow_file_table_mut();
-                    get_file_fast!(&mut file_table, fd).into_owned()
+                    get_file_fast!(
+                        &mut file_table,
+                        fd.cast_unsigned().try_into().map_err(|_| Errno::EBADF)?
+                    )
+                    .into_owned()
```
tatetian: [**`explain-variables`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/variables-expressions-and-statements.html#explain-variables): The `let fd = { ... }` block already exists to validate the raw fd -- the `FileDesc` conversion should happen there rather than being deferred into the macro call. With the `TryFrom<RawFileDesc>` impl suggested above, this simplifies to:

```rust
let fd = {
    let raw_fd = id.cast_signed();
    if raw_fd < 0 {
        return_errno_with_message!(Errno::EINVAL, "the pidfd is invalid");
    }
    FileDesc::try_from(raw_fd)?
};
let file = {
    get_file_fast!(&mut file_table, fd).into_owned()
};
```

StevenJiang1110:
> but I think a better way is to define struct FileDesc(RangedU32)

I agree. Compared with the approach in https://github.com/asterinas/asterinas/pull/3028#discussion_r2951793958, this approach would also help reduce duplicate code outside the `get_file_fast` macros.

> By the way, I think the panicking version of RangedU32::new is too risky. It's not good practice.

I share the same concern. It is difficult to reason about whether every call site of `RangedU32::new` is free from out-of-range panics. Though this can be addressed separately in another PR.

kernel/src/syscall/getdents64.rs
tatetian: I am bothered by this duplication, too. But I suggested adding `FileDesc::try_From`.

```rust
let fd = FileDesc::try_from(raw_fd)?;
let file = get_file_fast!(&mut file_table, fd);
```

This is cleaner and more flexible than adding a new macro.

chen1plus:
I appreciate your patience. I'll go through the suggestions and make updates accordingly.

chen1plus:
Hey team, quick question regarding integer types: are we ever planning to support 16-bit platforms?
If not, I’d like to assume `u32` fits into `usize` to avoid messy `try_into()` calls.

tatetian:
> Hey team, quick question regarding integer types: are we ever planning to support 16-bit platforms? If not, I’d like to assume `u32` fits into `usize` to avoid messy `try_into()` calls.

Only modern 64-bit CPU architectures will be supported.

kernel/src/net/socket/unix/ctrl_msg.rs
chen1plus: > Embedding the `RawFileDesc` → `FileDesc` conversion inside the macro call hurts readability.

As `TryFrom<RawFileDesc>` is implemented now, it becomes

```rs
get_file_fast!(&mut file_table, fd.try_into()?)
```

Do you think it clear enough?

> It chains the `?` operator inside a macro argument.

The extended code would be like

```rs
{
    // ...
    let file_desc: FileDesc = fd.try_into()?;
    // ...
}
```

I think that it is not a problem.

chen1plus:
Updated, thanks!

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +15,89 @@ use crate::{
...
+}
+
+impl TryFrom<usize> for FileDesc {
+    type Error = Error;
+
+    fn try_from(value: usize) -> Result<Self> {
+        Ok(Self(
+            RangedU32::try_from(value as u32).map_err(|_| Errno::EBADF)?,
+        ))
+    }
```
tatetian: **Silent truncation of values above `u32::MAX`**: Both `TryFrom<u64>` and `TryFrom<usize>` cast `value as u32` before passing to `RangedU32::try_from`. This silently discards the upper 32 bits. For example, `FileDesc::try_from(0x1_0000_0003u64)` truncates to `3u32`, passes the range check, and returns fd 3 instead of `EBADF`.

This is exercised in `handle_dupfd` (`fcntl.rs`), where `arg: u64` from the syscall is converted via `arg.try_into()`. A userspace caller could pass `0x1_0000_0003` as the `arg` to `F_DUPFD` and it would silently become fd 3. Fix:

```rust
impl TryFrom<u64> for FileDesc {
    type Error = Error;
    fn try_from(value: u64) -> Result<Self> {
        let value = u32::try_from(value).map_err(|_| Errno::EBADF)?;
        Ok(Self(RangedU32::try_from(value).map_err(|_| Errno::EBADF)?))
    }
}
```

Apply the same fix to `TryFrom<usize>`.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +15,89 @@ use crate::{
...
-pub type FileDesc = i32;
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub struct FileDesc(RangedU32<0, { i32::MAX as _ }>);
+pub type RawFileDesc = i32;
+
+impl FileDesc {
+    pub fn raw_fd(self) -> RawFileDesc {
+        self.0.get() as _
+    }
+}
```
tatetian, line 39: [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary): `FileDesc`, `RawFileDesc`, and `raw_fd()` all lack doc comments. These are the central types of this PR, used across 80+ files. Add:

```rust
/// Represents a validated, non-negative file descriptor.
///
/// The value is guaranteed to be in the range `[0, i32::MAX]`.
/// Use [`RawFileDesc`] at syscall boundaries, 
/// then convert to `FileDesc` via `TryFrom` for kernel-internal use.
pub struct FileDesc(RangedU32<0, { i32::MAX as _ }>);

/// A raw file descriptor as received from or returned to user space.
///
/// This is the `int` type from Linux syscall signatures.
/// It may hold negative sentinel values like `AT_FDCWD` (-100).
/// Convert to [`FileDesc`] via `TryFrom` before use.
pub type RawFileDesc = i32;
```

Also implement `Display` for `FileDesc` (delegating to the inner integer) to avoid the scattered `.raw_fd()` calls in `debug!()` / `write!()` formatting throughout the PR.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +15,89 @@ use crate::{
...
+    fn from(value: FileDesc) -> Self {
+        value.0.get() as _
+    }
+}
+
+impl From<u16> for FileDesc {
+    fn from(value: u16) -> Self {
+        Self(RangedU32::new(value as _))
+    }
+}
```
tatetian: **`From<u16>` is unnecessary and the sole use site (`0.into()`) is a magic literal**: The only place this is used is `sys_dup` where `0.into()` creates `FileDesc(0)`. Having an infallible `From<u16>` in the public API implies that `u16` file descriptors are a common input type, which is misleading. A named constant would be clearer:

```rust
impl FileDesc {
    /// File descriptor 0 (stdin).
    pub const ZERO: Self = Self(RangedU32::new(0));
}
```

Then use `FileDesc::ZERO` instead of `0.into()`.

kernel/src/fs/file/file_table.rs
```diff
@@ -132,27 +216,28 @@ impl FileTable {
...
-            .get(fd as usize)
+            .get(fd.into())
             .ok_or(Error::with_message(Errno::EBADF, "fd not exits"))
     }
 
     pub fn get_entry_mut(&mut self, fd: FileDesc) -> Result<&mut FileTableEntry> {
         self.table
-            .get_mut(fd as usize)
+            .get_mut(fd.into())
             .ok_or(Error::with_message(Errno::EBADF, "fd not exits"))
```
tatetian: [**`accurate-names`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#accurate-names): Pre-existing typo `"fd not exits"` should be `"fd does not exist"`. It appears in `get_file` (line 221), `get_entry` (line 227), and `get_entry_mut` (line 233). The adjacent `duplicate_entry` (line 166) already uses the correct wording. Since the PR touched these lines (changing `fd as usize` to `fd.into()`), this is the right time to fix it.

kernel/src/syscall/getpeername.rs
```diff
@@ -2,21 +2,27 @@
...
+    sockfd: RawFileDesc,
     addr: Vaddr,
     addrlen_ptr: Vaddr,
     ctx: &Context,
 ) -> Result<SyscallReturn> {
     debug!("sockfd = {sockfd}, addr = 0x{addr:x}, addrlen_ptr = 0x{addrlen_ptr:x}");
 
     let mut file_table = ctx.thread_local.borrow_file_table_mut();
-    let file = get_file_fast!(&mut file_table, sockfd);
+    let file = get_file_fast!(
```
tatetian: **Inconsistent conversion pattern**: Every other syscall in this PR converts `RawFileDesc` to `FileDesc` using `sockfd.try_into()?`. This one uniquely uses `.cast_unsigned().try_into().map_err(|_| Errno::EBADF)?`. The `TryFrom<RawFileDesc>` impl already handles negative values correctly (they wrap to large `u32` values that fail the range check, producing `EBADF`). This bespoke path is redundant and inconsistent. Use `sockfd.try_into()?` to match all other syscalls.

kernel/src/syscall/pipe.rs
```diff
@@ -38,14 +38,18 @@ pub fn sys_pipe2(fds: Vaddr, flags: u32, ctx: &Context) -> Result<SyscallReturn>
...
 
     if let Err(err) = ctx.user_space().write_val(fds, &pipe_fds) {
-        file_table_locked.close_file(pipe_fds.reader_fd).unwrap();
-        file_table_locked.close_file(pipe_fds.writer_fd).unwrap();
+        file_table_locked
+            .close_file(pipe_fds.reader_fd.try_into()?)
+            .unwrap();
+        file_table_locked
+            .close_file(pipe_fds.writer_fd.try_into()?)
+            .unwrap();
```
tatetian: **Unnecessary `FileDesc` → `RawFileDesc` → `FileDesc` round-trip in error cleanup**: The code inserts files into the table (getting `FileDesc` values), immediately converts them to `RawFileDesc` via `.into()` for the `PipeFds` struct, then converts back via `.try_into()?` to call `close_file` in the error path. This round-trip through the raw representation and a fallible conversion is wasteful and obscure. Keep the original `FileDesc` values for the cleanup path and build the `PipeFds` struct separately for writing to user space.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +15,89 @@ use crate::{
...
+}
+
+impl TryFrom<RawFileDesc> for FileDesc {
+    type Error = Error;
+
+    fn try_from(value: RawFileDesc) -> Result<Self> {
+        Ok(Self(
+            RangedU32::try_from(value as u32).map_err(|_| Errno::EBADF)?,
+        ))
+    }
```
tatetian, line 85: **`value as u32` on negative `i32` is correct but obscure**: `-1i32 as u32` yields `4294967295`, which is correctly rejected by `RangedU32<0, i32::MAX>` since `4294967295 > 2147483647`. This works but relies on wrapping arithmetic — a reader must mentally verify the range math. Consider an explicit check for clarity:

```rust
fn try_from(value: RawFileDesc) -> Result<Self> {
    if value < 0 {
        return Err(Errno::EBADF.into());
    }
    Ok(Self(RangedU32::new(value as u32)))
}
```

This is a minor readability suggestion, not a bug.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +15,89 @@ use crate::{
     },
 };
 
-pub type FileDesc = i32;
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub struct FileDesc(RangedU32<0, { i32::MAX as _ }>);
+pub type RawFileDesc = i32;
+
+impl FileDesc {
+    pub fn raw_fd(self) -> RawFileDesc {
```
tatetian: Rename this method to `as_raw_fd`. Simply naming this method with a noun makes it look like a getter method, but it is a zero-cost conversion method. Prefix its name with `as_` follows Rust std's convention, e.g., `NonNull::as_ptr`.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -735,13 +735,14 @@ impl<'a> FsPath<'a> {
...
-            FsPathInner::Fd(dirfd)
-        } else if dirfd == AT_FDCWD {
+    pub fn from_fd(dirfd: RawFileDesc) -> Result<Self> {
+        let fs_path_inner = if dirfd == AT_FDCWD {
             FsPathInner::Cwd
         } else {
-            return_errno_with_message!(Errno::EBADF, "the dirfd is invalid");
+            FsPathInner::Fd(
+                FileDesc::try_from(dirfd)
+                    .map_err(|_| Error::with_message(Errno::EBADF, "the dirfd is invalid"))?,
```
tatetian: **Unnecessary `map_err`**: `FileDesc::try_from` already returns `Errno::EBADF` on failure (via `RangedU32::try_from(...).map_err(|_| Errno::EBADF)?` in the `TryFrom<RawFileDesc>` impl). The `.map_err(...)` here just replaces `Error::new(Errno::EBADF)` with `Error::with_message(Errno::EBADF, "the dirfd is invalid")` -- same errno, only adding a message string. Consider dropping the `map_err` and using `?` directly to keep the code concise. The same applies to the second occurrence at lines 771-772.

kernel/src/syscall/poll.rs
```diff
@@ -144,7 +144,7 @@ impl<'a> PollFiles<'a> {
             .map(|poll_fd| {
                 poll_fd
                     .fd()
-                    .and_then(|fd| file_table.get_file(fd).ok().cloned())
+                    .and_then(|raw_fd| file_table.get_file(raw_fd.try_into().ok()?).ok().cloned())
```
tatetian: **Readability**: This chain packs too many `.ok()?` / `.ok().cloned()` conversions into a single expression, making the control flow hard to follow. Consider breaking it into a multi-line closure, e.g.:

```rust
.and_then(|raw_fd| {
    // Negative FDs are simply ignored by `poll`.
    let fd = raw_fd.try_into().ok()?;
    // Non-negative FDs are used to look up the file table.
    file_table.get_file(fd).ok().cloned()
})
```

kernel/src/fs/file/file_table.rs
lrh2000: Actually, Linux's behavior is to truncate:
https://github.com/torvalds/linux/blob/d0c3bcd5b8976159d835a897254048e078f447e6/fs/fcntl.c#L449

I think this should be handled on a per-syscall basis instead of polluting the `file_table.rs` file, i.e., something like `arg as RawFileDesc` in `fcntl.rs`.

`TryFrom<u64>`/`TryFrom<usize>` does not make sense because its behavior is likely system call-specific. So I think they need to be removed.

kernel/src/fs/file/file_table.rs
lrh2000: Let's use the uppercase ("FD" v.s. "fd") version: "the FD does not exist". I think your suggested comment below already uses the uppercase version. So let's be consistent.

Perhaps fix `FileTable::duplicate_entry`'s error message as well.

kernel/src/fs/file/file_table.rs
lrh2000: - `return Err(Errno::EBADF.into())` ->
```rust
return_errno_with_message(Errno::EBADF, "negative FDs are not valid");
```

 - `value as u32` -> `value.cast_unsigned()`

kernel/src/fs/file/file_table.rs
chen1plus: This method will no longer be needed after implementing `ostd::Pod`.

chen1plus:
I'm going to push new commits to make things easier to check.
Tell me if you want to rebase them back to previous commits.

kernel/src/fs/file/file_table.rs
chen1plus: Upon further review, I suggest removing this method and consistently using `RawFileDesc::from` for all cases.

tatetian: Make sense.

kernel/src/fs/file/file_table.rs
```diff
@@ -132,27 +214,28 @@ impl FileTable {
 
     pub fn get_file(&self, fd: FileDesc) -> Result<&Arc<dyn FileLike>> {
         self.table
-            .get(fd as usize)
+            .get(fd.into())
             .map(|entry| entry.file())
-            .ok_or(Error::with_message(Errno::EBADF, "fd not exits"))
+            .ok_or(Error::with_message(Errno::EBADF, "the FD does not exits"))
```
tatetian: **Typo: "exits" should be "exist"**: This appears in `get_file` (line 219), `get_entry` (line 225), and `get_entry_mut` (line 231). Note that `duplicate_entry` on line 164 has the correct spelling: `"the FD does not exist"`.

kernel/src/fs/file/file_table.rs
```diff
@@ -60,7 +140,7 @@ impl FileTable {
 
         let min_free_fd = get_min_free_fd();
         self.table.put_at(min_free_fd, entry);
-        Ok(min_free_fd as FileDesc)
+        (min_free_fd as u32).try_into()
```
tatetian: **Inconsistent conversion pattern in `dup_ceil`**: Other methods that convert a `usize` index to `FileDesc` (e.g., `insert` at line 171, `close_files` at line 201, `fds_and_files` at line 238) all use the pattern `(idx as u32).try_into().unwrap()` with an explicit comment: "Resource limits guarantee the table never exceeds `i32::MAX` entries." Here, `dup_ceil` uses a fallible `.try_into()` without the comment, suggesting uncertainty about the invariant. If the invariant holds (and it should, since the table size is bounded), add the comment and use `.unwrap()` for consistency. If it doesn't hold, note that `as u32` silently truncates `usize` values above `u32::MAX` before the range check.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +18,84 @@ use crate::{
...
+        }
+        Ok(Self(RangedU32::new(value.cast_unsigned())))
+    }
+}
+
+impl TryFrom<u32> for FileDesc {
+    type Error = Error;
+
+    fn try_from(value: u32) -> Result<Self> {
+        Ok(Self(RangedU32::try_from(value).map_err(|_| Errno::EBADF)?))
```
tatetian: **Missing error message**: The `TryFrom<RawFileDesc>` impl (line 84) uses `return_errno_with_message!(Errno::EBADF, "negative FDs are not valid")`, but `TryFrom<u32>` returns a bare `Errno::EBADF` with no message. Since `TryFrom<u32>` rejects values above `i32::MAX`, a message like `"FD value exceeds maximum"` would aid debugging and match the pattern established by the sibling impl.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +18,84 @@ use crate::{
...
+/// A raw file descriptor as received from or returned to user space.
+///
+/// This is the `int` type from Linux syscall signatures.
+/// It may hold negative sentinel values like `AT_FDCWD` (-100).
+/// Convert to [`FileDesc`] via `TryFrom` before use.
+pub type RawFileDesc = i32;
+
+impl FileDesc {
+    /// File descriptor 0 (stdin).
+    pub const ZERO: Self = Self(RangedU32::new(0));
```
tatetian, line 38: **Inaccurate comment.** Do not mention stdin. Although fd 0 is usually used for the so-called standard input stream, this is not only a convention, not a kernel-enforced rule. And this convention is made by the file table.

```rust
    /// File descriptor 0.
    pub const ZERO: Self = Self(RangedU32::new(0));
```

chen1plus:
Everything is done. I'm not sure if I should squash the refactor commits into the previous one.

I didn't squash the rename commit because the scope is huge and rebasing would be painful. I'm open to doing it if you insist.

kernel/src/fs/file/file_table.rs
```diff
@@ -15,7 +18,82 @@ use crate::{
...
+
+    fn try_from(value: RawFileDesc) -> Result<Self> {
+        if value < 0 {
+            return_errno_with_message!(Errno::EBADF, "negative FDs are not valid");
+        }
+        Ok(Self(RangedU32::new(value.cast_unsigned())))
+    }
+}
+
+impl TryFrom<u32> for FileDesc {
```
lrh2000, line 88: The existing `(id as u32).try_into()` should actually be
 1. `(id as RawFileDesc).try_into()`, or
 2. `id.cast_signed().try_into()`.

So I don't see why we want to have this `TryFrom` either?

kernel/src/syscall/close.rs
```diff
@@ -83,7 +86,7 @@ pub fn sys_close_range(
         let actual_last = last.min(table_len - 1);
 
         for fd in first..=actual_last {
-            let fd = fd as FileDesc;
+            let fd = fd.try_into()?;
```
lrh2000, line 89:
```rust
            let fd = fd.cast_signed().try_into().unwrap()
```

This is already bounded by the file table length. So it must be a valid FD. Right?

kernel/src/fs/file/file_table.rs
tatetian: Given a `raw_id: u32`, how should we convert it to `FileDesc`?

1. `raw_id.try_into()`, or
2. `(id as RawFileDesc).try_into()`?

It looks that you vote for the second one. But I can't see why it is better than the first one, which is more concise.

lrh2000: I don't vote for either 1 or 2. Instead, I vote for `raw_id.cast_signed().try_into()`.

The reason is that: If we change the type alias definition of `RawFileDesc`, the compiler will report errors.

---

For `usize`, I will vote for: `id.truncate::<RawFileDesc>().try_into()` once https://github.com/rust-lang/libs-team/issues/204 lands. But for now, let's use `(id as FileDesc).try_into()`.

What is happening is very clear: We want to truncate the bits to get a `RawFileDesc`. Even if we change `RawFileDesc`'s definitions, as long as the code compiles, our intention won't be altered.

lrh2000:
@tatetian You can merge it as it is if you think it is ready. I will submit the remaining two comments as a separate PR, and you can decide whether you like them or not in that PR.

---

Meanwhile, I also want to know what are your thoughts on https://github.com/asterinas/asterinas/issues/2907#issuecomment-4175356583?

Going through the changes in this PR, I am not sure if it makes much sense to check the validity of the old FDs (while useless, it won't harm much, I suppose), but it is important to check the validity of the new FDs. We do miss the important `RLIMIT_NOFILE` checks in several places. (This is of course not this PR's fault so this PR does not need to address it). For example,
 - Example:
https://github.com/asterinas/asterinas/blob/6a7c4b33eabb5bed45b20684a606cf38f806f620/kernel/src/syscall/fcntl.rs#L45-L49
 - `dup_ceil` is also problematic as it does not actually validate whether the new FD exceeds `RLIMIT_NOFILE` (again, not this PR's fault so this PR does not need to address this problem).

tatetian:
> @tatetian You can merge it as it is if you think it is ready. I will submit the remaining two comments as a separate PR, and you can decide whether you like them or not in that PR.

Ok. Since this is a quite large PR and vulnerable to conflicts, I think it is already good enough to get merged. Let's discuss and fix minor issues separately.

kernel/src/fs/file/file_table.rs
tatetian: No. I don't get it. I can't see the harm of doing `id.try_into()?` directly for `id: u32` or `id: usize`. Using `RawFileDesc` as a bridge during the conversion is not getting us anything, I suppose.

> The reason is that: If we change the type alias definition of `RawFileDesc`, the compiler will report errors.

I can't foresee the reason to change the type of `RawFileDesc`. I am reading the [comment](https://github.com/asterinas/asterinas/issues/2907#issuecomment-4175356583) you gave below.

tatetian:
> if it makes much sense to check the validity of the old FDs (while useless, it won't harm much, I suppose), but it is important to check the validity of the new FDs

You mentioned the term "old FDs" and "new FDs" here and the other issue. But what do they mean exactly? I am confused. I did skim the Linux source code links that you gave, but still don't know what "old FDs" and "new FDs" refer to.

kernel/src/fs/file/file_table.rs
chen1plus: Changes to `RawFileDesc` are unlikely because it's being used as an argument type in several syscalls.

lrh2000: > Changes to `RawFileDesc` are unlikely because it's being used as an argument type in several syscalls.

Of course, that's unlikely, if not impossible. However, `u32` is quite magical. I would be much happier if it looked like `UnsignedRawFileDesc`.

But the problem is: Introducing an unsigned version of `RawFileDesc` for this purpose seems excessive to me. Additionally, the name is too long and ugly. (I don't have an immediate idea to figure out a reasonable one.) Then, I'd rather just avoid this entirely since `_ as RawFileDesc` (instead of `_ as u32`) looks quite right to me.

lrh2000: (**EDIT:** Actually, I think I have written too much unnecessarily. @tatetian Could you please just check the diff in https://github.com/asterinas/asterinas/pull/3090? I don't think the diff is controversial. To me, it's a clear improvement.)

> I can't foresee the reason to change the type of `RawFileDesc`. I am reading the [comment](https://github.com/asterinas/asterinas/issues/2907#issuecomment-4175356583) you gave below.

Assuming it never changes, you can safely write `(id as i32).try_into()`, which will be correct forever. You would prefer `(id as FileDesc).try_into()` because the latter version is much clearer, right?

I don't really think `u32` makes sense. When we use it, we actually want to use `UnsignedFileDesc`, but that fact is hidden, which is odd to me.

> the reason to change the type of `RawFileDesc`

Very old versions of Linux uses 16-bit UIDs/GIDs, they later changed to 32-bit UIDs/GIDs. People fifty years ago would think 4 GiB memory is infinite, but now they are proved to be wrong.

Of course, changing the basic types is difficult these days because doing so would certainly break many programs. However, theoretically, it's possible, and I still think it's a valid argument.

> I can't see the harm of doing `id.try_into()?` directly for `id: u32` or `id: usize`. Using `RawFileDesc` as a bridge during the conversion is not getting us anything, I suppose.

For `usize`, it is ambiguous because it is unclear whether it will truncate. The exact semantics need to be expressed explicitly by a `truncate` method (https://github.com/rust-lang/libs-team/issues/204).
 - The `TryFrom` traits, as well as the `as` cast, typically have many different overloads with different semantics. Based on our (previous) discussions, you seem to be fans of them. However, I would argue that we should use them carefully because their semantics aren't always clear.
 - We should try to use methods with clear semantics whenever possible. For example, we can prefer the `cast_signed`/`cast_unsigned` method (already stable) or https://github.com/rust-lang/libs-team/issues/204 (soon nightly available).

For `u32`, I cannot see why we need to write `(_ as u32).into()` in several places in this PR. Perhaps it's because it's shorter than `(_ as RawFileDesc).into()`? But again, `u32` does not make much sense in these places.

lrh2000:
> You mentioned the term "old FDs" and "new FDs" here and the other issue. But what do they mean exactly? I am confused. I did skim the Linux source code links that you gave, but still don't know what "old FDs" and "new FDs" refer to.

Oh sorry during a rewrite, I lost some words in my previous version and I don't realize that..

 - Validity checks for old FDs **(for file lookup)** can be "omitted".
   - See https://github.com/asterinas/asterinas/issues/2907#issuecomment-4089417067. The only thing that needs to be checked is that the FD does not exceed the length of the file table, so it must already be valid.
 - Validity checks for new FDs **(for FD creation)** need to be checked against `RLIMIT_NOFILE`:
   - Some system calls accept an FD to install a new entry in the file table. In this case, the FD is referred to as a "new FD." This includes:
     - The `ceil_fd` parameter of `dup_ceil`, which specifies the lower bound of the newly created FD.
     - The return value of `dup_ceil`, which is a new FD that just installed to the file table.
     https://github.com/asterinas/asterinas/blob/318f374a770677ce971d6c7ee16f3f83083566eb/kernel/src/fs/file/file_table.rs#L116-L121
     - The `new_fd` argument of the `dup3` system call:
     https://github.com/asterinas/asterinas/blob/318f374a770677ce971d6c7ee16f3f83083566eb/kernel/src/syscall/dup.rs#L50-L55

In other words, old FDs refer to reading the file table, where the file table entry exists before the system call starts. New FDs refer to writing the file table, where the system call creates file table entries at a specific offset.

PR #3000, opened 2026-03-03
Keyboard enhancements

wyt8:
# Keyboard enhancements

## Description

This PR refactors the input processing logic to more closely mimic the Linux kernel's keyboard driver behavior. By adopting a more "Linux-aligned" approach, we are laying the groundwork for:

- Virtual Console (TTY) Switching: Support for Alt+Fn combinations.

- Custom Keymaps: Allowing users to redefine keys via a mechanism similar to loadkeys.

## Reference

To facilitate the review process, the default Linux keymap mappings were extracted using a script as follows:

| id  | key                  | plain                       | alt                          | shift                       | ctrl                        | shift_ctrl                  | ctrl_alt                    |
|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
| 0   | KEY_RESERVED         |                             |                              |                             |                             |                             |                             |
| 1   | KEY_ESC              | k_self(0x1b)(ESC)           | k_meta(0x1b)(ESC)            | k_self(0x1b)(ESC)           |                             |                             |                             |
| 2   | KEY_1                | k_self(0x31)('1')           | k_meta(0x31)('1')            | k_self(0x21)('!')           |                             |                             |                             |
| 3   | KEY_2                | k_self(0x32)('2')           | k_meta(0x32)('2')            | k_self(0x40)('@')           | k_self(0x0)(0x0)            | k_self(0x0)(0x0)            |                             |
| 4   | KEY_3                | k_self(0x33)('3')           | k_meta(0x33)('3')            | k_self(0x23)('#')           | k_self(0x1b)(ESC)           |                             |                             |
| 5   | KEY_4                | k_self(0x34)('4')           | k_meta(0x34)('4')            | k_self(0x24)('$')           | k_self(0x1c)(0x1c)          |                             |                             |
| 6   | KEY_5                | k_self(0x35)('5')           | k_meta(0x35)('5')            | k_self(0x25)('%')           | k_self(0x1d)(0x1d)          |                             |                             |
| 7   | KEY_6                | k_self(0x36)('6')           | k_meta(0x36)('6')            | k_self(0x5e)('^')           | k_self(0x1e)(0x1e)          |                             |                             |
| 8   | KEY_7                | k_self(0x37)('7')           | k_meta(0x37)('7')            | k_self(0x26)('&')           | k_self(0x1f)(0x1f)          |                             |                             |
[...]

PR #3029, opened 2026-03-11
Refactor the tty module

wyt8:
# Key Changes

**`n_tty.rs` Decoupling**: Previously, our `n_tty.rs` module incorrectly bundled specific device types (such as hvc0, ttyS0, and tty1). In Linux, n_tty is a generic Line Discipline (ldisc) rather than a device container. This PR extracts these device-specific implementations into their own modules.

**Unified the ioctl processing logic**: While various TTY types (hvc, tty, ttyS, pty) share common ioctls (e.g., TCGETS, TCSETS), they also require hardware-specific or protocol-specific ioctl handlers. I have refactored the dispatcher to handle common TTY ioctls at the layer above while allowing device drivers to implement their own specialized ioctl callbacks.

## Note

This pr should be reviewed after #3000 is merged.

### DIFF
--- a/kernel/src/device/pty/driver.rs
+++ b/kernel/src/device/pty/driver.rs
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use aster_console::AnyConsoleDevice;
 use ostd::sync::SpinLock;
 
 use super::file::PtySlaveFile;
@@ -16,7 +15,10 @@ use crate::{
     fs::file::FileIo,
     prelude::*,
     process::signal::Pollee,
-    util::ring_buffer::RingBuffer,
+    util::{
+        ioctl::{RawIoctl, dispatch_ioctl},
+        ring_buffer::RingBuffer,
+    },
 };
 
 const BUFFER_CAPACITY: usize = 8192;
@@ -171,10 +173,6 @@ impl TtyDriver for PtyDriver {
         self.pollee.notify(IoEvents::OUT);
     }
 
-    fn console(&self) -> Option<&dyn AnyConsoleDevice> {
-        None
-    }
-
     fn on_termios_change(&self, old_termios: &CTermios, new_termios: &CTermios) {
         // Reference: <https://elixir.bootlin.com/linux/v6.17/source/drivers/tty/pty.c#L246>.
         let extproc = old_termios.local_flags().contains(CLocalFlags::EXTPROC)
@@ -209,5 +207,22 @@ impl TtyDriver for PtyDriver {
             self.pollee
                 .notify(IoEvents::PRI | IoEvents::IN | IoEvents::RDNORM);
         }
+    }
+
+    fn ioctl(&self, tty: &Tty<Self>, raw_ioctl: RawIoctl) -> Result<bool>
+    where
+        Self: Sized,
+    {
+        use super::ioctl_defs::*;
+
+        dispatch_ioctl!(match raw_ioctl {
+            cmd @ GetPtyNumber => {
+                let idx = tty.index();
+                cmd.write(&idx)?;
+            }
+            _ => return Ok(false),
+        });
+
+        Ok(true)
     }
 }
--- a/kernel/src/device/pty/ioctl_defs.rs
+++ b/kernel/src/device/pty/ioctl_defs.rs
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use crate::util::ioctl::{InData, NoData, OutData, ioc};
+
+// Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/asm-generic/ioctls.h>
+
+pub(super) type SetPtyLock   = ioc!(TIOCSPTLCK,  b'T', 0x31, InData<i32>);
+pub(super) type GetPtyLock   = ioc!(TIOCGPTLCK,  b'T', 0x39, OutData<i32>);
+
+pub(super) type OpenPtySlave = ioc!(TIOCGPTPEER, b'T', 0x41, NoData);
+
+pub(super) type SetPktMode   = ioc!(TIOCPKT,     0x5420,     InData<i32>);
+pub(super) type GetPktMode   = ioc!(TIOCGPKT,    b'T', 0x38, OutData<i32>);
+
+pub(super) type GetPtyNumber = ioc!(TIOCGPTN,    b'T', 0x30, OutData<u32>);
--- a/kernel/src/device/tty/driver.rs
+++ b/kernel/src/device/tty/driver.rs
@@ -1,11 +1,10 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use aster_console::AnyConsoleDevice;
-
 use crate::{
     device::tty::{Tty, termio::CTermios},
     fs::file::FileIo,
     prelude::*,
+    util::ioctl::RawIoctl,
 };
 
 /// A TTY driver.
@@ -55,13 +54,27 @@ pub trait TtyDriver: Send + Sync + 'static {
     /// [`Tty::can_push`]: super::Tty::can_push
     fn notify_input(&self);
 
-    /// Returns the console device associated with the TTY.
-    ///
-    /// If the TTY is not associated with any console device, this method will return `None`.
-    fn console(&self) -> Option<&dyn AnyConsoleDevice>;
-
     /// Notifies that the TTY termios is changed.
     ///
     /// This method will be called with a spin lock held, so it cannot break atomic mode.
     fn on_termios_change(&self, old_termios: &CTermios, new_termios: &CTermios);
+
+    /// Handles driver-specific ioctl.
+    ///
+    /// This method allows a TTY driver to handle driver-specific
+    /// ioctl commands that are not processed by the generic TTY layer.
+    ///
+    /// The return value depends on whether the ioctl is recognized:
+    /// - If the driver recognizes and handles the ioctl, it should return
+    ///   `Ok(true)`.
+    /// - If an error occurs while processing the ioctl, it should return
+    ///   `Err(_)`.
+    /// - If the driver does not recognize the ioctl command, it should return
+    ///   `Ok(false)` to indicate that the ioctl command is not supported.
+    fn ioctl(&self, _tty: &Tty<Self>, _raw_ioctl: RawIoctl) -> Result<bool>
+    where
+        Self: Sized,
+    {
+        Ok(false)
+    }
 }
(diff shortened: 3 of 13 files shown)


kernel/src/device/tty/file.rs
```diff
@@ -0,0 +1,64 @@
...
+    },
+    prelude::*,
+    process::signal::{PollHandle, Pollable},
+    util::ioctl::RawIoctl,
+};
+
+pub struct TtyFile<D>(Arc<Tty<D>>);
+
+impl<D: TtyDriver> TtyFile<D> {
+    pub fn new(tty: Arc<Tty<D>>) -> Self {
```
lrh2000: Line 20/Line 23: `pub(super)` is enough?

kernel/src/device/tty/vt/mod.rs
```diff
@@ -28,5 +31,6 @@ impl VtIndex {
 
 pub(super) fn init_in_first_process() -> Result<()> {
     keyboard::init();
+    driver::init()?;
```
lrh2000: This should be `init_in_first_process`... as well as `keyboard::init`.

I don't catch this in your previous PR, but it seems to be the only instance in our codebase:
```
> grep -R "init_in_first_process" -A2 kernel | grep 'init('
kernel/src/device/tty/vt/mod.rs-    keyboard::init();
kernel/src/device/tty/vt/mod.rs-    driver::init()?;
```

kernel/src/device/pty/ioctl_defs.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub type SetPtyLock   = ioc!(TIOCSPTLCK,  b'T', 0x31, InData<i32>);
+pub type GetPtyLock   = ioc!(TIOCGPTLCK,  b'T', 0x39, OutData<i32>);
+
+pub type OpenPtySlave = ioc!(TIOCGPTPEER, b'T', 0x41, NoData);
+
+pub type SetPktMode   = ioc!(TIOCPKT,     0x5420,     InData<i32>);
+pub type GetPktMode   = ioc!(TIOCGPKT,    b'T', 0x38, OutData<i32>);
+
+pub type GetPtyNumber = ioc!(TIOCGPTN, b'T', 0x30, OutData<u32>);
```
lrh2000: This file should use `pub(super)`?

kernel/src/device/tty/vt/ioctl_defs.rs
```diff
@@ -0,0 +1,14 @@
...
+    util::ioctl::{InData, OutData, PassByVal, ioc},
+};
+
+pub type SetGraphicsMode = ioc!(KDSETMODE,  0x4B3A,     InData<i32, PassByVal>);
+pub type GetGraphicsMode = ioc!(KDGETMODE,  0x4B3B,     OutData<i32>);
+
+pub type GetKeyboardMode = ioc!(KDGKBMODE,  0x4B44,     OutData<i32>);
+pub type SetKeyboardMode = ioc!(KDSKBMODE,  0x4B45,     InData<i32, PassByVal>);
+
+pub type SetOrGetFont    = ioc!(KDFONTOP,   0x4B72,     InData<CFontOp>);
```
lrh2000: This file should use `pub(super)`?

Add a reference link like other `ioctl_defs.rs` files.

kernel/src/device/tty/driver.rs
```diff
@@ -58,10 +59,30 @@ pub trait TtyDriver: Send + Sync + 'static {
     /// Returns the console device associated with the TTY.
     ///
     /// If the TTY is not associated with any console device, this method will return `None`.
+    #[expect(dead_code)]
     fn console(&self) -> Option<&dyn AnyConsoleDevice>;
```
lrh2000: Remove this method if it is not useful anymore.

kernel/src/device/tty/driver.rs
```diff
@@ -58,10 +59,30 @@ pub trait TtyDriver: Send + Sync + 'static {
...
 
     /// Notifies that the TTY termios is changed.
     ///
     /// This method will be called with a spin lock held, so it cannot break atomic mode.
     fn on_termios_change(&self, old_termios: &CTermios, new_termios: &CTermios);
+
+    /// Driver-specific ioctl handler.
```
lrh2000:
```suggestion
    /// Handles driver-specific ioctl.
```

kernel/src/device/tty/driver.rs
```diff
@@ -58,10 +59,30 @@ pub trait TtyDriver: Send + Sync + 'static {
...
+    /// This method allows a TTY driver to handle driver-specific
+    /// ioctl commands that are not processed by the generic TTY layer.
+    ///
+    /// Semantics:
+    /// - If the driver recognizes and handles the ioctl, it should return
+    ///   `Ok(retval)`, where `retval` is the value returned to userspace.
+    /// - If an error occurs while processing the ioctl, it should return
+    ///   `Err(...)`.
+    /// - If the driver does not recognize the ioctl command, it should return
+    ///   `Err(Errno::ENOIOCTLCMD)` to indicate that the ioctl command is not supported.
```
lrh2000: `ENOIOCTLCMD` is an internal error code in the Linux kernel. In Rust, maybe we can have something different? For example, we can return `Ok(Some(_))` if the ioctl command is processed and `Ok(None)` if the ioctl command is not recognized.

kernel/src/device/tty/serial.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    fn on_termios_change(&self, _old_termios: &CTermios, _new_termios: &CTermios) {}
+}
+
+static SERIAL0: Once<Arc<Tty<SerialDriver>>> = Once::new();
+
+/// Returns the `ttyS0` device.
+///
+/// Returns `None` if the device is not found nor initialized.
+pub fn serial0_device() -> Option<&'static Arc<Tty<SerialDriver>>> {
```
lrh2000: It seems that `serial0_device`/`SerialDriver`/`hcv0_device`/`HcvDriver` can also be `pub(super)` (it's not a problem introduced in this PR though)?

kernel/src/device/tty/driver.rs
```diff
@@ -58,10 +59,30 @@ pub trait TtyDriver: Send + Sync + 'static {
...
+    /// Driver-specific ioctl handler.
+    ///
+    /// This method allows a TTY driver to handle driver-specific
+    /// ioctl commands that are not processed by the generic TTY layer.
+    ///
+    /// Semantics:
+    /// - If the driver recognizes and handles the ioctl, it should return
+    ///   `Ok(retval)`, where `retval` is the value returned to userspace.
+    /// - If an error occurs while processing the ioctl, it should return
+    ///   `Err(...)`.
```
lrh2000: Use `Err(_)` instead.

kernel/src/device/pty/ioctl_defs.rs
lrh2000: Insert some whitespace to align the arguments:
```rust
pub type GetPktMode   = ioc!(TIOCGPKT,    b'T', 0x38, OutData<i32>);

pub type GetPtyNumber = ioc!(TIOCGPTN,    b'T', 0x30, OutData<u32>);
```

kernel/src/device/tty/driver.rs
wyt8: In fact, I just use this pattern you mentioned to handle the `ioctl` returns. But I found that the return value handle inner this method can be fussy because our codebase usually uses `Result(...)` not `Result(Option(...))`.

I think I should not cause this sititution so I change the way. If you think we should still use this pattern, I will change it.

kernel/src/device/tty/driver.rs
```diff
@@ -55,13 +54,27 @@ pub trait TtyDriver: Send + Sync + 'static {
...
     ///
     /// This method will be called with a spin lock held, so it cannot break atomic mode.
     fn on_termios_change(&self, old_termios: &CTermios, new_termios: &CTermios);
+
+    /// Handles driver-specific ioctl.
+    ///
+    /// This method allows a TTY driver to handle driver-specific
+    /// ioctl commands that are not processed by the generic TTY layer.
+    ///
+    /// Semantics:
```
lrh2000:
```suggestion
    /// The return value depends on whether the ioctl is recognized:
```

kernel/src/device/tty/vt/driver.rs
```diff
@@ -51,11 +124,45 @@ impl TtyDriver for VtDriver {
...
+    fn ioctl(&self, _tty: &Tty<Self>, raw_ioctl: RawIoctl) -> Result<i32>
+    where
+        Self: Sized,
+    {
+        use super::ioctl_defs::*;
+
+        dispatch_ioctl!(match raw_ioctl {
+            cmd @ SetOrGetFont => {
+                let font_op = cmd.read()?;
+
```
lrh2000: Blank lines in other ioctl commands like `SetGraphicsMode` are removed, so let's also remove this?

kernel/src/device/tty/driver.rs
lrh2000: > But I found that the return value handle inner this method can be fussy because our codebase usually uses `Result(...)` not `Result(Option(...))`.

Could you please be more specific? Honestly, I'm not sure what you mean.

If it needs to return something depending on the ioctl command, the code will be:
```rust
let retval = dispatch_ioctl!(match raw_ioctl {
    // Some ioctl commands
    _ => return Ok(None);
});

Ok(Some(retval))
```

I don't see this pattern as problematic.

In fact, we now seem only to return zeros to userspace. So even `retval` isn't necessary:
```rust
dispatch_ioctl!(/* .. */);

Ok(Some(0))
```

Actually, if the return value is always zero, maybe we should return a boolean or a boolean wrapper (like [`NeedIfacePoll`](https://github.com/asterinas/asterinas/blob/3f3ad398cd8ec9016aa4f32cd18e44a56cc4a2c5/kernel/libs/aster-bigtcp/src/socket/bound/common.rs#L107-L110) (but `define_boolean_value` isn't public now) to indicate whether the ioctl command has been processed.

```rust
// Process job-control ioctls.
if terminal.job_ioctl(raw_ioctl, false)? {
    return Ok(0);
}

// Process driver-specific ioctls.
if self.driver.ioctl(self, raw_ioctl)? {
    return Ok(0);
}

return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown");
```

kernel/src/device/pty/driver.rs
```diff
@@ -210,4 +208,20 @@ impl TtyDriver for PtyDriver {
...
+    where
+        Self: Sized,
+    {
+        use super::ioctl_defs::*;
+
+        dispatch_ioctl!(match raw_ioctl {
+            cmd @ GetPtyNumber => {
+                let idx = tty.index();
+                cmd.write(&idx)?;
+                Ok(0)
```
lrh2000: Similar to other `ioctl` handlers. Move `Ok(0)` to the end of the method.

Issue #3030, opened 2026-03-12
ACPI tables cannot be parsed when memory is low under multiboot protocol


StevenJiang1110:
When using Asterinas as the guest kernel for Kata Containers(#2911), I found that ACPI table parsing always fails. After investigation, I traced the issue to Kata's default memory size of 2GB: when the memory is limited to 2GB, accessing the address corresponding to the RSDP register consistently triggers a kernel panic.

In the current codebase, the ACPI table is actually parsed multiple times. Specifically, get_acpi_tables() succeeds when called before the kernel page table is activated, but triggers a kernel panic when called after activation.

# To reproduce

First, add the following log output:

```diff
diff --git a/ostd/src/arch/x86/kernel/acpi/mod.rs b/ostd/src/arch/x86/kernel/acpi/mod.rs
index 99aa44795..90ca1ef1b 100644
--- a/ostd/src/arch/x86/kernel/acpi/mod.rs
+++ b/ostd/src/arch/x86/kernel/acpi/mod.rs
@@ -17,6 +17,7 @@ use spin::Once;
 
 use crate::{
     boot::{self, BootloaderAcpiArg},
+    early_println,
     mm::paddr_to_vaddr,
 };
 
@@ -59,9 +60,14 @@ pub(crate) fn get_acpi_tables() -> Option<AcpiTables<AcpiMemoryHandler>> {
             // We search by ourselves if the bootloader decides not to provide a rsdp location.
             let rsdp = unsafe { Rsdp::search_for_on_bios(AcpiMemoryHandler {}) };
             match rsdp {
-                Ok(map) => unsafe {
-                    AcpiTables::from_rsdp(AcpiMemoryHandler {}, map.physical_start()).unwrap()
-                },
+                Ok(map) => {
+                    early_println!("\nRSDP addr = 0x{:?}", map.virtual_start());
+                    let table = unsafe {
+                        AcpiTables::from_rsdp(AcpiMemoryHandler {}, map.physical_start()).unwrap()
+                    };
+                    early_println!("READ ACPI tables SUCCESS");
+                    table
+                }
                 Err(_) => {
                     warn!("ACPI info not found!");
                     return None;
diff --git a/ostd/src/lib.rs b/ostd/src/lib.rs
index 60b41bad6..2027e39d6 100644
--- a/ostd/src/lib.rs
+++ b/ostd/src/lib.rs
@@ -102,9 +102,13 @@ unsafe fn init() {
 
[...]

PR #3031, opened 2026-03-12
Init ACPI table only once

StevenJiang1110:
While running Asterinas as the guest kernel for Kata Containers, I found that ACPI table initialization always results in a panic. This panic issue has been documented in #3030.

I also noticed that `AcpiTables` is initialized multiple times across the codebase, even though all call sites ultimately use the same instance. Therefore, I think it makes sense to wrap it in a `Once` so that it is initialized only once.

One complication is that the `AcpiTables` type from the `acpi` crate does not implement `Sync`(https://docs.rs/acpi/5.2.0/acpi/struct.AcpiTables.html#impl-Sync-for-AcpiTables%3CH%3E). To address this, I introduced a wrapper type `SyncAcpiTables` to make it safe to share across threads.

### DIFF
--- a/ostd/src/arch/x86/boot/smp.rs
+++ b/ostd/src/arch/x86/boot/smp.rs
@@ -231,7 +231,7 @@ unsafe fn wake_up_aps_via_mailbox(num_cpus: u32) {
     let acpi_tables = get_acpi_tables().unwrap();
     for ap_num in 1..num_cpus {
         wakeup_aps(
-            &acpi_tables,
+            acpi_tables,
             AcpiMemoryHandler {},
             ap_num,
             (AP_BOOT_START_PA + offset) as u64,
--- a/ostd/src/arch/x86/kernel/acpi/mod.rs
+++ b/ostd/src/arch/x86/kernel/acpi/mod.rs
@@ -44,33 +44,47 @@ impl AcpiHandler for AcpiMemoryHandler {
     fn unmap_physical_region<T>(_region: &acpi::PhysicalMapping<Self, T>) {}
 }
 
-pub(crate) fn get_acpi_tables() -> Option<AcpiTables<AcpiMemoryHandler>> {
-    let acpi_tables = match boot::EARLY_INFO.get().unwrap().acpi_arg {
-        BootloaderAcpiArg::Rsdp(addr) => unsafe {
-            AcpiTables::from_rsdp(AcpiMemoryHandler {}, addr).unwrap()
-        },
-        BootloaderAcpiArg::Rsdt(addr) => unsafe {
-            AcpiTables::from_rsdt(AcpiMemoryHandler {}, 0, addr).unwrap()
-        },
-        BootloaderAcpiArg::Xsdt(addr) => unsafe {
-            AcpiTables::from_rsdt(AcpiMemoryHandler {}, 1, addr).unwrap()
-        },
-        BootloaderAcpiArg::NotProvided => {
-            // We search by ourselves if the bootloader decides not to provide a rsdp location.
-            let rsdp = unsafe { Rsdp::search_for_on_bios(AcpiMemoryHandler {}) };
-            match rsdp {
-                Ok(map) => unsafe {
-                    AcpiTables::from_rsdp(AcpiMemoryHandler {}, map.physical_start()).unwrap()
-                },
-                Err(_) => {
-                    warn!("ACPI info not found!");
-                    return None;
+struct SyncAcpiTables(Option<AcpiTables<AcpiMemoryHandler>>);
+
+// SAFETY: This relies on the current implementation of `AcpiTables`,
+// which provides thread-safe access to read-only ACPI table data,
+// so `Sync` is sound for the wrapper.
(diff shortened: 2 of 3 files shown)


lrh2000:
> One complication is that the `AcpiTables` type from the `acpi` crate does not implement `Sync`(https://docs.rs/acpi/5.2.0/acpi/struct.AcpiTables.html#impl-Sync-for-AcpiTables%3CH%3E). To address this, I introduced a wrapper type `SyncAcpiTables` to make it safe to share across threads.

Though unlikely, it is important to ensure that it does not contain a `RefCell` or something similar that cannot be shared across threads. Even unlikely, there may be some (unsaid) type invariants that prevent `Sync` to be implemented.

```rust
// SAFETY: `AcpiTables` is immutable after initialization,
// so sharing references to `SyncAcpiTables` across threads is safe.
```

I don't know what you mean by saying "immutable". By your rules, do you think `&'static RefCell<i32>` is immutable or mutable?

Maybe you want to say something like "interior mutability"? I don't know.

Admittedly, it may be impossible to write safety comments for types that you don't implemented yourself.

StevenJiang1110:
> I don't know what you mean by saying "immutable". By your rules, do you think &'static RefCell<i32> is immutable or mutable?

What I meant to say is that AcpiTables is read-only: once the struct is constructed, its contents can only be accessed through read-only methods — there is no way to write to the underlying table data. In fact, our current implementation already relies on this assumption. If the underlying memory region were writable having multiple `AcpiTables` instances pointing to the same region would inevitably introduce data races.

> Though unlikely, it is important to ensure that it does not contain a RefCell or something similar that cannot be shared across threads. Even unlikely, there may be some (unsaid) type invariants that prevent Sync to be implemented.

The reason `AcpiTables` does not implement `Sync` is that it contains raw pointers internally, which inherently prevents an automatic `Sync` implementation. However, I guess this is a limitation of the current version of the crate rather than a fundamental soundness concern — in version 6.1.1, `AcpiTables` does implement ``Sync (see [docs](https://docs.rs/acpi/6.1.1/acpi/struct.AcpiTables.html#impl-Sync-for-AcpiTables%3CH%3E)), which suggests that the absence of `Sync` in older versions is simply an oversight. That said, the overall code quality of this crate is not great — most unsafe blocks lack documentation, making it difficult to verify their safety requirements. I did consider bumping the acpi dependency to 6.1.1, but the API differences are significant enough that I decided against it for now.

I have updated the safety comments accordingly. I thin the soundness argument rests on two guarantees: (1) the memory regions accessed by `AcpiTables` are read-only, as provided by the third-party crate; and (2) the virtual address mappings for those regions remain unchanged after initialization. 

On a related note, this discussion suggests that the issue described in #3030 may actually be a soundness hole. Our current implementation implicitly assumes that the virtual address mappings for the ACPI table regions remain stable, but in practice these mappings can be invalidated under certain conditions. This means our implementation does not satisfy the safety requirement that the virtual address must always be valid (see [docs](https://docs.rs/acpi/5.2.0/acpi/handler/struct.PhysicalMapping.html#method.new)).

https://github.com/asterinas/asterinas/blob/3f3ad398cd8ec9016aa4f32cd18e44a56cc4a2c5/ostd/src/arch/x86/kernel/acpi/mod.rs#L26-L42

If we follow the intended usage of `PhysicalMapping` from the `acpi` crate, we should be constructing the mapping within the handler function and relying on the `Drop` implementation of `PhysicalMapping` to automatically unmap the region when it goes out of scope. This approach would also naturally resolve the issue described in #3030.
Under this API contract, the changes in this PR would be inappropriate — since the mapping should be established on every read access, it would not make sense to initialize `AcpiTables` only once at boot time. This ultimately comes down to a design decision about our ACPI table mapping strategy: should we maintain a permanent, fixed mapping for the ACPI table regions, or should we establish a temporary mapping on each access and tear it down immediately afterwards? I'm curious to hear your thoughts on this.

lrh2000:
> I'm curious to hear your thoughts on this.

See https://github.com/asterinas/asterinas/issues/3030#issuecomment-4059043540 for this.

> I thin the soundness argument rests on two guarantees:

Thank you for your detailed explanation. However, could you please tell me which part of your argument would be violated if I implemented the `AcpiTables` this way?

```rust
struct AcpiTables {
    counter: Cell<u32>,
}

impl AcpiTables {
    fn access_table(&self, typ: AcpiTableType) -> SomeAcpiTableReference {
        counter.set(self.counter.get() + 1);
        todo!()
    }

    fn num_accessed(&self) -> u32 {
        self.counter.get()
    }
}
```

Let me propose something differently, I think this is better:
```rust
// SAFETY: This relies on the current implementation of `AcpiTables`, which provides thread-safe
// access to read-only ACPI table data, so `Sync` is sound for the wrapper.
// FIXME: It depends on implementation details of `AcpiTables`, which should be avoided.
unsafe impl Sync for SyncAcpiTables {}
```

StevenJiang1110:
> Let me propose something differently, I think this is better:

Thanks for your advice! I agree the suggested form is better. I have changed the comment as suggested.

PR #3033, opened 2026-03-13
Support devices/FIFO in initramfs

StevenJiang1110:
Currently, we don't support device types in initramfs. However, kata initfs contais char devices. So asterinas will panic in this case. This PR adds support for devices/FIFO in initramfs.

### DIFF
--- a/kernel/src/fs/rootfs.rs
+++ b/kernel/src/fs/rootfs.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use core2::io::{Cursor, Read};
-use cpio_decoder::{CpioDecoder, FileMetadata, FileType};
+use cpio_decoder::{CpioDecoder, CpioEntry, FileMetadata, FileType};
 use device_id::{DeviceId, MajorId, MinorId};
 use lending_iterator::LendingIterator;
 use libflate::gzip::Decoder as GZipDecoder;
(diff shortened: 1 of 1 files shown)


kernel/src/fs/rootfs.rs
```diff
@@ -97,12 +98,32 @@ pub fn init_in_first_kthread(path_resolver: &PathResolver) -> Result<()> {
...
             }
         }
     }
 
     println!("[kernel] rootfs is ready");
     Ok(())
 }
+
+fn device_id_from_metadata(metadata: &FileMetadata) -> u64 {
+    let major = MajorId::new(metadata.rdev_maj() as _);
```
lrh2000: `MajorId::new` panics for out-of-bound marjor IDs. So to have consistent behavior for those IDs, we can just use:
```rust
    let major = MajorId::new(metadata.rdev_maj().try_into().unwrap());
```
?

kernel/src/fs/rootfs.rs
```diff
@@ -97,12 +98,32 @@ pub fn init_in_first_kthread(path_resolver: &PathResolver) -> Result<()> {
...
+                let device_id = device_id_from_metadata(metadata);
+                parent.mknod(name, mode, MknodType::BlockDevice(device_id))?;
+            }
+            FileType::FiFo => {
+                parent.mknod(name, mode, MknodType::NamedPipe)?;
+            }
+            FileType::Socket => {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "socket files are not supported in initramfs"
```
tatetian: Failing the entire kernel just because a socket file is unsupported sounds too strict. Maybe we should be more tolerable and just log a warning.

kernel/src/fs/rootfs.rs
tatetian: It is undesirable to allow user-given inputs to trigger panics. Panicking is the last resort when unrecoverable programming errors (most likely a bug of the kernel). I think we should make this function fallible and the caller only logs a warning if the function fails. Now that the function is fallible, rename the function to make it sound a bit more dangerous.

lrh2000: I don't think initramfs/kcmdline are "user-given" inputs. It's "administrator-given" input. In production environments, they're even signed by administrators. So we can assume that they're correct and panicking (refusing to boot) if they're incorrect.

So I don't think panicking here is problem. Even if we return an error, we will panic at:
https://github.com/asterinas/asterinas/blob/4d5af25230b6beabd9f548f7b9da1e870ebf0597/kernel/src/fs/mod.rs#L32-L33

kernel/src/fs/rootfs.rs
lrh2000: initramfs is something that should be signed by system administrators. Therefore, they shouldn't contain anything unsupported. It may be fine to log a warning, but panicking should also be a reasonable choice....

Regardless of our choice, we currently do fail the entire method in the following loop due to various ways:
https://github.com/asterinas/asterinas/blob/160e8dd4177b03a6619449aed6498f407155faa4/kernel/src/fs/rootfs.rs#L56-L59

I think this comment is out of the scope of this PR. You may want to see:
```rust
    loop {
        let Some(entry_result) = decoder.next() else {
            break;
        };

        let res = some_method_deal_with(entry_result);
        if res.is_err() {
            log::warn!("xxx");
        }
    }
```
So we won't abort if some entry is corrupted.

However, again, the initramfs should be signed by the system administrators, along with the kernel image itself. Is it worth continuing if errors occur when parsing the initramfs image?

kernel/src/fs/rootfs.rs
lrh2000: > kcmdline

Mention it because I also notice this:
https://github.com/asterinas/asterinas/blob/d27a16ec511da0a144a71a32a2eca685b1d8e3fb/kernel/src/process/process/init_proc.rs#L75
Or:
https://github.com/asterinas/asterinas/blob/d27a16ec511da0a144a71a32a2eca685b1d8e3fb/kernel/comps/cmdline/src/lib.rs#L171-L173

Do you think we need to deal with the situation when the cmdline is not valid? To make the code simpler, I think it is okay to just refuse to boot... We must trust the cmdline because it may contain security-sensitive (e.g., `root=`/`init=`) or debugging options. So handling the error does not sound quite necessary.

tatetian: Administrators are also human and human makes mistakes.

```rust
 if result.initproc.path.is_some() { 
     panic!("[KCmdline] Init process specified twice"); 
 } 
```

I am glad that you brought up this example. I was told by @taosue the other day that Linux's behavior is to allow an empty `init` CLI argument. If it is empty (not given), Linux will try to read several known paths such as `/init` and see if it can load a program there. So you see Linux is forgiven for administrators.

```rust
argv.insert(0, CString::new(executable_path).unwrap()); 
```

Given the immaturity of Asterinas, not all precedent cases can be used to justify design decisions. Some existing design decisions are simply not well thought.

Back to our case. Ignoring a broken file and logging an warning only adds a negligible complexity compared to simply throwing a result and eventually panicking.

This is all my personal preference vs your preference. Maybe we should stop arguing and simply study and follow Linux's behavior in the name of compatibility 😄

lrh2000: > This is all my personal preference vs your preference. Maybe we should stop arguing and simply study and follow Linux's behavior in the name of compatibility 😄

I have no strong opinion or preference here. I just want to emphasize that initramfs/kcmdline should be **signed** so that it's not bad to trust it. For example, you will not be able to change your smartphone's kernel command line or kernel initramfs unless you unlock your phone's bootloader.

For the initramfs, it would be fine if we just skip corrupted entries like the code snippet I mentioned below? (though I said it's out of the original scope of this PR, but we can do it in a separate commit). Then I agree that `device_id_from_metadata` should be made fallible.

tatetian: > I just want to emphasize that initramfs/kcmdline should be **signed** so that it's not bad to trust it.

Of course. The initramfs is trusted. But this fact is not quite relevant since we are not discussing a security issue, but a usability/compatibility issue.

kernel/src/fs/rootfs.rs
tatetian: Yes, your code snippet above reflected my intent.

My opinion does not really matter. This is a user-visible behavior, so following Linux's behavior is the safest bet.

lrh2000: > My opinion does not really matter. This is a user-visible behavior, so following Linux's behavior is the safest bet.

https://elixir.bootlin.com/linux/v6.19.8/source/init/initramfs.c#L385-L387

Linux will (silently!) ignore the errors here. So, we can do so and print warnings. I assume.

StevenJiang1110: > Linux will (silently!) ignore the errors here. So, we can do so and print warnings. I assume.

This problem is fixed in the second commit.

tatetian:
@boterinas codex

kernel/src/fs/rootfs.rs
```diff
@@ -28,81 +29,106 @@ impl Read for BoxedReader<'_> {
...
+
+    while let Some(entry_result) = decoder.next() {
         let mut entry = entry_result?;
+        if let Err(e) = try_append_entry_to_rootfs(&mut entry, path_resolver) {
+            warn!(
+                "[kernel] failed to add entry {} to rootfs: {:?}",
+                entry.name(),
+                e
+            );
+        }
```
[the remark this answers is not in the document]
StevenJiang1110: This is intentional. If the init process fails to spawn, the kernel will panic during subsequent checks. Any errors encountered while appending entries are logged.

Issue #3034, opened 2026-03-14
Remove `-Zbuild-std` from `COMMON_CARGO_ARGS` to reduce build time


zjp-CN:
### My Question

When osdk invokes cargo subcommands like build, check, test, or clippy, build-std arguments are passed to rustc.

https://github.com/asterinas/asterinas/blob/3f3ad398cd8ec9016aa4f32cd18e44a56cc4a2c5/osdk/src/commands/util.rs#L7-L10

Asterinas doesn't seem to ship its own libstd with custom target, right? Pulling the core, alloc, and compiler_builtins to the build dependency graph increases the build time on osdk/cargo subcommands.

I find these arguments have been around since osdk was [introduced](https://github.com/asterinas/asterinas/pull/642/changes#diff-2befa357ff3b25a46b5bc159310f1f159c0ca8d3e5e945fa27df8ba3b3932d7a).

tatetian:
Great. I think you have identified a point of improvement. I arrived at the same answer as Claude Code, but its answer gave more detail. So I just paste below.

---

**Short answer:** The `-Zbuild-std` in `COMMON_CARGO_ARGS` is a historical leftover from when OSDK used a custom `x86_64-custom.json` target. It could likely be removed for the main kernel/ostd build today.

**Details:**

OSDK originally (pre-Feb 2024) used a custom target specification `x86_64-custom.json`. Custom targets have no pre-compiled standard libraries shipped by rustup, so `-Zbuild-std=core,alloc,compiler_builtins` was required. When https://github.com/asterinas/asterinas/commit/7eac2772 ("Gone with x86_64-custom") switched to the standard `x86_64-unknown-none` target, the build-std flags in `COMMON_CARGO_ARGS` were carried over without being reconsidered.

Today, all three standard targets used by Asterinas have pre-compiled `core`, `alloc`, and `compiler_builtins` available via rustup:

- `x86_64-unknown-none`
- `riscv64imac-unknown-none-elf`
- `loongarch64-unknown-none-softfloat`

The `compiler-builtins-mem` feature (which provides `memcpy`, `memset`, `memmove`, etc.) is also already included in the pre-compiled `compiler_builtins` for these targets — I verified this by inspecting the rlib symbols. I also confirmed that `cargo check --target x86_64-unknown-none -p ostd` succeeds without any `-Zbuild-std` flags.

[...]

PR #3035, opened 2026-03-15
Remove `-Zbuild-std` from normal build

zjp-CN:
Close https://github.com/asterinas/asterinas/issues/3034

~~Asterinas uses the abort panic strategy by default (??)~~, while catching `#[should_panic]` tests requires the unwind strategy. Therefore, this PR removes `COMMON_CARGO_ARGS` from the normal build process and adjusts the `-Zbuild-std` arguments for testing.

The specification of bare-metal targets typically defines the abort panic strategy, causing the pre-compiled libcore to panic with abort. However, the ktest framework requires unwinding to catch `#[should_panic]` tests. Although OSDK passes `-Cpanic=unwind` to rustc when building crates, libcore must be rebuilt to apply the unwind strategy to itself. Consequently, `-Zbuild-std` is still passed when building the testing kernel.

**NOTE:** In my experiments, `compiler_builtins` did not require recompilation; this may have been a historical workaround for custom targets to handle missing symbols.

### DIFF
--- a/osdk/src/commands/build/mod.rs
+++ b/osdk/src/commands/build/mod.rs
@@ -13,7 +13,7 @@ use std::{
 
 use bin::make_elf_for_qemu;
 
-use super::util::{COMMON_CARGO_ARGS, DEFAULT_TARGET_RELPATH, cargo, profile_name_adapter};
+use super::util::{DEFAULT_TARGET_RELPATH, OSDK_TEST_CARGO_ARGS, cargo, profile_name_adapter};
 use crate::{
     arch::Arch,
     base_crate::{BaseCrateType, new_base_crate},
@@ -261,7 +261,7 @@ fn build_kernel_elf(
     command
         .arg("--target-dir")
         .arg(cargo_target_directory.as_ref());
-    command.args(COMMON_CARGO_ARGS);
+    command.args(OSDK_TEST_CARGO_ARGS);
     command.arg("--profile=".to_string() + profile);
     for override_config in override_configs {
         command.arg("--config").arg(override_config);
--- a/osdk/src/commands/mod.rs
+++ b/osdk/src/commands/mod.rs
@@ -26,7 +26,7 @@ use crate::{
 /// The `cfg_ktest` parameter controls whether `cfg(ktest)` is enabled.
 pub fn execute_forwarded_command(subcommand: &str, args: &Vec<String>, cfg_ktest: bool) {
     let mut cargo = util::cargo();
-    cargo.arg(subcommand).args(util::COMMON_CARGO_ARGS);
+    cargo.arg(subcommand);
     if !args.contains(&"--target".to_owned()) {
         cargo.arg("--target").arg(get_default_arch().triple());
     }
--- a/osdk/src/commands/util.rs
+++ b/osdk/src/commands/util.rs
@@ -4,10 +4,13 @@ use std::process::Command;
 
 use crate::util::{get_kernel_crate, new_command_checked_exists};
 
-pub const COMMON_CARGO_ARGS: &[&str] = &[
-    "-Zbuild-std=core,alloc,compiler_builtins",
-    "-Zbuild-std-features=compiler-builtins-mem",
-];
+/// Since the pre-compiled `libcore` distribution for bare-metal targets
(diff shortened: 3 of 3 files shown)


zjp-CN:
```rust
test ostd::mm::frame::test::segment::max_segment_creation ...An uncaught panic occurred: PanicInfo {
    message: "attempt to multiply with overflow",
    file: "/__w/asterinas/asterinas/ostd/src/mm/frame/allocator.rs",
    line: 90,
    col: 46,
}
ERROR: Testing ostd failed

test aster_util::fixed_point::tests::division_by_zero ...An uncaught panic occurred: PanicInfo {
    message: "attempt to divide by zero",
    file: "/__w/asterinas/asterinas/kernel/libs/aster-util/src/fixed_point.rs",
    line: 152,
    col: 22,
}
ERROR: Testing kernel/libs/aster-util failed
```

Thare are two "should_panic" tests failing: [max_segment_creation](https://github.com/asterinas/asterinas/blob/3f3ad398cd8ec9016aa4f32cd18e44a56cc4a2c5/ostd/src/mm/frame/test.rs#L311) and  [division_by_zero](https://github.com/asterinas/asterinas/blob/3f3ad398cd8ec9016aa4f32cd18e44a56cc4a2c5/kernel/libs/aster-util/src/fixed_point.rs#L243) in the CI. Strange  🤔

~~May I have a rerun on [basic-test (ktest)](https://github.com/asterinas/asterinas/actions/runs/23107884933/job/67120223477) to see if it's deterministic or not?~~ The fails are deterministic. I'm checking what's going on here.

zjp-CN:
I'm having trouble digging into this problem.

`should_panic` tests rely on the unwinding mechanism. Once a panic occurs in a test, the `panic_handler` takes over the `PanicInfo`. On x64, Asterinas delegates [`begin_panic`](https://github.com/asterinas/asterinas/blob/36b3d34a924db299e7c9955d0d6f7439f41c69f3/osdk/deps/test-kernel/src/lib.rs#L71) to the `unwinding` crate. Normally, `begin_panic` would jump back to [`test.run(catch_unwind)`](https://github.com/asterinas/asterinas/blob/36b3d34a924db299e7c9955d0d6f7439f41c69f3/osdk/deps/test-kernel/src/lib.rs#L151). However, after this commit, `begin_panic` returns [`END_OF_STACK`](https://docs.rs/unwinding/latest/unwinding/abi/struct.UnwindReasonCode.html#associatedconstant.END_OF_STACK), indicating the panic was not caught—causing QEMU to simply power off.

I spent some time tracing the exact unwinding behavior using GDB (and LLDB), but it seems debuggers hardly work with Asterinas? This might be a skill issue on my part, but I'd like to know if debugging is actually supported. I'm particularly stuck because `file <kernel-ELF>` reports `ELF 64-bit LSB executable, Intel 80386, version 1 (SYSV), statically linked, with debug_info, not stripped`. Some tools appear to misidentify the ELF as x32 (possibly due to the "Intel 80386" header field) and disassemble it using 32-bit registers.

junyang-zh:
> but I'd like to know if debugging is actually supported.

It's supported. Personally I do `make gdb_server` to run the kernel and `make gdb_client` in another terminal session to debug it. Just use the development image and it works. Perhaps you are using a custom debugger?

zjp-CN:
I've updated the PR description.

The failing workflow was caused by a transient network error. A rerun is needed to verify whether it passes.

> However, after this commit, `begin_panic` returns `END_OF_STACK`.

This is because certain panic symbols, such as `panic_fmt`, moved from `.eh_frame` to `.debug_frame` after build-std was removed, while the unwinding crate failed to find backtrace frames from `.eh_frame` when the `fde-gnu-eh-frame-hdr` feature is enabled. 

> Asterinas uses the abort panic strategy by default.

I'm uncertain whether this observation is valid, but the `unwinding` crate is not an optional dependency of `ostd`, which suggests that Asterinas uses the unwind strategy by default. If unwinding is an intentional design choice, this PR can be closed to preserve the current behavior.

zjp-CN:
Close this as unwinding is a feature of Asterinas. ([Evidence.](https://github.com/arceos-org/arceos/issues/320#issuecomment-4141539122))

PR #3038, opened 2026-03-16
Fallback to default init executables when `init=` is omitted

StevenJiang1110:
If the kernel command line does not specify `init=`, the kernel should fall back to trying a set of default paths to locate the init process, rather than panicking immediately. This PR fixes this issue.

This PR also addresses another minor issue: following the Linux convention, the executable file name should always be prepended as `argv[0]` when launching the init process.

Required by #2911.

### DIFF
--- a/kernel/src/process/process/init_proc.rs
+++ b/kernel/src/process/process/init_proc.rs
@@ -26,20 +26,54 @@ use crate::{
 
 /// Creates and schedules the init process to run.
 pub fn spawn_init_process(
-    executable_path: &str,
+    executable_path: Option<&str>,
     argv: Vec<CString>,
     envp: Vec<CString>,
 ) -> Result<Arc<Process>> {
-    // Ensure the path for init process executable is absolute.
-    debug_assert!(executable_path.starts_with('/'));
-
-    let process = create_init_process(executable_path, argv, envp)?;
+    let process = if let Some(executable_path) = executable_path {
+        create_init_process(
+            executable_path,
+            with_init_argv0(executable_path, argv),
+            envp,
+        )?
+    } else {
+        create_default_init_process(argv, envp)?
+    };
 
     set_session_and_group(&process);
 
     process.run();
 
     Ok(process)
+}
+
+fn create_default_init_process(argv: Vec<CString>, envp: Vec<CString>) -> Result<Arc<Process>> {
+    // Linux probes the fallback init executables in this order:
+    // <https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1634>.
+    const DEFAULT_INIT_EXEC_PATHS: &[&str] = &["/sbin/init", "/etc/init", "/bin/init", "/bin/sh"];
+
+    let mut last_error = None;
+
+    for default_init_exec_path in DEFAULT_INIT_EXEC_PATHS {
+        // FIXME: Avoid cloning `argv` and `envp` for each fallback candidate.
+        match create_init_process(
+            default_init_exec_path,
+            with_init_argv0(default_init_exec_path, argv.clone()),
+            envp.clone(),
+        ) {
+            Ok(process) => return Ok(process),
+            Err(error) => last_error = Some(error),
+        }
+    }
+
+    Err(last_error.unwrap())
+}
+
+fn with_init_argv0(executable_path: &str, mut argv: Vec<CString>) -> Vec<CString> {
+    // Linux prepends the init executable path as `argv[0]`.
+    // Reference: <https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1491>.
+    argv.insert(0, CString::new(executable_path).unwrap());
+    argv
 }
 
 fn create_init_process(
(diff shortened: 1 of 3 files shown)


kernel/src/process/process/init_proc.rs
```diff
@@ -26,14 +26,22 @@ use crate::{
...
     argv: Vec<CString>,
     envp: Vec<CString>,
 ) -> Result<Arc<Process>> {
-    // Ensure the path for init process executable is absolute.
-    debug_assert!(executable_path.starts_with('/'));
-
-    let process = create_init_process(executable_path, argv, envp)?;
+    let process = if let Some(executable_path) = executable_path {
+        // Ensure the path for init process executable is absolute.
+        debug_assert!(executable_path.starts_with('/'));
```
lrh2000: I share the same concern as Copilot. What is the purpose of using `debug_assert!` here? Will Linux accept a relative path?

kernel/src/process/process/init_proc.rs
```diff
@@ -42,6 +50,34 @@ pub fn spawn_init_process(
     Ok(process)
 }
 
+fn create_default_init_process(argv: Vec<CString>, envp: Vec<CString>) -> Result<Arc<Process>> {
+    // Linux probes the fallback init executables in this order:
+    // <https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1634>.
+    let default_init_exec_paths = ["/sbin/init", "/etc/init", "/bin/init", "/bin/sh"];
```
lrh2000: We can use `const` here?

If we use `const`, insert a blank line below.

kernel/src/process/process/init_proc.rs
```diff
@@ -42,6 +50,34 @@ pub fn spawn_init_process(
...
+            Err(error) => last_error = Some(error),
+        }
+    }
+
+    Err(last_error.unwrap())
+}
+
+fn with_init_argv0(executable_path: &str, mut argv: Vec<CString>) -> Vec<CString> {
+    // Linux prepends the init executable path as `argv[0]`.
+    // Reference: <https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1491>.
```
lrh2000, line 74: I assume doing so will break the qemu-direct boot CI. The details are discussed in https://github.com/asterinas/asterinas/issues/2982#issuecomment-3970964895.

However, those CIs have failed already due to https://github.com/asterinas/asterinas/pull/3019...

kernel/src/process/process/init_proc.rs
```diff
@@ -42,6 +50,34 @@ pub fn spawn_init_process(
...
 }
 
+fn create_default_init_process(argv: Vec<CString>, envp: Vec<CString>) -> Result<Arc<Process>> {
+    // Linux probes the fallback init executables in this order:
+    // <https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1634>.
+    let default_init_exec_paths = ["/sbin/init", "/etc/init", "/bin/init", "/bin/sh"];
+    let mut last_error = None;
+
+    for default_init_exec_path in default_init_exec_paths {
+        // FIXME: Avoid cloning `argv` and `envp` for each fallback candidate.
```
lrh2000, line 58: This is trivial to fix. I am not sure why we want to leave a FIXME.

kernel/src/process/process/init_proc.rs
StevenJiang1110: Sorry, I don't quite follow. The comment seems to be referring to an issue with the kernel command line, but that doesn't appear to be related to the `argv[0]` of the init process. Could you elaborate a bit more?

kernel/src/process/process/init_proc.rs
StevenJiang1110: This check was inherited from the old code and I haven't modified it. After looking at the Linux implementation, I found that Linux does not enforce whether the path is absolute or relative, so I'll remove this check.

kernel/src/process/process/init_proc.rs
StevenJiang1110: Since `create_new_task` and `ProgramToLoad` take ownership of the argv vector, and `ProgramToLoad` further parses the shebang's command line to append additional arguments to it, I haven't been able to come up with a clean fix for this. Cloning `envp` may indeed be unnecessary, but I don't see a good way to avoid cloning `argv`. Do you have any suggestions?

lrh2000: > `create_new_task` and `ProgramToLoad` take ownership of the argv vector, and `ProgramToLoad` further parses the shebang's command line to append additional arguments to it

Oh, I see the problem. It isn't as simple as I thought when I first skimmed the code. I should have thought about it more carefully. Thanks for your explanation.

kernel/src/process/process/init_proc.rs
lrh2000: > doesn't appear to be relate

My mistake. It doesn't seem to affect anything. I'm sorry for the noise.

PR #3053, opened 2026-03-19
Fix pipe handling and futex wake matching

cqs21:
## Summary

This PR fixes two issues exposed by running a small Java reproducer on Asterinas NixOS:

1. `FileInputStream.available()` on a pipe could fail with `Illegal seek`.
2. After fixing the pipe issue, the Java process could still hang in `Process.waitFor()` and never exit.

The fixes are split into two commits on purpose:

- the first commit fixes `FIONREAD` on pipe handles and removes the `Illegal seek` failure;
- the second commit fixes `futex_wake` matching across address spaces and removes the remaining hang.

Together they make the reproducer complete successfully.

## Bug Reproducer

The following Java program reproduces the problem. It starts a child Java process, reads its combined stdout/stderr through a pipe, and then waits for the child to exit.

```java
import java.io.BufferedReader;
import java.io.InputStreamReader;
import java.io.IOException;

public class GetJdkProps {
    public static void main(String[] args) throws Exception {
        String jdkHome = System.getenv("JAVA_HOME");
        if (jdkHome == null || jdkHome.isEmpty()) {
            System.err.println("JAVA_HOME is not set.");
            return;
        }

        String javaCmd = jdkHome + "/bin/java";
        ProcessBuilder pb = new ProcessBuilder(
                javaCmd,
                "-XshowSettings:properties",
                "-version"
        );
        pb.redirectErrorStream(true);

        Process p = pb.start();

        try (BufferedReader br = new BufferedReader(
                new InputStreamReader(p.getInputStream()))) {
            String line;
            while ((line = br.readLine()) != null) {
                System.out.println(line);
            }
        }

        int exitCode = p.waitFor();
        System.out.println("Exit code: " + exitCode);
    }
}
```

This can be reproduced in the NixOS environment as follows.

If Java is not already enabled in the NixOS image, add the following option to `distro/etc_nixos/configuration.nix`:

```nix
programs.java.enable = true;
```

Then build and run NixOS:

```bash
make nixos
make run_nixos
```

Inside the NixOS shell, save the program above as `GetJdkProps.java`, then run:

```bash
javac GetJdkProps.java
java GetJdkProps
```

### Without either commit

Running `java GetJdkProps` fails with:

```text
Exception in thread "main" java.io.IOException: Illegal seek
```

This happens while OpenJDK is reading the child process output pipe.

### With only the first commit

After applying the first commit, the `Illegal seek` error is gone and the child Java process output can be read successfully.

However, the reproducer still does not finish: it prints the child output, and then hangs instead of returning from `waitFor()`.

### With both commits

After additionally applying the second commit, the reproducer completes normally.

## Root Cause

This issue turned out to have two independent parts.

### 1. Pipe `FIONREAD` support was missing

OpenJDK probes process pipes via `available()`. On Unix, this path expects `FIONREAD` to work on pipe file descriptors.

Before the first commit, pipe handles did not implement this ioctl, so OpenJDK fell back to probing seekability and hit `lseek(..., SEEK_CUR)` on a pipe, which correctly returns `ESPIPE`. That surfaced in Java as `java.io.IOException: Illegal seek`.

The first commit fixes this by implementing `FIONREAD` for pipe handles and returning the current readable byte count from the pipe buffer.

### 2. Futex wakeups were not matched correctly across address spaces

After the pipe issue was fixed, the reproducer still hung while waiting for the child process to exit.

The remaining problem was in futex keying and wake matching. The old logic relied on a hash derived from virtual address and process-local information, which is insufficient for correct matching once different address spaces or shared mappings are involved. In particular, `clear_child_tid` wakeups need to wake the waiter in the right futex namespace rather than anything that happens to collide by hash.

The second commit fixes this by:

- distinguishing private and shared futex namespaces explicitly with `FutexIdentity`;
- comparing futex identity directly during wake matching instead of comparing only hashes;

This makes the `clear_child_tid` wake target the correct waiter, so `waitFor()` can return normally.

### DIFF
--- a/kernel/src/context.rs
+++ b/kernel/src/context.rs
@@ -13,7 +13,7 @@ use ostd::{
 use crate::{
     prelude::*,
     process::{
-        Process,
+        Process, VmarSnapshot,
         posix_thread::{PosixThread, ThreadLocal},
     },
     thread::Thread,
@@ -84,6 +84,11 @@ impl<'a> CurrentUserSpace<'a> {
     /// This method will panic if the current process has cleared its `Vmar`.
     pub fn vmar(&self) -> &Vmar {
         self.0.as_ref().unwrap()
+    }
+
+    /// Takes a snapshot of the current VMAR identity.
+    pub fn vmar_snapshot(&self) -> VmarSnapshot {
+        VmarSnapshot::from(Arc::downgrade(self.0.as_ref().unwrap()))
     }
 
     /// Returns whether the VMAR is shared with other processes or threads.
--- a/kernel/src/fs/pipe/common.rs
+++ b/kernel/src/fs/pipe/common.rs
@@ -23,7 +23,10 @@ use crate::{
             signals::user::{UserSignal, UserSignalKind},
         },
     },
-    util::ring_buffer::{RbConsumer, RbProducer, RingBuffer},
+    util::{
+        ioctl::{RawIoctl, dispatch_ioctl},
+        ring_buffer::{RbConsumer, RbProducer, RingBuffer},
+    },
 };
 
 /// A handle for a pipe that implements `FileIo`.
@@ -52,6 +55,10 @@ impl PipeHandle {
         debug_assert!(self.access_mode.is_writable());
 
         self.inner.writer.try_write(reader)
+    }
+
+    fn bytes_to_read(&self) -> usize {
+        self.inner.reader.buffer_len()
     }
 }
 
@@ -136,6 +143,20 @@ impl FileIo for PipeHandle {
 
     fn is_offset_aware(&self) -> bool {
         false
+    }
+
+    fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
+        use crate::util::ioctl::common_defs::GetNumBytesToRead;
+
+        dispatch_ioctl!(match raw_ioctl {
+            cmd @ GetNumBytesToRead => {
+                let bytes_to_read = self.bytes_to_read() as i32;
+
+                cmd.write(&bytes_to_read)?;
+                Ok(0)
+            }
+            _ => return_errno_with_message!(Errno::ENOTTY, "the ioctl command is unknown"),
+        })
     }
 }
 
@@ -384,6 +405,10 @@ impl PipeReader {
         };
 
         self.state.read_with(read)
+    }
+
+    fn buffer_len(&self) -> usize {
+        self.consumer.lock().len()
     }
 
     fn peer_shutdown(&self) {
--- a/kernel/src/process/posix_thread/exit.rs
+++ b/kernel/src/process/posix_thread/exit.rs
@@ -3,7 +3,9 @@
 use ostd::{mm::VmIo, task::Task};
 
 use super::{
-    AsPosixThread, AsThreadLocal, ThreadLocal, futex::futex_wake, robust_list::wake_robust_futex,
+    AsPosixThread, AsThreadLocal, ThreadLocal,
+    futex::{FutexVisibility, futex_wake},
+    robust_list::wake_robust_futex,
 };
 use crate::{
     current_userspace,
@@ -124,7 +126,7 @@ fn wake_clear_ctid(thread_local: &ThreadLocal) {
     let _ = current_userspace!()
         .write_val(clear_ctid, &0u32)
         .inspect_err(|err| debug!("exit: cannot clear the child TID: {:?}", err));
-    let _ = futex_wake(clear_ctid, 1, None)
+    let _ = futex_wake(clear_ctid, 1, FutexVisibility::Shared)
         .inspect_err(|err| debug!("exit: cannot wake the futex on the child TID: {:?}", err));
 
     thread_local.clear_child_tid().set(0);
--- a/kernel/src/process/posix_thread/futex.rs
+++ b/kernel/src/process/posix_thread/futex.rs
@@ -262,7 +262,7 @@ impl FutexWakeOpEncode {
             FutexWakeOp::FUTEX_OP_SET => oparg,
             FutexWakeOp::FUTEX_OP_ADD => oparg.wrapping_add(old_val),
             FutexWakeOp::FUTEX_OP_OR => oparg | old_val,
-            FutexWakeOp::FUTEX_OP_ANDN => oparg & !old_val,
+            FutexWakeOp::FUTEX_OP_ANDN => !oparg & old_val,
             FutexWakeOp::FUTEX_OP_XOR => oparg ^ old_val,
         }
     }
--- a/kernel/src/process/posix_thread/futex.rs
+++ b/kernel/src/process/posix_thread/futex.rs
@@ -8,22 +8,44 @@ use ostd::{
 use spin::Once;
 
 use crate::{
+    current_userspace,
     prelude::*,
-    process::Pid,
+    process::VmarSnapshot,
     time::wait::ManagedTimeout,
-    vm::{perms::VmPerms, vmar::PageFaultInfo},
+    vm::{perms::VmPerms, vmar::PageFaultInfo, vmo::Vmo},
 };
 
 type FutexBitSet = u32;
 
 const FUTEX_BITSET_MATCH_ANY: FutexBitSet = 0xFFFF_FFFF;
 
+/// Specifies whether a futex is scoped to the current process or shared through
+/// its backing mapping.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FutexVisibility {
+    /// Uses the calling process's private futex namespace.
+    Private,
+    /// Derives the futex identity from the backing mapping when possible, and
+    /// otherwise falls back to a process-local shared identity.
+    Shared,
+}
+
+impl From<FutexFlags> for FutexVisibility {
+    fn from(flags: FutexFlags) -> Self {
+        if flags.contains(FutexFlags::FUTEX_PRIVATE) {
+            Self::Private
+        } else {
+            Self::Shared
+        }
+    }
+}
+
 pub fn futex_wait(
     futex_addr: u64,
     futex_val: i32,
     timeout: Option<ManagedTimeout>,
     ctx: &Context,
-    pid: Option<Pid>,
+    visibility: FutexVisibility,
 ) -> Result<()> {
     futex_wait_bitset(
         futex_addr as _,
@@ -31,7 +53,7 @@ pub fn futex_wait(
         timeout,
         FUTEX_BITSET_MATCH_ANY,
         ctx,
-        pid,
+        visibility,
     )
 }
 
@@ -41,7 +63,7 @@ pub fn futex_wait_bitset(
     timeout: Option<ManagedTimeout>,
     bitset: FutexBitSet,
     ctx: &Context,
-    pid: Option<Pid>,
+    visibility: FutexVisibility,
 ) -> Result<()> {
     debug!(
         "futex_wait_bitset: addr = {:#x}, val = {}, bitset = {:#x}",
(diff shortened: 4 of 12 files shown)


kernel/src/fs/pipe/common.rs
```diff
@@ -137,6 +144,17 @@ impl FileIo for PipeHandle {
     fn is_offset_aware(&self) -> bool {
         false
     }
+
+    fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
+        dispatch_ioctl!(match raw_ioctl {
```
lrh2000, line 151: https://github.com/asterinas/asterinas/blob/36b3d34a924db299e7c9955d0d6f7439f41c69f3/kernel/src/fs/vfs/notify/inotify.rs#L342-L344

Let's use the same style here. Move the `ioctl_def` import in the method and insert a blank line before `dispatch_ioctl!`.

kernel/src/process/posix_thread/exit.rs
```diff
@@ -124,7 +124,8 @@ fn wake_clear_ctid(thread_local: &ThreadLocal) {
     let _ = current_userspace!()
         .write_val(clear_ctid, &0u32)
         .inspect_err(|err| debug!("exit: cannot clear the child TID: {:?}", err));
-    let _ = futex_wake(clear_ctid, 1, None)
+    let current_vmar = thread_local.vmar().borrow().as_ref().unwrap().clone();
```
lrh2000: Please read this note:
https://github.com/asterinas/asterinas/blob/36b3d34a924db299e7c9955d0d6f7439f41c69f3/kernel/src/process/process_vm/mod.rs#L163-L167

`Arc<Vmar>` cannot be cloned. Try to find a way to avoid the clone.

kernel/src/process/posix_thread/futex.rs
```diff
@@ -261,12 +276,23 @@ pub fn futex_wake_op(
...
 
-    let futex_key_1 = FutexKey::new(futex_addr_1, FUTEX_BITSET_MATCH_ANY, pid)?;
-    let futex_key_2 = FutexKey::new(futex_addr_2, FUTEX_BITSET_MATCH_ANY, pid)?;
+    let current_vmar = ctx.thread_local.vmar().borrow().as_ref().unwrap().clone();
+    let futex_key_1 = FutexKey::new(
+        futex_addr_1,
+        FUTEX_BITSET_MATCH_ANY,
+        &current_vmar,
+        is_private,
+    )?;
```
lrh2000: For example, I don't think it is really necessary to clone the VMAR here. You can use something like:
```rust
```
Then obtain access to the current VMAR via `user_space.vmar()`. There is no need to clone the `Arc`, which is no only problematic but also inefficient.

kernel/src/process/posix_thread/futex.rs
```diff
@@ -118,15 +121,27 @@ pub fn futex_wait_bitset(
...
+        is_private,
+    )
 }
 
 pub fn futex_wake_bitset(
     futex_addr: Vaddr,
     max_count: usize,
     bitset: FutexBitSet,
-    pid: Option<Pid>,
+    current_vmar: &Arc<Vmar>,
```
lrh2000: Could you please explain why you added `current_vmar` to `futex_*` methods? I am not sure about your motivation. Why not just using `current_userspace!().vmar()` in `futex_wake_bitset`?

kernel/src/process/posix_thread/futex.rs
```diff
@@ -515,21 +544,30 @@ impl FutexItem {
     }
 }
 
-/// The key of a futex used to mark a futex word.
+/// The identity used for futex namespace and backing selection.
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum FutexIdentity {
+    Private { vmar_id: usize, addr: Vaddr },
+    SharedLocal { vmar_id: usize, addr: Vaddr },
+    Shared { object_id: usize, offset: usize },
```
lrh2000: I was wondering if we should hold some weak references like `VmarSnapshot` and `Weak<Vmo>` here to avoid the `vmar_id` and `object_id` to be reused. Do you think it is a real problem? Perhaps it is not necessary if the spurious wakeup is acceptable?

kernel/src/process/posix_thread/futex.rs
```diff
@@ -540,21 +578,62 @@ impl FutexKey {
...
+        let identity = if is_private {
+            FutexIdentity::Private {
+                vmar_id: Arc::as_ptr(current_vmar) as usize,
+                addr,
+            }
+        } else {
+            current_vmar
+                .query(addr..addr + 1)
+                .iter()
+                .find(|mapping| mapping.map_to_addr() <= addr && addr < mapping.map_end())
```
lrh2000: Shouldn't you just need to `next()` since you `query(addr..addr + 1)`?

test/initramfs/src/apps/process/futex/clear_child_tid.c
lrh2000, line 1: As shared futex is an important feature, I would expect gVisor or LTP tests to have good coverage. Have you investigated these tests? Could you enable some of them? Will they cover this test?

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -171,6 +171,23 @@ impl VmMapping {
         }
     }
 
+    /// Returns the shared futex backing identity for the address if available.
+    pub(crate) fn futex_backing(&self, addr: Vaddr) -> Option<(usize, usize)> {
```
lrh2000: No need to use `pub(crate)`. The kernel crate has no downstream users. Use `pub` instead.

lrh2000 (commented):
I have wanted to fix the shared futex problem for a long time. Unfortunately, I have never found the time to take on that task. Thank you for adding support for shared futexes — I think this is an important feature.

kernel/src/process/posix_thread/exit.rs
cqs21: > `Arc<Vmar>` cannot be cloned. Try to find a way to avoid the clone.

Thanks for the input. The futex wake helpers no longer take a cloned `Arc<Vmar>` now.

test/initramfs/src/apps/ipc/pipe/process_pipe_available.c
```diff
@@ -0,0 +1,51 @@
...
+
+	TEST_SUCC(close(fildes[1]));
+	TEST_RES(waitpid(pid, NULL, 0), _ret == pid);
+
+	TEST_SUCC(fstat(fildes[0], &stat_buf));
+	TEST_RES(S_ISFIFO(stat_buf.st_mode), _ret != 0);
+
+	int pending = -1;
+	TEST_SUCC(ioctl(fildes[0], FIONREAD, &pending));
+	TEST_RES(pending, _ret == 3);
```
lrh2000:
```suggestion
	TEST_RES(ioctl(fildes[0], FIONREAD, &pending), pending == 3);
```

test/initramfs/src/apps/ipc/pipe/process_pipe_available.c
```diff
@@ -0,0 +1,51 @@
...
+		CHECK(close(fildes[0]));
+		CHECK_WITH(write(fildes[1], "abc", 3), _ret == 3);
+		_exit(0);
+	}
+
+	TEST_SUCC(close(fildes[1]));
+	TEST_RES(waitpid(pid, NULL, 0), _ret == pid);
+
+	TEST_SUCC(fstat(fildes[0], &stat_buf));
+	TEST_RES(S_ISFIFO(stat_buf.st_mode), _ret != 0);
```
lrh2000: Same. Merge `TEST_SUCC` and `TEST_RES`.

test/initramfs/src/apps/ipc/pipe/process_pipe_available.c
```diff
@@ -0,0 +1,51 @@
...
+ */
+FN_TEST(process_pipe_available_probe)
+{
+	int fildes[2];
+	char buf[4] = { 0 };
+	struct stat stat_buf;
+
+	CHECK(pipe(fildes));
+
+	int pid = CHECK(fork());
```
lrh2000: Use `TEST_SUCC` if possible.

kernel/src/process/posix_thread/robust_list.rs
```diff
@@ -158,7 +158,7 @@ pub fn wake_robust_futex(futex_addr: Vaddr, tid: Tid) -> Result<()> {
                 // Wake up one waiter and break out from the loop.
                 if new_val & FUTEX_WAITERS != 0 {
                     debug!("wake the robust futex at {:#x}", futex_addr);
-                    futex_wake(futex_addr, 1, None)?;
+                    futex_wake(futex_addr, 1, false)?;
```
lrh2000: Consider `futex_wake(futex_addr, 1, FutexVisibility::Private)`. It should be better than passing a boolean variable?

lrh2000 (commented):
Thanks for the revision. Seems good to me other than a few minor nits.

kernel/src/process/posix_thread/robust_list.rs
cqs21: > It should be better than passing a boolean variable?

Replaced with `FutexVisibility`. Thanks for the suggestion.

test/initramfs/src/apps/ipc/pipe/process_pipe_available.c
```diff
@@ -0,0 +1,55 @@
...
+ * pipe. Supporting `FIONREAD` on pipes avoids the fallback path that would
+ * otherwise probe seekability with `lseek(fd, 0, SEEK_CUR)`.
+ */
+FN_TEST(process_pipe_available_probe)
+{
+	int fildes[2];
+	char buf[4] = { 0 };
+	struct stat stat_buf;
+
+	CHECK(pipe(fildes));
```
lrh2000:
```suggestion
	TEST_SUCC(pipe(fildes));
```

test/initramfs/src/apps/ipc/pipe/process_pipe_available.c
```diff
@@ -0,0 +1,55 @@
...
+
+	int pid = TEST_SUCC(fork());
+	if (pid < 0) {
+		TEST_SUCC(close(fildes[0]));
+		TEST_SUCC(close(fildes[1]));
+		return;
+	}
+
```
lrh2000: Actually line 29-34 is redundant...
https://github.com/search?q=repo%3Aasterinas%2Fasterinas+%22TEST_SUCC%28fork%28%22&type=code

We should avoid manually returning from a `FN_TEST`. The end of an `FN_TEST` may contain trailing code that cannot be safely removed.

If TEST_SUCC fails, let the subsequent tests within the same `FN_TEST` fail too. However, we can still see the results of the other `FN_TEST`s  (but `CHECK` will terminate the entire test file).

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -171,6 +171,22 @@ impl VmMapping {
...
 
+    /// Returns the shared futex backing identity for the address if available.
+    pub fn futex_backing(&self, addr: Vaddr) -> Option<(Weak<Vmo>, usize)> {
+        if !self.is_shared {
+            return None;
+        }
+
+        let mapped_vmo = match &self.mapped_mem {
+            MappedMemory::Vmo(mapped_vmo) => mapped_vmo,
+            _ => return None,
```
lrh2000: What about shared device mappings?

Ideally, maybe we should fail since we don't support it. Similar to https://elixir.bootlin.com/linux/v6.17/source/kernel/futex/core.c#L673-L675.

Alternatively, having a FIXME will certainly also help.

kernel/src/context.rs
```diff
@@ -86,6 +86,11 @@ impl<'a> CurrentUserSpace<'a> {
         self.0.as_ref().unwrap()
     }
 
+    /// Takes a snapshot of the current VMAR identity.
+    pub fn vmar_snapshot(&self) -> VmarSnapshot {
+        VmarSnapshot::from(self.0.as_ref().map(Arc::downgrade).unwrap_or_default())
```
lrh2000: It does not seem appropriate to fall back to the default here. Defaulting is also a broken behaviour for this type, since VMAR should not be `None`.
```rust
        VmarSnapshot::from(Arc::dowgrade(self.0.as_ref().unwrap()))
```

lrh2000 (commented):
Thank you for addressing the comments.

lrh2000 (approved):

lrh2000:
Looks like you should resolve the conflicts.

cchanging (approved):

PR #3056, opened 2026-03-20
Enable `qemu-direct` support for `linux-efi-*` boot protocols

taosue:
This PR re-applies the changes originally introduced in #3019 and adds OVMF when using `linux-efi-*` boot protocols with `qemu-direct` boot method.

### DIFF
--- a/tools/qemu_args.sh
+++ b/tools/qemu_args.sh
@@ -165,6 +165,14 @@ if [ "$BOOT_METHOD" = "qemu-direct" ]; then
     fi
 fi
 
+# When using `grub-rescue-iso` or `grub-qcow2` boot, OVMF must be enabled.
+# Currently, the project's `grub-mkrescue` (in container image) only contained
+# `x86_64-efi` platform modules — no `i386-pc`. This meant the generated ISO/qcow2
+# could only be loaded by OVMF.
+if [ "$BOOT_METHOD" = "grub-rescue-iso" ] || [ "$BOOT_METHOD" = "grub-qcow2" ]; then
+    OVMF="on"
+fi
+
 if [ "$OVMF" = "on" ]; then
     if [ "$1" = "microvm" ]; then
         QEMU_ARGS="${QEMU_ARGS} \
--- a/tools/qemu_args.sh
+++ b/tools/qemu_args.sh
@@ -7,7 +7,8 @@
 #  - scheme: "normal", "test", "microvm" or "iommu";
 # Other arguments are configured via environmental variables:
 #  - OVMF: "on" or "off";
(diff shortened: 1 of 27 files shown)


tatetian:
@junyang-zh Could you review this PR?

OSDK.toml
```diff
@@ -33,7 +33,7 @@ args = "$(./tools/qemu_args.sh test)"
 [scheme."microvm"]
 boot.method = "qemu-direct"
 build.strip_elf = true
-qemu.args = "$(./tools/qemu_args.sh microvm)"
+qemu.args = "$(BOOT_METHOD=qemu-direct ./tools/qemu_args.sh microvm)"
```
StevenJiang1110: Move this to the root Makefile.

```Makefile
ifeq ($(SCHEME), microvm)
BOOT_METHOD=qemu-direct
endif

```

tools/qemu_args.sh
```diff
@@ -113,62 +114,65 @@ if [ "$1" = "iommu" ]; then
...
-    echo $QEMU_ARGS
-    exit 0
+# When using qemu-direct boot, OVMF depends on the boot protocol:
+# linux-efi-* protocols require OVMF; other protocols (e.g. multiboot) do not.
+if [ "$BOOT_METHOD" = "qemu-direct" ]; then
+    if [ "$BOOT_PROTOCOL" = "linux-efi-pe64" ] || [ "$BOOT_PROTOCOL" = "linux-efi-handover64" ]; then
+        OVMF="on"
+    else
+        OVMF="off"
+    fi
```
StevenJiang1110, line 165: The current handling of OVMF feels a bit puzzling — though this is likely a pre-existing issue. Could you clarify when OVMF is actually required? For example, if the boot protocol is linux-efi-handover64 but the boot method is grub-qcow2 or grub-rescue-iso, is OVMF still necessary? And in which cases is OVMF optional — i.e., where it may or may not be present?

StevenJiang1110 (commented):
Thanks for your contribution!

tools/qemu_args.sh
taosue: Currently, the project's `grub-mkrescue` (in container image) only contained `x86_64-efi` platform modules — no `i386-pc`. This meant the generated ISO/qcow2 could only be loaded by a UEFI firmware (OVMF), even when `BOOT_PROTOCOL=multiboot2`.

If we support `SeaBIOS+GRUB` in the future, `OVMF=off` is still fundamentally incompatible with `BOOT_PROTOCOL=linux-efi-handover64` or `linux-efi-pe64`, because those protocols require a live UEFI environment that OVMF provides. For `multiboot/multiboot2`, OVMF is optional.

StevenJiang1110: Thanks for your investigation! The new changes make sense to me.

StevenJiang1110 (approved):

Issue #2323, opened 2025-08-10
Tracking Issue for supporting GDB


vvvvsv:
This issue tracks all issues related to GDB support. 

Currently, Asterinas still lacks some features required to support GDB.

## Missing syscalls

- [ ] `ptrace`
- [ ] `personality`
- [ ] `tkill`
- [ ] `vfork`      

## Missing `/proc` files

 - [ ] `/proc/<pid>/task/<tid>/maps`
 - [ ] `/proc/<pid>/task/<tid>/mem`
 - [ ] `/proc/<pid>/task/<tid>/stat` is currently incomplete.

StevenJiang1110:
Does this issue also intend to track GDB’s breakpoint support (see http://www.nynaeve.net/?p=80)? If so, on x86 we may need to support either software breakpoints—by handling the debug exception caused by int3, or hardware breakpoints—by using the [x86 debug registers](https://en.wikipedia.org/wiki/X86_debug_register).

vvvvsv:
> Does this issue also intend to track GDB’s breakpoint support (see http://www.nynaeve.net/?p=80)?

Thanks for your advice! Yes, GDB mainly uses software breakpoints (int3), and this issue will also track that requirement.

vvvvsv:
# Current Progress

## A `ptrace` framework

PR #2380 provides an incomplete `ptrace` framework. According to the Linux [spec](https://man7.org/linux/man-pages/man2/ptrace.2.html), the tracer-tracee relationship is per-thread.

> ..."tracee" always means "(one) thread", never "a
> (possibly multithreaded) process"...

Meanwhile, according to experiments on Linux, different threads within the same process can trace distinct threads. GDB typically does not use `ptrace` in this way. I plan to organize the required information as follows:

```rust
struct PosixThread {
    // ...
    
    /// Status of being traced.
    tracee_status: Mutex<TraceeStatus>,

    /// Threads being traced by this thread.
    tracees: Mutex<Option<BTreeMap<Tid, Arc<Thread>>>>,
}

struct TraceeStatus {
    tracer: Option<Weak<Thread>>,
    // user_ctx, ...
}
```

## A breakpoint/single-step demo

This [branch](https://github.com/vvvvsv/asterinas/tree/1876da62dc538b65690aa4a3101fad9b836cd57a) contains a breakpoint/single-step demo, while the APIs are not aligned with Linux. It simulates the common GDB breakpoint workflow:
[...]

PR #3061, opened 2026-03-21
Support `PTRACE_SETOPTIONS`

vvvvsv:
Based on https://github.com/asterinas/asterinas/pull/3065, and required for #2323.

For reference, please refer to "PTRACE_SETOPTIONS" and "PTRACE_EVENT stops" in [spec description](https://man7.org/linux/man-pages/man2/ptrace.2.html#DESCRIPTION).

This PR does not yet implement clone-family ptrace-event-stops, because I want to focus on supporting GDB for single-threaded programs currently. In this PR, the clone calls that would trigger ptrace-event-stops are rejected.

### DIFF
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/process-and-thread-management/README.md
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/process-and-thread-management/README.md
@@ -98,11 +98,14 @@ Supported requests:
 * `PTRACE_SINGLESTEP` (x86-64 only)
 * `PTRACE_GETREGS` (x86-64 only)
 * `PTRACE_SETREGS` (x86-64 only)
+* `PTRACE_SETOPTIONS`
+* `PTRACE_GETEVENTMSG`
 * `PTRACE_GETSIGINFO`
 
-Additional limitations:
-* Only the main thread of a process can act as the tracer
-* `PTRACE_PEEKUSER` and `PTRACE_POKEUSER` only support offsets for general-purpose registers
+Limitations:
+* Only the main thread of a process can act as the tracer.
+* `PTRACE_PEEKUSER` and `PTRACE_POKEUSER` only support offsets for general-purpose registers.
+* If a tracee has clone-family options (`PTRACE_O_TRACEFORK`, `PTRACE_O_TRACEVFORK`, `PTRACE_O_TRACEVFORKDONE`, or `PTRACE_O_TRACECLONE`) and then performs a clone-family operation that would trigger a ptrace event, the clone-family operation returns `EOPNOTSUPP`.
 
 For more information,
 see [the man page](https://man7.org/linux/man-pages/man2/ptrace.2.html).
--- a/kernel/src/context.rs
+++ b/kernel/src/context.rs
@@ -23,16 +23,16 @@ use crate::{
 /// The context that can be accessed from the current POSIX thread.
 #[derive(Clone)]
 pub struct Context<'a> {
-    pub process: Arc<Process>,
-    pub thread_local: &'a ThreadLocal,
-    pub posix_thread: &'a PosixThread,
-    pub thread: &'a Thread,
+    pub(crate) process: Arc<Process>,
+    pub(crate) thread_local: &'a ThreadLocal,
+    pub(crate) posix_thread: &'a PosixThread,
+    pub(crate) thread: &'a Thread,
     pub task: &'a Task,
 }
 
 impl Context<'_> {
     /// Gets the userspace of the current task.
-    pub fn user_space(&self) -> CurrentUserSpace<'_> {
+    pub(crate) fn user_space(&self) -> CurrentUserSpace<'_> {
         CurrentUserSpace(self.thread_local.vmar().borrow())
     }
 }
--- a/kernel/src/process/clone.rs
+++ b/kernel/src/process/clone.rs
@@ -236,6 +236,16 @@ impl CloneArgs {
             );
         }
 
+        // TODO: Support clone-family ptrace events and remove this check.
+        if !clone_flags.contains(CloneFlags::CLONE_UNTRACED)
+            && ctx.posix_thread.needs_ptrace_clone_stop(self)
+        {
+            return_errno_with_message!(
+                Errno::EOPNOTSUPP,
+                "ptrace clone events are not supported currently"
+            );
+        }
+
         Ok(())
     }
 }
--- a/kernel/src/process/posix_thread/exit.rs
+++ b/kernel/src/process/posix_thread/exit.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use ostd::{mm::VmIo, task::Task};
+use ostd::{arch::cpu::context::UserContext, mm::VmIo, task::Task};
 
 use super::{
     AsPosixThread, AsThreadLocal, ThreadLocal,
@@ -25,8 +25,8 @@ use crate::{
 /// # Panics
 ///
 /// If the current thread is not a POSIX thread, this method will panic.
-pub fn do_exit(term_status: TermStatus) {
-    exit_internal(term_status, false);
+pub fn do_exit(term_status: TermStatus, ctx: &Context, user_ctx: &mut UserContext) {
+    exit_internal(term_status, false, ctx, user_ctx);
 }
 
 /// Kills all threads and exits the current POSIX process.
@@ -34,12 +34,17 @@ pub fn do_exit(term_status: TermStatus) {
 /// # Panics
 ///
 /// If the current thread is not a POSIX thread, this method will panic.
-pub fn do_exit_group(term_status: TermStatus) {
-    exit_internal(term_status, true);
+pub fn do_exit_group(term_status: TermStatus, ctx: &Context, user_ctx: &mut UserContext) {
+    exit_internal(term_status, true, ctx, user_ctx);
 }
 
 /// Exits the current POSIX thread or process.
-fn exit_internal(term_status: TermStatus, is_exiting_group: bool) {
+fn exit_internal(
+    term_status: TermStatus,
+    is_exiting_group: bool,
+    ctx: &Context,
+    user_ctx: &mut UserContext,
+) {
     let exit_code = term_status.as_u32();
 
     let current_task = Task::current().unwrap();
--- a/kernel/src/process/posix_thread/exit.rs
+++ b/kernel/src/process/posix_thread/exit.rs
@@ -5,6 +5,7 @@ use ostd::{arch::cpu::context::UserContext, mm::VmIo, task::Task};
 use super::{
     AsPosixThread, AsThreadLocal, ThreadLocal,
     futex::{FutexVisibility, futex_wake},
+    ptrace::PtraceEvent,
     robust_list::wake_robust_futex,
 };
 use crate::{
@@ -39,6 +40,10 @@ pub fn do_exit_group(term_status: TermStatus, ctx: &Context, user_ctx: &mut User
 }
 
 /// Exits the current POSIX thread or process.
+//
+// A mutable reference to `user_ctx` is needed because exiting a traced thread
+// may trigger a "exit" ptrace-event-stop, which snapshots the current user
+// register state and may later restore tracer-updated registers.
 fn exit_internal(
     term_status: TermStatus,
     is_exiting_group: bool,
@@ -46,6 +51,8 @@ fn exit_internal(
     user_ctx: &mut UserContext,
 ) {
     let exit_code = term_status.as_u32();
+    ctx.posix_thread
+        .ptrace_may_stop_on(PtraceEvent::Exit(exit_code), ctx, user_ctx);
 
     let current_task = Task::current().unwrap();
     let current_thread = current_task.as_thread().unwrap();
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -18,8 +18,7 @@ use crate::{
         signal::{
             DequeuedSignal, PauseReason,
             c_types::siginfo_t,
-            constants::{CLD_TRAPPED, SIGCHLD, SIGKILL},
-            sig_num::SigNum,
+            constants::{CLD_TRAPPED, SIGCHLD, SIGKILL, SIGTRAP},
             signals::{kernel::KernelSignal, raw::RawSignal, user::UserSignal},
         },
     },
@@ -29,7 +28,7 @@ use crate::{
 mod util;
 
 use util::StopDeliverySignal;
-pub use util::{PtraceContRequest, PtraceOptions};
+pub use util::{PtraceContRequest, PtraceOptions, PtraceWaitStatus};
 pub(in crate::process) use util::{PtraceEvent, PtraceStopResult};
 
 impl PosixThread {
@@ -85,8 +84,27 @@ impl PosixThread {
         }
     }
 
+    /// Stops this thread by ptrace on the `event` if it is currently traced,
+    /// and the corresponding option is enabled.
+    ///
+    /// May block in the event-stop until the tracer continues the stop,
+    /// or until a `SIGKILL` interrupts it.
+    pub(in crate::process) fn ptrace_may_stop_on(
+        &self,
+        event: PtraceEvent,
+        ctx: &Context,
+        user_ctx: &mut UserContext,
+    ) {
+        if let Some(status) = self.tracee_status.get() {
+            status.ptrace_may_stop_on(event, ctx, user_ctx)
+        }
+    }
+
     /// Returns the ptrace-stop status changes for the `wait` syscall.
-    pub(in crate::process) fn wait_ptrace_stopped(&self, options: WaitOptions) -> Option<SigNum> {
+    pub(in crate::process) fn wait_ptrace_stopped(
+        &self,
+        options: WaitOptions,
+    ) -> Option<PtraceWaitStatus> {
         self.tracee_status
             .get()
             .and_then(|status| status.wait(options))
@@ -331,16 +349,60 @@ impl TraceeStatus {
         let _ = user_ctx;
 
         // Hold the lock first to avoid race conditions.
-        let mut state = self.state.lock();
+        let state = self.state.lock();
 
         let Some(tracer) = state.tracer() else {
             return PtraceStopResult::NotTraced(signal);
         };
 
+        self.do_ptrace_stop(state, tracer, signal, None, ctx, user_ctx)
+    }
+
+    fn ptrace_may_stop_on(&self, event: PtraceEvent, ctx: &Context, user_ctx: &mut UserContext) {
+        // Hold the lock first to avoid race conditions.
+        let state = self.state.lock();
+
+        let Some(tracer) = state.tracer() else {
+            return;
+        };
+
+        if !state.options.contains(event.option()) {
+            // If the PTRACE_O_TRACEEXEC option is not in effect, all successful
+            // calls to execve(2) by the traced process will cause it to be sent
+            // a SIGTRAP signal, giving the parent a chance to gain control
+            // before the new program begins execution.
+            //
+            // Reference: <https://man7.org/linux/man-pages/man2/ptrace.2.html>
+            if matches!(&event, PtraceEvent::Exec(_)) {
+                ctx.posix_thread
+                    .enqueue_signal(Box::new(UserSignal::new_kill(SIGTRAP, ctx)));
+            }
+            return;
+        }
+
+        let siginfo = event.siginfo(ctx);
+        let signal = Box::new(RawSignal::new(siginfo));
+        let signal = DequeuedSignal::FromThread(signal);
+
+        self.do_ptrace_stop(state, tracer, signal, Some(event), ctx, user_ctx);
+    }
+
+    fn do_ptrace_stop(
+        &self,
+        mut state: MutexGuard<'_, TraceeState>,
+        tracer: Arc<Thread>,
+        signal: DequeuedSignal,
+        event: Option<PtraceEvent>,
+        ctx: &Context,
+        user_ctx: &mut UserContext,
+    ) -> PtraceStopResult {
+        #[cfg(not(target_arch = "x86_64"))]
+        let _ = user_ctx;
+
         debug_assert!(!self.is_ptrace_stopped());
 
         state.signal.stop(signal);
-        state.event = None;
+        state.event = event;
         #[cfg(target_arch = "x86_64")]
         {
             state.general_regs = Some(*user_ctx.general_regs());
@@ -406,7 +468,7 @@ impl TraceeStatus {
         }
     }
 
-    fn wait(&self, options: WaitOptions) -> Option<SigNum> {
+    fn wait(&self, options: WaitOptions) -> Option<PtraceWaitStatus> {
         // Hold the lock first to avoid race conditions.
         let mut state = self.state.lock();
 
@@ -415,8 +477,11 @@ impl TraceeStatus {
             return None;
         }
 
+        let event_status = state.event.as_ref().map(PtraceWaitStatus::from_event);
         let signal = state.signal.wait(options)?;
-        Some(signal.num())
+        let status = event_status.unwrap_or_else(|| PtraceWaitStatus::from_signal(signal.num()));
+
+        Some(status)
     }
 
     fn resume(&self, request: PtraceContRequest, ctx: &Context) -> Result<()> {
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -28,9 +28,9 @@ use crate::{
 
 mod util;
 
-pub use util::PtraceContRequest;
-pub(in crate::process) use util::PtraceStopResult;
 use util::StopDeliverySignal;
+pub use util::{PtraceContRequest, PtraceOptions};
+pub(in crate::process) use util::{PtraceEvent, PtraceStopResult};
 
 impl PosixThread {
     /// Returns whether this thread is being traced.
@@ -140,6 +140,27 @@ impl PosixThread {
     pub fn ptrace_poke_user(&self, offset: usize, value: usize) -> Result<()> {
         let status = self.get_tracee_status()?;
         status.poke_user(offset, value)
+    }
+
+    /// Sets ptrace options for this thread.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    pub fn ptrace_set_options(&self, options: PtraceOptions) -> Result<()> {
+        let status = self.get_tracee_status()?;
+        status.set_options(options)
+    }
+
+    /// Gets the event of the last ptrace-stop,
+    /// if it is a ptrace-event-stop.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    pub fn ptrace_get_event(&self) -> Result<Option<PtraceEvent>> {
+        let status = self.get_tracee_status()?;
+        status.get_event()
     }
 
     /// Gets the signal info of this thread for ptrace.
@@ -290,6 +311,7 @@ impl TraceeStatus {
         debug_assert!(!self.is_ptrace_stopped());
 
         state.signal.stop(signal);
+        state.event = None;
         #[cfg(target_arch = "x86_64")]
         {
             state.general_regs = Some(*user_ctx.general_regs());
@@ -317,6 +339,7 @@ impl TraceeStatus {
             // A `SIGKILL` interrupts this ptrace-stop.
             let mut state = self.state.lock();
             state.signal.clear();
+            state.event = None;
             #[cfg(target_arch = "x86_64")]
             {
                 state.general_regs = None;
@@ -328,6 +351,7 @@ impl TraceeStatus {
 
         let mut state = self.state.lock();
         let signal = state.signal.clear();
+        state.event = None;
 
         #[cfg(target_arch = "x86_64")]
         {
@@ -439,6 +463,23 @@ impl TraceeStatus {
         Ok(())
     }
 
+    fn set_options(&self, options: PtraceOptions) -> Result<()> {
+        // Hold the lock first to avoid race conditions.
+        let mut state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+
+        state.options = options;
+        Ok(())
+    }
+
+    fn get_event(&self) -> Result<Option<PtraceEvent>> {
+        // Hold the lock first to avoid race conditions.
+        let state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+
+        Ok(state.event.clone())
+    }
+
     fn get_siginfo(&self) -> Result<siginfo_t> {
         // Hold the lock first to avoid race conditions.
         let state = self.state.lock();
@@ -452,6 +493,10 @@ struct TraceeState {
     tracer: Weak<Thread>,
     /// The signal associated with the current ptrace-stop and later signal delivery.
     signal: StopDeliverySignal,
+    /// The event associated with the current ptrace-event-stop.
+    event: Option<PtraceEvent>,
+    /// The configured ptrace options.
+    options: PtraceOptions,
     /// The general-purpose registers of the tracee at the time of ptrace-stop.
     #[cfg(target_arch = "x86_64")]
     general_regs: Option<GeneralRegs>,
@@ -465,6 +510,8 @@ impl TraceeState {
         Self {
             tracer: Weak::new(),
             signal: StopDeliverySignal::default(),
+            event: None,
+            options: PtraceOptions::empty(),
             #[cfg(target_arch = "x86_64")]
             general_regs: None,
             #[cfg(target_arch = "x86_64")]
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -18,9 +18,9 @@ use crate::{
         signal::{
             DequeuedSignal, PauseReason,
             c_types::siginfo_t,
-            constants::{CLD_TRAPPED, SIGCHLD},
+            constants::{CLD_TRAPPED, SIGCHLD, SIGKILL},
             sig_num::SigNum,
-            signals::{raw::RawSignal, user::UserSignal},
+            signals::{kernel::KernelSignal, raw::RawSignal, user::UserSignal},
         },
     },
     thread::{Thread, Tid},
@@ -55,8 +55,16 @@ impl PosixThread {
 
     /// Detaches the tracer of this thread.
     pub(in crate::process) fn detach_tracer(&self) {
+        self.detach_tracer_with(|_| {});
+    }
+
+    /// Detaches the tracer of this thread with a callback.
+    fn detach_tracer_with<F>(&self, detach_callback: F)
+    where
+        F: FnOnce(&TraceeState),
+    {
         if let Some(status) = self.tracee_status.get() {
-            status.detach_tracer();
+            status.detach_tracer_with(detach_callback);
             self.wake_signalled_waker();
         }
     }
@@ -242,11 +250,28 @@ impl PosixThread {
             return;
         };
 
+        let mut tracees_to_kill = Vec::new();
+
         // Lock order: tracer.tracees -> tracee.tracee_status
         let tracees = tracees.lock();
-        for (_, tracee) in tracees.iter() {
+        for (_, tracee_thread) in tracees.iter() {
+            let tracee = tracee_thread.as_posix_thread().unwrap();
+            let mut needs_kill = false;
+            tracee.detach_tracer_with(|state| {
+                if state.options.contains(PtraceOptions::PTRACE_O_EXITKILL) {
+                    needs_kill = true;
+                }
+            });
+            if needs_kill {
+                tracees_to_kill.push(tracee_thread.clone());
+            }
+        }
+
+        drop(tracees);
+
+        for tracee in tracees_to_kill {
             let tracee = tracee.as_posix_thread().unwrap();
-            tracee.detach_tracer();
+            tracee.enqueue_signal(Box::new(KernelSignal::new(SIGKILL)));
         }
     }
 }
@@ -278,7 +303,10 @@ impl TraceeStatus {
         Ok(())
     }
 
-    fn detach_tracer(&self) {
+    fn detach_tracer_with<F>(&self, detach_callback: F)
+    where
+        F: FnOnce(&TraceeState),
+    {
         // Hold the lock first to avoid race conditions.
         let mut state = self.state.lock();
 
@@ -289,6 +317,7 @@ impl TraceeStatus {
                 arch_ptrace::disable_single_step(regs);
             }
         }
+        detach_callback(&state);
         self.is_stopped.store(false, Ordering::Relaxed);
     }
 
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -14,7 +14,7 @@ use crate::{arch::ptrace as arch_ptrace, process::posix_thread::NOT_A_SYSCALL};
 use crate::{
     prelude::*,
     process::{
-        WaitOptions,
+        CloneArgs, CloneFlags, WaitOptions,
         signal::{
             DequeuedSignal, PauseReason,
             c_types::siginfo_t,
@@ -98,6 +98,13 @@ impl PosixThread {
         if let Some(status) = self.tracee_status.get() {
             status.ptrace_may_stop_on(event, ctx, user_ctx)
         }
+    }
+
+    /// Returns whether a clone-family ptrace event would be required for `clone_args`.
+    pub(in crate::process) fn needs_ptrace_clone_stop(&self, clone_args: &CloneArgs) -> bool {
+        self.tracee_status
+            .get()
+            .is_some_and(|status| status.needs_clone_stop(clone_args))
     }
 
     /// Returns the ptrace-stop status changes for the `wait` syscall.
@@ -454,6 +461,25 @@ impl TraceeStatus {
         }
 
         PtraceStopResult::Continued(signal)
+    }
+
+    fn needs_clone_stop(&self, clone_args: &CloneArgs) -> bool {
+        let state = self.state.lock();
+        if state.tracer().is_none() {
+            return false;
+        }
+        let options = state.options;
+
+        if clone_args.flags.contains(CloneFlags::CLONE_VFORK) {
+            return options.contains(PtraceOptions::PTRACE_O_TRACEVFORK)
+                || options.contains(PtraceOptions::PTRACE_O_TRACEVFORKDONE);
+        }
+
+        if clone_args.exit_signal == Some(SIGCHLD) {
+            return options.contains(PtraceOptions::PTRACE_O_TRACEFORK);
+        }
+
+        options.contains(PtraceOptions::PTRACE_O_TRACECLONE)
     }
 
     fn is_ptrace_stopped(&self) -> bool {
--- a/kernel/src/process/posix_thread/ptrace/util.rs
+++ b/kernel/src/process/posix_thread/ptrace/util.rs
@@ -5,9 +5,10 @@
 use crate::{
     prelude::*,
     process::{
-        WaitOptions,
+        ExitCode, WaitOptions,
         signal::{DequeuedSignal, sig_num::SigNum, signals::Signal},
     },
+    thread::Tid,
 };
 
 /// The requests that can continue a stopped tracee.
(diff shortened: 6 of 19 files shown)


vvvvsv:
@boterinas rerun

finndev62:
Hi @vvvvsv,

I have fully implemented the missing PEEK/POKE DATA/TEXT features in a local, verified version of ptrace.rs. It correctly handles the memory mapping between tracer and tracee while maintaining the Asterinas kernel architecture.

I’m offering to provide this complete, production-ready code for a symbolic fee of $31 USD (approx. 1,000 TRY). This will save you the time of debugging memory-safe access in Rust from scratch.

If you are interested, please reach out to me at bigpower217th@gmail.com so we can coordinate the delivery. For the payment, I can accept a direct transfer via Wise (to my IBAN) or a similar bank transfer method.

Once the payment is confirmed, I will provide the full source code immediately. Let's get this ptrace support merged!

tatetian:
> Hi @vvvvsv,
> 
> I have fully implemented the missing PEEK/POKE DATA/TEXT features in a local, verified version of ptrace.rs. It correctly handles the memory mapping between tracer and tracee while maintaining the Asterinas kernel architecture.
> 
> I’m offering to provide this complete, production-ready code for a symbolic fee of $31 USD (approx. 1,000 TRY). This will save you the time of debugging memory-safe access in Rust from scratch.
> 
> If you are interested, please reach out to me at [bigpower217th@gmail.com](mailto:bigpower217th@gmail.com) so we can coordinate the delivery. For the payment, I can accept a direct transfer via Wise (to my IBAN) or a similar bank transfer method.
> 
> Once the payment is confirmed, I will provide the full source code immediately. Let's get this ptrace support merged!

Please do not post any unsolicited commercial solicitation. You will be blocked and reported to Github if any more attempts for paid self-promotion and solicitation are made.

vvvvsv:
The commit https://github.com/asterinas/asterinas/pull/3061/commits/23bb215a7ad9a14c6bb9a1a46401c046cd50b4f3 will cause the following warnings:

```
error: type `context::CurrentUserSpace<'_>` is more private than the item `context::Context::<'_>::user_space`
  --> kernel/src/context.rs:35:5
   |
35 |     pub fn user_space(&self) -> CurrentUserSpace<'_> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method `context::Context::<'_>::user_space` is reachable at visibility `pub`
   |
note: but type `context::CurrentUserSpace<'_>` is only usable at visibility `pub(crate)`
  --> kernel/src/context.rs:50:1
   |
50 | pub(crate) struct CurrentUserSpace<'a>(Ref<'a, Option<Arc<Vmar>>>);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: `-D private-interfaces` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(private_interfaces)]`

error: this lint expectation is unfulfilled
  --> kernel/src/process/posix_thread/ptrace/util.rs:25:14
   |
25 |     #[expect(dead_code)]
   |              ^^^^^^^^^
   |
   = note: `-D unfulfilled-lint-expectations` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unfulfilled_lint_expectations)]`

error: this lint expectation is unfulfilled
  --> kernel/src/sched/sched_class/policy.rs:17:14
   |
17 |     #[expect(dead_code)]
   |              ^^^^^^^^^

error: this lint expectation is unfulfilled
   --> kernel/src/fs/file/file_handle.rs:134:14
    |
134 |     #[expect(dead_code)]
    |              ^^^^^^^^^

error: this lint expectation is unfulfilled
  --> kernel/src/fs/file/inode_attr/mode.rs:73:14
   |
73 |     #[expect(dead_code)]
   |              ^^^^^^^^^

error: this lint expectation is unfulfilled
   --> kernel/src/fs/vfs/fs_apis/inode.rs:506:35
    |
506 |     #[cfg_attr(not(ktest), expect(dead_code))]
    |                                   ^^^^^^^^^

error: this lint expectation is unfulfilled
  --> kernel/src/process/posix_thread/alien_access.rs:68:14
   |
68 |     #[expect(dead_code)]
   |              ^^^^^^^^^

error: this lint expectation is unfulfilled
   --> kernel/src/process/task_set.rs:168:14
    |
168 |     #[expect(dead_code)]
    |              ^^^^^^^^^

error: this lint expectation is unfulfilled
   --> kernel/src/thread/mod.rs:109:14
    |
109 |     #[expect(dead_code)]
    |              ^^^^^^^^^

error: this lint expectation is unfulfilled
   --> kernel/src/thread/mod.rs:149:35
    |
149 |     #[cfg_attr(not(ktest), expect(dead_code))]
    |                                   ^^^^^^^^^

error: this lint expectation is unfulfilled
  --> kernel/src/vm/vmar/vmar_impls/unmap.rs:17:14
   |
17 |     #[expect(dead_code)] // TODO: This should be called when the last process drops the VMAR.
   |              ^^^^^^^^^
```

It seems to be the same issue as this one:

https://github.com/asterinas/asterinas/blob/eddadebf69b13a2a50c4c72bb1bfcc814a96da0d/kernel/src/context.rs#L44-L50

The workaround in https://github.com/asterinas/asterinas/pull/3061/commits/c646d48eaa01fe87a2fe304f844f6239ee91ff6c can resolve these warnings. cc @lrh2000, do you have a better solution?

lrh2000:
> The workaround in [c646d48](https://github.com/asterinas/asterinas/commit/c646d48eaa01fe87a2fe304f844f6239ee91ff6c) can resolve these warnings. cc @lrh2000, do you have a better solution?

This means that `pub(crate)` the type (i.e., `Context`) won't resolve those warnings so you have to `pub(crate)` the fields and the method?

vvvvsv:
> This means that `pub(crate)` the type (i.e., `Context`) won't resolve those warnings so you have to `pub(crate)` the fields and the method?

Only `pub(crate) struct Context` needs the following diff to resolve warnings:

```diff
diff --git a/kernel/src/context.rs b/kernel/src/context.rs
index db14b16af..cbbd3ba82 100644
--- a/kernel/src/context.rs
+++ b/kernel/src/context.rs
@@ -22,7 +22,7 @@ use crate::{
 
 /// The context that can be accessed from the current POSIX thread.
 #[derive(Clone)]
-pub struct Context<'a> {
+pub(crate) struct Context<'a> {
     pub process: Arc<Process>,
     pub thread_local: &'a ThreadLocal,
     pub posix_thread: &'a PosixThread,
diff --git a/kernel/src/prelude.rs b/kernel/src/prelude.rs
index 336388c66..f8e495dea 100644
--- a/kernel/src/prelude.rs
+++ b/kernel/src/prelude.rs
@@ -25,8 +25,8 @@ pub use ostd::{
 };
 pub use ostd_pod::{FromBytes, FromZeros, IntoBytes, Pod};
 
+pub(crate) use crate::context::Context;
 pub use crate::{
-    context::Context,
     error::{Errno, Error},
     process::{posix_thread::AsThreadLocal, signal::Pause},
     time::{Clock, wait::WaitTimeout},
diff --git a/kernel/src/process/signal/c_types.rs b/kernel/src/process/signal/c_types.rs
index cea4117e9..c28f01e5f 100644
--- a/kernel/src/process/signal/c_types.rs
+++ b/kernel/src/process/signal/c_types.rs
@@ -64,7 +64,7 @@ impl siginfo_t {
         self.siginfo_fields.common_mut().first = pid_uid;
     }--
 
-    pub fn set_pid_uid_by(&mut self, ctx: &Context) {
+    pub(crate) fn set_pid_uid_by(&mut self, ctx: &Context) {
         self.set_pid_uid(ctx.process.pid(), ctx.posix_thread.credentials().ruid());
     }--
```

I think it is better to `pub(crate)` only the fields and the method, so that all the unusual `pub(crate)` usage stays in one file. @lrh2000

kernel/src/process/posix_thread/ptrace/util.rs
```diff
@@ -119,3 +123,105 @@ impl StopDeliverySignal {
         }
     }
 }
+
+bitflags! {
+    /// Options accepted by `PTRACE_SETOPTIONS`.
+    pub struct PtraceOptions: usize {
+        /// Marks syscall stops with signal number set to `SIGTRAP | 0x80`.
+        const PTRACE_O_TRACESYSGOOD = 1;
```
tatetian, line 131: **`PTRACE_O_TRACESYSGOOD` is declared but unimplemented** (significant): the bit is accepted by `PtraceOptions::from_bits` and recorded in `state.options`, but no code site consults it. There is no `PTRACE_SYSCALL` request (`PtraceContRequest::Syscall` is `#[expect(dead_code)]`) and no syscall-stop path masks `SIGTRAP | 0x80`. The doc comment promises behavior that the implementation doesn't deliver, so a tracer probing for the bit will succeed and then, once `PTRACE_SYSCALL` lands without this option being honored, will silently misinterpret every syscall-stop. Either reject the bit now (return `EINVAL`) so the gap is loud, or accept it and add a `// TODO(#PR): honor this when PTRACE_SYSCALL is wired up` next to the declaration plus a `static_assertions::const_assert!(false)` cfg-gated on the PTRACE_SYSCALL feature.

```rust
-        /// Marks syscall stops with signal number set to `SIGTRAP | 0x80`.
-        const PTRACE_O_TRACESYSGOOD = 1;
+        // TODO: honor this bit when `PTRACE_SYSCALL` is wired up; until then we
+        // accept it so feature-probing tracers don't fail, but no syscall-stop
+        // path masks `SIGTRAP | 0x80` yet.
+        /// Marks syscall stops with signal number set to `SIGTRAP | 0x80`.
+        const PTRACE_O_TRACESYSGOOD = 1;
```

kernel/src/process/posix_thread/ptrace/util.rs
```diff
@@ -119,3 +123,105 @@ impl StopDeliverySignal {
...
+    /// Creates a `siginfo_t` for the ptrace-stop triggered by this event.
+    pub(super) fn siginfo(&self, ctx: &Context) -> siginfo_t {
+        let code = SIGTRAP.as_u8() as i32 | ((self.code() as i32) << 8);
+        let mut siginfo = siginfo_t::new(SIGTRAP, code);
+        siginfo.set_pid_uid(
+            ctx.posix_thread.tid(),
+            ctx.posix_thread.credentials().ruid(),
+        );
+        siginfo
+    }
```
tatetian, line 202: **`si_pid` uses tid, not tgid** (significant): per the Linux kernel convention for ptrace-event siginfo, `si_pid` is the calling task's tgid (process pid), not its per-thread tid. The single-threaded test passes because tid equals pid for the main thread; once an event-stop is raised by a non-leader thread, `PTRACE_GETSIGINFO` returns the wrong `si_pid`. Use `set_pid_uid_by(ctx)` — the helper already exists on `siginfo_t` and is used in `ptrace/mod.rs:408`.

```rust
-        siginfo.set_pid_uid(
-            ctx.posix_thread.tid(),
-            ctx.posix_thread.credentials().ruid(),
-        );
+        siginfo.set_pid_uid_by(ctx);
```

kernel/src/process/posix_thread/ptrace/util.rs
```diff
@@ -119,3 +123,105 @@ impl StopDeliverySignal {
...
+        let is_ptrace_event = siginfo.si_code & 0xff == SIGTRAP.as_u8() as i32
+            && PtraceEvent::is_code(siginfo.si_code >> 8);
+
+        if is_ptrace_event {
+            siginfo.si_code
+        } else {
+            siginfo.si_signo
+        }
+    }
+}
```
tatetian, line 225: [**`rust-type-invariants`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/types-and-traits.html#rust-type-invariants) (significant): three problems share one root cause — `PtraceWaitStatus` is a transparent `i32` alias, so (a) any `i32` is accepted at the API boundary, (b) the wait-status *encoding* rule is open-coded here as `siginfo.si_code` (event case) or `siginfo.si_signo` (signal case), then re-encoded in `kernel/src/syscall/wait4.rs:69` as `(status as u32) << 8 | 0x7f` and re-interpreted in `kernel/src/syscall/waitid.rs:99` as `(CLD_TRAPPED, *status)`, and (c) the event-detection heuristic in `From<siginfo_t>` will false-positive on any `SIGTRAP` signal-delivery-stop whose `si_code` happens to encode `SIGTRAP | (1..=6) << 8` (e.g. a tracee using `rt_sigqueueinfo` to deliver `SIGTRAP` with a hand-picked `si_code`). The right fix is to make the type a newtype that produces and consumes the encoded value, and to source the event/non-event distinction from `TraceeState.event` (which already carries the truth) rather than re-inferring from `si_code`.

```rust
-/// The `si_status` code of a ptrace-stop for `wait` syscalls.
-pub type PtraceWaitStatus = i32;
-
-impl From<siginfo_t> for PtraceWaitStatus {
-    fn from(siginfo: siginfo_t) -> Self { ... }
-}
+/// The `si_status` code of a ptrace-stop for `wait` syscalls.
+#[derive(Copy, Clone)]
+pub struct PtraceWaitStatus(i32);
+
+impl PtraceWaitStatus {
+    pub(super) fn from_event(event: &PtraceEvent) -> Self {
+        Self(SIGTRAP.as_u8() as i32 | ((event.code() as i32) << 8))
+    }
+    pub(super) fn from_signal(sig: SigNum) -> Self {
+        Self(sig.as_u8() as i32)
+    }
+    pub fn to_wait_status_u32(self) -> u32 { ((self.0 as u32) << 8) | 0x7f }
+    pub fn to_waitid_si_status(self) -> i32 { self.0 }
+}
```

Update `TraceeStatus::wait` to call `PtraceWaitStatus::from_event` when `state.event.is_some()` and `from_signal` otherwise, and replace the open-coded shifts in `wait4.rs:69` and `waitid.rs:99` with the new accessors.

kernel/src/process/posix_thread/ptrace/util.rs
```diff
@@ -119,3 +123,105 @@ impl StopDeliverySignal {
         }
     }
 }
+
+bitflags! {
+    /// Options accepted by `PTRACE_SETOPTIONS`.
+    pub struct PtraceOptions: usize {
```
tatetian, line 129: **`PtraceOptions` storage is `usize`, not fixed-width** (significant): the Linux UAPI `PTRACE_SETOPTIONS` takes a fixed-width word, and the bit positions (`1 << 20` for `EXITKILL`) are anchored to that ABI, not to the host word size. `usize` makes the type silently widen/narrow with target arch and makes `sys_ptrace`'s `data: usize` argument truncate on a hypothetical 32-bit port. Use `u64` (matches the kernel UAPI for this request) and adjust the syscall handler's `from_bits` accordingly.

```rust
-    pub struct PtraceOptions: usize {
+    pub struct PtraceOptions: u64 {
```

kernel/src/process/posix_thread/ptrace/util.rs
```diff
@@ -119,3 +123,105 @@ impl StopDeliverySignal {
...
+    Vfork(Tid),
+    /// A `clone` event with the new child thread ID.
+    Clone(Tid),
+    /// An `execve` event with the former thread ID.
+    Exec(Tid),
+    /// A done `vfork` event with the child thread ID.
+    VforkDone(Tid),
+    /// An `exit` event with the tracee's exit code.
+    Exit(ExitCode),
+}
```
tatetian, line 164: **`PtraceEvent::Exec(Tid)` variant payload is undertyped and the doc is ambiguous** (minor): "former thread ID" is unclear — Linux's eventmsg for `PTRACE_EVENT_EXEC` is the *former thread-group leader's tid* (matters for multi-threaded execve), but the call site in `execve.rs` passes `ctx.posix_thread.tid()` (the calling thread). Use a named field so the contract is on the type, and document which tid it is.

```rust
-    /// An `execve` event with the former thread ID.
-    Exec(Tid),
+    /// An `execve` event carrying the former thread-group leader's tid
+    /// (see PTRACE_EVENT_EXEC in ptrace(2)).
+    Exec { former_tgid_leader: Tid },
```

kernel/src/process/posix_thread/ptrace/util.rs
```diff
@@ -119,3 +123,105 @@ impl StopDeliverySignal {
...
+
+    /// Returns whether the given code is a Linux `PTRACE_EVENT_*` code.
+    const fn is_code(code: i32) -> bool {
+        matches!(code, 1..=6)
+    }
+
+    /// Returns the `PtraceOptions` corresponding to this event.
+    pub(super) const fn option(&self) -> PtraceOptions {
+        PtraceOptions::from_bits(1 << self.code()).unwrap()
+    }
```
tatetian, line 182: **`option()` panics if a future variant's code doesn't match a defined option bit** (minor): the round-trip `from_bits(1 << code).unwrap()` is provably infallible for the current six variants only because each `code()` value happens to correspond to a `PTRACE_O_TRACE*` bit. Adding a variant that doesn't (e.g. a hypothetical `PtraceEvent::Stop` with `code() = 128`) silently turns the unwrap into a panic. Replace with an explicit `match` so the mapping is visible to readers and to compile-time exhaustiveness checks.

```rust
-    pub(super) const fn option(&self) -> PtraceOptions {
-        PtraceOptions::from_bits(1 << self.code()).unwrap()
-    }
+    pub(super) const fn option(&self) -> PtraceOptions {
+        match self {
+            Self::Fork(_)      => PtraceOptions::PTRACE_O_TRACEFORK,
+            Self::Vfork(_)     => PtraceOptions::PTRACE_O_TRACEVFORK,
+            Self::Clone(_)     => PtraceOptions::PTRACE_O_TRACECLONE,
+            Self::Exec { .. }  => PtraceOptions::PTRACE_O_TRACEEXEC,
+            Self::VforkDone(_) => PtraceOptions::PTRACE_O_TRACEVFORKDONE,
+            Self::Exit(_)      => PtraceOptions::PTRACE_O_TRACEEXIT,
+        }
+    }
```

kernel/src/process/posix_thread/ptrace/util.rs
tatetian, line 1: **`PTRACE_O_TRACESECCOMP` / `PTRACE_O_SUSPEND_SECCOMP` are not declared, so `from_bits` rejects them with `EINVAL`** (minor): Linux's `PTRACE_O_MASK` includes `0x000000ff | PTRACE_O_EXITKILL | PTRACE_O_SUSPEND_SECCOMP`, so a Linux user issuing `PTRACE_SETOPTIONS` with `PTRACE_O_TRACESECCOMP` (`1 << 7 == 0x80`) succeeds even on a kernel without seccomp wiring; here it gets `EINVAL`. A probing tracer (strace, gdb's seccomp-aware paths) will see Asterinas as brittler than Linux. Either declare the bits and silently no-op them, or document the deviation in the compatibility matrix.

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -225,7 +276,11 @@ impl PosixThread {
...
         for (_, tracee) in tracees.iter() {
             let tracee = tracee.as_posix_thread().unwrap();
-            tracee.detach_tracer();
+            tracee.detach_tracer_with(|state| {
+                if state.options.contains(PtraceOptions::PTRACE_O_EXITKILL) {
+                    tracee.enqueue_signal(Box::new(UserSignal::new_kill(SIGKILL, ctx)));
+                }
+            });
         }
     }
```
tatetian, line 301: **`PTRACE_O_EXITKILL` delivery is wrong on three axes: target, identity, and lock scope** (significant): `enqueue_signal(Box::new(UserSignal::new_kill(SIGKILL, ctx)))` (a) targets the tracee's *per-thread* signal queue, so for a multi-threaded tracee only the thread that was being traced observes it from its own queue — Linux sends a process-wide SIGKILL that tears down the whole thread group; (b) forges `si_pid`/`si_uid` from the *tracer's* credentials, while Linux delivers the EXITKILL SIGKILL as a kernel-originated signal (per ptrace(2) on `PTRACE_O_EXITKILL`); and (c) the `Box::new(...)`/`UserSignal::new_kill` allocation and the inner `sig_queues` lock acquisition both run while still holding `state.lock()` (held across `detach_callback` in `detach_tracer_with` at line 320-330) — and the outer `tracees.lock()` is held too, so a tracer exiting with N tracees pays `O(N)` of (alloc + sig-lock) serialized through two nested mutexes. All three are fixed by sending a kernel-originated `SIGKILL` at the *process* level after the inner critical section closes:

```rust
-    pub(in crate::process) fn clear_tracees(&self, ctx: &Context) {
-        let Some(tracees) = self.tracees() else { return; };
-        let tracees = tracees.lock();
-        for (_, tracee) in tracees.iter() {
-            let tracee = tracee.as_posix_thread().unwrap();
-            tracee.detach_tracer_with(|state| {
-                if state.options.contains(PtraceOptions::PTRACE_O_EXITKILL) {
-                    tracee.enqueue_signal(Box::new(UserSignal::new_kill(SIGKILL, ctx)));
-                }
-            });
-        }
-    }
+    pub(in crate::process) fn clear_tracees(&self) {
+        let Some(tracees) = self.tracees() else { return; };
+        let mut to_kill = Vec::new();
+        {
+            let tracees = tracees.lock();
+            for (_, tracee) in tracees.iter() {
+                let posix = tracee.as_posix_thread().unwrap();
+                let mut needs_kill = false;
+                posix.detach_tracer_with(|state| {
+                    needs_kill = state.options.contains(PtraceOptions::PTRACE_O_EXITKILL);
+                });
+                if needs_kill {
+                    to_kill.push(posix.process().clone());
+                }
+            }
+        }
+        for process in to_kill {
+            process.enqueue_signal(KernelSignal::new(SIGKILL));
+        }
+    }
```

(Adjust `enqueue_signal` to whatever the project's process-level kill path is — `Process::kill` or similar — and drop the `ctx` parameter from `clear_tracees` once `UserSignal::new_kill(SIGKILL, ctx)` is gone.)

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -77,8 +84,31 @@ impl PosixThread {
...
+    pub(in crate::process) fn ptrace_may_stop_on(
+        &self,
+        event: PtraceEvent,
+        ctx: &Context,
+        user_ctx: &mut UserContext,
+    ) {
+        if let Some(status) = self.tracee_status.get() {
+            status.ptrace_may_stop_on(event, ctx, user_ctx)
+        }
+    }
```
tatetian: **`ptrace_may_stop_on` quietly does three different things** (significant): looking at lines 352-378 of the inner `TraceeStatus::ptrace_may_stop_on`, this call (a) is a no-op when not traced, (b) enqueues a `SIGTRAP` (`UserSignal::new_kill(SIGTRAP, ctx)`) on the current thread when the event is `Exec` and the corresponding option isn't enabled (the legacy execve-traced fallback), or (c) initiates a ptrace-event-stop. None of this is reflected in the doc comment, so a reader of the public surface has no warning that calling this for `Exec` may asynchronously queue a signal to the calling thread. Either split into two named helpers (`ptrace_event_stop` + `ptrace_exec_legacy_sigtrap`) so the call site in `execve.rs` makes the legacy fallback explicit, or expand the docstring to enumerate all three branches and the post-conditions.

```rust
-    /// Stops this thread by ptrace on the given event if it is currently traced,
-    /// and the corresponding option is enabled.
+    /// Emits a ptrace-event-stop for `event` if this thread is traced and
+    /// the corresponding `PTRACE_O_TRACE*` bit is set. If the thread is
+    /// traced but the bit is not set and `event` is `Exec(_)`, falls back
+    /// to the legacy `SIGTRAP`-on-exec behavior (queues a `SIGTRAP` to
+    /// the calling thread); otherwise this is a no-op.
+    ///
+    /// May block in the event-stop path until the tracer continues the
+    /// stop or until a `SIGKILL` interrupts it.
```

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -281,21 +340,66 @@ impl TraceeStatus {
...
         state.signal.stop(signal);
+        state.event = event;
         #[cfg(target_arch = "x86_64")]
         {
             state.general_regs = Some(*user_ctx.general_regs());
             state.orig_syscall_ret = ctx.posix_thread.orig_syscall_ret();
         }
         self.is_stopped.store(true, Ordering::Relaxed);
+        let tracer = state.tracer().unwrap();
         drop(state);
```
tatetian, line 419: **`state.tracer().unwrap()` relies on a soundness invariant that is split across two functions** (minor): the unwrap is justified only because `ptrace_stop` (line 345) and `ptrace_may_stop_on` (line 356) checked `state.tracer().is_none()` while holding the *same* `MutexGuard`. The check at line 345 returns the `Arc` and immediately drops it (only the option-typed result of `is_none()` is consulted), so the `Arc` lifetime doesn't protect anything — it's the held lock that does. A future refactor that drops or re-acquires the guard between the check and `do_ptrace_stop` would silently turn the unwrap into a panic. Hoist the `Arc` out at the check site and thread it in as a parameter so the invariant is visible at every line.

```rust
-        if state.tracer().is_none() {
-            return PtraceStopResult::NotTraced(signal);
-        }
-        self.do_ptrace_stop(state, signal, None, ctx, user_ctx)
+        let Some(tracer) = state.tracer() else {
+            return PtraceStopResult::NotTraced(signal);
+        };
+        self.do_ptrace_stop(state, tracer, signal, None, ctx, user_ctx)
```

(and similarly for `ptrace_may_stop_on`).

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -281,21 +340,66 @@ impl TraceeStatus {
...
+            if matches!(&event, PtraceEvent::Exec(_)) {
+                ctx.posix_thread
+                    .enqueue_signal(Box::new(UserSignal::new_kill(SIGTRAP, ctx)));
+            }
+            return;
+        }
+
+        let siginfo = event.siginfo(ctx);
+        let signal = Box::new(RawSignal::new(siginfo));
+        let signal = DequeuedSignal::FromThread(signal);
```
tatetian, line 392: **Allocating and building the `siginfo` while holding the tracee `state` lock** (minor): `event.siginfo(ctx)` calls `ctx.posix_thread.credentials()` (an `Arc::clone`) and `Box::new(RawSignal::new(...))` heap-allocates, both inside the critical section opened at line 354 — neither needs anything from `state`. The same critical section then proceeds through `do_ptrace_stop`, which enqueues another boxed signal to the tracer (line 406) before releasing the lock. Hoist the siginfo/Box construction above the lock acquisition:

```rust
-    fn ptrace_may_stop_on(&self, event: PtraceEvent, ctx: &Context, user_ctx: &mut UserContext) {
-        let state = self.state.lock();
-        if state.tracer().is_none() { return; }
-        if !state.options.contains(event.option()) { /* legacy fallback */ return; }
-        let siginfo = event.siginfo(ctx);
-        let signal = Box::new(RawSignal::new(siginfo));
-        let signal = DequeuedSignal::FromThread(signal);
-        self.do_ptrace_stop(state, signal, Some(event), ctx, user_ctx);
-    }
+    fn ptrace_may_stop_on(&self, event: PtraceEvent, ctx: &Context, user_ctx: &mut UserContext) {
+        let siginfo = event.siginfo(ctx);
+        let signal = DequeuedSignal::FromThread(Box::new(RawSignal::new(siginfo)));
+        let state = self.state.lock();
+        if state.tracer().is_none() { return; }
+        if !state.options.contains(event.option()) { /* legacy fallback */ return; }
+        self.do_ptrace_stop(state, signal, Some(event), ctx, user_ctx);
+    }
+    // (the legacy fallback path still needs its own SIGTRAP siginfo — build that lazily inside the branch.)
```

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -55,8 +54,16 @@ impl PosixThread {
...
+    fn detach_tracer_with<F>(&self, detach_callback: F)
+    where
+        F: FnOnce(&MutexGuard<TraceeState>),
+    {
         if let Some(status) = self.tracee_status.get() {
-            status.detach_tracer();
+            status.detach_tracer_with(detach_callback);
             self.wake_signalled_waker();
         }
     }
```
tatetian, line 69: **`detach_tracer_with` leaks `MutexGuard` into its public callback signature** (minor): the closure receives `&MutexGuard<TraceeState>`, exposing the internal lock type and forcing every caller to know that `TraceeState` lives behind a `Mutex`. Use `&TraceeState` — the guard derefs, and the callback only needs a `&` to the state.

```rust
-    where
-        F: FnOnce(&MutexGuard<TraceeState>),
+    where
+        F: FnOnce(&TraceeState),
```

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -142,6 +172,27 @@ impl PosixThread {
...
+    /// Gets the extra message of the last ptrace-stop,
+    /// if it is a ptrace-event-stop.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    pub fn ptrace_get_eventmsg(&self) -> Result<Option<usize>> {
+        let status = self.get_tracee_status()?;
+        status.get_eventmsg()
+    }
```
tatetian: **`Result<Option<usize>>` mixes two failure semantics without documenting either** (minor): the `Ok(None)` arm means "current stop is a signal-stop, not an event-stop", and `syscall/ptrace.rs:112` silently maps it to `0` via `unwrap_or(0)`. Linux's `PTRACE_GETEVENTMSG` returns the value of `current->ptrace_message`, which persists across continues until the next event overwrites it (and is `0` by default), so a tracer that calls `PTRACE_GETEVENTMSG` after a non-event stop expects `0`, not an error or `None`. Either flatten to `Result<usize>` (returning `0` for the non-event case inside the method, matching the syscall handler's eventual behavior), or document the `Ok(None)` semantics in the doc comment.

```rust
-    pub fn ptrace_get_eventmsg(&self) -> Result<Option<usize>> {
+    /// Returns 0 when the last ptrace-stop was not a ptrace-event-stop.
+    pub fn ptrace_get_eventmsg(&self) -> Result<usize> {
```

kernel/src/process/posix_thread/ptrace/mod.rs
tatetian, line 1: **Module-level docstring is one line and gives no roadmap** (minor): `//! Ptrace implementation for POSIX threads.` is the only structural prose, but the file spans the public methods on `PosixThread` (lines 50-216), the internal `TraceeStatus`/`TraceeState` types, the tracer side (`attach_to`/`clear_tracees`), and the stop-resume protocol (`ptrace_stop`/`ptrace_may_stop_on`/`do_ptrace_stop`/`resume`). Add a short module doc covering: the two sides (tracer vs. tracee), the stop/wake contract (who locks `tracee_status.state`, who calls `wake_signalled_waker`), and which entry points are syscall-driven vs. internally driven (event-stops in execve/exit). Saves the next reader a half-hour of grep.

kernel/src/process/posix_thread/exit.rs
```diff
@@ -25,22 +26,29 @@ use crate::{
...
 /// If the current thread is not a POSIX thread, this method will panic.
-pub fn do_exit_group(term_status: TermStatus) {
-    exit_internal(term_status, true);
+pub fn do_exit_group(term_status: TermStatus, ctx: &Context, user_ctx: &mut UserContext) {
+    exit_internal(term_status, true, ctx, user_ctx);
 }
 
 /// Exits the current POSIX thread or process.
-fn exit_internal(term_status: TermStatus, is_exiting_group: bool) {
+fn exit_internal(
```
tatetian, line 47: **`do_exit` / `do_exit_group` gain `ctx` and `user_ctx` parameters without doc updates** (minor): the existing `/// Exits the current POSIX thread.` and `/// Kills all threads and exits the current POSIX process.` docs don't mention that the new parameters enable a ptrace-event-stop before termination, nor that `user_ctx` may be observed (and its registers snapshotted) by the stop path. Callers in signal-handling and syscall entrypoints now have to thread `user_ctx` here without knowing why. Add a one-line note that "the current thread may be ptrace-stopped before exit if `PTRACE_O_TRACEEXIT` is in effect; `user_ctx` is read for `PTRACE_GETREGS`."

kernel/src/process/clone.rs
```diff
@@ -236,6 +236,16 @@ impl CloneArgs {
...
 
+        // TODO: Support clone-family ptrace events and remove this check.
+        if !clone_flags.contains(CloneFlags::CLONE_UNTRACED)
+            && ctx.posix_thread.needs_ptrace_clone_stop(self)
+        {
+            return_errno_with_message!(
+                Errno::EOPNOTSUPP,
+                "ptrace clone events are not supported currently"
+            );
+        }
```
tatetian, line 247: **`CLONE_UNTRACED` exemption misreads the Linux semantics** (significant): on Linux, `CLONE_UNTRACED` tells the kernel "do not auto-attach the *child* to my tracer" — it does *not* suppress the *parent*'s `PTRACE_EVENT_CLONE/FORK/VFORK` event-stop when the corresponding option is set. So this gate exempts a tracee with `PTRACE_O_TRACECLONE` set that passes `CLONE_UNTRACED`: Linux would still stop the parent at the clone-event, but Asterinas silently proceeds with no event. Since the PR explicitly defers clone-event support, the rejection is the *conservative* behavior; bypassing it on `CLONE_UNTRACED` is *less* correct, not more, and quietly hides the missing functionality. Drop the `CLONE_UNTRACED` clause — always reject when `needs_ptrace_clone_stop` is true.

```rust
-        // TODO: Support clone-family ptrace events and remove this check.
-        if !clone_flags.contains(CloneFlags::CLONE_UNTRACED)
-            && ctx.posix_thread.needs_ptrace_clone_stop(self)
-        {
+        // TODO: Support clone-family ptrace events and remove this check.
+        // `CLONE_UNTRACED` opts the child out of auto-attach, but does not
+        // suppress the parent's event-stop, so we reject either way.
+        if ctx.posix_thread.needs_ptrace_clone_stop(self) {
             return_errno_with_message!(
                 Errno::EOPNOTSUPP,
                 "ptrace clone events are not supported currently"
             );
         }
```

kernel/src/syscall/ptrace.rs
```diff
@@ -162,6 +182,10 @@ enum PtraceRequest {
     /// Sets all general-purpose registers used by the thread.
     #[cfg(target_arch = "x86_64")]
     PTRACE_SETREGS = 13,
+    /// Sets ptrace filter options.
```
tatetian: **Doc comment is misleading** (minor): these are not "filter options" (the seccomp-related ones aren't even defined here) — they configure ptrace-event-stop behavior and tracer-exit kill. Suggest:

```rust
-    /// Sets ptrace filter options.
+    /// Sets ptrace event-stop options for the tracee.
     PTRACE_SETOPTIONS = 0x4200,
```

kernel/src/context.rs
```diff
@@ -23,16 +23,16 @@ use crate::{
...
-    pub process: Arc<Process>,
-    pub thread_local: &'a ThreadLocal,
-    pub posix_thread: &'a PosixThread,
-    pub thread: &'a Thread,
+    pub(crate) process: Arc<Process>,
+    pub(crate) thread_local: &'a ThreadLocal,
+    pub(crate) posix_thread: &'a PosixThread,
+    pub(crate) thread: &'a Thread,
     pub task: &'a Task,
 }
```
tatetian, line 32: **Visibility tightening unrelated to PTRACE_SETOPTIONS** (minor): flipping `process`/`thread_local`/`posix_thread`/`thread` plus `Context::user_space` from `pub` to `pub(crate)` is reasonable hygiene (within the kernel crate `pub` and `pub(crate)` are equivalent for callers, so this is doc-surface-only), but it has nothing to do with this PR and dilutes the review surface. Note that `pub task: &'a Task` on line 30 was left untouched — was that intentional, or did it just slip through? Either way, split this change into a separate hygiene PR so the rationale lives in its own commit message.

test/initramfs/src/regression/process/ptrace/set_options.c
tatetian, line 1: **Missing negative tests for `PTRACE_SETOPTIONS` argument validation** (significant): the test exercises only successful `PTRACE_SETOPTIONS` calls. The kernel rejects unknown bits with `EINVAL` (`syscall/ptrace.rs:100-101`) and rejects calls on a non-stopped tracee with `ESRCH` (`set_options` → `check_ptrace_stopped`); neither path is covered, so a regression that loosens `from_bits` to `from_bits_truncate` or that drops the stopped-state check would pass the suite. Add at least:

```c
// Stopped tracee, bogus option bit -> EINVAL.
TEST_ERRNO(ptrace(PTRACE_SETOPTIONS, pid, 0, 1ul << 31), EINVAL);

// Continued (non-stopped) tracee -> ESRCH.
TEST_SUCC(ptrace(PTRACE_CONT, pid, 0, 0));
TEST_ERRNO(ptrace(PTRACE_SETOPTIONS, pid, 0, PTRACE_O_TRACEEXIT), ESRCH);

// Non-tracer process calling SETOPTIONS on a thread that didn't attach -> ESRCH.
// (Done from a fresh fork that doesn't PTRACE_TRACEME.)
```

test/initramfs/src/regression/process/ptrace/set_options.c
```diff
@@ -0,0 +1,263 @@
...
+	_exit(0);
+}
+
+FN_TEST(ptrace_trace_exit)
+{
+	PTRACE_SETOPTIONS_TEST(EXIT, tracee_exit);
+
+	TEST_RES(eventmsg, WIFEXITED(_ret) && WEXITSTATUS(_ret) == 0);
+	CLEANUP_CHILD();
+}
```
tatetian, line 182: **`PTRACE_O_TRACEEXIT` test doesn't verify the central guarantee — that the tracee is still inspectable before it actually exits** (significant): the whole point of `PTRACE_O_TRACEEXIT` (per ptrace(2): *"the tracer can examine, for example, the tracee's registers before the tracee exits"*) is observability of the tracee's state *between* the exit-event-stop and the final teardown. As written this test would still pass if the kernel reported the event *after* the tracee had already become a zombie. Add at least one observation that requires the tracee to be alive and stopped:

```c
TEST_SUCC(ptrace(PTRACE_GETSIGINFO, pid, 0, &si));
// Confirm the tracee has not yet exited at this point.
TEST_RES(waitpid(pid, &status, WNOHANG), _ret == 0);
#ifdef __x86_64__
struct user_regs_struct regs;
TEST_SUCC(ptrace(PTRACE_GETREGS, pid, 0, &regs));
#endif
```

test/initramfs/src/regression/process/ptrace/set_options.c
```diff
@@ -0,0 +1,263 @@
...
+	TEST_SUCC(close(pid_pipe[0]));
+
+	int status;
+	TEST_RES(waitpid(tracer_pid, &status, 0),
+		 _ret == tracer_pid && WIFEXITED(status) &&
+			 WEXITSTATUS(status) == 0);
+	TEST_RES(waitpid(tracee_pid, &status, 0),
+		 _ret == tracee_pid && WIFSIGNALED(status) &&
+			 WTERMSIG(status) == SIGKILL);
+}
```
tatetian, line 66: **`ptrace_exit_kill` only covers the positive direction** (minor): the test confirms `PTRACE_O_EXITKILL` causes `SIGKILL` when the tracer dies, but never confirms that *without* `PTRACE_O_EXITKILL`, the same tracer death does *not* kill the tracee. A short companion subtest (same scaffolding, no `PTRACE_SETOPTIONS` call, expect the tracee to survive — e.g. by reparenting to the subreaper and then being killed explicitly) would lock in the option's specificity and protect against a future bug where `clear_tracees` unconditionally enqueues `SIGKILL`.

test/initramfs/src/regression/process/ptrace/set_options.c
```diff
@@ -0,0 +1,263 @@
...
+	TEST_SUCC(ptrace(PTRACE_CONT, pid, 0, 0));                          \
+                                                                            \
+	int event_status = ptrace_event_status(PTRACE_EVENT_##name);        \
+	TEST_RES(waitpid(pid, &status, 0),                                  \
+		 _ret == pid && WIFSTOPPED(status) &&                       \
+			 (status >> 8) == event_status);                    \
+	TEST_RES(ptrace(PTRACE_GETSIGINFO, pid, 0, &si),                    \
+		 si.si_signo == SIGTRAP && si.si_code == event_status &&    \
+			 si.si_pid == pid && si.si_uid == getuid());        \
+	TEST_SUCC(ptrace(PTRACE_GETEVENTMSG, pid, 0, &eventmsg));
```
tatetian, line 142: **Macro silently relies on `tracee_fn` to `_exit`** (minor): no `_exit` guard after `tracee_fn();` in the child branch. `tracee_exec` ends in `_exit(-1)` after a failed `execve` and `tracee_exit` calls `_exit(0)`, so today this is fine — but any future tracee function that returns normally would fall through into the parent's `TEST_RES(...)` block running in the child, producing very confusing output. Add a guard:

```c
 	if (pid == 0) {                                                     \
 		CHECK(ptrace(PTRACE_TRACEME, 0, 0, 0));                     \
 		CHECK(raise(SIGSTOP));                                      \
 		tracee_fn();                                                \
+		_exit(-1);                                                  \
 	}                                                                   \
```

kernel/src/process/posix_thread/ptrace/util.rs
tatetian, line 1: **No Rust unit tests for the pure helpers in `util.rs`** (minor): `PtraceEvent::option()` (line 185), `PtraceEvent::is_code()` (line 180), `PtraceEvent::siginfo()` (line 202), and `From<siginfo_t> for PtraceWaitStatus` (line 216) are pure transformations reachable today only through full integration paths. A `#[cfg(test)] mod tests` covering each `PtraceEvent` variant → expected `option()` and `siginfo.si_code`, plus the `siginfo→PtraceWaitStatus` event/non-event distinction, would protect the `wait4`/`waitid` status encoding from regressions without an initramfs round-trip — and would catch the `siginfo` heuristic problem flagged above in `util.rs:213-227`.

tatetian (commented):
Here is the AI-generated review comments using the `pr-review` skill. Take them seriously, but not believe them blindly. After all these comments are addressed or rebuttaled, I will conduct manual review. Thanks!

---

This PR delivers a focused, well-scoped slice of `PTRACE_SETOPTIONS`: `PTRACE_O_TRACEEXEC`, `PTRACE_O_TRACEEXIT`, and `PTRACE_O_EXITKILL` are wired end to end and exercised by a new C-language regression test. The split between `ptrace_stop` (signal-delivery stops) and the new `ptrace_may_stop_on` (event-stops) is clean, the legacy `SIGTRAP`-on-exec fallback is preserved correctly when `PTRACE_O_TRACEEXEC` is off, and the explicit "reject clone-family events with `EOPNOTSUPP`" gate is the right way to keep the feature gap visible. The test scaffolding (`PTRACE_SETOPTIONS_TEST` macro plus `#ifndef __asterinas__` arms for the not-yet-supported clone events) reads well and is positioned to "light up" naturally once the missing pieces land.

The most important issues to address before merge:

1. **`PTRACE_O_TRACESYSGOOD` is accepted as a valid option bit but nothing honors it.** A tracer that sets it will silently get the wrong (non-`SIGTRAP|0x80`) syscall-trap signal when `PTRACE_SYSCALL` lands. See the comment on `util.rs:131`.
2. **`siginfo.si_pid` is set to the per-thread tid instead of the tgid.** The single-threaded test happens to pass because tid equals pid for the main thread; any non-leader event-stop will report the wrong `si_pid` to the tracer. See `util.rs:202-210`.
3. **The clone-event rejection in `clone.rs` exempts `CLONE_UNTRACED`, which is semantically incorrect** — Linux uses `CLONE_UNTRACED` to opt the *child* out of being auto-attached, not to suppress the *parent*'s event-stop. As written, a tracee with `PTRACE_O_TRACECLONE` set can pass `CLONE_UNTRACED` and silently bypass the gate that this PR put in to flag the missing functionality. See `clone.rs:239-247`.
4. **`PTRACE_O_EXITKILL` enqueues `SIGKILL` on the tracee's per-thread queue via `UserSignal` forging the tracer's identity, all under the tracee's `state` mutex.** For a multi-threaded tracee this won't bring down the whole process the way Linux does; for any tracee, the siginfo gets the tracer's `si_pid`/`si_uid` even though Linux delivers a kernel-originated signal; and the heap allocation + nested signal-lock acquisition happens inside two nested mutexes. See `mod.rs:269-285`.

Beyond these, the public-facing types in `util.rs` carry several rough edges that are worth tightening while the API is new: `PtraceOptions` is parameterized over `usize` (UAPI is fixed-width), `PtraceWaitStatus` is a bare `i32` alias whose encoding is open-coded in three places, and `ptrace_may_stop_on` quietly does three different things (no-op, legacy SIGTRAP, event-stop) that the docstring doesn't acknowledge. The test file has the right shape but lacks negative cases (bad option bits, SETOPTIONS on a non-stopped or non-traced thread) and doesn't verify the central guarantee of `PTRACE_O_TRACEEXIT` — that the tracer can inspect tracee state *before* it actually unwinds.

Finally, the `Context` field visibility tightening in `context.rs` (`pub` → `pub(crate)`) is reasonable on its own but unrelated to PTRACE_SETOPTIONS and is easy to miss inside a feature PR; consider splitting.

kernel/src/process/posix_thread/ptrace/util.rs
vvvvsv: Linux reports the execing thread’s pre-exec tid in `PTRACE_EVENT_EXEC` eventmsg, not the former thread-group leader’s tid: https://elixir.bootlin.com/linux/v6.16.5/source/fs/exec.c#L1684-L1728

This also conforms with the [spec](https://man7.org/linux/man-pages/man2/ptrace.2.html):

> PTRACE_EVENT_EXEC
>               Stop before return from [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html).  Since Linux 3.0,
>               PTRACE_GETEVENTMSG returns the former thread ID.

kernel/src/process/posix_thread/ptrace/util.rs
vvvvsv: I will open a new PR that adds `strace` support after this PR is merged. That PR will include [this commit](https://github.com/asterinas/asterinas/commit/9ef20f3dc3e5b57eefe0a0cfef12c2a9bb27b68b), which properly handles `PTRACE_O_TRACESYSGOOD`.

kernel/src/process/posix_thread/ptrace/mod.rs
vvvvsv: > (a) targets the tracee's _per-thread_ signal queue, so for a multi-threaded tracee only the thread that was being traced observes it from its own queue — Linux sends a process-wide SIGKILL that tears down the whole thread group;

Linux also sends `SIGKILL` to the [traced task](https://elixir.bootlin.com/linux/v6.16.5/source/kernel/ptrace.c#L600), and the whole thread group then dies because SIGKILL is fatal group-exit semantics. The other two points have been fixed.

kernel/src/process/posix_thread/ptrace/mod.rs
vvvvsv: Our goal is to minimize the performance impact of the ptrace subsystem on non-tracee threads, so I think the current approach is better.

kernel/src/process/clone.rs
vvvvsv: On Linux, `CLONE_UNTRACED` suppresses the parent’s `PTRACE_EVENT_CLONE/FORK/VFORK` stop as well.

https://elixir.bootlin.com/linux/v6.16.5/source/kernel/fork.c#L2581-L2597

test/initramfs/src/regression/process/ptrace/set_options.c
vvvvsv: This is already covered in:

https://github.com/asterinas/asterinas/blob/b19581bd65ef28a8644731a68e62566898f64763/test/initramfs/src/regression/process/ptrace/ptrace.c#L163-L207

book/src/kernel/linux-compatibility/syscall-flag-coverage/process-and-thread-management/README.md
```diff
@@ -98,11 +98,14 @@ Supported requests:
 * `PTRACE_SINGLESTEP` (x86-64 only)
 * `PTRACE_GETREGS` (x86-64 only)
 * `PTRACE_SETREGS` (x86-64 only)
+* `PTRACE_SETOPTIONS`
+* `PTRACE_GETEVENTMSG`
 * `PTRACE_GETSIGINFO`
 
 Additional limitations:
```
tatetian: "Additional limitations" --> "Limitations"

The previoius paragraph is called "Supported requests". So the more natural read for the following points are "Limitations" of the "Supported requests", rather than "Additional limitations".

kernel/src/process/posix_thread/exit.rs
tatetian, line 25: You might want to mention why a mutable reference to `user_ctx` is taken as the input. This could be surprising to a future reader.

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -142,6 +175,28 @@ impl PosixThread {
...
+    /// Gets the extra message of the last ptrace-stop,
+    /// if it is a ptrace-event-stop.
+    /// Otherwise returns `Ok(0)`.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    pub fn ptrace_get_eventmsg(&self) -> Result<Option<usize>> {
+        let status = self.get_tracee_status()?;
+        status.get_eventmsg()
```
tatetian: At this level, it is still better to make the method more high level (including its name) and return a rich-typed return value, rather than the raw and opaque `Option<usize>`. You can turn this value into a raw integer value at the syscall layer.

tatetian (commented):
The PR looks great. Just a few nits.

tatetian (approved):

PR #3064, opened 2026-03-24
Add /proc/sys/kernel/random/uuid Support

myml:
a random UUID that is generated once at boot time.

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/sys/kernel/random/boot_id.rs
+++ b/kernel/src/fs/fs_impls/procfs/sys/kernel/random/boot_id.rs
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! This module offers `/proc/sys/kernel/random/boot_id` file support, which provides
+//! a random UUID that is generated once at boot time.
+//!
+//! Reference: <https://man7.org/linux/man-pages/man5/proc.5.html>
+
+use alloc::format;
+use aster_util::printer::VmPrinter;
+use spin::Once;
+
+use crate::{
+    fs::{
+        file::mkmod,
+        procfs::template::{FileOps, ProcFileBuilder},
+        vfs::inode::Inode,
+    },
+    prelude::*,
+    util::random::getrandom,
+};
+
+/// Represents the inode at `/proc/sys/kernel/random/boot_id`.
+pub struct BootIdFileOps;
+
+impl BootIdFileOps {
+    pub fn new_inode(parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        ProcFileBuilder::new(Self, mkmod!(a+r))
+            .parent(parent)
+            .build()
+            .unwrap()
+    }
+}
+
+impl FileOps for BootIdFileOps {
+    fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        let mut printer = VmPrinter::new_skip(writer, offset);
+
+        // Get the boot ID (UUID)
+        let boot_id = get_boot_id();
+
+        writeln!(printer, "{}", boot_id)?;
+        Ok(printer.bytes_written())
+    }
+}
+
+/// Returns the boot ID as a UUID string.
+/// The boot ID is generated once at boot time and remains constant.
+fn get_boot_id() -> String {
+    static BOOT_ID: Once<String> = Once::new();
+
+    BOOT_ID.call_once(|| {
+        // Generate random bytes for UUID
+        let mut uuid_bytes = [0u8; 16];
+        getrandom(&mut uuid_bytes);
+
+        // Format as UUID string (8-4-4-4-12)
+        format_uuid(&uuid_bytes)
+    }).to_string()
+}
+
+/// Formats 16 bytes as a UUID string (8-4-4-4-12 format)
+fn format_uuid(bytes: &[u8; 16]) -> String {
+    format!(
+        "{:02x}{:02x}{:02x}{:02x}-{:02x}{:02x}-{:02x}{:02x}-{:02x}{:02x}-{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}",
+        bytes[0], bytes[1], bytes[2], bytes[3],
+        bytes[4], bytes[5],
+        bytes[6], bytes[7],
+        bytes[8], bytes[9],
+        bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15]
+    )
+}
--- a/kernel/src/fs/fs_impls/procfs/sys/kernel/random/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/sys/kernel/random/mod.rs
@@ -8,10 +8,7 @@ use crate::{
         file::mkmod,
         procfs::{
             ProcDir,
-            sys::kernel::{
-                cap_last_cap::CapLastCapFileOps, pid_max::PidMaxFileOps, random::RandomDirOps,
-                yama::YamaDirOps,
-            },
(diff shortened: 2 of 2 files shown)


StevenJiang1110:
Thanks for your contribution! It looks like some lint CI checks are failing. You can run `make format` locally to fix trivial formatting issues, and run `make check` to see if there are any remaining problems reported by other lint tools and address them accordingly.

kernel/src/fs/fs_impls/procfs/sys/kernel/random/boot_id.rs
```diff
@@ -0,0 +1,71 @@
...
+    static BOOT_ID: Once<String> = Once::new();
+
+    BOOT_ID.call_once(|| {
+        // Generate random bytes for UUID
+        let mut uuid_bytes = [0u8; 16];
+        getrandom(&mut uuid_bytes);
+
+        // Format as UUID string (8-4-4-4-12)
+        format_uuid(&uuid_bytes)
+    }).to_string()
```
StevenJiang1110, line 58: You don't need to call `to_string` here — doing so introduces unnecessary overhead by allocating a new String. Having `get_boot_id` return `&'static str` directly should suffice.

kernel/src/fs/fs_impls/procfs/sys/kernel/random/boot_id.rs
```diff
@@ -0,0 +1,71 @@
...
+        // Get the boot ID (UUID)
+        let boot_id = get_boot_id();
+
+        writeln!(printer, "{}", boot_id)?;
+        Ok(printer.bytes_written())
+    }
+}
+
+/// Returns the boot ID as a UUID string.
+/// The boot ID is generated once at boot time and remains constant.
```
StevenJiang1110, line 47: Add a blank line between the two comment lines, like this:

```rust
/// Returns the boot ID as a UUID string.
///
/// The boot ID is generated once at boot time and remains constant.
```

In Rust, API documentation conventionally begins with a single summary sentence, followed by a blank line, and then a more detailed description.

kernel/src/fs/fs_impls/procfs/sys/kernel/random/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+};
+
+mod boot_id;
+
+/// Represents the inode at `/proc/sys/kernel/random`.
+pub struct RandomDirOps;
+
+impl RandomDirOps {
+    pub fn new_inode(parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
+        ProcDirBuilder::new(Self, mkmod!(a+rx))
```
StevenJiang1110, line 28: To follow our codebase conventions, please add a Linux reference for the directory or file permission, like this.

https://github.com/asterinas/asterinas/blob/916ea0e0818cc4ca9d0745d02afab7a44f13594d/kernel/src/fs/fs_impls/procfs/pid/task/cmdline.rs#L19-L22

Similar changes are needed in other places as well.

kernel/src/fs/fs_impls/procfs/sys/kernel/random/boot_id.rs
```diff
@@ -0,0 +1,71 @@
...
+fn get_boot_id() -> String {
+    static BOOT_ID: Once<String> = Once::new();
+
+    BOOT_ID.call_once(|| {
+        // Generate random bytes for UUID
+        let mut uuid_bytes = [0u8; 16];
+        getrandom(&mut uuid_bytes);
+
+        // Format as UUID string (8-4-4-4-12)
+        format_uuid(&uuid_bytes)
```
tatetian, line 57: **UUID should be a reusable type in `kernel/src/util/`, not an inline helper**: UUID generation has uses beyond `boot_id` -- filesystems need it for superblock UUIDs, and `/proc/sys/kernel/random/uuid` (which generates a fresh UUID on every read) will need the same logic. In Linux, the generation function [`generate_random_uuid`](https://elixir.bootlin.com/linux/v6.18/source/lib/uuid.c#L33) lives in a utility module (`lib/uuid.c`), not inside procfs.

Please create `kernel/src/util/uuid.rs` with a `Uuid` struct that conforms to UUID version 4 (RFC 4122). The current code is also missing the version and variant bits that make it a valid UUID v4. Here is a reference implementation:

```rust
// kernel/src/util/uuid.rs

use crate::util::random::getrandom;
use core::fmt;

/// A version-4 (random) UUID as defined by RFC 4122.
///
/// A UUID is a 128-bit identifier displayed as 32 hex digits
/// in five groups separated by hyphens: `xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx`,
/// where `M` encodes the version (4 = random) and the upper bits of `N`
/// encode the variant (binary `10` = RFC 4122).
///
/// Reference: <https://datatracker.ietf.org/doc/html/rfc4122#section-4.4>.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub struct Uuid([u8; 16]);

impl Uuid {
    /// Generates a new random UUID (version 4, variant 1).
    ///
    /// All 128 bits are filled with cryptographic randomness,
    /// then the version and variant fields are stamped per RFC 4122:
    /// - Byte 6, upper nibble → `0100` (version 4).
    /// - Byte 8, upper two bits → `10` (variant 1).
    ///
    /// This is equivalent to Linux's
    /// [`generate_random_uuid`](https://elixir.bootlin.com/linux/v6.18/source/lib/uuid.c#L33).
    pub fn new() -> Self {
        let mut bytes = [0u8; 16];
        getrandom(&mut bytes);

        bytes[6] = (bytes[6] & 0x0f) | 0x40; // version 4
        bytes[8] = (bytes[8] & 0x3f) | 0x80; // variant 1

        Self(bytes)
    }
}

impl fmt::Display for Uuid {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let b = &self.0;
        write!(
            f,
            "{:02x}{:02x}{:02x}{:02x}-\
             {:02x}{:02x}-\
             {:02x}{:02x}-\
             {:02x}{:02x}-\
             {:02x}{:02x}{:02x}{:02x}{:02x}{:02x}",
            b[0], b[1], b[2], b[3],
            b[4], b[5],
            b[6], b[7],
            b[8], b[9],
            b[10], b[11], b[12], b[13], b[14], b[15],
        )
    }
}

// TODO: Add unit tests
```

With this, `boot_id.rs` reduces to:

```rust
fn get_boot_id() -> &'static Uuid {
    static BOOT_ID: Once<Uuid> = Once::new();
    BOOT_ID.call_once(Uuid::new)
}
```

and `read_at` becomes `writeln!(printer, "{}", get_boot_id())?;` -- no `String` allocation at all.

kernel/src/fs/fs_impls/procfs/sys/kernel/random/boot_id.rs
tatetian, line 58: [**`minimize-copies`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/performance.html#minimize-copies): `call_once` returns `&String`, so `.to_string()` allocates a new `String` on every read. With the `Uuid` type suggested above, this becomes zero-allocation: `Once<Uuid>` stores a 16-byte `Copy` type, and `Display` formats directly into the `VmPrinter` without an intermediate `String`.

tatetian (commented):
Thanks for the contribution!

Two issues: UUID generation should be a reusable type in `kernel/src/util/` (not an inline helper), and every read allocates a `String` unnecessarily.

tatetian:
@myml Are you still interested in pursuing this PR?

PR #2984, opened 2026-02-27
Add `ptrace` syscall

vvvvsv:
~Based on https://github.com/asterinas/asterinas/pull/2855.~

This PR and https://github.com/asterinas/asterinas/pull/3065 will complete the phase 1 of https://github.com/asterinas/asterinas/issues/2323. This makes it possible to run a [GDB-like debugger program](https://github.com/asterinas/asterinas/blob/bdebb99bd03b534fca001a214646d3e03f8b8c0d/test/initramfs/src/apps/process/ptrace/ptracer.c) in Asterinas.

This PR contains the basic ptrace stop / wait / continue support. #3065 supports debugging with breakpoints and single-step.

# Design Notes (of this PR and https://github.com/asterinas/asterinas/pull/3065)

Below, we refer to the **tracer** as the thread that traces another thread (e.g., GDB), and the **tracee** as the thread being traced.

## Goal

What exactly is our goal? 
 - Supporting GDB and ensuring that incorrect `ptrace` calls cannot corrupt the kernel,
 - Or fully supporting all `ptrace` calls?

I personally prefer the former. The following discussion is based on it.

## The `ptrace` Model

In Linux, each thread can act as a tracer or a tracee. However, in the GDB usage, the tracer is always a single-threaded process. When GDB debugs a multi-threaded program, its main thread performs `ptrace` on each thread individually. Therefore, we can simplify the implementation by only supporting the main thread as the tracer. 

A debugger interacts with the debuggee as follows:

```
[debugger]                        [debuggee]

                               enters ptrace-stop
returns from wait()
// do something
ptrace(PTRACE_CONT)
                               resumes from ptrace-stop
```

## Ptrace-stops

**Something I said in the community meeting on 2026.02.13 was incorrect. I described “ptrace-stop” as a process-level stop, but that was incorrect. Unlike `SIGSTOP`, which stops the entire process, `ptrace-stop` only stops a single thread.**

In Linux, a ptrace-stop can only be resumed by `PTRACE_CONT`-family ptrace calls, or interrupted by `SIGKILL`.

https://elixir.bootlin.com/linux/v6.16.5/source/include/linux/sched/signal.h#L439-L447
https://elixir.bootlin.com/linux/v6.16.5/source/kernel/signal.c#L721-L736
https://elixir.bootlin.com/linux/v6.16.5/source/kernel/signal.c#L2379

A ptrace-stop can occur in the middle of a system call:

> PTRACE_EVENT_VFORK
[...]

PR #3065, opened 2026-03-24
Support debugging with `ptrace`

vvvvsv:
Based on https://github.com/asterinas/asterinas/pull/2984.

#2984 and this PR complete the phase 1 of https://github.com/asterinas/asterinas/issues/2323. For design notes, please refer to #2984.

### DIFF
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -4,9 +4,13 @@
 
 use core::sync::atomic::{AtomicBool, Ordering};
 
-use ostd::sync::Waiter;
+#[cfg(target_arch = "x86_64")]
+use ostd::arch::cpu::context::GeneralRegs;
+use ostd::{arch::cpu::context::UserContext, sync::Waiter};
 
 use super::{AsPosixThread, PosixThread};
+#[cfg(target_arch = "x86_64")]
+use crate::{arch::ptrace as arch_ptrace, process::posix_thread::NOT_A_SYSCALL};
 use crate::{
     prelude::*,
     process::{
@@ -64,9 +68,10 @@ impl PosixThread {
         &self,
         signal: DequeuedSignal,
         ctx: &Context,
+        user_ctx: &mut UserContext,
     ) -> PtraceStopResult {
         if let Some(status) = self.tracee_status.get() {
-            status.ptrace_stop(signal, ctx)
+            status.ptrace_stop(signal, ctx, user_ctx)
         } else {
             PtraceStopResult::NotTraced(signal)
         }
@@ -91,6 +96,50 @@ impl PosixThread {
         self.wake_signalled_waker();
 
         Ok(())
+    }
+
+    /// Gets the general-purpose registers of this thread for ptrace.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    #[cfg(target_arch = "x86_64")]
+    pub fn ptrace_get_regs(&self) -> Result<arch_ptrace::CUserRegsStruct> {
+        let status = self.get_tracee_status()?;
+        status.get_regs()
+    }
+
+    /// Sets the general-purpose registers of this thread for ptrace.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    #[cfg(target_arch = "x86_64")]
+    pub fn ptrace_set_regs(&self, regs: arch_ptrace::CUserRegsStruct) -> Result<()> {
+        let status = self.get_tracee_status()?;
+        status.set_regs(regs)
+    }
+
+    /// Reads one word in the tracee's USER area.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    #[cfg(target_arch = "x86_64")]
+    pub fn ptrace_peek_user(&self, offset: usize) -> Result<usize> {
+        let status = self.get_tracee_status()?;
+        status.peek_user(offset)
+    }
+
+    /// Writes one word in the tracee's USER area.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    #[cfg(target_arch = "x86_64")]
+    pub fn ptrace_poke_user(&self, offset: usize, value: usize) -> Result<()> {
+        let status = self.get_tracee_status()?;
+        status.poke_user(offset, value)
     }
 
     /// Returns the tracee status of this thread if it has ever been traced.
@@ -206,7 +255,15 @@ impl TraceeStatus {
         self.is_stopped.store(false, Ordering::Relaxed);
     }
 
-    fn ptrace_stop(&self, signal: DequeuedSignal, ctx: &Context) -> PtraceStopResult {
+    fn ptrace_stop(
+        &self,
+        signal: DequeuedSignal,
+        ctx: &Context,
+        user_ctx: &mut UserContext,
+    ) -> PtraceStopResult {
+        #[cfg(not(target_arch = "x86_64"))]
+        let _ = user_ctx;
+
         // Hold the lock first to avoid race conditions.
         let mut state = self.state.lock();
 
@@ -217,6 +274,11 @@ impl TraceeStatus {
         debug_assert!(!self.is_ptrace_stopped());
 
         state.signal.stop(signal);
+        #[cfg(target_arch = "x86_64")]
+        {
+            state.general_regs = Some(*user_ctx.general_regs());
+            state.orig_syscall_ret = ctx.posix_thread.orig_syscall_ret();
+        }
         self.is_stopped.store(true, Ordering::Relaxed);
         drop(state);
 
@@ -239,12 +301,27 @@ impl TraceeStatus {
             // A `SIGKILL` interrupts this ptrace-stop.
             let mut state = self.state.lock();
             state.signal.clear();
+            #[cfg(target_arch = "x86_64")]
+            {
+                state.general_regs = None;
+                state.orig_syscall_ret = NOT_A_SYSCALL;
+            }
             self.is_stopped.store(false, Ordering::Relaxed);
             return PtraceStopResult::Interrupted;
         };
 
         let mut state = self.state.lock();
         let signal = state.signal.clear();
+
+        #[cfg(target_arch = "x86_64")]
+        {
+            let regs = state.general_regs.take().unwrap();
+            *user_ctx.general_regs_mut() = regs;
+            ctx.posix_thread
+                .set_orig_syscall_ret(state.orig_syscall_ret);
+            state.orig_syscall_ret = NOT_A_SYSCALL;
+        }
+
         PtraceStopResult::Continued(signal)
     }
 
@@ -289,12 +366,63 @@ impl TraceeStatus {
 
         Ok(())
     }
+
+    #[cfg(target_arch = "x86_64")]
+    fn get_regs(&self) -> Result<arch_ptrace::CUserRegsStruct> {
+        // Hold the lock first to avoid race conditions.
+        let state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+
+        let regs = state.general_regs.as_ref().unwrap();
+        let mut regs = arch_ptrace::CUserRegsStruct::from(regs);
+        regs.orig_rax = state.orig_syscall_ret;
+        Ok(regs)
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn set_regs(&self, regs: arch_ptrace::CUserRegsStruct) -> Result<()> {
+        // Hold the lock first to avoid race conditions.
+        let mut state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+
+        regs.apply_to(state.general_regs.as_mut().unwrap())?;
+        state.orig_syscall_ret = regs.orig_rax;
+
+        Ok(())
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn peek_user(&self, offset: usize) -> Result<usize> {
+        // Hold the lock first to avoid race conditions.
+        let state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+        let regs = state.general_regs.as_ref().unwrap();
+        arch_ptrace::read_user_word(regs, state.orig_syscall_ret, offset)
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn poke_user(&self, offset: usize, value: usize) -> Result<()> {
+        // Hold the lock first to avoid race conditions.
+        let mut state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+        let mut orig_syscall_ret = state.orig_syscall_ret;
+        let regs = state.general_regs.as_mut().unwrap();
+        arch_ptrace::write_user_word(regs, &mut orig_syscall_ret, offset, value)?;
+        state.orig_syscall_ret = orig_syscall_ret;
+        Ok(())
+    }
 }
 
 struct TraceeState {
     tracer: Weak<Thread>,
     /// The signal associated with the current ptrace-stop and later signal delivery.
     signal: StopDeliverySignal,
+    /// The general-purpose registers of the tracee at the time of ptrace-stop.
+    #[cfg(target_arch = "x86_64")]
+    general_regs: Option<GeneralRegs>,
+    /// The value of `PosixThread::orig_syscall_ret` at the time of ptrace-stop.
+    #[cfg(target_arch = "x86_64")]
+    orig_syscall_ret: usize,
 }
 
 impl TraceeState {
@@ -302,6 +430,10 @@ impl TraceeState {
         Self {
             tracer: Weak::new(),
             signal: StopDeliverySignal::default(),
+            #[cfg(target_arch = "x86_64")]
+            general_regs: None,
+            #[cfg(target_arch = "x86_64")]
+            orig_syscall_ret: NOT_A_SYSCALL,
         }
     }
 
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -142,6 +142,16 @@ impl PosixThread {
         status.poke_user(offset, value)
     }
 
+    /// Gets the signal info of this thread for ptrace.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    pub fn ptrace_get_siginfo(&self) -> Result<siginfo_t> {
+        let status = self.get_tracee_status()?;
+        status.get_siginfo()
+    }
+
     /// Returns the tracee status of this thread if it has ever been traced.
     ///
     /// # Errors
@@ -427,6 +437,14 @@ impl TraceeStatus {
         arch_ptrace::write_user_word(regs, &mut orig_syscall_ret, offset, value)?;
         state.orig_syscall_ret = orig_syscall_ret;
         Ok(())
+    }
+
+    fn get_siginfo(&self) -> Result<siginfo_t> {
+        // Hold the lock first to avoid race conditions.
+        let state = self.state.lock();
+        self.check_ptrace_stopped(&state)?;
+
+        Ok(state.signal.get().unwrap().to_info())
     }
 }
 
--- a/kernel/src/process/posix_thread/ptrace/mod.rs
+++ b/kernel/src/process/posix_thread/ptrace/mod.rs
@@ -252,6 +252,12 @@ impl TraceeStatus {
         let mut state = self.state.lock();
 
         state.tracer = Weak::new();
+        #[cfg(target_arch = "x86_64")]
+        {
+            if let Some(regs) = state.general_regs.as_mut() {
+                arch_ptrace::disable_single_step(regs);
+            }
+        }
         self.is_stopped.store(false, Ordering::Relaxed);
     }
 
@@ -362,6 +368,17 @@ impl TraceeStatus {
         } else {
             state.signal.clear();
         }
+
+        #[cfg(target_arch = "x86_64")]
+        {
+            let regs = state.general_regs.as_mut().unwrap();
+            if matches!(request, PtraceContRequest::SingleStep(_)) {
+                arch_ptrace::enable_single_step(regs);
+            } else {
+                arch_ptrace::disable_single_step(regs);
+            }
+        }
+
         self.is_stopped.store(false, Ordering::Relaxed);
 
         Ok(())
--- a/kernel/src/syscall/ptrace.rs
+++ b/kernel/src/syscall/ptrace.rs
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
 
-#[cfg(target_arch = "x86_64")]
 use ostd::mm::VmIo;
 
 use super::SyscallReturn;
@@ -10,7 +9,7 @@ use crate::{
     prelude::*,
     process::{
         posix_thread::{AsPosixThread, alien_access::AlienAccessMode, ptrace::PtraceContRequest},
-        signal::sig_num::SigNum,
+        signal::{constants::SIGKILL, sig_num::SigNum, signals::user::UserSignal},
     },
     thread::{Thread, Tid},
 };
@@ -60,6 +59,12 @@ pub fn sys_ptrace(
 
             tracee.ptrace_continue(PtraceContRequest::Continue(sig_num), ctx)?;
         }
+        PtraceRequest::PTRACE_KILL => {
+            let tracee = ctx.posix_thread.get_tracee(tid)?;
+            let tracee = tracee.as_posix_thread().unwrap();
+
+            tracee.enqueue_signal(Box::new(UserSignal::new_kill(SIGKILL, ctx)));
+        }
         #[cfg(target_arch = "x86_64")]
         PtraceRequest::PTRACE_SINGLESTEP => {
             let sig_num = parse_ptrace_injected_signal(data)?;
@@ -86,6 +91,12 @@ pub fn sys_ptrace(
                 .user_space()
                 .read_val::<arch_ptrace::CUserRegsStruct>(data)?;
             tracee.ptrace_set_regs(regs)?;
+        }
+        PtraceRequest::PTRACE_GETSIGINFO => {
+            let tracee = ctx.posix_thread.get_tracee(tid)?;
+            let siginfo = tracee.as_posix_thread().unwrap().ptrace_get_siginfo()?;
+
+            ctx.user_space().write_val(data, &siginfo)?;
         }
     }
 
@@ -140,6 +151,8 @@ enum PtraceRequest {
     PTRACE_POKEUSER = 6,
     /// Continue the thread.
     PTRACE_CONT = 7,
+    /// Kill the thread.
+    PTRACE_KILL = 8,
     /// Single step the thread.
     #[cfg(target_arch = "x86_64")]
     PTRACE_SINGLESTEP = 9,
@@ -149,6 +162,8 @@ enum PtraceRequest {
     /// Set all general purpose registers used by a thread.
     #[cfg(target_arch = "x86_64")]
     PTRACE_SETREGS = 13,
+    /// Get the siginfo of the last ptrace-stop.
+    PTRACE_GETSIGINFO = 0x4202,
     // TODO: Support other operations.
     // /// Return the word in the thread's text space at address ADDR.
     // PTRACE_PEEKTEXT = 1,
@@ -158,8 +173,6 @@ enum PtraceRequest {
     // PTRACE_POKETEXT = 4,
     // /// Write the word DATA into the thread's data space at address ADDR.
     // PTRACE_POKEDATA = 5,
-    // /// Kill the thread.
-    // PTRACE_KILL = 8,
     // /// Get all floating point registers used by a thread.
     // PTRACE_GETFPREGS = 14,
     // /// Set all floating point registers used by a thread.
@@ -180,8 +193,6 @@ enum PtraceRequest {
     // PTRACE_SYSEMU_SINGLESTEP = 32,
     // /// Set ptrace filter options.
     // PTRACE_SETOPTIONS = 0x4200,
-    // /// Get the siginfo of the last ptrace-stop.
-    // PTRACE_GETSIGINFO = 0x4202,
     // /// Get register content.
     // PTRACE_GETREGSET = 0x4204,
     // /// Set register content.
--- a/kernel/src/syscall/ptrace.rs
+++ b/kernel/src/syscall/ptrace.rs
@@ -56,10 +56,18 @@ pub fn sys_ptrace(
             let sig_num = parse_ptrace_injected_signal(data)?;
 
             let tracee = ctx.posix_thread.get_tracee(tid)?;
-            tracee
-                .as_posix_thread()
-                .unwrap()
-                .ptrace_continue(PtraceContRequest::Continue(sig_num), ctx)?;
+            let tracee = tracee.as_posix_thread().unwrap();
+
+            tracee.ptrace_continue(PtraceContRequest::Continue(sig_num), ctx)?;
+        }
+        #[cfg(target_arch = "x86_64")]
+        PtraceRequest::PTRACE_SINGLESTEP => {
+            let sig_num = parse_ptrace_injected_signal(data)?;
+
+            let tracee = ctx.posix_thread.get_tracee(tid)?;
+            let tracee = tracee.as_posix_thread().unwrap();
+
+            tracee.ptrace_continue(PtraceContRequest::SingleStep(sig_num), ctx)?;
         }
         #[cfg(target_arch = "x86_64")]
         PtraceRequest::PTRACE_GETREGS => {
@@ -132,6 +140,9 @@ enum PtraceRequest {
     PTRACE_POKEUSER = 6,
     /// Continue the thread.
     PTRACE_CONT = 7,
+    /// Single step the thread.
+    #[cfg(target_arch = "x86_64")]
+    PTRACE_SINGLESTEP = 9,
     /// Get all general purpose registers used by a thread.
     #[cfg(target_arch = "x86_64")]
     PTRACE_GETREGS = 12,
@@ -149,8 +160,6 @@ enum PtraceRequest {
     // PTRACE_POKEDATA = 5,
     // /// Kill the thread.
     // PTRACE_KILL = 8,
-    // /// Single step the thread.
-    // PTRACE_SINGLESTEP = 9,
     // /// Get all floating point registers used by a thread.
     // PTRACE_GETFPREGS = 14,
     // /// Set all floating point registers used by a thread.
--- a/kernel/src/syscall/ptrace.rs
+++ b/kernel/src/syscall/ptrace.rs
@@ -1,6 +1,11 @@
 // SPDX-License-Identifier: MPL-2.0
 
+#[cfg(target_arch = "x86_64")]
+use ostd::mm::VmIo;
+
 use super::SyscallReturn;
+#[cfg(target_arch = "x86_64")]
+use crate::arch::ptrace as arch_ptrace;
 use crate::{
     prelude::*,
     process::{
@@ -13,8 +18,8 @@ use crate::{
 pub fn sys_ptrace(
     request: u32,
     tid: Tid,
-    addr: u64,
-    data: u64,
+    addr: usize,
+    data: usize,
     ctx: &Context,
 ) -> Result<SyscallReturn> {
     let request = PtraceRequest::try_from(request)
@@ -32,6 +37,21 @@ pub fn sys_ptrace(
 
             do_ptrace_attach(&parent_main_thread, current_thread)?;
         }
+        #[cfg(target_arch = "x86_64")]
+        PtraceRequest::PTRACE_PEEKUSER => {
+            let tracee = ctx.posix_thread.get_tracee(tid)?;
+            let tracee = tracee.as_posix_thread().unwrap();
+
+            let val = tracee.ptrace_peek_user(addr)?;
+            ctx.user_space().write_val(data, &val)?;
+        }
+        #[cfg(target_arch = "x86_64")]
+        PtraceRequest::PTRACE_POKEUSER => {
+            let tracee = ctx.posix_thread.get_tracee(tid)?;
+            let tracee = tracee.as_posix_thread().unwrap();
+
+            tracee.ptrace_poke_user(addr, data)?;
+        }
         PtraceRequest::PTRACE_CONT => {
             let sig_num = parse_ptrace_injected_signal(data)?;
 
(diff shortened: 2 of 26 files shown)


vvvvsv:
@boterinas rerun

vvvvsv:
@tatetian Could you please help review this PR, especially the macro design of `ostd::for_all_general_regs!` in https://github.com/asterinas/asterinas/pull/3065/changes/0e6b2bca9a1b287773f430d38f7a969283e3ac44? Looking forward to your insightful suggestions!

tatetian:
I consolidated several major or minor issues that I found into this proposal to refactoring the design. The concrete code is AI-assisted but with my extensive review and revision.

My first impression with the `for_all_general_regs` macro was: "Wow, this is a great idea". By during the refactor, I found it is not essential as it may appear. It can be moved out of OSTD; in fact, the macro can be replaced with (mostly) const expressions.

There are other minor issues; I will post them separately.

# Proposed Refactor: Keep OSTD OS-Agnostic

The single largest concern with this PR is that it teaches OSTD a great deal about the Linux ptrace ABI: `c_user_regs_struct`, `RFLAGS_MASK`, the policy table baked into `for_all_general_regs!`, the segment-register pseudo-fields, `set_single_step`, the `orig_rax` field, and the assembly changes that go with it. None of these are hardware abstractions — they are Linux personality decisions. OSTD is a framekernel framework; Asterinas *happens* to implement the Linux ABI on top of it, but a future personality (a Plan 9 server, a unikernel, a research OS) should not be forced to inherit `struct user_regs_struct`'s layout or Linux's view of which RFLAGS bits the user may modify.

The refactor below moves all of this into `kernel/src/arch/x86/`, where Linux-ABI x86 code already lives (`cpu.rs`, `signal.rs`). After it, OSTD's diff for ptrace support is essentially zero, and the assembly is unchanged.

A secondary (minor) driver for the same refactor: `for_all_general_regs!` is a code-generating macro doing what plain data and an ordinary function can do. The project's [`macros-as-last-resort`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/macros-and-attributes.html#macros-as-last-resort) guideline says to prefer functions over macros where both are workable. The same per-register table can be expressed as a const slice of `RegRule` with a four-variant `Policy` enum and function-pointer accessors — the entries still pair each register with its policy, but dispatch is an ordinary `match` on `Policy` inside `RegRule::apply` rather than three nested layers of `paste::paste!` and X-macro callbacks. "What does ptrace do for `rsp`?" becomes answerable by reading one row of the slice and one match arm, with no macro expansion to mentally unfold. The concrete code below adopts this form.

Several smaller issues are also resolved as a side effect of the refactor — they would otherwise need their own fixes:

- **`orig_rax` is under-documented in the current PR** (the inline-comment review above expands on this). After the refactor the field disappears from `GeneralRegs` entirely; the value lives in the well-documented `PosixThread::orig_syscall_ret` (with its `NOT_A_SYSCALL` sentinel and full-contract doc), so there's nothing left to under-document on the OSTD side.
- **`orig_rax` overlaps with the existing `pre_syscall_ret`** local in the user-mode loop — the current PR has two pieces of state expressing the same thing (entry-time `rax`), one in the kernel's `pre_syscall_ret` parameter chain and one in `GeneralRegs.orig_rax`. The refactor merges them: `pre_syscall_ret` is retired and `handle_pending_signal` reads `orig_syscall_ret` directly, eliminating the duplication.
- **`RFLAGS_MASK` is an ambiguous name** — it doesn't say whether the mask names the modifiable or the unmodifiable bits, and a reader has to follow the macro expansion to find out. Renaming to `USER_MODIFIABLE_RFLAGS_MASK` makes the intent obvious at every call site, and the rename rides along with the move into `kernel/src/arch/x86/ptrace.rs`.

### File layout

```
ostd/src/arch/x86/
├── cpu/context/mod.rs   # GeneralRegs unchanged from main; no orig_rax,
│                        # no c_user_regs_struct, no for_all_general_regs!,
│                        # no set_single_step, no RFLAGS_MASK
└── trap/{syscall.S,trap.S}  # unchanged from main

kernel/src/arch/x86/
├── cpu.rs               # existing
├── signal.rs            # existing
└── ptrace.rs            # NEW — everything Linux-ptrace-specific

kernel/src/process/posix_thread/
├── mod.rs               # adds `orig_syscall_ret: AtomicUsize`
└── ptrace/{mod,util}.rs # arch-neutral; calls into crate::arch::x86::ptrace
```

### Minimal OSTD addition (the only one)

The kernel needs the GDT-defined user CS/SS values to validate user-supplied selectors. Expose them as plain constants — this is hardware-level, not Linux-level:

```rust
// ostd/src/arch/x86/cpu/context/mod.rs (or a sibling `selector.rs`)

/// User-space CS selector value, as configured by OSTD's GDT.
pub const USER_CS_VALUE: u16 = crate::arch::trap::gdt::USER_CS.0;
/// User-space SS selector value, as configured by OSTD's GDT.
pub const USER_SS_VALUE: u16 = crate::arch::trap::gdt::USER_SS.0;
```

Everything else this PR added to `ostd/src/arch/x86/cpu/context/mod.rs` reverts (no `orig_rax` field, no `for_all_general_regs!`, no `RFLAGS_MASK`, no `c_user_regs_struct`, no `set_single_step`, no segment-register getters). The original `cpu_context_impl_getter_setter!` macro stays as-is.

### `kernel/src/arch/x86/ptrace.rs` (NEW)

```rust

//! x86-64 ptrace ABI module.
//! 
//! This module provides Linux's `struct user_regs_struct` layout,
//! the per-register set-policy table,
//! and helpers used by the architecture-neutral `ptrace` module.

use ostd::{
    arch::cpu::context::{GeneralRegs, USER_CS_VALUE, USER_SS_VALUE},
    Pod,
};
use ostd_pod::IntoBytes;
use x86_64::registers::rflags::RFlags;

use crate::prelude::*;

// =====================================================================
// Public ABI mirror.
// =====================================================================

/// Mirror of Linux's `struct user_regs_struct` for x86-64.
//
// Reference:
// <https://elixir.bootlin.com/linux/v6.16.5/source/arch/x86/include/asm/user_64.h#L66-L97>
#[expect(missing_docs)]
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, Pod)]
pub struct CUserRegsStruct {
    pub r15: usize,
    pub r14: usize,
    pub r13: usize,
    pub r12: usize,
    pub rbp: usize,
    pub rbx: usize,
    pub r11: usize,
    pub r10: usize,
    pub r9: usize,
    pub r8: usize,
    pub rax: usize,
    pub rcx: usize,
    pub rdx: usize,
    pub rsi: usize,
    pub rdi: usize,
    pub orig_rax: usize,
    pub rip: usize,
    pub cs: usize,
    pub rflags: usize,
    pub rsp: usize,
    pub ss: usize,
    pub fsbase: usize,
    pub gsbase: usize,
    pub ds: usize,
    pub es: usize,
    pub fs: usize,
    pub gs: usize,
}

impl From<&GeneralRegs> for CUserRegsStruct {
    /// Builds the user-visible register snapshot
    /// from the kernel's saved general-purpose registers.
    ///
    /// `orig_rax` is left at zero; callers needing the syscall-entry
    /// value should assign it from `PosixThread::orig_syscall_ret`
    /// after the conversion (e.g. `out.orig_rax = ...`).
    fn from(regs: &GeneralRegs) -> Self {
        let mut out = Self::default();
        let bytes = out.as_bytes_mut();
        for rule in REG_RULES {
            let v = match rule.policy {
                Policy::Fixed(val) => val,
                _ => (rule.get.unwrap())(regs),
            };
            write_word(bytes, rule.offset, v);
        }
        out
    }
}

impl CUserRegsStruct {
    /// Validates this user-supplied register set
    /// and merges it into `regs`.
    ///
    /// Returns `Err(EIO)` on any invalid value;
    /// on `Err`, `*regs` is unmodified (the work is done on a copy).
    ///
    /// Note: this is a method, not a `TryFrom` conversion,
    /// because the `Policy::SetBitsTruncate` policy (used for `rflags`)
    /// preserves kernel-controlled bits from the *existing* `regs`,
    /// so the operation is a merge — not a pure value-to-value conversion.
    /// Callers handling `orig_rax` should read `self.orig_rax` directly
    /// and update their separate `orig_syscall_ret` storage.
    pub fn apply_to(&self, regs: &mut GeneralRegs) -> Result<()> {
        let bytes = self.as_bytes();
        let mut new_regs = *regs;
        for rule in REG_RULES {
            let v = read_word(bytes, rule.offset);
            rule.apply(&mut new_regs, v)?;
        }
        *regs = new_regs;
        Ok(())
    }
}

// =====================================================================
// Constants referenced by REG_RULES below.
// =====================================================================

const USER_CS: usize = USER_CS_VALUE as usize;
const USER_SS: usize = USER_SS_VALUE as usize;

/// RFLAGS bits the user may modify via ptrace.
//
// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/arch/x86/kernel/ptrace.c#L369>
const USER_MODIFIABLE_RFLAGS_MASK: usize = (RFlags::CARRY_FLAG.bits()
    | RFlags::PARITY_FLAG.bits()
    | RFlags::AUXILIARY_CARRY_FLAG.bits()
    | RFlags::ZERO_FLAG.bits()
    | RFlags::SIGN_FLAG.bits()
    | RFlags::TRAP_FLAG.bits()
    | RFlags::DIRECTION_FLAG.bits()
    | RFlags::OVERFLOW_FLAG.bits()
    | RFlags::RESUME_FLAG.bits()
    | RFlags::ALIGNMENT_CHECK.bits()
    | RFlags::NESTED_TASK.bits()) as usize;

// =====================================================================
// Public API (free functions; see also `CUserRegsStruct` methods above).
// =====================================================================

/// Reads one word from the USER area at `offset`.
pub fn read_user_word(regs: &GeneralRegs, orig_rax: usize, offset: usize) -> Result<usize> {
    check_user_offset(offset)?;
    if offset == core::mem::offset_of!(CUserRegsStruct, orig_rax) {
        return Ok(orig_rax);
    }
    let rule = RegRule::for_offset(offset)
        .expect("offset has been validated by `check_user_offset`");
    Ok(match rule.policy {
        Policy::Fixed(v) => v,
        _ => (rule.get.unwrap())(regs),
    })
}

/// Writes one word to the USER area at `offset`.
pub fn write_user_word(
    regs: &mut GeneralRegs,
    orig_rax: &mut usize,
    offset: usize,
    value: usize,
) -> Result<()> {
    check_user_offset(offset)?;
    if offset == core::mem::offset_of!(CUserRegsStruct, orig_rax) {
        *orig_rax = value;
        return Ok(());
    }
    let rule = RegRule::for_offset(offset)
        .expect("offset has been validated by `check_user_offset`");
    rule.apply(regs, value)
}

/// Enables single-step execution by setting the x86 Trap Flag.
pub fn enable_single_step(regs: &mut GeneralRegs) {
    regs.set_rflags(regs.rflags() | RFlags::TRAP_FLAG.bits() as usize);
}

/// Disables single-step execution by clearing the x86 Trap Flag.
pub fn disable_single_step(regs: &mut GeneralRegs) {
    regs.set_rflags(regs.rflags() & !(RFlags::TRAP_FLAG.bits() as usize));
}

// =====================================================================
// Per-register policy table.
// =====================================================================

/// Shorthand for `core::mem::offset_of!(CUserRegsStruct, $field)`.
/// Local to this module.
macro_rules! off {
    ($field:ident) => { core::mem::offset_of!(CUserRegsStruct, $field) };
}

/// One row per register in `CUserRegsStruct`,
/// **except** `orig_rax` — which lives in `PosixThread::orig_syscall_ret`,
/// not in `GeneralRegs`.
/// `From<&GeneralRegs>` and `CUserRegsStruct::apply_to` ignore it
/// (callers manipulate the public field directly);
/// `read_user_word` and `write_user_word` handle the `orig_rax` offset
/// in dedicated branches because they don't have a struct to read it from.
const REG_RULES: &[RegRule] = &[
    RegRule::rw(off!(rax), |r| r.rax(), |r, v| r.set_rax(v), Policy::Set),
    RegRule::rw(off!(rbx), |r| r.rbx(), |r, v| r.set_rbx(v), Policy::Set),
    RegRule::rw(off!(rcx), |r| r.rcx(), |r, v| r.set_rcx(v), Policy::Set),
    RegRule::rw(off!(rdx), |r| r.rdx(), |r, v| r.set_rdx(v), Policy::Set),
    RegRule::rw(off!(rsi), |r| r.rsi(), |r, v| r.set_rsi(v), Policy::Set),
    RegRule::rw(off!(rdi), |r| r.rdi(), |r, v| r.set_rdi(v), Policy::Set),
    RegRule::rw(off!(rbp), |r| r.rbp(), |r, v| r.set_rbp(v), Policy::Set),
    RegRule::rw(off!(r8),  |r| r.r8(),  |r, v| r.set_r8(v),  Policy::Set),
    RegRule::rw(off!(r9),  |r| r.r9(),  |r, v| r.set_r9(v),  Policy::Set),
    RegRule::rw(off!(r10), |r| r.r10(), |r, v| r.set_r10(v), Policy::Set),
    RegRule::rw(off!(r11), |r| r.r11(), |r, v| r.set_r11(v), Policy::Set),
    RegRule::rw(off!(r12), |r| r.r12(), |r, v| r.set_r12(v), Policy::Set),
    RegRule::rw(off!(r13), |r| r.r13(), |r, v| r.set_r13(v), Policy::Set),
    RegRule::rw(off!(r14), |r| r.r14(), |r, v| r.set_r14(v), Policy::Set),
    RegRule::rw(off!(r15), |r| r.r15(), |r, v| r.set_r15(v), Policy::Set),

    RegRule::rw(off!(rsp),    |r| r.rsp(),    |r, v| r.set_rsp(v),
                Policy::SetIf(user_addr)),
    RegRule::rw(off!(rip),    |r| r.rip(),    |r, v| r.set_rip(v),
                Policy::SetIf(user_addr)),
    RegRule::rw(off!(fsbase), |r| r.fsbase(), |r, v| r.set_fsbase(v),
                Policy::SetIf(user_addr)),
    RegRule::rw(off!(gsbase), |r| r.gsbase(), |r, v| r.set_gsbase(v),
                Policy::SetIf(user_addr)),

    RegRule::rw(off!(rflags), |r| r.rflags(), |r, v| r.set_rflags(v),
                Policy::SetBitsTruncate(USER_MODIFIABLE_RFLAGS_MASK)),

    // Pseudo-fields:
    // Asterinas does not run 32-bit user code,
    // so the segment selectors are always at fixed values.
    RegRule::fixed(off!(cs), USER_CS),
    RegRule::fixed(off!(ss), USER_SS),
    RegRule::fixed(off!(ds), 0),
    RegRule::fixed(off!(es), 0),
    RegRule::fixed(off!(fs), 0),
    RegRule::fixed(off!(gs), 0),
];

const _: () = {
    // Compile-time check
    // that REG_RULES (one entry per non-`orig_rax` field)
    // plus the explicit `orig_rax` slot
    // cover every word of `CUserRegsStruct`.
    assert!(
        (REG_RULES.len() + 1) * core::mem::size_of::<usize>()
            == core::mem::size_of::<CUserRegsStruct>()
    );
};

// =====================================================================
// RegRule and Policy.
// =====================================================================

/// One row of the per-register table.
#[derive(Clone, Copy)]
struct RegRule {
    /// Byte offset of this register inside `CUserRegsStruct`.
    offset: usize,
    /// Read accessor on `GeneralRegs`. `None` for `Policy::Fixed`,
    /// where the value is in the policy itself.
    get: Option<fn(&GeneralRegs) -> usize>,
    /// Write accessor on `GeneralRegs`. `None` for `Policy::Fixed`.
    set: Option<fn(&mut GeneralRegs, usize)>,
    policy: Policy,
}

impl RegRule {
    /// Builds a rule for a `GeneralRegs`-backed register
    /// with both read and write accessors and the given policy.
    const fn rw(
        offset: usize,
        get: fn(&GeneralRegs) -> usize,
        set: fn(&mut GeneralRegs, usize),
        policy: Policy,
    ) -> Self {
        Self { offset, get: Some(get), set: Some(set), policy }
    }

    /// Builds a rule for a pseudo-field that is always at `expected`
    /// and rejects any other write with `EIO`.
    const fn fixed(offset: usize, expected: usize) -> Self {
        Self { offset, get: None, set: None, policy: Policy::Fixed(expected) }
    }

    /// Looks up the rule whose offset matches `offset`.
    fn for_offset(offset: usize) -> Option<&'static Self> {
        REG_RULES.iter().find(|r| r.offset == offset)
    }

    /// Applies a user-supplied write according to this rule's policy.
    /// The match arms are the *only* place that knows how each `Policy`
    /// variant works; adding a new variant is a one-arm change.
    fn apply(&self, regs: &mut GeneralRegs, value: usize) -> Result<()> {
        match self.policy {
            Policy::Set => {
                (self.set.unwrap())(regs, value);
            }
            Policy::SetIf(check) => {
                if !check(value) {
                    return_errno_with_message!(Errno::EIO, "invalid register value");
                }
                (self.set.unwrap())(regs, value);
            }
            Policy::SetBitsTruncate(mask) => {
                let getter = self.get.unwrap();
                let setter = self.set.unwrap();
                let old = getter(regs);
                setter(regs, (old & !mask) | (value & mask));
            }
            Policy::Fixed(expected) => {
                if value != expected {
                    return_errno_with_message!(Errno::EIO, "invalid fixed register value");
                }
            }
        }
        Ok(())
    }
}

/// What to do with a user-supplied write to one register.
#[derive(Clone, Copy)]
enum Policy {
    /// Unrestricted set.
    Set,
    /// Set only if the predicate accepts the value;
    /// otherwise return `EIO`.
    SetIf(fn(usize) -> bool),
    /// Replace only the bits in `mask`,
    /// preserve the rest.
    SetBitsTruncate(usize),
    /// Pseudo-field that is always at `expected`;
    /// any write of a different value is rejected with `EIO`.
    Fixed(usize),
}

/// Predicate used by `Policy::SetIf` for the user-space-address fields
/// (`rsp`, `rip`, `fsbase`, `gsbase`).
fn user_addr(v: usize) -> bool {
    v < ostd::mm::MAX_USERSPACE_VADDR
}

// =====================================================================
// Small helpers.
// =====================================================================

fn check_user_offset(offset: usize) -> Result<()> {
    if !offset.is_multiple_of(core::mem::size_of::<usize>()) {
        return_errno_with_message!(Errno::EIO, "USER-area offset must be `usize`-aligned");
    }
    if offset >= core::mem::size_of::<CUserRegsStruct>() {
        // Returns `EIO` (not `EOPNOTSUPP`) for parity with Linux.
        return_errno_with_message!(Errno::EIO, "USER-area offset is outside `user_regs_struct`");
    }
    Ok(())
}

/// Reads a `usize` from the `Pod`-byte view of a struct.
/// `bytes[offset..offset + size_of::<usize>()]` must be in bounds and aligned;
/// `check_user_offset` guarantees both for caller-provided offsets.
fn read_word(bytes: &[u8], offset: usize) -> usize {
    usize::from_ne_bytes(
        bytes[offset..offset + core::mem::size_of::<usize>()]
            .try_into()
            .unwrap(),
    )
}

fn write_word(bytes: &mut [u8], offset: usize, value: usize) {
    bytes[offset..offset + core::mem::size_of::<usize>()]
        .copy_from_slice(&value.to_ne_bytes());
}
```

How the data-driven form works:

- **`REG_RULES` is pure data.** Each row is a `(offset, get, set, policy)` quadruple. Function pointers (`fn(&GeneralRegs) -> usize`, `fn(&mut GeneralRegs, usize)`) handle the field-name access that previously needed compile-time macro expansion; `Pod`-derived `as_bytes` / `as_bytes_mut` handle the `CUserRegsStruct` side without `unsafe`. To change "what does ptrace do for `rsp`?", edit one line.
- **`Policy` is the typed alphabet of what we'll do with a write.** Adding a new variant (e.g. a `SetMasked { mask, expected }`) is a one-arm change in `RegRule::apply` — every consumer goes through it.
- **The public entry points are short and uniform.** `read_user_word` / `write_user_word` look up by offset; `From<&GeneralRegs>` and `CUserRegsStruct::apply_to` iterate the slice. None of them touch `orig_rax` — it's a public field of `CUserRegsStruct` that callers manipulate directly, so it doesn't need to thread through any operation that lives below it in the layering.
- **`paste::paste!` is gone.** The kernel crate doesn't need the `paste` dep.
- **Compile-time invariants** still hold: `const _: () = assert!(...)` checks that the slice plus the `orig_rax` slot account for every word of `CUserRegsStruct`, so a future field addition that forgets to register a rule fails the build.

(The const-eval check is the same kind the PR has in OSTD today, just relocated.)

### `kernel/src/process/posix_thread/mod.rs` (small addition)

```rust
use core::sync::atomic::{AtomicUsize, Ordering};

/// Sentinel stored in [`PosixThread::orig_syscall_ret`]
/// when the most recent kernel entry was not a syscall.
/// Mirrors Linux's `orig_ax = (unsigned long)-1` convention.
pub const NOT_A_SYSCALL: usize = usize::MAX;

pub struct PosixThread {
    // ...existing fields...

    /// Original value of the syscall-return register
    /// at this thread's most recent kernel entry,
    /// or [`NOT_A_SYSCALL`] when that entry was not a syscall.
    /// See [`Self::orig_syscall_ret`] for the full contract;
    /// this field subsumes the `pre_syscall_ret` local of the user-mode loop.
    orig_syscall_ret: AtomicUsize,
}

impl PosixThread {
    /// Returns the original value of the syscall-return register
    /// (as read by [`UserContext::syscall_ret`])
    /// at this thread's most recent kernel entry.
    ///
    /// On the producer side,
    /// the user-mode main loop maintains this field
    /// by calling [`Self::set_orig_syscall_ret`] on every kernel entry:
    ///
    /// - On syscall entry,
    ///   it stores `user_ctx.syscall_num()`
    ///   *before* `handle_syscall` overwrites the syscall-return register
    ///   with the syscall's return value,
    ///   so the syscall number remains observable for the rest of the entry.
    /// - On every non-syscall entry (exceptions, interrupts),
    ///   it stores [`NOT_A_SYSCALL`],
    ///   mirroring Linux's `orig_rax = (unsigned long)-1` convention
    ///   for non-syscall stops.
    ///
    /// On the consumer side,
    /// two pieces of functionality depend on this value:
    ///
    /// 1. **Signal-driven syscall restart** in `handle_pending_signal`,
    ///    which restores the syscall-return register from this value
    ///    — which may have been modified by the syscall handling logic —
    ///    before re-executing an interrupted syscall.
    /// 2. **ptrace**,
    ///    where personality-specific code
    ///    maps this value to the `orig_rax` field of `struct user_regs_struct` on x86
    ///    (or the equivalent slot on another architecture).
    ///    The sentinel tells the tracer
    ///    that the current stop is not a syscall-stop.
    ///
    /// The value reflects only the most recent kernel entry,
    /// so it is meaningful when read from inside that entry's handlers
    /// (the syscall handler, the signal-delivery path)
    /// or at a ptrace-stop derived from it.
    pub(in crate::process) fn orig_syscall_ret(&self) -> usize {
        self.orig_syscall_ret.load(Ordering::Relaxed)
    }

    /// Sets the [`Self::orig_syscall_ret`] field;
    /// called by the user-mode loop on every kernel entry,
    /// and by ptrace `resume`
    /// to flush a tracer's modification back into thread state.
    pub(in crate::process) fn set_orig_syscall_ret(&self, v: usize) {
        self.orig_syscall_ret.store(v, Ordering::Relaxed);
    }
}
```

The user-mode loop becomes:

```rust
let user_ctx = user_mode.context_mut();
match return_reason {
    ReturnReason::UserException => {
        let exception = user_ctx.take_exception().unwrap();
        ctx.posix_thread.set_orig_syscall_ret(NOT_A_SYSCALL);
        handle_exception(&ctx, user_ctx, exception);
    }
        ctx.posix_thread.set_orig_syscall_ret(user_ctx.syscall_num());
    }
    ReturnReason::KernelEvent => {}
};

if current_thread.is_exited() { break; }

// `handle_pending_signal` reads `orig_syscall_ret` directly when it
// needs the syscall-restart "previous rax" value; the parameter goes
// away.
handle_pending_signal(user_ctx, &ctx);
```

### `kernel/src/process/posix_thread/ptrace/mod.rs` (the kernel side)

`TraceeState` snapshots both fields together at ptrace-stop entry. The existing `TraceeState::new()` gains one extra initialiser (`orig_syscall_ret: NOT_A_SYSCALL`); the value is overwritten on every entry to `ptrace_stop`, so the initial value is never observed by ptrace ops:

```rust
struct TraceeState {
    tracer: Weak<Thread>,
    signal: StopDeliverySignal,
    general_regs: Option<GeneralRegs>,
    /// Snapshot of `PosixThread::orig_syscall_ret` taken at `ptrace_stop` entry,
    /// initialised to [`NOT_A_SYSCALL`] in `TraceeState::new()`.
    orig_syscall_ret: usize,
}
```

The ptrace ops delegate to the arch module:

```rust
use crate::arch::x86::ptrace as arch_ptrace;

fn get_regs(&self) -> Result<arch_ptrace::CUserRegsStruct> {
    let state = self.state.lock();
    self.check_ptrace_stopped(&state)?;
    let regs = state.general_regs.as_ref().unwrap();
    let mut user_regs = arch_ptrace::CUserRegsStruct::from(regs);
    user_regs.orig_rax = state.orig_syscall_ret;
    Ok(user_regs)
}

fn set_regs(&self, user_regs: arch_ptrace::CUserRegsStruct) -> Result<()> {
    let mut state = self.state.lock();
    self.check_ptrace_stopped(&state)?;
    user_regs.apply_to(state.general_regs.as_mut().unwrap())?;
    state.orig_syscall_ret = user_regs.orig_rax;
    Ok(())
}

fn peek_user(&self, offset: usize) -> Result<usize> {
    let state = self.state.lock();
    self.check_ptrace_stopped(&state)?;
    let regs = state.general_regs.as_ref().unwrap();
    arch_ptrace::read_user_word(regs, state.orig_syscall_ret, offset)
}

fn poke_user(&self, offset: usize, value: usize) -> Result<()> {
    let mut state = self.state.lock();
    self.check_ptrace_stopped(&state)?;
    let regs = state.general_regs.as_mut().unwrap();
    let mut new = state.orig_syscall_ret;
    arch_ptrace::write_user_word(regs, &mut new, offset, value)?;
    state.orig_syscall_ret = new;
    Ok(())
}
```

`ptrace_stop` (the entry into ptrace-stop) snapshots `orig_syscall_ret` alongside the registers:

```rust
fn ptrace_stop(
    &self,
    signal: DequeuedSignal,
    ctx: &Context,
    user_ctx: &mut UserContext,
) -> PtraceStopResult {
    let mut state = self.state.lock();
    let Some(tracer) = state.tracer() else {
        return PtraceStopResult::NotTraced(signal);
    };
    debug_assert!(!self.is_ptrace_stopped());

    state.signal.stop(signal);
    {
        state.general_regs = Some(*user_ctx.general_regs());
        state.orig_syscall_ret = ctx.posix_thread.orig_syscall_ret();
    }
    self.is_stopped.store(true, Ordering::Relaxed);
    drop(state);

    // ...notify tracer, pause...
}
```

And on resume, `orig_rax` flows back to the per-thread cell:

```rust
let mut state = self.state.lock();
let signal = state.signal.clear();
{
    *user_ctx.general_regs_mut() = state.general_regs.take().unwrap();
    ctx.posix_thread.set_orig_syscall_ret(state.orig_syscall_ret);
}
PtraceStopResult::Continued(signal)
```

`detach_tracer` and `resume` use `arch_ptrace::disable_single_step` and `enable_single_step` directly:

```rust
fn detach_tracer(&self) {
    let mut state = self.state.lock();
    state.tracer = Weak::new();
    {
        if let Some(regs) = state.general_regs.as_mut() {
            arch_ptrace::disable_single_step(regs);
        }
    }
    self.is_stopped.store(false, Ordering::Relaxed);
}

fn resume(&self, request: PtraceContRequest, ctx: &Context) -> Result<()> {
    let mut state = self.state.lock();
    self.check_ptrace_stopped(&state)?;
    if let Some(sig_num) = request.sig_num() {
        state.signal.inject(Box::new(UserSignal::new_kill(sig_num, ctx)));
    } else {
        state.signal.clear();
    }
    {
        let regs = state.general_regs.as_mut().unwrap();
        if matches!(request, PtraceContRequest::SingleStep(_)) {
            arch_ptrace::enable_single_step(regs);
        } else {
            arch_ptrace::disable_single_step(regs);
        }
    }
    self.is_stopped.store(false, Ordering::Relaxed);
    Ok(())
}
```

### `kernel/src/syscall/ptrace.rs` (the only change)

Just swap the `c_user_regs_struct` import:

```rust
use crate::arch::x86::ptrace::CUserRegsStruct;
```

### What changes vs the current PR

| | Current PR | Proposal |
|---|---|---|
| OSTD lines added | ~150 | ~5 (two `pub const`s) |
| `paste` dependency added to | OSTD | nowhere (the X-macro is gone; const slice replaces it) |
| Assembly changes (`syscall.S`/`trap.S`) | yes (offset shifts + push/pop) | none |
| `orig_rax` field on `GeneralRegs` | yes | no — lives in `PosixThread::orig_syscall_ret` |
| `set_single_step` bool-arg violation | yes | gone (split into two free fns in `arch::x86::ptrace`) |
| `c_user_regs_struct` snake_case violation | yes | gone (renamed to `CUserRegsStruct` in kernel) |
| `EOPNOTSUPP` vs Linux's `EIO` divergence | yes | gone (`EIO`) |
| `pre_syscall_ret` parameter threading | yes | gone (read from `PosixThread`) |
| Per-syscall fast-path overhead | 3 ALU ops + push + pop | 1 atomic store (cheaper, already-cached cacheline) |
| `for_all_general_regs!` X-macro | exported via `#[macro_export]` | gone (replaced by `REG_RULES` const slice) |

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -71,6 +72,141 @@ pub struct GeneralRegs {
...
+        $macro!($($arg,)* [r11, set]);
+        $macro!($($arg,)* [r12, set]);
+        $macro!($($arg,)* [r13, set]);
+        $macro!($($arg,)* [r14, set]);
+        $macro!($($arg,)* [r15, set]);
+        // This is more strict than Linux.
+        $macro!($($arg,)* [rip, set_if(|rip| rip < $crate::mm::MAX_USERSPACE_VADDR)]);
+        $macro!($($arg,)* [rflags, set_bits_truncate($crate::arch::cpu::context::RFLAGS_MASK)]);
+        // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/arch/x86/kernel/ptrace.c#L389-L400>
+        $macro!($($arg,)* [fsbase, set_if(|fsbase| fsbase < $crate::mm::MAX_USERSPACE_VADDR)]);
```
tatetian: **`arch_prctl(ARCH_SET_FS/_GS)` is missing the same range check** (significant): The new ptrace check matches Linux exactly — `arch/x86/kernel/ptrace.c` rejects `value >= TASK_SIZE_MAX` on `fs_base`/`gs_base` with `EIO` (the snippet linked above), and `arch/x86/kernel/process_64.c` does the symmetric check on `arch_prctl(ARCH_SET_FS/_GS)` with `EPERM`. Linux is internally consistent. Asterinas isn't: `kernel/src/syscall/arch_prctl.rs:36` (`ARCH_SET_FS`) calls `user_ctx.set_tls_pointer(addr as usize)` with no bound on `addr`, so a process can legitimately leave its `fsbase` ≥ `MAX_USERSPACE_VADDR`. A debugger that then does `PTRACE_GETREGS` followed by an unmodified `PTRACE_SETREGS` (the canonical "change one register, write back" pattern) fails with `EIO` on the unchanged `fs_base`. The fix is on the `arch_prctl` side, not the ptrace side — the ptrace check is correct as written. Add the matching guard:

```rust
// kernel/src/syscall/arch_prctl.rs
ArchPrctlCode::ARCH_SET_FS => {
    if addr as usize >= ostd::mm::MAX_USERSPACE_VADDR {
        return_errno_with_message!(Errno::EPERM, "fsbase must be a user-space address");
    }
    user_ctx.set_tls_pointer(addr as usize);
    user_ctx.activate_tls_pointer();
    Ok(0)
}
```

(Note: this is an ABI-correctness fix, not a soundness fix — OSTD's `unsafe { wrfsbase(...) }` block remains sound regardless of the value passed. The kernel doesn't dereference `fs:`-relative addresses, and any user-mode access through a kernel-range `fsbase` faults on the page-table U/S bit. Linux validates here for symmetry with the ptrace path and to give debuggers a sane round-trip.)

kernel/src/process/posix_thread/ptrace/mod.rs
```diff
@@ -93,6 +96,60 @@ impl PosixThread {
         Ok(())
     }
 
+    /// Gets the general-purpose registers of this thread for ptrace.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ESRCH` if this thread is not ptrace-stopped.
+    #[cfg(target_arch = "x86_64")]
+    pub fn ptrace_get_regs(&self) -> Result<c_user_regs_struct> {
```
tatetian: [**`narrow-visibility`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#narrow-visibility) (minor): All five new methods on `PosixThread` (`ptrace_get_regs` line 105, `ptrace_set_regs` line 116, `ptrace_peek_user` line 127, `ptrace_poke_user` line 138, `ptrace_get_siginfo` line 148) are declared `pub` but are only called from `kernel/src/syscall/ptrace.rs`. The pre-existing siblings in this file already use `pub(in crate::process)` (e.g. `is_traced` line 35, `detach_tracer` line 55, `ptrace_stop` line 65) — match the local convention.

While here, the `# Errors` sections only mention `ESRCH`. `ptrace_set_regs` / `ptrace_poke_user` propagate `EIO` from the per-field setter (invalid `rsp` / `rip` / `fsbase` / `gsbase` / fixed-segment value), and `ptrace_peek_user` / `ptrace_poke_user` propagate `EIO` (unaligned offset) and `EOPNOTSUPP` (offset past `c_user_regs_struct`) from `check_user_offset`. List those.

```rust
    /// Gets the general-purpose registers of this thread for ptrace.
    ///
    /// # Errors
    ///
    /// Returns `ESRCH` if this thread is not ptrace-stopped.
    pub(in crate::process) fn ptrace_get_regs(&self) -> Result<c_user_regs_struct> {
```

kernel/src/syscall/ptrace.rs
```diff
@@ -86,29 +141,36 @@ fn parse_ptrace_injected_signal(data: u64) -> Result<Option<SigNum>> {
 enum PtraceRequest {
     /// Indicate that this thread should be traced by its parent.
     PTRACE_TRACEME = 0,
+    /// Return the word in the thread's user area at offset ADDR.
```
tatetian: [**`doc-summary-tense`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#doc-summary-tense) (minor): Doc-comment summary lines should use third-person singular present, not bare imperative. Also wrap identifiers (`ADDR`, `DATA`, `siginfo`) in backticks per [`backtick-identifiers`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#backtick-identifiers). The pre-existing `PTRACE_TRACEME` ("Indicate…") and the commented-out lines below have the same issue, but the new variants shouldn't extend the pattern.

```rust
    /// Returns the word in the thread's user area at offset `ADDR`.
    PTRACE_PEEKUSER = 3,
    /// Writes the word `DATA` into the thread's user area at offset `ADDR`.
    PTRACE_POKEUSER = 6,
    /// Continues the thread.
    PTRACE_CONT = 7,
    /// Kills the thread.
    PTRACE_KILL = 8,
    /// Single-steps the thread.
    PTRACE_SINGLESTEP = 9,
    /// Returns all general-purpose registers of the thread.
    PTRACE_GETREGS = 12,
    /// Sets all general-purpose registers of the thread.
    PTRACE_SETREGS = 13,
    /// Returns the `siginfo` of the last ptrace-stop.
    PTRACE_GETSIGINFO = 0x4202,
```

test/initramfs/src/regression/process/ptrace/read_write_regs.c
tatetian, line 1: **Test gaps for new edge cases** (significant): Three coverage holes worth filling alongside the existing register-round-trip tests:

1. **`orig_rax` semantics.** Currently `orig_rax` only appears inside `FOR_EACH_USER_REG` as a peek-vs-getregs equality check (line 63), which proves the two views agree but not that the kernel populates the field. Add an assertion that `orig_rax` holds a syscall number at the SIGSTOP-stop (the stop is reached via `raise(SIGSTOP)` → `tgkill`):
   ```c
   #include <sys/syscall.h>  // need __NR_*; not pulled in transitively
   TEST_RES(ptrace(PTRACE_PEEKUSER, pid, USER_REG_OFFSET(orig_rax), 0),
            _ret == __NR_tgkill || _ret == __NR_rt_tgsigqueueinfo);
   ```

2. **Out-of-range `PEEKUSER`/`POKEUSER` returns `EOPNOTSUPP`.** Today only the unaligned-offset case is tested (line 107–108). Add the in-bounds-but-too-large case so a regression that flips it back to `EIO` is caught:
   ```c
   TEST_ERRNO(ptrace(PTRACE_PEEKUSER, pid, sizeof(struct user_regs_struct), 0),
              EOPNOTSUPP);
   TEST_ERRNO(ptrace(PTRACE_POKEUSER, pid, sizeof(struct user_regs_struct), 0),
              EOPNOTSUPP);
   ```
   (If the EIO/EOPNOTSUPP debate above resolves to EIO, update both test and kernel together.)

3. **Signal-injecting `PTRACE_CONT`/`PTRACE_SINGLESTEP`.** Every `ptrace(PTRACE_CONT, …)` and `ptrace(PTRACE_SINGLESTEP, …)` in this file and in `debugger.c` passes `data == 0`. The `data != 0` path runs through `parse_ptrace_injected_signal` and `state.signal.inject()` (`util.rs:102`), which is a distinct code path and the typical real-debugger use (re-injecting an intercepted SIGSEGV, etc.). Add a small `FN_TEST` that catches a tracee-raised `SIGUSR1`, calls `ptrace(PTRACE_CONT, pid, 0, SIGUSR1)`, and asserts the tracee's installed handler observes the signal.

kernel/src/process/posix_thread/ptrace/mod.rs
tatetian, line 1: **Missing test: tracer exits while tracee is in ptrace-stop** (significant): The existing test set covers `PTRACE_KILL` interrupting a stop, but the symmetric path — tracer process `_exit`s (or simply dies) while the tracee sits in a ptrace-stop — flows through `clear_tracees` → `detach_tracer` (line 258) and is responsible for releasing the tracee. This is a frequent source of real bugs (lock-order, dangling weak refs, the `state.general_regs = Some(_)` left behind by `detach_tracer` and drained by the wake-up at line 330). Add an `FN_TEST` to `test/initramfs/src/regression/process/ptrace/ptrace.c` along the lines of:

```c
FN_TEST(tracer_exit_releases_stopped_tracee)
{
    SKIP_TEST_IF(read_yama_scope() == YAMA_SCOPE_NO_ATTACH);
    pid_t tracer = fork();
    if (tracer == 0) {
        pid_t tracee = fork();
        if (tracee == 0) {
            CHECK(ptrace(PTRACE_TRACEME, 0, 0, 0));
            CHECK(raise(SIGSTOP));
            exit(0);
        }
        int s; waitpid(tracee, &s, 0);  // observe the stop
        _exit(0);                        // tracer dies with tracee stopped
    }
    int s; waitpid(tracer, &s, 0);
    // Re-parented tracee should run to completion now that its tracer is gone.
    // (Walk /proc to find the tid, or use a pipe to plumb it up.)
}
```

test/initramfs/src/conformance/gvisor/blocklists/ptrace_test
```diff
@@ -0,0 +1,16 @@
+PtraceTest.Attach*
```
tatetian, line 1: **Blocklist entries lack rationale** (minor): 16 sub-tests are blocked with no comment explaining *why* each is blocked. Without that, a future contributor who implements (e.g.) `PTRACE_ATTACH` can't tell which entries to retire. Group with one-line comments:

```
# PTRACE_ATTACH / PTRACE_SEIZE not yet supported.
PtraceTest.Attach*
PtraceTest.Seize_Interrupt_Listen
PtraceTest.Interrupt_Listen_RequireSeize
PtraceTest.SeizeSetOptions

# PTRACE_GETREGSET / PTRACE_SETREGSET not yet supported.
PtraceTest.GetRegSet
PtraceTest.ChangeRegSetInOptSyscall

# PTRACE_O_TRACE* options not yet supported.
PtraceTest.Trace*
PtraceTest.ExitWhenParentIsNotTracer_Syscall_TraceVfork_TraceVforkDone
TraceExec*
# ...
```

The same suggestion applies to `test/initramfs/src/conformance/ltp/testcases/all.txt:1148` — a one-line comment above `ptrace05` explaining why that one but not `ptrace02-04` / `06-08`.

tatetian (commented):
Here are agent-generated comments with the pr-review skill. Take them with a great of salt. The first comment (significant) should be fixed; the rest are non-blockers.

It is always good to have more tests; it is your call to decide what to do with the test-related comments.

tatetian (approved):

PR #3068, opened 2026-03-27
Fix `readdir` whiteout handling in `overlayfs`

taosue:
Fix incorrect behavior in `overlayfs` `readdir`: during whiteout-only pre-scan non-`.wh` entries were not skipped and could be inserted into the merged view.

Also correct `readdir_at` to return the number of entries visited.

### DIFF
--- a/kernel/src/fs/fs_impls/overlayfs/fs.rs
+++ b/kernel/src/fs/fs_impls/overlayfs/fs.rs
@@ -328,8 +328,10 @@ impl OverlayInode {
             .read_at(offset, writer, status_flags)
     }
 
-    /// Returns the children objects in a unified view.
-    /// The object from the upper layer with the same name will mask the lower ones.
+    /// Visits the children objects in a unified view.
+    ///
+    /// Returns the offset increment needed to advance `offset` to the next unread
+    /// entry. If no entries have been visited, returns an error.
     pub fn readdir_at(&self, offset: usize, visitor: &mut dyn DirentVisitor) -> Result<usize> {
         if self.type_ != InodeType::Dir {
             return_errno!(Errno::ENOTDIR);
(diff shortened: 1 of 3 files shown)


kernel/src/fs/fs_impls/overlayfs/fs.rs
```diff
@@ -328,7 +328,7 @@ impl OverlayInode {
             .read_at(offset, writer, status_flags)
     }
 
-    /// Returns the children objects in a unified view.
+    /// Gets the children objects in a unified view.
```
cchanging: Using `Gets` here may mislead users that this method will return the children inodes. In addition, we prefer to insert a blank line between the summary sentence and the detail discription. You can also checks the other documents in this file.

Suggest:
```rust
/// Visits the directory entries in a unified view.
///
/// Returns the offset increment needed to advance `offset` to the next unread
/// entry. If no entries have been visited, returns an error.
```

kernel/src/fs/fs_impls/overlayfs/fs.rs
```diff
@@ -1467,6 +1486,72 @@ mod tests {
...
+            dir.new_fs_child(".wh.deleted", InodeType::File, mode)
+                .unwrap();
+            // a normal file that should appear exactly once
+            dir.new_fs_child("normal_file", InodeType::File, mode)
+                .unwrap();
+            dir
+        };
+
+        let lower = {
+            let r = Path::new_fs_root(new_dummy_mount());
```
cchanging: Here we can use a better name for `r`?

cchanging (commented):
Thanks for fixing the `readdir` bugs for overlayfs! The fix in this PR makes a lot of sense to me. However, there is currently no regression test for the case where `readdir` reads some entries but returns early because the user-space buffer is too small. 

Since `overlayfs` may undergo significant changes in the future, I think adding a corresponding regression test in this PR would help ensure correctness going forward. You can add an extra test for this case under `regression/fs/overlayfs`.

test/initramfs/src/regression/fs/overlayfs/ovl_test.c
```diff
@@ -185,5 +401,8 @@ int main()
 	printf("Cleaning up\n");
 	clean_up();
 
+	// Run additional tests
+	test_readdir_small_buffer();
```
cchanging, line 405: The changes of kernel source code looks good to me! However, this regression test is not implemented using our C test framework. That may be because the original overlayfs test itself was not written that way, but I would still prefer for tests under the `regression` directory to use a consistent testing framework. 

You could use the facilities provided by [here](https://github.com/asterinas/asterinas/pull/3109#issuecomment-4229416673) to rewrite this test file, while preserving the original test intent and basing it on our test framework.

tatetian (approved):

PR #3074, opened 2026-03-31
Avoid assigning the init process to any process group or session

StevenJiang1110:
The kata-agent requires that the first process (the init process) can call `setsid()` and have it succeed. However, our current implementation assigns a process group (PGID=1) and a session (SID=1) to the init process at creation time. This causes `setsid()` to fail, because a process that is already a process group leader is not permitted to create a new session.

https://github.com/asterinas/asterinas/blob/90607427ced656996fdd5d7423517b539e005670/kernel/src/process/process/mod.rs#L385-L390

On Linux, the init user process inherits `init_task` but is not explicitly assigned a process group or session — both are effectively None until user space takes action. It is only after the init process calls `setsid()` and `ioctl(TIOCSCTTY)` that it acquires a session, a process group, and a controlling terminal. However, strictly following this Linux behavior would break certain invariants in our implementation — for example, it would allow a process to exist without a process group. To avoid this, this PR adopts a pragmatic approach: the init process is placed in a dedicated bootstrap process group(PGID=0) and session(SID=0), distinct from any regular process group or session. So calling setsid will succeed for the init process.

Required by #2911.

### DIFF
--- a/kernel/src/process/process/init_proc.rs
+++ b/kernel/src/process/process/init_proc.rs
@@ -4,7 +4,7 @@
 
 use ostd::{arch::cpu::context::UserContext, task::Task, user::UserContextApi};
 
-use super::Process;
+use super::{Process, Session};
 use crate::{
     fs::{
         thread_info::ThreadFsInfo,
@@ -39,7 +39,14 @@ pub fn spawn_init_process(
         create_default_init_process(argv, envp)?
     };
 
-    set_session_and_group(&process);
+    // Linux starts the init process without placing it in a process group or session.
+    // It joins one only after userspace first calls `setsid()`.
+    //
+    // Asterinas instead requires every process to belong to both a process group and
+    // a session. The init process is therefore placed in a bootstrap process group
+    // and session with PGID and SID set to zero. This preserves the same user-visible
+    // behavior.
+    set_bootstrap_session_and_group(&process);
 
     process.run();
 
@@ -111,12 +118,15 @@ fn create_init_process(
     Ok(init_proc)
 }
 
-fn set_session_and_group(process: &Arc<Process>) {
+fn set_bootstrap_session_and_group(process: &Arc<Process>) {
     // Locking order: PID table -> process group
     let mut pid_table = pid_table::pid_table_mut();
 
-    // Create a new process group and session for the process
-    process.set_new_session(&mut process.process_group.lock(), &mut pid_table);
+    // Add the process to the bootstrap process group and session
+    let (session, process_group) = Session::new_bootstrap_pair(process);
+    pid_table.insert_session(session.sid(), &session);
+    pid_table.insert_process_group(process_group.pgid(), &process_group);
+    *process.process_group.lock() = Some(process_group);
 
     // Add the new process to the global table
     pid_table.insert_process(process.pid(), process);
--- a/kernel/src/process/process/mod.rs
+++ b/kernel/src/process/process/mod.rs
@@ -52,16 +52,23 @@ pub use process_group::ProcessGroup;
 pub use session::Session;
 pub use terminal::Terminal;
 
-/// Process id.
+/// Process ID.
 pub type Pid = u32;
+
 define_atomic_version_of_integer_like_type!(Pid, {
     #[derive(Debug)]
     pub struct AtomicPid(AtomicU32);
 });
-/// Process group id.
+
+/// Process group ID.
 pub type Pgid = u32;
-/// Session Id.
+/// Session ID.
 pub type Sid = u32;
+
+/// The process group ID for process group created by [`ProcessGroup::new_bootstrap`].
+const BOOTSTRAP_PGID: Pgid = 0;
+/// The session ID for the session created by [`Session::new_bootstrap_pair`].
+const BOOTSTRAP_SID: Sid = 0;
 
 pub type ExitCode = u32;
 
--- a/kernel/src/process/process/process_group.rs
+++ b/kernel/src/process/process/process_group.rs
@@ -2,7 +2,7 @@
 
 use alloc::collections::btree_map::Values;
 
-use super::{Pgid, Pid, Process, Session};
+use super::{BOOTSTRAP_PGID, BOOTSTRAP_SID, Pgid, Pid, Process, Session};
 use crate::{prelude::*, process::signal::signals::Signal};
 
 /// A process group.
@@ -27,16 +27,30 @@ impl ProcessGroup {
     ///
     /// The caller needs to ensure that the process does not belong to other process group.
     pub(super) fn new(process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
-        let pid = process.pid();
+        Self::new_with(process.pid(), process, session)
+    }
 
+    /// Creates the bootstrap process group for the init process.
+    ///
+    /// The caller needs to ensure that the process is the init process,
+    /// the process does not belong to other process group, and
+    /// the session is the bootstrap session.
+    pub(super) fn new_bootstrap(process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
+        debug_assert!(process.is_init_process());
+        debug_assert_eq!(session.sid(), BOOTSTRAP_SID);
+
+        Self::new_with(BOOTSTRAP_PGID, process, session)
+    }
+
+    fn new_with(pgid: Pgid, process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
         let inner = {
             let mut processes = BTreeMap::new();
-            processes.insert(pid, Arc::downgrade(process));
+            processes.insert(process.pid(), Arc::downgrade(process));
             Inner { processes }
         };
 
         Arc::new(ProcessGroup {
-            pgid: pid,
+            pgid,
             session,
             inner: Mutex::new(inner),
         })
--- a/kernel/src/process/process/session.rs
+++ b/kernel/src/process/process/session.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use super::{Pgid, Process, ProcessGroup, Sid, Terminal};
+use super::{BOOTSTRAP_SID, Pgid, Process, ProcessGroup, Sid, Terminal};
 use crate::prelude::*;
 
 /// A session.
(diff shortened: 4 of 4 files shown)


kernel/src/process/process/init_proc.rs
```diff
@@ -39,7 +39,11 @@ pub fn spawn_init_process(
         create_default_init_process(argv, envp)?
     };
 
-    set_session_and_group(&process);
+    // FIXME: Linux starts the first userspace process without session/group.
```
lrh2000: Will this cause a difference that users will notice? If not, then I don't think it's a FIXME or workaround. It's just an implementation detail of Asterinas and Linux. As long as the user space doesn't complain, it doesn't matter whether the init process belongs to a bootstrap session or no session.

kernel/src/process/process/init_proc.rs
```diff
@@ -39,7 +39,11 @@ pub fn spawn_init_process(
...
+    // The process will join a process group/session until the user first time calls setsid.
+    // However, this will break many existing code that assumes a live process always belongs
+    // to a process group/session. So we currently add the init process to a special boot strap process group session.
```
lrh2000: Break long lines.

`setsid` -> `` `setsid` ``

kernel/src/process/process/init_proc.rs
```diff
@@ -111,12 +115,15 @@ fn create_init_process(
...
-fn set_session_and_group(process: &Arc<Process>) {
+fn set_bootstrap_session_and_group(process: &Arc<Process>) {
     // Locking order: PID table -> process group
     let mut pid_table = pid_table::pid_table_mut();
 
-    // Create a new process group and session for the process
-    process.set_new_session(&mut process.process_group.lock(), &mut pid_table);
+    let (session, process_group) = Session::new_bootstrap_pair(process);
+
+    pid_table.insert_session(session.sid(), &session);
```
lrh2000, line 127:
```rust
    // Add the process to the bootstrap process group and session
```

kernel/src/process/process/mod.rs
```diff
@@ -62,6 +62,10 @@ define_atomic_version_of_integer_like_type!(Pid, {
 pub type Pgid = u32;
 /// Session Id.
 pub type Sid = u32;
+/// The process group ID inherited by the first userspace process.
+pub const BOOTSTRAP_PGID: Pgid = 0;
+/// The session ID inherited by the first userspace process.
```
lrh2000:
```
> grep -R "first userspace" kernel | wc -l
0
> grep -R "init process" kernel | wc -l
25
```

I am not sure why we want to introduce a new concept, i.e., `the first userspace process`.

---

Also, I'm not sure why the ID is inherited. Inherited from what? In Asterinas, I don't think the init process inherits from anything. It's simply set when the init process spawns. There is nothing for the process to inherit.

kernel/src/process/process/mod.rs
```diff
@@ -71,11 +75,29 @@ pub(super) fn init_on_each_cpu() {
 
 pub(super) fn init_in_first_process(ctx: &Context) {
     // FIXME: This should be done by the userspace init process.
```
lrh2000: Is this FIXME still correct?

Actually, if the init process in Linux does not have a session, the logic of this method seems meaningless. How can a session that doesn't exist become the controlling session of a terminal?

kernel/src/process/process/mod.rs
```diff
@@ -71,11 +75,29 @@ pub(super) fn init_on_each_cpu() {
...
-    (crate::device::tty::SystemConsole::singleton()
+    // The pattern used here is a workaround and is generally discouraged.
+    // We should call `Terminal::set_control()` instead,
+    // but that method requires the process to be the session leader.
+    // The init process, however, is not the leader of the bootstrap session,
+    // so we directly modify the internal state of the session and terminal here
+    // as a temporary workaround.
```
lrh2000: Introduce a new method in `Terminal` to bypass the check instead?

kernel/src/process/process/process_group.rs
```diff
@@ -27,16 +30,31 @@ impl ProcessGroup {
     ///
     /// The caller needs to ensure that the process does not belong to other process group.
     pub(super) fn new(process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
-        let pid = process.pid();
+        Self::new_with(process.pid(), process, session)
+    }
+
+    /// Creates the bootstrap process group for the first userspace process.
+    ///
+    /// The caller needs to ensure that the process does not belong to other process group.
```
lrh2000: "the process is the init process and does not ..."

kernel/src/process/process/session.rs
```diff
@@ -33,20 +33,36 @@ impl Session {
...
-        });
-
+        let session = Self::new_empty(process.pid());
         let process_group = ProcessGroup::new(process, session.clone());
         session.lock().insert_process_group(&process_group);
 
         (session, process_group)
     }
 
+    /// Creates the bootstrap session/process-group pair for the first userspace process.
```
lrh2000: Keep `new_pair`'s comments? "ensure that the process is the init process and ..."

kernel/src/process/process/mod.rs
```diff
@@ -62,6 +62,10 @@ define_atomic_version_of_integer_like_type!(Pid, {
 pub type Pgid = u32;
 /// Session Id.
 pub type Sid = u32;
```
lrh2000, line 66: Could you fix these comments as well?
```rust
/// Process ID.
pub type Pid = u32;

define_atomic_version_of_integer_like_type!(Pid, {
    #[derive(Debug)]
    pub struct AtomicPid(AtomicU32);
});

/// Process group ID.
pub type Pgid = u32;
/// Session ID.
pub type Sid = u32;

/// The process group ID inherited by the first userspace process.
pub const BOOTSTRAP_PGID: Pgid = 0;
/// The session ID inherited by the first userspace process. 
pub const BOOTSTRAP_SID: Sid = 0;Expand comment
```

lrh2000 (commented):
Thanks for fixing this behavior!

So what will `getsid`/`getpgrp` return before the init process belongs to a process group / session? I guess it will return zeros?

`setsid`/`setpgid` cannot join the bootstrap process group / session, right?

There is perhaps no need to leave FIXME for implementation details. Instead, let's focus on whether doing so will result in differences visible to users.

StevenJiang1110:
> So what will getsid/getpgrp return before the init process belongs to a process group / session? I guess it will return zeros?

Yes — I verified this experimentally by running kata-agent on Linux. Before the first setsid() call, both getpgid(0/1) and getsid() return 0, while getpid() returns 1. After setsid() is called, all three — getpid(), getpgid(0/1), and getsid() — return 1.

> setsid/setpgid cannot join the bootstrap process group / session, right?

Yes. For `setsid()`, it always creates a new session whose `SID` equals the calling process's `PID`, and since `PID`s are always at least 1, no process can ever end up in the bootstrap session (`SID=0`). For setpgid(), a `PGID` of `0` is intercepted at the syscall level and reinterpreted as the calling process's own `PID`, which is never 0 — so no process can join the bootstrap process group (`PGID=0`) through `setpgid()` either.

https://github.com/asterinas/asterinas/blob/38575e8fd61d9b0b39fc274858c89b611049f3f0/kernel/src/syscall/setpgid.rs#L23

kernel/src/process/process/mod.rs
StevenJiang1110: Yes, in theory we should remove it.

However, our current init process (`/bin/sh`) never calls `setsid()`, but does call `ioctl(TIOCGPGRP)` to query the foreground process group directly. If the init process has no controlling terminal, this ioctl will fail, resulting in the message `sh: can't access tty; job control turned off` when running `make run_kernel`. While this error does not appear to affect subsequent program execution, I am inclined to keep the current logic for now to avoid the noise.

kernel/src/process/process/init_proc.rs
StevenJiang1110: I cannot think up of the differences yet. So I reword the comment here to avoid FIXME and workaround.

kernel/src/process/process/mod.rs
lrh2000: > sh: can't access tty; job control turned off

Linux has exactly the same thing. So it should be expected.

> While this error does not appear to affect subsequent program execution,

This is not true. It means Ctrl+C won't work.

> I am inclined to keep the current logic for now to avoid the noise.

Let's see if https://github.com/asterinas/asterinas/pull/3077 works.

kernel/src/process/process/init_proc.rs
```diff
@@ -39,7 +39,12 @@ pub fn spawn_init_process(
         create_default_init_process(argv, envp)?
     };
 
-    set_session_and_group(&process);
+    // Linux starts the init process without assigning it to a process group or session.
+    // It joins a process group and session only when userspace first calls `setsid()`.
+    // That behavior breaks existing code that assumes every live process always belongs
+    // to both a process group and a session.
+    // For now, we place the init process in a special bootstrap process group and session.
```
lrh2000:
```rust
  // Linux starts the init process without placing it in a process group or session.
  // It joins one only after userspace first calls `setsid()`.
  //
  // Asterinas instead requires every process to belong to both a process group and
  // a session. The init process is therefore placed in a bootstrap process group
  // and session with PGID and SID set to zero. This preserves the same user-visible
  // behavior.
```

kernel/src/process/process/process_group.rs
```diff
@@ -2,8 +2,11 @@
 
 use alloc::collections::btree_map::Values;
 
-use super::{Pgid, Pid, Process, Session};
-use crate::{prelude::*, process::signal::signals::Signal};
+use super::{BOOTSTRAP_PGID, Pgid, Pid, Process, Session};
+use crate::{
+    prelude::*,
+    process::{process::BOOTSTRAP_SID, signal::signals::Signal},
```
lrh2000: I am not sure why you import `BOOTSTRAP_SID` and `BOOTSTRAP_PGID` in different ways. They locate in the same module. Use `super::BOOTSTRAP_SID` instead.

kernel/src/process/process/process_group.rs
```diff
@@ -27,16 +30,32 @@ impl ProcessGroup {
     ///
     /// The caller needs to ensure that the process does not belong to other process group.
     pub(super) fn new(process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
-        let pid = process.pid();
+        Self::new_with(process.pid(), process, session)
+    }
+
+    /// Creates the bootstrap process group for the first userspace process.
```
lrh2000:
```suggestion
    /// Creates the bootstrap process group for the init process.
```

kernel/src/process/process/process_group.rs
```diff
@@ -27,16 +30,32 @@ impl ProcessGroup {
...
+    ///
+    /// The caller needs to ensure that the process is the init process
+    /// and does not belong to other process group.
+    pub(in crate::process) fn new_bootstrap(
```
lrh2000: `pub(super)` should be enough?

kernel/src/process/process/mod.rs
```diff
@@ -52,17 +52,24 @@ pub use process_group::ProcessGroup;
...
+
 define_atomic_version_of_integer_like_type!(Pid, {
     #[derive(Debug)]
     pub struct AtomicPid(AtomicU32);
 });
-/// Process group id.
+
+/// Process group ID.
 pub type Pgid = u32;
 /// Session Id.
```
lrh2000:
```suggestion
/// Session ID.
```

kernel/src/process/process/mod.rs
```diff
@@ -52,17 +52,24 @@ pub use process_group::ProcessGroup;
...
 pub type Sid = u32;
 
+/// The process group ID for the bootstrap process group.
+pub const BOOTSTRAP_PGID: Pgid = 0;
+/// The session ID for the bootstrap session.
+pub const BOOTSTRAP_SID: Sid = 0;
```
lrh2000: No need to be `pub`.

Also it is somehow confusing because there is no explanation of what a bootstrap process group is when a reader first encounters these lines... Will `` for the session created by [`Session::new_bootstrap`] `` sightly better?

kernel/src/process/process/process_group.rs
```diff
@@ -27,16 +30,32 @@ impl ProcessGroup {
...
+        process: &Arc<Process>,
+        session: Arc<Session>,
+    ) -> Arc<Self> {
+        debug_assert_eq!(process.pid(), 1);
```
lrh2000: I assume similar assertions should use `process.is_init_process()` now.

kernel/src/process/process/process_group.rs
```diff
@@ -27,16 +27,29 @@ impl ProcessGroup {
...
     /// The caller needs to ensure that the process does not belong to other process group.
     pub(super) fn new(process: &Arc<Process>, session: Arc<Session>) -> Arc<Self> {
-        let pid = process.pid();
+        Self::new_with(process.pid(), process, session)
+    }
+
+    /// Creates the bootstrap process group for the init process.
+    ///
+    /// The caller needs to ensure that the process is the init process
+    /// and does not belong to other process group.
```
lrh2000: Find a way to add "the session is the bootstrap session". For example,
```rust
    /// The caller needs to ensure that the process is the init process,
    /// the process does not belong to other process group, and
    /// the session is the bootstrap session.
```

lrh2000:
@StevenJiang1110 Rebase this PR since #3077 has been merged?

StevenJiang1110:
> @StevenJiang1110 Rebase this PR since #3077 has been merged?

The PR is rebased now.

lrh2000 (approved):

PR #3078, opened 2026-04-02
Implement initial LSM framework and migrate Yama

zzj-5341:
## Summary

This PR introduces the first stage of Linux-style LSM support in Asterinas.

It establishes a small built-in LSM framework and migrates `Yama` onto it, so ptrace-style task access checks no longer rely on subsystem-local policy logic.

The resulting control flow is:

`alien_access` -> `security::ptrace_access_check()` -> `security::lsm` -> `Yama`

This PR is intentionally scoped to one real control path so that the framework direction can be reviewed before expanding the hook surface.

## Changes

- Add a new `security::lsm` framework under `kernel/src/security/lsm/`
- Define a built-in `LsmModule` trait for stackable LSM modules
- Add `PtraceAccessContext` and related ptrace-style access types
- Add ordered LSM dispatch through the `security` facade
- Migrate `Yama` into the LSM framework as the first built-in minor LSM
- Route the existing alien-access path through `security::ptrace_access_check()`
- Expose `/proc/sys/kernel/yama/ptrace_scope` through the new Yama module implementation
- Add an initramfs regression test for `pidfd_getfd` behavior under different `ptrace_scope` settings

## Why

Security-sensitive checks in Asterinas are still mostly subsystem-specific today.

This PR is the first step toward a Linux-style security structure where:

- subsystem code describes the operation through a common `security` facade
- security decisions are dispatched through one framework
- policy logic lives in modules instead of being embedded directly in subsystem code

`Yama` is a good first migration target because it already has a narrow scope and naturally maps to a minor LSM design.

## Testing

I ran the following checks:

- `cargo fmt --check` on the touched Rust files
- `cargo check -p aster-kernel --target x86_64-unknown-none --no-default-features`
- `make` in `test/initramfs/src/apps/security/yama`

I also added a regression test:

- `test/initramfs/src/apps/security/yama/pidfd_getfd_scope.c`

This test covers:

- reading and writing `/proc/sys/kernel/yama/ptrace_scope`
- rejecting sibling `pidfd_getfd` access under restrictive settings
- allowing access when Yama is disabled

## Non-goals

This PR does not yet introduce:

- file or exec policy hooks
- object labels
- `security.aster.*` xattr-based policy
- socket, mount, namespace, or mmap-related hooks
- a major LSM

These are planned as follow-up steps after the initial framework direction is reviewed.

## Follow-up work

Planned next steps include:

- adding file- and exec-related hooks
- introducing inode-backed policy state
- adding task labels and exec label propagation
- expanding the hook surface toward a more complete Linux-style LSM design

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/mod.rs
@@ -148,8 +148,7 @@ impl RootDirOps {
         ProcDir::new_root(Self, fs, PROC_ROOT_INO, sb, mkmod!(a+rx))
     }
 
-    #[expect(clippy::type_complexity)]
-    const STATIC_ENTRIES: &'static [StaticDirEntry<fn(Weak<dyn Inode>) -> Arc<dyn Inode>>] = &[
+    const STATIC_ENTRIES: &'static [StaticEntry] = &[
         ("cmdline", InodeType::File, CmdLineFileOps::new_inode),
         ("cpuinfo", InodeType::File, CpuInfoFileOps::new_inode),
         (
@@ -273,3 +272,6 @@ impl DirOps for RootDirOps {
         pid_entry.is_none_or(|pid_entry| pid_entry.type_().is_none())
     }
 }
+
+type StaticEntry = StaticDirEntry<fn(Weak<dyn Inode>) -> Arc<dyn Inode>>;
+type StaticEntryWithOps<T> = StaticDirEntry<fn(&T, Weak<dyn Inode>) -> Arc<dyn Inode>>;
--- a/kernel/src/fs/fs_impls/procfs/sys/kernel/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/sys/kernel/mod.rs
@@ -4,18 +4,19 @@ use crate::{
     fs::{
         file::{InodeType, mkmod},
         procfs::{
-            ProcDir,
+            ProcDir, StaticEntry,
             sys::kernel::{
                 cap_last_cap::CapLastCapFileOps, pid_max::PidMaxFileOps, yama::YamaDirOps,
             },
             template::{
-                DirOps, ReaddirEntry, StaticDirEntry, listed_entries_from_table,
+                DirOps, ListedEntry, ReaddirEntry, listed_entries_from_table,
                 lookup_child_from_table, visit_listed_entries,
             },
         },
         vfs::inode::Inode,
     },
     prelude::*,
+    security::lsm::is_yama_enabled,
 };
 
 mod cap_last_cap;
@@ -33,15 +34,13 @@ impl KernelDirOps {
         ProcDir::new(Self, parent, mkmod!(a+rx))
     }
 
-    #[expect(clippy::type_complexity)]
-    const STATIC_ENTRIES: &'static [StaticDirEntry<fn(Weak<dyn Inode>) -> Arc<dyn Inode>>] = &[
+    const STATIC_ENTRIES: &'static [StaticEntry] = &[
         (
             "cap_last_cap",
             InodeType::File,
             CapLastCapFileOps::new_inode,
         ),
         ("pid_max", InodeType::File, PidMaxFileOps::new_inode),
-        ("yama", InodeType::Dir, YamaDirOps::new_inode),
     ];
 }
 
@@ -53,6 +52,10 @@ impl DirOps for KernelDirOps {
             return Ok(child);
         }
 
+        if name == "yama" && is_yama_enabled() {
+            return Ok(YamaDirOps::new_inode(this_dir.this_weak().clone()));
+        }
+
         return_errno_with_message!(Errno::ENOENT, "the file does not exist");
     }
 
@@ -60,9 +63,11 @@ impl DirOps for KernelDirOps {
     where
         F: FnMut(ReaddirEntry<'a>) -> Result<()>,
     {
+        let yama_entry = is_yama_enabled().then(|| ListedEntry::new("yama", InodeType::Dir));
+
         visit_listed_entries(
             offset,
-            listed_entries_from_table(Self::STATIC_ENTRIES),
+            listed_entries_from_table(Self::STATIC_ENTRIES).chain(yama_entry),
             visit_fn,
         )
     }
--- a/kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
+++ b/kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
@@ -13,10 +13,10 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
-    process::posix_thread::alien_access::yama::{YamaScope, get_yama_scope, set_yama_scope},
+    security::lsm::{YamaScope, get_yama_scope, set_yama_scope},
 };
 
-/// Represents the inode at `/proc/sys/kernel/yama`.
+/// Directory operations for `/proc/sys/kernel/yama`.
 pub struct YamaDirOps;
 
 impl YamaDirOps {
@@ -58,7 +58,7 @@ impl DirOps for YamaDirOps {
     }
 }
 
-/// Represents the inode at `/proc/sys/kernel/yama/ptrace_scope`.
+/// File operations for `/proc/sys/kernel/yama/ptrace_scope`.
 struct PtraceScopeFileOps;
 
 impl PtraceScopeFileOps {
--- a/kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
+++ b/kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
@@ -5,18 +5,20 @@ use aster_util::printer::VmPrinter;
 use crate::{
     fs::{
         file::{InodeType, mkmod},
-        procfs::template::{
-            DirOps, FileOps, ProcDir, ProcFile, ReaddirEntry, StaticDirEntry,
-            listed_entries_from_table, lookup_child_from_table, read_i32_from,
-            visit_listed_entries,
+        procfs::{
+            StaticEntry,
+            template::{
+                DirOps, FileOps, ProcDir, ProcFile, ReaddirEntry, listed_entries_from_table,
+                lookup_child_from_table, read_i32_from, visit_listed_entries,
+            },
         },
         vfs::inode::Inode,
     },
     prelude::*,
-    security::lsm::{YamaScope, get_yama_scope, set_yama_scope},
+    security::lsm::yama,
 };
 
-/// Directory operations for `/proc/sys/kernel/yama`.
+/// Represents the inode at `/proc/sys/kernel/yama`.
 pub struct YamaDirOps;
 
 impl YamaDirOps {
@@ -27,8 +29,7 @@ impl YamaDirOps {
         ProcDir::new(Self, parent, mkmod!(a+rx))
     }
 
-    #[expect(clippy::type_complexity)]
-    const STATIC_ENTRIES: &'static [StaticDirEntry<fn(Weak<dyn Inode>) -> Arc<dyn Inode>>] = &[(
+    const STATIC_ENTRIES: &'static [StaticEntry] = &[(
         "ptrace_scope",
         InodeType::File,
         PtraceScopeFileOps::new_inode,
@@ -58,7 +59,7 @@ impl DirOps for YamaDirOps {
     }
 }
 
-/// File operations for `/proc/sys/kernel/yama/ptrace_scope`.
+/// Represents the inode at `/proc/sys/kernel/yama/ptrace_scope`.
 struct PtraceScopeFileOps;
 
 impl PtraceScopeFileOps {
@@ -72,16 +73,16 @@ impl FileOps for PtraceScopeFileOps {
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let mut printer = VmPrinter::new_skip(writer, offset);
 
-        writeln!(printer, "{}", get_yama_scope() as i32)?;
+        writeln!(printer, "{}", yama::get_scope() as i32)?;
 
         Ok(printer.bytes_written())
     }
 
     fn write_at(&self, _offset: usize, reader: &mut VmReader) -> Result<usize> {
         let (val, read_bytes) = read_i32_from(reader)?;
-        let new_scope = YamaScope::try_from(val)?;
+        let new_scope = yama::YamaScope::try_from(val)?;
 
-        set_yama_scope(new_scope)?;
+        yama::set_scope(new_scope)?;
 
         Ok(read_bytes)
     }
--- a/kernel/src/process/posix_thread/alien_access.rs
+++ b/kernel/src/process/posix_thread/alien_access.rs
@@ -4,11 +4,10 @@
 //!
 //! An alien thread is one outside the current thread's thread group (the process).
 
-use bitflags::bitflags;
-
 use crate::{
     prelude::*,
     process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::lsm::{self, AlienAccessContext},
 };
 
 impl PosixThread {
@@ -27,7 +26,7 @@ impl PosixThread {
         }
 
         let cred = accessor.credentials();
-        let (caller_uid, caller_gid) = if mode.1 == CredsSource::FsCreds {
+        let (caller_uid, caller_gid) = if mode.creds() == CredsSource::FsCreds {
             (cred.fsuid(), cred.fsgid())
         } else {
             (cred.ruid(), cred.rgid())
@@ -54,132 +53,59 @@ impl PosixThread {
             );
         }
 
-        yama::check_alien_access(accessor, self, mode, caller_has_cap)?;
+        lsm::alien_access_check(&AlienAccessContext::new(
+            accessor,
+            self,
+            mode,
+            caller_has_cap,
+        ))?;
 
         Ok(())
     }
 }
 
-/// The mode used by the alien access permission check.
-pub struct AlienAccessMode(AlienAccessFlags, CredsSource);
-
-impl AlienAccessMode {
-    /// Read-only alien access check, using real credentials (`ruid`/`rgid`).
-    #[expect(dead_code)]
-    pub const READ_WITH_REAL_CREDS: Self = Self(AlienAccessFlags::READ, CredsSource::RealCreds);
-    /// Attach-level alien access check, using real credentials (`ruid`/`rgid`).
-    pub const ATTACH_WITH_REAL_CREDS: Self = Self(AlienAccessFlags::ATTACH, CredsSource::RealCreds);
-    /// Read-only alien access check, using filesystem credentials (`fsuid`/`fsgid`).
-    pub const READ_WITH_FS_CREDS: Self = Self(AlienAccessFlags::READ, CredsSource::FsCreds);
-    /// Attach-level alien access check, using filesystem credentials (`fsuid`/`fsgid`).
-    pub const ATTACH_WITH_FS_CREDS: Self = Self(AlienAccessFlags::ATTACH, CredsSource::FsCreds);
-}
-
-bitflags! {
-    /// Access strength in the alien access permission check.
-    struct AlienAccessFlags: u32 {
-        const READ       = 0x01;
-        const ATTACH     = 0x02;
-    }
-}
-
-/// The credentials used in the alien access permission check.
-#[derive(PartialEq)]
-enum CredsSource {
+/// The credentials used by an alien access check.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum CredsSource {
     FsCreds,
     RealCreds,
 }
 
-// TODO: Rewrite Yama as a minor LSM module when Asterinas introduces the LSM subsystem.
-pub mod yama {
-    use core::sync::atomic::{AtomicI32, Ordering};
+/// The strength of an alien access check.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum AlienAccessKind {
+    Read,
+    Attach,
+}
 
-    use super::*;
-    use crate::process::{Process, UserNamespace, posix_thread::AsPosixThread};
+/// An alien access check mode.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct AlienAccessMode {
+    kind: AlienAccessKind,
+    creds: CredsSource,
+}
 
-    /// Returns the current Yama scope for alien access.
-    pub fn get_yama_scope() -> YamaScope {
-        YAMA_SCOPE.load(Ordering::Relaxed).try_into().unwrap()
+impl AlienAccessMode {
+    /// Read-only alien access check using real credentials.
+    #[expect(dead_code)]
+    pub const READ_WITH_REAL_CREDS: Self = Self::new(AlienAccessKind::Read, CredsSource::RealCreds);
+    /// Attach-level alien access check using real credentials.
+    pub const ATTACH_WITH_REAL_CREDS: Self =
+        Self::new(AlienAccessKind::Attach, CredsSource::RealCreds);
+    /// Read-only alien access check using filesystem credentials.
+    pub const READ_WITH_FS_CREDS: Self = Self::new(AlienAccessKind::Read, CredsSource::FsCreds);
+    /// Attach-level alien access check using filesystem credentials.
+    pub const ATTACH_WITH_FS_CREDS: Self = Self::new(AlienAccessKind::Attach, CredsSource::FsCreds);
+
+    pub const fn new(kind: AlienAccessKind, creds: CredsSource) -> Self {
+        Self { kind, creds }
     }
 
-    /// Sets the Yama scope for alien access.
-    pub fn set_yama_scope(new_scope: YamaScope) -> Result<()> {
-        UserNamespace::get_init_singleton().check_cap(
-            CapSet::SYS_PTRACE,
-            current_thread!().as_posix_thread().unwrap(),
-        )?;
-
-        YAMA_SCOPE
-            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current_scope| {
-                let is_downgrading_from_no_attach =
-                    current_scope == YamaScope::NoAttach as i32 && new_scope != YamaScope::NoAttach;
-                (!is_downgrading_from_no_attach).then_some(new_scope as i32)
-            })
-            .map_err(|_| {
-                Error::with_message(
-                    Errno::EINVAL,
-                    "`YamaScope::NoAttach` cannot be changed once set",
-                )
-            })?;
-
-        Ok(())
+    pub const fn kind(self) -> AlienAccessKind {
+        self.kind
     }
 
-    static YAMA_SCOPE: AtomicI32 = AtomicI32::new(YamaScope::Relational as i32);
-
-    /// The Yama scope levels.
-    #[repr(i32)]
-    #[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromInt)]
-    pub enum YamaScope {
-        /// No additional restrictions on alien attach.
-        Disabled = 0,
-        /// Only allow alien attach by ancestor processes, or processes with `CapSet::SYS_PTRACE`.
-        Relational = 1,
-        /// Only allow alien attach by processes with `CapSet::SYS_PTRACE`.
-        Capability = 2,
-        /// Disallow any alien attach.
-        NoAttach = 3,
-    }
-
-    pub(super) fn check_alien_access(
-        accessor: &PosixThread,
-        target: &PosixThread,
-        mode: AlienAccessMode,
-        caller_has_cap: bool,
-    ) -> Result<()> {
-        if !mode.0.contains(AlienAccessFlags::ATTACH) {
-            return Ok(());
-        }
-
-        let is_denied = match get_yama_scope() {
-            YamaScope::Disabled => false,
-            YamaScope::Relational => {
-                !caller_has_cap && !is_ancestor_of(accessor.weak_process(), target.process())
-            }
-            YamaScope::Capability => !caller_has_cap,
-            YamaScope::NoAttach => true,
-        };
-
-        if is_denied {
-            return_errno_with_message!(Errno::EPERM, "alien access is denied due to Yama scope");
-        }
-
-        Ok(())
-    }
-
-    fn is_ancestor_of(ancestor: &Weak<Process>, descendant: Arc<Process>) -> bool {
-        let mut current = descendant;
-        loop {
-            let parent_guard = current.parent().lock();
-            let parent = parent_guard.process();
-            if Weak::ptr_eq(parent, ancestor) {
-                return true;
-            }
-            let Some(parent) = parent.upgrade() else {
-                return false;
-            };
-            drop(parent_guard);
-            current = parent;
-        }
+    pub const fn creds(self) -> CredsSource {
+        self.creds
     }
 }
--- a/kernel/src/process/posix_thread/alien_access.rs
+++ b/kernel/src/process/posix_thread/alien_access.rs
@@ -7,7 +7,7 @@
 use crate::{
     prelude::*,
     process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
-    security::lsm::{self, AlienAccessContext},
+    security::lsm::{AlienAccessContext, hooks as lsm_hooks},
 };
 
 impl PosixThread {
@@ -53,7 +53,7 @@ impl PosixThread {
             );
         }
 
-        lsm::alien_access_check(&AlienAccessContext::new(
+        lsm_hooks::on_alien_access(&AlienAccessContext::new(
             accessor,
             self,
             mode,
--- a/kernel/src/security/lsm/hooks/alien_access.rs
+++ b/kernel/src/security/lsm/hooks/alien_access.rs
@@ -12,12 +12,13 @@ use crate::{
     process::posix_thread::{PosixThread, alien_access::AlienAccessMode},
 };
 
-/// Defines hooks for alien access checks.
-pub trait LsmAlienAccessCheck: Sync {
-    /// Checks whether the accessor may inspect or attach to the target.
-    fn alien_access_check(&self, _context: &AlienAccessContext) -> Result<()> {
-        Ok(())
+/// Runs alien access hooks in module order.
+pub fn on_alien_access(context: &AlienAccessContext) -> Result<()> {
+    for module in modules::active_modules() {
+        module.on_alien_access(context)?;
     }
+
+    Ok(())
 }
 
 /// The inputs for an alien access check through the LSM stack.
@@ -29,6 +30,7 @@ pub struct AlienAccessContext<'a> {
 }
 
 impl<'a> AlienAccessContext<'a> {
+    /// Creates an alien access context.
     pub const fn new(
         accessor: &'a PosixThread,
         target: &'a PosixThread,
@@ -43,28 +45,23 @@ impl<'a> AlienAccessContext<'a> {
         }
     }
 
+    /// Returns the thread requesting access.
     pub const fn accessor(&self) -> &PosixThread {
         self.accessor
     }
 
+    /// Returns the thread being accessed.
     pub const fn target(&self) -> &PosixThread {
         self.target
     }
 
+    /// Returns the requested access mode.
     pub const fn mode(&self) -> AlienAccessMode {
         self.mode
     }
 
+    /// Returns whether the accessor has `CapSet::SYS_PTRACE`.
     pub const fn accessor_has_cap_sys_ptrace(&self) -> bool {
         self.accessor_has_cap_sys_ptrace
     }
-}
-
-/// Runs alien access hooks in module order.
-pub fn alien_access_check(context: &AlienAccessContext) -> Result<()> {
-    for module in modules::active_modules() {
-        module.alien_access_check(context)?;
-    }
-
-    Ok(())
 }
--- a/kernel/src/security/lsm/mod.rs
+++ b/kernel/src/security/lsm/mod.rs
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! The Linux Security Module (LSM) framework.
+//!
+//! LSM lets the kernel route security-sensitive operations through a stack of
+//! built-in policy modules. Each module can implement shared hook traits and
+//! inspect common hook contexts before allowing or rejecting an operation.
+//!
+//! This module defines the common LSM traits, alien access hook contexts, and
+//! dispatch helpers shared by built-in modules such as `yama`.
+
+mod checks;
+mod modules;
+
+pub use self::{
+    checks::alien_access::{AlienAccessContext, LsmAlienAccessCheck, alien_access_check},
+    modules::yama::{YamaScope, get_yama_scope, set_yama_scope},
+};
+use crate::prelude::*;
+
+/// The kind of an LSM module.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LsmKind {
+    Minor,
+    #[expect(dead_code)]
+    Major,
+}
+
+impl LsmKind {
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::Minor => "minor",
+            Self::Major => "major",
+        }
+    }
+}
+
+/// Defines the common interface for built-in LSM modules.
+pub trait LsmModule: LsmAlienAccessCheck + Sync {
+    /// Returns the module name.
+    fn name(&self) -> &'static str;
+
+    /// Returns the module kind.
+    fn kind(&self) -> LsmKind;
+
+    /// Initializes the module during boot.
+    fn init(&self);
+}
+
+pub(super) fn init() {
+    for module in modules::active_modules() {
+        info!(
+            "[kernel] LSM module enabled: {} ({})",
+            module.name(),
+            module.kind().as_str()
+        );
+        module.init();
+    }
+}
--- a/kernel/src/security/lsm/mod.rs
+++ b/kernel/src/security/lsm/mod.rs
@@ -6,54 +6,49 @@
 //! built-in policy modules. Each module can implement shared hook traits and
 //! inspect common hook contexts before allowing or rejecting an operation.
 //!
-//! This module defines the common LSM traits, alien access hook contexts, and
-//! dispatch helpers shared by built-in modules such as `yama`.
+//! This module defines the common LSM traits and hook contexts shared by
+//! built-in modules such as `yama`. Module selection follows the `lsm=` and
+//! legacy `security=` kernel command-line parameters.
 
-mod checks;
+pub mod hooks;
 mod modules;
 
-pub use self::{
-    checks::alien_access::{AlienAccessContext, LsmAlienAccessCheck, alien_access_check},
-    modules::yama::{YamaScope, get_yama_scope, set_yama_scope},
-};
-use crate::prelude::*;
-
-/// The kind of an LSM module.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum LsmKind {
-    Minor,
-    #[expect(dead_code)]
-    Major,
+pub mod yama {
+    pub use super::modules::yama::{YamaScope, get_scope, set_scope};
 }
 
-impl LsmKind {
-    const fn as_str(self) -> &'static str {
-        match self {
-            Self::Minor => "minor",
-            Self::Major => "major",
-        }
+pub use self::hooks::AlienAccessContext;
+use self::hooks::LsmAlienAccessHook;
+use crate::prelude::*;
+
+bitflags! {
+    /// LSM module flags.
+    pub struct LsmFlags: u32 {
+        /// Marks a module as selectable through the legacy `security=` parameter.
+        const LEGACY_MAJOR = 1 << 0;
+        /// Marks a module as mutually exclusive with other exclusive modules.
+        const EXCLUSIVE = 1 << 1;
     }
 }
 
 /// Defines the common interface for built-in LSM modules.
-pub trait LsmModule: LsmAlienAccessCheck + Sync {
+trait LsmModule: LsmAlienAccessHook + Sync {
     /// Returns the module name.
     fn name(&self) -> &'static str;
 
-    /// Returns the module kind.
-    fn kind(&self) -> LsmKind;
+    /// Returns the module flags.
+    fn flags(&self) -> LsmFlags;
+}
 
-    /// Initializes the module during boot.
-    fn init(&self);
+/// Returns whether the Yama LSM is enabled.
+pub fn is_yama_enabled() -> bool {
+    modules::active_modules()
+        .iter()
+        .any(|module| module.name() == "yama")
 }
 
 pub(super) fn init() {
     for module in modules::active_modules() {
-        info!(
-            "[kernel] LSM module enabled: {} ({})",
-            module.name(),
-            module.kind().as_str()
-        );
-        module.init();
+        info!("[kernel] LSM module enabled: {}", module.name());
     }
 }
--- a/kernel/src/security/lsm/modules/mod.rs
+++ b/kernel/src/security/lsm/modules/mod.rs
@@ -1,11 +1,166 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! Built-in LSM module registration and boot-time selection.
+//!
+//! Linux provides two kernel command-line parameters for choosing enabled LSM
+//! modules: `lsm=` is the modern ordered module list, while `security=` is the
+//! legacy selector for one major LSM.
+//!
+//! When `lsm=` is specified, its comma-separated module names are used as the
+//! enabled LSM stack in that order. Unknown module names are ignored with a
+//! warning, and duplicate names are ignored after their first selection.
+//!
+//! Exclusive modules listed in `lsm=` are also processed in order: the first
+//! selected exclusive module is kept, and later exclusive entries are ignored.
+//!
+//! When only `security=` is specified, the default LSM stack is selected first.
+//! The named module is then selected if it is a legacy major LSM; if it is also
+//! exclusive, it replaces the currently selected exclusive module.
+//!
+//! This means `security=` applies after the default stack, while `lsm=` fully
+//! controls the explicit module order.
+//!
+//! If both parameters are specified, `security=` is ignored because `lsm=`
+//! fully describes the enabled stack. If neither parameter is specified, the
+//! default LSM stack is used.
+
 pub mod yama;
 
-use super::LsmModule;
+use spin::Once;
 
-static ACTIVE_MODULES: [&dyn LsmModule; 1] = [&yama::YAMA_LSM];
+use super::{LsmFlags, LsmModule};
+use crate::prelude::*;
+
+static LSM_PARAM: Once<String> = Once::new();
+static LEGACY_SECURITY_PARAM: Once<String> = Once::new();
+
+aster_cmdline::define_kv_param!("lsm", LSM_PARAM);
+aster_cmdline::define_kv_param!("security", LEGACY_SECURITY_PARAM);
+
+/// All LSM modules compiled into the kernel.
+static ALL_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
+
+/// The fallback LSM stack used when no boot-time selector is specified.
+pub(super) static DEFAULT_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
+
+static ALL_MODULES_BY_NAME: Once<BTreeMap<&'static str, &'static dyn LsmModule>> = Once::new();
+static ACTIVE_MODULES: Once<Box<[&'static dyn LsmModule]>> = Once::new();
 
 pub(super) fn active_modules() -> &'static [&'static dyn LsmModule] {
-    &ACTIVE_MODULES
+    ACTIVE_MODULES
+        .call_once(|| {
+            let modules_by_name = ALL_MODULES_BY_NAME.call_once(|| {
+                ALL_MODULES
+                    .iter()
+                    .map(|module| (module.name(), *module))
+                    .collect()
+            });
+            ModuleSelection::select_from_boot_params(modules_by_name).into_modules()
+        })
+        .as_ref()
+}
+
+#[derive(Default)]
+struct ModuleSelection {
+    selected_modules: Vec<&'static dyn LsmModule>,
+    selected_names: BTreeSet<&'static str>,
+    exclusive_module_name: Option<&'static str>,
+}
+
+impl ModuleSelection {
+    fn select_from_boot_params(
+        modules_by_name: &BTreeMap<&'static str, &'static dyn LsmModule>,
+    ) -> Self {
+        let mut selection = Self::default();
+
+        match (LSM_PARAM.get(), LEGACY_SECURITY_PARAM.get()) {
+            (Some(lsm_param), security_param) => {
+                if security_param.is_some() {
+                    warn!("`security=` is ignored because `lsm=` is specified");
+                }
+
+                for name in lsm_param
+                    .split(',')
+                    .map(str::trim)
+                    .filter(|name| !name.is_empty())
+                {
+                    let Some(module) = modules_by_name.get(name).copied() else {
+                        warn!("unknown LSM module `{}` in `lsm=`", name);
+                        continue;
+                    };
+
+                    selection.push(module);
+                }
+            }
+            (None, security_param) => {
+                for module in DEFAULT_MODULES.iter().copied() {
+                    selection.push(module);
+                }
+
+                let Some(security_param) = security_param else {
+                    return selection;
+                };
+                let name = security_param.trim();
+
+                if name.is_empty() {
+                    warn!("`security=` requires an LSM module name");
+                    return selection;
+                }
+
+                let Some(module) = modules_by_name.get(name).copied() else {
+                    warn!("unknown LSM module `{}` in `security=`", name);
+                    return selection;
+                };
+
+                if !module.flags().contains(LsmFlags::LEGACY_MAJOR) {
+                    warn!(
+                        "LSM module `{}` is ignored because `security=` only selects legacy major modules",
+                        name
+                    );
+                    return selection;
+                }
+
+                if module.flags().contains(LsmFlags::EXCLUSIVE)
+                    && let Some(module_name) = selection.exclusive_module_name.take()
+                {
+                    selection
+                        .selected_modules
+                        .retain(|module| module.name() != module_name);
+                    selection.selected_names.remove(module_name);
+                }
+
+                selection.push(module);
+            }
+        }
+
+        selection
+    }
+
+    fn into_modules(self) -> Box<[&'static dyn LsmModule]> {
+        self.selected_modules.into_boxed_slice()
+    }
+
+    fn push(&mut self, module: &'static dyn LsmModule) {
+        let name = module.name();
+
+        if !self.selected_names.insert(name) {
+            warn!("duplicate LSM module `{}` is ignored", name);
+            return;
+        }
+
+        if module.flags().contains(LsmFlags::EXCLUSIVE) {
+            if let Some(exclusive_name) = self.exclusive_module_name {
+                self.selected_names.remove(name);
+                warn!(
+                    "LSM module `{}` is ignored because exclusive module `{}` is already enabled",
+                    name, exclusive_name
+                );
+                return;
+            }
+
+            self.exclusive_module_name = Some(name);
+        }
+
+        self.selected_modules.push(module);
+    }
 }
--- a/kernel/src/security/lsm/modules/mod.rs
+++ b/kernel/src/security/lsm/modules/mod.rs
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MPL-2.0
+
+pub mod yama;
+
+use super::LsmModule;
+
+static ACTIVE_MODULES: [&dyn LsmModule; 1] = [&yama::YAMA_LSM];
+
+pub(super) fn active_modules() -> &'static [&'static dyn LsmModule] {
+    &ACTIVE_MODULES
+}
--- a/kernel/src/security/lsm/modules/yama.rs
+++ b/kernel/src/security/lsm/modules/yama.rs
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use core::sync::atomic::{AtomicI32, Ordering};
+
+use super::super::{AlienAccessContext, LsmAlienAccessCheck, LsmKind, LsmModule};
+use crate::{
+    prelude::*,
+    process::{
+        Process, UserNamespace,
+        credentials::capabilities::CapSet,
+        posix_thread::{AsPosixThread, alien_access::AlienAccessKind},
+    },
+};
+
+pub static YAMA_LSM: YamaLsm = YamaLsm;
+
+/// The Yama minor LSM.
+pub struct YamaLsm;
+
+impl LsmAlienAccessCheck for YamaLsm {
+    fn alien_access_check(&self, context: &AlienAccessContext) -> Result<()> {
+        if context.mode().kind() != AlienAccessKind::Attach {
+            return Ok(());
+        }
+
+        let is_denied = match get_yama_scope() {
+            YamaScope::Disabled => false,
+            YamaScope::Relational => {
+                !context.accessor_has_cap_sys_ptrace()
+                    && !is_ancestor_of(
+                        context.accessor().weak_process(),
+                        context.target().process(),
+                    )
+            }
+            YamaScope::Capability => !context.accessor_has_cap_sys_ptrace(),
+            YamaScope::NoAttach => true,
+        };
+
+        if is_denied {
+            return_errno_with_message!(Errno::EPERM, "alien access is denied due to Yama scope");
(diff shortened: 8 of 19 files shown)


kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,139 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Linux Security Module framework for Asterinas.
+//!
+//! The first goal of this framework is to provide a stable place to host
+//! minor LSMs such as Yama while keeping the hook surface small enough to
+//! evolve with the kernel subsystems.
```
StevenJiang1110: The module-level doc comment should be rewritten.

1. There is no need to mention `for Asterinas` — the entire project is already for Asterinas.
2. Rather than describing the implementation goals, the doc comment should explain what LSM is and outline the key abstractions introduced in this module. This gives readers an immediate high-level understanding of LSM before diving into the details.

test/initramfs/src/regression/security/lsm/yama.c
```diff
@@ -0,0 +1,260 @@
...
+{
+	TEST_SUCC(write_scope(YAMA_SCOPE_RELATIONAL));
+	pid_t target_pid = CHECK(spawn_target_process(true));
+
+	TEST_SUCC(run_sibling_pidfd_getfd(target_pid, false, true));
+
+	cleanup_target_process(target_pid);
+	TEST_SUCC(write_scope(saved_scope));
+}
+END_TEST()
```
StevenJiang1110, line 167: The current test does not conform to our existing conventions. The main issue is that almost all of the test logic has been moved into helper functions, leaving only a trivial entry point in `FN_TEST`. Within helper functions, you are limited to macros such as `CHECK`/`CHECK_WITH`. Instead, the majority of the test logic should reside directly in `FN_TEST`, where you can take full advantage of macros like `TEST_SUCC`/`TEST_RES` and benefit from more detailed output.

test/initramfs/src/regression/security/Makefile
```diff
@@ -3,5 +3,6 @@
 SUBDIRS := \
 	capability \
 	namespace \
+	yama \
```
StevenJiang1110: I think the directory should be renamed as `lsm`, and the test file should be renamed as `yama.c`.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+
+mod modules;
+
+pub(crate) use self::modules::yama::{YamaScope, get_yama_scope, set_yama_scope};
+use crate::{prelude::*, process::posix_thread::PosixThread};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum LsmKind {
```
StevenJiang1110: Replace all `pub(crate)` with `pub` in the `kernel` crate. Since the `kernel` crate has no external callers, `pub` and `pub(crate)` are effectively equivalent within this crate. We prefer the shorter form.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+}
+
+/// Describes which credentials should be used by a ptrace-style access check.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum PtraceAccessCreds {
+    Fs,
+    Real,
+}
```
StevenJiang1110: I don't think `PtraceAccessCreds` needs to be defined as a separate type, since it has exactly the same variants as `CredsSource`. Instead, you could either make `CredsSource` public, or move it to a more appropriate location. 

The same problem applies to `PtraceAccessKind` and `PtraceAccessMode`. If you think the original name is not accurate, you can rename them. But don't define duplicate structs.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Returns the short module name.
+    fn name(&self) -> &'static str;
+
+    /// Returns whether the module is a major or minor LSM.
+    fn kind(&self) -> LsmKind {
+        LsmKind::Minor
+    }
+
+    /// Initializes the module during kernel startup.
+    fn init(&self) {}
```
StevenJiang1110: `name`, `kind`, and `init` are all required methods and should not have default implementations. Each concrete module should be responsible for providing its own implementation.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// Checks ptrace-style access between unrelated tasks.
+    fn ptrace_access_check(&self, context: &PtraceAccessContext<'_>) -> Result<()> {
+        let _ = context;
+        Ok(())
+    }
```
StevenJiang1110: Defining ptrace-related structs and checks in `lsm/mod.rs` seems out of place. As the root module of LSM, it should ideally contain only the most fundamental and general logic. As LSM grows to cover more areas — such as file system and network checks — defining everything in a single file would make it unmaintainable.

I would suggest organizing the checks for different scopes into separate submodules:

```
lsm/
    mod.rs
    checks/
        mod.rs
        ptrace.rs  # ptrace-related checks
        fs.rs          # file system-related checks (to be added in the future)
        net.rs       # network-related checks (to be added in the future)

```

Each check file should only define the checks and data types relevant to that particular scope:

```rust
// lsm/checks/ptrace.rs
// lsm/checks/ptrace.rs
pub trait LsmPtraceCheck: Sync {
    fn ptrace_access_check(&self, context: &PtraceAccessContext<'_>) -> Result<()> {
        let _ = context;
        Ok(())
    }
}

pub struct PtraceAccessContext { ... }

```

The root module then only needs to have LsmModule inherit from LsmPtraceCheck. This way, we avoid concentrating all the logic in a single place:

```rust
// lsm/mod.rs
pub trait LsmModule: LsmPtraceCheck {
    fn name(&self);
    fn kind(&self);
    fn init(&self);
}

```

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        LsmKind::Minor
+    }
+
+    /// Initializes the module during kernel startup.
+    fn init(&self) {}
+
+    /// Checks ptrace-style access between unrelated tasks.
+    fn ptrace_access_check(&self, context: &PtraceAccessContext<'_>) -> Result<()> {
+        let _ = context;
+        Ok(())
```
StevenJiang1110:
```suggestion
    fn ptrace_access_check(&self, _context: &PtraceAccessContext<'_>) -> Result<()> {
        Ok(())
```

StevenJiang1110 (commented):
Thanks for your contribution!

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 7bc0948. I rewrote the module-level docs to describe what the LSM layer does and the key abstractions it introduces, instead of describing the initial implementation goal.

test/initramfs/src/regression/security/lsm/yama.c
zzj-5341: Done in 8903871. I moved the main pidfd_getfd test flow back into the FN_TEST bodies so the assertions can use TEST_SUCC/TEST_RES. The remaining helpers are only for shared setup/cleanup and syscall wrappers.

test/initramfs/src/regression/security/Makefile
zzj-5341: Done in 8903871. The regression test now lives under security/lsm/yama.c, and the Makefile/run script have been updated accordingly.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 7bc0948. I changed the LSM-facing items from pub(crate) to pub in the kernel crate.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 7bc0948. I removed the duplicated ptrace access types and reused the LSM ptrace mode directly. CredsSource is now public, and AlienAccessMode is a type alias of PtraceAccessMode.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 7bc0948. LsmModule now requires name, kind, and init, and YamaLsm implements them explicitly.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 7bc0948. I split ptrace-specific hooks and context types into lsm/checks/ptrace.rs, added LsmPtraceCheck, and made LsmModule inherit from it.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 7bc0948. The default ptrace hook now uses _context, so it no longer needs a dummy let _ = context.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -54,132 +53,16 @@ impl PosixThread {
...
+            self,
+            mode,
+            caller_has_cap,
+        ))?;
 
         Ok(())
     }
 }
 
 /// The mode used by the alien access permission check.
```
StevenJiang1110: This comment is outdated. Please remove it.

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::{
+    prelude::*,
+    process::{
+        Process, UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread,
+    },
+};
+
+pub static YAMA_LSM: YamaLsm = YamaLsm;
+
+/// Implements the Yama minor LSM.
```
StevenJiang1110:
```suggestion
/// The Yama minor LSM.
```

This is a struct not a method. The comment should not be started with a verb.

I found many struct and enum doc comments currently begin with a verb, which is inconsistent with our existing conventions. While trait and method doc comments should indeed begin with a verb, struct and enum doc comments should not. Please follow the style of the existing comments for those.

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,115 @@
...
+pub struct YamaLsm;
+
+impl LsmPtraceCheck for YamaLsm {
+    fn ptrace_access_check(&self, context: &PtraceAccessContext<'_>) -> Result<()> {
```
StevenJiang1110:
```suggestion
    fn ptrace_access_check(&self, context: &PtraceAccessContext) -> Result<()> {
```

Remove the unused lifetime.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+            "[kernel] LSM module enabled: {} ({})",
+            module.name(),
+            module.kind().as_str()
+        );
+        module.init();
+    }
+}
+
+/// Runs ptrace-style access hooks in module order.
+pub fn ptrace_access_check(context: &PtraceAccessContext<'_>) -> Result<()> {
```
StevenJiang1110:
```suggestion
pub fn ptrace_access_check(context: &PtraceAccessContext) -> Result<()> {
```

Further, this method should be moved to `checks/ptrace.rs` and pub use the check in this `lsm/mod.rs`.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Implements the Linux Security Module (LSM) framework used by the kernel.
```
StevenJiang1110:
```suggestion
//! The Linux Security Module (LSM) framework.
```

test/initramfs/src/regression/security/lsm/yama.c
```diff
@@ -0,0 +1,222 @@
...
+	pid_t target_pid = TEST_SUCC(spawn_target_process(true));
+	pid_t attacker_pid = TEST_SUCC(fork());
+
+	if (attacker_pid == 0) {
+		drop_cap_sys_ptrace();
+
+		int pidfd = pidfd_open_syscall(target_pid);
+		if (pidfd < 0) {
+			_exit(20);
+		}
```
StevenJiang1110: Check the return value and exit manually is an anti-pattern. You should use the `CHECK` macro instead. Please also fix other similar cases.

```suggestion
		int pidfd = CHECK(pidfd_open_syscall(target_pid));
```

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    Ok(())
+}
+
+static YAMA_SCOPE: AtomicI32 = AtomicI32::new(YamaScope::Relational as i32);
+
+/// The Yama scope levels.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
+#[repr(i32)]
+pub enum YamaScope {
```
StevenJiang1110:
```suggestion
#[repr(i32)]
#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
pub enum YamaScope {
```

The `#[derive(...)]` macro should always be placed as the last attribute. The original ordering was correct, but this PR has inadvertently mixed up the order.

kernel/src/security/mod.rs
```diff
@@ -9,10 +11,20 @@ cfg_if! {
...
     ostd::if_tdx_enabled!({
         tsm::init();
         tsm_mr::init();
     });
 }
+
+/// Runs the LSM stack for a ptrace-style access check.
+pub fn ptrace_access_check(context: &PtraceAccessContext<'_>) -> Result<()> {
+    lsm::ptrace_access_check(context)
+}
```
StevenJiang1110: I really can't see why introduce this method. Remove it. 

We should let the caller use `lsm::ptrace_access_check(context)` directly.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -54,132 +53,16 @@ impl PosixThread {
             );
         }
 
-        yama::check_alien_access(accessor, self, mode, caller_has_cap)?;
+        security::ptrace_access_check(&PtraceAccessContext::new(
```
StevenJiang1110: Use `lsm::ptrace_access_check` here, just as I said in another comment.

kernel/src/security/lsm/checks/ptrace.rs
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use crate::{prelude::*, process::posix_thread::PosixThread};
+
+/// Defines ptrace-style hooks supported by built-in LSM modules.
+pub trait LsmPtraceCheck: Sync {
+    /// Checks ptrace-style access between unrelated tasks.
```
StevenJiang1110: What does "ptrace-style" mean exactly? These hooks are specifically for the ptrace syscall itself, aren't they?

What do you mean by "unrelated tasks"? The two tasks involved are clearly the tracer and the tracee — why describe them as "unrelated"?

kernel/src/security/lsm/checks/ptrace.rs
```diff
@@ -0,0 +1,99 @@
...
+        self.creds
+    }
+}
+
+/// Carries the inputs for a ptrace-style access check through the LSM stack.
+pub struct PtraceAccessContext<'a> {
+    accessor: &'a PosixThread,
+    target: &'a PosixThread,
+    mode: PtraceAccessMode,
+    accessor_has_sys_ptrace: bool,
```
StevenJiang1110: Rename it as `accessor_has_cap_sys_ptrace` or simply `accessor_has_cap`.

StevenJiang1110 (commented):
Thanks for your efforts! It looks much better now. 

Please also remember to solve the lint errors that CI reports.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+/// Defines the common interface for built-in LSM modules.
+pub trait LsmModule: LsmPtraceCheck + Sync {
+    /// Returns the module name.
+    fn name(&self) -> &'static str;
+
+    /// Returns the module kind.
+    fn kind(&self) -> LsmKind;
```
StevenJiang1110: I think one key question this PR needs to address, as part of implementing an LSM framework, is how to control which LSMs are enabled or disabled.

Following Linux's approach, LSMs can be selected at kernel boot time via the `lsm=...` command-line parameter (e.g., `lsm=lockdown,yama,integrity,apparmor`), or through the legacy `security=...` interface (e.g., `security=apparmor`). If neither parameter is specified, the default set of enabled modules falls back to what was configured at compile time.

I think it would be ideal for this PR to support both of these command-line parameters for controlling module selection, as they both appear to be commonly used in practice. This would also make the methods you defined on `LsmModule` — such as `name` and `kind` — actually useful, rather than being dead code. At least for now, I can't see the usage of these two methods.

From this perspective, the methods currently defined on the `LsmModule` trait also seem insufficient. For instance, Linux defines `LSM_FLAG_EXCLUSIVE` to indicate that certain modules are mutually exclusive and only one of them may be enabled at a time. I think the module selection logic should be implemented more thoroughly to account for cases like this.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Initializes the module during boot.
+    fn init(&self);
```
StevenJiang1110: Remove this `init` function for now. The only module `yama`'s init is an empty function. You can add it back once some module actually needs it.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MPL-2.0
+
+pub mod yama;
+
+use super::LsmModule;
+
+static ACTIVE_MODULES: [&dyn LsmModule; 1] = [&yama::YAMA_LSM];
```
StevenJiang1110: Let's name it as `DEFAULT_MODULES`. The active modules should be decided by both kernel command line parameters and this default modules.

StevenJiang1110 (commented):
I have been studying LSM a bit more recently. I think one of the core problems that an LSM framework needs to address is determining which modules are enabled — something that this PR has not yet tackled. Please refer to my specific comments below for more details.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in a41577dad. I removed `LsmModule::init()` and the empty `YamaLsm::init()` implementation. We can add module-specific initialization back when a module actually needs it.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in a41577dad. I renamed the compiled-in list to `DEFAULT_MODULES`; `ACTIVE_MODULES` is now initialized at boot from `lsm=`, `security=`, or this default list.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done in 72fd0a98f and a0bcb6bf2. I changed `YamaLsm` to use `The Yama minor LSM.` and also cleaned up the other touched struct/enum doc comments so they no longer start with verbs.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done in 72fd0a98f. I removed the unnecessary explicit `'_` lifetime from the ptrace hook signatures.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 72fd0a98f. I moved `ptrace_access_check` into `checks/ptrace.rs` and re-exported it from `lsm/mod.rs`. I also removed the unnecessary explicit `'_` lifetime from the signature.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in 72fd0a98f. I updated the module-level summary to `The Linux Security Module (LSM) framework.`

test/initramfs/src/regression/security/lsm/yama.c
zzj-5341: Done in 72fd0a98f. I replaced the manual return-value checks in both affected child paths with `CHECK`/`CHECK_WITH`, including the `pidfd_open`, `pidfd_getfd`, `pread`, `close`, and content comparison checks.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done in 72fd0a98f. I restored the attribute order so `#[derive(...)]` is the last attribute.

kernel/src/security/mod.rs
zzj-5341: Done in 72fd0a98f. I removed the `security::ptrace_access_check` wrapper and updated the caller to use `lsm::ptrace_access_check` directly.

kernel/src/security/lsm/checks/ptrace.rs
zzj-5341: Done in 72fd0a98f. I replaced the ambiguous `ptrace-style` and `unrelated tasks` wording with direct ptrace access terminology.

kernel/src/security/lsm/mod.rs
zzj-5341: Done in a41577dad. The active LSM stack is now selected from the kernel command line: `lsm=` builds the module list directly, `security=` is handled as the legacy selector, and the code falls back to `DEFAULT_MODULES` when neither is specified. I also added `LsmFlags::EXCLUSIVE` and `LsmModule::flags()` so mutually exclusive modules can be accounted for.

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
```diff
@@ -13,10 +13,10 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
-    process::posix_thread::alien_access::yama::{YamaScope, get_yama_scope, set_yama_scope},
+    security::lsm::{YamaScope, get_yama_scope, set_yama_scope},
 };
 
-/// Represents the inode at `/proc/sys/kernel/yama`.
+/// Directory operations for `/proc/sys/kernel/yama`.
```
StevenJiang1110: Please avoid modifying the documentation in this file. The original writing follows the established style consistently used across all procfs files.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::prelude::*;
+
+static LSM_PARAM: Once<String> = Once::new();
+static LEGACY_SECURITY_PARAM: Once<String> = Once::new();
+
+aster_cmdline::define_kv_param!("lsm", LSM_PARAM);
+aster_cmdline::define_kv_param!("security", LEGACY_SECURITY_PARAM);
+
+static BUILTIN_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
+pub(super) static DEFAULT_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
```
StevenJiang1110, line 44: `BUILTIN_MODULES` should be renamed to `ALL_MODULES`, as it is intended to represent all available security modules. 

Additionally, please add documentation comments to explain the purpose of these two static variables.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+
+fn select_default_modules() -> Vec<&'static dyn LsmModule> {
+    let mut modules = Vec::new();
+
+    for module in DEFAULT_MODULES.iter().copied() {
+        push_selected_module(&mut modules, module);
+    }
+
+    modules
+}
```
StevenJiang1110, line 68: This method `select_default_modules` can be simplified as just one line: `DEFAULT_MODULES.to_vec()`. I don't think we need this method.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        return;
+    };
+
+    push_selected_module(modules, module);
+}
+
+fn find_builtin_module(name: &str) -> Option<&'static dyn LsmModule> {
+    BUILTIN_MODULES
+        .iter()
+        .copied()
```
StevenJiang1110: The `copied` should be moved behind `find`, you only need to copy the found module.

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
```diff
@@ -64,7 +64,7 @@ impl DirOps for YamaDirOps {
     }
 }
 
-/// Represents the inode at `/proc/sys/kernel/yama/ptrace_scope`.
+/// File operations for `/proc/sys/kernel/yama/ptrace_scope`.
```
StevenJiang1110: Since we now can disable yama by kernel cmd lines, should this file be visible and writeble for users if yama is disabled?

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        .map(str::trim)
+        .filter(|name| !name.is_empty())
+    {
+        select_module_by_name(&mut modules, name);
+    }
+
+    modules
+}
+
+fn select_modules_from_security_param(param: &str) -> Vec<&'static dyn LsmModule> {
```
StevenJiang1110: You should add some document to explain how `lsm=` and `security=` works.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        .any(|selected| selected.name() == module.name())
+    {
+        warn!("duplicate LSM module `{}` is ignored", module.name());
+        return;
+    }
+
+    if module.flags().contains(LsmFlags::EXCLUSIVE)
+        && let Some(selected) = modules
+            .iter()
+            .find(|selected| selected.flags().contains(LsmFlags::EXCLUSIVE))
```
StevenJiang1110: The current implementation has some inefficiencies. For example, checking whether any previously registered LSM is exclusive requires iterating over the entire list, and `find_builtin_module` also iterates over all modules to find one by name — both of which seem unnecessary. On top of that, lines 111–113 perform yet another full iteration just to check for duplicate module names.

I have drafted a rough rewrite locally that introduces a `ModuleSelection` abstraction, which eliminates these redundant iterations. You can use it as a reference when revising the implementation.

```diff
diff --git a/kernel/src/security/lsm/modules/mod.rs b/kernel/src/security/lsm/modules/mod.rs
index 45a7b87f1..79e032815 100644
--- a/kernel/src/security/lsm/modules/mod.rs
+++ b/kernel/src/security/lsm/modules/mod.rs
@@ -14,11 +14,20 @@ aster_cmdline::define_kv_param!("lsm", LSM_PARAM);
 aster_cmdline::define_kv_param!("security", LEGACY_SECURITY_PARAM);
 
 static BUILTIN_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
+static BUILTIN_MODULES_BY_NAME: Once<BTreeMap<&'static str, &'static dyn LsmModule>> = Once::new();
 pub(super) static DEFAULT_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
 
 static ACTIVE_MODULES: Once<Vec<&'static dyn LsmModule>> = Once::new();
 
 pub(super) fn init() {
+    BUILTIN_MODULES_BY_NAME.call_once(|| {
+        BUILTIN_MODULES
+            .iter()
+            .copied()
+            .map(|module| (module.name(), module))
+            .collect()
+    });
+
     ACTIVE_MODULES.call_once(select_active_modules);
 }
 
@@ -30,104 +39,103 @@ pub(super) fn active_modules() -> &'static [&'static dyn LsmModule] {
 }
 
 fn select_active_modules() -> Vec<&'static dyn LsmModule> {
-    if let Some(lsm_param) = LSM_PARAM.get() {
-        if LEGACY_SECURITY_PARAM.get().is_some() {
-            warn!("`security=` is ignored because `lsm=` is specified");
-        }
+    let selector = ModuleSelection::default();
 
-        return select_modules_from_lsm_param(lsm_param);
-    }
+    match (LSM_PARAM.get(), LEGACY_SECURITY_PARAM.get()) {
+        (Some(lsm_param), legacy_security_param) => {
+            if legacy_security_param.is_some() {
+                warn!("`security=` is ignored because `lsm=` is specified");
+            }
 
-    if let Some(security_param) = LEGACY_SECURITY_PARAM.get() {
-        return select_modules_from_security_param(security_param);
+            selector.select_from_lsm_param(lsm_param)
+        }
+        (None, Some(security_param)) => selector.select_from_security_param(security_param),
+        (None, None) => {
+            let mut selector = selector;
+            selector.push_default_modules();
+            selector.selected_modules
+        }
     }
-
-    select_default_modules()
 }
 
-fn select_default_modules() -> Vec<&'static dyn LsmModule> {
-    let mut modules = Vec::new();
-
-    for module in DEFAULT_MODULES.iter().copied() {
-        push_selected_module(&mut modules, module);
-    }
-
-    modules
+#[derive(Default)]
+struct ModuleSelection {
+    selected_modules: Vec<&'static dyn LsmModule>,
+    selected_names: BTreeSet<&'static str>,
+    exclusive_module_name: Option<&'static str>,
 }
 
-fn select_modules_from_lsm_param(param: &str) -> Vec<&'static dyn LsmModule> {
-    let mut modules = Vec::new();
+impl ModuleSelection {
+    fn select_from_lsm_param(mut self, param: &str) -> Vec<&'static dyn LsmModule> {
+        for name in param
+            .split(',')
+            .map(str::trim)
+            .filter(|name| !name.is_empty())
+        {
+            let Some(module) = BUILTIN_MODULES_BY_NAME.get().unwrap().get(name) else {
+                warn!("unknown LSM module `{}` in `lsm=`", name);
+                continue;
+            };
+
+            self.push(*module);
+        }
 
-    for name in param
-        .split(',')
-        .map(str::trim)
-        .filter(|name| !name.is_empty())
-    {
-        select_module_by_name(&mut modules, name);
+        self.selected_modules
     }
 
-    modules
-}
-
-fn select_modules_from_security_param(param: &str) -> Vec<&'static dyn LsmModule> {
-    let mut modules = select_default_modules();
-    let name = param.trim();
+    fn select_from_security_param(mut self, param: &str) -> Vec<&'static dyn LsmModule> {
+        self.push_default_modules();
 
-    if name.is_empty() {
-        warn!("`security=` requires an LSM module name");
-        return modules;
-    }
+        let name = param.trim();
+        if name.is_empty() {
+            warn!("`security=` requires an LSM module name");
+            return self.selected_modules;
+        }
 
-    let Some(module) = find_builtin_module(name) else {
-        warn!("unknown LSM module `{}` in `security=`", name);
-        return modules;
-    };
+        let Some(module) = BUILTIN_MODULES_BY_NAME.get().unwrap().get(name) else {
+            warn!("unknown LSM module `{}` in `security=`", name);
+            return self.selected_modules;
+        };
+
+        if module.flags().contains(LsmFlags::EXCLUSIVE) {
+            if let Some(exclusive_name) = self.exclusive_module_name.take() {
+                self.selected_modules
+                    .retain(|module| module.name() != exclusive_name);
+                self.selected_names.remove(exclusive_name);
+            }
+        }
 
-    if module.flags().contains(LsmFlags::EXCLUSIVE) {
-        modules.retain(|selected| !selected.flags().contains(LsmFlags::EXCLUSIVE));
+        self.push(*module);
+        self.selected_modules
     }
 
-    push_selected_module(&mut modules, module);
-    modules
-}
-
-fn select_module_by_name(modules: &mut Vec<&'static dyn LsmModule>, name: &str) {
-    let Some(module) = find_builtin_module(name) else {
-        warn!("unknown LSM module `{}` in `lsm=`", name);
-        return;
-    };
+    fn push_default_modules(&mut self) {
+        for module in DEFAULT_MODULES {
+            self.push(module);
+        }
+    }
 
-    push_selected_module(modules, module);
-}
+    fn push(&mut self, module: &'static dyn LsmModule) {
+        let name = module.name();
 
-fn find_builtin_module(name: &str) -> Option<&'static dyn LsmModule> {
-    BUILTIN_MODULES
-        .iter()
-        .copied()
-        .find(|module| module.name() == name)
-}
+        if !self.selected_names.insert(name) {
+            warn!("duplicate LSM module `{}` is ignored", name);
+            return;
+        }
 
-fn push_selected_module(modules: &mut Vec<&'static dyn LsmModule>, module: &'static dyn LsmModule) {
-    if modules
-        .iter()
-        .any(|selected| selected.name() == module.name())
-    {
-        warn!("duplicate LSM module `{}` is ignored", module.name());
-        return;
-    }
+        if module.flags().contains(LsmFlags::EXCLUSIVE) {
+            if let Some(exclusive_name) = self.exclusive_module_name {
+                self.selected_names.remove(name);
+                warn!(
+                    "LSM module `{}` is ignored because exclusive module `{}` is already enabled",
+                    name, exclusive_name
+                );
+                return;
+            }
+
+            self.exclusive_module_name = Some(name);
+        }
 
-    if module.flags().contains(LsmFlags::EXCLUSIVE)
-        && let Some(selected) = modules
-            .iter()
-            .find(|selected| selected.flags().contains(LsmFlags::EXCLUSIVE))
-    {
-        warn!(
-            "LSM module `{}` is ignored because exclusive module `{}` is already enabled",
-            module.name(),
-            selected.name()
-        );
-        return;
+        self.selected_modules.push(module);
     }
-
-    modules.push(module);
 }
```

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I restored the documentation in this procfs file to the existing procfs style.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I renamed `BUILTIN_MODULES` to `ALL_MODULES` and added documentation comments for both `ALL_MODULES` and `DEFAULT_MODULES`.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I removed `select_default_modules`; default modules are now added directly through `ModuleSelection`.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I removed `find_builtin_module` and replaced it with an `ALL_MODULES_BY_NAME` lookup, so the module reference is copied only after a successful lookup.

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I now hide `/proc/sys/kernel/yama` from `/proc/sys/kernel` when the Yama LSM is disabled, so `ptrace_scope` is neither visible nor writable in that case.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I added module-level documentation explaining how `lsm=` and the legacy `security=` selector work, including default fallback, exclusive-module replacement, unknown names, and duplicate names.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in c61a8d25f, and it is preserved after the latest merge in 1271d0109. I rewrote the module selection logic around `ModuleSelection`, added `ALL_MODULES_BY_NAME` for name-based lookup, and track selected module names and the current exclusive module explicitly. This removes `find_builtin_module` and avoids the repeated full-list scans for lookup, duplicate detection, and exclusive-module checks.

kernel/src/fs/fs_impls/procfs/sys/kernel/mod.rs
```diff
@@ -33,8 +36,16 @@ impl KernelDirOps {
...
+    const STATIC_ENTRIES: &'static [StaticEntry] = &[
+        (
+            "cap_last_cap",
+            InodeType::File,
+            CapLastCapFileOps::new_inode,
+        ),
+        ("pid_max", InodeType::File, PidMaxFileOps::new_inode),
+    ];
+
+    const STATIC_ENTRIES_WITH_YAMA: &'static [StaticEntry] = &[
```
StevenJiang1110: I don't think defining an entry like this is the right approach. As the name suggests, this list should only contain static entries — that is, entries whose existence is determined at compile time. Entries that can be toggled or adjusted at runtime(yama) should not be included in this list.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+
+pub(super) fn init() {
+    active_modules();
+}
+
+pub(super) fn active_modules() -> &'static [&'static dyn LsmModule] {
+    ACTIVE_MODULES.call_once(select_active_modules).as_slice()
+}
+
+pub(super) fn is_module_enabled(name: &str) -> bool {
```
StevenJiang1110: Please remove the `pub(super)` visibility marker, as this method is only called within the same module.

Also, since this function is very short and is only called once, I don't think it warrants being a separate function at all — it would be cleaner to just inline it at the call site.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+    let mut modules_by_name = BTreeMap::new();
+
+    for module in ALL_MODULES.iter().copied() {
+        if modules_by_name.contains_key(module.name()) {
+            warn!(
+                "duplicate built-in LSM module `{}` is ignored",
+                module.name()
+            );
+            continue;
+        }
```
StevenJiang1110: I don't think the checks on lines 54–60 are necessary. `ALL_MODULES` is a static variable, and there is no need to check for duplicate modules at runtime — this is something that can be trivially verified by reading the code. Please remove these checks.

The function itself can be simplified to a single line: `ALL_MODULES.iter().map(|module| (module.name(), *module)).collect()`. Furthermore, since this function is only called once, I don't think it warrants being a separate function — please inline it at the call site directly.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        modules_by_name.insert(module.name(), module);
+    }
+
+    modules_by_name
+}
+
+fn select_active_modules() -> Vec<&'static dyn LsmModule> {
```
StevenJiang1110: This function is also only called once, so please inline it at the call site directly — the caller is already just a single line, so this won't add any complexity.

Additionally, I think this function should return `Box<[&'static dyn LsmModule]>` to convey that the length of the collection is fixed after initialization and will never change. The type of `ACTIVE_MODULES` and `into_modules` should be updated accordingly.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        let Some(module_name) = self.exclusive_module_name.take() else {
+            return;
+        };
+
+        self.modules.retain(|module| module.name() != module_name);
+        self.selected_names.remove(module_name);
+    }
+}
+
+fn find_module_by_name(name: &str, param_name: &str) -> Option<&'static dyn LsmModule> {
```
StevenJiang1110: I don't see why param_name is being passed here — using it solely for a warning message does not justify its inclusion. Also, this method can be simplified to a single line: `all_modules_by_name().get(name).copied()`. It is too trivial to warrant being its own method.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        };
+
+        if module.flags().contains(LsmFlags::EXCLUSIVE) {
+            self.remove_current_exclusive_module();
+        }
+
+        self.push(module);
+    }
+
+    fn push_by_name(&mut self, name: &str, param_name: &str) {
```
StevenJiang1110: Methods like `push_by_name` and `remove_current_exclusive_module` are simple helper functions with only a few lines of code each, and both are called only once from a call site that is not particularly complex. I don't think either of them warrants being extracted into a dedicated function. More importantly, the semantics of these functions are not clear enough to make them good candidates for abstraction in the first place.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,189 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Built-in LSM module registration and boot-time selection.
+//!
+//! The `lsm=` kernel parameter explicitly lists enabled modules in stack order.
+//! The legacy `security=` parameter selects one module by name on top of the
+//! default stack. If the selected module is exclusive, it replaces the currently
+//! selected exclusive module. If neither parameter is specified, the default
+//! module stack is used. Unknown module names are ignored with a warning, and
+//! duplicate names are ignored after their first selection.
```
StevenJiang1110: The documentation is a bit hard to follow. One issue is that it is not clear which parts describe the behavior of lsm= and which parts describe the behavior of security=. I would like you to restructure the content into clearly separated paragraphs rather than writing everything in one block.

Paragraph 1: Introduce the two command-line parameters that Linux provides for selecting which LSM modules to enable. lsm= is the modern approach, while security= is the legacy interface.
Paragraph 2: Describe how lsm= is processed.
Paragraph 3: Describe how security= is processed.
Paragraph 4: Describe what happens when both parameters are specified, and what happens when neither is specified.

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,119 @@
...
+        Ok(())
+    }
+}
+
+impl LsmModule for YamaLsm {
+    fn name(&self) -> &'static str {
+        "yama"
+    }
+
+    fn kind(&self) -> LsmKind {
```
StevenJiang1110: The `kind` method is also dead code? Let's remove it. Or you can add it to `LsmFlags`, since Linux also encodes it into `LsmFlags`.

Could you double-check whether any logic has been accidentally omitted here? Specifically, why is `kind` never used? Linux does make use of this flag during initialization.

```C
      lsm_for_each_raw(lsm) {
			if ((lsm->flags & LSM_FLAG_LEGACY_MAJOR) &&
			     strcmp(lsm->id->name, lsm_order_legacy)) {
				lsm_enabled_set(lsm, false);
				lsm_pr_dbg("skip legacy LSM conflict %s:%s\n",
					   src, lsm->id->name);
			}
		}
```

StevenJiang1110 (commented):
Thanks for your efforts! Could you also click `Resolve conversations` button for all above comments that have already been addressed? That way they won't draw unnecessary attention.

kernel/src/fs/fs_impls/procfs/sys/kernel/mod.rs
zzj-5341: Done in 9185a6c54. I removed `STATIC_ENTRIES_WITH_YAMA`, so `STATIC_ENTRIES` now only contains entries whose existence is compile-time static. The `yama` directory is now handled dynamically in `lookup_child` and `visit_entries_from_offset` only when the Yama LSM is enabled.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in 9185a6c54. I removed `is_module_enabled` and inlined the only call site into `is_yama_enabled`.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in 9185a6c54. I removed the runtime duplicate checks for `ALL_MODULES`, and inlined the map construction at the only call site as `ALL_MODULES.iter().map(|module| (module.name(), *module)).collect()`.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in 9185a6c54. I inlined `select_active_modules` into `active_modules`, and changed `ACTIVE_MODULES` plus `ModuleSelection::into_modules` to use `Box<[&'static dyn LsmModule]>`.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in 9185a6c54. I removed `find_module_by_name` and no longer pass `param_name`; the lookup is now done directly at the call sites with `modules_by_name.get(name).copied()`.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in 9185a6c54. I removed `push_by_name` and `remove_current_exclusive_module`, and inlined their logic into the module-selection flow.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done in 9185a6c54. I restructured the module-level documentation into four separate paragraphs covering the two Linux command-line parameters, `lsm=` processing, `security=` processing, and the precedence/default behavior.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done in 9185a6c54. I removed `LsmKind` and the `kind()` method, and added `LsmFlags::LEGACY_MAJOR` instead. I also updated `security=` handling to only select modules with `LEGACY_MAJOR`, which matches the Linux logic that uses the legacy-major flag during initialization. Yama is a minor LSM, so it keeps empty flags.

kernel/src/fs/fs_impls/procfs/sys/kernel/mod.rs
```diff
@@ -9,19 +9,22 @@ use crate::{
...
     },
     prelude::*,
+    security::lsm::is_yama_enabled,
 };
 
 mod cap_last_cap;
 mod pid_max;
 mod yama;
 
+type StaticEntry = StaticDirEntry<fn(Weak<dyn Inode>) -> Arc<dyn Inode>>;
```
StevenJiang1110: Defining a type alias to simplify a complex type is a good idea. However, there are many other places in procfs that still use the underlying type directly and have not yet adopted this alias. For consistency, I would suggest moving the type alias to `procfs/mod.rs` and updating all other usages throughout procfs to use it as well.

kernel/src/security/lsm/checks/ptrace.rs
```diff
@@ -0,0 +1,109 @@
...
+    ) -> Self {
+        Self {
+            accessor,
+            target,
+            mode,
+            accessor_has_cap_sys_ptrace,
+        }
+    }
+
+    pub const fn accessor(&self) -> &'a PosixThread {
```
StevenJiang1110:
```suggestion
    pub const fn accessor(&self) -> &PosixThread {
```

kernel/src/security/lsm/checks/ptrace.rs
```diff
@@ -0,0 +1,109 @@
...
+        self.accessor
+    }
+
+    pub const fn target(&self) -> &'a PosixThread {
```
StevenJiang1110:
```suggestion
    pub const fn target(&self) -> &PosixThread {
```

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+
+/// The fallback LSM stack used when no boot-time selector is specified.
+pub(super) static DEFAULT_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
+
+static ALL_MODULES_BY_NAME: Once<BTreeMap<&'static str, &'static dyn LsmModule>> = Once::new();
+static ACTIVE_MODULES: Once<Box<[&'static dyn LsmModule]>> = Once::new();
+
+pub(super) fn init() {
+    active_modules();
+}
```
StevenJiang1110: Is the `init` function still necessary? Since accessing `active_modules` for the first time will trigger its initialization anyway, the explicit init call seems redundant.

If we do decide to keep it, it could be refactored to:

```rust
// Access active_modules to trigger its lazy initialization.
let _ = active_modules();
```

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+                            .retain(|module| module.name() != module_name);
+                        selection.selected_names.remove(module_name);
+                    }
+
+                    selection.push(module);
+                }
+            }
+
+            selection.into_modules()
+        })
```
StevenJiang1110, line 59: The closure has become quite long, spanning lines 49–115. Embedding such a lengthy block of logic inside a closure leads to excessive indentation and makes the code difficult to read. Consider extracting part of it — for example, lines 54–114 — into a dedicated method on `ModuleSelection`, such as `select_from_boot_params`.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+
+//! Built-in LSM module registration and boot-time selection.
+//!
+//! Linux provides two kernel command-line parameters for choosing enabled LSM
+//! modules: `lsm=` is the modern ordered module list, while `security=` is the
+//! legacy selector for one major LSM.
+//!
+//! When `lsm=` is specified, its comma-separated module names are used as the
+//! enabled LSM stack in that order. Unknown module names are ignored with a
+//! warning, and duplicate names are ignored after their first selection.
```
StevenJiang1110, line 11: Could you also add a comment here explaining the behavior of the exclusive flag for the `lsm=` option? It appears to differ from `security=` — for `lsm=`, it is the first exclusive entry that takes effect, whereas for `security=`, it is the last one. A brief explanation of this distinction would help clarify the intent.

StevenJiang1110 (commented):
LGTM! These should be my final comments — thank you for your contribution!

Please also clean up your commit history before merging. For instance, commits such as "[Merge upstream main into LSM branch](https://github.com/asterinas/asterinas/pull/3078/changes/bed42f1d340597413472d6bdee7f0ddf5e79ccef)" are not acceptable. You should rebase your changes onto the latest main branch and resolve any conflicts, rather than merging main into your feature branch. Additionally, any temporary commits created to address review comments should be squashed into the appropriate earlier commits. Keeping the project's commit history clean and linear is important to us.

kernel/src/fs/fs_impls/procfs/sys/kernel/mod.rs
zzj-5341: Done. I moved the aliases to procfs/mod.rs and updated the procfs users to use StaticEntry / StaticEntryWithOps consistently.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done. I removed the redundant modules::init; active_modules() now performs lazy initialization when first accessed.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done. I extracted the boot-parameter selection logic into ModuleSelection::select_from_boot_params.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done. I added documentation explaining that lsm= keeps the first selected exclusive module, while security= applies after the default stack and can replace the selected exclusive module.

zzj-5341:
> LGTM! These should be my final comments — thank you for your contribution!
> 
> Please also clean up your commit history before merging. For instance, commits such as "[Merge upstream main into LSM branch](https://github.com/asterinas/asterinas/pull/3078/changes/bed42f1d340597413472d6bdee7f0ddf5e79ccef)" are not acceptable. You should rebase your changes onto the latest main branch and resolve any conflicts, rather than merging main into your feature branch. Additionally, any temporary commits created to address review comments should be squashed into the appropriate earlier commits. Keeping the project's commit history clean and linear is important to us.

Done. I rebased the PR onto the latest main branch, removed the merge commit, and squashed the temporary review-fix commits into three logical commits. The branch history is now linear.

kernel/src/security/lsm/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+//! dispatch helpers shared by built-in modules such as `yama`. Module selection
+//! follows the `lsm=` and legacy `security=` kernel command-line parameters.
+
+mod checks;
+mod modules;
+
+pub use self::{
+    checks::ptrace::{
+        CredsSource, LsmPtraceCheck, PtraceAccessContext, PtraceAccessKind, PtraceAccessMode,
+        ptrace_access_check,
```
lrh2000: This reverts https://github.com/asterinas/asterinas/pull/2855#discussion_r2887306332 by reintroducing the `ptrace` module and `PtraceXxx` types. We should have a consistent convention for this, i.e., when to use `ptrace`/`alien`...?

test/initramfs/src/regression/security/Makefile
```diff
@@ -3,5 +3,6 @@
 SUBDIRS := \
 	capability \
 	namespace \
+	lsm \
```
lrh2000, line 5:
```
	capability \
	lsm \
	namespace \
```

This needs to be sorted. Same for `run_test.sh`:
```
./capability/execve

./lsm/yama

./namespace/cgroup_ns
```

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,113 @@
...
+        })?;
+
+    Ok(())
+}
+
+static YAMA_SCOPE: AtomicI32 = AtomicI32::new(YamaScope::Relational as i32);
+
+/// The Yama scope levels.
+#[repr(i32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
```
lrh2000: The removed code in `alien_access.rs` is:
```patch
-#[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromInt)]
```

I assume it is a rebase error. All `Debug, Clone, Copy, PartialEq, Eq, TryFromInt` in this file and in `kernel/src/security/lsm/checks/ptrace.rs` should be converted back to `Clone, Copy, Debug, Eq, PartialEq, TryFromInt`, which is sorted alphabetically.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -54,132 +53,15 @@ impl PosixThread {
...
+            mode,
+            caller_has_cap,
+        ))?;
 
         Ok(())
     }
 }
 
-/// The mode used by the alien access permission check.
-pub struct AlienAccessMode(AlienAccessFlags, CredsSource);
```
lrh2000, line 64: Also is there really a need to move this type out of this file? I don't immediately understand why it's necessary, but maybe I'm missing something.

If we do so, this module is almost empty. I suppose there is no need for the type alias and the module should be marked as private (e.g., `pub mod alien_access` -> `mod alien_access`)?

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    fn kind(&self) -> LsmKind {
+        LsmKind::Minor
+    }
+
+    fn init(&self) {}
+}
+
+/// Returns the current Yama scope for alien access.
+pub fn get_yama_scope() -> YamaScope {
```
tatetian: Rename the `get_yama_scope` and `set_yama_scope` to `get_scope` and `set_scope, respectively.

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
tatetian: According to the [`qualified-fn-imports`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#qualified-fn-imports) guideline, here should be rewritten as `yama::get_scope`.

The same argument applies to the subsequent `set_yama_scope`, which should be rewritten to `yama::set_scope`.

kernel/src/security/lsm/checks/ptrace.rs
```diff
@@ -0,0 +1,109 @@
...
+/// The inputs for a ptrace access check through the LSM stack.
+pub struct PtraceAccessContext<'a> {
+    accessor: &'a PosixThread,
+    target: &'a PosixThread,
+    mode: PtraceAccessMode,
+    accessor_has_cap_sys_ptrace: bool,
+}
+
+impl<'a> PtraceAccessContext<'a> {
+    pub const fn new(
```
tatetian: Add the missing doc comments for all methods.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -54,132 +53,15 @@ impl PosixThread {
             );
         }
 
-        yama::check_alien_access(accessor, self, mode, caller_has_cap)?;
+        lsm::ptrace_access_check(&PtraceAccessContext::new(
```
tatetian: My mental model is that `lsm` provides a number of _hook points_, upon which LSM module can add extra checks. To make this model more explicit, I suggest (1) renaming `lsm::checks` to `hooks`, (2) renaming each hook points as `on_xxx`. For example,

```rust
use security::lsm::hooks as lsm_hooks;

lsm_hooks::on_ptrace_access(&PtraceAccessContext::new(...))?;
```

kernel/src/security/lsm/checks/ptrace.rs
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use super::super::modules;
+use crate::{prelude::*, process::posix_thread::PosixThread};
+
+/// Defines hooks for ptrace access checks.
+pub trait LsmPtraceCheck: Sync {
```
tatetian: According to [`narrow-visibility`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html#narrow-visibility), mark this trait `pub(super)` because only LSM modules need to see and implement this trait.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -54,132 +53,15 @@ impl PosixThread {
...
-            }
-            let Some(parent) = parent.upgrade() else {
-                return false;
-            };
-            drop(parent_guard);
-            current = parent;
-        }
-    }
-}
+pub type AlienAccessMode = PtraceAccessMode;
```
tatetian: I think that you need to make up your mind as to whether this check targets "alien access" or "ptrace access". And be consistent with the terminology through the `posix_thread` and `lsm` modules. I think both have its merits, but using two different terms is confusing.

To make this decision, we need to ask this fundamental question: what are all the call sites of this hook? Does the `ptrace` module ever use the hook? In the current form, we have the `lsm::checks::ptrace` module. The natural expectation of people seeing this module is to believe that the module provides some checks used by the `ptrace` module. But the ptrace implementation in `posix_thread::ptrace` never uses the hooks! Is it because that the `posix_thread::ptrace` module lacks certain LSM hooks or it will never use any LSM hooks?

If the ptrace implementation never uses the hook, then I suppose that it makes more sense to call the hook `alien_access_check` and the `lsm::checks::ptrace` should be renamed to `lsm::checks::alien_access`. Otherwise, we should probably consolidate the concept of "alien access" with "ptrace access" and merge the `posix_thread::alien_access` module into `posix_thread::ptrace`.

kernel/src/security/lsm/mod.rs
zzj-5341: Done. I picked `alien access` as the consistent terminology and removed the reintroduced `ptrace` hook/module naming and `PtraceXxx` types.

test/initramfs/src/regression/security/Makefile
zzj-5341: Done. The security regression entries are now sorted as `capability`, `lsm`, and `namespace` in both the Makefile and `run_test.sh`.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done. The derive order is now `Clone, Copy, Debug, Eq, PartialEq, TryFromInt`.

kernel/src/process/posix_thread/alien_access.rs
zzj-5341: Done. I kept `CredsSource`, `AlienAccessKind`, and `AlienAccessMode` in `posix_thread::alien_access`, so there is no separate type alias or almost-empty moved module anymore.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done. I renamed them to `get_scope` and `set_scope`.

kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs
zzj-5341: Done. The procfs caller now imports the `yama` module and calls `yama::get_scope` / `yama::set_scope`.

kernel/src/security/lsm/checks/ptrace.rs
zzj-5341: Done. I added doc comments for all `AlienAccessContext` methods.

kernel/src/process/posix_thread/alien_access.rs
zzj-5341: Done. I renamed `lsm::checks` to `lsm::hooks` and changed the dispatch entry to `lsm_hooks::on_alien_access`.

kernel/src/security/lsm/checks/ptrace.rs
zzj-5341: Done. The LSM hook trait is now internal to `lsm::hooks` with `pub(super)` visibility.

kernel/src/process/posix_thread/alien_access.rs
zzj-5341: Done. I chose `alien access` as the consistent concept because this hook is called from `posix_thread::alien_access`, not from the ptrace module. The LSM side now uses `lsm::hooks::on_alien_access`, and the old `ptrace` hook/module naming plus `PtraceXxx` types have been removed.

kernel/src/fs/fs_impls/procfs/mod.rs
```diff
@@ -70,6 +70,11 @@ const PROC_ROOT_INO: u64 = 1;
 /// Block size.
 const BLOCK_SIZE: usize = 1024;
 
+pub(in crate::fs::fs_impls::procfs) type StaticEntry =
+    StaticDirEntry<fn(Weak<dyn Inode>) -> Arc<dyn Inode>>;
+pub(in crate::fs::fs_impls::procfs) type StaticEntryWithOps<T> =
```
tatetian: The visibility modifier should simply be private...

kernel/src/security/lsm/modules/yama.rs
```diff
@@ -0,0 +1,115 @@
...
+            Error::with_message(
+                Errno::EINVAL,
+                "`YamaScope::NoAttach` cannot be changed once set",
+            )
+        })?;
+
+    Ok(())
+}
+
+static YAMA_SCOPE: AtomicI32 = AtomicI32::new(YamaScope::Relational as i32);
```
tatetian: I believe you want to define an `AtomicYamaScope` type with the `define_atomic_version_of_integer_like_type` macro. Search the codebase and learn from the existing users of this macro.

kernel/src/fs/fs_impls/procfs/mod.rs
```diff
@@ -70,6 +70,11 @@ const PROC_ROOT_INO: u64 = 1;
...
+    StaticDirEntry<fn(&T, Weak<dyn Inode>) -> Arc<dyn Inode>>;
```
tatetian: Follow the [`top-down-reading`](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/#top-down-reading) guideline. Move less important concept or implementation detail closer to the bottom of this file.

kernel/src/security/lsm/hooks/alien_access.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Returns whether the accessor has `CapSet::SYS_PTRACE`.
+    pub const fn accessor_has_cap_sys_ptrace(&self) -> bool {
+        self.accessor_has_cap_sys_ptrace
+    }
+}
+
+/// Runs alien access hooks in module order.
+pub fn on_alien_access(context: &AlienAccessContext) -> Result<()> {
```
tatetian: Follow the [`top-down-reading`](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/#top-down-reading) guideline. Move the most high-level, public API closer to the top of this file.

kernel/src/fs/fs_impls/procfs/mod.rs
zzj-5341: Done in 2ee08c9. I changed the procfs static-entry aliases to private type aliases.

kernel/src/security/lsm/modules/yama.rs
zzj-5341: Done in 2ee08c9. I added AtomicYamaScope with define_atomic_version_of_integer_like_type and now use typed load/fetch_update operations for YamaScope.

tatetian (approved):

tatetian:
@lrh2000 Sorry, I should have waited for your approval... I hope that your concerns have all been addressed.

lrh2000:
> @lrh2000 Sorry, I should have waited for your approval... I hope that your concerns have all been addressed.

Not a problem -- they've been addressed. Thanks! (Also thank @zzj-5341 for addressing them.)

PR #3082, opened 2026-04-02
Bind mount ns files

StevenJiang1110:
This PR improves Linux compatibility for `nsfs` bind mounts and updates `proc_nsfs`.

 This PR contains 4 commits: 

  1. Enable `nsfs` bind-mount tests on Asterinas.
  2. Check only the destination mount namespace for bind mounts.
  3. Refactor the mount-loop check helper.
  4. Align mount-namespace `nsfs` bind-mount behavior with Linux.

### DIFF
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -7,6 +7,7 @@ use core::time::Duration;
 pub(in crate::fs) use dentry::Dentry;
 use dentry::DirDentry;
 use inherit_methods_macro::inherit_methods;
+use mount::MountNsFileCopying;
 pub use mount::{Mount, MountPropType, PerMountFlags};
 pub use mount_namespace::MountNamespace;
 pub use resolver::{AT_FDCWD, AbsPathResult, FsPath, LookupResult, PathResolver, SplitPath};
@@ -16,6 +17,7 @@ use crate::{
         file::{
             CreationFlags, InodeHandle, InodeMode, InodeType, OpenArgs, Permission, StatusFlags,
         },
+        pseudofs::NsInode,
         vfs::{
             file_system::{FileSystem, FsFlags},
             inode::{Inode, Metadata, MknodType},
@@ -242,6 +244,13 @@ impl Path {
     }
 }
 
+fn try_get_mnt_ns_inode(dentry: &Dentry) -> Option<&NsInode<MountNamespace>> {
+    dentry
+        .inode()
+        .as_ref()
+        .downcast_ref::<NsInode<MountNamespace>>()
+}
+
 impl Path {
     /// Mounts a filesystem at the current path.
     ///
@@ -353,9 +362,12 @@ impl Path {
     ///
     /// # Errors
     ///
-    /// Returns `ENOTDIR` if the `dst_path` is not a directory.
-    /// Returns `EINVAL` if either source or destination path is not in the
-    /// current mount namespace.
+    /// Returns `ENOTDIR` if one of the source and destination is a directory
+    /// and the other is not.
+    ///
+    /// Returns `EINVAL` if any of the following holds:
+    /// - The destination path is not in the current mount namespace.
+    /// - The source path is a mount namespace file that would create a namespace loop.
     pub fn bind_mount_to(&self, dst_path: &Self, recursive: bool, ctx: &Context) -> Result<()> {
         let can_bind = {
             let src_is_dir = self.type_() == InodeType::Dir;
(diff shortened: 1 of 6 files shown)


kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -54,6 +55,11 @@ impl MountNamespace {
         &self.root
     }
 
+    /// Returns the creation sequence of this mount namespace.
+    pub(in crate::fs) fn seq(&self) -> u64 {
+        self.stashed_dentry.ino()
+    }
```
cchanging: Can we implement `PartialEq` for `MountNameSpace` to avoid expose this?

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -22,6 +23,20 @@ use crate::{
...
+        .inode()
+        .as_ref()
+        .downcast_ref::<NsInode<MountNamespace>>()
+        .is_some()
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) enum MountNsFileCopying {
+    Copy,
+    Skip,
```
cchanging, line 32: Add documents for this to explain the detail semantics for each.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -236,6 +238,44 @@ impl Path {
...
+            return_errno_with_message!(
+                Errno::ELOOP,
+                "the mount tree contains a mount namespace file that would create a namespace loop"
+            );
+        }
+        let children: Vec<_> = {
+            let children = mount.children.read();
+            children.values().cloned().collect()
+        };
+        worklist.extend(children);
```
cchanging: Can we use `extend_from_slice` to avoid creating a new `Vec`?

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -236,6 +238,44 @@ impl Path {
...
+/// Ensures that a mount subtree contains no mount-namespace loop.
+fn check_no_mnt_ns_loop_in_tree(
+    root_mount: &Arc<Mount>,
+    current_mnt_ns: &Arc<MountNamespace>,
+) -> Result<()> {
+    let mut worklist = VecDeque::new();
+    worklist.push_back(root_mount.clone());
+
+    while let Some(mount) = worklist.pop_front() {
+        if is_mnt_ns_loop(mount.root_dentry(), current_mnt_ns) {
```
cchanging: It seems that we obtain the mount namespace of the `mount` here in every iteration, I think we can do some optimization here.

cchanging (commented):
Thanks for handling the case of performing a bind operation on a `procfs` file corresponding to a mount namespace! This is indeed a very unusual operation, and the changes in this PR make sense to me.

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -27,6 +28,26 @@ pub struct MountNamespace {
     stashed_dentry: StashedDentry,
 }
 
+impl PartialEq for MountNamespace {
```
cchanging, line 31: With this trait, you can re-implement the `owns` method of `MountNamespace` to avoid its atomic operation.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -22,6 +23,23 @@ use crate::{
     prelude::*,
 };
 
+fn is_mnt_ns_file(dentry: &Dentry) -> bool {
```
cchanging: we can rename the method something like `try_get_mnt_ns_inode` and let it return `Option<&Arc<NsInode<MountNamespace>>>`, then this method can also be used in `is_mnt_ns_loop`?

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -242,6 +244,46 @@ impl Path {
...
+    let inode = dentry.inode().as_ref();
+    let Some(mnt_ns_inode) = inode.downcast_ref::<NsInode<MountNamespace>>() else {
+        return false;
+    };
+
+    current_mnt_ns.as_ref() >= mnt_ns_inode.ns().as_ref()
+}
+
+/// Ensures that a mount subtree contains no mount-namespace loop.
+fn check_no_mnt_ns_loop_in_tree(
```
cchanging: Can we make this as a method of `MountNamespace`? The semantics can be more clear in this way.

test/initramfs/src/regression/security/namespace/proc_nsfs.c
```diff
@@ -361,6 +361,145 @@ END_TEST()
...
+	if (mount_ret == 0)
+		TEST_SUCC(umount(bind_mount_path));
+
+	struct clone_args args = {
+		.flags = CLONE_NEWNS,
+		.exit_signal = SIGCHLD,
+	};
+	pid_t child = TEST_SUCC(sys_clone3(&args));
+	if (child < 0)
+		goto out_unlink;
```
cchanging: Such `goto` pattern is unnecessary·

cchanging (approved):
LGTM! Only some minor comments left.

cqs21 (approved):

PR #3036, opened 2026-03-15
Introduce Klint's `not_using_prelude` check

zjp-CN:
The PR
* applies KLint to Asterinas, and fixes all the not_using_prelude warnings
* adds `--klint` argument for `cargo osdk check`
* declares `Result` in prelude module with default type parameter `E`
  * I can remove this breaking change, because it's unnecessary for the fix

The code requires asterinas container v0.17.2-20260508 to work.

Also see https://github.com/asterinas/asterinas/issues/2967 for the context.

<details><summary>Click to expand and view sample klint diagnostics suggesting that imports can be replaced by the prelude modules of `ostd`, `rand`, and `bitvec`.
</summary>
<p>

Full warnings are [here](https://github.com/os-checker/asterinas/actions/runs/23113661880/job/67135354514).

```rust
    Checking aster-virtio v0.1.0 (/asterinas/kernel/comps/virtio)
warning: this item is available via prelude
  --> kernel/comps/virtio/src/device/block/device.rs:26:10
   |
26 |     mm::{HasSize, VmIo, dma::DmaStream},
   |          ^^^^^^^
   |
   = help: import with `ostd::prelude::*` instead
   = note: `#[warn(klint::not_using_prelude)]` on by default

warning: this item is available via prelude
 --> kernel/comps/virtio/src/dma_buf.rs:8:15
  |
8 |     HasDaddr, HasSize,
  |               ^^^^^^^
  |
  = help: import with `ostd::prelude::*` instead

warning: this item is available via prelude
  --> kernel/comps/virtio/src/queue.rs:15:16
   |
15 | use ostd::mm::{HasPaddr, PodOnce, Split, dma::DmaCoherent};
   |                ^^^^^^^^
   |
   = help: import with `ostd::prelude::*` instead

warning: this item is available via prelude
 --> kernel/comps/virtio/src/transport/mmio/bus/common_device.rs:8:12
  |
8 |     Error, Result,
  |            ^^^^^^
  |
  = help: import with `ostd::prelude::*` instead

warning: this item is available via prelude
  --> kernel/comps/virtio/src/transport/mmio/bus/common_device.rs:11:10
   |
11 |     mm::{HasPaddr, VmIoOnce},
   |          ^^^^^^^^
   |
   = help: import with `ostd::prelude::*` instead

warning: `aster-virtio` (lib) generated 5 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.67s
Checking kernel/libs/aster-bigtcp
[...]

PR #3083, opened 2026-04-02
Install KLint in docker and bump image version to 0.17.2-20260508

zjp-CN:
cc https://github.com/asterinas/asterinas/pull/3036#issuecomment-4065699743

This PR
* installs klint 
* bumps docker image to 0.17.2-20260508


osdk/tools/docker/Dockerfile
```diff
@@ -166,6 +166,9 @@ RUN cargo install --locked \
     mdbook-mermaid@0.17.0 \
     typos-cli@1.39.0
 
+# Install klint-asterinas a forked klint.
+Run cargo install klint-asterinas --git https://github.com/os-checker/klint-asterinas --rev 40952a7
```
tatetian: > Yes. Klint's not_using_prelude was introduced on Jan 24 in https://github.com/Rust-for-Linux/klint/commit/8632dfad6a7b41cc54303ee40990f4914a28ffe4 , but the code doesn't compile on Asterinas' current Rust toolchain:

Is it possible to upgrade the version of Asterinas's Rust nightly toolchain to a more recent version compatible with klint? I am redundant to fork klint and maintain it.

If maintaining a fork is truly necessary, then this fork repo must be transferred to the `asterinas` Github org. As I understand it, this `os-checker/klint-asterinas` fork is specifically for Asterinas. If the Asterinas project needs to add a new dependency, I want to make sure that the dependency is either maintained by the Asterinas community itself or has a large enough user base so that the dependency will continue to be well maintained in the foreseeable future.

zjp-CN: I've tested that the original klint[^1] is compatible with Asterinas on nightly-2026-03-19, also submitted the toolchain update [PR #3089](https://github.com/asterinas/asterinas/pull/3089).

[^1]: But klint currently has a [bug](https://github.com/Rust-for-Linux/klint/pull/15), so need to wait for it to be fixed.

tatetian: In this PR, you can (1) upgrade the Rust toolchain installed in the Docker image, (2) install klint, and (3) bump the Docker image version. #3089 can then switch to the new Docker image with the more recent Rust toolchain.

tatetian (approved):

osdk/tools/docker/Dockerfile
tatetian: Sorry, I was wrong. The CI tests failed because the toolchain is upgraded but the code is not. It seems that we have to include the changes to the code in this PR.

junyang-zh:
Please merge #3144 before this. Otherwise, we lost the version 0.17.2

tatetian (approved):

zjp-CN:
Sorry, need to install sqlite3 library in the docker to fix the failed image publication [run](https://github.com/asterinas/asterinas/actions/runs/25650709238/job/75288664960).
I'll put forward a PR to fix it.

PR #3089, opened 2026-04-02
Adapt to rust-toolchain nightly-2026-04-03

zjp-CN:
cc https://github.com/asterinas/asterinas/pull/3083#discussion_r3026566584

The motivation of the rust-toolchain update is to use the [latest  klint](https://github.com/Rust-for-Linux/klint/commit/7d7c522b66a3e2456bc6024ff69dea619c4b9c2a) on Asterinas.

Changes include:
* bumping the toolchain channel field
* removing unused features which is triggered as a part of the unused code lint
* fixing [collapsible_match](https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_match)
* removing the unnecessary `.into_iter` which is warned against now
* using  absolute module path to `format_args_nl!` which is [hidden now](https://github.com/rust-lang/rust/pull/146314/changes)
* fixing the unnecessary unsafe block when calling `core::arch::x86_64::__cpuid_count` that is [safe now](https://github.com/rust-lang/rust/issues/152374)
* lifting the visibility of Dentry to fs scope due to a rising compilation error
  * the [SysTreeInodeTy](https://github.com/asterinas/asterinas/blob/318f374a770677ce971d6c7ee16f3f83083566eb/kernel/src/fs/utils/systree_inode.rs#L39) trait has that wider visibility, and its method new_dentry_iter [returns](https://github.com/asterinas/asterinas/blob/318f374a770677ce971d6c7ee16f3f83083566eb/kernel/src/fs/utils/systree_inode.rs#L262) a iterator of Dentry which was only acceesible in a narrow scope
* fixing the missing `-Ztarget-spec-json` 
  * the cargo flag was required since https://github.com/rust-lang/cargo/pull/16557  
* ignoring some pedantic warnings by adding `#[expect]`


StevenJiang1110:
Please open a separate PR to update the toolchain in our Docker image first, and then submit this PR to update the Rust toolchain in the project. Note that our Docker image already has the project toolchain installed, so both changes need to stay in sync.

zjp-CN:
I've submitted a PR to update toolchain and docker image in https://github.com/asterinas/asterinas/pull/3083 whose CI must fail without these changes...

StevenJiang1110:
> I've submitted a PR to update toolchain and docker image in #3083 whose CI must fail without these changes...

Oh, sorry, I missed the context in #3083....I only notice it installs klint, but haven't noticed it also upgrades toolchain.

zjp-CN:
I skimmed the output, but don't get the failure in the [failed conformance test](https://github.com/asterinas/asterinas/actions/runs/25481225035/job/74765612825?pr=3089):

```
All conformance tests passed.
Conformance test failed
make: *** [Makefile:262: run_kernel] Error 1
Error: Process completed with exit code 2.
```

What causes the error?

lrh2000:
> What causes the error?

It is caused by https://github.com/asterinas/asterinas/issues/3181.

kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
lrh2000, line 529: I think the `explicit_counter_loop` lint is suggesting that you use something like the following, right?
```patch
diff --git a/kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs b/kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
index a8b981721..7043514da 100644
--- a/kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
+++ b/kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
@@ -488,17 +488,16 @@ impl<L: BlockLog> MhtStorage<L> {
         let num_append = nodes.len();
         let mut node_entries = Vec::with_capacity(num_append);
         let mut cipher_buf = Buf::alloc(num_append)?;
-        let mut pos = self.block_log.nblocks() as BlockId;
-        let start_pos = pos;
+        let start_pos = self.block_log.nblocks() as BlockId;
         for (i, node) in nodes.iter().enumerate() {
             let plain = node.as_bytes();
             let cipher = &mut cipher_buf.as_mut_slice()[i * BLOCK_SIZE..(i + 1) * BLOCK_SIZE];
             let key = Key::random();
             let mac = Aead::new().encrypt(plain, &key, &Iv::new_zeroed(), &[], cipher)?;
 
+            let pos = start_pos + i;
             node_entries.push(MhtNodeEntry { pos, key, mac });
             self.node_cache.put(pos, node.clone());
-            pos += 1;
         }
 
         let append_pos = self.block_log.append(cipher_buf.as_ref())?;
```

osdk/src/commands/build/bin.rs
```diff
@@ -187,6 +187,11 @@ fn install_setup_with_arch(
         cmd.arg("--version").arg(env!("CARGO_PKG_VERSION"));
     }
     cmd.arg("--target").arg(arch);
+    // Custom json target requires this cargo flag.
+    // cc https://github.com/rust-lang/cargo/pull/16557
```
lrh2000: Actually it is better to link the tracking issue or the doc instead of linking to the PR..
```suggestion
    // https://doc.rust-lang.org/cargo/reference/unstable.html#target-spec-json
```
or:
```suggestion
    // https://github.com/rust-lang/rust/issues/151528
```

osdk/src/commands/new/mod.rs
```diff
@@ -228,7 +228,10 @@ fn check_rust_toolchain(toolchain: &toml::Table) {
     let expected_channel = expected.get("channel").unwrap().as_str().unwrap();
 
     if channel != expected_channel {
-        error_msg!("The current version of rust-toolchain.toml is not compatible with the osdk");
+        error_msg!(
+            "The current version of rust-toolchain.toml `{channel}` \
+             is not compatible with the osdk's `{expected_channel}`"
```
lrh2000:
```suggestion
             is not compatible with the OSDK's `{expected_channel}`"
```
(though it's not this PR's fault)

zjp-CN:
Given the failure is not caused by this PR, I think the code is ready to review.

Should I squash all commits into single one?

lrh2000 (commented):
Thanks for updating the toolchain version. Could you rebase this PR and squash commits that do the same thing into one commit?

For example,
 - Allow explicit_counter_loop in append_mht_nodes 
 - Expect explicit_counter_loop
 - Expect explicit_counter_loop iin append_data_nodes

There are quite a few others:
 - Remove unused features from ostd
 - Fix unused features by gating them under cfg

And:
 - Force installing osdk to override old osdk in CI
 - Install the toolchain components as early as possible

And:
 - Expect unused_assignments due to the lock guard
 - Expect unused_assignments

Maybe others. But I'm too lazy to copy their commit messages.

lrh2000:
> Should I squash all commits into single one?

I think a single one would work too. This PR is much more easy to review by just checking the whole diff. (I left my comment without seeing this comment so I did go through all individual commits.)

**EIDT:** If possible, it may be reasonable to split some CI changes (and some unrelated OSDK changes) that are not directly related to the specific toolchain update (i.e., `2026-04-03` toolchain update) to a separate commit.

kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
zjp-CN: The clippy suggests a less readable snippet here

```
error: the variable `pos` is used as a loop counter
   --> kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs:493:9
    |
493 |         for (i, node) in nodes.iter().enumerate() {
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `for (pos, (i, node)) in (self.block_log.nblocks() as BlockId..).zip(nodes.iter().enumerate())`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#explicit_counter_loop
    = note: `-D clippy::explicit-counter-loop` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::explicit_counter_loop)]`
```

so I just ignored the lint.

lrh2000: The clippy suggestion is indeed wrose than ignoring the lint...

kernel/src/security/tsm_mr.rs
```diff
@@ -205,6 +205,7 @@ inherit_sys_leaf_node!(Measurement, fields, {
         SysPerms::DEFAULT_RW_PERMS
     }
 
+    #[expect(unused_assignments)]
```
lrh2000: I was thinking about... Another approach is:
```rust
#[expect(unused_assignments)]
{ in_sync = in_sync_write.downgrade() };
```

It's not very elegant, either (due to the unnecessary brackets). But it's clear what we're expecting.

zjp-CN: Good idea! I'll do that.

junyang-zh:
A version bump is long overdue. @tatetian can you check #3144 ? Otherwise, this PR may be blocked.

kernel/src/security/tsm_mr.rs
zjp-CN: Or is `#![feature(stmt_expr_attributes)]` acceptable in this crate? We can attach the `expect` macro to the expression with that feature. I don't have a preference for both fixes.

kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
```diff
@@ -490,6 +490,8 @@ impl<L: BlockLog> MhtStorage<L> {
         let mut cipher_buf = Buf::alloc(num_append)?;
         let mut pos = self.block_log.nblocks() as BlockId;
         let start_pos = pos;
+        // Ignore the lint because clippy suggests a less readable fix.
+        #[expect(clippy::explicit_counter_loop)]
```
lrh2000: Well, the fact is that Clippy's suggestion is poor. I think my suggestion is better than using this `#[expect]`, and it won't trigger Clippy's warnings?

The problem is that Clippy isn't smart enough to recognize the relationship between `i` and `pos`. However, we are smarter than Clippy and can correctly recognize this relationship, so the code can be written in a more reasonable way with fewer mutable variables.

kernel/src/security/tsm_mr.rs
```diff
@@ -217,7 +217,11 @@ inherit_sys_leaf_node!(Measurement, fields, {
             let mut in_sync_write = in_sync.upgrade();
             let mr = tdxguest::get_tdx_mr_refresh(attr.reg);
             *in_sync_write = true;
-            in_sync = in_sync_write.downgrade();
+            // Ignore the lint here due to the value being a lock guard.
+            #[expect(unused_assignments)]
```
lrh2000:
```suggestion
            #[expect(unused_assignments, reason = "the value is a lock guard")]
```

kernel/src/security/tsm_mr.rs
```diff
@@ -217,7 +217,11 @@ inherit_sys_leaf_node!(Measurement, fields, {
...
+            {
+                in_sync = in_sync_write.downgrade()
```
lrh2000:
```suggestion
                in_sync = in_sync_write.downgrade();
```

kernel/comps/mlsdisk/src/layers/1-crypto/crypto_log.rs
zjp-CN: 😂You're right. Thanks!

.github/actions/test/action.yml
```diff
@@ -100,6 +100,8 @@ runs:
       shell: bash
       run: |
         [[ "${{ !contains(matrix.container, 'osdk') }}" == "true" ]] && make check_osdk
+        # Install toolchain and components as early as possible to avoid rustup's conflict problem.
+        rustup show
```
zjp-CN: I doubt this command is necessary, because `make check_osdk` has fully installed components.

The fix occurred due to a toolchain mismatch when I hadn't fixed all the stuff.

I'm trying to see if it still works without the command.

zjp-CN: It works. I'll remove this.

lrh2000 (approved):

tatetian (approved):

PR #3095, opened 2026-04-07
Update book to document init fallback paths when init parameter is omitted

taosue:


book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -15,7 +15,7 @@ init=/bin/busybox
 
 Notes:
 - The value is the path to the executable.
-- If omitted, Asterinas will fail to boot.
+- If omitted, Asterinas will try to execute from the following paths in order: `/sbin/init`, `/etc/init`, `/bin/init`, `/bin/sh`.
```
tatetian: To conform with semantics line break convention, break the line after `in order:`

tatetian (commented):
Thanks for updating the doc!

tatetian (approved):

PR #3107, opened 2026-04-08
Optimize make check speed

StevenJiang1110:
Currently, running `make check` takes a significant amount of time, often exceeding 3 minutes. This PR reduces the execution time of the main `make check` hotspots from 225.354s to 49.721s, a 77.9% reduction based on local measurements. (The percentage reduction in CI may be somewhat smaller due to the additional time spent installing OSDK, but the absolute time savings should be roughly the same.)

This PR consists of 4 commits:

1. The first optimization addresses non-osdk crates, where `cargo clippy --test <crate_name>` and `cargo clippy <crate_name>` were previously invoked separately. These are now replaced with a single `cargo clippy --lib --tests <crate_name>` call, avoiding repeated cargo invocations. This saves 7.424s.
2. The second commit contains the main optimization. The root cause is that `cargo osdk clippy` would by default invoke `cargo clippy` under every crate in the current workspace. Since we were already iterating over all crates in the Makefile, this resulted in a redundant double loop. After this fix, `cargo osdk clippy` serves purely as a forwarding call. This change saves 161.268s in local testing.
3. The third optimization addresses the fact that `make check` was previously building the initramfs, which is unnecessary. It should only depend on `exfat.img`. This is a pre-existing issue in our codebase — ideally, the dependency on `exfat.img` could also be removed in the future, but for now it has been changed to only build `exfat.img`. This saves 6.941s.

https://github.com/asterinas/asterinas/blob/965566496420a6e96a1955bded88ef0b2a178cce/kernel/src/fs/fs_impls/exfat/mod.rs#L124

4. The fourth commit is a pure refactor, extracting the clippy-related logic from the Makefile into a dedicated script.


tools/clippy_check.sh
```diff
@@ -0,0 +1,188 @@
...
+
+    split_crate_list "${NON_OSDK_CRATES:-}" non_osdk_crates
+    split_crate_list "${OSDK_CRATES:-}" osdk_crates
+
+    local crate_dir
+    for crate_dir in "${non_osdk_crates[@]}"; do
+        echo "Checking ${crate_dir}"
+        (
+            cd "$PROJECT_ROOT/$crate_dir"
+            # Check both non-test code and test code in one clippy run.
```
tatetian: **Target coverage narrowed from default to `--lib --tests`**: The old Makefile ran `cargo clippy --no-deps` (which checks all default targets: lib, bins, examples) followed by `cargo clippy --tests --no-deps`. The new code runs `cargo clippy --lib --tests --no-deps`, which explicitly excludes `[[bin]]` and `[[example]]` targets.

Today this does not cause a real regression because the only non-OSDK crate with a `[[bin]]` target (`kernel/libs/logo-ascii-art`) gates it behind `required-features = ["color"]`, so the old code silently skipped it too. However, the explicit `--lib` flag establishes a new contract: any future `[[bin]]` or `[[example]]` target added to a non-OSDK crate will be silently excluded from `make check` with no warning.

Consider using `--all-targets` instead of `--lib --tests`. This is equivalent to the old two-pass approach in a single invocation and covers bins and examples as well.

tools/clippy_check.sh
```diff
@@ -0,0 +1,188 @@
...
+run_check_osdk() {
+    ensure_command cargo
+
+    echo "Checking osdk"
+    (
+        cd "$PROJECT_ROOT/osdk"
+        RUSTFLAGS="-Dwarnings" cargo clippy --no-deps
+        RUSTFLAGS="-Dwarnings" cargo clippy --tests --no-deps
+    )
+}
```
tatetian, line 57: **Two-pass pattern not consolidated for the `osdk` crate**: Commit #1 eliminated the two-pass pattern for non-OSDK crates by combining into `--lib --tests`. The `osdk` crate is a binary package (no `[lib]` target), so `--lib` wouldn't apply, but `--all-targets --no-deps` would achieve the same consolidation here, halving the number of cargo invocations for the `osdk` crate.

Makefile
```diff
@@ -300,6 +298,16 @@ check_vdso:
 initramfs: check_vdso
 	@$(MAKE) --no-print-directory -C test/initramfs
 
+.PHONY: check_test_assets
```
tatetian: **`check_test_assets` is an unnecessary workaround; remove the exfat ktests instead**: The sole reason `make check` depends on any initramfs asset is the `include_bytes!("../../../../../test/initramfs/build/exfat.img")` in `kernel/src/fs/fs_impls/exfat/mod.rs:124`, which backs an in-memory block device for the exfat ktest module. Rather than introducing a new Makefile target (plus a corresponding `check_assets` target in `test/initramfs/Makefile`) to build just this one artifact, consider removing the exfat unit tests entirely. The project already has more comprehensive integration tests for exfat, and no other filesystem implementation (e.g., Ext2) carries ktest-only `include_bytes!` dependencies on build artifacts. Removing the tests would let `make check` drop all initramfs-related dependencies, eliminating this workaround and the associated TODO.

tatetian (commented):
Great job on identifying the double-loop overhead as the main bottleneck in `make check` and eliminates it. However, the current approach works around the problem with elaborate shell scripting (manually constructed crate lists, `cargo metadata | jq` pipelines, env-var-based interfaces) when a simpler structural fix is available.

**The root cause of the complexity is that `cargo osdk clippy` doesn't work from the workspace root because the workspace mixes OSTD-based kernel crates with proc-macro and host-native crates that are incompatible with the bare-metal `--target` and `-Zbuild-std` flags OSDK injects.** This is a pre-existing bug — it fails on both `main` and this branch.

**Recommended approach**: Add a `default-members` field to the root `Cargo.toml` listing only the OSTD-based kernel crates. This fixes `cargo osdk clippy` at the workspace level and enables a much simpler `make check` structure with three tiers:

1. **OSDK crates** (default workspace members): `cargo osdk clippy -- --no-deps` and `cargo osdk clippy --ktests -- --no-deps` from the project root. No crate list construction needed.
2. **Non-OSDK workspace crates** (members minus default members, i.e., proc-macro and host-native crates): `cargo clippy -p <packages> --no-deps` where the package list is derived automatically from `cargo metadata` (`workspace_members - workspace_default_members`). No manually maintained list needed.
3. **Standalone crates** not in the workspace (e.g., `osdk/`): `cd <crate> && cargo clippy --no-deps`.

One special case remains: `ostd/libs/linux-bzimage/setup` is x86_64-only and depends on a crate-local `.cargo/config.toml`, so it must be excluded from `default-members` and checked separately on x86_64.

This approach eliminates `NON_OSDK_CRATES` / `OSDK_CRATES` as Makefile variables, removes the `build_workspace_package_args` / `cargo metadata | jq` machinery, and vastly simplifies `clippy_check.sh`.

**This generalizes beyond `check`/`clippy`.** The same `default-members` fix benefits other Makefile targets that use the same per-crate iteration pattern:

- **`docs`**: Currently has two per-crate loops (`cargo doc` for non-OSDK, `cargo osdk doc` for OSDK). With `default-members`, the OSDK loop collapses to a workspace-level `cargo osdk doc` and the non-OSDK loop to `cargo doc -p <computed-set>`. One remaining special case: the `kernel` crate needs `--document-private-items`, requiring a separate pass.
- **`test`**: Currently loops over `NON_OSDK_CRATES` with `cargo test`. Collapses to `cargo test -p <computed-set>`.
- **`ktest`**: Fundamentally per-crate (each crate boots in a separate QEMU instance), but it can derive its crate list from `default-members` instead of the hardcoded `OSDK_CRATES` variable.
- **`format`**: Already optimal — uses a single `cargo fmt` at the workspace root.

A small helper script that computes `workspace_members - workspace_default_members` from `cargo metadata` can be shared across `check`, `docs`, and `test`, eliminating the manually maintained crate lists from the Makefile entirely.

Additionally, the `check_test_assets` workaround should be removed by deleting the exfat ktests instead (see comment below).

Makefile
```diff
@@ -395,34 +340,42 @@ profile_client: initramfs $(CARGO_OSDK)
...
-		(cd $$dir && RUSTDOCFLAGS="-Dwarnings" cargo doc --no-deps) || exit 1; \
-	done
-	@for dir in $(OSDK_CRATES); do \
-		EXTRA_DOC_FLAGS=""; \
-		# The kernel crate is primarily composed of private items. \
-		# We include the --document-private-items flag \
-		# to ensure documentation of the internal items is fully checked. \
-		if [ "$$dir" = "kernel" ]; then \
-			EXTRA_DOC_FLAGS="--document-private-items -Arustdoc::private_intra_doc_links"; \
+	@mapfile -t default_package_args < <(./tools/workspace_members.sh default package-args); \
```
tatetian, line 357: **`make docs` is broken — `default` is not a valid member set**: `tools/workspace_members.sh` only accepts `members` or `non-default` (see `case` at lines 107–114). Invoking it with `default` exits 1, but `mapfile < <(...)` silently discards the exit status, so `default_package_args` ends up empty, the guard evaluates to false, and the `cargo osdk doc` pass over default members never runs.

See the `workspace_members.sh` interface comment below for a suggested redesign that fixes this bug and the position-indexed filtering in one go. The short version: rename to `print_workspace_members.sh`, default to printing all members as paths, and add `--default-ones` / `--non-default-ones` / `--package-names` flags. Then the `docs` target becomes:

```makefile
DEFAULT_PKGS := $(shell ./tools/print_workspace_members.sh --default-ones --package-names)
# ...
docs:
	@RUSTDOCFLAGS="-Dwarnings" cargo osdk doc $(addprefix -p ,$(filter-out aster-kernel,$(DEFAULT_PKGS))) --no-deps
```

tools/workspace_members.sh
```diff
@@ -0,0 +1,129 @@
...
+
+print_workspace_members() {
+    local member_set="$1"
+    local output_kind="$2"
+    local jq_program
+    local workspace_excludes_json
+
+    ensure_command cargo
+    ensure_command jq
+    ensure_command tomlq
```
tatetian, line 45: **`tomlq` is not installed in the main Dockerfile**: `tools/workspace_members.sh` hard-depends on `tomlq` (Python `yq` package), but `tools/docker/Dockerfile` only installs `jq`. Every target that calls `workspace_members.sh` will abort in CI.

Mirror the `pipx install yq==3.4.3` block from `osdk/tools/docker/Dockerfile:12-17`, or switch to a toolchain already installed. The `workspace.exclude` filter (line 47) is load-bearing — `cargo metadata --no-deps` includes 3 `comp-sys` path deps despite `[workspace].exclude` — but `tomlq` is overkill for a 4-entry static list. Cheaper alternatives: hardcode the exclude list, use `python3 -c "import tomllib; ..."`, or filter by the actual invariant (crate lacks `[lints] workspace = true`).

Makefile
```diff
@@ -395,34 +340,42 @@ profile_client: initramfs $(CARGO_OSDK)
 
 .PHONY: test
 test:
-	@for dir in $(NON_OSDK_CRATES); do \
-		(cd $$dir && cargo test) || exit 1; \
-	done
+	@mapfile -t non_default_package_args < <(./tools/workspace_members.sh non-default package-args); \
```
tatetian, line 343: **`mapfile <(...)` silently swallows helper-script failures (root cause of the `make docs` bug)**: This pattern recurs in `test`, `docs`, `check`, and `clippy_check.sh`. If `workspace_members.sh` aborts, the outer shell gets an empty array and the `if (( N > 0 ))` guard silently skips work. The `default` bug and the missing `tomlq` are both manifestations.

Fix: define Make variables with `$(shell ...)` at parse time — just like the old `NON_OSDK_CRATES`/`OSDK_CRATES` but computed from `cargo metadata`. See the `workspace_members.sh` interface comment below for a concrete example.

tools/workspace_members.sh
```diff
@@ -0,0 +1,129 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+PROJECT_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")"
+WORKSPACE_MANIFEST="${PROJECT_ROOT}/Cargo.toml"
+LINUX_BZIMAGE_SETUP_DIR="ostd/libs/linux-bzimage/setup"
+
+usage() {
+    cat <<'EOF'
```
tatetian, line 12: **The interface is not self-describing**: `workspace_members.sh non-default package-args` forces readers to guess what "non-default" and "package-args" mean. `workspace_members.sh members dirs` stutters with the script name. The `package-args` output bakes `-p <name>` cargo conventions into a generic utility, forcing position-indexed array surgery at call sites (Makefile:357–367). The `non-default` set silently hardcodes a `linux-bzimage/setup` exclusion (line 75), leaking kernel knowledge into the utility.

Suggested redesign — rename to `print_workspace_members.sh`:

```
Usage: ./tools/print_workspace_members.sh [--default-ones | --non-default-ones] [--package-names]

Print the (relative) paths of all Cargo workspace members.

Options:
    --default-ones      Only print default workspace members.
    --non-default-ones  Only print non-default workspace members.
    --package-names     Print package names rather than paths.
```

No flags = all members as paths. Callers compose with standard Make idioms (`$(addprefix -p ,...)`, `$(filter-out ...)`). No hidden exclusions — `linux-bzimage/setup` filtering moves to call sites. Combined with `$(shell ...)` Make variables (see the `mapfile` comment above):

```makefile
DEFAULT_PKGS     := $(shell ./tools/print_workspace_members.sh --default-ones --package-names)
NON_DEFAULT_PKGS := $(shell ./tools/print_workspace_members.sh --non-default-ones --package-names)
MEMBER_DIRS      := $(shell ./tools/print_workspace_members.sh)

docs: $(CARGO_OSDK)
	@RUSTDOCFLAGS="-Dwarnings" cargo osdk doc \
		$(addprefix -p ,$(filter-out aster-kernel,$(DEFAULT_PKGS))) --no-deps
	@RUSTDOCFLAGS="-Dwarnings --document-private-items ..." \
		cargo osdk doc -p aster-kernel --no-deps
	@RUSTDOCFLAGS="-Dwarnings" cargo doc \
		$(addprefix -p ,$(NON_DEFAULT_PKGS)) --no-deps
```

Cargo.toml
```diff
@@ -7,6 +7,9 @@
 
 [workspace]
 resolver = "2"
+# When adding a new workspace crate that depends on OSTD, include it in
```
tatetian, line 10: [**`semantic-line-breaks`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#semantic-line-breaks): Several comments in this PR break lines mid-phrase rather than at semantic boundaries. Instances:

- `Cargo.toml:10-12` — breaks "include it in / `default-members`"
- `Makefile:351-352` — breaks "linux-bzimage-setup is / left out"
- `tools/clippy_check.sh:81-82` — breaks "may fail on / other architectures"

Also, the `Cargo.toml` comment only explains the criterion for `default-members`. A reader seeing three lists (`members`, `default-members`, `exclude`) would benefit from understanding the purpose of each:

```toml
# All in-kernel, `no_std` crates.
#
# These crates share a single Cargo workspace
# so that their `Cargo.toml` files do not have to repeat common configurations.
#
# Some of these crates depend on OSDK, some do not.
# If they do, they also belong to `default-members`.
members = [
    ...
]
# All OSDK-based, in-kernel crates.
#
# For these crates, use `cargo osdk <cmd>`.
default-members = [
    ...
]
# All user-space crates, including the OSDK tool.
#
# For these crates, use `cargo <cmd>` directly.
exclude = [
    ...
]
```

tatetian (commented):
The PR has been substantially reworked to adopt the `default-members` approach recommended in the previous review.

However, the rework introduced one critical bug that breaks `make docs`, plus one CI-breaking dependency gap:**

- **`make docs` is broken**: the target calls `workspace_members.sh default ...` but the script only accepts `members` or `non-default` — the `default` keyword is unsupported. The documentation pass for default-member crates is silently skipped.
- **`tomlq` is a new, undeclared CI dependency**: the main `tools/docker/Dockerfile` installs only `jq`, not the Python `yq` package. `make check`/`test`/`docs` will fail in CI with "required command 'tomlq' is not installed".

In addition, the `workspace_members.sh` interface is hard to read at call sites and the approach relies on bash `mapfile` via process substitution which looks ugly. A cleaner interface of `workspace_members.sh` is proposed. So is an alternative to `mapfile`.

tools/workspace_members.sh
StevenJiang1110: tomlq is actually available in our container environment. Although it is not explicitly listed in the Dockerfile, it was likely pulled in as an implicit dependency when installing certain other packages. That is why the previous CI runs did not report any errors, and I have also verified locally that tomlq is indeed available by default.

That said, your suggested approach is also valid — using Python's built-in tomllib is a more explicit and reliable choice. So I have updated the implementation to use tomllib instead.

tatetian (approved):

PR #3109, opened 2026-04-09
Support cgroup namespace

cchanging:
This PR introduces cgroup namespaces, which allow a task to observe the unified cgroup hierarchy from a namespace-specific root while preserving its underlying cgroup membership. 

This functionality is required by #2911.

### DIFF
--- a/kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
@@ -1,0 +1,149 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use alloc::sync::Arc;
+
+use aster_systree::SysBranchNode;
+use spin::Once;
+
+use crate::{
+    fs::{
+        cgroupfs::{CgroupNode, CgroupSysNode, CgroupSystem},
+        pseudofs::{NsCommonOps, NsType, StashedDentry},
+    },
+    prelude::*,
+    process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::PosixThread},
+};
+
+/// The cgroup namespace for the unified cgroup hierarchy.
+pub struct CgroupNamespace {
+    root: Arc<dyn CgroupSysNode>,
+    owner: Arc<UserNamespace>,
+    stashed_dentry: StashedDentry,
+}
+
+impl CgroupNamespace {
+    /// Returns a reference to the singleton initial cgroup namespace.
+    pub fn get_init_singleton() -> &'static Arc<Self> {
+        static INIT: Once<Arc<CgroupNamespace>> = Once::new();
+
+        INIT.call_once(|| {
+            Arc::new(Self {
+                root: CgroupSystem::singleton().clone(),
+                owner: UserNamespace::get_init_singleton().clone(),
+                stashed_dentry: StashedDentry::new(),
+            })
+        })
+    }
+
+    /// Creates a new cgroup namespace rooted at the given cgroup.
+    pub fn new_clone(
+        current_cgroup: Option<Arc<CgroupNode>>,
+        owner: Arc<UserNamespace>,
+        posix_thread: &PosixThread,
+    ) -> Result<Arc<Self>> {
+        owner.check_cap(CapSet::SYS_ADMIN, posix_thread)?;
+
+        let root: Arc<dyn CgroupSysNode> = match current_cgroup {
+            Some(current_cgroup) => current_cgroup,
+            None => CgroupSystem::singleton().clone(),
+        };
+
+        Ok(Arc::new(Self {
+            root,
+            owner,
+            stashed_dentry: StashedDentry::new(),
+        }))
+    }
+
+    /// Returns the cgroup subtree root exposed by this namespace.
+    pub fn root_node(&self) -> Arc<dyn SysBranchNode> {
+        self.root.clone()
+    }
+
+    /// Renders the cgroup path visible from this namespace.
+    ///
+    /// Linux reports `/proc/[pid]/cgroup` relative to the namespace root.
+    /// When the target lies outside that subtree, the path climbs up with
+    /// `..` components instead of failing.
+    pub fn virtualize_path(&self, cgroup: Arc<dyn CgroupSysNode>) -> String {
+        let root_path = self.root.path();
+        let cgroup_path = cgroup.path();
+        virtualize_path_from(root_path.as_ref(), cgroup_path.as_ref())
+    }
+}
+
+fn virtualize_path_from(root_path: &str, target_path: &str) -> String {
+    fn path_components(path: &str) -> impl Iterator<Item = &str> + Clone + '_ {
+        path.split('/').filter(|component| !component.is_empty())
+    }
+
+    let root_components = path_components(root_path);
+    let target_components = path_components(target_path);
+    let shared_len = root_components
+        .clone()
+        .zip(target_components.clone())
+        .take_while(|(root_component, target_component)| root_component == target_component)
+        .count();
+    let root_suffix = root_components.clone().skip(shared_len);
+    let target_suffix = target_components.clone().skip(shared_len);
+
+    // `SysObj::path()` already returns canonical absolute cgroup paths, so
+    // virtualization reduces to computing a relative path between them.
+    root_suffix
+        .map(|_| "..")
+        .chain(target_suffix)
+        .fold(String::from("/"), |mut path, component| {
+            if path.len() > 1 {
+                path.push('/');
+            }
+            path.push_str(component);
+            path
+        })
+}
+
+impl NsCommonOps for CgroupNamespace {
+    const TYPE: NsType = NsType::Cgroup;
+
+    fn owner_user_ns(&self) -> Option<&Arc<UserNamespace>> {
+        Some(&self.owner)
+    }
+
+    fn parent(&self) -> Result<&Arc<Self>> {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "a cgroup namespace does not have a parent namespace"
+        );
+    }
+
+    fn stashed_dentry(&self) -> &StashedDentry {
+        &self.stashed_dentry
+    }
+}
+
+#[cfg(ktest)]
+mod tests {
+    use ostd::prelude::ktest;
+
+    use super::virtualize_path_from;
+
+    #[ktest]
+    fn virtualize_cgroup_path_for_same_node() {
+        assert_eq!(virtualize_path_from("/", "/"), "/");
+        assert_eq!(virtualize_path_from("/base", "/base"), "/");
+    }
+
+    #[ktest]
+    fn virtualize_cgroup_path_for_descendant() {
+        assert_eq!(virtualize_path_from("/base", "/base/nested"), "/nested");
+    }
+
+    #[ktest]
+    fn virtualize_cgroup_path_for_sibling() {
+        assert_eq!(virtualize_path_from("/base", "/peer"), "/../peer");
+    }
+
+    #[ktest]
+    fn virtualize_cgroup_path_for_ancestor() {
+        assert_eq!(virtualize_path_from("/base", "/"), "/..");
+    }
+}
--- a/kernel/src/fs/fs_impls/cgroupfs/fs.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/fs.rs
@@ -4,13 +4,13 @@ use alloc::sync::Arc;
 
 use aster_block::BlockDevice;
 use aster_systree::EmptyNode;
+use ostd::task::Task;
 use spin::Once;
 
 use super::inode::CgroupInode;
 use crate::{
     fs::{
         Result,
-        cgroupfs::systree_node::CgroupSystem,
         pseudofs::AnonDeviceId,
         utils::systree_inode::SysTreeInodeTy,
         vfs::{
@@ -20,13 +20,13 @@ use crate::{
         },
     },
     prelude::*,
+    process::posix_thread::AsThreadLocal,
 };
 
 /// A file system for managing cgroups.
 pub(super) struct CgroupFs {
     _anon_device_id: AnonDeviceId,
     sb: SuperBlock,
-    root: Arc<dyn Inode>,
     fs_event_subscriber_stats: FsEventSubscriberStats,
 }
 
@@ -40,19 +40,17 @@ impl CgroupFs {
     pub(super) fn singleton() -> &'static Arc<CgroupFs> {
         static SINGLETON: Once<Arc<CgroupFs>> = Once::new();
 
-        SINGLETON.call_once(|| Self::new(CgroupSystem::singleton().clone()))
+        SINGLETON.call_once(Self::new)
     }
 
-    fn new(root_node: Arc<CgroupSystem>) -> Arc<Self> {
+    fn new() -> Arc<Self> {
         let anon_device_id =
             AnonDeviceId::acquire().expect("no device ID is available for cgroupfs");
         let sb = SuperBlock::new(MAGIC_NUMBER, BLOCK_SIZE, NAME_MAX, anon_device_id.id());
-        let root_inode = CgroupInode::new_root(root_node, &sb);
 
         Arc::new(Self {
             _anon_device_id: anon_device_id,
             sb,
-            root: root_inode,
             fs_event_subscriber_stats: FsEventSubscriberStats::new(),
         })
     }
@@ -69,7 +67,12 @@ impl FileSystem for CgroupFs {
     }
 
     fn root_inode(&self) -> Arc<dyn Inode> {
-        self.root.clone()
+        let current_task = Task::current().unwrap();
+        let thread_local = current_task.as_thread_local().unwrap();
+        let ns_proxy = thread_local.borrow_ns_proxy();
+        let cgroup_namespace = ns_proxy.unwrap().cgroup_ns();
+
+        CgroupInode::new_root(cgroup_namespace.root_node(), &self.sb)
     }
 
     fn sb(&self) -> SuperBlock {
--- a/kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use aster_systree::SysObj;
 use aster_util::printer::VmPrinter;
+use ostd::task::Task;
 
 use crate::{
     fs::{
+        cgroupfs::{CgroupSysNode, CgroupSystem},
         file::mkmod,
         procfs::{
             pid::TidDirOps,
@@ -13,6 +14,7 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
+    process::posix_thread::AsThreadLocal,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/cgroup` (and also `/proc/[pid]/cgroup`).
@@ -30,13 +32,16 @@ impl CgroupFileOps {
 
 impl FileOps for CgroupFileOps {
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
-        let path = self
-            .0
-            .process_ref
-            .cgroup()
-            .get()
-            .map(|cgroup| cgroup.path())
-            .unwrap_or_else(|| "/".into());
+        let cgroup: Arc<dyn CgroupSysNode> = match self.0.process_ref.cgroup().get() {
+            Some(cgroup) => cgroup.clone(),
+            None => CgroupSystem::singleton().clone(),
+        };
+
+        let current_task = Task::current().unwrap();
+        let thread_local = current_task.as_thread_local().unwrap();
+        let ns_proxy = thread_local.borrow_ns_proxy();
+        let cgroup_ns = ns_proxy.unwrap().cgroup_ns();
+        let path = cgroup_ns.virtualize_path(cgroup);
 
         let mut printer = VmPrinter::new_skip(writer, offset);
         writeln!(printer, "0::{}", path)?;
--- a/kernel/src/fs/vfs/fs_apis/file_system.rs
+++ b/kernel/src/fs/vfs/fs_apis/file_system.rs
@@ -22,6 +22,11 @@ pub trait FileSystem: Any + Sync + Send {
     fn sync(&self) -> Result<()>;
 
     /// Returns the root inode of this file system.
+    ///
+    /// For each mount, the VFS invokes this method only once and eagerly when
+    /// it creates the mount root. It never defers the lookup to a later path
+    /// walk. For mounts created by `mount()`, the call happens in the thread
+    /// context that performs the syscall.
     fn root_inode(&self) -> Arc<dyn Inode>;
 
     /// Returns the super block of this file system.
--- a/kernel/src/process/clone.rs
+++ b/kernel/src/process/clone.rs
@@ -259,7 +259,9 @@ impl CloneFlags {
             | CloneFlags::CLONE_CHILD_SETTID
             | CloneFlags::CLONE_CHILD_CLEARTID
             | CloneFlags::CLONE_VFORK
+            | CloneFlags::CLONE_NEWCGROUP
             | CloneFlags::CLONE_NEWNS
+            | CloneFlags::CLONE_NEWUTS
             | CloneFlags::CLONE_PARENT;
         let unsupported_flags = *self - supported_flags;
         if !unsupported_flags.is_empty() {
@@ -378,8 +380,9 @@ fn clone_child_task(
     let child_ns_proxy = clone_ns_proxy(
         thread_local.borrow_ns_proxy().unwrap(),
         &child_user_ns,
-        clone_flags,
+        process,
         posix_thread,
+        clone_flags,
     )?;
 
     // Clone default timer slack
@@ -492,8 +495,9 @@ fn clone_child_process(
     let child_ns_proxy = clone_ns_proxy(
         thread_local.borrow_ns_proxy().unwrap(),
         &child_user_ns,
-        clone_flags,
+        process,
         posix_thread,
+        clone_flags,
     )?;
 
     // Clone default timer slack
@@ -747,10 +751,11 @@ fn clone_user_ns(
 fn clone_ns_proxy(
     parent_ns_proxy: &Arc<NsProxy>,
     user_ns: &Arc<UserNamespace>,
-    clone_flags: CloneFlags,
+    process: &Process,
     posix_thread: &PosixThread,
+    clone_flags: CloneFlags,
 ) -> Result<Arc<NsProxy>> {
-    parent_ns_proxy.new_clone(user_ns, clone_flags, posix_thread)
+    parent_ns_proxy.new_clone(user_ns, process, posix_thread, clone_flags)
 }
 
 #[expect(clippy::too_many_arguments)]
--- a/kernel/src/process/namespace/nsproxy.rs
+++ b/kernel/src/process/namespace/nsproxy.rs
@@ -3,10 +3,10 @@
 use spin::Once;
 
 use crate::{
-    fs::vfs::path::MountNamespace,
+    fs::{cgroupfs::CgroupNamespace, vfs::path::MountNamespace},
     net::uts_ns::UtsNamespace,
     prelude::*,
-    process::{CloneFlags, UserNamespace, posix_thread::PosixThread},
+    process::{CloneFlags, Process, UserNamespace, posix_thread::PosixThread},
 };
 
 /// A struct that acts as a per-thread proxy to give access to most namespaces.
@@ -17,8 +17,9 @@ use crate::{
 /// 1. The user namespace, which is included in the `Process` struct.
 /// 2. The PID namespace, which is included in the `Process` struct (TODO).
 pub struct NsProxy {
-    uts_ns: Arc<UtsNamespace>,
+    cgroup_ns: Arc<CgroupNamespace>,
     mnt_ns: Arc<MountNamespace>,
+    uts_ns: Arc<UtsNamespace>,
 }
 
 impl NsProxy {
@@ -27,8 +28,9 @@ impl NsProxy {
         static INIT: Once<Arc<NsProxy>> = Once::new();
         INIT.call_once(|| {
             Arc::new(NsProxy {
-                uts_ns: UtsNamespace::get_init_singleton().clone(),
+                cgroup_ns: CgroupNamespace::get_init_singleton().clone(),
                 mnt_ns: MountNamespace::get_init_singleton().clone(),
+                uts_ns: UtsNamespace::get_init_singleton().clone(),
             })
         })
     }
@@ -45,8 +47,9 @@ impl NsProxy {
     pub(in crate::process) fn new_clone(
         self: &Arc<Self>,
         user_ns: &Arc<UserNamespace>,
-        clone_flags: CloneFlags,
+        process: &Process,
         posix_thread: &PosixThread,
+        clone_flags: CloneFlags,
     ) -> Result<Arc<Self>> {
         let clone_ns_flags = (clone_flags & CloneFlags::CLONE_NS_FLAGS) - CloneFlags::CLONE_NEWUSER;
 
@@ -63,9 +66,11 @@ impl NsProxy {
 
         let mut builder = NsProxyBuilder::new(self);
 
-        if clone_ns_flags.contains(CloneFlags::CLONE_NEWUTS) {
-            let uts_ns = self.uts_ns.new_clone(user_ns.clone(), posix_thread)?;
-            builder.uts_ns(uts_ns);
+        if clone_ns_flags.contains(CloneFlags::CLONE_NEWCGROUP) {
+            let current_cgroup = process.cgroup().get().as_deref().map(Arc::clone);
+            let new_cgroup_ns =
+                CgroupNamespace::new_clone(current_cgroup, user_ns.clone(), posix_thread)?;
+            builder.cgroup_ns(new_cgroup_ns);
         }
 
         if clone_ns_flags.contains(CloneFlags::CLONE_NEWNS) {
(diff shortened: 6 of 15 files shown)


kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
```diff
@@ -0,0 +1,121 @@
...
+        // `SysObj::path()` already returns canonical absolute cgroup paths, so
+        // virtualization reduces to computing a relative path between them.
+        let mut relative_components = Vec::new();
+        relative_components.extend((shared_len..base_components.len()).map(|_| ".."));
+        relative_components.extend(target_components[shared_len..].iter().copied());
+
+        debug_assert!(!relative_components.is_empty());
+
+        let mut path = String::from("/");
+        path.push_str(&relative_components.join("/"));
```
lrh2000: Can we rewrite this method to in an more efficient way? This creates many intermediate `Vec`s which I don't think are quite necessary. We can construct the final string with two `path.split` iterators I believe.

In addition, if we extract it to a method that accepts two `&str`s. We can add some unit tests.

kernel/src/fs/fs_impls/cgroupfs/fs.rs
```diff
@@ -69,7 +66,14 @@ impl FileSystem for CgroupFs {
     }
 
     fn root_inode(&self) -> Arc<dyn Inode> {
-        self.root.clone()
+        let thread = current_thread!();
+        let ns_proxy = thread.as_posix_thread().unwrap().ns_proxy().lock();
```
lrh2000: I suppose we should use `ThreadLocal` to access the NS proxy.

kernel/src/fs/fs_impls/cgroupfs/fs.rs
```diff
@@ -69,7 +66,14 @@ impl FileSystem for CgroupFs {
...
+        let cgroup_namespace = ns_proxy
+            .as_ref()
+            .map(|ns_proxy| ns_proxy.cgroup_ns())
+            .unwrap_or_else(|| CgroupNamespace::get_init_singleton());
```
lrh2000: What's the purpose of this `unwrap_or_else`?

kernel/src/fs/fs_impls/cgroupfs/fs.rs
```diff
@@ -69,7 +66,14 @@ impl FileSystem for CgroupFs {
...
+
+        CgroupInode::new_root(cgroup_namespace.root_node(), &self.sb)
```
lrh2000, line 75: Do we really need to construct a new `Inode` whenever this method is called? Will it cause user-visible differences, e.g., the inode number?

kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
```diff
@@ -30,13 +31,18 @@ impl CgroupFileOps {
...
+            Some(cgroup) => cgroup.clone(),
+            None => CgroupSystem::singleton().clone(),
+        };
+
+        let thread = current_thread!();
+        let ns_proxy = thread.as_posix_thread().unwrap().ns_proxy().lock();
+        let cgroup_ns = ns_proxy
+            .as_ref()
+            .map(|ns_proxy| ns_proxy.cgroup_ns())
+            .unwrap_or(CgroupNamespace::get_init_singleton());
```
lrh2000: Similar comments mentioned above apply to the code here.

kernel/src/process/namespace/nsproxy.rs
```diff
@@ -45,6 +50,7 @@ impl NsProxy {
     pub(in crate::process) fn new_clone(
         self: &Arc<Self>,
         user_ns: &Arc<UserNamespace>,
+        current_cgroup: Option<Arc<CgroupNode>>,
```
lrh2000: I don't really think this should accept `current_cgroup`. I believe accepting a `process: &Process` will make the code simpler and more reasonable (especially, we accept `posix_thread: &PosixThread` already).

kernel/src/syscall/setns.rs
```diff
@@ -154,33 +159,50 @@ fn try_apply_ns_from_inode<T: NsCommonOps>(
...
+    Ok(())
+}
+
+fn set_mnt_ns(
+    builder: &mut NsProxyBuilder,
+    target_ns: &Arc<MountNamespace>,
+    ctx: &Context,
+) -> Result<()> {
+    check_set_ns_perms(target_ns, ctx)?;
+
```
lrh2000, line 184: I think we can remove this blank line (line 179) but add blank lines before line 168 and line 200. So we can group permission checks in one code block.

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,706 @@
...
+	TEST_SUCC(setns(new_nsfd, 0));
+	/*
+	 * Now the caller sits above the namespace root (`/base`), so Linux
+	 * renders the current cgroup as `/..`.
+	 */
+	TEST_RES(read_proc_cgroup(getpid(), proc_buf, sizeof(proc_buf)),
+		 strcmp(proc_buf, "0::/..\n") == 0);
+
+out:
+	if (old_nsfd >= 0)
```
lrh2000: I think we should guide the code agent to be concise when writing regression tests.

For example, we should assume that all tests are successful and do not write code like `goto out`, `if (xxx_fd >= 0)`. I don't think any existing tests do things like that.

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,706 @@
...
+	pid_t tid;
+	int ready_state;
+	int error_number;
+	int should_exit;
+};
+
+static int write_all(int fd, const char *buf, size_t len)
+{
+	while (len > 0) {
+		ssize_t written = write(fd, buf, len);
```
lrh2000: The code agent should also assume that there are no partial reads/writes. When writing tests, we don't need to be that robust. Robustness is not the goal here, and it creates redundant code.

Additionally, the code is literally incorrect for file systems like procfs and cgroupfs. I don't believe we can recover from partial writes by writing the remaining content. I bet neither Asterinas nor Linux can understand this behavior.

lrh2000 (commented):
Thank you for adding support for cgroup namespaces!

kernel/src/fs/fs_impls/cgroupfs/fs.rs
cchanging: > Will it cause user-visible differences, e.g., the inode number?

`SysTreeInodeTy::new_root()` eventually derives `ino` from the underlying sysnode id at `kernel/src/fs/utils/systree_inode.rs:788`, so repeated reconstruction preserves inode numbers for the same root node.

To avoid this, we might also need to cache the inode, but I don’t think this inode access is a hot path, so I’m not sure this kind of caching optimization is worth it?

kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // `SysObj::path()` already returns canonical absolute cgroup paths, so
+    // virtualization reduces to computing a relative path between them.
+    root_suffix
+        .map(|_| "..")
+        .chain(target_suffix)
+        .fold(String::from("/"), |mut path, component| {
+            if path.len() > 1 {
+                path.push('/');
+            }
```
lrh2000, line 98: This can't reach. You initialize the path as `"/"`.

kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
lrh2000: I don't believe you address https://github.com/asterinas/asterinas/pull/3109#discussion_r3061900775.

Use `ThreadLocal`. Remove `unwrap_or` in the current process's NS proxy.

kernel/src/process/namespace/nsproxy.rs
```diff
@@ -121,20 +138,29 @@ impl<'a> NsProxyBuilder<'a> {
         self
     }
 
+    /// Sets the new UTS namespace.
```
lrh2000:
```suggestion
    /// Sets the new UTS namespace for the context being built.
```

kernel/src/syscall/setns.rs
```diff
@@ -42,7 +43,6 @@ pub fn sys_setns(fd: RawFileDesc, flags: u32, ctx: &Context) -> Result<SyscallRe
         build_proxy_from_ns_file(file.as_ref(), ns_type_flags, ctx)?
     };
 
-    // Install the newly created `NsProxy`.
```
lrh2000, line 45: Why removing this comment?

kernel/src/syscall/setns.rs
```diff
@@ -82,17 +82,20 @@ fn build_proxy_from_pid_file(
...
+        let target_ns = target_proxy.cgroup_ns();
+        set_cgroup_ns(&mut builder, target_ns, ctx)?;
     }
 
     if flags.contains(CloneFlags::CLONE_NEWNS) {
         let target_ns = target_proxy.mnt_ns();
         set_mnt_ns(&mut builder, target_ns, ctx)?;
     }
 
-    // TODO: Support setting other namespaces from the target process.
```
lrh2000, line 95: Why removing this comment?

kernel/src/syscall/setns.rs
```diff
@@ -119,13 +122,15 @@ fn build_proxy_from_ns_file(
...
+        || try_apply_ns_from_inode::<CgroupNamespace>(inode_handle, flags, |ns| {
+            set_cgroup_ns(&mut builder, &ns, ctx)
         })?
         || try_apply_ns_from_inode::<MountNamespace>(inode_handle, flags, |ns| {
             set_mnt_ns(&mut builder, &ns, ctx)
+        })?
+        || try_apply_ns_from_inode::<UtsNamespace>(inode_handle, flags, |ns| {
+            set_uts_ns(&mut builder, &ns, ctx)
         })?;
-    // TODO: Support setting other namespaces from the ns file.
```
lrh2000, line 128: Why removing this comment?

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,566 @@
...
+	}
+
+	child = fork();
+	if (child < 0) {
+		close(ready_pipe[0]);
+		close(ready_pipe[1]);
+		close(release_pipe[0]);
+		close(release_pipe[1]);
+		return -1;
+	}
```
lrh2000: I think we should just `CHECK()`.

Same for the `ext(1)`/`exit(2)`/... in the child process.

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,566 @@
...
+	TEST_RES(0, worker_ready);
+
+	if (worker_ready) {
+		main_after_fd = TEST_SUCC(open_task_cgroup_ns_fd(main_tid));
+		worker_fd = TEST_SUCC(open_task_cgroup_ns_fd(worker_tid));
+		TEST_SUCC(fstat(main_after_fd, &main_after_stat));
+		TEST_SUCC(fstat(worker_fd, &worker_stat));
+
+		TEST_RES(0, main_before_stat.st_ino == main_after_stat.st_ino);
+		TEST_RES(0, main_after_stat.st_ino != worker_stat.st_ino);
```
lrh2000: Merge these `TEST_RES` back the the `TEST_SUCC` statements.

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,566 @@
...
+	while (thread_args.unshare_errno < 0)
+		TEST_RES(pthread_cond_wait(&thread_args.cond,
+					   &thread_args.lock),
+			 _ret == 0);
+	worker_tid = thread_args.tid;
+	worker_ready = thread_args.unshare_errno == 0;
+	TEST_RES(pthread_mutex_unlock(&thread_args.lock), _ret == 0);
+	TEST_RES(0, worker_ready);
+
+	if (worker_ready) {
```
lrh2000: No need to check this.

In fact, no need to have this `worker_ready`. Just do:
```diff
-worker_ready = thread_args.unshare_errno == 0;
+TEST_RES(thread_args.unshare_errno, _ret == 0);
```

kernel/src/fs/fs_impls/cgroupfs/fs.rs
lrh2000: Perhaps either way is okay as long as it don't cause user-visible differences.

> so I’m not sure this kind of caching optimization is worth it?

I mean that the cgroup namespace can contain the root inode. So we don't need to create it repeatedly. Do you think implementing it would be difficult?

Currently, this is the only user of `root_node`. So it can indeed be replaced to an `root_inode()`?

---

Also `root_node`/`root_inode` perhaps should return a reference. So the caller can clone only if necessary.

---

**Edit:** Sorry, I think I missed something. According to the behavior confirmed in the test, only new mounts will recognize new root inodes. Is this method guaranteed to be called only once by the process that invokes the `mount()` system call?

If so, then it is really a cold path. I thought it would be called whenever a process tried to access the cgroup file system root.

In this case, I don't think the doc comments for this method are complete. At least from doc comments, there's no way to know what context this method will be called in.
```rust
    /// Returns the root inode of this file system.
    fn root_inode(&self) -> Arc<dyn Inode>;
```

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,566 @@
...
+static void init_cgroup_test_dirs(struct cgroup_test_dirs *dirs)
+{
+	snprintf(dirs->base, sizeof(dirs->base), CGROUP_ROOT "/cgns-base-%d",
+		 getpid());
+	snprintf(dirs->peer, sizeof(dirs->peer), CGROUP_ROOT "/cgns-peer-%d",
+		 getpid());
+	snprintf(dirs->mount, sizeof(dirs->mount), "/tmp/cgns-mnt-%d",
+		 getpid());
+	CHECK(join_path(dirs->nested, sizeof(dirs->nested), dirs->base,
+			"nested"));
```
lrh2000: If manually written, a better form is just to use:
```c
FN_SETUP(init)
{
        CHECK(mkdir("/proc/cgroup/test-peer"));
        CHECK(mkdir("/proc/cgroup/test-mnt"));
}
END_SETUP()
```

Something like:
```c
	TEST_RES(snprintf(expected, sizeof(expected), "0::%s\n",
			  relative_to_cgroup_root(dirs.base)),
		 _ret >= 0 && (size_t)_ret < sizeof(expected));
	TEST_RES(read_proc_cgroup(getpid(), proc_buf, sizeof(proc_buf)),
		 strcmp(proc_buf, expected) == 0);
```
can be simplified to:
```c
	TEST_RES(read_proc_cgroup(getpid(), proc_buf, sizeof(proc_buf)),
		 strcmp(proc_buf, "0::/test-base\n") == 0);
```

Even better, the `proc_buf` can be handled within `read_proc_cgroup`:
```c
const char *read_proc_cgroup()
{
        static char proc_buf[BUF_LEN];
        /* ... */
        return proc_buf;
}
```
```c
	TEST_RES(read_proc_cgroup(getpid()), strcmp(_ret, "0::/test-base\n") == 0);
```
Or maybe we can merge `strcmp` to the method altogether. Then the test can be made easier to read and write.

Telling the code agent to be less verbose will make the review easier. Currently, reading a large code snippet that is automatically generated in this way is still painful. However, I don't know if it's worth the cost to fix this at this moment. For this PR, I've already read it, so it may not matter much anyway.

kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
lrh2000: Sorry, ignore this. It seems that I misread the code. This branch is useful for preventing the "/" from being pushed twice at the beginning.

lrh2000:
I would expect https://github.com/asterinas/asterinas/pull/3109#discussion_r3061913200, https://github.com/asterinas/asterinas/pull/3109#discussion_r3063532324, e.t.c. to be the norm for AI-generated tests because I noticed similar patterns when using Codex to generate regression tests for my vsock rewrite.

However, using code agents to write regression tests is one of the most appropriate applications of AI (much more appropriate than allowing AI to write kernel code). I think we should fix something to make progress more flexible.

I spent some time creating a skill, which I then used to rewrite the test that this PR intends to add.

The end result is much better. I still adjust the generated tests manually, but only make minor adjustments. Then, in my opinion, the test is perfect. The new test is simpler, yet still covers everything this PR expects to test.

<details>
<summary> The new test </summary>

```c

#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <limits.h>
#include <pthread.h>
#include <sched.h>
#include <signal.h>
#include <stdio.h>
#include <string.h>
#include <sys/mount.h>
#include <sys/stat.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <unistd.h>

#include "../../common/test.h"

#define CGROUP_ROOT "/sys/fs/cgroup"
#define TEST_BASE_CGROUP CGROUP_ROOT "/cgns-base"
#define TEST_LEAF_CGROUP TEST_BASE_CGROUP "/leaf"
#define TEST_SIBLING_CGROUP CGROUP_ROOT "/cgns-sibling"
#define TEST_MOUNT_DIR "/tmp/cgns-mnt"
#define TEST_MOUNT_LEAF TEST_MOUNT_DIR "/leaf"
#define TEST_MOUNT_SIBLING TEST_MOUNT_DIR "/cgns-sibling"

static int g_initial_cgroup_ns_fd;
static int g_initial_mount_ns_fd;

static int move_pid_to_cgroup(const char *dir, pid_t pid)
{
	char path[PATH_MAX];
	char pid_text[32];
	int fd;
	ssize_t count;

	snprintf(path, sizeof(path), "%s/cgroup.procs", dir);
	fd = open(path, O_WRONLY);
	if (fd < 0)
		return -1;

	snprintf(pid_text, sizeof(pid_text), "%d", pid);
	count = write(fd, pid_text, strlen(pid_text));
	close(fd);
	if (count != (ssize_t)strlen(pid_text)) {
		if (count >= 0)
			errno = EIO;
		return -1;
	}

	return 0;
}

static int check_proc_cgroup(pid_t pid, const char *expected)
{
	char path[64];
	char content[PATH_MAX];
	int fd;
	ssize_t count;

	snprintf(path, sizeof(path), "/proc/%d/cgroup", pid);
	fd = open(path, O_RDONLY);
	if (fd < 0)
		return -1;

	count = read(fd, content, sizeof(content) - 1);
	close(fd);
	if (count < 0)
		return -1;

	content[count] = '\0';
	if (strcmp(content, expected) != 0) {
		errno = EIO;
		return -1;
	}

	return 0;
}

static int read_namespace_inode(const char *path, ino_t *inode)
{
	struct stat stat_buf;

	if (stat(path, &stat_buf) < 0)
		return -1;

	*inode = stat_buf.st_ino;
	return 0;
}

static int read_self_cgroup_ns_inode(ino_t *inode)
{
	return read_namespace_inode("/proc/self/ns/cgroup", inode);
}

static int read_task_cgroup_ns_inode(pid_t tid, ino_t *inode)
{
	char path[64];

	snprintf(path, sizeof(path), "/proc/self/task/%d/ns/cgroup", tid);
	return read_namespace_inode(path, inode);
}

static int read_process_cgroup_ns_inode(pid_t pid, ino_t *inode)
{
	char path[64];

	snprintf(path, sizeof(path), "/proc/%d/ns/cgroup", pid);
	return read_namespace_inode(path, inode);
}

static int check_mount_root(const char *mount_point, const char *expected_root)
{
	FILE *mountinfo = fopen("/proc/self/mountinfo", "r");
	char line[512];

	if (!mountinfo)
		return -1;

	while (fgets(line, sizeof(line), mountinfo)) {
		char root[256];
		char seen_mount_point[256];
		char fs_type[64];

		if (sscanf(line, "%*s %*s %*s %255s %255s %*s - %63s", root,
			   seen_mount_point, fs_type) != 3)
			continue;
		if (strcmp(seen_mount_point, mount_point) != 0)
			continue;
		if (strcmp(fs_type, "cgroup2") != 0)
			continue;

		fclose(mountinfo);
		if (strcmp(root, expected_root) != 0) {
			errno = EIO;
			return -1;
		}

		return 0;
	}

	fclose(mountinfo);
	errno = ENOENT;
	return -1;
}

static int reset_self(void)
{
	if (setns(g_initial_cgroup_ns_fd, CLONE_NEWCGROUP) < 0)
		return -1;
	if (move_pid_to_cgroup(CGROUP_ROOT, getpid()) < 0)
		return -1;
	if (setns(g_initial_mount_ns_fd, CLONE_NEWNS) < 0)
		return -1;

	return 0;
}

enum new_cgroup_ns_t {
	KEEP_OLD_CGROUP_NS = 0,
	MOVE_TO_NEW_CGROUP_NS = 1,
};

static pid_t spawn_paused_process(const char *cgroup_dir,
				  enum new_cgroup_ns_t new_cgroup)
{
	pid_t child = CHECK(fork());

	if (child == 0) {
		CHECK(move_pid_to_cgroup(cgroup_dir, getpid()));
		if (new_cgroup)
			CHECK(unshare(CLONE_NEWCGROUP));
		pause();
		_exit(0);
	}

	/* Wait until the child executes `pause()`. */
	CHECK(usleep(50000));

	return child;
}

static void kill_paused_process(pid_t pid)
{
	CHECK(kill(pid, SIGKILL));
	CHECK_WITH(waitpid(pid, NULL, 0), _ret == pid);
}

struct thread_probe {
	pthread_barrier_t ready;
	pthread_barrier_t release;
	pid_t tid;
};

static void *unshare_cgroup_ns_thread(void *arg)
{
	struct thread_probe *probe = arg;

	probe->tid = CHECK(syscall(SYS_gettid));
	CHECK(unshare(CLONE_NEWCGROUP));
	CHECK_WITH(pthread_barrier_wait(&probe->ready),
		   _ret == 0 || _ret == PTHREAD_BARRIER_SERIAL_THREAD);
	CHECK_WITH(pthread_barrier_wait(&probe->release),
		   _ret == 0 || _ret == PTHREAD_BARRIER_SERIAL_THREAD);

	return NULL;
}

FN_SETUP(init)
{
	g_initial_cgroup_ns_fd = CHECK(open("/proc/self/ns/cgroup", O_RDONLY));
	g_initial_mount_ns_fd = CHECK(open("/proc/self/ns/mnt", O_RDONLY));

	CHECK(mkdir(TEST_BASE_CGROUP, 0755));
	CHECK(mkdir(TEST_LEAF_CGROUP, 0755));
	CHECK(mkdir(TEST_SIBLING_CGROUP, 0755));
	CHECK(mkdir(TEST_MOUNT_DIR, 0755));
}
END_SETUP()

/*
 * `unshare(CLONE_NEWCGROUP)` only changes the calling thread's namespace.
 * Other threads in the same process keep the original namespace inode.
 */
FN_TEST(cgroup_namespace_stays_thread_local)
{
	struct thread_probe probe = {};
	pthread_t thread;
	ino_t main_before;
	ino_t main_after;
	ino_t worker_ns;

	TEST_SUCC(read_self_cgroup_ns_inode(&main_before));

	TEST_RES(pthread_barrier_init(&probe.ready, NULL, 2), _ret == 0);
	TEST_RES(pthread_barrier_init(&probe.release, NULL, 2), _ret == 0);
	TEST_RES(pthread_create(&thread, NULL, unshare_cgroup_ns_thread,
				&probe),
		 _ret == 0);

	TEST_RES(pthread_barrier_wait(&probe.ready),
		 _ret == 0 || _ret == PTHREAD_BARRIER_SERIAL_THREAD);
	TEST_RES(read_self_cgroup_ns_inode(&main_after),
		 main_before == main_after);
	TEST_RES(read_task_cgroup_ns_inode(probe.tid, &worker_ns),
		 worker_ns != main_after);
	TEST_RES(pthread_barrier_wait(&probe.release),
		 _ret == 0 || _ret == PTHREAD_BARRIER_SERIAL_THREAD);

	TEST_RES(pthread_join(thread, NULL), _ret == 0);
	TEST_RES(pthread_barrier_destroy(&probe.ready), _ret == 0);
	TEST_RES(pthread_barrier_destroy(&probe.release), _ret == 0);
}
END_TEST()

/*
 * `/proc/<pid>/cgroup` is rendered from the caller's active cgroup namespace:
 * the namespace root becomes `/`, descendants stay reachable beneath it, and
 * cgroups outside the root are shown through `..`.
 */
FN_TEST(proc_cgroup_follows_callers_namespace)
{
	int rooted_ns_fd;
	pid_t sibling_child;

	sibling_child =
		spawn_paused_process(TEST_SIBLING_CGROUP, KEEP_OLD_CGROUP_NS);
	TEST_SUCC(move_pid_to_cgroup(TEST_BASE_CGROUP, getpid()));
	TEST_SUCC(check_proc_cgroup(sibling_child, "0::/cgns-sibling\n"));
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/cgns-base\n"));

	TEST_SUCC(unshare(CLONE_NEWCGROUP));
	rooted_ns_fd = TEST_SUCC(open("/proc/self/ns/cgroup", O_RDONLY));
	TEST_SUCC(check_proc_cgroup(sibling_child, "0::/../cgns-sibling\n"));
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/\n"));

	TEST_SUCC(move_pid_to_cgroup(TEST_LEAF_CGROUP, getpid()));
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/leaf\n"));

	TEST_SUCC(setns(g_initial_cgroup_ns_fd, CLONE_NEWCGROUP));
	TEST_SUCC(move_pid_to_cgroup(CGROUP_ROOT, getpid()));
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/\n"));

	TEST_SUCC(setns(rooted_ns_fd, CLONE_NEWCGROUP));
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/..\n"));

	TEST_SUCC(reset_self());
	TEST_SUCC(close(rooted_ns_fd));
	kill_paused_process(sibling_child);
}
END_TEST()

/*
 * Joining another task's cgroup namespace should immediately re-virtualize
 * `/proc/self/cgroup` against that namespace's root.
 */
FN_TEST(setns_revirtualizes_proc_cgroup)
{
	ino_t child_ns;
	ino_t joined_ns;
	pid_t child;
	int pidfd;

	TEST_SUCC(move_pid_to_cgroup(TEST_BASE_CGROUP, getpid()));
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/cgns-base\n"));

	child = spawn_paused_process(TEST_BASE_CGROUP, MOVE_TO_NEW_CGROUP_NS);
	TEST_SUCC(read_process_cgroup_ns_inode(child, &child_ns));

	pidfd = TEST_SUCC(syscall(SYS_pidfd_open, child, 0));
	TEST_SUCC(setns(pidfd, CLONE_NEWCGROUP));

	TEST_RES(read_self_cgroup_ns_inode(&joined_ns), joined_ns == child_ns);
	TEST_SUCC(check_proc_cgroup(getpid(), "0::/\n"));

	TEST_SUCC(reset_self());
	TEST_SUCC(close(pidfd));
	kill_paused_process(child);
}
END_TEST()

/*
 * A fresh `cgroup2` mount inside a cgroup namespace starts at the namespace
 * root, so descendants below that root stay visible and siblings above it do
 * not appear in the mount tree.
 */
FN_TEST(cgroup_mount_starts_at_namespace_root)
{
	TEST_SUCC(move_pid_to_cgroup(TEST_BASE_CGROUP, getpid()));
	TEST_SUCC(unshare(CLONE_NEWCGROUP));

	TEST_SUCC(unshare(CLONE_NEWNS));
	TEST_SUCC(mount("none", TEST_MOUNT_DIR, "cgroup2", 0, NULL));

	TEST_SUCC(check_mount_root(TEST_MOUNT_DIR, "/"));
	TEST_SUCC(access(TEST_MOUNT_LEAF, F_OK));
	TEST_ERRNO(access(TEST_MOUNT_SIBLING, F_OK), ENOENT);

	TEST_SUCC(umount(TEST_MOUNT_DIR));
	TEST_SUCC(reset_self());
}
END_TEST()

FN_SETUP(cleanup)
{
	CHECK(rmdir(TEST_MOUNT_DIR));
	CHECK(rmdir(TEST_LEAF_CGROUP));
	CHECK(rmdir(TEST_BASE_CGROUP));
	CHECK(rmdir(TEST_SIBLING_CGROUP));
	CHECK(close(g_initial_mount_ns_fd));
	CHECK(close(g_initial_cgroup_ns_fd));
}
END_SETUP()
```

</details>

Skill: [asterinas-test-guardrails.zip](https://github.com/user-attachments/files/26645685/asterinas-test-guardrails.zip)

kernel/src/fs/fs_impls/cgroupfs/fs.rs
cchanging: >  Is this method guaranteed to be called only once by the process that invokes the mount() system call?

Yes, this is the only usage of `root_inode` method, so it should not be a performance-critical path.

> In this case, I don't think the doc comments for this method are complete. At least from doc comments, there's no way to know what context this method will be called in.

Thank you for the reminder, I will add the necessary details.

kernel/src/fs/vfs/fs_apis/file_system.rs
```diff
@@ -22,6 +22,9 @@ pub trait FileSystem: Any + Sync + Send {
     fn sync(&self) -> Result<()>;
 
     /// Returns the root inode of this file system.
+    ///
+    /// This method is typically called only during mount initialization to obtain
+    /// the filesystem root inode.
```
lrh2000: I don't think we should say `typically`. It needs to be guaranteed. Otherwise, how can you be sure that the cgroupfs root inode is obtained correctly?

In addition, something that must be made explicitly it that it needs to be called in the thread context that performs the `mount()`.
 - In other words, we need to rule out the possible scenarios that it is lazily called later, in a context belonging to an arbitrary process that tries to access the path.

kernel/src/syscall/setns.rs
```diff
@@ -181,6 +181,34 @@ fn set_mnt_ns(
     target_ns: &Arc<MountNamespace>,
     ctx: &Context,
 ) -> Result<()> {
+    check_set_ns_perms(target_ns, ctx)?;
+    if ctx.thread_local.is_fs_shared() {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "setting a mount namespace is not allowed with shared filesystem information"
```
lrh2000, line 188: I thought this was a permission check, but it seems not to be. So, it might be better to keep the original blank line before `ctx.thread_local.is_fs_shared()`. Sorry for the noise.

test/initramfs/src/regression/security/namespace/cgroup_ns.c
```diff
@@ -0,0 +1,356 @@
+// SPDX-License-Identifier: MPL-2.0
```
lrh2000, line 1: For credit/blame purposes, I think you can add `Co-authored-by: Ruihan Li <lrh2000@pku.edu.cn>` to the second commit's commit message since it is really changed significantly by me...

kernel/src/syscall/setns.rs
```diff
@@ -68,10 +69,10 @@ fn build_proxy_from_pid_file(
...
     check_unsupported_ns_flags(flags)?;
 
-    let target_thread = pid_file
+    let target_process = pid_file
         .process_opt()
-        .ok_or_else(|| Error::with_message(Errno::ESRCH, "the target process has been reaped"))?
-        .main_thread();
+        .ok_or_else(|| Error::with_message(Errno::ESRCH, "the target process has been reaped"))?;
+    let target_thread = target_process.main_thread();
     let target_proxy = target_thread.as_posix_thread().unwrap().ns_proxy().lock();
```
lrh2000, line 76: It seems that changes here are unrelated to this PR? I'm not sure if it's intended or not. Both the previous code and the new code is okay for me, though.

kernel/src/process/clone.rs
```diff
@@ -260,6 +260,8 @@ impl CloneFlags {
             | CloneFlags::CLONE_CHILD_CLEARTID
             | CloneFlags::CLONE_VFORK
             | CloneFlags::CLONE_NEWNS
+            | CloneFlags::CLONE_NEWCGROUP
```
lrh2000, line 262: Move `NEWCGROUP` before `NEWNS`?

kernel/src/fs/vfs/fs_apis/file_system.rs
```diff
@@ -22,6 +22,11 @@ pub trait FileSystem: Any + Sync + Send {
     fn sync(&self) -> Result<()>;
 
     /// Returns the root inode of this file system.
+    ///
+    /// The VFS invokes this method eagerly when it creates the mount root, and
```
lrh2000: I think it is better to change the first sentence to something like "For each mount, the VFS invokes this method only once and eagerly when it creates the mount root".

```rust
    /// For each mount, the VFS invokes this method only once and eagerly when
    /// it creates the mount root. It never defers the lookup to a later path
    /// walk. For mounts created by `mount()`, the call happens in the thread
    /// context that performs the syscall.
```

kernel/src/syscall/setns.rs
```diff
@@ -119,12 +125,16 @@ fn build_proxy_from_ns_file(
...
+            set_cgroup_ns(&mut builder, &ns, ctx)
         })?
         || try_apply_ns_from_inode::<MountNamespace>(inode_handle, flags, |ns| {
             set_mnt_ns(&mut builder, &ns, ctx)
+        })?
+        || try_apply_ns_from_inode::<UtsNamespace>(inode_handle, flags, |ns| {
+            set_uts_ns(&mut builder, &ns, ctx)
         })?;
+
     // TODO: Support setting other namespaces from the ns file.
```
lrh2000, line 137: This comment explains the previous code block. Remove the blank line before it. Especially, it does not exist before this PR anyway.

lrh2000 (approved):
Just two final minor nits. Otherwise LGTM. Thanks for your contribution!

cqs21 (approved):

PR #3111, opened 2026-04-09
Fix fork and vfork panic on pids.max exhaustion

Marsman1996:
This PR fixes a kernel panic in `fork` and `vfork` when process creation is rejected by the cgroup v2 pids controller.

In `clone` and `clone3`, the `clone_child()` uses `?` already.

https://github.com/asterinas/asterinas/blob/b05d53e9661f650ab21abd0bf367fe0632b2df39/kernel/src/syscall/clone.rs#L28

### DIFF
--- a/kernel/src/syscall/fork.rs
+++ b/kernel/src/syscall/fork.rs
@@ -10,12 +10,12 @@ use crate::{
 
 pub fn sys_fork(ctx: &Context, parent_context: &UserContext) -> Result<SyscallReturn> {
     let clone_args = CloneArgs::for_fork();
-    let child_pid = clone_child(ctx, parent_context, clone_args).unwrap();
+    let child_pid = clone_child(ctx, parent_context, clone_args)?;
     Ok(SyscallReturn::Return(child_pid as _))
 }
 
 pub fn sys_vfork(ctx: &Context, parent_context: &UserContext) -> Result<SyscallReturn> {
     let clone_args = CloneArgs::for_vfork();
-    let child_pid = clone_child(ctx, parent_context, clone_args).unwrap();
+    let child_pid = clone_child(ctx, parent_context, clone_args)?;
     Ok(SyscallReturn::Return(child_pid as _))
 }


Marsman1996:
## PoC
```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <linux/limits.h>
#include <signal.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <unistd.h>

static void write_text(const char *path, const char *text) {
    int fd = open(path, O_WRONLY | O_CLOEXEC);

    if (fd < 0 || write(fd, text, strlen(text)) < 0) {
        perror(path);
        exit(1);
    }

    close(fd);
}

static unsigned long read_ulong(const char *path) {
    FILE *file = fopen(path, "r");
    unsigned long value;

    if (file == NULL || fscanf(file, "%lu", &value) != 1) {
        perror(path);
        exit(1);
    }

    fclose(file);
    return value;
}

int main(void) {
    const char *root = "/sys/fs/cgroup";
    const char *name = "panic_pilot_poc";
    char path[PATH_MAX];
    char value[32];

    snprintf(path, sizeof(path), "%s/%s", root, name);
    if (mkdir(path, 0755) < 0 && errno != EEXIST) {
        perror("mkdir cgroup");
        return 1;
    }

    snprintf(path, sizeof(path), "%s/cgroup.subtree_control", root);
    if (access(path, W_OK) == 0) {
        int fd = open(path, O_WRONLY | O_CLOEXEC);

        if (fd >= 0 && write(fd, "+pids", 5) < 0 && errno != EBUSY) {
            perror("enable +pids in root subtree");
            close(fd);
            return 1;
        }
        if (fd >= 0) {
            close(fd);
        }
    } else if (errno != EROFS) {
        perror("enable +pids in root subtree");
        return 1;
    }

    snprintf(path, sizeof(path), "%s/%s/cgroup.procs", root, name);
    snprintf(value, sizeof(value), "%ld", (long)getpid());
    write_text(path, value);

    snprintf(path, sizeof(path), "%s/%s/pids.current", root, name);
    unsigned long current = read_ulong(path);

    if (current == 0) {
        fprintf(stderr, "unexpected pids.current\n");
        return 1;
    }

    snprintf(path, sizeof(path), "%s/%s/pids.max", root, name);
    snprintf(value, sizeof(value), "%lu", current);
    write_text(path, value);

    long ret = syscall(SYS_fork);
    if (ret < 0) {
        perror("fork");
        return 0;
    }

    if (ret == 0) {
        _exit(0);
    }

    kill((pid_t)ret, SIGKILL);
    return 0;
}
```

## Log
```log
~ # /etc/fork_pids_max_panic.c 
[    74.114] ERROR : Uncaught panic:
        called `Result::unwrap()` on an `Err` value: Error { errno: EAGAIN, msg: Some("the pids sub-controller limit is reached") }
        at /root/asterinas/kernel/src/syscall/fork.rs:13
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(0), real_time: RealTimeAttr { prio: 99, time_slice: 31491380 }, fair: FairAttr { weight: 1024, pending_weight: 0, vruntime: 417456063 } } })
```

lrh2000 (approved):
This is a good catch! Thanks for identifying it and submitting the fix!

PR #3114, opened 2026-04-10
Implement BLKSSZGET ioctl

Fischer0522:
This pull request has introduced the `BLKSSZGET` ioctl, which is similar to the existing `BLKGETSIZE64` ioctl. Some test cases within the `xfstests` suite depend on this particular ioctl for execution.

I'm not sure whether we will add a logical_block_size like Linux ([ref](https://elixir.bootlin.com/linux/v6.18/source/block/ioctl.c#L596)), so I add a TODO here.

### DIFF
--- a/kernel/src/device/registry/block.rs
+++ b/kernel/src/device/registry/block.rs
@@ -1,12 +1,13 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use aster_block::{BlockDevice, SECTOR_SIZE};
+use aster_block::{BLOCK_SIZE, BlockDevice, SECTOR_SIZE};
 use aster_nvme::NvmeBlockDevice;
 use aster_virtio::device::block::device::BlockDevice as VirtIoBlockDevice;
 use device_id::DeviceId;
 use ostd::mm::VmIo;
 
 use crate::{
+    context::current_userspace,
     device::{Device, DeviceType, DevtmpfsInodeMeta, add_node},
     events::IoEvents,
     fs::{
@@ -65,10 +66,27 @@ pub(super) fn init_in_first_process(path_resolver: &PathResolver) -> Result<()> 
 }
 
 mod ioctl_defs {
-    use crate::util::ioctl::{OutData, ioc};
+    use crate::util::ioctl::{NoData, OutData, ioc};
 
     // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fs.h>
+
+    /// Returns the device size in bytes.
     pub(super) type BlkGetSize64 = ioc!(BLKGETSIZE64, 0x12, 114, OutData<u64>);
+
+    /// Returns the logical sector size of the block device.
+    ///
+    /// This is the smallest unit of I/O the device can address and,
+    /// importantly, the minimum alignment required for `O_DIRECT` I/O on
+    /// files backed by this device. Both buffer address and offset must be a
+    /// multiple of this value.
+    ///
+    /// Benchmarks and filesystem tests (for example, `xfstests`, LTP
+    /// `preadv03`/`pwritev03`) rely on this ioctl to size `O_DIRECT` buffers.
+    /// If the effective alignment enforced by the filesystem layered on top
+    /// is larger than the hardware sector, such as `ext2`'s 4 KiB block, this
+    /// ioctl must return that larger value. Otherwise user programs will align
+    /// correctly for the device but still hit `EINVAL` at the filesystem.
+    pub(super) type BlkGetSectorSize = ioc!(BLKSSZGET, 0x12, 104, NoData);
 }
 
 /// Represents a block device inode in the filesystem.
@@ -156,6 +174,15 @@ impl PerOpenFileOps for OpenBlockFile {
         use ioctl_defs::*;
 
         dispatch_ioctl!(match raw_ioctl {
+            _cmd @ BlkGetSectorSize => {
+                // TODO: Query the per-device logical block size once block device metadata
+                // exposes it. For now, report the effective minimum I/O granularity enforced
+                // by Asterinas filesystems so userspace can use `BLKSSZGET` for `O_DIRECT`
+                // alignment.
+                let sector_size = SECTOR_SIZE.max(BLOCK_SIZE) as i32;
+                current_userspace!().write_val(raw_ioctl.arg(), &sector_size)?;
+                Ok(0)
+            }
             cmd @ BlkGetSize64 => {
                 let size = (self.0.metadata().nr_sectors * SECTOR_SIZE) as u64;
                 cmd.write(&size)?;
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/ioctl.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-descriptor-and-io-control/ioctl.scml
@@ -27,7 +27,7 @@ ioctl(
 );
 
 // Control block devices
-ioctl(fd, op = BLKGETSIZE64, ..);
+ioctl(fd, op = BLKGETSIZE64 | BLKSSZGET, ..);
 
 // Control Trust Domain Extensions (TDX) guest devices
 ioctl(fd, op = TDX_CMD_GET_REPORT0, ..);
--- a/test/initramfs/src/conformance/xfstests/full.list
+++ b/test/initramfs/src/conformance/xfstests/full.list
@@ -58,6 +58,7 @@ generic/393
 generic/401
 generic/406
 generic/412
+generic/418
 generic/428
 generic/437
 generic/438
--- a/test/initramfs/src/conformance/xfstests/short.list
+++ b/test/initramfs/src/conformance/xfstests/short.list
@@ -52,6 +52,7 @@ generic/393
 generic/401
 generic/406
 generic/412
+generic/418
 generic/428
 generic/437
 generic/438


kernel/src/device/registry/block.rs
```diff
@@ -48,10 +49,11 @@ pub(super) fn init_in_first_process(path_resolver: &PathResolver) -> Result<()>
 }
 
 mod ioctl_defs {
-    use crate::util::ioctl::{OutData, ioc};
+    use crate::util::ioctl::{NoData, OutData, ioc};
 
     // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fs.h>
     pub(super) type BlkGetSize64 = ioc!(BLKGETSIZE64, 0x12, 114, OutData<u64>);
+    pub(super) type BlkGetSectorSize = ioc!(BLKSSZGET, 0x1268, NoData);
```
tatetian, line 56: **Prefer the modern `ioc!` form**: `BLKSSZGET` is defined in Linux's `fs.h` as [`_IO(0x12, 104)`](https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fs.h#L272):

```c
#define BLKSSZGET  _IO(0x12,104)/* get block device sector size */
```

So we should also define it in the modern ioctl encoding:

```rust
pub(super) type BlkGetSectorSize = ioc!(BLKSSZGET, 0x12, 104, NoData);
```

Besides matching the neighbouring entry, this is easier for a reader to verify against the C definition without having to reconstruct `0x1268 = (0x12 << 8) | 104` in their head.

**Document the "`_IO` but actually writes data" Linux quirk**:

```rust
/// Gets the sector size of the underlying block device.
///
/// # Linux quirk
///
/// Linux defines this number as follows:
///
/// ```c
/// #define BLKSSZGET  _IO(0x12,104)/* get block device sector size */
/// ```
///
/// The data direction and size bits of this number DO NOT reflect 
/// the real behaviors of this ioctl.
///
/// To conform with Linux's encoding, our definition uses `NoData`,
/// but in principle it should have been `OutData<i32>`.
pub(super) type BlkGetSectorSize = ioc!(BLKSSZGET, 0x12, 104, NoData);
```

kernel/src/device/registry/block.rs
```diff
@@ -139,6 +141,16 @@ impl FileIo for OpenBlockFile {
         use ioctl_defs::*;
 
         dispatch_ioctl!(match raw_ioctl {
+            _cmd @ BlkGetSectorSize => {
+                // TODO: Query the per-device logical block size from block device
+                // metadata instead of using the hardcoded `SECTOR_SIZE`.
+                // BLKSSZGET is encoded as `_IO` (no data) in Linux, but actually writes
+                // an `int` to the user pointer. We use `NoData` to match the encoding and
+                // write manually.
+                let sector_size = SECTOR_SIZE as i32;
```
tatetian, line 150: **Returning 512 breaks O_DIRECT on ext2 and thus the xfstests motivation**: `BLKSSZGET` is the value userspace (xfstests, LTP `preadv03`/`pwritev03`, `tst_device.c`) uses to compute O_DIRECT alignment. Asterinas's ext2 hardcodes a 4 KiB filesystem block and `Ext2Inode::read_direct_at` / `write_direct_at` reject any non-block-aligned offset/length with `EINVAL` (`kernel/src/fs/fs_impls/ext2/inode.rs:718-725`). Concretely, the typical xfstests pattern is:

```c
int lbs;
ioctl(dev_fd, BLKSSZGET, &lbs);          // -> 512 with this PR
posix_memalign(&buf, lbs, lbs);          // 512-aligned buffer
pread(file_fd, buf, lbs, 0);             // with O_DIRECT: EINVAL from ext2
```

So every O_DIRECT test will fail at the first `pread`/`pwrite`, and the ioctl — although correctly plumbed — does not unblock xfstests as claimed in the PR description.

The cleanest fix is to make `BLKSSZGET` return the larger of the device's hardware sector size and the effective minimum I/O granularity the kernel actually enforces on top of it. Since ext2 mandates 4 KiB, returning `max(SECTOR_SIZE, 4096) = 4096` (or, equivalently, `BLOCK_SIZE`) matches what userspace must actually align to. This mirrors what Linux's block layer does via `queue_logical_block_size()` when a filesystem or stacking driver raises the effective LBS.

Three options worth weighing:
1. **Hardcode 4 KiB here** with a comment pointing at the ext2 alignment requirement. Fast, unblocks xfstests, but a lie about the underlying virtio-blk device.
2. **Query the block device's logical block size** (already TODO'd) and independently raise it to `BLOCK_SIZE` when the device is consumed by ext2. This is the direction Linux takes.
3. **Relax `read_direct_at` / `write_direct_at`** to accept `SECTOR_SIZE`-aligned I/O and do the internal block-level fan-out. This preserves the honest semantics of `BLKSSZGET` but is a larger change.

Whichever path is chosen, shipping 512 today makes the ioctl actively misleading for its stated consumer; a follow-up that just adds the `max(SECTOR_SIZE, BLOCK_SIZE)` clamp here would be enough to make the xfstests claim true.

kernel/src/device/registry/block.rs
tatetian, line 56: **Add doc comments to both aliases and contrast them with `st_blksize`**: `BLKGETSIZE64` and `BLKSSZGET` have deceptively similar names — both are "get something size" on a block device — and on top of that there is a *third* nearby concept, `struct stat::st_blksize`, which is yet another "block size" that userspace regularly consults. Readers of this file have no quick way to tell them apart. Crucially, had the "return value is the O_DIRECT alignment granularity" property been written down for `BLKSSZGET`, the ext2-alignment bug flagged in the previous comment would likely have been caught at implementation time rather than at review — proper semantics documentation is load-bearing here, not decoration.

A concrete suggestion:

```rust
/// `BLKGETSIZE64` — total size of the block device in bytes.
///
/// The reported value is the raw capacity of the underlying device, independent of
/// any filesystem mounted on top of it or of the partition table. It is the value
/// that tools such as `blockdev --getsize64` or `lsblk -b` report.
///
/// Do not confuse with `struct stat::st_size` on a block device node, which, in
/// Linux, also reports the total device size — the two happen to agree for a whole
/// disk, but `BLKGETSIZE64` is the canonical query and is the one userspace uses
/// on partitions and loop devices.
pub(super) type BlkGetSize64 = ioc!(BLKGETSIZE64, 0x12, 114, OutData<u64>);

/// `BLKSSZGET` — logical block (sector) size of the block device, in bytes.
///
/// This is the smallest unit of I/O the device can address, and — importantly —
/// it is the **minimum alignment required for O_DIRECT I/O** on files backed by
/// this device (both buffer address and offset must be a multiple of this value).
/// Benchmarks and filesystem tests (e.g. `xfstests`, LTP `preadv03`/`pwritev03`)
/// rely on this ioctl to size O_DIRECT buffers. If the effective alignment
/// enforced by the filesystem layered on top (e.g. ext2's 4 KiB block) is larger
/// than the hardware sector, this ioctl must return that larger value — otherwise
/// user programs will align correctly for the device but still hit `EINVAL` at
/// the filesystem.
///
/// Do not confuse with `struct stat::st_blksize`, reported by `fstat(2)`. The two
/// are related but distinct:
///
/// - `BLKSSZGET` is a *correctness* query: the device-layer minimum I/O granularity.
///   Violating it causes O_DIRECT and raw block-device I/O to fail with `EINVAL`.
/// - `st_blksize` is a *performance* hint: the "preferred" I/O block size for
///   buffered reads/writes. Violating it is slower but not an error.
///
/// For a raw block device, `st_blksize` is typically set to `BLKSSZGET`'s value
/// (or a small multiple thereof); for a regular file, it is the filesystem's
/// I/O block size.
pub(super) type BlkGetSectorSize = ioc!(BLKSSZGET, 0x12, 104, NoData);
```

Writing these down makes the `BLKGETSIZE64` / `BLKSSZGET` / `st_blksize` distinction impossible to miss, and — per the previous comment — pins down the "must match the effective O_DIRECT alignment" contract that the current implementation violates.

kernel/src/device/registry/block.rs
tatetian, line 1: **Wire in LTP tests that query `BLKSSZGET` before issuing O_DIRECT**: The ext2-alignment bug flagged above would have been caught by any test that actually uses the returned value to align a subsequent direct I/O. LTP already has two such tests that follow exactly this pattern and would exercise the contract end-to-end:

- [`testcases/kernel/syscalls/preadv/preadv03.c`](https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/preadv/preadv03.c) — opens a block device, calls `SAFE_IOCTL(dev_fd, BLKSSZGET, &ret)`, then issues `O_DIRECT` `preadv` at that alignment and asserts success.
- [`testcases/kernel/syscalls/pwritev/pwritev03.c`](https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/pwritev/pwritev03.c) — the symmetric case for `pwritev`.

With the PR's current 512-byte return, both tests would fail on ext2-backed block devices (the `preadv`/`pwritev` calls return `EINVAL` from the 4 KiB alignment check). With the fix suggested in the comment above (`max(SECTOR_SIZE, BLOCK_SIZE)`), they would pass. So these are near-perfect regression tests for the behaviour this PR is trying to deliver.

Concrete suggestion:

1. Add an LTP runner entry (or a small curated subset) to the test harness — something like `test/src/apps/ltp/preadv03.rs` / a Makefile line under `test/syscall/`, mirroring how existing LTP binaries are pulled in (grep the tree for `ltp` / `linux-test-project` to see the current plumbing).
2. At minimum, include `preadv03` and `pwritev03`; `ioctl_loop05` is a useful stretch goal but depends on loop-device support.

If full LTP integration is out of scope for this PR, a minimal C program mirroring `preadv03`'s flow (`open` → `BLKSSZGET` → `posix_memalign` → `O_DIRECT pread`) placed under `test/` would already catch the bug; point the follow-up at pulling in LTP proper later.

tatetian (commented):
It is kind of weird to see such a long review for such a short PR...

kernel/src/device/registry/block.rs
Fischer0522: I tried to enable related test cases (`preadv03.c`, `pwritev03.c`). But these cases failed because currently the block device doesn't support `lseek`, and this situation is mentioned in #3212.
The error msg is :
```
mkfs.ext2: lseek(0, 2): Invalid argument
```

kernel/src/device/registry/block.rs
```diff
@@ -65,10 +66,15 @@ pub(super) fn init_in_first_process(path_resolver: &PathResolver) -> Result<()>
 }
 
 mod ioctl_defs {
-    use crate::util::ioctl::{OutData, ioc};
+    use crate::util::ioctl::{NoData, OutData, ioc};
 
     // Reference: <https://elixir.bootlin.com/linux/v6.18/source/include/uapi/linux/fs.h>
+
+    /// Returns the device size in bytes (`BLKGETSIZE64`).
     pub(super) type BlkGetSize64 = ioc!(BLKGETSIZE64, 0x12, 114, OutData<u64>);
```
tatetian, line 74: Remove from the comment `BLKGETSIZE64`, which is a piece of redundant info as it already appears in `ioc!(BLKGETSIZE64, ...)`.

kernel/src/device/registry/block.rs
```diff
@@ -65,10 +66,15 @@ pub(super) fn init_in_first_process(path_resolver: &PathResolver) -> Result<()>
...
+
+    /// Returns the logical sector size of the block device (`BLKSSZGET`).
+    pub(super) type BlkGetSectorSize = ioc!(BLKSSZGET, 0x12, 104, NoData);
```
tatetian, line 89: I suggest persisting the following knowledge into our code as a doc comment:

```rust
/// Returns the logical sector size of the block device
///
/// This is the smallest unit of I/O the device can address, and — importantly —
/// it is the **minimum alignment required for O_DIRECT I/O** on files backed by
/// this device (both buffer address and offset must be a multiple of this value).
/// Benchmarks and filesystem tests (e.g. `xfstests`, LTP `preadv03`/`pwritev03`)
/// rely on this ioctl to size O_DIRECT buffers. If the effective alignment
/// enforced by the filesystem layered on top (e.g. ext2's 4 KiB block) is larger
/// than the hardware sector, this ioctl must return that larger value — otherwise
/// user programs will align correctly for the device but still hit `EINVAL` at
/// the filesystem.
```

tatetian (commented):
The PR looks mostly good. Just two minor issues.

kernel/src/device/registry/block.rs
tatetian: Ok. Thanks for trying.

kernel/src/device/registry/block.rs
Fischer0522: Done.

kernel/src/device/registry/block.rs
Fischer0522: Done.

tatetian (approved):

PR #3115, opened 2026-04-10
Fix races in sending parent death signals

lrh2000:
https://github.com/asterinas/asterinas/blob/ec037a4665f67c182e44013dca655006cf3a704c/kernel/src/process/exit.rs#L27-L29

There are two problems:
 1. A new child may appear between the `send_parent_death_signal` and `move_children_to_reaper_process`. For example, if the process is a reaper and a child becomes a zombie, then the grandchild becomes the process's child.
 2. The signal should be sent after updating the child's parent. Otherwise, when the process handles the signal, it will still see the old parent.

The first commit fixes the two problems.

---

Additionally, the order of the methods in `process/exit.rs` isn't ideal. It would be better to follow the [_top-down read_](https://github.com/asterinas/asterinas/blob/ec037a4665f67c182e44013dca655006cf3a704c/book/src/to-contribute/coding-guidelines/general-guidelines/README.md?plain=1#L198) rule to reorder them. This is done in the second commit.

### DIFF
--- a/kernel/src/process/exit.rs
+++ b/kernel/src/process/exit.rs
@@ -21,9 +21,11 @@ pub(super) fn exit_process(current_process: &Process) {
     // Drop fields in `Process`.
     current_process.lock_vmar().set_vmar(None);
 
-    send_parent_death_signal(current_process);
-
-    move_children_to_reaper_process(current_process);
+    // Move the children to the reaper process and send them signals. The children should see a new
+    // parent when they receive the signal.
+    let children = move_children_to_reaper_process(current_process);
+    send_parent_death_signal(&children);
+    drop(children);
 
     // Mark the current process as a zombie. After this point, it may be reaped concurrently and
     // removed from the process tree. For example, the process may no longer have a live parent if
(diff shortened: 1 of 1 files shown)


StevenJiang1110:
> A new child may appear between the send_parent_death_signal and move_children_to_reaper_process. For example, if the process is a reaper and a child becomes a zombie, then the grandchild becomes the process's child.

Sorry, I don't quite follow this statement — could you elaborate a bit more?

Suppose we have three processes: the current process(1), a child(2), and a grandchild(3) and process 1 is the reaper. Do you mean that when process 2 exits, its `children` field could gain a new entry?  I don't understand why that would happen. Or do you mean that when process 2 exits, process 1's `children` field could gain a new child? If so, I don't think that should be a problem — but I may be missing something.

lrh2000:
> Or do you mean that when process 2 exits, process 1's `children` field could gain a new child? If so, I don't think that should be a problem — but I may be missing something.

Yes. I mean this case. I don't know what do you mean by "not a problem".

I think the following case is somewhat problematic:
```
```text
[ Process 1 (reaper)]                     [ Process 2 ]                         [ Process 3 ]

exit_process()

   send_parent_death_signal()
                                          { handle parent-death signal }
                                             getppid() == Process 1

                                          exit_process()
                                          { exit_process() finishes }

                                                                                { handle parent-death signal }
                                                                                   X: getppid() == Process 1

   move_children_to_reaper_process()

                                                                                 Y: getppid() == Process 1's reaper
```

At _time X_, Process 3 thinks its parent is Process 1 (a reaper). However, Process 1 then dies. By _time Y_, Process 3's parent has already changed to Process 1's reaper (init process or whatever). But between _time X_ and _time Y_, no parent death signal is delivered. So I think this will cause Process 3 to miss the parent-death event?

StevenJiang1110:
> At time X, Process 3 thinks its parent is Process 1 (a reaper). However, Process 1 then dies. By time Y, Process 3's parent has already changed to Process 1's reaper (init process or whatever). But between time X and time Y, no parent death signal is delivered. So I think this will cause Process 3 to miss the parent-death event?

Got it, thanks for the explanation! So process 3 could observe its parent change twice — first from process 2 to process 1, and then from process 1 to process 1's reaper — yet only receive the parent-death signal once. This is indeed a subtle bug and would be very difficult to spot.

StevenJiang1110 (approved):

PR #3116, opened 2026-04-10
Support cgroup CPU sub-controller

cchanging:
This PR enable the cgroup CPU sub-controller and support a basic attribute `cpu.stat`.

Required by #2911.

Kata checks at startup whether the CPU sub-controller is supported, and it does so using a fairly aggressive `unwrap`-based check... So although Kata does not actually use operations on that sub-controller afterward, in order to avoid adding too many patches to Kata, we would still like to enable at least the most basic CPU sub-controller functionality.

### DIFF
--- a/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
@@ -1,0 +1,177 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use alloc::sync::Arc;
+use core::sync::atomic::{AtomicBool, Ordering};
+
+use aster_systree::{Error, Result, SysAttrSetBuilder, SysPerms, SysStr};
+use aster_util::{per_cpu_counter::PerCpuCounter, printer::VmPrinter};
+use ostd::{
+    cpu::CpuId,
+    mm::{VmReader, VmWriter},
+    task::atomic_mode::AsAtomicModeGuard,
+    timer::Jiffies,
+};
+
+use crate::{
+    fs::cgroupfs::systree_node::{CgroupSysNode, CgroupSystem},
+    process::Process,
+};
+
+/// A sub-controller responsible for CPU resource management in the cgroup subsystem.
+///
+/// This controller always exists so that `cpu.stat` remains readable even before `+cpu`
+/// is enabled. When inactive, only the base usage fields are exposed. The `user` and
+/// `system` counters store CPU time in units of one [`Jiffies`] per increment, and
+/// `usage_usec` is derived from their sum when `cpu.stat` is read.
+pub struct CpuController {
+    is_enabled: AtomicBool,
+    user: PerCpuCounter,
+    system: PerCpuCounter,
+}
+
+/// Specifies the cgroup CPU sub-controller receives one [`Jiffies`] of charge.
+#[derive(Clone, Copy)]
+pub enum CpuStatKind {
+    /// Charges one [`Jiffies`] to `user_usec`.
+    User,
+    /// Charges one [`Jiffies`] to `system_usec`.
+    System,
+}
+
+impl CpuController {
+    pub(super) fn init_attr_set(builder: &mut SysAttrSetBuilder, _is_root: bool) {
+        builder.add(SysStr::from("cpu.stat"), SysPerms::DEFAULT_RO_ATTR_PERMS);
+    }
+
+    fn new(is_enabled: bool) -> Self {
+        Self {
+            is_enabled: AtomicBool::new(is_enabled),
+            user: PerCpuCounter::new(),
+            system: PerCpuCounter::new(),
+        }
+    }
+
+    fn write_cpu_stat_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        /// Converts a per-CPU counter expressed in jiffies to microseconds.
+        fn counter_usec(counter: &PerCpuCounter) -> u64 {
+            let jiffies = u64::try_from(counter.sum_all_cpus()).unwrap_or(u64::MAX);
+            u64::try_from(Jiffies::new(jiffies).as_duration().as_micros()).unwrap_or(u64::MAX)
+        }
+
+        let mut printer = VmPrinter::new_skip(writer, offset);
+
+        let user_usec = counter_usec(&self.user);
+        let system_usec = counter_usec(&self.system);
+        writeln!(
+            printer,
+            "usage_usec {}",
+            user_usec.saturating_add(system_usec)
+        )?;
+        writeln!(printer, "user_usec {}", user_usec)?;
+        writeln!(printer, "system_usec {}", system_usec)?;
+
+        if self.is_enabled.load(Ordering::Relaxed) {
+            // TODO: Support CPU bandwidth control statistics. These fields are reported as `0`
+            // for now because the cgroup CPU sub-controller does not yet implement throttling or
+            // burst accounting.
+            writeln!(printer, "nr_periods 0")?;
+            writeln!(printer, "nr_throttled 0")?;
+            writeln!(printer, "throttled_usec 0")?;
+            writeln!(printer, "nr_bursts 0")?;
+            writeln!(printer, "burst_usec 0")?;
+        }
+
+        Ok(printer.bytes_written())
+    }
+
+    pub(super) fn enable(&self) {
+        self.is_enabled.store(true, Ordering::Relaxed);
+    }
+
+    pub(super) fn disable(&self) {
+        self.is_enabled.store(false, Ordering::Relaxed);
+    }
+
+    /// Accounts one [`Jiffies`] of CPU time on `cpu`.
+    fn account(&self, cpu: CpuId, stat_kind: CpuStatKind) {
+        match stat_kind {
+            CpuStatKind::User => self.user.add_on_cpu(cpu, 1),
+            CpuStatKind::System => self.system.add_on_cpu(cpu, 1),
+        }
+    }
+}
+
+impl super::SubControl for CpuController {
+    fn read_attr_at(&self, name: &str, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        if name != "cpu.stat" {
+            return Err(Error::AttributeError);
+        }
+
+        self.write_cpu_stat_at(offset, writer)
+    }
+
+    fn write_attr(&self, _name: &str, _reader: &mut VmReader) -> Result<usize> {
+        Err(Error::AttributeError)
+    }
+}
+
+impl super::SubControlStatic for CpuController {
+    fn new(is_root: bool, is_active: bool) -> Self {
+        Self::new(is_root || is_active)
+    }
+
+    fn type_() -> super::SubCtrlType {
+        super::SubCtrlType::Cpu
+    }
+
+    fn read_from(controller: &super::Controller) -> Arc<super::SubController<Self>> {
+        controller.cpu.read().get().clone()
+    }
+}
+
+impl super::SubController<CpuController> {
+    pub(super) fn enable(&self) {
+        self.inner.as_ref().unwrap().enable();
+    }
+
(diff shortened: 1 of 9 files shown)


kernel/src/process/process/timer_manager.rs
```diff
@@ -56,6 +62,7 @@ fn update_cpu_time() {
         // `None` here.
         return;
     };
+    let current_cpu = CpuId::current_racy();
```
lrh2000: This has nothing to do with `// Retrieve some info about the current task`.

Insert a blank line and have the comment:
```rust
    // No races because we're in IRQs.
    let current_cpu = CpuId::current_racy();
```

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -123,6 +138,7 @@ impl From<SubCtrlType> for SubCtrlSet {
         match ctrl_type {
             SubCtrlType::Memory => Self::MEMORY,
             SubCtrlType::CpuSet => Self::CPUSET,
+            SubCtrlType::Cpu => Self::CPU,
```
lrh2000, line 140: So `CpuSet` should be sorted before `Cpu`? Not sure...

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,168 @@
...
+        while let Some(node) = current {
+            node.inner.as_ref().unwrap().account(on_cpu, stat_kind);
+            current = node.parent.as_deref();
+        }
+    }
+}
+
+impl super::Controller {
+    /// Charges one jiffy in the CPU sub-controller hierarchy.
+    pub fn charge_cpu_time(&self, on_cpu: CpuId, stat_kind: CpuStatKind) {
```
lrh2000: I don't think it should be `pub`?

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,168 @@
...
+        let guard = self.cpu.read();
```
lrh2000: Actually we have `cgroup_guard` so we should be able to use `read_with`?

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::{
+    fs::cgroupfs::{CgroupSysNode, CgroupSystem},
+    process::Process,
+};
+
+/// A sub-controller responsible for CPU resource management in the cgroup subsystem.
+///
+/// This controller always exists so that `cpu.stat` remains readable even before `+cpu`
+/// is enabled. When inactive, only the base usage fields are exposed. Each counter stores
+/// CPU time in units of one `jiffy` per increment.
```
lrh2000: I think you should use either `one jiffy` or just say `` one [`Jiffies`] `` (although the latter may not be grammatically correct, it does provide more information).

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,168 @@
...
+        let jiffies = u64::try_from(counter.sum_all_cpus()).unwrap_or(u64::MAX);
+        u64::try_from(Jiffies::new(jiffies).as_duration().as_micros()).unwrap_or(u64::MAX)
+    }
+
+    fn write_cpu_stat_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
+        let mut printer = VmPrinter::new_skip(writer, offset);
+        let usage_usec = Self::counter_usec(&self.usage);
+        let user_usec = Self::counter_usec(&self.user);
+        let system_usec = Self::counter_usec(&self.system);
+
```
lrh2000, line 62: The following is more reasonable?
```rust

        let usage_usec = Self::counter_usec(&self.usage);
        let user_usec = Self::counter_usec(&self.user);
        let system_usec = Self::counter_usec(&self.system);
        writeln!(printer, "usage_usec {}", usage_usec)?;
        writeln!(printer, "user_usec {}", user_usec)?;
        writeln!(printer, "system_usec {}", system_usec)?;

        if self.is_enabled.load(Ordering::Relaxed) {
```

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,168 @@
...
+        Self {
+            is_enabled: AtomicBool::new(is_enabled),
+            usage: PerCpuCounter::new(),
+            user: PerCpuCounter::new(),
+            system: PerCpuCounter::new(),
+        }
+    }
+
+    /// Converts a per-CPU counter expressed in jiffies to microseconds.
+    fn counter_usec(counter: &PerCpuCounter) -> u64 {
```
lrh2000: Define it after `write_cpu_stat_at`? Or move it inside `write_cpu_stat_at`?

lrh2000 (commented):
Looks good. Just minor comments here.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
cchanging: Reading Linux `cgroup.controllers` shows: 

```
cpuset cpu io memory hugetlb pids rdma misc
```

I tried to find some reference for this ordering, but couldn’t find one... So I don’t know why Linux defines it in this order. For now, though, I’ve kept our ordering consistent with Linux.

test/initramfs/src/regression/process/cgroup.sh
```diff
@@ -118,34 +157,95 @@ verify "Initial cgroup.events populated=0" \
...
-    "memory pids"
+verify "child controllers contain cpu" \
+    "tr ' ' '\\n' < cgroup.controllers | grep -x cpu" \
+    "cpu"
+verify "child controllers contain memory" \
+    "tr ' ' '\\n' < cgroup.controllers | grep -x memory" \
+    "memory"
+verify "child controllers contain pids" \
+    "tr ' ' '\\n' < cgroup.controllers | grep -x pids" \
+    "pids"
```
lrh2000, line 230: I assume it's because you previously used an inconsistent order with Linux. Is now the right time to revert the changes?

test/initramfs/src/regression/process/cgroup.sh
```diff
@@ -118,34 +157,95 @@ verify "Initial cgroup.events populated=0" \
...
+echo "+cpu +memory +pids" > cgroup.subtree_control
+verify "root subtree_control contains cpu" \
+    "tr ' ' '\\n' < cgroup.subtree_control | grep -x cpu" \
+    "cpu"
+verify "root subtree_control contains memory" \
+    "tr ' ' '\\n' < cgroup.subtree_control | grep -x memory" \
+    "memory"
+verify "root subtree_control contains pids" \
+    "tr ' ' '\\n' < cgroup.subtree_control | grep -x pids" \
+    "pids"
```
lrh2000: Same here.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -10,17 +10,22 @@ use core::{
 use aster_systree::{Error, Result, SysAttrSetBuilder, SysBranchNode, SysObj};
 use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
 use bitflags::bitflags;
+pub use cpu::{CpuStatKind, charge_cpu_time};
```
lrh2000: Consider `pub(super) cpu` to avoid exporting the same function/type twice?

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    pub(super) fn enable(&self) {
+        self.is_enabled.store(true, Ordering::Relaxed);
+    }
+
+    pub(super) fn disable(&self) {
+        self.is_enabled.store(false, Ordering::Relaxed);
+    }
+
+    /// Accounts one [`Jiffies`] of CPU time on `on_cpu`.
```
lrh2000: ``` on `on_cpu` ``` seems a bit odd? Use `cpu`?

```rust
> grep -R "add_on_cpu(.*1" kernel
kernel/comps/softirq/src/stats.rs:    IRQ_COUNTERS.get().unwrap()[irq_num as usize].add_on_cpu(CpuId::current_racy(), 1);
kernel/comps/softirq/src/lib.rs:                .add_on_cpu(CpuId::current_racy(), 1);
kernel/src/process/clone.rs:            .add_on_cpu(CpuId::current_racy(), 1);
kernel/src/thread/mod.rs:        .add_on_cpu(CpuId::current_racy(), 1);
kernel/src/time/cpu_time_stats.rs:        self.user.add_on_cpu(cpu, 1);
kernel/src/time/cpu_time_stats.rs:        self.system.add_on_cpu(cpu, 1);
kernel/src/time/cpu_time_stats.rs:        self.idle.add_on_cpu(cpu, 1);
```

---

Also I don't realize why it is necessary to pass the CPU ID so many levels from `update_cpu_time` down to this method? Can't we just get the current CPU ID in `account_hierarchy`?

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,177 @@
...
+        on_cpu: CpuId,
+        stat_kind: CpuStatKind,
+    ) {
+        self.cpu
+            .read_with(guard)
+            .account_hierarchy(on_cpu, stat_kind);
+    }
+}
+
+pub fn charge_cpu_time(process: &Process, on_cpu: CpuId, stat_kind: CpuStatKind) {
```
lrh2000: This actually needs documentation comments since it's a public method exposed to another crate-level top module. We have good documentation comments for many private methods above, but this method lacks them. That's a bit odd.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
lrh2000: It makes sense. Thanks for your explanation!

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.is_enabled.store(true, Ordering::Relaxed);
+    }
+
+    pub(super) fn disable(&self) {
+        self.is_enabled.store(false, Ordering::Relaxed);
+    }
+
+    /// Accounts one [`Jiffies`] of CPU time on `cpu`.
+    fn account(&self, cpu: CpuId, stat_kind: CpuStatKind) {
+        self.usage.add_on_cpu(cpu, 1);
```
lrh2000: As a (minor) optimization, I found that this can be removed in favor of summing during the printing process.
```rust
        writeln!(printer, "usage_usec {}", user_usec.saturating_add(system_usec))?;
```

ostd/src/sync/rcu/mod.rs
```diff
@@ -379,6 +379,12 @@ impl<P: NonNullPtr + Send> RcuReadGuard<'_, P> {
     }
 }
 
+impl<P: NonNullPtr> AsAtomicModeGuard for RcuReadGuard<'_, P> {
+    fn as_atomic_mode_guard(&self) -> &dyn InAtomicMode {
+        self.0._inner_guard.as_atomic_mode_guard()
```
lrh2000:
```suggestion
        self.0.inner_guard.as_atomic_mode_guard()
```

Actually no need to suppress the unused fields warning now?

lrh2000 (approved):
LGTM other than two minor points. Thanks for your contribution!

cqs21 (approved):

PR #3117, opened 2026-04-10
Add `FsNotifyMode` and suppress notifications for `O_PATH` fds

BattiestStone4:
## Summary

In Linux, `O_PATH` file descriptors carry `FMODE_NONOTIFY` in `f_mode`, which suppresses all fsnotify events. Asterinas did not implement this — `O_PATH` fds would trigger `IN_OPEN` / `IN_CLOSE_NOWRITE` notifications like regular fds.

- `O_PATH` is already observable on the `FileLike` trait via `status_flags()`, so `on_open` / `on_close` now suppress notifications directly on that flag, with the shared `O_PATH`-check + subscriber lookup factored into a `notifiable_path` helper.
- No check is needed in `on_access` / `on_modify`: an `O_PATH` `InodeHandle` is built with empty `Rights`, so reads and writes fail with `EBADF` before any access / modify notification could fire.
- Add a regression test verifying that `O_PATH` opens suppress inotify events while normal opens still generate them, registered in `run_test.sh`.

## Test plan

- [x] `inotify_o_path` regression test: both subtests pass (`o_path_open_suppresses_events`, `normal_open_generates_events`)
- [x] Existing inotify regression tests unaffected (`inotify_align`, `inotify_poll`, `inotify_unlink`)

### DIFF
--- a/kernel/src/fs/file/file_attr/fs_notify_mode.rs
+++ b/kernel/src/fs/file/file_attr/fs_notify_mode.rs
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Filesystem notification mode flags.
+//!
+//! These flags control which fsnotify events are suppressed for an
+//! opened file descriptor. They correspond to the `FMODE_NONOTIFY`
+//! and `FMODE_NONOTIFY_PERM` flags in Linux's `f_mode` field.
+
+use bitflags::bitflags;
+
+bitflags! {
+    /// Controls which filesystem notification events are suppressed for a file.
+    pub struct FsNotifyMode: u8 {
+        /// Suppresses all fsnotify events for this file descriptor.
+        ///
+        /// Set on `O_PATH` file descriptors and fanotify-internal fds
+        /// to prevent recursive notification loops.
+        const NONOTIFY      = 1 << 0;
+        /// Suppresses permission-class fsnotify events only.
+        ///
+        /// Reserved for future fanotify support. When set without
+        /// `NONOTIFY`, only permission events (`FAN_ACCESS_PERM`,
+        /// `FAN_OPEN_PERM`) are suppressed.
+        const NONOTIFY_PERM = 1 << 1;
+    }
+}
--- a/kernel/src/fs/vfs/notify/mod.rs
+++ b/kernel/src/fs/vfs/notify/mod.rs
@@ -12,7 +12,7 @@ use ostd::sync::RwLock;
 
 use crate::{
     fs::{
-        file::{AccessMode, FileLike, InodeType},
+        file::{AccessMode, FileLike, FsNotifyMode, InodeType},
         vfs::path::Path,
     },
     prelude::*,
@@ -276,7 +276,9 @@ define_atomic_version_of_integer_like_type!(FsEvents, {
 
 /// Notifies that a file was accessed.
 pub fn on_access(file: &Arc<dyn FileLike>) {
-    // TODO: Check fmode flags (FMODE_NONOTIFY, FMODE_NONOTIFY_PERM).
+    if file.fs_notify_mode().contains(FsNotifyMode::NONOTIFY) {
+        return;
+    }
     let path = file.path();
 
     if !path.fs().fs_event_subscriber_stats().has_any_subscribers() {
@@ -287,7 +289,9 @@ pub fn on_access(file: &Arc<dyn FileLike>) {
 
 /// Notifies that a file was modified.
 pub fn on_modify(file: &Arc<dyn FileLike>) {
-    // TODO: Check fmode flags (FMODE_NONOTIFY, FMODE_NONOTIFY_PERM).
+    if file.fs_notify_mode().contains(FsNotifyMode::NONOTIFY) {
+        return;
+    }
     let path = file.path();
 
     if !path.fs().fs_event_subscriber_stats().has_any_subscribers() {
@@ -379,7 +383,9 @@ pub fn on_create(file_path: &Path, name: impl FnOnce() -> String) {
 
 /// Notifies that a file was opened.
 pub fn on_open(file: &Arc<dyn FileLike>) {
-    // TODO: Check fmode flags (FMODE_NONOTIFY, FMODE_NONOTIFY_PERM).
+    if file.fs_notify_mode().contains(FsNotifyMode::NONOTIFY) {
+        return;
+    }
     let path = file.path();
 
     if !path.fs().fs_event_subscriber_stats().has_any_subscribers() {
@@ -390,7 +396,9 @@ pub fn on_open(file: &Arc<dyn FileLike>) {
 
 /// Notifies that a file was closed.
 pub fn on_close(file: &Arc<dyn FileLike>) {
-    // TODO: Check fmode flags (FMODE_NONOTIFY, FMODE_NONOTIFY_PERM).
+    if file.fs_notify_mode().contains(FsNotifyMode::NONOTIFY) {
+        return;
+    }
     let path = file.path();
 
     if !path.fs().fs_event_subscriber_stats().has_any_subscribers() {
--- a/kernel/src/fs/vfs/notify/mod.rs
+++ b/kernel/src/fs/vfs/notify/mod.rs
@@ -7,7 +7,7 @@ use bitflags::bitflags;
 
 use crate::{
     fs::{
-        file::{AccessMode, FileLike, InodeType},
+        file::{AccessMode, FileLike, InodeType, StatusFlags},
         vfs::path::Path,
     },
     prelude::*,
@@ -271,7 +271,6 @@ define_atomic_version_of_integer_like_type!(FsEvents, {
 
 /// Notifies that a file was accessed.
 pub fn on_access(file: &Arc<dyn FileLike>) {
-    // TODO: Check fmode flags (FMODE_NONOTIFY, FMODE_NONOTIFY_PERM).
     let path = file.path();
 
     if !path.fs().fs_event_subscriber_stats().has_any_subscribers() {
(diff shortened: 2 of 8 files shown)


kernel/src/fs/file/file_attr/fs_notify_mode.rs
```diff
@@ -0,0 +1,26 @@
...
+use bitflags::bitflags;
+
+bitflags! {
+    /// Controls which filesystem notification events are suppressed for a file.
+    pub struct FsNotifyMode: u8 {
+        /// Suppresses all fsnotify events for this file descriptor.
+        ///
+        /// Set on `O_PATH` file descriptors and fanotify-internal fds
+        /// to prevent recursive notification loops.
+        const NONOTIFY      = 1 << 0;
```
tatetian: I am skeptical about the usefulness of adding this new mode type.

Currently, the only user of `FsNotifyMode` is `O_PATH`. But if an instance of `InodeHandle` is opened with `O_PATH`, then its `rights` would be [set](https://github.com/asterinas/asterinas/blob/497b617c427cc2d2b15f531e2b15f7a91e30273a/kernel/src/fs/file/inode_handle.rs#L61) to empty. This would make any attempt to read or write this `InodeHandle` fail. If the I/O operation fails in the first place, then `fs::vfs::notify::on_access` or `fs::vfs::notify::on_modify` would never be invoked. So why bother introducing this new mode and check it in `on_access` and `on_modify`?

----

Edit: Ok. I got it. It has no real effect for suppressing notifications on I/O operations for `O_PATH`-opened `InodeHandle`. But it is still useful to suppressing inotify events when the FD is opened and closed.

But this brings the question: why only checking the value of `FsNotifyMode` for `on_open`, `on_close`, `on_access`, and `on_modify`? If you want to this mode to be respected regardless of `InodeHandle::rights`, all `on_xxx` hooks of `vfs::notify` should check this mode as well.

kernel/src/fs/file/file_attr/fs_notify_mode.rs
```diff
@@ -0,0 +1,26 @@
...
+        /// Suppresses permission-class fsnotify events only.
+        ///
+        /// Reserved for future fanotify support. When set without
+        /// `NONOTIFY`, only permission events (`FAN_ACCESS_PERM`,
+        /// `FAN_OPEN_PERM`) are suppressed.
+        const NONOTIFY_PERM = 1 << 1;
```
tatetian: `NONOTIFY` and `NONOTIFY_PERM` are not symbols known to the userspace. We do not need to follow the naming of Linux's internals, which is not Rust style. So rename them:

* `NONOTIFY` --> `NO_NOTIFY`
* `NONOTIFY_PERM` --> `NO_NOTIFY_PERM`

kernel/src/fs/vfs/notify/mod.rs
```diff
@@ -271,7 +271,9 @@ define_atomic_version_of_integer_like_type!(FsEvents, {
 
 /// Notifies that a file was accessed.
 pub fn on_access(file: &Arc<dyn FileLike>) {
-    // TODO: Check fmode flags (FMODE_NONOTIFY, FMODE_NONOTIFY_PERM).
+    if file.fs_notify_mode().contains(FsNotifyMode::NONOTIFY) {
```
tatetian: **`NONOTIFY` guards in `on_access`/`on_modify` are dead code; the mode is over-built for its one user** (significant): An `O_PATH` `InodeHandle` is constructed with `Rights::empty()` (`inode_handle.rs:61`), so `InodeHandle::read`/`write` reject it with `EBADF` (`inode_handle.rs:260-261`, `:280-281`) — as do all the other rights-gated entry points (`readdir` `:124`, preadv/pwritev/pread64/pwrite64/fallocate, sendfile). The syscalls only notify *after* a successful transfer (`read.rs:41` guards `on_access` with `if read_len > 0`; `write.rs:41` guards `on_modify` with `if write_len > 0`). So for the sole current user the I/O fails first and these two guards can never fire. Only `on_open` (`open.rs:66`) and `on_close` (`close.rs:33`) are reached unconditionally for `O_PATH`, so only their guards do real work. Recommendation: drop the `NONOTIFY` checks from `on_access` and `on_modify` (here and at line 287), and reconsider the abstraction's size — `NONOTIFY_PERM` is read nowhere and the `InodeHandle.fs_notify_mode` field merely re-encodes `status_flags.contains(O_PATH)`. Either collapse to suppression in `on_open`/`on_close` only, or, if the `f_mode`-style mode is kept for future fanotify parity, drop `NONOTIFY_PERM` and document the field as intentionally decoupled from `O_PATH`. With the two live guards now non-trivial, factor their shared body (`NONOTIFY` check → `has_any_subscribers` → `notify_parent`) into one private helper.

kernel/src/fs/file/file_attr/fs_notify_mode.rs
```diff
@@ -0,0 +1,26 @@
...
+    /// Controls which filesystem notification events are suppressed for a file.
+    pub struct FsNotifyMode: u8 {
+        /// Suppresses all fsnotify events for this file descriptor.
+        ///
+        /// Set on `O_PATH` file descriptors and fanotify-internal fds
+        /// to prevent recursive notification loops.
+        const NONOTIFY      = 1 << 0;
+        /// Suppresses permission-class fsnotify events only.
+        ///
+        /// Reserved for future fanotify support. When set without
```
tatetian: **Doc comments describe behavior that isn't implemented** (minor): `NONOTIFY_PERM`'s doc promises concrete semantics, but no code reads the flag and Asterinas emits no permission events yet, so the prose will silently rot; the "Reserved for future fanotify support" line already suffices. The same over-description appears at `kernel/src/fs/file/file_handle.rs:144`, where the `fs_notify_mode()` default-method doc explains the impl ("returns an empty `FsNotifyMode`, meaning all notifications are enabled") instead of the contract — and "all notifications are enabled" is imprecise; empty means *no suppression*. Trim both to intent/contract, e.g.:

```rust
/// Suppresses permission-class fsnotify events only.
/// Reserved for future fanotify support; not yet honored.
const NONOTIFY_PERM = 1 << 1;
```

tatetian (commented):
The following comments were generated by my in-house review comments. Take them with a grain of salt.

test/initramfs/src/regression/fs/inotify/inotify_o_path.c
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: MPL-2.0
```
tatetian, line 1: This test does not run at all. It should be added [in this file](https://github.com/asterinas/asterinas/blob/f7ff85597d892ec7476489216672b0ad61b7090f/test/initramfs/src/regression/fs/run_test.sh#L117-L119) along with other existing inotify C test programs.

tatetian (commented):
The new test is

kernel/src/fs/file/file_attr/fs_notify_mode.rs
BattiestStone4: The reason only these four hooks check the mode is structural: `on_open`/`on_close`/`on_access`/`on_modify` are the only `vfs::notify` hooks that receive a `FileLike`, i.e. an open file description. The other eight take `&Path` or `&Arc<dyn Inode>` and act on the path/inode itself — there is no open instance in scope, so there is no per-fd `f_mode`-equivalent to consult. This matches Linux, where `FMODE_NONOTIFY` lives in `f_mode` and is only checked on the `fsnotify_file()` path; inode/dentry-level events (`fsnotify_create`/`mkdir`/`unlink`) never consult it because there is no `file` there. So "check it in all `on_xxx` hooks" does not really apply — the rest neither can (no `file`) nor should (the event is not tied to a specific fd).

That said, your follow-up is correct: for the only current user (`O_PATH`), the `on_access`/`on_modify` guards are dead code. An `O_PATH` `InodeHandle` is built with `Rights::empty()`, so `read`/`write` fail with `EBADF` before any access/modify notification can fire (and the syscalls only notify after a successful transfer). I have dropped those two guards and keep suppression only where it does real work — `on_open`/`on_close`. More on collapsing the abstraction in the other thread.

kernel/src/fs/vfs/notify/mod.rs
BattiestStone4: Agreed on all points — I verified each claim against the source before acting on it (the `EBADF` rejection in `InodeHandle::read`/`write`, the `if read_len > 0` / `if write_len > 0` guards in the read/write syscalls, and that the field just re-encodes `status_flags.contains(O_PATH)`), and they all hold. So I collapsed the abstraction rather than keeping it:

- Dropped the `NONOTIFY` guards from `on_access`/`on_modify` (dead code, as you noted).
- Removed the `FsNotifyMode` type entirely (including the unused `NONOTIFY_PERM`) and the `InodeHandle.fs_notify_mode` field.
- `on_open`/`on_close` now suppress directly on `file.status_flags().contains(StatusFlags::O_PATH)`, which is already exposed on the `FileLike` trait, and the shared `O_PATH`-check + `has_any_subscribers` prologue is factored into a `notifiable_path` helper.

This keeps the change scoped to the one real behavior (`IN_OPEN`/`IN_CLOSE_NOWRITE` suppression for `O_PATH`) without a mode type that has a single bit and a single caller. If/when fanotify lands, the `f_mode`-style mode can be reintroduced then, with actual readers. This also resolves the doc-comment thread, since the `FsNotifyMode` doc comments are gone.

kernel/src/fs/file/file_attr/fs_notify_mode.rs
BattiestStone4: The `FsNotifyMode` type is removed entirely in the latest revision (see the other thread), so these flag names are gone along with it.

kernel/src/fs/file/file_attr/fs_notify_mode.rs
BattiestStone4: Both doc comments are gone — `FsNotifyMode` (and `NONOTIFY_PERM`) is removed, and the `fs_notify_mode()` trait method along with it.

test/initramfs/src/regression/fs/inotify/inotify_o_path.c
BattiestStone4: Done — `inotify_o_path` is now registered in `run_test.sh` alongside the other inotify tests, and both subtests pass in a regression run.

tatetian (approved):

PR #3121, opened 2026-04-12
Fix invalid signal number panic for clone and clone3

### LINKED ISSUE #3112: Reachable unwrap panic caused by invalid exit signal in `clone` and `clone3`
### Describe the bug

There is a reachable unwrap panic in `SigNum::from_u8()` when calling system call `clone` and `clone3` with invalid exit signal

https://github.com/asterinas/asterinas/blob/b05d53e9661f650ab21abd0bf367fe0632b2df39/kernel/src/process/signal/sig_num.rs#L30

### To Reproduce

Compile the PoC and run

```C
#define _GNU_SOURCE
#include <errno.h>
#include <stdio.h>
#include <sys/syscall.h>
#include <unistd.h>

int main(void) {
  long ret = syscall(SYS_clone, 255UL, 0UL, 0UL, 0UL, 0UL);
  perror("clone");

  printf("clone returned %ld\n", ret);
  return 0;
}
```

```C
#define _GNU_SOURCE
#include <errno.h>
#include <linux/sched.h>
#include <stdint.h>
#include <stdio.h>
#include <string.h>
#include <sys/syscall.h>
#include <unistd.h>

int main(void) {
  struct clone_args args;
  memset(&args, 0, sizeof(args));
  args.exit_signal = 255;

  long ret = syscall(SYS_clone3, &args, sizeof(args));

  printf("clone3 returned %ld\n", ret);
  perror("clone3");
  return 0;
}
```

### Expected behavior

For  `clone`, it shall return success.
For `clone3`, it shall return with `EINVAL`

### Environment

- Official docker asterinas/asterinas:0.17.0-20251217
- Asterinas version: main b05d53e9

### Logs
```
~ # /etc/clone_invalid_exit_signal_panic.c 
[    54.320] ERROR : Uncaught panic:
        invalid signal number
        at /root/asterinas/kernel/src/process/signal/sig_num.rs:30
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(0), real_time: RealTimeAttr { prio: 99, time_slice: 31490620 }, fair: FairAttr { weight: 1024, pending_weight: 0, vruntime: 362744218 } } })
```

```
~ # ./etc/clone3_invalid_exit_signal_panic.c 
[    47.931] ERROR : Uncaught panic:
        invalid signal number
        at /root/asterinas/kernel/src/process/signal/sig_num.rs:30
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(0), real_time: RealTimeAttr { prio: 99, time_slice: 31491100 }, fair: FairAttr { weight: 1024, pending_weight: 0, vruntime: 352539464 } } })
```

Marsman1996:
This PR is going to fix #3112

### DIFF
--- a/kernel/src/syscall/clone.rs
+++ b/kernel/src/syscall/clone.rs
@@ -99,8 +99,15 @@ impl TryFrom<Clone3Args> for CloneArgs {
 
         let flags = CloneFlags::from_bits(value.flags as u32)
             .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid clone flags"))?;
-        let exit_signal =
-            (value.exit_signal != 0).then(|| SigNum::from_u8(value.exit_signal as u8));
+        let exit_signal = if value.exit_signal == 0 {
+            None
+        } else {
+            let exit_signal = u8::try_from(value.exit_signal)
+                .ok()
+                .and_then(|exit_signal| SigNum::try_from(exit_signal).ok())
+                .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid exit signal"))?;
+            Some(exit_signal)
+        };
         if flags.intersects(CloneFlags::CLONE_PARENT | CloneFlags::CLONE_THREAD)
             && exit_signal.is_some()
         {
--- a/test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
+++ b/test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+
+#include <linux/sched.h>
+#include <sys/wait.h>
+#include <sys/syscall.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "../../common/test.h"
+
+static pid_t sys_clone3(struct clone_args *args)
+{
+	return syscall(SYS_clone3, args, sizeof(struct clone_args));
+}
+
+FN_TEST(clone_normalizes_invalid_exit_signal)
+{
+	int status;
+	long pid;
+
+	pid = TEST_SUCC(syscall(SYS_clone, 255UL, 0UL, 0UL, 0UL, 0UL));
+
+	if (pid == 0) {
+		_exit(EXIT_SUCCESS);
+	}
+
+	TEST_RES(waitpid(pid, &status, 0),
+		 _ret == pid && WIFEXITED(status) &&
+			 WEXITSTATUS(status) == EXIT_SUCCESS);
+}
+END_TEST()
+
+FN_TEST(clone3_rejects_invalid_exit_signal)
+{
+	struct clone_args args = {};
+
+	args.exit_signal = 0xff;
+	TEST_ERRNO(sys_clone3(&args), EINVAL);
+
+	args.exit_signal = 0x101;
+	TEST_ERRNO(sys_clone3(&args), EINVAL);
+}
+END_TEST()
--- a/test/initramfs/src/regression/process/run_test.sh
+++ b/test/initramfs/src/regression/process/run_test.sh
@@ -8,6 +8,7 @@ set -e
 
 ./clone3/clone_exit_signal
 ./clone3/clone_files
+./clone3/clone_invalid_exit_signal
 ./clone3/clone_no_exit_signal
 ./clone3/clone_parent
 ./clone3/clone_process
--- a/kernel/src/process/clone.rs
+++ b/kernel/src/process/clone.rs
@@ -125,7 +125,10 @@ impl CloneArgs {
     ) -> Result<Self> {
         const FLAG_MASK: u64 = 0xff;
         let flags = CloneFlags::from(raw_flags & !FLAG_MASK);
-        let exit_signal = raw_flags & FLAG_MASK;
+        let exit_signal = match (raw_flags & FLAG_MASK) as u8 {
+            0 => None,
+            sig_num => SigNum::try_from(sig_num).ok(),
+        };
 
         // Disambiguate the `parent_tid` parameter. The field is used
         // both for `CLONE_PIDFD` and `CLONE_PARENT_SETTID`, so at
@@ -150,7 +153,7 @@ impl CloneArgs {
             pidfd,
             child_tid,
             parent_tid,
-            exit_signal: (exit_signal != 0).then(|| SigNum::from_u8(exit_signal as u8)),
+            exit_signal,
             stack: NonZeroU64::new(stack),
             tls,
             ..Default::default()


Marsman1996:
For now this PR only handle the panic of `clone3`.
In Linux, `clone3` does a strict validation for `exit_signal`:

https://elixir.bootlin.com/linux/v6.16.5/source/kernel/fork.c#L2784-L2786

However, for `clone`, Linux will accept the invalid `exit_signal`, and when the child exits, Linux only sends the signal if it is actually valid:

https://elixir.bootlin.com/linux/v6.16.5/source/kernel/signal.c#L2258-L2260

So my question is: shall we make Asterinas compate to the Linux in `clone`? Or we can just return `EINVAL` for `clone` as well.

kernel/src/syscall/clone.rs
```diff
@@ -99,8 +99,9 @@ impl TryFrom<Clone3Args> for CloneArgs {
 
         let flags = CloneFlags::from_bits(value.flags as u32)
             .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid clone flags"))?;
-        let exit_signal =
-            (value.exit_signal != 0).then(|| SigNum::from_u8(value.exit_signal as u8));
+        let exit_signal = (value.exit_signal != 0)
+            .then(|| SigNum::try_from(value.exit_signal as u8))
+            .transpose()?;
         if flags.intersects(CloneFlags::CLONE_PARENT | CloneFlags::CLONE_THREAD)
```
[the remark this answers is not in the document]
Marsman1996: Yeah, I know it, thank you for the advise. But can you tell how Linux do for `clone`?

kernel/src/syscall/clone.rs
```diff
@@ -99,8 +99,9 @@ impl TryFrom<Clone3Args> for CloneArgs {
 
         let flags = CloneFlags::from_bits(value.flags as u32)
             .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid clone flags"))?;
-        let exit_signal =
-            (value.exit_signal != 0).then(|| SigNum::from_u8(value.exit_signal as u8));
+        let exit_signal = (value.exit_signal != 0)
+            .then(|| SigNum::try_from(value.exit_signal as u8))
```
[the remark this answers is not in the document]
Marsman1996: you are right, I missed the `(args.exit_signal & ~((u64)CSIGNAL)` check here...

lrh2000:
> So my question is: shall we make Asterinas compate to the Linux in `clone`? Or we can just return `EINVAL` for `clone` as well.

Follow the Linux behavior. However, we don't need to follow the implementation level of Linux behavior. Following the semantics level should suffice. For example, ignore the exit signal if it is invalid for `clone`. We can ignore it in the `clone()` system call instead of when the process exits.

This PR should have regression tests to confirm both `clone` and `clone3` behavior. Use the regression test framework to add one.

Marsman1996:
> > So my question is: shall we make Asterinas compate to the Linux in `clone`? Or we can just return `EINVAL` for `clone` as well.
> 
> Follow the Linux behavior. However, we don't need to follow the implementation level of Linux behavior. Following the semantics level should suffice. For example, ignore the exit signal if it is invalid for `clone`. We can ignore it in the `clone()` system call instead of when the process exits.
> 
> This PR should have regression tests to confirm both `clone` and `clone3` behavior. Use the regression test framework to add one.

Thank you for the quick reply! I will update the PR.

kernel/src/syscall/clone.rs
```diff
@@ -99,8 +99,15 @@ impl TryFrom<Clone3Args> for CloneArgs {
...
-            (value.exit_signal != 0).then(|| SigNum::from_u8(value.exit_signal as u8));
+        if (value.exit_signal & !CSIGNAL_MASK) != 0 {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "the exit signal must be specified in the lower 8 bits of the exit_signal field"
+            )
+        }
+        let exit_signal = (value.exit_signal != 0)
+            .then(|| SigNum::try_from(value.exit_signal as u8))
+            .transpose()?;
```
lrh2000: Actually the logic here is more naturally (you mention the exact same thing in the error message):
```rust
    let exit_signal = u8::try_from(value.exit_signal)
        .ok()
        .and_then(|exit_signal| SigNum::try_from(exit_signal).ok())
        .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid exit signal"));
```

I don't see much point in exporting the CSIGNAL_MASK constant?

It makes more sense to me to have a mask inside `CloneArgs::for_clone` because it's the mask for the clone flags. However, `exit_signal` is not a flag here, so why perform bit operations?

test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
```diff
@@ -0,0 +1,44 @@
...
+			 WEXITSTATUS(status) == EXIT_SUCCESS);
+}
+END_TEST()
+
+FN_TEST(clone3_rejects_invalid_exit_signal)
+{
+	struct clone_args args = {
+		.exit_signal = 255,
+	};
+	TEST_ERRNO(sys_clone3(&args), EINVAL);
```
lrh2000, line 40: We can make the test more symmetric:
```c
	struct clone_args args = {};

	args.exit_signal = 255;
	TEST_ERRNO(sys_clone3(&args), EINVAL);

	args.exit_signal = 257;
	TEST_ERRNO(sys_clone3(&args), EINVAL);
```

test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
```diff
@@ -0,0 +1,44 @@
...
+
+static pid_t sys_clone3(struct clone_args *args)
+{
+	return syscall(SYS_clone3, args, sizeof(struct clone_args));
+}
+
+FN_TEST(clone_normalizes_invalid_exit_signal)
+{
+	int status = 0;
+	long ret = syscall(SYS_clone, 255UL, 0UL, 0UL, 0UL, 0UL);
```
lrh2000: This variable should be renamed to `pid`. Let's use `TEST_SUCC` for the system call.
```c
	int status;
	long pid;

	pid = TEST_SUCC(syscall(SYS_clone, 255UL, 0UL, 0UL, 0UL, 0UL));
	if (pid == 0) {
```

So there is no need to have the redundant `TEST_RES`.

lrh2000 (commented):
Thanks for fixing the bug.

kernel/src/syscall/clone.rs
Marsman1996: Well, I just follow the Linux code before, and it is updated now

kernel/src/syscall/clone.rs
```diff
@@ -99,8 +99,16 @@ impl TryFrom<Clone3Args> for CloneArgs {
...
-            (value.exit_signal != 0).then(|| SigNum::from_u8(value.exit_signal as u8));
+        let exit_signal = if value.exit_signal == 0 {
+            None
+        } else {
+            Some(
+                u8::try_from(value.exit_signal)
+                    .ok()
+                    .and_then(|exit_signal| SigNum::try_from(exit_signal).ok())
+                    .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid exit signal"))?,
+            )
```
lrh2000:
```suggestion
            let exit_signal = u8::try_from(value.exit_signal)
                .ok()
                .and_then(|exit_signal| SigNum::try_from(exit_signal).ok())
                .ok_or_else(|| Error::with_message(Errno::EINVAL, "invalid exit signal"))?;
            Some(exit_signal)
```
It should be more elegant than having these big brackets for `Some`...

test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
```diff
@@ -0,0 +1,42 @@
...
+	int status = 0;
+	long pid;
+
+	pid = TEST_SUCC(syscall(SYS_clone, 255UL, 0UL, 0UL, 0UL, 0UL));
+
+	if (pid == 0) {
+		_exit(EXIT_SUCCESS);
+	}
+
+	TEST_SUCC(waitpid(pid, &status, 0));
```
lrh2000: Sorry, maybe I wasn't clear in my previous comment. By "redundant TEST_RES," I mean `TEST_RES(pid, _ret > 0)` is redundant. Using `TEST_RES` for `waitpid` to check the return code is fine, and it is certainly not redundant.

test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
```diff
@@ -0,0 +1,42 @@
...
+}
+END_TEST()
+
+FN_TEST(clone3_rejects_invalid_exit_signal)
+{
+	struct clone_args args = {};
+	args.exit_signal = 0xff;
```
lrh2000, line 39: Insert a blank line before line 36. `args` will be used in two code blocks below so it should be in a separate code block.

test/initramfs/src/regression/process/clone3/clone_invalid_exit_signal.c
```diff
@@ -0,0 +1,42 @@
...
+#include "../../common/test.h"
+
+static pid_t sys_clone3(struct clone_args *args)
+{
+	return syscall(SYS_clone3, args, sizeof(struct clone_args));
+}
+
+FN_TEST(clone_normalizes_invalid_exit_signal)
+{
+	int status = 0;
```
lrh2000:
```suggestion
	int status;
```

Should be written by the kernel.

test/initramfs/src/regression/process/run_test.sh
```diff
@@ -11,6 +11,7 @@ set -e
 ./clone3/clone_no_exit_signal
 ./clone3/clone_parent
 ./clone3/clone_process
+./clone3/clone_invalid_exit_signal
```
lrh2000: Move it above and keep the list sorted in the alphabetic order.

lrh2000 (approved):

PR #3125, opened 2026-04-13
Add `/proc/[tid]`

### LINKED ISSUE #2940: Add support for accessing thread-specific information via /proc/<tid>
### Current Feature

Currently, in Asterinas, the `/proc/<pid>` directory is supported, allowing access to process-specific information. However, `/proc/<tid>` for accessing thread-specific information is not supported. This is in contrast to the behavior seen in Linux, where `/proc/[tid]` can be used to open and interact with thread-specific data. Additionally, the `pidfd_send_signal` system call in Linux allows sending signals to a specific thread by using `/proc/<tid>` to open the corresponding `pidfd`. In Asterinas, this functionality is not yet supported.

### Enhancement Description

The goal of this enhancement is to support accessing thread-specific information via `/proc/<tid>`, similar to how processes are accessed via `/proc/<pid>`. This would allow Asterinas to fully support operations like `pidfd_send_signal`, enabling signals to be sent to individual threads by opening `/proc/[tid]` and interacting with the corresponding `pidfd`. The implementation should ensure that thread directories are properly exposed and accessible in the `/proc` filesystem, allowing for more granular control over thread management.

### Motivation

Currently, Asterinas lacks support for opening `/proc/<tid>`, which is required for performing thread-specific operations, such as sending signals to threads via `pidfd_send_signal`. This limitation reduces the flexibility of interacting with threads at a lower level. By implementing this feature, Asterinas would align more closely with Linux's behavior, improving compatibility and providing users and developers with more control over thread management.

### Possible Implementations

* Implement the ability to open `/proc/[tid]` for individual threads, similar to how `/proc/[pid]` works for processes.
* Support for accessing `pidfd` for threads via `/proc/<tid>`, allowing `pidfd_send_signal` to send signals directly to specific threads.

### Additional Context
https://github.com/asterinas/asterinas/pull/2912#issuecomment-3815556322

vvvvsv:
Required for #2323. Fix https://github.com/asterinas/asterinas/issues/2940.

This PR adds `/proc/[tid]`s. They can be looked up from `/proc`, but not included in `readdir(/proc)`.

> /proc/tid subdirectories
>               Each one of these subdirectories contains files and
>               subdirectories exposing information about the thread with
>               the corresponding thread ID.  The contents of these
>               directories are the same as the corresponding
>               /proc/pid/task/tid directories.
>               The /proc/tid subdirectories are not visible when iterating
>               through /proc with [getdents(2)](https://man7.org/linux/man-pages/man2/getdents.2.html) (and thus are not visible
>               when one uses [ls(1)](https://man7.org/linux/man-pages/man1/ls.1.html) to view the contents of /proc).
> 

Reference: https://man7.org/linux/man-pages/man5/proc.5.html

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/mod.rs
@@ -7,9 +7,17 @@ use ostd::sync::RwMutexUpgradeableGuard;
 use template::{DirOps, ProcDir, lookup_child_from_table, populate_children_from_table};
 
 use self::{
-    cmdline::CmdLineFileOps, cpuinfo::CpuInfoFileOps, loadavg::LoadAvgFileOps,
-    meminfo::MemInfoFileOps, mounts::MountsSymOps, pid::PidDirOps, self_::SelfSymOps,
-    sys::SysDirOps, thread_self::ThreadSelfSymOps, uptime::UptimeFileOps, version::VersionFileOps,
+    cmdline::CmdLineFileOps,
+    cpuinfo::CpuInfoFileOps,
+    loadavg::LoadAvgFileOps,
+    meminfo::MemInfoFileOps,
+    mounts::MountsSymOps,
+    pid::{PidDirOps, TidDirOps},
+    self_::SelfSymOps,
+    sys::SysDirOps,
+    thread_self::ThreadSelfSymOps,
+    uptime::UptimeFileOps,
+    version::VersionFileOps,
 };
 use crate::{
     events::Observer,
@@ -28,6 +36,7 @@ use crate::{
     process::{
         Pid,
         pid_table::{self, PidEvent},
+        posix_thread::AsPosixThread,
     },
 };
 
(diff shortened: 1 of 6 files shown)


lrh2000:
Fix https://github.com/asterinas/asterinas/issues/2940.

In addition, I haven't thought about any code yet. However, please confirm a few quick questions first:
 1. Please confirm that this PR works according to the following: https://github.com/asterinas/asterinas/pull/2912#issuecomment-3815556322.
 2. Please confirm that this PR will not cause resource leakage and/or make dead processes/threads accessible. We have notifications to release process/thread resources in some places. See `grep -R "observer" kernel/src/fs/fs_impls/procfs`.
 3. Please confirm that this PR works when `execve()` is run in a non-main thread, which makes the non-main thread the main thread.

vvvvsv:
> * Please confirm that this PR works according to the following: [Add `pidfd_send_signal` syscall #2912 (comment)](https://github.com/asterinas/asterinas/pull/2912#issuecomment-3815556322).

Yes.

> * Please confirm that this PR will not cause resource leakage and/or make dead processes/threads accessible. We have notifications to release process/thread resources in some places. See `grep -R "observer" kernel/src/fs/fs_impls/procfs`.
> * Please confirm that this PR works when `execve()` is run in a non-main thread, which makes the non-main thread the main thread.

The implementation in this PR does not place the `ProcDir<TidDirOps>` representing `/proc/[tid]` into any cache, including the dentry cache and `ProcDir<RootDirOps>::cached_children`. Therefore, `/proc/[tid]`s are naturally invisible in `readdir` and does not require the `Subject`/`Observer` mechanism. Each lookup simply queries the thread from `pid_table` again. I think this tradeoff is acceptable, given that most applications are unlikely to access `/proc/[tid]` directly.

lrh2000:
> The implementation in this PR does not place the `ProcDir<TidDirOps>` representing `/proc/[tid]` into any cache, including the dentry cache and `ProcDir<RootDirOps>::cached_children`. 

Thanks for the clarification. I originally thought there were some difficulties. The approach in this PR seems clean.

lrh2000:
> The contents of these
> directories are the same as the corresponding
> /proc/pid/task/tid directories.

But `/proc/[tid]/task` exists. I don't know why.

Anyway, we can defer that to future work... I don't think it's a blocking problem.

---

Add some tests? For example, spawn a test and check:
 - `system("ls /proc/{tid}")`,
 - `system("ls /proc -al | grep ' {tid}$'")`, and
 - `system("ls /proc -al | grep ' {pid}$'")`.

cqs21 (approved):

lrh2000 (approved):

Issue #3126, opened 2026-04-13
Introduce DRM subsystem and migrate graphics stack from fbdev


Linermao:
# Introduce DRM subsystem and migrate graphics stack from fbdev

## Feature Description

Asterinas currently exposes graphics output through a basic fbdev-compatible interface (`/dev/fb0`). Userspace can `mmap` a single linear framebuffer and write pixels into it directly. This is sufficient for early boot graphics and simple framebuffer-based applications, but it is not the programming model expected by modern Linux graphics stacks.

The current fbdev path is fundamentally a fixed linear-scanout model. It provides direct access to framebuffer memory, but it does not provide a DRM/KMS-style device model with display objects, buffer objects, or atomic display control.

This model lacks:

- a DRM device interface such as `/dev/dri/card*`;
- kernel mode-setting abstractions such as connectors, encoders, CRTCs, and planes;
- userspace-visible buffer management primitives such as GEM-style objects;
- display update mechanisms such as page flip, vblank synchronization, and event delivery;
- modern permission and ownership semantics such as DRM master/device-scoped display control.

As a result, the current fbdev-based path cannot serve as a proper foundation for Xorg's modesetting path or for modern compositors, which expect DRM/KMS semantics rather than a single memory-mapped legacy framebuffer.

To support real Linux graphics userspace, Asterinas needs a DRM subsystem that can expose display resources, create mappable scanout buffers, and let userspace control presentation through DRM/KMS APIs instead of writing directly to `/dev/fb0`.

To successfully bring up Xorg (modesetting driver), Asterinas must provide a minimal but coherent DRM contract:

- a DRM device node (`/dev/dri/card*`);
- dumb buffer allocation and mmap capability;
- basic KMS support (CRTC + connector + mode setting);
- a framebuffer creation and scanout path;
- a page flip or equivalent presentation mechanism.

This issue focuses on reaching that minimal contract before extending toward full DRM feature parity.

## Goal

Introduce a DRM/KMS-based graphics path in Asterinas so that Xorg can run on a DRM device instead of depending on the legacy fbdev interface.

- Asterinas exposes a DRM device node to userspace;
- Xorg can start on top of the DRM/KMS path without requiring `/dev/fb0`;
- userspace can create and map scanout-capable buffers through DRM;
- display output is presented through DRM/KMS and remains stable during basic interactive use.

### Phase 0: DRM mmap semantics
  
Before any buffer-object or dumb-buffer support can work correctly,  
[...]

PR #3127, opened 2026-04-14
Add Minimal DRM core and Device Model

Linermao:
Phase 1 for #3126.

## Summary

  This PR introduces the initial DRM infrastructure in Asterinas and wires up the first userspace-facing DRM file/ioctl path.

  **On the component side**, it adds a new `kernel/comps/drm` crate that defines the core `DrmDevice` abstraction and a global DRM device registry.
  It also registers a `simpledrm` device when a framebuffer-backed display is available.

  **On the kernel side**, it adds initial DRM minor support under `kernel/src/device/drm`,
  including primary/render node registration, per-open `DrmFile` state, DRM master tracking,
  and generic ioctl permission/feature gating.

  ## What This PR Adds

  - Introduces the `aster_drm` component crate and hooks it into the build.
  - Defines `DrmDevice` as the common DRM device model.
  - Adds DRM device registration and lookup infrastructure.
  - Implements an initial `simpledrm` device backed by the existing framebuffer path.
  - Registers DRM device nodes under `/dev/dri/`.
  - Adds support for primary nodes (`cardX`) and render nodes (`renderDX`).
  - Implements initial `DrmFile` handling and DRM master ownership semantics.
  - Adds generic DRM ioctl dispatch and permission checks based on minor type and driver features.
  - Implements a basic ioctl set:
    - `DRM_IOCTL_VERSION`
    - `DRM_IOCTL_GET_CAP`
    - `DRM_IOCTL_SET_CLIENT_CAP`
    - `DRM_IOCTL_SET_MASTER`
    - `DRM_IOCTL_DROP_MASTER`
  - Adds `aster_drm::DrmError` to kernel error conversion.
  - Adds DRM initramfs regression tests under `test/initramfs/src/regression/device/drm`.

**On the testing side**, it adds initramfs regression coverage for the initial DRM userspace ABI, including device-node registration, basic file operations, ioctl query paths, client capability state, DRM master ownership, and render-node permission behavior where applicable.

> tips: simpledrm depends on the bootloader-provided framebuffer, so run DRM tests with a framebuffer-capable boot protocol: `BOOT_PROTOCOL=linux-efi-handover64`.

## Reference:

1. [https://docs.kernel.org/gpu/index.html](https://docs.kernel.org/gpu/index.html)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,6 +131,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "aster-drm"
+version = "0.1.0"
+dependencies = [
+ "aster-framebuffer",
+ "aster-time",
+ "bitflags 1.3.2",
+ "component",
+ "hashbrown 0.14.5",
+ "int-to-c-enum",
+ "log",
+ "ostd",
+ "ostd-pod",
+ "spin",
+ "zerocopy",
+]
+
+[[package]]
 name = "aster-framebuffer"
 version = "0.1.0"
 dependencies = [
@@ -186,6 +203,7 @@ dependencies = [
  "aster-block",
  "aster-cmdline",
  "aster-console",
+ "aster-drm",
  "aster-framebuffer",
  "aster-fuse",
  "aster-i8042",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,6 +206,7 @@ dependencies = [
  "component",
  "ostd",
  "ostd-pod",
+ "sparse-id-alloc",
  "spin",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,6 +147,7 @@ dependencies = [
  "aster-nvme",
  "aster-rights",
  "aster-rights-proc",
+ "aster-simpledrm",
  "aster-softirq",
  "aster-systree",
  "aster-time",
@@ -205,6 +206,7 @@ dependencies = [
  "component",
  "ostd",
  "ostd-pod",
+ "spin",
 ]
 
 [[package]]
@@ -345,6 +347,17 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.118",
+]
+
+[[package]]
+name = "aster-simpledrm"
+version = "0.1.0"
+dependencies = [
+ "aster-drm",
+ "aster-framebuffer",
+ "component",
+ "ostd",
+ "spin",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,6 +137,7 @@ dependencies = [
  "aster-block",
  "aster-cmdline",
  "aster-console",
+ "aster-drm",
  "aster-framebuffer",
  "aster-fuse",
  "aster-i8042",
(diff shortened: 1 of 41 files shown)


StevenJiang1110:
@Linermao, could you please rebase this PR to the latest main branch? 

According to the design in #3601, the kernel is currently moving toward a component-based architecture. The DRM subsystem could potentially serve as our first genuinely large-scale component. Note that this is actually the reverse dependency direction from what's implemented in the current PR: per the #3601 design, the DRM component would depend on `aster-core` (i.e., which is still called `aster-kernel` in this PR), whereas in the current implementation, `aster-kernel` depends on the DRM component.

We also currently have a PR (e.g., #3758) that's attempting to migrate components that can be modularized out of `aster-core` into layers above it. However, since this is a newly introduced system, I think we could design it as a standalone component right from the start—this would be much easier than extracting it later.

TBH, I'm not entirely sure whether fully componentizing DRM is feasible at this stage, since `aster-core` hasn't been completely decomposed yet, but perhaps we could give it a try? If it doesn't work out, we can fall back to the current approach and just place DRM under `kernel/core/comps`.

Linermao:
@StevenJiang1110  Thanks for the suggestion. I tried moving the DRM framework directly into kernel/comps, and I agree that making DRM a high-level component from the beginning would probably be the better direction.

One issue I ran into is that aster-core does not yet expose enough of the device/VFS interfaces for a DRM component to live above it. For example, the current DRM design has DrmMinor implementing Device, while each open() creates a per-open DrmFile implementing FileOps / PerOpenFileOps. These interfaces, together with parts of the device registration path, are currently pub(crate) inside aster-core.

So we first need to define an appropriate public extension surface from aster-core for high-level device components. DRM exposes this difference very early: starting from DrmMinor and DrmFile, the implementation under the new architecture is already structurally different from the current PR.

Personally, I would prefer to try the high-level component approach directly rather than implement DRM against the current core-private interfaces and extract it later. It also seems like a useful case for validating what APIs aster-core needs to export for larger components under the design in #3601.

Since this may involve a fairly substantial restructuring compared with the current PR, I am considering putting it in a new draft PR and keeping #3127 as a reference for the existing implementation.

StevenJiang1110:
> Personally, I would prefer to try the high-level component approach directly rather than implement DRM against the current core-private interfaces and extract it later.

Yes, that would be the best approach. After all, we're pushing for componentization in this release cycle, and I think DRM could serve as a great example/case study for this effort.

> One issue I ran into is that aster-core does not yet expose enough of the device/VFS interfaces for a DRM component to live above it. For example, the current DRM design has DrmMinor implementing Device, while each open() creates a per-open DrmFile implementing FileOps / PerOpenFileOps. These interfaces, together with parts of the device registration path, are currently pub(crate) inside aster-core.

This shouldn't be an issue. If DRM needs these interfaces, you can simply make them `pub` in `aster-core`. The `pub(crate)` visibility was always meant to be temporary, since there weren't previously any callers above `aster-core`. Feel free to make this change directly in the current PR.

> Since this may involve a fairly substantial restructuring compared with the current PR, I am considering putting it in a new draft PR and keeping https://github.com/asterinas/asterinas/pull/3127 as a reference for the existing implementation.

You can either open a new PR or continue working directly in this one—it's up to you.

PR #3129, opened 2026-04-14
  Add IPv6 support to the network stack  

yydawx:
Add initial IPv6 support across the network stack, covering the bigtcp                                                                                                                                                                                                                                                                                                      
  protocol layer, address handling utilities, and socket implementation.                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                              
  - Network interfaces: extend bigtcp to handle both IPv4 and IPv6                                                                                                                                                                                                                                                                                                            
    packets, add per-address port tracking, and support dual-stack socket                                                                                                                                                                                                                                                                                                     
    table lookups                                                                                                                                                                                                                                                                                                                                                             
  - Address handling: add IpAddressFamily::Ipv6, Ipv6Address wrapper, and                                                                                                                                                                                                                                                                                                     
    AF_INET6 recognition at the syscall entry                                                                                                                                                                                                                                                                                                                                 
  - Socket layer: wire up TCP and UDP sockets for IPv6, handle ephemeral                                                                                                                                                                                                                                                                                                      
    endpoint binding, return EADDRNOTAVAIL when no interface is available                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                              
  Known limitations (tracked as FIXMEs):                                                                                                                                                                                                                                                                                                                                      
  - EtherIface drops outbound IPv6 packets — neighbor discovery not yet                                                                                                                                                                                                                                                                                                       
    implemented                                                                                                                                                                                                                                                                                                                                                               
  - IPv4-mapped IPv6 addresses (::ffff:x.x.x.x) are normalized to IPv4;                                                                                                                                                                                                                                                                                                       
    proper dual-stack and IPV6_V6ONLY are not yet supported


yydawx:
I've noticed some lint issues, and I will make the necessary fixes.

kernel/libs/aster-bigtcp/src/iface/phy/ip.rs
```diff
@@ -62,7 +77,20 @@ impl<D: WithDevice + 'static, E: Ext> Iface<E> for IpIface<D, E> {
         self.driver.with(|device| {
             let next_poll = self.common.poll(
                 device,
-                |data, _iface_cx, tx_token| Some((Ipv4Packet::new_checked(data).ok()?, tx_token)),
+                |data, _iface_cx, tx_token| {
+                    // 尝试解析为IPv4数据包
+                    if let Ok(ipv4_pkt) = Ipv4Packet::new_checked(data) {
+                        return Some((ipv4_pkt, tx_token));
+                    }
+                    // 尝试解析为IPv6数据包
```
StevenJiang1110: Please avoid any Chinese comments. All comments should be written in English

kernel/libs/aster-bigtcp/src/socket_table.rs
```diff
@@ -31,28 +30,16 @@ pub(crate) struct ListenerKey {
...
-    pub(crate) const fn new(addr: IpAddress, port: PortNum) -> Self {
-        // FIXME: If the socket is listening on an unspecified address (0.0.0.0),
-        // Linux will get the hash value by port only.
+    pub(crate) fn new(addr: IpAddress, port: PortNum) -> Self {
         let hash = hash_addr_port(addr, port);
         Self { addr, port, hash }
     }
 
-    pub(crate) const fn hash(&self) -> SocketHash {
+    pub(crate) fn hash(&self) -> SocketHash {
```
StevenJiang1110: Please avoid including unrelated changes in this PR. For example, removing the `const` qualifier from the `hash` method is obviously not an improvement. Please double-check the PR for any other similar unintended changes.

test/initramfs/src/regression/network/tcp6_client.c
```diff
@@ -0,0 +1,79 @@
+// SPDX-License-Identifier: MPL-2.0
+
```
StevenJiang1110: We now have a testing framework for writing regression tests, which provides macros such as `FN_TEST`, `FN_SETUP`, etc. You can refer to `tcp_err.c` as an example. You may have been following some legacy regression test patterns, but all new tests should use this framework — so this test will need to be updated accordingly.

If you are using an agent to write the test, you might want to try the SKILL mentioned in this comment, which is specifically designed for writing tests:
https://github.com/asterinas/asterinas/pull/3109#issuecomment-4229416673

In addition, please consolidate the client and server into a single C program — there is no need to have two separate files. You can run the client and server in the child and parent processes respectively, eliminating the need for any artificial delays. This way, we get the results as soon as the program finishes. Having separate client and server files is a legacy pattern that we no longer follow.

StevenJiang1110 (commented):
Thanks for your contribution to Asterinas! 

I've done a quick scan of the PR and found a few issues that need to be addressed first.

test/initramfs/src/regression/network/tcp6_client.c
yydawx: I have received all your comments and greatly appreciate your time.I will revise and resubmit my work in accordance with your comments and relevant specifications.
Thank you again for your understanding and patience. I am still not very experienced with collaboration on such large and complex open-source projects.

StevenJiang1110: No worries — please go ahead with the changes. The project is quite complex, so it's natural for newcomers to run into issues. I'll switch this PR to draft status, and you can mark it as ready once the fixes are in place and all CI checks pass.

yydawx:
Regarding the const issue: using const in that fn imposes many restrictions and makes function modifications inflexible. We can keep const if needed, but I will need to adjust the code further to work around these limitations.

test/initramfs/src/regression/network/tcp6_client.c
yydawx: check the new comment please

yydawx:
Already rebase to main

StevenJiang1110:
Thanks for your contribution. I will take a deeper look when I find some time (perhaps this afternoon or tomorrow).

> Regarding the const issue: using const in that fn imposes many restrictions and makes function modifications inflexible. We can keep const if needed, but I will need to adjust the code further to work around these limitations.

The general rule is to avoid introducing unrelated changes. If you have made some modifications to the internal logic of the function that necessitate removing `const`, that is perfectly acceptable. However, if no such changes were made, please refrain from modifying the function signature. If you plan to make such changes in the future, that removal should be done in the future too. Therefore, I do not think these signature changes belong in this PR.

Also, is there any progress on supporting the gVisor tests? This PR seems to only add a single hand-written test. The gVisor tests for this area should be TCP-only, so we should be able to enable at least a subset of them — has that been looked into?

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -289,13 +314,27 @@ impl<E: Ext> BoundPort<E> {
...
+            IpAddress::Ipv4(_) => {
+                let ipv4_addr = self.iface().ipv4_addr()?;
+                IpAddress::Ipv4(ipv4_addr)
+            }
+            IpAddress::Ipv6(_) => {
+                let ipv6_addr = self
+                    .iface()
+                    .ipv6_addr()
+                    .unwrap_or_else(|| Ipv6Address::new(0, 0, 0, 0, 0, 0, 0, 0));
+                IpAddress::Ipv6(ipv6_addr)
```
StevenJiang1110, line 334: The logic here is very strange. Since you already stores addr in `BoundPort`, why query the iface for address?

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -303,16 +331,12 @@ impl<E: Ext> PollContext<'_, E> {
         }
 
         if self.is_unicast_local(ip_repr.src_addr()) {
-            // In this case, the generating ICMP message will have a local IP address as the
-            // destination. However, since we don't have the ability to handle ICMP messages, we'll
-            // just skip the generation.
-            //
-            // TODO: Generate the ICMP message here once we're able to handle incoming ICMP
-            // messages.
```
StevenJiang1110, line 311: Why removing the comments here? This PR has nothing related?

kernel/libs/aster-bigtcp/src/socket_table.rs
```diff
@@ -31,9 +30,7 @@ pub(crate) struct ListenerKey {
 }
 
 impl ListenerKey {
-    pub(crate) const fn new(addr: IpAddress, port: PortNum) -> Self {
-        // FIXME: If the socket is listening on an unspecified address (0.0.0.0),
-        // Linux will get the hash value by port only.
```
StevenJiang1110, line 36: Similarly, why the comments are removed....

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -60,30 +60,34 @@ impl<E: Ext> PollContext<'_, E> {
     pub(super) fn poll_ingress<D, P, Q>(
         &mut self,
         device: &mut D,
-        process_phy: &mut P,
+        _process_phy: &mut P,
```
StevenJiang1110, line 63: If the parameter is unused, remove it.

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -60,30 +60,34 @@ impl<E: Ext> PollContext<'_, E> {
...
-                    return;
-                };
-
-                let Some(reply) = self.parse_and_process_ipv4(pkt) else {
-                    return;
-                };
-
-                dispatch_phy(&reply, self.iface.context_mut(), tx_token);
+                // Try to parse as IPv4 packet
+                if let Ok(ipv4_pkt) = Ipv4Packet::new_checked(data) {
```
StevenJiang1110, line 78: The code here is incorrect. As indicated by the `new_checked` documentation, it only validates the packet length. Therefore, there is a non-trivial chance that an IPv6 packet could pass this check. I think the packet version should be explicitly verified in `process_phy` instead.

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -60,30 +60,34 @@ impl<E: Ext> PollContext<'_, E> {
...
-                let Some((pkt, tx_token)) = process_phy(data, self.iface.context_mut(), tx_token)
-                else {
-                    return;
-                };
-
-                let Some(reply) = self.parse_and_process_ipv4(pkt) else {
-                    return;
-                };
-
-                dispatch_phy(&reply, self.iface.context_mut(), tx_token);
```
StevenJiang1110, line 86: Typically, I hope the code should look like:

```rust
       let Some((pkt, tx_token)) = process_phy(data, self.iface.context_mut(), tx_token) else {
            return;
        };

        match packet {
              IpPacket::IPv4(ipv4_pkt) => {
                     let Some(reply) = self.parse_and_process_ipv4(ipv4_pkt) else {
                        return;
                    };
                    dispatch_phy(&reply, self.iface.context_mut(), tx_token);
              }
              IpPacket::IPv4(ipv6_pkt) => {
                     //.....
              }
        }
```

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -180,23 +200,27 @@ impl<E: Ext> IfaceCommon<E> {
...
+        {
             port_state.nsocket -= 1;
             if can_reuse {
                 port_state.nreuse -= 1;
             }
             if port_state.nsocket == 0 {
-                entry.remove_entry();
+                used_ports
+                    .get_mut(&addr)
+                    .and_then(|family_ports| family_ports.remove(&port));
```
StevenJiang1110, line 223: This is an inefficient pattern. On line 213, you perform a map lookup via `get_mut`, and then on line 223, there is yet another lookup. Compare this with the original implementation, which only requires a single lookup.

kernel/libs/aster-bigtcp/src/iface/phy/ip.rs
```diff
@@ -32,6 +32,18 @@ impl<D: WithDevice, E: Ext> IpIface<D, E> {
...
+    }
+
+    pub fn new_with_ipv6(
+        driver: D,
+        ip_cidr: Ipv4Cidr,
+        ipv6_cidr: Option<Ipv6Cidr>,
+        name: String,
+        sched_poll: E::ScheduleNextPoll,
+        type_: InterfaceType,
+        flags: InterfaceFlags,
```
StevenJiang1110, line 46: Merge this method to `new`. You can directly add `ipv6_cidr: Option<Ipv6Cidr>` to the `new` method.

kernel/src/net/socket/ip/stream/connecting.rs
```diff
@@ -34,26 +34,22 @@ impl ConnectingStream {
         let tcp_conn =
             match TcpConnection::new_connect(bound_port, remote_endpoint, option, observer) {
                 Ok(tcp_conn) => tcp_conn,
-                Err((bound_port, ConnectError::AddressInUse)) => {
+                Err((bp, ConnectError::AddressInUse)) => {
```
StevenJiang1110, line 37: No need to rename `bound_port` to `bp`.

kernel/src/net/socket/ip/stream/connecting.rs
```diff
@@ -105,7 +101,13 @@ impl ConnectingStream {
...
 
     pub(super) fn check_io_events(&self) -> IoEvents {
-        IoEvents::empty()
+        use aster_bigtcp::socket::ConnectState;
+
+        match self.tcp_conn.connect_state() {
+            ConnectState::Connected => IoEvents::OUT,
+            ConnectState::Connecting => IoEvents::empty(),
+            ConnectState::Refused => IoEvents::empty(),
+        }
```
StevenJiang1110, line 110: Are you sure the changes are intended? It seems unrelated to IPv6 support. 

It also looks incorrect to me. You can view the init stream `check_io_events` implementation. If a connection is refused, we should return `OUT | HUP | ERR`.

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -237,27 +237,34 @@ impl StreamSocket {
...
+                    (
+                        State::Connecting(connecting_stream),
+                        Some(iface_to_poll),
+                        None,
+                    )
                 }
                 Err((err, init_stream)) => {
-                    return (State::Init(init_stream), (Some(Err(err)), None));
+                    // Some connection errors should be reported immediately
+                    if err.error() == Errno::EADDRNOTAVAIL || err.error() == Errno::ECONNREFUSED {
```
StevenJiang1110, line 256: All the changes in this file should **not** be part of this PR. If you would like to improve the error handling, please open a separate PR for that purpose. This is because the logic involved covers various corner cases, each of which must be validated by dedicated tests.

kernel/src/syscall/socket.rs
```diff
@@ -85,6 +85,24 @@ pub fn sys_socket(domain: i32, type_: i32, protocol: i32, ctx: &Context) -> Resu
...
         }
+        (CSocketAddrFamily::AF_INET6, SockType::SOCK_STREAM) => {
+            let protocol = Protocol::try_from(protocol)?;
+            match protocol {
+                Protocol::IPPROTO_IP | Protocol::IPPROTO_TCP => {
+                    StreamSocket::new(is_nonblocking) as Arc<dyn FileLike>
+                }
+                _ => return_errno_with_message!(Errno::EAFNOSUPPORT, "unsupported protocol"),
+            }
+        }
```
StevenJiang1110, line 96: This branch can be merged into above SOCK_STREAM branch?

```rust
(CSocketAddrFamily::AF_INET | CSocketAddrFamily::AF_INET6, SockType::SOCK_STREAM)
```

test/initramfs/src/regression/network/run_test.sh
```diff
@@ -4,14 +4,25 @@
...
 ./udp_server &
 sleep 0.2
 ./udp_client
 
+# Test IPv6 UDP (skipped: not completed)
+if [ "${SKIP_IPV6_UDP}" != "1" ]; then
+    ./udp6_server &
+    sleep 0.2
+    ./udp6_client
+fi
```
StevenJiang1110, line 24: This should be dead code? Since there are no udp client and udp server.

test/initramfs/src/regression/network/tcp6.c
```diff
@@ -0,0 +1,103 @@
...
+
+	// Fork a child process for the client
+	pid = TEST_SUCC(fork());
+
+	if (pid == 0) {
+		// Child process (client)
+		close(server_fd);
+
+		// Create client socket
+		client_fd = TEST_SUCC(socket(AF_INET6, SOCK_STREAM, 0));
```
StevenJiang1110, line 59: In the child process, you should use `CHECK`/`CHECK_WITH` instead of `TEST_SUCC`/`TEST_RES`. This is because any state updated by `TEST_SUCC` in the child process will not be reflected in the parent process. Additionally, in the parent process, you should wait on the child process's exit code to confirm that it terminated successfully.

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -39,7 +45,7 @@ pub struct IfaceCommon<E: Ext> {
     flags: InterfaceFlags,
 
     interface: SpinLock<PollableIface<E>, BottomHalfDisabled>,
-    used_ports: SpinLock<BTreeMap<u16, PortState>, BottomHalfDisabled>,
+    used_ports: SpinLock<BTreeMap<IpAddress, BTreeMap<u16, PortState>>, BottomHalfDisabled>,
```
StevenJiang1110, line 48: Based on this structure, it appears that each IPv4 and IPv6 address has its own corresponding port. However, with this data structure, how would you handle the mapping between IPv4-mapped IPv6 addresses? For example, ::ffff:192.0.2.1 and 192.0.2.1 should be treated as the same address. If a socket is bound to one of them, the other should not be bindable as well (ignoring SO_REUSEADDR). The table should therefore be capable of detecting and handling this kind of behavior.

StevenJiang1110 (commented):
Thanks for your contribution again! This is a preliminary review, and there are still quite a few issues with the current implementation — please refer to my specific comments for details.

One major concern is that this PR, in addition to implementing basic IPv6 support, also mixes in a number of unrelated changes, such as fixes to error handling. These should be moved into separate PRs, each with dedicated tests to validate the corresponding behavior. This PR should remain focused on IPv6 support. Furthermore, it appears that some changes unrelated to this PR have been introduced as well — I have left comments on some of them, but you may need to do another pass and double-check.

yydawx:
Hi! @StevenJiang1110
I've improved this PR. Please have a look when you're available.

I've fixed the **IPv4 mapped IPv6** issue, details are as follows:

# IPv4-Mapped IPv6 Address Handling
## Problem
`::ffff:192.0.2.1` and `192.0.2.1` should be treated as the same address for binding.
Binding to one should block the other.

## Solution
Added a simple `normalize_ip_address` function, called in `bind_port` and `release_port`:

```rust
fn normalize_ip_address(addr: IpAddress) -> IpAddress {
    match addr {
        IpAddress::Ipv6(ipv6) => ipv6.to_ipv4_mapped()
            .map(IpAddress::Ipv4)
            .unwrap_or(addr),
        other => other,
    }
}
```

## Performance & Alternatives
- **Overhead**: Negligible - just a match check, only executed on bind/release (not hot path)
- **Custom key type**: More type-safe but adds boilerplate
- **Separate tables**: More complex, need to check both for mapped addresses

We chose simplicity. The overhead is trivial, and we can refactor later if needed.

## TODO
- Full dual-stack support
- Binding to `::`
- `IPV6_V6ONLY` option, etc.


btw， is this normal?
image

yydawx:
Hi！ @StevenJiang1110 . This pr is ready for review. I've solved previous probloms. Regarding to the mapping issue, I gave a possible way up, please check when you're available. Thanks for your time!

StevenJiang1110:
> Hi！ @StevenJiang1110 . This pr is ready for review. I've solved previous probloms. Regarding to the mapping issue, I gave a possible way up, please check when you're available. Thanks for your time!

Thanks for your efforts! I'll try to find some time to review it tomorrow. Could you also mark all the comments you have addressed as "Resolved"? This will help keep the discussion focused and avoid drawing unnecessary attention to already-solved issues.

> btw， is this normal?

This is a known issue tracked in #3181. Someone is already working on it, so you don't need to worry about it.

yydawx:
Copy that! I will order our conversition ASAP




青海白。
***@***.***

kernel/libs/aster-bigtcp/src/iface/phy/ip.rs
yydawx: No any Chinese comments now

kernel/libs/aster-bigtcp/src/iface/common.rs
yydawx: I've gave a simple but possible way to solve this problem below.

test/initramfs/src/regression/network/tcp6.c
yydawx: Shall I keep the regression test? Or I just open gVisor tests.

btw, I opened gVisor tests and block the same tests as ipv4 did.

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -15,9 +15,40 @@ use ostd::sync::{SpinLock, SpinLockGuard};
...
+        IpAddress::Ipv6(ipv6) => {
+            // Check if this is an IPv4-mapped IPv6 address
+            if let Some(ipv4) = ipv6.to_ipv4_mapped() {
+                IpAddress::Ipv4(ipv4)
+            } else {
+                IpAddress::Ipv6(ipv6)
+            }
+        }
+        other => other,
+    }
```
StevenJiang1110, line 43: This funtion can be simplified with the following form:

```rust
fn normalize_ip_address(addr: &IpAddress) -> IpAddress {
    if let IpAddress::Ipv6(ipv6_addr) = addr && let Some(normalized_addr) = ipv6_addr.to_ipv4_mapped() {
        return IpAddress::Ipv4(normalized_addr);
    }
    
    *addr
}
```

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -154,21 +193,24 @@ impl<E: Ext> IfaceCommon<E> {
...
         } else {
-            match Self::alloc_ephemeral_port(&mut used_ports, config_can_reuse) {
+            match Self::alloc_ephemeral_port(&mut used_ports, addr, config_can_reuse) {
                 Some(port) => port,
                 None => return Err(BindError::Exhausted),
             }
         };
 
-        if let Some(port_state) = used_ports.get_mut(&port) {
+        let family_ports = used_ports.entry(addr).or_default();
```
StevenJiang1110: `family_ports` seems like a slightly odd name. Since this variable represents all ports associated with a given address, `address_ports` might be a more fitting name.

kernel/libs/aster-bigtcp/src/iface/common.rs
StevenJiang1110: I think this problem is not resolved in the latest version.....

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -289,13 +338,14 @@ impl<E: Ext> BoundPort<E> {
         self.port
     }
 
+    /// Returns the bound IP address.
+    pub fn addr(&self) -> IpAddress {
+        self.addr
+    }
```
StevenJiang1110, line 337: This method should return `&IpAddress`? Let the caller decide whether to clone the address.

kernel/libs/aster-bigtcp/src/iface/port.rs
```diff
@@ -1,35 +1,49 @@
...
-            (_, false) => Self::Specified(port),
+            (0, can_reuse) => Self::Ephemeral(endpoint.addr, can_reuse),
+            (_, true) => Self::CanReuse(endpoint.addr, port),
+            (_, false) => Self::Specified(endpoint.addr, port),
         }
     }
 
     pub(super) fn can_reuse(&self) -> bool {
-        matches!(self, Self::CanReuse(_)) || matches!(self, Self::Ephemeral(true))
+        matches!(self, Self::CanReuse(_, _)) || matches!(self, Self::Ephemeral(_, true))
```
StevenJiang1110:
```suggestion
        matches!(self, Self::CanReuse(..)) || matches!(self, Self::Ephemeral(_, true))
```
When an enum variant has multiple fields, using `..` is more concise than listing individual wildcards like `_, _`.

kernel/src/syscall/socket.rs
```diff
@@ -35,17 +35,23 @@ pub fn sys_socket(domain: i32, type_: i32, protocol: i32, ctx: &Context) -> Resu
...
+                        _ => unreachable!(),
+                    };
+                    StreamSocket::new(is_nonblocking, family) as Arc<dyn FileLike>
                 }
                 _ => return_errno_with_message!(Errno::EAFNOSUPPORT, "unsupported protocol"),
             }
         }
-        (CSocketAddrFamily::AF_INET, SockType::SOCK_DGRAM) => {
+        (CSocketAddrFamily::AF_INET, SockType::SOCK_DGRAM)
+        | (CSocketAddrFamily::AF_INET6, SockType::SOCK_DGRAM) => {
```
StevenJiang1110: Let's remove line 54 for now since UDP ipv6 support is not ready now.

kernel/src/net/socket/ip/stream/init.rs
```diff
@@ -22,6 +25,8 @@ use crate::{
 
 pub(super) struct InitStream {
     bound_port: Option<BoundPort>,
+    /// The address family of this socket (IPv4 or IPv6).
+    family: AddressFamily,
```
StevenJiang1110: Have you validated the address family when binding or connecting to an address? For example, if a socket is created with `socket(AF_INET, ...)`, and the caller later tries to bind it to an IPv6 address, should we return an error?

I looked into the Linux behavior on this. Based on my research: if a socket is created with `AF_INET`, it only accepts `struct sockaddr_in` (i.e., IPv4 addresses); if created with `AF_INET6`, it only accepts `struct sockaddr_in6`. The `IPV6_V6ONLY` socket option controls whether the socket can handle IPv4-mapped IPv6 addresses. This is admittedly unrelated to the current PR — feel free to correct me if my understanding is wrong.

From this perspective, I think we should validate the address family in bind and connect. In Rust, a cleaner approach would be to encode this check in the type system. One way to do this is to introduce an `AddressFamily` trait:

```rust
pub trait AddressFamily {
    type Addr;
    const UNSPECIFIED_ADDR: Self::Addr;
}

pub struct Ipv4Family;

impl AddressFamily for Ipv4Family {
    type Addr = Ipv4Addr;
    const UNSPECIFIED_ADDR: Ipv4Addr = Ipv4Addr::new(0, 0, 0, 0);
}

pub struct Ipv6Family;

impl AddressFamily for Ipv6Family {
    type Addr = Ipv6Addr;
    const UNSPECIFIED_ADDR: Ipv6Addr = Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0);
}

```

We can then use this as a type parameter for BoundPort and InitStream:


```rust
pub struct BoundPort<Family: AddressFamily> {
    iface: Arc<dyn Iface>,
    addr: Family::Addr,
    // ...
}

pub struct InitStream<Family: AddressFamily> {
    bound_port: Option<BoundPort<Family>>,
    // ...
}

```

This way, the type system statically guarantees that the address type is always correct, catching mismatches at compile time rather than at runtime.

However, this change could have a wide blast radius. If the refactoring turns out to be too invasive, it would be acceptable to simply add a runtime check in bind and connect instead, without encoding it in the type system.

test/initramfs/src/regression/network/tcp6.c
```diff
@@ -0,0 +1,103 @@
...
+
+		// Receive response from server
+		CHECK_WITH(read(client_fd, buffer, 1024),
+			   _ret == (ssize_t)strlen(response));
+		CHECK_WITH(strcmp(buffer, response), _ret == 0);
+
+		// Close client socket
+		CHECK(close(client_fd));
+		_exit(0);
+	} else {
```
StevenJiang1110: The else branch is redundant here. Since the if block returns early, you can move the code out of the else branch to reduce unnecessary indentation

test/initramfs/src/regression/network/tcp6.c
StevenJiang1110: This test appears to be fairly basic, and since the relevant gVisor tests have already been enabled, their coverage should subsume this regression test. I think we can safely remove it.

kernel/libs/aster-bigtcp/src/iface/common.rs
StevenJiang1110: The solution looks good.

StevenJiang1110 (commented):
Thanks for your revision! The code looks much better now.

kernel/libs/aster-bigtcp/src/iface/common.rs
yydawx: now it's resolved

kernel/libs/aster-bigtcp/src/iface/common.rs
yydawx: resolved

kernel/libs/aster-bigtcp/src/iface/port.rs
yydawx: resoleved

kernel/src/net/socket/ip/stream/init.rs
yydawx: I take the solution to add a runtime check and add TODO comments

test/initramfs/src/regression/network/tcp6.c
yydawx: deleted this test file

yydawx:
Hi！Happy labour day. I've improved this PR as we discussed.

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -289,13 +331,14 @@ impl<E: Ext> BoundPort<E> {
...
+
     /// Returns the bound endpoint.
     pub fn endpoint(&self) -> Option<IpEndpoint> {
-        let ip_addr = {
-            let ipv4_addr = self.iface().ipv4_addr()?;
-            IpAddress::Ipv4(ipv4_addr)
-        };
-        Some(IpEndpoint::new(ip_addr, self.port))
+        Some(IpEndpoint::new(self.addr, self.port))
```
StevenJiang1110: This method doesn't need to return `Option<IpEndpoint>` now, return `IpEndpoint` should be enough.

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -15,9 +15,36 @@ use ostd::sync::{SpinLock, SpinLockGuard};
...
+/// IPv4-mapped IPv6 addresses (::ffff:x.x.x.x) should be treated as equivalent
+/// to their IPv4 counterparts for binding purposes. This ensures that binding
+/// to 192.0.2.1:80 and ::ffff:192.0.2.1:80 are treated as the same.
+///
+/// TODO: This function currently only handles port binding conflict detection.
+/// Full dual-stack support is not yet implemented, including:
+/// - Accepting IPv4 connections on IPv6 wildcard socket (binding to `::`)
+/// - Returning IPv4-mapped addresses in accept() for IPv4 clients
+/// - Proper handling of `IPV6_V6ONLY` socket option
+fn normalize_ip_address(addr: IpAddress) -> IpAddress {
```
StevenJiang1110, line 38: The TODO should not be placed in doc comments (i.e., ///), but rather in a regular comment (i.e., //).


```suggestion
/// Normalizes IPv4-mapped IPv6 addresses to IPv4 addresses.
///
/// IPv4-mapped IPv6 addresses (::ffff:x.x.x.x) should be treated as equivalent
/// to their IPv4 counterparts for binding purposes. This ensures that binding
/// to 192.0.2.1:80 and ::ffff:192.0.2.1:80 are treated as the same.
//
// TODO: This function currently only handles port binding conflict detection.
// Full dual-stack support is not yet implemented, including:
// - Accepting IPv4 connections on IPv6 wildcard socket (binding to `::`)
// - Returning IPv4-mapped addresses in accept() for IPv4 clients
// - Proper handling of `IPV6_V6ONLY` socket option
fn normalize_ip_address(addr: IpAddress) -> IpAddress {
```

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -154,21 +189,24 @@ impl<E: Ext> IfaceCommon<E> {
     fn bind_port(&self, config: BindPortConfig) -> Result<(u16, bool), BindError> {
         let mut used_ports = self.used_ports.lock();
         let config_can_reuse = config.can_reuse();
+        // Normalize IPv4-mapped IPv6 addresses to IPv4 for consistent key lookup
```
StevenJiang1110: This comment is useless, the method name already explains it. Another comment with the same content should be removed too.

kernel/libs/aster-bigtcp/src/iface/port.rs
```diff
@@ -1,35 +1,49 @@
...
+    CanReuse(IpAddress, u16),
     /// Binds to the specified reusable port.
-    Specified(u16),
+    Specified(IpAddress, u16),
     /// Allocates an ephemeral port to bind.
-    Ephemeral(bool),
+    Ephemeral(IpAddress, bool),
     /// Reuses the port of the listening socket.
-    Backlog(u16),
+    Backlog(IpAddress, u16),
```
StevenJiang1110: Duplicating `IpAddress` in every variant looks like not a good idea. Considering the following form.

```rust
pub struct BindPortConfig {
    addr: IpAddress,
    kind: PortKind,
}

pub enum PortKind {
    CanReuse(u16),
    Specified(u16),
    Ephemeral(bool),
    Backlog(u16),
}

impl BindPortConfig {
    pub fn new(endpoint: IpEndpoint, can_reuse: bool) -> Self { ... }
    pub fn new_backlog(endpoint: IpEndpoint) -> Self { ... }
    pub(super) fn addr(&self) -> IpAddress { self.addr }
    pub(super) fn port(&self) -> Option<u16> { ... }
    pub(super) fn can_reuse(&self) -> bool { ... }
}
```

kernel/libs/aster-bigtcp/src/iface/phy/ether.rs
```diff
@@ -213,6 +213,10 @@ impl<D, E: Ext> EtherIface<D, E> {
         // Resolve the next-hop IP address.
         let next_hop_ip = match iface_cx.route(&pkt.ip_repr().dst_addr(), iface_cx.now()) {
             Some(IpAddress::Ipv4(next_hop_ip)) => next_hop_ip,
+            Some(IpAddress::Ipv6(_)) => {
+                // For IPv6, we don't have neighbor discovery yet
```
StevenJiang1110: This should be a FIXME instead a normal comment:

```suggestion
                // FIXME: Drop outbound IPv6 packets until neighbor discovery is implemented;   
                // we have no way to resolve the next-hop link-layer address.
                ostd::warn!("neighbor discovery is not implemented for EtherIface");
```

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -115,6 +125,30 @@ impl<E: Ext> PollContext<'_, E> {
...
 
+    fn parse_and_process_ipv6<'pkt>(
+        &mut self,
+        pkt: smoltcp::wire::Ipv6Packet<&'pkt [u8]>,
+    ) -> Option<Packet<'pkt>> {
+        // Parse the IPv6 header. Ignore the packet if the header is ill-formed.
+        let repr = smoltcp::wire::Ipv6Repr::parse(&pkt).ok()?;
+
+        if !self.is_unicast_local(IpAddress::Ipv6(repr.dst_addr)) {
+            // Temporarily not handling IPv6 ICMP unreachable messages
```
StevenJiang1110: This also should be a FIXME

kernel/libs/aster-bigtcp/src/socket_table.rs
```diff
@@ -103,23 +103,48 @@ const fn hash_local_remote(
...
+                local_bits as u32,
+                (remote_bits >> 96) as u32,
+                (remote_bits >> 64) as u32,
+                (remote_bits >> 32) as u32,
+                remote_bits as u32,
+                (local_port as u32).wrapping_shl(16) | remote_port as u32,
+            ];
+            jhash_u32_array(hash_keys, HASH_SECRET.wrapping_add(NET_HASHMIX))
+        }
+        _ => panic!("Cannot mix IPv4 and IPv6 addresses"),
```
StevenJiang1110:
```suggestion
        _ => panic!("cannot mix IPv4 and IPv6 addresses"),
```

Panic messages should start with a lowercase letter.

kernel/src/net/iface/broadcast.rs
```diff
@@ -32,6 +32,8 @@ pub(super) fn init() {
 
 /// Determines if a given IP endpoint's address is a known broadcast address.
 pub fn is_broadcast_endpoint(endpoint: &IpEndpoint) -> bool {
-    let IpAddress::Ipv4(ipv4_addr) = &endpoint.addr;
+    let IpAddress::Ipv4(ipv4_addr) = &endpoint.addr else {
+        return false;
```
StevenJiang1110, line 39: Add some document about IPv6 does not have broadcast address.

```rust
/// IPv6 has no broadcast; multicast (`ff00::/8`) handles fan-out instead and
/// is intentionally not covered by this function.
```

kernel/src/net/socket/ip/common.rs
```diff
@@ -15,39 +15,74 @@ use crate::{
...
+fn get_ephemeral_iface(remote_ip_addr: &IpAddress) -> Option<Arc<Iface>> {
+    match remote_ip_addr {
+        IpAddress::Ipv4(remote_ipv4_addr) => {
+            if let Some(iface) = iter_all_ifaces().find(|iface| {
+                if let Some(iface_ipv4_addr) = iface.ipv4_addr() {
+                    iface_ipv4_addr == *remote_ipv4_addr
+                } else {
+                    false
+                }
+            }) {
```
StevenJiang1110, line 38: Let us simplify the condition in the find closure, similar for the IPv6 case:

```suggestion
            if let Some(iface) = iter_all_ifaces().find(|iface| {
                iface.ipv4_addr().is_some_and(|iface_ipv4_addr|
                    iface_ipv4_addr == *remote_ipv4_addr
                )
             })
      
```

kernel/src/net/socket/ip/common.rs
```diff
@@ -15,39 +15,74 @@ use crate::{
...
 /// If the remote address is the same as that of some iface, we will use the iface.
 /// Otherwise, we will use a default interface.
-fn get_ephemeral_iface(remote_ip_addr: &IpAddress) -> Arc<Iface> {
-    let IpAddress::Ipv4(remote_ipv4_addr) = remote_ip_addr;
-    if let Some(iface) = iter_all_ifaces().find(|iface| {
-        if let Some(iface_ipv4_addr) = iface.ipv4_addr() {
-            iface_ipv4_addr == *remote_ipv4_addr
-        } else {
-            false
+fn get_ephemeral_iface(remote_ip_addr: &IpAddress) -> Option<Arc<Iface>> {
```
StevenJiang1110: Why changing this method to return `Option<Arc<Iface>>`? This method will never return `None`, right?

Similar for `get_empheral_endpoint`, which should not return `Option`.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -35,3 +36,33 @@ impl From<IpEndpoint> for SocketAddr {
...
+/// IPv6 unspecified local endpoint.
+pub(super) const UNSPECIFIED_LOCAL_ENDPOINT_V6: IpEndpoint =
+    IpEndpoint::new(IpAddress::Ipv6(Ipv6Address::UNSPECIFIED), 0);
+
+/// Address family for IP sockets.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum AddressFamily {
+    IPv4,
+    IPv6,
+}
```
StevenJiang1110, line 49: Consider renaming this enum to `IpAddressFamily`. Linux's address family encompasses more than just IPv4 and IPv6, so a more precise name is warranted here.

test/initramfs/src/conformance/gvisor/blocklists/tcp_socket_test
```diff
@@ -29,3 +28,34 @@ AllInetTests/TcpSocketTest.TcpInq/0
 AllInetTests/TcpSocketTest.TcpSCMPriority/0
 AllInetTests/TcpSocketTest.Tiocinq/0
 AllInetTests/TcpSocketTest.ZeroWriteAllowed/0
+
+AllInetTests/SimpleTcpSocketTest.CleanupOnConnectionRefused/1
```
StevenJiang1110: The IPv6 block list entries are simply duplicates of their IPv4 counterparts. We can simplify this by using a wildcard, since gVisor supports wildcard patterns.

Replace the following two lines:

```
AllInetTests/SimpleTcpSocketTest.CleanupOnConnectionRefused/0
AllInetTests/SimpleTcpSocketTest.CleanupOnConnectionRefused/1
```
with one lines

```
AllInetTests/SimpleTcpSocketTest.CleanupOnConnectionRefused/*
```

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -68,22 +68,32 @@ impl<E: Ext> PollContext<'_, E> {
...
+                        };
+                        dispatch_phy(&reply, self.iface.context_mut(), tx_token);
+                    }
+                    IpPacket::Ipv6(ipv6_pkt) => {
+                        let Some(reply) = self.parse_and_process_ipv6(ipv6_pkt) else {
+                            return;
+                        };
+                        dispatch_phy(&reply, self.iface.context_mut(), tx_token);
+                    }
+                }
```
StevenJiang1110: This two branch has same patterns, so we can simplify it:

```suggestion
                let reply = match ip_packet {
                    IpPacket::Ipv4(p) => self.parse_and_process_ipv4(p),
                    IpPacket::Ipv6(p) => self.parse_and_process_ipv6(p),
                };
                let Some(reply) = reply else { return };
                dispatch_phy(&reply, self.iface.context_mut(), tx_token);
```

kernel/src/net/socket/ip/addr.rs
```diff
@@ -35,3 +36,33 @@ impl From<IpEndpoint> for SocketAddr {
...
+
+impl From<IpAddress> for AddressFamily {
+    fn from(addr: IpAddress) -> Self {
+        match addr {
+            IpAddress::Ipv4(_) => AddressFamily::IPv4,
+            IpAddress::Ipv6(_) => AddressFamily::IPv6,
+        }
+    }
+}
```
StevenJiang1110, line 71: Please add a short doc comment here explaining that this function does not handle IPv6 addresses that can be mapped to IPv4 addresses. This is because when `IPV6_V6ONLY` is set, IPv4-mapped IPv6 addresses are not permitted, and therefore this function cannot be used to determine whether such an address is acceptable.

kernel/src/net/socket/ip/stream/init.rs
StevenJiang1110: The runtime check is acceptable for now. I'm not yet sure that encoding the address family as a type parameter is the right approach without more experimentation.

StevenJiang1110 (commented):
Thanks for your revisions! I only have a few minor comments remaining, most of which should be straightforward to address.

Could you also update the PR title and description to reflect the current state of the changes? This will help other reviewers and readers follow along more easily.

Additionally, you may want to consider using the [pr-review-skill](https://github.com/tatetian/asterinas/blob/pr-review-skill/.agents/skills/pr-review/spec.md) to review the changes locally and fix the problems it found. It may occasionally produce false positives, but it can also catch issues that human reviewers might easily overlook!

Thanks for your cooperation!

kernel/libs/aster-bigtcp/src/iface/phy/ether.rs
yydawx: we should reconsider ostd::warn
Per-packet warn floods the log under sustained IPv6 traffic. One-shot gating adds complexity for a known limitation. Removed — the FIXME is sufficient.

kernel/src/net/socket/ip/common.rs
```diff
@@ -84,6 +105,18 @@ impl From<BindError> for Error {
...
+        IpAddress::Ipv4(_) => {
+            let ip_addr = iface
+                .ipv4_addr()
+                .expect("selected iface should have an IPv4 address");
+            IpEndpoint::new(IpAddress::Ipv4(ip_addr), 0)
+        }
+        IpAddress::Ipv6(_) => {
+            let ipv6_addr = iface
+                .ipv6_addr()
+                .expect("selected iface should have an IPv6 address");
```
StevenJiang1110: This panic is actually reachable. Since we currently require each interface to have an IPv4 address but not necessarily an IPv6 address, the kernel can be panicked if no interface has an IPv6 address configured.

Let's revert to the previous approach: have `get_ephemeral_endpoint` return `Option<IpEndpoint>` instead, so that this panic can be avoided.

kernel/libs/aster-bigtcp/src/iface/phy/ether.rs
StevenJiang1110: Consider switching to a lower log level, such as `ostd::debug!`, so that this message is only visible when debug-level logging is enabled.

StevenJiang1110 (commented):
LGTM! Thanks for your work. These should be my final comments.

Apart from the PR contents itself, please also tidy up the commit messages to align with this project's conventions: each commit message should begin with a capitalized verb, and prefixes such as `feat(bigtcp):` should be avoided. Any commits that address review feedback should be amended into their corresponding earlier commits.

> Could you also update the PR title and description to reflect the current state of the changes? This will help other reviewers and readers follow along more easily.

As I mentioned earlier, please update the PR description to reflect the current scope of the changes. The scope has shifted considerably from the original description — for example, the tests mentioned initially are no longer included — so the description should be revised accordingly.

kernel/libs/aster-bigtcp/src/iface/phy/ether.rs
```diff
@@ -213,6 +213,11 @@ impl<D, E: Ext> EtherIface<D, E> {
         // Resolve the next-hop IP address.
         let next_hop_ip = match iface_cx.route(&pkt.ip_repr().dst_addr(), iface_cx.now()) {
             Some(IpAddress::Ipv4(next_hop_ip)) => next_hop_ip,
+            Some(IpAddress::Ipv6(_)) => {
+                // FIXME: Drop outbound IPv6 packets until neighbor discovery is implemented;
+                // we have no way to resolve the next-hop link-layer address.
```
lrh2000:
```suggestion
                // FIXME: Currently, we drop outbound IPv6 packets because neighbor discovery is not
                // implemented and we have no way to resolve the next-hop link-layer address.
```

kernel/libs/aster-bigtcp/src/iface/phy/ip.rs
```diff
@@ -25,9 +25,15 @@ pub struct IpIface<D, E: Ext> {
 }
 
 impl<D: WithDevice, E: Ext> IpIface<D, E> {
+    // NOTE: `ip_cidr` is required; `ipv6_cidr` is optional.
+    // This means IPv6-only interfaces cannot be created with this constructor.
+    // TODO: Refactor to accept `Option<Ipv4Cidr>` or use an enum like `IpCidrConfig`
+    // to support IPv6-only interfaces (e.g. sit/gre tunnels, WireGuard).
+    // All callers must be updated when this changes.
```
lrh2000: > All callers must be updated when this changes.

This comment is useless.

> Refactor to accept `Option<Ipv4Cidr>` or use an enum like `IpCidrConfig`

I don't think we should propose a concrete solution in a TODO/FIXME comment. Let's say the problem.

```suggestion
    // TODO: Support interfaces with multiple IPv4/IPv6 addresses or without IPv4 addresses.
```

kernel/libs/aster-bigtcp/src/iface/poll.rs
```diff
@@ -115,6 +117,30 @@ impl<E: Ext> PollContext<'_, E> {
...
 
+    fn parse_and_process_ipv6<'pkt>(
+        &mut self,
+        pkt: smoltcp::wire::Ipv6Packet<&'pkt [u8]>,
+    ) -> Option<Packet<'pkt>> {
+        // Parse the IPv6 header. Ignore the packet if the header is ill-formed.
+        let repr = smoltcp::wire::Ipv6Repr::parse(&pkt).ok()?;
+
+        if !self.is_unicast_local(IpAddress::Ipv6(repr.dst_addr)) {
+            // FIXME: Temporarily not handling IPv6 ICMP unreachable messages
```
lrh2000:
```suggestion
            // TODO: Generate an IPv6 ICMP unreachable message.

File: kernel/libs/aster-bigtcp/src/iface/poll.rs [hunk 44]
```diff
@@ -115,6 +117,30 @@ impl<E: Ext> PollContext<'_, E> {
         }
     }
 
+    fn parse_and_process_ipv6<'pkt>(
+        &mut self,
+        pkt: smoltcp::wire::Ipv6Packet<&'pkt [u8]>,
+    ) -> Option<Packet<'pkt>> {
+        // Parse the IPv6 header. Ignore the packet if the header is ill-formed.
+        let repr = smoltcp::wire::Ipv6Repr::parse(&pkt).ok()?;
```
lrh2000:
```suggestion
        pkt: Ipv6Packet<&'pkt [u8]>,
    ) -> Option<Packet<'pkt>> {
        // Parse the IPv6 header. Ignore the packet if the header is ill-formed.
        let repr = Ipv6Repr::parse(&pkt).ok()?;
```

This is more consistent with `parse_and_process_ipv4`.

kernel/libs/aster-bigtcp/src/iface/phy/ip.rs
```diff
@@ -62,7 +71,22 @@ impl<D: WithDevice + 'static, E: Ext> Iface<E> for IpIface<D, E> {
...
+                    if data.is_empty() {
+                        return None;
+                    }
+                    let version = data[0] >> 4;
+
+                    if version == 4 {
+                        let pkt = Ipv4Packet::new_checked(data).ok()?;
+                        Some((IpPacket::Ipv4(pkt), tx_token))
+                    } else if version == 6 {
+                        let pkt = wire::Ipv6Packet::new_checked(data).ok()?;
```
lrh2000: To be consistent with the `version == 4` branch:
```suggestion
                        let pkt = Ipv6Packet::new_checked(data).ok()?;
```

kernel/src/net/socket/ip/addr.rs
```diff
@@ -35,3 +36,36 @@ impl From<IpEndpoint> for SocketAddr {
 /// unspecified endpoint helps with that.
 pub(super) const UNSPECIFIED_LOCAL_ENDPOINT: IpEndpoint =
     IpEndpoint::new(IpAddress::Ipv4(Ipv4Address::UNSPECIFIED), 0);
+
+/// IPv6 unspecified local endpoint.
```
lrh2000: Line 32:
```patch
-/// A local endpoint, which indicates that the local endpoint is unspecified.
+/// An IPv4 local endpoint, which indicates that the local endpoint is unspecified.
```

This line:
```suggestion
/// An IPv6 local endpoint, which indicates that the local endpoint is unspecified.
```

kernel/src/net/socket/ip/stream/init.rs
```diff
@@ -45,35 +52,52 @@ pub(super) struct InitStream {
...
     pub(super) fn bind(&mut self, endpoint: &IpEndpoint, can_reuse: bool) -> Result<()> {
         if self.bound_port.is_some() {
             return_errno_with_message!(Errno::EINVAL, "the socket is already bound to an address");
         }
 
+        // NOTE: When implementing IPV6_V6ONLY, this check must be extended:
+        // if IPV6_V6ONLY is set on an AF_INET6 socket, IPv4-mapped addresses
+        // (::ffff:x.x.x.x) should be rejected here. Currently normalize_ip_address()
+        // in IfaceCommon collapses mapped addresses to IPv4 before port binding,
+        // which silently accepts mapped addresses regardless of V6ONLY.
```
lrh2000: > Currently normalize_ip_address() in IfaceCommon collapses mapped addresses to IPv4 before port binding,

`normalize_ip_address` isn't visible here. Let's avoid mentioning another crate's implementation details.

This comment looks too verbose. I think the following is enough:
```rust
        // When we support `IPV6_V6ONLY` and if it is set, we should also reject IPv4-mapped
        // IPv6 addresses.
```

Same for line 123.

kernel/src/net/socket/ip/stream/init.rs
```diff
@@ -45,35 +52,52 @@ pub(super) struct InitStream {
...
+        if IpAddressFamily::from(endpoint.addr) != self.family {
+            return_errno_with_message!(Errno::EAFNOSUPPORT, "address family mismatch");
```
lrh2000:
```suggestion
            return_errno_with_message!(Errno::EAFNOSUPPORT, "the protocol family does not match the address family");
```

Same for line 126.

kernel/src/net/socket/ip/common.rs
```diff
@@ -15,39 +15,60 @@ use crate::{
...
+            if let Some(iface) = iter_all_ifaces().find(|iface| {
+                iface
+                    .ipv6_addr()
+                    .is_some_and(|addr| addr == *remote_ipv6_addr)
+            }) {
+                return iface.clone();
+            }
+
+            // Fallback to an interface with IPv6 address
+            // Prefer virtio over loopback for external traffic
```
lrh2000:
```suggestion
            // Fall back to an interface with an IPv6 address.
            // Prefer virtio over loopback for external traffic.

File: kernel/libs/aster-bigtcp/src/iface/common.rs [hunk 50]
```diff
@@ -15,9 +15,36 @@ use ostd::sync::{SpinLock, SpinLockGuard};
...
+/// Normalizes IPv4-mapped IPv6 addresses to IPv4 addresses.
+///
+/// IPv4-mapped IPv6 addresses (::ffff:x.x.x.x) should be treated as equivalent
+/// to their IPv4 counterparts for binding purposes. This ensures that binding
+/// to 192.0.2.1:80 and ::ffff:192.0.2.1:80 are treated as the same.
+//
+// TODO: This function currently only handles port binding conflict detection.
+// Full dual-stack support is not yet implemented, including:
+// - Accepting IPv4 connections on IPv6 wildcard socket (binding to `::`)
+// - Returning IPv4-mapped addresses in accept() for IPv4 clients
```
lrh2000:
```suggestion
// - Returning IPv4-mapped addresses in `accept()` for IPv4 clients
```

yydawx:
Done! I think it's ready! Thanks for your time. Hope we can work further. What shell I do next?

kernel/libs/aster-bigtcp/src/lib.rs
```diff
@@ -17,6 +17,13 @@
 // See <https://github.com/rust-lang/rust/issues/132952>.
 #![expect(rustdoc::private_intra_doc_links)]
 
+// Enable OSTD logging macros for this crate.
```
lrh2000:
```suggestion
// Set this crate's log prefix for `ostd::log`.
```

Be consistent with:
```
> grep -R "Set this crate" kernel
kernel/comps/block/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/framebuffer/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/input/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/time/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/virtio/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/logger/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/mlsdisk/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/pci/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/cmdline/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/i8042/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/comps/uart/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
kernel/src/lib.rs:// Set this crate's log prefix for `ostd::log`.
```

kernel/libs/aster-bigtcp/src/iface/phy/ether.rs
```diff
@@ -213,6 +213,12 @@ impl<D, E: Ext> EtherIface<D, E> {
...
+                ostd::debug!("neighbor discovery is not implemented for EtherIface");
```
lrh2000:
```suggestion
                ostd::debug!("IPv6 neighbor discovery is not implemented for Ethernet interfaces");
```

kernel/src/net/socket/ip/stream/init.rs
```diff
@@ -22,6 +25,10 @@ use crate::{
 
 pub(super) struct InitStream {
     bound_port: Option<BoundPort>,
+    /// The address family of this socket (IPv4 or IPv6).
+    // TODO: Encode the address family in the type system (e.g., `InitStream<IPv4Family>`)
```
lrh2000, line 30:
```suggestion
    //
    // TODO: Encode the address family in the type system (e.g., `InitStream<IPv4Family>`)
```

kernel/libs/aster-bigtcp/src/iface/phy/ether.rs
```diff
@@ -213,6 +213,12 @@ impl<D, E: Ext> EtherIface<D, E> {
         // Resolve the next-hop IP address.
         let next_hop_ip = match iface_cx.route(&pkt.ip_repr().dst_addr(), iface_cx.now()) {
             Some(IpAddress::Ipv4(next_hop_ip)) => next_hop_ip,
+            Some(IpAddress::Ipv6(_)) => {
+                // FIXME: Drop outbound IPv6 packets until neighbor discovery is implemented;
```
lrh2000: Also note that we do not support incoming IPv6 Packets for Ethernet packets as well...
```rust
        // Ignore the Ethernet frame if the protocol is not supported.
        match repr.ethertype {
            EthernetProtocol::Ipv4 => {
                Ok(Ipv4Packet::new_checked(frame.payload()).map_err(|_| None)?)
            }
            EthernetProtocol::Arp => {
                let pkt = ArpPacket::new_checked(frame.payload()).map_err(|_| None)?;
                let arp = ArpRepr::parse(&pkt).map_err(|_| None)?;
                Err(self.process_arp(&arp, iface_cx))
            }
            _ => Err(None),
```
(in `parse_ip_or_process_arp` above in the same file).

I think it is trivial to implement though?

lrh2000 (commented):
Thank you for your wonderful work adding IPv6 support, and thank @StevenJiang1110 for your review. This PR mostly looks good to me. I left a few minor comments that mostly adjust the wording and style of some comments.

When addressing these comments, please squash the changes to the original comment. Thanks!

yydawx:
Happy to see you! Good evening. All problems you mentioned were resolved. Please check again!

StevenJiang1110 (approved):

> What shell I do next?

We'll need to wait for reviews and approvals from the other reviewers.

lrh2000 (approved):

yydawx:
Hi @StevenJiang1110  Could you let me know what I should take on next as part of you guys planned workflow?

StevenJiang1110:
> Hi @StevenJiang1110 Could you let me know what I should take on next as part of you guys planned workflow?

It really depends on your personal plans and preferences. You can decide what tasks to take on based on your availability and interests. For example, if you have plenty of time, you might want to take on a more time-intensive project; if you are more limited on time, a smaller and more self-contained task might be a better fit. We do not directly assign tasks to contributors who are not working full-time in the Asterinas community, unless you voluntarily step up to take something on.

We are about to cut the 0.18.0 release (in this month), and all of the planned goals for this release already have corresponding PRs, so there are no pending tasks at the moment. The goals for the next 0.19.0 release have not been planned yet, so there are no short-term roadmap items available either.

If you would like to continue contributing, here are a few directions you might consider:

1. Building on top of this PR to further improve IPv6 support — for example, adding UDP support or IPv4-mapped IPv6 address support. The current IPv6 support is still fairly preliminary.
2. Working on other networking features outside of IPv6 — one option would be implementing IP raw socket support, though this is a fairly large feature and would require a significant time commitment.
3. Exploring other non-networking areas — you could look through the open issues for something that interests you, or identify a new task yourself, open an issue for it, and take it on from there.

yydawx:
Ok！ I will contiue working on ipv6-related staff.
btw: who should I discuss with about this PR since it's not been approved by other maintainers so far
Thanks!

StevenJiang1110:
> btw: who should I discuss with about this PR since it's not been approved by other maintainers so far

@tatetian, could you please take a look at this PR? Thanks.

tatetian (approved):

tatetian:
@yydawx Could you rebase this PR to resolve the conflicts? Thanks!

yydawx:
> @yydawx Could you rebase this PR to resolve the conflicts? Thanks!

Ok！Already rebased.

yydawx:
could this pr be merged so far? or we should wait for another reviewer.    :)

tatetian:
> could this pr be merged so far? or we should wait for another reviewer. :)

Thanks! We appreciate your effort!

yydawx:
Gooooooood!!

PR #2936, opened 2026-01-29
Redirect serial output if `hvc0` is on

lrh2000:
Fix #2935

See https://github.com/asterinas/asterinas/issues/2935#issuecomment-3815029847 for the discussion.

PR #3130, opened 2026-04-14
Fix duplicate kernel prints by selecting one console backend

le-monde-bleu:
**Problem**
In `/kernel/comps/logger/src`, there is a FIXME sign in line 5:

> //! FIXME: It will print to all `virtio-console` devices, which is not a good choice.

which indicates that the current set causes some problems. For example, in the RISC-V scheme, banner/logs can appear twice because logger broadcasts to all consoles. By running `make run_kernel OSDK_TARGET_ARCH=riscv64 SCHEME=riscv` to boot in docker, the "Asterinas" banner appears twice (see the image at the bottom).

**Root Cause**
Logger sends each print to every registered console. In the RISC-V scheme, QEMU wires both UART serial and virtio console to the same chardev mux. Consequently, there are two registered consoles.

**Fix**
Route to the configured console backend only, and fallback to first backend when available. Now `kernel/comps/logger/src/console.rs` maps from Linux console name to internal device name, and in `write_str`:
- If no device exists: fallback to early print (same idea as before).
- If devices exist:
- - Try selected device from console=...
- - If not found, fallback to first registered device
- - Send once only

This can prevent duplicated prints when multiple consoles are active. After the fix, the banner appears only once in the RISC-V scheme.

**WARNING**
1. Fallback uses first_key_value from a BTreeMap, so fallback target is deterministic by key order, not by boot priority;
2. If device list is empty, it still uses `early_print` path (as in the x86_64 scheme possibly, where no banner appears);
3. For future extensions, if more TTY consoles are supported in the future, they can be added to the device list in `selected_console_device_name`.

image


lrh2000:
As a quick solution, consider switching to an approach similar to the one described in https://github.com/asterinas/asterinas/pull/2936 (it solves the problem for x86 but forgets about RISC-V, sorry). It should be relatively simple and uncontroversial to merge.

https://github.com/asterinas/asterinas/blob/2a30158c3cd77427217581a39a971667b761576a/OSDK.toml#L62-L64

A long-term fix would require more complexity. As the Copilot comment above points out, the way the console parameter is handled in `aster-logger` isn't correct.  We already handle the parameter in the `aster-kernel`. Therefore, the interaction needs to be studied more carefully.

It isn't a problem if you have real hardware where each console works independently. I suggest fixing the configuration first.

le-monde-bleu:
@lrh2000 Thanks for your revision! I will make a new pull request after.

le-monde-bleu:
As for long-term fix, is it practical to add a single shared source for `console=` parsing under `kernel/comps/cmdline/src` and export it via `kernel/comps/cmdline/src/lib.rs`?

lrh2000:
> As for long-term fix, is it practical to add a single shared source for `console=` parsing under `kernel/comps/cmdline/src` and export it via `kernel/comps/cmdline/src/lib.rs`?

Putting it in `aster-cmdline` seems odd and parsing it in `aster-logger` might not be reasonable either, since you would need to hardcode the console name, such as "Uart-Console". However, we can do better here:
https://github.com/asterinas/asterinas/blob/72e395b5f10d6f87e28b813adb40370c7b54cbb4/kernel/src/device/tty/serial.rs#L75

`aster-kernel` will depend on all of those crates so that it can easily refer to their names. However, it may not be suitable for `aster-logger` to depend on those crates.

I think that, under our current design, `aster-kernel` should look up the consoles (i.e., translating the console names to the console objects) and instruct the `aster-logger` to print to them.

PR #3131, opened 2026-04-14
Add capability LSM module and migrate capability checks

zzj-5341:
## Summary

This PR builds on the LSM framework and boot-time module-selection logic introduced by #3078 and moves kernel capability authorization into the LSM stack.

It introduces a mandatory built-in capability LSM, keeps optional modules such as Yama selected through the existing `lsm=` / legacy `security=` flow, and routes existing capability-sensitive checks through `lsm_hooks::on_capable(...)`.

File capability support is not part of this PR. That work has been deferred to a follow-up PR so this change stays focused on the capability LSM itself and on migrating existing checks to the LSM stack.

### 1. Add a built-in mandatory capability LSM

- Add `kernel/src/security/lsm/modules/capability.rs`
- Add `kernel/src/security/lsm/hooks/capability.rs`
- Register capability in `MANDATORY_MODULES`
- Keep Yama optional and selected by the existing boot-time module-selection rules

### 2. Extend the LSM hook surface for capability checks

- Add `CapableContext`
- Expose capability checks through `lsm_hooks::on_capable(...)`
- Keep hook dispatch consistent with the existing `lsm_hooks::on_alien_access(...)` pattern from #3078
- Let the public dispatch functions consume one-shot hook contexts by value while each active module still borrows the shared context during dispatch

### 3. Route existing capability checks through the LSM stack

Migrate existing capability-sensitive paths, including:

- namespace-related checks (`setns`, mount, UTS, IPC, and cgroup paths)
- signal permission checks and alien-access / ptrace-like checks
- resource limit checks
- reboot authorization
- socket privilege checks (`NET_ADMIN`, `NET_BIND_SERVICE`, and SCM_RIGHTS UNIX socket passing)
- `capset`
- trusted xattr access and xattr namespace selection
- inode DAC override through a direct `CAP_DAC_OVERRIDE` capability check in `Inode::check_permission`
- ext2 reserved-block allocation

### 4. Add regression coverage and capability-test helper cleanup

- Add capability regression tests for `kill`, `reboot`, `setgroups`, and trusted xattr behavior
- Add an LSM module-selection regression test for Yama procfs visibility under the effective boot-time LSM selection
- Add shared capability helpers in `test/initramfs/src/regression/common/capability.h`
- Migrate repeated capget / capset / drop-capability helpers in existing tests to the shared header

### DIFF
--- a/.github/workflows/test_x86.yml
+++ b/.github/workflows/test_x86.yml
@@ -33,7 +33,6 @@ jobs:
         uses: ./.github/actions/test
         with:
           auto_test: 'general'
-          runs_on: 'ubuntu-latest'
 
   integration-test:
     runs-on: ubuntu-latest
--- a/kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
@@ -12,6 +12,7 @@ use crate::{
     },
     prelude::*,
     process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::CapabilityReason,
 };
 
 /// The cgroup namespace for the unified cgroup hierarchy.
@@ -41,7 +42,11 @@ impl CgroupNamespace {
         owner: Arc<UserNamespace>,
         posix_thread: &PosixThread,
     ) -> Result<Arc<Self>> {
-        owner.check_cap(CapSet::SYS_ADMIN, posix_thread)?;
+        owner.check_cap_with_reason(
+            CapSet::SYS_ADMIN,
+            posix_thread,
+            CapabilityReason::Namespace,
+        )?;
 
         let root: Arc<dyn CgroupSysNode> = match current_cgroup {
             Some(current_cgroup) => current_cgroup,
--- a/kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
@@ -10,6 +10,7 @@ use crate::{
     },
     prelude::*,
     process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::lsm::hooks as lsm_hooks,
 };
 
 /// The cgroup namespace for the unified cgroup hierarchy.
@@ -39,7 +40,11 @@ impl CgroupNamespace {
         owner: Arc<UserNamespace>,
         posix_thread: &PosixThread,
     ) -> Result<Arc<Self>> {
-        owner.check_cap(CapSet::SYS_ADMIN, posix_thread)?;
+        lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+            owner.as_ref(),
+            posix_thread,
+            CapSet::SYS_ADMIN,
+        ))?;
 
         let root: Arc<dyn CgroupSysNode> = match current_cgroup {
             Some(current_cgroup) => current_cgroup,
--- a/kernel/src/fs/vfs/fs_apis/inode.rs
+++ b/kernel/src/fs/vfs/fs_apis/inode.rs
@@ -22,7 +22,8 @@ use crate::{
         vfs::path::Path,
     },
     prelude::*,
-    process::{Gid, Uid, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{Gid, Uid, posix_thread::AsPosixThread},
+    security,
     time::clocks::RealTimeCoarseClock,
     vm::vmo::Vmo,
 };
@@ -424,37 +425,17 @@ pub trait Inode: Any + InodeIo + Send + Sync {
     /// Similar to Linux, using "fsuid" here allows setting filesystem permissions
     /// without changing the "normal" uids for other tasks.
     fn check_permission(&self, mut perm: Permission) -> Result<()> {
-        let creds = match Task::current() {
-            Some(task) => match task.as_posix_thread() {
-                Some(thread) => thread.credentials(),
-                None => return Ok(()),
-            },
+        let current = match Task::current() {
+            Some(task) => task,
             None => return Ok(()),
         };
+        let Some(posix_thread) = current.as_posix_thread() else {
+            return Ok(());
+        };
+        let creds = posix_thread.credentials();
 
-        // With DAC_OVERRIDE capability, the user can bypass some permission checks.
-        if creds.effective_capset().contains(CapSet::DAC_OVERRIDE) {
-            // Read/write DACs are always overridable.
-            perm -= Permission::MAY_READ | Permission::MAY_WRITE;
-
-            // Executable DACs are overridable when there is at least one exec bit set.
-            if perm.may_exec() {
-                let metadata = self.metadata();
-                let mode = metadata.mode;
-
-                if mode.is_owner_executable()
-                    || mode.is_group_executable()
-                    || mode.is_other_executable()
-                {
-                    perm -= Permission::MAY_EXEC;
-                } else {
-                    return_errno_with_message!(
-                        Errno::EACCES,
-                        "root execute permission denied: no execute bits set"
-                    );
-                }
-            }
-        }
+        let overridden = security::inode_dac_override(self.metadata().mode, perm, posix_thread)?;
+        perm -= overridden;
 
         perm =
             perm.intersection(Permission::MAY_READ | Permission::MAY_WRITE | Permission::MAY_EXEC);
--- a/kernel/src/fs/vfs/fs_apis/inode.rs
+++ b/kernel/src/fs/vfs/fs_apis/inode.rs
@@ -7,7 +7,7 @@ use core::time::Duration;
 
 use device_id::DeviceId;
 use no_std_io2::io::{Error as IoError, ErrorKind as IoErrorKind, Result as IoResult, Write};
-use ostd::task::Task;
+use ostd::task::{CurrentTask, Task};
 use spin::Once;
 
 use super::{
@@ -22,7 +22,12 @@ use crate::{
         vfs::path::Path,
     },
     prelude::*,
-    process::{Gid, Uid, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{
+        Gid, Uid,
+        credentials::capabilities::CapSet,
+        posix_thread::{AsPosixThread, PosixThread},
+    },
+    security::lsm::hooks as lsm_hooks,
     time::clocks::RealTimeCoarseClock,
     vm::page_cache::PageCache,
 };
@@ -531,24 +536,24 @@ pub trait Inode: Any + FileOps + Send + Sync {
     /// Similar to Linux, using "fsuid" here allows setting filesystem permissions
     /// without changing the "normal" uids for other tasks.
     fn check_permission(&self, mut perm: Permission) -> Result<()> {
-        let creds = match Task::current() {
-            Some(task) => match task.as_posix_thread() {
-                Some(thread) => thread.credentials(),
-                None => return Ok(()),
-            },
-            None => return Ok(()),
+        let Some(task) = Task::current() else {
+            return Ok(());
+        };
+        let Some(posix_thread) = task.as_posix_thread() else {
+            return Ok(());
         };
 
+        let creds = posix_thread.credentials();
+        let metadata = self.metadata();
+        let mode = metadata.mode;
+
         // With DAC_OVERRIDE capability, the user can bypass some permission checks.
-        if creds.effective_capset().contains(CapSet::DAC_OVERRIDE) {
+        if has_dac_override_capability(&task, posix_thread) {
             // Read/write DACs are always overridable.
             perm -= Permission::MAY_READ | Permission::MAY_WRITE;
 
             // Executable DACs are overridable when there is at least one exec bit set.
             if perm.may_exec() {
-                let metadata = self.metadata();
-                let mode = metadata.mode;
-
                 if mode.is_owner_executable()
                     || mode.is_group_executable()
                     || mode.is_other_executable()
@@ -562,11 +567,6 @@ pub trait Inode: Any + FileOps + Send + Sync {
                 }
             }
         }
-
-        perm =
-            perm.intersection(Permission::MAY_READ | Permission::MAY_WRITE | Permission::MAY_EXEC);
-        let metadata = self.metadata();
-        let mode = metadata.mode;
 
         if metadata.uid == creds.fsuid() {
             if (perm.may_read() && !mode.is_owner_readable())
@@ -591,6 +591,17 @@ pub trait Inode: Any + FileOps + Send + Sync {
 
         Ok(())
     }
+}
+
+fn has_dac_override_capability(task: &CurrentTask, posix_thread: &PosixThread) -> bool {
+    let thread_local = task.as_thread_local().unwrap();
+    let user_ns = thread_local.borrow_user_ns();
+    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        user_ns.as_ref(),
+        posix_thread,
+        CapSet::DAC_OVERRIDE,
+    ))
+    .is_ok()
 }
 
 impl dyn Inode {
--- a/kernel/src/net/socket/unix/ctrl_msg.rs
+++ b/kernel/src/net/socket/unix/ctrl_msg.rs
@@ -13,7 +13,8 @@ use crate::{
     },
     net::socket::util::{CControlHeader, ControlMessage},
     prelude::*,
-    process::{credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    security::{self, CapabilityReason},
     util::net::CSocketOptionLevel,
 };
 
@@ -253,13 +254,14 @@ impl AuxiliaryData {
         {
             warn!("UNIX sockets in SCM_RIGHTS messages can leak kernel resource");
 
-            let credentials = current_thread!().as_posix_thread().unwrap().credentials();
-            if !credentials.effective_capset().contains(CapSet::SYS_ADMIN) {
-                return_errno_with_message!(
-                    Errno::EPERM,
-                    "UNIX sockets in SCM_RIGHTS messages can leak kernel resource"
-                )
-            }
+            let current = current_thread!();
+            let posix_thread = current.as_posix_thread().unwrap();
+            security::capable(
+                UserNamespace::get_init_singleton().as_ref(),
+                CapSet::SYS_ADMIN,
+                posix_thread,
+                CapabilityReason::Socket,
+            )?;
         }
 
         Ok(Self { files, cred })
--- a/kernel/src/net/socket/unix/ctrl_msg.rs
+++ b/kernel/src/net/socket/unix/ctrl_msg.rs
@@ -13,7 +13,8 @@ use crate::{
     },
     net::socket::util::{CControlHeader, ControlMessage},
     prelude::*,
-    process::{credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    security::lsm::hooks as lsm_hooks,
     util::net::CSocketOptionLevel,
 };
 
@@ -253,13 +254,13 @@ impl AuxiliaryData {
         {
             warn!("UNIX sockets in SCM_RIGHTS messages can leak kernel resource");
 
-            let credentials = current_thread!().as_posix_thread().unwrap().credentials();
-            if !credentials.effective_capset().contains(CapSet::SYS_ADMIN) {
-                return_errno_with_message!(
-                    Errno::EPERM,
-                    "UNIX sockets in SCM_RIGHTS messages can leak kernel resource"
-                )
-            }
+            let current = current_thread!();
+            let posix_thread = current.as_posix_thread().unwrap();
+            lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+                UserNamespace::get_init_singleton().as_ref(),
+                posix_thread,
+                CapSet::SYS_ADMIN,
+            ))?;
         }
 
         Ok(Self { files, cred })
--- a/kernel/src/net/socket/util/options.rs
+++ b/kernel/src/net/socket/util/options.rs
@@ -18,7 +18,8 @@ use crate::{
         unix::{CUserCred, UNIX_DATAGRAM_DEFAULT_BUF_SIZE, UNIX_STREAM_DEFAULT_BUF_SIZE},
     },
     prelude::*,
-    process::{credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    security::{self, CapabilityReason},
 };
 
 #[derive(Clone, CopyGetters, Debug, Setters)]
@@ -261,17 +262,14 @@ impl SocketOptionSet {
 }
 
 fn check_current_privileged() -> Result<()> {
-    let credentials = {
-        let current = current_thread!();
-        let posix_thread = current.as_posix_thread().unwrap();
-        posix_thread.credentials()
-    };
-
-    if credentials.effective_capset().contains(CapSet::NET_ADMIN) {
-        return Ok(());
-    }
-
-    return_errno_with_message!(Errno::EPERM, "the process does not have permissions")
+    let current = current_thread!();
+    let posix_thread = current.as_posix_thread().unwrap();
+    security::capable(
+        UserNamespace::get_init_singleton().as_ref(),
+        CapSet::NET_ADMIN,
+        posix_thread,
+        CapabilityReason::Socket,
+    )
 }
 
 fn check_priority(priority: i32) -> Result<()> {
--- a/kernel/src/net/socket/util/options.rs
+++ b/kernel/src/net/socket/util/options.rs
@@ -18,7 +18,8 @@ use crate::{
         unix::{CUserCred, UNIX_DATAGRAM_DEFAULT_BUF_SIZE, UNIX_STREAM_DEFAULT_BUF_SIZE},
     },
     prelude::*,
-    process::{credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    security::lsm::hooks as lsm_hooks,
 };
 
 #[derive(Clone, CopyGetters, Debug, Setters)]
@@ -261,17 +262,13 @@ impl SocketOptionSet {
 }
 
 fn check_current_privileged() -> Result<()> {
-    let credentials = {
-        let current = current_thread!();
-        let posix_thread = current.as_posix_thread().unwrap();
-        posix_thread.credentials()
-    };
-
-    if credentials.effective_capset().contains(CapSet::NET_ADMIN) {
-        return Ok(());
-    }
-
-    return_errno_with_message!(Errno::EPERM, "the process does not have permissions")
+    let current = current_thread!();
+    let posix_thread = current.as_posix_thread().unwrap();
+    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        UserNamespace::get_init_singleton().as_ref(),
+        posix_thread,
+        CapSet::NET_ADMIN,
+    ))
 }
 
 fn check_priority(priority: i32) -> Result<()> {
--- a/kernel/src/process/credentials/credentials_.rs
+++ b/kernel/src/process/credentials/credentials_.rs
@@ -9,10 +9,15 @@ use super::{
 };
 use crate::{
     prelude::*,
-    process::credentials::{
-        AMBIENT_CAPSET,
-        capabilities::{AtomicCapSet, CapSet},
+    process::{
+        UserNamespace,
+        credentials::{
+            AMBIENT_CAPSET,
+            capabilities::{AtomicCapSet, CapSet},
+        },
+        posix_thread::AsPosixThread,
     },
+    security::{self, CapabilityReason},
 };
 
 #[derive(Debug)]
@@ -122,7 +127,7 @@ impl Credentials_ {
     }
 
     pub(super) fn set_uid(&self, uid: Uid) -> Result<()> {
-        if self.effective_capset().contains(CapSet::SETUID) {
+        if self.current_has_capability(CapSet::SETUID, CapabilityReason::CredentialsSetUid) {
             self.set_resuid_unchecked(Some(uid), Some(uid), Some(uid));
             Ok(())
         } else {
@@ -164,7 +169,7 @@ impl Credentials_ {
             return Ok(old_fsuid);
         };
 
-        if self.effective_capset().contains(CapSet::SETUID) {
+        if self.current_has_capability(CapSet::SETUID, CapabilityReason::CredentialsSetUid) {
             self.set_fsuid_unchecked(fsuid);
             return Ok(old_fsuid);
         }
@@ -224,7 +229,7 @@ impl Credentials_ {
         suid: Option<&Uid>,
         ruid_may_be_old_suid: bool,
     ) -> Result<()> {
-        if self.effective_capset().contains(CapSet::SETUID) {
+        if self.current_has_capability(CapSet::SETUID, CapabilityReason::CredentialsSetUid) {
             return Ok(());
         }
 
@@ -359,7 +364,7 @@ impl Credentials_ {
     }
 
     pub(super) fn set_gid(&self, gid: Gid) -> Result<()> {
-        if self.effective_capset().contains(CapSet::SETGID) {
+        if self.current_has_capability(CapSet::SETGID, CapabilityReason::CredentialsSetGid) {
             self.set_resgid_unchecked(Some(gid), Some(gid), Some(gid));
             Ok(())
         } else {
@@ -405,7 +410,7 @@ impl Credentials_ {
             return Ok(old_fsgid);
         }
 
-        if self.effective_capset().contains(CapSet::SETGID) {
+        if self.current_has_capability(CapSet::SETGID, CapabilityReason::CredentialsSetGid) {
             self.set_fsgid_unchecked(fsgid);
             return Ok(old_fsgid);
         }
@@ -437,7 +442,7 @@ impl Credentials_ {
         sgid: Option<&Gid>,
         rgid_may_be_old_sgid: bool,
     ) -> Result<()> {
-        if self.effective_capset().contains(CapSet::SETGID) {
+        if self.current_has_capability(CapSet::SETGID, CapabilityReason::CredentialsSetGid) {
             return Ok(());
         }
 
@@ -580,7 +585,7 @@ impl Credentials_ {
     }
 
     pub(super) fn set_securebits(&self, securebits: SecureBits) -> Result<()> {
-        if !self.effective_capset().contains(CapSet::SETPCAP) {
+        if !self.current_has_capability(CapSet::SETPCAP, CapabilityReason::CredentialsSetPcap) {
             return_errno_with_message!(
                 Errno::EPERM,
                 "only threads with CAP_SETPCAP can change secure bits"
@@ -588,6 +593,21 @@ impl Credentials_ {
         }
 
         self.securebits.try_store(securebits, Ordering::Relaxed)
+    }
+
+    fn current_has_capability(&self, capability: CapSet, reason: CapabilityReason) -> bool {
+        let current = current_thread!();
+        let Some(posix_thread) = current.as_posix_thread() else {
+            return self.effective_capset().contains(capability);
+        };
+
+        security::capable(
+            UserNamespace::get_init_singleton().as_ref(),
+            capability,
+            posix_thread,
+            reason,
+        )
+        .is_ok()
     }
 }
 
--- a/kernel/src/process/kill.rs
+++ b/kernel/src/process/kill.rs
@@ -8,6 +8,7 @@ use super::{
 use crate::{
     prelude::*,
     process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::CapabilityReason,
     thread::Tid,
 };
 
@@ -168,7 +169,7 @@ fn check_signal_perm(target: &PosixThread, ctx: &Context, signum: Option<SigNum>
     if target_process
         .user_ns()
         .lock()
-        .check_cap(CapSet::KILL, ctx.posix_thread)
+        .check_cap_with_reason(CapSet::KILL, ctx.posix_thread, CapabilityReason::Signal)
         .is_ok()
     {
         return Ok(());
--- a/kernel/src/process/kill.rs
+++ b/kernel/src/process/kill.rs
@@ -8,6 +8,7 @@ use super::{
 use crate::{
     prelude::*,
     process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::lsm::hooks as lsm_hooks,
     thread::Tid,
 };
 
@@ -174,11 +175,12 @@ fn check_signal_perm(target: &PosixThread, ctx: &Context, signum: Option<SigNum>
         return Ok(());
     }
 
-    if target_process
-        .user_ns()
-        .lock()
-        .check_cap(CapSet::KILL, ctx.posix_thread)
-        .is_ok()
+    if lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        target_process.user_ns().lock().as_ref(),
+        ctx.posix_thread,
+        CapSet::KILL,
+    ))
+    .is_ok()
     {
         return Ok(());
     }
--- a/kernel/src/process/namespace/user_ns.rs
+++ b/kernel/src/process/namespace/user_ns.rs
@@ -6,6 +6,7 @@ use crate::{
     fs::pseudofs::{NsCommonOps, NsType, StashedDentry},
     prelude::*,
     process::{Uid, credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::{self, CapabilityReason},
 };
 
 /// The user namespace.
@@ -27,20 +28,18 @@ impl UserNamespace {
 
     /// Checks whether the thread has the required capability in this user namespace.
     pub fn check_cap(&self, required: CapSet, posix_thread: &PosixThread) -> Result<()> {
-        // Since creating new user namespaces is not supported at the moment,
-        // there is effectively only one user namespace in the entire system.
-        // Therefore, the thread has a single set of capabilities used for permission checks.
-        // FIXME: Once support for creating new user namespaces is added,
-        // we should verify the thread's capabilities within the relevant user namespace.
-        let cap_set = posix_thread.credentials().effective_capset();
-        if cap_set.contains(required) {
-            return Ok(());
-        }
+        self.check_cap_with_reason(required, posix_thread, CapabilityReason::General)
+    }
 
-        return_errno_with_message!(
-            Errno::EPERM,
-            "the thread does not have the required capability"
-        )
+    /// Checks whether the thread has the required capability in this user namespace
+    /// for a specific kernel operation.
+    pub fn check_cap_with_reason(
+        &self,
+        required: CapSet,
+        posix_thread: &PosixThread,
+        reason: CapabilityReason,
+    ) -> Result<()> {
+        security::capable(self, required, posix_thread, reason)
     }
 
     /// Returns the owner UID of the user namespace.
--- a/kernel/src/process/namespace/user_ns.rs
+++ b/kernel/src/process/namespace/user_ns.rs
@@ -26,11 +26,6 @@ impl UserNamespace {
         })
     }
 
-    /// Checks whether the thread has the required capability in this user namespace.
-    pub fn check_cap(&self, required: CapSet, posix_thread: &PosixThread) -> Result<()> {
-        self.check_cap_with_reason(required, posix_thread, CapabilityReason::General)
-    }
-
     /// Checks whether the thread has the required capability in this user namespace
     /// for a specific kernel operation.
     pub fn check_cap_with_reason(
--- a/kernel/src/process/namespace/user_ns.rs
+++ b/kernel/src/process/namespace/user_ns.rs
@@ -5,7 +5,7 @@ use spin::Once;
 use crate::{
     fs::pseudofs::{NsCommonOps, NsType, StashedDentry},
     prelude::*,
-    process::{Uid, credentials::capabilities::CapSet, posix_thread::PosixThread},
+    process::Uid,
 };
 
 /// The user namespace.
@@ -23,24 +23,6 @@ impl UserNamespace {
                 stashed_dentry: StashedDentry::new(),
             })
         })
-    }
-
-    /// Checks whether the thread has the required capability in this user namespace.
-    pub fn check_cap(&self, required: CapSet, posix_thread: &PosixThread) -> Result<()> {
-        // Since creating new user namespaces is not supported at the moment,
-        // there is effectively only one user namespace in the entire system.
-        // Therefore, the thread has a single set of capabilities used for permission checks.
-        // FIXME: Once support for creating new user namespaces is added,
-        // we should verify the thread's capabilities within the relevant user namespace.
-        let cap_set = posix_thread.credentials().effective_capset();
-        if cap_set.contains(required) {
-            return Ok(());
-        }
-
-        return_errno_with_message!(
-            Errno::EPERM,
-            "the thread does not have the required capability"
-        )
     }
 
     /// Returns the owner UID of the user namespace.
--- a/kernel/src/process/posix_thread/alien_access.rs
+++ b/kernel/src/process/posix_thread/alien_access.rs
@@ -9,6 +9,7 @@ use bitflags::bitflags;
 use crate::{
     prelude::*,
     process::{credentials::capabilities::CapSet, posix_thread::PosixThread},
+    security::{self, PtraceAccessContext, PtraceAccessCreds, PtraceAccessKind, PtraceAccessMode},
 };
 
 impl PosixThread {
@@ -54,13 +55,19 @@ impl PosixThread {
             );
         }
 
-        yama::check_alien_access(accessor, self, mode, caller_has_cap)?;
+        security::ptrace_access_check(&PtraceAccessContext::new(
+            accessor,
+            self,
+            mode.into(),
+            caller_has_cap,
+        ))?;
 
         Ok(())
     }
 }
 
 /// The mode used by the alien access permission check.
+#[derive(Clone, Copy)]
 pub struct AlienAccessMode(AlienAccessFlags, CredsSource);
 
 impl AlienAccessMode {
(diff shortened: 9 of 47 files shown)


kernel/src/security/lsm/modules/capability.rs
```diff
@@ -0,0 +1,122 @@
...
+        if !ptr::eq(current_user_ns.as_ref(), context.user_namespace()) {
+            return_errno_with_message!(
+                Errno::EPERM,
+                "the target user namespace is outside the caller's capability scope"
+            );
+        }
+
+        // Capability bit checks are reason-independent. The reason stays in the
+        // shared context for modules that need operation-specific policy.
+        let _reason = context.reason();
```
StevenJiang1110: I am wondering what the `CapabilityReason` is intended for. You have defined it, but it is not used anywhere. I took a look at Linux's `capable_of` implementation, which has an `opts` parameter that accepts a few fixed values: `CAP_OPT_NONE`, `CAP_OPT_NOAUDIT`, and `CAP_OPT_INSETID`. However, these do not seem to correspond to what you have defined in `CapabilityReason`. Could you walk through a concrete example to explain why you need this reason in the first place?

kernel/src/security/lsm/modules/capability.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    fn flags(&self) -> LsmFlags {
+        LsmFlags::empty()
+    }
+}
+
+impl LsmCapabilityHook for CapabilityLsm {
+    fn on_capable(&self, context: &CapableContext) -> Result<()> {
+        let current_process = context.posix_thread().process();
```
StevenJiang1110: Retrieving the process from a `posix_thread` requires an `Arc::upgrade`, which involves an atomic operation. However, this atomic operation can be avoided for some cases, for example, if the caller has direct access to `ctx: &Context`. Since this is a general-purpose interface that will be called throughout the codebase, performance should be taken into consideration from the design stage.

kernel/src/security/mod.rs
```diff
@@ -20,3 +33,36 @@ pub(super) fn init() {
         tsm_mr::init();
     });
 }
+
+/// Runs the LSM stack for a capability check.
+pub fn capable(
```
StevenJiang1110: I think we can let the caller construct a `CapableContext`. Furthermore, this function is merely a thin wrapper around `lsm::capable`, so in the `security` module, a simple `pub use lsm::capable` should suffice.

kernel/src/security/mod.rs
```diff
@@ -20,3 +33,36 @@ pub(super) fn init() {
...
+        posix_thread,
+        capability,
+        reason,
+    ))
+}
+
+/// Updates security state after credentials are committed for a new executable.
+pub fn bprm_committed_creds(_path: &Path, _credentials: &Credentials<ReadWriteOp>) -> Result<()> {
+    Ok(())
+}
```
StevenJiang1110: Such no-op hook should not be added....

kernel/src/process/namespace/user_ns.rs
```diff
@@ -25,22 +26,15 @@ impl UserNamespace {
...
-        )
+    /// Checks whether the thread has the required capability in this user namespace
+    /// for a specific kernel operation.
+    pub fn check_cap_with_reason(
+        &self,
+        required: CapSet,
+        posix_thread: &PosixThread,
+        reason: CapabilityReason,
+    ) -> Result<()> {
+        security::capable(self, required, posix_thread, reason)
```
StevenJiang1110: I don't think we need `check_cap_with_reason` on `UserNamespace` anymore. The caller can directly use `security::capable`.

kernel/src/process/credentials/mod.rs
```diff
@@ -29,5 +31,5 @@ use crate::prelude::*;
 /// - secure bits.
 pub struct Credentials<R = FullOp>(Arc<Credentials_>, R);
 
-// TODO: Support the ambient capability set.
+// TODO: Support the bounding and ambient capability set.
```
StevenJiang1110: Bounding capset is already supported... What't the intent of this change?

kernel/src/process/credentials/file_capabilities.rs
```diff
@@ -0,0 +1,143 @@
+// SPDX-License-Identifier: MPL-2.0
```
StevenJiang1110: I think the whole file capability support should be removed from this PR. Supporting the capability module and centralizing the capability already introduce too many changes.

kernel/src/process/credentials/credentials_.rs
```diff
@@ -589,6 +625,21 @@ impl Credentials_ {
 
         self.securebits.try_store(securebits, Ordering::Relaxed)
     }
+
+    fn current_has_capability(&self, capability: CapSet, reason: CapabilityReason) -> bool {
+        let current = current_thread!();
+        let Some(posix_thread) = current.as_posix_thread() else {
```
StevenJiang1110: You call call `unwrap` for `as_posix_thread`, unless the method will be called by kernel thread.

kernel/src/process/credentials/credentials_.rs
```diff
@@ -589,6 +625,21 @@ impl Credentials_ {
 
         self.securebits.try_store(securebits, Ordering::Relaxed)
     }
+
+    fn current_has_capability(&self, capability: CapSet, reason: CapabilityReason) -> bool {
```
StevenJiang1110: I don't think having this helper method makes any sense. Just inline it.

kernel/src/fs/fs_impls/ext2/impl_for_vfs/inode.rs
```diff
@@ -255,6 +255,14 @@ impl Inode for Ext2Inode {
         self.get_xattr(name, value_writer)
     }
 
+    fn get_xattr_without_permission_check(
+        &self,
+        name: XattrName,
+        value_writer: &mut VmWriter,
+    ) -> Result<usize> {
+        self.get_xattr_without_permission_check(name, value_writer)
+    }
```
StevenJiang1110: This method looks quite odd. If we do want to skip the permission check for `get_xattr`, the right approach would be to move the permission check from the inode layer to the VFS layer, so that `inode::get_xattr` performs no permission checking itself.

That said, this PR does not need to address this, as I think it would be best to avoid supporting file capabilities in this PR.

kernel/src/net/socket/unix/ctrl_msg.rs
```diff
@@ -205,7 +206,7 @@ enum CControlType {
 #[derive(Default)]
 pub(super) struct AuxiliaryData {
     files: Vec<Arc<dyn FileLike>>,
-    cred: Option<SocketCred>,
+    cred: Option<CUserCred>,
```
StevenJiang1110: Is there any special reason to change the credential type? If not, please drop it from this PR.

kernel/src/process/credentials/static_cap.rs
```diff
@@ -307,41 +323,31 @@ impl<R: TRights> Credentials<R> {
         self.0.set_inheritable_capset(inheritable_capset);
     }
 
-    /// Sets the capabilities that a process can potentially be granted.
+    /// Sets the capabilities that are a process can potentially be granted.
```
StevenJiang1110: The changes seems incorrect.....

kernel/src/process/credentials/static_cap.rs
```diff
@@ -307,41 +323,31 @@ impl<R: TRights> Credentials<R> {
...
+    /// Gets keep capabilities flag.
     ///
     /// This method requires the `Read` right.
     #[require(R > Read)]
     pub fn keep_capabilities(&self) -> bool {
         self.0.keep_capabilities()
     }
 
-    /// Sets the keep-capabilities flag.
+    /// Sets keep capabilities flag.
```
StevenJiang1110: The all doc comments changes in this files seem to be incorrect.

kernel/src/security/lsm/modules/capability.rs
```diff
@@ -0,0 +1,122 @@
...
+        let current_user_ns = current_process.user_ns().lock();
+        if !ptr::eq(current_user_ns.as_ref(), context.user_namespace()) {
+            return_errno_with_message!(
+                Errno::EPERM,
+                "the target user namespace is outside the caller's capability scope"
+            );
```
StevenJiang1110: Is there a particular reason why the user namespaces need to be equal here? First, we currently do not support creating new user namespaces, so there is only one global user namespace. Second, even if we add support for user namespace creation in the future, there is no requirement for the two user namespaces to be equal — a process can hold certain capabilities within its own user namespace as well as all descendant namespaces. I do not think this check is necessary at all.

kernel/src/security/lsm/modules/capability.rs
```diff
@@ -0,0 +1,122 @@
...
+        // shared context for modules that need operation-specific policy.
+        let _reason = context.reason();
+        if context
+            .posix_thread()
+            .credentials()
+            .effective_capset()
+            .contains(context.capability())
+        {
+            return Ok(());
+        }
```
StevenJiang1110, line 41: I think you should move the original comment for the above check. Otherwise, directly checking the thread's capset is incorrect.

```rust
        // Since creating new user namespaces is not supported at the moment,
        // there is effectively only one user namespace in the entire system.
        // Therefore, the thread has a single set of capabilities used for permission checks.
        // FIXME: Once support for creating new user namespaces is added,
        // we should verify the thread's capabilities within the relevant user namespace.
```

kernel/src/security/lsm/modules/capability.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        return_errno_with_message!(
+            Errno::EPERM,
+            "the calling process does not have the required permissions"
+        );
+    }
+}
+
+impl LsmInodeHook for CapabilityLsm {
+    fn on_inode_dac_override(&self, context: &InodeDacOverrideContext) -> Result<Permission> {
```
StevenJiang1110: The semantics of this function look quite odd. All other `on_xxx` functions return `Result<()>` to indicate whether the check passed or failed, so why does this one return a permission? If possible, it would be preferable to have it return `Result<()>` as well to stay consistent.

StevenJiang1110 (commented):
Thanks for your contribution at first!

This PR is quite large, so I have not reviewed it in its entirety — I have only skimmed through the main changes so far.

One major concern is that this PR encompasses too many things at once, making it difficult to review. However, it can definitely be broken down into smaller, more focused pieces. I would suggest deferring file capabilities and all related changes to a follow-up PR. This PR should focus on introducing the capability module and consolidating capability-related checks under the LSM module, which is already a substantial and self-contained change in itself.

kernel/src/security/lsm/modules/capability.rs
zzj-5341: I originally added CapabilityReason to leave room for operation-specific policy in future LSM modules, but this PR does not have a concrete user for it. To keep the interface aligned with the current capability hook needs and avoid adding unused concepts, I removed CapabilityReason from this PR.

kernel/src/security/lsm/modules/capability.rs
zzj-5341: Good point. I changed the API so callers construct CapableContext and pass the relevant user namespace directly. The capability module no longer retrieves the process from PosixThread for this check.

kernel/src/security/mod.rs
zzj-5341: Done. security now re-exports lsm::capable, and callers construct CapableContext directly.

kernel/src/security/mod.rs
zzj-5341: Removed the no-op bprm_committed_creds hook.

kernel/src/process/namespace/user_ns.rs
zzj-5341: Removed it. Callers now use security::capable with CapableContext directly.

kernel/src/process/credentials/mod.rs
zzj-5341: Removed the file-capability-related changes and restored the unrelated bounding-capset changes.

kernel/src/process/credentials/file_capabilities.rs
zzj-5341: Done. File capability support has been removed from this PR and deferred to a follow-up PR.

kernel/src/process/credentials/credentials_.rs
zzj-5341: Removed this callsite together with the file-capability-related credential changes, so the unwrap concern no longer applies in this PR.

kernel/src/process/credentials/credentials_.rs
zzj-5341: Removed the helper and the related credential changes from this PR.

kernel/src/fs/fs_impls/ext2/impl_for_vfs/inode.rs
zzj-5341: Removed get_xattr_without_permission_check together with file capability support.

kernel/src/net/socket/unix/ctrl_msg.rs
zzj-5341: Dropped the credential type change. This file now only routes the existing CAP_SYS_ADMIN check through security::capable.

kernel/src/process/credentials/static_cap.rs
zzj-5341: Reverted the unrelated static_cap changes.

kernel/src/process/credentials/static_cap.rs
zzj-5341: Reverted the unrelated doc comment changes in static_cap.rs.

kernel/src/security/lsm/modules/capability.rs
zzj-5341: Removed the user namespace equality check. The capability module now keeps the current single-user-namespace assumption as a FIXME near the effective capset check.

kernel/src/security/lsm/modules/capability.rs
zzj-5341: Moved the single-user-namespace comment to the direct effective capset check, so the current simplified check is documented next to the code that relies on it.

kernel/src/security/lsm/modules/capability.rs
zzj-5341: Changed inode DAC override hooks to return Result<()> and updated the VFS permission check to pass only the missing DAC permissions to the hook.

zzj-5341:
> Thanks for your contribution at first!
> 
> This PR is quite large, so I have not reviewed it in its entirety — I have only skimmed through the main changes so far.
> 
> One major concern is that this PR encompasses too many things at once, making it difficult to review. However, it can definitely be broken down into smaller, more focused pieces. I would suggest deferring file capabilities and all related changes to a follow-up PR. This PR should focus on introducing the capability module and consolidating capability-related checks under the LSM module, which is already a substantial and self-contained change in itself.

Thanks for the detailed review. I've narrowed this PR to the capability LSM module and the migration of existing capability checks.

File capability support and related xattr/exec credential changes have been removed from this PR and will be handled separately. I also removed `CapabilityReason`, let callers construct `CapableContext` directly, re-exported `lsm::capable` from `security`, removed the `UserNamespace` helper and no-op hook, dropped the namespace equality check, and changed inode DAC override hooks to return `Result<()>`.

kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
```diff
@@ -39,7 +40,11 @@ impl CgroupNamespace {
         owner: Arc<UserNamespace>,
         posix_thread: &PosixThread,
     ) -> Result<Arc<Self>> {
-        owner.check_cap(CapSet::SYS_ADMIN, posix_thread)?;
+        security::capable(&security::CapableContext::new(
```
StevenJiang1110: I think `security::capable` should accept `CapableContext` by value rather than by reference (`&CapableContext`). Since `CapableContext` is a one-shot context, its ownership should be consumed upon use — passing it by reference does not properly reflect this semantics.

Other `on_xxx` that needs to accept some context needs to refactored similarly.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -531,65 +532,58 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
-        {
-            return_errno_with_message!(Errno::EACCES, "other permission check failed");
         }
 
-        Ok(())
+        if missing_perm.is_empty() {
+            return Ok(());
+        }
+
+        security::inode_dac_override(mode, missing_perm, posix_thread)
```
StevenJiang1110: In #3078, the usage pattern we established was `lsm_hooks::on_alien_access` — that is, other modules invoke LSM-related functions directly through the `lsm_hooks` module. However, this PR appears to introduce a new pattern where functions are called via `security::xxx` from the `security` module.

I think we should consolidate these into a single consistent pattern. Personally, I prefer standardizing on `lsm_hooks::on_xxx` for the following reasons:

1. `lsm_hooks` is a more specific and descriptive name than `security` — it immediately signals to the reader that this is a hook point in the LSM framework.
2. It eliminates the need to re-export items from the `lsm_hooks` module through `security`, reducing the amount of boilerplate `pub use` declarations.

That said, if you think `security::xxx` is the better choice, we can discuss it.

Please note that this change should be applied consistently across all existing call sites using `security::xxx`, not just the specific location I commented on.

kernel/src/syscall/setgroups.rs
```diff
@@ -5,13 +5,20 @@ use ostd::mm::VmIo;
...
 
 pub fn sys_setgroups(size: usize, group_list_addr: Vaddr, ctx: &Context) -> Result<SyscallReturn> {
     debug!("size = {}, group_list_addr = 0x{:x}", size, group_list_addr);
 
-    // TODO: check perm: the calling process should have the CAP_SETGID capability
+    security::capable(&security::CapableContext::new(
+        UserNamespace::get_init_singleton().as_ref(),
+        ctx.posix_thread,
+        CapSet::SETGID,
+    ))?;
```
StevenJiang1110, line 19: A common pattern in this PR is that capability checks, which previously only checked whether the current thread's credentials contain a certain capability, are now being updated to check whether that capability is held relative to a specific user namespace. This is semantically more correct — credentials should always be evaluated in the context of a user namespace.

However, we must be very careful to distinguish which user namespace is being used: the init user namespace, or the user namespace the current process belongs to. In this particular case, setgroups should check for the SETGID capability relative to the user namespace of the current process, not relative to the init user namespace. Therefore, `ctx.process.user_namespace()` should be used here.

Since we currently have only a single global user namespace, both the init user namespace and `ctx.process.user_namespace()` happen to point to the same object, so no bug would manifest in practice. Nevertheless, the two have different semantics, and we must ensure the semantics are correct regardless.

Please review all similar changes in this PR and double-check which user namespace each capability check is intended to be relative to.

kernel/src/syscall/setxattr.rs
```diff
@@ -186,18 +187,20 @@ pub(super) fn parse_xattr_name(name_str: &str) -> Result<XattrName<'_>> {
...
 }
+
+pub(super) fn current_can_access_trusted_xattr(ctx: &Context) -> bool {
+    security::capable(&security::CapableContext::new(
+        UserNamespace::get_init_singleton().as_ref(),
+        ctx.posix_thread,
+        CapSet::SYS_ADMIN,
+    ))
+    .is_ok()
+}
```
StevenJiang1110, line 211: I don't think introducing this helper makes any sense. Just inline it.

```rust
if namespace == XattrNamespace::Trusted {
    security::capable(&security::CapableContext::new(
        UserNamespace::get_init_singleton().as_ref(),
        ctx.posix_thread,
        CapSet::SYS_ADMIN,
    ))?
}
```

test/initramfs/src/regression/security/capability/kill.c
```diff
@@ -0,0 +1,81 @@
...
+
+FN_TEST(kill_requires_cap_kill_for_other_uid)
+{
+	int ready_pipe[2];
+	int release_pipe[2];
+	pid_t pid;
+	char ready_byte;
+	int status;
+
+	CHECK(pipe(ready_pipe));
```
StevenJiang1110: Use `TEST_SUCC` instead of `CHECK` in `FN_TEST` if possible.

test/initramfs/src/regression/security/capability/reboot.c
```diff
@@ -0,0 +1,61 @@
...
+	pid_t pid;
+	int status;
+
+	TEST_ERRNO(syscall(SYS_reboot, LINUX_REBOOT_MAGIC1, LINUX_REBOOT_MAGIC2,
+			   0xdeadbeefU, 0),
+		   EINVAL);
+
+	pid = TEST_SUCC(fork());
+	if (pid == 0) {
+		errno = 0;
```
StevenJiang1110: You don't need to manually reset `errno`. The below `syscall` will set it anyway.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -25,42 +28,21 @@ impl PosixThread {
...
-            );
-        }
+        let target_process = self.process();
+        let target_user_ns = target_process.user_ns().lock();
+        let accessor_has_cap_sys_ptrace = security::capable(&security::CapableContext::new(
+            target_user_ns.as_ref(),
+            accessor,
+            CapSet::SYS_PTRACE,
+        ))
+        .is_ok();
```
StevenJiang1110: Do we really need to call this here? Should `lsm_hooks::on_alien_access` instead delegate to `capability::on_alien_access` to perform the capability check?

I'm not entirely sure about this. However, performing a capability check outside of the LSM module does feel a bit out of place.

kernel/src/security/lsm/hooks/capability.rs
```diff
@@ -0,0 +1,57 @@
...
+    for module in modules::active_modules() {
+        module.on_capable(context)?;
+    }
+
+    Ok(())
+}
+
+/// The inputs for checking whether a thread has a capability.
+pub struct CapableContext<'a> {
+    user_namespace: &'a UserNamespace,
```
StevenJiang1110: Rename it to `target_user_ns` to make it explicit that this is the user namespace against which the capability check is being performed.

kernel/src/security/lsm/hooks/capability.rs
```diff
@@ -0,0 +1,57 @@
...
+    posix_thread: &'a PosixThread,
+    capability: CapSet,
```
StevenJiang1110: Renamed it to `required_cap`.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -140,6 +147,12 @@ impl ModuleSelection {
         self.selected_modules.into_boxed_slice()
     }
 
+    fn push_mandatory_modules(&mut self) {
+        for module in MANDATORY_MODULES {
+            self.push(module);
+        }
+    }
```
StevenJiang1110: This helper doesn't make much sense. Inline it.

kernel/src/syscall/listxattr.rs
```diff
@@ -94,12 +93,7 @@ fn listxattr(
 }
 
 fn get_current_xattr_namespace(ctx: &Context) -> XattrNamespace {
-    let credentials = ctx.posix_thread.credentials();
-    let permitted_capset = credentials.permitted_capset();
-    let effective_capset = credentials.effective_capset();
-
-    if permitted_capset.contains(CapSet::SYS_ADMIN) && effective_capset.contains(CapSet::SYS_ADMIN)
-    {
+    if current_can_access_trusted_xattr(ctx) {
```
StevenJiang1110: The new logic differs from the previous behavior: the original implementation checked both the permitted capability set and the effective capability set, whereas the updated one only checks the effective capability set. Is this an intentional change?

test/initramfs/src/regression/security/capability/run_test.sh
```diff
@@ -0,0 +1,28 @@
...
+./trusted_xattr
+
+echo "[capability] running kill"
+./kill
+
+echo "[capability] running reboot"
+./reboot
+
+echo "[capability] running execve"
+./execve
```
StevenJiang1110: I don't think you need to add a new `run_test.sh`. You can just include these test in `security/run_test.sh`.

But obviously, the test is outdated. For example. `execve` is missing....

test/initramfs/src/regression/security/lsm/module_selection.c
```diff
@@ -0,0 +1,126 @@
...
+
+FN_TEST(capability_static_procfs_entry_is_visible)
+{
+	int fd = TEST_SUCC(open(CAP_LAST_CAP_PATH, O_RDONLY));
+
+	if (fd >= 0) {
+		TEST_SUCC(close(fd));
+	}
+}
+END_TEST()
```
StevenJiang1110: This test has anything with module selection?

test/initramfs/src/regression/security/lsm/module_selection.c
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return module_list_contains(lsm_param, "yama");
+}
+
+FN_TEST(capability_static_procfs_entry_is_visible)
+{
+	int fd = TEST_SUCC(open(CAP_LAST_CAP_PATH, O_RDONLY));
+
+	if (fd >= 0) {
```
StevenJiang1110: You don't need to check `fd>=0` here. Please also check other similar cases.

Makefile
```diff
@@ -123,6 +128,9 @@ CARGO_OSDK_BUILD_ARGS += --kcmd-args="INTEL_TDX=$(INTEL_TDX)"
 CARGO_OSDK_BUILD_ARGS += --init-args="/test/run_regression_test.sh"
 else ifeq ($(AUTO_TEST), boot)
 CARGO_OSDK_BUILD_ARGS += --init-args="/test/boot_hello.sh"
+else ifeq ($(AUTO_TEST), lsm-module-selection)
+ENABLE_REGRESSION_TEST := true
+CARGO_OSDK_BUILD_ARGS += --init-args="/test/run_lsm_module_selection_test.sh"
```
StevenJiang1110: This does not like such a generic test that should live in the root Makefile. Please remove it.

test/initramfs/src/regression/scripts/run_lsm_module_selection_test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -e
+
+/test/security/lsm/module_selection
```
StevenJiang1110: I don't quite see what makes this test special enough to warrant a dedicated script. Could it not simply be included in `security/run_test.sh` instead?

StevenJiang1110 (commented):
Thanks for your revision.

kernel/src/fs/fs_impls/cgroupfs/cgroup_ns.rs
zzj-5341: Done. The public hook dispatch functions now consume CapableContext, AlienAccessContext, and InodeDacOverrideContext by value.

The module callbacks still borrow the context because the dispatcher needs to pass the same context through the active LSM modules.

kernel/src/fs/vfs/fs_apis/inode.rs
zzj-5341: Agreed. I standardized the LSM call sites on lsm_hooks::on_xxx and removed the corresponding security facade wrappers and re-exports.

kernel/src/syscall/setgroups.rs
zzj-5341: I audited the target user namespace used by all migrated capability checks.

setgroups and capset now use the current process user namespace. Checks associated with a target object or namespace use that target's owner user namespace, while genuinely global operations continue to use the init user namespace.

kernel/src/syscall/setxattr.rs
zzj-5341: Done. I removed the helper and inlined the trusted-xattr capability check at the call sites.

test/initramfs/src/regression/security/capability/kill.c
zzj-5341: Done. I replaced the applicable CHECK calls inside FN_TEST with TEST_SUCC.

test/initramfs/src/regression/security/capability/reboot.c
zzj-5341: Done. I removed the unnecessary errno reset.

kernel/src/process/posix_thread/alien_access.rs
zzj-5341: Agreed. AlienAccessContext now carries the target user namespace instead of a precomputed capability result.

The capability LSM performs the CAP_SYS_PTRACE check, and Yama performs its capability-based override through the capable hook.

kernel/src/security/lsm/hooks/capability.rs
zzj-5341: Done. I renamed the field to target_user_ns to make its semantics explicit.

kernel/src/security/lsm/hooks/capability.rs
zzj-5341: Done. I renamed the field to required_cap.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done. I removed push_mandatory_modules and inlined the mandatory-module insertion logic.

kernel/src/syscall/listxattr.rs
zzj-5341: Good catch. I restored the original trusted-xattr visibility semantics: CAP_SYS_ADMIN must still be present in both the permitted and effective capability sets.

The effective capability check is now routed through the capability LSM hook.

test/initramfs/src/regression/security/capability/run_test.sh
zzj-5341: Done. I removed the dedicated capability/run_test.sh and added the capability tests directly to security/run_test.sh, including execve.

test/initramfs/src/regression/security/lsm/module_selection.c
zzj-5341: Agreed. I removed the unrelated cap_last_cap check from the module-selection test.

test/initramfs/src/regression/security/lsm/module_selection.c
zzj-5341: Done. I removed the redundant fd >= 0 check and reviewed the other affected test code for similar cases.

Makefile
zzj-5341: Done. I removed the dedicated lsm-module-selection AUTO_TEST target and its Makefile boot-argument handling.

test/initramfs/src/regression/scripts/run_lsm_module_selection_test.sh
zzj-5341: Done. I removed the dedicated script and included the module-selection test in security/run_test.sh.

zzj-5341:
> Thanks for your revision.

Thanks for the detailed follow-up review. I addressed the comments in commit e8c9f5c28.

The changes align hook invocation with the #3078 lsm_hooks::on_xxx pattern, audit target user namespace semantics, move alien-access capability decisions into the LSM modules, preserve trusted-xattr behavior, and simplify the regression test organization.

I also verified the changes with cargo fmt --check, git diff --check, make kernel, and the full regression test run.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -2,28 +2,31 @@
 
 //! Built-in LSM module registration and boot-time selection.
 //!
-//! Linux provides two kernel command-line parameters for choosing enabled LSM
-//! modules: `lsm=` is the modern ordered module list, while `security=` is the
-//! legacy selector for one major LSM.
+//! The kernel always enables mandatory LSMs first. The capability module is
+//! mandatory because ordinary capability enforcement must not depend on boot
+//! parameters.
```
StevenJiang1110, line 7: Are all modules other than the `capability` module optional? Please add a comment here to make that explicit.

kernel/src/security/lsm/modules/mod.rs
```diff
@@ -37,10 +40,13 @@ static LEGACY_SECURITY_PARAM: Once<String> = Once::new();
...
+/// LSM modules that are always enabled before boot-selected modules.
+static MANDATORY_MODULES: [&'static dyn LsmModule; 1] = [&capability::CAPABILITY_LSM];
+
 /// All LSM modules compiled into the kernel.
-static ALL_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
+static ALL_MODULES: [&'static dyn LsmModule; 2] = [&capability::CAPABILITY_LSM, &yama::YAMA_LSM];
 
-/// The fallback LSM stack used when no boot-time selector is specified.
+/// The fallback optional LSM stack used when no boot-time selector is specified.
 pub(super) static DEFAULT_MODULES: [&'static dyn LsmModule; 1] = [&yama::YAMA_LSM];
```
StevenJiang1110: Rename the variable to `DEFAULT_OPTIOANAL_MODULES` to emphasize it does not include mandatory modules.

kernel/src/syscall/capset.rs
```diff
@@ -61,10 +63,17 @@ pub fn sys_capset(
...
-        && !credentials.effective_capset().contains(CapSet::SETPCAP)
+        && lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+            current_user_ns.as_ref(),
+            ctx.posix_thread,
+            CapSet::SETPCAP,
+        ))
+        .is_err()
     {
         return_errno_with_message!(Errno::EPERM, "inheritable capabilities are not permitted");
     }
```
StevenJiang1110, line 72: Why not use the following form? It's more natural.

```rust
f !(credentials.inheritable_capset() | credentials.permitted_capset())
        .contains(inheritable_capset)
        && !credentials.effective_capset().contains(CapSet::SETPCAP) {
    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
            current_user_ns.as_ref(),
            ctx.posix_thread,
            CapSet::SETPCAP,
        ))?;
} 
```

test/initramfs/src/regression/security/lsm/module_selection.c
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return false;
+}
+
+static bool expect_yama_enabled(void)
+{
+	char lsm_param[CMDLINE_BUFFER_SIZE] = "";
+
+	if (!find_last_lsm_param(lsm_param)) {
```
StevenJiang1110: Why is `lsm=yama` always the last module? Given that `find_last_lsm_param` only retrieves the value of the last `lsm=` parameter, would this assumption still hold if more modules are added in the future?

test/initramfs/src/regression/security/run_test.sh
```diff
@@ -4,10 +4,17 @@
...
 set -e
 
+cd "$(dirname "$0")"
+
 ./capability/capabilities
 ./capability/capset
+./capability/setgroups
+./capability/trusted_xattr
+./capability/kill
+./capability/reboot
```
StevenJiang1110, line 11: Keep the list ordered.

kernel/src/process/posix_thread/alien_access.rs
```diff
@@ -25,42 +21,14 @@ impl PosixThread {
...
-        if !caller_is_same && !caller_has_cap {
-            return_errno_with_message!(
-                Errno::EPERM,
-                "the calling process does not have the required permissions"
-            );
-        }
-
-        lsm_hooks::on_alien_access(&lsm_hooks::AlienAccessContext::new(
+        let target_process = self.process();
+        let target_user_ns = target_process.user_ns().lock();
```
StevenJiang1110: It looks like you don't need to retrieve `target_user_ns` here. Since `AlienAccessContext` already contains `self`, the user namespace can be obtained from `self` at the point where it is actually needed. There is no need to store `target_user_ns` as a field in `AlienAccessContext`.

StevenJiang1110 (approved):
LGTM! Thanks for your contribution! Just a few minor comments.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done. I clarified that capability is the only mandatory built-in LSM and that all other built-in LSM modules are optional and selected according to boot parameters.

kernel/src/security/lsm/modules/mod.rs
zzj-5341: Done. I renamed it to DEFAULT_OPTIONAL_MODULES to make it explicit that mandatory modules are not included.

kernel/src/syscall/capset.rs
zzj-5341: Done. I simplified the control flow and now propagate the on_capable result directly with `?`.

I did not keep a separate effective-capset fast path because the mandatory capability LSM performs that check, while routing the request through the hook allows every active LSM to participate.

test/initramfs/src/regression/security/lsm/module_selection.c
zzj-5341: There is no assumption that yama is the last module in the comma-separated module list. The helper retrieves the complete value of the effective `lsm=` parameter and then searches that list for yama.

I renamed it to `find_effective_lsm_param` and added a comment explaining the last-wins semantics of non-repeatable kernel parameters.

test/initramfs/src/regression/security/run_test.sh
zzj-5341: Done. I reordered the capability regression test list alphabetically.

kernel/src/process/posix_thread/alien_access.rs
zzj-5341: Done. I removed `target_user_ns` from `AlienAccessContext`. The capability and Yama modules now retrieve the target user namespace from `context.target()` at the point where it is needed.

kernel/src/process/kill.rs
```diff
@@ -174,11 +175,13 @@ fn check_signal_perm(target: &PosixThread, ctx: &Context, signum: Option<SigNum>
...
     }
 
-    if target_process
-        .user_ns()
-        .lock()
-        .check_cap(CapSet::KILL, ctx.posix_thread)
-        .is_ok()
+    let target_user_ns = target_process.user_ns().lock();
+    if lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        target_user_ns.as_ref(),
```
lrh2000:
```suggestion
    if lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
        target_process.user_ns().lock().as_ref(),
```

kernel/src/syscall/prlimit64.rs
```diff
@@ -94,11 +95,13 @@ fn do_prlimit64(
...
 // Reference: <https://man7.org/linux/man-pages/man2/prlimit.2.html>
 fn check_rlimit_perm(target_process: &Process, ctx: &Context) -> Result<()> {
-    if target_process
-        .user_ns()
-        .lock()
-        .check_cap(CapSet::SYS_RESOURCE, ctx.posix_thread)
-        .is_ok()
+    let target_user_ns = target_process.user_ns().lock();
+    if lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        target_user_ns.as_ref(),
```
lrh2000:
```suggestion
    if lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
        target_process.user_ns().lock().as_ref(),
```

kernel/src/syscall/setns.rs
```diff
@@ -232,13 +233,17 @@ fn set_uts_ns(
...
-        .borrow_user_ns()
-        .check_cap(CapSet::SYS_ADMIN, ctx.posix_thread)?;
+    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        target_ns.owner_user_ns().unwrap().as_ref(),
+        ctx.posix_thread,
+        CapSet::SYS_ADMIN,
+    ))?;
+    let current_user_ns = ctx.thread_local.borrow_user_ns();
+    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        current_user_ns.as_ref(),
```
lrh2000:
```suggestion
    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
        ctx.thread_local.borrow_user_ns().as_ref(),
```

kernel/src/syscall/setgroups.rs
```diff
@@ -5,13 +5,20 @@ use ostd::mm::VmIo;
...
+    security::lsm::hooks as lsm_hooks,
 };
 
 pub fn sys_setgroups(size: usize, group_list_addr: Vaddr, ctx: &Context) -> Result<SyscallReturn> {
     debug!("size = {}, group_list_addr = 0x{:x}", size, group_list_addr);
 
-    // TODO: check perm: the calling process should have the CAP_SETGID capability
+    let current_user_ns = ctx.process.user_ns().lock();
+    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        current_user_ns.as_ref(),
```
lrh2000:
```suggestion
    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
        ctx.thread_local.borrow_user_ns().as_ref(),
```

Note `ctx.process` -> `ctx.thread_local`. It is faster.

kernel/src/syscall/capset.rs
```diff
@@ -61,10 +63,15 @@ pub fn sys_capset(
...
     if !(credentials.inheritable_capset() | credentials.permitted_capset())
         .contains(inheritable_capset)
-        && !credentials.effective_capset().contains(CapSet::SETPCAP)
     {
-        return_errno_with_message!(Errno::EPERM, "inheritable capabilities are not permitted");
+        lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+            current_user_ns.as_ref(),
```
lrh2000, line 68:
```suggestion
        let current_user_ns = ctx.thread_local.borrow_user_ns();
        lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
            current_user_ns.as_ref(),
```

test/initramfs/src/regression/security/capability/kill.c
```diff
@@ -0,0 +1,81 @@
...
+		char byte;
+
+		TEST_SUCC(close(ready_pipe[0]));
+		TEST_SUCC(close(release_pipe[1]));
+		TEST_SUCC(setresgid(1000, 1000, 1000));
+		TEST_SUCC(setresuid(1000, 1000, 1000));
+		TEST_SUCC(write(ready_pipe[1], "r", 1));
+		TEST_SUCC(close(ready_pipe[1]));
+		TEST_SUCC(read(release_pipe[0], &byte, sizeof(byte)));
+		TEST_SUCC(close(release_pipe[0]));
```
lrh2000: In the child process, use `CHECK()`.

`TEST_SUCC` has no effects in a child process.

test/initramfs/src/regression/security/capability/trusted_xattr.c
```diff
@@ -0,0 +1,104 @@
...
+	cap_data[cap_index].effective &= ~cap_mask;
+	cap_data[cap_index].permitted &= ~cap_mask;
+	cap_data[cap_index].inheritable &= ~cap_mask;
+	CHECK(syscall(SYS_capset,
+		      &(struct __user_cap_header_struct){
+			      .version = _LINUX_CAPABILITY_VERSION_3,
+			      .pid = 0,
+		      },
+		      cap_data));
+}
```
lrh2000: There are too many repetitions of these code. Both in this PR and in existing code:
```
> grep -R "CHECK.*SYS_cap" test/initramfs
test/initramfs/src/regression/fs/procfs/pid_mem.c:	CHECK(syscall(SYS_capget, &hdr, &capdat));
test/initramfs/src/regression/fs/procfs/pid_mem.c:	CHECK(syscall(SYS_capset, &hdr, &capdat));
test/initramfs/src/regression/process/prctl/capbset.c:	CHECK(syscall(SYS_capget, &cap_header, cap_data));
test/initramfs/src/regression/security/lsm/yama.c:	CHECK(syscall(SYS_capget, &hdr, &capdat));
test/initramfs/src/regression/security/lsm/yama.c:	CHECK(syscall(SYS_capset, &hdr, &capdat));
```
And also:
```
> grep -R "static.*[^a-z]cap_.*(.*)$" test/**/*.c
test/initramfs/src/regression/fs/procfs/pid_mem.c:static void drop_cap_sys_ptrace(void)
test/initramfs/src/regression/process/prctl/capbset.c:static void read_cap_data(struct __user_cap_data_struct cap_data[2])
test/initramfs/src/regression/process/prctl/capbset.c:static int __write_cap_data(const struct __user_cap_data_struct cap_data[2])
test/initramfs/src/regression/process/prctl/capbset.c:static void write_cap_data(const struct __user_cap_data_struct cap_data[2])
test/initramfs/src/regression/security/lsm/yama.c:static void drop_cap_sys_ptrace(void)
```

Could you consider extract them to a common place, e.g., `test/initramfs/src/regression/common/capability.h`, so we don't need to repeat?

e.g.,
```c
// See below for a reference implementation.
static int __read_cap_data(struct __user_cap_data_struct cap_data[2]);
static void read_cap_data(struct __user_cap_data_struct cap_data[2]);

static int __write_cap_data(const struct __user_cap_data_struct cap_data[2])
{
	struct __user_cap_header_struct cap_header = {
		.version = _LINUX_CAPABILITY_VERSION_3,
		.pid = 0,
	};

	return syscall(SYS_capset, &cap_header, cap_data);
}

static void write_cap_data(const struct __user_cap_data_struct cap_data[2])
{
	CHECK(__write_cap_data(cap_data));
}

static void drop_capability(int capability)
{
	// Implement this via read_cap_data and write_cap_data
}
```

lrh2000 (commented):
Thanks for the work!

There are several changes that will prolong the existing `user_ns().lock()` guard lifetime. Is this intentional? If not, please try to revert them (see the suggestions below).

In addition, when accessing the current thread's user namespace, please use `thread_local.borrow_user_ns()` instead `process.user_ns.lock()`. They're equivalent, but the latter is faster since it does not need to take a lock. See the suggestions below.

There are also some suggestions about the regression tests. See below.

Could you please also tidy up the commit history? We prefer atomic commits, so the commit addressing previous reviewer comments can be squashed back to the commit for the original functionality.

kernel/src/process/kill.rs
zzj-5341: Done. I inlined the target user namespace guard into the capability-hook call so it is dropped immediately after the check.

kernel/src/syscall/prlimit64.rs
zzj-5341: Done. I inlined the target user namespace guard into the capability-hook call so it is dropped immediately after the check.

kernel/src/syscall/setns.rs
zzj-5341: Done. The current user namespace is now borrowed directly from ctx.thread_local for the capability check.

kernel/src/syscall/setgroups.rs
zzj-5341: Done. I switched the current user namespace lookup from ctx.process.user_ns().lock() to ctx.thread_local.borrow_user_ns().

kernel/src/syscall/capset.rs
zzj-5341: Done. capset now borrows the current user namespace from ctx.thread_local only in the branch that requires the SETPCAP check.

test/initramfs/src/regression/security/capability/kill.c
zzj-5341: Done. I replaced TEST_SUCC with CHECK in the child process because child-side test accounting is not reported to the parent.

test/initramfs/src/regression/security/capability/trusted_xattr.c
zzj-5341: Done. I added common/capability.h with shared capget, capset, and drop_capability helpers.

I migrated the repeated helpers in the new capability tests and the existing pid_mem, capbset, and Yama tests. Tests that directly validate capget/capset behavior continue to invoke those syscalls explicitly.

zzj-5341:
> Thanks for the work!
> 
> There are several changes that will prolong the existing `user_ns().lock()` guard lifetime. Is this intentional? If not, please try to revert them (see the suggestions below).
> 
> In addition, when accessing the current thread's user namespace, please use `thread_local.borrow_user_ns()` instead `process.user_ns.lock()`. They're equivalent, but the latter is faster since it does not need to take a lock. See the suggestions below.
> 
> There are also some suggestions about the regression tests. See below.
> 
> Could you please also tidy up the commit history? We prefer atomic commits, so the commit addressing previous reviewer comments can be squashed back to the commit for the original functionality.

Thanks for the review. I addressed all the namespace guard lifetime and regression-test comments.

I also rebased the PR onto the latest main branch and consolidated the history into two atomic commits: one for the capability LSM implementation and one for regression coverage.

The changes were verified with cargo fmt --check, clang-format, make kernel, and the full regression test suite.

test/initramfs/src/regression/security/run_test.sh
```diff
@@ -4,10 +4,17 @@
 
 set -e
 
+cd "$(dirname "$0")"
+
```
lrh2000: I think adding this is helpful, but if you want to add this, please add it to all other `run_test.sh` files for consistency purposes.

(Of course, you can also just remove this from this PR...)

kernel/src/syscall/capset.rs
```diff
@@ -61,10 +62,15 @@ pub fn sys_capset(
...
     {
-        return_errno_with_message!(Errno::EPERM, "inheritable capabilities are not permitted");
+        let current_user_ns = ctx.thread_local.borrow_user_ns();
+        lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+            current_user_ns.as_ref(),
+            ctx.posix_thread,
+            CapSet::SETPCAP,
+        ))?;
     }
+
```
lrh2000: Is this blank line added intentionally? If not, please remove it.

kernel/src/net/socket/util/options.rs
lrh2000, line 1: This is not updated:
https://github.com/asterinas/asterinas/blob/d01367944019248aaf958cd9155f4fe152e603c8/kernel/src/net/socket/util/port_privilege.rs#L18-L31

Should we also update this?
https://github.com/asterinas/asterinas/blob/d01367944019248aaf958cd9155f4fe152e603c8/kernel/src/fs/fs_impls/ext2/fs.rs#L564-L574

lrh2000 (commented):
Thanks for the revision. I only notice three minor problems. Otherwise look good!

test/initramfs/src/regression/security/run_test.sh
zzj-5341: Removed the unnecessary blank line.

kernel/src/syscall/capset.rs
zzj-5341: Removed the unnecessary blank line.

kernel/src/net/socket/util/options.rs
zzj-5341: Yes. I updated both remaining external capability checks to go through
`lsm_hooks::on_capable`: `CAP_NET_BIND_SERVICE` for privileged-port binding
and `CAP_SYS_RESOURCE` for ext2 reserved-block allocation.

The privileged-port path still preserves the existing `EACCES` behavior on
denial. I also checked that no external direct `effective_capset()` capability
checks remain outside credential internals and the capability LSM module.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -531,65 +532,62 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
-        {
-            return_errno_with_message!(Errno::EACCES, "other permission check failed");
         }
 
-        Ok(())
+        if missing_perm.is_empty() {
+            return Ok(());
+        }
+
+        lsm_hooks::on_inode_dac_override(lsm_hooks::InodeDacOverrideContext::new(
```
lrh2000: https://elixir.bootlin.com/linux/v7.1-rc7/source/fs/namei.c#L534-L537
```c
		if (capable_wrt_inode_uidgid(idmap, inode,
					     CAP_DAC_OVERRIDE))
			return 0;
		return -EACCES;
```

It does not appear that the Linux source code has an LSM hook in the DAC override path, except for the `ns_capable` hook (called in `capable_wrt_inode_uidgid`), which checks the capability itself and is not specific to the DAC override.

So... Is there some justification why we need `on_inode_dac_override` and `InodeDacOverrideContext`, as well as the changes in `Inode::check_permission`?

Maybe I'm missing some context for it? Not sure...

test/initramfs/src/regression/security/run_test.sh
lrh2000: > Removed the unnecessary blank line.

Emmm? I mean `cd "$(dirname "$0")"`. Not the blank line.

The blank line between `./capability/trusted_xattr` and `./lsm/module_selection` should be added back.

kernel/src/fs/vfs/fs_apis/inode.rs
zzj-5341: Good point. I removed the dedicated `on_inode_dac_override` / `InodeDacOverrideContext` hook and switched the DAC override path back to a direct `on_capable(..., CAP_DAC_OVERRIDE)` check in `Inode::check_permission`, while keeping the execute-bit restriction there.

This should be closer to Linux's `capable_wrt_inode_uidgid(..., CAP_DAC_OVERRIDE)` usage.

test/initramfs/src/regression/security/run_test.sh
zzj-5341: Fixed. I removed `cd "$(dirname "$0")"` and kept the blank line between the capability and LSM test groups.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -531,24 +536,21 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
-                Some(thread) => thread.credentials(),
-                None => return Ok(()),
-            },
+        let current = match Task::current() {
+            Some(task) => task,
             None => return Ok(()),
         };
+        let Some(posix_thread) = current.as_posix_thread() else {
+            return Ok(());
+        };
```
lrh2000, line 544: Use `if let Some(task) = Task::current() else {` to keep the format consistent:
```rust
        let Some(task) = Task::current() else {
            return Ok(());
        };
        let Some(posix_thread) = current.as_posix_thread() else {
            return Ok(());
        };
```

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -531,24 +536,21 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
 
-        // With DAC_OVERRIDE capability, the user can bypass some permission checks.
-        if creds.effective_capset().contains(CapSet::DAC_OVERRIDE) {
+        if has_dac_override_capability(posix_thread) {
```
lrh2000: Keep the original comment:
```rust
        // With DAC_OVERRIDE capability, the user can bypass some permission checks.
        if has_dac_override_capability(posix_thread) {
```

kernel/src/fs/vfs/fs_apis/inode.rs
lrh2000: The intersection seems redundant:
```rust
        perm =
            perm.intersection(Permission::MAY_READ | Permission::MAY_WRITE | Permission::MAY_EXEC);
```
Given that we only check `may_read()`/`may_write()`/`may_exec()` below anyway. 

So if the redundant intersection can be removed, we can define these variables together, which look better:
```rust
        let creds = posix_thread.credentials();
        let metadata = self.metadata();
        let mode = metadata.mode;
```

.github/workflows/test_x86.yml
```diff
@@ -33,7 +33,6 @@ jobs:
         uses: ./.github/actions/test
         with:
           auto_test: 'general'
-          runs_on: 'ubuntu-latest'
```
lrh2000, line 36: I don't believe changes under `.github/` are relevant given that the workflow passes in the main branch.
https://github.com/asterinas/asterinas/commit/e61a873f634e37d8a519ad63045914931f0fcd63

lrh2000 (commented):
Thanks for updates!

kernel/src/fs/vfs/fs_apis/inode.rs
zzj-5341: Good point. I changed this to the `let Some(task) = Task::current() else { ... }` form and used `task` consistently in the following lookup. This is included in the latest update.

kernel/src/fs/vfs/fs_apis/inode.rs
zzj-5341: Good point. I restored the original summary comment for this DAC override block in the latest update.

kernel/src/fs/vfs/fs_apis/inode.rs
zzj-5341: I removed the redundant `perm.intersection(...)` and grouped `creds`, `metadata`, and `mode` together in the latest update.

.github/workflows/test_x86.yml
zzj-5341: Good point. I removed the unrelated `.github/workflows` changes from this PR in the latest force-push

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -531,24 +536,22 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
+        let metadata = self.metadata();
+        let mode = metadata.mode;
+
         // With DAC_OVERRIDE capability, the user can bypass some permission checks.
-        if creds.effective_capset().contains(CapSet::DAC_OVERRIDE) {
-            // Read/write DACs are always overridable.
+        if has_dac_override_capability(&task, posix_thread) {
             perm -= Permission::MAY_READ | Permission::MAY_WRITE;
 
-            // Executable DACs are overridable when there is at least one exec bit set.
```
lrh2000, line 547: >            // Read/write DACs are always overridable.
>             // Executable DACs are overridable when there is at least one exec bit set.
Then you remove two different comments. Please add them back...

```rust
            // Read/write DACs are always overridable.
            perm -= Permission::MAY_READ | Permission::MAY_WRITE;

            // Executable DACs are overridable when there is at least one exec bit set.
            if perm.may_exec() {
```

zzj-5341: I had restored the summary comment but missed these two inline comments in the DAC override block, so I added them back in the latest update.

lrh2000: This should be done in the first commit instead of the second, right?

These are minor issues and perhaps not important enough to prevent merging. However, it's quite obvious if you review the diffs of the two commits...

zzj-5341: You're right, and sorry about that. This comment-only fix belongs to the first implementation commit rather than the regression-coverage commit, so I moved it into the first commit and force-pushed the updated two-commit history.

lrh2000 (approved):

lrh2000:
@StevenJiang1110 I will leave the merge to you if the merge requirement is met later.

cqs21 (approved):

Issue #3132, opened 2026-04-15
The kernel cannot be built due to core2 is yanked


StevenJiang1110:
It appears that the [core2 crate was yanked last night](https://crates.io/crates/core2). Unfortunately, our project depends on it heavily, which means all CI pipelines are currently failing as a result.

There are a few places where we rely on it:

1. Both the kernel and cpio-decoder depend on it directly.
2. We also depend on libflate for decompression, which transitively depends on core2.

For the first issue, the fix is straightforward — we can replace all direct usages with [no_std_io2](https://crates.io/crates/no_std_io2), which is a drop-in replacement with an identical API, so the migration should be relatively simple.

For the second issue, we can wait for an upstream fix. There is already a PR (https://github.com/sile/libflate/pull/86) addressing this. If the upstream fix is not landed in time, we could either switch to an alternative compression library or wrap the underlying primitives ourselves.

tatetian:
The maintainer of `libflate` says he will review the patch to replace `core2` in a way. So I think we should figure out a temp fix rather than replacing `libflate`.

And another person [suggested](https://github.com/sile/libflate/issues/85#issuecomment-4246788936) using `patch.crates-io` as a temp fix:

```
[patch.crates-io]
core2 = { git = "https://github.com/bbqsrc/core2", rev = "545e84bcb0f235b12e21351e0c69767958efe2a7" }
```

This should be good enough as a temp fix?

StevenJiang1110:
libflate has published a new version 35 minutes ago.....So the CI still fails now with #3134. I will fix this problem soom

PR #3134, opened 2026-04-15
Use patch section to fix core2 yank temporarily

StevenJiang1110:
The related discussion can be found in #3132.

This is a temporary fix. Once upstream libflate publishes a new release with the fix, this commit should be reverted.

Note that `cargo publish` [automatically strips the `[patch]` section](https://doc.rust-lang.org/cargo/commands/cargo-package.html#description), so the osdk and ostd publish CI jobs will still fail with this change. We must fix this problem before next release.

```plain
Create the compressed .crate file.

- The original Cargo.toml file is rewritten and normalized.
- [patch], [replace], and [workspace] sections are removed from the manifest.

.......
```

### DIFF
--- a/osdk/Cargo.toml
+++ b/osdk/Cargo.toml
@@ -33,3 +33,8 @@ rustc_version = "0.4.1"
 
 [dev-dependencies]
 assert_cmd = "2.0.13"
+
+[patch.crates-io]
+# Temporary workaround for the `core2` yank; see #3132.
+# TODO: Once upstream `libflate` publishes the fix and `core2` is available again on crates.io, revert the workaround introduced by PR #3134.
+core2 = { git = "https://github.com/bbqsrc/core2", rev = "545e84bcb0f235b12e21351e0c69767958efe2a7" }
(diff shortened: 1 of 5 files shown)


osdk/Cargo.toml
```diff
@@ -33,3 +33,8 @@ rustc_version = "0.4.1"
 
 [dev-dependencies]
 assert_cmd = "2.0.13"
+
+[patch.crates-io]
+# Temporary workaround for the `core2` yank; see #3132.
+# TODO: Remove this git rev pin once upstream `libflate` publishes the fix and `core2` is available again on crates.io.
```
tatetian: In the TODO, say "revert this entire commit XXXX" (give the commit number). Otherwise, we may forget to the `copy_core2_patch` change in `base_crate/mod.rs`

tatetian (commented):
Thanks for the quickfix

osdk/Cargo.toml
StevenJiang1110: Updated the TODO to point to this PR instead of a commit hash, since the commit hash always changes after `--amend`. This should still make it clear that the whole workaround needs to be reverted later.

tatetian (approved):

PR #3136, opened 2026-04-15
Bump libflate to 2.3.0 and replace core2 with no_std_io2

### LINKED ISSUE #3132: The kernel cannot be built due to core2 is yanked

StevenJiang1110:
Fixes #3132


tatetian (approved):

PR #3139, opened 2026-04-15
Replace `impl From<&CpuException> for FaultSignal` with `ToFaultSignal`

lrh2000:
Previously, we used `impl From<&CpuException> for FaultSignal` to build a fault signal from a CPU exception. However, this approach is problematic.
 1. The `unreachable!()` macro should not appear in the `From` trait. As quoted from [the standard library's documentation](https://doc.rust-lang.org/std/convert/trait.From.html), it does not make much sense:
https://github.com/asterinas/asterinas/blob/83edf72dd107d253ea46902db3ee71e6a823ec2f/kernel/src/arch/riscv/signal.rs#L31
> Note: This trait must not fail. The `From` trait is intended for perfect conversions. If the conversion can fail or is not perfect, use `TryFrom`.
 2. Some information is unavailable. See this FIXME:
https://github.com/asterinas/asterinas/blob/83edf72dd107d253ea46902db3ee71e6a823ec2f/kernel/src/arch/riscv/signal.rs#L21-L23

This PR proposes to add a `ToFaultSignal` trait instead.
```rust
/// A trait that converts CPU exceptions into fault signals.
///
/// This trait should be implemented by architecture-specific code.
pub trait ToFaultSignal {
    /// Converts a CPU exception into a fault signal.
    ///
    /// Returns `None` if the exception must be handled earlier and cannot be
    /// delivered as a signal.
    ///
    /// POSIX [requires] `SIGILL` and `SIGFPE` to report the address of the
    /// faulting instruction, and `SIGSEGV` and `SIGBUS` to report the address
    /// of the faulting memory reference. Linux behavior, however, is highly
    /// architecture-specific and does not always follow POSIX: for some
    /// exceptions, it reports neither the expected code nor the expected
    /// address. Linux may also attach an instruction address to `SIGTRAP`, but
    /// this behavior is not consistent across architectures.
    ///
    /// [requires]: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/signal.h.html
    fn to_fault_signal(&self, user_ctx: &UserContext) -> Option<FaultSignal>;
}
```

With the `user_ctx` argument, we can fill more information to the fault signals. So this PR also resolves the aforementioned TODO.

In addition, we previously can also trigger kernel panics by generating #SS and #OF exceptions from userspace:
```c
// This generates #OF. The kernel should respond with a SIGSEGV.
static void generate_of(void)
{
       asm volatile("int $4");
}

// This generates #SS. The kernel should respond with a SIGBUS.
static void generate_ss(void)
{
       asm volatile("movabs $0x8000000000000000, %%rax\n\t"
                    "mov %%rax, %%rsp\n\t"
                    "push %%rax\n\t"
                    :
                    :
                    : "rax", "memory");
}
```

This PR fixes the reachable kernel panics and adds some more regression tests.

However, generating #BR from the user space is impossible with our x86-64 settings. Although there are known methods, none of them work.
```c
// The `bound` instruction is not a valid instruction in 64-bit mode.
// So it generates SIGILL.
static void try_generate_br_via_bound(void)
{
	// bound %eax, (%eax)
	asm volatile(".byte 0x62, 0x00");
}

// The `int` instruction cannot trigger #BR: the vector-5 gate has DPL 0,
// so it is callable only from Ring 0.
static void try_generate_br_via_int5(void)
{
	asm volatile("int $5");
}

// "When Intel MPX is not enabled or not present, all Intel MPX
// instructions behave as NOP."
static void try_generate_br_via_mpx(void)
{
	// bndcl (%rax), %bnd0
	asm volatile(".byte 0xf3, 0x0f, 0x1a, 0x00");
}
```

As mentioned in the `ToFaultSignal` trait's documentation, the Linux implementation is highly architecture-specific and does not consistently adhere to the POSIX standard. I reviewed all the existing code manually and then with the help of Codex, adjusting the signal information accordingly. I added comments in one of the following two cases:
1. The behavior does not match that of Linux due to limitations in our implementation. I add `TODO`/`FIXME` comments for this case. Example:
```rust
            CpuException::X87FloatingPointException | CpuException::SIMDFloatingPointException => {
                // TODO: Derive the code from the floating-point status.
                (SIGFPE, FPE_FLTDIV, Some(rip))
            }
```
2. The behavior does not match POSIX standards, usually because the code and/or address is missing or incorrect. In this case, I include a reference to the Linux source code for confirmation. Example:
```rust
            LoadMisaligned(_) | StoreMisaligned(_) => {
                // The address should be the memory address, but Linux reports the instruction
                // address. So we follow it.
                //
                // Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/riscv/kernel/traps.c#L230-L232>
                (SIGBUS, BUS_ADRALN, sepc)
            }
```

### DIFF
--- a/kernel/src/arch/loongarch/signal.rs
+++ b/kernel/src/arch/loongarch/signal.rs
@@ -1,10 +1,14 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use loongArch64::register::estat::Exception;
-use ostd::arch::cpu::context::{CpuExceptionInfo, UserContext};
+use ostd::{
+    arch::cpu::context::{CpuExceptionInfo, UserContext},
+    user::UserContextApi,
+};
 
-use crate::process::signal::{
-    SignalContext, constants::*, sig_num::SigNum, signals::fault::FaultSignal,
+use crate::{
+    process::signal::{SignalContext, sig_num::SigNum, signals::fault::FaultSignal},
+    thread::exception::ToFaultSignal,
 };
 
 impl SignalContext for UserContext {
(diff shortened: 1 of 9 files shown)


StevenJiang1110 (approved):

cqs21 (approved):

PR #3122, opened 2026-04-13
Reimplement vsock

lrh2000:
This PR removes the old vsock implementation and reimplements it entirely.

#### The new implementation features:
 1. Better architecture. It is similar to our TCP/IP stack.
   - `kernel/comps/virtio/src/device/socket` (an equivalent to `kernel/comps/virtio/src/device/network`):
```rust
//! Basic virtio-vsock device support.
//!
//! This module implements the transmit and receive primitives of a virtio-vsock device, together
//! with the minimal queue management and callbacks required to drive the hardware. Higher layers
//! are responsible for protocol-level logic, such as interpreting received packet operations and
//! managing connection state.
//!
//! For a quick start, look up the device using the [`get_device`] function, then send packets via
//! the transmit queue using the [`SocketDevice::lock_tx`] method and receive packets via the
//! receive queue using the [`SocketDevice::lock_rx`] method. Notifications for newly received
//! packets happen via the callback specified in the [`SocketDevice::init_rx_callback`] method,
//! which is called in the bottom half.
```
   - `kernel/src/net/socket/vsock/transport` (an equivalent to `kernel/libs/aster-bigtcp`):
```rust
//! The virtio-vsock transport protocol.
//!
//! Built on top of the [`basic virtio-vsock device support`](`aster_virtio::device::socket`), this
//! module manages connections and listeners, handles connection establishment, data transfer, and
//! shutdown, and implements credit-based flow control plus I/O event checking and notification.
//! The socket layer is expected to build on these APIs to provide the user-visible socket
//! interface.
//!
//! For a quick start, bind to a port by creating a [`BoundPort`] instance.
//!  - To connect to a remote address, use the [`BoundPort::connect`] method and get a
//!    [`Connection`] instance. Data can be transmitted or received via the [`Connection::try_send`]
//!    and [`Connection::try_recv`] methods.
//!  - To listen to the local address, use the [`BoundPort::listen`] method and get a [`Listener`]
//!    instance. Incoming connections can be accepted via the [`Listener::try_accept`] method.
//!
//! Drop the [`BoundPort`], [`Connection`], or [`Listener`] instance will shut down the underlying
//! connection (if any) and release the resources.
```
   - `kernel/src/net/socket/vsock/stream` (an equivalent to `kernel/src/net/socket/ip/stream`): It implements the socket API via the `Socket` trait.
   - The socket state machine runs in the bottom half and appears to be the first user of our `Taskless` API.
 2. Better performance.
[...]

PR #3141, opened 2026-04-16
Reject binding to privileged ports without `CAP_NET_BIND_SERVICES`

lrh2000:
https://man7.org/linux/man-pages/man7/capabilities.7.html:
>       CAP_NET_BIND_SERVICE
>              Bind a socket to Internet domain >privileged ports (port
>              numbers less than 1024).

This PR adds a helper method to check privileged ports and places it under `kernel/src/net/socket/util/port_privilege.rs`. This is because vsock (https://github.com/asterinas/asterinas/pull/3122) will also use that logic.

https://man7.org/linux/man-pages/man7/vsock.7.html:
> The
>        port numbers below 1024 are called privileged ports.  Only a
>       process with the CAP_NET_BIND_SERVICE capability may bind(2) to
>       these port numbers.


StevenJiang1110 (approved):

cqs21 (approved):

PR #3143, opened 2026-04-16
Fix cross-compilation issues in OSDK and NixOS

### LINKED ISSUE #3138: `make run_nixos` failed with arm64 dev container
As title. It seems that QEMU cannot find the entry point.

```log
BdsDxe: failed to load Boot0002 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x2,0x0): Not Found
BdsDxe: failed to load Boot0003 "UEFI Misc Device 2" from PciRoot(0x0)/Pci(0x6,0x0): Not Found
BdsDxe: failed to load Boot0004 "UEFI Misc Device 3" from PciRoot(0x0)/Pci(0x7,0x0): Not Found
```

### To Reproduce

```sh
make nixos
make ENABLE_KVM=0 run_nixos
```

### Environment

- Hardware: MacBook Air M3 (macOS 26)
- Conteiner: OrbStack (Image: `0.17.2-20260407`)

junyang-zh:
fix #3138

The first commits reverts `OSDK_TARGET_ARCH` to `TARGET_ARCH`, since there's an OSDK argument `--target` for this purpose. Duplicating the functionality with another environment variable is weird. Letting other scripts unrelated with OSDK use this variable is also weird because of the `OSDK_` prefix.

The second commit enforces building `TARGET_ARCH` NixOS image, not the host-native image. The root cause of #3138 is failing to boot Arm64 NixOS on x86 VM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # =========================== Makefile options. ===============================
 
 # Global build options.
-OSDK_TARGET_ARCH ?= x86_64
+TARGET_ARCH ?= x86_64
 BENCHMARK ?= none
 BOOT_METHOD ?= grub-rescue-iso
 BOOT_PROTOCOL ?= multiboot2
@@ -88,12 +88,14 @@ DEV_TRUSTED_PUBLIC_KEY ?= aster-nixos-dev.cachix.org-1:xrCbE2flfliFTQCY/2HeJoT2t
 
 # ========================= End of Makefile options. ==========================
 
+export OSDK_TARGET_ARCH=$(TARGET_ARCH)
+
 SHELL := /bin/bash
 
 CARGO_OSDK := ~/.cargo/bin/cargo-osdk
 
 # Common arguments for `cargo osdk` `build`, `run` and `test` commands.
-CARGO_OSDK_COMMON_ARGS := --target-arch=$(OSDK_TARGET_ARCH)
+CARGO_OSDK_COMMON_ARGS :=
 # The build arguments also apply to the `cargo osdk run` command.
 CARGO_OSDK_BUILD_ARGS := --kcmd-args="ostd.log_level=$(LOG_LEVEL)"
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="console=$(CONSOLE)"
@@ -122,7 +124,7 @@ CARGO_OSDK_COMMON_ARGS += --profile release-lto
 OSTD_TASK_STACK_SIZE_IN_PAGES = 8
 else ifeq ($(RELEASE), 1)
 CARGO_OSDK_COMMON_ARGS += --release
-	ifeq ($(OSDK_TARGET_ARCH), riscv64)
+	ifeq ($(TARGET_ARCH), riscv64)
 	# FIXME: Unwinding in RISC-V seems to cost more stack space, so we increase
 	# the stack size for it. This may need further investigation.
 	# See https://github.com/asterinas/asterinas/pull/2383#discussion_r2307673156
@@ -151,9 +153,9 @@ BOOT_METHOD = qemu-direct
 endif
 
 ifeq ($(SCHEME), "")
-	ifeq ($(OSDK_TARGET_ARCH), riscv64)
+	ifeq ($(TARGET_ARCH), riscv64)
 	SCHEME = riscv
-	else ifeq ($(OSDK_TARGET_ARCH), loongarch64)
+	else ifeq ($(TARGET_ARCH), loongarch64)
 	SCHEME = loongarch
 	endif
 endif
@@ -188,7 +190,7 @@ CARGO_OSDK_COMMON_ARGS += --grub-boot-protocol=$(BOOT_PROTOCOL)
 endif
 
 ifeq ($(ENABLE_KVM), 1)
-	ifeq ($(OSDK_TARGET_ARCH), x86_64)
+	ifeq ($(TARGET_ARCH), x86_64)
 	CARGO_OSDK_COMMON_ARGS += --qemu-args="-accel kvm"
 	endif
 endif
@@ -381,7 +383,7 @@ docs: $(CARGO_OSDK)
 	@# Include --document-private-items to fully check internal documentation.
 	@RUSTDOCFLAGS="-Dwarnings --document-private-items -Arustdoc::private_intra_doc_links" \
 		cargo osdk doc -p aster-kernel --no-deps
-	@if [ "$(OSDK_TARGET_ARCH)" = "x86_64" ]; then \
+	@if [ "$(TARGET_ARCH)" = "x86_64" ]; then \
 		cd ostd/libs/linux-bzimage/setup && RUSTDOCFLAGS="-Dwarnings" cargo osdk doc --no-deps; \
 	fi
 
@@ -415,7 +417,7 @@ check: $(CARGO_OSDK)
 	done
 	@
 	@# Check compilation of the Rust code
-	@OSDK_TARGET_ARCH="$(OSDK_TARGET_ARCH)" ./tools/clippy_check.sh workspace
+	@./tools/clippy_check.sh workspace
 	@
 	@# Check formatting issues of the C code and Nix files (regression tests)
 	@$(MAKE) --no-print-directory -C test/initramfs check
--- a/tools/nixos/build_iso.sh
+++ b/tools/nixos/build_iso.sh
@@ -4,17 +4,22 @@
 
 set -e
 
+TARGET_ARCH=${TARGET_ARCH:-x86_64}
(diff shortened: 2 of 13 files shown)


junyang-zh:
The full nix test failure seems not the problem of this PR.

```
[   185.494] ERROR : Uncaught panic:
assertion failed: record_len >= self.actual_len()
at /__w/asterinas/asterinas/kernel/src/fs/fs_impls/ext2/dir.rs:360
```

tools/clippy_check.sh
```diff
@@ -99,12 +120,12 @@ run_workspace_clippy() {
...
     # It may fail on other architectures.
-    if [[ "${OSDK_TARGET_ARCH:-x86_64}" = "x86_64" ]]; then
+    if [[ "${TARGET_ARCH:-x86_64}" = "x86_64" ]]; then
         echo "Checking ${LINUX_BZIMAGE_SETUP_DIR}"
         (
             cd "$PROJECT_ROOT/$LINUX_BZIMAGE_SETUP_DIR"
-            RUSTFLAGS="-Dwarnings" cargo osdk clippy -- --no-deps
-            RUSTFLAGS="-Dwarnings" cargo osdk clippy --ktests -- --no-deps
+            RUSTFLAGS="-Dwarnings" cargo osdk clippy -- --no-deps --target "$target_triple"
+            RUSTFLAGS="-Dwarnings" cargo osdk clippy --ktests -- --no-deps --target "$target_triple"
```
tatetian: > The first commits reverts `OSDK_TARGET_ARCH` to `TARGET_ARCH`, since there's an OSDK argument `--target` for this purpose. Duplicating the functionality with another environment variable is weird. Letting other scripts unrelated with OSDK use this variable is also weird because of the `OSDK_` prefix.

From the perspective of the Asterinas build system as a whole, using the name `OSDK_TARGET_ARCH` looks weird. But from the perspective of OSDK itself, it is desirable to have both an environment variable and a CLI argument to specify the same configuration. This is quite common among CLI tools: the CLI argument takes priority over the environment variable.

```diff
-            RUSTFLAGS="-Dwarnings" cargo osdk clippy --ktests -- --no-deps
+            RUSTFLAGS="-Dwarnings" cargo osdk clippy --ktests -- --no-deps --target "$target_triple"
```

For example, the proposed change complicates every invocation of `cargo osdk clippy`, which is highly undesirable. The target CPU architecture is most likely to stay the same over a development session or within a Makefile invocation.

So I propose to (1) keep the `OSDK_TARGET_ARCH` env var for OSDK, which has its value, (2) use `TARCH_ARCH` for the remaining of the Asterinas build system (it specifies the build target across the project, beyond the OSDK-managed kernel itself), (3) set `OSDK_TARGET_ARCH` to `TARCH_ARCH` in the root Makefile.

tools/nixos/run.sh
```diff
@@ -4,36 +4,43 @@
...
 # Run a NixOS installation or NixOS ISO installer image built by the root Makefile inside a VM
 #
-# Usage: ./run.sh [nixos | iso]
+# Usage: ./run.sh [nixos | iso] [target-arch]
 
 set -e
 
 usage() {
-    echo "Usage: $0 [nixos | iso]"
+    echo "Usage: $0 [nixos | iso] [target-arch]"
```
tatetian: In the same spirit as my previous comment, use the `TARGET_ARCH` env var directly, which should have been set up by the Makefile, the only user of these helper scripts.

tools/nixos/target_arch.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+target_arch_to_nix_system() {
+    case "${1:-${TARGET_ARCH:-x86_64}}" in
+        x86_64)
+            echo "x86_64-linux"
+            ;;
+        aarch64|arm64)
```
tatetian: Do we really want to allow `TARGET_ARCH` to accept both? I don't think so.

tools/nixos/run.sh
```diff
@@ -4,36 +4,43 @@
...
-if [ "$#" -ne 1 ]; then
+if [ "$#" -ne 2 ]; then
     usage
 fi
 
 MODE=$1
+TARGET_ARCH=$2
 SCRIPT_DIR=$(dirname "$0")
 ASTERINAS_DIR=$(realpath "${SCRIPT_DIR}/../..")
+. "${SCRIPT_DIR}/target_arch.sh"
```
tatetian: Insert a blank line above this line. Setting variables and importing a helper function are two distinct purposes, and thus should not be grouped together.

tools/nixos/build_iso.sh
```diff
@@ -4,17 +4,26 @@
...
 SCRIPT_DIR=$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)
 ASTERINAS_DIR=$(realpath ${SCRIPT_DIR}/../..)
 DISTRO_DIR=$(realpath ${ASTERINAS_DIR}/distro)
 TARGET_DIR=${ASTERINAS_DIR}/target/nixos
 VERSION=$(cat ${ASTERINAS_DIR}/VERSION)
-# Accept config file name as parameter, default to "configuration.nix"
-CONFIG_FILE_NAME=${1:-"configuration.nix"}
+. "${SCRIPT_DIR}/target_arch.sh"
+
+NIX_SYSTEM=$(target_arch_to_nix_system "${TARGET_ARCH}") || {
```
tatetian: Instead of importing the function from `target_arch.sh`, why not make `target_arch.sh` an executable script and directly run this script? All other scripts in this directory are executable, except this newly-added `target_arch.sh`.

Rename this script to `show_target_nix_system.sh` (or `print_target_nix_system.sh`) and directly run this script here.

tatetian (commented):
Thanks for fixing the cross compilation issues.

Makefile
```diff
@@ -415,7 +415,7 @@ check: $(CARGO_OSDK)
 	done
 	@
 	@# Check compilation of the Rust code
-	@OSDK_TARGET_ARCH="$(OSDK_TARGET_ARCH)" ./tools/clippy_check.sh workspace
+	@TARGET_ARCH="$(TARGET_ARCH)" ./tools/clippy_check.sh workspace
```
tatetian: As a wrapper script for OSDK, `clippy_check.sh` is a typical example where target CPU architecture should be governed entirely with `OSDK_TARGET_ARCH`. If even this script used `TARGET_ARCH`, then I am not really sure about the value of `OSDK_TARGET_ARCH`.

Makefile
```diff
@@ -93,7 +93,7 @@ SHELL := /bin/bash
 CARGO_OSDK := ~/.cargo/bin/cargo-osdk
 
 # Common arguments for `cargo osdk` `build`, `run` and `test` commands.
-CARGO_OSDK_COMMON_ARGS := --target-arch=$(OSDK_TARGET_ARCH)
+CARGO_OSDK_COMMON_ARGS := --target-arch=$(TARGET_ARCH)
```
tatetian: Let's make the `OSDK_TARGET_ARCH` env var useful by exposing it in the Makefile 

```Makefile
...

# ========================= End of Makefile options. ==========================

# Specify the target CPU arch for all OSDK commands.
export OSDK_TARGET_ARCH=$(TARGET_ARCH)
```

This brings two benefits: (1) no OSDK commands need to have the `--target-arch` argument set explicitly, and (2) we don't need to a target arch env var to `./tools/clippy_check.sh`.

tatetian (commented):
Thanks for the revision. Just two minor issues.

tatetian:
Could you confirm if "Test AsterNixOS (full)" is caused by this PR or not? Other recent PRs do not seem to witness this failure.

tatetian:
> The full nix test failure seems not the problem of this PR.
> 
> ```
> [   185.494] ERROR : Uncaught panic:
> assertion failed: record_len >= self.actual_len()
> at /__w/asterinas/asterinas/kernel/src/fs/fs_impls/ext2/dir.rs:360
> ```

Yes, this appears to be a kernel bug. The problem is why the changes in this PR can reveal the problem...

> No need. Expected failure.

(_Originally posted by @junyang-zh in https://github.com/asterinas/asterinas/issues/3144#issuecomment-4411586760_)

Why did you say it is _expected_?

junyang-zh:
> Other recent PRs do not seem to witness this failure.

Because they do not run the test at all.

tatetian:
> > Other recent PRs do not seem to witness this failure.
> 
> Because they do not run the test at all.

Thanks for explaining. Then I have no concerns with the CI. Just fix the minor issues and we will be good to go.

tatetian (approved):

PR #3145, opened 2026-04-16
Only calls `on_first_process_startup` when the first user thread starts

StevenJiang1110:
Currently, the execution of `on_first_process_startup` depends on whether the current process is the init process. However, this condition is incorrect — if the init process spawns multiple threads, the startup routine would be executed multiple times. The condition should instead check whether the current thread ID is 1, ensuring that the startup routine is only executed when the very first thread starts up.

### DIFF
--- a/kernel/src/process/exit.rs
+++ b/kernel/src/process/exit.rs
@@ -2,7 +2,7 @@
 
 use core::sync::atomic::Ordering;
 
-use super::{Pid, Process, pid_table};
+use super::{INIT_PROCESS_PID, Pid, Process, pid_table};
 use crate::{
     events::IoEvents, fs::cgroupfs::CgroupMembership, prelude::*,
     process::signal::signals::kernel::KernelSignal,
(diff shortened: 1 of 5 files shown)


lrh2000:
Looks like it is introduced by https://github.com/asterinas/asterinas/commit/fa0ce729665a3e88dde390ed8dc2ccb64fc88361 (#3088)? My fault. Thanks for catching this.

The previous code does not clearly indicate that it wants to rule out other threads in the init process. It actually does rule them out. So it is correct. However, I didn't realize that.

StevenJiang1110:
> Looks like it is introduced by [fa0ce72](https://github.com/asterinas/asterinas/commit/fa0ce729665a3e88dde390ed8dc2ccb64fc88361) (#3088)? 

Yes, it looks like that commit introduced the regression. I also missed it during the review. The original code did not clearly convey the intent. I discovered this issue when running Kata against the main branch — it was encountering errors related to double initialization.

lrh2000 (approved):

PR #2951, opened 2026-02-04
Support `virtio-rng` and expose it as `/dev/hwrng`

li041:
## Changes in this PR
- Add configurations to include the `virtio-rng` device in qemu args.
- Detect `virtio-rng` devices and initialize.
- Expose `virtio-rng` device through `/dev/hwrng`, making it accessible for reading random numbers.
## Unsupported Feature
- Linux supports selecting the `hwrng` backend via `/sys/class/misc/hw_random/rng_current`. This feature was not implemented in this PR due to the need for **_locking_**, which could impact performance. The feature is noted but not currently included in the implementation.

PR #3146, opened 2026-04-16
Fix short TCP stream reads across wrapped receive buffer segments

taosue:
Currently, Asterinas returns a short TCP stream read when the TCP receive ring wrapped around, which is harmful in async edge-triggered code.  After seeing a short read, userspace may clear readiness and go back to waiting for the next event, even though unread TCP payload is still buffered in the socket. If no new packet arrives, the task can sleep indefinitely while data is already available. **This directly caused CoCo guest image pulling to stall on Asterinas.**

Conceptually, the unread data in a TCP stream is continuous:
```text
  ... A A A A B B B B ...
```
 But after the receive ring wraps, the same unread bytes may be stored like this:
```text
  [ B B B B . . . . . . . . . . . . . A A A A ]
          ^ tail of ring              ^ head of ring
```
In this situation, the first contiguous fragment visible from the current read pointer is only the tail fragment (`A A A A`), even though the stream still has more unread bytes after the wrap (`B B B B`).

I.e., a single `read()` could stop at that first contiguous fragment and return only the short tail piece. This is inconsistent with Linux stream socket behavior: a stream read should not expose the kernel receive ring's internal segmentation to userspace.

This PR tries to fix in `ConnectedStream::try_recv()` so that if the current contiguous fragment is fully consumed and the userspace buffer still has remaining space, the same syscall continues reading across the wrap boundary.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
+++ b/kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
@@ -4,7 +4,10 @@ use alloc::{
     boxed::Box,
     sync::{Arc, Weak},
 };
-use core::ops::{Deref, DerefMut};
+use core::{
+    num::NonZeroUsize,
+    ops::{Deref, DerefMut},
+};
 
 use aster_softirq::BottomHalfDisabled;
 use ostd::sync::{SpinLock, SpinLockGuard};
@@ -20,7 +23,7 @@ use super::{
 };
 use crate::{
     define_boolean_value,
-    errors::tcp::{ConnectError, RecvError, SendError},
+    errors::tcp::{ConnectError, IoError, RecvError, SendError},
     ext::Ext,
     iface::{BoundPort, PollKey, PollableIfaceMut},
     socket::{
@@ -370,9 +373,12 @@ impl<E: Ext> TcpConnection<E> {
     /// Sends some data.
     ///
     /// Polling the iface _may_ be required after this method succeeds.
-    pub fn send<F, R>(&self, f: F) -> Result<(R, NeedIfacePoll), SendError>
+    pub fn send<F, CopyErr>(
+        &self,
+        mut f: F,
+    ) -> Result<(NonZeroUsize, NeedIfacePoll), IoError<SendError, CopyErr>>
     where
-        F: FnOnce(&mut [u8]) -> (usize, R),
+        F: FnMut(&mut [u8]) -> Result<usize, (CopyErr, usize)>,
     {
         let common = self.iface().common();
         let mut iface = common.interface();
@@ -381,10 +387,56 @@ impl<E: Ext> TcpConnection<E> {
 
         if socket.is_rst_closed {
             socket.is_rst_closed = false;
-            return Err(SendError::ConnReset);
+            return Err(IoError::Socket(SendError::ConnReset));
         }
-        let result = socket.send(f)?;
 
+        let mut total_sent_bytes = 0;
+        let mut need_wrap_continuation = true;
+
+        let result = loop {
+            let mut sent_bytes = 0;
+            let mut socket_buffer_len = 0;
+
+            // `socket.send` exposes the largest contiguous writable range. If the free space in
+            // the transmit ring buffer wraps, one logical stream write may span two such ranges,
+            // so continue at most once while holding the same socket lock to avoid an avoidable
+            // short write.
+            let send_result = socket.send(|socket_buffer| {
+                socket_buffer_len = socket_buffer.len();
+                match f(socket_buffer) {
+                    Ok(current_sent_bytes) => {
+                        sent_bytes = current_sent_bytes;
+                        (current_sent_bytes, Ok(()))
+                    }
+                    Err((err, current_sent_bytes)) => {
+                        sent_bytes = current_sent_bytes;
+                        (current_sent_bytes, Err(err))
+                    }
+                }
+            });
+
+            let copy_result = match send_result {
+                Err(_) if total_sent_bytes > 0 => break total_sent_bytes,
+                res => res?,
+            };
+
+            total_sent_bytes += sent_bytes;
+
+            if let Err(err) = copy_result {
+                if total_sent_bytes == 0 {
+                    return Err(IoError::Copy(err));
+                }
+                break total_sent_bytes;
+            }
+
+            if sent_bytes < socket_buffer_len || !need_wrap_continuation {
+                break total_sent_bytes;
+            }
+
+            need_wrap_continuation = false;
+        };
+
+        let result = NonZeroUsize::new(result).ok_or(IoError::NoProgress)?;
         let poll_at = socket.poll_at(iface.context_mut());
         let need_poll = iface.update_next_poll_at_ms(&self.0, poll_at);
 
@@ -394,9 +446,12 @@ impl<E: Ext> TcpConnection<E> {
     /// Receives some data.
     ///
     /// Polling the iface _may_ be required after this method succeeds.
-    pub fn recv<F, R>(&self, f: F) -> Result<(R, NeedIfacePoll), RecvError>
+    pub fn recv<F, CopyErr>(
+        &self,
+        mut f: F,
+    ) -> Result<(NonZeroUsize, NeedIfacePoll), IoError<RecvError, CopyErr>>
     where
-        F: FnOnce(&mut [u8]) -> (usize, R),
+        F: FnMut(&mut [u8]) -> Result<usize, (CopyErr, usize)>,
     {
         let common = self.iface().common();
         let mut iface = common.interface();
@@ -404,16 +459,59 @@ impl<E: Ext> TcpConnection<E> {
         let mut socket = self.0.inner.lock();
 
         if socket.is_recv_shut && socket.recv_queue() == 0 {
-            return Err(RecvError::Finished);
+            return Err(IoError::Socket(RecvError::Finished));
         }
-        let result = match socket.recv(f) {
-            Err(_) if socket.is_rst_closed => {
-                socket.is_rst_closed = false;
-                return Err(RecvError::ConnReset);
-            }
-            res => res,
-        }?;
 
+        let mut total_recv_bytes = 0;
+        let mut need_wrap_continuation = true;
+
+        let result = loop {
+            let mut recv_bytes = 0;
+            let mut socket_buffer_len = 0;
+
+            // `socket.recv` exposes the largest contiguous readable range. If the receive ring
+            // buffer wraps, one logical stream read may span two such ranges, so continue at most
+            // once while holding the same socket lock to avoid an avoidable short read.
+            let recv_result = socket.recv(|socket_buffer| {
+                socket_buffer_len = socket_buffer.len();
+                match f(socket_buffer) {
+                    Ok(current_recv_bytes) => {
+                        recv_bytes = current_recv_bytes;
+                        (current_recv_bytes, Ok(()))
+                    }
+                    Err((err, current_recv_bytes)) => {
+                        recv_bytes = current_recv_bytes;
+                        (current_recv_bytes, Err(err))
+                    }
+                }
+            });
+
+            let copy_result = match recv_result {
+                Err(_) if socket.is_rst_closed && total_recv_bytes == 0 => {
+                    socket.is_rst_closed = false;
+                    return Err(IoError::Socket(RecvError::ConnReset));
+                }
+                Err(_) if total_recv_bytes > 0 => break total_recv_bytes,
+                res => res?,
+            };
+
+            total_recv_bytes += recv_bytes;
+
+            if let Err(err) = copy_result {
+                if total_recv_bytes == 0 {
+                    return Err(IoError::Copy(err));
+                }
+                break total_recv_bytes;
+            }
+
+            if recv_bytes < socket_buffer_len || !need_wrap_continuation {
+                break total_recv_bytes;
+            }
+
+            need_wrap_continuation = false;
+        };
+
+        let result = NonZeroUsize::new(result).ok_or(IoError::NoProgress)?;
         let poll_at = socket.poll_at(iface.context_mut());
         let need_poll = iface.update_next_poll_at_ms(&self.0, poll_at);
 
--- a/kernel/src/net/socket/ip/stream/connected.rs
+++ b/kernel/src/net/socket/ip/stream/connected.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use aster_bigtcp::{
-    errors::tcp::{RecvError, SendError},
+    errors::tcp::{IoError, RecvError, SendError},
     socket::{NeedIfacePoll, RawTcpSetOption},
     wire::IpEndpoint,
 };
@@ -74,29 +74,22 @@ impl ConnectedStream {
         writer: &mut dyn MultiWrite,
         _flags: SendRecvFlags,
     ) -> Result<(usize, NeedIfacePoll)> {
-        let result = self.tcp_conn.recv(|socket_buffer| {
-            match writer.write(&mut VmReader::from(&*socket_buffer)) {
-                Ok(len) => (len, Ok(len)),
-                Err(e) => (0, Err(e)),
-            }
-        });
+        let result = self
+            .tcp_conn
+            .recv(|socket_buffer| writer.write(&mut VmReader::from(&*socket_buffer)));
 
         match result {
-            Ok((Ok(0), need_poll)) => {
-                debug_assert!(!*need_poll);
+            Ok((recv_bytes, need_poll)) => Ok((recv_bytes.get(), need_poll)),
+            Err(IoError::NoProgress) => {
                 return_errno_with_message!(Errno::EAGAIN, "the receive buffer is empty")
             }
-            Ok((Ok(recv_bytes), need_poll)) => Ok((recv_bytes, need_poll)),
-            Ok((Err(e), need_poll)) => {
-                debug_assert!(!*need_poll);
-                Err(e)
-            }
-            Err(RecvError::Finished) | Err(RecvError::InvalidState) => {
+            Err(IoError::Copy(e)) => Err(e.into()),
+            Err(IoError::Socket(RecvError::Finished | RecvError::InvalidState)) => {
                 // `InvalidState` occurs when the connection is reset but `ECONNRESET` was reported
                 // earlier. Linux returns EOF in this case, so we follow it.
                 Ok((0, NeedIfacePoll::FALSE))
             }
-            Err(RecvError::ConnReset) => {
+            Err(IoError::Socket(RecvError::ConnReset)) => {
                 return_errno_with_message!(Errno::ECONNRESET, "the connection is reset")
             }
         }
@@ -107,27 +100,20 @@ impl ConnectedStream {
         reader: &mut dyn MultiRead,
         _flags: SendRecvFlags,
     ) -> Result<(usize, NeedIfacePoll)> {
-        let result = self.tcp_conn.send(|socket_buffer| {
-            match reader.read(&mut VmWriter::from(socket_buffer)) {
-                Ok(len) => (len, Ok(len)),
-                Err(e) => (0, Err(e)),
-            }
-        });
+        let result = self
+            .tcp_conn
+            .send(|socket_buffer| reader.read(&mut VmWriter::from(socket_buffer)));
 
         match result {
-            Ok((Ok(0), need_poll)) => {
-                debug_assert!(!*need_poll);
+            Ok((sent_bytes, need_poll)) => Ok((sent_bytes.get(), need_poll)),
+            Err(IoError::NoProgress) => {
                 return_errno_with_message!(Errno::EAGAIN, "the send buffer is full")
             }
-            Ok((Ok(sent_bytes), need_poll)) => Ok((sent_bytes, need_poll)),
-            Ok((Err(e), need_poll)) => {
-                debug_assert!(!*need_poll);
-                Err(e)
-            }
-            Err(SendError::InvalidState) => {
+            Err(IoError::Copy(e)) => Err(e.into()),
+            Err(IoError::Socket(SendError::InvalidState)) => {
                 return_errno_with_message!(Errno::EPIPE, "the connection is closed");
             }
-            Err(SendError::ConnReset) => {
+            Err(IoError::Socket(SendError::ConnReset)) => {
                 return_errno_with_message!(Errno::ECONNRESET, "the connection is reset");
             }
         }
--- a/kernel/src/net/socket/vsock/transport/connection/send.rs
+++ b/kernel/src/net/socket/vsock/transport/connection/send.rs
@@ -111,7 +111,7 @@ impl Connection {
             let packet_mut = packet_opt.as_mut().unwrap();
             let bytes_written = packet_mut.copy_payload(|mut writer| {
                 writer.limit(remaining_bytes);
-                Ok(reader.read(&mut writer)?)
+                reader.read(&mut writer).map_err(|(err, _)| err)
             })?;
             remaining_bytes -= bytes_written;
         }
--- a/kernel/src/util/iovec.rs
+++ b/kernel/src/util/iovec.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use ostd::mm::{Infallible, VmSpace};
+use ostd::{
+    Error as OstdError,
+    mm::{Infallible, VmSpace},
+};
 
 use crate::prelude::*;
 
@@ -179,9 +182,13 @@ pub trait MultiRead: ReadCString {
     ///
     /// # Errors
     ///
-    /// This method returns [`Errno::EFAULT`] if a page fault occurs.
-    /// The position of `self` and the `writer` is left unspecified when this method returns error.
-    fn read(&mut self, writer: &mut VmWriter<'_, Infallible>) -> Result<usize>;
+    /// This method returns [`OstdError::PageFault`] if a page fault occurs, along with
+    /// the number of bytes copied before the error occurs. When an error is returned,
+    /// both `self` and `writer` are advanced by the returned byte count.
+    fn read(
+        &mut self,
+        writer: &mut VmWriter<'_, Infallible>,
+    ) -> core::result::Result<usize, (OstdError, usize)>;
 
     /// Calculates the total length of data remaining to read.
     fn sum_lens(&self) -> usize;
@@ -206,9 +213,13 @@ pub trait MultiWrite {
     ///
     /// # Errors
     ///
-    /// This method returns [`Errno::EFAULT`] if a page fault occurs.
-    /// The position of `self` and the `reader` is left unspecified when this method returns error.
-    fn write(&mut self, reader: &mut VmReader<'_, Infallible>) -> Result<usize>;
+    /// This method returns [`OstdError::PageFault`] if a page fault occurs, along with
+    /// the number of bytes copied before the error occurs. When an error is returned,
+    /// both `self` and `reader` are advanced by the returned byte count.
+    fn write(
+        &mut self,
+        reader: &mut VmReader<'_, Infallible>,
+    ) -> core::result::Result<usize, (OstdError, usize)>;
 
     /// Calculates the length of space available to write.
     fn sum_lens(&self) -> usize;
@@ -224,11 +235,16 @@ pub trait MultiWrite {
 }
 
 impl MultiRead for VmReaderArray<'_> {
-    fn read(&mut self, writer: &mut VmWriter<'_, Infallible>) -> Result<usize> {
+    fn read(
+        &mut self,
+        writer: &mut VmWriter<'_, Infallible>,
+    ) -> core::result::Result<usize, (OstdError, usize)> {
         let mut total_len = 0;
 
         for reader in &mut self.0 {
-            let copied_len = reader.read_fallible(writer)?;
+            let copied_len = reader
+                .read_fallible(writer)
+                .map_err(|(err, copied_len)| (err, total_len + copied_len))?;
             total_len += copied_len;
             if !writer.has_avail() {
                 break;
@@ -255,8 +271,11 @@ impl MultiRead for VmReaderArray<'_> {
 }
 
 impl MultiRead for VmReader<'_> {
-    fn read(&mut self, writer: &mut VmWriter<'_, Infallible>) -> Result<usize> {
-        Ok(self.read_fallible(writer)?)
+    fn read(
+        &mut self,
+        writer: &mut VmWriter<'_, Infallible>,
+    ) -> core::result::Result<usize, (OstdError, usize)> {
+        self.read_fallible(writer)
     }
 
     fn sum_lens(&self) -> usize {
@@ -283,11 +302,16 @@ impl dyn MultiRead + '_ {
 }
 
 impl MultiWrite for VmWriterArray<'_> {
-    fn write(&mut self, reader: &mut VmReader<'_, Infallible>) -> Result<usize> {
+    fn write(
+        &mut self,
+        reader: &mut VmReader<'_, Infallible>,
+    ) -> core::result::Result<usize, (OstdError, usize)> {
         let mut total_len = 0;
 
         for writer in &mut self.0 {
-            let copied_len = writer.write_fallible(reader)?;
+            let copied_len = writer
+                .write_fallible(reader)
+                .map_err(|(err, copied_len)| (err, total_len + copied_len))?;
             total_len += copied_len;
             if !reader.has_remain() {
                 break;
@@ -314,8 +338,11 @@ impl MultiWrite for VmWriterArray<'_> {
 }
 
 impl MultiWrite for VmWriter<'_> {
-    fn write(&mut self, reader: &mut VmReader<'_, Infallible>) -> Result<usize> {
-        Ok(self.write_fallible(reader)?)
+    fn write(
+        &mut self,
+        reader: &mut VmReader<'_, Infallible>,
+    ) -> core::result::Result<usize, (OstdError, usize)> {
+        self.write_fallible(reader)
     }
 
     fn sum_lens(&self) -> usize {
--- a/test/initramfs/src/regression/network/tcp_err.c
+++ b/test/initramfs/src/regression/network/tcp_err.c
@@ -493,7 +493,37 @@ FN_TEST(sendmsg_and_recvmsg)
 	iov[1].iov_len = 1;
 	msg.msg_iov = iov;
 	msg.msg_iovlen = 2;
+
+	char recv_buffer[4096] = { 0 };
+	/*
+	 * FIXME: Linux TCP reports `EFAULT` if a copy fault occurs before
+	 * completing the first SKB touched by the call. Later faults may return
+	 * a short count at SKB granularity. Asterinas has no SKBs, so it reports
+	 * byte-granular progress instead.
+	 *
+	 * Reference:
+	 * <https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L1328-L1329>
+	 * <https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L1441-L1442>
+	 */
+#ifdef __asterinas__
+	/*
+	 * The good bytes are queued in the TCP stream and must be drained
+	 * before the next case.
+	 */
+	TEST_RES(sendmsg(sk_accepted, &msg, 0), _ret == strlen(good_buffer));
+
+	sleep(1);
+
+	iov[0].iov_base = recv_buffer;
+	iov[0].iov_len = sizeof(recv_buffer);
+	msg.msg_iovlen = 1;
+	TEST_RES(recvmsg(sk_connected, &msg, 0),
+		 _ret == strlen(good_buffer) &&
+			 memcmp(recv_buffer, good_buffer,
+				strlen(good_buffer)) == 0);
+#else
 	TEST_ERRNO(sendmsg(sk_accepted, &msg, 0), EFAULT);
+#endif
 
 	// TEST CASE 4: Receive via a partially bad receive buffer
 
@@ -506,7 +536,7 @@ FN_TEST(sendmsg_and_recvmsg)
 
 	sleep(1);
 
-	char recv_buffer[4096] = { 0 };
+	memset(recv_buffer, 0, sizeof(recv_buffer));
 	iov[0].iov_base = recv_buffer;
 	iov[0].iov_len = 1;
 	TEST_RES(recvmsg(sk_connected, &msg, 0), _ret == 1);
@@ -516,13 +546,36 @@ FN_TEST(sendmsg_and_recvmsg)
 	iov[1].iov_base = (char *)1;
 	iov[1].iov_len = 1;
 	msg.msg_iovlen = 2;
-	TEST_ERRNO(recvmsg(sk_connected, &msg, 0), EFAULT);
 
+	/* FIXME: Refer to the above FIXME regarding sending with a bad buffer. */
+#ifdef __asterinas__
+	/*
+	 * The bytes copied are consumed from the TCP stream, so only the
+	 * remaining bytes are left for the next receive.
+	 */
+	TEST_RES(recvmsg(sk_connected, &msg, 0),
+		 _ret == 1 && recv_buffer[0] == good_buffer[1]);
+
+	memset(recv_buffer, 0, sizeof(recv_buffer));
 	iov[0].iov_base = recv_buffer;
-	iov[0].iov_len = 4096;
+	iov[0].iov_len = sizeof(recv_buffer);
 	msg.msg_iovlen = 1;
 	TEST_RES(recvmsg(sk_connected, &msg, 0),
-		 _ret == strlen(good_buffer) - 1);
+		 _ret == strlen(good_buffer) - 2 &&
+			 memcmp(recv_buffer, good_buffer + 2,
+				strlen(good_buffer) - 2) == 0);
+#else
+	TEST_ERRNO(recvmsg(sk_connected, &msg, 0), EFAULT);
+
+	memset(recv_buffer, 0, sizeof(recv_buffer));
+	iov[0].iov_base = recv_buffer;
+	iov[0].iov_len = sizeof(recv_buffer);
+	msg.msg_iovlen = 1;
+	TEST_RES(recvmsg(sk_connected, &msg, 0),
+		 _ret == strlen(good_buffer) - 1 &&
+			 memcmp(recv_buffer, good_buffer + 1,
+				strlen(good_buffer) - 1) == 0);
+#endif
 
 	// TEST CASE 5: Send a large buffer
 
--- a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
+++ b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "../common/test.h"
+
+#define LISTEN_PORT 9999
+#define LISTEN_BACKLOG 4
+#define RECEIVE_BUFFER_FILL_DRAIN_ROUNDS 16
+#define TCP_SETTLE_USEC 100000
+
+/*
+ * Tokio's PollEvented clears readiness after a positive short read on
+ * epoll/kqueue targets because such a read is treated as evidence that the
+ * socket buffer has been drained. If a TCP stream read stops at the tail
+ * fragment of a wrapped receive buffer while bytes remain after the wrap, a
+ * reactor can miss the remaining data and leave the task waiting.
+ *
+ * The same applies to a positive short write, which is not yet covered by
+ * this test.
+ *
+ * See:
+ * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213
+ * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L240-L267
+ *
+ * This test focuses on Asterinas's simple, ring-buffered TCP behavior. On
+ * Linux, there are many other complexities related to SKB management and
+ * receive window advertisement, so this test is likely to fail.
+ *
+ * See:
+ * https://github.com/asterinas/asterinas/pull/3146#issuecomment-4351171818
+ */
+FN_TEST(tcp_read_wrap_receive_buffer_tail)
+{
+	struct sockaddr_in listen_addr = {
+		.sin_family = AF_INET,
+		.sin_addr.s_addr = htonl(INADDR_LOOPBACK),
+		.sin_port = htons(LISTEN_PORT),
+	};
+	int listen_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
+	int client_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
+	int server_fd;
+	int tcp_recv_buf_len = 0;
+	socklen_t optlen = sizeof(tcp_recv_buf_len);
+	size_t one_third_buf_len;
+	size_t two_thirds_buf_len;
+	char *buf;
+
+	TEST_SUCC(bind(listen_fd, (struct sockaddr *)&listen_addr,
+		       sizeof(listen_addr)));
+	TEST_SUCC(listen(listen_fd, LISTEN_BACKLOG));
+	TEST_SUCC(connect(client_fd, (struct sockaddr *)&listen_addr,
+			  sizeof(listen_addr)));
+	server_fd = TEST_SUCC(accept(listen_fd, NULL, NULL));
+
+	TEST_RES(getsockopt(client_fd, SOL_SOCKET, SO_RCVBUF, &tcp_recv_buf_len,
+			    &optlen),
+		 _ret == 0 && tcp_recv_buf_len >= 3);
+	one_third_buf_len = tcp_recv_buf_len / 3;
+	two_thirds_buf_len = one_third_buf_len * 2;
+
+	buf = TEST_RES(malloc(two_thirds_buf_len), _ret != NULL);
+	memset(buf, 'a', two_thirds_buf_len);
+
+	TEST_RES(send(server_fd, buf, two_thirds_buf_len, 0),
+		 _ret == (ssize_t)two_thirds_buf_len);
+	usleep(TCP_SETTLE_USEC);
+	TEST_RES(recv(client_fd, buf, one_third_buf_len, 0),
+		 _ret == (ssize_t)one_third_buf_len);
+	usleep(TCP_SETTLE_USEC);
+
+	for (int i = 0; i < RECEIVE_BUFFER_FILL_DRAIN_ROUNDS; i++) {
+		TEST_RES(send(server_fd, buf, two_thirds_buf_len, 0),
+			 _ret == (ssize_t)two_thirds_buf_len);
+		usleep(TCP_SETTLE_USEC);
+		TEST_RES(recv(client_fd, buf, two_thirds_buf_len, 0),
+			 _ret == (ssize_t)two_thirds_buf_len);
+		usleep(TCP_SETTLE_USEC);
+	}
+
+	TEST_RES(recv(client_fd, buf, two_thirds_buf_len, 0),
+		 _ret == (ssize_t)one_third_buf_len);
+
+	free(buf);
+	TEST_SUCC(close(client_fd));
+	TEST_SUCC(close(server_fd));
+	TEST_SUCC(close(listen_fd));
+}
+END_TEST()
--- a/kernel/libs/aster-bigtcp/src/errors.rs
+++ b/kernel/libs/aster-bigtcp/src/errors.rs
@@ -10,6 +10,22 @@ pub enum BindError {
 }
 
 pub mod tcp {
+    /// An error returned by a TCP stream I/O operation before any byte is transferred.
+    ///
+    /// If some bytes are transferred before a socket or copy error is observed, the
+    /// operation succeeds with the transferred byte count instead.
+    #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+    pub enum IoError<SocketError, CopyError> {
+        /// The operation made no progress.
+        ///
+        /// This usually means the send buffer is full or the receive buffer is empty.
+        NoProgress,
+        /// The underlying TCP socket failed the operation.
+        Socket(SocketError),
+        /// The caller-provided copy function failed.
+        Copy(CopyError),
+    }
+
     /// An error returned by [`TcpListener::new_listen`].
     ///
     /// [`TcpListener::new_listen`]: crate::socket::TcpListener::new_listen
@@ -68,6 +84,12 @@ pub mod tcp {
         }
     }
 
+    impl<CopyError> From<smoltcp::socket::tcp::SendError> for IoError<SendError, CopyError> {
+        fn from(value: smoltcp::socket::tcp::SendError) -> Self {
+            Self::Socket(value.into())
+        }
+    }
+
     /// An error returned by [`TcpConnection::recv`].
     ///
     /// [`TcpConnection::recv`]: crate::socket::TcpConnection::recv
@@ -85,6 +107,12 @@ pub mod tcp {
                 smoltcp::socket::tcp::RecvError::InvalidState => Self::InvalidState,
                 smoltcp::socket::tcp::RecvError::Finished => Self::Finished,
             }
+        }
+    }
+
+    impl<CopyError> From<smoltcp::socket::tcp::RecvError> for IoError<RecvError, CopyError> {
+        fn from(value: smoltcp::socket::tcp::RecvError) -> Self {
+            Self::Socket(value.into())
         }
     }
 }
--- a/kernel/src/net/socket/ip/datagram/bound.rs
+++ b/kernel/src/net/socket/ip/datagram/bound.rs
@@ -58,7 +58,9 @@ impl datagram_common::Bound for BoundDatagram {
         _flags: SendRecvFlags,
     ) -> Result<(usize, Self::Endpoint)> {
         let result = self.bound_socket.recv(|packet, udp_metadata| {
-            let copied_res = writer.write(&mut VmReader::from(packet));
+            let copied_res = writer
+                .write(&mut VmReader::from(packet))
+                .map_err(Into::into);
             let endpoint = udp_metadata.endpoint;
             (copied_res, endpoint)
         });
@@ -91,6 +93,7 @@ impl datagram_common::Bound for BoundDatagram {
                     .inspect_err(|e| {
                         warn!("unexpected UDP packet {e:#?} will be sent");
                     })
+                    .map_err(Into::into)
             });
 
         match result {
--- a/test/initramfs/src/regression/network/run_test.sh
+++ b/test/initramfs/src/regression/network/run_test.sh
@@ -26,6 +26,7 @@ sleep 0.2
 ./tcp_err
 ./tcp_poll
 ./tcp_reuseaddr
+./tcp_wrapped_buffer_io
 ./udp_broadcast
 ./udp_err
 ./unix_datagram_err


kernel/src/net/socket/ip/stream/connected.rs
```diff
@@ -74,30 +79,61 @@ impl ConnectedStream {
...
-            Err(RecvError::Finished) | Err(RecvError::InvalidState) => {
-                // `InvalidState` occurs when the connection is reset but `ECONNRESET` was reported
-                // earlier. Linux returns EOF in this case, so we follow it.
-                Ok((0, NeedIfacePoll::FALSE))
-            }
-            Err(RecvError::ConnReset) => {
-                return_errno_with_message!(Errno::ECONNRESET, "the connection is reset")
+
+            let result = if total_recv_bytes == 0 {
+                self.try_recv_contiguous_segment(writer)
```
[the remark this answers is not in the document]
taosue: Previous code returns `Errno::EAGAIN` if destination buffer is zero-length and some tests validate this behavior. E.g.,
https://github.com/asterinas/asterinas/blob/18b1fb6255456808b5b42867344f7216179a1cc5/test/initramfs/src/regression/network/tcp_err.c#L187

lrh2000:
> If no new packet arrives, the task can sleep indefinitely while data is already available.

What? I don't believe your analysis is right. Could you please show concrete user-space code? I think either the userspace code is buggy or your analysis is wrong.

taosue:
> > If no new packet arrives, the task can sleep indefinitely while data is already available.
> 
> What? I don't believe your analysis is right. Could you please show concrete user-space code? I think either the userspace code is buggy or your analysis is wrong.

The concrete userspace path in our case is:

  1. [CDH](https://github.com/confidential-containers/guest-components/tree/main/confidential-data-hub) waits in `pull_image(...).await`:
     https://github.com/confidential-containers/guest-components/blob/7475b8d00cdb9598205d257680e985e5406a957d/confidential-data-hub/hub/src/hub.rs#L104-L108

  2. `image-rs` pulls layers via `async_pull_layers(...).await`:
     https://github.com/confidential-containers/guest-components/blob/7475b8d00cdb9598205d257680e985e5406a957d/image-rs/src/image.rs#L425-L432

  3. For each layer, `image-rs` calls `pull_blob_stream(...).await`, wraps the returned `layer_stream.stream` with `StreamReader::new(...)`, and passes it to `async_handle_layer(...)`:
     https://github.com/confidential-containers/guest-components/blob/7475b8d00cdb9598205d257680e985e5406a957d/image-rs/src/pull.rs#L131-L145

4. `async_handle_layer(...)` eventually calls `stream_processing(...)`:                                                          
  https://github.com/confidential-containers/guest-components/blob/7475b8d00cdb9598205d257680e985e5406a957d/image-rs/src/pull.rs#L231-L233

5. `stream_processing(...)` calls `async_processing(...).await`. Inside `async_processing(...)`, the layer reader is wrapped with `HashReader`, and then passed to `unpack(...).await`:
  https://github.com/confidential-containers/guest-components/blob/7475b8d00cdb9598205d257680e985e5406a957d/image-rs/src/stream/mod.rs#L101-L108

  6. `unpack(...)` consumes the tar stream through `entries.next().await`:
  https://github.com/confidential-containers/guest-components/blob/7475b8d00cdb9598205d257680e985e5406a957d/image-rs/src/stream/unpack.rs#L211

 The whole path is:
  `pull_blob_stream() -> StreamReader -> stream_processing() -> unpack() -> entries.next().await`

The observed stall is at the final consumption point: `entries.next().await` does not make progress because the underlying async HTTP body reader is not woken again. There is no special CDH-side retry loop or custom read-until-full logic here. I'm not sure the specific underlying implementation, but perhaps we can refer to the edge-triggered `epoll`.

 Linux [`epoll(7)`](https://man7.org/linux/man-pages/man7/epoll.7.html) explicitly documents this requirement. In the `EPOLLET` example, it says that if the reader does not consume the whole input buffer, a later `epoll_wait()` may block indefinitely even though data is still buffered.

Logically, the TCP stream still has more unread bytes, but physically the receive ring is split into two fragments after wraparound. Asterinas could return only the first contiguous fragment from a single stream read instead of continuing across the wrap boundary. Once such  short read happens in an `EPOLLET` async stack, the task can stop making progress for exactly the reason described in `epoll(7)`: the current wakeup did not actually drain all readable data to `EAGAIN`, but if no new packet arrives, there may be no later edge to wake the task again.

 So the claim is not that CDH has a buggy custom loop. The claim is that a kernel short-read bug can break a normal `EPOLLET` async read path, and the concrete path above is where the image layer body is consumed.

lrh2000:
> The whole path is: `pull_blob_stream() -> StreamReader -> stream_processing() -> unpack() -> entries.next().await`

This looks mostly unrelated. The related part is that it uses `tokio::io::StreamReader`. So if there is a problem, it's likely in `StreamReader`.

> The claim is that a kernel short-read bug can break a normal `EPOLLET` async read path

A signal will also cause short read and will break the logic for the exact same reason. If it happens, I believe the userspace is buggy: https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L2791

I don't believe Linux ever guarantees that `recv` will receive all the bytes or exhaust the given buffer. In practice, this is also not the case due to the existence of signals. Therefore, userspace that relies on this fact may need to fix itself.

lrh2000:
https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213

I guess we can continue our discussion based on this.

lrh2000:
> https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213

This is also found by Codex. It is relatively easy, here are the logs:
[rollout-2026-04-17T10-16-57-019d9af1-98db-7ba2-b52c-eaaaf026f4de.jsonl.zip](https://github.com/user-attachments/files/26822309/rollout-2026-04-17T10-16-57-019d9af1-98db-7ba2-b52c-eaaaf026f4de.jsonl.zip)

I only used it for two rounds, or two interactions. It completes the task in under an hour. But it exhausts my 5-hour token limit...

As a result, I am now convinced that your analysis is correct based on the code snippet shown above.

However, I am yet to be convinced that why Linux won't break the logic due to signals? My Codex token has been exhausted due to the previous analysis. So I cannot let it continue.

There are two possibilities:
 1. If Linux won't break that logic, then we need to fix our kernel.
 2. If Linux also breaks that logic, then it's a userspace bug. We should report it to the Tokio people and see how they respond. However, having this workaround as a quick solution may be unavoidable.

taosue:
> > https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213
> 
> This is also found by Codex. It is relatively easy, here are the logs: [rollout-2026-04-17T10-16-57-019d9af1-98db-7ba2-b52c-eaaaf026f4de.jsonl.zip](https://github.com/user-attachments/files/26822309/rollout-2026-04-17T10-16-57-019d9af1-98db-7ba2-b52c-eaaaf026f4de.jsonl.zip)
> 
> I only used it for two rounds, or two interactions. It completes the task in under an hour. But it exhausts my 5-hour token limit...
> 

Thank you for your investigation! Because the user-space software layer is indeed quite complex, the AI's analysis of this problem within my context gradually diverged, and I also lost patience with it...

> However, having this workaround as a quick solution may be unavoidable.

That's great, hopefully this won't hinder our support for CoCo too much. Should I drop this regression test?

kernel/src/net/socket/ip/stream/connected.rs
```diff
@@ -74,30 +79,61 @@ impl ConnectedStream {
...
+            } else {
+                self.try_recv_contiguous_segment_without_consuming_rst(writer)
```
lrh2000, line 93: ## Problem 1

First, calling `TcpConnection::recv` (and its related methods) multiple times is incorrect and introduces unnecessary complexity. This PR tries to introduce `recv_without_consuming_rst` and add a loop in `stream/connected.rs`, but I see no reason for such changes and they are **incorrect**.

Consider two threads that call `recvmsg()` concurrently. We only take a read lock on the `StreamSocket` so that `ConnectedStream::recv` can be called concurrently. Then, `recvmsg()` can return an interleaved buffer to userspace:
```
           buffer        [ACCC....BBBB]
Thread A   recv(5, ..             ^^^^  (we release all exclusive locks after `TcpConnection::recv` returns)
Thread B   recv(1)        ^              => "A"
Thread A   .. recv(5)      ^             => "BBBBC"
```

This means that `recvmsg` is no longer atomic. It returns nonsense data ("BBBBC"), which never appeared in the stream (it contains "BBBBAC" instead).

So why not fix `TcpConnection::recv` instead? Currently, it requires `F` to be an `FnOnce`, but I think we can convert it to an `FnMut` and try to invoke `F` two times in case the ring buffer wraps. Note, however, that we can only do so if the first call consumes up all the data in the first buffer.

## Problem 2

What should we do if we start receiving the second part of the buffer, but it fails due to an unresolvable page fault in the middle? We will end up in an embarrassing situation:

- We should not return `EFAULT`, because some bytes have already been received. The first part of the buffer has been used up, and it is not possible to undo this.
- We cannot safely return only the bytes from the fully received packets, because `writer` would then be left at the wrong position.
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/src/util/iovec.rs#L210-L211
- We also cannot return the bytes from the first part of the buffer plus the partial bytes from the faulting part, because `MultiWrite` does not report how many bytes were written before `EFAULT` occurred.

It is unlikely that this problem will be resolved without first fixing the `MultiRead` and `MultiWrite` trait.

For more information, see also the dicussion at https://github.com/asterinas/asterinas/pull/3122#discussion_r3076956113 and https://github.com/asterinas/asterinas/pull/2951#discussion_r2999868223.

## Problem 3

If you fix `recv`, then you should also fix `send`.

One reason is for consistency. A more serious issue, however, is that the tokio code has the same exact logic. https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L240-L267

test/initramfs/src/regression/network/tcp_wrapped_receive_buffer_read.c
```diff
@@ -0,0 +1,227 @@
...
+	fprintf(stderr, "%s: first_read=%zd second_read=%zd\n", __func__,
+		first_read, second_read);
+
+	TEST_RES(first_read, _ret == (ssize_t)read_request);
+	TEST_RES(second_read, _ret > 0);
+
+	close(epfd);
+	close_fd_if_open(&client_fd);
+	close_fd_if_open(&server_fd);
+	close_fd_if_open(&listen_fd);
```
lrh2000, line 225: For the regression test, please read the discussion at https://github.com/asterinas/asterinas/pull/3109#issuecomment-4229416673. No one will write things like `close_if_open` in **tests**. They are clearly AI-generated and serve no purpose except to create redundancy and make the code harder to read.

Please try to use skills and other measures to avoid these. I tried the skills mentioned in the above comment and they worked for me. You can take a look.

There are several other problems in the test as well. For example,
 1. We should avoid using `CHECK()` in `FN_TEST()`. Use `TEST_SUCC()` instead.
 2. Merge `TEST_RES` to `TEST_SUCC` if possible.
```c
	TEST_RES(read(client_fd, recv_buf, sizeof(recv_buf)),
			      _ret == (ssize_t)read_request);
```
  3. Remove unnecessary `first_read`/`second_read`/`fprintf(first_read, second_read)`/`TEST_RES(first_read)`/`TEST_RES(second_read)`.
  4. Many other things could be simplified. This test is unnecessarily complex.
    - `make_nonblocking` is unnecessary because we can simply pass `SOCK_NONBLOCK`.
    - `send_bytes` is unnecessary after fixing send (it should just be `malloc` + `memset` + `send`). 
    - `drain_exact_bytes` is also unnecessary because you can simply use `recv` on the newly allocated buffer.
    - `init_fds` also makes no sense. Well, almost every function in this file is nonsense.
  5. Let's use macros to define constants in C instead of using `static const`. In fact, a better solution for `TCP_RECV_BUF_LEN` might be to query the kernel using socket options.

lrh2000 (commented):
Based on the two major comments below, I assume that almost every line in this PR needs to be rewritten. Honestly speaking, it would be better to revert all the changes and start from scratch. The current code, including the Rust code and the regression test, is not very helpful.

> A signal will also cause short read and will break the logic for the exact same reason. If it happens, I believe the userspace is buggy: https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L2791

I tried to make a minimal PoC to show the tokio code is buggy.

It seems that the signal path cannot cause short reads. Existing data takes precedence over pending signals, so pending signals cannot cause a `recvmsg()` to return early. https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L2772-L2773

However, out-of-band (OOB) data can. https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L2743-L2751

In practice, out-of-band (OOB) data is almost never used. It's an obsolete TCP feature. But the feature does exist and can be used without extra configurations. Therefore, creating a PoC to demonstrate that the tokio `TcpStream` can get stuck on Linux is possible.

PoC: [Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f7b3cf34b4e17f9a2adc472c8d5351dd)

It shows:
```
tokio read_exact timed out after copying A=4096, B=0
tokio try_read reports WouldBlock
direct libc recv on the same fd read 4096 bytes, first=Some('B')
reproduced: Tokio cleared readiness after an urgent-mark short read while ordinary data remained readable
```

So, the tokio `try_recv` function reports a `WouldBlock` error, but the underlying TCP socket is readable. Then the tokio code is buggy, at least under this scenario. However, given that OOB data is obsolete and has no modern users, I'm not sure if it has a real impact or if tokio wants to fix it.

kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
```diff
@@ -383,7 +386,52 @@ impl<E: Ext> TcpConnection<E> {
             socket.is_rst_closed = false;
             return Err(SendError::ConnReset);
         }
-        let result = socket.send(f)?;
+        let mut total_sent_bytes = 0;
+        let mut allow_wrap_continuation = true;
```
lrh2000: - If you have a blank line before `total_recv_bytes`, you should also insert a blank line before `total_sent_bytes`?
 - Is `need_wrap_continuation` better than `allow_wrap_continuation`?

kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
```diff
@@ -370,9 +370,12 @@ impl<E: Ext> TcpConnection<E> {
     /// Sends some data.
     ///
     /// Polling the iface _may_ be required after this method succeeds.
-    pub fn send<F, R>(&self, f: F) -> Result<(R, NeedIfacePoll), SendError>
+    pub fn send<F, CopyErr>(
+        &self,
+        mut f: F,
+    ) -> Result<(Result<usize, CopyErr>, NeedIfacePoll), SendError>
```
lrh2000: This type is becoming too complex and obscure.

Something like:
```rust
  /// An error returned by a TCP stream I/O operation before any byte is transferred.
  ///
  /// If some bytes are transferred before a socket or copy error is observed, the
  /// operation succeeds with the transferred byte count instead.
  pub enum IoError<SocketError, CopyError> {
      /// The operation made no progress.
      ///
      /// This usually means the send buffer is full or the receive buffer is empty.
      NoProgress,
      /// The underlying TCP socket failed the operation.
      Socket(SocketError),
      /// The caller-provided copy function failed.
      Copy(CopyError),
  }
```

Then using `Result<(NonZeroUsize, NeedIfacePoll), IoError<SendError, CopyError>>` here is better? It's still somewhat complex, but at least it should be clearer now.

As a side note, we do not need to return `NeedIfacePoll` on `IoError::NoProgress`/`IoError::Copy`. We have `debug_assert!` in `connected.rs`.

kernel/src/util/iovec.rs
```diff
@@ -179,9 +179,13 @@ pub trait MultiRead: ReadCString {
     ///
     /// # Errors
     ///
-    /// This method returns [`Errno::EFAULT`] if a page fault occurs.
-    /// The position of `self` and the `writer` is left unspecified when this method returns error.
-    fn read(&mut self, writer: &mut VmWriter<'_, Infallible>) -> Result<usize>;
+    /// This method returns [`Errno::EFAULT`] if a page fault occurs, along with the number of
+    /// bytes copied before the error occurs. The position of `self` and the `writer` is left
+    /// unspecified when this method returns error.
```
lrh2000: > The position of `self` and the `writer` is left unspecified when this method returns error.

This isn't quite ideal? Since we return `(Error, usize)`, I think we need to guarantee the position now. So the caller can handle the short-read situation correctly.

kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
```diff
@@ -394,9 +442,12 @@ impl<E: Ext> TcpConnection<E> {
...
     ///
     /// Polling the iface _may_ be required after this method succeeds.
-    pub fn recv<F, R>(&self, f: F) -> Result<(R, NeedIfacePoll), RecvError>
+    pub fn recv<F, CopyErr>(
+        &self,
+        mut f: F,
+    ) -> Result<(Result<usize, CopyErr>, NeedIfacePoll), RecvError>
     where
-        F: FnOnce(&mut [u8]) -> (usize, R),
+        F: FnMut(&mut [u8]) -> core::result::Result<usize, (CopyErr, usize)>,
```
lrh2000: I think using `Result` instead of `core::result::Result` is okay here.

test/initramfs/src/regression/network/tcp_err.c
```diff
@@ -493,7 +493,27 @@ FN_TEST(sendmsg_and_recvmsg)
 	iov[1].iov_len = 1;
 	msg.msg_iov = iov;
 	msg.msg_iovlen = 2;
+	char recv_buffer[4096] = { 0 };
```
lrh2000, line 497: Add a blank line before this variable definition.

test/initramfs/src/regression/network/tcp_err.c
```diff
@@ -493,7 +493,27 @@ FN_TEST(sendmsg_and_recvmsg)
...
+#ifdef __asterinas__
```
lrh2000, line 508:
```c
	/*
	 * FIXME: Linux TCP reports `EFAULT` if a copy fault occurs before
	 * completing the first SKB touched by the call. Later faults may return
	 * a short count at SKB granularity. Asterinas has no SKBs, so it reports
	 * byte-granular progress instead.
	 */
```
Add a FIXME before `#ifdef` to explain the behavior difference between Linux and Asterinas in a clearer way.

>Asterinas reports the bytes copied before the fault instead of failing
> the whole call with `EFAULT`. Therefore the good bytes are actually
> queued in the TCP stream and must be drained before the next case.

Simplify this comment given that some information is covered by the FIXME above.

---

Do the same for the following `#ifdef` as well.

test/initramfs/src/regression/network/tcp_err.c
```diff
@@ -516,13 +536,34 @@ FN_TEST(sendmsg_and_recvmsg)
...
+	msg.msg_iovlen = 1;
+	TEST_RES(recvmsg(sk_connected, &msg, 0),
+		 _ret == strlen(good_buffer) - 2 &&
+			 recv_buffer[0] == good_buffer[2]);
+#else
 	TEST_ERRNO(recvmsg(sk_connected, &msg, 0), EFAULT);
 
+	memset(recv_buffer, 0, sizeof(recv_buffer));
 	iov[0].iov_base = recv_buffer;
 	iov[0].iov_len = 4096;
```
lrh2000: Covert "4096" in line 550/560 to `sizeof(recv_buffer)`?

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,110 @@
...
+	TEST_SUCC(shutdown(server_fd, SHUT_WR));
+	usleep(300 * 1000);
+
+	epfd = TEST_SUCC(epoll_create1(0));
+	event.events = EPOLLIN | EPOLLET;
+	event.data.fd = client_fd;
+	TEST_SUCC(epoll_ctl(epfd, EPOLL_CTL_ADD, client_fd, &event));
+	TEST_RES(epoll_wait(epfd, &out_event, 1, 1000),
+		 _ret == 1 && out_event.data.fd == client_fd &&
+			 (out_event.events & EPOLLIN));
```
lrh2000: I don't understand why we need epoll for this test. It doesn't seem to help at all.

Let's provide some context by writing comments before the test. Not in the code, which is confusing.

We can include the Tokio reference links that I provided earlier. They are really helpful for understanding why we need the fix and the regression test.

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,110 @@
...
+		 _ret == 0 && tcp_recv_buf_len >
+				      TAIL_UNREAD + TAIL_ROOM_BEFORE_WRAP);
+	initial_send_bytes = tcp_recv_buf_len - TAIL_ROOM_BEFORE_WRAP;
+	initial_drain_bytes = initial_send_bytes - TAIL_UNREAD;
+
+	/*
+	 * Leave `TAIL_UNREAD` bytes near the end of the TCP receive ring, then send
+	 * `WRAP_SEND` more bytes so the queue wraps to the beginning of the ring.
+	 * An EPOLLET reader should still receive `READ_REQUEST` bytes in one read
+	 * instead of stopping at the wrapped tail fragment.
```
lrh2000: I still think the test is overly complex. Why not just:
```rust
TEST_RES(send(), _ret == tcp_recv_buf_len / 3 * 2);
TEST_RES(recv(), _ret == tcp_recv_buf_len / 3);

for (i = 0; i < 16; ++i)
{
    TEST_RES(send(), _ret == tcp_recv_buf_len / 3 * 2);
    TEST_RES(recv(), _ret == tcp_recv_buf_len / 3 * 2);
}

TEST_RES(recv(), _ret == tcp_recv_buf_len / 3);
```

 - This proposal is simpler because we don't need as many magic numbers.
 - Additionally, the proposal is more robust because it exercises more cases than this test does: The test in this PR assumes that the `tcp_recv_buf_len` is the accurate size for the receive buffer, but that may not always be the case. (At least Linux will report one that is far from accurate.)

We also don't need so many `malloc`s. One is enough.

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,110 @@
...
+		 _ret == (ssize_t)initial_drain_bytes);
+	free(drain_buf);
+
+	send_buf = TEST_RES(malloc(WRAP_SEND), _ret != NULL);
+	memset(send_buf, 'b', WRAP_SEND);
+	TEST_RES(send(server_fd, send_buf, WRAP_SEND, 0), _ret == WRAP_SEND);
+	free(send_buf);
+
+	TEST_SUCC(shutdown(server_fd, SHUT_WR));
+	usleep(300 * 1000);
```
lrh2000: > TEST_SUCC(shutdown(server_fd, SHUT_WR));

`shutdown` is unrelated to this test.

> usleep(100 * 1000);
> usleep(300 * 1000);

Unless you have a specific reason to do otherwise, use the same sleep timeout in one test.

lrh2000 (commented):
Thank you for revising the code and addressing the comments and issues I raised.

The new code is in much better shape. There are some points that could be improved.

kernel/src/net/socket/vsock/transport/connection/send.rs
```diff
@@ -111,7 +111,9 @@ impl Connection {
             let packet_mut = packet_opt.as_mut().unwrap();
             let bytes_written = packet_mut.copy_payload(|mut writer| {
                 writer.limit(remaining_bytes);
-                Ok(reader.read(&mut writer)?)
+                reader
+                    .read(&mut writer)
+                    .map_err(|(err, _)| ostd::Error::from(err))
```
lrh2000: Actually, I think you can return the error type from the OSTD in the `MultiRead`/`MultiWrite` methods, so we won't need to convert the error types back and forth.

kernel/src/util/iovec.rs
```diff
@@ -314,8 +336,12 @@ impl MultiWrite for VmWriterArray<'_> {
 }
 
 impl MultiWrite for VmWriter<'_> {
-    fn write(&mut self, reader: &mut VmReader<'_, Infallible>) -> Result<usize> {
-        Ok(self.write_fallible(reader)?)
+    fn write(
+        &mut self,
+        reader: &mut VmReader<'_, Infallible>,
+    ) -> core::result::Result<usize, (Error, usize)> {
+        FallibleVmWrite::write_fallible(self, reader)
```
lrh2000: Be consistent with yourself. You use `self.read_fallible` in `impl MultiRead for VmReader<'_> {`, so I'm not sure why you use `FallibleVmWrite::write_fallible` instead of `self.write_fallible` here?

test/initramfs/src/regression/network/tcp_err.c
```diff
@@ -516,13 +542,33 @@ FN_TEST(sendmsg_and_recvmsg)
 	iov[1].iov_base = (char *)1;
 	iov[1].iov_len = 1;
 	msg.msg_iovlen = 2;
+#ifdef __asterinas__
```
lrh2000, line 551:
```rust
/* FIXME: Refer to the above FIXME regarding sending with a bad buffer. */
```

test/initramfs/src/regression/network/tcp_err.c
```diff
@@ -493,7 +493,33 @@ FN_TEST(sendmsg_and_recvmsg)
...
 	msg.msg_iov = iov;
 	msg.msg_iovlen = 2;
+
+	char recv_buffer[4096] = { 0 };
+	/*
+	 * FIXME: Linux TCP reports `EFAULT` if a copy fault occurs before
+	 * completing the first SKB touched by the call. Later faults may return
+	 * a short count at SKB granularity. Asterinas has no SKBs, so it reports
+	 * byte-granular progress instead.
+	 */
```
lrh2000, line 507:
```suggestion
	 *
	 * Reference:
	 * <https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L1328-L1329>
	 * <https://elixir.bootlin.com/linux/v7.0/source/net/ipv4/tcp.c#L1441-L1442>
	 */

File: test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c [hunk 18]
```diff
@@ -0,0 +1,110 @@
...
+	size_t two_thirds_buf_len;
+	char *buf;
+
+	TEST_SUCC(setsockopt(listen_fd, SOL_SOCKET, SO_REUSEADDR, &one,
+			     sizeof(one)));
+	TEST_SUCC(bind(listen_fd, (struct sockaddr *)&listen_addr,
+		       sizeof(listen_addr)));
+	TEST_SUCC(listen(listen_fd, LISTEN_BACKLOG));
+	TEST_SUCC(getsockname(listen_fd, (struct sockaddr *)&listen_addr,
+			      &addr_len));
```
lrh2000: This can be simplified:
```diff
diff --git a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
index c5d8ba230..9a5e06d98 100644
--- a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
+++ b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
@@ -10,6 +10,7 @@
 
 #include "../common/test.h"
 
+#define LISTEN_PORT 9999
 #define LISTEN_BACKLOG 4
 #define RECEIVE_BUFFER_FILL_DRAIN_ROUNDS 16
 
@@ -30,10 +31,9 @@ FN_TEST(read_should_not_stop_at_wrapped_tail_fragment)
        struct sockaddr_in listen_addr = {
                .sin_family = AF_INET,
                .sin_addr.s_addr = htonl(INADDR_LOOPBACK),
-               .sin_port = htons(0),
+               .sin_port = htons(LISTEN_PORT),
        };
        socklen_t addr_len = sizeof(listen_addr);
-       int one = 1;
        int listen_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
        int client_fd;
        int server_fd;
@@ -43,13 +43,9 @@ FN_TEST(read_should_not_stop_at_wrapped_tail_fragment)
        size_t two_thirds_buf_len;
        char *buf;
 
-       TEST_SUCC(setsockopt(listen_fd, SOL_SOCKET, SO_REUSEADDR, &one,
-                            sizeof(one)));
        TEST_SUCC(bind(listen_fd, (struct sockaddr *)&listen_addr,
                       sizeof(listen_addr)));
        TEST_SUCC(listen(listen_fd, LISTEN_BACKLOG));
-       TEST_SUCC(getsockname(listen_fd, (struct sockaddr *)&listen_addr,
-                             &addr_len));
 
        client_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
        TEST_SUCC(connect(client_fd, (struct sockaddr *)&listen_addr,
```

I ran the test locally after making my modifications and did not see any issues.

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,110 @@
...
+
+	for (int i = 0; i < RECEIVE_BUFFER_FILL_DRAIN_ROUNDS; i++) {
+		TEST_RES(send_exact_bytes(server_fd, buf, two_thirds_buf_len),
+			 _ret == (ssize_t)two_thirds_buf_len);
+		TEST_RES(recv(client_fd, buf, two_thirds_buf_len, MSG_WAITALL),
+			 _ret == (ssize_t)two_thirds_buf_len);
+	}
+
+	TEST_RES(recv(client_fd, buf, one_third_buf_len, MSG_WAITALL),
+		 _ret == (ssize_t)one_third_buf_len);
```
lrh2000: Do not pass `MSG_WAITALL` to `recv` calls.
 1. This changes the semantics of the `recv` calls, which is not what we want to test.
 2. The affected use case (tokio code) does not supply the `MSG_WAITALL` flag.
 3. Asterinas does not support this flag yet.

You may want to add some `usleep()` timeouts to ensure that the both endpoint sees all the packets sent/ACKed by the remote endpoint.

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,110 @@
...
+	TEST_SUCC(close(server_fd));
+	TEST_SUCC(close(listen_fd));
+}
+END_TEST()
+
+/*
+ * Sends exactly `len` bytes unless `send` reports an error or no progress.
+ *
+ * A TCP stream write first queues data in the local TCP send buffer. If earlier
+ * data is still waiting for ACKs, the send buffer may not have enough free space
```
lrh2000: Just add some `usleep()` to prevent this case from happening?

This PR fixes the send behavior and the receive buffer. As I mentioned earlier, tokio also treats a partial send when the send buffer is full. Therefore, this test must ensure that both the send and receive operations complete within one call.

lrh2000 (commented):
Thanks for revising the code! Most of it looks perfect. Here are some follow-up comments that should be easy to address.

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,86 @@
...
+ * Tokio's PollEvented clears readiness after a positive short read on
+ * epoll/kqueue targets because such a read is treated as evidence that the
+ * socket buffer has been drained. If a TCP stream read stops at the tail
+ * fragment of a wrapped receive buffer while bytes remain after the wrap, a
+ * reactor can miss the remaining data and leave the task waiting.
+ *
+ * See:
+ * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213
+ */
+FN_TEST(read_should_not_stop_at_wrapped_tail_fragment)
```
lrh2000: We should adjust the test names and the comments to properly reflect https://github.com/asterinas/asterinas/pull/3146#discussion_r3165348339:

Something like:
```patch
diff --git a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
index bfa3f4dd8..e61edf11e 100644
--- a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
+++ b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
@@ -22,10 +22,13 @@
  * fragment of a wrapped receive buffer while bytes remain after the wrap, a
  * reactor can miss the remaining data and leave the task waiting.
  *
+ * The same applies to a positive short write.
+ *
  * See:
  * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213
+ * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L240-L267
  */
-FN_TEST(read_should_not_stop_at_wrapped_tail_fragment)
+FN_TEST(tcp_read_write_wrap_buffer_tail_fragment)
 {
        struct sockaddr_in listen_addr = {
                .sin_family = AF_INET,
```

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,86 @@
...
+	socklen_t optlen = sizeof(tcp_recv_buf_len);
+	size_t one_third_buf_len;
+	size_t two_thirds_buf_len;
+	char *buf;
+
+	TEST_SUCC(bind(listen_fd, (struct sockaddr *)&listen_addr,
+		       sizeof(listen_addr)));
+	TEST_SUCC(listen(listen_fd, LISTEN_BACKLOG));
+
+	client_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
```
lrh2000: We can just do:
```c
	int listen_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
	int client_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
```

test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
```diff
@@ -0,0 +1,86 @@
...
+	for (int i = 0; i < RECEIVE_BUFFER_FILL_DRAIN_ROUNDS; i++) {
+		TEST_RES(send(server_fd, buf, two_thirds_buf_len, 0),
+			 _ret == (ssize_t)two_thirds_buf_len);
+		usleep(TCP_SETTLE_USEC);
+		TEST_RES(recv(client_fd, buf, two_thirds_buf_len, 0),
+			 _ret == (ssize_t)two_thirds_buf_len);
+		usleep(TCP_SETTLE_USEC);
+	}
+
+	TEST_RES(recv(client_fd, buf, one_third_buf_len, 0),
```
lrh2000:
```suggestion
	TEST_RES(recv(client_fd, buf, two_third_buf_len, 0),
```


We can receive two-thirds here. However, the result should be one third, as we only have one third remaining.

test/initramfs/src/regression/network/tcp_err.c
```diff
@@ -516,13 +546,35 @@ FN_TEST(sendmsg_and_recvmsg)
...
+	TEST_RES(recvmsg(sk_connected, &msg, 0),
+		 _ret == 1 && recv_buffer[0] == good_buffer[1]);
+
+	memset(recv_buffer, 0, sizeof(recv_buffer));
+	iov[0].iov_base = recv_buffer;
+	iov[0].iov_len = sizeof(recv_buffer);
+	msg.msg_iovlen = 1;
+	TEST_RES(recvmsg(sk_connected, &msg, 0),
+		 _ret == strlen(good_buffer) - 2 &&
+			 recv_buffer[0] == good_buffer[2]);
```
lrh2000: I think we should use `memcmp(recv_buffer, good_buffer + 2, strlen(good_buffer) - 2) == 0);` here.

kernel/src/util/iovec.rs
```diff
@@ -206,9 +213,13 @@ pub trait MultiWrite {
     ///
     /// # Errors
     ///
-    /// This method returns [`Errno::EFAULT`] if a page fault occurs.
-    /// The position of `self` and the `reader` is left unspecified when this method returns error.
-    fn write(&mut self, reader: &mut VmReader<'_, Infallible>) -> Result<usize>;
+    /// This method returns [`Errno::EFAULT`] if a page fault occurs, along with the number of
```
lrh2000: `Errno::EFAULT` is not accurate if we are using the OSTD's error type.

lrh2000 (approved):

lrh2000:
Hmmm... Sorry, but Linux still fails the test for some unknown reason.
```
test_tcp_read_write_wrap_buffer_tail_fragment: `send(server_fd, buf, two_thirds_buf_len, 0)` passed [got Success]
test_tcp_read_write_wrap_buffer_tail_fragment: `recv(client_fd, buf, one_third_buf_len, 0)` passed [got Success]
test_tcp_read_write_wrap_buffer_tail_fragment: `send(server_fd, buf, two_thirds_buf_len, 0)` passed [got Success]
test_tcp_read_write_wrap_buffer_tail_fragment: `recv(client_fd, buf, two_thirds_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)two_thirds_buf_len` is false]
```

It seems to return one-third of the buffer length. But that shouldn't happen. I'm not sure what I'm missing.

taosue:
> It seems to return one-third of the buffer length. But that shouldn't happen. I'm not sure what I'm missing.

That's really interesting... But right now our servers are all down, I'll probably have to wait and debug it on my own computer later.

lrh2000:
I suspect the issue is related to some Linux TCP internals. For example, it may delay ACKs, which prevents the other endpoint from being aware of the peer's available buffer in time. With `TEST_SUCC(setsockopt(client_fd, IPPROTO_TCP, TCP_QUICKACK, &one, sizeof(one)))`, the original error disappears, but other errors appear later.

Let me ask Codex to see if it has some clues..

cqs21 (approved):

lrh2000:
> Let me ask Codex to see if it has some clues..

The Codex does identify the root cause. However, it fails to provide any valid solutions, even when prompted.

After spending some time trying by hand, I came up with the following patch:
```patch
diff --git a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
index 594878d64..c07cc93e2 100644
--- a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
+++ b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <linux/tcp.h>
 #include <unistd.h>
 
 #include "../common/test.h"
@@ -22,13 +23,14 @@
  * fragment of a wrapped receive buffer while bytes remain after the wrap, a
  * reactor can miss the remaining data and leave the task waiting.
  *
- * The same applies to a positive short write.
+ * The same applies to a positive short write, which is not yet covered by
+ * this test.
  *
  * See:
  * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213
  * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L240-L267
  */
-FN_TEST(tcp_read_write_wrap_buffer_tail_fragment)
+FN_TEST(tcp_read_wrap_receive_buffer_tail)
 {
 	struct sockaddr_in listen_addr = {
 		.sin_family = AF_INET,
@@ -38,46 +40,65 @@ FN_TEST(tcp_read_write_wrap_buffer_tail_fragment)
 	int listen_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
 	int client_fd = TEST_SUCC(socket(AF_INET, SOCK_STREAM, 0));
 	int server_fd;
-	int tcp_recv_buf_len = 0;
+	int maxseg = 1024;
+	int tcp_recv_buf_len;
 	socklen_t optlen = sizeof(tcp_recv_buf_len);
-	size_t one_third_buf_len;
-	size_t two_thirds_buf_len;
+	size_t one_fifth_buf_len;
+	size_t two_fifths_buf_len;
 	char *buf;
 
 	TEST_SUCC(bind(listen_fd, (struct sockaddr *)&listen_addr,
 		       sizeof(listen_addr)));
 	TEST_SUCC(listen(listen_fd, LISTEN_BACKLOG));
+
+	/*
+	 * Linux accounts receive-buffer pressure by each SKB's allocated size,
+	 * not just by unread payload bytes. If a read consumes only part of an
+	 * SKB, the SKB remains queued and its allocation stays charged. Linux
+	 * may therefore advertise no new receive window even though payload
+	 * bytes were consumed, so the next send() can stay queued on the sender
+	 * instead of reaching the receiver before the measured recv().
+	 *
+	 * Loopback normally uses a large MSS, which makes this effect easy to
+	 * hit. Set a small TCP_MAXSEG before connecting so Linux builds smaller
+	 * receive SKBs and can advertise enough window after every recv().
+	 *
+	 * Asterinas does not have SKB-based receive-buffer accounting.
+	 * TCP_MAXSEG is currently unsupported there and is a no-op.
+	 */
+	TEST_SUCC(setsockopt(client_fd, IPPROTO_TCP, TCP_MAXSEG, &maxseg, sizeof(maxseg)));
+
 	TEST_SUCC(connect(client_fd, (struct sockaddr *)&listen_addr,
 			  sizeof(listen_addr)));
 	server_fd = TEST_SUCC(accept(listen_fd, NULL, NULL));
 
 	TEST_RES(getsockopt(client_fd, SOL_SOCKET, SO_RCVBUF, &tcp_recv_buf_len,
 			    &optlen),
-		 _ret == 0 && tcp_recv_buf_len >= 3);
-	one_third_buf_len = tcp_recv_buf_len / 3;
-	two_thirds_buf_len = one_third_buf_len * 2;
+		 _ret == 0 && tcp_recv_buf_len >= 5);
+	one_fifth_buf_len = tcp_recv_buf_len / 5;
+	two_fifths_buf_len = one_fifth_buf_len * 2;
 
-	buf = TEST_RES(malloc(two_thirds_buf_len), _ret != NULL);
-	memset(buf, 'a', two_thirds_buf_len);
+	buf = TEST_RES(malloc(two_fifths_buf_len), _ret != NULL);
+	memset(buf, 'a', two_fifths_buf_len);
 
-	TEST_RES(send(server_fd, buf, two_thirds_buf_len, 0),
-		 _ret == (ssize_t)two_thirds_buf_len);
+	TEST_RES(send(server_fd, buf, two_fifths_buf_len, 0),
+		 _ret == (ssize_t)two_fifths_buf_len);
 	usleep(TCP_SETTLE_USEC);
-	TEST_RES(recv(client_fd, buf, one_third_buf_len, 0),
-		 _ret == (ssize_t)one_third_buf_len);
+	TEST_RES(recv(client_fd, buf, one_fifth_buf_len, 0),
+		 _ret == (ssize_t)one_fifth_buf_len);
 	usleep(TCP_SETTLE_USEC);
 
 	for (int i = 0; i < RECEIVE_BUFFER_FILL_DRAIN_ROUNDS; i++) {
-		TEST_RES(send(server_fd, buf, two_thirds_buf_len, 0),
-			 _ret == (ssize_t)two_thirds_buf_len);
+		TEST_RES(send(server_fd, buf, two_fifths_buf_len, 0),
+			 _ret == (ssize_t)two_fifths_buf_len);
 		usleep(TCP_SETTLE_USEC);
-		TEST_RES(recv(client_fd, buf, two_thirds_buf_len, 0),
-			 _ret == (ssize_t)two_thirds_buf_len);
+		TEST_RES(recv(client_fd, buf, two_fifths_buf_len, 0),
+			 _ret == (ssize_t)two_fifths_buf_len);
 		usleep(TCP_SETTLE_USEC);
 	}
 
-	TEST_RES(recv(client_fd, buf, two_thirds_buf_len, 0),
-		 _ret == (ssize_t)one_third_buf_len);
+	TEST_RES(recv(client_fd, buf, two_fifths_buf_len, 0),
+		 _ret == (ssize_t)one_fifth_buf_len);
 
 	free(buf);
 	TEST_SUCC(close(client_fd));
```

 - See the comment in the patch for the root cause.
 - To resolve the issue, the following changes are needed:
   - Set TCP_MAXSEG to a smaller value, as explained in the comment in the patch.
   - Use a smaller buffer each round. Otherwise, it is too easy to hit the zero receive window if the receiver does not advertise a tight receive window in time. There are many possible reasons for this, such as the allocation not being freed as the exact bytes consumed.
 - I also noted that my previous review commit was incorrect. This test may not be able to determine if `send()` can put bytes into two wrapped buffer fragments. Testing that case is even more challenging since we need to fill the receive buffer first. So I simply note this fact.

@taosue Please help verify that this also passes in your Linux machine and in Asterinas after this PR? Thanks!

taosue:
> @taosue Please help verify that this also passes in your Linux machine and in Asterinas after this PR? Thanks!

Thanks for the investigation. I learned a lot from your explanation.

I tested this on my local Linux machine and it still does not pass, although it passes on Asterinas.
```
test_tcp_read_wrap_receive_buffer_tail: `socket(2, SOCK_STREAM, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `socket(2, SOCK_STREAM, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `bind(listen_fd, (struct sockaddr *)&listen_addr, sizeof(listen_addr))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `listen(listen_fd, 4)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `setsockopt(client_fd, IPPROTO_TCP, 2, &maxseg, sizeof(maxseg))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `connect(client_fd, (struct sockaddr *)&listen_addr, sizeof(listen_addr))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `accept(listen_fd, ((void *)0), ((void *)0))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `getsockopt(client_fd, 1, 8, &tcp_recv_buf_len, &optlen)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `malloc(two_fifths_buf_len)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, one_fifth_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)two_fifths_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)two_fifths_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)two_fifths_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)two_fifths_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)one_fifth_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `close(client_fd)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `close(server_fd)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `close(listen_fd)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail summary: 42 tests passed, 5 tests failed
```

The number of failures seems related to the sleep duration. But even after increasing the delay to 3 seconds, I can still observe two failures.
```
test_tcp_read_wrap_receive_buffer_tail: `socket(2, SOCK_STREAM, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `socket(2, SOCK_STREAM, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `bind(listen_fd, (struct sockaddr *)&listen_addr, sizeof(listen_addr))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `listen(listen_fd, 4)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `setsockopt(client_fd, IPPROTO_TCP, 2, &maxseg, sizeof(maxseg))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `connect(client_fd, (struct sockaddr *)&listen_addr, sizeof(listen_addr))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `accept(listen_fd, ((void *)0), ((void *)0))` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `getsockopt(client_fd, 1, 8, &tcp_recv_buf_len, &optlen)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `malloc(two_fifths_buf_len)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, one_fifth_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)two_fifths_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `send(server_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `recv(client_fd, buf, two_fifths_buf_len, 0)` failed [got Success, but `_ret == (ssize_t)one_fifth_buf_len` is false]
test_tcp_read_wrap_receive_buffer_tail: `close(client_fd)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `close(server_fd)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail: `close(listen_fd)` passed [got Success]
test_tcp_read_wrap_receive_buffer_tail summary: 45 tests passed, 2 tests failed
```

My current understanding is that a successful `send()` only means the data has been accepted into the sender-side TCP path. It does not necessarily mean that all bytes are immediately visible to the receiver-side `recv()` call. There are still several steps in between, including segmentation, loopback delivery, receive queue insertion, ACK/window updates, and Linux’s SKB-based receive-buffer accounting.

Given that this test is specifically targeting Asterinas’s wrapped receive-buffer behavior, maybe we can make it an Asterinas-specific regression test?

lrh2000:
> Given that this test is specifically targeting Asterinas’s wrapped receive-buffer behavior, maybe we can make it an Asterinas-specific regression test?

This is also a choice. There are indeed quite some complexities so let's note this discussion:
```patch
diff --git a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
index 594878d64..5cce8c5ba 100644
--- a/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
+++ b/test/initramfs/src/regression/network/tcp_wrapped_buffer_io.c
@@ -22,13 +22,21 @@
  * fragment of a wrapped receive buffer while bytes remain after the wrap, a
  * reactor can miss the remaining data and leave the task waiting.
  *
- * The same applies to a positive short write.
+ * The same applies to a positive short write, which is not yet covered by
+ * this test.
  *
  * See:
  * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L182-L213
  * https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/io/poll_evented.rs#L240-L267
+ *
+ * This test focuses on Asterinas's simple, ring-buffered TCP behavior. On
+ * Linux, there are many other complexities related to SKB management and
+ * receive window advertisement, so this test is likely to fail.
+ *
+ * See:
+ * https://github.com/asterinas/asterinas/pull/3146#issuecomment-4351171818
  */
-FN_TEST(tcp_read_write_wrap_buffer_tail_fragment)
+FN_TEST(tcp_read_wrap_receive_buffer_tail)
 {
        struct sockaddr_in listen_addr = {
                .sin_family = AF_INET,
```

> I tested this on my local Linux machine and it still does not pass, although it passes on Asterinas.

What version of the Linux kernel are you using? I am using a recent version of Linux (6.19).

> But even after increasing the delay to 3 seconds, I can still observe two failures.

I was wondering if it is caused by delayed ACKs or some other "delayed" mechanisms. However, that is unlikely given this claim (three seconds is a very long time; all reasonable timers should have timed out by then). So I don't have many clues as to why it happens at this moment...

taosue:
> What version of the Linux kernel are you using? I am using a recent version of Linux (6.19).
> 
I'm using Linux 6.16.

lrh2000:
> I'm using Linux 6.16.

Thanks for the information. Codex attributes our observation to https://github.com/torvalds/linux/commit/aa251c84636c326471ca9d53723816ba8fffe2bf

<details>
<summary> Analysis (in Chinese) </summary>

• 看起来是 Linux 6.16 和 6.19 之间的 TCP receive-buffer 行为差异，不是 recv() 那条“queued bytes 足够时尽量读满”的结论变了。

  关键点在“recv 前是否真的 queued 了足够 payload”。6.16 这类 workload 下更容易被 SKB truesize 和 receive-buffer autotuning 卡住：payload 读掉一部分以后，部分消
  费的 SKB 还留在 sk_receive_queue，它的 truesize 仍然计入 sk_rmem_alloc，直到整个 SKB 被 tcp_eat_recv_skb() 释放时才由 sock_rfree() 扣掉。所以 send() 成功只表
  示写进了 server 端发送缓冲，不表示这些 bytes 已经进了 client receive queue。

  版本差异里最可疑的是这些 TCP rcvbuf 改动：

  - v6.16 的 tcp_rcvbuf_grow() 还是比较保守的 rcvq_space.space << 1 模型；tcp_rcv_space_adjust() 又会用 copied -= inq，而你的测试每轮都故意留下 one_fifth，会让
    grow 更慢。
  - 6.19 之后包含 aa251c84636c ("tcp: fix too slow tcp_rcvbuf_grow() action")，commit message 明确说之前会让 sender 受 RWIN 限制。
  - 6.19 还包含 f017c1f768b ("tcp: use skb->len instead of skb->truesize in tcp_can_ingest()")，说明这段时间 upstream 正在修“小 MSS 但 skb->truesize 很大导致
    receiver drop/限窗”的问题。

  所以 6.16 下即使 sleep 到 3 秒也可能失败：不是 ACK 没等到，而是 receive queue/window/memory accounting 的状态没有保证下一轮 recv() 前已有 two_fifths payload。
  失败一次后，最后那个期望 one_fifth 的 recv 也会跟着错位，通常会读到比 one_fifth 更多的残留数据。

  源码入口：

  - v6.16 tcp_rcvbuf_grow / tcp_try_rmem_schedule: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/plain/net/ipv4/tcp_input.c?h=v6.16
  - v6.19 tcp_can_ingest / 新 tcp_rcvbuf_grow: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/plain/net/ipv4/tcp_input.c?h=v6.19
  - aa251c84636c: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/patch/?id=aa251c84636c326471ca9d53723816ba8fffe2bf
  - f017c1f768b: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/patch/?id=f017c1f768b670bced4464476655b27dfb937e67

  要确认的话，让 6.16 那边跑一次 strace -e sendto,recvfrom,getsockopt,setsockopt 看失败的实际 _ret，再用 ss -tinm sport = :9999 or dport = :9999 抓失败前后的
  skmem；我预期会看到 rmem_alloc 接近/超过 rcvbuf，而不是单纯睡眠不足。

• tcp_rcvbuf_grow() 是 Linux TCP receive-buffer autotuning 的一部分。它不负责 recv() 拷贝数据，也不负责把 skb 放进 receive queue；它负责在应用读得比较快时，动态
  扩大接收端的 sk_rcvbuf 和可通告窗口上限，让 sender 后续能继续把数据送进来。

  简化成一句话：

  应用最近一段时间读了多少 payload
  => Linux 估算下一 RTT 应该允许 sender 发多少 payload
  => 换算成需要多少 receive-buffer memory
  => 必要时增大 sk_rcvbuf / window_clamp

  关键字段是这些：

  - sk->sk_rcvbuf：TCP receive queue 的内存预算，不是纯 payload bytes。
  - sk->sk_rmem_alloc：当前 receive queue 已经占用的内存，按 skb truesize 算。
  - tp->window_clamp：接收端最多愿意向 peer 通告多大的 receive window。
  - tp->rcv_ssthresh：receive window 的 slow-start 限制。
  - tp->rcvq_space.space：Linux 对“应用每 RTT 能消耗多少 receive data”的估计，不是当前 queue 长度。

  它的上游调用点是 tcp_rcv_space_adjust()。应用 recv() 以后，Linux 会算：

  copied = tp->copied_seq - tp->rcvq_space.seq;
  inq = tp->rcv_nxt - tp->copied_seq;
  copied -= inq;

  意思是：这段时间应用向前读了多少，但如果 receive queue 里还有 backlog，就把 backlog 扣掉，避免过度估计应用消费速度。你的测试每轮故意留下 one_fifth 在 receive
  queue 里，所以这个 inq 长期非零，autotuning 会比较保守。

  v6.16 的 tcp_rcvbuf_grow() 核心逻辑很简单：

  rcvwin = tp->rcvq_space.space << 1;
  rcvbuf = min(tcp_space_from_win(sk, rcvwin), tcp_rmem[2]);
  if (rcvbuf > sk->sk_rcvbuf) {
      sk->sk_rcvbuf = rcvbuf;
      tp->window_clamp = tcp_win_from_space(sk, rcvbuf);
  }

  也就是：目标窗口大概是“最近测到的应用消费能力”的 2 倍，然后把 payload window 换算成内存预算。

  这里为什么要换算？因为 TCP window 是 payload bytes，但 Linux receive buffer 账本算的是 skb->truesize。include/net/tcp.h 里的 tcp_space() 大概是：

  tcp_win_from_space(sk, sk_rcvbuf - backlog - sk_rmem_alloc)

  所以如果 52KB payload 被拆成很多小 skb，sk_rmem_alloc 可能远大于 52KB。应用只读掉 payload 的一部分时，没被完全读完的 skb 还在队列里，truesize 也还占着预算。

  和你的测试的关系是：

  1. send(server_fd, two_fifths) 成功，只说明 server 端把数据放进了发送路径。
  2. client 端是否已经真的收进 receive queue，取决于它之前通告的 window，以及 sk_rcvbuf/sk_rmem_alloc 是否允许继续收。
  3. tcp_rcvbuf_grow() 如果没把 sk_rcvbuf/window_clamp 扩到足够大，sender 可能被 receive window 卡住。
  4. 于是下一次 recv(client_fd, two_fifths) 时，client receive queue 里未必已经有 two_fifths，就会短读。

  v6.19 这块改过。aa251c84636c ("tcp: fix too slow tcp_rcvbuf_grow() action") 明确就是修 tcp_rcvbuf_grow() 增长太慢导致 sender 被 RWIN 限制的问题。v6.19 的版本
  不只是 space * 2，还根据 old/new drain rate 加额外 grow。另一个相关 commit f017c1f768b 又把部分 receive admission 从 skb->truesize 放宽到 skb->len，进一步降
  低“小 payload 但 truesize 大”导致收不进去的概率。

  源码：

  - v6.16 tcp_rcvbuf_grow() / tcp_rcv_space_adjust()：https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/plain/net/ipv4/tcp_input.c?h=v6.16
  - v6.19 对应代码：https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/plain/net/ipv4/tcp_input.c?h=v6.19
  - aa251c84636c patch：https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/patch/?id=aa251c84636c326471ca9d53723816ba8fffe2bf

</details>


Anyway, I think the current version is acceptable, given the many complexities of Linux's TCP implementation. So I'm going to merge this. Thank you for contributing to all of the discussions in this PR.

PR #3147, opened 2026-04-16
Fix error code for CLOCK_THREAD_CPUTIME_ID in clock_nanosleep

cqs21:
This PR fixes `clock_nanosleep` to return `EOPNOTSUPP` for `CLOCK_THREAD_CPUTIME_ID`, aligning with Linux's behavior, see <https://elixir.bootlin.com/linux/v6.18/source/kernel/time/posix-timers.c#L1379-L1385>.

### DIFF
--- a/kernel/src/syscall/nanosleep.rs
+++ b/kernel/src/syscall/nanosleep.rs
@@ -100,11 +100,17 @@ fn do_clock_nanosleep(
             ClockId::CLOCK_PROCESS_CPUTIME_ID => {
                 ctx.process.timer_manager().prof_timer().timer_manager()
             }
-            // FIXME: From the manual,
-            // the CPU clock IDs returned by clock_getcpuclockid(3)
-            // and pthread_getcpuclockid(3) can also be passed in clockid.
-            // But it's not covered here.
-            _ => return_errno_with_message!(Errno::EINVAL, "unknown clockid for clock_nanosleep"),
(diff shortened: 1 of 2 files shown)


kernel/src/syscall/nanosleep.rs
```diff
@@ -100,6 +100,12 @@ fn do_clock_nanosleep(
             ClockId::CLOCK_PROCESS_CPUTIME_ID => {
                 ctx.process.timer_manager().prof_timer().timer_manager()
             }
+            ClockId::CLOCK_THREAD_CPUTIME_ID => {
+                return_errno_with_message!(
+                    Errno::EOPNOTSUPP,
+                    "unsupported clockid for clock_nanosleep"
+                );
+            }
```
StevenJiang1110, line 113: The fix looks correct. However, it appears that many other ClockId variants also do not support nanosleep — for example, CLOCK_MONOTONIC_RAW, CLOCK_REALTIME_COARSE, and CLOCK_MONOTONIC_COARSE. These should all return EOPNOTSUPP, but currently they are still returning EINVAL.

Reference: https://elixir.bootlin.com/linux/v6.18/source/kernel/time/posix-timers.c#L1467

Could you take a look at the other clock IDs and fix the remaining cases as part of this patch as well?

It would also be good to add a comment explaining that these clocks are low-resolution and therefore do not support nanosleep.

StevenJiang1110 (commented):
Thanks for your fix!

StevenJiang1110 (approved):

PR #3048, opened 2026-03-18
Introduce `Dentry` revalidate mechanism

cchanging:
This PR implements `Dentry` revalidation mechanism, replacing the original `is_dentry_cacheable` API. With this PR, all dentries will first enter the `DentryCache`, and then be revalidated on demand upon a cache hit. This PR also provide more complete support for negative dentry in `DentryCache`.

Through this PR, the original validation logic in `procfs` has been handed over to the VFS layer, allowing the internal cache to be removed. The offset consistency semantics for `readdir` operations in `procfs`, previously provided by the `SlotVec` cache, are now handled through an additional API to prevent offset inconsistencies caused by changes in directory contents during consecutive `getdents` operations.

This PR does not modify the existing caching mechanisms for `devpts` and `cgroupfs`, preserving their original semantics. Their more specific revalidation mechanisms will be implemented later.

PR #3148, opened 2026-04-16
Fix `rename`/`renameat2` and enable more LTP test cases

cqs21:
This PR fixes Linux compatibility for `rename` and `renameat2`.

- Add a scratch test disk to initramfs setup so LTP cases that require a block device can run.
- Enable additional `rename*` and `renameat2*` LTP cases, while keeping the unsupported `exfat` cases explicitly blocked.
- Validate rename basename lengths against each filesystem's `namelen` limit and enforce sticky-bit ownership checks during rename.
- Make `ramfs` create new inodes with the caller's `fsuid` and `fsgid`, which aligns ownership semantics with the sticky-bit rename tests.
- Add `RENAME_NOREPLACE` and `RENAME_EXCHANGE` support for `renameat2`.

### DIFF
--- a/kernel/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/src/fs/fs_impls/ramfs/fs.rs
@@ -33,7 +33,8 @@ use crate::{
         },
     },
     prelude::*,
-    process::{Gid, Uid},
+    process::{Gid, Uid, posix_thread::AsPosixThread},
+    thread::Thread,
     time::clocks::RealTimeCoarseClock,
     vm::vmo::Vmo,
 };
@@ -843,15 +844,18 @@ impl Inode for RamInode {
         }
 
         let fs = self.fs.upgrade().unwrap();
+        let (uid, gid) = Thread::current()
+            .and_then(|thread| {
+                let posix_thread = thread.as_posix_thread()?;
+                let credentials = posix_thread.credentials();
+                Some((credentials.fsuid(), credentials.fsgid()))
+            })
+            .unwrap_or((Uid::new_root(), Gid::new_root()));
         let new_inode = match type_ {
-            InodeType::File => RamInode::new_file(&fs, mode, Uid::new_root(), Gid::new_root()),
-            InodeType::SymLink => {
-                RamInode::new_symlink(&fs, mode, Uid::new_root(), Gid::new_root())
-            }
-            InodeType::Socket => RamInode::new_socket(&fs, mode, Uid::new_root(), Gid::new_root()),
-            InodeType::Dir => {
-                RamInode::new_dir(&fs, mode, Uid::new_root(), Gid::new_root(), &self.this)
-            }
+            InodeType::File => RamInode::new_file(&fs, mode, uid, gid),
+            InodeType::SymLink => RamInode::new_symlink(&fs, mode, uid, gid),
+            InodeType::Socket => RamInode::new_socket(&fs, mode, uid, gid),
+            InodeType::Dir => RamInode::new_dir(&fs, mode, uid, gid, &self.this),
             _ => {
                 panic!("unsupported inode type");
             }
--- a/kernel/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/src/fs/fs_impls/ramfs/fs.rs
@@ -28,7 +28,7 @@ use crate::{
             file_system::{FileSystem, FsEventSubscriberStats, SuperBlock},
             inode::{
                 Extension, FallocMode, FileOps, HardLinkability, Inode, Metadata, MknodType,
-                SymbolicLink,
+                RenameMode, SymbolicLink,
             },
             path::{is_dot, is_dot_or_dotdot, is_dotdot},
             registry::{FsCreationCtx, FsProperties, FsType},
@@ -464,6 +464,18 @@ impl DirEntry {
         Some(substitute)
     }
 
+    fn exchange_entry_inodes(&mut self, idx_a: usize, idx_b: usize) {
+        assert!(idx_a >= NUM_SPECIAL_ENTRIES && idx_b >= NUM_SPECIAL_ENTRIES);
+        let child_idx_a = idx_a - NUM_SPECIAL_ENTRIES;
+        let child_idx_b = idx_b - NUM_SPECIAL_ENTRIES;
+
+        let (name_a, inode_a) = self.children.remove(child_idx_a).unwrap();
+        let (name_b, inode_b) = self.children.remove(child_idx_b).unwrap();
+
+        self.children.put_at(child_idx_a, (name_a, inode_b));
+        self.children.put_at(child_idx_b, (name_b, inode_a));
+    }
+
     fn visit_entry(&self, idx: usize, visitor: &mut dyn DirentVisitor) -> Result<usize> {
         let try_visit = |idx: &mut usize, visitor: &mut dyn DirentVisitor| -> Result<()> {
             // Read the two special entries("." and "..").
@@ -668,6 +680,14 @@ impl RamInode {
             .get_entry(name)
             .ok_or(Error::new(Errno::ENOENT))?;
         Ok(inode)
+    }
+
+    fn set_parent_if_dir(&self, parent: Weak<RamInode>) {
+        if self.typ != InodeType::Dir {
+            return;
+        }
+
+        self.inner.as_direntry().unwrap().write().set_parent(parent);
     }
 }
 
@@ -1097,28 +1117,13 @@ impl Inode for RamInode {
         Ok(inode as _)
     }
 
-    fn rename(&self, old_name: &str, target: &Arc<dyn Inode>, new_name: &str) -> Result<()> {
-        if is_dot_or_dotdot(old_name) {
-            return_errno_with_message!(Errno::EISDIR, "old_name is . or ..");
-        }
-        if is_dot_or_dotdot(new_name) {
-            return_errno_with_message!(Errno::EISDIR, "new_name is . or ..");
-        }
-
-        let target = target
-            .downcast_ref::<RamInode>()
-            .ok_or(Error::new(Errno::EXDEV))?;
-
-        if !Arc::ptr_eq(&self.fs(), &target.fs()) {
-            return_errno_with_message!(Errno::EXDEV, "not same fs");
-        }
-        if self.typ != InodeType::Dir {
-            return_errno_with_message!(Errno::ENOTDIR, "self is not dir");
-        }
-        if target.typ != InodeType::Dir {
-            return_errno_with_message!(Errno::ENOTDIR, "target is not dir");
-        }
-
+    fn rename(
+        &self,
+        old_name: &str,
+        target: &Arc<dyn Inode>,
+        new_name: &str,
+        mode: RenameMode,
+    ) -> Result<()> {
         // Perform necessary checks to ensure that `dst_inode` can be replaced by `src_inode`.
         let check_replace_inode =
             |src_inode: &Arc<RamInode>, dst_inode: &Arc<RamInode>| -> Result<()> {
@@ -1149,39 +1154,40 @@ impl Inode for RamInode {
                 Ok(())
             };
 
+        let target = target.downcast_ref::<RamInode>().unwrap();
+
         // Rename in the same directory
         if self.ino == target.ino {
             let mut self_dir = self.inner.as_direntry().unwrap().write();
-            let (src_idx, src_inode) = self_dir
-                .get_entry(old_name)
-                .ok_or(Error::new(Errno::ENOENT))?;
-            let is_dir = src_inode.typ == InodeType::Dir;
+            // The source is guaranteed to exist (checked by VFS layer).
+            let (src_idx, src_inode) = self_dir.get_entry(old_name).unwrap();
 
-            if let Some((dst_idx, dst_inode)) = self_dir.get_entry(new_name) {
+            if mode == RenameMode::Exchange {
+                // The destination is guaranteed to exist for `RenameMode::Exchange` (checked by VFS layer).
+                let (dst_idx, dst_inode) = self_dir.get_entry(new_name).unwrap();
+                self_dir.exchange_entry_inodes(src_idx, dst_idx);
+                drop(self_dir);
+
+                let now = now();
+                DirChange::touch().apply(self, now);
+                src_inode.set_ctime(now);
+                dst_inode.set_ctime(now);
+            } else if let Some((dst_idx, dst_inode)) = self_dir.get_entry(new_name) {
                 check_replace_inode(&src_inode, &dst_inode)?;
                 self_dir.remove_entry(dst_idx);
                 self_dir.substitute_entry(src_idx, (CStr256::from(new_name), src_inode.clone()));
                 drop(self_dir);
 
                 let now = now();
-                let mut self_meta = self.metadata.lock();
-                self_meta.dec_size();
-                if is_dir {
-                    self_meta.dec_nlinks();
-                }
-                self_meta.set_mtime(now);
-                self_meta.set_ctime(now);
-                drop(self_meta);
+                DirChange::del(&src_inode).apply(self, now);
                 src_inode.set_ctime(now);
                 dst_inode.set_ctime(now);
             } else {
                 self_dir.substitute_entry(src_idx, (CStr256::from(new_name), src_inode.clone()));
                 drop(self_dir);
+
                 let now = now();
-                let mut self_meta = self.metadata.lock();
-                self_meta.set_mtime(now);
-                self_meta.set_ctime(now);
-                drop(self_meta);
+                DirChange::touch().apply(self, now);
                 src_inode.set_ctime(now);
             }
         }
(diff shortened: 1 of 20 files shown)


kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -466,6 +466,11 @@ impl DirDentry<'_> {
 
         let old_dir_inode = old_dir.inode();
         let new_dir_inode = new_dir.inode();
+        if old_name.len() > old_dir_inode.fs().sb().namelen
+            || new_name.len() > new_dir_inode.fs().sb().namelen
```
cchanging, line 470: `new_dir_inode` and `old_dir_inode` must be in the same file systems (having been checked in `Path::rename`), so we can obtain `old_dir_inode.fs().sb().namelen` first and reuse it.

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -482,6 +504,15 @@ impl DirDentry<'_> {
             let old_dentry = children.check_mountpoint_then_find(old_name)?;
             children.check_mountpoint(new_name)?;
 
+            let old_inode = match old_dentry.as_ref() {
+                Some(dentry) => dentry.inode().clone(),
+                None => old_dir_inode.lookup(old_name)?,
```
cchanging, line 509: Here misses an opportunity to synchronize the lookup result into the `DentryCache`. It may require the new API introduced in #3048 ...

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -482,6 +504,15 @@ impl DirDentry<'_> {
...
+            };
+            Self::check_sticky_bit(old_dir_inode, &old_inode)?;
+            if let Ok(Some(new_dentry)) = children.find(new_name) {
+                Self::check_sticky_bit(old_dir_inode, new_dentry.inode())?;
```
cchanging, line 513: `check_sticky_bit` contains two pieces of logic: one checks whether the `dir` directory has the sticky bit set, and the other checks whether the target inode has the corresponding permissions. I think these two checks could be separated. Otherwise, we would end up checking whether `dir` has the sticky bit set twice here.

cchanging (commented):
Thanks for enhancing the `rename`/`renameat2` syscalls! The new feature and error handling added in this PR both look correct. However, it also involves some logic around operations on the `DentryCache`, especially in the last commit. Personally, I think the implementation would become clearer if it were revised later based on #3048 .

tatetian:
@cqs21 A kind ping to continue working on this PR.

tatetian:
@cchanging Could you review this PR again?

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -458,6 +459,50 @@ impl DirDentry<'_> {
...
+    /// Resolves a child dentry for a rename operation.
+    ///
+    /// Probes the cache with revalidation and mountpoint checks, then falls
+    /// back to a filesystem lookup and caches the result.
+    fn resolve_child_for_rename(
+        &self,
+        children: &mut DentryChildren,
+        name: &str,
+        parent_arc: &Arc<Dentry>,
+    ) -> Result<Arc<Dentry>> {
```
cchanging, line 500: We do not need to have the `parent_arc` parameter. Using `self.this()` is enough.

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -458,6 +459,50 @@ impl DirDentry<'_> {
...
+        if let Some(dentry) = children.probe_cached_child_for_rename(self, name)? {
+            return Ok(dentry);
+        }
```
cchanging: After removing the `parent_arc` parameter, I think `resolve_child_for_rename` and `probe_cached_child_for_rename` can be merged into a single method, or at least made methods on the same object. Their semantics are very similar; the only difference is that `resolve_child_for_rename` falls back to an FS lookup when the dentry is not present in the cache.

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -705,44 +750,59 @@ impl DirDentry<'_> {
 
             let mut children = old_dir.children.write();
             children.check_mountpoint(new_name)?;
```
cchanging: I just realized it seems that all of the `check_mountpoint` operation is redundant with the logics in `resolve_child_for_rename`? We can remove `check_mountpoint` method.

kernel/src/syscall/rename.rs
```diff
@@ -28,12 +28,19 @@ pub fn sys_renameat2(
...
         return_errno_with_message!(Errno::EINVAL, "invalid flags");
     };
-    // TODO: Add support for handling the `NOREPLACE`, `EXCHANGE`, and `WHITEOUT` flags.
-    if !flags.is_empty() {
-        warn!("unsupported flags: {:?}", flags);
-        return_errno_with_message!(Errno::EINVAL, "unsupported flags");
+    // TODO: Add support for handling the `WHITEOUT` flag.
+    if flags.contains(Flags::WHITEOUT) || flags.contains(Flags::NOREPLACE | Flags::EXCHANGE) {
+        return_errno_with_message!(Errno::EINVAL, "invalid renameat2 flag combination");
     }
```
cchanging: I suggest handling the error messages for `WHITEOUT`, which is "unsupport flag", and invalid combination separately.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -35,6 +35,17 @@ mod mount;
 mod mount_namespace;
 mod resolver;
 
+/// Specifies the behavior of `renameat2()` when renaming a path.
```
cchanging: I suggest just say "Specifies the behavior of the rename operation".

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -754,59 +768,191 @@ impl DirDentry<'_> {
...
+                    old_dentry
+                        .name_and_parent
+                        .set(new_name, old_dir_arc.clone())
+                        .unwrap();
+                    old_dir.insert_positive_child(&mut children, new_name, old_dentry);
+                }
+                RenameMode::Exchange => {
+                    let new_dentry = new_dentry.unwrap();
+                    children.delete(old_name);
+                    children.delete(new_name);
```
cchanging: These delete operations can be removed.

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -754,59 +768,191 @@ impl DirDentry<'_> {
...
+            RenameMode::Replace | RenameMode::NoReplace => {
+                old_dir_inode.rename(old_name, new_dir_inode, new_name)?;
+            }
+            RenameMode::Exchange => {
+                let temp_name = temp_name.unwrap();
+                old_dir_inode.rename(old_name, old_dir_inode, temp_name)?;
+                new_dir_inode.rename(new_name, old_dir_inode, old_name)?;
+                old_dir_inode.rename(temp_name, new_dir_inode, new_name)?;
+            }
         }
```
cchanging: I think we should add the `mode` parameter to `Inode::rename` and implement this logic in `Inode`.

kernel/src/fs/fs_impls/ext2/inode/dir/mod.rs
cchanging, line 323: With this PR, we have obtained the old dentry and new dentry in the VFS layer, so we can perform this check in the VFS layer.

After that, I think we can rename the `replaced_inode` to `target_inode`, and then we can merge `validate_exchange_invariants` to `validate_rename_invariants`.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -1155,33 +1241,33 @@ impl Inode for RamInode {
             let (src_idx, src_inode) = self_dir
                 .get_entry(old_name)
                 .ok_or(Error::new(Errno::ENOENT))?;
-            let is_dir = src_inode.typ == InodeType::Dir;
 
-            if let Some((dst_idx, dst_inode)) = self_dir.get_entry(new_name) {
+            if mode == RenameMode::Exchange {
+                let (dst_idx, dst_inode) = self_dir
+                    .get_entry(new_name)
+                    .ok_or(Error::new(Errno::ENOENT))?;
```
cchanging: We have checked the existence of the `new_entry` so I think we can directly unwrap here and add a comment.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -464,6 +464,18 @@ impl DirEntry {
...
+    fn exchange_entries(&mut self, idx_a: usize, idx_b: usize) {
+        assert!(idx_a >= NUM_SPECIAL_ENTRIES && idx_b >= NUM_SPECIAL_ENTRIES);
+        let ca = idx_a - NUM_SPECIAL_ENTRIES;
+        let cb = idx_b - NUM_SPECIAL_ENTRIES;
+
+        let inode_a = self.children.get(ca).unwrap().1.clone();
+        let inode_b = self.children.get(cb).unwrap().1.clone();
+
+        self.children.get_mut(ca).unwrap().1 = inode_b;
+        self.children.get_mut(cb).unwrap().1 = inode_a;
```
cchanging: We can use `SlotVec::remove` and `SlotVec::put_at` method to avoid additional `clone`.

Besides, can we give `ca`/`cb` better names? I don't know what they mean.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -669,6 +681,74 @@ impl RamInode {
...
+            if removed_entry_type == InodeType::Dir {
+                meta.dec_nlinks();
+            }
+        });
+    }
+
+    fn update_parent_metadata_after_entry_addition(
+        &self,
+        added_entry_type: InodeType,
+        now: Duration,
```
cchanging: Here introduces too many helper methods which have similar behaviors. I  think maybe we can only keep `update_parent_metadata_for_rename` and replace the `update_meta_fn` with an enum?

```rust
enum DirChange {
    DecEntry(InodeType),
    IncEntry(InodeType),
    ExchangeEntry {old_type: InodeType, new_type: InodeType},
    NoChange,
}
```

Can this signature make the code simpler?

Besides, I think we may need to add some documents to `update_parent_metadata_for_rename` or give it a better name. At the first glance, I thought it will modify the metadate for the parent of `self`.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -464,6 +464,19 @@ impl DirEntry {
...
+        assert!(idx_a >= NUM_SPECIAL_ENTRIES && idx_b >= NUM_SPECIAL_ENTRIES);
+        let child_idx_a = idx_a - NUM_SPECIAL_ENTRIES;
+        let child_idx_b = idx_b - NUM_SPECIAL_ENTRIES;
+
+        let mut entry_a = self.children.remove(child_idx_a).unwrap();
+        let mut entry_b = self.children.remove(child_idx_b).unwrap();
+        core::mem::swap(&mut entry_a.1, &mut entry_b.1);
+
+        self.children.put_at(child_idx_a, entry_a);
+        self.children.put_at(child_idx_b, entry_b);
```
cchanging: I'm not sure why `mem::swap` is used here. Can we do this:

```rust
let (name_a, entry_a) = self.children.remove(child_idx_a).unwrap();
let (name_b, entry_b) = self.children.remove(child_idx_b).unwrap();

self.children.put_at(child_idx_a, (name_a, entry_b));
self.children.put_at(child_idx_b, (name_b, entry_a));
```

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -1155,33 +1213,32 @@ impl Inode for RamInode {
             let (src_idx, src_inode) = self_dir
                 .get_entry(old_name)
                 .ok_or(Error::new(Errno::ENOENT))?;
-            let is_dir = src_inode.typ == InodeType::Dir;
 
-            if let Some((dst_idx, dst_inode)) = self_dir.get_entry(new_name) {
+            if mode == RenameMode::Exchange {
+                // The destination is guaranteed to exist for RENAME_EXCHANGE (checked by VFS layer).
```
cchanging: Suggest: RENAME_EXCHANGE -> `RenameMode::Exchange`

kernel/src/fs/fs_impls/ramfs/fs.rs
cchanging: Since you have modified this method, you could do a favor to remove these redundant checks and error handlings which have been implied in VFS layer.

cchanging (approved):
Just some minor issues, otherwise LGTM!

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -35,6 +35,8 @@ mod mount;
 mod mount_namespace;
 mod resolver;
 
+pub use super::inode::RenameMode;
```
lrh2000: What is it.... Move it to a correct location and remove `pub`.

```patch
-            inode::{HardLinkability, Inode, Metadata, MknodType},
+            inode::{HardLinkability, Inode, Metadata, MknodType, RenameMode},
```

`inode` is already a public module I think. So there is no need to re-export it in the path module.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -669,6 +681,45 @@ impl RamInode {
...
+                    meta.inc_nlinks();
+                }
+            }
+            DirChange::ExchangeEntry { old_type, new_type } if old_type != new_type => {
+                if old_type == InodeType::Dir {
+                    meta.dec_nlinks();
+                }
+                if new_type == InodeType::Dir {
+                    meta.inc_nlinks();
+                }
```
lrh2000: It sounds like... Is this better?
```rust
struct DirChange {
    new_type: Option<InodeType>,
    old_type: Option<InodeType>,
}

DirChange::add(src_inode).apply(self, now);
DirChange::del(src_inode).apply(target, now);
DirChange::exchange(src_inode, dst_inode).apply(...);
```

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -1192,17 +1228,41 @@ impl Inode for RamInode {
...
+                    now,
+                );
+                target.apply_dir_change_for_rename(
+                    DirChange::ExchangeEntry {
+                        old_type: dst_inode.typ,
+                        new_type: src_inode.typ,
+                    },
+                    now,
+                );
+
```
lrh2000: Remove this blank line.

I think all other code does not have this blank line. Be consistent:
```rust
                let now = now();
                self.apply_dir_change_for_rename(DirChange::DecEntry(src_inode.typ), now);
                target.apply_dir_change_for_rename(DirChange::IncEntry(src_inode.typ), now);
                src_inode.set_ctime(now);
```

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -771,23 +783,41 @@ impl DirDentry<'_> {
...
+                    new_dentry
+                        .name_and_parent
+                        .set(old_name, old_dir_arc.clone())
+                        .unwrap();
+                    old_dir.insert_positive_child(&mut children, new_name, old_dentry);
+                    old_dir.insert_positive_child(&mut children, old_name, new_dentry);
+                }
+            }
         } else {
-            // The two are different dentries
```
lrh2000, line 717: The original comment should be kept or revised?
```rust
        if Arc::ptr_eq(old_dir_arc, new_dir_arc) {
            // The two are the same dentry, we just modify the name
            ...
        } else {
            // The two are different dentries
        }
```

kernel/src/syscall/rename.rs
```diff
@@ -26,13 +26,26 @@ pub fn sys_renameat2(
...
-        warn!("unsupported flags: {:?}", flags);
-        return_errno_with_message!(Errno::EINVAL, "unsupported flags");
+    // TODO: Add support for handling the `WHITEOUT` flag.
+    if flags.contains(Flags::WHITEOUT) {
+        return_errno_with_message!(Errno::EINVAL, "unsupported flag: WHITEOUT");
     }
+    if flags.contains(Flags::NOREPLACE | Flags::EXCHANGE) {
+        return_errno_with_message!(
+            Errno::EINVAL,
+            "NOREPLACE and EXCHANGE are mutually exclusive"
```
lrh2000: Just comment out the `WHITEOUT` flag.
```diff
-        const WHITEOUT  = 1 << 2;
+        // TODO: Add support for handling the `WHITEOUT` flag.
+        // const WHITEOUT  = 1 << 2;
```

Also for consistency with existing code:
```suggestion
            "NOREPLACE and EXCHANGE cannot be used together"
```
```
> grep -R "cannot be used together" kernel
kernel/src/process/clone.rs:                "`CLONE_FS` cannot be used together with `CLONE_NEWNS` or `CLONE_NEWUSER`"
kernel/src/process/clone.rs:                    "`CLONE_PARENT` cannot be used together with `CLONE_NEWUSER` or `CLONE_NEWPID`"
kernel/src/process/clone.rs:                    "`CLONE_THREAD` cannot be used together with `CLONE_PIDFD` or `CLONE_NEWUSER`"
kernel/src/syscall/fallocate.rs:                "PUNCH_HOLE and ZERO_RANGE cannot be used together"
```

lrh2000 (commented):
(Only some random comments. I don't plan on checking the other details here...)

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -1145,39 +1185,39 @@ impl Inode for RamInode {
...
-                self_meta.set_mtime(now);
-                self_meta.set_ctime(now);
-                drop(self_meta);
+                self.apply_dir_change_for_rename(DirChange::DecEntry(src_inode.typ), now);
                 src_inode.set_ctime(now);
                 dst_inode.set_ctime(now);
             } else {
                 self_dir.substitute_entry(src_idx, (CStr256::from(new_name), src_inode.clone()));
                 drop(self_dir);
                 let now = now();
```
lrh2000, line 1189: And we should have a blank line here since you have it for other places:
```rust
                drop(self_dir);

                let now = now();
```

tatetian:
@cqs21 Needs to resolve the conflicts. Thanks

PR #3149, opened 2026-04-16
Add `devcontainer.json`

chen1plus:
‌‌This PR adds a `devcontainer.json` configuration to allow the editor to automatically activate the environment.
It reduces onboarding time for new contributors by providing a pre-configured workspace.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,15 @@
+{
+  "name": "Asterinas Dev Container",
+  "image": "asterinas/asterinas:0.17.2-20260407",
+  // Mount the workspace and the `/dev` directory
+  "workspaceFolder": "/root/asterinas",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/root/asterinas,type=bind",
+  "mounts": ["source=/dev,target=/dev,type=bind"],
+  // Run with privileged mode and host networking
+  "privileged": true,
+  "runArgs": ["--network=host"],
+  // Enable extensions for Rust development
+  "customizations": {
+    "vscode": { "extensions": ["rust-lang.rust-analyzer"] }
+  }
+}
--- a/.github/workflows/test_devcontainer.yml
+++ b/.github/workflows/test_devcontainer.yml
@@ -1,0 +1,32 @@
+name: Test devcontainer
+
+on:
+  pull_request:
+    paths:
+      - ".devcontainer/**"
+  push:
+    branches: [main]
+    paths:
+      - ".devcontainer/**"
+  workflow_run:
+    workflows: ["Publish Docker images"]
+    types:
+      - completed
+
+jobs:
+  test-devcontainer:
+    # DO NOT run when the "Publish Docker images" workflow failed.
+    if: ${{ github.event_name != 'workflow_run'
+      || github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Verify image tag matches DOCKER_IMAGE_VERSION
+        run: |
+          EXPECTED=$(cat DOCKER_IMAGE_VERSION)
+          grep -q "asterinas/asterinas:${EXPECTED}" .devcontainer/devcontainer.json \
+            || { echo "devcontainer.json image doesn't match DOCKER_IMAGE_VERSION"; exit 1; }
+      - name: Build and verify dev container
+        uses: devcontainers/ci@v0.3
+        with:
+          runCmd: rustc --version && cargo --version && make --version
--- a/README.md
+++ b/README.md
@@ -154,6 +154,10 @@ Follow the steps below to get Asterinas up and running.
     docker run -it --privileged --network=host -v /dev:/dev -v $(pwd)/asterinas:/root/asterinas asterinas/asterinas:0.17.1-20260319
     ```
 
+    Alternatively, if you use VS Code with the
+    [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
+    extension, open the cloned folder and select "Reopen in Container".
+
 3. Inside the container,
 go to the project folder (`/root/asterinas`) and run:
 
--- a/book/src/kernel/README.md
+++ b/book/src/kernel/README.md
@@ -83,6 +83,10 @@ since `distro/README.md` references them -->
                 asterinas/asterinas:0.17.1-20260319
     ```
 
+    Alternatively, if you use VS Code with the
+    [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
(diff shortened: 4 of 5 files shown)


junyang-zh:
I wonder, which tool reads the `.devcontainer/` files?

chen1plus:
> I wonder, which tool reads the `.devcontainer/` files?

VS Code, the JetBrains suite, and Zed all provide native support for Dev Containers.
See https://containers.dev/supporting for more detail.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "Asterinas Dev Container",
+  "image": "asterinas/asterinas:0.17.2-20260407",
```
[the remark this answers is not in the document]
chen1plus: The document should be updated instead.

tatetian, line 3: **Wire this file into `tools/bump_version.sh`**: `update_all_docker_version_refs()` (lines 118-144) has a fixed allowlist of files to rewrite on `--docker_version_refs`. This new file is not in it, so future bumps will skip it. Please add:
```
update_image_versions ${ASTER_SRC_DIR}/.devcontainer/devcontainer.json
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
...
+  // Mount the workspace and the `/dev` directory
+  "workspaceFolder": "/root/asterinas",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/root/asterinas,type=bind",
+  "mounts": ["source=/dev,target=/dev,type=bind"],
+  // Run with privileged mode and host networking
+  "privileged": true,
+  "runArgs": ["--network=host"],
```
tatetian, line 10: **Trailing comma**: JSONC (VS Code) tolerates it, but strict JSON parsers (`jq`, `python -m json.tool`) will reject it. Suggest removing.

.devcontainer/devcontainer.json
tatetian, line 1: **Consider adding VS Code extensions**: Adding a `customizations` block would give new contributors a working IDE out of the box. For example:

```jsonc
  "customizations": {
    "vscode": {
      "extensions": [
        "rust-lang.rust-analyzer"
      ]
    }
  }
```

tatetian, line 1: **Consider a CI workflow to validate the devcontainer**: A [`devcontainers/ci`](https://github.com/devcontainers/ci) job would catch stale image tags and broken configs. Something like:

```yaml
# .github/workflows/test_devcontainer.yml
name: Test devcontainer
on:
  pull_request:
    paths:
      - '.devcontainer/**'
      - 'DOCKER_IMAGE_VERSION'
  push:
    branches: [main]
    paths:
      - '.devcontainer/**'
      - 'DOCKER_IMAGE_VERSION'

jobs:
  test-devcontainer:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Verify image tag matches DOCKER_IMAGE_VERSION
        run: |
          EXPECTED=$(cat DOCKER_IMAGE_VERSION)
          grep -q "asterinas/asterinas:${EXPECTED}" .devcontainer/devcontainer.json \
            || { echo "devcontainer.json image doesn't match DOCKER_IMAGE_VERSION"; exit 1; }
      - name: Build and verify dev container
        uses: devcontainers/ci@v0.3
        with:
          runCmd: rustc --version && cargo --version && make --version
```

The `DOCKER_IMAGE_VERSION` path trigger is the key part — it fires when the image is bumped but `devcontainer.json` isn't updated.

tatetian, line 1: **Mention "Reopen in Container" in the README**: Step 2 of "For Kernel
Developers" (`README.md` line 151) only shows `docker run`. Consider
adding a note like:

> **Alternatively**, if you use VS Code with the
> [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
> extension, open the cloned folder and select **"Reopen in Container"**.

Without this, most contributors won't discover the devcontainer.

tatetian (commented):
This sounds a useful addition.

.devcontainer/devcontainer.json
chen1plus: The trigger should depend on the Docker CI, or this will fail because the image hasn't been built and published.

.github/workflows/test_devcontainer.yml
```diff
@@ -0,0 +1,29 @@
+name: Test Dev Container
```
tatetian: This workflow tests the devcontainer configuration, not the Dev Container. So using "devcontainer" in the name is more accurate.

.github/workflows/test_devcontainer.yml
```diff
@@ -0,0 +1,29 @@
...
+    paths:
+      - ".devcontainer/**"
+  push:
+    branches: [main]
+    paths:
+      - ".devcontainer/**"
+  workflow_run:
+    workflows: ["Publish Docker images"]
+    types:
+      - completed
```
tatetian, line 14: **Guard against failed publishes**: `completed` fires regardless of
whether the publish succeeded or failed. Add a job-level condition so
the test only runs after a successful publish:

```yaml
jobs:
  test-devcontainer:
    if: >-
      # DO NOT run when the "Publish Docker images" workflow failed.
      github.event_name != 'workflow_run'
      || github.event.workflow_run.conclusion == 'success'
    runs-on: ubuntu-latest
```

Without this, a failed Docker image publish will trigger a devcontainer
test that also fails (image doesn't exist), producing a confusing double
failure.

README.md
```diff
@@ -150,6 +150,10 @@ Follow the steps below to get Asterinas up and running.
 
 2. Run a Docker container as the development environment:
 
+    > **Alternatively**, if you use VS Code with the
+    > [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
+    > extension, open the cloned folder and select **"Reopen in Container"**.
```
tatetian: **Don't use a blockquote here**: The `>` prefix renders as a
callout/admonition, giving too much visual weight to what's just an
alternative. Place it as plain text *after* the `docker run` code block
instead:

```markdown
    ```bash
    docker run -it --privileged ...
    ```

    Alternatively, if you use VS Code with the
    [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
    extension, open the cloned folder and select "Reopen in Container".
```

Same applies to `book/src/kernel/README.md`.

tatetian (commented):
All 5 previous comments have been addressed — nice work. The
`bump_version.sh` wiring, VS Code extensions, CI workflow, README
mentions, and trailing comma fix all look good.

A few small issues remain in the CI workflow and README formatting.

README.md
chen1plus: I also removed the unnecessary bold text.

chen1plus:
Thanks for the guidance, much appreciated.

tatetian (approved):

PR #3150, opened 2026-04-16
Add `tkill` syscall

vvvvsv:
The `tkill` syscall is [deprecated](https://man7.org/linux/man-pages/man2/tgkill.2.html). However GDB uses it as a legacy dependency. Required for #2323.

Additionally, `tgkill` should return EINVAL when either `tgid` or `tid` equals to 0.

---

Adding this syscall causes the LTP test `signal06` to fail. This is because, on Linux, if writing the signal frame fails when handling a user signal, the kernel retries a SIGSEGV delivery, instead of killing the process directly. We should fix this issue in a separate PR.

### DIFF
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -220,7 +220,7 @@ which are summarized in the table below.
 | 197     | removexattr            | ✅             | 💯 |
 | 198     | lremovexattr           | ✅             | 💯 |
 | 199     | fremovexattr           | ✅             | 💯 |
-| 200     | tkill                  | ❌             | N/A |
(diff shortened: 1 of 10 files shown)


tatetian:
Looks good. Please add the new system call to the "[Linux Compatibility](https://asterinas.github.io/book/kernel/linux-compatibility/index.html#linux-compatibility)" page of Asterinas book as well as the SCML file.

vvvvsv:
> Looks good. Please add the new system call to the "[Linux Compatibility](https://asterinas.github.io/book/kernel/linux-compatibility/index.html#linux-compatibility)" page of Asterinas book as well as the SCML file.

Done. :)

tatetian (approved):

Issue #2935, opened 2026-01-28
Duplicate output during kernel initialization


Fischer0522:
### Describe the bug

Currently, kernel initialization prints all output twice, including logs and the Asterinas logo.

### To Reproduce

```
make kernel
make run_kernel
```

### Screenshots

<img width="2272" height="2514" alt="Image" src="https://github.com/user-attachments/assets/87f9a38e-76c5-4df4-9692-164adde62d32" />

lrh2000:
Hmmm, it's because all consoles print kernel logs:
https://github.com/asterinas/asterinas/blob/a599a2c563b857c8c69ab512d3405c8fa6a12932/kernel/comps/logger/src/console.rs#L14-L36

However, both the virtio and serial consoles are bound to standard output:
https://github.com/asterinas/asterinas/blob/a599a2c563b857c8c69ab512d3405c8fa6a12932/tools/qemu_args.sh#L78
https://github.com/asterinas/asterinas/blob/a599a2c563b857c8c69ab512d3405c8fa6a12932/tools/qemu_args.sh#L52

We can remove the serial console from the standard output by default? However, if we do this, we will lose the `ostd::early_println` logs, which is not ideal.
```patch
diff --git a/tools/qemu_args.sh b/tools/qemu_args.sh
index 92696f498..1b69cd956 100755
--- a/tools/qemu_args.sh
+++ b/tools/qemu_args.sh
@@ -49,7 +49,9 @@ else
 fi
 
 if [ "$CONSOLE" = "hvc0" ]; then
-    CONSOLE_ARGS="-device virtconsole,chardev=mux"
+    CONSOLE_ARGS="-device virtconsole,chardev=mux -serial /dev/null"
+else
+    CONSOLE_ARGS="-serial chardev:mux"
 fi
 
 if [ "$1" = "tdx" ]; then
@@ -75,7 +77,6 @@ if [ "$1" = "tdx" ]; then
         $CONSOLE_ARGS \
         -device isa-debug-exit,iobase=0xf4,iosize=0x04 \
         -monitor chardev:mux \
-        -serial chardev:mux \
         -d guest_errors \
     "
     echo $QEMU_ARGS
@@ -89,7 +90,6 @@ COMMON_QEMU_ARGS="\
     --no-reboot \
     -nographic \
     -display vnc=0.0.0.0:${VNC_PORT:-42} \
-    -serial chardev:mux \
     -monitor chardev:mux \
[...]

PR #3151, opened 2026-04-16
Fix RISC-V duplicate output during kernel initialization by redirecting serial output

le-monde-bleu:
This PR fixes the problem of duplicate output during kernel initialization in RISC-V scheme by redirecting serial output to `qemu-serial.log`, and moves the RISC-V QEMU run configuration to `qemu_args.sh` in compliance with x86_64 scheme.

Refer to discussions in #3130 and #2936 .

For discussions about long-term modification, refer to #2935.


junyang-zh (approved):
LGTM! Thanks for contribution.

PR #3152, opened 2026-04-17
Remove remaining `Ordering::SeqCst` usages

lrh2000:
`Ordering::SeqCst` is used to establish a global order between two atomic variables that cannot be accurately modeled via the "happens-before" relation. See the example at https://sabrinajewson.org/rust-nomicon/atomics/seqcst.html.

I don't believe we have a valid usecase in our current code base. It only exists for historical reasons because we didn't choose the proper order for our project in the early days. Most of them have already been removed. Only a few use cases remain. This PR removes the remaining uses.


cchanging (approved):

tatetian (approved):

PR #3153, opened 2026-04-17
Fix VFS permission checks and `unlink` path resolution

cqs21:
This PR aligns VFS `link` and `unlink` behavior with Linux.

- add a directory-entry modification permission check for `mknod`, `link`, `unlink`, and `rmdir`
- enforce hard-link source permission checks and return `EISDIR` for `unlink` on `.` or `..`
- resolve the full `unlinkat` path before removing the parent entry so trailing-slash and dot-component errors match Linux
- update gVisor and LTP blocklists to reflect the newly passing `link`, `symlink`, and `unlink` cases


tatetian:
@cchanging Could you review this PR? Thanks.

kernel/src/syscall/unlink.rs
```diff
@@ -26,17 +26,24 @@ pub fn sys_unlinkat(
...
+
+        // Linux resolves the full unlink path before invoking the directory operation.
+        // This preserves path-walk errors for trailing slash and dot components, such as
+        // `file/`, `file/.`, and `file/..`, which should fail with `ENOTDIR` instead of
+        // being reduced to unlinking the parent directory entry.
+        path_resolver.lookup_unresolved_no_follow(&fs_path)?;
+
+        let (parent_path_name, target_name) = path_name.split_dirname_and_basename()?;
+        let parent_fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
+        (path_resolver.lookup(&parent_fs_path)?, target_name)
```
cchanging: This approach here may have some issues:
1. We currently perform two lookups: one full lookup and one lookup to the parent, which is somewhat redundant.
2. The early full lookup still cannot guarantee the deletion semantics in the presence of races, because after the full lookup, a directory could have been replaced with a file of the same name.

I think we may need to perform the `ENOTDIR` check while holding the lock in `Dentry`, or change `unlink` so that it takes a `Path` rather than a name as its argument. How does Linux handle this? Does it rely on locking at some other layer?

Besides, this can be a separate commit.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -266,6 +254,50 @@ impl Path {
     fn this(&self) -> Self {
         self.clone()
     }
+
+    /// Checks whether a directory entry may be modified in this directory.
+    fn check_dir_entry_modify_permission(&self) -> Result<()> {
```
cchanging: I don't think it's necessary to mention `dir_entry`. `check_modify_permission` should be enough.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -266,6 +254,50 @@ impl Path {
...
+                "a directory entry cannot be modified on a read-only mount or filesystem"
+            );
+        }
+
+        self.inode()
+            .check_permission(Permission::MAY_WRITE | Permission::MAY_EXEC)
+    }
+
+    /// Checks whether an inode may be used as the source of a hard link.
+    fn check_hardlink_source(inode: &Arc<dyn Inode>) -> Result<()> {
```
cchanging: I think this can be a method of `Path`.

cchanging (commented):
Thanks for Strengthening the relevant checks in the VFS layer!

kernel/src/syscall/unlink.rs
```diff
@@ -26,20 +26,24 @@ pub fn sys_unlinkat(
...
+        let fs_path = FsPath::from_fd_at(dirfd, &path_name, EmptyPathStr::Reject)?;
+        let fs_ref = ctx.thread_local.borrow_fs();
+        let path_resolver = fs_ref.resolver().read();
+
+        let result = path_resolver.lookup_parentat(&fs_path)?;
+        if result.target_is_dir() {
+            return_errno_with_message!(Errno::ENOTDIR, "the path refers to a directory");
+        }
+
+        result.into_parent_and_basename()
```
cchanging: After this change, it seems the target dentry can no longer be resolved, so it’s impossible to report `ENOENT` early (`ENOENT` on Linux seems to have higher priority?). I think we could keep some of the original logic and do something similar instead: 
```rust
    let fs_path = FsPath::from_fd_at(dirfd, &path_name, EmptyPathStr::Reject)?;
    let fs_ref = ctx.thread_local.borrow_fs();
    let path_resolver = fs_ref.resolver().read();

    let (dir_path, name) = {
        let (parent_path_name, target_name) = path_name.split_dirname_and_filename()?;
        let fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
        (path_resolver.lookup(&fs_path)?, target_name)
    };
    path_resolver.lookup_at_path(&dir_path, name)?;
    drop(path_resolver);
    
    let target_is_dir = path_name.end_with('/');
    if target_is_dir { ... }
```

In this way, it seems that we don't have to add additional logics in the `lookup` operation.

lrh2000:
This PR and #3148 need to switch to LSM hooks, see e8f3fd171ad1dff87adee0e4af5241ad6d2352c2 (https://github.com/asterinas/asterinas/pull/3131).

Admittedly, I don't know which user namespace to use in the VFS and FS code. Based on current practice, it seems the initial user namespace is used. However, this may not be correct, though it perhaps does not matter much before we have actual user namespace support.

Together with some other random revisions (in both PRs)... Review them and fix them in your commits if you feel the new changes are reasonable. I made the changes in the final merged version for convenience.

<details>
<summary> Use LSM hooks and other minor revisions </summary>

```patch
diff --git a/kernel/src/fs/vfs/path/dentry.rs b/kernel/src/fs/vfs/path/dentry.rs
index 3e5a2ae8c..a85588b4a 100644
--- a/kernel/src/fs/vfs/path/dentry.rs
+++ b/kernel/src/fs/vfs/path/dentry.rs
@@ -19,7 +19,8 @@ use crate::{
         },
     },
     prelude::*,
-    process::{credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    security::lsm::hooks as lsm_hooks,
     util::random::getrandom,
 };
 
@@ -465,20 +466,23 @@ impl DirDentry<'_> {
     }
 
     fn check_sticky_bit_permission(&self, child_inode: &Arc<dyn Inode>) -> Result<()> {
+        let current_thread = current_thread!();
+        let Some(posix_thread) = current_thread.as_posix_thread() else {
+            return Ok(());
+        };
+
         let dir_metadata = self.inode.metadata();
         let child_metadata = child_inode.metadata();
-        let creds = current_thread!().as_posix_thread().unwrap().credentials();
-        if creds.effective_capset().contains(CapSet::FOWNER)
-            || creds.fsuid() == dir_metadata.uid
-            || creds.fsuid() == child_metadata.uid
-        {
+        let fsuid = posix_thread.credentials().fsuid();
+        if fsuid == dir_metadata.uid || fsuid == child_metadata.uid {
             return Ok(());
         }
 
-        return_errno_with_message!(
-            Errno::EPERM,
-            "sticky bit set and caller is neither file owner nor directory owner"
-        );
+        lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+            UserNamespace::get_init_singleton().as_ref(),
+            posix_thread,
+            CapSet::FOWNER,
+        ))
     }
 
     /// Resolves a child dentry for a rename operation.
@@ -737,12 +741,11 @@ impl DirDentry<'_> {
             return_errno_with_message!(Errno::EBUSY, "old_name is . or ..");
         }
         if is_dot_or_dotdot(new_name) {
-            let errno = if mode == RenameMode::NoReplace {
-                Errno::EEXIST
+            if mode == RenameMode::NoReplace {
+                return_errno_with_message!(Errno::EEXIST, "new_name is . or ..");
             } else {
-                Errno::EBUSY
-            };
-            return_errno_with_message!(errno, "new_name is . or ..");
+                return_errno_with_message!(Errno::EBUSY, "new_name is . or ..");
+            }
         }
 
         let old_dir_inode = old_dir.inode();
@@ -767,7 +770,6 @@ impl DirDentry<'_> {
 
             let old_dentry =
                 old_dir.resolve_child_for_rename(&mut children, old_name, old_dir_arc)?;
-
             let new_dentry =
                 match old_dir.resolve_child_for_rename(&mut children, new_name, old_dir_arc) {
                     Ok(dentry) => Some(dentry),
@@ -828,7 +830,6 @@ impl DirDentry<'_> {
 
             let old_dentry =
                 old_dir.resolve_child_for_rename(&mut old_children, old_name, old_dir_arc)?;
-
             let new_dentry =
                 match new_dir.resolve_child_for_rename(&mut new_children, new_name, new_dir_arc) {
                     Ok(dentry) => Some(dentry),
@@ -949,7 +950,7 @@ impl DirDentry<'_> {
 
         return_errno_with_message!(
             Errno::EBUSY,
-            "failed to generate an unused temporary name for exchange"
+            "no temporary file name for exachange is available"
         )
     }
 
@@ -1125,6 +1126,7 @@ impl DentryChildren {
         self.entries.get(name).cloned()
     }
 
+    /// Returns whether the cache contains the positive dentry by name.
     fn contains_positive(&self, name: &str) -> bool {
         self.entries
             .get(name)
diff --git a/kernel/src/fs/vfs/path/mod.rs b/kernel/src/fs/vfs/path/mod.rs
index a0d9e6552..06b906f99 100644
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -27,7 +27,10 @@ use crate::{
         },
     },
     prelude::*,
-    process::{Gid, Uid, credentials::capabilities::CapSet, posix_thread::AsPosixThread},
+    process::{
+        Gid, Uid, UserNamespace, credentials::capabilities::CapSet, posix_thread::AsPosixThread,
+    },
+    security::lsm::hooks as lsm_hooks,
 };
 
 mod dentry;
@@ -35,7 +38,7 @@ mod mount;
 mod mount_namespace;
 mod resolver;
 
-/// Specifies the behavior of `renameat2()` when renaming a path.
+/// The behavior of [`Path::rename`].
 pub enum RenameMode {
     /// Replaces the destination if it already exists.
@@ -283,17 +286,25 @@ impl Path {
 
     /// Checks whether an inode may be used as the source of a hard link.
     fn check_hardlink_source(&self) -> Result<()> {
-        let Some(creds) = current_thread!()
-            .as_posix_thread()
-            .map(|thread| thread.credentials())
-        else {
+        let current_thread = current_thread!();
+        let Some(posix_thread) = current_thread.as_posix_thread() else {
             return Ok(());
         };
 
         let inode = self.inode();
         let metadata = inode.metadata();
 
-        if metadata.uid == creds.fsuid() || creds.effective_capset().contains(CapSet::FOWNER) {
+        let creds = posix_thread.credentials();
+        if metadata.uid == creds.fsuid() {
+            return Ok(());
+        }
+        if lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+            UserNamespace::get_init_singleton().as_ref(),
+            posix_thread,
+            CapSet::FOWNER,
+        ))
+        .is_ok()
+        {
             return Ok(());
         }
 
@@ -308,7 +319,10 @@ impl Path {
             return Ok(());
         }
 
-        return_errno_with_message!(Errno::EPERM, "hard link source is not permitted");
+            Errno::EPERM,
+            "the source is not permitted to be hard-linked"
+        );
     }
 }
 
diff --git a/kernel/src/syscall/rename.rs b/kernel/src/syscall/rename.rs
index 2a12fdf51..0efa0e7b1 100644
--- a/kernel/src/syscall/rename.rs
+++ b/kernel/src/syscall/rename.rs
@@ -26,7 +26,7 @@ pub fn sys_renameat2(
         old_dirfd, old_path_name, new_dirfd, new_path_name
     );
     let Some(flags) = Flags::from_bits(flags) else {
-        return_errno_with_message!(Errno::EINVAL, "invalid flags");
+        return_errno_with_message!(Errno::EINVAL, "invalid renameat2 flags");
     };
     // TODO: Add support for handling the `WHITEOUT` flag.
     if flags.contains(Flags::WHITEOUT) || flags.contains(Flags::NOREPLACE | Flags::EXCHANGE) {
diff --git a/kernel/src/syscall/unlink.rs b/kernel/src/syscall/unlink.rs
index e26c205f8..395539b9c 100644
--- a/kernel/src/syscall/unlink.rs
+++ b/kernel/src/syscall/unlink.rs
@@ -30,22 +30,25 @@ pub fn sys_unlinkat(
         return_errno_with_message!(Errno::EISDIR, "unlink on root directory");
     }
 
-    let fs_ref = ctx.thread_local.borrow_fs();
-    let path_resolver = fs_ref.resolver().read();
+    let (dir_path, target, name) = {
+        let fs_ref = ctx.thread_local.borrow_fs();
+        let path_resolver = fs_ref.resolver().read();
 
-    let (dir_path, name) = {
         let (parent_path_name, target_name) = path_name.split_dirname_and_basename()?;
         let fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
-        (path_resolver.lookup(&fs_path)?, target_name)
+        let dir = path_resolver.lookup(&fs_path)?;
+        let target = path_resolver.lookup_at_path(&dir_path, target_name)?;
+        (dir, target, target_name)
     };
-    let target = path_resolver.lookup_at_path(&dir_path, name)?;
-    drop(path_resolver);
 
     if target.type_() == InodeType::Dir {
-        return_errno_with_message!(Errno::EISDIR, "the path refers to a directory");
+        return_errno_with_message!(Errno::EISDIR, "the path is a directory");
     }
     if path_name.ends_with('/') {
-        return_errno_with_message!(Errno::ENOTDIR, "trailing slash but not a directory");
+            Errno::ENOTDIR,
+            "the path ends with a slash but is not a directory"
+        );
     }
 
     dir_path.unlink(name)?;
````
</details>

**Review by cchanging — APPROVED** [review]

File: kernel/src/fs/vfs/path/mod.rs [hunk 5]
```diff
@@ -266,6 +257,74 @@ impl Path {
     fn this(&self) -> Self {
         self.clone()
     }
+
+    /// Checks whether a directory entry may be modified in this directory.
+    fn check_modify_permission(&self) -> Result<()> {
```
lrh2000: Someone will want to call this at `resize()`, `set_atime()`, e.t.c. if you name this method `check_modify_permission`.

This method should be renamed to something like `check_dir_modification`?

---

And we should have `check_file_modification` in the future? Though it's outside the scope of this PR.

In addition, `open(O_WRONLY)` and `open(O_RDWR)` should also report `EROFS`...

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -266,6 +257,74 @@ impl Path {
...
+    fn check_hardlink_source(&self) -> Result<()> {
+        let current_thread = current_thread!();
+        let Some(posix_thread) = current_thread.as_posix_thread() else {
+            return Ok(());
+        };
+
+        let metadata = self.metadata();
+        let fsuid = posix_thread.credentials().fsuid();
+
+        // Source inode owner can hardlink all they like.
```
lrh2000:
```suggestion
        // The source inode owner can hardlink all they like.
```

kernel/src/syscall/unlink.rs
```diff
@@ -26,19 +26,31 @@ pub fn sys_unlinkat(
     debug!("dirfd = {}, path = {:?}", dirfd, path_name);
 
     let path_name = path_name.to_string_lossy();
-    let (dir_path, name) = {
-        let (parent_path_name, target_name) = path_name.split_dirname_and_filename()?;
+    if path_name.starts_with('/') && path_name.trim_end_matches('/').is_empty() {
+        return_errno_with_message!(Errno::EISDIR, "unlink on root directory");
+    }
```
lrh2000: This is also a bit odd. I questioned myself several times about the necessity of this, and later figured out that it's to avoid the wrong error code in `split_dirname_and_basename`.

Should we fix `split_dirname_and_basename` instead? `EBUSY` isn't a natural error code from a string splitting method, but it was previously chosen for simplicity. However, it will cause more confusion now.

```patch
diff --git a/kernel/src/fs/vfs/path/resolver.rs b/kernel/src/fs/vfs/path/resolver.rs
index 68ed1dc93..394aa556c 100644
--- a/kernel/src/fs/vfs/path/resolver.rs
+++ b/kernel/src/fs/vfs/path/resolver.rs
@@ -930,8 +930,8 @@ pub trait SplitPath {
     /// If the original path is an empty string, an error will be returned.
     ///
     /// If the original path directly points to the root directory (e.g., `/` or `//`, but not `/.`
-    /// or `/./`), an error will be returned.
-    fn split_dirname_and_basename(&self) -> Result<(&Self, &Self)>;
+    /// or `/./`), `None` will be returned.
+    fn split_dirname_and_basename(&self) -> Result<Option<(&Self, &Self)>>;
 }
 
 impl SplitPath for str {
diff --git a/kernel/src/syscall/rename.rs b/kernel/src/syscall/rename.rs
index b1b134b15..5f1b13e28 100644
--- a/kernel/src/syscall/rename.rs
+++ b/kernel/src/syscall/rename.rs
@@ -39,7 +39,9 @@ pub fn sys_renameat2(
 
     let old_path_name = old_path_name.to_string_lossy();
     let (old_parent_path, old_name) = {
-        let (old_parent_path_name, old_name) = old_path_name.split_dirname_and_basename()?;
+        let Some((old_parent_path_name, old_name)) = old_path_name.split_dirname_and_basename()? else {
+            return_errno_with_message!(Errno::EBUSY, "the root directory cannot be renamed from");
+        };
         let old_fs_path =
             FsPath::from_fd_at(old_dirfd, old_parent_path_name, EmptyPathStr::Reject)?;
         (path_resolver.lookup(&old_fs_path)?, old_name)
@@ -54,7 +56,9 @@ pub fn sys_renameat2(
         if old_path.type_() != InodeType::Dir && new_path_name.ends_with('/') {
             return_errno_with_message!(Errno::EISDIR, "the new path is a directory");
         }
-        let (new_parent_path_name, new_name) = new_path_name.split_dirname_and_basename()?;
+        let Some((new_parent_path_name, new_name)) = new_path_name.split_dirname_and_basename()? else {
+            return_errno_with_message!(Errno::EBUSY, "the root directory cannot be renamed to");
+        };
         let new_parent_fs_path =
             FsPath::from_fd_at(new_dirfd, new_parent_path_name, EmptyPathStr::Reject)?;
         (
diff --git a/kernel/src/syscall/rmdir.rs b/kernel/src/syscall/rmdir.rs
index a55b66f52..d7156fd18 100644
--- a/kernel/src/syscall/rmdir.rs
+++ b/kernel/src/syscall/rmdir.rs
@@ -24,7 +24,9 @@ pub(super) fn sys_rmdirat(
 
     let path_name = path_name.to_string_lossy();
     let (dir_path, name) = {
-        let (parent_path_name, target_name) = path_name.split_dirname_and_basename()?;
+        let Some((parent_path_name, target_name)) = path_name.split_dirname_and_basename()? else {
+            return_errno_with_message!(Errno::EBUSY, "the root directory cannot be removed");
+        };
         let fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
         (
             ctx.thread_local
diff --git a/kernel/src/syscall/unlink.rs b/kernel/src/syscall/unlink.rs
index c3ab68f32..76dc03340 100644
--- a/kernel/src/syscall/unlink.rs
+++ b/kernel/src/syscall/unlink.rs
@@ -27,7 +27,9 @@ pub fn sys_unlinkat(
 
     let path_name = path_name.to_string_lossy();
     let (dir_path, name) = {
-        let (parent_path_name, target_name) = path_name.split_dirname_and_filename()?;
+        let Some((parent_path_name, target_name)) = path_name.split_dirname_and_base()? else {
+            return_errno_with_message!(Errno::EISDIR, "the root directory cannot be unlinked");
+        };
         let fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
         (
             ctx.thread_local

```

tatetian: Yes, I agree with @lrh2000 that special handling for `/` or `////` should be removed as `split_dirname_and_filename` already takes care of this edge case and that returning `EBUSY` is silly.

But personally, I always frown upon returning `Result<Option<T>>`. For this specific case, I think `split_dirname_and_basename` should return `Result<(&Self, &Self), SplitPathError>`, where `SplitPathError` is defined as:

```rust
pub enum SplitPathError {
    /// The path to be split is empty (``).
    PathIsEmpty,
    /// The path to be split is root
    /// (any number of consecutive slashes such `/` or `////`).
    PathIsRoot,
}
```

Let the caller to decide how to map a `SplitPathError` to `Errno`.

kernel/src/syscall/unlink.rs
```diff
@@ -26,19 +26,31 @@ pub fn sys_unlinkat(
...
-        )
+        let dir = path_resolver.lookup(&fs_path)?;
+        let target = path_resolver.lookup_at_path(&dir, target_name)?;
+        (dir, target, target_name)
     };
 
+    if target.type_() == InodeType::Dir {
+        return_errno_with_message!(Errno::EISDIR, "the path is a directory");
+    }
+    if path_name.ends_with('/') {
```
tatetian: I am surprised to see this check is placed after doing the path split and resolution. This check can be done upfront at the beginning of this function. If this is done on purpose, add a comment as it is worthy explaining.

kernel/src/syscall/unlink.rs
```diff
@@ -26,19 +26,31 @@ pub fn sys_unlinkat(
...
+    if path_name.starts_with('/') && path_name.trim_end_matches('/').is_empty() {
+        return_errno_with_message!(Errno::EISDIR, "unlink on root directory");
+    }
+
+    let (dir_path, target, name) = {
+        let fs_ref = ctx.thread_local.borrow_fs();
+        let path_resolver = fs_ref.resolver().read();
+
+        let (parent_path_name, target_name) = path_name.split_dirname_and_basename()?;
         let fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
```
tatetian: Both `sys_unlinkat` and `sys_rmdirat` syscall adopt the following pattern of `split_dirname_and_filename` + `from_fd_at`:

```rust
        let (parent_path_name, target_name) = path_name.split_dirname_and_filename()?;
        let fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
```

But `mkdirat` adopts the pattern of `from_fd_at` + `into_parent_and_basename`:

https://github.com/asterinas/asterinas/blob/42d38f9af388c28c2dda8cb4d6e62de465f23e8e/kernel/src/syscall/mkdir.rs#L24-L32

That the latter one is the recommended pattern since `EmptyPathStr` is designed to tell if the empty path is allowed at the _syscall_ level. But in the `split_dirname_and_filename` + `from_fd_at` pattern, the empty path is rejected by `split_dirname_and_filename`, not by `from_fd_at`. This causes some confusion.

kernel/src/fs/vfs/path/mod.rs
tatetian: I have a similar concern. I thought `check_modify_permission` applies to other setter methods inherited from `Inode` as well:

```
impl Path {
    pub fn sync_all(&self) -> Result<()>;
    pub fn sync_data(&self) -> Result<()>;
    pub fn set_mode(&self, mode: InodeMode) -> Result<()>;
    pub fn resize(&self, size: usize) -> Result<()>;
    ...
}
```

And the following checks


```rust
        if self.mount.flags().contains(PerMountFlags::RDONLY)
            || self.fs().flags().contains(FsFlags::RDONLY)
```

appear to be applicable to regular files as well.

cqs21: I split the `EROFS` check out into a `check_mount_writable` helper, `check_dir_entry_mutation` then builds on top of it and adds the directory-specific `W|X` permission check that is required for mutating directory entries.

For file-side mutations, the situation is more complicated.

Operations such as `open(O_WRONLY)`, `open(O_RDWR)`, truncation, xattr updates, and metadata changes should also reject read-only mounts with `EROFS`, so they should eventually reuse `check_mount_writable`. However, beyond the `EROFS` part, their permission and policy checks are not uniform: some are about file data writes, some are about inode metadata, some have xattr-specific rules, and some also depend on DAC/capability/LSM checks. Because of that, I did not introduce a generic `check_file_data_mutation` helper in this PR...

kernel/src/syscall/unlink.rs
cqs21: The `unlinkat` / `rmdirat` cannot directly reuse `mkdirat`'s `lookup_unresolved_no_follow` + `into_parent_and_basename` chain. That chain is suitable for creation syscalls because it expects the final component to be absent and returns `AtParent` in that case. By contrast, `unlinkat` / `rmdirat` normally operate on an existing final component, so the same lookup would often return `Resolved(target)` instead of a parent+basename pair.

I agree with the direction here. In an earlier version of this PR, I tried adding a `LookupFlags::Parent`-style mode, similar to Linux `LOOKUP_PARENT` (<https://elixir.bootlin.com/linux/v6.17.4/source/include/linux/namei.h#L31>), so the caller could first resolve the path and then convert the lookup result into a parent+basename pair.

 That approach can make the pattern closer to `mkdirat`, but it also requires restructuring the lookup flow. For this fix PR, that refactor felt larger than necessary, since the main goal here is to correct the `unlinkat` behavior. I think it makes sense to revisit this if we later want broader support for more `LookupFlags`-style semantics. At that point, a dedicated resolver refactor would be more justified.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -913,44 +913,41 @@ impl<'a> TryFrom<&'a str> for FsPath<'a> {
...
+///
+/// These errors intentionally describe only what the path string looks like.
+/// Callers are responsible for translating them into syscall- or subsystem-specific `Errno`s.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum SplitPathError {
+    /// The path to be split is empty (``).
+    PathIsEmpty,
+    /// The path to be split is root
+    /// (any number of consecutive slashes such `/` or `////`).
+    PathIsRoot,
```
lrh2000:
```suggestion
    /// The path to be split is empty (``).
    Empty,
    /// The path to be split is root
    /// (any number of consecutive slashes such `/` or `////`).
    Root,
```

kernel/src/syscall/rename.rs
```diff
@@ -39,7 +39,18 @@ pub fn sys_renameat2(
...
+                .split_dirname_and_basename()
+                .map_err(|error| match error {
+                    SplitPathError::PathIsEmpty => {
+                        Error::with_message(Errno::ENOENT, "the old path is empty")
+                    }
+                    SplitPathError::PathIsRoot => Error::with_message(
+                        Errno::EBUSY,
+                        "the root directory cannot be renamed from",
+                    ),
+                })?;
```
lrh2000:
```suggestion
                .map_err(SplitPathError::reject_root_as_busy)?;
```

I am not sure whether this is the best naming. Do you have a better suggestion?
```rust
impl SplitPathError {
    pub fn reject_root_as_busy(self) -> Error;
    pub fn reject_root_as_invalid(self) -> Error;
}
```

kernel/src/syscall/unlink.rs
```diff
@@ -27,16 +27,39 @@ pub fn sys_unlinkat(
...
+        let parent_fs_path = FsPath::from_fd_at(dirfd, parent_path_name, EmptyPathStr::Reject)?;
+        let dir_path = path_resolver.lookup(&parent_fs_path)?;
+
+        // Linux resolves the full unlink path before removing the parent directory entry.
+        // This preserves path-walk `ENOTDIR` errors for trailing slashes and dot components,
+        // such as `file/`, `file/.`, and `file/..`, instead of collapsing the request into
+        // unlinking an entry in the parent directory.
+        let target = path_resolver.lookup_at_path(&dir_path, target_name)?;
+        if target.type_() == InodeType::Dir {
+            return_errno_with_message!(Errno::EISDIR, "cannot unlink a directory");
```
lrh2000:
```rust
        // Linux resolves the full unlink path before removing the parent directory entry.
        // This preserves path-walk `ENOTDIR` errors for trailing slashes and dot components,
        // such as `file/`, `file/.`, and `file/..`, instead of collapsing the request into
        // unlinking an entry in the parent directory.
```

I found this comment is quite confusing in several ways. Please try to rewrite it.

> removing the parent directory entry

What does it want to mean? "parent"? Why should we remove the parent directory?

> `file/.`, and `file/..`

Are they related here?

---

The code is also racy and involves unnecessary checks for the common case. I think you want to mean:
```rust
if path_name.ends_with('/') {
    let target = path_resolver.lookup_at_path(&dir_path, target_name)?;
    if target.type_() == InodeType::Dir {
        return_errno_with_message!(Errno::EISDIR, "a directory cannot be unlinked");
    } else {
        return_errno_with_message!(
            Errno::ENOTDIR,
            "the path ends with a slash but is not a directory"
        );
    }
}
```

Note that this logic should not be executed if the path name does not end with "/". Conversely, if the path name does end in "/", it should always fail early and return immediately to avoid racing with a latter `dir_path.unlink()`.

cqs21: > Note that this logic should not be executed if the path name does not end with "/". 

Thanks for the suggestion! 

You're right that this `lookup_at_path` is mainly here to handle the trailing-`/` case. However, it also covers `file/.` and `file/..`, which do not end with `/`. If we gate the whole block behind `if path_name.ends_with('/')`, these cases fall through to the `dir_path.unlink` call below. There, `check_dir_entry_mutation`  returns `EACCES` instead of the expected `ENOTDIR`:

```
[ RUN      ] UnlinkTest.CannotRemoveDots
test/syscalls/linux/unlink.cc:195: Failure
Value of: unlink(self.c_str())
Expected: -1 (failure), with errno PosixError(errno=20 Not a directory)
  Actual: -1 (of type int), with errno PosixError(errno=13 Permission denied)
```

But this can also be fixed by:

```rust
pub fn unlink(&self, name: &str) -> Result<()> {
    let dir_dentry = self.dentry.as_dir_dentry_or_err()?;
    self.check_dir_entry_mutation()?;
    dir_dentry.unlink(name)
}
```

lrh2000: > ```rust
> pub fn unlink(&self, name: &str) -> Result<()> {
>     let dir_dentry = self.dentry.as_dir_dentry_or_err()?;
>     self.check_dir_entry_mutation()?;
>     dir_dentry.unlink(name)
> }
> ```

To be honest, I think all `check_dir_entry_mutation` should be called after `as_dir_dentry_or_err`...

Logically, if you haven't checked whether the path is a directory, why can you check whether the directory can be mutated?

kernel/src/net/socket/unix/ns/path.rs
```diff
@@ -38,23 +38,24 @@ pub fn lookup_socket_file(path: &str) -> Result<Path> {
...
-                err
-            }
-        })
+        let fs_path = FsPath::try_from(path_name)?;
+        let (parent, file_name) = path_resolver
+            .lookup_unresolved_no_follow(&fs_path)?
+            .into_parent_and_filename()?;
+
+        parent.new_fs_child(&file_name, InodeType::Socket, mkmod!(u+rw))
+    })();
```
lrh2000, line 52: Why is this a closure?

Also, it does not seem to preserve the original errno if the path ends with `/`. Is the new errno correct?

cqs21: Both good points, thanks.

> Why is this a closure?

It lets the `EEXIST → EADDRINUSE` mapping apply to the whole operation rather than just `new_fs_child`. After the refactor, `EEXIST` can come from `into_parent_and_filename()` too. Mapping just `new_fs_child` would leak a raw `EEXIST` for the common already-exists case.

> Also, it does not seem to preserve the original errno if the path ends with /. Is the new errno correct?

You're right that the errno changed: the old `split_dirname_and_filename` returned `EISDIR` for any trailing slash, whereas the new path returns `ENOENT` / `EADDRINUSE`. I verified the behavior against Linux 6.8:

```
bind("nonexist_dir/")  -> ENOENT
bind("nonexist_file/") -> ENOENT
bind("existing_dir/")  -> EADDRINUSE
```

The new code matches Linux for these: `ENOENT` from `into_parent_and_filename`, and `EADDRINUSE` from the mapped `EEXIST`. So the old `EISDIR` was incorrect.

There is one remaining edge case that still diverges from Linux:

```
bind("existing_file/") -> Linux: EADDRINUSE,  Asterinas: ENOTDIR
```

Linux lets "the path already exists" (`EEXIST`) win over the trailing-slash directory requirement, so it reports `EADDRINUSE` regardless of whether the existing entry is a directory or a regular file. Our shared `lookup_from_parent` instead enforces the trailing-slash "must be a directory" check on the resolved tail and returns `ENOTDIR`.

Fixing it cleanly means letting create-style lookups defer the tail directory check to the consumer (so existence maps to `EEXIST`) while keeping `ENOTDIR` for pure lookups like `open("file/")`; that touches the shared resolver and is orthogonal to this PR, so I'd prefer to fix it as a follow-up.

lrh2000: Thanks for your detailed explanation!

kernel/src/syscall/unlink.rs
lrh2000: Just to be clear. In case that you miss what I mean here.

What I mean here is that we may also try to fix other methods:
https://github.com/asterinas/asterinas/blob/9eb51bc53265d92c00181f43057b78396c03ae70/kernel/src/fs/vfs/path/mod.rs#L587-L633

Instead of only addressing the `unlink` method.

> To be honest, I think all check_dir_entry_mutation should be called after as_dir_dentry_or_err...

(Unless, of course, you can show that we shouldn't fix them because doing so would produce incorrect error codes for corner cases.)

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -86,13 +80,7 @@ impl Path {
...
-        if self
-            .inode()
-            .check_permission(Permission::MAY_WRITE)
-            .is_err()
-        {
-            return_errno!(Errno::EACCES);
-        }
+        self.check_dir_entry_mutation()?;
         let tmp_inode = self.inode().create_tmpfile(mode, hard_linkability)?;
         let tmp_dentry = Dentry::new_anonymous(tmp_inode, self.dentry.clone());
```
lrh2000: This should accept `dir_dentry` as well. Right?
```rust
        let dir_dentry = self.as_dir_dentry_or_err()?;
        self.check_dir_entry_mutation()?;
        let tmp_inode = self.inode().create_tmpfile(mode, hard_linkability)?;
        let tmp_dentry = Dentry::new_anonymous(tmp_inode, &dir_dentry);
```

lrh2000 (approved):
Otherwise LGTM. Thanks for your continuous efforts.

See if @tatetian has more comments?

kernel/src/fs/vfs/path/mod.rs
cqs21: The caller already rejects non-directories before calling it, so `as_dir_dentry_or_err` is not required for correctness.
https://github.com/asterinas/asterinas/blob/56bf21533b4278663e817b124f4dc9f4afcdc7d5/kernel/src/syscall/open.rs#L140-L145

lrh2000: Yes. On the other hand, from an API design perspective, it would be better to:
 1. have the `Path::create_tmpfile` function check if the path refers to a directory instead of relying on the system call layer. I don't foresee any problems arising if we remove the (redundant) system call layer.
 2. have the `Dentry::new_anonymous` function accept a `DirDentry` so that it ensures the directory check has been performed by the caller. It does not make sense to create a (anonymous) dentry under a normal file.

However, this is mostly an existing problem. Feel free to close this thread if you think that adjusting the API is outside the scope of this PR.

tatetian (approved):

kernel/src/syscall/unlink.rs
tatetian: Could you create an issue and assign yourself to follow up this improvement direction?

PR #3154, opened 2026-04-17
Remove `log!` from prelude & Remove `#[macro_export]` in `aster-kernel`

lrh2000:
1. Remove `log!` from prelude
   - It does not seem useful. It has no users. Since a user needs to manually import the `LogLevel` type anyway, `log!` should be imported manually as well.
2. Remove `#[macro_export]` in `aster-kernel`
    - `aster-kernel` has no downstream users. So I don't think having `#[macro_export]` in it makes any sense..

### DIFF
--- a/kernel/src/prelude.rs
+++ b/kernel/src/prelude.rs
@@ -16,7 +16,7 @@ pub(crate) use core::{any::Any, ffi::CStr, fmt::Debug};
 pub(crate) use bitflags::bitflags;
 pub(crate) use int_to_c_enum::TryFromInt;
 pub(crate) use ostd::{
-    alert, crit, debug, emerg, error, info, log,
+    alert, crit, debug, emerg, error, info,
     mm::{FallibleVmRead, FallibleVmWrite, PAGE_SIZE, Vaddr, VmReader, VmWriter},
     notice,
     sync::{Mutex, MutexGuard, RwLock, RwMutex, SpinLock, SpinLockGuard},
--- a/kernel/src/prelude.rs
+++ b/kernel/src/prelude.rs
@@ -13,6 +13,7 @@ pub(crate) use alloc::{
 };
 pub(crate) use core::{any::Any, ffi::CStr, fmt::Debug};
 
+pub(crate) use aster_logger::{print, println};
 pub(crate) use bitflags::bitflags;
 pub(crate) use int_to_c_enum::TryFromInt;
 pub(crate) use ostd::{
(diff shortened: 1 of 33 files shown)


kernel/src/prelude.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) use alloc::{
 };
 pub(crate) use core::{any::Any, ffi::CStr, fmt::Debug};
 
+pub(crate) use aster_logger::{print, println};
 pub(crate) use bitflags::bitflags;
 pub(crate) use int_to_c_enum::TryFromInt;
```
tatetian: All the macro rules touched by this PR are marked `pub(crate)`. In the kernel crate, `pub` and `pub(crate)` have no real difference; so the current code has chosen `pub` overwhelmingly. The `prelude.rs` file is one major exception. I think we should stick with `pub` given that it is the majority. The `prelude.rs` file should be rewritten to use `pub` rather than `pub(crate)`.

tatetian (commented):
Just one issue.

kernel/src/prelude.rs
lrh2000: We cannot mark macros as `pub` because it will cause compiler errors:
```
error[E0364]: `current_thread` is only public within the crate, and cannot be re-exported outside
   --> kernel/src/context.rs:317:9
    |
317 | pub use current_thread;
    |         ^^^^^^^^^^^^^^
    |
help: consider adding a `#[macro_export]` to the macro in the imported module
   --> kernel/src/context.rs:311:1
    |
311 | / macro_rules! current_thread {
312 | |     () => {
313 | |         $crate::thread::Thread::current().expect("the current task is not associated with a thread")
314 | |     };
315 | | }
    | |_^
help: in case you want to use the macro within this crate only, reduce the visibility to `pub(crate)`
    |
317 | pub(crate) use current_thread;
    |    +++++++
```

`context::CurrentUserspace` cannot be marked as `pub` due to this FIXME:
https://github.com/asterinas/asterinas/blob/5b93972e6fd61efcca18772431a92313005d2d6a/kernel/src/context.rs#L44-L50

Otherwise, I agree with you. So I remove all remaining `pub(crate)` in `aster-kernel`.

Note that the initialization methods of the top-level modules sometimes use `pub(super)` and sometimes use `pub`. Resolving this inconsistency is beyond the scope of this PR.
https://github.com/asterinas/asterinas/blob/5b93972e6fd61efcca18772431a92313005d2d6a/kernel/src/fs/mod.rs#L23-L26
https://github.com/asterinas/asterinas/blob/5b93972e6fd61efcca18772431a92313005d2d6a/kernel/src/vdso.rs#L364-L366

tatetian (approved):

PR #3155, opened 2026-04-18
Update Rust guidelines on visibility modifiers

tatetian:
Codifies the style preference from #3154 as a coding guideline: inside `aster-kernel`, `pub(crate)` and `pub` are equivalent since the crate has no downstream consumers, so prefer the shorter `pub`. Also refreshes the "See also" links with related PRs.


tatetian:
Merging a trivial change to the coding guideline

PR #3158, opened 2026-04-20
Fix LTP `readlink` test cases

cqs21:
This PR fixes LTP `readlink` test failures for two edge cases:
- return `EINVAL` when `readlink()` is called with a zero-length buffer;
- support `readlinkat(fd, "")` on a symlink opened with `O_PATH | O_NOFOLLOW`.

It also enables the related LTP `readlink*` test cases and blocks them on `exfat`.

### DIFF
--- a/kernel/src/syscall/readlink.rs
+++ b/kernel/src/syscall/readlink.rs
@@ -22,6 +22,10 @@ pub fn sys_readlinkat(
     usr_buf_len: usize,
     ctx: &Context,
 ) -> Result<SyscallReturn> {
+    if usr_buf_len == 0 {
+        return_errno_with_message!(Errno::EINVAL, "the buffer length is zero");
+    }
+
     let user_space = ctx.user_space();
     let path_name = user_space.read_cstring(path_addr, MAX_FILENAME_LEN)?;
     debug!(
@@ -34,7 +38,11 @@ pub fn sys_readlinkat(
         let path_resolver = fs_ref.resolver().read();
 
         let path_name = path_name.to_string_lossy();
-        let fs_path = FsPath::from_fd_and_path(dirfd, &path_name)?;
+        let fs_path = if path_name.is_empty() && dirfd != AT_FDCWD {
+            FsPath::from_fd(dirfd)?
+        } else {
+            FsPath::from_fd_and_path(dirfd, &path_name)?
+        };
         let path = path_resolver.lookup_no_follow(&fs_path)?;
 
         match path.inode().read_link()? {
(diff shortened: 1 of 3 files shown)


kernel/src/syscall/readlink.rs
```diff
@@ -29,12 +29,20 @@ pub fn sys_readlinkat(
         dirfd, path_name, usr_buf_addr, usr_buf_len
     );
 
+    if usr_buf_len == 0 {
+        return_errno_with_message!(Errno::EINVAL, "the buffer length is zero");
+    }
```
tatetian: **Place the `usr_buf_len == 0` check before `read_cstring`** (minor): Per Linux's `do_readlinkat` (`fs/stat.c`), `bufsiz <= 0` is rejected before any path work. The current placement runs `read_cstring` and the `debug!` first, so a call with `usr_buf_len == 0` and a bogus `path_addr` returns `EFAULT` instead of `EINVAL` — a detectable divergence from Linux even though the LTP cases happen to pass valid path pointers. Move the guard to the very top of the function:

```rust
pub fn sys_readlinkat(
    dirfd: RawFileDesc,
    path_addr: Vaddr,
    usr_buf_addr: Vaddr,
    usr_buf_len: usize,
    ctx: &Context,
) -> Result<SyscallReturn> {
    if usr_buf_len == 0 {
        return_errno_with_message!(Errno::EINVAL, "the buffer length is zero");
    }

    let path_name = user_space.read_cstring(path_addr, MAX_FILENAME_LEN)?;
    ...
}
```

test/initramfs/src/conformance/ltp/testcases/all.txt
```diff
@@ -1196,12 +1196,11 @@ read04
...
 
-# readlink01
-# readlink03
+readlink01
+readlink03
 
-#readlinkat test cases
-# readlinkat01
-# readlinkat02
+readlinkat01
```
tatetian, line 1203: **Preserve the `#readlinkat test cases` section header** (minor): The surrounding file convention keeps per-syscall section-header comments (`#symlinkat test cases`, `#linkat test cases`, `#mkdirat test cases`, `#renameat test cases`, `#unlinkat test cases`) even when the tests below them are enabled — they act as group labels. Dropping `#readlinkat test cases` while enabling the tests diverges from that pattern. Restore the header:

```
 readlink01
 readlink03

+# readlinkat test cases
 readlinkat01
 readlinkat02
```

tatetian (commented):
Thanks for the fix. There are two minor issues to address.

A follow-up PR is needed to handle all at-family syscall more gracefully. See the design proposal, #3172.

tatetian (approved):

PR #3160, opened 2026-04-20
Model `virtqueue` as untrusted and use fallible allocation in `aster-virtio`

lrh2000:
The first few commits are clean-up commits.

The following two sections describe *the last two commits*:

## Commit: Specify virtqueue API contract

The current API design and the implementation of `VirtQueue` assumes that the queue can be trusted. For example, it stores some indexes in a DMA-coherent region:
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/comps/virtio/src/queue.rs#L178-L186

If the device writes garbage to the region, our kernel will crash at `&self.descs[self.free_head as usize]`.

The related APIs, such as `pop_used`, do not state whether the returned token or the buffer length is valid. Therefore, it is unclear whether the queue or the device should verify this.

This PR proposes the following API design:
```rust
    /// Adds input and output DMA buffers to the virtqueue and returns a token.
    ///
    /// This method will succeed if the arguments contain at least one buffer and there is
    /// sufficient space in the queue. If the above condition is met, users can then unwrap the
    /// result.
    ///
    /// When successful, the result token is guaranteed to be valid. It will not exceed the queue
    /// size, and the same token will not be returned twice, unless it has been removed from the
    /// queue by [`Self::pop_used`] in the meantime.
    pub fn add_dma_bufs<I: DmaBuf, O: DmaBuf>(
        &mut self,
        inputs: &[&I],
        outputs: &[&O],
    ) -> Result<u16, QueueError>;

    /// Pops a device-used buffer and returns the token and the buffer length.
    ///
    /// When successful, the following is guaranteed:
    /// - The token is valid.  It will not exceed the queue size. It was previously returned by
    ///   [`Self::add_dma_bufs`], [`Self::add_input_bufs`], or [`Self::add_output_bufs`], and it has
    ///   not yet been removed from the queue by this method.
    /// - The length is valid. It will not exceed the length of the original DMA buffer.
    ///
    /// If the device malfunctions, it may report a token or length that violates these guarantees.
    /// Such reports are logged as errors and ignored; the reported token is not returned to the
    /// caller, preventing an invalid token from corrupting upper-layer state. If the device
    /// continues to malfunction, the queue may become stuck because the affected buffer cannot be
    /// reclaimed.
    pub fn pop_used(&mut self) -> Result<(u16, u32), QueueError>;

    /// Pops a device-used buffer containing a header and returns the token and the buffer length.
    ///
    /// This is the same as [`Self::pop_used`], except it also guarantees that the used buffer is at
    /// least `header_len` bytes.
    ///
    /// The note about malfunctioning devices in [`Self::pop_used`] applies here as well, including
    /// extra cases where the used buffer is shorter than `header_len`.
    pub fn pop_used_with_header(&mut self, header_len: usize) -> Result<(u16, u32), QueueError>;
``` 

## Commit: Make queue/memory allocation faillible 

This PR also avoids `unwrap()` on memory allocation APIs, including the two most common scenarios:
 - Allocating a `DmaStream`/`DmaCoherent`.
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/comps/virtio/src/device/console/device.rs#L100
 - Allocating a `TxBuffer`/`RxBuffer`.
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/comps/virtio/src/device/network/device.rs#L96

### Notes

 1. This PR intentionally skips all the code under `device/socket`. I will address this in my vsock PR.
 2. APIs that cannot fail should continue to use `unwrap()`. These include the following:
   - `DmaStream::reader()`/`DmaStream::writer()`/`DmaStream::sync_from_device()`/`DmaStream::sync_to_device()`.
   - `SafePtr::read_once()`/`SafePtr::write_once()`.
   - `VirtQueue::add_dma_bufs()`/`VirtQueue::add_input_bufs()`/`VirtQueue::add_output_bufs()` (if the caller knows that it has enough space for the new buffers).
   - `VirtioTransport::register_queue_callback()`/`VirtioTransport::register_cfg_callback()`.
   - `Once::get` (for use after initialization methods).

### DIFF
--- a/kernel/comps/virtio/src/queue.rs
+++ b/kernel/comps/virtio/src/queue.rs
@@ -184,6 +184,16 @@ impl VirtQueue {
         })
     }
 
+    /// Adds only input DMA buffers to the virtqueue and returns a token.
+    pub fn add_input_bufs<I: DmaBuf>(&mut self, inputs: &[&I]) -> Result<u16, QueueError> {
+        self.add_dma_bufs(inputs, &[] as &[&I])
+    }
+
+    /// Adds only output DMA buffers to the virtqueue and returns a token.
+    pub fn add_output_bufs<O: DmaBuf>(&mut self, outputs: &[&O]) -> Result<u16, QueueError> {
+        self.add_dma_bufs(&[] as &[&O], outputs)
+    }
+
     /// Adds input and output DMA buffers to the virtqueue and returns a token.
     ///
     /// Ref: linux virtio_ring.c virtqueue_add
@@ -199,8 +209,9 @@ impl VirtQueue {
             return Err(QueueError::BufferTooSmall);
         }
 
-        // Allocate descriptors from the free list.
         let head = self.free_head;
+
+        // Allocate descriptors from the free list.
         let mut last = self.free_head;
         for input in inputs.iter() {
             let desc = &self.descs[self.free_head as usize];
@@ -234,9 +245,8 @@ impl VirtQueue {
         }
         self.num_used += (inputs.len() + outputs.len()) as u16;
 
-        let avail_slot = self.avail_idx & (self.device_queue_size - 1);
-
         {
+            let avail_slot = self.avail_idx & (self.device_queue_size - 1);
             let ring_ptr: SafePtr<[u16; 64], &Arc<DmaCoherent>> =
                 field_ptr!(&self.avail, AvailRing, ring);
             let mut ring_slot_ptr = ring_ptr.cast::<u16>();
@@ -251,19 +261,10 @@ impl VirtQueue {
         field_ptr!(&self.avail, AvailRing, idx)
             .write_once(&self.avail_idx)
             .unwrap();
-
+        // Write barrier.
         fence(Ordering::SeqCst);
+
         Ok(head)
-    }
-
-    /// Adds only input DMA buffers to the virtqueue and returns a token.
-    pub fn add_input_bufs<I: DmaBuf>(&mut self, inputs: &[&I]) -> Result<u16, QueueError> {
-        self.add_dma_bufs(inputs, &[] as &[&I])
-    }
-
-    /// Adds only output DMA buffers to the virtqueue and returns a token.
-    pub fn add_output_bufs<O: DmaBuf>(&mut self, outputs: &[&O]) -> Result<u16, QueueError> {
-        self.add_dma_bufs(&[] as &[&O], outputs)
     }
 
     /// Returns whether there is a used element that can pop.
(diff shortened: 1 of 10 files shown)


lrh2000:
> 1. This PR intentionally skips all the code under `device/socket`. I will address this in my vsock PR.

Actually, I now have to modify some lines under `device/socket` (vsock code) to fix compiler errors in the commit **Make queue/memory allocation failible**.

However, since I will fix the rewritten version, perhaps there is no need to review those vsock changes in this PR. The changes are for compiler fixes only and may not be 100% reasonable.

---

Additionally, there are other virtio bugs, such as the order of steps during device initialization and certain features negotiated during feature negotiation, such as the queue's features. To control the scope and size of this PR, I am treating those bugs as outside the scope for now.

kernel/comps/virtio/src/queue.rs
```diff
@@ -282,55 +343,96 @@ impl VirtQueue {
...
-        let index = field_ptr!(&element_ptr, UsedElem, id).read_once().unwrap();
-        let len = field_ptr!(&element_ptr, UsedElem, len).read_once().unwrap();
+    /// Pops a device-used buffer containing a header and returns the token and the buffer length.
+    ///
+    /// This is the same as [`Self::pop_used`], except it also guarantees that the used buffer is at
+    /// least `header_len` bytes.
+    ///
+    /// The note about malfunctioning devices in [`Self::pop_used`] applies here as well, including
+    /// extra cases where the used buffer is shorter than `header_len`.
+    pub fn pop_used_with_header(&mut self, header_len: usize) -> Result<(u16, u32), QueueError> {
```
StevenJiang1110: I would suggest renaming it to `pop_used_with_min_bytes`. From the virtqueue's perspective, it should have no notion of what constitutes a header.

What do you think about merging this method into `pop_used`?

```rust
pub fn pop_used(&mut self, min_bytes: Option<usize>) -> Result<(u16, u32), QueueError>
```

I don't have a strong opinion either way — it would just reduce the amount of documentation needed, and from the caller's perspective, it eliminates the need to choose between two separate APIs.

kernel/comps/virtio/src/queue.rs
```diff
@@ -282,55 +343,96 @@ impl VirtQueue {
...
+    /// - The token is valid.  It will not exceed the queue size. It was previously returned by
+    ///   [`Self::add_dma_bufs`], [`Self::add_input_bufs`], or [`Self::add_output_bufs`], and it has
+    ///   not yet been removed from the queue by this method.
+    /// - The length is valid. It will not exceed the length of the original DMA buffer.
+    ///
+    /// If the device malfunctions, it may report a token or length that violates these guarantees.
+    /// Such reports are logged as errors and ignored; the reported token is not returned to the
+    /// caller, preventing an invalid token from corrupting upper-layer state. If the device
+    /// continues to malfunction, the queue may become stuck because the affected buffer cannot be
+    /// reclaimed.
```
StevenJiang1110, line 384: Could you add an Errors section to document when the method returns an error? This would help readers understand how to handle the error case. For example:

```rust
/// # Errors
///
/// This method will return error only if no data is ready.
```

StevenJiang1110 (approved):

kernel/comps/virtio/src/queue.rs
lrh2000: > I would suggest renaming it to `pop_used_with_min_bytes`. From the virtqueue's perspective, it should have no notion of what constitutes a header.

Thanks for the suggestion. It seems a good idea to me!

> What do you think about merging this method into `pop_used`?

`self.pop_used(None)` may not be very readable. I think having a separate method should have better clarity. In addition, the `Option` seems odd. Currently, `pop_used` just delegates to `pop_used_with_min_bytes(0)`. If we only have one API, we should end up with `self.pop_used(0)`. However, it is not very pretty because the purpose of 0 is unclear.

tatetian (approved):

PR #3163, opened 2026-04-21
Align metadata of devtmpfs inodes with Linux semantics

li041:
## Summary

This change aligns Asterinas devtmpfs node creation more closely with Linux.

It replaces the old path-only device-node interface with a `DevtmpfsNode`
description that carries the metadata required to accurately model Linux
defaults and per-subsystem overrides.

## Changes

* Introduce `DevtmpfsNode` for devtmpfs node metadata
* Switch device registration from `devtmpfs_path()` to `devtmpfs_node()`
* Make devtmpfs default to `root:root` with mode `0600`, matching Linux
* Allow subsystems to override devtmpfs node metadata when needed

## Notes

Linux devtmpfs defaults device nodes to `root:root` with mode `0600`.
Subsystems may override metadata via `device_get_devnode()`. This PR models that behavior explicitly in the kernel device layer.

### DIFF
--- a/kernel/src/device/fb.rs
+++ b/kernel/src/device/fb.rs
@@ -6,7 +6,7 @@ use aster_framebuffer::{ColorMapEntry, FRAMEBUFFER, FrameBuffer, MAX_CMAP_SIZE, 
 use device_id::{DeviceId, MajorId, MinorId};
 use ostd::mm::{HasPaddr, HasSize, VmIo, VmReader, VmWriter};
 
-use super::{Device, DeviceType, registry::char};
+use super::{Device, DeviceType, DevtmpfsInodeMeta, registry::char};
 use crate::{
     context::current_userspace,
     events::IoEvents,
@@ -226,8 +226,12 @@ impl Device for Fb {
         DeviceId::new(MajorId::new(29), MinorId::new(0))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some("fb0".into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>> {
+        // Linux names framebuffer device nodes as `fbN`.
+        // TODO: We currently expose only one framebuffer device,
+        // so the devtmpfs node is fixed to `fb0`.
+        // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/video/fbdev/core/fbsysfs.c#L482>.
+        Some(DevtmpfsInodeMeta::new("fb0"))
     }
 
     fn open(&self) -> Result<Box<dyn FileIo>> {
--- a/kernel/src/device/mem/mod.rs
+++ b/kernel/src/device/mem/mod.rs
@@ -28,10 +28,13 @@ pub use file::{getrandom, geturandom};
 use spin::Once;
 
 use super::{
-    Device, DeviceType,
+    Device, DeviceType, DevtmpfsInodeMeta,
     registry::char::{MajorIdOwner, acquire_major, register},
 };
-use crate::{fs::file::FileIo, prelude::*};
+use crate::{
+    fs::file::{FileIo, mkmod},
+    prelude::*,
+};
 
 /// A memory device.
 #[derive(Debug)]
@@ -61,8 +64,18 @@ impl Device for MemDevice {
         self.id
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.file.name().into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>> {
+        // Linux's memory-device table uses nonzero modes only for devices
+        // that override devtmpfs's default `u+rw` permissions.
+        // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/char/mem.c#L690>.
+        // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/char/mem.c#L734>.
+        Some(match self.file {
+            MemFile::Full | MemFile::Null | MemFile::Random | MemFile::Urandom | MemFile::Zero => {
+                DevtmpfsInodeMeta::with_mode(self.file.name(), mkmod!(a+rw))
+            }
+            MemFile::Kmsg => DevtmpfsInodeMeta::with_mode(self.file.name(), mkmod!(a+r, u+w)),
+            _ => DevtmpfsInodeMeta::new(self.file.name()),
+        })
     }
 
     fn open(&self) -> Result<Box<dyn FileIo>> {
--- a/kernel/src/device/mod.rs
+++ b/kernel/src/device/mod.rs
@@ -9,6 +9,8 @@ mod registry;
 mod shm;
 pub mod tty;
 
+use alloc::borrow::Cow;
+
 use device_id::DeviceId;
 pub use mem::{getrandom, geturandom};
 pub use pty::{PtyMaster, PtySlave, new_pty_pair};
@@ -16,7 +18,7 @@ pub use registry::lookup;
 
 use crate::{
     fs::{
-        file::{FileIo, InodeType, mkmod},
+        file::{FileIo, InodeMode, InodeType, mkmod},
         ramfs::RamFs,
         vfs::{
             inode::MknodType,
@@ -34,8 +36,8 @@ pub trait Device: Send + Sync + 'static {
     /// Returns the device ID.
     fn id(&self) -> DeviceId;
 
-    /// Returns the path where the device should appear in devtmpfs (usually under `/dev`), if any.
-    fn devtmpfs_path(&self) -> Option<String>;
+    /// Returns the metadata that specifies a device inode to be created in devtmpfs, if any.
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>>;
 
     /// Opens the device, returning a file-like object that the userspace can interact with by
     /// doing I/O.
@@ -47,7 +49,7 @@ impl Debug for dyn Device {
         f.debug_struct("Device")
             .field("type", &self.type_())
             .field("id", &self.id())
-            .field("devtmpfs_path", &self.devtmpfs_path())
+            .field("devtmpfs_meta", &self.devtmpfs_meta())
             .finish_non_exhaustive()
     }
 }
@@ -59,6 +61,49 @@ pub enum DeviceType {
     Block,
 }
 
+/// The metadata that describes a device inode in devtmpfs.
+///
+/// The metadata contains the inode path relative to `/dev` and the
+/// permission bits used when creating the inode. Device subsystems can use this
+/// type to override the default mode.
+///
+/// If a device does not specify a mode explicitly, we use `mkmod!(u+rw)`,
+/// matching Linux devtmpfs's default device inode permissions.
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/base/devtmpfs.c#L11>.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DevtmpfsInodeMeta<'a> {
+    path: Cow<'a, str>,
+    mode: InodeMode,
+}
+
+impl<'a> DevtmpfsInodeMeta<'a> {
+    /// Creates the metadata for a devtmpfs inode with the default mode (`u+rw`).
+    pub fn new(path: impl Into<Cow<'a, str>>) -> Self {
+        Self {
+            path: path.into(),
+            mode: mkmod!(u+rw),
+        }
+    }
+
+    /// Creates the metadata for a devtmpfs inode with the specified path and mode.
+    pub fn with_mode(path: impl Into<Cow<'a, str>>, mode: InodeMode) -> Self {
+        Self {
+            path: path.into(),
+            mode,
+        }
+    }
+
+    /// Returns the device inode path relative to `/dev`.
+    pub fn path(&self) -> &str {
+        &self.path
+    }
+
+    /// Returns the permission bits of the device inode.
+    pub fn mode(&self) -> InodeMode {
+        self.mode
+    }
+}
+
 /// Adds a device node in `/dev`.
 ///
 /// If the parent path does not exist, it will be created as a directory.
@@ -68,12 +113,12 @@ pub enum DeviceType {
 pub fn add_node(
     dev_type: DeviceType,
     dev_id: u64,
-    path: &str,
+    meta: &DevtmpfsInodeMeta<'_>,
     path_resolver: &PathResolver,
 ) -> Result<Path> {
     let mut dev_path = path_resolver.lookup(&FsPath::try_from("/dev").unwrap())?;
     let mut relative_path = {
-        let relative_path = path.trim_start_matches('/');
+        let relative_path = meta.path().trim_start_matches('/');
         if relative_path.is_empty() {
             return_errno_with_message!(Errno::EINVAL, "the device path is invalid");
         }
@@ -102,7 +147,7 @@ pub fn add_node(
                         DeviceType::Block => MknodType::BlockDevice(dev_id),
                         DeviceType::Char => MknodType::CharDevice(dev_id),
                     };
-                    dev_path = dev_path.mknod(next_name, mkmod!(a+rw), mknod_type)?;
+                    dev_path = dev_path.mknod(next_name, meta.mode(), mknod_type)?;
                 } else {
                     // Create the parent directory
                     dev_path =
--- a/kernel/src/device/registry/block.rs
+++ b/kernel/src/device/registry/block.rs
@@ -6,7 +6,7 @@ use device_id::DeviceId;
 use ostd::mm::VmIo;
 
 use crate::{
-    device::{Device, DeviceType, add_node},
+    device::{Device, DeviceType, DevtmpfsInodeMeta, add_node},
     events::IoEvents,
     fs::{
         file::{FileIo, StatusFlags},
@@ -38,9 +38,9 @@ pub(super) fn init_in_first_kthread() {
 pub(super) fn init_in_first_process(path_resolver: &PathResolver) -> Result<()> {
     for device in aster_block::collect_all() {
         let device = Arc::new(BlockFile::new(device));
-        if let Some(devtmpfs_path) = device.devtmpfs_path() {
+        if let Some(devtmpfs_meta) = device.devtmpfs_meta() {
             let dev_id = device.id().as_encoded_u64();
-            add_node(DeviceType::Block, dev_id, &devtmpfs_path, path_resolver)?;
+            add_node(DeviceType::Block, dev_id, &devtmpfs_meta, path_resolver)?;
         }
     }
 
@@ -77,8 +77,8 @@ impl Device for BlockFile {
         self.0.id()
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.0.name().into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>> {
+        Some(DevtmpfsInodeMeta::new(self.0.name()))
     }
 
     fn open(&self) -> Result<Box<dyn FileIo>> {
--- a/kernel/src/device/tty/device.rs
+++ b/kernel/src/device/tty/device.rs
@@ -11,7 +11,7 @@ use spin::Once;
 
 use crate::{
     device::{
-        Device, DeviceType,
+        Device, DeviceType, DevtmpfsInodeMeta,
         registry::char,
         tty::{
             Tty,
@@ -20,7 +20,7 @@ use crate::{
             vt::{VtDriver, tty1_device},
         },
     },
-    fs::file::FileIo,
+    fs::file::{FileIo, mkmod},
     prelude::*,
     process::{JobControl, Terminal},
 };
@@ -45,8 +45,8 @@ impl Device for Tty0Device {
         DeviceId::new(MajorId::new(4), MinorId::new(0))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some("tty0".into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>> {
+        Some(DevtmpfsInodeMeta::new("tty0"))
     }
 
     fn open(&self) -> Result<Box<dyn FileIo>> {
@@ -73,8 +73,9 @@ impl Device for TtyDevice {
         DeviceId::new(MajorId::new(5), MinorId::new(0))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some("tty".into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>> {
+        // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/tty/tty_io.c#L3511>.
+        Some(DevtmpfsInodeMeta::with_mode("tty", mkmod!(a+rw)))
     }
 
     fn open(&self) -> Result<Box<dyn FileIo>> {
@@ -135,8 +136,8 @@ impl Device for SystemConsole {
         DeviceId::new(MajorId::new(5), MinorId::new(1))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some("console".into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta<'_>> {
+        Some(DevtmpfsInodeMeta::new("console"))
     }
 
     fn open(&self) -> Result<Box<dyn FileIo>> {
--- a/kernel/src/device/tty/driver.rs
+++ b/kernel/src/device/tty/driver.rs
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use crate::{
-    device::tty::{Tty, termio::CTermios},
+    device::{
+        DevtmpfsInodeMeta,
+        tty::{Tty, termio::CTermios},
+    },
     fs::file::FileIo,
     prelude::*,
     util::ioctl::RawIoctl,
@@ -19,8 +22,8 @@ pub trait TtyDriver: Send + Sync + 'static {
     /// The device major ID.
     const DEVICE_MAJOR_ID: u32;
 
-    /// Returns the path where the TTY should appear under devtmpfs, if any.
-    fn devtmpfs_path(&self, index: u32) -> Option<String>;
+    /// Returns the metadata that specifies a TTY device inode to be created in devtmpfs, if any.
+    fn devtmpfs_meta(&self, index: u32) -> Option<DevtmpfsInodeMeta<'_>>;
 
     /// Opens the TTY.
     ///
--- a/kernel/src/device/tty/mod.rs
+++ b/kernel/src/device/tty/mod.rs
@@ -5,7 +5,7 @@ use ostd::sync::LocalIrqDisabled;
 
 use self::{line_discipline::LineDiscipline, termio::CFontOp};
 use crate::{
-    device::{Device, DeviceType},
+    device::{Device, DeviceType, DevtmpfsInodeMeta},
     events::IoEvents,
     fs::file::{FileIo, StatusFlags},
     prelude::*,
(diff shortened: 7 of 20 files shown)


lrh2000:
The comments added in this PR is very confusing. I think they're all generated by LLM. Could you please avoid those?

I'd prefer no comments at all than comments referring to unnecessary Linux implementation details. Only write useful comments if they're necessary and provide good information.

Please do not use links under "https://codebrowser.dev/linux/linux/drivers/char/mem.c.html" at all. Run `grep -R "Reference:" kernel` yourself to see what we should use. At least remove the links for the well-known ones. I don't think they make much sense.

kernel/src/device/pty/mod.rs
```diff
@@ -34,7 +34,6 @@ pub fn init_in_first_process(path_resolver: &PathResolver, ctx: &Context) -> Res
 
     DEV_PTS.call_once(|| Path::new_fs_root(devpts_mount));
 
-    // Create the "ptmx" symlink.
```
lrh2000, line 37: Do not delete existing comments in an unrelated PR without good reason.

This has never happened before, but the buggy LLM is now constantly trying to delete existing comments. So I am quite confused when seeing this... If possible, please try to avoid these diffs in your self-review.

kernel/src/device/mem/mod.rs
```diff
@@ -61,8 +64,19 @@ impl Device for MemDevice {
         self.id
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.file.name().into())
+    fn devtmpfs_node(&self) -> Option<DevtmpfsNode> {
+        // Linux's `/dev/null`, `/dev/zero`, `/dev/full`, `/dev/random`, and
+        // `/dev/urandom` are exposed with explicit `0666` modes, overriding
+        // devtmpfs's default `0600`.
+        // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/char/mem.c#L710>.
```
lrh2000: This points to:
```c
	const struct memdev *dev;
```

How is that line related here?

kernel/src/device/mem/mod.rs
```diff
@@ -61,8 +64,19 @@ impl Device for MemDevice {
...
+        // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/char/mem.c#L750>.
```
lrh2000: This points to an empty line. How is the empty line related here?

kernel/src/device/registry/block.rs
```diff
@@ -77,8 +77,8 @@ impl Device for BlockFile {
         self.0.id()
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.0.name().into())
+    fn devtmpfs_node(&self) -> Option<crate::device::DevtmpfsNode> {
+        Some(crate::device::DevtmpfsNode::new(self.0.name()))
```
lrh2000: Import the `DevtmpfsNode` type and avoid typing the full-qualified path.

kernel/src/device/mod.rs
```diff
@@ -59,6 +59,37 @@ pub enum DeviceType {
...
+    /// Creates a devtmpfs node description with default metadata.
+    pub fn new(path: impl Into<String>) -> Self {
+        Self {
+            path: path.into(),
+            mode: mkmod!(u+rw),
+        }
+    }
+
+    // Sets the permission bits of the device node.
+    fn set_mode(mut self, mode: InodeMode) -> Self {
```
lrh2000: I think it is better just to provide a `with_mode(path, mode)` constructor.

kernel/src/device/mod.rs
```diff
@@ -59,6 +59,37 @@ pub enum DeviceType {
     Block,
 }
 
+/// Linux devtmpfs defaults device nodes to mode `0600`. Subsystems may then
+/// override the path and mode via `device_get_devnode()`.
```
lrh2000: This makes no sense. Asterinas do not have the `device_get_devnode` method.

Doc comments should describe what the type does. Instead of describing what Linux has.

kernel/src/device/mod.rs
```diff
@@ -59,6 +59,37 @@ pub enum DeviceType {
...
+/// Linux also supports overriding the node owner UID and GID there, but
+/// Asterinas does not need that today, so this type intentionally models only
+/// the path and mode for now.
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/base/devtmpfs.c#L11>.
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/base/core.c#L3963>.
```
lrh2000: The doc comments should start with a short summary sentence, instead of starting with a paragraph.
```rust
/// Short summary sentence.
///
/// Detailed explanation.
///
/// Detailed explanation.
```

kernel/src/fs/fs_impls/devpts/ptmx.rs
```diff
@@ -175,7 +175,16 @@ impl Device for Inner {
...
+    fn devtmpfs_node(&self) -> Option<crate::device::DevtmpfsNode> {
+        // `/dev/ptmx` is created separately during PTY initialization rather
+        // than through the generic devtmpfs device-registration path. Keep
+        // returning `None` here to preserve that layout.
+        //
+        // Linux accepts a `ptmx` entry that is outside the devpts mount and
+        // resolves it to the appropriate devpts instance after path
+        // resolution, including a symlinked `/dev/ptmx`.
+        // Reference: <https://codebrowser.dev/linux/linux/fs/devpts/inode.c.html#135>
+        // Reference: <https://codebrowser.dev/linux/linux/drivers/tty/tty_io.c.html#3511>
```
lrh2000: Could you please explain what this comment means and how it helps us understand the code?

kernel/src/fs/fs_impls/devpts/ptmx.rs
```diff
@@ -175,7 +175,16 @@ impl Device for Inner {
         DeviceId::new(MajorId::new(PTMX_MAJOR_NUM), MinorId::new(PTMX_MINOR_NUM))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
+    fn devtmpfs_node(&self) -> Option<crate::device::DevtmpfsNode> {
+        // `/dev/ptmx` is created separately during PTY initialization rather
```
lrh2000: Use `pty` instead of `PTY` for consistency:
```
> grep -R " pty " kernel | wc -l
8
> grep -R " PTY " kernel | wc -l
0
```

kernel/src/device/fb.rs
```diff
@@ -226,8 +226,13 @@ impl Device for Fb {
         DeviceId::new(MajorId::new(29), MinorId::new(0))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some("fb0".into())
+    fn devtmpfs_node(&self) -> Option<DevtmpfsNode> {
+        // Linux names framebuffer device nodes as `fbN`.
+        // Todo: We currently expose only one framebuffer device,
```
lrh2000:
```suggestion
        // TODO: We currently expose only one framebuffer device,
```

kernel/src/device/mod.rs
```diff
@@ -34,8 +34,8 @@ pub trait Device: Send + Sync + 'static {
     /// Returns the device ID.
     fn id(&self) -> DeviceId;
 
-    /// Returns the path where the device should appear in devtmpfs (usually under `/dev`), if any.
-    fn devtmpfs_path(&self) -> Option<String>;
+    /// Returns the device node metadata created in devtmpfs, if any.
```
lrh2000: I think it is called before the node creation. So perhaps:
```suggestion
    /// Returns the metadata that specifies a device node to be created in devtmpfs, if any.
```

lrh2000 (commented):
I am really disappointed that someone that works on our project for months submitting a PR with such low quality. I'm not blaming you for anything. The fact is, though, that it disappoints me.

By low quality, I mean:
 - The link provided in this PR is broken and points to nonsense lines. This happens after I request a update to the link. So it seems that a clear thing is that the new changes are pushed without extra checks.
 - I request to remove the `codebrowser` links. But there are still some in the new revision.
 - Some unrelated comments have been deleted. Given this PR's small size, I assume it can easily be caught by a self-review. But it is not. I assume it was not self-reviewed after it was generated by some tools (assuming AI).
 - The doc comments are written in an odd way. It says some Linux behavior instead of describing a type that defined in Asterinas' code. I don't believe you wrote them. Have you even read them?

Please try to reply to the following questions. I am really curious. Thanks.
 - Could you please explain how you wrote this PR? For example, what tools did you use?
 - If the diffs were generated by an automated tool (e.g., LLM), did you self-review the changes?
 - How much time did you spend writing or reviewing the code changes **manually**, if any? (Rough estimate)
 - If you used a tool to generate most of the changes, how many lines did you change by hand **before opening this PR**? (Rough estimate)
 - When I asked you to remove the nonsense comments, how did you respond? If you are still using a tool to update this PR, have you checked that the new changes are correct? For example, do the links point to the correct location? Are the links aligned correctly with the comment text?

li041:
I sincerely apologize for the confusion and the issues in this PR. I should have double-checked the generated content more carefully, especially the comments reconstructed by the LLM. Thank you for pointing these problems out — I take responsibility for the oversight and will make sure to review such changes more thoroughly going forward.

> Could you please explain how you wrote this PR? For example, what tools did you use?

gpt-5.4 medium

> If the diffs were generated by an automated tool (e.g., LLM), did you self-review the changes?

It's my fault that I didn't recheck when LLM recounstruct comments. I'll review whatever changes made by LLM next time

> How much time did you spend writing or reviewing the code changes **manually**, if any? (Rough estimate)

Less than half an hour. 
Most of my time spent on exploring how linux create inode for device that exists in Asterinas. Those semantics are fresh for me.
> If you used a tool to generate most of the changes, how many lines did you change by hand **before opening this PR**? (Rough estimate)

Less than 50 lines. 
I just narrow visiblity of DevtmpfsNode, and remove unnessary fields and methods.
> When I asked you to remove the nonsense comments, how did you respond? If you are still using a tool to update this PR, have you checked that the new changes are correct? For example, do the links point to the correct location? Are the links aligned correctly with the comment text?

It's indeed surprising when an LLM misses something as seemingly straightforward as changing comments. Maybe It's due to my prompt. However, What makes it a mess is the absence of rechecking.

kernel/src/fs/fs_impls/devpts/ptmx.rs
li041: This comment mainly explain why returning `None` here which is abnormal. It's not obviours to me. But it seems that this is only due to my knowledge.

lrh2000:
OK. Thanks for your reply.

Since reviews take time, hopefully you can do more self-reviews before opening PRs or pushing new changes. Thanks for understanding.

It's too easy for the LLM to reduce coding time compared to review time. Otherwise, the review burden becomes unmanageable.

The simplest rule is that if you wouldn't write the code or comments a certain way, neither should the LLM. Ultimately, you are the one authoring the diff, not the LLM.

kernel/src/fs/fs_impls/devpts/ptmx.rs
lrh2000: I mean the second paragraph:
```rust
        // Linux accepts a `ptmx` entry that is outside the devpts mount and
        // resolves it to the appropriate devpts instance after path
        // resolution, including a symlinked `/dev/ptmx`.
```
To be honest, I have read this paragraph several times and I don't know what it is saying..

The first paragraph may help you understand why we return `None`. However, I don't know what the second paragraph is talking about. Perhaps I'm missing something. Depending on its usefulness, the paragraph should probably be removed or rewritten.

I don't understand it. Therefore, I cannot judge its usefulness.

Should I be more concrete? Okay:

> Linux accepts a `ptmx` entry

What does it mean to say that Linux **accepts an entry**? I don't understand this phrase.

> resolves it to the appropriate devpts instance after path resolution

Hmmm. I'm not sure how it relates to the code here. I don't know what an **appropriate devpts instance** is. I have no idea how **path resolution** is involved.

> including a symlinked `/dev/ptmx`.

I am really confused. It is **not** a symbolic link, AFAICK:
```
> ls /dev/ptmx -al
crw-rw-rw- 1 root tty 5, 2 Apr 21 18:52 /dev/ptmx
```

The second paragraph really confuses me. Could you explain what it wants to mean?

kernel/src/device/tty/mod.rs
```diff
@@ -326,8 +326,10 @@ impl<D: TtyDriver> Device for Tty<D> {
         )
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        self.driver.devtmpfs_path(self.index)
+    fn devtmpfs_node(&self) -> Option<crate::device::DevtmpfsNode> {
+        self.driver
+            .devtmpfs_path(self.index)
+            .map(crate::device::DevtmpfsNode::new)
```
lrh2000: Import `DevtmpfsNode` in scope.

test/initramfs/src/regression/device/devtmpfs_mode.c
```diff
@@ -0,0 +1,39 @@
...
+	unsigned int minor;
+};
+
+static const struct device_node device_nodes[] = {
+	{ "/dev/null", 0666, 1, 3 },
+	{ "/dev/zero", 0666, 1, 5 },
+	{ "/dev/tty", 0666, 5, 0 },
+};
+
+FN_TEST(check_rdev_and_mode)
```
lrh2000:
```suggestion
FN_TEST(rdev_and_mode)
```

`FN_TEST` will append a `test_` prefix. So the full name is `test_rdev_and_mode`, which is reasonable. But `test_check_rdev_and_mode` does not make much sense.

test/initramfs/src/regression/device/random.c
```diff
@@ -3,10 +3,39 @@
...
+	unsigned int major;
+	unsigned int minor;
+};
+
+static const struct random_device random_devices[] = {
+	{ "/dev/random", 1, 8 },
+	{ "/dev/urandom", 1, 9 },
+};
+
+FN_TEST(check_rdev_and_mode)
```
lrh2000: Same here.

kernel/src/device/tty/device.rs
```diff
@@ -73,8 +73,8 @@ impl Device for TtyDevice {
         DeviceId::new(MajorId::new(5), MinorId::new(0))
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some("tty".into())
+    fn devtmpfs_node(&self) -> Option<DevtmpfsNode> {
+        Some(DevtmpfsNode::with_mode("tty", mkmod!(a+rw)))
```
lrh2000:
```rust
// Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/tty/tty_io.c#L3511>.
```

kernel/src/device/mod.rs
```diff
@@ -59,6 +58,41 @@ pub enum DeviceType {
     Block,
 }
 
+/// Describes a device node in devtmpfs.
+///
+/// The description contains the node path relative to `/dev` and the
+/// permission bits used when creating the node. Device subsystems can use this
+/// type to override the default mode.
+///
+/// If a device does not specify a mode explicitly, Asterinas uses `0600`,
```
lrh2000:
```suggestion
/// If a device does not specify a mode explicitly, Asterinas uses `mkmod!(u+rw)`,
```

I think it would be better if we aligned our code style. We use `mkmod` and `u+rw`.

kernel/src/device/mod.rs
```diff
@@ -59,6 +58,41 @@ pub enum DeviceType {
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DevtmpfsNode {
+    // The device node path relative to `/dev`.
+    path: String,
+    // The permission bits of the device node.
+    mode: InodeMode,
+}
+
+impl DevtmpfsNode {
+    /// Creates a devtmpfs node description with default metadata.
```
lrh2000:
```suggestion
    /// Creates a devtmpfs node description with the default mode (`u+rw`).
```

kernel/src/device/mod.rs
```diff
@@ -59,6 +58,41 @@ pub enum DeviceType {
...
+/// type to override the default mode.
+///
+/// If a device does not specify a mode explicitly, Asterinas uses `0600`,
+/// matching Linux devtmpfs's default device-node permissions.
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/base/devtmpfs.c#L11>.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DevtmpfsNode {
+    // The device node path relative to `/dev`.
+    path: String,
+    // The permission bits of the device node.
```
lrh2000: I guess we can make them doc comments. `//` -> `///`.

kernel/src/device/mem/mod.rs
```diff
@@ -61,8 +64,18 @@ impl Device for MemDevice {
         self.id
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.file.name().into())
+    fn devtmpfs_node(&self) -> Option<DevtmpfsNode> {
+        // Linux's memory-device table uses nonzero modes only for devices
+        // that override devtmpfs's default `0600` permissions.
```
lrh2000:
```suggestion
        // that override devtmpfs's default `u+rw` permissions.
```

lrh2000 (commented):
Thanks for the revision. This version is much better.

I just have some minor comments here.

test/initramfs/src/regression/device/devtmpfs_mode.c
```diff
@@ -0,0 +1,39 @@
...
+	const char *path;
+	mode_t mode;
+	unsigned int major;
+	unsigned int minor;
+};
+
+static const struct device_node device_nodes[] = {
+	{ "/dev/null", 0666, 1, 3 },
+	{ "/dev/zero", 0666, 1, 5 },
+	{ "/dev/tty", 0666, 5, 0 },
```
lrh2000, line 23: Should we have some other, like, `console`/`tty0`/`tty1`/`ttyS0`?

kernel/src/device/mod.rs
```diff
@@ -33,9 +33,8 @@ pub trait Device: Send + Sync + 'static {
 
     /// Returns the device ID.
     fn id(&self) -> DeviceId;
-
```
lrh2000, line 36: The blank line on line 36 should be kept? I have no idea why it should be removed.

kernel/src/device/mod.rs
```diff
@@ -33,9 +33,8 @@ pub trait Device: Send + Sync + 'static {
...
-    /// Returns the path where the device should appear in devtmpfs (usually under `/dev`), if any.
-    fn devtmpfs_path(&self) -> Option<String>;
+    /// Returns the metadata that specifies a device node to be created in devtmpfs, if any.
+    fn devtmpfs_node(&self) -> Option<DevtmpfsNode>;
```
tatetian: [`accurate-name`](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#accurate-names): As you said in the comment, it is "the metadata that specifies a device node". So I suggest renaming `DevtmpfsNode` to `DevtmpFsInodeMeta`. If you has to choose between "concise-but-misleading" and "longer-but-accurate", choose the latter.

kernel/src/device/mod.rs
```diff
@@ -59,6 +58,41 @@ pub enum DeviceType {
...
+/// permission bits used when creating the node. Device subsystems can use this
+/// type to override the default mode.
+///
+/// If a device does not specify a mode explicitly, Asterinas uses `0600`,
+/// matching Linux devtmpfs's default device-node permissions.
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/base/devtmpfs.c#L11>.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DevtmpfsNode {
+    // The device node path relative to `/dev`.
+    path: String,
```
tatetian: [`minimize-copies`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/select-topics/performance.html#minimize-copies): Around half callers of `DevtmpfsNode` have a name that is statically known, i.e., `&'static str`. Turning static strings to heap allocated form (`String`) is a complete waste of time.

So I suggest replacing `String` with `Cow<'static, str>`. The latter one has the same size as the former. So you can avoid one heap allocation--whenever possible---without paying any extra cost:

```rust
use std::borrow::Cow;

fn main() {
    let a: Cow<'static, str> = "borrowed".into();
    let b: Cow<'static, str> = "owned".to_string().into();
    println!("size of String: {}", core::mem::size_of::<String>());
    // Output: size of String: 24
    println!("size of Cow<'static, str>: {}", core::mem::size_of::<Cow<'static, str>>());
    // Output: size of Cow<'static, str>: 24
}
```

The new version of `DevtmpfsNode` would look like below:

```rust
pub struct DevtmpfsNode {
    path: Cow<'static, str>,
    //...
}

impl DevtmpfsNode {
    pub fn new<S: Into<Cow<'static, str>>>(path: S) -> Self {
        ...
    }
}
```

test/initramfs/src/regression/device/devtmpfs_mode.c
li041: > https://elixir.bootlin.com/linux/v6.18/source/drivers/tty/tty_io.c#L3511

Only `tty(5, 0)` and `ptmx(5, 2)` devices overwrite default mode.

lrh2000: So, actually, you only want to write tests for nodes that overwrite the default mode?

I originally thought having tests for the other nodes would also be helpful. Not sure if it's intended here to have `"/dev/tty0", 0600`, etc.

kernel/src/device/registry/block.rs
```diff
@@ -77,8 +77,8 @@ impl Device for BlockFile {
         self.0.id()
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.0.name().into())
+    fn devtmpfs_node(&self) -> Option<DevtmpfsInodeMeta> {
```
lrh2000: I was thought whether to change the method name as well, such as:
```suggestion
    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta> {
```

We may want to rename variables:
```patch
-    node: &DevtmpfsInodeMeta,
+    meta: &DevtmpfsInodeMeta,
```

Perhaps use "metadata" instead of "description" in the doc comments as well.
```patch
diff --git a/kernel/src/device/mod.rs b/kernel/src/device/mod.rs
index e0888f5b6..25b27318f 100644
--- a/kernel/src/device/mod.rs
+++ b/kernel/src/device/mod.rs
@@ -61,9 +61,9 @@ pub enum DeviceType {
     Block,
 }
 
-/// Describes a device node in devtmpfs.
+/// The metadata that describes a device node in devtmpfs.
 ///
-/// The description contains the node path relative to `/dev` and the
+/// The metadata contains the node path relative to `/dev` and the
 /// permission bits used when creating the node. Device subsystems can use this
 /// type to override the default mode.
 ///
@@ -79,7 +79,7 @@ pub struct DevtmpfsInodeMeta {
 }
 
 impl DevtmpfsInodeMeta {
-    /// Creates a devtmpfs node description with default mode (`u+rw`).
+    /// Creates the metadata for a devtmpfs node with the default mode (`u+rw`).
     pub fn new(path: impl Into<Cow<'static, str>>) -> Self {
         Self {
             path: path.into(),
@@ -87,7 +87,7 @@ impl DevtmpfsInodeMeta {
         }
     }
 
-    /// Creates a devtmpfs node description with the specified path and mode.
+    /// Creates the metadata for a devtmpfs node with the specified path and mode.
     pub fn with_mode(path: impl Into<Cow<'static, str>>, mode: InodeMode) -> Self {
         Self {
             path: path.into(),
```

test/initramfs/src/regression/device/devtmpfs_mode.c
li041: Oh, it's truly helpful. The mode of those devices is not as simple as I think. It seems that [udev](https://github.com/systemd/systemd/blob/main/rules.d/50-udev-default.rules.in) in userspace can also overwrite those files, so the behaviour may diff between Asterinas and Linux.

li041: So should this test take `systemd` into account? 
Those modes are 0600 from pure kernel path, but I test it on Linux and get:
```
  /dev/console  expected mode=0600 rdev=5:1   actual mode=0620 rdev=136:2
  /dev/tty0     expected mode=0600 rdev=4:0   actual mode=0620 rdev=4:0
  /dev/tty1     expected mode=0600 rdev=4:1   actual mode=0620 rdev=4:1
  /dev/fb0      expected mode=0600 rdev=29:0  actual mode=0660 rdev=29:0
  ```

Comment (lrh2000) [reply test/initramfs/src/regression/device/devtmpfs_mode.c:23]: > It seems that [udev](https://github.com/systemd/systemd/blob/main/rules.d/50-udev-default.rules.in) in userspace can also overwrite those files, so the behaviour may diff between Asterinas and Linux.
> So should this test take `systemd` into account?

I don't think we should consider `systemd`. Maybe we should just leave a comment to prevent it from being "fixed" in the future?

Not sure if we can do better in some other ways... Indeed, it also seems to be a bit more complex than I originally expected.

File: kernel/src/device/registry/block.rs (hunk 24 above)
Comment (tatetian) [reply kernel/src/device/registry/block.rs]: Totally agree.

Comment (tatetian) [reply kernel/src/device/registry/block.rs]: In addition, the doc should replace the word "node" with "inode".

File: kernel/src/device/tty/driver.rs [hunk 25]
```diff
@@ -19,8 +22,8 @@ pub trait TtyDriver: Send + Sync + 'static {
     /// The device major ID.
     const DEVICE_MAJOR_ID: u32;
 
-    /// Returns the path where the TTY should appear under devtmpfs, if any.
-    fn devtmpfs_path(&self, index: u32) -> Option<String>;
+    /// Returns the devtmpfs node where the TTY should appear, if any.
```
lrh2000: This line should be fixed as well. 
```suggestion
    /// Returns the metadata that specifies where the TTY should appear under devtmpfs, if any.
```

test/initramfs/src/regression/device/devtmpfs_mode.c
```diff
@@ -0,0 +1,51 @@
...
+
+struct device_node {
+	const char *path;
+	mode_t mode;
+	unsigned int major;
+	unsigned int minor;
+};
+
+static const struct device_node device_nodes[] = {
+	// Explicit Linux overrides (non-default mode).
```
lrh2000:
```suggestion
	/* Explicit Linux overrides (non-default mode). */
```

for consistency with other comments in this file.

kernel/src/device/registry/block.rs
```diff
@@ -77,8 +77,8 @@ impl Device for BlockFile {
         self.0.id()
     }
 
-    fn devtmpfs_path(&self) -> Option<String> {
-        Some(self.0.name().into())
+    fn devtmpfs_meta(&self) -> Option<DevtmpfsInodeMeta> {
+        Some(DevtmpfsInodeMeta::new(self.0.name().to_string()))
```
lrh2000: If you're already using `Cow<'static, str>`, then this `to_string` shouldn't be necessary, since we can always use
```rust
pub struct DevtmpfsInodeMeta<'a> {
    path: Cow<'a, str>,
    mode: InodeMode,
}
```
?

test/initramfs/src/regression/device/run_test.sh
```diff
@@ -9,6 +9,7 @@ set -e
 ./pty/open_pty
 ./pty/pty_blocking
 ./pty/pty_packet_mode
+./devtmpfs_mode
```
lrh2000, line 13: Let's add a blank line before this line. Given that `./pty` block is sorted independently.

test/initramfs/src/regression/device/devtmpfs_mode.c
```diff
@@ -0,0 +1,51 @@
...
+	/* Explicit Linux overrides (non-default mode). */
+	{ "/dev/null", 0666, 1, 3 },
+	{ "/dev/zero", 0666, 1, 5 },
+	{ "/dev/tty", 0666, 5, 0 },
+	/*
+	 * Devices that fall back to the kernel devtmpfs default (0600).
+	 *
+	 * Do not replace these modes with values observed on a host Linux.
+	 * Systemd may later change the default modes of these devices in user space.
+	 * Reference: <https://github.com/systemd/systemd/blob/main/rules.d/50-udev-default.rules.in>
```
lrh2000: Replace it with a permanent link.

Although a permanent link is longer, it is better than one that becomes invalid over time.

```suggestion
	 * Reference: <https://github.com/systemd/systemd/blob/9149c7595305a7c4d105d5d33ba25733af4302eb/rules.d/50-udev-default.rules.in>
```

(e.g., if the upstream renames the directory or moves the file, we will get a broken link...)

lrh2000 (approved):

tatetian (approved):

PR #3164, opened 2026-04-21
Set `procfs` `fd` ownership from task credentials

taosue:
Set the ownership of `/proc/[pid]/fd`, `/proc/[pid]/fd/<n>`, and `/proc/[pid]/fdinfo/<n>` from the target task credentials instead of leaving them as the `procfs` template default `root:root`.

This fixes cases where a child process changes its `uid/gid` and then opens an inherited FIFO through `/proc/self/fd/<fd>`. Without the correct `procfs` inode ownership, permission checks can reject the open even though Linux allows this path.

This aligns Linux's behavior since Linux updates `proc fd inode` ownership from the target task when instantiating `fd` and `fdinfo` entries:

```c
task_dump_owner(task, 0, &inode->i_uid, &inode->i_gid);
```
This is done through [`tid_fd_update_inode()`](https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/fd.c#L127-L141), which is called when creating both [`/proc/[pid]/fd/<n>`](https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/fd.c#L220) and [`/proc/[pid]/fdinfo/<n>`](https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/fd.c#L384).

Linux also documents why `/proc/[pid]/fd` needs special handling, see [here](https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/fd.c#L327-L330):
```c
/* /proc/pid/fd needs a special permission handler so that a process can still
 * access /proc/self/fd after it has executed a setuid().
 */
```

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/pid/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/mod.rs
@@ -12,6 +12,7 @@ use crate::{
     },
     prelude::*,
     process::pid_table::{PidEntry, PidEntryType},
+    thread::Thread,
 };
 
 mod task;
@@ -52,6 +53,10 @@ impl PidDirOps {
 }
 
 impl DirOps for PidDirOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn lookup_child(&self, this_dir: &ProcDir<Self>, name: &str) -> Result<Arc<dyn Inode>> {
         if self.0.pid_entry().type_().is_none() {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
@@ -15,6 +15,7 @@ use crate::{
     },
     prelude::*,
     process::posix_thread::AsThreadLocal,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/cgroup` (and also `/proc/[pid]/cgroup`).
@@ -28,6 +29,10 @@ impl CgroupFileOps {
 }
 
 impl FileOps for CgroupFileOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(process) = self.0.process() else {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/mountstats.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/mountstats.rs
@@ -14,6 +14,7 @@ use crate::{
     },
     prelude::*,
     process::posix_thread::AsPosixThread,
+    thread::Thread,
 };
 
 /// A single entry in the `mountstats` file.
@@ -81,6 +82,10 @@ impl MountStatsFileOps {
 }
 
 impl FileOps for MountStatsFileOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(thread) = self.0.thread() else {
             return_errno_with_message!(Errno::ESRCH, "the thread does not exist");
--- a/kernel/src/fs/fs_impls/procfs/template/file.rs
+++ b/kernel/src/fs/fs_impls/procfs/template/file.rs
@@ -16,6 +16,7 @@ use crate::{
     },
     prelude::*,
     process::{Gid, Uid},
+    thread::Thread,
 };
 
 pub struct ProcFile<F: FileOps> {
@@ -70,7 +71,6 @@ impl<F: FileOps + 'static> InodeIo for ProcFile<F> {
 #[inherit_methods(from = "self.common")]
 impl<F: FileOps + 'static> Inode for ProcFile<F> {
     fn size(&self) -> usize;
-    fn metadata(&self) -> Metadata;
     fn extension(&self) -> &Extension;
     fn ino(&self) -> u64;
     fn mode(&self) -> Result<InodeMode>;
@@ -86,6 +86,11 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
     fn ctime(&self) -> Duration;
     fn set_ctime(&self, time: Duration);
     fn fs(&self) -> Arc<dyn FileSystem>;
+
+    fn metadata(&self) -> Metadata {
+        let owner_thread = self.inner.owner_thread();
+        self.common.metadata_with_owner(owner_thread)
+    }
 
     fn resize(&self, _new_size: usize) -> Result<()> {
         // Resizing files under `/proc` will succeed, but will do nothing.
@@ -119,6 +124,11 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
 }
 
 pub trait FileOps: Sync + Send {
+    /// Returns the thread whose credentials own this procfs inode.
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        None
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize>;
 
     fn write_at(&self, _offset: usize, _reader: &mut VmReader) -> Result<usize> {
@@ -135,10 +145,19 @@ pub trait FileOps: Sync + Send {
 }
 
 pub trait FileOpsByHandle: Sync + Send {
+    /// Returns the thread whose credentials own this procfs inode.
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        None
+    }
+
     fn open(&self, access_mode: AccessMode, status_flags: StatusFlags) -> Result<Box<dyn FileIo>>;
 }
 
 impl<T: FileOpsByHandle> FileOps for T {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        FileOpsByHandle::owner_thread(self)
+    }
+
     fn read_at(&self, _offset: usize, _writer: &mut VmWriter) -> Result<usize> {
         unreachable!("`read_at` is never called when `open` returns `Some(_)`")
     }
--- a/kernel/src/fs/fs_impls/procfs/template/mod.rs
+++ b/kernel/src/fs/fs_impls/procfs/template/mod.rs
@@ -20,13 +20,19 @@ use crate::{
         },
     },
     prelude::*,
-    process::{Gid, Uid},
+    process::{Gid, Uid, posix_thread::AsPosixThread},
+    thread::Thread,
 };
 
 mod dir;
 mod file;
 mod sym;
 
+/// Shared procfs inode state.
+///
+/// FIXME: Procfs permissions should be checked during each operation, not by relying on mutable
+/// inode ownership. See Linux comment:
+/// <https://elixir.bootlin.com/linux/v6.13/source/fs/proc/base.c#L107>.
 struct Common {
     metadata: RwLock<Metadata>,
     extension: Extension,
@@ -48,6 +54,23 @@ impl Common {
 
     fn metadata(&self) -> Metadata {
         *self.metadata.read()
+    }
+
+    fn metadata_with_owner(&self, owner_thread: Option<Arc<Thread>>) -> Metadata {
+        let Some(owner_thread) = owner_thread else {
+            return self.metadata();
+        };
+
+        let credentials = owner_thread.as_posix_thread().unwrap().credentials();
+        let mut metadata = self.metadata.write();
+        // Cache the dynamic owner into the metadata so that if the thread
+        // later exits, subsequent calls fall back to the last known owner
+        // instead of the root user. This is a best-effort attempt to align
+        // with Linux behavior. See:
+        // <https://github.com/asterinas/asterinas/pull/3164#discussion_r3212307770>.
+        metadata.uid = credentials.euid();
+        metadata.gid = credentials.egid();
+        *metadata
     }
 
     fn ino(&self) -> u64 {
--- a/test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
+++ b/test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+
+#include <fcntl.h>
+#include <sys/prctl.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "../../common/test.h"
+
+#define FIFO_PATH "/tmp/proc-fd-open-fifo-after-setid"
+
+FN_TEST(open_fifo_via_proc_fd_after_setid)
+{
+	char proc_fd_path[64];
+	int fifo_fd;
+	pid_t child;
+	int writer_fd;
+	int status;
+	char buf = 0;
+
+	TEST_SUCC(mkfifo(FIFO_PATH, 0644));
+
+	fifo_fd = TEST_SUCC(open(FIFO_PATH, O_PATH));
+	snprintf(proc_fd_path, sizeof(proc_fd_path), "/proc/self/fd/%d",
+		 fifo_fd);
+
+	child = TEST_SUCC(fork());
+	if (child == 0) {
+		int proc_fd;
+		int reopened_fd;
+		struct stat proc_dir_stat;
+		struct stat proc_stat;
+
+		CHECK(setresgid(-1, 65535, -1));
+		CHECK(setresuid(-1, 65535, -1));
+
+		/*
+		 * FIXME: Asterinas does not support the "dumpable" attribute
+		 * yet. So procfs files are always owned by the user
+		 * corresponding to the thread's effective UID.
+		 *
+		 * When resolving this FIXME, keep in mind that if the
+		 * "dumpable" attribute is not set, the procfs fd entry below
+		 * will be owned by the root user. However, the re-open should
+		 * still succeed because Linux uses special logic to bypass
+		 * the permission check.
+		 */
+#ifdef __asterinas__
+		CHECK(prctl(PR_SET_DUMPABLE, 1));
+		CHECK_WITH(prctl(PR_GET_DUMPABLE), _ret != 1);
+#else
+		CHECK_WITH(prctl(PR_GET_DUMPABLE), _ret != 1);
+		CHECK_WITH(stat("/proc/self/fd", &proc_dir_stat),
+			   proc_dir_stat.st_uid == 0 &&
+				   proc_dir_stat.st_gid == 0);
+		CHECK(prctl(PR_SET_DUMPABLE, 1));
+		CHECK_WITH(prctl(PR_GET_DUMPABLE), _ret == 1);
+#endif
+
+		CHECK_WITH(stat("/proc/self/fd", &proc_dir_stat),
+			   proc_dir_stat.st_uid == 65535 &&
+				   proc_dir_stat.st_gid == 65535);
+
+		// Open the procfs fd entry itself. With `O_PATH | O_NOFOLLOW`, `fstat`
+		// reports the owner of `/proc/self/fd/<n>` instead of the FIFO target.
+		proc_fd = CHECK(open(proc_fd_path, O_PATH | O_NOFOLLOW));
+		CHECK_WITH(fstat(proc_fd, &proc_stat),
+			   proc_stat.st_uid == 65535 &&
+				   proc_stat.st_gid == 65535);
+		CHECK(close(proc_fd));
+
+		// The FIFO target is owned by the root user, yet it is world-readable.
+		// So it can be re-opened.
+		reopened_fd = CHECK(open(proc_fd_path, O_RDONLY));
+		CHECK_WITH(fstat(reopened_fd, &proc_stat),
+			   proc_stat.st_uid == 0 && proc_stat.st_gid == 0);
+
+		CHECK_WITH(read(reopened_fd, &buf, 1), _ret == 1 && buf == 'X');
+		CHECK(close(reopened_fd));
+
+		exit(EXIT_SUCCESS);
+	}
+
+	writer_fd = TEST_SUCC(open(FIFO_PATH, O_WRONLY));
+	TEST_RES(write(writer_fd, "X", 1), _ret == 1);
+	TEST_SUCC(close(writer_fd));
+
+	TEST_RES(waitpid(child, &status, 0), _ret == child &&
+						     WIFEXITED(status) &&
+						     WEXITSTATUS(status) == 0);
+
+	TEST_SUCC(close(fifo_fd));
+	TEST_SUCC(unlink(FIFO_PATH));
+}
+END_TEST()
--- a/kernel/src/fs/fs_impls/procfs/pid/task/auxv.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/auxv.rs
@@ -8,6 +8,7 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/auxv` (and also `/proc/[pid]/auxv`).
@@ -21,6 +22,10 @@ impl AuxvFileOps {
 }
 
 impl FileOps for AuxvFileOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(process) = self.0.process() else {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/cmdline.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/cmdline.rs
@@ -8,6 +8,7 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/cmdline` (and also `/proc/[pid]/cmdline`).
@@ -21,6 +22,10 @@ impl CmdlineFileOps {
 }
 
 impl FileOps for CmdlineFileOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(process) = self.0.process() else {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/comm.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/comm.rs
@@ -8,6 +8,7 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/comm` (and also `/proc/[pid]/comm`).
@@ -21,6 +22,10 @@ impl CommFileOps {
 }
 
 impl FileOps for CommFileOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(process) = self.0.process() else {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/environ.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/environ.rs
@@ -8,6 +8,7 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/environ` (and also `/proc/[pid]/environ`).
@@ -21,6 +22,10 @@ impl EnvironFileOps {
 }
 
 impl FileOps for EnvironFileOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(process) = self.0.process() else {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/exe.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/exe.rs
@@ -8,6 +8,7 @@ use crate::{
         vfs::inode::{Inode, SymbolicLink},
     },
     prelude::*,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/exe` (and also `/proc/[pid]/exe`).
@@ -23,6 +24,10 @@ impl ExeSymOps {
 }
 
 impl SymOps for ExeSymOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.0.thread()
+    }
+
     fn read_link(&self) -> Result<SymbolicLink> {
         let Some(process) = self.0.process() else {
             return_errno_with_message!(Errno::ESRCH, "the process does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
@@ -20,6 +20,7 @@ use crate::{
     },
     prelude::*,
     process::posix_thread::AsPosixThread,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/fd` (and also `/proc/[pid]/fd`).
@@ -43,6 +44,10 @@ impl<T: FdOps> FdDirOps<T> {
 }
 
 impl<T: FdOps> DirOps for FdDirOps<T> {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.dir.thread()
+    }
+
     fn lookup_child(&self, this_dir: &ProcDir<Self>, name: &str) -> Result<Arc<dyn Inode>> {
         let file_desc = if let Ok(raw_fd) = name.parse::<RawFileDesc>()
             && let Ok(file_desc) = FileDesc::try_from(raw_fd)
@@ -225,6 +230,10 @@ impl FdOps for FileSymOps {
 }
 
 impl SymOps for FileSymOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.tid_dir_ops.thread()
+    }
+
     fn read_link(&self) -> Result<SymbolicLink> {
         let Some(thread) = self.tid_dir_ops.thread() else {
             return_errno_with_message!(Errno::ESRCH, "the thread does not exist");
@@ -288,6 +297,10 @@ impl FdOps for FileInfoOps {
 }
 
 impl FileOps for FileInfoOps {
+    fn owner_thread(&self) -> Option<Arc<Thread>> {
+        self.tid_dir_ops.thread()
+    }
+
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
         let Some(thread) = self.tid_dir_ops.thread() else {
             return_errno_with_message!(Errno::ESRCH, "the thread does not exist");
--- a/kernel/src/fs/fs_impls/procfs/pid/task/gid_map.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/gid_map.rs
@@ -11,10 +11,10 @@ use crate::{
     },
     prelude::*,
     process::Gid,
+    thread::Thread,
 };
 
 /// Represents the inode at `/proc/[pid]/task/[tid]/gid_map` (and also `/proc/[pid]/gid_map`).
-#[expect(dead_code)]
 pub struct GidMapFileOps(TidDirOps);
 
 impl GidMapFileOps {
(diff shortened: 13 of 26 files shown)


lrh2000:
> Linux also documents why `/proc/[pid]/fd` needs special handling, see [here](https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/fd.c#L327-L330):
> 
> ```c
> /* /proc/pid/fd needs a special permission handler so that a process can still
>  * access /proc/self/fd after it has executed a setuid().
>  */
> ```

This just misunderstands the Linux code. The Linux code wants to have that special handling when accessing the inode, not when the inode is created. The special handling has nothing to do with changes in this PR at all.

> the `procfs` template default `root:root`.

Why not fix this instead? Defaulting to `root:root` is not correct for many other files. `ls /proc/self` will report that all files are owned by the current user and not the root user.

Because a process's UID/GID can change while it is running, we may need to generate the metadata dynamically by overriding `Inode::metadata`:
https://github.com/asterinas/asterinas/blob/aa44d6e4975b86427e686b1996cbd0967da7b264/kernel/src/fs/fs_impls/procfs/template/file.rs#L77-L79

Even then, the security impact is still complex due to:
https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/base.c#L107-L114

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,72 @@
...
+
+		CHECK(close(ready_pipe[0]));
+		CHECK(setresgid(65535, 65535, 65535));
+		CHECK(setresuid(65535, 65535, 65535));
+
+		reopened_fd = CHECK(
+			open(proc_fd_path, O_RDONLY | O_CLOEXEC | O_NONBLOCK));
+		CHECK_WITH(fcntl(reopened_fd, F_SETFL,
+				 fcntl(reopened_fd, F_GETFL) & ~O_NONBLOCK),
+			   _ret != -1);
```
StevenJiang1110: I do not see any test coverage for the changes you made to `/proc/pid/fd`. Since you have reset the owner and GID of the files under `/proc/pid/fd`, here is what I would expect the tests to verify:

1. The parent process creates a file, so the file's owner is 0.
2. The child process calls setuid, so its UID is no longer 0.
3. The child process opens the file and obtains a file descriptor.
4. The child process reopens the file via `/proc/self/fd`.
5. The child process calls `fstat` on the reopened file and checks the returned UID — it should match the child process's UID, not 0.

The same checks should also be performed for GID.

kernel/src/fs/fs_impls/procfs/template/mod.rs
```diff
@@ -97,19 +145,25 @@ impl Common {
...
 
     fn group(&self) -> Result<Gid> {
-        Ok(self.metadata.read().gid)
+        Ok(self.dynamic_metadata().gid)
     }
 
     fn set_group(&self, gid: Gid) -> Result<()> {
+        // FIXME: Procfs permissions should be checked during each operation, not by relying on
+        // mutable inode ownership. See Linux comment:
+        // <https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/base.c#L107-L114>.
```
StevenJiang1110: Could you use a reference from https://elixir.bootlin.com instead? This is the site we consistently use across the repository for referencing Linux source code.

https://elixir.bootlin.com/linux/v6.13/source/fs/proc/base.c#L107

kernel/src/fs/fs_impls/procfs/template/mod.rs
```diff
@@ -14,41 +14,89 @@ use crate::{
...
+                    .upgrade()
+                    .expect("procfs child inode should not outlive its parent inode");
+                let parent_metadata = parent.metadata();
+                metadata.uid = parent_metadata.uid;
+                metadata.gid = parent_metadata.gid;
+            }
+            Self::Thread(thread) => {
+                let posix_thread = thread
+                    .as_posix_thread()
+                    .expect("procfs task owner should be backed by a POSIX thread");
```
StevenJiang1110: Please use `unwrap` instead of `expect`. Based on our current code conventions, `as_posix_thread` is always followed by `unwrap`. The additional panic message provided by expect doesn't add much value here.

kernel/src/fs/fs_impls/procfs/template/mod.rs
```diff
@@ -97,19 +145,25 @@ impl Common {
...
 
     fn owner(&self) -> Result<Uid> {
-        Ok(self.metadata.read().uid)
+        Ok(self.dynamic_metadata().uid)
     }
 
     fn set_owner(&self, uid: Uid) -> Result<()> {
+        // FIXME: Procfs permissions should be checked during each operation, not by relying on
+        // mutable inode ownership. See Linux comment:
+        // <https://github.com/torvalds/linux/blob/b4e07588e743c989499ca24d49e752c074924a9a/fs/proc/base.c#L107-L114>.
```
StevenJiang1110: Please avoid repeating the same comment in multiple places. Instead, consider putting it in a central location — for example, in the struct-level or module-level documentation, or in a relevant function's doc comment. Other places can then simply reference it.

kernel/src/fs/fs_impls/procfs/template/mod.rs
```diff
@@ -14,41 +14,89 @@ use crate::{
...
     fn fs(&self) -> Arc<dyn FileSystem> {
         self.fs.upgrade().unwrap()
     }
 
-    fn metadata(&self) -> Metadata {
-        *self.metadata.read()
+    fn dynamic_metadata(&self) -> Metadata {
+        let mut metadata = *self.metadata.read();
+        self.owner_source.apply_to(&mut metadata);
+        metadata
```
StevenJiang1110: Why do we have both `dynamic_metadata` and `metadata` (inherited from the `FileLike` trait)? This is quite confusing. When I first saw `dynamic_metadata`, I assumed there would be a separate function for directly accessing the metadata field. Since direct access to the `metadata` field is not intended, I think this function should simply be named `metadata`.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,69 @@
...
+	TEST_SUCC(close(writer_fd));
+
+	TEST_RES(waitpid(child, &status, 0), _ret == child &&
+						     WIFEXITED(status) &&
+						     WEXITSTATUS(status) == 0);
+
+	TEST_SUCC(close(fifo_fd));
+	TEST_SUCC(unlink(FIFO_PATH));
+}
+END_TEST()
```
StevenJiang1110, line 98: Since you mentioned that the owner of the `/proc/self/fd` directory also changes, could you add a test to cover the ownership of directories as well?

kernel/src/fs/fs_impls/procfs/template/mod.rs
```diff
@@ -14,41 +14,89 @@ use crate::{
...
+/// The source used to fill the dynamic owner fields of procfs metadata.
+pub(super) enum OwnerSource {
+    /// Keeps the owner stored in the inode metadata, which defaults to `root:root`.
+    Static,
+    /// Inherits the current owner from the parent inode.
+    ///
+    /// Most procfs child nodes use the same owner as their parent. This avoids storing duplicate
+    /// task references in child builders while still following a dynamic task owner through the
+    /// parent's metadata.
+    Inherit(Weak<dyn Inode>),
```
StevenJiang1110: I'm not quite sure why the `Inherit` variant is needed. If the parent's ownership source is `FromThread`, couldn't we just initialize the child as `FromThread` directly? And if the parent is `Static`, shouldn't the child just be `Static` as well?

On a related note, I'm not sure `Static` is necessary either. If the owner is a fixed value, it should just be written directly into the `metadata` field at initialization time.

Would it suffice to simply put an `owner_source: Option<Arc<Thread>>` field in the `Common` struct instead?

StevenJiang1110 (commented):
Thanks for fixing this problem!

kernel/src/fs/fs_impls/procfs/pid/task/mountstats.rs
```diff
@@ -43,6 +43,7 @@ impl MountStatsFileOps {
     pub fn new_inode(dir: &TidDirOps, parent: Weak<dyn Inode>) -> Arc<dyn Inode> {
         ProcFileBuilder::new(Self(dir.clone()), mkmod!(a+r))
             .parent(parent)
+            .owner_from_thread(dir.thread())
```
lrh2000: This is incorrect because executing a new program in a non-main thread replaces the main thread with the non-main thread under the same TID.

This means that `/proc/12345` does not correspond to one `Arc<Thread>`. The underlying thread may change to a different `Thread` object at any time.

Additionally, caching the thread in `Common` is odd. This information is duplicated in the `FileOps` or `DirOps`.

So it means that we really need to query the associated thread when we generate the metadata, instead of caching the `Thread`.

Note also that https://github.com/asterinas/asterinas/pull/3048 is refactoring the procfs architecture. This PR will have significant conflicts with it.

taosue: Thank you for the information. I didn't realize this PR would conflict significantly with #3048. Could you clarify whether the conflicts arise primarily from rebasing or from a substantial design change? If it's the latter, how about we apply the minor fix first and defer the redesign task until after the new release?

lrh2000: There are no substantial design changes related to this PR. The conflict arises because they touch similar lines, and both PRs modify many files under procfs. I think the conflicts are not difficult to address; they're just a bit annoying.

kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs
```diff
@@ -31,6 +32,10 @@ impl CgroupFileOps {
 }
 
 impl FileOps for CgroupFileOps {
+    fn associated_thread(&self) -> Option<Arc<Thread>> {
+        Some(self.0.thread())
```
tatetian: Is it "associated thread" or "owner thread"?

kernel/src/fs/fs_impls/procfs/template/file.rs
```diff
@@ -93,6 +93,17 @@ impl<F: FileOps + 'static> Inode for ProcFile<F> {
     fn set_ctime(&self, time: Duration);
     fn fs(&self) -> Arc<dyn FileSystem>;
 
+    fn metadata(&self) -> Metadata {
+        let mut metadata = self.common.metadata();
+        if let Some(thread) = self.inner.associated_thread() {
+            let posix_thread = thread.as_posix_thread().unwrap();
+            let credentials = posix_thread.credentials();
```
tatetian: euid and egid come from credentials, which in turn come from the "associated" or "owner" thread. So I am wondering if the `XxxOps` traits should return a thread, a set of credentials, or a pair of ids?

StevenJiang1110:
@taosue, this PR needs to be rebased. Since #3048 has been merged, the changes in this PR may also need to be adapted to the new API, which could result in fairly significant modifications compared to the current state.

taosue:
> @taosue, this PR needs to be rebased. Since #3048 has been merged, the changes in this PR may also need to be adapted to the new API, which could result in fairly significant modifications compared to the current state.

Thanks for the reminder, I have rebased this PR.

kernel/src/fs/fs_impls/procfs/pid/mod.rs
```diff
@@ -52,6 +56,12 @@ impl PidDirOps {
 }
 
 impl DirOps for PidDirOps {
+    fn owner_ids(&self) -> Option<(Uid, Gid)> {
+        let thread = self.0.thread()?;
+        let credentials = thread.as_posix_thread().unwrap().credentials();
+        Some((credentials.euid(), credentials.egid()))
```
lrh2000: This is repeated too many times in various file impls. Why are those lines repeated?

In fact, this logic will become more complex in the future. For example, https://man7.org/linux/man-pages/man5/proc_pid.5.html:

              The files inside each /proc/pid directory are normally
              owned by the effective user and effective group ID of the
              process.  However, as a security measure, the ownership is
              made root:root if the process's "dumpable" attribute is set
              to a value other than 1.

I think returning an `Arc<Thread>` is the simplest way to centralize handling of all complexities. It looks like this has been changed to `owner_ids` per @tatetian's request. Could @tatetian explain more about how to deal with this case in the future?

kernel/src/fs/fs_impls/procfs/template/mod.rs
```diff
@@ -50,6 +55,20 @@ impl Common {
         *self.metadata.read()
     }
 
+    fn metadata_with_owner(&self, owner_ids: Option<(Uid, Gid)>) -> Metadata {
+        if let Some((uid, gid)) = owner_ids {
+            let mut metadata = self.metadata.write();
+            // Cache the dynamic owner into the metadata so that if the task
+            // later exits and `owner_ids` returns `None`, subsequent calls
+            // fall back to the last known owner instead of root:root.
```
lrh2000: I think `root:root` is better than the last known owner. Instead of guessing an answer that is probably incorrect when we don't know it, I think we should choose the most conservative and safest option.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+			open(proc_fd_path, O_RDONLY | O_CLOEXEC | O_NONBLOCK));
+		TEST_SUCC(fcntl(reopened_fd, F_SETFL,
+				fcntl(reopened_fd, F_GETFL) & ~O_NONBLOCK));
+
+		TEST_RES(write(ready_pipe[1], "R", 1), _ret == 1);
+		TEST_SUCC(close(ready_pipe[1]));
+		TEST_RES(read(reopened_fd, &buf, 1), _ret == 1);
+		TEST_SUCC(close(reopened_fd));
+
+		_exit(__tests_failed == 0 && buf == 'X' ? 0 : 1);
```
lrh2000: - We do not use the `TEST_*` macros in the child process. Follow the convention of all other test cases by using the `CHECK` macros instead.
 - We do not use the test framework's internal variable `__tests_failed` outside the test framework.
 - We do not write code like `exit(cond ? 0 : 1)`, just use `CHECK_WITH(read(), cond)` instead.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+
+		TEST_SUCC(close(ready_pipe[0]));
+		TEST_SUCC(setresgid(65535, 65535, 65535));
+		TEST_SUCC(setresuid(65535, 65535, 65535));
+		uid = TEST_SUCC(getuid());
+		gid = TEST_SUCC(getgid());
+
+		TEST_RES(stat("/proc/self/fd", &proc_dir_stat),
+			 proc_dir_stat.st_uid == uid &&
+				 proc_dir_stat.st_gid == gid);
```
lrh2000: Have you ever run this test on a Linux system?
```
test_open_fifo_via_proc_fd_after_setid: `stat("/proc/self/fd", &proc_dir_stat)` failed [got Success, but `proc_dir_stat.st_uid == uid && proc_dir_stat.st_gid == gid` is false]
```

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+		TEST_SUCC(close(ready_pipe[1]));
+		TEST_RES(read(reopened_fd, &buf, 1), _ret == 1);
+		TEST_SUCC(close(reopened_fd));
+
+		_exit(__tests_failed == 0 && buf == 'X' ? 0 : 1);
+	}
+
+	TEST_SUCC(close(ready_pipe[1]));
+	TEST_RES(read(ready_pipe[0], &buf, 1), _ret == 1 && buf == 'R');
+	TEST_SUCC(close(ready_pipe[0]));
```
lrh2000: This test is far too complex. Please try to simplify it. For example, `ready_pipe` is unnecessary. You can simply use `open(FIFO_PATH)` without `O_NONBLOCK`.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+		struct stat proc_dir_stat;
+		struct stat proc_stat;
+		uid_t uid;
+		gid_t gid;
+
+		TEST_SUCC(close(ready_pipe[0]));
+		TEST_SUCC(setresgid(65535, 65535, 65535));
+		TEST_SUCC(setresuid(65535, 65535, 65535));
+		uid = TEST_SUCC(getuid());
+		gid = TEST_SUCC(getgid());
```
lrh2000: I think it is unnecessary.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+		// reports the owner of `/proc/self/fd/<n>` instead of the FIFO target.
+		proc_fd = TEST_SUCC(
+			open(proc_fd_path, O_PATH | O_NOFOLLOW | O_CLOEXEC));
+		TEST_RES(fstat(proc_fd, &proc_stat),
+			 proc_stat.st_uid == uid && proc_stat.st_gid == gid);
+		TEST_SUCC(close(proc_fd));
+
+		// Reopen the FIFO by following the procfs fd entry to its target. The
+		// reopened fd refers to the root-owned FIFO itself, so only reopen and I/O
+		// behavior is checked here, not its uid or gid.
```
lrh2000: Instead of writing a comment, why not just check if the owner is root?

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+		reopened_fd = TEST_SUCC(
+			open(proc_fd_path, O_RDONLY | O_CLOEXEC | O_NONBLOCK));
+		TEST_SUCC(fcntl(reopened_fd, F_SETFL,
+				fcntl(reopened_fd, F_GETFL) & ~O_NONBLOCK));
```
lrh2000: Why do you need to set the FD to blocking mode? And if you really need to, why did you open the file in non-blocking mode?

kernel/src/fs/fs_impls/procfs/pid/mod.rs
taosue: It seems hard to avoid having a helper here if we want to remove the repeated lines. Such a helper also gives us a single place to extend the ownership logic later, for example to handle dumpable. With that in mind, maybe keeping `metadata()` focused on owner IDs, instead of exposing the owner thread to it directly, is better?

kernel/src/fs/fs_impls/procfs/template/mod.rs
taosue: This cache is meant to align with Linux, because our tests show that Linux does not change the uid/gid when a thread is dead. So I’m not sure a root:root fallback would be the safer choice here, since it would introduce a state that Linux does not seem to use and could affect permission checks or user-visible ownership reporting.

kernel/src/fs/fs_impls/procfs/pid/mod.rs
lrh2000: Why not have a `owner_thread()` method here? That way, we'll have a centralized location to handle the conversion from the owner thread to the owner IDs. Then, no (public) helper methods would be needed.

kernel/src/fs/fs_impls/procfs/template/mod.rs
lrh2000: > This cache is meant to align with Linux

If the intention is to align with Linux, just say so. I don't see the point from the words you wrote in the code comments....

taosue: > If the intention is to align with Linux, just say so.

OK, I'll add it to the comments.

lrh2000: In addition,
 - `the task` -> `the thread` (We usually don't mention `task` in `aster-kernel`. Here, procfs shouldn't be too closely related to the OSTD's tasks anyway);
 - `root:root` -> `the root user`.

> Linux does not change the uid/gid when a thread is dead

If a thread initially belongs to user A and a file handle is opened under its procfs directory, `fstat` reports user A as the owner. If the thread's UID later changes to user B and the thread dies, `fstat` continues to report user A (instead of user B or the root user) as the owner.

I've confirmed this behavior locally. Maybe we can have a regression test? Thanks!

taosue: > I've confirmed this behavior locally.

Could you share your local demo? Per my test, `fstat` reports user B...

[proc_task_owner_demo.c](https://github.com/user-attachments/files/27516932/proc_task_owner_demo.c)

```
after set A        uid=65534 gid=65534
after set B        uid=65535 gid=65535
after exit         uid=65535 gid=65535
```

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
taosue: Thanks! I ran it on Linux and Linux reports `/proc/self/fd` and `/proc/self/fd/<n>` as `root:root` after `setuid/setgid`, so the previous expectation was wrong.

However, Linux can still reopen `/proc/self/fd/<n>` because procfs does not rely only on the inode owner/mode bits for these paths. It performs procfs-specific runtime permission checks for task-related entries. This is also why Linux has the warning...

Can we add a temporary FIXME for now? E.g.,
```diff
diff --git a/kernel/src/fs/fs_impls/procfs/pid/task/fd.rs b/kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
index 4c7f90bac..f9fcbc823 100644
--- a/kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/fd.rs
@@ -38,7 +38,13 @@ impl<T: FdOps> FdDirOps<T> {
             },
             parent,
             // Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/fs/proc/base.c#L3317>
-            mkmod!(u+rx),
+            //
+            // FIXME: Linux reports `/proc/[pid]/fd` as root user with mode 0500, but procfs
+            // performs runtime permission checks instead of relying only on inode ownership.
+            // Linux comment: <https://elixir.bootlin.com/linux/v6.13/source/fs/proc/base.c#L107>.
+            // Asterinas currently uses generic DAC checks during path lookup, so use searchable
+            // directory permissions here while keeping the inode owner as root.
+            mkmod!(a+rx),
         )
     }
 }
```

kernel/src/fs/fs_impls/procfs/template/mod.rs
lrh2000: > after set B        uid=65535 gid=65535

The cache be updated if you query after `setresuid`? If we don't query after `setresuid`, the UID will remain A even after the thread exits, which is what I mean. Am I misunderstanding anything?

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
lrh2000: > Thanks! I ran it on Linux and Linux reports `/proc/self/fd` and `/proc/self/fd/<n>` as `root:root` after `setuid/setgid`, so the previous expectation was wrong.

Well, I think... Not really. This is the behavior I mentioned in https://github.com/asterinas/asterinas/pull/3164#discussion_r3200697552.

What we're missing here is an explicitly `prctl(PR_SET_DUMPABLE, 1)` after `setresuid` in the regression test. The FIXME should be written as:
```c
// FIXME: ...
#ifdef __asterinas__
// Aserinas does not support the "dumpable" attribute yet.
TEST_SUCC(prctl(PR_SET_DUMPABLE, 1));
TEST_SUCC(prctl(PR_GET_DUMPABLE), _ret == 0);

// So the procfs files are always owned by
// the user corresponding to the thread's effective UID.
TEST_RES(stat, uid == guest);
#else
// The procfs files are owned by the root user
// if the "dumpable" attribute is not set.
TEST_SUCC(prctl(PR_GET_DUMPABLE), _ret == 0);
TEST_RES(stat, uid == root);

// The procfs files are owned by
// the user corresponding to the thread's effective UID
// if the "dumpable" attribute is set.
TEST_SUCC(prctl(PR_SET_DUMPABLE, 1));
TEST_RES(stat, uid == guest);
#endif
```

kernel/src/fs/fs_impls/procfs/template/mod.rs
taosue: > If we don't query after `setresuid`, the UID will remain A even after the thread exits

Do you mean that we should not query the fd after switching to user B? I tested that case, but after the thread exits, `fstat()` on `/proc/self/task/<tid>` still reports user B.

```
after set A              uid=65534 gid=65534
after exit               uid=65535 gid=65535
```

lrh2000: Oh I see the problem...

Well, it seems that the Linux implementation is complex. Perhaps it's just full of implementation details that are hard to figure out, and it may be unnecessary to implement all the corner cases.

This case is what I originally meant:
```patch
diff --git a/proc_task_owner_demo.c b/proc_task_owner_demo2.c
index aa0b1bd..f886a2e 100644
--- a/proc_task_owner_demo.c
+++ b/proc_task_owner_demo2.c
@@ -6,6 +6,8 @@
 #include <stdlib.h>
 #include <sys/stat.h>
 #include <sys/syscall.h>
+#include <sys/prctl.h>
+#include <linux/prctl.h>
 #include <unistd.h>
 
 #define USER_A 65534
@@ -60,6 +62,7 @@ static void *thread_fn(void *arg)
 		perror("set A");
 		exit(1);
 	}
+	prctl(PR_SET_DUMPABLE, 1);
 	send_byte(to_main[1], 'A');
 
 	recv_byte(to_thread[0], 'B');
@@ -68,6 +71,7 @@ static void *thread_fn(void *arg)
 		perror("set B");
 		exit(1);
 	}
+	prctl(PR_SET_DUMPABLE, 1);
 	send_byte(to_main[1], 'B');
 
 	recv_byte(to_thread[0], 'X');
@@ -90,8 +94,8 @@ int main(void)
 	}
 
 	recv_byte(to_main[0], 'A');
-	snprintf(path, sizeof(path), "/proc/self/task/%ld", tid);
-	fd = open(path, O_RDONLY | O_DIRECTORY);
+	snprintf(path, sizeof(path), "/proc/self/task/%ld/stat", tid);
+	fd = open(path, O_RDONLY);
 	if (fd < 0) {
 		perror("open");
 		return 1;
@@ -100,6 +104,9 @@ int main(void)
 
 	send_byte(to_thread[1], 'B');
 	recv_byte(to_main[0], 'B');
+#ifdef REFRESH_B
+	close(open(path, O_RDONLY));
+#endif
 	show_owner("after set B", fd);
 
 	send_byte(to_thread[1], 'X');
```

I got:
```
> gcc proc_task_owner_demo2.c -o /tmp/x && sudo /tmp/x
after set A        uid=65534 gid=65534
after set B        uid=65534 gid=65534
after exit         uid=65534 gid=65534
> gcc proc_task_owner_demo2.c -DREFRESH_B -o /tmp/x && sudo /tmp/x
after set A        uid=65534 gid=65534
after set B        uid=65535 gid=65535
after exit         uid=65535 gid=65535
```

So:
 - `task/[tid]` behaves differently from `task/[tid]/stat`.
 - The ownership of `task/[tid]` seems to be automatically refreshed, but the ownership of `task/[tid]/stat` is cached and only refreshed upon `open()`, `readdir()`, etc., but not `fstat()`.

I don't believe we can implement this specific Linux behavior with our current architecture. Refactoring the code for this corner case may also be unnecessary. So, a best-effort implementation in this PR is fine.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+
+	child = TEST_SUCC(fork());
+	if (child == 0) {
+		int proc_fd;
+		int reopened_fd;
+		struct stat proc_dir_stat;
+		struct stat proc_stat;
+
+		CHECK(setresgid(65535, 65535, 65535));
+		CHECK(setresuid(65535, 65535, 65535));
```
lrh2000: I don't think you need to change the real IDs or the saved-set IDs, as this test is designed to verify that the procfs files are owned by the user corresponding to the effective IDs.

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+
+/*
+ * FIXME: Asterinas does not support the "dumpable" attribute
+ * yet. So procfs files are always owned by the user
+ * corresponding to the thread's effective UID.
+ */
```
lrh2000: Should we indent those lines?

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,90 @@
...
+
+		// Open the procfs fd entry itself. With `O_PATH | O_NOFOLLOW`, `fstat`
+		// reports the owner of `/proc/self/fd/<n>` instead of the FIFO target.
+		proc_fd = CHECK(open(proc_fd_path, O_PATH | O_NOFOLLOW));
+		CHECK_WITH(fstat(proc_fd, &proc_stat),
+			   proc_stat.st_uid == 65535 &&
+				   proc_stat.st_gid == 65535);
+		CHECK(close(proc_fd));
+
+		reopened_fd = CHECK(open(proc_fd_path, O_RDONLY));
```
lrh2000, line 77:
```c
		// The FIFO target is owned by the root user, yet it is world-readable.
		// So it can be reopened.
		reopened_fd = CHECK(open(proc_fd_path, O_RDONLY));
		CHECK_WITH(fstat(reopened_fd, &proc_stat),
			   proc_stat.st_uid == 0 && proc_stat.st_gid == 0);
```

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
lrh2000:
```suggestion
 *
 * When resolving this FIXME, keep in mind that if the
 * "dumpable" attribute is not set, the procfs fd entry below
 * will be owned by the root user. However, the re-open should
 * still succeed because Linux uses special logic to bypass
 * the permission check.
 */
```

lrh2000 (commented):
Thanks for the revision. Just some minor comments about the tests and/or the comment. Otherwise the changes look good to me.

lrh2000:
Oh but you really should run the kernel yourself and verify that you don't miss anything under `/proc/[pid]`.

From what I can see, you are missing `/proc/[pid]/ns/*`. This behavior differs from that of Linux.
```
~ # adduser test
adduser: /home/test: No such file or directory
Changing password for test
New password: 
Bad password: too short
Retype password: 
passwd: password for test changed by root
~ # su test
/ $ ls -al /proc/self/*
-r--------    1 test     test             0 May 10 09:00 /proc/self/auxv
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/cgroup
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/cmdline
-rw-r--r--    1 test     test             0 May 10 09:00 /proc/self/comm
-r--------    1 test     test             0 May 10 09:00 /proc/self/environ
lrwxrwxrwx    1 test     test             0 May 10 09:00 /proc/self/exe -> /usr/bin/busybox
-rw-r--r--    1 test     test             0 May 10 09:00 /proc/self/gid_map
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/maps
-rw-------    1 test     test             0 May 10 09:00 /proc/self/mem
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/mountinfo
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/mounts
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/mountstats
-rw-r--r--    1 test     test             0 May 10 09:00 /proc/self/oom_score_adj
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/stat
-r--r--r--    1 test     test             0 May 10 09:00 /proc/self/status
-rw-r--r--    1 test     test             0 May 10 09:00 /proc/self/uid_map

/proc/self/fd:
total 0
dr-x------    2 test     test             2 May 10 09:00 .
dr-xr-xr-x    2 test     test             2 May 10 09:00 ..
lrwx------    1 test     test             0 May 10 09:00 0 -> /dev/pts/0
lrwx------    1 test     test             0 May 10 09:00 1 -> /dev/pts/0
lrwx------    1 test     test             0 May 10 09:00 2 -> /dev/pts/0
ls: /proc/self/fd/3: cannot read link: No such file or directory
lr-x------    1 test     test             0 May 10 09:00 3

/proc/self/fdinfo:
total 0
dr-x------    2 test     test             2 May 10 09:00 .
dr-xr-xr-x    2 test     test             2 May 10 09:00 ..
-r--r--r--    1 test     test             0 May 10 09:00 0
-r--r--r--    1 test     test             0 May 10 09:00 1
-r--r--r--    1 test     test             0 May 10 09:00 2
-r--r--r--    1 test     test             0 May 10 09:00 3

/proc/self/ns:
total 0
dr-x--x--x    2 test     test             2 May 10 09:00 .
dr-xr-xr-x    2 test     test             2 May 10 09:00 ..
lrwxrwxrwx    1 root     root             0 May 10 09:00 cgroup -> cgroup:[4]
lrwxrwxrwx    1 root     root             0 May 10 09:00 ipc -> ipc:[5]
lrwxrwxrwx    1 root     root             0 May 10 09:00 mnt -> mnt:[3]
lrwxrwxrwx    1 root     root             0 May 10 09:00 user -> user:[2]
lrwxrwxrwx    1 root     root             0 May 10 09:00 uts -> uts:[6]

/proc/self/task:
total 0
dr-xr-xr-x    2 test     test             2 May 10 09:00 .
dr-xr-xr-x    2 test     test             2 May 10 09:00 ..
dr-xr-xr-x    2 test     test             2 May 10 09:00 13
```

test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c
```diff
@@ -0,0 +1,98 @@
...
+		// Open the procfs fd entry itself. With `O_PATH | O_NOFOLLOW`, `fstat`
+		// reports the owner of `/proc/self/fd/<n>` instead of the FIFO target.
+		proc_fd = CHECK(open(proc_fd_path, O_PATH | O_NOFOLLOW));
+		CHECK_WITH(fstat(proc_fd, &proc_stat),
+			   proc_stat.st_uid == 65535 &&
+				   proc_stat.st_gid == 65535);
+		CHECK(close(proc_fd));
+
+		// The FIFO target is owned by the root user, yet it is world-readable.
+		// So it can be reopened.
```
lrh2000: This issue is not worth addressing separately, but since you need another push anyway, let's fix the wording:
```suggestion
		// So it can be re-opened.
```
to align with:
```
kernel/src/fs/fs_impls/pseudofs/mod.rs:            "the pseudo inode is not re-openable",
test/initramfs/src/regression/fs/procfs/proc_fd_open_fifo_after_setid.c:                * will be owned by the root user. However, the re-open should
```

(My bad in the suggestion..)

taosue:
> From what I can see, you are missing `/proc/[pid]/ns/*`. This behavior differs from that of Linux.

On my Linux host, I get a different result for `/proc/self/ns/*`:

```
/proc/self/ns:
total 0
dr-x--x--x 2 test test 0 May 11 01:45 .
dr-xr-xr-x 9 test test 0 May 11 01:45 ..
lrwxrwxrwx 1 test test 0 May 11 01:45 cgroup -> 'cgroup:[4026531835]'
lrwxrwxrwx 1 test test 0 May 11 01:45 ipc -> 'ipc:[4026531839]'
lrwxrwxrwx 1 test test 0 May 11 01:45 mnt -> 'mnt:[4026531841]'
lrwxrwxrwx 1 test test 0 May 11 01:45 net -> 'net:[4026531840]'
lrwxrwxrwx 1 test test 0 May 11 01:45 pid -> 'pid:[4026531836]'
lrwxrwxrwx 1 test test 0 May 11 01:45 pid_for_children -> 'pid:[4026531836]'
lrwxrwxrwx 1 test test 0 May 11 01:45 time -> 'time:[4026531834]'
lrwxrwxrwx 1 test test 0 May 11 01:45 time_for_children -> 'time:[4026531834]'
lrwxrwxrwx 1 test test 0 May 11 01:45 user -> 'user:[4026531837]'
lrwxrwxrwx 1 test test 0 May 11 01:45 uts -> 'uts:[4026531838]'
```

Maybe our BusyBox `ls` implementations behave differently here. On my host, `lstat("/proc/self/ns/*")` reports the task owner, while `stat("/proc/self/ns/*")` follows the magic link and reports `root:root` for the namespace inode.

So I think there are two different objects involved here:
  - the procfs symlink itself: `lstat("/proc/self/ns/*")`
  - the followed namespace inode: `stat("/proc/self/ns/*")`

Could you confirm whether your observation is for the symlink itself or for the followed namespace inode?

lrh2000:
> On my Linux host, I get a different result for `/proc/self/ns/*`:

This is what I mean.

https://github.com/asterinas/asterinas/pull/3164#issuecomment-4414903702 is the output of Asterinas after this PR, which is incorrect.

Sorry for the confusion.

lrh2000:
> [#3164 (comment)](https://github.com/asterinas/asterinas/pull/3164#issuecomment-4414903702) is the output of Asterinas after this PR

I thought it was obvious because you can see that the procfs mentioned in this comment is far from complete, so it cannot be Linux's output. But I should make this clearer.

taosue:
> > [#3164 (comment)](https://github.com/asterinas/asterinas/pull/3164#issuecomment-4414903702) is the output of Asterinas after this PR
> 

Thanks for catching this. You are right, this is indeed a missing case in my change. I will fix `/proc/[pid]/ns/*` so that the procfs symlink itself is also owned by the task owner.

> I thought it was obvious because you can see that the procfs mentioned in this comment is far from complete, so it cannot be Linux's output. But I should make this clearer.

I mistakenly thought that the output was from Linux and that you had only copied the subset of procfs entries currently supported by Asterinas...

lrh2000 (approved):

PR #3166, opened 2026-04-21
Fix `Cursor::jump` documentation

rikosellic:
- Document missing panic conditions for `vm_space::Cursor::jump` and `vm_space::CursorMut::jump`.
- Fix the internal documentation for `page_table::CursorMut::jump`.

### DIFF
--- a/ostd/src/mm/vm_space.rs
+++ b/ostd/src/mm/vm_space.rs
@@ -291,6 +291,12 @@ impl Cursor<'_> {
     }
 
     /// Jumps to the virtual address.
+    ///
+    /// If the target address is out of the range, this method will return `Err`.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if the address has bad alignment.
     pub fn jump(&mut self, va: Vaddr) -> Result<()> {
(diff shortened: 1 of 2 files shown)


ostd/src/mm/vm_space.rs
```diff
@@ -291,6 +291,11 @@ impl Cursor<'_> {
     }
 
     /// Jumps to the virtual address.
+    /// If the target address is out of the range, this method will return `Err`.
```
junyang-zh, line 295: Need to add a blank line before line 294.

lrh2000: Seems to be copied from:
https://github.com/asterinas/asterinas/blob/aa44d6e4975b86427e686b1996cbd0967da7b264/ostd/src/mm/page_table/cursor/mod.rs#L247-L248

So let's add a blank line there as well.

rikosellic: Added!

junyang-zh (approved):
LGTM! Thanks for your contribution.

PR #3168, opened 2026-04-22
Refactor mount filesystem creation with `MountContext`

li041:
## Summary

  Refactor filesystem creation during `mount(2)` by introducing
  `MountContext` and passing it through `FsType::create`.

  Instead of threading mount flags, source/data, and block-device lookup
  through separate parameters, the mount path now builds a single context
  object and each filesystem implementation pulls the inputs it needs
  from that context.

  ## Why

  The old interface resolved `source` into a block device in the generic
  mount path and passed the result into `FsType::create`.

  That is too specific for a generic mount API. `source` is interpreted
  differently by different filesystems: for example, ext2/exfat treat it
  as a device path, while virtiofs treats it as a tag. Those semantics
  belong to the filesystem implementation, not to the shared mount layer.

  This change keeps mount inputs in their original form and passes them
  through `MountContext`, so each filesystem can interpret `source` in
  the way that matches its own mount semantics. Common code no longer
  assumes that `source` always means “a block device”.
  ## Notes

  - block-backed filesystems now resolve their backing device through the
    shared mount context helper
  - virtiofs now uses the mount source directly as its tag input

### DIFF
--- a/kernel/src/fs/fs_impls/cgroupfs/fs.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/fs.rs
@@ -2,7 +2,6 @@
 
 use alloc::sync::Arc;
 
-use aster_block::BlockDevice;
 use aster_systree::EmptyNode;
 use ostd::task::Task;
 use spin::Once;
@@ -14,12 +13,11 @@ use crate::{
         pseudofs::AnonDeviceId,
         utils::systree_inode::SysTreeInodeTy,
         vfs::{
-            file_system::{FileSystem, FsEventSubscriberStats, FsFlags, SuperBlock},
+            file_system::{FileSystem, FsEventSubscriberStats, SuperBlock},
             inode::Inode,
-            registry::{FsProperties, FsType},
+            registry::{FsCreationCtx, FsProperties, FsType},
         },
     },
-    prelude::*,
     process::posix_thread::AsThreadLocal,
 };
 
@@ -95,13 +93,8 @@ impl FsType for CgroupFsType {
         FsProperties::empty()
     }
 
-    fn create(
-        &self,
-        _flags: FsFlags,
-        _args: Option<CString>,
-        _disk: Option<Arc<dyn BlockDevice>>,
-    ) -> Result<Arc<dyn FileSystem>> {
-        Ok(CgroupFs::singleton().clone() as _)
+    fn create(&self, _fs_creation_ctx: &FsCreationCtx) -> Result<Arc<dyn FileSystem>> {
+        Ok(CgroupFs::singleton().clone())
     }
 
     fn sysnode(&self) -> Option<Arc<dyn aster_systree::SysNode>> {
--- a/kernel/src/fs/fs_impls/exfat/fs.rs
+++ b/kernel/src/fs/fs_impls/exfat/fs.rs
@@ -27,10 +27,10 @@ use crate::{
     fs::{
         exfat::{constants::*, inode::Ino},
         vfs::{
-            file_system::{FileSystem, FsEventSubscriberStats, FsFlags, SuperBlock},
+            file_system::{FileSystem, FsEventSubscriberStats, SuperBlock},
             inode::Inode,
             page_cache::{CachePage, PageCache, PageCacheBackend},
-            registry::{FsProperties, FsType},
+            registry::{FsCreationCtx, FsProperties, FsType},
         },
     },
     prelude::*,
@@ -478,13 +478,11 @@ impl FsType for ExfatType {
         FsProperties::NEED_DISK
     }
 
-    fn create(
-        &self,
-        _flags: FsFlags,
-        _args: Option<CString>,
-        disk: Option<Arc<dyn BlockDevice>>,
-    ) -> Result<Arc<dyn FileSystem>> {
-        ExfatFs::open(disk.unwrap(), ExfatMountOptions::default()).map(|fs| fs as _)
+    fn create(&self, fs_creation_ctx: &FsCreationCtx) -> Result<Arc<dyn FileSystem>> {
+        Ok(ExfatFs::open(
+            fs_creation_ctx.resolve_block_device()?,
+            ExfatMountOptions::default(),
+        )?)
     }
 
     fn sysnode(&self) -> Option<Arc<dyn aster_systree::SysNode>> {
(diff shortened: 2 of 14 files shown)


kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -43,6 +40,73 @@ pub trait FsType: Send + Sync + 'static {
     fn sysnode(&self) -> Option<Arc<dyn SysNode>>;
 }
 
+/// Mount inputs for creating a filesystem instance.
+pub struct MountContext<'a> {
```
cchanging: Since this abstraction includes the input for creating the filesystem, would `FsCreationCtx` be a better name? `MountContext` gives the impression that it would also contain things like mount flags.

```rust
/// A context that describes the inputs used to create a filesystem instance.
///
/// This context will be used by [`FsType::create`].
pub struct FsCreationCtx<'a> 
```

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -43,6 +40,73 @@ pub trait FsType: Send + Sync + 'static {
...
+        Self {
+            flags,
+            source,
+            args,
+            task_ctx,
+        }
+    }
+
+    /// Returns the user-supplied mount source.
+    pub fn source(&self) -> Option<&'a str> {
```
cchanging: I think the visibility of these APIs could be `pub(in crate::fs)`.

cchanging (commented):
Thanks for further cleaning up the `FsType::create` interface! This change makes sense to me. It gives us more flexibility in how we use these parameters when creating certain filesystems, such as `virtiofs`.

kernel/src/fs/fs_impls/cgroupfs/fs.rs
```diff
@@ -95,12 +93,7 @@ impl FsType for CgroupFsType {
         FsProperties::empty()
     }
 
-    fn create(
-        &self,
-        _flags: FsFlags,
-        _args: Option<CString>,
-        _disk: Option<Arc<dyn BlockDevice>>,
-    ) -> Result<Arc<dyn FileSystem>> {
+    fn create(&self, _mount_ctx: &FsCreationCtx<'_>) -> Result<Arc<dyn FileSystem>> {
```
StevenJiang1110:
```suggestion
    fn create(&self, _mount_ctx: &FsCreationCtx) -> Result<Arc<dyn FileSystem>> {
```

The lifetime can be removed. Please also fix similar cases.

kernel/src/fs/fs_impls/exfat/fs.rs
```diff
@@ -478,13 +478,12 @@ impl FsType for ExfatType {
...
-        _args: Option<CString>,
-        disk: Option<Arc<dyn BlockDevice>>,
-    ) -> Result<Arc<dyn FileSystem>> {
-        ExfatFs::open(disk.unwrap(), ExfatMountOptions::default()).map(|fs| fs as _)
+    fn create(&self, mount_ctx: &FsCreationCtx<'_>) -> Result<Arc<dyn FileSystem>> {
+        ExfatFs::open(
+            mount_ctx.resolve_block_device()?,
+            ExfatMountOptions::default(),
+        )
+        .map(|fs| fs as _)
```
StevenJiang1110:
```suggestion
        Ok(ExfatFs::open(
            mount_ctx.resolve_block_device()?,
            ExfatMountOptions::default(),
        )?)
```

We can let the compiler perfrom the type conversion. Please also fix similar cases.

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -43,6 +40,75 @@ pub trait FsType: Send + Sync + 'static {
...
+    /// Returns the user-supplied mount source.
+    pub(in crate::fs) fn source(&self) -> Option<&'a str> {
+        self.source
+    }
+
+    /// Returns the user-supplied mount flags.
+    #[expect(
+        dead_code,
+        reason = "`FsFlags` is part of the fs creation context, but current filesystem implementations do not consume it yet"
+    )]
```
StevenJiang1110: Remove the reason message. It seems meaningless.

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -43,6 +40,75 @@ pub trait FsType: Send + Sync + 'static {
...
+        Self {
+            flags,
+            source,
+            args,
+            task_ctx,
+        }
+    }
+
+    /// Returns the user-supplied mount source.
+    pub(in crate::fs) fn source(&self) -> Option<&'a str> {
```
StevenJiang1110:
```suggestion
    pub(in crate::fs) fn source(&self) -> Option<&str> {
```

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -43,6 +40,75 @@ pub trait FsType: Send + Sync + 'static {
...
+    #[expect(
+        dead_code,
+        reason = "`FsFlags` is part of the fs creation context, but current filesystem implementations do not consume it yet"
+    )]
+    pub(in crate::fs) fn flags(&self) -> FsFlags {
+        self.flags
+    }
+
+    /// Returns the filesystem-specific mount arguments.
+    pub(in crate::fs) fn args(&self) -> Option<&'a CStr> {
```
StevenJiang1110:
```suggestion
    pub(in crate::fs) fn args(&self) -> Option<&CStr> {
```

kernel/src/syscall/mount.rs
```diff
@@ -242,32 +242,8 @@ fn open_fs(
...
-            return_errno_with_message!(Errno::ENODEV, "the device is not found");
-        }
-
-        device
-    } else {
-        None
-    };
-
-    fs_type.create(flags.into(), data, disk)
+    let mount_ctx = FsCreationCtx::new(source, flags.into(), data.as_deref(), ctx);
```
StevenJiang1110: Since the struct is `FsCreationCtx`, the variable should be named as `fs_creation_ctx` instead of `mount_ctx`. Please also fix similar cases.

StevenJiang1110 (approved):
LGTM! Just a few minor comments.

Let @cchanging perfrom the final check and merge.

cchanging (approved):

PR #3169, opened 2026-04-22
Add a read_random implementation for riscv architecture

le-monde-bleu:
This PR adds a implementation of `read_random` for RISC-V architecture. It is a function to read hardware generated 64-bit random value. This implementation uses the RISC-V Entropy Source Extension (Zkr) which provides access to a physical entropy source via the `seed` CSR and follows a similar retry pattern as the x86 `RDRAND` version, adapted for RISC-V's 16-bit entropy output. 

The source of reference is [RISC-V Cryptography Extensions Volume I: Scalar & Entropy Source Instructions](https://docs.riscv.org/reference/isa/extensions/crypto-scalar/_attachments/riscv-crypto-spec-scalar.pdf)

The logic goes like:

1. Check hardware support for `Zkr` extension
2. Prepare to build a 64-bit random value from 16-bit pieces.  
- The `seed` CSR only gives 16 bits of entropy per successful read (`ES16` status). So it needs 4 successful chunks to assemble one `u64`.
3. For each of the 4 chunks, poll with retry.
4. Decode status from bits 31:30.
5. Fail if any chunk cannot be obtained in time.
6. Return complete 64-bit result.

### DIFF
--- a/ostd/src/arch/riscv/mod.rs
+++ b/ostd/src/arch/riscv/mod.rs
@@ -103,9 +103,75 @@ pub fn read_tsc() -> u64 {
 /// Reads a hardware generated 64-bit random value.
 ///
 /// Returns `None` if no random value was generated.
+///
+/// This implementation uses the RISC-V Entropy Source Extension (Zkr) which
+/// provides access to a physical entropy source via the `seed` CSR.
+///
+/// Reference:
+/// - "RISC-V Cryptography Extensions Volume I: Scalar & Entropy Source Instructions"
+///   - <https://docs.riscv.org/reference/isa/extensions/crypto-scalar/_attachments/riscv-crypto-spec-scalar.pdf>
+///   - Section 2.9: Zkr - Entropy Source Extension
+///   - Section 4.1: The `seed` CSR
 pub fn read_random() -> Option<u64> {
-    // FIXME: Implement a hardware random number generator on RISC-V platforms.
-    None
+    use cpu::extension::{IsaExtensions, has_extensions};
+
+    // Check if the Zkr (Entropy Source) extension is available.
+    if !has_extensions(IsaExtensions::ZKR) {
+        return None;
+    }
+
+    const RETRY_LIMIT: usize = 10;
+
+    const OPST_SHIFT: usize = 30;
+    const OPST_MASK: usize = 0b11 << OPST_SHIFT;
+    const ENTROPY_MASK: usize = 0xFFFF;
+
+    const OPST_ES16: usize = 0b10; // Status indicating entropy is available
+    const OPST_WAIT: usize = 0b01; // Status indicating entropy source is busy (waiting)
+    const OPST_BIST: usize = 0b00; // Status indicating entropy source is in self-test
+    const OPST_DEAD: usize = 0b11; // Status indicating entropy source has failed
+
+    // Read 16-bit entropy from hardware.
+    let read_entropy = || -> Option<u16> {
+        for _ in 0..RETRY_LIMIT {
+            let seed_val: usize;
+            // SAFETY: We've checked that the Zkr extension is available.
+            // Therefore, this is a valid instruction to access the `seed` CSR.
+            unsafe {
+                // The Zkr spec requires using `csrrw` to access `seed` CSR,
+                // which signals polling and flushing. See the spec for more
+                // information.
+                core::arch::asm!(
+                    "csrrw {0}, seed, zero",
+                    out(reg) seed_val,
+                    options(nomem, nostack)
+                );
+            }
+
+            // Extract status bits (bits 31:30).
+            let status = (seed_val & OPST_MASK) >> OPST_SHIFT;
+
+            if status == OPST_ES16 {
+                // Entropy is available: extract 16 bits (bits 15:0).
+                return Some((seed_val & ENTROPY_MASK) as u16);
+            } else if status == OPST_DEAD {
+                // Entropy source failed permanently.
+                return None;
+            } else if status == OPST_WAIT || status == OPST_BIST {
+                // Entropy source busy or running self-test: retry.
+                core::hint::spin_loop();
+            }
+        }
+        None
+    };
+
+    // Collect 4 chunks of 16-bit entropy to form a 64-bit random value.
+    let result = (read_entropy()? as u64)
+        | ((read_entropy()? as u64) << 16)
+        | ((read_entropy()? as u64) << 32)
+        | ((read_entropy()? as u64) << 48);
+
+    Some(result)
 }
 
 pub(crate) fn enable_cpu_features() {
--- a/tools/qemu_args.sh
+++ b/tools/qemu_args.sh
@@ -61,7 +61,7 @@ if [ "$1" = "riscv" ]; then
     # The ordering below ensures `x1` (ext2.img) is discovered before `x0`, maintaining this assumption.
     # TODO: Once UUID-based mounting is implemented, this strict ordering will no longer be required.
     QEMU_ARGS="\
-        -cpu rv64,svpbmt=true \
+        -cpu rv64,svpbmt=true,zkr=true \
         -machine virt \
         -m ${MEM:-8G} \
         -smp ${SMP:-1} \


le-monde-bleu:
Switched from `"csrrw {0}, 0x015, x0"` to `"csrrw {0}, 0x015, zero"` for consistency and robustness.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
 /// Returns `None` if no random value was generated.
+///
+/// This implementation uses the RISC-V Entropy Source Extension (Zkr) which
+/// provides access to a physical entropy source via the `seed` CSR.
+///
+/// Note: The Zkr specification explicitly requires using `csrrw` (not read-only
+/// CSR instructions) to access the `seed` CSR. The write signals polling and
+/// triggers the wipe-on-read security mechanism. Using `csrrs`/`csrrc` with rs1=x0
+/// would raise an illegal-instruction exception per the spec.
+///
```
lrh2000:
```suggestion
```

There is no need to include implementation details in documentation comments. Doc comments should only provide information that is useful to the caller. A caller should not need to learn how the method is implemented.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+/// Reference:
+/// - "RISC-V Cryptography Extensions Volume I: Scalar & Entropy Source Instructions"
+///   - <https://docs.riscv.org/reference/isa/extensions/crypto-scalar/_attachments/riscv-crypto-spec-scalar.pdf>
+///   - Section 2.9: Zkr - Entropy Source Extension
+///   - Section 4.1: The `seed` CSR
+///   - The `seed` CSR at address 0x015 provides up to 16 bits of entropy
```
lrh2000:
```suggestion
```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+///
+/// The implementation follows a similar retry pattern as the x86 RDRAND version,
+/// adapted for RISC-V's 16-bit entropy output.
```
lrh2000:
```suggestion
```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
-    None
+    use cpu::extension::{IsaExtensions, has_extensions};
+
+    // Check if the Zkr (Entropy Source) extension is available
+    if !has_extensions(IsaExtensions::ZKR) {
+        return None;
+    }
+
+    // The seed CSR provides 16 bits of entropy at a time.
+    // We need 4 successful reads to construct a 64-bit value.
```
lrh2000: This comment is in the wrong place. It is not related to the constants defined below.

Actually I think it should be removed since you've already written "Collect 4 chunks of 16-bit entropy to form a 64-bit value" below.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+            // (per Zkr spec, a write signals that the core is ready for entropy).
+            // We read into usize to match the native XLEN register width (RV64/RV32).
+            let seed_val: usize;
+            // IMPORTANT: The Zkr spec requires using `csrrw` to access seed CSR.
+            // The write operation signals polling and triggers wipe-on-read.
+            // Read-only CSR instructions (csrrs/csrrc with rs1=x0) would trap.
+            // The value written (zero) is ignored by hardware per the spec.
+            unsafe {
+                core::arch::asm!(
+                    // Use `zero` instead of `x0` here for consistency and robustness
```
lrh2000: > for consistency and robustness

What is its consistency, and why is it robust?

Actually, I think using `x0` is fine. After all, it's the exact instruction suggested by the spec.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+                    "csrrw {0}, 0x015, zero",
```
lrh2000: Instead of introducing the magic number 0x015, using `seed` is just fine? I ran a trial, and the compilation succeeded. So, I don't understand why you want to write 0x015 explicitly.
```suggestion
                    "csrrw {0}, seed, x0",
```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+            // Read the seed CSR using inline assembly.
+            // The seed CSR is at address 0x015.
+            // We use csrrw with x0 (zero) to perform a read while signaling polling intent
+            // (per Zkr spec, a write signals that the core is ready for entropy).
+            // We read into usize to match the native XLEN register width (RV64/RV32).
+            let seed_val: usize;
+            // IMPORTANT: The Zkr spec requires using `csrrw` to access seed CSR.
+            // The write operation signals polling and triggers wipe-on-read.
+            // Read-only CSR instructions (csrrs/csrrc with rs1=x0) would trap.
+            // The value written (zero) is ignored by hardware per the spec.
```
lrh2000: There isn't much need for so many comments here. Even worse, it appears to mention the same thing multiple times. I am not sure why.
 - Simply stating what the spec requires is sufficient. Anyone wondering why should just read the spec, which is written in much more detail.

Something like `We read into usize to match the native XLEN register width (RV64/RV32).` provides zero useful information and should be removed.

See also the next comment: https://github.com/asterinas/asterinas/pull/3169#discussion_r3182164760.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+            unsafe {
```
lrh2000, line 140: Every `unsafe` block should have a safety comment.

In combination with the above comments, we should end up with something like:
```rust
            // SAFETY: We've checked that the Zkr extension is available.
            // Therefore, this is a valid instruction to access the `seed` CSR.
            unsafe {
                // The Zkr spec requires using `csrrw` to access `seed` CSR,
                // which signals polling and flushing. See the spec for more
                // information.
                core::arch::asm!(
                    "csrrw {0}, seed, x0",
                    out(reg) seed_val,
                    options(nomem, nostack)
                );
            }
```

Extra comments should be deleted.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+///   - The `seed` CSR at address 0x015 provides up to 16 bits of entropy
+///
+/// The implementation follows a similar retry pattern as the x86 RDRAND version,
+/// adapted for RISC-V's 16-bit entropy output.
 pub fn read_random() -> Option<u64> {
-    // FIXME: Implement a hardware random number generator on RISC-V platforms.
-    None
+    use cpu::extension::{IsaExtensions, has_extensions};
+
+    // Check if the Zkr (Entropy Source) extension is available
```
lrh2000: For the sake of consistency, let's add final periods to the "do something" sentence as well.
```suggestion
    // Check if the Zkr (Entropy Source) extension is available.
```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+    const OPST_WAIT: usize = 0b01; // Status indicating entropy source is busy (waiting)
+    const OPST_BIST: usize = 0b00; // Status indicating entropy source is in self-test
+    const OPST_DEAD: usize = 0b11; // Status indicating entropy source has failed
+    const OPST_MASK: usize = 0b11 << 30;
+    const OPST_SHIFT: usize = 30;
+    const ENTROPY_MASK: usize = 0xFFFF;
+
+    let mut result: u64 = 0;
+
+    // Collect 4 chunks of 16-bit entropy to form a 64-bit value
```
lrh2000: I think the following is a clearer version:
```rust
    // Read 16-bit entropy from the hardware.
    let read_entropy = || -> Option<u16> { todo!() };

    // Collect 4 chunks of 16-bit entropy to form a 64-bit value.
    let result = ((read_entropy()? as u64) << 48)
        | ((read_entropy()? as u64) << 32)
        | ((read_entropy()? as u64) << 16)
        | (read_entropy()? as u64);

    Some(result)

```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+    use cpu::extension::{IsaExtensions, has_extensions};
+
+    // Check if the Zkr (Entropy Source) extension is available
+    if !has_extensions(IsaExtensions::ZKR) {
+        return None;
+    }
+
+    // The seed CSR provides 16 bits of entropy at a time.
+    // We need 4 successful reads to construct a 64-bit value.
+    const RETRY_LIMIT: usize = 10;
```
lrh2000, line 123: Add a blank line below `RETRY_LIMIT`, as the limit is not related to the subsequent constants (which define how to intercept the bits in the `seed` CSR).

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+            unsafe {
+                core::arch::asm!(
+                    // Use `zero` instead of `x0` here for consistency and robustness
+                    "csrrw {0}, 0x015, zero",
+                    out(reg) seed_val,
+                    options(nomem, nostack)
+                );
+            }
+
+            // Extract status bits (bits 31:30)
```
lrh2000:
```suggestion
            // Extract status bits (bits 31:30).
```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,93 @@ pub fn read_tsc() -> u64 {
...
+/// Reference:
+/// - "RISC-V Cryptography Extensions Volume I: Scalar & Entropy Source Instructions"
+///   - <https://docs.riscv.org/reference/isa/extensions/crypto-scalar/_attachments/riscv-crypto-spec-scalar.pdf>
+///   - Section 2.9: Zkr - Entropy Source Extension
+///   - Section 4.1: The `seed` CSR
+///   - The `seed` CSR at address 0x015 provides up to 16 bits of entropy
+///
+/// The implementation follows a similar retry pattern as the x86 RDRAND version,
+/// adapted for RISC-V's 16-bit entropy output.
 pub fn read_random() -> Option<u64> {
```
lrh2000, line 115: This method will not be used for non-x86 architectures.

https://github.com/asterinas/asterinas/blob/8c3a0d362674f1f9235df20470cc62d596fe63a9/kernel/src/util/random.rs#L18-L19

https://github.com/asterinas/asterinas/blob/8c3a0d362674f1f9235df20470cc62d596fe63a9/kernel/src/util/random.rs#L49-L60

So this commit does not produce any effects. Then,
 1. I wonder if you have tested that the method works correctly? How?
 2. The logic in `util/random.rs` may also need to be updated to select the random seed from either the hardware or the device tree. Currently, I am unsure which source should be prioritized.

lrh2000 (commented):
Thank you for implementing the `read_random` method for RISC-V.

Actually, this method is not called anywhere for non-x86 architectures. I think the `kernel/src/util/random.rs` file should be updated to call it as well. See https://github.com/asterinas/asterinas/pull/3169#discussion_r3182283316 below.

ostd/src/arch/riscv/mod.rs
lrh2000: Oh, I saw the Copilot comment. Using `zero` is fine, but the comment should be deleted since it's unclear and causes confusion.

(I think either "zero" or "x0" is fine. There is no clear winner between the two. So there is also no need to write a comment for it.)

The Copilot's claim that `x0` is not portable is just AI slop. Ignore it. Inline assembly is never portable.

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the comment!
1. Yes, it is true that the implementation will not be used. Temporarily, I tested the method with `ktest` locally and it works properly. I may submit a commit further if necessary. In fact, this implementation is a step for solving the problem of SiFive initialization. For now, `make run_kernel OSDK_TARGET_ARCH=riscv64 SCHEME=sifive_u` triggers uncaught panic at the `get_random_seed` for the lack of `/chosen/rng-seed` in the device tree for `sifive_u`. Initially, I tried to implement the `read_random` to provide a random seed. Unfortunately, it seems that the `Zkr` extension is not supported on `SiFive HiFive Unleashed A00` platform. It is supported on `qemu-virtio, riscv64` though so the implementation works. I am submitting this PR for advice as well.
2. As I illustrated above, the `Zkr` extension may not be supported on certain platforms, so I guess the device tree may be prioritized? But again, the device tree and hadware does not work for `sifive_u` for now, so I am not sure either. I may work on the device tree way to support `sifive_u` initialization later, and any suggestion is welcomed.

lrh2000: > 1. In fact, this implementation is a step for solving the problem of SiFive initialization.

Thanks for the detailed explanation. This important context should help readers understand the purpose of this PR if it were included in the description.

> As I illustrated above, the Zkr extension may not be supported on certain platforms, so I guess the device tree may be prioritized? 

This actually doesn't matter much. For example, we can let the hardware generate the seed first. If that is not available, we can fall back to the device tree approach. I guess hardware random generators can produce higher-quality seeds, so maybe we should try it first.

> I may submit a commit further if necessary.

Since it isn't a complex task, I think we can implement it in this PR. My current proposal is to try `read_random` first for all architectures and then try to view the device tree for non-x86 architectures.

> But again, the device tree and hadware does not work for sifive_u for now

I think the kernel should continue to boot even if no random seed is available, unless it involves TDX guests (see the quoted code snippet below).

https://github.com/asterinas/asterinas/blob/3c35e15a91431d98c8d40096925b6578bef9d617/kernel/src/util/random.rs#L30-L33

However, further discussion and study of concrete approaches is needed. My understandings are:
 - At the very least, the kernel should boot. It should not fail to boot due to a lack of randomness.
 - Real randomness is required if userspace tries to read from `/dev/random`. If the kernel lacks randomness, the user's `read` will be blocked until it acquires some.
 - Even if the kernel cannot get real randomness from the hardware, it can gather from the environment. For example, Linux has [add_interrupt_randomness](https://elixir.bootlin.com/linux/v7.0.5/source/drivers/char/random.c#L1103) which collects randomness from addresses of instructions that were interrupted.

However, fully implementing all of these may be challenging. It needs to address some existing problems like our `/dev/random` reads currently never blocks. We may also need to implement mechanisms like `add_interrupt_randomness` outside of OSTD, as we don't want to increase our TCB size (perhaps we should do so within handlers outside OSTD like `register_bottom_half_handler_l2`).

I don't mean that you have to implement all of these mechanisms. Implementing some and deferring others for future work is also okay. But we should try to be self-contained.

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!
And yes, using `seed` should be fine, the reference doc explicitly writes

> The entropy source extension defines the `seed` CSR at address `0x015`.

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the suggestion!

ostd/src/arch/riscv/mod.rs
le-monde-bleu: Thank you for the revision and all the suggestions about proper notations! Should have been more careful about this.

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,74 @@ pub fn read_tsc() -> u64 {
...
+            if status == OPST_ES16 {
+                // Entropy is available: extract 16 bits (bits 15:0).
+                return Some((seed_val & ENTROPY_MASK) as u16);
+            } else if status == OPST_DEAD {
+                // Entropy source failed permanently.
+                return None;
+            } else if status == OPST_WAIT || status == OPST_BIST {
+                // Entropy source busy or running self-test: retry.
+                core::hint::spin_loop();
+            }
```
tatetian, line 163: Is it possible that `status` has values other than `OPST_ES16`, `OPST_DEAD`, `OPST_WAIT`, and `OPST_BIST`? From the code, this seems to be the case. You might to be explicit about this point:

```rust
            if status == OPST_ES16 {
                // Entropy is available: extract 16 bits (bits 15:0).
                return Some((seed_val & ENTROPY_MASK) as u16);
            } else if status == OPST_DEAD {
                // Entropy source failed permanently.
                return None;
            } else if status == OPST_WAIT || status == OPST_BIST {
                // Entropy source busy or running self-test: retry.
                core::hint::spin_loop();
            } else {
                unreachable!("say your reason; probably because of the RISC-V spec?");
            }
```

lrh2000: It cannot happen because the status has two bits, so there are only four possibilities: 00, 01, 10, and 11. 

image


> From the code, this seems to be the case.

I have no idea how you could think that... Perhaps without checking the definition of the mask number?

> You might to be explicit about this point:

Ideally, the compiler would be able to check this in a `match` expression. However, I don't believe the Rust compiler can do so at this time.

tatetian: Yes, I did not read the definitions of these bits. I just skimmed to that piece of logic and it triggered my brain's detector of "code smell". It happens to have 2-bit length. So the code is correct, as I expected. I thought it has more bits and it is the spec that eliminate other possible values. It turns out that the values only has two bits.

With that said, it wouldn't be more explicit by adding:

```rust
} else {
    unreachable!("we checked all four possible values of the 2-bit status");
}
```

ostd/src/arch/riscv/mod.rs
```diff
@@ -103,9 +103,74 @@ pub fn read_tsc() -> u64 {
...
+    }
+
+    const RETRY_LIMIT: usize = 10;
+
+    const OPST_ES16: usize = 0b10; // Status indicating entropy is available
+    const OPST_WAIT: usize = 0b01; // Status indicating entropy source is busy (waiting)
+    const OPST_BIST: usize = 0b00; // Status indicating entropy source is in self-test
+    const OPST_DEAD: usize = 0b11; // Status indicating entropy source has failed
+    const OPST_MASK: usize = 0b11 << 30;
+    const OPST_SHIFT: usize = 30;
```
tatetian: A slightly better version:

```rust
const OPST_MASK: usize = 0b11 << OPST_SHIFT;
```

le-monde-bleu:
I noticed the failure of CI test, which comes up as mentioned in #3209 and seems flaky/intermittent.

tatetian:
> I noticed the failure of CI test, which comes up as mentioned in #3209 and seems flaky/intermittent.

No worry. They are not your fault.

tatetian (approved):

PR #3170, opened 2026-04-22
Add kata containers documentation in book

StevenJiang1110:
This PR adds documentation on how to use Asterinas as the guest kernel for Kata Containers.

The workflow in the guide has been tested in a [separate repository](https://github.com/asterinas/kata-containers/actions/runs/24870586587/workflow).


book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
+# Using Asterinas as a Kata Guest Kernel
+
+This guide explains how to use Asterinas
+as the guest kernel for Kata Containers.
+
+Kata Containers is a VM-based container runtime.
```
tatetian: Make "Kata Containers" a hyperlink to the project repo/website.

book/src/kernel/vm-based-containers/README.md
```diff
@@ -0,0 +1,26 @@
...
+Its framekernel architecture helps reduce the trusted computing base
+of the guest kernel.
+
+At the same time,
+Asterinas provides a Linux-compatible ABI.
+This allows many existing Linux workloads
+to migrate to Asterinas-based VM environments seamlessly,
+without requiring changes to the applications themselves.
+
+## Available Guides
```
tatetian: **Section heading is generic and the list lacks framing** (minor): "Available Guides" describes the *form* (a list of guides) rather than the *subject* (what kinds of runtimes Asterinas works with). Renaming it to "Supported Runtimes" frames the chapter as a compatibility statement — readers scanning the page learn what Asterinas integrates with, not just that more pages exist. A one-line intro before the list also signals that entries here have been verified, not just collected.

```diff
-## Available Guides
+## Supported Runtimes
+
+Here is the list of secure container runtimes that have been verified to work with Asterinas:

 - [Kata Containers](kata.md)
```

book/src/kernel/vm-based-containers/kata.md
tatetian, line 1: **Heavy duplication between `End Users` and `Kernel Developers`** (significant): The two sections share roughly 80% of their commands — `docker run` with the same device/tmpfs flags (lines 22-32 vs 92-103), `kata_services.sh start`, and `nerdctl run` (lines 51-59 vs 127-133) all appear twice. The genuine differences are: (a) image choice (`asterinas/kata` vs `asterinas/asterinas`), (b) the kernel-developer flow needs `-v` mounts and `kata_env.sh install`, (c) the optional locally-built-kernel section. Reading top-to-bottom, the reader has to compare two near-identical command blocks to find the diff.

Restructure as one linear walkthrough that follows the "Prepare → Build → Run → Debug" shape established by `kernel/intel-tdx.md`:

1. **Prepare the host** (KVM, vhost modules, arch).
2. **Run the dev container** — one `docker run` recipe, with a tabbed/branched note: "Operators use `asterinas/kata`; contributors who want to swap in a locally built kernel use `asterinas/asterinas` plus the bind-mount flags shown below."
3. **Start a Kata workload** (`kata_services.sh start`, `nerdctl run`, verification).
4. **Optional: Build and use a local kernel** — only this subsection has `kata_env.sh install`, the `configuration.toml` edit, and `make kernel BOOT_METHOD=qemu-direct`.

This roughly halves the page and makes the two audiences read the same first three sections.

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
...
+docker run --rm -it \
+    --cgroupns host \
+    --privileged \
+    --device /dev/kvm \
+    --device /dev/vhost-net \
+    --device /dev/vhost-vsock \
+    --tmpfs /tmp:exec,mode=1777,size=8g \
+    --tmpfs /var/lib/containerd:exec,mode=755,size=8g \
+    asterinas/kata:0.17.2-20260407
+```
```
tatetian, line 95: **No host-prerequisite section** (significant): The guide jumps straight into `docker run --device /dev/kvm --device /dev/vhost-net --device /dev/vhost-vsock` without stating what the host must provide. A reader on a non-x86_64 host, on a kernel without `vhost_net`/`vhost_vsock` modules loaded, without permissions on `/dev/kvm`, or running the guide on a CI box without nested virtualization will get a cryptic Docker error rather than a clear diagnosis. `kernel/intel-tdx.md` has a dedicated "Prepare the Intel TDX Environment" section; this guide should mirror it.

Add a section before the first `docker run`:

````markdown
## Prepare the host

Kata Containers requires a host with KVM and vhost support.
The current guide is x86_64-only.

Verify that the required device nodes exist:

```bash
ls /dev/kvm /dev/vhost-net /dev/vhost-vsock
```

If any are missing, load the matching kernel modules
(`kvm`, `kvm_intel`/`kvm_amd`, `vhost_net`, `vhost_vsock`)
and ensure your user is in the `kvm` group, or run the
container as root.
````

File: book/src/kernel/vm-based-containers/kata.md [hunk 4]
```diff
@@ -0,0 +1,152 @@
...
+## End Users
+
+For end users,
+Asterinas provides the `asterinas/kata` container image.
+This image includes the dependencies and scripts required by Kata.
+
+Run it with:
+
+```bash
+docker run --rm -it \
```
tatetian: (Also applies to the equivalent block in the kernel-developer flow at lines 92-103.)

**`docker run` flags are cryptic and `--rm` shouldn't be prescribed** (significant): The `docker run` recipe stacks several non-obvious flags with no explanation, and `--rm` in particular shouldn't be there at all — it's a lifecycle choice that belongs to the user. A reader who hits an error mid-walkthrough loses the container (and any state in `/tmp`, `/var/lib/containerd`, etc.) the moment they `exit`, which makes troubleshooting harder and rules out re-entering the same instance to retry a failed step. The other flags (`--cgroupns host`, `--privileged`, the two `--tmpfs` overrides) are likely required, but the doc should say so once.

Two changes:

1. Drop `--rm` from both `docker run` blocks. Let the user decide whether to clean up; mention `docker rm <id>` (or `docker run --rm` as an opt-in) at the end of the section.

2. Add a one-line note per non-obvious flag, e.g.:

```markdown
The flags in the command above:

- `--cgroupns host` shares the host's cgroup namespace so that containerd
  inside the container can manage Kata workloads correctly.
- `--privileged` is required for KVM and nested container management.
- `--device /dev/{kvm,vhost-net,vhost-vsock}` exposes the virtualization
  devices Kata needs.
- `--tmpfs /tmp:exec,...` and `--tmpfs /var/lib/containerd:exec,...`
  give containerd writable, exec-permitted scratch space; without `exec`,
  Kata's runtime binaries fail to launch.
```

Tighten the wording per flag once the actual rationale is confirmed (the bullets above are the most plausible reading; the author can correct any that are wrong).

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
...
+these services are `containerd` and `syslogd`:
+
+```bash
+tools/kata/kata_services.sh start
+```
+
+Check their status to make sure they started successfully:
+
+```bash
+tools/kata/kata_services.sh status
```
tatetian, line 160: **Provenance of `tools/kata/*.sh` is implicit** (significant): No `tools/kata/` directory exists in the asterinas tree (`find tools/ -name 'kata_*.sh'` returns nothing), so a reader who lands on this guide via the asterinas repo will be confused. In the End Users flow these scripts presumably ship inside the `asterinas/kata` image; in the Kernel Developers flow they resolve relative to `/root/kata-containers` (the `-w` working directory set by `docker run`), i.e. they live in the `asterinas/kata-containers` fork. Both are reasonable, but the doc never says so.

Add one sentence per flow that names where the scripts come from, so readers know where to look when something fails:

```diff
 For end users,
 Asterinas provides the `asterinas/kata` container image.
-This image includes the dependencies and scripts required by Kata.
+This image bundles the dependencies and helper scripts required by Kata.
+The `tools/kata/` scripts referenced below are pre-installed in the
+image's working directory.
```

```diff
 After entering the container,
 you will already be in `/root/kata-containers`.
+The `tools/kata/` scripts referenced below come from this checkout
+(the Asterinas fork of Kata Containers), not the asterinas tree.
```

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
...
+
+```bash
+nerdctl run \
+    --cgroup-manager cgroupfs \
+    --net none \
+    --runtime io.containerd.kata.v2 \
+    --name foo \
+    -it \
+    docker.io/alpine:latest
+```
```
tatetian, line 175: **Unexplained workflow constraints** (significant): `--net none` and `--cgroup-manager cgroupfs` are non-obvious choices that a reader cannot guess from context. `--net none` is the surprising one: a user who tries `apk add` or `curl` inside the resulting Alpine container will get cryptic failures and assume Kata is broken — they have no way to know networking was deliberately omitted from the demo. `--cgroup-manager cgroupfs` is presumably required because the host's cgroups are managed by systemd while we're running nerdctl inside a Docker container that doesn't see systemd; again, the reader cannot infer that. Compounding this, the kernel-developer flow (line 127) repeats both flags and never calls `kata_services.sh status`, so a developer hitting a startup failure gets no early signal.

Add a short paragraph just after the code block:

```markdown
The flags above are not optional:

- `--cgroup-manager cgroupfs` is required because we are nested inside a
  Docker container where the systemd cgroup driver is not available.
- `--net none` keeps this example minimal — the demo only checks that the
  workload boots inside an Asterinas guest. Adding networking is out of
  scope for this guide.
```

Also add `tools/kata/kata_services.sh status` to the kernel-developer flow before `nerdctl run`, for parity with the end-user flow.

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
...
+
+After the container starts,
+you are inside Alpine.
+Run the following commands to confirm that the workload
+is no longer running in the host environment:
+
+```bash
+cat /proc/version
+cat /etc/alpine-release
+```
```
tatetian: **No "what success looks like"** (significant): Confirming that the workload runs on Asterinas is the entire point of this guide, but the verification step says only "run these commands" without telling the reader what the output should contain. A reader cannot distinguish a successful Asterinas boot from accidentally falling back to a Linux guest kernel. A single load-bearing substring is enough.

```diff
 Run the following commands to confirm that the workload
-is no longer running in the host environment:
+is running inside an Asterinas guest, not the host kernel:
 
 ```bash
 cat /proc/version
 cat /etc/alpine-release
 ```
+
+The `/proc/version` output should mention `Asterinas` rather than the host's
+Linux version, e.g.:
+
+```text
+Asterinas version 0.17.2 (...)
+```
+
+`/etc/alpine-release` should print the Alpine version of the rootfs in the
+container, confirming you are inside the Alpine workload (and not, say,
+still inside the asterinas dev container).
```

Replace the version string above with the actual format Asterinas prints — confirm it once locally before merging.

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
...
+
+You can also point Kata to a locally built guest kernel.
+In `/etc/kata-containers/configuration.toml`,
+set `kernel` to the path of the local kernel image:
+
+```toml
+[hypervisor.qemu]
+kernel = "/root/asterinas/target/osdk/aster-kernel-osdk-bin.qemu_elf"
+```
```
tatetian: **Configuration-file provenance and ordering are unstated** (significant): The reader is told to edit `/etc/kata-containers/configuration.toml` but never told how it gets there (presumably written by `kata_env.sh install`). Three loose ends a reader will hit:

1. **Order:** the section sits *after* the basic `nerdctl run` walkthrough, but the edit must happen *after* `kata_env.sh install` (so the file exists) and *before* re-running `nerdctl run`. State that explicitly.
2. **Idempotence:** if `kata_env.sh install` clobbers user edits when re-run, mention it. Ideally the script should accept `--kernel <path>` and avoid the hand-edit.
3. **Kernel binary filename:** confirm `aster-kernel-osdk-bin.qemu_elf` is what `make kernel BOOT_METHOD=qemu-direct` actually produces. The OSDK code at `osdk/src/commands/build/bin.rs:83-92` constructs `<input-elf-name>.qemu_elf`; the input ELF in this repo is installed as `aster-kernel-osdk-bin` (visible in `target/osdk/iso_root/boot/`), which matches — but I could not verify by running a `qemu-direct` build, so the author should run it once to confirm rather than relying on the static analysis.

Suggested rewrite:

```markdown
You can also point Kata to a locally built guest kernel. After running
`kata_env.sh install` (which writes `/etc/kata-containers/configuration.toml`),
edit the file and set `kernel` under `[hypervisor.qemu]`:

```toml
[hypervisor.qemu]
kernel = "/root/asterinas/target/osdk/aster-kernel-osdk-bin.qemu_elf"
```

Note: re-running `kata_env.sh install` will overwrite this edit (TODO: confirm
whether the script preserves user edits, or add a `--kernel <path>` flag).

Build the kernel with:
…
```

File: book/src/kernel/vm-based-containers/kata.md [hunk 9]
```diff
@@ -0,0 +1,152 @@
...
+Kernel developers usually use the `asterinas/asterinas` container image.
+When starting the container,
+also pass the additional arguments required by Kata:
+
+```bash
+# Assumes the Asterinas source has already been cloned locally.
+ASTERINAS_SRC=$HOME/asterinas
+
+# Clones the kata-containers repository locally as well.
+cd $HOME && git clone https://github.com/asterinas/kata-containers.git
```
tatetian, line 116: **Forked repo cloned without rationale** (minor): The kernel-developer flow tells the reader to clone an Asterinas fork of Kata Containers and bind-mount it, but never says why a fork is needed, what diverges from upstream, or whether contributions should go upstream or to the fork. A contributor evaluating Kata support naturally asks both questions.

```diff
+# Clones the Asterinas fork of kata-containers locally as well.
+# We use a fork because it carries patches that let the runtime boot an
+# Asterinas guest kernel; these are being upstreamed.
 cd $HOME && git clone https://github.com/asterinas/kata-containers.git
```

Replace the second sentence with the actual reason the fork exists (e.g., specific upstream PRs in flight, or a note that the fork tracks upstream main plus a small patch series).

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,152 @@
...
+
+After entering the container,
+you will already be in `/root/kata-containers`.
+The examples below assume that the current `kata-containers` checkout
+is mounted there,
+and that the local Asterinas tree is mounted at `/root/asterinas`.
+
+```bash
+pwd
+```
```
tatetian, line 242: **Stray `pwd` block** (minor): The previous paragraph already states the working directory; the bare `pwd` block has no surrounding instruction and shows no expected output. As written it reads like leftover scaffolding. Either delete it, or turn it into an explicit verification:

````markdown
Verify the working directory:

```console
$ pwd
/root/kata-containers
```
````

Deleting is cleaner — the prose already communicates the cwd.

File: book/src/kernel/vm-based-containers/kata.md [hunk 11]
```diff
@@ -0,0 +1,152 @@
...
+
+Kata Containers is a VM-based container runtime.
+It runs each container inside its own virtual machine,
+so containers do not share the host kernel.
+This design combines the deployment model of containers
+with the stronger isolation boundary of virtual machines.
+It reduces the risk that a compromised workload can affect
+the host or other containers.
+
+## End Users
```
tatetian: (Also applies to `## Kernel Developers` at line 78.)

**H2 capitalization is inconsistent with neighbors** (minor): The closest analogue mixes sentence and title case (`Why choose Asterinas for Intel TDX` and `Build and run Asterinas` are sentence case, but `Prepare the Intel TDX Environment` and `Using GDB to Debug` are Title Case — `intel-tdx.md` is itself inconsistent). This file goes uniformly Title Case. Pick one style and apply chapter-wide so the book's H2s converge. If the section split survives the duplication-cleanup comment above:

```diff
-## End Users
+## For end users
…
-## Kernel Developers
+## For kernel developers
```

(If the file is restructured per the duplication comment, this issue dissolves into the new section names.)

tatetian (commented):
The documentation needs more work — please address the comments below and do another pass before the next push. (Most of these were generated by my coding agents via the `pr-review` skill, with a few dropped; you can run the skill yourself to catch this kind of thing earlier.

book/src/kernel/vm-based-containers/kata.md
StevenJiang1110: H1 headers now use title case, while H2 and H3 headers use sentence case.

StevenJiang1110:
> The documentation needs more work — please address the comments below and do another pass before the next push.

I have run the PR review checks three times and addressed all identified issues locally. The latest review shows no remaining issues, so I have marked the PR as ready again.

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,272 @@
...
+It reduces the risk that a compromised workload can affect
+the host or other containers.
+
+This guide uses the
+[Asterinas fork of Kata Containers](https://github.com/asterinas/kata-containers),
+which carries Asterinas-specific patches, helper scripts,
+and configuration for building, installing, and testing Kata
+with Asterinas as the guest kernel.
+
+## Prepare the host
```
tatetian: Rename this title to "Prepare the host kernel" to be more precise. The next section, "Enter the Kata environment", is also meant to prepare the "host"——if everything outside the Kata container is considered the "host".

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,272 @@
...
+```bash
+sudo modprobe kvm
+sudo modprobe kvm_intel  # Or use kvm_amd on AMD hosts.
+sudo modprobe vhost_vsock
+```
+
+Then make sure the user running Docker can access
+`/dev/kvm` and `/dev/vhost-vsock`.
+
+## Enter the Kata environment
```
tatetian: "Kata environment" is a bit of confusing. People may mistake it for the environment inside the Kata container.

How about this way?

```md
## Enter the Kata-ready environment

We provide two Docker images with Asterinas and/or Kata preinstalled
for end users and kernel developers, respectively.
End users should use the prebuilt `asterinas/kata` image.
Kernel developers should use the `asterinas/asterinas` image
with source mounts.
```

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,272 @@
...
+Choose one of the two workflows below.
+End users should use the prebuilt `asterinas/kata` image.
+Kernel developers should use the `asterinas/asterinas` image
+with source mounts.
+
+### Shared Docker arguments
+
+All workflows need the same Docker access to KVM, vhost, cgroups,
+and temporary storage.
+Define the shared Docker arguments first:
```
tatetian: End users and kernel developers share some steps. So it is unclear which "two workflows" refer to. I suggest avoiding using the term "workflows". Another reason that you don't need to mention "two workflows" explicitly is that the different part between "two workflows" is put under "For end users" and "For kernel developers".

```md
### Prepare Docker arguments

Define the following Docker argument:

```bash
KATA_DOCKER_ARGS=(
    --cgroupns host
    --privileged
    --device /dev/kvm
    --device /dev/vhost-vsock
    --tmpfs /tmp:exec,mode=1777,size=8g
    --tmpfs /var/lib/containerd:exec,mode=755,size=8g
)
```

The Docker arguments apply to both end users and kernel devleopers.

book/src/kernel/vm-based-containers/kata.md
```diff
@@ -0,0 +1,272 @@
...
+- `--tmpfs /tmp:exec,...` and `--tmpfs /var/lib/containerd:exec,...`
+  give Kata enough temporary storage space to create containers.
+
+### For end users
+
+Use the `asterinas/kata` image.
+It already includes Kata, the Asterinas guest kernel,
+the Kata configuration,
+and the `tools/kata/` helper scripts in `/root/kata-containers`.
+The command below makes `/root/kata-containers` the working directory:
```
tatetian: This page is titled "Using Asterinas as a Kata Guest Kernel". The entire point of this page is to provide concrete instructions so that people can try start Asterinas-powered Kata containers. So to maximize readability, put instructions first and explanations later. Do not provide explanations upfront. This forces readers to comprehend the explanations.

For example, take a look at the root README of Asterinas:

Image

Explanation or supplementary info (e.g., "Alternatively, if you use VS Code..") is given _after_ the instruction. People can skip it if they want.

Back to our documentation here. Do not start by describing the image:

```md
Use the `asterinas/kata` image.
It already includes Kata, the Asterinas guest kernel,
the Kata configuration,
and the `tools/kata/` helper scripts in `/root/kata-containers`.
The command below makes `/root/kata-containers` the working directory:
```

Start with the action:

```md
### For end users

Run a `asterinas/kata` Docker container 
to enter an environment with Kata and Asterinas pre-installed:

```bash
docker run -it \
    "${KATA_DOCKER_ARGS[@]}" \
    -w /root/kata-containers \
    asterinas/kata:0.17.2-20260407
```

The Docker image includes Kata, the Asterinas guest kernel,
the Kata configuration,
and the `tools/kata/` helper scripts in `/root/kata-containers`.
The command below makes `/root/kata-containers` the working directory:
```

The same principle applies to "For kernel developers" and other action-oriented sections.

---

In a similar spirit, I suggest numbering all section to make steps more clear:

```md
## Step 1: Prepare the host

## Step 2: Enter the Kata environment

### Step 2.1: Prepare Docker arguments

### Step 2.2: Start the Docker container

#### For end users

#### For kernel users

## Step 3: Start a Kata workload

### Step 3.1: Start services

### Step 3.2: Run Alpine with Kata

### Step 3.3: Verify the guest

### Step 3.4: Clean up

## Step 4: Use a local kernel (optional)
```

It would be much easier for readers to follow the instructions by telling them which step they are currently at and how many steps in total.

tatetian (commented):
Thanks for the revision! This version looks much better.

tatetian (approved):

PR #3174, opened 2026-04-24
Fix RISC-V regression test portability issues

TankTechnology:
## Summary
- Avoid failing the hello_world regression on non-x86 when the optional x86_64 assembly test is skipped.
- Relax mmap_err low-address mmap expectations so architectures whose image already occupies mmap_min_addr can still pass with a valid higher mapping.
- Skip the x86_64-specific execve_err regression on non-x86 architectures until that test is ported.

## Test Plan
- `git diff --check`
- `CARGO_TARGET_DIR=/tmp/asterinas-target make run_kernel OSDK_TARGET_ARCH=riscv64 AUTO_TEST=regression INITRAMFS_SKIP_GZIP=1`

The RISC-V regression run now gets past hello_world, memory/mmap_err, and process/execve_err. It later fails in the network tests at interface enumeration (`test_if_nameindex` and `test_get_link_by_libnl` expect both lo and eth0). I kept that as a follow-up issue rather than broadening this PR.

`make -C test/initramfs/src/regression check` is currently blocked by an unrelated pre-existing clang-format issue in `fs/pseudofs/pseudo_file_create.h`.

### DIFF
--- a/test/initramfs/src/regression/hello_world/run_test.sh
+++ b/test/initramfs/src/regression/hello_world/run_test.sh
@@ -6,6 +6,8 @@ set -e
 
 ./in_c/hello
 ./in_c_pie/hello
-if [ "$(uname -m)" = "x86_64" ]; then
-    ./in_assembly/hello
-fi
+case "$(uname -m)" in
+    x86_64|riscv64)
+        ./in_assembly/hello
+        ;;
+esac
--- a/test/initramfs/src/regression/hello_world/run_test.sh
+++ b/test/initramfs/src/regression/hello_world/run_test.sh
@@ -6,4 +6,6 @@ set -e
 
 ./in_c/hello
 ./in_c_pie/hello
-[ "$(uname -m)" = "x86_64" ] && ./in_assembly/hello
+if [ "$(uname -m)" = "x86_64" ]; then
+    ./in_assembly/hello
+fi
--- a/test/initramfs/src/regression/memory/mmap/mmap_err.c
+++ b/test/initramfs/src/regression/memory/mmap/mmap_err.c
@@ -3,7 +3,6 @@
 #define _GNU_SOURCE
 #include <sys/mman.h>
 #include <sys/fcntl.h>
-#include <stdint.h>
 #include <unistd.h>
 
 #include "../../common/test.h"
@@ -12,11 +11,6 @@
 
 // The value in `/proc/sys/vm/mmap_min_addr`.
 #define MMAP_MIN_ADDR ((void *)65536)
-
-static int is_valid_low_hint_result(void *addr)
-{
-	return (uintptr_t)addr >= (uintptr_t)MMAP_MIN_ADDR;
-}
 
 static void *valid_addr;
 static void *avail_addr;
@@ -121,22 +115,16 @@ FN_TEST(underflow_addr)
 	void *addr = (void *)PAGE_SIZE;
 	void *addr2;
 
-	// `mmap` without MAP_FIXED treats `addr` as a hint. Linux first
-	// rounds the hint down to a page boundary. If the rounded hint is
-	// non-null and below `mmap_min_addr`, Linux raises it to
-	// `mmap_min_addr`. The adjusted address is still only a hint: it is
-	// returned only if that range is free. The current RISC-V regression
-	// binary is an ET_EXEC whose first load segment starts at `0x10000`,
-	// which equals `mmap_min_addr`, so the rounded hint can already be
-	// occupied. The portable expectation is therefore any address that
-	// respects the minimum mapping address.
+	// `mmap` without MAP_FIXED. The hint address will be rounded
+	// to MMAP_MIN_ADDR, unless it is in the first page, in which
+	// case the hint will be ignored.
 	addr2 = TEST_RES(mmap(addr, PAGE_SIZE, PROT_READ,
 			      MAP_PRIVATE | MAP_ANONYMOUS, 0, 0),
-			 is_valid_low_hint_result(_ret));
+			 _ret == MMAP_MIN_ADDR);
 	TEST_SUCC(munmap(addr2, PAGE_SIZE));
 	addr2 = TEST_RES(mmap(addr + 1, PAGE_SIZE, PROT_READ,
 			      MAP_PRIVATE | MAP_ANONYMOUS, 0, 0),
-			 is_valid_low_hint_result(_ret));
+			 _ret == MMAP_MIN_ADDR);
 	TEST_SUCC(munmap(addr2, PAGE_SIZE));
 	addr2 = TEST_RES(mmap(addr - 1, PAGE_SIZE, PROT_READ,
 			      MAP_PRIVATE | MAP_ANONYMOUS, 0, 0),
--- a/test/initramfs/src/regression/memory/mmap/mmap_err.c
+++ b/test/initramfs/src/regression/memory/mmap/mmap_err.c
@@ -13,8 +13,10 @@
 // The value in `/proc/sys/vm/mmap_min_addr`.
 #define MMAP_MIN_ADDR ((void *)65536)
 
-#define IS_ABOVE_MMAP_MIN_ADDR(addr) \
-	((uintptr_t)(addr) >= (uintptr_t)MMAP_MIN_ADDR)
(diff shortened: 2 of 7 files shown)


test/initramfs/src/regression/hello_world/run_test.sh
```diff
@@ -6,4 +6,6 @@ set -e
 
 ./in_c/hello
 ./in_c_pie/hello
-[ "$(uname -m)" = "x86_64" ] && ./in_assembly/hello
+if [ "$(uname -m)" = "x86_64" ]; then
+    ./in_assembly/hello
```
tatetian: How about adding a RISC-V version of the `hello.S` file and cross-compiling it?

test/initramfs/src/regression/memory/mmap/mmap_err.c
```diff
@@ -115,16 +119,17 @@ FN_TEST(underflow_addr)
 	void *addr = (void *)PAGE_SIZE;
 	void *addr2;
 
-	// `mmap` without MAP_FIXED. The hint address will be rounded
-	// to MMAP_MIN_ADDR, unless it is in the first page, in which
-	// case the hint will be ignored.
+	// `mmap` without MAP_FIXED. The hint address will be rounded to
+	// MMAP_MIN_ADDR unless it is in the first page, in which case the
+	// hint will be ignored. If MMAP_MIN_ADDR is already occupied, Linux
+	// may choose another available address.
```
tatetian: What is the underlying reason to make this change? What makes the kernel behavior on RISC-V different?

test/initramfs/src/regression/process/run_test.sh
```diff
@@ -16,7 +16,9 @@ set -e
 ./cpu_affinity/cpu_affinity
 
 ./execve/execve
-./execve/execve_err
+if [ "$(uname -m)" = "x86_64" ]; then
+    ./execve/execve_err
```
tatetian: Again, I would suggest fixing the source code so that it can be compiled for risc-v.

tatetian (commented):
Good job on identifying these problems and fixing them.

test/initramfs/src/regression/process/execve/execve_err.c
```diff
@@ -288,9 +297,40 @@ static int do_execve_fatal(void)
...
 #include <sys/syscall.h>
 
+#if defined(__x86_64__)
+#include <sys/user.h>
+#elif defined(__riscv) && __riscv_xlen == 64
+#include <asm/ptrace.h>
+#include <sys/uio.h>
+#endif
+
+static void get_execve_regs(pid_t pid, struct user_regs_struct *regs)
```
tatetian: Both `get_execve_regs` and `execve_result_errno` are misnomers. They are not specific to exceve and thus having "exceve" in their names are simply misleading. The former function can get the register of _any_ traced process/thread, whereas the latter can get the returned errno by _any_ syscall.

Fix:
* `get_execve_regs` --> `get_tracee_regs`
* `execve_result_errno` --> `get_syscall_errno`

test/initramfs/src/regression/memory/mmap/mmap_err.c
```diff
@@ -115,16 +121,22 @@ FN_TEST(underflow_addr)
...
+	// `mmap_min_addr`. The adjusted address is still only a hint: it is
+	// returned only if that range is free. The current RISC-V regression
+	// binary is an ET_EXEC whose first load segment starts at `0x10000`,
+	// which equals `mmap_min_addr`, so the rounded hint can already be
+	// occupied. The portable expectation is therefore any address that
+	// respects the minimum mapping address.
 	addr2 = TEST_RES(mmap(addr, PAGE_SIZE, PROT_READ,
 			      MAP_PRIVATE | MAP_ANONYMOUS, 0, 0),
-			 _ret == MMAP_MIN_ADDR);
+			 is_valid_low_hint_result(_ret));
```
tatetian: The test is weaken for x86 and too many words are spent on a nuance. A simpler fix is to link the test as PIE in `test/initramfs/src/regression/memory/mmap/Makefile` so the binary is loaded above `mmap_min_addr` on every arch and the original strict check works unchanged.

test/initramfs/src/regression/process/execve/execve_err.c
```diff
@@ -288,9 +297,40 @@ static int do_execve_fatal(void)
...
+{
+#if defined(__x86_64__)
+	CHECK_WITH(ptrace(PTRACE_GETREGS, pid, NULL, regs),
+		   _ret >= 0 && regs->orig_rax == __NR_execve);
```
tatetian: The x86 branch has the check `regs->orig_rax == __NR_execve`, but the risc-v one lacks the equivalent check. Since I already suggested renaming this function, you should ask the caller of this function to do the check.

test/initramfs/src/regression/hello_world/run_test.sh
TankTechnology: Done. Added a RISC-V 64 assembly version in `hello.S` using `#if defined(__riscv)` with `ecall`-based syscalls (write=64 via a7, exit=93 via a7). The Makefile now also builds `in_assembly` for `riscv64-linux`, and `run_test.sh` runs it on riscv64. Verified on RISC-V QEMU: the assembly test prints "Hello, world" x3 correctly.

test/initramfs/src/regression/memory/mmap/mmap_err.c
TankTechnology: The root cause: on RISC-V the test binary (ET_EXEC) has its first load segment at `0x10000`, which equals `mmap_min_addr`. So the mmap hint address got rounded up to `mmap_min_addr` but the range was already occupied, causing the test to fail. 

Following your later suggestion about PIE (comment below), I linked the test as PIE so the binary loads above `mmap_min_addr` on all architectures. This eliminates the address conflict and allows the original strict `_ret == MMAP_MIN_ADDR` assertion to work unchanged on both x86 and RISC-V.

test/initramfs/src/regression/process/run_test.sh
TankTechnology: Done. Added RISC-V support: `ELF_MACHINE` macro to select `EM_X86_64`/`EM_RISCV`, architecture-specific `illegal_instr[]` (ud2 for x86, zero-filled instruction for RISC-V), and `run_test.sh` now unconditionally runs `execve_err` on all architectures. All execve error tests pass on RISC-V QEMU.

test/initramfs/src/regression/process/execve/execve_err.c
TankTechnology: Good catch. Renamed as suggested:
- `get_execve_regs` → `get_tracee_regs` (just reads tracee registers via ptrace, no policy)
- `execve_result_errno` → `get_syscall_errno` (returns the errno from syscall return value)

Also extracted `get_syscall_number()` as a separate accessor so the `__NR_execve` check can be done by the caller (see reply to your next comment).

test/initramfs/src/regression/memory/mmap/mmap_err.c
TankTechnology: Agreed, PIE is the cleaner fix. Changed `Makefile` to add `EXTRA_C_FLAGS := -fPIE -pie`, and reverted the mmap_err.c assertion back to the strict `_ret == MMAP_MIN_ADDR` (removed the `is_valid_low_hint_result` helper and the verbose comment). With PIE the binary loads well above `mmap_min_addr` on every arch, so the original strict check works correctly on both x86 and RISC-V.

test/initramfs/src/regression/process/execve/execve_err.c
TankTechnology: Good point. I moved the syscall number check out of `get_tracee_regs` and into the caller `do_execve_fatal` via a separate `get_syscall_number()` helper:

```c
get_tracee_regs(pid, &regs);
CHECK_WITH(get_syscall_number(&regs), _ret == __NR_execve);
```

`get_syscall_number()` returns `regs->orig_rax` on x86 and `regs->a7` on RISC-V, so the check is now architecture-agnostic and covers both equally.

tatetian (approved):

PR #3175, opened 2026-04-24
Support dummy `cpu.weight` and `cpu.max` attributes for cgroup cpu sub-controller

cchanging:
Required by #2939 .

### DIFF
--- a/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
@@ -1,32 +1,64 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use alloc::sync::Arc;
-use core::sync::atomic::{AtomicBool, Ordering};
+use core::sync::atomic::{AtomicU32, Ordering};
 
-use aster_systree::{Error, Result, SysAttrSetBuilder, SysPerms, SysStr};
+use aster_systree::{Error, MAX_ATTR_SIZE, Result, SysAttrSetBuilder, SysPerms, SysStr};
 use aster_util::{per_cpu_counter::PerCpuCounter, printer::VmPrinter};
 use ostd::{
     cpu::CpuId,
     mm::{VmReader, VmWriter},
+    sync::SpinLock,
     task::atomic_mode::AsAtomicModeGuard,
     timer::Jiffies,
+    warn,
 };
 
 use crate::{
     fs::cgroupfs::systree_node::{CgroupSysNode, CgroupSystem},
     process::Process,
+    util::ReadCString,
 };
 
 /// A sub-controller responsible for CPU resource management in the cgroup subsystem.
 ///
 /// This controller always exists so that `cpu.stat` remains readable even before `+cpu`
-/// is enabled. When inactive, only the base usage fields are exposed. The `user` and
-/// `system` counters store CPU time in units of one [`Jiffies`] per increment, and
-/// `usage_usec` is derived from their sum when `cpu.stat` is read.
+/// is enabled. When inactive, only the base usage fields are exposed and the non-root
+/// CPU control files are hidden.
 pub struct CpuController {
-    is_enabled: AtomicBool,
+    /// Persistent CPU usage accounting for this cgroup.
+    stats: CpuStats,
+    /// Optional CPU resource control attributes exposed only when `+cpu` is active.
+    control: Option<CpuControl>,
+}
+
+/// CPU usage accounting that survives CPU controller toggles.
+struct CpuStats {
+    /// A counter accumulates CPU time spent executing user-space code.
+    ///
+    /// The counter is in units of one [`Jiffies`] per increment.
     user: PerCpuCounter,
+    /// A counter accumulates CPU time spent executing kernel code on behalf of tasks.
+    ///
+    /// The counter is in units of one [`Jiffies`] per increment.
     system: PerCpuCounter,
+}
+
+/// CPU resource controls that are reset whenever `+cpu` is re-enabled.
+struct CpuControl {
+    /// A value stores the configured relative CPU share for scheduler integration.
+    weight: AtomicU32,
+    /// A value stores the configured CPU bandwidth limit for scheduler integration.
+    max: SpinLock<CpuMax>,
+}
+
+/// A CPU bandwidth limit.
+#[derive(Clone, Copy)]
+struct CpuMax {
+    /// The configured CPU runtime budget for bandwidth enforcement.
+    quota_usec: u64,
+    /// The window over which the CPU runtime budget is intended to apply.
+    period_usec: u64,
 }
 
 /// Specifies the cgroup CPU sub-controller receives one [`Jiffies`] of charge.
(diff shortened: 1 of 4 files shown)


kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -1,32 +1,56 @@
...
+    ///
+    /// The counter is in units of one [`Jiffies`] per increment.
     user: PerCpuCounter,
+    /// A counter accumulates CPU time spent executing kernel code on behalf of tasks.
+    ///
+    /// The counter is in units of one [`Jiffies`] per increment.
     system: PerCpuCounter,
+    /// A value stores the configured relative CPU share for future scheduler integration.
+    weight: AtomicU32,
+    /// A value stores the configured CPU bandwidth limit for future scheduler integration.
```
lrh2000: There's no need to say "future" in the documentation comments. We'll have to remove it when the functionality is implemented, which does not seem necessary.

```diff
diff --git a/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs b/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
index aba7972d6..108a8e1ca 100644
--- a/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
@@ -25,8 +25,6 @@ use crate::{
 /// This controller always exists so that `cpu.stat` remains readable even before `+cpu`
 /// is enabled. When inactive, only the base usage fields are exposed and the non-root
 /// CPU control files are hidden.
-//
-// The CPU control knobs are currently stored but not enforced.
 pub struct CpuController {
     /// Tracks whether the optional CPU control interface is visible for this cgroup.
     is_enabled: AtomicBool,
@@ -38,16 +36,16 @@ pub struct CpuController {
     ///
     /// The counter is in units of one [`Jiffies`] per increment.
     system: PerCpuCounter,
-    /// A value stores the configured relative CPU share for future scheduler integration.
+    /// A value stores the configured relative CPU share for scheduler integration.
     weight: AtomicU32,
-    /// A value stores the configured CPU bandwidth limit for future scheduler integration.
+    /// A value stores the configured CPU bandwidth limit for scheduler integration.
     max: SpinLock<CpuMax>,
 }
 
 /// A CPU bandwidth limit.
 #[derive(Clone, Copy)]
 struct CpuMax {
-    /// The configured CPU runtime budget for future bandwidth enforcement.
+    /// The configured CPU runtime budget for bandwidth enforcement.
     quota_usec: u64,
     /// The window over which the CPU runtime budget is intended to apply.
     period_usec: u64,
```

But you can add some TODOs around the warnings.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -1,32 +1,56 @@
...
+    max: SpinLock<CpuMax>,
```
lrh2000: `weight` and `max` need to be reset after disabling and re-enable the controller.

Should we replace the controller object the same way we do with other types of controllers?

Otherwise, there are quite some race conditions to handle if we reset the properties manually.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -102,16 +142,125 @@ impl CpuController {
...
+                    .map_err(|_| Error::InvalidOperation)?
+                    .trim()
+                    .parse::<u32>()
+                    .map_err(|_| Error::InvalidOperation)?;
+                if !(MIN_WEIGHT..=MAX_WEIGHT).contains(&weight) {
+                    return Err(Error::InvalidOperation);
+                }
+
+                warn!(
+                    "cpu.weight is accepted but not enforced yet; configured weight = {}",
```
lrh2000:
```suggestion
                    "`cpu.weight` is accepted but not enforced yet; weight = {}",
```


In fact, you can add a `cgroup: ` prefix to the log messages (via `__log_prefix` macro), to make it clear where the log messages come from?

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -102,16 +142,125 @@ impl CpuController {
...
+                };
+
+                // Check whether `period_usec` is valid.
+                let period_usec = if let Some(period_usec) =
+                    period_token.and_then(|period_token| period_token.parse::<u64>().ok())
+                {
+                    if !(MIN_PERIOD_USEC..=MAX_PERIOD_USEC).contains(&period_usec) {
+                        return Err(Error::InvalidOperation);
+                    }
+
```
lrh2000: Remove this blank line for consistency with code above.

lrh2000 (commented):
Thanks for adding these cgroup attributes.

It seems that these attributes are only used by the CFS scheduler. All other schedulers ignore them. I am not sure if we can enforce them, as I am not familiar with the scheduler code. However, adding dummy interfaces is fine with me.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -1,34 +1,66 @@
...
 /// This controller always exists so that `cpu.stat` remains readable even before `+cpu`
-/// is enabled. When inactive, only the base usage fields are exposed. The `user` and
-/// `system` counters store CPU time in units of one [`Jiffies`] per increment, and
-/// `usage_usec` is derived from their sum when `cpu.stat` is read.
+/// is enabled. When inactive, only the base usage fields are exposed and the non-root
+/// CPU control files are hidden.
 pub struct CpuController {
-    is_enabled: AtomicBool,
+    /// Persistent CPU usage accounting for this cgroup.
+    stats: Arc<CpuStats>,
```
lrh2000: If we don't have this `Arc`, we can just copy the value:
```rust
let user = PerCpuCounter::new();
// No one cares which CPU the count is on. Therefore, we chose the BSP for simplicity.
user.add_on_cpu(CpuId::bsp(), old.user.sum_all_cpus());
```

It's racy, but perhaps no one really cares (i.e., the counters added during controller enable/disable can be lost). For many other reasons, enabling and disabling the controller cannot be frequent operations.

Without this `Arc`, we can reduce the amount of indirection when updating the counters. It is on the fast path, which improves cache locality.

What do you think of this approach?

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -396,18 +400,13 @@ impl Controller {
...
-                    } else {
-                        guard.get().disable();
-                    }
+                    let previous_controller = guard.get();
+                    let new_controller = Arc::new(SubController::new_with_existing_stats(
+                        Some(parent_controller),
+                        previous_controller.deref_target(),
+                    ));
+                    child_node.controller().cpu.update(new_controller);
                 }
```
lrh2000, line 415: I don't think we need to perform memory allocation (`CpuStats` allocates `PerCpuCounter`) under the RCU guard (`guard.get()`).

In addition, I prefer more consistency to existing code.

So what I think is better is:
```rust
                    let mut new_controller: SubController<PidsController> =
                        SubController::new(Some(parent_controller));
                    new_controller.inner.as_mut().init_stats(...);
                    child_node.controller().cpu.update(new_controller);
```

`init_stats` should take `&mut self`, similar to `PidsController::init_count`, so it cannot be called by anything other than the initialization code.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -1,34 +1,66 @@
...
+/// CPU resource controls that are reset whenever `+cpu` is re-enabled.
+struct CpuControl {
+    /// A value stores the configured relative CPU share for scheduler integration.
+    weight: AtomicU32,
+    /// A value stores the configured CPU bandwidth limit for scheduler integration.
+    max: SpinLock<CpuMax>,
+}
+
+/// CPU usage accounting that survives CPU controller toggles.
+struct CpuStats {
```
lrh2000, line 36: Why did you put `CpuControl` before `CpuStats`? I can't think of any valid reason.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -84,34 +139,169 @@ impl CpuController {
...
+                    .to_str()
+                    .map_err(|_| Error::InvalidOperation)?
+                    .trim()
+                    .parse::<u32>()
+                    .map_err(|_| Error::InvalidOperation)?;
+                if !(MIN_WEIGHT..=MAX_WEIGHT).contains(&weight) {
+                    return Err(Error::InvalidOperation);
+                }
+
+                // TODO: Wire this value into group scheduling before removing this warning.
```
lrh2000: We should remove the warning. So let's rephrase it as:
```suggestion
                // TODO: Wire this value into group scheduling and remove this warning.
```

"Wire this value into group scheduling" also sounds odd. Is "group scheduling" a well-known term?

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -396,18 +400,16 @@ impl Controller {
...
-                        guard.get().enable();
-                    } else {
-                        guard.get().disable();
+                    let mut new_controller: SubController<CpuController> =
+                        SubController::new(Some(parent_controller));
+                    if let Some(inner) = new_controller.inner.as_mut() {
+                        let guard = child_node.controller().cpu.read();
+                        let previous_controller = guard.get();
+                        let previous_inner =
+                            previous_controller.deref_target().inner.as_ref().unwrap();
```
lrh2000: I think `deref_target` is redundant.

kernel/src/fs/fs_impls/cgroupfs/controller/mod.rs
```diff
@@ -396,18 +400,16 @@ impl Controller {
...
-                        .active_set()
-                        .contains_type(SubCtrlType::Cpu);
-                    let guard = child_node.controller().cpu.read();
-                    if is_enabled {
-                        guard.get().enable();
-                    } else {
-                        guard.get().disable();
+                    let mut new_controller: SubController<CpuController> =
+                        SubController::new(Some(parent_controller));
+                    if let Some(inner) = new_controller.inner.as_mut() {
```
lrh2000: This is inconsistent with how you accessed the previous controller's inner. Can we be consistent?

For example, something like?
```rust
                    {
                        let guard = child_node.controller().cpu.read();
                        let previous_controller = guard.get();
                        new_controller
                            .inner
                            .as_ref()
                            .unwrap()
                            .init_stats(previous_controller.inner.as_ref().unwrap());
                    }
```

lrh2000 (commented):
Thanks for revising the code. Just some minor follow-up comments.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
cchanging: > "Wire this value into group scheduling" also sounds odd. Is "group scheduling" a well-known term?

"group scheduling” is a known Linux scheduler term, especially around cgroups and CFS group scheduling. Yet this concept does not currently exist in our system. I’ll rephrase it in a simpler way.

lrh2000 (approved):

cqs21 (approved):

PR #3176, opened 2026-04-24
Refactor empty-path handling for `*at` syscalls

### LINKED ISSUE #3172: Introducing `EmptyPathStr` enum for `*at` syscalls
This issue propose a new abstraction: `EmptyPathStr` enum for `*at` syscalls.

### Three `FsPath` construction patterns have accumulated

Historically, every `*at` syscall in Asterinas built its `FsPath` the same way, using one call to `FsPath::from_fd_and_path(dirfd, &path)`. That constructor returns `ENOENT` when `path == ""`, so this blanket pattern baked in a "no empty paths, ever" rule across the whole `*at` surface.

Over time three inconsistent patterns have accumulated in `kernel/src/syscall/`:

- **Pattern A** — legacy "path is mandatory". Still used by `openat`, `mkdirat`, `mknodat`, `symlinkat`, `unlinkat`, `renameat[2]`, and was previously used by `readlink` / `readlinkat`:

  ```rust
  let fs_path = FsPath::from_fd_and_path(dirfd, &path_name)?;
  ```

- **Pattern B** — "empty path accepted only if caller sets `AT_EMPTY_PATH`". Used by `faccessat2`, `fchmodat2`, `fchownat`, `fstatat`, `statx`, `execveat`, `linkat` (oldname), and `utimensat`. Representative code from `kernel/src/syscall/access.rs`:

  ```rust
  let fs_path = if flags.contains(FaccessatFlags::AT_EMPTY_PATH)
      && path_name.is_empty()
  {
      FsPath::from_fd(dirfd)?
  } else {
      FsPath::from_fd_and_path(dirfd, &path_name)?
  };
  ```

  Every such file also redefines its own local `const AT_EMPTY_PATH = 0x1000;` inside a per-syscall `bitflags!` block — the same magic number, copy-pasted.

- **Pattern C** — "empty path always accepted, no flag gate." Recently adopted by `readlinkat`:

  ```rust
  let fs_path = if path_name.is_empty() {
      FsPath::from_fd(dirfd)?
  } else {
      FsPath::from_fd_and_path(dirfd, &path_name)?
  };
  ```

### Why three patterns is too many

Asterinas implements (or plans to implement) the following `*at` syscalls, and every one of them must pick a pattern:

```
openat, openat2, faccessat, faccessat2, fchmodat, fchmodat2, fchownat,
fstatat / newfstatat, statx, readlinkat, linkat, mkdirat, mknodat,
symlinkat, unlinkat, renameat, renameat2, utimensat, futimesat,
name_to_handle_at, execveat
```

Patterns A, B, and C differ only in whether an empty path is rejected, gated on a flag, or always accepted — a two-line diff that is easy to overlook in review. The consequences of picking the wrong one are real:

- **A where C is required** — the `readlinkat` bug: on an `O_PATH | O_NOFOLLOW` symlink fd, `readlinkat(dirfd, "", ...)` rejected with `ENOENT` instead of reading the link target.
- **C where B is required** — silently accepts `""` even when the caller didn't pass `AT_EMPTY_PATH`, violating the Linux ABI.
- **A where B is required** — blocks a legitimate `AT_EMPTY_PATH` caller with `ENOENT`.

Today the decision lives in prose (man pages, kernel source) and copy-pasted code. Reviewers have no mechanical check, and there is no single place to look up "which pattern is right for this new `*at` syscall?" This proposal distills that knowledge — derived directly from the Linux v6.8 source — into one type that every `*at` syscall must pass through.

### The rule the abstraction must encode

Two first principles together determine which pattern is correct for a given `*at` syscall. Both must hold for an empty path to be accepted:

**Principle 1 (semantic) — inode op vs. dentry op.** An empty path means "operate on whatever `dirfd` points at, directly." That is only meaningful if the syscall's target is an **inode** (the data or metadata an fd can identify). If the target is a **directory entry** — a name living inside a parent directory — an empty path is meaningless, because you cannot create, rename, remove, or open a nameless entry. So inode ops (`stat`, `statx`, `chmod`, `chown`, `access`, `utimens`, `readlink`, `execve`, `name_to_handle_at`, `linkat` oldname) are candidates for accepting an empty path; dentry ops (`open`, `mkdir`, `mknod`, `symlink`, `unlink`, `rename`, `linkat` newname) never are.

**Principle 2 (ABI) — is there room to carry `AT_EMPTY_PATH`?** The `AT_EMPTY_PATH` bit was added in Linux 2.6.39, after most `*at` syscalls had already shipped. Existing callers that passed `""` by accident had to keep getting `ENOENT`, so the new behaviour had to be **opt-in via a flag bit** — and that bit can only live in a `flags` argument. Four cases fall out: inode ops with a `flags` argument gate on `AT_EMPTY_PATH` (opt-in); dentry ops always reject regardless of shape; inode ops without a `flags` argument were handled in one of two ways — either Linux made the empty-path case always accepted (the sole example is `readlinkat`, because its semantics are safe and the ABI cost of introducing a `readlinkat2` was deemed not worth it), or a new `…2` variant was added to expose the flag (`faccessat2`, `fchmodat2`).

Composed into a single rule:

> An `*at` syscall accepts an empty path **iff** its target is addressable as a bare inode **and** the ABI has a way to say "I mean it" — either a `flags` word carrying `AT_EMPTY_PATH`, or (rarely) a lone syscall that Linux has chosen to always accept.

**Corollary for two-path syscalls.** Apply the rule per path. `linkat`'s *oldname* is an inode input (gate on `AT_EMPTY_PATH`); its *newname* is a new dentry being created (always reject). `renameat*`'s two paths are both dentries — both reject.

### Per-syscall classification (Linux v6.8)

| Syscall                | Empty `path` accepted?                                            | Target      |
|------------------------|-------------------------------------------------------------------|-------------|
| `readlinkat`           | **yes, always** — the sole always-accept case (no `flags` argument) | inode     |
| `faccessat2`           | yes, iff `AT_EMPTY_PATH` set                                      | inode       |
| `fchmodat2`            | yes, iff `AT_EMPTY_PATH` set                                      | inode       |
| `fchownat`             | yes, iff `AT_EMPTY_PATH` set                                      | inode       |
| `newfstatat` / `fstatat` | yes, iff `AT_EMPTY_PATH` set                                    | inode       |
| `statx`                | yes, iff `AT_EMPTY_PATH` set                                      | inode       |
| `utimensat`            | yes, iff `AT_EMPTY_PATH` set (also `path == NULL`, fd-only trick) | inode       |
| `name_to_handle_at`    | yes, iff `AT_EMPTY_PATH` set                                      | inode       |
| `execveat`             | yes, iff `AT_EMPTY_PATH` set                                      | inode       |
| `linkat` (oldname)     | yes, iff `AT_EMPTY_PATH` set **and** `CAP_DAC_READ_SEARCH`        | inode       |
| `linkat` (newname)     | **no**                                                            | dentry      |
| `openat`, `openat2`    | **no**                                                            | dentry      |
| `faccessat`            | **no** — predates `AT_EMPTY_PATH`, no `flags` arg                 | (inode)     |
| `fchmodat`             | **no** — predates `AT_EMPTY_PATH`, no `flags` arg                 | (inode)     |
| `mkdirat`              | **no**                                                            | dentry      |
| `mknodat`              | **no**                                                            | dentry      |
| `symlinkat`            | **no** (both args)                                                | dentry      |
| `unlinkat`             | **no** (`AT_EMPTY_PATH` passed → `EINVAL` from flag validator)    | dentry      |
| `renameat`, `renameat2` | **no** (both args)                                               | dentry      |
| `futimesat`            | **no** (but `path == NULL` works — orthogonal fd-only trick)      | (inode)     |
| `open_by_handle_at`    | n/a — no path argument                                            | —           |

The composed rule predicts every row. The three `(inode)` rows in parentheses — `faccessat`, `fchmodat`, `futimesat` — are inode ops without a `flags` argument that were *not* chosen to always accept, which is why Linux added `faccessat2` and `fchmodat2` to expose the opt-in.

### Definition

Add to `kernel/src/fs/vfs/path/resolver.rs`:

```rust
/// Policy for how [`FsPath::from_fd_at`] treats an empty `path_str`.
///
/// [`FsPath::from_fd_at`] is the entry point that 
/// every `*at` syscall uses to resolve its `(dirfd, path_str)` argument pair.
/// Whether an empty `path_str` should be rejected, conditionally accepted, 
/// or always accepted depends on the specific syscall's Linux semantics.
/// This enum makes that decision explicit at every call site.
///
/// # Examples
///
/// One call per variant, showing the three patterns side by side:
///
/// ```ignore
/// // mkdirat — dentry op, a name is mandatory.
/// let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::Reject)?;
///
/// // faccessat2 — inode op with a `flags` argument; `""` accepted
/// // only if the caller passed AT_EMPTY_PATH.
/// let fs_path = FsPath::from_fd_at(
///     dirfd,
///     &path_str,
///     EmptyPathStr::AllowIfFlag(flags.bits()),
/// )?;
///
/// // readlinkat — inode op with no `flags` argument; `""` always
/// // accepted and means "read the symlink `dirfd` points at".
/// let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::Allow)?;
/// ```
///
/// # Design Rationales
///
/// Two questions about a syscall determine which variant is correct:
///
/// 1. **Target.** Does the syscall operate on an *inode* 
///    (the data or metadata behind `dirfd`)
///    or on a *directory entry* (a name inside a parent directory)?
///    An empty `path_str` means "operate on `dirfd` directly" —
///    eaningful only for inode ops,
///    since there is no such thing as a nameless dentry 
///    to create, remove, rename, or open.
///
/// 2. **ABI.** Does the syscall carry a `flags` argument 
///    that can hold the [`AT_EMPTY_PATH`] opt-in bit?
///    `AT_EMPTY_PATH` post-dates most `*at` syscalls,
///    and callers historically passed `""` by accident (and received `ENOENT`),
///    so the accept-empty behaviour is opt-in —
///    and the opt-in must live in a `flags` word.
///
/// Quick chooser:
///
/// - dentry op → [`Reject`](Self::Reject)
/// - inode op with a `flags` argument → [`AllowIfFlag`](Self::AllowIfFlag)
/// - inode op that accepts `""` without a `flags` argument → [`Allow`](Self::Allow)
///   (rare; see the variant's docs)
pub enum EmptyPathStr {
    /// Always reject an empty `path_str` with `ENOENT`; a name is mandatory.
    ///
    /// Use for syscalls whose target is a **directory entry** —
    /// create, remove, rename, or open a named file.
    /// Examples: `openat`, `mkdirat`, `mknodat`, `symlinkat`, `unlinkat`,
    /// `renameat[2]`, and `linkat`'s *newname*.
    ///
    /// Also the right choice for the 3-argument forms
    /// `faccessat`, `fchmodat`, and `futimesat`:
    /// they are inode ops,
    /// but they predate [`AT_EMPTY_PATH`] and
    /// have no `flags` word to carry the opt-in.
    /// Newer code should prefer `faccessat2` / `fchmodat2` with
    /// [`AllowIfFlag`](Self::AllowIfFlag).
    Reject,

    /// Accept an empty `path_str`
    /// iff [`AT_EMPTY_PATH`] is set in the enclosed raw flag bits;
    /// otherwise behave like [`Reject`](Self::Reject).
    ///
    /// Callers pass their syscall's raw flag bits —
    /// typically `flags.bits()` after parsing the flag word into a typed `BitFlags`.
    /// Only the [`AT_EMPTY_PATH`] bit is inspected; the rest are ignored.
    ///
    /// Use for **inode-target** syscalls that carry a `flags` argument —
    /// the common case for attribute queries and modifications on the inode behind `dirfd`.
    /// Examples: `faccessat2`, `fchmodat2`, `fchownat`, `fstatat` / `newfstatat`,
    /// `statx`, `utimensat`, `name_to_handle_at`, `execveat`, and `linkat`'s *oldname*.
    ///
    /// Any additional gating on top of [`AT_EMPTY_PATH`]
    /// (for instance, `linkat` also requires `CAP_DAC_READ_SEARCH`)
    /// stays in the syscall implementation;
    /// this enum only decides whether an empty `path_str` is syntactically acceptable.
    AllowIfFlag(u32),

    /// Always accept an empty `path_str` and operate on `dirfd` directly.
    /// No flag is consulted.
    ///
    /// Use only for inode-target syscalls
    /// that Linux has chosen to accept an empty `path_str` unconditionally,
    /// because they have no `flags` argument
    /// in which [`AT_EMPTY_PATH`] could live.
    /// This is deliberately rare:
    /// as of Linux 6.8 the sole such syscall is **`readlinkat`**.
    Allow,
}

/// The `AT_EMPTY_PATH` flag bit, as defined by Linux. 
pub const AT_EMPTY_PATH: u32 = 0x1000;
```

### The helper that consumes it

```rust
impl<'a> FsPath<'a> {
    /// Construct an `FsPath` from `(dirfd, path_str)`,
    /// applying the syscall-specific empty-path-string policy.
    ///
    /// An empty `path_str` is handled per the variant of `policy`;
    /// see [`EmptyPathStr`] for the full rule.
    pub fn from_fd_at(
        dirfd: RawFileDesc,
        path_str: &'a str,
        policy: EmptyPathStr,
    ) -> Result<Self> {
        if path_str.is_empty() {
            let allowed = matches!(policy, EmptyPathStr::Allow)
                || matches!(policy, EmptyPathStr::AllowIfFlag(f) if f & AT_EMPTY_PATH != 0);
            return if allowed {
                Self::from_fd(dirfd)
            } else {
                return_errno_with_message!(Errno::ENOENT, "the path is empty")
            };
        }
        Self::from_fd_and_path(dirfd, path_str)
    }
}
```

### Usage across the `*at` surface

Every `*at` syscall picks the variant that matches its target and ABI:

| Syscall                                                                                                  | Variant                                             |
|----------------------------------------------------------------------------------------------------------|-----------------------------------------------------|
| `readlinkat`                                                                                             | `EmptyPathStr::Allow`                                  |
| `faccessat2`, `fchmodat2`, `fchownat`, `fstatat`, `statx`, `utimensat`, `name_to_handle_at`, `execveat`  | `EmptyPathStr::AllowIfFlag(flags.bits())`                   |
| `linkat`                                                                                                 | oldname: `AllowIfFlag(flags.bits())`; newname: `Reject`  |
| `openat`, `openat2`, `faccessat`, `fchmodat`, `mkdirat`, `mknodat`, `symlinkat`, `unlinkat`, `renameat`, `renameat2`, `futimesat` | `EmptyPathStr::Reject`                     |

Concrete calls:

```rust
// readlinkat.rs — now a one-liner:
let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::Allow)?;

// access.rs (faccessat2) — no more inline `if empty && flag`:
let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::AllowIfFlag(flags.bits()))?;

// mkdir.rs — still rejects empty paths, but now says so explicitly:
let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::Reject)?;
```

### Why this is the right shape

The three variants correspond 1:1 to the three rows of the principle table, so picking the wrong one is as visible at a callsite as picking the wrong syscall implementation — there is no silent default. The doc comments encode the survey directly: the "inode vs dentry" justification lives next to the call, not in a separate document a reviewer has to remember to consult. The copy-pasted flag constants go away: seven-plus local `const AT_EMPTY_PATH = 0x1000;` declarations collapse into one shared constant. The `readlinkat` empty-path fix becomes a one-variant change — it switches from `Reject`-equivalent behaviour to `Allow`, and the fix is visible in the type signature of the call rather than buried in an if-chain. And unrelated quirks stay where they belong: `linkat`'s `CAP_DAC_READ_SEARCH` check, `utimensat` / `futimesat`'s `NULL`-pathname short-circuit, and `unlinkat`'s flag validator all live in their own syscall modules, because `EmptyPathStr` only answers one question — "is `""` a valid path here?"

`EmptyPathStr::AllowIfFlag(u32)` takes raw flag bits rather than a typed `BitFlags`. This keeps the helper from depending on every syscall's local flag enum, at the cost of one `flags.bits()` conversion per callsite. A plainer `bool allow_empty` parameter would be simpler still but less self-documenting and easy to flip by mistake. The enum is recommended; swap to the bool if the cast is judged too noisy.

## Migration plan

1. Add `EmptyPathStr`, `AT_EMPTY_PATH`, and `FsPath::from_fd_at` in `resolver.rs` without touching any syscall.
2. Migrate `readlinkat` first — this lands the empty-path fix.
3. Migrate the Pattern-B syscalls (`faccessat2`, `fchmodat2`, `fchownat`, `fstatat`, `statx`, `execveat`, `linkat`, `utimensat`), deleting their local `AT_EMPTY_PATH` bitflag constants.
4. Migrate the Pattern-A syscalls last, mechanically replacing `FsPath::from_fd_and_path(...)?` with `FsPath::from_fd_at(..., EmptyPathStr::Reject)?`.
5. Consider demoting `FsPath::from_fd_and_path` from the public API (make it `pub(super)` or inline it into `from_fd_at`) so new syscalls cannot bypass the policy.

cqs21:
Fixes #3172

This PR implements the empty-path handling design proposed in #3172 and then enables more LTP syscall cases.

### DIFF
--- a/kernel/src/fs/vfs/path/resolver.rs
+++ b/kernel/src/fs/vfs/path/resolver.rs
@@ -24,6 +24,115 @@ use crate::{
 /// The file descriptor of the current working directory.
 pub const AT_FDCWD: RawFileDesc = -100;
 
+/// The `AT_EMPTY_PATH` flag bit, as defined by Linux.
+pub const AT_EMPTY_PATH: u32 = 0x1000;
+
+/// Policy for how [`FsPath::from_fd_at`] treats an empty `path_str`.
+///
+/// [`FsPath::from_fd_at`] is the entry point that
+/// every `*at` syscall uses to resolve its `(dirfd, path_str)` argument pair.
+/// Whether an empty `path_str` should be rejected, conditionally accepted,
+/// or always accepted depends on the specific syscall's Linux semantics.
+/// This enum makes that decision explicit at every call site.
+///
+/// # Examples
+///
+/// One call per variant, showing the three patterns side by side:
+///
+/// ```ignore
+/// // mkdirat — dentry op, a name is mandatory.
+/// let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::Reject)?;
+///
+/// // faccessat2 — inode op with a `flags` argument; `""` accepted
+/// // only if the caller passed AT_EMPTY_PATH.
+/// let fs_path = FsPath::from_fd_at(
+///     dirfd,
+///     &path_str,
+///     EmptyPathStr::AllowIfFlag(flags.bits()),
+/// )?;
+///
+/// // readlinkat — inode op with no `flags` argument; `""` accepted
+/// // only when `dirfd` is a real fd rather than `AT_FDCWD`.
+/// let fs_path = FsPath::from_fd_at(dirfd, &path_str, EmptyPathStr::Allow)?;
+/// ```
+///
+/// # Design Rationales
+///
+/// Two questions about a syscall determine which variant is correct:
+///
+/// 1. **Target.** Does the syscall operate on an *inode*
+///    (the data or metadata behind `dirfd`)
+///    or on a *directory entry* (a name inside a parent directory)?
+///    An empty `path_str` means "operate on `dirfd` directly" —
+///    meaningful only for inode ops,
+///    since there is no such thing as a nameless dentry
+///    to create, remove, rename, or open.
+///
+/// 2. **ABI.** Does the syscall carry a `flags` argument
+///    that can hold the [`AT_EMPTY_PATH`] opt-in bit?
+///    `AT_EMPTY_PATH` post-dates most `*at` syscalls,
+///    and callers historically passed `""` by accident (and received `ENOENT`),
+///    so the accept-empty behaviour is opt-in —
+///    and the opt-in must live in a `flags` word.
+///
+/// Quick chooser:
+///
+/// - dentry op → [`Reject`](Self::Reject)
+/// - inode op with a `flags` argument → [`AllowIfFlag`](Self::AllowIfFlag)
+/// - inode op that accepts `""` without a `flags` argument → [`Allow`](Self::Allow)
+///   (rare; see the variant's docs)
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum EmptyPathStr {
+    /// Always reject an empty `path_str` with `ENOENT`; a name is mandatory.
+    ///
+    /// Use for syscalls whose target is a **directory entry** —
+    /// create, remove, rename, or open a named file.
+    /// Examples: `openat`, `mkdirat`, `mknodat`, `symlinkat`, `unlinkat`,
+    /// `renameat[2]`, and `linkat`'s *newname*.
+    ///
+    /// Also the right choice for the 3-argument forms
+    /// `faccessat`, `fchmodat`, and `futimesat`:
+    /// they are inode ops,
+    /// but they predate [`AT_EMPTY_PATH`] and
+    /// have no `flags` word to carry the opt-in.
+    /// Newer code should prefer `faccessat2` / `fchmodat2` with
+    /// [`AllowIfFlag`](Self::AllowIfFlag).
+    Reject,
+
+    /// Accept an empty `path_str`
+    /// iff [`AT_EMPTY_PATH`] is set in the enclosed raw flag bits;
+    /// otherwise behave like [`Reject`](Self::Reject).
+    ///
+    /// Callers pass their syscall's raw flag bits —
+    /// typically `flags.bits()` after parsing the flag word into a typed `BitFlags`.
+    /// Only the [`AT_EMPTY_PATH`] bit is inspected; the rest are ignored.
+    ///
+    /// Use for **inode-target** syscalls that carry a `flags` argument —
+    /// the common case for attribute queries and modifications on the inode behind `dirfd`.
+    /// Examples: `faccessat2`, `fchmodat2`, `fchownat`, `fstatat` / `newfstatat`,
+    /// `statx`, `utimensat`, `name_to_handle_at`, `execveat`, and `linkat`'s *oldname*.
+    ///
+    /// Any additional gating on top of [`AT_EMPTY_PATH`]
+    /// (for instance, `linkat` also requires `CAP_DAC_READ_SEARCH`)
+    /// stays in the syscall implementation;
+    /// this enum only decides whether an empty `path_str` is syntactically acceptable.
+    AllowIfFlag(u32),
+
+    /// Accept an empty `path_str` and operate on `dirfd` directly.
+    /// No flag is consulted.
+    ///
+    /// `dirfd` must be a real file descriptor; if it is [`AT_FDCWD`],
+    /// [`FsPath::from_fd_at`] still rejects `""` with `ENOENT`.
+    ///
+    /// Use only for inode-target syscalls
+    /// that Linux has chosen to accept an empty `path_str` unconditionally,
+    /// because they have no `flags` argument
+    /// in which [`AT_EMPTY_PATH`] could live.
+    /// This is deliberately rare:
+    /// as of Linux 6.8 the sole such syscall is **`readlinkat`**.
+    Allow,
+}
+
 /// A resolver for [`Path`]s.
 ///
 /// `PathResolver` provides a context for resolving paths, defined by a root directory
@@ -741,14 +850,32 @@ impl<'a> FsPath<'a> {
         })
     }
 
-    /// Creates a new `FsPath` from the given `dirfd` and `path`.
+    /// Constructs an `FsPath` from `(dirfd, path_str)`,
+    /// applying the syscall-specific empty-path-string `policy`.
     ///
-    /// If the FD is not valid (i.e., it's negative and it's not [`AT_FDCWD`]) or the path is empty
-    /// or too long, an error will be returned.
-    pub fn from_fd_and_path(dirfd: RawFileDesc, path: &'a str) -> Result<Self> {
+    /// An empty `path_str` is handled per the variant of `policy`;
+    /// see [`EmptyPathStr`] for the full rule. A non-empty absolute `path`
+    /// ignores `dirfd`; otherwise `dirfd` must be either [`AT_FDCWD`] or a
+    /// valid file descriptor.
+    ///
+    /// # Errors
+    ///
+    /// - `ENOENT` if `path` is empty and `policy` does not permit it.
+    /// - `ENAMETOOLONG` if `path.len() > PATH_MAX`.
+    /// - `EBADF` if `dirfd` is negative and is not [`AT_FDCWD`].
+    pub fn from_fd_at(dirfd: RawFileDesc, path: &'a str, policy: EmptyPathStr) -> Result<Self> {
         if path.is_empty() {
-            return_errno_with_message!(Errno::ENOENT, "the path is empty")
+            let allowed = match policy {
+                EmptyPathStr::Reject => false,
+                EmptyPathStr::AllowIfFlag(f) => f & AT_EMPTY_PATH != 0,
+                EmptyPathStr::Allow => dirfd != AT_FDCWD,
+            };
+            if !allowed {
+                return_errno_with_message!(Errno::ENOENT, "the path is empty");
+            }
+            return Self::from_fd(dirfd);
         }
+
         if path.len() > PATH_MAX {
             return_errno_with_message!(Errno::ENAMETOOLONG, "the path is too long");
         }
@@ -771,7 +898,7 @@ impl<'a> TryFrom<&'a str> for FsPath<'a> {
     type Error = crate::error::Error;
 
     fn try_from(path: &'a str) -> Result<FsPath<'a>> {
-        FsPath::from_fd_and_path(AT_FDCWD, path)
+        FsPath::from_fd_at(AT_FDCWD, path, EmptyPathStr::Reject)
     }
 }
 
@@ -923,5 +1050,46 @@ mod test {
             (" //b//a/ ", Ok((" //b//a", " "))),
         ];
         assert_split_results(&cases, SplitPath::split_dirname_and_basename);
+    }
+
+    #[ktest]
+    fn fs_path_from_fd_at_policy_matrix() {
+        // Reject: empty always ENOENT
+        assert!(
+            FsPath::from_fd_at(AT_FDCWD, "", EmptyPathStr::Reject)
+                .is_err_and(|e| e.error() == Errno::ENOENT)
+        );
+        assert!(
+            FsPath::from_fd_at(3, "", EmptyPathStr::Reject)
+                .is_err_and(|e| e.error() == Errno::ENOENT)
+        );
+
+        // AllowIfFlag: only AT_EMPTY_PATH bit opts in
+        assert!(
+            FsPath::from_fd_at(3, "", EmptyPathStr::AllowIfFlag(0))
+                .is_err_and(|e| e.error() == Errno::ENOENT)
+        );
+        assert!(FsPath::from_fd_at(3, "", EmptyPathStr::AllowIfFlag(AT_EMPTY_PATH)).is_ok());
+        // Unrelated bits do not opt in.
+        assert!(
+            FsPath::from_fd_at(3, "", EmptyPathStr::AllowIfFlag(0x100))
+                .is_err_and(|e| e.error() == Errno::ENOENT)
+        );
+
+        // Allow: rejected when dirfd == AT_FDCWD, permitted otherwise
+        assert!(
+            FsPath::from_fd_at(AT_FDCWD, "", EmptyPathStr::Allow)
+                .is_err_and(|e| e.error() == Errno::ENOENT)
+        );
+        assert!(FsPath::from_fd_at(3, "", EmptyPathStr::Allow).is_ok());
+    }
+
+    #[ktest]
+    fn fs_path_from_fd_at_rejects_too_long() {
+        let long = "a".repeat(PATH_MAX + 1);
+        assert!(
+            FsPath::from_fd_at(AT_FDCWD, &long, EmptyPathStr::Reject)
+                .is_err_and(|e| e.error() == Errno::ENAMETOOLONG)
+        );
     }
 }
--- a/kernel/src/syscall/utimens.rs
+++ b/kernel/src/syscall/utimens.rs
@@ -9,7 +9,7 @@ use crate::{
     fs,
     fs::{
         file::file_table::RawFileDesc,
-        vfs::path::{AT_FDCWD, FsPath, Path},
+        vfs::path::{AT_FDCWD, EmptyPathStr, FsPath, Path},
     },
     prelude::*,
     time::{clocks::RealTimeCoarseClock, timespec_t, timeval_t},
@@ -20,6 +20,7 @@ use crate::{
 /// and `times[1]` represents the modification time.
 /// The `flags` argument is a bit mask that can include the following values:
 /// - `AT_SYMLINK_NOFOLLOW`: If set, the file is not dereferenced if it is a symbolic link.
+/// - `AT_EMPTY_PATH`: If set, an empty pathname means operating on `dirfd` directly.
 pub fn sys_utimensat(
     dirfd: RawFileDesc,
     pathname_ptr: Vaddr,
@@ -172,8 +173,15 @@ fn do_utimes(
 
     let path = {
         let fs_path = if let Some(pathname) = pathname.as_ref() {
-            FsPath::from_fd_and_path(dirfd, pathname)?
+            FsPath::from_fd_at(dirfd, pathname, EmptyPathStr::AllowIfFlag(flags.bits()))?
         } else {
+            // Matches Linux `do_utimes_fd`: no flags are accepted when pathname is NULL.
+            if !flags.is_empty() {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "flags must be zero when pathname is NULL"
+                );
+            }
             FsPath::from_fd(dirfd)?
         };
 
@@ -255,6 +263,7 @@ const UTIME_OMIT: i64 = (1i64 << 30) - 2i64;
 
 bitflags::bitflags! {
     struct UtimensFlags: u32 {
+        const AT_EMPTY_PATH = 0x1000;
         const AT_SYMLINK_NOFOLLOW = 0x100;
     }
 }
--- a/test/initramfs/src/conformance/ltp/testcases/all.txt
+++ b/test/initramfs/src/conformance/ltp/testcases/all.txt
@@ -8,8 +8,8 @@
 # accept4_01
 
 # access01
-# access02
-# access03
+access02
+access03
 # access04
 
 # acct01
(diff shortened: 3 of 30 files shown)


kernel/src/syscall/utimens.rs
```diff
@@ -172,7 +173,7 @@ fn do_utimes(
 
     let path = {
         let fs_path = if let Some(pathname) = pathname.as_ref() {
-            FsPath::from_fd_and_path(dirfd, pathname)?
+            FsPath::from_fd_at(dirfd, pathname, EmptyPathStr::AllowIfFlag(flags.bits()))?
         } else {
             FsPath::from_fd(dirfd)?
         };
```
tatetian, line 186: **`utimensat` NULL-pathname branch no longer rejects non-zero flags** (significant): The NULL-pathname short-circuit (`pathname_ptr == 0 && dirfd != AT_FDCWD`) skips `from_fd_at` and calls `FsPath::from_fd(dirfd)` directly, which consults no flags. In Linux, NULL pathname dispatches to `do_utimes_fd`, which rejects any non-zero `flags` with `EINVAL` (per Linux 6.x `fs/utimes.c` — the same file whose NULL-handling line range is already cited at `utimens.rs:163-164`). Adding `AT_EMPTY_PATH` to `UtimensFlags` in this PR widens the gap: `utimensat(validfd, NULL, times, AT_EMPTY_PATH)` now silently succeeds where Linux returns `EINVAL`. The same is true for `AT_SYMLINK_NOFOLLOW` with NULL pathname, which was a latent divergence already but is worth fixing together. Reject non-zero `flags` on the NULL branch to match `do_utimes_fd`.

Fix:

```rust
let path = {
    let fs_path = if let Some(pathname) = pathname.as_ref() {
        FsPath::from_fd_at(dirfd, pathname, EmptyPathStr::AllowIfFlag(flags.bits()))?
    } else {
        // Matches Linux `do_utimes_fd`: no flags are accepted when pathname is NULL.
        if !flags.is_empty() {
            return_errno_with_message!(Errno::EINVAL, "flags must be zero when pathname is NULL");
        }
        FsPath::from_fd(dirfd)?
    };
    // ...
};
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -747,14 +856,22 @@ impl<'a> FsPath<'a> {
...
         if path.is_empty() {
-            return_errno_with_message!(Errno::ENOENT, "the path is empty")
+            let allowed = matches!(policy, EmptyPathStr::Allow if dirfd != AT_FDCWD)
+                || matches!(policy, EmptyPathStr::AllowIfFlag(f) if f & AT_EMPTY_PATH != 0);
+            return if allowed {
+                Self::from_fd(dirfd)
+            } else {
+                return_errno_with_message!(Errno::ENOENT, "the path is empty")
+            };
         }
```
tatetian, line 877: [**`minimize-nesting`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/functions-and-methods.html#minimize-nesting) (minor): Two small readability nits in the empty-path branch. The two `matches!` arms scan `policy` twice where a single `match` covers all three variants exhaustively. And `return_errno_with_message!` already expands to `return Err(...)`, so wrapping it in `return if allowed { ... } else { ... };` is an odd `return`-expression whose `else` arm itself returns; an early-return guard reads more idiomatically.

Fix:

```rust
if path.is_empty() {
    let allowed = match policy {
        EmptyPathStr::Reject => false,
        EmptyPathStr::AllowIfFlag(f) => f & AT_EMPTY_PATH != 0,
        EmptyPathStr::Allow => dirfd != AT_FDCWD,
    };
    if !allowed {
        return_errno_with_message!(Errno::ENOENT, "the path is empty");
    }
    return Self::from_fd(dirfd);
}
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -747,14 +856,22 @@ impl<'a> FsPath<'a> {
...
 
-    /// Creates a new `FsPath` from the given `dirfd` and `path`.
+    /// Construct an `FsPath` from `(dirfd, path_str)`,
+    /// applying the syscall-specific empty-path-string policy.
     ///
-    /// If the FD is not valid (i.e., it's negative and it's not [`AT_FDCWD`]) or the path is empty
-    /// or too long, an error will be returned.
-    pub fn from_fd_and_path(dirfd: RawFileDesc, path: &'a str) -> Result<Self> {
+    /// An empty `path_str` is handled per the variant of `policy`;
+    /// see [`EmptyPathStr`] for the full rule.
```
tatetian: [**`rfc1574-summary`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/comments-and-documentation.html#rfc1574-summary) (minor): Summary line uses the imperative "Construct" where the sibling `from_fd` at line 847 correctly uses "Creates". While here, this is the primary entry point for every `*at` syscall, so the doc deserves an `# Errors` section listing the three failure modes (`ENOENT`, `ENAMETOOLONG`, and `EBADF` via `FileDesc::try_from`) rather than delegating everything to `EmptyPathStr`.

Fix:

```rust
/// Constructs an `FsPath` from `(dirfd, path_str)`,
/// applying the syscall-specific empty-path-string `policy`.
///
/// An empty `path_str` is handled per the variant of `policy`;
/// see [`EmptyPathStr`] for the full rule. A non-empty absolute `path`
/// ignores `dirfd`; otherwise `dirfd` must be either [`AT_FDCWD`] or a
/// valid file descriptor.
///
/// # Errors
///
/// - `ENOENT` if `path` is empty and `policy` does not permit it.
/// - `ENAMETOOLONG` if `path.len() > PATH_MAX`.
/// - `EBADF` if `dirfd` is negative and is not [`AT_FDCWD`].
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -24,6 +24,115 @@ use crate::{
 /// The file descriptor of the current working directory.
 pub const AT_FDCWD: RawFileDesc = -100;
 
+/// The `AT_EMPTY_PATH` flag bit, ad defined by Linux.
```
tatetian: **Typo** (minor): "ad defined" should be "as defined".

Fix:

```rust
/// The `AT_EMPTY_PATH` flag bit, as defined by Linux.
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -24,6 +24,115 @@ use crate::{
...
+///
+/// # Design Rationales
+///
+/// Two questions about a syscall determine which variant is correct:
+///
+/// 1. **Target.** Does the syscall operate on an *inode*
+///    (the data or metadata behind `dirfd`)
+///    or on a *directory entry* (a name inside a parent directory)?
+///    An empty `path_str` means "operate on `dirfd` directly" —
+///    eaningful only for inode ops,
```
tatetian: **Typo** (minor): Missing leading `m` — should be "meaningful".

Fix:

```rust
///    An empty `path_str` means "operate on `dirfd` directly" —
///    meaningful only for inode ops,
```

kernel/src/fs/vfs/path/resolver.rs
tatetian, line 1: [**`add-regression-tests`**](https://asterinas.github.io/book/to-contribute/coding-guidelines/testing-guidelines.html#add-regression-tests) (significant): No unit tests were added for the new `from_fd_at` + `EmptyPathStr` API, despite an existing `#[cfg(ktest)]` module in the same file covering `split_dirname_*` thoroughly. The three-variant policy × three dirfd kinds (AT_FDCWD, real fd, invalid negative) × (empty, non-empty) matrix is cheap to cover in-tree and would pin down the subtle rules — "`Allow` still rejects when `dirfd == AT_FDCWD`", "`AllowIfFlag` inspects only the `AT_EMPTY_PATH` bit" — against future regressions. A bit-mask typo or a flipped `!=` in either pattern guard would slip past all currently-enabled LTP tests (only `statx02` exercises an `AT_EMPTY_PATH` code path end-to-end).

Fix — add tests next to the existing ones in the ktest module:

```rust
#[ktest]
fn fs_path_from_fd_at_policy_matrix() {
    // Reject: empty always ENOENT
    assert_eq!(
        FsPath::from_fd_at(AT_FDCWD, "", EmptyPathStr::Reject).map_err(|e| e.error()),
        Err(Errno::ENOENT)
    );
    assert_eq!(
        FsPath::from_fd_at(3, "", EmptyPathStr::Reject).map_err(|e| e.error()),
        Err(Errno::ENOENT)
    );

    // AllowIfFlag: only AT_EMPTY_PATH bit opts in
    assert_eq!(
        FsPath::from_fd_at(3, "", EmptyPathStr::AllowIfFlag(0)).map_err(|e| e.error()),
        Err(Errno::ENOENT)
    );
    assert!(FsPath::from_fd_at(3, "", EmptyPathStr::AllowIfFlag(AT_EMPTY_PATH)).is_ok());
    // Unrelated bits do not opt in.
    assert_eq!(
        FsPath::from_fd_at(3, "", EmptyPathStr::AllowIfFlag(0x100))
            .map_err(|e| e.error()),
        Err(Errno::ENOENT)
    );

    // Allow: rejected when dirfd == AT_FDCWD, permitted otherwise
    assert_eq!(
        FsPath::from_fd_at(AT_FDCWD, "", EmptyPathStr::Allow).map_err(|e| e.error()),
        Err(Errno::ENOENT)
    );
    assert!(FsPath::from_fd_at(3, "", EmptyPathStr::Allow).is_ok());
}

#[ktest]
fn fs_path_from_fd_at_rejects_too_long() {
    let long = "a".repeat(PATH_MAX + 1);
    assert_eq!(
        FsPath::from_fd_at(AT_FDCWD, &long, EmptyPathStr::Reject).map_err(|e| e.error()),
        Err(Errno::ENAMETOOLONG)
    );
}
```

test/initramfs/src/conformance/ltp/testcases/all.txt
tatetian, line 1: **Coverage gap: most newly-enabled LTP tests don't exercise the changed code** (minor): The 16 newly-uncommented entries (`access02`, `access03`, `chmod03`, `chmod07`, `faccessat201`, `fchmod02`, `fchmod03`, `fchmod04`, `fchmodat02`, `fchmodat2_02`, `lstat02`, `lstat02_64`, `mknod02`, `mknod06`, `symlink03`, `unlink07`) appear (per the testing sub-agent's read of the LTP sources) to mostly cover unrelated validation paths (sticky-bit, ownership, ELOOP, ENAMETOOLONG); only the `Reject`-branch empty-path case (empty → `ENOENT`, the pre-PR behavior) is exercised, if any. The code paths this PR actually adds — `AT_EMPTY_PATH` on `fchownat`, `fstatat`, `statx` (only indirectly via the already-enabled `statx02`), `utimensat`, `linkat`'s oldname, and `execveat` — have no enabled LTP coverage. Consider enabling the relevant LTP tests that specifically cover AT_EMPTY_PATH (`execveat01`, `statx12`, `utimensat01` where they pass on Asterinas). If there are relevant Gvisor test cases, also enable them.

tatetian (commented):
Thanks for the quick implementation.

test/initramfs/src/conformance/ltp/testcases/all.txt
cqs21: > Consider enabling the relevant LTP tests that specifically cover AT_EMPTY_PATH (`execveat01`, `statx12`, `utimensat01` where they pass on Asterinas).

These tests are not enabled because they still fail for reasons unrelated to `AT_EMPTY_PATH`.

tatetian (approved):

PR #3177, opened 2026-04-24
Refine rfc1574-summary coding guideline to distinguish items and types

tatetian:
The previous wording applied "third-person singular present indicative" uniformly, which matches function/method conventions but conflicts with the Rust stdlib practice of using noun phrases to describe types, modules, and fields. Split the rule along that axis and add a type-level example.


tatetian:
Merge a trivial change to the doc without extra review

Issue #3178, opened 2026-04-24
RISC-V regression: clone3(CLONE_FILES) does not share file table as expected


TankTechnology:
## Summary

While validating RISC-V regression tests, `/test/process` fails in `test_clone_files`. The test expects a file descriptor closed by a `clone3(CLONE_FILES)` child to also be closed in the parent, but the parent can still close it successfully.

## Reproduction

Environment:
- Docker image: `asterinas/asterinas:0.17.2-20260407`
- Target: `riscv64`
- Branch used for testing: `investigate-riscv-network-device`, which includes #3174 and a local RISC-V virtio-net QEMU config change

Command:

```bash
CARGO_NET_GIT_FETCH_WITH_CLI=true \
CARGO_HTTP_MULTIPLEXING=false \
CARGO_TARGET_DIR=/tmp/asterinas-target \
make run_kernel OSDK_TARGET_ARCH=riscv64 AUTO_TEST=regression INITRAMFS_SKIP_GZIP=1
```

## Observed Result

The run reaches `/test/process` and fails at `test_clone_files`:

```text
test_clone_files: `syscall(59, pipefds, 02000000)` passed [got Success]
test_clone_files: `syscall(24, pipefds[0], dupped_pipe_fd, 02000000)` passed [got Success]
test_clone_files: `close(pipefds[0])` passed [got Success]
test_clone_files: `sys_clone3(&args)` passed [got Success]
test_clone_files: `wait4(pid, &status, 0, ((void *)0))` passed [got Success]
test_clone_files: `close(dupped_pipe_fd)` passed [got Success]
test_clone_files: `close(pipefds[1])` failed [got Success, but expected Bad file descriptor]
test_clone_files summary: 8 tests passed, 1 tests failed
Regression test failed
```

## Expected Result

`clone3(CLONE_FILES)` should share the file descriptor table between parent and child. After the child closes `pipefds[1]`, the parent should get `EBADF` when closing the same fd.

## Notes

The nearby `clone(CLONE_FILES)` regression test (`test_clone_files_and_close` in `process/clone3/clone_files.c`) passes, so this appears specific to the `clone3` path or to how the `clone3` arguments/flags are handled on RISC-V.

Relevant test source:
- `test/initramfs/src/regression/process/execve/execve_mt_parent.c`, `FN_TEST(clone_files)`
- `test/initramfs/src/regression/process/clone3/clone_files.c`, `FN_TEST(clone_files_and_close)`

Relevant kernel paths:
[...]

PR #3179, opened 2026-04-24
Enable virtio-net for the RISC-V QEMU scheme

TankTechnology:
## Summary

- Add a user-mode QEMU netdev to the RISC-V scheme.
- Attach a `virtio-net-device` so the RISC-V guest exposes a virtio network interface.
- This aligns the RISC-V QEMU scheme with network regression tests that expect both `lo` and `eth0` to be present.

## Test Plan

- `git diff --check upstream/main..HEAD`
- `CARGO_TARGET_DIR=/tmp/asterinas-target make run_kernel OSDK_TARGET_ARCH=riscv64 AUTO_TEST=boot INITRAMFS_SKIP_GZIP=1`

Additional validation while investigating RISC-V regression tests:

- With this change stacked on #3174, `/test/network` passed, including `test_if_nameindex` and `test_get_link_by_libnl`.
- A full RISC-V regression run is currently blocked before the network tests by #3178 (`clone3(CLONE_FILES)` does not share the file table as expected on RISC-V).

### DIFF
--- a/OSDK.toml
+++ b/OSDK.toml
@@ -62,6 +62,7 @@ qemu.args = """\
     -serial chardev:mux \
     -monitor chardev:mux \
     -chardev stdio,id=mux,mux=on,signal=off,logfile=qemu.log \
+    -netdev user,id=net01 \
     -drive if=none,format=raw,id=x0,file=./test/initramfs/build/ext2.img \
     -drive if=none,format=raw,id=x1,file=./test/initramfs/build/exfat.img \
     # NOTE: The `/etc/profile.d/init.sh` assumes that `ext2.img` appears as the first block device (`/dev/vda`).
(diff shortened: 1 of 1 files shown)


tatetian:
@TankTechnology Does `virtio-net` work correctly on RISC-V? This new QEMU argument is enabled without a test. What are the gaps to enable network CI tests on RISC-V? I would feel more comfortable when virtio-net is enabled with CI tests that show it actually work. Accepting this PR as is now would benefit no real user (code or human).

tatetian:
> @TankTechnology Does `virtio-net` work correctly on RISC-V? This new QEMU argument is enabled without a test. What are the gaps to enable network CI tests on RISC-V? I would feel more comfortable when virtio-net is enabled with CI tests that show it actually work. Accepting this PR as is now would benefit no real user (code or human).

@TankTechnology A kind reminder

PR #3180, opened 2026-04-27
Fix `cargo osdk test` exiting after the first default member

StevenJiang1110:
#3107 changed `make ktest` from iterating over each crate in the Makefile and invoking cargo osdk run individually, to having OSDK internally iterate over all default members and run ktests for each crate. However, the previous OSDK behavior was to exit immediately after each ktests run invocation, which caused `make ktest` to exit after running only the first crate. This PR fixes that behavior by deferring the exit until all crates have been processed.

https://github.com/asterinas/asterinas/blob/7bcb2fe3849b64788a945d1b916c52a16bce8a0c/osdk/src/bundle/mod.rs#L306-L321

### DIFF
--- a/osdk/src/bundle/mod.rs
+++ b/osdk/src/bundle/mod.rs
@@ -9,7 +9,7 @@ use file::{BundleFile, Initramfs};
 use std::{
     io::{BufRead, BufReader, Write},
     os::unix::net::UnixStream,
-    process,
+    process::{self, ExitStatus},
     time::Duration,
 };
 use tempfile::NamedTempFile;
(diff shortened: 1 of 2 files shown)


osdk/src/bundle/mod.rs
```diff
@@ -52,6 +52,39 @@ pub struct BundleManifest {
     pub last_modified: SystemTime,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum VmExit {
```
junyang-zh: VM exit is a specific term. Please change to `QemuExit`. The same applies to other function and variable names.

osdk/src/bundle/mod.rs
```diff
@@ -345,6 +381,7 @@ impl Bundle {
                 std::thread::sleep(Duration::from_millis(100));
             }
         }
+        handle_exit_status(exit_status)
```
junyang-zh: `handle_exit_status` is only called at the end of `do_run`. This means that callers can do it just after the call to `do_run`. Adding the handler as a parameter to `do_run` only adds complexity. Please remove the `run_for_test` and `do_run` functions and keep the old way.

junyang-zh (approved):
LGTM

PR #3182, opened 2026-04-28
Fix RISC-V SIGSEGV si_code for page faults

TankTechnology:
## Summary

- Report whether an unhandled page fault address is covered by an existing VMAR mapping.
- Use that VMAR address state on RISC-V to return `SEGV_ACCERR` for mapped permission faults and `SEGV_MAPERR` for unmapped faults.
- Keep x86_64 and LoongArch page-fault `si_code` behavior unchanged while updating the shared `ToFaultSignal` trait signature.
- Add a cross-arch regression test for `SIGSEGV` `si_code` covering unmapped and `PROT_NONE` faults.

## Tests

- `git diff --check`
- `cargo fmt --check`
- `clang-format --dry-run --Werror test/initramfs/src/regression/process/signal/fault_info.c`
- `make -C test/initramfs/src/regression/process/signal fault_info && ./test/initramfs/src/regression/process/signal/fault_info`
- Docker `make check`
- Docker `make run_kernel OSDK_TARGET_ARCH=riscv64 AUTO_TEST=regression INITRAMFS_SKIP_GZIP=1`
  - Fails before reaching this new test, in `/test/hello_world`, after printing `hello world from hello_c!` and `hello world from hello_pie!`.
  - I reproduced the same `/test/hello_world` failure on a clean `upstream/main` worktree at `d00df2b88`, so this appears to be a baseline/environment issue rather than this patch.

### DIFF
--- a/kernel/src/arch/loongarch/signal.rs
+++ b/kernel/src/arch/loongarch/signal.rs
@@ -9,6 +9,7 @@ use ostd::{
 use crate::{
     process::signal::{SignalContext, sig_num::SigNum, signals::fault::FaultSignal},
     thread::exception::ToFaultSignal,
+    vm::vmar::PageFaultAddressState,
 };
 
 impl SignalContext for UserContext {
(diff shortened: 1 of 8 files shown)


TankTechnology:
Closing this for now because it was opened before the implementation plan and verification status were fully discussed and approved. We will revisit the changes locally before submitting a PR again.

PR #3183, opened 2026-04-28
Add spaces in log messages after `:`

lrh2000:
I think we prefer:
```rust
info!("Found device: {:?}", device_type);
```
than:
```rust
info!("Found device:{:?}", device_type);
```

Right?

If so, this PR is generated by:
```bash
sed -i 's|\([^:]\):{|\1: {|' kernel/**/*.rs ostd/**/*.rs osdk/src/**/*.rs
```

I manually go through the changes produced by the above command and remove any that are incorrect (e.g., command lines, Linux-compatible procfs files, line/column numbers, etc.). The remaining changes focus solely on log messages.


tatetian (approved):

PR #3184, opened 2026-04-28
Propagate more `VirtioTransportError`s

lrh2000:
In https://github.com/asterinas/asterinas/pull/3160, it was decided to continue using `unwrap` when communicating with the transport in some places.

> 2. APIs that cannot fail should continue to use `unwrap()`. These include the following: [..]
>   - `VirtioTransport::register_queue_callback()`/`VirtioTransport::register_cfg_callback()`. [..]

However, this may not be reasonable since unexpected errors can occur when communicating with the device through the transport implementation, which is behind a trait.

Propagating errors with `?` in such places is trivial. This PR converts the `unwrap` to error propagation.

The merged `virtio-rng` implementation is fine. No need to fix it. All other devices are fixed in this PR.

### DIFF
--- a/kernel/comps/virtio/src/device/block/device.rs
+++ b/kernel/comps/virtio/src/device/block/device.rs
@@ -270,12 +270,8 @@ impl DeviceInner {
 
         {
             let mut transport = device.transport.lock();
-            transport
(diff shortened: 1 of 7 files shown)


tatetian (approved):

Issue #3181, opened 2026-04-27
`sys_sendfile()` drops data after a short write to a PTY


cqs21:
### Describe the bug

When a shell is running on a pseudoterminal, `cat` may fail to display the full contents of a large file. The file data is intact, but the last part of the output may be silently lost on the terminal.

This issue can be observed in some CI runs, for example <https://github.com/asterinas/asterinas/actions/runs/24954101917/job/73069244092> and <https://github.com/asterinas/asterinas/actions/runs/24978320756/job/73134763494>, and it can also be reproduced locally with a few simple steps. 

The actual bug is in Asterinas `sys_sendfile()`.

In the initramfs environment, `/bin/cat` comes from BusyBox. BusyBox `cat` uses `bb_copyfd_eof()`, which prefers `sendfile(2)` on Linux. When the output fd is a PTY slave, the PTY write path may legally return a positive short write if only part of the requested data can be queued.

Asterinas `sys_sendfile()` currently mishandles that case:

1. It reads a full page-sized chunk from the input file.
2. It performs one write to the output fd.
3. If that write returns a positive short write, `sys_sendfile()` stops the syscall instead of retrying the unwritten tail of the already-read chunk.
4. The remaining bytes of that chunk are therefore silently dropped, while the input file offset has already advanced.

As a result, `sendfile()` can skip ranges in the middle of the file, not just truncate the end.

### To Reproduce

1. Start the VM:

```bash
make run_kernel
```

2. Inside the VM:

```bash
seq 3226 > lines.txt
cat lines.txt
tail lines.txt
```

3. Observe that `cat` may skip some lines in the middle and print `3225` as the last line, while `tail` correctly prints 3226.

### Expected behavior

`cat lines.txt` should print the complete file contents.

PR #3171, opened 2026-04-23
Refactor the ext2 file system implementation

Fischer0522:
## Summary

This PR completely refactors the ext2 implementation, focusing on the correctness and consistency of on-disk state. It is rebased from the page cache refactor (#2953), as it relies on the new APIs introduced there.

### Sparse File Support
Block allocation strategy is now aligned with Linux's ext2, with native sparse file support, improving disk utilization and reducing space overhead.

### Block Allocation
Replaced the legacy delayed allocation strategy (diverged from Linux's ext2) with a cleaner abstraction, `BlockPtrTree`, which manages all block mapping and allocation in a more principled way.

### Inode Lifecycle Management
Inode lifecycle is now managed via `Arc<Inode>` reference counting, ensuring resources are reclaimed when the last reference is dropped. This aligns with POSIX semantics — I/O remains valid as long as any `fd` referencing the file is open.

### Sync Path
Introduced a dedicated sync path to ensure all dirty data is persisted to disk correctly.

### Other Improvements
- **Rollback on error**: Introduced a RAII-style `BlockAllocGuard` that automatically deallocates blocks when an error occurs during allocation.
- **Partial block handling**: Improved handling of partial blocks in `write_at`, `resize`, and `fallocate` to prevent stale data from being exposed to upper layers.
- **Bug fixes**: Fixed `i_blocks` overflow on large files and incorrect directory entry offset (`d_off`) computation.

## Results

The refactored ext2 driver significantly improves conformance test coverage across xfstests, LTP, and gVisor:

| Test Suite | Newly Enabled |
|------------|---------------|
| **xfstests** | 33 more test cases (~60% improvement) |
| **LTP**      | 33 test cases  |
| **gVisor**   | 109 subtests across 9 test suites |

## TODO
- [x] Documentation improvements
- [x] Refactor ktests and regression tests — currently in a messy state
- [ ] Fix timestamps
- [x] Improve the `xattr` implementation

PR #3185, opened 2026-04-29
Implement `O_TMPFILE` support for `open/openat`

YanLien:
## Summary

This PR implements basic `O_TMPFILE` support for `open`/`openat`.

The change adds parsing and validation for `O_TMPFILE`, creates unnamed regular files through the VFS inode API, implements tmpfile creation for ramfs, and tracks whether an unnamed file may later be hard-linked. It also adds regression tests for successful tmpfile creation, invalid flag combinations, directory visibility, `linkat(..., AT_EMPTY_PATH)`, `O_EXCL` unlinkability, and cross-mount linking.

## Changes

- Add `O_TMPFILE` parsing and validation, including writable access requirements and rejection of invalid combinations such as `O_CREAT` or `O_PATH`.
- Extend the VFS inode/path APIs to create unnamed temporary files and track whether they may later receive a hard link.
- Implement ramfs tmpfile support with zero hard links, no directory entry, and `O_EXCL` tmpfiles marked unlinkable.
- Add fs regression tests for tmpfile creation, invalid flags, directory invisibility, `linkat(..., AT_EMPTY_PATH)`, `O_EXCL`, and cross-mount behavior.
                                                                                                                                                                                                   
## Testing

Added fs regression test:

- `test/initramfs/src/regression/fs/tmpfile/tmpfile.c`

The test is wired into the fs regression suite via `run_test.sh` and the fs regression `Makefile`.

### DIFF
--- a/kernel/src/fs/file/file_attr/creation_flags.rs
+++ b/kernel/src/fs/file/file_attr/creation_flags.rs
@@ -3,7 +3,6 @@
 use bitflags::bitflags;
 
 bitflags! {
-    // TODO: Add O_TMPFILE
     pub struct CreationFlags: u32 {
         /// create file if it does not exist
         const O_CREAT = 1 << 6;
@@ -19,5 +18,7 @@ bitflags! {
         const O_NOFOLLOW = 1 << 17;
         /// close on exec
         const O_CLOEXEC = 1 << 19;
+        /// create an unnamed temporary file
+        const O_TMPFILE = 1 << 22;
     }
 }
--- a/kernel/src/fs/file/file_attr/open_args.rs
+++ b/kernel/src/fs/file/file_attr/open_args.rs
@@ -19,7 +19,27 @@ impl OpenArgs {
     pub fn from_flags_and_mode(flags: u32, inode_mode: InodeMode) -> Result<Self> {
         let creation_flags = CreationFlags::from_bits_truncate(flags);
         let status_flags = StatusFlags::from_bits_truncate(flags);
-        if creation_flags.contains(CreationFlags::O_CREAT)
+        let access_mode = AccessMode::from_u32(flags)?;
+
+        // When `O_PATH` is set, all other flags (including `O_TMPFILE`) are
+        // ignored, so the `O_TMPFILE` validations are skipped.
+        // Reference: <https://man7.org/linux/man-pages/man2/open.2.html>.
+        if creation_flags.contains(CreationFlags::O_TMPFILE)
+            && !status_flags.contains(StatusFlags::O_PATH)
+        {
+            if !creation_flags.contains(CreationFlags::O_DIRECTORY) {
+                return_errno_with_message!(Errno::EINVAL, "O_TMPFILE requires O_DIRECTORY");
+            }
+            if !access_mode.is_writable() {
+                return_errno_with_message!(Errno::EINVAL, "O_TMPFILE requires O_RDWR or O_WRONLY");
+            }
+            if creation_flags.contains(CreationFlags::O_CREAT) {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "O_TMPFILE and O_CREAT are mutually exclusive"
+                );
+            }
+        } else if creation_flags.contains(CreationFlags::O_CREAT)
             && creation_flags.contains(CreationFlags::O_DIRECTORY)
         {
             return_errno_with_message!(
@@ -27,7 +47,7 @@ impl OpenArgs {
                 "O_CREAT and O_DIRECTORY cannot be specified together"
             );
         }
-        let access_mode = AccessMode::from_u32(flags)?;
+
         Ok(Self {
             creation_flags,
             status_flags,
@@ -51,5 +71,11 @@ impl OpenArgs {
         !(self.creation_flags.contains(CreationFlags::O_NOFOLLOW)
             || self.creation_flags.contains(CreationFlags::O_CREAT)
                 && self.creation_flags.contains(CreationFlags::O_EXCL))
+    }
+
+    /// Returns whether this is an `O_TMPFILE` open request.
+    pub fn is_tmpfile(&self) -> bool {
+        self.creation_flags.contains(CreationFlags::O_TMPFILE)
+            && !self.status_flags.contains(StatusFlags::O_PATH)
     }
 }
--- a/kernel/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/src/fs/fs_impls/ramfs/fs.rs
@@ -26,7 +26,10 @@ use crate::{
         utils::{CStr256, DirentVisitor},
         vfs::{
             file_system::{FileSystem, FsEventSubscriberStats, SuperBlock},
-            inode::{Extension, FallocMode, FileOps, Inode, Metadata, MknodType, SymbolicLink},
+            inode::{
+                Extension, FallocMode, FileOps, HardLinkability, Inode, Metadata, MknodType,
+                SymbolicLink,
+            },
             path::{is_dot, is_dot_or_dotdot, is_dotdot},
             registry::{FsCreationCtx, FsProperties, FsType},
             xattr::{XattrName, XattrNamespace, XattrSetFlags},
@@ -108,6 +111,7 @@ impl RamFs {
                 this: weak_root.clone(),
                 fs: weak_fs.clone(),
                 container_dev_id: root_dev_id,
+                hard_linkability: HardLinkability::Linkable,
                 extension: Extension::new(),
                 xattr: RamXattr::new(),
             }),
@@ -162,6 +166,10 @@ pub(super) struct RamInode {
     /// Detached inodes such as `memfd` store it directly
     /// because they do not have a valid fs reference.
     container_dev_id: DeviceId,
+    /// Hard linkability.
+    /// All inodes except temporary files are set to linkable
+    /// Linkability of temporary files is specified via [`RamInode::create_tmpfile`]
+    hard_linkability: HardLinkability,
     /// Extensions
     extension: Extension,
     /// Extended attributes
@@ -321,6 +329,12 @@ impl InodeMeta {
             uid,
             gid,
         }
+    }
+
+    pub fn new_tmpfile(mode: InodeMode, uid: Uid, gid: Gid) -> Self {
+        let mut meta = Self::new(mode, uid, gid);
+        meta.nlinks = 0;
+        meta
     }
 
     pub fn resize(&mut self, new_size: usize) {
@@ -504,6 +518,7 @@ impl RamInode {
             this: weak_self.clone(),
             fs: Arc::downgrade(fs),
             container_dev_id: fs.sb.container_dev_id,
+            hard_linkability: HardLinkability::Linkable,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         })
@@ -518,6 +533,28 @@ impl RamInode {
             this: weak_self.clone(),
             fs: Arc::downgrade(fs),
             container_dev_id: fs.sb.container_dev_id,
+            hard_linkability: HardLinkability::Linkable,
+            extension: Extension::new(),
+            xattr: RamXattr::new(),
+        })
+    }
+
+    fn new_tmpfile(
+        fs: &Arc<RamFs>,
+        mode: InodeMode,
+        uid: Uid,
+        gid: Gid,
+        hard_linkability: HardLinkability,
+    ) -> Arc<Self> {
+        Arc::new_cyclic(|weak_self| RamInode {
+            inner: Inner::new_file(),
+            metadata: SpinLock::new(InodeMeta::new_tmpfile(mode, uid, gid)),
+            ino: fs.alloc_id(),
+            typ: InodeType::File,
+            this: weak_self.clone(),
+            fs: Arc::downgrade(fs),
+            container_dev_id: fs.sb.container_dev_id,
+            hard_linkability,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         })
@@ -539,6 +576,7 @@ impl RamInode {
             this: Weak::new(),
             fs: Weak::new(),
             container_dev_id: dev_id,
+            hard_linkability: HardLinkability::Linkable,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         }
@@ -553,6 +591,7 @@ impl RamInode {
             this: weak_self.clone(),
             fs: Arc::downgrade(fs),
             container_dev_id: fs.sb.container_dev_id,
+            hard_linkability: HardLinkability::Linkable,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         })
@@ -579,6 +618,7 @@ impl RamInode {
             this: weak_self.clone(),
             fs: Arc::downgrade(fs),
             container_dev_id: fs.sb.container_dev_id,
+            hard_linkability: HardLinkability::Linkable,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         })
@@ -593,6 +633,7 @@ impl RamInode {
             this: weak_self.clone(),
             fs: Arc::downgrade(fs),
             container_dev_id: fs.sb.container_dev_id,
+            hard_linkability: HardLinkability::Linkable,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         })
@@ -607,6 +648,7 @@ impl RamInode {
             this: weak_self.clone(),
             fs: Arc::downgrade(fs),
             container_dev_id: fs.sb.container_dev_id,
+            hard_linkability: HardLinkability::Linkable,
             extension: Extension::new(),
             xattr: RamXattr::new(),
         })
@@ -911,6 +953,25 @@ impl Inode for RamInode {
         Ok(new_inode)
     }
 
+    fn create_tmpfile(
+        &self,
+        mode: InodeMode,
+        hard_linkability: HardLinkability,
+    ) -> Result<Arc<dyn Inode>> {
+        if self.typ != InodeType::Dir {
+            return_errno_with_message!(Errno::ENOTDIR, "self is not dir");
+        }
+
+        let fs = self.fs.upgrade().unwrap();
+        Ok(RamInode::new_tmpfile(
+            &fs,
+            mode,
+            Uid::new_root(),
+            Gid::new_root(),
+            hard_linkability,
+        ))
+    }
+
     fn link(&self, old: &Arc<dyn Inode>, name: &str) -> Result<()> {
         if !Arc::ptr_eq(&self.fs(), &old.fs()) {
             return_errno_with_message!(Errno::EXDEV, "not same fs");
@@ -925,24 +986,29 @@ impl Inode for RamInode {
         if old.typ == InodeType::Dir {
             return_errno_with_message!(Errno::EPERM, "old is a dir");
         }
+        if old.hard_linkability == HardLinkability::Unlinkable {
+            return_errno_with_message!(Errno::ENOENT, "tmpfile is not linkable");
+        }
 
         let mut self_dir = self.inner.as_direntry().unwrap().write();
         if self_dir.contains_entry(name) {
             return_errno_with_message!(Errno::EEXIST, "entry exist");
         }
         self_dir.append_entry(name, old.this.upgrade().unwrap());
+        let now = now();
+
+        // An `O_TMPFILE` inode starts with zero links. Bump the link count
+        // before exposing the new entry to concurrent lookup or unlink.
+        let mut old_meta = old.metadata.lock();
+        old_meta.inc_nlinks();
+        old_meta.set_ctime(now);
+        drop(old_meta);
         drop(self_dir);
 
-        let now = now();
         let mut self_meta = self.metadata.lock();
         self_meta.set_mtime(now);
         self_meta.set_ctime(now);
         self_meta.inc_size();
-        drop(self_meta);
-
-        let mut old_meta = old.metadata.lock();
-        old_meta.inc_nlinks();
-        old_meta.set_ctime(now);
 
         Ok(())
     }
--- a/kernel/src/fs/vfs/fs_apis/inode.rs
+++ b/kernel/src/fs/vfs/fs_apis/inode.rs
@@ -119,6 +119,19 @@ pub struct Metadata {
     pub self_dev_id: Option<DeviceId>,
 }
 
+/// Describes whether an inode may get new hard links.
+///
+/// This does not control symbolic links. A symbolic link is a separate inode
+/// whose target is a pathname string, so creating one does not link the target
+/// inode directly.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum HardLinkability {
+    /// Allows creating hard links to the inode.
+    Linkable,
+    /// Prevents creating hard links to the inode.
+    Unlinkable,
+}
+
 impl Metadata {
     pub fn new_dir(ino: u64, mode: InodeMode, blk_size: usize, container_dev_id: DeviceId) -> Self {
         let now = RealTimeCoarseClock::get().read_time();
@@ -363,6 +376,14 @@ pub trait Inode: Any + FileOps + Send + Sync {
 
     fn create(&self, name: &str, type_: InodeType, mode: InodeMode) -> Result<Arc<dyn Inode>> {
         Err(Error::new(Errno::ENOTDIR))
+    }
+
+    fn create_tmpfile(
+        &self,
+        mode: InodeMode,
+        hard_linkability: HardLinkability,
+    ) -> Result<Arc<dyn Inode>> {
+        Err(Error::new(Errno::EOPNOTSUPP))
     }
 
     fn mknod(&self, name: &str, mode: InodeMode, type_: MknodType) -> Result<Arc<dyn Inode>> {
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -22,7 +22,7 @@ use crate::{
         pseudofs::NsInode,
         vfs::{
             file_system::{FileSystem, FsFlags},
-            inode::{Inode, Metadata, MknodType},
+            inode::{HardLinkability, Inode, Metadata, MknodType},
             xattr::{XattrName, XattrNamespace, XattrSetFlags},
         },
     },
@@ -74,6 +74,28 @@ impl Path {
             .as_dir_dentry_or_err()?
             .create(name, type_, mode)?;
         Ok(Self::new(self.mount.clone(), new_child_dentry))
+    }
+
+    /// Creates a new `Path` to represent an unnamed temporary file.
+    ///
+    /// The returned inode has no directory entry and is invisible to `readdir`.
+    /// If created hard-linkable, it may later be given a name via a link
+    /// operation; otherwise it can never be linked.
+    pub fn create_tmpfile(
+        &self,
+        mode: InodeMode,
+        hard_linkability: HardLinkability,
+    ) -> Result<Self> {
+        if self
+            .inode()
+            .check_permission(Permission::MAY_WRITE)
+            .is_err()
+        {
+            return_errno!(Errno::EACCES);
+        }
+        let tmp_inode = self.inode().create_tmpfile(mode, hard_linkability)?;
+        let tmp_dentry = Dentry::new_anonymous(tmp_inode, self.dentry.clone());
+        Ok(Self::new(self.mount.clone(), tmp_dentry))
     }
 
     /// Creates a new pseudo `Path`.
--- a/kernel/src/syscall/open.rs
+++ b/kernel/src/syscall/open.rs
@@ -9,7 +9,10 @@ use crate::{
             StatusFlags,
             file_table::{FdFlags, RawFileDesc},
         },
-        vfs::path::{AT_FDCWD, EmptyPathStr, FsPath, LookupResult, PathResolver},
+        vfs::{
+            inode::HardLinkability,
+            path::{AT_FDCWD, EmptyPathStr, FsPath, LookupResult, PathResolver},
+        },
     },
     prelude::*,
     syscall::constants::MAX_FILENAME_LEN,
@@ -82,6 +85,10 @@ fn do_open(
 ) -> Result<Arc<dyn FileLike>> {
     let open_args = OpenArgs::from_flags_and_mode(flags, mode)?;
 
+    if open_args.is_tmpfile() {
+        return do_open_tmpfile(path_resolver, fs_path, &open_args);
+    }
+
     let lookup_res = if open_args.follow_tail_link() {
         path_resolver.lookup_unresolved(fs_path)?
     } else {
@@ -118,4 +125,38 @@ fn do_open(
     };
 
     Ok(file_handle)
+}
+
+fn do_open_tmpfile(
+    path_resolver: &PathResolver,
+    fs_path: &FsPath,
+    open_args: &OpenArgs,
+) -> Result<Arc<dyn FileLike>> {
+    let dir_path = if open_args.follow_tail_link() {
+        path_resolver.lookup(fs_path)?
+    } else {
+        path_resolver.lookup_no_follow(fs_path)?
+    };
+    if dir_path.type_() != InodeType::Dir {
+        return_errno_with_message!(
+            Errno::ENOTDIR,
+            "O_TMPFILE requires the path to be a directory"
+        );
+    }
+
+    // `O_EXCL` with `O_TMPFILE` is allowed by Linux, but it prevents the tmpfile
+    // from being linked later by `linkat(..., AT_EMPTY_PATH)`.
+    // Reference: <https://man7.org/linux/man-pages/man2/open.2.html>.
+    let hard_linkability = if open_args.creation_flags.contains(CreationFlags::O_EXCL) {
+        HardLinkability::Unlinkable
+    } else {
+        HardLinkability::Linkable
+    };
+    let tmpfile_path = dir_path.create_tmpfile(open_args.inode_mode, hard_linkability)?;
+
+    Ok(Arc::new(InodeHandle::new_unchecked_access(
+        tmpfile_path,
+        open_args.access_mode,
+        open_args.status_flags,
+    )?))
 }
--- a/test/initramfs/src/regression/fs/run_test.sh
+++ b/test/initramfs/src/regression/fs/run_test.sh
@@ -130,4 +130,6 @@ echo "All mount bind file test passed."
 
 ./symlink/symlink
 
+./tmpfile/tmpfile
+
 ./utimensat/utimensat
--- a/test/initramfs/src/regression/fs/tmpfile/tmpfile.c
+++ b/test/initramfs/src/regression/fs/tmpfile/tmpfile.c
@@ -1,0 +1,292 @@
+/* SPDX-License-Identifier: MPL-2.0 */
+
+#define _GNU_SOURCE
+
+#include <dirent.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "../../common/test.h"
+
+#define TEST_DIR "/tmp/aster_tmpfile_test"
+#define CROSS_LINK_DIR "/ext2"
+#define CROSS_LINK_NAME "cross_mount_tmpfile"
+#define LINKED_NAME "linked_file"
+#define LINKED_SECOND_NAME "linked_second_file"
+#define LINKED_O_EXCL_NAME "linked_o_excl_file"
+#define SYMLINK_NAME "tmpfile_symlink"
+#define DATA "hello from tmpfile"
+#define DATA_LEN (sizeof(DATA) - 1)
+
+#define RAW_O_TMPFILE 020000000
+
+/* O_TMPFILE may not be defined on older glibc. */
+#ifndef __O_TMPFILE
+#define __O_TMPFILE RAW_O_TMPFILE
+#endif
+
+#ifndef O_TMPFILE
+#define O_TMPFILE (__O_TMPFILE | O_DIRECTORY)
+#endif
+
+#ifndef AT_EMPTY_PATH
+#define AT_EMPTY_PATH 0x1000
+#endif
+
+static void cleanup_test_files(void)
+{
+	unlink(TEST_DIR "/" LINKED_NAME);
+	unlink(TEST_DIR "/" LINKED_SECOND_NAME);
+	unlink(TEST_DIR "/" LINKED_O_EXCL_NAME);
+	unlink(TEST_DIR "/" SYMLINK_NAME);
+	unlink(CROSS_LINK_DIR "/" CROSS_LINK_NAME);
+	rmdir(TEST_DIR);
+}
+
+static int timespec_equal(struct timespec left, struct timespec right)
+{
+	return left.tv_sec == right.tv_sec && left.tv_nsec == right.tv_nsec;
+}
+
+static int dir_is_unavailable_or_same_mount(const char *source_path,
+					    const char *target_path)
+{
+	struct stat source_stat;
+	struct stat target_stat;
+
+	if (stat(source_path, &source_stat) < 0 ||
+	    stat(target_path, &target_stat) < 0) {
+		return 1;
+	}
+
+	return source_stat.st_dev == target_stat.st_dev;
+}
+
+FN_SETUP(prepare)
+{
+	cleanup_test_files();
+	CHECK(mkdir(TEST_DIR, 0755));
+}
+END_SETUP()
+
+FN_TEST(tmpfile_open_succeeds)
+{
+	int fd;
+
+	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_RDWR, 0666));
+	TEST_SUCC(close(fd));
+}
+END_TEST()
+
+FN_TEST(tmpfile_open_write_only_succeeds)
+{
+	int fd;
+
+	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_WRONLY, 0666));
+	TEST_SUCC(close(fd));
+}
+END_TEST()
+
+FN_TEST(tmpfile_open_read_only_returns_einval)
+{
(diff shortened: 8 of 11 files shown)


tatetian:
Welcome to the Asterinas community! Since you are a first-time contributor, I'd like to share a few pointers:

1. **Read the [Coding Guidelines](https://asterinas.github.io/book/to-contribute/coding-guidelines/index.html).** They cover the preferred style and common pitfalls in Asterinas development — well worth a read before your first contribution.
2. **Do a thorough self-review, especially for AI-assisted code.** We welcome AI-assisted contributions, but human authors are responsible for the code. As in most open-source communities, contributors outnumber reviewers, so we encourage contributors to polish their code as much as possible before requesting review.
3. **Consider using the [`pr-review` skill](https://github.com/tatetian/asterinas/blob/pr-review-skill/.agents/skills/pr-review/spec.md).** It isn't merged into the upstream repo yet, but it's tuned for the specific needs of Asterinas and should give you useful feedback.

I've marked this PR as "Draft". Once you've completed the self-review, feel free to mark it "Ready for Review". Thanks!

YanLien:
This CI failure appears to be a false negative in the log-checking logic, not an actual LTP failure.

In the CI log for `conformance-test (ltp, ext2-microvm)`, the LTP run itself completed successfully:

- `INFO: ltp-pan reported all tests PASS`
- `Total Tests: 430`
- `Total Failures: 0`
- `All conformance tests passed.` is printed by the guest test script

The job failed afterward because the Makefile checks `qemu.log` with a line-anchored pattern: `grep -q "^All conformance tests passed."` But in this run, the serial output was concatenated with the previous line: `Machine ArchitectuAll conformance tests passed.` So the success marker exists, but not at the beginning of a line, causing the outer CI wrapper to report `Conformance test failed`.

I think this is a CI/log parsing false negative rather than a regression in this PR.

tatetian:
> This CI failure appears to be a false negative in the log-checking logic, not an actual LTP failure.
> 
> In the CI log for `conformance-test (ltp, ext2-microvm)`, the LTP run itself completed successfully:
> 
> * `INFO: ltp-pan reported all tests PASS`
> * `Total Tests: 430`
> * `Total Failures: 0`
> * `All conformance tests passed.` is printed by the guest test script
> 
> The job failed afterward because the Makefile checks `qemu.log` with a line-anchored pattern: `grep -q "^All conformance tests passed."` But in this run, the serial output was concatenated with the previous line: `Machine ArchitectuAll conformance tests passed.` So the success marker exists, but not at the beginning of a line, causing the outer CI wrapper to report `Conformance test failed`.
> 
> I think this is a CI/log parsing false negative rather than a regression in this PR.

No worry, this is due to a bug described by #3181

YanLien:
Hi @tatetian, could you help take a look at this PR, or help loop in the right reviewer? Thanks!

tatetian:
> Hi @tatetian, could you help take a look at this PR, or help loop in the right reviewer? Thanks!

I shimmed through the code. It looks quite promising.

Before raising any minor issues, I would like to see this PR rebased to #3235 (~~once the latter one gets merged. Your comments on #3235 are also welcome.~~ Merged). One major problem with this PR is that temporary file are created via `Dentry::new_pseudo`. This does not make sense as temporary files are NOT pseudo files. #3235 adds a new dentry variant called `Anonymous`, which better suites temporary files.

YanLien:
> > Hi @tatetian, could you help take a look at this PR, or help loop in the right reviewer? Thanks!
> 
> I shimmed through the code. It looks quite promising.
> 
> Before raising any minor issues, I would like to see this PR rebased to #3235 (~once the latter one gets merged. Your comments on #3235 are also welcome.~ Merged). One major problem with this PR is that temporary file are created via `Dentry::new_pseudo`. This does not make sense as temporary files are NOT pseudo files. #3235 adds a new dentry variant called `Anonymous`, which better suites temporary files.

Thanks for the review.

Rebased onto main after #3235.

O_TMPFILE no longer uses Dentry::new_pseudo. It now creates a Dentry::Anonymous with the containing directory as parent, so the tmpfile is a real filesystem inode but remains unreachable by normal lookup/readdir.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -925,6 +960,9 @@ impl Inode for RamInode {
         if old.typ == InodeType::Dir {
             return_errno_with_message!(Errno::EPERM, "old is a dir");
         }
+        if old.metadata.lock().hard_linkability() == HardLinkability::Unlinkable {
+            return_errno_with_message!(Errno::ENOENT, "tmpfile is not linkable");
+        }
 
         let mut self_dir = self.inner.as_direntry().unwrap().write();
         if self_dir.contains_entry(name) {
```
tatetian: **`link` makes the new directory entry visible before incrementing `nlinks`, underflowing a tmpfile's link count** (blocker): `RamInode::link` calls `self_dir.append_entry(name, ...)` and `drop(self_dir)` (lines 971-972) before `old.metadata.lock().inc_nlinks()` runs (line 982). For a normal `create` this is harmless because the inode is born with `nlinks == 1`, but an `O_TMPFILE` inode is created with `nlinks == 0` (`create_tmpfile`, line 942). So between line 972 and line 982 the tmpfile is fully reachable via `lookup`/`readdir` under its new name while `nlinks` is still 0. A concurrent `unlink` of that name takes `target.metadata.lock()` and calls `dec_nlinks()` (line 1014), which hits `debug_assert!(self.nlinks > 0)` at line 371 — a userspace-triggerable panic in debug builds — and underflows `nlinks` to `usize::MAX` in release builds, permanently corrupting the link count and leaking the inode.

Fix: increment `nlinks` while still holding the directory write lock, so the entry is never observable with `nlinks == 0`. This introduces a directory-lock-then-inode-metadata-lock ordering; no existing path acquires those in the reverse order, so it is deadlock-free.

```rust
        self_dir.append_entry(name, old.this.upgrade().unwrap());

        // Bump `nlinks` before releasing the directory lock, so the new entry
        // is never observable while the target still has `nlinks == 0`.
        let now = now();
        let mut old_meta = old.metadata.lock();
        old_meta.inc_nlinks();
        old_meta.set_ctime(now);
        drop(old_meta);
        drop(self_dir);

        let mut self_meta = self.metadata.lock();
        self_meta.set_mtime(now);
        self_meta.set_ctime(now);
        self_meta.inc_size();
```

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -911,6 +925,27 @@ impl Inode for RamInode {
...
+        }
+
+        let fs = self.fs.upgrade().unwrap();
+        let new_inode = RamInode::new_file(&fs, mode, Uid::new_root(), Gid::new_root());
+
+        {
+            let mut new_inode_meta = new_inode.metadata.lock();
+            new_inode_meta.nlinks = 0;
+            new_inode_meta.set_hard_linkability(hard_linkability);
+        }
```
tatetian: **`create_tmpfile` overwrites freshly-constructed metadata instead of building it correctly** (minor): `RamInode::new_file` builds `InodeMeta` with `nlinks: 1` and `hard_linkability: Linkable`; `create_tmpfile` then re-locks `metadata` to set `nlinks = 0` and the requested linkability, poking the `nlinks` field directly rather than through an accessor (every other site mutates it via `inc_nlinks`/`dec_nlinks`). This is race-free only because the inode is still local to `create_tmpfile` and unreachable by other threads — a non-obvious invariant. Prefer a dedicated constructor that establishes the tmpfile invariant (`nlinks == 0` + linkability) at construction time, mirroring how `InodeMeta::new_dir` exists to encode the directory `nlinks` invariant. This drops the transient `nlinks == 1` state and the second lock acquisition:

```rust
// InodeMeta
fn new_tmpfile(mode: InodeMode, uid: Uid, gid: Gid, hard_linkability: HardLinkability) -> Self {
    let mut meta = Self::new(mode, uid, gid);
    meta.nlinks = 0;
    meta.hard_linkability = hard_linkability;
    meta
}
```

with a matching `RamInode::new_tmpfile` that `create_tmpfile` calls directly.

kernel/src/syscall/open.rs
```diff
@@ -119,3 +126,34 @@ fn do_open(
...
+        path_resolver.lookup(fs_path)?
+    } else {
+        path_resolver.lookup_no_follow(fs_path)?
+    };
+    if dir_path.type_() != InodeType::Dir {
+        return_errno_with_message!(
+            Errno::ENOTDIR,
+            "O_TMPFILE requires the path to be a directory"
+        );
+    }
```
tatetian, line 145: **`O_TMPFILE` + `O_NOFOLLOW` on a final symlink returns `ENOTDIR` instead of `ELOOP`** (significant): With `O_NOFOLLOW`, `do_open_tmpfile` resolves via `lookup_no_follow`, so a path whose final component is a symlink yields the symlink inode itself; the `dir_path.type_() != InodeType::Dir` check then returns `ENOTDIR`. Asterinas's own regular open path handles the identical case in `Path::open` (`kernel/src/fs/vfs/path/mod.rs:151-156`) by returning `ELOOP` ("the file is a symlink") *before* the non-directory check — which is also Linux's behavior for a final-component symlink under `O_NOFOLLOW` (see `openat(2)`). The new test `tmpfile_open_symlink_with_o_nofollow_returns_enotdir` (`tmpfile.c:104-117`, assertion at 110-112) bakes in the divergent `ENOTDIR`; since the regression suite is also built and run for the `linux` platform (see `test/initramfs/Makefile`), that test will fail the Linux baseline run.

Fix: return `ELOOP` for a final symlink before the directory-type check, then update the test to expect `ELOOP`:

```rust
    if dir_path.type_() == InodeType::SymLink {
        return_errno_with_message!(Errno::ELOOP, "O_TMPFILE with O_NOFOLLOW hit a final symlink");
    }
    if dir_path.type_() != InodeType::Dir {
        return_errno_with_message!(Errno::ENOTDIR, "O_TMPFILE requires the path to be a directory");
    }
```

kernel/src/fs/file/file_attr/open_args.rs
```diff
@@ -19,15 +19,39 @@ impl OpenArgs {
...
+            }
+            // O_EXCL with O_TMPFILE is allowed by Linux, but it prevents
+            // the tmpfile from being linked later by linkat(..., AT_EMPTY_PATH).
+            // Reference: <https://man7.org/linux/man-pages/man2/open.2.html>.
+            if status_flags.contains(StatusFlags::O_PATH) {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "O_TMPFILE and O_PATH are mutually exclusive"
+                );
+            }
```
tatetian: **Misplaced comment makes the `O_PATH` rejection confusing, and the rejection diverges from Linux** (significant): The comment at lines 37-39 explains `O_EXCL` behavior ("O_EXCL with O_TMPFILE is allowed by Linux, but it prevents the tmpfile from being linked later..."), but it sits directly above the `O_PATH` check at lines 40-45. A reader naturally attributes it to that branch, where it makes no sense — it describes `O_EXCL`, not `O_PATH`, and says its subject "is allowed" while the code immediately below *rejects* it. The actual `O_EXCL`→`HardLinkability` mapping lives in `do_open_tmpfile` (`open.rs:147-151`). Separately, rejecting `O_TMPFILE | O_PATH` with `EINVAL` diverges from Linux: per `open(2)`, `O_PATH` causes the other open flags (including `__O_TMPFILE`) to be ignored rather than producing an error, so `open(dir, O_TMPFILE|O_PATH|O_RDWR)` succeeds there as a plain `O_PATH` descriptor.

Fix: move the `O_EXCL` commentary down to `do_open_tmpfile` next to the `HardLinkability` decision, and either drop the `O_PATH` rejection (to match Linux) or replace the comment here with one that justifies the intentional divergence:

```rust
-            // O_EXCL with O_TMPFILE is allowed by Linux, but it prevents
-            // the tmpfile from being linked later by linkat(..., AT_EMPTY_PATH).
-            // Reference: <https://man7.org/linux/man-pages/man2/open.2.html>.
+            // Unlike Linux, which silently ignores O_TMPFILE when O_PATH is
+            // set, Asterinas rejects the combination outright.
             if status_flags.contains(StatusFlags::O_PATH) {
```

kernel/src/fs/file/file_attr/creation_flags.rs
```diff
@@ -19,5 +18,7 @@ bitflags! {
         const O_NOFOLLOW = 1 << 17;
         /// close on exec
         const O_CLOEXEC = 1 << 19;
+        /// creates an unnamed temporary file.
+        const O_TMPFILE = 1 << 22;
```
tatetian: **`O_TMPFILE` constant is the `__O_TMPFILE` bit alone; the doc should say so** (minor): `O_TMPFILE = 1 << 22` (`020000000`) is what Linux calls `__O_TMPFILE` — the internal bit. The userspace `O_TMPFILE` macro is `__O_TMPFILE | O_DIRECTORY` (the PR's own test header defines it exactly that way, `tmpfile.c:25-30`). The code correctly re-checks `O_DIRECTORY` separately in `open_args.rs`, but a reader who sees the constant named `O_TMPFILE` could wrongly assume `CreationFlags::O_TMPFILE` alone is a complete tmpfile request. The doc comment also breaks the file's style — every sibling flag uses a lowercase fragment with no trailing period. Tighten the doc to fix both:

```rust
-        /// creates an unnamed temporary file.
+        /// create an unnamed temporary file (this is the `__O_TMPFILE` bit; a
+        /// complete `O_TMPFILE` request also carries `O_DIRECTORY`)
         const O_TMPFILE = 1 << 22;
```

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -365,6 +378,20 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
+    /// directory iteration. If `hard_linkability` is `HardLinkability::Linkable`,
+    /// the inode may be linked explicitly later; otherwise, it must remain
+    /// unlinkable.
+    fn create_tmpfile(
+        &self,
+        mode: InodeMode,
+        hard_linkability: HardLinkability,
+    ) -> Result<Arc<dyn Inode>> {
+        Err(Error::new(Errno::EOPNOTSUPP))
+    }
```
tatetian: **`create_tmpfile`'s default-impl errno differs from its siblings without explanation** (minor): The default `create_tmpfile` returns `EOPNOTSUPP`, whereas the sibling directory-mutating defaults in this trait (`create`, `mknod`, `link`, ...) return `ENOTDIR`. `EOPNOTSUPP` is the *right* choice here — unlike `create`/`mknod`, whose default models "this inode is not a directory", `create_tmpfile`'s default is hit on a genuine directory whose filesystem simply does not support `O_TMPFILE` (e.g. ext2), and `EOPNOTSUPP` is the conventional errno for an unsupported operation. But the divergence is silent, inviting a future reader to "fix" it to `ENOTDIR` for consistency and thereby return the wrong errno for `O_TMPFILE` on an unsupporting filesystem. Document the contract in the doc comment:

```rust
    /// Creates an unnamed regular file in this directory.
    ///
    /// ... (existing text) ...
    ///
    /// A filesystem that supports directories but not `O_TMPFILE` returns
    /// `EOPNOTSUPP` via this default implementation.
```

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -76,6 +76,28 @@ impl Path {
         Ok(Self::new(self.mount.clone(), new_child_dentry))
     }
 
+    /// Creates an unnamed temporary file in the directory represented by this path.
+    ///
+    /// The returned inode has no directory entry and is invisible to `readdir`.
+    /// It can get a hard link later via `linkat` with `AT_EMPTY_PATH` only if
+    /// it is created as hard-linkable.
```
tatetian: **`Path::create_tmpfile` doc references syscall-layer concepts and omits its error contract** (minor): The doc says the inode "can get a hard link later via `linkat` with `AT_EMPTY_PATH`". `linkat` and `AT_EMPTY_PATH` are syscall-layer concepts; at the VFS `Path` layer a later hard link is established through the `Path`/`Inode` link operation, so the doc should describe it in layer-neutral terms. The doc also omits that the method returns `EACCES` when the directory lacks write permission (lines 89-95). Suggest:

```rust
    /// Creates an unnamed temporary file in the directory represented by this path.
    ///
    /// The returned inode has no directory entry and is invisible to `readdir`.
    /// If created hard-linkable, it may later be given a name via a link
    /// operation; otherwise it can never be linked.
    ///
    /// Returns `EACCES` if the directory is not writable.
```

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
tatetian, line 1: **Validation branches added by this PR are untested** (significant): The new test covers `O_RDONLY`→`EINVAL` and `O_CREAT`→`EINVAL`, but misses three branches:

- **`O_TMPFILE` without `O_DIRECTORY` → `EINVAL`** (`open_args.rs:25-27`) — the most important gap. Every test uses the `O_TMPFILE` macro, which bakes in `O_DIRECTORY`, so this rejection is never exercised. Linux rejects this combination too (`openat(2)` documents `O_TMPFILE` as requiring `O_DIRECTORY`), so it is testable on both platforms: `TEST_ERRNO(open(TEST_DIR, __O_TMPFILE | O_RDWR, 0666), EINVAL)`. Hoist the `__O_TMPFILE` fallback `#define` (currently nested inside the `#ifndef O_TMPFILE` block) so it is always available.
- **`O_TMPFILE | O_PATH` → `EINVAL`** (`open_args.rs:40-45`) and **`O_TMPFILE` on a non-ramfs filesystem → `EOPNOTSUPP`** (the `inode.rs` default impl; ext2 does not override it). Both encode Asterinas-specific behavior — Linux instead succeeds — so a test for either must be guarded with `#ifdef __asterinas__` (see `test/initramfs/Makefile`). Worth adding for the O_PATH-rejection and ramfs-only contracts, but lower priority than the first item, and the O_PATH test depends on whether that rejection is kept (see the `open_args.rs` comment).

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
```diff
@@ -0,0 +1,313 @@
...
+	if (dirfd < 0) {
+		goto out;
+	}
+
+	if (TEST_SUCC(linkat(fd, "", dirfd, LINKED_NAME, AT_EMPTY_PATH)) < 0) {
+		goto out;
+	}
+
+	struct stat stat_after_link;
+	TEST_RES(fstat(fd, &stat_after_link), stat_after_link.st_nlink == 1);
```
tatetian, line 223: **`tmpfile_write_and_linkat` does not exercise the tmpfile's own fd or post-link normalcy** (minor): The test writes through the tmpfile fd, links it, then opens the *linked path* to read back — it never reads back through the original (still-anonymous) fd, which is the core "an `O_TMPFILE` fd is a fully usable file before it has any name" guarantee. It also never confirms the file behaves as an ordinary file after linking. Add, before the `linkat` call, an `lseek(fd, 0, SEEK_SET)` + `read(fd, ...)` that verifies the written content; and after linking, confirm the file is now visible in `readdir` of `TEST_DIR` and accepts a second hard link (`st_nlink` becomes 2).

tatetian (commented):
The following comments were generated by my coding agent using the `pr-review` skill. I skimmed some (but not all) of the comments, which make sense to me. Take the comments with a grain of salt. Either address or rebuttal the comments, but not ignore. Thanks!

---

This PR adds a clean, well-scoped `O_TMPFILE` implementation: flag parsing and validation live in `OpenArgs`, the syscall path forks early into a dedicated `do_open_tmpfile`, and ramfs gains a `create_tmpfile` inode method. The `HardLinkability` named enum is good API design (far clearer at call sites than a bool), the stale `#[expect(dead_code)]` on `Dentry::new_anonymous` is correctly retired, and the C regression test is reasonably thorough.

One **blocker** should be fixed before merge: `RamInode::link` makes the new directory entry observable before it increments `nlinks`. Because an `O_TMPFILE` inode starts at `nlinks == 0`, a concurrent `unlink` racing a `linkat` can trip `dec_nlinks`'s `debug_assert!` (a userspace-triggerable kernel panic in debug builds) or underflow the link count in release builds.

Two **significant** items also warrant attention. First, `do_open_tmpfile` returns `ENOTDIR` where both Asterinas's own `Path::open` and Linux return `ELOOP` for a final-symlink `O_NOFOLLOW` open; the bundled test bakes in the divergent errno, so it will fail the Linux-platform regression build. Second, a comment in `open_args.rs` that documents `O_EXCL` behavior is misplaced above an unrelated `O_PATH` check, and the `O_TMPFILE | O_PATH` rejection itself diverges from Linux. The test suite also leaves the newly added `O_TMPFILE`-without-`O_DIRECTORY` validation branch entirely uncovered.

Structurally, consider giving ramfs a dedicated tmpfile constructor so the `nlinks == 0` + linkability invariant is established at construction time rather than patched in after `new_file`.

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
```diff
@@ -0,0 +1,313 @@
...
+
+out:
+	if (dirfd >= 0) {
+		close(dirfd);
+	}
+	if (fd >= 0) {
+		close(fd);
+	}
+}
+END_TEST()
```
lrh2000, line 262: Please remove all error handling from regression tests. For example, this test should be:
```c
FN_TEST(tmpfile_open_with_o_excl_cannot_be_linked)
{
	int fd;
	int dirfd;

	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_RDWR | O_EXCL, 0666));
	dirfd = TEST_SUCC(open(TEST_DIR, O_RDONLY | O_DIRECTORY));

	TEST_ERRNO(linkat(fd, "", dirfd, LINKED_O_EXCL_NAME, AT_EMPTY_PATH),
		   ENOENT);

	TEST_SUCC(close(dirfd));
	TEST_SUCC(close(fd));
}
END_TEST()
```

There should be no `if (...) goto`, `if (fd ..) close(fd)`, etc.

YanLien: Cleaned all of those up. The `goto out` / `if (fd < 0)` / cleanup-on-error patterns are gone everywhere; each test now just runs straight down with `TEST_SUCC` / `TEST_ERRNO` / `TEST_RES`.

kernel/src/fs/file/file_attr/creation_flags.rs
YanLien: Done. I fixed the doc style and kept it as a short one-liner to match the surrounding flags.

kernel/src/fs/fs_impls/ramfs/fs.rs
YanLien: Done. I added `InodeMeta::new_tmpfile` and `RamInode::new_tmpfile`, and `create_tmpfile` now delegates to that path.

I also removed `set_hard_linkability`, since the metadata is now built correctly from the start.

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
YanLien: Done. I added all three checks to `tmpfile_write_and_linkat`: read back through the original fd before `linkat`, verify the linked file appears in `readdir`, and create a second hard link to confirm `st_nlink == 2`.

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
YanLien: Added all three tests. I used `RAW_O_TMPFILE` for the no-`O_DIRECTORY` case so the test really hits that branch, updated the `O_PATH` case to match the Linux-compatible behavior, and kept the ext2 `EOPNOTSUPP` case under `#ifdef __asterinas__`.

kernel/src/fs/file/file_attr/open_args.rs
```diff
@@ -52,4 +72,10 @@ impl OpenArgs {
             || self.creation_flags.contains(CreationFlags::O_CREAT)
                 && self.creation_flags.contains(CreationFlags::O_EXCL))
     }
+
+    /// Returns whether this is an O_TMPFILE open request.
```
lrh2000:
```suggestion
    /// Returns whether this is an `O_TMPFILE` open request.
```

kernel/src/syscall/open.rs
```diff
@@ -119,3 +126,43 @@ fn do_open(
...
+    if dir_path.type_() != InodeType::Dir {
+        return_errno_with_message!(
+            Errno::ENOTDIR,
+            "O_TMPFILE requires the path to be a directory"
+        );
+    }
+
+    // O_EXCL with O_TMPFILE is allowed by Linux, but it prevents the tmpfile
+    // from being linked later by linkat(..., AT_EMPTY_PATH).
```
lrh2000:
```suggestion
    // `O_EXCL` with `O_TMPFILE` is allowed by Linux, but it prevents the tmpfile
    // from being linked later by `linkat(..., AT_EMPTY_PATH)`.
```

kernel/src/fs/vfs/path/mod.rs
YanLien: Reworded at `path/mod.rs` to talk about a "link operation" instead of `linkat` / `AT_EMPTY_PATH`.

Skipped the `EACCES` line though. The sibling `Path` methods (`create`, `mknod`, `link`, `unlink`) all return `EACCES` from `check_permission(MAY_WRITE)` and none of them document it, so adding it only here creates a different inconsistency.

kernel/src/fs/fs_impls/ramfs/fs.rs
YanLien: Fixed at `kernel/src/fs/fs_impls/ramfs/fs.rs` — `inc_nlinks()` now runs while the directory write lock is still held, so the entry is never observable with `nlinks == 0`.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -365,6 +378,26 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
+    /// The returned inode must start with zero hard links and must not appear in
+    /// directory iteration. If `hard_linkability` is `HardLinkability::Linkable`,
+    /// the inode may be linked explicitly later; otherwise, it must remain
+    /// unlinkable.
+    ///
+    /// Unlike sibling directory-mutating defaults (`create`, `mknod`, `link`)
+    /// which return `ENOTDIR`, this default returns `EOPNOTSUPP`: callers are
+    /// expected to verify the inode is a directory before invoking, so the only
+    /// reason to hit the default is a filesystem (e.g. ext2) that supports
+    /// directories but not `O_TMPFILE`.
```
lrh2000:
```suggestion
    /// reason to hit the default is a filesystem that supports directories but
    /// not `O_TMPFILE`.
```

It seems that ext2 does support `O_TMPFILE`. Just our current implementation does not support it.

kernel/src/syscall/open.rs
YanLien: Fixed. I added the final-symlink check before the non-directory check, so `O_TMPFILE | O_NOFOLLOW` now returns `ELOOP` instead of `ENOTDIR`.

I also renamed the test and updated it to expect `ELOOP`, which keeps the behavior aligned with `Path::open`, Linux, and the Linux baseline run.

kernel/src/fs/vfs/fs_apis/inode.rs
YanLien: Done. I added a doc note explaining that `create_tmpfile` intentionally returns `EOPNOTSUPP` for filesystems that support directories but not `O_TMPFILE`, instead of matching the sibling defaults' `ENOTDIR`.

kernel/src/fs/file/file_attr/open_args.rs
YanLien: Done. I took the Linux-compatible option and dropped the `O_PATH` rejection. I also moved the `O_EXCL` comment next to the hard-linkability decision.

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
```diff
@@ -0,0 +1,288 @@
...
+	struct dirent *entry;
+	struct stat stat_after_link;
+	struct stat stat_after_second_link;
+
+	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_RDWR, 0666));
+
+	const char *data = "hello from tmpfile";
+	TEST_RES(write(fd, data, strlen(data)), _ret == strlen(data));
+
+	char buf[64];
```
lrh2000: - I'm not sure why some variables are defined at the beginning, while others are defined in the middle. Pick a style and stick with it.

 - In the Asterinas test environment, `SKIP_TEST_IF` does not make much sense because we want to remind ourselves in case we later move `/ext2` to a different location. I think it should be used when `__asterinas__` is not defined.

 - `#ifdef __asterinas__` should have a FIXME comment if it represents a missing feature.

- `buf` tests are too complex. I believe there are simpler ways.

See:
```patch
diff --git a/test/initramfs/src/regression/fs/tmpfile/tmpfile.c b/test/initramfs/src/regression/fs/tmpfile/tmpfile.c
index 4f5aff0db..2efbdf85d 100644
--- a/test/initramfs/src/regression/fs/tmpfile/tmpfile.c
+++ b/test/initramfs/src/regression/fs/tmpfile/tmpfile.c
@@ -3,10 +3,8 @@
 #define _GNU_SOURCE
 
 #include <dirent.h>
-#include <errno.h>
 #include <fcntl.h>
 #include <stdio.h>
-#include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -21,6 +19,7 @@
 #define LINKED_SECOND_NAME "linked_second_file"
 #define LINKED_O_EXCL_NAME "linked_o_excl_file"
 #define SYMLINK_NAME "tmpfile_symlink"
+#define DATA "hello from tmpfile"
 
 #define RAW_O_TMPFILE 020000000
 
@@ -107,20 +106,19 @@ END_TEST()
 
 FN_TEST(tmpfile_open_with_o_path_yields_path_fd)
 {
-	int fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_PATH | O_RDWR, 0666));
-
+	int fd;
 	char buf[1];
-	TEST_ERRNO(read(fd, buf, sizeof(buf)), EBADF);
 
+	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_PATH | O_RDWR, 0666));
+	TEST_ERRNO(read(fd, buf, sizeof(buf)), EBADF);
 	TEST_SUCC(close(fd));
 }
 END_TEST()
 
+// FIXME: Add `O_TMPFILE` support for ext2.
 #ifdef __asterinas__
 FN_TEST(tmpfile_open_on_ext2_returns_eopnotsupp)
 {
-	SKIP_TEST_IF(access(CROSS_LINK_DIR, F_OK) < 0);
-
 	TEST_ERRNO(open(CROSS_LINK_DIR, O_TMPFILE | O_RDWR, 0666), EOPNOTSUPP);
 }
 END_TEST()
@@ -169,18 +167,18 @@ END_TEST()
 
 FN_TEST(tmpfile_invisible_in_readdir)
 {
-	DIR *dir;
 	int fd;
+	DIR *dir;
+	int found;
+	struct dirent *entry;
 
 	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_RDWR, 0666));
 
-	const char *data = "hello tmpfile";
-	TEST_RES(write(fd, data, strlen(data)), _ret == strlen(data));
+	TEST_RES(write(fd, DATA, sizeof(DATA)), _ret == sizeof(DATA));
 
 	dir = TEST_RES(opendir(TEST_DIR), _ret != NULL);
 
-	struct dirent *entry;
-	int found = 0;
+	found = 0;
 	while ((entry = readdir(dir)) != NULL) {
 		if (strcmp(entry->d_name, ".") != 0 &&
 		    strcmp(entry->d_name, "..") != 0) {
@@ -196,29 +194,27 @@ END_TEST()
 
 FN_TEST(tmpfile_write_and_linkat)
 {
-	DIR *dir;
-	int dirfd;
 	int fd;
+	char buf[64];
+	int dirfd;
+	struct stat stat_after_link;
+	DIR *dir;
 	int found;
-	int linked_fd;
 	struct dirent *entry;
-	struct stat stat_after_link;
 	struct stat stat_after_second_link;
+	int linked_fd;
 
 	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_RDWR, 0666));
 
-	const char *data = "hello from tmpfile";
-	TEST_RES(write(fd, data, strlen(data)), _ret == strlen(data));
+	TEST_RES(write(fd, DATA, sizeof(DATA)), _ret == sizeof(DATA));
 
-	char buf[64];
 	TEST_SUCC(lseek(fd, 0, SEEK_SET));
-	ssize_t n = TEST_SUCC(read(fd, buf, sizeof(buf) - 1));
-	buf[n] = 0;
-	TEST_RES(strcmp(buf, data), _ret == 0);
+	TEST_RES(read(fd, buf, sizeof(buf) - 1),
+		 _ret == sizeof(DATA) && memcmp(buf, DATA, sizeof(DATA)) == 0);
 
 	dirfd = TEST_SUCC(open(TEST_DIR, O_RDONLY | O_DIRECTORY));
-	TEST_SUCC(linkat(fd, "", dirfd, LINKED_NAME, AT_EMPTY_PATH));
 
+	TEST_SUCC(linkat(fd, "", dirfd, LINKED_NAME, AT_EMPTY_PATH));
 	TEST_RES(fstat(fd, &stat_after_link), stat_after_link.st_nlink == 1);
 
 	dir = TEST_RES(opendir(TEST_DIR), _ret != NULL);
@@ -235,10 +231,10 @@ FN_TEST(tmpfile_write_and_linkat)
 		 stat_after_second_link.st_nlink == 2);
 
 	linked_fd = TEST_SUCC(open(TEST_DIR "/" LINKED_NAME, O_RDONLY));
+
 	TEST_SUCC(lseek(linked_fd, 0, SEEK_SET));
-	n = TEST_SUCC(read(linked_fd, buf, sizeof(buf) - 1));
-	buf[n] = 0;
-	TEST_RES(strcmp(buf, data), _ret == 0);
+	TEST_RES(read(linked_fd, buf, sizeof(buf) - 1),
+		 _ret == sizeof(DATA) && memcmp(buf, DATA, sizeof(DATA)) == 0);
 
 	TEST_SUCC(close(linked_fd));
 	TEST_SUCC(close(dirfd));
@@ -264,12 +260,17 @@ END_TEST()
 
 FN_TEST(tmpfile_linkat_cross_mount_returns_exdev)
 {
-	SKIP_TEST_IF(
-		dir_is_unavailable_or_same_mount(TEST_DIR, CROSS_LINK_DIR));
-
 	int fd;
 	int tmpfd;
 
+#ifdef __asterinas__
+	TEST_RES(dir_is_unavailable_or_same_mount(TEST_DIR, CROSS_LINK_DIR),
+		 _ret == 0);
+#else
+	SKIP_TEST_IF(
+		dir_is_unavailable_or_same_mount(TEST_DIR, CROSS_LINK_DIR));
+#endif
+
 	fd = TEST_SUCC(open(TEST_DIR, O_TMPFILE | O_RDWR, 0666));
 	tmpfd = TEST_SUCC(open(CROSS_LINK_DIR, O_RDONLY | O_DIRECTORY));
 
```

test/initramfs/src/regression/fs/run_test.sh
```diff
@@ -130,4 +130,5 @@ echo "All mount bind file test passed."
 
 ./symlink/symlink
 
+./tmpfile/tmpfile
 ./utimensat/utimensat
```
lrh2000, line 135: A blank line should be added below `tmpfile/tmpfile`.

kernel/src/syscall/open.rs
lrh2000: Is https://github.com/asterinas/asterinas/pull/3185#discussion_r3285316864 correct, or does the behavior differ depending on the Linux version?

The test fails in my Linux machine (`Linux localhost 7.0.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 15 May 2026 16:45:15 +0000 x86_64 GNU/Linux`):
> test_tmpfile_open_symlink_with_o_nofollow_returns_eloop: `open("/tmp/aster_tmpfile_test" "/" "tmpfile_symlink", (020000000 | 0200000) | 02 | 0400000, 0666)` failed [got Not a directory, but expected Too many levels of symbolic links]

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -288,6 +291,7 @@ struct InodeMeta {
     ctime: Duration,
     mode: InodeMode,
     nlinks: usize,
+    hard_linkability: HardLinkability,
```
tatetian: I don't think this property is mutable after an inode is created. As such, it should join the league of the immutable fields in `RamInode`. This way, we won't need to acquire the lock on the `InodeMeta` in order to retrieve the value of `HardLinkability`.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -76,6 +76,28 @@ impl Path {
         Ok(Self::new(self.mount.clone(), new_child_dentry))
     }
 
+    /// Creates an unnamed temporary file in the directory represented by this path.
```
tatetian: Rewrite the summary sentence to be symmetric to the ones of the existing `new_xxx` methods.

tatetian (commented):
Thanks for the revision. The PR looks almost perfect. Just two nits.

test/initramfs/src/regression/fs/run_test.sh
YanLien: Done. Added the blank line.

test/initramfs/src/regression/fs/tmpfile/tmpfile.c
YanLien: Applied all your suggestions.

kernel/src/fs/vfs/path/mod.rs
YanLien: Done. Updated the summary sentence.

kernel/src/syscall/open.rs
YanLien: > 

Thanks for the suggestions, all updated.

kernel/src/fs/file/file_attr/open_args.rs
YanLien: > 

Thanks for the suggestions, all updated.

kernel/src/fs/vfs/fs_apis/inode.rs
YanLien: Already updated.

kernel/src/syscall/open.rs
YanLien: > Is [#3185 (comment)](https://github.com/asterinas/asterinas/pull/3185#discussion_r3285316864) correct, or does the behavior differ depending on the Linux version?
> 
> The test fails in my Linux machine (`Linux localhost 7.0.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 15 May 2026 16:45:15 +0000 x86_64 GNU/Linux`):
> 
> > test_tmpfile_open_symlink_with_o_nofollow_returns_eloop: `open("/tmp/aster_tmpfile_test" "/" "tmpfile_symlink", (020000000 | 0200000) | 02 | 0400000, 0666)` failed [got Not a directory, but expected Too many levels of symbolic links]

You're right. I initially followed the pr-review suggestion to return `ELOOP`, and did run the test locally but overlooked this failure. After your comment I re-checked and confirmed Linux returns `ENOTDIR`, not `ELOOP`. I've removed the ELOOP check from the kernel and updated the test to expect `ENOTDIR`.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -150,6 +154,8 @@ pub(super) struct RamInode {
     inner: Inner,
     /// Inode metadata
     metadata: SpinLock<InodeMeta>,
+    /// Whether new hard links may be created to this inode.
+    hard_linkability: HardLinkability,
```
tatetian: The fields are ordered roughly in their significance. This `HardLinkability` property is less important than `ino` or `typ`, right? I suggest moving it behind `container_dev_id: DeviceId` but before `extension: Extension`.

And I suggest augmenting the doc comments:

```rust
    /// Whether new hard links may be created to this inode.
    ///
    /// All inodes except temporary files are set to linkable.
    /// For temporary files, the linkability is specified via [`RamInode::create_tmpfile`].
    hard_linkability: HardLinkability,
```

tatetian (commented):
One more nit.

kernel/src/fs/fs_impls/ramfs/fs.rs
YanLien: Yes, that makes sense. I moved it there and added doc comments, with the wording adjusted slightly to match the surrounding field-comment style.

tatetian (approved):

tatetian:
Could you update the "Syscall Flag Coverage" part of the Asterinas Book? In the [`open` and `openat`](https://asterinas.github.io/book/kernel/linux-compatibility/syscall-flag-coverage/file-and-directory-operations/#open-and-openat) section, `O_TMPFILE` is still marked as "Unsupported flags". Need to update both the MdBook page and the corresponding SCML file.

YanLien:
> Could you update the "Syscall Flag Coverage" part of the Asterinas Book? In the [`open` and `openat`](https://asterinas.github.io/book/kernel/linux-compatibility/syscall-flag-coverage/file-and-directory-operations/#open-and-openat) section, `O_TMPFILE` is still marked as "Unsupported flags". Need to update both the MdBook page and the corresponding SCML file.

Sure, I can update the "Syscall Flag Coverage" section in the Asterinas Book, including both the MdBook page and the corresponding SCML file.

Currently, Asterinas' ext2 implementation does not support `O_TMPFILE`. I will first update the documentation. By the way, would you agree to adding ext2 support for `O_TMPFILE`? I can implement it in a follow-up PR if that sounds good to you.

tatetian:
> Sure, I can update the "Syscall Flag Coverage" section in the Asterinas Book, including both the MdBook page and the corresponding SCML file.

Thanks a lot!

> By the way, would you agree to adding ext2 support for `O_TMPFILE`? I can implement it in a follow-up PR if that sounds good to you.

Sure. That would be a fantastic follow-up PR. Please do so upon the new Ext2 implementation #3171.

PR #3186, opened 2026-04-29
Fix popular applications links and package names

cqs21:
This PR fixes navigation in the “Popular Applications” section of the Asterinas Book, and rename some packages.


tatetian (approved):

Issue #3119, opened 2026-04-11
Netfilter Hooks for eBPF


chen1plus:
Related to #2423. I propose adding Netfilter hooks to the networking stack to support `BPF_PROG_TYPE_NETFILTER`. I am implementing this feature as my university project.

### Phase 1 Roadmap

- Netfilter
  - Hook Points
    - Send UDP
  - Registry Management
  - Verdict System
    - Accept / Drop
- eBPF (support `BPF_PROG_TYPE_NETFILTER` only)
  - `BPF_PROG_LOAD`
  - `BPF_LINK_CREATE`
  - Execution upon triggering (w/o JIT)

By implementing these features, we wiil be able to use a custom eBPF program to determine whether to send a UDP packet. Although some critical eBPF features are still missing, we need to take it one step at a time.

### Potential Problem

Regarding the concerns raised by @tatetian about `sk_buff` (refer to #2423), I believe we don't need to change the core networking stack. Instead of modifying our existing internal packet structure, we can define a dedicated `SkBuff` struct specifically for BPF contexts. The transformation (or wrapping) would only occur on-demand when a BPF hook is actually active. So the original packet handling path remains clean and efficient.

In fact, this might be our only option for now. We cannot easily determine the overhead of adopting a C ABI. That said, some performance loss is an acceptable compromise for those who need BPF functionality.

chen1plus:
Even though our phase 1 goals still need some key improvements, laying out phase 2 now is reasonable. Phase 2 should target the ring buffer mechanism for kernel-user space communication and map implementation for eBPF state. These can be handled in separate PRs.

tatetian:
Thanks for working on bringing eBPF to Asterinas.

Before discussing _how_ to implement eBPF support, I want to first raise the question of _why_. This is not a trivial question. Normally, it is a no-brainer to implement a Linux feature in Asterinas, but I am afraid that eBPF is an exception.

There are two reasons that **I am currently reluctant to introducing eBPF to Asterinas**.

[...]

PR #3189, opened 2026-04-29
Netfilter Hook for eBPF

chen1plus:
PR for phase 1 of #3119: a Netfilter hook and minimal eBPF support for `BPF_PROG_TYPE_NETFILTER`.
With this, a user-supplied eBPF program can accept/drop outbound UDP packets.

## Implemented

- Netfilter: UDP send hook point, registry, `ACCEPT` / `DROP` verdicts
- eBPF: `BPF_PROG_LOAD`, `BPF_LINK_CREATE`, interpreter execution

The hook lives in `aster-bigtcp`; the context is only materialized when a hook is active.
I add a minimal test `bpf_udp_hook.c` (I've tried several tricks but I only add this minimal test)

## Feedback wanted

- What additional tests should I include
- The current approach is not scalable for multiple hook points. Could it be handled in a future iteration
- How can I better implement the global hook registry
- Whether passing metadata into the BPF program should be done now or in a follow-up

Will rebase and clean up once direction is settled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,7 @@ dependencies = [
  "int-to-c-enum",
  "jhash",
  "ostd",
+ "rbpf",
  "smoltcp",
  "spin",
  "takeable",
@@ -500,6 +501,15 @@ name = "cobs"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
+
+[[package]]
+name = "combine"
+version = "4.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba5a308b75df32fe02788e748662718f03fde005016435c444eea572398219fd"
+dependencies = [
+ "memchr",
+]
 
 [[package]]
 name = "component"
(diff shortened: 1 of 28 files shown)


chen1plus:
Some more feedback wanted:

- The validator isn't implemented yet, and we still need to figure out how to rebuild it. Since this is a massive amount of work, letting it block these early foundational tasks doesn't seem practical. I think it makes more sense to merge eBPF as an optional feature for now (though I haven't put it behind a feature flag just yet).

- The eBPF VM crate (`rbpf`) is currently pulled into the kernel's `Cargo.toml`. Since it contains unsafe code, should we move it to ostd instead?

chen1plus:
> The validator isn't implemented yet, ...

There's a simple verifier in this PR. It is incomplete but provides memory bounds checking. This guarantees memory safety, even if it cannot prevent infinite loops yet.

> The eBPF VM crate (`rbpf`) is currently pulled into the kernel's `Cargo.toml`. Since it contains unsafe code, should we move it to ostd instead?

I am updating my recommendation to fork `rbpf` into `aster-ebpf`. After further review, the interpreter can be made fully memory-safe by replacing the current host memory address checks with sandbox addresses. This approach is completely feasible and ensures strict Rust memory safety.

PR #3190, opened 2026-04-30
Fix weight update races in `sched_class/fair.rs`

lrh2000:
If a thread invokes `FairAttr::update` twice, there is no guarantee that `FairAttr::fetch_weight` will detect the second update. It may clear the `HAS_PENDING` bit based on the first update:
```
  [ Thread 1 ]                               [ Thread 2 ]

  pending_weight.store(100)
  weight.fetch_or(HAS_PENDING)
                                             weight.load() = HAS_PENDING
                                             pending_weight.load() = 100
  pending_weight.store(200)
  weight.fetch_or(HAS_PENDING)

                                             weight.compare_and_exchange(HAS_PENDING, 100) = Ok(())
```

I think using a spin lock to solve this race condition is unavoidable. However, we can keep the fast path lock-free. Therefore, there shouldn't be a performance penalty.

### DIFF
--- a/kernel/src/sched/sched_class/fair.rs
+++ b/kernel/src/sched/sched_class/fair.rs
@@ -8,6 +8,7 @@ use core::{
 
 use ostd::{
     cpu::{CpuId, num_cpus},
+    sync::SpinLock,
     task::{
         Task,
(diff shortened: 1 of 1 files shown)


StevenJiang1110 (approved):

PR #3191, opened 2026-04-30
Fix sendfile handling of short writes

### LINKED ISSUE #3181: `sys_sendfile()` drops data after a short write to a PTY

### LINKED ISSUE #3209: Bug of CI conformance test (multiboot2-smp4, ltp)
### Describe the bug

There seems to be a bug in the CI conformance test (multiboot2-smp4, ltp). The `All conformance tests passed.` is present, but it is not a standalone line (see the CI test result of #3206 ,  [https://github.com/asterinas/asterinas/actions/runs/25627732487/job/75225761132?pr=3206#step:5:15612](url)). While the `Makefile` check was anchored to the start of a line:
https://github.com/asterinas/asterinas/blob/5f41ff02d681e047f9420bb18c162cfe76116606/Makefile#L266
So it would not match that line, even though the phrase is there. After removing the `^` in the match, the test passed locally.
```qemu.log
futex_wait01                                       PASS       0    
futex_wait02                                       PASS       0    
futex_wait03                                       PASS       0    
futex_wait04                                       PASS       0    
futex_wake01                                       PASS       0    
futex_wake02                                       PASS       0    
futex_wake03                                       PASS       0    
futex_wait_bitset01                                PASS       0    
statx02                                            PASS       0    
statx03                                            PASS       0    

-----------------------------------------------
Total Tests: 500
Total Skipped Tests: 0
Total Failures: 0
Kernel Version: 5.13.0
Machine Architecture: x86_64
Hostname: (none)

All conformance tests passed.
```
Besides, the local test log above shows that there should be more test after the `futex_wait04` test. I am not sure why the test stops here.

### To Reproduce

See the CI test result of #3206  [https://github.com/asterinas/asterinas/actions/runs/25627732487/job/75225761132?pr=3206](url)

cqs21:
Fixes #3181 
Fixes #3209

### DIFF
--- a/kernel/src/syscall/sendfile.rs
+++ b/kernel/src/syscall/sendfile.rs
@@ -4,8 +4,13 @@ use ostd::mm::VmIo;
 
 use super::SyscallReturn;
 use crate::{
-    fs,
-    fs::file::file_table::{RawFileDesc, WithFileTable},
+    fs::{
+        self,
+        file::{
+            InodeHandle, InodeType, SeekFrom,
+            file_table::{RawFileDesc, WithFileTable},
+        },
+    },
     prelude::*,
 };
 
@@ -55,10 +60,36 @@ pub fn sys_sendfile(
         count = MAX_COUNT;
     }
 
+    let origin_f_pos = if offset.is_none() {
+        let outfile_is_pipe = out_file
+            .downcast_ref::<InodeHandle>()
+            .is_some_and(|inode_handle| {
+                inode_handle.path().inode().type_() == InodeType::NamedPipe
+            });
+
+        // If the output file is a pipe, Linux requires the input file to be seekable.
+        // See <https://elixir.bootlin.com/linux/v7.0/source/fs/splice.c#L1038>.
+        // If the output file is not a pipe, the input file does not need to be seekable,
+        // so we simply store the offset if available.
+        // The saved offset is used to restore the correct file position in case a short write occurs.
+        if outfile_is_pipe {
+            Some(
+                in_file.seek(SeekFrom::Current(0)).map_err(|_| {
+                    Error::with_message(Errno::EINVAL, "the in_file is not seekable")
+                })?,
+            )
+        } else {
+            in_file.seek(SeekFrom::Current(0)).ok()
+        }
+    } else {
+        None
+    };
+
     const BUFFER_SIZE: usize = PAGE_SIZE;
     let mut buffer = vec![0u8; BUFFER_SIZE].into_boxed_slice();
     let mut total_len = 0;
     let mut offset = offset.map(|offset| offset as usize);
+    let mut short_write_occurs = false;
 
     while total_len < count {
         // The offset decides how to read from `in_file`.
@@ -70,12 +101,8 @@ pub fn sys_sendfile(
         let max_readlen = buffer.len().min(count - total_len);
 
         // Read from `in_file`
-        let read_res = if let Some(offset) = offset.as_mut() {
-            let res = in_file.read_bytes_at(*offset, &mut buffer[..max_readlen]);
-            if let Ok(len) = res.as_ref() {
-                *offset += *len;
-            }
-            res
+        let read_res = if let Some(offset) = offset {
+            in_file.read_bytes_at(offset, &mut buffer[..max_readlen])
         } else {
             in_file.read_bytes(&mut buffer[..max_readlen])
         };
@@ -102,15 +129,31 @@ pub fn sys_sendfile(
         match write_res {
             Ok(len) => {
                 total_len += len;
-                if len < BUFFER_SIZE {
+                if let Some(offset) = offset.as_mut() {
+                    *offset += len;
+                }
+                if len < read_len {
+                    short_write_occurs = true;
                     break;
                 }
             }
             Err(e) => {
                 if total_len > 0 {
                     warn!("error occurs when trying to write file: {:?}", e);
+                    short_write_occurs = true;
                     break;
                 }
+
+                if offset.is_none() {
+                    if let Some(origin_f_pos) = origin_f_pos {
+                        in_file.seek(SeekFrom::Start(origin_f_pos))?;
+                    } else {
+                        error!(
+                            "the file position of in_file may be incorrect due to a write error"
+                        );
+                    }
+                }
+
                 return Err(e);
             }
         }
@@ -118,6 +161,17 @@ pub fn sys_sendfile(
 
     if let Some(offset) = offset {
         ctx.user_space().write_val(offset_ptr, &(offset as isize))?;
+    } else if short_write_occurs {
+        if let Some(origin_f_pos) = origin_f_pos {
+            // Seek `in_file` to the position corresponding to the last byte actually transferred.
+            // Note: since the file offset lock is not held between saving `origin_f_pos` and this
+            // point, a race condition may occur if another thread reads from the same file
+            // concurrently. Linux permits this behavior and leaves it to userspace to avoid such races.
+            let new_f_pos = origin_f_pos + total_len;
+            in_file.seek(SeekFrom::Start(new_f_pos))?;
+        } else {
+            error!("the file position of in_file may be incorrect due to a short write");
+        }
     }
 
     fs::vfs::notify::on_access(&in_file);


lrh2000:
I am a bit confused how this PR resolves https://github.com/asterinas/asterinas/issues/3181.

From the logs:
<details>
<summary> Logs  </summary>

```
[     9.826] ERROR : actually write 1024
[     9.827] ERROR : try to write 1024

1861
1862
1863
1864
1865
1866
1867
1868
1869
1870
1871
1872
1873
1874
1875
1876
1877
1878
1879
1880
1881
1882
1883
1884
1885
1886
1887
1888
1889
1890
1891
1892
1893
1894
1895
1896
1897
1898
1899
1900
1901
1902
1903
1904
1905
1906
1907
1908
1909
1910
1911
1912
1913
1914
1915
1916
1917
1918
1919
1920
1921
1922
1923
1924
1925
1926
1927
1928
1929
1930
1931
1932
1933
1934
1935
1936
1937
1938
1939
1940
1941
1942
1943
1944
1945
1946
1947
1948
1949
1950
1951
1952
1953
1954
1955
1956
1957
1958
1959
1960
1961
1962
1963
1964
1965
1966
1967
1968
1969
1970
1971
1972
1973
1974
1975
1976
1977
1978
1979
1980
1981
1982
1983
1984
1985
1986
1987
1988
1989
1990
1991
1992
1993
1994
1995
1996
1997
1998
1999
2000
2001
2002
2003
2004
2005
2006
2007
2008
2009
2010
2011
2012
2013
2014
2015
2016
2017
2018
2019
2020
2021
2022
2023
2024
2025
2026
2027
2028
2029
2030
203[     9.828] ERROR : actually write 1024
[     9.829] ERROR : try to write 1024
2680
2681
2682
2683
2684
2685
2686
2687
2688
2689
2690
2691
2692
2693
2694
2695
2696
2697
2698
2699
2700
2701
2702
2703
2704
2705
2706
2707
2708
2709
2710
2711
2712
2713
2714
2715
2716
2717
2718
2719
2720
2721
2722
2723
2724
2725
2726
2727
2728
2729
2730
2731
2732
2733
2734
2735
2736
2737
2738
2739
2740
2741
2742
2743
2744
2745
2746
2747
2748
2749
2750
2751
2752
2753
2754
2755
2756
2757
2758
2759
2760
2761
2762
2763
2764
2765
2766
2767
2768
2769
2770
2771
2772
2773
2774
2775
2776
2777
2778
2779
2780
2781
2782
2783
2784
2785
2786
2787
2788
2789
2790
2791
2792
2793
2794
2795
2796
2797
2798
2799
2800
2801
2802
2803
2804
2805
2806
2807
2808
2809
2810
2811
2812
2813
2814
2815
2816
2817
2818
2819
2820
2821
2822
2823
2824
2825
2826
2827
2828
2829
2830
2831
2832
2833
2834
2835
2836
2837
2838
2839
2840
2841
2842
2843
2844
2845
2846
2847
2848
2849
2850[     9.830] ERROR : actually write 1024
[     9.831] ERROR : try to write 4
~ # [     9.832] ERROR : actually write 4
```
</details>

<details>
<summary> Patch </summary>

```diff
diff --git a/kernel/src/device/tty/mod.rs b/kernel/src/device/tty/mod.rs
index d5bb86625..44267818a 100644
--- a/kernel/src/device/tty/mod.rs
+++ b/kernel/src/device/tty/mod.rs
@@ -206,6 +206,8 @@ impl<D: TtyDriver> Tty<D> {
             return_errno_with_message!(Errno::EIO, "the TTY is closed");
         }
 
+        ostd::error!("try to write {:?}", reader.remain());
+
         let mut buf = vec![0u8; reader.remain().min(IO_CAPACITY)];
         let write_len = reader.read_fallible(&mut buf.as_mut_slice().into())?;
 
@@ -219,6 +221,9 @@ impl<D: TtyDriver> Tty<D> {
             })?
         };
         self.pollee.invalidate();
+
+        ostd::error!("actually write {:?}", len);
+
         Ok(len)
     }
 
```
</details>

The userspace only tries to write 1024 bytes here.

There are also no partial writes (`actually write == try to write` ).

cqs21:
> There are also no partial writes (actually write == try to write ).

I tried this patch, and at the beginning of the log, there will be the following output:
```
[    17.876] ERROR : actually write 4096
[    17.877] ERROR : try to write 4096
[    17.878] ERROR : actually write 2546
[    17.879] ERROR : try to write 1024
1
2
3
4
```

lrh2000:
> [    17.878] ERROR : actually write 2546
> [    17.879] ERROR : try to write 1024

Thanks for the explanation.

It seems this is caused by the output buffer being full? It is not related to the `IO_CAPABILITY` settings.

But that seems odd to me, too. I have several more questions.
 1. So, even if the output buffer is full, we can't perform a partial write? That sounds ridiculous to me. Could you provide more evidence from the Linux kernel and/or userspace code?
 2. Does userspace silently count a partial write as a full write and drop some bytes at the end instead of the middle? So:
    - Why does userspace treat a partial write as a full write? I assume it is glibc, the de facto standard infrastructure of the Linux ecosystem, so bugs are unlikely.
    - Why does userspace drop bytes from the end instead of the middle?
    - What happens if a signal interrupts the system call and a partial write occurs?

cqs21:
@lrh2000 Thanks for raising these questions. You were right to push on that point, and my earlier explanation was wrong.

The root cause is not that TTY/PTY short writes are inherently broken or that partial writes must not happen when the output buffer is full. A positive short write on a PTY is valid behavior. The real bug should be in Asterinas `sys_sendfile`: after reading a chunk from the input file, it does not correctly handle a short write on the output side. When the write returns fewer bytes than were read, the input file offset has already advanced by the full read size instead of the actual written size. As a result, the next read starts from an incorrect offset and skips the unwritten tail of the previous chunk. 

I will push a proper fix afterwards.

lrh2000:
It was difficult to determine if the changes in this PR are correct because our previous `sys_sendfile` implementation was also incorrect. I checked the Linux implementation, which is quite complex. It does not actually go through the normal `read`/`write` interfaces to implement `sys_sendfile`, instead, it uses `splice_read`/`splice_write`.

Unfortunately, we may ultimately need something similar to `splice_read`/`splice_write` to ensure the file offset is placed correctly in case of failure. I also assume that the Linux implementation of `splice_read`/`splice_write` will also have different blocking behavior: Currently, our `sys_sendfile` implementation can be blocked multiple times in the middle of system calls, which I assume is incorrect.

But we need to fix the CI failure as soon as possible. I don't hold strong opinion here since they're mostly existing problems. Could @StevenJiang1110 check this PR since you're the original author (c04fd0c9a0dffc54d2cfc985583569bdc7b7f629)?

tatetian:
> instead, it uses `splice_read`/`splice_write`.

It is time to implement the `splice` syscall...

kernel/src/syscall/sendfile.rs
```diff
@@ -95,25 +91,42 @@ pub fn sys_sendfile(
...
             }
         }
+
+        if let Some(offset) = offset.as_mut() {
+            *offset += write_offset;
+        }
+
+        if !remaining.is_empty() {
+            break;
+        }
```
StevenJiang1110, line 159: I check the Linux implementation about sendfile. It looks like Linux also allows short write. So the problem should be that we should correct the in file position if short write occurs.

```diff
diff --git a/kernel/src/syscall/sendfile.rs b/kernel/src/syscall/sendfile.rs
index a0c93479f..2cd25f07c 100644
--- a/kernel/src/syscall/sendfile.rs
+++ b/kernel/src/syscall/sendfile.rs
@@ -4,8 +4,13 @@ use ostd::mm::VmIo;
 
 use super::SyscallReturn;
 use crate::{
-    fs,
-    fs::file::file_table::{RawFileDesc, WithFileTable},
+    fs::{
+        self,
+        file::{
+            InodeHandle, SeekFrom,
+            file_table::{RawFileDesc, WithFileTable},
+        },
+    },
     prelude::*,
 };
 
@@ -55,10 +60,37 @@ pub fn sys_sendfile(
         count = MAX_COUNT;
     }--
 
+    // FIXME: We should also check whether the out file is a named pipe.
+    let outfile_is_pipe = out_file
+        .downcast_ref::<InodeHandle>()
+        .is_some_and(|inode_handle| {
+            inode_handle
+                .path()
+                .inode()
+                .downcast_ref::<AnonPipeInode>()
+                .is_some()
+        });
+
+    // If the output file is a pipe, Linux requires the input file to be seekable.
+    // See <https://elixir.bootlin.com/linux/v7.0/source/fs/splice.c#L1038>.
+    // If the output file is not a pipe, the input file does not need to be seekable,
+    // so we simply store the offset if available.
+    // The saved offset is used to restore the correct file position in case a short write occurs.
+    let origin_f_pos = if outfile_is_pipe {
+        Some(
+            in_file
+                .seek(SeekFrom::Current(0))
+                .map_err(|_| Error::with_message(Errno::EINVAL, "the in_file is not seekable"))?,
+        )
+    } else {
+        in_file.seek(SeekFrom::Current(0)).ok()
+    };
+
     const BUFFER_SIZE: usize = PAGE_SIZE;
     let mut buffer = vec![0u8; BUFFER_SIZE].into_boxed_slice();
     let mut total_len = 0;
     let mut offset = offset.map(|offset| offset as usize);
+    let mut short_write_occurs = false;
 
     while total_len < count {
         // The offset decides how to read from `in_file`.
@@ -103,12 +135,14 @@ pub fn sys_sendfile(
             Ok(len) => {
                 total_len += len;
                 if len < BUFFER_SIZE {
+                    short_write_occurs = true;
                     break;
                 }
             }
             Err(e) => {
                 if total_len > 0 {
                     warn!("error occurs when trying to write file: {:?}", e);
+                    short_write_occurs = true;
                     break;
                 }
                 return Err(e);
@@ -118,6 +152,17 @@ pub fn sys_sendfile(
 
     if let Some(offset) = offset {
         ctx.user_space().write_val(offset_ptr, &(offset as isize))?;
+    } else if short_write_occurs {
+        if let Some(origin_f_pos) = origin_f_pos {
+            // Seek `in_file` to the position corresponding to the last byte actually transferred.
+            // Note: since the file offset lock is not held between saving `origin_f_pos` and this
+            // point, a race condition may occur if another thread reads from the same file
+            // concurrently. Linux permits this behavior and leaves it to userspace to avoid such races.
+            let new_f_pos = origin_f_pos + total_len;
+            in_file.seek(SeekFrom::Start(new_f_pos))?;
+        } else {
+            error!("the file position of in_file may be incorrect due to a short write");
+        }
     }
 
     fs::vfs::notify::on_access(&in_file);
```

StevenJiang1110: If no bytes are successfully written, we should also correct the in file position. The above patch forgets about this case, you can also fix it...

StevenJiang1110:
> It is time to implement the `splice` syscall...

Yes, in the long run we will need to implement the splice_read/splice_write methods — otherwise it will be very difficult to get sendfile fully correct. However, this could be a significant undertaking, as we would need to implement splice_read/splice_write for most file types, including device files, sockets, and various filesystems, each of which may require its own specialized implementation.

Do you think it would be acceptable to apply a temporary patch similar to the one discussed in https://github.com/asterinas/asterinas/pull/3191#discussion_r3232097182 for now? This would ensure the file offset is handled correctly in the vast majority of cases, while emitting an error message for the edge cases that cannot be properly handled yet. This way, we can unblock the CI issue first, and leave the full splice implementation to a separate follow-up PR.

kernel/src/syscall/sendfile.rs
```diff
@@ -102,22 +129,45 @@ pub fn sys_sendfile(
...
                     warn!("error occurs when trying to write file: {:?}", e);
+                    short_write_occurs = true;
                     break;
                 }
+
+                if let Some(origin_f_pos) = origin_f_pos
+                    && offset.is_none()
+                {
+                    in_file.seek(SeekFrom::Start(origin_f_pos))?;
+                }
```
StevenJiang1110, line 155: If `origin_f_pos` is `None` and offset is not `None`, we should also print an error message like line 169 for consistency?

StevenJiang1110 (commented):
LGTM! Just one small issue.

StevenJiang1110 (approved):

lrh2000:
Well... **If the output file is a pipe**

```rust
        // If the output file is a pipe, Linux requires the input file to be seekable.
        // See <https://elixir.bootlin.com/linux/v7.0/source/fs/splice.c#L1038>.
```

The Linux code is **between two non-pipes**:
```
 * splice_direct_to_actor - splices data directly between two non-pipes
```
```c
	/*
	 * We require the input to be seekable, as we don't want to randomly
	 * drop data for eg socket -> socket splicing. Use the piped splicing
	 * for that!
	 */
	if (unlikely(!(in->f_mode & FMODE_LSEEK)))
		return -EINVAL;
```

Are we correctly interpreting the Linux code here....?

---

In addition, I suppose Linux allows non-seekable `in_file` if the destination is a pipe because Linux specializes in pipe handling to avoid dropping data. However, since we don't have the special handling, perhaps we should not follow the Linux checks. Instead, we should reject all non-seekable `in_file`s.

StevenJiang1110:
> Are we correctly interpreting the Linux code here....?

Sorry, the condition here is inverted — my mistake.

> In addition, I suppose Linux allows non-seekable in_file if the destination is a pipe because Linux specializes in pipe handling to avoid dropping data. However, since we don't have the special handling, perhaps we should not follow the Linux checks. Instead, we should reject all non-seekable in_files.

My original concern was that we might break nginx's use of sendfile — specifically, if nginx ever calls sendfile from a socket to somewhere else, we would need to allow a non-seekable in_file. I just used Codex to check how nginx uses sendfile, and it appears that nginx only uses it to copy data from regular files to sockets. So there should be no need to keep this special case open.

Thanks for point out these problems. I will fix them in a separate PR.

PR #3192, opened 2026-05-01
Add confidential containers documentation

taosue:
This PR adds documentation on how to use Asterinas as the guest kernel for Confidential Containers.

### DIFF
--- a/book/src/kernel/vm-based-containers/coco.md
+++ b/book/src/kernel/vm-based-containers/coco.md
@@ -1,0 +1,279 @@
+# Using Asterinas as a Confidential Containers Guest Kernel
+
+This guide explains how to use Asterinas
+as the guest kernel for Confidential Containers (CoCo).
+
+[Confidential Containers](https://github.com/confidential-containers/confidential-containers)
+is a VM-based containers stack built on [Kata Containers](https://github.com/kata-containers/kata-containers)
+and protected with [confidential computing](https://en.wikipedia.org/wiki/Confidential_computing) technology.
+It runs pods inside confidential virtual machines (CVMs)
+while integrating with Kubernetes and confidential computing workflows.
+If you only need VM-based isolation without confidential computing,
+see [Using Asterinas as a Kata Guest Kernel](kata.md) instead.
+
+This guide uses the
+[Asterinas fork of Confidential Containers](https://github.com/asterinas/confidential-containers),
+which carries Asterinas-specific patches, helper scripts,
+and configuration for building, installing, and testing CoCo
+with Asterinas as the guest kernel.
+
+The current Asterinas-based CoCo setup supports two paths:
+
+- a regular CoCo runtime for local development and debugging
+- a [Intel TDX](../intel-tdx.md)-powered CoCo runtime for CVMs
+
+## Step 1: Prepare the host kernel
+
+CoCo requires a host with KVM and vhost support.
+The commands in this guide are currently written for x86-64 hosts.
+
+Verify that the required device nodes exist:
+
+```bash
+ls /dev/kvm /dev/vhost-vsock
+```
+
+If any of them are missing,
+load the matching kernel modules:
+
+```bash
+sudo modprobe kvm
+sudo modprobe kvm_intel  # Or use kvm_amd on AMD hosts.
+sudo modprobe vhost_vsock
+```
+
+Then make sure the user running Docker can access
+`/dev/kvm` and `/dev/vhost-vsock`.
+
+If you plan to use the TDX-powered CoCo runtime,
+your host additionally requires Intel TDX hardware support,
+a TDX-capable host kernel, and the TDX firmware modules.
+See [Intel TDX](../intel-tdx.md) for the full list of TDX prerequisites.
+
+## Step 2: Enter the CoCo environment
+
+We provide the `asterinas/coco` image with CoCo and Asterinas preinstalled.
+End users should use it directly.
+Kernel developers should additionally mount the local Asterinas source.
+
+### Step 2.1: Prepare Docker arguments
+
+Define the following Docker arguments:
+
+```bash
+COCO_DOCKER_ARGS=(
+    --cgroupns host
+    --privileged
+    --device /dev/kvm
+    --device /dev/vhost-vsock
+    --tmpfs /var/lib/containerd-nydus:rw,size=512m
+)
+```
+
+These flags are required for CoCo:
+
+- `--cgroupns host` shares the host cgroup namespace
+  so that `containerd` inside the container can manage CoCo workloads.
+- `--privileged` is required for KVM and nested container management.
+- `--device /dev/kvm` and `--device /dev/vhost-vsock`
+  expose the virtualization devices
+  that CoCo needs.
+- `--tmpfs /var/lib/containerd-nydus:rw,size=512m`
+  gives the nydus image service enough temporary storage space for guest-side image pulling.
+
+### Step 2.2: Start the Docker container
+
+#### For end users
+
+Run an `asterinas/coco` Docker container
+to enter an environment with CoCo and Asterinas preinstalled:
+
+```bash
+docker run -it \
+    "${COCO_DOCKER_ARGS[@]}" \
+    asterinas/coco:0.17.2-20260407
+```
+
+The image bundles the CoCo runtime, the Asterinas guest kernel,
+the cluster setup script `/opt/coco/setup-coco-k8s.sh`,
+and the pod manifests under `/opt/coco/manifests/`
+used in the steps below.
+
+After entering the container,
+continue with [Step 3](#step-3-start-a-coco-workload).
+
+#### For kernel developers
+
+Clone the Asterinas source,
+then run an `asterinas/coco` Docker container
+with the source tree mounted:
+
+```bash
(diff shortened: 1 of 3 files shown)


tools/bump_version.sh
```diff
@@ -36,12 +36,14 @@ update_dep_version() {
...
     echo "Updating file $1"
-    # Update the version of the Asterinas development container
+    # Update the version of the asterinas/asterinas container
     sed -i "s/asterinas\/asterinas:[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+\(-[[:digit:]]\+\)\?/asterinas\/asterinas:${new_version}/g" $1
     # Update the version of asterinas/osdk container
     sed -i "s/asterinas\/osdk:[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+\(-[[:digit:]]\+\)\?/asterinas\/osdk:${new_version}/g" $1
     # Update the version of asterinas/kata container
     sed -i "s/asterinas\/kata:[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+\(-[[:digit:]]\+\)\?/asterinas\/kata:${new_version}/g" $1
+    # Update the version of asterinas/coco container
+    sed -i "s/asterinas\/coco:[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+\(-[[:digit:]]\+\)\?/asterinas\/coco:${new_version}/g" $1
```
tatetian: @StevenJiang1110 On a second thought, I think that it was a mistake to include the `asterinas/kata` and `asterinas/coco` Docker images into the `bump_version.sh` script. Although project version bumps are infrequent, Docker version bumps are frequent. If all three version numbers must be kept in sync, we need to generate `asterinas/coco` and `asterinas/kata` Docker images frequently with human intervention (we don't have an automation mechanism in place), which is quite a maintenance burden.

On the other hand, since the `asterinas/kata` and `asterinas/coco` Docker images are self-contained (do not rely on the kernel project to work), using an (outdated) version of these images can still succeed. So it is tolerable for the versions of the `asterinas/kata` and `asterinas/coco` Docker images lag behind that of the `asterinas/asterinas`.

Our workflow for generating these Docker images is:
1. The Asterinas main project version bumps;
2. The new `asterinas/asterinas` Docker image generates;
3. If this is a major release, the maintainers for `asterinas/kata` and `asterinas/coco` trigger the generation of a new release.
4. After the two Docker images generate successfully, submit a PR to update the `asterinas/kata` and `asterinas/coco` Docker image versions in the Asterinas Book.

@StevenJiang1110 @taosue What do you think? If the two of your agree, you can persist this workflow by writing it down into the README of our forked Kata and CoCo repos.

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
+# Using Asterinas as a Confidential Containers Guest Kernel
+
+This guide explains how to use Asterinas
+as the guest kernel for Confidential Containers (CoCo).
+
+[Confidential Containers](https://github.com/confidential-containers/confidential-containers)
+is a VM-based containers stack built on [Kata Containers](https://github.com/kata-containers/kata-containers).
```
tatetian: In this definition, you need to mention "confidential computing". Otherwise, how can the containers be considered confidential?

```
is a VM-based containers stack built on [Kata Containers](https://github.com/kata-containers/kata-containers)
and protected with [confidential computing](https://en.wikipedia.org/wiki/Confidential_computing) technology.
```

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+It runs pods inside lightweight virtual machines
```
tatetian: lightweight virtual machines --> confidential virtual machines (CVMs)

Defining the "CVM" abbreviation would come in handy for use in the later part of the page.

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+while integrating with Kubernetes and confidential-computing workflows.
+Compared with plain Kata usage,
+Confidential Containers adds Kubernetes runtime integration,
+guest-side image pulling,
+and the configuration needed to manage confidential workloads.
```
tatetian: I am wondering if mentioning "Kubernetes runtime integration", "guest-side image pulling", and "the configuration needed to manage confidential workloads" is a favor to the reader or a noise.

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+and configuration for building, installing, and testing CoCo
+with Asterinas as the guest kernel.
+
+The current Asterinas-based CoCo setup supports two paths:
+
+- a regular CoCo runtime for local development and debugging
+- a TDX-based CoCo runtime for confidential VMs
+
+So this guide is about both using Asterinas with CoCo in general
+and using Asterinas inside Intel TDX-backed confidential VMs in particular.
```
tatetian: Remove this paragraph; it basically restates having "two paths". And the writing sounds weird: I don't know why it emphasizes "in general" and "in particular". I can't see the point.

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+```bash
+sudo modprobe kvm
+sudo modprobe kvm_intel  # Or use kvm_amd on AMD hosts.
+sudo modprobe vhost_vsock
+```
+
+Then make sure the user running Docker can access
+`/dev/kvm` and `/dev/vhost-vsock`.
+
+If you plan to use the TDX-based CoCo runtime,
```
tatetian: I am not sure that saying "TDX-based" is appropriate. If CoCo runtime has a TDX build, then you can name it "TDX-based CoCo runtime". But if the same CoCo runtime may be deployed in normal VM or TDX VM, then I would say "TDX-powered".

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+This guide uses the
+[Asterinas fork of Confidential Containers](https://github.com/asterinas/confidential-containers),
+which carries Asterinas-specific patches, helper scripts,
+and configuration for building, installing, and testing CoCo
+with Asterinas as the guest kernel.
+
+The current Asterinas-based CoCo setup supports two paths:
+
+- a regular CoCo runtime for local development and debugging
+- a TDX-based CoCo runtime for confidential VMs
```
tatetian: a TDX-based CoCo --> a [Intel TDX](../intel-tdx)-based

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+#### For kernel developers
+
+Clone the Asterinas source,
+then run an `asterinas/coco` Docker container
+with the source tree mounted:
+
+```bash
+# Clones the Asterinas source locally.
+ASTERINAS_SRC=$HOME/asterinas
+cd $HOME && git clone https://github.com/asterinas/asterinas.git
```
tatetian: The writing here differs from Kata; you might want to use the same expression.

book/src/kernel/vm-based-containers/coco.md
```diff
@@ -0,0 +1,287 @@
...
+
+docker run -it \
+    "${COCO_DOCKER_ARGS[@]}" \
+    -v "${ASTERINAS_SRC}:/root/asterinas" \
+    asterinas/coco:0.17.2-20260407
+```
+
+Since running CoCo requires many additional dependencies
+that would be tedious to install manually,
+the `asterinas/coco` image is built on top of the `asterinas/asterinas` image.
```
tatetian: This paragraph is illogical: it says because "CoCo requires many additional dependencies", we built "on top of the `asterinas/asterinas` image".

tatetian (commented):
Thanks you for the excellent work on supporting Confidential Containers.

tools/bump_version.sh
taosue: I agree with this workflow.

Otherwise, we would need to generate `asterinas/coco` automatically via CI, and there would still inevitably be a period of time when `asterinas/coco` is not in sync with `asterinas/asterinas`.

First, end users may not be able to pull the version described in the documentation before the corresponding `asterinas/coco` image is generated. Second, CI test could easily break if it tries to pull the same version as `asterinas/asterinas`. Third, an automatically generated `asterinas/coco` image may introduce regressions, so we should validate it before submitting a version update.

StevenJiang1110: > @StevenJiang1110 @taosue What do you think? If the two of your agree, you can persist this workflow by writing it down into the README of our forked Kata and CoCo repos.

That makes sense to me. With the current approach, there is a gap between the release of the asterinas/asterinas image and the subsequent release of the asterinas/kata image. Furthermore, the CoCo release process likely depends on the Kata release being completed first, since it needs to use the TDX-enabled kernel built as part of the Kata release, so the gap is larger. For these reasons, having the image published and the version updated manually by a person is probably the better approach.

I will update the README of asterinas/kata to include this workflow.

I remove the bump asterinas/kata code in #3220

tatetian (approved):

PR #3197, opened 2026-05-06
Fix error codes for `/proc/self/fd/-1`

lrh2000:
This is not quite correct...
```
~ # cat /proc/self/fd/-1
cat: can't open '/proc/self/fd/-1': Bad file descriptor
```


tatetian (approved):

PR #3198, opened 2026-05-06
Non-functionality improvements to System V semaphores

lrh2000:
This PR contains a series of non-functionality improvements to System V semaphores.

There are quite a few commits. However, each one is small, so I hope that will help with the review process. I don't know if it's worth writing descriptions to explain each commit. I think all the commits are obvious improvements. Please let me know if you have any questions.

 - "Convert some spin locks to mutexes": This may be the only commit that does not belong to strict Non-Functionality Changes (NFC). The underlying reason is that there are quite a few semaphores and pending operations, so the execution time under those locks is relatively long. `Mutex` is a better fit than `SpinLock`.

Perhaps I should squash some commits here? I'm not sure, but I'm happy to follow any guidance.


StevenJiang1110 (approved):

PR #3199, opened 2026-05-06
(Lots of) bug fixes to System V semaphores

### LINKED ISSUE #1370: SysV semaphore can be decreased below 0
### Describe the bug

Semaphore's value is restricted to being greater than or equal to 0. However, when the set of operations (`sops`) contains duplicate sop, there is no mechanism to handle this properly. For example, subtracting a number twice from a semaphore, each sop doesn't reduce it below zero, but the two sops can result in a negative value.

### To Reproduce

1. Create a semaphore
2. Set/increase the value to `1`
3. Decrease the value twice in one `semop()` syscall.
4. Check the value, which shows `-1`

e2d702813fa87b806383a1b84e1faec

### Expected behavior

Step 3 should block (like Linux):

a4fc4506851dc6f28c818711a349dee

### Additional context

In Asterinas, it warns but continues to execute when dupsop is encountered:
https://github.com/asterinas/asterinas/blob/4d36dd541f3ea4ec6ac72681feb6d587f87bab6d/kernel/src/ipc/semaphore/system_v/sem.rs#L157-L159

In Linux, the situation is handled by [perform_atomic_semop_slow](https://elixir.bootlin.com/linux/v6.0.9/C/ident/perform_atomic_semop_slow) function:
<https://elixir.bootlin.com/linux/v6.0.9/source/ipc/sem.c#L731-L732>

lrh2000:
~~This PR is based on #3198. Please review it first.~~

There are many bugs in the System V semaphore implementation. This PR contains a series of bug fixes.

Each commit fixes one bug, or two if they're closely related. Each commit also adds a regression test for the fixed problem. I hope I won't need to explain each commit, as the bug is obvious to me. Again, please let me know if you have any questions.

 - "Fix race on timed out `semop`": This is the only commit in this PR that does not have a regression test. It's a race condition with an extremely small window, so it's difficult to test. The problem is illustrated in the following example:
```
[ Thread A ]              [ Thread B ]                    Timer

semop()

   waiter.wait()

                                                          waker.wake_up()

   status.load()
     = Pending

                          semctl(SETVAL)

                              perform_atomic_semop()
                              status.store(Normal)

   pending_op.retain()

   return EAGAIN
```

In this case, Thread A sees the `semop` system call return `EAGAIN`. However, the operation has already been performed in Thread B, which is problematic.

Perhaps I should squash some commits here? I'm not sure, but I'm happy to follow any guidance.

Fixes https://github.com/asterinas/asterinas/issues/1370. After this PR, the test case mentioned in that issue will be rejected as unsupported.

### DIFF
--- a/kernel/src/ipc/ipc_ns.rs
+++ b/kernel/src/ipc/ipc_ns.rs
@@ -13,8 +13,8 @@ use aster_rights::ReadOp;
 use spin::Once;
 
 use super::{
+    IPC_PRIVATE, IpcFlags, IpcId, IpcKey,
     ipc_ids::IpcIds,
-    key_t,
     semaphore::system_v::{
         PermissionMode,
         sem_set::{SEMMNI, SemaphoreSet},
@@ -22,7 +22,6 @@ use super::{
 };
 use crate::{
     fs::pseudofs::{NsCommonOps, NsType, StashedDentry},
-    ipc::IpcFlags,
     prelude::*,
     process::{
         Credentials, UserNamespace, credentials::capabilities::CapSet, posix_thread::PosixThread,
@@ -59,10 +58,16 @@ impl IpcNamespace {
     }
 
     fn new(owner: Arc<UserNamespace>) -> Arc<Self> {
+        const MAX_SEM_ID: IpcId = {
+            assert!(SEMMNI <= u32::MAX as usize);
+            IpcId::new(SEMMNI as u32)
+        };
+
+        let sem_ids = IpcIds::new(MAX_SEM_ID);
         let stashed_dentry = StashedDentry::new();
 
         Arc::new(Self {
-            sem_ids: IpcIds::new(SEMMNI),
+            sem_ids,
             owner,
             stashed_dentry,
         })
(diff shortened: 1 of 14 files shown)


lrh2000:
The number of commits in this PR is still a bit large...

I verified that the regression test passes for each commit and fails if we revert the kernel changes for that commit. Here is the script:
<details>
<summary> script </summary>

```bash
#!/bin/bash

set -e

commits=$(git rev-list HEAD^1 | head -13)

for commit in $commits; do
	git checkout $commit

	cat <<'EOF' | git apply
diff --git a/test/initramfs/src/regression/scripts/run_regression_test.sh b/test/initramfs/src/regression/scripts/run_regression_test.sh
index d6e821b5c..50f4e4d88 100755
--- a/test/initramfs/src/regression/scripts/run_regression_test.sh
+++ b/test/initramfs/src/regression/scripts/run_regression_test.sh
@@ -4,6 +4,13 @@
 
 set -e
 
+./test/ipc/sem/sem
+
+echo "Sempahore regression tests passed."
+echo "All regression tests passed."
+exit 0
+
+
 SCRIPT_DIR=/test
 
 for dir in $(find -L "${SCRIPT_DIR}" -mindepth 1 -maxdepth 1 -type d); do
EOF

	git show -q | tee -a semaphore_test_logs

	echo "" >> semaphore_test_logs
	if make run_kernel AUTO_TEST=regression >> semaphore_test_logs; then
		echo "commit $commit [                                ] semaphore tests pass" | \
			tee -a semaphore_test_result
	else
		echo "commit $commit [                                ] semaphore tests fail" | \
			tee -a semaphore_test_result
	fi

	echo "" >> semaphore_test_logs
	echo 'Reverting all changes under kernel/' >> semaphore_test_logs
	git show kernel | git apply -R

	echo "" >> semaphore_test_logs
	if make run_kernel AUTO_TEST=regression >> semaphore_test_logs; then
		echo "commit $commit [ reverted changes under kernel/ ] semaphore tests pass" | \
			tee -a semaphore_test_result
	else
		echo "commit $commit [ reverted changes under kernel/ ] semaphore tests fail" | \
			tee -a semaphore_test_result
	fi

	git diff | git apply -R

	echo "" >> semaphore_test_logs
	echo "" >> semaphore_test_logs
	echo "" >> semaphore_test_logs
	echo "" >> semaphore_test_logs
	echo "" >> semaphore_test_logs
	echo "" >> semaphore_test_logs
done
```
</details>

`git rev-list HEAD^1 | head -13` excludes the last commit because it does not have changes under `kernel/`. This script tests all 13 other commits in this PR.

The output is:
```
commit 3e4b7b185c9fa45d6e59c7f27686b32d85f3e54c [                                ] semaphore tests pass
commit 3e4b7b185c9fa45d6e59c7f27686b32d85f3e54c [ reverted changes under kernel/ ] semaphore tests fail
commit 7402584b0c134be10633486f2c7f8205fdcb4588 [                                ] semaphore tests pass
commit 7402584b0c134be10633486f2c7f8205fdcb4588 [ reverted changes under kernel/ ] semaphore tests fail
commit 643e0c153dd1b9ed0b08c435a6e8cfd41784d911 [                                ] semaphore tests pass
commit 643e0c153dd1b9ed0b08c435a6e8cfd41784d911 [ reverted changes under kernel/ ] semaphore tests fail
commit 00d56c4def885f3128aa33cf082349da8428578d [                                ] semaphore tests pass
commit 00d56c4def885f3128aa33cf082349da8428578d [ reverted changes under kernel/ ] semaphore tests fail
commit 193730a9cf20af872345ac942e5e669a70a54188 [                                ] semaphore tests pass
commit 193730a9cf20af872345ac942e5e669a70a54188 [ reverted changes under kernel/ ] semaphore tests fail
commit 3bcdbda6e83547a032692c6e2411416adefa6d82 [                                ] semaphore tests pass
commit 3bcdbda6e83547a032692c6e2411416adefa6d82 [ reverted changes under kernel/ ] semaphore tests fail
commit a5bff5427df4207b26ed4062b85ba3ec5bbf93fa [                                ] semaphore tests pass
commit a5bff5427df4207b26ed4062b85ba3ec5bbf93fa [ reverted changes under kernel/ ] semaphore tests fail
commit 20067f4beb3c6a77815e2ae960f5bc979f0bf22e [                                ] semaphore tests pass
commit 20067f4beb3c6a77815e2ae960f5bc979f0bf22e [ reverted changes under kernel/ ] semaphore tests fail
commit 12ba928d18c50c257edc24e2c34be367de6d623c [                                ] semaphore tests pass
commit 12ba928d18c50c257edc24e2c34be367de6d623c [ reverted changes under kernel/ ] semaphore tests fail
commit 3fc5b6d52fbe876f005f0fb7a0973cc2f2349db1 [                                ] semaphore tests pass
commit 3fc5b6d52fbe876f005f0fb7a0973cc2f2349db1 [ reverted changes under kernel/ ] semaphore tests pass
commit 8cbcf389a27161b8aacb7525a9158eba51bc5a1d [                                ] semaphore tests pass
commit 8cbcf389a27161b8aacb7525a9158eba51bc5a1d [ reverted changes under kernel/ ] semaphore tests fail
commit 587db3f45d18aaed0c043979dc6e0c1f8c716802 [                                ] semaphore tests pass
commit 587db3f45d18aaed0c043979dc6e0c1f8c716802 [ reverted changes under kernel/ ] semaphore tests fail
commit e1335c29c67b24ec802f0f053691d12097f739d3 [                                ] semaphore tests pass
commit e1335c29c67b24ec802f0f053691d12097f739d3 [ reverted changes under kernel/ ] semaphore tests fail
```

Note that commit 3fc5b6d52fbe876f005f0fb7a0973cc2f2349db1 does not have a regression test. The test succeeds regardless of whether the changes under `kernel/` are reverted. The reason is explained in the PR description.


Some tests got stuck without kernel changes, requiring manual interaction with the command `pkill qemu`. Those tests are counted as `failed`.

kernel/src/ipc/ipc_ns.rs
```diff
@@ -59,10 +58,16 @@ impl IpcNamespace {
     }
 
     fn new(owner: Arc<UserNamespace>) -> Arc<Self> {
+        const MAX_SEM_ID: IpcId = {
+            assert!(SEMMNI <= u32::MAX as usize);
```
StevenJiang1110, line 62: Can the assertion be removed?

The reason is that `IpcId` is defined as `pub type IpcId = RangedU32<1, { i32::MAX as u32 }>;`, so `IpcId::new` already asserts that `SEMMNI < i32::MAX`, which is a strictly tighter bound.

kernel/src/ipc/semaphore/system_v/sem.rs
```diff
@@ -248,33 +353,50 @@ fn do_smart_update(inner: &mut SemSetInner, pending_op: &PendingOp) -> LinkedLis
...
-        update_pending_alter(sems, pending_alter, pending_const, &mut wake_queue);
+        let _ = update_pending_alter(sems, pending_alter, pending_const, &mut wake_queue);
     }
 
     wake_queue
 }
 
 /// Looks for alteration operations that can be completed and then completes them.
 ///
+/// This method returns whether there is at least one operation has been completed. The caller
```
StevenJiang1110:
```suggestion
/// This method returns whether at least one operation has been completed. The caller
```

The same issue applies to line 422.

test/initramfs/src/conformance/gvisor/blocklists/semaphore_test
```diff
@@ -4,14 +4,5 @@ SemaphoreTest.SemCtlValAll
 SemaphoreTest.SemInfo
 SemaphoreTest.SemIpcSet
 SemaphoreTest.SemopGetncnt
-SemaphoreTest.SemopGetncntOnSetRemoval
-SemaphoreTest.SemopGetncntOnSignal
 SemaphoreTest.SemopGetzcnt
-SemaphoreTest.SemopGetzcntOnSetRemoval
-SemaphoreTest.SemopGetzcntOnSignal
-# SemOpMultiNoBlock requires handling the dupsop situation
```
StevenJiang1110, line 12: Was removing this comment intentional? The `SemOpMultiNoBlock` test still appears to be blocked.

kernel/src/ipc/semaphore/system_v/sem_set.rs
```diff
@@ -120,32 +122,37 @@ impl SemSetInner {
...
-        let inner = self.inner.lock();
-        let pending_const = &inner.pending_const;
-        let mut count = 0;
-        for i in pending_const.iter() {
-            for sem_buf in i.sops_iter() {
-                if sem_buf.sem_num() == sem_num {
-                    count += 1;
-                }
-            }
+    pub fn pending_const_count(&self, sem_num: usize) -> Result<usize> {
```
StevenJiang1110: The method name seems to have become inappropriate after this PR, since it now counts operations across both the `pending_const` and `pending_alter` lists.

How about renaming it to pending_zero_count?

```rust
/// Returns the number of pending operations waiting for the semaphore at
/// `sem_num` to become zero.
pub fn pending_zero_count(&self, sem_num: usize) -> Result<usize> {
```

The same issue applies to `pending_alter_count` — perhaps it could be renamed to `pending_decrease_count`.

kernel/src/ipc/semaphore/system_v/sem.rs
```diff
@@ -71,6 +94,43 @@ impl PendingOp {
...
+                    return Err(Status::ErrWouldBlock);
+                }
+                return Ok(PendingBlocker::Decrease(sem_num));
+            }
+            if i32::from(op.sem_op) > SEMVMX - val {
+                return Err(Status::ErrValRange);
+            }
+        }
+
+        Err(Status::Normal)
```
StevenJiang1110: This is not really an error case — it simply means no semaphore operation is blocking, right?

Instead of returning an error, I would suggest adding a new variant to `PendingBlocker`, so that this becomes `Ok(PendingBlocker::NotBlock)`, which would be clearer:

```rust
pub enum PendingBlocker {
     Zero(usize), 
     Decrease(usize),
     NotBlock,
}
```

Another benefit of this approach is that we could revert the function signature back to returning `core::result::Result<PendingBlocker, Error>`, eliminating the need to keep the `ErrWouldBlock` and `ErrValRange` variants in `Status`. I'm not entirely sure whether other parts of the code require these two variants to remain in `Status`, but it would be preferable to remove them if possible, as they feel somewhat out of place there.

StevenJiang1110 (approved):
Thanks for your excellent work! The PR looks good to me. Only a few minor comments.

test/initramfs/src/conformance/gvisor/blocklists/semaphore_test
lrh2000: It was caused because I deleted the file and regenerated it from scratch. Sorry, I think we can re-add it...

kernel/src/ipc/semaphore/system_v/sem.rs
lrh2000: > I'm not entirely sure whether other parts of the code require these two variants to remain in `Status`

It is needed because we need to prograte the error after the thread gets blocked...

Consider the following case:
```
 [ Thread A ]                           [ Thread B ]

 semop()

   wait()
     => blocked

                                         semctl(SETVAL)

                                             peform_semop() = ErrValRange

                                             wake(Thread A)

   status.load() = ErrValRange

   return Err(ERANGE)
```

We need to pass the error code from Thread B to Thread A to explain why the previously blocked semaphore operation is failing. This is passed as an atomic value, so the fat `Error` type is not appropriate.

So, we need to store the error type in the `Status` type. That seems a bit odd to me, either, but I think the strategy makes sense in general.

> This is not really an error case — it simply means no semaphore operation is blocking, right?
>
> Instead of returning an error, I would suggest adding a new variant to PendingBlocker, so that this becomes Ok(PendingBlocker::NotBlock), which would be clearer:

Thanks for the suggestion! It makes sense, and I will try it out to see if it works.

kernel/src/ipc/ipc_ns.rs
lrh2000: This is meant to prevent the `as u32` cast in the next line from dropping bits:
```rust
            assert!(SEMMNI <= u32::MAX as usize); 
            IpcId::new(SEMMNI as u32)
```

Otherwise, if `SEMMNI == (1usize << 32) + 1`, we will end up with a `IpcId::new(1)`...

Also note that `SEMMNI.try_into().unwrap()` cannot be used here because trait methods cannot be called during constant evaluation. So, I manually added an assertion to fail the complication if it occurs.

cqs21 (approved):

PR #3201, opened 2026-05-06
Fix book links to use directory-style chapter paths

cqs21:
This PR normalizes mdBook links for README-based chapters to use directory-style paths.

The main purpose is to make these links work not only in the generated mdBook site, but also when browsing the documentation directly in the GitHub repository. With this change, clicking those links in GitHub opens the corresponding document instead of landing on an error page.


tatetian (approved):

PR #3203, opened 2026-05-08
Check `has_pending` before entering user mode

vvvvsv:
Required for #2323.

This avoids a race condition where a new signal arrives after `handle_pending_signal` but before `user_mode.execute`. The current main branch's implementation returns directly to user space, and the new signal will not be handled until the next trap occurs.

The implementation in this PR still has a similar race condition, but the race window is reduced.

### DIFF
--- a/ostd/src/arch/x86/cpu/context/mod.rs
+++ b/ostd/src/arch/x86/cpu/context/mod.rs
@@ -273,6 +273,10 @@ impl UserContextApiInternal for UserContext {
 
         // Return when it is syscall or cpu exception type is Fault or Trap.
         loop {
+            if has_kernel_event() {
+                break ReturnReason::KernelEvent;
+            }
+
             crate::task::scheduler::might_preempt();
             self.user_context.run();
 
@@ -311,10 +315,6 @@ impl UserContextApiInternal for UserContext {
                     );
                     crate::arch::irq::enable_local();
                 }
-            }
-
-            if has_kernel_event() {
-                break ReturnReason::KernelEvent;
             }
         }
     }


lrh2000:
> This avoids a race condition where a new signal arrives after `handle_pending_signal` but before `user_mode.execute`. The current main branch's implementation returns directly to user space, and the new signal will not be handled until the next trap occurs.
> 
> The implementation in this PR still has a similar race condition, but the race window is reduced.

A signal delivered while a program is running in userspace won't be handled until the next trap. I believe this is standard Linux behavior. This is not a race, so the PR description is incorrect.

> Required for https://github.com/asterinas/asterinas/issues/2323.

You need to explain more on why it helps. Again, I don't really think the description makes any sense.

> ostd/src/arch/x86/cpu/context/mod.rs

Also, you are only changing the code for the x86 architecture. Please check the code for all the other architectures as well.

tatetian:
I think that the proposed change is wrong.

For our discussion, the main loop of a user task (in `create_new_user_task`) can be simplified to the following form:

```rust
while !current_thread.is_exited() {
   let return_reason = user_mode.execute(has_kernel_event_fn);

    match return_reason {
        ReturnReason::UserException => {
            handle_exception(&ctx, user_ctx, exception)
        }
        }
        ReturnReason::KernelEvent => {}
    };   

   handle_pending_signal(user_ctx, &ctx, pre_syscall_ret);
}
```

If a signal has set a user-defined signal handler, then `handle_pending_signal` modifies `user_ctx` so that the subsequent `user_mode.execute` transfers the control of the CPU to the userspace signal handler.

This relies on the fact that `user_mode.execute` is guaranteed to jump to the userspace **at least once** with `user_ctx` whose value has been set to jump to the userspace signal handler, regardless of the return value of `has_kernel_event_fn`.

But the proposal change breaks this assumption. If there is another pending signal, then `user_mode.execute` would immediately return, without going to the userspace to handle the delivered signal.

tatetian:
@vvvvsv Do you have any update on this PR? Do you feel it is correct?

vvvvsv:
> @vvvvsv Do you have any update on this PR? Do you feel it is correct?

In my previous testing, applying this patch made a multi-threaded GDB session work correctly. 

However, according to the discussion in the PR, this does not appear to be a fundamental bug. I need more time to figure out the root cause of the bug. I will re-open the PR if I find that this patch is indeed required.

PR #3216, opened 2026-05-12
Make tmpfs visible to statfs-based filesystem probes

Fischer0522:
## Overview

This PR creates a dedicated `SuperBlock` for tmpfs instead of reusing ramfs’s `SuperBlock`.  
And it fixes `statfs` to query the filesystem of the mount point rather than that of the underlying inode.  
For tmpfs, inodes belong to ramfs, so the old `statfs` returned the wrong superblock.
With this change, tmpfs becomes visible to `statfs`-based filesystem probes.

This is needed for tmpfs support in xfstests.
Previously, tmpfs in Asterinas was a thin wrapper around `RamFs`. Because tmpfs reused ramfs inodes, `statfs()` returned information about the inner ramfs instead of the mounted tmpfs.  
As a result:
- `df -T -P` did not list tmpfs as a normal filesystem.
- xfstests `_fs_type tmpfs_test` returned empty.

This made xfstests believe `TEST_DEV=tmpfs_test` was not mounted, so it mounted tmpfs again on the same target. The duplicate stacked mount caused `findmnt -S tmpfs_test` to return duplicate rows, breaking xfstests mount validation.

Now, tmpfs is properly visible via `statfs`, so xfstests detects the existing mount and no longer remounts it.

## TODO 

`TmpFs::sb()` still contains a TODO because this is a workaround.  
Tmpfs currently wraps `RamFs`, and its inodes still belong to the inner ramfs instance. We override `statfs`‑visible fields in `TmpFs::sb()` to preserve normal `RamFs::new()` semantics.

A cleaner long‑term fix is to implement tmpfs as its own independent filesystem, rather than teaching ramfs about tmpfs‑specific behavior.

### DIFF
--- a/kernel/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/src/fs/fs_impls/ramfs/fs.rs
@@ -22,6 +22,7 @@ use crate::{
         file::{AccessMode, FileIo, InodeMode, InodeType, Permission, StatusFlags, mkmod},
         pipe::Pipe,
         pseudofs::AnonDeviceId,
+        tmpfs::{self, TMPFS_MAGIC},
         utils::{CStr256, DirentVisitor},
         vfs::{
             file_system::{FileSystem, FsEventSubscriberStats, SuperBlock},
@@ -61,13 +62,41 @@ impl RamFs {
         Self::new_internal("rootfs")
     }
 
+    // TODO: Remove this tmpfs-specific constructor once `TmpFs` no longer
+    // aliases `RamFs`.
+    pub(in crate::fs) fn new_tmpfs() -> Arc<Self> {
+        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for tmpfs");
+        let sb = {
+            let mut super_block =
+                SuperBlock::new(TMPFS_MAGIC, BLOCK_SIZE, NAME_MAX, anon_device_id.id());
+            let max_blocks = tmpfs::default_max_blocks();
+            let max_inodes = tmpfs::default_max_inodes();
+            super_block.blocks = max_blocks;
+            super_block.bfree = max_blocks;
+            super_block.bavail = max_blocks;
+            super_block.files = max_inodes;
+            super_block.ffree = max_inodes;
+            super_block
+        };
+        Self::new_internal_with_sb("tmpfs", anon_device_id, sb)
+    }
+
     fn new_internal(name: &'static str) -> Arc<Self> {
         let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for ramfs");
+        let sb = SuperBlock::new(RAMFS_MAGIC, BLOCK_SIZE, NAME_MAX, anon_device_id.id());
+        Self::new_internal_with_sb(name, anon_device_id, sb)
+    }
+
+    fn new_internal_with_sb(
+        name: &'static str,
+        anon_device_id: AnonDeviceId,
+        sb: SuperBlock,
+    ) -> Arc<Self> {
         let root_dev_id = anon_device_id.id();
         Arc::new_cyclic(move |weak_fs| Self {
             name,
             _anon_device_id: anon_device_id,
-            sb: SuperBlock::new(RAMFS_MAGIC, BLOCK_SIZE, NAME_MAX, root_dev_id),
+            sb,
             root: Arc::new_cyclic(|weak_root| RamInode {
                 inner: Inner::new_dir(weak_root.clone(), weak_root.clone()),
                 metadata: SpinLock::new(InodeMeta::new_dir(
--- a/kernel/src/fs/fs_impls/tmpfs/fs.rs
+++ b/kernel/src/fs/fs_impls/tmpfs/fs.rs
@@ -4,8 +4,7 @@ use crate::{
     fs::{
         ramfs::RamFs,
         vfs::{
-            file_system::{FileSystem, FsEventSubscriberStats, SuperBlock},
-            inode::Inode,
+            file_system::FileSystem,
             registry::{FsCreationCtx, FsProperties, FsType},
         },
     },
@@ -14,42 +13,20 @@ use crate::{
 
 /// The temporary file system (tmpfs) structure.
 //
-// TODO: Currently, tmpfs is implemented as a thin wrapper around ramfs.
-// In the future we need to implement tmpfs-specific features such as
-// memory limits and swap support.
-pub struct TmpFs {
-    inner: Arc<RamFs>,
+// TODO: `TmpFs` currently aliases `RamFs` and relies on `RamFs::new_tmpfs()`
+// to create tmpfs-flavored ramfs instances. In the future we need to
+// implement a dedicated tmpfs with tmpfs-specific features such as memory
+// limits and swap support.
+pub type TmpFs = RamFs;
+
+// FIXME: These defaults are only a rough approximation for tmpfs-over-ramfs.
+// A dedicated tmpfs implementation should replace them with real tmpfs limit
+// and accounting semantics.
+pub(in crate::fs) fn default_max_blocks() -> usize {
+    crate::vm::mem_total() / PAGE_SIZE / 2
 }
-
-impl TmpFs {
-    pub fn new() -> Arc<Self> {
-        Arc::new(TmpFs {
-            inner: RamFs::new(),
-        })
-    }
-}
-
-impl FileSystem for TmpFs {
-    fn name(&self) -> &'static str {
-        "tmpfs"
-    }
-
-    fn sync(&self) -> Result<()> {
-        // do nothing
-        Ok(())
-    }
-
-    fn root_inode(&self) -> Arc<dyn Inode> {
-        self.inner.root_inode()
-    }
-
-    fn sb(&self) -> SuperBlock {
-        self.inner.sb()
-    }
-
-    fn fs_event_subscriber_stats(&self) -> &FsEventSubscriberStats {
-        self.inner.fs_event_subscriber_stats()
-    }
+pub(in crate::fs) fn default_max_inodes() -> usize {
+    crate::vm::mem_total() / PAGE_SIZE / 2
 }
 
 pub(super) struct TmpFsType;
@@ -64,7 +41,7 @@ impl FsType for TmpFsType {
     }
 
     fn create(&self, _fs_creation_ctx: &FsCreationCtx) -> Result<Arc<dyn FileSystem>> {
-        Ok(TmpFs::new())
+        Ok(TmpFs::new_tmpfs())
     }
 
     fn sysnode(&self) -> Option<Arc<dyn aster_systree::SysNode>> {
--- a/kernel/src/fs/fs_impls/ramfs/memfd.rs
+++ b/kernel/src/fs/fs_impls/ramfs/memfd.rs
@@ -321,7 +321,7 @@ impl MemfdTmpFs {
     pub(self) fn singleton() -> &'static Arc<TmpFs> {
         static MEMFD_TMPFS: Once<Arc<TmpFs>> = Once::new();
 
-        MEMFD_TMPFS.call_once(TmpFs::new)
+        MEMFD_TMPFS.call_once(TmpFs::new_tmpfs)
     }
 
     pub(self) fn new_path(memfd_inode: Arc<MemfdInode>) -> Path {
--- a/kernel/src/fs/fs_impls/tmpfs/mod.rs
+++ b/kernel/src/fs/fs_impls/tmpfs/mod.rs
@@ -2,13 +2,12 @@
 
 //! Temporary file system (tmpfs) based on ramfs.
 
-pub(super) use fs::TmpFs;
 use fs::TmpFsType;
+pub(super) use fs::{TmpFs, default_max_blocks, default_max_inodes};
 
 mod fs;
 
-#[expect(dead_code)]
-const TMPFS_MAGIC: u64 = 0x0102_1994;
+pub(super) const TMPFS_MAGIC: u64 = 0x0102_1994;
 
 pub(super) fn init() {
     crate::fs::vfs::registry::register(&TmpFsType).unwrap();


kernel/src/fs/fs_impls/tmpfs/fs.rs
```diff
@@ -44,14 +45,35 @@ impl FileSystem for TmpFs {
     }
 
     fn sb(&self) -> SuperBlock {
-        self.inner.sb()
+        // TODO: Tmpfs currently wraps RamFs, whose inodes still point to the
+        // inner RamFs. Override statfs-visible fields here as a workaround
+        // rather than changing the semantics of ordinary RamFs instances.
```
cchanging:
```suggestion
        // TODO: `Tmpfs` currently wraps `RamFs`, whose inodes still point to the
        // inner `RamFs`. Override statfs-visible fields here as a workaround
        // rather than changing the semantics of ordinary `RamFs` instances.
```

kernel/src/fs/fs_impls/tmpfs/fs.rs
```diff
@@ -44,14 +45,35 @@ impl FileSystem for TmpFs {
...
     }
 }
 
+fn default_max_blocks() -> usize {
+    crate::vm::mem_total() / PAGE_SIZE / 2
+}
+
+fn default_max_inodes() -> usize {
+    crate::vm::mem_total() / PAGE_SIZE / 2
+}
```
cchanging: If the default value is aligned with Linux, you can add a reference for it; If the default value is only a workaround, you should add a `TODO` for it.

cchanging (commented):
Thanks for fixing the `statfs` display for `Tmpfs`! I only raised some minor issues

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -100,6 +100,11 @@ impl Path {
         &self.dentry
     }
 
+    // Gets the file system of current `Path`.
+    pub fn fs(&self) -> &Arc<dyn FileSystem> {
+        self.mount.fs()
```
lrh2000: Is it really a good idea to make `path.fs()` differ from `inode.fs()`?

Why not make the ramfs superblock configurable? For now, it seems like a better idea to have something like `RamFs::new_tmpfs()`, `RamFs::new_with_sb()`, or whatever to fix the real super block instead of stacking an incomplete shim.

cchanging: > Is it really a good idea to make path.fs() differ from inode.fs()?

I think if we didn’t have the current setup where `Ramfs` is used as a dummy implementation of `Tmpfs`, then the filesystem obtained from the `mount` should always be the same as the one obtained from the `inode`. In that case, getting the filesystem from the `mount` would avoid one `Weak::upgrade` compared with going through the inode path.

However, in the current case, if we obtain it from the inode for a `Tmpfs` `Path`, we end up getting the `Ramfs` instance, which causes the problem. The mismatch here is not an issue with this API, but rather with our `Tmpfs` implementation.

> Why not make the ramfs superblock configurable? For now, it seems like a better idea to have something like RamFs::new_tmpfs(), RamFs::new_with_sb(), or whatever to fix the real super block instead of stacking an incomplete shim.

I think this would indeed work as well. We could also add a `TODO` comment to such API and remove it after `Tmpfs` is fully implemented.

lrh2000: > The mismatch here is not an issue with this API, but rather with our `Tmpfs` implementation.

Yes. I think something like `TmpFs(Arc<RamFs>)` is fundamentally wrong and we shoud just use:
```rust
let tmp_fs = RamFs::new_tmpfs();
```
or
```rust
let tmp_fs = RamFs::new_with_sb(tmpfs_sb);
```

Otherwise `inode.fs()` reports `Arc<TmpFs>` but `path.mount` stores `Arc<RamFs>`, which are two objects, despite that the associated methods may function identically.

---

Also this method should use doc comments instead of normal comments even if you're going to add it, right?
```patch
-    // Gets the file system of current `Path`.
+    /// Gets the file system of current `Path`.
```

cchanging (approved):

kernel/src/fs/fs_impls/tmpfs/fs.rs
```diff
@@ -14,42 +13,21 @@ use crate::{
...
+// TODO: `TmpFs` currently aliases `RamFs` and relies on `RamFs::new_tmpfs()`
+// to create tmpfs-flavored ramfs instances. In the future we need to
+// implement a dedicated tmpfs with tmpfs-specific features such as memory
+// limits and swap support.
+pub type TmpFs = RamFs;
+
+// FIXME: These defaults are only a rough approximation for tmpfs-over-ramfs.
+// A dedicated tmpfs implementation should replace them with real tmpfs limit
+// and accounting semantics.
+pub(in crate::fs) fn default_max_blocks() -> usize {
```
lrh2000, line 25: I think that if you `pub(in crate::fs)` a method that constructs a tmpfs super block, then all the visibility markers of `default_max_blocks`, `default_max_inodes`, and `TMPFS_MAGIC` can be removed, and we will end up with more self-contained APIs.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -61,6 +62,43 @@ impl RamFs {
...
+                this: weak_root.clone(),
+                fs: weak_fs.clone(),
+                container_dev_id: root_dev_id,
+                extension: Extension::new(),
+                xattr: RamXattr::new(),
+            }),
+            inode_allocator: AtomicU64::new(ROOT_INO + 1),
+            fs_event_subscriber_stats: FsEventSubscriberStats::new(),
+        })
+    }
```
lrh2000, line 82: This method has quite a few duplicate lines with `new_internal`, so can we figure out a way to avoid duplication? One option would be to create a method that accepts the name and the super block. Then, the old `new_internal` implementation could delegate to it.

Of course, we need to decide on proper names for these methods. Simple choices are `new_with_name` and `new_with_name_and_sb`. I don't know if there are better options at the moment.

lrh2000 (commented):
The code looks acceptable. Thanks for addressing the previous comments. I leave two small nits that hope you can address as well.

kernel/src/fs/fs_impls/ramfs/fs.rs
Fischer0522: I tried several methods, but all of them do damage to `new_interval`. If we add `tmpfs::create_super_block`, it will allocate `anon_device_id` but the `SuperBlock` doesn't keep it. If we allocate again in `new_internal`, the `anon_device_id` diverges between `SuperBlock` and `RamFs` (we can not create `AnonDeviceId` from the `device_id` in `SuperBlock` because `AnonDeviceId::new(id: DeviceId)` is private).

So I think accepting the super block doesn't work well. If we want to avoid this duplication, the new method might look like: `new_internal(name: &'static str, device_id: AnonDeviceId)`, but the duplication becomes each method needs to create `AnonDeviceId` on its own. And since the `TmpFs` will be refactored, I think it is acceptable to keep such duplication for now.

```rust
   pub(in crate::fs) fn new_tmpfs() -> Arc<Self> {
        // Allocate the `anon_device_id` in it.
        let super_block = tmpfs::create_super_block();
        Self::new_internal("tmpfs", Some(super_block))
    }

    fn new_internal(name: &'static str, sb: Option<SuperBlock>) -> Arc<Self> {
        // the anon_device_id is different from the id kept in `SuperBlock`
        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for ramfs");
        let root_dev_id = anon_device_id.id();
        Arc::new_cyclic(move |weak_fs| Self {
            name,
            _anon_device_id: anon_device_id,
            sb: sb.unwrap_or_else(||SuperBlock::new(RAMFS_MAGIC, BLOCK_SIZE, NAME_MAX, root_dev_id)),
            root: Arc::new_cyclic(|weak_root| RamInode {
                inner: Inner::new_dir(weak_root.clone(), weak_root.clone()),
                metadata: SpinLock::new(InodeMeta::new_dir(
                    mkmod!(a+rx, u+w),
                    Uid::new_root(),
                    Gid::new_root(),
                )),
                ino: ROOT_INO,
                typ: InodeType::Dir,
                this: weak_root.clone(),
                fs: weak_fs.clone(),
                container_dev_id: root_dev_id,
                extension: Extension::new(),
                xattr: RamXattr::new(),
            }),
            inode_allocator: AtomicU64::new(ROOT_INO + 1),
            fs_event_subscriber_stats: FsEventSubscriberStats::new(),
        })
    }
```

Or if we want completely to remove the duplication and not allocate `device_id` in `tmpfs::create_super_block`, the `new_internal` will look like:
```rust
    fn new_internal(name: &'static str, super_block: Option<SuperBlock>) -> Arc<Self> {
        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for ramfs");
        let root_dev_id = anon_device_id.id();
        let super_block = super_block
            .map(|mut super_block| {
                super_block.container_dev_id = root_dev_id;
                super_block
            })
            .unwrap_or_else(|| SuperBlock::new(RAMFS_MAGIC, BLOCK_SIZE, NAME_MAX, root_dev_id));
        })
         // ...
    }
```

lrh2000: Thank you for your detailed explanation! I got the complexity now...

So how about:
```patch
diff --git a/kernel/src/fs/fs_impls/ramfs/fs.rs b/kernel/src/fs/fs_impls/ramfs/fs.rs
index 8bf6219fd..a65ec8773 100644
--- a/kernel/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/src/fs/fs_impls/ramfs/fs.rs
@@ -61,50 +61,41 @@ impl RamFs {
         Self::new_internal("rootfs")
     }
 
+    fn new_internal(name: &'static str) -> Arc<Self> {
+        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for ramfs");
+        let sb = SuperBlock::new(RAMFS_MAGIC, BLOCK_SIZE, NAME_MAX, anon_device_id.id());
+        Self::new_internal_with_sb(name, anon_device_id, sb)
+    }
+
     // TODO: Remove this tmpfs-specific constructor once `TmpFs` no longer
     // aliases `RamFs`.
     pub(in crate::fs) fn new_tmpfs() -> Arc<Self> {
-        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for ramfs");
-        let root_dev_id = anon_device_id.id();
-        let mut super_block = SuperBlock::new(TMPFS_MAGIC, BLOCK_SIZE, NAME_MAX, root_dev_id);
-        let max_blocks = tmpfs::default_max_blocks();
-        let max_inodes = tmpfs::default_max_inodes();
-        super_block.blocks = max_blocks;
-        super_block.bfree = max_blocks;
-        super_block.bavail = max_blocks;
-        super_block.files = max_inodes;
-        super_block.ffree = max_inodes;
-        Arc::new_cyclic(move |weak_fs| Self {
-            name: "tmpfs",
-            _anon_device_id: anon_device_id,
-            sb: super_block,
-            root: Arc::new_cyclic(|weak_root| RamInode {
-                inner: Inner::new_dir(weak_root.clone(), weak_root.clone()),
-                metadata: SpinLock::new(InodeMeta::new_dir(
-                    mkmod!(a+rx, u+w),
-                    Uid::new_root(),
-                    Gid::new_root(),
-                )),
-                ino: ROOT_INO,
-                typ: InodeType::Dir,
-                this: weak_root.clone(),
-                fs: weak_fs.clone(),
-                container_dev_id: root_dev_id,
-                extension: Extension::new(),
-                xattr: RamXattr::new(),
-            }),
-            inode_allocator: AtomicU64::new(ROOT_INO + 1),
-            fs_event_subscriber_stats: FsEventSubscriberStats::new(),
-        })
+        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for tmpfs");
+        let sb = {
+            let mut super_block =
+                SuperBlock::new(TMPFS_MAGIC, BLOCK_SIZE, NAME_MAX, anon_device_id.id());
+            let max_blocks = tmpfs::default_max_blocks();
+            let max_inodes = tmpfs::default_max_inodes();
+            super_block.blocks = max_blocks;
+            super_block.bfree = max_blocks;
+            super_block.bavail = max_blocks;
+            super_block.files = max_inodes;
+            super_block.ffree = max_inodes;
+            super_block
+        };
+        Self::new_internal_with_sb("tmpfs", anon_device_id, sb)
     }
 
-    fn new_internal(name: &'static str) -> Arc<Self> {
-        let anon_device_id = AnonDeviceId::acquire().expect("no device ID is available for ramfs");
+    fn new_internal_with_sb(
+        name: &'static str,
+        anon_device_id: AnonDeviceId,
+        sb: SuperBlock,
+    ) -> Arc<Self> {
         let root_dev_id = anon_device_id.id();
         Arc::new_cyclic(move |weak_fs| Self {
             name,
             _anon_device_id: anon_device_id,
-            sb: SuperBlock::new(RAMFS_MAGIC, BLOCK_SIZE, NAME_MAX, root_dev_id),
+            sb,
             root: Arc::new_cyclic(|weak_root| RamInode {
                 inner: Inner::new_dir(weak_root.clone(), weak_root.clone()),
                 metadata: SpinLock::new(InodeMeta::new_dir(
diff --git a/kernel/src/fs/fs_impls/tmpfs/fs.rs b/kernel/src/fs/fs_impls/tmpfs/fs.rs
index b38aaca31..ce44624b0 100644
--- a/kernel/src/fs/fs_impls/tmpfs/fs.rs
+++ b/kernel/src/fs/fs_impls/tmpfs/fs.rs
@@ -25,7 +25,6 @@ pub type TmpFs = RamFs;
 pub(in crate::fs) fn default_max_blocks() -> usize {
     crate::vm::mem_total() / PAGE_SIZE / 2
 }
-
 pub(in crate::fs) fn default_max_inodes() -> usize {
     crate::vm::mem_total() / PAGE_SIZE / 2
 }
```

 - Remove the blank line in `tmpfs/fs.rs` because I assume that the TODO comments apply to both functions.
 - I found that `tmpfs::new_superblock` isn't a valid suggestion since the superblock construction needs to access some ramfs constants in our current implementation. So my previous comment can be ignored.. Sorry for the confusion, if any.

Fischer0522: Fixed, I think this might be the best solution.

lrh2000 (approved):

PR #3217, opened 2026-05-12
sysfs: Implement `cpu_byteorder` and `address_bits` attributes

taosue:
Add two attributes to `/sys/kernel`:                                                                                     
- cpu_byteorder: reports the endianness of the running kernel   
- address_bits: reports the kernel address size in bits

### DIFF
--- a/kernel/src/fs/fs_impls/sysfs/kernel.rs
+++ b/kernel/src/fs/fs_impls/sysfs/kernel.rs
@@ -1,11 +1,24 @@
 // SPDX-License-Identifier: MPL-2.0
 
+//! Implementation of the `/sys/kernel` sysfs directory.
+//!
+//! This module provides the `/sys/kernel` directory in sysfs, which contains
+//! kernel-specific attributes and information. Currently implemented attributes:
+//!
+//! - `cpu_byteorder`: The endianness of the running kernel ("little" or "big")
+//! - `address_bits`: The address size of the running kernel in bits
+//!
+//! These attributes follow the Linux kernel sysfs specification:
+//! - [cpu_byteorder](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-kernel-cpu_byteorder)
+//! - [address_bits](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-kernel-address_bits)
+
 use alloc::sync::Arc;
 
 use aster_systree::{
     BranchNodeFields, Error, Result, SysAttrSetBuilder, SysBranchNode, SysNode, SysPerms, SysStr,
     inherit_sys_branch_node,
 };
+use aster_util::printer::VmPrinter;
 use inherit_methods_macro::inherit_methods;
 use ostd::mm::{VmReader, VmWriter};
 use spin::Once;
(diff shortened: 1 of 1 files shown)


lrh2000:
Are the changes from `AttributeError` to `NotFound` intentional? I can't find an explanation for it. As far as I know, cgroupfs uses `AttributeError` here, so I'm not sure why you changed it. Should `read`/`write` ever return `ENOENT`?

https://github.com/asterinas/asterinas/blob/035226075a6219713c374b9af51f4fe0b4df1691/kernel/src/fs/fs_impls/cgroupfs/controller/cpuset.rs#L49-L50

Also I think some removed TODOs should be added back:
```rust
        let name = SysStr::from("kernel");

        let mut builder = SysAttrSetBuilder::new();
        builder.add(
            SysStr::from("cpu_byteorder"),
            SysPerms::DEFAULT_RO_ATTR_PERMS,
        );
        builder.add(
            SysStr::from("address_bits"),
            SysPerms::DEFAULT_RO_ATTR_PERMS,
        );
        // TODO: Add more kernel-specific attributes.
        let attrs = builder
            .build()
            .expect("Failed to build kernel attribute set");

        Arc::new_cyclic(|weak_self| {
            let fields = BranchNodeFields::new(name, attrs, weak_self.clone());
            KernelSysNodeRoot { fields }
        })
```

```rust
            "address_bits" => {
                writeln!(printer, "{}", usize::BITS)?;
                Ok(printer.bytes_written())
            }
            // TODO: Add support for reading other attributes.
            _ => Err(Error::NotFound),
```

(Adjust some blank lines as well since I don't know why they're arranged in the current form...)

taosue:
> Are the changes from `AttributeError` to `NotFound` intentional?I can't find an explanation for it. As far as I know, cgroupfs uses `AttributeError` here, so I'm not sure why you changed it. Should `read`/`write` ever return `ENOENT`?
> 

Thanks for the review! The change from `AttributeError` to `NotFound` was following the pattern in `tsm_mr.rs`:                        
https://github.com/asterinas/asterinas/blob/035226075a6219713c374b9af51f4fe0b4df1691/kernel/src/security/tsm_mr.rs#L212

But you're right, `AttributeError` is the correct choice here (I will submit a PR to fix `tsm_mr.rs` soon). I checked the Linux implementation:

Linux sysfs `read`/`write` can return `-ENOENT`, but it's very rare and only as specific callback business logic — e.g., [`uevent_helper_store`](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/kernel/ksysfs.c#L74-L75) returns `-ENOENT` when the input exceeds `UEVENT_HELPER_PATH_LEN`. This is not a framework-level behavior.

For our case where the attribute name doesn't match any arm in the match block, it may be semantically equivalent to `kattr->show ==NULL` in Linux's [`kobj_attr_show`](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/lib/kobject.c#L820):

```c
  static ssize_t kobj_attr_show(struct kobject *kobj, struct attribute *attr, char *buf)
  {                                                                                                                                
      struct kobj_attribute *kattr;
      ssize_t ret = -EIO;                                                                                                          
                            
      kattr = container_of(attr, struct kobj_attribute, attr);
      if (kattr->show)                         
          ret = kattr->show(kobj, kattr, buf);
      return ret;
  }
```

Linux returns `-EIO` when the callback is missing, which corresponds to our `Error::AttributeError`.

tatetian (approved):

PR #3218, opened 2026-05-12
Suppress find error in ASTER_SCML when directory is missing

taosue:
Suppress the find error in the Dockerfile when the `/root/asterinas/book/src/kernel/linux-compatibility/` directory does not exist. Some Docker images (e.g. Kata or CoCo) are built on top of the Asterinas development image but used by end users who don't mount the Asterinas repository. In such cases, the find command in `.bashrc` produces a confusing error message on every shell startup:
```
find: '/root/asterinas/book/src/kernel/linux-compatibility/': No such file or directory
```
Redirect `find`'s stderr to `/dev/null` so that ASTER_SCML is silently set to an empty string when the directory is absent.


tatetian (approved):

PR #3219, opened 2026-05-12
Fix reachable unwrap panic in mount ID allocator

### LINKED ISSUE #3205: Reachable unwrap panic caused by `mount` ID allocator
### Describe the bug

There is a reachable unwrap panic in `Mount::new` when calling system call `mount` lots of times.

https://github.com/asterinas/asterinas/blob/c884afe23aac6e3934647520396a207e3f640ccf/kernel/src/fs/vfs/path/mount.rs#L238

The allocator is initialized with a fixed capacity
https://github.com/asterinas/asterinas/blob/c884afe23aac6e3934647520396a207e3f640ccf/kernel/src/fs/vfs/path/mount.rs#L57

### To Reproduce
Compile and run:
```C
#define _GNU_SOURCE

#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mount.h>
#include <sys/stat.h>
#include <unistd.h>

static const char *make_base_dir(void) {
  static char base[128];

  snprintf(base, sizeof(base), "/tmp/aster-mount-id-poc-%ld", (long)getpid());
  if (mkdir(base, 0700) == 0 || errno == EEXIST) {
    return base;
  }

  snprintf(base, sizeof(base), "aster-mount-id-poc-%ld", (long)getpid());
  if (mkdir(base, 0700) == 0 || errno == EEXIST) {
    return base;
  }

  fprintf(stderr, "failed to create base directory: %s\n", strerror(errno));
  return NULL;
}

int main(void) {
  const char *base = make_base_dir();
  if (base == NULL) {
    return 1;
  }

  for (unsigned long i = 0;; i++) {
    char target[256];
    snprintf(target, sizeof(target), "%s/m%05lu", base, i);

    if (mkdir(target, 0700) != 0 && errno != EEXIST) {
      fprintf(stderr, "mkdir(%s) failed after %lu mounts: %s\n", target, i,
              strerror(errno));
      return 1;
    }

    if (mount("none", target, "ramfs", 0, NULL) != 0) {
      fprintf(stderr, "mount(%s) failed after %lu mounts: %s\n", target, i,
              strerror(errno));
      return 1;
    }

    if ((i & 0xffUL) == 0) {
      printf("created %lu ramfs mounts under %s\n", i + 1, base);
    }
  }
}
```

### Expected behavior

Asterinas shall not crash.

### Environment

- Official docker asterinas/asterinas:0.17.0-20251217
- Asterinas version: main c884afe23aac6e3934647520396a207e3f640ccf

### Logs
```
~ # /etc/mount_id_alloc_panic.c.bin 
created 1 ramfs mounts under /tmp/aster-mount-id-poc-10
created 257 ramfs mounts under /tmp/aster-mount-id-poc-10
created 513 ramfs mounts under /tmp/aster-mount-id-poc-10
created 769 ramfs mounts under /tmp/aster-mount-id-poc-10
created 1025 ramfs mounts under /tmp/aster-mount-id-poc-10
created 1281 ramfs mounts under /tmp/aster-mount-id-poc-10
created 1537 ramfs mounts under /tmp/aster-mount-id-poc-10
created 1793 ramfs mounts under /tmp/aster-mount-id-poc-10
created 2049 ramfs mounts under /tmp/aster-mount-id-poc-10
created 2305 ramfs mounts under /tmp/aster-mount-id-poc-10
created 2561 ramfs mounts under /tmp/aster-mount-id-poc-10
created 2817 ramfs mounts under /tmp/aster-mount-id-poc-10
created 3073 ramfs mounts under /tmp/aster-mount-id-poc-10
created 3329 ramfs mounts under /tmp/aster-mount-id-poc-10
created 3585 ramfs mounts under /tmp/aster-mount-id-poc-10
created 3841 ramfs mounts under /tmp/aster-mount-id-poc-10
created 4097 ramfs mounts under /tmp/aster-mount-id-poc-10
created 4353 ramfs mounts under /tmp/aster-mount-id-poc-10
created 4609 ramfs mounts under /tmp/aster-mount-id-poc-10
created 4865 ramfs mounts under /tmp/aster-mount-id-poc-10
created 5121 ramfs mounts under /tmp/aster-mount-id-poc-10
created 5377 ramfs mounts under /tmp/aster-mount-id-poc-10
created 5633 ramfs mounts under /tmp/aster-mount-id-poc-10
created 5889 ramfs mounts under /tmp/aster-mount-id-poc-10
created 6145 ramfs mounts under /tmp/aster-mount-id-poc-10
created 6401 ramfs mounts under /tmp/aster-mount-id-poc-10
created 6657 ramfs mounts under /tmp/aster-mount-id-poc-10
created 6913 ramfs mounts under /tmp/aster-mount-id-poc-10
created 7169 ramfs mounts under /tmp/aster-mount-id-poc-10
created 7425 ramfs mounts under /tmp/aster-mount-id-poc-10
created 7681 ramfs mounts under /tmp/aster-mount-id-poc-10
created 7937 ramfs mounts under /tmp/aster-mount-id-poc-10
created 8193 ramfs mounts under /tmp/aster-mount-id-poc-10
created 8449 ramfs mounts under /tmp/aster-mount-id-poc-10
created 8705 ramfs mounts under /tmp/aster-mount-id-poc-10
created 8961 ramfs mounts under /tmp/aster-mount-id-poc-10
created 9217 ramfs mounts under /tmp/aster-mount-id-poc-10
created 9473 ramfs mounts under /tmp/aster-mount-id-poc-10
created 9729 ramfs mounts under /tmp/aster-mount-id-poc-10
[    38.120] ERROR : Uncaught panic:
        called `Option::unwrap()` on a `None` value
        at /root/asterinas/kernel/src/fs/vfs/path/mount.rs:238
        on CPU 0 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [1], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(0), real_time: RealTimeAttr { prio: 99, time_slice: 31490700 }, fair: FairAttr { weight: 1024, pending_weight: UnsafeCell { .. }, vruntime: 3154338150 } } })
```

avasummer:
This PR fixes a kernel panic caused by `.unwrap()` on mount ID allocation when the fixed-capacity ID pool is exhausted by repeated `mount` syscalls.

The two panic sites (`Mount::new` and `Mount::clone_mount`) now return `Errno::ENOMEM` instead of panicking, matching Linux's behavior in `fs/namespace.c`.

To allow `MountNamespace::new_clone` to propagate the error with `?`, `MountNamespace::root` is changed from `Arc<Mount>` to `Once<Arc<Mount>>`, replacing `Arc::new_cyclic` with a two-phase init pattern (allocate namespace first, then clone mount tree). This mirrors Linux's `alloc_mnt_ns` + `copy_tree` approach in `copy_mnt_ns`.

Closes #3205

### DIFF
--- a/kernel/src/fs/fs_impls/pseudofs/anon_inodefs.rs
+++ b/kernel/src/fs/fs_impls/pseudofs/anon_inodefs.rs
@@ -15,6 +15,11 @@ use crate::{
     process::{Gid, Uid},
 };
 
+pub(super) fn init() {
+    // Touch the mount node to trigger the initialization of all singletons.
+    let _ = AnonInodeFs::mount_node();
+}
+
 pub struct AnonInodeFs {
     _private: (),
 }
--- a/kernel/src/fs/fs_impls/pseudofs/anon_inodefs.rs
+++ b/kernel/src/fs/fs_impls/pseudofs/anon_inodefs.rs
@@ -40,7 +40,7 @@ impl AnonInodeFs {
     pub fn mount_node() -> &'static Arc<Mount> {
         static ANON_INODEFS_MOUNT: Once<Arc<Mount>> = Once::new();
 
-        ANON_INODEFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()))
+        ANON_INODEFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()).unwrap())
     }
 
     /// Returns the shared inode of the anonymous inode file system singleton.
--- a/kernel/src/fs/vfs/path/mount.rs
+++ b/kernel/src/fs/vfs/path/mount.rs
@@ -60,6 +60,16 @@ pub(super) fn init() {
     ID_ALLOCATOR.call_once(|| SpinLock::new(id_allocator));
 }
 
+/// Allocates a new mount ID from the global pool.
+fn alloc_mount_id() -> Result<usize> {
+    ID_ALLOCATOR
+        .get()
+        .unwrap()
+        .lock()
+        .alloc()
+        .ok_or_else(|| Error::with_message(Errno::ENOMEM, "mount ID pool exhausted"))
+}
+
 bitflags! {
     pub struct PerMountFlags: u32 {
         /// Mount read-only.
@@ -235,12 +245,7 @@ impl Mount {
         mnt_ns: Weak<MountNamespace>,
         source: Option<String>,
     ) -> Result<Arc<Self>> {
-        let id = ID_ALLOCATOR
-            .get()
-            .unwrap()
-            .lock()
-            .alloc()
-            .ok_or_else(|| Error::new(Errno::ENOMEM))?;
+        let id = alloc_mount_id()?;
 
         Ok(Arc::new_cyclic(|weak_self| Self {
             id,
@@ -331,12 +336,7 @@ impl Mount {
         root_dentry: &Arc<Dentry>,
         new_ns: &Weak<MountNamespace>,
     ) -> Result<Arc<Self>> {
-        let id = ID_ALLOCATOR
-            .get()
-            .unwrap()
-            .lock()
-            .alloc()
-            .ok_or_else(|| Error::new(Errno::ENOMEM))?;
+        let id = alloc_mount_id()?;
 
         Ok(Arc::new_cyclic(|weak_self| Self {
             id,
--- a/kernel/src/fs/vfs/path/mount.rs
+++ b/kernel/src/fs/vfs/path/mount.rs
@@ -207,7 +207,7 @@ impl Mount {
     pub(in crate::fs) fn new_root(
         fs: Arc<dyn FileSystem>,
         mnt_ns: Weak<MountNamespace>,
-    ) -> Arc<Self> {
+    ) -> Result<Arc<Self>> {
         let source = fs.name().to_string();
         Self::new(fs, PerMountFlags::default(), None, mnt_ns, Some(source))
     }
@@ -216,7 +216,7 @@ impl Mount {
     ///
     /// This pseudo mount is not mounted on other mount nodes, has no parent, and does not
     /// belong to any mount namespace.
-    pub(in crate::fs) fn new_pseudo(fs: Arc<dyn FileSystem>) -> Arc<Self> {
+    pub(in crate::fs) fn new_pseudo(fs: Arc<dyn FileSystem>) -> Result<Arc<Self>> {
         Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new(), None)
     }
 
@@ -234,10 +234,15 @@ impl Mount {
         parent_mount: Option<Weak<Mount>>,
         mnt_ns: Weak<MountNamespace>,
         source: Option<String>,
-    ) -> Arc<Self> {
-        let id = ID_ALLOCATOR.get().unwrap().lock().alloc().unwrap();
+    ) -> Result<Arc<Self>> {
+        let id = ID_ALLOCATOR
+            .get()
+            .unwrap()
+            .lock()
+            .alloc()
+            .ok_or_else(|| Error::new(Errno::ENOMEM))?;
 
-        Arc::new_cyclic(|weak_self| Self {
+        Ok(Arc::new_cyclic(|weak_self| Self {
             id,
             root_dentry: Dentry::new_root(fs.root_inode()),
             mountpoint: RwLock::new(None),
@@ -249,7 +254,7 @@ impl Mount {
             mnt_ns,
             flags: AtomicPerMountFlags::new(flags),
             this: weak_self.clone(),
-        })
+        }))
     }
 
     /// Gets the mount ID.
@@ -293,7 +298,7 @@ impl Mount {
             Some(Arc::downgrade(self)),
             self.mnt_ns.clone(),
             source,
-        );
+        )?;
         self.children.write().insert(key, child_mount.clone());
         child_mount.set_mountpoint(mountpoint);
 
@@ -321,9 +326,20 @@ impl Mount {
     /// have no parent and children. We should set the parent and children manually.
     ///
     /// The new mount will belong to the given mount namespace.
-    fn clone_mount(&self, root_dentry: &Arc<Dentry>, new_ns: &Weak<MountNamespace>) -> Arc<Self> {
-        Arc::new_cyclic(|weak_self| Self {
-            id: ID_ALLOCATOR.get().unwrap().lock().alloc().unwrap(),
+    fn clone_mount(
+        &self,
+        root_dentry: &Arc<Dentry>,
+        new_ns: &Weak<MountNamespace>,
+    ) -> Result<Arc<Self>> {
+        let id = ID_ALLOCATOR
+            .get()
+            .unwrap()
+            .lock()
+            .alloc()
+            .ok_or_else(|| Error::new(Errno::ENOMEM))?;
+
+        Ok(Arc::new_cyclic(|weak_self| Self {
+            id,
             root_dentry: root_dentry.clone(),
             mountpoint: RwLock::new(None),
             parent: RwLock::new(None),
@@ -334,7 +350,7 @@ impl Mount {
             mnt_ns: new_ns.clone(),
             flags: AtomicPerMountFlags::new(self.flags.load(Ordering::Relaxed)),
             this: weak_self.clone(),
-        })
+        }))
     }
 
     /// Clones a mount tree starting from the specified root `Dentry`.
@@ -356,10 +372,10 @@ impl Mount {
         new_ns: &Weak<MountNamespace>,
         recursive: bool,
         mnt_ns_file_copying: MountNsFileCopying,
-    ) -> Arc<Self> {
-        let new_root_mount = self.clone_mount(root_dentry, new_ns);
+    ) -> Result<Arc<Self>> {
+        let new_root_mount = self.clone_mount(root_dentry, new_ns)?;
         if !recursive {
-            return new_root_mount;
+            return Ok(new_root_mount);
         }
 
         let mut stack = vec![self.this()];
@@ -379,7 +395,7 @@ impl Mount {
                     continue;
                 }
                 let new_child_mount =
-                    old_child_mount.clone_mount(old_child_mount.root_dentry(), new_ns);
+                    old_child_mount.clone_mount(old_child_mount.root_dentry(), new_ns)?;
                 let key = mountpoint.key();
                 new_parent_mount
                     .children
@@ -392,7 +408,7 @@ impl Mount {
             }
         }
 
-        new_root_mount
+        Ok(new_root_mount)
     }
 
     /// Sets the propagation type of this mount.
@@ -615,6 +631,7 @@ impl Debug for Mount {
 
 impl Drop for Mount {
     fn drop(&mut self) {
+        self.clear_mountpoint();
         ID_ALLOCATOR.get().unwrap().lock().free(self.id);
     }
 }
--- a/kernel/src/fs/vfs/path/mount_namespace.rs
+++ b/kernel/src/fs/vfs/path/mount_namespace.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use alloc::sync::UniqueArc;
+
 use spin::Once;
 
 use super::{mount::MountNsFileCopying, try_get_mnt_ns_inode};
@@ -21,7 +23,13 @@ use crate::{
 /// rejected and return an `Err`.
 pub struct MountNamespace {
     /// The root mount of this namespace.
-    root: Arc<Mount>,
+    ///
+    /// This field is wrapped within an `Option<_>`
+    /// because the root mount is unknown
+    /// in the beginning of the constructor method (see `new_clone`).
+    /// But if the constructor method completes,
+    /// this field is guaranteed to be `Some(_)`.
+    root: Option<Arc<Mount>>,
     /// The user namespace that owns this mount namespace.
     owner: Arc<UserNamespace>,
     /// The stashed dentry in nsfs.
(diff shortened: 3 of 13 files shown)


kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -186,8 +184,11 @@ impl MountNamespace {
 // detached from their parents and cleared of their mountpoints.
 impl Drop for MountNamespace {
     fn drop(&mut self) {
+        let Some(root) = self.root.get() else {
+            return;
+        };
         let mut worklist = VecDeque::new();
```
tatetian, line 206: Add some comments and insert a blank line:

```rust
        let Some(root) = self.root.get() else {
            // The constructor must be incomplete
            // and thus the subsequent cleanup logic can be skipped.
            return;
        };

```

kernel/src/fs/fs_impls/pseudofs/anon_inodefs.rs
```diff
@@ -40,7 +40,7 @@ impl AnonInodeFs {
     pub fn mount_node() -> &'static Arc<Mount> {
         static ANON_INODEFS_MOUNT: Once<Arc<Mount>> = Once::new();
 
-        ANON_INODEFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()))
+        ANON_INODEFS_MOUNT.call_once(|| Mount::new_pseudo(Self::singleton().clone()).unwrap())
```
tatetian, line 48: We can't be sure when such lazy initialization might be called. If this is done after the kernel already gets up and running for a while, the mount ID may still be exhausted and cause panicking. So I suggest for each and every the anon inode FS modules under `pseoudofs`, add an `init` function and let the `init` function of the `pseoduofs` module invokes them.

```rust
// file: pseudofs.rs

pub(super) fn init() {
    crate::fs::vfs::registry::register(&PipeFsType).unwrap();
    crate::fs::vfs::registry::register(&SockFsType).unwrap();
    // Note: `AnonInodeFs` does not need to be registered in the FS registry.
    // Reference: <https://elixir.bootlin.com/linux/v6.16.5/A/ident/anon_inode_fs_type>

    anon_inodefs::init();
    nsfs::init();
    pidfdfs::init();
    pipefs::init();
    sockfs::init();
}
```

Each per-pseudofs init function calls the `call_once` of all static `ONCE` objects of the module to ensure their initialization eagerly, rather than the current lazy-initialization on first use.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -234,10 +234,15 @@ impl Mount {
         parent_mount: Option<Weak<Mount>>,
         mnt_ns: Weak<MountNamespace>,
         source: Option<String>,
-    ) -> Arc<Self> {
-        let id = ID_ALLOCATOR.get().unwrap().lock().alloc().unwrap();
-
-        Arc::new_cyclic(|weak_self| Self {
+    ) -> Result<Arc<Self>> {
+        let id = ID_ALLOCATOR
+            .get()
```
tatetian: **Two-phase namespace init duplication; same alloc block also duplicated** (significant): Two related cleanups land naturally together. First, the five-line `ID_ALLOCATOR.get().unwrap().lock().alloc().ok_or_else(|| Error::new(Errno::ENOMEM))?` block now appears verbatim in `Mount::new` (lines 238-243) and `Mount::clone_mount` (lines 334-339). Extracting a `fn alloc_mount_id() -> Result<usize>` keeps the error mapping in one place and shrinks both constructors to one line:

```rust
fn alloc_mount_id() -> Result<usize> {
    ID_ALLOCATOR
        .get()
        .unwrap()
        .lock()
        .alloc()
        .ok_or_else(|| Error::with_message(Errno::ENOMEM, "mount ID pool exhausted"))
}
```

Second, `MountNamespace::get_init_singleton` and `MountNamespace::new_clone` now share an identical four-step pattern: (1) allocate an `Arc<MountNamespace>` with `root: Once::new()`; (2) build a root mount weakly-referencing it; (3) `call_once` to install the root; (4) return the `Arc`. A helper hides the `Once` plumbing in one place and makes it structurally impossible for a future namespace constructor to forget the `call_once`:

```rust
fn new_with_root<F>(owner: Arc<UserNamespace>, build_root: F) -> Result<Arc<Self>>
where
    F: FnOnce(&Weak<Self>) -> Result<Arc<Mount>>,
{
    let new_ns = Arc::new(Self {
        root: Once::new(),
        owner,
        stashed_dentry: StashedDentry::new(),
    });
    let root = build_root(&Arc::downgrade(&new_ns))?;
    new_ns.root.call_once(|| root);
    Ok(new_ns)
}
```

With this in place, `Drop`'s `if self.root.get().is_none() { return; }` becomes load-bearing only for the helper-internal failure window — currently it exists *because* construction is open-coded twice.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -379,7 +395,7 @@ impl Mount {
                     continue;
                 }
                 let new_child_mount =
-                    old_child_mount.clone_mount(old_child_mount.root_dentry(), new_ns);
+                    old_child_mount.clone_mount(old_child_mount.root_dentry(), new_ns)?;
```
tatetian, line 398: **Partial new tree on `Err` leaks mountpoint counts on source dentries** (significant): When `old_child_mount.clone_mount(...)?` at line 398 returns `Err` mid-loop, prior successful iterations of the loop body have already called `new_child_mount.set_mountpoint(...)` (line 405), which invokes `inc_mount_count()` on the source dentries (the *old* tree's dentries, shared with the new mounts). On error, the partial new tree drops via local-variable scope, but `Mount::Drop` (line 632) only frees the mount ID; it does not call `clear_mountpoint()`, so the matching `dec_mount_count()` never runs. The result is permanently elevated `mount_count` on the source dentries, which keeps `DentryFlags::MOUNTED` set and corrupts subsequent `is_mountpoint()` checks. The original code couldn't trigger this because it `.unwrap()`ed and aborted; making `clone_mount_tree` fallible exposes the latent ordering bug.

Fixe: Make `Mount::Drop` idempotently clear its own mountpoint — `clear_mountpoint()` already no-ops if the mountpoint is `None`, and `MountNamespace::Drop` calls `clear_mountpoint()` before any `Mount` is dropped, so the new behavior is a no-op on the success path:

```rust
impl Drop for Mount {
    fn drop(&mut self) {
        self.clear_mountpoint();
        ID_ALLOCATOR.get().unwrap().lock().free(self.id);
    }
}
```

tatetian (commented):
Thanks for fixing the bug. This is a minimal fix to eliminate ID exhaustion.

While reviewing the code, I reached the conclusion that the entire mount ID allocation logic must be overhauled to (1) get rid of the hardcoded upper bound for mount IDs, (2) support the 64-bit unique mount IDs in addition to the legacy 32-bit ones, and (3) support looking up `Mount` with unique mount IDs. In Linux, the mount ID allocator is based on Xarray. And Linux's mount ID allocator may return `ENOSPC` (if the number of mount IDs reaches an upper bound specified by a user-specified policy) or `ENOMEM` (if Xarray fails to allocate new nodes).

lrh2000:
> To allow `MountNamespace::new_clone` to propagate the error with `?`, `MountNamespace::root` is changed from `Arc<Mount>` to `Once<Arc<Mount>>`, replacing `Arc::new_cyclic` with a two-phase init pattern (allocate namespace first, then clone mount tree).

I think this should be avoided. It does not make much sense to use `Once` to store a value that must be initialized during constructors. We can try to think about other approaches as well.

For example, we can add a type like:
https://github.com/asterinas/asterinas/blob/035226075a6219713c374b9af51f4fe0b4df1691/kernel/src/fs/fs_impls/pseudofs/allocator.rs#L8-L32

Then we can do something like:
```rust
        let mnt_id = MountId::acquire()?;
        let new_mnt_ns = Arc::new_cyclic(|weak_self| {
            let new_root = root_mount.clone_mount_tree(
                mnt_id, /* ... */
            );
        );
```

So we don't need to deal with something like https://github.com/asterinas/asterinas/pull/3219#discussion_r3231073246...

avasummer:
> I think this should be avoided. It does not make much sense to use `Once` to store a value that must be initialized during constructors. We can try to think about other approaches as well.
> 
> For example, we can add a type like:
> 
> https://github.com/asterinas/asterinas/blob/035226075a6219713c374b9af51f4fe0b4df1691/kernel/src/fs/fs_impls/pseudofs/allocator.rs#L8-L32
> 
> Then we can do something like:
> 
> ```rust
>         let mnt_id = MountId::acquire()?;
>         let new_mnt_ns = Arc::new_cyclic(|weak_self| {
>             let new_root = root_mount.clone_mount_tree(
>                 mnt_id, /* ... */
>             );
>         );
> ```
> 
> So we don't need to deal with something like [#3219 (comment)](https://github.com/asterinas/asterinas/pull/3219#discussion_r3231073246)...

The motivation for changing `root` to `Once<Arc<Mount>>` was to eliminate `Arc::new_cyclic` in `MountNamespace::new_clone`. The core issue is that `clone_mount_tree` (which calls `clone_mount` multiple times internally) is now fallible, but `Arc::new_cyclic`'s closure cannot return a `Result`. This removes `Arc::new_cyclic` and uses a two-phase init, where the namespace is allocated first with root unset, then the tree is cloned, and root is assigned on success.

Could you clarify the intended approach for `new_clone`? Specifically, with the `MountId` pattern, `Mount::new` and `clone_mount` can pre-acquire IDs before construction, but `clone_mount_tree` with `recursive: true` allocates IDs for each child mount internally. Thanks.

tatetian:
> It does not make much sense to use `Once` to store a value that must be initialized during constructors.

Yes, this pattern should be frown upon. I also dislike, although I can't think of a better alternative. 

```rust
        let mnt_id = MountId::acquire()?;
        let new_mnt_ns = Arc::new_cyclic(|weak_self| {
            let new_root = root_mount.clone_mount_tree(
                mnt_id, /* ... */
            );
        );
```

I thought about this pattern of passing a new ID explicitly. But I immediately dismissed it in my head because logically a new ID is part of the new object. So it should be the job of the object's constructor to allocate the ID, not the caller. So I feel this pattern is a little bit weird, but may be acceptable, if using `Once` as a field is a bigger concern.

tatetian:
> Could you clarify the intended approach for `new_clone`? Specifically, with the `MountId` pattern, `Mount::new` and `clone_mount` can pre-acquire IDs before construction, but `clone_mount_tree` with `recursive: true` allocates IDs for each child mount internally. Thanks.

Yeah, this one is hard...

lrh2000:
> but `clone_mount_tree` with `recursive: true` allocates IDs for each child mount internally.

Thank you for your explanation. I missed this case when I skimmed the code. Sorry for my carelessness.

We will need `ArcUninit` to resolve this problem. It has [an accepted ACP](https://github.com/rust-lang/libs-team/issues/660#issuecomment-3457751531) but no one has implemented it.

For now, having this `Once` seems very unfortunate, but I have no better solutions.

> But I immediately dismissed it in my head because logically a new ID is part of the new object. So it should be the job of the object's constructor to allocate the ID, not the caller. 

Personally, I'd prefer to avoid the `Once` if possible. However, it's subjective because it involves personal judgments between two imperfect solutions. I also agree with everything you said.

tatetian:
> We will need `ArcUninit` to resolve this problem. It has [an accepted ACP](https://github.com/rust-lang/libs-team/issues/660#issuecomment-3457751531) but no one has implemented it.

@lrh2000 This looks like a powerful primitive.

---

@avasummer If we have to use a `Once`-like wrapper, then I suppose that we should use `Option`. The `get` method of `Once` has to rely on atomic operation for thread safety, but `Option` is only modifiable when one gets a mutable reference to it. So `Option` is cheaper.

```rust
pub struct MountNamespace {
    /// The root mount.
    ///
    /// This field is wrapped within an `Option<_>`
    /// because the root mount is unknown 
    /// in the beginning of the constructor method (see `new_clone`).
    /// But if the constructor method completes,
    /// this field is guaranteed to be `Some(_)`.
    root: Option<Arc<Mount>>,
    // ...
}

impl MountNamespace {
    pub fn new_clone(
        &self,
        owner: Arc<UserNamespace>,
        posix_thread: &PosixThread,
    ) -> Result<Arc<MountNamespace>> {
        owner.check_cap(CapSet::SYS_ADMIN, posix_thread)?;

        let root_mount = self.root();
        let new_mnt_ns = Arc::new(MountNamespace {
            root: None,
            owner,
            stashed_dentry: StashedDentry::new(),
        });
        let new_root = root_mount.clone_mount_tree(
            root_mount.root_dentry(),
            &Arc::downgrade(&new_mnt_ns),
            true,
            MountNsFileCopying::Skip,
        )?;
        new_mnt_ns
            .as_mut()
            .unwrap()
            .root = Some(new_root);

        Ok(new_mnt_ns)            
    }
}
```

lrh2000:
> ```rust
> 
>         new_mnt_ns
>             .as_mut()
>             .unwrap()
> ```

It will just fail because [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) requires:
>> Returns a mutable reference into the given Arc, **if there are no other Arc or Weak pointers to the same allocation.**

What you want here is: [`UniqueArc`](https://doc.rust-lang.org/std/sync/struct.UniqueArc.html), where:
 1. [`UniqueArc::downgrade`](https://doc.rust-lang.org/std/sync/struct.UniqueArc.html#method.downgrade) allows to create `Weak`s that cannot be upgraded before [`UniqueArc::into_arc`](https://doc.rust-lang.org/std/sync/struct.UniqueArc.html#method.into_arc) is called, 
 2. but you can still [mutate the value via `DerefMut`](https://doc.rust-lang.org/std/sync/struct.UniqueArc.html#impl-DerefMut-for-UniqueArc%3CT,+A%3E) in the meantime.

avasummer:
Thanks for the thorough review. I've updated the PR with three commits addressing the feedback.

The kernel no longer panics:

```
created 9729 ramfs mounts under /tmp/aster-mount-id-poc-279
mount(/tmp/aster-mount-id-poc-279/m09979) failed after 9979 mounts: Cannot allocate memory
```

kernel/src/fs/fs_impls/pseudofs/anon_inodefs.rs
```diff
@@ -15,6 +15,10 @@ use crate::{
     process::{Gid, Uid},
 };
 
+pub(super) fn init() {
+    AnonInodeFs::mount_node();
```
tatetian: A comment should be inserted here. A reader should be surprised to see calling a getter method without using the returned object:

```rust
    // Touch the mount node to trigger the initialization of all singletons.
    let _ = AnonInodeFs::mount_node();
```

This argument applies to all the affected `init` functions.

tatetian (approved):

PR #3225, opened 2026-05-14
Implement ext4 read-only filesystem support foundation

1604141326:
Add ext4 filesystem read-only foundation

Implement the skeleton for ext4 read-only support:

- Superblock: full on-disk layout (1024 bytes), feature flag sets (compat/incompat/ro_compat), 64-bit block counts, backup group detection
- Block group: 64-byte group descriptor with 64-bit fields, inode bitmap loading, BlockGroup with lazy inode loading and caching
- Extent tree: ExtentHeader/ExtentIndex/ExtentLeaf parsing, recursive tree traversal for logical-to-physical block mapping
- Inode: 256-byte on-disk format, InodeDesc with extent-based block mapping via InodeBlockMapper, PageCacheBackend integration
- Directory: DirEntry/DirEntryHeader/DirEntryReader for entry iteration and name lookup, FILETYPE support
- VFS adapter: FileSystem trait (sb, sync, root_inode) and Inode trait (only read operations; writes return EROFS)
- Registry: ext4 registered as NEED_DISK filesystem type

Also move Dirty<T> and IsPowerOf trait from ext2 to shared fs/utils for cross-filesystem reuse.
@

### DIFF
--- a/kernel/src/fs/fs_impls/ext4/block_group.rs
+++ b/kernel/src/fs/fs_impls/ext4/block_group.rs
@@ -1,0 +1,228 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Block group and group descriptor for Ext4.
+
+use super::{
+    inode::Inode as Ext4Inode,
+    prelude::*,
+    super_block::SuperBlock,
+};
+
+/// The in-memory rust block group descriptor.
+#[derive(Clone, Copy, Debug)]
+pub(super) struct GroupDescriptor {
+    pub block_bitmap_bid: u64,
+    pub inode_bitmap_bid: u64,
+    pub inode_table_bid: u64,
+    pub free_blocks_count: u32,
+    pub free_inodes_count: u32,
+    pub dirs_count: u16,
+    pub flags: u16,
+}
+
+/// The raw ext4 group descriptor on disk (64 bytes with 64-bit support).
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Pod)]
+pub(super) struct RawGroupDescriptor {
+    pub block_bitmap_lo: u32,
+    pub inode_bitmap_lo: u32,
+    pub inode_table_lo: u32,
+    pub free_blocks_count_lo: u16,
+    pub free_inodes_count_lo: u16,
+    pub dirs_count_lo: u16,
+    pub flags: u16,
+    pub exclude_bitmap_lo: u32,
+    pub block_bitmap_csum_lo: u16,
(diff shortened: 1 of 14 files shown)


tatetian:
Welcome to the Asterinas community.

Since implementing Ext4 in its full capacity is a huge effort and involve quite some complexity, I would like see a plan to move it forward. For example, I don't know why the first PR is to implement a read-only Ext4. What will be the subsequent PRs? An alternative approach is to first implement Ext2 and then add more Ext3 and Ext4 features on top of it.

In addition, you will need to ensure the **quality** of the code and follow the **established convention** of the community. Contributors outnumber reviewers. So each PR should be polished as much as possible before requesting for reviews. Here are some specific points: 

* A PR needs to a thorough self-review, preferably with a help from the [`pr-review`](https://github.com/tatetian/asterinas/blob/pr-review-skill/.agents/skills/pr-review/SKILL.md) skill. 
* A PR must pass all CI tests and new tests should be added for new features. The proposed implementation of Ext4 in this PR is not tested at all. This is unacceptable. The xftests test suite has been merged into the mainline by #2945. It is highly recommended to test your Ext4 implementation against the test suite.
* A PR must follow the [Coding Guidelines](https://asterinas.github.io/book/to-contribute/coding-guidelines/index.html). For example, the [`imperative-subject`](https://asterinas.github.io/book/to-contribute/coding-guidelines/git-guidelines.html#imperative-subject) guideline requires each Git commit message start with a _verb_, but this PR starts with "@".

In addition, there is an on-going PR, #3171, that rewrites Ext2, which vastly improves the quality of the code over the current implementation. Ext4 is backward-compatible to Ext2. So if this PR was accepted, we would have two competing Ext2 implementations, which does not make much sense from the perspective of maintenance. Could you take a look at the new Ext2 PR? Is it possible to consolidate your effort with the new Ext2 implementation? Adding Ext4 features on top of the new Ext2 is a viable option. You probably need to make a case as to why Ext4 should be built from scratch, as done in this PR. I understand that your development is done in parallel. But it looks that the new Ext2 will be merged earlier than yours. So you need to take it into consideration.

We welcome all contributors, especially newcomers. But I want to be clear upfront: Asterinas puts a high quality bar for code contributions. It would be a great testimony and commend for the professionalism and ingenuity shown in your code if it gets merged. But the road ahead would be long and bumpy. Be prepared 😄 

I will mark this PR as "Draft". Feel free to mark this "Ready for Review" when you got all the answers for the above questions.

tatetian:
As I explained above, this PR has no way to get merged as-is. So I am closing this PR.

PR #3226, opened 2026-05-14
Fix some sendfile problems

StevenJiang1110:
This PR addresses the issues raised in https://github.com/asterinas/asterinas/pull/3191#issuecomment-4448076510. As suggested, we now reject all non-seekable `in_file`s, which ensures that the file position of `in_file` can be correctly updated based on the number of bytes actually written.

We also fix all other failing sendfile test cases in the gVisor and LTP test suites, primarily by adding missing boundary checks. Two gVisor test cases still fail, but they are not directly related to the sendfile implementation itself — comments have been added to document the reasons.

Additionally, we remove the comment `FIXME: the in_file must support mmap-like operations (i.e., it cannot be a socket)`., as it was based on a misreading of the [Linux manual](https://man7.org/linux/man-pages/man2/sendfile.2.html):

```plain
       The in_fd argument must correspond to a file which supports
       mmap(2)-like operations (i.e., it cannot be a socket).  Except
       since Linux 5.12 and if out_fd is a pipe, in which case sendfile()
       desugars to a splice(2) and its restrictions apply.
```

As the manual clarifies, when out_fd is a pipe, in_fd is allowed to be a socket.

### DIFF
--- a/kernel/src/syscall/sendfile.rs
+++ b/kernel/src/syscall/sendfile.rs
@@ -7,7 +7,7 @@ use crate::{
     fs::{
         self,
         file::{
-            InodeHandle, InodeType, SeekFrom,
+            InodeHandle, InodeType, SeekFrom, StatusFlags,
             file_table::{RawFileDesc, WithFileTable},
         },
     },
@@ -38,10 +38,14 @@ pub fn sys_sendfile(
         out_fd, in_fd, offset, count
     );
 
+    if offset.is_some_and(|off| off.checked_add(count).is_none()) {
+        return_errno_with_message!(Errno::EINVAL, "offset + count overflows");
+    }
+
     let mut count = if count < 0 {
         return_errno_with_message!(Errno::EINVAL, "count cannot be negative");
     } else {
-        count as usize
+        count.cast_unsigned()
     };
 
     let (out_file, in_file) = ctx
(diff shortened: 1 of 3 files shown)


test/initramfs/src/conformance/gvisor/blocklists/sendfile_test
```diff
@@ -1,7 +1,5 @@
...
-SendFileTest.SendFileToSelf
+# Requires fcntl(pipe_fd, F_GETPIPE_SZ) support
 SendFileTest.SendPipeBlocks
-SendFileTest.SendPipeWouldBlock
-SendFileTest.SendToNotARegularFile
-SendFileTest.SendToSpecialFile
\ No newline at end of file
+# Linux returns EISDIR when the output file is a directory;
+# Asterinas currently returns EBADF instead.
+SendFileTest.SendToNotARegularFile
```
lrh2000: This isn't very correct? https://github.com/google/gvisor/blob/7e7d35b8a0aefee15c719ae176f34ffec07a21fc/test/syscalls/linux/sendfile.cc#L462-L476

The input file is a directory. What said here may be correct, but it isn't related to the test because `sendfile` should return `EINVAL` instead of `EISDIR` or `EBADF`.

test/initramfs/src/conformance/gvisor/blocklists/sendfile_test
```diff
@@ -1,7 +1,5 @@
-SendFileTest.DoNotSendfileIfOutfileIsAppendOnly
-SendFileTest.Overflow
-SendFileTest.SendFileToSelf
+# Requires fcntl(pipe_fd, F_GETPIPE_SZ) support
```
lrh2000: Maybe follow:
```
test/initramfs/src/conformance/gvisor/blocklists/pipe_test:# TODO: Support `F_SETPIPE_SZ`/`F_GETPIPE_SZ` for `fcntl` syscall.
```

Also some other minor adjustments if they make sense...
```diff
diff --git a/kernel/src/syscall/sendfile.rs b/kernel/src/syscall/sendfile.rs
index c7d37bf09..bd84c8cc6 100644
--- a/kernel/src/syscall/sendfile.rs
+++ b/kernel/src/syscall/sendfile.rs
@@ -45,7 +45,7 @@ pub fn sys_sendfile(
     let mut count = if count < 0 {
         return_errno_with_message!(Errno::EINVAL, "count cannot be negative");
     } else {
-        count as usize
+        count.cast_unsigned()
     };
 
     let (out_file, in_file) = ctx
@@ -75,7 +75,7 @@ pub fn sys_sendfile(
         // Note: Linux allows a non-seekable `in_file` when the output file is a pipe,
         // relying on specialized pipe handling to avoid data loss.
         // We do not currently support this case and reject it for simplicity.
-        // TODO: Support sendfile from a non-seekable in_file to a pipe.
+        // TODO: Support `sendfile` from a non-seekable `in_file` to a pipe.
         Some(in_file.seek(SeekFrom::Current(0)).map_err(|_| {
             if outfile_is_pipe {
                 warn!("sendfile from a non-seekable in_file to a pipe is not yet supported");
@@ -99,19 +99,19 @@ pub fn sys_sendfile(
     const BUFFER_SIZE: usize = PAGE_SIZE;
     let mut buffer = vec![0u8; BUFFER_SIZE].into_boxed_slice();
     let mut total_len = 0;
-    let mut offset = offset.map(|offset| offset as usize);
+    let mut offset = offset.map(|offset| offset.cast_unsigned());
     let mut short_write_occurs = false;
 
     while total_len < count {
         // The offset decides how to read from `in_file`.
-        // If offset is `Some(_)`, the data will be read from the given offset,
+        // If the offset is `Some(_)`, the data will be read from the given offset,
         // and after reading, the file offset of `in_file` will remain unchanged.
-        // If offset is `None`, the data will be read from the file offset,
+        // If the offset is `None`, the data will be read from the file offset,
         // and the file offset of `in_file` is adjusted
         // to reflect the number of bytes read from `in_file`.
         let max_readlen = buffer.len().min(count - total_len);
 
-        // Read from `in_file`
+        // Read from `in_file`.
         let read_res = if let Some(offset) = offset {
             in_file.read_bytes_at(offset, &mut buffer[..max_readlen])
         } else {
@@ -134,7 +134,7 @@ pub fn sys_sendfile(
         }
 
         // Note: `sendfile` allows sending partial data,
-        // so short reads and short writes are all acceptable
+        // so short reads and short writes are all acceptable.
         let write_res = out_file.write_bytes(&buffer[..read_len]);
 
         match write_res {
@@ -154,11 +154,9 @@ pub fn sys_sendfile(
                     short_write_occurs = true;
                     break;
                 }
-
                 if offset.is_none() {
                     in_file.seek(SeekFrom::Start(origin_f_pos.unwrap()))?;
                 }
-
                 return Err(e);
             }
         }
@@ -168,7 +166,7 @@ pub fn sys_sendfile(
         ctx.user_space().write_val(offset_ptr, &(offset as isize))?;
     } else if short_write_occurs {
         // Seek `in_file` to the position corresponding to the last byte actually transferred.
-        // Note: since the file offset lock is not held
+        // Note: Since the file offset lock is not held
         // between saving `origin_f_pos` and this point,
         // a race condition may occur if another thread reads from the same file concurrently.
         // Linux permits this behavior and leaves it to userspace to avoid such races.
diff --git a/test/initramfs/src/conformance/gvisor/blocklists/sendfile_test b/test/initramfs/src/conformance/gvisor/blocklists/sendfile_test
index 73680890f..3c67db850 100644
--- a/test/initramfs/src/conformance/gvisor/blocklists/sendfile_test
+++ b/test/initramfs/src/conformance/gvisor/blocklists/sendfile_test
@@ -1,5 +1,5 @@
-# Requires fcntl(pipe_fd, F_GETPIPE_SZ) support
+# TODO: Support `F_GETPIPE_SZ` for `fcntl` syscall.
 SendFileTest.SendPipeBlocks
 # Linux returns EISDIR when the output file is a directory;
 # Asterinas currently returns EBADF instead.
-SendFileTest.SendToNotARegularFile
\ No newline at end of file
+SendFileTest.SendToNotARegularFile
```

lrh2000 (commented):
Thanks for your efforts to fix `sendfile` behavior!

test/initramfs/src/conformance/gvisor/blocklists/sendfile_test
StevenJiang1110: I do some changes to make the test case pass in the latest update.

kernel/src/syscall/sendfile.rs
```diff
@@ -38,69 +38,97 @@ pub fn sys_sendfile(
...
+            }
+            Error::with_message(Errno::EINVAL, "in_file is not seekable")
+        })?)
     } else {
         None
     };
 
+    // Verify that `in_file` is readable and `out_file` is writable upfront,
+    // even if the total transfer count is zero.
+    in_file.read_bytes(&mut []).map_err(|e| {
```
tatetian: In Linux, the `sendfile` system call performs the check by probing the mode bits.

```c
/*
	 * Get input file, and verify that it is ok..
	 */
    ...
	if (!(fd_file(in)->f_mode & FMODE_READ))
		return -EBADF;

	/*
	 * Get output file, and verify that it is ok..
	 */
	...
	if (!(fd_file(out)->f_mode & FMODE_WRITE))
		return -EBADF;
```

In Asterinas, our counterpart of `f_mode` is `FileLike::access_mode`. The access mode should be the authoritative info to answer the question if a file allows reads or writes——and it is more efficient than doing a zero-length read/write.

The problem is that I don't think all file types have faithfully implemented the `FileLike::access_mode` method. Currently, the trait provides a default implementation that returns `AccessMode::O_RDWR`. I believe this is a mistake. So I suggest removing the default implementation, adding a proper implementation or examine the existing one for each and every file type.

You can implement it in a prerequisite PR.

lrh2000 (approved):
LGTM! Thanks for your contribution.

kernel/src/syscall/sendfile.rs
StevenJiang1110: I reimplement the `access_mode` method in #3231.

PR #3227, opened 2026-05-15
Rename `io_util` to `io-util` and fix some other minor problems

lrh2000:
```
> ls kernel/libs
aster-bigtcp  aster-rights-proc  atomic-integer-wrapper  cpio-decoder  io-util  keyable-arc     ring-buffer  typeflags-util
aster-rights  aster-util         comp-sys                device-id     jhash    logo-ascii-art  typeflags    xarray
```

All other crate names use `-` instead of `_`.

Fix some minor problems:
 - `kernel/src/fs/vfs/path/dentry.rs` has warnings when `RELEASE=1` due to unused imports and unused variables.
 - `VmoCommitError::pending_index` can return `Result<>` instead of `Option<>` so we don't need to write `unreachable!()`.
 - ...


cchanging (approved):

PR #3229, opened 2026-05-15
Remove `PageCacheBackend::npages`

lrh2000:
When reading the VMO code, it is difficult to determine whether the `npages()` method is used correctly, i.e., whether it will cause races.

##### Problem 1

https://github.com/asterinas/asterinas/blob/818cfd9da9748d587b8346918b2261bed4e4bda2/kernel/src/vm/page_cache/vmo/mod.rs#L792-L797

It isn't immediately clear to me whether the backend `npages()` may change between the query and the operation (i.e., `read_pages`). One possibility is that the change to the `npages()` function can be synchronized via the reverse mapping/invalidation lock, which will be added in the future. However, the code would still be difficult to analyze.

##### Problem 2

https://github.com/asterinas/asterinas/blob/818cfd9da9748d587b8346918b2261bed4e4bda2/kernel/src/vm/page_cache/vmo/mod.rs#L841-L848

Also, I don't understand why we simply return the page when the offset is outside of `npages()`. It appears to leak uninitialized page content to the user space, which is terrible.

##### Problem 3

https://github.com/asterinas/asterinas/blob/818cfd9da9748d587b8346918b2261bed4e4bda2/kernel/src/vm/page_cache/vmo/mod.rs#L832-L842

Another problem is that `npages()` willl break the atomic mode. XArray `Cursor` can only be obtained in the atomic mode, but the code calls `npages()`, which in turns will lock a mutex for exfat:

https://github.com/asterinas/asterinas/blob/818cfd9da9748d587b8346918b2261bed4e4bda2/kernel/src/fs/fs_impls/exfat/inode.rs#L181-L182

---

As a result, I think we should remove `PageCacheBackend::npages()` to solve all three problems. `submit_read_bio` and `submit_write_bio` should validate the `idx` before reading or writing to the disk.

Additionally, submit_read_bio should fill the page with zeros if a file range (in the middle or at the end) isn't allocated on the disk. I found that exfat already takes advantage of this, so I added a `BioStatus::Zeros` to achieve this goal.

### DIFF
--- a/kernel/src/vm/page_cache/mod.rs
+++ b/kernel/src/vm/page_cache/mod.rs
@@ -86,7 +86,7 @@ use core::{
 use align_ext::AlignExt;
 use aster_block::bio::{BioCompleteFn, BioDirection, BioSegment, BioStatus};
 use io_util::batch::IoBatch;
-use ostd::mm::{Segment, VmIo, io::util::HasVmReaderWriter};
+use ostd::mm::{Segment, VmIo, VmIoFill, io::util::HasVmReaderWriter};
 
 use crate::prelude::*;
 
@@ -345,6 +345,11 @@ impl VmIo for PageCache {
 /// devices usually implement [`BlockAsPageCacheBackend`] instead.
 pub trait PageCacheBackend: Sync + Send {
     /// Reads a page from the backend asynchronously.
+    ///
+    /// The caller may try to pass an index that exceeds the size of the
+    /// underlying backend (e.g., the file size). This can occur if file
+    /// truncation and a page fault occur at the same time. If this happens,
+    /// this method should fail with `EINVAL`.
     fn read_page_async(
         &self,
         idx: usize,
@@ -353,18 +358,20 @@ pub trait PageCacheBackend: Sync + Send {
     ) -> Result<()>;
 
     /// Writes a page to the backend asynchronously.
+    ///
+    /// If the caller tries to pass an index that exceeds the size of the
+    /// underlying backend (e.g. the file size), this method should fail with
+    /// `EINVAL`. Note that this cannot happen until we support concurrent file
+    /// truncation and page writeback.
+    //
+    // TODO: Revise this behavior once file truncation and page writeback can
+    // happen concurrently.
     fn write_page_async(
         &self,
         idx: usize,
         locked_page: cache_page::LockedCachePage,
         io_batch: &mut IoBatch,
     ) -> Result<()>;
-
-    /// Returns the number of pages addressable in the backend storage.
-    ///
-    /// This defines the upper bound of valid page indices for I/O operations.
-    /// Requests beyond this limit are treated as holes (zero-filled).
-    fn npages(&self) -> usize;
 }
 
 impl dyn PageCacheBackend {
@@ -399,6 +406,9 @@ pub trait BlockAsPageCacheBackend: Sync + Send {
     /// Implementations must attach `complete_fn`, when present, to the
     /// underlying asynchronous I/O so the page-cache layer can mark successful
     /// reads up to date and release the page lock.
+    ///
+    /// Implementations should fail with `EINVAL` for an out-of-bounds `idx`.
+    /// See also [`PageCacheBackend::read_page_async`].
     fn submit_read_bio(
         &self,
         idx: usize,
@@ -413,6 +423,9 @@ pub trait BlockAsPageCacheBackend: Sync + Send {
     /// Implementations must attach `complete_fn`, when present, to the
     /// underlying asynchronous I/O so the page-cache layer can finish writeback
     /// bookkeeping and report failures.
+    ///
+    /// Implementations should fail with `EINVAL` for an out-of-bounds `idx`.
+    /// See also [`PageCacheBackend::write_page_async`].
     fn submit_write_bio(
         &self,
         idx: usize,
@@ -420,12 +433,6 @@ pub trait BlockAsPageCacheBackend: Sync + Send {
         complete_fn: Option<BioCompleteFn>,
         io_batch: &mut IoBatch,
     ) -> Result<()>;
-
-    /// Returns the number of pages addressable in the backend storage.
-    ///
-    /// This defines the upper bound of valid page indices for I/O operations.
-    /// Requests beyond this limit are treated as holes (zero-filled).
-    fn npages(&self) -> usize;
 }
 
 impl<T: BlockAsPageCacheBackend> PageCacheBackend for T {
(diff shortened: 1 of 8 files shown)


tatetian:
I agree with the motivation.

@cchanging Could you review the changes carefully?

kernel/src/vm/page_cache/mod.rs
```diff
@@ -420,12 +415,6 @@ pub trait BlockAsPageCacheBackend: Sync + Send {
         complete_fn: Option<BioCompleteFn>,
         io_batch: &mut IoBatch,
     ) -> Result<()>;
-
-    /// Returns the number of pages addressable in the backend storage.
-    ///
-    /// This defines the upper bound of valid page indices for I/O operations.
-    /// Requests beyond this limit are treated as holes (zero-filled).
-    fn npages(&self) -> usize;
 }
```
cchanging, line 429: After removing `npages`, I think we could mention in the documentation for `PageCacheBackend` that implementers are responsible for handling valid-page checks themselves, including zero-filling.

kernel/src/vm/page_cache/vmo/mod.rs
cchanging, line 720: After removing `npages`, it seems that `flush` could end up writing back a dirty page beyond EOF. In `Ext2`, it looks like this case is currently handled by directly returning an error, but I wonder whether treating it as a no-op would be more reasonable. I did a quick check on how Linux handles similar cases:

https://github.com/torvalds/linux/blob/5200f5f493f79f14bbdc349e402a40dfb32f23c8/fs/buffer.c#L2635-L2639

cchanging (commented):
Thanks for further improving the `PageCache` API! Removing the `npages` API makes sense to me. It makes the overall implementation clearer. I only have a few minor questions.

kernel/src/vm/page_cache/vmo/mod.rs
lrh2000: It appears that truncation can only occur after the inode write lock is held. Currently, we cannot trigger writeback without holding the inode lock. Therefore, they cannot happen concurrently.

I do wonder if we should relax the lock usage for writeback. What rules should we use? Do you have any insights?

Perhaps we only need to require the reverse mapping lock for writeback? But truncation also needs to hold the reverse mapping lock....

cchanging: > It appears that truncation can only occur after the inode write lock is held. Currently, we cannot trigger writeback without holding the inode lock. Therefore, they cannot happen concurrently.

Yes, currently writeback is only synchronous, so there is no race.

> I do wonder if we should relax the lock usage for writeback. What rules should we use? Do you have any insights?

The locking rules for writeback are indeed looser; they are different from those for the direct interface to the page cache. In Linux, writeback can be performed by dedicated background threads that can flush pages. Such background writeback operations take the page lock, but not the inode mutex, and when the page is not mapped, they also do not take the reverse-mapping lock. It is probably only after this full mechanism is implemented that concurrent writeback and truncate could actually occur...

lrh2000: > In Linux, writeback can be performed by dedicated background threads that can flush pages. Such background writeback operations take the page lock, but not the inode mutex, and when the page is not mapped, they also do not take the reverse-mapping lock.

It doesn't sound like something we'll have in the near future.

So I am a bit unsure what I should do here...

---

I tried to follow the suggestion. I think it means that we should return `Ok(())` for an out-of-bounds `idx` in `submit_write_bio`. However, when I wrote the doc comments to explain why `Ok(())` is correct, I found that the case does not exist in our current codebase.

The current version still returns `EINVAL` with the following doc comments.

```patch
diff --git a/kernel/src/vm/page_cache/mod.rs b/kernel/src/vm/page_cache/mod.rs
index 9e31ace37..6739a5782 100644
--- a/kernel/src/vm/page_cache/mod.rs
+++ b/kernel/src/vm/page_cache/mod.rs
@@ -345,6 +345,11 @@ impl VmIo for PageCache {
 /// devices usually implement [`BlockAsPageCacheBackend`] instead.
 pub trait PageCacheBackend: Sync + Send {
     /// Reads a page from the backend asynchronously.
+    ///
+    /// The caller may try to pass an index that exceeds the size of the
+    /// underlying backend (e.g., the file size). This can occur if file
+    /// truncation and a page fault occur at the same time. If this happens,
+    /// this method should fail with `EINVAL`.
     fn read_page_async(
         &self,
         idx: usize,
@@ -353,18 +358,17 @@ pub trait PageCacheBackend: Sync + Send {
     ) -> Result<()>;
 
     /// Writes a page to the backend asynchronously.
+    ///
+    /// If the caller tries to pass an index that exceeds the size of the
+    /// underlying backend (e.g. the file size), this method should fail with
+    /// `EINVAL`. Note that this cannot happen until we support concurrent file
+    /// truncation and page writeback.
     fn write_page_async(
         &self,
         idx: usize,
         locked_page: cache_page::LockedCachePage,
         io_batch: &mut IoBatch,
     ) -> Result<()>;
-
-    /// Returns the number of pages addressable in the backend storage.
-    ///
-    /// This defines the upper bound of valid page indices for I/O operations.
-    /// Requests beyond this limit are treated as holes (zero-filled).
-    fn npages(&self) -> usize;
 }
 
 impl dyn PageCacheBackend {
```

But I don't really have a preference either way...

lrh2000:
Due to the newly added `BioStatus::Zeros` logic, I think `submit_read_bio` should accept `BioCompleteFn` instead of `Option<BioCompleteFn>`. Therefore, I added the second commit.

kernel/src/vm/page_cache/vmo/mod.rs
cchanging: > It doesn't sound like something we'll have in the near future.

Yes there’s still quite a long way to go before this is actually implemented. 

> The current version still returns EINVAL with the following doc comments.

The current version is also OK for me, since a race condition won’t actually happen in practice. Perhaps we could add an additional `TODO` to remind ourselves to handle the behavior here if/when asynchronous flush is implemented in the future?

lrh2000: I think the fact has already been noted, though it is not an explicit TODO.

Do you mean this?
```rust
    /// Writes a page to the backend asynchronously.
    ///
    /// If the caller tries to pass an index that exceeds the size of the
    /// underlying backend (e.g. the file size), this method should fail with
    /// `EINVAL`. Note that this cannot happen until we support concurrent file
    /// truncation and page writeback.
    //
    // TODO: Revise this behavior once file truncation and page writeback can
    // happen concurrently.
    fn write_page_async(
        &self,
        idx: usize,
        locked_page: cache_page::LockedCachePage,
        io_batch: &mut IoBatch,
    ) -> Result<()>;
```

cchanging (approved):

PR #3230, opened 2026-05-15
Support unaligned block device I/O

### LINKED ISSUE #3214: Support short or non-sector-sized `pread` on block device
### Describe the bug

Asterinas guest block devices do not yet provide Linux-compatible `pread` semantics for short and non-sector-sized metadata reads.

### To Reproduce

1. Boot an Asterinas guest with a virtio-blk test disk exposed as `/dev/vdb`.
2. Run direct `pread` probes against that device, for example:

```c
#include <errno.h>
#include <fcntl.h>
#include <stdint.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

int main(void) {
    int fd = open("/dev/vdb", O_RDONLY);
    unsigned char boot[64];
    uint32_t entry = 0;

    if (fd < 0) {
        perror("open");
        return 1;
    }

    errno = 0;
    ssize_t n1 = pread(fd, boot, sizeof(boot), 0);
    printf("pread boot: n=%zd errno=%d (%s)\n", n1, errno, strerror(errno));

    errno = 0;
    ssize_t n2 = pread(fd, &entry, sizeof(fat_entry), 1048592);
    printf("pread entry: n=%zd errno=%d (%s)\n", n2, errno, strerror(errno));
    return 0;
}
```

3. Both `pread`s will return with `errno=22` (`EINVAL`, "Invalid argument").

### Expected behavior

These reads should succeed on a readable block device.

### Screenshots

Image

Image

### Environment

- Hardware platform: x86_64 QEMU/KVM guest with a virtio-blk data disk

### Additional context

Related issues:

- Tracker: [`#3211`](https://github.com/asterinas/asterinas/issues/3211)

cchanging:
Fixes #3214

### DIFF
--- a/kernel/comps/block/src/impl_block_device.rs
+++ b/kernel/comps/block/src/impl_block_device.rs
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use align_ext::AlignExt;
 use io_util::batch::IoBatch;
 use ostd::mm::{VmIo, VmReader, VmWriter};
 
 use super::{
-    BLOCK_SIZE, BlockDevice,
+    BLOCK_SIZE, BlockDevice, SECTOR_SIZE,
     bio::{Bio, BioCompleteFn, BioEnqueueError, BioSegment, BioStatus, BioType},
     id::{Bid, Sid},
 };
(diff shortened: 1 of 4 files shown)


kernel/comps/block/src/impl_block_device.rs
```diff
@@ -118,43 +126,70 @@ impl VmIo for dyn BlockDevice {
...
+        if !is_sector_aligned(offset) || !is_sector_aligned(write_len) {
+            let read_bio = Bio::new(
+                BioType::Read,
+                Sid::from_offset(aligned_offset),
+                vec![bio_segment.clone()],
+                None,
+            );
+            if read_bio.submit_and_wait(self)? != BioStatus::Complete {
+                return Err(ostd::Error::IoError);
+            }
```
StevenJiang1110: I don't think you need to read the entire bio segment, as the write range could be quite large. Instead, reading only the first and/or last sectors should be sufficient.

StevenJiang1110 (approved):
LGTM! Only a small nit.

kernel/comps/block/src/impl_block_device.rs
```diff
@@ -118,43 +127,89 @@ impl VmIo for dyn BlockDevice {
...
+            let is_the_same_sector = last_sector_offset == aligned_offset;
+            !(is_last_sector_aligned || need_read_first_sector && is_the_same_sector)
+        };
+        let middle_end = if need_read_last_sector {
+            last_sector_offset
+        } else {
+            aligned_end
         };
 
+        let mut bio_segments = SmallVec::<[BioSegment; 3]>::new();
```
StevenJiang1110: There is no need for `SmallVec` here — you convert it to a `Vec` at line 210 anyway. A `Vec::with_capacity(3)` should be sufficient.

cqs21 (approved):

PR #3231, opened 2026-05-15
Fix `access_mode` and `FileLike::read/write` default values

StevenJiang1110:
The first commit addresses https://github.com/asterinas/asterinas/pull/3226#discussion_r3245707097. Each file type now implements `access_mode` individually. A reference to the corresponding Linux behavior is also added.

The second commit fixes another issue discovered during the refactor. In Linux, whether a file is readable or writable is not determined solely by `FMODE_READ`/`FMODE_WRITE`. Linux also checks whether the file has read/write methods implemented; if so, it sets `FMODE_CAN_READ` and `FMODE_CAN_WRITE` accordingly when initializing a `struct file`. If a file has `FMODE_READ`/`FMODE_WRITE` set but lacks the corresponding read/write method, Linux returns `EINVAL`. This behavior can serve as a default implementation shared across all file types, allowing us to remove several redundant per-type implementations.

https://elixir.bootlin.com/linux/v7.0/source/fs/file_table.c#L330
```C
if ((file->f_mode & FMODE_WRITE) &&
       likely(fop->write || fop->write_iter))
      file->f_mode |= FMODE_CAN_WRITE;
```

https://elixir.bootlin.com/linux/v7.0/source/fs/read_write.c#L672
```C
if (!(file->f_mode & FMODE_WRITE))
      return -EBADF;
if (!(file->f_mode & FMODE_CAN_WRITE))
      return -EINVAL;
```

### DIFF
--- a/kernel/src/fs/file/file_handle.rs
+++ b/kernel/src/fs/file/file_handle.rs
@@ -79,9 +79,27 @@ pub trait FileLike: Pollable + Send + Sync + Any {
         return_errno_with_message!(Errno::EINVAL, "set_status_flags is not supported");
     }
 
-    fn access_mode(&self) -> AccessMode {
-        AccessMode::O_RDWR
-    }
+    /// Returns the access mode of this file.
+    ///
+    /// The access mode indicates whether the file descriptor is readable or writable.
+    /// Readability is required for operations
+    /// such as [`read`], [`read_at`], and read-only memory mappings.
+    /// Writability is required for operations
+    /// such as [`write`], [`write_at`], [`resize`], [`fallocate`],
+    /// and writable memory mappings.
+    ///
+    /// For inode-backed files,
+    /// their access modes are determined dynamically for each opened file.
+    /// For special files such as epoll files, eventfd files, and pid files,
+    /// they are determined statically by the file types.
+    ///
+    /// [`read`]: FileLike::read
+    /// [`read_at`]: FileLike::read_at
+    /// [`write`]: FileLike::write
+    /// [`write_at`]: FileLike::write_at
+    /// [`resize`]: FileLike::resize
+    /// [`fallocate`]: FileLike::fallocate
+    fn access_mode(&self) -> AccessMode;
 
     fn seek(&self, seek_from: SeekFrom) -> Result<usize> {
         return_errno_with_message!(Errno::ESPIPE, "seek is not supported");
--- a/kernel/src/fs/file/file_handle.rs
+++ b/kernel/src/fs/file/file_handle.rs
@@ -20,12 +20,48 @@ use crate::{
 
 /// The basic operations defined on a file
 pub trait FileLike: Pollable + Send + Sync + Any {
+    /// Reads data from this file.
+    ///
+    /// By default, this method returns `EBADF`
+    /// if the file is not readable,
+    /// or `EINVAL` if the file type does not support `read`.
+    /// These are distinct conditions:
+    /// [`access_mode`] may say that a file is readable,
+    /// while the file type still does not support `read`,
+    /// as with epoll files and pid files.
+    ///
+    /// Implementors should override this method if the file type supports reads.
+    /// An overriding implementation must check whether [`access_mode`] is readable
+    /// before performing the operation.
+    ///
+    /// [`access_mode`]: FileLike::access_mode
     fn read(&self, writer: &mut VmWriter) -> Result<usize> {
-        return_errno_with_message!(Errno::EBADF, "the file is not valid for reading");
+        if !self.access_mode().is_readable() {
+            return_errno_with_message!(Errno::EBADF, "the file is not opened for reading");
+        }
+        return_errno_with_message!(Errno::EINVAL, "read is not supported for this file type");
     }
 
+    /// Writes data to this file.
+    ///
+    /// By default, this method returns `EBADF`
+    /// if the file is not writable,
+    /// or `EINVAL` if the file type does not support `write`.
+    /// These are distinct conditions:
+    /// [`access_mode`] may say that a file is writable,
+    /// while the file type still does not support `write`,
+    /// as with epoll files and pid files.
+    ///
+    /// Implementors should override this method if the file type supports writes.
+    /// An overriding implementation must check whether [`access_mode`] is writable
+    /// before performing the operation.
+    ///
+    /// [`access_mode`]: FileLike::access_mode
     fn write(&self, reader: &mut VmReader) -> Result<usize> {
-        return_errno_with_message!(Errno::EBADF, "the file is not valid for writing");
+        if !self.access_mode().is_writable() {
+            return_errno_with_message!(Errno::EBADF, "the file is not opened for writing");
+        }
+        return_errno_with_message!(Errno::EINVAL, "write is not supported for this file type");
     }
 
     /// Read at the given file offset.
--- a/kernel/src/events/epoll/file.rs
+++ b/kernel/src/events/epoll/file.rs
@@ -261,14 +261,6 @@ impl Pollable for EpollFile {
 
 // Implement the common methods required by FileHandle
 impl FileLike for EpollFile {
-    fn read(&self, _writer: &mut VmWriter) -> Result<usize> {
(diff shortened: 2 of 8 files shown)


kernel/src/fs/file/file_handle.rs
```diff
@@ -79,9 +85,7 @@ pub trait FileLike: Pollable + Send + Sync + Any {
         return_errno_with_message!(Errno::EINVAL, "set_status_flags is not supported");
     }
 
-    fn access_mode(&self) -> AccessMode {
-        AccessMode::O_RDWR
-    }
+    fn access_mode(&self) -> AccessMode;
```
tatetian, line 138:
```rust
   /// Returns the access mode of this file.
   ///
   /// This access mode indicates
   /// whether it is possible to read or write this file.
   /// For inode-backed files,
   /// their access modes are determined _dynamically_
   /// on a per-file basis when they are opened.
   /// For special files (such as epoll files, eventfd, and pid files),
   /// their access modes are determined _statically_,
   /// by the implementation of each file type.
   ///
   /// # Implementators
   ///
   /// The implementator of `FileLike` should override the `read` method
   /// if the access mode returned by this method may suggest readable.
   /// Similarly,
   /// the implementator should override the `write` method
   /// if the access mode returned by this method may suggest writable.
   fn access_mode(&self) -> AccessMode;
```

kernel/src/fs/file/file_handle.rs
tatetian, line 74: If the `read` method needs to add the `!self.access_mode().is_readable()` check, then I suppose that `read_at` is the same?

kernel/src/fs/file/file_handle.rs
```diff
@@ -21,11 +21,17 @@ use crate::{
 /// The basic operations defined on a file
 pub trait FileLike: Pollable + Send + Sync + Any {
     fn read(&self, writer: &mut VmWriter) -> Result<usize> {
-        return_errno_with_message!(Errno::EBADF, "the file is not valid for reading");
+        if !self.access_mode().is_readable() {
+            return_errno_with_message!(Errno::EBADF, "the file is not opened readable");
+        }
+        return_errno_with_message!(Errno::EINVAL, "read is not supported");
```
tatetian: The change does not make sense for Asterinas. The change was made to mimic Linux's logic:

```c
	if ((file->f_mode & FMODE_READ) &&
	     likely(fop->read || fop->read_iter))
		file->f_mode |= FMODE_CAN_READ;
```

But I can conclude the refactored logic is flawed by a proof of contradiction. Assuming new logic is correct and has been put into effect. If a file type implements `access_mode` in a way that it is possible to return a value indicating file being readable (hitting the returning-`EINVAL` line), then the file type must have _overridden_ the default `read` method. This means that we will never reach the newly-added, returning-`EINVAL` line! So this change will never take effect.

I suppose that the correct fix is to keep this default implementation, but remove the overridden implementation that returns the incorrect error code.

And you might consider adding some doc comments to provide guidance to future file type implementators.

kernel/src/fs/file/file_handle.rs
StevenJiang1110: I think `read_at` is slightly different from `read` here.

In Linux, `ksys_pread64()` first checks whether the file supports positional reads through `FMODE_PREAD`; if that bit is not set, it returns `-ESPIPE` directly. Only after that check passes does it call `vfs_read()`, where Linux then checks `FMODE_READ` and `FMODE_CAN_READ` and returns `-EBADF` or `-EINVAL` as appropriate.

So I think the current split is closer to Linux's structure:

- the default `read_at` means that this file type does not support positional I/O at all, so it returns `ESPIPE`;
- for file types that do support positional I/O, their `read_at` implementation should perform the usual readability check before doing the actual read.

kernel/src/fs/file/file_handle.rs
StevenJiang1110: I think the new default implementation is still useful here, because Linux treats a file descriptor's open mode and the availability of file operations as two separate concepts.

Even if a file's access_mode is `O_RDWR`, it may still _not_ implement read or write — for example, pid files and epoll files.

The new default `read` and `write` implementations are intended to model this same distinction in Asterinas: `access_mode()` describes whether the descriptor was opened for reading or writing. Overriding `read` or `write` indicates whether the file type actually supports the corresponding operation.

Under this interpretation, the default `EINVAL` path is reached for file typeswhose `access_mode` reports readable or writable, but that do not override the corresponding method. For example, an epoll file may report `O_RDWR` while not supporting ordinary read or write calls.

I have added more documentation make it clearer.

kernel/src/fs/file/file_handle.rs
```diff
@@ -79,9 +109,20 @@ pub trait FileLike: Pollable + Send + Sync + Any {
...
 
-    fn access_mode(&self) -> AccessMode {
-        AccessMode::O_RDWR
-    }
+    /// Returns the access mode of this file.
+    ///
+    /// This access mode indicates whether the file descriptor is opened for reading or writing.
+    /// It does not indicate whether the file type actually supports the corresponding operation.
+    /// For inode-backed files, it is determined dynamically for each opened file.
+    /// For special files, it is determined statically by each file type.
```
tatetian: It is always a good idea to give some examples:

```rust
/// For special files such as epoll, eventfd, and pid files,
/// it is determined statically by each file type.
```

kernel/src/fs/file/file_handle.rs
tatetian: I got it. So even if `access_mode` returns `O_RDWR`, the real implementations of `read` and `write` may still be missing.

> Even if a file's access_mode is `O_RDWR`, it may still _not_ implement read or write — for example, pid files and epoll files.

```rust
    /// Returns the access mode of this file.
    ///
    /// This access mode indicates whether the file descriptor is opened for reading or writing.
```

But this understanding is still limited and inaccurate. Actually, `access_mode` indicates whether the file supports read-only or/and mutable operations, which includes but are not limited to `read` and `write` methods. This is why `AccessMode` has three variants: `O_RDWR`, `O_WRONLY`, and `O_RDONLY`. There is no the fourth variant that corresponds to the concept of an "empty" mode!

So you are right to change the default implementation, but the doc comments need to be revised.

kernel/src/fs/file/file_handle.rs
```diff
@@ -79,9 +109,20 @@ pub trait FileLike: Pollable + Send + Sync + Any {
...
+    ///
+    /// If the access mode allows an operation
+    /// but the file type does not override the corresponding method,
+    /// the default [`read`] or [`write`] implementation returns `EINVAL`.
+    ///
```
tatetian, line 131: The following revise version is better in my opinion:

```rust
/// Returns the access mode of this file.
///
/// The access mode indicates whether the file descriptor is readable or writable.
/// A readable file allows operations
/// such as `read`, `read_at`, and read-only memory mappings.
/// A writable file allows operations
/// such as `write`, `write_at`, `resize`, `fallocate`,
/// and writable memory mappings.
///
/// For inode-backed files, their access modes are determined dynamically for each opened file.
/// For special files, they are determined statically by the file types.
```

kernel/src/fs/file/file_handle.rs
```diff
@@ -20,12 +20,42 @@ use crate::{
 
 /// The basic operations defined on a file
 pub trait FileLike: Pollable + Send + Sync + Any {
+    /// Reads data from this file.
+    ///
+    /// The default implementation models Linux's distinction between
+    /// a file's open mode and its supported operations:
+    /// it returns `EBADF` if the file was not opened for reading,
+    /// or `EINVAL` if reading is not supported by this file type.
```
tatetian: This paragraph explains what the "distinction" is but does not explicit answer what the default implementation actually is.

kernel/src/fs/file/file_handle.rs
```diff
@@ -20,12 +20,42 @@ use crate::{
...
+    ///
+    /// Implementors should override this method if the file type supports reads.
+    /// An overriding implementation must check whether [`access_mode`] is readable
+    /// before performing the operation.
```
tatetian, line 35: This paragraph is good.

tatetian (approved):

PR #3232, opened 2026-05-15
Enable and fix `redundant_imports` lint

zjp-CN:
See https://github.com/asterinas/asterinas/pull/3036#issuecomment-4458156155

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,6 +109,7 @@ exclude = [
 ]
 
 [workspace.lints.rust]
+redundant_imports = "warn"
 unsafe_op_in_unsafe_fn = "warn"
 unexpected_cfgs = { level = "allow", check-cfg = ['cfg(ktest)'] }
 
--- a/ostd/src/prelude.rs
+++ b/ostd/src/prelude.rs
@@ -5,7 +5,7 @@
 /// A specialized [`Result`] type for this crate.
 ///
 /// [`Result`]: core::result::Result
-pub type Result<T> = core::result::Result<T, crate::error::Error>;
+pub type Result<T, E = crate::error::Error> = core::result::Result<T, E>;
 
 pub(crate) use alloc::{boxed::Box, sync::Arc, vec::Vec};
 
--- a/kernel/comps/block/src/bio.rs
+++ b/kernel/comps/block/src/bio.rs
@@ -304,7 +304,7 @@ impl BioMetadata {
 }
 
 impl IoCompletion for BioMetadata {
-    fn wait(&self) -> core::result::Result<(), IoError> {
+    fn wait(&self) -> Result<(), IoError> {
         let status = self.wait_queue.wait_until(|| {
             let status = self.status();
             (status != BioStatus::Submit).then_some(status)
--- a/kernel/comps/mlsdisk/src/layers/0-bio/block_buf.rs
+++ b/kernel/comps/mlsdisk/src/layers/0-bio/block_buf.rs
@@ -19,11 +19,9 @@
 //! and `&mut [u8]` with `BufMut<[u8]>`.
 
 use alloc::vec;
-use core::convert::TryFrom;
 
 use lending_iterator::prelude::*;
 
-use super::BLOCK_SIZE;
 use crate::prelude::*;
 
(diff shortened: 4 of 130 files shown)


lrh2000:
Thanks! Please try to add it as a workspace lint instead:
https://github.com/asterinas/asterinas/blob/818cfd9da9748d587b8346918b2261bed4e4bda2/Cargo.toml#L111-L117

ostd/src/prelude.rs
```diff
@@ -5,7 +5,7 @@
 /// A specialized [`Result`] type for this crate.
 ///
 /// [`Result`]: core::result::Result
-pub type Result<T> = core::result::Result<T, crate::error::Error>;
+pub type Result<T, E = crate::error::Error> = core::result::Result<T, E>;
```
lrh2000, line 8:
```
> grep -R -- "-> core::result::Result" ostd/**/*.rs kernel/**/*.rs | wc -l
103
```

And this isn't related to this PR, right? In any case, I think doing this means that all of the `core::result::Result` becomes unnecessary, and we should just use `Result<>` instead. Either the core `Result` or the `crate::prelude` `Result` is fine.

Cargo.toml
```diff
@@ -109,6 +109,7 @@ exclude = [
 ]
 
 [workspace.lints.rust]
+redundant_imports = "deny"
```
lrh2000: Why `deny`? We should `warn` all lints. `make check` will deny all warnings. But `make run_kernel` can still work even if the lint hasn't been fixed...

ostd/src/prelude.rs
zjp-CN: Using the new form of declaration can make `Result` compatible with or without `crate::prelude::*`.

The current `Result<T>` results in compilation error:

```rust
error[E0107]: type alias takes 1 generic argument but 2 generic arguments were supplied
  --> kernel/src/thread/oops.rs:44:44
   |
44 | pub fn catch_panics_as_oops<F, R>(f: F) -> Result<R, OopsInfo>
   |                                            ^^^^^^  ---------- help: remove the unnecessary generic argument
   |                                            |
   |                                            expected 1 generic argument
   |
note: type alias defined here, with 1 generic parameter: `T`
  --> kernel/src/prelude.rs:40:10
   |
40 | pub type Result<T> = core::result::Result<T, Error>;
   |          ^^^^^^ -

error[E0308]: mismatched types
   --> kernel/src/thread/oops.rs:64:17
    |
 64 |             Err(*info)
    |             --- ^^^^^ expected `Error`, found `OopsInfo`
    |             |
    |             arguments to this enum variant are incorrect
    |
help: the type constructed contains `thread::oops::OopsInfo` due to the type of the argument passed
   --> kernel/src/thread/oops.rs:64:13
    |
 64 |             Err(*info)
    |             ^^^^-----^
    |                 |
    |                 this argument influences the type of `Err`
note: tuple variant defined here
   --> /home/zjp/.rustup/toolchains/nightly-2026-04-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:566:5
    |
566 |     Err(#[stable(feature = "rust1", since = "1.0.0")] E),
    |     ^^^
```

Another less great fix is using the absolute path 

```diff
-pub fn catch_panics_as_oops<F, R>(f: F) -> Result<R, OopsInfo>
+pub fn catch_panics_as_oops<F, R>(f: F) -> core::result::Result<R, OopsInfo>
```

But that conflicts with the idea "all of the core::result::Result becomes unnecessary".

lrh2000: Okay, thanks for explaining the relationship.

I meant that you could try to add another commit that performs things similar to `sed 's/core::result::Result/Result/' **/*.rs` because your changes make the full path `Result` unnecessary, and the simpler version should be preferred. Since this PR is already tree-wide, can you just include that commit?

```
> grep -R -- "-> core::result::Result" ostd/**/*.rs kernel/**/*.rs | wc -l
103
```
In other words, I mean that you can remove those ^^^

zjp-CN: Done~

lrh2000 (approved):

tatetian (approved):

PR #2081, opened 2025-05-15
Warn `unsafe_op_in_unsafe_fn` by default

lrh2000:
This PR concludes a series of efforts to enable the `unsafe_op_in_unsafe_fn` lint. This lint will be warned by default in Rust 2024. This lint is really good lint that I think deserves to be handled carefully rather than just suppressed.

If we have carefully written all the unsafe code, this lint should not cause much trouble, because we just need to add some `unsafe` blocks!

But the problem is that in some cases I have found that I cannot write proper safety comments for the newly added unsafe block, which means that the original code is buggy and requires some refactoring to fix the underlying unsoundness.

So the refactoring has to be done over several months:
 - https://github.com/asterinas/asterinas/pull/1836
 - https://github.com/asterinas/asterinas/pull/1866
 - https://github.com/asterinas/asterinas/pull/1786
 - https://github.com/asterinas/asterinas/pull/2030
 - https://github.com/asterinas/asterinas/pull/2068

So this PR finally enables the lint, with all newly added `unsafe` blocks properly documented.

---

This PR contains 9 commits, which sounds very awful. This PR touches many modules in the OSTD, and most of the changes are fairly trivial, so I'm not sure what the best way to create commits is. I'm currently grouping similar changes into one commit, rather than grouping changes to a specific module into one commit. If you're more interested in changes to a specific module, you can just check the whole diff and locate the module you're interested in.

Again, almost all the changes are trivial. `` Use `wrapping_add` to add userspace pointers `` can be an exception which needs some more explanation. We previously use `ptr::add` to operate userspace pointers, however, the safety requirements of `ptr::add` [claim](https://doc.rust-lang.org/std/primitive.pointer.html#safety-7) that:
> If the computed offset is non-zero, then self must be [derived from](https://doc.rust-lang.org/std/ptr/index.html#provenance) a pointer to some [allocated object](https://doc.rust-lang.org/std/ptr/index.html#allocated-object)

[...]

PR #3234, opened 2026-05-15
Remove explicit `unsafe_op_in_unsafe_fn`

lrh2000:
The `unsafe_op_in_unsafe_fn` lint was added in https://github.com/asterinas/asterinas/pull/2081 when we were still using Rust 2021.

It is now a `warn-by-default` lint in Rust 2024, so we don't need to list the lint explicitly. I have verified that the lint fires normally after removing it from the explicit warning list.

There are three other minor problems.
 1. `ostd::prelude` re-exports `ostd::error`. It is meant to re-export the log macro, but it also brings the `error` module into scope. This module is not visible to users outside of OSTD, so it does not seem to have much impact. However, I'd like to explicitly note this fact since the `unused_qualifications` lint successfully made me aware of this issue.
    - I could not find a valid solution to this. It appears that Rust does not have a syntax that allows us to re-export only the macro.
 2. Add a missing blank line in `cpio-decoder`.
 3. Sort the list of lint in alphabetical order.

cc @zjp-CN as it's a small follow-up to your PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,9 +110,8 @@ exclude = [
 
 [workspace.lints.rust]
 redundant_imports = "warn"
-unused_qualifications = "warn"
-unsafe_op_in_unsafe_fn = "warn"
 unexpected_cfgs = { level = "allow", check-cfg = ['cfg(ktest)'] }
+unused_qualifications = "warn"
 
 [workspace.lints.clippy]
 allow_attributes = "warn"
--- a/kernel/libs/cpio-decoder/src/lib.rs
+++ b/kernel/libs/cpio-decoder/src/lib.rs
@@ -19,6 +19,7 @@
 #![deny(unsafe_code)]
 
 extern crate alloc;
+
 #[cfg(not(test))]
 use alloc::string::{String, ToString};
 use alloc::vec;
--- a/ostd/src/prelude.rs
+++ b/ostd/src/prelude.rs
@@ -5,7 +5,11 @@
 /// A specialized [`Result`] type for this crate.
 ///
 /// [`Result`]: core::result::Result
-pub type Result<T, E = error::Error> = core::result::Result<T, E>;
+#[expect(
+    unused_qualifications,
+    reason = "`error` below is intended to re-export the macro but brings the module into scope"
+)]
+pub type Result<T, E = crate::error::Error> = core::result::Result<T, E>;
 
 pub(crate) use alloc::{boxed::Box, sync::Arc, vec::Vec};
 


zjp-CN:
Thanks for the styling improvement.



> * I could not find a valid solution to this. It appears that Rust does not have a syntax that allows us to re-export only the macro.

For the ambiguity of `ostd::error` and the intent of reexporting the macro only, we can declare `error!` via [macro 2.0](https://lukaswirth.dev/tlborm/decl-macros/macros2.html#macro-are-proper-items):

```diff
-#[macro_export]
-macro_rules! error {
-    ($($arg:tt)+) => { $crate::log!($crate::log::Level::Error, $($arg)+) };
+pub macro error ($($arg:tt)+) {
+    $crate::log!($crate::log::Level::Error, $($arg)+)
}
```

Then we can `pub use crate::log::macros::error` in `ostd::prelude`, and error is not a module so that `E = crate::error::Error` must be used in the Result type alias.

The newly added decl macros in libstd are all macro 2.0 to have path-based accessibity: [pin!](https://doc.rust-lang.org/stable/std/pin/macro.pin.html), [addr_of!](https://doc.rust-lang.org/stable/std/ptr/macro.addr_of.html), [iter!](https://doc.rust-lang.org/stable/std/iter/macro.iter.html) etc.

tatetian:
> 1. `ostd::prelude` re-exports `ostd::error`. It is meant to re-export the log macro, but it also brings the `error` module into scope. This module is not visible to users outside of OSTD, so it does not seem to have much impact. However, I'd like to explicitly note this fact since the `unused_qualifications` lint successfully made me aware of this issue.
> * I could not find a valid solution to this. It appears that Rust does not have a syntax that allows us to re-export only the macro.

I am not aware of this ambiguity. Thanks for sharing.

> For the ambiguity of `ostd::error` and the intent of reexporting the macro only, we can declare `error!` via [macro 2.0](https://lukaswirth.dev/tlborm/decl-macros/macros2.html#macro-are-proper-items):

This is news to me, too! I think this is the right choice. Since the [hygiene](https://lukaswirth.dev/tlborm/decl-macros/macros2.html#hygiene) rule has subtle differences between version 1.0 and 2.0, we cannot replace every macro will the new `macro` keyword, but in the specific case of the `log`-family macros, we can.

zjp-CN:
> Since the [hygiene](https://lukaswirth.dev/tlborm/decl-macros/macros2.html#hygiene) rule has subtle differences between version 1.0 and 2.0, we cannot replace every macro will the new `macro` keyword

For the linked example, we can still make it compile by providing explicit tokens on definition site: [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=e119d3d74e13bef40fa21409f5ca7ddf).

tatetian:
> > Since the [hygiene](https://lukaswirth.dev/tlborm/decl-macros/macros2.html#hygiene) rule has subtle differences between version 1.0 and 2.0, we cannot replace every macro will the new `macro` keyword
> 
> For the linked example, we can still make it compile by providing explicit tokens on definition site: [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=e119d3d74e13bef40fa21409f5ca7ddf).

Does it mean that we can and should replace all `macro_rules!` with the `macro` keyword?

zjp-CN:
No, I didn't mean to refactor all macro declarations.

But I indeed meet some obscure macro path problems when applying short_vis_path. 

I didn't fully figure out the compilation regression, but a quick [fix](https://github.com/asterinas/asterinas/compare/main...os-checker:asterinas:short_vis_path#diff-80b9340a8675a2d6a1075730f2bba82d73776d61cb75bd64a841bc07a46a7565) for me is marking parent modules `#[macro_use]` and moving it before any other modules to globally use `#[macro_export] macro_rules!` macros, instead of using them based on module paths. 


Not blocking this PR, I will share more details in short_vis_path implementation. The crate has been made, and asterinas compiles locally. Just need some clean-ups before opening the PR.

lrh2000:
> the specific case of the `log`-family macros, we can.

Really? I don't think so.

[Rust playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=0ccee58f3f20d5a9fb5ec1c4d4701764):
```rust
#![feature(decl_macro)]

// Set this crate's log prefix for `ostd::log`.
macro_rules! __log_prefix {
    () => {
        "i8042: "
    };
}

macro foo {
    ($name:ident : $new:ident) => {
        __log_prefix!()
    }
}

fn main() {
    let _prefix = foo!(Foo: new);
}
```
produces:
```
error: cannot find macro `__log_prefix` in this scope
  --> src/main.rs:12:9
   |
12 |         __log_prefix!()
   |         ^^^^^^^^^^^^
...
17 |     let _prefix = foo!(Foo: new);
   |                   -------------- in this macro invocation
   |
   = note: this error originates in the macro `foo` (in Nightly builds, run with -Z macro-backtrace for more info)
```

tatetian:
> > the specific case of the `log`-family macros, we can.
> 
> Really? I don't think so.

Good experiment. Macro 2.0 cannot reference the textual-scoped Macro 1.0 and that makes sense. This specific example can be easily fixed by turning `__log_prefix` to a [Macro 2.0 version](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=7fbed328fd853c1eb184c833726ad6ec).

However, the `__log_prefix` fundamentally relies on the nature of textual scoping to make it possible to specify a module-level log prefix, overriding the crate-level one. I don't think this can be done with Macro 2.0.

tatetian:
> * However, I'd like to explicitly note this fact since the `unused_qualifications` lint successfully made me aware of this issue.
>   
>   * I could not find a valid solution to this. It appears that Rust does not have a syntax that allows us to re-export only the macro.

Another option is to rename the `error` module, or even remove the module. But I suppose this would be an overkill. Let's keep the `expect` attribute for the lint error.

zjp-CN:
It's surprising to me! Looks like macro 2.0 is hard to use when [referencing some identifiers](https://github.com/rust-lang/rust/issues/39412#issuecomment-3476907414). 

Another fix is defining `error!` as  a proc macro in ostd-macros, and `pub use ostd_macros::error` in prelude, which means `crate::error`  only refers to the module.

tatetian (approved):

tatetian:
> Another fix is defining `error!` as a proc macro in ostd-macros

This would be an overkill (defining a proc macro just to avoid a false-positive lint error) and a surprise (all log macros are declarative macros except this one).

zjp-CN:
Just realized `crate::Error` is usable, so the perfect fix can be

```diff
-pub type Result<T, E = crate::error::Error> = core::result::Result<T, E>;
+pub type Result<T, E = crate::Error> = core::result::Result<T, E>;
```

lrh2000:
> Just realized `crate::Error` is usable, so the perfect fix can be

I intentionally want to point out that the following `error` brings the module into scope as well. Inside OSTD, for example, using the statement `use crate::prelude::*` also brings the `error` module into scope. It is just a bit surprising.

It could be noted with a comment, though. I think either noting it with a comment or an `expect` reason is fine. Both have their own rationale.

zjp-CN:
Yeah, that makes sense.

PR #3235, opened 2026-05-17
Refactor `Dentry` variants to pave the way for `O_TMPFILE`

tatetian:
## Motivation

PR #3185 implements `O_TMPFILE` support for `open`/`openat`. While designing that work it became clear that an `O_TMPFILE` file does not fit any existing `Dentry` shape: it is a *real* filesystem inode that has a parent directory and lives on a real mount, yet it has no name and is unreachable by path lookup until a later `linkat(2)` promotes it. Forcing it through the existing `Pseudo` shape (pipes/sockets/anon-inode/memfd, which live on private pseudo-mounts) would be semantically wrong and would mis-render `/proc/<pid>/fd/<n>`.

This PR extracts the purely structural `Dentry` groundwork out of #3185 so it can be reviewed on its own, with **no behavior change**. The actual `O_TMPFILE` syscall logic stays in #3185 and builds on top of this.

## What this PR does

- **Rename the `Leaf` dentry option to `Named`.** `Leaf` was a misnomer (a directory that has children is still a `Leaf`); `Named` states the real invariant (a named entry under a parent). Purely internal rename — no public API uses the word "leaf".
- **Add an `Anonymous` dentry option.** A real inode parented to a directory but deliberately kept out of the parent's children cache, so path lookup cannot reach it — the shape an `O_TMPFILE` file needs. It is currently unconstructed (annotated `#[expect(dead_code)]`) and is wired up by #3185; `linkat`-based promotion to `Named` is likewise left to #3185.
- **Document the taxonomy and the dentry/inode relationship.** The type-level docs now describe the four variants (`Root`, `Named`, `Anonymous`, `Pseudo`) with a decision-tree diagram, plus a new "Dentries vs inodes" section.

The three pre-existing shapes behave exactly as before; only the rename, the new (currently unused) variant, and documentation are added.

Related: #3185.

### DIFF
--- a/kernel/src/fs/vfs/path/dentry.rs
+++ b/kernel/src/fs/vfs/path/dentry.rs
@@ -90,6 +90,44 @@ use crate::{
 /// |  ordinary file or directory |
 /// +-----------------------------+
 /// ```
+///
+/// # Dentries vs inodes
+///
+/// An [`Inode`] *is* the filesystem object:
+/// it owns the file's data and metadata
+/// (type, size, mode, owner, timestamps, link count)
+/// and is implemented per filesystem (e.g. ramfs, ext2).
+/// An inode has no inherent name and no location in the directory tree.
+///
+/// A `Dentry` is the VFS-level cache node that *names* an inode
+/// and places it in the path namespace.
+/// Every `Dentry` references exactly one inode (held as `Arc<dyn Inode>`),
+/// but the reverse is not one-to-one:
+///
+/// - **One inode, many dentries.**
+///   Hard links are several `Named` dentries referencing the same inode;
+///   the inode's link count tracks how many.
+/// - **One inode, no reachable dentry.**
+///   A file unlinked while still open,
+///   or an `Anonymous` (`O_TMPFILE`) inode,
+///   lives on with no `Named` dentry, so path lookup cannot reach it.
+/// - **Directories are one-to-one.**
+///   A directory inode cannot be hard-linked,
+///   so it has exactly one `Named` (or `Root`) dentry;
+///   only the `Dentry` layer holds the parent/child structure
+///   and the children cache that path lookup walks.
+///
+/// Rule of thumb:
+/// operations on *content and metadata*
+/// (`read`, `write`, `stat`, `chmod`)
+/// act on the [`Inode`];
+/// operations on *names and tree shape*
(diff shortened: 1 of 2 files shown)


tatetian:

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -22,6 +22,110 @@ use crate::{
...
+/// # Dentries vs inodes
+///
+/// An [`Inode`] *is* the filesystem object:
+/// it owns the file's data and metadata
+/// (type, size, mode, owner, timestamps, link count)
+/// and is implemented per filesystem (e.g. ramfs, ext2).
+/// An inode has no inherent name and no location in the directory tree.
+///
+/// A `Dentry` is the VFS-level cache node that *names* an inode
+/// and places it in the path namespace.
```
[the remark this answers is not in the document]
tatetian: At a high level, the job of `Dentry` is to give inode names. I think it is ok to say this in the summary line.

kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -92,6 +212,20 @@ impl Dentry {
         Self::new(inode, DentryOptions::Root)
     }
 
+    /// Creates a new anonymous `Dentry`:
+    /// a real inode parented to `parent`
+    /// but kept out of `parent`'s children cache,
+    /// so it is unreachable by path lookup.
```
cchanging: This definition of anonymous `Dentry` already appears in the documentation for `Dentry`. I think we could just reference that directly to avoid repeating the explanation.

kernel/src/fs/vfs/path/dentry.rs
cchanging, line 793: After introducing anonymous `Dentry`, I think the documentation for `DentryKey` should be updated accordingly: for anonymous dentries, their `DentryKey` is meaningless and should not be used. As things stand, if an anonymous dentry were to actually have a `DentryKey`, it might not be unique.

With that clarified, we could also add a `debug_assert!` in `Dentry::key()` to ensure that `key()` is never called on an anonymous `Dentry`.

cchanging (commented):
Thanks for providing such a detailed classification and definition of `Dentry`! The new changes and definitions introduced in this PR make sense to me.

cchanging (approved):

PR #3236, opened 2026-05-18
Add patches for GDB

vvvvsv:
Goes after https://github.com/asterinas/asterinas/pull/3246. Required for #2323.

With these patches, we can debug single-threaded processes with GDB.

The first patch emulates the default state of the x86 debug registers, so it can correctly respond when a tracer reads the tracee’s debug registers via `PTRACE_PEEKUSER`. In the current main branch, the tracee’s x86 debug registers are never actually modified, so they always remain at their default values.

The second patch suppresses the GDB warning "Error disabling address space randomization". This patch will be removed once `personality(ADDR_NO_RANDOMIZE)` is supported.

### DIFF
--- a/.github/workflows/push_cachix_dev.yml
+++ b/.github/workflows/push_cachix_dev.yml
@@ -21,7 +21,7 @@ jobs:
     container:
       image: asterinas/asterinas:0.17.2-20260508
       options: -v /dev:/dev --privileged
-    timeout-minutes: 60
+    timeout-minutes: 120
     steps:
       - uses: actions/checkout@v4
 
--- a/distro/aster_nixos_installer/templates/aster_configuration.nix
+++ b/distro/aster_nixos_installer/templates/aster_configuration.nix
@@ -32,6 +32,7 @@
   # please refer to https://nixos.org/manual/nixpkgs/stable/#sec-overlays-definition.
   config.nixpkgs.overlays = [
     (import ./overlays/desktop/default.nix)
+    (import ./overlays/gdb/default.nix)
     (import ./overlays/hello-asterinas/default.nix)
     (import ./overlays/podman/default.nix)
     (import ./overlays/switch-to-configuration-ng/default.nix)
--- a/distro/cachix/default.nix
+++ b/distro/cachix/default.nix
@@ -7,6 +7,7 @@ let
   nixos = pkgs.nixos (import "${installer}/etc_nixos/configuration.nix");
   cachixPkgs = with nixos.pkgs;
     [
+      gdb
       hello-asterinas
       xfce.xfdesktop
       xfce.xfwm4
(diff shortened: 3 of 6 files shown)


tatetian:
@vvvvsv Could you also add a GDB documentation to the "Popular Applications" section of the Asterinas book as well as the CI tests? This way we can be sure that the patches actually work.

tatetian:
Could you diagnose the root cause of the failed `push-pkgs`? It looks an timeout error. Is it unavoidable or is there anything we can do about it (such as speeding up the build process or optimizing the CI workflow)?

tatetian:
> The first patch emulates the default state of the x86 debug registers, so it can correctly respond when a tracer reads the tracee’s debug registers via `PTRACE_PEEKUSER`. In the current main branch, the tracee’s x86 debug registers are never actually modified, so they always remain at their default values.
> 
> The second patch suppresses the GDB warning "Error disabling address space randomization". This patch will be removed once `personality(ADDR_NO_RANDOMIZE)` is supported.

These are some very useful and well written explanation for the patches. Could you move the text into the patches?

tatetian:
> @vvvvsv Could you also add a GDB documentation to the "Popular Applications" section of the Asterinas book as well as the CI tests? This way we can be sure that the patches actually work.

I am looking forward to the new documentation and CI. GDB test cases should be added to the [development-tools](https://github.com/asterinas/asterinas/blob/main/test/nixos/tests/development-tools/src/main.rs) test suite.

vvvvsv:
> I am looking forward to the new documentation and CI. GDB test cases should be added to the [development-tools](https://github.com/asterinas/asterinas/blob/main/test/nixos/tests/development-tools/src/main.rs) test suite.

I will add them after successfully pushing cacheix. I tried adding them yesterday, but that caused the Test AsterNixOS (full) to re-compile everything and timeout.

vvvvsv:
> Could you diagnose the root cause of the failed `push-pkgs`? It looks an timeout error. Is it unavoidable or is there anything we can do about it (such as speeding up the build process or optimizing the CI workflow)?

Seems like this timeout is unavoidable, since GDB is depended on by too many packages. For example, the log shows that gtk4 was rebuilt, and Codex found this dependency chain:

gdb -> valgrind -> pipewire -> openal-soft -> gst-plugins-bad -> gtk4

Could we set this CI timeout to 2 hours? @tatetian 

---

Or could we add a package similar to `aster_systemd`, called `aster_gdb`? In this way, users would need to add `aster_gdb` to `environment.systemPackages` if they want to install gdb, which seems a bit unnatural.

vvvvsv:
It seems that even 2h is not enough, since `webkitgtk` is huge: https://github.com/asterinas/asterinas/actions/runs/26145123047/job/76898790056?pr=3236

It would be better to add the patches on the kernel side instead.

PR #3237, opened 2026-05-18
Fix book links and add link check

zjp-CN:
This PR improves book links by
* updating the version in api-docs links from v0.17.1 to v0.17.2
* fixing dead links
* adding `check_dead_links` and `check_latest_version` in publish_website workflow
  * the VERSION file change will trigger the workflow 
* notifying the api-docs containing latest version are available, and triggering publish_website again

### DIFF
--- a/.github/workflows/publish_website.yml
+++ b/.github/workflows/publish_website.yml
@@ -4,14 +4,16 @@ on:
   workflow_dispatch:
   pull_request:
     paths:
+      - VERSION
       - book/**
-      - .github/workflows/update-website.yml
+      - .github/workflows/publish_website.yml
   push:
     branches:
       - main
     paths:
+      - VERSION
       - book/**
-      - .github/workflows/update-website.yml
+      - .github/workflows/publish_website.yml
 
 jobs:
   build_and_deploy:
@@ -23,12 +25,29 @@ jobs:
         with:
           repository: 'asterinas/asterinas'
           path: 'asterinas'
-      
+
+      - name: Cache .lycheecache
+        uses: actions/cache@v6
+        with:
+          path: asterinas/book/.lycheecache
+          key: lychee-${{ runner.os }}-${{ github.ref_name }}-${{ github.sha }}
+          # Cache lookup strategy:
+          # 1) Exact key (os + ref_name + sha), hit on rerun of the same commit.
+          # 2) If exact key misses, restore by os + ref_name prefix; newest match wins.
+          # 3) Final fallback is os-only prefix; newest match wins.
+          # Multiple restore-keys lines are intentional for progressive fallback (from specific to broad).
+          restore-keys: |
+            lychee-${{ runner.os }}-${{ github.ref_name }}
+            lychee-${{ runner.os }}-
+
       - name: Build the website
+        env:
+          # Avoid hitting rate limit when lychee accesses github links.
+          GITHUB_TOKEN: ${{ github.token }}
         run: |
           cd asterinas/book
-          mdbook-mermaid install
-          mdbook build
+          make check
+          make
 
       - name: Deploy website
         if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
--- a/.github/workflows/publish_website.yml
+++ b/.github/workflows/publish_website.yml
@@ -4,14 +4,16 @@ on:
   workflow_dispatch:
   pull_request:
     paths:
+      - VERSION
       - book/**
-      - .github/workflows/update-website.yml
+      - .github/workflows/publish_website.yml
   push:
     branches:
       - main
     paths:
+      - VERSION
       - book/**
-      - .github/workflows/update-website.yml
+      - .github/workflows/publish_website.yml
 
 jobs:
   build_and_deploy:
@@ -24,11 +26,28 @@ jobs:
           repository: 'asterinas/asterinas'
           path: 'asterinas'
 
+      - name: Cache .lycheecache
+        uses: actions/cache@v6
+        with:
+          path: asterinas/book/.lycheecache
+          key: lychee-${{ runner.os }}-${{ github.ref_name }}-${{ github.sha }}
+          # Cache lookup strategy:
+          # 1) Exact key (os + ref_name + sha), hit on rerun of the same commit.
+          # 2) If exact key misses, restore by os + ref_name prefix; newest match wins.
+          # 3) Final fallback is os-only prefix; newest match wins.
+          # Multiple restore-keys lines are intentional for progressive fallback (from specific to broad).
+          restore-keys: |
+            lychee-${{ runner.os }}-${{ github.ref_name }}
+            lychee-${{ runner.os }}-
+
       - name: Build the website
+        env:
+          # Avoid hitting rate limit when lychee accesses github links.
+          GITHUB_TOKEN: ${{ github.token }}
         run: |
           cd asterinas/book
-          mdbook-mermaid install
-          mdbook build
+          make
+          make check
 
       - name: Deploy website
         if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
--- a/book/Makefile
+++ b/book/Makefile
@@ -1,0 +1,19 @@
+# SPDX-License-Identifier: MPL-2.0
+
+.PHONY: build
+build: mermaid.min.js mermaid-init.js
+	@mdbook build
+
+mermaid.min.js mermaid-init.js:
+	@mdbook-mermaid install
+
+.PHONY: check
+check: check_rust_api_doc_version check_invalid_or_broken_links
+
+.PHONY: check_rust_api_doc_version
+check_rust_api_doc_version:
+	@./check_rust_api_doc_version.sh
+
+.PHONY: check_invalid_or_broken_links
+check_invalid_or_broken_links:
+	@./check_invalid_or_broken_links.sh
--- a/book/Makefile
+++ b/book/Makefile
@@ -1,0 +1,19 @@
+# SPDX-License-Identifier: MPL-2.0
+
+.PHONY: build
+build: mermaid.min.js mermaid-init.js
+	@mdbook build
+
+mermaid.min.js mermaid-init.js:
+	@mdbook-mermaid install
+
+.PHONY: check
+check: check_rust_api_doc_version check_invalid_or_broken_links
+
+.PHONY: check_rust_api_doc_version
+check_rust_api_doc_version:
+	@./check_rust_api_doc_version.sh
+
+.PHONY: check_invalid_or_broken_links
+check_invalid_or_broken_links: build
+	@./check_invalid_or_broken_links.sh
--- a/book/check_invalid_or_broken_links.sh
+++ b/book/check_invalid_or_broken_links.sh
@@ -1,0 +1,116 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+# Temporary files:
+# - JSON_FILE stores lychee JSON output.
+# - URLS_FILE stores failed http/https links extracted from lychee output.
+# - CACHE_UPDATES_FILE stores recovered links to write back into .lycheecache.
+# - CACHE_MERGED_FILE is the merged cache output before replace.
+JSON_FILE="$(mktemp /tmp/lychee-result-XXXXXX.json)"
+URLS_FILE="$(mktemp /tmp/lychee-failed-urls-XXXXXX.txt)"
+CACHE_UPDATES_FILE="$(mktemp /tmp/lychee-cache-updates-XXXXXX.txt)"
+CACHE_MERGED_FILE="$(mktemp /tmp/lychee-cache-merged-XXXXXX.txt)"
+
+cleanup() {
+  rm -f "$JSON_FILE" "$URLS_FILE" "$CACHE_UPDATES_FILE" "$CACHE_MERGED_FILE"
+}
+trap cleanup EXIT
+
+print_lychee_summary() {
+  echo "lychee summary:"
+  echo "  total.......$LYCHEE_TOTAL"
+  echo "  unique......$LYCHEE_UNIQUE"
+  echo "  successful..$LYCHEE_SUCCESSFUL"
+  echo "  timeouts....$LYCHEE_TIMEOUTS"
+  echo "  redirected..$LYCHEE_REDIRECTS"
+  echo "  excluded....$LYCHEE_EXCLUDED"
+  echo "  unknown.....$LYCHEE_UNKNOWN"
+  echo "  errors......$LYCHEE_ERRORS"
+  echo "  unsupported.$LYCHEE_UNSUPPORTED"
+}
+
+# Run lychee once and capture JSON output for fallback processing.
+echo "[1/3] Running lychee..."
+LYCHEE_EXIT=0
+if ! lychee --config lychee.toml --format json --no-progress src/ >"$JSON_FILE"; then
+  LYCHEE_EXIT=$?
+fi
+
+# Collect unique failed http/https links from lychee error/timeout maps.
+jq -r '
+  [
+    (.error_map // {} | to_entries[]? | .value[]? | .url?),
+    (.timeout_map // {} | to_entries[]? | .value[]? | .url?)
+  ]
+  | map(select(type == "string" and test("^https?://")))
+  | unique
+  | .[]
+' "$JSON_FILE" >"$URLS_FILE"
+
+# Print lychee raw stats (same data source as detailed format summary).
+LYCHEE_TOTAL="$(jq -r '.total // 0' "$JSON_FILE")"
+LYCHEE_UNIQUE="$(jq -r '.unique // 0' "$JSON_FILE")"
+LYCHEE_SUCCESSFUL="$(jq -r '.successful // 0' "$JSON_FILE")"
+LYCHEE_TIMEOUTS="$(jq -r '.timeouts // 0' "$JSON_FILE")"
+LYCHEE_REDIRECTS="$(jq -r '.redirects // 0' "$JSON_FILE")"
+LYCHEE_EXCLUDED="$(jq -r '.excludes // 0' "$JSON_FILE")"
+LYCHEE_UNKNOWN="$(jq -r '.unknown // 0' "$JSON_FILE")"
+LYCHEE_ERRORS="$(jq -r '.errors // 0' "$JSON_FILE")"
+LYCHEE_UNSUPPORTED="$(jq -r '.unsupported // 0' "$JSON_FILE")"
+
+TOTAL_FAILED_BY_LYCHEE="$(wc -l <"$URLS_FILE" | tr -d ' ')"
+echo "lychee failed links: $TOTAL_FAILED_BY_LYCHEE"
+if [ "$TOTAL_FAILED_BY_LYCHEE" -eq 0 ]; then
+  print_lychee_summary
+  exit "$LYCHEE_EXIT"
+fi
+
+# Retry failed links using plain curl; report recovered/failed links.
+echo "[2/3] Curl fallback on lychee-failed links..."
+RECOVERED=0
+FAILED=0
+while IFS= read -r url; do
+  [ -z "$url" ] && continue
+  STATUS_CODE="$(curl -o /dev/null -w '%{http_code}' "$url" 2>/dev/null)"
+  CURL_EXIT=$?
+  if [ "$CURL_EXIT" -eq 0 ]; then
+    RECOVERED=$((RECOVERED + 1))
+    echo "RECOVERED: $url (status=$STATUS_CODE)"
+    # Status code is recorded as 200 in .lycheecache to let lychee skip the link next time.
+    printf '%s,%s,%s\n' "$url" "200" "$(date +%s)" >>"$CACHE_UPDATES_FILE"
+  else
+    FAILED=$((FAILED + 1))
+    echo "FAILED:    $url"
+  fi
+done <"$URLS_FILE"
+
+# Write recovered links into .lycheecache as url,status,timestamp.
+UPDATED_CACHE_LINES=0
+if [ -s "$CACHE_UPDATES_FILE" ]; then
+  if [ -f .lycheecache ]; then
+    # Merge cache with URL-level deduplication:
+    # - First pass (updates): keep the latest recovered record by URL.
+    # - Second pass (existing cache): keep only URLs not present in updates.
+    # - End: append updated records so recovered entries replace old ones.
+    awk -F, '
+      NR==FNR { new[$1]=$0; order[++n]=$1; next }
+      !($1 in new) { print }
+      END { for (i=1; i<=n; i++) print new[order[i]] }
+    ' "$CACHE_UPDATES_FILE" .lycheecache >"$CACHE_MERGED_FILE"
+  else
+    cat "$CACHE_UPDATES_FILE" >"$CACHE_MERGED_FILE"
+  fi
+  mv "$CACHE_MERGED_FILE" .lycheecache
+  UPDATED_CACHE_LINES="$(wc -l <"$CACHE_UPDATES_FILE" | tr -d ' ')"
+fi
+
+echo "[3/3] Summary: lychee_failed=$TOTAL_FAILED_BY_LYCHEE recovered=$RECOVERED still_failed=$FAILED"
+echo "[3/3] Cache updated lines: $UPDATED_CACHE_LINES"
+print_lychee_summary
+if [ "$FAILED" -gt 0 ]; then
+  exit 1
+fi
+exit 0
--- a/book/check_invalid_or_broken_links.sh
+++ b/book/check_invalid_or_broken_links.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: MPL-2.0
+
+# Temporary files:
+# - `JSON_FILE` stores lychee JSON output.
+# - `URLS_FILE` stores failed http/https links extracted from lychee output.
+# - `CACHE_UPDATES_FILE` stores recovered links to write back into .lycheecache.
+# - `CACHE_MERGED_FILE` is the merged cache output before replace.
+OUT_DIR="$(mktemp -d /tmp/lychee-XXXXXX)"
+JSON_FILE="$OUT_DIR"/result.json
+URLS_FILE="$OUT_DIR"/failed-urls.txt
+CACHE_UPDATES_FILE="$OUT_DIR"/cache-updates.txt
+CACHE_MERGED_FILE="$OUT_DIR"/cache-merged.txt
+
+# Remove `/tmp/lychee` files when the script exits.
+cleanup() {
+  rm -rf "$OUT_DIR"
+}
+trap cleanup EXIT
+
+print_lychee_summary() {
+  echo "lychee summary:"
+  echo "  total.......$LYCHEE_TOTAL"
+  echo "  unique......$LYCHEE_UNIQUE"
+  echo "  successful..$LYCHEE_SUCCESSFUL"
+  echo "  timeouts....$LYCHEE_TIMEOUTS"
+  echo "  redirected..$LYCHEE_REDIRECTS"
+  echo "  excluded....$LYCHEE_EXCLUDED"
+  echo "  unknown.....$LYCHEE_UNKNOWN"
+  echo "  errors......$LYCHEE_ERRORS"
+  echo "  unsupported.$LYCHEE_UNSUPPORTED"
+}
+
+# Run lychee once and capture JSON output for fallback processing.
+echo "[1/3] Running lychee..."
+LYCHEE_EXIT=0
+if ! lychee --config lychee.toml --format json --no-progress book/**/*.html >"$JSON_FILE"; then
+  LYCHEE_EXIT=$?
+fi
+
+# Collect unique failed http/https links from lychee error/timeout maps.
+jq -r '
+  [
+    (.error_map // {} | to_entries[]? | .value[]? | .url?),
+    (.timeout_map // {} | to_entries[]? | .value[]? | .url?)
+  ]
+  | map(select(type == "string"))
+  | unique
+  | .[]
+' "$JSON_FILE" >"$URLS_FILE"
+
+# Print lychee raw stats (same data source as detailed format summary).
+LYCHEE_TOTAL="$(jq -r '.total // 0' "$JSON_FILE")"
+LYCHEE_UNIQUE="$(jq -r '.unique // 0' "$JSON_FILE")"
+LYCHEE_SUCCESSFUL="$(jq -r '.successful // 0' "$JSON_FILE")"
+LYCHEE_TIMEOUTS="$(jq -r '.timeouts // 0' "$JSON_FILE")"
+LYCHEE_REDIRECTS="$(jq -r '.redirects // 0' "$JSON_FILE")"
+LYCHEE_EXCLUDED="$(jq -r '.excludes // 0' "$JSON_FILE")"
+LYCHEE_UNKNOWN="$(jq -r '.unknown // 0' "$JSON_FILE")"
+LYCHEE_ERRORS="$(jq -r '.errors // 0' "$JSON_FILE")"
+LYCHEE_UNSUPPORTED="$(jq -r '.unsupported // 0' "$JSON_FILE")"
+
+TOTAL_FAILED_BY_LYCHEE="$(wc -l <"$URLS_FILE" | tr -d ' ')"
+echo "lychee failed links: $TOTAL_FAILED_BY_LYCHEE"
+if [ "$TOTAL_FAILED_BY_LYCHEE" -eq 0 ]; then
+  print_lychee_summary
+  echo "All links pass!"
+  exit "$LYCHEE_EXIT"
+fi
+
+# Retry failed links using plain curl; report recovered/failed links.
+echo "[2/3] Curl fallback on lychee-failed links..."
+RECOVERED=0
+FAILED=0
+while IFS= read -r url; do
+  [ -z "$url" ] && continue
+  if [[ $url == file://* ]]; then
+    # Just warn against this local file link is not accessible without further check.
+    FAILED=$((FAILED + 1))
+    echo "FAILED:    $url"
+    continue
+  fi
+
+  STATUS_CODE="$(curl -o /dev/null -w '%{http_code}' "$url" 2>/dev/null)"
+  CURL_EXIT=$?
+  if [ "$CURL_EXIT" -eq 0 ]; then
+    RECOVERED=$((RECOVERED + 1))
+    echo "RECOVERED: $url (status=$STATUS_CODE)"
+    # Status code is recorded as 200 in .lycheecache to let lychee skip the link next time.
+    printf '%s,%s,%s\n' "$url" "200" "$(date +%s)" >>"$CACHE_UPDATES_FILE"
+  else
+    FAILED=$((FAILED + 1))
+    echo "FAILED:    $url"
+  fi
+done <"$URLS_FILE"
+
+# Write recovered links into .lycheecache as url,status,timestamp.
+UPDATED_CACHE_LINES=0
+if [ -s "$CACHE_UPDATES_FILE" ]; then
+  if [ -f .lycheecache ]; then
+    # Merge cache with URL-level deduplication:
+    # - First pass (updates): keep the latest recovered record by URL.
+    # - Second pass (existing cache): keep only URLs not present in updates.
+    # - End: append updated records so recovered entries replace old ones.
+    awk -F, '
+      NR==FNR { new[$1]=$0; order[++n]=$1; next }
+      !($1 in new) { print }
+      END { for (i=1; i<=n; i++) print new[order[i]] }
+    ' "$CACHE_UPDATES_FILE" .lycheecache >"$CACHE_MERGED_FILE"
+  else
+    cat "$CACHE_UPDATES_FILE" >"$CACHE_MERGED_FILE"
+  fi
+  mv "$CACHE_MERGED_FILE" .lycheecache
+  UPDATED_CACHE_LINES="$(wc -l <"$CACHE_UPDATES_FILE" | tr -d ' ')"
+fi
+
+echo "[3/3] Summary: lychee_failed=$TOTAL_FAILED_BY_LYCHEE recovered=$RECOVERED still_failed=$FAILED"
+echo "[3/3] Cache updated lines: $UPDATED_CACHE_LINES"
+print_lychee_summary
+if [ "$FAILED" -gt 0 ]; then
+  exit 1
+fi
+exit 0
--- a/book/check_rust_api_doc_version.sh
+++ b/book/check_rust_api_doc_version.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && cd .. && pwd)
+VERSION_FILE="${SCRIPT_DIR}/VERSION"
+BOOK_SRC_DIR="${SCRIPT_DIR}/book/src"
+
+TARGET_VERSION=$(tr -d '[:space:]' <"${VERSION_FILE}")
+echo "Target version for validation: ${TARGET_VERSION}"
+
+echo "Scanning directory: ${BOOK_SRC_DIR}"
+
+# Define the search pattern.
+# This regex looks for 'asterinas.github.io/api-docs/'
+# followed by any version number that IS NOT '${TARGET_VERSION}/'.
+# (?!${ESCAPED_TARGET_VERSION}/) is a negative lookahead.
+ESCAPED_TARGET_VERSION=${TARGET_VERSION//./\\.}
+PATTERN="asterinas\\.github\\.io/api-docs/(?!${ESCAPED_TARGET_VERSION}/)[0-9]+\\.[0-9]+\\.[0-9]+"
+MISMATCHES=$(grep -rPn "${PATTERN}" "${BOOK_SRC_DIR}" || true)
+
+if [ -n "${MISMATCHES}" ]; then
+  echo "----------------------------------------------------------------"
+  echo "ERROR: Found links with outdated or incorrect versions:"
+  echo "${MISMATCHES}"
+  echo "----------------------------------------------------------------"
+  echo "Please update the links above to match version ${TARGET_VERSION}."
+  exit 1
+fi
+
+echo "SUCCESS: All found links match version ${TARGET_VERSION}."
--- a/book/check_rust_api_doc_version.sh
+++ b/book/check_rust_api_doc_version.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+ASTERINAS_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")"/..)
+VERSION_FILE="${ASTERINAS_ROOT}/VERSION"
+BOOK_SRC_DIR="${ASTERINAS_ROOT}/book/src"
+
+TARGET_VERSION=$(tr -d '[:space:]' <"${VERSION_FILE}")
+echo "Target version for validation: ${TARGET_VERSION}"
+
+echo "Scanning directory: ${BOOK_SRC_DIR}"
+
+# Define the search pattern.
+# This regex looks for 'asterinas.github.io/api-docs/'
+# followed by any version number that IS NOT '${TARGET_VERSION}/'.
+# (?!${ESCAPED_TARGET_VERSION}/) is a negative lookahead.
+ESCAPED_TARGET_VERSION=${TARGET_VERSION//./\\.}
+PATTERN="asterinas\\.github\\.io/api-docs/(?!${ESCAPED_TARGET_VERSION}/)[0-9]+\\.[0-9]+\\.[0-9]+"
+MISMATCHES=$(grep -rPn "${PATTERN}" "${BOOK_SRC_DIR}" || true)
+
+if [ -n "${MISMATCHES}" ]; then
+  echo "----------------------------------------------------------------"
+  echo "ERROR: Found links with outdated or incorrect versions:"
+  echo "${MISMATCHES}"
+  echo "----------------------------------------------------------------"
+  echo "Please update the links above to match version ${TARGET_VERSION}."
+  exit 1
+fi
+
+echo "SUCCESS: All found links match version ${TARGET_VERSION}."
--- a/book/lychee.toml
+++ b/book/lychee.toml
@@ -1,0 +1,45 @@
+# This is the configuration for lychee to check invalid or broken links.
+# See https://lychee.cli.rs/guides/config/ for the meanings.
+
+# Verbose program output
+# Accepts log level: "error", "warn", "info", "debug", "trace"
+verbose = "error"
+
+# Output format of final status report.
+format = "detailed"
+
+# Don't show interactive progress bar while checking links.
+no_progress = true
+
+# Proceed for server connections considered insecure (invalid TLS).
+insecure = true
+
+# Exclude the paths from getting checked. The values are treated as regular expressions.
+exclude_path = [
+  "SUMMARY.md", # The valid empty links are flagged out by lychee.
+]
+
+# Request method
+method = "get"
+
+# Custom request headers
+header = { "accept" = "text/html" }
+
+user_agent = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"
+
+# Enable link caching. This can be helpful to avoid checking the same links on multiple runs.
+# .lycheecache will be generated.
+cache = true
+
+# Discard all cached requests older than this duration.
+max_cache_age = "90d"
+
+# Number of threads to utilize. Defaults to number of cores available to the system if omitted.
+threads = 1
+# Maximum number of concurrent link checks.
+max_concurrency = 3
+
+# More strictly limit the access rate to gnu websites.
+[hosts."www.gnu.org"]
+concurrency = 1
+request_interval = "2s"
--- a/book/lychee.toml
+++ b/book/lychee.toml
@@ -1,0 +1,57 @@
+# This is the configuration for lychee to check invalid or broken links.
+# See https://lychee.cli.rs/guides/config/ for the meanings.
+
+# Verbose program output
+# Accepts log level: "error", "warn", "info", "debug", "trace"
+verbose = "error"
+
+# Output format of final status report.
+format = "detailed"
+
+# Don't show interactive progress bar while checking links.
+no_progress = true
+
+# Proceed for server connections considered insecure (invalid TLS).
+insecure = true
+
+# Exclude the paths from getting checked. The values are treated as regular expressions.
+exclude_path = [
+  "SUMMARY.md", # The valid empty links are flagged out by lychee.
+]
+
+# Root path to use when checking absolute local links.
+root_dir = "book"
+# Enable the checking of fragments in links, both anchor fragments and text fragments.
+include_fragments = "full"
+# Remap URI via regex, mainly to replace local `page#anchor` with `page/index.html#anchor`.
+# The first remap keeps `page.html#anchor`, otherwise it would be made into the invalid `page.html/index.html#anchor`.
+remap = [
+  "(file:///.+)\\.html#(.+) $0",
+  "(file:///.+)#(.+) $1/index.html#$2",
+  "(file:///.+)/README.html $1/index.html",
+]
+
+# Request method
+method = "get"
+
+# Custom request headers
+header = { "accept" = "text/html" }
+
+user_agent = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"
+
+# Enable link caching. This can be helpful to avoid checking the same links on multiple runs.
+# .lycheecache will be generated.
+cache = true
+
+# Discard all cached requests older than this duration.
+max_cache_age = "90d"
+
+# Number of threads to utilize. Defaults to number of cores available to the system if omitted.
+threads = 1
+# Maximum number of concurrent link checks.
+max_concurrency = 3
+
+# More strictly limit the access rate to gnu websites.
+[hosts."www.gnu.org"]
+concurrency = 1
+request_interval = "2s"
--- a/book/src/kernel/the-framekernel-architecture.md
+++ b/book/src/kernel/the-framekernel-architecture.md
@@ -70,8 +70,7 @@ if doing it outside is possible.
 * **Efficiency.**
 The safe API provided by the framework is only allowed
 to introduce minimal overheads.
-Ideally, these APIs should be realized
-as [zero-cost abstractions](https://monomorph.is/posts/zero-cost-abstractions/).
+Ideally, these APIs should be realized as zero-cost abstractions.
 
 Fortunately, our efforts
 to design and implement an OS framework meeting these standards
--- a/book/src/to-contribute/version-bump.md
+++ b/book/src/to-contribute/version-bump.md
@@ -40,19 +40,20 @@ except during the version bump process.
 
 ## How to Bump Versions
 
-We recommend a three-commit procedure to bump versions:
+We recommend a four-commit procedure to bump versions:
 1. **Commit 1 "Bump the Docker image version"** triggers the generation of a new Docker image.
 2. **Commit 2 "Switch to a new Docker image"** makes the codebase use the new Docker image.
 3. **Commit 3 "Bump the project version"** triggers the release of new crates.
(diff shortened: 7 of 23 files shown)


tatetian:
Thanks for fixing these links to API docs. But I hope you could devise a long-term solution. It is unsustainable to update these links manually after every version bump.

On project version bump, we want to make sure that all links to API docs (1) use the up-to-date version number and (2) the referenced language items still exist. One complication is that you can't simply fetch the API link upon the version bump because at that time, the new API doc is not generated, yet.

zjp-CN:
> (2) the referenced language items still exist

Before submitting the PR, I [tried](https://github.com/os-checker/asterinas/actions/runs/26015774985/job/76465401415) using [mdbook-linkcheck2](https://github.com/marxin/mdbook-linkcheck2) to find dead links, but the tool is not great:
* some links produce false positive request errors; these errors depend on local network conditions or request settings — websites that can be accessed successfully through a browser may still fail when checked by the tool
* the tool misses real broken links: the diagnostics fail to report the link which is fixed in the PR and ought to have been flagged

<details><summary>some of the mdbook-linkcheck2 warnings</summary>
<p>

```
 INFO Book building has started
Warning: The mdbook-mermaid preprocessor was built against version 0.5.0 of mdbook, but we're being called from version 0.5.2
 INFO Running the html backend
 INFO HTML book written to `/home/runner/work/asterinas/asterinas/book/book/html`
 INFO Running the linkcheck2 backend
 INFO Invoking the "linkcheck2" renderer
[2026-05-18T05:44:49Z ERROR rustls_platform_verifier::verification::others] failed to verify TLS certificate: invalid peer certificate: certificate not valid for name "www.info-zip.org"; certificate is only valid for DnsName("alicecoopertourprograms.com"), DnsName("dangerousdwarf.com"), DnsName("glennchadbourne.com"), DnsName("highbaugh.goatley.com"), DnsName("hunter.goatley.com"), DnsName("kong.goatley.com"), DnsName("matthewcorbettsworld.com"), DnsName("pota.goatley.com"), DnsName("rickhautala.com"), DnsName("robertmccammon.com"), DnsName("ronaldkelly.com"), DnsName("sickthings.goatley.com"), DnsName("webbwilder.goatley.com"), DnsName("www.alicecoopertourprograms.com"), DnsName("www.dangerousdwarf.com"), DnsName("www.glennchadbourne.com"), DnsName("www.goatley.com"), DnsName("www.lounge.org"), DnsName("www.matthewcorbettsworld.com"), DnsName("www.rickhautala.com"), DnsName("www.robertmccammon.com") or DnsName("www.ronaldkelly.com")
[2026-05-18T05:45:57Z ERROR rustls_platform_verifier::verification::others] failed to verify TLS certificate: invalid peer certificate: certificate expired: verification time 1779083157 (UNIX), but certificate is not valid after 1779070776 (12381 seconds ago)
error: error sending request for url (https://www.gnu.org/software/bash/)
  ┌─ distro/popular-applications/system-core/index.md:9:1
  │
9 │ [Bash](https://www.gnu.org/software/bash/) is the GNU Project's shell and command language, enabled by default in Asterinas NixOS.
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error sending request for url (https://www.gnu.org/software/bash/)

error: error sending request for url (https://www.gnu.org/software/coreutils/)
    ┌─ distro/popular-applications/system-core/index.md:188:1
    │
188 │ [coreutils](https://www.gnu.org/software/coreutils/) includes basic file, shell and text manipulation utilities.
    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error sending request for url (https://www.gnu.org/software/coreutils/)

...

warning: Potential incomplete link
   ┌─ osdk/reference/commands/build.md:24:1
   │
24 │ [default: dev]
   │ ^^^^^^^^^^^^^^ Did you forget to define a URL for `default: dev`?
   │
   = hint: declare the link's URL. For example: `[default: dev]: http://example.com/`
```

</p>
</details> 

> you can't simply fetch the API link upon the version bump because at that time, the new API doc is not generated, yet

The webhook event [repository_dispatch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#repository_dispatch)[^1] seems like a solution to notify doc deployment and trigger CI: 
* for the api-docs repo, emit a `docs_deployed` event after the API docs are successfully deployed to GitHub Pages
* for the asterinas repo, set up a workflow that starts the link check when receiving the `docs_deployed` event

[^1]: I haven't run any experiments; this idea comes from Gemini ([conversation link](https://aistudio.google.com/app/prompts/1SaH9c-pNcadPSYq2ufN8p8iSnzUT7j0u)).


I'll put in a bit more effort to see if it pans out.

zjp-CN:
I'll use [lychee-action](https://github.com/lycheeverse/lychee-action) to check links. All link connections are summarized in the action pane like [this](https://github.com/asterinas/asterinas/actions/runs/26148314452?pr=3237).

By default, it finds the following asterinas dead links, and dozens of dead links from other projects.

We can write a [config toml file](https://lychee.cli.rs/guides/config/) to filter out links that we're not interested.

### Errors in src/ostd/soundness/safe-kernel-logic.md

* [404] <https://asterinas.github.io/api-docs/0.17.1/ostd/sync/struct.LocalIrqDisabled.html> | Rejected status code: 404 Not Found (configurable with "accept" option)
* [404] <https://asterinas.github.io/api-docs/0.17.1/ostd/sync/struct.PreemptDisabled.html> | Rejected status code: 404 Not Found (configurable with "accept" option)

### Errors in src/ostd/soundness/safe-policy-injection.md

* [404] <https://asterinas.github.io/api-docs/0.17.1/ostd/log/logger/fn.inject_logger.html> | Rejected status code: 404 Not Found (configurable with "accept" option)

### Errors in src/to-contribute/coding-guidelines/rust-guidelines/select-topics/logging.md

* [404] <https://asterinas.github.io/ostd/ostd/log/> | Rejected status code: 404 Not Found (configurable with "accept" option)


<details><summary>Full summary of lychee link check</summary>
<p>

# Summary

| Status         | Count |
|----------------|-------|
| 🔍 Total       | 870   |
| ✅ Successful  | 816   |
| ⏳ Timeouts    | 4     |
| 🔀 Redirected  | 15    |
| 👻 Excluded    | 22    |
| ❓ Unknown     | 0     |
| 🚫 Errors      | 13    |
| ⛔ Unsupported | 0     |

## Errors per input

### Errors in src/distro/popular-applications/file-management-and-terminal/README.md

* [ERROR] <http://mama.indstate.edu/users/ice/tree/> | Network error: Connection failed. Check network connectivity and firewall settings (error sending request for url (http://mama.indstate.edu/users/ice/tree/)): Connection failed. Check network connectivity and firewall settings
* [TIMEOUT] <https://darwinsys.com/file/> | Timeout
* [ERROR] <https://www.info-zip.org/> | Network error: SSL certificate error. Check certificate validity (error sending request for url (https://www.info-zip.org/)): SSL certificate error. Check certificate validity

### Errors in src/kernel/intel-tdx.md

* [404] <https://github.com/canonical/tdx/tree/noble-24.04?tab=readme-ov-file#4-setup-host-os> | Rejected status code: 404 Not Found (configurable with "accept" option)
* [403] <https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html> | Rejected status code: 403 Forbidden (configurable with "accept" option)

### Errors in src/kernel/roadmap.md

* [403] <https://dl.acm.org/doi/10.1145/3274694.3274720> | Rejected status code: 403 Forbidden (configurable with "accept" option)
* [403] <https://dl.acm.org/doi/10.1145/3373376.3378507> | Rejected status code: 403 Forbidden (configurable with "accept" option)

### Errors in src/kernel/the-framekernel-architecture.md

* [ERROR] <https://monomorph.is/posts/zero-cost-abstractions/> | Network error: Connection failed. Check network connectivity and firewall settings (error sending request for url (https://monomorph.is/posts/zero-cost-abstractions/)): Connection failed. Check network connectivity and firewall settings

### Errors in src/osdk/guide/intel-tdx.md

* [404] <https://github.com/canonical/tdx/tree/noble-24.04?tab=readme-ov-file#4-setup-host-os> | Rejected status code: 404 Not Found (configurable with "accept" option)
* [403] <https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html> | Rejected status code: 403 Forbidden (configurable with "accept" option)

### Errors in src/ostd/soundness/safe-kernel-logic.md

* [404] <https://asterinas.github.io/api-docs/0.17.1/ostd/sync/struct.LocalIrqDisabled.html> | Rejected status code: 404 Not Found (configurable with "accept" option)
* [404] <https://asterinas.github.io/api-docs/0.17.1/ostd/sync/struct.PreemptDisabled.html> | Rejected status code: 404 Not Found (configurable with "accept" option)

### Errors in src/ostd/soundness/safe-policy-injection.md

* [404] <https://asterinas.github.io/api-docs/0.17.1/ostd/log/logger/fn.inject_logger.html> | Rejected status code: 404 Not Found (configurable with "accept" option)

### Errors in src/rfcs/0002-asterinas-nixos.md

* [TIMEOUT] <https://nixos.org/download/#nix-install-linux> | Timeout
* [TIMEOUT] <https://nixos.org/download/#nixos-iso> | Timeout
* [TIMEOUT] <https://nixos.org/guides/nix-pills/14-override-design-pattern.html> | Timeout

### Errors in src/to-contribute/coding-guidelines/rust-guidelines/select-topics/logging.md

* [404] <https://asterinas.github.io/ostd/ostd/log/> | Rejected status code: 404 Not Found (configurable with "accept" option)

## Redirects per input

### Redirects in src/distro/popular-applications/databases-and-middleware/README.md

* [200] <https://influxdata.com/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://influxdata.com/ --[301]--> https://www.influxdata.com/

### Redirects in src/distro/popular-applications/development-tools/README.md

* [200] <https://golang.org/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://golang.org/ --[301]--> https://go.dev/
* [200] <https://jqlang.github.io/jq/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://jqlang.github.io/jq/ --[301]--> https://jqlang.org/
* [200] <https://nodejs.org/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://nodejs.org/ --[307]--> https://nodejs.org/en
* [200] <https://openjdk.java.net/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://openjdk.java.net/ --[301]--> https://openjdk.org/

### Redirects in src/distro/popular-applications/file-management-and-terminal/README.md

* [200] <https://p7zip.sourceforge.io/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://p7zip.sourceforge.io/ --[301]--> https://p7zip.sourceforge.net/
* [200] <https://www.sourceware.org/bzip2> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://www.sourceware.org/bzip2 --[301]--> https://www.sourceware.org/bzip2/

### Redirects in src/distro/popular-applications/multimedia/README.md

* [200] <http://www.imagemagick.org/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: http://www.imagemagick.org/ --[301]--> https://imagemagick.org/
* [200] <https://sox.sourceforge.net/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://sox.sourceforge.net/ --[302]--> https://sourceforge.net/projects/sox/

### Redirects in src/distro/popular-applications/networking/README.md

* [200] <https://packages.qa.debian.org/w/whois.html> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://packages.qa.debian.org/w/whois.html --[301]--> https://tracker.debian.org/pkg/whois
* [200] <https://www.nlnetlabs.nl/projects/ldns/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://www.nlnetlabs.nl/projects/ldns/ --[301]--> https://www.nlnetlabs.nl/projects/ldns/about/

### Redirects in src/distro/popular-applications/web-servers-and-proxies/README.md

* [200] <https://openresty.org/> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://openresty.org/ --[302]--> https://openresty.org/en/

### Redirects in src/osdk/guide/README.md

* [200] <https://www.rust-lang.org/tools/install> | Redirect: Followed 2 redirects resolving to the final status of: OK. Redirects: https://www.rust-lang.org/tools/install --[301]--> https://rust-lang.org/tools/install --[301]--> https://rust-lang.org/tools/install/

### Redirects in src/ostd/README.md

* [200] <https://docs.rs/ostd/latest/ostd> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://docs.rs/ostd/latest/ostd --[302]--> https://docs.rs/ostd/latest/ostd/

### Redirects in src/to-contribute/coding-guidelines/rust-guidelines/select-topics/logging.md

* [200] <https://docs.rs/log> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://docs.rs/log --[302]--> https://docs.rs/log/latest/log/


</p>
</details>

.github/workflows/publish_api_docs.yml
```diff
@@ -49,3 +49,27 @@ jobs:
           KEY_FILE=./api_docs_publish_key
           echo "$API_DOCS_PUBLISH_KEY\n" > ${KEY_FILE}
           ./tools/github_workflows/publish_api_docs.sh release ${KEY_FILE} 
+
+  notify:
+    needs: check_api_docs
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
zjp-CN: The contributor may lack write permission to perform repository-dispatch, so the job must [fail](https://github.com/asterinas/asterinas/actions/runs/26160031118/job/76950191314).

`${{ secrets.API_DOCS_NIGHTLY_PUBLISH_KEY }}` seems a github token which has been used to deploy nightly api-docs, we can use it to gain the write permission. (Edit: I've set github token to  API_DOCS_NIGHTLY_PUBLISH_KEY.)

```
Run peter-evans/repository-dispatch@v2
Error: Resource not accessible by integration
```


> One complication is that you can't simply fetch the API link upon the version bump because at that time, the new API doc is not generated, yet.

But I tend to remove the notify mechanism as implemented here, and let the contributor just get a workflow error when check_dead_links or check_latest_version fails, because the cost to wait for the api-docs finish deployment is small.

.github/workflows/publish_api_docs.yml
```diff
@@ -49,3 +49,27 @@ jobs:
...
+    steps:
+        - name: Check the URL availability
+          env:
+            URL: https://asterinas.github.io/api-docs
```
zjp-CN: The URL should contain latest version. I'll fix it if this notify job is acceptable.

.github/workflows/publish_website.yml
tatetian: Rather adding two jobs to this file, I prefer encapsulating this complexity by adding a `Makefile` for the book sub-project. This way, the workflow file can be simplified to 

```yml
      - name: Build the website
        run: |
          cd asterinas/book
          make check
          make
```

The `check` target hides all the complexity of checking hyperlinks and API versions. The default Makefile target hides the complexity of installing `mdbook-mermaid` if it is not installed already.

In addition to managing complexity, the extra benefit of this approach is that the check can be done locally (need to install [`lychee`](https://github.com/lycheeverse/lychee) in the Docker image). This allows both human or a local coding agent to detect bugs before CI.

tools/bump_version.sh
```diff
@@ -186,6 +197,7 @@ sync_project_version() {
     fi
 
     update_project_dependencies
+    update_all_book_links
```
tatetian: I am not happy with the current approach because it **addresses a lessor concern on the prices of increased complexity with a limitation**. It is a lessor concern (outdated version in the book) as it only occurs on project version bumps, which is an infrequent event.

This PR introduces the following new complexity:
1. A new `check-latest-version.sh` script;
2. A new dependency between two otherwise independent Github workflows;
3. Enhancing the `bump_version.sh` script.

And the current approach of this PR comes with the limitation that the version bump PR is guaranteed to fail. This is because `bump_version.sh --version_file` updates the api doc hyperlinks _before_ the API docs are published. The second complexity eventually fixes the failed CI, but still the project version bump PR is doomed to fail temporarily.

Therefore, I propose adopting a new approach that addresses this concern with minimal changes.

The first change is similar to Point 3, but what's difference is that `update_all_book_links` is only invoked upon a new argument, `--api_doc_hyperlinks_in_book`

```bash
# Usage: ./tools/bump_version.sh command [options]
# Commands:
#   --docker_version_file [major|minor|patch|date]   Bump the Docker image version in the DOCKER_IMAGE_VERSION file under the project root
#   --docker_version_refs                            Update all references to the Docker image version throughout the codebase
#   --version_file                                   Bump the project version to match the Docker image version
#   --api_doc_hyperlinks_in_book                     Update all API doc hyperlinks to the latest project version
```

This is the only code change, which is quite minimal.

The second change is the version bump process. 

```md
We recommend a three-commit procedure to bump versions:

* Commit 1 “Bump the Docker image version” triggers the generation of a new Docker image.
* Commit 2 “Switch to a new Docker image” makes the codebase use the new Docker image.
* Commit 3 “Bump the project version” triggers the release of new crates.
```

Currently, the project version bump process involves three commits. Since we need to wait for the new API doc generation due to Commit 3, a four commit is introduced.

```
* Commit 4 “Update the API doc hyperlinks in the book” triggers the release of a new book, where all hyperlinks to the OSTD API documentation updated to the latest version.
```

Obviously, Commit 4 is generated mechanically with `./bump_verion.sh --api_doc_hyperlinks_in_book`. The "Version Bump" page of the Book needs to be revised with this new workflow.

The only downside of this alternative approach is demanding an extra commit upon project version bump. But this is a good deal for us because as I pointed out, we only bump project version _occasionally_. So this is just an infrequent inconvenience.

tatetian (commented):
Thanks for improving the book! I raised two major concerns. Both are about getting things done as simple as possible.

.github/workflows/publish_api_docs.yml
zjp-CN: The job notification is removed. 

Update the api docs in separate commit after the links are available. See [this](https://github.com/asterinas/asterinas/pull/3237#discussion_r3428597131) comment.

tools/bump_version.sh
zjp-CN: Good idea. I've improved the PR as you suggested.

zjp-CN:
@boterinas rerun

tatetian:
The commit messages are a complete mess.

The first commit's message is:

```
Add check_latest_version.sh to check if api doc links point to current version
The output if the check fails:

/home/zjp/asterinas/book/src/ostd/soundness/safe-kernel-logic.md:118:[`CpuLocal::get_on_cpu(target_cpu)`](https://asterinas.github.io/api-docs/0.17.1/ostd/cp
u/local/struct.CpuLocal.html#method.get_on_cpu) allows remote access without a guard
----------------------------------------------------------------
Please update the links above to match version 0.18.0.
```

The second one is:

```
Add check_dead_links.md used by lychee

The command and output of lychee for book:
asterinas $ lychee --config tools/book/check_dead_links.toml book/src

📝 Summary
---------------------
🔍 Total..........816
🔗 Unique.........725
✅ Successful.....775
⏳ Timeouts.........0
🔀 Redirected......18
👻 Excluded........31
❓ Unknown..........0
🚫 Errors..........10
⛔ Unsupported.....10

Errors in book/src/distro/popular-applications/development-tools/README.md
   [504] https://www.php.net/ (at 193:1) | Rejected status code: 504 Gateway Timeout

Redirects in book/src/distro/popular-applications/development-tools/README.md
https://golang.org/ --[301]--> https://go.dev/
https://jqlang.github.io/jq/ --[301]--> https://jqlang.org/
https://nodejs.org/ --[307]--> https://nodejs.org/en
https://openjdk.java.net/ --[301]--> https://openjdk.org/

Errors in book/src/distro/popular-applications/file-management-and-terminal/README.md
   [502] http://mama.indstate.edu/users/ice/tree/ (at 460:1) | Rejected status code: 502 Bad Gateway

Redirects in book/src/distro/popular-applications/file-management-and-terminal/README.md
https://p7zip.sourceforge.io/ --[301]--> https://p7zip.sourceforge.net/
https://www.info-zip.org/ --[301]--> https://infozip.sourceforge.net/
https://www.sourceware.org/bzip2 --[301]--> https://www.sourceware.org/bzip2/

Errors in book/src/kernel/intel-tdx.md
   [404] https://github.com/canonical/tdx/tree/noble-24.04?tab=readme-ov-file#4-setup-host-os (at 40:5) | Rejected status code: 404 Not Found

Errors in book/src/kernel/README.md
 [ERROR] error: (at 21:45) | Empty URL found but a URL must not be empty

Errors in book/src/kernel/the-framekernel-architecture.md
 [ERROR] https://monomorph.is/posts/zero-cost-abstractions/ (at 74:4) | Connection closed unexpectedly. Server terminated early

Errors in book/src/osdk/guide/intel-tdx.md
   [404] https://github.com/canonical/tdx/tree/noble-24.04?tab=readme-ov-file#4-setup-host-os (at 10:5) | Rejected status code: 404 Not Found

Errors in book/src/ostd/soundness/safe-kernel-logic.md
   [404] https://asterinas.github.io/api-docs/0.17.1/ostd/sync/struct.LocalIrqDisabled.html (at 72:18) | Rejected status code: 404 Not Found
   [404] https://asterinas.github.io/api-docs/0.17.1/ostd/sync/struct.PreemptDisabled.html (at 71:18) | Rejected status code: 404 Not Found

Errors in book/src/ostd/soundness/safe-policy-injection.md
   [404] https://asterinas.github.io/api-docs/0.17.1/ostd/log/logger/fn.inject_logger.html (at 127:5) | Rejected status code: 404 Not Found

Errors in book/src/to-contribute/coding-guidelines/rust-guidelines/select-topics/logging.md
   [404] https://asterinas.github.io/ostd/ostd/log/ (at 9:17) | Rejected status code: 404 Not Found

Redirects in book/src/to-contribute/coding-guidelines/rust-guidelines/select-topics/logging.md
https://docs.rs/log --[302]--> https://docs.rs/log/latest/log/
```

I marked this PR as "Draft". After doing a careful self-review, including the commit messages, feel free to mark this PR as "Ready for Review".

---

Edit: My first impression was that the raw output included in the commit messages were included by mistake. But after taking a second look, they seems to be included on purpose. So I am sorry to say it was "a mess".

But still, it is not a good idea to pack "log"-style info into commit messages, which are supposed to be concise. And they are error messages or command output; preserving them has little value in the long run.

zjp-CN:
They are included in the commits to show what link checks emit.

I can remove them from the commit message.

tatetian:
> They are included in the commits to show what link checks emit.
> 
> I can remove them from the commit message.

Yes, they should be removed... See my edit above.

.github/workflows/publish_website.yml
```diff
@@ -27,8 +29,10 @@ jobs:
       - name: Build the website
         run: |
           cd asterinas/book
-          mdbook-mermaid install
-          mdbook build
+          # Check dead hyperlinks and outdated api doc links.
+          make check
+          # Render the book to HTML artifacts.
```
tatetian: Remove worthless comments like the ones above. Using one sentence to explain one obvious command does not improve readability; it is wasting people's time to read these trivial comments.

book/src/to-contribute/version-bump.md
```diff
@@ -40,19 +40,20 @@ except during the version bump process.
...
+4. **Commit 4 "Update API doc hyperlinks in the Book"** updates all Book links to use the latest API doc version.
 
 Depending on your exact purpose,
-you may complete the version bump process with at most three commits within two PRs.
+you may complete the version bump process with at most four commits within two PRs.
 * **To make non-breaking changes to the Docker images**,
 submit Commit 1 in a PR, then Commit 2 in another.
 * **To make breaking changes to the Docker images and the crates' APIs**,
-submit Commit 1 in a PR, then Commit 2 and 3 in another.
+submit Commit 1 in a PR, then Commit 2, 3, and 4 in another.
```
tatetian: Commit 2 and 3 should go to the second PR, whereas Commit 4 goes to the third. This is because you need the second PR to trigger API doc generation.

> you may complete the version bump process with at most four commits within two PRs.

Change "within two PRs" to "two or three PRs".

book/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+all: build
+
+build: ensure_mdbook_mermaid
+	mdbook-mermaid install
+	mdbook build
+
+ensure_mdbook_mermaid:
+	@if ! command -v mdbook-mermaid >/dev/null 2>&1; then \
+		cargo install mdbook-mermaid; \
```
tatetian: This step can be skipped because the whole Asterinas project assumes that the development environment has been set up properly, including tools like mdbook and mdbook-mermaid. There is no point in worrying that this single tool is not pre-installed.

book/Makefile
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: MPL-2.0
+
+.PHONY: all check check_dead_links check_latest_version build ensure_mdbook_mermaid
+
+all: build
+
+build: ensure_mdbook_mermaid
+	mdbook-mermaid install
```
tatetian: The root `Makefile` contains these lines:

```
.PHONY: book
book: book/mermaid.min.js book/mermaid-init.js
        @cd book && mdbook build
 
book/mermaid.min.js book/mermaid-init.js:
        @mdbook-mermaid install book/
```

Now that the `book` project has its own Makefile, these lines should be moved to this Makefile.

Note that these original lines are strictly better than running `mdbook-mermaid install` unconditionally. You should remove `mdbook-mermaid install`.

tools/book/check_dead_links.toml
```diff
@@ -0,0 +1,30 @@
...
+# Output format
+format = "detailed"
+
+# Don't show interactive progress bar while checking links.
+no_progress = true
+
+# Proceed for server connections considered insecure (invalid TLS).
+insecure = true
+
+# Exclude URLs and mail addresses from checking. The values are treated as regular expressions.
```
tatetian: Exclude URLs and mail addresses from checking --> Exclude _these_ URLs and mail addresses from checking.

book/lychee.toml
```diff
@@ -0,0 +1,30 @@
+# See https://lychee.cli.rs/guides/config/ for the meanings.
+
+# Verbose program output
+# Accepts log level: "error", "warn", "info", "debug", "trace"
+verbose = "info"
+
+# Output format
+format = "detailed"
```
tatetian, line 9: What is the rationale to prefer the most verbose output?

```
verbose = "info"

# Output format
format = "detailed"
```

book/lychee.toml
```diff
@@ -0,0 +1,30 @@
...
+insecure = true
+
+# Exclude URLs and mail addresses from checking. The values are treated as regular expressions.
+exclude = [
+  "https://www.gnu.org/software/.*",                                                               # 429 Too Many Requests
+  "https://dl.acm.org/doi/.*",                                                                     # 403 Forbidden
+  "https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html", # 403 Forbidden
+  "https://web.archive.org/web/20230407233736/https://monomorph.is/posts/zero-cost-abstractions/", # timeout
+  "https://ftp.gnu.org/.*",                                                                        # timeout
+  "https://nixos.org/.*",                                                                          # timeout
```
tatetian: I am confused. If fetching these URLS ended up with timeout, how comes these URLs be valid?

book/lychee.toml
```diff
@@ -0,0 +1,30 @@
...
+
+# Don't show interactive progress bar while checking links.
+no_progress = true
+
+# Proceed for server connections considered insecure (invalid TLS).
+insecure = true
+
+# Exclude URLs and mail addresses from checking. The values are treated as regular expressions.
+exclude = [
+  "https://www.gnu.org/software/.*",                                                               # 429 Too Many Requests
```
tatetian: "Too many" is not a good reason to skip them. In the future, any new URLs starting with `https://www.gnu.org/software/` will not be checked for validity.

Can we check only the URLs in Git diff somehow? Checking the same URLs that are ready checked before is mostly fruitless (unless the web page pointed by the URLs are gone).

book/lychee.toml
zjp-CN: Timeouts and other various problems happen on different machines (CI runner vs URL server vs your computer vs other maintainer's computer) or environments (CLI vs browser).

A valid url is accessible, but not always accessible, or  accesible in any circumstance. 

So timeout can be the status on CI, but the link is accesible through my browser.

book/lychee.toml
zjp-CN: "Too Many Requests" is the server's response with status code 429. I copy & paste it to here.

```
Errors in src/distro/popular-applications/system-core/README.md
   [429] https://www.gnu.org/software/coreutils/ (at 237:1) | Rejected status code: 429 Too Many Requests
   [429] https://www.gnu.org/software/diffutils/ (at 353:1) | Rejected status code: 429 Too Many Requests
   [429] https://www.gnu.org/software/findutils/ (at 375:1) | Rejected status code: 429 Too Many Requests
   [429] https://www.gnu.org/software/grep/ (at 405:1) | Rejected status code: 429 Too Many Requests
   [429] https://www.gnu.org/software/texinfo/ (at 524:1) | Rejected status code: 429 Too Many Requests

Redirects in src/distro/popular-applications/system-core/README.md
https://www.gnu.org/software/which/ --[302]--> https://carlowood.github.io/which/


Errors in src/kernel/linux-compatibility/README.md
   [429] https://www.gnu.org/software/grub/manual/multiboot/multiboot.html (at 394:3) | Rejected status code: 429 Too Many Requests
   [429] https://www.gnu.org/software/grub/manual/multiboot2/multiboot.html (at 395:3) | Rejected status code: 429 Too Many Requests
```


> unless the web page pointed by the URLs are gone

That has indeed happened.



> Can we check only the URLs in Git diff somehow?

You'll miss the dead links like [this](https://github.com/asterinas/asterinas/commit/3d27562591bcaf0362d3397c7b080a2b3b3e8323).

book/lychee.toml
zjp-CN: [`--format`](https://lychee.cli.rs/guides/cli/#-f---format) is the output format of final status report.

```
# compact (default)
🔍 816 Total (in 2m 11s 707ms) 🔗 724 Unique ✅ 790 OK 🚫 10 Errors 👻 16 Excluded 🔀 18 Redirects
```

vs

```
# detailed
📝 Summary
---------------------
🔍 Total..........816
🔗 Unique.........724
✅ Successful.....787
⏳ Timeouts.........0
🔀 Redirected......18
👻 Excluded........16
❓ Unknown..........0
🚫 Errors..........13
⛔ Unsupported.....13
```

`detailed` format is more readable.

tatetian: > [`--format`](https://lychee.cli.rs/guides/cli/#-f---format) is the output format of final status report.

Ok. Both look fine with me. The latter one is arguably more readable.

What about `verbose = "info"`? Why do people run `make check` want to see a verbose output?

book/src/to-contribute/version-bump.md
```diff
@@ -104,3 +105,15 @@ Pack these changes into a third commit and
...
+
+After new API docs are published, run:
+
+```
+./bump_version.sh --api_doc_hyperlinks_in_book
+```
+
+This command mechanically updates API doc hyperlinks in Book pages
+to point to the current project version.
+Submit the result as a separate commit.
```
tatetian: > Submit the result as a separate commit.

The heading of this section is "Commit 4". So I suppose that it is obvious that this is a commit. Rewrite the sentence as below:

> Submit the changes in a single-commit PR.

book/lychee.toml
zjp-CN: These settings are stolen from lychees' documentation.

I made experiments to see how verbose value differs:


<details><summary>[verbos = "error"] summary + failure status</summary>
<p>

```
📝 Summary
---------------------
🔍 Total..........816
🔗 Unique.........724
✅ Successful.....779
⏳ Timeouts.........0
🔀 Redirected......17
👻 Excluded........32
❓ Unknown..........0
🚫 Errors...........5
⛔ Unsupported......5

Errors in src/distro/popular-applications/cicd-and-devops/README.md
 [ERROR] https://taskfile.dev/ (at 29:1) | Connection closed unexpectedly. Server terminated early

Errors in src/distro/popular-applications/development-tools/README.md
 [ERROR] https://www.nano-editor.org/ (at 531:1) | Connection closed unexpectedly. Server terminated early

Errors in src/distro/popular-applications/file-management-and-terminal/README.md
 [ERROR] https://www.openwall.com/john/ (at 562:1) | Connection closed unexpectedly. Server terminated early

Errors in src/distro/popular-applications/system-core/README.md
 [ERROR] https://www.greenwoodsoftware.com/less/ (at 469:1) | Connection closed unexpectedly. Server terminated early

Errors in src/README.md
 [ERROR] https://www.tldrlegal.com/license/mozilla-public-license-2-0-mpl-2 (at 69:19) | Connection closed unexpectedly. Server terminated early

Hint: Followed 17 redirects. You might want to consider replacing redirecting URLs with the resolved URLs. Use verbose mode (`-v`/`-vv`) to see redirection details.
make: *** [Makefile:8: check_dead_links] Error 2
```

</p>
</details> 

<details><summary>[verbose = "warn"] summary + failure status + exlusion</summary>
<p>

```
[EXCLUDED] https://nixos.org/ (at 10:165) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/ (at 44:282) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/guides/nix-pills/14-override-design-pattern.html (at 64:148) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/download/#nixos-iso (at 66:138) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/manual/nixos/stable/#sec-configuration-file (at 94:87) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/download/#nix-install-linux (at 113:64) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/ (at 4:11) | This is due to your 'exclude' values
[EXCLUDED] https://nixos.org/manual/nixos/stable/#ch-configuration (at 76:21) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/wget/ (at 165:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/gzip/ (at 39:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/tar/ (at 98:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/screen/ (at 188:1) | This is due to your 'exclude' values
[EXCLUDED] https://darwinsys.com/file/ (at 219:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/gawk/ (at 268:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/sed/ (at 308:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/make/ (at 365:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/emacs/ (at 485:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/bash/ (at 9:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/coreutils/ (at 237:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/diffutils/ (at 353:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/findutils/ (at 375:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/grep/ (at 405:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/texinfo/ (at 524:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/which/ (at 599:1) | This is due to your 'exclude' values
 [ERROR] https://www.tldrlegal.com/license/mozilla-public-license-2-0-mpl-2 (at 69:19) | Connection closed unexpectedly. Server terminated early
[EXCLUDED] https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html (at 3:41) | This is due to your 'exclude' values
[EXCLUDED] https://web.archive.org/web/20230407233736/https://monomorph.is/posts/zero-cost-abstractions/ (at 74:4) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/grub/manual/multiboot/multiboot.html (at 394:3) | This is due to your 'exclude' values
[EXCLUDED] https://www.gnu.org/software/grub/manual/multiboot2/multiboot.html (at 395:3) | This is due to your 'exclude' values
[EXCLUDED] https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_28.html (at 98:1) | This is due to your 'exclude' values
[EXCLUDED] https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html (at 144:17) | This is due to your 'exclude' values
[EXCLUDED] https://dl.acm.org/doi/10.1145/3373376.3378507 (at 22:4) | This is due to your 'exclude' values
[EXCLUDED] https://dl.acm.org/doi/10.1145/3274694.3274720 (at 50:10) | This is due to your 'exclude' values
📝 Summary
---------------------
🔍 Total..........816
🔗 Unique.........724
✅ Successful.....783
⏳ Timeouts.........0
🔀 Redirected......17
👻 Excluded........32
❓ Unknown..........0
🚫 Errors...........1
⛔ Unsupported......1

Errors in src/README.md
 [ERROR] https://www.tldrlegal.com/license/mozilla-public-license-2-0-mpl-2 (at 69:19) | Connection closed unexpectedly. Server terminated early
```

</p>
</details> 

<details><summary>[verbose = "info"] summary + failure status + exlusion + success status with redirection info</summary>
<p>

```
   [200] https://openresty.org/ (at 60:1) | 200 OK | Followed 1 redirect. Redirects: https://openresty.org/ --[302]--> https://openresty.org/en/
   [200] https://openjdk.java.net/ (at 153:1) | 200 OK | Followed 1 redirect. Redirects: https://openjdk.java.net/ --[301]--> https://openjdk.org/
   [200] https://nodejs.org/ (at 107:1) | 200 OK | Followed 1 redirect. Redirects: https://nodejs.org/ --[307]--> https://nodejs.org/en
   [200] https://www.nlnetlabs.nl/projects/ldns/ (at 187:1) | 200 OK | Followed 1 redirect. Redirects: https://www.nlnetlabs.nl/projects/ldns/ --[301]--> https://www.nlnetlabs.nl/projects/ldns/about/
...
```

</p>
</details> 

For CI, `verbose = "error"` is preferable for concision.

zjp-CN: Verbose is set to error now.

book/src/kernel/the-framekernel-architecture.md
```diff
@@ -71,7 +71,7 @@ if doing it outside is possible.
 The safe API provided by the framework is only allowed
 to introduce minimal overheads.
 Ideally, these APIs should be realized
-as [zero-cost abstractions](https://monomorph.is/posts/zero-cost-abstractions/).
+as [zero-cost abstractions](https://web.archive.org/web/20230407233736/https://monomorph.is/posts/zero-cost-abstractions/).
```
tatetian: I think we can simply remove this link. This link is not the authoritative source for this concept anyway.

book/Makefile
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: MPL-2.0
+
```
tatetian, line 2: Each Makefile should have a default target, which is usually about building the project.

The book-related targets in the root Makefile:

```make
.PHONY: book
book: book/mermaid.min.js book/mermaid-init.js
	@cd book && mdbook build

book/mermaid.min.js book/mermaid-init.js:
	@mdbook-mermaid install book/
```

should be migrated into this Makefile. This way, this local Makefile would have a sensible default Makefile target. In addition, it is better to keep all book-related targets in its own Makefile.

book/lychee.toml
```diff
@@ -0,0 +1,31 @@
+# See https://lychee.cli.rs/guides/config/ for the meanings.
```
tatetian, line 2: It is better to keep book-related config files and scripts like `check_dead_links.toml` and `check_latest_version.sh` under the `<ASTER_PROJECT>/book/` directory, not under `<ASTER_PROJECT>/tools/book`. This way, the book project is self-contained.

---

Rename this file to `lychee.toml`. The name of an executable script typically starts with a verb; but the name of a configuration file should be in the noun form.

book/check_rust_api_doc_version.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
tatetian, line 1: This script does not check the latest version, but the API doc links to see if its version is up-to-date. Saying "check the latest version" is inaccurate (in a subtle way); saying either "check the version" or "ensure the version up-to-date" is correct in logic. I suggest renaming this file to `check_rust_api_doc_version.sh`.

tatetian (commented):
Thanks to your revision, this PR looks much better. I only raised a few minor issues; most about naming and styles.

The commit history needs to be reworked as it contains too many small, fragmented commits. There are 8 commits for a small PR of 193 additions and 82 deletions. This means each commit only includes about 50 lines of changes.

The `atomic-commit` guideline says that "Each commit should represent one logical change". Adding `check_latest_version.sh` and `check_dead_links.toml` are not two logical changes because they have no users! Seeing each file alone does not make much sense. For example, `check_dead_links.toml` (to be renamed to `lychee.toml`) has no value at all if we do not add the `book/Makefile` file that invokes `lychee`. So these two changes must be included in the same commit. So is `check_dead_links.toml`: its purpose and value can only be justified with `book/Makefile`.

Use this standard to reconsider the organization of commits. We pursue logically-atomic commits, not the smallest ones.

book/lychee.toml
```diff
@@ -0,0 +1,33 @@
+# This is the configuration for lychee to check dead links.
```
tatetian: "dead links" --> "invalid or broken links".

Dead links sounds like the link is valid, but the website is brought down, so the link became dead. But a more common situation is that the link itself is incorrect (e.g., due to a typo) in the first place. So "invalid or broken links" sounds more accurate.

book/lychee.toml
tatetian: > The links work in browsers but not via the CLI, which will likely cause CI failures.

This argument is a weak argument. From the perspective of a Web server, all it sees are HTTP requests. It cannot differentiate between HTTP requests sent from a CLI tool or a Web server. Advanced methods such as reCAPTCHA must be put in place to deny visits from bots.

> "Too Many Requests" 

Maybe the tool sends too many requests too fast to the GNU website?

> "https://nixos.org/.*"

Again, I have a reservation about skipping all URLs to a Website.

> "https://www.tldrlegal.com/license/mozilla-public-license-2-0-mpl-2 # connection failed

This comment simply does not make sense. Does "connection failed" mean that the link is broken?

Overall, the `exclude` list is not well justified.

---

> The links work in browsers but not via the CLI, which will likely cause CI failures.

As I understand it, the primary reason for excluding these URLs is because visiting them could fail temporarily. But since we have succeeded in visiting them in one setting or another (from a web server), we know these URLs are valid.

I think fundamentally, checking the same URLs again and again is quite wasteful. This makes me think an alternative approach: maintaining a list of validated URLs and only revisiting those URLs after a period of time (like three months). Maybe `lychee` support such a mode?

zjp-CN:
> alternative approach: maintaining a list of validated URLs and only revisiting those URLs after a period of time (like three months). Maybe lychee support such a mode?


Thanks for the feedback. I agree we should avoid broad exclusions and keep checks strict.

I reran the problematic links with conservative settings for gun websites:
- concurrency = 1
- threads = 1
- request_interval = 2s

Even then, responses are still non-deterministic (the same machine can get different outcomes for the same URL under different rates/time windows).  In [this](https://github.com/asterinas/asterinas/actions/runs/28000965513/job/82873043417) CI failure, the status is timeout instead of "too many requests". So I won’t attach a single fixed failure reason per URL.

For periodic re-validation: lychee does not provide a built-in validated-URL cache with TTL.
My current plan is:
1. Keep a minimal explicit exclude list in config for CI stability.
2. Apply that config in `make check`.
3. Use that list as a record of manually verified valid URLs (reachable in browser).

This keeps CI stable, avoids broad domain-level skips, and makes each exception explicit.

tatetian:
> Keep a minimal explicit exclude list in config for CI stability

This does not change the fact that the URLs in the exclude list will never be revalidated. They could go broken because the Website is shut down. Maintaining such an exception list also brings maintenance burden.

> For periodic re-validation: lychee does not provide a built-in validated-URL cache with TTL.

I talked with AI. It turns out that Lychee has built-in support for caching. Here is the full response from AI:

## So yes — I agree, and your hypothesized "cache mode" of Lychee exists

You asked "Maybe lychee supports such a mode?" — **it does, almost exactly as you described.** lychee has first-class caching:

- `--cache` (config: `cache = true`) writes a `.lycheecache` file: plaintext, one line per URL with its **status code + a UNIX timestamp**.
- `--max-cache-age` (config: `max_cache_age = "..."`, default `1d`) — only re-checks URLs whose cached entry is older than this. Set it to `"90d"` and you get exactly your "validate, then don't revisit for three months" behavior.
- `cache_exclude_status = [429, 500, 502, 503, 504]` — crucially, this stops lychee from caching *transient* failures, so a one-off rate-limit/timeout isn't frozen into the cache as a permanent "bad."

So your direction was correct on the merits and is directly buildable. The only wrinkle is persistence: `.lycheecache` must survive between CI runs via `actions/cache` (restore before, `save` with `if: always()` after), since each run is a fresh container.

## But caching alone isn't the cleanest fix — split the check

Caching cures the *waste*. It doesn't cure the *root cause of the flakiness*: **a third party's server being down should never block an Asterinas PR.** The deeper fix is to separate two kinds of links that have completely different reliability profiles:

1. **Internal links** (relative paths between book pages, `SUMMARY.md` entries, `#anchor` fragments). These are **deterministic and offline** — they break only when *you* move/rename a file, which is the single most common and most actionable doc bug. lychee checks these with `--offline` (no network at all).

2. **External `http(s)` links.** Flaky, slow, and not the author's fault when they die.

That suggests this architecture:

|            | PR check (blocking)                         | Scheduled check (non-blocking)                                                                     |
| ---------- | ------------------------------------------- | -------------------------------------------------------------------------------------------------- |
| What       | `lychee --offline --include-fragments src/` | full lychee, network on                                                                            |
| When       | every `book/**` PR                          | weekly cron (`schedule:`)                                                                          |
| Network    | none → never flaky                          | yes, but cached                                                                                    |
| On failure | fail the PR (it's the author's bug)         | **open/update a GitHub issue**, don't block anyone                                                 |
| Caching    | n/a                                         | `cache=true`, `max_cache_age="14d"`, persisted via `actions/cache`, `cache_exclude_status=[429,…]` |

lychee-action even has a documented recipe for the issue-on-failure pattern (pipe its output to `peter-evans/create-issue-from-file`), so a dead external link becomes a tracked issue someone fixes on their own time, instead of red CI on an unrelated PR.

With this split, the `lychee_links_excluded.toml` denylist shrinks to just the handful of hosts that are *permanently* bot-hostile from CI (e.g. intel.com, dl.acm.org paywalls) — not the dozen-plus "usually works but rate-limits" hosts, which the cache now handles correctly.

Sources:
- [lychee caching docs](https://lychee.cli.rs/recipes/caching/)
- [lychee-action README (cache + create-issue recipe)](https://github.com/lycheeverse/lychee-action/blob/master/README.md)
- [lychee.example.toml (cache, max_cache_age, cache_exclude_status keys)](https://github.com/lycheeverse/lychee/blob/master/lychee.example.toml)
- [lychee rate-limits troubleshooting](https://lychee.cli.rs/troubleshooting/rate-limits/)

zjp-CN:
@boterinas rerun

zjp-CN:
Thanks for the detailed feedback @tatetian . I addressed your concern in commit https://github.com/asterinas/asterinas/pull/3237/commits/03ab28892468c32766feb0a4f4d0a61b0a1d295a (the first commit in this PR).

Key updates:

1) No URL exclude list now
    - I removed the URL/host exclude-list approach for external links.

2) Re-check failed links with curl
    - `book/check_invalid_or_broken_links.sh` now runs lychee first, then retries lychee-failed HTTP(S) URLs with plain `curl`.
    - Recovered links are treated as valid. With this retry flow, the link check now passes completely.

3) Enable and persist lychee cache in CI
    - `book/lychee.toml` now enables cache with:
      - `cache = true`
      - `max_cache_age = "90d"`
    - In `.github/workflows/publish_website.yml`, `.lycheecache` is restored/saved via `actions/cache`.
    - On each CI run, the latest available `.lycheecache` is restored into `book/` and reused by lychee; cached entries are used according to timestamps and remain valid within the 90-day window.

So the current behavior is: no external-link exclude list, retry transient failures, and use persisted 90-day cache to reduce repeated checks/flakiness.


```bash
# The first time:
asterinas/book $ make check_invalid_or_broken_links 
[1/3] Running lychee...
Hint: Followed 1 redirect. You might want to consider replacing redirecting URLs with the resolved URLs. Use verbose mode (`-v`/`-vv`) to see redirection det
ails.
Hint: You can configure accepted/rejected response codes with `-a` or `--accept`
lychee failed links: 5
[2/3] Curl fallback on lychee-failed links...
RECOVERED: http://www.info-zip.org/ (status=301)
RECOVERED: https://dl.acm.org/doi/10.1145/3274694.3274720 (status=403)
RECOVERED: https://dl.acm.org/doi/10.1145/3373376.3378507 (status=403)
RECOVERED: https://p7zip.sourceforge.io/ (status=301)
RECOVERED: https://sox.sourceforge.net/ (status=302)
[3/3] Summary: lychee_failed=5 recovered=5 still_failed=0
[3/3] Cache updated lines: 5
lychee summary:
  total.......815
  unique......723
  successful..810
  timeouts....0
  redirected..1
  excluded....0
  unknown.....0
  errors......5
  unsupported.0

# The second time:
asterinas/book $ make check_invalid_or_broken_links
[1/3] Running lychee...
lychee failed links: 0
lychee summary:
  total.......815
  unique......723
  successful..815
  timeouts....0
  redirected..0
  excluded....0
  unknown.....0
  errors......0
  unsupported.0
```

Items in  .lycheecache are incrementally added (comments are my own):

```
# 1782268705 is a unix timestamp for 2026-06-24 10:38:25 (UTC+8)
https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html,200,1782268705
# 1782287133 == 2026-06-24 15:45:33 (UTC+8)
https://wipe.sourceforge.net/,200,1782287133
```

Github workflow cache for .lycheecache are used:

image

zjp-CN:
One more point on why we keep `curl` as a recheck step:

Some websites enforce rate limits, and some perform request fingerprinting (i.e., they derive identity from multiple request-header fields, not only User-Agent). In practice, this means a link checker like lychee may be rejected even with a browser-like UA, while the same URL can still be reachable with `curl` in CI.

Also, the recheck does not require HTTP 200 only. Status codes like 301/302 (redirect) or even some 403 responses through curl can still indicate that the URL endpoint exists and is reachable, depending on site policy. If we required strict 200-only semantics, we would inevitably need and maintain an explicit allow/deny exception list again, which is exactly what we are trying to avoid.

For now, we treat a successful transport-level recheck as evidence that the link is not clearly broken. If the team prefers stricter semantics (e.g., 2xx/3xx only), we can enforce that, but then an explicit exception list becomes necessary for CI stability.

martinholovsky:
I tried the offline half of the split suggested above against current `main`, using the lychee already pinned in the image (0.24.2), in case measurements help.

```
lychee --offline --include-fragments 'book/src/**/*.md'

950 total / 817 unique   314 OK   209 errors   427 excluded   424 ms
```

The speed and determinism look right for a blocking check — no network, sub-second. But the error count is almost entirely an mdBook convention mismatch rather than broken links:

- **203** are `Cannot find fragment`, across 27 unique targets. All 27 are directories containing a `README.md`. mdBook resolves `some-dir#anchor` to `some-dir/README.md#anchor`; lychee resolves the directory itself and so cannot find the fragment. For example `.../ai-and-machine-learning#codex` is reported broken, though `### Codex` is at line 67 of that directory's `README.md`.
- **5** are `Empty URL` in `SUMMARY.md`, which is mdBook's draft-chapter syntax (`[Title]()`) and legal.
- **1** is a genuine broken link — `book/src/kernel/README.md:21`, `[OSDK]()` — which this PR already fixes.

So 208 of the 209 are false positives against the source markdown, and the single real finding is already handled here.

One thought: the workflow already runs `mdbook build` (`publish_website.yml:31`), and in the generated HTML mdBook has resolved directory links to `index.html` and rendered draft chapters as non-links. Would running the offline check over `book/book/` rather than `book/src/` avoid the convention problem without needing any exclude list? lychee also has `--remap`, `--base` and `--fallback-extensions` if checking the source side is preferred.

I have not tested the built-HTML variant — only flagging it, since the build already happens in that workflow.

zjp-CN:
> * **203** are `Cannot find fragment`

Checking anchor fragments is pretty reasonable and practical.

> Would running the offline check over `book/book/` rather than `book/src/` avoid the convention problem without needing any exclude list?

Yes, it's a good idea to check html files directly.

> lychee also has `--remap`, `--base` and `--fallback-extensions` if checking the source side is preferred.

Thanks for the suggestion. I've confirmed the following command works:


```
cd book
mdbook build
lychee --offline --include-fragments --remap '(file:///.+)\.html#(.+) $0'  --remap '(file:///.+)#(.+) $1/index.html#$2' 'book/**/*.html' --root-dir book
```

These configuration values are written to `lychee.toml` now, and I also adjusted the `check_invalid_or_broken_links.sh` script to report failed local files.

book/check_rust_api_doc_version.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && cd .. && pwd)
+VERSION_FILE="${SCRIPT_DIR}/VERSION"
+BOOK_SRC_DIR="${SCRIPT_DIR}/book/src"
```
tatetian: First of all, `SCRIPT_DIR` actually refers to the project directory. So calling it `SCRIPT_DIR` is misleading. It should be renamed to `ASTERINAS_ROOT` or `ASTER_SRC_DIR`. This is the convention chosen by existing scripts in `tools/`.

Second, the pattern `$(SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && cd .. && pwd))` is kind of silly and indirect. Just write `PROJECT_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")"`,

book/check_invalid_or_broken_links.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# Temporary files:
+# - JSON_FILE stores lychee JSON output.
+# - URLS_FILE stores failed http/https links extracted from lychee output.
+# - CACHE_UPDATES_FILE stores recovered links to write back into .lycheecache.
+# - CACHE_MERGED_FILE is the merged cache output before replace.
+JSON_FILE="$(mktemp /tmp/lychee-result-XXXXXX.json)"
+URLS_FILE="$(mktemp /tmp/lychee-failed-urls-XXXXXX.txt)"
+CACHE_UPDATES_FILE="$(mktemp /tmp/lychee-cache-updates-XXXXXX.txt)"
+CACHE_MERGED_FILE="$(mktemp /tmp/lychee-cache-merged-XXXXXX.txt)"
```
tatetian: Instead of creating four temporary files, a more elegant approach would be creating one temporary directory, which contains the four files. This way, you can enter one directory to examine all intermediate files in case you want to debug the script. The `cleanup` function would also be simplified because it only needs to remove one directory, rather than four files.

book/check_invalid_or_broken_links.sh
```diff
@@ -0,0 +1,123 @@
...
+  echo "  excluded....$LYCHEE_EXCLUDED"
+  echo "  unknown.....$LYCHEE_UNKNOWN"
+  echo "  errors......$LYCHEE_ERRORS"
+  echo "  unsupported.$LYCHEE_UNSUPPORTED"
+}
+
+# Run lychee once and capture JSON output for fallback processing.
+echo "[1/3] Running lychee..."
+LYCHEE_EXIT=0
+if ! lychee --config lychee.toml --format json --no-progress book/**/*.html >"$JSON_FILE"; then
```
tatetian, line 38: It looks that you are using lychee to check the rendered HTML files, not the source Markdown files. This means the `check_invalid_or_broken_links` Makefile target depends on the `build` target. You should establish this dependency in the Makefile.

book/src/to-contribute/version-bump.md
```diff
@@ -104,3 +105,15 @@ Pack these changes into a third commit and
...
+### Commit 4: "Update API doc hyperlinks in the Book"
+
+After new API docs are published, run:
+
+```
+./bump_version.sh --api_doc_hyperlinks_in_book
+```
+
+This command mechanically updates API doc hyperlinks in Book pages
+to point to the current project version.
```
tatetian, line 118: You probably want to provide the validation command:

````md
This command mechanically updates API doc hyperlinks in Book pages
to point to the current project version.

Then, validate all hyperlinks in the Book are valid:

```
cd ../book && make check
```

Submit the changes in a single-commit PR.
````

tatetian (commented):
Great job! Our use of lychee has been made much more robust. Like it.

tatetian (approved):

PR #3241, opened 2026-05-18
Add `ptrace(PTRACE_SYSCALL)`

vvvvsv:
Required for supporting `strace`.

>        PTRACE_SYSCALL
>        PTRACE_SINGLESTEP
>               Restart the stopped tracee as for PTRACE_CONT, but arrange
>               for the tracee to be stopped at the next entry to or exit
>               from a system call, or after execution of a single
>               instruction, respectively.  (The tracee will also, as
>               usual, be stopped upon receipt of a signal.)  From the
>               tracer's perspective, the tracee will appear to have been
>               stopped by receipt of a SIGTRAP.  So, for PTRACE_SYSCALL,
>               for example, the idea is to inspect the arguments to the
>               system call at the first stop, then do another
>               PTRACE_SYSCALL and inspect the return value of the system
>               call at the second stop.  The data argument is treated as
>               for PTRACE_CONT.  (addr is ignored.)
https://man7.org/linux/man-pages/man2/ptrace.2.html


kernel/src/thread/task.rs
```diff
@@ -87,7 +89,13 @@ pub fn create_new_user_task(
                 ReturnReason::UserSyscall => {
                     ctx.posix_thread
                         .set_orig_syscall_ret(user_ctx.syscall_ret());
-                    handle_syscall(&ctx, user_ctx);
+                    if !matches!(
+                        ctx.posix_thread.ptrace_may_stop_on_syscall(&ctx, user_ctx),
```
tatetian: This line (a blocking operation) is important enough to have its own line.

```rust
let res = ctx.posix_thread.ptrace_may_stop_on_syscall(&ctx, user_ctx);
if !matches!(res, PtraceStopResult::Interrupted) {

	ctx.posix_thread.ptrace_may_stop_on_syscall(&ctx, user_ctx);
}
```

tatetian (commented):
Looks most great. Just one nit. Thanks!

tatetian (approved):

PR #3242, opened 2026-05-18
Add `ptrace(PTRACE_PEEK/POKE_TEXT/DATA)`

vvvvsv:
Required for supporting `strace`.

>        PTRACE_PEEKTEXT
>        PTRACE_PEEKDATA
>               Read a word at the address addr in the tracee's memory,
>               returning the word as the result of the ptrace() call.
>               Linux does not have separate text and data address spaces,
>               so these two operations are currently equivalent.  (data is
>               ignored; but see NOTES.)
> 
>        PTRACE_POKETEXT
>        PTRACE_POKEDATA
>               Copy the word data to the address addr in the tracee's
>               memory.  As for PTRACE_PEEKTEXT and PTRACE_PEEKDATA, these
>               two operations are currently equivalent.

>    C library/kernel differences
>        At the system call level, the PTRACE_PEEKTEXT, PTRACE_PEEKDATA,
>        and PTRACE_PEEKUSER operations have a different API: they store
>        the result at the address specified by the data parameter, and the
>        return value is the error flag.  The glibc wrapper function
>        provides the API given in DESCRIPTION above, with the result being
>        returned via the function return value.
> 
https://man7.org/linux/man-pages/man2/ptrace.2.html

### DIFF
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/process-and-thread-management/README.md
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/process-and-thread-management/README.md
@@ -91,7 +91,11 @@ Supported functionality in SCML:
 
 Supported requests:
 * `PTRACE_TRACEME`
+* `PTRACE_PEEKTEXT`
+* `PTRACE_PEEKDATA`
 * `PTRACE_PEEKUSER` (x86-64 only)
+* `PTRACE_POKETEXT`
+* `PTRACE_POKEDATA`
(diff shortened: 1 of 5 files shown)


vvvvsv:
These two issues (https://github.com/asterinas/asterinas/pull/3242#discussion_r3260101597 and https://github.com/asterinas/asterinas/pull/3242#discussion_r3260101690) do exist, but they are pre-existing problems and are unrelated to this PR. I will revise `vmar_impls::access_alien` later.

tatetian (approved):

PR #3243, opened 2026-05-18
Add unused KCODE32 slot in x86 GDT to align USER_CS and USER_SS with Linux

vvvvsv:
Running GDB on current main branch's x86-64 builds produces the following warning:

```
warning: Selected architecture i386:x86-64 is not compatible with reported target architecture i386
warning: Architecture rejected target-supplied description
```

This is because [GDB expects the value of CS on x86-64 Linux to be `0x33`](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/nat/x86-linux.c#l124). On the current main branch, it is `0x2b`. Adding an empty entry before `UCODE64` and `UDATA` makes the values of CS and SS match Linux: `cs = 0x33, ss = 0x2b`.

Alternative: add a patch to GDB to suppress this warning. This would be less elegant.

Required for #2323.

### DIFF
--- a/kernel/src/arch/x86/ptrace.rs
+++ b/kernel/src/arch/x86/ptrace.rs
@@ -348,3 +348,14 @@ fn write_word(bytes: &mut [u8], offset: usize, value: usize) {
 const fn word_range(offset: usize) -> Range<usize> {
     offset..offset + size_of::<usize>()
 }
+
+// Reference: <https://elixir.bootlin.com/linux/v6.16.5/source/arch/x86/include/asm/segment.h#L188-L189>
+const LINUX_USER_CS: usize = 0x33;
+const LINUX_USER_SS: usize = 0x2b;
+
+// Some applications (e.g., GDB), rely on `ptrace` exposing Linux's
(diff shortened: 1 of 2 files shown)


lrh2000:
Add some assertions in `aster-kernel`? e.g., in `kernel/src/arch/x86/ptrace.rs`? This will prevent it from breaking silently in the future.
```rust
// Reference: <...>.
const LINUX_USER_CS;
const LINUX_USER_SS;

// Explain here: Some applications (e.g., GDB) expect / rely ...
// See <...>.
ostd::const_assert!(LINUX_USER_CS == USER_CS);
ostd::const_assert!(LINUX_USER_SS == USER_SS);
```

Also note that our kernel CS/DS values do not match those of Linux. But I suppose no one really cares about that anyway.

vvvvsv:
> Also note that our kernel CS/DS values do not match those of Linux. But I suppose no one really cares about that anyway.

Yes, I think so 😀

lrh2000 (approved):

PR #3244, opened 2026-05-19
Extend `InodeIo` with more file opeartions on per open data

li041:
Fixes #3071.

The current implementation in Asterinas delegates specific file operations (like seek/readdir) down to the Inode layer (via InodeHandle).

The Problem: 
- Many FUSE operations require a File Handle (fh), which is stored in the `file->private_data` in Linux. 

Maybe we should delegates those file opeartions to `FileIo`(which is renamed to `OpenFileOps` in this PR) for special files.

See https://github.com/asterinas/asterinas/issues/3071 for details

### DIFF
--- a/kernel/src/device/registry/block.rs
+++ b/kernel/src/device/registry/block.rs
@@ -9,8 +9,8 @@ use crate::{
     device::{Device, DeviceType, DevtmpfsInodeMeta, add_node},
     events::IoEvents,
     fs::{
-        file::{FileIo, StatusFlags},
-        vfs::{inode::InodeIo, path::PathResolver},
+        file::{PerOpenFileOps, StatusFlags},
+        vfs::{inode::FileOps, path::PathResolver},
     },
     prelude::*,
     process::signal::{PollHandle, Pollable},
@@ -81,19 +81,19 @@ impl Device for BlockFile {
         Some(DevtmpfsInodeMeta::new(self.0.name()))
     }
 
-    fn open(&self) -> Result<Box<dyn FileIo>> {
+    fn open(&self) -> Result<Box<dyn PerOpenFileOps>> {
         Ok(Box::new(OpenBlockFile(self.0.clone())))
     }
 }
 
 /// Represents an opened block device file ready for I/O operations.
 //
-// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Box` just to implement the `FileIo`
-// trait. It leads to redundant vtable dispatch and heap allocation. We should devise a better
-// strategy to eliminate the unnecessary intermediate `Box`.
+// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Box` just to implement the
+// `PerOpenFileOps` trait. It leads to redundant vtable dispatch and heap allocation. We should
+// devise a better strategy to eliminate the unnecessary intermediate `Box`.
 struct OpenBlockFile(Arc<dyn BlockDevice>);
 
-impl InodeIo for OpenBlockFile {
+impl FileOps for OpenBlockFile {
     fn read_at(
         &self,
         offset: usize,
@@ -126,7 +126,7 @@ impl Pollable for OpenBlockFile {
     }
 }
 
-impl FileIo for OpenBlockFile {
+impl PerOpenFileOps for OpenBlockFile {
     fn check_seekable(&self) -> Result<()> {
         Ok(())
     }
--- a/kernel/src/fs/file/inode_handle.rs
+++ b/kernel/src/fs/file/inode_handle.rs
@@ -16,7 +16,7 @@ use crate::{
         pipe::PipeHandle,
         utils::DirentVisitor,
         vfs::{
-            inode::{FallocMode, InodeIo},
+            inode::{FallocMode, FileOps},
             inode_ext::InodeExt,
             path::Path,
             range_lock::{FileRange, OFFSET_MAX, RangeLockItem, RangeLockType},
@@ -29,10 +29,10 @@ use crate::{
 
 pub struct InodeHandle {
     path: Path,
-    /// `file_io` is similar to the `file_private` field in Linux's `file` structure. If `file_io`
-    /// is `Some(_)`, typical file operations including `read`, `write`, `poll`, and `ioctl` will
-    /// be provided by `file_io`, instead of `path`.
-    file_io: Option<Box<dyn FileIo>>,
+    /// `open_file` is similar to the `file_private` field in Linux's `file` structure. If
+    /// `open_file` is `Some(_)`, typical file operations including `read`, `write`, `poll`,
+    /// and `ioctl` will be provided by the per-open file object instead of `path`.
+    open_file: Option<Box<dyn PerOpenFileOps>>,
     offset: Mutex<usize>,
     status_flags: AtomicStatusFlags,
     rights: Rights,
@@ -57,19 +57,19 @@ impl InodeHandle {
         status_flags: StatusFlags,
     ) -> Result<Self> {
         let inode = path.inode();
-        let (file_io, rights) = if status_flags.contains(StatusFlags::O_PATH) {
+        let (open_file, rights) = if status_flags.contains(StatusFlags::O_PATH) {
             (None, Rights::empty())
         } else if inode.type_() == InodeType::Dir && access_mode.is_writable() {
             return_errno_with_message!(Errno::EISDIR, "a directory cannot be opened writable");
         } else {
-            let file_io = inode.open(access_mode, status_flags).transpose()?;
+            let open_file = inode.open(access_mode, status_flags).transpose()?;
             let rights = Rights::from(access_mode);
-            (file_io, rights)
+            (open_file, rights)
         };
 
         Ok(Self {
             path,
-            file_io,
+            open_file,
             offset: Mutex::new(0),
             status_flags: AtomicStatusFlags::new(status_flags),
             rights,
@@ -89,10 +89,10 @@ impl InodeHandle {
         self.rights
     }
 
-    fn inode_io_and_is_offset_aware(&self) -> (&dyn InodeIo, bool) {
-        if let Some(ref file_io) = self.file_io {
-            let is_offset_aware = file_io.is_offset_aware();
-            return (file_io.as_ref(), is_offset_aware);
+    fn file_ops_and_is_offset_aware(&self) -> (&dyn FileOps, bool) {
+        if let Some(ref open_file) = self.open_file {
+            let is_offset_aware = open_file.is_offset_aware();
+            return (open_file.as_ref(), is_offset_aware);
         }
 
         let inode = self.path.inode();
@@ -100,12 +100,12 @@ impl InodeHandle {
         (inode.as_ref(), is_offset_aware)
     }
 
-    /// Returns the `InodeIo` for positional I/O, rejecting files
+    /// Returns the `FileOps` for positional I/O, rejecting files
     /// that do not support `pread`/`pwrite`.
-    fn inode_io_for_positional_io(&self) -> Result<&dyn InodeIo> {
-        if let Some(ref file_io) = self.file_io {
-            file_io.check_positional_io()?;
-            return Ok(file_io.as_ref());
+    fn file_ops_for_positional_io(&self) -> Result<&dyn FileOps> {
+        if let Some(ref open_file) = self.open_file {
+            open_file.check_positional_io()?;
+            return Ok(open_file.as_ref());
         }
 
         let inode = self.path.inode();
@@ -221,23 +221,23 @@ impl InodeHandle {
         Ok(())
     }
 
-    pub fn downcast_file_io<T: 'static>(&self) -> Result<Option<&T>> {
+    pub fn downcast_open_file<T: 'static>(&self) -> Result<Option<&T>> {
         if self.rights.is_empty() {
             return_errno_with_message!(Errno::EBADF, "the file is opened as a path");
         }
 
-        let Some(file_io) = self.file_io.as_ref() else {
+        let Some(open_file) = self.open_file.as_ref() else {
             return Ok(None);
         };
 
-        Ok((file_io.as_ref() as &dyn Any).downcast_ref::<T>())
+        Ok((open_file.as_ref() as &dyn Any).downcast_ref::<T>())
     }
 }
 
 impl Pollable for InodeHandle {
     fn poll(&self, mask: IoEvents, poller: Option<&mut PollHandle>) -> IoEvents {
-        if let Some(ref file_io) = self.file_io {
-            return file_io.poll(mask, poller);
+        if let Some(ref open_file) = self.open_file {
+            return open_file.poll(mask, poller);
         }
 
         if self.rights.is_empty() {
@@ -255,16 +255,16 @@ impl FileLike for InodeHandle {
             return_errno_with_message!(Errno::EBADF, "the file is not opened readable");
         }
 
-        let (inode_io, is_offset_aware) = self.inode_io_and_is_offset_aware();
+        let (file_ops, is_offset_aware) = self.file_ops_and_is_offset_aware();
         let status_flags = self.status_flags();
 
         if !is_offset_aware {
-            return inode_io.read_at(0, writer, status_flags);
+            return file_ops.read_at(0, writer, status_flags);
         }
 
         let mut offset = self.offset.lock();
 
-        let len = inode_io.read_at(*offset, writer, status_flags)?;
+        let len = file_ops.read_at(*offset, writer, status_flags)?;
         *offset += len;
 
         Ok(len)
@@ -275,56 +275,56 @@ impl FileLike for InodeHandle {
             return_errno_with_message!(Errno::EBADF, "the file is not opened writable");
         }
 
-        let (inode_io, is_offset_aware) = self.inode_io_and_is_offset_aware();
+        let (file_ops, is_offset_aware) = self.file_ops_and_is_offset_aware();
         let status_flags = self.status_flags();
 
         if !is_offset_aware {
-            return inode_io.write_at(0, reader, status_flags);
+            return file_ops.write_at(0, reader, status_flags);
         }
 
         let mut offset = self.offset.lock();
 
-        // FIXME: How can we deal with the `O_APPEND` flag if `file_io` is set?
-        if status_flags.contains(StatusFlags::O_APPEND) && self.file_io.is_none() {
+        // FIXME: How can we deal with the `O_APPEND` flag if `open_file` is set?
+        if status_flags.contains(StatusFlags::O_APPEND) && self.open_file.is_none() {
             // FIXME: `O_APPEND` should ensure that new content is appended even if another process
             // is writing to the file concurrently.
             *offset = self.path.size();
         }
 
-        let len = inode_io.write_at(*offset, reader, status_flags)?;
+        let len = file_ops.write_at(*offset, reader, status_flags)?;
         *offset += len;
 
         Ok(len)
     }
 
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
-        let inode_io = self.inode_io_for_positional_io()?;
+        let file_ops = self.file_ops_for_positional_io()?;
         if !self.rights.contains(Rights::READ) {
             return_errno_with_message!(Errno::EBADF, "the file is not opened readable");
         }
 
         let status_flags = self.status_flags();
 
-        inode_io.read_at(offset, writer, status_flags)
+        file_ops.read_at(offset, writer, status_flags)
     }
 
     fn write_at(&self, mut offset: usize, reader: &mut VmReader) -> Result<usize> {
-        let inode_io = self.inode_io_for_positional_io()?;
+        let file_ops = self.file_ops_for_positional_io()?;
         if !self.rights.contains(Rights::WRITE) {
             return_errno_with_message!(Errno::EBADF, "the file is not opened writable");
         }
 
         let status_flags = self.status_flags();
 
-        // FIXME: How can we deal with the `O_APPEND` flag if `file_io` is set?
-        if status_flags.contains(StatusFlags::O_APPEND) && self.file_io.is_none() {
+        // FIXME: How can we deal with the `O_APPEND` flag if `open_file` is set?
+        if status_flags.contains(StatusFlags::O_APPEND) && self.open_file.is_none() {
             // If the file has the `O_APPEND` flag, the offset is ignored.
             // FIXME: `O_APPEND` should ensure that new content is appended even if another process
             // is writing to the file concurrently.
             offset = self.path.size();
         }
 
-        inode_io.write_at(offset, reader, status_flags)
+        file_ops.write_at(offset, reader, status_flags)
     }
 
     fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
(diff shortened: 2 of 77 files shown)


kernel/src/fs/file/inode_handle.rs
```diff
@@ -521,17 +550,17 @@ pub enum SeekFrom {
     Current(isize),
 }
 
-/// A trait for file-like objects that provide custom I/O operations.
+/// A trait for file-like objects that provide custom file operations
```
StevenJiang1110:
```suggestion
/// A trait for file-like objects that provide custom file operations.
```

Don't missing the final period.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -542,11 +571,23 @@ pub trait FileIo: Pollable + InodeIo + Any + Send + Sync + 'static {
...
+    /// [`check_seekable`]: OpenFileOps::check_seekable
     fn check_positional_io(&self) -> Result<()> {
         self.check_seekable()
     }
 
+    /// Seeks the file position.
+    ///
+    /// `end` is the inode-provided end position used for [`SeekFrom::End`].
+    fn seek(&self, offset: &Mutex<usize>, pos: SeekFrom, end: Option<usize>) -> Result<usize> {
+        do_seek_util(offset, pos, end)
```
StevenJiang1110: It feels a bit odd to use `do_seek_util` as the default implementation of `seek`. The fact that `do_seek_util` does not even take `self` as a receiver suggests that it is not inherently tied to `OpenFileOps` — it simply operates on fields inside `InodeHandle`.

To be honest, without seeing an actual caller, it is hard to say whether the current form of `seek` is the most appropriate design.

Ideally, I would prefer `do_seek_util` to live inside `InodeHandle`, since it exclusively operates on the `offset` field within `InodeHandle`, and `OpenFileOps::seek` should have a default no-op implementation. However, I am not sure whether that would be sufficient for virtiofs. If it works, feel free to refactor it accordingly; if not, let's remove the `seek` method from `OpenFileOps` in this PR and add it back in the virtiofs PR, where we can revisit and discuss the appropriate shape of `seek` in that context.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -542,11 +571,23 @@ pub trait FileIo: Pollable + InodeIo + Any + Send + Sync + 'static {
...
+    }
+
+    /// Reads directory entries and advances the offset.
+    fn readdir(&self, _offset: &Mutex<usize>, _visitor: &mut dyn DirentVisitor) -> Result<usize> {
```
StevenJiang1110: Could we adopt the same signature as `Inode::readdir_at`?

```rust
fn readdir_at(&self, offset: usize, visitor: &mut dyn DirentVisitor) -> Result<usize> {
```

Or at the very least, have this method accept `_offset: &mut usize` or `_offset: usize`, so as to avoid exposing the internal lock.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -521,17 +550,17 @@ pub enum SeekFrom {
...
 
-/// A trait for file-like objects that provide custom I/O operations.
+/// A trait for file-like objects that provide custom file operations
 ///
-/// This trait is typically implemented for special files like devices or
-/// named pipes (FIFOs), which have behaviors different from regular on-disk files.
-pub trait FileIo: Pollable + InodeIo + Any + Send + Sync + 'static {
+/// The trait may carry per-open backend, such as a device handle, pipe endpoint,
+/// FUSE file handle. Inode-backed files use the inode implementation directly instead.
+pub trait OpenFileOps: Pollable + InodeIo + Any + Send + Sync + 'static {
```
StevenJiang1110: One thing I find myself going back and forth on is what this trait should be named. In Linux, `file_private` serves a wide variety of purposes — for instance, in FUSE or NFS it holds a remote file handle, while in certain filesystems like ext4 it may cache directory metadata. This makes it difficult to establish a single, general-purpose abstraction.

In our current implementation, it looks like this trait is primarily used to short-circuit certain file operations in order to support special file types, such as devices, while regular files would not need to override it at all. With that in mind, I would suggest defining the trait as follows:

```rust
/// Defines operations for special files that need per-open state.
///
/// Regular files are served directly by their inode. Special files such as
/// devices, pipes, namespace files, and procfs files may need a backend object
/// created at `open()` time to hold file-description-specific state or to
/// override operations whose behavior is not purely inode-backed.
pub trait SpecialFileOps {}
```

kernel/src/fs/file/inode_handle.rs
```diff
@@ -29,15 +29,37 @@ use crate::{
...
-    file_io: Option<Box<dyn FileIo>>,
+    /// Backs operations attached to this opened file description.
+    open_file_backend: OpenFileBackend,
     offset: Mutex<usize>,
     status_flags: AtomicStatusFlags,
     rights: Rights,
 }
 
+/// A backend for opened-file operations.
+enum OpenFileBackend {
```
StevenJiang1110: Simplifying it to just `Backend` should be sufficient. Alternatively, you could name it `InodeHandleBackend`, given that `InodeHandle` carries the semantics of an opened file. Either way, please make sure the naming is kept consistent throughout the same file.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -29,15 +29,37 @@ use crate::{
...
+    InodeBacked,
+    /// The filesystem-specific opened-file implementation.
+    ///
+    /// This is analogous to Linux's `file->private_data`: the object can carry
+    /// the backend created at open time.
+    Custom(Box<dyn OpenFileOps>),
+}
+
+impl OpenFileBackend {
+    fn custom_ops(&self) -> Option<&dyn OpenFileOps> {
```
StevenJiang1110: Naming it `as_custom_ops` or `as_special_file`. Prefix it with `as` so we can know it's not a simple getter method.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -221,23 +251,23 @@ impl InodeHandle {
         Ok(())
     }
 
-    pub fn downcast_file_io<T: 'static>(&self) -> Result<Option<&T>> {
+    pub fn downcast_open_file_ops<T: 'static>(&self) -> Result<Option<&T>> {
```
StevenJiang1110: Name it as `downcast_special_file`. We don't need to mention the `ops` suffix, which is more clear.

StevenJiang1110 (commented):
Thanks for your contribution!

kernel/src/fs/file/inode_handle.rs
li041: The main motivation for having seek delegated to the file-private implementation is virtio-fs specific behavior around `SEEK_END`.

For virtio-fs, the cached inode size can become stale because the server-side file size may change independently. Before handling `SeekFrom::End`, we need to revalidate the inode attributes through the file handle (fh) to obtain the latest size from the server. This logic naturally belongs to the file-private side rather than a generic helper operating purely on InodeHandle state.

Anyway, I will remove the `seek` method in this PR.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -404,12 +430,10 @@ impl FileLike for InodeHandle {
             return_errno_with_message!(Errno::EBADF, "the file is opened as a path");
         }
 
-        if let Some(ref file_io) = self.file_io {
-            file_io.check_seekable()?;
-            if file_io.is_offset_aware() {
-                // TODO: Figure out whether we need to add support for seeking from the end of
-                // special files.
```
StevenJiang1110, line 411: The TODO should still be kept. Don't remove it.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -502,6 +526,36 @@ impl Drop for InodeHandle {
     }
 }
 
+impl InodeHandle {
+    fn seek_offset(&self, pos: SeekFrom, end: Option<usize>) -> Result<usize> {
```
StevenJiang1110: I think `do_seek_util` is still a better name than `seek_offset`. I can't understand why the method contains an `offset`.

StevenJiang1110 (approved):
LGTM! Thanks for your contribution! Only two small nits.

This PR will close #3071, right? I have linked this PR to #3071.

kernel/src/fs/file/inode_handle.rs
li041: It seems that `do_seek_util` takes `offset: &Mutex<usize>` from self, so why not implement as a method of InodeHandle?

StevenJiang1110: I mean you can make it a method of `InodeHandle`, but the method name should be `do_seek_util` or `do_seek`, which is better that `seek_offset`.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -542,11 +600,16 @@ pub trait FileIo: Pollable + InodeIo + Any + Send + Sync + 'static {
...
     /// but not seeking (e.g., nsfs).
     ///
-    /// [`check_seekable`]: FileIo::check_seekable
+    /// [`check_seekable`]: SpecialFileOps::check_seekable
     fn check_positional_io(&self) -> Result<()> {
         self.check_seekable()
     }
 
+    /// Reads directory entries from the given offset.
+    fn readdir_at(&self, _offset: usize, _visitor: &mut dyn DirentVisitor) -> Result<usize> {
```
lrh2000: The original design of `InodeIo` is to represent the common methods of `FileIo` and `Inode`. This PR makes `readdir_at` the common part. I wonder, therefore, if it should live in `InodeIo` instead.

li041: I think the main issue is that `readdir` in FUSE is not purely `inode-based`.

Each `READDIR` request needs [fuse_read_in](https://github.com/libfuse/libfuse/blob/d999bd63afee30949443dc237e3a1f210d0c43e0/include/fuse_kernel.h#L827) which is the same as `READ` Operation, and it carries the opened directory handle (`fh`) from the corresponding `opendir`. The inode itself does not carry this information. In Linux FUSE, the `fh` is stored in `file->private_data` rather than in the inode.

So while `readdir_at` looks conceptually similar to inode operations, the actual virtio-fs implementation depends on per-open state from the file/handle layer. That is why I currently keep it on the handle side instead of moving it into `InodeIo`.

Actually, most operations for virtio-fs works on `file->private_data` which is [fuse_file](https://elixir.bootlin.com/linux/v6.19.3/source/fs/fuse/fuse_i.h#L254), the reason that only added a `readdir` there is the incompletion of viriot-fs implementation.

`FileIo` used to acted as I/O methods for specila files, but it can't support virtio-fs which interact with `file->private_data` closely.

lrh2000: Sorry, I wasn't sure if we were talking about the same thing.

https://github.com/asterinas/asterinas/blob/0f50d8dabf4438a90853ec356bc5201f0e01face/kernel/src/fs/file/inode_handle.rs#L528

`InodeIo` is a super trait of `FileIo`. If you have `InodeIo::readdir_at`, then you can provide a file-handle-based implementation.

```rust
impl InodeIo for YourFuseFileHandle {
    fn readdir_at(&self) {}
}

impl FileIo for YourFuseFileHandle {
    // ...
}
```

My original comment says:

> The original design of `InodeIo` is to represent **the common methods of `FileIo` and `Inode`**. This PR makes `readdir_at` the common part. I wonder, therefore, if it should live in `InodeIo` instead.

while means that we can remove `readdir_at` from `FileIo` and `Inode`. Instead we can add `readdir_at` to `InodeIo`.

---

To be honest, the name `IndeIo` may be misleading as well. It was previously a combination of `Inode` and `FileIo`. However, this PR renames `FileIo`, making it even more confusing. I don't know what the proper name is.

https://github.com/asterinas/asterinas/blob/0f50d8dabf4438a90853ec356bc5201f0e01face/kernel/src/fs/vfs/fs_apis/inode.rs#L297-L301

kernel/src/device/registry/block.rs
```diff
@@ -81,14 +81,14 @@ impl Device for BlockFile {
...
         Ok(Box::new(OpenBlockFile(self.0.clone())))
     }
 }
 
 /// Represents an opened block device file ready for I/O operations.
 //
-// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Box` just to implement the `FileIo`
+// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Box` just to implement the `SpecialFileOps`
 // trait. It leads to redundant vtable dispatch and heap allocation. We should devise a better
 // strategy to eliminate the unnecessary intermediate `Box`.
```
lrh2000:
```suggestion
// TODO: This type wraps an `Arc<dyn BlockDevice>` in another `Box` just to implement the
// `SpecialFileOps` trait. It leads to redundant vtable dispatch and heap allocation. We should
// devise a better strategy to eliminate the unnecessary intermediate `Box`.
```

kernel/src/fs/file/inode_handle.rs
StevenJiang1110: Yes, I agree that `InodeIo` should be renamed.

How about `FileBackendOps`? This PR introduces `InodeHandleBackend`, and we essentially have two types of file backends: special files and inodes. So `FileBackendOps` could serve as a common abstraction representing the shared methods of both `SpecialFileOps` and `Inode`.

If we rename InodeIo to `FileBackendOps`, then `readdir_at` could also be moved into `FileBackendOps`, rather than being defined directly on `SpecialFileOps`.

@lrh2000, what do you think of this idea?

lrh2000: > what do you think of this idea?

Thank you for the suggestion. It sounds reasonable to me. I can't think of any better ones at the moment, at least.

tatetian: Where should we draw the line for the interface of special files? With the addition of `readdir_at`, the special file backend takes a partial role of a directory inode. But this interface does not include other directory-related operations such as `create`, `open`, `rename`, and `unlink`. It is unclear to me why `SpecialFileOps` only includes `readdir_at`, but not other directory-related operations.

I want to understand whether this is a temp solution or a long-term one. Since we are rushing for the integration of virtio-fs, I am ok with adding this method as a temp solution. But I need some clarification.

kernel/src/fs/file/inode_handle.rs
```diff
@@ -29,15 +29,34 @@ use crate::{
 
 pub struct InodeHandle {
     path: Path,
-    /// `file_io` is similar to the `file_private` field in Linux's `file` structure. If `file_io`
-    /// is `Some(_)`, typical file operations including `read`, `write`, `poll`, and `ioctl` will
-    /// be provided by `file_io`, instead of `path`.
-    file_io: Option<Box<dyn FileIo>>,
+    /// Backs operations attached to this opened file description.
+    backend: InodeHandleBackend,
```
tatetian: Why `InodeHandleBackend` is an improvement over `Option<Box<dyn SpecialFileOps>>`?

```rust
enum InodeHandleBackend {
    /// The inode-backed implementation.
    InodeBacked,
    /// The opened special-file implementation.
    SpecialFile(Box<dyn SpecialFileOps>),
}

impl InodeHandleBackend {
    fn as_special_file(&self) -> Option<&dyn SpecialFileOps> {
        match self {
            Self::InodeBacked => None,
            Self::SpecialFile(special_file) => Some(special_file.as_ref()),
        }
    }

    fn is_inode_backed(&self) -> bool {
        matches!(self, Self::InodeBacked)
    }
}
```

I think this backend type is misleading because when the file is inode-backed, `InodeHandleBackend` is actually not a backend as it does not provide any useful operations or keep any state. So it should be modeled with `Option<_>`. In addition, the implemented methods, `as_special_file` and `is_inode_backed`, have counterparts on `Option<_>`. So I see no real value of `InodeHandleBackend` over `Option<_>`.

kernel/src/fs/file/inode_handle.rs
li041: > But this interface does not include other directory-related operations such as create, open, rename, and unlink. It is unclear to me why SpecialFileOps only includes readdir_at, but not other directory-related operations.

`create`, `open`, `rmdir` operates on dir inode, and I haven't implement `rename`.  As far as I know, `seek`, `fallocate`, `readdir`, `read`, `write` requires `fh` which is stored in `file->private_data`, and those operations can't implements under inode level.

> I want to understand whether this is a temp solution or a long-term one.

A long-term  abstraction which takes `file->private_data` into account is better. However, this maybe a temp solution due to the lack of my knowledge about this region.

lrh2000: > But this interface does not include other directory-related operations such as `create`, `open`, `rename`, and `unlink`.

I think there is a key difference. `readdir` needs to be called on an opened FD. But other operations do not.

```c
       ssize_t getdents64(size_t count;
                          int fd, void dirp[count], size_t count);
```

```c
       int rename(const char *oldpath, const char *newpath);
       int renameat(int olddirfd, const char *oldpath,
                    int newdirfd, const char *newpath);
       int renameat2(int olddirfd, const char *oldpath,
                    int newdirfd, const char *newpath, unsigned int flags);
```

 - For `rename`, there are no FD arguments. Therefore, we have to provide an inode-based operation.
 - Even for `renameat`, the file to be renamed is `olddirfd/oldpath`. I think it is usually implemented as an inode-based operation, too.

StevenJiang1110: It seems the naming of the `SpecialFileOps` trait is causing the confusion here. The name implies that this trait is intended exclusively for special files such as devices and named pipes, which makes it difficult to reason about whether "special files" encompass directories. As a result, placing `readdir_at` on `SpecialFileOps` feels out of place.

In fact, this trait is intended to correspond to Linux's `file_private` field. The field exists because each file descriptor may have its own per-fd attributes or behaviors — for example, `readdir` may need to cache some directory entries(it depends on the specific filesystem implementation), and `seek` may need to maintain an offset on the fd. Therefore, a name like `OpenFileOps`, `OpenedFileBackend`, or `PerFdOps` would better reflect the intent. Under such a name, the methods defined on this trait would naturally be understood as per-fd operations that may need to access or manipulate per-fd state.

From a longer-term perspective, I am not yet certain what methods should live on this trait, nor whether it will end up resembling `Inode` or `FileLike` more closely.

lrh2000: > `OpenedFileBackend`

Makes me wonder:

```rust
/// A backend that provides basic file operations.
///
/// For inode-backed files without per-`open()` state, [`Inode`] implements this
/// trait directly. For files whose behavior depends on state created by `open`,
/// the per-`open()` backend implements this trait through [`OpenFileBackend`].
trait FileBackend {
    fn read_at(...);
    fn write_at(...);
    fn readdir_at(...) -> Result<usize> { ENOTDIR }
}

/// A per-`open()` backend that provides file operations.
///
/// An open-file backend represents one opened file description. It can hold
/// file-description-specific state and override operations that are not purely
/// inode-backed, such as state and operations for devices, pipes, namespace
/// files, and procfs files.
trait OpenFileBackend: FileBackend {
    fn check_seekable(&self) -> Result<()>;
    fn is_offset_aware(&self) -> bool;
    fn check_positional_io(&self) -> Result<()> { ... }
    fn mappable(&self) -> Result<Mappable> { ... }
    fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> { ... }
}
```

tatetian: @lrh2000 and @StevenJiang1110 Thanks you! I agree that names like `PerFdOps` better describe the mental model. And have two traits, `FileBackend` and `OpenFileBackend`, make this mental model even more clear. I am not sure that `FileBackend` and `OpenFileBackend` are the best names, but they are acceptable.

tatetian: I think we are kind of abusing the word "backend". How about `FileOps` and `PerOpenFileOps`?

lrh2000: `FileOps` is going to clash with procfs though...
```
> grep -R "impl FileOps" kernel
kernel/src/fs/fs_impls/procfs/pid/task/auxv.rs:impl FileOps for AuxvFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/cmdline.rs:impl FileOps for CmdlineFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/comm.rs:impl FileOps for CommFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/environ.rs:impl FileOps for EnvironFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/fd.rs:impl FileOps for FileInfoOps {
kernel/src/fs/fs_impls/procfs/pid/task/gid_map.rs:impl FileOps for GidMapFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/maps.rs:impl FileOpsByHandle for MapsFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/mem.rs:impl FileOpsByHandle for MemFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/mountinfo.rs:impl FileOps for MountInfoFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/mounts.rs:impl FileOps for MountsFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/mountstats.rs:impl FileOps for MountStatsFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/oom_score_adj.rs:impl FileOps for OomScoreAdjFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/stat.rs:impl FileOps for StatFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/status.rs:impl FileOps for StatusFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/uid_map.rs:impl FileOps for UidMapFileOps {
kernel/src/fs/fs_impls/procfs/pid/task/cgroup.rs:impl FileOps for CgroupFileOps {
kernel/src/fs/fs_impls/procfs/sys/kernel/cap_last_cap.rs:impl FileOps for CapLastCapFileOps {
kernel/src/fs/fs_impls/procfs/sys/kernel/pid_max.rs:impl FileOps for PidMaxFileOps {
kernel/src/fs/fs_impls/procfs/sys/kernel/yama.rs:impl FileOps for PtraceScopeFileOps {
kernel/src/fs/fs_impls/procfs/cmdline.rs:impl FileOps for CmdLineFileOps {
kernel/src/fs/fs_impls/procfs/cpuinfo.rs:impl FileOps for CpuInfoFileOps {
kernel/src/fs/fs_impls/procfs/loadavg.rs:impl FileOps for LoadAvgFileOps {
kernel/src/fs/fs_impls/procfs/meminfo.rs:impl FileOps for MemInfoFileOps {
kernel/src/fs/fs_impls/procfs/stat.rs:impl FileOps for StatFileOps {
kernel/src/fs/fs_impls/procfs/uptime.rs:impl FileOps for UptimeFileOps {
kernel/src/fs/fs_impls/procfs/version.rs:impl FileOps for VersionFileOps {
kernel/src/fs/fs_impls/procfs/filesystems.rs:impl FileOps for FileSystemsFileOps {
```

It will occur in the same file...
```
kernel/src/fs/fs_impls/procfs/template/file.rs:impl<F: FileOps + 'static> InodeIo for ProcFile<F> {
```

StevenJiang1110: > FileOps is going to clash with procfs though...

Perhaps we can rename `FileOps` in the procfs module to `ProcFileOps` to resolve the naming conflict. For consistency, the existing `DirOps` and `SymOps` in procfs should also be renamed to `ProcDirOps` and `ProcSymOps` respectively.

If this approach is acceptable, the renaming can be done in a separate commit within this PR.

lrh2000: Note that we have `FileOpsByHandle` in procfs. It may need to be renamed, too.

tatetian: I agree. @li041 Could you make the revision as suggested above?

tatetian (approved):

PR #3187, opened 2026-04-29
Add tests for popular applications

cqs21:
This PR adds tests for popular applications on Asterinas NixOS.

The new coverage is organized by application category and aligned with `book/src/distro/popular-applications/`. It currently includes:

- `system-core`
- `nix-and-nixos-tools`
- `containerization-and-virtualization`
- `networking`
- `web-servers-and-proxies`
- `databases-and-middleware`
- `development-tools`
- `cicd-and-devops`
- `monitoring-and-observability`
- `web-browsers`
- `office-and-productivity`
- `multimedia`
- `communication`
- `file-management-and-terminal`
- `ai-and-machine-learning`

PR #3246, opened 2026-05-19
Fix alias handling in `merge_nixos_config.sh`

vvvvsv:
This PR fixes `test/nixos/common/merge_nixos_config.sh` when the input and output paths alias the same file.

Now, we want `merge_nixos_config.sh <base_file> <extra_file> <merged_file>` to generate a `merged_file` that looks like this:

```nix
{ config, lib, pkgs, ... }:
let
  # The content of the original modules is embedded here.
  baseModule = (
    # The full content of <base_file>
  );
  extraModule = import # The absolute path of <extra_file>
  # Evaluate each module with the standard NixOS arguments 
  base = baseModule { inherit config lib pkgs; }; 
  extra = extraModule { inherit config lib pkgs; }; 
in 
  # Deeply merge the base and extra configurations. 
  # The value from extra_file takes precedence. 
  lib.recursiveUpdate base extra
```

The current implementation breaks when the input and output files refer to the same file. This bug was introduced by https://github.com/asterinas/asterinas/pull/3187#discussion_r3194688661, and will break some [CI](https://github.com/asterinas/asterinas/actions/runs/26025044413/job/76496086117). cc @tatetian @cqs21.

The expected behavior is:

 - `base_file == extra_file == merged_file`: no-op
 - `base_file == extra_file != merged_file`: allow
 - `base_file == merged_file != extra_file`: allow, but read base_file before overwriting it
 - `extra_file == merged_file != base_file`: reject
 - `base_file != extra_file != merged_file != base_file`: allow

### DIFF
--- a/.github/workflows/publish_aster_nixos.yml
+++ b/.github/workflows/publish_aster_nixos.yml
@@ -5,8 +5,9 @@ on:
   push:
     branches:
       - main
-    paths: 
+    paths:
       - VERSION
+      - test/nixos/common/merge_nixos_config.sh
 
 jobs:
   push-cachix:
--- a/.github/workflows/push_cachix_dev.yml
+++ b/.github/workflows/push_cachix_dev.yml
@@ -6,12 +6,14 @@ on:
     paths:
       - .github/workflows/push_cachix_pkgs.yml
       - distro/**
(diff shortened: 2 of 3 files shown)


cqs21 (approved):

PR #3049, opened 2026-03-18
Multiple TTYs

wyt8:
# Multi TTY (Virtual Terminal) Support

This PR implements multi TTY (virtual terminal) support in the Asterinas kernel, extending the existing single-TTY console model to support multiple independent virtual terminals. It brings the TTY/VT behavior of Asterinas closer to Linux-like expectations, enables dynamic VT switching, and lays the foundation for further virtual console improvements.

## Overview

1. Virtual Terminal Manager

  - Introduced a central VT manager responsible for maintaining all created TTY instances.
  
  - Each VT holds its own screen state, cursor position, and console attributes.
  
  - Only the active VT is rendered to the framebuffer at any given time.

2. Runtime VT Switching

  - Supported runtime VT switching between virtual terminals.
  
  - Switching triggers a redraw of the framebuffer using the newly active VT’s stored state.
  
  - Switching behavior follows Linux VT semantics where practical.

3. ioctl & TTY Control Enhancements

  - Enhanced TTY control paths to support VT-related operations.
  
  - Improved handling of terminal control ioctls related to virtual terminal behavior.

## Additional Improvements

1. Remove some hardcode of xorg 

## How to verify it

1. Run with `xorg`

  - Enable `xorg`: Modify the `distro/etc_nixos/configuration.nix`
  
    ```
      # Uncomment the two options below to enable the X11 (X.Org) desktop (XFCE).
      services.xserver.enable = true;
      services.xserver.desktopManager.xfce.enable = true;
    ```
  
  - Run nixos
  
    ```
    make run_nixos
    ```

   - Press `Ctrl + Alt + Fn` or use `chvt` command to switch other TTYs
  
2. Run without `xorg`

  - Run nixos
  
    ```
    make run_nixos
    ```

   - Press `Ctrl + Alt + Fn` or use `chvt` command to switch other TTYs

Issue #2947, opened 2026-02-02
Merge `README.md` with crate-level Rust doc


tatetian:
This issue proposes to merge the `README.md` of a published crate with its crate-level Rust doc. This eliminates content duplication and thus improves maintainability.

## Motivation

The `ostd` crate along with its first-party dependency crates (under [`ostd/libs`](https://github.com/asterinas/asterinas/tree/main/ostd/libs)) are published to crates.io. A published crate needs to have a properly-written `README.md` (displayed on crates.io) and crate-level Rust documentation (usually displayed on docs.rs).

In a published crate in Asterinas, the content of `README.md` is the same as--or can be made same as--the crate-level Rust documentation in `lib.rs`. This means that we maintain duplicated content and require updating both locations for a change, which a fiction for maintainance.

## Solution

To promote a "single source of truth", this issue proposes to embed the content of `README.md` as the crate-level Rust doc using the following technique:

```rust
#![doc = include_str!("../README.md")
```

## A template project

Here is a template project to showcase how to use this technique. This crate has only one source code file, `lib.rs`:

```rust
#![doc = include_str!("../README.md")]

/// My struct.
pub struct MyStruct;

/// Adds two integers.
pub fn add(left: u64, right: u64) -> u64 {
    left + right
}
```

The `README.md` file is shown below:

````markdown
<!--
To promote a "single source of truth",
the content of `README.md` is also included in `lib.rs` as the crate-level documentation.
So when writing this README,
bear in mind that its content should be recognized correctly by
both a Markdown renderer and the rustdoc tool.
-->

# My Crate

My crate aims to showcase the technique of `#[doc = include_str!("../README.md")`.

## Introduction

This is the introduction.

## APIs

The crate provides only one struct, [`MyStruct`], and one function, [`add`].

## Examples

<!-- This code snippet is rendered correctly by a Markdown renderer 
and also runnable with `cargo test`. -->
```rust
let res = my_crate::add(1, 2);
assert!(res == 3);
```

<!--
Links like `[MyStruct]` and `[add]` that work fine with `rustdoc` do not work 
with a Markdown renderer. So we need to give their external links.
-->
[`MyStruct`]: https://docs.rs/my-crate/latest/my_crate/struct.MyStruct.html
[`add`]: https://docs.rs/my-crate/latest/my_crate/fn.add.html
````

## Goal

Use the technique shown above to refactor all published crates in Asterinas.

PR #3247, opened 2026-05-19
Add `short-vis-path` crate and replace restricted visibility paths by short subsystem names

### LINKED ISSUE #3188: Concise subsystem-scoped visibility for `aster-kernel`
## Summary

Introduce a small, general-purpose proc-macro crate, `short-vis-path`, that lets a contributor declare short names for visibility paths at the top of a file:

```rust
#![::short_vis_path::add(procfs)]

pub(in procfs) struct ProcDir { ... }
pub(in procfs) fn next_ino() -> u64 { ... }
```

The inner attribute `add` declares short names. The macro walks the file, finds every `pub(in <name>)` whose name was declared, and rewrites it to the full `pub(in crate::a::b::...::<name>)` path. Other visibilities (`pub`, `pub(crate)`, `pub(super)`, `pub(in crate::...)`, private) are untouched.

`add` accepts three call shapes:

```rust
#![::short_vis_path::add(procfs)]                                  // infer path from file location
#![::short_vis_path::add(procfs, fs)]                              // multiple short names
#![::short_vis_path::add(procfs = crate::fs::fs_impls::procfs)]    // explicit override
```

A pilot of the design was integrated into one procfs file (see "Pilot" below). `cargo osdk build` produces the kernel image; rust-analyzer's CLI diagnostics report no errors on the pilot file.

## Motivation

The Asterinas coding guideline asks every item to carry the *narrowest* visibility it needs. The intent is local readability: a reader can determine who may use an item without inspecting the parent module, the grandparent module, or any re-export chain.

That guideline collides head-on with the structure of `aster-kernel`. The kernel mimics a monolithic Linux kernel, so subsystems live deep in the tree and are deliberately not extracted into sub-crates (procfs surfaces in-kernel state of *every* other subsystem; mm, net, ipc, sys are all similarly entangled). Today these subsystems express "internal to this subsystem" by writing the full path:

```rust
pub(in crate::fs::fs_impls::procfs) struct ProcDir { ... }
pub(in crate::fs::fs_impls::procfs) fn next_ino() -> u64 { ... }
pub(in crate::fs::fs_impls::procfs) mod helpers { ... }
```

Three concrete costs:

1. **Verbosity.** The path repeats on every item — 35 characters before the keyword (`fn`, `struct`, …) for a procfs item, vs. 11 for `pub(crate)`.
2. **Refactor friction.** Renaming or moving a subsystem (e.g., promoting `procfs` out of `fs_impls/`) is a sweeping multi-file edit.
3. **Reading friction.** A reader has to parse a long path to learn what `pub(crate)` would tell them at a glance — and has to know where the subsystem lives in the tree, which the path itself was supposed to abstract away.

The pain is broad. On `main` today, the `kernel` crate alone has 89 `pub(in crate::...)` occurrences spread over 7 distinct subsystem paths.

<details>
<summary>Reproduce: <code>git grep</code> commands and output</summary>

```console
$ git grep -c 'pub(in crate::' main -- 'kernel/src/**/*.rs' \
    | awk -F: '{s+=$NF} END {print s}'
89

$ git grep -h 'pub(in crate::' main -- 'kernel/src/**/*.rs' \
    | grep -oE 'pub\(in crate::[a-zA-Z_:]+\)' | sort -u
pub(in crate::device::tty::vt::keyboard)
pub(in crate::fs)
pub(in crate::fs::fs_impls::procfs)
pub(in crate::net)
pub(in crate::net::socket::netlink)
pub(in crate::process)
pub(in crate::vm::vmar)

$ git grep -c 'pub(in crate::' main -- 'kernel/src/**/*.rs' \
    | sort -t: -k3 -rn | head -10
main:kernel/src/fs/vfs/path/mount.rs:8
main:kernel/src/net/socket/util/options.rs:7
main:kernel/src/process/posix_thread/thread_local.rs:5
main:kernel/src/net/socket/netlink/common/bound.rs:5
main:kernel/src/device/tty/vt/keyboard/keysym.rs:5
main:kernel/src/fs/vfs/path/mod.rs:4
main:kernel/src/vm/vmar/vmar_impls/page_fault.rs:3
main:kernel/src/process/signal/sig_queues.rs:3
main:kernel/src/process/process/session.rs:3
main:kernel/src/process/process/process_group.rs:3
```

</details>

## Goals

We want a new annotation mechanism for **subsystem-scoped visibility** that is:

- **Short**: the annotation at the use site is short and sweet.
- **Local**: a reader can see *which subsystem* the item is private to directly from the visibility annotation.
- **Friendly to `rustfmt`**: the annotation is fully formatted by `rustfmt` (the kernel runs `make format` and `make check`).
- **Non-intrusive**: it does not change the semantics of any existing visibility (`pub`, `pub(crate)`, `pub(super)`, private), and adoption requires no central registry of subsystem names to maintain.

### Recommended: file-level inner attribute

A contributor declares short names once at the top of a file with `add(...)`. Bodies then use the natural Rust syntax `pub(in <name>)`:

```rust
#![::short_vis_path::add(procfs)]

pub(in procfs) struct ProcDir { ... }                  // → pub(in crate::fs::fs_impls::procfs)
pub(in procfs) fn next_ino() -> u64 { ... }            // → pub(in crate::fs::fs_impls::procfs)
pub fn kernel_public() -> u64 { ... }                  // untouched
pub(crate) fn keep_pub_crate() -> u64 { ... }          // untouched
fn private_helper() -> u64 { ... }                     // untouched
impl ProcDir {
    pub(in procfs) fn new(...) -> Self { ... }
    fn inner(&self) -> u64 { ... }
}
```

`add` accepts three call shapes:

```rust
#![::short_vis_path::add(procfs)]                                  // infer path from file location
#![::short_vis_path::add(procfs, fs)]                              // multiple short names
#![::short_vis_path::add(procfs = crate::fs::fs_impls::procfs)]    // explicit override
```

The single-ident form runs the standard heuristic (find the named ancestor in the file's module path). The list form is for files that need to expose items at multiple subsystem levels — for instance, a file inside `procfs` that declares both `add(procfs, fs)` to expose some items at procfs scope and others at the wider `fs` scope. The override form (`add(name = crate::full::path)`) is the escape hatch for cases the heuristic gets wrong: `#[path = "..."]` overrides, name collisions, files outside the standard layout, or any case where the contributor wants to be explicit.

What the design gives:

- **One declaration per file**, not per item. With 89 occurrences of subsystem-visibility in the kernel today, that is a meaningful reduction in source-line overhead.
- **Locality of intent.** A reader scanning the file header sees exactly which short names are in play and treats `pub(in procfs)` in the body without surprise. Conversely, on encountering `pub(in procfs)` first, scanning to the file header explains the non-standard syntax.
- **Items in the body read as ordinary Rust.** `pub(in <ident>)` is valid Rust syntax (rustc only validates the path during name resolution, *after* attribute-macro expansion). No new keywords, no magic sentinels: the visibility annotation literally says "visible within `procfs`".
- **All other visibilities are untouched.** The macro touches exactly `pub(in <name>)` where `<name>` was declared in an `add(...)` invocation. Bare `pub`, `pub(crate)`, `pub(super)`, `pub(in crate::...)`, and private all retain their normal Rust meaning.
- **Explicit override** for cases the heuristic gets wrong.

### Concessions

- **Two nightly feature flags** at the crate root, `custom_inner_attributes` and `proc_macro_hygiene`. The kernel is already on a recent nightly with 13 `#![feature(...)]` flags in `kernel/src/lib.rs`; this adds two more.
- **rust-analyzer's inner-attribute proc-macro path is a real risk.** The pilot shows clean diagnostics, but the IDE experience under active development should be re-checked on each rust-analyzer upgrade. See limitation 2.

### Comparison vs. alternatives

| | Status quo | Block wrapper | Bare-`pub` inner | **Recommended** |
|---|---|---|---|---|
| Short at use site         | no  | yes | yes | **yes** |
| Locally readable          | no  | yes | yes | **yes** |
| `rustfmt`-clean           | yes | yes | yes | **yes** |
| No bookkeeping per subsys | yes | yes | yes | **yes** |
| Preserves real `pub`      | yes | no  | no  | **yes** |
| No extra indentation      | yes | no  | yes | **yes** |
| File-level intent         | n/a | yes | yes | **yes** |
| Override for edge cases   | n/a | no  | no  | **yes** |

The recommended design is the only option that scores positively on every row. Each alternative is detailed in "Alternatives considered" below.

## Pilot

A pilot was applied to `kernel/src/fs/fs_impls/procfs/template/dir.rs`, which previously contained the kernel's only `pub(in crate::fs::fs_impls::procfs)` annotation in the procfs subtree.

Diff (essentials):

```diff
 // SPDX-License-Identifier: MPL-2.0

+#![::short_vis_path::add(procfs)]
+
 use core::time::Duration;
 // ... unchanged imports and struct definition ...

 impl<D: DirOps> ProcDir<D> {
-    pub(in crate::fs::fs_impls::procfs) fn new_root(
+    pub(in procfs) fn new_root(
         dir: D,
         fs: Weak<dyn FileSystem>,
         ...
     ) -> Arc<Self> { ... }
 }
```

Plumbing (one-time, in four places):

- `kernel/libs/short-vis-path/` — new proc-macro crate, ~110 lines.
- `Cargo.toml` (workspace root): `short-vis-path` added to `workspace.members` and `workspace.dependencies`.
- `kernel/Cargo.toml`: `short-vis-path.workspace = true`.
- `kernel/src/lib.rs`: `#![feature(custom_inner_attributes)]` and `#![feature(proc_macro_hygiene)]`.

Verifications:

1. **`cargo osdk build` succeeds end-to-end** and produces `target/osdk/aster-kernel-osdk-bin.iso`. The fact that `pub(in procfs)` resolved at all proves the inner-attribute proc-macro expanded correctly — without expansion, `procfs` is not directly in scope from inside `procfs::template::dir`, so name resolution would fail with E0742.
2. **`rust-analyzer diagnostics` and `analysis-stats`** report **zero diagnostics** referencing the pilot file, the `short-vis-path` crate, or the `add` macro. (rust-analyzer's full type-check on the kernel codebase has many *pre-existing* diagnostics for unrelated reasons — including a `proc-macro-srv` serialisation bug that currently affects most existing proc-macros — but none of them are introduced by this work.)

### Public API

A new general-purpose crate, `kernel/libs/short-vis-path`, exposes one attribute proc-macro: an inner attribute `add` invoked at the top of a file.

```rust
#![::short_vis_path::add(procfs)]
#![::short_vis_path::add(procfs, fs)]
#![::short_vis_path::add(procfs = crate::fs::fs_impls::procfs)]

pub(in procfs) struct ProcDir { ... }
```

The macro accepts a comma-separated list of declarations. Each declaration is either a single bare ident (resolved to the nearest ancestor module of that name) or an `<ident> = <full::path>` form (used verbatim, for the cases where the heuristic does not apply).

### Resolution algorithm

For a single-ident argument `name`:

1. `let file = proc_macro::Span::call_site().local_file()?;`
2. Split `file` into path components.
3. Find the **last** `src` segment. Treat everything after it as module-path parts.
4. Strip a trailing `.rs` extension; drop a trailing `mod`, `lib`, or `main` segment.
5. `rposition` the components for `name`. If found, the prefix up to and including that position is the resolved path. If not found, the macro emits a clear error — see limitation 7.
6. Emit `pub(in crate::a::b::...::<name>)`.

For an explicit `name = path` argument, steps 1–5 are skipped and `path` is used directly.

The inner attribute (`add`) collects all declared names into a small table, then walks every item in the surrounding scope. For each item whose outer visibility matches the form `pub(in <single-ident>)` with an ident in the table, it replaces that visibility with the resolved full path. For `Item::Impl`, it descends into the impl items and applies the same rule to methods, associated constants, and associated types. Other item kinds, struct fields, enum variants, and trait items are left untouched.

### Crate dependencies

`syn = "2"`, `quote = "1"`, `proc-macro2 = "1"`. All already declared in `workspace.dependencies` and used by other kernel proc-macro crates (`aster-rights-proc`, `int-to-c-enum/derive`, `ostd-pod/macros`, `ostd-macros`).

### Adoption / migration

Adoption is **opt-in and incremental**:

- A subsystem maintainer can switch their files one at a time. The explicit-path form (status quo) continues to compile alongside the new form, so a half-migrated subsystem is fine.
- Once a subsystem is fully migrated, a `clippy` or grep-based lint can flag any remaining `pub(in crate::<subsys-path>)` occurrences.

No existing code requires changes. The crate is purely additive.

## Known limitations

1. **Two nightly feature flags.** `custom_inner_attributes` and `proc_macro_hygiene` must be enabled at the consuming crate's root (`kernel/src/lib.rs`). The asterinas kernel is already on nightly with many other feature flags; these two are well-trodden, with long-standing tracking issues. Their stabilisation status should be monitored.
2. **IDE / rust-analyzer compatibility is a real risk.** rust-analyzer expands proc-macros via an out-of-process server that has historically been roughest around inner attributes (`#![attr]`). The pilot shows clean output from `rust-analyzer diagnostics` on the converted file, but a CLI run cannot fully simulate an interactive editor session. If the inner-attribute proc-macro fails to expand in the editor, contributors will see spurious "unresolved module" reports on `pub(in procfs)`. A contributor encountering this can fall back to writing the explicit `pub(in crate::full::path)` form (the status quo) for the affected file. IDE behaviour should be re-checked on each rust-analyzer update.
3. **Field-level subsystem visibility is not yet rewritten.** The inner-attribute macro currently walks items but does not descend into struct fields. So `pub(in procfs) field: T` inside a struct will *not* be expanded, and rustc will reject it at name resolution. Contributors who need field-level subsystem visibility must spell the full path: `pub(in crate::fs::fs_impls::procfs) field: T`. This is a property of the macro's current implementation, not a Rust limitation; extending the walk to fields is a small follow-up if the kernel's procfs ever needs it. (Today most procfs structs keep fields private, so this rarely bites.)
4. **The heuristic depends on `<crate>/src/...` layout.** The file-path-to-module-path mapping assumes the standard Cargo layout. Files outside `src/` (`tests/`, `examples/`, `benches/`) yield a clear error from the macro. For files where the heuristic does not apply — `#[path = "..."]` overrides, inline `mod foo { ... }` blocks, or non-standard layouts — the override form `add(name = crate::full::path)` lets the contributor specify the path explicitly. So the heuristic's limits are mitigated, not forced.
5. **Same-name collisions.** If a subsystem name appears more than once in the path (e.g., `crate::procfs::foo::procfs::bar.rs`), the deepest match wins (`rposition`). This is the right default for subsystems, and the override form is available for contributors who want a specific match.
6. **Stable Rust 1.88 floor for `Span::local_file()`.** Already met by the kernel toolchain.
7. **Diagnostics on bad subsystem names.** A misspelled `add(procf)` (note the typo) emits a clear error from the macro: *subsystem `procf` not found in module path of this file*. A misspelled `pub(in procf)` inside a file whose `add(...)` declares `procfs` is left unrewritten by the macro and then rejected by rustc with E0742 (*visibilities can only be restricted to ancestor modules*). Both errors point at the relevant token; neither macro currently offers a "did you mean ...?" suggestion. Acceptable; can be improved later.

## Alternatives considered

Each option below was implemented and tested against a sample crate that mirrors the kernel's procfs layout. The recommended design (the inner attribute `add(...)`, described above) is what survived the comparison; everything below was rejected.

### Status quo

Keep writing `pub(in crate::fs::fs_impls::procfs)`. Zero machinery, but the costs in the Motivation stand.

### Relative paths: `pub(in super::super[::super]*)`

Built into Rust, no machinery. Two problems: the depth differs per file, so the form is not uniform across a subsystem; and moving a file across directories silently changes the meaning. Hard to enforce mechanically.

### `macro_rules!` wrapper per subsystem

A subsystem-local function-like macro that prepends the long path:

```rust
pub_subsys! { struct ProcDir { ... } }
```

**Empirically rules itself out.** `rustfmt` does not format the body of arbitrary user `macro_rules!` invocations — even with the nightly options `format_macro_bodies = true` and `format_macro_matchers = true`. Wrapped items become a formatting blind spot. Verified against the kernel's `rustfmt.toml` with `rustfmt 1.8.0-nightly (2025-12-05)`.

### Per-item attribute proc-macro: `#[short_vis_path::pub_in(procfs)]`

A small attribute macro that takes a subsystem name as its argument and rewrites the item's outer visibility to `pub(in crate::a::b::...::<name>)`.

Empirically, `rustfmt` always splits an item attribute onto its own line. So the realistic call-site shape is two lines per item:

```rust
#[short_vis_path::pub_in(procfs)]
struct ProcDir { ino: u64, name: &'static str }
```

Compared with the status-quo prefix `pub(in crate::fs::fs_impls::procfs) struct ProcDir { ... }` (one line, ~35 characters of visibility text), the attribute form takes two source lines and the `short_vis_path::pub_in(procfs)` qualifier is barely shorter than the path it replaces. The form is dominated by status quo on every axis worth measuring (line count, character count, locality). Not adopted.

### Function-like wrapper: `pub_subsys!(name, { items })`

A function-like proc-macro that wraps a block of items, rewriting every `pub` inside to `pub(in <full path>)` and recursing into `impl` blocks.

```rust
pub_subsys!(procfs, {
    pub struct ProcDir { ... }
    pub fn next_ino() -> u64 { ... }
    pub(crate) fn keep_pub_crate() -> u64 { 2 } // not rewritten
    impl ProcDir {
        pub fn new(...) -> Self { ... }
        fn inner(&self) -> u64 { self.ino }
    }
});
```

Merit over the per-item attribute: the subsystem name is written once per block, not per item.

`rustfmt` did format the body of this form (verified empirically; the `(name, { items })` shape is recognised, unlike the bare `name! { items }` form of the `macro_rules!` wrapper above). However, every wrapped item is **indented one level deeper**, which is a step backwards relative to status quo and was an explicit non-starter for this RFC.

### Inner attribute that rewrites bare `pub`: `#![short_vis_path::rewrite_pub(name)]`

Same engine as the function-like wrapper but applied as an inner attribute on the file:

```rust
#![::short_vis_path::rewrite_pub(procfs)]

pub struct ProcDir { ... }                  // rewritten to pub(in ...::procfs)
pub fn next_ino() -> u64 { ... }            // rewritten
pub(crate) fn keep_pub_crate() -> u64 { 2 } // untouched
fn private_helper() -> u64 { 3 }            // untouched
```

Zero extra indentation. `rustfmt` formats normally. Requires the nightly feature flags `custom_inner_attributes` and `proc_macro_hygiene`.

The drawback: contributors lose the ability to write a real `pub` in the file, because every bare `pub` gets rewritten to subsystem-internal. To expose something `pub` to the rest of the kernel from inside the file, you must spell `pub(crate)` instead — which works but is mildly contortionist. The recommended `add(...)` design avoids this trade-off by using a `pub(in <name>)` declaration that is only triggered by named opt-in.

## Open questions

- **Crate name and location.** The pilot uses `kernel/libs/short-vis-path`, exporting `short_vis_path::add`. The crate is general-purpose — it has nothing to do with the kernel domain — and could equally be a standalone crates.io publication if that ever became useful.
- **Should we offer `#[pub_in_self]` or similar shorthands** for `pub(in self)` / `pub(super)`? Author's view: no — those forms are already short. This proposal stays narrowly focused on the long-path problem.
- **Should `make check` add a lint** that flags `pub(in crate::...)` paths for which the short form would be equivalent? Out of scope for this RFC, but a natural follow-up once adoption is broad.

## Reference

- Proof-of-concept and design notebook: `worktrees/subsys-vis-poc/`, including `subsys_vis/src/lib.rs` (the design's first incarnation, ~150 lines), `sample_kernel/` (mirrors the deep procfs layout, exercises every alternative above), and `tests/ui/` (compile-fail snapshots verifying the rewritten visibility correctly rejects extra-subsystem access).
- Pilot in the real kernel: `kernel/libs/short-vis-path/` and the conversion of `kernel/src/fs/fs_impls/procfs/template/dir.rs`.

zjp-CN:
## Summary

Implements the design proposed in #3188 — concise subsystem-scoped visibility for `aster-kernel` — and rolls it out beyond the pilot. Three commits:

1. **Add short-vis-path crate** — new `ostd/libs/short-vis-path` proc-macro crate.
2. **Replace `pub(in crate::*subsystem)` by `pub(in subsystem)` in kernel** — applies the macro across `kernel` (54 files).
3. **Replace `pub(in crate::*::subsystem)` by `pub(in subsystem)` in ostd** — applies the macro across `ostd` (48 files).

Closes #3188.

## Background

#3188 (RFC, `C-design-proposal`) argues that the kernel's `pub(in crate::a::b::subsystem)` paths impose three concrete costs:

1. **Verbosity** — 35+ characters before the item keyword, vs. 11 for `pub(crate)`.
2. **Refactor friction** — promoting or renaming a subsystem forces sweeping multi-file edits.
3. **Reading friction** — a long path obscures the very locality information it is supposed to convey.

The RFC's grep on `main` counted 89 occurrences in `kernel/` alone across 7 subsystem paths. The same pain exists in `ostd/`, which is why this PR widens the rollout.

The proposed mechanism is a file-level inner attribute:

```rust
#![short_vis_path::add(procfs)]

pub(in procfs) struct ProcDir { ... }  // → pub(in crate::fs::fs_impls::procfs)
```

The macro infers the full path from the source file's location and rewrites only `pub(in <name>)` whose `<name>` was declared in `add(...)`. Other visibilities (`pub`, `pub(crate)`, `pub(super)`, `pub(in crate::...)`, private) are untouched.

## Implementation

`ostd/libs/short-vis-path` exposes a single attribute proc-macro `add` used as an inner attribute. It accepts the three call shapes specified in the RFC:

```rust
#![short_vis_path::add(procfs)]                                  // infer from file location
#![short_vis_path::add(procfs, fs)]                              // multiple short names
#![short_vis_path::add(procfs = crate::fs::fs_impls::procfs)]    // explicit override
```

Resolution algorithm (`ostd/libs/short-vis-path/src/args.rs`, `ExpandedPath::new`):

1. `proc_macro::Span::call_site().local_file()` + `CARGO_MANIFEST_DIR` ⇒ relative source path.
2. Strip the leading `src/`; strip `/mod.rs` or `.rs` from the tail.
3. Split on `/` to get the module path starting at `crate`.
4. For each bare-ident argument, `rposition` the named segment (deepest match wins). On miss, panic with a message naming both the unknown ident and the file's inferred path.

Rewriting (`AddArguments::visit_visibility_mut`):

- A `VisitMut` traversal touches **every** `Visibility::Restricted` whose path is a single ident in the table — items, methods, associated items, **and struct fields**. (This goes beyond Known Limitation 3 in the RFC, which stated field-level rewriting was not implemented in the POC.)
- `pub(in path) macro foo` is not yet parseable by `syn`, so `Item::Verbatim` is patched at the token-tree level (`replace_verbatim_vis_path`).

## Differences from the RFC

- **Field-level support.** Implemented; the RFC's Known Limitation 3 no longer applies.
- **Extra usage guidelines.** `short_vis_path` should only be used for deep subsystem paths that need repeated restricted visibility, when `pub(super)`/`pub(self)` are not enough, and the path must still expand to an ancestor module.

What is unchanged from the RFC: the user-visible syntax, the resolution heuristic (deepest match, override form as escape hatch), the visibility set the macro is allowed to touch, the two nightly features (`custom_inner_attributes`, `proc_macro_hygiene`), and the "opt-in, additive, no central registry" property.

## Known limitations (unchanged from the RFC where applicable)

- **Nightly feature flags.** `custom_inner_attributes` + `proc_macro_hygiene` are now load-bearing for `ostd` and `kernel`. Both have long-standing tracking issues; stabilisation should be monitored.
- **Obscure error diagnostics.** In some edge cases (listed below), the unstable `custom_inner_attributes` is incomplete and has bugs.
  - https://github.com/os-checker/short-vis-path/issues/1
  - https://github.com/os-checker/short-vis-path/issues/2
  - https://github.com/os-checker/short-vis-path/issues/3
- **`<crate>/src/...` layout assumed.** Files outside that layout, `#[path = "..."]` overrides, or name collisions can use the `name = crate::path` override form.

---

<details><summary>Stats of  short subsys names and touched files.</summary>
<p>

| Submodule | Long module path (`crate::...`) | Number of substitutions | Number of files changed |
|---|---|---:|---:|
| `vt` | `crate::device::tty::vt` | 11 | 1 |
| `ext2` | `crate::fs::fs_impls::ext2` | 38 | 6 |
| `inode` | `crate::fs::fs_impls::ext2::inode` | 14 | 1 |
| `virtiofs` | `crate::fs::fs_impls::virtiofs` | 6 | 1 |
| `netlink` | `crate::net::socket::netlink` | 6 | 2 |
| `vsock` | `crate::net::socket::vsock` | 23 | 8 |
| `vmar` | `crate::vm::vmar` | 3 | 1 |

- This commit makes **101** short visibility-path substitutions in total.
- Those changes are spread across **20 files**.

</p>
</details>


kernel/libs/short-vis-path/tests/tests.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use expect_test::expect_file;
+
+macro_rules! snapshot {
+    ($name:ident $success:literal: $stdout:expr; $stderr:expr) => {
+        #[test]
+        fn $name() {
+            let dir = stringify!($name);
+            let output = Command::new("cargo")
+                .arg("expand")
```
zjp-CN: The test job requires `cargo-expand` being available in the docker.  I'll open another PR to install the tool and bump image version.

In other words, basic-test must fail for now.

kernel/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
+#![short_vis_path::add(fs = crate::fs)]
```
zjp-CN: Note: `![short_vis_path::add(fs)]` expands `pub(in fs)` to a path referring to the current module `crate::fs::fs_impls::ramfs::fs` **by default**, which makes some code not compile any more, so I override the module path here.

ostd/src/arch/riscv/trap/trap.rs
```diff
@@ -14,6 +14,8 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
+#![short_vis_path::add(arch)]
```
zjp-CN: For the failed [riscv compile](https://github.com/asterinas/asterinas/actions/runs/26088170843/job/76706582443?pr=3247) job, multiple `global_asm!` statements are not guaranteed to be concatenated continuously, on the contrary, it's [discouraged](https://dirname.github.io/rust-std-doc/unstable-book/library-features/global-asm.html) to share states

> You may use global_asm! multiple times, anywhere in your crate, in whatever way suits you. However, you should not rely on assembler state (e.g. assembler macros) defined in one global_asm! to be available in another one. It is implementation-defined whether the multiple usages are concatenated into one or assembled separately.

This seems to be a good [blind spot](https://github.com/rust-embedded/riscv/issues/321) in low-level crates.

ostd/libs/short-vis-path/tests/err_bad_diagnostic/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "err_bad_diagnostic"
```
zjp-CN: Seems all Asterinas crate names should have kebab-case instead of snake_case, but this rule is not clear in the [naming guidelines](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/naming.html) which refer to [this](https://rust-lang.github.io/api-guidelines/naming.html).

tatetian:
Thanks for implementing this new feature so promptly!

In my original design proposal, I raised the following concern:

> rust-analyzer's inner-attribute proc-macro path is a real risk. The pilot shows clean diagnostics, but the IDE experience under active development should be re-checked on each rust-analyzer upgrade. See limitation 2.

Did try this macro in an IDE equipped with rust-analyzer? I want to make sure that rust-analyzer does not report false positive errors because this unconventional macro.

ostd/libs/short-vis-path/tests/err_bad_diagnostic/Cargo.toml
tatetian: Yes, Asterinas prefers kebab-case for crate names and this is unspecified in Rust's official documentation. Maybe we should add this rule as a coding guideline in Asterinas Book.

lrh2000:
> `keyboard` 	`crate::device::tty::vt::keyboard` 	1 	5

They were added by mistake. They should use `pub(super)` instead. Those will be fixed in https://github.com/asterinas/asterinas/pull/3049.

> `fs` 	`crate::fs` 	16 	43

I think `crate::subsystem` is fine because it's straightforward to read and write.

But `crate::sub::sys::tem` is indeed very painful to read and write.

So, I'm leaning towards replacing all occurrences of `pub(in crate::xx::yy)` with at least two `::` separators, but not those with only one.

In addition, if the file contains only one instance, perhaps it does not need to be replaced. We can replace the ones that occur multiple times (e.g., `>= 3` times?)

If we follow this rule, perhaps there won't be so many visibility markers that need to be replaced. But I think it's fine. There aren't many visibility markers available for replacement because `pub(in crate::xx::yy)` is too embarrassing to write. In the past, we just wrote `pub` instead.

In the future, there should be more occurrences of `short-vis-path` because `pub(in zz)` is much easier to write compared to `pub(in crate::xx::yy::zz)`.

> * **Obscure error diagnostics.** In some edge cases (listed below), the unstable `custom_inner_attributes` is incomplete and has bugs.

This is a bit unfortunate. Does it work well with rust-analyzer? Same question as https://github.com/asterinas/asterinas/pull/3247#issuecomment-4514240208.

tatetian:
> So, I'm leaning towards replacing all occurrences of `pub(in crate::xx::yy)` with at least two `::` separators, but not those with only one.

Agree. Prefer `pub(in crate::xxx)` over `pub(in xxx)` enabled with `#![short_vis_path::add(xxx)]`. We should persist this style into the coding guildeline.

zjp-CN:
> I'm leaning towards replacing all occurrences of `pub(in crate::xx::yy)` with at least two `::` separators, but not those with only one.

This means only 20% of the current replacement should be kept, as 80% targets submodules at just one level below the root.

<details><summary>Details</summary>
<p>

| Level | Module | Count | Percentage | Crate |
|-------|--------|-------|------------|-------|
| L1 | arch | 44 | 20.8% | ostd |
| L1 | fs | 43 | 20.3% | kernel |
| L1 | process | 32 | 15.1% | kernel |
| L1 | net | 24 | 11.3% | kernel |
| L3 | vsock | 24 | 11.3% | kernel |
| L1 | mm | 19 | 9.0% | ostd |
| L1 | task | 7 | 3.3% | ostd |
| L3 | netlink | 6 | 2.8% | kernel |
| L4 | keyboard | 5 | 2.4% | kernel |
| L2 | vmar | 3 | 1.4% | kernel |
| L1 | io | 2 | 0.9% | ostd |
| L2 | page_table | 2 | 0.9% | ostd |
| L1 | ipc | 1 | 0.5% | kernel |
| **Total** | | **212** | **100%** | |
| _L1 sub_ | |  **172** | **81.1%** |
| | _ostd_ | 72 | 41.9% |
| | _kernel_ | 100 | 58.1% |
| _L2 sub_ | |  **5** | **2.4%** |
| | _ostd_ | 2 | 40.0% |
| | _kernel_ | 3 | 60.0% |
| _L3 sub_ | |  **30** | **14.2%** |
| | _kernel_ | 30 | 100.0% |
| _L4 sub_ | |  **5** | **2.4%** |
| | _kernel_ | 5 | 100.0% |

</p>
</details> 


> We should persist this style into the coding guildeline.


I'll update the commits to restore simple visibility paths and add the corresponding rules to https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.html .

zjp-CN:
> Did try this macro in an IDE equipped with rust-analyzer? I want to make sure that rust-analyzer does not report false positive errors because this unconventional macro.

The `short-vis-path` crate and rust-analyzer (RA) generally work well together in most scenarios. Friction only occurs in certain edge cases—for instance, using an inner attribute macro in an outer attribute form can trigger diagnostics that appear exclusively in RA. That said, fundamental compiler-side issues regarding macro expansion remain much more challenging to address, as detailed in the links provided in the description.

<details><summary>Good with RA</summary>
<p>

https://github.com/user-attachments/assets/1cef96f1-dcb4-4f2a-88ff-a82b1c6aee82

</p>
</details> 


<details><summary>Edgecase: outer `#[short_vis_path::add]` compiles and works, but RA complains
</summary>
<p>

image


</p>
</details> 

By the way, I didn't see spurious "unresolved module" reports as mentioned in the RFC.

zjp-CN:
So the rules  to use `short-vis-path` are

* The submodule depth exceeds 2 levels: the target path must contain at least two `::` separators
* **AND** the restricted visibility path is used at least 3 times

**Question:** For a long module path that appears only once, should we still avoid `short-vis-path` and keep the original full path? I think the answer is yes, the path keeps untouched. But it causes inconsistensy: you'll have to handle two cases: `pub(in vsock)` and `pub(in crate::net::socket::vsock)`.

Edit:  ostd submodules don't need short-vis-path under the rules, so I'll remove the changes on ostd.

kernel/src/fs/fs_impls/ramfs/fs.rs
zjp-CN: This won't happen due to the not-use-short-vis-path [rules](https://github.com/asterinas/asterinas/pull/3247#issuecomment-4517792328).

tatetian:
Let's wait for the new Docker image: https://github.com/asterinas/asterinas/actions/runs/26325594827. This time it should work.

tatetian:
> * **AND** the restricted visibility path is used at least 3 times

I think we can lower the par to "at least twice". If it appears twice, we have good reason to expect the long restricted visibility path to appear more as the module get evolved.

tatetian:
> The `short-vis-path` crate and rust-analyzer (RA) generally work well together in most scenarios.

Thanks for the clarification! Our coding guideline can advise against using the outer attribute form to avoid complaints from rust-analyzer.

kernel/libs/short-vis-path/tests/err-decl-macro-absolute-path/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#![feature(decl_macro)]
+#![feature(custom_inner_attributes)]
+
+mod mm {
+    // This produces macro-expanded error in cpu module.
+    #![short_vis_path::add(mm = crate::mm)]
+
+    #[macro_export]
+    macro_rules! impl_frame_meta_for {
```
tatetian: I appreciate that you have added multiple test cases. But they are not written in a very readable form. For one thing, the current tests contain a lot of "noises" that distract the reader from getting the true intention of the tests. Preferably, each test case should test one specific feature, command, aspect, property, or historic bug of a target system. Test cases should be written in a way that make this single point clear. As such, these test cases should not include macro items like `impl_frame_meta_for`, modules like `mm`, and files like `procfs`. Obviously, they are not essential for testing the `short-vis-path` and thus are distractions.

Furthermore, the tests deserve some documentation. I have no idea why multiple `cargo_expand.rs` files are included. I don't think simply running `cargo test` is sufficient to run these test, is it? You may want to explain how the macro is tested and how the test files are organized by adding a README file under `tests/`. The `ok-single-indent` test especially deserves more explanation because it contains multiple sub-directories.

kernel/libs/short-vis-path/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MPL-2.0
```
tatetian, line 1: This crate deserves a `README.md` and a crate-level doc comments.

Use the technique introduced in #2947 to consolidate the doc in `README.md` and `lib.rs`. One example crate that has adopted this technique is [`ostd-pod`](https://github.com/asterinas/asterinas/tree/main/ostd/libs/ostd-pod).

kernel/libs/short-vis-path/tests/ok-decl-macro-2/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+                };
+            }
+        }
+    }
+}
+
+pub use crate::mm::frame::meta::{check_frame_meta_layout, impl_frame_meta_for};
+
+mod cpu {
+    crate::mm::frame::meta::impl_frame_meta_for!(S);
```
tatetian: I am confused as to the purpose of this test case? None of the defined macros use the short visibility `mm`.

kernel/libs/short-vis-path/tests/err-single-ident/src/procfs.rs
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#![short_vis_path::add(procfs)]
+
+pub(in procfs) fn foo() {}
```
tatetian: Need documentation to state explicitly what kind of error is expected and why.

kernel/libs/short-vis-path/src/args.rs
```diff
@@ -0,0 +1,179 @@
...
+            self.replace_verbatim_vis_path(ts);
+            return;
+        }
+        visit_mut::visit_item_mut(self, item);
+    }
+}
+
+impl AddArguments {
+    /// Replace `pub(in subsystem)` by `pub(in crate::to::subsystem)`.
+    pub fn replace_restricted_vis_path(&self, vis: &mut Visibility) {
```
tatetian: Use the narrowest the visibility. I don't there are any users of this method outside this module.

kernel/libs/short-vis-path/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+pub fn add(attr: TokenStream, item: TokenStream) -> TokenStream {
+    if attr.is_empty() {
+        // Do nothing if argument hasn't been provided yet.
+        return item;
+    }
+
+    let mut args = syn::parse_macro_input!(attr as args::AddArguments);
+    let mut file = syn::parse_macro_input!(item as File);
+
+    args.visit_file_mut(&mut file);
```
tatetian, line 25: This macro transforms a given file into a new form with short visibility paths expanded. This function is deterministic and compute-only. So I think it is perfectly fine to write unit tests, rather than integration tests. Sometimes, unit tests could be more light-weight; thus, it could be easier to add more unit tests to cover all sorts of corner cases. You could decompose this crate into several well-defined modules or components to facilitate testing.

book/src/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.md
```diff
@@ -28,6 +28,43 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
...
+pub(in subsystem) fn foo() {}
+
+// Bad
+pub(in crate::a::long::subsystem) fn foo() {}
+```
+
+**Guidelines for applying the attribute:**
+
+1. The subsystem must be a parent or ancestor module of the current module; the
+   Rust compiler does not allow restricting visibility to unrelated modules.
```
tatetian: I don't think that you need to mention this rule. This is not restriction related to the macro.

book/src/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.md
```diff
@@ -28,6 +28,43 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
...
+
+2. When the restricted path is straightforward or used infrequently, keep the
+   original long path to avoid overusing this attribute. Use the
+   attribute if **all** of the following conditions are met:
+   * the submodule depth exceeds 2 levels (i.e., the target path contains at
```
tatetian: Given examples ("Good" and "Bad") to make this rule more concrete.

In addition, this rule should also mention that `pub(super)` is preferred over `pub(in short-name)`.

book/src/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.md
```diff
@@ -28,6 +28,43 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
...
+     least two `::` separators); **and**
+   * the restricted visibility path is used at least 3 times.
+
+Refer to [#3247] for the `short-vis-path` implementation and [#3188] for the design.
```
tatetian: For brevity, only point to the design issue here. From the issue,  people can find the PR, if they want.

book/src/to-contribute/coding-guidelines/for-maintainability/rust-specific/crates-and-modules.md
```diff
@@ -28,6 +28,43 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
 [#2605](https://github.com/asterinas/asterinas/pull/2605#discussion_r2720506912),
 and [#3154](https://github.com/asterinas/asterinas/pull/3154#discussion_r3100905375).
 
+### Restrict subsystem visibility in a short name (`short-vis-path`) {#short-vis-path}
+
+For a kernel subsystem with a long module path, use
+`#![short_vis_path::add(subsystem)]` and `pub(in subsystem)` to restrict an
+item's visibility to that subsystem scope.
```
tatetian: This sentence combines the problem and solution. I suggests dedicating the first sentence (or paragraph) to present the problem, with a concrete example. Then, present the solution in the second sentence (or paragraph).

tatetian (commented):
The PR is in a great shape. I only raised a few minor issues.

The `short-vis-path` macro was originally proposed to handle the long visibility paths in the `procfs`, where most items defined within its sub-modules are only visible/usable to the `procfs` itself. But these items are currently often marked as `pub`, which violates the narrowest-visibility guideline. I hope you could also refactor the `procfs` module with the new macro (some should be marked as `pub(super)`). Doing so would raise the awareness of how the macro can be used to narrow the visibility conveniently.

book/src/to-contribute/coding-guidelines/rust-guidelines/language-items/modules-and-crates.md
```diff
@@ -28,6 +28,43 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
...
+
+// Good
+#![short_vis_path::add(subsystem)]
+pub(in subsystem) fn foo() {}
+
+// Bad
+pub(in crate::a::long::subsystem) fn foo() {}
+```
+
+**Guidelines for applying the attribute:**
```
tatetian: Do not use the word "guidelines" here because we already call "Restrict subsystem visibility in a short name" a guideline.

kernel/libs/short-vis-path/tests/ok-decl-macro-2/src/main.rs
zjp-CN: To keep things simple, I've removed some tests that were written to show that item import or usage is good to work.

kernel/libs/short-vis-path/tests/err-single-ident/src/procfs.rs
zjp-CN: The error reasons were written in `tests.rs`. But I've removed error testcases for simplicity.

kernel/libs/short-vis-path/src/lib.rs
zjp-CN: Done: I've added the both crate and tests README.

ostd/libs/short-vis-path/tests/err_bad_diagnostic/Cargo.toml
zjp-CN: Done: I've add such a guideline in the book.

kernel/libs/short-vis-path/tests/tests.rs
zjp-CN: Need to wait for https://github.com/asterinas/asterinas/pull/3265 to merge.

kernel/src/fs/fs_impls/procfs/template/dir.rs
```diff
@@ -2,6 +2,8 @@
 
 //! Reusable procfs directory inode templates and `readdir` helpers.
 
+#![short_vis_path::add(procfs)]
```
zjp-CN: I think this is unnecessary, because we have restricted the `dir` module [here](https://github.com/asterinas/asterinas/blob/3cde7a6fe3090de7cd94858382bff61c708625b0/kernel/src/fs/fs_impls/procfs/template/mod.rs#L5-L10) by
* making the dir module [private](https://github.com/asterinas/asterinas/blob/3cde7a6fe3090de7cd94858382bff61c708625b0/kernel/src/fs/fs_impls/procfs/template/mod.rs#L27)
* making the item exports [visible to super (procfs)](https://github.com/asterinas/asterinas/blob/3cde7a6fe3090de7cd94858382bff61c708625b0/kernel/src/fs/fs_impls/procfs/template/mod.rs#L5-L10)

So this brings the question: do we really need fine-grained visibility restriction on items in a module when the module has been already restricted? Put it differently, we can write the restriction only once on the module, and then get every thing as expected to have narrow scope.

kernel/libs/short-vis-path/src/lib.rs
zjp-CN: The main reason to make integration tests is to make sure inner attribute really works on the given inputs and file layouts. I use  [macrotest](https://docs.rs/macrotest/latest/macrotest/) to test proc-macro. It constructs a Cargo project behinde the scenes, but makes tested files more compact in the tree. The expanded rs files are also no longer stored.

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,76 @@
...
+            $( expanded.check_contains($target_str); )+
+        }
+    };
+    ($( $name:ident: $($target_str:literal),+ );+ $(;)?) => {
+        $( check! { $name: $($target_str),+ } )+
+    };
+}
+
+// Share entry point `fs/procfs/` to test usages on items and the attribute syntax.
+check! { integration_test:
```
zjp-CN: I'm making this `fs/procfs` to be unit tests. Mark the PR as draft for now.

lrh2000:
image

^^^

I believe that this PR was closed by an accident...

zjp-CN:
Thanks for reopening this. I didn't find a button to do it.

Yeah, I didn't expect it's valid to resolve a PR or a comment in a PR...

StevenJiang1110:
Since ext2 and virtio-fs have been merged into the repository, could this PR take the opportunity to clean up the overly verbose visibility specifiers in these two modules? For example:

```plain
pub(in crate::fs::fs_impls::ext2)
pub(in crate::fs::fs_impls::ext2::inode)
pub(in crate::fs::fs_impls::virtiofs)
```

Thanks!

lrh2000:
> Replace pub(in crate::fs::fs_impls::ext2) by pub(in ext2)
> Replace pub(in crate::fs::fs_impls::ext2::inode) by pub(in inode)
> Replace pub(in crate::fs::fs_impls::virtiofs) by pub(in virtiofs)

Shouldn't you do that in your original `Replace multi-level submodules with short-vis-path in kernel` commit?

In addition, https://github.com/asterinas/asterinas/pull/3049 also adds some `pub(in crate::device::tty::vt)` usages...

zjp-CN:
I added the requested short path changes in commits since https://github.com/asterinas/asterinas/pull/3247/changes/fb6e3d7d415a06f65fe5b561335c44638a9579e3 .

For the third  visibility, I only modified the vis paths in `kernel/src/fs/fs_impls/virtiofs/inode/ops.rs`, leaving the only usage in `metadata.rs` and `page_cache.rs` unchanged, because short path doesn't happen for single occurence.
```
# grep result of `pub(in crate::fs::fs_impls::virtiofs)`
    kernel/src/fs/fs_impls/virtiofs/inode/metadata.rs   ┃195┃pub(in crate::fs::fs_impls::virtiofs) fn metadata_from_attr(
    kernel/src/fs/fs_impls/virtiofs/inode/page_cache.rs ┃159┃    pub(in crate::fs::fs_impls::virtiofs) fn invalidate_whole_page_cache(&self) -> Result<()> {
    kernel/src/fs/fs_impls/virtiofs/inode/ops.rs        ┃ 49┃    pub(in crate::fs::fs_impls::virtiofs) fn cached_read_at(
    kernel/src/fs/fs_impls/virtiofs/inode/ops.rs        ┃ 85┃    pub(in crate::fs::fs_impls::virtiofs) fn direct_read_at(
    kernel/src/fs/fs_impls/virtiofs/inode/ops.rs        ┃134┃    pub(in crate::fs::fs_impls::virtiofs) fn cached_write_at(
    kernel/src/fs/fs_impls/virtiofs/inode/ops.rs        ┃184┃    pub(in crate::fs::fs_impls::virtiofs) fn direct_write_at(
    kernel/src/fs/fs_impls/virtiofs/inode/ops.rs        ┃466┃    pub(in crate::fs::fs_impls::virtiofs) fn readdir(
    kernel/src/fs/fs_impls/virtiofs/inode/ops.rs        ┃623┃    pub(in crate::fs::fs_impls::virtiofs) fn revalidate_attr(
```

zjp-CN:
> Shouldn't you do that in your original `Replace multi-level submodules with short-vis-path in kernel` commit?

Ok, I'll fixup these commits.

---

By the way, the exact commit https://github.com/asterinas/asterinas/pull/3247/changes/ef93f3a386290d3d06686ff75a6d724101ae70ee will be dropped due to a lack of feedback.

kernel/libs/short-vis-path/tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Add a testcase
+
+The entry point of all tests are `tests/expand.rs`, which runs `cargo-expand`
+on direct rs files under `tests/expand` directory, and compares the expected
+expansions with the real ones.
+
+If the test case to be added can share the same module layout with
+`integration_test.rs`, add a submodule in the `expand/fs/procfs` directory and
+append a new matching string in `check! { integration_test }`, with the module
+file listed above the string.
```
tatetian: This file breaks a paragraph with manual line breaks. But these lines do not breaks at sentence or clause boundaries. Adopt semantic line breaks.

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn drop(&mut self) {
+        // Remove the expanded file, because we don't need the file.
+        std::fs::remove_file(&self.file_path).unwrap();
+    }
+}
+
+macro_rules! check {
+    // # Arguments
+    //
+    // * `$name` refers to a file stem under `expand` directory.
```
tatetian: a file "stem"? What does it mean?

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+impl Drop for Expanded {
+    fn drop(&mut self) {
+        // Remove the expanded file, because we don't need the file.
+        std::fs::remove_file(&self.file_path).unwrap();
+    }
+}
+
+macro_rules! check {
```
tatetian: Add a summary line to describe the function of this macro at a high level.

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#![doc = include_str!("../README.md")]
+
+struct Expanded {
```
tatetian: Add doc comments to explain what this struct is supposed to do. Also, figure out a better name; this name is not self-describing at all. See the [`descriptive-names`](https://asterinas.github.io/book/to-contribute/coding-guidelines/general-guidelines/index.html#descriptive-names) coding guideline.

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,72 @@
...
+            let expanded = Expanded::run(stringify!($name));
+            $( expanded.check_contains($target_str); )+
+        }
+    };
+    ($( $name:ident: $($target_str:literal),+ );+ $(;)?) => {
+        $( check! { $name: $($target_str),+ } )+
+    };
+}
+
+// Share entry point `fs/procfs/` to test usages on items and the attribute syntax.
```
tatetian: I don't understand what this line tries to say...

kernel/libs/short-vis-path/tests/expand/file_based_module.rs
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#![feature(proc_macro_hygiene)]
+#![feature(custom_inner_attributes)]
+
+pub mod ancestor {
+    mod child;
```
tatetian: This file does not contain `short_vis_path` at all. So I am very surprised that it tests anything at all. There is a file `ancestor/child.rs`. It seems that somehow the content of `ancestor/child.rs` can be compiled as part of `file_base_module.rs`. But `file_based_module.rs` and `ancestor/` are under the _same_ directory. How come `ancestor/` become a sub-module of `file_based_module`? So my only conclusion is that `file_based_module.rs` is treated like a `lib.rs`.

Our purpose here is to test the macro, not to demonstrate some magic tricks to compile stuff. This entire file organization and naming are unconventional, confusing, and shift the focus of a reader for the wrong reason. Make the test code and structure straightforward.

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,72 @@
...
+check! { integration_test:
+    // override.rs
+    "pub(in crate::procfs) enum Override",
+    // fs.rs
+    "pub(in crate::fs::procfs::fs) fn deepest_wins",
+    // multi_args.rs
+    "pub(in crate::fs) fn foo",
+    "pub(in crate::fs::procfs) fn bar"
```
tatetian: I don't think that this macro is doing us any good. It makes simple tasks obsolete. I find the following form perfectly fine to write and read:

```rust
#[test]
fn integration_test() {
    let expanded = Expanded::run("integration_test");
    expanded.check_contains("pub(in crate::procfs) enum Override");
    expanded.check_contains("pub(in crate::fs::procfs::fs) fn deepest_wins");
    expanded.check_contains("pub(in crate::fs) fn foo");
    expanded.check_contains("pub(in crate::fs::procfs) fn bar");
}
```

The coding guideline advises against over-using macros. See [`macros-as-last-resort`](https://asterinas.github.io/book/to-contribute/coding-guidelines/rust-guidelines/language-items/macros-and-attributes.html#macros-as-last-resort).

kernel/libs/short-vis-path/tests/expand.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        // This function runs `cargo expand`, and generate a local expanded file.
+        macrotest::expand(&src);
+
+        let file_path = format!("{PREFIX_DIR}/{name}.expanded.rs");
+        let content = std::fs::read_to_string(&file_path).unwrap();
+        Expanded { file_path, content }
+    }
+
+    fn check_contains(&self, target_str: &str) {
```
tatetian: Use the verb "assert" (rather than "check") in this name to be consistent with the naming convention of existing APIs (e.g., `assert` and `debug_assert`)

kernel/libs/short-vis-path/tests/README.md
```diff
@@ -0,0 +1,51 @@
+# How to test the crate
+
+## Add a testcase
+
+The entry point of all tests are `tests/expand.rs`, which runs `cargo-expand`
+on direct rs files under `tests/expand` directory, and compares the expected
+expansions with the real ones.
+
+If the test case to be added can share the same module layout with
+`integration_test.rs`, add a submodule in the `expand/fs/procfs` directory and
```
tatetian: You need try harder on naming stuffs. Names such as `integration_test.rs` and `expand/fs/procfs` are not helpful at all.

First, **do not embed "domain knowledge" such as procfs into test cases**. This `short-vis-path` crate is solving a problem that is decoupled from kernel development. One day, we may even extract it to a separate Github repo and publish to crates.io. Test cases should be named with Rust terminology, not OS terminology. Each test case should focus on whether the macro works or not for a particular Rust usage pattern. In this regard, `outer_add_inline_modules.rs` is a good one.

Second, at which logical level or from which angel can `integration_test.rs` be considered an integration test? In Cargo's terminology, every file under `tests/` is considered an integration test. So I have no idea why `tests/expand/integration_test.rs` is singled out and named "integration test".

Last, do you think people who see the name `integration_test.rs` expect this file to include `procfs` as a sub-module? Shouldn't the parent of `procfs` be the `fs` module? These names do not match with one another and are simply confusing.

kernel/libs/short-vis-path/tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+append a new matching string in `check! { integration_test }`, with the module
+file listed above the string.
+
+If the test case has a separate module layout (e.g., inline vs. file-based
+modules), add the submodule in the `expand/ancestor` directory.
```
tatetian: Again, you really need to try harder on naming stuffs.

There is a file called `expand.rs` and there is a directory named `expand`. From these names, one can only get that the `expand.rs` file is related to macro expansion and the `expand/` directory as well. These names do not give a clue as to what are their connections and differences.

At the beginning, I even thought that they are the same module, but this does not what you want actually. In Rust 2018+ Edition, this structure

```
foo.rs          # Include a line: `mod bar;`
foo/
    bar.rs
```

is treated the same as

```
foo/
    mod.rs      # Include a line: `mod bar;`
    bar.rs
```

So you see how confusing one can get when seeing both `expand.rs` and `expand/`?

kernel/libs/short-vis-path/tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Set `SHORT_VIS_PATH_DIR` to run `tests/expand.rs`
+
+`#![short_vis_path::add]` requires a correct base directory, while the tested
+Rust files reside in a non-standard Cargo project structure, so
+`SHORT_VIS_PATH_DIR` must be set to specify the correct file prefix.
```
tatetian: Figure out a way to get rid of this env var. The rationale of having this env var is "a non-standard Cargo project structure". I cannot see any fundamental reason why we can't adopt a standard Cargo project structure for the target source code.

tatetian (commented):
I appreciate that more test cases are added, but this entire `tests/` directory needs an overhaul. Its `README`, file organization, and individual file content are very confusing. See my comments below.

kernel/libs/short-vis-path/tests/README.md
zjp-CN: I've used normal cargo-based project for testing. Also removes `SHORT_VIS_PATH_DIR` from code.

New testing instructions are written at the top of integration_test.rs file.

kernel/libs/short-vis-path/tests/README.md
zjp-CN: Sorry, that's indeed confusing.

I've renamed the test file to `integration_test.rs`, and separated test cases into two crates `case-three-level-of-submodules` and `case-syntax-and-behavior` for clarity, with inner items renamed.

`integration_test.rs` has been used in asterinas:

```bash
asterinas $ fd integration_test.rs
kernel/libs/short-vis-path/tests/integration_test.rs
ostd/libs/padding-struct/tests/integration_test.rs
tools/sctrace/tests/integration_test.rs
```

kernel/libs/short-vis-path/tests/README.md
zjp-CN: Thanks for the feedbacks.

I've simplified the file layouts and module names as follows:

```bash
libs/short-vis-path/tests $ tree --gitignore
.
├── case-syntax-and-behavior
│   ├── Cargo.toml
│   └── src
│       ├── lib.rs
│       ├── test_deepest_module_wins
│       │   └── parent
│       │       └── child
│       │           └── parent
│       │               └── child.rs
│       └── test_override
│           └── parent
│               └── child
│                   └── parent
│                       └── child.rs
├── case-three-level-of-submodules
│   ├── Cargo.toml
│   └── src
│       ├── ancestor
│       │   ├── parent
│       │   │   └── child
│       │   │       └── mod.rs
│       │   └── parent.rs
│       └── main.rs
└── integration_test.rs

16 directories, 9 files
```

kernel/libs/short-vis-path/tests/expand.rs
zjp-CN: Good idea. I've fixed it.

kernel/libs/short-vis-path/tests/expand.rs
zjp-CN: Thanks. Now I've used plain functions and methods for simplicity.

kernel/libs/short-vis-path/tests/expand.rs
zjp-CN: [file_stem](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.file_stem) refers to the non-extension portion of file_name.

This file doesn't exist anymore, so resolve the conversation.

kernel/libs/short-vis-path/tests/expand/file_based_module.rs
zjp-CN: I agree that the module layout is hard to follow. 

I've used normal cargo-based projects as test cases.

kernel/libs/short-vis-path/tests/expand.rs
zjp-CN: [file_stem](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.file_stem) refers to the non-extension portion of file_name.


This file doesn't exist anymore, so resolve the conversation.

kernel/libs/short-vis-path/tests/expand.rs
zjp-CN: The macro is removed. Resolve this.

kernel/libs/short-vis-path/tests/expand.rs
zjp-CN: This is renamed to `Testcase` , storing a directory path to case crate and the output of `cargo expand`.

kernel/libs/short-vis-path/tests/README.md
zjp-CN: This file is removed, and testing approach has changed.

How to adding a new test case is documented in the `short-vis-path/tests/integration_test.rs` header docstring.

kernel/libs/short-vis-path/src/lib.rs
zjp-CN: I discard the `macrotest` dev dependency, and now the crate exhibites a straightforward layout by having  `test_` prefixed modules:

```
.
├── case-syntax-and-behavior
│   ├── Cargo.toml
│   └── src
│       ├── lib.rs
│       ├── test_deepest_module_wins
│       │   └── parent
│       │       └── child
│       │           └── parent
│       │               └── child.rs
│       ├── test_mod_rs_flavor
│       │   └── child
│       │       └── mod.rs
│       ├── test_multiple_arguments
│       │   ├── one_ident_and_one_override.rs
│       │   └── two_idents.rs
│       └── test_override
│           └── parent
│               └── child
│                   └── parent
│                       └── child.rs
└── integration_test.rs
```

Resolve this conversation.

kernel/libs/short-vis-path/tests/err-decl-macro-absolute-path/src/main.rs
zjp-CN: The testing way has changed. Now
* each `test_` prefixed module serves as a macro behavior to be tested against.
* stdout of `cargo expand` is captured without writing to a local file. 
 
Resolve this.

kernel/src/fs/fs_impls/procfs/template/dir.rs
zjp-CN: The file change has been discarded. The change was initially as a discussion to my concerns, but it turns out that they don't seem to be a problem for project maintainers.

Resolve this.

zjp-CN:
@boterinas rerun

kernel/libs/short-vis-path/tests/tests.rs
zjp-CN: Resolve this: `cargo-expand` has been available in the docker image.

kernel/libs/short-vis-path/README.md
```diff
@@ -0,0 +1,68 @@
+# short-vis-path
+
+A proc-macro attribute for using short visibility paths in Rust.
```
tatetian: This summary is not quite helpful as it is like saying `short-vis-path` is about short visibility paths. But what is short visibility paths anyway?

I suggest revising the text in the following way:

> A proc-macro attribute for shortening the module path in a visibility modifier (`pub (in crate::a::very::deep::module)`) when the module hierarchy is deep.

tatetian:
@zjp-CN `![#![short_vis_path::add(short_name)] ]` obscuring all Rust compiler error messages for the entire module, as shown by https://github.com/os-checker/short-vis-path/issues/1, is a dealbreaker. We need to figure out how to overcome this problem or we have to drop this PR.

zjp-CN:
> We need to figure out how to overcome this problem

I opened the [bug report ](https://github.com/rust-lang/rust/issues/157881) to rustc. Two contributors are trying to solve the rustc bug:

* https://github.com/rust-lang/rust/pull/158046
* https://github.com/rust-lang/rust/pull/157597

I've locally tested short-vis-path against compiler with the fix from the first link, diagnostics are good:

```rust
# Current
short-vis-path/tests/err_bad_diagnostic ✗ cargo b
error[E0599]: no method named `detach_tracer_with` found for reference `&PosixThread` in the current scope
 --> err_bad_diagnostic/src/main.rs:6:1
  |
6 | pub mod ptrace;
  | ^^^^^^^^^^^^^^^
  |
help: there is a method `detach_tracer` with a similar name
  |
6 - pub mod ptrace;
6 + detach_tracer
  |
```
```rust
# Fixed
short-vis-path/tests/err_bad_diagnostic ✗ cargo +stage1 b
error[E0599]: no method named `detach_tracer_with` found for reference `&PosixThread` in the current scope
  --> err_bad_diagnostic/src/ptrace.rs:10:12
   |
10 |     thread.detach_tracer_with()
   |            ^^^^^^^^^^^^^^^^^^
   |
help: there is a method `detach_tracer` with a similar name
   |
10 -     thread.detach_tracer_with()
10 +     thread.detach_tracer()
   |
```

So we need to wait  for the upstream to merge it, and bump the toolchain in Asterinas.

tatetian:
> So we need to wait for the upstream to merge it, and bump the toolchain in Asterinas.

Thank you for follow-up with the latest move in the Rust community! Glad to hear that the Rust community is working this problem.

zjp-CN:
The compilation error message problem is [solved](https://github.com/rust-lang/rust/pull/158046) in the current toolchain. 

I've rebased the commits. The PR is ready for review.

kernel/libs/short-vis-path/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Background
+
+When working with deep module hierarchies, writing `pub(in crate::very::long::path)` repeatedly can be cumbersome. This library allows you to use short identifiers in visibility restrictions, which are automatically expanded to full paths at compile time. See [Asterinas#3188](https://github.com/asterinas/asterinas/issues/3188) for the design.
+
+### Basic Example
+
+In a file at `src/fs/procfs.rs`:
```
tatetian, line 11: We have referenced procfs as a main user of `short-vis-path` in documentation and tests, but this PR never touches procfs. I think that there are many Rust files under `procfs/` should have benefited from this macro.

For example, the following code should have used the syntax `pub(in procfs)`:

https://github.com/asterinas/asterinas/blob/c4368ac4f6c2b05eafcf363f76ab1d66e8bf1085/kernel/src/fs/fs_impls/procfs/pid/task/cmdline.rs#L15-L18

How about adding a commit that modifies the visibilities modifiers in procfs code to use `pub(in procfs)`.

book/src/to-contribute/coding-guidelines/for-maintainability/rust-specific/crates-and-modules.md
```diff
@@ -65,6 +65,72 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
 [#2605](https://github.com/asterinas/asterinas/pull/2605#discussion_r2720506912),
 and [#3154](https://github.com/asterinas/asterinas/pull/3154#discussion_r3100905375).
 
+### Restrict subsystem visibility with a short name (`short-vis-path`) {#short-vis-path}
+
+ It's a good practice to restrict the visibility of a subsystem item to the
+ subsystem itself using the [`pub(in path)`][pub-in-path] syntax. However,
+ writing `pub(in crate::very::long::subsystem)` becomes verbose, reduces
```
tatetian: First of all, adopt the semantic line break convention.

Second, refer to the `narrow-visibility` guideline as a justification to long visibility modifiers like `pub(in crate::very::long::subsystem)`. I suggest rewriting the first and second paragraphs this way:

````md
To conform with the [`narrow-visibility`](path/to/this/rule#narrow-visibility) guideline,
a common pattern is to expose Rust items inside a subsystem only
to the subsystem itself.
But when the subsystem sits deep in the module hierarchy, 
we have to write long visibility modifiers
like `pub(in crate::a::very::deep::subsystem)`,
which is tedious for writers and unfriendly for readers.

Instead, you can use the [`short_vis_path`] macro
to create a short name for the long full path of a subsystem:

```rust
// In kernel/src/a/very/deep/subsystem/lib.rs

// Good
#![short_vis_path::add(subsystem)]

pub(in subsystem) fn struct Foo {}

// Bad: violating the narrow-visibility guideline.
pub struct Foo {}

// Bad: the visibility path is too long to read.
pub(in crate::a::very::deep::subsystem) struct Foo {}
```
````

File: book/src/to-contribute/coding-guidelines/for-maintainability/rust-specific/crates-and-modules.md [hunk 32]
```diff
@@ -65,6 +65,72 @@ PR [#2951](https://github.com/asterinas/asterinas/pull/2951),
...
+// Bad: the visibility is not restricted, violating the narrow-visibility guideline.
+pub struct ProcDir {}
+
+// Bad: the visibility path is too long to read.
+pub(in crate::fs::fs_impls::procfs) struct ProcDir {}
+```
+
+When the restricted path is straightforward or used infrequently, keep the
+original long path to avoid overusing this attribute. Only follow this
+guideline when **all** of the following three conditions are met:
```
tatetian: These three conditions are quite simple. Providing examples make the text too verbose. Make the writing more concise.

tatetian (commented):
This PR looks in a great shape! I appreciate your effort.

Initially, when we designed the `short-vis-path` macro, the motivation was very strong because the `kernel/src/` crate is quite deep. Yet with #3601, we will be able to move many subsystems out of `kernel/src/`. Such subsystems can then use `pub(crate)` in its own crate.

With this said, I believe the `short-vis-path` macro is still useful to the relatively large knot of `aster-core`.

kernel/libs/short-vis-path/README.md
zjp-CN: > adding a commit that modifies the visibilities modifiers in procfs code to use `pub(in procfs)`

Done in https://github.com/asterinas/asterinas/pull/3247/changes/33718cc585a364228d0802b298e6b8ea713f879f .

zjp-CN:
@boterinas rerun

tatetian (approved):

PR #3248, opened 2026-05-19
Install `cargo-expand` and bump docker version to 0.17.2-20260519

zjp-CN:
See https://github.com/asterinas/asterinas/pull/3247#discussion_r3264959235


tatetian (approved):

zjp-CN:
Sorry, the argument is [wrong](https://github.com/asterinas/asterinas/actions/runs/26263934163/job/77303189954):

```
0.085 error: invalid value 'cargo-expand@v1.0.122' for '[CRATE[@<VER>]]...': the version provided, `v1.0.122` is not a valid SemVer requirement
0.086 
0.086 help: try changing the version to `1.0.122`
```

I need to remove the `v` prefix from the version number. 😅

Should I bump the Docker version again when fixing this bug, or leave it as-is and have someone manually trigger the image build CI?

I expected modifying the `DOCKER_IMAGE_VERSION` file to trigger the image build workflow in the current PR, but it doesn't seem to, instead the workflow starts only on pushing to the main branch.

PR #3249, opened 2026-05-20
 Add `VmarHandle` to count `Vmar` users 

lrh2000:
The purpose of this PR is to prepare reverse mappings of `Vmo` to `Vmar` (`Weak<Vmar>`).

We should be able to unmap pages from `Vmo`. For example, when a file is truncated, all the pages in the truncated range need to be unmapped.

At the very least, we will need access to the `VmSpace` that contains the mapped pages. However, that is not enough. Unmapping pages directly from a `VmSpace` will cause the RSS counters in the `Vmar` to become out of sync:
https://github.com/asterinas/asterinas/blob/09760e3c201821c63502c93f2b10e80b50a27280/kernel/src/vm/vmar/vmar_impls/mod.rs#L40-L41

We can resolve this problem by obtaining a `Vmar` from the reverse mappings. However, this is problematic due to:
https://github.com/asterinas/asterinas/blob/09760e3c201821c63502c93f2b10e80b50a27280/kernel/src/process/process_vm/mod.rs#L163-L166

The purpose of this PR is to relax the limitation. This is achieved by tracking the number of VMAR users separately (e.g. the number of POSIX threads that use that VMAR), rather than relying on the VMAR's reference count.
```rust
struct Vmar {
    // ...
    num_users: AtomicUsize,
}
```

This PR also introduces a strongly typed `VmarHandle` which represents a `Vmar` user. Each POSIX thread holds this handle in its `ThreadLocal`.
```rust
// vmar/handle.rs:

struct VmarHandle(Arc<Vmar>);

impl Drop for VmarHandle {
    fn drop(&mut self) {
        self.0.unmark_shared();
    }
}

// thread_local.rs:

struct ThreadLocal {
    // ...

    vmar: RefCell<Option<VmarHandle>>,
}
```

---

There are some other alternative solutions:
 - Find a way to add RSS counters to `VmSpace`. This is difficult because `VmSpace` lives in OSTD.
 - Wrap the RSS counters in another `Arc` and store them in the reverse mappings. I dislike this solution because it misuses `Arc`.

### DIFF
--- a/kernel/src/vm/vmar/vmar_impls/mod.rs
+++ b/kernel/src/vm/vmar/vmar_impls/mod.rs
@@ -9,7 +9,11 @@ mod query;
 mod remap;
 mod unmap;
 
-use core::{array, ops::Range};
+use core::{
+    array,
+    ops::Range,
+    sync::atomic::{AtomicUsize, Ordering},
+};
 
 use align_ext::AlignExt;
 use aster_util::per_cpu_counter::PerCpuCounter;
@@ -37,15 +41,19 @@ pub struct Vmar {
     inner: RwMutex<VmarInner>,
     /// The attached `VmSpace`
     vm_space: Arc<VmSpace>,
-    /// The RSS counters.
+    /// The RSS counters
     rss_counters: [PerCpuCounter; NUM_RSS_COUNTERS],
     /// The process VM
     process_vm: ProcessVm,
+    /// The number of handles that this `Vmar` has (see [`super::VmarHandle`])
+    num_handles: AtomicUsize,
 }
 
 impl Vmar {
     /// Creates a new VMAR.
-    pub fn new(process_vm: ProcessVm) -> Arc<Self> {
+    ///
+    /// This method should only be invoked by [`super::VmarHandle`].
+    pub(super) fn new(process_vm: ProcessVm) -> Arc<Self> {
         let inner = VmarInner::new();
         let vm_space = VmSpace::new();
         let rss_counters = array::from_fn(|_| PerCpuCounter::new());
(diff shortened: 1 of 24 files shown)


kernel/src/vm/vmar/vmar_impls/mod.rs
```diff
@@ -37,15 +41,17 @@ pub struct Vmar {
     inner: RwMutex<VmarInner>,
     /// The attached `VmSpace`
     vm_space: Arc<VmSpace>,
-    /// The RSS counters.
+    /// The RSS counters
     rss_counters: [PerCpuCounter; NUM_RSS_COUNTERS],
     /// The process VM
     process_vm: ProcessVm,
+    /// The number of threads that use this VMAR
+    num_users: AtomicUsize,
```
tatetian: If I understand it correctly, the only purpose of this counter is to serve the `VmarHandle`. And `Vmar` and `VmarHandle` are within the `vmar` module. So I think we can say that the two concepts are inevitably coupled with one another. So we call simply call this field `num_handles`, which would be more specific than `num_users`:

```rust
    /// The number of handles that this `Vmar` has. See [`super::VmarHandle`].
    num_handles: AtomicUsize,
```

By introducing the concept of handle, the three newly-added methods can reference this concept and enjoy more self-describing names:

```rust
    /// Returns whether this VMAR has multiple handles.
    pub fn have_multiple_handles(&self) -> bool { ... }
    // or
    /// Returns the number of handles.
    ///
    /// See [`super::VmarHandle`].
    pub fn num_handles(&self) -> usize { ... }

    /// Increases the number of handles.
    ///
    /// Should only be invoked by [`super::VmarHandle`].
    pub(super) fn inc_num_handles(&self) { ... }

    /// Decreases the number of handles.
    ///
    /// Should only be invoked by [`super::VmarHandle`].
    pub(super) fn dec_num_handles(&self) { ... }
```

tatetian (commented):
Looks mostly great. Just a few nits.

tatetian (approved):

PR #3253, opened 2026-05-20
Use local vsock endpoint for QEMU-to-QGS quote generation

Hsy-Intel:
QEMU uses `quote-generation-socket` as the VMM-side endpoint for connecting to QGS, not as a guest AF_VSOCK destination.

The TDX quote flow does not send the request from guest userspace through a guest `AF_VSOCK` socket. Instead, the guest submits the quote request through the TDX attestation path, QEMU reads the request from shared memory, forwards it to QGS, and writes the response back to shared memory.

In Linux vsock addressing, CID 1 is the local endpoint and CID 2 is the host endpoint. Because `quote-generation-socket` is used by QEMU itself when it opens the connection to QGS, the CID must be interpreted in the vsock context of the QEMU process.


tatetian (approved):

PR #3254, opened 2026-05-20
Add (kernel-side) patches, verified usage docs, and CI for GDB

vvvvsv:
Currently, running GDB on the main branch requires the two patches in #3236 https://github.com/asterinas/asterinas/pull/3236/changes/3f893512085d0c3acb7a5d23afea6282151e060e:

> The first patch emulates the default state of the x86 debug registers, so it can correctly respond when a tracer reads the tracee’s debug registers via PTRACE_PEEKUSER. In the current main branch, the tracee’s x86 debug registers are never actually modified, so they always remain at their default values.
> 
> The second patch suppresses the GDB warning "Error disabling address space randomization". This patch will be removed once personality(ADDR_NO_RANDOMIZE) is supported.
> 

However, GDB is a dependency of too many packages, including some huge ones such as WebKitGTK. Rebuilding them would take a looooooong time. [Even 2 hours is not enough](https://github.com/asterinas/asterinas/actions/runs/26145123047/job/76898790056?pr=3236), so it is difficult to do in CI. Therefore, it would be better not to patch GDB at the application side, but instead move the patches to the kernel side.

For the first patch, supporting GDB debugging of multithreaded programs will require correctly reading and writing the x86 debug registers. I already have a draft.

For the second patch, @fxySoleil already has a draft.

However, the complete implementation of both patches may not be ready before the 18.0 release. So we may have to leave a few `FIXME`s here. I did not add SCML for these dummy implementations, because we should indicate to `sctrace` users that these features are not supported yet.

Additionally, this PR adds verified usage docs and CI for GDB.

Required for #2323.


book/src/distro/popular-applications/development-tools/README.md
```diff
@@ -427,6 +427,48 @@ ninja -C build
...
+
+# Inspect the current stop
+(gdb) backtrace
+(gdb) frame 0
+(gdb) print x
+(gdb) info registers rip rsp
+
+# Modify the register and memory
+(gdb) set var x = 1000
+(gdb) print heap_value
```
tatetian: I am not familiar wit GDB. It looks that `x` and `heap_value` are variables defined in `hello.c`. I suggest giving the C file in the doc so that people can reproduce the results by following the exact steps.

test/nixos/tests/development-tools/src/main.rs
```diff
@@ -304,6 +304,88 @@ fn meson_build(nixos_shell: &mut Session) -> Result<(), Error> {
...
+// ============================================================================
+// Debugging Tools - GDB
+// ============================================================================
+
+#[nixos_test]
+fn gdb_debug(nixos_shell: &mut Session) -> Result<(), Error> {
+    nixos_shell.run_cmd(concat!(
+        r#"printf '%b' '"#,
+        r#"#include <stdio.h>\n"#,
+        r#"#include <stdlib.h>\n"#,
```
tatetian: Save the C source code in a file so that both the test code and the doc can directly include the source code without duplication.

test/nixos/tests/development-tools/src/main.rs
```diff
@@ -304,6 +304,88 @@ fn meson_build(nixos_shell: &mut Session) -> Result<(), Error> {
...
+        r#"    printf("memory check passed: %d\\n", *heap_value);\n"#,
+        r#"    free(heap_value);\n"#,
+        r#"    return 0;\n"#,
+        r#"}\n' > /tmp/gdb_sample.c"#,
+    ))?;
+    nixos_shell.run_cmd("gcc -g -O0 /tmp/gdb_sample.c -o /tmp/gdb_sample")?;
+
+    nixos_shell.run_cmd_and_expect(
+        concat!(
+            r#"gdb -q -batch /tmp/gdb_sample "#,
```
tatetian: Defining the string in the `concat` form causes too much visual noise. I guess the following, more readable form can work?

```
let text = r#"gdb -q -batch /tmp/gdb_sample \
    -ex 'set pagination off' \
    -ex 'break hello_world' \
    ...
    -ex 'continue' > /tmp/gdb.out 2>&1 && echo GDB_OK"#;
```

tatetian (commented):
Thanks for the quick workaround!

vvvvsv:
@tatetian Thanks for the suggestions! I have moved both the C code sample and the GDB command script into `book/src/distro/popular-applications/development-tools/`, and now both the CI test and the docs reference the files directly.

tatetian (approved):

PR #3256, opened 2026-05-21
Fix a page cache bug that leaks uninitialized meory to userspace (and more)

lrh2000:
There is a bug in the page cache that will leak uninitialized content to user space:
https://github.com/asterinas/asterinas/blob/ee133428d96a0a2f726b883fed1aa5e092c0e684/kernel/src/vm/page_cache/vmo/mod.rs#L600-L603

This is a partial write. Even if the commit mode can be set to `CommitMode::Overwrite`, we cannot set the page status to up-to-date.

I added a regression test for this case. This test also revealed a minor bug in the ext2 file system. I have fixed it together to make the test pass.

---

The third commit removes the `CommittedPage`. The main reason I did this is that, despite its good intentions to avoid lock costs, it messes up the lock orders among the pages.

If there are three pages in the page cache — Page 1, Page 2, and Page 3 — and I need to take the locks for all of those pages (which I plan to do when implementing the reverse mappings), I want to take the locks in order from first to last so that the lock order is consistent.
```rust
let pages: Vec<CachePage> = collect_pages();
let locked_pages = pages.into_iter().map(CachePage::lock).collect::<Vec<_>>();
```

But this design is incompatible with the current implementation. The problem is that `collect_pages` can report something like:
```rust
let mut pages = Vec::new();
collect_pages(&mut pages);
// pages = [CommittedPage::Unlocked(1), CommittedPage::Unlocked(2), CommittedPage::Locked(3)];
```

Then, when we write sequentially to the three pages, we will try to lock the first page while holding the lock on the third page.

Removing `CommittedPage` has no impact on performance. `CommittedPage::Locked` is only constructed here.
https://github.com/asterinas/asterinas/blob/ee133428d96a0a2f726b883fed1aa5e092c0e684/kernel/src/vm/page_cache/vmo/mod.rs#L796-L797

In this scenario, we don't need to lock the page. This may make the code even faster.


tatetian:

kernel/src/vm/page_cache/vmo/mod.rs
```diff
@@ -766,40 +754,38 @@ impl<'a> BackedVmo<'a> {
...
             // The page will be completely overwritten, no need to read.
-            Ok(locked_page.into())
+            Ok(uninit_page)
         } else {
-            let new_page = locked_page.clone();
             // Read the page from the backend storage.
-            self.backend.read_page(page_idx, locked_page)?;
-            Ok(new_page.into())
+            uninit_page.ensure_init(|locked_page| self.backend.read_page(page_idx, locked_page))?;
+            Ok(uninit_page)
```
cchanging, line 788: I'd like to explain why we previously returned with the lock held here: For a page that is about to be overwritten, if we don't keep it locked after returning from commit, there may be other commit operations trying to acquire it between the time this page is committed and when it gets overwritten — for example, during a page fault. This would trigger initialization logic and result in an unnecessary I/O operation. 

Alternatively, we could introduce an additional commit mode that keeps all returned pages locked to avoid lock ordering issues?

Other commits LGTM!

lrh2000: > if we don't keep it locked after returning from commit, there may be other commit operations trying to acquire it between the time this page is committed and when it gets overwritten — for example, during a page fault. This would trigger initialization logic and result in an unnecessary I/O operation.

This is a race condition that does not affect the correctness of the kernel or the userspace program. It only affects performance. As a race condition, it happens rarely. So I don't think it is necessary to optimize for it..

> Alternatively, we could introduce an additional commit mode that keeps all returned pages locked to avoid lock ordering issues?

That's a different matter, though. We can't always get the lock under the `XArray` guard. Therefore, we will need to retry. It is possible, but it also introduces some complexity. Thus, I don't know if it's worth the effort.

Also, https://github.com/asterinas/asterinas/issues/3257 is an issue with a different lock order that I don't know how to resolve. So, the related logic may need to be changed anyway. However, I don't know what the final result should look like.

cchanging: > As a race condition, it happens rarely. So I don't think it is necessary to optimize for it..

This makes sense. It seems that there is indeed no mandatory requirement for commit to return a locked page.

cchanging (approved):

cqs21 (approved):

PR #3259, opened 2026-05-22
Install virtiofsd in Docker image

li041:
Install the Ubuntu noble `virtiofsd` package in the Asterinas development Docker image so virtio-fs regression tests can start the vhost-user daemon from the CI container.

The package comes from Ubuntu universe (`Source: rust-virtiofsd`, `Origin: Ubuntu`) and installs the daemon under `/usr/libexec/virtiofsd`; the image adds a `/usr/local/bin/virtiofsd` symlink so tests can invoke it directly.

Also bump `DOCKER_IMAGE_VERSION` to trigger publishing the updated image.

Validation:
- `apt-cache show virtiofsd` confirms package source/origin
- `/usr/libexec/virtiofsd --version` reports `virtiofsd 1.10.0`
- `git diff --check`


tools/docker/Dockerfile
```diff
@@ -83,8 +83,10 @@ RUN apt update && apt-get install -y --no-install-recommends \
     strace \
     sudo \
     unzip \
+    virtiofsd \
     vim \
     zip
+RUN ln -s /usr/libexec/virtiofsd /usr/local/bin/virtiofsd
```
tatetian: Give this line a separate paragraph. And add a comment line to explain why this symbolic link should be created. Which software requires this symbolic link?

li041: The purpose of this symbolic link is that `/usr/libexec` is typically not included in the default `PATH`, while `/usr/local/bin` is globally accessible from the command line and scripts. 

I will remove it, and call `virtiofsd` with an absolute path.

tatetian (approved):

PR #3260, opened 2026-05-22
Move `orig_syscall_ret` to `ThreadLocal`

lrh2000:
It seems that we can move `orig_syscall_ret` from `PosixThread` to `ThreadLocal`. We can use `Cell<Option<usize>>` in `ThreadLocal`, which seems more reasonable.

@vvvvsv Maybe help to confirm that this aligns with your intention? Thanks!


vvvvsv:
Thanks! This aligns with the original intention, and it does make more sense to keep `orig_syscall_ret` in `ThreadLocal`.

tatetian (approved):

tatetian:
@lrh2000 Could you fix the conflict?

PR #3261, opened 2026-05-22
Add cmdline to disable legacy com1 for x86

liulog:
I tried to deploy Asterinas on a physical machine (Thinkpad X13 Gen4) and encountered some problems. I eventually located the issue to the UART component, specifically the `uart.flush()` function, which was stuck in an infinite loop. This is because on modern laptops, COM1 0x3F8 almost no longer corresponds to the physical serial port.

I added a control implementation for cmdline to support disabling the use of this serial port.

compile：
```
make kernel BOOT_METHOD=qemu-direct BOOT_PROTOCOL=linux-efi-handover64 CONSOLE=tty0
```

cmdline in grub.cfg：
```
...
menuentry "Boot Asterinas Kernel" {
    echo "Loading Asterinas Kernel..."
    linux /asterinas/aster-kernel-osdk-bin SHELL=/bin/sh LOGNAME=root HOME=/ USER=root PATH=/bin:/benchmark init=/init console=tty0 i8042.exist ostd.legacy_com1=off -- sh -l
    echo "Loading Initramfs..."
    initrd /asterinas/aster-initramfs.cpio.gz
    echo "Booting Asterinas..."
}
...
```

90af0dd80fc5486cd4f79171a5a32b9e


ostd/src/arch/x86/serial.rs
```diff
@@ -84,5 +80,26 @@ impl Ns16550aAccess for SerialAccess {
...
+    //    enabled, it is assumed to exist and be accessible via the I/O registers.
+    //    (FIXME: This should be confirmed by checking the ACPI table.)
+    // 2. `reserve_io_port_range` guarantees exclusive ownership of the I/O registers.
+    let access = unsafe { SerialAccess::new(0x3F8) };
+    let mut serial = Ns16550aUart::new(access);
+    serial.init();
+    SERIAL_PORT.call_once(move || SpinLock::new(serial));
+}
+
+fn should_initialize_legacy_com1() -> bool {
```
lrh2000:
```suggestion
fn is_legacy_com1_disabled() -> bool {
```

To match the argument name.

We should also invert the returned boolean if we rename the method.

ostd/src/arch/x86/serial.rs
```diff
@@ -6,21 +6,17 @@ use spin::Once;
...
-    Once::initialized(SpinLock::new(Ns16550aUart::new(
-        // SAFETY:
-        // 1. It is assumed that the serial port exists and can be accessed via the I/O registers.
-        //    (FIXME: This needs to be confirmed by checking the ACPI table or using kernel
-        //    parameters to obtain early information for building the early console.)
-        // 2. `reserve_io_port_range` guarantees exclusive ownership of the I/O registers.
-        unsafe { SerialAccess::new(0x3F8) },
-    )));
+pub static SERIAL_PORT: Once<SpinLock<Ns16550aUart<SerialAccess>, LocalIrqDisabled>> = Once::new();
 reserve_io_port_range!(0x3F8..0x400);
```
lrh2000: I think this line should be moved below the `init()` method so that the several 0x3F8 occurrences are grouped together.

ostd/src/arch/x86/serial.rs
```diff
@@ -84,5 +80,26 @@ impl Ns16550aAccess for SerialAccess {
...
+        return;
+    }
+    // SAFETY:
+    // 1. The legacy COM1 serial port at 0x3F8 can be disabled via the command line. If it is
+    //    enabled, it is assumed to exist and be accessible via the I/O registers.
+    //    (FIXME: This should be confirmed by checking the ACPI table.)
+    // 2. `reserve_io_port_range` guarantees exclusive ownership of the I/O registers.
+    let access = unsafe { SerialAccess::new(0x3F8) };
+    let mut serial = Ns16550aUart::new(access);
+    serial.init();
```
lrh2000: Let's move these statements into the `call_once` closure, so that "exclusive ownership of the I/O registers" is always true (otherwise if you call `init()` multiple times, the ownership isn't really exclusive.

ostd/src/arch/x86/serial.rs
```diff
@@ -6,21 +6,17 @@ use spin::Once;
 use x86_64::instructions::port::ReadWriteAccess;
 
 use crate::{
+    boot::EARLY_INFO,
     console::uart_ns16650a::{Ns16550aAccess, Ns16550aRegister, Ns16550aUart},
     io::{IoPort, reserve_io_port_range},
     sync::{LocalIrqDisabled, SpinLock},
 };
 
+/// Kernel command line argument that disables the legacy COM1 serial port.
```
lrh2000:
```suggestion
/// A kernel command line argument that disables the legacy COM1 serial port.
```

lrh2000: The commit title should be rewritten to something like `Add cmdline to disable [..]` to match the existing conventions.

lrh2000 (commented):
Thank you for diagnosing this issue. Debugging early boot issues on commodity hardware is difficult. I appreciate your efforts.

> const DISABLE_LEGACY_COM1_ARG: &str = "ostd.legacy_com1=off";

I think the current form is acceptable as we already have a `ostd.log_level` cmdline in OSTD.
https://github.com/asterinas/asterinas/blob/b166a03c79ac0b7962efad25043ff5921563e6e7/ostd/src/log/mod.rs#L140

It makes me think a little more, though. We can inject an early cmdline parser (`const fn`) from `aster-kernel`, which will make these arguments consistent with the Linux one.

```rust
#[ostd::early_cmdline_parser]
const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
    // Parse "loglevel=X"
    // Parse "earlycon=uart8250,io,0x3f8,115200"
    // ...
    //
    // They're Linux command lines so the parser lives in `aster-kernel`.

    EarlyCmdline {
        loglevel,
        // ...
    }
}
```

This technique is similar to the following:
https://github.com/asterinas/asterinas/blob/b166a03c79ac0b7962efad25043ff5921563e6e7/ostd/libs/ostd-macros/src/lib.rs#L239-L265

It will be called very early in the boot stage, even before the heap is initialized. Therefore, it must be a const function. I think a `const fn` should be safe to call.

I don't think we need to block this PR because of that. We can accept the PR and do the refactoring later.

liulog:
Thank you very much for your review and suggestions. I will take your advice into consideration and make appropriate revisions.

tatetian:
Thanks for raising the bug.

But it seems to me that this PR only fixes a surface problem. The deeper problem is how to determine the meaning of I/O ports (such as the one we are discussing about) on different hardware/platforms.

Furthermore, I don't think it is a good idea to introduce an OSTD-specific parameter when Linux already accepts a different parameter (`earlycon`) for the same task.

I am not an expert on selecting consoles, so I discussed with AI. And here is a conclusion that makes sense to me.

---

**The hang is at a lower layer than the knob.** Two cooperating bugs in `Ns16550aUart` / `serial::init()` cause it:

1. `serial::init()` configures the UART and starts `send()`-ing before checking that anything is at `0x3F8`.
2. `Ns16550aUart::send()` polls THRE with `while !contains(THRE) { spin_loop() }` and no timeout. On a chipset where the unmapped port reads `0x00`, THRE is never set, and that loop never returns.

Linux doesn't hit this because its `8250` driver does a scratch-register round-trip presence test before registering a port, and never TX-spins during detection. The two protections compose: probing screens out dead addresses, and bounded waits ensure that even a misprobe can't wedge boot. Either one alone would have prevented the ThinkPad hang.

**How Linux actually "knows" about `0x3F8`.** It mostly doesn't — `0x3F8` / IRQ 4 is a 1981 IBM PC convention, not something the OS discovers. Linux layers two strategies on top:

- **Probe with presence test.** `autoconfig()` in `drivers/tty/serial/8250/8250_port.c` writes the scratch register, reads it back, and only registers the port if it round-trips. Detection never spins on TX.
- **Enumerate from firmware:**
  - **ACPI SPCR** — a dedicated ACPI table describing the console UART (interface type, base address as a GAS, IRQ, baud). It's a static table — no AML interpretation needed — so it can be parsed extremely early; Linux uses it to bring up `earlycon` before the driver model exists.
  - **ACPI DSDT `PNP0501`** — a legacy COM port that genuinely exists is enumerated as a PnP device with `_CRS`; its absence is itself a signal.
  - **Device tree** on non-ACPI platforms — `/chosen/stdout-path` plus a `compatible = "ns16550a"` node.
- **Cmdline overrides on top, not instead of:** `console=ttyS0,…`, `earlycon=uart8250,io,0x3f8`, `8250.nr_uarts=0`. User escape hatches, not the discovery mechanism.

**What I'd suggest for Asterinas**, from cheapest to deepest — each step lands independently:

1. **Fix the hang at the source.** Add a scratch-register presence test in `Ns16550aUart` (or `SerialAccess`) and bound the THRE poll in `send()` with a small iteration cap that bails instead of spinning forever. This alone removes the hang class on every platform, with no cmdline knob and no firmware enumeration.
2. **SPCR-based discovery on x86.** The `acpi` crate already in `ostd/Cargo.toml` (`acpi = "=5.2.0"`) ships an `Spcr` table type with `base_address()`, `interface_type()`, `irq()`, and `baud_rate()`. It slots in beside the existing `find_table::<Fadt>()` / `find_table::<Mcfg>()` calls in `arch/x86/kernel/acpi/mod.rs`.

**On `ostd.legacy_com1=off`.** If we still want a user-visible escape hatch after step 1, I'd prefer to reuse Linux's existing vocabulary — `console=`, `earlycon=`, `8250.nr_uarts=` — rather than introduce a new `ostd.`-prefixed parameter for a concern Linux has already named. Adding `ostd.legacy_com1=off` cements a workaround into our kernel command-line ABI in exactly the place we'd later want to deprecate in favor of proper enumeration.

---

So my preference for this PR: keep the lazy-init restructure (it's a clean improvement and aligns x86 with the loongarch pattern), drop the cmdline plumbing, and replace it with the bounded-probe fix in `Ns16550aUart`. Steps 2 become a separate follow-up.

**`ostd.legacy_com1=` vs `earlycon=`:** Here is a dilemma. Using `ostd.legacy_com1` would create a parameter overlapping with Linux's. Yet, OSTD's recognizing `earlycon=` would hardcode Linux's knowledge into OSTD, which is designed to be OS agnostic.

lrh2000:
> I am not an expert on selecting consoles, so I discussed with AI. And here is a conclusion that makes sense to me.

AI reviews like this are quite confusing.

> **ACPI SPCR** — a dedicated ACPI table describing the console UART (interface type, base address as a GAS, IRQ, baud). It's a static table — no AML interpretation needed — so it can be parsed extremely early; Linux uses it to bring up `earlycon` before the driver model exists.

We're talking about the **early console** here. At this point, we don't even have a heap allocator. At this point, requiring ACPI to be parsed can be a non-trivial change.

> **Cmdline overrides on top, not instead of:** `console=ttyS0,…`, `earlycon=uart8250,io,0x3f8`, `8250.nr_uarts=0`. User escape hatches, not the discovery mechanism.

`earlycon=` **is** a method of discovering an earlycon device. It is an alternative to the ACPI SPCR tables. Supporting the latter is fine, but supporting the former is also correct.

> 1. **Fix the hang at the source.** Add a scratch-register presence test in `Ns16550aUart` (or `SerialAccess`) and bound the THRE poll in `send()` with a small iteration cap that bails instead of spinning forever. This alone removes the hang class on every platform, with no cmdline knob and no firmware enumeration.

The loop in question is `Uart::flush`, not `Ns16550aUart::send`.

Both loops may need to be bounded. However, the device should not be accessed if it is not present. So I think adding a command line to prevent the device from being initialized is a correct decision.

> **On `ostd.legacy_com1=off`.** If we still want a user-visible escape hatch after step 1, I'd prefer to reuse Linux's existing vocabulary — `console=`, `earlycon=`, `8250.nr_uarts=` — rather than introduce a new `ostd.`-prefixed parameter for a concern Linux has already named. Adding `ostd.legacy_com1=off` cements a workaround into our kernel command-line ABI in exactly the place we'd later want to deprecate in favor of proper enumeration.
> [ .. ]
> **`ostd.legacy_com1=` vs `earlycon=`:** Here is a dilemma. Using `ostd.legacy_com1` would create a parameter overlapping with Linux's. Yet, OSTD's recognizing `earlycon=` would hardcode Linux's knowledge into OSTD, which is designed to be OS agnostic.

https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441 clearly shows how to move the early command line parsing logic outside of OSTD. For now, I think we should keep the `ostd.*` and follow the aforementioned roadmap.

> keep the lazy-init restructure (it's a clean improvement and aligns x86 with the loongarch pattern),

I don't understand what this wants to say.

> So my preference for this PR: keep the lazy-init restructure (it's a clean improvement and aligns x86 with the loongarch pattern), drop the cmdline plumbing, and replace it with the bounded-probe fix in `Ns16550aUart`. Steps 2 become a separate follow-up.

I think the current form is fine. Creating bounded device loops can be a separate task. But a device should not be accessed if it is not present. So we should have a way to prevent it from being accessed. Using the command line is one of the approaches here.

lrh2000:
> the bounded-probe fix

Since it's a simple task, if you feel it is really necessary, we could combine it with this PR (in a separate commit) (for `Uart::flush` and `Ns16550aUart::send`).

> But it seems to me that this PR only fixes a surface problem.

However, I think the original changes are good and I don't think they "only fix a surface problem".

lrh2000:
> * **Cmdline overrides on top, not instead of:** `console=ttyS0,…`, `earlycon=uart8250,io,0x3f8`, `8250.nr_uarts=0`. User escape hatches, not the discovery mechanism.

This is factually incorrect. Linux logic is here.
https://elixir.bootlin.com/linux/v7.0.9/source/drivers/tty/serial/earlycon.c#L230

 1. ACPI SPDR isn't used unless `earlycon` is specified without an argument.
 2. When `earlycon=` is specified with an argument, the device is registered with the specified PIO/MMIO region.

So `earlycon` **is** a discovery mechanism. In addition, it is a prerequisite to ACPI SPCR.

> At this point, we don't even have a heap allocator. 

But I should also correct my comment here: This is not a problem as we also need to count the number of CPUs before heap is initialized.

liulog:
First, adding ACPI SPCR support requires extra design work and introduces more complexity to the code.
Second, based on my discussion with AI, SPCR tables are usually missing on laptops and desktops anyway. These devices use the framebuffer for output.

Right now, `early_print` in ostd and the `uart` component in the kernel are actually sharing the exact same serial port (0x3F8). Because this I/O port is defined inside ostd, using command-line arguments that start with ostd makes total sense to me. 

Looking ahead, if Asterinas targets physical server deployments instead of just running inside VMs, implementing SPCR support will definitely be a required task on our future roadmap.

Considering the current practical situation, I'm leaning towards using cmdline to provide a flexible and compatible way to disable COM1 when needed, just as @lrh2000 pointed out.

liulog:
I have updated the commit title and modified the relevant code according to the suggestions above.

`linux /asterinas/aster-kernel-osdk-bin SHELL=/bin/sh LOGNAME=root HOME=/ USER=root PATH=/bin:/benchmark init=/init console=tty0 i8042.exist ostd.legacy_com1=off -- sh -l`

Boot behavior analysis on ThinkPad X13 gen4 with/without specific kernel parameters. 

With the relevant command-line arguments added, the boot process on the ThinkPad behaves as follows:
1. GRUB loads the kernel.
2. The kernel's EFI Stub executes.
3. The kernel clears the screen.
4. Subsequent output is correctly handled via the framebuffer.

Without these arguments, the symptoms are as follows:
1. GRUB loads the kernel and the EFI Stub executes.
2. The kernel clears the framebuffer.
3. After that, the screen goes completely blank with no further output, which corresponds to the infinite loop in uart.flush() mentioned earlier.

img1
img2

tatetian:
Let me clarify my point.

Consider two possible solutions to fix this hang problem on your ThinkPad. One solution requires modifying the kernel parameter manually, the other does not. I think we can agree that a solution that can automatically avoid this problem is better than manually asking the kernel, right?

So can we devise such a simple solution to do just that and not to rely on ACPI? I think we can——with two measures.

## The first measure: testing the existence of UART

If a COM port is not mapped to a UART-driven chip at all on a PC, then there is no point in creating a `UartConsole` for it. If we do so, the kernel won't call `UartConsole::flush`.

In Linux, the 8250 driver has [an existence test](https://github.com/torvalds/linux/blob/4cbfe4502e3d4bda48eb4b83dfad8d7da3b22e90/drivers/tty/serial/8250/8250_port.c#L1090-L1127) for the UART silicon. 

The basic idea of Linux's UART existence test is to write to the `UART_IER` port of UART and see if valid values that are written to it can be read back as expected. It does two writes: `serial_out(up, UART_IER, 0);` and `serial_out(up, UART_IER, UART_IER_ALL_INTR);`. The first write enables no interrupts, while the second enables all the basic interrupts portable across 8250-family chips. If the port is driven by any UART chip at all, then we should be able to read back the written values.

I think we should and can adopt the same simple and effective strategy.

## The second measure: making I/O a bounded operation

The existence test could give false result or a serial device connected to the UART-driven COM port could go malfunction. So it is desirable to avoid unbounded loop that waits for I/O completion. Linux adopts this strategy. See [`wait_for_lsr`](https://github.com/torvalds/linux/blob/master/drivers/tty/serial/8250/8250_port.c#L1955-L1981) as an example.

## Conclusion

I suppose that the first measure is sufficient to make your hang-up problem on ThinkPad go away. The second measure would make the code even more robust.

This is why I am against the idea of adding a kernel argument, which only workarounds the deeper problems (enabling COM ports blindly and waiting for unbounded amount of time), as described shown above. Whenever possible, we should aim for a fix, not a workaround.

lrh2000:
> [an existence test](https://github.com/torvalds/linux/blob/4cbfe4502e3d4bda48eb4b83dfad8d7da3b22e90/drivers/tty/serial/8250/8250_port.c#L1090-L1127)

This is good. The problem is that writing to an unknown port will [have possibilities to hang the device](https://wiki.osdev.org/I8042_PS/2_Controller#Step_2:_Determine_if_the_PS/2_Controller_Exists):

>> Before you touch the PS/2 controller at all, you should determine if it exists. On some systems (e.g. 80x86 Apple machines) it doesn't exist and any attempt to touch it can cause a system crash.

So while it works in this case and is a reasonable addition, it does not solve the root problem..

If we can have a chance to align our command line behavior to the Linux one. It does seem a better solution to me.

In other words, all of these measures are definitely good to have. However, they are not the ultimate solution that can prevent a non-existent device from being accessed.

> only workarounds the deeper problems

The measures work around a deeper problem, **a nonexistent device should not be accessible**, either.

---

The problem with the command line is that we cannot reasonably align its behavior with that of Linux. If you say it's a major concern, I agree. I don't like the `ostd.*` parsers either.

If it's difficult to reach a consensus on this, we can implement the command line fixes in the future. It does seem like a proper fix. However, at the moment, we don't have the technique implemented to specify it in `kernel/`, which is where it should be to match the Linux behavior.

Let's add the `earlycon` command line from kernel after implementing an early command line parser injection in future PRs. I assume you don't object it, right? We can align the behavior with Linux. There are no early consoles if `earlycon` is not specified. Therefore, we don't need to worry about touching unknown I/O ports.

tatetian:
> The problem with the command line is that we cannot reasonably align its behavior with that of Linux. If you say it's a major concern, I agree. I don't like the `ostd.*` parsers either.

Yes, that is another reason that I don't like the current form of this PR.

> If it's difficult to reach a consensus on this, we can implement the command line fixes in the future. It does seem like a proper fix. However, at the moment, we don't have the technique implemented to specify it in `kernel/`, which is where it should be to match the Linux behavior.

Yes, we need to invent "the technique" of . But it is not immediately clear to me how to do this properly. All existing injection techniques only take effect _after_ OSTD is initialized. At this point, all OSTD public APIs are safe and sound.

But this new injected kernel parameter parser must be invoked _before_ the early console itself. The latter one is enabled in a very stage of the OSTD, when OSTD is not fully initialized yet. So we cannot rule out the possibility that the arbitrary Rust code in the injected kernel parameter parser cause soundness problems by invoking the OSTD APIs of some uninitialized OSTD modules.

> Let's add the `earlycon` command line from kernel after implementing an early command line parser injection in future PRs. I assume you don't object it, right?

Sure. But how should we proceed with this PR? Since adding a kernel parameter properly needs more preparatory steps, I think we should implement the two measures suggested by my previous comment; they are simple and do not rely on unimplemented infra.

> There are no early consoles if `earlycon` is not specified.

Thanks for bringing this up! This is where Asterinas is unaligned with Linux. Asterinas always enables an early console, yet with Linux, the user must opt in the early console with a kernel parameter. Linux's strategy does make sense for two reasons: (1) the kernel cannot reliably detect or enable a simple device for the early console in a very early phase and (2) most users do not care about losing some log messages in the early phase. 

Therefore, we should changing Asterinas' behavior to opt _in_ the early console with a new kernel parameter, rather than opt _out_.

<details>
<summary>
<b>About `earlycon=`</b>
</summary>

# Appendix: About `earlycon=`

I ask the following question to my coding agent: "Does Linux enable the early console by default, when no `earlycon=` argument is given?" Here is the answer.

**Short answer: no, not by default — except on ACPI systems where firmware exposes an SPCR table.** SPCR is the one mechanism that auto-enables an earlycon without anything on the cmdline. Everything else is opt-in.

## Why the question needs unpacking

There are actually three distinct early-output mechanisms in Linux, each gated differently. They're often conflated under "early console":

1. **`CONFIG_EARLY_PRINTK` / `earlyprintk=`** — the older x86-specific mechanism. Outputs via VGA text mode, 8250 serial, or USB-DBGP. **Not automatic.** Has to be on the cmdline.

2. **`CONFIG_SERIAL_EARLYCON` / `earlycon=`** — the generic, cross-architecture framework you're reading about. Polled-mode, lives on the kernel stack with no driver model. **Not automatic from the cmdline alone.** `earlycon=uart8250,io,0x3f8,115200` is explicit; bare `earlycon` (no args) uses DT's `/chosen/stdout-path`.

3. **`CONFIG_ACPI_SPCR_TABLE` / SPCR auto-earlycon** — when the kernel is built with this kconfig and firmware provides a populated SPCR (Serial Port Console Redirection) table, the early arch code calls `acpi_parse_spcr(enable_earlycon=true, …)` and **automatically registers an earlycon from SPCR's base address, interface type, and baud rate** — no cmdline needed. This is the only path that fires on its own.

The SPCR path lives in `drivers/acpi/spcr.c::acpi_parse_spcr()`; the architecture call sites that pass `enable_earlycon = true` are `arch/arm64/kernel/acpi.c::acpi_boot_table_init()` and the equivalents on x86 / RISC-V (where the arch supports it). The function reads `interface_type`, validates it against a known list (Full 16550, PL011, BCM2835, NVIDIA-16550, …), and calls `setup_earlycon()` with the right driver.

## What you get with no `earlycon=` on the cmdline

- **ACPI system with valid SPCR** (ARM64 servers are the archetype; modern x86 servers increasingly do too): **earlycon comes up automatically.** The kernel can print before the full console subsystem initializes.
- **ACPI system without SPCR**: no early console. The kernel boots silently from bootloader handoff until the regular `printk` console is registered around `register_console()` — well after early arch setup, memory init, and IRQ setup.
- **Device-tree system**: DT's `/chosen/stdout-path` is **not consulted** unless the bare `earlycon` keyword is on the cmdline. A DT-only board with no `earlycon` boots silently early.
- **Legacy x86 with neither SPCR nor `earlyprintk=`/`earlycon=`**: silent. This is the most common desktop/laptop case — and it is also why a desktop Linux on the ThinkPad X13 Gen4 doesn't hang. The kernel just doesn't try to use any UART until the regular serial driver runs, by which time the ACPI-PnP enumeration has reported "no `PNP0501` device, no `ttyS0`."

## Why Linux made this choice — the same lesson as #3261

Two reasons that interlock:

1. **Blindly touching 0x3F8 in early boot is unsafe.** That's the exact trap that bit Asterinas on the ThinkPad. If the kernel unconditionally probed the legacy COM range during early boot, every legacy-COM-less machine would risk the same hang. SPCR exists precisely to give firmware an authoritative way to say "yes, there is a console UART here, at *this* address, with *these* parameters" — and Linux trusts SPCR enough to auto-enable an earlycon from it. But it pointedly doesn't auto-probe `0x3F8` just because tradition says COM1 lives there.
2. **Early console is debug verbosity, not a correctness requirement.** Silent boot is fine for production; loud boot is for debugging. The default is silent, and verbosity is opt-in — either by the user (`earlycon=`) or by firmware (SPCR).

Notice what Linux explicitly does *not* do: feed the legacy `old_serial_port[]` table into earlycon. That table is consumed only later by `serial8250` proper, which goes through `autoconfig()` with its scratch/IER existence tests before actually registering a `ttyS*`. The early-console code path doesn't get the chance to wedge on a phantom legacy port because **the early-console code path is never given a phantom legacy port to wedge on** — it only gets ports that came from firmware (SPCR) or from the user (cmdline).

## The mirror for OSTD

The Linux pattern relevant to the Asterinas design discussion: **default to silent until something authoritative says "here is the console UART."** "Authoritative" here is one of:

- The user (`earlycon=…` / `console=…` on cmdline),
- The firmware (SPCR auto-detect),
- Or — on DT platforms — the user *acknowledging* DT (`earlycon` bare keyword). DT presence alone isn't enough; the user has to opt in.

That's the design knob I'd point at if you wanted to make OSTD's behavior look principled rather than legacy-driven: don't unconditionally init the legacy COM, parse SPCR, and if SPCR is silent, *be silent*. Cmdline override is the user's explicit acknowledgment that they accept the risk of poking a legacy address.
</details>

lrh2000:
> Yes, we need to invent "the technique" of . But it is not immediately clear to me how to do this properly. All existing injection techniques only take effect _after_ OSTD is initialized. At this point, all OSTD public APIs are safe and sound.

Let me ask again. Did you read https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441?

> But this new injected kernel parameter parser must be invoked _before_ the early console itself. The latter one is enabled in a very stage of the OSTD, when OSTD is not fully initialized yet. So we cannot rule out the possibility that the arbitrary Rust code in the injected kernel parameter parser cause soundness problems by invoking the OSTD APIs of some uninitialized OSTD modules.

Did you read https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441? A `const fn` won't cause soundness issues. Or you can prove me wrong?

lrh2000:
> Sure. But how should we proceed with this PR?

Either of the followings:
 1. Falling back to your proposals are acceptable to me, though I don't think your proposals can fix the root cause: **a nonexistent device should not be accessed**. I think there are workarounds for that, but the defensive measures are worth adding given that the device may be buggy even if it does exist.
 2. Add a command line like `ostd.disable_legacy_com1`, `ostd.earlycon`, or whatever. Defer the implementation of https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441 to a future PR. So we have ways to **prevent a nonexistence device from being accessed**.
 3. Or implement https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441 in this PR first and correctly implement the `earlycon` semantics in both OSTD and `aster-kernel` (or `aster-cmdline`?).

I think @liulog can make the final decision. If the second option is unacceptable to you, we can choose the first or third option instead. I think the third option is the right fix, but the first option is worth considering too.

tatetian:
> > Yes, we need to invent "the technique" of . But it is not immediately clear to me how to do this properly. All existing injection techniques only take effect _after_ OSTD is initialized. At this point, all OSTD public APIs are safe and sound.
> 
> Let me ask again. Did you read [#3261 (review)](https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441)?
> 
> > But this new injected kernel parameter parser must be invoked _before_ the early console itself. The latter one is enabled in a very stage of the OSTD, when OSTD is not fully initialized yet. So we cannot rule out the possibility that the arbitrary Rust code in the injected kernel parameter parser cause soundness problems by invoking the OSTD APIs of some uninitialized OSTD modules.
> 
> Did you read [#3261 (review)](https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441)? A `const fn` won't cause soundness issues. Or you can prove me wrong?

Somehow I missed the const function-based proposal. This approach should work.

> I think the third option is the right fix, but the first option is worth considering too.

Agree. Both the first and third options are meaningful improvements towards the right direction.

@liulog You could pick either, preferably both ❤️

liulog:
Before I begin, I'd like to confirm the following with you:

The following is required:

- Change the command-line argument to `earlycon` to indicate whether the SERIAL_PORT instance will be created in OSTD. This currently only indicates the legacy COM1 at 0x3F8. To prevent OSTD from accessing non-existent serial ports, `earlycon` needs to be parsed in OSTD. Currently, only one `earlycon` is provided for this indication; richer parameters, such as `earlycon=uartxxx,xxx`, are not yet available.

- Perform an existence test for SERIAL_PORT. Since it's defined and initialized in OSTD, this test should also be implemented in OSTD. 


More comprehensive `earlycon` configurations should be provided as future support options. (later PRs ?)

The schematic logic in OSTD is as follows:

```rust
// ostd/src/arch/x86/serial.rs
pub static SERIAL_PORT: Once<SpinLock<Ns16550aUart<SerialAccess>, LocalIrqDisabled>> = Once::new();

pub(crate) fn init() {
    // If earlycon is not declared in cmdline, skip initialization.
    ...
    // Check if 0x3F8 legacy COM1 exists.
    ...
    // SERIAL_PORT.call_once().
    ...
}
reserve_io_port_range!(0x3F8..0x400);
```

Does it meet the above requirements?

tatetian:
> Does it meet the above requirements?

Yes, I think so.

lrh2000:
> * `earlycon` needs to be parsed in OSTD.

The third option means to implement the aforementioned technique of injecting an early command line parser (https://github.com/asterinas/asterinas/pull/3261#pullrequestreview-4349635441) and parsing this option in `kernel/comps/cmdline` (or `kernel/src/`...? But I suppose `comps/cmdline` seems to be a better location

If the parser lives in OSTD, shouldn't the argument be called `ostd.earlycon`, in line with the existing `ostd.log_level`?
 - If so, this is the second option, not the third option (see https://github.com/asterinas/asterinas/pull/3261#issuecomment-4531231886).
 - The Linux-compatible ones should be named `earlycon`, `earlycon=uartXXX`, and `loglevel=8`, but it is not very related here as we already have a `ostd.log_level` command line argument.

lrh2000:
> The third option means to implement the aforementioned technique of injecting an early command line parser

If you have some interest to implement it, here are some more guidance:
 - Add a new marco in `ostd/libs/ostd-macros`: `ostd::early_cmdline_parser`, which can be annotated on a constant function and provides a `no_manged` symbol `__early_cmdline_parser`. See `global_heap_allocator_slot_map` for an example.
 - Add a default `fn __early_cmdline_parser(cmdline: &str) -> EarlyCmdline` implementation in `ostd/src/boot/mod.rs`. It should be marked as `#[linkage = "weak"]` since it is the default implementation and can be overridden by the kernel.
   - The default implementation shouldn't parse anything and just return a default `EarlyCmdline`, perhaps:
```rust
EarlyCmdline {
    log_level: log::Level::Debug,
    has_early_console: true,
}
```
 - Add a `pub(crate) fn parse_early_cmdline() -> EarlyCmdline` method in `ostd/src/boot/mod.rs`. It should get the early command line from `EARLY_INFO.unwrap().cmdline` and parse it with `__early_cmdline_parser(cmdline)`.
 - Call the `fn parse_early_cmdline()` method in `ostd::init()`. Perhaps the final code can look like:
```rust
let early_cmdline = boot::parse_early_cmdline();

arch::serial::init(&early_cmdline);

log::init(&early_cmdline);
```
So that `arch::serial` can know if there is an early console and `log::init` can know the log level.
 - Use the `ostd::early_cmdline_parser` to define an early command line parser in `aster-cmdline` (`kernel/comps/cmdline/src/early.rs`). It should return:
```rust
EarlyCmdline {
    log_level: /* parse it from something like `loglevel=8` */,
    has_early_console: /* whether  `earlycon` exists */,
}
```
 - We also need to prevent the parsed early command lines from being passed to the init process (as the init process's `argv`). We can record the known early params as an array (e.g., `EARLY_PARAMS: &["earlycon", "loglevel"]` and filter them out in `dispatch_params()`: https://github.com/asterinas/asterinas/blob/2e600838f35d36a6866bd2d75ed8573bfbf47a92/kernel/comps/cmdline/src/dispatch.rs#L139-L143

liulog:
You're right, I need to take a closer look at `kernel/comps/cmdline`. 

And it may indicate that an initial version is to be provided, just like to what you mentioned in the previous review.

> ```rust
> #[ostd::early_cmdline_parser]
> const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
>    // Parse "loglevel=X"
>    // Parse "earlycon=uart8250,io,0x3f8,115200"
>    // ...
>    //
>    // They're Linux command lines so the parser lives in `aster-kernel`.
>
>    EarlyCmdline {
>        loglevel,
>        // ...
>    }
>}
>```

It might only detect earlycon and loglevel, and perform some Linux-like parsing on earlycon.

I saw the relevant function in the Linux code.
https://github.com/torvalds/linux/blob/master/drivers/tty/serial/serial_core.c#L2108

However, there are some exceptions, namely those that are not standard earlycon, such as riscv, which has a valid cmdline with `earlycon=sbi`.
https://github.com/torvalds/linux/blob/master/drivers/tty/serial/earlycon-riscv-sbi.c#L52

---

My idea is that earlycon could be defined as a quadruple of `(name, type, addr, options)` for now? More complex parsing features could be left for the future? Currently, there are no more use cases, or to be simpler, for the current version, only one interface should be provided, and earlycon should only use a `Boolean` value to indicate whether earlycon is enabled.

liulog:
> > The third option means to implement the aforementioned technique of injecting an early command line parser第三种方案是实现前述注入早期命令行解析器的技术
> 
> If you have some interest to implement it, here are some more guidance:如果你有兴趣实施，这里有一些更多指导：
> 
> * Add a new marco in `ostd/libs/ostd-macros`: `ostd::early_cmdline_parser`, which can be annotated on a constant function and provides a `no_manged` symbol `__early_cmdline_parser`. See `global_heap_allocator_slot_map` for an example.在 `ostd/libs/ostd-macros` 中添加一个新的 marco：`ostd：：early_cmdline_parser`，它可以在常量函数上做注释，并提供`一个 no_manged` 符号 `__early_cmdline_parser`。请参见 `global_heap_allocator_slot_map` 示例。
> * Add a default `fn __early_cmdline_parser(cmdline: &str) -> EarlyCmdline` implementation in `ostd/src/boot/mod.rs`. It should be marked as `#[linkage = "weak"]` since it is the default implementation and can be overridden by the kernel.
>   在 `ostd/src/boot/mod.rs` 中添加默认`的 fn 实现__early_cmdline_parser（cmdline： &str） -> EarlyCmdline` 实现。应标记为 `#[linkage = “weak”]`，因为它是默认实现，内核可以覆盖。
>   * The default implementation shouldn't parse anything and just return a default `EarlyCmdline`, perhaps:默认实现不应该解析任何内容，可能只返回一个默认`的 EarlyCmdline`：
> 
> ```rust
> EarlyCmdline {
>     log_level: log::Level::Debug,
>     has_early_console: true,
> }
> ```
> 
> * Add a `pub(crate) fn parse_early_cmdline() -> EarlyCmdline` method in `ostd/src/boot/mod.rs`. It should get the early command line from `EARLY_INFO.unwrap().cmdline` and parse it with `__early_cmdline_parser(cmdline)`.在 `ostd/src/boot/mod.rs` 中添加`一个 pub（crate） fn parse_early_cmdline（） -> EarlyCmdline` 方法。它应该从 `EARLY_INFO.unwrap（）.cmdline` 获取早期命令行，并用 `__early_cmdline_parser（cmdline）` 解析。
> * Call the `fn parse_early_cmdline()` method in `ostd::init()`. Perhaps the final code can look like:在 `ostd：：init（）` 中调用 `fn parse_early_cmdline（）` 方法。也许最终的代码可以是：
> 
> ```rust
> let early_cmdline = boot::parse_early_cmdline();
> 
> arch::serial::init(&early_cmdline);
> 
> log::init(&early_cmdline);
> ```
> 
> So that `arch::serial` can know if there is an early console and `log::init` can know the log level.这样 arch`：：serial` 就能知道是否有早期控制台，`log：：init` 也能知道日志级别。
> 
> * Use the `ostd::early_cmdline_parser` to define an early command line parser in `aster-cmdline` (`kernel/comps/cmdline/src/early.rs`). It should return:使用 `ostd：：early_cmdline_parser` 在 `aster-cmdline` 中定义早期命令行解析器（`kernel/comps/cmdline/src/early.rs`）。它应该会回归：
> 
> ```rust
> EarlyCmdline {
>     log_level: /* parse it from something like `loglevel=8` */,
>     has_early_console: /* whether  `earlycon` exists */,
> }
> ```
> 
> * We also need to prevent the parsed early command lines from being passed to the init process (as the init process's `argv`). We can record the known early params as an array (e.g., `EARLY_PARAMS: &["earlycon", "loglevel"]` and filter them out in `dispatch_params()`: 我们还需要防止解析后的早期命令行传递给初始化进程（作为初始化进程的 `argv`）。我们可以将已知的早期参数记录为数组（例如，`EARLY_PARAMS： &[“earlycon”， “loglevel”]`，并在 `dispatch_params（）` 中过滤掉：https://github.com/asterinas/asterinas/blob/2e600838f35d36a6866bd2d75ed8573bfbf47a92/kernel/comps/cmdline/src/dispatch.rs#L139-L143

Sorry, I was editing and didn't check the email or refresh the page. I've seen it now, thank you so much for your guidance!

liulog:
The cmdline has been modified to `earlycon loglevel=N`, where earlycon indicates whether to enable early UART logging and determines whether to initialize the serial port. `loglevel=N` replaces the original `ostd.log_level`, similar to the Linux format, and here indicates the log level filter in ostd.

I tried adding an existence test to COM1, but the complexity exceeded my expectations. I tried adding the existence test, but the laptop didn't even show any screen refresh, meaning the OS crashed much earlier (previously I could indirectly debug using the kernel's framebuffer, but now the kernel might not have been entered, so I removed the existence test). Therefore, I think the existence test can be added in the future, possibly by an expert more familiar with x86. I added a TODO line to serial.rs.

Therefore, the current commit only contains changes to the command line. It works fine on my laptop.

5c6d90794d574c7145b9f5c56baddad8

tatetian:
> I think the existence test can be added in the future, possibly by an expert more familiar with x86. I added a TODO line to serial.rs.

That's fine. Thanks for the try.

Before reviewing the changes in detail, could you ensure that all CI tests pass? I marked this PR as "Draft" and feel free to turn it to "Ready for Review" once all CI tests pass.

liulog:
I'm having a CI issue that I can't resolve.

I need to modify `ostd-macros` and `ostd`.

The `Publish OSDK and OSTD`  test pulls `ostd-macros` from crates.io instead of my local one, causing related errors. I haven't come up with a suitable solution.

The error is below:
```
error[E0432]: unresolved import `ostd_macros::early_cmdline_parser`
  --> src/lib.rs:60:5
   |
60 |     early_cmdline_parser, global_frame_allocator, global_heap_allocator,
   |     ^^^^^^^^^^^^^^^^^^^^ no `early_cmdline_parser` in the root
```

I need to confirm this with you.

@tatetian @lrh2000

lrh2000:
> The `Publish OSDK and OSTD` test pulls `ostd-macros` from crates.io instead of my local one, causing related errors. I haven't come up with a suitable solution.

This problem does not need to be solved in this PR. Once it's merged, we'll need to [bump the versions](https://asterinas.github.io/book/to-contribute/version-bump.html) (e.g., tagging and publishing 0.18.1) to resolve the issue. For now, though, ignore it so that the normal review process can continue.

(So you can mark this PR as ready for review if there are no other issues...)

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,66 @@ console=ttyS0
...
+  `earlycon=uart8250,io,0x3f8`; Asterinas does not parse such suffixes yet. When
+  `earlycon` is present, the kernel uses a built-in NS16550A setup (on x86,
+  COM1 at `0x3F8`).
+- When omitted, the early UART console stays disabled.
+
+### `loglevel`
+
+Set the maximum log severity.
+
+Unlike Linux console printk (which prints only messages with level **strictly
```
lrh2000: Use [Semantic Line Breaks](https://sembr.org/) in Markdown files. I think it is the existing convention. Check other files for examples, e.g.,
https://github.com/asterinas/asterinas/blob/944778d7f9fd3821d4df1eb63b78ce1a456090c0/book/src/kernel/linux-compatibility/README.md?plain=1#L7-L19

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,66 @@ console=ttyS0
...
+less than** `console_loglevel`), Asterinas filtering is **inclusive**: `loglevel=N`
```
lrh2000: Is there a reason why we must differ from Linux here?

~~I think we just need to subtract one from the numeric value to match the Linux behavior. Am I missing anything?~~

**EDIT:** Oh, after checking the implementation I found we should store a `LevelFilter` in `EarlyCmdline` instead. `LevelFilter::from_u8` does exactly what we need here:
https://github.com/asterinas/asterinas/blob/944778d7f9fd3821d4df1eb63b78ce1a456090c0/ostd/src/log/level.rs#L154-L160

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,66 @@ console=ttyS0
...
 Example:
 ```text
-ostd.log_level=error
+loglevel=3
+```
+
+Notes:
+- Parsed during early boot and not forwarded to `init`.
+- When omitted, invalid, or greater than `7`, the default is `7` (debug).
+- Replaces the former Asterinas-specific `ostd.log_level` parameter.
```
lrh2000: In fact, since Asterinas is still in the early stages and has no stability guarantee, I don't think we need to include history in the persistent documents.

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,66 @@ console=ttyS0
...
-- `emerg`
-- `alert`
-- `crit`
-- `error`
-- `warn` (alias: `warning`)
-- `notice`
-- `info`
-- `debug`
+Notes:
+- Parsed during early boot (before the heap is available) and not forwarded to
```
lrh2000: > Parsed during early boot (before the heap is available)

I doubt we need to mention implementation details here.

The way it is implemented does not really matter. This is more of a user-facing document.

(It is necessary to mention this, in the API doc comments for the `early_cmdline_parser` macro. See my comments below.)

> not forwarded to `init`

This applies to all the arguments here. From a user-facing perspective, nothing is special. So, I find the words odd.

> OSTD prints its own log messages on this UART; the kernel prints early boot logs through OSTD on the same UART.

 1. I am not sure if we need to mention OSTD.
 2. After the kernel boots, messages printed by OSTD via log macros will go to the real console instead of the UART. So strictly speaking, "OSTD prints its own log messages on this UART" is just incorrect.

I believe we can explain the behavior without going through implementation concepts...

> [ .. ]
>
> ### `loglevel`
>
> [ .. ]
>
> Parsed during early boot and not forwarded to `init`.

My comment also applies to some other places of the docs. We could say that the log level applies to early messages (assuming that we have defined the term, "early messages", in the `earlycon` documentation). This is user-facing, but when it is parsed isn't.

kernel/comps/cmdline/src/dispatch.rs
```diff
@@ -204,6 +209,13 @@ mod tests {
         assert_eq!(args.argv[0].to_bytes(), b"unknown_key");
     }
 
+    #[ktest]
+    fn early_params_not_forwarded_to_init() {
```
lrh2000:
```suggestion
    fn early_params_not_forwarded() {
```

Be consistent with other test names:
```rust
    fn registered_params_not_forwarded() {
```

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,114 @@
...
+//! OSTD calls the function registered with [`ostd::early_cmdline_parser`] during
+//! bootstrap, before serial and logging are initialized. The full cmdline parser
+//! in the parent crate runs later, once the heap and component system are ready.
+
+use ostd::{boot::EarlyCmdline, log::Level};
+
+/// Kernel command-line keys consumed by [`early_cmdline_parser`].
+///
+/// These are stripped during dispatch so they are not forwarded to init.
+pub(super) const EARLY_PARAMS: &[&str] = &["earlycon", "loglevel"];
```
lrh2000:
```suggestion
const EARLY_PARAMS: &[&str] = &["earlycon", "loglevel"];
```

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,114 @@
...
+/// # Recognized tokens
+///
+/// - `earlycon` — enables the early UART console (exact token match).
+/// - `loglevel=N` — sets the maximum OSTD log level (`N` in `0..=7`).
+///
+/// Other tokens are ignored. When `loglevel` is absent, the default is `7`
+/// (debug). When `earlycon` is absent, the early console stays disabled.
+#[ostd::early_cmdline_parser]
+const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
+    let bytes = cmdline.as_bytes();
```
lrh2000: I think we can move this below since this is only used below:
```rust
    let mut has_early_console = false;
    let mut log_level_num = 7;

    let bytes = cmdline.as_bytes();
    let mut index = 0;
    while index < bytes.len() {
```

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+        index += 1;
+    }
+    true
+}
+
+/// Returns whether `bytes[start..end]` equals `target` byte-for-byte.
+///
+/// Requires equal lengths, so tokens such as `earlyconxxxx` do not match `earlycon`.
+const fn match_token(bytes: &[u8], start: usize, end: usize, target: &[u8]) -> bool {
```
lrh2000: Define `match_token` before `starts_with_at`. This is more consistent to the call order (`match_token` calls `starts_with_at`).

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+    starts_with_at(bytes, start, end, target)
+}
+
+/// Parses a decimal log level from the suffix of a `loglevel=` token.
+///
+/// Reads consecutive ASCII digits starting at `start` and stops at the first
+/// non-digit or at `end`. Returns `7` when no digits are present or the value
+/// exceeds `7`, consistent with [`Level::from_u8`].
+const fn parse_loglevel_at(bytes: &[u8], mut start: usize, end: usize) -> u8 {
```
lrh2000: This method is named `parse_loglevel_at`. So I think it should just return `log::Level`.

**EDIT:** Should be `log::Filter` now. See comments above...

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -264,6 +264,35 @@ pub fn global_heap_allocator_slot_map(_attr: TokenStream, item: TokenStream) ->
...
+    let fn_name = &item.sig.ident;
+
+    // Reject if the input is not a `const fn`.
+    assert!(
+        item.sig.constness.is_some(),
+        "the annotated function must be `const`"
+    );
+
+    quote!(
+        /// SAFETY: The name does not collide with other symbols.
```
lrh2000:
```suggestion
        // SAFETY: The name does not collide with other symbols.
```

ostd/src/lib.rs
```diff
@@ -82,15 +82,16 @@ unsafe fn init() {
     // and after memory regions are initialized.
     unsafe { mm::frame::allocator::init_early_allocator() };
 
+    let early_cmdline = boot::parse_early_cmdline();
```
lrh2000, line 85: Add a blank line below, since `early_cmdline` will be used in multiple code blocks.

ostd/src/boot/mod.rs
```diff
@@ -113,6 +115,28 @@ pub(crate) fn init_after_heap() {
...
+/// The early command line arguments.
+pub struct EarlyCmdline {
+    /// The log level.
+    pub log_level: Level,
+    /// Whether to enable the early console.
+    pub has_early_console: bool,
+}
+
+#[linkage = "weak"]
+#[unsafe(no_mangle)]
```
lrh2000, line 151:
```rust
// SAFETY: The name does not collide with other symbols.
#[unsafe(no_mangle)]
```

tools/nixos/log_level_to_num.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+# Maps LOG_LEVEL names to kernel loglevel=0..7 (see Makefile and early_cmdline_parser).
+
+level="${1:-error}"
+
+case "${level}" in
+0 | 1 | 2 | 3 | 4 | 5 | 6 | 7) echo "${level}" ;;
```
lrh2000: I believe this needs indents. Check existing examples:
https://github.com/asterinas/asterinas/blob/944778d7f9fd3821d4df1eb63b78ce1a456090c0/tools/nixos/print_target_nix_system.sh#L7-L10

(Note that I mostly say the indent before "xxx)". It does not seem necessary to move `echo` to a separate line and indent it separately.)

tools/nixos/log_level_to_num.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+# Maps LOG_LEVEL names to kernel loglevel=0..7 (see Makefile and early_cmdline_parser).
```
lrh2000:
```suggestion
# Maps LOG_LEVEL names to kernel `loglevel=0..7`.
# See Makefile and `ostd::early_cmdline_parser`.
```

ostd/src/log/mod.rs
```diff
@@ -119,74 +119,16 @@ pub use self::{
...
+use crate::boot::EarlyCmdline;
 
 /// Initializes the OSTD logging subsystem.
 ///
-/// Parses the `ostd.log_level` kernel command line parameter, sets the
-/// runtime max level, and registers the `log` crate bridge.
-pub(crate) fn init() {
-    let filter = parse_log_level_from_cmdline().unwrap_or(LevelFilter::Off);
+/// Applies the `loglevel` value from [`EarlyCmdline`] (parsed during early boot),
+/// sets the runtime max level, and registers the `log` crate bridge.
```
lrh2000: "Applies..." and "sets..." are actually duplicate.

```suggestion
/// Sets the runtime max level from [`EarlyCmdline::log_level`] and
/// registers the `log` crate bridge.
```

ostd/src/boot/mod.rs
```diff
@@ -113,6 +115,28 @@ pub(crate) fn init_after_heap() {
     });
 }
 
+/// The early command line arguments.
+pub struct EarlyCmdline {
```
lrh2000, line 125: Refer to the macro:
```rust
/// The early command line arguments.
///
/// [`crate::early_cmdline_parser`] can be used to specify how this is parsed
/// from the kernel command line. If it is not specified, we will use the
/// default values (see the field documentation).
```

Also the default value should be documented below:
```rust
    /// The log level.
    ///
    /// The default value is [`Level::Debug`].
    pub log_level: Level,
    /// Whether to enable the early console.
    ///
    /// The default value is `true`.
    pub has_early_console: bool,
```

ostd/libs/ostd-macros/src/lib.rs
```diff
@@ -264,6 +264,35 @@ pub fn global_heap_allocator_slot_map(_attr: TokenStream, item: TokenStream) ->
     .into()
 }
 
+/// A macro attribute to parse the early command line arguments.
+///
+/// The attributed function will be used to parse the early command line arguments.
+/// The function takes a single parameter of type `&str` and returns a `EarlyCmdline` struct.
```
lrh2000: Explain why we require the function to be `const`.

I also fix some grammar issues.

```suggestion
/// A macro attribute to register a parser for early command line arguments.
///
/// The attributed function will be used to parse the early command line
/// arguments. The function takes a single parameter of type `&str` and
/// returns an `EarlyCmdline` struct.
///
/// Note that the function will be called during an early boot stage. It
/// must not cause soundness issues even if OSTD has not been fully initialized
/// (e.g., the heap may be unavailable). OSDK enforces this by only allowing
/// the function to be `const`.
```

ostd/src/boot/mod.rs
```diff
@@ -113,6 +115,28 @@ pub(crate) fn init_after_heap() {
...
+fn __early_cmdline_parser(_cmdline: &str) -> EarlyCmdline {
+    EarlyCmdline {
+        log_level: Level::Debug,
+        has_early_console: true,
+    }
+}
+
+/// Parses the early command line arguments.
+pub(crate) fn parse_early_cmdline() -> EarlyCmdline {
+    __early_cmdline_parser(EARLY_INFO.get().unwrap().kernel_cmdline)
```
lrh2000:
```suggestion
    let kernel_cmdline = EARLY_INFO.get().unwrap().kernel_cmdline;
    __early_cmdline_parser(kernel_cmdline)
```

A bit easier to read...

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,114 @@
...
+///
+/// # Recognized tokens
+///
+/// - `earlycon` — enables the early UART console (exact token match).
+/// - `loglevel=N` — sets the maximum OSTD log level (`N` in `0..=7`).
+///
+/// Other tokens are ignored. When `loglevel` is absent, the default is `7`
+/// (debug). When `earlycon` is absent, the early console stays disabled.
+#[ostd::early_cmdline_parser]
+const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
```
lrh2000, line 47: We should have some tests for this? Similar to what you've seen in `kernel/comps/cmdline/src/dispatch.rs` (`#[ktest]`s).

lrh2000 (commented):
Thank you for your efforts in implementing the early cmdline parser framework. The implementation is overall of high quality, though I left a few (minor) comments below.

Remember to click the "Resolve conversation" button once the comment has been resolved, to help keep the discussion focused.

liulog:
Thank you so much for your detailed review, I learned a lot!

However, I have another small problem. I added ktest to `early.rs` like `dispatch.rs`, but when I run make ktest, I don't see any test pass output related to `kernel/`, unlike `ostd`/`osdk` which would have a green ok.

The following line comes from CI.
`test ostd::boot::memory_region::test::sort_full_non_overlapping ... ok`

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,54 @@ console=ttyS0
...
+
+Notes:
+- If omitted, the early console stays disabled.
+- Only the bare `earlycon` token is supported; complex Linux forms such as `earlycon=uart8250,io,0x3f8,115200` are not supported yet.
 
-### `ostd.log_level`
+### `loglevel`
 
-Set the verbosity level for Asterinas's logs.
+Set the OSTD log filter (`ostd::log::LevelFilter`).
```
tatetian: This doc is written for kernel users, who are not aware of any OSTD concepts (why should they?). Do not mention OSTD or `LevelFilter` in the doc.

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,54 @@ console=ttyS0
...
-- `off`
-- `emerg`
-- `alert`
-- `crit`
-- `error`
-- `warn` (alias: `warning`)
-- `notice`
-- `info`
-- `debug`
+Valid values are decimal integers `0` through `8`:
```
tatetian: Although Linux only supports symbolic values, we can do better by supporting symbolic values.

It is really nice of you to add the `log_level_to_num.sh` script so that we can do `make kernel LOG_LEVEL=debug`. But accepting only numeric values still complicates other parts of the system:

```diff
      log-level = lib.mkOption {
-        type = lib.types.enum [ "error" "warn" "info" "debug" "trace" ];
+        type = lib.types.enum [ "0" "1" "2" "3" "4" "5" "6" "7" "8" ];
        default = "@aster-log-level@";
-        description = "The log level of Asterinas NixOS.";
+        description =
          "Kernel `loglevel=` value (0..=8), aligned with `ostd::log::LevelFilter` and `LOG_LEVEL` / `LOG_LEVEL_NUM` in the Asterinas Makefile: 0 off, 1 emerg, 2 alert, 3 crit, 4 error, 5 warn, 6 notice, 7 info, 8 debug.";
      };
```

So why not supports symbolic values like `off`, `emerg`, and `warn` (and its aliases `warning`) at the kernel level? This way, you don't have to add `log_level_to_num.sh` or duplicating the symbol-to-value mapping (e.g., 0 -> off) in `description`.

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,54 @@ console=ttyS0
...
-ostd.log_level=error
+loglevel=4
+```
+
+Notes:
+- When omitted, invalid, or greater than `8`, the default is `8` (debug).
+
+Combined example:
+```text
+earlycon loglevel=4 console=ttyS0
```
tatetian: This combined example looks weird. Why mention this combination in particular? There are simply so many possible combinations. Why single it out? Simply saying "combined example" is not enough. No other sections give "combined example". Remove this example or justify gives the reader a hint as to why this "combined example" is so interesting.

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,174 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Parses a subset of the kernel command line before the heap is available.
```
tatetian: Saying "before the heap is available" misses the point. The correct way to frame it is:

```rust
//! A set of the kernel command line parameters to be handled early on by OSTD.
//!
//! Most parameters are recognized and handled by the Asterinas kernel,
//! but a few must be done early on by OSTD.
//! For example,
//! `earlycon` specifies why an early console (most commonly based on UART) should be initialized.
//! The early console is the only viable sink for log messages
//! generated by OSTD in the early stage,
//! before it handles the control of CPU to the Asterinas kernel.
//!
//! OSTD allows injecting a parser for these early command line parameters
//! via [`ostd::early_cmdline_parser`].
//! The injected parser is invoked so early
//! even the heap allocator is not ready.
//! As such, all parser logic in this module has to be written as const expressions.
```

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,174 @@
...
+///
+/// # Recognized tokens
+///
+/// - `earlycon` — enables the early UART console (exact token match).
+/// - `loglevel=N` — sets the OSTD log filter (`N` in `0..=8`, see [`LevelFilter`]).
+///
+/// Other tokens are ignored. When `loglevel` is absent, the default is `8`
+/// ([`LevelFilter::Debug`]). When `earlycon` is absent, the early console stays disabled.
+#[ostd::early_cmdline_parser]
+const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
```
tatetian, line 47: ## Flawed parser logic

This parser logic falls short of considering a few special but important cases in Linux CLI. Per [Linux's doc](https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html):

> The kernel parses parameters from the kernel command line up to “--”... Everything after “--” is passed as an argument to init.

So the parser logic should detect ` "--" `. If ` "--" ` is detected, ignore `earlycon` and `loglevel` afterwards.

> Double-quotes can be used to protect spaces in values, e.g.:
> ```
> param="spaces in here"
> ```

The current parse logic would mistakenly recognize `param="a earlycon b"` as if it is an `earlycon` parameter. 

In a way, both bugs are preexisting. The original parser logic in OSTD is also naive:

```rust
    let value = kcmdline
        .split(' ')
        .find(|arg| arg.starts_with("ostd.log_level="))
        .map(|arg| arg.split('=').next_back().unwrap_or_default())?;
```

But this is somewhat forgivable given that the CLI parser logic resides in OSTD, which can choose a different CLI parameter format compared to Linux. But the `cmdline` component is intended to be Linux compatible. So deviating from Linux's format is unacceptable.

## Duplicated tokenizer logic

Both `cmdline/src/early.rs` and `cmdline/src/dispatch.rs` contain tokenizer logic (splitting the CLI input string into key-value pairs). Is it possible to extract the tokenizer logic out so that the two pieces of parser logic can share the tokenizer logic?

The tokenizer is more complex than calling `String::split(' ')`. In particular, it needs to handle `param="spaces in here"`. Now that we have quoted strings, it follows that we need to handle **escaped characters** in quoted strings. So there is quite some complexity. And thus it is highly desirable to avoid duplicating the complexity in two different files.

One may argue that this bug should be fixed in a follow-up PR given that (1) handling quoted strings is non-trivial, and (2) it involves fixing a pre-existing bug in `dispatch.rs`. But I prefer to fix it now in this PR given that (1) this PR is relatively short, and (2) more importantly, sharing a quoted string-aware tokenizer would mean a structural change to `early.rs` and `dispatch.rs`. I would like to see the structure of the `cmdline` component _stable_ after this PR is merged.

tatetian (commented):
Thanks for implementing our idea of an injected CLI parameter parser. The code looks good, but I am afraid that two major concerns (see "Flawed parser logic" and "Duplicated tokenizer logic" below) demand a structural change to this PR.

tatetian:
@boterinas author

boterinas:

liulog:
@boterinas ready

kernel/comps/cmdline/src/early.rs
liulog: Thank you for pointing out the defects; the logic for -- and "" has now been added.

`early_cmdline_parser` scans the cmdline to parse relevant parameters into an `EarlyCmdline` struct, which is then passed via the stack. This is because heap-dependent data structures like Vec are unavailable at this early stage.

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,301 @@
...
+        Some(LevelFilter::Warning)
+    } else if match_token(bytes, start, end, b"notice") {
+        Some(LevelFilter::Notice)
+    } else if match_token(bytes, start, end, b"info") {
+        Some(LevelFilter::Info)
+    } else if match_token(bytes, start, end, b"debug") {
+        Some(LevelFilter::Debug)
+    } else {
+        None
+    }
```
lrh2000: We can simplify the logic and make the code much prettier by using the (const) [`slice::split_at`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at) method.

Here is an **incomplete** patch that demonstrates what I mean:
<details>
<summary> An incomplete patch </summary>

```patch
diff --git a/kernel/comps/cmdline/src/early.rs b/kernel/comps/cmdline/src/early.rs
index 0d0cba3eb..c93f4473a 100644
--- a/kernel/comps/cmdline/src/early.rs
+++ b/kernel/comps/cmdline/src/early.rs
@@ -72,16 +72,17 @@ const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
             index += 1;
         }
 
-        let (token_start, token_end) = strip_linux_double_quotes(bytes, start, index);
+        let token = strip_linux_double_quotes(bytes, start, index);
 
         // Only early-only keys are consumed here; other tokens are ignored.
-        if match_token(bytes, token_start, token_end, b"--") {
+        if matches!(token, b"--") {
             // `--` terminates kernel-side parsing; remaining tokens belong to init.
             break;
-        } else if match_token(bytes, token_start, token_end, b"earlycon") {
+        } else if matches!(token, b"earlycon") {
             has_early_console = true;
-        } else if starts_with_at(bytes, token_start, token_end, b"loglevel=")
-            && let Some(parsed_log_level) = parse_loglevel_at(bytes, token_start + 9, token_end)
+        } else if let Some((key, value)) = token.split_at_checked(9)
+            && matches!(key, b"earlycon=")
+            && let Some(parsed_log_level) = parse_loglevel_at(value)
         {
             log_level = parsed_log_level;
         }
@@ -94,8 +95,8 @@ const fn early_cmdline_parser(cmdline: &str) -> EarlyCmdline {
 }
 
 /// Strips Linux-style surrounding quotes from a token or value.
-const fn strip_linux_double_quotes(bytes: &[u8], start: usize, end: usize) -> (usize, usize) {
-    if start < end && bytes[start] == b'"' {
+const fn strip_linux_double_quotes(bytes: &[u8], start: usize, end: usize) -> &[u8] {
+    let (new_start, new_end) = if start < end && bytes[start] == b'"' {
         let new_start = start + 1;
         if new_start < end && bytes[end - 1] == b'"' {
             (new_start, end - 1)
@@ -104,7 +105,9 @@ const fn strip_linux_double_quotes(bytes: &[u8], start: usize, end: usize) -> (u
         }
     } else {
         (start, end)
-    }
+    };
+
+    bytes.split_at(new_start).1.split_at(new_end - new_start).0
 }
 
 /// Parses a `loglevel` value from `bytes[start..end]`.
```
</details>

After doing that, we can:
 1. replace `parse_loglevel_name_at` by something similar to `match ... { b"warn" => ... }`.
 2. use [`u8::from_ascii_radix`](https://doc.rust-lang.org/std/primitive.u8.html#method.from_ascii_radix) in `parse_loglevel_at`.

kernel/comps/cmdline/src/dispatch.rs
```diff
@@ -139,6 +141,9 @@ fn dispatch_params(cmdline: &str) -> InitprocArgs {
         if registry.contains_key(normalized.as_str()) {
             // Group by normalized name
             grouped.entry(normalized).or_default().push(value);
+        } else if is_early_param(normalized.as_str()) {
+            // Handled by OSTD during bootstrap; do not forward to init.
```
lrh2000:
```suggestion
            // Handled by OSTD during bootstrap: do not forward to init
```

Prefer to be consistent with existing code:
```rust
            // Unknown parameter: forward to init
```

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,57 @@ console=ttyS0
...
+Asterinas currently supports a simplified form.
+
+Example:
+```text
+earlycon
+```
+
+Notes:
+- If omitted, the early console stays disabled.
+- Only the bare `earlycon` token is supported; complex Linux forms such as `earlycon=uart8250,io,0x3f8,115200` are not supported yet.
```
lrh2000: Too long, break it:
```suggestion
- Only the bare `earlycon` token is supported;
  complex Linux forms such as `earlycon=uart8250,io,0x3f8,115200` are not supported yet.
```

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -35,28 +35,57 @@ console=ttyS0
...
 ```text
-ostd.log_level=error
+loglevel=4
+loglevel=error
 ```
 
+Notes:
+- Level names are case-sensitive; use lowercase names.
+- If omitted, the default is `8` (`debug`). Invalid values are ignored.
+- Use `info` for normal operation, `debug` when troubleshooting, and `error` or lower for a quieter console.
```
lrh2000: I hope this is true. But at least for now, `info` is unusable for normal operation because it prints system call logs.
https://github.com/asterinas/asterinas/blob/c6284a9106f5a4c87deb7c8a990af6211dc5f540/kernel/src/syscall/mod.rs#L399-L416

Maybe we should fix the problem since `info` should indeed be used for normal operations. But for now, let's at least not claim wrong things:
```suggestion
- Use `warn` for normal operation, `info`/`debug` when troubleshooting,
  and `error` or lower for a quieter console.
```

lrh2000 (commented):
During my previous round of review, I tried to see if the early parser could be simpler, even under the `const` restrictions. But I didn't find many clues at that time.

I've now checked the problem again, as the early parser is becoming more complex. It seems that the function `[u8]::split_at` can help. See the detailed comments below.

Otherwise, I only have a few nits. Thanks for the updates!

ostd/src/log/mod.rs
```diff
@@ -119,74 +119,14 @@ pub use self::{
         __write_log_record, Log, Record, STATIC_MAX_LEVEL, inject_logger, max_level, set_max_level,
     },
 };
+use crate::boot::EarlyCmdline;
```
tatetian, line 122: Combine this import with `use self::bridge::LogCrateBridge;` at a few line above.

ostd/src/boot/mod.rs
```diff
@@ -115,6 +115,38 @@ pub(crate) fn init_after_heap() {
...
+/// default values (see the field documentation).
+pub struct EarlyCmdline {
+    /// The log level filter.
+    ///
+    /// The default value is [`LevelFilter::Debug`].
+    pub log_level: LevelFilter,
+    /// Whether to enable the early console.
+    ///
+    /// The default value is `true`.
+    pub has_early_console: bool,
```
tatetian, line 146: Why the default value for `has_early_console` is set to `true`, whereas the `kernel/comps/cmdline/src/early.rs` set the default value to `false`? I think that the default value should be `false` because we have made the consensus that it is not quite safe to guess the early console.

kernel/comps/cmdline/src/early.rs
```diff
@@ -0,0 +1,301 @@
...
+use ostd::{boot::EarlyCmdline, log::LevelFilter};
+
+/// Kernel command-line keys consumed by [`early_cmdline_parser`].
+///
+/// These are stripped during dispatch so they are not forwarded to init.
+const EARLY_PARAMS: &[&str] = &["earlycon", "loglevel"];
+
+/// Returns whether `normalized_key` is handled only by the early parser.
+pub(super) fn is_early_param(normalized_key: &str) -> bool {
+    EARLY_PARAMS.contains(&normalized_key)
```
tatetian: The three lines here are the only locations that refer to this concept of `normalized_key`. Do not use terms without defining them first.

ostd/src/boot/mod.rs
lrh2000: Both `ktest` and `write_a_kernel_in_100_lines.rs` will have a problem if we set the default value to `false`. There are no other consoles for them. What's the best way to resolve this problem?

(I want to make the default value `false` either. But I don't know how to resolve this problem, so [the initial proposal from me](https://github.com/asterinas/asterinas/pull/3261#issuecomment-4535474261) contains a code snippet to set the default value to `true`...)

tatetian: Thanks for explaining! And we should explain this point to future developers:

```rust
    /// Whether to enable the early console.
    ///
    /// The default value is `true`.
    ///
    /// We choose `true` as the default value
    /// in order to give a minimal OSTD-based kernel
    /// (e.g., the one created with `osdk test`)
    /// access to an early console and thus enable logging.
    /// This is convenient for development purpose.
    ///
    /// On the other hand,
    /// blindly assuming a deployment platform is attached
    /// to a UART-based console is
    /// unacceptable for a production-grade kernel,
    /// which should instead register `crate::early_cmdline_parser`
    /// to acquire this information from the kernel parameter.
    pub has_early_console: bool,
```

tatetian:
@liulog A kind reminder

tatetian (approved):

PR #3263, opened 2026-05-22
Overhaul mount ID allocation: drop upper bound, add unique IDs

avasummer:
This PR introduces a new allocator, and overhauls VFS mount ID allocation.

- Add `SparseIdAlloc`, a range-bounded sparse `u32` ID allocator.
- Overhaul mount ID allocation: drop the upper bound, add unique IDs.

Per PR #3219, the reviewer asked for a deeper overhaul matching Linux:

> the entire mount ID allocation logic must be overhauled to (1) get rid of the hardcoded upper bound for mount IDs, (2) support the 64-bit unique mount IDs in addition to the legacy 32-bit ones, and (3) support looking up Mount with unique mount IDs. In Linux, the mount ID allocator is based on Xarray. And Linux's mount ID allocator may return ENOSPC (if the number of mount IDs reaches an upper bound specified by a user-specified policy) or ENOMEM (if Xarray fails to allocate new nodes).

## Follow-up

- Migrate the namespace lookup table from `BTreeMap<u64, Weak<Mount>>` to `XArray` once `ostd/` provides `NonNullPtr for Weak<T>`. Linux unifies allocation and lookup of unique IDs in a single xarray; we split them because Asterinas' `XArray<P>` currently requires `P: NonNullPtr`.
- Kernel-side `#[ktest]` for `unique_id` monotonicity / `lookup_by_unique_id` round-trip / `Drop` ID return.
- Userspace regression tests for `statx(STATX_MNT_ID_UNIQUE)` semantics and `/proc/self/mountinfo` field parsing.

### DIFF
--- a/kernel/libs/sparse-id-alloc/src/lib.rs
+++ b/kernel/libs/sparse-id-alloc/src/lib.rs
@@ -1,0 +1,180 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! A sparse u32 ID allocator over a configurable range.
+//!
+//! The [`SparseIdAlloc`] type allocates the smallest available ID in an
+//! inclusive `u32` range and supports returning IDs later.
+
+#![cfg_attr(not(test), no_std)]
+#![deny(unsafe_code)]
+
+extern crate alloc;
+
+use alloc::collections::BTreeMap;
+
+/// A sparse `u32` ID allocator over a configurable `[min, max]` range.
+///
+/// # Allocation policy
+///
+/// Always returns the smallest unallocated ID in the range, or `None`
+/// when every ID is in use.
+#[derive(Clone, Debug)]
+pub struct SparseIdAlloc {
+    min: u32,
+    max: u32,
+    /// Maximal runs of allocated IDs: start -> inclusive end.
+    /// Runs are disjoint and non-adjacent.
+    allocated_runs: BTreeMap<u32, u32>,
+}
+
+impl SparseIdAlloc {
+    /// Creates a new allocator that hands out IDs from the inclusive range
+    /// `[min, max]`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `min > max`.
+    pub const fn new(min: u32, max: u32) -> Self {
+        assert!(min <= max, "min must be <= max");
+        Self {
+            min,
+            max,
+            allocated_runs: BTreeMap::new(),
+        }
+    }
+
+    /// Allocates and returns a new ID from the configured range.
+    ///
+    /// Returns `None` when every ID in `[min, max]` is in use.
+    pub fn alloc(&mut self) -> Option<u32> {
+        let first = self
+            .allocated_runs
+            .first_key_value()
+            .map(|(&start, &end)| (start, end));
+
+        match first {
+            // The range's lowest IDs are allocated; the smallest free ID
+            // is just past the first run (unless the range is exhausted).
+            Some((start, end)) if start == self.min => {
+                if end == self.max {
+                    return None;
+                }
+
+                let id = end + 1;
+                let new_end = id
+                    .checked_add(1)
+                    .and_then(|next_start| self.allocated_runs.remove(&next_start))
+                    .unwrap_or(id);
+                *self.allocated_runs.get_mut(&start).unwrap() = new_end;
+                Some(id)
+            }
+            // `min` itself is free.
+            _ => {
+                let id = self.min;
+                let end = id
+                    .checked_add(1)
+                    .and_then(|next_start| self.allocated_runs.remove(&next_start))
+                    .unwrap_or(id);
+                self.allocated_runs.insert(id, end);
+                Some(id)
+            }
+        }
+    }
+
+    /// Releases the given ID.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `id` is not currently allocated (out of range, never
+    /// issued, or already freed).
+    pub fn free(&mut self, id: u32) {
+        let run = self
+            .allocated_runs
+            .range(..=id)
+            .next_back()
+            .map(|(&start, &end)| (start, end));
+        let Some((start, end)) = run.filter(|&(_, end)| id <= end) else {
+            panic!("free({id}) of unallocated ID");
+        };
+
+        if id == start {
+            self.allocated_runs.remove(&start);
+        } else {
+            *self.allocated_runs.get_mut(&start).unwrap() = id - 1;
+        }
+        if id < end {
+            self.allocated_runs.insert(id + 1, end);
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::SparseIdAlloc;
+
+    #[test]
+    fn sparse_alloc_is_monotonic_from_min() {
+        let mut a = SparseIdAlloc::new(1, u32::MAX);
+        assert_eq!(a.alloc(), Some(1));
+        assert_eq!(a.alloc(), Some(2));
+        assert_eq!(a.alloc(), Some(3));
+    }
+
+    #[test]
+    fn sparse_alloc_reuses_freed_ids_immediately() {
+        let mut a = SparseIdAlloc::new(1, u32::MAX);
+        for _ in 0..3 {
+            let _ = a.alloc();
+        }
+        a.free(3);
+        a.free(1);
+        a.free(2);
+        assert_eq!(a.alloc(), Some(1));
+        assert_eq!(a.alloc(), Some(2));
+        assert_eq!(a.alloc(), Some(3));
+    }
+
+    #[test]
+    fn sparse_alloc_reuses_smallest_gap_first() {
+        let mut a = SparseIdAlloc::new(1, 4);
+        for _ in 0..4 {
+            let _ = a.alloc();
+        }
+        assert_eq!(a.alloc(), None);
+        a.free(4);
+        a.free(2);
+        assert_eq!(a.alloc(), Some(2));
+        assert_eq!(a.alloc(), Some(4));
+        assert_eq!(a.alloc(), None);
+    }
+
+    #[test]
+    fn sparse_alloc_handles_single_id_range() {
+        let mut a = SparseIdAlloc::new(u32::MAX, u32::MAX);
+        assert_eq!(a.alloc(), Some(u32::MAX));
+        assert_eq!(a.alloc(), None);
+    }
+
+    #[test]
+    #[should_panic]
+    fn sparse_alloc_panics_when_min_greater_than_max() {
+        let _ = SparseIdAlloc::new(5, 4);
+    }
+
+    #[test]
+    #[should_panic(expected = "unallocated")]
+    fn sparse_alloc_panics_on_free_never_issued() {
+        let mut a = SparseIdAlloc::new(1, 100);
+        let _ = a.alloc();
+        a.free(5);
+    }
+
+    #[test]
+    #[should_panic(expected = "unallocated")]
+    fn sparse_alloc_panics_on_double_free() {
+        let mut a = SparseIdAlloc::new(1, 100);
+        let id = a.alloc().unwrap();
+        a.free(id);
+        a.free(id);
+    }
+}
--- a/kernel/src/fs/vfs/path/mount.rs
+++ b/kernel/src/fs/vfs/path/mount.rs
@@ -1,12 +1,11 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use core::sync::atomic::{AtomicU32, Ordering};
+use core::sync::atomic::{AtomicU32, AtomicU64, Ordering};
 
 use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
 use hashbrown::HashMap;
-use id_alloc::IdAlloc;
 use ostd::sync::{RwMutexReadGuard, RwMutexWriteGuard};
-use spin::Once;
+use sparse_id_alloc::SparseIdAlloc;
 
 use super::try_get_mnt_ns_inode;
 use crate::{
@@ -75,29 +74,64 @@ pub enum MountPropType {
     // TODO: Implement other propagation types.
 }
 
-static ID_ALLOCATOR: Once<SpinLock<IdAlloc>> = Once::new();
+/// 32-bit recyclable mount IDs.
+///
+/// IDs start at 1; 0 is never issued and denotes an invalid mount.
+/// Exhaustion returns `ENOMEM`.
+/// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L282>
+static MOUNT_ID_ALLOCATOR: SpinLock<SparseIdAlloc> =
+    SpinLock::new(SparseIdAlloc::new(1, i32::MAX as u32));
 
-/// The reserved mount ID, which represents an invalid mount.
-static RESERVED_MOUNT_ID: usize = 0;
+/// The first 64-bit unique mount ID.
+///
+/// The minimum keeps unique IDs above the recyclable ID range, ensuring
+/// the two ID spaces never overlap numerically.
+///
+/// This value is chosen to align with Linux.
+/// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L73>
+pub const MNT_UNIQUE_ID_MIN: u64 = (1 << 31) + 1;
 
-pub(super) fn init() {
-    // TODO: Make it configurable.
-    const MAX_MOUNT_NUM: usize = 10000;
+/// Monotonically increasing 64-bit unique mount IDs.
+static NEXT_FREE_UNIQUE_ID: AtomicU64 = AtomicU64::new(MNT_UNIQUE_ID_MIN);
 
-    let mut id_allocator = IdAlloc::with_capacity(MAX_MOUNT_NUM);
-    let _ = id_allocator.alloc_specific(RESERVED_MOUNT_ID).unwrap(); // Reserve mount ID 0.
+pub(super) fn init() {}
 
-    ID_ALLOCATOR.call_once(|| SpinLock::new(id_allocator));
+/// A mount's pair of identifiers.
+///
+/// Allocates from [`MOUNT_ID_ALLOCATOR`] and [`NEXT_FREE_UNIQUE_ID`] on
+/// construction; on drop, releases the recyclable ID back to the pool.
+/// The unique ID is monotonic and is not freed.
+pub(super) struct MountId {
+    recyclable_id: u32,
+    unique_id: u64,
 }
 
-/// Allocates a new mount ID from the global pool.
-fn alloc_mount_id() -> Result<usize> {
-    ID_ALLOCATOR
-        .get()
-        .unwrap()
-        .lock()
-        .alloc()
-        .ok_or_else(|| Error::with_message(Errno::ENOMEM, "mount ID pool exhausted"))
+impl MountId {
+    /// Allocates a new pair of mount identifiers.
+    ///
+    /// Returns `None` if the recyclable ID range is exhausted.
+    pub(super) fn alloc() -> Option<Self> {
+        let recyclable_id = MOUNT_ID_ALLOCATOR.lock().alloc()?;
+        let unique_id = NEXT_FREE_UNIQUE_ID.fetch_add(1, Ordering::Relaxed);
+        Some(Self {
+            recyclable_id,
+            unique_id,
+        })
+    }
+
+    pub(super) fn recyclable_id(&self) -> u32 {
+        self.recyclable_id
+    }
+
+    pub(super) fn unique_id(&self) -> u64 {
+        self.unique_id
+    }
+}
+
+impl Drop for MountId {
+    fn drop(&mut self) {
+        MOUNT_ID_ALLOCATOR.lock().free(self.recyclable_id);
+    }
 }
 
 bitflags! {
@@ -206,8 +240,9 @@ define_atomic_version_of_integer_like_type!(PerMountFlags, {
 /// Each `Mount` can be viewed as a node in the mount tree, maintaining
 /// mount-related information and the structure of the mount tree.
 pub struct Mount {
-    /// Global unique identifier for the mount node.
-    id: usize,
+    /// Pair of recyclable and unique mount identifiers; the recyclable
+    /// half is returned to the pool when the `Mount` drops.
+    id: MountId,
     /// Root dentry.
     root_dentry: Arc<Dentry>,
     /// Mountpoint dentry. A mount node can be mounted on one dentry of another mount node,
@@ -293,9 +328,10 @@ impl Mount {
         mnt_ns: Weak<MountNamespace>,
         source: Option<String>,
     ) -> Result<Arc<Self>> {
-        let id = alloc_mount_id()?;
+        let id = MountId::alloc()
+            .ok_or_else(|| Error::with_message(Errno::ENOMEM, "mount ID space exhausted"))?;
 
-        Ok(Arc::new_cyclic(|weak_self| Self {
+        let mount = Arc::new_cyclic(|weak_self| Self {
             id,
             root_dentry: Dentry::new_root(fs.root_inode()),
             mountpoint: RwLock::new(None),
@@ -307,12 +343,28 @@ impl Mount {
             propagation: RwLock::new(MountPropType::default()),
             flags: AtomicPerMountFlags::new(flags),
             this: weak_self.clone(),
-        }))
+        });
+        // `upgrade` returns `None` for pseudo mounts (no namespace) and for
+        // mounts still being built by `MountNamespace::new_with_root` (the
+        // namespace is a `UniqueArc` at that point). The latter are
+        // registered by `new_with_root` after the namespace is finalized.
+        if let Some(ns) = mount.mnt_ns.upgrade() {
+            ns.register_mount(&mount);
+        }
+        Ok(mount)
     }
 
-    /// Gets the mount ID.
-    pub fn id(&self) -> usize {
-        self.id
+    /// Gets the recyclable 32-bit mount ID.
+    pub fn id(&self) -> u32 {
+        self.id.recyclable_id()
+    }
+
+    /// Gets the unique 64-bit mount ID.
+    ///
+    /// Unique IDs start at [`MNT_UNIQUE_ID_MIN`], increase monotonically,
+    /// and are never reused.
+    pub fn unique_id(&self) -> u64 {
+        self.id.unique_id()
     }
 
     /// Returns the mount source.
@@ -389,9 +441,10 @@ impl Mount {
         root_dentry: &Arc<Dentry>,
         new_ns: &Weak<MountNamespace>,
     ) -> Result<Arc<Self>> {
-        let id = alloc_mount_id()?;
+        let id = MountId::alloc()
+            .ok_or_else(|| Error::with_message(Errno::ENOMEM, "mount ID space exhausted"))?;
 
-        Ok(Arc::new_cyclic(|weak_self| Self {
+        let mount = Arc::new_cyclic(|weak_self| Self {
             id,
             root_dentry: root_dentry.clone(),
             mountpoint: RwLock::new(None),
@@ -403,7 +456,11 @@ impl Mount {
             propagation: RwLock::new(MountPropType::default()),
             flags: AtomicPerMountFlags::new(self.flags.load(Ordering::Relaxed)),
             this: weak_self.clone(),
-        }))
+        });
+        if let Some(ns) = mount.mnt_ns.upgrade() {
+            ns.register_mount(&mount);
+        }
+        Ok(mount)
     }
 
     /// Clones a mount tree starting from the specified root `Dentry`.
@@ -703,6 +760,12 @@ impl Debug for Mount {
 impl Drop for Mount {
     fn drop(&mut self) {
         self.clear_mountpoint();
-        ID_ALLOCATOR.get().unwrap().lock().free(self.id);
+        // `upgrade` returns `None` for pseudo mounts (no namespace) and for
+        // mounts whose namespace is itself being dropped (the `mounts` map
+        // will be freed in a moment).
+        if let Some(ns) = self.mnt_ns.upgrade() {
+            ns.deregister_mount(self.id.unique_id());
+        }
+        // The recyclable ID is returned to the pool by `MountId`'s `Drop`.
     }
 }
--- a/kernel/src/fs/vfs/path/mount_namespace.rs
+++ b/kernel/src/fs/vfs/path/mount_namespace.rs
@@ -164,10 +164,6 @@ impl MountNamespace {
     ///
     /// No recyclable-`id` counterpart exists: the 32-bit ID space is reused
     /// on drop, so a keyed lookup would race the next allocation.
-    #[expect(
-        dead_code,
-        reason = "no in-tree consumer yet; staged for statmount/listmount"
-    )]
     pub fn lookup_by_unique_id(&self, unique_id: u64) -> Option<Arc<Mount>> {
         let mount = {
             let mounts = self.mounts.lock();
@@ -178,6 +174,34 @@ impl MountNamespace {
         mount
             .is_equal_or_descendant_of(self.root(), &topology_guard)
             .then_some(mount)
+    }
+
+    /// Returns live mounts in this namespace that are strict descendants of
+    /// `parent`, ordered by [`Mount::unique_id`].
+    ///
+    /// Live mounts are snapshotted under `mounts.lock`; ancestry is checked
+    /// afterward so per-`Mount` locks are never acquired while the table lock
+    /// is held.
+    pub fn descendant_mounts_of(
+        &self,
+        parent: &Arc<Mount>,
+    ) -> impl DoubleEndedIterator<Item = Arc<Mount>> {
+        let snapshot: Vec<_> = {
+            let guard = self.mounts.lock();
+            guard.values().filter_map(|weak| weak.upgrade()).collect()
+        };
+
+        let topology_guard = MountTopology::read_lock();
+
+        let descendants: Vec<_> = snapshot
+            .into_iter()
+            .filter(|mount| {
+                !Arc::ptr_eq(mount, parent)
+                    && mount.is_equal_or_descendant_of(parent, &topology_guard)
+            })
+            .collect();
+
+        descendants.into_iter()
     }
 
     /// Walks the in-place mount tree and registers every mount in
--- a/kernel/src/fs/vfs/path/mount_namespace.rs
+++ b/kernel/src/fs/vfs/path/mount_namespace.rs
@@ -38,6 +38,8 @@ pub struct MountNamespace {
     owner: Arc<UserNamespace>,
     /// The stashed dentry in nsfs.
     stashed_dentry: StashedDentry,
+    /// Live mounts that belong to this namespace, keyed by [`Mount::unique_id`].
+    mounts: SpinLock<BTreeMap<u64, Weak<Mount>>>,
 }
 
 impl PartialEq for MountNamespace {
@@ -74,10 +76,16 @@ impl MountNamespace {
             root: None,
             owner,
             stashed_dentry: StashedDentry::new(),
+            mounts: SpinLock::new(BTreeMap::new()),
         });
         let root = build_root_fn(&UniqueArc::downgrade(&new_ns))?;
         new_ns.root = Some(root);
-        Ok(UniqueArc::into_arc(new_ns))
+        let ns = UniqueArc::into_arc(new_ns);
+        // Mounts created inside `build_root_fn` saw a `Weak` to a
+        // `UniqueArc`, whose `upgrade` returns `None`; they could not
+        // self-register at construction. Register the finished tree.
+        ns.register_existing_mount_tree();
+        Ok(ns)
     }
 
     /// Returns a reference to the singleton initial mount namespace.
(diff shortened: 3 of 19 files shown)


ostd/libs/id-alloc/src/lib.rs
```diff
@@ -169,9 +172,71 @@ impl Debug for IdAlloc {
     }
 }
 
+/// A sparse `u32` ID allocator over a configurable `[min, max]` range.
+#[derive(Clone, Debug)]
+pub struct SparseIdAlloc {
```
tatetian: Per the [framekernel](https://asterinas.github.io/book/kernel/the-framekernel-architecture.html#requirements-for-the-os-framework) architecture, the complexity of OSTD is ought to be kept minimal. This means if a module can be extract out of OSTD without compromising functionality and efficiency, then it should be moved out. I can't see why `SparseIdAlloc` should be implemented as a library that the OSTD relies on. `IdAlloc` is an OSTD library because OSTD needs basic ID allocator. But all current users of `SparseIdAlloc` live outside OSTD. So `SparseIdAlloc` should also be implemented outside OSTD.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -169,9 +172,71 @@ impl Debug for IdAlloc {
...
+    min: u32,
+    max: u32,
+    next_id: Option<u32>,
+    freed: BTreeSet<u32>,
```
tatetian: It is ok to consume memory space in $O(N)$ complexity, where $N$ is the number of allocated IDs. But consuming $O(N)$ complexity, where $N$ is the number of free IDs, is unreasonable. So the field should be `allocated: BTreeSet<u32>`.

---

This `SparseIdAlloc` needs to reuse previously-freed IDs, but does it need to use the minimum available free IDs? This is user-visible behavior. You better document this behavior in the doc comments of this type and make sure that the Linux-compatible kernel user code (e.g., mount IDs) want this behavior.

---

At a high level, we are keeping an allocation (or free) bitmap. Yes, it is a sparse bitmap. The most space-efficient data structure for a sparse bitmap is XArray (a radix tree). [XArray](https://github.com/asterinas/asterinas/tree/main/kernel/libs/xarray) is already provided as a kernel library, but there is a problem: currently its leaf nodes can only store pointers, not values. As such, the current `XArray` can only be used as a sparse key-pointer map, not a sparse bitmap. `XArray` as a sparse bitmap could find use cases beyond managing mount IDs. For example, it is common in file systems to manage a large amount of IDs for objects like inodes and blocks. So I think it would be highly useful to extend `Xarray` so that it can support both `XArray<P: NonNullPtr>` and `XArray<u64>`. There could be some non-trivial design to achieve this goal. Feel like to take on this challenge?

Since this is mostly an optimization, it is ok to extend Xarray in a followup PR.

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -34,6 +34,12 @@ pub struct MountNamespace {
     owner: Arc<UserNamespace>,
     /// The stashed dentry in nsfs.
     stashed_dentry: StashedDentry,
+    /// Live mounts that belong to this namespace, keyed by [`Mount::unique_id`].
+    ///
+    /// Mirrors Linux's per-namespace `mnt_namespace->mounts` rb-tree.
+    ///
+    /// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/mount.h#L14>
+    mounts: SpinLock<BTreeMap<u64, Weak<Mount>>>,
```
tatetian, line 42: Consider adding the `listmount` syscall as a user of this map. Otherwise, this map would be introduced without any real user code.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -45,29 +45,38 @@ pub enum MountPropType {
...
+    // `STATX_MNT_ID`'s legacy semantics.
+    // Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L282>
+    MOUNT_ID_ALLOCATOR.call_once(|| SpinLock::new(SparseIdAlloc::new(1, i32::MAX as u32)));
 }
 
-/// Allocates a new mount ID from the global pool.
-fn alloc_mount_id() -> Result<usize> {
-    ID_ALLOCATOR
+/// Allocates the `(id, unique_id)` pair for a new mount.
+fn alloc_mount_id() -> Result<(u32, u64)> {
```
tatetian: I suggest introducing a new type:

```rust
pub MountId {
    recyclable_id: u32,
    unique_id: u64,
}

impl MountId {
    pub fn alloc() -> Option<Self> { ... }

    pub fn recyclable_id(&self) -> u32 { ... }

    pub fn unique_id(&self) -> u64 { ... }
}

impl Drop for MountId { ... }
```

The `Mount` struct can then (1) include this struct rather than two separate values, and (2) enjoy RAII rather than manually deallocating IDs.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -259,14 +273,27 @@ impl Mount {
...
+        if let Some(ns) = mount.mnt_ns.upgrade() {
+            ns.register_mount(&mount);
+        }
+        Ok(mount)
     }
 
-    /// Gets the mount ID.
-    pub fn id(&self) -> usize {
+    /// Gets the recyclable 32-bit mount ID.
+    pub fn id(&self) -> u32 {
```
tatetian, line 358: Maybe calling it `recyclable_id` would be better to emphasize the fact that there are two kinds of IDs and this method only returns the recyclable one.

tatetian (commented):
Thanks for following up with your previous PR. And I really appreciate it.

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -34,6 +34,12 @@ pub struct MountNamespace {
     owner: Arc<UserNamespace>,
     /// The stashed dentry in nsfs.
     stashed_dentry: StashedDentry,
+    /// Live mounts that belong to this namespace, keyed by [`Mount::unique_id`].
+    ///
+    /// Mirrors Linux's per-namespace `mnt_namespace->mounts` rb-tree.
+    ///
+    /// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/mount.h#L14>
```
lrh2000: I don't think we should write comments / reference link like this. Aligning the internal structures to Linux is a non-goal.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -45,29 +45,38 @@ pub enum MountPropType {
...
+/// Recyclable 32-bit mount IDs.
+/// `0` is reserved as the invalid mount ID. Exhaustion returns `ENOSPC`.
+/// Leaf lock; no other lock may be acquired while this is held.
+static MOUNT_ID_ALLOCATOR: Once<SpinLock<SparseIdAlloc>> = Once::new();
 
-/// The reserved mount ID, which represents an invalid mount.
-static RESERVED_MOUNT_ID: usize = 0;
+/// Keeps `unique_id` above the recyclable `id` range so the two ID spaces
+/// never numerically overlap.
+/// Mirrors Linux's `MNT_UNIQUE_ID_OFFSET`.
```
lrh2000: Add some blank lines. The first line should be a summary line and should be a noun for static constants.
```rust
/// 32-bit recyclable mount IDs.
///
/// `0` is reserved as the invalid mount ID. Exhaustion returns `ENOSPC`.
/// The lock is a leaf lock; no other lock may be acquired while this is held.
static MOUNT_ID_ALLOCATOR: Once<SpinLock<SparseIdAlloc>> = Once::new();

/// An offset from which 64-bit unique mount IDs start.
///
/// The offset keeps the unique ID above the recyclable ID range, ensuring
/// the two ID spaces never overlap numerically.
/// 
/// This value is chosen to align with Linux.
/// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L73>
const MNT_UNIQUE_ID_OFFSET: u64 = 1 << 31;
```

kernel/src/syscall/statx.rs
```diff
@@ -138,9 +138,23 @@ impl Statx {
             stx_attributes |= STATX_ATTR_MOUNT_ROOT;
         }
 
+        // Mirrors Linux's `vfs_statx_path` selection on `STATX_MNT_ID_UNIQUE`.
```
lrh2000: Say the behavior. Mentioning Linux method name provides no real value. Readers can check the Linux implemenation via the link.
```rust
        // If `STATX_MNT_ID_UNIQUE` is specified, a 64-bit unique ID is provided
        // instead of a 32-bit recyclable ID.
        // Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/stat.c#L303>
```

kernel/src/fs/vfs/path/mount.rs
avasummer: Considered this — I lean toward keeping Mount::id() to mirror Linux's mnt_id / mnt_id_unique asymmetry. Call-site comparisons like self.mount.id() == other.mount.id() stay clearer; using a "recyclable" id as an identity key would read against itself. The distinction still lives on the MountId field (recyclable_id: u32) and in Mount::id()'s doc, with Mount::unique_id() next to it making both kinds visible. What do you think? :)

tatetian: I am still leaning toward `recyclable_id` (or `legacy_id`), but this is just a personal preference. If you think calling `id` is better, then so be it 😄

avasummer:
Force-pushed addressing the review. Also adds the `listmount(2)` syscall and a regression test.

avasummer:
Rebased onto the latest main to resolve conflicts.

avasummer:
A gentle ping — this has been rebased and CI is green. Ready for another round of review whenever you have time. :)

kernel/libs/sparse-id-alloc/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+/// past them so that every newly issued ID is greater than every
+/// previously issued one while fresh IDs remain. Once the watermark
+/// has covered the entire range, [`Self::alloc`] returns the smallest
+/// unallocated ID in `[min, max]`, or `None` when every ID is in use.
+#[derive(Clone, Debug)]
+pub struct SparseIdAlloc {
+    min: u32,
+    max: u32,
+    /// Next never-issued ID. `None` once the watermark has passed `max`.
+    next_id: Option<u32>,
```
tatetian: The allocation strategy is hybrid: at first, it allocates

> IDs are issued in increasing order starting from `min`

then 

> Once the watermark has covered the entire range, [`Self::alloc`] returns the smallest unallocated ID in `[min, max]`.

This hybrid strategy is weird to me. In the first phase, the strategy keeps the monotonicity of allocated IDs, in the second phase, it gives up the monotonicity completely.

A more natural strategy would be an unified approach:

> Always allocate the smallest free ID that is greater than the previously-allocated ID.

If there no free IDs greater than the previously-allocated ID, then take (min - 1) as if it is the previously-allocated ID to search for free IDs from the beginning. This unified approach tries its best to keep the monotonicity of allocated IDs, which reduces the odds of the userspace getting confused with reused IDs.

Since the ID allocation strategy is visible to the userspace, could you dig into Linux's implementation?

kernel/libs/sparse-id-alloc/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+impl SparseIdAlloc {
+    /// Creates a new allocator that hands out IDs from the inclusive range
+    /// `[min, max]`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `min > max`.
+    pub fn new(min: u32, max: u32) -> Self {
```
tatetian: If a constructor method can be marked `const`, then do so. This allows for defining a static variable of this type.

---

Interestingly, I wrote the above suggestion when reviewing the first commit. Then, when I did the second commit, such a use case indeed emerged:

```rust
static MOUNT_ID_ALLOCATOR: Once<SpinLock<SparseIdAlloc>> = Once::new();
```

If `SparseIdAlloc::new` is const, then you can write:

```rust
static MOUNT_ID_ALLOCATOR: SpinLock<SparseIdAlloc> = {
    let id_alloc = SparseIdAlloc::new(1, i32::max() as u32);
    SpinLock::new(id_alloc);
};
```

This is a bit of more efficient as it avoids runtime check to see if the variable is initialized.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -45,29 +45,72 @@ pub enum MountPropType {
...
+///
+/// The offset keeps the unique ID above the recyclable ID range, ensuring
+/// the two ID spaces never overlap numerically.
+///
+/// This value is chosen to align with Linux.
+/// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L73>
+const MNT_UNIQUE_ID_OFFSET: u64 = 1 << 31;
+
+/// Monotonically increasing 64-bit unique mount IDs.
+static UNIQUE_ID_COUNTER: AtomicU64 = AtomicU64::new(MNT_UNIQUE_ID_OFFSET + 1);
```
tatetian: This is not a "counter" as it does not reflect the count of unique IDs. Rename it `NEXT_FREE_UNIQUE_ID`.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -45,29 +45,72 @@ pub enum MountPropType {
...
-/// The reserved mount ID, which represents an invalid mount.
-static RESERVED_MOUNT_ID: usize = 0;
+/// An offset from which 64-bit unique mount IDs start.
+///
+/// The offset keeps the unique ID above the recyclable ID range, ensuring
+/// the two ID spaces never overlap numerically.
+///
+/// This value is chosen to align with Linux.
+/// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L73>
+const MNT_UNIQUE_ID_OFFSET: u64 = 1 << 31;
```
tatetian: Why defining `MNT_UNIQUE_ID_OFFSET` as `1 << 31` then `UNIQUE_ID_COUNTER` as `1 << 31 + `? Why not simply define `MNT_UNIQUE_ID_MIN` as `1 << 31 + 1`? This avoid the potential off-by-one mistake.

In addition, the `Mount::unique_id` method returns the 64-bit value of the unique mount ID. So as part of the contract of `unique_id` method, the method should state that the unique ID is allocated starting from `MNT_UNIQUE_ID_MIN`, rather than 0 or 1. This point would be unobvious without explicit documentation.

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -129,6 +137,45 @@ impl MountNamespace {
...
+        self.mounts.lock().get(&unique_id).and_then(Weak::upgrade)
+    }
+
+    /// Walks the in-place mount tree and registers every mount in
+    /// [`Self::mounts`].
+    ///
+    /// The `children` read guard is released before the next iteration so
+    /// that `register_mount` (which takes `mounts.lock`) is never called
+    /// under a per-`Mount` lock.
+    fn register_existing_mount_tree(self: &Arc<Self>) {
```
tatetian, line 209: This function is vulnerable to a race condition where concurrent reader tries to inspect the mounts in the mount namespace. Because this function as a whole is not protected with a lock, the concurrent reader might see some mounts registered, while some are not. Rather than calling `register_mount` repeatedly, the method should acquire the guard from `self.mounts.lock()` at the beginning of this method.

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -153,14 +153,52 @@ impl MountNamespace {
...
+    /// descendants of `parent`, ordered by `unique_id`.
+    ///
+    /// `after` is exclusive: entries with `unique_id == after` are skipped;
+    /// passing `0` yields the whole namespace. With `reverse = true` the order
+    /// is reversed and `after` becomes the strict upper bound. At most `limit`
+    /// IDs are returned.
+    ///
+    /// Ancestry is checked outside the `mounts` lock so that the per-`Mount`
+    /// parent locks are never acquired while the table lock is held.
+    pub fn descendant_ids_of(
```
tatetian: This API has been hardwired to satisfy the semantics of the syscall. A more desirable design is to this API _flexible_. For example, this method can be defined to 

```rust
pub fn child_mounts_of(
    &self,
    parent: &Arc<Mount>,
) -> impl Iterator<Arc<Mount>> {
    ...
}
```

The caller can use the iterator to limit, reverse, or filter `Mount` objects as the caller wants.

tatetian (commented):
Thanks for revising this PR. This PR looks in a good shape. I only found a major issue (about `SparseIdAlloc`) the and a few minor issues.

tatetian:
In addition, I am testing and revising the new code review skill. Here is the output for your reference. I skimmed through the AI comments, which seem sensible. Take these comments with a grain of salt.

## `kernel/src/syscall/listmount.rs` line 68

**`req.size` validation is stricter than Linux and the struct is missing `mnt_ns_id`** (significant): Linux's `copy_mnt_id_req` accepts any `size` in `[MNT_ID_REQ_SIZE_VER0, PAGE_SIZE]` via `copy_struct_from_user` (`-EINVAL` only below 24, `-E2BIG` above a page), giving forward/backward compatibility; the current `struct mnt_id_req` is 32 bytes (`{size, spare, mnt_id, param, mnt_ns_id}`). Rejecting `size != 24` here means a caller built against current headers (size 32) gets `EINVAL` where Linux succeeds. The `book` doc honestly records this as a limitation, so this is a conformance gap rather than a bug, but the most future-proof fix is to model the full struct and validate a range. If cross-namespace support is intentionally deferred, at minimum accept `size >= MNT_ID_REQ_SIZE_VER0` and require the trailing `mnt_ns_id` to be zero. Fix:

```rust
if req.size < MNT_ID_REQ_SIZE_VER0 {
    return_errno_with_message!(Errno::EINVAL, "mnt_id_req too small");
}
// Accept larger (future) structs but require the unsupported tail to be zero,
// mirroring Linux's copy_struct_from_user forward-compat contract.
```

---

## `kernel/src/syscall/listmount.rs` line 71

**Missing `EINVAL` guard for `mnt_id`/`param` at or below `MNT_UNIQUE_ID_OFFSET`** (significant): Linux rejects `kreq->mnt_id <= MNT_UNIQUE_ID_OFFSET` (with `LSMT_ROOT` exempt) and `param != 0 && param <= MNT_UNIQUE_ID_OFFSET` with `EINVAL`, because no real unique ID can fall in that low range. Here such a `mnt_id` instead falls through to `lookup_by_unique_id` and surfaces as `ENOENT`, and such a `param` is silently treated as a valid empty cursor — both differ from Linux's `EINVAL`. The existing test `listmount_rejects_unknown_mnt_id` actually encodes the wrong expectation (`mnt_id = 0` → `ENOENT`) where Linux returns `EINVAL`. Fix: after parsing `req`, add

```rust
if req.mnt_id != LSMT_ROOT && req.mnt_id <= MNT_UNIQUE_ID_OFFSET {
    return_errno_with_message!(Errno::EINVAL, "mnt_id out of range");
}
if req.param != 0 && req.param <= MNT_UNIQUE_ID_OFFSET {
    return_errno_with_message!(Errno::EINVAL, "param out of range");
}
```

and update the C test (and import `MNT_UNIQUE_ID_OFFSET` or inline `1 << 31`).

---

## `kernel/src/fs/vfs/path/mount.rs` line 413

**`clone_mount` registers the clone before its parent is linked** (significant): `clone_mount` calls `ns.register_mount(&mount)` while the clone still has `parent: None`; the caller `clone_mount_tree` sets the real parent later via `set_parent` (line 467). This is safe today only because the sole caller path (`new_clone` → `new_with_root`) holds the namespace as a `UniqueArc`, so `mnt_ns.upgrade()` returns `None` and no registration happens during the parent-less window — `register_existing_mount_tree` registers the finished tree afterward. But nothing enforces that `clone_mount_tree` is never called with a finalized (upgradeable) namespace; if it ever were, a concurrent `descendant_ids_of` could momentarily misclassify the parent-less clone as a non-descendant. Fix: either register clones after parent linkage is established (move registration out of `clone_mount` into `clone_mount_tree`/`do_mount` after `set_parent`), or document the invariant explicitly on `clone_mount_tree`:

```rust
/// # Invariant
///
/// Must only be called while `new_ns` is still under construction (a
/// `UniqueArc`), so that clones do not self-register before `set_parent`.
```

---

## `kernel/libs/sparse-id-alloc/src/lib.rs` line 80

**`free` of an unallocated ID is a silent no-op in release** (significant): `free` only `debug_assert!`s that the ID was present, so a double-free or stale-ID free in a release build silently does nothing. For mount IDs that is mostly benign, but the more dangerous direction is undetected: if two `MountId`s ever held the same recyclable ID (an allocator bug), the second `free` would be silent in release and could later hand the live ID to a third mount. The unique-ID indirection makes accidental aliasing unlikely, but consider returning a `bool` so callers can assert, or keeping the `debug_assert` while documenting that release builds rely on callers never double-freeing. No code change strictly required; flag for a design decision on whether silent release-mode no-op is acceptable for an allocator primitive.

---

## `kernel/src/fs/vfs/path/mount.rs` line 50

**Doc comment overstates the "reserved" ID 0** (minor): The comment says "`0` is reserved as the invalid mount ID", but the allocator is constructed as `SparseIdAlloc::new(1, i32::MAX as u32)`, so `0` is simply outside the range and never issued — there is no explicit reservation step anymore (unlike the old `alloc_specific(0)`). Fix: reword to "IDs start at `1`; `0` is never issued and denotes an invalid mount."

---

## `kernel/src/fs/vfs/path/mount.rs` line 293

**Inconsistent errno for ID exhaustion vs. the old code** (minor): Exhaustion now returns `ENOSPC` ("mount ID space exhausted"); the previous code returned `ENOMEM`. Linux's mount-id allocation failure path returns `ENOMEM` (the `xa_alloc` failure in `alloc_mnt_id`). `ENOSPC` is defensible but diverges from Linux. Confirm `ENOSPC` is intended; if matching Linux is preferred, use `Errno::ENOMEM`. The same literal appears at line 397 (`clone_mount`) and should match whichever is chosen.

---

## `kernel/src/fs/vfs/path/mount_namespace.rs` line 179

**`descendant_ids_of` snapshots the whole range before applying `limit`** (minor): The snapshot collects and upgrades every entry in the (possibly whole-namespace) range into a `Vec`, then filters by ancestry, then `take(limit)`. Because `listmount` paginates by repeatedly calling this with a small `nr_mnt_ids`, a namespace of N mounts costs O(N) per call and O(N²) to page through — Linux walks the ordered list and stops early. For the strict-descendants-of-root common case the filter keeps nearly everything so the cost is unavoidable, but for a deep query of a small subtree this materializes the entire forward range needlessly. Fix: apply the descendancy filter and `limit` lazily while still holding the snapshot of `Weak`s, or short-circuit once `limit` descendants are collected:

```rust
let mut out = Vec::with_capacity(limit.min(snapshot.len()));
for (id, m) in snapshot {
    if !Arc::ptr_eq(&m, parent) && m.is_equal_or_descendant_of(parent) {
        out.push(id);
        if out.len() == limit { break; }
    }
}
out
```

---

## `kernel/src/fs/vfs/path/mount_namespace.rs` line 181

**`Box<dyn Iterator>` allocates on every `descendant_ids_of` call** (minor): The reverse/forward branch is unified by boxing the iterator, adding a heap allocation and dynamic dispatch on a syscall hot path. The two arms can instead drive a single loop without boxing, e.g. by collecting under each branch directly, or by using `itertools::Either` if available in the crate. Low impact given the syscall is rare; noting for craft. Fix: collect the snapshot inside each `if`/`else` arm rather than producing a `Box<dyn Iterator>` first.

---

## `test/initramfs/src/regression/fs/mount/listmount.c` line 245

**Test encodes a non-Linux expectation for an out-of-range `mnt_id`** (minor): `listmount_rejects_unknown_mnt_id` expects `ENOENT` for `mnt_id = 0`, but on real Linux `mnt_id = 0` (≤ `MNT_UNIQUE_ID_OFFSET`, not `LSMT_ROOT`) returns `EINVAL`. This locks in the divergence from the `mnt_id` guard comment above. Once the `EINVAL` guard is added, change the expected errno to `EINVAL` and rename the test (e.g. `listmount_rejects_out_of_range_mnt_id`). This is the regression test for the fix in `listmount.rs`.

tatetian:
@boterinas author

boterinas:

tatetian:
@avasummer A friendly ping

kernel/libs/sparse-id-alloc/src/lib.rs
avasummer: I checked Linux's implementation. `mnt_alloc_id()` uses `__xa_alloc()`, whose docs say "Finds an empty entry in @xa between @limit.min and @limit.max".

I think `SparseIdAlloc` could behave like:
```rust
    /// Allocates and returns a new ID from the configured range.
    ///
    /// Returns `None` when every ID in `[min, max]` is in use.
    pub fn alloc(&mut self) -> Option<u32> {
        let mut candidate = self.min;

        for &allocated_id in self.allocated.range(self.min..=self.max) {
            if candidate != allocated_id {
                break;
            }
            if candidate == self.max {
                return None;
            }
            candidate += 1;
        }

        self.allocated.insert(candidate);
        Some(candidate)
    }
```

kernel/src/fs/vfs/path/mount.rs
avasummer: Agree. I just followed linux's definition before, where `MNT_UNIQUE_ID_OFFSET` is used for boundary checks. In this case, using `MNT_UNIQUE_ID_MIN` is clearer.

kernel/libs/sparse-id-alloc/src/lib.rs
tatetian: So you are saying that the user-visible ID allocation strategy of Linux is to **always allocate the smallest available ID**. If this is the case, then we should adopt the same strategy. Write down this strategy in the doc comments of `SparseIdAlloc`.

tatetian:
By the way, please rebase this PR to the latest main branch.

test/initramfs/src/regression/fs/mount/listmount.c
```diff
@@ -0,0 +1,265 @@
...
+#define LSMT_ROOT 0xffffffffffffffffULL
+#endif
+
+#ifndef LISTMOUNT_REVERSE
+#define LISTMOUNT_REVERSE (1U << 0)
+#endif
+
+#ifndef MNT_UNIQUE_ID_MIN
+#define MNT_UNIQUE_ID_MIN ((1ULL << 31) + 1)
+#endif
```
lrh2000:
```suggestion
```

Try to include `#include <linux/mount.h>` instead.

test/initramfs/src/regression/fs/mount/listmount.c
```diff
@@ -0,0 +1,265 @@
...
+	struct linux_mnt_id_req req = {
+		.size = MNT_ID_REQ_SIZE_VER0,
+		.mnt_id = LSMT_ROOT,
+	};
+	uint64_t first[16];
+	long n_total =
+		TEST_RES(syscall(SYS_listmount, &req, first, 16, 0), _ret >= 2);
+	if (n_total < 2) {
+		goto cleanup;
+	}
```
lrh2000: Just assume the test pass. Making regression tests robust is a non-goal.
```suggestion
```

test/initramfs/src/regression/fs/mount/listmount.c
```diff
@@ -0,0 +1,265 @@
...
+		.mnt_id = LSMT_ROOT,
+	};
+	uint64_t asc[16];
+	long n_asc =
+		TEST_RES(syscall(SYS_listmount, &req, asc, 16, 0), _ret >= 2);
+	uint64_t desc[16];
+	long n_desc = TEST_RES(syscall(SYS_listmount, &req, desc, 16,
+				       LISTMOUNT_REVERSE),
+			       _ret == n_asc);
+	if (n_asc >= 2 && n_asc == n_desc) {
```
lrh2000:
```suggestion
```

test/initramfs/src/regression/fs/mount/listmount.c
```diff
@@ -0,0 +1,265 @@
...
+}
+END_TEST()
+
+FN_TEST(listmount_reverse_inverts_order)
+{
+	TEST_SUCC(unshare(CLONE_NEWNS));
+	ensure_dir("/tmp/lm_r1");
+	TEST_SUCC(mount("tmpfs", "/tmp/lm_r1", "tmpfs", 0, NULL));
+	ensure_dir("/tmp/lm_r2");
+	TEST_SUCC(mount("tmpfs", "/tmp/lm_r2", "tmpfs", 0, NULL));
```
lrh2000: Why not just:
```c
FN_SETUP(setup)
{
    CHECK(unshare);
    CHECK(mkdir);
    CHECK(mount);
    CHECK(mkdir);
    CHECK(mount);
}
END_SETUP()

// Add tests here.

FN_SETUP(cleanup)
{
    CHECK(umount);
    CHECK(rmdir);
    CHECK(umount);
    CHECK(rmdir);
}
```

I don't see purposes to repeat the code in every test.

test/initramfs/src/regression/fs/mount/listmount.c
```diff
@@ -0,0 +1,265 @@
...
+{
+	TEST_SUCC(unshare(CLONE_NEWNS));
+	ensure_dir("/tmp/lm_parent");
+	TEST_SUCC(mount("tmpfs", "/tmp/lm_parent", "tmpfs", 0, NULL));
+	ensure_dir("/tmp/lm_parent/child");
+	TEST_SUCC(mount("tmpfs", "/tmp/lm_parent/child", "tmpfs", 0, NULL));
+
+	uint64_t parent_uid = mount_unique_id("/tmp/lm_parent");
+	uint64_t child_uid = mount_unique_id("/tmp/lm_parent/child");
+	SKIP_TEST_IF(parent_uid == 0 || child_uid == 0);
```
lrh2000: Why not:
```c
uin64_t parent_uid = TEST_RES(..., ret != 0);
```

kernel/src/syscall/listmount.rs
```diff
@@ -0,0 +1,122 @@
...
+        return_errno_with_message!(Errno::EINVAL, "mnt_id is not a valid unique mount ID");
+    }
+    if req.param != 0 && req.param < MNT_UNIQUE_ID_MIN {
+        return_errno_with_message!(Errno::EINVAL, "param is not a valid unique mount ID");
+    }
+
+    let mnt_ns: Arc<MountNamespace> = {
+        let proxy = ctx.thread_local.borrow_ns_proxy();
+        proxy.unwrap().mnt_ns().clone()
+    };
```
lrh2000: Don't see why we need this clone. Try:
```suggestion
    let ns_proxy = ctx.thread_local.borrow_ns_proxy();
    let mnt_ns = ns_proxy.unwrap().mnt_ns();
```

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -75,29 +74,64 @@ pub enum MountPropType {
     // TODO: Implement other propagation types.
 }
 
-static ID_ALLOCATOR: Once<SpinLock<IdAlloc>> = Once::new();
+/// 32-bit recyclable mount IDs.
+///
+/// IDs start at 1; 0 is never issued and denotes an invalid mount.
+/// Exhaustion returns `ENOMEM`.
+// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L282>
```
lrh2000:
```suggestion
/// Reference: <https://elixir.bootlin.com/linux/v6.17/source/fs/namespace.c#L282>
```

Be consistent with the code below.

test/initramfs/src/regression/fs/mount/listmount.c
```diff
@@ -0,0 +1,265 @@
...
+
+	struct linux_mnt_id_req req = {
+		.size = MNT_ID_REQ_SIZE_VER0,
+		.mnt_id = parent_uid,
+	};
+	uint64_t ids[16];
+	long n = TEST_RES(syscall(SYS_listmount, &req, ids, 16, 0), _ret == 1);
+	if (n == 1) {
```
lrh2000:
```suggestion
```

test/initramfs/src/regression/fs/run_test.sh
```diff
@@ -124,6 +124,7 @@ echo "All mount bind file test passed."
 
 ./mount/mount_api
 ./mount/mount_move
+./mount/listmount
```
lrh2000:
```suggestion
./mount/listmount
./mount/mount_api
./mount/mount_move
```

Keep the list sorted.

avasummer:
@boterinas ready

lrh2000 (commented):
Mostly minor style comments.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -75,29 +74,64 @@ pub enum MountPropType {
...
+    unique_id: u64,
+}
+
+impl MountId {
+    /// Allocates a new pair of mount identifiers.
+    ///
+    /// Returns `None` if the recyclable ID range is exhausted.
+    pub(super) fn alloc() -> Option<Self> {
+        let recyclable_id = MOUNT_ID_ALLOCATOR.lock().alloc()?;
+        let unique_id = NEXT_FREE_UNIQUE_ID.fetch_add(1, Ordering::Relaxed);
```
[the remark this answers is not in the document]
avasummer: We intentionally follow Linux here and do not add overflow handling for a practically unreachable case.

avasummer:
Force-pushed.

kernel/libs/sparse-id-alloc/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+            min,
+            max,
+            allocated: BTreeSet::new(),
+        }
+    }
+
+    /// Allocates and returns a new ID from the configured range.
+    ///
+    /// Returns `None` when every ID in `[min, max]` is in use.
+    pub fn alloc(&mut self) -> Option<u32> {
```
tatetian, line 49: The complexity of this `alloc` method is obviously sub-optimal. So I instructed AI to come up with a more efficient implementation.

---

The current `alloc` scans the allocated set from `min` until it finds a gap. Since the policy is smallest-first, live IDs tend to form a dense prefix, so each `alloc` is O(live) and filling the range with n IDs costs O(n²).

I'd suggest changing the representation from individual allocated IDs to *maximal runs* of allocated IDs:

- `allocated_runs: BTreeMap<u32, u32>` — keyed by run start, value = inclusive run end. Runs are kept disjoint and non-adjacent (always fully coalesced).

With smallest-first allocation, the smallest free ID is derivable from the first run alone: if the map is empty or the first run starts above `min`, it's `min`; otherwise it's one past the first run's end (or the range is exhausted). So `alloc` never scans.

```rust
pub struct SparseIdAlloc {
    min: u32,
    max: u32,
    /// Maximal runs of allocated IDs: start -> inclusive end.
    /// Runs are disjoint and non-adjacent.
    allocated_runs: BTreeMap<u32, u32>,
}

impl SparseIdAlloc {
    pub const fn new(min: u32, max: u32) -> Self {
        assert!(min <= max, "min must be <= max");
        Self {
            min,
            max,
            allocated_runs: BTreeMap::new(),
        }
    }

    pub fn alloc(&mut self) -> Option<u32> {
        let first = self.allocated_runs.first_key_value().map(|(&s, &e)| (s, e));
        match first {
            // The range's lowest IDs are allocated; the smallest free ID
            // is just past the first run (unless the range is exhausted).
            Some((start, end)) if start == self.min => {
                if end == self.max {
                    return None;
                }
                let id = end + 1;
                let new_end = id
                    .checked_add(1)
                    .and_then(|next_start| self.allocated_runs.remove(&next_start))
                    .unwrap_or(id);
                *self.allocated_runs.get_mut(&start).unwrap() = new_end;
                Some(id)
            }
            // `min` itself is free.
            _ => {
                let id = self.min;
                let end = id
                    .checked_add(1)
                    .and_then(|next_start| self.allocated_runs.remove(&next_start))
                    .unwrap_or(id);
                self.allocated_runs.insert(id, end);
                Some(id)
            }
        }
    }

    pub fn free(&mut self, id: u32) {
        let run = self
            .allocated_runs
            .range(..=id)
            .next_back()
            .map(|(&s, &e)| (s, e));
        let Some((start, end)) = run.filter(|&(_, end)| id <= end) else {
            panic!("free({id}) of unallocated ID");
        };

        if id == start {
            self.allocated_runs.remove(&start);
        } else {
            *self.allocated_runs.get_mut(&start).unwrap() = id - 1;
        }
        if id < end {
            self.allocated_runs.insert(id + 1, end);
        }
    }
}
```

In both `alloc` arms, the `checked_add(1)` + `remove` handles coalescing: if a run begins right after the newly allocated ID, it's absorbed. In `free`, the four cases (whole run / at start / at end / middle) collapse into two independent statements: trim or remove the left part, reinsert the right part if any.

Properties:

- **Complexity.** `alloc` and `free` are both O(log R), where R is the number of allocated runs — i.e., the fragmentation of the ID space, not the number of live IDs.
- **Memory.** O(R) instead of O(live). Fully coalesced: a dense prefix of a million live IDs is a single map entry.
- **Canonical state.** The map is a pure function of the currently-allocated set — no history-dependent bookkeeping. Freeing everything returns the allocator to its pristine empty state, and `new` stays a zero-allocation `const fn`.
- **Panic contract preserved and unified.** Every kind of invalid free — never issued, out of range, double free — is the same condition: no run contains `id`. One containment check.
- **Overflow edges covered.** `end + 1` is guarded by the `end == max` exhaustion check; the two coalescing `+1`s use `checked_add`; `id - 1` and `id + 1` in `free` are safe by the case analysis (`id > start` and `id < end` respectively). Verified at `max == u32::MAX`.

tatetian (approved):

PR #3264, opened 2026-05-22
Limit some ptrace regression tests to compile only on x86

vvvvsv:
I did not restrict some ptrace regression tests to only be built on x86, which caused [some CI failures](https://github.com/asterinas/asterinas/actions/runs/26280931134/job/77360302868).

This PR fixes the issue and has been tested with a [dry run](https://github.com/vvvvsv/asterinas/actions/runs/26302908930/job/77432429380).


tatetian (approved):

Issue #264, opened 2023-06-19
Support binding/listening/connecting at 0.0.0.0


StevenJiang1110:
The current implemention cannot deal with 0.0.0.0, which means matching any existing ipv4 addr in the system.

PR #3268, opened 2026-05-23
Add IPv6 dual-stack and IPV6_V6ONLY support for TCP (StreamSocket) [TCP only]

yydawx:
Add the foundational IPv6 dual-stack layer and IPV6_V6ONLY socket option
for TCP connections via StreamSocket. 

### Option infrastructure
- Register `SOL_IPV6` level and `IPV6_V6ONLY` (value 26 per Linux in6.h)
- `Ipv6OptionSet` with get/set for v6only, default `true`
- IPv6 and IP options coexist on the same AF_INET6 socket

### Address validation (`SocketFamily`)
- `normalize_endpoint` — bare IPv4 → IPv4-mapped IPv6 for dual-stack
- `validate_endpoint` — v6only check + family check; skips unspecified addresses
- `unmap_ipv4_addr` — mapped → bare for smoltcp (idempotent)
- `resolve_remote_endpoint` — `::` → `::1`, `0.0.0.0` → `127.0.0.1`
- `present_to_user` — bare IPv4 → mapped presentation per RFC 4038
- Replace `From<IpAddress>` trait with `from_raw_addr` (footgun documented)

### StreamSocket integration
- `prepare_endpoint` and `present_addr` delegating to `SocketFamily`
- `InitStream` runtime validation replaced with `debug_assert`
- `unmap_ipv4_addr` before `ConnectingStream` in connect path
- `resolve_bind_iface_and_config` and `get_ephemeral_endpoint` unmapped internally

### Tests (10 scenarios)
- v6only default=1, toggle get/set, AF_INET rejection
- v6only=1 rejects bind to mapped address and connect with bare IPv4
- Dual-stack bind/listen/connect/accept with mapped getpeername
- `connect(::ffff:0.0.0.0)` resolution to loopback
- Dual-stack bind to bare IPv4 address
- Dual-stack broadcast connect without SO_BROADCAST

## What's not included
- DatagramSocket (UDP) dual-stack — deferred to follow-up PR

### DIFF
--- a/kernel/src/net/socket/ip/addr.rs
+++ b/kernel/src/net/socket/ip/addr.rs
@@ -58,9 +58,100 @@ impl IpAddressFamily {
     }
 }
 
-// Note: This does not handle IPv4-mapped IPv6 addresses. When `IPV6_V6ONLY` is set,
-// IPv4-mapped addresses are not permitted, so this function cannot be used to determine
-// whether such an address is acceptable.
+/// Returns `true` if the address is an IPv4-mapped IPv6 address (`::ffff:x.x.x.x`).
+pub(super) fn is_ipv4_mapped(addr: IpAddress) -> bool {
+    matches!(addr, IpAddress::Ipv6(v6) if v6.to_ipv4_mapped().is_some())
+}
+
+/// Maps a bare IPv4 endpoint to an IPv4-mapped IPv6 [`SocketAddr`].
+///
+/// Native IPv6 endpoints pass through unchanged.
+// Used by `present_to_user` to present dual-stack addresses
+// to the user in IPv4-mapped IPv6 form per RFC 4038.
+pub(super) fn ipv4_to_ipv4_mapped(endpoint: IpEndpoint) -> SocketAddr {
+    if let IpAddress::Ipv4(ipv4) = endpoint.addr {
+        let mapped = IpAddress::Ipv6(ipv4.to_ipv6_mapped());
+        return SocketAddr::from(IpEndpoint::new(mapped, endpoint.port));
+    }
+    SocketAddr::from(endpoint)
+}
+
+/// Strips the IPv4-mapped prefix (`::ffff:x.x.x.x` → `x.x.x.x`).
+///
+/// Native IPv4 and native IPv6 pass through unchanged.
+/// Must be called before handing an address to smoltcp, which does not
+/// understand mapped addresses. Idempotent — safe to call defensively.
+pub(crate) fn unmap_ipv4_addr(addr: IpAddress) -> IpAddress {
+    match addr {
+        IpAddress::Ipv6(addr) => match addr.to_ipv4_mapped() {
+            Some(ipv4) => IpAddress::Ipv4(ipv4),
+            None => IpAddress::Ipv6(addr),
+        },
+        other => other,
+    }
+}
+
+/// Normalizes an endpoint for a socket's address family.
+///
+/// In dual-stack mode (IPv6 + !v6only), maps bare IPv4 addresses to
+/// IPv4-mapped IPv6 so the socket layer can process them uniformly.
+pub(super) fn normalize_endpoint(
+    family: IpAddressFamily,
+    v6only: bool,
+    endpoint: IpEndpoint,
+) -> IpEndpoint {
+    if family == IpAddressFamily::IPv6
+        && !v6only
+        && let IpAddress::Ipv4(ipv4) = endpoint.addr
+    {
+        return IpEndpoint::new(IpAddress::Ipv6(ipv4.to_ipv6_mapped()), endpoint.port);
+    }
+    endpoint
+}
+
+/// Normalizes and validates the endpoint for a socket's address family.
+/// Returns `Err` if the endpoint is incompatible with the socket.
+pub(super) fn validate_endpoint(
+    family: IpAddressFamily,
+    v6only: bool,
+    endpoint: IpEndpoint,
+) -> Result<IpEndpoint> {
+    let endpoint = normalize_endpoint(family, v6only, endpoint);
+
+    if is_ipv4_mapped(endpoint.addr) && v6only {
+        return_errno_with_message!(
+            Errno::EAFNOSUPPORT,
+            "IPv4-mapped IPv6 addresses are not allowed when IPV6_V6ONLY is set"
+        );
+    }
+
+    if IpAddressFamily::from(endpoint.addr) != family {
+        return_errno_with_message!(
+            Errno::EAFNOSUPPORT,
+            "the protocol family does not match the address family"
+        );
+    }
+
+    Ok(endpoint)
+}
+
+/// Presents a stored endpoint to the user per RFC 4038.
+///
+/// For AF_INET6 sockets, bare IPv4 addresses are mapped to IPv4-mapped IPv6 form.
+/// The `IPV6_V6ONLY` setting is intentionally **ignored** — per RFC 4038,
+/// `getsockname`/`getpeername` always present dual-stack addresses in mapped form
+/// regardless of the `IPV6_V6ONLY` socket option.
+pub(super) fn present_to_user(family: IpAddressFamily, endpoint: IpEndpoint) -> SocketAddr {
+    if family == IpAddressFamily::IPv6 && matches!(endpoint.addr, IpAddress::Ipv4(_)) {
+        return ipv4_to_ipv4_mapped(endpoint);
+    }
+    SocketAddr::from(endpoint)
+}
+
+// Note: This does not handle IPv4-mapped IPv6 addresses — it returns `IPv6`
+// for them. Callers that need to reject IPv4-mapped addresses when
+// `IPV6_V6ONLY` is set must combine this with an explicit `is_ipv4_mapped`
+// check (see `validate_endpoint`).
 impl From<IpAddress> for IpAddressFamily {
     fn from(addr: IpAddress) -> Self {
         match addr {
--- a/kernel/src/net/socket/ip/options.rs
+++ b/kernel/src/net/socket/ip/options.rs
@@ -110,6 +110,7 @@ impl_socket_options!(
     pub struct Ttl(IpTtl);
     pub struct Hdrincl(bool);
     pub struct Recverr(bool);
+    pub struct Ipv6V6only(bool);
 );
 
 #[derive(Clone, Copy, Debug)]
--- a/kernel/src/net/socket/ip/stream/init.rs
+++ b/kernel/src/net/socket/ip/stream/init.rs
@@ -7,7 +7,10 @@ use core::{
 
 use aster_bigtcp::{socket::RawTcpOption, wire::IpEndpoint};
 
-use super::{connecting::ConnectingStream, listen::ListenStream, observer::StreamObserver};
+use super::{
+    super::unmap_ipv4_addr, connecting::ConnectingStream, listen::ListenStream,
+    observer::StreamObserver,
+};
 use crate::{
     events::IoEvents,
     net::{
@@ -80,18 +83,17 @@ impl InitStream {
         }
     }
 
-    /// Returns the address family of this socket.
-    pub(super) fn family(&self) -> IpAddressFamily {
-        self.family
-    }
-
     pub(super) fn bind(&mut self, endpoint: &IpEndpoint, can_reuse: bool) -> Result<()> {
         if self.bound_port.is_some() {
             return_errno_with_message!(Errno::EINVAL, "the socket is already bound to an address");
         }
 
-        // When we support `IPV6_V6ONLY` and if it is set, we should also reject IPv4-mapped
-        // IPv6 addresses.
+        // The endpoint must already be validated by the caller
+        // (`StreamSocket::bind` validates under the state write lock).
+        debug_assert!(
+            IpAddressFamily::from(endpoint.addr) == self.family,
+            "bind: endpoint family mismatch (caller must validate first)"
+        );
         if IpAddressFamily::from(endpoint.addr) != self.family {
             return_errno_with_message!(
                 Errno::EAFNOSUPPORT,
@@ -120,8 +122,10 @@ impl InitStream {
             "`finish_last_connect()` should be called before calling `connect()`"
         );
 
-        // When we support `IPV6_V6ONLY` and if it is set, we should also reject IPv4-mapped
-        // IPv6 addresses.
+        debug_assert!(
+            IpAddressFamily::from(remote_endpoint.addr) == self.family,
+            "connect: endpoint family mismatch (caller must validate first)"
+        );
         if IpAddressFamily::from(remote_endpoint.addr) != self.family {
             return Err((
                 Error::with_message(
@@ -135,6 +139,8 @@ impl InitStream {
         let bound_port = if let Some(bound_port) = self.bound_port {
             bound_port
         } else {
+            // `get_ephemeral_endpoint` internally calls `unmap_ipv4_addr`,
+            // so it handles both bare IPv4 and IPv4-mapped IPv6 correctly.
             let endpoint = match get_ephemeral_endpoint(remote_endpoint) {
                 Some(ep) => ep,
                 None => {
@@ -153,7 +159,14 @@ impl InitStream {
             }
         };
 
-        ConnectingStream::new(bound_port, *remote_endpoint, option, observer).map_err(
+        // `ConnectingStream` expects a bare IPv4 address (mapped addresses are
+        // not understood by the low-level network stack). The caller in
+        // `StreamSocket::start_connect` already validated the address.
+        // We unmap here to ensure the low-level stack always sees native addresses.
+        let remote_endpoint =
+            IpEndpoint::new(unmap_ipv4_addr(remote_endpoint.addr), remote_endpoint.port);
+
+        ConnectingStream::new(bound_port, remote_endpoint, option, observer).map_err(
             |(err, bound_port)| {
                 if err.error() == Errno::ECONNREFUSED {
                     (err, InitStream::new_refused(bound_port, self.family))
--- a/kernel/src/net/socket/ip/stream/mod.rs
+++ b/kernel/src/net/socket/ip/stream/mod.rs
@@ -21,7 +21,8 @@ use takeable::Takeable;
 use util::{Retrans, TcpOptionSet};
 
 use super::{
-    addr::IpAddressFamily,
+    addr::{IpAddressFamily, present_to_user, validate_endpoint},
+    ipv6_options::Ipv6OptionSet,
     options::{IpOptionSet, SetIpLevelOption},
 };
 use crate::{
@@ -62,6 +63,7 @@ pub struct StreamSocket {
     state: RwLock<Takeable<State>>,
     options: RwLock<OptionSet>,
 
+    family: IpAddressFamily,
     is_nonblocking: AtomicBool,
     pollee: Pollee,
     pseudo_path: Path,
@@ -82,6 +84,7 @@ enum State {
 struct OptionSet {
     socket: SocketOptionSet,
     ip: IpOptionSet,
+    ipv6: Ipv6OptionSet,
     tcp: TcpOptionSet,
 }
 
@@ -89,8 +92,14 @@ impl OptionSet {
     fn new() -> Self {
         let socket = SocketOptionSet::new_tcp();
         let ip = IpOptionSet::new_tcp();
+        let ipv6 = Ipv6OptionSet::new();
         let tcp = TcpOptionSet::new();
-        OptionSet { socket, ip, tcp }
+        OptionSet {
+            socket,
+            ip,
+            ipv6,
+            tcp,
+        }
     }
 
     fn raw(&self) -> RawTcpOption {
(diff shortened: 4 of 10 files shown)


yydawx:
Hi! I think this PR is ready to be reviewed

yydawx:
It's better to support tcp and udp at the same time, I'm working on it

StevenJiang1110:
It looks like you are still actively working on this — thanks again for your hard work!

> It's better to support tcp and udp at the same time, I'm working on it

It seems this PR has already grown to over 1,300 lines of changes. There is no need to pack too many features into a single PR. To make reviewing more manageable, we generally recommend keeping each PR self-contained, well-scoped, and minimal — focusing on the smallest coherent unit of change. This can significantly speed up the review process: a 100-line PR is much easier to review than a 1,000-line one, especially when reviewers have limited bandwidth, which can cause the PR to remain blocked for an extended period. 

It would be best to keep the scope of this PR more focused and limited. If you have additional changes in progress, consider deferring them to follow-up PRs.

yydawx:
> It looks like you are still actively working on this — thanks again for your hard work!
> 
> > It's better to support tcp and udp at the same time, I'm working on it
> 
> It seems this PR has already grown to over 1,300 lines of changes. There is no need to pack too many features into a single PR. To make reviewing more manageable, we generally recommend keeping each PR self-contained, well-scoped, and minimal — focusing on the smallest coherent unit of change. This can significantly speed up the review process: a 100-line PR is much easier to review than a 1,000-line one, especially when reviewers have limited bandwidth, which can cause the PR to remain blocked for an extended period.
> 
> It would be best to keep the scope of this PR more focused and limited. If you have additional changes in progress, consider deferring them to follow-up PRs.

Ok! I'm departing them rn.
Congratulation for publishing 0.18!

yydawx:
@StevenJiang1110 it's small and clean now. Ready for review

StevenJiang1110:
> @StevenJiang1110 it's small and clean now. Ready for review

To make this PR self-contained, the newly-added socket option should at minimum be functional. When adding a new socket option, you need to support at least the following:

1. The user can call `setsockopt(sockfd, SOL_IPV6, IPV6_ONLY)` or `getsockopt(...)` to set or retrieve the value.
2. The option should actually take effect where supported. For example, if `IPV6_ONLY` is enabled, binding to an IPv4-mapped IPv6 address should be rejected.

Some regression tests are also required. You can refer to the existing `sockoption.c` as an example.

One additional question: should IPv6 options and IP options coexist? That is, for an `AF_INET6` stream socket, should it be possible to call both `getsockopt(sockfd, SOL_IPV6, ...)` and `getsockopt(sockfd, SOL_IP, ...)` on the same socket?

yydawx:
Hi! @StevenJiang1110 I undersood what you said. This PR only implement ipv4-mapped-ipv6 infra and tcp-dual stack functions.And strong tests I think.

yydawx:
> > @StevenJiang1110 it's small and clean now. Ready for review
> 
> To make this PR self-contained, the newly-added socket option should at minimum be functional. When adding a new socket option, you need to support at least the following:
> 
> 1. The user can call `setsockopt(sockfd, SOL_IPV6, IPV6_ONLY)` or `getsockopt(...)` to set or retrieve the value.
> 2. The option should actually take effect where supported. For example, if `IPV6_ONLY` is enabled, binding to an IPv4-mapped IPv6 address should be rejected.
> 
> Some regression tests are also required. You can refer to the existing `sockoption.c` as an example.
> 
> One additional question: should IPv6 options and IP options coexist? That is, for an `AF_INET6` stream socket, should it be possible to call both `getsockopt(sockfd, SOL_IPV6, ...)` and `getsockopt(sockfd, SOL_IP, ...)` on the same socket?

Regarding coexistence: yes, IP and IPv6 options already coexist. On the same AF_INET6 socket, getsockopt(sk, SOL_IP, IP_TTL) hits the IP layer directly, while getsockopt(sk, SOL_IPV6, IPV6_V6ONLY) falls through IP to the IPv6 layer. They're dispatched independently, matching Linux behavior — no extra work needed.
Shell I add a test on this issue?

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -460,12 +498,17 @@ impl Socket for StreamSocket {
...
+        let remote_endpoint = self.prepare_endpoint(remote_endpoint)?;
+        debug!("connect: after prepare = {:?}", remote_endpoint);
+        let remote_endpoint = resolve_remote_endpoint(self.family, remote_endpoint);
+        debug!("connect: after resolve = {:?}", remote_endpoint);
         if let Some(result) = self.start_connect(&remote_endpoint) {
+            debug!("connect: start_connect returned Err({:?})", result);
             return result;
         }
-
+        debug!("connect: blocking wait");
```
StevenJiang1110: Remove all these debug log.

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -712,15 +763,17 @@ impl Socket for StreamSocket {
 
         let need_iface_poll = match socket_option_result {
             Err(err) if err.error() == Errno::ENOPROTOOPT => {
-                // Deal with IP-level options
```
StevenJiang1110, line 715: Why remove this comment? Some for `// Deal with TCP-level options`. I think these comments should be kept.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,11 +58,155 @@ impl IpAddressFamily {
...
 }
 
-// Note: This does not handle IPv4-mapped IPv6 addresses. When `IPV6_V6ONLY` is set,
-// IPv4-mapped addresses are not permitted, so this function cannot be used to determine
-// whether such an address is acceptable.
-impl From<IpAddress> for IpAddressFamily {
-    fn from(addr: IpAddress) -> Self {
+/// Returns `true` if the address is an IPv4-mapped IPv6 address (`::ffff:x.x.x.x`).
+pub(super) fn is_ipv4_mapped(addr: IpAddress) -> bool {
+    matches!(addr, IpAddress::Ipv6(v6) if v6 != Ipv6Address::UNSPECIFIED && v6.to_ipv4_mapped().is_some())
```
StevenJiang1110: An unspecified IPv6 address cannot simultaneously be an IPv4-mapped address, so the unspecified check is redundant. The condition should be simplified to:

```rust
matches!(addr, IpAddress::Ipv6(v6) if v6.is_ipv4_mapped())
```

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,11 +58,155 @@ impl IpAddressFamily {
...
+    ///
+    /// # Footgun
+    ///
+    /// IPv4-mapped IPv6 addresses (`::ffff:x.x.x.x`) return [`IpAddressFamily::IPv6`],
+    /// even though they represent embedded IPv4 addresses. Callers that use this for
+    /// routing, binding, or broadcast decisions **must** call [`unmap_ipv4_addr`] first.
+    /// For socket-level address validation, use [`SocketFamily::validate_endpoint`]
+    /// instead — it combines this check with `IPV6_V6ONLY` handling.
+    #[must_use]
+    pub fn from_raw_addr(addr: IpAddress) -> Self {
```
StevenJiang1110: Since we already have `impl From<IpAddress> for IpAddressFamily`, why introduce this method? It also fails to handle the IPv6-only case. The existing `From` trait implementation is equally sufficient, so this method is unnecessary.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,11 +58,155 @@ impl IpAddressFamily {
...
+        family: IpAddressFamily,
+        v6only: bool,
+        endpoint: IpEndpoint,
+    ) -> Result<IpEndpoint> {
+        Self::with_v6only(family, v6only).validate_endpoint(endpoint)
+    }
+}
+/// Resolves a remote endpoint for connect.
+///
+/// An unspecified address (`0.0.0.0` or `::`) as a destination is not
```
StevenJiang1110: This should not be handled here. At least for now, connecting, binding, and listening on an unspecified address (0.0.0.0) is not supported — see #264.

If we do want to support unspecified addresses in the future, this is not the right place to handle it. I would prefer this PR to leave unspecified address handling out of scope. Note that it's not just this location — I see several other places in this PR that also attempt to handle unspecified addresses. This is a separate concern that I would suggest deferring to a dedicated PR, keeping the current behavior intact for now.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,11 +58,155 @@ impl IpAddressFamily {
...
+    /// Normalizes and validates the endpoint for this socket's address family.
+    /// Returns `Err` if the endpoint is incompatible with the socket.
+    pub fn validate_endpoint(&self, endpoint: IpEndpoint) -> Result<IpEndpoint> {
+        let endpoint = self.normalize_endpoint(endpoint);
+
+        let effective = unmap_ipv4_addr(endpoint.addr);
+        let is_unspecified = match effective {
+            IpAddress::Ipv4(addr) => addr == Ipv4Address::UNSPECIFIED,
+            IpAddress::Ipv6(addr) => addr == Ipv6Address::UNSPECIFIED,
+        };
```
StevenJiang1110: Similarly, don't handle unspecified address here. It looks very strange.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,11 +58,155 @@ impl IpAddressFamily {
...
+                    "the protocol family does not match the address family"
+                );
+            }
+        }
+
+        Ok(endpoint)
+    }
+
+    /// Convenience wrapper for `with_v6only` + `validate_endpoint`.
+    pub fn prepare_endpoint(
```
StevenJiang1110: The current `SocketFamily` struct is a poor abstraction. First, it bundles two seemingly unrelated concerns into a single struct. Second, the method names on this struct are quite confusing — for instance,`prepare_endpoint` is difficult to understand at first glance, and it is unclear why `present_to_user` is defined on `SocketFamily` at all, given that the method doesn't even use the v6only field.

I would suggest removing this struct entirely. You could instead define a few standalone helper functions with more descriptive names. For example, `prepare` is a poor name.

test/initramfs/src/regression/network/ipv6_v6only.c
```diff
@@ -0,0 +1,262 @@
...
+
+	val = 1;
+	TEST_SUCC(setsockopt(sk, IPPROTO_IPV6, IPV6_V6ONLY, &val, sizeof(val)));
+
+	val = 0;
+	len = sizeof(val);
+	TEST_RES(getsockopt(sk, IPPROTO_IPV6, IPV6_V6ONLY, &val, &len),
+		 val == 1);
+
+	close(sk);
```
StevenJiang1110:
```suggestion
	TEST_SUCC(close(sk));
```

Add `TEST_SUCC`/`TEST_RES` for syscall/library call in `FN_TEST` if possible.

test/initramfs/src/regression/network/ipv6_v6only.c
```diff
@@ -0,0 +1,262 @@
...
+	TEST_RES(getsockopt(sk, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, &len),
+		 v6only == 1);
+
+	close(sk);
+}
+END_TEST()
+
+FN_TEST(v6only_toggle)
+{
+	int sk = CHECK(socket(AF_INET6, SOCK_STREAM, 0));
```
StevenJiang1110:
```suggestion
	int sk = TEST_SUCC(socket(AF_INET6, SOCK_STREAM, 0));
```

In `FN_TEST`, always prefer `TEST_SUCC` than `CHECK`, if possible.

kernel/src/net/socket/ip/options.rs
```diff
@@ -105,11 +105,53 @@ impl IpOptionSet {
     }
 }
 
+/// IPv6-level socket options.
+#[derive(Clone, Copy, CopyGetters, Debug, Setters)]
+#[get_copy = "pub"]
+#[set = "pub"]
+pub(super) struct Ipv6OptionSet {
```
StevenJiang1110: Move this struct to a separate module, for example: `kernel/src/net/socket/ip/ipv6_options.rs`

StevenJiang1110 (commented):
Thanks for your contribution!

> Regarding coexistence: yes, IP and IPv6 options already coexist. On the same AF_INET6 socket, getsockopt(sk, SOL_IP, IP_TTL) hits the IP layer directly, while getsockopt(sk, SOL_IPV6, IPV6_V6ONLY) falls through IP to the IPv6 layer. They're dispatched independently, matching Linux behavior — no extra work needed.
Shell I add a test on this issue?

Yes, feel free to add a test to confirm the behavior. My main reason for asking was to better understand how the API should be designed in a principled way.

tatetian:
@boterinas author

boterinas:

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -427,6 +444,27 @@ impl StreamSocket {
...
+        let v6only = if self.family == IpAddressFamily::IPv6 {
+            self.options.read().ipv6.v6only()
+        } else {
+            false
+        };
+        validate_endpoint(self.family, v6only, endpoint)
+    }
+
+    /// Presents a stored endpoint to the user per RFC 4038.
+    fn present_addr(&self, endpoint: IpEndpoint) -> SocketAddr {
```
StevenJiang1110, line 458: This thin wrapper method is meaningless to me. Just remove it.

yydawx:
## Changes

- Inlined endpoint validation into `bind()` and `start_connect()`
  under `state.write()` → `options.read()`, making validation atomic
  with the operation. Removed the standalone `prepare_endpoint`.
- Unified lock order to `state → options`, consistent with
  `setsockopt`, eliminating the TOCTOU window.
- `InitStream::bind()` / `connect()` keep both runtime `EAFNOSUPPORT`
  and `debug_assert!` for defense-in-depth.
- `IPV6_V6ONLY` defaults to `false` (matching Linux
  `net.ipv6.bindv6only=0`).

yydawx:
@boterinas  ready

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,9 +58,100 @@ impl IpAddressFamily {
...
+pub(super) fn is_ipv4_mapped(addr: IpAddress) -> bool {
+    matches!(addr, IpAddress::Ipv6(v6) if v6.to_ipv4_mapped().is_some())
+}
+
+/// Maps a bare IPv4 endpoint to an IPv4-mapped IPv6 [`SocketAddr`].
+///
+/// Native IPv6 endpoints pass through unchanged.
+// Used by `present_to_user` to present dual-stack addresses
+// to the user in IPv4-mapped IPv6 form per RFC 4038.
+pub(super) fn ipv4_to_ipv4_mapped(endpoint: IpEndpoint) -> SocketAddr {
```
StevenJiang1110, line 71: This method serves the same purpose as `NormalizedAddress` in aster-bigtcp — normalizing an IP address. I think it would be cleaner to make `NormalizedAddress` public and reuse it here instead of duplicating the logic.

If so, also remember to adjust the document on `NormalizedAddress`.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,9 +58,100 @@ impl IpAddressFamily {
...
+    }
+    SocketAddr::from(endpoint)
+}
+
+/// Strips the IPv4-mapped prefix (`::ffff:x.x.x.x` → `x.x.x.x`).
+///
+/// Native IPv4 and native IPv6 pass through unchanged.
+/// Must be called before handing an address to smoltcp, which does not
+/// understand mapped addresses. Idempotent — safe to call defensively.
+pub(crate) fn unmap_ipv4_addr(addr: IpAddress) -> IpAddress {
```
StevenJiang1110, line 84: `pub(crate)` is not suggested to use in `kernel` crate. Further, `unmap` is a very confusing name, you can directly name this method as `to_ipv4_mapped`.

kernel/src/util/net/options/ipv6.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use int_to_c_enum::TryFromInt;
+
+use super::{RawSocketOption, SocketOption, impl_raw_socket_option};
+use crate::{net::socket::ip::options::Ipv6V6only, prelude::*};
+
+/// Socket options for IPv6 socket.
+///
+/// The raw definitions can be found at:
+/// <https://elixir.bootlin.com/linux/v6.0.19/source/include/uapi/linux/in6.h#L170>.
```
StevenJiang1110, line 11: The link points to an unexpected location. Please manually verify all external links generated by AI.

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -427,6 +453,11 @@ impl StreamSocket {
         self.pollee.invalidate();
         error
     }
+
+    /// Presents a stored endpoint to the user per RFC 4038.
+    fn present_addr(&self, endpoint: IpEndpoint) -> SocketAddr {
+        present_to_user(self.family, endpoint)
```
StevenJiang1110, line 459: `present_addr`/`present_to_user` is a very vague name. Depending on the actual semantics, consider renaming it to something more descriptive like `to_ipv4_mapped` or `normalized`, which I think would be much clearer.

kernel/src/net/socket/ip/options.rs
```diff
@@ -110,6 +110,7 @@ impl_socket_options!(
     pub struct Ttl(IpTtl);
     pub struct Hdrincl(bool);
     pub struct Recverr(bool);
+    pub struct Ipv6V6only(bool);
```
StevenJiang1110, line 113: Please move this to a separate module for IPv6-specific options, as they should not be mixed with other IP-level options.

Also, please rename this option to V6only, dropping the Ipv6 prefix.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,9 +58,100 @@ impl IpAddressFamily {
...
+///
+/// For AF_INET6 sockets, bare IPv4 addresses are mapped to IPv4-mapped IPv6 form.
+/// The `IPV6_V6ONLY` setting is intentionally **ignored** — per RFC 4038,
+/// `getsockname`/`getpeername` always present dual-stack addresses in mapped form
+/// regardless of the `IPV6_V6ONLY` socket option.
+pub(super) fn present_to_user(family: IpAddressFamily, endpoint: IpEndpoint) -> SocketAddr {
+    if family == IpAddressFamily::IPv6 && matches!(endpoint.addr, IpAddress::Ipv4(_)) {
+        return ipv4_to_ipv4_mapped(endpoint);
+    }
+    SocketAddr::from(endpoint)
```
StevenJiang1110, line 148: Why does this function only handle the address on the AF_INET6 socket side? For the other side, shouldn't the address on the AF_INET socket also be converted to an IPv4 address?

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,9 +58,100 @@ impl IpAddressFamily {
...
+        && let IpAddress::Ipv4(ipv4) = endpoint.addr
+    {
+        return IpEndpoint::new(IpAddress::Ipv6(ipv4.to_ipv6_mapped()), endpoint.port);
+    }
+    endpoint
+}
+
+/// Normalizes and validates the endpoint for a socket's address family.
+/// Returns `Err` if the endpoint is incompatible with the socket.
+pub(super) fn validate_endpoint(
```
StevenJiang1110, line 114: It is not clear to me why this method is named `validate_endpoint`. Its primary purpose appears to be normalization, so I think it should be renamed to `normalize_endpoint` instead. The original `normalize_endpoint` should then be inlined into it.

kernel/src/net/socket/ip/addr.rs
```diff
@@ -58,9 +58,100 @@ impl IpAddressFamily {
...
+            "IPv4-mapped IPv6 addresses are not allowed when IPV6_V6ONLY is set"
+        );
+    }
+
+    if IpAddressFamily::from(endpoint.addr) != family {
+        return_errno_with_message!(
+            Errno::EAFNOSUPPORT,
+            "the protocol family does not match the address family"
+        );
+    }
```
StevenJiang1110, line 133: Should these two check be performed before `normalize_endpoint`?

kernel/src/net/socket/ip/stream/init.rs
```diff
@@ -153,7 +159,14 @@ impl InitStream {
             }
         };
 
-        ConnectingStream::new(bound_port, *remote_endpoint, option, observer).map_err(
+        // `ConnectingStream` expects a bare IPv4 address (mapped addresses are
+        // not understood by the low-level network stack). The caller in
```
StevenJiang1110, line 163: Do you mean that two IPv4-mapped IPv6 addresses cannot be handled by smoltcp(i.e., build connection)? Could you provide a reference to the relevant smoltcp documentation or implementation code to support this?

StevenJiang1110 (commented):
This is not a thorough review.

The current code is quite difficult to follow, so I may have misunderstood some parts of it. First and foremost, I would like you to establish consistent terminology. For example, the conversion from an IPv4 address to an IPv6 address should be referred to as "normalize", in line with our existing code — please stick with this term and avoid using other verbs. The conversion from an IPv6 address to an IPv4 address should be called to_ipv4_mapped. I think this would make things much clearer.

Also, please try to reuse the existing conversion functions as much as possible, rather than introducing similar new helpers.

StevenJiang1110:
@boterinas author

PR #3269, opened 2026-05-24
Concatenate multiple `global_asm!` statements in rv trap

zjp-CN:
solve https://github.com/asterinas/asterinas/pull/3247#discussion_r3265750466

Since the github link to a specific PR discussion is broken, here's what was written:

For the failed [riscv compile](https://github.com/asterinas/asterinas/actions/runs/26088170843/job/76706582443?pr=3247) job, multiple global_asm! statements are not guaranteed to be concatenated continuously, on the contrary, it's [discouraged](https://dirname.github.io/rust-std-doc/unstable-book/library-features/global-asm.html) to share states:

> You may use global_asm! multiple times, anywhere in your crate, in whatever way suits you. However, you should not rely on assembler state (e.g. assembler macros) defined in one global_asm! to be available in another one. It is implementation-defined whether the multiple usages are concatenated into one or assembled separately.

This seems to be a good blind spot in low-level crates.  (e.g. https://github.com/rust-embedded/riscv/issues/321)

### DIFF
--- a/ostd/src/arch/riscv/trap/trap.rs
+++ b/ostd/src/arch/riscv/trap/trap.rs
@@ -9,6 +9,7 @@
 //
 // We make the following new changes:
 // * Implement the `trap_handler` of Asterinas.
+// * Concatenate multiple `global_asm!` statements.
 //
 // These changes are released under the following license:
 //
@@ -21,8 +22,8 @@ use crate::arch::cpu::{
     extension::{IsaExtensions, has_extensions},
 };
 
-#[cfg(target_arch = "riscv32")]
 global_asm!(
+    #[cfg(target_arch = "riscv32")]
     r"
     .equ XLENB, 4
     .macro LOAD_SP a1, a2
@@ -31,10 +32,8 @@ global_asm!(
     .macro STORE_SP a1, a2
         sw \a1, \a2*XLENB(sp)
     .endm
-"
-);
-#[cfg(target_arch = "riscv64")]
-global_asm!(
+    ",
+    #[cfg(target_arch = "riscv64")]
     r"
     .equ XLENB, 8
     .macro LOAD_SP a1, a2
@@ -43,7 +42,10 @@ global_asm!(
     .macro STORE_SP a1, a2
         sd \a1, \a2*XLENB(sp)
     .endm
-"
+    ",
+    include_str!("trap.S"),
+    SSTATUS_FS_MASK = const SSTATUS_FS_MASK,
+    SSTATUS_SUM = const SSTATUS_SUM
 );
 
 /// FPU status bits.
@@ -52,8 +54,6 @@ pub(in crate::arch) const SSTATUS_FS_MASK: usize = 0b11 << 13;
 /// Supervisor User Memory access bit.
 /// Reference: <https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#sstatus>.
 pub(in crate::arch) const SSTATUS_SUM: usize = 0b1 << 18;
-
-global_asm!(include_str!("trap.S"), SSTATUS_FS_MASK = const SSTATUS_FS_MASK, SSTATUS_SUM = const SSTATUS_SUM);
 
 /// Initialize interrupt handling for the current HART.
 ///


lrh2000:
At the top of the file:
```patch
 // We make the following new changes:
 // * Implement the `trap_handler` of Asterinas.
+// * Concatenate multiple `global_asm!` statements.
```

though I don't know if it is really necessary.

ostd/src/arch/riscv/trap/trap.rs
```diff
@@ -43,7 +42,9 @@ global_asm!(
     .macro STORE_SP a1, a2
         sd \a1, \a2*XLENB(sp)
     .endm
-"
+    ",
+    include_str!("trap.S"),
+    SSTATUS_FS_MASK = const SSTATUS_FS_MASK, SSTATUS_SUM = const SSTATUS_SUM
```
lrh2000:
```suggestion
    SSTATUS_FS_MASK = const SSTATUS_FS_MASK,
    SSTATUS_SUM = const SSTATUS_SUM,
```

Similar to:
https://github.com/asterinas/asterinas/blob/040707d0f33ddb4c03aa954f101245d4f2907cb2/ostd/src/arch/x86/boot/mod.rs#L31-L33

lrh2000 (commented):
I have had a look at the other files containing multiple `global_asm!`. They seem fine.

ostd/src/arch/riscv/trap/trap.rs
```diff
@@ -31,10 +32,8 @@ global_asm!(
...
     .endm
-"
-);
-#[cfg(target_arch = "riscv64")]
-global_asm!(
+    ",
+    #[cfg(target_arch = "riscv64")]
     r"
     .equ XLENB, 8
     .macro LOAD_SP a1, a2
```
zjp-CN, line 39: Maybe we can do this to avoid macro duplicates:

```rust
global_asm!(
    #[cfg(target_arch = "riscv32")] ".equ XLENB, 4",
    #[cfg(target_arch = "riscv64")] ".equ XLENB, 8",

    r".macro LOAD_SP a1, a2
        lw \a1, \a2*XLENB(sp)
    .endm
    .macro STORE_SP a1, a2
        sw \a1, \a2*XLENB(sp)
    .endm
    ",

    include_str!("trap.S"),
    SSTATUS_FS_MASK = const SSTATUS_FS_MASK,
    SSTATUS_SUM = const SSTATUS_SUM,
)
```

zjp-CN:
I'm also considering to add a guideline for this.

ostd/src/arch/riscv/trap/trap.rs
lrh2000: > Maybe we can do this to avoid macro duplicates:

We don't intend to support `riscv32`. Another option would be to simply remove the `riscv32` support. If we do that, we can move the macro to the assembly file?

(Or we can remove the macro? But it is really aggressive.)

But I don't really have a preference here. So I previously treated this PR as a bug fix PR, without being asked to clean up this file further.

lrh2000: > ```rust
>     r".macro LOAD_SP a1, a2
>         lw \a1, \a2*XLENB(sp)
>     .endm
>     .macro STORE_SP a1, a2
>         sw \a1, \a2*XLENB(sp)
>     .endm
>     ",
> ```

Wait a minute. It does not work. Please pay attention to **`lw`&`sw` AND `ld`&`sd`**.

lrh2000: > I previously treated this PR as a bug fix PR, without being asked to clean up this file further.

This should be still the simplest choice if no one has a strong opinion whether the `riscv32` code should end up with (completely removed or not).

lrh2000:
> I'm also considering to add a guideline for this.

Shouldn't a guideline be added to the official Rust's `global_asm!` reference? Nothing is special to Asterinas here.

ostd/src/arch/riscv/trap/trap.rs
zjp-CN: > Please pay attention to **`lw`&`sw` AND `ld`&`sd`**.

You have sharp eyes 👀 



> This should be still the simplest choice if no one has a strong opinion whether the `riscv32` code should end up with (completely removed or not).

I added the second commit to remove rv32 directives, and also adjusted the PR title.

ostd/src/arch/riscv/trap/trap.rs
```diff
@@ -23,16 +24,6 @@ use crate::arch::cpu::{
...
-    #[cfg(target_arch = "riscv32")]
-    r"
-    .equ XLENB, 4
-    .macro LOAD_SP a1, a2
-        lw \a1, \a2*XLENB(sp)
-    .endm
-    .macro STORE_SP a1, a2
-        sw \a1, \a2*XLENB(sp)
-    .endm
-    ",
```
lrh2000: If we only remove this... It will confuse readers, who will wonder why we have `#[cfg(target_arch = "riscv64")]` below and why we define the macro in this file instead of in `trap.S`.

I don't think it's a proper clean-up. I prefer the original form, which is at least less confusing.

ostd/src/arch/riscv/trap/trap.rs
lrh2000: > I added the second commit to remove rv32 directives, and also adjusted the PR title.

But I don't think I suggested to only remove the rv32 directives?

The problem here is that if we remove rv32 directives, the rv64 directives are also odd. So properly cleaning it up may require more work to be done.


>> We don't intend to support `riscv32`. Another option would be to simply remove the `riscv32` support.  **If we do that, we can move the macro to the assembly file?**
>> 
>> **(Or we can remove the macro? But it is really aggressive.)**
>>
>> [ .. ]
>>
>> **without being asked to clean up this file further [ .. ] should be still the simplest choice** if no one has a strong opinion whether the `riscv32` code should end up with (completely removed or not).

ostd/src/arch/riscv/trap/trap.rs
zjp-CN: Ok. I'll only leave the first commit, and raise another PR to clean up rv32 code.

lrh2000 (approved):

PR #3273, opened 2026-05-25
Add system-core docs and tests for bash, busybox and systemd

cqs21:


tatetian (approved):

PR #3274, opened 2026-05-25
Remove riscv32 trap code

zjp-CN:
This PR
* removes riscv32 code from trap.rs
  * because Asterinas doesn't target riscv32
* moves XLENB, LOAD_SP, and STORE_SP into riscv trap.S

See https://github.com/asterinas/asterinas/pull/3269#discussion_r3296508240

### DIFF
--- a/ostd/src/arch/riscv/trap/trap.S
+++ b/ostd/src/arch/riscv/trap/trap.S
@@ -16,10 +16,14 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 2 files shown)


zjp-CN:
There's also a line for rv32, should I remove the first operand as well?
https://github.com/asterinas/asterinas/blob/e3fd4eee7accd2730875eb04e8e3d08db16c928a/ostd/src/arch/riscv/cpu/extension.rs#L58

tatetian (approved):

PR #3275, opened 2026-05-25
Fix virtio-blk flush feature detection and handling

hburaylee:
Fix two bugs about VIRTIO_BLK_F_FLUSH :

- 1. In `VirtioBlockFeature::new()` , `== 1` should be `!= 0`; since `FLUSH` is bit 9 `support_flush` may always be false.

- 2. The condition in `flush()` that checks whether `VIRTIO_BLK_F_FLUSH ` is supported appears to be written backwards.

### DIFF
--- a/kernel/comps/virtio/src/device/block/mod.rs
+++ b/kernel/comps/virtio/src/device/block/mod.rs
@@ -200,7 +200,7 @@ impl ConfigManager<VirtioBlockConfig> {
 
 impl VirtioBlockFeature {
     pub(self) fn new(transport: &dyn VirtioTransport) -> Self {
-        let support_flush = transport.read_device_features() & BlockFeatures::FLUSH.bits() == 1;
+        let support_flush = (transport.read_device_features() & BlockFeatures::FLUSH.bits()) != 0;
         VirtioBlockFeature { support_flush }
     }
 }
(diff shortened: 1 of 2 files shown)


kernel/comps/virtio/src/device/block/mod.rs
```diff
@@ -200,7 +200,7 @@ impl ConfigManager<VirtioBlockConfig> {
 
 impl VirtioBlockFeature {
     pub(self) fn new(transport: &dyn VirtioTransport) -> Self {
-        let support_flush = transport.read_device_features() & BlockFeatures::FLUSH.bits() == 1;
+        let support_flush = transport.read_device_features() & BlockFeatures::FLUSH.bits() != 0;
```
tatetian: I suggest rewriting this line as

```rust
(transport.read_device_features() & BlockFeatures::FLUSH.bits()) != 0
```

for improved readability.

tatetian (commented):
Thanks for identifying the two bugs!

kernel/comps/virtio/src/device/block/mod.rs
hburaylee: > I suggest rewriting this line as
> 
> ```rust
> (transport.read_device_features() & BlockFeatures::FLUSH.bits()) != 0
> ```
> 
> for improved readability.

Done

tatetian (approved):

PR #3276, opened 2026-05-26
Update supported Linux system calls count to 240

tatetian:


tatetian:
Merging a trivial change without extra review.

PR #3277, opened 2026-05-26
Overlay Codex with a working nixpkgs version

taosue:
The `codex` package from the current NixOS 25.05 package set is too old for our usage. This PR pins Codex to a newer, tested nixpkgs version that works in Asterinas NixOS.


tatetian:
Need to add a CI test for Codex. Any difficulties?

taosue:
> Need to add a CI test for Codex. Any difficulties?

Yes. We do not have an API key that can be provided to CI, so we cannot test an actual Codex request or interactive workflow. The test can only verify that `codex` is installed correctly on Asterinas NixOS and that simple offline commands such as `codex --help` work.

tatetian:
> simple offline commands such as `codex --help` work.

Ok. Got it. Then, let's at the very least run `codex --help` to ensure that `codex` is installed correctly. And this trivial test case is a perfect place to add a comment that explains why we cannot test non-trivial `codex` functions. Without such a comment, people may get the impression that a codex test case is missing.

tatetian (approved):

PR #3278, opened 2026-05-26
Fix some thread name problems

StevenJiang1110:
This PR fixes the process names shown by BusyBox `ps` on Asterinas.

Before this change, after running `make run_kernel`, `ps aux` could show command names wrapped in braces. BusyBox prints `{comm} cmdline` when the process `comm` name does not match the basename of `argv[0]`. In this code path, BusyBox gets `comm` from `/proc/[pid]/stat`(from thread name) and the command line from `/proc/[pid]/cmdline`（by reading from init stack）.

```plain
~ # ps aux
PID   USER     TIME  COMMAND
    1 root      0:00 {init} /bin/sh /init sh -l
    8 root      0:00 {busybox} script /dev/null -q -c sh -l
    9 root      0:00 {busybox} sh -l
   10 root      0:00 {busybox} ps aux
```

The root cause was that Asterinas did not preserve Linux-compatible thread names in several cases:

1. `execve` through a symlink derived the thread name from the resolved path. Linux derives it from the user-supplied exec path before symlink resolution. For `execveat` with `AT_EMPTY_PATH`, there is no user-supplied path, so this PR keeps using the resolved file name.

2. `fork` derived the child thread name from the executable path. It should inherit the parent's thread name instead, otherwise names set by `PR_SET_NAME` are lost.

3. `/proc/[pid]/comm` and `/proc/[pid]/task/[tid]/comm` reported the executable file name. They should report the thread name, so names set by `PR_SET_NAME` remain visible and procfs is consistent with `/proc/[pid]/stat`.

After this PR, `ps aux` shows:

```plain
~ # ps aux
PID   USER     TIME  COMMAND
    1 root      0:00 {init} /bin/sh /init sh -l
    8 root      0:00 script /dev/null -q -c sh -l
    9 root      0:00 sh -l
   10 root      0:00 ps aux
```

The remaining `{init}` is expected. Asterinas boots `/init` via `init=/init` and passes `sh -l` as init arguments. The kernel prepends the init executable path, so the script starts with `/init sh -l`. Since `/init` is a shebang script with `#!/bin/sh`, the loader runs the interpreter with the command line `/bin/sh /init sh -l`.

The thread name, however, is still derived from the executed script path `/init`, so `comm` is `init`. BusyBox compares this with `argv[0]`, whose basename is `sh`, sees that they differ, and intentionally prints `{init} /bin/sh /init sh -l`. This matches normal Linux behavior for shebang scripts and is not the bug fixed by this PR.

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/pid/task/comm.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/comm.rs
@@ -8,6 +8,7 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
+    process::posix_thread::AsPosixThread,
     thread::Thread,
 };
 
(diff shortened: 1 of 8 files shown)


tatetian (approved):

Issue #3239, opened 2026-05-18
Fully implement initial random seed generation


le-monde-bleu:
### Description

At present, Asterinas randomness generation is relatively not robust (at least, for RISC-V). As discussed [here](https://github.com/asterinas/asterinas/pull/3169#discussion_r3182283316). This tracking issue:

- proposes to strengthen the `get_random_seed` implementation for non x86_64 architecture;
- proposes to add a kernel entropy pool and collect runtime entropy outside `OSTD`;
- proposes to fix `random` semantics;
- calls for suggestions about this issue.

### Current Status

- The implementation of `get_random_seed` for x86_64:
https://github.com/asterinas/asterinas/blob/b19581bd65ef28a8644731a68e62566898f64763/kernel/src/util/random.rs#L24-L47
The implementation takes hardware as the only source of random seed. As illustrated in the notes for future factorings, this could be a problem if hardware randomness cannot be generated.
- The implementation of `get_random_seed` for non x86_64:
https://github.com/asterinas/asterinas/blob/b19581bd65ef28a8644731a68e62566898f64763/kernel/src/util/random.rs#L49-L60
The implementation takes device tree as the only source of random seed, and it does not use the implemented hardware entropy accessor in `ostd/src/arch/riscv/mod.rs`. Consequently, the method panics if the device tree is not available.
- The boot issue of RISC-V `sifive_u` mode:
At present, the initialization of `sifive_u` scheme:
`make run_kernel TARGET_ARCH=riscv64 SCHEME=sifive_u`
triggers uncaught panic:
```
[     0.267] ERROR: Uncaught panic:
	called `Option::unwrap()` on a `None` value
	at /root/asterinas/kernel/src/util/random.rs:56
	on CPU 0 by thread None
```
due to the lack of device tree, and fails the boot process.
- `random` semantic problem
At present, `/dev/random` and `/dev/urandom` devices both map to the same implementation alias:
https://github.com/asterinas/asterinas/blob/cdca74f8f06afbd5dd6cbd20bfd4feca8d74a516/kernel/src/device/mem/file.rs#L36-L37
So, there is no true blocking entropy-pool distinction yet, and the `/dev/random` never blocks:
https://github.com/asterinas/asterinas/blob/cdca74f8f06afbd5dd6cbd20bfd4feca8d74a516/kernel/src/syscall/getrandom.rs#L21-L33

### Proposed tasks

- [x] Boot-safe seeding for non x86_64 (necessary)

see #3281 

[...]

PR #3281, opened 2026-05-27
Modify the get_random_seed implementation for non x86_64 arch

le-monde-bleu:
**Description**
This PR modifies the implementation of `get_random_seed` for non x86_64 architectures.
https://github.com/asterinas/asterinas/blob/9fab50ef564ee66a7b50bbd32dddf564831327ef/kernel/src/util/random.rs#L49-L60
At present, `get_random_seed` for non x86_64 takes device tree as the only source of random seed. This PR tries to add the hierarchical sources:
- hardware (`read_random`), see #3169 
- device tree (the current source)
- pseudorandomness (`read_tsc`)
https://github.com/asterinas/asterinas/blob/9fab50ef564ee66a7b50bbd32dddf564831327ef/ostd/src/arch/riscv/mod.rs#L98-L101

 This is also a step in #3239 , which solves the uncaught panic during `sifive_u` scheme initialization.

**Influence**
- The `riscv` scheme initialization now takes hardware as its source of random seed.
- The `sifive_u` scheme initialization now does not trigger uncaught panic due to the lack of `/chosen/rng-seed`

**Note**
The pseudo randomness is not safe enough for the initialization, and is a temporary step for letting the booting process of `sifive_u` scheme run successfully without the uncaught panic. For further refactoring, multiple sources might be considered, e.g. environment noise.

**Alternative**
An alternative implementation is based on function call, which may reduce duplicate variable definition. But I am not quite sure which is better. Any suggestion is welcomed.
```Rust
#[cfg(not(target_arch = "x86_64"))]
fn get_random_seed() -> <StdRng as SeedableRng>::Seed {
    use ostd::arch::{boot::DEVICE_TREE, read_random, read_tsc};

    fn fill_seed(
        seed: &mut <StdRng as SeedableRng>::Seed,
        mut next: impl FnMut() -> Option<u64>,
    ) -> bool {
        let mut chunks = seed.as_mut().chunks_exact_mut(size_of::<u64>());
        for chunk in chunks.by_ref() {
            let Some(src) = next() else {
                return false;
            };
            chunk.copy_from_slice(&src.to_ne_bytes());
        }

        let tail = chunks.into_remainder();
        if !tail.is_empty() {
            let Some(src) = next() else {
                return false;
            };
            tail.copy_from_slice(&src.to_ne_bytes()[..tail.len()]);
        }

        true
    }

    // First, try to fill the seed using hardware based source `read_random()`.
    let mut seed = <StdRng as SeedableRng>::Seed::default();
    if fill_seed(&mut seed, || read_random()) {
        return seed;
    }

    // If hardware based method is not available, try to fill the seed using
    // the device tree.
    if let Some(seed) = DEVICE_TREE
        .get()
        .and_then(|device_tree| device_tree.find_node("/chosen"))
        .and_then(|chosen| chosen.property("rng-seed"))
        .and_then(|rng_seed| <StdRng as SeedableRng>::Seed::try_from(rng_seed.value).ok())
    {
        return seed;
    }

    // If neither hardware based source nor device tree is available, use TSC as a fallback.
    let _ = fill_seed(&mut seed, || Some(read_tsc()));
    seed
}
```


kernel/src/util/random.rs
```diff
@@ -48,13 +48,60 @@ fn get_random_seed() -> <StdRng as SeedableRng>::Seed {
...
+    }
+    if fill_from_read_random {
+        let tail = chunks.into_remainder();
+        if !tail.is_empty() {
+            match read_random() {
+                Some(src) => tail.copy_from_slice(&src.to_ne_bytes()[..tail.len()]),
+                None => fill_from_read_random = false,
+            }
+        }
+    }
```
lrh2000: We can try to write the code in a way that makes it (1) reusable and (2) easier to read. This code snippet replicates `x86_64` version of `get_random_seed` for no good reason.

Try to do the following:
```rust
fn get_random_seed() {
    if let Some(seed) = read_seed_from_hardware() {
        ostd::info!("use randomness generated by the hardware");
        return seed;
    }

    ostd::if_tdx_enabled({
        // If hardware randomness cannot be generated, we should stop if we are
        // TD guests. For more details, see
        // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
        panic!("hardware randomness is mandatory for TD guests");
    });

    #[cfg(not(target_arch = "x86_64"))]
    if let Some(seed) = read_seed_from_device_tree() {
        ostd::info!("use randomness provided by the device tree");
        return seed;
    }

    // Some platforms (e.g., RISC-V `sifive_u`) have neither hardware
    // randomness nor randomness from the device tree. For now, we will
    // continue with pseudorandomness.
    // TODO: Collect entropy from environment noise for better security.
    ostd::warn!("use randomness based on the timestamp, which is insecure");
    read_seed_from_timestamp()
}

fn read_seed_from_hardware();

#[cfg(not(target_arch = "x86_64"))]
fn read_seed_from_device_tree();

fn read_seed_from_timestamp();
```

> An alternative implementation is based on function call, which may reduce duplicate variable definition. But I am not quite sure which is better. Any suggestion is welcomed.

Yeah. I think you have noted that that the structure of `read_seed_from_hardware` and `read_seed_from_timestamp` appears to be similar. If so, adding a new method to avoid repeating the same logic in two separate methods is certainly better. But perhaps it should just be something like `read_seed_from(ostd::arch::read_random)` and `read_seed_from(ostd::arch::read_tsc)`...

lrh2000 (commented):
Thank you for enhancing `get_random_seed` logic!

le-monde-bleu:
**Changes**
Now the implementation is unified for x86_64 and non x86_64 arch. The changes are:
- The `get_random_seed()` is unified for x86_64 and non x86_64 arch;
- Isolate the implementation of `read_seed_from_hardware()` for x86_64 and non x86_64 archs (seperated), which uses the hardware source for random seed;
- Isolate the implementation of `read_seed_from_device_ tree()` for non x86_64 arch, which uses the device tree for random seed;
- Isolate the implementation of `read_seed_from_timestamp()`, which is pseudo-randomness.
- Seperate the common logic of random seed generation and compact it into one common helper function `read_seed_from()`, so that `read_seed_from_hardware()` (for both archs, by calling `read_seed_from(read_random())`) and `read_seed_from_device_tree()` (by calling `read_seed_from(read_tsc())`) can utilize it for simplicity and reduce duplication.

**Current Situation**
To be noted, this causes a change in the logic of random seed generation for x86_64 arch. Previously, it uses only hardware as the source. Now, the process is aborted if no hardware and running inside TDX, and a pseudo randomness fallback (which uses `read_tsc()`) is added (as a shared fallback for all archs). 

So the logic for x86_64 arch is like:
- hardware;
- abort if no hardware and TDX;
- pseudo randomness fallback.

While the logic for non x86_64 arch is like:
- hardware;
- device tree;
- pseudo randomness fallback.

**Notes**
- The use of `ostd::info!` or maybe `ostd::warn!` seems to trigger makefile error during x86_64 booting, I am not quite sure why. So I use `ostd::early_println!` instead, for now, which eliminates the problem.
- The CI test of x86_64 boot-test(legacy32, linux-legacy32, 4) seem to occur in different PRs, which does not seem to be the code's problem. @lrh2000

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
...
+
+    read_seed_from(read_random)
+}
+
+#[cfg(not(target_arch = "x86_64"))]
+fn read_seed_from_hardware() -> Option<<StdRng as SeedableRng>::Seed> {
     use ostd::arch::read_random;
 
+    read_seed_from(read_random)
+}
```
lrh2000: Aren't the two `read_seed_from_hardware()` identical? Remove `#[cfg(...)]` and one implementation.

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
...
+        // TD guests. For more details, see
+        // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
+        ostd::if_tdx_enabled!({
+            panic!("hardware randomness is mandatory for TD guests");
+        });
+    }
+
+    // Non-x86_64: fall back to device-tree seed if present.
+    #[cfg(not(target_arch = "x86_64"))]
+    {
```
lrh2000: I think we should remove unnecessary indent.
Use this version:
```rust
    #[cfg(not(target_arch = "x86_64"))]
    if let Some(seed) = read_seed_from_device_tree() {
        // ...
    }
```

Also:
```rust
    ostd::if_tdx_enabled!({
        // If hardware randomness cannot be generated, we should stop if we are
        // TD guests. For more details, see
        // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
        panic!("...");
    });
```

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
...
+            None => return None,
+        }
     }
     let tail = chunks.into_remainder();
-    let n = tail.len();
-    if n > 0 {
-        let src = read_random().expect("`read_random` failed").to_ne_bytes();
-        tail.copy_from_slice(&src[..n]);
+    if !tail.is_empty() {
+        match next_random() {
```
lrh2000: For simplicity, use:
```rust
        let val = next_random()?;
``` 

Same for the previous `match read_random()`.

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
...
+    read_seed_from(read_random)
+}
+
+#[cfg(not(target_arch = "x86_64"))]
+fn read_seed_from_device_tree() -> Option<<StdRng as SeedableRng>::Seed> {
+    use ostd::arch::boot::DEVICE_TREE;
+
+    DEVICE_TREE
+        .get()
+        .and_then(|device_tree| device_tree.find_node("/chosen"))
```
lrh2000: `DEVICE_TREE` is currently assumed to be present for non-x86 archs.
```
> grep -R "DEVICE_TREE.*unwrap" kernel
kernel/comps/time/src/rtc/loongson.rs:        let chosen = DEVICE_TREE.get().unwrap().find_node("/rtc").unwrap();
kernel/comps/uart/src/arch/riscv/mod.rs:    let device_tree = DEVICE_TREE.get().unwrap();
kernel/comps/virtio/src/transport/mmio/bus/arch/riscv.rs:    let device_tree = DEVICE_TREE.get().unwrap();
kernel/src/arch/loongarch/power.rs:    let device_tree = DEVICE_TREE.get().unwrap();
kernel/src/util/random.rs:    let chosen = DEVICE_TREE.get().unwrap().find_node("/chosen").unwrap();
```

To align with existing code, prefer:
```rust
DEVICE_TREE
    .get()
    .unwrap()
    .find_node("/chosen")
    .and_then(/* ... */)
```

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
     RNG.call_once(|| SpinLock::new(StdRng::from_seed(seed)));
 }
 
-#[cfg(target_arch = "x86_64")]
 fn get_random_seed() -> <StdRng as SeedableRng>::Seed {
+    // Prefer hardware RNG if available.
+    if let Some(seed) = read_seed_from_hardware() {
+        ostd::early_println!("use randomness generated by hardware");
```
lrh2000: > > * The use of `ostd::info!` or maybe `ostd::warn!` seems to trigger makefile error during x86_64 booting, I am not quite sure why

What is a Makefile error? Do you mean compile-time or runtime errors? Could you please specify the exact error?

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
     RNG.call_once(|| SpinLock::new(StdRng::from_seed(seed)));
 }
 
-#[cfg(target_arch = "x86_64")]
 fn get_random_seed() -> <StdRng as SeedableRng>::Seed {
+    // Prefer hardware RNG if available.
```
lrh2000:
```suggestion
    // Prefer the hardware RNG if available.
```

kernel/src/util/random.rs
```diff
@@ -21,40 +21,91 @@ pub fn init() {
...
+    {
+        // If hardware randomness cannot be generated, we should stop if we are
+        // TD guests. For more details, see
+        // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
+        ostd::if_tdx_enabled!({
+            panic!("hardware randomness is mandatory for TD guests");
+        });
+    }
+
+    // Non-x86_64: fall back to device-tree seed if present.
```
lrh2000:
```suggestion
    // Non-x86_64: fall back to the device-tree seed if present.
```

lrh2000 (commented):
Thanks for revising the code. I just have some minor comments.

Also, I don't know what you mean by saying a Makefile error. Could you please be more specific?

kernel/src/util/random.rs
le-monde-bleu: Yes, indeed. I will fix this.

le-monde-bleu:
I made the fixes. By the way, the error disappeared now. I guess it was due to some other occasional problems. Now I switch back to `ostd::info!` and `ostd::warn!` and everything is fine.

kernel/src/util/random.rs
```diff
@@ -21,40 +21,76 @@ pub fn init() {
...
+    }
 
-    // Notes for future refactorings: If hardware randomness cannot be generated (i.e., if
-    // `read_random` fails), we can usually continue with pseudorandomness. However, we should stop
-    // if we are TD guests. For more details, see
+    // x86_64: if no hardware RNG and running inside TDX, abort.
+    #[cfg(target_arch = "x86_64")]
+    // If hardware randomness cannot be generated, we should stop if we are
+    // TD guests. For more details, see
     // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
```
lrh2000, line 34: It seems odd to insert a `cfg` attribute between two paragraphs of comments.

I think we can either move the comment to the `if` body, or move this above. Use either of the two versions:
 - Version 1:
```rust
    // x86_64: if no hardware RNG and running inside TDX, abort.
    ostd::if_tdx_enabled!({
        // If hardware randomness cannot be generated, we should stop if we are
        // TD guests. For more details, see
        // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
        panic!("hardware randomness is mandatory for TD guests");
    });
```
 - Version 2:
```rust
    // x86_64: if no hardware RNG and running inside TDX, abort.
    //
    // If hardware randomness cannot be generated, we should stop if we are
    // TD guests. For more details, see
    // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
    ostd::if_tdx_enabled!({
        panic!("hardware randomness is mandatory for TD guests");
    });
```

In fact, the comments seem a bit repetitive if we use the second version. We should simplify it to:
```rust
    // x86_64: if no hardware RNG and running inside TDX, abort.
    //
    // For more details, see
    // <https://intel.github.io/ccc-linux-guest-hardening-docs/security-spec.html#randomness-inside-tdx-guest>.
    ostd::if_tdx_enabled!({
        panic!("hardware randomness is mandatory for TD guests");
    });
```

lrh2000 (commented):
Thanks for your updates. Just a nit.

kernel/src/util/random.rs
le-monde-bleu: I suppose so, have made an update.

lrh2000 (approved):

PR #3282, opened 2026-05-27
Fix some UDP problems

lrh2000:
This PR resolves some bugs that were encountered during UDP speed testing with `iperf3`.

 1. The first bug was that we incorrectly checked the number of bytes against the number of packets.
https://github.com/asterinas/asterinas/blob/9fab50ef564ee66a7b50bbd32dddf564831327ef/kernel/libs/aster-bigtcp/src/socket/bound/udp.rs#L150

It should be:
```diff
-        if size > socket.packet_send_capacity() {
+        if size > socket.payload_send_capacity() {
```

 2. The second is to resolve the following FIXME item:
https://github.com/asterinas/asterinas/blob/9fab50ef564ee66a7b50bbd32dddf564831327ef/kernel/libs/aster-bigtcp/src/iface/common.rs#L314-L315

This PR introduces the `PortKey` struct:
```rust
pub struct IfaceCommon<E: Ext> {
    used_ports: SpinLock<BTreeMap<PortKey, PortState>, BottomHalfDisabled>,
    // ..
}

#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
struct PortKey {
    addr: NormalizedAddress,
    port: u16,
    protocol: PortProtocol,
}
```

It also introduces the strongly typed bound ports `BoundTcpPort` and `BoundUdpPort` to avoid creating a TCP socket on a UDP port.
```rust
/// A TCP port bound to an iface.
pub struct BoundTcpPort<E: Ext>(BoundPort<E>);
/// A UDP port bound to an iface.
pub struct BoundUdpPort<E: Ext>(BoundPort<E>);
```

`iperf -u -l 1400` can work under `NETDEV=tap` after this PR. There are some other problems:
 - It does not work under `NETDEV=user` because the UDP port forwarding is not configuerd.
 - `-l 1400` is required. Otherwise either the data rate is zero or the kernel will panic. We support neither IPv4 fragments nor virtio-net jumbo packets. 65535-byte Ethernet frames won't work.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/iface/iface.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/iface.rs
@@ -4,7 +4,7 @@ use alloc::sync::Arc;
 
 use smoltcp::wire::{Ipv4Address, Ipv4Cidr, Ipv6Address};
 
-use super::{BoundPort, InterfaceFlags, InterfaceType, port::BindPortConfig};
+use super::{BindPortConfig, BoundTcpPort, BoundUdpPort, InterfaceFlags, InterfaceType};
 use crate::{errors::BindError, ext::Ext};
 
 /// A network interface.
@@ -22,20 +22,32 @@ pub trait Iface<E>: internal::IfaceInternal<E> + Send + Sync {
 }
 
 impl<E: Ext> dyn Iface<E> {
-    /// Binds a socket to the iface.
+    // FIXME: The reason for binding the socket and the iface together is because there are
+    // limitations inside smoltcp. See discussion at
+    // <https://github.com/smoltcp-rs/smoltcp/issues/779>.
+
+    /// Binds a TCP port to the iface.
     ///
-    /// After binding the socket to the iface, the iface will handle all packets to and from the
-    /// socket.
-    ///
-    /// If no specific port is given in [`BindPortConfig`], the iface will pick up an ephemeral port
-    /// for the socket.
-    ///
-    /// FIXME: The reason for binding the socket and the iface together is because there are
-    /// limitations inside smoltcp. See discussion at
-    /// <https://github.com/smoltcp-rs/smoltcp/issues/779>.
-    pub fn bind(self: &Arc<Self>, config: BindPortConfig) -> Result<BoundPort<E>, BindError> {
+    /// If no specific port is given in [`BindPortConfig`], the iface will pick up an ephemeral
+    /// port.
+    pub fn bind_tcp(
+        self: &Arc<Self>,
+        config: BindPortConfig,
+    ) -> Result<BoundTcpPort<E>, BindError> {
         let common = self.common();
-        common.bind(self.clone(), config)
+        common.bind_tcp(self.clone(), config)
+    }
+
+    /// Binds a UDP port to the iface.
+    ///
+    /// If no specific port is given in [`BindPortConfig`], the iface will pick up an ephemeral
+    /// port.
+    pub fn bind_udp(
+        self: &Arc<Self>,
+        config: BindPortConfig,
+    ) -> Result<BoundUdpPort<E>, BindError> {
+        let common = self.common();
+        common.bind_udp(self.clone(), config)
     }
 
     /// Returns the interface index.
--- a/kernel/src/net/socket/ip/common.rs
+++ b/kernel/src/net/socket/ip/common.rs
@@ -8,7 +8,7 @@ use aster_bigtcp::{
 
 use crate::{
     net::{
-        iface::{BoundPort, Iface, iter_all_ifaces, loopback_iface, virtio_iface},
+        iface::{Iface, iter_all_ifaces, loopback_iface, virtio_iface},
         socket::util::check_port_privilege,
     },
     prelude::*,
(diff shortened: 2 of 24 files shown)


test/initramfs/src/regression/network/udp_err.c
```diff
@@ -420,3 +420,45 @@ FN_TEST(self_connect)
...
+	TEST_SUCC(
+		connect(sender, (struct sockaddr *)&sk_addr, sizeof(sk_addr)));
+
+	memset(send_buf, 'x', sizeof(send_buf));
+	TEST_RES(send(sender, send_buf, sizeof(send_buf), 0),
+		 _ret == sizeof(send_buf));
+
+	memset(receive_buf, 'y', sizeof(receive_buf));
+	TEST_RES(recv(receiver, receive_buf, sizeof(receive_buf), 0),
+		 _ret == sizeof(receive_buf));
```
StevenJiang1110, line 441: Calling `recv` immediately after `send` may return nothing if the data transfer has not yet completed. Consider either putting the socket into blocking mode or calling `poll` before `recv` to ensure data is available.

kernel/src/net/socket/ip/common.rs
```diff
@@ -69,7 +69,10 @@ fn get_ephemeral_iface(remote_ip_addr: &IpAddress) -> Arc<Iface> {
     }
 }
 
-pub(super) fn bind_port(endpoint: &IpEndpoint, can_reuse: bool) -> Result<BoundPort> {
+pub(super) fn prepare_bind_port(
```
StevenJiang1110: Perhaps `resolve_iface_for_bind` is a better name?

kernel/libs/aster-bigtcp/src/iface/iface.rs
```diff
@@ -22,20 +22,32 @@ pub trait Iface<E>: internal::IfaceInternal<E> + Send + Sync {
...
-    /// After binding the socket to the iface, the iface will handle all packets to and from the
-    /// socket.
+    // FIXME: The reason for binding the socket and the iface together is because there are
+    // limitations inside smoltcp. See discussion at
+    // <https://github.com/smoltcp-rs/smoltcp/issues/779>.
+
+    /// Binds a TCP port to the iface.
     ///
     /// If no specific port is given in [`BindPortConfig`], the iface will pick up an ephemeral port
     /// for the socket.
```
StevenJiang1110:
```suggestion
    /// Binds a TCP port to the iface.
   ///
   /// If no specific port is given in [`BindPortConfig`], the iface will pick up an ephemeral port.
```

Since you change "Bind a socket" to "Binds a TCP port", perhaps we should remove "for the socket" too.

StevenJiang1110 (approved):

kernel/src/net/socket/ip/common.rs
lrh2000: There is already a `get_iface_to_bind` in this file. Doesn't `resolve_iface_for_bind` sound too similar and difficult to express the difference between the two?

But `get_iface_to_bind` isn't public anymore. So maybe the concern is indeed minimal.

I renamed this method to `resolve_bind_iface_and_config`. While it still does not express that it also checks port privileges, perhaps this name suggests that it performs more actions than the `get_iface_to_bind` method.

lrh2000:
Fix another minor problem in the first commit... See the range diff below.

`run_test.sh` should be sorted alphabetically, and `nanosleep_err` should be able to test more bits.

<details>
<summary> range-diff </summary>

```
1:  5b7360358 ! 1:  c51ac079a Adjust comments and visibility
    @@ kernel/src/process/execve.rs: fn do_execve_no_return(
      
          // Unshare and reset signal dispositions to their default actions.
     
    + ## test/initramfs/src/regression/process/clock_nanosleep/nanosleep_err.c ##
    +@@ test/initramfs/src/regression/process/clock_nanosleep/nanosleep_err.c: FN_TEST(clock_nanosleep_unknown_flag_bit)
    + {
    + 	struct timespec req = { .tv_sec = 0, .tv_nsec = 0 };
    + 
    +-	// `flags = 2` is neither 0 nor `TIMER_ABSTIME`. Used to trigger
    +-	// `unreachable!()` and panic the kernel.
    +-	TEST_SUCC(syscall(SYS_clock_nanosleep, CLOCK_MONOTONIC, 2, &req, NULL));
    ++	// The flag (`0xdeadbeef`) is neither 0 nor `TIMER_ABSTIME`.
    ++	// Linux will ignore unknown flag bits.
    ++	TEST_SUCC(syscall(SYS_clock_nanosleep, CLOCK_MONOTONIC, 0xdeadbeef,
    ++			  &req, NULL));
    + }
    + END_TEST()
    +
      ## test/initramfs/src/regression/process/execve/execve_comm.c ##
     @@ test/initramfs/src/regression/process/execve/execve_comm.c: FN_TEST(execve_symlink)
      	int status;
    @@ test/initramfs/src/regression/process/prctl/thread_name.c: FN_TEST(fork_inherits
      	}
      
      	TEST_RES(waitpid(pid, &status, 0),
    +
    + ## test/initramfs/src/regression/process/run_test.sh ##
    +@@
    + 
    + set -e
    + 
    +-./cgroup.sh
    ++./clock_nanosleep/nanosleep_err
    + 
    + ./clone3/clone_exit_signal
    + ./clone3/clone_files
    +@@ test/initramfs/src/regression/process/run_test.sh: set -e
    + 
    + ./cpu_affinity/cpu_affinity
    + 
    +-./clock_nanosleep/nanosleep_err
    +-
    + ./execve/execve
    + ./execve/execve_comm
    + ./execve/execve_err
    +@@ test/initramfs/src/regression/process/run_test.sh: if [ "$(uname -m)" = "x86_64" ]; then
    +     ./signal/sigtrap
    + fi
    + 
    ++./cgroup.sh
    + ./group_session
    + ./job_control
    + ./pidfd
2:  1c9aecc9c = 2:  538cba939 Strongly type `NormalizedAddress`
3:  d1ae1026f = 3:  3ed83f7b6 Fix large UDP `send()`
4:  42c519004 = 4:  52ce042c4 Bind to TCP and UDP ports separately
```

</details>

cqs21 (approved):

PR #3285, opened 2026-05-27
Unify `virtio-net` features in QEMU cmdline

lrh2000:
The three configurations differ from each other.
https://github.com/asterinas/asterinas/blob/82c2ff94e11a504f245c7adf07a4b882946a4310/test/initramfs/src/benchmark/bench_linux_and_aster.sh#L147
https://github.com/asterinas/asterinas/blob/82c2ff94e11a504f245c7adf07a4b882946a4310/tools/qemu_args.sh#L47
https://github.com/asterinas/asterinas/blob/82c2ff94e11a504f245c7adf07a4b882946a4310/tools/qemu_args.sh#L53

From what I can see, the difference is:
```patch
diff --git a/virtio-net.benchmark b/virtio-net.tap
index 0a86623..448bb60 100644
--- a/virtio-net.benchmark
+++ b/virtio-net.tap
@@ -17,5 +17,6 @@ host_ecn=off
 host_tso4=off
 host_tso6=off
 host_ufo=off
+host_ufo=off
 host_uso=off
 mrg_rxbuf=off
```
```patch
diff --git a/virtio-net.benchmark b/virtio-net.user
index 0a86623..4cfe0ba 100644
--- a/virtio-net.benchmark
+++ b/virtio-net.user
@@ -1,21 +1,11 @@
-csum=off
 ctrl_guest_offloads=off
 ctrl_mac_addr=off
 ctrl_rx_extra=off
 ctrl_rx=off
 ctrl_vlan=off
 ctrl_vq=off
+event_idx=off
 guest_announce=off
-guest_csum=off
-guest_ecn=off
-guest_tso4=off
-guest_tso6=off
-guest_ufo=off
-guest_uso4=off
-guest_uso6=off
-host_ecn=off
-host_tso4=off
-host_tso6=off
-host_ufo=off
-host_uso=off
+indirect_desc=off
 mrg_rxbuf=off
+queue_reset=off
```

One real problem is that we don't specify `indirect_desc`/`event_idx` in `bench_linux_and_aster.sh`, meaning that Linux will try to enable and use that feature. However, we don't support it. Therefore, the comparison may not be very fair.

(We do disable those features consistently when using `virtio-blk-pci`: `disable-legacy=on,disable-modern=off,queue-size=64,num-queues=1,request-merging=off,backend_defaults=off,disctestingard=off,write-zeroes=off,event_idx=off,indirect_desc=off,queue_reset=off`.)
https://github.com/asterinas/asterinas/blob/30bd5a173eb7f6d84a35fd1133413ab795f83cc9/tools/qemu_args.sh#L175
https://github.com/asterinas/asterinas/blob/30bd5a173eb7f6d84a35fd1133413ab795f83cc9/test/initramfs/src/benchmark/bench_linux_and_aster.sh#L146

This PR combines and unifies the existing (disabled) virtio-net features. I think this will cause less confusion in future.

### DIFF
--- a/test/initramfs/src/benchmark/bench_linux_and_aster.sh
+++ b/test/initramfs/src/benchmark/bench_linux_and_aster.sh
@@ -144,7 +144,7 @@ run_benchmark() {
         -initrd "${BENCHMARK_ROOT}/../../build/initramfs.cpio.gz"
         -drive "if=none,format=raw,id=x0,file=${BENCHMARK_ROOT}/../../build/ext2.img"
         -device "virtio-blk-pci,bus=pcie.0,addr=0x6,drive=x0,serial=vext2,disable-legacy=on,disable-modern=off,queue-size=64,num-queues=1,request-merging=off,backend_defaults=off,discard=off,write-zeroes=off,event_idx=off,indirect_desc=off,queue_reset=off"
-        -device "virtio-net-pci,netdev=net01,disable-legacy=on,disable-modern=off,csum=off,guest_csum=off,ctrl_guest_offloads=off,guest_tso4=off,guest_tso6=off,guest_ecn=off,guest_ufo=off,host_tso4=off,host_tso6=off,host_ecn=off,mrg_rxbuf=off,ctrl_vq=off,ctrl_rx=off,ctrl_vlan=off,ctrl_rx_extra=off,guest_announce=off,ctrl_mac_addr=off,host_ufo=off,guest_uso4=off,guest_uso6=off,host_uso=off"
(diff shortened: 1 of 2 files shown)


martinholovsky:
CI is green here. I ran `tools/qemu_args.sh` from this branch and from its parent and diffed the emitted `-device virtio-net-pci,...` strings, in case an independent check helps this along.

The unification is an exact set union — 24 features, the union of the three originals, with nothing added or dropped. Per call site:

```
NETDEV=user   14 -> 27 properties   +13
NETDEV=tap    24 -> 27 properties    +3
benchmark     24 -> 27 properties    +3
```

Nothing is removed anywhere, and no value changes.

The `+13` on the user path looks larger than it is. In QEMU 10.2.1 (`hw/net/virtio-net.c:3090-3118`) those exact thirteen — `CSUM`, `GUEST_CSUM`, `GUEST_TSO4`, `GUEST_TSO6`, `HOST_TSO4`, `HOST_TSO6`, `GUEST_ECN`, `HOST_ECN`, `GUEST_UFO`, `HOST_UFO`, `GUEST_USO4`, `GUEST_USO6`, `HOST_USO` — are cleared whenever the peer has no vnet_hdr, and the `user` backend never declares one (`net/slirp.c:171`: `net_slirp_info` has no `.has_vnet_hdr` member, so `qemu_has_vnet_hdr()` returns false at `net/net.c:534`). They were already unavailable under SLIRP, so setting them `=off` changes the command line rather than the negotiated feature set.

The `+3` is the real change: `event_idx`, `indirect_desc` and `queue_reset` are transport features rather than offloads, so peer capability does not clear them. That is the benchmark fairness fix described here, and it now applies to the tap path as well.

Minor: the unified list is nearly alphabetical but not quite — `ctrl_rx_extra` precedes `ctrl_rx`, and `indirect_desc` sits between `guest_announce` and `guest_csum` rather than after `host_uso`.

tatetian (approved):

PR #3286, opened 2026-05-27
Add Firefox to the verified web browsers

cqs21:
This PR contains two changes:
1. Add x86-64 support for `ARCH_SET_GS` and `ARCH_GET_GS`, and align user FS/GS base handling with the task-switch and user-entry paths.
2. Add Firefox to the verified web browsers list in the Asterinas NixOS documentation and test suite.


ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -258,6 +277,44 @@ impl UserContext {
...
+    /// On x86-64, the kernel runs with its own `gsbase` and returns to user mode via `swapgs`.
+    /// Therefore the user `gsbase` must be staged in `IA32_KERNEL_GS_BASE` while the kernel is
+    /// running.
+    pub fn activate_gs_pointer(&self) {
+        KernelGsBase::write(VirtAddr::new(self.gsbase() as u64));
+    }
+
+    /// Sets whether the arch context needs to be loaded.
+    pub fn set_needs_arch_context_load(&mut self, needs_load: bool) {
+        self.needs_arch_context_load.set(needs_load);
```
lrh2000: Why are you introducing this? I don't think it's necessary. We can activate the FS/GS pointers immediately.

The existing pattern is to manually call `activate_tls_pointer` when the TLS pointer changes and before executing a new task. I think we should follow this pattern instead of creating a new one.

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -258,6 +277,44 @@ impl UserContext {
...
+    }
+
+    /// Synchronizes the saved FSGSBASE pointers with the current CPU state.
+    ///
+    /// This must be called after returning from user mode because userspace may update either
+    /// base directly through the FSGSBASE instructions.
+    pub fn sync_fsgsbase_pointers_from_cpu(&mut self) {
+        // SAFETY: The kernel enables FSGSBASE during boot and uses these instructions in its
+        // context-switch path. Reading the current task's inactive user FS/GS base does not affect
+        // kernel execution.
```
lrh2000: Be consistent with the existing code (either fix this or fix the existing SAFETY comments as well):
```rust
        // SAFETY: Setting `fsbase` won't affect kernel code.
```

 - "The kernel enables FSGSBASE during boot" This is okay, but we may want to have the same in the existing code.
 - "uses these instructions in its context-switch path" This should be removed. It is unrelated at all.
 - "Reading the current task's inactive user FS/GS base does not affect kernel execution." This can be simplified.

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -258,6 +277,44 @@ impl UserContext {
...
+    /// Sets whether the arch context needs to be loaded.
+    pub fn set_needs_arch_context_load(&mut self, needs_load: bool) {
+        self.needs_arch_context_load.set(needs_load);
+    }
+
+    /// Synchronizes the saved FSGSBASE pointers with the current CPU state.
+    ///
+    /// This must be called after returning from user mode because userspace may update either
+    /// base directly through the FSGSBASE instructions.
+    pub fn sync_fsgsbase_pointers_from_cpu(&mut self) {
```
lrh2000: This does not need to be `pub`. Also it fixes a separate bug. Better to be a separate commit and with some a regression test...

kernel/src/syscall/arch_prctl.rs
```diff
@@ -9,9 +9,11 @@ use crate::prelude::*;
 #[repr(u64)]
 #[derive(Debug, TryFromInt)]
 enum ArchPrctlCode {
+    #[cfg(target_arch = "x86_64")]
     ARCH_SET_GS = 0x1001,
     ARCH_SET_FS = 0x1002,
     ARCH_GET_FS = 0x1003,
+    #[cfg(target_arch = "x86_64")]
```
lrh2000: In fact, the entire file may need to be gated to `x86_64`?

lrh2000 (commented):
Thanks for adding support for `ARCH_SET_GS`.

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -299,6 +308,7 @@ impl UserContextApiInternal for UserContext {
         loop {
             crate::task::scheduler::might_preempt();
             self.user_context.run();
+            self.sync_fsgsbase_pointers_from_cpu();
```
lrh2000: On second thought, I still find this a little odd.

There are two types of register:
 1. Registers are saved and restored during syscall enter and exit.
 2. Registers are saved and restored during context switches.

Which type of register are the GS and FS bases?
 - If performance is not a concern, the first option can be chosen for simplicity.
 - However, given that we currently choose the second option, it is a bit odd to save them during syscall enter. Most syscalls do not access the saved values, after all.

So, I want to know if we want the following:

```rust
/// Activates per-task userspace state to the current task.
///
/// On x86, the TLS pointer and the GS base are per-task states that are saved
/// and restored during context switches. This method should be called after
/// setting them via [`Self::set_tls_pointer`], [`Self::set_gs_pointer`], or
/// [`Self::general_regs_mut`], before the next [`Self::execute`], and on the
/// same task as the next [`Self::execute`].
///
/// Otherwise, [`Self::set_tls_pointer`] and other setters will not take
/// effect, and [`Self::execute`] will run the userspace program with outdated
/// states.
pub fn activate_task_state() {}

/// Synchronizes per-task userspace state from the current task.
///
/// On x86, the TLS pointer and the GS base are per-task states that are saved
/// and restored during context switches. This method should be called before
/// getting them via [`Self::tls_pointer`], [`Self::gs_pointer`], or
/// [`Self::general_regs`], after the last [`Self::execute`], and on the same
/// task as the last [`Self::execute`].
///
/// Otherwise, [`Self::tls_pointer`] and other getters may return outdated
/// states because the userspace program can change them without entering the
/// kernel.
pub fn sync_task_state() {}
```

For other architectures:
```rust
/// Activates per-task userspace state to the current task.
///
/// On RISC-V, there is no per-task userspace state and this method does nothing.
pub fn activate_task_state() {}

/// Synchronizes per-task userspace state from the current task.
///
/// On RISC-V, there is no per-task userspace state and this method does nothing.
pub fn sync_task_state() {}
```

`sync_task_state()` (or `activate_task_state()`) should only be called if we want to read (or write) the userspace FS/GS bases. For example, this would be necessary for `arch_prctl`, signals, and ptrace.

---

I have also considered other alternatives. One of them is similar to the previous version in this PR.
```rust
pub fn set_tls_pointer(&mut self) {
    self.need_activate_tls = true;
    self.set_fsbase(base);
}

pub fn execute(&mut self) {
    if self.need_activate_tls {
       self.activate_tls();
    }

    // Enter userspace
}
```

This works really well for setters. However, I can't figure out how to make it work well for getters.
```rust
pub fn tls_pointer(&self) {
     if !self.has_sync_tls {
         // What should we do here?
     }
     self.fsbase()
}
```

If the TLS pointer is automatically synchronized in cases where it isn't, the following code will not work properly:
```rust
let user_context_1 = UserContext::new();
let user_context_2 = UserContext::new();

user_context_1.execute();
user_context_2.execute();

// This will wrongly save the TLS pointer
// of `user_context_2` to `user_context_1`.
dbg!(user_context_1.tls_pointer());
```

So `sync_task_state()` must be an explicit operation. But if it has to be an explicit operation, maybe we should just use the proposed `sync_task_state()`/`activate_task_state()` approach.

---

Do you have any thoughts on this? Requiring the caller to manually invoke `sync`/`activate` is certainly an additional burden, and it's quite easy to forget. I believe the recent PRs about `ptrace` just ignore the `activate()` requirements (fortunately this PR catches that). However, I can't think of a better, but still self-explanatory solution.

test/initramfs/src/regression/process/arch_prctl/fsgsbase.c
```diff
@@ -0,0 +1,77 @@
...
+	asm volatile("wrgsbase %0" : : "r"(gs_base) : "memory");
+}
+
+FN_TEST(sync_gsbase_from_cpu)
+{
+	SKIP_TEST_IF(!cpu_has_fsgsbase());
+
+	long original_gs = arch_get_gs();
+	SKIP_TEST_IF(original_gs < 0);
+	SKIP_TEST_IF(arch_set_gs(original_gs) != 0);
```
lrh2000: What is the purpose of the two `SKIP_TEST_IF`s? I doubt they're useful...

Can't this test be made as simple as:
```rust
	SKIP_TEST_IF(!cpu_has_fsgsbase());

	uintptr_t syscall_gs = (uintptr_t)&gs_slots[0];
	TEST_SUCC(arch_set_gs(syscall_gs));
	usleep(100); // Trigger a syscall and context switch.
	TEST_RES(read_gsbase(), _ret == syscall_gs);
	TEST_RES(arch_get_gs(), _ret == syscall_gs);

	uintptr_t fsgsbase_gs = (uintptr_t)&gs_slots[1];
	write_gsbase(fsgsbase_gs);
	usleep(100); // Trigger a syscall and context switch.
	TEST_RES(read_gsbase(), _ret == fsgsbase_gs);
	TEST_RES(arch_get_gs(), _ret == fsgsbase_gs);
```

This is also much more symmetric than the current version.

(glibc won't use GS bases so we may not have to restore it as well? But correct me if I'm wrong)

kernel/src/process/execve.rs
```diff
@@ -283,6 +283,12 @@ fn set_cpu_context(
     *user_context.general_regs_mut() = GeneralRegs::default();
     // Clear the TLS pointer.
     user_context.set_tls_pointer(0);
+    user_context.activate_tls_pointer();
+    #[cfg(target_arch = "x86_64")]
```
lrh2000, line 290: I think your previous version has a comment here. Maybe the comment should be kept since `// Clear the TLS pointer.` isn't related to `set_gs_pointer(0)`.

kernel/src/syscall/arch_prctl.rs
```diff
@@ -41,8 +41,14 @@ fn do_arch_prctl(code: ArchPrctlCode, addr: u64, user_ctx: &mut UserContext) ->
             Ok(0)
         }
         ArchPrctlCode::ARCH_GET_FS => Ok(user_ctx.tls_pointer() as u64),
-        ArchPrctlCode::ARCH_GET_GS | ArchPrctlCode::ARCH_SET_GS => {
-            return_errno_with_message!(Errno::EINVAL, "GS cannot be accessed from the user space")
+        ArchPrctlCode::ARCH_SET_GS => {
```
lrh2000: I was thinking about whether we should sort the actions according to the numeric values:
```rust
    ARCH_SET_GS = 0x1001,
    ARCH_SET_FS = 0x1002,
    ARCH_GET_FS = 0x1003,
    ARCH_GET_GS = 0x1004,
```
So:
```
        ArchPrctlCode::ARCH_SET_GS => (),
        ArchPrctlCode::ARCH_SET_FS => (),
        ArchPrctlCode::ARCH_GET_FS => (),
        ArchPrctlCode::ARCH_GET_GS => (),
```

On the other hand, the Linux magic numbers are a bit unusual....

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -258,6 +262,26 @@ impl UserContext {
...
+    /// Gets the user-space GS base.
+    pub fn gs_pointer(&self) -> usize {
+        self.gsbase()
+    }
+
+    /// Activates the user-space GS base for the current task.
+    ///
+    /// On x86-64, the kernel runs with its own `gsbase` and returns to user mode via `swapgs`.
+    /// Therefore the user `gsbase` must be staged in `IA32_KERNEL_GS_BASE` while the kernel is
+    /// running.
```
lrh2000: Comments related to `IA32_KERNEL_GS_BASE` should be moved to the method body and converted to normal comments. They describe the implementation (i.e., `KernelGsBase::write`). Doc comments are written for callers instead.

lrh2000 (commented):
Thanks for revising the code!

ostd/src/arch/x86/cpu/context/mod.rs
tatetian: We should never do the expensive `sync_fsgsbase_pointers_from_cpu`, which involves a MSR write, on the performance-critical syscall path. 

> There are two types of register:
> 
> 1. Registers are saved and restored during syscall enter and exit.
> 2. Registers are saved and restored during context switches.

This is the right mental model to think about saving/storing CPU registers. For CPU states that are not touched by the kernel, such as FPU states, FS base, and GS base, the kernel should not save/restore them on syscall enter/exit. We should postpone the save and restore to context switching, at which point there is no way but to do the save and store.

With this observation, I suppose that we should:
1. Move fsbase and gsbase out of `GeneralRegs`, only keeping the registers that may be touched by the kernel in this struct. As fsbase and gsbase are moved out of `GeneralRegs`, they are not saved or restored on syscall or traps.
2. Add helper functions for reading and write the current values of userspace fsbase and gsbase. This way, we can access or manipulate these values on demand, even though they are not saved on syscall or traps.
3. Mimic what we do with FPU registers:
    a. A dedicated struct: Add a new `ThreadLocalStorageRegs` struct that consists of fsbase and gsbase. This struct represents architecture-specific, userspace thread storage pointer registers.
    b. Attaching it to per-thread objects: Add `ThreadLocalStorageRegs` as a field to both `PosixThread` (the initial value) and `ThreadLocal` (the up-to-date value).
    c. Do saving/restoring in `pre_schedule_handler` and `post_schedule_handler.

We do not implement the above strategy because the overhead of always saving/restoring FS base is minimal. But saving/restoring GS base is more expensive. So now is the time to adopt a more efficient strategy.

I believe that your proposal

```
pub fn activate_task_state() {}

pub fn sync_task_state() {}
```

can work, but I think treating FS/GS bases in a similar way as FPU states is more natural and consistent than inventing a new set of methods/abstractions for FS/GS bases.

ostd/src/arch/x86/task/switch.S
```diff
@@ -18,6 +18,11 @@ context_switch: # (nxt: *const TaskContext, cur: *mut TaskContext)
   mov [rsi + 48], r15
   rdfsbase r15
   mov [rsi + 64], r15 # 64 = offsetof(TaskContext, fsbase)
+  mov ecx, 0xc0000102 # IA32_KERNEL_GS_BASE
+  rdmsr
```
tatetian: The serializing `rwmsr` and `wrmsr` instructions are so expensive that even doing `swapgs` + `rdgsbase`/`wrgsbase` + `swapgs` would lead to a performance gain (estimated to be 3X-5X faster). Check out how [Linux does this](https://github.com/torvalds/linux/blob/8fde5d1d47f69db6082dfa34500c27f8485389a5/arch/x86/kernel/process_64.c#L195-L197).

ostd/src/arch/x86/cpu/context/mod.rs
lrh2000: > 1. Move fsbase and gsbase out of `GeneralRegs`

If we can do this, it's certainly a better solution. I have once thought this, but it looks to require more refactoring.

One important benefit of this is that the `activate()` and `sync()` calls are difficult to be missed. They are easy to miss if they are in `GeneralRegs`. If they're missed, the `fsbase`/`gsbase` fields in `GeneralRegs` are outdated or ineffective.

> ThreadLocalStorageRegs

This idea is good. Also note that some other architectures do save and restore TLS pointers on syscall entry and exit. Therefore, we cannot remove the `UserContext::set_tls_pointer` function for those architectures. Then we will end up with more `#[cfg(target_arch = "x86_64")]`. This is a (minor?) downside, though I'm not sure if there are ways to mitigate the disadvantages.

tatetian: > Also note that some other architectures do save and restore TLS pointers on syscall entry and exit. Therefore, we cannot remove the `UserContext::set_tls_pointer` function for those architectures.

This is certainly a valid concern. Or we can just save/restore gsbase (not fsbase) lazily on `pre_schedule_handler` and `post_schedule_handler`.

lrh2000: > save/restore gsbase (not fsbase)

I think it's even odder. We still need `activate_tls_pointer` and `sync_tls_pointer` if we keep `fsbase` in `UserContext`. However, if we have those two, I would prefer to just have `activate_task_state` and `sync_task_state`....

cqs21: I’m more inclined to have this PR focused on adding `activate_task_state()` and `sync_task_state()`, and leave those as follow-up refactoring:

> 1. Move `fsbase` and `gsbase` out of GeneralRegs
> 2. Add a new `ThreadLocalStorageRegs` struct

tatetian: > I’m more inclined to have this PR focused on adding `activate_task_state()` and `sync_task_state()`, and leave those as follow-up refactoring:

The thing is that I don't know who will do this follow-up and when. This optimization does not seem to be complex. If there is no technical blocker, then why not do it the right way in the first attempt?

The latest revision has avoided adding overhead to the hot syscall path. This is definitely a great improvement. But the added overhead is still suboptimal:

```diff
.global context_switch
.type context_switch, @function
context_switch: # (nxt: *const TaskContext, cur: *mut TaskContext)
   rdfsbase r15
   mov [rsi + 64], r15 # 64 = offsetof(TaskContext, fsbase)
+  swapgs
+  rdgsbase rax
+  swapgs
+  mov [rsi + 72], rax # 72 = offsetof(TaskContext, gsbase)
```

As far as OSTD is concerned, a task is not destined to touch FSBASE or GSBASE. So hardcoding saving/restoring FSBASE and GSBASE in OSTD's context switching logic is unreasonable.

From the perspective of the Asterinas kernel, only `PosixThread` might switche to the userspace, while `KernelThread` does not. So saving (or restoring) FSBASE/GSBASE when switching-out (or switching-in) a `KernelThread` is a complete waste of time.

If we maintain FSBASE and GSBASE in `ThreadLocal` and save/restore them in `pre_schedule_handler` and `post_schedule_handler`, then the above cost would be naturally eliminated. Context switching 

@cqs21 Could you try work out this optimized approach? You can keep the code of the current approach. If the optimized approach encounters unexpected or difficult obstacles, we can always change to the current approach.

tatetian:
By the way, the Linux Compatibility section (as well as the SCML file) should be updated.

ostd/src/arch/riscv/cpu/context.rs
```diff
@@ -28,6 +28,10 @@ pub struct UserContext {
     exception: Option<CpuException>,
 }
 
+/// Userspace thread-local storage registers.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub struct ThreadLocalStorageRegs;
```
tatetian: I don't think it is desirable to add an empty `ThreadLocalStorageRegs`. It is useless and confusing.

I suggest adding a new abstraction called `SuppCpuContext` and its friends to `thread_local.rs`. For most call sites, `SuppCpuContext` hides the technical detail whether a CPU arch uses `ThreadLocalStorageRegs` or not.

```rust
pub struct ThreadLocal {
    // fpu_context: RefCell<FpuContext>,
    // fpu_state: Cell<FpuState>,
    
    // Replace the old `fpu_context` and `fpu_state` fields with `supp_cpu_context`
    supp_cpu_context: SuppUserContext,
}

impl ThreadLocal {
    pub fn supp_user_context(&self) -> ThreadSuppUserContext<'a> {
        ThreadSuppUserContext(&self.supp_cpu_context)
    }
}

/// The current thread's supplementary userspace CPU context.
/// 
/// This struct provides access to the current thread's [`SuppUserContext`].
pub struct ThreadSuppUserContext<'a>(&'a SuppUserContext);

impl<'a> ThreadSuppUserContext<'a> {
    pub fn fpu(&self) -> ThreadFpu<'a> { ... }

    pub fn x86_tls_regs(&self) -> Threadx86TlsRegs<'a> { ... }
    
    pub fn before_schedule(&self) { ... }

    pub fn after_schedule(&self) { ... }
}

/// Supplementary userspace CPU context.
/// 
/// # `UserContext` vs `SuppUserContext`.
/// 
/// The entire userspace CPU state is split into two structs:
/// `UserContext` and `SuppUserContext`.
/// `UserContext` is a set of essential CPU registers,
/// including the general-purpose ones,
/// that are used by the kernel as well.
/// Thus, to avoid conflicts between the userspace and kernel space use,
/// `UserContext` has to be saved/restored _eagerly_ upon every exit/enter from/to the userspace.
/// For best performance,
/// the number of registers in `UserContext` is kept as small as possible.
/// 
/// In contrast,
/// `SuppUserContext` is a set of supplementary CPU registers,
/// such as FPU registers,
/// that are modifiable by the userpsace
/// but might or might not be used in the kernel space.
/// As such, 
/// the saving and restoring of `SuppUserContext` can be delayed to 
/// the point of context switching,
/// when saving and restoring are unavoidable.
/// 
/// In conclusion,
/// dividing userspace CPU states into `UserContext` and `SuppUserContext`
/// allows for better performance.
pub struct SuppUserContext {
    fpu_context: RefCell<FpuContext>,
    fpu_state: Cell<FpuState>,
    x86_tls_regs: X86TlsRegs,
}

impl SuppUserContext {
    // Add methods...
}

mod x86_tls_regs {
    pub struct Threadx86TlsRegs<'a>(&'a x86TlsRegs);

    /// Thread Local Storage (TLS) Registers on x86-64.
    /// 
    /// Unlike most CPU architectures,
    /// x86 has two registers that point to the base addresses of TLS storage:
    /// `FSBASE` and `GSBASE`.
    /// Their uses in the user and kernel spaces are summarized in the table below:
    /// 
    /// |              | FSBASE                    | GSBASE                   |
    /// | ------------ | ------------------------- | ------------------------- |
    /// | Userspace    | Thread-local data pointer | (Optional) App-controlled |
    /// | Kernel space | None                      | CPU-local data pointer    |
    /// 
    pub struct x86TlsRegs {
        fsbase: usize,
        gsbase: usize,
    }
 
    impl x86TlsRegs {
        pub fn save(&mut self) { ... }
        pub fn save_fs_base(&mut self) { ... }
        pub fn save_gs_base(&mut self) { ... }
        

        pub fn load(&mut self) { ... }
        pub fn load_fs_base(&mut self) { ... }
        pub fn load_gs_base(&mut self) { ... }
    }
}
```

With this new abstraction that combines FPU and x86 TLS registers, the call sites can be simplified. For example, the scheduler hook is simplified from

```rust
fn pre_schedule_handler() {
    // ...

    thread_local
        .thread_local_storage_regs()
        .borrow_mut()
        .sync_task_state();
    thread_local.fpu().before_schedule();
}
```

to

```rust
fn pre_schedule_handler() {
    // ...

    thread_local.supp_user_context().before_schedule();
}
```

As another example, the `exceve.rs` file can be simiplified from

```rust
fn set_cpu_context(
    thread_local: &ThreadLocal,
    user_context: &mut UserContext,
    elf_load_info: &ElfLoadInfo,
) {
    // Reset FPU context.
    thread_local.fpu().set_context(FpuContext::new());
    // ..
     {
         let mut thread_local_storage_regs = thread_local.thread_local_storage_regs().borrow_mut();
         thread_local_storage_regs.set_fsbase(0);
         // Clear the GS base pointer.
         thread_local_storage_regs.set_gsbase(0);
         thread_local_storage_regs.activate_task_state();
     }
     // ..
}
```

to

```rust
fn set_cpu_context(
    thread_local: &ThreadLocal,
    user_context: &mut UserContext,
    elf_load_info: &ElfLoadInfo,
) {
    let supp_user_ctx = thread_local.supp_user_context()
    supp_user_ctx.set(SuppCpuContext::new());
    supp_user_ctx.activate();
    // ..
}
```

The `posix_thread/build.rs` can be simplied from

```rust
/// The builder to build a POSIX thread
pub struct PosixThreadBuilder {
    // ...
    fpu_context: FpuContext,
    thread_local_storage_regs: ThreadLocalStorageRegs,
    // ...
}
```

to

```rust
/// The builder to build a POSIX thread
pub struct PosixThreadBuilder {
    // ...
    supp_user_context: SuppUserContext,
    // ...
}
```

The code that touches FPU or FSBASE/GSBASE should be adapted to get `ThreadFpu` and `ThreadX86TlsRegs` out of `ThreadSuppUserContext` to manipulate the two sets of registers, respectively.

kernel/src/thread/mod.rs
```diff
@@ -33,6 +33,10 @@ fn pre_schedule_handler() {
         return;
     };
 
+    thread_local
+        .thread_local_storage_regs()
```
tatetian: Simplify this name to `tls_regs` and `ThreadLocalStorageRegs` to `TlsRegs`.

kernel/src/thread/task.rs
tatetian: Do we really need to activate/deactivate FPU registers upon every enter to and exit from the userspace? We do not do so for FSBASE/GSBASE, so I believe the same optimization can be done for FPU registers.

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -67,8 +71,108 @@ pub struct GeneralRegs {
...
+        self.fsbase = fsbase;
+    }
+
+    /// Gets the userspace FS base.
+    pub fn fsbase(&self) -> usize {
+        self.fsbase
+    }
+
+    /// Synchronizes the saved userspace FS base from the current task.
+    pub fn sync_fsbase(&mut self) {
```
tatetian: For virtually the same action, `FpuContext` uses the verb "save". And for the "activate" action, `FpuContext` uses the verb "load". If `ThreadLocalStorageRegs` and `FpuContext` are always bundled (as I suggested above), then the terms adopted by them should be unified.

tatetian (commented):
Thanks for revising the PR for improved performance.

tatetian:
The second an fourth commits look to be intermediate ones that amend the first one. I think they should be squashed into one.

book/src/ostd/soundness/safe-user-kernel-interactions.md
```diff
@@ -8,7 +8,10 @@ and accessing user memory ([`VmReader<Fallible>`](https://asterinas.github.io/ap
 ## UserMode and UserContext
 
 `UserContext` stores the user-space register state
-(general-purpose registers, RFLAGS, FS/GS base, and exception info).
+(general-purpose registers, RFLAGS, and exception info).
+On x86, the userspace FS/GS base is stored separately in `X86TlsRegs`
+as part of the supplementary user context (`SuppUserContext`),
+which also includes the FPU state.
```
tatetian: Remove the above three lines. The target of soundness analysis is OSTD, yet `SuppUserContext` are part of the kernel. Mentioning `SuppUserContext` won't contribute to the soundness analysis.

book/src/ostd/soundness/safe-user-kernel-interactions.md
```diff
@@ -47,10 +50,12 @@ are enforced every time before entering user space:
    the interrupt handler would see the user GS base instead of the kernel GS base,
    corrupting CPU-local storage access.
 
-3. **FS base**:
-   The user's `fsbase` is stored in `UserContext` and restored via `wrfsbase`.
-   The kernel uses GS (not FS) for CPU-local storage,
-   so a user-controlled FS base cannot affect kernel state.
+3. **FS/GS base**:
+   On x86, the user's `fsbase` and `gsbase` are stored in `X86TlsRegs`,
+   which is part of `SuppUserContext` (the supplementary userspace CPU context).
```
tatetian: I think the third point "FS/GS base" can be removed given that `UserContext` does not contain FS and GS bases any more.

ostd/src/arch/x86/cpu/context/mod.rs
```diff
@@ -67,8 +71,100 @@ pub struct GeneralRegs {
...
+    }
+
+    /// Saves the GS base from the CPU into this struct.
+    pub fn save_gs_base(&mut self) {
+        // On x86-64, the kernel runs with its own `gsbase` and returns to user mode via `swapgs`.
+        // Therefore the user `gsbase` is staged in `IA32_KERNEL_GS_BASE` while the kernel is
+        // running, so it must be read back from there.
+        // Do not use `swapgs` + `rdgsbase` + `swapgs` here: this method can be called with
+        // interrupts enabled, and an interrupt between the `swapgs` instructions would run with
+        // the user GS base instead of the kernel GS base.
```
tatetian: So why not first disable the local IRQs then do `swapgs` + `rdgsbase` + `swapgs`? Disabling local IRQs first does not change the equation of  `swapgs` + `rdgsbase` + `swapgs` vs `rdmsr`.

kernel/src/process/posix_thread/thread_local.rs
```diff
@@ -199,19 +199,88 @@ impl ThreadLocal {
...
+/// The current thread's supplementary userspace CPU context.
 ///
-/// - `Activated`: The FPU context is currently loaded onto the CPU and it must be loaded
-///   while the associated task is running. If preemption occurs in between, the context switch
-///   must load FPU context again.
-/// - `Loaded`: The FPU context is currently loaded onto the CPU. It may or may not still
-///   be loaded in CPU after a context switch.
-/// - `Unloaded`: The FPU context is not currently loaded onto the CPU.
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-enum FpuState {
```
lrh2000, line 211: Do we intend to remove the FPU state machine so that the FPU always needs to be saved and restored during a context switch, even though this may not be necessary for some context switches when handling a system call?

lrh2000: Oh, is this related to https://github.com/asterinas/asterinas/pull/3286#discussion_r3338356838?
> Do we really need to activate/deactivate FPU registers upon every enter to and exit from the userspace? We do not do so for FSBASE/GSBASE, so I believe the same optimization can be done for FPU registers.

But removing the state machine is not an optimization?

test/initramfs/src/regression/process/run_test.sh
```diff
@@ -15,6 +15,10 @@ set -e
 
 ./cpu_affinity/cpu_affinity
 
+if [ "$(uname -m)" = "x86_64" ]; then
+    ./arch_prctl/fsgsbase
```
lrh2000: I think `arch_prctl` < `cpu_affinity`. This should be moved to the beginning of the file?

kernel/src/process/clone.rs
```diff
@@ -452,6 +458,10 @@ fn clone_child_task(
         .user_ns(child_user_ns)
         .ns_proxy(child_ns_proxy)
         .default_timer_slack_ns(default_timer_slack_ns);
+        #[cfg(target_arch = "x86_64")]
+        {
+            thread_builder = thread_builder.tls_regs(child_tls_regs);
+        }
```
lrh2000, line 464: Is the bracket really necessary?

kernel/src/process/clone.rs
```diff
@@ -580,6 +594,10 @@ fn clone_child_process(
             .ns_proxy(child_ns_proxy)
             .default_timer_slack_ns(default_timer_slack_ns)
         };
+        #[cfg(target_arch = "x86_64")]
+        {
+            child_thread_builder = child_thread_builder.tls_regs(child_tls_regs);
+        }
```
lrh2000, line 602: Same here.

kernel/src/process/clone.rs
cqs21: Without the bracket, there is an error:
```
error[E0658]: attributes on expressions are experimental
```
But it can be fixed by adding `#![feature(stmt_expr_attributes)]` to the crate attributes.

lrh2000: Thanks for the explanation! Ignore my comments about this...

kernel/src/process/posix_thread/thread_local.rs
cqs21: > Do we intend to remove the FPU state machine so that the FPU always needs to be saved and restored during a context switch.

This commit removed the FPU state machine in order to unify FPU and FS/GS base handling under `before_schedule/after_schedule`.

I agree that the `FpuState` tracking is a worthwhile optimization — it marks whether the FPU context is live on the CPU and defers the actual `xrstor` until we truly need to return to userspace. This matters especially for I/O-heavy workloads where a blocking syscall may trigger multiple context switches, each of which would otherwise do a redundant FPU save/load cycle.

For FS/GS base registers, the switching cost is relatively small — `rdfsbase/wrfsbase` for FS base, and `rdmsr/wrmsr` for GS base. In an earlier iteration of this PR, I did add a flag to defer the FS/GS base load until the actual return to userspace, but decided it wasn't worth the added complexity given the much lower cost compared to `xsave/xrstor`.

So I'll add the `FpuState` machine back and keep "activate/deactivate FPU registers upon every enter to and exit from the userspace".

tatetian: No worry. `CpuSync<R: UserReg>` comes to the rescue.

# New abstraction: `CpuSync<R: UserReg>`

Introducing a new abstraction called `CpuSync<R: UserReg>`, which decouples the state machine from `FpuState` and `FpuContext` to any user-mode, supplementary CPU registers such as FPU, FSBASE, and GSBASE. Using this new abstraction, we can minimize the expensive save/restore traffic between in-memory Rust objects and the CPU's hardware registers without harming correctness or readability.

```rust
/// A smart cache of one user-mode CPU register
/// to minimize the expensive save/restore traffic 
/// between memory and the CPU's hardware register.
///
/// # How it works
///
/// ## Canonical values
///
/// To minimize the expensive saves/restores from/to the CPU hardware registers,
/// `CpuSync<R>` tracks *where* the latest **canonical value** of 
/// a register of type `R` currently lives 
/// and using that to elide redundant work.
/// At every moment in time, the canonical value lives 
/// in one or both of two places,
/// tracked by [`CanonicalValueLocation`]:
///
/// | Location   | Memory canonical | CPU canonical |
/// | ---------- | :--------------: | :-----------: |
/// | `InMemory` | ✓                | ✗             |
/// | `OnCpu`    | ✗                | ✓             |
/// | `Both`     | ✓                | ✓             |
///
/// ## Invariant
///
/// `CpuSync<R>` preserves the invariant that
/// **the recorded location reflects where the canonical value actually resides.**
/// Four methods cooperate to maintain it:
///
/// - The [`before_schedule`] event hook method runs
///   just before swithcing out the current task.
/// - The [`before_user_exec`] event hook method runs
///   just before entering the user space.
/// - The [`get`] accessor method returns the canonical value.
/// - The [`set`] accessor method writes a new canonical value into memory.
///
/// The two event hooks are deliberately placed at *the latest moments*
/// where save and restore are still correct.
/// Delaying as far as possible allows `CpuSync`
/// to do as many optimizations as possible
///
/// ## Optimizations
/// 
/// Thanks to the invariant,
/// `CpuSync<R>` takes the optimization opportunities below:
///
/// - At [`before_schedule`]: 
///   an expensive `<R as UserReg>::save_from_cpu` is needed
///   *only if* the canonical value is `OnCpu`;
///   if it's `InMemory` or `Both`, then the save is skipped safely.
///   Either way the location transitions to `InMemory`
///   because the imminent context switch clobbers the CPU.
/// - At [`before_user_exec`]:
///   an expensive `<R as UserReg>::restore_to_cpu` is needed
///   *only if* the value is `InMemory`;
///   if it's `OnCpu` (successive loop iterations) or
///   `Both` (a prior `get` followed by no schedule),
///   then the load is skipped safely.
/// - At [`get`]:
///   an expensive `<R as UserReg>::save_from_cpu` is needed
///   *only if* the value is `OnCpu`;
///   afterwards the location transitions to `Both`,
///   so a subsequent `before_schedule` can skip the save.
/// - At [`set`]:
///   nothing touches the CPU; the value is written to memory and
///   the location transitions to `InMemory`, so the new value reaches the CPU
///   lazily at the next `before_user_exec` (batching multiple sets into one
///   load, and performing no CPU write at all if the task exits first).
///
/// # Concurrency
///
/// `CpuSync<R>` is intended to live inside a per-thread `ThreadLocal` 
/// and uses `Cell`/`RefCell` for interior mutability accordingly.
/// It is **not** `Sync` and must not be accessed across threads.
///
/// All methods assume kernel preemption cannot occur
/// for the duration of the call (the existing `ThreadLocal` discipline).
///
/// [`before_schedule`]:  Self::before_schedule
/// [`before_user_exec`]: Self::before_user_exec
/// [`get`]:              Self::get
/// [`set`]:              Self::set
pub struct CpuSync<R> {
    location: Cell<CanonicalValueLocation>,
    reg: RefCell<R>,
}

impl<R: UserReg> CpuSync<R> {
    /// Constructs a `CpuSync` holding `reg`,
    /// with the canonical value initially `InMemory`.
    /// 
    /// The CPU register is presumed to 
    /// hold some other task's bytes (or nobody's)
    /// until the first [`Self::before_user_exec`] loads `reg` onto it.
    pub const fn new(reg: R) -> Self {
        Self {
            location: Cell::new(CanonicalValueLocation::InMemory),
            reg: RefCell::new(reg),
        }
    }

    /// Gets a copy of the current canonical value.
    ///
    /// This method guarantees that 
    /// the returned value reflects the very latest state.
    pub fn get(&self) -> R
    where
        R: Clone,
    {
        if self.location.get() == CanonicalValueLocation::OnCpu {
            self.reg.borrow_mut().save_from_cpu(); // sync CPU → memory
            self.location.set(CanonicalValueLocation::Both);
        }
        self.reg.borrow().clone()
    }

    /// Sets the canonical value.
    ///
    /// The new value reaches the CPU on the next [`before_user_exec`].
    /// The `set` method itself only writes the in-memory copy and marks `InMemory`.
    /// This avoids a CPU write when:
    ///
    /// 1. `set` is followed by more `set`s before user-entry —
    ///    they collapse into a single load.
    /// 2. The task never re-enters user mode
    ///    after this `set` (e.g. exits in the kernel) — no CPU write happens at all.
    /// 3. A context switch intervenes — the new value rides into the CPU on
    ///    the eventual `before_user_exec`, with `before_schedule` a no-op
    ///    along the way (since the location is already `InMemory`).
    ///
    /// [`before_user_exec`]: Self::before_user_exec
    pub fn set(&self, new_reg: R) {
        *self.reg.borrow_mut() = new_reg;
        if self.location.get() != CanonicalValueLocation::InMemory {
            self.location.set(CanonicalValueLocation::InMemory);
        }
    }

    /// Returns where the canonical value currently lives.
    ///
    /// Only useful for diagnostics.
    pub fn location(&self) -> CanonicalValueLocation {
        self.location.get()
    }

    // ---- Lifecycle callbacks ----

    /// Scheduler hook: this task is about to be switched off the CPU.
    ///
    /// This hook method ensures that
    /// the canonical value of the register `R` is saved into `self`
    /// and the canonical value location is marked `InMemory`. 
    pub fn before_schedule(&self, guard: &DisabledLocalIrqGuard) {
        let loc = self.location.get();
        if loc == CanonicalValueLocation::OnCpu {
            self.reg.borrow_mut().save_from_cpu_with_irq_disabled(guard);
        }
        if loc != CanonicalValueLocation::InMemory {
            self.location.set(CanonicalValueLocation::InMemory);
        }
    }

    /// User-task-loop hook: this task is about to enter user mode.
    ///
    /// Must be called immediately before `UserMode::execute()`,
    /// before any user-visible CPU instruction runs.
    ///
    /// This hook method ensures that 
    /// the canonical value of the register `R` is loaded to the CPU hardware
    /// and the canonical value location is marked `OnCpu`. 
    pub fn before_user_exec(&self) {
        if self.location.get() == CanonicalValueLocation::InMemory {
            self.reg.borrow().restore_to_cpu();
        }
        self.location.set(CanonicalValueLocation::OnCpu);
    }
}

/// A user-mode CPU register
/// whose in-memory value may be moved to and from the CPU's hardware register.
///
/// Implementors typically wrap one or a small bundle of registers/MSRs
/// (e.g. the user FS base, the user GS base, or the FPU XSAVE area) 
/// into a struct that implements this trait.
pub trait UserReg {
    /// Saves the current value from the CPU register(s) into `self`.
    ///
    /// May be invoked in any IRQ state.
    fn save_from_cpu(&mut self);

    /// Saves the current value from the CPU register(s) into `self`,
    /// with the caller's guarantee that local IRQs are disabled.
    ///
    /// Implementors may override to take a cheaper path 
    /// that is only safe with IRQs disabled
    /// The default falls back to [`Self::save_from_cpu`].
    fn save_from_cpu_with_irq_disabled(&mut self, _guard: &DisabledLocalIrqGuard) {
        self.save_from_cpu();
    }

    /// Writes `self`'s value into the CPU register(s).
    ///
    /// May be invoked in any IRQ state.
    fn restore_to_cpu(&self);

    /// Writes `self`'s value into the CPU register(s), 
    /// with the caller's guarantee that local IRQs are disabled.
    ///
    /// Symmetric to [`Self::save_from_cpu_with_irq_disabled`].
    fn restore_to_cpu_with_irq_disabled(&self, _guard: &DisabledLocalIrqGuard) {
        self.restore_to_cpu()
    }
}

/// Where the latest canonical value of a [`UserReg`] currently lives.
pub enum CanonicalValueLocation {
    /// Only the in-memory copy holds the canonical value.
    InMemory,
    /// Only the CPU register holds the canonical value.
    OnCpu,
    /// Both the in-memory copy and the CPU register hold the canonical value.
    Both,
}
```

**I carefully reviewed the above design of `CpuSync`, but not the rest of the related changes, as shown below. Take them with a grain of salt.**

```rust
// In kernel/src/process/posix_thread/thread_local.rs.
pub struct SuppUserContext {
    fpu: CpuSync<FpuContext>,
    fs_base: CpuSync<FsBase>,
    gs_base: CpuSync<GsBase>,
}

impl SuppUserContext {
    pub(crate) fn new() -> Self {
        Self {
            fpu: CpuSync::new(FpuContext::new()),
            fs_base: CpuSync::new(FsBase::default()),
            gs_base: CpuSync::new(GsBase::default()),
        }
    }

    pub fn fpu(&self)     -> &CpuSync<FpuContext> { &self.fpu }
    pub fn fs_base(&self) -> &CpuSync<FsBase>     { &self.fs_base }
    pub fn gs_base(&self) -> &CpuSync<GsBase>     { &self.gs_base }

    pub fn before_schedule(&self, guard: &DisabledLocalIrqGuard) {
        self.fpu.before_schedule(guard);
        {
            self.fs_base.before_schedule(guard);
            self.gs_base.before_schedule(guard);
        }
    }

    pub fn before_user_exec(&self) {
        self.fpu.before_user_exec();
        {
            self.fs_base.before_user_exec();
            self.gs_base.before_user_exec();
        }
    }
}
```

```rust
// In ostd/src/arch/x86/cpu/context/mod.rs, replacing X86TlsRegs.
#[derive(Clone, Copy, Debug, Default)]
pub struct FsBase(usize);
impl FsBase {
    pub fn new(addr: usize) -> Self { Self(addr) }
    pub fn addr(&self) -> usize { self.0 }
}
impl SuppUserReg for FsBase {
    fn save_from_cpu(&mut self) {
        // SAFETY: `CR4.FSGSBASE` is enabled by `enable_cpu_features`; reading the user FS base
        // does not affect kernel state.
        self.0 = unsafe { rdfsbase() as usize };
    }
    fn restore_to_cpu(&self) {
        // SAFETY: setting fsbase to any value cannot affect kernel state — the kernel does not
        // use FS.
        unsafe { wrfsbase(self.0 as u64) }
    }
}

#[derive(Clone, Copy, Debug, Default)]
pub struct GsBase(usize);
impl GsBase {
    pub fn new(addr: usize) -> Self { Self(addr) }
    pub fn addr(&self) -> usize { self.0 }
}
impl SuppUserReg for GsBase {
    /// Precondition: called from kernel mode after a user→kernel `swapgs`, so the user gsbase
    /// is staged in `IA32_KERNEL_GS_BASE`.
    fn save_from_cpu(&mut self) {
        self.0 = KernelGsBase::read().as_u64() as usize;
    }
    fn restore_to_cpu(&self) {
        KernelGsBase::write(VirtAddr::new(self.0 as u64));
    }
}

impl SuppUserReg for FpuContext {
    fn save_from_cpu(&mut self) { /* existing FpuContext::save body */ }
    fn restore_to_cpu(&self)    { /* existing FpuContext::load body, taking &self — see note below */ }
}
```

`FpuContext::load` currently takes `&mut self` but doesn't actually mutate (XRSTOR doesn't need a mutable area pointer — `as_bytes()`/`as_ptr()` suffice). Switch its receiver to `&self` so it fits the `SuppUserReg::restore_to_cpu(&self)` signature; this matches the earlier comment about `after_schedule` only borrowing `RefCell<FpuContext>`.

```rust
// In ostd/src/task/mod.rs — widen the pre-schedule handler's signature so the
// scheduler can pass the IRQ guard down into the SuppUserContext path.
pub fn inject_pre_schedule_handler(handler: fn(&DisabledLocalIrqGuard));

// In ostd/src/task/processor.rs::before_switching_to — pass the guard:
if let Some(handler) = PRE_SCHEDULE_HANDLER.get() {
    handler(irq_guard);
}

// In kernel/src/thread/mod.rs::pre_schedule_handler — accept and forward the guard:
fn pre_schedule_handler(guard: &DisabledLocalIrqGuard) {
    let Some(task) = Task::current() else { return };
    let Some(thread_local) = task.as_thread_local() else { return };
    thread_local.supp_user_context().before_schedule(guard);
}

// In kernel/src/thread/mod.rs::post_schedule_handler — drop the after_schedule
// call entirely (the new CpuSync has no after-schedule method). VMar activation
// and the counter stay.

// In kernel/src/thread/task.rs — new pre-call in the user-task loop:
while !current_thread.is_exited() {
    ctx.thread_local.supp_user_context().before_user_exec();
    let return_reason = user_mode.execute(has_kernel_event_fn);
    // No after_user_exec needed: the CPU keeps the (possibly user-modified)
    // bytes, marked OnCpu — the next iteration's before_user_exec short-circuits,
    // or the next before_schedule picks them up.
    // … existing handling unchanged …
}
```

tatetian (approved):

PR #3287, opened 2026-05-27
Fix linux-legacy32 boot with stripped payload

li041:
## Summary

  The linux-legacy32 boot path embeds the kernel ELF into the Linux bzImage
  setup payload. On this branch, the unstripped kernel ELF can become very large
  because it contains debug and symbol sections such as `.debug_info`,
  `.debug_str`, and `.strtab`.

  These sections are not needed for booting, but they still increase the payload
  size. For legacy32 boot, the oversized setup payload can overlap with the
  kernel load range and make the setup code fail before entering Asterinas.

  This PR enables `--strip-elf` for `BOOT_PROTOCOL=linux-legacy32` and derives a
  separate stripped `boot_elf` from the original kernel ELF. The stripped
  `boot_elf` is used only as the Linux bzImage payload for boot loading.

  The original unstripped kernel ELF is still kept in the bundle. It preserves
  debug information for post-run panic analysis and symbolization.

## Test
`make run_kernel AUTO_TEST=boot SMP=4 BOOT_PROTOCOL=linux-legacy32`

### DIFF
--- a/osdk/src/commands/build/mod.rs
+++ b/osdk/src/commands/build/mod.rs
@@ -11,7 +11,7 @@ use std::{
     time::SystemTime,
 };
 
-use bin::{make_elf_for_qemu, make_install_bzimage};
+use bin::{make_elf_for_qemu, make_install_bzimage, make_stripped_boot_elf};
 
 use super::util::{COMMON_CARGO_ARGS, DEFAULT_TARGET_RELPATH, cargo, profile_name_adapter};
 use crate::{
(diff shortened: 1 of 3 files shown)


osdk/src/commands/build/mod.rs
```diff
@@ -200,6 +206,45 @@ pub fn do_cached_build(
...
+    strip_elf: bool,
+) -> AsterBin {
+    if !strip_elf || aster_elf.stripped() {
+        return aster_elf.clone();
+    }
+
+    make_stripped_boot_elf(output_dir, aster_elf)
+}
+
+fn make_stripped_boot_elf(output_dir: impl AsRef<Path>, aster_elf: &AsterBin) -> AsterBin {
```
junyang-zh: https://github.com/junyang-zh/asterinas/blob/8decda9c9a3bc2b38debc722593dcd5b807579b3/osdk/src/commands/build/bin.rs#L95-L120

`--strip-elf` was only used by QEMU direct booting. Instead of adding stripping for legacy boot, maybe we should make all boot methods respect this argument. A right fix would be to unify this code into a single function and call it from a boot method-agnostic point.

junyang-zh (commented):
I think it's indeed a bug to fix.

osdk/src/commands/build/mod.rs
li041: Thanks. I now put `make_stripped_boot_elf` at the entrance of `do_cached_build`, and it use stripped elf to boot for all boot methods and origin elf  to privide info during panic tracing seperately.

tatetian (approved):

PR #2265, opened 2025-07-15
Support syslog(2): add syscall(2), logger integration and tests

TankTechnology:
- Add `sys_syslog` module with circular 64 KB kernel log buffer and `dmesg_restrict`.
- Extend `AsterLogger` to push logs into the buffer and honor console log levels.
- Register callback in syscall init, dispatch on x86/RISCV (`SYS_SYSLOG`).

PR #3288, opened 2026-05-27
Support `syslog` syscall

cchanging:
This PR is a further improvement on PR #2265. Thanks the contribution of @TankTechnology !

### DIFF
--- a/kernel/comps/logger/src/klog.rs
+++ b/kernel/comps/logger/src/klog.rs
@@ -1,0 +1,395 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Provides the kernel log buffer and Linux-compatible log controls.
+//!
+//! This module stores formatted kernel log records in a fixed-size ring buffer
+//! and exposes methods used by Linux `syslog` and `dmesg` operations.
+
+use core::{
+    fmt::Write,
+    num::Wrapping,
+    sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering},
+    time::Duration,
+};
+
+use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
+use int_to_c_enum::TryFromInt;
+use ostd::{
+    log::{Level, Record},
+    mm::{Fallible, FallibleVmWrite, VmReader, VmWriter},
+    sync::{LocalIrqDisabled, SpinLock, WaitQueue},
+};
+use ring_buffer::RingBuffer;
+use spin::Once;
+
+/// Size of the kernel log ring buffer (64 KB).
+///
+/// This matches the default `CONFIG_LOG_BUF_SHIFT=16` in Linux.
+/// The buffer stores formatted log messages including timestamps.
+const LOG_BUFFER_CAPACITY: usize = 64 * 1024;
+
+/// Maximum size of a single formatted log record.
+///
+/// Linux bounds formatted printk records with `PRINTKRB_RECORD_MAX`, currently
+/// 1024 bytes. This limit applies before the record is appended to the ring
+/// buffer, so an oversized record cannot evict the whole log buffer by itself.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v6.13/source/kernel/printk/internal.h#L52>
+const MAX_RECORD_SIZE: usize = 1024;
+
+/// Minimum Linux console log level accepted by syslog console controls.
+///
+/// Linux clamps `SYSLOG_ACTION_CONSOLE_LEVEL` to `minimum_console_loglevel`
+/// and uses the same level for `SYSLOG_ACTION_CONSOLE_OFF`.
+pub const MINIMUM_CONSOLE_LOG_LEVEL: LinuxConsoleLogLevel = LinuxConsoleLogLevel::Emerg;
+
+/// Linux console log level.
+///
+/// Controls which log messages are broadcast to the console. Only messages
+/// with a priority level less than the console log level will be displayed.
+/// For example, setting the console level to `Error` (4) will display messages
+/// of levels `Emerg`, `Alert`, `Crit`, and `Error`.
+///
+/// These values correspond to Linux's console log level as documented in
+/// `man 2 syslog` and used by the `SYSLOG_ACTION_CONSOLE_LEVEL` action.
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, TryFromInt)]
+pub enum LinuxConsoleLogLevel {
+    /// Allow messages with priority `KERN_EMERG` (0) or higher.
+    Emerg = 1,
+    /// Allow messages with priority `KERN_ALERT` (1) or higher.
+    Alert = 2,
+    /// Allow messages with priority `KERN_CRIT` (2) or higher.
+    Crit = 3,
+    /// Allow messages with priority `KERN_ERR` (3) or higher.
+    Err = 4,
+    /// Allow messages with priority `KERN_WARNING` (4) or higher.
+    Warn = 5,
+    /// Allow messages with priority `KERN_NOTICE` (5) or higher.
+    Notice = 6,
+    /// Allow messages with priority `KERN_INFO` (6) or higher.
+    Info = 7,
+    /// Allow messages with priority `KERN_DEBUG` (7) or higher (all messages).
+    Debug = 8,
+}
+
+impl LinuxConsoleLogLevel {
+    /// Checks if a log level should be printed at this console log level.
+    fn should_print(self, level: Level) -> bool {
+        (level as u8) < (self as u8)
+    }
+}
+
+impl From<LinuxConsoleLogLevel> for u8 {
+    fn from(level: LinuxConsoleLogLevel) -> Self {
+        level as u8
+    }
+}
+
+define_atomic_version_of_integer_like_type!(LinuxConsoleLogLevel, try_from = true, {
+    /// An atomic version of `LinuxConsoleLogLevel`.
+    #[derive(Debug)]
+    struct AtomicLinuxConsoleLogLevel(AtomicU8);
+});
+
+static KLOG: Once<KernelLog> = Once::new();
+
+/// Returns the global kernel log.
+pub fn klog() -> &'static KernelLog {
+    KLOG.call_once(KernelLog::new)
+}
+
+/// Stores kernel log records and Linux-compatible log controls.
+///
+/// `KernelLog` owns the fixed-size ring buffer used by `dmesg`/`syslog`,
+/// console log-level state, and the `dmesg_restrict` access-control flag.
+/// Callers should obtain the global instance with [`klog`].
+pub struct KernelLog {
+    buffer: SpinLock<RingBuffer<u8>, LocalIrqDisabled>,
+    dmesg_restrict: AtomicBool,
+    /// First byte visible to non-destructive latest-pick operations.
+    pick_head: AtomicUsize,
+    /// Current console log level.
+    console_level: AtomicLinuxConsoleLogLevel,
+    /// Saved console log level for restore after disable/enable.
+    saved_console_level: SpinLock<Option<LinuxConsoleLogLevel>, LocalIrqDisabled>,
+    waitq: WaitQueue,
+}
+
+impl KernelLog {
+    fn new() -> Self {
+        Self {
+            buffer: SpinLock::new(RingBuffer::new(LOG_BUFFER_CAPACITY)),
+            dmesg_restrict: AtomicBool::new(false),
+            pick_head: AtomicUsize::new(0),
+            // TODO: Make the initial console level be configurable.
+            console_level: AtomicLinuxConsoleLogLevel::new(LinuxConsoleLogLevel::Err),
+            saved_console_level: SpinLock::new(None),
+            waitq: WaitQueue::new(),
+        }
+    }
+
+    /// Appends a log record to the kernel log buffer.
+    ///
+    /// The record is formatted with a timestamp prefix before being stored.
+    /// This method does not allocate memory and can be called from interrupt
+    /// context.
+    pub fn append(&self, record: &Record, timestamp: &Duration) {
+        let bytes_written = {
+            let mut buffer = self.buffer.lock();
+            let mut formatter = buffer.formatter().limit(MAX_RECORD_SIZE);
+            let _ = writeln!(
+                formatter,
+                "[{:>6}.{:03}] {:<6}: {}{}",
+                timestamp.as_secs(),
+                timestamp.subsec_millis(),
+                record.level(),
+                record.prefix(),
+                record.args()
+            );
+            formatter.bytes_written()
+        };
+
+        if bytes_written > 0 {
+            self.waitq.wake_all();
+        }
+    }
+
+    /// Reads log messages from the kernel log buffer destructively.
+    ///
+    /// Returns the number of bytes written into `writer`.
+    pub fn read(&self, writer: &mut VmWriter<'_, Fallible>) -> ostd::Result<usize> {
+        let mut total_written = 0;
+        let mut temp_buffer = [0u8; MAX_RECORD_SIZE];
+
+        while writer.has_avail() {
+            let bytes_read = {
+                let mut buffer = self.buffer.lock();
+                let read_len = buffer.len().min(temp_buffer.len()).min(writer.avail());
+                if read_len == 0 {
+                    return Ok(total_written);
+                }
+
+                let read_start = buffer.head();
+                let read_end = read_start + Wrapping(read_len);
+                let mut temp_writer = VmWriter::from(&mut temp_buffer[..read_len]).to_fallible();
+                // The writer points to a valid kernel stack buffer, so the
+                // fallible copy cannot fault.
+                let written = buffer
+                    .pick_range(read_start..read_end, &mut temp_writer)
+                    .unwrap();
+                buffer.commit_read(written);
+                written
+            };
+
+            let mut temp_reader = VmReader::from(&temp_buffer[..bytes_read]);
+            let bytes_written = match writer.write_fallible(&mut temp_reader) {
+                Ok(written) => written,
+                Err((err, 0)) if total_written == 0 => return Err(err),
+                Err((_, written)) => written,
+            };
+            total_written += bytes_written;
+
+            if bytes_written < bytes_read {
+                return Ok(total_written);
+            }
+        }
+
+        Ok(total_written)
+    }
+
+    /// Picks the last log messages from the kernel log buffer.
+    ///
+    /// Writes the last available log bytes that fit in `writer`, ending at the
+    /// current buffer tail. The data remains in the buffer after picking.
+    pub fn pick_latest(&self, writer: &mut VmWriter<'_, Fallible>) -> ostd::Result<usize> {
+        self.pick_latest_impl(writer, false)
+    }
+
+    /// Picks the last log messages and clears the visible range.
+    ///
+    /// Writes the last visible log bytes that fit in `writer`, ending at the
+    /// current buffer tail. The pick head is advanced under the same buffer
+    /// lock used for picking, so messages appended after the pick are not
+    /// accidentally hidden.
+    pub fn pick_latest_and_clear(
+        &self,
+        writer: &mut VmWriter<'_, Fallible>,
+    ) -> ostd::Result<usize> {
+        self.pick_latest_impl(writer, true)
+    }
+
+    fn pick_latest_impl(
+        &self,
+        writer: &mut VmWriter<'_, Fallible>,
+        clear_after_pick: bool,
+    ) -> ostd::Result<usize> {
+        let mut total_written = 0;
+        let mut temp_buffer = [0u8; MAX_RECORD_SIZE];
+        let (mut next_read_start, read_end, mut remaining) = {
+            let buffer = self.buffer.lock();
+
+            let tail = buffer.tail();
+            let len = (tail - buffer.head()).0;
+            let len_from_pick_head = {
+                let pick_head = Wrapping(self.pick_head.load(Ordering::Relaxed));
+                (tail - pick_head).0
+            };
+
+            let pick_len = len.min(len_from_pick_head).min(writer.avail());
+            if pick_len == 0 {
+                return Ok(total_written);
+            }
+
+            if clear_after_pick {
+                self.pick_head.store(tail.0, Ordering::Relaxed);
+            }
+
+            (tail - Wrapping(pick_len), tail, pick_len)
+        };
+
+        while remaining > 0 && writer.has_avail() {
+            let bytes_read = {
+                let buffer = self.buffer.lock();
+                let head = buffer.head();
+                let len = (buffer.tail() - head).0;
+
+                let read_end_offset = (read_end - head).0;
+                // The chosen end has fallen behind the current readable
+                // range, so none of the remaining snapshot bytes survived.
+                if read_end_offset > len {
+                    return Ok(total_written);
+                }
+
+                let read_start_offset = (next_read_start - head).0;
+                // The next byte to read was overwritten. Best-effort reads
+                // skip the lost prefix and resume from the oldest live byte.
+                if read_start_offset > len {
+                    next_read_start = head;
+                    remaining = read_end_offset;
+                    // The lost prefix reached the original end of this read.
+                    if remaining == 0 {
+                        return Ok(total_written);
+                    }
+                }
+
+                let chunk_len = remaining.min(temp_buffer.len()).min(writer.avail());
+                let chunk_end = next_read_start + Wrapping(chunk_len);
+                let mut temp_writer = VmWriter::from(&mut temp_buffer[..chunk_len]).to_fallible();
+                // The writer points to a valid kernel stack buffer, so the
+                // fallible copy cannot fault.
+                buffer
+                    .pick_range(next_read_start..chunk_end, &mut temp_writer)
+                    .unwrap()
+            };
+
+            let mut temp_reader = VmReader::from(&temp_buffer[..bytes_read]);
+            let bytes_written = match writer.write_fallible(&mut temp_reader) {
+                Ok(written) => written,
+                Err((err, 0)) if total_written == 0 => return Err(err),
+                Err((_, written)) => written,
+            };
+            total_written += bytes_written;
+
+            if bytes_written < bytes_read {
+                return Ok(total_written);
+            }
+
+            remaining -= bytes_written;
+            next_read_start += Wrapping(bytes_written);
+        }
+
+        Ok(total_written)
+    }
+
+    /// Marks the current buffer position as the next pick head.
+    ///
+    /// After this call, [`Self::pick_latest`] only returns messages logged after
+    /// this mark, but the data is not removed from the buffer.
+    pub fn mark_clear(&self) {
+        let buffer = self.buffer.lock();
+        self.pick_head.store(buffer.tail().0, Ordering::Relaxed);
+    }
+
+    /// Returns the number of unread bytes in the kernel log buffer.
+    pub fn size_unread(&self) -> usize {
+        self.buffer.lock().len()
+    }
+
+    /// Returns the total capacity of the kernel log buffer in bytes.
+    pub fn capacity(&self) -> usize {
+        LOG_BUFFER_CAPACITY
+    }
+
+    /// Returns the current value of the `dmesg_restrict` flag.
+    pub fn dmesg_restrict(&self) -> bool {
+        self.dmesg_restrict.load(Ordering::Relaxed)
+    }
+
+    /// Sets the `dmesg_restrict` flag.
+    ///
+    /// When enabled, reading all log messages requires `CAP_SYSLOG` or
+    /// `CAP_SYS_ADMIN`.
+    pub fn set_dmesg_restrict(&self, val: bool) {
+        self.dmesg_restrict.store(val, Ordering::Relaxed);
+    }
+
+    /// Sets the console log level.
+    ///
+    /// Returns the previous console log level.
+    pub fn set_console_level(&self, level: LinuxConsoleLogLevel) -> LinuxConsoleLogLevel {
+        let mut saved = self.saved_console_level.lock();
+        let old = self.console_level.load(Ordering::Relaxed);
+        self.console_level.store(level, Ordering::Relaxed);
+        *saved = None;
+        old
+    }
+
+    /// Disables console output by setting the log level to the Linux minimum.
+    ///
+    /// The previous log level is saved and can be restored with
+    /// [`Self::enable_console`]. Returns the previous console log level.
+    pub fn disable_console(&self) -> LinuxConsoleLogLevel {
+        let mut saved = self.saved_console_level.lock();
+        let old = self.console_level.load(Ordering::Relaxed);
+        self.console_level
+            .store(MINIMUM_CONSOLE_LOG_LEVEL, Ordering::Relaxed);
+        if saved.is_none() {
+            *saved = Some(old);
+        }
+        old
+    }
+
+    /// Restores the console log level that was saved by [`Self::disable_console`].
+    ///
+    /// Returns the restored console log level.
+    pub fn enable_console(&self) -> LinuxConsoleLogLevel {
+        let mut saved = self.saved_console_level.lock();
+        if let Some(prev) = saved.take() {
+            self.console_level.store(prev, Ordering::Relaxed);
+            prev
+        } else {
+            self.console_level()
+        }
+    }
+
+    /// Returns the current console log level.
+    pub fn console_level(&self) -> LinuxConsoleLogLevel {
+        self.console_level.load(Ordering::Relaxed)
+    }
+
+    /// Checks if a log message at the given level should be printed to console.
+    pub fn should_print(&self, level: Level) -> bool {
+        self.console_level().should_print(level)
+    }
+
+    /// Returns whether the destructive-read side of the buffer is empty.
+    pub fn is_empty(&self) -> bool {
+        self.buffer.lock().is_empty()
+    }
+
+    /// Returns the wait queue notified when new log bytes are appended.
+    pub fn wait_queue(&self) -> &WaitQueue {
+        &self.waitq
+    }
+}
--- a/kernel/libs/ring-buffer/src/lib.rs
+++ b/kernel/libs/ring-buffer/src/lib.rs
@@ -6,14 +6,21 @@ extern crate alloc;
 
 use alloc::sync::Arc;
 use core::{
+    fmt,
     marker::PhantomData,
     num::Wrapping,
-    ops::Deref,
+    ops::{Deref, Range},
     sync::atomic::{AtomicUsize, Ordering},
 };
 
 use inherit_methods_macro::inherit_methods;
-use ostd::mm::{FrameAllocOptions, PAGE_SIZE, Segment, VmIo, io::util::HasVmReaderWriter};
+use ostd::{
+    Error,
+    mm::{
+        FallibleVmRead, FrameAllocOptions, PAGE_SIZE, Segment, VmIo, VmWriter,
+        io::{Fallible, util::HasVmReaderWriter},
+    },
+};
 use ostd_pod::Pod;
 
 /// A lock-free single-producer single-consumer (SPSC) FIFO ring buffer.
@@ -217,6 +224,58 @@ impl<T> RingBuffer<T> {
 }
 
 impl RingBuffer<u8> {
+    /// Returns a formatter that appends text directly to the ring buffer.
+    ///
+    /// The formatter discards the oldest buffered bytes when new formatted
+    /// bytes exceed the available capacity.
+    pub fn formatter(&mut self) -> RingBufferFormatter<'_> {
+        RingBufferFormatter::new(self)
+    }
+
+    /// Picks bytes from the specified absolute ring counter range into `writer`.
+    ///
+    /// This method does not consume bytes from the buffer. The caller must
+    /// ensure that `range` is currently readable. The method handles wraparound
+    /// internally and advances `writer` by the number of bytes copied.
+    pub fn pick_range(
+        &self,
+        range: Range<Wrapping<usize>>,
+        writer: &mut VmWriter<'_, Fallible>,
+    ) -> Result<usize, (Error, usize)> {
+        fn copy_segment_to_writer(
+            segment: &Segment<()>,
+            offset: usize,
+            len: usize,
+            writer: &mut VmWriter<'_, Fallible>,
+        ) -> Result<usize, (Error, usize)> {
+            let mut reader = segment.reader();
+            reader.skip(offset).limit(len);
+            reader.read_fallible(writer)
+        }
+
+        let len = (range.end - range.start).0.min(writer.avail());
+        if len == 0 {
+            return Ok(0);
+        }
+
+        let offset = range.start.0 & (self.capacity - 1);
+        let mut copied = 0;
+        if offset + len > self.capacity {
+            let first_len = self.capacity - offset;
+            let first_copied = copy_segment_to_writer(self.segment(), offset, first_len, writer)
+                .map_err(|(err, copied_len)| (err, copied + copied_len))?;
+            copied += first_copied;
+            let second_copied = copy_segment_to_writer(self.segment(), 0, len - first_len, writer)
+                .map_err(|(err, copied_len)| (err, copied + copied_len))?;
+            copied += second_copied;
+        } else {
+            let segment_copied = copy_segment_to_writer(self.segment(), offset, len, writer)
+                .map_err(|(err, copied_len)| (err, copied + copied_len))?;
+            copied += segment_copied;
+        }
+        Ok(copied)
+    }
+
     /// Commits a read operation by advancing the head pointer.
     ///
     /// This method is intended for advanced use cases where the caller reads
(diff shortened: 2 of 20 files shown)


kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,303 @@
...
+    /// Returns the number of bytes written into `writer`.
+    pub fn read(&self, writer: &mut VmWriter<'_, Fallible>) -> ostd::Result<usize> {
+        let mut buf = self.buffer.lock();
+        let take = core::cmp::min(writer.avail(), buf.len());
+        if take == 0 {
+            return Ok(0);
+        }
+
+        let head = buf.head().0;
+        match buf.read_at(head, take, writer) {
```
lrh2000: Atomic mode break...

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+        if n == 0 {
+            // If we raced with another reader after waiting, wait again for the first byte.
+            if copied == 0 {
+                // Retry waiting for data.
+                continue;
+            }
+            break;
+        }
+        copied += n;
+    }
```
lrh2000: What is the purpose of the code? `pause_until(|| read())` will work just fine and cannot cause any observable behavior differences.

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /// Sets the console log level.
+    ///
+    /// Returns the previous console log level.
+    pub fn set_console_level(&self, level: LinuxConsoleLogLevel) -> LinuxConsoleLogLevel {
+        let mut saved = self.saved_console_level.lock();
+        let old_raw = self.console_level.swap(level as u8, Ordering::Relaxed);
+        let old = LinuxConsoleLogLevel::try_from(old_raw)
+            .expect("console_level stores only valid LinuxConsoleLogLevel values");
```
lrh2000: Can't we use `atomic_integer_wrapper`?

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub fn set_dmesg_restrict(&self, val: bool) {
+        self.dmesg_restrict.store(val, Ordering::Relaxed);
+    }
+
+    /// Sets the console log level.
+    ///
+    /// Returns the previous console log level.
+    pub fn set_console_level(&self, level: LinuxConsoleLogLevel) -> LinuxConsoleLogLevel {
+        let mut saved = self.saved_console_level.lock();
+        let old_raw = self.console_level.swap(level as u8, Ordering::Relaxed);
```
lrh2000: If an atomic variable can only be mutated under a lock, `store` + `load` is more efficient than `swap` and equally correct.

kernel/src/fs/fs_impls/procfs/sys/kernel/dmesg_restrict.rs
```diff
@@ -0,0 +1,57 @@
...
+            } else {
+                0
+            }
+        )?;
+        Ok(printer.bytes_written())
+    }
+
+    fn write_at(&self, offset: usize, reader: &mut VmReader) -> Result<usize> {
+        if offset != 0 {
+            return_errno_with_message!(Errno::EINVAL, "writes must start at offset 0");
```
lrh2000: I think other `procfs` files just ignore the offset (`_offset: usize`).

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+            // The set level must be an integer between 1 and 8 (inclusive).
+            //
+            // Reference: <https://man7.org/linux/man-pages/man2/syslog.2.html>
+            if !(1..=8).contains(&len) {
+                return_errno_with_message!(Errno::EINVAL, "console level out of range");
+            }
+
+            let level = len as u8;
+            let new_level = LinuxConsoleLogLevel::try_from(level)
+                .map_err(|_| Error::with_message(Errno::EINVAL, "invalid console level"))?;
```
lrh2000: This can be simplified. With some other minor fixes. I did not test the code and it may not even compile... Correct me if I misunderstood anything?
```patch
diff --git a/kernel/comps/logger/src/klog.rs b/kernel/comps/logger/src/klog.rs
index 6bf390ab2..f0b607a3a 100644
--- a/kernel/comps/logger/src/klog.rs
+++ b/kernel/comps/logger/src/klog.rs
@@ -51,8 +51,8 @@ pub const MINIMUM_CONSOLE_LOG_LEVEL: LinuxConsoleLogLevel = LinuxConsoleLogLevel
 ///
 /// These values correspond to Linux's console log level as documented in
 /// `man 2 syslog` and used by the `SYSLOG_ACTION_CONSOLE_LEVEL` action.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, TryFromInt)]
 #[repr(u8)]
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, TryFromInt)]
 pub enum LinuxConsoleLogLevel {
     /// Suppress all console output.
     Off = 0,
@@ -149,24 +149,13 @@ impl KernelLog {
     /// Returns the number of bytes written into `writer`.
     pub fn read(&self, writer: &mut VmWriter<'_, Fallible>) -> ostd::Result<usize> {
         let mut buf = self.buffer.lock();
-        let take = core::cmp::min(writer.avail(), buf.len());
-        if take == 0 {
-            return Ok(0);
-        }
 
-        let head = buf.head().0;
-        match buf.read_at(head, take, writer) {
-            Ok(written) => {
-                buf.commit_read(written);
-                Ok(written)
-            }
-            Err((err, written)) => {
-                if written > 0 {
-                    buf.commit_read(written);
-                }
-                Err(err)
-            }
-        }
+        let written = buf
+            .read_at(buf.head().0, buf.len(), writer)
+            .map_err(|(err, _)| err)?;
+        buf.commit_read(written);
+
+        Ok(written)
     }
 
     /// Reads log messages from the kernel log buffer non-destructively.
@@ -174,35 +163,23 @@ impl KernelLog {
     /// Writes the newest available log bytes that fit in `writer`. The data
     /// remains in the buffer after reading.
     pub fn read_all(&self, writer: &mut VmWriter<'_, Fallible>) -> ostd::Result<usize> {
-        /// Returns the first byte visible to non-destructive read-all operations.
-        fn read_all_visible_head(
-            head: Wrapping<usize>,
-            tail: Wrapping<usize>,
-            clear_head: Wrapping<usize>,
-        ) -> Wrapping<usize> {
-            let len = (tail - head).0;
-            if (clear_head - head).0 <= len {
-                clear_head
-            } else {
-                head
-            }
-        }
-
         let buf = self.buffer.lock();
-        let head = buf.head();
+
         let tail = buf.tail();
-        let clear_head = Wrapping(self.clear_head.load(Ordering::Relaxed));
-        let visible_head = read_all_visible_head(head, tail, clear_head);
-        let available = (tail - visible_head).0;
-        let take = core::cmp::min(available, writer.avail());
-        if take == 0 {
-            return Ok(0);
-        }
+        let len = (tail - buf.head()).0;
+        let len_from_clear_head = {
+            let clear_head = Wrapping(self.clear_head.load(Ordering::Relaxed));
+            (tail - clear_head).0
+        };
+
+        let read_len = len.min(len_from_clear_head).min(writer.remain());
+        let read_start = (tail - Wrapping(read_len)).0;
 
-        let start = (tail - Wrapping(take)).0;
-        let copied = buf.read_at(start, take, writer).map_err(|(err, _)| err)?;
+        let written = buf
+            .read_at(read_start, read_len, writer)
+            .map_err(|(err, _)| err)?;
 
-        Ok(copied)
+        Ok(written)
     }
 
     /// Marks the current buffer position as cleared.
diff --git a/kernel/src/fs/fs_impls/procfs/sys/kernel/dmesg_restrict.rs b/kernel/src/fs/fs_impls/procfs/sys/kernel/dmesg_restrict.rs
index e72b62064..08fc00526 100644
--- a/kernel/src/fs/fs_impls/procfs/sys/kernel/dmesg_restrict.rs
+++ b/kernel/src/fs/fs_impls/procfs/sys/kernel/dmesg_restrict.rs
@@ -23,15 +23,14 @@ impl DmesgRestrictFileOps {
 impl ProcFileOps for DmesgRestrictFileOps {
     fn read_at(&self, offset: usize, writer: &mut VmWriter) -> Result<usize> {
-        writeln!(
-            printer,
-            "{}",
-            if aster_logger::klog().dmesg_restrict() {
-                1
-            } else {
-                0
-            }
-        )?;
+
+        let dmesg_restrict = if aster_logger::klog().dmesg_restrict() {
+            1
+        } else {
+            0
+        };
+        writeln!(printer, "{}", dmesg_restrict)?;
+
         Ok(printer.bytes_written())
     }
 
@@ -45,13 +44,15 @@ impl ProcFileOps for DmesgRestrictFileOps {
         }
 
         let (val, read_len) = read_i32_from(reader)?;
-        let val = match val {
+        let dmesg_restrict = match val {
             0 => false,
             1 => true,
-            _ => return_errno_with_message!(Errno::EINVAL, "expected 0 or 1"),
+            _ => {
+                return_errno_with_message!(Errno::EINVAL, "`dmesg_restrict` must be either 0 or 1")
+            }
         };
 
-        aster_logger::klog().set_dmesg_restrict(val);
+        aster_logger::klog().set_dmesg_restrict(dmesg_restrict);
         Ok(read_len)
     }
 }
diff --git a/kernel/src/syscall/syslog.rs b/kernel/src/syscall/syslog.rs
index 14ee93d37..7a345a22f 100644
--- a/kernel/src/syscall/syslog.rs
+++ b/kernel/src/syscall/syslog.rs
@@ -1,5 +1,7 @@
 
+use core::num::NonZeroUsize;
+
 use aster_logger::LinuxConsoleLogLevel;
 use int_to_c_enum::TryFromInt;
 
@@ -10,8 +12,8 @@ use crate::{prelude::*, process::credentials::capabilities::CapSet};
 ///
 /// These actions control how the kernel log buffer is accessed and managed.
 /// See `man 2 syslog` for detailed documentation.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromInt)]
 #[repr(i32)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromInt)]
 enum SysLogAction {
     /// Close the log (currently a no-op).
     Close = 0,
@@ -25,13 +27,13 @@ enum SysLogAction {
     ReadClear = 4,
     /// Clear the ring buffer.
     Clear = 5,
-    /// Disable printk to console.
+    /// Disable printing of messages to the console.
     ConsoleOff = 6,
-    /// Enable printk to console.
+    /// Enable printing of messages to the console.
     ConsoleOn = 7,
     /// Set the console log level.
     ConsoleLevel = 8,
-    /// Return number of unread characters in the log buffer.
+    /// Return the number of unread characters in the log buffer.
     SizeUnread = 9,
     /// Return the size of the log buffer.
     SizeBuffer = 10,
@@ -42,26 +44,33 @@ pub fn sys_syslog(action: i32, buf: Vaddr, len: i32, ctx: &Context) -> Result<Sy
         check_cap(ctx)?;
     }
 
-    let action = SysLogAction::try_from(action)
-        .map_err(|_| Error::with_message(Errno::EINVAL, "unknown syslog action"))?;
+    let action = SysLogAction::try_from(action)?;
 
     match action {
         SysLogAction::Close | SysLogAction::Open => Ok(SyscallReturn::Return(0)),
         SysLogAction::Read => {
             let len = validate_read_args(buf, len)?;
+            let Some(len) = NonZeroUsize::new(len) else {
+                return Ok(SyscallReturn::Return(0));
+            };
+
             Ok(SyscallReturn::Return(
                 read_destructive(buf, len, ctx)? as isize
             ))
         }
         SysLogAction::ReadAll => {
             let len = validate_read_args(buf, len)?;
+            let Some(len) = NonZeroUsize::new(len) else {
+                return Ok(SyscallReturn::Return(0));
+            };
+
             Ok(SyscallReturn::Return(read_all(buf, len, ctx)? as isize))
         }
         SysLogAction::ReadClear => {
             let len = validate_read_args(buf, len)?;
-            if len == 0 {
+            let Some(len) = NonZeroUsize::new(len) else {
                 return Ok(SyscallReturn::Return(0));
-            }
+            };
 
             let copied = read_all(buf, len, ctx)?;
             aster_logger::klog().mark_clear();
@@ -81,16 +90,13 @@ pub fn sys_syslog(action: i32, buf: Vaddr, len: i32, ctx: &Context) -> Result<Sy
             Ok(SyscallReturn::Return(0))
         }
         SysLogAction::ConsoleLevel => {
-            // The set level must be an integer between 1 and 8 (inclusive).
-            //
-            // Reference: <https://man7.org/linux/man-pages/man2/syslog.2.html>
-            if !(1..=8).contains(&len) {
-                return_errno_with_message!(Errno::EINVAL, "console level out of range");
-            }
-
-            let level = len as u8;
-            let new_level = LinuxConsoleLogLevel::try_from(level)
-                .map_err(|_| Error::with_message(Errno::EINVAL, "invalid console level"))?;
+            let new_level = if let Ok(raw_level) = u8::try_from(len)
+                && let Ok(level) = LinuxConsoleLogLevel::try_from(level)
+            {
+                level
+            } else {
+                return_errno_with_message!(Errno::EINVAL, "the console level is out of range");
+            };
 
             let _old = aster_logger::klog().set_console_level(new_level);
             Ok(SyscallReturn::Return(0))
@@ -125,18 +131,21 @@ fn action_requires_cap(action: i32) -> bool {
 /// Checks whether the current thread may perform privileged syslog actions.
 fn check_cap(ctx: &Context) -> Result<()> {
     let credentials = ctx.posix_thread.credentials();
-    if credentials
+    if !credentials
         .effective_capset()
         .intersects(CapSet::SYS_ADMIN | CapSet::SYSLOG)
     {
-        return Ok(());
+            Errno::EPERM,
+            "the thread does not have the required capability"
+        )
     }
-    return_errno_with_message!(Errno::EPERM, "operation not permitted")
+    Ok(())
 }
 
 fn validate_read_args(buf: Vaddr, len: i32) -> Result<usize> {
     if buf == 0 || len < 0 {
-        return_errno_with_message!(Errno::EINVAL, "invalid syslog read arguments");
+        return_errno_with_message!(Errno::EINVAL, "syslog read arguments are invalid");
     }
 
     Ok(len as usize)
@@ -146,13 +155,10 @@ fn validate_read_args(buf: Vaddr, len: i32) -> Result<usize> {
 ///
 /// This action blocks until at least one byte is available, then drains up to
 /// `len` bytes into the user buffer. A zero-length read returns immediately.
-fn read_destructive(buf: Vaddr, len: usize, ctx: &Context) -> Result<usize> {
-    if len == 0 {
-        return Ok(0);
-    }
+fn read_destructive(buf: Vaddr, len: NonZeroUsize, ctx: &Context) -> Result<usize> {
     let mut copied = 0;
-    let mut writer = user_space.writer(buf, len)?;
+    let mut writer = user_space.writer(buf, len.get())?;
     let klog = aster_logger::klog();
 
     while copied < len {
@@ -180,13 +186,9 @@ fn read_destructive(buf: Vaddr, len: usize, ctx: &Context) -> Result<usize> {
 }
 
 /// Reads kernel log bytes without consuming them.
-fn read_all(buf: Vaddr, len: usize, ctx: &Context) -> Result<usize> {
-    if len == 0 {
-        return Ok(0);
-    }
-
+fn read_all(buf: Vaddr, len: NonZeroUsize, ctx: &Context) -> Result<usize> {
-    let mut writer = user_space.writer(buf, len)?;
+    let mut writer = user_space.writer(buf, len.get())?;
     let klog = aster_logger::klog();
 
     Ok(klog.read_all(&mut writer)?)
```

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+fn read_all(buf: Vaddr, len: usize, ctx: &Context) -> Result<usize> {
+    if len == 0 {
+        return Ok(0);
+    }
+
+    let user_space = ctx.user_space();
+    let mut writer = user_space.writer(buf, len)?;
+    let klog = aster_logger::klog();
+
+    Ok(klog.read_all(&mut writer)?)
```
lrh2000: I don't think `read_all` is an accurate name. Perhaps `read_last`? Or other better names?

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+            Ok(SyscallReturn::Return(read_all(buf, len, ctx)? as isize))
+        }
+        SysLogAction::ReadClear => {
+            let len = validate_read_args(buf, len)?;
+            if len == 0 {
+                return Ok(SyscallReturn::Return(0));
+            }
+
+            let copied = read_all(buf, len, ctx)?;
+            aster_logger::klog().mark_clear();
```
lrh2000: I think this is racy.

kernel/src/syscall/syslog.rs
cchanging: This function corresponds to `SysLogAction::ReadAll`. The main difference between it and the `Read` action is that `ReadAll` does not remove bytes from the kernel log, whereas `Read` does. I’m thinking maybe the operation corresponding to `Read` could be called `read_destructive`, and `ReadAll` could be called `read_non_destructive`? Or perhaps we could distinguish them as `read` and `copy`? `read_last` only reflects the behavior when the user buffer is too small, which doesn’t feel general enough.

lrh2000: > `read_last` only reflects the behavior when the user buffer is too small, which doesn’t feel general enough.

Actually, when I see a function named `read_all`, my first question is what will happen if the buffer is small. How can all messages be read into a smaller buffer?

> I’m thinking maybe the operation corresponding to `Read` could be called `read_destructive`, and `ReadAll` could be called `read_non_destructive`?

I think a more common naming rule is: `read_destructive` -> `read`, `read_non_destructive` -> `peek`.

If the buffer contains "123" and we want to peek at one byte, I assume the default is to peek from the beginning, so we would see the "1." In this case, however, we actually peek the last byte, "3." Therefore, I think it should be "peek" plus a word that indicates we start from the end.

No matter whether we peek from the start or the end, we'll get the same result if the user buffer is large enough.

Note, however, that we also have `clear_head`. Therefore, things may be slightly complex, and "peek" may not be 100% accurate. It should be accurate if we only consider the buffer head.

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// Returns whether the raw syslog action requires privilege.
+///
+/// When `dmesg_restrict` is enabled, every syslog action requires
+/// `CAP_SYS_ADMIN` or `CAP_SYSLOG`. Otherwise, Linux allows
+/// `SYSLOG_ACTION_READ_ALL` and `SYSLOG_ACTION_SIZE_BUFFER` without these
+/// capabilities, while all other actions, including unknown action numbers,
+/// remain privileged.
+fn action_requires_cap(action: i32) -> bool {
+    if aster_logger::klog().dmesg_restrict() {
```
lrh2000: I'm wondering if we should allow the methods in this file to accept `klog` so that we won't need to query `aster_logger::klog()` repeatedly.

We only need to do it once at the beginning of the system call. All other usages of `aster_logger::klog()` should be removed.

kernel/comps/logger/src/klog.rs
lrh2000: https://github.com/asterinas/asterinas/pull/3288#discussion_r3315561737 To avoid this atomic mode break, consider putting the producer in a spin lock and the consumer in a mutex, as shown below:
https://github.com/asterinas/asterinas/blob/5831c49a9a1c9c7d512fe12f200dd1c7575fd0bf/kernel/src/device/evdev/mod.rs#L55
https://github.com/asterinas/asterinas/blob/5831c49a9a1c9c7d512fe12f200dd1c7575fd0bf/kernel/src/device/evdev/file.rs#L125

I am not  sure if there are other blockers of doing so.

cchanging: > To avoid this atomic mode break, consider putting the producer in a spin lock and the consumer in a mutex, as shown below

I tried it, but it doesn’t seem to work. The main issue is that the produce behavior here may fill up the buffer, so consuming needs to happen at the same time. This means we would need to hold both locks while producing, which could cause log operations performed in interrupt context to potentially sleep, leading to problems.

I think we might still need to use an extra buffer during read operations to avoid the atomic mode issue...

lrh2000: > The main issue is that the produce behavior here may fill up the buffer, so consuming needs to happen at the same time.

This is true. Alternatively, we could copy the logs to userspace without holding the lock. Then, we can retry if we find that the copied region has been overwritten by new logs. We don't need to worry about leaking information because, even if they are overwritten, the copied logs are still kernel logs.

However, I don't know how this can be implemented via reasonable and safe Rust APIs. Do you have any ideas?

In addition, have you checked how Linux handles this situation?

cchanging: > In addition, have you checked how Linux handles this situation?

https://codebrowser.dev/linux/linux/kernel/printk/printk.c.html#syslog_print

I just checked, and Linux seems to allocate a temporary cache as well. Linux's copy granularity appears to be based on records. We don't yet have a complete definition of records, including their headers, metadata, etc.

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,214 @@
...
+                .pause_until(|| match klog.read(&mut writer) {
+                    Ok(0) => None,
+                    res => Some(res),
+                })
+                .map_err(|err| match err.error() {
+                    Errno::EINTR => Error::new(Errno::ERESTARTSYS),
+                    _ => err,
+                })??
+        } else {
+            match klog.read(&mut writer) {
```
lrh2000: Can you reply to my question in https://github.com/asterinas/asterinas/pull/3288#discussion_r3315566786? I don't think we need a loop here.

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,214 @@
...
+            Ok(SyscallReturn::Return(0))
+        }
+        SysLogAction::ConsoleOn => {
+            klog.enable_console();
+            Ok(SyscallReturn::Return(0))
+        }
+        SysLogAction::ConsoleLevel => {
+            // The set level must be an integer between 1 and 8 (inclusive).
+            //
+            // Reference: <https://man7.org/linux/man-pages/man2/syslog.2.html>
```
lrh2000: Can you explain why you think it is useful and preferred than `u8::try_from` version?
https://github.com/asterinas/asterinas/pull/3288#discussion_r3315744889

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,394 @@
...
+                match buffer.pick_range(read_start..read_end, &mut temp_writer) {
+                    Ok(written) => {
+                        buffer.commit_read(written);
+                        written
+                    }
+                    Err((err, 0)) => return Err(err),
+                    Err((_, written)) => {
+                        buffer.commit_read(written);
+                        written
+                    }
```
lrh2000: How can this fail?

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Reads log messages from the kernel log buffer destructively.
+    ///
+    /// Returns the number of bytes written into `writer`.
+    pub fn read(&self, writer: &mut VmWriter<'_, Fallible>) -> ostd::Result<usize> {
+        let mut total_written = 0;
+        let mut temp_buffer = [0u8; MAX_RECORD_SIZE];
+
+        while writer.has_avail() {
+            let bytes_read = {
+                let mut buffer = self.buffer.lock();
```
lrh2000, line 167: The problem with this solution is that we will end up with mixed log messages if two processes try to read the kernel logs at the same time.

```
info!("abcd");
info!("def");
```

Two users will see `abcef` (User A) and `dd` (User B). It does not seem very correct.

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,394 @@
...
+            }
+
+            let chunk_end = next_read_start + Wrapping(chunk_len);
+            let bytes_read = {
+                let buffer = self.buffer.lock();
+                let mut temp_writer = VmWriter::from(&mut temp_buffer[..chunk_len]).to_fallible();
+                match buffer.pick_range(next_read_start..chunk_end, &mut temp_writer) {
+                    Ok(written) => written,
+                    Err((err, 0)) => return Err(err),
+                    Err((_, written)) => written,
```
lrh2000: How can this fail?

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,394 @@
...
+                if read_len == 0 {
+                    return Ok(total_written);
+                }
+
+                let read_start = buffer.head();
+                let read_end = read_start + Wrapping(read_len);
+                let mut temp_writer = VmWriter::from(&mut temp_buffer[..read_len]).to_fallible();
+                match buffer.pick_range(read_start..read_end, &mut temp_writer) {
+                    Ok(written) => {
+                        buffer.commit_read(written);
```
lrh2000: Also should we commit even if the following `write_fallible` fails? But if we don't do this, it is also racy.

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,394 @@
...
+                    }
+                    Err((err, 0)) => return Err(err),
+                    Err((_, written)) => {
+                        buffer.commit_read(written);
+                        written
+                    }
+                }
+            };
+
+            if bytes_read == 0 {
```
lrh2000: I feel that there are too many `if ... == 0` checks in this PR. I removed some in https://github.com/asterinas/asterinas/pull/3288#discussion_r3315744889, but the logic was rewritten, so new checks appeared. Have you checked that every `if ... == 0` is not redundant and can provide real values?

kernel/libs/ring-buffer/src/lib.rs
```diff
@@ -236,6 +296,85 @@ impl RingBuffer<u8> {
...
+    /// longer be present in the ring buffer if later writes through this
+    /// formatter overwrote older data.
+    pub fn bytes_written(&self) -> usize {
+        self.bytes_written
+    }
+
+    fn push_bytes(&mut self, mut bytes: &[u8]) {
+        let cap = self.rb.capacity();
+        let current_len = self.rb.len();
+        let overflow = current_len.saturating_add(bytes.len()).saturating_sub(cap);
```
lrh2000: Is the first `saturating_add` necessary?

Also I found some variables are defined without good reason, e.g., why we define `cap` here? It is only used once. I fixed some occurrences in https://github.com/asterinas/asterinas/pull/3288#discussion_r3315744889, but there are others..

`cap` does not provide new information compared to `self.rb.capacity()`. We should define a variable if it is used multiple times or the variable name provides new information.

For example, `new_len` will also be used once below, but I think it provides new information, and it serves a counterpart as the `current_len` variable (which will be used multiple times), so it is a good variable to define:
```rust
        let current_len = self.rb.len();
        let new_len = current_len + bytes.len();
        let overflow = new_len.saturating_sub(self.rb.capacity());
        if overflow > 0 {
```

kernel/libs/ring-buffer/src/lib.rs
```diff
@@ -236,6 +296,85 @@ impl RingBuffer<u8> {
...
+            }
+            let drop_from_input = overflow - drop_from_buffer;
+            if drop_from_input > 0 {
+                bytes = &bytes[drop_from_input..];
+            }
+        }
+
+        self.rb
+            .push_slice(bytes)
+            .expect("push_slice must succeed after dropping enough bytes");
```
lrh2000:
```suggestion
            .expect("`push_slice` must succeed after dropping enough bytes");
```

kernel/src/syscall/syslog.rs
cchanging: Yes, this loop is indeed redundant. I will correct it.

kernel/src/syscall/syslog.rs
cchanging: My original intention was to filter out invalid values with level 0. But after checking again, I feel that `LinuxConsoleLogLevel::Off` shouldn't exist at all, so this case really doesn't need to be handled. I'll make a correction.

kernel/comps/logger/src/klog.rs
cchanging: https://codebrowser.dev/linux/linux/kernel/printk/printk.c.html#1646

Our behavior here should be consistent with Linux. Linux also consumes one record first, and then calls `copy_to_user`. Even if that fails, the record has already been consumed.

kernel/comps/logger/src/klog.rs
cchanging: In Linux, the read operation consumes data byte by byte, so it is possible that a record is half-read, and then the next user reads the remaining part.

I asked an LLM to write a test, and the output was like this: the first read only read the first half of the first record, and the second read read the second half of the first record along with the second record. So I think this kind of race might not cause a bad impact?

```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/syscall.h>
#include <time.h>
#include <unistd.h>

#ifndef SYS_syslog
#error "SYS_syslog is not available on this platform"
#endif

#define SYSLOG_ACTION_READ 2
#define SYSLOG_ACTION_SIZE_UNREAD 9

static long ksyslog(int type, char *buf, int len)
{
	return syscall(SYS_syslog, type, buf, len);
}

static void die(const char *what)
{
	perror(what);
	exit(1);
}

static void drain_klog(void)
{
	for (;;) {
		long unread = ksyslog(SYSLOG_ACTION_SIZE_UNREAD, NULL, 0);
		if (unread < 0)
			die("syslog(SIZE_UNREAD)");
		if (unread == 0)
			return;

		char buf[4096];
		int to_read = unread > (long)sizeof(buf) ? (int)sizeof(buf) : (int)unread;
		long ret = ksyslog(SYSLOG_ACTION_READ, buf, to_read);
		if (ret < 0)
			die("syslog(READ) while draining");
	}
}

static void write_kmsg(int fd, const char *message)
{
	char line[512];
	int len = snprintf(line, sizeof(line), "<6>%s\n", message);
	if (len < 0 || len >= (int)sizeof(line)) {
		fprintf(stderr, "message too long\n");
		exit(1);
	}

	ssize_t written = write(fd, line, (size_t)len);
	if (written != len)
		die("write(/dev/kmsg)");
}

static void dump_read(const char *name, const char *buf, long len)
{
	printf("%s: %ld bytes\n", name, len);
	fwrite(buf, 1, (size_t)len, stdout);
	if (len == 0 || buf[len - 1] != '\n')
		putchar('\n');
	putchar('\n');
}

int main(void)
{
	char tag[64];
	snprintf(tag, sizeof(tag), "syslog-partial-poc-%ld-%ld", (long)getpid(), (long)time(NULL));

	char first[384];
	snprintf(first, sizeof(first),
		 "%s FIRST begin "
		 "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
		 " bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
		 " cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"
		 " FIRST end",
		 tag);

	char second[160];
	snprintf(second, sizeof(second), "%s SECOND complete line", tag);

	drain_klog();

	int fd = open("/dev/kmsg", O_WRONLY | O_CLOEXEC);
	if (fd < 0)
		die("open(/dev/kmsg)");

	write_kmsg(fd, first);
	write_kmsg(fd, second);
	close(fd);

	char read1[80];
	char read2[4096];

	long len1 = ksyslog(SYSLOG_ACTION_READ, read1, sizeof(read1));
	if (len1 < 0)
		die("syslog(READ #1)");

	long len2 = ksyslog(SYSLOG_ACTION_READ, read2, sizeof(read2));
	if (len2 < 0)
		die("syslog(READ #2)");

	dump_read("read #1, intentionally smaller than the first record", read1, len1);
	dump_read("read #2, continues first record and then reads the second", read2, len2);

	if (memmem(read1, (size_t)len1, "FIRST begin", strlen("FIRST begin")) == NULL) {
		fprintf(stderr, "did not find the beginning of the first message in read #1\n");
		return 1;
	}
	if (memmem(read2, (size_t)len2, "FIRST end", strlen("FIRST end")) == NULL) {
		fprintf(stderr, "did not find the end of the first message in read #2\n");
		return 1;
	}
	if (memmem(read2, (size_t)len2, "SECOND complete line", strlen("SECOND complete line")) == NULL) {
		fprintf(stderr, "did not find the complete second message in read #2\n");
		return 1;
	}

	return 0;
}
```

```
read #1, intentionally smaller than the first record: 80 bytes
<14>[2087334.235366] syslog-partial-poc-3203908-1780037013 FIRST begin aaaaaaaaa

read #2, continues first record and then reads the second: 268 bytes
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc FIRST end
<14>[2087334.235373] syslog-partial-poc-3203908-1780037013 SECOND complete line
```

kernel/src/syscall/syslog.rs
lrh2000: If it has to exist, we can make the check more explicit by using either `raw_level != 0` or `level != LinuxConsoleLogLevel::Off`.

I was unable to understand the intention of the previous one.

Of course, just removing the `Off` variant may be even better.

kernel/comps/logger/src/klog.rs
lrh2000: Sorry, I edited my previous reply. I am not sure if you see the latest version:

>> Two users will see `abcef` (User A) and `dd` (User B). It does not seem very correct.

The first read can see `abcef` which is the head part of the first record plus the tail part of the second record.

The first record: $\color{#FF0000}{abc}$ $\color{#00FF00}{d}$
The second record: $\color{#00FF00}{d}$ $\color{#FF0000}{ef}$

Well, on the other hand, Linux does drop the mutex in the middle, so perhaps it is acceptable either: https://codebrowser.dev/linux/linux/kernel/printk/printk.c.html#1660

kernel/comps/logger/src/klog.rs
lrh2000: It seems to be the exact case. Thanks for your explanation!

kernel/libs/ring-buffer/src/lib.rs
```diff
@@ -236,6 +295,84 @@ impl RingBuffer<u8> {
...
+    ///
+    /// This counts bytes passed through the formatter limit. The bytes may no
+    /// longer be present in the ring buffer if later writes through this
+    /// formatter overwrote older data.
+    pub fn bytes_written(&self) -> usize {
+        self.bytes_written
+    }
+
+    fn push_bytes(&mut self, mut bytes: &[u8]) {
+        let current_len = self.rb.len();
```
lrh2000: Shouldn't this be moved to the code block below?
```rust
        let overflow = bytes.len().saturating_sub(self.rb.free_len());
        if overflow > 0 {
            let current_len = self.rb.len();
            let drop_from_buffer = overflow.min(current_len);
```
Or just:
```rust
        let overflow = bytes.len().saturating_sub(self.rb.free_len());
        if overflow > 0 {
            let drop_from_buffer = overflow.min(self.rb.len());
```

kernel/libs/ring-buffer/src/test.rs
```diff
@@ -4,6 +4,58 @@ use ostd::prelude::ktest;
...
+    rb.push_slice(b"123456").unwrap();
+
+    let mut output = [0u8; 8];
+    let mut writer = VmWriter::from(&mut output[..]).to_fallible();
+    let pick_start = rb.head();
+    let pick_end = pick_start + Wrapping(rb.len());
+    let copied = rb.pick_range(pick_start..pick_end, &mut writer).unwrap();
+    assert_eq!(copied, output.len());
+    assert_eq!(&output, b"ef123456");
+}
```
lrh2000: I think all the new tests should be moved to the end of the file. They do not seem to be more important than the first test `rb_basics` (which is the "basics" test!)

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,374 @@
...
+                    .pick_range(read_start..read_end, &mut temp_writer)
+                    .unwrap();
+                buffer.commit_read(written);
+                written
+            };
+
+            let mut temp_reader = VmReader::from(&temp_buffer[..bytes_read]);
+            let bytes_written = match writer.write_fallible(&mut temp_reader) {
+                Ok(written) => written,
+                Err((err, 0)) => return Err(err),
```
lrh2000: I assume that we shouldn't return `Err(err)` if `total_written > 0`. We should return `Ok(total_written)` instead.

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,374 @@
...
+                // fallible copy cannot fault.
+                buffer
+                    .pick_range(next_read_start..chunk_end, &mut temp_writer)
+                    .unwrap()
+            };
+
+            let mut temp_reader = VmReader::from(&temp_buffer[..bytes_read]);
+            let bytes_written = match writer.write_fallible(&mut temp_reader) {
+                Ok(written) => written,
+                Err((err, 0)) => return Err(err),
```
lrh2000: Same here.

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,374 @@
...
+            let chunk_len = remaining.min(temp_buffer.len()).min(writer.avail());
+
+            let chunk_end = next_read_start + Wrapping(chunk_len);
+            let bytes_read = {
+                let buffer = self.buffer.lock();
+                let mut temp_writer = VmWriter::from(&mut temp_buffer[..chunk_len]).to_fallible();
+                // The writer points to a valid kernel stack buffer, so the
+                // fallible copy cannot fault.
+                buffer
+                    .pick_range(next_read_start..chunk_end, &mut temp_writer)
```
lrh2000, line 282: What if the range has been overwritten after the lock is released?

`pick_latest_impl` acquires the same lock multiple times. This means that there are many potential causes of the range being overwritten. Should we address this situation properly?

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,196 @@
...
+            };
+
+            Ok(SyscallReturn::Return(
+                read_destructive(buf, len, ctx, klog)? as isize,
+            ))
+        }
+        SysLogAction::ReadAll => {
+            let len = validate_read_args(buf, len)?;
+            let Some(len) = NonZeroUsize::new(len) else {
+                return Ok(SyscallReturn::Return(0));
```
lrh2000, line 66: Confirm that this is actually a no-op instead of `ReadClear`?

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,196 @@
...
+        return_errno_with_message!(Errno::EINVAL, "syslog read arguments are invalid");
+    }
+
+    Ok(len as usize)
+}
+
+/// Reads and removes kernel log bytes for `SYSLOG_ACTION_READ`.
+///
+/// This action blocks until at least one byte is available, then drains up to
+/// `len` bytes into the user buffer. A zero-length read returns immediately.
```
lrh2000:
```suggestion
/// `len` bytes into the user buffer.
```

This method accepts a `len: NonZeroUsize` now.

lrh2000 (commented):
Thanks for revising the code. I am still a bit concerned about the races in `pick_latest_impl`. Other comments seem to be minor.

kernel/src/syscall/syslog.rs
cchanging: https://github.com/torvalds/linux/blob/f5e5d3509bffb95c6648eb9795f7f236852ae62d/kernel/printk/printk.c#L1774

Linux returns directly when `len == 0`, and currently it should be aligned with Linux.

kernel/comps/logger/src/klog.rs
cchanging: https://github.com/torvalds/linux/blob/master/kernel/printk/printk.c#L1561C1-L1566C5

I checked the relevant Linux implementation. Linux's related operations for this are actually best effort as well. I think our operations also need to check whether the current interval to be read is valid after each re-lock. If it's not valid, we need to skip it.

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,408 @@
...
+            let mut temp_reader = VmReader::from(&temp_buffer[..bytes_read]);
+            let bytes_written = match writer.write_fallible(&mut temp_reader) {
+                Ok(written) => written,
+                Err((err, 0)) => {
+                    return if total_written > 0 {
+                        Ok(total_written)
+                    } else {
+                        Err(err)
+                    };
+                }
```
lrh2000:
```suggestion
                Err((err, 0)) if total_written == 0 => return Err(err),
```

I think we only need to have this. Otherwise, we will hit the logic below:
```rust
            if bytes_written < bytes_read {
                return Ok(total_written);
            }
```

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,408 @@
...
```
lrh2000: Same here.

kernel/comps/logger/src/klog.rs
lrh2000: Thanks for your explanation and revision. The current version looks fine to me.

kernel/src/syscall/syslog.rs
```diff
@@ -0,0 +1,196 @@
...
+fn check_cap(ctx: &Context) -> Result<()> {
+    let credentials = ctx.posix_thread.credentials();
+    if !credentials
+        .effective_capset()
+        .intersects(CapSet::SYS_ADMIN | CapSet::SYSLOG)
+    {
+        return_errno_with_message!(
+            Errno::EPERM,
+            "the thread does not have the required capability"
+        )
```
lrh2000:
```suggestion
        );
```

kernel/src/fs/fs_impls/procfs/sys/kernel/dmesg_restrict.rs
```diff
@@ -0,0 +1,54 @@
...
+        if reader.remain() == 0 {
+            return Ok(0);
+        }
+
+        let (val, read_len) = read_i32_from(reader)?;
+        let dmesg_restrict = match val {
+            0 => false,
+            1 => true,
+            _ => {
+                return_errno_with_message!(Errno::EINVAL, "`dmesg_restrict` must be either 0 or 1")
```
lrh2000:
```suggestion
                return_errno_with_message!(Errno::EINVAL, "`dmesg_restrict` must be either 0 or 1");
```

kernel/comps/logger/src/klog.rs
```diff
@@ -0,0 +1,408 @@
...
+
+            (tail - Wrapping(pick_len), tail, pick_len)
+        };
+
+        while remaining > 0 && writer.has_avail() {
+            let bytes_read = {
+                let buffer = self.buffer.lock();
+                let head = buffer.head();
+                let tail = buffer.tail();
+                let len = (tail - head).0;
```
lrh2000:
```suggestion
                let len = (buffer.tail() - head).0;
```

lrh2000 (approved):
Thanks for the updates. Just some minor nits. Otherwise LGTM.

tatetian:
I thought this syscall is close to completion. But after a careful review, I find that more rounds of revision are needed to land this PR. So let's not rush it for the 0.18.0 release. I will post full review comments later. Thanks for the patience!

PR #3290, opened 2026-05-28
Complete `/proc/pid/stat` output for ps compatibility

StevenJiang1110:
This PR fixes a ps compatibility issue I hit when running Asterinas as the kata guest kernel: running `ps` in the guest could trigger a segmentation fault because `/proc/[pid]/stat` was incomplete.

Previously, only a subset of the 52 `/proc/[pid]/stat` fields was implemented. This PR fills in the missing fields and also reviews the existing ones. Wherever the kernel can provide a real value, the field now uses it. Where no real value is available yet, the field remains `0` and is annotated with a comment to mark it as a placeholder.

The second commit is a cleanup commit that refactors some scheduler-related code by replacing the previously defined constants with the `LinuxSchedPolicy` enum.

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use core::sync::atomic::Ordering;
+use core::{sync::atomic::Ordering, time::Duration};
 
 use aster_util::printer::VmPrinter;
+use ostd::timer::TIMER_FREQ;
 
 use super::{super::PidDirOps, TidDirOps};
 use crate::{
@@ -12,7 +13,12 @@ use crate::{
         vfs::inode::Inode,
     },
     prelude::*,
-    process::posix_thread::{AsPosixThread, SleepingState},
+    process::{
+        Process, ResourceType,
+        posix_thread::{AsPosixThread, SleepingState},
+        signal::{HandlePendingSignal, sig_action::SigAction, sig_mask::SigMask},
+    },
+    sched::{LinuxSchedPolicy, RealTimePriority, SchedPolicy},
     thread::Thread,
     vm::vmar::RssType,
 };
@@ -42,16 +48,17 @@ use crate::{
 /// - stime            : Time spent in kernel mode (clock ticks).
 /// - cutime           : Child processes' user mode time.
 /// - cstime           : Child processes' kernel mode time.
-/// - priority         : Process priority or nice value.
+/// - priority         : Kernel scheduling priority.
 /// - nice             : Nice value (-20 to 19; lower is higher priority).
 /// - num_threads      : Number of threads.
+/// - itrealvalue      : Time before the next `SIGALRM`.
 /// - starttime        : Process start time since boot (clock ticks).
 /// - vsize            : Virtual memory size (bytes).
 /// - rss              : Resident Set Size (pages in real memory).
 /// - rsslim           : Soft memory limit (bytes).
 /// - startcode        : Start address of executable code.
 /// - endcode          : End address of executable code.
-/// - startstack       : Bottom address of process stack.
+/// - startstack       : Initial top address of process stack.
 /// - kstkesp          : Current stack pointer (ESP).
 /// - kstkeip          : Current instruction pointer (EIP).
 /// - signal           : Bitmap of pending signals.
@@ -159,7 +166,9 @@ impl ProcFileOps for StatFileOps {
             (0, -1)
         };
 
+        // Placeholder: kernel flags are not exposed yet.
         let flags = 0;
+        // Placeholder: page-fault counters are not tracked per process/thread yet.
         let min_flt = 0;
         let cmin_flt = 0;
         let maj_flt = 0;
@@ -176,28 +185,107 @@ impl ProcFileOps for StatFileOps {
             )
         };
 
-        let cutime = 0;
-        let cstime = 0;
-        let priority = 0;
+        let (cutime, cstime) = {
+            let (user_time, kernel_time) = process.reaped_children_stats().lock().get();
+            (
+                duration_to_jiffies(user_time),
+                duration_to_jiffies(kernel_time),
+            )
+        };
+
+        let (priority, rt_priority, policy) = sched_values(&thread);
         let nice = process.nice().load(Ordering::Relaxed).value().get();
         let num_threads = process.tasks().lock().as_slice().len();
-        let itrealvalue = 0;
-        let starttime = 0;
+        let itrealvalue =
+            duration_to_jiffies(process.timer_manager().alarm_timer().lock().remain());
+        let starttime = process.start_time().as_u64();
 
-        let (vsize, rss) = if let Some(vmar_ref) = process.lock_vmar().as_ref() {
+        let (
+            vsize,
+            rss,
+            startcode,
+            endcode,
+            startstack,
+            start_data,
+            end_data,
+            start_brk,
+            arg_start,
+            arg_end,
+            env_start,
+            env_end,
+        ) = if let Some(vmar_ref) = process.lock_vmar().as_ref() {
             let vsize = vmar_ref.get_mappings_total_size();
             let anon = vmar_ref.get_rss_counter(RssType::Anon);
             let file = vmar_ref.get_rss_counter(RssType::File);
             let rss = anon + file;
-            (vsize, rss)
+            let process_vm = vmar_ref.process_vm();
+            let code_range = process_vm.code_range();
+            let data_range = process_vm.data_range();
+            let init_stack = process_vm.init_stack();
+            let argv_range = init_stack.argv_range();
+            let envp_range = init_stack.envp_range();
+            let start_brk = process_vm.heap().lock().heap_low();
+
+            (
+                vsize,
+                rss,
+                code_range.start,
+                code_range.end,
+                init_stack.user_stack_top(),
+                data_range.start,
+                data_range.end,
+                start_brk,
+                argv_range.start,
+                argv_range.end,
+                envp_range.start,
+                envp_range.end,
+            )
         } else {
-            (0, 0)
+            (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
+        };
+
+        let rsslim = process
+            .resource_limits()
+            .get_rlimit(ResourceType::RLIMIT_RSS)
+            .get_cur();
+
+        // Placeholder: saved SP/IP and wait-channel addresses are not exposed by `Thread`.
+        let kstkesp = 0;
+        let kstkeip = 0;
+        let wchan = 0;
+
+        let signal = u64::from(posix_thread.pending_signals());
+        let blocked = u64::from(posix_thread.sig_mask());
+        let (sigignore, sigcatch) = signal_disposition_masks(&process);
+
+        // Placeholder: swap accounting is not implemented yet.
+        let nswap = 0;
+        let cnswap = 0;
+
+        let exit_signal = process
+            .exit_signal()
+            .map(|sig_num| sig_num.as_u8() as i32)
+            .unwrap_or(0);
+        let processor = thread
+            .sched_attr()
+            .last_cpu()
+            .map(|cpu| u32::from(cpu) as usize)
+            .unwrap_or(0);
+
+        // Placeholder: block I/O delay and guest-time accounting are not implemented yet.
+        let delayacct_blkio_ticks = 0;
+        let guest_time = 0;
+        let cguest_time = 0;
+
+        let exit_code = match self.mode {
+            StatMode::Process => process.status().exit_code(),
+            StatMode::Thread => posix_thread.exit_code(),
         };
 
         let mut printer = VmPrinter::new_skip(writer, offset);
         writeln!(
             printer,
-            "{} ({}) {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {}",
+            "{} ({}) {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {} {}",
             pid,
             comm,
             state,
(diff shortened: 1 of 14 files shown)


kernel/src/syscall/sched_getattr.rs
```diff
@@ -103,12 +94,15 @@ impl TryFrom<LinuxSchedAttr> for SchedPolicy {
     type Error = Error;
 
     fn try_from(value: LinuxSchedAttr) -> Result<Self> {
-        Ok(match value.sched_policy {
-            SCHED_FIFO | SCHED_RR => SchedPolicy::RealTime {
+        let linux_policy = LinuxSchedPolicy::try_from(value.sched_policy)
+            .map_err(|_| Error::with_message(Errno::EINVAL, "invalid scheduling policy"))?;
```
lrh2000: I think we should prefer `try_from()?` if possible. For simplicity.

All `LinuxSchedPolicy::try_from` in this PR may need to be converted to `?`.

kernel/src/sched/sched_class/policy.rs
```diff
@@ -32,16 +32,28 @@ impl Default for SchedPolicy {
...
     RoundRobin = 2,
+    /// Not supported.
+    Batch = 3,
+    /// Reserved but not implemented yet on Linux.
+    Iso = 4,
     Idle = 5,
+    /// Not supported.
+    Deadline = 6,
+    /// Not supported.
+    Ext = 7,
```
lrh2000: This is a bit odd as doc comments. How about:
```rust
    Ext = 7, // Not supported.
```

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -333,13 +343,40 @@ fn map_segment_vmos(
...
+        start..(start + range.len())
+    });
+
+    let data_segment_size = data_range.len();
 
     Ok(ElfMappedInfo {
         full_range: relocated_range,
+        code_range,
+        data_range,
         data_segment_size,
```
lrh2000: Then `data_segment_size` is redundant?

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -333,13 +343,40 @@ fn map_segment_vmos(
...
+            let end = start
+                .checked_add(range.len())
+                .expect("relocated segment range should not overflow");
+            start..end
+        })
+        .reduce(|acc_range, range| acc_range.start.min(range.start)..acc_range.end.max(range.end))
+        .unwrap_or(0..0);
+
+    // The data range is the relocated memory range of the last loadable segment.
     // According to Linux behavior, the data segment only includes the last loadable segment.
```
lrh2000: This second sentence is redundant. The same fact is repeated twice.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -333,13 +343,40 @@ fn map_segment_vmos(
...
+        .filter(|phdr| phdr.vm_perms().contains(VmPerms::EXEC))
+        .map(|phdr| {
+            let range = phdr.virt_range();
+            let start = relocated_range
+                .relocated_addr_of(range.start)
+                .expect("`calc_total_vaddr_bounds()` should cover all segments");
+            let end = start
+                .checked_add(range.len())
+                .expect("relocated segment range should not overflow");
+            start..end
```
lrh2000: Should use `start..(start + range.len())` instead.

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -221,9 +314,91 @@ impl ProcFileOps for StatFileOps {
...
+            env_start,
+            env_end,
+            exit_code
         )?;
 
         Ok(printer.bytes_written())
     }
 }
+
+const CLOCK_TICKS_PER_SEC: u64 = 100;
```
lrh2000: What does this code want to do here? I think the previous code without `jiffies_to_clock_t` is correct. Could you explain your reasoning for adding `jiffies_to_clock_t`?

The reference link shows that Linux uses `HZ` as the clock units. Do we need to convert anything here? `HZ` is the configurable time interrupt frequency, which can be 100, 300, or other values. It is our `TIMER_FREQ`.

kernel/src/process/process/mod.rs
```diff
@@ -145,6 +147,8 @@ pub struct Process {
 
     /// A profiling clock measures the user CPU time and kernel CPU time of the current process.
     prof_clock: Arc<ProfClock>,
+    /// Process start time since boot.
+    start_time: Jiffies,
```
lrh2000: Maybe:
```patch
diff --git a/kernel/src/process/process/mod.rs b/kernel/src/process/process/mod.rs
index 95a5ffe7e..34486dcdc 100644
--- a/kernel/src/process/process/mod.rs
+++ b/kernel/src/process/process/mod.rs
@@ -124,7 +124,6 @@ pub struct Process {
     /// Instead of letting the init process to reap all orphan zombie processes,
     /// a subreaper can reap orphan zombie processes among its descendants.
     is_child_subreaper: AtomicBool,
-
     /// Whether the process has a subreaper that will reap it when the
     /// process becomes orphaned.
     ///
@@ -139,15 +138,16 @@ pub struct Process {
     sig_queues: SigQueues,
     /// The signal that the process should receive when parent process exits.
     parent_death_signal: AtomicSigNum,
-
     /// The signal that should be sent to the parent when this process exits.
     exit_signal: AtomicSigNum,
 
+    // Time
     /// A profiling clock measures the user CPU time and kernel CPU time of the current process.
     prof_clock: Arc<ProfClock>,
-
     /// A manager that manages timer resources and utilities of the process.
     timer_manager: PosixTimerManager,
+    /// Process start time since boot.
+    start_time: Jiffies,
 
     // Namespaces
     /// The user namespace
```

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -171,33 +178,119 @@ impl ProcFileOps for StatFileOps {
...
+                duration_to_clock_t(kernel_time),
             )
         };
 
-        let cutime = 0;
-        let cstime = 0;
+        // Placeholder: scheduler priority is not exposed yet.
         let priority = 0;
+        // FIXME: `process.nice()` is not synchronized with the actual thread scheduling
+        // policy. The scheduler and procfs should eventually use the same source of truth.
```
lrh2000: Is this the right place for the FIXME?

lrh2000 (commented):
Thanks for enhancing `/proc/[pid]/stat` output.

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
StevenJiang1110: Sorry about that. I defined a new constant here because I could not find the equivalent of Linux's `HZ` in our repository. If `TIMER_FREQ` already serves this purpose, then there is indeed no need to introduce a new constant.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -333,14 +341,34 @@ fn map_segment_vmos(
...
+            let range = phdr.virt_range();
+            let start = relocated_range
+                .relocated_addr_of(range.start)
+                .expect("`calc_total_vaddr_bounds()` should cover all segments");
+            start..(start + range.len())
+        })
+        .reduce(|acc_range, range| acc_range.start.min(range.start)..acc_range.end.max(range.end))
+        .unwrap_or(0..0);
+
     // According to Linux behavior, the data segment only includes the last loadable segment.
```
lrh2000:
```suggestion
    // According to Linux behavior, the data range only includes the last loadable segment.
```

Update the comment since you renamed the variable name.

kernel/src/process/process_vm/mod.rs
```diff
@@ -137,6 +156,16 @@ impl ProcessVm {
             .map_and_init_heap(vmar, data_segment_size, heap_base)
     }
 
+    /// Updates the executable code range.
+    pub(super) fn set_code_range(&self, range: Range<Vaddr>) {
+        *self.code_range.lock() = range;
+    }
+
+    /// Updates the executable data range.
```
lrh2000: "executable data range" is really odd.

Can we remove the "executable" word, or use something like (assuming that I understand the word "executable" correctly):
```rust
    /// Updates the data range from the executable file.

    /// Returns the data range of the executable file.
```

There are (at least) three occurrences of "executable data range". Remember to update all of them. Maybe we should update "executable code range" for consistency as well.

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -176,28 +183,112 @@ impl ProcFileOps for StatFileOps {
...
+                duration_to_jiffies(user_time),
+                duration_to_jiffies(kernel_time),
+            )
+        };
+
+        // Placeholder: scheduler priority is not exposed yet.
         let priority = 0;
         let nice = process.nice().load(Ordering::Relaxed).value().get();
         let num_threads = process.tasks().lock().as_slice().len();
+        // Placeholder: interval timers are not included in `/proc/<pid>/stat` yet.
```
lrh2000: We can just implement this via `process.timer_manager().alarm_timer()`?

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -176,28 +183,112 @@ impl ProcFileOps for StatFileOps {
...
-            (0, 0)
+            (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
+        };
+
+        let rsslim = process
+            .resource_limits()
+            .get_rlimit(ResourceType::RLIMIT_RSS)
+            .get_cur();
+
+        // Placeholder: saved kernel SP/IP and wait-channel addresses are not exposed by `Thread`.
```
lrh2000:
```suggestion
        // Placeholder: saved SP/IP and wait-channel addresses are not exposed by `Thread`.
```

SP/IP are the user-space ones instead of kernel ones. In fact, zeros are correct values for most scenarios. https://github.com/torvalds/linux/blob/ba3e43a9e601636f5edb54e259a74f96ca3b8fd8/fs/proc/array.c#L492-L507

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -176,28 +183,112 @@ impl ProcFileOps for StatFileOps {
...
+            .exit_signal()
+            .map(|sig_num| sig_num.as_u8() as i32)
+            .unwrap_or(0);
+        let processor = thread
+            .sched_attr()
+            .last_cpu()
+            .map(|cpu| u32::from(cpu) as usize)
+            .unwrap_or(0);
+
+        // Placeholder: real-time priority and scheduler policy are not implemented yet.
```
lrh2000:
```rust
        // Placeholder: scheduler priority is not exposed yet.
```
```rust
        // Placeholder: real-time priority and scheduler policy are not implemented yet.
```

Why are they "not exposed" or "not implemented"? It seems that `thread.sched_attr()` is available to call. Am I missing anything?

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -221,9 +312,73 @@ impl ProcFileOps for StatFileOps {
...
     }
 }
+
+/// Converts a duration into kernel clock ticks.
+fn duration_to_jiffies(duration: Duration) -> u64 {
+    const NSEC_PER_SEC: u128 = 1_000_000_000;
+
+    let sec_jiffies = duration.as_secs().saturating_mul(TIMER_FREQ);
+    let subsec_jiffies =
+        u128::from(duration.subsec_nanos()) * u128::from(TIMER_FREQ) / NSEC_PER_SEC;
```
lrh2000: We can use `const { assert!(NSEC_PER_SEC %  TIMER_FREQ == 0) }`. Then use `* (NSEC_PER_SEC / TIMER_FREQ)`? `u64` should be enough?

kernel/src/syscall/sched_getattr.rs
```diff
@@ -183,7 +181,8 @@ pub(super) fn write_linux_sched_attr_to_user(
 
     attr.size = (size_of::<LinuxSchedAttr>() as u32).min(user_size);
 
-    let range = &SCHED_PRIORITY_RANGE[attr.sched_policy as usize];
+    let linux_policy = LinuxSchedPolicy::try_from(attr.sched_policy)?;
```
lrh2000: This shouldn't fail. The code below is:
```
    let attr: LinuxSchedAttr = policy
        .try_into()
        .expect("all user-visible scheduling attributes should be valid");
```

So we can align with it:
```suggestion
    let linux_policy = LinuxSchedPolicy::try_from(attr.sched_policy)
        .expect("all user-visible scheduling policies should be vaild");
```

lrh2000 (commented):
Thanks for the revision!

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -221,9 +309,109 @@ impl ProcFileOps for StatFileOps {
...
+        SchedPolicy::Stop => (
+            -i32::from(MAX_RT_PRIORITY) - 1,
+            MAX_RT_PRIORITY,
+            linux_policy as i32,
+        ),
+        SchedPolicy::RealTime { rt_prio, rt_policy } => {
+            let linux_policy = match rt_policy {
+                RealTimePolicy::Fifo => LinuxSchedPolicy::Fifo,
+                RealTimePolicy::RoundRobin { .. } => LinuxSchedPolicy::RoundRobin,
+            };
```
lrh2000: I don't understand how this code snippet was written. It seems to be the same logic that is already present and executed during `LinuxSchedPolicy::from`. What is the purpose of the repetition, then?
```rust
    let linux_policy = LinuxSchedPolicy::from(policy);
```
```rust
            // In `LinuxSchedPolicy::from`:
            SchedPolicy::RealTime {
                rt_policy: RealTimePolicy::Fifo,
                ..
            } => LinuxSchedPolicy::Fifo,
            SchedPolicy::RealTime {
                rt_policy: RealTimePolicy::RoundRobin { .. },
                ..
            } => LinuxSchedPolicy::RoundRobin,
```

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -221,9 +309,109 @@ impl ProcFileOps for StatFileOps {
...
+    const NSEC_PER_SEC: u64 = 1_000_000_000;
+    const NSEC_PER_JIFFY: u64 = NSEC_PER_SEC / TIMER_FREQ;
+    const { assert!(NSEC_PER_SEC.is_multiple_of(TIMER_FREQ)) };
+
+    let sec_jiffies = duration.as_secs().saturating_mul(TIMER_FREQ);
+    let subsec_jiffies = u64::from(duration.subsec_nanos()) / NSEC_PER_JIFFY;
+    sec_jiffies.saturating_add(subsec_jiffies)
+}
+
+fn sched_values(thread: &Thread) -> (i32, u8, i32) {
```
lrh2000, line 359:
```suggestion
/// Returns the `priority`, `rt_priority`, and `policy` values for `/proc/<pid>/stat`.
fn sched_values(thread: &Thread) -> (i32, u8, i32) {
```

Otherwise, it is hard to learn what `(i32, u8, i32)` represents...

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -221,9 +309,109 @@ impl ProcFileOps for StatFileOps {
...
+            -i32::from(MAX_RT_PRIORITY) - 1,
+            MAX_RT_PRIORITY,
+            linux_policy as i32,
+        ),
+        SchedPolicy::RealTime { rt_prio, rt_policy } => {
+            let linux_policy = match rt_policy {
+                RealTimePolicy::Fifo => LinuxSchedPolicy::Fifo,
+                RealTimePolicy::RoundRobin { .. } => LinuxSchedPolicy::RoundRobin,
+            };
+            let rt_priority = RT_PRIORITY_LIMIT - rt_prio.get();
```
lrh2000, line 377: Please also double-confirm that this logic is correct?

From what I can see,
 1. in the `Stop` branch, you're trying to return `-i32::from(MAX_RT_PRIORITY) - 1`;
 2. however, in the `RealTime` branch, if `rt_priority` is equal to `RealTimePriority::MAX`, `rt_priority` will be assigned `1`, and the return value will be `-2`.

But I think the two values need to match. Otherwise, it is somewhat odd.

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
lrh2000: Also note that if this comment is valid, maybe we can just simplify the code to:
```rust
    let policy = thread.sched_attr().policy();

    let (priority, rt_priority) = match /* .. */;
    let linux_policy = LinuxSchedPolicy::from(policy);

    (priority, rt_priority, linux_policy as i32)
```

So we don't need to repeat `linux_policy as i32` in three branches...

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
StevenJiang1110: I checked the Linux logic. First, `rt_priority` is computed as `-priority - 1`. For the `Stop` case, the `priority` is at its maximum value of 99, so the return value is `(99, -100)`, which is correct.

Regarding the `Realtime` branch, the main issue is that the `rt_priority` we store internally is not the same value as what the user passes in — it is `100 - user_value`. That is, if the user passes `1` via `sched_setattr`, we store `99` internally, but `/proc/pid/stat` should still return 1 to the user.

https://github.com/asterinas/asterinas/blob/b27cca2263b5512a83be3d7781da369ee19b53fe/kernel/src/syscall/sched_getattr.rs#L105-L109
https://github.com/asterinas/asterinas/blob/b27cca2263b5512a83be3d7781da369ee19b53fe/kernel/src/syscall/sched_get_priority_max.rs#L12-L18

I am not very familiar with the scheduler code, so I am not sure why this conversion is needed. However, if the internally stored `rt_priority` matched the user-provided value directly, it would clearly avoid a lot of confusion.

So I think the current calculation is correct. I have added a comment in the code for now to explain this subtlety.

kernel/src/sched/sched_class/policy.rs
```diff
@@ -54,8 +62,7 @@ impl From<SchedPolicy> for LinuxSchedPolicy {
                 rt_policy: RealTimePolicy::RoundRobin { .. },
                 ..
             } => LinuxSchedPolicy::RoundRobin,
-            SchedPolicy::Fair(_) => LinuxSchedPolicy::Normal,
-            SchedPolicy::Idle => LinuxSchedPolicy::Idle,
+            SchedPolicy::Fair(_) | SchedPolicy::Idle => LinuxSchedPolicy::Normal,
         }
     }
 }
```
lrh2000, line 68: I was wondering why you changed the code introduced in the first commit in your second commit. I think we should reorder the commits to avoid problems like this (or just move the related changes back to the first commit)? Though it's not a big problem...

All changes in `kernel/src/sched/sched_class/policy.rs` in the second commit seems to modify the code introduced in the first commit.

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
```diff
@@ -221,9 +309,104 @@ impl ProcFileOps for StatFileOps {
...
+            // `syscall` module.
+            let rt_priority = RT_PRIORITY_LIMIT - rt_prio.get();
+            (-i32::from(rt_priority) - 1, rt_priority)
+        }
+        SchedPolicy::Fair(nice) => (NICE_TO_PRIORITY_OFFSET + nice.value().get() as i32, 0),
+        SchedPolicy::Idle => (NICE_TO_PRIORITY_OFFSET, 0),
+    };
+
+    let linux_policy = LinuxSchedPolicy::from(policy);
+    (priority, rt_priority, linux_policy as i32)
```
lrh2000, line 385:
```suggestion
    };
    let linux_policy = LinuxSchedPolicy::from(policy);

    (priority, rt_priority, linux_policy as i32)
```

```rust
// Block 1: Get `policy`.

// Block 2: Compute `priority`, `rt_policy`, and `linux_policy` from `policy`.

// Block 3: Return `priority`, `rt_policy`, and `linux_policy`.
```

But it's not a big problem either way, so it can stay as it is if you have a different opinion.

kernel/src/fs/fs_impls/procfs/pid/task/stat.rs
lrh2000: Thanks for your explanation. It makes sense to me!

lrh2000 (commented):
Two nits.. Otherwise LGTM.

lrh2000 (approved):

PR #3291, opened 2026-05-28
Fix virtio-blk DMA sync for writes

Hsy-Intel:
In the virtio-blk write path, the code collected each BIO segment's `inner_dma_slice()` and directly passed those slices into the virtqueue descriptor chain.

That missed the contract of `DmaStream`: when the CPU updates a DMA buffer, `sync_to_device()` must be called before the device is notified to consume it.


Hsy-Intel:
@boterinas rerun

Hsy-Intel:
@boterinas rerun

tatetian (approved):

PR #3294, opened 2026-05-28
Some small improvements for the terminal display

wyt8:
These changes are all visible to verify.

1. The first commit makes the color display normal
2. The second commit makes the software know the terminal size, which can improve the result of display

Before:

image

After:

image

### DIFF
--- a/kernel/comps/framebuffer/src/ansi_escape.rs
+++ b/kernel/comps/framebuffer/src/ansi_escape.rs
@@ -394,7 +394,7 @@ impl EscapeFsm {
                 100..=107 => op.set_bg_color(COLORS[op_code as usize - 100 + 8]),
 
                 // Invalid or unsupported
-                _ => return,
+                _ => {}
             }
         }
     }
--- a/kernel/comps/framebuffer/src/ansi_escape.rs
+++ b/kernel/comps/framebuffer/src/ansi_escape.rs
@@ -20,6 +20,17 @@ pub(super) enum EraseInDisplay {
     EntireScreenAndScrollback,
 }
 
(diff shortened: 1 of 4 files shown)


lrh2000:
> After:
   
It does not seem to correctly detect the correct window size? Otherwise, why do the `ls` results wrap around the line? Am I missing something?

lrh2000:
Additionally, the second commit is fundamentally wrong because you cannot and should not call `push_input` after locking console. This will cause deadlocks.

The existing lock order is:
 1. `ldisc`
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/mod.rs#L126-L128
 2. `console`
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/vt/driver.rs#L123-L124

However, calling `push_input` after locking `console` will lock `ldisc`, which is a lock order violation that can cause a deadlock.

So, we must re-study how to lock `ldisc` in `Tty::write`. Should we hold the lock during `TtyDriver::push_output` or not?
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/mod.rs#L214-L220

```rust
let ldisc = self.ldisc.lock();
self.driver.push_output(&buf[..write_len], &mut ldisc);
// So that `push_output` can use `ldisc.push_char`... without lock order violations?
```

This relates to how we implement https://github.com/asterinas/asterinas/pull/3341 as well.

PR #3297, opened 2026-05-29
Introduce `InodeRollbackGuard` to handle write errors more elegantly for ext2

Fischer0522:
## Summary

This PR introduces `InodeRollbackGuard` to handle partial failures on the write path (`write_at`, `write_direct_at`, `fallocate`, and `slow_symlink`) for ext2.

`InodeRollbackGuard` acquires `&mut InodeInner`, allowing the write path to be structured with a guard pattern that captures the current state before fallible operations and restores it on error, with `commit` finalizing success.

With `InodeRollbackGuard`, the write path can be structured as follows:
```rust
      let mut guard = InodeRollbackGuard::new(self, RollbackKind::PageCacheAndBlocks { end: target_len });
      let inner = guard.inner_mut();
      inner.prepare_write(fs.as_ref(), 0, target_len)?;
      // perform some write operations here
      guard.commit();
```

## What changed

- Added `InodeRollbackGuard` and `RollbackKind` in `inode/file.rs`:
  - `PageCacheAndBlocks { end }` — restores both page-cache capacity and block mappings to the captured `old_size` (counterpart of `prepare_write`)
  - `BlocksOnly` — restores only block mappings (counterpart of `fallocate`, which never grows the page cache)
- Converted the following call sites to use the guard, propagating errors via `?`:
  - `InodeInner::write_at`, `write_direct_at`, `Inode::fallocate` — `file.rs`
  - `InodeInner::make_empty` — `dir/mod.rs`
  - `InodeInner::write_link` slow path — `symlink.rs`

## TODO

This PR is currently marked as a draft due to the following open questions:

- **`grow_dir_block`**: This method allocates a new block when the current directory block is full. I did not include it in `InodeRollbackGuard` because an empty block is harmless, and Linux does not roll it back either.

- **`rename`**: This operation involves multiple inodes, making rollback across them complicated. Partial failures are also harmless — the operation adds a new directory entry first, then deletes the old one. If a failure occurs, only a deleted entry remains, which can be fixed by `fsck`.

### DIFF
--- a/kernel/src/fs/fs_impls/ext2/inode/dir/mod.rs
+++ b/kernel/src/fs/fs_impls/ext2/inode/dir/mod.rs
@@ -10,7 +10,11 @@
 mod dir_entry;
 
 use self::dir_entry::{DOT_BYTE, DOT_DOT_BYTE, DirBlockView, DirEntryFileType, DirEntryHeader};
-use super::{super::Ext2, FileFlags, FilePerm, Inode, InodeInner, MAX_LINK_COUNT};
+use super::{
+    super::Ext2,
+    FileFlags, FilePerm, Inode, InodeInner, MAX_LINK_COUNT,
+    file::{InodeRollbackGuard, RollbackKind},
+};
 use crate::fs::ext2::{prelude::*, utils};
 
 /// Information about a candidate directory entry slot.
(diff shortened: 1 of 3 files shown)


kernel/src/fs/fs_impls/ext2/inode/dir/mod.rs
```diff
@@ -410,38 +414,35 @@ impl Inode {
...
-        // Allocate one block for the directory.
-        self.prepare_write(fs, 0, BLOCK_SIZE)?;
-
-        let page_cache = self.page_cache();
+        // Allocate one block for the directory; the guard rolls back the
+        // page-cache expansion and block allocation if entry writes fail.
+        let mut guard =
+            InodeRollbackGuard::new(self, RollbackKind::PageCacheAndBlocks { end: BLOCK_SIZE });
+        let inner = guard.inner_mut();
+        inner.prepare_write(fs, 0, BLOCK_SIZE)?;
```
cchanging, line 422: `InodeRollbackGuard` should not use a constructor like 'new'. 

Once a guard is created, its semantics should imply that the transaction it guards has already occurred. For example, with a lock guard, when you acquire it, it means the lock has already been taken. However, when 'new' happens here, the actual allocation logic has not yet occurred. 

Therefore, what should happen here is that `prepare_write` generates a `PageCacheAndBlocks` guard, and `allocate_range_blocks` generates a `BlocksOnly` guard. Of course, the names of these two functions and some details can be adjusted to make them appear more symmetric.

kernel/src/fs/fs_impls/ext2/inode/file.rs
```diff
@@ -213,6 +203,100 @@ impl Inode {
...
+    PageCacheAndBlocks { end: usize },
+    /// Block mappings only, with the page cache left untouched.
+    ///
+    /// The counterpart of `fallocate`, which allocates blocks without
+    /// expanding the page cache.
+    BlocksOnly,
+}
+
+/// RAII guard that restores an inode to its pre-mutation state when dropped,
+/// unless `commit` is called first.
```
cchanging, line 221: Explain the semantics instead of the details. 

```rust
  /// Inode rollback guard for prepared write-side mutations.
  ///
  /// ...
```

kernel/src/fs/fs_impls/ext2/inode/symlink.rs
```diff
@@ -111,8 +113,14 @@ impl InodeInner {
...
             }
-            self.prepare_write(fs.as_ref(), 0, target_len)?;
-            self.page_cache().write_bytes(0, target.as_bytes())?;
+            // The guard rolls back the block allocation if writing the target
+            // bytes fails partway through.
+            let mut guard =
+                InodeRollbackGuard::new(self, RollbackKind::PageCacheAndBlocks { end: target_len });
+            let inner = guard.inner_mut();
+            inner.prepare_write(fs.as_ref(), 0, target_len)?;
+            inner.page_cache().write_bytes(0, target.as_bytes())?;
```
cchanging, line 122: We prefer to restrict these write operations so that they can only be initiated by a guard, thereby preventing misuse. So I'm thinking whether we can move the implementations of these methods to the guard abstraction, and then turn different `RollbackKinds` into different generic parameters, implementing the required mutable methods for guards with different generic parameters.

PR #3298, opened 2026-05-29
Support multi file systems for `xfstests`

Fischer0522:
## Summary

  Generalize the xfstests harness so it can run against multiple file systems instead of being hard-wired to ext2. `tmpfs` is added as the first non-ext2 example.

  File systems are now classified into two categories, selected via `XFSTESTS_FS_TYPE` (default `ext2`):

  - `XFSTESTS_BLOCK_BACKED_FS_LIST` — file systems that need backing block devices (e.g. `ext2`). For these, `qemu_args.sh` attaches the `xfstests_test.img` / `xfstests_scratch.img` drives and the runner mounts them.
  - `XFSTESTS_NON_BLOCK_BACKED_FS_LIST` — file systems that do not need block devices (e.g. `tmpfs`, and virtio-fs in the future). For these, the runner mounts the test/scratch directories directly.

The two lists combine into `XFSTESTS_SUPPORTED_FS_LIST`, which gates the selected type with a clear `Unsupported XFSTESTS_FS_TYPE=...` error.

  ### Per-filesystem test lists
  Each file system now owns a directory holding its own `block.list`, `short.list`, and `full.list`:
  - `test/initramfs/src/conformance/xfstests/ext2/{block,short,full}.list`
  - `test/initramfs/src/conformance/xfstests/tmpfs/{block,short,full}.list`
    (new)
   
`XFSTESTS_RUNLIST` defaults to `/opt/xfstests/$(XFSTESTS_FS_TYPE)/short.list`,
  and the Nix build copies the per-fs directories into the image.


Fischer0522:
Should I enable tmpfs in CI?

test/initramfs/src/conformance/xfstests/run_xfstests.sh
```diff
@@ -7,31 +7,43 @@ set -eu
...
+    mkdir -p "$XFSTESTS_DIR/test" "$XFSTESTS_DIR/scratch"
+    mount -t tmpfs tmpfs_test "$XFSTESTS_DIR/test"
+    mount -t tmpfs tmpfs_scratch "$XFSTESTS_DIR/scratch"
+else
+    # The xfstests test/scratch images are attached by qemu_args.sh as the 3rd and
+    # 4th virtio-blk devices, which Linux names /dev/vdc and /dev/vdd respectively
+    # (after /dev/vda=ext2 and /dev/vdb=exfat).
+    TEST_DEV=${XFSTESTS_TEST_DEV:-/dev/vdc}
+    SCRATCH_DEV=${XFSTESTS_SCRATCH_DEV:-/dev/vdd}
+    export TEST_DEV SCRATCH_DEV
```
cchanging: I think we can manage the `xfstests` directory like:

```
xfstests
     |- tmpfs
     |- ext2
     |- templete
     |       |- prepare.sh
     |       |- config
     |       |- block.list
     |       |- full.list
     |       |- short.list
     |- ...
     |- README.md

```

test/initramfs/src/conformance/xfstests/tmpfs/short.list
```diff
@@ -0,0 +1,73 @@
+
```
cchanging: Why have a blank line here?

test/initramfs/src/conformance/xfstests/tmpfs/full.list
```diff
@@ -0,0 +1,75 @@
+
```
cchanging: Why have a blank line here?

test/initramfs/src/conformance/xfstests/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+- `config/build_config.mk`: Makefile fragment for build-time settings.
+- `config/xfstests.config`: xfstests host options file.
+- `prepare.sh`: prepares the configured devices and mount points before
+  `./check` starts.
+- `run_list/block.list`: tests excluded from every run for this filesystem.
+- `run_list/short.list`: default quick run list.
+- `run_list/full.list`: broader run list for manual validation.
+
+Copy `template/` when adding a new filesystem.
```
cchanging: This README still needs further polishing. Since it is placed under the `conformance/xfstests` directory, it should provide a fairly comprehensive introduction to running xfstests on Asterinas.

At a minimum, it should answer these questions:

1. How does a user run xfstests?
2. How does a user add new xfstests and perform the related configuration?

At the moment, this README only covers the second part, and even in its current form it does not really read like a README. It feels more like a collection of notes or comments.

I think you could try reading this document from the perspective of a user with absolutely no prior knowledge of xfstests. You would probably notice some issues. For example:

- `config/xfstests.config`: xfstests host options file.  
  What is this host options file used for?

- `prepare.sh`: prepares the configured devices and mount points before `./check` starts.  
  What exactly is `./check` here?

- `run_list/short.list`: default quick run list.  
  Why do we need a quick run list? What kinds of tests should be considered “quick runs”? How should I use these lists locally?

You can take `test/nixos/README.md` as an example and further polish this README accordingly.

cchanging (approved):
Please resolve the conflicts, otherwise LGTM!

Fischer0522:
> Please resolve the conflicts, otherwise LGTM!

Done.

Makefile
```diff
@@ -53,10 +53,13 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
 EXTRA_BLOCKLISTS ?= ""
 # Parameters for xfstests.
-XFSTESTS_RUNLIST ?= /opt/xfstests/short.list
+XFSTESTS_RUNLIST ?= /opt/xfstests/$(XFSTESTS_FS_TYPE)/run_list/short.list
 XFSTESTS_DISK_SIZE ?= 12G
 XFSTESTS_TEST_DEV ?= /dev/vdd
 XFSTESTS_SCRATCH_DEV ?= /dev/vde
+XFSTESTS_FS_TYPE ?= ext2
+XFSTESTS_FS_DIR := test/initramfs/src/conformance/xfstests/$(XFSTESTS_FS_TYPE)
+XFSTESTS_BUILD_CONFIG := $(XFSTESTS_FS_DIR)/config/build_config.mk
```
tatetian: `XFSTESTS_FS_DIR` and `XFSTESTS_BUILD_CONFIG` can be derived from `XFSTESTS_FS_TYPE`. Do you have any practical reason to specify these two parameters in this `Makefile`?

Makefile
```diff
@@ -53,10 +53,13 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
 #   directory, and appends that file directly.
 EXTRA_BLOCKLISTS ?= ""
 # Parameters for xfstests.
-XFSTESTS_RUNLIST ?= /opt/xfstests/short.list
+XFSTESTS_RUNLIST ?= /opt/xfstests/$(XFSTESTS_FS_TYPE)/run_list/short.list
 XFSTESTS_DISK_SIZE ?= 12G
 XFSTESTS_TEST_DEV ?= /dev/vdd
 XFSTESTS_SCRATCH_DEV ?= /dev/vde
+XFSTESTS_FS_TYPE ?= ext2
```
tatetian: `XFSTESTS_RUNLIST` relies on the value of `XFSTESTS_FS_TYPE`. And I think `XFSTESTS_FS_TYPE` is the single most important parameter. Move `XFSTESTS_FS_TYPE` to the first position, before `XFSTESTS_RUNLIST`.

Makefile
```diff
@@ -130,6 +134,20 @@ export VSOCK=on
 CARGO_OSDK_BUILD_ARGS += --init-args="/test/run_vsock_test.sh"
 endif
 
+XFSTESTS_NEEDS_BLOCK_DEVICES := false
+XFSTESTS_MKFS :=
+ifeq ($(ENABLE_CONFORMANCE_TEST), true)
+ifeq ($(CONFORMANCE_TEST_SUITE), xfstests)
+ifeq ($(wildcard $(XFSTESTS_FS_DIR)),)
+$(error Unsupported XFSTESTS_FS_TYPE=$(XFSTESTS_FS_TYPE))
```
tatetian: The same logic is repeated in `test/initramfs/Makefile`. Find a way to avoid duplication. One idea is to move the check into `$(XFSTESTS_BUILD_CONFIG)` so that `include` it in the Makefile has no effect when xfstests is not enabled.

.github/actions/test/action.yml
```diff
@@ -56,6 +56,9 @@ inputs:
   xfstests_runlist:
     description: 'Run list file for xfstests'
     required: false
+  xfstests_fs_type:
```
tatetian: In another comment, I suggest placing `xfstests_fs_type` as the first of the xfstests-related parameters. Make sure that xfstest parameters in different files follow a consistent order.

.github/workflows/test_xfstests_full.yml
```diff
@@ -23,4 +23,5 @@ jobs:
           release: true
           mem: '12G'
           xfstests_disk_size: '12G'
-          xfstests_runlist: '/opt/xfstests/full.list'
+          xfstests_runlist: '/opt/xfstests/ext2/run_list/full.list'
```
tatetian: This PR requires the xfstest-related files for a given FS type to be organized in a specific structure (see the `template` directory), which includes a `run_list` sub-directory. So we don't need to provide the full path for the full list any more. This simplifies the interface for running xfstests.

test/initramfs/Makefile
```diff
@@ -34,6 +34,22 @@ SSD_IMAGE := $(BUILD_DIR)/nvme0n1.img
...
+XFSTESTS_BUILD_CONFIG := $(XFSTESTS_FS_DIR)/config/build_config.mk
+XFSTESTS_NEEDS_BLOCK_DEVICES := false
+XFSTESTS_MKFS :=
+ifeq ($(ENABLE_CONFORMANCE_TEST), true)
+ifeq ($(CONFORMANCE_TEST_SUITE), xfstests)
+ifeq ($(wildcard $(XFSTESTS_FS_DIR)),)
+$(error Unsupported XFSTESTS_FS_TYPE=$(XFSTESTS_FS_TYPE))
+endif
+ifeq ($(wildcard $(XFSTESTS_BUILD_CONFIG)),)
+$(error Missing xfstests build config: $(XFSTESTS_BUILD_CONFIG))
```
tatetian: This PR adds four defensive checks to this Makefile, which look quite annoying. What is the true value of these checks? Makefile is only intended to be used by developers, not end users. So it is less likely that they make mistakes.

In addition, there is a double standard problem: no other Makefile parameters in this Makefile, e.g., `TARGET_ARCH`, are guarded by such defensive-programming checks.

Last, we need distinguish the input parameters given by the Makefile caller from those that are derived from these input parameters. For example, `XFSTESTS_FS_TYPE` is definitely an input parameter. But why would a Makefile caller provide  `XFSTESTS_FS_DIR` and `XFSTESTS_BUILD_CONFIG`? They can be determined by `XFSTESTS_FS_TYPE`.

I suggest removing all four annoying, defensive-programming checks from this file.

test/initramfs/Makefile
```diff
@@ -34,6 +34,22 @@ SSD_IMAGE := $(BUILD_DIR)/nvme0n1.img
 XFSTESTS_TEST_IMAGE := $(BUILD_DIR)/xfstests_test.img
 XFSTESTS_SCRATCH_IMAGE := $(BUILD_DIR)/xfstests_scratch.img
 XFSTESTS_DISK_SIZE ?= 12G
+XFSTESTS_FS_TYPE ?= ext2
+XFSTESTS_FS_DIR := $(CUR_DIR)/src/conformance/xfstests/$(XFSTESTS_FS_TYPE)
+XFSTESTS_BUILD_CONFIG := $(XFSTESTS_FS_DIR)/config/build_config.mk
+XFSTESTS_NEEDS_BLOCK_DEVICES := false
+XFSTESTS_MKFS :=
```
tatetian: These two parameters are already specified in `src/conformance/xfstests/$(XFSTESTS_FS_TYPE)/config/build_config.mk`. Why would a Makefile caller want to override the values given in `build_config.mk`?

tatetian (commented):
Great job on extending the xfstest test suite. I like the template-based approach. With that said, the implementation can be simplified. See comments below.

test/initramfs/nix/conformance/xfstests.nix
```diff
@@ -32,10 +32,15 @@ in stdenvNoCC.mkDerivation {
   buildCommand = ''
     mkdir -p $out/xfstests
     cp -r ${pkgs.xfstests}/lib/xfstests/* $out/xfstests/
+    chmod -R u+w $out/xfstests
```
tatetian, line 36: Why making everything under `xfstests` executable? What if there are configuration files? You might want to explain this point. It is unclear from here alone why it is safe to do `chmod -R u+w $out/xfstests`.

tatetian (commented):
The interface looks much better! I got only one minor issue.

test/initramfs/nix/conformance/xfstests.nix
Fischer0522: The copied xfstests tree inherits read-only permissions from the Nix store. This is needed so `tmpfs/prepare.sh` can patch `common/config` at runtime. I’ve added a comment to clarify it.

tatetian: Oh. I see. It is about being writable, not executable.

tatetian (approved):

tatetian:
I got no problem with this PR! Thanks for your work @Fischer0522. I appreciate it. Let's wait for the CI results.

PR #3299, opened 2026-05-29
Fix ext2 block-mapping writeback: persist allocations made during page-cache writeback

Fischer0522:
## Problem

Normally all block allocations for a file happen in the foreground via `prepare_write` during `write_at`, which marks the inode descriptor dirty, so `sync` correctly persists the updated block mappings.
  
However, when a file is mapped via `mmap` and a write lands on a hole, the allocation is deferred until page-cache writeback and performed inside `BlockAsPageCacheBackend::submit_write_bio`. In this path `RawBlockPtrs` (`block_ptrs` / `sector_count`) is modified, but the inode descriptor is never marked dirty. Because `fsync` / `fdatasync` gated descriptor writeback solely on `InodeInner::is_dirty()`, they would skip persisting the descriptor: the freshly allocated data blocks reach the disk, but the inode never records the mappings pointing at them, so the data is effectively lost after a remount.

  ## Fix

Give the on-disk block-pointer state its own dirty tracking by wrapping `RawBlockPtrs` in `Dirty<RawBlockPtrs>`, so allocations made anywhere — including during page-cache writeback — are no longer missed at sync time. The flag is exposed through `BlockPtrTree` and `InodeBlockManager` as `is_dirty()` / `clear_dirty()`.
  
`write_back_inode_desc is now:
  
  - treats the inode as needing writeback if either the descriptor dirty flag or the block-pointer dirty flag is set;
  - copies `block_ptrs` / `sector_count` into the descriptor only when the block pointers are dirty;
  - skips the write entirely when nothing is dirty, returning whether a write actually occurred;
  - clears both dirty flags after a successful write.

### DIFF
--- a/kernel/src/fs/fs_impls/ext2/fs.rs
+++ b/kernel/src/fs/fs_impls/ext2/fs.rs
@@ -256,7 +256,7 @@ impl Ext2 {
     }
 
     /// Writes an inode descriptor to the group's `PageCache`.
-    pub(super) fn write_inode_desc(&self, ino: Ext2Ino, raw_inode: &RawInode) -> Result<()> {
+    pub(super) fn write_back_inode_desc(&self, ino: Ext2Ino, raw_inode: &RawInode) -> Result<()> {
         {
             let sb = self.super_block.read();
             // Apply ext2 inode-number validity rules before indexing groups.
@@ -269,7 +269,7 @@ impl Ext2 {
             .find_group(ino)
             .ok_or_else(|| Error::with_message(Errno::EIO, "block group index out of range"))?;
 
-        group.write_inode_desc(ino, raw_inode)
+        group.write_back_inode_desc(ino, raw_inode)
     }
 
     /// Allocates up to `count` contiguous blocks.
@@ -393,7 +393,7 @@ impl Ext2 {
             .find_group(ino)
             .ok_or_else(|| Error::with_message(Errno::EIO, "block group index out of range"))?;
 
-        if let Err(err) = self.write_inode_desc(ino, &raw_inode) {
+        if let Err(err) = self.write_back_inode_desc(ino, &raw_inode) {
             if let Ok(was_allocated) = block_group.free_inode(ino, inode_type)
                 && was_allocated
             {
@@ -911,7 +911,7 @@ mod test {
 
         // Free path now takes caller-provided inode type; no inode-table read is needed.
         let raw_dir = make_raw_inode(0o040755, 1, 0);
-        f.ext2.write_inode_desc(ino, &raw_dir).unwrap();
+        f.ext2.write_back_inode_desc(ino, &raw_dir).unwrap();
         f.ext2.free_inode(ino, InodeType::Dir).unwrap();
         assert_eq!(f.ext2.super_block().free_inodes_count(), before_sb_free);
         assert_eq!(f.ext2.block_group(0).free_inodes_count(), before_group_free);
@@ -966,7 +966,10 @@ mod test {
         // Already-free inode: should return Ok and keep counters unchanged.
         let target_ino = f_free.sb.first_ino();
         let raw_file = make_raw_inode(0o100644, 1, 0);
-        f_free.ext2.write_inode_desc(target_ino, &raw_file).unwrap();
+        f_free
+            .ext2
+            .write_back_inode_desc(target_ino, &raw_file)
+            .unwrap();
 
         let before_sb = f_free.ext2.super_block().free_inodes_count();
         let before_group = f_free.ext2.block_group(0).free_inodes_count();
--- a/kernel/src/fs/fs_impls/ext2/inode/block_manager/block_ptr_tree.rs
+++ b/kernel/src/fs/fs_impls/ext2/inode/block_manager/block_ptr_tree.rs
@@ -30,7 +30,7 @@ const MAX_BLOCK_POINTER_LEVELS: usize = 4;
 /// or to indirect metadata blocks at the level selected by the slot.
 #[derive(Debug)]
 pub(in crate::fs::fs_impls::ext2::inode) struct BlockPtrTree {
-    raw_block_ptrs: RawBlockPtrs,
+    raw_block_ptrs: Dirty<RawBlockPtrs>,
     indirect_blocks_manager: Mutex<IndirectBlockManager>,
 }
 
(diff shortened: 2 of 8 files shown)


kernel/src/fs/fs_impls/ext2/inode/sync.rs
```diff
@@ -99,23 +88,32 @@ impl Inode {
...
 
 impl InodeInner {
-    /// Serializes the descriptor to disk and clears the dirty flag.
-    pub(super) fn write_back_desc(&mut self, fs: &Ext2, ino: Ext2Ino) -> Result<()> {
-        let raw_block_ptrs = self.raw_block_ptrs();
-        self.desc.block_ptrs = raw_block_ptrs.block_ptrs;
-        self.desc.sector_count = raw_block_ptrs.sector_count;
+    /// Serializes the descriptor to disk if dirty. Returns whether a write occurred.
+    pub(super) fn write_back_inode_desc(&mut self, fs: &Ext2, ino: Ext2Ino) -> Result<bool> {
+        let block_ptrs_is_dirty = self.block_manager().is_ok_and(|bm| bm.is_dirty());
```
cchanging: This logic should be put in `InodeInner::is_dirty`.

kernel/src/fs/fs_impls/ext2/inode/sync.rs
```diff
@@ -99,23 +88,32 @@ impl Inode {
...
+            return Ok(false);
+        }
         let raw_inode = RawInode::from(&*self.desc);
-        fs.write_inode_desc(ino, &raw_inode)?;
+        fs.write_back_inode_desc(ino, &raw_inode)?;
         self.clear_dirty();
-        Ok(())
+        if block_ptrs_is_dirty {
+            self.block_manager()?.clear_dirty();
+        }
```
cchanging: This logic should be put in `InodeInner::clear_dirty`.

kernel/src/fs/fs_impls/ext2/inode/sync.rs
cchanging: It is not necessary to let `write_back_inode_desc` return `bool`, which will make the semantic weird.

```rust
let is_dirty = inner.is_dirty;
if is_dirty {
    inner.write_back_inode_desc(...)?;
}
```

Or I think `sync_all` doesn't need to return `is_dirty`. For the `Inode::sync_all` operation, it should always call `sync_inode_table`. After all, the current is_dirty judgment in the `Inode::sync_all` operation has a race condition. The `sync_inode_table` operation only flushes dirty pages; when clean, it should just incur an extra scanning cost.

cchanging (commented):
Thanks for fixing this sync bug!

kernel/src/fs/fs_impls/ext2/inode/block_manager/block_ptr_tree.rs
```diff
@@ -51,6 +51,16 @@ impl BlockPtrTree {
...
 
+    /// Returns whether the `RawBlockPtrs` is dirty.
+    pub(in crate::fs::fs_impls::ext2::inode) fn is_dirty(&self) -> bool {
+        self.raw_block_ptrs.is_dirty()
+    }
+
+    /// Clears the dirty flag for the raw on-disk block pointer state.
+    pub(in crate::fs::fs_impls::ext2::inode) fn clear_dirty(&mut self) {
+        self.raw_block_ptrs.clear_dirty();
+    }
```
cchanging, line 62: `pub(super)` is enough?

kernel/src/fs/fs_impls/ext2/inode/sync.rs
```diff
@@ -99,21 +81,26 @@ impl Inode {
...
 
         fs.free_inode(self.ino, self.type_)?;
         Ok(true)
     }
 }
 
 impl InodeInner {
-    /// Serializes the descriptor to disk and clears the dirty flag.
-    pub(super) fn write_back_desc(&mut self, fs: &Ext2, ino: Ext2Ino) -> Result<()> {
+    /// Serializes the descriptor to disk if dirty. Returns whether a write occurred.
```
cchanging: The document should also be adjusted.

kernel/src/fs/fs_impls/ext2/inode/sync.rs
```diff
@@ -11,62 +11,44 @@ use super::{super::Ext2, Inode, InodeInner, RawInode};
...
-    /// Flushes xattr, data pages, and indirect blocks, then writes the inode descriptor back to
-    /// the block group's inode table page cache.
+    /// Flushes xattr, dirty data pages, indirect metadata, and dirty inode metadata.
     ///
-    /// Returns `true` if the descriptor was dirty and written back. The caller is responsible for
-    /// flushing the inode table and the block device.
-    pub(in crate::fs::fs_impls::ext2) fn sync_all(&self) -> Result<bool> {
-        // `fsync` step 1: flush the xattr.
+    /// This stages inode-table updates in the block group's inode-table page cache.
+    /// The caller is responsible for flushing the inode table and block device.
```
cchanging:
```suggestion
    /// Flushes all dirty state owned by the inode.
    ///
    /// Writes xattrs, data pages, and indirect blocks back before staging dirty
    /// inode metadata in the block group's inode-table page cache.
```

kernel/src/fs/fs_impls/ext2/inode/sync.rs
```diff
@@ -11,62 +11,44 @@ use super::{super::Ext2, Inode, InodeInner, RawInode};
...
+        inner.write_back_inode_desc(&fs, self.ino)
     }
 
     /// Flushes dirty data pages, indirect blocks, and dirty inode metadata.
     ///
-    /// Returns `true` if the descriptor was dirty and written back. The caller is responsible for
-    /// flushing the inode table and the block device.
-    pub(in crate::fs::fs_impls::ext2) fn sync_data(&self) -> Result<bool> {
+    /// This stages inode-table updates in the block group's inode-table page cache.
+    /// The caller is responsible for flushing the inode table and block device.
```
cchanging:
```suggestion
    /// Flushes dirty file data and the metadata required to retrieve it.
    ///
    /// Writes data pages and indirect blocks back before staging dirty inode
    /// metadata in the block group's inode-table page cache. This does not flush
    /// xattrs.
```

kernel/src/fs/fs_impls/ext2/fs.rs
cchanging: The name of this method shoud keep consistent.

cchanging (approved):

PR #3300, opened 2026-05-31
Avoid immediate ephemeral port reuse

StevenJiang1110:
This PR addresses a port reuse issue encountered when running Ollama on Kata/CoCo.

Ollama follows a specific usage pattern: a socket is first bound to 127.0.0.1:0, which causes an ephemeral port to be assigned — say, 32768. This socket is then closed, and another socket attempts to bind to 127.0.0.1:32768, i.e., the port that was just released. However, in between these two operations, another socket binding to 127.0.0.1:0 may have already claimed that port, resulting in an "address already in use" error.

The root cause is our current ephemeral port allocation strategy, which always starts scanning from the beginning of the range (i.e., from port 32768) on every allocation. This means a just-released port can be immediately reassigned to another socket. In contrast, Linux starts scanning from the port following the last allocated one — so if 32768 was the last assigned port, the next allocation starts from 32769, ensuring that a recently released port is not immediately reused.

This PR aligns our port allocation strategy more closely with Linux's behavior to fix this issue.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/iface/common.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/common.rs
@@ -42,7 +42,7 @@ pub struct IfaceCommon<E: Ext> {
     flags: InterfaceFlags,
 
     interface: SpinLock<PollableIface<E>, BottomHalfDisabled>,
-    used_ports: SpinLock<BTreeMap<PortKey, PortState>, BottomHalfDisabled>,
+    used_ports: SpinLock<PortTable, BottomHalfDisabled>,
     sockets: SpinLock<SocketTable<E>, BottomHalfDisabled>,
     sched_poll: E::ScheduleNextPoll,
 }
@@ -95,7 +95,7 @@ impl<E: Ext> IfaceCommon<E> {
             type_,
             flags,
             interface: SpinLock::new(PollableIface::new(interface)),
-            used_ports: SpinLock::new(BTreeMap::new()),
+            used_ports: SpinLock::new(PortTable::new()),
             sockets: SpinLock::new(SocketTable::new()),
             sched_poll,
         }
@@ -181,7 +181,7 @@ impl<E: Ext> IfaceCommon<E> {
         protocol: PortProtocol,
     ) -> Result<BoundPort<E>, BindError> {
         let addr = config.addr();
-        let (port, can_reuse) = self.bind_port(config, protocol)?;
+        let (port, can_reuse) = self.used_ports.lock().bind(config, protocol)?;
         Ok(BoundPort {
             iface,
             addr,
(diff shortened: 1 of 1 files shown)


kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -69,6 +69,7 @@ pub struct IfaceCommon<E: Ext> {
     used_ports: SpinLock<BTreeMap<IpAddress, BTreeMap<u16, PortState>>, BottomHalfDisabled>,
     sockets: SpinLock<SocketTable<E>, BottomHalfDisabled>,
     sched_poll: E::ScheduleNextPoll,
+    next_ephemeral_port: AtomicU32,
```
lrh2000: I don't think it should be an atomic variable. However, I understand that this is a quick fix that results in minimal code changes.

Would it make more sense to add a new `PortTable` type and make `next_ephemeral_port` a field?
```rust
used_ports: SpinLock<PortTable<E>, BottomHalfDisabled>
```

What do you think?

lrh2000 (commented):
Thanks for revising the ephemeral port allocation logic.

kernel/libs/aster-bigtcp/src/iface/common.rs
StevenJiang1110: It sounds reasonable. I will refactor it as suggested.

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -191,102 +191,19 @@ impl<E: Ext> IfaceCommon<E> {
...
-            }
-        }
-
-        // FIXME: If `can_reuse` is `true`, we should also check all in-use ephemeral ports
-        // to see if any can be reused instead of directly returning `None`.
-
-        None
-    }
-
     fn bind_port(
```
lrh2000: Looks like this method should be removed now?

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -456,6 +367,146 @@ enum PortProtocol {
...
+    }
+
+    /// Allocates an ephemeral port.
+    ///
+    /// We follow the port range that many Linux kernels use by default, which is 32768-60999.
+    ///
+    /// Each allocation starts scanning from the port immediately
+    /// after the last successfully allocated ephemeral port.
+    /// This ensures that a recently released port is not immediately
+    /// reused by a new socket, which mirrors Linux's ephemeral port allocation strategy.
```
lrh2000: This is actually the most reasonable way to allocate ephemeral ports. The previous method is sub-optimal because:
 1. It is inefficient because we scan from the start each time.
 2. It is easy to cause port conflicts for in-flight packets (packets may be delayed, duplicated, or reordered in the network);
 3. It is also possible to cause port conflicts for applications, as reported by this PR. I'm surprised that this may happen.

So minimizing the Linux strategy is not the most important reason here....
```suggestion
    /// reused by a new socket, which avoids potential port conflicts.
```

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -456,6 +367,146 @@ enum PortProtocol {
     Udp,
 }
 
+struct PortTable {
```
lrh2000, line 379: Should this type and its methods be moved after `PortState`?

kernel/libs/aster-bigtcp/src/iface/common.rs
```diff
@@ -456,6 +367,146 @@ enum PortProtocol {
...
+                port_state.nreuse -= 1;
+            }
+            if port_state.nsocket == 0 {
+                occupied.remove();
+            }
+        }
+    }
+
+    fn set_can_reuse(&mut self, key: PortKey, can_reuse: bool) {
+        if let Some(port_state) = self.used_ports.get_mut(&key) {
```
lrh2000: I think we can use early returns for `release` and `set_can_reuse` now.

lrh2000 (commented):
Just some minor nits.

lrh2000 (approved):

PR #3301, opened 2026-05-31
Inspect output flags (`OPOST`/`ONLCR`/`OCRNL`) when echoing characters in TTY line discipline

Solicey:
## Problem

The TTY line discipline's `output_char` method, which handles character echo, was ignoring the termios output flags. By default, `OPOST | ONLCR` is set, meaning a `\n` should be output as `\r\n` (carriage return + line feed). However, the code unconditionally echoed bare `\n` for newlines. This might lead to terminal cursor moving down one line without resetting to the first column.

The input side was already correct — `ICRNL` properly converts `\r` (Enter key) to `\n` on input. The bug was only on the echo/output side.

## Changes

- **`kernel/src/device/tty/termio.rs`**: Added `output_flags()` accessor to `CTermios` to expose the `c_oflags` field.
- **`kernel/src/device/tty/line_discipline.rs`**:
  - `output_char` now checks `OPOST` and `ONLCR` flags when echoing `\n`, outputting `\r\n` when both are set (the default).
  - `output_char` now checks `OPOST` and `OCRNL` flags when echoing `\r`, previously it unconditionally output `\r\n`.
  - Removed the `// TODO: respect output flags` comment as this is now addressed for the echo path.

### DIFF
--- a/ostd/src/arch/x86/cpu/context/mod.rs
+++ b/ostd/src/arch/x86/cpu/context/mod.rs
@@ -371,7 +371,6 @@ impl UserContextApiInternal for UserContext {
             rsi: self.user_context.general.rsi,
             rdi: self.user_context.general.rdi,
             rbp: self.user_context.general.rbp,
-            rsp: self.user_context.general.rsp,
             r8: self.user_context.general.r8,
             r9: self.user_context.general.r9,
             r10: self.user_context.general.r10,
@@ -380,12 +379,13 @@ impl UserContextApiInternal for UserContext {
             r13: self.user_context.general.r13,
             r14: self.user_context.general.r14,
             r15: self.user_context.general.r15,
-            _pad: 0,
             trap_num: self.user_context.trap_num,
             error_code: self.user_context.error_code,
             rip: self.user_context.general.rip,
             cs: 0,
             rflags: self.user_context.general.rflags,
+            rsp: self.user_context.general.rsp,
+            ss: 0,
         }
     }
 }
--- a/ostd/src/arch/x86/trap/mod.rs
+++ b/ostd/src/arch/x86/trap/mod.rs
@@ -73,7 +73,6 @@ pub struct TrapFrame {
     pub rsi: usize,
     pub rdi: usize,
     pub rbp: usize,
-    pub rsp: usize,
     pub r8: usize,
     pub r9: usize,
     pub r10: usize,
@@ -82,7 +81,6 @@ pub struct TrapFrame {
     pub r13: usize,
     pub r14: usize,
     pub r15: usize,
-    pub _pad: usize,
 
     pub trap_num: usize,
     pub error_code: usize,
@@ -91,10 +89,11 @@ pub struct TrapFrame {
     pub rip: usize,
     pub cs: usize,
     pub rflags: usize,
+    pub rsp: usize,
+    pub ss: usize,
 }
 
-// The padding field in `TrapFrame` is necessary. Do not attempt to remove it
-// without considering this note. Otherwise, the code will be **unsound**.
+// Be careful: This assertion is a **soundness** requirement.
 //
 // According to the System V AMD64 ABI, the stack pointer should be aligned to
 // at least 16 bytes. The hardware will align the stack pointer to a 16-byte
--- a/ostd/src/arch/x86/trap/trap.S
+++ b/ostd/src/arch/x86/trap/trap.S
@@ -10,8 +10,8 @@
  * We make the following new changes:
  * * Add the `trap_handler_table`.
  * * Adjust some symbol names.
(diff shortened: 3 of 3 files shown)


kernel/src/device/tty/line_discipline.rs
```diff
@@ -157,11 +157,25 @@ impl LineDiscipline {
...
+            b'\n' => {
+                if opost && oflags.contains(COutputFlags::ONLCR) {
+                    echo_callback(b"\r\n");
+                } else {
+                    echo_callback(b"\n");
+                }
+            }
+            b'\r' => {
+                if opost && oflags.contains(COutputFlags::OCRNL) {
+                    echo_callback(b"\r\n");
```
StevenJiang1110: `OCRNL` will map `\r` to `\n`, not to `\r\n`.

https://elixir.bootlin.com/linux/v7.0/source/drivers/tty/n_tty.c#L427

kernel/src/device/tty/line_discipline.rs
```diff
@@ -157,11 +157,25 @@ impl LineDiscipline {
         Ok(())
     }
 
-    // TODO: respect output flags
```
StevenJiang1110, line 160: Not all output flags are handled yet. I suggest narrowing the comment to explicitly state which flags are currently supported (e.g., ONLCR, OCRNL) and which are not, and placing it inside the `output_char` method.

StevenJiang1110 (commented):
Thanks for your contribution!

Could you also please add some regression test for it? Perhaps under `test/initramfs/src/regression/device/pty`.

Solicey:
Thanks for the review. I’ve addressed the requested changes in the latest update:
- added a regression test covering relevant output-flags' behavior,
- fixed the `OCRNL` handling so the conversion logic matches the expected Linux semantics,
- updated the comments to make the behavior clearer.

kernel/src/device/tty/line_discipline.rs
```diff
@@ -157,11 +157,30 @@ impl LineDiscipline {
...
+    /// Echoes a character back to the output, respecting output processing flags.
+    ///
+    /// Currently supported output flags: `OPOST`, `ONLCR`, `OCRNL`.
+    ///
+    /// TODO: Implement remaining output flags: `OLCUC`, `ONOCR`, `ONLRET`, `OFILL`, `OFDEL`.
     fn output_char<F: FnMut(&[u8])>(&self, ch: u8, mut echo_callback: F) {
```
StevenJiang1110: This method might be better renamed to `echo_char`, as it only handles the echo path rather than general output processing.

test/initramfs/src/regression/device/pty/output_flags_echo.c
```diff
@@ -0,0 +1,89 @@
...
+#include "../../common/test.h"
+
+static int master;
+static int slave;
+
+// Verify that with OPOST|ONLCR (the default), echoing a newline produces \r\n.
+FN_TEST(onlcr_enabled)
+{
+	struct termios term;
+	CHECK(openpty(&master, &slave, NULL, NULL, NULL));
```
StevenJiang1110: Always prefer `TEST_SUCC`/`TEST_RES` in `FN_TEST` than `CHECK`, if possible. So we should use `TEST_SUCC` here.

test/initramfs/src/regression/device/pty/output_flags_echo.c
```diff
@@ -0,0 +1,89 @@
...
+
+	// Use default termios: OPOST|ONLCR is on, ECHO is on, ICANON is on.
+	CHECK(tcgetattr(slave, &term));
+	TEST_RES(term.c_oflag & OPOST, _ret);
+	TEST_RES(term.c_oflag & ONLCR, _ret);
```
StevenJiang1110: For the common usage pattern, the first param in `TEST_RES` is an operation, usually call library functions or syscalls, the second param is the condition. The above three lines can be simplified as one:

`TEST_RES(tcgetattr(slave, &term), (term.c_oflag & OPOST != 0) && (term.c_oflag & ONLCR != 0))`

StevenJiang1110 (commented):
After taking a closer look at the current implementation, I found that the main issue is that the output flags are only applied to the echo path, but not to the data read by the other end (i.e., when the slave writes and the master reads, the output flags should also take effect). This makes the value of this PR somewhat difficult to assess, because the primary goal of handling output flags should be to ensure that what the other end reads is correct. Moreover, with the current implementation, the echoed characters and the characters read by the other end may end up being inconsistent.

Therefore, I would prefer that if you are going to handle output flags, you address the issue of the other end's read path being controlled by the output flags first. It would be great if you could fix both problems together in this PR. If you run into any difficulties during implementation, feel free to discuss them in the comments below.

PR #3303, opened 2026-06-01
Implement `TCP_KEEPINTVL` and stub `TCP_KEEPCNT`

taosue:
This PR adds `TCP_KEEPINTVL` and `TCP_KEEPCNT` socket options, which are needed while adding support for OpenClaw. `TCP_KEEPINTVL` is stored per socket and applied to the underlying TCP socket when `SO_KEEPALIVE` is enabled, including updates made while keepalive is already active. `TCP_KEEPCNT` follows the existing `TCP_KEEPIDLE` behavior and remains a dummy implementation for now: the value is validated, stored, and returned through getsockopt, but it is not yet enforced by the TCP stack.

### DIFF
--- a/kernel/src/net/socket/ip/stream/mod.rs
+++ b/kernel/src/net/socket/ip/stream/mod.rs
@@ -4,6 +4,7 @@ use core::sync::atomic::{AtomicBool, Ordering};
 
 use aster_bigtcp::{
     socket::{NeedIfacePoll, RawTcpOption, RawTcpSetOption},
+    time::Duration,
     wire::IpEndpoint,
 };
 use connected::{ConnectedStream, close_and_linger};
@@ -12,7 +13,7 @@ use init::InitStream;
 use listen::ListenStream;
 use observer::StreamObserver;
 use options::{
-    Congestion, DeferAccept, Inq, KEEPALIVE_INTERVAL, KeepIdle, MaxSegment, NoDelay, SynCnt,
+    Congestion, DeferAccept, Inq, KeepCnt, KeepIdle, KeepIntvl, MaxSegment, NoDelay, SynCnt,
     UserTimeout, WindowClamp,
 };
 use ostd::sync::{PreemptDisabled, RwLockReadGuard, RwLockWriteGuard};
@@ -93,8 +94,10 @@ impl OptionSet {
     }
 
     fn raw(&self) -> RawTcpOption {
+        let keep_alive_interval = Duration::from_secs(self.tcp.keep_intvl() as u64);
+
         RawTcpOption {
-            keep_alive: self.socket.keep_alive().then_some(KEEPALIVE_INTERVAL),
+            keep_alive: self.socket.keep_alive().then_some(keep_alive_interval),
             is_nagle_enabled: !self.tcp.no_delay(),
         }
     }
@@ -112,12 +115,27 @@ impl StreamSocket {
         })
     }
 
-    fn new_accepted(connected_stream: ConnectedStream) -> Arc<Self> {
+    fn new_accepted(connected_stream: ConnectedStream, listener_options: &OptionSet) -> Arc<Self> {
         let options = connected_stream.raw_with(|raw_tcp_socket| {
             let mut options = OptionSet::new();
 
-            if raw_tcp_socket.keep_alive().is_some() {
+            // Inherit socket options from `raw_tcp_socket` first, then fall
+            // back to `listener_options` for options the raw socket cannot
+            // expose.
+            //
+            // The raw socket is created when a connection arrives, before
+            // `accept()` returns it. If the listener's options are changed
+            // after that but before `accept()` returns, using only
+            // `listener_options` would give the accepted socket "new" options
+            // while its raw socket still has the "old" ones.
+
+            if let Some(interval) = raw_tcp_socket.keep_alive() {
                 options.socket.set_keep_alive(true);
+                options.tcp.set_keep_intvl(interval.secs() as u32);
+            } else {
+                options
+                    .tcp
+                    .set_keep_intvl(listener_options.tcp.keep_intvl());
             }
 
             if !raw_tcp_socket.nagle_enabled() {
@@ -303,7 +321,8 @@ impl StreamSocket {
 
         let accepted = listen_stream.try_accept().map(|connected_stream| {
             let remote_endpoint = connected_stream.remote_endpoint();
-            let accepted_socket = Self::new_accepted(connected_stream);
+            let listener_options = self.options.read();
+            let accepted_socket = Self::new_accepted(connected_stream, &listener_options);
             (accepted_socket as _, remote_endpoint.into())
         });
         let iface_to_poll = listen_stream.iface().clone();
@@ -639,6 +658,14 @@ impl Socket for StreamSocket {
                 let keep_idle = options.tcp.keep_idle();
                 tcp_keep_idle.set(keep_idle);
             }
+            tcp_keep_intvl @ KeepIntvl => {
+                let keep_intvl = options.tcp.keep_intvl();
+                tcp_keep_intvl.set(keep_intvl);
+            }
+            tcp_keep_cnt @ KeepCnt => {
+                let keep_cnt = options.tcp.keep_cnt();
+                tcp_keep_cnt.set(keep_cnt);
+            }
             tcp_syn_cnt @ SynCnt => {
                 let syn_cnt = options.tcp.syn_cnt();
                 tcp_syn_cnt.set(syn_cnt);
@@ -678,7 +705,12 @@ impl Socket for StreamSocket {
         let mut options = self.options.write();
 
         // Deal with socket-level options
-        let need_iface_poll = match options.socket.set_option(option, state.as_ref()) {
+        let socket_option_result = {
+            let OptionSet { socket, tcp, .. } = &mut *options;
+            socket.set_option(option, &(state.as_ref(), &*tcp))
+        };
+
+        let need_iface_poll = match socket_option_result {
             Err(err) if err.error() == Errno::ENOPROTOOPT => {
                 // Deal with IP-level options
                 match options.ip.set_option(option, state.as_ref()) {
@@ -748,6 +780,35 @@ fn do_tcp_setsockopt(
             options.tcp.set_keep_idle(*keepidle);
 
             // TODO: Track when the socket becomes idle to actually support keep idle.
+        }
+        tcp_keep_intvl @ KeepIntvl => {
+            // Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h#L168>
+            const MIN_KEEP_INTVL: u32 = 1;
+            const MAX_KEEP_INTVL: u32 = 32767;
+
+            let keepintvl = tcp_keep_intvl.get().unwrap();
+            if *keepintvl < MIN_KEEP_INTVL || *keepintvl > MAX_KEEP_INTVL {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "the keepalive interval is out of bounds"
+                );
+            }
+            options.tcp.set_keep_intvl(*keepintvl);
+
+            return Ok(state.set_keep_alive(options.socket.keep_alive(), *keepintvl));
+        }
+        tcp_keep_cnt @ KeepCnt => {
+            // Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h#L169>
+            const MIN_KEEP_CNT: u8 = 1;
+            const MAX_KEEP_CNT: u8 = 127;
+
+            let keepcnt = tcp_keep_cnt.get().unwrap();
+            if *keepcnt < MIN_KEEP_CNT || *keepcnt > MAX_KEEP_CNT {
+                return_errno_with_message!(Errno::EINVAL, "the keepalive count is out of bounds");
+            }
+            options.tcp.set_keep_cnt(*keepcnt);
+
+            // TODO: Track the keepalive count in the underlying TCP socket.
         }
         tcp_syn_cnt @ SynCnt => {
             const MAX_TCP_SYN_CNT: u8 = 127;
@@ -821,15 +882,7 @@ impl State {
             State::Listen(listen_stream) => Some(listen_stream.iface()),
         }
     }
-}
 
-impl GetSocketLevelOption for State {
-    fn is_listening(&self) -> bool {
-        matches!(self, Self::Listen(_))
-    }
-}
-
-impl SetSocketLevelOption for State {
     fn set_reuse_addr(&self, reuse_addr: bool) {
         let bound_port = match self {
             State::Init(init_stream) => {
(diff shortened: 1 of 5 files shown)


test/initramfs/src/regression/network/sockoption.c
```diff
@@ -282,6 +282,57 @@ FN_TEST(keepidle)
...
+
+FN_TEST(keepcnt)
+{
+	int keepcnt;
+	socklen_t keepcnt_len = sizeof(keepcnt);
+
+	refresh_connection();
+	TEST_RES(getsockopt(sk_connected, IPPROTO_TCP, TCP_KEEPCNT, &keepcnt,
+			    &keepcnt_len),
+		 keepcnt == 9);
```
StevenJiang1110, line 341: Please follow the pattern of the existing tests in this file and add comments to describe each test segment. For example:

```C
// 1. Check default values
// 2. Set and get values
```

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -848,19 +914,6 @@ impl SetSocketLevelOption for State {
...
-            Some(KEEPALIVE_INTERVAL)
-        } else {
-            None
-        };
-
-        let set_keepalive = |raw_socket: &dyn RawTcpSetOption| raw_socket.set_keep_alive(interval);
-
-        self.set_raw_option(set_keepalive)
-            .unwrap_or(NeedIfacePoll::FALSE)
-    }
```
StevenJiang1110, line 863: I think directly removing the `set_keep_alive` implementation on `SetSocketLevelOption` trait may cause some issues. This method is currently used by `SocketOptionSet::set_option`:

```rust
          socket_keepalive @ KeepAlive => {
                let keep_alive = socket_keepalive.get().unwrap();
                self.set_keep_alive(*keep_alive);
                return Ok(socket.set_keep_alive(*keep_alive));
            }
```

Once removed, the `set_keep_alive` method on the `SetSocketLevelOption` trait becomes dead code, since no socket would implement it anymore. To go down this path, you would need to remove both the method from the `SetSocketLevelOption` trait and the corresponding handling in `SocketOptionSet::set_option`.

But after doing all that, you then add a special case in `StreamSocket::set_option`:

```rust
               if let Some(socket_keep_alive) = option.as_any().downcast_ref::<KeepAlive>() {
                    // Sync the raw TCP socket using the current `TCP_KEEPINTVL`.
                    let keep_alive = socket_keep_alive.get().unwrap();
                    state.set_keep_alive(*keep_alive, options.tcp.keep_intvl())
                }
```

This does not feel ideal, as it breaks the existing abstraction. The reason you are doing this is that `SetSocketLevelOption` is currently implemented on `State`, but `State` does not have access to the interval information, which is stored in `OptionSet`.

My suggestion is to implement `SetSocketLevelOption` not just on `State` alone, but on a tuple of `(&State, &OptionSet)` instead. This way, the original code structure can be preserved:

```rust
    impl SetSocketLevelOption for (&State, &OptionSet) {
         fn keep_alive(....) {}
    }
```

You can try whether it works.

test/initramfs/src/regression/network/sockoption.c
```diff
@@ -282,6 +282,57 @@ FN_TEST(keepidle)
...
+	int enabled = 1;
+	TEST_SUCC(setsockopt(sk_listen, IPPROTO_TCP, TCP_KEEPINTVL, &seconds,
+			     sizeof(seconds)));
+	TEST_SUCC(setsockopt(sk_listen, SOL_SOCKET, SO_KEEPALIVE, &enabled,
+			     sizeof(enabled)));
+	refresh_connection();
+	TEST_RES(getsockopt(sk_accepted, IPPROTO_TCP, TCP_KEEPINTVL, &keepintvl,
+			    &keepintvl_len),
+		 keepintvl == 30);
+}
```
StevenJiang1110, line 329: Please add a test for the following case: the listening socket sets a keep interval but disables keep-alive, and then verifies whether the accepted socket inherits the keep interval:

```C
    enabled = 0;
	seconds = 50;
	TEST_SUCC(setsockopt(sk_listen, IPPROTO_TCP, TCP_KEEPINTVL, &seconds,
			     sizeof(seconds)));
	TEST_SUCC(setsockopt(sk_listen, SOL_SOCKET, SO_KEEPALIVE, &enabled,
			     sizeof(enabled)));
	refresh_connection();
	TEST_RES(getsockopt(sk_accepted, IPPROTO_TCP, TCP_KEEPINTVL, &keepintvl,
			    &keepintvl_len),
		 keepintvl == 50);
```

My local testing shows that Linux does inherit this value, but the changes in this PR do not support this behavior.

You may want to fix this at the location marked by the following comment:

```rust
// TODO: Update other options for a newly-accepted socket
```

StevenJiang1110 (commented):
Thanks for your contribution!

kernel/libs/aster-bigtcp/src/socket/option.rs
```diff
@@ -36,3 +48,53 @@ impl RawTcpOption {
...
 }
+
+/// A raw TCP socket that preserves `TCP_KEEPINTVL` independently of keepalive state.
+///
+/// The underlying smoltcp socket stores TCP keepalive as `Option<Duration>`. When keepalive is
+/// disabled, that representation becomes `None` and cannot carry the configured interval. Linux
+/// still lets accepted sockets inherit `TCP_KEEPINTVL` from a listener even when `SO_KEEPALIVE` is
+/// disabled, so this wrapper keeps the interval separately until the accepted stream socket can
+/// initialize its TCP option set.
+pub(super) struct RawTcpSocketWithKeepAliveInterval {
```
StevenJiang1110: I think introducing this struct is unnecessary and feels quite awkward. I understand why you did it — when keep-alive is None on RawTcpSocket, there is no way to retrieve the interval from the listener. However, we do not have to get this information from RawTcpSocket at all; it can be retrieved from the upper-level StreamSocket instead.

Considering the following change:

```diff
iff --git a/kernel/src/net/socket/ip/stream/mod.rs b/kernel/src/net/socket/ip/stream/mod.rs
index 9e5f12dc6..cdb681525 100644
--- a/kernel/src/net/socket/ip/stream/mod.rs
+++ b/kernel/src/net/socket/ip/stream/mod.rs
@@ -112,7 +112,7 @@ impl StreamSocket {
         })
     }
 
-    fn new_accepted(connected_stream: ConnectedStream) -> Arc<Self> {
+    fn new_accepted(connected_stream: ConnectedStream, listener_options: &OptionSet) -> Arc<
Self> {-
         let options = connected_stream.raw_with(|raw_tcp_socket| {
             let mut options = OptionSet::new();
 
@@ -124,6 +124,9 @@ impl StreamSocket {
                 options.tcp.set_no_delay(true);
             }
 
+            // FIXME: This is just a placeholder example.
+            options.tcp.set_keep_idle(listener_options.tcp.keep_idle());
+
             // TODO: Update other options for a newly-accepted socket
 
             options
@@ -303,7 +306,8 @@ impl StreamSocket {
 
         let accepted = listen_stream.try_accept().map(|connected_stream| {
             let remote_endpoint = connected_stream.remote_endpoint();
-            let accepted_socket = Self::new_accepted(connected_stream);
+            let listener_options = self.options.read();
+            let accepted_socket = Self::new_accepted(connected_stream, &listener_options);
             (accepted_socket as _, remote_endpoint.into())
         });
         let iface_to_poll = listen_stream.iface().clone();
```

With this approach, the change is minimal and there is no need to introduce the `RawTcpSocketWithKeepAliveInterval` abstraction at all. Note that this patch uses `keep_idle` as a placeholder example — you should replace it with `keep_interval` accordingly.

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -112,10 +115,13 @@ impl StreamSocket {
...
     }
 
-    fn new_accepted(connected_stream: ConnectedStream) -> Arc<Self> {
+    fn new_accepted(connected_stream: ConnectedStream, listener_options: &OptionSet) -> Arc<Self> {
         let options = connected_stream.raw_with(|raw_tcp_socket| {
             let mut options = OptionSet::new();
 
+            options
+                .tcp
+                .set_keep_intvl(listener_options.tcp.keep_intvl());
```
StevenJiang1110: Please move this code to after the block that retrieves options from the raw TCP socket, with blank lines added on both sides to keep it consistent with how other options are formatted.

After the changes in this PR, it occurs to me that exposing socket option-related interfaces on the raw TCP socket — at least the getter interfaces — may actually be unnecessary. However, that is out of scope for this PR, so let's just add a comment for now to document the issue.

```diff
diff --git a/kernel/src/net/socket/ip/stream/mod.rs b/kernel/src/net/socket/ip/stream/mod.rs
index 618df582c..f52c91a7a 100644
--- a/kernel/src/net/socket/ip/stream/mod.rs
+++ b/kernel/src/net/socket/ip/stream/mod.rs
@@ -119,9 +119,6 @@ impl StreamSocket {
         let options = connected_stream.raw_with(|raw_tcp_socket| {
             let mut options = OptionSet::new();
 
-            options
-                .tcp
-                .set_keep_intvl(listener_options.tcp.keep_intvl());
             if raw_tcp_socket.keep_alive().is_some() {
                 options.socket.set_keep_alive(true);
             }
@@ -130,6 +127,17 @@ impl StreamSocket {
                 options.tcp.set_no_delay(true);
             }
 
+            // Some options cannot be retrieved directly from the raw TCP socket,
+            // but still need to be inherited from the listening socket.
+            // Such options are handled here.
+            // TODO: Determine whether all options can be obtained 
+            // from the listening socket, 
+            // so that socket option-related getter interfaces on the raw socket 
+            // can be removed.
+            options
+                .tcp
+                .set_keep_intvl(listener_options.tcp.keep_intvl());
+
             // TODO: Update other options for a newly-accepted socket
 
             options
```

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -849,20 +891,34 @@ impl SetSocketLevelOption for State {
...
 
-    fn set_keep_alive(&self, keep_alive: bool) -> NeedIfacePoll {
-        let interval = if keep_alive {
-            Some(KEEPALIVE_INTERVAL)
-        } else {
-            None
-        };
+    fn set_keep_alive(&self, is_keep_alive_enabled: bool, keep_intvl: u32) -> NeedIfacePoll {
+        let interval = Duration::from_secs(keep_intvl as u64);
+        let keep_alive = is_keep_alive_enabled.then_some(interval);
```
StevenJiang1110:
```suggestion
        let interval = is_keep_alive_enabled.then(|| Duration::from_secs(keep_intvl as u64));
```

The resulted variable should be named `interval` instead of `keep_alive`.

StevenJiang1110 (approved):
LGTM! Thanks for your contribution! Just two small nits.

lrh2000:
Existing socket options are copied from raw sockets because the socket option in `accept()` may not be the same as when the raw socket was created. For example:

```rust
listen(sk)

getsockopt(sk) = old

// incoming connection (sk2)

setsockopt(sk, new)

sk2 = accept(sk);

getsockopt(sk) = new ? old ?
```

Ideally, the `setsockopt()` system call should not be called after `listen()` and `connect()`. The question is, if applications try to do so, how should we respond? Copying the socket option in `accept()` would result in inconsistency between the option set and the raw socket. Is this considered okay?

Though not ideal, perhaps we could try some best-effort approaches. For example, we could prefer the values from the raw sockets first and then fall back to the listener's socket option value at the time of `accept()`.

I'm not sure if we can come up with better ways. Any ideas?

taosue:
> perhaps we could try some best-effort approaches. For example, we could prefer the values from the raw sockets first and then fall back to the listener's socket option value at the time of `accept()`.
> 

Indeed, we can make this behavior more consistent by using raw-socket values whenever they are available, and only falling back to the listener's `OptionSet` for options that the raw socket cannot expose, such as `TCP_KEEPINTVL` when `SO_KEEPALIVE` is disabled.

lrh2000:
> only falling back to the listener's `OptionSet` for options that the raw socket cannot expose, such as `TCP_KEEPINTVL` when `SO_KEEPALIVE` is disabled.

Yes, this is the so-called best-effort approach that I mentioned earlier. I think it would be better than the current approach. It may not be 100% ideal, though.

---

Also could you align the reference URLs to the codebase's existing conventions? Thanks.
```patch
diff --git a/kernel/src/net/socket/ip/stream/mod.rs b/kernel/src/net/socket/ip/stream/mod.rs
index 306c4f699..96c6e133c 100644
--- a/kernel/src/net/socket/ip/stream/mod.rs
+++ b/kernel/src/net/socket/ip/stream/mod.rs
@@ -777,8 +777,7 @@ fn do_tcp_setsockopt(
             // TODO: Track when the socket becomes idle to actually support keep idle.
         }
         tcp_keep_intvl @ KeepIntvl => {
-            // The range matches Linux TCP_KEEPINTVL:
-            // https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h#L168
+            // Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h#L168>
             const MIN_KEEP_INTVL: u32 = 1;
             const MAX_KEEP_INTVL: u32 = 32767;
 
@@ -794,12 +793,12 @@ fn do_tcp_setsockopt(
             return Ok(state.set_keep_alive(options.socket.keep_alive(), *keepintvl));
         }
         tcp_keep_cnt @ KeepCnt => {
-            // The range matches Linux TCP_KEEPCNT:
-            // https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h#L169
+            // Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h#L169>
+            const MIN_KEEP_CNT: u8 = 1;
             const MAX_KEEP_CNT: u8 = 127;
 
             let keepcnt = tcp_keep_cnt.get().unwrap();
-            if *keepcnt < 1 || *keepcnt > MAX_KEEP_CNT {
+            if *keepcnt < MIN_KEEP_CNT || *keepcnt > MAX_KEEP_CNT {
                 return_errno_with_message!(Errno::EINVAL, "the keepalive count is out of bounds");
             }
             options.tcp.set_keep_cnt(*keepcnt);
diff --git a/kernel/src/net/socket/ip/stream/util.rs b/kernel/src/net/socket/ip/stream/util.rs
index e0cf1fa19..a468c539d 100644
--- a/kernel/src/net/socket/ip/stream/util.rs
+++ b/kernel/src/net/socket/ip/stream/util.rs
@@ -22,7 +22,7 @@ pub(super) struct TcpOptionSet {
     receive_inq: bool,
 }
 
-// See https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h
+// Reference: <https://elixir.bootlin.com/linux/v6.13/source/include/net/tcp.h>
 pub(super) const DEFAULT_MAXSEG: u32 = 536;
 pub(super) const DEFAULT_KEEP_IDLE: u32 = 7200;
 pub(super) const DEFAULT_KEEP_INTVL: u32 = 75;
```

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -112,18 +115,33 @@ impl StreamSocket {
...
             if !raw_tcp_socket.nagle_enabled() {
                 options.tcp.set_no_delay(true);
             }
 
+            // Some options cannot be retrieved directly from the raw TCP socket,
+            // but still need to be inherited from the listening socket. Such
+            // options are handled here.
+            // TODO: Determine whether all options can be obtained from the
+            // listening socket, so that socket option-related getter interfaces
+            // on the raw socket can be removed.
```
lrh2000: I believe I have already explained why we can't do this. It would be better to replace this TODO with my explanation.
```rust
// Explain how we inherit socket options
// from `raw_tcp_socket` and `listener_options`.

if let Some(interval) = raw_tcp_socket.keep_alive() {
    todo!()
} else {
    todo!()
}

if !raw_tcp_socket.nagle_enabled() {
    todo!()
}
```

Also there is no need to check `keep_alive()` twice.

kernel/src/net/socket/ip/stream/mod.rs
```diff
@@ -112,12 +115,24 @@ impl StreamSocket {
...
             let mut options = OptionSet::new();
 
-            if raw_tcp_socket.keep_alive().is_some() {
+            // Inherit socket options from `raw_tcp_socket` first, then use a
+            // best-effort fallback to `listener_options` for options the raw
+            // socket cannot expose. The raw socket is created when a connection
+            // arrives, before `accept()` returns it. If the listener's options are
+            // changed after that but before `accept()` returns, using only
+            // `listener_options` would give the accepted socket "new" options
+            // while its raw socket still has the "old" ones.
```
lrh2000: 1. Break this to two paragraphs.
 2. Add a blank line after it (this comment should apply to all socket options, not only the `keep_alive` one).

```suggestion
            // Inherit socket options from `raw_tcp_socket` first, then fall
            // back to `listener_options` for options that the raw socket
            // cannot expose.
            //
            // The raw socket is created when a connection arrives, before
            // `accept()` returns it. If the listener's options are changed
            // after that but before `accept()` returns, using only
            // `listener_options` would give the accepted socket "new" options
            // while its raw socket still has the "old" ones.

```

lrh2000 (approved):
Just a minor comment. Otherwise LGTM. Thanks for your contribution!

cqs21 (approved):

PR #3304, opened 2026-06-02
Fix OSDK self-copy truncation for hard-linked artifacts

taosue:
This fixes an OSDK artifact corruption issue that can be reproduced by running `make kernel BOOT_METHOD=qemu-direct` followed by `make run_kernel BOOT_METHOD=qemu-direct INTEL_TDX=1`: the first build may create `target/osdk/aster-kernel-osdk-bin.qemu_elf` as a hard link to the Cargo-built kernel ELF, so both paths share the same inode; on the second build, `hard_link_or_copy()` sees the destination already exists, `fs::hard_link()` fails, and the fallback `fs::copy(src, dst)` is invoked even though `src` and `dst` refer to the same inode, **which truncates the file and leads to an empty bzImage payload (see [here](https://doc.rust-lang.org/std/fs/fn.copy.html))**.

The fix adds a same-file fast path before attempting the hard link/copy fallback.

### DIFF
--- a/osdk/src/util.rs
+++ b/osdk/src/util.rs
@@ -5,6 +5,7 @@ use std::{
     ffi::OsStr,
     fs::{self, File},
     io::{BufRead, BufReader, Result, Write},
+    os::unix::fs::MetadataExt,
     os::unix::net::UnixStream,
     path::{Path, PathBuf},
     process::Command,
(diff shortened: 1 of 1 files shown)


StevenJiang1110 (approved):

PR #3305, opened 2026-06-02
Update Kata networking guide

StevenJiang1110:
This PR updates the commands for starting Kata with networking enabled. I have modified the Kata runtime to attach a virtio-net device when launching QEMU, so there is no need to support network device hotplug for now.

I have also modified Kata to bypass the network configuration flow that would normally be handled by the Kata agent on the guest kernel side. Instead, networking is configured by creating a network via `nerdctl network create` (which is one step inside `kata-service.sh start`) and specifying the IP address via `nerdctl run`. This ensures that the host-side network configuration matches the gateway and IP values hardcoded in Asterinas, allowing network connectivity to be established.

The updated demo introduced by this PR has been integrated into the CI pipeline of asterinas/kata-containers.

### DIFF
--- a/book/src/kernel/vm-based-containers/kata.md
+++ b/book/src/kernel/vm-based-containers/kata.md
@@ -29,7 +29,7 @@ The commands in this guide are currently written for x86-64 hosts.
 Verify that the required device nodes exist:
 
 ```bash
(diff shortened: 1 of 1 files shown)


tatetian (approved):

PR #3306, opened 2026-06-02
Increase NixOS test default timeout from 5min to 10min

cqs21:
Increase the default NixOS test timeout from 5 minutes to 10 minutes, since the `nix-and-nixos-tools` test suite is prone to timing out under the previous limit.


tatetian (approved):

PR #3308, opened 2026-06-02
net/netlink/route: implement RTM_NEWADDR, RTM_NEWLINK, RTM_NEWROUTE

Rodrigo-Ichaso:
## Summary

This PR implements handlers for three write-type rtnetlink requests that were previously unhandled, enabling standard `iproute2` commands to configure network interfaces at runtime on Asterinas:

```sh
ip link set eth0 up
ip addr add 192.168.0.119/24 dev eth0
ip route add default via 192.168.0.1
```

Without this, all three commands either hung indefinitely waiting for a response or returned `EOPNOTSUPP`.

### `aster-bigtcp`
- Add `set_ipv4_cidr()` and `set_ipv4_gateway()` to `Iface`, `IfaceCommon`, and `PollableIface` so the smoltcp interface can be reconfigured at runtime from kernel netlink handlers.

### RTM_NEWLINK (`do_new_link`)
- Returns ACK (error=0) so the caller does not block on a third `recvmsg`.
- Multicasts a `RTM_NEWLINK` notification to `RTMGRP_LINK` to unblock any monitoring socket subscribed to link events.

### RTM_NEWADDR (`do_new_addr`)
- Parses `IFA_ADDRESS` / `IFA_LOCAL` attributes from the request.
- Calls `iface.set_ipv4_cidr()` to assign the address to the smoltcp interface.

### RTM_NEWROUTE (`do_new_route`)
- New `RouteSegment` / `RouteAttr` types to fully parse the `rtmsg` body and `RTA_*` attributes.
- Parses `RTA_GATEWAY` and `RTA_OIF`, then calls `iface.set_ipv4_gateway()` to install the default route in smoltcp.

### Base netlink message types (NOOP / ERROR / DONE / OVERRUN)
- These were falling through to the default `SkippedErr(EOPNOTSUPP)` arm.
- Now return `Skipped` (no error sent back), which matches Linux kernel behaviour.

### MSG_TRUNC in `recvmsg`
- Honour `MSG_TRUNC` and return the actual message length even when the writer buffer is zero-length.
- `iproute2` uses `MSG_PEEK | MSG_TRUNC` with `iov_len=0` to probe message size before the real read; without this fix the size was reported as 0 and the subsequent read was mishandled.

## Testing

Tested by booting Asterinas with a NixOS guest that runs the three `iproute2` commands above during `stage-2-init`:
- All three commands complete without hanging.
- The interface responds correctly to ARP for the configured address.
- TCP connections reach the smoltcp stack (RST for unbound ports confirms layer-3 routing is functional).

Only AF_INET is handled; AF_INET6 support is left as a follow-up.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/iface/common.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/common.rs
@@ -129,6 +129,14 @@ impl<E: Ext> IfaceCommon<E> {
         self.interface.lock().prefix_len()
     }
 
+    pub(super) fn set_ipv4_cidr(&self, cidr: smoltcp::wire::Ipv4Cidr) {
+        self.interface.lock().set_ipv4_cidr(cidr);
+    }
+
+    pub(super) fn set_ipv4_gateway(&self, gateway: smoltcp::wire::Ipv4Address) {
+        self.interface.lock().set_ipv4_gateway(gateway);
+    }
+
     pub(super) fn sched_poll(&self) -> &E::ScheduleNextPoll {
         &self.sched_poll
     }
--- a/kernel/libs/aster-bigtcp/src/iface/iface.rs
+++ b/kernel/libs/aster-bigtcp/src/iface/iface.rs
@@ -103,6 +103,16 @@ impl<E: Ext> dyn Iface<E> {
         cidr.broadcast()
     }
 
+    /// Sets the IPv4 CIDR of the interface.
+    pub fn set_ipv4_cidr(&self, cidr: smoltcp::wire::Ipv4Cidr) {
+        self.common().set_ipv4_cidr(cidr);
+    }
+
+    /// Sets the default IPv4 gateway.
+    pub fn set_ipv4_gateway(&self, gateway: smoltcp::wire::Ipv4Address) {
+        self.common().set_ipv4_gateway(gateway);
+    }
+
     /// Returns a reference to the associated [`ScheduleNextPoll`].
     ///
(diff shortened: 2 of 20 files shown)


Rodrigo-Ichaso:
Gentle ping on this one — with v0.18.0 expanding the networking stack (IPv6, vsock, namespace support), RTM_NEWADDR, RTM_NEWLINK and RTM_NEWROUTE feel increasingly relevant for tools like `ip addr` and `ip route` to work correctly.

Happy to rebase on main and address any review feedback if this is something the team wants to move forward.

tatetian:
@StevenJiang1110 Could you review this PR?

StevenJiang1110 (commented):
Thanks for your contribution!

I have done a quick scan of this PR. It adds some support for user-space network configuration. Being able to get the `ip` command running is certainly a worthwhile goal, but the current implementation is quite incomplete and conflicts with our planned design in several ways. For instance, Linux's routing table is significantly more complex than smoltcp's and is global in scope, whereas smoltcp's routing table is per-interface. We will need to introduce a compatibility layer in between to bridge these semantic gaps.

However, there are some useful changes in this PR. For example, returning an error segment to user space for unsupported netlink message types — rather than silently returning nothing — is a reasonable improvement. As for `MSG_TRUNC` support, I would suggest dropping it from this PR, as #3202 will provide a more complete implementation. If the author wishes to continue working on this PR, I would recommend focusing on the netlink-side changes rather than prioritizing support for manipulating network interfaces.

PR #3309, opened 2026-06-03
Fix atomic mode violation in Condvar::wait_timeout timeout path

### LINKED ISSUE #3157: Potential violations of kernel atomic mode
### Describe the bug

As documented in `atomic_mode.rs`, triggering sleep-like actions is not allowed when holding a `SpinLock` / `RwLock` (which also disables preempt).
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/ostd/src/task/atomic_mode.rs#L16-L26

However, several potential violation is observed in the current code base:

 ### Case 1: `VsockStreamSocket::bind()` 
L144 in `bind()` first holds `self.status` which is a `RwLock`. The function then calls `init.bind()` at L146, acquiring `self.bound_addr` at L25 (and also L50 in the same function) which is a `Mutex`.
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/src/net/socket/vsock/stream/socket.rs#L142-L146
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/src/net/socket/vsock/stream/init.rs#L24-L27

### Case 2: `VsockStreamSocket::addr()`
Similar to case 1, L290 in `addr()` first holds `self.status (RwLock)` and then calls `init.bound_addr()` which acquires `self.bound_addr (Mutex)` at L55.
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/src/net/socket/vsock/stream/socket.rs#L290-L293
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/src/net/socket/vsock/stream/init.rs#L54-L56

### Case 3: `Condvar::wait_timeout()`
In the `Err(_)` branch, `lock.lock()` (which is a `Mutex`) at L172 is called while hoding `self.counter` (which is a `SpinLock`), since `counter` has not gone out of its scope yet.
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/kernel/src/process/sync/condvar.rs#L169-L173

Thanks for your time and attention. Could you please kindly confirm whether this is a valid issue, or let me know if I have misunderstood anything in my description?

sximelon:
Fixes #3157.

## Problem
    
In `Condvar::wait_timeout`, the `Err(_)` (timeout) path held `self.counter` (a `SpinLock` with `PreemptDisabled` guard) across a call to `lock.lock()` (a `Mutex`). 
When the mutex is contested, `Mutex::lock` may sleep (`WaitQueue::wait_until` → `park_current` → `switch_to_task` →
`might_sleep`), which panics if `guard_count > 0`.
    
```
match res {
    Ok(()) => Ok((lock.lock(), false)),
    Err(_) => {
        let mut counter = self.counter.lock();  // guard_count +1
        counter.waiter_count -= 1;
        Err(LockErr::Timeout((lock.lock(), true))) // ⚠️ might_sleep with guard held
    }  // counter dropped here — too late
}
```
    
## Solution

Replace the named `SpinLockGuard` binding with an inline temporary, which Rust drops at the statement semicolon before `lock.lock()` runs.
    
```
Err(_) => {
    - let mut counter = self.counter.lock();
    - counter.waiter_count -= 1;
    +    self.counter.lock().waiter_count -= 1;
    Err(LockErr::Timeout((lock.lock(), true)))
}
```
    
## Related

- Issue #3157 (Case 3)

### DIFF
--- a/kernel/src/process/sync/condvar.rs
+++ b/kernel/src/process/sync/condvar.rs
@@ -167,8 +167,7 @@ impl Condvar {
         match res {
             Ok(()) => Ok((lock.lock(), false)),
             Err(_) => {
-                let mut counter = self.counter.lock();
-                counter.waiter_count -= 1;
+                self.counter.lock().waiter_count -= 1;
                 Err(LockErr::Timeout((lock.lock(), true)))
             }
         }


StevenJiang1110 (approved):

PR #3310, opened 2026-06-03
Preserve EOF during cached writeback for virtiofs

li041:
This PR fixes a virtiofs cached-write bug where writing a small file could incorrectly make the file size become one page, typically 4096 bytes.

## Root Cause 
Virtiofs page-cache writeback always submitted a full-page `FUSE_WRITE`, even when the dirty range ended before `EOF`. For example, writing a few bytes to a new file dirtied one cache page, and writeback sent the whole 4096-byte page to the virtiofs server.

## Solution
This PR updates virtiofs writeback to cap each page write by the inode EOF:
  - Cached writes temporarily publish the target EOF before flushing page cache data.
  - `write_page_async` uses the current inode size to calculate the valid writeback length for the page.
  - The final page is written only up to `EOF` instead of always writing `PAGE_SIZE`.

### DIFF
--- a/kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
@@ -156,26 +156,32 @@ impl VirtioFsInode {
         let page_cache = inner.page_cache().unwrap();
 
         if requested_end > old_size {
-            page_cache.resize(requested_end, old_size)?;
+            // Extend the visible EOF before growing the page cache.
+            self.set_size(requested_end);
+            page_cache
+                .resize(requested_end, old_size)
+                .expect("expanding the page cache should not fail");
         }
 
-        // If a later step fails, the enlarged page-cache capacity does not
-        // publish an EOF extension. The cached inode size advances only after
-        // writeback succeeds, and cached reads are limited by that size.
+        let mut write_through_page_cache = || -> Result<()> {
+            page_cache.write(offset, reader).map_err(Error::from)?;
+            page_cache.flush_range(offset..requested_end)
+        };
 
-        page_cache.write(offset, reader)?;
-
-        // FIXME: The page cache writeback API reports only the page index, not
-        // the dirty byte range within the page. Although this call flushes the
-        // requested byte range, the virtio-fs client currently only submits
-        // whole-page writes to the server. This can write bytes outside the
-        // user request.
-        page_cache.flush_range(offset..requested_end)?;
+        if let Err(err) = write_through_page_cache() {
+            if requested_end > old_size {
+                // Roll back in truncate order: shrink the page cache before
+                // restoring the old visible EOF.
+                page_cache.resize(old_size, requested_end)?;
+                self.set_size(old_size);
+            }
+            return Err(err);
+        }
 
         let new_size = self.size().max(requested_end);
         let attr_version = self.fs_ref().session().bump_attr_version();
+
         inner.commit_local_write(new_size, attr_version);
-        self.set_size(new_size);
 
         Ok(write_len)
     }
--- a/kernel/src/fs/fs_impls/virtiofs/inode/page_cache.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/inode/page_cache.rs
@@ -76,9 +76,17 @@ impl PageCacheBackend for VirtioFsInode {
     ) -> Result<()> {
         locked_page.wait_until_finish_writing_back();
 
+        let page_start = page_offset(idx)?;
+        let file_size = self.size();
+        if page_start >= file_size {
+            return_errno_with_message!(Errno::EINVAL, "virtiofs writeback page is beyond EOF");
+        }
+        let writeback_len = PAGE_SIZE.min(file_size - page_start);
+
         let fs = self.fs_ref();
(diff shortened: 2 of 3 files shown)


kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
```diff
@@ -159,20 +159,16 @@ impl VirtioFsInode {
...
+
+        if let Err(err) = page_cache.flush_range(offset..requested_end) {
+            self.set_size(old_size);
+            return Err(err);
+        }
 
-        let new_size = self.size().max(requested_end);
         let attr_version = self.fs_ref().session().bump_attr_version();
         inner.commit_local_write(new_size, attr_version);
         self.set_size(new_size);
```
cchanging: We can remove the `set_size` here?

kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
```diff
@@ -159,20 +159,16 @@ impl VirtioFsInode {
...
         page_cache.write(offset, reader)?;
 
-        // FIXME: The page cache writeback API reports only the page index, not
-        // the dirty byte range within the page. Although this call flushes the
-        // requested byte range, the virtio-fs client currently only submits
-        // whole-page writes to the server. This can write bytes outside the
-        // user request.
-        page_cache.flush_range(offset..requested_end)?;
+        let new_size = old_size.max(requested_end);
+        self.set_size(new_size);
```
cchanging: Keep consistency in size updates.

```rust
if requested_end > old_size {
    self.set_size(requested_end);
    page_cache.resize(requested_end, old_size)?;
}

page_cache.write(offset, reader)?;
```

kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
```diff
@@ -159,20 +159,16 @@ impl VirtioFsInode {
...
+
+        if let Err(err) = page_cache.flush_range(offset..requested_end) {
+            self.set_size(old_size);
+            return Err(err);
+        }
```
cchanging: Also resize the page cache?

kernel/src/fs/fs_impls/virtiofs/inode/page_cache.rs
```diff
@@ -76,9 +76,18 @@ impl PageCacheBackend for VirtioFsInode {
     ) -> Result<()> {
         locked_page.wait_until_finish_writing_back();
 
+        let page_start = page_offset(idx)?;
+        let writeback_end = self.size();
+        if page_start >= writeback_end {
+            locked_page.set_up_to_date();
+            return Ok(());
+        }
```
cchanging: See the document of `write_page_async`, currently we prefer to return `EINVAL`.

kernel/src/fs/fs_impls/virtiofs/inode/page_cache.rs
```diff
@@ -76,9 +76,18 @@ impl PageCacheBackend for VirtioFsInode {
     ) -> Result<()> {
         locked_page.wait_until_finish_writing_back();
 
+        let page_start = page_offset(idx)?;
+        let writeback_end = self.size();
```
cchanging: I think we can directly name this variable `file_size`.

cchanging (commented):
Thanks for fixing this bug!

kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
```diff
@@ -156,26 +156,38 @@ impl VirtioFsInode {
         let page_cache = inner.page_cache().unwrap();
 
         if requested_end > old_size {
-            page_cache.resize(requested_end, old_size)?;
+            // Extend the visible EOF before growing the page cache.
+            self.set_size(requested_end);
+            if let Err(err) = page_cache.resize(requested_end, old_size) {
```
cchanging: I found that when the `resize` is an expand operation, it will not fail. So I think we can add a document to `resize` and unwrap expand `resize` directly.

kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
```diff
@@ -156,26 +156,38 @@ impl VirtioFsInode {
...
-        page_cache.flush_range(offset..requested_end)?;
+        if let Err(err) = page_cache.flush_range(offset..requested_end) {
+            if requested_end > old_size {
+                // Roll back in truncate order: shrink the page cache before
+                // restoring the old visible EOF.
+                page_cache.resize(old_size, requested_end)?;
+                self.set_size(old_size);
+            }
+            return Err(err);
+        }
```
cchanging:
```suggestion
            let mut write_through_page_cache = || -> Result<()> {
                page_cache.write(offset, reader).map_err(Error::from)?;
                page_cache.flush_range(offset..requested_end)
            };

            if let Err(err) = write_through_page_cache() {
               if requested_end > old_size {
                    // Roll back in truncate order: shrink the page cache before
                    // restoring the old visible EOF.
                    page_cache.resize(old_size, requested_end)?;
                    self.set_size(old_size);
                }
                return Err(err);
            }
```

In the future, we may need to explore an approach similar to #3297 and use an RAII mechanism to implement rollback.

cchanging (approved):

PR #3311, opened 2026-06-03
Correct and optimize the create-related operations in `Ext2`

cchanging:
Currently, when the VFS performs create-related operations, it only checks whether there is a child with the same name in the dentry cache, and then directly calls the filesystem layer’s `create` operation. At the same time, our filesystem-layer implementation does not properly check for duplicate names. As a result, due to races or later eviction under memory pressure, the corresponding child may be absent from the dentry cache even though a child with the same name actually exists in the filesystem layer, which can lead to creating duplicate entries.

Therefore, the first commit in this PR moves the complete duplicate-name check into the VFS layer, which reduces the complexity of the filesystem-layer logic and is consistent with Linux’s behavior. 

The second commit optimizes the slot lookup in `Ext2` file creation, reducing many unnecessary page cache read operations.


StevenJiang1110 (approved):

PR #3315, opened 2026-06-03
Fix virtiofs request queue descriptor exhaustion

li041:
## Summary
- wait for free virtqueue descriptors before submitting virtio-fs requests
- propagate submit failures through the FUSE submit path instead of unwrapping
- reject virtio-fs queues that cannot fit the maximum descriptors needed by one request

### DIFF
--- a/kernel/comps/virtio/src/device/filesystem/device/queue.rs
+++ b/kernel/comps/virtio/src/device/filesystem/device/queue.rs
@@ -9,11 +9,12 @@
 use alloc::{sync::Arc, vec::Vec};
 use core::mem;
 
+use aster_fuse::FuseError;
 use aster_softirq::Taskless;
 use aster_util::{mem_obj_slice::Slice, slot_vec::SlotVec};
 use ostd::{
     mm::dma::{FromDevice, ToDevice},
-    sync::{LocalIrqDisabled, SpinLock},
+    sync::{LocalIrqDisabled, SpinLock, WaitQueue},
 };
 use smallvec::SmallVec;
 use spin::Once;
@@ -21,9 +22,17 @@ use spin::Once;
 use super::request::FuseRequest;
 use crate::{device::filesystem::pool::FsDmaStorage, queue::VirtQueue};
 
+/// Maximum virtqueue descriptors used by one FUSE request.
+///
+/// Current requests use at most two to-device buffers (request header/body and
+/// optional write payload) and two from-device buffers (reply header and
+/// optional read payload).
+pub(super) const MAX_DMA_BUFS_PER_REQUEST: usize = 4;
+
 /// A virtio-fs request queue and its in-flight request state.
 pub(super) struct FsRequestQueue {
     inner: SpinLock<FsRequestQueueInner, LocalIrqDisabled>,
+    wait_queue: WaitQueue,
     taskless: Once<Arc<Taskless>>,
 }
 
@@ -47,6 +56,7 @@ impl FsRequestQueue {
                 in_flight_requests: SlotVec::new(),
                 pending_completions: Vec::new(),
             }),
+            wait_queue: WaitQueue::new(),
             taskless: Once::new(),
         })
     }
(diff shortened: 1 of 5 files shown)


kernel/comps/virtio/src/device/filesystem/device/queue.rs
```diff
@@ -47,6 +57,8 @@ impl FsRequestQueue {
                 in_flight_requests: SlotVec::new(),
                 pending_completions: Vec::new(),
             }),
+            wait_queue: WaitQueue::new(),
+            max_dma_bufs_per_request: MAX_DMA_BUFS_PER_REQUEST,
```
cchanging: Why store the constant in the structure? We can always directly use `MAX_DMA_BUFS_PER_REQUEST`.

kernel/comps/virtio/src/device/filesystem/device/queue.rs
```diff
@@ -73,43 +85,70 @@ impl FsRequestQueue {
...
-            inner
-                .queue
-                .add_dma_bufs(request_bufs.as_slice(), reply_bufs.as_slice())
-                .unwrap()
-        };
+    pub(super) fn add_request(&self, request: FuseRequest) -> Result<(), FuseError> {
+        let num_dma_bufs = request.num_dma_bufs();
+        if num_dma_bufs > self.max_dma_bufs_per_request {
+            return Err(FuseError::BufferTooSmall);
+        }
```
cchanging: I think the `num_dma_bufs` will always be smaller than `MAX_DMA_BUFS_PER_REQUEST`. Can’t it only be 3 or 4?

kernel/comps/virtio/src/device/filesystem/device/queue.rs
```diff
@@ -73,43 +85,70 @@ impl FsRequestQueue {
...
+                    return Ok(());
+                }
+            }
 
-        if inner.queue.should_notify() {
-            inner.queue.notify();
+            self.wait_queue.wait_until(|| {
+                let inner = self.inner.lock();
+                (num_dma_bufs <= inner.queue.available_desc()).then_some(())
+            });
```
cchanging:
```suggestion
        let mut request = Some(request);
        self.wait_queue.wait_until(|| {
            let mut inner = self.inner.lock();
            if num_dma_bufs > inner.queue.available_desc() {
                return None;
            }

            let request = request.take().unwrap();
            let request_bufs: SmallVec<[_; 2]> = request
                .request_bufs()
                .iter()
                .map(|buf| buf.as_dma_slice())
                .collect();
            let reply_bufs: SmallVec<[_; 2]> = request
                .waiter()
                .reply_bufs()
                .iter()
                .map(|buf| buf.as_dma_slice())
                .collect();

            let token = inner
                .queue
                .add_dma_bufs(request_bufs.as_slice(), reply_bufs.as_slice())
                .unwrap();

            inner.in_flight_requests.put_at(token as usize, request);

            if inner.queue.should_notify() {
                inner.queue.notify();
            }

            Some(())
        });

        Ok(())
    }
```

kernel/comps/virtio/src/device/filesystem/device/queue.rs
```diff
@@ -73,43 +85,70 @@ impl FsRequestQueue {
...
+            while let Ok((token, len)) = inner.queue.pop_used() {
+                has_freed_descs = true;
+                let reply_len = len as usize;
+                let request = inner.in_flight_requests.remove(token as usize).unwrap();
+                inner
+                    .pending_completions
+                    .push(CompletedFuseRequest { request, reply_len });
+            }
+            has_freed_descs
+        };
```
cchanging:
```suggestion
    let mut inner = self.inner.lock();
    let mut has_freed_descs = false;
    while let Ok((token, len)) = inner.queue.pop_used() {
        has_freed_descs = true;
        let reply_len = len as usize;
        let request = inner.in_flight_requests.remove(token as usize).unwrap();
        inner
            .pending_completions
            .push(CompletedFuseRequest { request, reply_len });
    }
    drop(inner);
      
    if has_freed_descs {
          self.wait_queue.wake_all();
    }
```

In fact we can do `wake_all` under a `SpinLock`. Yet doing an early drop here is also fine.

cchanging (commented):
Thanks for fixing this panic bug!

kernel/comps/virtio/src/device/filesystem/device/queue.rs
cchanging: This comment doesn’t seem to have been resolved. I think this could be changed to a `debug_assert!`.

cchanging (approved):

PR #3316, opened 2026-06-03
Fix block device error codes and enhance virtio-blk feature negotiation

hburaylee:
## Summary
This PR fix block device error codes and enhance virtio-blk feature negotiation.

## Changes
- Fix bio error reporting to use EOPNOTSUPP for unsupported operations
- Enhance virtio-blk feature negotiation for BLK_SIZE and FLUSH
- Rename `support_features` to `supported_features` for NetworkFeatures to better reflect its purpose

### DIFF
--- a/kernel/comps/virtio/src/device/block/device.rs
+++ b/kernel/comps/virtio/src/device/block/device.rs
@@ -27,7 +27,7 @@ use ostd::{
     sync::SpinLock,
 };
 
-use super::{BlockFeatures, VirtioBlockConfig, VirtioBlockFeature};
+use super::{BlockFeatures, VirtioBlockConfig};
 use crate::{
     VIRTIO_BLOCK_MAJOR_ID,
     device::{
@@ -124,10 +124,8 @@ impl BlockDevice {
     }
 
     /// Negotiate features for the device specified bits 0~23
-    pub(crate) fn negotiate_features(features: u64) -> u64 {
-        let mut support_features = BlockFeatures::from_bits_truncate(features);
-        support_features.remove(BlockFeatures::MQ);
-        support_features.bits
+    pub(crate) fn negotiate_features(device_features: u64) -> u64 {
+        BlockFeatures::negotiated_with_device(device_features).bits()
     }
 }
 
@@ -199,7 +197,7 @@ impl aster_block::BlockDevice for BlockDevice {
 #[derive(Debug)]
 struct DeviceInner {
     config_manager: ConfigManager<VirtioBlockConfig>,
-    features: VirtioBlockFeature,
+    features: BlockFeatures,
     queue: SpinLock<VirtQueue>,
     transport: SpinLock<Box<dyn VirtioTransport>>,
     block_requests: Arc<DmaStream>,
@@ -218,7 +216,16 @@ impl DeviceInner {
         let config = config_manager.read_config();
         debug!("virio_blk_config = {:?}", config);
 
-        let block_size = config_manager.block_size();
+        let features = BlockFeatures::negotiated_with_device(transport.read_device_features());
+
+        let block_size = if features.contains(BlockFeatures::BLK_SIZE) {
+            config_manager.block_size()
+        } else {
+            // Fall back to standard sector size SECTOR_SIZE (512 bytes)
+            // Reference:
+            // <https://docs.oasis-open.org/virtio/virtio/v1.2/virtio-v1.2.html#x1-2790004>
+            VirtioBlockConfig::sector_size()
+        };
         if block_size != VirtioBlockConfig::sector_size() {
             ostd::error!("block size {} is not supported yet", block_size);
             return Err(VirtioDeviceError::UnsupportedConfig);
@@ -233,8 +240,6 @@ impl DeviceInner {
                 "Multi-Queue Block IO Queueing Mechanism is not supported yet; using the first queue"
             );
         }
-
-        let features = VirtioBlockFeature::new(transport.as_ref());
 
         let queue = VirtQueue::new(0, Self::QUEUE_SIZE, transport.as_mut())?;
 
@@ -477,7 +482,7 @@ impl DeviceInner {
     /// Flushes any cached data from the guest to the persistent storage on the host.
     /// This will be ignored if the device doesn't support the `VIRTIO_BLK_F_FLUSH` feature.
     fn flush(&self, bio_request: BioRequest) {
-        if !self.features.support_flush {
+        if !self.features.contains(BlockFeatures::FLUSH) {
             bio_request.into_bios().for_each(|bio| {
                 bio.complete(BioStatus::Complete);
             });
--- a/kernel/comps/virtio/src/device/block/mod.rs
+++ b/kernel/comps/virtio/src/device/block/mod.rs
@@ -30,6 +30,14 @@ bitflags! {
         const MQ            = 1 << 12;
         const DISCARD       = 1 << 13;
         const WRITE_ZEROES  = 1 << 14;
+
+        const ALL_SUPPORTED = Self::BLK_SIZE.bits() | Self::FLUSH.bits();
+    }
+}
+
+impl BlockFeatures {
+    pub(super) fn negotiated_with_device(device_features: u64) -> Self {
+        BlockFeatures::from_bits_truncate(device_features) & Self::ALL_SUPPORTED
     }
 }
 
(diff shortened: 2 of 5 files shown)


hburaylee:
@boterinas codex

zjp-CN:
Codex review [fails](https://github.com/asterinas/asterinas/actions/runs/27132367166/job/80076231206) for the author @hburaylee due to [check-write-access](https://github.com/search?q=repo%3Aopenai%2Fcodex-action%20check-write-access&type=code) in codex action, because he doesn't have write permission as a *contributor*. Someone with write access needs to trigger codex review for the author.

StevenJiang1110:
@boterinas codex

kernel/comps/virtio/src/device/block/mod.rs
```diff
@@ -120,6 +127,7 @@ struct VirtioBlockTopology {
 #[derive(Clone, Copy, Debug)]
 struct VirtioBlockFeature {
     support_flush: bool,
+    support_blk_size: bool,
 }
```
StevenJiang1110: I don't quite see the need for `VirtioBlockFeature`. All usages of `VirtioBlockFeature` can be replaceable with `BlockFeature`, right? The struct design also seems a bit odd — if every feature requires a dedicated `support_xxx` field, why not use a bitflag instead?

This issue predates the current PR. If possible, it would still be great to include this cleanup here as well.

kernel/comps/virtio/src/device/block/mod.rs
```diff
@@ -33,6 +33,13 @@ bitflags! {
     }
 }
 
+impl BlockFeatures {
+    pub(super) fn support_features() -> Self {
```
StevenJiang1110: Maybe we should rename it to `supported_features`, which is grammatically more correct. It also looks like `support_features` is used in another place — on `NetworkFeatures` — so it would be best to fix both at the same time.

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -218,7 +218,9 @@ impl DeviceInner {
         let config = config_manager.read_config();
         debug!("virio_blk_config = {:?}", config);
 
-        let block_size = config_manager.block_size();
+        let features = VirtioBlockFeature::new(transport.as_ref());
+
+        let block_size = config_manager.block_size(features.support_blk_size);
```
StevenJiang1110: Why not simplify this to:

```rust
let block_size = if features.support_blk_size {
    config_manager.block_size()
} else {
    SECTOR_SIZE
}
```

There doesn't seem to be a need to pass `features.support_blk_size` into `config_manager.block_size()`.

StevenJiang1110 (commented):
Mostly looks good to me! Only a few minor comments.

kernel/comps/virtio/src/device/block/mod.rs
hburaylee: > I don't quite see the need for `VirtioBlockFeature`. All usages of `VirtioBlockFeature` can be replaceable with `BlockFeature`, right? The struct design also seems a bit odd — if every feature requires a dedicated `support_xxx` field, why not use a bitflag instead?
> 
> This issue predates the current PR. If possible, it would still be great to include this cleanup here as well.

Good point. I removed `VirtioBlockFeature` and now use `BlockFeatures` directly.

kernel/comps/virtio/src/device/block/mod.rs
hburaylee: > Maybe we should rename it to `supported_features`, which is grammatically more correct. It also looks like `support_features` is used in another place — on `NetworkFeatures` — so it would be best to fix both at the same time.

Renamed it to `supported_features`. I also updated in `NetworkFeatures`.

kernel/comps/virtio/src/device/block/device.rs
hburaylee: > Why not simplify this to:
> 
> ```rust
> let block_size = if features.support_blk_size {
>     config_manager.block_size()
> } else {
>     SECTOR_SIZE
> }
> ```
> 
> There doesn't seem to be a need to pass `features.support_blk_size` into `config_manager.block_size()`.

Agreed! fixed.

hburaylee:
@boterinas rerun

StevenJiang1110 (approved):

hburaylee:
@sdww0 @lrh2000 @tangruize Gentle ping!

Could you please take a look when you have a moment? Thank you!

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -218,7 +216,17 @@ impl DeviceInner {
...
-        let block_size = config_manager.block_size();
+        let features = BlockFeatures::negotiate_features(transport.read_device_features());
+
+        let block_size = if features.contains(BlockFeatures::BLK_SIZE) {
+            config_manager.block_size()
+        } else {
+            // Fall back to standard sector size SECTOR_SIZE (512 bytes)
+            // Reference:
+            // <https://elixir.bootlin.com/linux/v7.0/source/drivers/block/virtio_blk.c#L1441>
+            // <https://elixir.bootlin.com/linux/v7.0/source/drivers/block/virtio_blk.c#L1286>
```
tatetian: Virtio Specfication is the authoritative source; Linux source code is implemented to comply with the spec. So the code should reference the Virtio Spec.

kernel/comps/virtio/src/device/block/mod.rs
```diff
@@ -33,6 +33,17 @@ bitflags! {
...
 }
 
+impl BlockFeatures {
+    pub(super) fn supported_features() -> Self {
+        // Can be added gradually later: CONFIG_WCE, CONFIG_UNMAP, etc.
+        BlockFeatures::FLUSH | BlockFeatures::BLK_SIZE
+    }
+
+    pub(super) fn negotiate_features(device_features: u64) -> Self {
+        BlockFeatures::from_bits_truncate(device_features) & Self::supported_features()
```
tatetian: This method name does not read like a constructor. A noun form method name is usually only for simply getter, but this method should be considered a constructor. I suggest calling it `negotiated_with_device`.

kernel/comps/virtio/src/device/block/mod.rs
```diff
@@ -33,6 +33,17 @@ bitflags! {
     }
 }
 
+impl BlockFeatures {
+    pub(super) fn supported_features() -> Self {
+        // Can be added gradually later: CONFIG_WCE, CONFIG_UNMAP, etc.
+        BlockFeatures::FLUSH | BlockFeatures::BLK_SIZE
```
tatetian: In this specific case, a better choice would be associating a const to `BlockFeatures`:

```rust
bitflags! {
    /// features for virtio block device
    struct BlockFeatures : u64 {
        const BARRIER       = 1 << 0;
        const SIZE_MAX      = 1 << 1;
        const SEG_MAX       = 1 << 2;
        const GEOMETRY      = 1 << 4;
        const RO            = 1 << 5;
        const BLK_SIZE      = 1 << 6;
        const SCSI          = 1 << 7;
        const FLUSH         = 1 << 9;
        const TOPOLOGY      = 1 << 10;
        const CONFIG_WCE    = 1 << 11;
        const MQ            = 1 << 12;
        const DISCARD       = 1 << 13;
        const WRITE_ZEROES  = 1 << 14;

        const ALL_SUPPORTED = 0 |
            Self::BLK_SIZE::bits |
            Self::FLUSH::bits;
    }
}
```

Using a const conveys the idea that the supported features are compiled time determined.

tatetian (approved):
Only a few nits

kernel/comps/virtio/src/device/block/device.rs
hburaylee: > Virtio Specfication is the authoritative source; Linux source code is implemented to comply with the spec. So the code should reference the Virtio Spec.

Done. Changed the link to the corresponding chapter of Virtio Specification.

kernel/comps/virtio/src/device/block/mod.rs
hburaylee: > This method name does not read like a constructor. A noun form method name is usually only for simply getter, but this method should be considered a constructor. I suggest calling it `negotiated_with_device`.

`negotiate_features` has been renamed to `negotiated_with_device` as suggested, the semantics are clearer, thank you!

kernel/comps/virtio/src/device/block/mod.rs
hburaylee: It has been modified to define `const ALL_SUPPORTED` in `bitflags!`, which is more consistent with the semantics of the feature set determined at compile time.

hburaylee:
@tatetian  
All three nits addressed. Also rebased onto main. PTAL, thanks!

tatetian (approved):

PR #3320, opened 2026-06-03
Enable and fix `clippy::inconsistent_struct_constructor`

lrh2000:
When reviewing https://github.com/asterinas/asterinas/pull/3290, I found that the order of fields used to construct a `Process` is different from the order used to define a `Process`.

https://github.com/asterinas/asterinas/blob/6b183be003ab3dac666f18f8599f77b8b7253b11/kernel/src/process/process/mod.rs#L254-L255

https://github.com/asterinas/asterinas/blob/6b183be003ab3dac666f18f8599f77b8b7253b11/kernel/src/process/process/mod.rs#L106-L109

So let's fix this by enabling [`clippy::inconsistent_struct_constructor`](https://rust-lang.github.io/rust-clippy/master/index.html#inconsistent_struct_constructor)...


ostd/src/arch/riscv/cpu/context.rs
```diff
@@ -415,6 +415,7 @@ pub struct QFpuContext {
 //
 // See <https://github.com/rust-lang/rust/issues/61415>.
 impl Default for QFpuContext {
+    #[expect(clippy::inconsistent_struct_constructor)]
```
tatetian: Could you shed light as to why there are any inconsistency in field ordering? The same question pops up in my head when I see `expect(clippy::inconsistent_struct_constructor)` for `Stat` and `ucontext_t`.

tatetian (commented):
Thanks for adding this new lint.

ostd/src/arch/riscv/cpu/context.rs
lrh2000: The correct order is to insert the padding between the fields. For example:
```rust
            f: [0; 64],
            __pad1: [0; _],
            fcsr: 0,
            __pad2: [0; _],
```

I think we may not want to follow it because it harms the readability.

> ucontext_t

This is same.

> Stat

This is because x86 and non-x86 have different field orders, so we cannot make them both happy. So the expectation is:
```rust
    #[cfg_attr(
        not(target_arch = "x86_64"),
        expect(clippy::inconsistent_struct_constructor)
    )]
```

Here are the exact fields:

 - non-x86
```
    /// Link count.
    st_nlink: u64,
    /// File mode.
    st_mode: u32,
```

 - x86
```
    /// File mode.
    st_mode: u32,
    /// Link count.
    st_nlink: u32,
```

---

Given that you ask this question, not sure if you feel this is better or do you have better words:
```rust
    #[expect(
        clippy::inconsistent_struct_constructor,
        reason = "padding fields are reordered to the end for readability"
    )]
```

(I don't think `Stat` needs a reason. It is already gated by `target_arch`, so the reason seems much more obvious...)

tatetian: Thanks for the explanation!

```rust
impl Default for QFpuContext {
    #[expect(
        clippy::inconsistent_struct_constructor,
        reason = "padding fields are reordered to the end for readability"
    )]
    fn default() -> Self {
        Self {
            f: [0; 64],
            fcsr: 0,
            __pad1: [0; _],
            __pad2: [0; _],
        }
    }
}
```

Positioning the padding fields to the last is out of good intention. And I appreciate that you added the `reason` field.

But I think re-positioning the padding fields to the last is doing the opposite of your intention, which is to improve readability. This is because **the position of a padding field carries significance**. Moving all padding fields to the last makes it unclear where the paddings are intended to be inserted.

The following form feels more clear to me and requires no `reason` field to explain.

```rust
impl Default for QFpuContext {
    fn default() -> Self {
        Self {
            f: [0; 64],
            __pad1: [0; _],
            fcsr: 0,
            __pad2: [0; _],
        }
    }
}
```

lrh2000: > Moving all padding fields to the last makes it unclear where the paddings are intended to be inserted.

OK. I can see your point as well, which does make sense. I'll just fix the field order, then. Thanks for the comments.

tatetian (approved):

PR #3321, opened 2026-06-03
Remove an unncessary `fs_base.get()` during `clone()`

lrh2000:
Another minor improvement (cc https://github.com/asterinas/asterinas/pull/3286)..

We shouldn't need to call `fs_base().get()` if `CLONE_SETTLS` is specified, right?


tatetian (approved):

PR #3323, opened 2026-06-04
Revert `TrapFrame::_pad` removal for soundness reasons

lrh2000:
https://github.com/asterinas/asterinas/pull/3286 removes `TrapFrame::_pad`. But **doing so is unsound** because it violates the calling conventions specified in the System V ABI.

Read this note:
```rust
// The padding field in `TrapFrame` is necessary. Do not attempt to remove it
// without considering this note. Otherwise, the code will be **unsound**.
//
// According to the System V AMD64 ABI, the stack pointer should be aligned to
// at least 16 bytes. The hardware will align the stack pointer to a 16-byte
// boundary for exceptions and interrupts ("In IA-32e mode, the RSP is aligned
// to a 16-byte boundary before pushing the stack frame"), so we only need to
// ensure the size of a `TrapFrame` is also aligned.
crate::const_assert!(size_of::<TrapFrame>().is_multiple_of(16));
```

### DIFF
--- a/ostd/src/arch/x86/cpu/context/mod.rs
+++ b/ostd/src/arch/x86/cpu/context/mod.rs
@@ -380,6 +380,7 @@ impl UserContextApiInternal for UserContext {
             r13: self.user_context.general.r13,
             r14: self.user_context.general.r14,
             r15: self.user_context.general.r15,
+            _pad: 0,
             trap_num: self.user_context.trap_num,
             error_code: self.user_context.error_code,
             rip: self.user_context.general.rip,
--- a/ostd/src/arch/x86/trap/mod.rs
+++ b/ostd/src/arch/x86/trap/mod.rs
@@ -82,6 +82,7 @@ pub struct TrapFrame {
     pub r13: usize,
     pub r14: usize,
     pub r15: usize,
+    pub _pad: usize,
 
     pub trap_num: usize,
     pub error_code: usize,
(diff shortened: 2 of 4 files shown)


tatetian (approved):

tatetian:
Did you find the bug by review or you encounter a mysterious crash?

zjp-CN:
It'd be good to leave a short comment on the padding field and refer to the alignment check. 

In this case, it's fine because the assertion code is nearby.

lrh2000:
> Did you find the bug by review or you encounter a mysterious crash?

By review.

I have previously experienced some strange crashes, particularly during CTF competitions, when the exploit mysteriously stops working. So I am now extremely sensitive about removing a single 8-byte field from the stack frame. This is likely to cause problems at some point due to code changes or just compiler updates.

> In this case, it's fine because the assertion code is nearby.

It is very close (just below the struct) so that's the reason why I omit the comments... But if there is a comment, it is certainly better. I agree.

---

But indeed, I still have to make a follow-up PR because I found the RSP value reported in `TrapFrame` cannot be correct in the current way due to (see the PR description):
>> In IA-32e mode, the RSP is aligned to a 16-byte boundary before pushing the stack frame

In the follow-up PR, `_pad` and the original `rsp` field will be removed together (so their 16 bytes go away)....

I'm really sorry for changing the code back and forth. I should catch this a bit earlier so I can just submit the desired changes within this PR. I hope this won't cause too much noise.

See: https://github.com/asterinas/asterinas/pull/3326.

PR #3325, opened 2026-06-04
Replace `has_kernel_event` and `inject_pre_user_run_handler` with `UserModeHooks`

lrh2000:
Emmm... This should be my last follow-up to https://github.com/asterinas/asterinas/pull/3286. Hopefully!

`UserModeHooks` sounds like a better idea than `inject_pre_user_run_handler`, doesn't it?
```rust
/// Hooks that will be called during [`UserMode::execute`].
pub trait UserModeHooks {
    /// Checks whether a kernel event is pending.
    ///
    /// This method will be called after user space is interrupted
    /// by external interrupts. If the result is `true`,
    /// [`UserMode::execute`] will return with [`ReturnReason::KernelEvent`].
    fn has_kernel_event(&self) -> bool {
        false
    }

    /// Prepares user space execution.
    ///
    /// This method will be called just before entering user space.
    /// Local IRQs are disabled and will only be enabled after entering user space.
    fn pre_user_run(&self, _guard: &DisabledLocalIrqGuard) {}
}

impl UserMode {
    pub fn execute<T: UserModeHooks>(&mut self, hooks: &T) -> ReturnReason;
}
```

### DIFF
--- a/kernel/src/thread/mod.rs
+++ b/kernel/src/thread/mod.rs
@@ -55,18 +55,10 @@ fn post_schedule_handler() {
     }
 }
 
-fn pre_user_run_handler(guard: &DisabledLocalIrqGuard) {
-    let task = Task::current().unwrap();
-    let thread_local = task.as_thread_local().unwrap();
-
-    thread_local.supp_user_context().before_user_exec(guard);
-}
-
 pub(super) fn init() {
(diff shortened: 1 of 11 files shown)


tatetian (approved):

PR #3326, opened 2026-06-04
Report real RSP before the trap in `TrapFrame`

lrh2000:
My mistake. I should submit this as https://github.com/asterinas/asterinas/pull/3323 instead. Otherwise, the code keeps changing back and forth....

Although https://github.com/asterinas/asterinas/pull/3323 correctly fixes the soundness issues, it highlights another issue that I wasn't aware of until now.

The CPU aligns the stack pointer to a 16-byte boundary upon an exception/interrupt. The existing method of getting the RSP value is incorrect. We should rely on what the CPU pushes onto the stack.

From the Intel docs:
> 64-bit interrupt service routines that exit with an IRET unconditionally pop SS:RSP off of the interrupt stack frame, even if the target code segment is running in 64-bit mode or at CPL = 0. This is because the original interrupt always pushes SS:RSP.

The new code can correctly obtain the RSP value **before** the trap occurs. The existing code cannot do so because it has no way of learning the RSP value **before the CPU aligns it**.

I have verified the code works correct using GDB:
```
ostd::arch::irq::ops::enable_local_and_halt () at src/arch/x86/irq/ops.rs:38
⚠️ warning: 38	src/arch/x86/irq/ops.rs: No such file or directory
(gdb) p /x $rsp
$1 = 0xffffdfffeffdfa28
(gdb) b trap_handler
Breakpoint 1 at 0xffffffff88b1686c: file src/arch/x86/trap/mod.rs, line 159.
(gdb) c
Continuing.

Breakpoint 1, ostd::arch::trap::trap_handler (f=0xffffdfffeffdf970) at src/arch/x86/trap/mod.rs:159
⚠️ warning: 159	src/arch/x86/trap/mod.rs: No such file or directory
(gdb) p /x f.rsp
$2 = 0xffffdfffeffdfa28
(gdb) p /x f.ss
$3 = 0x10
```

As you can see,
 - `f.rsp` now correctly reports the RSP value at `enable_local_and_halt`;
 - `f.ss` is also the correct kernel SS.

### DIFF
--- a/ostd/src/arch/x86/cpu/context/mod.rs
+++ b/ostd/src/arch/x86/cpu/context/mod.rs
@@ -371,7 +371,6 @@ impl UserContextApiInternal for UserContext {
             rsi: self.user_context.general.rsi,
             rdi: self.user_context.general.rdi,
             rbp: self.user_context.general.rbp,
-            rsp: self.user_context.general.rsp,
             r8: self.user_context.general.r8,
             r9: self.user_context.general.r9,
             r10: self.user_context.general.r10,
@@ -380,12 +379,13 @@ impl UserContextApiInternal for UserContext {
             r13: self.user_context.general.r13,
             r14: self.user_context.general.r14,
             r15: self.user_context.general.r15,
-            _pad: 0,
             trap_num: self.user_context.trap_num,
             error_code: self.user_context.error_code,
             rip: self.user_context.general.rip,
             cs: 0,
             rflags: self.user_context.general.rflags,
+            rsp: self.user_context.general.rsp,
+            ss: 0,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


junyang-zh:
Just to ensure that we are on the same page... From your description, it seems that the existing code only works when interrupts happen when `rsp` is aligned? Is that the reason why it currently just works?

lrh2000:
> From your description, it seems that the existing code only works when interrupts happen when `rsp` is aligned?

The existing code works perfectly well except for the incorrect value in `TrapFrame::rsp`.

`TrapFrame::rsp` is a dead field that no one will access. Therefore, having an incorrect value does not have a real impact.
 - Except: There is one scenario in which the kernel may log the entire `TrapFrame` on the console when it cannot handle the exception. The logged RSP will be incorrect.
   https://github.com/asterinas/asterinas/blob/356dd6e71bc6f980f7ebb883728f2e912553bb12/ostd/src/arch/x86/trap/mod.rs#L193-L196

This PR corrects the value of the `TrapFrame::rsp` field. Again, however, no one accesses the field. I confirm the fix via GDB, as you can see from the PR description.

junyang-zh (approved):
LGTM! Thanks for the fix.

PR #3327, opened 2026-06-04
Add inode cache for virtio-fs

li041:
## Overview
Add an inode cache for virtio-fs, keyed by `(FuseNodeId, FuseGeneration)`, so multiple directory entries or hard links that refer to the same FUSE inode share one `VirtioFsInode` instance.

  ## Motivation

 Before this change, each successful lookup constructed a fresh `VirtioFsInode`, even when the FUSE server returned the samenode ID and generation. This meant hard links to the same file could be represented by different in-memory inode objects.

  ## Changes
  - Add `InodeCache` for virtio-fs, storing weak references by FUSE node ID and generation.
  - Route lookup replies through the inode cache and reuse an existing live inode when possible. Refresh cached inode metadata, entry TTL, and lookup references when a cached inode is reused.
  - Insert newly created virtio-fs inodes into the cache.
  - Remove cached inodes after `unlink`/`rmdir` when the local hard-link count reaches zero.

### DIFF
--- a/kernel/src/fs/fs_impls/virtiofs/fs.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/fs.rs
@@ -118,11 +118,11 @@ impl VirtioFs {
     /// Reads an inode from a FUSE entry reply via the inode cache.
     pub(super) fn lookup_inode_from_cache(
         self: &Arc<Self>,
-        entry_reply: EntryReply,
+        lookup_reply: EntryReply,
         request_attr_version: AttrVersion,
     ) -> Result<Arc<VirtioFsInode>> {
         self.inode_cache
-            .lookup_inode(entry_reply, request_attr_version, self)
+            .lookup_inode(lookup_reply, request_attr_version, self)
     }
 
     /// Inserts a newly created inode into the inode cache.
--- a/kernel/src/fs/fs_impls/virtiofs/fs.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/fs.rs
@@ -4,13 +4,13 @@
 //!
 //! This module defines the filesystem type and objects for virtio-fs.
 
-use core::time::Duration;
-
-use aster_fuse::{FUSE_ROOT_ID, Kstatfs, StatfsOperation, ops::lookup::LookupOperation};
-use aster_virtio::device::filesystem::device::{self, FileSystemDevice, FuseSession};
+use aster_fuse::{
+    EntryReply, FUSE_ROOT_ID, Kstatfs, StatfsOperation, ops::lookup::LookupOperation,
+};
+use aster_virtio::device::filesystem::device::{self, AttrVersion, FileSystemDevice, FuseSession};
 use device_id::DeviceId;
 
-use super::{inode::VirtioFsInode, valid_until};
+use super::inode::{InodeCache, VirtioFsInode};
 use crate::{
     fs::{
         pseudofs::AnonDeviceId,
@@ -65,6 +65,7 @@ pub(super) struct VirtioFs {
     root: Arc<VirtioFsInode>,
     tag: String,
     session: Arc<FuseSession>,
+    inode_cache: InodeCache,
     fs_event_subscriber_stats: FsEventSubscriberStats,
 }
 
@@ -84,25 +85,22 @@ impl VirtioFs {
         let sb = SuperBlock::from((container_dev_id, statfs));
 
         let root_entry = session.do_fuse_op(FUSE_ROOT_ID, LookupOperation::new("."))?;
-        let root_metadata = super::inode::metadata_from_attr(root_entry.attr(), container_dev_id);
-        let attr_valid_until = valid_until(root_entry.attr_valid(), root_entry.attr_valid_nsec());
 
         Ok(Arc::new_cyclic(|weak_fs| {
-            let root = VirtioFsInode::new(
-                FUSE_ROOT_ID,
-                root_entry.generation(),
-                root_metadata,
+            let root = VirtioFsInode::new_root(
+                root_entry,
                 weak_fs.clone(),
-                Duration::MAX,
-                attr_valid_until,
+                container_dev_id,
                 session.bump_attr_version(),
             );
+            let inode_cache = InodeCache::new(&root);
 
             Self {
                 sb,
                 root,
                 tag,
                 session,
+                inode_cache,
                 fs_event_subscriber_stats: FsEventSubscriberStats::new(),
             }
         }))
@@ -110,6 +108,34 @@ impl VirtioFs {
 
     pub(super) fn session(&self) -> &Arc<FuseSession> {
         &self.session
+    }
+
+    /// Returns the device ID of this virtio-fs mount.
+    pub(super) fn container_device_id(&self) -> DeviceId {
+        self.sb.container_dev_id
+    }
+
+    /// Reads an inode from a FUSE entry reply via the inode cache.
+    pub(super) fn lookup_inode_from_cache(
+        self: &Arc<Self>,
+        entry_reply: EntryReply,
+        request_attr_version: AttrVersion,
+    ) -> Result<Arc<VirtioFsInode>> {
+        self.inode_cache
+            .lookup_inode(entry_reply, request_attr_version, self)
+    }
+
+    /// Inserts a newly created inode into the inode cache.
+    pub(super) fn insert_inode_to_cache(&self, inode: &Arc<VirtioFsInode>) {
+        self.inode_cache.insert_inode(inode);
+    }
+
+    /// Removes an inode from the cache if it is still the cached entry.
+    pub(super) fn remove_inode_from_cache(
+        &self,
+        inode: &VirtioFsInode,
+    ) -> Option<Weak<VirtioFsInode>> {
+        self.inode_cache.remove_inode(inode)
     }
 }
 
--- a/kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
@@ -23,11 +23,11 @@ impl InodeCache {
     /// Looks up an inode by FUSE node ID and validates its generation.
     pub(in crate::fs::fs_impls::virtiofs) fn lookup_inode(
         &self,
-        entry_reply: EntryReply,
+        lookup_reply: EntryReply,
         request_attr_version: AttrVersion,
         fs: &Arc<VirtioFs>,
     ) -> Result<Arc<VirtioFsInode>> {
-        let nodeid = entry_reply.nodeid();
+        let nodeid = lookup_reply.nodeid();
 
         // An `EntryReply` carries a new lookup reference even when the inode is
         // already cached, so cache hits still need to commit the reply.
@@ -36,14 +36,14 @@ impl InodeCache {
             .read()
             .get(&nodeid)
             .and_then(Weak::upgrade)
-            .filter(|inode| inode.generation() == entry_reply.generation())
+            .filter(|inode| inode.generation() == lookup_reply.generation())
         {
             // Reusing a cached inode can still fail while committing the
             // returned attributes, because a size or mtime change may require
             // page-cache resize or invalidation. There is no local recovery
             // path for those page-cache errors, so propagate them to the
             // lookup caller.
-            inode.update_from_entry_reply(&entry_reply, request_attr_version)?;
+            inode.update_from_entry_reply(&lookup_reply, request_attr_version)?;
             return Ok(inode);
         }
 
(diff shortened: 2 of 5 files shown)


kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
```diff
@@ -0,0 +1,101 @@
...
+            valid_until(entry_reply.attr_valid(), entry_reply.attr_valid_nsec()),
+            fs.session().bump_attr_version(),
+        );
+        inode_cache.insert(key, Arc::downgrade(&inode));
+
+        Ok(inode)
+    }
+
+    /// Inserts an inode by FUSE node ID and generation.
+    pub(in crate::fs::fs_impls::virtiofs) fn insert_inode(&self, inode: Arc<VirtioFsInode>) {
```
StevenJiang1110: There's no need to pass `Arc<VirtioFsInode>` here, using `&Arc<VirtioFsInode>` should be enough.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -186,6 +223,10 @@ impl InodeInner {
     fn page_cache(&self) -> Option<&PageCache> {
         self.page_cache.as_ref()
     }
+
+    fn dec_link_count(&mut self, delta: usize) {
+        self.metadata.nr_hard_links = self.metadata.nr_hard_links.saturating_sub(delta);
```
StevenJiang1110: Will overflow occur here? If not, I think we should not use `saturating_sub`. Directly using `self.metadata.nr_hard_links -= delta`.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -373,15 +410,41 @@ impl Inode for VirtioFsInode {
...
     fn rmdir(&self, name: &str) -> Result<()> {
         let fs = self.fs_ref();
+        let child = self.lookup_child_inode(name)?;
+
         fs.session()
             .do_fuse_op(self.nodeid(), RmdirOperation::new(name))?;
+
+        {
+            let mut child_inner = child.inner.write();
+            child_inner.dec_link_count(2);
```
StevenJiang1110: Please add a comment to explain the magic number 2.

To be honest, I am not sure why we need to update the link count here. On the next line, we have already invalidated the local attribute, so the next time we access this inode, we should be fetching the hard link count from the remote anyway, shouldn't we?

kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
```diff
@@ -0,0 +1,101 @@
...
+            .and_then(|inode| inode.upgrade())
+    }
+}
+
+/// Key for identifying a FUSE inode generation in the inode cache.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub(in crate::fs::fs_impls::virtiofs) struct InodeCacheKey {
+    nodeid: FuseNodeId,
+    generation: FuseGeneration,
+}
```
StevenJiang1110: I think using `nodeid + generation` as the key is problematic. In the FUSE protocol, `nodeid` is meant to uniquely identify an inode. If two inodes share the same nodeid but have different generation values, the one with the older generation should be considered stale and evicted from the inode cache. However, with the current design, the inode cache can end up containing multiple inodes with the same `nodeid` but different generation values.

Therefore, I think `nodeid` alone should be used as the key, but during lookup, the generation should also be validated to detect stale entries and ensure that the inode cache always holds the most up-to-date inode.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -373,15 +410,41 @@ impl Inode for VirtioFsInode {
...
+
+        {
+            let mut child_inner = child.inner.write();
+            child_inner.dec_link_count(1);
+            child_inner.attr_valid_until = Duration::ZERO;
+        }
+
+        if child.metadata().nr_hard_links == 0 {
+            let _ = fs.remove_inode(child.cache_key());
+        }
```
StevenJiang1110: I think the logic for removing an inode from the inode cache can be moved to the `Drop` implementation of `VirtioFsInode`. Additionally, relying on `nr_hard_links` here is not reliable, as its value may be stale.

```rust
impl Drop for VirtioFsInode {
     fn drop(&mut self) {
          let fs = self.fs_ref();
          let _ = fs.remove_inode(self.cache_key());
     } 
}
```

kernel/src/fs/fs_impls/virtiofs/fs.rs
```diff
@@ -111,6 +120,31 @@ impl VirtioFs {
...
         &self.session
     }
+
+    /// Returns the device ID of this virtio-fs mount.
+    pub(super) fn container_device_id(&self) -> DeviceId {
+        self.sb.container_dev_id
+    }
+
+    /// Reads an inode from a FUSE entry reply via the inode cache.
+    pub(super) fn read_inode(
```
StevenJiang1110: `read_inode` is a strange name. You can use `lookup_inode`.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -373,15 +410,41 @@ impl Inode for VirtioFsInode {
...
+
         fs.session()
             .do_fuse_op(self.nodeid(), RmdirOperation::new(name))?;
+
+        {
+            let mut child_inner = child.inner.write();
+            child_inner.dec_link_count(2);
+            child_inner.attr_valid_until = Duration::ZERO;
+        }
+
```
StevenJiang1110: ext2 modifies the parent's `link_count`, but here you are only invalidating `attr_valid_until` for the child without doing the same for the parent. I think you need to mark the parent's attributes as stale as well. The same applies to `unlink` — the parent's attributes should also be marked as stale there.

kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
```diff
@@ -0,0 +1,101 @@
...
+        entry_reply: EntryReply,
+        request_attr_version: AttrVersion,
+        fs: &Arc<super::super::fs::VirtioFs>,
+    ) -> Result<Arc<VirtioFsInode>> {
+        let key = InodeCacheKey::new(entry_reply.nodeid(), entry_reply.generation());
+
+        // An `EntryReply` carries a new lookup reference even when the inode is
+        // already cached, so cache hits still need to commit the reply.
+        if let Some(inode) = self.inodes.read().get(&key).and_then(Weak::upgrade) {
+            inode.reuse_from_entry_reply(&entry_reply, request_attr_version)?;
```
StevenJiang1110: Under what circumstances would an inode exist in the inode cache yet a `reuse_from_entry_reply` still fail? 

At least for now, it seems the only errors  come from page cache operations. I think you can add some comment to explain about it. We cannot handle such page cache errors, so just return error should be accepted here.

kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
```diff
@@ -0,0 +1,101 @@
...
+        let mut inode_cache = self.inodes.write();
+        if let Some(inode) = inode_cache.get(&key).and_then(Weak::upgrade) {
+            drop(inode_cache);
+            inode.reuse_from_entry_reply(&entry_reply, request_attr_version)?;
+            return Ok(inode);
+        }
+
+        // Drop stale weak entries before installing a new live inode for the
+        // same FUSE generation.
+        inode_cache.remove(&key);
```
StevenJiang1110: The `remove` is useless? The below `insert` will remove the stale entry anyway.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -123,10 +125,45 @@ impl VirtioFsInode {
...
         self.size.load(Ordering::Acquire)
     }
 
+    /// Reuses a cached inode for a fresh FUSE entry reply.
+    ///
+    /// A successful entry reply gives the client one additional lookup
+    /// reference and a refreshed directory-entry TTL. Even though the inode
+    /// object already exists, the cached lookup count and metadata still need
+    /// to observe the reply before the inode is returned to VFS.
+    pub(super) fn reuse_from_entry_reply(
```
StevenJiang1110: This method may be better renamed to `update_from_entry_reply`. This method is to update the lookup reference count and TTL for the inode.

kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
```diff
@@ -0,0 +1,101 @@
...
+        )]));
+        Self { inodes }
+    }
+
+    /// Looks up an inode by FUSE node ID and generation.
+    pub(in crate::fs::fs_impls::virtiofs) fn lookup_inode(
+        &self,
+        entry_reply: EntryReply,
+        request_attr_version: AttrVersion,
+        fs: &Arc<super::super::fs::VirtioFs>,
```
StevenJiang1110: Avoid using `super::super` in the visibility marker.

kernel/src/fs/fs_impls/virtiofs/inode/cache.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub(in crate::fs::fs_impls::virtiofs) fn insert_inode(&self, inode: Arc<VirtioFsInode>) {
+        self.inodes
+            .write()
+            .insert(inode.cache_key(), Arc::downgrade(&inode));
+    }
+
+    pub(in crate::fs::fs_impls::virtiofs) fn remove_inode(
+        &self,
+        key: InodeCacheKey,
+    ) -> Option<Arc<VirtioFsInode>> {
```
StevenJiang1110: I think this method should return `Option<Weak<VirtioFsInode>>` instead. Since this is a remove operation, there is a good chance the caller does not actually need the return value, in which case the `upgrade` would be an unnecessary operation. If the caller does need the upgraded value, it can perform the `upgrade` itself.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -123,10 +125,45 @@ impl VirtioFsInode {
...
+        )?;
+        *self.entry_valid_until.lock() =
+            valid_until(entry_reply.entry_valid(), entry_reply.entry_valid_nsec());
+        Ok(())
+    }
+
+    fn lookup_child_inode(&self, name: &str) -> Result<Arc<VirtioFsInode>> {
+        let fs = self.fs_ref();
+        let request_attr_version = fs.session().snapshot_attr_version();
+        let entry_reply = fs
```
StevenJiang1110: `entry_reply` is not very clear. I think renaming it to `lookup_reply` or `lookup_res` would make the code easier to follow. At least within the kernel crate, I think either of these names would be an improvement over `entry_reply`.

That said, `entry_reply` is used extensively throughout the current virtio-fs code, so this change may be out of scope for this PR. You can consider addressing it in a follow-up PR.

StevenJiang1110 (commented):
Thanks for your contribution!

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
li041: `entry_reply` can be the payload of `lookup`, `create` and `mknod`. I will change these names according to the context, e.g. `xxx_reply`.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -348,7 +365,19 @@ impl Inode for VirtioFsInode {
...
+
+        let child = VirtioFsInode::new(
+            create_reply.nodeid(),
+            create_reply.generation(),
+            metadata_from_attr(create_reply.attr(), fs.container_device_id()),
+            Arc::downgrade(&fs),
+            valid_until(create_reply.entry_valid(), create_reply.entry_valid_nsec()),
+            valid_until(create_reply.attr_valid(), create_reply.attr_valid_nsec()),
+            fs.session().bump_attr_version(),
+        );
```
StevenJiang1110: It looks like this pattern is repeated several times throughout the current code. There seem to be only two cases where a `VirtioFsInode` is constructed: one for creating a root inode, and one for creating a new inode from a reply. I think we should add two helper methods to `VirtioFsInode` and make the original constructor a private internal method:

```rust
impl VirtioFsInode {
    pub(in ...) new_root(fs: &Weak<VirtioFs>) -> Self;
    pub(in ...) new_from_entry_reply(entry_reply: EntryReply, fs: &Weak<VirtioFs>) -> Self { 
         // ....
    }
    fn new(...) -> Self {}
}
```

kernel/src/fs/fs_impls/virtiofs/fs.rs
```diff
@@ -111,6 +120,31 @@ impl VirtioFs {
...
         &self.session
     }
+
+    /// Returns the device ID of this virtio-fs mount.
+    pub(super) fn container_device_id(&self) -> DeviceId {
+        self.sb.container_dev_id
+    }
+
+    /// Reads an inode from a FUSE entry reply via the inode cache.
+    pub(super) fn lookup_inode(
```
StevenJiang1110: Let's add some suffix to the three methods to make it more clear. 

```rust
lookup_inode_from_cache
insert_inode_to_cache
remove_inode_to_cache
```

StevenJiang1110 (commented):
Thanks for your revision! Just two small nits.

StevenJiang1110 (approved):

PR #3328, opened 2026-06-04
Introduce a cache mechanism for `FsType`

cchanging:
Currently, every time we mount a device, we create a new file system instance. However, having multiple file system instances for the same device means each instance has its own independent locks, and these instances could concurrently modify the contents of the device.

This PR addresses the above situation. With this PR, when the same device is mounted multiple times in different locations, these mount instances can actually share the same file system instance, which is aligned with Linux.

### DIFF
--- a/kernel/src/fs/fs_impls/ext2/fs_type.rs
+++ b/kernel/src/fs/fs_impls/ext2/fs_type.rs
@@ -36,8 +36,9 @@ impl FsType for Ext2Type {
 
     fn create(&self, fs_creation_ctx: &FsCreationCtx) -> Result<Arc<dyn FileSystem>> {
         let disk = fs_creation_ctx.resolve_block_device()?.clone();
+        let flags = fs_creation_ctx.flags();
         let args = fs_creation_ctx.args();
-        Ext2::open(disk, args).map(|fs| fs as Arc<dyn FileSystem>)
+        Ext2::open(disk, flags, args).map(|fs| fs as Arc<dyn FileSystem>)
     }
 
     fn obtain_key_and_cache(
--- a/kernel/src/fs/fs_impls/ext2/fs_type.rs
+++ b/kernel/src/fs/fs_impls/ext2/fs_type.rs
@@ -6,17 +6,26 @@
 //! discover and mount ext2 volumes by name (`"ext2"`).
 
 use aster_systree::SysNode;
+use device_id::DeviceId;
 
 use super::{fs::Ext2, prelude::*};
 use crate::fs::vfs::{
     file_system::FileSystem,
-    registry::{FsCreationCtx, FsProperties, FsType},
+    registry::{FsCache, FsCreationCtx, FsProperties, FsType},
 };
 
 /// VFS-visible Ext2 filesystem type.
-pub(super) struct Ext2Type;
+pub(super) struct Ext2Type {
+    cache: FsCache<DeviceId>,
+}
+
+pub(super) static EXT2_TYPE: Ext2Type = Ext2Type {
+    cache: FsCache::new(),
+};
 
 impl FsType for Ext2Type {
+    type Key = DeviceId;
+
     fn name(&self) -> &'static str {
         "ext2"
     }
@@ -26,9 +35,21 @@ impl FsType for Ext2Type {
     }
 
     fn create(&self, fs_creation_ctx: &FsCreationCtx) -> Result<Arc<dyn FileSystem>> {
-        let disk = fs_creation_ctx.resolve_block_device()?;
+        let disk = fs_creation_ctx.resolve_block_device()?.clone();
         let args = fs_creation_ctx.args();
         Ext2::open(disk, args).map(|fs| fs as Arc<dyn FileSystem>)
+    }
+
+    fn obtain_key_and_cache(
+        &self,
+        fs_creation_ctx: &FsCreationCtx,
+    ) -> Option<(DeviceId, &FsCache<DeviceId>)> {
+        let key = fs_creation_ctx
+            .resolve_block_device()
+            .ok()
+            .map(|disk| disk.id())?;
+
+        Some((key, &self.cache))
     }
 
     fn sysnode(&self) -> Option<Arc<dyn SysNode>> {
--- a/kernel/src/fs/vfs/fs_apis/registry.rs
+++ b/kernel/src/fs/vfs/fs_apis/registry.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use alloc::collections::btree_map::Entry;
+
 use aster_block::BlockDevice;
 use aster_systree::{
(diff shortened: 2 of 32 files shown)


kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -161,8 +254,68 @@ pub fn init() {
...
+            if let Some((fs, root_dentry)) = entries
+                .get(&key)
+                .and_then(|cached| cached.0.upgrade().zip(cached.1.upgrade()))
+            {
+                return Ok(FsAndRoot { fs, root_dentry });
+            }
+        }
+
+        let fs = create_fn()?;
+        let fs_and_root = FsAndRoot::new(fs);
```
StevenJiang1110: The current approach of releasing the lock before invoking `create_fn` may introduce correctness issues. Specifically, it can lead to multiple instances of `create_fn` executing concurrently, which may be problematic for certain filesystems. For example, in the case of virtio-fs, this could result in multiple `FUSE_INIT` requests being sent simultaneously. Therefore, we should ensure that at most one instance of `create_fn` is executing at any given time.

The simplest fix is to hold the lock on `entries` while calling `create_fn`. Alternatively, if you prefer to release the lock before invoking it, we would need to track whether a `create_fn` is currently in progress to prevent concurrent executions.

test/initramfs/src/regression/fs/run_test.sh
```diff
@@ -99,6 +99,7 @@ test_ext2 "/ext2" "test_file.txt"
 ./ext2/rmdir
 ./ext2/short_rw
 ./ext2/sparse
+./ext2/shared_block_device
```
StevenJiang1110: Move it before `./ext2/short_rw`.

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -161,8 +254,68 @@ pub fn init() {
...
+        let mut entries = self.entries.lock();
+        // Recheck whether the entry exists.
+        if let Some((fs, root_dentry)) = entries
+            .get(&key)
+            .and_then(|cached| cached.0.upgrade().zip(cached.1.upgrade()))
+        {
+            return Ok(FsAndRoot { fs, root_dentry });
+        }
+
+        entries.insert(
```
StevenJiang1110: line 295 can be moved to just before line 306?

kernel/src/fs/fs_impls/ext2/fs_type.rs
```diff
@@ -31,6 +44,17 @@ impl FsType for Ext2Type {
         Ext2::open(disk, args).map(|fs| fs as Arc<dyn FileSystem>)
     }
 
+    fn obtain_key(&self, fs_creation_ctx: &FsCreationCtx) -> Option<DeviceId> {
+        fs_creation_ctx
+            .resolve_block_device()
+            .ok()
+            .map(|disk| disk.id())
+    }
```
StevenJiang1110: For an ext2 filesystem that has not yet been cached, we first obtain the key to look up the cache, and then invoke create to instantiate it. This means `resolve_block_device` ends up being called twice. Given that this operation involves a path lookup and is therefore not particularly cheap, is there any way to avoid the redundant call?

StevenJiang1110 (commented):
Thanks for your contribution!

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -28,6 +66,20 @@ pub trait FsType: Send + Sync + 'static {
...
+    fn obtain_key(&self, _fs_creation_ctx: &FsCreationCtx) -> Option<Self::Key> {
+        None
+    }
+
+    /// Returns this type's per-instance cache.
+    ///
+    /// Must be `Some` whenever `obtain_key` can return `Some`.
+    fn cache(&self) -> Option<&FsCache<Self::Key>> {
+        None
+    }
```
StevenJiang1110: It looks like we could merge `obtain_key` and `cache` into a single method, eliminating the need for trait implementers to maintain the invariant "must be `Some` whenever `obtain_key` returns `Some`."

```rust
fn cache(&self, _fs_creation_ctx: &FsCreationCtx) -> Option<&FsCache<Self::Key>>;
```

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -82,6 +181,10 @@ impl<'a> FsCreationCtx<'a> {
 
     /// Resolves the mount source into a block device.
     pub(in crate::fs) fn resolve_block_device(&self) -> Result<Arc<dyn BlockDevice>> {
```
StevenJiang1110: Perhaps we can return `Result<&Arc<dyn BlockDevice>>` now?

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -161,8 +266,79 @@ pub fn init() {
...
+    /// `create_fn` and derives a root `Dentry`.
+    fn get_or_create<F>(&self, key: K, required_flags: FsFlags, create_fn: F) -> Result<FsAndRoot>
+    where
+        F: FnOnce() -> Result<Arc<dyn FileSystem>>,
+    {
+        let mut entries = self.entries.lock();
+        match entries.entry(key) {
+            Entry::Occupied(mut entry) => {
+                if let Some((fs, root_dentry)) =
+                    entry.get().0.upgrade().zip(entry.get().1.upgrade())
```
StevenJiang1110: In this part of the code, if either `fs` or `root_dentry` fails to upgrade, you call `create_fn` to create a new filesystem. But that doesn't seem necessary — as long as `fs` upgrades successfully, even if `root_dentry` fails to upgrade, you could construct a new `root_dentr`y from the existing fs, rather than calling `create_fn` to create an entirely new filesystem.

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -161,8 +266,79 @@ pub fn init() {
...
 
+/// A cache of file system instances, keyed by `K`.
+///
+/// `FsType` implementations use this cache to deduplicate mounts that refer to
+/// the same logical file system instance, such as the same block device.
+pub struct FsCache<K: Eq + Ord + Clone + Send + Sync + 'static> {
+    entries: Mutex<BTreeMap<K, CachedEntry>>,
+}
+
+type CachedEntry = (Weak<dyn FileSystem>, Weak<Dentry>);
```
StevenJiang1110: Do you think it would be reasonable to define it in the following form instead:

```rust
struct WeakFsAndRoot {
    fs: Weak<dyn FileSystem>, 
    root_dentry: Weak<Dentry>
}
```

This way, we could implement a `downgrade` method on `FsAndRoot`, which better reflects the relationship between the two types — I feel this is slightly cleaner. However, it's up to you; feel free to keep it as is if you prefer.

StevenJiang1110 (approved):
LGTM! Thanks for your contribution! Only a few minor comments.

cqs21 (approved):

PR #3329, opened 2026-06-04
Check for address overflow in CursorMut::unmap

### LINKED ISSUE #3159: Can `CursorMut::unmap` silently become a no-op in release builds on overflow?
I may be missing an intended invariant here, but I wanted to check whether this is considered a bug.

`CursorMut::unmap` appears to compute the end of the range with unchecked addition:
https://github.com/asterinas/asterinas/blob/c1e6cfe448508b93aef4e2d6bcdbc644e102b719/ostd/src/mm/vm_space.rs#L451-L457

From the comments, `CursorMut::unmap` is expected to panic if:
- `len` is longer than the remaining range of the cursor
- `len` is not page-aligned

My question is whether the first guarantee still holds in release builds.

If `len` is extremely large, then:

1. `self.virt_addr() + len` may wrap
2. `self.va + len` in the lower layer may also wrap
3. `assert!(end <= self.barrier_va.end)` may then pass unexpectedly

mvanhorn:
## Summary
`CursorMut::unmap` now computes its end address with `checked_add`, so a `len` large enough to wrap `usize` panics with a clear message instead of silently wrapping. The documented panic contract ("panics if the length is longer than the remaining range of the cursor") previously did not hold in release builds: `self.virt_addr() + len` wrapped, the lower layer's `assert!(end <= self.barrier_va.end)` could then pass unexpectedly, and the unmap became a no-op or operated on a wrong range.

## Why this matters
#3159 traced this exact path (`ostd/src/mm/vm_space.rs`), and junyang-zh confirmed "the wrapping behavior is unintended. We should check it." The unmap path is memory-safety adjacent: an unmap that silently does nothing while the caller believes the range is gone undermines the TLB-flush contract described in the method docs.

This change is scoped to the `unmap` entry path per the issue; the sibling sites Marsman1996 listed (`page_table/cursor/mod.rs`, `kernel/src/vm/vmo/mod.rs`) have the same pattern but separate call contracts, so they are left for follow-ups.

## Testing
- New `#[ktest]` `vmspace_unmap_overflowing_len` locks the panic in: a page-aligned `len` of `usize::MAX & !(PAGE_SIZE - 1)` added to the cursor's address wraps, and the test expects the "the unmap length overflows the address space" panic
- `cargo check -p ostd --target x86_64-unknown-none` (with and without `--cfg ktest`) and `cargo fmt -p ostd -- --check` pass on toolchain nightly-2026-04-03
- Valid page-aligned `len` within range takes the same path as before (`checked_add` returns `Some`), so existing `vmspace_map_unmap` / `vmspace_unmap_twice` behavior is unchanged

Fixes #3159

### DIFF
--- a/ostd/src/mm/test.rs
+++ b/ostd/src/mm/test.rs
@@ -6,8 +6,8 @@ use crate::{
     Error,
     io::IoMem,
     mm::{
-        CachePolicy, FallibleVmRead, FallibleVmWrite, FrameAllocOptions, PageFlags, PageProperty,
-        UFrame, VmSpace,
+        CachePolicy, FallibleVmRead, FallibleVmWrite, FrameAllocOptions, PAGE_SIZE, PageFlags,
+        PageProperty, UFrame, VmSpace,
         io::{VmIo, VmIoFill, VmReader, VmWriter, util::HasVmReaderWriter},
         tlb::TlbFlushOp,
         vm_space::{VmQueriedItem, get_activated_vm_space},
@@ -608,6 +608,25 @@ mod vmspace {
                 .expect("failed to create the mutable cursor");
             cursor_mut.map(frame.clone(), prop);
         }
+    }
+
+    /// Ensures `unmap` panics when `len` overflows the address space
+    /// instead of silently wrapping (issue #3159).
+    #[ktest]
+    #[should_panic(expected = "the unmap length overflows the address space")]
+    fn vmspace_unmap_overflowing_len() {
+        let vmspace = VmSpace::default();
+        let range = 0x1000..0x2000;
+        let preempt_guard = disable_preempt();
+
+        let mut cursor_mut = vmspace
+            .cursor_mut(&preempt_guard, &range)
+            .expect("failed to create the mutable cursor");
+        // Page-aligned `len` that wraps past `usize::MAX` when added to the
+        // cursor's virtual address. Without the checked addition this used
+        // to wrap and bypass the documented out-of-range panic.
+        let overflowing_len = usize::MAX & !(PAGE_SIZE - 1);
+        cursor_mut.unmap(overflowing_len);
     }
 
     /// Unmaps twice using `CursorMut`.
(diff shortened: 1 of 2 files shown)


lrh2000:
I believe this is fully automated, given (1) the overly verbose PR description for such a small change and (2) GitHub profile.

I propose to just close this PR, regardless whether the code is correct or not. I don't think we should welcome fully automated contributions. Sorry.

1. 
image

2.
image

mvanhorn:
Sad face.  I use AI in my workflow, but picked this problem to solve and reviewed and submitted myself.  Would love a real look, if it's low quality kill it..

lrh2000:
I expect someone working on this problem to carefully review the code within a single module (e.g., the page table module) and fix similar issues. Fixing one issue without understanding the surrounding code is not very valuable. **Especially, this integer overflow is not reachable from user space.**

Moreover, I do not mean for you to use AI to finish this task. **As a first-time contributor, I expect you to learn about the code.** However, since you have dozens of other projects and hundreds of other PRs to work on, I don't think you'll have much time to learn this project's codebase.

I'd like to quote [rust-lang's words](https://redirect.github.com/rust-lang/rust/pull/157369#issuecomment-4611079917), which explain my position very well:
> We are a community of contributors, not just a code repository. Thus we focus on contributors who desire to stay around and put in the work to produce high quality contributions or learn to do so.

The bottom line is: if you would still like to contribute, please open a new PR and **write the description by hand** (based on your understanding of the code, not based on AI's analysis). However, this is the minimum requirement. For the aforementioned reasons, I suggest you stop working on this problem unless you (yourself, not an AI) want to learn.

Thank you for your understanding and interest in contributing, though.

mvanhorn:
That's fair, and I appreciate you explaining the reasoning instead of just closing without comment. Point taken that the valuable unit of work here is a module-level review, not a single call site. If I pick asterinas up again it'll be on your terms, hand-written and after actually learning the page table code.

PR #3334, opened 2026-06-05
Add rename support for virtio-fs

li041:
## Summary

This PR implements `rename` support for virtio-fs.

It adds the `FUSE_RENAME` request type to `aster-fuse`, wires it into the virtio-fs inode implementation, and sends the rename operation to the FUSE server with the source name, target parent node ID, and target name.

The PR also updates VFS dentry rename handling to invalidate cached dentries when a remote filesystem reports `ENOENT`, so stale cached entries are removed for revalidation-based filesystems.

  ## Changes

  - Add `RenameReq` and `RenameOperation` for `FUSE_RENAME`.
  - Implement `Inode::rename` for `VirtioFsInode`.
  - Expire virtio-fs attribute caches after successful rename.
  - Remove stale cached dentries when rename fails with `ENOENT` under revalidation policy.

### DIFF
--- a/kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
@@ -22,6 +22,7 @@ use aster_fuse::{
         mknod::{MknodOperation, MknodReq},
         open::{OpenReq, OpendirOperation},
         release::ReleaseOptions,
+        rename::{RenameOperation, RenameReq},
         rmdir::RmdirOperation,
         unlink::UnlinkOperation,
     },
@@ -445,6 +446,27 @@ impl Inode for VirtioFsInode {
 
         self.expire_attr_cache();
         child.expire_attr_cache();
+
+        Ok(())
+    }
+
+    fn rename(&self, old_name: &str, target: &Arc<dyn Inode>, new_name: &str) -> Result<()> {
+        let target = target.downcast_ref::<VirtioFsInode>().unwrap();
+
+        let fs = self.fs_ref();
+
+        // TODO: Invalidate the stale positive dentry for `old_name`
+        // when encountering `ENOENT` once the rename interface can
+        // pass the cached old dentry down.
+        fs.session().do_fuse_op(
+            self.nodeid(),
+            RenameOperation::new(RenameReq::new(target.nodeid()), old_name, new_name),
+        )?;
+
+        self.expire_attr_cache();
+        if self.nodeid() != target.nodeid() {
+            target.expire_attr_cache();
+        }
 
         Ok(())
     }
--- a/kernel/libs/aster-fuse/src/lib.rs
+++ b/kernel/libs/aster-fuse/src/lib.rs
@@ -51,6 +51,7 @@ pub use self::{
         readdir::ReaddirOperation,
         readlink::{MAX_READLINK_LEN, ReadlinkOperation},
         release::{ReleaseFlags, ReleaseKind, ReleaseOperation, ReleaseReq},
+        rename::{RenameOperation, RenameReq},
         rmdir::RmdirOperation,
         setattr::{SetattrOperation, SetattrReq, SetattrValid},
         statfs::{Kstatfs, StatfsOperation, StatfsReply},
--- a/kernel/libs/aster-fuse/src/ops/mod.rs
+++ b/kernel/libs/aster-fuse/src/ops/mod.rs
@@ -28,6 +28,7 @@ pub mod read;
 pub mod readdir;
 pub mod readlink;
 pub mod release;
+pub mod rename;
 pub mod rmdir;
 pub mod setattr;
 pub mod statfs;
--- a/kernel/libs/aster-fuse/src/ops/rename.rs
+++ b/kernel/libs/aster-fuse/src/ops/rename.rs
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! `FUSE_RENAME` renames a directory entry.
+//!
+//! The request body contains [`RenameReq`] followed by the null-terminated old
+//! name and the null-terminated new name. Successful replies do not carry a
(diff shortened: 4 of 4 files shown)


kernel/src/fs/vfs/path/dentry.rs
```diff
@@ -692,7 +692,17 @@ impl DirDentry<'_> {
...
+                if err.error() == Errno::ENOENT
+                    && old_dir
+                        .revalidation_policy
+                        .contains(RevalidationPolicy::REVALIDATE_EXISTS)
+                {
+                    children.remove(old_name);
+                    children.remove(new_name);
+                }
+                return Err(err);
+            }
```
cchanging: Why add this rollback logic? We have locked the dentry cache here and obtained the `old_dentry`, why should we recheck the `ENOENT` error?

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -449,6 +450,29 @@ impl Inode for VirtioFsInode {
...
 
+    fn rename(&self, old_name: &str, target: &Arc<dyn Inode>, new_name: &str) -> Result<()> {
+        let target = target
+            .downcast_ref::<VirtioFsInode>()
+            .ok_or_else(|| Error::with_message(Errno::EXDEV, "not same fs"))?;
+
+        let fs = self.fs_ref();
+        if !Arc::ptr_eq(&fs, &target.fs_ref()) {
+            return_errno_with_message!(Errno::EXDEV, "not same fs");
+        }
```
cchanging: These checks have been done in VFS layer. I think we should have an intuition that all generic, per-filesystem checks ought to be done at the VFS layer.

Actually, #3118 is intended to try to address this issue, but I’m not sure whether you still have time to keep pushing this PR forward.

What I hope #3118 PR can do is:
1. Improve the documentation for each method in the `Inode` trait, making it clear what checks the VFS layer has already performed before the method is called;
2. Move all checks that can be moved to the VFS layer there.

If you don’t have time to keep working on it, please let me know, and I can help take care of it.

cchanging (commented):
Thanks for adding the `rename` functionality for `virtiofs`!

kernel/src/fs/vfs/path/dentry.rs
li041: This is not required for correctness if `REVALIDATE_EXISTS` is enabled, becausethe next cache hit will revalidate the positive dentry and drop it if the server has removed the entry.

 The reason for doing it here is that `rename` returning `ENOENT` is already a strong signal that the cached source entry is stale. Since the current dentry cache has no separate invalid/stale state, removing the entry is the way to invalidate it immediately and avoid keeping stale cache until a later lookup.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
li041: I may focus on the completeness of virtio-fs. I really appreciate if you can make VFS checks clean.

kernel/src/fs/vfs/path/dentry.rs
cchanging: > The reason for doing it here is that rename returning ENOENT is already a strong signal that the cached source entry is stale.

This situation would only occur with virtiofs, right? I don’t think behavior specific to a particular filesystem should be placed in the VFS layer. And as you said, virtiofs does perform `REVALIDATE_EXISTS`, so this early removal doesn’t seem strictly necessary either.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -449,6 +455,35 @@ impl Inode for VirtioFsInode {
         Ok(())
     }
 
+    fn rename(&self, old_name: &str, target: &Arc<dyn Inode>, new_name: &str) -> Result<()> {
+        let target = target
+            .downcast_ref::<VirtioFsInode>()
+            .ok_or_else(|| Error::with_message(Errno::EXDEV, "not same fs"))?;
```
cchanging: We can call `unwrap` here.

kernel/src/fs/fs_impls/virtiofs/inode/mod.rs
```diff
@@ -449,6 +455,35 @@ impl Inode for VirtioFsInode {
...
+            RenameOperation::new(RenameReq::new(target.nodeid()), old_name, new_name),
+        ) {
+            let err = Error::from(err);
+
+            if err.error() == Errno::ENOENT
+                && let Some(old_inode) = self.lookup_child_inode(old_name).ok()
+            {
+                old_inode.expire_entry_cache();
+            }
+            return Err(err);
```
cchanging: I’m still a bit confused: why do we need to explicitly expire it when returning `ENOENT`? Does Linux do the same thing? If we don’t proactively expire it, wouldn’t the next lookup detect it during revalidation anyway?

li041: `revalidate_exists()` currently first checks whether `entry_valid_until` has expired. If we don't explicitly expire the cache here, it's very possible that the old cached dentry is still considered valid locally even though it has already been removed on the server side.

In that case, the next lookup/revalidation path will hit the existing cache entry and skip issuing a new `FUSE_LOOKUP`, because `entry_valid_until` is still valid. As a result, userspace may continue observing a stale positive dentry for `old_name`.

So the explicit `expire_entry_cache()` on `ENOENT` is intended to invalidate that stale positive cache immediately after the server tells us the entry no longer exists.

cchanging: Will `lookup_child_inode` always return the `old_inode`? Could this update the inode cache here, causing you to actually operate on an inode that is not expired?

cchanging (approved):

StevenJiang1110 (approved):

PR #3337, opened 2026-06-05
net/tcp: dispatch incoming connections to INADDR_ANY listeners

Rodrigo-Ichaso:
## Problem

When a server binds to `0.0.0.0:port` (INADDR_ANY), incoming TCP packets addressed to a specific interface IP are not dispatched to that listener. `lookup_listener()` in `socket_table.rs` performs only an exact `(addr, port)` match, so a listener registered under `UNSPECIFIED:port` is never found when the incoming key carries a concrete IP.

This means any server using `bind(0.0.0.0, port)` — the standard POSIX pattern — never accepts connections on Asterinas.

## Fix

Add a two-stage lookup in `lookup_listener()`:

1. **Exact match** — try the key as-is (specific addr + port)
2. **Wildcard fallback** — if no exact match, retry with `UNSPECIFIED` addr and the same port

This mirrors the approach described in RFC 793 and used by Linux (which maintains a separate wildcard-listener hash table keyed by port only).

## Testing

Validated with a Rust TCP echo server bound to `0.0.0.0:8080` inside Asterinas (Proxmox KVM, virtio-net), receiving connections from the host network. Before this fix: connections timed out. After: accepted correctly.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/socket_table.rs
+++ b/kernel/libs/aster-bigtcp/src/socket_table.rs
@@ -262,16 +262,33 @@ impl<E: Ext> SocketTable<E> {
     }
 
     pub(crate) fn lookup_listener(&self, key: &ListenerKey) -> Option<&Arc<TcpListenerBg<E>>> {
+        // First: try an exact match on (addr, port).
         let bucket = {
             let hash = key.hash();
             let bucket_index = hash & LISTENER_BUCKET_MASK;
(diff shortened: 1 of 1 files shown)


lrh2000:
This patch is already included in https://github.com/asterinas/asterinas/pull/3338. It does not make sense without #3338 anyway. So I think we should close this PR and continue the discussion in #3338.

PR #3338, opened 2026-06-05
net/socket: fix INADDR_ANY bind returning EADDRNOTAVAIL

Rodrigo-Ichaso:
## Problem

`bind(0.0.0.0, port)` on Asterinas always returned `EADDRNOTAVAIL` (errno 99).

The root cause is in `get_iface_to_bind()` in `kernel/src/net/socket/ip/common.rs`. The function searches all interfaces for one whose IP address matches the requested bind address. When the address is `0.0.0.0` (INADDR_ANY), no interface has that address, so `find()` returns `None`, and `resolve_bind_iface_and_config()` returns `EADDRNOTAVAIL`.

This affects any server that binds to `0.0.0.0` — standard practice for most TCP servers.

## Fix

Detect the unspecified address (0.0.0.0 for IPv4, :: for IPv6) before the interface search and return the default outbound iface (virtio if present, loopback otherwise). This mirrors the pattern already used in `get_ephemeral_iface()`.

The `BindPortConfig` is still constructed with the original `0.0.0.0:port` endpoint, so the wildcard listener is registered correctly in bigtcp and the dispatch layer can route incoming connections to it.

## Verification

Tested with a minimal static Rust binary on an Asterinas VM (kernel serial log):

```
=== bind test start ===
OK  bind(192.168.0.200:9999)   ← specific IP (was already working)
OK  bind(0.0.0.0:9998)         ← INADDR_ANY (was EADDRNOTAVAIL before this fix)
socket() fd=3
bind(0.0.0.0:9997) raw ret=0   ← raw libc::bind confirms syscall succeeds
=== bind test end ===
```

## Relation to #3337

PR #3337 fixes the **receive** side: incoming TCP connections are dispatched to INADDR_ANY listeners. This PR fixes the **bind** side: `bind(0.0.0.0)` no longer returns an error. Both are required for a complete INADDR_ANY fix.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/socket/bound/tcp_listen.rs
+++ b/kernel/libs/aster-bigtcp/src/socket/bound/tcp_listen.rs
@@ -7,7 +7,7 @@ use ostd::sync::SpinLock;
 use smoltcp::{
     socket::PollAt,
     time::Duration,
-    wire::{IpEndpoint, IpRepr, TcpRepr},
+    wire::{IpAddress, IpEndpoint, IpListenEndpoint, IpRepr, TcpRepr},
 };
 
 use super::{
@@ -90,7 +90,14 @@ impl<E: Ext> TcpListener<E> {
 
             option.apply(&mut socket);
 
-            if let Err(err) = socket.listen(local_endpoint) {
+            // Convert unspecified addr (0.0.0.0/::) to addr:None so smoltcp
+            // accepts SYN packets destined to any local interface address.
+            let listen_endpoint = if local_endpoint.addr.is_unspecified() {
+                IpListenEndpoint { addr: None, port: local_endpoint.port }
+            } else {
+                IpListenEndpoint::from(local_endpoint)
+            };
+            if let Err(err) = socket.listen(listen_endpoint) {
                 return Err((bound, err.into()));
             }
 
--- a/kernel/libs/aster-bigtcp/src/socket_table.rs
+++ b/kernel/libs/aster-bigtcp/src/socket_table.rs
@@ -262,16 +262,33 @@ impl<E: Ext> SocketTable<E> {
     }
 
     pub(crate) fn lookup_listener(&self, key: &ListenerKey) -> Option<&Arc<TcpListenerBg<E>>> {
+        // First: try an exact match on (addr, port).
         let bucket = {
             let hash = key.hash();
             let bucket_index = hash & LISTENER_BUCKET_MASK;
             &self.listener_buckets[bucket_index as usize]
         };
+        if let Some(listener) = bucket.listeners.iter().find(|l| l.listener_key() == key) {
+            return Some(listener);
+        }
 
-        bucket
+        // Fallback: try an INADDR_ANY listener on the same port.
+        // Linux uses a separate hash table keyed by port only for wildcard listeners;
+        // here we reuse the same table with a wildcard key.
+        let wildcard_addr = match key.addr {
+            IpAddress::Ipv4(_) => IpAddress::Ipv4(Ipv4Addr::UNSPECIFIED),
+            IpAddress::Ipv6(_) => IpAddress::Ipv6(smoltcp::wire::Ipv6Address::UNSPECIFIED),
+        };
+        let wildcard_key = ListenerKey::new(wildcard_addr, key.port);
+        let wildcard_bucket = {
+            let hash = wildcard_key.hash();
+            let bucket_index = hash & LISTENER_BUCKET_MASK;
+            &self.listener_buckets[bucket_index as usize]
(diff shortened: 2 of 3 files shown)


lrh2000:
Unfortunately, the original behavior is desired because we don't currently support `0.0.0.0`. So binding to `0.0.0.0` should return an error to indicate that the kernel does not support it...

Adding proper support for `0.0.0.0` is much more complex because the socket must handle incoming connections from all interfaces, including both `eth0` and `lo`. Therefore, it requires refactoring, significantly.

Rodrigo-Ichaso:
Thank you for the feedback, @lrh2000.

I understand the concern about multi-interface support. You are right that a fully general INADDR_ANY implementation would need to handle multiple interfaces (e.g. `eth0` and `lo` independently).

However, I would like to propose a pragmatic, incremental path that fixes the most common use case without the full refactor:

The fix actually requires three layers working together:

1. **`common.rs`** — `bind(0.0.0.0)` no longer returns `EADDRNOTAVAIL`; selects the virtio interface (the only non-loopback one in practice).
2. **`socket_table.rs`** — incoming SYN packets to the specific interface address are dispatched to the wildcard listener, not dropped.
3. **`tcp_listen.rs`** — passes `IpListenEndpoint { addr: None }` to smoltcp instead of `Some(0.0.0.0)`, so smoltcp accepts SYN packets destined to any local address.

All three commits are now in this PR. Without all three, the socket binds but every incoming SYN is rejected with RST.

**Tested on real hardware**: LLM-OS (a Rust web server) running on Asterinas VM responds correctly on port 8080 after binding to `0.0.0.0` — without any `LLM_OS_BIND_ADDR` workaround.

**Known limitation**: if a system has multiple non-loopback interfaces, a wildcard-bound socket will only accept connections through the first virtio interface. This is documented behavior for this iteration.

If you would prefer this as a documented "basic/single-interface" feature rather than closing, I am happy to add a comment to `get_iface_to_bind()` making the limitation explicit. Otherwise, if you believe the correct approach is to return an explicit `ENOTSUPP` or similar instead of `EADDRNOTAVAIL`, I can adjust accordingly.

Happy to iterate on whatever direction makes most sense for the project.

StevenJiang1110:
> Happy to iterate on whatever direction makes most sense for the project.

As a kernel project, we strive to implement features in the most correct way when possible. If a particular feature is not supported, returning an error to the user as early as possible is clearly the right approach. In the early stages of the project, we may have introduced various workarounds to quickly get important workloads running, but at this stage, where the core functionality is largely in place, we are focused on progressively improving Linux compatibility. We require that all implemented features are correct, or at least more correct than their previous implementations. Indroducing some workaround requires strong motivation. Temporarily passing a few demos or applications is not our goal — we already support a wide range of applications.

If your primary goal is to get a specific application running, consider checking whether it allows you to specify the host address explicitly. In many cases, something like `XXX_HOST=127.0.0.1 ./app` (if you want to listen on localhost) or `XXX_HOST=10.0.2.15 ./app` (if you want to listen on virtio-net) should work. Most applications support this. If not, Asterinas may simply not support that application in the short term, or you could consider modifying the application's source code.

If you are interested in contributing support for `INADDR_ANY` in Asterinas, we would absolutely welcome that. It is indeed a long-standing pain point, though it requires a significant amount of effort.

Rodrigo-Ichaso:
Thanks for the feedback, @StevenJiang1110 and @lrh2000.

Before investing time in a full multi-interface INADDR_ANY implementation, I wanted to check on direction: is the approach from #3337 (dispatching incoming connections to INADDR_ANY listeners) something worth building on, or was it closed because the team is planning a different internal approach?

If there's a preferred design already in mind, I'm happy to align with it rather than work in a parallel direction. Just want to avoid duplicating effort.

StevenJiang1110:
> If there's a preferred design already in mind, I'm happy to align with it rather than work in a parallel direction. Just want to avoid duplicating effort.

We haven't settled on a specific approach yet, so feel free to propose your own design. If the proposal is fairly involved, consider opening a dedicated design proposal issue so we can discuss it there.

> is the approach from https://github.com/asterinas/asterinas/pull/3337 (dispatching incoming connections to INADDR_ANY listeners) something worth building on, or was it closed because the team is planning a different internal approach?

I guess the PR was closed primarily because #3337 is incomplete and cannot be meaningfully reviewed in isolation. Before we can assess whether any individual step is on the right track, we need to understand the overall plan. A large task can certainly be broken down into multiple commits or PRs — and we actually encourage that, as long as the breakdown is well-reasoned — but we need to see the big picture to ensure the full plan can be carried through to completion. Once code is merged, it becomes the responsibility of the community's maintainers and code owners to maintain it going forward.

PR #3339, opened 2026-06-06
Fix TDX MapGPA retry handling

Hsy-Intel:


Hsy-Intel:
@boterinas rerun

lrh2000:
Why do we need to retry? Due to race conditions caused by other code, or it is just required by some spec?

Hsy-Intel:
> Why do we need to retry? Due to race conditions caused by other code, or it is just required by some spec?

The retry is explicitly required by the GHCI spec, not added as a workaround for a local race in our code. `TDG.VP.VMCALL<MapGPA>` defines a `RETRY` status, and on that status the TD must retry the operation starting from the GPA returned in R11. So the retry logic belongs at the `MapGPA` handling layer, because this is leaf-specific ABI semantics rather than generic TDVMCALL behavior. 

Linux retry logic: https://github.com/torvalds/linux/blob/master/arch/x86/coco/tdx/tdx.c#L908-L957
Real case: https://github.com/asterinas/asterinas/actions/runs/27153648296/job/80150810486

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -162,3 +159,115 @@ impl TdxTrapFrame for TrapFrameWrapper<'_> {
...
+    }
+}
+
+fn convert_gpa_range(
+    start_gpa: u64,
+    size: u64,
+    target_state: TargetPageState,
+) -> Result<(), PageConvertError> {
+    debug_assert!(start_gpa.is_multiple_of(PAGE_SIZE as u64));
+    debug_assert!(size.is_multiple_of(PAGE_SIZE as u64));
```
lrh2000: `debug_assert!(start_gpa...)` is already done in:
```rust
pub unsafe fn unprotect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), PageConvertError> {
    debug_assert!(gpa.is_multiple_of(PAGE_SIZE));
```

`debug_assert!(size...)` should be moved there.

This two conditions, as well as the overflow check below, are safety pre-conditions:
```rust
///  - The provided physical address is page aligned.
///  - The provided physical address range is in bounds, i.e., it should fall
///    within the maximum Guest Physical Address (GPA) limit.
```

So debug assertions are fine and there is also no need to introduce a `Overflow` error variant.

Perhaps we could just accept `end_gpa` here:
```rust
pub unsafe fn protect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), PageConvertError> {
    debug_assert!(gpa.is_multiple_of(PAGE_SIZE));
    debug_assert!(size.is_multiple_of(PAGE_SIZE));

    map_gpa(gpa as u64, (gpa + size) as u64).map_err(|_| PageConvertError::TdVmcall)?;
    convert_gpa_range(gpa as u64, size as u64, TargetPageState::Private)?;
```
```rust
fn convert_gpa_range(
    start_gpa: u64,
    end_gpa: u64,
    target_state: TargetPageState,
) -> Result<(), PageConvertError> {
```

Given that `size` isn't really used in this method.

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -162,3 +159,115 @@ impl TdxTrapFrame for TrapFrameWrapper<'_> {
...
+    TdCall(TdCallError),
+    TdVmcall {
+        output_value: u64,
+        error: TdVmcallError,
+        retry_count: usize,
+    },
+}
+
+impl fmt::Display for PageConvertError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
lrh2000: I think `Debug` is sufficient. Why do we need to write a `Display` impl by hand?

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -162,3 +159,115 @@ impl TdxTrapFrame for TrapFrameWrapper<'_> {
         self.0.rbp = rbp;
     }
 }
+
+#[derive(Debug)]
+pub(crate) enum PageConvertError {
```
lrh2000: This should have the same visibility as the existing method `pub unsafe fn protect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), PageConvertError> {` to avoid confusion.

In addition, the new type/method order is odd. Why should a unrelated `TrapFrameWrapper` be inserted in between? The new types and methods added in this PR should be moved just below `protect_gpa_tdvm_call`..

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -162,3 +159,115 @@ impl TdxTrapFrame for TrapFrameWrapper<'_> {
...
+                "TdVmcall {{ output_value: {output_value:#x}, error: {error:?}, retry_count: {retry_count} }}"
+            ),
+        }
+    }
+}
+
+const SHARED_BIT: u8 = 51;
+const SHARED_MASK: u64 = 1u64 << SHARED_BIT;
+// Retrying the same page a second time should succeed; use 3 just in case.
+const MAX_MAP_GPA_RETRIES_PER_PAGE: usize = 3;
```
lrh2000: Move these constants to a more suitable place, e.g.,

```rust
impl TargetPageState {
    const SHARED_BIT: u8 = 51;
    const SHARED_MASK: u64 = 1u64 << SHARED_BIT;
```

and

```rust
fn convert_gpa_range(
    start_gpa: u64,
    end_gpa: u64,
    target_state: TargetPageState,
) -> Result<(), PageConvertError> {
    // Retrying the same page a second time should succeed; use 3 just in case.
    const MAX_MAP_GPA_RETRIES_PER_PAGE: usize = 3;

    let mut next_gpa = start_gpa;
    let mut retry_count = 0;
```

lrh2000 (commented):
> `TDG.VP.VMCALL<MapGPA>` defines a `RETRY` status, and on that status the TD must retry the operation starting from the GPA returned in R11.

Thanks for the information! I misunderstood the code previously..

ostd/src/mm/dma/util.rs
```diff
@@ -301,3 +300,28 @@ fn unmap_dma_remap(daddr_range: Option<Range<Daddr>>) {
...
+                "failed to {action} the DMA segment in TDX guest: gpa={gpa:#x}, size={size:#x}, tdcall_error={error:?}"
+            )
+        }
+        crate::arch::tdx_guest::PageConvertError::TdVmcall {
+            output_value,
+            error,
+            retry_count,
+        } => {
+            panic!(
+                "failed to {action} the DMA segment in TDX guest: gpa={gpa:#x}, size={size:#x}, output_value={output_value:#x}, tdvmcall_error={error:?}, retry_count={retry_count}"
```
lrh2000: Sorry, but I still don't understand why don't you just use the automatically generated version of the `Debug` trait? What's wrong with it? Do you want to print hex values only for `output_value`, but not for the other fields? Or you have other reasons?

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -50,18 +46,100 @@ pub unsafe fn unprotect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), Pag
...
+            Self::Private => gpa,
+            Self::Shared => {
+                debug_assert_eq!(SHARED_MASK, 1u64 << SHARED_BIT);
+                gpa | SHARED_MASK
+            }
+        }
+    }
+}
+
+fn convert_gpa_range(
```
lrh2000: Is this safe? I believe not. If so,

```suggestion
/// # Safety
///
/// The caller must ensure that `start_gpa..end_gpa` represents a valid GPA
/// region that can be safely converted to `target_state`.
unsafe fn convert_gpa_range(
```

Also at the caller:
```c
        // SAFETY: The caller ensures the safety of this operation.
```

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -50,18 +46,100 @@ pub unsafe fn unprotect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), Pag
...
+    Private,
+    Shared,
+}
+
+impl TargetPageState {
+    fn to_map_gpa(self, gpa: u64) -> u64 {
+        match self {
+            Self::Private => gpa,
+            Self::Shared => {
+                debug_assert_eq!(SHARED_MASK, 1u64 << SHARED_BIT);
```
lrh2000:
```suggestion
                const { assert!(SHARED_MASK == 1u64 << SHARED_BIT) };
```

Honestly, we could even do this at compile time.

ostd/src/mm/dma/util.rs
Hsy-Intel: I’m just not quite used to writing `#[expect(dead_code)]` for struct fields. I’ve switched to using the `Debug` trait instead.

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -50,18 +47,107 @@ pub unsafe fn unprotect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), Pag
...
 
-    map_gpa(gpa as u64, size as u64).map_err(|_| PageConvertError::TdVmcall)?;
+    // SAFETY: The caller ensures the safety of this operation.
+    unsafe { convert_gpa_range(gpa as u64, (gpa + size) as u64, TargetPageState::Private)? };
     for page_gpa in (gpa..gpa + size).step_by(PAGE_SIZE) {
         // SAFETY: The caller ensures the safety of this operation.
         unsafe {
-            accept_page(0, page_gpa as u64).map_err(|_| PageConvertError::TdCall)?;
+            accept_page(0, page_gpa as u64).map_err(PageConvertError::TdCall)?;
         }
```
lrh2000: Try this:
```patch
diff --git a/ostd/src/arch/x86/tdx_guest.rs b/ostd/src/arch/x86/tdx_guest.rs
index 4c8b359f6..010ce9f72 100644
--- a/ostd/src/arch/x86/tdx_guest.rs
+++ b/ostd/src/arch/x86/tdx_guest.rs
@@ -54,9 +54,7 @@ pub unsafe fn protect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), PageC
     map_gpa(gpa as u64, size as u64).map_err(|_| PageConvertError::TdVmcall)?;
     for page_gpa in (gpa..gpa + size).step_by(PAGE_SIZE) {
         // SAFETY: The caller ensures the safety of this operation.
-        unsafe {
-            accept_page(0, page_gpa as u64).map_err(|_| PageConvertError::TdCall)?;
-        }
+        unsafe { accept_page(0, page_gpa as u64).map_err(|_| PageConvertError::TdCall)? };
     }
 
     Ok(())
```

If rustfmt is happy with it, I think it is more consistent and concise.

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -50,18 +47,107 @@ pub unsafe fn unprotect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), Pag
...
+    target_state: TargetPageState,
+) -> Result<(), PageConvertError> {
+    // Retrying the same page a second time should succeed; use 3 just in case.
+    const MAX_MAP_GPA_RETRIES_PER_PAGE: usize = 3;
+
+    let mut next_gpa = start_gpa;
+    let mut retry_count = 0;
+
+    loop {
+        let request_gpa = target_state.to_map_gpa(next_gpa);
```
lrh2000: To be honest, `to_map_gpa` and `request_gpa` sound like bad names to me. The result is not a GPA, it is a GPA with some masks. I suggest:
```rust
let gpa_with_mask = next_gpa | target_state.as_gpa_mask();
```

ostd/src/arch/x86/tdx_guest.rs
```diff
@@ -50,18 +47,107 @@ pub unsafe fn unprotect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), Pag
...
+            Err((retry_gpa, TdVmcallError::TdxRetry)) => {
+                if !(next_gpa..end_gpa).contains(&retry_gpa)
+                    || !retry_gpa.is_multiple_of(PAGE_SIZE as u64)
+                {
+                    return Err(PageConvertError::TdVmcall {
+                        output_value: retry_gpa,
+                        error: TdVmcallError::TdxRetry,
+                        retry_count,
+                    });
+                }
```
lrh2000: I think you can just do:
```
Err((retry_gpa, TdVmcallError::TdxRetry))
    if (next_gpa..end_gpa).contains(&retry_gpa) &&
       retry_gpa.is_multiple_of(PAGE_SIZE as u64)
=> {}
```

We can fall back to the error branch below naturally in all other cases, right?
```rust
            Err((output_value, error)) => {
                return Err(PageConvertError::TdVmcall {
                    output_value,
                    error,
                    retry_count,
                });
            }
```

ostd/src/mm/dma/util.rs
```diff
@@ -301,3 +300,15 @@ fn unmap_dma_remap(daddr_range: Option<Range<Daddr>>) {
         // FIXME: Flush IOTLBs to prevent any future DMA access to the frames.
     }
 }
+
+#[cfg(all(target_arch = "x86_64", any(debug_assertions, feature = "cvm_guest")))]
+fn panic_page_convert_error(
```
lrh2000: This should be moved below `dealloc_protect_physical_range`. It has no relationship to `dma_remap` and `unmap_dma_remap`.

lrh2000 (commented):
Just some minor nits. Otherwise the changes look good!

ostd/src/arch/x86/tdx_guest.rs
Hsy-Intel: Seems only this format works fine with `make check`:

```rust
    for page_gpa in (gpa..gpa + size).step_by(PAGE_SIZE) {
        // SAFETY: The caller ensures the safety of this operation.
        unsafe { accept_page(0, page_gpa as u64).map_err(PageConvertError::TdCall)? };
    }
```

Written this way will throw out an error:

```rust
    for page_gpa in (gpa..gpa + size).step_by(PAGE_SIZE) {
        // SAFETY: The caller ensures the safety of this operation.
        unsafe { accept_page(0, page_gpa as u64).map_err(|_| PageConvertError::TdCall)? };
    }
```

```txt
$:~/asterinas# make check
Checking default workspace members
    Checking ostd v0.18.0 (/root/asterinas/ostd)
error[E0277]: `?` couldn't convert the error to `arch::tdx_guest::PageConvertError`
  --> ostd/src/arch/x86/tdx_guest.rs:56:87
   |
48 | pub unsafe fn protect_gpa_tdvm_call(gpa: Paddr, size: usize) -> Result<(), PageConvertError> {
   |                                                                 ---------------------------- expected `arch::tdx_guest::PageConvertError` because of this
...
56 |         unsafe { accept_page(0, page_gpa as u64).map_err(|_| PageConvertError::TdCall)? };
   |                  ------------------------------- -------------------------------------^ the trait `core::convert::From<fn(tdx_guest::tdcall::TdCallError) -> arch::tdx_guest::PageConvertError {arch::tdx_guest::PageConvertError::TdCall}>` is not implemented for `arch::tdx_guest::PageConvertError`
   |                  |                               |
   |                  |                               this can't be annotated with `?` because it has type `Result<_, fn(TdCallError) -> PageConvertError {PageConvertError::TdCall}>`
   |                  this has type `Result<_, tdx_guest::tdcall::TdCallError>`
   |
note: `arch::tdx_guest::PageConvertError` needs to implement `From<fn(tdx_guest::tdcall::TdCallError) -> arch::tdx_guest::PageConvertError {arch::tdx_guest::PageConvertError::TdCall}>`
  --> ostd/src/arch/x86/tdx_guest.rs:63:1
   |
63 | pub enum PageConvertError {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
   = note: the full name for the type has been written to '/root/asterinas/target/x86_64-unknown-none/debug/deps/ostd-48262ca29e257736.long-type-13095716261815850816.txt'
   = note: consider using `--verbose` to print the full type name to the console

For more information about this error, try `rustc --explain E0277`.
error: could not compile `ostd` (lib) due to 1 previous error
[Error]: Command RUSTFLAGS="-Dwarnings --check-cfg cfg(ktest)" "cargo" "clippy" "-Zbuild-std=core,alloc,compiler_builtins" "-Zbuild-std-features=compiler-builtins-mem" "--target" "x86_64-unknown-none" "--no-deps" failed with status: ExitStatus(unix_wait_status(25856))
make: *** [Makefile:425: check] Error 101
```

lrh2000 (approved):

PR #3341, opened 2026-06-07
Inspect output flags (OPOST/ONLCR/OCRNL) in TTY line discipline for echo path and write path

Solicey:
## Problem

The TTY line discipline was ignoring termios output flags (`OPOST`, `ONLCR`, `OCRNL`) when echoing characters and processing writes. With the default `OPOST | ONLCR`, a `\n` should be output as `\r\n`, but the code echoed bare `\n` — which might cause the terminal cursor to move down a line without returning to column 0.

The previous PR (#3301) only fixed the echo path and was accidentally closed by me, so I'm opening this new PR. This version adds the missing write path and modifies regression tests.

### Write path processing

The write path (`Tty::write`) now runs user data through `process_output` before handing it to the driver, so output flags are respected regardless of whether data comes from echo or from a process writing to the terminal.

### PTY driver cleanup

The PTY driver's `push_output` had a hardcoded `\n → \r\n` expansion with a TODO saying it belonged in the TTY layer. Since `process_output` now handles this, the driver's duplicate expansion is removed. Without this fix, writes through a PTY would double-process newlines (`\n` → `\r\r\n`) and return incorrect byte counts to userspace, causing `EFAULT`.

### Regression tests

Added 6 regression tests covering ONLCR and OCRNL in both echo and write paths, using `TEST_SUCC`/`TEST_RES` rather than `CHECK`:

| Test | Path | Scenario |
|------|------|----------|
| `echo_onlcr_enabled` | Echo | `\n` echoed as `\r\n` with default flags |
| `echo_onlcr_disabled` | Echo | `\n` kept as `\n` when ONLCR is off |
| `echo_ocrnl_enabled` | Echo | `\r` echoed as `\n` with OCRNL, ICRNL off |
| `write_onlcr_enabled` | Write | Slave writes `A\n`, master reads `A\r\n` |
| `write_onlcr_disabled` | Write | Slave writes `A\n`, master reads `A\n` (ONLCR off) |
| `write_ocrnl_enabled` | Write | Slave writes `\r`, master reads `\n` (OCRNL on) |

## Summary

- **`kernel/src/device/tty/termio.rs`** — added `output_flags()` accessor
- **`kernel/src/device/tty/line_discipline.rs`** — `echo_char` respects `OPOST`/`ONLCR`/`OCRNL`, added `process_output`
- **`kernel/src/device/tty/mod.rs`** — write path calls `process_output`, returns correct byte count
- **`kernel/src/device/pty/driver.rs`** — removed hardcoded `\n → \r\n` expansion, cleaned up `can_push`
- **`test/initramfs/src/regression/device/pty/output_flags.c`** — new regression tests
- **`test/initramfs/src/regression/device/run_test.sh`** — registers the new tests

## Behavior reference

Matches Linux `n_tty.c`:
- [ONLCR](https://elixir.bootlin.com/linux/v7.0/source/drivers/tty/n_tty.c#L414) — `\n` → `\r\n`
- [OCRNL](https://elixir.bootlin.com/linux/v7.0/source/drivers/tty/n_tty.c#L427) — `\r` → `\n`

### DIFF
--- a/kernel/src/device/pty/driver.rs
+++ b/kernel/src/device/pty/driver.rs
@@ -128,12 +128,7 @@ impl TtyDriver for PtyDriver {
 
         let mut len = 0;
         for ch in chs {
-            // TODO: This is termios-specific behavior and should be part of the TTY implementation
-            // instead of the TTY driver implementation. See the ONLCR flag for more details.
-            if *ch == b'\n' && output.capacity() - output.len() >= 2 {
-                output.push(b'\r').unwrap();
-                output.push(b'\n').unwrap();
-            } else if *ch != b'\n' && !output.is_full() {
+            if !output.is_full() {
                 output.push(*ch).unwrap();
             } else if len == 0 {
                 return_errno_with_message!(Errno::EAGAIN, "the output buffer is full");
@@ -164,8 +159,7 @@ impl TtyDriver for PtyDriver {
     }
 
     fn can_push(&self) -> bool {
-        let output = self.output.lock();
-        output.capacity() - output.len() >= 2
+        !self.output.lock().is_full()
     }
 
     fn notify_input(&self) {
--- a/kernel/src/device/tty/line_discipline.rs
+++ b/kernel/src/device/tty/line_discipline.rs
@@ -2,7 +2,7 @@
 
 use ostd::const_assert;
 
-use super::termio::{CCtrlCharId, CTermios, CWinSize};
+use super::termio::{CCtrlCharId, COutputFlags, CTermios, CWinSize};
 use crate::{
     device::tty::termio::{CInputFlags, CLocalFlags},
     prelude::*,
@@ -113,7 +113,7 @@ impl LineDiscipline {
         // Typically, a TTY in raw mode does not echo. But the TTY can also be in a CBREAK mode,
         // with ICANON closed and ECHO opened.
         if self.termios.local_flags().contains(CLocalFlags::ECHO) {
-            self.output_char(ch, echo_callback);
+            self.echo_char(ch, echo_callback);
         }
 
         if self.is_full() {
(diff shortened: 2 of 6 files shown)


Solicey:
Hello, I've been debugging the last [CI failure](https://github.com/asterinas/asterinas/actions/runs/27093835857/job/79962214334?pr=3341) on this PR and noticed something odd — the actual test output shows "All conformance tests passed." printed in the log. The issue appears to be that this line doesn't start at column 0, so the Makefile grep (`^All conformance tests passed.`) doesn't match it.

I suspect this is related to the `\r` / `\n` handling in my changes, but I haven't been able to trace the exact cause yet. Would appreciate your help looking into this.

StevenJiang1110:
First of all, thank you for your contribution — it is greatly appreciated!

> Hello, I've been debugging the last [CI failure](https://github.com/asterinas/asterinas/actions/runs/27093835857/job/79962214334?pr=3341) on this PR and noticed something odd — the actual test output shows "All conformance tests passed." printed in the log. The issue appears to be that this line doesn't start at column 0, so the Makefile grep (`^All conformance tests passed.`) doesn't match it.
> 
> I suspect this is related to the `\r` / `\n` handling in my changes, but I haven't been able to trace the exact cause yet. Would appreciate your help looking into this.

I have not yet had the time to take a close look at your changes.

Regarding the issue you encountered, we have actually run into this before — it is tracked in #3181. At the time, we traced it back to our sendfile implementation, and it was subsequently fixed in #3191. Since the problem stopped occurring after that fix, I suspect this PR may have reintroduced it.

I would suggest investigating from two angles:
1. Check whether the current handling of output flags is causing some content to be dropped.
2. Check whether the length returned by the write/read functions matches the amount of data actually written to userspace. I think this is the most likely culprit — if the return value is inconsistent with the actual number of bytes written, the position for the next read/write will be off.

You can also try the reproducer code from #3181 to see if the issue can be replicated with that simpler case, which would make it easier to validate the fix. Additionally, you might consider using an AI coding agent such as Codex or Claude Code to assist with debugging — in our experience, agents tend to be quite effective at pinpointing this kind of issue, especially when given a reproducible test case.

If you run into any further difficulties, please feel free to continue the discussion here. I will mark this PR as a draft for now; you can mark it as ready once the issue is resolved and CI passes.

lrh2000:
https://github.com/asterinas/asterinas/blob/22f59db61ec7783281a4f6088a6530173f5a5ac5/kernel/src/device/tty/mod.rs#L215-L220

Fewer bytes may be pushed due to a full buffer. It is incorrect to ignore the return value. This seems like a likely reason for the CI to fail.

---

For what I can see, we likely need to handle the output flags in the driver layer instead of in the TTY layer. Otherwise we cannot know whether "\r\n" should be treated as a whole or as two separate bytes. Our previous approach of treating them as a whole when they come from the ONLCR flag was correct and should not be changed:
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/pty/driver.rs#L133-L135

It needs be something like:
```patch
- if *ch == b'\n' && && output.capacity() - output.len() >= 2 { 
+ if *ch == b'\n' && ONLCR && && output.capacity() - output.len() >= 2 { 
      output.push(b'\r').unwrap(); 
      output.push(b'\n').unwrap(); 
```

Duplicating the logic in each TTY does not sound like a good idea, of course. Therefore, we will need to introduce a new abstraction to help. Something like:
```rust
struct OutputBehavior;

impl OutputBehavior {
    // (This should be used in `output.capacity() - output.len() >= 2`).
    const PROCESS_MULTIPLIER: usize = 2;

    // (This should be used in `TtyDriver::push_output`).
    fn process<'a>(&self, ch: &'a u8) -> &'a [u8] {
        match *ch {
            b'\n' if ONLCR => b"\r\n",
            _ => core::slice::from_ref(ch)
        }
    }
}
```

Also we should query the output flags just before `driver.push_output`. Querying it before `wait_events` may not be suitable because a long time can pass during waiting. Currently, releasing the `ldisc` lock before `push_output` should be fine, but we will likely need to hold the `ldisc` lock during `push_output` due to https://github.com/asterinas/asterinas/pull/3294#issuecomment-4656575058.
```rust
let try_write = || todo!("query output flags and push output");

if non_blocking {
    try_write();
} else {
    self.wait_events(try_write);
}
```

PR #3346, opened 2026-06-08
Avoid redundant checks when implementing the `Inode` trait

cchanging:
Based on #3311.

When reviewing filesystem-related implementations, I often find many checks that are redundant with those already performed in the VFS layer. The reason this happens is that we have not clearly defined what checks are handled on the VFS side. By explicitly documenting these completed pre-checks in the `Inode` documentation, this PR aims to give future filesystem implementers a clearer understanding of them.

At the same time, this PR removes some existing redundant checks and consolidates certain checks into the VFS layer wherever possible. The guiding principle here is that checks that are general and required by every filesystem can usually be handled in the VFS layer.

Includes #3118.


lrh2000:
Then this is the right time for https://github.com/asterinas/asterinas/pull/2521#discussion_r2545039085.

Issue #455, opened 2023-10-23
Jinux network develepment state (2023-10-23)


StevenJiang1110:
This issue is to record the development state of jinux network, for communication use.

# Done:
### Syscalls
socket, bind, connect, listen, accept, shutdown, etc.
### Socket 
TCP socket，UDP socket，Unix domain socket
### TCP/IP stack
smoltcp, a third-party crate
### driver
virtio-net
# TODOs:
### Syscalls
support different flags (e.g., nonblocking, accept4 flags) and options
### Socket 
- basic IPv6 support: IPv6 localhost; IPv4 - IPv6 mapping
- Raw socket: ping; dhcp client
- Netlink
- Other sockets
### TCP/IP stack:
- UDP high packet loss rate.
- Maintain `smoltcp`: bug fix; implememt missing features
- Turn to our own stack: `smoltcp` is designed for real-time applications.
### drivers: 
- Virtio-net driver enhancement:
    - single queue -> multiple queue
    - interrupt-driven packet sending
    - Others......
- e1000 driver
- RDMA driver
- other drivers  
### Performance
 - softirq + tasklet
 - others

PR #3347, opened 2026-06-09
Support MSG_PEEK for all sockets

StevenJiang1110:
This PR is part of #3202. I moved the `MSG_PEEK` implementation here since #3202 has grown quite large.

`MSG_PEEK` allows receiving a message without consuming the data from the buffer.

One issue I encountered during implementation involves how TCP sockets handle wrap-around in the ring buffer — that is, when a read reaches the end of the ring buffer but the remaining data wraps around to the beginning. Currently, this is handled by issuing two separate receive calls to ensure both segments are fully read. However, this becomes problematic for MSG_PEEK: with the current approach, a peek can only return the segment up to the end of the ring buffer. To also retrieve the wrapped-around segment at the beginning, we would need to first consume the first segment via a regular receive, and then issue another peek — which defeats the purpose.

I also considered maintaining a temporary peek buffer inside the socket, but this does not seem ideal either, as it would introduce an additional memory copy. For now, this wrap-around case is left unhandled.

### DIFF
--- a/kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
+++ b/kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
@@ -18,6 +18,7 @@ use smoltcp::{
 };
 
 use super::{
+    ReceiveBehavior,
     common::{Inner, NeedIfacePoll, Socket, SocketBg},
     tcp_listen::TcpListenerBg,
 };
@@ -445,12 +446,13 @@ impl<E: Ext> TcpConnection<E> {
     /// Receives some data.
     ///
     /// Polling the iface _may_ be required after this method succeeds.
-    pub fn recv<F, CopyErr>(
+    pub fn recv<CopyFn, CopyErr>(
         &self,
-        mut f: F,
+        behavior: ReceiveBehavior,
+        mut copy_fn: CopyFn,
     ) -> Result<(NonZeroUsize, NeedIfacePoll), IoError<RecvError, CopyErr>>
     where
-        F: FnMut(&mut [u8]) -> Result<usize, (CopyErr, usize)>,
+        CopyFn: FnMut(&[u8]) -> Result<usize, (CopyErr, usize)>,
     {
         let common = self.iface().common();
         let mut iface = common.interface();
@@ -462,7 +464,7 @@ impl<E: Ext> TcpConnection<E> {
         }
 
         let mut total_recv_bytes = 0;
-        let mut need_wrap_continuation = true;
+        let mut need_wrap_continuation = behavior.will_consume_data();
 
         let result = loop {
             let mut recv_bytes = 0;
@@ -471,9 +473,9 @@ impl<E: Ext> TcpConnection<E> {
             // `socket.recv` exposes the largest contiguous readable range. If the receive ring
             // buffer wraps, one logical stream read may span two such ranges, so continue at most
             // once while holding the same socket lock to avoid an avoidable short read.
-            let recv_result = socket.recv(|socket_buffer| {
+            let mut copy_socket_buffer = |socket_buffer: &[u8]| {
                 socket_buffer_len = socket_buffer.len();
-                match f(socket_buffer) {
+                match copy_fn(socket_buffer) {
                     Ok(current_recv_bytes) => {
                         recv_bytes = current_recv_bytes;
                         (current_recv_bytes, Ok(()))
@@ -483,7 +485,23 @@ impl<E: Ext> TcpConnection<E> {
                         (current_recv_bytes, Err(err))
                     }
                 }
-            });
+            };
+
+            let recv_result = match behavior {
+                ReceiveBehavior::Recv => {
+                    socket.recv(|socket_buffer| copy_socket_buffer(socket_buffer))
+                }
+                ReceiveBehavior::Peek => {
+                    // FIXME: This implementation relies on `socket.peek()`, which only returns the
+                    // first contiguous readable range. We need more powerful smoltcp APIs to peek the
+                    // wrapped receive buffer without extra caching and copying.
+                    let recv_queue = socket.recv_queue();
+                    socket.peek(recv_queue).map(|socket_buffer| {
+                        let (_, copy_result) = copy_socket_buffer(socket_buffer);
+                        copy_result
+                    })
+                }
+            };
 
             let copy_result = match recv_result {
                 Err(_) if socket.is_rst_closed && total_recv_bytes == 0 => {
--- a/kernel/libs/ring-buffer/src/lib.rs
+++ b/kernel/libs/ring-buffer/src/lib.rs
@@ -466,6 +466,11 @@ impl<T, R: Deref<Target = RingBuffer<T>>> Consumer<T, R> {
 }
 
 impl<T, R: Deref<Target = RingBuffer<T>>> Consumer<T, R> {
+    /// Returns the underlying ring buffer.
+    pub fn ring_buffer(&self) -> &RingBuffer<T> {
+        &self.rb
+    }
+
     /// Commits a read operation by advancing the head pointer.
     ///
     /// This method is intended for advanced use cases where the caller reads
@@ -475,7 +480,7 @@ impl<T, R: Deref<Target = RingBuffer<T>>> Consumer<T, R> {
     /// # Panics
     ///
     /// Panics if `len` exceeds the number of available items in the buffer.
-    pub fn commit_read(&self, len: usize) {
+    pub fn commit_read(&mut self, len: usize) {
         assert!(
             len <= self.len(),
             "commit_read: len exceeds available items"
--- a/kernel/src/fs/utils/endpoint.rs
+++ b/kernel/src/fs/utils/endpoint.rs
@@ -122,7 +122,7 @@ impl AsRef<EndpointState> for EndpointState {
 }
 
 impl<T: AsRef<EndpointState>> Endpoint<T> {
-    /// Reads from the endpoint and updates the local/remote [`Pollee`]s.
+    /// Reads from the endpoint.
     ///
     /// Note that if `read` returns `Ok(0)`, it is assumed that there is no data to read and an
     /// [`Errno::EAGAIN`] error will be returned instead.
@@ -139,8 +139,6 @@ impl<T: AsRef<EndpointState>> Endpoint<T> {
         let read_len = read()?;
 
         if read_len > 0 {
-            self.peer_end().as_ref().pollee.notify(IoEvents::OUT);
-            self.this_end().as_ref().pollee.invalidate();
             Ok(read_len)
         } else if is_shutdown {
             Ok(0)
@@ -149,7 +147,7 @@ impl<T: AsRef<EndpointState>> Endpoint<T> {
         }
     }
 
-    /// Writes to the endpoint and updates the local/remote [`Pollee`]s.
+    /// Writes to the endpoint.
     ///
     /// Note that if `write` returns `Ok(0)`, it is assumed that there is no space to write and an
     /// [`Errno::EAGAIN`] error will be returned instead.
@@ -166,13 +164,23 @@ impl<T: AsRef<EndpointState>> Endpoint<T> {
 
         let written_len = write()?;
 
-        if written_len > 0 {
-            self.peer_end().as_ref().pollee.notify(IoEvents::IN);
-            self.this_end().as_ref().pollee.invalidate();
-            Ok(written_len)
-        } else {
+        if written_len == 0 {
             return_errno_with_message!(Errno::EAGAIN, "the channel is full");
         }
+
+        Ok(written_len)
+    }
+
+    /// Notifies the local/remote [`Pollee`]s that data have been read.
+    pub fn notify_read(&self) {
+        self.peer_end().as_ref().pollee.notify(IoEvents::OUT);
+        self.this_end().as_ref().pollee.invalidate();
+    }
+
+    /// Notifies the local/remote [`Pollee`]s that data have been written.
+    pub fn notify_write(&self) {
+        self.peer_end().as_ref().pollee.notify(IoEvents::IN);
+        self.this_end().as_ref().pollee.invalidate();
     }
 
     /// Polls the I/O events in the local [`Pollee`].
--- a/kernel/src/net/socket/unix/ctrl_msg.rs
+++ b/kernel/src/net/socket/unix/ctrl_msg.rs
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use core::fmt;
+use core::{fmt, mem};
 
+use aster_bigtcp::socket::ReceiveBehavior;
 use aster_rights::ReadOp;
 use ostd::task::Task;
 
@@ -274,7 +275,11 @@ impl AuxiliaryData {
     }
 
     /// Generates the control messages from the auxiliary data.
-    pub(super) fn generate_control(&mut self, is_pass_cred: bool) -> Vec<ControlMessage> {
+    pub(super) fn generate_control(
+        &mut self,
+        behavior: ReceiveBehavior,
+        is_pass_cred: bool,
+    ) -> Vec<ControlMessage> {
         let mut ctrl_msgs = Vec::new();
 
         let Self { files, cred } = self;
@@ -290,9 +295,11 @@ impl AuxiliaryData {
         }
 
         if !files.is_empty() {
-            let unix_ctrl_msg = UnixControlMessage(Message::Files(FileMessage {
-                files: core::mem::take(files),
-            }));
+            let files = match behavior {
+                ReceiveBehavior::Recv => mem::take(files),
+                ReceiveBehavior::Peek => files.clone(),
+            };
+            let unix_ctrl_msg = UnixControlMessage(Message::Files(FileMessage { files }));
             ctrl_msgs.push(ControlMessage::Unix(unix_ctrl_msg));
         }
 
--- a/kernel/src/net/socket/unix/datagram/message.rs
+++ b/kernel/src/net/socket/unix/datagram/message.rs
@@ -13,7 +13,7 @@ use crate::{
             addr::{UnixSocketAddrBound, UnixSocketAddrKey},
             ctrl_msg::AuxiliaryData,
         },
-        util::ControlMessage,
+        util::{ControlMessage, SendRecvFlags},
     },
     prelude::*,
     process::signal::Pollee,
@@ -163,6 +163,7 @@ impl MessageReceiver {
     pub(super) fn try_recv(
         &self,
         writer: &mut dyn MultiWrite,
+        flags: SendRecvFlags,
     ) -> Result<(usize, Vec<ControlMessage>, UnixSocketAddr)> {
         let mut inner = self.queue.inner.lock();
         let inner = inner.as_mut().unwrap();
@@ -180,18 +181,26 @@ impl MessageReceiver {
             warn!("setting MSG_TRUNC is not supported");
         }
 
-        let mut msg = inner.messages.pop_front().unwrap();
-        inner.total_length -= msg.bytes.len();
-
         let is_pass_cred = self.queue.is_pass_cred.load(Ordering::Relaxed);
-        let ctrl_msgs = msg.aux.generate_control(is_pass_cred);
+        let src = msg.src.clone();
 
-        self.queue.pollee.invalidate();
-        // A writer may still fail if the free space is not enough.
-        // So we have to wake up all the writers here.
-        self.queue.send_wait_queue.wake_all();
+        let behavior = flags.receive_behavior();
+        let ctrl_msgs = if behavior.will_consume_data() {
+            let mut message = inner.messages.pop_front().unwrap();
+            inner.total_length -= message.bytes.len();
 
-        Ok((len, ctrl_msgs, msg.src))
+            self.queue.pollee.invalidate();
+            // A writer may still fail if the free space is not enough.
+            // So we have to wake up all the writers here.
+            self.queue.send_wait_queue.wake_all();
+
+            message.aux.generate_control(behavior, is_pass_cred)
+        } else {
+            let message = inner.messages.front_mut().unwrap();
+            message.aux.generate_control(behavior, is_pass_cred)
+        };
+
+        Ok((len, ctrl_msgs, src))
     }
 
     pub(super) fn shutdown(&self) {
--- a/kernel/src/net/socket/unix/stream/connected.rs
+++ b/kernel/src/net/socket/unix/stream/connected.rs
@@ -14,13 +14,13 @@ use crate::{
         unix::{
             UnixSocketAddr, addr::UnixSocketAddrBound, cred::SocketCred, ctrl_msg::AuxiliaryData,
         },
-        util::{ControlMessage, SockShutdownCmd},
+        util::{ControlMessage, SendRecvFlags, SockShutdownCmd},
     },
     prelude::*,
     process::signal::Pollee,
     util::{
         MultiRead, MultiWrite,
-        ring_buffer::{ConsumerU8Ext, ProducerU8Ext, RbConsumer, RbProducer, RingBuffer},
+        ring_buffer::{ProducerU8Ext, RbConsumer, RbProducer, RingBuffer, RingBufferU8Ext},
     },
 };
 
@@ -117,6 +117,7 @@ impl Connected {
         &self,
         writer: &mut dyn MultiWrite,
         is_seqpacket: bool,
+        flags: SendRecvFlags,
     ) -> Result<(usize, Vec<ControlMessage>)> {
         let is_empty = writer.is_empty();
         if is_empty && !is_seqpacket {
@@ -135,14 +136,24 @@ impl Connected {
         let no_aux_len = reader.len();
 
         let is_pass_cred = this_end.is_pass_cred.load(Ordering::Relaxed);
+        let behavior = flags.receive_behavior();
 
         // Fast path: There are no auxiliary data to receive.
         if !peer_end.has_aux.load(Ordering::Relaxed) {
-            let read_len = self
-                .inner
-                .read_with(move || reader.read_fallible_with_max_len(writer, no_aux_len))?;
+            let read_len = self.inner.read_with(|| {
+                let head = reader.head();
+                let copy_range = head..head + Wrapping(no_aux_len);
+                reader.ring_buffer().pick_fallible(copy_range, writer)
+            })?;
+            if behavior.will_consume_data() {
+                reader.commit_read(read_len);
+                if read_len > 0 {
+                    self.inner.notify_read();
+                }
+            }
+
             let ctrl_msgs = if is_pass_cred {
-                AuxiliaryData::default().generate_control(is_pass_cred)
+                AuxiliaryData::default().generate_control(behavior, is_pass_cred)
             } else {
                 Vec::new()
             };
@@ -150,79 +161,116 @@ impl Connected {
         }
 
         let mut all_aux = peer_end.all_aux.lock();
-        let mut aux_prev_data: Option<AuxiliaryData> = None;
-        let mut read_tot_len = 0;
+        let mut aux_pos = 0;
 
-        let aux_data = loop {
-            let read_start = reader.head();
-            let (aux_len, aux_front) = if let Some(front) = all_aux.front_mut() {
+        let read_base = reader.head();
+        let mut read_tot_len = 0;
+        let mut trunc_len = 0;
+
+        loop {
+            let read_start = read_base + Wrapping(read_tot_len);
+            let (aux_len, aux_front) = if let Some(front) = all_aux.get(aux_pos) {
                 if front.start == read_start {
                     ((front.end - read_start).0, Some(front))
                 } else {
                     ((front.start - read_start).0, None)
                 }
             } else {
-                (usize::MAX, None)
+                ((reader.tail() - read_start).0, None)
             };
 
             // Unless the auxiliary data we have already received is a subset of the current
             // auxiliary data, we cannot receive additional bytes.
-            if let Some(prev) = aux_prev_data.as_mut() {
-                let is_subset = if let Some(front) = aux_front.as_ref() {
-                    prev.is_subset_of(&front.data, is_pass_cred)
-                } else {
-                    prev.is_subset_of(&AuxiliaryData::default(), is_pass_cred)
+            if read_tot_len > 0 {
+                let is_subset = match (aux_pos.checked_sub(1), aux_front.as_ref()) {
+                    (Some(prev_pos), Some(front)) => {
+                        let prev = all_aux.get(prev_pos).unwrap();
+                        prev.data.is_subset_of(&front.data, is_pass_cred)
+                    }
+                    (Some(prev_pos), None) => {
+                        let prev = all_aux.get(prev_pos).unwrap();
+                        prev.data
+                            .is_subset_of(&AuxiliaryData::default(), is_pass_cred)
+                    }
+                    (None, Some(front)) => {
+                        AuxiliaryData::default().is_subset_of(&front.data, is_pass_cred)
+                    }
+                    (None, None) => true,
                 };
                 if !is_subset {
-                    break prev;
+                    break;
                 }
             }
 
             // Read the payload bytes of the current auxiliary data.
-            let read_res = if !is_empty && aux_len > 0 {
-                self.inner
-                    .read_with(|| reader.read_fallible_with_max_len(writer, aux_len))
+            let read_res = if !is_empty && (aux_len > 0 || aux_front.is_none()) {
+                self.inner.read_with(|| {
+                    let copy_range = read_start..read_start + Wrapping(aux_len);
+                    reader.ring_buffer().pick_fallible(copy_range, writer)
+                })
             } else {
                 Ok(0)
             };
             let read_len = match read_res {
                 Ok(read_len) => read_len,
-                Err(_) if read_tot_len > 0 => break aux_prev_data.as_mut().unwrap(),
+                Err(_) if read_tot_len > 0 => break,
                 Err(err) => return Err(err),
             };
+
             read_tot_len += read_len;
+            if aux_front.is_some() {
+                aux_pos += 1;
+            }
 
             // Record the current auxiliary data. Break if the read is incomplete or this is a
             // `SOCK_SEQPACKET` socket.
             if is_seqpacket {
-                aux_prev_data = Some(all_aux.pop_front().unwrap().data);
                 if read_len < aux_len {
                     warn!("setting MSG_TRUNC is not supported");
-                    reader.skip(aux_len - read_len);
+                    trunc_len = aux_len - read_len;
                 }
-                break aux_prev_data.as_mut().unwrap();
-            } else if let Some(front) = aux_front {
-                if read_len < aux_len {
-                    front.start += read_len;
-                    break &mut front.data;
-                }
-                aux_prev_data = Some(all_aux.pop_front().unwrap().data);
-            } else {
-                aux_prev_data = Some(AuxiliaryData::default());
-                if read_len < aux_len {
-                    break aux_prev_data.as_mut().unwrap();
-                }
+                break;
+            } else if read_len < aux_len {
+                break;
             }
-        };
+        }
 
+        // Consume the payload bytes that we've read.
+        if behavior.will_consume_data() {
+            let consume_tot_len = read_tot_len + trunc_len;
+            reader.commit_read(consume_tot_len);
+            if consume_tot_len > 0 {
+                self.inner.notify_read();
+            }
+        }
         drop(reader);
 
-        let ctrl_msgs = aux_data.generate_control(is_pass_cred);
-        debug_assert!(is_seqpacket || read_tot_len != 0);
-        peer_end
-            .has_aux
-            .store(!all_aux.is_empty(), Ordering::Relaxed);
+        // Consume the auxiliary data that we've read.
+        let ctrl_msgs = if aux_pos >= 1 {
+            let aux_data = all_aux.get_mut(aux_pos - 1).unwrap();
+            debug_assert!((aux_data.start - read_base).0 <= read_tot_len);
 
+            let ctrl_msgs = aux_data.data.generate_control(behavior, is_pass_cred);
+            if behavior.will_consume_data() {
+                let remaining_aux_count = all_aux.len() - (aux_pos - 1);
+                all_aux.truncate_front(remaining_aux_count);
+                let consume_len = read_tot_len + trunc_len;
+                if (all_aux.front().unwrap().end - read_base).0 <= consume_len {
+                    all_aux.pop_front();
+                } else {
+                    all_aux.front_mut().unwrap().start = read_base + Wrapping(consume_len);
+                }
+                peer_end
+                    .has_aux
+                    .store(!all_aux.is_empty(), Ordering::Relaxed);
+            }
+            ctrl_msgs
+        } else {
+            let mut default_aux_data = AuxiliaryData::default();
+            default_aux_data.generate_control(behavior, is_pass_cred)
+        };
+
+        debug_assert!(is_seqpacket || read_tot_len != 0);
         Ok((read_tot_len, ctrl_msgs))
     }
 
@@ -253,12 +301,11 @@ impl Connected {
         // Fast path: There are no auxiliary data to transmit.
         if aux_data.is_empty() && !is_seqpacket && !need_pass_cred {
             let mut writer = this_end.writer.lock();
-            return self.inner.write_with(move || {
-                if is_seqpacket && writer.free_len() < reader.sum_lens() {
-                    return Ok(0);
-                }
-                writer.write_fallible(reader)
-            });
+            let write_len = self
+                .inner
+                .write_with(move || writer.write_fallible(reader))?;
+            self.inner.notify_write();
+            return Ok(write_len);
         }
 
         let mut all_aux = this_end.all_aux.lock();
@@ -299,6 +346,8 @@ impl Connected {
             end: write_start + Wrapping(write_len),
         };
         all_aux.push_back(aux_range);
+
+        self.inner.notify_write();
 
         Ok(write_len)
     }
--- a/kernel/src/net/socket/vsock/transport/connection/recv.rs
+++ b/kernel/src/net/socket/vsock/transport/connection/recv.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
+use aster_bigtcp::socket::ReceiveBehavior;
 use aster_virtio::device::socket::{header::VirtioVsockOp, packet::RxPacket};
 
 use crate::{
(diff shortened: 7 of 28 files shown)


lrh2000:
This is great work! Currently, there seem to be some problems with the commit history that make the review a bit annoying. Could you fix them?
 - Some regression tests (`test/initramfs/src/regression/network/msg_peek.c`) are added in `Implement XXX` commit, but the entire file is refactored in the later commit (`Add MSG_PEEK socket regression tests`).
 - The first commit `Implement MSG_PEEK for TCP sockets` does not define `ReceiveBehavior`, which is introduced in the second commit.
 - The commit `Implement MSG_PEEK for UDP sockets` touches netlink code for no good reason. I suppose the changes to netlink code should belong to another commit.

Also for netlink code, this needs to be updated:
```rust
        // TODO: Deal with other flags. Only MSG_PEEK is handled here.
        if !flags.sub(SendRecvFlags::MSG_PEEK).is_all_supported() {
            warn!("unsupported flags: {:?}", flags);
        }
```

After this PR, we no longer need to check `MSG_PEEK` specially in netlink.

StevenJiang1110:
> Could you fix them?

I have cleaned up the commit history. `ReceiveBehavior` has been moved to the first commit, the netlink cleanup has been extracted into a separate commit, and the regression tests have been split into their corresponding commits with the final refactored version.

kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
```diff
@@ -445,12 +446,42 @@ impl<E: Ext> TcpConnection<E> {
...
+        mut copy_fn: CopyFn,
+        mut need_wrap_continuation: bool,
+        mut recv_fn: RecvFn,
+    ) -> Result<(NonZeroUsize, NeedIfacePoll), IoError<RecvError, CopyErr>>
+    where
+        CopyFn: FnMut(&[u8]) -> Result<usize, (CopyErr, usize)>,
+        RecvFn: FnMut(
+            &mut RawTcpSocketExt<E>,
+            &mut dyn FnMut(&[u8]) -> (usize, Result<(), CopyErr>),
+        ) -> Result<Result<(), CopyErr>, smoltcp::socket::tcp::RecvError>,
```
lrh2000: I do not really think we need this. We can finish this task equally well within the `recv` method by:
```rust
let recv_result = if will_consume_data {
    // receive logic
} else {
    // peek logic
}
```

If this will lead to code duplication and too many levels of indentation, I believe there are other, more reasonable ways to extract the logic. But I don't really think it is needed to use `FnMut(..., &mut dyn FnMut)`.

kernel/libs/aster-bigtcp/src/socket/bound/tcp_conn.rs
```diff
@@ -445,12 +446,42 @@ impl<E: Ext> TcpConnection<E> {
...
+        CopyFn: FnMut(&[u8]) -> Result<usize, (CopyErr, usize)>,
+    {
+        match behavior {
+            ReceiveBehavior::Recv => self.recv_internal(copy_fn, true, |socket, copy_fn| {
+                socket.recv(|socket_buffer| copy_fn(socket_buffer))
+            }),
+            ReceiveBehavior::Peek => self.recv_internal(copy_fn, false, |socket, copy_fn| {
+                // `socket.peek()` only returns the first contiguous readable range.
+                // We intentionally do not stitch a wrapped receive buffer together here:
+                // supporting that would need extra caching and copying.
```
lrh2000:
```suggestion
                // FIXME: This implementation relies on `socket.peek()`, which only returns the
                // first contiguous readable range. We need more powerful smoltcp APIs to peek the
                // wrapped receive buffer without extra caching and copying.
```

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -0,0 +1,242 @@
...
+static int tcp_listener;
+static struct sockaddr_in tcp_addr = { .sin_family = AF_INET };
+static socklen_t tcp_addr_len = sizeof(tcp_addr);
+static struct pollfd pfd = { .events = POLLIN };
+
+static ssize_t peek_message(int fd, char *buf, size_t len, int *msg_flags)
+{
+	struct iovec iov = { .iov_base = buf, .iov_len = len };
+	struct msghdr msg = { .msg_iov = &iov, .msg_iovlen = 1 };
+	ssize_t ret = CHECK(recvmsg(fd, &msg, MSG_PEEK));
```
lrh2000: There is no need to use `CHECK()` if the entire helper function is just a wrapper for a single system call.

You use `TEST_RES(peek_message(...))` below, so everything will work fine even if you don't `CHECK()`.

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -240,3 +240,36 @@ FN_SETUP(close_tcp_listener)
...
+
+	addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+	TEST_SUCC(bind(recv_fd, (struct sockaddr *)&addr, sizeof(addr)));
+	TEST_SUCC(getsockname(recv_fd, (struct sockaddr *)&addr, &addr_len));
+	TEST_SUCC(connect(send_fd, (struct sockaddr *)&addr, addr_len));
+
+	// Peeking a datagram must not consume the datagram.
+	TEST_RES(send(send_fd, PAYLOAD, PAYLOAD_LEN, 0), _ret == PAYLOAD_LEN);
+
+	TEST_RES(peek_message(recv_fd, buf, PAYLOAD_LEN, &msg_flags),
```
lrh2000:
```suggestion
	TEST_RES(send(send_fd, PAYLOAD, PAYLOAD_LEN, 0), _ret == PAYLOAD_LEN);

	// Peeking a datagram must not consume the datagram.
	TEST_RES(peek_message(recv_fd, buf, PAYLOAD_LEN, &msg_flags),
```

kernel/src/net/socket/vsock/transport/connection/recv.rs
```diff
@@ -50,15 +51,17 @@ impl Connection {
...
+        let result = packets.copy_to_userspace(writer, behavior);
         let recv_len = *result.as_ref().unwrap_or(&0);
 
-        self.inner
-            .state
-            .lock()
-            .ungrab_packets_and_finish_recv(&self.inner, packets, recv_len);
-
-        self.inner.pollee.invalidate();
+        let mut state = self.inner.state.lock();
```
lrh2000: There seem to be no need to do this. We can continue to lock the `state` when calling its methods?

kernel/src/net/socket/unix/datagram/message.rs
```diff
@@ -180,18 +181,25 @@ impl MessageReceiver {
...
-        inner.total_length -= msg.bytes.len();
-
         let is_pass_cred = self.queue.is_pass_cred.load(Ordering::Relaxed);
-        let ctrl_msgs = msg.aux.generate_control(is_pass_cred);
-
-        self.queue.pollee.invalidate();
-        // A writer may still fail if the free space is not enough.
-        // So we have to wake up all the writers here.
-        self.queue.send_wait_queue.wake_all();
+        let msg = msg.src.clone();
```
lrh2000: Why do you name a `src: UnixSocketAddr` variable as `msg`... This is very confusing.

kernel/src/net/socket/unix/ctrl_msg.rs
```diff
@@ -273,10 +274,17 @@ impl AuxiliaryData {
...
+    pub(super) fn generate_control(
+        &mut self,
+        behavior: ReceiveBehavior,
+        is_pass_cred: bool,
+    ) -> Vec<ControlMessage> {
         let Self { files, cred } = self;
+        let files = match behavior {
+            ReceiveBehavior::Recv => mem::take(files),
+            ReceiveBehavior::Peek => files.clone(),
+        };
```
lrh2000: I don't believe we should do this here. This should be done in the `files.is_empty()` branch?

In addition, minimize the diff by keeping the original code layout:
```rust
        let mut ctrl_msgs = Vec::new();

        let Self { files, cred } = self;

        if ... {}
```

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -387,4 +423,94 @@ struct RangedAuxiliaryData {
...
+    read_len: usize,
+    selected_aux_data_index: Option<usize>,
+}
+
+/// Returns how many bytes and which auxiliary data a `MSG_PEEK` read should expose.
+///
+/// The returned length is the largest prefix that fits in `writer_len` and remains compatible with
+/// the control messages at the current read position. For `SOCK_SEQPACKET`, the returned length is
+/// further capped to the first packet.
+fn peek_read_plan(
```
lrh2000: https://github.com/asterinas/asterinas/pull/3288 proposes to add `pick*` APIs to the ring buffer. With it, there is a possibility to avoid duplication of the receive loop.

```rust
let read_len = loop {
   // Use pick APIs here.
};

if will_consume_data {
    ring_buffer.commit_read();
    aux_data.truncate_front();
}
```

Give this a thought?

kernel/src/fs/utils/endpoint.rs
```diff
@@ -130,6 +131,29 @@ impl<T: AsRef<EndpointState>> Endpoint<T> {
...
+
+    /// Reads from the endpoint without updating the local/remote [`Pollee`]s.
+    ///
+    /// The `read` operation may either consume data or inspect it without consumption. Note that if
+    /// `read` returns `Ok(0)`, it is assumed that there is no data to read and an [`Errno::EAGAIN`]
+    /// error will be returned instead.
+    ///
+    /// However, if the remote endpoint has shut down, this method will return `Ok(0)` to indicate
+    /// the end-of-file (EOF).
+    pub fn read_without_pollee_update<F>(&self, read: F) -> Result<usize>
```
lrh2000: Another option is to add `notify_read`/`notify_write` methods... Which is better?

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -0,0 +1,242 @@
...
+		      tcp_addr_len));
+	accepted_fd = CHECK(accept(tcp_listener, NULL, NULL));
+	status_flags = CHECK(fcntl(accepted_fd, F_GETFL, 0));
+	CHECK(fcntl(accepted_fd, F_SETFL, status_flags | O_NONBLOCK));
+
+	*send_fd = connected_fd;
+	*recv_fd = accepted_fd;
+}
+
+FN_TEST(tcp_short_peek_short_reads)
```
lrh2000: There are many TCP tests, which seem hard to follow.

With some common variable names, we can possibly simply it to:
```c
int sk;
int read_pos = 0;
int write_pos = 0;

INIT_TCP();

SEND(PAYLOAD_LEN);
PEEK(PAYLOAD_LEN / 2);
RECV(PAYLOAD_LEN / 2);
RECV(PAYLOAD_LEN / 2);

FINI_TCP();
```

It would be even better if we could make this work with other socket types. However, there may be some complexities with control messages.

lrh2000 (commented):
Thanks for the excellent work. Adding support for more `SendRecvFlags` was one of the first things I heard about when I started this project (https://github.com/asterinas/asterinas/issues/455). However, it seems that we have made zero progress in this area in the past three years. I didn't expect a dummy `SendRecvFlags` to work for that long...

kernel/src/net/socket/unix/stream/connected.rs
StevenJiang1110: `pick_range` sounds like a more powerful API than `peek*`, since it can access any range within the buffer rather than only from the read head. I therefore think using `pick_range` here is well-suited. However, since #3288 does not appear likely to be merged in the near future, I have cherry-picked the ring buffer-related commit here and refactored on top of it.

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -135,90 +137,153 @@ impl Connected {
...
-                } else {
-                    ((front.start - read_start).0, None)
+        let mut read_tot_len = 0;
+        let mut consume_tot_len = 0;
+        let read_start = reader.head();
+        let mut cursor = read_start;
+        let mut aux_index = 0;
+        let mut has_prev_aux_data = false;
+        let mut prev_aux_index = None;
+        let mut selected_aux_index = None;
```
lrh2000: I don't think we need such a complex logic for the UNIX stream socket, and I found no valid reason for it after reading the code.

Could you please explain why the following simpler solution won't work?
```patch
diff --git a/kernel/src/net/socket/unix/stream/connected.rs b/kernel/src/net/socket/unix/stream/connected.rs
index 4b4177278..8fb129ea5 100644
--- a/kernel/src/net/socket/unix/stream/connected.rs
+++ b/kernel/src/net/socket/unix/stream/connected.rs
@@ -150,13 +150,17 @@ impl Connected {
         }
 
         let mut all_aux = peer_end.all_aux.lock();
-        let mut aux_prev_data: Option<AuxiliaryData> = None;
+        let mut aux_pos = 0;
+
+        let read_base = reader.head();
         let mut read_tot_len = 0;
+        let mut trunc_len = 0;
 
-        let aux_data = loop {
-            let read_start = reader.head();
-            let (aux_len, aux_front) = if let Some(front) = all_aux.front_mut() {
+        loop {
+            let read_start = reader_base + read_tot_len;
+            let (aux_len, aux_front) = if let Some(front) = all_aux.get(aux_pos) {
                 if front.start == read_start {
+                    aux_pos += 1;
                     ((front.end - read_start).0, Some(front))
                 } else {
                     ((front.start - read_start).0, None)
@@ -167,7 +171,8 @@ impl Connected {
 
             // Unless the auxiliary data we have already received is a subset of the current
             // auxiliary data, we cannot receive additional bytes.
-            if let Some(prev) = aux_prev_data.as_mut() {
+            if aux_pos >= 2 {
+                let prev = all_aux.get(prev_idx - 2).unwrap();
                 let is_subset = if let Some(front) = aux_front.as_ref() {
                     prev.is_subset_of(&front.data, is_pass_cred)
                 } else {
@@ -180,14 +185,16 @@ impl Connected {
 
             // Read the payload bytes of the current auxiliary data.
             let read_res = if !is_empty && aux_len > 0 {
-                self.inner
-                    .read_with(|| reader.read_fallible_with_max_len(writer, aux_len))
+                todo!("pick from `read_start` with `aux_len`");
             } else {
                 Ok(0)
             };
             let read_len = match read_res {
                 Ok(read_len) => read_len,
-                Err(_) if read_tot_len > 0 => break aux_prev_data.as_mut().unwrap(),
+                Err(_) if read_tot_len > 0 => {
+                    aux_pos = aux_pos.wrapping_sub(1);
+                    break;
+                }
                 Err(err) => return Err(err),
             };
             read_tot_len += read_len;
@@ -195,34 +202,48 @@ impl Connected {
             // Record the current auxiliary data. Break if the read is incomplete or this is a
             // `SOCK_SEQPACKET` socket.
             if is_seqpacket {
-                aux_prev_data = Some(all_aux.pop_front().unwrap().data);
                 if read_len < aux_len {
                     warn!("setting MSG_TRUNC is not supported");
-                    reader.skip(aux_len - read_len);
-                }
-                break aux_prev_data.as_mut().unwrap();
-            } else if let Some(front) = aux_front {
-                if read_len < aux_len {
-                    front.start += read_len;
-                    break &mut front.data;
+                    trunc_len = aux_len - read_len;
                 }
-                aux_prev_data = Some(all_aux.pop_front().unwrap().data);
+                break;
             } else {
-                aux_prev_data = Some(AuxiliaryData::default());
                 if read_len < aux_len {
-                    break aux_prev_data.as_mut().unwrap();
+                    break;
                 }
             }
-        };
+        }
 
+        // Consume the payload bytes that we've read.
+        if will_consume {
+            reader.commit_read(read_tot_len + trunc_len);
+            if read_tot_len + trunc_len > 0 {
+                self.inner.notify_read();
+            }
+        }
         drop(reader);
 
-        let ctrl_msgs = aux_data.generate_control(is_pass_cred);
-        debug_assert!(is_seqpacket || read_tot_len != 0);
+        // Consume the auxiliary data that we've read.
+        let ctrl_msgs = if aux_pos >= 1
+            && let Some(aux_data) = all_aux.get(aux_pos - 1)
+            && aux_data.start - read_base >= read_tot_len
+        {
+            aux_data.generate_control();
+            if will_consume {
+                if aux_data.end - read_base <= read_tot_len + trunc_len {
+                    all_aux.truncate_front(aux_pos);
+                } else {
+                    all_aux.truncate_front(aux_pos - 1);
+                }
+            }
+        } else {
+            todo!("default path")
+        };
         peer_end
             .has_aux
             .store(!all_aux.is_empty(), Ordering::Relaxed);
 
+        debug_assert!(is_seqpacket || read_tot_len != 0);
         Ok((read_tot_len, ctrl_msgs))
     }
 
```

kernel/src/net/socket/unix/datagram/message.rs
```diff
@@ -180,18 +181,25 @@ impl MessageReceiver {
...
+        let behavior = flags.receive_behavior();
+        let ctrl_msgs = if behavior.will_consume_data() {
+            let mut message = inner.messages.pop_front().unwrap();
+            inner.total_length -= message.bytes.len();
+
+            self.queue.pollee.invalidate();
+            // A writer may still fail if the free space is not enough.
+            // So we have to wake up all the writers here.
+            self.queue.send_wait_queue.wake_all();
+            message.aux.generate_control(behavior, is_pass_cred)
```
lrh2000, line 197: Add a blank line below `wake_all`.

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -425,3 +425,46 @@ FN_TEST(unix_seqpacket_msg_peek_with_scm_rights)
...
 END_TEST()
 
+FN_TEST(unix_dgram_msg_peek)
+{
+	int fds[2] = { -1, -1 };
+	int msg_flags = 0;
+	char buf[PAYLOAD_LEN] = {};
+
+	TEST_SUCC(socketpair(AF_UNIX, SOCK_DGRAM | SOCK_NONBLOCK, 0, fds));
+	TEST_RES(send(fds[0], PAYLOAD, PAYLOAD_LEN, 0), _ret == PAYLOAD_LEN);
```
lrh2000: Add a blank line before `send()`. This comment applies to many tests for UNIX datagram/raw/stream/seqpacket sockets.

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -233,3 +264,164 @@ FN_TEST(udp_msg_peek)
...
+
+	TEST_SUCC(socketpair(AF_UNIX, SOCK_SEQPACKET | SOCK_NONBLOCK, 0, fds));
+	TEST_SUCC(pipe(pipe_fds));
+
+	chdr->cmsg_level = SOL_SOCKET;
+	chdr->cmsg_type = SCM_RIGHTS;
+	chdr->cmsg_len = CMSG_LEN(sizeof(int));
+	cdata = (int *)CMSG_DATA(chdr);
+	cdata[0] = pipe_fds[0];
+
```
lrh2000: Remove this blank line. The logic prepare to the next `send`, right?

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -233,3 +264,164 @@ FN_TEST(udp_msg_peek)
...
+	int msg_flags = 0;
+	char buf[PAYLOAD_LEN] = {};
+
+	SOCKETPAIR_CONNECT(SOCK_STREAM);
+	SOCKETPAIR_SEND(0, PAYLOAD_LEN);
+	SOCKETPAIR_PEEK(0, SHORT_LEN);
+	SOCKETPAIR_RECV(0, PAYLOAD_LEN);
+	SOCKETPAIR_CLOSE();
+}
+END_TEST()
```
lrh2000: So the defined macros are only used once...

In fact, the TCP tests and the UNIX stream tests should work for each other? I am not sure if this idea works:
```c
#define SEND TCP_SEND
#define RECV TCP_RECV
#include "msg_peek_streamlike.h"
#undef SEND TCP_SEND
#undef RECV TCP_RECV

#define SEND UNIX_STREAM_SEND
#define RECV UNIX_STREAM_RECV
#include "msg_peek_streamlike.h"
#undef SEND UNIX_STREAM_SEND
#undef RECV UNIX_STREAM_RECV
```

kernel/src/util/ring_buffer.rs
```diff
@@ -92,29 +136,10 @@ impl<R: Deref<Target = RingBuffer<u8>>> ConsumerU8Ext for Consumer<u8, R> {
         max_len: usize,
     ) -> Result<usize> {
         let len = self.len().min(max_len);
-
         let head = self.head();
-        let offset = head.0 & (self.capacity() - 1);
-
-        let read_len = if offset + len > self.capacity() {
-            // Read from two separate parts
```
lrh2000, line 100: Add this comment back.

kernel/src/util/ring_buffer.rs
```diff
@@ -64,6 +67,47 @@ impl<R: Deref<Target = RingBuffer<u8>>> ProducerU8Ext for Producer<u8, R> {
...
+    ) -> Result<usize>;
+}
+
+impl RingBufferU8Ext for RingBuffer<u8> {
+    fn pick_fallible(
+        &self,
+        range: Range<Wrapping<usize>>,
+        writer: &mut dyn MultiWrite,
+    ) -> Result<usize> {
+        let len = (range.end - range.start).0.min(writer.sum_lens());
```
lrh2000: It does not seem like we need to call `sum_lens()`. We don't do this previously?

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -233,3 +264,164 @@ FN_TEST(udp_msg_peek)
...
+
+	memset(buf, 0, sizeof(buf));
+	TEST_RES(recv(fds[1], buf, sizeof(buf), 0),
+		 _ret == PAYLOAD_LEN && memcmp(buf, PAYLOAD, PAYLOAD_LEN) == 0);
+
+	memset(buf, 0, sizeof(buf));
+	TEST_RES(recv(fds[1], buf, sizeof(buf), 0),
+		 _ret == PAYLOAD_LEN - SHORT_LEN &&
+			 memcmp(buf, PAYLOAD + SHORT_LEN,
+				PAYLOAD_LEN - SHORT_LEN) == 0);
```
lrh2000: This seems to be covered by `SOCKETPAIR_*` macros?

lrh2000 (commented):
Most other code seems good.

kernel/src/net/socket/unix/stream/connected.rs
StevenJiang1110: The suggested approach is indeed much simpler — I have refactored the code accordingly.

I did run into two minor issues when running the tests, both related to the `is_subset` check, so I have adjusted the condition slightly to address them.

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,154 @@ impl Connected {
...
+            let consume_len = if (aux_data.end - read_base).0 <= read_tot_len + trunc_len {
+                read_tot_len + trunc_len
+            } else {
+                read_tot_len
+            };
+            let ctrl_msgs = aux_data.data.generate_control(behavior, is_pass_cred);
+            if behavior.will_consume_data() {
+                truncate_aux_data_front(&mut all_aux, read_base, consume_len);
+            }
+            ctrl_msgs
```
lrh2000: I am not sure why you don't use `VecDeque::truncate_front` based on `aux_pos`?

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,154 @@ impl Connected {
...
             // auxiliary data, we cannot receive additional bytes.
-            if let Some(prev) = aux_prev_data.as_mut() {
+            if read_tot_len > 0 {
                 let is_subset = if let Some(front) = aux_front.as_ref() {
-                    prev.is_subset_of(&front.data, is_pass_cred)
+                    if aux_pos >= 2 {
+                        let prev = all_aux.get(aux_pos - 2).unwrap();
+                        prev.data.is_subset_of(&front.data, is_pass_cred)
+                    } else {
+                        AuxiliaryData::default().is_subset_of(&front.data, is_pass_cred)
```
lrh2000, line 196: This is always true, right?

My previous suggestion is just a prototype. It can be revised, for example, if you remove `aux_pos += 1` in:
```rust
                if front.start == read_start {
                    aux_pos += 1;
```

Then here we just need to write `if read_tot_len > 0 && aux_pos >= 1 {` and check `front` against `all_aux.get(aux_pos - 1)`.

Other logic like the following should also be removed:
```rust
                    if aux_front.is_some() {
                        aux_pos -= 1;
                    }
```
```rust
                    aux_pos = aux_pos.wrapping_sub(1);
```

(In fact, `wrapping_sub` is a typo in my previous suggestion. Wrapping the index to `usize::MAX` makes absolutely no sense. It seems you didn't identify this problem, though.)

It is better to do the following at the end of the loop:
```rust
if aux_front.is_some() {
    aux_pos += 1;
}
```

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -29,48 +29,60 @@ static struct pollfd pfd = { .events = POLLIN };
...
-	do {                               \
-		TEST_SUCC(close(send_fd)); \
-		TEST_SUCC(close(recv_fd)); \
-	} while (0)
-
-#define TCP_SEND(offset, len)                                 \
-	TEST_RES(send(send_fd, PAYLOAD + (offset), (len), 0), \
-		 _ret == (ssize_t)(len))
-
 #define TCP_WAIT_READABLE()                          \
```
lrh2000: You're refactoring TCP regression tests significantly in the UNIX stream commit.

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -299,6 +353,9 @@ impl Connected {
             end: write_start + Wrapping(write_len),
         };
         all_aux.push_back(aux_range);
+        if write_len > 0 || is_seqpacket {
```
lrh2000: Make a separate code block.

Also you don't seem to check `write_len > 0` in all other scenarios. This seems to be the only exception?

lrh2000 (commented):
Thanks for the revision.

kernel/libs/ring-buffer/src/lib.rs
```diff
@@ -217,6 +224,58 @@ impl<T> RingBuffer<T> {
 }
 
 impl RingBuffer<u8> {
+    /// Returns a formatter that appends text directly to the ring buffer.
+    ///
+    /// The formatter discards the oldest buffered bytes when new formatted
+    /// bytes exceed the available capacity.
+    pub fn formatter(&mut self) -> RingBufferFormatter<'_> {
```
lrh2000: Let's not add an unrelated, unused API here.

I suggest removing the code related to the `RingBufferFormatter` and squashing the first two commits. You can co-author that commit.

kernel/src/net/socket/unix/stream/connected.rs
StevenJiang1110: I revised the index handling as suggested: `aux_pos` is no longer advanced when the current aux entry is found.

> This is always true, right?

Actually, I found a corner case when running  gVisor's `CredsNotCoalescedUp`: it first writes 20 bytes before enabling `SO_PASSCRED`, then enables `SO_PASSCRED` and writes another 20 bytes. The first 20 bytes have default auxiliary data and should produce default credentials. The second 20 bytes have real credentials.

At the boundary after reading the first 20 bytes, the loop state is:

```text
read_tot_len = 20
aux_pos = 0
aux_front = Some(...)
```

If we use:

```rust
if read_tot_len > 0 && aux_pos >= 1 {
    ...
}
```

then `aux_pos >= 1` is false, so the subset check is skipped. The loop then continues to read the second 20-byte chunk too, causing the first `recvmsg` to return 40 bytes with the real credentials. But the test expects the first `recvmsg` to return only the first 20 bytes with default credentials. This is why `CredsNotCoalescedUp/*` fails with that exact condition.

kernel/libs/ring-buffer/src/lib.rs
```diff
@@ -217,6 +223,50 @@ impl<T> RingBuffer<T> {
 }
 
 impl RingBuffer<u8> {
+    /// Picks bytes from the specified absolute ring counter range into `writer`.
+    ///
+    /// This method does not consume bytes from the buffer. The caller must
+    /// ensure that `range` is currently readable. The method handles wraparound
+    /// internally and advances `writer` by the number of bytes copied.
+    pub fn pick_range(
```
lrh2000: Sorry, but it also seems that this API is (also) not used in this PR. I'm confused as to why you cherry-picked the commit from the `syslog` PR. The commit seems entirely unrelated here because we're using `RingBufferU8Ext::pick_fallible`.

test/initramfs/src/regression/network/msg_peek_streamlike.h
```diff
@@ -0,0 +1,139 @@
...
+	PEEK(0, SHORT_LEN);
+
+	SEND(SHORT_LEN, PAYLOAD_LEN - SHORT_LEN);
+	WAIT_APPENDED_READABLE();
+	RECV(0, PAYLOAD_LEN);
+	CLOSE();
+}
+END_TEST()
+
+DEFINE_TEST(full_peek_short_reads)
```
lrh2000: Move this test below `short_peek_full_read`.

It is not related to the `append_*` tests, right?

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,159 @@ impl Connected {
...
                 } else {
                     ((front.start - read_start).0, None)
                 }
             } else {
-                (usize::MAX, None)
+                ((reader.tail() - read_start).0, None)
             };
+            if aux_len == 0 && aux_front.is_none() {
+                break;
+            }
```
lrh2000: Is this correct? We need to return `EAGAIN` if the buffer is empty. It was previously handled by `read_with`. But now this will break.... It will hit:
```rust
        debug_assert!(is_seqpacket || read_tot_len != 0);
```

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,159 @@ impl Connected {
...
+                })
             } else {
                 Ok(0)
             };
             let read_len = match read_res {
                 Ok(read_len) => read_len,
-                Err(_) if read_tot_len > 0 => break aux_prev_data.as_mut().unwrap(),
+                Err(_) if read_tot_len > 0 => {
+                    break;
+                }
```
lrh2000:
```suggestion
                Err(_) if read_tot_len > 0 => break,
```

If applicable (otherwise ignore this comment).

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,159 @@ impl Connected {
...
+            && (aux_data.start - read_base).0 <= read_tot_len
+        {
+            let consume_len = if (aux_data.end - read_base).0 <= read_tot_len + trunc_len {
+                read_tot_len + trunc_len
             } else {
-                aux_prev_data = Some(AuxiliaryData::default());
-                if read_len < aux_len {
-                    break aux_prev_data.as_mut().unwrap();
+                read_tot_len
+            };
```
lrh2000: What is the purpose of this check? `consume_len` should always be `read_tot_len + trunc_len`.

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -126,3 +165,148 @@ FN_TEST(udp_msg_peek)
...
+
+	SOCKETPAIR_SEND(0, PAYLOAD_LEN);
+
+	SOCKETPAIR_PEEK(0, PAYLOAD_LEN);
+
+	SOCKETPAIR_RECV(0, PAYLOAD_LEN);
+
+	memset(buf, 0, sizeof(buf));
+	SOCKETPAIR_SEND(0, PAYLOAD_LEN);
+	SOCKETPAIR_SEND(SHORT_LEN, PAYLOAD_LEN - SHORT_LEN);
```
lrh2000: Why do you insert `memset(buf)` here. I think it is unrelated to `SEND`. Do you mean to have it for `peek_message`?
```c
	SOCKETPAIR_SEND(0, PAYLOAD_LEN);
	SOCKETPAIR_SEND(SHORT_LEN, PAYLOAD_LEN - SHORT_LEN);

	memset(buf, 0, sizeof(buf));
	msg_flags = 0;
	TEST_RES(peek_message(fds[1], buf, sizeof(buf), &msg_flags),
		 _ret == PAYLOAD_LEN && (msg_flags & MSG_TRUNC) == 0 &&
			 memcmp(buf, PAYLOAD, PAYLOAD_LEN) == 0);
```

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -126,3 +165,148 @@ FN_TEST(udp_msg_peek)
...
+
+	msg_flags = 0;
+	TEST_RES(peek_message(fds[1], buf, sizeof(buf), &msg_flags),
+		 _ret == PAYLOAD_LEN && (msg_flags & MSG_TRUNC) == 0 &&
+			 memcmp(buf, PAYLOAD, PAYLOAD_LEN) == 0);
+
+	SOCKETPAIR_RECV(0, PAYLOAD_LEN);
+
+	SOCKETPAIR_RECV(SHORT_LEN, PAYLOAD_LEN - SHORT_LEN);
```
lrh2000, line 230: Remove this blank line 242 since you don't have a blank line when calling `SEND`.

kernel/src/net/socket/unix/stream/connected.rs
lrh2000: Thanks for your detailed explanation. The current behavior should be correct.

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,159 @@ impl Connected {
...
+            if consume_tot_len > 0 {
+                self.inner.notify_read();
+            }
+        }
+        drop(reader);
+
+        // Consume the auxiliary data that we've read.
+        let ctrl_msgs = if aux_pos >= 1
+            && let Some(aux_data) = all_aux.get_mut(aux_pos - 1)
+            && (aux_data.start - read_base).0 <= read_tot_len
```
lrh2000: If `aux_pos >= 1`, it means that we've already consumed some aux data (`front.is_some()`). It seems that checks in line 257 and line 258 will always evaluate to true.

lrh2000 (commented):
Thanks for updating the code!

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,155 @@ impl Connected {
...
                 } else {
                     ((front.start - read_start).0, None)
                 }
             } else {
-                (usize::MAX, None)
+                ((reader.tail() - read_start).0, None)
             };
+            if aux_len == 0 && aux_front.is_none() {
+                if !is_seqpacket && read_tot_len == 0 {
+                    self.inner.read_with(|| Ok(0))?;
```
lrh2000: I mean that this case can be handled naturally in the following code:
```rust
                self.inner.read_with(|| {
                    let copy_range = read_start..read_start + Wrapping(aux_len);
                    reader.ring_buffer().pick_fallible(copy_range, writer)
                })
```

All corner cases, including `is_seqpacket` and `read_tot_len == 0`, are handled correctly below.

So why did you change the code like this? What problems will be caused by removing these `if`s?

StevenJiang1110: Sorry, I don't quite follow your point.

As I understand it, you are saying that we should return EAGAIN when there is no data available to read. That is why I added an extra `read_with` call before the `break` to ensure `EAGAIN` can be returned. But if we handle it at the location you suggested, wouldn't the `break` above still cause us to exit the loop, making the subsequent `read_with` call unreachable?

StevenJiang1110: I understood it. I removed the fast break path. You can check the latest force push diff about it.

```diff
diff --git a/kernel/src/net/socket/unix/stream/connected.rs b/kernel/src/net/socket/unix/stream/connected.rs

index 26758a6bc..eba5b7a74 100644
--- a/kernel/src/net/socket/unix/stream/connected.rs
+++ b/kernel/src/net/socket/unix/stream/connected.rs
@@ -180,12 +180,6 @@ impl Connected {
             } else {
                 ((reader.tail() - read_start).0, None)
             };
-            if aux_len == 0 && aux_front.is_none() {
-                if !is_seqpacket && read_tot_len == 0 {
-                    self.inner.read_with(|| Ok(0))?;
-                }
-                break;
-            }
 
             // Unless the auxiliary data we have already received is a subset of the current
             // auxiliary data, we cannot receive additional bytes.
@@ -211,7 +205,7 @@ impl Connected {
             }
 
             // Read the payload bytes of the current auxiliary data.
-            let read_res = if !is_empty && aux_len > 0 {
+            let read_res = if !is_empty && (aux_len > 0 || aux_front.is_none()) {
                 self.inner.read_with(|| {
                     let copy_range = read_start..read_start + Wrapping(aux_len);
                     reader.ring_buffer().pick_fallible(copy_range, writer)
```

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,149 @@ impl Connected {
...
+            }
+        }
         drop(reader);
 
-        let ctrl_msgs = aux_data.generate_control(is_pass_cred);
-        debug_assert!(is_seqpacket || read_tot_len != 0);
+        // Consume the auxiliary data that we've read.
+        let ctrl_msgs = if aux_pos >= 1 {
+            debug_assert!((all_aux[aux_pos - 1].start - read_base).0 <= read_tot_len);
+            let aux_data = &mut all_aux[aux_pos - 1];
```
lrh2000:
```suggestion
            let aux_data = all_aux.get_mut(aux_pos - 1).unwrap();
            debug_assert!((aux_data.start - read_base).0 <= read_tot_len);
            
```

(I am using `get_mut` because the rest of the code uses `get().unwrap()`. You can prefer to use `[]` as well, just be consistent.)

Also add a blank line below to avoid making this code block too large.

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,149 @@ impl Connected {
...
+            let ctrl_msgs = aux_data.data.generate_control(behavior, is_pass_cred);
+            if behavior.will_consume_data() {
+                let remaining_aux_count = all_aux.len() - (aux_pos - 1);
+                let consume_len = read_tot_len + trunc_len;
+                all_aux.truncate_front(remaining_aux_count);
```
lrh2000: Reorder. This variable should only be used in the next line after reordering.
```suggestion
                all_aux.truncate_front(remaining_aux_count);
                let consume_len = read_tot_len + trunc_len;
```

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,149 @@ impl Connected {
         let this_end = self.inner.this_end();
         let peer_end = self.inner.peer_end();
 
-        let mut reader = this_end.reader.lock();
+        let reader = this_end.reader.lock();
```
lrh2000: `commit_read` should take `&mut self`. It must not be called in a concurrent way.

It seems like an existing bug, but let's fix it since this PR touches the ring buffer code anyway?

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,149 @@ impl Connected {
...
+            }
+            ctrl_msgs
+        } else {
+            let mut default_aux_data = AuxiliaryData::default();
+            default_aux_data.generate_control(behavior, is_pass_cred)
+        };
+
         peer_end
             .has_aux
             .store(!all_aux.is_empty(), Ordering::Relaxed);
```
lrh2000: Better to be moved after truncating the queue:
```rust
                if (all_aux.front().unwrap().end - read_base).0 <= consume_len {
                    all_aux.pop_front();
                } else {
                    all_aux.front_mut().unwrap().start = read_base + Wrapping(consume_len);
                }
                peer_end
                    .has_aux
                    .store(!all_aux.is_empty(), Ordering::Relaxed);
```

kernel/src/net/socket/unix/stream/connected.rs
```diff
@@ -129,100 +130,149 @@ impl Connected {
...
-        let mut aux_prev_data: Option<AuxiliaryData> = None;
+
+        let mut aux_pos = 0;
+
+        let read_base = reader.head();
         let mut read_tot_len = 0;
+        let mut trunc_len = 0;
+
+        loop {
+            let read_start = read_base + Wrapping(read_tot_len);
```
lrh2000: Remove the blank line 164 and line 173, which do not exist before this PR and it seems not very useful to add them.

test/initramfs/src/regression/network/msg_peek_stream.h
```diff
@@ -0,0 +1,134 @@
...
+
+	CONNECT();
+
+	SEND(0, SHORT_LEN);
+	PEEK(0, SHORT_LEN);
+
+	SEND(SHORT_LEN, PAYLOAD_LEN - SHORT_LEN);
+	WAIT_APPENDED_READABLE();
+	RECV(0, PAYLOAD_LEN);
+	CLOSE();
```
lrh2000, line 126: Add a blank line before `CLOSE()` for all tests in `test/initramfs/src/regression/network/msg_peek_streamlike.h`.

`CONNECT()` lives in a separate code block anyway.

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -331,3 +331,73 @@ FN_TEST(unix_raw_msg_peek)
...
+			.nlmsg_type = RTM_GETLINK,
+			.nlmsg_flags = NLM_F_REQUEST,
+			.nlmsg_seq = seq,
+		},
+		.info = {
+			.ifi_family = AF_UNSPEC,
+			.ifi_change = 0xffffffff,
+		},
+	};
+	return send(fd, &req, sizeof(req), 0) == sizeof(req) ? 0 : -1;
```
lrh2000, line 352: `TEST_*` marcos work based on `errno`. If you return `-1` without setting an `errno`, use:
```diff
-	TEST_SUCC(send_getlink_request(fd, 1));
+	TEST_RES(send_getlink_request(fd, 1), _ret == 0);
```

kernel/src/net/socket/unix/stream/connected.rs
lrh2000: Sorry. My words may have been too confusing. I also admit that I missed some of the complexities. But your final update looks good!

lrh2000 (commented):
Here are a few more minor comments. I hope these are my last comments. Thanks for continuing to work on this.

test/initramfs/src/regression/network/msg_peek.c
```diff
@@ -0,0 +1,82 @@
...
+	accepted_fd = CHECK(accept(tcp_listener, NULL, NULL));
+
+	*send_fd = connected_fd;
+	*recv_fd = accepted_fd;
+}
+
+#define PREFIX tcp_
+#define CONNECT() TCP_CONNECT()
+#define WAIT_APPENDED_READABLE() TCP_WAIT_APPENDED_READABLE()
+#include "msg_peek_streamlike.h"
```
lrh2000: I just realized that `test/initramfs/src/regression/network/unix_streamlike_prologue.h` uses "streamlike" to denote that `SOCK_STREAM` and `SOCK_SEQPACKET`.

To avoid confusion, maybe we shouldn't use `streamlike` here. Just use `stream`?

lrh2000 (approved):

cqs21 (approved):

taosue (approved):

PR #3348, opened 2026-06-09
Introduce an atomic symlink creation API

### LINKED ISSUE #3312: Make symlink creation atomic by adding an inode symlink operation
### Current Feature
The current symbolic link creation path is implemented as a two-stage operation:

  1. Create a new symlink inode/dentry with `new_fs_child(..., InodeType::SymLink, ...)`.
  2. Write the symlink target with `write_link(target)`.

 This means symlink creation is not modeled as an atomic filesystem operation. If the first step succeeds but `write_link()` fails, the partially-created symlink may remain visible.

 For ext2, this failure window exists today, although it is relatively hard to trigger from normal user space. For example, it requires the symlink target to take the slow-symlink path and then fail during data block allocation or I/O. Short fast symlinks are stored inline, and overly long targetsare rejected before inode creation, so the bug is not easy to hit with ordinary inputs. However, the operation is still structurally fragile.

### Enhancement Description
  Symlink creation should be represented as an atomic inode operation instead of being emulated by create-then-write.

  ```rust
  fn symlink(&self, name: &str, target: &str, mode: InodeMode) -> Result<Arc<dyn Inode>>;
```
  Then `sys_symlinkat` should call the parent directory inode’s symlink operation directly.

### Motivation

Each filesystem can implement the operation atomically:

  - ext2 can create the inode, write the target, and roll back the directory entry/inode if any step fails.
  - virtiofs can map this directly to the FUSE/virtiofs symlink request, which already receives both the link name and target.
  - filesystems without symlink support can return EOPNOTSUPP or an appropriate filesystem-specific error.

  This removes the VFS-level partial-create failure window and lets each filesystem preserve its own consistency rules.

### PR DESCRIPTION
**Pull request by cchanging**
Fixes #3312

This PR introduces a `symlink` API to the `Inode` trait for atomically creating symlinks, replacing the functionality previously provided by the `write_link` API.

### DIFF
--- a/kernel/core/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/core/src/fs/fs_impls/ramfs/fs.rs
@@ -203,8 +203,8 @@ impl Inner {
         Self::File(Mutex::new(PageCache::new_anon(0).unwrap()))
     }
 
-    pub(self) fn new_symlink() -> Self {
-        Self::SymLink(SpinLock::new(String::from("")))
+    pub(self) fn new_symlink(target: &str) -> Self {
+        Self::SymLink(SpinLock::new(String::from(target)))
     }
 
     pub(self) fn new_block_device(dev_id: u64) -> Self {
@@ -622,11 +622,16 @@ impl RamInode {
         mode: InodeMode,
         uid: Uid,
         gid: Gid,
+        target: &str,
         to_be_revalidated: ToBeRevalidated,
     ) -> Arc<Self> {
         Arc::new_cyclic(|weak_self| RamInode {
-            inner: Inner::new_symlink(),
-            metadata: SpinLock::new(InodeMeta::new(mode, uid, gid)),
+            inner: Inner::new_symlink(target),
+            metadata: SpinLock::new({
+                let mut metadata = InodeMeta::new(mode, uid, gid);
+                metadata.size = target.len();
+                metadata
+            }),
             ino: fs.alloc_id(),
             typ: InodeType::SymLink,
             dir_revalidation_policy: RevalidationPolicy::empty(),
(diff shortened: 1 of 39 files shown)

### CODE REVIEW

File: kernel/core/src/fs/vfs/path/mod.rs
Comment (tatetian) [inline kernel/core/src/fs/vfs/path/mod.rs]: This is a separate issue, but I am bother by the names `new_fs_root` and `new_fs_child`. I think the `fs` part is redundant. They should simply be `new_root` and `new_child`.

File: kernel/core/src/fs/vfs/fs_apis/inode.rs [hunk 1]
```diff
@@ -420,6 +420,10 @@ pub(crate) trait Inode: Any + FileOps + Send + Sync {
         Err(Error::new(Errno::ENOTDIR))
     }
 
+    fn symlink(&self, name: &str, target: &str, mode: InodeMode) -> Result<Arc<dyn Inode>> {
```
tatetian: This name reads like a reader method and is against our practice of beginning a method name with a verb. I suggest renaming this method to `create_symlink`.

kernel/core/src/fs/fs_impls/ramfs/fs.rs
```diff
@@ -907,9 +913,16 @@ impl RamInode {
...
             InodeType::File => RamInode::new_file(&fs, mode, uid, gid, to_be_revalidated),
-            InodeType::SymLink => RamInode::new_symlink(&fs, mode, uid, gid, to_be_revalidated),
             InodeType::Socket => RamInode::new_socket(&fs, mode, uid, gid, to_be_revalidated),
             InodeType::Dir => RamInode::new_dir(&fs, mode, uid, gid, &self.this, to_be_revalidated),
+            InodeType::SymLink => RamInode::new_symlink(
+                &fs,
+                mode,
+                uid,
+                gid,
+                symlink_target.expect("a symlink target must be provided"),
```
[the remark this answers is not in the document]
cchanging: This panic is unreachable now. If it is reached, it indicates a problem in the usages of this API, and the panic can alert us to fix it.

tatetian (approved):

PR #3350, opened 2026-06-09
Implement NVMe PRP list support for multi-page I/O

Cautreoxit:
This PR is an improvement on NVMe driver. Previously, the NVMe block driver only used `prp1` for I/O (`prp2` was always 0). Transfers longer than one page were split into many small commands, each covering at most the remainder of the current page. 
This PR adds PRP list support: for contiguous DMA buffers it builds `prp1`/`prp2` and chained PRP list pages, and submits larger read/write commands in one go. I/O is chunked by both PRP chain limits and controller MDTS. This enables multi-page transfers such as 64KB `O_DIRECT` R/W in the NVMe regression test.

### DIFF
--- a/kernel/comps/nvme/src/device/block_device.rs
+++ b/kernel/comps/nvme/src/device/block_device.rs
@@ -48,6 +48,7 @@ use crate::{
     },
     nvme_regs::{NvmeRegs32, NvmeRegs64},
     nvme_spec::{NvmeCommand, NvmeCompletion},
+    prp::PrpPointers,
     transport::pci::transport::{NvmePciTransport, NvmePciTransportLock},
 };
 
@@ -146,7 +147,9 @@ struct IdentifyControllerData {
     serial: [u8; 20],
     model: [u8; 40],
     firmware: [u8; 8],
-    _rest: [u8; 56],
+    _bytes_72_76: [u8; 5],
+    mdts: u8,
+    _rest: [u8; 55],
 }
 
 #[repr(C)]
@@ -191,6 +194,7 @@ struct InitContext {
     dstrd: u16,
     cc_mps_value: u32,
     controller_ready_timeout: Duration,
+    max_data_transfer: Option<usize>,
 }
 
 struct IoMsixVectors([u16; QUEUE_NUM - 1]);
@@ -202,6 +206,7 @@ struct NvmeDeviceInner {
     transport: NvmePciTransportLock,
     namespace: NvmeNamespace,
     dstrd: u16,
+    max_data_transfer: Option<usize>,
     stats: NvmeStats,
 }
 
@@ -284,6 +289,7 @@ impl NvmeDeviceInner {
             transport: NvmePciTransportLock::new(init_ctx.transport),
             namespace,
             dstrd: init_ctx.dstrd,
+            max_data_transfer: init_ctx.max_data_transfer,
             stats: NvmeStats::new(),
         };
         Ok((device, io_msix_vectors))
@@ -371,6 +377,7 @@ impl InitContext {
             dstrd,
             cc_mps_value,
             controller_ready_timeout,
+            max_data_transfer: None,
         })
     }
 
@@ -518,9 +525,18 @@ impl InitContext {
         let model = bytes_to_cstr_string(&result.model);
         let firmware = bytes_to_cstr_string(&result.firmware);
 
+        self.max_data_transfer = max_data_transfer_bytes(result.mdts);
+
         info!(
-            "Controller identified - Serial: {}, Model: {}, Firmware: {}",
-            serial, model, firmware
+            "Controller identified - Serial: {}, Model: {}, Firmware: {}, MDTS={}",
+            serial,
+            model,
+            firmware,
+            if let Some(bytes) = self.max_data_transfer {
+                format!("{} bytes", bytes)
+            } else {
+                "unlimited".to_owned()
+            }
         );
         Ok(())
     }
@@ -719,9 +735,39 @@ impl NvmeDeviceInner {
         }
     }
 
+    /// Returns the maximum bytes transferable in one NVMe command for a buffer at `dma_addr`.
+    fn max_io_bytes(&self, dma_addr: u64) -> usize {
+        let prp_limit = crate::prp::max_transfer_bytes(dma_addr);
+        match self.max_data_transfer {
+            Some(mdts_limit) => prp_limit.min(mdts_limit),
+            None => prp_limit,
+        }
+    }
+
+    /// Submits one read or write covering `length` bytes at contiguous `dma_addr`.
+    fn submit_io_chunk(
+        &self,
+        nsid: u32,
+        lba: u64,
+        dma_addr: u64,
+        length: usize,
+        io_op: IoOp,
+    ) -> Result<(), NvmeDeviceError> {
+        let prp = PrpPointers::build_prp(dma_addr, length)?;
+        let nlb = (length / SECTOR_SIZE - 1) as u16;
+        let entry = match io_op {
+            IoOp::Read => nvme_cmd::io_read(nsid, lba, nlb, prp.prp1, prp.prp2),
+            IoOp::Write => nvme_cmd::io_write(nsid, lba, nlb, prp.prp1, prp.prp2),
+        };
+
+        let _list_pages = prp.list_pages();
+
+        self.submit_and_wait(IO_QID, entry)
+    }
+
     /// Performs read or write I/O for a `BioRequest` on I/O queue [`IO_QID`].
     ///
-    /// Splits work into chunks; each chunk is built from `io_op` and submitted synchronously.
+    /// Each segment is transferred with PRP lists when needed, chunked by MDTS and PRP limits.
     fn io_rw_request(&self, request: BioRequest, io_op: IoOp) {
         const { assert!(LBA_SIZE == SECTOR_SIZE) };
 
@@ -737,35 +783,26 @@ impl NvmeDeviceInner {
                 debug_assert!(dma_slice.daddr().is_multiple_of(SECTOR_SIZE));
                 debug_assert!(dma_slice.size().is_multiple_of(SECTOR_SIZE));
 
-                let seg_sectors = (dma_slice.size() / SECTOR_SIZE) as u64;
-                let mut remaining = seg_sectors;
-                let mut ptr0 = dma_slice.daddr() as u64;
+                let mut dma_addr = dma_slice.daddr() as u64;
+                let mut remaining_bytes = dma_slice.size();
 
-                while remaining > 0 {
-                    // TODO: Support PRP lists / `ptr1`. For now we only use `ptr0` and keep
-                    // `ptr1` at 0, so each command is limited to a page.
-                    let sectors_to_io = {
-                        let bytes_in_page = (PAGE_SIZE as u64) - (ptr0 & (PAGE_SIZE as u64 - 1));
-                        let sectors_in_page = bytes_in_page / (SECTOR_SIZE as u64);
-                        sectors_in_page.min(remaining)
-                    };
+                while remaining_bytes > 0 {
+                    let max_chunk = self.max_io_bytes(dma_addr);
+                    let chunk_bytes = remaining_bytes.min(max_chunk);
+                    debug_assert!(chunk_bytes.is_multiple_of(SECTOR_SIZE));
 
-                    let entry = match io_op {
-                        IoOp::Read => {
-                            nvme_cmd::io_read(nsid, lba, (sectors_to_io - 1) as u16, ptr0, 0u64)
-                        }
-                        IoOp::Write => {
-                            nvme_cmd::io_write(nsid, lba, (sectors_to_io - 1) as u16, ptr0, 0u64)
-                        }
-                    };
                     // TODO: This path submits and waits synchronously, which may block.
-                    if self.submit_and_wait(IO_QID, entry).is_err() {
+                    if self
+                        .submit_io_chunk(nsid, lba, dma_addr, chunk_bytes, io_op)
+                        .is_err()
+                    {
                         status = BioStatus::IoError;
                     }
 
-                    lba += sectors_to_io;
-                    remaining -= sectors_to_io;
-                    ptr0 += (SECTOR_SIZE as u64) * sectors_to_io;
+                    let chunk_sectors = (chunk_bytes / SECTOR_SIZE) as u64;
+                    lba += chunk_sectors;
+                    dma_addr += chunk_bytes as u64;
+                    remaining_bytes -= chunk_bytes;
                 }
             }
             bio.complete(status);
@@ -791,6 +828,15 @@ impl NvmeDeviceInner {
         for bio in request.into_bios() {
             bio.complete(status);
         }
+    }
+}
+
+/// Converts the Identify Controller MDTS field to a byte limit.
+fn max_data_transfer_bytes(mdts: u8) -> Option<usize> {
+    if mdts == 0 {
+        None
+    } else {
+        Some((1usize << mdts) * PAGE_SIZE)
     }
 }
 
--- a/kernel/comps/nvme/src/prp.rs
+++ b/kernel/comps/nvme/src/prp.rs
@@ -1,0 +1,229 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! NVMe PRP (Physical Region Page) setup for contiguous DMA buffers.
+
+use alloc::vec::Vec;
+
+use ostd::mm::{HasDaddr, PAGE_SIZE, VmIo, dma::DmaCoherent, io::util::HasVmReaderWriter};
+
+use crate::device::NvmeDeviceError;
+
+/// Number of 64-bit PRP entries that fit in one PRP list page.
+const PRP_ENTRIES_PER_PAGE: usize = PAGE_SIZE / size_of::<u64>();
+
+pub(crate) struct PrpPointers {
+    pub prp1: u64,
+    pub prp2: u64,
+    list_pages: Vec<DmaCoherent>,
+}
+
+impl PrpPointers {
+    /// Builds PRP pointers for a physically contiguous `[dma_addr, dma_addr + length)` range.
+    pub(crate) fn build_prp(dma_addr: u64, length: usize) -> Result<Self, NvmeDeviceError> {
+        if length == 0 {
+            return Err(NvmeDeviceError::InvalidIoLength);
+        }
+
+        let prp1 = dma_addr;
+        let remaining = remaining_bytes_after_first_page(dma_addr, length);
+        if remaining == 0 {
+            return Ok(Self::without_list(prp1, 0));
+        }
+
+        let second_page = next_page_addr(dma_addr);
+        if remaining <= PAGE_SIZE as u64 {
+            return Ok(Self::without_list(prp1, second_page));
+        }
+
+        Self::with_list(prp1, second_page, remaining)
+    }
+
+    fn without_list(prp1: u64, prp2: u64) -> Self {
+        Self {
+            prp1,
+            prp2,
+            list_pages: Vec::new(),
+        }
+    }
+
+    fn with_list(prp1: u64, mut addr: u64, mut remaining: u64) -> Result<Self, NvmeDeviceError> {
+        let mut list_pages = Vec::new();
+        list_pages
+            .push(DmaCoherent::alloc(1, true).map_err(|_| NvmeDeviceError::DmaAllocationFailed)?);
+        let prp2 = list_pages[0].daddr() as u64;
+        let mut active_list_index = 0usize;
+        let mut list_index = 0usize;
+        let page_size = PAGE_SIZE as u64;
+
+        while remaining > 0 {
+            if list_index == PRP_ENTRIES_PER_PAGE {
+                active_list_index = chain_prp_list_page(&mut list_pages, active_list_index)?;
+                list_index = 1;
+            }
+
+            write_prp_entry(&list_pages[active_list_index], list_index, addr)?;
+            list_index += 1;
+            addr = addr
+                .checked_add(page_size)
+                .ok_or(NvmeDeviceError::InvalidIoLength)?;
+            remaining = remaining.saturating_sub(page_size);
+        }
+
+        Ok(Self {
+            prp1,
+            prp2,
+            list_pages,
+        })
+    }
+
+    pub(crate) fn list_pages(&self) -> &[DmaCoherent] {
+        &self.list_pages
+    }
+}
+
+/// Returns how many bytes of `length` lie after the page containing `dma_addr`.
+fn remaining_bytes_after_first_page(dma_addr: u64, length: usize) -> u64 {
+    let page_size = PAGE_SIZE as u64;
+    let offset = dma_addr & (page_size - 1);
+    (length as u64).saturating_sub(page_size - offset)
+}
+
+/// Returns the physical address of the page immediately after the one containing `dma_addr`.
+fn next_page_addr(dma_addr: u64) -> u64 {
+    let page_size = PAGE_SIZE as u64;
+    let page_mask = page_size - 1;
+    (dma_addr & !page_mask) + page_size
+}
+
+/// Links the full list page at `active_list_index` to a newly allocated page; returns the new index.
+fn chain_prp_list_page(
+    list_pages: &mut Vec<DmaCoherent>,
+    active_list_index: usize,
+) -> Result<usize, NvmeDeviceError> {
+    let last_entry = read_prp_entry(&list_pages[active_list_index], PRP_ENTRIES_PER_PAGE - 1)?;
+    let new_list = DmaCoherent::alloc(1, true).map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
+    let chain_addr = new_list.daddr() as u64;
+    write_prp_entry(&new_list, 0, last_entry)?;
+    write_prp_entry(
+        &list_pages[active_list_index],
+        PRP_ENTRIES_PER_PAGE - 1,
+        chain_addr,
+    )?;
+    list_pages.push(new_list);
+    Ok(list_pages.len() - 1)
+}
+
+fn read_prp_entry(list: &DmaCoherent, index: usize) -> Result<u64, NvmeDeviceError> {
+    debug_assert!(index < PRP_ENTRIES_PER_PAGE);
+    let mut entry = [0u8; 8];
+    list.read_bytes(index * size_of::<u64>(), &mut entry)
+        .map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
+    Ok(u64::from_le_bytes(entry))
+}
+
+/// Returns the maximum transfer size (in bytes) supported by a single PRP chain from `prp1`.
+pub(crate) fn max_transfer_bytes(prp1: u64) -> usize {
+    let page_size = PAGE_SIZE as u64;
+    let page_mask = page_size - 1;
+    let offset = prp1 & page_mask;
+    let first_page_bytes = (page_size - offset) as usize;
+    first_page_bytes.saturating_add(PRP_ENTRIES_PER_PAGE * PAGE_SIZE)
+}
+
+fn write_prp_entry(list: &DmaCoherent, index: usize, addr: u64) -> Result<(), NvmeDeviceError> {
+    debug_assert!(index < PRP_ENTRIES_PER_PAGE);
+    list.writer()
+        .skip(index * size_of::<u64>())
+        .write_val(&addr)
+        .map_err(|_| NvmeDeviceError::DmaAllocationFailed)
+}
+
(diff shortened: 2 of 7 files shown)


kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,229 @@
...
+
+use crate::device::NvmeDeviceError;
+
+/// Number of 64-bit PRP entries that fit in one PRP list page.
+const PRP_ENTRIES_PER_PAGE: usize = PAGE_SIZE / size_of::<u64>();
+
+pub(crate) struct PrpPointers {
+    pub prp1: u64,
+    pub prp2: u64,
+    list_pages: Vec<DmaCoherent>,
```
lrh2000, line 21: From
```rust
pub(crate) struct PrpPointers {
    // ...
    list_pages: Vec<DmaCoherent>,
```
and
```rust
/// Links the full list page at `active_list_index` to a newly allocated page; returns the new index.
fn chain_prp_list_page(
    list_pages: &mut Vec<DmaCoherent>,
    active_list_index: usize,
) -> Result<usize, NvmeDeviceError> {
```

It seems that you want to express the idea that you can link as many PRP lists as possible. Therefore, there is effectively no "maximum transfer size".

However, you also wrote:
```rust
/// Returns the maximum transfer size (in bytes) supported by a single PRP chain from `prp1`.
pub(crate) fn max_transfer_bytes(prp1: u64) -> usize {
    // ...
    first_page_bytes.saturating_add(PRP_ENTRIES_PER_PAGE * PAGE_SIZE)
```

This is confusing. Could you explain why you added support for arbitrary lengths of the PRP list, yet only use it for a single PRP page? There should be an explanation in the code comments as well (or a reference to NVMe spec)?

kernel/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,229 @@
...
+fn read_prp_entry(list: &DmaCoherent, index: usize) -> Result<u64, NvmeDeviceError> {
+    debug_assert!(index < PRP_ENTRIES_PER_PAGE);
+    let mut entry = [0u8; 8];
+    list.read_bytes(index * size_of::<u64>(), &mut entry)
+        .map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
+    Ok(u64::from_le_bytes(entry))
+}
+
+/// Returns the maximum transfer size (in bytes) supported by a single PRP chain from `prp1`.
+pub(crate) fn max_transfer_bytes(prp1: u64) -> usize {
```
lrh2000: It is very confusing why you want to put this method between `read_prp_entry` and `write_prp_entry`.

kernel/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,229 @@
...
+    )?;
+    list_pages.push(new_list);
+    Ok(list_pages.len() - 1)
+}
+
+fn read_prp_entry(list: &DmaCoherent, index: usize) -> Result<u64, NvmeDeviceError> {
+    debug_assert!(index < PRP_ENTRIES_PER_PAGE);
+    let mut entry = [0u8; 8];
+    list.read_bytes(index * size_of::<u64>(), &mut entry)
+        .map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
```
lrh2000: Please use errors for the following cases:
 1. Check untrusted inputs, such as **inputs from userspace, DMA buffers, and device MMIO results**.
    - Depending on the specific case, input from other crates or modules can also qualify this rule.
 2. Propagate **environmental errors** that you know are possible regardless of how the code is written, e.g., out-of-memory errors (allocation failures), hardware errors, e.t.c.

However, please do not use errors for cases that you can easily correct within a module or crate.

---

`read_prp_entry` is private, so the caller should always pass a valid index. There is no need to propagate errors. It should not return a `Result`. This `map_err` should be replaced with a `unwrap()`.

This PR includes several similar cases, including some uses of `checked_add` and `saturating_add`. You should try to fix others as well. However, **not all uses** of these methods are invalid. You should make the final judgment yourself.

---

Also, if you're using an LLM, it's "great" at defensive programming so it can easily generate code like this. However, I don't believe that defensive programming is a valid goal...

kernel/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,229 @@
...
+        chain_addr,
+    )?;
+    list_pages.push(new_list);
+    Ok(list_pages.len() - 1)
+}
+
+fn read_prp_entry(list: &DmaCoherent, index: usize) -> Result<u64, NvmeDeviceError> {
+    debug_assert!(index < PRP_ENTRIES_PER_PAGE);
+    let mut entry = [0u8; 8];
+    list.read_bytes(index * size_of::<u64>(), &mut entry)
```
lrh2000: Use `read_val::<[0u8; 8]>` instead.

kernel/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,229 @@
...
+    let offset = prp1 & page_mask;
+    let first_page_bytes = (page_size - offset) as usize;
+    first_page_bytes.saturating_add(PRP_ENTRIES_PER_PAGE * PAGE_SIZE)
+}
+
+fn write_prp_entry(list: &DmaCoherent, index: usize, addr: u64) -> Result<(), NvmeDeviceError> {
+    debug_assert!(index < PRP_ENTRIES_PER_PAGE);
+    list.writer()
+        .skip(index * size_of::<u64>())
+        .write_val(&addr)
```
lrh2000: `read_prp_entry` uses little-endian, while `write_prp_entry` uses native-endian. Therefore, one of them is incorrect.

kernel/comps/nvme/src/device/block_device.rs
```diff
@@ -719,9 +735,39 @@ impl NvmeDeviceInner {
...
+    fn max_io_bytes(&self, dma_addr: u64) -> usize {
+        let prp_limit = crate::prp::max_transfer_bytes(dma_addr);
+        match self.max_data_transfer {
+            Some(mdts_limit) => prp_limit.min(mdts_limit),
+            None => prp_limit,
+        }
+    }
+
+    /// Submits one read or write covering `length` bytes at contiguous `dma_addr`.
+    fn submit_io_chunk(
```
lrh2000: We should prefer `top-down-reading`.

https://github.com/asterinas/asterinas/blob/2e2b3468f07815be2c372fd5cd103bb37664ad5c/book/src/to-contribute/coding-guidelines/for-maintainability/layout.md?plain=1#L3-L14

lrh2000 (commented):
Thanks for implementing multi-page IO for NVMe.

kernel/core/comps/nvme/src/prp.rs
Cautreoxit: Oh, sorry. With PRP list chaining there is indeed no transfer-size limit from PRP itself, so `max_transfer_bytes` should not have been there. I added it earlier for debugging during development and forgot to remove it afterward. I’ve dropped it.

Cautreoxit:
Thank you for your valuable suggestions! I have fixed the previous issues and made these changes based on your feedback.
If you have any further suggestions or comments, please let me know, and I'll be happy to make additional improvements. Thanks again for your help!

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -629,30 +645,100 @@ impl InitContext {
...
-        qid: usize,
-    ) -> NvmeSubmissionQueueAccess<'_, SpinLockGuard<'_, NvmeSubmissionQueue, LocalIrqDisabled>>
-    {
-        NvmeSubmissionQueueAccess::new(
-            qid as u16,
-            self.dstrd,
-            self.submission_queues[qid].lock(),
-            self.transport.dbregs(),
-        )
+    fn read(&self, request: BioRequest) {
```
lrh2000: If you want to reorder the existing methods, you can do so in a separate commit. This will make it much easier to check the changes.

If you don't want to do it in this PR, that's fine, but please try to do so next time. Thanks.

(You can try reading the diffs yourself and you'll understand why I'm saying this. It's too difficult to understand the changes if you mix reordering with new changes.)

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -629,30 +645,100 @@ impl InitContext {
...
+        io_op: IoOp,
+    ) -> Result<(), NvmeDeviceError> {
+        let prp = PrpPointers::build_prp(dma_addr, length)?;
+        let nlb = (length / SECTOR_SIZE - 1) as u16;
+        let entry = match io_op {
+            IoOp::Read => nvme_cmd::io_read(nsid, lba, nlb, prp.prp1, prp.prp2),
+            IoOp::Write => nvme_cmd::io_write(nsid, lba, nlb, prp.prp1, prp.prp2),
+        };
+
+        let _list_pages = prp.list_pages();
```
lrh2000: I don't think this method accomplishes anything. It doesn't seem to be used elsewhere either. Can we remove it?

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -719,78 +805,39 @@ impl NvmeDeviceInner {
...
-            .map_or(BioStatus::IoError, |_| BioStatus::Complete);
-        for bio in request.into_bios() {
-            bio.complete(status);
-        }
+/// Converts the Identify Controller MDTS field to a byte limit.
+fn max_data_transfer_bytes(mdts: u8) -> Option<usize> {
+    if mdts == 0 {
+        None
+    } else {
+        Some((1usize << mdts) * PAGE_SIZE)
```
lrh2000: This is missing at least some checks. If the device provides an invalid MDTS, either fail the initialization or fall back to a reasonable value and print a warning.

(e.g., `(1 << 255) * PAGE_SIZE` does not seem reasonable)

---

>         // `NLB` is zero-based in a `u16`, so one command covers at most 65536 LBAs.

We can use this as a threshold and cap the shift at a reasonable value (i.e., merging `max_data_transfer_bytes` and `max_io_bytes`). This will result in simpler code. [Unless you're considering cases without the NLB limit, which can be implemented in the future if necessary].

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+        let prp2 = list_pages[0].daddr() as u64;
+        let mut active_list_index = 0usize;
+        let mut list_index = 0usize;
+        let page_size = PAGE_SIZE as u64;
+
+        while remaining > 0 {
+            if list_index == PRP_ENTRIES_PER_PAGE {
+                active_list_index = chain_prp_list_page(&mut list_pages, active_list_index)?;
+                list_index = 1;
+            }
```
lrh2000: The code can be made simpler by:
```rust
if list_index == PRP_ENTRIES_PER_PAGE - 1 && remaining > PAGE_SIZE {
    write_prp_entry();
    list_pages.push();
    active_list_index += 1;
    list_index = 0;
}
```

I think there is no need to have `read_prp_entry`..

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+        let prp = PrpPointers::build_prp(addr, PAGE_SIZE * 4).unwrap();
+        assert_eq!(prp.list_pages.len(), 1);
+
+        for page_index in 1..4 {
+            let mut entry = [0u8; 8];
+            prp.list_pages[0]
+                .read_bytes((page_index - 1) * 8, &mut entry)
+                .unwrap();
+            assert_eq!(
+                u64::from_le_bytes(entry),
```
lrh2000: Why don't you use `read_list_entry`?

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+                addr + page_index as u64 * PAGE_SIZE as u64
+            );
+        }
+    }
+
+    fn read_list_entry(list: &DmaCoherent, index: usize) -> u64 {
```
lrh2000: Use `read_prp_entry` or rename this method to `read_prp_entry` if you remove `read_prp_entry` (as said in comments above).

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    fn with_list(prp1: u64, mut addr: u64, mut remaining: u64) -> Result<Self, NvmeDeviceError> {
+        let mut list_pages = Vec::new();
+        list_pages
+            .push(DmaCoherent::alloc(1, true).map_err(|_| NvmeDeviceError::DmaAllocationFailed)?);
+        let prp2 = list_pages[0].daddr() as u64;
+        let mut active_list_index = 0usize;
+        let mut list_index = 0usize;
+        let page_size = PAGE_SIZE as u64;
```
lrh2000: Sounds like you want a `const PAGE_SIZE_U64: u64 = PAGE_SIZE as u64` as a global variable.
```
kernel/core/comps/nvme/src/prp.rs:        if remaining <= PAGE_SIZE as u64 {
kernel/core/comps/nvme/src/prp.rs:        let page_size = PAGE_SIZE as u64;
kernel/core/comps/nvme/src/prp.rs:    let page_size = PAGE_SIZE as u64;
kernel/core/comps/nvme/src/prp.rs:    let page_size = PAGE_SIZE as u64;
kernel/core/comps/nvme/src/prp.rs:        assert_eq!(prp.prp2, addr + PAGE_SIZE as u64);
kernel/core/comps/nvme/src/prp.rs:                addr + page_index as u64 * PAGE_SIZE as u64
kernel/core/comps/nvme/src/prp.rs:                addr + page_index * PAGE_SIZE as u64
kernel/core/comps/nvme/src/prp.rs:            addr + 512 * PAGE_SIZE as u64
kernel/core/comps/nvme/src/prp.rs:                addr + page_index * PAGE_SIZE as u64
```

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    fn with_list(prp1: u64, mut addr: u64, mut remaining: u64) -> Result<Self, NvmeDeviceError> {
+        let mut list_pages = Vec::new();
+        list_pages
+            .push(DmaCoherent::alloc(1, true).map_err(|_| NvmeDeviceError::DmaAllocationFailed)?);
+        let prp2 = list_pages[0].daddr() as u64;
+        let mut active_list_index = 0usize;
+        let mut list_index = 0usize;
```
lrh2000: I think this can be a clear layout and variable names:
```rust
        let mut list_pages = Vec::new();

        let first_list_page = DmaCoherent::alloc()?;
        let prp2 = first_list_page.addr() as u64;
        list_pages.push_back(first_list_page);

        let mut cur_list_page = &list_pages[0];
        let mut cur_list_index = 0;

        while remaining > 0 {
            if cur_list_index == PRP_ENTRIES_PER_PAGE - 1 && remaining > PAGE_SIZE {
                let new_list_page = DmaCoherent::alloc()?;
                write_prp_entry(cur_list_page, active_list_index, new_list_page.daddr() as u64);
                list_pages.push_back(new_list_page);

                cur_list_page = list_pages.last().unwrap();
                cur_list_index = 0;
            }

            // ...
        }       
```

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn without_list(prp1: u64, prp2: u64) -> Self {
+        Self {
+            prp1,
+            prp2,
+            list_pages: Vec::new(),
+        }
+    }
+
+    fn with_list(prp1: u64, mut addr: u64, mut remaining: u64) -> Result<Self, NvmeDeviceError> {
+        let mut list_pages = Vec::new();
```
lrh2000: I think we can reserve the capacity `remaining.div_ceil(PAGE_SIZE * (PRP_ENTRIES_PER_PAGE - 1)`?

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub prp2: u64,
+    list_pages: Vec<DmaCoherent>,
+}
+
+impl PrpPointers {
+    /// Builds PRP pointers for a physically contiguous `[dma_addr, dma_addr + length)` range.
+    pub(crate) fn build_prp(dma_addr: u64, length: usize) -> Result<Self, NvmeDeviceError> {
+        if length == 0 {
+            return Err(NvmeDeviceError::InvalidIoLength);
+        }
```
lrh2000: It seems that this method should just accept a `length: NonZeroUsize`.

The caller should not pass a zero length. In fact, it has already checked `remaining > 0` and `max_io_size` is guaranteed to be non-zero as well.

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,217 @@
...
+use ostd::mm::{HasDaddr, PAGE_SIZE, VmIo, dma::DmaCoherent};
+
+use crate::device::NvmeDeviceError;
+
+/// Number of 64-bit PRP entries that fit in one PRP list page.
+const PRP_ENTRIES_PER_PAGE: usize = PAGE_SIZE / size_of::<u64>();
+
+pub(crate) struct PrpPointers {
+    pub prp1: u64,
+    pub prp2: u64,
```
lrh2000: Let's avoid exposing the internal fields. Add getters for them instead.

lrh2000 (commented):
Thanks for your quick updates!

kernel/core/comps/nvme/src/device/block_device.rs
Cautreoxit: Oh yes, I agree with that. I will reorder in another PR and pay attention to this. For now, I have dropped them.

kernel/core/comps/nvme/src/device/block_device.rs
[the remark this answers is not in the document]
Cautreoxit: I think this issue has already been addressed in this round.

Cautreoxit:
Thanks for your advice! I have addressed these issues. Please check them and see if there are any problems. Thanks again!

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,204 @@
...
+            if cur_list_index == PRP_ENTRIES_PER_PAGE - 1 && remaining > PAGE_SIZE_U64 {
+                let new_list_page = DmaCoherent::alloc(1, true)
+                    .map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
+                write_prp_entry(
+                    list_pages.last().unwrap(),
+                    cur_list_index,
+                    new_list_page.daddr() as u64,
+                );
+                list_pages.push(new_list_page);
+                cur_list_index = 0;
```
lrh2000, line 80: Since you didn't follow my exact suggestion, I thought maybe my suggestion had some problems, such as problems with borrow checks.

But I have had a trial, it does not seem to have problems, I think the code is better because we have fewer `unwrap()`s?

```patch
diff --git a/kernel/core/comps/nvme/src/prp.rs b/kernel/core/comps/nvme/src/prp.rs
index 0357ce681..c35daadd1 100644
--- a/kernel/core/comps/nvme/src/prp.rs
+++ b/kernel/core/comps/nvme/src/prp.rs
@@ -65,6 +65,7 @@ impl PrpPointers {
         let prp2 = first_list_page.daddr() as u64;
         list_pages.push(first_list_page);
 
+        let mut cur_list_page = &list_pages[0];
         let mut cur_list_index = 0usize;
 
         while remaining > 0 {
@@ -72,16 +73,14 @@ impl PrpPointers {
             if cur_list_index == PRP_ENTRIES_PER_PAGE - 1 && remaining > PAGE_SIZE_U64 {
                 let new_list_page = DmaCoherent::alloc(1, true)
                     .map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
-                write_prp_entry(
-                    list_pages.last().unwrap(),
-                    cur_list_index,
-                    new_list_page.daddr() as u64,
-                );
+                write_prp_entry(cur_list_page, cur_list_index, new_list_page.daddr() as u64);
                 list_pages.push(new_list_page);
+
+                cur_list_page = list_pages.last().unwrap();
                 cur_list_index = 0;
             }
 
-            write_prp_entry(list_pages.last().unwrap(), cur_list_index, addr);
+            write_prp_entry(cur_list_page, cur_list_index, addr);
             cur_list_index += 1;
 
             if remaining <= PAGE_SIZE_U64 {
```

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        let first_list_page =
+            DmaCoherent::alloc(1, true).map_err(|_| NvmeDeviceError::DmaAllocationFailed)?;
+        let prp2 = first_list_page.daddr() as u64;
+        list_pages.push(first_list_page);
+
+        let mut cur_list_index = 0usize;
+
+        while remaining > 0 {
+            // Last entry of a full PRP list page must point to the next list page.
```
lrh2000:
```suggestion
            // The last entry of a non-final PRP list page must point to the next list page.
```

The last entry of the final PRP list page can point to a data page?

Not sure if "final" is the best wording. You can try to use another word if you have a better one.

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert_eq!(prp.prp2(), addr + PAGE_SIZE_U64);
+        assert!(prp.list_pages.is_empty());
+    }
+
+    #[ktest]
+    fn four_page_transfer_fills_prp_list() {
+        let dma = DmaCoherent::alloc(4, true).unwrap();
+        let addr = dma.daddr() as u64;
+        let prp = PrpPointers::build_prp(addr, NonZeroUsize::new(PAGE_SIZE * 4).unwrap()).unwrap();
+        assert_eq!(prp.list_pages.len(), 1);
```
lrh2000, line 159: I think this test misses checks to `prp1` and `prp2`. Try this:
```rust
        let dma = DmaCoherent::alloc(4, true).unwrap();
        let addr = dma.daddr() as u64;
        let prp = PrpPointers::build_prp(addr, NonZeroUsize::new(PAGE_SIZE * 4).unwrap()).unwrap();

        assert_eq!(prp.prp1(), addr);
        assert_eq!(prp.prp2(), prp.list_pages[0].daddr() as u64);
        assert_eq!(prp.list_pages.len(), 1);
```

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,204 @@
...
+                read_prp_entry(first_list, (page_index - 1) as usize),
+                addr + page_index * PAGE_SIZE_U64
+            );
+        }
+        assert_eq!(
+            read_prp_entry(first_list, PRP_ENTRIES_PER_PAGE - 1),
+            second_list.daddr() as u64
+        );
+        assert_eq!(read_prp_entry(second_list, 0), addr + 512 * PAGE_SIZE_U64);
+        for page_index in 513..NR_PAGES {
```
lrh2000:
```suggestion
        for page_index in 512..NR_PAGES {
```

No idea why 512 is a special case. It does not seem to be.

kernel/core/comps/nvme/src/prp.rs
```diff
@@ -0,0 +1,204 @@
...
+    prp1: u64,
+    prp2: u64,
+    #[cfg_attr(not(ktest), expect(dead_code))]
+    list_pages: Vec<DmaCoherent>,
+}
+
+impl PrpPointers {
+    /// Builds PRP pointers for a physically contiguous `[dma_addr, dma_addr + length)` range.
+    pub(crate) fn build_prp(dma_addr: u64, length: NonZeroUsize) -> Result<Self, NvmeDeviceError> {
+        let length = length.get();
```
lrh2000: I don't think we need to define this local variable. `length.get()` is only used once below. Move the expression to the place where it is used.

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -721,12 +732,13 @@ impl NvmeDeviceInner {
...
     /// Performs read or write I/O for a `BioRequest` on I/O queue [`IO_QID`].
     ///
-    /// Splits work into chunks; each chunk is built from `io_op` and submitted synchronously.
+    /// Each segment is transferred with PRP lists when needed.
     fn io_rw_request(&self, request: BioRequest, io_op: IoOp) {
         const { assert!(LBA_SIZE == SECTOR_SIZE) };
 
         let nsid = self.namespace.id;
         let mut lba = request.sid_range().start.to_raw();
+        let max_chunk = self.max_io_bytes;
```
lrh2000: I don't think we need to define this local variable. `self.max_io_bytes` is only used once below. Move the expression to the place where it is used.

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -794,6 +815,33 @@ impl NvmeDeviceInner {
...
 
+/// Maximum bytes transferable in one NVMe command.
+///
+/// `NLB` is zero-based in a `u16`, so one command covers at most 65536 LBAs.
+const MAX_BYTES_BY_NLB: usize = (u16::MAX as usize + 1) * SECTOR_SIZE;
+
+/// Converts Identify Controller MDTS into an effective per-command byte limit.
+///
+/// MDTS `0` means no device-side limit; we still cap at [`MAX_BYTES_BY_NLB`].
+fn max_io_bytes_from_mdts(mdts: u8) -> usize {
```
lrh2000: Move `MAX_BYTES_BY_NLB`/`max_io_bytes_from_mdts` after `bytes_to_cstr_string`.

Doing so will match the order that they're used.

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -737,41 +749,50 @@ impl NvmeDeviceInner {
...
-                        }
-                        IoOp::Write => {
-                            nvme_cmd::io_write(nsid, lba, (sectors_to_io - 1) as u16, ptr0, 0u64)
-                        }
-                    };
+                let mut dma_addr = dma_slice.daddr() as u64;
+                let mut remaining_bytes = dma_slice.size();
+
+                while remaining_bytes > 0 {
+                    let chunk_bytes = NonZeroUsize::new(remaining_bytes.min(max_chunk)).unwrap();
```
lrh2000: As an experiment, you can try to see if you can result in better code by removing this `unwrap()`.

I think the following code works:
```rust
const MAX_BYTES_BY_NLB: NonZeroUsize;

fn max_io_bytes_from_mdts(mdts: u8) -> NonZeroUsize {
    // ...

    const ONE_NONZERO: usize = NonZeroUsize::new(1).unwrap();
    const PAGE_SIZE_NONZERO: usze = NonZeroUsize::new(PAGE_SIZE).unwrap();

    // Use `NonZero`'s `checked_shl` and `checked_mul` below.
}
```

Then you can write:
```rust
     while let Some(remaining) = NonZeroUsize::new(remining_bytes) {
         let chunk_bytes = remaining.min(self.max_io_size);
```

This will give you a non-zero `chunk_bytes` without any `unwrap()`s.

lrh2000 (commented):
Thanks again for your quick updates! Here are some remaining comments. Otherwise I think the changes mostly look good!

kernel/core/comps/nvme/src/prp.rs
Cautreoxit: Oh, at the time I noticed a few small issues in the snippet, so I treated it as pseudocode describing the idea and only followed the approach in my own rewrite. I've now updated the code to match this version.

Cautreoxit:
Thank you for the quick review! I’ve addressed all the issues above. Could you please check whether the changes look reasonable (such as the wording of the comments)? Please let me know if there are any issues. Thanks!

kernel/core/comps/nvme/src/device/block_device.rs
```diff
@@ -803,6 +827,42 @@ fn bytes_to_cstr_string(bytes: &[u8]) -> String {
...
+    if mdts == 0 {
+        return MAX_BYTES_BY_NLB;
+    }
+
+    const ONE_NONZERO: NonZeroUsize = NonZeroUsize::new(1).unwrap();
+    const PAGE_SIZE_NONZERO: NonZeroUsize = NonZeroUsize::new(PAGE_SIZE).unwrap();
+
+    let Some(pages) = ONE_NONZERO
+        .get()
+        .checked_shl(u32::from(mdts))
```
lrh2000: Sorry, this should be:
```suggestion
    let Some(pages) = 1usize
        .checked_shl(u32::from(mdts))
```

I was thought there is a `NonZeroUsize::checked_shl` but there isn't. So let's remove `ONE_NONZERO`.

lrh2000 (commented):
Just one nit. Thanks for your contribution!

Cautreoxit:
Thanks! This has been addressed.

lrh2000 (approved):

PR #3353, opened 2026-06-09
Reject `MAP_SHARED_VALIDATE` for anonymous mappings

Marsman1996:
When `MAP_ANONYMOUS` is set and the mapping type is `MAP_SHARED_VALIDATE` (0x03), we now return `EINVAL` before creating the mapping.

The corresponding Linux code is in
https://elixir.bootlin.com/linux/v7.0/source/mm/mmap.c#L493-L542
```c
switch (flags & MAP_TYPE) {
  case MAP_SHARED:
      ...
      break;
  case MAP_PRIVATE:
      ...
      break;
  default:
      return -EINVAL;             // <-- MAP_SHARED_VALIDATE hits this
  }
```

### Describe the bug
On Linux, `mmap()` with `MAP_SHARED_VALIDATE | MAP_ANONYMOUS` returns `EINVAL`. On Asterinas, this combination is accepted and creates a mapping. This is a compatibility difference — programs relying on this error to detect `MAP_SHARED_VALIDATE` support will behave differently.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <sys/mman.h>

#ifndef MAP_SHARED_VALIDATE
#define MAP_SHARED_VALIDATE 0x03
#endif

int main(void) {
    errno = 0;
    void *p = mmap(NULL, 4096, PROT_READ | PROT_WRITE,
                   MAP_SHARED_VALIDATE | MAP_ANONYMOUS, -1, 0);
    if (p == MAP_FAILED) {
        printf("MAP_SHARED_VALIDATE|MAP_ANONYMOUS: errno=%d (%s)\n", errno, strerror(errno));
    } else {
        printf("MAP_SHARED_VALIDATE|MAP_ANONYMOUS: success (addr=%p)\n", p);
        munmap(p, 4096);
    }
    return 0;
}
```

### Expected behavior
Should return `EINVAL` (matching Linux behavior).

### Log
- In Asterinas:
```
MAP_SHARED_VALIDATE|MAP_ANONYMOUS: success (addr=0x7ffffeff9000)
```
- In Linux:
```
MAP_SHARED_VALIDATE|MAP_ANONYMOUS: errno=22 (Invalid argument)
```


junyang-zh (approved):

kernel/src/syscall/mmap.rs
```diff
@@ -99,6 +99,13 @@ fn do_sys_mmap(
         }
 
         if option.flags().contains(MMapFlags::MAP_ANONYMOUS) {
+            // Linux rejects MAP_SHARED_VALIDATE for anonymous mappings.
+            if option.typ() == MMapType::SharedValidate {
+                return_errno_with_message!(
+                    Errno::EINVAL,
+                    "invalid flag combination: MAP_SHARED_VALIDATE with MAP_ANONYMOUS"
```
lrh2000:
```suggestion
                    "MAP_SHARED_VALIDATE and MAP_ANONYMOUS cannot be used together"
```

Ref:
```
> grep -R "cannot be used together" kernel/**/*.rs
kernel/src/process/clone.rs:                "`CLONE_FS` cannot be used together with `CLONE_NEWNS` or `CLONE_NEWUSER`"
kernel/src/process/clone.rs:                    "`CLONE_PARENT` cannot be used together with `CLONE_NEWUSER` or `CLONE_NEWPID`"
kernel/src/process/clone.rs:                    "`CLONE_THREAD` cannot be used together with `CLONE_PIDFD` or `CLONE_NEWUSER`"
kernel/src/syscall/fallocate.rs:                "PUNCH_HOLE and ZERO_RANGE cannot be used together"
```

lrh2000 (approved):
Just a minor error message nit for consistency. Otherwise LGTM!

Thanks for fixing this.

cqs21 (approved):

PR #3355, opened 2026-06-09
Introduce `RawUid` and `Uid` to reject invalid UIDs (and GIDs)

Marsman1996:
Cast all non-(-1) values to u32 to match Linux behavior.

`chown()` rejects any negative uid or gid value other than -1 with `EINVAL`. On Linux, negative values are treated as unsigned ~~(e.g., `uid=-2` becomes `uid_t(65534)` = `nobody`). This affects programs that use the `nobody` user (uid 65534, which is `-2` when passed as `int`).~~

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <errno.h>
#include <string.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/stat.h>

int main(void) {
    const char *path = "/tmp/chown_test";
    int fd = open(path, O_CREAT | O_WRONLY | O_TRUNC, 0644);
    close(fd);

    errno = 0;
    int ret = chown(path, -2, -1); /* -2 = uid 65534 (nobody) */
    printf("chown(path, -2, -1): ret=%d errno=%d (%s)\n",
           ret, errno, ret == 0 ? "success" : strerror(errno));

    if (ret == 0) {
        struct stat st;
        stat(path, &st);
        printf("new uid=%d\n", st.st_uid);
    }

    unlink(path);
    return 0;
}
```

### Expected behavior
`chown(path, -2, -1)` should succeed and set the uid to ~~65534 (nobody).~~

### Log
- In Asterinas:
```
chown(path, -2, -1): ret=-1 errno=22 (Invalid argument)
```
- In Linux:
```
chown(path, -2, -1): ret=0 errno=0 (success)
new uid=-2
```

### DIFF
--- a/kernel/src/fs/fs_impls/exfat/inode.rs
+++ b/kernel/src/fs/fs_impls/exfat/inode.rs
@@ -1441,8 +1441,8 @@ impl Inode for ExfatInode {
             type_: inner.inode_type,
             mode: inner.make_mode(),
             nr_hard_links: nlinks,
-            uid: Uid::new(inner.fs().mount_option().fs_uid as u32),
-            gid: Gid::new(inner.fs().mount_option().fs_gid as u32),
+            uid: Uid::from_raw(inner.fs().mount_option().fs_uid as u32),
+            gid: Gid::from_raw(inner.fs().mount_option().fs_gid as u32),
             container_dev_id: inner.fs().container_device_id(),
             self_dev_id: None,
         }
@@ -1486,7 +1486,7 @@ impl Inode for ExfatInode {
     }
 
     fn owner(&self) -> Result<Uid> {
-        Ok(Uid::new(
+        Ok(Uid::from_raw(
             self.inner.read().fs().mount_option().fs_uid as u32,
         ))
     }
@@ -1497,7 +1497,7 @@ impl Inode for ExfatInode {
     }
 
     fn group(&self) -> Result<Gid> {
-        Ok(Gid::new(
+        Ok(Gid::from_raw(
             self.inner.read().fs().mount_option().fs_gid as u32,
         ))
     }
--- a/kernel/src/process/credentials/user.rs
+++ b/kernel/src/process/credentials/user.rs
@@ -6,6 +6,9 @@ use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
 
 use crate::prelude::*;
 
+/// The raw UID type at the syscall ABI boundary, matching Linux's `uid_t`.
+pub type RawUid = u32;
+
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Pod)]
 pub struct Uid(u32);
@@ -13,10 +16,10 @@ pub struct Uid(u32);
 const ROOT_UID: u32 = 0;
 
 impl Uid {
-    /// The invalid UID, typically used to indicate that no valid UID is found when returning to user space.
+    /// The raw value representing an invalid UID (`(uid_t)-1` in Linux).
     ///
     /// Reference: <https://elixir.bootlin.com/linux/v6.15/source/include/linux/uidgid.h#L50>.
-    pub const INVALID: Uid = Self::new(u32::MAX);
+    pub const RAW_INVALID: RawUid = u32::MAX;
 
     /// The overflow UID, typically used to indicate that user mappings between namespaces fail.
     ///
@@ -24,24 +27,41 @@ impl Uid {
     /// configured via `/proc/sys/kernel/overflowuid`.
     ///
     /// Reference: <https://elixir.bootlin.com/linux/v6.15/source/kernel/sys.c#L166>.
-    pub const OVERFLOW: Uid = Self::new(65534);
+    pub const OVERFLOW: Uid = Self(65534);
 
     pub const fn new_root() -> Self {
         Self(ROOT_UID)
     }
 
-    pub const fn new(uid: u32) -> Self {
+    /// Creates a `Uid` from a raw value, returning `None` if the value is the
+    /// invalid sentinel (i.e., [`Self::RAW_INVALID`]).
+    pub const fn new(uid: RawUid) -> Option<Self> {
+        if uid == Self::RAW_INVALID {
+            None
+        } else {
+            Some(Self(uid))
+        }
+    }
+
+    /// Creates a `Uid` from a raw value without checking for the invalid sentinel.
+    ///
+    /// This is intended for filesystem use where any raw UID stored on disk is valid.
+    pub const fn from_raw(uid: RawUid) -> Self {
         Self(uid)
+    }
+
+    /// Returns whether this UID has a valid mapping (i.e., is not the invalid sentinel).
+    pub const fn has_valid_mapping(&self) -> bool {
+        self.0 != Self::RAW_INVALID
+    }
+
+    /// Returns the underlying raw UID value.
+    pub const fn as_raw(&self) -> RawUid {
+        self.0
     }
 
     pub const fn is_root(&self) -> bool {
         self.0 == ROOT_UID
-    }
-}
-
-impl From<u32> for Uid {
-    fn from(value: u32) -> Self {
-        Self::new(value)
     }
 }
 
@@ -51,7 +71,15 @@ impl From<Uid> for u32 {
     }
 }
 
-define_atomic_version_of_integer_like_type!(Uid, {
+impl TryFrom<u32> for Uid {
+    type Error = Error;
+
+    fn try_from(value: u32) -> Result<Self> {
+        Self::new(value).ok_or_else(|| Error::with_message(Errno::EINVAL, "the UID is invalid"))
+    }
+}
+
+define_atomic_version_of_integer_like_type!(Uid, try_from = true, {
     #[derive(Debug)]
     pub(super) struct AtomicUid(AtomicU32);
 });
--- a/kernel/src/syscall/chown.rs
+++ b/kernel/src/syscall/chown.rs
@@ -8,14 +8,22 @@ use crate::{
         vfs::path::{AT_FDCWD, EmptyPathStr, FsPath},
     },
     prelude::*,
-    process::{Gid, Uid},
+    process::{Gid, RawGid, RawUid, Uid},
 };
 
-pub fn sys_fchown(raw_fd: RawFileDesc, uid: i32, gid: i32, ctx: &Context) -> Result<SyscallReturn> {
-    debug!("raw_fd = {}, uid = {}, gid = {}", raw_fd, uid, gid);
+pub fn sys_fchown(
+    raw_fd: RawFileDesc,
+    raw_uid: RawUid,
+    raw_gid: RawGid,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    debug!(
+        "raw_fd = {}, raw_uid = {}, raw_gid = {}",
+        raw_fd, raw_uid, raw_gid
+    );
 
-    let uid = to_optional_id(uid, Uid::new)?;
-    let gid = to_optional_id(gid, Gid::new)?;
+    let uid = Uid::new(raw_uid);
+    let gid = Gid::new(raw_gid);
     if uid.is_none() && gid.is_none() {
         return Ok(SyscallReturn::Return(0));
     }
@@ -32,16 +40,26 @@ pub fn sys_fchown(raw_fd: RawFileDesc, uid: i32, gid: i32, ctx: &Context) -> Res
     Ok(SyscallReturn::Return(0))
 }
 
-pub fn sys_chown(path_ptr: Vaddr, uid: i32, gid: i32, ctx: &Context) -> Result<SyscallReturn> {
-    sys_fchownat(AT_FDCWD, path_ptr, uid, gid, 0, ctx)
+pub fn sys_chown(
+    path_ptr: Vaddr,
+    raw_uid: RawUid,
+    raw_gid: RawGid,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    sys_fchownat(AT_FDCWD, path_ptr, raw_uid, raw_gid, 0, ctx)
 }
 
-pub fn sys_lchown(path_ptr: Vaddr, uid: i32, gid: i32, ctx: &Context) -> Result<SyscallReturn> {
+pub fn sys_lchown(
+    path_ptr: Vaddr,
+    raw_uid: RawUid,
+    raw_gid: RawGid,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
     sys_fchownat(
         AT_FDCWD,
         path_ptr,
-        uid,
-        gid,
+        raw_uid,
+        raw_gid,
         ChownFlags::AT_SYMLINK_NOFOLLOW.bits(),
         ctx,
     )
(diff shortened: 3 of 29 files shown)


test/initramfs/src/regression/fs/chown/chown.c
```diff
@@ -0,0 +1,64 @@
...
+
+	TEST_SUCC(chown(TEST_FILE, 0, 0));
+}
+END_TEST()
+
+FN_SETUP(cleanup)
+{
+	unlink(TEST_FILE);
+}
+END_SETUP()
```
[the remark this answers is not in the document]
Marsman1996: Can't you see the other code in `test/initramfs/src/regression`, they all use this pattern

kernel/src/syscall/chown.rs
```diff
@@ -96,17 +96,10 @@ pub fn sys_fchownat(
...
-        None
-    } else {
-        return_errno!(Errno::EINVAL);
-    };
-
-    Ok(id)
+fn to_optional_id<T>(id: i32, f: impl Fn(u32) -> T) -> Option<T> {
+    // If the owner or group is specified as -1, then that ID is not changed.
+    // All other values are cast to u32, matching Linux's unsigned uid_t/gid_t ABI.
+    if id == -1 { None } else { Some(f(id as u32)) }
```
lrh2000:
```suggestion
    // All other values are cast to `u32`, matching Linux's unsigned `uid_t`/`gid_t` ABI.
    if id == -1 { None } else { Some(f(id.cast_unsigned())) }
```

test/initramfs/src/regression/fs/chown/chown.c
```diff
@@ -0,0 +1,61 @@
...
+FN_SETUP(create_test_file)
+{
+	int fd = CHECK(open(TEST_FILE, O_CREAT | O_WRONLY | O_TRUNC, 0644));
+	CHECK(close(fd));
+}
+END_SETUP()
+
+FN_TEST(chown_negative_uid_as_unsigned)
+{
+	// -2 cast to uid_t (32-bit unsigned) is 4294967294.
```
lrh2000:
```suggestion
	// -2 cast to `uid_t` (32-bit unsigned) is 4294967294.

File: test/initramfs/src/regression/fs/chown/chown.c [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+
+	TEST_SUCC(chown(TEST_FILE, 0, 0));
+}
+END_TEST()
+
+FN_TEST(chown_minus_one_no_change)
+{
+	TEST_SUCC(chown(TEST_FILE, 1000, 1000));
+
+	// -1 means "no change" for both uid and gid
```
lrh2000:
```suggestion
	// -1 means "no change" for both UID and GID.
```

test/initramfs/src/regression/fs/run_test.sh
```diff
@@ -104,6 +104,10 @@ test_ext2 "/ext2" "test_file.txt"
 ./ext2/xattr
 echo "All ext2 fs test passed."
 
+echo "Start chown test......"
+./chown/chown
+echo "All chown test passed."
```
lrh2000: So should it belong to the following block?
```
./chown/chown

./inotify/inotify_align
./inotify/inotify_poll
./inotify/inotify_unlink

./isolation/chroot
./isolation/pivot_root
```

lrh2000 (commented):
> `chown()` rejects any negative uid or gid value other than -1 with `EINVAL`. On Linux, negative values are treated as unsigned (e.g., `uid=-2` becomes `uid_t(65534)` = `nobody`). This affects programs that use the `nobody` user (uid 65534, which is `-2` when passed as `int`).

But why should `-1` (`4294967294`) be `nobody` (`65534`)? I believe $65534 \not= 4294967294$. Am I missing anything?

Marsman1996:
> > `chown()` rejects any negative uid or gid value other than -1 with `EINVAL`. On Linux, negative values are treated as unsigned (e.g., `uid=-2` becomes `uid_t(65534)` = `nobody`). This affects programs that use the `nobody` user (uid 65534, which is `-2` when passed as `int`).
> 
> But why should `-1` (`4294967294`) be `nobody` (`65534`)? I believe 65534 ≠ 4294967294 . Am I missing anything?

My bad, when opus generates bug report, it treats `-2` as `nobody`.
And the Linux output is 
```
chown(path, -2, -1): ret=0 errno=0 (success)
new uid=-2
```

test/initramfs/src/regression/fs/chown/chown.c
```diff
@@ -0,0 +1,61 @@
...
+	struct stat st;
+	TEST_RES(stat(TEST_FILE, &st), st.st_uid == 1000 && st.st_gid == 1000);
+
+	TEST_SUCC(chown(TEST_FILE, 0, 0));
+}
+END_TEST()
+
+FN_SETUP(cleanup)
+{
+	unlink(TEST_FILE);
```
lrh2000:
```suggestion
	CHECK(unlink(TEST_FILE));
```

lrh2000 (commented):
Need to resolve conflicts...

lrh2000:
Having a closer look, this is also wrong:
https://github.com/asterinas/asterinas/blob/ecabd4c7ddaf26ad9b9f0a98e81db1ed914d0507/kernel/src/syscall/setuid.rs#L10-L12

Essentially, I believe that we should redesign our `Uid` and `Gid` types. Perhaps we should introduce a `RawUid` type for all UIDs and restrict the `Uid` specifically to valid UIDs (i.e., not `-1`)?
```rust
type RawUid = u32;

struct Uid(u32);

impl Uid {
    fn new(raw_uid: RawUid) -> Option<Self>;
}
```
So this should be:
```rust
let Some(uid) = Uid::new(raw_uid) else {
    return_errno_with_message!(Errno::EINVAL, "the UID is invalid");
};
```

This pattern should be used when implementing the `setuid` and `chown` system calls instead.

Marsman1996:
> Having a closer look, this is also wrong:
> 
> https://github.com/asterinas/asterinas/blob/ecabd4c7ddaf26ad9b9f0a98e81db1ed914d0507/kernel/src/syscall/setuid.rs#L10-L12
> 
> Essentially, I believe that we should redesign our `Uid` and `Gid` types. Perhaps we should introduce a `RawUid` type for all UIDs and restrict the `Uid` specifically to valid UIDs (i.e., not `-1`)?
> 
> ```rust
> type RawUid = u32;
> 
> struct Uid(u32);
> 
> impl Uid {
>     fn new(raw_uid: RawUid) -> Option<Self>;
> }
> ```
> 
> So this should be:
> 
> ```rust
> let Some(uid) = Uid::new(raw_uid) else {
>     return_errno_with_message!(Errno::EINVAL, "the UID is invalid");
> };
> ```
> 
> This pattern should be used when implementing the `setuid` and `chown` system calls instead.

Yes, you are right. My tool haven't audited `setuid` yet.

Shall I make the refactor in this PR?

lrh2000:
> Shall I make the refactor in this PR?

I don't really have a strong opinion. But I think you can give it a try. If it works, it's preferred that you do it in this PR.

You can also check the existing code, `RawFileDesc` and `FileDesc`, for a reference. I believe they're similar to `RawUid` and `Uid`.

Currently, I have a question: What would happen if we executed a setuid binary file owned by a `raw_uid=-1` user? Note that although `fchown` rejects setting the file UID to `-1`, the UID must be stored on the disk in raw format. Therefore, a `raw_uid=-1` file can theoretically exist if we edit the disk image. So, how should we behave?

Marsman1996:
> Currently, I have a question: What would happen if we executed a setuid binary file owned by a `raw_uid=-1` user? Note that although `fchown` rejects setting the file UID to `-1`, the UID must be stored on the disk in raw format. Therefore, a `raw_uid=-1` file can theoretically exist if we edit the disk image. So, how should we behave?

Seems it is ignored.

https://elixir.bootlin.com/linux/v6.15/source/fs/mnt_idmapping.c#L80-L97
```C
vfsuid_t make_vfsuid(struct mnt_idmap *idmap,
		     struct user_namespace *fs_userns,
		     kuid_t kuid)
{
	uid_t uid;

	if (idmap == &nop_mnt_idmap)
		return VFSUIDT_INIT(kuid);
	if (idmap == &invalid_mnt_idmap)
		return INVALID_VFSUID;
	if (initial_idmapping(fs_userns))
		uid = __kuid_val(kuid);
	else
		uid = from_kuid(fs_userns, kuid);
	if (uid == (uid_t)-1)
		return INVALID_VFSUID;
	return VFSUIDT_INIT_RAW(map_id_down(&idmap->uid_map, uid));
}
```

https://elixir.bootlin.com/linux/v6.15/source/include/linux/fs.h#L1493-L1497
```C
static inline vfsuid_t i_uid_into_vfsuid(struct mnt_idmap *idmap,
					 const struct inode *inode)
{
	return make_vfsuid(idmap, i_user_ns(inode), inode->i_uid);
}
```

https://elixir.bootlin.com/linux/v6.15/source/include/linux/mnt_idmapping.h#L158-L162
```C
static inline bool vfsuid_has_mapping(struct user_namespace *userns,
				      vfsuid_t vfsuid)
{
	return from_kuid(userns, AS_KUIDT(vfsuid)) != (uid_t)-1;
}
```

https://elixir.bootlin.com/linux/v6.15/source/fs/exec.c#L1642-L1681
```C
static void bprm_fill_uid(struct linux_binprm *bprm, struct file *file)
{
	/* Handle suid and sgid on files */
	struct mnt_idmap *idmap;
	struct inode *inode = file_inode(file);
	unsigned int mode;
	vfsuid_t vfsuid;
	vfsgid_t vfsgid;
	int err;

	if (!mnt_may_suid(file->f_path.mnt))
		return;

	if (task_no_new_privs(current))
		return;

	mode = READ_ONCE(inode->i_mode);
	if (!(mode & (S_ISUID|S_ISGID)))
		return;

	idmap = file_mnt_idmap(file);

	/* Be careful if suid/sgid is set */
	inode_lock(inode);

	/* Atomically reload and check mode/uid/gid now that lock held. */
	mode = inode->i_mode;
	vfsuid = i_uid_into_vfsuid(idmap, inode);
	vfsgid = i_gid_into_vfsgid(idmap, inode);
	err = inode_permission(idmap, inode, MAY_EXEC);
	inode_unlock(inode);

	/* Did the exec bit vanish out from under us? Give up. */
	if (err)
		return;

	/* We ignore suid/sgid if there are no mappings for them in the ns */
	if (!vfsuid_has_mapping(bprm->cred->user_ns, vfsuid) ||
	    !vfsgid_has_mapping(bprm->cred->user_ns, vfsgid))
		return;
```

kernel/src/syscall/setregid.rs
```diff
@@ -7,17 +7,8 @@ use crate::{
...
-        None
-    };
-
-    let egid = if egid >= 0 {
-        Some(Gid::new(egid.cast_unsigned()))
-    } else {
-        None
-    };
+    let rgid = Gid::new(rgid.cast_unsigned());
+    let egid = Gid::new(egid.cast_unsigned());
```
lrh2000: Following the existing conventions:
```rust
pub fn sys_some_system_call(raw_uid: RawUid) {
    let uid = Uid::new(raw_uid);
}
```

Same for many other system calls in modified in this PR.

kernel/src/syscall/setgid.rs
```diff
@@ -7,11 +7,9 @@ use crate::{
 };
 
 pub fn sys_setgid(gid: i32, ctx: &Context) -> Result<SyscallReturn> {
-    if gid < 0 {
-        return_errno_with_message!(Errno::EINVAL, "GIDs cannot be negative");
-    }
-
-    let gid = Gid::new(gid.cast_unsigned());
+    let Some(gid) = Gid::new(gid.cast_unsigned()) else {
```
lrh2000: Use `Gid::try_from()?`.

Same for `setuid`/`setgid`/`setgroups`.

kernel/src/fs/fs_impls/exfat/inode.rs
```diff
@@ -1441,8 +1441,8 @@ impl Inode for ExfatInode {
             type_: inner.inode_type,
             mode: inner.make_mode(),
             nr_hard_links: nlinks,
-            uid: Uid::new(inner.fs().mount_option().fs_uid as u32),
-            gid: Gid::new(inner.fs().mount_option().fs_gid as u32),
+            uid: Uid::new(inner.fs().mount_option().fs_uid as u32).unwrap_or(Uid::OVERFLOW),
+            gid: Gid::new(inner.fs().mount_option().fs_gid as u32).unwrap_or(Gid::OVERFLOW),
```
lrh2000: If we repeat this pattern many times, we should create a separate method, e.g., `Uid::new_or_overflow`.

However, I am not convinced that this is the right way to handle invalid UIDs from the FS side. As I said before, the FS inode can contain invalid UIDs if we edit the disk image. Should we fall back to `Uid::OVERFLOW` here? Another solution is to try to store/report raw IDs in the FS module and reject them when applied (e.g., when handling a `setuid` program).

Especially, you already figure out that this case is ignored? https://github.com/asterinas/asterinas/pull/3355#issuecomment-4670234759

kernel/src/syscall/get_priority.rs
```diff
@@ -101,7 +101,12 @@ impl PriorityTarget {
                 let uid = if who == 0 {
                     ctx.posix_thread.credentials().ruid()
                 } else {
-                    Uid::new(who)
+                    // Linux does not validate the UID here; an unmapped UID
+                    // simply finds no matching processes and returns ESRCH.
+                    let Some(uid) = Uid::new(who) else {
```
lrh2000: - `an unmapped UID` -> `An invalid UID`
 - Remove unnecessary indention.
```rust
                let uid = if who == 0 {
                    ctx.posix_thread.credentials().ruid()
                } else if let Some(uid) = Uid::new(who) {
                    uid
                } else {
                    // Linux does not validate the UID here. An invalid UID
                    // simply finds no matching processes and returns `ESRCH`.
                    return_errno_with_message!(Errno::ESRCH, "invalid UID");
                }
```

kernel/src/process/credentials/user.rs
```diff
@@ -6,42 +6,50 @@ use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
...
+    pub const OVERFLOW: Uid = Self(65534);
 
     pub const fn new_root() -> Self {
         Self(ROOT_UID)
     }
 
-    pub const fn new(uid: u32) -> Self {
-        Self(uid)
+    /// Creates a `Uid` from a raw value, returning `None` if the value is the
+    /// invalid sentinel (`u32::MAX`, i.e. `(uid_t)-1` in Linux).
```
lrh2000:
```suggestion
    /// invalid sentinel (i.e., [`Self::RAW_INVALID`]).
```

Same for `group.rs`.

lrh2000 (commented):
Thanks for the updates!

kernel/src/fs/fs_impls/exfat/inode.rs
Marsman1996: In the latest commit I add an unchecked `Uid/Gid::from_raw`.

Shall I also change`fs_uid: usize,` to `fs_uid: RawUid`?

lrh2000: > In the latest commit I add an unchecked `Uid/Gid::from_raw`.

A separate type should be created, such as `InodeUid`, or whatever the more appropriate name is. (Reusing `RawUid` is also an option, but it might not be ideal as it is not strongly typed.)

The UID is always meant to be valid. Don't break the type invariant.

> Shall I also change`fs_uid: usize,` to `fs_uid: RawUid`?

`RawUid` is certainly better. The code is dead code and the field is never populated with a value other than `Default::default()`, so it's difficult to say whether using `Uid` would be more suitable (The type is named `MountOptions`. It does not come from the disk image...)

PR #3361, opened 2026-06-10
Mark ACPI region reclaimable for Multiboot

### LINKED ISSUE #3030: ACPI tables cannot be parsed when memory is low under multiboot protocol

StevenJiang1110:
Fixes #3030.

For more background, you can see the discussion under #3030.

The address of the ACPI tables can potentially be higher than all usable memory regions, particularly when the boot memory is small (e.g., less than 4GB). In such cases, since the kernel page table's linear mapping only covers up to the last usable memory region, the ACPI table area may not be mapped, leading to a kernel page fault and panic when accessed. This can be reproduced by running `make run_kernel BOOT_PROTOCOL=multiboot MEM=2G`, and the panic message can be observed in `qemu-serial.log`.

For boot protocols other than Multiboot, the ACPI table regions are passed in as boot parameters, which allows us to know the ACPI region ranges and establish the necessary mappings accordingly. However, the Multiboot protocol does not include this information in its boot parameters, so we resort to searching the BIOS region to locate the ACPI tables. This approach is based on the existing `get_acpi_tables` implementation, but it remains unsound — especially when booting via UEFI. I have added comments to document this limitation.

### DIFF
--- a/ostd/src/arch/x86/boot/multiboot/mod.rs
+++ b/ostd/src/arch/x86/boot/multiboot/mod.rs
@@ -62,7 +62,7 @@ fn parse_initramfs(mb1_info: &MultibootLegacyInfo) -> Option<&[u8]> {
 fn parse_acpi_arg(_mb1_info: &MultibootLegacyInfo) -> BootloaderAcpiArg {
     // Multiboot v1 is BIOS-oriented and does not define a standard field for
     // the RSDP or the EFI System Table.
-    BootloaderAcpiArg::NotProvided
+    BootloaderAcpiArg::ScanBios
 }
 
 fn parse_framebuffer_info(mb1_info: &MultibootLegacyInfo) -> Option<BootloaderFramebufferArg> {
--- a/ostd/src/arch/x86/boot/multiboot/mod.rs
+++ b/ostd/src/arch/x86/boot/multiboot/mod.rs
@@ -1,8 +1,11 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use core::arch::global_asm;
+use core::{arch::global_asm, num::NonZeroUsize};
+
+use acpi::rsdp::Rsdp;
 
 use crate::{
+    arch::kernel::acpi::AcpiMemoryHandler,
     boot::{
         BootloaderAcpiArg, BootloaderFramebufferArg,
         memory_region::{MemoryRegion, MemoryRegionArray, MemoryRegionType},
@@ -57,8 +60,8 @@ fn parse_initramfs(mb1_info: &MultibootLegacyInfo) -> Option<&[u8]> {
 }
 
 fn parse_acpi_arg(_mb1_info: &MultibootLegacyInfo) -> BootloaderAcpiArg {
-    // The multiboot protocol does not contain RSDP address.
-    // TODO: What about UEFI?
+    // Multiboot v1 is BIOS-oriented and does not define a standard field for
+    // the RSDP or the EFI System Table.
     BootloaderAcpiArg::NotProvided
 }
 
@@ -78,13 +81,30 @@ fn parse_framebuffer_info(mb1_info: &MultibootLegacyInfo) -> Option<BootloaderFr
 fn parse_memory_regions(mb1_info: &MultibootLegacyInfo) -> MemoryRegionArray {
     let mut regions = MemoryRegionArray::new();
 
+    // The Multiboot protocol does not report ACPI regions explicitly. An ACPI
+    // region may live above the highest usable physical memory region when the
+    // total memory size is below 2G. We therefore mark it as reclaimable to
+    // include it in the linear mappings and allow access.
+    //
+    // FIXME: The ACPI specification does not guarantee that all ACPI tables
+    // are contiguous and do not cross region boundaries. See ACPI 6.4, Section
+    // 5.1, "Overview of the System Description Table Architecture".
+    // A full ACPI table graph scan may eventually be unavoidable.
+    let acpi_root_table_address = find_acpi_root_table_address();
+
     // Add the regions in the multiboot protocol.
     for entry in mb1_info.get_memory_map() {
-        let region = MemoryRegion::new(
-            entry.base_addr().try_into().unwrap(),
-            entry.length().try_into().unwrap(),
-            entry.memory_type(),
-        );
+        let base = entry.base_addr().try_into().unwrap();
+        let len = entry.length().try_into().unwrap();
+        let mut typ = entry.memory_type();
+        if typ == MemoryRegionType::Reserved
+            && acpi_root_table_address
+                .is_some_and(|address| (base..(base + len)).contains(&address.get()))
+        {
+            typ = MemoryRegionType::Reclaimable;
+        }
+
+        let region = MemoryRegion::new(base, len, typ);
         regions.push(region).unwrap();
     }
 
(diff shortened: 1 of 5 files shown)


ostd/src/arch/x86/boot/multiboot/mod.rs
```diff
@@ -121,6 +155,30 @@ fn parse_memory_regions(mb1_info: &MultibootLegacyInfo) -> MemoryRegionArray {
...
+    //
+    // FIXME: This call is not actually sound yet
+    // because we do not distinguish UEFI boots from BIOS boots.
+    // On UEFI systems,
+    // the RSDP should be obtained from the EFI System Table's
+    // configuration table.
+    // Probing these BIOS-specific addresses on UEFI systems
+    // may have unexpected side effects.
+    // The current `get_acpi_tables` implementation
+    // has the same limitation.
```
junyang-zh: Maybe there's no problem here since UEFI firmware cannot boot a multiboot kernel. And here this is multiboot specific. Am I wrong?

ostd/src/arch/x86/kernel/acpi/mod.rs
```diff
@@ -75,7 +75,18 @@ pub(crate) fn get_acpi_tables() -> Option<&'static AcpiTables<AcpiMemoryHandler>
...
+                //
+                // FIXME: This call is not actually sound yet
+                // because we do not distinguish UEFI boots from BIOS boots.
+                // On UEFI systems,
+                // the RSDP should be obtained from the EFI System Table's
+                // configuration table.
+                // Probing these BIOS-specific addresses on UEFI systems
+                // may have unexpected side effects.
+                // The current `find_acpi_root_table_address` implementation
+                // has the same limitation.
```
junyang-zh: Perhaps a better solution is to ensure `BootloaderAcpiArg::NotProvided` never happens in `EARLY_INFO` if we are booted through UEFI. Seems not very hard. In contrast, the FIXME you write here is... way too verbose.

ostd/src/arch/x86/boot/multiboot/mod.rs
StevenJiang1110: I think you are right. After studying the [Multiboot specification](https://www.gnu.org/software/grub/manual/multiboot/multiboot.html#Specification), it appears that Multiboot is BIOS-oriented. While it does not explicitly state that it is BIOS-only, most of the specification's descriptions implicitly assume a BIOS environment. I have updated the documentation accordingly.

ostd/src/arch/x86/kernel/acpi/mod.rs
StevenJiang1110: I have added a new variant, `ScanBios`, to `BootloaderAcpiArg` to distinguish between the two cases.

ostd/src/arch/x86/boot/multiboot/mod.rs
```diff
@@ -78,13 +81,44 @@ fn parse_framebuffer_info(mb1_info: &MultibootLegacyInfo) -> Option<BootloaderFr
...
+    // and later mark the reserved region that contains it as reclaimable.
+    // At this point, the boot page table is still active
+    // and linearly maps physical addresses below 4 GiB,
+    // so the RSDP structure found by the BIOS scan is accessible.
+    //
+    // FIXME: The ACPI specification does not guarantee
+    // that all ACPI tables are contiguous.
+    // See ACPI 6.4, Section 5.1,
+    // "Overview of the System Description Table Architecture".
+    // A full ACPI table graph scan may eventually be unavoidable.
```
junyang-zh: This paragraph is verbose and wraps badly. Don't repeat yourself, and try to wrap to a length closest to, but not exceeding, 80 chars.

```
    // The Multiboot protocol does not report ACPI regions explicitly. An ACPI
    // region may live above the highest usable physical memory memory region
    // when the total memory size is below 2G. We therefore mark it as
    // reclaimable to include it in the linear mappings to allow access.
    //
    // FIXME: The ACPI specification does not guarantee that all ACPI tables
    // are contiguous and do not cross region boundaries. See ACPI 6.4, Section
    // 5.1, "Overview of the System Description Table Architecture".
    // A full ACPI table graph scan may eventually be unavoidable.
```

junyang-zh (approved):

PR #3364, opened 2026-06-10
Implement group scheduling for cgroup

Plucky923:
# Summary

This PR implements **CPU-weight-based group scheduling** for **cgroups**.

Asterinas already exposed CPU controller files such as `cpu.weight` and `cpu.max`. However, the **fair scheduler** still treated all runnable threads as **independent scheduling entities**. As a result, container runtimes could configure CPU settings through `cgroupfs`, but `cpu.weight` did not affect **cgroup-level CPU allocation**.

This PR introduces **task groups** into the fair scheduler. It maps **CPU cgroups** to **scheduler task groups**. The scheduler can now apportion CPU time among groups according to their weights. It then selects runnable tasks inside the chosen group.

This PR only enforces **`cpu.weight`**. CPU bandwidth control for **`cpu.max`** is left for future work.

# Motivation

Container runtimes such as **Podman** and **Kata Containers** rely on **cgroups** as the basic interface for CPU resource management.Podman exposes CPU quota, period, and share-like options. These options are mapped to cgroup CPU controller files such as `cpu.max` and `cpu.weight`. Kata Containers also places CPU-constrained containers into CPU cgroups.

Without **group scheduling**, these cgroup settings are accepted by the `cgroupfs` interface, but they do not affect actual CPU allocation. A container with more runnable tasks can still receive more CPU time simply because each task competes directly in the **root runqueue**.

This PR fixes that gap by making **cgroup CPU weight effective in the scheduler**.

# What is Group Scheduling

A cgroup describes a group of processes and its CPU policy. For example, a container runtime can create a cgroup for a container, move the container's processes into it through `cgroup.procs`, and configure its relative CPU weight through `cpu.weight`. However, these cgroup files are only the **userspace control interface**. To make the policy effective, the scheduler must translate the cgroup into an internal **schedulable object**.

Without group scheduling, the fair scheduler sees only **individual runnable tasks**. If three tasks have the same weight, each task receives roughly one third of the CPU time. This is fair at the **task level**, but not at the **workload level**. If container A has two runnable tasks and container B has one runnable task, container A receives about two thirds of the CPU time simply because it has more tasks.

**Group scheduling changes the unit of fairness.** Instead of placing all tasks directly into the root runqueue, the scheduler represents each cgroup as a **task group**. A task group has its own **scheduling entity**, which competes with other tasks or task groups in the parent runqueue. Once a task group is selected, the scheduler enters that group and selects one runnable task from the group's own runqueue.

Conceptually, scheduling changes from this:

```text
root runqueue
├── task_1
├── task_2
└── task_3
```

to this:

```text
root runqueue
├── group_a.sched_entity
│   └── group_a runqueue
│       ├── task_1
│       └── task_2
└── group_b.sched_entity
    └── group_b runqueue
        └── task_3
```

With this structure, CPU time is first apportioned among groups according to their weights. The CPU time received by a group is then shared among the tasks inside that group.

For example, if container A has two runnable tasks, container B has one runnable task, and both containers have the same `cpu.weight`, the scheduler first gives each container roughly half of the CPU time. The two tasks inside container A then share container A's half, so each receives roughly one quarter. The single task inside container B receives roughly one half.

Therefore, CPU allocation is no longer determined by the number of runnable tasks in each container. Instead, it follows the **workload-level policy** expressed by the **cgroup CPU controller**.

# Design

This PR integrates the **cgroup CPU controller** with the **FAIR scheduler** by introducing **scheduler task groups** and **hierarchical FAIR runqueues**.

The implementation is based on three main pieces:

1. a **TaskGroup abstraction** that represents a scheduler-visible CPU cgroup;
2. **fair scheduling entities** that can represent either a thread or a task group;
3. **cgroupfs integration** that keeps thread and task group membership and `cpu.weight` synchronized with scheduler state.

## Cgroup to TaskGroup Mapping

Each CPU cgroup owns a scheduler **TaskGroup**.

The root CPU cgroup uses the global **root task group**. A non-root CPU cgroup creates a child TaskGroup under its parent cgroup's task group. This mirrors the cgroup hierarchy in the scheduler, while keeping cgroup management and scheduler internals separated.

However, the TaskGroup owned by a cgroup is not always its **effective scheduling group**. In cgroup v2, a cgroup receives CPU control only when its parent enables **`+cpu`** through `cgroup.subtree_control`.

If `+cpu` is enabled for a cgroup, tasks attached to that cgroup are scheduled under its own TaskGroup. If `+cpu` is not enabled, those tasks use the nearest ancestor that has effective CPU control. If no such ancestor exists, they use the root task group.

As a result, changing `cgroup.subtree_control` can change the **effective scheduling group** of descendant tasks even when the tasks themselves are not moved. When CPU control is enabled or disabled, the implementation resynchronizes descendant processes, recomputes their effective task groups, and re-enqueues queued tasks through the updated scheduling hierarchy.

## TaskGroup

**TaskGroup** represents a workload in the FAIR scheduler.

Each task group contains **per-CPU scheduler state**:

- per-CPU FAIR runqueues for direct member threads and child group entities;
- per-CPU FAIR entity attributes for the group entity that appears in the parent runqueue;
- a parent pointer used to walk the scheduling hierarchy.

The root task group is special: it has per-CPU FAIR runqueues, but it has no parent and therefore no group entity attributes.

For a non-root task group, its **group entity** is what competes in the parent group's runqueue. Its own runqueue stores the runnable threads and child group entities that belong to that group.

Conceptually:

```text
root task group
└── root fair_rq[cpu]
    ├── task_a.se
    └── group_b.se
        └── group_b fair_rq[cpu]
            ├── task_b1.se
            └── task_b2.se
```

## FAIR Entity

The FAIR scheduler now treats both **threads** and **task groups** as scheduling entities.

A FAIR entity can be either:

```text
Thread(task)
Group(task_group)
```

Both entity types have FAIR scheduling attributes such as **weight** and **vruntime**. This allows the existing **vruntime-based scheduling logic** to be reused for both task-level fairness and group-level fairness.

The runqueue orders entities by `(vruntime, entity_id)`. The `vruntime` provides fairness, while the stable entity ID provides deterministic tie-breaking.

## Weight

The cgroup v2 `cpu.weight` value is mapped into a **FAIR scheduler weight**.

The default cgroup CPU weight is `100`, which corresponds to the scheduler's normal FAIR weight `1024`. A cgroup weight is scaled as:

```text
scheduler_weight = cpu.weight * 1024 / 100
```

When userspace writes `cpu.weight`, the CPU controller validates and stores the new value. The cgroup node write path then synchronizes this value into the corresponding task group. If the group entity is already queued in a parent FAIR runqueue, the runqueue entry is refreshed so that future scheduling decisions use the new weight.

This makes `cpu.weight` effective for runnable cgroups without requiring userspace to recreate or remigrate the processes.

## Enqueue

When a FAIR task is enqueued, the scheduler checks the thread's current **task group**.

If the task belongs directly to the runqueue's task group, it is enqueued as a thread entity in that runqueue.

If the task belongs to a descendant task group, the scheduler first enqueues the task into the descendant group's leaf runqueue. If that leaf group was previously empty, the scheduler activates the corresponding group entity in its parent runqueue. This activation can propagate upward until an already-active ancestor is reached.

This ensures that a task group competes in its parent runqueue only while it has **runnable work**.

## Pick Next Task

Task selection proceeds from **parent to child**.

At each FAIR runqueue, the scheduler selects the entity with the smallest `vruntime`.

If the selected entity is a thread, that thread is returned as the next task to run.

If the selected entity is a task group, the scheduler enters the child group's FAIR runqueue and repeats the same selection process. This continues until a concrete runnable thread is found.

## Vruntime Accounting

When a thread runs, the scheduler updates two kinds of `vruntime`:

1. the thread's own FAIR `vruntime` in its group runqueue;
2. the group entity `vruntime` for each non-root ancestor task group.

Each entity's `vruntime` is updated according to its weight:

```text
vruntime += runtime_delta * 1024 / weight
```

A larger weight causes `vruntime` to grow more slowly, so that entity remains eligible to run more often over time.

The preemption decision is also derived from the same hierarchy. If either the thread entity or one of its ancestor group entities exceeds its fair time slice, the scheduler can request preemption.

## Cgroup Migration and Controller Changes

When a process is moved through `cgroup.procs`, the implementation updates the task group of all threads in that process.

If a task is already queued in an old FAIR runqueue, the scheduler removes it from that runqueue and re-enqueues it through the new task group. During this update, the scheduler preserves the task's `vruntime` lag relative to the source runqueue's `min_vruntime`. This prevents migration from resetting the task's fair scheduling position.

Running tasks are not forcibly dequeued from the current CPU. Sleeping tasks keep the updated task-group pointer and use it when they later enter the enqueue path.

The same mechanism is reused when `cgroup.subtree_control` changes the effective CPU controller state of descendant cgroups.

## Fork and Thread Inheritance

New threads inherit the task group of their parent thread.

This is required because threads in the same process should remain under the same effective CPU control unless the process is explicitly moved to another cgroup.

For forked processes, cgroup membership is applied through the normal cgroup membership path. This path also assigns the effective CPU task group to the new process.

### DIFF
--- a/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
+++ b/kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use alloc::sync::Arc;
+use alloc::{sync::Arc, vec::Vec};
 use core::sync::atomic::{AtomicU32, Ordering};
 
 use aster_systree::{Error, MAX_ATTR_SIZE, Result, SysAttrSetBuilder, SysPerms, SysStr};
@@ -9,7 +9,7 @@ use ostd::{
     cpu::CpuId,
     mm::{VmReader, VmWriter},
     sync::SpinLock,
-    task::atomic_mode::AsAtomicModeGuard,
+    task::{Task, atomic_mode::AsAtomicModeGuard},
     timer::Jiffies,
     warn,
 };
@@ -17,6 +17,8 @@ use ostd::{
 use crate::{
     fs::cgroupfs::systree_node::{CgroupSysNode, CgroupSystem},
     process::Process,
+    sched::{self, TaskGroup},
+    thread::AsThread,
     util::ReadCString,
 };
 
@@ -28,6 +30,8 @@ use crate::{
 pub struct CpuController {
     /// Persistent CPU usage accounting for this cgroup.
     stats: CpuStats,
+    /// Scheduler state used by hierarchical fair group scheduling for this CPU cgroup.
+    task_group: Arc<TaskGroup>,
     /// Optional CPU resource control attributes exposed only when `+cpu` is active.
     control: Option<CpuControl>,
 }
@@ -48,16 +52,16 @@ struct CpuStats {
 struct CpuControl {
     /// A value stores the configured relative CPU share for scheduler integration.
     weight: AtomicU32,
-    /// A value stores the configured CPU bandwidth limit for scheduler integration.
+    /// A value stores the configured CPU bandwidth limit.
     max: SpinLock<CpuMax>,
 }
 
 /// A CPU bandwidth limit.
 #[derive(Clone, Copy)]
 struct CpuMax {
-    /// The configured CPU runtime budget for bandwidth enforcement.
+    /// The configured CPU runtime budget.
     quota_usec: u64,
-    /// The window over which the CPU runtime budget is intended to apply.
+    /// The window over which the CPU runtime budget is configured.
     period_usec: u64,
 }
 
@@ -80,19 +84,21 @@ impl CpuController {
         }
     }
 
-    fn new(is_active: bool) -> Self {
+    fn new(is_active: bool, task_group: Arc<TaskGroup>) -> Self {
         Self {
             stats: CpuStats::new(),
+            task_group,
             control: is_active.then(CpuControl::new),
         }
     }
 
-    /// Initializes CPU usage statistics from the previous sub-controller.
+    /// Inherits persistent state from the previous sub-controller.
     ///
     /// Toggling `+cpu` must reset some controller attributes such as `cpu.weight`
     /// and `cpu.max`, but `cpu.stat` remains visible and continues to report the
-    /// accumulated CPU usage.
-    pub(super) fn init_stats(&mut self, previous: &Self) {
+    /// accumulated CPU usage. The scheduler state is also reused so that existing
+    /// tasks remain associated with the same CPU cgroup scheduling entity.
+    pub(super) fn inherit_state_from(&mut self, previous: &Self) {
         // No one cares which CPU the count is on. Therefore, choose the BSP for simplicity.
         self.stats.user.add_on_cpu(
             CpuId::bsp(),
@@ -102,10 +108,25 @@ impl CpuController {
             CpuId::bsp(),
             isize::try_from(previous.stats.system.sum_all_cpus()).unwrap_or(isize::MAX),
         );
+
+        self.task_group = previous.task_group.clone();
     }
 
     fn control(&self) -> Result<&CpuControl> {
         self.control.as_ref().ok_or(Error::AttributeError)
+    }
+
+    fn sync_weight(&self) {
+        let Some(control) = &self.control else {
+            return;
+        };
+
+        let weight = control.weight.load(Ordering::Relaxed);
+        self.task_group.update_weight(weight);
+    }
+
+    pub(super) fn is_active(&self) -> bool {
+        self.control.is_some()
     }
 
     fn read_cpu_stat(&self, printer: &mut VmPrinter) -> Result<usize> {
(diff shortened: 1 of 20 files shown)


kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -102,12 +108,27 @@ impl CpuController {
             CpuId::bsp(),
             isize::try_from(previous.stats.system.sum_all_cpus()).unwrap_or(isize::MAX),
         );
+
+        self.task_group = previous.task_group.clone();
```
cchanging, line 112: Could you document the inheritance rules for task groups somewhere?

My understanding is that only an enabled CPU sub-controller should create a child task group; otherwise, it should directly inherit the parent’s task group. The current logic looks a bit strange to me, because `inherit_state_from` seems to overwrite the `task_group` field every time.

Besides, the logic mentioned above could all be placed inside `inherit_state_from`; it doesn’t necessarily need to be in the `new` function.

kernel/src/fs/fs_impls/cgroupfs/controller/cpu.rs
```diff
@@ -149,17 +170,18 @@ impl CpuController {
 }
 
 impl CpuControl {
+    const DEFAULT_WEIGHT: u32 = 100;
```
cchanging, line 173: Why move it here?

kernel/src/thread/mod.rs
```diff
@@ -85,22 +88,35 @@ pub struct Thread {
     /// Thread CPU affinity
     cpu_affinity: AtomicCpuSet,
     sched_attr: SchedAttr,
+    /// The task group this thread belongs to.
+    task_group: Rcu<Arc<TaskGroup>>,
```
cchanging, line 92: Do we really need to store an extra copy of `task_group` in `Thread`?

`Process` already contains the cgroup information, so I think we can access it via `process -> cgroup -> task_group`. That would also help avoid operations like `apply_cpu_control_to_process` that you introduced in the cgroup module, which looks unnecessarily complicated and not really needed.

cchanging (commented):
Thanks for further improving the cgroup functionality! First, I’d like to raise a few questions about the related changes in the cgroup module.

PR #3368, opened 2026-06-10
Fix eventfd2 initval type width and ULLONG_MAX write handling

Marsman1996:
refer: https://elixir.bootlin.com/linux/v6.15/source/fs/eventfd.c

- The Linux `eventfd2` syscall declares `initval` as `unsigned int` (32 bits).
  ```c
  SYSCALL_DEFINE2(eventfd2, unsigned int, count, int, flags)
  ```
- for u64::MAX (ULLONG_MAX) in the write method, returning EINVAL immediately
  ```c
	  if (ucnt == ULLONG_MAX)
		  return -EINVAL;
  ```

### DIFF
--- a/kernel/src/syscall/eventfd.rs
+++ b/kernel/src/syscall/eventfd.rs
@@ -30,13 +30,13 @@ use crate::{
     process::signal::{PollHandle, Pollable, Pollee},
 };
 
-pub fn sys_eventfd(init_val: u64, ctx: &Context) -> Result<SyscallReturn> {
+pub fn sys_eventfd(init_val: u32, ctx: &Context) -> Result<SyscallReturn> {
     debug!("init_val = 0x{:x}", init_val);
 
     do_sys_eventfd2(init_val, Flags::empty(), ctx)
 }
 
-pub fn sys_eventfd2(init_val: u64, flags: u32, ctx: &Context) -> Result<SyscallReturn> {
+pub fn sys_eventfd2(init_val: u32, flags: u32, ctx: &Context) -> Result<SyscallReturn> {
     debug!("raw flags = {}", flags);
     let flags = Flags::from_bits(flags)
         .ok_or_else(|| Error::with_message(Errno::EINVAL, "unknown flags"))?;
@@ -45,8 +45,8 @@ pub fn sys_eventfd2(init_val: u64, flags: u32, ctx: &Context) -> Result<SyscallR
     do_sys_eventfd2(init_val, flags, ctx)
 }
 
-fn do_sys_eventfd2(init_val: u64, flags: Flags, ctx: &Context) -> Result<SyscallReturn> {
-    let event_file = EventFile::new(init_val, flags);
+fn do_sys_eventfd2(init_val: u32, flags: Flags, ctx: &Context) -> Result<SyscallReturn> {
+    let event_file = EventFile::new(init_val as u64, flags);
     let file_table = ctx.thread_local.borrow_file_table();
     let mut file_table_locked = file_table.unwrap().write();
     let fd_flags = if flags.contains(Flags::EFD_CLOEXEC) {
@@ -192,6 +192,10 @@ impl FileLike for EventFile {
         }
 
         let supplied_value = reader.read_val::<u64>()?;
+
+        if supplied_value == u64::MAX {
+            return_errno_with_message!(Errno::EINVAL, "write value must not be ULLONG_MAX");
+        }
 
         // Try to add counter val at first
         if self.add_counter_val(supplied_value).is_ok() {
--- a/test/initramfs/src/regression/io/eventfd2/eventfd2_err.c
+++ b/test/initramfs/src/regression/io/eventfd2/eventfd2_err.c
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include <sys/eventfd.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "../../common/test.h"
+
+static int efd;
+
+FN_SETUP(create_eventfd_with_large_initval)
+{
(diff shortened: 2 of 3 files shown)


Marsman1996:
# `eventfd2` initval parameter accepts 64-bit values instead of truncating to 32-bit

### Describe the bug
The Linux `eventfd2` syscall declares `initval` as `unsigned int` (32 bits). On x86-64, values larger than UINT32_MAX passed via the syscall ABI are truncated to 32 bits. Asterinas declares `init_val` as `u64` in `kernel/src/syscall/eventfd.rs:39`, allowing the counter to be initialized with values exceeding UINT32_MAX.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>
#include <sys/eventfd.h>
#include <stdint.h>
#include <sys/syscall.h>

/*
 * Test: param.initval.type_width
 *
 * Linux eventfd2 takes initval as "unsigned int" (32-bit).
 * On x86-64, the kernel sign-extends/zero-extends arguments per ABI, so
 * passing a value > UINT32_MAX should be truncated to 32 bits.
 *
 * Asterinas takes init_val as u64, so it would accept values > UINT32_MAX.
 */
int main(void) {
    int fd;
    ssize_t s;
    uint64_t val;

    /* Pass 0x100000001 as initval via raw syscall.
     * If the kernel treats it as unsigned int, it becomes 1 (truncated).
     * If treated as u64, the counter would be 0x100000001 (4294967297).
     */
    fd = syscall(SYS_eventfd2, 0x100000001ULL, 0);
    if (fd < 0) {
        printf("VERDICT: UNKNOWN — eventfd2 failed: %s\n", strerror(errno));
        return 1;
    }

    s = read(fd, &val, sizeof(val));
    if (s != 8) {
        printf("VERDICT: UNKNOWN — read failed\n");
        close(fd);
        return 1;
    }

    if (val == 1) {
        printf("VERDICT: LINUX behavior — initval truncated to u32 (read back %lu)\n", (unsigned long)val);
    } else if (val == 0x100000001ULL) {
        printf("VERDICT: ASTERINAS behavior — initval accepted as u64 (read back %lu)\n", (unsigned long)val);
    } else {
        printf("VERDICT: UNKNOWN — read back unexpected value 0x%lx\n", (unsigned long)val);
    }

    close(fd);
    return 0;
}
```

### Expected behavior
When `eventfd2` is called with `initval = 0x100000001`, the kernel should truncate it to `unsigned int`, resulting in an initial counter value of 1. A subsequent read should return 1.

# `eventfd2` write of ULLONG_MAX returns EAGAIN instead of EINVAL

### Describe the bug
When writing the value 0xFFFFFFFFFFFFFFFF (ULLONG_MAX) to an eventfd file descriptor, Linux immediately returns EINVAL regardless of the blocking mode. Asterinas instead returns EAGAIN in nonblocking mode, or blocks indefinitely in blocking mode. The root cause is in `kernel/src/syscall/eventfd.rs` lines 196-203: the `write` method does not check whether the supplied value is inherently invalid (ULLONG_MAX) before falling through to the overflow/blocking path.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>
#include <sys/eventfd.h>
#include <stdint.h>

/*
 * Test: error.EINVAL.write_max_value
 *
 * Linux returns EINVAL when attempting to write ULLONG_MAX (0xFFFFFFFFFFFFFFFF)
 * to an eventfd, regardless of whether the fd is blocking or nonblocking.
 *
 * Asterinas does not have an explicit check for ULLONG_MAX writes. Instead,
 * it falls through to the overflow check (add_counter_val fails), and then:
 * - In nonblocking mode: returns EAGAIN
 * - In blocking mode: blocks forever (since the counter can never make room for ULLONG_MAX)
 */
int main(void) {
    int fd;
    ssize_t s;
    uint64_t val;

    /* Test with nonblocking fd */
    fd = eventfd(0, EFD_NONBLOCK);
    if (fd < 0) { perror("eventfd"); return 1; }

    val = UINT64_MAX;  /* 0xFFFFFFFFFFFFFFFF */
    errno = 0;
    s = write(fd, &val, sizeof(val));
    if (s == -1 && errno == EINVAL) {
        printf("VERDICT: LINUX behavior — write of ULLONG_MAX returns EINVAL (errno=%d)\n", errno);
    } else if (s == -1 && errno == EAGAIN) {
        printf("VERDICT: ASTERINAS behavior — write of ULLONG_MAX returns EAGAIN (errno=%d)\n", errno);
    } else {
        printf("VERDICT: UNKNOWN behavior — write returned s=%zd, errno=%s\n", s, strerror(errno));
    }

    close(fd);
    return 0;
}
```

### Expected behavior
Writing the value 0xFFFFFFFFFFFFFFFF to an eventfd should immediately return -1 with errno set to EINVAL, regardless of whether the file descriptor is in blocking or nonblocking mode. This value is unconditionally invalid per the Linux kernel implementation.

StevenJiang1110 (approved):

cqs21 (approved):

lrh2000:
This should use `EventFile::MAX_COUNTER_VALUE`. I believe they're the same thing, which more clearly indicates why we need this check. `ULLONG_MAX` is confusing and comes from nowhere.
https://github.com/asterinas/asterinas/blob/e61a873f634e37d8a519ad63045914931f0fcd63/kernel/src/syscall/eventfd.rs#L154-L160

Also please remember to add `CHECK()` for system calls in `FN_SETUP()`, e.g., `CHECK(close(ep_file))`.

No action is needed from you. I think I can add these fixes to my periodical "Fix some minor problems" commit.

PR #3370, opened 2026-06-11
Fix `getsockopt` and `setsockopt` returning EINVAL instead of EFAULT for NULL optlen

Marsman1996:
Removed the explicit null pointer check for optlen_addr in kernel/src/syscall/getsockopt.rs (lines 21-23). 
Previously, when optlen_addr == 0 (NULL), Asterinas returned EINVAL. Now the NULL address falls through to user_space.read_val(optlen_addr) which naturally produces EFAULT when attempting to read from address 0, matching Linux behavior.

### Describe the bug
When `getsockopt` is called with `optlen=NULL`, Asterinas returns EINVAL (errno=22) with message "optlen_addr is null pointer", while Linux returns EFAULT (errno=14). The root cause is an explicit null check in `kernel/src/syscall/getsockopt.rs:21-23` that returns EINVAL before the pointer dereference that would naturally produce EFAULT.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <errno.h>
#include <string.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <unistd.h>

int main(void) {
    int sock = socket(AF_INET, SOCK_STREAM, 0);
    if (sock < 0) {
        printf("SKIP: could not create socket: %s\n", strerror(errno));
        return 1;
    }

    int val = 0;
    errno = 0;
    int ret = getsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &val, NULL);
    if (ret == -1 && errno == EFAULT) {
        printf("VERDICT: LINUX behavior — NULL optlen returns EFAULT\n");
    } else if (ret == -1 && errno == EINVAL) {
        printf("VERDICT: ASTERINAS behavior — NULL optlen returns EINVAL\n");
    } else {
        printf("VERDICT: UNKNOWN — ret=%d, errno=%d (%s)\n",
               ret, errno, strerror(errno));
    }

    close(sock);
    return 0;
}
```

### Expected behavior
On Linux, `getsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &val, NULL)` returns -1 with `errno=EFAULT` (14).

### Log
- In Asterinas:
```
VERDICT: ASTERINAS behavior — NULL optlen returns EINVAL
```
- In Linux:
```
VERDICT: LINUX behavior — NULL optlen returns EFAULT
```

### DIFF
--- a/kernel/src/syscall/getsockopt.rs
+++ b/kernel/src/syscall/getsockopt.rs
@@ -18,9 +18,6 @@ pub fn sys_getsockopt(
     ctx: &Context,
 ) -> Result<SyscallReturn> {
     let level = CSocketOptionLevel::try_from(level).map_err(|_| Errno::EOPNOTSUPP)?;
-    if optlen_addr == 0 {
-        return_errno_with_message!(Errno::EINVAL, "optlen_addr is null pointer");
-    }
 
     let user_space = ctx.user_space();
     let optlen: u32 = user_space.read_val(optlen_addr)?;
--- a/kernel/src/syscall/setsockopt.rs
+++ b/kernel/src/syscall/setsockopt.rs
@@ -16,9 +16,6 @@ pub fn sys_setsockopt(
     ctx: &Context,
 ) -> Result<SyscallReturn> {
     let level = CSocketOptionLevel::try_from(level).map_err(|_| Errno::EOPNOTSUPP)?;
-    if optval == 0 {
-        return_errno_with_message!(Errno::EINVAL, "optval is null pointer");
-    }
 
     debug!(
         "level = {:?}, sockfd = {}, optname = {}, optval = {}",
--- a/test/initramfs/src/regression/network/sockoption.c
+++ b/test/initramfs/src/regression/network/sockoption.c
@@ -65,6 +65,15 @@ FN_TEST(invalid_socket_option)
 }
 END_TEST()
 
+FN_TEST(null_optlen)
+{
+	int val;
+	TEST_ERRNO(getsockopt(sk_connected, SOL_SOCKET, SO_KEEPALIVE, &val,
+			      NULL),
+		   EFAULT);
+}
+END_TEST()
+
 int refresh_connection()
 {
 	close(sk_connected);
--- a/test/initramfs/src/regression/network/sockoption.c
+++ b/test/initramfs/src/regression/network/sockoption.c
@@ -74,6 +74,18 @@ FN_TEST(null_optlen)
 }
 END_TEST()
 
+FN_TEST(null_optval)
+{
+	socklen_t len = sizeof(int);
+	TEST_ERRNO(setsockopt(sk_connected, SOL_SOCKET, SO_KEEPALIVE, NULL,
+			      sizeof(int)),
+		   EFAULT);
+	TEST_ERRNO(getsockopt(sk_connected, SOL_SOCKET, SO_KEEPALIVE, NULL,
+			      &len),
+		   EFAULT);
+}
+END_TEST()
+
 int refresh_connection()
 {
 	close(sk_connected);


Marsman1996:
And there is another inconsist problem, to fix this it seems I have to modify the code in `write_to_user` (it is now only called by `getsockopt`) instead of in `getsockopt` syscall impl. I'm not sure whether it is the best solution.

# `getsockopt` rejects small optlen instead of allowing partial reads

### Describe the bug
When `getsockopt` is called with an `optlen` value smaller than the option's actual size, Asterinas returns EINVAL ("max_len is too short"), while Linux performs a partial copy of `min(optlen, actual_size)` bytes and succeeds. This affects all int-sized socket options (4 bytes) when called with `optlen < 4`, including `optlen=0`. 

The root cause is in the `WriteToUser` implementations in `kernel/src/util/net/options/utils.rs:54-65` which unconditionally check `max_len < write_len` and return EINVAL.
https://github.com/asterinas/asterinas/blob/45e0026bff7d020ed311126319b24b069b1d5947/kernel/src/util/net/options/utils.rs#L53-L59

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <errno.h>
#include <string.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <unistd.h>

int main(void) {
    int sock = socket(AF_INET, SOCK_STREAM, 0);
    if (sock < 0) {
        printf("SKIP: could not create socket: %s\n", strerror(errno));
        return 1;
    }

    /* Set SO_KEEPALIVE to 1 first so we have known state */
    int val = 1;
    setsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &val, sizeof(val));

    /* Now try to get with optlen=2 (less than sizeof(int)=4) */
    int buf = 0;
    socklen_t len = 2;
    errno = 0;
    int ret = getsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &buf, &len);
    if (ret == 0) {
        printf("VERDICT: LINUX behavior — optlen=2 succeeds, partial read (len=%u, buf=0x%08x)\n",
               len, buf);
    } else if (ret == -1 && errno == EINVAL) {
        printf("VERDICT: ASTERINAS behavior — optlen=2 returns EINVAL\n");
    } else {
        printf("VERDICT: UNKNOWN — ret=%d, errno=%d (%s), len=%u\n",
               ret, errno, strerror(errno), len);
    }

    /* Also test optlen=0 */
    buf = 0xDEADBEEF;
    len = 0;
    errno = 0;
    ret = getsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &buf, &len);
    if (ret == 0 && len == 0) {
        printf("VERDICT: LINUX behavior (optlen=0) — succeeds with no data copied (buf unchanged: 0x%08x)\n", buf);
    } else if (ret == -1 && errno == EINVAL) {
        printf("VERDICT: ASTERINAS behavior (optlen=0) — returns EINVAL\n");
    } else {
        printf("VERDICT: UNKNOWN (optlen=0) — ret=%d, errno=%d (%s), len=%u\n",
               ret, errno, strerror(errno), len);
    }

    close(sock);
    return 0;
}
```

### Expected behavior
On Linux, `getsockopt` with a small `optlen` succeeds and performs a partial copy:
- `optlen=2`: copies 2 bytes of the 4-byte int value, sets `*optlen=2`, returns 0.
- `optlen=0`: copies no data, sets `*optlen=0`, returns 0.

### Log
- In Asterinas:
```
VERDICT: ASTERINAS behavior — optlen=2 returns EINVAL
VERDICT: ASTERINAS behavior (optlen=0) — returns EINVAL
```
- In Linux:
```
VERDICT: LINUX behavior — optlen=2 succeeds, partial read (len=2, buf=0x00000001)
VERDICT: LINUX behavior (optlen=0) — succeeds with no data copied (buf unchanged: 0xdeadbeef)
```

Marsman1996:
After thinking, maybe we shall 
```rust
if optlen_addr == 0 {
    return_errno_with_message!(Errno::EFAULT, "optlen_addr is null pointer");
}
```
to return error early?

lrh2000:
> to return error early?

There is no need to create a fast path for error cases. It is better to simplify.

Additionally, this fast path is incorrect. Depending on the system settings, creating a mapping at the zero address is possible (if you set `min_mmap_addr` to zero, which is possible in Linux, though Asterinas does not currently allow that).

---

What about `setsockopt`? There is a similar null pointer check which does not seem to have a Linux equivalent either.

Marsman1996:
> What about `setsockopt`? There is a similar null pointer check which does not seem to have a Linux equivalent either.

Yes, `setsockopt` has the similar problem.

lrh2000 (approved):

cqs21 (approved):

PR #3371, opened 2026-06-11
Fix `getcwd` not returning ENAMETOOLONG when path exceeds PATH_MAX

Marsman1996:
Added a `PATH_MAX` check on the resolved absolute path before checking the user buffer size.

https://elixir.bootlin.com/linux/v6.15/source/fs/d_path.c#L429-L438
```c
		DECLARE_BUFFER(b, page, PATH_MAX);

		prepend_char(&b, 0);
		if (unlikely(prepend_path(&pwd, &root, &b) > 0))
			prepend(&b, "(unreachable)", 13);
		rcu_read_unlock();

		len = PATH_MAX - b.len;
		if (unlikely(len > PATH_MAX))
			error = -ENAMETOOLONG;
```

### Describe the bug
Linux's `getcwd` syscall uses an internal buffer of `PATH_MAX` (4096) bytes to resolve the current working directory path. If the resolved absolute pathname exceeds this limit, the kernel returns `-ENAMETOOLONG`. Asterinas's `sys_getcwd` implementation (`kernel/src/syscall/getcwd.rs`) uses dynamic allocation via Rust's `CString` and `make_abs_path()` (which builds the path using `VecDeque<String>` and `format!`), so there is no PATH_MAX enforcement. This means paths longer than 4096 bytes may succeed on Asterinas where they would fail on Linux.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>
#include <sys/syscall.h>
#include <sys/stat.h>
#include <stdlib.h>

/*
 * Test: getcwd should return ENAMETOOLONG when the path exceeds PATH_MAX (4096).
 *
 * Linux uses an internal buffer of PATH_MAX bytes. If the resolved path
 * exceeds this, it returns -ENAMETOOLONG regardless of the user-provided size.
 * Asterinas uses dynamic allocation (CString) with no PATH_MAX limit.
 */
int main(void) {
    /* Create deeply nested directories to exceed PATH_MAX */
    char basedir[256];
    snprintf(basedir, sizeof(basedir), "/tmp/getcwd_long_%d", getpid());
    mkdir(basedir, 0755);
    chdir(basedir);

    /* Create directory names of 200 chars each. 25 levels * 201 = 5025 > 4096 */
    char dirname[256];
    memset(dirname, 'x', 200);
    dirname[200] = '\0';

    int depth = 0;
    for (int i = 0; i < 25; i++) {
        if (mkdir(dirname, 0755) != 0) {
            printf("INFO: mkdir failed at depth %d: %s\n", i, strerror(errno));
            break;
        }
        if (chdir(dirname) != 0) {
            printf("INFO: chdir failed at depth %d: %s\n", i, strerror(errno));
            break;
        }
        depth++;
    }

    printf("INFO: created %d levels of nesting\n", depth);

    /* Now try getcwd with a large buffer */
    char buf[8192];
    errno = 0;
    long ret = syscall(SYS_getcwd, buf, sizeof(buf));
    int err = errno;

    if (ret == -1 && err == ENAMETOOLONG) {
        printf("VERDICT: LINUX behavior — getcwd returns ENAMETOOLONG when path exceeds PATH_MAX (depth=%d)\n", depth);
    } else if (ret > 0) {
        printf("VERDICT: ASTERINAS behavior — getcwd succeeds with long path (len=%ld, depth=%d, no PATH_MAX limit)\n", ret, depth);
    } else {
        printf("VERDICT: UNKNOWN — ret=%ld, errno=%d (%s)\n", ret, err, strerror(err));
    }

    /* Cleanup */
    chdir("/tmp");
    /* Best-effort cleanup of deep dirs */
    char cmd[512];
    snprintf(cmd, sizeof(cmd), "rm -rf %s", basedir);
    system(cmd);

    return 0;
}
```

### Expected behavior
Linux returns `ENAMETOOLONG` (errno 36) when the resolved absolute pathname of the CWD exceeds PATH_MAX (4096 bytes). This happens regardless of the user-provided `size` argument because the kernel's internal resolution buffer is limited to PATH_MAX.

### Log
- In Asterinas:
```
INFO: created 25 levels of nesting
VERDICT: ASTERINAS behavior — getcwd succeeds with long path (len=5045, depth=25, no PATH_MAX limit)
rm: can't stat '/tmp/getcwd_long_10/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx': File name too long
```
- In Linux:
```
INFO: created 25 levels of nesting
VERDICT: LINUX behavior — getcwd returns ENAMETOOLONG when path exceeds PATH_MAX (depth=25)
```

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/pid/task/mountinfo.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/mountinfo.rs
@@ -30,9 +30,12 @@ pub(super) fn make_mount_point_path(
         "/".to_string()
     } else if let Some(parent) = parent {
         if let Some(mount_point_dentry) = mount.mountpoint() {
+            // TODO: Support paths longer than the path max length.
+            // Reference: <https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L239>
             path_resolver
                 .make_abs_path(&Path::new(parent.clone(), mount_point_dentry))
-                .into_string()
+                .map(|r| r.into_path_buf().to_string())
+                .unwrap_or_default()
         } else {
             "".to_string()
         }
--- a/kernel/src/fs/vfs/path/resolver.rs
+++ b/kernel/src/fs/vfs/path/resolver.rs
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use alloc::str;
-
 use ostd::task::Task;
 
 use super::{Mount, Path};
@@ -188,14 +186,24 @@ impl PathResolver {
     /// For pseudo paths, the returned string is simply the path's name. For other
     /// unreachable paths, the returned string starts with `/` but represents a partial
     /// path that could not reach the root.
-    pub fn make_abs_path(&self, path: &Path) -> AbsPathResult {
+    pub fn make_abs_path(&self, path: &Path) -> Result<AbsPathResult> {
+        let buf = PathBuffer::with_capacity(PATH_MAX);
+        self.make_abs_path_into(path, buf).ok_or_else(|| {
+            Error::with_message(Errno::ENAMETOOLONG, "the absolute path is too long")
+        })
+    }
+
+    fn make_abs_path_into(&self, path: &Path, mut buf: PathBuffer) -> Option<AbsPathResult> {
         // Handle the root path.
         if path == &self.root {
-            return AbsPathResult::Reachable("/".to_string());
+            buf.prepend_slash().then_some(())?;
+            return Some(AbsPathResult::Reachable(buf));
         }
         // Handle pseudo paths.
         if path.is_pseudo() {
-            return AbsPathResult::Unreachable(path.name());
+            let name = path.name();
+            buf.prepend_bytes(name.as_bytes()).then_some(())?;
+            return Some(AbsPathResult::Unreachable(buf));
         }
 
         // TODO: The paths reported via `/proc/<pid>/fd/<fd>` are currently
@@ -210,37 +218,32 @@ impl PathResolver {
         // the pseudo-path special-case above); the anonymous side is wired up
         // together with the `O_TMPFILE` open path in PR #3185.
 
-        let mut components = VecDeque::new();
-        components.push_front(self.resolve_name(path));
+        buf.prepend_component(&self.resolve_name(path))
+            .then_some(())?;
 
         let mut parent_path = self.resolve_parent(path);
         let mut reach_resolver_root = false;
 
         while let Some(parent_dir) = parent_path {
-            // Stop if we reach the resolver's root.
             if parent_dir == self.root {
                 reach_resolver_root = true;
                 break;
             }
 
             let parent_name = self.resolve_name(&parent_dir);
-            // Stop if we reach an absolute root.
             if parent_name == "/" {
                 break;
             }
 
-            components.push_front(parent_name);
+            buf.prepend_component(&parent_name).then_some(())?;
 
             parent_path = self.resolve_parent(&parent_dir);
         }
 
-        let path_name = alloc::format!("/{}", components.make_contiguous().join("/"));
-        debug_assert!(path_name.starts_with('/'));
-
         if reach_resolver_root {
-            AbsPathResult::Reachable(path_name)
+            Some(AbsPathResult::Reachable(buf))
         } else {
-            AbsPathResult::Unreachable(path_name)
+            Some(AbsPathResult::Unreachable(buf))
         }
     }
 
@@ -469,22 +472,90 @@ impl PathResolver {
     }
 }
 
+/// A buffer that builds an absolute path by prepending components from back to front.
+pub struct PathBuffer {
+    bytes: Vec<u8>,
+    start: usize,
+}
+
+impl PathBuffer {
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            bytes: vec![0u8; capacity],
+            start: capacity,
+        }
+    }
+
+    /// Returns `false` if there is not enough room.
+    pub fn prepend_bytes(&mut self, data: &[u8]) -> bool {
+        if data.len() > self.start {
+            return false;
+        }
+        self.start -= data.len();
+        self.bytes[self.start..self.start + data.len()].copy_from_slice(data);
+        true
+    }
+
+    fn prepend_component(&mut self, name: &str) -> bool {
+        let name_bytes = name.as_bytes();
+        if name_bytes.len() + 1 > self.start {
+            return false;
+        }
+        self.start -= name_bytes.len();
+        self.bytes[self.start..self.start + name_bytes.len()].copy_from_slice(name_bytes);
+        self.start -= 1;
+        self.bytes[self.start] = b'/';
+        true
+    }
+
+    fn prepend_slash(&mut self) -> bool {
+        self.prepend_bytes(b"/")
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes[self.start..]
+    }
+
+    pub fn as_str(&self) -> &str {
+        core::str::from_utf8(self.as_bytes()).expect("path is valid UTF-8")
+    }
+}
+
+impl Debug for PathBuffer {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_tuple("PathBuffer").field(&self.as_str()).finish()
+    }
+}
+
+impl core::fmt::Display for PathBuffer {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
 /// The result of resolving an absolute path name.
 ///
 /// If the path can be traced back to the root of the resolver, it is `Reachable`.
 /// Otherwise, it is `Unreachable`.
-#[derive(Clone, Debug)]
 pub enum AbsPathResult {
-    Reachable(String),
-    Unreachable(String),
+    Reachable(PathBuffer),
+    Unreachable(PathBuffer),
+}
+
+impl Debug for AbsPathResult {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            AbsPathResult::Reachable(buf) => f.debug_tuple("Reachable").field(buf).finish(),
+            AbsPathResult::Unreachable(buf) => f.debug_tuple("Unreachable").field(buf).finish(),
+        }
+    }
 }
 
 impl AbsPathResult {
-    /// Converts the `AbsPathResult` into a `String`.
-    pub fn into_string(self) -> String {
+    /// Unwraps the inner [`PathBuffer`] regardless of reachability.
+    pub fn into_path_buf(self) -> PathBuffer {
         match self {
-            AbsPathResult::Reachable(s) => s,
-            AbsPathResult::Unreachable(s) => s,
+            AbsPathResult::Reachable(buf) | AbsPathResult::Unreachable(buf) => buf,
         }
     }
 }
--- a/kernel/src/process/process/init_proc.rs
+++ b/kernel/src/process/process/init_proc.rs
@@ -8,7 +8,7 @@ use super::{Process, Session};
 use crate::{
     fs::{
         thread_info::ThreadFsInfo,
-        vfs::path::{FsPath, MountNamespace, Path},
+        vfs::path::{FsPath, MountNamespace},
     },
     prelude::*,
     process::{
@@ -95,7 +95,7 @@ fn create_init_process(
     let elf_path = fs.resolver().read().lookup(&fs_path)?;
 
     let pid = allocate_posix_tid();
-    let vmar = VmarHandle::new(ProcessVm::new(elf_path.clone()));
+    let vmar = VmarHandle::new(ProcessVm::new(elf_path));
     let resource_limits = new_resource_limits_for_init();
     let nice = Nice::default();
     let oom_score_adj = 0;
@@ -112,7 +112,8 @@ fn create_init_process(
         user_ns,
     );
 
-    let init_task = create_init_task(pid, &init_proc, fs, vmar, elf_path, argv, envp)?;
+    let thread_name = ThreadName::new_from_executable_path(executable_path);
+    let init_task = create_init_task(pid, &init_proc, fs, vmar, thread_name, argv, envp)?;
     init_proc.tasks().lock().insert(init_task).unwrap();
 
     Ok(init_proc)
@@ -138,29 +139,29 @@ fn create_init_task(
     process: &Arc<Process>,
     fs: ThreadFsInfo,
     vmar: VmarHandle,
-    elf_path: Path,
+    thread_name: ThreadName,
     argv: Vec<CString>,
     envp: Vec<CString>,
 ) -> Result<Arc<Task>> {
     let credentials = Credentials::new_root();
 
-    let (elf_load_info, elf_abs_path) = {
+    let elf_load_info = {
         let path_resolver = fs.resolver().read();
+        let elf_path = process
+            .lock_vmar()
+            .unwrap()
+            .process_vm()
+            .executable_file()
+            .clone();
 
-        let program_to_load =
-            ProgramToLoad::build_from_file(elf_path.clone(), &path_resolver, argv, envp)?;
+        let program_to_load = ProgramToLoad::build_from_file(elf_path, &path_resolver, argv, envp)?;
         let vmar = process.lock_vmar();
-        let elf_load_info = program_to_load.load_to_vmar(vmar.unwrap(), &path_resolver)?;
-        let elf_abs_path = path_resolver.make_abs_path(&elf_path).into_string();
-
-        (elf_load_info, elf_abs_path)
+        program_to_load.load_to_vmar(vmar.unwrap(), &path_resolver)?
     };
 
     let mut user_ctx = UserContext::default();
     user_ctx.set_instruction_pointer(elf_load_info.entry_point as _);
     user_ctx.set_stack_pointer(elf_load_info.user_stack_top as _);
-
-    let thread_name = ThreadName::new_from_executable_path(&elf_abs_path);
 
     let thread_builder =
         PosixThreadBuilder::new(tid, thread_name, Box::new(user_ctx), credentials, vmar)
--- a/kernel/src/syscall/getcwd.rs
+++ b/kernel/src/syscall/getcwd.rs
@@ -9,7 +9,7 @@ pub fn sys_getcwd(buf: Vaddr, len: usize, ctx: &Context) -> Result<SyscallReturn
     let abs_path = {
         let fs_ref = ctx.thread_local.borrow_fs();
         let path_resolver = fs_ref.resolver().read();
-        path_resolver.make_abs_path(path_resolver.cwd())
+        path_resolver.make_abs_path(path_resolver.cwd())?
     };
 
     debug!("getcwd: {:?}", abs_path);
@@ -20,17 +20,21 @@ pub fn sys_getcwd(buf: Vaddr, len: usize, ctx: &Context) -> Result<SyscallReturn
     // to conform to POSIX. Here follows Linux's behavior to keep compatibility.
     //
     // Reference: <https://man7.org/linux/man-pages/man3/getcwd.3.html>
-    let abs_path = match abs_path {
-        AbsPathResult::Reachable(s) => s,
-        AbsPathResult::Unreachable(s) => alloc::format!("(unreachable){}", s),
+    let path_buf = match abs_path {
+        AbsPathResult::Reachable(buf) => buf,
+        AbsPathResult::Unreachable(mut buf) => {
+            buf.prepend_bytes(b"(unreachable)");
+            buf
+        }
     };
 
-    let cwd = CString::new(abs_path)?;
-    let bytes = cwd.as_bytes_with_nul();
-    if bytes.len() > len {
+    let path_bytes = path_buf.as_bytes();
+    let total_len = path_bytes.len() + 1;
+    if total_len > len {
         return_errno_with_message!(Errno::ERANGE, "the CWD buffer is too small");
     }
-    ctx.user_space().write_bytes(buf, bytes)?;
+    ctx.user_space().write_bytes(buf, path_bytes)?;
+    ctx.user_space().write_bytes(buf + path_bytes.len(), &[0])?;
 
-    Ok(SyscallReturn::Return(bytes.len() as _))
+    Ok(SyscallReturn::Return(total_len as _))
 }
--- a/kernel/src/syscall/readlink.rs
+++ b/kernel/src/syscall/readlink.rs
@@ -43,7 +43,10 @@ pub fn sys_readlinkat(
 
         match path.inode().read_link()? {
             SymbolicLink::Plain(s) => s,
-            SymbolicLink::Path(path) => path_resolver.make_abs_path(&path).into_string(),
+            SymbolicLink::Path(path) => path_resolver
+                .make_abs_path(&path)?
+                .into_path_buf()
+                .to_string(),
         }
     };
 
--- a/kernel/src/vm/vmar/vm_mapping.rs
+++ b/kernel/src/vm/vmar/vm_mapping.rs
@@ -300,8 +300,12 @@ impl VmMapping {
                 }
             }
 
-            if let Some(path) = &self.path {
-                return Some(Cow::Owned(path_resolver.make_abs_path(path).into_string()));
+            // TODO: Support paths longer than the path max length.
+            // Reference: <https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L239>
+            if let Some(path) = &self.path
+                && let Ok(abs_path) = path_resolver.make_abs_path(path)
+            {
+                return Some(Cow::Owned(abs_path.into_path_buf().to_string()));
             }
 
             // Reference: <https://github.com/google/gvisor/blob/38123b53da96ff6983fcc103dfe2a9cc4e0d80c8/test/syscalls/linux/proc.cc#L1158-L1172>
--- a/test/initramfs/src/regression/fs/getcwd/getcwd.c
+++ b/test/initramfs/src/regression/fs/getcwd/getcwd.c
@@ -3,6 +3,9 @@
 #define _GNU_SOURCE
 
 #include <errno.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
 #include <unistd.h>
 
 #include "../../common/test.h"
@@ -13,3 +16,42 @@ FN_TEST(getcwd_small_buffer_returns_erange)
 	TEST_ERRNO(getcwd(small, 1), ERANGE);
 }
 END_TEST()
+
+FN_SETUP(getcwd_long_path)
+{
+	CHECK_WITH(mkdir("/tmp/getcwd_long", 0755),
+		   _ret == 0 || errno == EEXIST);
+	CHECK(chdir("/tmp/getcwd_long"));
+
+	char dirname[256];
+	memset(dirname, 'x', 200);
+	dirname[200] = '\0';
+
+	for (int i = 0; i < 25; i++) {
+		CHECK(mkdir(dirname, 0755));
+		CHECK(chdir(dirname));
+	}
+}
+END_SETUP()
+
+FN_TEST(getcwd_enametoolong_when_path_exceeds_path_max)
+{
+	char buf[8192];
+	TEST_ERRNO(syscall(SYS_getcwd, buf, sizeof(buf)), ENAMETOOLONG);
+}
+END_TEST()
+
+FN_SETUP(getcwd_long_path_cleanup)
+{
+	char dirname[256];
+	memset(dirname, 'x', 200);
+	dirname[200] = '\0';
+
+	for (int i = 0; i < 25; i++) {
+		CHECK(chdir(".."));
+		CHECK(rmdir(dirname));
+	}
+	CHECK(chdir("/"));
+	CHECK(rmdir("/tmp/getcwd_long"));
+}
+END_SETUP()
--- a/kernel/src/device/pty/master.rs
+++ b/kernel/src/device/pty/master.rs
@@ -187,8 +187,8 @@ impl PerOpenFileOps for PtyMaster {
 
                     let slave_name = {
                         let devpts_path = path_resolver
-                            .make_abs_path(super::DEV_PTS.get().unwrap())
-                            .into_string();
+                            .make_abs_path(super::DEV_PTS.get().unwrap())?
+                            .into_path_buf();
                         format!("{}/{}", devpts_path, self.slave.index())
                     };
 
--- a/kernel/src/process/execve.rs
+++ b/kernel/src/process/execve.rs
@@ -52,7 +52,7 @@ pub fn do_execve(
 
     debug!(
         "file path: {:?}, argv = {:?}, envp = {:?}",
-        path_resolver.make_abs_path(&elf_file).into_string(),
+        path_resolver.make_abs_path(&elf_file).ok(),
         argv,
         envp
     );


lrh2000:
`CHECK(... == 1)` does not make sense and will not work correctly. Check the impl. Why was this syntax invented?

There are nothing similar in the codebase (the following example is correct because it converts a NULL poniter to `-1`):
```
> grep -R "CHECK([^)]*==" test/**/*.c
test/initramfs/src/regression/ipc/pipe/short_rw.c:	CHECK(page == NULL ? -1 : 0);
```

Marsman1996:
> `CHECK(... == 1)` does not make sense and will not work correctly. Check the impl. Why was this syntax invented?
> 
> There are nothing similar in the codebase (the following example is correct because it converts a NULL poniter to `-1`):
> 
> ```
> > grep -R "CHECK([^)]*==" test/**/*.c
> test/initramfs/src/regression/ipc/pipe/short_rw.c:	CHECK(page == NULL ? -1 : 0);
> ```

I’m using a skill to generate tests, and this is the first time I’ve asked it to add `CHECK`. 
It seems to be interpreting `CHECK` as if it is `assert`...

lrh2000:
Could you please check how `make_abs_path` is used elsewhere and determine how it should behave if the path length exceeds `PATH_MAX`? I think we should fix `make_abs_path` instead.
```
> grep -R "[.]make_abs_path" kernel
kernel/src/device/pty/master.rs:                            .make_abs_path(super::DEV_PTS.get().unwrap())
kernel/src/fs/fs_impls/procfs/pid/task/mountinfo.rs:                .make_abs_path(&Path::new(parent.clone(), mount_point_dentry))
kernel/src/process/process/init_proc.rs:        let elf_abs_path = path_resolver.make_abs_path(&elf_path).into_string();
kernel/src/process/clone.rs:                    .make_abs_path(executable_path)
kernel/src/process/execve.rs:        path_resolver.make_abs_path(&elf_file).into_string(),
kernel/src/process/execve.rs:    let executable_path = path_resolver.make_abs_path(&elf_file).into_string();
kernel/src/syscall/getcwd.rs:        path_resolver.make_abs_path(path_resolver.cwd())
kernel/src/syscall/readlink.rs:            SymbolicLink::Path(path) => path_resolver.make_abs_path(&path).into_string(),
kernel/src/vm/vmar/vm_mapping.rs:                return Some(Cow::Owned(path_resolver.make_abs_path(path).into_string()));
```

`make_abs_path` and `getcwd` can be implemented more efficiently, e.g.,
```rust
let cwd = make_abs_path(PATH_MAX);
/* Should `cwd` be a `PathBuffer { data: Vec<u8>, start: usize }`,
 * which is similar to Linux's `prepend_buffer`? */
write_bytes(buf, cwd,as_bytes());
```

---

> I’m using a skill to generate tests, and this is the first time I’ve asked it to add `CHECK`.
> It seems to be interpreting `CHECK` as if it is `assert`...

Based on your previous PRs, I think you know how to use `CHECK()`. Therefore, you should take responsibility for catching AI's errors before submitting PRs. Anyway, tools are not to blame.

Marsman1996:
> Could you please check how `make_abs_path` is used elsewhere and determine how it should behave if the path length exceeds `PATH_MAX`? I think we should fix `make_abs_path` instead.

I'm adding check in `make_abs_path`. However, not all callers of `make_abs_path` should be subject to `PATH_MAX`.
For example, for the `/proc/pid/maps`, the linux uses dynamically-sized output:
https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L134-L139
```c
Eoverflow:
	m->op->stop(m, p);
	kvfree(m->buf);
	m->count = 0;
	m->buf = seq_buf_alloc(m->size <<= 1);
	return !m->buf ? -ENOMEM : -EAGAIN;
```

Maybe adding a `make_abs_path_unchecked`?

Marsman1996:
The test case:
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>
#include <fcntl.h>
#include <sys/syscall.h>
#include <sys/stat.h>
#include <sys/mman.h>
#include <stdlib.h>

/*
 * PoC: PATH_MAX enforcement across syscalls that use make_abs_path.
 *
 * Creates a deeply nested directory (path > 4096 bytes) and exercises:
 *   1. getcwd                     — should return ENAMETOOLONG
 *   2. readlink(/proc/self/fd/N)  — should return ENAMETOOLONG
 *   3. /proc/self/maps            — should show full path (no PATH_MAX limit)
 *
 * Note: /proc/self/cwd is NOT tested because Asterinas has not implemented
 * that symlink yet (returns ENOENT at directory lookup, unrelated to make_abs_path).
 *
 * Compile: gcc -o getcwd6 getcwd6.c
 * Run on both Linux and Asterinas to compare.
 */

#define DIRNAME_LEN 200
#define NEST_DEPTH  25  /* 25 * 201 = 5025 > 4096 */

static int saved_root_fd;

static int setup_deep_path(void) {
    char basedir[256];
    snprintf(basedir, sizeof(basedir), "/tmp/pathmax_test_%d", getpid());
    mkdir(basedir, 0755);

    saved_root_fd = open("/", O_RDONLY | O_DIRECTORY);
    if (saved_root_fd < 0) {
        perror("open /");
        return -1;
    }

    if (chdir(basedir) != 0) {
        perror("chdir basedir");
        return -1;
    }

    char dirname[256];
    memset(dirname, 'a', DIRNAME_LEN);
    dirname[DIRNAME_LEN] = '\0';

    int depth = 0;
    for (int i = 0; i < NEST_DEPTH; i++) {
        if (mkdir(dirname, 0755) != 0)
            break;
        if (chdir(dirname) != 0)
            break;
        depth++;
    }
    printf("[setup] nested %d levels, path ~%d bytes\n",
           depth, depth * (DIRNAME_LEN + 1) + (int)strlen(basedir));
    return depth;
}

static void cleanup(void) {
    fchdir(saved_root_fd);
    close(saved_root_fd);
    chdir("/tmp");
    char cmd[256];
    snprintf(cmd, sizeof(cmd), "rm -rf /tmp/pathmax_test_%d", getpid());
    system(cmd);
}

static void test_getcwd(void) {
    printf("\n[1] getcwd(buf, 8192)\n");
    char buf[8192];
    errno = 0;
    long ret = syscall(SYS_getcwd, buf, sizeof(buf));
    if (ret == -1)
        printf("    => errno=%d (%s)\n", errno, strerror(errno));
    else
        printf("    => OK, len=%ld\n", ret);
}

static void test_readlink_proc_fd(void) {
    printf("\n[2] readlink(/proc/self/fd/N) for file at deep path\n");
    int fd = open("testfile", O_CREAT | O_RDWR, 0644);
    if (fd < 0) {
        printf("    SKIP: open failed: %s\n", strerror(errno));
        return;
    }

    char link[64];
    snprintf(link, sizeof(link), "/proc/self/fd/%d", fd);
    char buf[8192];
    errno = 0;
    ssize_t ret = readlink(link, buf, sizeof(buf));
    if (ret == -1)
        printf("    => errno=%d (%s)\n", errno, strerror(errno));
    else
        printf("    => OK, len=%zd\n", ret);

    close(fd);
    unlink("testfile");
}

static void test_proc_maps(void) {
    printf("\n[3] /proc/self/maps for mmap'd file at deep path\n");
    int fd = open("mapfile", O_CREAT | O_RDWR, 0644);
    if (fd < 0) {
        printf("    SKIP: open failed: %s\n", strerror(errno));
        return;
    }
    ftruncate(fd, 4096);

    void *addr = mmap(NULL, 4096, PROT_READ, MAP_PRIVATE, fd, 0);
    if (addr == MAP_FAILED) {
        printf("    SKIP: mmap failed: %s\n", strerror(errno));
        close(fd);
        return;
    }

    FILE *f = fopen("/proc/self/maps", "r");
    if (!f) {
        printf("    SKIP: cannot open maps: %s\n", strerror(errno));
        munmap(addr, 4096);
        close(fd);
        return;
    }

    char line[8192];
    char target[64];
    snprintf(target, sizeof(target), "%lx-", (unsigned long)addr);

    while (fgets(line, sizeof(line), f)) {
        if (strncmp(line, target, strlen(target)) == 0) {
            line[strcspn(line, "\n")] = '\0';
            char *slash = strrchr(line, '/');
            if (slash)
                printf("    => path present, total line len=%zu\n", strlen(line));
            else
                printf("    => path ABSENT\n");
            break;
        }
    }

    fclose(f);
    munmap(addr, 4096);
    close(fd);
    unlink("mapfile");
}

int main(void) {
    printf("=== PATH_MAX enforcement PoC (PATH_MAX=4096 incl. nul) ===\n");

    int depth = setup_deep_path();
    if (depth < 20) {
        printf("ERROR: insufficient nesting (%d < 20)\n", depth);
        cleanup();
        return 1;
    }

    test_getcwd();
    test_readlink_proc_fd();
    test_proc_maps();

    printf("\n=== Expected Linux behavior ===\n");
    printf("[1] getcwd:              ENAMETOOLONG\n");
    printf("[2] readlink(proc/fd/N): ENAMETOOLONG\n");
    printf("[3] /proc/self/maps:     path present (NO PATH_MAX limit)\n");

    cleanup();
    return 0;
}

```

lrh2000:
> https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L134-L139

This is the file buffer. You need to figure out who calls `d_path`:
https://elixir.bootlin.com/linux/v6.15/source/fs/proc/task_mmu.c#L562
https://elixir.bootlin.com/linux/v6.15/source/fs/proc/task_mmu.c#L547

So:
```
   size_t name_buf_sz = min_t(size_t, PATH_MAX, karg.vma_name_size);
```

I don't think it is unbounded.

Marsman1996:
> who calls `d_path`:

`do_procmap_query` (`size_t name_buf_sz = min_t(size_t, PATH_MAX, karg.vma_name_size);`) is called by `procfs_procmap_ioctl`, which is accessed by `ioctl`.


For the `/proc/self/maps` in my testcase, `seq_path` calls the `d_path`
https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L461-L487
```c
int seq_path(struct seq_file *m, const struct path *path, const char *esc)
{
	char *buf;
	size_t size = seq_get_buf(m, &buf);
	int res = -1;

	if (size) {
		char *p = d_path(path, buf, size);
		if (!IS_ERR(p)) {
			char *end = mangle_path(buf, p, esc);
			if (end)
				res = end - buf;
		}
	}
	seq_commit(m, res);

	return res;
}
```

And the callsites are like
```
fgets() → read() syscall
    → seq_read()                              // proc_pid_maps_operations.read
      → seq_read_iter()                          // fs/seq_file.c:230
          → m->op->show(m, p)
            → show_map(m, v)                     // fs/proc/task_mmu.c:361
              → show_map_vma(m, vma)            // fs/proc/task_mmu.c:348-350
                → seq_path(m, path, "\n")       // fs/seq_file.c:470
                  → d_path(path, buf, size)
```

Eventually `seq_read_iter` will double the buffer and retry when `seq_path` overflows
 (`seq_commit(m, res)` marks the overflow):
https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L239-L248
```c
		if (!seq_has_overflowed(m)) // got it
			goto Fill;
		// need a bigger buffer
		m->op->stop(m, p);
		kvfree(m->buf);
		m->count = 0;
		m->buf = seq_buf_alloc(m->size <<= 1);
		if (!m->buf)
			goto Enomem;
		p = m->op->start(m, &m->index);
```

So the initial seq_get_buf size is PAGE_SIZE (4096), but it grows unboundedly on overflow.

lrh2000:
> For the `/proc/self/maps` in my testcase, `seq_path` calls the `d_path`

Oh sorry I misread the code. So we may really need a separate method, though I'm not sure about the names at this moment (`make_abs_path`/`make_abs_path_unlimited` or `make_abs_path`/`make_abs_path_with_max_len`...)

Marsman1996:
Add `make_abs_path_unlimited` and find in Linux `/proc/self/mountinfo` also calls `seq_read_iter` to make buffer grow.

kernel/src/process/process/init_proc.rs
```diff
@@ -151,7 +151,7 @@ fn create_init_task(
             ProgramToLoad::build_from_file(elf_path.clone(), &path_resolver, argv, envp)?;
         let vmar = process.lock_vmar();
         let elf_load_info = program_to_load.load_to_vmar(vmar.unwrap(), &path_resolver)?;
-        let elf_abs_path = path_resolver.make_abs_path(&elf_path).into_string();
+        let elf_abs_path = path_resolver.make_abs_path(&elf_path)?.into_string();
```
lrh2000: The code here is wrong. There is nothing to do with the absolute path.
https://github.com/asterinas/asterinas/blob/e4f8507720924b42e04dfa40b8b10e2c0bf11941/kernel/src/syscall/execve.rs#L87-L94
```
> cat /proc/1/comm
systemd
> ls /proc/1/exe -al
lrwxrwxrwx 1 root root 0 Jun 17 01:46 /proc/1/exe -> /usr/lib/systemd/systemd
```

Add a separate commit to fix this issue, before refactoring `make_abs_path`.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -188,14 +188,30 @@ impl PathResolver {
...
+    pub fn make_abs_path(&self, path: &Path) -> Result<AbsPathResult> {
+        self.make_abs_path_inner(path, true)
+    }
+
+    /// Like [`Self::make_abs_path`], but without the `PATH_MAX` enforcement.
+    ///
+    /// This mirrors Linux's `d_path` / `seq_path` behavior: contexts such as
+    /// `/proc/[pid]/maps` and `/proc/[pid]/mountinfo` render arbitrarily long
+    /// paths without truncation. The seq_file buffer grows unboundedly on
+    /// overflow rather than failing with `ENAMETOOLONG`.
```
lrh2000: I'm really annoyed by LLM-generated comments. Please do not mention Linux internal type and function names. Providing a reference link describing the behavior is sufficient.

 - **Bad:** `Matches Linux xxx method, which yyy`
 - **Good:** `Linux will yyy. Reference: <Linux source code link>`

This is also not the correct place to document what `mountinfo` and `maps` behave. Their behavior should be documented at the caller site.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -188,14 +188,30 @@ impl PathResolver {
     /// For pseudo paths, the returned string is simply the path's name. For other
     /// unreachable paths, the returned string starts with `/` but represents a partial
     /// path that could not reach the root.
-    pub fn make_abs_path(&self, path: &Path) -> AbsPathResult {
+    pub fn make_abs_path(&self, path: &Path) -> Result<AbsPathResult> {
+        self.make_abs_path_inner(path, true)
+    }
+
+    /// Like [`Self::make_abs_path`], but without the `PATH_MAX` enforcement.
```
lrh2000: Incorrect documentation style. Additionally, `make_abs_path` never mentions `PATH_MAX`. This documentation is not self-contained. Did you read it after the LLM generated it?

Good examples:
https://github.com/asterinas/asterinas/blob/e4f8507720924b42e04dfa40b8b10e2c0bf11941/ostd/src/mm/dma/dma_coherent.rs#L42
https://github.com/asterinas/asterinas/blob/e4f8507720924b42e04dfa40b8b10e2c0bf11941/ostd/src/mm/dma/dma_coherent.rs#L58-L62

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -237,10 +253,14 @@ impl PathResolver {
         let path_name = alloc::format!("/{}", components.make_contiguous().join("/"));
         debug_assert!(path_name.starts_with('/'));
 
+        if enforce_path_max && path_name.len() >= PATH_MAX {
```
lrh2000: Doing so has minimal benefits. I think we can have a path buffer to remove lots of heap allocations here, including:
 - `let mut components = VecDeque::new();`
 - `into_string()`
 - `CString::new()`

```rust
pub fn make_abs_path_unlimited() {
    let mut capability = PATH_MAX;

    loop {
        let pathbuf = PathBuffer::with_capability(capability);
        if let Some(result) = self.max_abs_path_inner(pathbuf)) {
            return result;
        }

        capability *= 2;
    }
}

struct PathBuffer {
    bytes: Vec<u8>,
    start: usize,
}

enum AbsPathResult {
     Reachable(PathBuffer),
     Unreachable(PathBuffer),
}

impl AbsPathResult {
    // This should be renamed since it no longer returns a string?
    fn into_string(self) -> PathBuffer {} 
}
```

One problem is what we should do if the `into_string` function finds that the buffer does not have enough capacity to prepend `(unreachable) `. However, there should be available solutions or workarounds (e.g. reversing the capability in advance or making the operation fallible).

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -300,8 +300,10 @@ impl VmMapping {
                 }
             }
 
-            if let Some(path) = &self.path {
-                return Some(Cow::Owned(path_resolver.make_abs_path(path).into_string()));
+            if let Some(path) = &self.path
+                && let Ok(abs_path) = path_resolver.make_abs_path_unlimited(path)
```
lrh2000: At the second though, I realize that Linux can use an unlimited buffer here because it has proper support of fallible allocation and memory accounting..

https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L38

We don't have anything like that at the moment. Another option is to drop the unlimited `make_abs_path` support. Instead, we could truncate the mountinfo and maps request. We could leave a TODO note explaining what we're missing and why we don't mimic the Linux unlimited behavior for now.

test/initramfs/src/regression/fs/getcwd/getcwd.c
```diff
@@ -13,3 +16,33 @@ FN_TEST(getcwd_small_buffer_returns_erange)
...
+FN_TEST(getcwd_enametoolong_when_path_exceeds_path_max)
+{
+	char buf[8192];
+	TEST_ERRNO(syscall(SYS_getcwd, buf, sizeof(buf)), ENAMETOOLONG);
+}
+END_TEST()
+
+FN_SETUP(getcwd_long_path_cleanup)
+{
+	CHECK(chdir("/tmp"));
```
lrh2000: I am not sure why this is a clean-up. The test may not be run from the `/tmp` directory. Why is it a `cleanup`? You may want to delete the newly created directories, which is a form of cleanup. `chdir` is just a side effect that does not matter.

lrh2000 (commented):
Thanks for the efforts!

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -210,37 +218,32 @@ impl PathResolver {
...
 
         while let Some(parent_dir) = parent_path {
-            // Stop if we reach the resolver's root.
             if parent_dir == self.root {
                 reach_resolver_root = true;
                 break;
             }
 
             let parent_name = self.resolve_name(&parent_dir);
-            // Stop if we reach an absolute root.
```
lrh2000, line 227: Could you please explain why you deleted this line?

The tools are not to blame. If the new changes are generated by an LLM, could you also explain why you could not identify this problem in your self-review? This PR isn't that large so this is not a thing that is hard to catch.

Check for other comments as well.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
+        self.start -= name_bytes.len();
+        self.bytes[self.start..self.start + name_bytes.len()].copy_from_slice(name_bytes);
+        self.start -= 1;
+        self.bytes[self.start] = b'/';
+        true
+    }
+
+    fn prepend_slash(&mut self) -> bool {
+        self.prepend_bytes(b"/")
+    }
```
lrh2000, line 513: No need to add these things. Just call `if !prepend(filename) || !prepend(b"/")`.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
 /// Otherwise, it is `Unreachable`.
-#[derive(Clone, Debug)]
 pub enum AbsPathResult {
-    Reachable(String),
-    Unreachable(String),
+    Reachable(PathBuffer),
+    Unreachable(PathBuffer),
+}
+
+impl Debug for AbsPathResult {
```
lrh2000, line 545: What's wrong with the original `#[derive(Debug)]`? No need to manually implement this.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
+/// A buffer that builds an absolute path by prepending components from back to front.
+pub struct PathBuffer {
+    bytes: Vec<u8>,
+    start: usize,
+}
+
+impl PathBuffer {
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            bytes: vec![0u8; capacity],
```
lrh2000, line 484: I missed the zero initialization part in my previous design. But when I see this, I think we should fix our design by using [`VecDeque::prepend`](https://doc.rust-lang.org/beta/std/collections/struct.VecDeque.html#method.prepend) instead.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
+    fn prepend_slash(&mut self) -> bool {
+        self.prepend_bytes(b"/")
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes[self.start..]
+    }
+
+    pub fn as_str(&self) -> &str {
+        core::str::from_utf8(self.as_bytes()).expect("path is valid UTF-8")
```
lrh2000, line 520: This will panic if we call `prepend_bytes` with non-UTF8 bytes. I think we should remove the `prepend_bytes` method. Having just a `prepend` method that accepts `&str` is sufficient.

```suggestion
        core::str::from_utf8(self.as_bytes()).expect("`prepend` only accepts valid UTF-8 characters")
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
+
+impl PathBuffer {
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            bytes: vec![0u8; capacity],
+            start: capacity,
+        }
+    }
+
+    /// Returns `false` if there is not enough room.
```
lrh2000, line 489: 1. We should also return `ENAMETOOLANG` if `prepend("unreachable")` fails in `sys_getcwd`.
 2. It would be better to return a truncated path than nothing when handling procfs files.

The current API design appears unable to fulfill these goals in a reasonable manner. See the following comment for an alternative design.

test/initramfs/src/regression/fs/getcwd/getcwd.c
```diff
@@ -13,3 +16,42 @@ FN_TEST(getcwd_small_buffer_returns_erange)
...
+{
+	char dirname[256];
+	memset(dirname, 'x', 200);
+	dirname[200] = '\0';
+
+	for (int i = 0; i < 25; i++) {
+		CHECK(chdir(".."));
+		CHECK(rmdir(dirname));
+	}
+	CHECK(chdir("/"));
```
lrh2000, line 54: Add a blank line before line 54.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
+            return false;
+        }
+        self.start -= data.len();
+        self.bytes[self.start..self.start + data.len()].copy_from_slice(data);
+        true
+    }
+
+    fn prepend_component(&mut self, name: &str) -> bool {
+        let name_bytes = name.as_bytes();
+        if name_bytes.len() + 1 > self.start {
```
lrh2000, line 501: Can we add some characters even though the buffer is almost full? So the following fact is more clear:
> It would be better to return a truncated path than nothing when handling procfs files.

Something like:
```rust
if self.is_overflowed {
    return false;
}

// Fast path: The whole string can be prepended.
let bytes = str.as_bytes();
if bytes.len() <= self.start {
    todo!("prepend `bytes`")
}

// Try to prepend as many characters as possible
for ch in str.rev() {
    let bytes = ch.as_bytes();
    if bytes.len() >= self.start {
        self.is_overflowed = true;
        return false;
    }

    todo!("prepend `ch`");
}

true
```

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -469,22 +472,90 @@ impl PathResolver {
...
 }
 
+/// A buffer that builds an absolute path by prepending components from back to front.
+pub struct PathBuffer {
+    bytes: Vec<u8>,
+    start: usize,
+}
+
+impl PathBuffer {
+    fn with_capacity(capacity: usize) -> Self {
```
lrh2000, line 482: In my original design, it is for `make_abs_path_unlimited`. Given that we do not want to have the unlimited version, let's do:
```rust
/// A buffer that builds an absolute path by prepending components from back to front.
///
/// This buffer can contain a maximum of [`PATH_MAX`] bytes. If more than this
/// amount of data is prepended, the buffer will be marked as overflowed, and
/// all of the additional characters will be ignored.
pub struct PathBuffer {
    bytes: VecDeque<u8>,
    is_overflowed: bool,
}

impl PathBuffer {
    fn new() {} // This just use `VecDeque::with_capability(PATH_MAX)`.

    // Other methods.

    /// Prepends a string.
    ///
    /// Characters that overflow the buffer's [`PATH_MAX`] capability will be
    /// lost. But the buffer will always contain a valid UTF-8 string because
    /// we discard entire UTF-8 characters.
    pub fn prepend(&mut self, str: &str);

    /// Returns the path as a `&[u8]`.
    ///
    /// This method will fail with [`Errno::ENAMETOOLANG`] if the buffer has
    /// been marked as overflowed.
    pub fn as_bytes(&self) -> Result<&str>;

    /// Returns the path as a `&str`.
    ///
    /// This method will fail with [`Errno::ENAMETOOLANG`] if the buffer has
    /// been marked as overflowed.
    pub fn as_str(&self) -> Result<&str>;

    /// Returns the path as a `&str`.
    ///
    /// This method will always succeed. However, any characters prepended to
    /// the buffer after it has been marked as overflowed are silently lost.
    pub fn as_str_trunc(&self) -> &str;
}

// `impl Display` should be removed. Use `as_str_trunc` or `as_str` explicitly.
```

kernel/src/fs/fs_impls/procfs/pid/task/mountinfo.rs
```diff
@@ -30,9 +30,12 @@ pub(super) fn make_mount_point_path(
         "/".to_string()
     } else if let Some(parent) = parent {
         if let Some(mount_point_dentry) = mount.mountpoint() {
+            // TODO: Support paths longer than the path max length.
+            // Reference: <https://elixir.bootlin.com/linux/v6.15/source/fs/seq_file.c#L239>
```
lrh2000, line 34:
```rust
            // TODO: Paths that are longer than `PATH_MAX` should not be
            // truncated. However, we need to implement memory accounting
            // before we can allocate a buffer of an arbitrary length.
```

Same for the `/proc/[pid]/maps` file.

kernel/src/syscall/readlink.rs
```diff
@@ -43,7 +43,10 @@ pub fn sys_readlinkat(
 
         match path.inode().read_link()? {
             SymbolicLink::Plain(s) => s,
-            SymbolicLink::Path(path) => path_resolver.make_abs_path(&path).into_string(),
+            SymbolicLink::Path(path) => path_resolver
+                .make_abs_path(&path)?
+                .into_path_buf()
+                .to_string(),
```
lrh2000, line 49: No need to call `to_string` here. Define an enum at the end of the file:
```rust
enum LinkPath {
    String(String),
    Buffer(PathBuffer),
}
```

Then you only need to do:
```rust
let bytes = match link_path {
    LinkPath::String(s) => s.as_bytes(),
    LinkPath::Buffer(buf) => buf.as_bytes()?,
};
```

kernel/src/process/process/init_proc.rs
```diff
@@ -138,30 +139,30 @@ fn create_init_task(
...
     let credentials = Credentials::new_root();
 
-    let (elf_load_info, elf_abs_path) = {
+    let elf_load_info = {
         let path_resolver = fs.resolver().read();
-
-        let program_to_load =
-            ProgramToLoad::build_from_file(elf_path.clone(), &path_resolver, argv, envp)?;
+        let elf_path = process
+            .lock_vmar()
```
lrh2000, line 151: There is no need to lock the VMAR twice.

```rust
    let elf_load_info = {
        let vmar = process.lock_vmar();
        let path_resolver = fs.resolver().read();

        let elf_path = vmar.unwrap().process_vm().executable_file().clone();
        let program_to_load = ProgramToLoad::build_from_file(elf_path, &path_resolver, argv, envp)?;
        program_to_load.load_to_vmar(vmar.unwrap(), &path_resolver)?
```

Regarding the lock order issue, I believe it's fine since we're constructing the initialization process. No one will contend with the VMAR lock, so we don't need to worry about lock order issues or holding the lock for too long.

kernel/src/syscall/getcwd.rs
```diff
@@ -20,17 +20,21 @@ pub fn sys_getcwd(buf: Vaddr, len: usize, ctx: &Context) -> Result<SyscallReturn
...
-    let bytes = cwd.as_bytes_with_nul();
-    if bytes.len() > len {
+    let path_bytes = path_buf.as_bytes();
+    let total_len = path_bytes.len() + 1;
+    if total_len > len {
         return_errno_with_message!(Errno::ERANGE, "the CWD buffer is too small");
     }
-    ctx.user_space().write_bytes(buf, bytes)?;
+    ctx.user_space().write_bytes(buf, path_bytes)?;
+    ctx.user_space().write_bytes(buf + path_bytes.len(), &[0])?;
```
lrh2000, line 37: Use `write_val(buf + path_bytes.len(), 0u8)` instead.

lrh2000 (commented):
Thanks for the updates. After checking the implementation, I found the previous design isn't quite satisfying. Please check the comments below (specifically https://github.com/asterinas/asterinas/pull/3371#discussion_r3440994492).

PR #3376, opened 2026-06-11
Fix `pidfd_send_signal` rejecting signal 0

Marsman1996:
Modified sys_pidfd_send_signal to accept signal 0 as a null signal, matching Linux behavior.
https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L1418-L1419
```C
	if (!ret && sig)
		ret = do_send_sig_info(sig, info, p, type);
```

### Describe the bug
Asterinas rejects signal number 0 in `pidfd_send_signal` with EINVAL, 
while Linux accepts it as a "null signal" that only performs permission and existence checks without actually delivering any signal. 

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <signal.h>
#include <unistd.h>
#include <sys/syscall.h>

#ifndef SYS_pidfd_send_signal
#define SYS_pidfd_send_signal 424
#endif

#ifndef SYS_pidfd_open
#define SYS_pidfd_open 434
#endif

int main(void) {
    int pidfd = syscall(SYS_pidfd_open, getpid(), 0);
    if (pidfd < 0) {
        perror("pidfd_open");
        return 1;
    }

    errno = 0;
    int ret = syscall(SYS_pidfd_send_signal, pidfd, 0, NULL, 0);
    if (ret == 0) {
        printf("VERDICT: LINUX behavior — signal 0 accepted (null signal for permission check)\n");
    } else {
        printf("VERDICT: ASTERINAS behavior — signal 0 rejected with errno=%d (%s)\n",
               errno, strerror(errno));
    }

    close(pidfd);
    return 0;
}
```

### Expected behavior
On Linux, `pidfd_send_signal(pidfd, 0, NULL, 0)` returns 0 (success) when the target process exists and the caller has permission to signal it. It returns -1 with ESRCH if the target is dead, or EPERM if lacking permission. No signal is delivered.

### Log
- In Asterinas:
```
VERDICT: ASTERINAS behavior — signal 0 rejected with errno=22 (Invalid argument)
```
- In Linux:
```
VERDICT: LINUX behavior — signal 0 accepted (null signal for permission check)
```

### DIFF
--- a/kernel/src/syscall/pidfd_send_signal.rs
+++ b/kernel/src/syscall/pidfd_send_signal.rs
@@ -28,44 +28,54 @@ pub fn sys_pidfd_send_signal(
     ctx: &Context,
 ) -> Result<SyscallReturn> {
     let flags = PidfdSendSignalFlags::try_from(flags)?;
-    let sig_num = SigNum::try_from(sig_num as u8)?;
+    let sig_num = if sig_num == 0 {
+        None
+    } else {
+        Some(SigNum::try_from(sig_num as u8)?)
+    };
     debug!(
         "pidfd={}, info_ptr={:#x}, flags={:?}",
         pidfd, info_ptr, flags
     );
 
-    let siginfo = read_siginfo_from_user(info_ptr, sig_num, ctx)?;
-    let signal = RawSignal::new(siginfo);
-
     let target = get_target_from_pidfd(pidfd, flags, ctx)?;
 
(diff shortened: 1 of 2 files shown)


StevenJiang1110 (approved):

cqs21 (approved):

PR #3362, opened 2026-06-10
Fix `alarm` remaining time rounding to match Linux behavior

Marsman1996:
Update the rounding logic to match Linux's alarm_setitimer() implementation in 
https://elixir.bootlin.com/linux/v6.15/source/kernel/time/itimer.c#L311-L319
```C
	/*
	 * We can't return 0 if we have an alarm pending ...  And we'd
	 * better return too much than too little anyway
	 */
	if ((!it_old.it_value.tv_sec && it_old.it_value.tv_nsec) ||
	      it_old.it_value.tv_nsec >= (NSEC_PER_SEC / 2))
		it_old.it_value.tv_sec++;

	return it_old.it_value.tv_sec;
```

### Describe the bug
When `alarm()` returns the remaining time of a previously set alarm, Asterinas rounds up to the next whole second (e.g., 1.5s remaining -> returns 2), 
while Linux truncates to the lower second (1.5s remaining -> returns 1). 
This can cause programs that use the return value of `alarm()` for timing calculations to behave differently.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <unistd.h>
#include <signal.h>

static void handler(int sig) { (void)sig; }

int main(void) {
    struct sigaction sa = { .sa_handler = handler, .sa_flags = 0 };
    sigemptyset(&sa.sa_mask);
    sigaction(SIGALRM, &sa, NULL);

    alarm(3);
    usleep(1500000); /* 1.5 seconds */
    unsigned int rem = alarm(0);

    printf("alarm(3), wait 1.5s, alarm(0) -> remaining = %u\n", rem);
    return 0;
}
```

### Expected behavior
`alarm(0)` should return `1` (1.5 seconds remaining, truncated to 1).

### Log
- In Asterinas:
```
alarm(3), wait 1.5s, alarm(0) -> remaining = 2
```
- In Linux:
```
alarm(3), wait 1.5s, alarm(0) -> remaining = 1
```

PR #3377, opened 2026-06-11
Refactor sigset size validation

Marsman1996:
Changed the size validation from `sigset_size != 8` to `sigset_size > 8`, matching Linux behavior.
https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L3390-L3395
```c
SYSCALL_DEFINE2(rt_sigpending, sigset_t __user *, uset, size_t, sigsetsize)
{
	sigset_t set;

	if (sigsetsize > sizeof(*uset))
		return -EINVAL;
```

### Describe the bug
Asterinas's `rt_sigpending` implementation requires `sigsetsize` to be exactly 8, returning `EINVAL` for any other value. Linux only rejects `sigsetsize > sizeof(sigset_t)` (i.e., > 8 on x86-64) and allows smaller sizes, copying only `sigsetsize` bytes to userspace.

Root cause: `kernel/src/syscall/rt_sigpending.rs:17-18` uses `sigset_size != 8` instead of `sigset_size > 8`.

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <signal.h>
#include <unistd.h>
#include <sys/syscall.h>
#include <stdint.h>

int main(void) {
    sigset_t blocked;
    unsigned char buf[8];

    /* Block SIGUSR1 and raise it so there's something pending */
    sigemptyset(&blocked);
    sigaddset(&blocked, SIGUSR1);
    sigprocmask(SIG_BLOCK, &blocked, NULL);
    raise(SIGUSR1);

    /* Test 1: sigsetsize=4 (partial copy) */
    memset(buf, 0xAA, sizeof(buf));
    errno = 0;
    long ret = syscall(SYS_rt_sigpending, buf, 4);
    if (ret == 0) {
        unsigned int val = *(unsigned int *)buf;
        printf("VERDICT: LINUX behavior — sigsetsize=4 returns 0, partial copy val=0x%x\n", val);
    } else {
        printf("VERDICT: ASTERINAS behavior — sigsetsize=4 returns %ld errno=%s\n",
               ret, strerror(errno));
    }

    /* Test 2: sigsetsize=0 */
    errno = 0;
    ret = syscall(SYS_rt_sigpending, buf, 0);
    if (ret == 0) {
        printf("VERDICT: LINUX behavior — sigsetsize=0 returns 0\n");
    } else {
        printf("VERDICT: ASTERINAS behavior — sigsetsize=0 returns %ld errno=%s\n",
               ret, strerror(errno));
    }

    /* Clean up */
    signal(SIGUSR1, SIG_IGN);
    sigprocmask(SIG_UNBLOCK, &blocked, NULL);
    return 0;
}
```

### Expected behavior
On Linux, `rt_sigpending` with `sigsetsize` less than 8 (but >= 0) should succeed (return 0) and copy only `sigsetsize` bytes of the pending signal set to the userspace buffer. The kernel checks `sigsetsize > sizeof(sigset_t)` and only rejects values larger than 8.

### Log
- In Asterinas:
```
VERDICT: ASTERINAS behavior — sigsetsize=4 returns -1 errno=Invalid argument
VERDICT: ASTERINAS behavior — sigsetsize=0 returns -1 errno=Invalid argument
```
- In Linux:
```
VERDICT: LINUX behavior — sigsetsize=4 returns 0, partial copy val=0x200
VERDICT: LINUX behavior — sigsetsize=0 returns 0
```

### DIFF
--- a/kernel/src/process/signal/sig_mask.rs
+++ b/kernel/src/process/signal/sig_mask.rs
@@ -28,7 +28,7 @@ pub type SigMask = SigSet;
 /// Because that all the signal numbers are in the range of 1 to 64, casting
 /// a signal set from `u64` to `SigSet` will always succeed.
 #[repr(C)]
-#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Pod)]
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub struct SigSet {
     bits: u64,
 }
@@ -126,6 +126,13 @@ impl ops::Not for SigSet {
     }
 }
 
+// This is to allow hexadecimally formatting a `SigSet` when debug printing it.
+impl LowerHex for SigSet {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        LowerHex::fmt(&self.bits, f)
+    }
+}
+
 impl SigSet {
     pub fn new_empty() -> Self {
         SigSet { bits: 0 }
@@ -155,13 +162,6 @@ impl SigSet {
     pub fn intersects(&self, other: impl Into<Self>) -> bool {
         let other = other.into();
         self.bits & other.bits != 0
-    }
-}
-
-// This is to allow hexadecimally formatting a `SigSet` when debug printing it.
-impl LowerHex for SigSet {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        LowerHex::fmt(&self.bits, f) // delegate to u64's implementation
     }
 }
 
--- a/kernel/src/syscall/rt_sigaction.rs
+++ b/kernel/src/syscall/rt_sigaction.rs
@@ -2,7 +2,7 @@
 
 use ostd::mm::VmIo;
 
-use super::SyscallReturn;
+use super::{SyscallReturn, rt_sigprocmask::RequireFullSize};
 use crate::{
     prelude::*,
     process::{
@@ -22,7 +22,7 @@ pub fn sys_rt_sigaction(
     sig_num: u8,
     sig_action_addr: Vaddr,
     old_sig_action_addr: Vaddr,
-    sigset_size: u64,
+    sigset_size: usize,
     ctx: &Context,
 ) -> Result<SyscallReturn> {
     let sig_num = SigNum::try_from(sig_num)?;
@@ -34,9 +34,8 @@ pub fn sys_rt_sigaction(
         sigset_size
     );
 
-    if sigset_size != 8 {
-        return_errno_with_message!(Errno::EINVAL, "sigset size is not equal to 8");
-    }
+    // Validate the size of the signal set
+    RequireFullSize::new(sigset_size)?;
 
     let sig_dispositions = ctx.process.sig_dispositions().lock();
     let mut sig_dispositions = sig_dispositions.lock();
--- a/kernel/src/syscall/rt_sigpending.rs
+++ b/kernel/src/syscall/rt_sigpending.rs
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use ostd::mm::VmIo;
-
-use super::SyscallReturn;
+use super::{
+    SyscallReturn,
+    rt_sigprocmask::{AllowTruncSize, UserSigSetPtr},
+};
 use crate::{prelude::*, process::signal::HandlePendingSignal};
 
 pub fn sys_rt_sigpending(
@@ -14,21 +15,20 @@ pub fn sys_rt_sigpending(
         "u_set_ptr = 0x{:x},  sigset_size = {}",
         u_set_ptr, sigset_size
     );
-    if sigset_size != 8 {
-        return_errno_with_message!(Errno::EINVAL, "sigset size is not equal to 8")
-    }
-    do_rt_sigpending(u_set_ptr, ctx)?;
+    let size_policy = AllowTruncSize::new(sigset_size)?;
+    do_rt_sigpending(u_set_ptr, size_policy, ctx)?;
     Ok(SyscallReturn::Return(0))
 }
 
-fn do_rt_sigpending(set_ptr: Vaddr, ctx: &Context) -> Result<()> {
+fn do_rt_sigpending(set_ptr: Vaddr, size_policy: AllowTruncSize, ctx: &Context) -> Result<()> {
     let combined_signals = {
         let sig_mask_value = ctx.posix_thread.sig_mask();
         let sig_pending_value = ctx.pending_signals();
         sig_mask_value & sig_pending_value
     };
 
-    ctx.user_space()
-        .write_val(set_ptr, &u64::from(combined_signals))?;
+    let user_space = ctx.user_space();
+    let set_ptr = UserSigSetPtr::new(&user_space, set_ptr, size_policy);
+    set_ptr.write_val(&combined_signals)?;
     Ok(())
 }
--- a/kernel/src/syscall/rt_sigtimedwait.rs
+++ b/kernel/src/syscall/rt_sigtimedwait.rs
@@ -4,13 +4,16 @@ use core::time::Duration;
 
 use ostd::{mm::VmIo, sync::Waiter};
 
-use super::SyscallReturn;
+use super::{
+    SyscallReturn,
+    rt_sigprocmask::{RequireFullSize, UserSigSetPtr},
+};
 use crate::{
     prelude::*,
     process::signal::{
         HandlePendingSignal,
         constants::{SIGKILL, SIGSTOP},
-        sig_mask::{SigMask, SigSet},
+        sig_mask::SigMask,
         signals::Signal,
         with_sigmask_changed,
     },
@@ -29,14 +32,14 @@ pub fn sys_rt_sigtimedwait(
         set_ptr, info_ptr, timeout_ptr, sigset_size
     );
 
-    // Validate sigset size
-    if sigset_size != size_of::<SigMask>() {
-        return_errno_with_message!(Errno::EINVAL, "invalid sigset size");
-    }
+    // Validate the size of the signal set.
+    let size_policy = RequireFullSize::new(sigset_size)?;
 
+    let user_space = ctx.user_space();
+    let sigmask_ptr = UserSigSetPtr::new(&user_space, set_ptr, size_policy);
     // Read the signal set
     let mask = {
-        let mut set: SigSet = ctx.user_space().read_val(set_ptr)?;
+        let mut set = sigmask_ptr.read_val()?;
 
         // Remove SIGKILL and SIGSTOP as they cannot be waited for
         set -= SIGKILL;
--- a/kernel/src/syscall/signalfd.rs
+++ b/kernel/src/syscall/signalfd.rs
@@ -12,9 +12,11 @@ use core::{
 };
 
 use bitflags::bitflags;
-use ostd::mm::VmIo;
 
-use super::SyscallReturn;
+use super::{
+    SyscallReturn,
+    rt_sigprocmask::{RequireFullSize, UserSigSetPtr},
+};
 use crate::{
     events::IoEvents,
     fs::{
@@ -60,11 +62,11 @@ pub fn sys_signalfd4(
         raw_fd, mask_ptr, sizemask, flags
     );
 
-    if sizemask != size_of::<SigMask>() {
-        return Err(Error::with_message(Errno::EINVAL, "invalid mask size"));
-    }
+    let size_policy = RequireFullSize::new(sizemask)?;
 
-    let mut mask = ctx.user_space().read_val::<SigMask>(mask_ptr)?;
+    let user_space = ctx.user_space();
+    let sigmask_ptr = UserSigSetPtr::new(&user_space, mask_ptr, size_policy);
+    let mut mask = sigmask_ptr.read_val()?;
     mask -= SIGKILL;
     mask -= SIGSTOP;
 
--- a/kernel/src/syscall/signalfd.rs
+++ b/kernel/src/syscall/signalfd.rs
@@ -9,9 +9,11 @@
 use core::{fmt::Display, sync::atomic::Ordering};
 
 use bitflags::bitflags;
-use ostd::mm::VmIo;
 
-use super::SyscallReturn;
+use super::{
+    SyscallReturn,
+    rt_sigprocmask::{RequireFullSize, UserSigSetPtr},
+};
 use crate::{
     events::IoEvents,
     fs::{
@@ -56,11 +58,11 @@ pub fn sys_signalfd4(
         raw_fd, mask_ptr, sizemask, flags
     );
 
-    if sizemask != size_of::<SigMask>() {
-        return_errno_with_message!(Errno::EINVAL, "invalid mask size");
-    }
+    let size_policy = RequireFullSize::new(sizemask)?;
 
-    let mut mask = ctx.user_space().read_val::<SigMask>(mask_ptr)?;
+    let user_space = ctx.user_space();
+    let sigmask_ptr = UserSigSetPtr::new(&user_space, mask_ptr, size_policy);
+    let mut mask = sigmask_ptr.read_val()?;
     mask -= SIGKILL;
     mask -= SIGSTOP;
 
--- a/test/initramfs/src/regression/process/signal/rt_sigpending.c
+++ b/test/initramfs/src/regression/process/signal/rt_sigpending.c
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
+
+#include <signal.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#include "../../common/test.h"
+
+static long rt_sigpending(void *set, size_t sigsetsize)
+{
+	return syscall(SYS_rt_sigpending, set, sigsetsize);
+}
+
+sigset_t blocked;
+
+FN_SETUP(block_sigusr1)
+{
+	sigemptyset(&blocked);
+	sigaddset(&blocked, SIGUSR1);
+	CHECK(sigprocmask(SIG_BLOCK, &blocked, NULL));
+	CHECK(raise(SIGUSR1));
+}
+END_SETUP()
+
+FN_TEST(sigsetsize_larger_than_eight_rejected)
+{
+	unsigned char buf[16];
+
+	TEST_ERRNO(rt_sigpending(buf, 9), EINVAL);
+	TEST_ERRNO(rt_sigpending(buf, 16), EINVAL);
+}
+END_TEST()
+
+FN_TEST(sigsetsize_equal_to_eight)
+{
+	unsigned long val = 0;
+
+	TEST_SUCC(rt_sigpending(&val, 8));
+	TEST_RES(rt_sigpending(&val, 8), val & (1UL << (SIGUSR1 - 1)));
+}
+END_TEST()
+
+FN_TEST(sigsetsize_smaller_than_eight)
+{
+	unsigned char buf[8];
+
+	// sigsetsize = 4: should write the lower 4 bytes containing SIGUSR1
+	memset(buf, 0, sizeof(buf));
+	TEST_RES(rt_sigpending(buf, 4),
+		 buf[((SIGUSR1 - 1) / 8)] & (1U << ((SIGUSR1 - 1) % 8)));
+
+	// sigsetsize = 0: no-op, should succeed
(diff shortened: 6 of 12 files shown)


kernel/src/syscall/rt_sigpending.rs
```diff
@@ -14,21 +14,22 @@ pub fn sys_rt_sigpending(
...
     let combined_signals = {
         let sig_mask_value = ctx.posix_thread.sig_mask();
         let sig_pending_value = ctx.pending_signals();
         sig_mask_value & sig_pending_value
     };
 
+    let sigset_bytes = u64::from(combined_signals).to_ne_bytes();
     ctx.user_space()
-        .write_val(set_ptr, &u64::from(combined_signals))?;
+        .write_bytes(set_ptr, &sigset_bytes[..sigset_size])?;
```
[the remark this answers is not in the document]
Marsman1996: Linux copy it by `copy_to_user(uset, &set, sigsetsize)`

test/initramfs/src/regression/process/signal/rt_sigpending.c
```diff
@@ -0,0 +1,63 @@
...
+
+FN_TEST(sigsetsize_smaller_than_eight)
+{
+	unsigned char buf[8];
+
+	memset(buf, 0, sizeof(buf));
+	TEST_SUCC(rt_sigpending(buf, 4));
+
+	unsigned int partial = *(unsigned int *)buf;
+	TEST_RES(rt_sigpending(buf, 4), partial & (1U << (SIGUSR1 - 1)));
```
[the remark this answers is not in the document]
Marsman1996: 🤔

lrh2000:
Please confirm the following:
```
> grep -R "if sig\(mask\|set\)_size" kernel
kernel/src/syscall/ppoll.rs:        if sigmask_size != size_of::<SigMask>() {
kernel/src/syscall/rt_sigaction.rs:    if sigset_size != 8 {
kernel/src/syscall/rt_sigpending.rs:    if sigset_size != 8 {
kernel/src/syscall/rt_sigprocmask.rs:    if sigset_size != 8 {
kernel/src/syscall/rt_sigsuspend.rs:    if sigmask_size != size_of::<SigMask>() {
kernel/src/syscall/rt_sigtimedwait.rs:    if sigset_size != size_of::<SigMask>() {
kernel/src/syscall/epoll.rs:        if sigmask_size != size_of::<SigMask>() {
```

lrh2000:
@Marsman1996 Thank you for your efforts to resolve the many inconsistencies in our system call implementations. After reviewing these PRs, I have a few suggestions for you.

In general, I think you should be more thoughtful when submitting PRs like this one. Many system calls are never carefully audited, resulting in many behaviors that are inconsistent with Linux. It's not surprising that a simple AI agent can identify and generate fixes for many problems.

But when you check the AI's fix, I hope you:

 1. Understand the system call's implementation and check for obvious problems that haven't been fixed. This will prevent you from ending up with multiple PRs that fix multiple issues within a single system call, some of which may touch the same code. For example,
    - https://github.com/asterinas/asterinas/pull/3371 and https://github.com/asterinas/asterinas/pull/3356
    - https://github.com/asterinas/asterinas/pull/3376 and https://github.com/asterinas/asterinas/pull/3360

(Note that small PRs are easier to review than large ones. So, I'm not suggesting that you blindly combine multiple PRs. There is a trade-off. At least, I don't think submitting multiple PRs that touch the same code is really helpful.)

 2. Do some work to check if there are similar problems in our codebase. This includes:
    - Whether the system call layer is the most suitable layer for the fix, e.g.,
       - https://github.com/asterinas/asterinas/pull/3371#issuecomment-4687771436
       - https://github.com/asterinas/asterinas/pull/3355#issuecomment-4668781676
    - Whether other system calls that accept similar arguments have the same problem, e.g.,
       - https://github.com/asterinas/asterinas/pull/3370#issuecomment-4677706004
       - https://github.com/asterinas/asterinas/pull/3377#issuecomment-4689929773
  
I understand that you may not be familiar with the codebase, which may make you feel unsure about your decisions. That's fine. If you're not sure, you can submit the simplest fix and wait for the reviewers to decide. But you should learn to think about these things.

Hopefully, your knowledge will grow in this process, leading you to make better decisions in the future. Otherwise, flooding PRs without thinking will make the review bandwidth unmanageable.

Marsman1996:
> Please confirm the following

`rt_sigpending` is the only syscall that uses `>` (allows smaller sigsetsize). All others use `!=`.

For `rt_sigaction`, the Linux code is https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L4634:
```c
if (sigsetsize != sizeof(sigset_t))
    return -EINVAL;
```

For `rt_sigprocmask`, the Linux code is https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L3325:
```c
if (sigsetsize != sizeof(sigset_t))
    return -EINVAL;
```

For `rt_sigsuspend`, the Linux code is https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L4853:
```c
if (sigsetsize != sizeof(sigset_t))
    return -EINVAL;
```

For `rt_sigtimedwait`, the Linux code is https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L3815:
```c
if (sigsetsize != sizeof(sigset_t))
    return -EINVAL;
```

For `ppoll` and `epoll_pwait`, they both go through `set_user_sigmask` at https://elixir.bootlin.com/linux/v6.15/source/kernel/signal.c#L3278:
```c
if (sigsetsize != sizeof(sigset_t))
    return -EINVAL;
```

Marsman1996:
> Hopefully, your knowledge will grow in this process, leading you to make better decisions in the future. Otherwise, flooding PRs without thinking will make the review bandwidth unmanageable.

Thank you for the detailed feedback and guidance.
I'm opening PRs directly from individual differential fuzzing reports. 
And I will group and merge related reports in the future and open one PR to reduce the review burden.

lrh2000:
> `rt_sigpending` is the only syscall that uses `>` (allows smaller sigsetsize). All others use `!=`.

Thanks for the investigation. It turns out that `rt_sigpending` is the real exception. 

Since we have audited the code, this should be a good opportunity to unify the code and make the fact explicit. A potential design is (`SigSet` should not be `#[derive(Pod)]` if this works, all reads/writes should go through `SigSet{Full,Trunc}Size::{read,write}_val`):
```rust
impl SigSet {
    /// Checks whether the user-provided size matches the full size of [`SigSet`].
    ///
    /// This is used for syscalls whose Linux ABI requires the size to be exactly
    /// `size_of::<SigSet>()`.
    pub fn check_full_size(size: usize) -> Result<SigSetFullSize>;

    /// Checks whether the user-provided size is not larger than [`SigSet`].
    ///
    /// This is used for syscalls such as `rt_sigpending`, whose Linux ABI
    /// rejects only sizes greater than `size_of::<SigSet>()` and copies at most
    /// the requested number of bytes.
    pub fn check_trunc_size(size: usize) -> Result<SigSetTruncSize>;
}

/// A checked user-space size for a full [`SigSet`].
///
/// See [`SigSet::check_full_size`].
pub struct SigSetFullSize {
    _private: (),
}

impl SigSetFullSize {
    /// Reads a full [`SigSet`] from user space.
    pub fn read_val(
        self,
        user_space: &CurrentUserSpace<'_>,
        addr: Vaddr,
    ) -> Result<SigSet>;

    /// Writes a full [`SigSet`] to user space.
    pub fn write_val(
        self,
        user_space: &CurrentUserSpace<'_>,
        addr: Vaddr,
        sigset: &SigSet,
    ) -> Result<()>;
}

/// A checked user-space size for a possibly truncated [`SigSet`].
///
/// See [`SigSet::check_trunc_size`].
pub struct SigSetTruncSize {
    size: usize,
}

impl SigSetTruncSize {
    /// Writes a possibly truncated [`SigSet`] to user space.
    pub fn write_val(
        self,
        user_space: &CurrentUserSpace<'_>,
        addr: Vaddr,
        sigset: &SigSet,
    ) -> Result<()>;
}

// Maybe also define `SigMaskFullSize` and `SigMaskTruncSize` as type aliases...?
// Check the existing code and be consistent with them.
```

kernel/src/process/signal/sig_mask.rs
```diff
@@ -156,8 +157,88 @@ impl SigSet {
...
+///
+/// See [`SigSet::check_full_size`].
+#[derive(Clone, Copy, Debug)]
+pub struct SigSetFullSize {
+    _private: (),
+}
+
+impl SigSetFullSize {
+    /// Reads a full [`SigSet`] from user space.
+    pub(crate) fn read_val(self, user_space: &CurrentUserSpace<'_>, addr: Vaddr) -> Result<SigSet> {
```
lrh2000: No `pub(crate)` in the kernel crate. https://github.com/asterinas/asterinas/pull/3362#issuecomment-4668509490

kernel/src/process/signal/sig_mask.rs
```diff
@@ -156,8 +157,88 @@ impl SigSet {
...
+        let mut bytes = [0u8; size_of::<SigSet>()];
+        user_space.read_bytes(addr, &mut bytes)?;
+        Ok(SigSet::from(u64::from_ne_bytes(bytes)))
```
lrh2000: No need to do so. Use `read_val::<u64>` instead.

kernel/src/process/signal/sig_mask.rs
```diff
@@ -156,8 +157,88 @@ impl SigSet {
...
+
+    /// Writes a full [`SigSet`] to user space.
+    pub(crate) fn write_val(
+        self,
+        user_space: &CurrentUserSpace<'_>,
+        addr: Vaddr,
+        sigset: &SigSet,
+    ) -> Result<()> {
+        let bytes = u64::from(*sigset).to_ne_bytes();
+        user_space.write_bytes(addr, &bytes)?;
```
lrh2000: Same here.

kernel/src/syscall/rt_sigaction.rs
```diff
@@ -34,9 +34,7 @@ pub fn sys_rt_sigaction(
         sigset_size
     );
 
-    if sigset_size != 8 {
-        return_errno_with_message!(Errno::EINVAL, "sigset size is not equal to 8");
-    }
+    SigMask::check_full_size(sigset_size as usize)?;
```
lrh2000: https://elixir.bootlin.com/linux/v7.1/source/kernel/signal.c#L4631-L4634

The system call argument should use `usize` instead.

---

Also we can add a comment since it is the only case where we ignore the `check_*` results.
```rust
    // This is a check for the signal mask contained in `SigAction`.
    let _ = SigMask::check_full_size(sigset_size)?;
```

kernel/src/process/signal/sig_mask.rs
```diff
@@ -156,8 +157,88 @@ impl SigSet {
...
+        sigset: &SigSet,
+    ) -> Result<()> {
+        let bytes = u64::from(*sigset).to_ne_bytes();
+        user_space.write_bytes(addr, &bytes[..self.size])?;
+        Ok(())
+    }
+}
+
+/// A checked user-space size for a full [`SigMask`].
+pub type SigMaskFullSize = SigSetFullSize;
```
lrh2000: According to other type aliases defined in this module. This should be defined just above the corresponding aliased type, i.e., 
```rust
/// A checked user-space size for a full [`SigMask`].
pub type SigMaskFullSize = SigSetFullSize;

#[derive(Clone, Copy, Debug)]
pub struct SigSetFullSize {
```

kernel/src/syscall/rt_sigtimedwait.rs
```diff
@@ -29,14 +29,11 @@ pub fn sys_rt_sigtimedwait(
         set_ptr, info_ptr, timeout_ptr, sigset_size
     );
 
-    // Validate sigset size
-    if sigset_size != size_of::<SigMask>() {
-        return_errno_with_message!(Errno::EINVAL, "invalid sigset size");
-    }
+    let checked_size = SigMask::check_full_size(sigset_size)?;
```
lrh2000: There is a comment before this PR. Keep it or:
```suggestion
    // Validate the size of the signal set
    let checked_size = SigMask::check_full_size(sigset_size)?;
```

kernel/src/syscall/signalfd.rs
```diff
@@ -60,11 +59,8 @@ pub fn sys_signalfd4(
         raw_fd, mask_ptr, sizemask, flags
     );
 
-    if sizemask != size_of::<SigMask>() {
-        return Err(Error::with_message(Errno::EINVAL, "invalid mask size"));
-    }
-
-    let mut mask = ctx.user_space().read_val::<SigMask>(mask_ptr)?;
+    let checked_size = SigMask::check_full_size(sizemask)?;
+    let mut mask = checked_size.read_val(&ctx.user_space(), mask_ptr)?;
```
lrh2000: Keep the original blank line. You seem to preserve it for most other cases, so I don't know why it is removed here.

kernel/src/process/signal/sig_mask.rs
```diff
@@ -156,8 +157,88 @@ impl SigSet {
...
         self.bits & other.bits != 0
     }
+
+    /// Checks whether the user-provided size matches the full size of [`SigSet`].
+    ///
+    /// This is used for syscalls whose Linux ABI requires the size to be exactly
+    /// `size_of::<SigSet>()`.
+    pub fn check_full_size(size: usize) -> Result<SigSetFullSize> {
+        if size != size_of::<Self>() {
+            return_errno_with_message!(Errno::EINVAL, "invalid sigset size");
```
lrh2000:
```suggestion
            return_errno_with_message!(Errno::EINVAL, "the sigset size is invalid");

File: kernel/src/process/signal/sig_mask.rs [hunk 11]
```diff
@@ -156,8 +157,88 @@ impl SigSet {
...
+    }
+
+    /// Checks whether the user-provided size is not larger than [`SigSet`].
+    ///
+    /// This is used for syscalls such as `rt_sigpending`, whose Linux ABI
+    /// rejects only sizes greater than `size_of::<SigSet>()` and copies at most
+    /// the requested number of bytes.
+    pub fn check_trunc_size(size: usize) -> Result<SigSetTruncSize> {
+        if size > size_of::<Self>() {
+            return_errno_with_message!(Errno::EINVAL, "sigset size is too large");
```
lrh2000:
```suggestion
            return_errno_with_message!(Errno::EINVAL, "the sigset size is too large");
```

kernel/src/process/signal/sig_mask.rs
lrh2000, line 166: Move all the `*Size` definitions below this trait impl.

Maybe even better, this `impl` seems to be in an odd location. We can move it near other trait impls:
```rust
impl ops::Not for SigSet {
    type Output = Self;

    fn not(self) -> Self {
        SigSet { bits: !self.bits }
    }
}

// This is to allow hexadecimally formatting a `SigSet` when debug printing it.
impl LowerHex for SigSet {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        LowerHex::fmt(&self.bits, f) // delegate to u64's implementation
    }
}
```

lrh2000 (commented):
Thanks for the quick implementation!

kernel/src/process/signal/sig_mask.rs
Marsman1996: Without `(crate)` there will be a warning
```
warning: type `context::CurrentUserSpace<'_>` is more private than the item `SigSetFullSize::read_val`
   --> /root/asterinas/kernel/src/process/signal/sig_mask.rs:195:5
    |
195 |     pub fn read_val(self, user_space: &CurrentUserSpace<'_>, addr: Vaddr) -> Result<SigSet> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method `SigSetFullSize::read_val` is reachable at visibility `pub`
```

lrh2000: Oh really. Then ignore this. `CurrentUserSpace` has to be `pub(crate)` due to some compiler bugs.... Thanks for reminding me that fact...

lrh2000 (approved):

tatetian:
Hey guys, thanks for improving the correctness of handling all signal mask pointers! I like the idea of persisting the two strategies (full size vs trunc size) into the type system so that we won't make sloppy mistakes in the syscall layer.

But I think it still has some rooms for improvements due to the concerns below:

First, adding `check_xxx_size` methods to `SigSet` violates the **single-responsibility principle**. Whether the user-space provides a full or truncated signal set and whether this is acceptable for a particular syscall is a separate concern from the kernel representation of a signal mask, which is what `SigSet` is all about. Thus, seeing these two methods on `SigSet` would be a surprise.

Second, as the two `check_xxx_size` methods is about a syscall layer concern, such utility functions (types) should appear in the syscall layer, not in `signal/sig_mask.rs`.

Third, the interface design causes confusing usage:

```rust
let checked_size = SigMask::check_full_size(sigmask_with_size.size)?;
let sigmask = checked_size.read_val(&user_space, sigmask_with_size.addr)?;
```

The line `checked_size.read_val` reads strange: read value from a checked size? One can read a value from a file, a VM space, or a reader, but not a checked value.

For the reasons above, I propose introducing an alternative design (draft code):

```rust
pub(super) struct UserSigSetPtr<'a, P> {
    user_space: &'a CurrentUserSpace<'a>,
    addr: Vaddr,
    size_policy: P,
}

pub(super) struct RequireFullSize;

pub(super) struct AllowTruncSize {
    real_size: usize,
}

impl<P> UserSigSetPtr<P> {
    pub(super) fn new(
        user_space: &'a CurrentUserSpace<'a>,
        addr: Vaddr,
        size_policy: P
    ) -> Self {
        ...
    }
}

impl UserSigSetPtr<RequireFullSize> {
    pub(super) fn read_val(&self) -> Result<SigSet> { ... }

    pub(super) fn write_val(&self, new_val: &SigSet) -> Result<()> { ... }
}

impl UserSigSetPtr<AllowTruncSize> {
    pub(super) fn read_val(&self) -> Result<SigSet> { ... }
}
```

And this new abstraction is added to `syscall/rt_sigprocmask.rs` or another syscall file regarding signal handling.

Marsman1996:
Sorry for busy with my thesis recently and didn’t get to this PR in time.

> For the reasons above, I propose introducing an alternative design (draft code):

Refactored the code accordingly.

> And this new abstraction is added to `syscall/rt_sigprocmask.rs` or another syscall file regarding signal handling.

The `aster-code-review` skill suggests

```markdown
Move `UserSigSetPtr`, `RequireFullSize`, and `AllowTruncSize` into a dedicated module such as `kernel/src/syscall/sigset_ptr.rs` (or a `sigset` submodule)
```

Currently they are all in `syscall/rt_sigprocmask.rs`, which is a little weird, but creating a new `syscall/sigset_ptr.rs` is more  weird to me.

tatetian:
@Marsman1996 Could you rebase this PR to the latest main?

Marsman1996:
> @Marsman1996 Could you rebase this PR to the latest main?

Rebase to the latest commit

kernel/core/src/syscall/rt_sigtimedwait.rs
```diff
@@ -29,14 +32,14 @@ pub(super) fn sys_rt_sigtimedwait(
...
 
-    // Validate sigset size
-    if sigset_size != size_of::<SigMask>() {
-        return_errno_with_message!(Errno::EINVAL, "invalid sigset size");
-    }
+    // Validate the size of the signal set.
+    let size_policy = RequireFullSize::new(sigset_size)?;
 
+    let user_space = ctx.user_space();
+    let sigmask_ptr = UserSigSetPtr::new(&user_space, set_ptr, size_policy);
```
tatetian: The name `RequireFullSize` would be confusing without the context of creating new signal set. So I suggest refactoring all code to the following pattern:

```rust
let sigmask_ptr = {
    let size_policy = RequireFullSize::new(sigset_size)?;
    UserSigSetPtr::new(&user_space, set_ptr, size_policy)
};
```

With this pattern, the intention of `RequireFullSize::new` would be apparent. You don't even need to write comment such as `// Validate the size of the signal set.`.

tatetian (commented):
Thanks for the revision. I have only one styling concern.

kernel/core/src/syscall/rt_sigtimedwait.rs
Marsman1996: All updated except the `rt_sigaction`, since it does not create a pointer and I keep the comment for this as well.

PR #3380, opened 2026-06-11
Fix inotify_poll test race on SMP

cqs21:
Fix a race condition in the `inotify_poll` regression test that caused sporadic failures on SMP systems (observed with `INTEL_TDX=1 SMP=4`).

The child process's `IN_MODIFY` event could arrive after the parent's `read()`, leaving it in the queue when the subsequent `poll(0)` check expects an empty state. Moving `waitpid` before `read` ensures all events are queued before draining.

See CI failure: https://github.com/asterinas/asterinas/actions/runs/27310628212/job/80679644490

### DIFF
--- a/test/initramfs/src/regression/fs/inotify/inotify_poll.c
+++ b/test/initramfs/src/regression/fs/inotify/inotify_poll.c
@@ -99,8 +99,6 @@ int main(void)
 		return 3;
 	}
 
-	char buf[4096]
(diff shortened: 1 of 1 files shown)


lrh2000 (approved):
Ideally, we would want to die if `waitpid` fails. However, I assume this is also okay, given that we do not die on cleanup operations.

The fix is a good catch. LGTM. Thanks for your contribution!

PR #3382, opened 2026-06-11
Fix `gettimeofday` not returning EFAULT for invalid timezone pointer

Marsman1996:
Add the `timezone_t` struct and handle  the `tz` parameter in the syscall handler.

https://elixir.bootlin.com/linux/v6.15/source/kernel/time/time.c#L140-L155
```C
SYSCALL_DEFINE2(gettimeofday, struct __kernel_old_timeval __user *, tv,
		struct timezone __user *, tz)
{
	if (likely(tv != NULL)) {
		struct timespec64 ts;

		ktime_get_real_ts64(&ts);
		if (put_user(ts.tv_sec, &tv->tv_sec) ||
		    put_user(ts.tv_nsec / 1000, &tv->tv_usec))
			return -EFAULT;
	}
	if (unlikely(tz != NULL)) {
		if (copy_to_user(tz, &sys_tz, sizeof(sys_tz)))
			return -EFAULT;
	}
	return 0;
}
```

The original comment says
```rust
// The use of the timezone structure is obsolete.
// Glibc sets the timezone_addr argument to NULL, so just ignore it.
```
But I cannot find the corresponding glibc code logic, 
and for glibc-2.43, it set `tz` to zero not set it to NULL
https://elixir.bootlin.com/glibc/glibc-2.43/source/sysdeps/unix/sysv/linux/gettimeofday.c#L31-L37
```C
static int
__gettimeofday_syscall (struct timeval *restrict tv, void *restrict tz)
{
  if (__glibc_unlikely (tz != NULL))
    memset (tz, 0, sizeof *tz);
  return INLINE_SYSCALL_CALL (gettimeofday, tv, tz);
}
```

### Describe the bug
When `gettimeofday` is called with a valid (or NULL) `tv` pointer but an invalid `tz` pointer (e.g., `0xDEAD0000`), 
Linux correctly returns -1 with `errno=EFAULT`. 
Asterinas does not return EFAULT because the `tz` parameter is never received by the syscall handler -- the dispatch at `kernel/src/syscall/arch/x86.rs:271` only passes one argument (`args[..1]`), so the invalid pointer is silently ignored and the syscall returns 0. 

### To Reproduce
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <errno.h>
#include <string.h>
#include <sys/time.h>
#include <sys/syscall.h>
#include <unistd.h>

/*
 * DIFF PoC: Test whether EFAULT is returned for bad tz pointer.
 *
 * Linux: gettimeofday(NULL, 0xDEAD0000) returns -1 with EFAULT.
 * Asterinas: tz parameter is ignored, so bad tz never causes EFAULT.
 *
 * Detection: Pass an invalid tz pointer and check if we get EFAULT.
 */
int main(void) {
    int ret;

    errno = 0;
    ret = syscall(SYS_gettimeofday, NULL, (void*)0xDEAD0000);

    if (ret == -1 && errno == EFAULT) {
        printf("VERDICT: LINUX behavior — bad tz pointer returns EFAULT (ret=%d, errno=EFAULT)\n", ret);
    } else if (ret == 0) {
        printf("VERDICT: ASTERINAS behavior — bad tz pointer is ignored, returns 0\n");
    } else {
        printf("VERDICT: UNKNOWN — ret=%d, errno=%s\n", ret, strerror(errno));
    }

    return 0;
}
```

### Expected behavior
On Linux, passing an invalid pointer as the `tz` argument to `gettimeofday` causes the kernel to return -1 with `errno` set to `EFAULT`, indicating that the pointer is outside the accessible address space.

### Log
- In Asterinas:
```
Expected: VERDICT: ASTERINAS behavior — bad tz pointer is ignored, returns 0
```
- In Linux:
```
VERDICT: LINUX behavior — bad tz pointer returns EFAULT (ret=-1, errno=EFAULT)
```


kernel/src/syscall/gettimeofday.rs
```diff
@@ -5,25 +5,27 @@ use ostd::mm::VmIo;
...
 
-    let time_val = {
-        let now = SystemTime::now();
-        let time_duration = now.duration_since(&SystemTime::UNIX_EPOCH)?;
-        timeval_t::from(time_duration)
-    };
-    ctx.user_space().write_val(timeval_addr, &time_val)?;
+    if timezone_addr != 0 {
+        let tz = timezone_t::default();
+        ctx.user_space().write_val(timezone_addr, &tz)?;
```
cchanging, line 28: I agree with Copilot that we can add a brief comment here explaining why we write the default value—for example, that this parameter is obsolete on Linux, and that we also haven’t implemented `settimeofday` yet.

test/initramfs/src/regression/time/gettimeofday/gettimeofday.c
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#define _GNU_SOURCE
```
cchanging, line 3: "regression/misc/gettimeofday.c" or "regression/time/gettimeofday.c"  may be a better place?

cchanging (commented):
Thanks for fixing this inconsistent behavior!

test/initramfs/src/regression/time/gettimeofday/gettimeofday.c
Marsman1996: maybe `regression/time/gettimeofday.c`

test/initramfs/src/regression/Makefile
```diff
@@ -21,6 +21,7 @@ SUBDIRS := \
 	network \
 	process \
 	scripts \
+	time \
 	security \
```
lrh2000, line 24: Keep it sorted.

test/initramfs/src/regression/time/run_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -e
+
+./gettimeofday/gettimeofday
```
lrh2000, line 9: Move `process/itimer` to this directory?

test/initramfs/src/regression/time/gettimeofday/gettimeofday.c
```diff
@@ -0,0 +1,35 @@
...
+}
+END_TEST()
+
+FN_TEST(gettimeofday_both_null_success)
+{
+	TEST_SUCC(syscall(SYS_gettimeofday, NULL, NULL));
+}
+END_TEST()
+
+FN_TEST(gettimeofday_valid_tv)
```
lrh2000, line 30: Add `gettimeofday_valid_tz` for consistency purposes?

test/initramfs/src/regression/time/run_test.sh
Marsman1996: Then how about clock_nanosleep and alarm?

lrh2000: Good catch! I think they also seem to be more related to time.

I didn't realize this because I just use `find . -name '*time*'`.

Marsman1996:
Ok, we never run the alarm test before, it was a dead test...
https://github.com/asterinas/asterinas/blob/f0958799ad220ef2f926e15b0bad07372d16a103/test/initramfs/src/regression/process/run_test.sh#L1-L83

kernel/src/time/mod.rs
```diff
@@ -171,3 +171,10 @@ pub struct itimerspec_t {
     pub it_interval: timespec_t,
     pub it_value: timespec_t,
 }
+
+#[repr(C)]
```
lrh2000, line 176:
```suggestion
/// This struct is corresponding to the `timezone` struct in Linux.
#[repr(C)]
```

Be consistent with:
```rust
/// This struct is corresponding to the `itimerval` struct in Linux.
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, Pod)]
pub struct itimerval_t {
    pub it_interval: timeval_t,
    pub it_value: timeval_t,
}

/// This struct is corresponding to the `itimerspec` struct in Linux.
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, Pod)]
pub struct itimerspec_t {
    pub it_interval: timespec_t,
    pub it_value: timespec_t,
}
```

lrh2000 (approved):

lrh2000:
> Ok, we never run the alarm test before, it was a dead test...

We have some dead tests (for various reasons). This is not the only one. However, fixing them is not within the scope of this PR...

cqs21 (approved):

cchanging (approved):

PR #3385, opened 2026-06-12
Replace `libflate` with `zune-inflate` for initramfs decompression

cqs21:
Decompression time comparison (average of 5 runs, in seconds):

### `make run_kernel` (initramfs.cpio.gz: 12M)

| Configuration | libflate | zune-inflate | Speedup |
|---|---|---|---|
| dev | 2.215s | 1.762s | 20.4% faster |
| release-lto | 0.133s | 0.094s | 29.3% faster |

### `make run_kernel ENABLE_BENCHMARK_TEST=true` (initramfs.cpio.gz: 32M)

| Configuration | libflate | zune-inflate | Speedup |
|---|---|---|---|
| dev | 5.941s | 4.675s | 21.3% faster |
| release-lto | 0.379s | 0.245s | 35.4% faster |

### `make run_kernel ENABLE_CONFORMANCE_TEST=true CONFORMANCE_TEST_SUITE=kselftest` (initramfs.cpio.gz: 121M)

| Configuration | libflate | zune-inflate | Speedup |
|---|---|---|---|
| dev | 22.958s | 22.881s | 0.3% faster |
| release-lto | 1.632s | 0.975s | 40.3% faster |

In release-lto builds, `zune-inflate` achieves **29-40% faster** decompression across all tested image sizes.

Dev builds gain ~20% for smaller images, but the advantage nearly disappears at 121M — likely because at that size the bottleneck shifts to memory allocation/copying rather than pure inflate computation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,7 +221,6 @@ dependencies = [
  "jhash",
  "keyable-arc",
  "lending-iterator",
- "libflate",
  "logo-ascii-art",
  "loongArch64",
  "lru",
@@ -245,6 +244,7 @@ dependencies = [
  "xarray",
  "xmas-elf",
  "zerocopy",
+ "zune-inflate",
 ]
 
 [[package]]
@@ -1145,15 +1145,14 @@ name = "linux-bzimage-setup"
 version = "0.18.0"
 dependencies = [
  "cfg-if",
- "libflate",
  "linux-boot-params",
  "log",
- "no_std_io2",
  "tdx-guest",
  "uart_16550",
  "uefi",
  "uefi-raw 0.14.0",
  "xmas-elf",
+ "zune-inflate",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


lrh2000:
Why not remove `libflate`? It is still used in `linux-bzimage` I think.

In addition, could you keep the crate names sorted in `Cargo.toml`?

cqs21:
> Why not remove `libflate`? It is still used in `linux-bzimage` I think.

Thanks for the reminder! However, `zune-inflate` currently only supports decompression yet (though it's mentioned as a future feature in its docs). So we can't replace `libflate` with `zune-inflate` in `linux-bzimage`.

lrh2000:
You still need to sort `kernel/Cargo.toml` I think...

---

> However, `zune-inflate` currently only supports decompression yet (though it's mentioned as a future feature in its docs). So we can't replace `libflate` with `zune-inflate` in `linux-bzimage`.

We can try to replace `bzimage/setup`. `bzimage/builder` is only used at build time so the impact is quite minimal.

But it is indeed a bit unfortunate that we cannot elimate `libflate` from `/Cargo.toml`. Thanks for the information.

tatetian:
Good job. This is definitely an improvement.

> likely because at that size the bottleneck shifts to memory allocation/copying rather than pure inflate computation.

What is the largest reasonable initramfs size in real-world scenarios? And under such scenarios, what is the overall gap of Asterinas vs Linux? Is the gap already bridged with this PR or the gap still exists and is caused by the inefficient memory allocation. 

```
pub fn decode_payload(payload: &[u8]) -> Vec<u8> {
```

The interface returning a `Vec` mandates repeatedly growing the capacity of the underlying buffer and doing memory copies. Does this design cause a major overhead?

tatetian (approved):

PR #3386, opened 2026-06-12
Fix futimesat with O_PATH fd

WaterWhisperer:
## Summary

Fix `futimesat(fd, NULL, times)` to return `EBADF` when `fd` was opened with `O_PATH`.

## Changes

- Return `EBADF` for the legacy NULL-path `futimesat` case on `O_PATH` file descriptors.
- Add regression for `futimesat(fd, NULL, times)` with an `O_PATH` fd.
- Unblock the gVisor `FutimesatTest.OnNullPathWithOPath` case.

## Testing

- `./utimes_test --gtest_filter=FutimesatTest.OnNullPathWithOPath`

  ```text
  [  PASSED  ] 1 test.
  ```

  ```text
  79 of 79 test cases passed.
  ```

  ```text
  All regression tests passed.
  ```

### DIFF
--- a/kernel/src/syscall/utimens.rs
+++ b/kernel/src/syscall/utimens.rs
@@ -8,7 +8,10 @@ use super::{SyscallReturn, constants::MAX_FILENAME_LEN};
 use crate::{
     fs,
     fs::{
-        file::file_table::RawFileDesc,
+        file::{
+            StatusFlags,
+            file_table::{RawFileDesc, get_file_fast},
+        },
         vfs::path::{AT_FDCWD, EmptyPathStr, FsPath, Path},
     },
     prelude::*,
@@ -171,20 +174,8 @@ fn do_utimes(
         Some(pathname.to_string_lossy().into_owned())
     };
 
-    let path = {
-        let fs_path = if let Some(pathname) = pathname.as_ref() {
-            FsPath::from_fd_at(dirfd, pathname, EmptyPathStr::AllowIfFlag(flags.bits()))?
-        } else {
-            // Matches Linux `do_utimes_fd`: no flags are accepted when pathname is NULL.
-            if !flags.is_empty() {
-                return_errno_with_message!(
-                    Errno::EINVAL,
-                    "flags must be zero when pathname is NULL"
-                );
-            }
-            FsPath::from_fd(dirfd)?
-        };
-
+    let path = if let Some(pathname) = pathname.as_ref() {
+        let fs_path = FsPath::from_fd_at(dirfd, pathname, EmptyPathStr::AllowIfFlag(flags.bits()))?;
         let fs_ref = ctx.thread_local.borrow_fs();
         let path_resolver = fs_ref.resolver().read();
         if flags.contains(UtimensFlags::AT_SYMLINK_NOFOLLOW) {
@@ -192,6 +183,21 @@ fn do_utimes(
         } else {
             path_resolver.lookup(&fs_path)?
         }
+    } else {
+        // Linux will reject non-zero flags and `O_PATH` files if the pathname is null.
+        // Reference: <https://elixir.bootlin.com/linux/v7.1/source/fs/utimes.c#L108>
+        if !flags.is_empty() {
+            return_errno_with_message!(
+                Errno::EINVAL,
+                "flags must be zero when the pathname is NULL"
+            );
+        }
+        let mut file_table = ctx.thread_local.borrow_file_table_mut();
+        let file = get_file_fast!(&mut file_table, dirfd.try_into()?);
+        if file.status_flags().contains(StatusFlags::O_PATH) {
+            return_errno_with_message!(Errno::EBADF, "the file is opened as a path");
+        }
+        file.path().clone()
     };
 
     vfs_utimes(&path, times)
--- a/test/initramfs/src/regression/fs/utimensat/utimensat.c
+++ b/test/initramfs/src/regression/fs/utimensat/utimensat.c
@@ -70,6 +70,21 @@ FN_TEST(legacy_null_pathname_with_nonzero_flag_returns_einval)
 }
 END_TEST()
 
+FN_TEST(utimensat_null_pathname_with_o_path_returns_ebadf)
+{
+	int opath_fd = TEST_SUCC(open(TEST_FILE, O_PATH));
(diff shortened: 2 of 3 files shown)


lrh2000:
This is wrong because the check is racy (the file is immediately dropped after the check, so the path may not really correspond to the checked file).

Also I don't think we need `ifdef SYS_futimesat` per our previous discussion. If the code was generated by an LLM, did you conduct sufficient self-review before submitting it?

WaterWhisperer:
> This is wrong because the check is racy (the file is immediately dropped after the check, so the path may not really correspond to the checked file).

Thanks, and sorry for missing the fd lifetime/concurrency issue. I'll fix it.

WaterWhisperer:
> Also I don't think we need `ifdef SYS_futimesat` per our previous discussion. If the code was generated by an LLM, did you conduct sufficient self-review before submitting it?

I did self-review the patch before submitting it, but I saw `SYS_futimesat` defined for x86-64 and not in the generic syscall list, and forgot the previous discussion that syscall macros from `<sys/syscall.h>` could be used directly. :(

lrh2000:
> defined for x86-64 and not in the generic syscall list

Oh, that seems to be a good point and I cannot get this point without your explanation. Thanks for your explanation..

If that's your intention, it's better to include this point in the code:
```c
// `futimesat` is a syscall that is only present in certain architectures (e.g., x86-64).
#ifdef SYS_futimesat
```

Alternatively, we could just test `SYS_utimensat`. (I assume the newly enabled gvisor test may already cover the `futimesat` path. Exploring a different path may be even better, if it is feasible.)
```
> grep -R "sys_utimensat" kernel/src/syscall/arch/
kernel/src/syscall/arch/generic.rs:            utimens::sys_utimensat,
kernel/src/syscall/arch/generic.rs:            SYS_UTIMENSAT = 88               => sys_utimensat(args[..4]);
kernel/src/syscall/arch/x86.rs:    utimens::{sys_futimesat, sys_utime, sys_utimensat, sys_utimes},
kernel/src/syscall/arch/x86.rs:    SYS_UTIMENSAT = 280        => sys_utimensat(args[..4]);
```

`SYS_utimensat` should be present in all architectures, so we won't have to deal with architecture-specific issues. Let the gvisor test handle it.

WaterWhisperer:
> Alternatively, we could just test `SYS_utimensat`

Thanks for the help and guidance!

test/initramfs/src/regression/fs/utimensat/utimensat.c
```diff
@@ -70,6 +70,19 @@ FN_TEST(legacy_null_pathname_with_nonzero_flag_returns_einval)
...
+FN_TEST(utimensat_null_pathname_with_o_path_returns_ebadf)
+{
+	int opath_fd = TEST_SUCC(open(TEST_FILE, O_PATH));
+	struct timespec times[2] = {
+		{ .tv_sec = 10, .tv_nsec = 0 },
+		{ .tv_sec = 20, .tv_nsec = 0 },
+	};
+
+	TEST_ERRNO(syscall(SYS_utimensat, opath_fd, NULL, times, 0), EBADF);
+	TEST_SUCC(close(opath_fd));
```
lrh2000, line 84: Actually I am wondering what will happen if we use `AT_EMPTY_PATH` and `path = ""`. It seems to pass, so what was changed in this PR is correct.

Let's make this explicit:
```c
	TEST_ERRNO(syscall(SYS_utimensat, opath_fd, NULL, times, 0), EBADF);
	TEST_SUCC(syscall(SYS_utimensat, opath_fd, "", times, AT_EMPTY_PATH));

	TEST_SUCC(close(opath_fd));
```

(Also add a blank line to move the `close()` call to a separate code block).

kernel/src/syscall/utimens.rs
```diff
@@ -171,27 +174,26 @@ fn do_utimes(
...
+        let fs_path = FsPath::from_fd_at(dirfd, pathname, EmptyPathStr::AllowIfFlag(flags.bits()))?;
         let fs_ref = ctx.thread_local.borrow_fs();
         let path_resolver = fs_ref.resolver().read();
         if flags.contains(UtimensFlags::AT_SYMLINK_NOFOLLOW) {
             path_resolver.lookup_no_follow(&fs_path)?
         } else {
             path_resolver.lookup(&fs_path)?
         }
+    } else {
+        // Matches Linux `do_utimes_fd`: no flags are accepted when pathname is NULL.
```
lrh2000: This needs to be updated. It seems that the special path in Linux is the exact reason why we need to reject non-zero flags and `O_PATH` files.

Also, we should add a reference link instead of mentioning the Linux function name. In this case, it's an existing problem. However, since we need to change the code here anyway, just fix it here.

```patch
diff --git a/kernel/src/syscall/utimens.rs b/kernel/src/syscall/utimens.rs
index b50a0d0db..111596281 100644
--- a/kernel/src/syscall/utimens.rs
+++ b/kernel/src/syscall/utimens.rs
@@ -184,9 +184,10 @@ fn do_utimes(
             path_resolver.lookup(&fs_path)?
         }
     } else {
-        // Matches Linux `do_utimes_fd`: no flags are accepted when pathname is NULL.
+        // Linux will reject non-zero flags and `O_PATH` files if the pathname is null.
+        // Reference: <https://elixir.bootlin.com/linux/v7.1/source/fs/utimes.c#L108>
         if !flags.is_empty() {
-            return_errno_with_message!(Errno::EINVAL, "flags must be zero when pathname is NULL");
+            return_errno_with_message!(Errno::EINVAL, "flags must be zero when the pathname is NULL");
         }
         let file = get_file_fast!(&mut file_table, dirfd.try_into()?);
```

lrh2000 (commented):
Thanks for the revision! Just two minor nits.

kernel/src/syscall/utimens.rs
WaterWhisperer: Updated and thanks a lot

lrh2000 (approved):

WaterWhisperer:
Hi @StevenJiang1110, sorry for the ping. This PR has been waiting for another maintainer review since the last approval. Could you take a look when you have time? Thanks!

StevenJiang1110 (approved):

cqs21 (approved):

PR #3387, opened 2026-06-13
Add route/kernel per-namespace socket

co63oc:
https://www.gitlink.org.cn/competitions/track1_2026Asterinas

image

Currently, there is a TODO indicating that it should be a per-namespace socket (kernel/src/net/socket/netlink/route/kernel/mod.rs:73)
If the NetlinkRouteKernelSocket returns a pointer, unsafe should be used, and it contains PhantomData, which is a zero-size type. Here, Copy is used

### DIFF
--- a/kernel/src/fs/fs_impls/procfs/pid/task/ns.rs
+++ b/kernel/src/fs/fs_impls/procfs/pid/task/ns.rs
@@ -20,7 +20,7 @@ use crate::{
         },
     },
     ipc::IpcNamespace,
-    net::uts_ns::UtsNamespace,
+    net::{net_ns::NetNamespace, uts_ns::UtsNamespace},
     prelude::*,
     process::{NsProxy, UserNamespace, posix_thread::AsPosixThread},
     thread::Thread,
@@ -52,13 +52,15 @@ enum NsProxyEntry {
     Ipc,
     /// The mount namespace.
     Mnt,
+    /// The network namespace.
+    Net,
     /// The UTS namespace.
     Uts,
 }
 
 impl NsProxyEntry {
     /// All supported `NsProxy`-backed namespace entries.
-    const ALL: &[Self] = &[Self::Cgroup, Self::Ipc, Self::Mnt, Self::Uts];
+    const ALL: &[Self] = &[Self::Cgroup, Self::Ipc, Self::Mnt, Self::Net, Self::Uts];
 
     /// Returns the filename of this namespace entry under `/proc/[pid]/ns/`.
     fn as_str(self) -> &'static str {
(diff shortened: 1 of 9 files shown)


StevenJiang1110:
Thank you for your contribution at first!

The issue you addressed is a real one, and introducing a network namespace to centrally manage network resources is indeed the right direction. However, the current implementation is quite incomplete. A key concern is that introducing a network namespace requires isolating resources such as interfaces, sockets, and more — yet this PR only isolates the rtnetlink socket, which is far from sufficient. It is therefore difficult to say that this PR establishes a correct network namespace abstraction.

Additionally, this PR supports creating new network namespaces from an existing one, which is actually a fairly complex topic. If I recall correctly, each newly created network namespace should come with its own loopback device, while physical devices (e.g., virtio-net) are only accessible from the root network namespace. None of this logic is currently in place.

My suggestion would be to narrow the scope of this PR to implementing the init network namespace — that is, placing all resources that should be isolated by a network namespace (such as interfaces, socket tables covering not just TCP and UDP but potentially also UNIX sockets and other resources, and devices) into it. This would be analogous to our current init user namespace support: we don't yet support creating new user namespaces, but the abstraction is already used in the right places (e.g., capability checks). If you also want to support creating new network namespaces, I would recommend leaving that for a follow-up PR.

lrh2000:
> yet this PR only isolates the rtnetlink socket

This PR does not even do that. It introduces a `Copy` `NetlinkRouteKernelSocket`, which makes no sense and cannot isolate anything.

StevenJiang1110:
> This PR does not even do that. It introduces a `Copy` `NetlinkRouteKernelSocket`, which makes no sense and cannot isolate anything.

You are right — this appears to be an incorrect implementation. I hadn't looked at it carefully.

co63oc:
> > This PR does not even do that. It introduces a `Copy` `NetlinkRouteKernelSocket`, which makes no sense and cannot isolate anything.
> 
> You are right — this appears to be an incorrect implementation. I hadn't looked at it carefully.

image
Because it is marked as "中" here, I think the scope of modification should not be too large, returning different objects or pointers.

If that's the case, I'll look at other parts that are easier to modify。  

https://asterinas.github.io/book/to-contribute/coding-guidelines/how-guidelines-are-written.html
image
There's no need to take up time to say much. According to Rust's rules, the code is like this, and the PR (Pull Request) is also like this, what a PR should do should be simple and effective. If changes are necessary, they should be minimal and it should be easy to determine whether they are acceptable. If significant changes are required, it can be closed
If it's a larger task, we can design it as multiple PRs

Issue #3390, opened 2026-06-13
Procfs should dynamically generate user-visible metadata


li041:
Procfs metadata is not always a plain cached inode attribute. Some procfs inodes need to generate user-visible metadata from current task, thread, or file descriptor state each time metadata is queried. This includes **_dynamic field updates_**, and in some cases can also **_fail_**.

Asterinas already has part of this pattern today. For example, procfs updates owner fields through `metadata_with_owner()` directly in `metadata()`. That works with the old `metadata() -> Metadata`  interface because this owner update path has **_no failure path_**. However, there are failure paths in `getattr` for proc inode metadata (See `Linux examples` for details).

If VFS distinguishes `cached_metadata()` from user-visible `metadata()`, procfs can keep cached metadata separate and placedynamic user-visible field generation in `metadata()` implementations.

  ## Linux examples:

  - `/proc/<pid>/task` uses` proc_task_getattr()` to dynamically add the current thread count to nlink.
    https://codebrowser.dev/linux/linux/fs/proc/base.c.html#3955

  - `/proc/<pid>/fd` uses `proc_fd_getattr()` to dynamically compute the fd directory size. This path can also return an `ENOENT` through `proc_readfd_count()` when task doesn't exist.
    https://codebrowser.dev/linux/linux/fs/proc/fd.c.html#351

  - `/proc/<tid>` base inode operations also install `.getattr = pid_getattr`.
    https://codebrowser.dev/linux/linux/fs/proc/base.c.html#3776
    - `pid_getattr()` first calls `generic_fillattr()`, then dynamically updates `uid/gid` from the current task and checks hidepid policy. If permission checks fail, it returns `ENOENT`.(Although, we didn't implement relative ptrace capacity.)
    https://codebrowser.dev/linux/linux/fs/proc/base.c.html#2018

  ## Expected outcome:

  - Audit procfs inode implementations that need dynamic user-visible metadata generation.
  - Keep cached metadata separate from user-visible metadata.
  - Move dynamic field generation such as uid, gid, nlink, and size into procfs `metadata() -> Result<Metedata>`
    implementations where appropriate.
  - Preserve Linux semantics for fd disappearance, and hidepid-related ENOENT.

PR #3391, opened 2026-06-13
Make inode metadata queries fallible

### LINKED ISSUE #3333: Add inode attribute revalidation before returning VFS metadata
### Current Feature

`Path::metadata()` is currently inherited directly from `Inode::metadata()`, and `Inode::metadata()` returns `Metadata` rather than `Result<Metadata>`.

This works well for filesystems whose inode attributes are always locally authoritative. However, it does not give filesystems a clean way to report failures when metadata must be revalidated before being returned to userspace.

Asterinas already has `RevalidationPolicy`, including `REVALIDATE_EXISTS`, but that policy currently describes dentry cache revalidation: whether a parent directory should revalidate cached positive children. It does not explicitly model inode attribute or metadata freshness.

### Enhancement Description
  Add a separate inode attribute metadata revalidation mechanism.

  The VFS should distinguish between:

  - cached metadata, which can be returned without negotiation;
  - user-visible metadata, which may require inode attribute revalidation and may fail.

 One possible API shape is to make `Inode::metadata()` return `Result<Metadata>` and add a separate cached metadata accessor:
```rust
  pub trait Inode {
      fn cached_metadata(&self) -> Metadata;

      fn attr_revalidation_policy(&self) -> AttrRevalidationPolicy {
          AttrRevalidationPolicy::empty()
      }

      fn revalidate_metadata(&self) -> Result<()> {
          Ok(())
      }

      fn metadata(&self) -> Result<Metadata> {
          if self
              .attr_revalidation_policy()
              .contains(AttrRevalidationPolicy::REVALIDATE_METADATA)
          {
              self.revalidate_metadata()?;
          }

          Ok(self.cached_metadata())
      }
  }
```

### Motivation

-   For filesystems whose metadata can change outside local VFS operations, cached inode attributes may become stale. The current `metadata() -> Metadata` interface cannot express revalidation failure, so callers cannot distinguish fresh metadata from stale cached metadata returned after a failed refresh.

### Possible Implementations

Introduce a new policy type dedicated to inode attribute metadata revalidation:
```rust
bitflags! {
    pub struct AttrRevalidationPolicy: u8 {
        const REVALIDATE_METADATA = 1 << 0;
    }
}
```
Then update the inode trait along these lines:
```rust
pub trait Inode {
    fn cached_metadata(&self) -> Metadata;

    fn attr_revalidation_policy(&self) -> AttrRevalidationPolicy {
        AttrRevalidationPolicy::empty()
    }

    fn revalidate_metadata(&self) -> Result<()> {
        Ok(())
    }

    fn metadata(&self) -> Result<Metadata> {
        if self
            .attr_revalidation_policy()
            .contains(AttrRevalidationPolicy::REVALIDATE_METADATA)
        {
            self.revalidate_metadata()?;
        }

        Ok(self.cached_metadata())
    }
}
```
### Additional Context
- TODO: Evaluate whether APIs such as size(), mode(), owner(), group(), and timestamp getters should eventually return Result or be routed through the same metadata revalidation path.

li041:
Close #3333

## Summary

  - Changes `Inode::metadata()` and `Path::metadata()` to return `Result<Metadata>`.
  - Documents that convenience inode metadata getters return `cached values`, while `metadata()` may be more expensive and should be used when callers need a `filesystem-validated` metadata snapshot.

## Note

- This PR only make inode metadata quiries falliable. Filesystem users of metadata revalidation are not implemented yet. Planned users include procfs (https://github.com/asterinas/asterinas/issues/3390) and virtio-fs.

### DIFF
--- a/kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
@@ -58,7 +58,7 @@ impl VirtioFsInode {
         // so cached reads refresh attributes before checking the cached file size.
         // If this flag becomes optional, reads that stay below EOF should be allowed
         // to use still-valid cached attributes.
-        self.revalidate_attr(fh)?;
+        self.revalidate_attr(Some(fh))?;
 
         let inner = self.inner.read();
         let file_size = self.size();
@@ -89,7 +89,7 @@ impl VirtioFsInode {
         // FIXME: Direct reads do not support short reads yet, so `FUSE_READ`
         // must request the actual bytes to read instead of the caller's maximum
         // buffer size. Refresh attributes before using the cached file size.
-        self.revalidate_attr(fh)?;
+        self.revalidate_attr(Some(fh))?;
 
         let _inner = self.inner.read();
         let file_size = self.size();
@@ -605,7 +605,7 @@ impl VirtioFsInode {
     /// so the server may return handle-specific attributes.
     pub(in crate::fs::fs_impls::virtiofs) fn revalidate_attr(
         &self,
-        fh: FuseFileHandle,
+        fh: Option<FuseFileHandle>,
     ) -> Result<()> {
         let now = MonotonicCoarseClock::get().read_time();
         if self.inner.read().is_attr_valid(now) {
@@ -614,10 +614,13 @@ impl VirtioFsInode {
 
         let fs = self.fs_ref();
         let request_attr_version = fs.session().snapshot_attr_version();
-        let attr_reply = fs.session().do_fuse_op(
-            self.nodeid(),
-            GetattrOperation::new(GetattrReq::new(GetattrFlags::GETATTR_FH, fh)),
-        )?;
+        let getattr_req = match fh {
+            Some(fh) => GetattrReq::new(GetattrFlags::GETATTR_FH, fh),
+            None => GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0)),
+        };
+        let attr_reply = fs
+            .session()
+            .do_fuse_op(self.nodeid(), GetattrOperation::new(getattr_req))?;
 
         self.commit_attr_reply(attr_reply, request_attr_version, StaleAttrAction::Discard)?;
 
--- a/kernel/src/fs/vfs/fs_apis/inode.rs
+++ b/kernel/src/fs/vfs/fs_apis/inode.rs
@@ -357,7 +357,24 @@ pub trait Inode: Any + FileOps + Send + Sync {
 
     fn resize(&self, new_size: usize) -> Result<()>;
 
-    fn metadata(&self) -> Metadata;
+    /// Returns all inode metadata.
+    ///
+    /// This method returns all metadata fields and is typically used to serve
+    /// syscalls such as `fstat`. Callers must not assume that this operation
+    /// is always cheap. For example, remote filesystems may need to perform
+    /// I/O when getting or updating metadata, and such operations may fail.
+    ///
+    /// If only a specific field is needed, prefer using the dedicated getters
+    /// such as [`size`], [`ino`], [`type_`], [`mode`], [`owner`], and [`group`], which
+    /// are generally cheaper as they return cached data from the inode.
+    ///
+    /// [`size`]: Inode::size
+    /// [`ino`]: Inode::ino
+    /// [`type_`]: Inode::type_
+    /// [`mode`]: Inode::mode
+    /// [`owner`]: Inode::owner
+    /// [`group`]: Inode::group
+    fn metadata(&self) -> Result<Metadata>;
 
     fn ino(&self) -> u64;
 
@@ -562,7 +579,7 @@ pub trait Inode: Any + FileOps + Send + Sync {
         };
 
         let creds = posix_thread.credentials();
-        let metadata = self.metadata();
+        let metadata = self.metadata()?;
         let mode = metadata.mode;
 
         // With DAC_OVERRIDE capability, the user can bypass some permission checks.
--- a/kernel/src/fs/fs_impls/devpts/mod.rs
+++ b/kernel/src/fs/fs_impls/devpts/mod.rs
@@ -239,8 +239,8 @@ impl Inode for RootInode {
         Err(Error::new(Errno::EISDIR))
     }
 
-    fn metadata(&self) -> Metadata {
-        *self.metadata.read()
+    fn metadata(&self) -> Result<Metadata> {
+        Ok(*self.metadata.read())
     }
 
     fn extension(&self) -> &Extension {
--- a/kernel/src/fs/fs_impls/devpts/ptmx.rs
+++ b/kernel/src/fs/fs_impls/devpts/ptmx.rs
@@ -79,8 +79,8 @@ impl Inode for Ptmx {
         Ok(())
     }
 
-    fn metadata(&self) -> Metadata {
-        *self.metadata.read()
+    fn metadata(&self) -> Result<Metadata> {
+        Ok(*self.metadata.read())
     }
 
     fn extension(&self) -> &Extension {
--- a/kernel/src/fs/fs_impls/devpts/slave.rs
+++ b/kernel/src/fs/fs_impls/devpts/slave.rs
@@ -68,8 +68,8 @@ impl Inode for PtySlaveInode {
         Err(Error::new(Errno::EPERM))
     }
 
-    fn metadata(&self) -> Metadata {
-        *self.metadata.read()
+    fn metadata(&self) -> Result<Metadata> {
+        Ok(*self.metadata.read())
     }
 
     fn extension(&self) -> &Extension {
--- a/kernel/src/fs/fs_impls/exfat/inode.rs
+++ b/kernel/src/fs/fs_impls/exfat/inode.rs
@@ -1422,7 +1422,7 @@ impl Inode for ExfatInode {
         Ok(())
     }
 
-    fn metadata(&self) -> Metadata {
+    fn metadata(&self) -> Result<Metadata> {
         let inner = self.inner.read();
 
         let blk_size = inner.fs().sector_size();
@@ -1433,7 +1433,7 @@ impl Inode for ExfatInode {
             1
         };
 
-        Metadata {
+        Ok(Metadata {
             ino: inner.ino,
             size: inner.size,
             optimal_block_size: blk_size,
@@ -1449,7 +1449,7 @@ impl Inode for ExfatInode {
             container_dev_id: inner.fs().container_device_id(),
             self_dev_id: None,
             birth_at: Some(inner.crtime.as_duration().unwrap_or_default()),
-        }
+        })
     }
 
     fn type_(&self) -> InodeType {
--- a/kernel/src/fs/fs_impls/ext2/impl_for_vfs/inode.rs
+++ b/kernel/src/fs/fs_impls/ext2/impl_for_vfs/inode.rs
@@ -73,8 +73,8 @@ impl Inode for Ext2Inode {
         self.resize(new_size)
     }
 
-    fn metadata(&self) -> Metadata {
-        self.metadata()
+    fn metadata(&self) -> Result<Metadata> {
+        Ok(self.metadata())
     }
 
     fn ino(&self) -> u64 {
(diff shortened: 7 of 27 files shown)


kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -345,7 +345,15 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
-    fn metadata(&self) -> Metadata;
+    /// Returns a snapshot of the inode metadata.
+    ///
+    /// Convenience metadata getters such as `size`, `ino`, `type_`, `mode`,
+    /// `owner`, `group`, and timestamp getters return cached values.
+    ///
+    /// Callers must not assume that this operation is always cheap.
+    /// For example, remote filesystems may need to perform I/O when
+    /// getting or setting metadata, and such operations may fail.
+    fn metadata(&self) -> Result<Metadata>;
```
StevenJiang1110, line 377: To be honest, the current comment is quite confusing and does not convey the right information.

First, we need to distinguish between mutable and immutable metadata. As I understand it, even for virtiofs, `ino` and `type_` are immutable — if `ino` changes, we should clearly be creating a new `VirtioFsInode` rather than reusing the existing one. On the other hand, `size`, `mode`, `owner`, `group`, and `timestamps` are all mutable. Second, the notion of whether an operation is "cheap" or not is really about whether it involves time-consuming operations — not necessarily I/O. Acquiring a mutex can also be costly due to lock contention. So I don't think framing it in terms of whether the caller considers it cheap is particularly meaningful.

What actually needs to be explained here is why `metadata` returns a `Result`, while some other metadata-related APIs — such as `size()` — do not. I find this a bit puzzling myself. Could you explain why virtiofs only requires `metadata` to return a `Result`, but not `size`?

If this is intended to be a temporary arrangement, you might consider the following approach:

```rust
/// Returns the inode metadata.
///
/// This operation is fallible. For remote filesystems, for example, fetching
/// metadata may require synchronization with a remote source, which can fail.
///
/// However, some metadata getter methods on `Inode` are currently infallible,
/// such as `ino`, `type_`, `size`, `ctime`, `atime`, and `mtime`. This is not
/// entirely correct in principle. For mutable metadata like `size`, the corresponding
/// getter methods should ideally be fallible. For immutable fields like `ino`
/// and `type_`, however, we can assume a cache that remains valid for the
/// lifetime of the inode, justifying an infallible getter (does this also
/// apply to `ctime`?).
///
/// FIXME: For `size`, `atime`, `mtime`, and similar mutable fields, their getter
/// methods should also be made fallible.
```

But I'm not sure whether it's suitable. So I'd like to return to my earlier question to decide how to proceed: why is only `metadata` fallible? What specific problem were you trying to solve? And how do you plan to handle other methods like `size`?

li041: >why is only metadata fallible?

Most direct calls to individual metadata-related getters, such as `size()`, `atime()`, `mtime()`, and `ctime()`, are internal VFS/filesystem uses. For those uses, I think returning the cached inode metadata is acceptable. These getters are not intended to be the authoritative path for refreshing all remote metadata.

For virtiofs, freshness of those fields is handled at the operation level. For example, when an operation needs an up-to-date file size, it should revalidate attributes before using the cached `size()`. The revalidation should not be hidden inside `size()` itself, because many internal users(page cache) only need the current cached snapshot.

By contrast, `metadata()` is the stat-like API that returns the full user-visible metadata snapshot. Since user-visible metadata includes multiple mutable fields, virtiofs should make sure this snapshot is valid before returning it. That is why `metadata()` needs to be fallible: refreshing remote metadata can fail, and the failure should be propagated to callers such as `stat`/`statx`.

> What specific problem were you trying to solve?
 
Also, virtiofs revalidation does not mean issuing `GETATTR` on every metadata access. The cache can first be marked expired, and the actual refresh happens when a later operation needs valid attributes. `metadata()` is one of those places where we need to refresh, because it returns the complete user-visible metadata.

> how do you plan to handle other methods like size

So my current plan is:
  - `ino()` and `type_()` remain infallible because they are immutable for the lifetime of the inode.
  - `size()`, `atime()`, `mtime()`, and `ctime()` remain infallible cached getters for internal use,  
  - And operation-specific paths that require fresh mutable metadata, such as size-sensitive operations, should explicitly revalidate before reading the cached fields, e.g. `VirtioFsInode::seek_end()`.
  - `metadata()` becomes fallible because it is the full user-visible metadata query path and may need remote synchronization.

StevenJiang1110: > Most direct calls to individual metadata-related getters, such as size(), atime(), mtime(), and ctime(), are internal VFS/filesystem uses. For those uses, I think returning the cached inode metadata is acceptable.

But this is incorrect, isn't it? If the metadata has expired, we should always refresh it on the next access rather than serving stale data. It might be worth checking Linux's behavior to see whether there are any circumstances under which expired metadata is intentionally tolerated, and if so, for what reason. If such cases exist, I think we could adopt a policy-based approach to distinguish when stale metadata is acceptable — for instance, as you suggested, allowing stale metadata for internal use while enforcing a refresh when the user explicitly requests the full metadata.

I am not necessarily asking for this change to be addressed in this PR. However, it would be good to align on the future direction before moving forward. After all, this PR is already a clear improvement over the existing code.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -357,7 +357,15 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
```
StevenJiang1110: I suggest rewording the comments as follows:

```suggestion
    /// Returns all inode metadata.
    ///
    /// This method returns all metadata fields,
    /// and is typically used to serve syscalls such as `fstat`.
    /// Callers must not assume that this operation is always cheap.
    /// For example, remote filesystems may need to perform I/O
    /// when getting or updating metadata, and such operations may fail.
    ///
    /// If only a specific field is needed,
    /// prefer using the dedicated getters such as
    /// `size`, `ino`, `type_`, `mode`, `owner`, and `group`,
    /// which are generally cheaper and infallible
    /// as they return cached data from the inode.
    fn metadata(&self) -> Result<Metadata>;
```

StevenJiang1110 (approved):
LGTM! Thanks for your contribution! Only one small nit left.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -357,7 +357,17 @@ pub trait Inode: Any + FileOps + Send + Sync {
...
+    /// Returns all inode metadata.
+    ///
+    /// This method returns all metadata fields and is typically used to serve
+    /// syscalls such as `fstat`. Callers must not assume that this operation
+    /// is always cheap. For example, remote filesystems may need to perform
+    /// I/O when getting or updating metadata, and such operations may fail.
+    ///
+    /// If only a specific field is needed, prefer using the dedicated getters
+    /// such as `size`, `ino`, `type_`, `mode`, `owner`, and `group`, which
+    /// are generally cheaper as they return cached data from the inode.
```
StevenJiang1110, line 369:
```suggestion
    /// If only a specific field is needed, prefer using the dedicated getters
    /// such as [`size`], [`ino`], `type_`, `mode`, `owner`, and `group`, which
    /// are generally cheaper as they return cached data from the inode.
    /// 
    /// [`size`]: Inode::size
    /// [`ino`]: Inode::ino
```

I just realized that all these method names can be linked to their corresponding methods. I only provided two as examples, so please update the remaining ones accordingly.

cchanging:
> This PR only make inode metadata quiries falliable. Filesystem users of metadata revalidation are not implemented

In this PR, you introduced a fallible metadata API, but there is no actual user of it yet. That feels a bit odd and doesn’t really align with our requirements for merging PRs. Is there any obstacle to implementing the fallible metadata method for virtiofs in this PR as well?

kernel/src/fs/fs_impls/virtiofs/inode/ops.rs
```diff
@@ -614,10 +614,13 @@ impl VirtioFsInode {
 
         let fs = self.fs_ref();
         let request_attr_version = fs.session().snapshot_attr_version();
-        let attr_reply = fs.session().do_fuse_op(
-            self.nodeid(),
-            GetattrOperation::new(GetattrReq::new(GetattrFlags::GETATTR_FH, fh)),
-        )?;
+        let getattr_req = match fh {
+            Some(fh) => GetattrReq::new(GetattrFlags::GETATTR_FH, fh),
+            None => GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0)),
```
cchanging, line 619: What is the semantic of using `GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0))` here? I thought we can use the `fh` in `open_handles` of a `VirtiofsInode`.

li041: `FUSE_GETATTR_FH` is only used when the VFS already has a file handle available (e.g., `fstat()`); ordinary `stat()` requests use the inode nodeid without `FUSE_GETATTR_FH`. Since this is an inode-level getattr, there may be no open handle available, so falling back to a plain `GETATTR` is consistent with FUSE implementation.

cchanging: So based on your description, in `VirtiofsInode`'s `metadata`, should the `fh` we pass in be the `Option` returned by `open_handles.find_handle`, rather than passing `None` directly? I'm not sure what the exact behavioral difference would be here.

li041: I suggest keeping current behavior for two reasons:

1. From the implementation perspective, `open_handles` is not intended to provide an arbitrary file handle for `GETATTR`. Its lookup APIs are **_access-mode aware_** (find_readable_handle / find_writable_handle), and there is no notion of "find any handle". Since `GETATTR` fundamentally identifies the inode by its `nodeid`, probing the open-handle set here is unnecessary.
2. This also matches Linux's FUSE semantics. `FUSE_GETATTR_FH` is optional, and Linux exposes the same model: callers provide an fh only when they already have one (e.g., file-based operations), while a regular inode GETATTR is issued without an fh. The open-handle table is primarily maintained for writeback/write operations rather than as a source of file handles for GETATTR.

cchanging: > callers provide an fh only when they already have one (e.g., file-based operations), while a regular inode GETATTR is issued without an fh. 

Yes, but in the current implementation, `sys_fstat` locates the corresponding file and retrieves its `metadata` via the `Path` of the file, eventually calling directly into `inode::metadata`. As a result, even though it is a file operation and we have a valid `fh`, the `fh` isn't actually used—which appears inconsistent with Linux semantics. 

The question I've been consistently concerned about is: for this particular case, are there any semantic differences between using `GetattrReq::new(GetattrFlags::GETATTR_FH, fh)` versus `GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0))`? If there are, I think we should still leverage `open_handles` to attempt retrieving the handle; but if there aren't any differences, why not just use `GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0))` across the board? This is precisely where my main confusion lies.

li041: >are there any semantic differences between using GetattrReq::new(GetattrFlags::GETATTR_FH, fh) versus GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0))

It seems that fh doesn't influence the result for [virtiofsd](https://virtio-fs.gitlab.io/virtiofsd/doc/src/virtiofsd/passthrough/mod.rs.html#). 

> why not just use GetattrReq::new(GetattrFlags::empty(), FuseFileHandle::new(0)) across the board?

Protocol-wise, yes, other filesystem implementation may choose to verify and interpret the handle.

> As a result, even though it is a file operation and we have a valid fh, the fh isn't actually used—which appears inconsistent with Linux semantics. 

I think the solution is to pass the `fh` if needed instead of using `open_handles` as an alternative, btw, current interface of `Openhandles` doesn't fit this scenario. It requires `access_mode` which is naive for I/O path, while seems strange for `getattr`.

cchanging: Thank you for the explanation! I took a look at the Linux source code and found that both `stat` and `fstat` eventually call `fuse_do_getattr`, so my previous understanding was incorrect. It seems that, in Linux, `GETATTR_FH` is more something used internally by the implementation, although I still do not fully understand the rationale behind doing it that way. 

Anyway, the current implementation should be consistent with Linux.

cchanging (approved):

Need to resolve the conflicts.

PR #3395, opened 2026-06-15
Fix EpollTest.CycleOfOneDisallowed

co63oc:
Fix EpollTest.CycleOfOneDisallowed, EpollTest.CycleOfThreeDisallowed, update test/initramfs/src/conformance/gvisor/blocklists/epoll_test

### DIFF
--- a/kernel/src/events/epoll/file.rs
+++ b/kernel/src/events/epoll/file.rs
@@ -26,6 +26,10 @@ use crate::{
     util::ioctl::RawIoctl,
 };
 
+/// Global mutex to prevent parallel cycle formation in epoll topologies.
+/// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/eventpoll.c#L258>
+static EPNESTED_MUTEX: Mutex<()> = Mutex::new(());
+
 /// A file-like object that provides epoll API.
 ///
 /// Conceptually, we maintain two lists: one consists of all interesting files,
@@ -48,6 +52,19 @@ pub struct EpollFile {
     ready: Arc<ReadySet>,
     /// The pseudo path associated with this epoll file.
     pseudo_path: Path,
+    /// Serializes epoll_ctl operations (check + insert) to prevent TOCTOU races.
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/eventpoll.c#L2316>
+    mtx: Mutex<()>,
+}
+
+/// Result of the reachability check.
+enum ReachResult {
+    /// Cycle detected.
+    Found,
+    /// Nesting depth exceeded limit.
+    TooDeep,
+    /// Not found.
+    NotFound,
 }
 
 impl EpollFile {
@@ -59,10 +76,12 @@ impl EpollFile {
             interest: Mutex::new(BTreeSet::new()),
             ready: Arc::new(ReadySet::new()),
             pseudo_path,
+            mtx: Mutex::new(()),
         })
     }
 
     /// Controls the interest list of the epoll file.
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/eventpoll.c#L2316>
     pub fn control(&self, thread_local: &ThreadLocal, cmd: &EpollCtl) -> Result<()> {
         let fd = match cmd {
             EpollCtl::Add(fd, ..) => *fd,
@@ -73,6 +92,27 @@ impl EpollFile {
         let mut file_table = thread_local.borrow_file_table_mut();
         let file = get_file_fast!(&mut file_table, fd).into_owned();
         drop(file_table);
+
+        // Reject cycles: an epoll file cannot monitor itself or form indirect cycles.
+        if let EpollCtl::Add(..) = cmd {
+            if (self as *const _ as *const ()) == (file.as_ref() as *const _ as *const ()) {
+                return_errno_with_message!(Errno::EINVAL, "epoll file cannot be added to itself");
+            }
+            if let Some(target_epoll) = file.downcast_ref::<EpollFile>() {
+                // Lock ordering: epnested_mutex -> mtx
+                let _epnested_guard = EPNESTED_MUTEX.lock();
+                self.check_cycle(target_epoll)?;
+
+                let _mtx_guard = self.mtx.lock();
+                let EpollCtl::Add(fd, ep_event, ep_flags) = *cmd else {
+                    unreachable!()
+                };
+                return self.add_interest(fd, file, ep_event, ep_flags);
+            }
+        }
+
+        // For non-epoll ADD or DEL/MOD operations
+        let _mtx_guard = self.mtx.lock();
 
         match *cmd {
             EpollCtl::Add(fd, ep_event, ep_flags) => {
@@ -249,6 +289,84 @@ impl EpollFile {
     fn warn_unsupported_flags(&self, flags: &EpollFlags) {
         if flags.intersects(EpollFlags::EXCLUSIVE | EpollFlags::WAKE_UP) {
             warn!("{:?} contains unsupported flags", flags);
+        }
+    }
+
+    /// Maximum number of nesting allowed inside epoll sets.
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/eventpoll.c#L94>
+    const EP_MAX_NESTS: usize = 4;
+
+    /// Checks whether adding `target` as an interest of `self` would create a cycle
+    /// or exceed the maximum nesting depth.
+    /// Reference: <https://elixir.bootlin.com/linux/v6.18/source/fs/eventpoll.c#L2132>
+    fn check_cycle(&self, target: &EpollFile) -> Result<()> {
+        let self_ptr = self as *const Self as *const ();
+        let mut visited = Vec::new();
+        match target.can_reach(self_ptr, &mut visited, 0) {
+            ReachResult::Found => {
+                return_errno_with_message!(
+                    Errno::ELOOP,
+                    "adding this fd would create an epoll cycle"
+                );
+            }
+            ReachResult::TooDeep => {
+                return_errno_with_message!(
+                    Errno::ELOOP,
+                    "adding this fd would exceed the maximum epoll nesting depth"
+                );
+            }
+            ReachResult::NotFound => Ok(()),
+        }
+    }
+
+    /// Checks whether an epoll file at the given raw pointer is reachable by
+    /// recursively traversing the epoll files monitored by `self`.
+    fn can_reach(
+        &self,
+        target_ptr: *const (),
+        visited: &mut Vec<*const ()>,
+        depth: usize,
+    ) -> ReachResult {
+        let self_ptr = self as *const Self as *const ();
+
+        if self_ptr == target_ptr {
+            return ReachResult::Found;
+        }
+
+        if depth > Self::EP_MAX_NESTS {
+            return ReachResult::TooDeep;
+        }
+
+        if visited.contains(&self_ptr) {
+            return ReachResult::NotFound;
+        }
+
+        visited.push(self_ptr);
+
+        let files: Vec<Arc<dyn FileLike>> = {
+            let interest = self.interest.lock();
+            interest.iter().filter_map(|entry| entry.0.file()).collect()
+        };
+
+        let mut max_depth_reached = false;
+        for file in &files {
+            if let Some(ep) = file.downcast_ref::<EpollFile>() {
+                let ep_ptr = ep as *const EpollFile as *const ();
+                if ep_ptr == target_ptr {
+                    return ReachResult::Found;
+                }
+                match ep.can_reach(target_ptr, visited, depth + 1) {
+                    ReachResult::Found => return ReachResult::Found,
+                    ReachResult::TooDeep => max_depth_reached = true,
+                    ReachResult::NotFound => {}
+                }
+            }
+        }
+
+        if max_depth_reached {
+            ReachResult::TooDeep
+        } else {
+            ReachResult::NotFound
         }
     }
 }
--- a/kernel/src/events/epoll/entry.rs
+++ b/kernel/src/events/epoll/entry.rs
@@ -104,7 +104,7 @@ impl Entry {
     ///
     /// Since an epoll entry only holds a weak reference to the file,
     /// it is possible (albeit unlikely) that the file has been dropped.
-    fn file(&self) -> Option<Arc<dyn FileLike>> {
+    pub(super) fn file(&self) -> Option<Arc<dyn FileLike>> {
         self.key.file.upgrade().map(KeyableArc::into)
     }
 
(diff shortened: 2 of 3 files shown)


kernel/src/syscall/epoll.rs
```diff
@@ -64,6 +64,10 @@ pub fn sys_epoll_ctl(
     const EPOLL_CTL_DEL: i32 = 2;
     const EPOLL_CTL_MOD: i32 = 3;
 
+    if epfd == fd {
+        return_errno_with_message!(Errno::EINVAL, "epfd and fd are the same");
+    }
```
StevenJiang1110: https://elixir.bootlin.com/linux/v7.0/source/fs/eventpoll.c#L2272

```C
    /*
	 * We have to check that the file structure underneath the file descriptor
	 * the user passed to us _is_ an eventpoll file. And also we do not permit
	 * adding an epoll file descriptor inside itself.
	 */
	error = -EINVAL;
	if (fd_file(f) == fd_file(tf) || !is_file_epoll(fd_file(f)))
		goto error_tgt_fput;
```

Looking at the Linux implementation, it does not directly compare file descriptors for equality — instead, it checks whether both file descriptors refer to the same underlying file. We should do the same:

```rust
    let file = get_file_fast!(&mut file_table, epfd.try_into()?).into_owned();

    let controlled_file = get_file_fast!(&mut file_table, fd.try_into()?);
    if Arc::ptr_eq(&controlled_file, &file) {
        return_errno_with_message!(Errno::EINVAL, ".....");
    }
```

It might also be cleaner to add this check right after line 90?

StevenJiang1110 (approved):
Thanks for fixing this problem!

kernel/src/syscall/epoll.rs
co63oc: Thanks. Actually there are two tests here. The other one is EpollTest.CycleOfThreeDisallowed. I tried to modify it as such, but CycleOfThreeDisallowed still had errors. Therefore, I only modified CycleOfOneDisallowed. 
 This modification was not very good, and I have fixed it now

lrh2000: The new check is still racy because the file behind a FD can change concurrently. So this is a [TOCTOU problem](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use).

In other words, the file can be changed immediately after it is obtained from `get_file_fast`, so the check won't make any sense if you drop it. Take care with races in kernel programming.

This check has to be done in `Epolll::control`, e.g.,
```patch
diff --git a/kernel/src/events/epoll/file.rs b/kernel/src/events/epoll/file.rs
index 5675bb3e2..db37965dc 100644
--- a/kernel/src/events/epoll/file.rs
+++ b/kernel/src/events/epoll/file.rs
@@ -74,6 +74,14 @@ impl EpollFile {
         let file = get_file_fast!(&mut file_table, fd).into_owned();
         drop(file_table);
 
+        // A deadlock can occur if an epoll file becomes ready, wakes up other epoll files, and then
+        // loops back to itself. So we must reject all such loops in advance.
+        //
+        // FIXME: A complete loop detection mechanism needs to be implemented here.
+        if core::ptr::addr_eq(self, file.as_ref()) {
+            return_errno_with_message!(Errno::EINVAL, "epoll files must not form a loop");
+        }
+
         match *cmd {
             EpollCtl::Add(fd, ep_event, ep_flags) => {
                 self.add_interest(fd, file, ep_event, ep_flags)
```

> CycleOfThreeDisallowed

We need to have a complete loop detection mechanism here, as you've noted and I wrote in the comment above. It is of course a hard job and the Linux implementation is also quite complex.

co63oc: done, EpollTest.CycleOfThreeDisallowed can pass the test @lrh2000

lrh2000: Firstly, please try to be more concise when writing comments. Most of the comments you have added here are clearly generated by an LLM and have no value.


 - Please read my previous comments. https://github.com/asterinas/asterinas/pull/3371#discussion_r3425708960
> I'm really annoyed by LLM-generated comments. Please do not mention Linux internal type and function names. Providing a reference link describing the behavior is sufficient.
> 
>     * **Bad:** `Matches Linux xxx method, which yyy`
> 
>     * **Good:** `Linux will yyy. Reference: <Linux source code link>`
 - If there is a good reason for doing something --- for instance, forming a loop would cause a deadlock --- then explaining the reason is sufficient. There is no need to mention the behavior of the Linux operating system. At least, we should not use too many words to describe the Linux implementation details, as that makes no sense. The Linux side may have a similar implementation, but that may not be the main reason why we do something.
 - You don't need to repeat what I said in the code comments. This is a basic error to avoid, and there is no need to include it in the code comments. Otherwise, our entire codebase would be filled with such useless comments.

---

Could you please read the code yourself and write some comments by hand? This will demonstrate your understanding of the code. First, you need to learn how to do things yourself so that you can fix the LLM's output. If you don't know how to do something, you won't be able to fix the LLM either.

lrh2000 (changes requested):
Changes generated by LLM that require further review. Please try to self-review the changes first.

kernel/src/syscall/epoll.rs
co63oc: So, I understand the current steps as follows:
The code and its associated tests are without issue and do not require modification
The comment needs to be corrected and modified to a link in the Linux code. It has been modified
Because LLM is currently being widely promoted, I am gradually getting familiar with it
In interactions with numerous LLMs and Agents, it seems somewhat out of place to write comments completely

co63oc:
> Changes generated by LLM that require further review. Please try to self-review the changes first.

Ok, because LLM is currently being widely promoted, I am gradually getting familiar with it

kernel/src/syscall/epoll.rs
lrh2000: > The code and its associated tests are without issue and do not require modification

I didn't even start checking it. I skimmed the changes and found too many comments generated by an LLM. There may be (significant) problems with the implementation as well.

> Because LLM is currently being widely promoted, I am gradually getting familiar with it
> In interactions with numerous LLMs and Agents, it seems somewhat out of place to write comments completely

I'm not sure what you mean by that. My point is that you can't use LLMs to generate code and comments that you don't understand and can't write by hand.

It's fine if you **are able to** write some code by hand in 60 minutes but you just use LLMs to write **the same piece of code** in only 10 minutes. But if you don't know how to write the code, then you shouldn't use an LLM to generate code for you. Instead, you should learn and understand the existing code as much as possible. Only after you know exactly what you want should you start using an LLM.

In other words, you should be able to review the LLM's output. By "being able", you need to understand the existing code. Simply forwarding a problem, such as a failed gVisor test or a reviewer's comment, to an LLM will not help you learn.

kernel/src/events/epoll/file.rs
```diff
@@ -251,6 +273,61 @@ impl EpollFile {
...
+        {
+            return false;
+        }
+
+        // Collect the files in the interest list, then drop the lock before recursing.
+        // We keep `Arc<dyn FileLike>` alive so that `downcast_ref` references remain valid.
+        let files: Vec<Arc<dyn FileLike>> = {
+            let interest = self.interest.lock();
+            interest.iter().filter_map(|entry| entry.0.file()).collect()
+        };
```
lrh2000, line 349: The implementation is likely to be fundamentally wrong either. How can you prove the code to be race-free?

kernel/src/events/epoll/file.rs
```diff
@@ -251,6 +273,61 @@ impl EpollFile {
...
+    fn can_reach(&self, target: &EpollFile, visit_id: usize) -> bool {
+        if (self as *const _ as *const ()) == (target as *const _ as *const ()) {
+            return true;
+        }
+
+        // Skip already-visited epoll files.
+        if self
+            .visit_id
+            .swap(visit_id, core::sync::atomic::Ordering::Relaxed)
+            == visit_id
```
lrh2000: Fundamentally wrong. How can this code be correct if the system calls are called concurrently?

lrh2000:
So please double-check that you understand and follow this guideline. I am now very skeptical that you understand what the code is doing.

> It's fine if you **are able to** write some code by hand in 60 minutes but you just use LLMs to write **the same piece of code** in only 10 minutes. But if you don't know how to write the code, then you shouldn't use an LLM to generate code for you. Instead, you should learn and understand the existing code as much as possible. Only after you know exactly what you want should you start using an LLM.
> 
> In other words, you should be able to review the LLM's output. By "being able", you need to understand the existing code. Simply forwarding a problem, such as a failed gVisor test or a reviewer's comment, to an LLM will not help you learn.

co63oc:
 I searched from the list of issues, but the race condition issue was not mentioned in the list. Perhaps I need to check against the code

lrh2000:
> https://github.com/asterinas/ccf-open-inno-competition I searched from the list of issues, but the race condition issue was not mentioned in the list.

We are a kernel project. Our goal is to develop a kernel that can function properly in a production environment. There's no need to list all the rules a correct kernel should follow. It is important to become familiar with kernel programming.

---

Also: https://www.gitlink.org.cn/competitions/track1_2026Asterinas?login=ppr2eztn8&websiteName=educoder

> 三、评审标准框架
> 
> 参赛者按照通常的开源贡献惯例和流程通过PR提交代码，只有被上游仓库合并的PR才会被算作参赛评奖的依据。上游仓库有严格的开源代码审查制度，会从代码的正确性、高效性、安全性、可读性、可维护性等多个方面做代码评审，并给出具体反馈意见。与普通开源贡献者一样，参赛者需要根据这些代码评审意见完善和迭代PR，直到被上游接收。
> 
> 在提交PR之前，参赛同学应该认真阅读编码指南（https://asterinas.github.io/book/to-contribute/coding-guidelines/index.html）、研读代码库现有风格和规范，遵循最佳实践，保证提交代码的贡献。
> 
> 对于AI辅助生成的代码，作者必须认真履行自我评审的义务，完全理解代码、确保代码质量之后，再提交正式PR。
> 
> 对于不符合规范的代码或明显低质量的PR，上游仓库的维护者有权拒绝评审。

Especially:

 1. `正确性、高效性、安全性、可读性、可维护性`

This includes ensuring the code is correct. Racy code is not correct code.

 2. `对于AI辅助生成的代码，作者必须认真履行自我评审的义务，完全理解代码、确保代码质量之后，再提交正式PR。`

So you need to understand the code.

kernel/src/events/epoll/file.rs
co63oc: Here the modification has added the use of an EpollFile mutex lock for check_cycle.

kernel/src/events/epoll/file.rs
co63oc: Here the modification has added the use of an EpollFile mutex lock

co63oc:
> So please double-check that you understand and follow this guideline. I am now very skeptical that you understand what the code is doing.
> 
> > It's fine if you **are able to** write some code by hand in 60 minutes but you just use LLMs to write **the same piece of code** in only 10 minutes. But if you don't know how to write the code, then you shouldn't use an LLM to generate code for you. Instead, you should learn and understand the existing code as much as possible. Only after you know exactly what you want should you start using an LLM.
> > In other words, you should be able to review the LLM's output. By "being able", you need to understand the existing code. Simply forwarding a problem, such as a failed gVisor test or a reviewer's comment, to an LLM will not help you learn.

Refer to https://elixir.bootlin.com/linux/v6.18/source/fs/eventpoll.c#L2247
The logic in the code has been modified, and a global lock and an EpollFile mutex have been added. A depth judgment for the loop has been added. If it exceeds a certain number, it is also determined as not being able to add.
The global lock is used to check for cycles, while the EpollFile mutex is used to check for additions or modifications. The logic now seems clearer

Global Lock (EPNESTED_MUTEX)
Prevents parallel cycle formation across multiple epoll instances. When multiple threads concurrently execute EPOLL_CTL_ADD on different epoll files, each thread's cycle check might pass individually.
Lock ordering: epnested_mutex → mtx (acquired before the per-instance lock)
Example scenario without this lock:
Thread A: epoll_ctl(ep1, ADD, ep2) - cycle check passes
Thread B: epoll_ctl(ep2, ADD, ep1) - cycle check passes
Result: ep1→ep2→ep1 

Per-Instance Lock (mtx)
Serializes operations on a single EpollFile to prevent TOCTOU (Time-of-Check-Time-of-Use) races. It protects the entire check-then-insert sequence in epoll_ctl, ensuring that between the cycle check and the actual insertion.

Combined Usage
For EPOLL_CTL_ADD on an epoll file:
Acquire EPNESTED_MUTEX (global)
Execute check_cycle (traverse graph, no mtx held)
Acquire mtx (per-instance)
Execute add_interest
Release both locks

tatetian:
@co63oc We welcome first-time contributors like you, whether you are participating in a competition or not. But you must show that you have the sufficient expertise and professionalism. If you can't fully understand how the code works, I suggest that you pick an easier problem that is more manageable to your skill level. Epoll is one of the key kernel mechanism and due to its event-driven nature, the implementation must be carefully crafted to deal with all kinds of concurrency conditions. 

> the race condition issue was not mentioned in the list.

Given the fact that you are not even aware of the importance of handling race conditions, I can safely say that you have little prior experience in kernel programing and can hardly take the full responsibility for the code changing epoll.

In addition, I skimmed through the code. It is full of references to Linux's internal logic and use of pointers. This suggests that you instructed AI to implement the Rust code to mirror Linux's C code. No Asterinas code is written by "translating" Linux's code code function by function. Translation of GPL code must also be licensed as GPL, but Asterinas is MPL licensed.

Asterinas code does include references to Linux's source code, but we do so only when we try to match Linux's definitions, ABIs, or user-visible behaviors that are not documented and can only be understood by reading the source code.

Similar to other open-source projects, we reserve the right to shut down or deny reviewing low-quality AI-generated PRs. For the reasons above, I am using this right to close this PR.

You are still welcome to pick easier problems that match your skill level. Thank you.

co63oc:
> @co63oc We welcome first-time contributors like you, whether you are participating in a competition or not. But you must show that you have the sufficient expertise and professionalism. If you can't fully understand how the code works, I suggest that you pick an easier problem that is more manageable to your skill level. Epoll is one of the key kernel mechanism and due to its event-driven nature, the implementation must be carefully drafted to deal with all kinds of concurrency conditions.
> 
> > the race condition issue was not mentioned in the list.
> 
> Given the fact that you are not even aware of the importance of handling race conditions, I can safely say that you have little prior experience in kernel programing and can hardly take the full responsibility for the code changing epoll.
> 
> In addition, I skimmed through the code. It is full of references to Linux's internal logic and use of pointers. This suggests that you instructed AI to implement the Rust code to mirror Linux's C code. No Asterinas code is written by "translating" Linux's code code function by function. Translation of GPL code must also be licensed as GPL, but Asterinas is MPL licensed.
> 
> Asterinas code does include references to Linux's source code, but we do so only when we try to match Linux's definitions, ABIs, or user-visible behaviors that are not documented and can only be understood by reading the source code.
> 
> Similar to other open-source projects, we reserve the right to shut down or deny reviewing low-quality AI-generated PRs. For the reasons above, I am using this right to close this PR.
> 
> You are still welcome to pick easier problems that match your skill level. Thank you.

This is very wise and will be of great help to those who need to use AI in the future. Each self-contained system should have a set of rules and gradually move forward. Thank you for your reply. I will do my best to create contributions

PR #3396, opened 2026-06-16
Add conformance test config `CONFORMANCE_TEST_SELECTOR`

co63oc:
Add the configuration `CONFORMANCE_TEST_BINARY` to `CONFORMANCE_TEST` to test a specific test item
We can specify the `CONFORMANCE_TEST_BINARY` parameter to run the test
```
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=gvisor CONFORMANCE_TEST_BINARY="epoll_test" BLOCK="EpollTest.CloseFile:EpollTest.Oneshot"
```

The check has been passed for /aster-code-review


co63oc:
@cqs21 Please review whether this PR is valuable. Sometimes, it is necessary to test a specific item separately during testing, so variables have been added

Makefile
```diff
@@ -111,6 +111,7 @@ ENABLE_CONFORMANCE_TEST := true
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="CONFORMANCE_TEST_SUITE=$(CONFORMANCE_TEST_SUITE)"
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="CONFORMANCE_TEST_WORKDIR=$(CONFORMANCE_TEST_WORKDIR)"
 CARGO_OSDK_BUILD_ARGS += --kcmd-args="EXTRA_BLOCKLISTS=$(EXTRA_BLOCKLISTS)"
+CARGO_OSDK_BUILD_ARGS += --kcmd-args="TEST_ITEM=$(TEST_ITEM)"
```
tatetian: Give this variable a prefix to make it clear the effect scope of this variable. A candidate name would be `CONFORMANCE_TEST_CASE`. (We never used the term test item before. I suppose that test item means a test case? If so, then call it test case for consistency.)

test/initramfs/src/conformance/gvisor/run_gvisor_test.sh
```diff
@@ -53,8 +53,15 @@ run_one_test(){
...
 rm -rf $TEST_TMP_DIR/*
 
-for syscall_test in $(find $TEST_BIN_DIR/. -name \*_test) ; do
-    test_name=$(basename "$syscall_test")
+# If the TEST_ITEM environment variable is set, only run that specific test;
+# otherwise run all *_test binaries found in the test directory.
+if [ -n "$TEST_ITEM" ]; then
+    TEST_LIST="$TEST_ITEM"
+else
+    TEST_LIST=$(find $TEST_BIN_DIR/. -name \*_test -exec basename {} \;)
```
tatetian: There are three conformance test suites: gvisor, LTP, and kselftest. If gvisor gains this ability to pick a specific test, then I think that it is reasonable to say that LTP and kselftest should also gain the same ability.

tatetian (commented):
Sounds like a reasonable env var.

test/initramfs/src/conformance/gvisor/run_gvisor_test.sh
co63oc:
```
CREATE_ENTRIES=1 $LTP_DIR/runltp -f syscalls -Q -p -d $TEST_TMP_DIR -l $LOG_FILE

TESTS="$KSELFTEST_DIR/kselftest-list.txt"
```
ltp和kselftest传递的是文件，没有单独参数设置，实现更复杂一些，需要再查看

Makefile
co63oc: 已修改为 CONFORMANCE_TEST_CASE

WaterWhisperer: It's useful for iterating on gVisor tests, but this selects a test binary such as `access_test`/`open_test`, not a single gtest case. Would `CONFORMANCE_TEST_BINARY` be clearer? 
I often need to run filters like `./access_test --gtest_filter=AccessTest.InvalidModeRoot:AccessTest.InvalidModes`, where `AccessTest.InvalidModeRoot` and `AccessTest.InvalidModes` are the actual gtest cases.

co63oc: 已修改为 `CONFORMANCE_TEST_BINARY` 。如果经常使用，可以写入脚本执行，脚本里也可以修改其他名称
已修改 `gtest_filter` 参数使用`BLOCK`传递，主要因为 `test/initramfs/src/conformance/gvisor/run_gvisor_test.sh` 中使用的是`BLOCK`参数
```
    if [ -f $TEST_BIN_DIR/$1 ]; then
        get_blocklist_subtests $1
        cd $TEST_BIN_DIR && ./$1 --gtest_filter=-$BLOCK
        ret=$?
```
参数调用示例
```
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=gvisor CONFORMANCE_TEST_BINARY="epoll_test" BLOCK="EpollTest.CloseFile:EpollTest.Oneshot"
```

cqs21:
Thanks for working on this. This is a useful workflow: when developing a new feature or fixing a bug, we often need to rerun only one or a few conformance tests, and for gVisor we may also need to run only a few gtest cases inside one test binary.

I think the current PR is still gVisor-specific for a `CONFORMANCE_*` interface, and `BLOCK` mixes two different concepts:

- selecting tests to run;
- excluding known-bad tests through blocklists.

`BLOCK` is also passed as a negative gtest filter (`--gtest_filter=-...`), so it skips the listed gtest cases instead of running only those cases. More importantly, if user-provided `BLOCK` bypasses the existing blocklists, the behavior becomes hard to reason about.

I suggest a general conformance filtering mechanism:

```make
# Comma-separated tests to run within the selected conformance suite.
CONFORMANCE_TESTS ?=

# Controls whether conformance blocklists are applied.
# Values:
# - auto: apply blocklists for full-suite runs, ignore blocklists for explicit selections.
# - apply: always apply blocklists.
# - ignore: never apply blocklists.
CONFORMANCE_BLOCKLIST_MODE ?= auto

# gVisor-only positive gtest filter, used inside one selected gVisor test binary.
GVISOR_GTEST_FILTER ?=
```

The key behavior is:

| Mode | `CONFORMANCE_TESTS` empty | `CONFORMANCE_TESTS` non-empty |
| --- | --- | --- |
| `auto` | apply blocklists | ignore blocklists |
| `apply` | apply blocklists | apply blocklists |
| `ignore` | ignore blocklists | ignore blocklists |

This keeps CI/default full-suite runs unchanged, while making the development workflow natural: if a developer explicitly selects a test, the default should be to run it even if it is currently blocklisted.

`CONFORMANCE_TESTS` would mean the top-level test identity of the selected suite:

| Suite | `CONFORMANCE_TESTS` item | Example |
| --- | --- | --- |
| `gvisor` | test binary | `epoll_test` |
| `ltp` | LTP testcase name | `open01` |
| `kselftest` | full `kselftest-list.txt` entry | `timers:posix_timers` |
| `xfstests` | xfstests case id | `generic/001` |

Examples:

```sh
# Run one gVisor binary, ignoring blocklists by default.
make run_kernel AUTO_TEST=conformance \
    CONFORMANCE_TEST_SUITE=gvisor \
    CONFORMANCE_TESTS=epoll_test

# Run selected gtest cases inside one gVisor binary.
make run_kernel AUTO_TEST=conformance \
    CONFORMANCE_TEST_SUITE=gvisor \
    CONFORMANCE_TESTS=epoll_test \
    GVISOR_GTEST_FILTER=EpollTest.CloseFile:EpollTest.Oneshot

# Run selected LTP tests, including tests that are currently blocklisted.
make run_kernel AUTO_TEST=conformance \
    CONFORMANCE_TEST_SUITE=ltp \
    CONFORMANCE_TESTS=open01,open02,epoll_wait01

# Run selected gVisor binary but still apply blocklists.
make run_kernel AUTO_TEST=conformance \
    CONFORMANCE_TEST_SUITE=gvisor \
    CONFORMANCE_TESTS=epoll_test \
    CONFORMANCE_BLOCKLIST_MODE=apply
```

Implementation-wise, LTP needs special care because its current Makefile applies filesystem-specific blocklists at packaging time and only copies the selected binaries into initramfs. So `CONFORMANCE_TESTS` and `CONFORMANCE_BLOCKLIST_MODE` should be passed both to the initramfs/Nix packaging stage and to the runtime environment. Otherwise, a selected-but-blocklisted LTP test may fail simply because its binary was not packaged.

For gVisor, `GVISOR_GTEST_FILTER` should be a positive filter. If blocklists are applied, the runner can combine it with the computed negative blocklist filter:

```sh
selected_filter=${GVISOR_GTEST_FILTER:-*}
blocked_filter=$(get_gvisor_blocked_filter "$test_name")

if [ "$blocklist_mode" = "apply" ] && [ -n "$blocked_filter" ]; then
    gtest_filter="${selected_filter}-${blocked_filter}"
else
    gtest_filter="$selected_filter"
fi
```

This separates selection from exclusion, keeps the default CI behavior stable, and gives developers an explicit way to run previously blocklisted tests after a fix.

co63oc:
> Thanks for working on this. This is a useful workflow: when developing a new feature or fixing a bug, we often need to rerun only one or a few conformance tests, and for gVisor we may also need to run only a few gtest cases inside one test binary.
> 
> I think the current PR is still gVisor-specific for a `CONFORMANCE_*` interface, and `BLOCK` mixes two different concepts:
> 
> * selecting tests to run;
> * excluding known-bad tests through blocklists.
> 
> `BLOCK` is also passed as a negative gtest filter (`--gtest_filter=-...`), so it skips the listed gtest cases instead of running only those cases. More importantly, if user-provided `BLOCK` bypasses the existing blocklists, the behavior becomes hard to reason about.
> 
> I suggest a general conformance filtering mechanism:
> 
> ```makefile
> # Comma-separated tests to run within the selected conformance suite.
> CONFORMANCE_TESTS ?=
> 

已修改，大概测试了没什么问题，如果有问题改动应该也不大 @cqs21 
```
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=ltp CONFORMANCE_TESTS="open01,read01,write01"
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=gvisor CONFORMANCE_TESTS="epoll_test"
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=gvisor CONFORMANCE_TESTS="epoll_test" GVISOR_GTEST_FILTER="EpollTest.CloseFile:EpollTest.Oneshot"
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest CONFORMANCE_TESTS="exec:binfmt_script.py,timers:posix_timers"
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=xfstests CONFORMANCE_TESTS="generic/001,generic/002,ext2/001"
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=gvisor CONFORMANCE_TESTS="epoll_test" CONFORMANCE_BLOCKLIST_MODE=apply
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=gvisor CONFORMANCE_BLOCKLIST_MODE=ignore
```

test/initramfs/src/conformance/ltp/run_ltp_test.sh
```diff
@@ -7,10 +7,88 @@ TEST_TMP_DIR=${CONFORMANCE_TEST_WORKDIR:-/tmp}
 LOG_FILE=$TEST_TMP_DIR/result.log
 RESULT=0
 
+CONFORMANCE_BLOCKLIST_MODE=${CONFORMANCE_BLOCKLIST_MODE:-auto}
```
cqs21: The `CONFORMANCE_BLOCKLIST_MODE` is read but never used in the LTP runner.

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -31,54 +56,71 @@ else
...
+                exit 1
+                ;;
+        esac
+    done < "$COMBINED_BLOCKLISTS"
+
+    sort -u -o "$RESOLVED_BLOCKLISTS" "$RESOLVED_BLOCKLISTS"
+    if [ -s "$RESOLVED_BLOCKLISTS" ]; then
+        blocked_count=$(wc -l < "$RESOLVED_BLOCKLISTS")
+        # Filter out blocked tests only if no tests were explicitly selected
+        if [ -z "$CONFORMANCE_TESTS" ]; then
```
cqs21: `CONFORMANCE_BLOCKLIST_MODE=apply` + `CONFORMANCE_TESTS=...` does **not** apply blocklists in kselftest, contradicting the documented `apply` = "always apply". Remove the inner guard and let `should_apply_blocklists` be the single source of truth.

test/initramfs/src/conformance/gvisor/run_gvisor_test.sh
```diff
@@ -7,24 +7,51 @@ TEST_TMP_DIR=${CONFORMANCE_TEST_WORKDIR:-/tmp}
...
-    if [ -f $BLOCKLIST_DIR/$1 ]; then
-        BLOCK=$(grep -v '^#' $BLOCKLIST_DIR/$1 | tr '\n' ':')
-    else
-        BLOCK=""
+CONFORMANCE_BLOCKLIST_MODE=${CONFORMANCE_BLOCKLIST_MODE:-auto}
+CONFORMANCE_TESTS=${CONFORMANCE_TESTS:-}
+GVISOR_GTEST_FILTER=${GVISOR_GTEST_FILTER:-}
+
+# Determine whether to apply blocklists
+should_apply_blocklists() {
```
cqs21: The identical function is duplicated in the gVisor, kselftest, and xfstests runners.  The fix could be:
```sh
# run_conformance_test.sh — define the policy once, decide, then dispatch
CONFORMANCE_BLOCKLIST_MODE=${CONFORMANCE_BLOCKLIST_MODE:-auto}
CONFORMANCE_TESTS=${CONFORMANCE_TESTS:-}

should_apply_blocklists() {
    case "$CONFORMANCE_BLOCKLIST_MODE" in
        apply)  return 0 ;;
        ignore) return 1 ;;
        auto|*) [ -z "$CONFORMANCE_TESTS" ] ;;
    esac
}

if should_apply_blocklists; then
    export CONFORMANCE_APPLY_BLOCKLISTS=1
else
    export CONFORMANCE_APPLY_BLOCKLISTS=0
fi
# ... existing per-suite dispatch, unchanged ...
```

test/initramfs/src/conformance/ltp/run_ltp_test.sh
```diff
@@ -7,10 +7,88 @@ TEST_TMP_DIR=${CONFORMANCE_TEST_WORKDIR:-/tmp}
...
+            echo "Executing: CREATE_ENTRIES=1 $LTP_DIR/runltp -f \"$TEMP_RUNTEST_FILE\" -Q -p -d $TEST_TMP_DIR -l $LOG_FILE"
+            CREATE_ENTRIES=1 $LTP_DIR/runltp -f "$TEMP_RUNTEST_FILE" -Q -p -d $TEST_TMP_DIR -l $LOG_FILE
+        else
+            echo "Error: No valid tests found to run"
+            RESULT=1
+        fi
+    else
+        # Single test or pattern
+        echo "Executing: CREATE_ENTRIES=1 $LTP_DIR/runltp -f syscalls -Q -p -d $TEST_TMP_DIR -l $LOG_FILE -s \"$CONFORMANCE_TESTS\""
+        CREATE_ENTRIES=1 $LTP_DIR/runltp -f syscalls -Q -p -d $TEST_TMP_DIR -l $LOG_FILE -s "$CONFORMANCE_TESTS"
```
cqs21: A `CONFORMANCE_TESTS` value without a comma takes a different code path that calls `runltp ... -s "$CONFORMANCE_TESTS"`, while a comma-containing value builds a temp runtest file. Prefer always building the temp runtest file (exact match) for one mental model.

cqs21 (commented):
Thanks for the rework — this is a big step in the right direction, gVisor and xfstests implement the `auto`/`apply`/`ignore` table correctly. The remaining problems are in LTP and kselftest. Especially, **the LTP packaging-time blocklist concern is not addressed**.

This is the specific risk I flagged in the original note, and it is still open. The LTP Makefile filters at packaging time by
```make
@if [ "$(CONFORMANCE_TEST_WORKDIR)" = "/ext2" ]; then \
    grep -vxF -f $(EXT2_BLOCKLIST) $@/all.txt > $@/filtered.txt; \
elif [ "$(CONFORMANCE_TEST_WORKDIR)" = "/exfat" ]; then \
    grep -vxF -f $(EXFAT_BLOCKLIST) $@/all.txt > $@/filtered.txt; \
...
@while read -r syscall binary params; do \
    if grep -q "^$$syscall$$" $@/filtered.txt; then \
        ... cp binary ...; echo "$$syscall $$binary $$params" >> $@/runtest/syscalls; \
```

Only non-blocklisted binaries are copied into the image, and only their lines are written into `runtest/syscalls`. The runtime script then greps the *packaged* list.

So if a developer wants to run a test that is on the blocklist, its binary was never packaged. The runner would print `Warning: Test 'xxx' not found` and silently does **not** run it `CONFORMANCE_BLOCKLIST_MODE=ignore` cannot recover this, because it is a runtime knob and the binary is already gone.

test/initramfs/src/conformance/ltp/run_ltp_test.sh
co63oc: 已修改

test/initramfs/src/conformance/kselftest/run_kselftest.sh
co63oc: 已去掉判断

test/initramfs/src/conformance/gvisor/run_gvisor_test.sh
co63oc: 已修改 `should_apply_blocklists` 到 `run_conformance_test.sh`

test/initramfs/src/conformance/ltp/run_ltp_test.sh
co63oc: 已合并单例和多例处理，都写入临时文件

co63oc:
> Thanks for the rework — this is a big step in the right direction, gVisor and xfstests implement the `auto`/`apply`/`ignore` table correctly. The remaining problems are in LTP and kselftest. Especially, **the LTP packaging-time blocklist concern is not addressed**.
> 
> This is the specific risk I flagged in the original note, and it is still open. The LTP Makefile filters at packaging time by
> 
> ```makefile
> @if [ "$(CONFORMANCE_TEST_WORKDIR)" = "/ext2" ]; then \
>     grep -vxF -f $(EXT2_BLOCKLIST) $@/all.txt > $@/filtered.txt; \
> elif [ "$(CONFORMANCE_TEST_WORKDIR)" = "/exfat" ]; then \
>     grep -vxF -f $(EXFAT_BLOCKLIST) $@/all.txt > $@/filtered.txt; \
> ...
> @while read -r syscall binary params; do \
>     if grep -q "^$$syscall$$" $@/filtered.txt; then \
>         ... cp binary ...; echo "$$syscall $$binary $$params" >> $@/runtest/syscalls; \
> ```
> 
> Only non-blocklisted binaries are copied into the image, and only their lines are written into `runtest/syscalls`. The runtime script then greps the _packaged_ list.
> 
> So if a developer wants to run a test that is on the blocklist, its binary was never packaged. The runner would print `Warning: Test 'xxx' not found` and silently does **not** run it `CONFORMANCE_BLOCKLIST_MODE=ignore` cannot recover this, because it is a runtime knob and the binary is already gone.

已修改 `Makefile`，实现 `run_conformance_test.sh` 中相同逻辑 @cqs21

cqs21:
Thanks for the update, but I think the core issue is still not addressed: **the LTP packaging-time blocklist concern**. The runtime `CONFORMANCE_BLOCKLIST_MODE` knob is currently broken for LTP because the binaries are stripped out before the runtime even starts.

To illustrate the problem, you can test this locally:
```shell
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=ltp CONFORMANCE_TESTS=rename01
```
**Expected behavior**: The test runner executes `rename01` (and either passes or fails).
**Actual behavior**: The runner prints **Warning: Test 'rename01' not found** and silently skips it, because the `rename01` binary was never packaged into the image.

To fix this, the `CONFORMANCE_TESTS` and `CONFORMANCE_BLOCKLIST_MODE` variables need to penetrate the Nix build boundary so the packaging Makefile can respect them. Specifically:
1. In `test/initramfs/Makefile`: Pass `CONFORMANCE_TESTS` and `CONFORMANCE_BLOCKLIST_MODE` as arguments to the `nix-build` invocation.
2. In `test/initramfs/nix/conformance/default.nix`: Export these two variables in `buildCommand` so they are available as environment variables during `make`.
3. In `test/initramfs/src/conformance/ltp/Makefile`:  Use these environment variables to determine which binaries to copy into the initramfs. To avoid bloating the image, we should not copy all binaries; instead, only copy the required ones.

co63oc:
已修改在文件中设置环境变量  
`test/initramfs/Makefile`, 
`test/initramfs/nix/conformance/default.nix`，
`test/initramfs/src/conformance/ltp/Makefile` 按 CONFORMANCE_TESTS 复制文件
```
		if [ -n "$(CONFORMANCE_TESTS)" ]; then \
			echo "$(CONFORMANCE_TESTS)" | tr ',' '\n' > $@/filtered.txt; \
		else \
			cp -f $@/all.txt $@/filtered.txt; \
```

测试  
```
make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=ltp CONFORMANCE_TESTS=rename01
```
image

CI已完成
@cqs21

Makefile
```diff
@@ -52,6 +52,20 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
 # - `kselftest` treats each entry as a blocklist file relative to its runner
 #   directory, and appends that file directly.
 EXTRA_BLOCKLISTS ?= ""
+
+# Comma-separated tests to run within the selected conformance suite.
```
cqs21, line 55: The token format differs by suite and is not obvious, add a per-suite example to the comment.
```suggestion
# Comma-separated tests to run within the selected conformance suite.
# - gvisor:    a test *binary* name, e.g. `epoll_test` (optionally narrow the
#              cases inside it with `GVISOR_GTEST_FILTER`).
# - kselftest: a `<collection>:<case>` entry, e.g. `timers:posix_timers`.
# - ltp:       a syscall testcase id, e.g. `rename01`.
# - xfstests:  a test id, e.g. `generic/001`.
CONFORMANCE_TESTS ?= ""
```

test/initramfs/src/conformance/ltp/run_ltp_test.sh
```diff
@@ -7,10 +7,78 @@ TEST_TMP_DIR=${CONFORMANCE_TEST_WORKDIR:-/tmp}
...
+            local test_line=""
+            if grep -q "^$test_name " "$LTP_DIR/runtest/syscalls"; then
+                test_line=$(grep "^$test_name " "$LTP_DIR/runtest/syscalls")
+            elif grep -q "^$test_name$" "$LTP_DIR/runtest/syscalls"; then
+                test_line=$(grep "^$test_name$" "$LTP_DIR/runtest/syscalls")
+            fi
+            
+            if [ -n "$test_line" ]; then
+                if [ -n "$content" ]; then
+                    content="$content"$'\n'"$test_line"
```
cqs21: `run_ltp_test.sh` uses `content=...$'\n'...` (ANSI-C quoting) and `echo -e "$content"`. Neither is POSIX. It works today only because the initramfs `/bin/sh` is BusyBox `ash` with bash-compat, but running the same script under `dash` produces a corrupted file. Suggested change:
```diff
     local temp_file=$(mktemp /tmp/ltp_runtest.XXXXXX)
-    local content=""
     local old_IFS="$IFS"

     # Split the tests string using IFS and set as positional parameters
     IFS=','
     set -- $tests
     IFS="$old_IFS"

     # Process each test in the list
     for test_name; do
         if [ -n "$test_name" ]; then
             # Find the test in the original syscalls file
             local test_line=""
             if grep -q "^$test_name " "$LTP_DIR/runtest/syscalls"; then
                 test_line=$(grep "^$test_name " "$LTP_DIR/runtest/syscalls")
             elif grep -q "^$test_name$" "$LTP_DIR/runtest/syscalls"; then
                 test_line=$(grep "^$test_name$" "$LTP_DIR/runtest/syscalls")
             fi

             if [ -n "$test_line" ]; then
-                if [ -n "$content" ]; then
-                    content="$content"$'\n'"$test_line"
-                else
-                    content="$test_line"
-                fi
+                printf '%s\n' "$test_line" >> "$temp_file"
             else
                 echo "Warning: Test '$test_name' not found in syscalls file"
             fi
         fi
     done

-    echo -e "$content" > "$temp_file"
     TEMP_RUNTEST_FILE="$temp_file"
```

cqs21 (commented):
Thanks for the work on this! It's working as intended now. Just two minor comments left, otherwise LGTM.

test/initramfs/src/conformance/ltp/run_ltp_test.sh
co63oc: 已修改

Makefile
```diff
@@ -52,6 +52,26 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
+# - gvisor:    a test *binary* name, e.g. `epoll_test` (optionally narrow the
+#              cases inside it with `GVISOR_GTEST_FILTER`).
+# - kselftest: a `<collection>:<case>` entry, e.g. `timers:posix_timers`.
+# - ltp:       a syscall testcase id, e.g. `rename01`.
+# - xfstests:  a test id, e.g. `generic/001`.
+CONFORMANCE_TESTS ?= ""
+CONFORMANCE_TESTS ?= ""
```
cqs21: The doc edit left `CONFORMANCE_TESTS ?= ""` twice — please drop one.

test/initramfs/src/conformance/ltp/Makefile
```diff
@@ -19,12 +19,30 @@ $(TARGET_DIR): $(RUN_BASH) $(ALL_TESTS) $(EXT2_BLOCKLIST) $(EXFAT_BLOCKLIST)
...
+		should_apply=1; \
+	elif [ "$(CONFORMANCE_BLOCKLIST_MODE)" = "ignore" ]; then \
+		should_apply=0; \
+	elif [ -z "$(CONFORMANCE_TESTS)" ]; then \
+		should_apply=1; \
 	else \
-		cp -f $@/all.txt $@/filtered.txt; \
+		should_apply=0; \
+	fi; \
+	if [ "$$should_apply" = "1" ]; then \
```
cqs21: On a second thought: `CONFORMANCE_TESTS` is consumed at build time by `ltp/Makefile`, which already packs only the selected tests into `runtest/syscalls`, and changing it forces an initramfs rebuild. So build-time and run-time values are always the same — the runtime `create_temp_runtest` just re-derives what the build already produced. Perhaps we should **reverte all the changes of `run_ltp_test.sh`**.

Besides, one adjustment is also required in `ltp/Makefile`: its `apply` branch currently ignores `CONFORMANCE_TESTS`, so `CONFORMANCE_BLOCKLIST_MODE=apply CONFORMANCE_TESTS=rename01` would run the whole suite instead of just `rename01`. Fix: narrow to the selection first, then subtract the blocklist when `should_apply=1`.
```diff
diff --git a/test/initramfs/src/conformance/ltp/Makefile b/test/initramfs/src/conformance/ltp/Makefile
index 1abd84d1d..24bd2da6e 100644
--- a/test/initramfs/src/conformance/ltp/Makefile
+++ b/test/initramfs/src/conformance/ltp/Makefile
@@ -18,10 +18,27 @@ $(TARGET_DIR): $(RUN_BASH) $(ALL_TESTS) $(EXT2_BLOCKLIST) $(EXFAT_BLOCKLIST)
 	@# Prepare tests dir for test binaries
 	@mkdir -p $@/testcases/bin
 	@mkdir -p $@/runtest
-	@awk '!/^#/ && NF' $(ALL_TESTS) > $@/all.txt
-	@if [ "$(CONFORMANCE_TEST_WORKDIR)" = "/ext2" ]; then \
+	@# Use the explicit selection verbatim (it may name tests disabled in
+	@# all.txt), otherwise fall back to the default enabled set.
+	@if [ -n "$(CONFORMANCE_TESTS)" ]; then \
+		echo "$(CONFORMANCE_TESTS)" | tr ',' '\n' | grep -v '^$$' > $@/all.txt; \
+	else \
+		awk '!/^#/ && NF' $(ALL_TESTS) > $@/all.txt; \
+	fi
+	@# Then subtract the blocklist when it should be applied
+	@# (same should_apply logic as run_conformance_test.sh).
+	@if [ "$(CONFORMANCE_BLOCKLIST_MODE)" = "apply" ]; then \
+		should_apply=1; \
+	elif [ "$(CONFORMANCE_BLOCKLIST_MODE)" = "ignore" ]; then \
+		should_apply=0; \
+	elif [ -z "$(CONFORMANCE_TESTS)" ]; then \
+		should_apply=1; \
+	else \
+		should_apply=0; \
+	fi; \
+	if [ "$$should_apply" = "1" ] && [ "$(CONFORMANCE_TEST_WORKDIR)" = "/ext2" ]; then \
 		grep -vxF -f $(EXT2_BLOCKLIST) $@/all.txt > $@/filtered.txt; \
-	elif [ "$(CONFORMANCE_TEST_WORKDIR)" = "/exfat" ]; then \
+	elif [ "$$should_apply" = "1" ] && [ "$(CONFORMANCE_TEST_WORKDIR)" = "/exfat" ]; then \
 		grep -vxF -f $(EXFAT_BLOCKLIST) $@/all.txt > $@/filtered.txt; \
 	else \
 		cp -f $@/all.txt $@/filtered.txt; \
```

cqs21 (commented):
Thanks for the updates! A couple of follow-ups:

test/initramfs/src/conformance/ltp/Makefile
co63oc: run_ltp_test.sh 修改已撤销

ltp/Makefile 这里有问题，比如rename01在 all.txt 中已注释，测试可能不是依赖 all.txt
```
 make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=ltp CONFORMANCE_TESTS=rename01
```
运行会找不到syscalls
image

cqs21: You're right. When `CONFORMANCE_TESTS` is set, use that list directly; don't intersect with `all.txt`. Just remove `grep -xF -f $@/sel.txt $@/all.txt > $@/selected.txt || true; \` to fix it. See the edited diff above.

co63oc: 已修改

co63oc:
@cqs21 还有什么需要修改的没有

cqs21:
LGTM! Thanks for your continuous efforts! See if @tatetian has more comments?

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
 # - `kselftest` treats each entry as a blocklist file relative to its runner
 #   directory, and appends that file directly.
 EXTRA_BLOCKLISTS ?= ""
+
```
tatetian: Remove these blank lines; they break the established convention:

```Makefile
# XXX options.
VAR_1= val1
...
# No blank lines in between "XXX options" and "End of XXX options"
VAR_N= valN
# End of XXX options.
```

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
+# Comma-separated tests to run within the selected conformance suite.
+# - gvisor:    a test *binary* name, e.g. `epoll_test` (optionally narrow the
+#              cases inside it with `GVISOR_GTEST_FILTER`).
+# - kselftest: a `<collection>:<case>` entry, e.g. `timers:posix_timers`.
+# - ltp:       a syscall testcase id, e.g. `rename01`.
+# - xfstests:  a test id, e.g. `generic/001`.
+CONFORMANCE_TESTS ?= ""
```
tatetian: Rename `CONFORMANCE_TESTS` to `CONFORMANCE_TEST_SELECTOR`. Use the term "test selector" consistently and define the meaning of "test selector" for different conformance test suite.

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
+
+# Controls whether conformance blocklists are applied.
+# Values:
+# - auto: apply blocklists for full-suite runs, ignore blocklists for explicit selections.
+# - apply: always apply blocklists.
+# - ignore: never apply blocklists.
+CONFORMANCE_BLOCKLIST_MODE ?= auto
```
tatetian: Rename `CONFORMANCE_BLOCKLIST_MODE` to `CONFORMANCE_TEST_BLOCKLIST_MODE` so that the prefix is consistent with other related variables. Rename "apply" to "on" and "ignore" to "off"; the new names are more concise and even more clear.

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
+
+# gVisor-only positive gtest filter, used inside one selected gVisor test binary.
+GVISOR_GTEST_FILTER ?= ""
```
tatetian: Rename this one to `CONFORMANCE_TEST_GVISOR_FILTER"

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
+CONFORMANCE_TESTS ?= ""
+
+# Controls whether conformance blocklists are applied.
+# Values:
+# - auto: apply blocklists for full-suite runs, ignore blocklists for explicit selections.
+# - apply: always apply blocklists.
+# - ignore: never apply blocklists.
+CONFORMANCE_BLOCKLIST_MODE ?= auto
+
+# gVisor-only positive gtest filter, used inside one selected gVisor test binary.
```
tatetian, line 67: Provide an example. It is unclear here the format of gtest filter.

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
...
+
+# Comma-separated tests to run within the selected conformance suite.
+# - gvisor:    a test *binary* name, e.g. `epoll_test` (optionally narrow the
+#              cases inside it with `GVISOR_GTEST_FILTER`).
+# - kselftest: a `<collection>:<case>` entry, e.g. `timers:posix_timers`.
+# - ltp:       a syscall testcase id, e.g. `rename01`.
+# - xfstests:  a test id, e.g. `generic/001`.
+CONFORMANCE_TESTS ?= ""
+
+# Controls whether conformance blocklists are applied.
```
tatetian: It is under-specified what is the expected behaviors when both `CONFORMANCE_BLOCKLIST_MODE` is `apply` and `CONFORMANCE_TESTS` is non-empty. Make the effect of this combination explicit.

In fact, I found that this combination is (mostly) useless: now that you already select test cases to run, why still need to respect the blocked list? It looks that the only thing to respect an non-empty `CONFORMANCE_TESTS` and `CONFORMANCE_BLOCKLIST_MODE=apply` is for gvisor because `CONFORMANCE_TESTS` would be too corse-grained. Then, why not merge `GVISOR_GTEST_FILTER` into `CONFORMANCE_TESTS`? Let the gvisor test runner script to parse `CONFORMANCE_TESTS` into two parts: a list of gvisor binaries and `GVISOR_GTEST_FILTER`.

Therefore, I think that we do not need `CONFORMANCE_BLOCKLIST_MODE` at all. Just add `CONFORMANCE_TESTS`: if it is specified, run the selected test cases, regardless of the block list; otherwise, run all test cases except those in the block list. In other words, `CONFORMANCE_TESTS` take a higher priority than the blocklist. This results in a simpler interface and implementation.

Makefile
```diff
@@ -52,6 +52,25 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
 # - `kselftest` treats each entry as a blocklist file relative to its runner
 #   directory, and appends that file directly.
 EXTRA_BLOCKLISTS ?= ""
```
tatetian: For consistency, I suggest renaming `EXTRA_BLOCKLISTS` to `CONFORMANCE_TEST_EXTRA_BLOCKLISTS`.

tatetian (commented):
I have a major concern about the semantics of the new Makefile variables. See the comments below. The rest are minor issues.

By the way, please rebase to the latest main branch.

Makefile
co63oc: 因为这些修改是给以后开发测试使用的，比赛结束后我就不用了，所以建议问下其他开发人员是否要改动

co63oc:
image
CI测试错误，看上去和修改没有关联，kernel并没有修改

Makefile
tatetian: > 建议问下其他开发人员是否要改动

What is your attitude and opinion? It is YOUR proposal, so it is YOUR responsibility to defend YOUR idea. You can't ask ME to ask OTHER people's opinions.

co63oc: Based on the current status of other submitted PRs, fixing the unit test is not the main issue. The modification of the unit test relies on the design modifications of other existing systems, and there is no need to design too many test cases for testing. Therefore, this approach is also feasible.

tatetian: My main concern is:

> I think that we do not need `CONFORMANCE_BLOCKLIST_MODE` at all.

and I don't think that I get any answer on it:

> fixing the unit test is not the main issue.

Why are you talking about "unit test"? I am confused.

tatetian:
> CI测试错误，看上去和修改没有关联，kernel并没有修改

Let's not draw a conclusion this soon. This PR changed the test harness. What if it was caused by the changes to the conformance test infra? I didn't see any such errors on the main branch. Let me trigger another run of the CI to see if the failures are random.

tatetian:
@boterinas rerun

To see if the last three failed CI workflows still would still fail:

image

Makefile
co63oc: The modification to CONFORMANCE_BLOCKLIST_MODE has been made, but there might be another reason behind it. In Rust, bool is rarely used in parameter design, and CONFORMANCE_BLOCKLIST_MODE has a value similar to bool and is passed between multiple modules, which is not quite consistent with Rust style. The switching of CONFORMANCE_BLOCKLIST_MODE is mainly used to verify whether the tests in the blocklist have passed. This can also be achieved by modifying the configuration, and the impact is not significant. CONFORMANCE_TEST_GVISOR_FILTER will not be merged into CONFORMANCE_TEST_SELECTOR for now, because when used, CONFORMANCE_TEST_SELECTOR and CONFORMANCE_TEST_GVISOR_FILTER are copied separately. If they are combined and parsed, it adds some steps @tatetian

co63oc:
> > CI测试错误，看上去和修改没有关联，kernel并没有修改
> 
> Let's not draw a conclusion this soon. This PR changed the test harness. What if it was caused by the changes to the conformance test infra? I didn't see any such errors on the main branch. Let me trigger another run of the CI to see if the failures are random.

It's a code issue. I'll look into it and make some modifications.
The variable names in .github/actions/test/action.yml need to be modified.
Now CI has been completed.

test/initramfs/Makefile
```diff
@@ -77,6 +78,7 @@ $(INITRAMFS_IMAGE): $(INITRAMFS)
 		--arg enableRegressionTest $(ENABLE_REGRESSION_TEST) \
 		--argstr conformanceTestSuite $(CONFORMANCE_TEST_SUITE) \
 		--argstr conformanceTestWorkDir $(CONFORMANCE_TEST_WORKDIR) \
+		--argstr testSelector "$(CONFORMANCE_TEST_SELECTOR)" \
```
tatetian: Rename this parameter to `conformanceTestSelector` to be consistent with other parameters that start with `conformanceTest`.

test/initramfs/src/conformance/run_conformance_test.sh
```diff
@@ -5,11 +5,22 @@
...
 CONFORMANCE_TEST_SUITE=${CONFORMANCE_TEST_SUITE:-ltp}
+CONFORMANCE_TEST_SELECTOR=${CONFORMANCE_TEST_SELECTOR:-}
 LTP_DIR=/opt/ltp
 GVISOR_DIR=/opt/gvisor
 KSELFTEST_DIR=/opt/kselftest
 XFSTESTS_DIR=/opt/xfstests
 
+should_apply_blocklists() {
+    [ -z "$CONFORMANCE_TEST_SELECTOR" ]
+}
```
cqs21: Now that `CONFORMANCE_BLOCKLIST_MODE` has been removed, this helper function is no longer necessary. Since it's only a single condition used in one place, it can be inlined directly:

```
if [ -z "$CONFORMANCE_TEST_SELECTOR" ]; then
    export CONFORMANCE_APPLY_BLOCKLISTS=1
else
    export CONFORMANCE_APPLY_BLOCKLISTS=0
fi
```

test/initramfs/src/conformance/xfstests/run_xfstests.sh
```diff
@@ -61,7 +61,13 @@ while [ $# -gt 0 ]; do
   esac
 done
 
-if [ -n "$RUNLIST_FILE" ]; then
+# If CONFORMANCE_TEST_SELECTOR is set, use it instead of the runlist or other args
```
cqs21:
```suggestion
# If `CONFORMANCE_TEST_SELECTOR` is set, use it instead of the runlist or other args
```

cqs21 (commented):
Two minor comments.

test/initramfs/src/conformance/xfstests/run_xfstests.sh
```diff
@@ -61,7 +61,13 @@ while [ $# -gt 0 ]; do
...
+if [ -n "$CONFORMANCE_TEST_SELECTOR" ]; then
+    # Convert comma-separated to space-separated
+    TEST_ARGS=$(echo "$CONFORMANCE_TEST_SELECTOR" | tr ',' ' ')
+fi
+
+if [ -z "$TEST_ARGS" ] && [ -n "$RUNLIST_FILE" ]; then
```
tatetian: **Improve the readability of the code.** There are three `if ...; then ... fi` code blocks. They are exclusive to each other, yet the code is not structured to reflect this point. One has to dive into each code block, decode the intern logic, and understand why they are exclusive to each other. Refactor the code to make it clear that (1) they are three exclusive branches, and (2) they also serve the purpose of setting `$TEST_ARGS`.

test/initramfs/src/conformance/run_conformance_test.sh
```diff
@@ -5,11 +5,22 @@
...
+
+if should_apply_blocklists; then
+    export CONFORMANCE_APPLY_BLOCKLISTS=1
```
tatetian: **Removing this variable.** It is useless: `CONFORMANCE_APPLY_BLOCKLISTS` is same as `CONFORMANCE_TEST_SELECTOR` being non-empty. So why not check `CONFORMANCE_TEST_SELECTOR` directly?

Makefile
```diff
@@ -51,7 +51,17 @@ CONFORMANCE_TEST_WORKDIR ?= /tmp
 #   and loads a per-test blocklist file from that directory.
 # - `kselftest` treats each entry as a blocklist file relative to its runner
 #   directory, and appends that file directly.
-EXTRA_BLOCKLISTS ?= ""
+CONFORMANCE_TEST_EXTRA_BLOCKLISTS ?= ""
+# Comma-separated tests to run within the selected conformance suite.
```
tatetian, line 55: Add the following comments to make the behaviors of this parameter explicit:

```Makefile
# Comma-separated tests to run within the selected conformance suite.
#
# This variable allows one to select one or multiple tests to run.
# When this environment variable is present,
# the blocklist of a conformance test will be ignored.
#
# - gvisor:    a test *binary* name, e.g. `epoll_test` (optionally narrow the
#              cases inside it with `CONFORMANCE_TEST_GVISOR_FILTER`).
# - kselftest: a `<collection>:<case>` entry, e.g. `timers:posix_timers`.
# - ltp:       a syscall testcase id, e.g. `rename01`.
# - xfstests:  a test id, e.g. `generic/001`.
```

.github/actions/test/action.yml
```diff
@@ -98,7 +98,7 @@ runs:
         [[ -n "${{ inputs.netdev }}" ]] && CMD+=" NETDEV=${{ inputs.netdev }}"
         [[ -n "${{ inputs.scheme }}" ]] && CMD+=" SCHEME=${{ inputs.scheme }}"
         [[ -n "${{ inputs.arch }}" ]] && CMD+=" TARGET_ARCH=${{ inputs.arch }}"
-        [[ -n "${{ inputs.extra_blocklists }}" ]] && CMD+=" EXTRA_BLOCKLISTS=${{ inputs.extra_blocklists }}"
+        [[ -n "${{ inputs.extra_blocklists }}" ]] && CMD+=" CONFORMANCE_TEST_EXTRA_BLOCKLISTS=${{ inputs.extra_blocklists }}"
         [[ -n "${{ inputs.conformance_test_suite }}" ]] && CMD+=" CONFORMANCE_TEST_SUITE=${{ inputs.conformance_test_suite }}"
```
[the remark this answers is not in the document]
co63oc: CI 不使用 CONFORMANCE_TEST_SELECTOR

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -31,54 +31,68 @@ else
...
-cat "$DEFAULT_BLOCKLISTS" > "$COMBINED_BLOCKLISTS"
-for extra_file in $EXTRA_BLOCKLISTS ; do
-    extra_blocklists="$KSELFTEST_DIR/$extra_file"
-    if [ -r "$extra_blocklists" ]; then
-        printf '\n' >> "$COMBINED_BLOCKLISTS"
-        cat "$extra_blocklists" >> "$COMBINED_BLOCKLISTS"
-    else
-        echo "Warning: extra blocklist not found: $extra_blocklists" >&2
+# Process blocklists only if needed
+if [ "$CONFORMANCE_APPLY_BLOCKLISTS" = "1" ]; then
```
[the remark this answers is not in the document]
co63oc: CONFORMANCE_APPLY_BLOCKLISTS 已取消

co63oc:
@tatetian 已修改CI完成

test/initramfs/src/conformance/kselftest/run_kselftest.sh
```diff
@@ -31,54 +31,68 @@ else
...
-cat "$DEFAULT_BLOCKLISTS" > "$COMBINED_BLOCKLISTS"
-for extra_file in $EXTRA_BLOCKLISTS ; do
-    extra_blocklists="$KSELFTEST_DIR/$extra_file"
-    if [ -r "$extra_blocklists" ]; then
-        printf '\n' >> "$COMBINED_BLOCKLISTS"
-        cat "$extra_blocklists" >> "$COMBINED_BLOCKLISTS"
-    else
-        echo "Warning: extra blocklist not found: $extra_blocklists" >&2
+# Process blocklists only if needed
+if [ -z "$CONFORMANCE_TEST_SELECTOR" ]; then
```
tatetian, line 46: Defend your code:

```
if [ -n "$CONFORMANCE_TEST_SELECTOR" ]; then
    # Do A
    ...
else
    # Do B
    ...
fi

if [ -z "$CONFORMANCE_TEST_SELECTOR" ]; then
    # Do C
    ...
fi
```

Why would anyone--or any AI--write code like this? It should be as simple as:

```
if [ -n "$CONFORMANCE_TEST_SELECTOR" ]; then
    # Do A
    ...
else
    # Do B
    ...
    # Do C
    ...
fi
```

Did you take any time to do serious self-review? I have made it a clear in another comment for you that "AI is welcome, but human is responsible". Are your really responsible for your code?

We tried to provide constructive feedbacks for you in hope that you can do meaningful and quality contributions and learn something from it. But unfortunately, I have lost confidence in your attitude or/and ability.

tatetian (commented):
I refuse to review this PR anymore because I lost faith in the attitude and/or ability of the author.

PR #3399, opened 2026-06-16
Make `VmMapping` owns `FileLike`

cchanging:
Currently, `VmMapping` only holds a `Path` and not a file, which means some per-file state cannot be associated with `VmMapping`. For example:

1. For Virtiofs, the `fh` used to communicate with the backend is per-file state and is stored in the file. When implementing mappings for it, this can lead to a situation where the mapping is still alive but the file has already been closed, leaving the mapping unable to access backend data.
2. For the readahead mechanism, its state is also a per-file property. If we want to trigger readahead on page faults as well, we cannot reuse that per-file state.

Besides, Linux mappings also hold a file.

Therefore, this PR aims to align the semantics at this layer with Linux, so as to address several potential issues such as those mentioned above.

---

This PR also includes some preparatory work. Although these changes are relatively independent, they are fairly trivial, so I grouped them together in this PR:

1. It fixes the incorrect behavior where `execve` did not actually open the file.
2. Once `execve` starts opening files as it should, some previously hidden issues become visible, mainly around the timing of `notify close` and `release range lock`. This PR fixes those as well.

### DIFF
--- a/kernel/src/fs/vfs/notify/mod.rs
+++ b/kernel/src/fs/vfs/notify/mod.rs
@@ -7,7 +7,7 @@ use bitflags::bitflags;
 
 use crate::{
     fs::{
-        file::{AccessMode, FileLike, InodeType, StatusFlags},
+        file::{AccessMode, FileCommon, FileLike, InodeType, StatusFlags},
         vfs::path::Path,
     },
     prelude::*,
@@ -372,18 +372,18 @@ pub fn on_create(file_path: &Path, name: impl FnOnce() -> String) {
 
 /// Notifies that a file was opened.
 pub fn on_open(file: &Arc<dyn FileLike>) {
-    let Some(path) = notifiable_path(file) else {
+    let Some(path) = notifiable_path(file.common()) else {
         return;
     };
     notify_parent(path, FsEvents::OPEN);
 }
 
 /// Notifies that a file was closed.
-pub fn on_close(file: &Arc<dyn FileLike>) {
-    let Some(path) = notifiable_path(file) else {
+pub fn on_close(common: &FileCommon) {
+    let Some(path) = notifiable_path(common) else {
         return;
     };
-    let events = match file.access_mode() {
+    let events = match common.access_mode() {
         AccessMode::O_RDONLY => FsEvents::CLOSE_NOWRITE,
         _ => FsEvents::CLOSE_WRITE,
     };
@@ -395,11 +395,11 @@ pub fn on_close(file: &Arc<dyn FileLike>) {
 /// `O_PATH` file descriptors carry `FMODE_NONOTIFY` in Linux's `f_mode` and
 /// thus suppress all fsnotify events, so they yield `None`. `None` is also
 /// returned when the filesystem has no event subscribers.
-fn notifiable_path(file: &Arc<dyn FileLike>) -> Option<&Path> {
-    if file.status_flags().contains(StatusFlags::O_PATH) {
+fn notifiable_path(common: &FileCommon) -> Option<&Path> {
+    if common.status_flags().contains(StatusFlags::O_PATH) {
         return None;
     }
-    let path = file.path();
+    let path = common.path();
     if !path.fs().fs_event_subscriber_stats().has_any_subscribers() {
         return None;
     }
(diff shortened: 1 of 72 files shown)


kernel/src/process/posix_thread/exit.rs
```diff
@@ -114,7 +114,13 @@ fn exit_internal(
 
     // Drop fields in `ThreadLocal`.
     drop_after!(thread_local.vmar().borrow_mut().take());
-    drop_after!(thread_local.borrow_file_table_mut().remove());
+    drop_after!({
+        let file_table = thread_local.borrow_file_table_mut().remove();
+        if is_last_thread && let Some(file_table) = file_table.as_ref() {
+            file_table.read().release_all_range_locks();
```
lrh2000: I don't believe this is correct. What about closing a file? What about closing a file that has been `dup()`ed before? What about closing a file that has been sent to the current process twice (via two `recvmsg` calls)?

In all of the above situations, when should the lock be released (if there is lock that has not been released manually)?

Also, the file table may be shared by other processes, which may close the file concurrently. The code here does not seem to make sense.

It might be better to put it in a separate code block instead of mixing it in the "Drop fields" section. But if it can be fixed, the final form may differ significantly from the current one.

cchanging: > What about closing a file? 

Currently we will invoke `release_range_lock` when closing a file:

```rust
impl FileTable
    pub fn close_file(&mut self, fd: FileDesc) -> Option<Arc<dyn FileLike>> {
        let removed_entry = self.table.remove(fd.into())?;
        // POSIX record locks are process-associated and Linux drops them when any fd for the inode is
        // closed by that process, even if duplicated descriptors still exist.
        //
        // Reference: <https://man7.org/linux/man-pages/man2/fcntl_locking.2.html>
        if let Ok(inode_handle) = removed_entry.file.as_inode_handle_or_err() {
            inode_handle.release_range_locks();
        }
        Some(removed_entry.file)
    }
}
```

> What about closing a file that has been dup()ed before?

We have a test `close_dup_fd_releases_locks`. The intended semantics here should be consistent with the comments in `close_file`.

> What about closing a file that has been sent to the current process twice (via two recvmsg calls)?

I guess this is probably similar to the `dup` case? This operation should also fit the description of “POSIX record locks are process-associated and Linux drops them when any fd for the inode is closed by that process”

> Also, the file table may be shared by other processes, which may close the file concurrently. The code here does not seem to make sense.

This is indeed a problem. Linux seems to release it [when the last `FileTable` reference is dropped](https://elixir.bootlin.com/linux/v6.16.5/source/fs/file.c#L506).  I found that the lock owner model here isn’t fully consistent with Linux. When releasing, we probably shouldn’t get it from `current`. In Linux, processes that share the same file table also share the same lock owner.

This may require some additional refactoring. If the changes here turn out to be fairly involved, it might make sense to split them out into a separate PR.

kernel/src/fs/vfs/notify/mod.rs
```diff
@@ -372,18 +372,18 @@ pub fn on_create(file_path: &Path, name: impl FnOnce() -> String) {
 
 /// Notifies that a file was opened.
 pub fn on_open(file: &Arc<dyn FileLike>) {
```
StevenJiang1110: For consistency with `on_close`, let `on_open` accept file_common as well.

kernel/src/process/program_loader/elf/load_elf.rs
```diff
@@ -343,12 +348,21 @@ fn map_segment_vmos(
...
             .relocated_addr_of(loadable_phdr.virt_range().start)
             .expect("`calc_total_vaddr_bounds()` should cover all segments");
-        map_segment_vmo(loadable_phdr, elf_file, vmar, map_at)?;
+        map_segment_vmo(loadable_phdr, elf_file, vmar, map_at)
+    };
+
+    for loadable_phdr in other_loadable_phdrs {
+        map_loadable_phdr(loadable_phdr, elf_file.clone())?;
     }
+    map_loadable_phdr(last_loadable_phdr, elf_file)?;
```
StevenJiang1110: I think it may not be worth introducing extra complexity just to avoid cloning `elf_file` one more time. We've already cloned `elf_file` for the loop, so one additional clone seems acceptable — and it helps avoid unnecessary code complexity.

StevenJiang1110 (approved):
LGTM! This PR is in good shape for me. Only two small nits.

cqs21 (approved):

cchanging:
@junyang-zh Could you help review this PR?

junyang-zh (approved):

PR #3400, opened 2026-06-16
Fix Intel TDX section of asterinas-book

VZstless:
As an intern who is working in TDX directly related field, I feel angry about your docs. If the only work for your documentation writer is copy white papers from other vendor you can now stop your work and have a lifelong rest. Archwiki even could have a better way to "copy" these kinds of materials.

Ref: https://github.com/intel/lkvs/blob/main/KVM/qemu/tdx_basic.cfg

### DIFF
--- a/book/src/kernel/intel-tdx.md
+++ b/book/src/kernel/intel-tdx.md
@@ -8,7 +8,6 @@ Intel TDX (Trust Domain Extensions) is a Trusted Execution Environment (TEE) tec
 that enhances VM security
 by creating isolated, hardware-enforced trust domains
 with encrypted memory, secure initialization, and attestation mechanisms.
-For more information about Intel TDX, jump to the last section.
 
 ## Why choose Asterinas for Intel TDX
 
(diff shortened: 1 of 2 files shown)


tatetian:
Thank you for improving the doc.

> As an intern who is working in TDX directly related field, I feel angry about your docs. If the only work for your documentation writer is copy white papers from other vendor you can now stop your work and have a lifelong rest.

Being rude and offensive to prior contributors are not helpful. I won't accept this PR unless these unnecessarily impolite and offensive words are taken back.

The person who contributed the paragraphs you want to delete is an Intel employee. The paragraph was written with good intention to promote Intel TDX by giving the reader a quick introduction to Intel TDX. Arguably, the writing has some room for improvements. And simply deleting them and citing another source is probably better. But this does not warrant being rude against other contributors.

book/src/kernel/intel-tdx.md
```diff
@@ -53,7 +57,26 @@ The following result is an example:
 [   20.507296] tdx: TDX module initialized.
 ```
 
-`TDX module initialized` means TDX module is loaded successfully.
+Third, check the output of `rdmsr` command:
+
+```bash
+rdmsr 0xfe --bitfield 15:15
```
tatetian: Now that checking if the host platform supports Intel TDX or not become non-trivial, I think a script `check-tdx-support.sh` should be added to `tools/` so that the reader of this doc can simply call this script, rather than taking care of all the technical details.

book/src/kernel/intel-tdx.md
```diff
@@ -107,38 +130,7 @@ make gdb_client INTEL_TDX=1
...
-
-Intel TDX ensures data integrity, confidentiality, and authenticity,
-which empowers engineers and tech professionals
-to create and maintain secure systems,
-enhancing trust in virtualized environments.
-
-For more information,
+
+## References
+For more information about Intel Trust Domain Extension,
```
tatetian: I suggest simply removing this section. The second paragraph is the location where Intel TDX is introduced:

> Intel TDX (Trust Domain Extensions) is a Trusted Execution Environment (TEE) technology

Just make "Intel TDX (Trust Domain Extensions)" a hyperlink to an authorative source. One good candidate is as Intel's official documentation.

VZstless:
:triumph:

tatetian:
> I won't accept this PR unless these unnecessarily impolite and offensive words are taken back.

This is where I stand.

VZstless:
DO AS U LIKE
I have no time to fix your so-called value.
farewell.

PR #3402, opened 2026-06-16
Add global lock for mount tree topology

### LINKED ISSUE #3392: Concurrent umount on parent/child mounts causes kernel panic
The VFS mount tree lacks global serialization for structural mutations (mount/umount/move_mount). Concurrent umount operations on a parent and its child mount can cause a kernel panic due to a `Weak::upgrade().unwrap()` on a dangling reference. 

## Reproduction

Compile the program below and run with `SMP=2 ENABLE_KVM=1`.

```
[   114.464] ERROR : Uncaught panic:
        called `Option::unwrap()` on a `None` value
        at /root/asterinas/kernel/src/fs/vfs/path/mod.rs:348
        on CPU 1 by thread Some(Thread { task: (Weak), data: Any { .. }, is_exited: false, cpu_affinity: AtomicIdSet { bits: [3], phantom: PhantomData<ostd::cpu::id::CpuId> }, sched_attr: SchedAttr { policy: SchedPolicyState { kind: AtomicSchedPolicyKind(2), policy: UnsafeCell { .. } }, last_cpu: AtomicCpuId(1), real_time: RealTimeAttr { prio: 99, time_slice: 35893960 }, fair: FairAttr { weight: 1024, pending_weight: UnsafeCell { .. }, vruntime: 126391798147 } } })
```

```c
#define _GNU_SOURCE

#include <errno.h>
#include <pthread.h>
#include <sched.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mount.h>
#include <sys/stat.h>
#include <unistd.h>

#define ITERATIONS 100000
#define RACE_DIR "/tmp/mount_race"
#define MOUNT_A RACE_DIR "/a"
#define MOUNT_B RACE_DIR "/a/b"

static volatile int running = 1;
static volatile int started = 0;

static void ensure_dir(const char *path)
{
      if (mkdir(path, 0755) < 0 && errno != EEXIST) {
              fprintf(stderr, "mkdir(%s): %s\n", path, strerror(errno));
              exit(1);
      }
}

static void *thread_parent(void *arg)
{
      (void)arg;
      while (!started)
              ;
      for (int i = 0; i < ITERATIONS && running; i++) {
              if (mount("tmpfs", MOUNT_A, "tmpfs", 0, "size=1M") != 0)
                      continue;
              mkdir(MOUNT_B, 0755);
              umount2(MOUNT_A, MNT_DETACH);
      }
      running = 0;
      return NULL;
}

static void *thread_child(void *arg)
{
      (void)arg;
      while (!started)
              ;
      for (int i = 0; i < ITERATIONS && running; i++) {
              if (mount("tmpfs", MOUNT_B, "tmpfs", 0, "size=1M") == 0) {
                      umount2(MOUNT_B, MNT_DETACH);
              }
      }
      running = 0;
      return NULL;
}

int main(void)
{
      if (unshare(CLONE_NEWNS) != 0) {
              fprintf(stderr, "unshare(CLONE_NEWNS): %s\n", strerror(errno));
              return 1;
      }
      ensure_dir(RACE_DIR);
      ensure_dir(MOUNT_A);
      if (mount("tmpfs", MOUNT_A, "tmpfs", 0, "size=1M") != 0) {
              fprintf(stderr, "initial mount: %s\n", strerror(errno));
              return 1;
      }
      ensure_dir(MOUNT_B);

      pthread_t t1, t2;
      pthread_create(&t1, NULL, thread_parent, NULL);
      pthread_create(&t2, NULL, thread_child, NULL);
      started = 1;
      pthread_join(t1, NULL);
      pthread_join(t2, NULL);

      fprintf(stderr, "mount_race: survived %d iterations without kernel panic\n", ITERATIONS);
      return 0;
}
```

avasummer:
This PR adds a `RwMutex`(`namespace_lock`) to serialize mount tree structural mutations, and prevents panic in concurrent umount operations caused by a `Weak::upgrade().unwrap()`, returning `EINVAL` instead.

A regression test is included to reproduce the race on SMP=2.

Closes #3392

### DIFF
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -7,8 +7,8 @@ use core::time::Duration;
 pub(in crate::fs) use dentry::Dentry;
 use dentry::DirDentry;
 use inherit_methods_macro::inherit_methods;
-use mount::MountNsFileCopying;
 pub use mount::{Mount, MountPropType, PerMountFlags};
+use mount::{MountNsFileCopying, MountTopology};
 pub use mount_namespace::MountNamespace;
 pub use resolver::{
     AT_FDCWD, AbsPathResult, EmptyPathStr, FsPath, LookupResult, PathResolver, SplitPath,
@@ -216,8 +216,12 @@ impl Path {
     }
 
     /// Finds the corresponding `Path` in the given mount namespace.
-    fn find_corresponding_mount(&self, mnt_ns: &Arc<MountNamespace>) -> Option<Self> {
-        let corresponding_mount = self.mount.find_corresponding_mount(mnt_ns)?;
+    fn find_corresponding_mount(
+        &self,
+        mnt_ns: &Arc<MountNamespace>,
+        topology: &MountTopology,
+    ) -> Option<Self> {
+        let corresponding_mount = self.mount.find_corresponding_mount(mnt_ns, topology)?;
         let corresponding_path = Self::new(corresponding_mount, self.dentry.clone());
 
         Some(corresponding_path)
@@ -229,7 +233,7 @@ impl Path {
     /// of the `root` path. The check traverses upwards from the current path,
     /// crossing mount point boundaries as necessary, until it either finds
     /// the `root` path or reaches the global root.
-    fn is_reachable_from(&self, root: &Path) -> bool {
+    fn is_reachable_from(&self, root: &Path, _topology: &MountTopology) -> bool {
         let mut owned;
         let mut current = self;
 
@@ -313,7 +317,10 @@ impl Path {
             return_errno_with_message!(Errno::EINVAL, "the path is not in this mount namespace");
         }
 
-        let child_mount = self.mount.do_mount(fs, flags, &self.dentry, source)?;
+        let mut topology_guard = MountTopology::write_lock();
+        let child_mount =
+            self.mount
+                .do_mount(fs, flags, &self.dentry, source, &mut topology_guard)?;
 
         Ok(child_mount)
     }
@@ -333,7 +340,7 @@ impl Path {
             return_errno_with_message!(Errno::EINVAL, "the path is not a mount root");
         }
 
-        let Some(mountpoint) = self.mount.mountpoint() else {
+        let Some(_mountpoint) = self.mount.mountpoint() else {
             return_errno_with_message!(Errno::EINVAL, "the root mount cannot be unmounted");
         };
 
@@ -345,8 +352,12 @@ impl Path {
 
         self.mount.sync()?;
 
+        let mut topology_guard = MountTopology::write_lock();
+        let Some(mountpoint) = self.mount.mountpoint() else {
+            return_errno_with_message!(Errno::EINVAL, "the mount has been detached");
+        };
         let parent_mount = self.mount.parent().unwrap().upgrade().unwrap();
-        let child_mount = parent_mount.do_unmount(&mountpoint)?;
+        let child_mount = parent_mount.do_unmount(&mountpoint, &mut topology_guard)?;
 
         Ok(child_mount)
     }
@@ -378,7 +389,9 @@ impl Path {
             return_errno_with_message!(Errno::EINVAL, "the path is not in this mount namespace");
         }
 
-        self.mount.remount(mount_flags, fs_flags, data, ctx)
+        let mut topology_guard = MountTopology::write_lock();
+        self.mount
+            .remount(mount_flags, fs_flags, data, ctx, &mut topology_guard)
     }
 
     /// Creates a bind mount from the current path to the destination path.
@@ -426,14 +439,16 @@ impl Path {
             );
         }
 
+        let mut topology_guard = MountTopology::write_lock();
         let current_mnt_ns_weak = Arc::downgrade(current_mnt_ns);
         let new_mount = self.mount.clone_mount_tree(
             &self.dentry,
             &current_mnt_ns_weak,
             recursive,
             MountNsFileCopying::Copy,
+            &topology_guard,
         )?;
-        new_mount.graft_mount_tree(dst_path);
+        new_mount.graft_mount_tree(dst_path, &mut topology_guard);
         Ok(())
     }
 
@@ -473,10 +488,11 @@ impl Path {
                 "the destination path is not in this mount namespace"
             );
         }
-        current_mnt_ns.check_no_mnt_ns_loop_in_tree(self.mount_node())?;
+        let mut topology_guard = MountTopology::write_lock();
+        current_mnt_ns.check_no_mnt_ns_loop_in_tree(self.mount_node(), &topology_guard)?;
         if dst_path
             .mount_node()
-            .is_equal_or_descendant_of(self.mount_node())
+            .is_equal_or_descendant_of(self.mount_node(), &topology_guard)
         {
             // Reject moves that would place a mount beneath itself, because the mount tree
             // must remain acyclic.
@@ -486,7 +502,7 @@ impl Path {
             );
         }
 
-        self.mount.graft_mount_tree(dst_path);
+        self.mount.graft_mount_tree(dst_path, &mut topology_guard);
 
         Ok(())
     }
@@ -508,7 +524,9 @@ impl Path {
             return_errno_with_message!(Errno::EINVAL, "the path is not in this mount namespace");
         }
 
-        self.mount.set_propagation(prop, recursive);
+        let mut topology_guard = MountTopology::write_lock();
+        self.mount
+            .set_propagation(prop, recursive, &mut topology_guard);
 
         Ok(())
     }
--- a/kernel/src/fs/vfs/path/mount.rs
+++ b/kernel/src/fs/vfs/path/mount.rs
@@ -560,12 +560,6 @@ impl Mount {
         ctx: &Context,
         _topology: &mut MountTopology,
     ) -> Result<()> {
-        // TODO: This lock is a workaround to guarantee the atomicity of remount operation.
-        // We need to re-design the lock mechanism of `Mount` and file system in the future.
-        static REMOUNT_LOCK: Mutex<()> = Mutex::new(());
-
-        let _guard = REMOUNT_LOCK.lock();
-
         if let Some(flags) = fs_flags {
             self.fs.set_fs_flags(flags, data, ctx)?;
         }
--- a/kernel/src/fs/vfs/path/mount.rs
+++ b/kernel/src/fs/vfs/path/mount.rs
@@ -5,6 +5,7 @@ use core::sync::atomic::{AtomicU32, Ordering};
 use atomic_integer_wrapper::define_atomic_version_of_integer_like_type;
 use hashbrown::HashMap;
 use id_alloc::IdAlloc;
+use ostd::sync::{RwMutexReadGuard, RwMutexWriteGuard};
 use spin::Once;
 
 use super::try_get_mnt_ns_inode;
@@ -22,6 +23,35 @@ use crate::{
     },
     prelude::*,
 };
+
+/// Provides synchronized access to mount topology.
+pub(super) struct MountTopology {
+    _private: (),
+}
+
+fn global_mount_topology() -> &'static RwMutex<MountTopology> {
+    static MOUNT_TOPOLOGY: RwMutex<MountTopology> = RwMutex::new(MountTopology { _private: () });
+    &MOUNT_TOPOLOGY
+}
+
+impl MountTopology {
+    /// Acquires the write side of the mount topology lock.
+    ///
+    /// Use this for operations that may change the mount topology,
+    /// including the parent-child links, mountpoints, mount propagation state,
+    /// or namespace-visible mount trees.
+    pub(super) fn write_lock() -> RwMutexWriteGuard<'static, Self> {
+        global_mount_topology().write()
+    }
+
+    /// Acquires the read side of the mount topology lock.
+    ///
+    /// Use this for operations that need a stable view of mount topology
+    /// without changing it.
+    pub(super) fn read_lock() -> RwMutexReadGuard<'static, Self> {
+        global_mount_topology().read()
+    }
+}
 
 /// Controls how recursive mount-tree cloning handles mount-namespace files.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
@@ -291,6 +321,7 @@ impl Mount {
         flags: PerMountFlags,
         mountpoint: &Arc<Dentry>,
         source: Option<String>,
+        _topology: &mut MountTopology,
     ) -> Result<Arc<Self>> {
         if mountpoint.type_() != InodeType::Dir {
             return_errno!(Errno::ENOTDIR);
@@ -313,14 +344,18 @@ impl Mount {
     /// Unmounts a child mount node from the mountpoint and returns it.
     ///
     /// The mountpoint should belong to this mount node, or an error is returned.
-    pub(super) fn do_unmount(&self, mountpoint: &Dentry) -> Result<Arc<Self>> {
+    pub(super) fn do_unmount(
+        &self,
+        mountpoint: &Dentry,
+        topology: &mut MountTopology,
+    ) -> Result<Arc<Self>> {
         let child_mount = self
             .children
             .write()
             .remove(&mountpoint.key())
             .ok_or_else(|| Error::with_message(Errno::ENOENT, "can not find child mount"))?;
 
-        child_mount.clear_mountpoint();
+        child_mount.clear_topology_link(topology);
 
         Ok(child_mount)
     }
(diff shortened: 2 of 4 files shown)


kernel/src/fs/vfs/path/mod.rs
```diff
@@ -345,7 +346,12 @@ impl Path {
 
         self.mount.sync()?;
 
-        let parent_mount = self.mount.parent().unwrap().upgrade().unwrap();
+        let _tree_guard = current_mnt_ns.namespace_lock().write();
```
cchanging: The `mountpoint` is obtained without holding the `namespace_lock`, so there maybe some operations like `move_mount` make the `mountpoint` out of date?

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -96,6 +98,11 @@ impl MountNamespace {
         self.root.as_ref().unwrap()
     }
 
+    /// Returns the lock that serializes mount tree structural mutations.
+    pub(super) fn namespace_lock(&self) -> &RwMutex<()> {
```
cchanging: It seems that you only use the write lock of the `namespace_lock`. 

I think we need some semantic-level documentation here to describe what properties this lock is intended to protect. You may also need to consider whether there are places that require a consistent view (such as clone operations) and therefore should acquire the read lock here.

cchanging (commented):
Thanks for trying to address the existing race issues in mount! First, I think you may need to provide a clear semantic-level definition of this newly introduced lock.

kernel/src/fs/vfs/path/mod.rs
avasummer: This scenario was considered, it would return `ENOENT` from `do_unmount` if the mountpoint stale, though `EINVAL` is more appropriate. I can move the `mountpoint()` inside the lock, while keeping the original outside for root mount case.

kernel/src/fs/vfs/path/mount_namespace.rs
avasummer: The comment now is "the lock that serializes mount tree structural mutations". Would it make sense to extend the semantics to "protects mount tree topology" that write side for mutations, read side for consistent snapshots? I'll add the read lock to `new_clone`. Happy to discuss the documentation wording in a follow-up comment.

cchanging: > Would it make sense to extend the semantics to "protects mount tree topology" that write side for mutations, read side for consistent snapshots? 

This makes more senses to me. 

>  I'll add the read lock to `new_clone`

Besides `new_clone`, I think some other opeartions like `PathResolver::collect_visible_mounts` should also take the read lock.

---

I have checked the Linux's source code:
https://codebrowser.dev/linux/linux/fs/namespace.c.html#namespace_sem

It seems that Linux defines such lock as a global lock. I think it should be fine for now to implement the lock directly as a global lock, and this would also help later when implementing a cross-namespace propagation mechanism for mounts. What do you think?

avasummer:
Force-pushed. Changed to a global `MOUNT_TREE_LOCK` instead of per namespace lock. Added read lock to `PathResolver::collect_visible_mounts` and `new_clone`, and write lock to `Drop`. Replaced more `unwrap()` calls with graceful error handling.

By the way, it seems that `REMOUNT_LOCK` is redundant since the introduction of `MOUNT_TREE_LOCK` and should be removed later.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -246,14 +247,10 @@ impl Path {
...
             };
 
-            owned = Path::new(
-                parent_mount,
-                current
-                    .mount
-                    .mountpoint()
-                    .expect("Mounts with parents must have a mount point")
-                    .clone(),
-            );
```
cchanging, line 241: Here we actually prefer the original `expect`/`unwrap` style, because in this case an actual panic would only happen if there is a bug in the mount system. In such situations, a panic helps us detect the bug. Besides, such pattern will have a better performance since it does not have error handling. We only prefer error handling in cases that may legitimately fail during execution.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -345,7 +343,15 @@ impl Path {
 
         self.mount.sync()?;
 
-        let parent_mount = self.mount.parent().unwrap().upgrade().unwrap();
```
cchanging, line 405: Same here, if it should not fail, just use `expect`/`unwrap`. You can also check the other similar changes.

test/initramfs/src/regression/fs/mount/mount_race.c
```diff
@@ -0,0 +1,99 @@
...
+	started = 1;
+
+	pthread_join(t1, NULL);
+	pthread_join(t2, NULL);
+
+	fprintf(stderr,
+		"mount_race: survived %d iterations without kernel panic\n",
+		ITERATIONS);
+	return 0;
+}
```
cchanging: You can check the other tests in `/regression`. We prefer to use `FN_TEST` framework (see `.../regression/common/test.h` for details).

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -208,6 +208,9 @@ impl Drop for MountNamespace {
             // and thus the subsequent cleanup logic can be skipped.
             return;
         };
+
+        let _tree_guard = MOUNT_TREE_LOCK.write();
+
```
cchanging, line 218: This locking approach is semantically correct, but a better, more Rust-idiomatic way would be to use an abstraction to enforce that certain operations can only be performed while holding this global write lock or read lock. You can refer to the `CgroupMembership` abstraction; similarly, here we could perhaps have:

```rust
 // in mount.rs

 /// Provides synchronized access to mount topology.
  pub(super) struct MountTopology {
      _private: (),
  }

  fn global_mount_topology() -> &'static RwMutex<MountTopology> {
      static MOUNT_TOPOLOGY: RwMutex<MountTopology> =
          RwMutex::new(MountTopology { _private: () });
      &MOUNT_TOPOLOGY
  }

  impl MountTopology {
      /// Acquires the write side of the mount topology lock.
      ///
      /// Use this for operations that may change the mount topology,
      /// including the parent-child links, mountpoints, mount propagation state, 
      /// or namespace-visible mount trees.
      pub(super) fn write_lock() -> RwMutexWriteGuard<'static, Self> {
          global_mount_topology().write()
      }

      /// Acquires the read side of the mount topology lock.
      ///
      /// Use this for operations that need a stable view of mount topology
      /// without changing it.
      pub(super) fn read_lock() -> RwMutexReadGuard<'static, Self> {
          global_mount_topology().read()
      }
  }
```

cchanging (commented):
Thanks for continuing to move this PR forward. I have some suggestions that I hope will help further improve it. If you have any questions about these comments or run into any difficulties with the implementation, please let me know.

kernel/src/fs/vfs/path/mod.rs
avasummer: I think `mountpoint().unwrap()` can panic in a legitimate case. `do_unmount` clears the child's mountpoint but does not clear its parent pointer. In `find_corresponding_mount() `, it traces the parent chain and panics.
```rust
    pub(super) fn find_corresponding_mount(
        &self,
        mnt_ns: &Arc<MountNamespace>,
    ) -> Option<Arc<Self>> {
        // Collect the ancestors from self to the root mount (The root mount is not included).
        let mut ancestors = VecDeque::new();
        let mut current = self.this();
        while let Some(weak_parent) = current.parent() {
            ancestors.push_front(current.clone());
            current = weak_parent.upgrade().unwrap();
        }

        let mut target_mount = mnt_ns.root().clone();
        while let Some(ancestor) = ancestors.pop_front() {
            // Find the child mount that matches the mountpoint of ancestor.
            let mount_point = ancestor.mountpoint().unwrap();
```

 The code below can reproduce it:
```
[    23.237] ERROR : Uncaught panic:
        called `Option::unwrap()` on a `None` value
        at /root/asterinas/kernel/src/fs/vfs/path/mount.rs:601
```

```c
#define _GNU_SOURCE

#include <errno.h>
#include <pthread.h>
#include <sched.h>
#include <stdatomic.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mount.h>
#include <sys/stat.h>
#include <sys/wait.h>
#include <unistd.h>

#define ITERATIONS 50000
#define MNT_A "/tmp/unshare_race/a"
#define MNT_B "/tmp/unshare_race/a/b"

static atomic_int running = 1;
static atomic_int started = 0;

static void ensure_dir(const char *path)
{
	if (mkdir(path, 0755) < 0 && errno != EEXIST) {
		fprintf(stderr, "mkdir(%s): %s\n", path, strerror(errno));
		exit(1);
	}
}

// Thread A: repeatedly umount/mount at MNT_A, clearing its mountpoint.
static void *thread_umount(void *arg)
{
	(void)arg;

	while (!started)
		sched_yield();

	for (int i = 0; i < ITERATIONS && running; i++) {
		umount2(MNT_A, MNT_DETACH);
		mount("tmpfs", MNT_A, "tmpfs", 0, "size=1M");
		mkdir(MNT_B, 0755);
	}

	running = 0;
	return NULL;
}

// Thread B: repeatedly unshare(CLONE_NEWNS), which internally calls
// switch_to_mnt_ns → find_corresponding_mount, traversing the old
// namespace's parent chain.
static void *thread_unshare(void *arg)
{
	(void)arg;

	while (!started)
		sched_yield();

	for (int i = 0; i < ITERATIONS && running; i++) {
		pid_t pid = fork();
		if (pid == 0) {
			// Set cwd inside the mount tree so find_corresponding_mount
			// traverses through mount A's parent chain.
			if (chdir(MNT_B) == 0)
				unshare(CLONE_NEWNS);
			_exit(0);
		}
		if (pid > 0)
			waitpid(pid, NULL, 0);
	}

	running = 0;
	return NULL;
}

int main(void)
{
	ensure_dir("/tmp/unshare_race");
	ensure_dir(MNT_A);

	if (mount("tmpfs", MNT_A, "tmpfs", 0, "size=1M") != 0) {
		fprintf(stderr, "initial mount: %s\n", strerror(errno));
		return 1;
	}
	ensure_dir(MNT_B);
	if (mount("tmpfs", MNT_B, "tmpfs", 0, "size=1M") != 0) {
		fprintf(stderr, "initial mount B: %s\n", strerror(errno));
		return 1;
	}

	pthread_t t1, t2;
	if (pthread_create(&t1, NULL, thread_umount, NULL) != 0 ||
	    pthread_create(&t2, NULL, thread_unshare, NULL) != 0) {
		perror("pthread_create");
		return 1;
	}

	started = 1;

	pthread_join(t1, NULL);
	pthread_join(t2, NULL);

	fprintf(stderr,
		"unshare_race: survived %d iterations without kernel panic\n",
		ITERATIONS);
	return 0;
}
```

kernel/src/fs/vfs/path/mod.rs
avasummer: Running the regression test([Add regression test for concurrent umount race](https://github.com/asterinas/asterinas/pull/3402/commits/831bfa8113d75d41fdcbaa2917d5469a35c86ed6)) with the lock commit can cause this panic.
```
[    22.444] ERROR : Uncaught panic:
        called `Option::unwrap()` on a `None` value
        at /root/asterinas/kernel/src/fs/vfs/path/mod.rs:354
```

kernel/src/fs/vfs/path/mod.rs
cchanging: So this panic is serving its purpose: it has exposed another area where the current implementation is still incomplete — `find_corresponding_mount` also needs a stable view, so a global read lock should be held during this period. After this fix, I think the `unwrap` here will not panic in a legitimate case?

In fact, once the global lock is in place, one of the existing `FIXME`s related to `find_corresponding_mount` can be addressed here:
```rust
    // FIXME: This cannot fail if we clone mount namespaces and update resolvers in an atomic way.
    // We currently leak this error to userspace, which is not a correct behavior.
    pub fn switch_to_mnt_ns(&mut self, mnt_ns: &Arc<MountNamespace>) -> Result<()> {
```

However, I don’t think this change necessarily needs to be completed in the current PR since it involves more modifications.

---
EDIT: At the same time, the issue you mentioned also seems valid: we probably need to clear the `parent` pointer when calling `Mount::clear_mountpoint`.

kernel/src/fs/vfs/path/mod.rs
cchanging: Similarly, I guess the panic here is likely caused by the fact that we do not check whether there are still child mounts during `umount`, and we have not correctly implemented the `MNT_DETACH` flag. So the proper fix should be to revise the `umount` logic.

avasummer: I guess you mean that we should keep `expect`/`unwrap`, and consider fix `umount` in a follow-up PR.

cchanging: > I guess you mean that we should keep expect/unwrap, and consider fix umount in a follow-up PR.

Yeah, fixing it in another PR also makes sense, since this change is relatively independent. However, the added tests may need some adjustment.

kernel/src/fs/vfs/path/mod.rs
avasummer: > EDIT: At the same time, the issue you mentioned also seems valid: we probably need to clear the parent pointer when calling Mount::clear_mountpoint.

I would like to fix this issue first to prevent panic, and we can implement `umount` more properly later. What do you think?

cchanging: I think it is ok.

kernel/src/fs/vfs/path/mount_namespace.rs
avasummer: This approach is much better. I still have a question about that: should the lock be acquired inside the function as it is, or should we pass the lock guard as parameter. If it's the latter, does it make sense to change the related function signatures in this PR?

cchanging: I think we can pass this lock's guard as a parameter to the corresponding function, so the hidden meaning is that this operation needs to be performed while holding the global lock.

We can do something like:
```rust
pub(super) fn find_corresponding_mount(
    &self,
    mnt_ns: &Arc<MountNamespace>,
    _guard: &MountTopology,
) -> Option<Arc<Self>> {
    ....
}

pub(super) fn do_mount(
    self: &Arc<Self>,
    ...
    _guard: &mut MountTopology,
) -> Result<Arc<Self>> {
}
```

But I haven't tried it, so I'm not sure if I'd run into any problems.

cchanging:
@avasummer A kind ping to move this PR forward, or did you run into some other issue?

avasummer:
Sorry for late response. I've been a bit tied up lately, but I will try to get this updated this afternoon.

avasummer:
Reverted the panic-handling workaround and added a `MountTopology` abstraction. The topology lock guard is now passed to functions that read or mutate mount topology. I am not sure whether additional tests are needed here.

The test run still fails due to a known, separate `umount` issue:

> We do not check whether there are still child mounts during `umount`, and `MNT_DETACH` is not implemented correctly. The proper fix should revise the `umount` logic.

kernel/src/fs/vfs/path/mount.rs
cchanging: We can remove these code now.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -440,8 +481,17 @@ impl Mount {
...
     }
 
+    /// Detaches the mount node from its namespace-visible topology.
+    ///
+    /// The parent pointer and mountpoint describe the same topology edge, so
+    /// they must be cleared together while holding the mount topology lock.
+    pub(super) fn detach_from_namespace(&self, _topology: &mut MountTopology) {
+        self.set_parent(None);
+        self.clear_mountpoint();
+    }
```
cchanging: We can use `detach_from_namespace` in `detach_from_parent`:
```rust
if let Some(child) = child {
    child.detach_from_namespace(topology);
}
```

Besides, this two methods have similar names, I suggest we can rename `detach_from_namespace` to `clear_topology_link` or some better names. Emphasize that it only mutates its own state, to avoid users confusing the two functions.

cchanging (commented):
Thanks for the changes! Most of them look good to me; I only have a few minor comments.

Also, we generally require every PR to pass CI. So I think any newly added tests that are currently failing can be moved to a follow-up PR, and this PR can keep only the tests that pass. If it’s difficult to write a race test that passes reliably, I think it would also be fine to defer all of those tests to a later PR.

avasummer:
Thanks for the review. I have removed the test and `REMOUNT_LOCK`, renamed `detach_from_namespace` to `clear_topology_link` and reused it in `detach_from_parent`.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -411,13 +419,15 @@ impl PathResolver {
                 "`new_root` or the current root is on the rootfs mount"
             );
         }
-        if !put_old_path.is_reachable_from(&new_root_path) {
+        let mut tree_guard = MountTopology::write_lock();
```
cchanging: Maybe change all `tree_guard` to `topology_guard` will be better?

cchanging (approved):
Only one small nit, otherwise LGTM!

PR #3403, opened 2026-06-17
Implement SEEK_END for block devices

### LINKED ISSUE #3212: Support SEEK_END syscall on block device
### Describe the bug

Block device file descriptor does not provide Linux-compatible `lseek(fd, 0, SEEK_END)` behavior. Some userspace block device consumers such as `fsck.exfat` rely on this operation to discover device size.

### To Reproduce

1. Prepare a clean exFAT image such as `test.img`.
2. Boot a NixOS guest and attach the same image as a virtio-blk disk. For example, QEMU can expose it with:

```text
-drive if=none,format=raw,id=xfstest0,file=test.img
-device virtio-blk-pci,...,drive=xfstest0,...
```

In the guest, this device appears as `/dev/vdb`.

3. Open the device as as file, and run a direct `lseek(..., SEEK_END)` probe against that device, for example:

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
    int fd = open("/dev/vdb", O_RDONLY);
    if (fd < 0) {
        perror("open");
        return 1;
    }

    errno = 0;
    off_t end = lseek(fd, 0, SEEK_END);
    printf("lseek_end=%lld errno=%d\n", (long long)end, errno);
    return 0;
}
```

4. The syscall will fail with `errno=22` (`EINVAL`, "Invalid argument").

### Expected behavior

`lseek(fd, 0, SEEK_END)` on an `fd` of block device should return the device size.

### Screenshots

Image

### Environment

- Hardware platform: x86_64 QEMU/KVM guest with a virtio-blk data disk

### Additional context

Related issues:

- Tracker: [`#3211`](https://github.com/asterinas/asterinas/issues/3211)

hburaylee:
## Summary
This PR adds support for SEEK_END positioning on block device files, aligning the behavior with standard Linux block device semantics.

In Linux, testing with loop devices. If the environment lacks `/dev/loop-control`,  `/dev/loopxxx` , or sufficient permissions, the test will be skipped.

**Related Issue :** This PR fixes #3212

### DIFF
--- a/kernel/src/device/registry/block.rs
+++ b/kernel/src/device/registry/block.rs
@@ -154,6 +154,16 @@ impl FileOps for OpenBlockFile {
     }
 }
 
+impl OpenBlockFile {
+    fn device_size(&self) -> Result<usize> {
+        self.0
+            .metadata()
+            .nr_sectors
+            .checked_mul(SECTOR_SIZE)
+            .ok_or_else(|| Error::with_message(Errno::EOVERFLOW, "block device size overflows"))
+    }
+}
+
 impl Pollable for OpenBlockFile {
     fn poll(&self, mask: IoEvents, _: Option<&mut PollHandle>) -> IoEvents {
         let events = IoEvents::IN | IoEvents::OUT;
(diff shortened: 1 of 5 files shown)


hburaylee:
@Halifuda Thanks for the work!
I also noticed issue #3212  earlier and submitted this PR (#3403 ). However, your PR (#3420 ) is more complete, and is based on the latest main branch.
I agree that yours should be prioritized, so I'm closing this one. Thank you for taking over and improving the fix! 👍

Issue #3295, opened 2026-05-29
Reverse mappings for file-backed `Vmo`s


lrh2000:
## Motivation

Correct semantics of page cache operations if the page is mapped to userspace.

 - direct `read()`/`write()`
 - `resize()`
 - `flush()`
 - `evict()`

## Design

Add a new filed to `Vmo`:

```rust
pub struct Vmo {
    //...

    /// Reserve mappings.
    pub(super) rmap: Mutex<Rmap>,
}
```

Implement `Rmap` as the reverse mappings from a `Vmo` to `Vmar`s:

```rust
/// Reverse mappings from a [`Vmo`] to [`Vmar`]s.
///
/// [`Vmo`]: super::Vmo
pub struct Rmap {
    entries: BTreeMap<KeyableWeak<Vmar>, Vec<RmapEntry>>,
}

/// A reverse mapping entry.
#[derive(Copy, Clone, Debug)]
pub struct RmapEntry {
    /// The virtual address.
    pub vaddr: Vaddr,
    /// The VMO offset.
    pub offset: usize,
    /// The mapping size.
    pub size: usize,
}
```

`Vmar` will maintain the reverse mapping when creating or destroying mappings.
```rust
    /// Inserts a new reverse mapping entry.
    pub(super) fn insert(&mut self, vmar: Weak<Vmar>, entry: RmapEntry);

    /// Removes a reverse mapping entry.
    ///
    /// # Panics
    ///
    /// This method will panic if the reverse mapping entry does not exist.
    pub(super) fn remove(&mut self, vmar: Weak<Vmar>, vaddr: Vaddr);
```

Then, `Rmap` provides APIs that can be used to either remove all the mapped pages or mark them as read-only for a given `Vmo` offset range.
```rust
impl Rmap {
    /// Iterates over all reverse mappings and unmaps the given offset range.
    ///
    /// # Panics
    ///
    /// This method may panic if the offset range is not aligned to the page boundary.
    pub fn unmap(&mut self, offset: Range<usize>);

    /// Iterates over all reverse mappings and freezes (i.e., makes read-only) the given offset
    /// range.
    ///
    /// # Panics
    ///
    /// This method may panic if the offset range is not aligned to the page boundary.
    pub fn freeze(&mut self, offset: Range<usize>);
}
```

## Operations

**direct `read()`/`write()`:**

The caller should hold the inode read/write locks to serialize. 

[...]

Issue #3142, opened 2026-04-16
Tracking issue for gaps found by `xfstests`


Fischer0522:
This issue tracks kernel/filesystem gaps exposed by `xfstests` failures. Each entry describes the current state, a brief solution overview, and the affected tests.

### 1. Overview

| Metric | Legacy Ext2 | Refactored Ext2 | Improvement |
|--------|-------------|-----------------|-------------|
| Passing xfstests | 53 | 86 | +62.2% |

> **Note:** The "53 passing tests" for the legacy implementation is an inflated metric achieved by disabling `fs.sync()` on unmount. Under standard-compliant conditions (sync-on-unmount enabled), the legacy code passes only **6** tests. The refactored Ext2, conversely, achieves 86 passes while maintaining full on-disk consistency.

#### The Impact of Redesign

The 22 / 33 improvement reflects a shift from localized patches to architectural alignment with Linux. We have moved from handling edge cases to guaranteeing behavioral invariants:

| Area | Tests Unlocked |
|------|----------------|
| Sparse File & Page Cache | 14 |
| Block Allocation & fallocate | 5 |
| Sync Path & Durability | 3 |
| **Total** |  **22** |

Crucially, the redesigned inode lifecycle and synchronization path ensure on-disk state consistent. This resolves the persistent data-loss issues of the legacy implementation, moving the filesystem closer to production-grade reliability.

The other 11 unlocked tests are due to supporting new features or bug fixes.

---

#### Sparse File Support
*   **Problem**: Legacy implementation lacked sparse file support; every byte range was eagerly allocated, causing significant space amplification and divergence from Linux semantics.
*   **Outcome**: Implemented hole-based storage for unwritten regions. Integrated precise partial-block handling and zero-fill logic in the page cache to prevent stale data exposure during `write`, `fallocate`, and `truncate`.

#### Inode Lifecycle Management
*   **Problem**: Previous eager reclamation upon `sync()` violated POSIX semantics, causing unlinked-but-open files to become inaccessible prematurely.
*   **Outcome**: Migrated to an `Arc`-based lifecycle. Inodes are now strictly reclaimed only when both `link_count` is zero **and** all in-kernel references are released, ensuring POSIX-compliant file persistence.

#### Block Allocation
*   **Problem**: 
    1. Delayed allocation diverged from standard Ext2 semantics.
    2. Redundant, synchronized state between `InodeDesc` and `InodeBlockManager` introduced deadlock risks and consistency overhead.
[...]

PR #3405, opened 2026-06-17
Atomically allocate and zero ext2 data blocks

lrh2000:
ext2 code calls `Vmo::commit_on`. Doing so will break the lock contract used in the reverse mapping (see https://github.com/asterinas/asterinas/issues/3295).

https://github.com/asterinas/asterinas/blob/f7ff85597d892ec7476489216672b0ad61b7090f/kernel/src/fs/fs_impls/ext2/inode/file.rs#L534-L537

To get a page from a VMO in a racy free way, we must lock the page table lock and use `try_commit_on`. Calling `commit_on` directly cannot prevent the page from being evicting concurrently.

---

Even without this issue, the related code is problematic for some other reasons.

 1. `allocate_range_blocks` and `zero_new_blocks` are racy against concurrent page faults. If a page fault occurs in between, an uninitialized block will be loaded from the disk, which may expose sensitive information. https://github.com/asterinas/asterinas/blob/f7ff85597d892ec7476489216672b0ad61b7090f/kernel/src/fs/fs_impls/ext2/inode/file.rs#L172-L182
 2. `allocate_range_blocks` and `zero_new_blocks` are also racy against concurrent page cache eviction. Currently, we don't have concurrent page cache eviction, but it would be problematic if we did.
 3. It seems that the page cache layer is intended to address EOF-related issues. I am unsure why the ext2 implementation contains duplicate logic. https://github.com/asterinas/asterinas/blob/f7ff85597d892ec7476489216672b0ad61b7090f/kernel/src/fs/fs_impls/ext2/inode/file.rs#L480-L482 https://github.com/asterinas/asterinas/blob/f7ff85597d892ec7476489216672b0ad61b7090f/kernel/src/vm/page_cache/mod.rs#L229-L232

---

This PR proposes zeroing the data blocks (1) **when they are allocated** (2) **while the block manager lock is held**.

Therefore, a page fault (i.e., `Vmo::commit_on`) will see either an unallocated or an initialized block.

Doing so also eliminates the need for the `zero_partial_block_edge`, meaning that there is no longer a `Vmo::commit_on` being called from the FS side.

The downside is that holding the block manager lock for longer may result in less efficient zeroing (e.g., it may be redundant or less batched compared to the state before this PR). But I don't think this is really a big concern. For example, writing to unallocated blocks with `O_DIRECT` incurs a one-off cost due to redundant zeroing, which seems acceptable.

### DIFF
--- a/kernel/comps/nvme/src/lib.rs
+++ b/kernel/comps/nvme/src/lib.rs
@@ -13,7 +13,7 @@ extern crate alloc;
 #[macro_use]
 extern crate ostd_pod;
 
-// Set this module's log prefix for `ostd::log`.
+// Set this crate's log prefix for `ostd::log`.
 macro_rules! __log_prefix {
     () => {
         "nvme: "
--- a/kernel/src/fs/fs_impls/ext2/block_group.rs
+++ b/kernel/src/fs/fs_impls/ext2/block_group.rs
@@ -347,7 +347,7 @@ impl BlockGroup {
         for block_bit in range_start..range_end {
             if !metadata.block_bitmap.is_allocated(block_bit) {
                 warn!(
-                    "ext2_free_blocks: bit already cleared for block {}",
+                    "free_blocks: bit already cleared for block {}",
                     self.first_block + start_bit + (block_bit - range_start) as u32
                 );
             } else {
@@ -407,7 +407,7 @@ impl BlockGroup {
         let mut metadata = self.metadata.write();
 
         if !metadata.inode_bitmap.is_allocated(inode_idx) {
-            warn!("ext2_free_inode: inode idx {} already freed", inode_idx);
+            warn!("free_inode: inode idx {} already freed", inode_idx);
             return Ok(false);
         }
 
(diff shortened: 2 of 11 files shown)


cchanging (commented):
The first four commits look good to me!

I think the changes in the last commit are also acceptable, but in the long term, we should probably make further API changes, because block allocation currently happens on the write path. With some API adjustments, we should be able to avoid zeroing the portions that are going to be fully overwritten.

Besides, this PR seems to address the page cache architectural gap mentioned in #3142. You could try enabling xfstests generic/127 and generic/346？

lrh2000:
> With some API adjustments, we should be able to avoid zeroing the portions that are going to be fully overwritten.

This should be possible. I'm not sure how to design the APIs yet.
```rust
        // We must wait until the blocks are initialized before we can proceed.
        // Otherwise, concurrent page faults may see uninitialized blocks.
        //
        // TODO: In the write path, if the entire block is going to be
        // overwritten, we should write the real payload instead of zeroing it.
        Self::zero_new_blocks(fs, &data_blocks_range)?;
```

> generic/346

This is already enabled. https://github.com/asterinas/asterinas/blob/22ca7ccf30757c8ece7dcffbe933b8e1072f052a/test/initramfs/src/conformance/xfstests/short.list#L47

> generic/127

This test fails before the last commit and passes after the last commit. I've enabled it.

cchanging (approved):

StevenJiang1110 (approved):

cqs21 (approved):

PR #3407, opened 2026-06-17
Dispatch kernel params by command-line order

taosue:
This PR updates kernel command-line parameter dispatching to preserve command-line order across different registered keys.

Linux has command-line parameters where multiple flag keys control the same behavior. A typical example is `ro` and `rw`: both configure the root filesystem mount mode, and Linux applies them with **last-wins** semantics. For example, `ro` `rw` results in read-write, while `rw` `ro` results in read-only.

Previously, Asterinas grouped parameters by key and dispatched groups in map order. That works for repeated occurrences of the same key, but it loses ordering between different keys. As a result, parameters like `ro` and `rw` cannot cleanly share one underlying state while preserving Linux-compatible last-wins behavior.

This PR records the last command-line position for each recognized parameter key. Before dispatching, it partitions early and regular parameters as before, then sorts each group by the recorded last position. This keeps the existing early/non-early dispatch boundary while making cross-key flag interactions deterministic and compatible with Linux-style command-line semantics.

### DIFF
--- a/kernel/comps/cmdline/src/dispatch.rs
+++ b/kernel/comps/cmdline/src/dispatch.rs
@@ -9,6 +9,7 @@ use alloc::{
     collections::BTreeMap,
     ffi::CString,
     string::{String, ToString},
+    vec,
     vec::Vec,
 };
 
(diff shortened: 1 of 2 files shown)


StevenJiang1110 (approved):

PR #3351, opened 2026-06-09
Remove VT methods from `AnyConsoleDevice`

lrh2000:
Based on the current VT support architecture, the implementation is located in the `kernel/src/device/tty/vt` directory and works solely with the `FramebufferConsole`. Therefore, there is no need for VT interfaces on `AnyConsoleDevice` (other consoles cannot support VT after all).

https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/comps/console/src/lib.rs#L32-L33
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/comps/console/src/lib.rs#L43-L78

These methods are no longer implemented by any console devices, so removing them is a trivial effort. I also moved some type definitions to more suitable locations.

@cqs21 originally commented in https://github.com/asterinas/asterinas/pull/2211#discussion_r2184686373:
> Downcasting `AnyConsoleDevice` to `FramebufferConsole` is weird here, especially when I run `setfont -C /dev/tty0 32x12.psf.gz`. To address this, I propose adding a `set_font` method to the `AnyConsoleDevice` trait, relocating `font.rs` to `aster_console`, and simply calling `self.console.set_font(font)?` here.

There is no longer a need for a downcast because the ioctl is now implemented directly on the VT TTY, which contains an internal `FramebufferConsole`. I moved the font type back to `aster-framebuffer` because the framebuffer console is the only possible user in our current architecture.
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/vt/driver.rs#L32-L35
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/vt/driver.rs#L174-L183
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/vt/manager/console.rs#L187-L197
https://github.com/asterinas/asterinas/blob/dde91f0ccbe8d3fb0b239d2ed30da1053a3a8824/kernel/src/device/tty/vt/manager/console.rs#L24-L31

---

The second commit has been moved to https://github.com/asterinas/asterinas/pull/3408

<details>
<summary> The original description of the second commit </summary>

The second commit makes some other cleanups, e.g.,
[...]

PR #3408, opened 2026-06-17
Clean up `eventfd` code, enhance some regression tests, and others

lrh2000:
Move out from #3351 since the code will soon cause conflicts, and I have collected a few other minor fixes that so I feel they worth a separate PR.

I don't think this PR needs a more detailed explanation. Just in case, here are a few things worth noting:
  - The `eventfd` code has some comment-style fixes, as well as some identical code transforms, and
  - some regression tests have been enhanced, and some usages of the `TEST_*` macros have been corrected.

Other fixes should be quite minor.


StevenJiang1110 (approved):

cqs21 (approved):

PR #3413, opened 2026-06-19
Fix O_TRUNC permission check

WaterWhisperer:
## Summary

Fix permission checks for truncation.

## Changes

- Check `MAY_WRITE` in `Path::resize`, covering `O_TRUNC` and `truncate`.
- Add regression for `O_RDONLY | O_TRUNC` and `truncate` on a file without write permission.
- Unblock the gVisor truncation permission cases from common blocklists, while keeping exFAT-specific chmod-dependent cases in `blocklists.exfat`.
- Remove now-passing or stale gVisor blocklist entries in `open_create_test` and `truncate_test`.

## Testing

- `make check`

  ```text
  All regression tests passed.
  ```

  ```text
  79 of 79 test cases passed.
  All conformance tests passed.
  ```

### DIFF
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -556,6 +556,13 @@ impl Path {
 
         DirDentry::rename(&self.dentry, old_name, &new_dir.dentry, new_name)
     }
+
+    /// Resizes the file.
+    pub fn resize(&self, size: usize) -> Result<()> {
+        let inode = self.inode();
+        inode.check_permission(Permission::MAY_WRITE)?;
+        inode.resize(size)
+    }
 }
 
 // Methods inherited from `Inode`.
@@ -568,7 +575,6 @@ impl Path {
     pub fn mode(&self) -> Result<InodeMode>;
     pub fn set_mode(&self, mode: InodeMode) -> Result<()>;
     pub fn size(&self) -> usize;
-    pub fn resize(&self, size: usize) -> Result<()>;
     pub fn owner(&self) -> Result<Uid>;
     pub fn set_owner(&self, uid: Uid) -> Result<()>;
     pub fn group(&self) -> Result<Gid>;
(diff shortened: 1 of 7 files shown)


WaterWhisperer:
The CI failure appears exFAT-specific. Should I restore only this case to the blocklist while keeping the O_TRUNC permission fix?

WaterWhisperer:
It seems that this case depends on chmod changing mode bits, while Asterinas exFAT currently does not preserve Unix mode changes.

https://github.com/asterinas/asterinas/blob/546e32e231429bbb0b5b40ba0fe9e75f38387c9f/kernel/src/fs/fs_impls/exfat/inode.rs#L1459-L1462

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -165,6 +165,7 @@ impl Path {
             && creation_flags.contains(CreationFlags::O_TRUNC)
             && !status_flags.contains(StatusFlags::O_PATH)
         {
+            inode.check_permission(Permission::MAY_WRITE)?;
```
cchanging: Can we move this check to `Path::resize`? It seems that in this way we can also fix the permission check for `truncate` syscall.

cchanging (commented):
> The CI failure appears exFAT-specific. Should I restore only this case to the blocklist while keeping the O_TRUNC permission fix?

We have `test/src/comformance/gvisor/blocklists.exfat`. I think you can add this test to it.

kernel/src/fs/vfs/path/mod.rs
WaterWhisperer: Okay, I'll try.

cchanging (approved):

cqs21 (approved):

lrh2000:
```rust
// Methods inherited from `Dentry`.
#[inherit_methods(from = "self.dentry")]
impl Path {
    // ...


    /// Resizes the file.
    pub fn resize(&self, size: usize) -> Result<()> {
        let inode = self.inode();
        inode.check_permission(Permission::MAY_WRITE)?;
        inode.resize(size)
    }
}


// Methods inherited from `Inode`.
impl Path {
    // ....
}
```

Well, no idea why this method is put under:
```c
// Methods inherited from `Dentry`.
```

It does not seem to be related...

WaterWhisperer:
> 
> Well, no idea why this method is put under:
> 
> ```c
> // Methods inherited from `Dentry`.
> ```
> 
> It does not seem to be related...

You're right, it shouldn't be here. I'll move it near `open` in a follow-up PR.

PR #3419, opened 2026-06-22
Generalize fio benchmarks and add virtio-fs coverage

li041:
## Motivation

The existing fio benchmarks duplicate workload logic across ext2 configurations and only report direct I/O results. They also cannot configure or launch virtiofsd, so virtio-fs is not covered by the benchmark workflow.

This PR generalizes the fio benchmark suite, adds cached and direct measurements, and introduces virtio-fs coverage for both Asterinas and Linux.

Related issue: #3616.

## Changes

- Reorganize fio benchmarks by workload and filesystem:
  - Share workload logic through `fio/common/run.sh`.
  - Move filesystem setup into `fio/fs/{ext2,virtiofs}`.
  - Preserve the existing ext2 and ext2-without-IOMMU configurations.

- Measure cached and direct I/O separately:
  - Run each sequential workload with both `direct=0` and `direct=1`.
  - Add separate result definitions for cached and direct bandwidth.
  - Expand the fio summary to include all filesystem and I/O mode combinations.

- Add virtio-fs sequential read and write benchmarks:
  - Mount virtio-fs inside Asterinas and Linux guests.
  - Configure the tag, socket, shared directory, and virtiofsd binary through `runtime_config`.
  - Add the shared memory backend and `vhost-user-fs-pci` device to Linux benchmark runs.

- Add `tools/start_virtiofsd.sh` for virtiofsd lifecycle management:
  - Validate the virtiofsd binary and prepare required directories.
  - Remove stale sockets and stop the previous daemon associated with the socket.
  - Wait for the vhost-user socket before starting QEMU.
  - Stop the benchmark virtiofsd instance after the Linux run.

- Add Makefile options for enabling and configuring virtio-fs.
- Use a virtio-fs-enabled Linux kernel for benchmarks.
- Generate result filenames from the complete benchmark path to avoid collisions.
- Normalize fio bandwidth values such as `MiB/s` and `GiB/s` to decimal `MB/s`.
- Update the x86-64 and TDX benchmark workflow paths for the new layout.

## Benchmark configuration

```text
fio -rw={read,write} -filename=<test-file> -size=1G -bs=1M \
    -ioengine=sync -direct={0,1} -numjobs=1 -fsync_on_close=1 \
    -time_based=1 -ramp_time=60 -runtime=100
```

Environment:

- QEMU 10.2.1
- virtiofsd 1.10.0 with `--cache auto`
- Linux benchmark kernel 6.16.0
- 1 vCPU
- 8 GiB guest memory

## Benchmark results

All results below were measured from commit `61b6e07e`.

### Virtio-fs

The following single-run results were measured:

| Workload | Mode | Linux (MB/s) | Asterinas (MB/s) |
|---|---|---:|---:|
| read | cached | 32534.4 | 13529.1 |
| read | direct | 10449.1 | 5081.4 |
| write | cached | 3901.75 | 594.543 |
| write | direct | 2156.92 | 3378.51 |

### Ext2

The following single-run results were measured with the default IOMMU configuration:

| Workload | Mode | Linux (MB/s) | Asterinas (MB/s) |
|---|---|---:|---:|
| read | cached | 35326.1 | 19327.4 |
| read | direct | 9888.07 | 6132.07 |
| write | cached | 7875.85 | 18146.2 |
| write | direct | 1405.09 | 1367.34 |

The virtio-fs raw JSON, metadata, and checksums are available in the [benchmark results](https://github.com/li041/asterinas/tree/benchmark_reults/benchmark_results/fio-benchmark/2026-07-21).

These are single-run local measurements intended to validate the benchmark workflow. Repeated CI runs are needed before using them as regression thresholds.


StevenJiang1110 (commented):
I scan the whole PR and it looks like you've substantially refactored the existing tests under `benchmark/fio`, including the directory structure — it seems like you're trying to reuse some logic in `fio/common/run.sh`?

However, I think the cost here outweighs the benefit. First, I don't think the logic in `run.sh` is complex enough to warrant this kind of reuse. It looks very trivial. Also, after this restructuring, the logic for running the tests has become significantly more complicated — at least I can't easily follow the logic of these scripts in a short amount of time. Additionally, we don't have CI in place to validate benchmark correctness, so we have no way to verify whether your changes are correct just by looking at the results. In fact, modifying the main repo alone isn't sufficient here, especially since you've renamed many existing tests. This would break the rendering of results for existing tests, requiring changes to the corresponding historical data keys in the https://github.com/asterinas/benchmark repo as well.

So my overall suggestion is: don't rename/refactor existing benchmarks unless doing so provides a substantial benefit. Instead, you could add the new benchmarks you want under a new directory. For example, if there's already `ext2_seq_read_bw` and you want to test a version with page cache enabled, you could add `ext2_cached_seq_read_bw`. I think that should be clear enough to distinguish between the two. If renaming becomes necessary later, we can reconsider it as a whole at that point.

Also, I'd suggest putting the new ext2 benchmarks and the corresponding changes to the existing data-parsing script in the same PR.

The newly added virtiofs benchmark should go into a separate PR. From what I can tell, these two changes don't really overlap — although, without the script fix, the virtiofs data might not parse correctly, submitting them separately should still be fine.

li041:
I understand the compatibility concern around renaming existing benchmarks and breaking historical result keys.

One concern I have with adding every variant as a separate top-level directory is how quickly the directory structure grows. For each filesystem, we currently have: {direct, cached} × {read, write} × {IOMMU, no IOMMU}. That means eight directories per filesystem. With n filesystems, fio/ would contain roughly `8n` directories, while most of their `run.sh` files would differ only in a few fio options or runtime settings. This seems difficult to scan and may also make the duplicated scripts drift over time.

This was the main motivation for organizing the benchmarks by workload and treating filesystem, cache mode, and IOMMU mode as test dimensions. Reusing the small amount of shell logic was secondary.

Would it make sense to preserve the existing externally visible benchmark names and historical data keys, while still grouping the internal implementation by workload?

StevenJiang1110:
> For each filesystem, we currently have: {direct, cached} × {read, write} × {IOMMU, no IOMMU}.

TBH, I can't quite recall why we need to test with IOMMU both enabled and disabled. Under Asterinas's threat model, IOMMU should always be kept on — otherwise, the soundness of OSTD could be compromised.

Have you tested whether enabling IOMMU has any impact on virtiofs's performance?

li041:
> Have you tested whether enabling IOMMU has any impact on virtiofs's performance? 

Not yet. I’ll look into how to enable IOMMU properly for virtiofs first, then run a performance comparison once it works.

li041:
I investigated enabling IOMMU for virtio-fs, but the bundledvirtiofsd 1.13.3 does not advertise `VIRTIO_F_ACCESS_PLATFORM` or implement IOTLB handling. Consequently, QEMU rejects `vhost-user-fs-pci,iommu_platform=on` as unsupported. Supporting this requires changes in the virtiofsd/vhost-user backend rather than just a QEMU option, so I cannot run a valid IOMMU performance comparison with the current stack.

StevenJiang1110:
> I investigated enabling IOMMU for virtio-fs, but the bundledvirtiofsd 1.13.3 does not advertise `VIRTIO_F_ACCESS_PLATFORM` or implement IOTLB handling. Consequently, QEMU rejects `vhost-user-fs-pci,iommu_platform=on` as unsupported. Supporting this requires changes in the virtiofsd/vhost-user backend rather than just a QEMU option, so I cannot run a valid IOMMU performance comparison with the current stack.

Ok, thanks for looking into this. This seems to be a limitation we hadn't realized before.

What is `bundledvirtiofsd 1.13.3`? Is that the daemon running on the host? Is it the most "official" virtiofsd implementation? Is the issue that this particular virtiofsd wasn't built with IOMMU support? Could we potentially switch to a different virtiofsd implementation?

However, the main reason I asked about IOMMU was to figure out whether a filesystem really needs all `2×2×2 = 8` different configurations to be tested, since my sense is that not every configuration actually needs to be covered. For example, for cached reads/writes, if the data never actually needs to be flushed to disk and only lives in the page cache, then IOMMU vs. no-IOMMU shouldn't matter at all. Of course, I'd need you to confirm this — specifically, whether fio without `O_DIRECT` ever causes data to be flushed to disk (e.g., via a `sync` call).

Because if we really want to account for generality here, I think it gets fairly complex quickly — there are many factors to consider, such as whether we should test the filesystem and bio layers separately (e.g., the filesystem using cache mode, while bio directly reads/writes to a device like `/dev/vda1`), which might also help avoid an explosion of configuration combinations. 

So as long as things don't grow to an unmanageable scale, I'd prefer we keep the current directory structure as-is.

test/initramfs/src/benchmark/fio/summary.yaml
```diff
@@ -3,3 +3,11 @@ benchmarks:
...
   - ext2_seq_write_bw_no_iommu
   - ext2_seq_read_bw_no_iommu
+  - virtiofs_seq_write_bw
+  - virtiofs_seq_read_bw
+  - ext2_seq_write_bw_cached
+  - ext2_seq_read_bw_cached
+  - ext2_seq_write_bw_no_iommu_cached
+  - ext2_seq_read_bw_no_iommu_cached
+  - virtiofs_seq_write_bw_cached
+  - virtiofs_seq_read_bw_cached
```
cchanging, line 13: Let's group entries from the same filesystem together first.

test/initramfs/src/benchmark/fio/summary.yaml
```diff
@@ -3,3 +3,11 @@ benchmarks:
   - ext2_seq_read_bw
   - ext2_seq_write_bw_no_iommu
   - ext2_seq_read_bw_no_iommu
+  - virtiofs_seq_write_bw
+  - virtiofs_seq_read_bw
+  - ext2_seq_write_bw_cached
+  - ext2_seq_read_bw_cached
+  - ext2_seq_write_bw_no_iommu_cached
+  - ext2_seq_read_bw_no_iommu_cached
```
cchanging, line 11: I suggest rename these modules to `ext2_seq_*_bw_cached_no_iommu`

.github/workflows/benchmark_x86.yml
```diff
@@ -59,6 +59,14 @@ jobs:
...
           - fio/ext2_seq_write_bw_no_iommu
           - fio/ext2_seq_read_bw_no_iommu
+          - fio/virtiofs_seq_write_bw
+          - fio/virtiofs_seq_read_bw
+          - fio/ext2_seq_write_bw_cached
+          - fio/ext2_seq_read_bw_cached
+          - fio/ext2_seq_write_bw_no_iommu_cached
+          - fio/ext2_seq_read_bw_no_iommu_cached
+          - fio/virtiofs_seq_write_bw_cached
+          - fio/virtiofs_seq_read_bw_cached
```
cchanging, line 69: Same here, group entries from the same filesystem together first.

cchanging (approved):
Some small issues, otherwise LGTM!

.github/workflows/benchmark_x86.yml
```diff
@@ -59,6 +59,14 @@ jobs:
           - fio/ext2_seq_read_bw
           - fio/ext2_seq_write_bw_no_iommu
           - fio/ext2_seq_read_bw_no_iommu
+          - fio/ext2_seq_write_bw_cached
+          - fio/ext2_seq_read_bw_cached
+          - fio/ext2_seq_write_bw_cached_no_iommu
```
tatetian: The opposite of direct I/O is more commonly referred to as buffered I/O. So I think the `_cached` suffix should be replaced with `_buffered`.

Compared to direct I/O, buffered I/O is the standard or default I/O mode. People would be surprised to see buffered I/O benchmark is marked in the name ( `ext2_seq_write_bw_cached`), yet direct I/O is NOT highlighted in the name (`ext2_seq_write_bw`).

For the reasons above, I suggest renaming the test names in the following pattern:

```
fio/ext2_seq_write_bw                  # buffered I/O, with IOMMU
fio/ext2_seq_write_bw_direct           # direct I/O, with IOMMU
fio/ext2_seq_write_bw_no_iommu         # buffered I/O, without IOMMU
fio/ext2_seq_write_bw_direct_no_iommu  # direct I/O, with IOMMU
```

test/initramfs/src/benchmark/fio/virtiofs_seq_read_bw_cached/run.sh
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+echo "*** Running the FIO cached sequential read test (virtio-fs) ***"
+
+mkdir -p /virtiofs
+mountpoint -q /virtiofs || mount -t virtiofs aster-virtiofs /virtiofs
+/benchmark/bin/fio -rw=read -filename=/virtiofs/fio-test -name=seqread \
+-size=1G -bs=1M \
+-ioengine=sync -direct=0 -numjobs=1 -fsync_on_close=1 \
+-time_based=1 -ramp_time=60 -runtime=100
```
tatetian: Indent the above three lines.

test/initramfs/src/benchmark/fio/ext2_seq_write_bw_cached_no_iommu/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: MPL-2.0
+
+set -e
+
+echo "*** Running the FIO cached sequential write test (Ext2) ***"
```
tatetian: Replace "cached" with "buffered" throughout this PR.

test/initramfs/src/benchmark/fio/virtiofs_seq_read_bw/bench_result.yaml
```diff
@@ -0,0 +1,17 @@
+alert:
+  bigger_is_better: true
+  threshold: 125%
+chart:
+  description: fio -filename=/virtiofs/fio-test -size=1G -bs=1M -direct=1
+  legend: Average file read bandwidth on {system}
+  title: '[virtio-fs] The bandwidth of sequential reads'
```
tatetian: In same spirit as the suggest to renaming the test names, rewrite all file I/O titles (both Ext2 and virtiofs) in the following patterns:

```
[virtio-fs] The bandwidth of sequential reads (buffered I/O, w/t IOMMU)
[virtio-fs] The bandwidth of sequential reads (buffered I/O, w/o IOMMU)
[virtio-fs] The bandwidth of sequential reads (direct I/O, w/t IOMMU)
[virtio-fs] The bandwidth of sequential reads (direct I/O, w/o IOMMU)
```

PR #3420, opened 2026-06-22
Implement SEEK_END for block devices

Halifuda:
## Summary

This PR adds `SEEK_END` support for block device files.

Current `upstream/main` already has the per-open-file `SEEK_END` hook: `InodeHandle::seek()` calls `PerOpenFileOps::seek_end()` when a file is backed by a seekable, offset-aware per-open file object, and the trait default returns `Ok(None)` for implementations that do not support seeking from the end.

Block device files are seekable and offset-aware, but `OpenBlockFile` still uses the default `seek_end()` implementation. As a result, `lseek(fd, 0, SEEK_END)` on a block device still fails with `EINVAL` even though the same device size is already exposed through the `BLKGETSIZE64` ioctl path.

This PR fills that remaining gap by implementing `PerOpenFileOps::seek_end()` for `OpenBlockFile`. The returned end position is computed from the block device metadata as `nr_sectors * SECTOR_SIZE`, using checked arithmetic and returning `Ok(None)` on overflow.

The PR also adds a block-device regression test that compares `lseek(..., SEEK_END)` with the existing `BLKGETSIZE64` byte-size ioctl result.

No other per-open file type is changed. Pipes, pty files, evdev files, `/dev/mem`, framebuffer files, procfs task files, and other implementations keep the existing default behavior unless they add their own `seek_end()` implementation later.

## Testing

- Rebased onto `upstream/main` at `c6284a9106f5a4c87deb7c8a990af6211dc5f540`.
- Added regression coverage in `test/initramfs/src/regression/io/file_io/block_device.c`.
- `make -C test/initramfs/src/regression check`
- Focused regression test to `/test/io/file_io/block_device`
  Result: `test_seek_end_matches_block_device_size summary: 5 tests passed, 0 tests failed`; existing block-device tests also passed (`7/0` and `9/0`).

### DIFF
--- a/kernel/src/device/registry/block.rs
+++ b/kernel/src/device/registry/block.rs
@@ -205,6 +205,10 @@ impl PerOpenFileOps for OpenBlockFile {
         true
     }
 
+    fn seek_end(&self) -> Result<Option<usize>> {
+        Ok(Some(self.0.metadata().nr_sectors * SECTOR_SIZE))
+    }
+
     fn ioctl(&self, raw_ioctl: RawIoctl) -> Result<i32> {
         use ioctl_defs::*;
 
--- a/test/initramfs/src/regression/io/file_io/block_device.c
+++ b/test/initramfs/src/regression/io/file_io/block_device.c
@@ -4,7 +4,9 @@
 
 #include "../../common/test.h"
 #include <fcntl.h>
+#include <linux/fs.h>
 #include <stdint.h>
+#include <sys/ioctl.h>
 #include <unistd.h>
 
 #define SECTOR_SIZE 512
@@ -21,7 +23,6 @@ static int close_block_device(int fd)
 }
 #else
 #include <linux/loop.h>
-#include <sys/ioctl.h>
 
 #define LOOP_BACKING_FILE "/tmp/block_device_file_io.img"
 #define LOOP_BACKING_FILE_SIZE (SECTOR_SIZE * 8)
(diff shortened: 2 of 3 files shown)


Halifuda:
`lseek(..., SEEK_END)` and `ioctl(BLKGETSIZE64)` now use checked size calculation and return `EOVERFLOW` if the block-device byte-size computation overflows. This should be practically unreachable for real devices, but keeps the two paths consistent and avoids treating overflow as “SEEK_END not supported”.
The regression test now casts the successful `lseek()` return value to `uint64_t` before comparing it with the `BLKGETSIZE64` result.

kernel/src/device/registry/block.rs
```diff
@@ -206,7 +206,15 @@ impl PerOpenFileOps for OpenBlockFile {
...
     fn seek_end(&self) -> Result<Option<usize>> {
-        Ok(self.0.metadata().nr_sectors.checked_mul(SECTOR_SIZE))
+        let size = self
+            .0
+            .metadata()
+            .nr_sectors
+            .checked_mul(SECTOR_SIZE)
+            .ok_or_else(|| {
+                Error::with_message(Errno::EOVERFLOW, "the block device size overflows")
+            })?;
```
lrh2000: `nr_sectors * SECTOR_SIZE` should be just fine. Be consistent with existing code:
```
kernel/comps/block/src/impl_block_device.rs:        let device_size = self.metadata().nr_sectors * SECTOR_SIZE;
kernel/comps/block/src/impl_block_device.rs:        let device_size = self.metadata().nr_sectors * SECTOR_SIZE;
kernel/comps/mlsdisk/src/lib.rs:        let end = host_disk.metadata().nr_sectors * SECTOR_SIZE / BLOCK_SIZE;
kernel/src/device/registry/block.rs:        let device_size = self.0.metadata().nr_sectors * SECTOR_SIZE;
kernel/src/device/registry/block.rs:        let device_size = self.0.metadata().nr_sectors * SECTOR_SIZE;
kernel/src/device/registry/block.rs:                let size = (self.0.metadata().nr_sectors * SECTOR_SIZE) as u64;
```

If buggy devices and/or buggy partition tables provide us with buggy sector numbers that could lead to overflow, I think we would prefer to fix this when `metadata()` is constructed (e.g., when querying the device or when parsing the partition table). For example, if a device says it has `u64::MAX` blocks, we should round the number down to the highest number our kernel can handle. This means that the rest of the code can remain unchanged, i.e., no users should perform `checked_mul`.

test/initramfs/src/regression/io/file_io/block_device.c
```diff
@@ -5,10 +5,15 @@
...
 #include <fcntl.h>
 #include <stdint.h>
+#include <sys/ioctl.h>
 #include <unistd.h>
 
 #define SECTOR_SIZE 512
 
+#ifndef BLKGETSIZE64
+#define BLKGETSIZE64 _IOR(0x12, 114, size_t)
+#endif
```
lrh2000: This should not be necessary. Use `#include <linux/fs.h>` instead. https://github.com/asterinas/asterinas/pull/3403 seems to do so and have passed the CI.

lrh2000 (commented):
Thanks for implementing `SEEK_END` for block devices.

Please squash the commits into one. We prefer atomic commits within a single PR.

lrh2000 (approved):

cqs21 (approved):

Issue #3257, opened 2026-05-22
Locking same page in `Vmo::write` and page fault handler will deadlock


lrh2000:
### Describe the bug

We lock a page in `Vmo::write`, and then we perform a fallible read/write. However, the same page may need to be locked again in the page fault handler, which could result in a deadlock.

https://github.com/asterinas/asterinas/blob/a4e808cc7edebcade78b28796202ef554199dde4/kernel/src/vm/page_cache/vmo/mod.rs#L593-L597

### To Reproduce

```c
/* SPDX-License-Identifier: MPL-2.0 */

#define _GNU_SOURCE

#include <unistd.h>
#include <sys/mman.h>
#include <sys/fcntl.h>

#include "../../common/test.h"

#define PAGE_SIZE 4096

#define TEST_FILE "/ext2/mmap_write"
#define TEST_DATA "abcdefg"

FN_TEST(write_from_mmap_same_page)
{
	char *buf;
	int fd;

	buf = TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE,
			     MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
	strcpy(buf, TEST_DATA);

	// Write a page to the disk via O_DIRECT.
	fd = TEST_SUCC(
		open(TEST_FILE, O_WRONLY | O_DIRECT | O_CREAT | O_TRUNC, 0644));
	TEST_RES(write(fd, buf, PAGE_SIZE), _ret == 4096);
	TEST_SUCC(close(fd));

	TEST_SUCC(munmap(buf, PAGE_SIZE));

	fd = TEST_SUCC(open(TEST_FILE, O_RDWR));
	buf = TEST_SUCC(mmap(NULL, PAGE_SIZE, PROT_READ,
			     MAP_SHARED, fd, 0));

	// This writes to a page in the page cache from the same page in the page table.
	// Therefore, if we try to lock both pages, a deadlock will occur.
	TEST_SUCC(write(fd, buf, PAGE_SIZE));

	TEST_SUCC(munmap(buf, PAGE_SIZE));
	TEST_SUCC(close(fd));

	TEST_SUCC(unlink(TEST_FILE));

}
END_TEST()
```

```
~ # /test/fs/ext2/mmap_write
test_write_from_mmap_same_page: `mmap(((void *)0), 4096, 0x1 | 0x2, 0x02 | 0x20, -1, 0)` passed [got Success]
[...]

PR #3421, opened 2026-06-22
Reverse mappings for file-backed memory mappings

### LINKED ISSUE #3295: Reverse mappings for file-backed `Vmo`s

lrh2000:
This PR adds reverse mappings for file-backed memory mappings.

### Commit "Make `PageCache` a unique handle"

`PageCache` is not long `Clone`able. It is a unique handle that locates behind the inode lock. Resizing requires `&mut self`.

Ideally, both `fill_zeros` and `write` would also require `&mut self`. However, this has been left as a TODO because the extfat implementations still need to be able to write without a mutable reference.

### Commit "Add basic reverse mapping implementation"

This adds `Rmap` and ensure that all shared file-backed mappings are recorded in it.

```rust
/// Reverse mappings from a [`Vmo`] to [`Vmar`]s.
///
/// [`Vmo`]: super::Vmo
pub struct Rmap {
    entries: BTreeMap<KeyableWeak<Vmar>, Vec<RmapEntry>>,
}
```

### Commit "Integrate reverse mappings to page cache"

This implements https://github.com/asterinas/asterinas/issues/3295. For more details, see the issue description.

Closes https://github.com/asterinas/asterinas/issues/3295.

### More tests.

I originally expected xfstests to contain a number of useful reverse mapping tests. However, according to https://github.com/asterinas/asterinas/issues/3142, there do not seem to be many.

I have enabled `generic/029`, but `generic/647` still needs to remain disabled due to #3257.

I have also added some regression tests but there are not many because I cannot figure out more. It seems that `sync()` and its interaction with the page table entries are difficult to test, and their effects are less observable from userspace.

### Concucrency tests.

As shown in #3295, the concurrency model is complex because of its various interactions with the page table, reverse mappings, and the inode. However, writing reasonable tests is difficult because it is hard to inject a page fault (e.g., lock the page table and call `Vmo::try_commit_page`) at the right time and customize it (e.g., in the middle of file operations such as `evict`, `sync`, `read`, `write`). Additionally, we lack the ability to create a `Vmar` in ktest.

I haven't found a solution yet, so this PR lacks concurrency tests.

### Documentations.

I added comments to various page cache operations to explain when to unmap pages and why releasing a lock is OK. I have also inserted the table in #3295 as module-level documentation. Is this enough or is this the best way to explain the concurrency model?

### Efficiency.

Unless there are trivial optimizations that I missed, I'd like to treat efficiency problems outside the scope of this PR. Having reverse mappings is essential for correctness, which always outweighs efficiency.

For example, we currently unmap the entire VMO offset range when evicting pages, regardless of whether the page itself is evicted or not. Dirty pages will always be kept, since losing data is not an option. We may want to optimize this in the future to unmap only clean pages.

### Private mappings.

Linux maintains reverse mappings, even for private ones. I am quite surprised that private mappings are also affected by file operations. Even after being COWed, a page in a private mapping will be unmapped if the corresponding file range is truncated.

However, I did not implement this behavior in this PR. Besides invoking additional complexities (e.g., a COWed page not being unmapped when the page is evicted from the page cache but only when the file range is truncated), our private mapping behavior is already broken, even without considering reverse mappings.

The following code, which existed before this PR (and this PR does not touch), is incorrect. If the file range is not present, we should trigger a SIGBUS, even for private mappings:
https://github.com/asterinas/asterinas/blob/e80904666fcc97085605f1e8118b86bc738c13c8/kernel/src/vm/vmar/vm_mapping.rs#L544-L547

In other words, if we don't trigger a SIGBUS when mapping a new page, then there is no reason to trigger one at file truncation either. So, I decided to leave this for future work.

### DIFF
--- a/kernel/src/fs/fs_impls/exfat/inode.rs
+++ b/kernel/src/fs/fs_impls/exfat/inode.rs
@@ -26,7 +26,10 @@ use super::{
 };
 use crate::{
     fs::{
-        exfat::{dentry::ExfatDentryIterator, fat::ExfatChain, fs::ExfatFs},
+        exfat::{
+            bitmap::ExfatBitmap, dentry::ExfatDentryIterator, fat::ExfatChain, fs::ExfatFs,
+            upcase_table::ExfatUpcaseTable,
+        },
         file::{InodeMode, InodeType, StatusFlags, mkmod},
         utils::DirentVisitor,
         vfs::{
@@ -37,10 +40,10 @@ use crate::{
     },
     prelude::*,
     process::{Gid, Uid},
-    vm::page_cache::{BlockAsPageCacheBackend, PageCache},
+    vm::page_cache::{BlockAsPageCacheBackend, PageCache, Vmo},
 };
 
-///Inode number
+/// Inode number
 pub type Ino = u64;
 
 bitflags! {
@@ -325,12 +328,14 @@ impl ExfatInodeInner {
         }
 
         let parent = self.get_parent_inode().unwrap_or_else(|| unimplemented!());
-        let page_cache = parent.page_cache().unwrap();
+        let parent_inner = parent.inner.read();
 
         // Need to read the latest dentry set from parent inode.
 
-        let mut dentry_set =
-            ExfatDentrySet::read_from(&page_cache, self.dentry_entry as usize * DENTRY_SIZE)?;
+        let mut dentry_set = ExfatDentrySet::read_from(
+            &parent_inner.page_cache,
+            self.dentry_entry as usize * DENTRY_SIZE,
+        )?;
 
         let mut file_dentry = dentry_set.get_file_dentry();
         let mut stream_dentry = dentry_set.get_stream_dentry();
@@ -364,9 +369,11 @@ impl ExfatInodeInner {
         let start_off = self.dentry_entry as usize * DENTRY_SIZE;
         let bytes = dentry_set.to_le_bytes();
 
-        page_cache.write_bytes(start_off, &bytes)?;
+        parent_inner.page_cache.write_bytes(start_off, &bytes)?;
         if sync {
-            page_cache.flush_range(start_off..start_off + bytes.len())?;
+            parent_inner
+                .page_cache
+                .flush_range(start_off..start_off + bytes.len())?;
         }
 
         Ok(())
@@ -836,7 +843,8 @@ impl ExfatInode {
         let fs = inner.fs();
         let fs_guard = fs.lock();
         let mut inner = self.inner.write();
-        inner.page_cache.resize(0, inner.size)?;
+        let old_size = inner.size;
+        inner.page_cache.resize(0, old_size)?;
         inner.resize(0, &fs_guard)?;
         Ok(())
     }
@@ -1037,6 +1045,37 @@ impl ExfatInode {
             parent_hash,
             fs_guard,
         )
+    }
+
+    pub(super) fn read_bitmap(&self) -> Result<ExfatBitmap> {
+        let inner = self.inner.read();
+        let dentry_iterator = ExfatDentryIterator::new(&inner.page_cache, 0, None)?;
+
+        for dentry_result in dentry_iterator {
+            let dentry = dentry_result?;
+            if let ExfatDentry::Bitmap(bitmap_dentry) = dentry {
+                // If the last bit is 0, it is a valid bitmap.
+                if (bitmap_dentry.flags & 0x1) == 0 {
+                    return ExfatBitmap::load_bitmap_from_dentry(inner.fs.clone(), &bitmap_dentry);
+                }
+            }
+        }
+
+        return_errno_with_message!(Errno::EINVAL, "bitmap not found")
+    }
+
+    pub(super) fn read_upcase_table(&self) -> Result<ExfatUpcaseTable> {
+        let inner = self.inner.read();
+        let dentry_iterator = ExfatDentryIterator::new(&inner.page_cache, 0, None)?;
+
+        for dentry_result in dentry_iterator {
+            let dentry = dentry_result?;
+            if let ExfatDentry::Upcase(upcase_dentry) = dentry {
+                return ExfatUpcaseTable::load_table_from_dentry(inner.fs.clone(), &upcase_dentry);
+            }
+        }
+
+        return_errno_with_message!(Errno::EINVAL, "upcase table not found")
     }
 
     /// Find empty dentry. If not found, expand the cluster chain.
@@ -1256,7 +1295,8 @@ impl ExfatInode {
         if delete_contents {
             if is_dir {
                 let mut inner = inode.inner.write();
-                inner.page_cache.resize(0, inner.size)?;
+                let old_size = inner.size;
+                inner.page_cache.resize(0, old_size)?;
                 inner.resize(0, fs_guard)?;
             }
             // Set the delete flag.
@@ -1515,8 +1555,8 @@ impl Inode for ExfatInode {
         self.inner.read().fs()
     }
 
-    fn page_cache(&self) -> Option<PageCache> {
-        Some(self.inner.read().page_cache.clone())
+    fn page_cache(&self) -> Option<Arc<Vmo>> {
+        Some(self.inner.read().page_cache.as_vmo().clone())
     }
 
     fn create(&self, name: &str, type_: InodeType, mode: InodeMode) -> Result<Arc<dyn Inode>> {
--- a/kernel/src/fs/fs_impls/overlayfs/fs.rs
+++ b/kernel/src/fs/fs_impls/overlayfs/fs.rs
@@ -35,7 +35,7 @@ use crate::{
     },
     prelude::*,
     process::{Gid, Uid},
-    vm::page_cache::PageCache,
+    vm::page_cache::Vmo,
 };
 
 const OVERLAY_FS_MAGIC: u64 = 0x794C7630;
@@ -489,7 +489,7 @@ impl OverlayInode {
         &self.extension
     }
 
-    pub fn page_cache(&self) -> Option<PageCache> {
+    pub fn page_cache(&self) -> Option<Arc<Vmo>> {
         let _ = self.get_top_valid_inode().page_cache()?;
         // Do copy-up for the potential memory mapping operations
         let upper = self.build_upper_recursively_if_needed().unwrap();
@@ -990,7 +990,7 @@ impl Inode for OverlayInode {
     fn set_mtime(&self, time: Duration);
     fn ctime(&self) -> Duration;
     fn set_ctime(&self, time: Duration);
-    fn page_cache(&self) -> Option<PageCache>;
+    fn page_cache(&self) -> Option<Arc<Vmo>>;
     fn create(&self, name: &str, type_: InodeType, mode: InodeMode) -> Result<Arc<dyn Inode>>;
     fn mknod(&self, name: &str, mode: InodeMode, type_: MknodType) -> Result<Arc<dyn Inode>>;
     fn open(
@@ -1242,7 +1242,7 @@ impl FsType for OverlayFsType {
 // TODO: Enrich the tests to cover more cases.
 #[cfg(ktest)]
 mod tests {
-    use ostd::{mm::VmIo, prelude::ktest};
+    use ostd::prelude::ktest;
 
     use super::*;
     use crate::fs::{
@@ -1578,7 +1578,10 @@ mod tests {
         let f1 = root.lookup("f1").unwrap();
         assert_eq!(f1.size(), 0);
         f1.resize(PAGE_SIZE).unwrap();
-        f1.page_cache().unwrap().write_val(0, &3u8).unwrap();
+        f1.page_cache()
+            .unwrap()
+            .write(0, &mut VmReader::from([3].as_slice()).to_fallible())
+            .unwrap();
         f1.set_atime(Duration::default());
         f1.sync_data().unwrap();
         let mut data = [0u8; 1];
--- a/kernel/src/fs/vfs/fs_apis/inode.rs
+++ b/kernel/src/fs/vfs/fs_apis/inode.rs
@@ -29,7 +29,7 @@ use crate::{
     },
     security::lsm::hooks as lsm_hooks,
     time::clocks::RealTimeCoarseClock,
-    vm::page_cache::PageCache,
+    vm::page_cache::Vmo,
 };
 
 #[derive(Clone, Copy, Debug)]
@@ -387,7 +387,7 @@ pub trait Inode: Any + FileOps + Send + Sync {
 
     fn set_ctime(&self, time: Duration);
 
-    fn page_cache(&self) -> Option<PageCache> {
+    fn page_cache(&self) -> Option<Arc<Vmo>> {
         None
     }
 
--- a/kernel/src/vm/page_cache/vmo/mod.rs
+++ b/kernel/src/vm/page_cache/vmo/mod.rs
@@ -591,7 +591,7 @@ impl Vmo {
                     Ok(written_size) => written_size,
                     Err((err, written_size)) => {
                         // If the page is not initialized, keep it as it is on a partial write.
-                        if written_size > 0 && locked_page.is_up_to_date() {
+                        if written_size > 0 && locked_page.state().is_up_to_date() {
                             locked_page.set_dirty();
                         }
                         return Err(Error::from(err));
@@ -702,8 +702,9 @@ impl<'a> BackedVmo<'a> {
         }
 
         let page_idx_range = get_page_idx_range(range);
-        let dirty_pages =
-            self.collect_pages_if(locked_pages, page_idx_range, |_, page| page.is_dirty());
+        let dirty_pages = self.collect_pages_if(locked_pages, page_idx_range, |_, page| {
+            page.state().is_dirty()
+        });
 
         let mut io_batch = IoBatch::with_capacity(dirty_pages.len());
         for (idx, page) in dirty_pages {
@@ -730,8 +731,9 @@ impl<'a> BackedVmo<'a> {
         }
 
         let page_idx_range = get_page_idx_range(range);
-        let pages_to_evict =
-            self.collect_pages_if(locked_pages, page_idx_range, |_, page| page.is_up_to_date());
+        let pages_to_evict = self.collect_pages_if(locked_pages, page_idx_range, |_, page| {
+            page.state().is_up_to_date()
+        });
         self.wait_for_writeback_and_remove_pages(pages_to_evict);
 
         Ok(())
@@ -806,7 +808,7 @@ impl<'a> BackedVmo<'a> {
         };
 
         // Check if the page is initialized.
-        if !commit_mode.skips_backend_read() && page.is_uninit() {
+        if !commit_mode.skips_backend_read() && page.state().is_uninit() {
             return Err(VmoCommitError::WaitUntilInit {
                 index: page_idx,
                 page: page.clone(),
--- a/kernel/src/vm/page_cache/vmo/mod.rs
+++ b/kernel/src/vm/page_cache/vmo/mod.rs
@@ -31,7 +31,10 @@ use xarray::{Cursor, LockedXArray, XArray};
 
 use crate::{
     prelude::*,
-    vm::page_cache::{CachePage, CachePageExt, PageCacheBackend},
+    vm::{
+        page_cache::{CachePage, CachePageExt, PageCacheBackend},
+        vmar::Rmap,
+    },
 };
 
 mod options;
@@ -131,6 +134,8 @@ pub struct Vmo {
     // not have the knowledge to determine if they belong to memfd. We may want to enhance
     // `VmoOptions` to make VMOs aware of whether its writable mappings should be tracked.
     pub(super) writable_mapping_status: WritableMappingStatus,
+    /// Reserve mappings.
+    pub(super) rmap: Mutex<Rmap>,
 }
 
 impl Debug for Vmo {
@@ -343,6 +348,18 @@ impl Vmo {
         // VMOs with a backend do not use this field.
         debug_assert!(!self.has_backend());
         &self.writable_mapping_status
+    }
+
+    /// Returns reverse mappings of the VMO.
+    ///
+    /// Holding either this lock or the VMAR lock ensures the stability of the
+    /// associated VM mappings. In other words, when adding, removing, or
+    /// moving a VM mapping, both this lock and the VMAR lock must be held.
+    ///
+    /// To avoid deadlocks, acquire this lock after the VMAR lock in cases both
+    /// locks need to be acquired.
+    pub fn rmap(&self) -> &Mutex<Rmap> {
+        &self.rmap
     }
 
     /// Decommits anonymous pages in the specified byte range.
--- a/kernel/src/vm/page_cache/vmo/mod.rs
+++ b/kernel/src/vm/page_cache/vmo/mod.rs
@@ -214,8 +214,9 @@ impl Vmo {
     ///
     /// For anonymous VMOs the page is zero-filled on first access.
     /// For VMOs with a backend this may perform synchronous I/O.
-    pub fn commit_on(&self, page_idx: usize) -> Result<CachePage> {
-        self.commit_on_internal(page_idx, CommitMode::Read)
+    pub fn commit_on(&self, page_idx: usize) -> Result<()> {
+        self.commit_on_internal(page_idx, CommitMode::Read)?;
+        Ok(())
     }
 
     fn commit_on_internal(&self, page_idx: usize, commit_mode: CommitMode) -> Result<CachePage> {
@@ -416,7 +417,8 @@ impl Vmo {
                 page_offset = 0;
             }
 
-            // `current_idx < page_idx_range.end` guarantees at least one page is successfully collected here.
+            // `current_idx < page_idx_range.end` guarantees at least one page is successfully
+            // collected here.
             current_idx = page_batch.last().unwrap().0 + 1;
         }
 
@@ -548,7 +550,8 @@ impl Vmo {
                 *page_offset = 0;
             }
 
-            // `current_idx < page_idx_range.end` guarantees at least one page is successfully collected here.
+            // `current_idx < page_idx_range.end` guarantees at least one page is successfully
+            // collected here.
             current_idx = page_batch.last().unwrap().0 + 1;
         }
 
@@ -573,7 +576,8 @@ impl Vmo {
         while current_idx < page_idx_range.end {
             self.collect_pages(current_idx, page_idx_range.end, commit_mode, page_batch)?;
 
-            // `current_idx < page_idx_range.end` guarantees at least one page is successfully collected here.
+            // `current_idx < page_idx_range.end` guarantees at least one page is successfully
+            // collected here.
             let next_idx = page_batch.last().unwrap().0 + 1;
 
             for (_, page) in page_batch.drain(..) {
--- a/kernel/src/vm/page_cache/vmo/mod.rs
+++ b/kernel/src/vm/page_cache/vmo/mod.rs
@@ -32,7 +32,7 @@ use xarray::{Cursor, LockedXArray, XArray};
 use crate::{
     prelude::*,
     vm::{
-        page_cache::{CachePage, CachePageExt, PageCacheBackend},
+        page_cache::{CachePage, CachePageExt, PageCacheBackend, cache_page::PageState},
         vmar::Rmap,
     },
 };
@@ -103,8 +103,11 @@ pub use options::VmoOptions;
 ///   materialized as zeros.
 /// - `Uninit -> Dirty`: a full-page overwrite commits and keeps the page
 ///   locked until the writer has copied its new contents.
-/// - `UpToDate -> Dirty`: buffered writes modify the page under the page lock.
+/// - `UpToDate -> Dirty`: buffered writes modify the page under the page lock,
+///   or the page is mapped writably to userspace.
 /// - `Dirty -> UpToDate`: the VMO writeback path hands the page to the backend.
+/// - `UpToDate -> Evicted`: the page is evicted and the page lock is held until
+///   the page is removed from the VMO.
 ///
 /// The auxiliary `is_writing_back` bit is set under the page lock, then cleared
 /// later by the BIO completion callback after the writeback state has been
@@ -214,13 +217,84 @@ impl CommitMode {
     }
 }
 
+/// A mode specified when committing a new [`Vmo`] page to the map in a page table.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VmoMapMode {
+    /// The page is private. It can be a snapshot.
+    Private,
+    /// The page is shared. It needs to be read.
+    SharedRead,
+    /// The page is shared. It needs to be read and written.
+    SharedWrite,
+}
+
+impl VmoMapMode {
+    /// Ensures that the page can be mapped to a page table.
+    fn ensure(self, page: CachePage) {
+        match page.state() {
+            PageState::UpToDate if self.is_write() => {
+                let locked_page = page.lock_guard();
+                if locked_page.state().is_up_to_date() {
+                    locked_page.set_dirty();
+                }
+            }
+            PageState::UpToDate | PageState::Dirty => (),
+            PageState::Evicted if !self.is_shared() => (),
+            state @ (PageState::Uninit | PageState::Evicted) => {
+                // We shouldn't reach this point if the page is not initialized. `CommitMode::Read`
+                // should have been specified earlier to either read the page from the backend
+                // storage or report an error if reading fails.
+                debug_assert_eq!(state, PageState::Evicted);
+                // The page may be evicted. It still contains valid data. However, it cannot be
+                // mapped to a page table. We acquire the page lock here so that, when the caller
+                // retries, it won't see this page anymore.
+                let _ = page.lock_guard();
+            }
+        }
+    }
+
+    /// Tries to ensure that the page can be mapped to a page table without blocking.
+    ///
+    /// On success, returns a mode that satisfies `self` but may grant the caller more permissions
+    /// when mapping `page`. See also the "Map modes" section in [`Vmo::try_commit_page`].
+    fn try_ensure(self, page_idx: usize, page: &CachePage) -> Result<Self, VmoCommitError> {
+        match page.state() {
+            PageState::UpToDate if self.is_write() => {
+                if let Some(locked_page) = page.try_lock_guard()
+                    && locked_page.state().is_up_to_date()
+                {
+                    locked_page.set_dirty();
+                    Ok(Self::SharedWrite)
+                } else {
+                    Err(VmoCommitError::NeedIo { index: page_idx })
+                }
+            }
+            PageState::UpToDate => Ok(Self::SharedRead),
+            PageState::Dirty => Ok(Self::SharedWrite),
+            PageState::Evicted if !self.is_shared() => Ok(Self::Private),
+            PageState::Uninit | PageState::Evicted => {
+                Err(VmoCommitError::NeedIo { index: page_idx })
+            }
+        }
+    }
+
+    fn is_write(self) -> bool {
+        matches!(self, Self::SharedWrite)
+    }
+
+    fn is_shared(self) -> bool {
+        matches!(self, Self::SharedRead | Self::SharedWrite)
+    }
+}
+
 impl Vmo {
     /// Commits the page at `page_idx`, blocking on backend I/O if required.
     ///
     /// For anonymous VMOs the page is zero-filled on first access.
     /// For VMOs with a backend this may perform synchronous I/O.
-    pub fn commit_on(&self, page_idx: usize) -> Result<()> {
-        self.commit_on_internal(page_idx, CommitMode::Read)?;
+    pub fn commit_on(&self, page_idx: usize, map_mode: VmoMapMode) -> Result<()> {
+        let page = self.commit_on_internal(page_idx, CommitMode::Read)?;
+        map_mode.ensure(page);
         Ok(())
     }
 
@@ -252,7 +326,28 @@ impl Vmo {
     ///
     /// If the commit operation needs to perform I/O, it will return a
     /// [`VmoCommitError::NeedIo`].
-    pub fn try_commit_page(&self, offset: usize) -> Result<CachePage, VmoCommitError> {
+    ///
+    /// If the VMO corresponds to a page cache, callers must hold the page table
+    /// lock and insert the committed page atomically to the page table after
+    /// this method returns.
+    ///
+    /// # Map modes
+    ///
+    /// `map_mode` specifies the requested mapping mode for the page in the page
+    /// table. If successful, this method returns another [`VmoMapMode`] that
+    /// grants at least the requested mode, but may grant additional permissions
+    /// according to the page state. Callers can use this information to prevent
+    /// some future page faults.
+    ///
+    /// For example, if the page is already dirty, mapping it as writable is
+    /// permitted even if only read access is required in `map_mode`. In this
+    /// case, the requested mode (`map_mode`) is [`VmoMapMode::SharedRead`] but
+    /// the granted mode (the return value) is [`VmoMapMode::SharedWrite`].
+    pub fn try_commit_page(
+        &self,
+        offset: usize,
+        map_mode: VmoMapMode,
+    ) -> Result<(CachePage, VmoMapMode), VmoCommitError> {
         let page_idx = offset / PAGE_SIZE;
         if offset >= self.size() {
             return Err(VmoCommitError::Err(Error::with_message(
@@ -261,10 +356,14 @@ impl Vmo {
             )));
         }
 
-        let guard = disable_preempt();
-        let mut cursor = self.pages.cursor(&guard, page_idx as u64);
-        self.try_commit_with_cursor(&mut cursor, CommitMode::Read)
-            .map(|(_, page)| page)
+        let (_, page) = {
+            let guard = disable_preempt();
+            let mut cursor = self.pages.cursor(&guard, page_idx as u64);
+            // `VmoMapMode::try_ensure` checks if the page has been initialized.
+            self.try_commit_with_cursor(&mut cursor, CommitMode::Overwrite)?
+        };
+        let mode = map_mode.try_ensure(page_idx, &page)?;
+        Ok((page, mode))
     }
 
     fn try_commit_with_cursor(
@@ -293,17 +392,34 @@ impl Vmo {
     ///
     /// Once a commit operation needs to perform I/O, it will return a
     /// [`VmoCommitError::NeedIo`].
+    ///
+    /// If the VMO corresponds to a page cache, callers must hold the page table
+    /// lock and insert the committed page atomically to the page table inside
+    /// the operation.
     pub fn try_operate_on_range<F>(
         &self,
         range: &Range<usize>,
-        operate: F,
+        mut operate: F,
+        map_mode: VmoMapMode,
     ) -> Result<(), VmoCommitError>
     where
         F: FnMut(
             &mut dyn FnMut() -> Result<(usize, CachePage), VmoCommitError>,
         ) -> Result<(), VmoCommitError>,
     {
-        self.try_operate_on_range_internal(range, operate, CommitMode::Read)
+        self.try_operate_on_range_internal(
+            range,
+            |commit_fn| {
+                let mut new_commit_fn = || {
+                    let (idx, page) = commit_fn()?;
+                    map_mode.try_ensure(idx, &page)?;
+                    Ok((idx, page))
+                };
+                operate(&mut new_commit_fn)
+            },
+            // `VmoMapMode::try_ensure` checks if the page has been initialized.
+            CommitMode::Overwrite,
+        )
     }
 
     fn try_operate_on_range_internal<F>(
@@ -365,6 +481,7 @@ impl Vmo {
     /// Decommits anonymous pages in the specified byte range.
     pub(super) fn decommit_anon_pages(
         &self,
+        mut locked_rmap: MutexGuard<Rmap>,
         mut locked_pages: LockedXArray<CachePage>,
         range: Range<usize>,
     ) -> Result<()> {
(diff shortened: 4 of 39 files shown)


kernel/src/fs/fs_impls/exfat/inode.rs
```diff
@@ -37,7 +40,7 @@ use crate::{
     },
     prelude::*,
     process::{Gid, Uid},
-    vm::page_cache::{BlockAsPageCacheBackend, PageCache},
+    vm::page_cache::{BlockAsPageCacheBackend, PageCache, Vmo},
 };
 
 ///Inode number
```
junyang-zh: Missed a blank space here. It's close, so maybe fix it as well.

kernel/src/fs/fs_impls/overlayfs/fs.rs
```diff
@@ -1563,7 +1563,10 @@ mod tests {
         let f1 = root.lookup("f1").unwrap();
         assert_eq!(f1.size(), 0);
         f1.resize(PAGE_SIZE).unwrap();
-        f1.page_cache().unwrap().write_val(0, &3u8).unwrap();
+        f1.page_cache()
+            .unwrap()
+            .write(0, &mut VmReader::from([3].as_slice()).to_fallible())
+            .unwrap();
```
junyang-zh, line 1584: Didn't get why `write_val` doesn't work for `Arc<Vmo>`. Anything missing?

kernel/src/vm/vmar/vmar_impls/fork.rs
```diff
@@ -48,12 +51,12 @@ impl Vmar {
                 let num_copied =
                     cow_copy_pt(&mut cur_cursor, &mut new_cursor, vm_mapping.map_size());
 
+                cur_cursor.flusher().issue_tlb_flush(TlbFlushOp::for_all());
+                cur_cursor.flusher().dispatch_tlb_flush();
+                cur_cursor.flusher().sync_tlb_flush();
```
junyang-zh: Seems a lot of redundant synced flushes are performed here. One possible optimization would be to flush only with the last cursor. Here, the flush is for CoW. The only thing to ensure before exiting this function is that all the parent threads see a write-protected page table. Missing flushes of the former `VmMapping`s only indicate that there are some CPUs that no longer activate this `VmSpace`, which is fine. So flushing only with the last cursor ensures that all threads see the protected page table.

junyang-zh (commented):
I read the previous 5 commits and skimmed the last two. Will come back later. Thanks for implementing this! Liked it.

kernel/src/fs/fs_impls/overlayfs/fs.rs
lrh2000: `Vmo` does not implement `VmIo`? It is only implemented for `PageCache`. This is not introduced in this PR anyway.

In general, writing a value to a Vmo seems much more dangerous as it can easily cause lock misuse. Therefore, making it slightly inconvenient seems fine.

kernel/src/vm/vmar/vmar_impls/fork.rs
lrh2000: Thanks. This is a good catch! The previous way did cause too many (unnecessary) TLB flushes..

kernel/src/vm/vmar/rmap.rs
```diff
@@ -0,0 +1,164 @@
...
+        let Entry::Occupied(mut map_entry) = self.entries.entry(key) else {
+            panic!("the entry to remove does not exist")
+        };
+
+        let entries = map_entry.get_mut();
+        let index = entries
+            .iter()
+            .position(|entry| entry.vaddr == vaddr)
+            .expect("the entry to remove does not exist");
+        entries.swap_remove(index);
```
junyang-zh, line 65: When the entries vector becomes empty, we do not seem to remove the `Vmar` key. This should be a leakage.

kernel/src/vm/vmar/rmap.rs
```diff
@@ -0,0 +1,164 @@
...
+                    .unwrap();
+                rss_delta.add(
+                    RssType::File,
+                    -(cursor_mut.unmap(addr_range.len()) as isize),
+                );
+                cursor_mut.flusher().dispatch_tlb_flush();
+                cursor_mut.flusher().sync_tlb_flush();
+            }
+
+            true
```
junyang-zh, line 111: Here too. We seem not to have removed any keys. Any concerns?

lrh2000: `Rmap::unmap` unmaps pages from the page table. The mapping should still exist. I don't think we need to remove any keys here?

lrh2000:
Here is the range-diff (rebase + address the comments above):
<details>
<summary> range-diff </summary>

Command: `git range-diff a24c3ea7534d22425c17693c518f2d1aaffad475..bf7fe7f993a2c635ce74e5d5faba05e77afc0d89 main..HEAD`

Output:
```
1:  2135011d4 ! 1:  9ade87123 Simplify some VMO APIs
    @@ Metadata
     Author: Ruihan Li <lrh2000@pku.edu.cn>
     
      ## Commit message ##
    -    Simplify some VMO APIs
    +    Simplify VMO APIs and clean up imports
     
      ## kernel/src/vm/page_cache/vmo/mod.rs ##
     @@ kernel/src/vm/page_cache/vmo/mod.rs: impl Vmo {
    @@ kernel/src/vm/page_cache/vmo/mod.rs: impl Vmo {
      
          fn commit_on_internal(&self, page_idx: usize, commit_mode: CommitMode) -> Result<CachePage> {
     
    + ## kernel/src/vm/vmar/mod.rs ##
    +@@ kernel/src/vm/vmar/mod.rs: use ostd::mm::Vaddr;
    + pub use self::{
    +     handle::VmarHandle,
    +     vmar_impls::{
    +-        RemapOldMappingAction, RssType, Vmar, map::VmarMapOffset, page_fault::PageFaultInfo,
    ++        RssType, Vmar, map::VmarMapOffset, page_fault::PageFaultInfo, remap::RemapOldMappingAction,
    +     },
    + };
    + 
    +
      ## kernel/src/vm/vmar/vm_mapping.rs ##
     @@ kernel/src/vm/vmar/vm_mapping.rs: impl MappedVmo {
          ///
    @@ kernel/src/vm/vmar/vm_mapping.rs: impl MappedVmo {
          }
      
          /// Traverses the indices within a specified range of a VMO sequentially.
    +
    + ## kernel/src/vm/vmar/vmar_impls/mod.rs ##
    +@@ kernel/src/vm/vmar/vmar_impls/mod.rs: pub(super) mod map;
    + pub(super) mod page_fault;
    + mod protect;
    + mod query;
    +-mod remap;
    ++pub(super) mod remap;
    + mod unmap;
    ++
    + use core::{
    +     array,
    +     ops::Range,
    +@@ kernel/src/vm/vmar/vmar_impls/mod.rs: use core::{
    + use align_ext::AlignExt;
    + use aster_util::per_cpu_counter::PerCpuCounter;
    + use ostd::{cpu::CpuId, mm::VmSpace};
    +-pub use remap::RemapOldMappingAction;
    + 
    + use super::{
    +     VMAR_CAP_ADDR, VMAR_LOWEST_ADDR,
2:  d0cd7fd34 = 2:  b1234c973 Make `PageCache` a unique handle
3:  5cbc00c8f ! 3:  29dbbe234 Update FS impls to new `PageCache` APIs
    @@ kernel/src/fs/fs_impls/exfat/inode.rs: use crate::{
     +    vm::page_cache::{BlockAsPageCacheBackend, PageCache, Vmo},
      };
      
    - ///Inode number
    +-///Inode number
    ++/// Inode number
    + pub type Ino = u64;
    + 
    + bitflags! {
     @@ kernel/src/fs/fs_impls/exfat/inode.rs: impl ExfatInodeInner {
              }
      
    @@ kernel/src/fs/fs_impls/overlayfs/fs.rs: mod tests {
     
      ## kernel/src/fs/fs_impls/ramfs/fs.rs ##
     @@ kernel/src/fs/fs_impls/ramfs/fs.rs: use crate::{
    -     prelude::*,
    -     process::{Gid, Uid},
    +     process::{Gid, Uid, posix_thread::AsPosixThread},
    +     thread::Thread,
          time::clocks::RealTimeCoarseClock,
     -    vm::page_cache::PageCache,
     +    vm::page_cache::{PageCache, Vmo},
4:  31ba421fe = 4:  e4bf36879 Check page states on `PageState`
5:  69627c2d3 ! 5:  e8762f3c5 Add basic reverse mapping implementation
    @@ kernel/src/vm/page_cache/vmo/options.rs: fn alloc_vmo(
      }
      
     
    + ## kernel/src/vm/vmar/interval_set.rs ##
    +@@ kernel/src/vm/vmar/interval_set.rs: where
    +         }
    +     }
    + 
    ++    /// Returns the number of elements in the interval set.
    ++    pub fn len(&self) -> usize {
    ++        self.btree.len()
    ++    }
    ++
    +     /// Inserts an interval item into the interval set.
    +     pub fn insert(&mut self, item: V) {
    +         let range = item.range();
    +
      ## kernel/src/vm/vmar/mod.rs ##
     @@
      
    @@ kernel/src/vm/vmar/mod.rs: use ostd::mm::Vaddr;
      pub use self::{
          handle::VmarHandle,
     +    rmap::{Rmap, RmapEntry},
    -     vmar_impls::{RssType, Vmar, map::VmarMapOffset, page_fault::PageFaultInfo},
    - };
    - 
    +     vmar_impls::{
    +         RssType, Vmar, map::VmarMapOffset, page_fault::PageFaultInfo, remap::RemapOldMappingAction,
    +     },
     
      ## kernel/src/vm/vmar/rmap.rs (new) ##
     @@
    @@ kernel/src/vm/vmar/rmap.rs (new)
     +            .position(|entry| entry.vaddr == vaddr)
     +            .expect("the entry to remove does not exist");
     +        entries.swap_remove(index);
    ++
    ++        if entries.is_empty() {
    ++            map_entry.remove();
    ++        }
     +    }
     +
     +    /// Iterates over all reverse mappings and unmaps the given offset range.
    @@ kernel/src/vm/vmar/vmar_impls/fork.rs: impl Vmar {
     -            let cur_vmspace = vmar.vm_space();
     -            let mut cur_cursor = cur_vmspace.cursor_mut(&preempt_guard, &range).unwrap();
                  let mut rss_delta = RssDelta::new(&new_vmar);
    ++            let mut remaining = inner.vm_mappings.len();
      
     +            // Clone mappings.
                  for vm_mapping in inner.vm_mappings.iter() {
    ++                remaining -= 1;
    ++
                      let base = vm_mapping.map_to_addr();
     +                let range = base..vm_mapping.map_end();
     +
    @@ kernel/src/vm/vmar/vmar_impls/fork.rs: impl Vmar {
                      let num_copied =
                          cow_copy_pt(&mut cur_cursor, &mut new_cursor, vm_mapping.map_size());
      
    -+                cur_cursor.flusher().issue_tlb_flush(TlbFlushOp::for_all());
    -+                cur_cursor.flusher().dispatch_tlb_flush();
    -+                cur_cursor.flusher().sync_tlb_flush();
    ++                // We need to ensure that no writes can be performed to COW
    ++                // pages only after `fork()` returns. So we can perform a full
    ++                // TLB flush only when handling the last mapping.
    ++                if remaining == 0 {
    ++                    cur_cursor.flusher().issue_tlb_flush(TlbFlushOp::for_all());
    ++                    cur_cursor.flusher().dispatch_tlb_flush();
    ++                    cur_cursor.flusher().sync_tlb_flush();
    ++                }
     +
                      rss_delta.add(vm_mapping.rss_type(), num_copied as isize);
                  }
    @@ kernel/src/vm/vmar/vmar_impls/mod.rs: use aster_util::per_cpu_counter::PerCpuCou
     +    vm::page_cache::Vmo,
      };
      
    - /// The VMAR (used to be Virtual Memory Address Region, but now an orphan
    + /// The upper bound for allocations under the `MAP_32BIT` mmap flag.
     @@ kernel/src/vm/vmar/vmar_impls/mod.rs: pub struct Vmar {
          process_vm: ProcessVm,
          /// The number of handles that this `Vmar` has (see [`super::VmarHandle`])
    @@ kernel/src/vm/vmar/vmar_impls/remap.rs: impl Vmar {
     -            )?
     +            inner.alloc_free_region_exact_truncate(self, new_addr, new_size, &mut rss_delta)?
              } else {
    -             debug_assert!(new_size > old_size);
    - 
    +             // Fast path: expand the old mapping in place to the new size.
    +             // Skip when `action` is `RemapOldMappingAction::Keep` since we must
     @@ kernel/src/vm/vmar/vmar_impls/remap.rs: impl Vmar {
                          .is_ok()
                  {
    @@ kernel/src/vm/vmar/vmar_impls/remap.rs: impl Vmar {
                  }
      
     @@ kernel/src/vm/vmar/vmar_impls/remap.rs: impl Vmar {
    -         };
      
    -         // Create a new `VmMapping`.
    -+        let old_mapping_addr = inner.check_lies_in_single_mapping(old_addr, old_size)?;
    -+        let (vm_mapping, rmap_to_remove) = inner.remove(&old_mapping_addr).unwrap();
    -+        let mut rmap = rmap_to_remove.remove(self, old_mapping_addr);
    -         let old_mapping = {
    --            let old_mapping_addr = inner.check_lies_in_single_mapping(old_addr, old_size)?;
    +         // Create a new `VmMapping` at the target address.
    +         let old_mapping_addr = inner.check_lies_in_single_mapping(old_addr, old_size)?;
    +-        let new_mapping = if action == RemapOldMappingAction::Unmap {
     -            let vm_mapping = inner.remove(&old_mapping_addr).unwrap();
    ++        let mut _rmap_to_remove = super::RmapToRemove::new(None); // Suppress `unused_assignments`.
    ++        let mut _vmo_for_rmap = None; // Suppress `unused_assignments`.
    ++        let (new_mapping, mut rmap) = if action == RemapOldMappingAction::Unmap {
    ++            let (vm_mapping, rmap_to_remove) = inner.remove(&old_mapping_addr).unwrap();
    ++            _rmap_to_remove = rmap_to_remove;
    ++            let mut rmap = _rmap_to_remove.remove(self, old_mapping_addr);
                  let (left, old_mapping, right) = vm_mapping.split_range(&old_range);
                  if let Some(left) = left {
     -                inner.insert_without_try_merge(left);
    @@ kernel/src/vm/vmar/vmar_impls/remap.rs: impl Vmar {
     -                inner.insert_without_try_merge(right);
     +                inner.insert_without_try_merge(self, right, rmap.as_deref_mut());
                  }
    -             old_mapping
    +             // Note that we have ensured that `new_size >= old_size` at the beginning.
    +-            old_mapping.clone_for_remap_at(new_range.start)
    ++            (old_mapping.clone_for_remap_at(new_range.start), rmap)
    +         } else {
    +             let old_mapping = inner.vm_mappings.find_one(&old_mapping_addr).unwrap();
    +-            old_mapping.clone_range_for_remap_at(old_addr..old_addr + old_size, new_range.start)
    ++            _vmo_for_rmap = old_mapping.vmo_for_rmap().cloned();
    ++            (
    ++                old_mapping
    ++                    .clone_range_for_remap_at(old_addr..old_addr + old_size, new_range.start),
    ++                _vmo_for_rmap.as_ref().map(|vmo| vmo.rmap().lock()),
    ++            )
              };
    -         // Note that we have ensured that `new_size >= old_size` at the beginning.
    --        let new_mapping = old_mapping.clone_for_remap_at(new_range.start);
    +-
     -        inner.insert_try_merge(new_mapping.enlarge(new_size - old_size));
    -+        let new_mapping = old_mapping
    -+            .clone_for_remap_at(new_range.start)
    -+            .enlarge(new_size - old_size);
    -+        inner.insert_try_merge(self, new_mapping, rmap.as_deref_mut());
    ++        inner.insert_try_merge(
    ++            self,
    ++            new_mapping.enlarge(new_size - old_size),
    ++            rmap.as_deref_mut(),
    ++        );
      
              let preempt_guard = disable_preempt();
              let total_range = old_range.start.min(new_range.start)..old_range.end.max(new_range.end);
    @@ kernel/src/vm/vmar/vmar_impls/remap.rs: impl Vmar {
     +            .unwrap();
      
              // Move the mapping.
    ++        // We must hold the reverse mapping lock to avoid race conditions with
    ++        // unmapping from reverse mappings.
              let mut current_offset = 0;
    +         while current_offset < old_size {
    +             cursor.jump(old_range.start + current_offset).unwrap();
     
      ## kernel/src/vm/vmar/vmar_impls/unmap.rs ##
     @@ kernel/src/vm/vmar/vmar_impls/unmap.rs: impl Vmar {
6:  3fae1fb4a = 6:  1eb789137 Integrate reverse mappings to page cache
7:  bf7fe7f99 ! 7:  5b34826f2 Update page cache tests
    @@ test/initramfs/src/conformance/xfstests/short.list: generic/013
      generic/035
      generic/069
     
    - ## test/initramfs/src/regression/fs/ext2/xattr.c ##
    -@@ test/initramfs/src/regression/fs/ext2/xattr.c: FN_TEST(xattr_set_get_roundtrip)
    - }
    - END_TEST()
    - 
    --// TODO: xattr_list test is disabled because sys_listxattr only lists one
    --// namespace (Trusted when running as root), so user.* xattrs are invisible.
    --// This is a VFS-level bug, not ext2-specific.
    -+// Currently, the `listxattr()` syscall in Asterinas only lists xattrs within
    -+// one namespace ("Trusted" when running as root), so `user.*` xattrs are
    -+// invisible. This is a VFS-level bug, not ext2-specific.
    -+//
    -+// TODO: Add some tests for xattr listing after this bug has been fixed.
    - 
    - FN_TEST(xattr_remove_enodata)
    - {
    -
      ## test/initramfs/src/regression/memory/mmap/rev_map_common.h (new) ##
     @@
```
</details>

junyang-zh (approved):
LGTM! I don't have further issues. Please rebase the changes and let @cchanging take a look :>

kernel/src/vm/page_cache/vmo/mod.rs
```diff
@@ -209,13 +217,82 @@ impl CommitMode {
...
+                }
+            }
+            PageState::UpToDate => Ok(Self::SharedRead),
+            PageState::Dirty => Ok(Self::SharedWrite),
+            PageState::Evicted if !self.is_shared() => Ok(Self::Private),
+            PageState::Uninit | PageState::Evicted => {
+                Err(VmoCommitError::NeedIo { index: page_idx })
+            }
+        }
+    }
```
cchanging, line 279: Thank you for implementing the reverse mapping mechanism, the overall implementation looks great! Yet I may still need some extra time to fully go through some of the details in this PR.

From my quick pass so far, the main thing I’m confused about is this API — more specifically, its return value. Based on the API documentation and signature, I find it a bit hard to infer the intended semantics of the return value. 

I looked at the related usage, and it seems that it is only used in `VmMapping::prepare_page`, where, for shared mappings, if the page fault is not a write fault and the corresponding page is not dirty, the code avoids setting write permission in the page table for now? I’m not sure whether I’ve understood that correctly.

Anyway, I feel this part is currently a bit hard to follow when reading, and it might benefit from some additional comments to clarify the semantics. For example, maybe it would help to explicitly mention here that the return value indicates the recommended mapping mode?

lrh2000: > if the page fault is not a write fault and the corresponding page is not dirty, the code avoids setting write permission in the page table for now
    
This is correct. However, it does not seem related to having a mode as a return value.

Suppose this method returns a `Result<()>` without the mode type:
 - If the page fault is not a write fault, we request the page from the VMO in `SharedRead` mode. If the VMO approves, we can map the page with read-only permissions. 
 - We cannot map the page with read-write permissions because we never request such from the VMO!
 - But what if the page is already a dirty page? We can map it with read-write permissions without any cost. We cannot do so just because we don't know if it is actually allowed.

This can be fixed by returning the granted mode. The request mode is `SharedRead`, but the granted mode can be `SharedWrite`:
 - If the page fault is not a write fault and the corresponding page is dirty, the code will allow setting write permission in the page table (as long as the memory mapping is also a writable mapping).

How about revising the comments using the following patch:
<details>
<summary> Revision </summary>

```patch
diff --git a/kernel/src/vm/page_cache/vmo/mod.rs b/kernel/src/vm/page_cache/vmo/mod.rs
index 12b8d1806..a4d7c27a6 100644
--- a/kernel/src/vm/page_cache/vmo/mod.rs
+++ b/kernel/src/vm/page_cache/vmo/mod.rs
@@ -254,6 +254,9 @@ impl VmoMapMode {
     }
 
     /// Tries to ensure that the page can be mapped to a page table without blocking.
+    ///
+    /// On success, returns a mode that satisfies `self` but may grant the caller more permissions
+    /// when mapping `page`. See also the "Map modes" section in [`Vmo::try_commit_page`].
     fn try_ensure(self, page_idx: usize, page: &CachePage) -> Result<Self, VmoCommitError> {
         match page.state() {
             PageState::UpToDate if self.is_write() => {
@@ -327,6 +330,19 @@ impl Vmo {
     /// If the VMO corresponds to a page cache, callers must hold the page table
     /// lock and insert the committed page atomically to the page table after
     /// this method returns.
+    ///
+    /// # Map modes
+    ///
+    /// `map_mode` specifies the requested mapping mode for the page in the page
+    /// table. If successful, this method returns another [`VmoMapMode`] that
+    /// grants at least the requested mode, but may grant additional permissions
+    /// according to the page state. Callers can use this information to prevent
+    /// some future page faults.
+    ///
+    /// For example, if the page is already dirty, mapping it as writable is
+    /// permitted even if only read access is required in `map_mode`. In this
+    /// case, the requested mode (`map_mode`) is [`VmoMapMode::SharedRead`] but
+    /// the granted mode (the return value) is [`VmoMapMode::SharedWrite`].
     pub fn try_commit_page(
         &self,
         offset: usize,
```
</details>

kernel/src/vm/vmar/vm_mapping.rs
```diff
@@ -174,6 +176,28 @@ impl VmMapping {
...
+    ///
+    /// This method will return `Some(_)` if this mapping is shared and
+    /// VMO-backed.
+    //
+    // FIXME: According to Linux behavior, private mappings and COWed pages
+    // are also affected by file operations (e.g., truncation). So reverse
+    // mappings should also include them. We don't do this for now, as the
+    // man pages say that private mappings have unspecified behavior if the
+    // file content changes later on.
+    pub(super) fn vmo_for_rmap(&self) -> Option<&Arc<Vmo>> {
```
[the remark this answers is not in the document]
lrh2000: This is outside the scope of this PR. I discussed this in the PR description.

The correct fix is not trivial. The suggested fix above ("include private file-backed mappings in reverse mapping") cannot make the regression tests pass. We have to resolve some other existing issues to ensure that the regression tests pass in both Linux and Asterinas. Let's defer this to a future PR.

kernel/src/fs/vfs/fs_apis/inode.rs
```diff
@@ -387,7 +387,7 @@ pub trait Inode: Any + FileOps + Send + Sync {
 
     fn set_ctime(&self, time: Duration);
 
-    fn page_cache(&self) -> Option<PageCache> {
+    fn page_cache(&self) -> Option<Arc<Vmo>> {
```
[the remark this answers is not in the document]
lrh2000: I don't know if I should rename this method. After all, this method can only return the VMO behind the page cache.

kernel/src/vm/page_cache/vmo/mod.rs
```diff
@@ -689,46 +841,142 @@ pub struct BackedVmo<'a> {
     backend: Arc<dyn PageCacheBackend>,
 }
 
+enum CollectPage {
```
cchanging: This name doesn’t really sound like a condition. Would it be better to rename it to something like `PageSelection`?

kernel/src/vm/vmar/vmar_impls/mod.rs
```diff
@@ -167,6 +170,27 @@ impl Drop for RssDelta<'_> {
...
+impl RmapToRemove {
+    pub(self) fn new(vmo: Option<Arc<Vmo>>) -> Self {
+        Self { vmo }
+    }
+
+    pub(self) fn remove(&self, vmar: &Vmar, addr: Vaddr) -> Option<MutexGuard<'_, Rmap>> {
+        let mut rmap = self.vmo.as_ref()?.rmap().lock();
+        rmap.remove(vmar.weak_self.clone(), addr);
+        Some(rmap)
+    }
```
cchanging: Should this API have consume semantics? From the signature alone, it feels easy to misuse by calling it repeatedly and causing a deadlock. At the very least, should `&self` be changed to `&mut self`?

kernel/src/vm/page_cache/vmo/mod.rs
cchanging: This change makes sense, The semantics of returning a possibly elevated permission here feels fairly natural.

cchanging (commented):
I took a fairly careful look through the logic and didn’t find any obvious potential correctness issues. The only thing I noticed is that the direct write path seems to be missing some of the extra handling Linux has — for example, checking for re-dirty cases during invalidation. But as you also pointed out earlier when discussing the design, Linux itself does not really define the behavior of direct I/O, and those additional Linux-side checks do not fundamentally solve the problem either. So I think some inconsistency here is acceptable.

What I’m more concerned about is the potential performance impact of using a single `Mutex` for rmap in place of Linux’s separate rmap lock and invalidate lock, especially since Linux’s invalidate lock is an rwlock. Although you mentioned in the PR description that efficiency is not a goal of this PR, I’m not sure whether you’ve considered this part specifically.

One fairly obvious example is that when inserting a new page into the page cache, Linux only takes the read side of the invalidate lock, whereas we currently take the `Mutex`. It feels like this could cause some performance issues in workloads that concurrently read and write different offsets of a large file. Is this something you plan to optimize further in follow-up work?

lrh2000:
> One fairly obvious example is that when inserting a new page into the page cache, Linux only takes the read side of the invalidate lock, whereas we currently take the `Mutex`.

Alternatively, we can make the reverse mapping lock a `RwMutex` and only take a read lock when inserting a new page. As long as we require all other operations to take a write lock (including `unmap(&mut self)` and `freeze(&mut self)`. This seems correct.

As for splitting the reverse mapping lock and the invalidation lock, I can't see why we need to do that at the moment. Maybe you have some input?

> Is this something you plan to optimize further in follow-up work?

I don't have a specific plan at the moment. My main motivation is ensuring correctness. I also need (part of) reverse mapping implementations for my research project, so I'm just tidying up the work for integration into the main line.

kernel/src/vm/vmar/vmar_impls/mod.rs
lrh2000: > Should this API have consume semantics?

It cannot. Otherwise, the lifetime in `MutexGuard<'_, Rmap>` does not live long enough because the value has been moved.

> At the very least, should `&self` be changed to `&mut self`?

I suppose this is fine. Thanks for the suggestion.

cchanging:
> Alternatively, we can make the reverse mapping lock a RwMutex and only take a read lock when inserting a new page. As long as we require all other operations to take a write lock (including unmap(&mut self) and freeze(&mut self). This seems correct.

Yeah, I think this way is better. It shouldn’t involve major changes, but it can help avoid some performance bottlenecks in concurrent scenarios.

> As for splitting the reverse mapping lock and the invalidation lock, I can't see why we need to do that at the moment. Maybe you have some input?

I also haven’t found a strong reason to split it into two locks at this point, so I only mentioned cases where rw_mutex and mutex might lead to performance differences. In fact, I previously spent some time reviewing the relevant Linux code and also had an LLM analyze whether merging the two locks could cause performance or correctness issues, but I didn’t find any typical cases. In many scenarios, you need to hold both locks continuously anyway, which even feels like it could introduce extra overhead.

I think we can wait to split them until we see a stronger need for it later. That kind of split probably wouldn’t have much impact on the current code structure anyway.

lrh2000:
I just thought about this a bit more. I realized that locking the reverse mapping lock when committing a new page might not be necessary in our current implementation.

In this PR, the reverse mapping lock is held while committing a new page.
https://github.com/asterinas/asterinas/blob/f50256436cf193544e6ea4821cea942e07c31402/kernel/src/vm/page_cache/vmo/mod.rs#L1024-L1025

This is only for:
https://github.com/asterinas/asterinas/blob/f50256436cf193544e6ea4821cea942e07c31402/kernel/src/vm/page_cache/vmo/mod.rs#L998-L1001


However, upon further review, I think it is unnecessary. Even without the reverse mapping lock, one of the following cases must happen for concurrent page faults:
 1. Get the page and atomically put it in the page table. This can only happen before `rmap.unmap()` (serialized on the page table lock). So `rmap.unmap()` will unmap the page.
 2. Fail to get the page. Then, try acquiring the lock of the VMO pages. After acquiring the VMO pages, the new size will be seen. It will then fail if the index does not fit the new page size.

So I think the following patch is correct. We do not need to use an `RwMutex` for `Rmap`.
<details>
<summary> Patch </summary>

```patch
diff --git a/kernel/src/vm/page_cache/mod.rs b/kernel/src/vm/page_cache/mod.rs
index 4cad7f890..67c016465 100644
--- a/kernel/src/vm/page_cache/mod.rs
+++ b/kernel/src/vm/page_cache/mod.rs
@@ -69,8 +69,8 @@
 //! | resize         | inode           | inode           | inode         | inode           | inode       |             |             |              |               |
 //! | flush          | N/A             | page            | (flush)       | (flush + evict) | rmap + page | rmap + page |             |              |               |
 //! | evict          | N/A             | page            | (flush)       | (flush + evict) | rmap + page | rmap + page | rmap + page |              |               |
-//! | fault (load)   | N/A             | N/A             | N/A           | (flush + evict) | PT + rmap   | N/A         | PT + page   | N/A          |               |
-//! | fault (store)  | N/A             | N/A             | (flush)       | (flush + evict) | PT + rmap   | PT + page   | PT + page   | N/A          | N/A           |
+//! | fault (load)   | N/A             | N/A             | N/A           | (flush + evict) | PT + xarray | N/A         | PT + page   | N/A          |               |
+//! | fault (store)  | N/A             | N/A             | (flush)       | (flush + evict) | PT + xarray | PT + page   | PT + page   | N/A          | N/A           |
 //!
 //! The annotations in the table are explained below.
 //!
@@ -100,8 +100,8 @@
 //!    page is always in a consistent state, even when operations are called
 //!    concurrently. The corresponding entry is noted as `LockA + LockB`. We
 //!    use `rmap` to denote the reverse-mapping lock, `PT` to denote the page
-//!    table lock, and `page` to denote the page lock. For more details, see
-//!    the comments in the implementation.
+//!    table lock, `page` to denote the page lock, and `xarray` to denote the
+//!    lock of VMO pages. For details, see the comments in the implementation.
 
 use core::{
     ops::{Deref, Range},
diff --git a/kernel/src/vm/page_cache/vmo/mod.rs b/kernel/src/vm/page_cache/vmo/mod.rs
index 004d46b8b..3c2bda3a7 100644
--- a/kernel/src/vm/page_cache/vmo/mod.rs
+++ b/kernel/src/vm/page_cache/vmo/mod.rs
@@ -997,7 +997,7 @@ impl<'a> BackedVmo<'a> {
 
         // We remove pages before unmapping them in the page table. So a concurrent page fault will
         // synchronize via the page table lock. Either the mapped item will be unmapped below or the
-        // page fault handler will commit the page and be blocked on the reverse-mapping lock.
+        // page fault handler will see the new size after acquiring the lock in `self.pages`.
         let page_idx_range = get_page_idx_range(range);
         let removed_pages = self.remove_pages_in(locked_pages, &page_idx_range);
         if removed_pages.is_empty() {
@@ -1006,9 +1006,6 @@ impl<'a> BackedVmo<'a> {
 
         locked_rmap.unmap(page_idx_range.start * PAGE_SIZE..page_idx_range.end * PAGE_SIZE);
 
-        // We can allow new pages to be committed before waiting for the decommitted pages to be
-        // written back. There is no relationship because the contents of the old pages should be
-        // discarded. Therefore, we can release the reverse-mapping lock now.
         drop(locked_rmap);
 
         for (_, page) in removed_pages {
@@ -1021,9 +1018,6 @@ impl<'a> BackedVmo<'a> {
 
     /// Commits a page at the given index for a VMO with a backend.
     fn commit_on_internal(&self, page_idx: usize, commit_mode: CommitMode) -> Result<CachePage> {
-        // The reverse-mapping lock must be held when adding a new page.
-        let locked_rmap = self.rmap.lock();
-
         let mut locked_pages = self.pages.lock();
         if page_idx >= self.size().div_ceil(PAGE_SIZE) {
             return_errno_with_message!(Errno::EINVAL, "the offset is outside the VMO");
@@ -1034,7 +1028,6 @@ impl<'a> BackedVmo<'a> {
         if let Some(page) = cursor.load() {
             let page = page.clone();
             drop(locked_pages);
-            drop(locked_rmap);
 
             if !commit_mode.skips_backend_read() {
                 page.ensure_init(|locked_page| self.backend.read_page(page_idx, locked_page))?;
@@ -1048,7 +1041,6 @@ impl<'a> BackedVmo<'a> {
         let uninit_page = CachePage::alloc_uninit()?;
         cursor.store(uninit_page.clone());
         drop(locked_pages);
-        drop(locked_rmap);
 
         if commit_mode.skips_backend_read() {
             // The page will be completely overwritten, no need to read.
```

</details>

Am I right? Could @cchanging also help to check this? Thanks!

cchanging:
> Am I right? Could @cchanging also help to check this? Thanks!

I think you’re right. What we’ve implemented so far indeed doesn’t require holding the rmap lock when committing pages, just as you said.

We may only really need to hold the rmap lock during commit after we implement operations like `FallocMode::PunchHoleKeepSize`—for example, on ext4, where the data in the corresponding range is removed but the page cache size does not change. We can handle this later.

lrh2000:
> I think you’re right. What we’ve implemented so far indeed doesn’t require holding the rmap lock when committing pages, just as you said.
> 
> We may only really need to hold the rmap lock during commit after we implement operations like `FallocMode::PunchHoleKeepSize`—for example, on ext4, where the data in the corresponding range is removed but the page cache size does not change. We can handle this later.

Thanks for the confirmation!

kernel/src/vm/vmar/vmar_impls/mod.rs
```diff
@@ -167,6 +170,27 @@ impl Drop for RssDelta<'_> {
...
+    pub(self) fn new(vmo: Option<Arc<Vmo>>) -> Self {
+        Self { vmo }
+    }
+
+    pub(self) fn remove(&mut self, vmar: &Vmar, addr: Vaddr) -> Option<MutexGuard<'_, Rmap>> {
+        let mut rmap = self.vmo.as_ref()?.rmap().lock();
+        rmap.remove(vmar.weak_self.clone(), addr);
+        Some(rmap)
+    }
+}
```
cchanging, line 192: The other parts LGTM! But the VMAR-related changes in this PR still look a bit odd to me.

What bothers me is still this `RmapToRemove`. As it is used currently:

1. it is first constructed from a VMO,
2. then `remove` is called to remove the `VmMapping` associated with the abstract VMO and obtain the rmap guard,
3. and optionally, that rmap guard is then used to perform further `remove` or `insert` operations.

In actual usage, steps 2 and 3 sometimes feel more like peer operations, but the current API forces the caller to carry around two different handles (`RmapToRemove` and the guard returned by `remove`). Besides, the changes around `Vmar::remap` feel somewhat hard to read, mainly because they require variables like:

```rust
let mut _rmap_to_remove = super::RmapToRemove::new(None); // Suppress `unused_assignments`.
let mut _vmo_for_rmap = None; // Suppress `unused_assignments`.
```

I’m wondering whether we could use a different abstraction to encapsulate the rmap update semantics more cleanly like:

```rust
/// A scoped editor that keeps VM mappings and reverse mappings consistent.
struct MappingEdit<'a> {
    inner: &'a mut VmarInner,
    rmap: Option<&'a mut Rmap>,
    vmar: Weak<Vmar>,
}

impl MappingEdit<'_> {
    fn insert_try_merge(&mut self, vm_mapping: VmMapping);
    fn insert_without_try_merge(&mut self, vm_mapping: VmMapping);
    fn remove(&mut self, key: &Vaddr) -> Option<VmMapping>;
}

impl VmarInner {
    fn with_mapping_edit<R>(
        &mut self,
        vmar: &Vmar,
        rmap_vmo: Option<Arc<Vmo>>,
        op: impl FnOnce(&mut MappingEdit<'_>) -> R,
    ) -> R;
}
```

This way, we can also ensure something like the mapping removal and the rmap removal happen together when executing `MappingEdit::remove`. However, I’m not sure whether an approach like this might run into other issues.

cchanging (approved):

kernel/src/vm/vmar/vmar_impls/mod.rs
lrh2000: Sorry. I don't quite understand the proposed solution.

>         rmap_vmo: Option<Arc<Vmo>>,

How can the caller know the exact VMO before removing the VM mapping?

I don't think there is enough information to call `VmarInner::with_mapping_edit` (which happens before `MappingEdit::remove`).

> ```rust
> let mut _rmap_to_remove = super::RmapToRemove::new(None); // Suppress `unused_assignments`.
> let mut _vmo_for_rmap = None; // Suppress `unused_assignments`.
> ```

I agree that the code is ugly and could be hard to understand.

I have found another solution. How do you like [this update](https://github.com/asterinas/asterinas/compare/a6e289d54ac1d15e43e835974769e2d6425267ab..098a96a0be55e72c30f0da7531ce8ff9f5f68279)?

cchanging: > How can the caller know the exact VMO before removing the VM mapping?

The usage I had in mind for my proposal was to first find the `VmMapping` to operate on, get and lock its `rmap`, and then create a `MappingEdit` for the subsequent operations. But thinking about it now, that would introduce extra lookup overhead, so it probably isn’t a good idea.

> I have found another solution. How do you like [this update](https://github.com/asterinas/asterinas/compare/a6e289d54ac1d15e43e835974769e2d6425267ab..098a96a0be55e72c30f0da7531ce8ff9f5f68279)?

This version is indeed much clearer, and I think it looks good. Thanks for the revision!

StevenJiang1110 (approved):

cqs21 (approved):

PR #3423, opened 2026-06-22
Fix fallocate errno for special files

WaterWhisperer:
## Summary

Fix `fallocate()` on special `FileLike` objects to return `ENODEV` instead of `EOPNOTSUPP`.

## Changes

- Return `ENODEV` from the default `FileLike::fallocate()` implementation.
- Add regression for `fallocate()` on special `FileLike` objects.
- Unblock the fixed gVisor fallocate cases.

## Testing

- `make check`

  ```text
  All regression tests passed.
  ```

  ```text
  79 of 79 test cases passed.
  All conformance tests passed.
  ```

## Additional

Linux's `vfs_fallocate()` returns `ENODEV` for targets that are neither regular files nor block devices:
https://elixir.bootlin.com/linux/v7.1/source/fs/open.c#L324

### DIFF
--- a/kernel/src/fs/file/file_handle.rs
+++ b/kernel/src/fs/file/file_handle.rs
@@ -141,8 +141,11 @@ pub trait FileLike: Pollable + Send + Sync + Any {
         return_errno_with_message!(Errno::ESPIPE, "seek is not supported");
     }
 
-    fn fallocate(&self, mode: FallocMode, offset: usize, len: usize) -> Result<()> {
-        return_errno_with_message!(Errno::EOPNOTSUPP, "fallocate is not supported");
+    fn fallocate(&self, _mode: FallocMode, _offset: usize, _len: usize) -> Result<()> {
+        return_errno_with_message!(
+            Errno::ENODEV,
+            "fallocate is not supported for this file type"
+        );
     }
 
     fn as_socket(&self) -> Option<&dyn Socket> {
--- a/kernel/src/fs/vfs/notify/inotify.rs
+++ b/kernel/src/fs/vfs/notify/inotify.rs
@@ -15,7 +15,7 @@ use crate::{
         file::{AccessMode, CreationFlags, FileLike, StatusFlags, file_table::FdFlags},
         pseudofs::AnonInodeFs,
         vfs::{
-            inode::Inode,
+            inode::{FallocMode, Inode},
             inode_ext::InodeExt,
             notify::{FsEventSubscriber, FsEvents},
             path::Path,
@@ -359,6 +359,10 @@ impl FileLike for InotifyFile {
             Ordering::Relaxed,
         );
         Ok(())
+    }
+
+    fn fallocate(&self, _mode: FallocMode, _offset: usize, _len: usize) -> Result<()> {
+        return_errno_with_message!(Errno::EBADF, "inotify file is not opened writable");
     }
 
     fn access_mode(&self) -> AccessMode {
--- a/test/initramfs/src/conformance/gvisor/blocklists/fallocate_test
+++ b/test/initramfs/src/conformance/gvisor/blocklists/fallocate_test
@@ -1,4 +1,1 @@
-AllocateTest.FallocatePipe
-AllocateTest.FallocateChar
 AllocateTest.FallocateRlimit
-AllocateTest.FallocateOtherFDs
--- a/test/initramfs/src/regression/fs/pseudofs/fallocate.c
+++ b/test/initramfs/src/regression/fs/pseudofs/fallocate.c
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include "pseudo_file_create.h"
+
+FN_TEST(special_fds)
+{
+	int udp_socket = TEST_SUCC(socket(AF_INET, SOCK_DGRAM, 0));
+
+	struct fd_fallocate_case {
+		int fd;
+		int expected_errno;
+	};
+
+	struct fd_fallocate_case cases[] = {
+		{ pipe_1[0], EBADF },	{ pipe_1[1], ESPIPE },
+		{ pipe_2[0], EBADF },	{ pipe_2[1], ESPIPE },
+		{ sock[0], ENODEV },	{ sock[1], ENODEV },
+		{ epoll_fd, ENODEV },	{ event_fd, ENODEV },
+		{ timer_fd, ENODEV },	{ signal_fd, ENODEV },
+		{ inotify_fd, EBADF },	{ pid_fd, ENODEV },
+		{ mem_fd, 0 },		{ ns_uts_fd, EBADF },
+		{ udp_socket, ENODEV },
+	};
+
+	for (size_t i = 0; i < sizeof(cases) / sizeof(cases[0]); i++) {
+		if (cases[i].expected_errno == 0) {
+			TEST_SUCC(fallocate(cases[i].fd, 0, 0, 10));
+		} else {
+			TEST_ERRNO(fallocate(cases[i].fd, 0, 0, 10),
+				   cases[i].expected_errno);
+		}
+	}
+
+	TEST_SUCC(close(udp_socket));
+}
+END_TEST()
+
+#include "pseudo_file_cleanup.h"
--- a/test/initramfs/src/regression/fs/run_test.sh
+++ b/test/initramfs/src/regression/fs/run_test.sh
@@ -134,6 +134,7 @@ echo "All mount bind file test passed."
 ./procfs/proc_fd_open_fifo_after_setid
 ./procfs/tid
 
+./pseudofs/fallocate
 ./pseudofs/memfd_access_err
 ./pseudofs/memfd_create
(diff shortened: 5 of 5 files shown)


kernel/src/fs/file/file_handle.rs
```diff
@@ -142,7 +142,10 @@ pub trait FileLike: Pollable + Send + Sync + Any {
     }
 
     fn fallocate(&self, mode: FallocMode, offset: usize, len: usize) -> Result<()> {
-        return_errno_with_message!(Errno::EOPNOTSUPP, "fallocate is not supported");
+        return_errno_with_message!(
+            Errno::ENODEV,
+            "the file is not a regular file or a block device"
+        );
```
cchanging: I think a better approach would be to override `fallocate` in the `FileLike` implementations of those non-regular pseudo files and make it return `ENODEV`, rather than changing the default `fallocate` implementation.

As a default behavior, `EOPNOTSUPP` is more appropriate because it expresses that the operation is simply not supported by the file object unless a particular implementation chooses to provide different semantics. If we change the default to `ENODEV`, we will risk applying that behavior too broadly and making unrelated file types report a misleading error in the future.

cchanging (commented):
Thanks for fixing the returned errno!

kernel/src/fs/file/file_handle.rs
WaterWhisperer: Thanks! I didn't think through the boundary of the default impl carefully enough.

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include "pseudo_file_create.h"
+
+FN_TEST(special_fds)
+{
+	int inet_sock = TEST_SUCC(socket(AF_INET, SOCK_DGRAM, 0));
+	int fds[] = { timer_fd, signal_fd, event_fd, inet_sock, sock[0] };
```
StevenJiang1110: Why not test all pseudo file types defined in `pseudo_file_create.h`? Do all of them need to return `ENODEV`?

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include "pseudo_file_create.h"
+
+FN_TEST(special_fds)
+{
+	int inet_sock = TEST_SUCC(socket(AF_INET, SOCK_DGRAM, 0));
```
StevenJiang1110: Perhaps you can move the `inet_sock` definition also to `pseudo_file_create.h`? And`udp_socket` is the more accurate name than `inet_sock`.

kernel/src/time/timerfd.rs
```diff
@@ -250,6 +250,13 @@ impl FileLike for TimerfdFile {
         Ok(())
     }
 
+    fn fallocate(&self, _mode: FallocMode, _offset: usize, _len: usize) -> Result<()> {
+        return_errno_with_message!(
+            Errno::ENODEV,
+            "the file is not a regular file or a block device"
```
StevenJiang1110:
```suggestion
            "timerfd does not support fallocate"
```

Similar for other error messages.

StevenJiang1110 (approved):
LGTM! Only a few minor issues. Let @cchanging perform the final review and merge.

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
WaterWhisperer: No, not all of them should return `ENODEV`. For example, pipes should still return `ESPIPE`.

WaterWhisperer: Oh, did you mean that this regression should cover all pseudo file types from `pseudo_file_create.h` with their expected `fallocate()` results, rather than only the `ENODEV` cases?

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
WaterWhisperer: Thanks, I'll rename it. But I think keeping it here is better since this socket is only used by this test?

kernel/src/time/timerfd.rs
WaterWhisperer: Thanks! I think changing `the file` to the specific fd type should be enough here. Using `does not support fallocate` sounds closer to `EOPNOTSUPP`, while this path returns `ENODEV`.

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
StevenJiang1110: > Oh, did you mean that this regression should cover all pseudo file types from pseudo_file_create.h with their expected fallocate() results, rather than only the ENODEV cases?

Yes, I think all file types should be included. For example, `inotify_fd`, `ns_uts_fd`, and `pid_fd` are missing from the current test — don't they also return `ENODEV`?

```C
int epoll_fd, event_fd, timer_fd, signal_fd, inotify_fd, pid_fd, mem_fd;
int ns_uts_fd;
```

Some may return a different errno, but it would be best to cover all of them in one go since this PR is titled "Fix fallocate errno for special files".

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
StevenJiang1110: OK, it is acceptable to keep it here.

The reason I raised this is that in your `fds` array, `inet_sock` is placed before `sock[0]`, which is quite confusing. If it is defined in this file, `inet_sock` should clearly come last, and the file descriptors sourced from `pseudo_file_create.h` should appear in the array in their original order.

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
WaterWhisperer: Okay

test/initramfs/src/regression/fs/pseudofs/fallocate_err.c
WaterWhisperer: Got it. I originally followed the gVisor test order. I'll update it.

kernel/src/time/timerfd.rs
StevenJiang1110: The message "the file is not a regular file or a block device" is misleading. It implies that fallocate is unsupported for everything other than regular files and block devices, which is clearly not the case — for instance, memfd supports fallocate despite being neither a regular file nor a block device.

WaterWhisperer: Understood, the memfd example makes it clear

tatetian:
> Fix `fallocate()` on special `FileLike` objects to return `ENODEV` instead of `EOPNOTSUPP`.

The default implementation for `FileLike::fallocate` still returns `EOPNOTSUPP`:

https://github.com/asterinas/asterinas/blob/83ace475af27d36168b3b9e0372bd14730cb0477/kernel/src/fs/file/file_handle.rs#L144-L146

Should we fix the default one so that all current and future special file types do not override the default implementation to return a different errno?

WaterWhisperer:
> The default implementation for `FileLike::fallocate` still returns `EOPNOTSUPP`:

@tatetian Sorry for not updating the PR body yet. I originally changed the default impl, but changed it back after this review thread: https://github.com/asterinas/asterinas/pull/3423#discussion_r3465030861

> Should we fix the default one so that all current and future special file types do not override the default implementation to return a different errno?

Initially I changed the default one for the same reason, since Linux also handles this in the common `vfs_fallocate()` path. But @cchanging pointed out that `FileLike::fallocate()` is a generic trait default, so `ENODEV` there may be too broad. That made sense to me, so I switched to explicit overrides.

WaterWhisperer:
Hmm, I’m still a bit unsure about this tradeoff: the default impl is simpler and avoids repeated overrides, while explicit overrides keep the generic `FileLike` default more conservative. 🤔

cchanging:
> Initially I changed the default one for the same reason, since Linux also handles this in the common vfs_fallocate() path. But @cchanging pointed out that FileLike::fallocate() is a generic trait default, so ENODEV there may be too broad. That made sense to me, so I switched to explicit overrides.

I checked the [manual](https://www.man7.org/linux/man-pages/man2/fallocate.2.html):
>EOPNOTSUPP
              The filesystem containing the file referred to by fd does
              not support this operation; or the mode is not supported by
              the filesystem containing the file referred to by fd.

So this errno should mainly result from lack of support at the filesystem level, and no single file type would return this errno by default. It makes sense to have it as the default implementation of `Inode::fallocate`.

For `FileLike::fallocate`, most `FileLike` types added later in addition to `InodeHandle` can likely be considered non-regular files, and it does seem that in most cases they would just return `ENODEV`.

So now I think we can indeed directly modify the default implementation of `FileLike`. Sorry for the noise.

tatetian (approved):

PR #3425, opened 2026-06-23
Support `root` boot parameter

taosue:
This PR adds Linux-compatible `root` parameter support for selecting the `initramfs` entry point and mounting a root filesystem.

Main changes:
  - Correct the previous `init=` usage to `rdinit=` for selecting the initramfs entry point, matching Linux semantics.
  - Resolve Linux-style `/dev/...` root device specs through block-device lookup before `/dev` is mounted, aligning with Linux where `root=/dev/...` is parsed before `devtmpfs` is mounted.
  - Document the supported `root`-related kernel parameters.
  - Clean up the NixOS init handoff command line by relying on the default `/init` entry point in `initramfs` instead of spelling it out explicitly.

### DIFF
--- a/book/src/kernel/linux-compatibility/kernel-parameters.md
+++ b/book/src/kernel/linux-compatibility/kernel-parameters.md
@@ -4,19 +4,18 @@ This section documents kernel command-line parameters supported by Asterinas.
 
 ## Inherited from Linux
 
-### `init`
+### `rdinit`
 
-Run the specified binary as `init`.
+Run the specified initramfs binary as the first userspace process.
 
 Example:
 ```text
-init=/bin/busybox
+rdinit=/bin/busybox
 ```
 
 Notes:
-- The value is the path to the executable.
-- If omitted, Asterinas will try to execute from the following paths in order:
-  `/sbin/init`, `/etc/init`, `/bin/init`, `/bin/sh`.
+- The value is the path to the executable in the initramfs root.
+- If omitted, Asterinas will try to execute `/init` from the initramfs root.
 
 ### `console`
 
--- a/book/src/kernel/linux-compatibility/kernel-parameters.md
+++ b/book/src/kernel/linux-compatibility/kernel-parameters.md
@@ -16,6 +16,81 @@ rdinit=/bin/busybox
 Notes:
 - The value is the path to the executable in the initramfs root.
 - If omitted, Asterinas will try to execute `/init` from the initramfs root.
+- If the specified path cannot be accessed,
+  Asterinas ignores it and proceeds with root filesystem initialization.
+  It does not fall back to `/init` in this case.
+
+### `root`
+
+Mount the specified block device as the root filesystem.
+
+Example:
+```text
+root=/dev/vda2
+```
+
+Notes:
+- The value currently must name a registered block device under `/dev`,
+  such as `/dev/vda2`.
+- Asterinas first determines the boot source. If the path specified by
+  `rdinit` is accessible, or if `rdinit` is absent and `/init` exists, it
+  boots from the initramfs and ignores `root`. Otherwise, it boots from the
+  rootfs, for which `root` is required: Asterinas mounts the specified device
+  as the root filesystem and panics if the parameter is absent.
+
+### `rootfstype`
+
+Select the filesystem type used for the root filesystem.
+
+Example:
+```text
+root=/dev/vda2 rootfstype=ext2
+```
+
+Valid values:
+- `ext2`
+
+### `ro`
+
+Mount the root filesystem read-only.
+This is the default when Asterinas mounts the root filesystem via `root`.
+
+Example:
+```text
+root=/dev/vda2 ro
+```
+
+Notes:
+- This parameter has no effect when Asterinas boots from the initramfs.
+- If both `ro` and `rw` are specified, the last one takes precedence.
+
+### `rw`
+
+Mount the root filesystem read-write.
+
+Example:
+```text
+root=/dev/vda2 rw
+```
+
+Notes:
+- This parameter has no effect when Asterinas boots from the initramfs.
+- `rw` overrides the default read-only root mount.
+- If both `ro` and `rw` are specified, the last one takes precedence.
+
+### `init`
+
+Run the specified executable as the first userspace process from the root filesystem.
+
+Example:
+```text
+root=/dev/vda2 rootfstype=ext2 init=/nix/var/nix/profiles/system/init
+```
+
+Notes:
+- `init` is used only after Asterinas mounts the root filesystem via `root=`.
+- If omitted,
+  Asterinas tries `/sbin/init`, `/etc/init`, `/bin/init`, and `/bin/sh`, in that order.
 
 ### `console`
 
--- a/distro/etc_nixos/modules/core.nix
+++ b/distro/etc_nixos/modules/core.nix
@@ -79,17 +79,14 @@ in {
   # TODO: Fix errors and warnings from systemd and remove this setting.
   environment.sessionVariables = { SYSTEMD_LOG_LEVEL = "crit"; };
   system.systemBuilderCommands = ''
-    echo "PATH=/bin:/nix/var/nix/profiles/system/sw/bin earlycon loglevel=${config.aster_nixos.log-level} console=${config.aster_nixos.console} -- sh /init root=/dev/vda2 init=/nix/var/nix/profiles/system/stage-2-init rd.break=${
+    echo "PATH=/bin:/nix/var/nix/profiles/system/sw/bin earlycon loglevel=${config.aster_nixos.log-level} console=${config.aster_nixos.console} -- root=/dev/vda2 init=/nix/var/nix/profiles/system/init rd.break=${
       if config.aster_nixos.break-into-stage-1-shell then "1" else "0"
     }"  > $out/kernel-params
-    mv $out/init $out/stage-2-init
-    sed -i 's_^\([[:space:]]*\)\(exec > >(tee -i /run/log/stage-2-init.log) 2>&1\)$_\1# \2_' $out/stage-2-init
+    sed -i 's_^\([[:space:]]*\)\(exec > >(tee -i /run/log/stage-2-init.log) 2>&1\)$_\1# \2_' $out/init
     if [ "${config.aster_nixos.disable-systemd}" = "true" ]; then
-      sed -i 's/^[[:space:]]*echo "starting systemd..."$/# &/' $out/stage-2-init
-      sed -i 's/^[[:space:]]*exec \/run\/current-system\/systemd\/lib\/systemd\/systemd "$@"$/# &/' $out/stage-2-init
+      sed -i 's/^[[:space:]]*echo "starting systemd..."$/# &/' $out/init
+      sed -i 's/^[[:space:]]*exec \/run\/current-system\/systemd\/lib\/systemd\/systemd "$@"$/# &/' $out/init
     fi
-    rm -rf $out/init
-    ln -s /bin/busybox $out/init
     ln -s ${kernel} $out/kernel
     ln -s ${initramfs}/initrd $out/initrd
   '';
--- a/kernel/core/src/fs/rootfs.rs
+++ b/kernel/core/src/fs/rootfs.rs
@@ -1,119 +1,203 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use alloc::borrow::Cow;
+//! Root filesystem mounting during boot.
+//!
+//! Asterinas first tries to boot from the initramfs. It checks the path specified by `rdinit`, or
+//! `/init` when `rdinit` is not provided. If the selected path is accessible, Asterinas runs it as
+//! the first userspace process. Otherwise, Asterinas boots from the root filesystem.
+//!
+//! For root filesystem boot, this module mounts a supported filesystem from the block device
+//! specified by `root`. The mounted filesystem replaces the bootstrap root in the initial mount
+//! namespace and becomes the root and working directory of the first userspace process. The
+//! bootstrap root is then detached. The `root` parameter is required when this boot path is
+//! selected, and [`SUPPORTED_ROOTFS_TYPES`] lists the supported filesystem types.
+//!
+//! The following kernel parameters configure root filesystem boot:
+//!
+//! - `root` specifies the block device to mount.
+//! - `rootfstype` specifies a comma-separated list of filesystem type candidates. All supported
+//!   types are tried when it is not provided.
+//! - `ro` and `rw` select read-only or read-write mounting. The root filesystem is read-only by
+//!   default. They use last-wins semantics: the last `ro` or `rw` parameter determines the mount
+//!   mode.
+//! - `init` specifies the init executable on the mounted filesystem. If it is not provided,
+//!   Asterinas tries `/sbin/init`, `/etc/init`, `/bin/init`, and `/bin/sh` in order.
 
-use cpio_decoder::{CpioDecoder, CpioEntry, FileMetadata, FileType};
-use device_id::{DeviceId, MajorId, MinorId};
-use lending_iterator::LendingIterator;
-use no_std_io2::io::{Cursor, Read};
-use ostd::boot::boot_info;
-use zune_inflate::DeflateDecoder;
+// Set this module's log prefix for `ostd::log`.
+macro_rules! __log_prefix {
+    () => {
+        "rootfs: "
+    };
+}
+
+use core::{
+    str::FromStr,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+use aster_cmdline::parse::ParamStorage;
+use spin::once::Once;
 
 use super::{
-    file::{InodeMode, InodeType},
-    vfs::path::{FsPath, PathResolver, is_dot},
+    ext2,
+    vfs::{
+        file_system::FsFlags,
+        path::{FsPath, Mount, MountNamespace, Path, PathResolver, PerMountFlags},
+        registry::{DynFsType, FsAndRoot, FsCreationCtx},
+    },
 };
-use crate::{fs::vfs::inode::MknodType, prelude::*};
+use crate::prelude::*;
 
-/// Unpack and prepare the rootfs from the initramfs CPIO buffer.
-pub fn init_in_first_kthread(path_resolver: &PathResolver) -> Result<()> {
-    let initramfs_buf = boot_info()
-        .initramfs
-        .ok_or_else(|| Error::with_message(Errno::EINVAL, "no initramfs found"))?;
+/// Filesystem types supported for the root filesystem.
+pub static SUPPORTED_ROOTFS_TYPES: &[&dyn DynFsType] = &[&ext2::EXT2_TYPE];
 
-    let (reader, suffix) = match &initramfs_buf[..4] {
-        // Gzip magic number: 0x1F 0x8B
-        &[0x1F, 0x8B, _, _] => {
-            let decompressed = DeflateDecoder::new(initramfs_buf)
-                .decode_gzip()
-                .map_err(|_| Error::with_message(Errno::EINVAL, "gzip decompression failed"))?;
-            (Cow::Owned(decompressed), ".gz")
-        }
-        _ => (Cow::Borrowed(initramfs_buf), ""),
-    };
-
-    println!("[kernel] unpacking initramfs.cpio{} to rootfs ...", suffix);
-
-    let mut decoder = CpioDecoder::new(Cursor::new(reader));
-
-    while let Some(entry_result) = decoder.next() {
-        let mut entry = entry_result?;
-        if let Err(e) = try_append_entry_to_rootfs(&mut entry, path_resolver) {
-            warn!("failed to add entry {} to rootfs: {:?}", entry.name(), e);
-        }
-    }
-
-    println!("[kernel] rootfs is ready");
+/// Mounts and switches to the root filesystem configured by the kernel command line.
+///
+/// The filesystem replaces the bootstrap root mount and becomes the resolver's root and working
+/// directory.
+///
+/// Returns an error if `root` is not provided or the root filesystem cannot be opened or mounted.
+pub fn switch_to_rootfs(path_resolver: &mut PathResolver) -> Result<()> {
+    let root = ROOT_PATH.get().ok_or_else(|| {
+        Error::with_message(Errno::EINVAL, "the `root` parameter was not provided")
+    })?;
+    let (fs_flags, mount_flags) = rootfs_flags();
+    let fs_and_root = open_rootfs(root, fs_flags)?;
+    let rootfs_mount = Mount::new_detached(
+        fs_and_root,
+        mount_flags,
+        Arc::downgrade(MountNamespace::get_init_singleton()),
+        Some(root.to_string()),
+    )?;
+    path_resolver.switch_root_for_boot(rootfs_mount);
+    println!("[kernel] rootfs is ready (mounted from {})", root);
     Ok(())
 }
 
-fn try_append_entry_to_rootfs<R: Read>(
-    entry: &mut CpioEntry<R>,
-    path_resolver: &PathResolver,
-) -> Result<()> {
-    // Make sure the name is a relative path, and is not end with "/".
-    let entry_name = entry.name().trim_start_matches('/').trim_end_matches('/');
-    if entry_name.is_empty() {
-        return_errno_with_message!(Errno::EINVAL, "invalid entry name");
-    }
-    if is_dot(entry_name) {
-        return Ok(());
-    }
-
-    // Here we assume that the directory referred by "prefix" must has been created.
-    // The basis of this assumption is：
-    // The mkinitramfs script uses `find` command to ensure that the entries are
-    // sorted that a directory always appears before its child directories and files.
-    let (parent, name) = if let Some((prefix, last)) = entry_name.rsplit_once('/') {
-        (path_resolver.lookup(&FsPath::try_from(prefix)?)?, last)
-    } else {
-        (path_resolver.root().clone(), entry_name)
+/// Finds the init program specified for booting from the root filesystem.
+///
+/// Resolves the path specified by `init` and returns the resolved path together with the original
+/// pathname. Returns `Ok(None)` when `init` is not provided, and an error if the configured
+/// pathname is invalid or cannot be resolved.
+pub fn find_init(path_resolver: &PathResolver) -> Result<Option<(Path, &'static str)>> {
+    let Some(init_path) = INIT_PATH.get().map(String::as_str) else {
+        return Ok(None);
     };
 
-    let metadata = entry.metadata();
-    let mode = InodeMode::from_bits_truncate(metadata.permission_mode());
-    match metadata.file_type() {
-        FileType::File => {
-            let path = parent.new_fs_child(name, InodeType::File, mode)?;
-            entry.read_all(path.inode().writer(0))?;
-        }
-        FileType::Dir => {
-            let _ = parent.new_fs_child(name, InodeType::Dir, mode)?;
-        }
-        FileType::Link => {
-            let path = parent.new_fs_child(name, InodeType::SymLink, mode)?;
-            let link_content = {
-                let mut link_data: Vec<u8> = Vec::new();
-                entry.read_all(&mut link_data)?;
-                core::str::from_utf8(&link_data)?.to_string()
-            };
-            path.inode().write_link(&link_content)?;
-        }
-        FileType::Char => {
-            let device_id = try_device_id_from_metadata(metadata)?;
-            parent.mknod(name, mode, MknodType::CharDevice(device_id))?;
-        }
-        FileType::Block => {
-            let device_id = try_device_id_from_metadata(metadata)?;
-            parent.mknod(name, mode, MknodType::BlockDevice(device_id))?;
-        }
-        FileType::FiFo => {
-            parent.mknod(name, mode, MknodType::NamedPipe)?;
-        }
-        FileType::Socket => {
-            return_errno_with_message!(Errno::EINVAL, "socket files are not supported in initramfs")
+    let path = path_resolver.lookup(&FsPath::try_from(init_path)?)?;
+    Ok(Some((path, init_path)))
+}
+
+fn open_rootfs(root: &str, fs_flags: FsFlags) -> Result<FsAndRoot> {
+    // Treat a `/dev/...` value of `root` as a Linux-compatible root device spec, not as a VFS path
+    // lookup.
+    // Reference: <https://elixir.bootlin.com/linux/v6.18/source/drivers/base/devtmpfs.c#L358-L359>.
+    let device_name = root
+        .strip_prefix("/dev/")
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "root must name a /dev block device"))?;
+    let device = aster_block::lookup_by_name(device_name)
+        .ok_or_else(|| Error::with_message(Errno::ENODEV, "root block device not found"))?;
+    open_rootfs_from_candidates(device, rootfs_types(), fs_flags)
+}
+
+fn open_rootfs_from_candidates(
+    device: Arc<dyn aster_block::BlockDevice>,
+    rootfs_types: &[&dyn DynFsType],
+    fs_flags: FsFlags,
+) -> Result<FsAndRoot> {
+    let mut fs_creation_ctx = FsCreationCtx::from_block_device(device, fs_flags, None);
+    for rootfs_type in rootfs_types {
+        match rootfs_type.get_or_create(&mut fs_creation_ctx) {
+            Ok(fs_and_root) => return Ok(fs_and_root),
+            Err(err) if err.error() == Errno::EINVAL => continue,
+            Err(err) => return Err(err),
         }
     }
 
-    Ok(())
+    return_errno_with_message!(Errno::ENODEV, "no root filesystem type could mount root")
 }
 
-fn try_device_id_from_metadata(metadata: &FileMetadata) -> Result<u64> {
-    let major = {
-        let dev_maj = u16::try_from(metadata.rdev_maj())?;
-        MajorId::try_from(dev_maj).map_err(|msg| Error::with_message(Errno::EINVAL, msg))?
-    };
-    let minor = MinorId::try_from(metadata.rdev_min())
-        .map_err(|msg| Error::with_message(Errno::EINVAL, msg))?;
-    Ok(DeviceId::new(major, minor).as_encoded_u64())
+fn rootfs_flags() -> (FsFlags, PerMountFlags) {
+    let mut fs_flags = FsFlags::empty();
+    let mut mount_flags = PerMountFlags::default();
+    if ROOT_MOUNT_READ_ONLY.load(Ordering::Relaxed) {
+        fs_flags.insert(FsFlags::RDONLY);
+        mount_flags.insert(PerMountFlags::RDONLY);
+    }
+    (fs_flags, mount_flags)
 }
+
+fn rootfs_types() -> &'static [&'static dyn DynFsType] {
+    ROOTFS_TYPES
+        .get()
+        .map(RootFsTypes::as_slice)
+        .unwrap_or(SUPPORTED_ROOTFS_TYPES)
+}
+
+static ROOT_PATH: Once<String> = Once::new();
+aster_cmdline::define_kv_param!("root", ROOT_PATH);
+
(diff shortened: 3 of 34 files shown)


kernel/src/init.rs
```diff
@@ -135,13 +143,66 @@ fn first_kthread() {
...
     INIT_PROCESS.call_once(|| {
         let karg = INIT_PROC_ARGS.get().unwrap();
-        let init_path = INIT_PATH.get().map(|s| s.as_str());
-        spawn_init_process(init_path, karg.argv().to_vec(), karg.envp().to_vec())
-            .expect("Failed to run the init process")
+        let argv = karg.argv().to_vec();
+        let envp = karg.envp().to_vec();
+        match boot_init {
+            BootInit::Initramfs(init_path) => {
+                println!("[kernel] running {} as the initramfs init", init_path);
```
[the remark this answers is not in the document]
taosue: Keeping this as `println!` intentionally. These boot status messages are printed unconditionally, consistent with the nearby boot banner/rootfs messages, and would be hidden by the default `LOG_LEVEL=error` if converted to `info!`.

kernel/src/fs/rootfs.rs
```diff
@@ -27,11 +33,40 @@ impl Read for BoxedReader<'_> {
...
+        impl FromStr for RootFsType {
+            type Err = core::convert::Infallible;
+
+            fn from_str(type_name: &str) -> Result<Self, Self::Err> {
+                match type_name {
+                    $($name => Ok(Self::$variant),)+
+                    _ => panic!("unsupported root filesystem type '{}'", type_name),
+                }
+            }
+        }
```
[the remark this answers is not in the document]
taosue: Linux does not validate `rootfstype=` during command-line parsing. It keeps the string as mount candidates, then tries them during root mount. If none can mount the root filesystem, [the kernel panics](https://github.com/torvalds/linux/blob/ef0c9f75a19532d7675384708fc8621e10850104/init/do_mounts.c#L242). So panicking eventually is Linux-compatible, but panicking inside `FromStr` is earlier than Linux and inconsistent with the parser contract.

hburaylee:
I'm very glad to see this PR. I think this PR can be part of the work for issue #3429: "Support Asterinas as guest kernel in Firecracker microVMs". Looking forward to seeing this PR merged soon.

kernel/core/src/process/process/init_proc.rs
```diff
@@ -53,28 +50,6 @@ pub fn spawn_init_process(
     Ok(process)
 }
 
-fn create_default_init_process(argv: Vec<CString>, envp: Vec<CString>) -> Result<Arc<Process>> {
-    // Linux probes the fallback init executables in this order:
-    // <https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1634>.
-    const DEFAULT_INIT_EXEC_PATHS: &[&str] = &["/sbin/init", "/etc/init", "/bin/init", "/bin/sh"];
```
cchanging, line 59: I think the original reference makes sense?

> Clean up the NixOS init handoff command line by relying on the default /init entry point in initramfs instead of spelling it out explicitly.

So why was the previous approach wrong? If it’s meant as a workaround, for compatibility reasons maybe we could add `"/init"` to `DEFAULT_INIT_EXEC_PATHS` and include some extra comments explaining it.

kernel/src/fs/rootfs.rs
```diff
@@ -10,16 +23,70 @@ use ostd::boot::boot_info;
...
+
+macro_rules! define_rootfs_types {
+    (
+        $(
+            $variant:ident => {
+                name: $name:literal,
+                open: $open:expr,
+            }
+        ),+ $(,)?
+    ) => {
```
cchanging: I think we could potentially store only the corresponding `FsType` here, but we might need to wait until #3328 is merged.

kernel/src/fs/rootfs.rs
```diff
@@ -47,6 +114,46 @@ pub fn init_in_first_kthread(path_resolver: &PathResolver) -> Result<()> {
     Ok(())
 }
 
+/// Mounts the specified block device as a root filesystem in a new mount namespace.
+pub fn mount(
```
cchanging: Some of the APIs involved here look quite strange. For example, a method called `mount` (including `mount_rootfs`) returns a `MountNamespace`, which feels very counterintuitive.

cchanging (commented):
Thanks for the implementation! I haven’t finished reviewing the entire PR yet, I’ll leave more detailed comments later.

kernel/core/src/process/process/init_proc.rs
taosue: The original Linux fallback list is correct. The problem was not the list itself, but the workaround in `core.nix`, which mixed the initramfs entry point with the real-root `init`.

`$out` is the NixOS system closure that is eventually linked from `/nix/var/nix/profiles/system`, and its original `$out/init` is the stage-2 `init` on the real root filesystem. The previous code renamed it to `$out/stage-2-init` only to free `$out/init` and replace it with a symlink to `/bin/busybox`. This effectively made the generated init entry resolve to `/bin/busybox` (i.e. `init=/bin/busybox` in kernel parameter), even though that path was consumed as the first process while the kernel was still using the initramfs. In other words, we were modifying the **real-root** system closure solely to select an executable from the **initramfs**.

This PR removes that workaround. The initramfs already installs its stage-1 script at `/init`, so initramfs boot should select either the path specified by `rdinit=` or the default `/init`. We do not need to specify `rdinit=/init`, and we do not need to pass `sh /init` explicitly because the script already has a `#!/bin/sh` shebang. This also allows `$out/init` to remain the normal NixOS stage-2 `init`.

Adding `/init` to `DEFAULT_INIT_EXEC_PATHS` would mix the two stages again. That list is used only after booting from the real root filesystem, where Linux probes `/sbin/init`, `/etc/init`, `/bin/init`, and `/bin/sh`. `/init` is the initramfs default and should remain handled separately.

kernel/src/fs/rootfs.rs
taosue: The function does not mount the filesystem into the caller’s current mount namespace. It creates a new mount namespace whose root mount is backed by the selected root filesystem, and that namespace is later assigned to the init process. Returning a `MountNamespace` is therefore intentional, but I agree that `mount`/`mount_rootfs` hides this important behavior.

Would renaming the API to something like `create_mount_namespace` make the construction and isolation semantics sufficiently explicit?

kernel/src/fs/rootfs.rs
taosue: I agree that storing only the corresponding `FsType` here would be a better approach after #3328 is merged.

One remaining issue is that `FsType::get_or_create()` requires an `FsCreationCtx`. Currently, `FsCreationCtx` accepts a source path and resolves it through a task’s VFS context. However, the rootfs boot path already has an `Arc<dyn BlockDevice>` and runs before a userspace task context and the `/dev` filesystem are available.

Could `FsCreationCtx` support being constructed from an already resolved `Arc<dyn BlockDevice>`? E.g., introduce `FsCreationCtx::from_block_device()`. This would allow the rootfs boot path to call `FsType::get_or_create()` and reuse the filesystem instance cache, instead of bypassing `FsType` and calling `Ext2::open()` directly.

kernel/core/src/process/process/init_proc.rs
cchanging: Should the fallback logic here be moved to the rootfs boot path instead? It looks like this PR removes it outright.

taosue: Yes, the fallback logic has been moved to `kernel/src/init.rs::spawn_default_rootfs_init()` rather than removed.

kernel/src/fs/rootfs.rs
cchanging: > introduce FsCreationCtx::from_block_device()

Yes I think this method makes sense.

.github/workflows/test_riscv.yml
```diff
@@ -53,9 +53,17 @@ jobs:
...
+          - test_id: 'boot-debug-smp4-rootfs'
+            smp: 4
+            rootfs_boot: true
+
           - test_id: 'boot-release'
             release: true
 
+          - test_id: 'boot-release-rootfs'
+            release: true
+            rootfs_boot: true
```
StevenJiang1110: You don't need to test rootfs booting under every combination of boot parameters. Since there are many possible combinations of startup parameters, our current principle is to ensure that each individual parameter value is covered — not that every possible combination is covered.

For example, we have two build modes (debug and release) and two SMP configurations (smp1 and smp4), giving 4 possible combinations in total. However, we only require testing `debug + smp4` and `release + smp1`. This way, debug, release, smp1, and smp4 are all covered, without testing every combination. So, similarly, I think adding just one test for rootfs should be sufficient.

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -4,19 +4,93 @@ This section documents kernel command-line parameters supported by Asterinas.
...
+Example:
+```text
+root=/dev/vda2
+```
+
+Notes:
+- The value currently must name a registered block device under `/dev`,
+  such as `/dev/vda2`.
+- By default, the initramfs init is `/init`;
+  `rdinit` overrides it with another path.
```
StevenJiang1110: This should be put under the `rdinit` section?

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -4,19 +4,93 @@ This section documents kernel command-line parameters supported by Asterinas.
...
+- If the selected initramfs init is present,
+  that program is responsible for switching to the root filesystem.
```
StevenJiang1110: I cannot understand line 38-39. It seems to be the userspace behavior? Should we include it here?

Or should we reword it as "If rdinit is provided, root= is ignored...."

kernel/core/src/process/process/init_proc.rs
cchanging: Sorry I missed that, since I search by the original reference link. However I think the original link is right? The modified link "https://elixir.bootlin.com/linux/v6.19/source/init/main.c#L1604-L1607" refers to a weird place.

book/src/kernel/linux-compatibility/kernel-parameters.md
```diff
@@ -4,19 +4,93 @@ This section documents kernel command-line parameters supported by Asterinas.
...
+Mount the root filesystem read-only.
+This is the default when Asterinas mounts the root filesystem via `root=`.
+
+Example:
+```text
+root=/dev/vda2 ro
+```
+
+Notes:
+- This parameter is ignored when an initramfs init is selected to run.
```
StevenJiang1110:
```suggestion
- This parameter is ignored when `rdinit=` is provided, or when `root=` is not provided.
```

Would this be more accurate? By the way, do parameters like `ro`/`rw` need to immediately follow `root=`, or can other parameters appear in between?

kernel/core/src/process/process/init_proc.rs
taosue: I pasted the modified line reference incorrectly; the original link is the correct one... I’ll restore the reference to #L1634. Thanks for catching this issue!

distro/etc_nixos/modules/core.nix
```diff
@@ -78,17 +78,14 @@ in {
   # TODO: Fix errors and warnings from systemd and remove this setting.
   environment.sessionVariables = { SYSTEMD_LOG_LEVEL = "crit"; };
   system.systemBuilderCommands = ''
-    echo "PATH=/bin:/nix/var/nix/profiles/system/sw/bin earlycon loglevel=${config.aster_nixos.log-level} console=${config.aster_nixos.console} -- sh /init root=/dev/vda2 init=/nix/var/nix/profiles/system/stage-2-init rd.break=${
+    echo "PATH=/bin:/nix/var/nix/profiles/system/sw/bin earlycon loglevel=${config.aster_nixos.log-level} console=${config.aster_nixos.console} -- root=/dev/vda2 init=/nix/var/nix/profiles/system/init rd.break=${
```
StevenJiang1110, line 82: It looks like you've changed NixOS's boot method from initramfs to rootfs. However, NixOS's default boot process is inherently a two-stage process, consisting of initramfs stage 1 and rootfs stage 2. Could we keep the original boot method instead? For example:

```suggestion
    echo "PATH=/bin:/nix/var/nix/profiles/system/sw/bin earlycon loglevel=${config.aster_nixos.log-level} console=${config.aster_nixos.console} -- rdinit=/init root=/dev/vda2 init=/nix/var/nix/profiles/system/stage-2-init rd.break=${
```

taosue: > It looks like you've changed NixOS's boot method from initramfs to rootfs.

The boot process is still two-stage. We omit `rdinit=/init` because `/init` is already the default initramfs entry point when `rdinit` is not specified. Also, placing `rdinit=/init` after `--`, as in the suggested change, would only pass it to the stage-1 script rather than make it a kernel parameter.

`stage-2-init` was not the original NixOS path. `$out/init` is the NixOS stage-2 init script. The previous code renamed it to `$out/stage-2-init` only so that `$out/init` could be replaced with a symlink to `/bin/busybox`, which caused the old boot path to select `init=/bin/busybox` and then invoke `sh /init` in the initramfs. Now that the kernel can execute the initramfs `/init` directly through its shebang, that workaround is no longer needed, and `$out/init` can retain its original role as the stage-2 init.

StevenJiang1110: Because the documentation currently reads:

```md
### root

Mount the specified block device as the root filesystem
when no initramfs init is selected to run.
```

My understanding is that this means: if `rdinit` is not specified, the kernel itself is responsible for mounting the root filesystem, so the system boots directly into the root filesystem — is that not the correct interpretation?

taosue: > Mount the specified block device as the root filesystem
> when no initramfs init is selected to run.

The current wording is indeed misleading. Sorry for the confusion.

Omitting `rdinit` does not immediately cause the kernel to mount the root filesystem. The kernel first checks for the default `/init` in the initramfs and runs it if present. It mounts the root filesystem only when no initramfs init is selected—normally when `rdinit` is not specified and `/init` is absent.

This selection order is described in more detail in the module documentation for `rootfs.rs`. I will update the Book to make the same order explicit here as well.

book/src/kernel/linux-compatibility/kernel-parameters.md
taosue: Sorry for the confusion. By “an initramfs init is selected,” I mean that the kernel chooses the initramfs path rather than the rootfs path. In this case, `ro` and `rw` have no effect.  I will expand the note to make “selected” explicit.

“When root= is not provided” is not an ignored case if no initramfs init was selected—the kernel will panic because it has neither an initramfs init nor a root filesystem to boot from.

`ro` and `rw` are parsed independently and do not need to immediately follow `root=`. Other parameters may appear between them, and if both are specified, the last one takes precedence.

kernel/src/fs/vfs/fs_apis/registry.rs
```diff
@@ -103,11 +103,18 @@ impl<T: FsType> DynFsType for T {
...
+}
+
+enum FsCreationSource<'a> {
+    Vfs {
+        source: Option<&'a str>,
+        task_ctx: &'a Context<'a>,
+        resolved_block_device: Option<Arc<dyn BlockDevice>>,
+    },
+    BlockDevice(Arc<dyn BlockDevice>),
 }
```
cchanging, line 115: Both variants of `FsCreationSource` include `BlockDevice`, which feels a bit odd at first glance. Would it be better to use the following form instead?

```rust
pub struct FsCreationCtx<'a> {
    source: Option<&'a str>,
    flags: FsFlags,
    args: Option<&'a str>,
    block_device: BlockDeviceResolution<'a>,
}

 enum BlockDeviceResolution<'a> {
      Pending(&'a Context<'a>),
      Resolved(Arc<dyn BlockDevice>),
  }
```

kernel/src/init.rs
```diff
@@ -127,26 +138,156 @@ fn ap_idle_loop() {
...
 
+fn prepare_boot_init(path_resolver: &PathResolver) -> BootInit {
+    if let Some(init_path) = find_initramfs_init(path_resolver) {
+        return BootInit::Initramfs(init_path);
+    }
+
+    let mnt_ns = create_rootfs_mount_namespace();
+    let init_path = INIT_PATH.get().map(String::as_str);
+    BootInit::RootFs { mnt_ns, init_path }
+}
```
cchanging: I think the logics for preparing the `RootFs` should be also put in `rootfs.rs`? On the one hand, the filename itself is `rootfs`; on the other hand, the preparation for its peer `initramfs` is also placed under `rootfs`. Or, to make the distinction clearer, should we move the `initramfs` logic into `initramfs.rs`?

kernel/src/init.rs
```diff
@@ -127,26 +138,156 @@ fn ap_idle_loop() {
...
+            } => {
+                println!("[kernel] running {} as the rootfs init", init_path);
+                spawn_init_process(mnt_ns, init_path, argv, envp)
+            }
+            BootInit::RootFs {
+                mnt_ns,
+                init_path: None,
+            } => spawn_default_rootfs_init(mnt_ns, argv, envp),
+        }
+        .expect("failed to run the init process")
```
cchanging: This logics can be wrapped as a method of `BootInit`.

kernel/src/init.rs
```diff
@@ -127,26 +138,156 @@ fn ap_idle_loop() {
...
+
+    let mnt_ns = create_rootfs_mount_namespace();
+    let init_path = INIT_PATH.get().map(String::as_str);
+    BootInit::RootFs { mnt_ns, init_path }
+}
+
+fn find_initramfs_init(path_resolver: &PathResolver) -> Option<&'static str> {
+    const DEFAULT_INITRAMFS_INIT_PATH: &str = "/init";
+
+    if let Some(init_path) = RDINIT_PATH.get().map(String::as_str) {
```
cchanging: This way will be better?

```rust
let init_path = RDINIT_PATH.get().map(String::as_str).unwrap_or(DEFAULT_INITRAMFS_INIT_PATH)
```

kernel/src/init.rs
```diff
@@ -172,3 +313,47 @@ pub(super) fn on_first_process_startup(ctx: &Context) {
...
+aster_cmdline::define_kv_param!("rootfstype", ROOTFS_TYPES);
+
+static INIT_PATH: Once<String> = Once::new();
+aster_cmdline::define_kv_param!("init", INIT_PATH);
+
+static ROOT_MOUNT_READ_ONLY: AtomicBool = AtomicBool::new(true);
+static RO_PARAM: SetRootMountReadOnly = SetRootMountReadOnly;
+aster_cmdline::define_flag_param!("ro", RO_PARAM);
+static RW_PARAM: SetRootMountReadWrite = SetRootMountReadWrite;
+aster_cmdline::define_flag_param!("rw", RW_PARAM);
```
cchanging: Would it be better to keep these `define_kv_param` variables together in one place within the file? Right now, they’re separated from `ROOT_PATH` and `RDINIT_PATH`.

kernel/src/fs/rootfs.rs
```diff
@@ -47,6 +70,49 @@ pub fn init_in_first_kthread(path_resolver: &PathResolver) -> Result<()> {
...
 
+/// Creates a new mount namespace rooted at the specified block-backed filesystem.
+pub fn create_mount_namespace(
+    root: &str,
+    rootfs_types: &[&dyn DynFsType],
+    mount_flags: PerMountFlags,
+) -> Result<Arc<MountNamespace>> {
+    // Treat `root=/dev/...` as a Linux-compatible root device spec, not as a
+    // VFS path lookup. Linux also mounts the root filesystem before
+    // auto-mounting devtmpfs on `/dev`.
```
cchanging: > Linux also mounts the root filesystem before auto-mounting devtmpfs on `/dev`.

This description is for the implementation details of Linux. We can just remove it since the reference is enough.

kernel/src/init.rs
```diff
@@ -127,26 +138,156 @@ fn ap_idle_loop() {
...
+
+impl ParamStorage for SetRootMountReadWrite {
+    type Value = bool;
+
+    fn store_param(&self, value: Self::Value) {
+        if value {
+            ROOT_MOUNT_READ_ONLY.store(false, Ordering::Relaxed);
+        }
+    }
+}
```
cchanging: Can we use `define_flag_param` directly just like:

```rust
static I8042_EXIST: AtomicBool = AtomicBool::new(false);
aster_cmdline::define_flag_param!("i8042.exist", I8042_EXIST);
```

kernel/src/fs/rootfs.rs
```diff
@@ -47,6 +70,49 @@ pub fn init_in_first_kthread(path_resolver: &PathResolver) -> Result<()> {
...
+    let device_name = root
+        .strip_prefix("/dev/")
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "root must name a /dev block device"))?;
+    let device = aster_block::lookup_by_name(device_name)
+        .ok_or_else(|| Error::with_message(Errno::ENODEV, "root block device not found"))?;
+    let fs_flags = FsFlags::from_bits_truncate(mount_flags.bits());
+    let fs_and_root = open_rootfs_from_candidates(device, rootfs_types, fs_flags)?;
+
+    let owner = UserNamespace::get_init_singleton().clone();
+    let mount_namespace = MountNamespace::new_with_root(owner, |weak_ns| {
```
cchanging: To be honest, I don’t quite understand why it’s necessary to create an additional mount namespace here. Why can’t we just use the initial mount namespace?

cchanging (commented):
Thanks for the revision!

kernel/src/fs/rootfs.rs
taosue: The initial mount namespace is already rooted at the bootstrap **RamFs** because it is created before boot-source selection and is used to unpack and inspect the initramfs. See: https://github.com/asterinas/asterinas/blob/0c0b37ca6c437aab1ae65f744504912cc50229e3/kernel/src/fs/vfs/path/mount_namespace.rs#L93-L103

kernel/src/init.rs
taosue: The direct form works for `i8042.exist` because one flag directly represents one `Boolean` state. `ro` and `rw` are different: they are two parameter names controlling the same state with opposite meanings, and they require last-wins semantics across the two names.

Preserving this cross-key ordering is exactly the use case addressed by #3407, so `ro` and `rw` cannot both use `define_flag_param!` with an `AtomicBool` directly.

kernel/src/fs/rootfs.rs
cchanging: My concern is that we are creating an extra mount namespace here to serve as the post-boot mount namespace, which means the initial namespace becomes useless after boot, yet its associated memory will not be released.

I looked into Linux’s behavior, and it seems that Linux does not create a new namespace here. Instead, it mounts the new `rootfs` at `/root` (https://github.com/torvalds/linux/blob/master/init/do_mounts.c#L157), and then performs internal operations to switch to the new `rootfs` (https://github.com/torvalds/linux/blob/master/init/do_mounts.c#L486).

In theory, we should be able to do something similar. One additional requirement is that, like Linux, we would need to mount a `null` mount on top of the original root in the initial mount namespace in order to satisfy the requirements of `pivot_root`.

kernel/src/init.rs
cchanging: Thanks for the explanation! I just feel this implementation is somewhat complicated.

At least `SetRootMountReadOnly` is not necessary? We can directly:
```rust
define_flag_param!("ro", ROOT_MOUNT_READ_ONLY);
```

But this makes things slightly inconsistent. Actually, I feel it might be better to slightly extend the macro to support a similar pattern. However, I haven’t looked into whether there are other similar use cases besides the `ro`/`rw` case here. If not, extending the macro in a more complex way might be a bit of an overkill, so keeping the current approach would also be acceptable.

kernel/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -68,7 +71,10 @@ impl MountNamespace {
     /// The closure receives a `Weak<Self>` so that mounts in the new tree can
     /// reference the namespace being constructed. Construction uses `UniqueArc`
     /// to allow mutable initialization while still providing `Weak` references.
-    fn new_with_root<F>(owner: Arc<UserNamespace>, build_root_fn: F) -> Result<Arc<Self>>
+    pub(in crate::fs) fn new_with_root<F>(
```
cchanging: Now this method can be private.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -282,20 +282,23 @@ impl Mount {
     ///
     /// It is allowed to create a mount node even if the fs has been provided to another
     /// mount node. It is the fs's responsibility to ensure the data consistency.
+    #[cfg(ktest)]
     pub(in crate::fs) fn new_root(
         fs: Arc<dyn FileSystem>,
         mnt_ns: Weak<MountNamespace>,
     ) -> Result<Arc<Self>> {
+        Self::new_root_with_flags(FsAndRoot::new(fs), PerMountFlags::default(), mnt_ns)
+    }
```
cchanging: Since this API is only used in ktest, I think we can remove this API and rename `new_root_with_flags` with `new_root`.

kernel/core/src/fs/vfs/path/mount_namespace.rs
```diff
@@ -95,10 +101,31 @@ impl MountNamespace {
...
+                .do_mount(
+                    FsAndRoot::new(rootfs),
+                    PerMountFlags::default(),
+                    namespace.root().root_dentry(),
+                    Some("rootfs".to_string()),
+                    &mut topology_guard,
+                )
+                .expect("failed to mount rootfs on nullfs");
+
+            namespace
```
cchanging, line 121: We can obtain `namespace.root()` first and reuse it.

kernel/src/fs/vfs/path/resolver.rs
```diff
@@ -361,6 +361,36 @@ impl PathResolver {
...
+        let mut topology_guard = MountTopology::write_lock();
+        let old_root = self.root.mount.clone();
+        let old_root_parent = old_root
+            .parent()
+            .and_then(|parent| parent.upgrade())
+            .expect("the bootstrap root should have a parent mount");
+        let old_root_mountpoint = old_root
+            .mountpoint()
+            .expect("the bootstrap root should have a mountpoint");
+        let target_path = Path::new(old_root_parent, old_root_mountpoint);
```
cchanging: Follow the style in `pivot_root` is better?

```rust
        let parent_path = {
            let parent_mount = self.root.mount.parent().unwrap().upgrade().unwrap();
            let mountpoint = self.root.mount.mountpoint().unwrap();
            Path::new(parent_mount, mountpoint)
        };
```

kernel/src/fs/rootfs.rs
```diff
@@ -1,119 +1,213 @@
...
+        let mut topology_guard = MountTopology::write_lock();
+        let mount = mountpoint.mount_node().do_mount(
+            fs_and_root,
+            mount_flags,
+            mountpoint.dentry(),
+            Some(root.to_string()),
+            &mut topology_guard,
+        )?;
+        Path::new_fs_root(mount)
+    };
```
cchanging: Use `Path::mount()` can avoid make the `MountTopology` public to the whole `crate::fs`?

Or if you want to save some unnecessary checks, I think it is better to find some way to wrap related operations into the `path` module.

kernel/src/init.rs
```diff
@@ -134,17 +143,83 @@ fn first_kthread() {
...
+            path_resolver,
+            init_path: Some(init_path),
+        };
+    }
+
+    rootfs::switch_to_rootfs(&mut path_resolver)
+        .expect("neither an initramfs init nor a usable root filesystem was available");
+    BootInit {
+        path_resolver,
+        init_path: rootfs::find_init(),
```
cchanging: `initramfs::find_init` and `rootfs::find_init()` are not really symmetrical: `initramfs::find_init` includes extra logic to check whether the target actually exists. Could we also make `rootfs::find_init()` perform the actual lookup, and then have `BootInit` uniformly store `init_path: Option<Path>`? That would avoid subsequent repeated lookups, for example in the `initramfs` branch.

test/initramfs/src/regression/fs/isolation/pivot_root.c
```diff
@@ -62,13 +62,6 @@ FN_TEST(pivot_root)
 		 CHROOT_DIR, PIVOT_TARGET_DIR);
 	ensure_dir(pivot_target_full_path);
 
-	// Negative test: pivot_root should fail if the root mount is the rootfs mount.
-	// We skip this in Linux because common test environments (e.g., Docker) do not
-	// use rootfs as the root mount.
-#ifdef __asterinas__
-	TEST_ERRNO(syscall(SYS_pivot_root, CHROOT_DIR, CHROOT_DIR), EINVAL);
-#endif
-
```
cchanging, line 71: Can we change this to a positive test?

kernel/core/src/init.rs
```diff
@@ -34,6 +37,12 @@ pub(super) fn main() {
         .spawn();
 }
 
+pub(super) fn on_first_process_startup(ctx: &Context) {
+    component::init_all(InitStage::Process, component::parse_metadata!()).unwrap();
+    crate::device::init_in_first_process(ctx).unwrap();
+    crate::fs::init_in_first_process(ctx);
+}
+
```
cchanging, line 45: Why move this method after the `main`? I couldn’t think of a good reason.

kernel/core/src/fs/rootfs.rs
```diff
@@ -1,119 +1,213 @@
...
+
+    fn store_param(&self, value: Self::Value) {
+        if value {
+            ROOT_MOUNT_READ_ONLY.store(false, Ordering::Relaxed);
+        }
+    }
 }
+
+static RW_PARAM: SetRootMountReadWrite = SetRootMountReadWrite;
+aster_cmdline::define_flag_param!("rw", RW_PARAM);
```
cchanging, line 203:
```rust
static ROOT_MOUNT_READ_ONLY: AtomicBool = AtomicBool::new(true);

struct SetRootMountReadOnly;
struct SetRootMountReadWrite;

impl ParamStorage for SetRootMountReadOnly {
    type Value = bool;

    fn store_param(&self, value: Self::Value) {
        if value {
            ROOT_MOUNT_READ_ONLY.store(true, Ordering::Relaxed);
        }
    }
}

impl ParamStorage for SetRootMountReadWrite {
    type Value = bool;

    fn store_param(&self, value: Self::Value) {
        if value {
            ROOT_MOUNT_READ_ONLY.store(false, Ordering::Relaxed);
        }
    }
}

static RO_PARAM: SetRootMountReadOnly = SetRootMountReadOnly;
static RW_PARAM: SetRootMountReadWrite = SetRootMountReadWrite;
aster_cmdline::define_flag_param!("ro", RO_PARAM);
aster_cmdline::define_flag_param!("rw", RW_PARAM);
```

or

```rust
static ROOT_MOUNT_READ_ONLY: AtomicBool = AtomicBool::new(true);
aster_cmdline::define_flag_param!("ro", ROOT_MOUNT_READ_ONLY);

struct SetRootMountReadWrite;

impl ParamStorage for SetRootMountReadWrite {
    type Value = bool;

    fn store_param(&self, value: Self::Value) {
        if value {
            ROOT_MOUNT_READ_ONLY.store(false, Ordering::Relaxed);
        }
    }
}

static RW_PARAM: SetRootMountReadWrite = SetRootMountReadWrite;
aster_cmdline::define_flag_param!("rw", RW_PARAM);
```

Will these ways be better?

cchanging (commented):
Thanks for the updating!

kernel/src/init.rs
taosue: Resolving the init path early makes sense, but we should retain the original pathname alongside the resolved `Path`. Linux uses the configured pathname as `argv[0]`, even when it is a symlink.              
                                                    
The default rootfs fallback should remain unchanged because each candidate must be actually executed in order; a successful lookup does not guarantee successful execution.

kernel/core/src/init.rs
taosue: I moved it to follow our top-down-reading guideline, which asks us to place public entry points before private helpers.

test/initramfs/src/regression/fs/isolation/pivot_root.c
taosue: Making it positive here would change the test process’s root and invalidate the remaining steps. Since `pivot_root_dot_dot` already covers the same behavior, would it be acceptable to simply remove this obsolete negative assertion?

cchanging: Thanks for the explanation! That makes sense. I forgot that `pivot_root_dot_dot` already covers the relevant behavior.

kernel/src/fs/rootfs.rs
```diff
@@ -1,119 +1,211 @@
...
+/// and working directory, and the bootstrap root is detached.
+///
+/// Returns an error if `root` is not provided or the root filesystem cannot be opened or mounted.
+pub fn switch_to_rootfs(path_resolver: &mut PathResolver) -> Result<()> {
+    const ROOTFS_MOUNTPOINT: &str = "/root";
+
+    let root = ROOT_PATH.get().ok_or_else(|| {
+        Error::with_message(Errno::EINVAL, "the `root` parameter was not provided")
+    })?;
+    let mount_flags = root_mount_flags();
```
cchanging: I suggest rename `root_mount_flags` to `rootfs_flags` and let it return `(FsFlags, PerMountFlags)`.

Because `FsFlags` is not fundamentally derived from `PerMountFlags`—the two are relatively independent—I think it’s fine to obtain them separately based on whether it is `ro` or not. Using `let fs_flags = FsFlags::from_bits_truncate(mount_flags.bits());` feels ambiguous.

kernel/src/fs/rootfs.rs
```diff
@@ -1,119 +1,211 @@
...
+            mkmod!(u+rwx),
+        )?,
+        Err(error) => return Err(error),
+    };
+
+    let mount = mountpoint.do_mount(fs_and_root, mount_flags, Some(root.to_string()))?;
+    let new_root = Path::new_fs_root(mount);
+    path_resolver.switch_root_for_boot(new_root);
+    println!("[kernel] rootfs is ready (mounted from {})", root);
+    Ok(())
```
cchanging: Sorry I forgot the `Path::mount` requires the parameter `ctx`, but here add a `do_mount` API for `Path` is not ideal —— it has almost the same name with `mount` and we cannot distinguish their differences easily.

I think a better way is use `Mount::new_root` to create a `rootfs_mount` here and pass both `mountpoint` and `rootfs_mount` to `path_resolver.switch_root_for_boot`. In this way we can obtain topology only once in `switch_root_for_boot` and use `Mount::graft_mount_tree` to attach the `Mount`.

But this thinking raises another question for me: why do we first mount this rootfs mount at "/root"? Since this is kernel-side initialization logic, we could just detach the original child mount of nullfs, attach the rootfs mount directly onto nullfs, and then set up the `PathResolver`.

I can’t think of a good reason for doing it the current way. At the moment, my impression is that Linux’s main motivation for doing this may simply be code reuse, but I’m not sure that is particularly compelling here, since the most straightforward approach would also require only a small amount of code.

What do you think? Right now, I’m actually more inclined toward handling it directly.

taosue: I agree that the direct approach is better. Both approaches produce the same final mount tree, while the intermediate state where the root filesystem is mounted at `/root` cannot be observed by userspace because it occurs before the first userspace process is created.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -283,19 +283,13 @@ impl Mount {
     /// It is allowed to create a mount node even if the fs has been provided to another
     /// mount node. It is the fs's responsibility to ensure the data consistency.
     pub(in crate::fs) fn new_root(
-        fs: Arc<dyn FileSystem>,
+        fs_and_root: FsAndRoot,
```
cchanging: I looked at the relevant usages, and it seems there may not be a need to change `fs` to `fs_and_root`, since the current usages only hold `fs` anyway.

cchanging (approved):
One small issue, otherwise LGTM!

@cqs21 @StevenJiang1110 Could you help review the related NixOS changes?

cqs21 (approved):

StevenJiang1110 (approved):

taosue:
@tatetian Could you help review this PR?

PR #3426, opened 2026-06-23
Support SiFive UART console

le-monde-bleu:
## Summary
This PR improves Asterinas support for QEMU's `sifive_u` machine, which emulates the SiFive HiFive Unleashed board based on the FU540-C000 SoC.
This PR is also a reimplementation of the unfinished work in #2529 , with the support and under the framework of #2837 .

## Background
Asterinas already had RISC-V boot support and NS16550A UART support, but QEMU's `sifive_u` machine exposes a SiFive UART rather than an NS16550A-compatible UART. As a result, the kernel could print early output and the Asterinas banner, but the boot flow would not reliably reach an interactive shell through the expected serial console path.

## References
- [QEMU RISC-V board model documentation](https://www.qemu.org/docs/master/system/target-riscv.html#choosing-a-board-model)
- [QEMU `sifive_u` board documentation](https://www.qemu.org/docs/master/system/riscv/sifive_u.html)
- [SiFive HiFive Unleashed documentation](https://www.sifive.com/boards/hifive-unleashed#documentation)
- [SiFive FU540-C000 Manual](https://pdos.csail.mit.edu/6.828/2025/readings/FU540-C000-v1.0.pdf)

The FU540-C000 manual describes the UART instances, MMIO register layout, TX/RX control registers, interrupt enable/pending registers, and baud-rate divisor register. The implementation follows those register definitions for `txdata`, `rxdata`, `txctrl`, `rxctrl`, `ie`, `ip`, and `div`.
The QEMU `sifive_u` documentation also notes that QEMU generates a DTB for the guest and that guest software should discover devices through that DTB. This PR follows that model by discovering the UART from the device tree rather than hardcoding its MMIO address.

## Main Behavior
- Discover a SiFive UART from the RISC-V device tree.
- Support both `sifive,uart0` and `sifive,fu540-c000-uart` compatible strings.
- Register the SiFive UART as the kernel serial console.
- Enable RX interrupts so the interactive shell can receive input.
- Use `ttyS0` as the default console for the `sifive_u` scheme.
- Avoid expensive initramfs gzip decompression for `sifive_u`.
- Make optional test block-device mounts conditional.

## Testing
The PR is validated by the following check:
```
make run_kernel TARGET_ARCH=riscv64 SCHEME=sifive_u
```
Before this PR, the kernel boot halts without interactive shell prompt. Now, the kernel boots on QEMU `sifive_u` scheme normally.
The following checks have been made:
```
sh -n test/initramfs/src/init
cargo fmt --check
CARGO_TARGET_DIR=/tmp/asterinas-cargo-target cargo check -p aster-uart --target riscv64imac-unknown-none-elf
make run_kernel TARGET_ARCH=riscv64 SCHEME=riscv RELEASE=1 AUTO_TEST=boot
```
The generic RISC-V boot test was included to make sure the existing NS16550A-based riscv scheme still works after adding the SiFive-specific UART path.

## Gaps and Future Work
This PR enables the current QEMU sifive_u boot path and interactive serial console, but there are still follow-up items for fuller HiFive Unleashed / FU540 support:
- The UART clock handling is still incomplete for real FU540-style device trees. The driver reads a direct `clock-frequency` property when available, but real FU540 device trees may provide the UART clock through the PRCI clock provider instead. A future PR should parse the `clocks` phandle or add proper PRCI clock support.
- The `500 MHz` UART clock fallback is suitable for the current HiFive Unleashed/QEMU path, but it should not be treated as a general SiFive UART clock solution.
- Exiting the init shell still reaches the current poweroff failure path on `sifive_u`. This PR intentionally does not work around that by forcing reboot; proper shutdown support should be addressed separately.
- This PR focuses on the serial console path. Other HiFive Unleashed peripherals, such as SD card, Ethernet, SPI, GPIO, DMA, and full PRCI support, remain future work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,11 @@ COVERAGE ?= 0
 # Asterinas will automatically fall back to tty0 if hvc0 is not available.
 # Note that currently the virtual terminal (tty0) can only work with
 # linux-efi-handover64 and linux-efi-pe64 boot protocol.
+ifeq ($(SCHEME), sifive_u)
+CONSOLE ?= ttyS0
+else
 CONSOLE ?= hvc0
+endif
 # End of global build options.
 
 # GDB debugging and profiling options.
--- a/kernel/comps/uart/src/arch/riscv/mod.rs
+++ b/kernel/comps/uart/src/arch/riscv/mod.rs
@@ -3,11 +3,17 @@
 use ostd::arch::boot::DEVICE_TREE;
 
 mod ns16550a;
+mod sifive;
 
 pub(super) fn init() {
     let device_tree = DEVICE_TREE.get().unwrap();
 
-    if let Some(ns16550a_node) = device_tree.find_compatible(&["ns16550a"]) {
+    if let Some(ns16550a_node) = device_tree.find_compatible(&ns16550a::FDT_COMPATIBLES) {
         ns16550a::init(ns16550a_node);
+        return;
+    }
+
+    if let Some(sifive_node) = device_tree.find_compatible(&sifive::FDT_COMPATIBLES) {
+        sifive::init(sifive_node);
     }
 }
--- a/kernel/comps/uart/src/arch/riscv/sifive.rs
+++ b/kernel/comps/uart/src/arch/riscv/sifive.rs
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! SiFive UART console support.
+
+use alloc::string::ToString;
+
+use aster_util::{field_ptr, safe_ptr::SafePtr};
+use fdt::node::FdtNode;
+use ostd::{
+    arch::irq::{self, InterruptSourceInFdt, MappedIrqLine},
+    io::IoMem,
+    irq::IrqLine,
+    sync::{LocalIrqDisabled, SpinLock},
+};
+use spin::Once;
+
+use crate::{
+    CONSOLE_NAME,
+    console::{Uart, UartConsole, UartMut},
+};
+
+pub(super) const FDT_COMPATIBLES: [&str; 2] = ["sifive,uart0", "sifive,fu540-c000-uart"];
+
+/// IRQ line for UART serial.
+static IRQ_LINE: Once<MappedIrqLine> = Once::new();
+
+pub(super) fn init(fdt_node: FdtNode) {
+    let Some(reg) = fdt_node.reg().and_then(|mut regs| regs.next()) else {
+        ostd::info!("Failed to read 'reg' property from SiFive UART node");
+        return;
+    };
+    let Some(reg_size) = reg.size else {
+        ostd::info!("Incomplete 'reg' property found in SiFive UART node");
+        return;
+    };
+
+    let reg_addr = reg.starting_address as usize;
+    let Some(reg_end) = reg_addr.checked_add(reg_size) else {
+        ostd::info!("Invalid I/O memory range found in SiFive UART node");
+        return;
+    };
+    let Ok(io_mem) = IoMem::acquire(reg_addr..reg_end) else {
+        ostd::info!("I/O memory is not available for SiFive UART");
+        return;
+    };
+
+    let Some(intr_parent) = fdt_node
+        .property("interrupt-parent")
+        .and_then(|prop| prop.as_usize())
+    else {
+        ostd::info!("Failed to read 'interrupt-parent' property from SiFive UART node");
+        return;
+    };
+    let Some(intr) = fdt_node.interrupts().and_then(|mut intrs| intrs.next()) else {
+        ostd::info!("Failed to read 'interrupts' property from SiFive UART node");
+        return;
+    };
+
+    let Ok(mut irq_line) = IrqLine::alloc().and_then(|irq_line| {
+        irq::IRQ_CHIP.get().unwrap().map_fdt_pin_to(
+            InterruptSourceInFdt {
+                interrupt_parent: intr_parent as u32,
+                interrupt: intr as u32,
+            },
+            irq_line,
+        )
+    }) else {
+        ostd::info!("IRQ line is not available for SiFive UART");
+        return;
+    };
+
+    let clock_hz = fdt_node
+        .property("clock-frequency")
+        .and_then(|prop| prop.as_usize())
+        .and_then(|freq| u32::try_from(freq).ok())
+        // FIXME: Real FU540 device trees may provide the UART clock through
+        // a PRCI clock provider instead of a direct `clock-frequency` property.
+        .unwrap_or(SifiveUart::DEFAULT_CLOCK_HZ);
+
+    let mut uart: SpinLock<SifiveUart, LocalIrqDisabled> = SpinLock::new(SifiveUart::new(io_mem));
+    uart.get_mut().init(clock_hz);
+
+    let uart_console = UartConsole::new(uart);
+
+    aster_console::register_device(CONSOLE_NAME.to_string(), uart_console.clone());
+
+    let cloned_uart_console = uart_console.clone();
+    irq_line.on_active(move |_| cloned_uart_console.trigger_input_callbacks());
+    IRQ_LINE.call_once(move || irq_line);
+    uart_console.uart().flush();
+
+    ostd::info!("Registered SiFive UART as a console");
+}
+
+struct SifiveUart {
+    registers: SafePtr<Registers, IoMem>,
+}
+
+impl SifiveUart {
+    const DEFAULT_CLOCK_HZ: u32 = 500_000_000;
+    const TARGET_BAUD: u32 = 115_200;
+
+    const TXDATA_FULL: u32 = 1 << 31;
+    const RXDATA_EMPTY: u32 = 1 << 31;
+    const TXCTRL_TXEN: u32 = 1;
+    const RXCTRL_RXEN: u32 = 1;
+    const IE_RXWM: u32 = 1 << 1;
+
+    fn new(io_mem: IoMem) -> Self {
+        Self {
+            registers: SafePtr::new(io_mem, 0),
+        }
+    }
+
+    fn init(&mut self, clock_hz: u32) {
+        let divisor = clock_hz
+            .checked_div(Self::TARGET_BAUD)
+            .and_then(|quotient| quotient.checked_sub(1))
+            .unwrap_or(0);
+
+        field_ptr!(&self.registers, Registers, div)
+            .write_once(&divisor)
+            .unwrap();
+        field_ptr!(&self.registers, Registers, txctrl)
+            .write_once(&Self::TXCTRL_TXEN)
+            .unwrap();
+        field_ptr!(&self.registers, Registers, rxctrl)
+            .write_once(&Self::RXCTRL_RXEN)
+            .unwrap();
+        field_ptr!(&self.registers, Registers, ie)
+            .write_once(&Self::IE_RXWM)
+            .unwrap();
+    }
+}
+
+impl UartMut for SifiveUart {
+    fn send_byte(&mut self, byte: u8) {
+        let txdata = field_ptr!(&self.registers, Registers, txdata);
+        while txdata.read_once().unwrap() & Self::TXDATA_FULL != 0 {
+            core::hint::spin_loop();
+        }
+
+        txdata.write_once(&u32::from(byte)).unwrap();
+    }
+
+    fn recv_byte(&mut self) -> Option<u8> {
+        let rxdata = field_ptr!(&self.registers, Registers, rxdata)
+            .read_once()
+            .unwrap();
+        if rxdata & Self::RXDATA_EMPTY != 0 {
+            return None;
+        }
+
+        Some(rxdata as u8)
+    }
+}
+
+// Register layout and bit definitions are from the SiFive FU540-C000 Manual,
+// Chapter 13 "Universal Asynchronous Receiver/Transmitter (UART)".
+// Reference: <https://pdos.csail.mit.edu/6.828/2025/readings/FU540-C000-v1.0.pdf>
+#[repr(C)]
+struct Registers {
+    txdata: u32,
+    rxdata: u32,
+    txctrl: u32,
+    rxctrl: u32,
+    ie: u32,
+    _ip: u32,
+    div: u32,
+}
--- a/kernel/comps/uart/src/console.rs
+++ b/kernel/comps/uart/src/console.rs
@@ -86,7 +86,16 @@ pub(super) trait Uart {
     fn flush(&self);
 }
 
-impl<A: Ns16550aAccess> Uart for SpinLock<Ns16550aUart<A>, LocalIrqDisabled> {
+/// A trait that abstracts byte-wise UART device operations.
+pub(super) trait UartMut {
+    /// Sends one byte to UART.
+    fn send_byte(&mut self, byte: u8);
+
+    /// Receives one byte from UART.
+    fn recv_byte(&mut self) -> Option<u8>;
+}
+
+impl<T: UartMut> Uart for SpinLock<T, LocalIrqDisabled> {
     fn send(&self, buf: &[u8]) {
         let mut uart = self.lock();
 
@@ -94,9 +103,9 @@ impl<A: Ns16550aAccess> Uart for SpinLock<Ns16550aUart<A>, LocalIrqDisabled> {
             // TODO: This is termios-specific behavior and should be part of the TTY implementation
             // instead of the serial console implementation. See the ONLCR flag for more details.
             if *byte == b'\n' {
-                uart.send(b'\r');
+                uart.send_byte(b'\r');
             }
-            uart.send(*byte);
+            uart.send_byte(*byte);
         }
     }
 
@@ -104,7 +113,7 @@ impl<A: Ns16550aAccess> Uart for SpinLock<Ns16550aUart<A>, LocalIrqDisabled> {
         let mut uart = self.lock();
 
         for (i, byte) in buf.iter_mut().enumerate() {
-            let Some(recv_byte) = uart.recv() else {
+            let Some(recv_byte) = uart.recv_byte() else {
                 return i;
             };
             *byte = recv_byte;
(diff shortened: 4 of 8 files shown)


kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,219 @@
...
+        ostd::info!("Failed to read 'interrupt-parent' property from SiFive UART node");
+        return;
+    };
+    let Some(intr) = fdt_node.interrupts().and_then(|mut intrs| intrs.next()) else {
+        ostd::info!("Failed to read 'interrupts' property from SiFive UART node");
+        return;
+    };
+
+    let Some(irq_chip) = irq::IRQ_CHIP.get() else {
+        ostd::info!("IRQ chip is not available for SiFive UART");
```
lrh2000: Be consistent with existing code. `IRQ_CHIP.get().unwrap()` is fine because we cannot have a working system without an IRQ chip.
https://github.com/asterinas/asterinas/blob/e80904666fcc97085605f1e8118b86bc738c13c8/kernel/comps/uart/src/arch/riscv/ns16550a.rs#L67-L68

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,219 @@
...
+        ostd::info!("Failed to read 'reg' property from SiFive UART node");
+        return;
+    };
+    let Some(reg_size) = reg.size else {
+        ostd::info!("Incomplete 'reg' property found in SiFive UART node");
+        return;
+    };
+
+    let reg_addr = reg.starting_address as usize;
+    let Some(reg_end) = reg_addr.checked_add(reg_size) else {
```
lrh2000, line 38: Using `checked_add` here is a good choice. Could you please also fix the code in `ns16550a.rs` so we don't end up with some unnecessary inconsistency?
https://github.com/asterinas/asterinas/blob/e80904666fcc97085605f1e8118b86bc738c13c8/kernel/comps/uart/src/arch/riscv/ns16550a.rs#L49-L50

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    let uart_console = UartConsole::new(uart);
+
+    aster_console::register_device(CONSOLE_NAME.to_string(), uart_console.clone());
+
+    let cloned_uart_console = uart_console.clone();
+    irq_line.on_active(move |_| cloned_uart_console.trigger_input_callbacks());
+    IRQ_LINE.call_once(move || irq_line);
+    uart_console.uart().flush();
+    uart_console.uart().lock().enable_rx_interrupt();
```
lrh2000: https://github.com/asterinas/asterinas/blob/e80904666fcc97085605f1e8118b86bc738c13c8/kernel/comps/uart/src/console.rs#L82-L86

`flush()` is used to ensure RX interrupts won't be missed. It should be called after `enable_rx_interrupt()`.

Otherwise, if a new character appears between the calls to `flush()` and `enable_rx_interrupt()`, it will remain in the device's receive buffer for a long time.
```rust
    uart_console.uart().flush();
    // <- a new character comes
    uart_console.uart().lock().enable_rx_interrupt();
```

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,219 @@
...
+            .unwrap_or(0);
+
+        self.write_register(offset_of!(Registers, div), divisor);
+        self.write_register(offset_of!(Registers, txctrl), Self::TXCTRL_TXEN);
+        self.write_register(offset_of!(Registers, rxctrl), Self::RXCTRL_RXEN);
+        self.write_register(offset_of!(Registers, ie), 0);
+    }
+
+    fn enable_rx_interrupt(&mut self) {
+        self.write_register(offset_of!(Registers, ie), Self::IE_RXWM);
```
lrh2000: We already allocate and map the IRQ line at `init()`. Why not just enable the RX interrupt by default?

Similar to `Ns16550aUart::init`, which enables interrupts in `init()`.

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,219 @@
...
+        self.io_mem
+            .read_once(offset)
+            .expect("SiFive UART register offset should be within the acquired I/O memory")
+    }
+
+    fn write_register(&mut self, offset: usize, value: u32) {
+        self.io_mem
+            .write_once(offset, &value)
+            .expect("SiFive UART register offset should be within the acquired I/O memory");
+    }
```
lrh2000: We can use:
https://github.com/asterinas/asterinas/blob/e80904666fcc97085605f1e8118b86bc738c13c8/kernel/libs/aster-util/src/safe_ptr.rs#L439-L441

Example usage:
https://github.com/asterinas/asterinas/blob/e80904666fcc97085605f1e8118b86bc738c13c8/kernel/comps/virtio/src/queue.rs#L328-L334

No need to use `expect`. `Option::unwrap` for `read_once` and `write_once` should be fine.

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    fn flush(&self) {
+        let mut uart = self.lock();
+
+        while uart.recv_byte().is_some() {}
+    }
+}
+
+#[repr(C)]
+struct Registers {
```
lrh2000, line 162: Could you provide a reference? Is there a manual, specification, or datasheet for the SiFive UART device that explains the meaning of the registers? Add it as a comment.

Makefile
```diff
@@ -28,7 +28,14 @@ COVERAGE ?= 0
...
 # Note that currently the virtual terminal (tty0) can only work with
 # linux-efi-handover64 and linux-efi-pe64 boot protocol.
+ifeq ($(SCHEME), sifive_u)
+CONSOLE ?= ttyS0
+INITRAMFS_SKIP_GZIP ?= 1
+INIT_EXIT_ACTION ?= reboot
+else
 CONSOLE ?= hvc0
+INIT_EXIT_ACTION ?= poweroff
+endif
```
lrh2000, line 35: > INIT_EXIT_ACTION=reboot works around the lack of a working poweroff path for this scheme. Proper poweroff support can be revisited separately if the board model exposes a usable shutdown mechanism.

This is strange. I don't think this workaround is worth it. I think it should be removed from this PR.

lrh2000 (commented):
Thanks for continuing the work in https://github.com/asterinas/asterinas/pull/2529.

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
The 
```rust
let Some(irq_chip) = irq::IRQ_CHIP.get() else { ... };
``` 
is now replaced by 
```rust
irq::IRQ_CHIP.get().unwrap().map_fdt_pin_to(...)
```

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
The relevant code in `ns165501.rs` is fixed, using `checked_add`.

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
I will fix this along with the next one, by removing the `enable_rx_interrupt()` here, putting
```rust
self.write_register(offset_of!(Registers, ie), Self::IE_RXWM);
```
into `SifiveUart::init()`. So that the initialization order goes like:
```rust
uart.get_mut().init(clock_hz);

register console
install irq callback
store irq line
uart_console.uart().flush();
```
now.

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
The fix is along with the one above.

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
I will fix this by adjusting the `SiFiveUart` and using:
```rust
field_ptr!(&self.registers, Registers, txdata).read_once().unwrap()
field_ptr!(&self.registers, Registers, div).write_once(&divisor).unwrap()
```

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
The reference comes from the [SiFive FU540-C000 Manual](https://pdos.csail.mit.edu/6.828/2025/readings/FU540-C000-v1.0.pdf), in chapter 13, Table 58, and some following tables. I will add a comment here.

Makefile
le-monde-bleu: Thanks for your revision!
It is true that this workaround is not for this PR. This workaround is for solving the power off failure problem of `sifive_u` scheme. To be more specific, in the docker container, when running:
```shell
make run_kernel TARGET_ARCH=riscv64 SCHEME=sifive_u
```
the boot process does normally and reaches the interactive shell, however, when trying to `exit`, the following problem appears:
```shell
ERROR : Failed to power off the system because the poweroff handler fails
ERROR : Halting the machine...
```
This is because there is no applicable poweroff backend under `sifive_u` scheme on QEMU. The FU540-C000 manual documents reset, clocking, and peripheral reset control through PRCI, but it does not define a software poweroff mechanism. Therefore, the current `sifive_u` poweroff failure after exiting `init` is a separate platform shutdown-support gap, not part of the UART console support added by this PR.
I shall remove this workaround and leave this problem as a separate gap for future work. However, the power off failure problem above will emerge again.
To fix this, we need to study and choose the software mechanism in `sifive_u` scheme.

Makefile
```diff
@@ -28,7 +28,12 @@ COVERAGE ?= 0
 # Asterinas will automatically fall back to tty0 if hvc0 is not available.
 # Note that currently the virtual terminal (tty0) can only work with
 # linux-efi-handover64 and linux-efi-pe64 boot protocol.
+ifeq ($(SCHEME), sifive_u)
+CONSOLE ?= ttyS0
+INITRAMFS_SKIP_GZIP ?= 1
```
lrh2000: 1. `INITRAMFS_SKIP_GZIP` is defined at a wrong location (it is not related to the console parameter).
2. The definition lacks of docs to explain why.
3. Also I don't believe it is related to `sifive_u`. It is slow for all non-native archs (e.g., what about other RISC-V/LoongArch/ARM platforms), right?

kernel/comps/uart/src/arch/riscv/mod.rs
```diff
@@ -3,10 +3,21 @@
...
+mod sifive;
 
 pub(super) fn init() {
     let device_tree = DEVICE_TREE.get().unwrap();
 
+    if let Some(sifive_node) = device_tree
+        .chosen()
+        .stdout()
+        .filter(|node| sifive::is_compatible(node))
+        .or_else(|| device_tree.find_compatible(&sifive::FDT_COMPATIBLES))
```
lrh2000: Is checking for `stdout` really necessary?

I think the `ttyS*` nodes should be created regardless of whether "stdout" is specified. Ideally, if we have multiple Sifive UARTs and NS16550A UARTs, we should create `ttyS*` nodes for each device.

For now, the problem is that we can only create one ttyS0 node. Therefore, preferring "stdout" might not be a bad idea. However, the current code does not make much sense because you still prefer the Sifive UART in the following case:
 - "stdout" -> a NS16559A node (does not pass the check in line 14)
 - a Sifive UART node (pass the check in line 15)

Either fix this case, or just prefer a simpler solution to follow the existing code:
```rust
    if let Some(ns16550a_node) = device_tree.find_compatible(&["ns16550a"]) {
        ns16550a::init(ns16550a_node);
    }
```

kernel/comps/uart/src/arch/riscv/mod.rs
```diff
@@ -3,10 +3,21 @@
...
+    {
+        sifive::init(sifive_node);
+        return;
+    }
+
     if let Some(ns16550a_node) = device_tree.find_compatible(&["ns16550a"]) {
         ns16550a::init(ns16550a_node);
```
lrh2000: For consistency:
 - Define `ns16550a::FDT_COMPATIBLES: [&str; 1]` as well.
 - Add a `return` (if the rust toolchain does not warn; otherwise ignore this comment).
```rust
    if let Some(ns16550a_node) = device_tree.find_compatible(&ns16550a::FDT_COMPATIBLES) {
        ns16550a::init(ns16550a_node);
        return;
```

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,213 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! SiFive UART console support.
+//!
+//! This module registers SiFive FU540-compatible UART devices with the
+//! architecture-independent [`UartConsole`] layer. It discovers the UART from
+//! the RISC-V device tree and maps the device interrupt through the platform
+//! IRQ chip.
```
lrh2000: This seems mostly unrelated to this module; it explains the facts for all architectures.

If you can find a more suitable location, move it there. Otherwise, just delete the redundant comments. In this case, a simple line is enough.
```rust
//! SiFive UART console support.
```

Also in `ns16550a.rs`:
```rust
//! NS16550A UART console support.
```

kernel/comps/uart/src/arch/riscv/sifive.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        buf.len()
+    }
+
+    fn flush(&self) {
+        let mut uart = self.lock();
+
+        while uart.recv_byte().is_some() {}
+    }
+}
```
lrh2000: Is there a way to avoid duplicating the code from `console.rs`? Define a trait like this?
```rust
trait UartMut {}

impl<T: UartMut> Uart for SpinLock<T> {}

impl<T: UartMut> Uart for &SpinLock<T> {}
```

lrh2000 (commented):
Thanks for updates!

Makefile
le-monde-bleu: Thanks for your revision!
Yes, this is added here because the test time is a little too long. I should have removed it.

kernel/comps/uart/src/arch/riscv/mod.rs
le-monde-bleu: Thanks for your revision!
I suppose I should follow the simpler solution, since there is no base for enumerating all UART devices and creating multiple `ttyS*`.

kernel/comps/uart/src/arch/riscv/mod.rs
le-monde-bleu: Thanks for your revision!
I will fix this.

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
I will fix this.

kernel/comps/uart/src/arch/riscv/sifive.rs
le-monde-bleu: Thanks for your revision!
I will try to implement this.

lrh2000 (approved):

---

I just realized that we really need a utility crate for device tree stuff. The similar logic is repeated all over the world:
1. `kernel/comps/uart/src/arch/riscv/ns16550a.rs`
2. `kernel/comps/uart/src/arch/riscv/sifive.rs`
3. `kernel/comps/time/src/rtc/goldfish.rs`
4. `kernel/comps/time/src/rtc/loongson.rs`
5. `kernel/comps/pci/src/arch/loongarch/mod.rs`
6. `kernel/comps/pci/src/arch/riscv/mod.rs`

They also need to acquire the I/O memory (and interrupt line if applicable) from the device tree.

The world will become much better if we just have something like:
```rust
trait FdtNodeExt {
    fn acquire_io_mems<const N: usize>(&self) -> Result<[IoMem; N]>;
    fn map_interrupt_pins<const N: usize>(&self) -> Result<[MappedIrqLine; N]>;
}

impl FdtNodeExt for FdtNode {}
```

All corner cases, including overflow checks and logging error messages, should be handled within the `FdtNodeExt` methods.

We will soon add support for ARM platforms, and I found that this is becoming a bigger problem. The current state is already quite bad, given the looooong list above.

However, this is a bit outside the scope of this PR anyway. So I don't think we should block this PR due to that.

le-monde-bleu:
@tatetian 
The merge of this PR has been delayed for a long time. I solved the emerged conflict and some minor semantic issues. Could you please check and review this PR?

le-monde-bleu:
@tatetian 
The merge of this PR has been delayed for a period of time. I solved the emerged conflicts and some minor issues. Could you please check and review?

kernel/comps/uart/src/console.rs
```diff
@@ -116,13 +125,23 @@ impl<A: Ns16550aAccess> Uart for SpinLock<Ns16550aUart<A>, LocalIrqDisabled> {
...
 
 #[inherit_methods(from = "(**self)")]
-impl<A: Ns16550aAccess> Uart for &SpinLock<Ns16550aUart<A>, LocalIrqDisabled> {
+impl<T: UartMut> Uart for &SpinLock<T, LocalIrqDisabled> {
     fn send(&self, buf: &[u8]);
     fn recv(&self, buf: &mut [u8]) -> usize;
     fn flush(&self);
 }
+
+impl<A: Ns16550aAccess> UartMut for Ns16550aUart<A> {
```
lrh2000, line 139: Now this trait should be used in `arm/mod.rs` as well?
https://github.com/asterinas/asterinas/blob/a5449e62b0a5a0affccb6087ea3543a2fdf66052/kernel/comps/uart/src/arch/arm/mod.rs#L33-L55

le-monde-bleu: Thank you for your supplement!
Yes indeed. I will apply the trait.

le-monde-bleu:
Updated the implementation for AArch64.
Removed the duplicated `Uart` implementation of `Pl011Uart` and implement `UartMut` for it.

lrh2000:
No other comments from me. I will simplify the code in https://github.com/asterinas/asterinas/pull/3821 anyway.

tatetian:
@le-monde-bleu Sorry for the delay in processing in PR. Please clean up the Git history; we do not allow merge commits.

le-monde-bleu:
> @le-monde-bleu Sorry for the delay in processing in PR. Please clean up the Git history; we do not allow merge commits.

Sorry for the mistake. The history is clean now.

le-monde-bleu:
The x86_64 osdk test kernel-dev is cancelled without reporting the cause. I re-checked the code but I am not sure why. Perhaps it is occasional?

PR #3431, opened 2026-06-25
Make getsockopt support short integer optlen buffers

ljb90071-ship-it:
Title: Make getsockopt support short integer optlen buffers

Branch: fix-getsockopt-partial-optlen

Summary:
- Allows integer-valued getsockopt options to return as many bytes as the caller-provided optlen can hold, matching Linux behavior for short buffers.
- Keeps setsockopt validation strict, so undersized input buffers still return EINVAL.
- Adds regression coverage for a two-byte SO_KEEPALIVE getsockopt buffer, a zero-length SO_ERROR buffer, and the unchanged undersized setsockopt failure.

Test plan:
- `git diff --check`
- `gcc -fsyntax-only -Wall -Wextra -Itest/initramfs/src/regression test/initramfs/src/regression/network/sockoption.c`
- `gcc -Wall -Wextra -Itest/initramfs/src/regression test/initramfs/src/regression/network/sockoption.c -o /tmp/asterinas_sockoption_regression`
- `/tmp/asterinas_sockoption_regression`

Pending upstream validation:
- `make kernel` and VM regression testing are blocked locally by Docker Hub TLS EOF when pulling `asterinas/asterinas:0.18.0-20260618`.
- Pinned Rust toolchain `nightly-2026-04-03` installation is blocked locally by TLS EOF from `static.rust-lang.org`.

PR #3460, opened 2026-06-26
Support short getsockopt integer buffers

ljb90071-ship-it:
## Summary

This resubmits the narrow getsockopt short-buffer fix from #3431.

Linux allows integer-valued getsockopt options to copy only the number of bytes provided by optlen and then report that copied length. Asterinas returned EINVAL when optlen was smaller than the integer size. This updates the integer output helpers to copy min(sizeof(value), optlen). The input path used by setsockopt still requires a full integer-sized buffer, so undersized setsockopt keeps returning EINVAL.

## Test Plan

- git diff --check
- gcc -fsyntax-only -Wall -Wextra -Itest/initramfs/src/regression test/initramfs/src/regression/network/sockoption.c
- gcc -Wall -Wextra -Itest/initramfs/src/regression test/initramfs/src/regression/network/sockoption.c -o /tmp/asterinas_sockoption_regression
- /tmp/asterinas_sockoption_regression
- cargo fmt --check
- VDSO_LIBRARY_DIR=/root/linux_vdso cargo check -p aster-kernel --target x86_64-unknown-none

### DIFF
--- a/kernel/src/util/net/options/utils.rs
+++ b/kernel/src/util/net/options/utils.rs
@@ -52,13 +52,10 @@ macro_rules! impl_read_write_for_32bit_type {
 
         impl WriteToUser for $pod_ty {
             fn write_to_user(&self, addr: Vaddr, max_len: u32) -> Result<usize> {
-                let write_len = size_of::<$pod_ty>();
+                let bytes = self.to_ne_bytes();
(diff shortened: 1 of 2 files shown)


ljb90071-ship-it:
Withdrawing this resubmission to slow down and re-review the change more carefully before any future submission.

Issue #3429, opened 2026-06-24
Support Asterinas as guest kernel in Firecracker microVMs


hburaylee:
### Feature Description

Firecracker is a lightweight, KVM-based Virtual Machine Monitor (VMM) optimized for serverless workloads, containers, and secure multi-tenancy. It expects a minimal guest kernel that boots quickly and works with a very small set of virtio devices.

**Expected Outcome:**
- Asterinas kernel image (`vmlinux`) can be directly loaded and booted by Firecracker.
- Full compatibility with Firecracker’s virtio device model (`virtio-net`, `virtio-block`, `virtio-vsock`, serial console, etc.).
- Preserve Asterinas’ Linux ABI compatibility so user-space applications and distributions (e.g., Alpine Linux) run seamlessly inside the microVM.
- Achieve fast cold-start times and low memory footprint consistent with Firecracker’s design goals.

---

### Problem or Need

Asterinas is already being developed and tested primarily with QEMU/KVM. However, there is currently no official validation or support for running it as a guest kernel in Firecracker — one of the most popular production microVM solutions.

**Why it matters:**
- Firecracker is widely used in cloud-native and serverless environments (AWS Lambda, Firecracker-based container runtimes, etc.).
- Asterinas’ memory-safe Rust implementation and small TCB can significantly improve the security posture of microVMs.
- This would broaden Asterinas’ applicability in production scenarios such as secure sandboxes, CI/CD isolation, edge computing, and function-as-a-service platforms.

**Current gaps**

| Area             | Current status                                               | Gap                                                          | PR                                                           | Completion Status |
| ---------------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------- |
| Direct boot      | Supports multiboot、multiboot2、linux                        | PVH direct boot                                              | [3623](https://github.com/asterinas/asterinas/pull/3623): Add firecracker bin in docker. <br /> [3461](https://github.com/asterinas/asterinas/pull/3461): Support PVH boot protocol. | **👀**             |
| virtio-mmio      | virtio-mmio transport exists                                 | Exposes virtio devices through virtio-mmio instead of PCI discovery | [cf9c093d0](https://github.com/asterinas/asterinas/commit/cf9c093d014cf47d18277cc0a640aac83adff884): Support virtio-mmio cmdline.(by @WangYue)<br /><br />[3465](https://github.com/asterinas/asterinas/pull/3465): Fix init sequence. | **✅**             |
[...]

PR #2395, opened 2025-09-01
Better x86 CPU feature detection by rewriting all CPUID-related code

lrh2000:
## Summary

This PR proposes the following two major changes:
 - Implement `arch/x86/cpu/cpuid.rs`, which replaces [the `cpuid` module from the `x86` crate](https://docs.rs/x86/latest/x86/cpuid/index.html).
   - In the future, we should not use the `cpuid` module from the `x86` crate anymore.
 - Implement `arch/x86/cpu/extension.rs`, which replaces [`arch::CPU_FEATURES`](https://github.com/asterinas/asterinas/blob/f56d2f3bad8ff769b50507356c0306488926d1cb/ostd/src/arch/x86/mod.rs#L53).
   - This mirrors the APIs of `arch/riscv/cpu/extension.rs`.

Since the generation of `/proc/cpuinfo` is closely related to the way to execute the CPUID instruction, it was also completely rewritten in this PR. I carefully examined the Linux implementation to ensure more accurate results than the previous implementation.

## Motivation

We previously detect x86 CPU features with the help of the `CPU_FEATURES` static variable:
https://github.com/asterinas/asterinas/blob/f56d2f3bad8ff769b50507356c0306488926d1cb/ostd/src/arch/x86/mod.rs#L53

This is far from being ideal due to the following reasons:
 - The [`FeatureInfo` type](https://docs.rs/x86/latest/x86/cpuid/struct.FeatureInfo.html) collects the feature information only from the CPUID 0x01 leaf, but there is other feature information from many other CPUID leaves. We can take a look at [the Linux implementation](https://github.com/torvalds/linux/blob/aaf724ed69264719550ec4f194d3ab17b886af9a/arch/x86/include/asm/cpufeatures.h):
```
/* Intel-defined CPU features, CPUID level 0x00000001 (EDX), word 0 */
/* Don't duplicate feature flags which are redundant with Intel! */
/* Intel-defined CPU features, CPUID level 0x00000001 (ECX), word 4 */
/* Intel-defined CPU features, CPUID level 0x00000007:0 (EBX), word 9 */
/* Intel-defined CPU features, CPUID level 0x00000007:1 (EAX), word 12 */
/* Intel-defined CPU features, CPUID level 0x00000007:0 (ECX), word 16 */
/* Intel-defined CPU features, CPUID level 0x00000007:0 (EDX), word 18 */

/* AMD-defined CPU features, CPUID level 0x80000001, word 1 */
[...]

PR #3461, opened 2026-06-26
Add Firecracker direct PVH boot support

hburaylee:
## Summary

This PR adds initial support for booting Asterinas on Firecracker with an initrd. 

It is an initial step towards supporting Asterinas as a guest kernel in Firecracker microVMs, see #3429 .

The current implementation is able to build a Firecracker-loadable kernel with:
```bash
make kernel BOOT_PROTOCOL=pvh
```

## Dependency

The Firecracker support depends on the following two PRs being merged first:
- #3623  : adds Firecracker to the Docker image to support the CI test.
- #3765  : adds the PVH boot protocol.

## Changes

- Use KVM pvclock as the preferred source for TSC frequency detection, since Firecracker only supports KVM clock.
- Supplement I/O port probing for the legacy PIC, since Firecracker's ACPI support is very limited. 
- Add a Firecracker CI boot test and usage documentation.
  
  
## Build And Run

Build:

```bash
make kernel BOOT_PROTOCOL=pvh
```

Expected kernel artifact:

```text
target/osdk/aster-kernel/aster-kernel-osdk-bin.elf
```

Firecracker binaries can be downloaded from:

```text
https://github.com/firecracker-microvm/firecracker/releases
```

Run with the Firecracker helper `fc-run.sh`:

<details>

```bash
#!/bin/bash

cat > vmconfig.json << EOF
{
  "boot-source": {
    "kernel_image_path": "target/osdk/aster-kernel/aster-kernel-osdk-bin.elf",
    "initrd_path": "$(ls target/osdk/aster-kernel/*-initramfs-image)",
    "boot_args": "console=ttyS0 earlycon loglevel=error i8042.exist"
  },
  "drives": [],
  "machine-config": {
    "vcpu_count": 2,
    "mem_size_mib": 1024
  }
}
EOF

"$FIRECRACKER" --config-file vmconfig.json
```

</details>

## Missing Work

- The current implementation requires adding the `i8042.exist` parameter to Firecracker's kernel command line, so ACPI-based probing for the i8042 controller needs to be added.
- The current implementation only obtains the TSC clock frequency via kvmclock, so a proper wall-clock time source needs to be added for KVM/Firecracker.

### DIFF
--- a/.github/workflows/test_x86_fc.yml
+++ b/.github/workflows/test_x86_fc.yml
@@ -1,0 +1,45 @@
+name: "Test x86-64 (Firecracker)"
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  boot-test:
+    runs-on: ubuntu-latest
+    container:
+      image: asterinas/asterinas:0.18.0-20260702
+      options: -v /dev:/dev --privileged
+    env:
+      FC_SERIAL_LOG: /var/log/fc-serial.log
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build kernel
+        run: make kernel BOOT_METHOD=direct-elf
+
+      - name: Run boot test with Firecracker
+        shell: bash
+        run: |
+          set -euo pipefail
+          trap 'kill $FC_PID 2>/dev/null || true; wait $FC_PID 2>/dev/null || true' EXIT
+
+          firecracker $(./tools/firecracker_args.sh boot) > "$FC_SERIAL_LOG" 2>&1 &
+          FC_PID=$!
+
+          TIMEOUT=60
+          for i in $(seq 1 $TIMEOUT); do
+              if grep -q "^Entered userspace" "$FC_SERIAL_LOG" 2>/dev/null; then
+                  echo "Firecracker boot test passed"
+                  exit 0
+              fi
+              sleep 1
+          done
+
+          echo "Firecracker boot test failed (timeout)"
+          echo "--- Firecracker serial log ---"
+          cat "$FC_SERIAL_LOG"
+          exit 1
--- a/ostd/src/arch/x86/irq/chip/mod.rs
+++ b/ostd/src/arch/x86/irq/chip/mod.rs
@@ -4,10 +4,13 @@ use alloc::{boxed::Box, vec::Vec};
 use core::{
     fmt,
     ops::{Deref, DerefMut},
+    pin::Pin,
 };
 
+use acpi::madt::{Madt, MadtEntry};
 use ioapic::IoApic;
 use spin::Once;
+use x86_64::instructions::port::{PortGeneric, ReadOnlyAccess, WriteOnlyAccess};
 
 use crate::{
     Error, Result, arch::kernel::acpi::get_acpi_tables, info, io::IoMemAllocatorBuilder,
@@ -160,9 +163,53 @@ impl Drop for MappedIrqLine {
 /// The [`IrqChip`] singleton.
 pub static IRQ_CHIP: Once<IrqChip> = Once::new();
 
-pub(in crate::arch) fn init(io_mem_builder: &IoMemAllocatorBuilder) {
-    use acpi::madt::{Madt, MadtEntry};
+fn should_init_and_disable_pic(madt_table: Pin<&Madt>) -> bool {
+    // "A one indicates that the system also has a PC-AT-compatible dual-8259 setup. The 8259
+    // vectors must be disabled (that is, masked) when enabling the ACPI APIC operation."
+    const PCAT_COMPAT: u32 = 1;
+    if madt_table.flags & PCAT_COMPAT != 0 {
+        return true;
+    }
 
+    is_legacy_pic_present()
+}
+
+/// Detects whether a legacy 8259A PIC is present via I/O port probing.
+///
+/// If the I/O ports return the written mask value, we assume the PIC is present.
+/// This serves as a fallback when MADT does not declare `PCAT_COMPAT` but the hardware
+/// (e.g., in some Hypervisors like Firecracker) still emulates it.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/kernel/i8259.c#L306>
+fn is_legacy_pic_present() -> bool {
+    // Mask all except cascade (IRQ 2) on master IMR.
+    const PROBE_VAL: u8 = 0xFB;
+
+    unsafe {
+        // SAFETY: We are in early boot and writing to the legacy PIC ports (0x21, 0xA1) is safe
+        // as long as we verify they respond correctly.
+
+        // Mask slave PIC first to avoid interrupts during probe
+        let mut slave_port = PortGeneric::<u8, WriteOnlyAccess>::new(0xA1);
+        slave_port.write(0xFF);
+
+        // Mask master PIC except cascade (IRQ 2)
+        let mut master_write_port = PortGeneric::<u8, WriteOnlyAccess>::new(0x21);
+        master_write_port.write(PROBE_VAL);
+
+        // Read back from master PIC. If the port exists, it returns what we wrote.
+        // If it's a memory hole (no PIC), it usually returns 0xFF or 0x00.
+        let mut master_read_port = PortGeneric::<u8, ReadOnlyAccess>::new(0x21);
+        let read_val = master_read_port.read();
+
+        if read_val == PROBE_VAL {
+            return true;
+        }
+    }
+    false
+}
+
+pub(in crate::arch) fn init(io_mem_builder: &IoMemAllocatorBuilder) {
     // If there are no ACPI tables, or the ACPI tables do not provide us with information about
     // the I/O APIC, we may need to find another way to determine the I/O APIC address
     // correctly and reliably (e.g., by parsing the MultiProcessor Specification, which has
@@ -170,10 +217,7 @@ pub(in crate::arch) fn init(io_mem_builder: &IoMemAllocatorBuilder) {
     let acpi_tables = get_acpi_tables().unwrap();
     let madt_table = acpi_tables.find_table::<Madt>().unwrap();
 
-    // "A one indicates that the system also has a PC-AT-compatible dual-8259 setup. The 8259
-    // vectors must be disabled (that is, masked) when enabling the ACPI APIC operation"
-    const PCAT_COMPAT: u32 = 1;
-    if madt_table.get().flags & PCAT_COMPAT != 0 {
+    if should_init_and_disable_pic(madt_table.get()) {
         pic::init_and_disable();
     }
 
--- a/ostd/src/arch/x86/kernel/kvm_clock.rs
+++ b/ostd/src/arch/x86/kernel/kvm_clock.rs
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use core::sync::atomic::{Ordering, compiler_fence};
+
+use spin::Once;
+use x86::msr;
+
+use crate::mm::{Frame, FrameAllocOptions, HasPaddr, paddr_to_vaddr};
+
+// KVM CPUID leaves, signature, feature bits, and MSR values.
+// Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/include/uapi/asm/kvm_para.h>.
+const KVM_CPUID_SIGNATURE: u32 = 0x4000_0000;
+const KVM_CPUID_FEATURES: u32 = 0x4000_0001;
+const KVM_SIGNATURE: &[u8; 12] = b"KVMKVMKVM\0\0\0";
+const KVM_FEATURE_CLOCKSOURCE2: u32 = 3;
+const MSR_KVM_SYSTEM_TIME_NEW: u32 = 0x4b56_4d01;
+const KVM_MSR_ENABLED: u64 = 1;
+
+static PVCLOCK_FRAME: Once<Option<Frame<()>>> = Once::new();
+
+#[repr(C, align(4))]
+struct PvclockVcpuTimeInfo {
+    version: u32,
+    _pad0: u32,
+    _tsc_timestamp: u64,
+    _system_time: u64,
+    tsc_to_system_mul: u32,
+    tsc_shift: i8,
+    _flags: u8,
+    _pad: [u8; 2],
+}
+
+struct PvclockTimeSnapshot {
+    tsc_to_system_mul: u32,
+    tsc_shift: i8,
+}
+
+/// Determines the TSC frequency from KVM's paravirtual clock.
+pub(super) fn determine_tsc_freq() -> Option<u64> {
+    if !has_kvm_clocksource2() {
+        return None;
+    }
+
+    let pvclock_info = enable_system_time_msr()?;
+    let time_info = read_stable_time_info(pvclock_info)?;
+    freq_from_time_info(&time_info)
+}
+
+fn has_kvm_clocksource2() -> bool {
+    let Some(signature_leaf) = crate::arch::cpu::cpuid::cpuid(KVM_CPUID_SIGNATURE, 0) else {
+        return false;
+    };
+
+    let mut signature = [0u8; 12];
+    signature[0..4].copy_from_slice(&signature_leaf.ebx.to_le_bytes());
+    signature[4..8].copy_from_slice(&signature_leaf.ecx.to_le_bytes());
+    signature[8..12].copy_from_slice(&signature_leaf.edx.to_le_bytes());
+    if &signature != KVM_SIGNATURE {
+        return false;
+    }
+
+    let Some(features) = crate::arch::cpu::cpuid::cpuid(KVM_CPUID_FEATURES, 0) else {
+        return false;
+    };
+    features.eax & (1 << KVM_FEATURE_CLOCKSOURCE2) != 0
+}
+
+fn enable_system_time_msr() -> Option<*const PvclockVcpuTimeInfo> {
+    let frame = PVCLOCK_FRAME.call_once(|| FrameAllocOptions::new().alloc_frame().ok());
+    let frame = frame.as_ref()?;
+    let paddr = frame.paddr();
+
+    // SAFETY: `paddr` is the physical address of a live, zeroed page retained by
+    // `PVCLOCK_FRAME`. KVM owns updates to the `PvclockVcpuTimeInfo` fields after
+    // this MSR is enabled.
+    unsafe {
+        msr::wrmsr(MSR_KVM_SYSTEM_TIME_NEW, paddr as u64 | KVM_MSR_ENABLED);
+        Some(paddr_to_vaddr(paddr) as *const PvclockVcpuTimeInfo)
+    }
+}
+
+fn read_stable_time_info(pvclock_info: *const PvclockVcpuTimeInfo) -> Option<PvclockTimeSnapshot> {
+    const MAX_RETRIES: usize = 1_000_000;
+
+    for _ in 0..MAX_RETRIES {
+        // KVM marks an in-progress pvclock update with an odd `version`.
+        // Accept fields only when `version` is even and unchanged across the read.
+        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/Documentation/virt/kvm/x86/msr.rst#L87-L90>.
+        let version_before = volatile_read_version(pvclock_info);
+        if version_before & 1 != 0 {
+            core::hint::spin_loop();
+            continue;
+        }
+
+        compiler_fence(Ordering::Acquire);
+        let tsc_to_system_mul = volatile_read_tsc_to_system_mul(pvclock_info);
+        let tsc_shift = volatile_read_tsc_shift(pvclock_info);
+        compiler_fence(Ordering::Acquire);
+
+        let version_after = volatile_read_version(pvclock_info);
+        if version_before == version_after && tsc_to_system_mul != 0 {
+            return Some(PvclockTimeSnapshot {
+                tsc_to_system_mul,
+                tsc_shift,
+            });
+        }
+
+        core::hint::spin_loop();
+    }
+
+    None
+}
+
+fn volatile_read_version(pvclock_info: *const PvclockVcpuTimeInfo) -> u32 {
+    // SAFETY: `pvclock_info` points to a live KVM pvclock page.
+    unsafe { core::ptr::addr_of!((*pvclock_info).version).read_volatile() }
+}
+
+fn volatile_read_tsc_to_system_mul(pvclock_info: *const PvclockVcpuTimeInfo) -> u32 {
+    // SAFETY: `pvclock_info` points to a live KVM pvclock page.
+    unsafe { core::ptr::addr_of!((*pvclock_info).tsc_to_system_mul).read_volatile() }
+}
+
+fn volatile_read_tsc_shift(pvclock_info: *const PvclockVcpuTimeInfo) -> i8 {
+    // SAFETY: `pvclock_info` points to a live KVM pvclock page.
+    unsafe { core::ptr::addr_of!((*pvclock_info).tsc_shift).read_volatile() }
+}
+
+fn freq_from_time_info(time_info: &PvclockTimeSnapshot) -> Option<u64> {
+    let mut numerator = 1_000_000_000u128.checked_shl(32)?;
+    let mut denominator = time_info.tsc_to_system_mul as u128;
+
+    let shift = i32::from(time_info.tsc_shift);
+    // Reverse KVM's TSC-to-ns formula. A positive shift scales the multiplier,
+    // while a negative shift scales the input cycles before multiplication.
+    // Reference: <https://elixir.bootlin.com/linux/v7.0/source/Documentation/virt/kvm/x86/msr.rst#L106-L117>.
+    if shift >= 0 {
+        denominator = denominator.checked_shl(shift as u32)?;
+    } else {
+        numerator = numerator.checked_shl((-shift) as u32)?;
+    }
+
+    let freq = numerator.checked_div(denominator)?;
+    u64::try_from(freq).ok().filter(|freq| *freq != 0)
+}
--- a/ostd/src/arch/x86/kernel/tsc.rs
+++ b/ostd/src/arch/x86/kernel/tsc.rs
@@ -16,9 +16,12 @@ use crate::{
 pub(in crate::arch) static TSC_FREQ: AtomicU64 = AtomicU64::new(0);
 
 pub fn init_tsc_freq() {
+    use super::kvm_clock::determine_tsc_freq as determine_tsc_freq_via_kvmclock;
     use crate::arch::cpu::cpuid::query_tsc_freq as determine_tsc_freq_via_cpuid;
 
-    let tsc_freq = determine_tsc_freq_via_cpuid().unwrap_or_else(determine_tsc_freq_via_pit);
+    let tsc_freq = determine_tsc_freq_via_kvmclock()
+        .or_else(determine_tsc_freq_via_cpuid)
+        .unwrap_or_else(determine_tsc_freq_via_pit);
     TSC_FREQ.store(tsc_freq, Ordering::Relaxed);
     info!("TSC frequency: {:?} Hz", tsc_freq);
 }
--- a/tools/firecracker_args.sh
+++ b/tools/firecracker_args.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: MPL-2.0
+
+# This script generates Firecracker command-line arguments.
+# Usage: `firecracker_args.sh [scheme]`
+#  - scheme: "boot" (default) — boot test with a minimal initrd.
+# Other arguments are configured via environmental variables:
+#  - KERNEL_PATH: path to the kernel ELF binary;
+#  - INITRD_PATH: path to the initramfs cpio archive;
+#  - BOOT_ARGS: kernel command line arguments;
+#  - VCPU: number of vCPUs (default: 2);
+#  - MEM: memory size in MiB (default: 1024);
+#  - FC_CONFIG_FILE: path to the generated Firecracker VM config JSON.
+#    Defaults to /tmp/fc-vm.json.
+
+SCHEME=${1:-"boot"}
+
+KERNEL_PATH=${KERNEL_PATH:-"./target/osdk/aster-kernel/aster-kernel-osdk-bin.elf"}
+INITRD_PATH=${INITRD_PATH:-"./test/initramfs/build/initramfs.cpio.gz"}
+BOOT_ARGS=${BOOT_ARGS:-"console=ttyS0 earlycon loglevel=error init=/bin/echo Entered userspace"}
+VCPU=${VCPU:-2}
+MEM=${MEM:-1024}
+FC_CONFIG_FILE=${FC_CONFIG_FILE:-/tmp/fc-vm.json}
+
+# Generate the VM configuration JSON as a side effect.
+cat > "$FC_CONFIG_FILE" << EOF
+{
+  "boot-source": {
+    "kernel_image_path": "$KERNEL_PATH",
+    "initrd_path": "$INITRD_PATH",
+    "boot_args": "$BOOT_ARGS"
+  },
+  "drives": [],
+  "machine-config": {
+    "vcpu_count": $VCPU,
+    "mem_size_mib": $MEM
+  }
+}
+EOF
+
+# Output Firecracker CLI arguments.
+FC_ARGS="--no-api --config-file $FC_CONFIG_FILE"
+
+echo $FC_ARGS
--- a/Makefile
+++ b/Makefile
@@ -156,11 +156,11 @@ CARGO_OSDK_COMMON_ARGS += --scheme tdx
 endif
 
 ifeq ($(BOOT_PROTOCOL), multiboot)
-BOOT_METHOD = qemu-direct
+BOOT_METHOD = direct-elf
 endif
 
 ifeq ($(SCHEME), microvm)
-BOOT_METHOD = qemu-direct
+BOOT_METHOD = direct-elf
 endif
 
 ifeq ($(SCHEME), "")
--- a/OSDK.toml
+++ b/OSDK.toml
@@ -31,7 +31,7 @@ args = "$(./tools/qemu_args.sh test)"
 # Other Schemes
 
 [scheme."microvm"]
-boot.method = "qemu-direct"
+boot.method = "direct-elf"
 build.strip_elf = true
 qemu.args = "$(./tools/qemu_args.sh microvm)"
 
@@ -48,14 +48,14 @@ qemu.args = "$(./tools/qemu_args.sh tdx)"
 
 [scheme."riscv"]
 supported_archs = ["riscv64"]
-boot.method = "qemu-direct"
+boot.method = "direct-elf"
 build.strip_elf = false
 qemu.args = "$(./tools/qemu_args.sh riscv)"
 
 [scheme."sifive_u"]
 supported_archs = ["riscv64"]
 build.features = ["riscv_sv39_mode"]
-boot.method = "qemu-direct"
+boot.method = "direct-elf"
 build.strip_elf = false
 qemu.with_monitor = true
 
@@ -72,7 +72,7 @@ qemu.args = """\
 """
 
 [scheme."loongarch"]
-boot.method = "qemu-direct"
+boot.method = "direct-elf"
 build.strip_elf = false
 
 qemu.args = """\
--- a/book/src/kernel/vm-based-containers/coco.md
+++ b/book/src/kernel/vm-based-containers/coco.md
@@ -237,16 +237,16 @@ Build the kernel and verify the output images:
 
 ```bash
 # Build the regular guest kernel
-cd /root/asterinas && make kernel BOOT_METHOD=qemu-direct
+cd /root/asterinas && make kernel BOOT_METHOD=direct-elf
 
 # Verify the regular build output
-ls /root/asterinas/target/osdk/aster-kernel-osdk-bin.qemu_elf
+ls /root/asterinas/target/osdk/aster-kernel-osdk-bin.elf
 
 # Build the TDX guest kernel (only needed for the CVM path)
-cd /root/asterinas && make kernel BOOT_METHOD=qemu-direct INTEL_TDX=1
+cd /root/asterinas && make kernel BOOT_METHOD=direct-elf INTEL_TDX=1
 
 # Verify the TDX build output
-# (Note that the TDX kernel uses a bzImage format without the `.qemu_elf` extension.)
+# (Note that the TDX kernel uses a bzImage format without the `.elf` extension.)
 ls /root/asterinas/target/osdk/aster-kernel-osdk-bin
 ```
 
@@ -260,7 +260,7 @@ and set `kernel` under `[hypervisor.qemu]`:
 
 ```toml
 [hypervisor.qemu]
-kernel = "/root/asterinas/target/osdk/aster-kernel-osdk-bin.qemu_elf"
+kernel = "/root/asterinas/target/osdk/aster-kernel-osdk-bin.elf"
 ```
 
 For the TDX runtime, edit:
--- a/book/src/kernel/vm-based-containers/kata.md
+++ b/book/src/kernel/vm-based-containers/kata.md
@@ -199,14 +199,14 @@ If you are using the image's default kernel,
 look for:
 
 ```text
-/opt/kata/share/kata-containers/aster-kernel-osdk-bin.qemu_elf
+/opt/kata/share/kata-containers/aster-kernel-osdk-bin.elf
 ```
 
 If you configured a locally built kernel in Step 4,
 look for:
 
 ```text
-/root/asterinas/target/osdk/aster-kernel-osdk-bin.qemu_elf
+/root/asterinas/target/osdk/aster-kernel-osdk-bin.elf
 ```
 
 This is the most direct check that Kata booted an Asterinas guest kernel.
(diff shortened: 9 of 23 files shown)


StevenJiang1110:
Just some initial comments after a quick look — I haven't had time to review the correctness in detail yet.

However, directly extending OSDK with a new boot method called `LinuxDirect64` doesn't feel quite right. For one thing, this `BootMethod`, as implemented, does not support QEMU — yet that limitation isn't reflected in the name at all. A more fitting name might be `FirecrackerDirect`, as a counterpart to our existing `QemuDirect`. 

Before landing this boot method, I'd also suggest taking care of two things first:

1. Install Firecracker in our Docker image so that we can test it in CI. At a minimum, a basic boot test should be added once this PR is complete.
2. Refactor OSDK to support multiple VMMs. Currently, many parts of OSDK — from configuration files to field naming — are hardcoded to QEMU. Before adding Firecracker support, it would be worth refactoring OSDK to properly accommodate multiple VMMs. If the changes are small, they can be included as a separate commit within this PR; otherwise, a dedicated PR would be more appropriate.

> Patches the ELF header e_entry field to 0x08001200, the fixed
__linux64_boot entry point used by the x86 Linux 64-bit boot protocol path.

I'm also not quite following this part. What does it mean to "patch the ELF header field"? Is the Linux 64-bit boot protocol a new boot protocol? If so, why is this being introduced as a new `BOOT_METHOD` rather than a new `BOOT_PROTOCOL`? And which VMMs support this protocol — is Firecracker the only one? I think this deserves some clarification, as the answer would have a meaningful impact on the design decisions here.

hburaylee:
> Just some initial comments after a quick look — I haven't had time to review the correctness in detail yet.
> 
> However, directly extending OSDK with a new boot method called `LinuxDirect64` doesn't feel quite right. For one thing, this `BootMethod`, as implemented, does not support QEMU — yet that limitation isn't reflected in the name at all. A more fitting name might be `FirecrackerDirect`, as a counterpart to our existing `QemuDirect`.
> 
> Before landing this boot method, I'd also suggest taking care of two things first:
> 
> 1. Install Firecracker in our Docker image so that we can test it in CI. At a minimum, a basic boot test should be added once this PR is complete.
> 2. Refactor OSDK to support multiple VMMs. Currently, many parts of OSDK — from configuration files to field naming — are hardcoded to QEMU. Before adding Firecracker support, it would be worth refactoring OSDK to properly accommodate multiple VMMs. If the changes are small, they can be included as a separate commit within this PR; otherwise, a dedicated PR would be more appropriate.
> 
> > Patches the ELF header e_entry field to 0x08001200, the fixed
> > __linux64_boot entry point used by the x86 Linux 64-bit boot protocol path.
> 
> I'm also not quite following this part. What does it mean to "patch the ELF header field"? Is the Linux 64-bit boot protocol a new boot protocol? If so, why is this being introduced as a new `BOOT_METHOD` rather than a new `BOOT_PROTOCOL`? And which VMMs support this protocol — is Firecracker the only one? I think this deserves some clarification, as the answer would have a meaningful impact on the design decisions here.


I have updated the PR. Here is the current status:

**Naming:** Changed to direct-elf (matching the old qemu-direct), not LinuxDirect64 anymore. This name means "a directly-loadable ELF artifact", and it is not tied to one VMM. Other PVH-capable VMMs can use the same ELF later, so I did not use the name FirecrackerDirect.
  
**CI test:** Firecracker is now installed in the Docker image (see #3623), and a GitHub Actions workflow was added to run a basic boot test.

**About the "patch e_entry" part:** That was the old design, and it is not used anymore. Now the PR uses the PVH boot protocol instead, through a PT_NOTE ELF note. PVH is a common protocol, and it is supported by several VMMs, including Firecracker and Cloud Hypervisor.

  [cloud-hypervisor: The x86-64 kernel must be built with PVH support or be a bzImage](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/README.md)
  [firecracker: PVH boot mode](https://github.com/firecracker-microvm/firecracker/blob/main/docs/pvh.md)
  [qemu: Xen PVH machine](https://www.qemu.org/docs/master/system/i386/xenpvh.html)

hburaylee:
The current commits differ from the earlier discussion in this PR.

- The **PVH boot protocol** has already been moved out and is implemented in **#3765**. This PR now **depends on** that work rather than containing it.

- The old design (patching `e_entry`, `LinuxDirect64` / direct-ELF boot method naming, etc.) is gone. We simply consume the standard PVH + `vmm-direct` path provided by [#3765](https://github.com/asterinas/asterinas/pull/3765).

- What remains in **this** PR are the Firecracker-specific adaptations only:

  (1) Prefer KVM pvclock for TSC frequency (Firecracker only exposes KVM clock), it‘s the most accurate and cheapest TSC frequency source if KVM exist.
  (2) I/O-port probing for the legacy PIC (Firecracker’s ACPI is limited).
  (3) CI boot test workflow + helper script + documentation.
  
@junyang-zh  Could you please have a look ? Thanks!

tools/firecracker_args.sh
```diff
@@ -0,0 +1,53 @@
...
+# The "ci-boot" scheme appends a boot marker for the CI boot test.
+if [ "$SCHEME" = "boot" ]; then
+    BOOT_ARGS="$BOOT_ARGS"
+elif [ "$SCHEME" = "ci-boot" ]; then
+    BOOT_ARGS="$BOOT_ARGS rdinit=/bin/echo Entered userspace"
+fi
+
+# Generate the VM configuration JSON as a side effect.
+cat > "$FC_CONFIG_FILE" << EOF
+{
```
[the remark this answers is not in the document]
hburaylee: I used the `EOF` syntax here. With proper escaping, the issue can be avoided. There’s no need to introduce `jq`. These variables generally don’t contain complex special characters.

ostd/src/arch/x86/kernel/tsc.rs
```diff
@@ -16,9 +16,12 @@ use crate::{
 pub(in crate::arch) static TSC_FREQ: AtomicU64 = AtomicU64::new(0);
 
 pub fn init_tsc_freq() {
+    use super::kvm_clock::determine_tsc_freq as determine_tsc_freq_via_kvmclock;
```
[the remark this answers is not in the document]
hburaylee: The original code has used `use ... as` to directly import the free function. This modification is to be consistent with the existing writing method.

https://github.com/asterinas/asterinas/blob/d77e70d48c15deaa0debe700283c93ed2d024736/ostd/src/arch/x86/kernel/tsc.rs#L18-L21

book/src/kernel/firecracker.md
```diff
@@ -0,0 +1,63 @@
+# Running Asterinas on Firecracker
+
+Asterinas can run as a guest kernel inside a
```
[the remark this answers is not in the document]
hburaylee:
```md
Asterinas can run as a guest kernel inside a
[Firecracker](https://github.com/firecracker-microvm/firecracker) microVM
through the PVH boot protocol.
This page shows how to build the kernel and boot it with Firecracker.
```
However, `[Firecracker](...)` is a complete Markdown link syntax, and the URL is also very long, so it should not be split in the middle.

tools/firecracker_args.sh
```diff
@@ -0,0 +1,53 @@
...
+FC_CONFIG_FILE=${FC_CONFIG_FILE:-/tmp/fc-vm.json}
+
+# The "ci-boot" scheme appends a boot marker for the CI boot test.
+if [ "$SCHEME" = "boot" ]; then
+    BOOT_ARGS="$BOOT_ARGS"
+elif [ "$SCHEME" = "ci-boot" ]; then
+    BOOT_ARGS="$BOOT_ARGS rdinit=/bin/echo Entered userspace"
+fi
+
+# Generate the VM configuration JSON as a side effect.
```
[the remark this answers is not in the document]
hburaylee: Fixed. Added `set -euo pipefail` near the top of `tools/firecracker_args.sh`.

ostd/src/arch/x86/irq/chip/pic.rs
```diff
@@ -24,6 +24,24 @@ pub fn init_and_disable() {
     set_mask(0xff, 0xff);
 }
 
+/// Detects whether a legacy 8259A PIC is present via I/O port probing.
+///
```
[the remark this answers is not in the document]
hburaylee: Rejected. `is_legacy_present()` itself derives its result via I/O port probing, it requires writing to the IMR and then reading it back for comparison.
This inevitably alters the IMR state, a fact already conveyed by the function name and the "via I/O port probing" note in the documentation.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,166 @@
...
+        let frame = PVCLOCK_FRAME.call_once(|| FrameAllocOptions::new().alloc_frame().ok());
+        let frame = frame.as_ref()?;
+        let paddr = frame.paddr();
+
+        // SAFETY: `paddr` is the physical address of a live, zeroed page retained by
+        // `PVCLOCK_FRAME`. KVM owns updates to the `PvclockVcpuTimeInfo` fields after
+        // this MSR is enabled.
+        unsafe {
+            msr::wrmsr(MSR_KVM_SYSTEM_TIME_NEW, paddr as u64 | KVM_MSR_ENABLE_BIT);
+        }
```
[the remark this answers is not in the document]
hburaylee: Fixed. To allow safe's `PvclockPage::setup()` to establish all necessary conditions itself, the feature check `has_kvm_clocksource2()` has been moved into `PvclockPage::setup()`.

hburaylee:
The following force push is to resolve the conflicts.
```rust
<<<<<<< HEAD                                                                       
pub(in crate::arch) fn init_tsc_freq() {                                           
=======                                                                            
pub fn init_tsc_freq() {                                                           
    use super::kvm_clock::determine_tsc_freq as determine_tsc_freq_via_kvmclock;
>>>>>>> efb97a51d (Add KVM pvclock TSC frequency detection) 
```

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,165 @@
...
+fn has_kvm_clocksource2() -> bool {
+    let Some(signature_leaf) = cpuid::cpuid(KVM_CPUID_SIGNATURE, 0) else {
+        return false;
+    };
+
+    let mut signature = [0u8; 12];
+    signature[0..4].copy_from_slice(&signature_leaf.ebx.to_ne_bytes());
+    signature[4..8].copy_from_slice(&signature_leaf.ecx.to_ne_bytes());
+    signature[8..12].copy_from_slice(&signature_leaf.edx.to_ne_bytes());
+    if &signature != KVM_SIGNATURE {
```
lrh2000: Find a way to avoid repeating the code with:
https://github.com/asterinas/asterinas/blob/ac790aa89b2f39e700d263882e1a1301e0a77a25/ostd/src/arch/x86/cpu/cpuid.rs#L107-L120

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,165 @@
...
+        return false;
+    }
+
+    let Some(features) = cpuid::cpuid(KVM_CPUID_FEATURES, 0) else {
+        return false;
+    };
+    features.eax & (1 << KVM_FEATURE_CLOCKSOURCE2) != 0
```
lrh2000: I think this should be a `cpuid::query_xxx` function? See the existing examples in the `cpuid` module.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,165 @@
...
+    tsc_to_system_mul: u32,
+    tsc_shift: i8,
+}
+
+impl PvclockTimeSnapshot {
+    /// Returns the TSC frequency in Hz derived from this snapshot.
+    fn tsc_freq_hz(&self) -> Option<u64> {
+        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/kernel/pvclock.c#L27>.
+
+        // 10^6 << 32 / tsc_to_system_mul gives the TSC frequency in kHz.
```
lrh2000:
```suggestion
        // This expression gives the TSC frequency in kHz.
```

We don't need to repeat the code?

In addition, I do not think the comment is correct: `the TSC frequency in kHz` should be `tsc_khz`. What is the difference between `tsc_khz` and `base_khz`?

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,165 @@
...
+// KVM CPUID leaves, signature, feature bits, and MSR values.
+// Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/include/uapi/asm/kvm_para.h>.
+const KVM_CPUID_SIGNATURE: u32 = 0x4000_0000;
+const KVM_CPUID_FEATURES: u32 = 0x4000_0001;
+const KVM_SIGNATURE: &[u8; 12] = b"KVMKVMKVM\0\0\0";
+const KVM_FEATURE_CLOCKSOURCE2: u32 = 3;
+const MSR_KVM_SYSTEM_TIME_NEW: u32 = 0x4b56_4d01;
+const KVM_MSR_ENABLE_BIT: u64 = 1;
+
+static PVCLOCK_FRAME: Once<Option<Frame<()>>> = Once::new();
```
lrh2000: Please clarify whether this frame contains sensitive data that OSTD users should not be able to access.

 - If you choose to user `Frame<()>`, then:
   - OSTD users can get this frame via [`Frame::from_in_use`](https://github.com/asterinas/asterinas/blob/02c530c1c2da95e29b5bdf93660df88a0aee31b1/ostd/src/mm/frame/mod.rs#L131).
   - OSTD users can _safely_ read from and write to this frame via `VmReader` and `VmWriter`.
   - This PR should also use `VmReader::read_once` and `VmWriter::write_once` to access the KVM time information.
 - Otherwise, you can make the frame typed:
https://github.com/asterinas/asterinas/blob/ac790aa89b2f39e700d263882e1a1301e0a77a25/ostd/src/mm/dma/util.rs#L24-L36

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,165 @@
...
+            let version_before = self.read_version();
+            if version_before & 1 != 0 {
+                core::hint::spin_loop();
+                continue;
+            }
+
+            compiler_fence(Ordering::Acquire);
+            let tsc_to_system_mul = self.read_tsc_to_system_mul();
+            let tsc_shift = self.read_tsc_shift();
+            compiler_fence(Ordering::Acquire);
```
lrh2000, line 104: This needs to be an atomic fence.

How to synchronize with the Linux code is an open problem. Note that:
 1. If we use `read_volatile`, then [atomic fences have nothing to do with it](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#mandatory-atomic).
 2. If we use atomic reads, then the code is UB because [the Linux code does not use atomic operations](https://elixir.bootlin.com/linux/v7.3-rc3/source/arch/x86/kvm/x86.c#L1745-L1755) and [mixing atomic and non-atomic operations are considered UB](https://doc.rust-lang.org/std/sync/atomic/index.html#memory-model-for-atomic-accesses).

So it's difficult to decide which is the "more" correct one.

In any case, we need documentation explaining how synchronization works here. Basically, it is as follows:
 - The first read should act as an acquire operation to ensure that we see the up-to-date data.
 - The data access should also act as an acquire operation to ensure that we see the up-to-date sequence in the second version read. If there are new changes, we'll reject the data because it changed in the meantime.

ostd/src/arch/x86/irq/chip/mod.rs
```diff
@@ -160,20 +162,26 @@ impl Drop for MappedIrqLine {
 /// The [`IrqChip`] singleton.
 pub static IRQ_CHIP: Once<IrqChip> = Once::new();
 
-pub(in crate::arch) fn init(io_mem_builder: &IoMemAllocatorBuilder) {
-    use acpi::madt::{Madt, MadtEntry};
+fn should_init_and_disable_pic(madt_table: Pin<&Madt>) -> bool {
```
lrh2000: Follow `top-down-reading`, define this method after `fn init`.

In addition: I think we should combine the logic that calls the `pic::is_legacy_present` and `pci::init_and_disable` methods into a single method to make all PIC-related logic self-contained.

ostd/src/arch/x86/irq/chip/pic.rs
```diff
@@ -24,6 +24,24 @@ pub(super) fn init_and_disable() {
...
+    const PROBE_VAL: u8 = 0xFB;
+
+    // Mask slave PIC first to avoid interrupts during probe.
+    SLAVE_DATA.write(0xFF);
+
+    // Mask master PIC except cascade (IRQ 2).
+    MASTER_DATA.write(PROBE_VAL);
+
+    // Read back from master PIC. If the port exists, it returns what we wrote.
+    // If it's a memory hole (no PIC), it usually returns 0xFF or 0x00.
```
lrh2000, line 34: Perhaps add final periods to the comments in `set_mask` as well, so that the styles are consistent within the same file.

lrh2000 (commented):
Thanks for your work! The first two commits look mostly good, but I have raised some issues below.

ostd/src/arch/x86/irq/chip/pic.rs
```diff
@@ -24,6 +24,24 @@ pub(super) fn init_and_disable() {
     set_mask(0xff, 0xff);
 }
 
+/// Detects whether a legacy 8259A PIC is present via I/O port probing.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/kernel/i8259.c#L306>
+pub(super) fn is_legacy_present() -> bool {
```
junyang-zh: I think the function name should be `probe_legacy_pic`. `is_legacy_present` implies readonly query. However we probe by writing to I/O ports here.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed. Added `query_hypervisor_signature()` and reused it for both QEMU and KVM detection; added a `TODO` noting the `KVMKVMKVM` signature is not unique to QEMU, so `query_is_running_in_qemu()` may need rewrite in the future.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed. Added `cpuid::query_is_running_under_kvm()` and `cpuid::query_kvm_feature()`.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed. Rewrote the comment to explain the ABI.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed.  Fixed. Made the frame typed with `PvclockPageMeta` so OSTD users cannot read or write it.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed. 
Added required synchronization documentation in `read_time_snapshot()`. Instead of switching to atomic fences, we continue to use `read_volatile()` + `compiler_fence()` because the two alternatives you listed are not sound, the specific reasons are written in the code comments.

ostd/src/arch/x86/irq/chip/mod.rs
hburaylee: Fixed. Moved the helper after `init()` and merged the check + action into `init_and_disable_pic_if_present()`.

ostd/src/arch/x86/irq/chip/pic.rs
hburaylee: Fixed. Added trailing periods to the `set_mask()` comments.

ostd/src/arch/x86/irq/chip/pic.rs
hburaylee: Fixed. Renamed to `probe_legacy_pic()`. The new name is more accurate: the function writes to I/O ports to probe, so the read-only sense of `is_legacy_present` was inaccurate.

ostd/src/arch/x86/cpu/cpuid.rs
```diff
@@ -101,20 +102,45 @@ pub(in crate::arch) fn query_xsave_area_size() -> Option<u32> {
...
     // Reference: <https://wiki.osdev.org/QEMU_fw_cfg#Detecting_QEMU>
+    //
+    // TODO: The "KVMKVMKVM\0\0\0" signature is commonly shared by KVM-based VMMs
+    // (not unique to QEMU). Detecting QEMU via its ACPI tables
+    // (e.g., OEM IDs or the QEMU0002 device) would generally be more accurate.
     matches!(&signature, b"TCGTCGTCGTCG" | b"KVMKVMKVM\0\0\0")
 }
+
+/// Queries if the system is running under KVM.
+pub(in crate::arch) fn query_is_running_under_kvm() -> bool {
```
junyang-zh, line 135: The function name doesn't read like natural English. If we need to keep `query_` prefix, `query_if_running_under_qemu` would be better. However `is_running_under_kvm` is just OK. It is also the case for line 120.

junyang-zh (approved):
LGTM!

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,172 @@
...
+            // Reference: <https://elixir.bootlin.com/linux/v7.0/source/Documentation/virt/kvm/x86/msr.rst#L86-L90>.
+            let version_before = self.read_version();
+            if version_before & 1 != 0 {
+                core::hint::spin_loop();
+                continue;
+            }
+
+            // KVM writes these fields non-atomically. We read them with
+            // `read_volatile`. The fences mark the acquire order, like Linux's
+            // `virt_rmb()` does on x86. Atomic loads would race with KVM's
```
lrh2000, line 96: This is fundamentally incorrect. We cannot compare a fence to Linux barriers because they operate within two different memory models. Any interaction between the two memory models is essentially undefined, so this is a significant open question. We cannot write comments this way.

We should rephrase this as a `FIXME`. I will try later to see what I can write best.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,172 @@
...
+        if !has_kvm_clocksource2() {
+            return None;
+        }
+
+        let frame = PVCLOCK_FRAME.call_once(|| {
+            FrameAllocOptions::new()
+                .alloc_frame_with(PvclockPageMeta)
+                .ok()
+        });
+        let frame = frame.as_ref()?;
```
lrh2000, line 67: I think this is better:
```rust
let frame = if let Some(frame) = FRAME.get() {
    frame
} else {
    let frame = alloc()?;
    FRAME.call_once(|| frame)
};
```

So we do not really need to have `Once<Option<Frame>>`.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,172 @@
...
+        let paddr = frame.paddr();
+
+        // SAFETY: `paddr` is a live page retained by `PVCLOCK_FRAME`, and this
+        // MSR is supported per `has_kvm_clocksource2()` above.
+        unsafe {
+            msr::wrmsr(MSR_KVM_SYSTEM_TIME_NEW, paddr as u64 | KVM_MSR_ENABLE_BIT);
+        }
```
lrh2000:
```suggestion
        unsafe { msr::wrmsr(MSR_KVM_SYSTEM_TIME_NEW, paddr as u64 | KVM_MSR_ENABLE_BIT) };
```

ostd/src/arch/x86/cpu/cpuid.rs
```diff
@@ -101,20 +102,45 @@ pub(in crate::arch) fn query_xsave_area_size() -> Option<u32> {
     cpuid(Leaf::Xstate as u32, 1).map(|res| res.ebx)
 }
 
+/// Returns the 12-byte hypervisor signature.
+fn query_hypervisor_signature() -> Option<[u8; 12]> {
```
lrh2000: Again, follow `top-down-reading`, this should be defined below `query_is_running_under_kvm`,

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,172 @@
...
+            // `read_volatile`. The fences mark the acquire order, like Linux's
+            // `virt_rmb()` does on x86. Atomic loads would race with KVM's
+            // non-atomic stores, so we avoid them.
+            // Reference:
+            // <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/kvm/x86.c#L3294-L3304>
+            // <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/kernel/pvclock.c#L75-L79>
+            compiler_fence(Ordering::Acquire);
+            let tsc_to_system_mul = self.read_tsc_to_system_mul();
+            let tsc_shift = self.read_tsc_shift();
+            compiler_fence(Ordering::Acquire);
```
lrh2000, line 104: As I said before, this should be atomic fences. Do you disagree?

lrh2000 (commented):
Thanks for the updates!

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: I agree that fences and Linux barriers cannot be considered equivalent under the same memory model, the use of "like" instead of "match" is precisely to express this.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: > If we use read_volatile, then [atomic fences have nothing to do with it](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#mandatory-atomic).

Sorry, I didn't understand what you meant. If atomic fences are used with read_volatile, then don't "atomic fences" become meaningless?

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,172 @@
...
+/// [`VmReader`]: crate::mm::VmReader
+/// [`VmWriter`]: crate::mm::VmWriter
+struct PvclockPageMeta;
+impl_frame_meta_for!(PvclockPageMeta);
+
+static PVCLOCK_FRAME: Once<Option<Frame<PvclockPageMeta>>> = Once::new();
+
+/// Guest-visible layout of the KVM pvclock page, mirroring the ABI's
+/// `struct pvclock_vcpu_time_info`. KVM writes this structure directly
+/// into the page we hand it via `MSR_KVM_SYSTEM_TIME_NEW`.
```
lrh2000:
```suggestion
/// Guest-visible layout of the KVM pvclock page.
///
/// This is Linux's `struct pvclock_vcpu_time_info`. KVM writes this structure
/// directly into the page we hand it via `MSR_KVM_SYSTEM_TIME_NEW`.
```

Keep the summary sentence short.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,172 @@
...
+    tsc_to_system_mul: u32,
+    tsc_shift: i8,
+    _flags: u8,
+    _pad: [u8; 2],
+}
+
+/// A handle to the KVM pvclock page shared with the hypervisor.
+struct PvclockPage {
+    info: *const PvclockVcpuTimeInfo,
+}
```
lrh2000: I am also wondering if you like this:
```rust
struct PvclockPage {
    Frame<PvclockPageMeta>
}

impl PvclockPage {
    fn setup() -> Option<&'static Self> {
        if let Some(page) = PVCLOCK_PAGE.get() {
            return Some(page);
        }

        let frame = alloc()?;
        PVCLOCK_PAGE.call_once(|| {
            unsafe { wrmsr() };
            Self(frame)
        })
    }

     fn as_ptr(&self) -> *const PvclockVcpuTimeInfo;
}
```

I think it is more reasonable than the current form.

ostd/src/arch/x86/kernel/kvm_clock.rs
lrh2000: > the use of "like" instead of "match" is precisely to express this

However, I am unsure of how you define `like` and whether you have any authoritative sources claiming that they behave similarly.

Even if there are, I don't think you can argue for correctness based on `like`. We need to take a serious approach when discussing fundamental concepts related to correctness.

---

I propose this version:
```rust
            // FIXME: All synchronization below is based on the assumption that we are performing
            // atomic operations under the same memory model as the KVM side. However:
            // 1. The Linux memory model is incompatible with Rust's.
            // 2. We use non-atomic volatile operations to safely communicate with untrusted
            //    parties. But these operations may not be able to establish synchronization.

            // Synchronize with the (even) version update so that we can see the data written before
            // the version is written.
            atomic::fence(Ordering::Acquire);

            let tsc_to_system_mul = self.read_tsc_to_system_mul();
            let tsc_shift = self.read_tsc_shift();

            // Synchronize with the data update so that we can see the (odd) version update before
            // the data is written (if the data is changed concurrently).
            atomic::fence(Ordering::Acquire);
```

ostd/src/arch/x86/kernel/kvm_clock.rs
lrh2000: Compiler fences have nothing to do with CPU reordering instructions. Without considering CPU out-of-order execution, the code can never be correct.

> > If we use read_volatile, then [atomic fences have nothing to do with it](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#mandatory-atomic).
> 
> Sorry, I didn't understand what you meant. If atomic fences are used with read_volatile, then don't "atomic fences" become meaningless?

That sentence means using `read_volatile` is a wrong choice because it cannot work correctly with `atomic::fence`. However, it does not say that we should not use `atomic::fence`.

We still need to use `atomic::fence` because it is the only way to address CPU out-of-order execution. You must take this into consideration this when communicating between threads (the KVM and the guest are of course two threads). `compiler_fence` makes no sense here because we cannot consider the KVM and the guest to be within one thread.

ostd/src/arch/x86/cpu/cpuid.rs
hburaylee: Agreed. Renamed `query_is_running_under_kvm` to `is_running_under_kvm` and `query_is_running_in_qemu` to `is_running_in_qemu`, as suggested.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Thanks for the patient correction. My mistake,  I overlooked something. `atomic::fence(Ordering::Acquire)` can stop the CPU from reordering, but `compiler_fence(Ordering::Acquire)` can't. That means `compiler_fence` can't guarantee the data is actually correct. 

Fixed. Replaced `compiler_fence` with the atomic `fence`, and removed that comment, the synchronization is now documented by the FIXME.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: I rewrote this following the approach used in the code linked here: https://github.com/asterinas/asterinas/pull/3461#discussion_r4082724778

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Done. Moved the semicolon to after the `unsafe` block.

ostd/src/arch/x86/cpu/cpuid.rs
hburaylee: Sorry, I missed this one. Fixed now.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Thanks for the patient correction. My mistake,  I overlooked something. `atomic::fence(Ordering::Acquire)` can stop the CPU from reordering, but `compiler_fence(Ordering::Acquire)` can't. That means `compiler_fence` can't guarantee the data is actually correct. 

Fixed. Replaced `compiler_fence` with the atomic `fence`, and removed that comment, the synchronization is now documented by the FIXME.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed. `PvclockPage` now owns `Frame<PvclockPageMeta>` instead of a raw pointer, and the page address is derived via `addr()`.

ostd/src/arch/x86/cpu/cpuid.rs
```diff
@@ -101,20 +102,45 @@ pub(in crate::arch) fn query_xsave_area_size() -> Option<u32> {
     cpuid(Leaf::Xstate as u32, 1).map(|res| res.ebx)
 }
 
-/// Queries if the system is running in QEMU.
+/// Returns whether the system is running in QEMU.
 ///
 /// This function uses the CPUID instruction to detect the QEMU hypervisor signature.
-pub(in crate::arch) fn query_is_running_in_qemu() -> bool {
-    let Some(result) = cpuid(Leaf::HypervisorBase as u32, 0) else {
+pub(in crate::arch) fn is_running_in_qemu() -> bool {
```
lrh2000: Please keep the `query_` prefix for methods in this file: https://github.com/asterinas/asterinas/pull/2395#discussion_r2324025899

ostd/src/arch/x86/cpu/cpuid.rs
```diff
@@ -101,20 +102,45 @@ pub(in crate::arch) fn query_xsave_area_size() -> Option<u32> {
...
+/// Returns whether the system is running under KVM.
+pub(in crate::arch) fn is_running_under_kvm() -> bool {
+    let Some(signature) = query_hypervisor_signature() else {
+        return false;
+    };
+
+    matches!(&signature, b"KVMKVMKVM\0\0\0")
+}
+
+/// Returns the 12-byte hypervisor signature.
```
lrh2000:
```suggestion
/// Queries the 12-byte hypervisor signature.
```

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,184 @@
...
+                    MSR_KVM_SYSTEM_TIME_NEW,
+                    frame.paddr() as u64 | KVM_MSR_ENABLE_BIT,
+                )
+            };
+            Self(frame)
+        }))
+    }
+
+    /// Returns the virtual address of the pvclock page.
+    fn addr(&self) -> *const PvclockVcpuTimeInfo {
```
lrh2000: Sorry, I edited my GitHub comment, but it might not have been refreshed on your end. Let's rename this to `as_ptr` because I found that `addr()` conflicts with [`ptr::addr`](https://doc.rust-lang.org/std/primitive.pointer.html#method.addr) (which should return `usize` instead).

---

Also please move this method to be the last method in this `impl` block:
```rust
fn read_time_snapshot();
fn read_version();
fn read_xxx();
fn read_yyy();
fn as_ptr();
```

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,184 @@
...
+
+            // Synchronize with the (even) version update so that we can see the data written before
+            // the version is written.
+            fence(Ordering::Acquire);
+
+            let tsc_to_system_mul = self.read_tsc_to_system_mul();
+            let tsc_shift = self.read_tsc_shift();
+
+            // Synchronize with the data update so that we can see the (odd) version update before
+            // the data is written.
```
lrh2000:
```suggestion
            // the data is written (if the data is changed concurrently).
```

I edited this after posting my comment to emphasize that an update of the odd version happens if the data is changed concurrently. I'm not sure if you missed it or deleted it for some reason.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,184 @@
...
+/// The frame is written by the hypervisor and read by the guest clock code,
+/// so it is kept typed to prevent OSTD users from reading or writing it as
+/// untyped memory via [`VmReader`] and [`VmWriter`].
+///
+/// [`VmReader`]: crate::mm::VmReader
+/// [`VmWriter`]: crate::mm::VmWriter
+struct PvclockPageMeta;
+impl_frame_meta_for!(PvclockPageMeta);
+
+static PVCLOCK_PAGE: Once<PvclockPage> = Once::new();
```
lrh2000: This should be moved to:
```rust
/// A handle to the KVM pvclock page shared with the hypervisor.
struct PvclockPage(Frame<PvclockPageMeta>);

static PVCLOCK_PAGE: Once<PvclockPage> = Once::new();

impl PvclockPage {
```

ostd/src/arch/x86/irq/chip/pic.rs
```diff
@@ -24,26 +24,44 @@ pub(super) fn init_and_disable() {
     set_mask(0xff, 0xff);
 }
 
+/// Probes whether a legacy 8259A PIC is present.
+///
+/// Reference: <https://elixir.bootlin.com/linux/v7.0/source/arch/x86/kernel/i8259.c#L306>
+pub(super) fn probe_legacy_pic() -> bool {
```
lrh2000, line 23: This should be moved above `init_and_disable`? It happens before the initialization process after all.

lrh2000 (commented):
Thanks for the quick updates.

ostd/src/arch/x86/cpu/cpuid.rs
hburaylee: Fixed. 

Kept the `query_` prefix and renamed them to `query_if_running_in_qemu` and `query_if_running_under_kvm` so the predicate reads more naturally.
As suggested: https://github.com/asterinas/asterinas/pull/3461#discussion_r4081289219

ostd/src/arch/x86/cpu/cpuid.rs
hburaylee: Fixed. Changed "Returns "to "Queries".

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Oh, sorry! Perhaps I didn't see the final edited version. 
Fixed. Renamed `addr` to `as_ptr` and moved it to be the last method in the `impl` block.

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Sorry, perhaps I didn't see the final edited version, I should have checked more carefully.
Fixed. Updated the comment to "the data is written (if the data is changed concurrently)."

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Fixed. Moved `PVCLOCK_PAGE` below the `PvclockPage` definition.

ostd/src/arch/x86/irq/chip/pic.rs
hburaylee: Fixed. Moved `probe_legacy_pic` above `init_and_disable`.

ostd/src/arch/x86/kernel/kvm_clock.rs
```diff
@@ -0,0 +1,184 @@
...
+        // SAFETY: `self.as_ptr()` points to a live KVM pvclock page.
+        unsafe { core::ptr::addr_of!((*self.as_ptr()).tsc_to_system_mul).read_volatile() }
+    }
+
+    fn read_tsc_shift(&self) -> i8 {
+        // SAFETY: `self.as_ptr()` points to a live KVM pvclock page.
+        unsafe { core::ptr::addr_of!((*self.as_ptr()).tsc_shift).read_volatile() }
+    }
+
+    /// Returns the virtual address of the pvclock page.
```
lrh2000, line 142: Update the comment:
```suggestion
    /// Returns a pointer that points to the pvclock page.
```

lrh2000 (approved):
**The first two commits** LGTM except a minor comment revision. (I do not plan to check the last two commits.)
 -  Add KVM pvclock TSC frequency detection.
 - Add legacy PIC detection via IO port.

Thanks for your contribution!

ostd/src/arch/x86/kernel/kvm_clock.rs
hburaylee: Done.

.github/workflows/test_x86_fc.yml
```diff
@@ -0,0 +1,45 @@
+name: "Test x86-64 (Firecracker)"
```
tatetian: Please name this file to `test_x86_firecracker.yml` to improved clarity.

.github/workflows/test_x86_fc.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Run boot test with Firecracker
+        shell: bash
+        run: |
+          set -euo pipefail
+          trap 'kill $FC_PID 2>/dev/null || true; wait $FC_PID 2>/dev/null || true' EXIT
+
+          firecracker $(./tools/firecracker_args.sh ci-boot) > "$FC_SERIAL_LOG" 2>&1 &
+          FC_PID=$!
+
+          TIMEOUT=60
```
tatetian: Don't you think 60 seconds is too long?

Issue #3430, opened 2026-06-24
Tracking issue for supporting Asterinas as the host kernel of Kata Containers


volcaxiao:
# Roadmap

This issue tracks the work required to run `containerd` and Kata Containers on top of Asterinas as the host kernel. This is different from using Asterinas as a Kata guest kernel or as a confidential-computing guest kernel: here the Asterinas system is the host that runs `containerd`, `containerd-shim-kata-v2`, QEMU, `virtiofsd`, and the host-side vhost/vsock backend used by the nested Kata VM.

The initial target is a single-container Kata path with QEMU using TCG inside the Asterinas host. The roadmap then extends the same path to image-based containers, KVM-backed QEMU through the Asterinas hypervisor support work, guest networking, common OCI security settings, and finally performance and semantic hardening.

This issue intentionally does not track kvm device initialization or VMM bring-up work. The KVM-backed phase below should connect to the upstream Asterinas hypervisor effort tracked by #3418.

# Current Status

Some of the checked items below have already been split into PRs. As of 2026-07-23, [#3462](https://github.com/asterinas/asterinas/pull/3462), [#3467](https://github.com/asterinas/asterinas/pull/3467), [#3486](https://github.com/asterinas/asterinas/pull/3486), [#3494](https://github.com/asterinas/asterinas/pull/3494), [#3552](https://github.com/asterinas/asterinas/pull/3552), and [#3600](https://github.com/asterinas/asterinas/pull/3600) have been merged. [#3508](https://github.com/asterinas/asterinas/pull/3508) remains under review, while [#3418](https://github.com/asterinas/asterinas/pull/3418) is open and approved. The rest are either local prototypes that still need polishing and splitting, or areas where the community is already working on the upstream implementation.

The current branch validates three user-visible paths and one optional restricted-network reproduction path:

- Minimal Kata `--rootfs` execution through `containerd -> shim-kata-v2 -> runtime-rs -> QEMU -> kata-agent`, including non-zero exit-code propagation.
- Offline OCI image import into containerd, overlayfs snapshot preparation, and Kata execution without `--rootfs`, using a bind-mounted shared rootfs path.
- A local KVM-backed integration path: nested QEMU `microvm` with KVM runs the BusyBox image test and propagates `HI-IMG` plus exit code 7. This uses the local Asterinas hypervisor and runtime-rs MMIO changes and is not yet upstream support.
[...]

PR #3463, opened 2026-06-26
Add namespace plumbing for container runtimes

volcaxiao:
This PR adds namespace identity/plumbing needed by container runtime probes.

## Changes

- Expose `/proc/[pid]/ns/*` entries through procfs/nsfs.
- Add namespace metadata plumbing for `clone`, `unshare`, and `setns` paths.

This is part of #3430, but the PR is scoped to namespace compatibility only.


volcaxiao:
this PR is not ready as an upstreamable namespace implementation.

The current change mostly adds namespace identity/plumbing for runtime probes, but it does not yet define and implement the actual resources that should be isolated by a new namespace. I will close this PR and revisit it only after the namespace scope and isolation semantics are designed and implemented properly.

PR #3464, opened 2026-06-26
Support mount propagation for shared rootfs setup

volcaxiao:
This PR adds mount propagation handling needed by Linux-compatible shared rootfs setup.

## Changes

- Add support for propagation flags such as shared, slave, and private in the mount path.
- Propagate bind mounts in the VFS mount tree where Linux-compatible runtime setup expects it.

Part of #3430.


volcaxiao:
#3430 
Thanks for the update. I will close this PR to avoid duplicating the ongoing mount propagation work.

I may revisit the containerd/Kata shared-rootfs use case later on top of the upstream mount propagation implementation, with focused regression tests if any gaps remain.

PR #3465, opened 2026-06-26
Fix virtio device initialization sequence

hburaylee:
## Summary

This PR updates VirtIO device initialization to follow the sequence required by VirtIO 1.3 section 3.1.1.

This fixes initialization that previously worked on QEMU (more lenient) but failed on Firecracker VMM, which strictly follows the VirtIO specification. The change improves compatibility with strict VMMs.

This PR is also a step towards supporting Asterinas as a guest kernel in Firecracker microVMs, see #3429 .

  - Set `ACKNOWLEDGE` and `DRIVER` in separate steps.
  - Re-read device status after setting `FEATURES_OK`.
  - Mark the device as `FAILED` and stop initializing it if any of the steps go wrong.
  - Move initial receive/event queue notifications until after `DRIVER_OK` is set.

> Reference: https://docs.oasis-open.org/virtio/virtio/v1.3/virtio-v1.3.html#x1-1230001

### DIFF
--- a/kernel/comps/virtio/src/device/console/device.rs
+++ b/kernel/comps/virtio/src/device/console/device.rs
@@ -16,12 +16,12 @@ use super::{DEVICE_NAME, config::VirtioConsoleConfig};
 use crate::{
     device::{VirtioDeviceError, console::config::ConsoleFeatures},
     queue::VirtQueue,
-    transport::{ConfigManager, VirtioTransport},
+    transport::{ConfigManager, DeviceTransport},
 };
 
 pub struct ConsoleDevice {
     config_manager: ConfigManager<VirtioConsoleConfig>,
-    transport: SpinLock<Box<dyn VirtioTransport>>,
+    transport: SpinLock<DeviceTransport>,
     receive_queue: SpinLock<VirtQueue>,
     transmit_queue: SpinLock<VirtQueue>,
     send_buffer: DmaStream,
@@ -85,18 +85,21 @@ impl ConsoleDevice {
         features.bits()
     }
 
-    pub(crate) fn init(mut transport: Box<dyn VirtioTransport>) -> Result<(), VirtioDeviceError> {
-        let config_manager = VirtioConsoleConfig::new_manager(transport.as_ref());
+    pub(crate) fn init(mut device_transport: DeviceTransport) -> Result<(), VirtioDeviceError> {
+        let config_manager = VirtioConsoleConfig::new_manager(device_transport.as_ref());
         debug!("virtio_console_config = {:?}", config_manager.read_config());
 
         const RECV0_QUEUE_INDEX: u16 = 0;
         const TRANSMIT0_QUEUE_INDEX: u16 = 1;
-        let receive_queue =
-            SpinLock::new(VirtQueue::new(RECV0_QUEUE_INDEX, 2, transport.as_mut())?);
+        let receive_queue = SpinLock::new(VirtQueue::new(
+            RECV0_QUEUE_INDEX,
+            2,
+            device_transport.as_mut(),
+        )?);
         let transmit_queue = SpinLock::new(VirtQueue::new(
             TRANSMIT0_QUEUE_INDEX,
             2,
-            transport.as_mut(),
+            device_transport.as_mut(),
         )?);
 
         let send_buffer = DmaStream::alloc(1, false).map_err(VirtioDeviceError::ResourceAlloc)?;
@@ -105,7 +108,7 @@ impl ConsoleDevice {
 
         let device = Arc::new(Self {
             config_manager,
-            transport: SpinLock::new(transport),
+            transport: SpinLock::new(device_transport),
             receive_queue,
             transmit_queue,
             send_buffer,
@@ -131,6 +134,8 @@ impl ConsoleDevice {
         transport.finish_init();
         drop(transport);
 
+        Self::notify_receive_queue(&mut device.receive_queue.disable_irq().lock());
+
         aster_console::register_device(DEVICE_NAME.to_string(), device);
 
         Ok(())
@@ -155,6 +160,7 @@ impl ConsoleDevice {
         drop(callbacks);
 
         self.activate_receive_buffer(&mut receive_queue);
+        Self::notify_receive_queue(&mut receive_queue);
     }
 
     fn activate_receive_buffer(&self, receive_queue: &mut VirtQueue) {
@@ -168,7 +174,9 @@ impl ConsoleDevice {
             // <https://lore.kernel.org/qemu-devel/20240707111940.232549-3-lrh2000@pku.edu.cn/T/#u>.
             .add_output_bufs(&[&Slice::new(&self.receive_buffer, 0..1)])
             .unwrap();
+    }
 
+    fn notify_receive_queue(receive_queue: &mut VirtQueue) {
         if receive_queue.should_notify() {
             receive_queue.notify();
         }
--- a/kernel/comps/virtio/src/device/network/device.rs
+++ b/kernel/comps/virtio/src/device/network/device.rs
@@ -18,7 +18,7 @@ use crate::{
         },
     },
     queue::{self, VirtQueue},
-    transport::{ConfigManager, VirtioTransport},
+    transport::{ConfigManager, DeviceTransport},
 };
 
 pub struct NetworkDevice {
@@ -34,7 +34,7 @@ pub struct NetworkDevice {
     tx_buffers: Vec<Option<TxBuffer>>,
     rx_buffers: SlotVec<RxBuffer>,
     new_rx_buffer: Option<RxBuffer>,
-    transport: Box<dyn VirtioTransport>,
+    transport: DeviceTransport,
     poll_stat: PollStatistics,
 }
 
@@ -70,22 +70,22 @@ impl NetworkDevice {
         network_features.bits()
     }
 
-    pub(crate) fn init(mut transport: Box<dyn VirtioTransport>) -> Result<(), VirtioDeviceError> {
-        let config_manager = VirtioNetConfig::new_manager(transport.as_ref());
+    pub(crate) fn init(mut device_transport: DeviceTransport) -> Result<(), VirtioDeviceError> {
+        let config_manager = VirtioNetConfig::new_manager(device_transport.as_ref());
         let config = config_manager.read_config();
         debug!("virtio_net_config = {:?}", config);
         let mac_addr = config.mac;
         let features = NetworkFeatures::from_bits_truncate(Self::negotiate_features(
-            transport.read_device_features(),
+            device_transport.read_device_features(),
         ));
         debug!("features = {:?}", features);
 
         let caps = init_caps(&features, &config);
 
-        let mut send_queue = VirtQueue::new(QUEUE_SEND, QUEUE_SIZE, transport.as_mut())?;
+        let mut send_queue = VirtQueue::new(QUEUE_SEND, QUEUE_SIZE, device_transport.as_mut())?;
         send_queue.disable_callback();
 
-        let mut recv_queue = VirtQueue::new(QUEUE_RECV, QUEUE_SIZE, transport.as_mut())?;
+        let mut recv_queue = VirtQueue::new(QUEUE_RECV, QUEUE_SIZE, device_transport.as_mut())?;
 
         let tx_buffers = (0..QUEUE_SIZE).map(|_| None).collect();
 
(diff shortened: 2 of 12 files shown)


kernel/comps/virtio/src/lib.rs
```diff
@@ -52,18 +52,27 @@ fn virtio_component_init() -> Result<(), ComponentInitError> {
     device::socket::init();
 
     while let Some(mut transport) = pop_device_transport() {
-        // Reset device
+        // Follow VirtIO 1.3, 3.1.1 "Driver Requirements: Device Initialization"
+        // Reference: <https://docs.oasis-open.org/virtio/virtio/v1.3/virtio-v1.3.html>
```
lrh2000: Make the link jump to the section?
```suggestion
        // Reference: <https://docs.oasis-open.org/virtio/virtio/v1.3/virtio-v1.3.html#x1-1230001>
```

Also add a blank line below since it's the reference for all the operations (instead of only the "Reset the device" part).

kernel/comps/virtio/src/lib.rs
lrh2000: Could you also please check the code here? I don't think it's 100% correct. Audit it according to the specifications and update the code as necessary.

For example, it seems that we need to re-check `FEATURES_OK` and we many need to write `DRIVER_OK`.

Also rewrite the comment `// negotiate features` and `// change to features ok status`. Align them with your newly added comments (following the correct comment style in our codebase).

lrh2000 (commented):
Thanks for fixing the virtio device initialization process.

kernel/comps/virtio/src/lib.rs
hburaylee: Done.

kernel/comps/virtio/src/lib.rs
hburaylee: > Could you also please check the code here? I don't think it's 100% correct. Audit it according to the specifications and update the code as necessary.
> 
> For example, it seems that we need to re-check `FEATURES_OK` and we many need to write `DRIVER_OK`.
> 
> Also rewrite the comment `// negotiate features` and `// change to features ok status`. Align them with your newly added comments (following the correct comment style in our codebase).

Fixed according to the specification: re-read the device status after setting `FEATURES_OK` to verify; if it is not set, mark the device as `FAILED` and skip it. `DRIVER_OK` is set after the device completes specific setup (including virtqueue discovery).

kernel/comps/virtio/src/device/console/device.rs
```diff
@@ -131,6 +131,9 @@ impl ConsoleDevice {
         transport.finish_init();
         drop(transport);
 
+        // Notify the device only after `DRIVER_OK`.
```
lrh2000: Should say `finish_init` instead of `DIRVER_OK`? This is an API contract, which is better than exposing the implementation details of `finish_init`.

In addition, let's be consistent. If you choose to add a comment, ensure it is added to every location. It seems that you sometimes include this comment but sometimes omit it, but it is unclear why.

kernel/comps/virtio/src/device/socket/queue.rs
```diff
@@ -163,17 +163,19 @@ impl RxQueue {
...
-        }
-
         Ok(Self {
             queue,
             buffers,
             pending: None,
         })
     }
 
+    pub(super) fn notify_if_needed(&mut self) {
```
lrh2000: If you add this method, please use this method to replace other similar code.

For example, in this file, there is still code likes `if should_notify { notify() }` in two `recv()` methods. Convert them.

kernel/comps/virtio/src/lib.rs
```diff
@@ -52,30 +52,52 @@ fn virtio_component_init() -> Result<(), ComponentInitError> {
...
+
+            let status = transport.read_device_status();
+            if !status.contains(DeviceStatus::FEATURES_OK) {
+                error!(
+                    "Device rejected negotiated features, device type: {:?}",
+                    device_type
+                );
+                transport
+                    .write_device_status(status | DeviceStatus::FAILED)
+                    .unwrap();
```
lrh2000, line 95: Should we also set `FAILED` in:
```rust
            _ => {
                warn!("Found unimplemented device: {:?}", device_type);
                Ok(())
            }
```
And
```rust
        if res.is_err() {
            error!(
                "Device initialization error: {:?}, device type: {:?}",
                res, device_type
            );
```
?

According to the specifications, it seems necessary because they say that we should set the failed flag if any of the preceding steps fail.

lrh2000 (commented):
Thanks for the revision!

kernel/comps/virtio/src/device/console/device.rs
hburaylee: Thanks for pointing this out. The old comment referencing `DRIVER_OK` has been removed.

kernel/comps/virtio/src/device/socket/queue.rs
hburaylee: A `notify_if_needed()` method has been added to each queue struct (`TxQueue`, `RxQueue`, `EventQueue`).

kernel/comps/virtio/src/lib.rs
hburaylee: Now, for initialization error cases, the `init()` method of each device receives a `TransportGuard`, which internally manages the FAILED state as needed.

It should be noted that `TransportGuard::into_transport()` takes ownership of the transport, so after this call, if subsequent operations (such as `register_xxx_callback`) fail, it is no longer possible to set the FAILED state via the transport. Fortunately, the calls after `into_transport()` (`register_xxx_callback`) do not actually fail in practice (they always return `Ok`).

kernel/comps/virtio/src/device/socket/queue.rs
```diff
@@ -82,6 +82,10 @@ impl TxQueue {
             self.inflight[token as usize] = Some(packet);
         }
 
+        self.notify_if_needed();
+    }
+
+    pub(super) fn notify_if_needed(&mut self) {
```
lrh2000: `TxQueue::notify_if_needed` does not need to be `pub(super)`.

Even as a non-`pub(super)` helper method, it is located at a wrong position. Why should it be defined between `free_processed_tx_buffers` and `try_send`? This is not reasonable.... Follow the [`top-down-reading`](https://github.com/asterinas/asterinas/blob/1192a330ce31302e829f17168120d13500f9d40a/book/src/to-contribute/coding-guidelines/for-maintainability/layout.md?plain=1#L3) rule to move all `notify_if_needed` to the end of the `impl` block, which is more consistent.

kernel/comps/virtio/src/lib.rs
lrh2000: > the `init()` method of each device receives a `TransportGuard`

What's the advantage of doing this instead of setting the FAILED bit in `virtio_component_init`?

Sorry if I'm missing something. Introducing `TransportGuard` is a much more complex solution, so it should provide additional benefits (e.g., it is more correct in certain scenarios). However, I can't find any justification now.

hburaylee: Thanks for your good question! Let me explain why I wrote the `TransportGuard`.

**The main problem is ownership.** When a device's `init()` function runs, it needs to take full ownership of the `transport`. This means that the caller (in virtio_component_init) no longer has it. So the caller cannot set the `FAILED` status bit anymore.

This is exactly what Linux does:
https://elixir.bootlin.com/linux/v7.1.2/source/drivers/virtio/virtio.c#L362
```
static int virtio_dev_probe(struct device *_d)
{
    struct virtio_device *dev = dev_to_virtio(_d);
    // ... 
    return 0;
err:
    virtio_add_status(dev, VIRTIO_CONFIG_S_FAILED);
    return err;
}
```
This works because Linux uses pointer, The caller always retains access.

Here are some other ways i thought about:
- (1) Return the `transport` back on `init()` failure. 
We could use `Result<_, (Box<dyn VirtioTransport>, Error)>` — is a bit complicated and requires a lot of changes. Also, for the `init()` function's responsibility, returning a `transport` as the result feels a bit strange.

- (2) We could revert the code, keep things as they are, and avoid premature optimization. 
Although this would not strictly comply with the Virtio spec, it should be fine in most cases.

If you have any good ideas, I'd really like to hear them and think them through carefully.

lrh2000: Oh, I see this point now. I failed to realize that we lose ownership once we pass the transport to the initialization method.

Another solution is to simply store the guard type (or whatever it should be named --- I'm not 100% sure that "guard" is still a suitable name) in the device struct?

The current design is:
```rust
pub struct TransportGuard {
    inner: Option<Box<dyn VirtioTransport>>,
    is_completed: bool,
}
```

If we keep the current design, it should be simplified to:
```rust
pub struct TransportGuard {
    inner: Option<Box<dyn VirtioTransport>>,
    // Remove `is_completed`!
}
```
This is because `is_completed == inner.is_some()`. So there is no need to maintain `is_completed` separately.

In the alternative design, the device struct just owns the newly introduced guard type:
```rust
pub struct TransportGuard {
    inner: Box<dyn VirtioTransport>, // Remove `Option`!
    is_completed: bool,
}

impl TransportGuard {
    fn finish_init(&mut self) [
        self.is_completed = true;
        // Set DRIVER_OK.
    }
}

impl Drop for TransportGuard {
    // If `!is_completed`, set FAILED.
}
```

> It should be noted that `TransportGuard::into_transport()` takes ownership of the transport, so after this call, if subsequent operations (such as `register_xxx_callback`) fail, it is no longer possible to set the FAILED state via the transport.

This problem will go away.

I feel it is slightly better. The only drawback is that a byte is wasted in the device structure because the completed boolean is useless once initialization finishes. What do you think about it?

kernel/comps/virtio/src/device/socket/queue.rs
hburaylee: Thanks for the careful review!

1. Moved `notify_if_needed()` to the end of each impl block in `TxQueue / RxQueue /EventQueue`, following top-down reading order.
2. Dropped `pub(super)` from `TxQueue::notify_if_needed()`, it's only called internally within `TxQueue`.

kernel/comps/virtio/src/lib.rs
hburaylee: Thank you for the detailed suggestion.

For the alternative design you proposed where the device struct owns the 'guard' directly .I think it is a cleaner approach. 

And i renamed `TransportGuard` to `DeviceTransport` because 'guard' was not the right name, this field stays alive for the whole lifetime of the device.

I agree that `is_completed` wastes a byte, but I think it's still necessary. It allows us to follow the VirtIO spec more strictly and accurately. The design is cleaner this way, and I think the cost is worth it.

I pushed the code already so we can see the whole diff clearly.

kernel/comps/virtio/src/transport/mod.rs
```diff
@@ -48,13 +51,8 @@ pub trait VirtioTransport: Sync + Send + Debug {
...
-        self.write_device_status(
-            DeviceStatus::ACKNOWLEDGE
-                | DeviceStatus::DRIVER
-                | DeviceStatus::FEATURES_OK
-                | DeviceStatus::DRIVER_OK,
-        )
-        .unwrap();
+        let status = self.read_device_status() | DeviceStatus::DRIVER_OK;
+        self.write_device_status(status).unwrap();
     }
```
lrh2000: I believe we can remove this method now. Just call `read_device_status` and `write_device_status` in `DeviceTransport::finish_init` is fine?

kernel/comps/virtio/src/device/network/device.rs
```diff
@@ -142,6 +137,10 @@ impl NetworkDevice {
             .register_queue_callback(QUEUE_RECV, Box::new(handle_recv_event), true)?;
 
         device.transport.finish_init();
+        if device.recv_queue.should_notify() {
```
lrh2000, line 141: Add a blank line before this line. It seems that you have this blank line for all other case.

lrh2000 (commented):
Thanks for the revision. Only two minor nits.

kernel/comps/virtio/src/transport/mod.rs
hburaylee: Agreed. Removed the `finish_init` method and moved the status read/write logic into `DeviceTransport::finish_init`.

kernel/comps/virtio/src/device/network/device.rs
hburaylee: Done. Added the blank line.

lrh2000 (approved):

sdww0 (approved):
LGTM

tatetian (approved):

PR #3466, opened 2026-06-26
Use mainline `inventory` dependency

lrh2000:
https://github.com/asterinas/asterinas/pull/2529 says that the mainline `inventory` dependency works. Let's use it.


lrh2000:
@tatetian Could you merge this trivial PR? This PR only modifies one line in `Cargo.toml` and I think it is not controversial.

tatetian (approved):

PR #3468, opened 2026-06-27
Increase vDSO clock update frequency  to fix `__vdso_time`

### LINKED ISSUE #3459: System time does not update smoothly
### Describe the bug

When running Asterinas with `make run_kernel`, the system time inside the guest VM does not advance continuously. The `date` / `date -R` command often returns the exact same timestamp even after `sleep 1` (or longer), and only jumps forward after a significant delay.

### Steps to reproduce

1. Use the latest main branch (commit range shown below).
2. Build and run: `make run_kernel`
3. Inside the guest shell, repeatedly run:
   ```bash
   sleep 1; date -R
   ```

### Expected behavior

Time should advance roughly every second (or at the timer tick rate).

### Actual behavior

```bash
~ # sleep 1 ; date -R
Fri, 26 Jun 2026 01:53:59 +0000
~ # 
~ # 
~ # sleep 1 ; date -R
Fri, 26 Jun 2026 01:53:59 +0000
~ # 
~ # 
~ # sleep 1 ; date -R
Fri, 26 Jun 2026 01:53:59 +0000
~ # 
~ # 
~ # sleep 1 ; date -R
Fri, 26 Jun 2026 01:53:59 +0000
~ # 
~ # 
~ # sleep 1 ; date -R
Fri, 26 Jun 2026 01:54:49 +0000   <--- Jump suddenly for about 50 seconds
~ # 
~ # 
~ # sleep 1 ; date -R
Fri, 26 Jun 2026 01:54:49 +0000
~ # 
```

### Version / Commit information

```bash
asterinas $ git log --pretty=format:"%h %ad | %s"
e80904666 Mon Jun 22 16:49:54 2026 +0800 | Fix fallocate errno for special files
a24c3ea75 Fri Jun 12 09:47:29 2026 +0000 | Replace libflate with zune-inflate for initramfs decompression
83ace475a Wed Jun 24 17:20:46 2026 +0800 | Update gvisor/blocklists/open_test
e8e12d194 Wed Jun 24 16:39:47 2026 +0800 | Add birth time (birth_at) support to inode metadata
```

hxjerry:
#3459 is because __vdso_time directly reads basetime from vDSO, which has accurate seconds in Linux. This increases clock update frequency to provide accurate seconds.

Fixes #3459

### DIFF
--- a/kernel/comps/time/src/clocksource.rs
+++ b/kernel/comps/time/src/clocksource.rs
@@ -98,9 +98,17 @@ impl ClockSource {
             (self.read_cycles(), last_instant, last_cycles)
         };
 
-        let delta_nanos = {
-            let delta_cycles = instant_cycles - last_cycles;
-            self.cycles_to_nanos_lossy(delta_cycles)
+        let delta_nanos = match instant_cycles.checked_sub(last_cycles) {
+            Some(delta) => self.cycles_to_nanos_lossy(delta),
+            None => {
+                ostd::warn!(
+                    "The clock source becomes not reliable since TSC \
+                    has moved backwards from {} to {}",
+                    last_cycles,
+                    instant_cycles
+                );
+                0
+            }
         };
         let duration = Duration::from_nanos(delta_nanos);
         (last_instant + duration, instant_cycles)
--- a/kernel/comps/time/src/tsc.rs
+++ b/kernel/comps/time/src/tsc.rs
@@ -66,14 +66,13 @@ fn update_clocksource() {
 static TSC_UPDATE_COUNTER: AtomicU64 = AtomicU64::new(1);
 
 fn init_timer() {
-    // The `max_delay_secs` should be set as `clock.max_delay_secs() >> 1` or something much smaller than `max_delay_secs`.
-    // This is because the initialization of this timer occurs during system startup,
-    // and the system will also undergo other initialization processes, during which time interrupts are disabled.
-    // This results in the actual trigger time of the timer being delayed by about 5 seconds compared to the set time.
-    // If without KVM, the delayed time will be larger.
-    // TODO: This is a temporary solution, and should be modified in the future.
-    let max_delay_secs = CLOCK.get().unwrap().max_delay_secs() >> 1;
-    let delay_counts = TIMER_FREQ * max_delay_secs;
+    // This must be frequent enough to provide values accurate to the second
+    // for the time fields in vDSO. We choose 10 Hz, which results in a
+    // worst-case staleness of ~100 ms.
+    // TODO: Implement a more complete and efficient timekeeping mechanism,
+    // then align this update frequency with Linux.
+    const VDSO_UPDATE_FREQ: u64 = 10;
+    let delay_counts = TIMER_FREQ / VDSO_UPDATE_FREQ;
 
     let update = move || {
         let counter = TSC_UPDATE_COUNTER.fetch_add(1, Ordering::Relaxed);
--- a/kernel/src/vdso.rs
+++ b/kernel/src/vdso.rs
@@ -274,9 +274,7 @@ impl Vdso {
         data.update_high_res_instant(instant, instant_cycles);
 
         // Update begins.
-        self.data_frame
-            .write_once(vdso_data_field_offset!(seq), &1)
-            .unwrap();
+        self.begin_update_data_frame(&mut data);
 
         self.data_frame
             .write_val(vdso_data_field_offset!(last_cycles), &instant_cycles)
@@ -286,11 +284,7 @@ impl Vdso {
         }
 
         // Update finishes.
-        // FIXME: To synchronize with the vDSO library, this needs to be an atomic write with the
-        // Release memory order.
-        self.data_frame
-            .write_once(vdso_data_field_offset!(seq), &0)
-            .unwrap();
+        self.finish_update_data_frame(&mut data);
     }
 
     fn update_coarse_res_instant(&self, instant: Instant) {
@@ -299,19 +293,22 @@ impl Vdso {
         data.update_coarse_res_instant(instant);
 
         // Update begins.
-        self.data_frame
-            .write_once(vdso_data_field_offset!(seq), &1)
-            .unwrap();
+        self.begin_update_data_frame(&mut data);
 
         for clock_id in COARSE_RES_CLOCK_IDS {
             self.update_data_frame_instant(clock_id, &mut data);
         }
 
         // Update finishes.
-        // FIXME: To synchronize with the vDSO library, this needs to be an atomic write with the
-        // Release memory order.
+        self.finish_update_data_frame(&mut data);
+    }
+
+    fn begin_update_data_frame(&self, data: &mut VdsoData) {
+        debug_assert!(data.seq.is_multiple_of(2));
+
+        data.seq = data.seq.wrapping_add(1);
         self.data_frame
-            .write_once(vdso_data_field_offset!(seq), &0)
+            .write_once(vdso_data_field_offset!(seq), &data.seq)
             .unwrap();
     }
 
@@ -329,6 +326,17 @@ impl Vdso {
             .unwrap();
         self.data_frame
             .write_val(nanos_info_offset, &data.basetime[clock_index].nanos_info)
+            .unwrap();
+    }
+
+    fn finish_update_data_frame(&self, data: &mut VdsoData) {
+        debug_assert!(!data.seq.is_multiple_of(2));
+
+        data.seq = data.seq.wrapping_add(1);
+        // FIXME: To synchronize with the vDSO library, this needs to be an atomic write with the
+        // Release memory order.
+        self.data_frame
+            .write_once(vdso_data_field_offset!(seq), &data.seq)
             .unwrap();
     }
 }


kernel/comps/time/src/tsc.rs
```diff
@@ -73,7 +73,9 @@ fn init_timer() {
     // If without KVM, the delayed time will be larger.
     // TODO: This is a temporary solution, and should be modified in the future.
     let max_delay_secs = CLOCK.get().unwrap().max_delay_secs() >> 1;
-    let delay_counts = TIMER_FREQ * max_delay_secs;
+
+    const MAX_REFRESH_TICKS: u64 = TIMER_FREQ / 4; // Cap refresh ticks to satisfy precision requirements of vDSO
+    let delay_counts = (TIMER_FREQ * max_delay_secs).min(MAX_REFRESH_TICKS);
```
cchanging: Making this timestamp update more frequently would be more in line with Linux’s approach, but some changes may be needed here:

1. Previously, since we only considered vDSO usage such as `clock_gettime` and only considered the case where the vDSO data segment is used as the base time, we used an update frequency related to `max_delay_secs`. However, once the update frequency is increased, it is no longer really tied to `max_delay_secs`, so the above calculation can be removed.

2. The new update frequency, `TIMER_FREQ / 4`, needs an explanation for why this value was chosen. Ideally, there should be either a reference or a reasonable comment explaining the rationale.

cchanging (commented):
Thanks for fixing the bug of vdso usages!

It seems that the `vdso_clock_gettime_test` CI fails. I suspect that increasing the update frequency has made some previously overlooked races occur more consistently.

My guess is that the issue may be with how we handle `seq` when writing the vDSO data (not sure, you can do some additional work to figure out). In Asterinas, the current update logic is:

```text
seq = 1
write last_cycles / basetime
seq = 0
```

Linux, on the other hand, uses a monotonically increasing sequence number:

```text
even -> odd -> next even
```

On the read side, it records the initial even `seq`, and if it sees that `seq` has changed by the end, it retries the read. With Asterinas’s `0 -> 1 -> 0` scheme, if the reader starts when `seq=0` and a full update happens in between, then sees `seq=0` again afterward, it may incorrectly assume nothing changed and thus potentially observe mixed data.

You could try fixing this issue so that CI can pass reliably.

hxjerry:
Thanks. It was indeed caused by the reader seeing seq0->seq0 while a update has occurred. Implemented incremental seq number to fix this. I was not able to reproduce the current `gettimeofday02` error, still working on it.

cchanging:
>  I was not able to reproduce the current gettimeofday02 error, still working on it.

I looked at the CI log, and the time difference was nearly 100 seconds. I suspect this may be related to `ClockSource::calculate_instant`. We currently don’t have any mechanism to synchronize TSC counters across CPUs, so in a multicore scenario, when calculating `delta_nanos`, it’s possible that the `instant_cycles` read on the current CPU is smaller than the `last_cycles` that was just recorded on another CPU. That would cause the subtraction to underflow, which then triggers the overflow handling in `cycles_to_nanos_lossy`, and the resulting time discrepancy happens to be around 100 seconds.

So I think we may need to change it like this:
```rust
        let delta_nanos = match instant_cycles.checked_sub(last_cycles) {
            Some(delta_cycles) => self.cycles_to_nanos_lossy(delta_cycles),
            None => {
                 ostd::warn!("...");
                 0,
            }
        };
```

kernel/comps/time/src/tsc.rs
hxjerry: Linux updates vDsO every kernel tick, thus commonly 100Hz or 1000Hz. Previously 4Hz was chosen as a relatively low value that can provide good precision for the second. Added a explicit constant for the update frequency. Also removed `max_delay_secs`.

kernel/comps/time/src/clocksource.rs
```diff
@@ -99,7 +99,18 @@ impl ClockSource {
...
+                    ostd::warn!(
+                        "The clock source becomes not reliable since TSC \
+                        has moved backwards from {} to {}",
+                        last_cycles,
+                        instant_cycles
+                    );
+                    0
+                }
+            };
             self.cycles_to_nanos_lossy(delta_cycles)
```
cchanging: Why not use:
```rust
 let delta_nanos = match instant_cycles.checked_sub(last_cycles) {
    Some(delta_cycles) => self.cycles_to_nanos_lossy(delta_cycles),
    None => {
         ostd::warn!("...");
         0,
    }
};
```

It will be a bit simpler and can avoid calling `cycles_to_nanos_lossy` when the result of `checked_sub` is `None`.

kernel/comps/time/src/tsc.rs
```diff
@@ -72,8 +72,9 @@ fn init_timer() {
     // This results in the actual trigger time of the timer being delayed by about 5 seconds compared to the set time.
     // If without KVM, the delayed time will be larger.
     // TODO: This is a temporary solution, and should be modified in the future.
```
cchanging: These can be removed.

kernel/comps/time/src/tsc.rs
```diff
@@ -72,8 +72,9 @@ fn init_timer() {
...
-    let max_delay_secs = CLOCK.get().unwrap().max_delay_secs() >> 1;
-    let delay_counts = TIMER_FREQ * max_delay_secs;
+
+    const VDSO_UPDATE_FREQ: u64 = 10; // Sets frequency for vDsO updates.
+    let delay_counts = TIMER_FREQ / VDSO_UPDATE_FREQ;
```
cchanging: Well, since we currently don’t have a complete timekeeping module, nor a dedicated CPU for timekeeping, maybe updating too frequently might hurt performance. I think it would help if you added a comment explaining the current choice of `VDSO_UPDATE_FREQ`—using a magic number without any comment is always confusing:

```rust
// This update frequency provides reasonably good time accuracy 
// without causing noticeable performance impact.
// TODO: Implement a more complete and efficient timekeeping mechanism,
// then align this update frequency with Linux.
const VDSO_UPDATE_FREQ: u64 = 10;
```

kernel/src/vdso.rs
```diff
@@ -273,23 +277,27 @@ impl Vdso {
...
+
         // Update finishes.
-        // FIXME: To synchronize with the vDSO library, this needs to be an atomic write with the
-        // Release memory order.
+        data.seq = data.seq.wrapping_add(1);
         self.data_frame
-            .write_once(vdso_data_field_offset!(seq), &0)
+            .write_once(vdso_data_field_offset!(seq), &data.seq)
             .unwrap();
     }
```
cchanging: Can we provide two methods for `Vdso` to aovid some duplicate code snippets?

```rust
impl Vdso {
    fn begin_update_data_frame(&self, &mut data) {
        debug_assert!(data.seq.is_multiple_of(2));

        data.seq = data.seq.wrapping_add(1);
        self.data_frame
            .write_once(vdso_data_field_offset!(seq), &data.seq)
            .unwrap();

        fence(Ordering::Release);
    }
    
    fn finish_updata_data_frame(&self, &mut data) {
        ...
    }
}
```

cchanging (commented):
Thanks for the revision!

kernel/src/vdso.rs
lrh2000: `fence(Ordering::Release)` has nothing to do with `read_once`. Memory model is a **very** complex topic. Please avoid removing the original FIXME and adding the `fence(Ordering::Release)`.

1. `read_once` is **not** an atomic operation. It is a single-instruction (but non-atomic) operations. 
2. `fence()` [has nothing to do without atomic operations](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#mandatory-atomic).
3. Even we treat `read_once` as an atomic operation, its relationship with `fence` is **still not** clear because we need to consider if they're in the same memory model.
4. Having only `fence(Release)` does nothing. Some acquire operations are mandatory to establish the synchronization. Here, the `Acquire` operation is in the vDSO library, so we need to determine if it uses the same memory model as the Rust code to ensure proper synchronization.

hxjerry:
Apologies for not being familiar with Rust's memory model. Done some reading on that. Fixed issues and added a explanatory comment on `VDSO_UPDATE_FREQ`.

kernel/src/vdso.rs
```diff
@@ -268,15 +268,33 @@ impl Vdso {
...
+
+    fn finish_update_data_frame(&self, data: &mut VdsoData) {
+        debug_assert!(!data.seq.is_multiple_of(2));
+
+        data.seq = data.seq.wrapping_add(1);
+        self.data_frame
+            .write_once(vdso_data_field_offset!(seq), &data.seq)
+            .unwrap();
+        // FIXME: To synchronize with the vDSO library, this needs to be an atomic write with the
+        // Release memory order.
```
cchanging: Move this FIXME above, like it was in the previous version of the code.

kernel/comps/time/src/tsc.rs
```diff
@@ -66,14 +66,10 @@ fn update_clocksource() {
...
-    // This is because the initialization of this timer occurs during system startup,
-    // and the system will also undergo other initialization processes, during which time interrupts are disabled.
-    // This results in the actual trigger time of the timer being delayed by about 5 seconds compared to the set time.
-    // If without KVM, the delayed time will be larger.
-    // TODO: This is a temporary solution, and should be modified in the future.
-    let max_delay_secs = CLOCK.get().unwrap().max_delay_secs() >> 1;
-    let delay_counts = TIMER_FREQ * max_delay_secs;
+    // `__vdso_time` uses `basetime` vDSO field directly, so this must update frequent enough
+    // to provide second accurate values, chose 10Hz for ~100ms worst case staleness.
+    const VDSO_UPDATE_FREQ: u64 = 10;
```
cchanging: Maybe also add a `TODO` like the [comment](https://github.com/asterinas/asterinas/pull/3468#discussion_r3497333774)?

cchanging (approved):
Only some minor comments, otherwise LGTM!

hxjerry:
Thanks. Fixed the comments.

kernel/comps/time/src/tsc.rs
```diff
@@ -66,14 +66,12 @@ fn update_clocksource() {
...
-    // and the system will also undergo other initialization processes, during which time interrupts are disabled.
-    // This results in the actual trigger time of the timer being delayed by about 5 seconds compared to the set time.
-    // If without KVM, the delayed time will be larger.
-    // TODO: This is a temporary solution, and should be modified in the future.
-    let max_delay_secs = CLOCK.get().unwrap().max_delay_secs() >> 1;
-    let delay_counts = TIMER_FREQ * max_delay_secs;
+    // `__vdso_time` uses `basetime` vDSO field directly, so this must update frequent enough
+    // to provide second accurate values, chose 10Hz for ~100ms worst case staleness.
+    // TODO: Implement a more complete and efficient timekeeping mechanism,
+    // then align this update frequency with Linux.
```
lrh2000:
```suggestion
    // This must be frequent enough to provide values accurate to the second
    // for the time fields in vDSO. We choose 10 Hz, which results in a
    // worst-case staleness of ~100 ms.
    // TODO: Implement a more complete and efficient timekeeping mechanism,
    // then align this update frequency with Linux.
```

No need to mention implementation details of vDSO routines here, which does not help understand.

In addition, there are some grammar issues...

kernel/src/vdso.rs
```diff
@@ -268,15 +268,33 @@ impl Vdso {
...
+    fn finish_update_data_frame(&self, data: &mut VdsoData) {
+        debug_assert!(!data.seq.is_multiple_of(2));
+
+        data.seq = data.seq.wrapping_add(1);
+        // FIXME: To synchronize with the vDSO library, this needs to be an atomic write with the
+        // Release memory order.
+        self.data_frame
+            .write_once(vdso_data_field_offset!(seq), &data.seq)
+            .unwrap();
+    }
```
lrh2000: Define `begin_update_data_frame` and `finish_update_data_frame` after `update_coarse_res_instant`. We prefer [`top-down reading`](https://github.com/hxjerry/asterinas/blob/13d2460fc2cc66c5e7a79d5015672e88dfc735f7/book/src/to-contribute/coding-guidelines/for-maintainability/layout.md?plain=1#L3).

The ideal way to arrange may be: `begin_update_data_frame`, `update_data_frame_instant`, `finish_update_data_frame`, so the call order and the method layout will match.

```rust
    fn update_high_res_instant(&self, instant: Instant, instant_cycles: u64) {
        begin_update_data_frame();
        update_data_frame_instant();
        finish_update_data_frame();
    }

    fn update_coarse_res_instant(&self, instant: Instant) {
        begin_update_data_frame();
        update_data_frame_instant();
        finish_update_data_frame();
    }

    fn begin_update_data_frame() {}

    fn update_data_frame_instant() {}

    fn finish_update_data_frame() {}
```

lrh2000 (commented):
Please squash the commit history and ensure that each commit contains atomic changes. You can just squash all commits here into one commit.

lrh2000 (approved):

PR #3469, opened 2026-06-27
Support clone-family ptrace-event-stops

vvvvsv:
This PR adds support for the clone-family ptrace-event-stops (`PTRACE_EVENT_{FORK,VFORK,VFORK_DONE,CLONE}`): the tracee enters the matching event-stop and the new child is auto-attached to the tracer and stopped with `SIGSTOP`.

This PR also includes a cleaner fix than #3203. A freshly cloned thread should handle its pending signals before its first return to user space, else a cloned child who exits immediately may skip the SIGSTOP-delivery-stop.

Enables the clone-family cases in `set_options.c` on Asterinas.

Required for #2323.


vvvvsv:
Sorry for the long absence, I've been very BUSY this month......

PR #3418, opened 2026-06-21
RFC: hypervisor support.

Cosh-y:
RFC in draft stage of hypervisor support in asterinas.

[Original markdown file](https://github.com/Cosh-y/asterinas/blob/hypervisor-rfc/book/src/rfcs/0000-hypervisor-module.md)
[Rendered preview](https://cosh-y.github.io/asterinas/rfcs/0003-hypervisor-support.html)

PR #3471, opened 2026-06-28
Feat(hypervisor): vmx operations.

Cosh-y:
I am developing the hypervisor module for Asterinas. To support the hyprvisor function, I extended ostd and the design is in RFC: https://github.com/asterinas/asterinas/pull/3418.
This PR introduces a very basic module from my extension of ostd, and its function is to perform **vmx operations(Intel Virtual Machine Extensions)**. The code in this PR does not expose interfaces to the kernel. All the functions are used to support the subsequent ostd implementation.
I added ktests as unit tests.

### DIFF
--- a/ostd/src/arch/x86/mod.rs
+++ b/ostd/src/arch/x86/mod.rs
@@ -15,6 +15,7 @@ pub mod serial;
 pub(crate) mod task;
 mod timer;
 pub mod trap;
+pub(crate) mod vm;
 
 #[cfg(feature = "cvm_guest")]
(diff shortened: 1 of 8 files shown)


Cosh-y:
- Two failed CI/CD checks seem to have nothing to do with my code.
- Is the size of this pr appropriate? To ensure that the code can pass the compilation, I need to submit all the files that the code dependent. Additionally, only one pull request has been submitted so far. Excessive pull requests would increase the difficulty for me to synchronize the code while developing.
@tatetian Thanks.

PR #3472, opened 2026-06-28
Add support for generating Asterinas kernel ISO installer images

bet4it:
## Description
This PR implements bootable ISO image generation support for Asterinas NixOS. It introduces configuration parameters and custom boot scripts to package the Asterinas kernel alongside a NixOS installation closure into a bootable ISO (utilizing a ramfs root with a pre-populated Nix store). 

This allows users and developers to test the Asterinas NixOS within QEMU VMs using a CD-ROM image.

### 1. Refactored NixOS Stage-1 Init & Base Configurations
* Extracted shared configurations from `core.nix` into `asterinas-base.nix` and `asterinas-stage-1-init.nix`.
* Unified the stage-1 init script to dynamically handle two boot modes:
  * **Disk mode** (when `root=` is specified): mounts the physical root ext2 partition and performs `switch_root`.
  * **ISO mode** (when `root=` is absent): mounts a `ramfs` root filesystem, copies Nix store/system closures, and executes the stage-2 init directly.

### 2. Custom ISO Packaging (`iso_image/default.nix`)
* Added the `useAsterinasKernel` parameter to toggle between standard Linux ISO and Asterinas ISO generation.
* Created a modified version of NixOS's `iso-image.nix` to configure GRUB to load the Asterinas kernel (`aster-kernel`) and boot via the custom `initrd`.
* Packaged the full NixOS store closure into the ISO's initrd (using `closureInfo`), including the Nix database registration so Nix commands are functional upon boot.

### 3. Patched Systemd & Build Optimizations
* Updated `systemd.nix` to supply stubs and placeholder services for unsupported systemd components (`systemd-logind`, `systemd-user-sessions`, `systemd-bsod`, etc.) to prevent service startup failures in the ramfs environment.

### 4. Kernel Probe & Installation Script Updates
* Added `/init` to the default init executable fallback probe list in `init_proc.rs` to support booting directly from `initramfs`/`cpio`.
* Enhanced `aster-nixos-install` script to read `/proc/partitions` and manually create partition block device nodes (via `mknod`) if `udev` is not active in the installer environment.

---

## Verification & Testing

The ISO image was successfully built and tested in a KVM VM:
1. **Build ISO**:
   ```bash
   USE_ASTERINAS_KERNEL=true make iso
   ```
2. **Run in QEMU**:
   ```bash
   ENABLE_KVM=1 ./tools/nixos/run.sh iso
   ```
3. **Observation**:
   The VM boots correctly, enters the stage-2 shell, auto-logs into the root prompt on `hvc0`.

### DIFF
--- a/.github/workflows/test_iso_image.yml
+++ b/.github/workflows/test_iso_image.yml
@@ -35,3 +35,38 @@ jobs:
       - name : Run Asterinas NixOS
         run: |
           make run_nixos NIXOS_TEST_SUITE=hello
+
+  iso-test-asterinas-kernel:
+    runs-on: ubuntu-4-cores-150GB-ssd
+    container:
+      image: asterinas/asterinas:0.18.0-20260618
+      options: -v /dev:/dev --privileged
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build Asterinas NixOS ISO installer image with Asterinas kernel
+        run: |
+          AUTO_INSTALL=false USE_ASTERINAS_KERNEL=true make iso NIXOS_TEST_SUITE=hello
+      - name: Run ISO image with Asterinas kernel
+        run: |
+          set -e
+          mkfifo qemu.input
+          (
+            sleep 900
+            echo uname -a
+            sleep 10
+            echo poweroff
+          ) > qemu.input &
+          input_pid=$!
+          trap 'kill "${input_pid}" 2>/dev/null || true; rm -f qemu.input' EXIT
+
+          ENABLE_KVM=0 timeout 30m make run_iso < qemu.input || true
+          if ! grep -a -q "Linux nixos 5.13.0 .* x86_64 GNU/Linux" qemu.log; then
+            echo "Test ISO failed"
+            echo "===== qemu.log ====="
+            tail --lines 200 qemu.log
+            echo "===== qemu-serial.log ====="
+            tail --lines 200 qemu-serial.log || true
+            exit 1
+          fi
+          echo "Test ISO succeeds!"
--- a/book/src/distro/README.md
+++ b/book/src/distro/README.md
@@ -32,11 +32,11 @@ using the Asterinas NixOS installer ISO.
     docker run -it --privileged --network=host ubuntu:latest bash
     ```
 
-2. Inside the container, install QEMU:
+2. Inside the container, install QEMU and OVMF:
 
     ```bash
     apt update
-    apt install -y qemu-system-x86 qemu-utils
+    apt install -y qemu-system-x86 qemu-utils ovmf
     ```
 
 3. Download the latest Asterinas NixOS installer ISO from [GitHub Releases](https://github.com/asterinas/asterinas/releases).
@@ -54,12 +54,17 @@ one is the installer CD-ROM and the other is the target disk:
     ```bash
     export INSTALLER_ISO=/path/to/your/downloaded/installer.iso
     qemu-system-x86_64 \
+    -bios /usr/share/qemu/OVMF.fd \
     -cpu host -m 8G -enable-kvm \
+    -machine q35,kernel-irqchip=split \
     -drive file="$INSTALLER_ISO",media=cdrom -boot d \
-    -drive if=virtio,format=raw,file=aster_nixos_disk.img \
+    -drive if=none,format=raw,id=hd0,file=aster_nixos_disk.img \
+    -device virtio-blk-pci,drive=hd0,disable-legacy=on,disable-modern=off \
     -chardev stdio,id=mux,mux=on,logfile=qemu.log \
     -device virtio-serial-pci -device virtconsole,chardev=mux \
     -serial chardev:mux -monitor chardev:mux \
+    -netdev user,id=net0 \
+    -device virtio-net-pci,netdev=net0,disable-legacy=on,disable-modern=off,mrg_rxbuf=off,ctrl_rx=off,ctrl_rx_extra=off,ctrl_vlan=off,ctrl_vq=off,ctrl_guest_offloads=off,ctrl_mac_addr=off,event_idx=off,queue_reset=off,guest_announce=off,indirect_desc=off \
     -nographic
     ```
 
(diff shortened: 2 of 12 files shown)


cqs21:
Thanks for the work here! A few questions and observations before I can fully evaluate this.

1. **Could you clarify the motivation or target use case?**

I'm not yet clear on what concrete problem this PR is meant to solve. It would help to spell out the intended workflow this unlocks that we can't already do today.

2. **The ramfs switch_root path looks close to what `make run_kernel` already gives us.**

As I read it, the ISO path with `useAsterinasKernel` boots into a `ramfs` rootfs via `switch_root`. But functionally that seems very close to the shell we can get from `make run_kernel` — that also produces an ISO containing just the Asterinas kernel + an initramfs (the differences being mainly **no** `switch_root` and a possibly different set of mounts). For development and testing I don't think this changes much, since our current test matrix already covers both ends:
- `make run_kernel` → busybox shell + ramfs
- `make run_nixos` → NixOS shell + ext2

3. **The Asterinas-kernel-in-ISO-image change is genuinely useful, though.**

The part that swaps the Linux kernel in the ISO image for the Asterinas kernel is valuable on its own: the end goal is to use Asterinas kernel + ISO installer to install NixOS, and this is a necessary building block for that.

4. **It doesn't boot to a shell in my local test.**

I tried this PR locally with:
```
USE_ASTERINAS_KERNEL=true make iso
ENABLE_KVM=1 ./tools/nixos/run.sh iso
```
It reaches "Spawn the first kernel thread", then after a short while the VM exits on its own without ever reaching a shell. I haven't pinned down the cause yet.

bet4it:
Thanks for your review and feedback!

Regarding your questions:

#### Motivation / Target Use Case

The ultimate goal is to support installing NixOS from an Asterinas-running ISO installer onto a raw disk. This PR provides the installer ISO environment that makes this flow possible.

#### Why the  ramfs  copy and  switch_root  in Stage-1 is necessary
Under vanilla Linux, a NixOS Live CD (ISO) boots from a read-only SquashFS, and Linux uses overlayfs  to overlay a writeable  tmpfs  on top of it, making the rootfs writable.
However, Asterinas currently does not support  overlayfs . Without a writable root filesystem:

• Essential system services and PAM login modules (e.g.,  pam_unix  and  pam_lastlog ) will fail to write to `/var/log/lastlog`  or  `/run/utmp` .
• This causes PAM to crash during boot, throwing  Error in service module  and preventing users from logging into the installer environment.

Therefore, copying the system files into a writeable ramfs and performing `switch_root` is the only viable way to boot a functional, interactive NixOS installation environment under Asterinas at this stage.


   
   
#### Local Testing and Booting

I have verified the boot flow by running the following commands in a fresh, clean Docker
  container environment:

    docker run -it --privileged --network=host -v /dev:/dev -v $(pwd):/root/asterinas asterinas/asterinas:0.18.0-20260618

Inside the container:

    USE_ASTERINAS_KERNEL=true make iso
    ENABLE_KVM=1 ./tools/nixos/run.sh iso

In my test, the system booted successfully all the way to the interactive shell without any premature VM exits.

Could you please try doing a clean build (or rebuilding in a fresh Docker container using the command above) to see if there might be any cached build artifacts or environment discrepancies causing the early exit on your end?

cqs21:
Thanks for the detailed clarification.

> However, Asterinas currently does not support overlayfs . Without a writable root filesystem

Replacing `squashfs` with an initrd and `switch_root`-ing into a `ramfs` is a nice idea and works as an interim workaround.
Longer term, though, I'd still prefer using `overlayfs` to build the writable root. Asterinas is actively working on `overlayfs` support and already has an initial implementation (`kernel/src/fs/fs_impls/overlayfs/`), though a few things are still incomplete (e.g. `rename` currently returns `EOPNOTSUPP`).

On the boot issue:

> Could you please try doing a clean build (or rebuilding in a fresh Docker container using the command above) to see if there might be any cached build artifacts or environment discrepancies causing the early exit on your end?

I started a fresh Docker environment using the `asterinas/asterinas:0.18.0-20260618` image and ran the two commands exactly as you described.
Unfortunately, it still cannot boot into the shell. In addition, my testing is based on commit `95d5bd5`.

bet4it:
I found why you may not be able to reproduce.

  #3385 changed initramfs decompression from streaming `libflate` to whole-buffer `zune-inflate`. For the ISO boot path, since Asterinas does not support parsing ISO9660 yet (#3484), most files are packed into the initrd. That makes the initrd very large (~729 MiB compressed, ~2 GiB decompressed).

  With `zune-inflate`, the kernel tries to decompress the whole initrd into memory before unpacking it, and rootfs initialization fails. With the old `libflate` streaming path, the same ISO can boot.

  I noticed that #3385 was authored by you. Could we revert this commit, or switch to https://github.com/rust-lang/flate2-rs instead?

cqs21:
Thanks — great catch.

> That makes the initrd very large (~729 MiB compressed, ~2 GiB decompressed).

A ~2 GiB initrd is unusual, and stage-1 then `cp -a`s the whole closure into a second ramfs before `switch_root`, so the closure is resident twice at peak — the memory cost roughly doubles. I think it's worth reconsidering this approach: long-term I'd still prefer adding `squashfs` support and completing `overlayfs`, so the closure stays read-only on the ISO (never fully resident) and the writable layer is a `tmpfs` upper rather than a full copy.

> Could we revert this commit, or switch to https://github.com/rust-lang/flate2-rs instead?

I'm not against switching decoder, but `flate2` won't work directly here: its public API is built on `std::io::{Read, Write}`, so it isn't usable in the kernel's `no_std` context. Its default backend `miniz_oxide` *is* fully `no_std` (with `alloc`) and exposes a low-level streaming inflate API, so it (or another efficient `no_std` streaming decoder) could be a candidate. Which one we land on should be backed by more test data, though.

bet4it:
Thanks for the insight into the memory footprint! I completely agree that for the long term, supporting `squashfs` and completing `overlayfs` is the right architectural path to avoid pinning the entire closure in memory.

Regarding the streaming decoder, I noticed that according to `libflate`'s [informal benchmark](https://github.com/sile/libflate#an-informal-benchmark), `flate2`'s decoding speed is significantly faster than `libflate`'s. Since `miniz_oxide` serves as the core decompressor for `flate2`, it likely inherits this performance advantage. Would it be possible to run a quick test on `miniz_oxide`'s streaming API within your benchmark environment to see how it compares under the same conditions?

Beyond this specific ISO use case, I feel that having streaming decompression capability in the kernel is highly valuable. For instance, if Asterinas is ported to memory-constrained or embedded environments in the future, a streaming decoder would be crucial for reducing peak memory usage during boot, even if it comes with a slight trade-off in raw decompression speed. 

Looking forward to your thoughts and benchmark data!

Issue #3474, opened 2026-06-28
Tracking issue: Add RISC-V IOMMU support


le-monde-bleu:
## Description

This issue tracks the plan to add RISC-V IOMMU support to Asterinas.

The target is to make Asterinas boot and run correctly on the RISC-V QEMU `virt` platform with IOMMU enabled, especially with:

```bash
make run_kernel TARGET_ARCH=riscv64 SCHEME=iommu
```

The planned work includes:

- enabling the QEMU RISC-V IOMMU test scheme;
- adding the RISC-V IOMMU core data structures and register/queue support;
- wiring DMA remapping into the generic DMA layer;
- adding the RISC-V interrupt delivery pieces needed by QEMU's AIA/IMSIC/APLIC configuration;
- enabling PCI MSI-X message programming on RISC-V;
- adding validation coverage for the pure IOMMU logic and the QEMU boot path.

The immediate goal is functional DMA remapping for RISC-V devices under QEMU, with the minimum interrupt/MSI support needed for devices to keep making progress when IOMMU is enabled. This is not intended to implement every advanced RISC-V IOMMU feature in one step.

## Background

Asterinas already has IOMMU support on x86_64. The current x86 IOMMU code is organized around DMA remapping, interrupt remapping, invalidation, fault handling, and VT-d register access under:

- `ostd/src/arch/x86/iommu/`
- `kernel/comps/pci/src/arch/x86/mod.rs`

That support allows the x86_64 architecture code to provide the architecture side of DMA address translation and interrupt remapping. RISC-V currently lacks an equivalent implementation.

RISC-V IOMMU support is needed because RISC-V systems, including QEMU's `virt` machine, can expose an IOMMU between PCI/MMIO devices and memory. Once the IOMMU is enabled, device DMA addresses are no longer necessarily physical addresses. The kernel must program IOMMU translation structures before devices can DMA safely and correctly.

This also interacts with interrupt delivery. On the QEMU RISC-V `virt` platform, the IOMMU scheme uses AIA with APLIC and IMSIC. PCI devices use MSI/MSI-X-style message interrupts. Therefore, enabling IOMMU for RISC-V is not just a matter of mapping DMA buffers: the PCI/MSI-X and AIA/IMSIC paths must also be sufficient for virtio and other PCI devices to receive completions reliably.

[...]

PR #3475, opened 2026-06-28
Make request state visible before device notification

le-monde-bleu:
## Summary

This PR fixes a request-completion race in the Virtio block driver.

This PR is also a step for RISC-V IOMMU implementation, see #3474 .

Before this PR, the driver notified the device before recording the submitted block request in `submitted_requests`. A fast Virtio device may complete the request and raise an interrupt immediately after the notification, so the completion path can observe a used-ring entry before the driver has recorded the corresponding request metadata.

This PR changes the read, write, and flush request paths to:

1. add descriptors to the virtqueue;
2. compute whether the device should be notified;
3. record the submitted request by token;
4. notify the device.

This makes the driver's internal request state visible before the device is allowed to complete the request.

## Background

The Virtio block completion path pops a used descriptor token from the queue and then removes the corresponding request from `submitted_requests`. This assumes that every token visible in the used ring has already been inserted into the submitted-request table.

The old submission order violated that assumption:

```text
add descriptors -> notify device -> record submitted request
```

https://github.com/asterinas/asterinas/blob/a19974df7f48dda7d1f6540350d787f5d7df7860/kernel/comps/virtio/src/device/block/device.rs#L390-L409

This is racy because Virtio devices are allowed to consume available descriptors after notification. With fast backends, or with interrupt paths that deliver completion interrupts quickly, the completion handler may run before the submitted request has been recorded.

This race is a general Virtio block correctness issue. It is also important for the later RISC-V IOMMU series because the QEMU RISC-V IOMMU setup uses PCI MSI-X/AIA/IMSIC interrupt delivery, where block completion interrupts can arrive quickly once the device is notified.

## References

- Virtio 1.2 specification:
  - https://docs.oasis-open.org/virtio/virtio/v1.2/virtio-v1.2.html

The relevant Virtio behavior is that a driver makes buffers available through a virtqueue and then notifies the device. After notification, the device may consume the available buffers and report completion through the used ring.

## Main Behavior

This PR updates the Virtio block request submission paths in `kernel/comps/virtio/src/device/block/device.rs`.

For read requests, write requests, and flush requests, the driver now:

- adds the DMA buffers to the virtqueue and obtains the queue token;
- reads `queue.should_notify()` after updating the virtqueue;
- inserts the `SubmittedRequest` into `submitted_requests` before notification;
- calls `queue.notify()` only after the request has been recorded.

This preserves the existing virtqueue notification logic while fixing the ordering between driver-side bookkeeping and device notification.

## Testing

The focused branch was validated with:

```bash
cargo check -p aster-virtio --target riscv64imac-unknown-none-elf
```

## Gaps and Future Work

This PR intentionally focuses only on the Virtio block request notification ordering.

It does not add new runtime tests for the race because the failure depends on timing between device notification, request completion, and interrupt handling. The issue is therefore difficult to reproduce deterministically with a simple unit test.

This PR also does not implement the RISC-V IOMMU support itself. It is a standalone Virtio block correctness fix and serves as preparation for the later RISC-V IOMMU PRs, where fast device completion through PCI MSI-X/AIA paths makes this ordering more important.

### DIFF
--- a/kernel/comps/virtio/src/device/block/device.rs
+++ b/kernel/comps/virtio/src/device/block/device.rs
@@ -396,16 +396,16 @@ impl DeviceInner {
             let token = queue
                 .add_dma_bufs(&[&req_slice], outputs.as_slice())
                 .expect("add queue failed");
-            if queue.should_notify() {
-                queue.notify();
-            }
-
-            // Records the submitted request
+            // Record the submitted request before notifying the device so the
+            // driver-side state is published before the device can complete it.
             let submitted_request = SubmittedRequest::new(id as u16, bio_request);
             self.submitted_requests
                 .disable_irq()
                 .lock()
                 .insert(token, submitted_request);
+            if queue.should_notify() {
+                queue.notify();
+            }
             return;
         }
     }
@@ -466,16 +466,16 @@ impl DeviceInner {
             let token = queue
                 .add_dma_bufs(inputs.as_slice(), &[&resp_slice])
                 .expect("add queue failed");
-            if queue.should_notify() {
-                queue.notify();
-            }
-
-            // Records the submitted request
+            // Record the submitted request before notifying the device so the
+            // driver-side state is published before the device can complete it.
             let submitted_request = SubmittedRequest::new(id as u16, bio_request);
             self.submitted_requests
                 .disable_irq()
                 .lock()
                 .insert(token, submitted_request);
+            if queue.should_notify() {
+                queue.notify();
+            }
             return;
         }
     }
@@ -521,16 +521,16 @@ impl DeviceInner {
             let token = queue
                 .add_dma_bufs(&[&req_slice], &[&resp_slice])
                 .expect("add queue failed");
-            if queue.should_notify() {
-                queue.notify();
-            }
-
-            // Records the submitted request
+            // Record the submitted request before notifying the device so the
+            // driver-side state is published before the device can complete it.
             let submitted_request = SubmittedRequest::new(id as u16, bio_request);
             self.submitted_requests
                 .disable_irq()
                 .lock()
                 .insert(token, submitted_request);
+            if queue.should_notify() {
+                queue.notify();
+            }
             return;
         }
     }


le-monde-bleu:
@sdww0 Would you please review this PR? Thanks!

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -395,16 +395,19 @@ impl DeviceInner {
...
                 .add_dma_bufs(&[&req_slice], outputs.as_slice())
                 .expect("add queue failed");
-            if queue.should_notify() {
-                queue.notify();
-            }
+            let should_notify = queue.should_notify();
 
-            // Records the submitted request
+            // Record the submitted request before notifying the device: with
+            // fast devices, the completion interrupt may arrive immediately.
```
sdww0: My understanding is that you deferred the device notification to avoid a race where the device services the request and sends an interrupt before the request is recorded, causing the completion path to fail to find the request entry. 

However, I'm not sure if this race can actually happen, because interrupts are already disabled via `disable_irq` here; even if the virtio backend completes the request, the interrupt won't be handled until after the record is written and we return. (Also, on x86, MSI‑X is used for interrupts.)

kernel/comps/virtio/src/device/block/device.rs
```diff
@@ -395,16 +395,19 @@ impl DeviceInner {
...
             let submitted_request = SubmittedRequest::new(id as u16, bio_request);
             self.submitted_requests
                 .disable_irq()
                 .lock()
                 .insert(token, submitted_request);
+            if should_notify {
```
sdww0: Since we defer notification, could we use `if queue.should_notify()` here?

le-monde-bleu:
> My understanding is that you deferred the device notification to avoid a race where the device services the request and sends an interrupt before the request is recorded, causing the completion path to fail to find the request entry.
> However, I'm not sure if this race can actually happen, because interrupts are already disabled via disable_irq here; even if the virtio backend completes the request, the interrupt won't be handled until after the record is written and we return. (Also, on x86, MSI‑X is used for interrupts.)

@sdww0 Thanks for pointing this out. After looking at the current locking path again, I agree that the concrete completion race is unlikely to happen with the current implementation: the submission path holds the virtqueue lock with local IRQs disabled, and the IRQ completion path also needs to acquire the same queue lock before popping used descriptors and looking up `submitted_requests`.
So this PR should not be presented as fixing a currently reproducible race. The main reason for the change is to make the submission ordering more explicit and robust: the driver records the request metadata before notifying the device, so the driver-side state is already published before the device is asked to consume and complete the descriptor.
This avoids relying too much on the current lock lifetime and interrupt serialization, and makes the code safer for future changes such as moving notification outside the queue lock, adding polling-based completion, or refactoring the completion path. I have also updated the comment to describe this ordering invariant more accurately.

sdww0:
Thanks for the justification! I have one question left, which is about the performance: If we defer the notification, the device will be notified after [L401 ~ L405](https://github.com/asterinas/asterinas/pull/3475/changes#diff-35fe13c352e4363aa572e2fcb3d9977d89d71e8997151ef7383c76ec7a9a9214R401-R405), which may increases the latency of device request completion. 

I think the performance impact of just these two lines shouldn't be significant, but just in case, could you run the [fio benchmark](https://github.com/asterinas/asterinas/tree/main/test/initramfs/src/benchmark/fio) before and after this PR and give us some numbers? That should be helpful.

le-monde-bleu:
@sdww0 
I ran the fio benchmarks before and after this PR.

| Benchmark | System | Before | After | Change |
|---|---|---:|---:|---:|
| fio/ext2_seq_read_bw | Linux | 870 MB/s | 858 MB/s | -1.38% |
| fio/ext2_seq_read_bw | Asterinas | 490 MB/s | 659 MB/s | +34.49% |
| fio/ext2_seq_write_bw | Linux | 640 MB/s | 602 MB/s | -5.94% |
| fio/ext2_seq_write_bw | Asterinas | 769 MB/s | 725 MB/s | -5.72% |
| fio/ext2_seq_read_bw_no_iommu | Linux | 811 MB/s | 916 MB/s | +12.95% |
| fio/ext2_seq_read_bw_no_iommu | Asterinas | 667 MB/s | 608 MB/s | -8.85% |
| fio/ext2_seq_write_bw_no_iommu | Linux | 209 MB/s | 196 MB/s | -6.22% |
| fio/ext2_seq_write_bw_no_iommu | Asterinas | 736 MB/s | 616 MB/s | -16.30% |

For the IOMMU benchmarks, I do not see a clear regression: read improves in this run, and write changes by about the same magnitude as the Linux baseline. The no-IOMMU results are more mixed, especially the write benchmark.

le-monde-bleu:
@sdww0 
I tried a second run, both runs are executed locally on VMware hypervisor.

| Benchmark | System | Before | After | Change |
|---|---|---:|---:|---:|
| fio/ext2_seq_read_bw | Linux | 1040 MB/s | 997 MB/s | -4.13% |
| fio/ext2_seq_read_bw | Asterinas | 826 MB/s | 877 MB/s | +6.17% |
| fio/ext2_seq_write_bw | Linux | 236 MB/s | 225 MB/s | -4.66% |
| fio/ext2_seq_write_bw | Asterinas | 665 MB/s | 746 MB/s | +12.18% |
| fio/ext2_seq_read_bw_no_iommu | Linux | 1126 MB/s | 983 MB/s | -12.70% |
| fio/ext2_seq_read_bw_no_iommu | Asterinas | 820 MB/s | 945 MB/s | +15.24% |
| fio/ext2_seq_write_bw_no_iommu | Linux | 599 MB/s | 219 MB/s | -63.44% |
| fio/ext2_seq_write_bw_no_iommu | Asterinas | 677 MB/s | 639 MB/s | -5.61% |

The results seem intermittent, I am not sure whether there is a robust environment to run the benchmark.

sdww0:
> @sdww0 I ran the fio benchmarks before and after this PR.
> 
> Benchmark	System	Before	After	Change
> fio/ext2_seq_read_bw	Linux	870 MB/s	858 MB/s	-1.38%
> fio/ext2_seq_read_bw	Asterinas	490 MB/s	659 MB/s	+34.49%
> fio/ext2_seq_write_bw	Linux	640 MB/s	602 MB/s	-5.94%
> fio/ext2_seq_write_bw	Asterinas	769 MB/s	725 MB/s	-5.72%
> fio/ext2_seq_read_bw_no_iommu	Linux	811 MB/s	916 MB/s	+12.95%
> fio/ext2_seq_read_bw_no_iommu	Asterinas	667 MB/s	608 MB/s	-8.85%
> fio/ext2_seq_write_bw_no_iommu	Linux	209 MB/s	196 MB/s	-6.22%
> fio/ext2_seq_write_bw_no_iommu	Asterinas	736 MB/s	616 MB/s	-16.30%
> For the IOMMU benchmarks, I do not see a clear regression: read improves in this run, and write changes by about the same magnitude as the Linux baseline. The no-IOMMU results are more mixed, especially the write benchmark.

Maybe try several (like 3 or 5, enable KVM) runs and get the average? 3 out of 4 in Asterinas's numbers are decreased. BTW, Linux benchmarks are unnecessary; you may disable them in the benchmark suite to speed it up.

le-monde-bleu:
@sdww0 
| Benchmark | Main Runs | PR Runs | Main Avg | PR Avg | Change | Main Stddev | PR Stddev |
|---|---|---|---|---|---|---|---|
| fio_ext2_seq_read_bw | 630, 644, 689, 737, 321 | 624, 557, 632, 639, 434 | 604.2 | 577.2 | -4.47% | 163.8 | 86.5 |
| fio_ext2_seq_write_bw | 610, 544, 619, 499, 777 | 659, 658, 634, 733, 862 | 609.8 | 709.2 | +16.30% | 105.6 | 93.2 |
| fio_ext2_seq_read_bw_no_iommu | 758, 677, 763, 698, 728 | 742, 739, 713, 713, 780 | 724.8 | 737.4 | +1.74% | 37.3 | 27.5 |
| fio_ext2_seq_write_bw_no_iommu | 656, 719, 730, 639, 721 | 645, 718, 636, 562, 630 | 693.0 | 638.2 | -7.91% | 42.2 | 55.4 |

The pair result:
| Benchmark | PR-Main Difference by Round | Avg Difference | PR Better Rounds |
|---|---|---|---|
| fio_ext2_seq_read_bw | -6, -87, -57, -98, +113 | -27.0 | 1/5 |
| fio_ext2_seq_write_bw | +49, +114, +15, +234, +85 | +99.4 | 5/5 |
| fio_ext2_seq_read_bw_no_iommu | -16, +62, -50, +15, +52 | +12.6 | 3/5 |
| fio_ext2_seq_write_bw_no_iommu | -11, -1, -94, -77, -91 | -54.8 | 0/5 |

The result for **no-IOMMU sequential write** indicates a potential regression signal. I may add a few more rounds for this benchmark. If the result still indicates a regression, this PR may need reconsideration.

le-monde-bleu:
@sdww0 
I ran 5 more rounds (therefore 10 rounds in total), the results for asterinas benchmarks in round 6-10 is as follow:

| Benchmark | Main Runs | PR Runs | Main Avg | PR Avg | Changes | PR Better Rounds |
| -- | -- | -- | -- | -- | -- | -- |
| fio_ext2_seq_read_bw | 623, 650, 703, 811, 743 | 712, 737, 653, 743, 601 | 706.0 | 689.2 | -2.38% | 2/5 |
| fio_ext2_seq_write_bw | 588, 546, 616, 648, 557 | 795, 558, 509, 569, 550 | 591.0 | 596.2 | +0.88% | 2/5 |
| fio_ext2_seq_read_bw_no_iommu | 655, 769, 846, 682, 875 | 856, 645, 689, 823, 813 | 765.4 | 765.2 | -0.03% | 2/5 |
| fio_ext2_seq_write_bw_no_iommu | 653, 592, 727, 538, 552 | 679, 686, 545, 632, 557 | 612.4 | 619.8 | +1.21% | 4/5 |

The results are still a bit noisy by round, but much better on average. However, there appears regressions in two benchmarks.

Specifically, the results for no-IOMMU Sequential Write benchmark is:

| Round | Main | PR | Changes |
| -- | -- | -- | -- |
| 6 | 653 | 679 | +3.98% |
| 7 | 592 | 686 | +15.88% |
| 8 | 727 | 545 | -25.03% |
| 9 | 538 | 632 | +17.47% |
| 10 | 552 | 557 | +0.91% |

There are 4 out of 5 rounds, in which the result for PR is better than main.

Combined with the previous 5 rounds (so 10 rounds in total):

| Benchmark | Main Avg | PR Avg | Changes | PR Better Rounds |
| -- | -- | -- | -- | -- |
| fio_ext2_seq_read_bw | 655.1 | 633.2 | -3.34% | 3/10 |
| fio_ext2_seq_write_bw | 600.4 | 652.7 | +8.71% | 7/10 |
| fio_ext2_seq_read_bw_no_iommu | 745.1 | 751.3 | +0.83% | 5/10 |
| fio_ext2_seq_write_bw_no_iommu | 652.7 | 629.0 | -3.63% | 4/10 |

Conclusively, there is minor regression on two benchmarks. To be prudent, I guess the changes of this PR should be reconsidered. I may close this PR ater if there is no further comment.

Issue #2546, opened 2025-10-30
Fully enable CI test for RISC-V Asterinas


jellllly420:
### Description

We'd like to support RISC-V as a well-tested architecture, just like we do for x86-64.

### Current Status

Currently the `Test x86-64` action consists of the following test. The status of `Test riscv64` is,
- [x] general test
    - [x] lint
    - [x] compile
    - [x] usermode_test
    - [x] ktest
- [ ] integration test
    - [x] boot
    - [ ] syscall
        - [x] ltp
        - [ ] gvisor
            gvisor officially supports x86_64 and arm64 only. The syscall test may or may not work for riscv64 target.
    - [ ] general
        Some of the tests are x86_64 specific. Besides, some of the tests fail to pass on riscv64. Here are two failed tests I know,
        - if_nameindex
        - get_link_by_libnl
- [ ] osdk test
    Some test are for OSDK itself, which we don't need a per-arch run. Others involve spawning a little kernel base on OSTD. and this part is hard-coded in OSDK test code to be tested on x86_64 only.

### Additional Information 

    
- LTP test under debug mode would consistently timeout, which indicates further implementation problems of some system calls. One who's interested could check the following job status.
https://github.com/jellllly420/asterinas/actions/runs/18929528653/job/54043599324
- We are also working on real board support. Existing work are mostly towards SiFive HiFive Unleashed. Qemu supports emulation of this specific board through "sifive_u" machine so we may also want that in CI.
https://www.qemu.org/docs/master/system/riscv/sifive_u.html

alitariq4589:
If RISC-V CI is something needed for this project, I can provide it (on multiple RISC-V hardware machines). We also have kernelci if this requires power setup and remote power control for flashing OS through host x86 (Reference link: https://cloud-v.co).

tatetian:
> If RISC-V CI is something needed for this project, I can provide it (on multiple RISC-V hardware machines). We also have kernelci if this requires power setup and remote power control for flashing OS through host x86 (Reference link: https://cloud-v.co).

@alitariq4589 Thanks for letting us know. This is an interesting option. We will investigate it and see if it fits the needs of Asterinas.

PR #3476, opened 2026-06-28
Enable riscv64 kselftest baseline build

le-monde-bleu:
## Summary

This PR enables the existing kselftest conformance baseline to build and run on `riscv64`.

This PR is also a step to expand the RISC-V CI test, enabling the `kselftest`, see #2546 .

Before this PR, the kselftest Nix package always included the `exec` selftest target and `glibc_multi.static`. That works for x86_64, but blocks the RISC-V build: `glibc_multi.static` pulls in x86/i686 multilib assumptions, and the upstream `exec` selftests build static/static-pie helper binaries that require static libc support from the target toolchain. Consequently, when running:
```shell
make run_kernel TARGET_ARCH=riscv64 AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest RELEASE=1 INITRAMFS_SKIP_GZIP=1
```
it triggers:
```bash
error:
       … while calling the 'derivationStrict' builtin
         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'initramfs'
         whose name attribute is located at /nix/store/qwsqazbxi81mnj3hf8kpambr0fh687zl-source/pkgs/stdenv/generic/make-derivation.nix:480:13

       … while evaluating attribute 'buildCommand' of derivation 'initramfs'
         at /root/asterinas/test/initramfs/nix/initramfs.nix:28:3:
           27|   name = "initramfs";
           28|   buildCommand = ''
             |   ^
           29|     mkdir -p $out/{dev,etc,root,usr,opt,tmp,var,proc,sys}

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: i686 Linux package set can only be used with the x86 family.
make[1]: *** [Makefile:89: /root/asterinas/test/initramfs/build/initramfs] Error 1
make: *** [Makefile:263: initramfs] Error 2
```

This PR keeps the RISC-V baseline focused on the architecture-neutral kselftest targets that currently build and run: `lsm`, `proc`, `signal`, `splice`, `timers`, and `vDSO`.

## Background

Asterinas already packages Linux kselftest from the pinned Linux tree and builds a selected target set through `test/initramfs/nix/conformance/kselftest.nix`.

Linux kselftest is target-based: upstream supports selecting subsets via `TARGETS`, and installed kselftests are run from the generated test list. This PR uses that target-selection model to avoid making the initial RISC-V baseline depend on tests that currently require unavailable target static libc support.

The upstream `exec` selftests generate `-static` and `-static-pie` helpers, so this PR keeps `exec` in the x86_64-specific target set for now.

## Changes

- Keep `baseKselftestTargets` architecture-neutral by removing `exec`.
- Add `exec` only for x86_64, together with the existing x86-specific target.
- Keep `glibc_multi.static` only in the x86_64 build inputs.
- Leave `libcap.dev` as a common kselftest build input.

## References

- Linux kselftest documentation: https://docs.kernel.org/dev-tools/kselftest.html
- Linux selftests top-level target list: https://github.com/torvalds/linux/blob/v6.18/tools/testing/selftests/Makefile
- Linux `exec` selftest Makefile: https://github.com/torvalds/linux/blob/v6.18/tools/testing/selftests/exec/Makefile
- Asterinas kselftest doc: `test/initramfs/README.md`

## Validation
- `make run_kernel AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest RELEASE=1 INITRAMFS_SKIP_GZIP=1`

runs normally as before.

- `make run_kernel TARGET_ARCH=riscv64 AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest RELEASE=1 INITRAMFS_SKIP_GZIP=1`

Observed result:

- `Found 48 available test cases`
- `Total blocklist entries processed: 91`
- `Test cases to be executed: 11`
- `11 of 11 test cases passed.`
- `All kselftest tests passed.`
- `All conformance tests passed.`

## Gaps and Future Work

This PR intentionally does not expand the RISC-V kselftest coverage beyond the current baseline.

Follow-up work can triage and unblock additional kselftest entries. In particular, enabling the `exec` target on RISC-V should first address the target static libc/toolchain requirement exposed by its upstream `-static` and `-static-pie` helper binaries.

### DIFF
--- a/test/initramfs/nix/conformance/kselftest.nix
+++ b/test/initramfs/nix/conformance/kselftest.nix
@@ -31,17 +31,18 @@ in stdenv.mkDerivation rec {
   # a regression in an unlisted subsystem upstream cannot silently change our conformance surface.
   # To enable a new subsystem,
   # add it here and triage any resulting build/run failures before landing.
-  baseKselftestTargets =
-    [ "exec" "lsm" "proc" "signal" "splice" "timers" "vDSO" ];
+  baseKselftestTargets = [ "lsm" "proc" "signal" "splice" "timers" "vDSO" ];
 
   kselftestTargets = lib.concatStringsSep " " (baseKselftestTargets
-    ++ lib.optionals stdenv.hostPlatform.isx86_64 [ "x86" ]);
+    ++ lib.optionals stdenv.hostPlatform.isx86_64 [ "exec" "x86" ]);
 
   enableParallelBuilding = true;
 
   nativeBuildInputs = with pkgsBuildBuild; [ rsync ];
 
-  buildInputs = with pkgs; [ glibc_multi.static libcap.dev ];
+  buildInputs = with pkgs;
+    [ libcap.dev ]
+    ++ lib.optionals stdenv.hostPlatform.isx86_64 [ glibc_multi.static ];
 
   buildPhase = ''
     runHook preBuild


le-monde-bleu:
@cqs21 Could you please review this PR? Thanks!

tatetian:
> * make run_kernel TARGET_ARCH=riscv64 AUTO_TEST=conformance CONFORMANCE_TEST_SUITE=kselftest RELEASE=1 INITRAMFS_SKIP_GZIP=1

Now that at least one conformance test works for RISC-V, is this PR a good opportunity to enable the conformance test CI for RISC-V?

le-monde-bleu:
@tatetian 

I added a dedicated RISC-V kselftest CI job, included the kselftest package in the prebuilt RISC-V package set, and made the conformance suite configurable in the CI matrix.

During validation, there is an independent issue in the pinned Linux v6.18 source: `proc-pidns.c` uses `open()` and `O_*` flags without including `<fcntl.h>`. I backported the [upstream one-line fix](https://github.com/torvalds/linux/commit/879b3353d04d043a9e01525c520d9b81339421b2) rather than bumping the entire Linux source version. I also added a completeness check in `/test/initramfs/nix/conformace/kselftest.nix` because the upstream build could otherwise skip the failed proc target without failing the package build.

```makefile
    for target in $kselftestTargets; do
      if [ ! -d "$out/$target" ] || ! grep -q "^$target:" "$out/kselftest-list.txt"; then
        echo "kselftest target was not installed completely: $target" >&2
        exit 1
      fi
    done
    runHook postInstall
  '';
```

An alternative would be to move `exec` into the x86_64-specific target list. That would avoid adding the RISC-V static libc dependencies and provide a smaller 11-test RISC-V baseline, but it would also lose useful RISC-V `exec` coverage. The `proc-pidns.c` patch would still be required because that issue is independent of `exec`.

To reduce CI time, `conformance.kselftest` was added to the `riscv64_pkgs` prebuild target, so newly built development images will include the RISC-V kselftest package and its Nix dependencies. CI jobs using such an image can reuse those artifacts instead of rebuilding the dependency closure from scratch; the current PR run remains slower because it still uses an older development image.

PR #3477, opened 2026-06-28
Enable virtio-pci transport on RISCV64

YanLien:
# Motivation

The virtio-pci modern transport was unusable on RISC-V64 for two reasons:

1. **Early panic during probe** at `common_cfg.memory_bar().unwrap()`. Device BARs were never programmed: unlike x86 (where SeaBIOS performs PCI enumeration) or the existing LoongArch path, RISC-V PCI initialization only mapped the ECAM configuration space and left every device BAR uninitialized. RISC-V boots via OpenSBI, which performs no PCI enumeration.
2. **No usable interrupt path.** QEMU's `virt` machine exposes only a PLIC (no IMSIC/APLIC), so MSI-X writes are never delivered — the device's completion interrupt never reaches the CPU, and any block I/O hangs.

The broader goal of this work is to **enable testing of the RISC-V IOMMU DMA-remapping driver**. The IOMMU translates DMA for PCI devices, so exercising it requires a working virtio-pci device that performs DMA — which did not exist on RISC-V64 before this PR. Bringing up virtio-pci (with identity-mapped DMA for now) removes that blocker and is the prerequisite for the IOMMU validation that follows.

# What this PR does

**1. OS-side PCI BAR allocation** (`riscv/mod.rs`, `cfg_space.rs`, `Cargo.toml`)
  - Port the LoongArch MMIO bump allocator into the RISC-V PCI arch module (`MmioAllocator` + `alloc_mmio`), initialized from the host bridge's `ranges` DT property.
  - Extend the `cfg` gates in `cfg_space.rs` from `loongarch64` to `any(loongarch64, riscv64)` so RISC-V takes the manual BAR-allocation path instead of expecting a firmware-programmed base.
  - Add the `fdt` dependency under `[target.riscv64imac-unknown-none-elf]`.

**2. Legacy INTx interrupt delivery** (`riscv/mod.rs`, `common_device.rs`, `device.rs`) 
Since RISC-V `virt` has no IMSIC, MSI-X cannot be delivered, so the transport falls back to legacy INTx routed through the PLIC:
  - `map_intx_interrupt()`: parse the host bridge's `interrupt-map` / `interrupt-map-mask` to resolve a device's INTx pin to a PLIC interrupt source, then map it to an `IrqLine` via `IRQ_CHIP`.
  - `VirtioPciIntxManager`: a shared, level-triggered INTx line with ISR-register-driven dispatch (bit 0 = queue, bit 1 = config-change; reading the ISR register also deasserts INTx).
  - `create_interrupt_manager()` prefers INTx on RISC-V64 and falls back to MSI-X on other architectures.

**3. QEMU switch** (`tools/qemu_args.sh`)
  - Add a `VIRTIO_PCI=on|off` switch (default `off` = MMIO) for the RISC-V scheme that attaches virtio devices as `virtio-*-pci` instead of `virtio-*-device`.

# How to test

```bash
# Boot smoke test
TARGET_ARCH=riscv64 VIRTIO_PCI=on LOG_LEVEL=info AUTO_TEST=boot make run_kernel

# Interactive shell
TARGET_ARCH=riscv64 VIRTIO_PCI=on make run_kernel
```

### DIFF
--- a/kernel/comps/pci/Cargo.toml
+++ b/kernel/comps/pci/Cargo.toml
@@ -15,7 +15,10 @@ ostd.workspace = true
 spin.workspace = true
 
 [target.loongarch64-unknown-none-softfloat.dependencies]
-fdt = { version = "0.1.5", features = ["pretty-printing"] }
+fdt.workspace = true
+
+[target.riscv64imac-unknown-none-elf.dependencies]
+fdt.workspace = true
 
 [lints]
 workspace = true
--- a/kernel/comps/pci/src/arch/riscv/mod.rs
+++ b/kernel/comps/pci/src/arch/riscv/mod.rs
@@ -2,14 +2,29 @@
 
 //! PCI bus access
 
-use core::ops::RangeInclusive;
+use alloc::vec::Vec;
+use core::{alloc::Layout, ops::RangeInclusive};
 
-use ostd::{Error, arch::boot::DEVICE_TREE, io::IoMem, mm::VmIoOnce, warn};
+use fdt::node::FdtNode;
+use ostd::{
+    Error,
+    arch::{
+        boot::DEVICE_TREE,
+        irq::{IRQ_CHIP, InterruptSourceInFdt, MappedIrqLine},
+    },
+    io::IoMem,
+    irq::IrqLine,
+    mm::VmIoOnce,
+    prelude::Paddr,
+    sync::SpinLock,
+    warn,
+};
 use spin::Once;
 
-use crate::PciDeviceLocation;
+use crate::{PciDeviceLocation, cfg_space::PciCommonCfgOffset};
 
 static PCI_ECAM_CFG_SPACE: Once<IoMem> = Once::new();
+static PCI_INTX_MAPPER: Once<PciIntxMapper> = Once::new();
 
 pub(crate) fn write32(location: &PciDeviceLocation, offset: u32, value: u32) -> Result<(), Error> {
     if offset > PCI_ECAM_MAX_OFFSET {
@@ -101,6 +116,12 @@ pub(crate) fn init() -> Option<RangeInclusive<u8>> {
         Some(0..=255)
     };
 
+    // Initialize the MMIO allocator used to assign base addresses to device BARs.
+    // On RISC-V the firmware (OpenSBI) does not perform PCI enumeration, so the
+    // OS must allocate BAR addresses within the host bridge's memory-mapped window.
+    init_mmio_allocator_from_fdt(&pci);
+    init_intx_mapper_from_fdt(&pci);
+
     let addr_start = region.starting_address as usize;
     let addr_end = addr_start.checked_add(region.size.unwrap()).unwrap();
     PCI_ECAM_CFG_SPACE.call_once(|| IoMem::acquire(addr_start..addr_end).unwrap());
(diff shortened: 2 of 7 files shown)


tatetian:
@lrh2000 Could you review this PR?

kernel/comps/pci/src/arch/riscv/mod.rs
```diff
@@ -101,13 +116,227 @@ pub(crate) fn init() -> Option<RangeInclusive<u8>> {
...
+/// A simple MMIO allocator managing a linear region.
+///
+/// RISC-V platforms booted via OpenSBI do not have firmware that performs PCI
+/// enumeration, so the base addresses of PCI device BARs must be allocated by
+/// the OS within the host bridge's memory-mapped window.
+///
+/// This is a bump allocator: addresses are handed out monotonically and never
+/// reclaimed. That is intentional — BAR base addresses are assigned once during
+/// enumeration and are not freed afterwards.
+struct MmioAllocator {
```
lrh2000, line 141: > * Port the LoongArch MMIO bump allocator into the RISC-V PCI arch module (`MmioAllocator` + `alloc_mmio`), initialized from the host bridge's `ranges` DT property.

This is the wrong direction in which to work. We should avoid duplicating code for no good reason. Make the allocator arch-agnostic.

The code in `MemoryBar` should be changed to something like:
```rust
let base = if let Some(mmio_allocator) = MMIO_ALLOCATOR.get() {
    // allocating
} else {
    raw64 & MEMORY_ADDRESS_MASK
}
```

Even in x86, we will need an MMIO allocator as well. Not all BIOSes will initialize all PCI devices, some BIOSes leave some PCI devices uninitialized.
https://github.com/asterinas/asterinas/blob/cee211c4232d8628aa781c2193d462c400dbba81/kernel/comps/pci/src/cfg_space.rs#L450-L462

But the x86 case is extremely complex because the MMIO range is encoded in a complex ACPI table. So it can be left as a TODO. But the correct direction to work with the MMIO allocator is still to make it arch agnostic.

kernel/comps/pci/src/common_device.rs
```diff
@@ -64,6 +66,17 @@ impl PciCommonDevice {
...
 
+    /// Enables legacy PCI INTx interrupts.
+    pub fn enable_intx_interrupt(&self) {
+        self.write_command(self.read_command() & !Command::INTERRUPT_DISABLE)
+    }
+
+    /// Maps the legacy PCI INTx interrupt pin to an IRQ line.
+    #[cfg(target_arch = "riscv64")]
+    pub fn map_intx_interrupt(&self) -> Result<MappedPciIrqLine> {
+        crate::arch::map_intx_interrupt(&self.location)
```
lrh2000, line 77: This seems to be a wrong way to model INTx interrupts. There are a total of four INTx interrupts, but since there can be many PCI devices, the interrupts must be shared among them.

However, mapping two IRQ lines to the same interrupt pin will not work. Therefore, we cannot model INTx interrupts this way.

There can be more reasonable ways to implement this. For example, we could acquire four IRQ lines and map them at `arch::init()`. Then, we should only return `Option<IRQLine>`.

> #[cfg(target_arch = "riscv64")]

Also, please try to make the code architecture-agnostic. The x86 code should simply return `None` because it does not support the INTx pins, which are encoded in the complex ACPI table. However, the API should be available.

kernel/comps/pci/src/arch/riscv/mod.rs
```diff
@@ -101,13 +116,227 @@ pub(crate) fn init() -> Option<RangeInclusive<u8>> {
...
 }
 
+/// A simple MMIO allocator managing a linear region.
+///
+/// RISC-V platforms booted via OpenSBI do not have firmware that performs PCI
+/// enumeration, so the base addresses of PCI device BARs must be allocated by
+/// the OS within the host bridge's memory-mapped window.
+///
+/// This is a bump allocator: addresses are handed out monotonically and never
+/// reclaimed. That is intentional — BAR base addresses are assigned once during
```
lrh2000, line 139: LLM generated comments are odd. Try to avoid this: "That is intentional — BAR base addresses ..." -> "This is because BAR base addresses ..."

kernel/comps/pci/src/cfg_space.rs
```diff
@@ -421,14 +424,20 @@ impl MemoryBar {
         let size = decode_size(size_encoded64, BarKind::Memory);
 
         // Restore the original base address.
-        #[cfg(not(target_arch = "loongarch64"))]
+        #[cfg(not(any(target_arch = "loongarch64", target_arch = "riscv64")))]
         let base = raw64 & MEMORY_ADDRESS_MASK;
-        // In LoongArch, the BAR base address needs to be allocated manually.
-        #[cfg(target_arch = "loongarch64")]
+        // On LoongArch and RISC-V the BAR base address needs to be allocated
```
lrh2000, line 429: In this PR, you're keeping removing the comma after the architecture names, why?
```diff
- In LoongArch, the BAR ...
+ On LoongArch and RISC-V the BAR ...
```

Please do not do this. It harms the readability. Review your changes and revert such changes.

The same also applies to your newly added comments. Please add a comma after the architecture name (i.e., `In RISC-V, `).

kernel/comps/virtio/src/transport/pci/device.rs
```diff
@@ -55,7 +59,68 @@ pub struct VirtioPciModernTransport {
...
+}
+
+enum VirtioPciInterrupt {
+    Msix(VirtioMsixManager),
+    #[cfg(target_arch = "riscv64")]
+    Intx(Arc<RwLock<VirtioPciIntxManager>>),
+}
+
+#[cfg(target_arch = "riscv64")]
+struct VirtioPciIntxManager {
```
lrh2000, line 72: This needs to be moved to a correct location. `VirtioMsixManager` is defined in `kernel/comps/virtio/src/transport/pci/msix.rs`. Create a new file for the INTx manager.

kernel/comps/virtio/src/transport/pci/device.rs
```diff
@@ -312,17 +397,42 @@ impl VirtioPciModernTransport {
         let common_cfg = common_cfg.unwrap();
         let device_cfg = device_cfg.unwrap();
 
-        // TODO: Support interrupt without MSI-X.
```
lrh2000, line 315: What about the legacy path then? In `legacy.rs`:
https://github.com/asterinas/asterinas/blob/a19974df7f48dda7d1f6540350d787f5d7df7860/kernel/comps/virtio/src/transport/pci/legacy.rs#L116-L117

kernel/comps/pci/Cargo.toml
```diff
@@ -15,7 +15,10 @@ ostd.workspace = true
 spin.workspace = true
 
 [target.loongarch64-unknown-none-softfloat.dependencies]
-fdt = { version = "0.1.5", features = ["pretty-printing"] }
+fdt.workspace = true
```
lrh2000, line 18: Please do not mix unrelated style changes here.

Currently, architecture dependencies are **not** workspace ones.
```
kernel/comps/pci/Cargo.toml:fdt = { version = "0.1.5", features = ["pretty-printing"] }
kernel/comps/uart/Cargo.toml:fdt = { version = "0.1.5", features = ["pretty-printing"] }
ostd/Cargo.toml:fdt = { version = "0.1.5", features = ["pretty-printing"] }
ostd/Cargo.toml:fdt = { version = "0.1.5", features = ["pretty-printing"] }
ostd/Cargo.toml:fdt = { version = "0.1.5", features = ["pretty-printing"] }
```

You can propose to change them to workspace ones. But please do not mix the changes here. Other dependencies (e.g., `unwinding`) need to change as well if you propose to do so.

kernel/comps/virtio/src/transport/pci/device.rs
```diff
@@ -312,17 +397,42 @@ impl VirtioPciModernTransport {
...
     }
 }
+
+fn create_interrupt_manager(
+    common_device: &mut PciCommonDevice,
+    #[cfg_attr(not(target_arch = "riscv64"), expect(unused_variables))] isr_cfg: Option<
+        VirtioPciCapabilityData,
+    >,
+) -> Result<VirtioPciInterrupt, VirtioTransportError> {
+    #[cfg(target_arch = "riscv64")]
```
lrh2000, line 421: The changes in `virtio/src/transport` should also be arch-agnostic. The usual way is:
```rust
    // Prefer MSI-X if it is available.
    if let Ok(Some(msix)) = common_device.acquire_msix_capability() {
        return Ok(VirtioPciInterrupt::Msix(VirtioMsixManager::new(msix)));
    }

    // Fall back to INTx.
    if let Some(isr_cfg) = isr_cfg
        && let Ok(mapped_irq) = common_device.map_intx_interrupt()
    {
        common_device.enable_intx_interrupt();
        return Ok(VirtioPciInterrupt::Intx(VirtioPciIntxManager::new(
            mapped_irq, isr_cfg,
        )));
    }
```

Try MSI-X first, if it not available, fallback to INTx. (INTx is less efficient than MSI-X, so it should be a fallback).

On RISC-V, we'd likely want to make `acquire_msix_capability` fail if the MSI-X is unavailable for some reason (e.g., if `IMSIC/APLIC` is absent).

`map_intx_interrupt` will always fail on x86 (before we adding support for it), but it should succeed on device-tree-based architectures (e.g., RISC-V and others).

kernel/comps/virtio/src/transport/pci/device.rs
```diff
@@ -55,7 +59,68 @@ pub struct VirtioPciModernTransport {
...
+            if interrupt_status & 0x01 != 0 {
+                for callback in intx.queue_callbacks.iter() {
+                    callback(trap_frame);
+                }
+            }
+            if interrupt_status & 0x02 != 0 {
+                for callback in intx.cfg_callbacks.iter() {
+                    callback(trap_frame);
+                }
+            }
```
lrh2000, line 110: This implementation is also very similar to `MultiplexIrq` in `kernel/comps/virtio/src/transport/mmio/multiplex.rs`. I think there should be a way to extract the common logic:
```rust
struct MultiplexIrq<T>;

trait IsrAccess {
    // For MMIO code:
    // ```
    //            let interrupt_status = irq.interrupt_status.read_once().unwrap();
    //            // Acknowledging before invoking the callbacks to prevent the loss of unhandled
    //            // interrupts due to race conditions.
    //            irq.interrupt_ack.write_once(&interrupt_status).unwrap();
    // ```
    //
    // For INTx code:
    // ```
    //            let interrupt_status = intx.isr_io_memory.read_once::<u8>(intx.isr_offset).unwrap();
    // ```
    fn read(&self) -> IsrStatus;
}

bitflags! {
    struct IsrStatus {
        // Used buffer notification
        // Configuration change notification
    }
}

impl<T: IsrAccess> {}
```

kernel/comps/pci/src/arch/riscv/mod.rs
```diff
@@ -101,13 +116,227 @@ pub(crate) fn init() -> Option<RangeInclusive<u8>> {
...
+}
+
+/// The default target physical address written into each MSI-X Table entry's
+/// `Message Address` field. A device raises an MSI by writing to this address.
+///
+/// On the QEMU `virt` platform this is `0x2400_0000`, the IMSIC (Incoming MSI
+/// Controller) doorbell base defined by the RISC-V AIA. Per-vector remapping
+/// ([`construct_remappable_msix_address`]) is not yet implemented, so every
+/// vector currently shares this single address.
 pub(crate) const MSIX_DEFAULT_MSG_ADDR: u32 = 0x2400_0000;
```
lrh2000, line 340: We should not hard-code a QEMU address. I believe the address needs to be obtained from the device tree.

Also for now we never configure `IMSIC` and handle interrupts from `IMSIC`, I doubt writing comments like this provides any useful information. It is better to:
```rust
/// The default target physical address written into each MSI-X Table entry's
/// `Message Address` field. A device raises an MSI by writing to this address.
//
// TODO: MSI-X is not currently supported in RISC-V. So this is `None`.
pub(crate) const MSIX_DEFAULT_MSG_ADDR: Option<u32> = None;
```

kernel/comps/pci/src/arch/riscv/mod.rs
```diff
@@ -101,13 +116,227 @@ pub(crate) fn init() -> Option<RangeInclusive<u8>> {
...
+        InterruptSourceInFdt {
+            interrupt_parent: entry.interrupt_parent,
+            interrupt: entry.interrupt,
+        },
+        irq_line,
+    )
+}
+
+/// The default target physical address written into each MSI-X Table entry's
+/// `Message Address` field. A device raises an MSI by writing to this address.
```
lrh2000, line 334: If you add this, please add this to `loongarch/mod.rs` and `x86/mod.rs` as well.

kernel/comps/pci/src/arch/riscv/mod.rs
```diff
@@ -101,13 +116,227 @@ pub(crate) fn init() -> Option<RangeInclusive<u8>> {
...
+
+    let mut entries = Vec::new();
+    // Each `interrupt-map` entry is laid out as:
+    //   child unit address (3 cells) + child interrupt pin (1 cell)
+    //   + interrupt parent phandle (1 cell) + parent interrupt specifier (1 cell)
+    // = 6 cells = 24 bytes.
+    //
+    // This assumes the interrupt parent is a PLIC with `#interrupt-cells = 1`, which
+    // matches ostd's `InterruptSourceInFdt` model (a single interrupt source number).
+    // A controller with multi-cell specifiers (e.g. a GIC) is not supported.
```
lrh2000, line 274:
```suggestion
    // This assumes the interrupt parent is a PLIC with `#interrupt-cells = 1`, which
    // provides a single interrupt source value (`InterruptSourceInFdt::interrupt`).
```

lrh2000 (commented):
Adding support for INTx to the PCI is a great feature addition. Thank you for your efforts.

The biggest problem here is deciding which parts of the code should be arch-agnostic. Consider whether x86 should have similar functionalities in the future. If so, try to make the code arch-agnostic.

Additionally, the INTx API should support multiple PCI devices sharing the same IRQ line.

For more details, see comments below.

PR #3480, opened 2026-06-29
Replace `cfg_if` with `cfg_select`

lrh2000:
There is no need to use the third-party `cfg_if` dependency anymore. Rust 1.95 now has stable built-in support for [`cfg_select`](https://doc.rust-lang.org/beta/std/macro.cfg_select.html).

Example:

```rust
    pub const MACHINE: &str = {
        cfg_if::cfg_if! {
            if #[cfg(target_arch = "x86_64")] {
                "x86_64"
            } else if #[cfg(target_arch = "riscv64")] {
                "riscv64"
            } else if #[cfg(target_arch = "loongarch64")] {
                "loongarch64"
            } else if #[cfg(target_arch = "aarch64")] {
                "aarch64"
            } else {
                compile_error!("unsupported target")
            }
        }
    };
```

can be converted to:

```rust
    pub const MACHINE: &str = cfg_select! {
        target_arch = "x86_64" => "x86_64",
        target_arch = "riscv64" => "riscv64",
        target_arch = "loongarch64" => "loongarch64",
        target_arch = "aarch64" => "aarch64",
        _ => compile_error!("unsupported target"),
    };
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,7 +205,6 @@ dependencies = [
  "atomic-integer-wrapper",
  "bitflags 1.3.2",
  "bitvec",
- "cfg-if",
  "component",
  "const_format",
  "controlled",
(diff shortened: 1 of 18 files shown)


tatetian (approved):

PR #3481, opened 2026-06-29
Fix inode append and memfd seal races

li041:
## Summary

This PR fixes several inode metadata **TOCTOU** issues around `append writes`, `resize`, and `fallocate` which is mainly asociated with `size()`.

The VFS layer no longer resolves `O_APPEND` by reading `path.size()` before entering filesystem-specific write logic. 

Instead, ramfs, ext2, and exfat resolve append offsets while holding their inode/internal locks, so the size check and write operation are serialized by the filesystem implementation.

For ramfs and memfd, this also moves size-dependent seal checks into `RamInode` operations. `MemfdInode` now snapshots seals while holding the memfd seal lock and passes them into the shared ramfs write/resize/fallocate paths. This keeps seal validation and size updates in the same locked inode operation.

 ## Changes

- Resolve `O_APPEND` in filesystem write implementations instead of `InodeHandle`.
- Move ramfs write, resize, and fallocate size checks under inode/page-cache locking.
- Apply memfd seal checks inside the corresponding ramfs operation branches.
- Add regression coverage for memfd fallocate seal behavior.

### DIFF
--- a/kernel/src/device/evdev/file.rs
+++ b/kernel/src/device/evdev/file.rs
@@ -17,7 +17,7 @@ use crate::{
     events::IoEvents,
     fs::{
         file::{PerOpenFileOps, SettableStatusFlags, StatusFlags},
-        vfs::inode::FileOps,
+        vfs::inode::{FileOps, WriteOffset},
     },
     prelude::*,
     process::signal::{PollHandle, Pollable, Pollee},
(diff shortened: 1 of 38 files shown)


cchanging:
For this PR, I think we can introduce an abstraction:
```rust
pub enum WriteOffset {
    Absolute(usize),
    Append,
}
```

and take this to replace the original `offset` parameter, will this make the semantics clearer? 

I don’t have a good solution for the other metadata-related races at the moment. If we can’t refactor it in the short term, I think we can move this PR forward for now.

PR #3482, opened 2026-06-29
Update `birth_at` field type from `Duration` to `Option<Duration>`

co63oc:
Update `birth_at` field type from `Duration` to `Option<Duration>`
Update some comments

Refer to comments
https://github.com/asterinas/asterinas/pull/3394#issuecomment-4829386230

### DIFF
--- a/kernel/src/fs/fs_impls/exfat/inode.rs
+++ b/kernel/src/fs/fs_impls/exfat/inode.rs
@@ -1448,7 +1448,7 @@ impl Inode for ExfatInode {
             gid: Gid::new(inner.fs().mount_option().fs_gid as u32),
             container_dev_id: inner.fs().container_device_id(),
             self_dev_id: None,
-            birth_at: inner.crtime.as_duration().unwrap_or_default(),
+            birth_at: Some(inner.crtime.as_duration().unwrap_or_default()),
         }
     }
 
--- a/kernel/src/syscall/statx.rs
+++ b/kernel/src/syscall/statx.rs
@@ -138,16 +138,18 @@ impl Statx {
             stx_attributes |= STATX_ATTR_MOUNT_ROOT;
         }
 
-        // Build `stx_mask` based on what's supported and what was requested
+        // Build `stx_mask` based on what's supported and what was requested.
+        // TODO: Pass the request `stx_mask` to the filesystem. It can determine which information
+        // to return and omit the information that is not requested and expensive to query.
         let mut stx_mask = StatxMask::STATX_BASIC_STATS.bits() | StatxMask::STATX_MNT_ID.bits();
 
-        // Include `STATX_BTIME` if the birth time is available
-        // TODO: The filesystem supports populating `stx_mask`
-        if !info.birth_at.is_zero() {
+        // Include `STATX_BTIME` if the birth time is available.
+        let stx_btime = if let Some(birth_at) = info.birth_at {
             stx_mask |= StatxMask::STATX_BTIME.bits();
-        }
-
-        let stx_btime = StatxTimestamp::from(info.birth_at);
+            StatxTimestamp::from(birth_at)
+        } else {
+            StatxTimestamp::default()
+        };
 
         Self {
             // FIXME: All zero fields below are dummy implementations that need to be improved in the future.
--- a/kernel/src/fs/fs_impls/ext2/inode/attrs.rs
+++ b/kernel/src/fs/fs_impls/ext2/inode/attrs.rs
@@ -74,7 +74,7 @@ impl Inode {
             gid: Gid::new(inner.gid()),
             container_dev_id,
             self_dev_id,
-            birth_at: Duration::ZERO,
+            birth_at: None,
         }
     }
 
--- a/kernel/src/fs/fs_impls/pseudofs/mod.rs
+++ b/kernel/src/fs/fs_impls/pseudofs/mod.rs
@@ -224,7 +224,7 @@ impl PseudoInode {
             gid,
             container_dev_id: dev_id,
             self_dev_id: None,
-            birth_at: Duration::ZERO,
+            birth_at: None,
         };
 
         PseudoInode {
--- a/kernel/src/fs/fs_impls/ramfs/fs.rs
+++ b/kernel/src/fs/fs_impls/ramfs/fs.rs
@@ -1309,7 +1309,7 @@ impl Inode for RamInode {
             } else {
                 DeviceId::from_encoded_u64(rdev)
             },
-            birth_at: Duration::ZERO,
+            birth_at: None,
         }
     }
 
--- a/kernel/src/fs/fs_impls/virtiofs/inode/metadata.rs
+++ b/kernel/src/fs/fs_impls/virtiofs/inode/metadata.rs
@@ -215,6 +215,6 @@ pub(in crate::fs::fs_impls::virtiofs) fn metadata_from_attr(
         } else {
             DeviceId::from_encoded_u64(attr.rdev() as u64)
         },
-        birth_at: Duration::ZERO,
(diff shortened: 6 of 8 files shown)


co63oc:
@boterinas rerun

kernel/src/syscall/statx.rs
```diff
@@ -138,16 +138,18 @@ impl Statx {
...
+        // Build `stx_mask` based on what's supported and what was requested.
         let mut stx_mask = StatxMask::STATX_BASIC_STATS.bits() | StatxMask::STATX_MNT_ID.bits();
 
-        // Include `STATX_BTIME` if the birth time is available
-        // TODO: The filesystem supports populating `stx_mask`
-        if !info.birth_at.is_zero() {
+        // Include `STATX_BTIME` if the birth time is available.
+        // TODO: Currently, we manually check for available fields here. A better approach would be
+        // to have the filesystem populate `stx_mask` directly in the returned metadata, indicating
+        // which fields are actually supported and valid.
```
lrh2000: > A better approach would be to have the filesystem populate `stx_mask` directly in the returned metadata,

This sounds like a C-style suggestion. Is it really a better approach?

In C, one may write:
```c
struct foo_t {
    int flags; // X_VALID | Y_VALID | Z_VALID
    int x;
    int y;
    int z;
};
```

It naturally corresponds to:
```rust
struct Foo {
    x: Option<u32>,
    y: Option<u32>,
    z: Option<u32>,
}
```

I'm not sure if the first approach is better. One advantage is that it may occupy less space due to a flag containing multiple bits. However, this may be its only advantage. It cannot prevent misuse of `x`, `y`, and `z` when the bit is not set.

Ideally, we should create Rust types that render non-existent states unpresentable.
```rust
// In C, this is a state that does not make any sense.
foo_t { flags: 0, x: 3 }

// In Rust, you cannot write:
Foo { x: None(3) }

// Instead, you can only write:
Foo { x: None }
// or
Foo { x: Some(3) }
```

kernel/src/syscall/statx.rs
```diff
@@ -138,16 +138,18 @@ impl Statx {
...
+        if info.birth_at.is_some() {
             stx_mask |= StatxMask::STATX_BTIME.bits();
         }
 
-        let stx_btime = StatxTimestamp::from(info.birth_at);
+        let stx_btime = StatxTimestamp::from(info.birth_at.unwrap_or(Duration::ZERO));
```
lrh2000:
```rust
let stx_btime = if let Some(birth_at) = info.birth_at {
    stx_mask |= StatxMask::STATX_BTIME.bits();
    StatxTimestamp::from(birth_at)
} else {
    StatxTimestamp::default()
};
```

Avoid checking the same thing twice if possible. We only need to check `info.birth_at.is_some()` once here.

lrh2000 (commented):
Thanks for revising the code!

kernel/src/syscall/statx.rs
cchanging: I think the `TODO` here should probably indicate that, for fields that were not requested, each filesystem should decide on its own whether it needs to return `Some`. In [manual](https://man7.org/linux/man-pages/man2/statx.2.html):

> It should be noted that the kernel may return fields that weren't
       requested and may fail to return fields that were requested

Although I don’t know the exact cases, based on the manual, it seems that some filesystems may return the value whenever they support it (it seems that exfat will do this), while others may return it only when it is explicitly requested. This is likely behavior that the interface will need to support in the future.

kernel/src/syscall/statx.rs
co63oc: 已修改

kernel/src/syscall/statx.rs
co63oc: 现在修改是引用manual内容，因为用其他不确定怎么描述

lrh2000: > It should be noted that the kernel may return fields that weren't
> requested and may fail to return fields that were requested

Yes. This is a valid TODO. it means that the filesystem layer should handle `stx_mask` requested by the userspace. So we should pass what was requested to the filesystem layer.

The man pages also say: 
>       A filesystem may also fill in fields that the caller didn't ask
>       for if it has values for them available and the information is
>       available at no extra cost.  If this happens, the corresponding
>       bits will be set in stx_mask.

So what I suggested in https://github.com/asterinas/asterinas/pull/3394#issuecomment-4829458118 is that:
```rust
        // TODO: Pass what was requested to the filesystem to avoid returning unnecessary information.
```

Or if we want to be a bit more verbose:
```rust
        // TODO: Pass the request `stx_mask` to the filesystem. It can determine which information
        // to return and omit the information that is not requested and expensive to query.
```

Either way, this TODO has nothing to do with `STATX_BTIME`. So I also move it below:
```rust
        // Build `stx_mask` based on what's supported and what was requested
        // TODO: Pass the request `stx_mask` to the filesystem. It can determine which information
        // to return and omit the information that is not requested and expensive to query.
        let mut stx_mask = StatxMask::STATX_BASIC_STATS.bits() | StatxMask::STATX_MNT_ID.bits();
```

kernel/src/fs/fs_impls/exfat/inode.rs
```diff
@@ -1448,7 +1448,7 @@ impl Inode for ExfatInode {
             gid: Gid::new(inner.fs().mount_option().fs_gid as u32),
             container_dev_id: inner.fs().container_device_id(),
             self_dev_id: None,
-            birth_at: inner.crtime.as_duration().unwrap_or_default(),
+            birth_at: inner.crtime.as_duration().ok(),
```
lrh2000: If this is Linux behavior, it's fine ([but I believe it is not](https://elixir.bootlin.com/linux/v7.1.2/source/fs/exfat/file.c#L319)).

Otherwise, I will prefer:
```suggestion
            birth_at: Some(inner.crtime.as_duration().unwrap_or_default()),
```

Currently, we return `Duration::ZERO` as `last_access_at`, `last_modify_at`, ... if the time is not invalid.
```
            last_access_at: inner.atime.as_duration().unwrap_or_default(),
            last_modify_at: inner.mtime.as_duration().unwrap_or_default(),
            last_meta_change_at: inner.ctime.as_duration().unwrap_or_default(),
```

So let's do so for `birth_at`, too. This way, every inode in extfat will have a birth time. If the birth time is not valid, it falls back to the epoch, just like the access time, creation time, etc

kernel/src/syscall/statx.rs
co63oc: 已修改 `let mut stx_mask` 增加注释

co63oc:
@boterinas rerun

kernel/src/syscall/statx.rs
```diff
@@ -138,16 +138,22 @@ impl Statx {
...
         let mut stx_mask = StatxMask::STATX_BASIC_STATS.bits() | StatxMask::STATX_MNT_ID.bits();
 
-        // Include `STATX_BTIME` if the birth time is available
-        // TODO: The filesystem supports populating `stx_mask`
-        if !info.birth_at.is_zero() {
+        // Include `STATX_BTIME` if the birth time is available.
+        // TODO: Reference: <https://man7.org/linux/man-pages/man2/statx.2.html>.
+        // The kernel may return fields not explicitly requested or fail to return
+        // requested ones, depending on what the backing filesystem supports.
+        // Filesystems like exfat may return supported fields regardless of request.
```
lrh2000: So what is TODO here? Still no idea. This is at least not a TODO.
```suggestion
```

lrh2000:
Please squash the two commits... Thanks. Otherwise LGTM.

lrh2000 (approved):

PR #3484, opened 2026-06-29
Add filesystem and block device compatibility for NixOS installation

bet4it:
This PR fills in the Asterinas kernel capabilities needed to install an Asterinas-kernel NixOS system to disk from an Asterinas-kernel NixOS ISO.

If #3472 and this PR are both merged, a follow-up PR can add the remaining NixOS ISO/installer integration to complete the full workflow: booting a NixOS ISO with the Asterinas kernel, installing an Asterinas-kernel NixOS system to a hard disk, and then booting from that installed disk.

Specifically, this PR adds the following kernel-side support:

- Add [hadris](https://github.com/hxyulin/hadris/blob/main/crates/hadris-fat)-based FAT filesystem support.
  This is needed for mounting and accessing the EFI System Partition used by GRUB/UEFI installation.

- Add [hadris](https://github.com/hxyulin/hadris/blob/main/crates/hadris-iso)-based ISO9660 filesystem support.
  This allows the kernel to mount an ISO image as a filesystem, which is useful for boot/install media workflows.

- Add a block I/O adapter shared by the `hadris` filesystem integrations.

- Add block device compatibility ioctls used by Linux userspace tooling:
  - `BLKGETSIZE`
  - `BLKSSZGET`
  - `BLKRRPART`
  - `HDIO_GETGEO`

- Add partition metadata needed by compatibility tooling.
  In particular, partition devices can now report their start sector, which is used by tools such as GRUB to map Linux partition devices back to disk partitions when sysfs does not provide the corresponding information.

- Improve GPT partition parsing for disks with 4 KiB sectors.

- Add `/proc/devices`.
  This exposes registered character and block device classes in a Linux-compatible format, allowing early userspace scripts and installer tooling to discover block device major numbers.

Together, these changes are intended as reusable kernel functionality rather than NixOS-specific installer logic. The actual NixOS ISO and installer changes will be layered on top in a follow-up PR.


lrh2000:
We discussed at https://github.com/asterinas/asterinas/pull/3488#issuecomment-4839941504. This is great work, but the current form is better suited to a tracking issue. The code combines too many unrelated modifications to various subsystems, so it's unlikely that we can proceed. Therefore, I'm closing this.

bet4it:
I understand, and thank you for the feedback. This PR wasn't meant to be merged right away in its current state; my main goal was to gather some early feedback from you.

Specifically, I wanted to ask if introducing the `hadris` library to add support for FAT and ISO9660 filesystems is an acceptable or appropriate direction for the project?

lrh2000:
> Specifically, I wanted to ask if introducing the `hadris` library to add support for FAT and ISO9660 filesystems is an acceptable or appropriate direction for the project?

This is a good question to ask... One of our project goal is to minimize the TCB (Trusted Computing Base) size, so we want to avoid use `unsafe` code outside OSTD (i.e., the kernel and the kernel's dependency).

`hadris` seems to contain some unsafe code, so that will be one concern.
https://github.com/search?q=repo%3Ahxyulin%2Fhadris+%22unsafe%22&type=code&p=1

Otherwise, it seems well-maintained. So, if using the dependency really helps us a lot (I guess so?), this approach can be acceptable.

lrh2000:
cc @tatetian for this question. I don't have a really strong opinion for it after a quick look.

> Specifically, I wanted to ask if introducing the `hadris` library to add support for FAT and ISO9660 filesystems is an acceptable or appropriate direction for the project?

PR #3485, opened 2026-06-29
Move Path methods to inherited-method blocks

WaterWhisperer:
https://github.com/asterinas/asterinas/pull/3413#issuecomment-4830169273

Move `Path::name` to the Dentry inherited-method list and `Path::resize` to the Inode inherited-method block.

### DIFF
--- a/kernel/src/fs/vfs/path/mod.rs
+++ b/kernel/src/fs/vfs/path/mod.rs
@@ -171,11 +171,6 @@ impl Path {
         InodeHandle::new(self.clone(), open_args.access_mode, *status_flags)
     }
 
-    /// Gets the real name of the `Path` from its dentry.
-    pub fn name(&self) -> String {
-        self.dentry.name()
-    }
-
     /// Gets the parent `Path` within the same mount.
     ///
     /// This method returns the parent path within the same filesystem/mount.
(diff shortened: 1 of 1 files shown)


lrh2000:
Shouldn't it be placed at the end of the block "Methods inherited from `Inode`"?

Sorry. I don't really get why the existing code is just that confusing. From my perspective, the existing code suggests:
```rust
// Methods inherited from `dentry`.
#[inherit_methods(from = "self.dentry")]
impl Path {
   fn dentry_method1();
   fn dentry_method2();
   fn dentry_method3();

   fn dentry_method4() {
       do_necessary_checks();
       self.dentry.method4();
   }

   fn dentry_method5() {
       do_necessary_checks();
       self.dentry.method5();
   }
}

// Methods inherited from `Inode`.
impl Path {
   fn inode_method1();
   fn inode_method2();
   fn inode_method3();

   fn inode_method4() {
       do_necessary_checks();
       self.inode().method4();
   }

   fn inode_method5() {
       do_necessary_checks();
       self.inode().method5();
   }
}
```

We will also need to add permission checks for `set_atime`, etc. in the future. This will make the code more extensible.

However, the permissions for `set_atime`, `set_owner`, etc. are more complex than `resize()`. A single `check_permission(MAYWRITE)` is not enough, so it's a separate job. But I think it's a good idea to leave some room here.

lrh2000:
Oh, but the new location of the `resize()` method in this PR makes me aware of `Path::name`. It seems that, it should be moved to the `Dentry` block using the short inheritance format (`fn name(&self);`).

WaterWhisperer:
> Shouldn't it be placed at the end of the block "Methods inherited from `Inode`"?

Thanks, I misunderstood the grouping here. I'll update it.

lrh2000 (approved):

PR #3486, opened 2026-06-29
Support SO_TYPE socket option

volcaxiao:
Part of #3430.

## Summary

This PR adds `getsockopt(SOL_SOCKET, SO_TYPE)` support for:

- TCP/UDP sockets
- Unix stream/datagram/seqpacket sockets
- Netlink raw/datagram sockets

## Validation

Tested with:

```sh
make -C test/initramfs/src/regression check
make kernel
make run_kernel AUTO_TEST=regression LOG_LEVEL=error
```

The network regression tests passed, including the added SO_TYPE coverage in sockoption, sockoption_unix, and rtnl_err.

### DIFF
--- a/kernel/src/net/socket/netlink/common/mod.rs
+++ b/kernel/src/net/socket/netlink/common/mod.rs
@@ -25,7 +25,7 @@ use crate::{
     },
     prelude::*,
     process::signal::{PollHandle, Pollable, Pollee},
-    util::{MultiRead, MultiWrite},
+    util::{MultiRead, MultiWrite, net::SockType},
 };
 
 mod bound;
@@ -34,6 +34,7 @@ mod unbound;
 pub struct NetlinkSocket<P: SupportedNetlinkProtocol> {
     inner: RwMutex<Inner<UnboundNetlink<P>, BoundNetlink<P::Message>>>,
     options: RwLock<OptionSet>,
+    socket_type: SockType,
 
     is_nonblocking: AtomicBool,
     pollee: Pollee,
@@ -57,11 +58,14 @@ impl<P: SupportedNetlinkProtocol> NetlinkSocket<P>
 where
     BoundNetlink<P::Message>: Bound<Endpoint = NetlinkSocketAddr>,
 {
-    pub fn new(is_nonblocking: bool) -> Arc<Self> {
+    pub fn new(is_nonblocking: bool, socket_type: SockType) -> Arc<Self> {
+        debug_assert!(socket_type == SockType::SOCK_RAW || socket_type == SockType::SOCK_DGRAM);
+
         let unbound = UnboundNetlink::new();
         Arc::new(Self {
             inner: RwMutex::new(Inner::Unbound(unbound)),
             options: RwLock::new(OptionSet::new()),
+            socket_type,
             is_nonblocking: AtomicBool::new(is_nonblocking),
             pollee: Pollee::new(),
             pseudo_path: SockFs::new_path(),
@@ -200,7 +204,9 @@ where
         let options = self.options.read();
 
         // Deal with socket-level options
-        options.socket.get_option(option, &*inner)
+        options
+            .socket
+            .get_option(option, &(&*inner, self.socket_type))
 
         // TODO: Deal with netlink-level options
     }
@@ -250,8 +256,15 @@ where
 }
 
 impl<P: SupportedNetlinkProtocol> GetSocketLevelOption
-    for Inner<UnboundNetlink<P>, BoundNetlink<P::Message>>
+    for (
+        &Inner<UnboundNetlink<P>, BoundNetlink<P::Message>>,
+        SockType,
+    )
 {
+    fn socket_type(&self) -> SockType {
+        self.1
+    }
+
     fn is_listening(&self) -> bool {
         false
     }
--- a/kernel/src/net/socket/options/mod.rs
+++ b/kernel/src/net/socket/options/mod.rs
@@ -3,7 +3,7 @@
 use macros::impl_socket_options;
 
 use super::util::LingerOption;
-use crate::{net::socket::unix::CUserCred, prelude::*, process::Gid};
+use crate::{net::socket::unix::CUserCred, prelude::*, process::Gid, util::net::SockType};
 
 pub(in crate::net) mod macros;
 
@@ -16,6 +16,7 @@ pub trait SocketOption: Any + Send + Sync + Debug {
 
 impl_socket_options!(
     pub struct ReuseAddr(bool);
+    pub struct SocketType(SockType);
     pub struct Error(Option<crate::error::Error>);
     pub struct Broadcast(bool);
     pub struct SendBuf(u32);
--- a/kernel/src/net/socket/unix/stream/socket.rs
+++ b/kernel/src/net/socket/unix/stream/socket.rs
@@ -30,7 +30,7 @@ use crate::{
         Gid,
         signal::{PollHandle, Pollable, Pollee},
     },
-    util::{MultiRead, MultiWrite},
+    util::{MultiRead, MultiWrite, net::SockType},
 };
 
 pub struct UnixStreamSocket {
@@ -41,7 +41,7 @@ pub struct UnixStreamSocket {
     pollee: Pollee,
     is_nonblocking: AtomicBool,
 
-    is_seqpacket: bool,
+    socket_type: SockType,
     pseudo_path: Path,
 }
 
@@ -163,22 +163,30 @@ impl OptionSet {
 }
 
 impl UnixStreamSocket {
-    pub fn new(is_nonblocking: bool, is_seqpacket: bool) -> Arc<Self> {
-        Self::new_init(Init::new(), is_nonblocking, is_seqpacket)
+    pub fn new(is_nonblocking: bool, socket_type: SockType) -> Arc<Self> {
+        debug_assert!(
+            socket_type == SockType::SOCK_STREAM || socket_type == SockType::SOCK_SEQPACKET
+        );
+
+        Self::new_init(Init::new(), is_nonblocking, socket_type)
     }
 
-    fn new_init(init: Init, is_nonblocking: bool, is_seqpacket: bool) -> Arc<Self> {
+    fn new_init(init: Init, is_nonblocking: bool, socket_type: SockType) -> Arc<Self> {
         Arc::new(Self {
             state: RwMutex::new(Takeable::new(State::Init(init))),
             options: RwLock::new(OptionSet::new()),
             pollee: Pollee::new(),
             is_nonblocking: AtomicBool::new(is_nonblocking),
-            is_seqpacket,
+            socket_type,
             pseudo_path: SockFs::new_path(),
         })
     }
 
-    pub fn new_pair(is_nonblocking: bool, is_seqpacket: bool) -> (Arc<Self>, Arc<Self>) {
+    pub fn new_pair(is_nonblocking: bool, socket_type: SockType) -> (Arc<Self>, Arc<Self>) {
+        debug_assert!(
+            socket_type == SockType::SOCK_STREAM || socket_type == SockType::SOCK_SEQPACKET
+        );
+
         let cred = SocketCred::<ReadDupOp>::new_current();
 
         let (conn_a, conn_b) = Connected::new_pair(
@@ -190,8 +198,8 @@ impl UnixStreamSocket {
             cred.restrict(),
         );
         (
-            Self::new_connected(conn_a, OptionSet::new(), is_nonblocking, is_seqpacket),
-            Self::new_connected(conn_b, OptionSet::new(), is_nonblocking, is_seqpacket),
+            Self::new_connected(conn_a, OptionSet::new(), is_nonblocking, socket_type),
+            Self::new_connected(conn_b, OptionSet::new(), is_nonblocking, socket_type),
         )
     }
 
@@ -199,7 +207,7 @@ impl UnixStreamSocket {
         connected: Connected,
         options: OptionSet,
         is_nonblocking: bool,
-        is_seqpacket: bool,
+        socket_type: SockType,
     ) -> Arc<Self> {
         let cloned_pollee = connected.cloned_pollee();
         Arc::new(Self {
@@ -207,7 +215,7 @@ impl UnixStreamSocket {
             options: RwLock::new(options),
             pollee: cloned_pollee,
             is_nonblocking: AtomicBool::new(is_nonblocking),
-            is_seqpacket,
+            socket_type,
             pseudo_path: SockFs::new_path(),
         })
     }
@@ -219,7 +227,7 @@ impl UnixStreamSocket {
         _flags: SendRecvFlags,
     ) -> Result<usize> {
         match self.state.read().as_ref() {
-            State::Connected(connected) => connected.try_write(buf, aux_data, self.is_seqpacket),
+            State::Connected(connected) => connected.try_write(buf, aux_data, self.is_seqpacket()),
             State::Init(_) | State::Listen(_) => {
                 return_errno_with_message!(Errno::ENOTCONN, "the socket is not connected")
             }
@@ -232,7 +240,7 @@ impl UnixStreamSocket {
         flags: SendRecvFlags,
     ) -> Result<(usize, Vec<ControlMessage>)> {
         match self.state.read().as_ref() {
-            State::Connected(connected) => connected.try_read(buf, self.is_seqpacket, flags),
+            State::Connected(connected) => connected.try_read(buf, self.is_seqpacket(), flags),
             State::Init(_) | State::Listen(_) => {
                 return_errno_with_message!(Errno::EINVAL, "the socket is not connected")
             }
@@ -269,7 +277,7 @@ impl UnixStreamSocket {
                 init,
                 self.pollee.clone(),
                 &self.options.read(),
-                self.is_seqpacket,
+                self.is_seqpacket(),
             ) {
                 Ok(connected) => connected,
                 Err((err, init)) => return (State::Init(init), Err(err)),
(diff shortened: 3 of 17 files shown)


StevenJiang1110:
I think you can split `SO_TYPE` and timeout support as two separate PRs (for example, keep `SO_TYPE` in this PR, and open a new PR for socket timeout). They are not very related, right? Smaller PRs can be reviewed more easy.

volcaxiao:
> I think you can split `SO_TYPE` and timeout support as two separate PRs (for example, keep `SO_TYPE` in this PR, and open a new PR for socket timeout). They are not very related, right? Smaller PRs can be reviewed more easy.

Thanks for the suggestion. I have split the timeout support out of this PR.
This PR now only keeps `SO_TYPE`, and the socket timeout support is moved to a separate PR: #3494.

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -429,8 +429,10 @@ impl Socket for UnixStreamSocket {
             res => return res,
         }
 
+        let stream_level_option = (state.as_ref(), self.is_seqpacket);
+
         // Deal with socket-level options
-        match options.socket.get_option(option, state.as_ref()) {
```
StevenJiang1110, line 433: Moves line 432 below line 434. It should also be under the comment.

```rust
// Deal with socket-level options
let stream_level_option = (state.as_ref(), self.is_seqpacket);
```

test/initramfs/src/regression/network/rtnl_err.c
```diff
@@ -79,6 +80,24 @@ FN_TEST(getpeername)
...
+	int type = -1;
+	socklen_t type_len = sizeof(type);
+
+	TEST_RES(getsockopt(sk_unbound, SOL_SOCKET, SO_TYPE, &type, &type_len),
+		 type == SOCK_DGRAM && type_len == sizeof(type));
+
+	int sk_raw = TEST_SUCC(
+		socket(PF_NETLINK, SOCK_RAW | SOCK_NONBLOCK, NETLINK_ROUTE));
+	type = -1;
+	type_len = sizeof(type);
```
StevenJiang1110: Please remove lines 93 and 94. They are redundant if getsockopt on line 88 already returns the correct values. In tests, we should avoid excessive defensive code, as it adds unnecessary noise — if an earlier assertion fails, it is perfectly fine to let the test fail at that point directly.

There are likely a number of similar issues elsewhere; please fix them all accordingly.

test/initramfs/src/regression/network/sockoption.c
```diff
@@ -122,6 +122,49 @@ FN_TEST(socket_error)
...
+	TEST_ERRNO(setsockopt(sk_unbound, SOL_SOCKET, SO_TYPE, &type, type_len),
+		   ENOPROTOOPT);
+
+	TEST_RES(getsockopt(sk_unbound, SOL_SOCKET, SO_TYPE, &type, &type_len),
+		 type == SOCK_STREAM && type_len == sizeof(type));
+
+	type = -1;
+	type_len = sizeof(type);
+	TEST_RES(getsockopt(sk_listen, SOL_SOCKET, SO_TYPE, &type, &type_len),
+		 type == SOCK_STREAM && type_len == sizeof(type));
```
StevenJiang1110: There is no need to test on every socket type if they do not behave any differently. For TCP sockets, testing on an unbound socket alone is sufficient.

test/initramfs/src/regression/network/sockoption.c
```diff
@@ -122,6 +122,49 @@ FN_TEST(socket_error)
...
+		 type == SOCK_DGRAM && type_len == sizeof(type));
+
+	int sk_nonblocking =
+		TEST_SUCC(socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0));
+	type = -1;
+	type_len = sizeof(type);
+	TEST_RES(getsockopt(sk_nonblocking, SOL_SOCKET, SO_TYPE, &type,
+			    &type_len),
+		 type == SOCK_STREAM && type_len == sizeof(type));
+	TEST_SUCC(close(sk_nonblocking));
```
StevenJiang1110: Similarly, I don't think we need to test non-blocking UDP sockets separately. It just brings some noise.

kernel/src/net/socket/util/options.rs
```diff
@@ -286,6 +291,9 @@ pub const MIN_RECVBUF: u32 = 2304;
 
 /// A trait used for getting socket level options on actual sockets.
 pub(in crate::net) trait GetSocketLevelOption {
+    /// Returns the type used to create the socket.
```
StevenJiang1110:
```suggestion
    /// Returns the socket type.
```

There is no need to emphasize that this is the type used to create the socket. The socket type is immutable once the socket has been created.

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -429,8 +429,10 @@ impl Socket for UnixStreamSocket {
             res => return res,
         }
 
+        let stream_level_option = (state.as_ref(), self.is_seqpacket);
```
StevenJiang1110: What does `stream_level` mean? There is no socket option level named `stream_level`.

kernel/src/net/socket/netlink/common/mod.rs
```diff
@@ -57,11 +58,12 @@ impl<P: SupportedNetlinkProtocol> NetlinkSocket<P>
 where
     BoundNetlink<P::Message>: Bound<Endpoint = NetlinkSocketAddr>,
 {
-    pub fn new(is_nonblocking: bool) -> Arc<Self> {
+    pub fn new_with_type(is_nonblocking: bool, socket_type: SockType) -> Arc<Self> {
```
StevenJiang1110: Is there any particular reason why we cannot continue using `new` as the method name here?

kernel/src/net/socket/options/mod.rs
```diff
@@ -16,6 +16,7 @@ pub trait SocketOption: Any + Send + Sync + Debug {
 
 impl_socket_options!(
     pub struct ReuseAddr(bool);
+    pub struct SocketType(i32);
```
StevenJiang1110: I think we should use `pub struct SocketType(SockType)` here.

The `kernel/src/net` module maintains internal Rust data structures and is not directly responsible for interacting with user space. We should therefore prefer more type-safe types wherever possible.

StevenJiang1110 (commented):
Thanks for your contribution!

volcaxiao:
Thanks for the review. I’ve pushed the latest update to the branch.

kernel/src/net/socket/netlink/common/mod.rs
```diff
@@ -34,6 +34,7 @@ mod unbound;
 pub struct NetlinkSocket<P: SupportedNetlinkProtocol> {
     inner: RwMutex<Inner<UnboundNetlink<P>, BoundNetlink<P::Message>>>,
     options: RwLock<OptionSet>,
+    socket_type: SockType,
```
StevenJiang1110, line 37: Since you stores `SockType` in netlink socket, for consistency, we can store `SockType` instead of `is_seqpacket` in `UnixStreamSocket`. You can perform this change in a separate commit.

kernel/src/net/socket/netlink/common/mod.rs
```diff
@@ -57,11 +58,12 @@ impl<P: SupportedNetlinkProtocol> NetlinkSocket<P>
 where
     BoundNetlink<P::Message>: Bound<Endpoint = NetlinkSocketAddr>,
 {
-    pub fn new(is_nonblocking: bool) -> Arc<Self> {
+    pub fn new(is_nonblocking: bool, socket_type: SockType) -> Arc<Self> {
         let unbound = UnboundNetlink::new();
```
StevenJiang1110, line 64: Please add a `debug_assert!(socket_type == SockType::SOCK_RAW || socket_type == SockType::SOCK_DGRAM)`. While this invariant is enforced by the syscall module, the net module should also guard against invalid socket types being passed in.

kernel/src/net/socket/netlink/common/mod.rs
```diff
@@ -249,9 +250,11 @@ where
     }
 }
 
-impl<P: SupportedNetlinkProtocol> GetSocketLevelOption
-    for Inner<UnboundNetlink<P>, BoundNetlink<P::Message>>
-{
+impl<P: SupportedNetlinkProtocol> GetSocketLevelOption for NetlinkSocket<P> {
+    fn socket_type(&self) -> SockType {
+        self.socket_type
+    }
```
StevenJiang1110, line 266: Let us keep this trait implementation close to its original form for now. Both handling netlink-level socket options and handling socket-level options would need to hold the `Inner` lock (even if that is not strictly necessary at the moment). We would therefore need to acquire the Inner lock upfront from the start. So let us stick with `&Inner` here for now, even though accepting a `NetlinkSocket` directly might also be possible in the future.

```rust
impl<P: SupportedNetlinkProtocol> GetSocketLevelOption
    for (&Inner<UnboundNetlink<P>, BoundNetlink<P::Message>>, SockType) {}
```

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -538,9 +542,22 @@ fn do_unix_getsockopt(option: &mut dyn SocketOption, state: &State) -> Result<()
     Ok(())
 }
 
-impl GetSocketLevelOption for State {
+struct UnixStreamSocketLevelOption<'a> {
+    state: &'a State,
+    is_seqpacket: bool,
+}
+
+impl GetSocketLevelOption for UnixStreamSocketLevelOption<'_> {
```
StevenJiang1110: Let us directly use the following pattern here:

```rust
impl GetSocketLevelOption for (&State, SockType) {}
```

This is more consistent with the current pattern used in TCP socket

```rust
impl SetSocketLevelOption for (&State, &TcpOptionSet) {
```

kernel/src/util/net/options/socket.rs
```diff
@@ -86,6 +89,25 @@ impl_raw_sock_option_get_only!(AcceptConn);
...
+    }
+
+    fn as_sock_option_mut(&mut self) -> &mut dyn SocketOption {
+        self
+    }
+
+    fn as_sock_option(&self) -> &dyn SocketOption {
+        self
+    }
+}
```
StevenJiang1110: You don't need to manually implement `RawSocketOption` trait. Instead, you can implement `write_to_user` for `SockType` then you can use `impl_raw_sock_option_get_only` macro. Below is an example diff: 

```diff
diff --git a/kernel/src/util/net/options/socket.rs b/kernel/src/util/net/options/socket.rs
index 2c2e5803d..e1727dee0 100644
--- a/kernel/src/util/net/options/socket.rs
+++ b/kernel/src/util/net/options/socket.rs
@@ -75,6 +75,7 @@ pub fn new_socket_option(name: i32) -> Result<Box<dyn RawSocketOption>> {
 }
 
 impl_raw_socket_option!(ReuseAddr);
+impl_raw_sock_option_get_only!(SocketType);
 impl_raw_sock_option_get_only!(Error);
 impl_raw_socket_option!(Broadcast);
 impl_raw_socket_option!(SendBuf);
@@ -89,25 +90,6 @@ impl_raw_sock_option_get_only!(AcceptConn);
 impl_raw_socket_option!(SendBufForce);
 impl_raw_socket_option!(RecvBufForce);
 
-impl RawSocketOption for SocketType {
-    fn read_from_user(&mut self, _addr: Vaddr, _max_len: u32) -> Result<()> {
-        return_errno_with_message!(Errno::ENOPROTOOPT, "the option is getter-only");
-    }
-
-    fn write_to_user(&self, addr: Vaddr, max_len: &mut u32) -> Result<usize> {
-        let output = *self.get().unwrap() as i32;
-        output.write_to_user(addr, *max_len)
-    }
-
-    fn as_sock_option_mut(&mut self) -> &mut dyn SocketOption {
-        self
-    }
-
-    fn as_sock_option(&self) -> &dyn SocketOption {
-        self
-    }
-}
-
 // SO_PEERGROUPS is a read-only option. However, calling setsockopt on SO_PEERGROUPS will return EINVAL
 // instead of ENOPROTOOPT like other options. Therefore, we manually implement `RawSocketOption` for it.

 impl RawSocketOption for PeerGroups {
diff --git a/kernel/src/util/net/options/utils.rs b/kernel/src/util/net/options/utils.rs
index d67e6d60f..5de61aab6 100644
--- a/kernel/src/util/net/options/utils.rs
+++ b/kernel/src/util/net/options/utils.rs
@@ -5,13 +5,11 @@ use core::{num::NonZeroU8, time::Duration};
 use ostd::mm::VmIo;
 
 use crate::{
-    context::current_userspace,
-    net::socket::{
+    context::current_userspace, net::socket::{
         ip::{options::IpTtl, stream_options::CongestionControl},
         unix::CUserCred,
         util::LingerOption,
-    },
-    prelude::*,
+    }, prelude::*, util::net::SockType,
 };
 
 /// Create an object by reading its C counterpart from the user space.
@@ -200,6 +198,12 @@ impl WriteToUser for CongestionControl {
     }
 }
 
+impl WriteToUser for SockType {
+    fn write_to_user(&self, addr: Vaddr, max_len: u32) -> Result<usize> {
+        todo!()
+    }
+}
+
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Pod)]
 struct CLinger {
```

StevenJiang1110 (commented):
Thanks for your revision. Only a few minor comments.

volcaxiao:
Thanks, I have addressed the latest review comments in the updated push:

- `UnixStreamSocket` now stores `SockType` directly and derives the seqpacket behavior from it where needed.
- Netlink sockets now assert the expected `SOCK_RAW`/`SOCK_DGRAM` invariant at construction time.
- Netlink socket-level option handling now acquires the `Inner` lock upfront and uses `(&Inner, SockType)` as the `GetSocketLevelOption` context.
- UNIX stream socket option handling now uses `(&State, SockType)` directly.
- `SocketType` now uses `impl_raw_sock_option_get_only!(SocketType)` with `WriteToUser for SockType`, instead of a manual `RawSocketOption` implementation.

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -164,22 +164,24 @@ impl OptionSet {
 
 impl UnixStreamSocket {
     pub fn new(is_nonblocking: bool, is_seqpacket: bool) -> Arc<Self> {
```
StevenJiang1110: The caller should pass `socket_type` directly, just like netlink.

Similarly, add a `debug_assert` like netlink.

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -164,22 +164,24 @@ impl OptionSet {
...
             options: RwLock::new(OptionSet::new()),
             pollee: Pollee::new(),
             is_nonblocking: AtomicBool::new(is_nonblocking),
-            is_seqpacket,
+            socket_type,
             pseudo_path: SockFs::new_path(),
         })
     }
 
     pub fn new_pair(is_nonblocking: bool, is_seqpacket: bool) -> (Arc<Self>, Arc<Self>) {
```
StevenJiang1110: Similarly, this method should accept `socket_type`. Adds a `debug_assert` for the same reason.

StevenJiang1110 (approved):
LGTM! Thanks for your contribution. Only two minor issues left.

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -163,22 +163,26 @@ impl OptionSet {
 }
 
 impl UnixStreamSocket {
-    pub fn new(is_nonblocking: bool, is_seqpacket: bool) -> Arc<Self> {
-        Self::new_init(Init::new(), is_nonblocking, is_seqpacket)
+    pub fn new(is_nonblocking: bool, socket_type: SockType) -> Arc<Self> {
+        debug_assert_unix_stream_socket_type(socket_type);
```
StevenJiang1110: Please avoid introducing such thin wrapper methods — they serve no meaningful purpose. Keep in mind that when abstracting logic into a standalone function/struct/trait, it should at least provide a clear and useful abstraction that aids code comprehension; otherwise, it is better to leave the code as is without the abstraction.

Simply use `debug_assert` directly here instead of wrapping it in a new function.

I have noticed that this is a recurring issue in AI-assisted PRs I have reviewed recently, so please be mindful of similar patterns going forward.

kernel/src/net/socket/unix/stream/socket.rs
```diff
@@ -190,24 +194,26 @@ impl UnixStreamSocket {
...
     }
 
     pub(super) fn new_connected(
         connected: Connected,
         options: OptionSet,
         is_nonblocking: bool,
-        is_seqpacket: bool,
+        socket_type: SockType,
     ) -> Arc<Self> {
+        debug_assert_unix_stream_socket_type(socket_type);
```
StevenJiang1110: Please remove this `debug_assert`.

`new` and `new_pair` are called from outside the `net` module, so adding a `debug_assert` there to ensure the invariant is not violated by external callers. However, `new_connected` is only called internally within the `net` module — as long as the invariant is upheld in `new` and `new_pair`, it is guaranteed to hold here as well, making the assertion unnecessary.

volcaxiao:
Thanks for pointing this out. I agree that the helper was too thin and did not add a useful abstraction.

I have removed `debug_assert_unix_stream_socket_type` and now use `debug_assert!` directly in `UnixStreamSocket::new` and `UnixStreamSocket::new_pair`. I also removed the redundant assertion from `new_connected`, since the invariant is already enforced at the external construction points.

I will avoid introducing such one-line wrapper helpers in future PRs unless they provide a clear abstraction or encode a reusable invariant.

cqs21 (approved):

PR #3488, opened 2026-06-30
Upgrade NixOS distro to 26.05

bet4it:
This PR upgrades the NixOS distro configuration from 25.05 to 26.05, enabling modern systemd configurations in QEMU. To make NixOS 26.05 bootable and support console logins, this PR implements several required kernel features and ports runtime configurations.

## Key Changes:
1. **Implement new mount API**:
   - Introduces modern Linux mount system calls (`fsopen`, `fsconfig`, `fsmount`, `move_mount`) in the VFS layer.
   - Allows configuring and mounting filesystems in discrete, transactional steps, which is required by newer systemd and container managers.
2. **Support mount propagation type changes**:
   - Parses mount propagation flags (`MS_SHARED`, `MS_SLAVE`, `MS_UNBINDABLE`, `MS_PRIVATE`, `MS_REC`) in the `mount` syscall.
   - Implements VFS layer placeholders to support changing mount propagation types, preventing systemd boot failures during root-mount propagation setup.
3. **Support TTY ioctls for systemd getty**:
   - Implements terminal ioctls (`TCGETS2`/`TCSETS2`) and `CTermios2` struct support in the TTY layer.
   - Required by systemd getty to initialize terminal console settings on `/dev/hvc0`.
4. **Add minimal PR_CAP_AMBIENT support**:
   - Implements minimal support for ambient capabilities (`PR_CAP_AMBIENT`), allowing processes to retain capabilities across `execve`.
   - Prevents systemd from hanging during early initialization due to missing capability support.
5. **Bump NixOS distro to 26.05**:
   - Upgrades pinned NixOS pkgs to 26.05.
   - Configuration changes to launch `getty@hvc0.service` for automatic login on console, bypassing desktop configurations.
   - Imports required upstream systemd backports (`Fallback-from-pidfd_spawn-to-posix_spawn.patch`, etc.).

PR #3492, opened 2026-06-30
Implement fsopen, fsconfig, fsmount, and move_mount

bet4it:
This is part of the groundwork for #3488.

## Summary

Implement the initial Linux fd-based mount API support for `fsopen`, `fsconfig`, `fsmount`, and `move_mount`. This supports configuring mount contexts through discrete syscall steps, which is required by newer systemd and container managers.

### Kernel

- Implement `fsopen`, `fsconfig`, `fsmount`, and `move_mount`.
- Add filesystem configuration contexts and detached mount file descriptors.
- Add `CAP_SYS_ADMIN` checks for fd-based mount API operations.
- Validate `fsconfig` arguments, including NULL key/value cases and duplicate `source` parameters.
- Forward filesystem-specific `fsconfig` options to `FsCreationCtx` instead of hard-coding filesystem option names in the syscall layer.
- Return `EOPNOTSUPP` for unsupported `fsconfig` commands and parameter types.
- Make `fsmount` consume a created filesystem context once, matching Linux `EBUSY` behavior on repeated mounts.
- Support mount attributes such as `MOUNT_ATTR_RDONLY` through per-mount flags.
- Reject moving submounts out of detached mount trees.

### LTP Tests

- Enable `fsopen01`, `fsopen02`, `fsconfig01`, `fsconfig02`, `fsmount01`, and `fsmount02`.
- Keep `fsconfig03` disabled because Asterinas lacks `/proc/sys/kernel/tainted`, which the test uses to check kernel taint status.
- Provide a dedicated LTP block device image and set `LTP_DEV=/dev/vdc` for mount-related LTP coverage.

### gVisor Tests

- Enable `mount_fd_test` in the gVisor syscall test list.
- Cover the upstream `FsOpenTest`, `FsConfigTest`, most `FsMountTest`, and `MoveMountTest` cases added by gVisor.
- Keep `FsMountTest.FsMountSuccess` blocklisted because it requires tmpfs to apply creation options such as `nr_inodes`, which is outside this PR.
- Keep `OpenTreeTest.*` blocklisted because `open_tree` is outside the implemented subset in this PR.

### Regression Tests

- Add regression coverage for basic `fsopen`, `fsconfig`, `fsmount`, and `move_mount` flows.
- Add regression coverage for detached mount movement and empty-path handling.

### SCML Coverage

- Add SCML coverage files for `fsopen`, `fsconfig`, `fsmount`, and `move_mount`.
- Update the filesystem and mount-control syscall coverage README.

### DIFF
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -345,6 +345,7 @@ which are summarized in the table below.
 | 332     | statx                  | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#statx) |
 | 424     | pidfd_send_signal      | ✅             | 💯 |
 | 430     | fsopen                 | ✅             | 💯 |
+| 431     | fsconfig               | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#fsconfig) |
 | 434     | pidfd_open             | ✅             | 💯 |
 | 435     | clone3                 | ✅             | [⚠️](syscall-flag-coverage/process-and-thread-management/#clone-and-clone3) |
 | 436     | close_range            | ✅             | 💯 |
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -344,6 +344,7 @@ which are summarized in the table below.
 | 328     | pwritev2               | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#preadv2-and-pwritev2) |
 | 332     | statx                  | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#statx) |
 | 424     | pidfd_send_signal      | ✅             | 💯 |
+| 430     | fsopen                 | ✅             | 💯 |
 | 434     | pidfd_open             | ✅             | 💯 |
 | 435     | clone3                 | ✅             | [⚠️](syscall-flag-coverage/process-and-thread-management/#clone-and-clone3) |
 | 436     | close_range            | ✅             | 💯 |
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -346,6 +346,7 @@ which are summarized in the table below.
 | 424     | pidfd_send_signal      | ✅             | 💯 |
 | 430     | fsopen                 | ✅             | 💯 |
 | 431     | fsconfig               | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#fsconfig) |
+| 432     | fsmount                | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#fsmount) |
 | 434     | pidfd_open             | ✅             | 💯 |
 | 435     | clone3                 | ✅             | [⚠️](syscall-flag-coverage/process-and-thread-management/#clone-and-clone3) |
 | 436     | close_range            | ✅             | 💯 |
--- a/book/src/kernel/linux-compatibility/README.md
+++ b/book/src/kernel/linux-compatibility/README.md
@@ -344,6 +344,7 @@ which are summarized in the table below.
 | 328     | pwritev2               | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#preadv2-and-pwritev2) |
 | 332     | statx                  | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#statx) |
 | 424     | pidfd_send_signal      | ✅             | 💯 |
+| 429     | move_mount             | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#move_mount) |
 | 430     | fsopen                 | ✅             | 💯 |
 | 431     | fsconfig               | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#fsconfig) |
 | 432     | fsmount                | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#fsmount) |
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
@@ -80,6 +80,23 @@ Unsupported commands:
 For more information,
 see [the man page](https://man7.org/linux/man-pages/man2/fsconfig.2.html).
 
+### `fsmount`
+
+Supported functionality in SCML:
+
+```c
+{{#include fsmount.scml}}
+```
+
+Silently-ignored mount attributes:
+* `MOUNT_ATTR_NOATIME`
+* `MOUNT_ATTR_NODIRATIME`
+* `MOUNT_ATTR_RELATIME`
+* `MOUNT_ATTR_STRICTATIME`
+
+For more information,
+see [the man page](https://man7.org/linux/man-pages/man2/fsmount.2.html).
+
 ## Event notifications
 
 ### `inotify_init` and `inotify_init1`
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
@@ -60,6 +60,8 @@ Silently-ignored flags:
 For more information,
 see [the man page](https://man7.org/linux/man-pages/man2/umount.2.html).
 
+## New mount API
+
 ## Event notifications
 
 ### `inotify_init` and `inotify_init1`
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
@@ -97,6 +97,26 @@ Silently-ignored mount attributes:
 For more information,
 see [the man page](https://man7.org/linux/man-pages/man2/fsmount.2.html).
 
+### `move_mount`
+
+Supported functionality in SCML:
+
+```c
+{{#include move_mount.scml}}
+```
+
+Unsupported flags:
+* `MOVE_MOUNT_F_SYMLINKS`
+* `MOVE_MOUNT_F_AUTOMOUNTS`
+* `MOVE_MOUNT_T_SYMLINKS`
+* `MOVE_MOUNT_T_AUTOMOUNTS`
+* `MOVE_MOUNT_T_EMPTY_PATH`
+* `MOVE_MOUNT_SET_GROUP`
+* `MOVE_MOUNT_BENEATH`
+
+For more information,
+see [the man page](https://man7.org/linux/man-pages/man2/move_mount.2.html).
+
 ## Event notifications
 
 ### `inotify_init` and `inotify_init1`
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
@@ -62,6 +62,24 @@ see [the man page](https://man7.org/linux/man-pages/man2/umount.2.html).
 
 ## New mount API
 
+### `fsconfig`
+
+Supported functionality in SCML:
+
+```c
+{{#include fsconfig.scml}}
+```
+
+Unsupported commands:
+* `FSCONFIG_SET_BINARY`
+* `FSCONFIG_SET_PATH`
+* `FSCONFIG_SET_PATH_EMPTY`
+* `FSCONFIG_SET_FD`
+* `FSCONFIG_CMD_CREATE_EXCL`
+
+For more information,
+see [the man page](https://man7.org/linux/man-pages/man2/fsconfig.2.html).
+
 ## Event notifications
 
 ### `inotify_init` and `inotify_init1`
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsconfig.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsconfig.scml
@@ -1,0 +1,7 @@
+// Configure a filesystem context
+fsconfig(
+    fs_fd,
+    cmd = FSCONFIG_SET_FLAG | FSCONFIG_SET_STRING |
+          FSCONFIG_CMD_CREATE | FSCONFIG_CMD_RECONFIGURE,
+    key, value, aux
+);
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsmount.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsmount.scml
@@ -1,0 +1,7 @@
+mount_attrs = MOUNT_ATTR_RDONLY | MOUNT_ATTR_NOSUID | MOUNT_ATTR_NODEV |
+              MOUNT_ATTR_NOEXEC | MOUNT_ATTR_RELATIME | MOUNT_ATTR_NOATIME |
+              MOUNT_ATTR_STRICTATIME | MOUNT_ATTR_NODIRATIME |
+              MOUNT_ATTR_NOSYMFOLLOW;
+
+// Create a mount object from a filesystem context
+fsmount(fs_fd, flags = FSMOUNT_CLOEXEC, mount_attrs = <mount_attrs>);
--- a/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/move_mount.scml
+++ b/book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/move_mount.scml
@@ -1,0 +1,2 @@
+// Move or attach a mount object to the filesystem
+move_mount(from_dfd, from_path, to_dfd, to_path, flags = MOVE_MOUNT_F_EMPTY_PATH);
--- a/kernel/src/fs/file/fs_config_file.rs
+++ b/kernel/src/fs/file/fs_config_file.rs
@@ -269,7 +269,6 @@ impl DetachedMountFile {
     }
 
     /// Returns the detached mount.
-    #[expect(dead_code)]
     pub fn mount(&self) -> Arc<Mount> {
         self.mount.clone()
     }
--- a/kernel/src/fs/file/fs_config_file.rs
+++ b/kernel/src/fs/file/fs_config_file.rs
@@ -2,13 +2,17 @@
 
 use core::fmt::Display;
 
-use super::{AccessMode, CreationFlags, FileLike};
+use super::{AccessMode, CreationFlags, FileLike, InodeMode};
 use crate::{
     events::IoEvents,
     fs::{
         file::file_table::FdFlags,
         pseudofs::AnonInodeFs,
-        vfs::{path::Path, registry::FsType},
+        vfs::{
+            file_system::{FileSystem, FsFlags},
+            path::Path,
+            registry::{FsCreationCtx, FsType},
+        },
     },
     prelude::*,
     process::signal::{PollHandle, Pollable},
@@ -20,15 +24,126 @@ use crate::{
 /// filesystem is created. Once creation succeeds, further configuration is
 /// rejected unless it is an explicit reconfiguration request.
 pub struct FsConfigFile {
+    context: Mutex<FsConfigContext>,
     pseudo_path: Path,
+}
+
+/// Stores the mutable state of a filesystem configuration context.
+struct FsConfigContext {
+    fs_type: &'static dyn FsType,
+    fs_flags: FsFlags,
+    source: Option<String>,
+    mode: Option<InodeMode>,
+    fs: Option<Arc<dyn FileSystem>>,
+    /// Accumulates filesystem-specific mount options as a comma-separated
+    /// string, so they can be forwarded to `FsCreationCtx`.
+    extra_options: String,
 }
 
 impl FsConfigFile {
     /// Creates a filesystem configuration file for a filesystem type.
-    pub fn new(_fs_type: &'static dyn FsType) -> Self {
+    pub fn new(fs_type: &'static dyn FsType) -> Self {
         Self {
+            context: Mutex::new(FsConfigContext {
+                fs_type,
+                fs_flags: FsFlags::empty(),
+                source: None,
+                mode: None,
+                fs: None,
+                extra_options: String::new(),
+            }),
             pseudo_path: AnonInodeFs::new_path(|_| "anon_inode:[fscontext]".to_string()),
         }
+    }
+
+    /// Sets a boolean filesystem configuration option.
+    pub fn set_flag(&self, key: &str) -> Result<()> {
+        let mut context = self.context.lock();
+        if context.fs.is_some() {
+            return_errno_with_message!(Errno::EBUSY, "the file system has already been created");
+        }
+        match key {
+            "ro" => {
+                context.fs_flags |= FsFlags::RDONLY;
+                Ok(())
+            }
+            _ => {
+                append_option(&mut context.extra_options, key, None);
+                Ok(())
+            }
+        }
+    }
+
+    /// Sets a string filesystem configuration option.
+    pub fn set_string(&self, key: &str, value: &str) -> Result<()> {
+        let mut context = self.context.lock();
+        if context.fs.is_some() {
+            return_errno_with_message!(Errno::EBUSY, "the file system has already been created");
+        }
+        match key {
+            "source" => {
+                if context.source.is_some() {
+                    return_errno_with_message!(Errno::EINVAL, "the source is already specified");
+                }
+                context.source = Some(value.to_string());
+                Ok(())
+            }
+            "mode" => {
+                context.mode = Some(parse_octal_mode(value)?);
+                Ok(())
+            }
+            _ => {
+                append_option(&mut context.extra_options, key, Some(value));
+                Ok(())
+            }
+        }
+    }
+
+    /// Creates the configured filesystem.
+    pub fn create_fs(&self, ctx: &Context) -> Result<()> {
+        let mut context = self.context.lock();
+        if context.fs.is_some() {
+            return_errno_with_message!(Errno::EBUSY, "the file system has already been created");
+        }
+
+        let args_cstr = if context.extra_options.is_empty() {
+            None
+        } else {
+            Some(CString::new(context.extra_options.as_str()).map_err(|_| {
+                Error::with_message(Errno::EINVAL, "mount options contain null byte")
+            })?)
+        };
+        let args_ref = args_cstr.as_deref();
+        let fs_creation_ctx =
+            FsCreationCtx::new(context.source.as_deref(), context.fs_flags, args_ref, ctx);
+        let fs = context.fs_type.create(&fs_creation_ctx)?;
+        if Arc::strong_count(&fs) > 1 {
+            let extant_readonly = fs.flags().contains(FsFlags::RDONLY);
+            let context_readonly = context.fs_flags.contains(FsFlags::RDONLY);
+            if extant_readonly != context_readonly {
+                return_errno_with_message!(
+                    Errno::EBUSY,
+                    "the read-only flag of the extant filesystem does not match"
+                );
+            }
+        } else {
+            if let Some(mode) = context.mode {
+                fs.root_inode().set_mode(mode)?;
+            }
+        }
+        context.fs = Some(fs);
+        Ok(())
+    }
+
+    /// Reconfigures the created filesystem with the current flags.
+    pub fn reconfigure_fs(&self, ctx: &Context) -> Result<()> {
+        let context = self.context.lock();
+        let fs = context
+            .fs
+            .as_ref()
+            .ok_or_else(|| Error::with_message(Errno::EINVAL, "the file system is not created"))?;
+
+        fs.set_fs_flags(context.fs_flags, None, ctx)
     }
 }
 
@@ -71,5 +186,27 @@ impl Display for MountApiFdInfo {
         writeln!(f, "flags:\t0{:o}", flags)?;
         writeln!(f, "mnt_id:\t{}", AnonInodeFs::mount_node().id())?;
         writeln!(f, "ino:\t{}", AnonInodeFs::shared_inode().ino())
+    }
+}
+
+fn parse_octal_mode(value: &str) -> Result<InodeMode> {
+    const MAX_MODE_BITS: u16 = 0o7777;
+
+    let mode = u16::from_str_radix(value, 8)
+        .map_err(|_| Error::with_message(Errno::EINVAL, "invalid octal mode"))?;
+    if mode & !MAX_MODE_BITS != 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid mode bits");
+    }
+    Ok(InodeMode::from_bits_truncate(mode))
+}
+
+fn append_option(options: &mut String, key: &str, value: Option<&str>) {
+    if !options.is_empty() {
+        options.push(',');
+    }
+    options.push_str(key);
+    if let Some(value) = value {
+        options.push('=');
+        options.push_str(value);
     }
 }
--- a/kernel/src/fs/file/fs_config_file.rs
+++ b/kernel/src/fs/file/fs_config_file.rs
@@ -221,6 +221,7 @@ impl Pollable for FsConfigFile {
 
 impl FileLike for FsConfigFile {
     fn access_mode(&self) -> AccessMode {
+        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/fsopen.c#L97>.
         AccessMode::O_RDWR
     }
 
@@ -282,6 +283,7 @@ impl Pollable for DetachedMountFile {
 
 impl FileLike for DetachedMountFile {
     fn access_mode(&self) -> AccessMode {
+        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/namespace.c#L4497>.
         AccessMode::O_RDONLY
     }
 
--- a/kernel/src/fs/file/fs_config_file.rs
+++ b/kernel/src/fs/file/fs_config_file.rs
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use core::fmt::Display;
+
+use super::{AccessMode, CreationFlags, FileLike};
+use crate::{
+    events::IoEvents,
+    fs::{
+        file::file_table::FdFlags,
+        pseudofs::AnonInodeFs,
+        vfs::{path::Path, registry::FsType},
+    },
+    prelude::*,
+    process::signal::{PollHandle, Pollable},
+};
+
+/// Represents a filesystem configuration context opened by `fsopen`.
+///
+/// The file stores configuration supplied through `fsconfig` until the
+/// filesystem is created. Once creation succeeds, further configuration is
+/// rejected unless it is an explicit reconfiguration request.
+pub struct FsConfigFile {
+    pseudo_path: Path,
+}
+
+impl FsConfigFile {
+    /// Creates a filesystem configuration file for a filesystem type.
+    pub fn new(_fs_type: &'static dyn FsType) -> Self {
+        Self {
+            pseudo_path: AnonInodeFs::new_path(|_| "anon_inode:[fscontext]".to_string()),
+        }
+    }
+}
+
+impl Pollable for FsConfigFile {
+    fn poll(&self, _mask: IoEvents, _poller: Option<&mut PollHandle>) -> IoEvents {
+        IoEvents::empty()
+    }
+}
+
+impl FileLike for FsConfigFile {
+    fn access_mode(&self) -> AccessMode {
+        AccessMode::O_RDWR
+    }
+
+    fn path(&self) -> &Path {
+        &self.pseudo_path
+    }
+
+    fn dump_proc_fdinfo(self: Arc<Self>, fd_flags: FdFlags) -> Box<dyn Display> {
+        Box::new(MountApiFdInfo {
+            access_mode: self.access_mode(),
+            fd_flags,
+        })
+    }
+}
+
+struct MountApiFdInfo {
+    access_mode: AccessMode,
+    fd_flags: FdFlags,
+}
+
+impl Display for MountApiFdInfo {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let mut flags = self.access_mode as u32;
+        if self.fd_flags.contains(FdFlags::CLOEXEC) {
+            flags |= CreationFlags::O_CLOEXEC.bits();
+        }
+
+        writeln!(f, "pos:\t{}", 0)?;
+        writeln!(f, "flags:\t0{:o}", flags)?;
+        writeln!(f, "mnt_id:\t{}", AnonInodeFs::mount_node().id())?;
+        writeln!(f, "ino:\t{}", AnonInodeFs::shared_inode().ino())
+    }
+}
--- a/kernel/src/fs/file/fs_config_file.rs
+++ b/kernel/src/fs/file/fs_config_file.rs
@@ -10,7 +10,7 @@ use crate::{
         pseudofs::AnonInodeFs,
         vfs::{
             file_system::{FileSystem, FsFlags},
-            path::Path,
+            path::{Mount, MountNamespace, Path, PerMountFlags},
             registry::{FsCreationCtx, FsType},
         },
     },
@@ -24,102 +24,123 @@ use crate::{
 /// filesystem is created. Once creation succeeds, further configuration is
 /// rejected unless it is an explicit reconfiguration request.
 pub struct FsConfigFile {
-    context: Mutex<FsConfigContext>,
+    fs_type: &'static dyn FsType,
+    state: Mutex<FsConfigState>,
     pseudo_path: Path,
 }
 
-/// Stores the mutable state of a filesystem configuration context.
-struct FsConfigContext {
-    fs_type: &'static dyn FsType,
-    fs_flags: FsFlags,
+enum FsConfigState {
+    Configuring(FsCreationConfig),
+    Created(Option<CreatedFs>),
+}
+
+struct FsCreationConfig {
+    flags: FsFlags,
     source: Option<String>,
     mode: Option<InodeMode>,
-    fs: Option<Arc<dyn FileSystem>>,
     /// Accumulates filesystem-specific mount options as a comma-separated
     /// string, so they can be forwarded to `FsCreationCtx`.
     extra_options: String,
+}
+
+struct CreatedFs {
+    fs: Arc<dyn FileSystem>,
+    flags: FsFlags,
+    source: Option<String>,
 }
 
 impl FsConfigFile {
     /// Creates a filesystem configuration file for a filesystem type.
     pub fn new(fs_type: &'static dyn FsType) -> Self {
         Self {
-            context: Mutex::new(FsConfigContext {
-                fs_type,
-                fs_flags: FsFlags::empty(),
+            fs_type,
+            state: Mutex::new(FsConfigState::Configuring(FsCreationConfig {
+                flags: FsFlags::empty(),
                 source: None,
                 mode: None,
-                fs: None,
                 extra_options: String::new(),
-            }),
+            })),
             pseudo_path: AnonInodeFs::new_path(|_| "anon_inode:[fscontext]".to_string()),
         }
     }
 
-    /// Sets a boolean filesystem configuration option.
+    /// Sets a flag-style filesystem configuration option.
+    ///
+    /// This is only valid while the context is still configuring the filesystem.
+    /// It returns `EBUSY` after the filesystem has been created.
     pub fn set_flag(&self, key: &str) -> Result<()> {
-        let mut context = self.context.lock();
-        if context.fs.is_some() {
+        let mut state = self.state.lock();
+        let FsConfigState::Configuring(config) = &mut *state else {
             return_errno_with_message!(Errno::EBUSY, "the file system has already been created");
-        }
+        };
+
         match key {
             "ro" => {
-                context.fs_flags |= FsFlags::RDONLY;
+                config.flags |= FsFlags::RDONLY;
                 Ok(())
             }
             _ => {
-                append_option(&mut context.extra_options, key, None);
+                append_option(&mut config.extra_options, key, None);
                 Ok(())
             }
         }
     }
 
     /// Sets a string filesystem configuration option.
+    ///
+    /// This is only valid while the context is still configuring the filesystem.
+    /// It returns `EBUSY` after the filesystem has been created. The `source`
+    /// option may only be specified once.
     pub fn set_string(&self, key: &str, value: &str) -> Result<()> {
-        let mut context = self.context.lock();
-        if context.fs.is_some() {
+        let mut state = self.state.lock();
+        let FsConfigState::Configuring(config) = &mut *state else {
             return_errno_with_message!(Errno::EBUSY, "the file system has already been created");
-        }
+        };
+
         match key {
             "source" => {
-                if context.source.is_some() {
+                if config.source.is_some() {
                     return_errno_with_message!(Errno::EINVAL, "the source is already specified");
                 }
-                context.source = Some(value.to_string());
+                config.source = Some(value.to_string());
                 Ok(())
             }
             "mode" => {
-                context.mode = Some(parse_octal_mode(value)?);
+                config.mode = Some(parse_octal_mode(value)?);
                 Ok(())
             }
             _ => {
-                append_option(&mut context.extra_options, key, Some(value));
+                append_option(&mut config.extra_options, key, Some(value));
                 Ok(())
             }
         }
     }
 
     /// Creates the configured filesystem.
+    ///
+    /// This consumes the current creation configuration and moves the context
+    /// into the created state. It returns `EBUSY` if the filesystem has already
+    /// been created.
     pub fn create_fs(&self, ctx: &Context) -> Result<()> {
-        let mut context = self.context.lock();
-        if context.fs.is_some() {
+        let mut state = self.state.lock();
+        let FsConfigState::Configuring(config) = &mut *state else {
             return_errno_with_message!(Errno::EBUSY, "the file system has already been created");
-        }
+        };
 
-        let args_cstr = if context.extra_options.is_empty() {
+        let args_cstr = if config.extra_options.is_empty() {
             None
         } else {
-            Some(CString::new(context.extra_options.as_str()).map_err(|_| {
+            Some(CString::new(config.extra_options.as_str()).map_err(|_| {
                 Error::with_message(Errno::EINVAL, "mount options contain null byte")
             })?)
         };
         let args_ref = args_cstr.as_deref();
         let fs_creation_ctx =
-            FsCreationCtx::new(context.source.as_deref(), context.fs_flags, args_ref, ctx);
-        let fs = context.fs_type.create(&fs_creation_ctx)?;
+            FsCreationCtx::new(config.source.as_deref(), config.flags, args_ref, ctx);
+        let fs = self.fs_type.create(&fs_creation_ctx)?;
         if Arc::strong_count(&fs) > 1 {
             let extant_readonly = fs.flags().contains(FsFlags::RDONLY);
-            let context_readonly = context.fs_flags.contains(FsFlags::RDONLY);
+            let context_readonly = config.flags.contains(FsFlags::RDONLY);
             if extant_readonly != context_readonly {
                 return_errno_with_message!(
                     Errno::EBUSY,
@@ -127,23 +148,68 @@ impl FsConfigFile {
                 );
             }
         } else {
-            if let Some(mode) = context.mode {
+            if let Some(mode) = config.mode {
                 fs.root_inode().set_mode(mode)?;
             }
         }
-        context.fs = Some(fs);
+        let flags = config.flags;
+        let source = config.source.clone();
+        *state = FsConfigState::Created(Some(CreatedFs { fs, flags, source }));
         Ok(())
     }
 
     /// Reconfigures the created filesystem with the current flags.
+    ///
+    /// This is only valid after the filesystem has been created and before it
+    /// has been consumed by `create_detached_mount`. It returns `EINVAL` if the
+    /// filesystem has not been created yet, and `EBUSY` if it has already been
+    /// consumed.
     pub fn reconfigure_fs(&self, ctx: &Context) -> Result<()> {
-        let context = self.context.lock();
-        let fs = context
-            .fs
-            .as_ref()
-            .ok_or_else(|| Error::with_message(Errno::EINVAL, "the file system is not created"))?;
+        let state = self.state.lock();
+        let (fs, flags) = match &*state {
+            FsConfigState::Created(Some(created)) => (&created.fs, created.flags),
+            FsConfigState::Created(None) => {
+                return_errno_with_message!(
+                    Errno::EBUSY,
+                    "the file system has already been mounted"
+                );
+            }
+            FsConfigState::Configuring(_) => {
+                return_errno_with_message!(Errno::EINVAL, "the file system is not created");
+            }
+        };
 
-        fs.set_fs_flags(context.fs_flags, None, ctx)
+        fs.set_fs_flags(flags, None, ctx)
+    }
+
+    /// Creates a detached mount from the created filesystem.
+    ///
+    /// This is only valid after `create_fs` has succeeded and before a
+    /// detached mount has already been created from this file. On success, the
+    /// filesystem is consumed and subsequent calls return `EBUSY`. On failure,
+    /// the filesystem remains available for a later `fsmount`.
+    pub fn create_detached_mount(
+        &self,
+        flags: PerMountFlags,
+        mnt_ns: Weak<MountNamespace>,
+    ) -> Result<Arc<Mount>> {
+        let mut state = self.state.lock();
+        let FsConfigState::Created(created) = &mut *state else {
+            return_errno_with_message!(Errno::EINVAL, "the file system is not created");
+        };
+        let Some(created_fs) = created.as_ref() else {
+            return_errno_with_message!(Errno::EBUSY, "the file system has already been mounted");
+        };
+
+        let detached_mount = Mount::new_detached(
+            created_fs.fs.clone(),
+            flags,
+            mnt_ns,
+            created_fs.source.clone(),
+        )?;
+        *created = None;
+
+        Ok(detached_mount)
     }
 }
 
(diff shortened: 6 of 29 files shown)


kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+        .unwrap()
+        .write()
+        .insert(file, fd_flags);
+    Ok(SyscallReturn::Return(fd.into()))
+}
+
+pub fn sys_fsconfig(
+    fs_fd: RawFileDesc,
+    cmd: u32,
+    key_addr: Vaddr,
```
cchanging: Asterinas usually prefers to place each syscall in a separate module. The syscalls in `mount_api.rs` are somewhat complex, and I think each of them deserves its own `.rs` module.

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+    prelude::*,
+    process::signal::{PollHandle, Pollable},
+};
+
+pub fn sys_fsopen(fs_name_addr: Vaddr, flags: u32, ctx: &Context) -> Result<SyscallReturn> {
+    let flags = FsOpenFlags::from_bits(flags)
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "unknown fsopen flags"))?;
+    let fs_name = ctx
+        .user_space()
+        .read_cstring(fs_name_addr, MAX_FILENAME_LEN)?;
```
cchanging: We’d better first check whether `fs_name_addr` is non-null.

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+        .lookup(&fs_path)?
+        .get_top_path();
+
+    target_path.attach_detached_mount(&detached_mount, ctx)?;
+    Ok(SyscallReturn::Return(0))
+}
+
+fn read_user_str(addr: Vaddr, ctx: &Context) -> Result<CString> {
+    ctx.user_space().read_cstring(addr, MAX_FILENAME_LEN)
+}
```
cchanging: This helper function is too trival.

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+    let flags = MoveMountFlags::from_bits(flags)
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "unknown move_mount flags"))?;
+    if flags != MoveMountFlags::MOVE_MOUNT_F_EMPTY_PATH {
+        return_errno_with_message!(Errno::EINVAL, "unsupported move_mount flags");
+    }
+
+    let from_path = read_user_str(from_path_addr, ctx)?;
+    if !from_path.is_empty() {
+        return_errno_with_message!(Errno::EINVAL, "only empty from_path is supported");
+    }
```
cchanging: So currently, if the flags is empty or `from_path` is empty, we will return `EINVAL`? This is weird. I suggest remove the implementation of `sys_move_mount` in this PR, and  implement a more complete version in another PR.

book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsmount.scml
```diff
@@ -0,0 +1,11 @@
...
+fsmount(fs_fd, flags = 0, mount_attrs = 0);
+fsmount(fs_fd, flags = FSMOUNT_CLOEXEC, mount_attrs = 0);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_RDONLY);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_NOSUID);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_NODEV);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_NOEXEC);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_RELATIME);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_NOATIME);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_STRICTATIME);
+fsmount(fs_fd, flags = 0, mount_attrs = MOUNT_ATTR_NODIRATIME);
```
cchanging: This is too redundant. We can do something like this:

```
predefined_clockid = CLOCK_REALTIME | CLOCK_MONOTONIC | CLOCK_MONOTONIC_RAW |
                     CLOCK_REALTIME_COARSE | CLOCK_MONOTONIC_COARSE | CLOCK_BOOTTIME |
                     CLOCK_PROCESS_CPUTIME_ID | CLOCK_THREAD_CPUTIME_ID;

// Get the time of a clock specified by a static ID
clock_gettime(clockid = <predefined_clockid>, tp);

// Get the time of a clock specified by a dynamic ID
clock_gettime(clockid = <INTEGER>, tp);
```

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+}
+
+struct MountApiFdInfo {
+    fd_flags: FdFlags,
+}
+
+impl Display for MountApiFdInfo {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let mut flags = AccessMode::O_RDONLY as u32;
```
cchanging: `DetachedMountFile` is `O_RDONLY` but `FsContextFile` is `O_RDWR`

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+const FSCONFIG_CMD_RECONFIGURE: u32 = 7;
+
+const MOUNT_ATTR_RDONLY: u32 = 0x0000_0001;
+const MOUNT_ATTR_NOSUID: u32 = 0x0000_0002;
+const MOUNT_ATTR_NODEV: u32 = 0x0000_0004;
+const MOUNT_ATTR_NOEXEC: u32 = 0x0000_0008;
+const MOUNT_ATTR_RELATIME: u32 = 0x0000_0040;
+const MOUNT_ATTR_NOATIME: u32 = 0x0000_0010;
+const MOUNT_ATTR_STRICTATIME: u32 = 0x0000_0020;
+const MOUNT_ATTR_NODIRATIME: u32 = 0x0000_0080;
```
cchanging: Declare an `FsConfigOps` enum and a `MountAttrs` bitflags?

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,517 @@
...
+}
+
+fn read_user_str(addr: Vaddr, ctx: &Context) -> Result<CString> {
+    ctx.user_space().read_cstring(addr, MAX_FILENAME_LEN)
+}
+
+struct FsContextFile {
+    state: Mutex<FsContextState>,
+    pseudo_path: Path,
+}
```
cchanging: I suggest moving the implementation of this abstraction to `fs/file/fs_config_file.rs`, renaming `FsContextFile` to `FsConfigFile`, `FsContextState` to `FsConfigContext`, and adding appropriate documentation to the relevant abstractions and interfaces to describe their semantics. This would further separate the logic layer from the syscall layer and better align with the current code organization.

cchanging (commented):
Thanks for working on support for these new syscalls! I did a somewhat incomplete review, and my first impression is that this PR includes quite a lot of changes, some of which may not necessarily need to be merged all at once. Also, all the current implementation is packed into a single commit. Even if you do want to land support for multiple syscalls in one go, I think the commits should at least be split into finer-grained pieces.

I’d suggest further splitting the commits, and also pulling out any parts that could reasonably be separated into their own PRs. That would make this PR easier to review and merge. Thanks!

bet4it:
Thanks for the review!

My understanding is that these new mount API syscalls are closely related: `fsopen`, `fsconfig`, `fsmount`, and `move_mount` work together to provide a usable end-to-end workflow. So I think keeping the minimal implementation and the necessary tests/docs in one PR is reasonable.

I agree that the single commit is too coarse though. How would you prefer me to split the commits?

kernel/src/syscall/mount_api.rs
bet4it: Thanks for pointing this out. I agree the previous behavior was too narrow.

  I’d prefer to keep `move_mount` in this PR because without the attach step, the other new mount API syscalls are not very useful end-to-end, and this syscall is also needed by #3488.

  I updated the implementation so `flags == 0` is now supported. When `from_path` is non-empty, it is handled through the existing `move_mount_to` path-move logic. When `from_path` is empty, it is accepted only with `MOVE_MOUNT_F_EMPTY_PATH`, matching the empty-path source case used with `fsmount`.

cchanging:
> How would you prefer me to split the commits?

The simplest way is to classify them according to the implementation of the system calls.

bet4it:
Done, split into one commit per syscall. Does this look good? Anything else you'd like me to adjust?

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use super::{AccessMode, CreationFlags, FileLike};
+use crate::{
+    events::IoEvents,
+    fs::{file::file_table::FdFlags, pseudofs::AnonInodeFs, vfs::path::Path},
+    prelude::*,
+    process::signal::{PollHandle, Pollable},
+};
+
+pub(crate) struct FsConfigFile {
```
cchanging: In `kernel` crate, we generally use `pub` instead of `pub(crate)`. In addition, the documents for this abstraction should be put in the commit which introduce this abstraction.

kernel/src/syscall/mount_api/fsopen.rs
```diff
@@ -0,0 +1,51 @@
...
+            "the filesystem is not configured in the kernel"
+        );
+    }
+
+    let file = Arc::new(FsConfigFile::new()) as Arc<dyn FileLike>;
+    let fd_flags = if flags.contains(FsOpenFlags::FSOPEN_CLOEXEC) {
+        FdFlags::CLOEXEC
+    } else {
+        FdFlags::empty()
+    };
```
cchanging: You can implement `From<FsOpenFlags>` for `FdFlags`.

kernel/src/syscall/mount_api/fsopen.rs
```diff
@@ -0,0 +1,51 @@
...
+    let fd = ctx
+        .thread_local
+        .borrow_file_table()
+        .unwrap()
+        .write()
+        .insert(file, fd_flags);
```
cchanging: Should we call `fs::vfs::notify::on_open` here? I am not entirely sure.

kernel/src/syscall/mod.rs
```diff
@@ -87,6 +87,7 @@ mod mkdir;
 mod mknod;
 mod mmap;
 mod mount;
+mod mount_api;
```
cchanging: Currently all syscall implementation files put under the syscall module directly. I think we should follow the current style in this PR to keep consistent.

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -63,3 +188,23 @@ impl Display for MountApiFdInfo {
         writeln!(f, "ino:\t{}", AnonInodeFs::shared_inode().ino())
     }
 }
+
+fn parse_octal_mode(value: &str) -> Result<InodeMode> {
+    let mode = u16::from_str_radix(value, 8)
+        .map_err(|_| Error::with_message(Errno::EINVAL, "invalid octal mode"))?;
+    if mode & !0o7777 != 0 {
```
cchanging: Use an additional variable declaration to avoid use magic number `0o777` directly.

kernel/src/syscall/mount_api/fsmount.rs
```diff
@@ -0,0 +1,113 @@
...
+    let fd = ctx
+        .thread_local
+        .borrow_file_table()
+        .unwrap()
+        .write()
+        .insert(file, fd_flags);
+    Ok(SyscallReturn::Return(fd.into()))
+}
+
+fn mount_attrs_to_per_mount_flags(attrs: u32) -> Result<PerMountFlags> {
```
cchanging: Prefer use `TryFrom` trait.

kernel/src/syscall/mount_api/fsmount.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+    if attrs.contains(MountAttrs::NOSUID) {
+        flags |= PerMountFlags::NOSUID;
+    }
+    if attrs.contains(MountAttrs::NODEV) {
+        flags |= PerMountFlags::NODEV;
+    }
+    if attrs.contains(MountAttrs::NOEXEC) {
+        flags |= PerMountFlags::NOEXEC;
+    }
```
cchanging: Can we use `from_truncate` to simplify these logics?

kernel/src/syscall/mount_api/move_mount.rs
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+        let from_path = from_path.to_string_lossy();
+        MoveMountSource::Path(lookup_path(from_dfd, &from_path, ctx)?)
+    };
+
+    let to_path = ctx
+        .user_space()
+        .read_cstring(to_path_addr, MAX_FILENAME_LEN)?;
+    let to_path = to_path.to_string_lossy();
+    let target_path = lookup_path(to_dfd, &to_path, ctx)?;
```
cchanging: These two `lookup_path` lock the `PathResolver` twice. I think we can remove `lookup_path` method, then obtain the `PathResolver` only once.

kernel/src/syscall/move_mount.rs
```diff
@@ -0,0 +1,83 @@
...
+        .downcast_ref::<DetachedMountFile>()
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "the file is not a detached mount"))?;
+
+    Ok(mount_file.mount())
+}
+
+bitflags! {
+    struct MoveMountFlags: u32 {
+        const MOVE_MOUNT_F_EMPTY_PATH = 0x0000_0004;
+    }
```
cchanging, line 101: We’d better define a complete set of flags to distinguish invalid flags from unsupported flags.

cchanging (commented):
Thanks for the updating! To make the commit history clearer, I suggest splitting the final overwrite fix commit and folding its changes into the corresponding earlier commits.

kernel/src/syscall/mount_api/fsmount.rs
bet4it: Done. I now use `PerMountFlags::from_bits_truncate` for the mount attributes whose bit values match `PerMountFlags` directly: `RDONLY`, `NOSUID`, `NODEV`, and `NOEXEC`.

The remaining attributes still need explicit mapping. Their Linux `MOUNT_ATTR_*` values do not match our internal `PerMountFlags` bit values.

kernel/src/syscall/mount_api/fsopen.rs
bet4it: I don't think we should call `fs::vfs::notify::on_open` here.

`fsopen` and `fsmount` create anonymous API file descriptors, similar to other anon-inode based descriptors such as eventfd, timerfd, signalfd, epoll, and inotify. Those paths currently do not call `notify::on_open` either. In contrast, `sys_openat` calls `on_open` after opening a real filesystem path.

The pseudo path under `anon_inodefs` is only an implementation detail, so emitting an open notification for it would be inconsistent with the current anon-inode behavior and probably not meaningful to userspace. If we want anon-inode file creation to emit fsnotify events, I think that should be considered as a separate VFS-wide policy change.

bet4it:
I have updated the syscall organization as suggested: `fsopen`, `fsconfig`, `fsmount`, and `move_mount` are now implemented as separate syscall modules instead of being grouped under `syscall/mount_api/`.

One thing I am not fully sure about is where to place the small shared helper for the new mount API capability check. Currently it is kept in `syscall/mount_api.rs`:

```rust
pub(super) fn check_mount_api_capability(ctx: &Context) -> Result<()>
```

and the individual syscall implementations call it via `super::mount_api::check_mount_api_capability(ctx)?`.

This means `mod mount_api;` still exists in `syscall/mod.rs`, but it is only a small shared helper module now, not the old grouped syscall implementation directory.

Would you prefer keeping this helper in `mount_api.rs`, renaming it to something more explicit like `mount_api_common.rs`, or moving the helper into one of the syscall files such as `fsopen.rs`?

kernel/src/syscall/fsconfig.rs
```diff
@@ -0,0 +1,119 @@
...
+        .ok_or_else(|| Error::with_message(Errno::EINVAL, "the file is not a fs context"))?;
+
+    match cmd {
+        FsConfigOps::SetFlag => {
+            if key_addr == 0 {
+                return_errno_with_message!(Errno::EINVAL, "fsconfig key is NULL");
+            }
+            if value_addr != 0 || aux != 0 {
+                return_errno_with_message!(Errno::EINVAL, "invalid fsconfig arguments");
+            }
```
cchanging: Some blocks check `key_addr` separately, while others include it in the same boolean expression. I think it would be better to make this consistent, and we could use a helper function to handle these checks.

```rust
fn ensure_valid_args(valid: bool) -> Result<()> {
    if !valid {
        return_errno_with_message!(Errno::EINVAL, "invalid fsconfig arguments");
    }
    Ok(())
}

...

FsConfigOps::SetFd => {
      ensure_valid_args(key_addr != 0 && value_addr == 0 && aux >= 0)?;
      return_errno_with_message!(Errno::EOPNOTSUPP, "fsconfig command not supported");
 }
...
```

kernel/src/syscall/fsconfig.rs
```diff
@@ -0,0 +1,119 @@
...
+enum FsConfigOps {
+    SetFlag,
+    SetString,
+    SetBinary,
+    SetPath,
+    SetPathEmpty,
+    SetFd,
+    Create,
+    Reconfigure,
+}
```
cchanging: Can you add some documents to describe the semantics and behaviors for each operations?

kernel/src/syscall/fsconfig.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl TryFrom<u32> for FsConfigOps {
```
cchanging: You can leverage the trait `TryFromInt`, like:

```rust
/// The type of `Bio`.
#[repr(u8)]
#[derive(Clone, Copy, Debug, PartialEq, TryFromInt)]
pub enum BioType {
    /// Read sectors from the device.
    Read = 0,
    /// Write sectors into the device.
    Write = 1,
    /// Flush the volatile write cache.
    Flush = 2,
    // TODO: Add support for other BIO types, such as discarding sectors.
}
```

kernel/src/syscall/mount_api.rs
```diff
@@ -0,0 +1,13 @@
...
+    prelude::*, process::credentials::capabilities::CapSet, security::lsm::hooks as lsm_hooks,
+};
+
+pub(super) fn check_mount_api_capability(ctx: &Context) -> Result<()> {
+    lsm_hooks::on_capable(lsm_hooks::CapableContext::new(
+        ctx.thread_local.borrow_user_ns().as_ref(),
+        ctx.posix_thread,
+        CapSet::SYS_ADMIN,
+    ))
+}
```
cchanging: Each sub-module in `syscall` module should represent a `syscall` implementation. So I suggest you can move this helper function to one of the mount api syscall module such as `fsmount.rs`, and reference it in the other mount api syscall module.

Besides, I have checked some related manual. In the [manual](https://man7.org/linux/man-pages/man2/fsconfig.2.html) of `fsconfig`, it says:
```
EPERM  The command given by cmd was FSCONFIG_CMD_CREATE,
              FSCONFIG_CMD_CREATE_EXCL, or FSCONFIG_CMD_RECONFIGURE, but
              the calling process does not have the required
              CAP_SYS_ADMIN capability.
```

So it seems that some operations does not need to check `CAP_SYS_ADMIN`?

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,265 @@
...
+            })?)
+        };
+        let args_ref = args_cstr.as_deref();
+        let fs_creation_ctx =
+            FsCreationCtx::new(context.source.as_deref(), context.fs_flags, args_ref, ctx);
+        let fs = context.fs_type.create(&fs_creation_ctx)?;
+        if let Some(mode) = context.mode {
+            fs.root_inode().set_mode(mode)?;
+        }
+        context.fs = Some(fs);
```
cchanging: From the manual:

```
One side-effect of this behavior is that if an
 extant filesystem instance is reused, all parameters
 configured for this filesystem configuration context
 are silently ignored (with the exception of the ro
 and rw flag parameters; if the state of the read-
 only flag in the extant filesystem instance and the
 filesystem configuration context do not match, this
 operation will return EBUSY).
```

So I think we may need to perform additional checks here, such as check the reference count of `fs` to distinguish whether it is an extant filesystem.

kernel/src/syscall/fsmount.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn try_from(value: u32) -> Result<Self> {
+        Self::from_bits(value)
+            .ok_or_else(|| Error::with_message(Errno::EINVAL, "unsupported mount attributes"))
+    }
+}
+
+impl TryFrom<MountAttrs> for PerMountFlags {
+    type Error = Error;
+
+    fn try_from(attrs: MountAttrs) -> Result<Self> {
```
cchanging, line 99: Can you give some reference for this check? I am not sure whether it is correct, since I remember that we can set both of these bits at the same time during `mount`.

kernel/src/syscall/move_mount.rs
```diff
@@ -0,0 +1,101 @@
...
+bitflags! {
+    struct MoveMountFlags: u32 {
+        const MOVE_MOUNT_F_SYMLINKS = 0x0000_0001;
+        const MOVE_MOUNT_F_AUTOMOUNTS = 0x0000_0002;
+        const MOVE_MOUNT_F_EMPTY_PATH = 0x0000_0004;
+        const MOVE_MOUNT_T_SYMLINKS = 0x0000_0010;
+        const MOVE_MOUNT_T_AUTOMOUNTS = 0x0000_0020;
+        const MOVE_MOUNT_T_EMPTY_PATH = 0x0000_0040;
+        const MOVE_MOUNT_SET_GROUP = 0x0000_0100;
+        const MOVE_MOUNT_BENEATH = 0x0000_0200;
```
cchanging: You can use spaces to make the bit flag definitions more aligned.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -375,6 +375,64 @@ impl Path {
...
+    ///
+    /// Returns `ENOTDIR` if one of the detached mount root and destination is a
+    /// directory and the other is not.
+    ///
+    /// Returns `EINVAL` if either path is not in the current mount namespace.
+    ///
+    /// Returns `ELOOP` if the destination path is inside the detached mount
+    /// subtree or the mount tree contains a mount namespace file that would
+    /// create a namespace loop.
+    pub(crate) fn attach_detached_mount(
```
cchanging: Use `pub` instead of `pub(crate)`.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -375,6 +375,64 @@ impl Path {
...
+        &self,
+        detached_mount: &Arc<Mount>,
+        ctx: &Context,
+    ) -> Result<()> {
+        let detached_root = Path::new_fs_root(detached_mount.clone());
+        let source_is_dir = detached_root.type_() == InodeType::Dir;
```
cchanging: Can a detached mount created through this new API end up with a source that is not a directory?

If we can get rid of this check, I think we should find a way to reduce the code duplication between `mount` and `attach_detached_mount`. One option is to encapsulate the differences in their parameters with an enum and merge them into a single function. Another possibility is to have `mount` create a new mount first and then reuse `attach_detached_mount`.

test/initramfs/src/regression/fs/mount/mount_api.c
```diff
@@ -0,0 +1,164 @@
...
+	TEST_SUCC(rmdir(dst));
+}
+END_TEST()
+
+FN_TEST(move_mount_empty_path_nonempty_from_path)
+{
+	const char *src = "/tmp/move_mount_empty_flag_src";
+	const char *dst = "/tmp/move_mount_empty_flag_dst";
+
+	CHECK(unshare(CLONE_NEWNS));
```
cchanging: We prefer use `TEST_*` instead of `CHECK` in `FN_TEST`.

test/initramfs/Makefile
```diff
@@ -29,6 +29,7 @@ INITRAMFS_COMPRESSED := true
 endif
 EXT2_IMAGE := $(BUILD_DIR)/ext2.img
 EXFAT_IMAGE := $(BUILD_DIR)/exfat.img
+LTP_DEV_IMAGE := $(BUILD_DIR)/ltp_dev.img
```
cchanging, line 32: Could you explain why this is necessary here? I think we’d better avoid adding an extra image by default just for testing purposes. I don’t think `LTP_DEV_IMAGE` is necessary.

kernel/src/syscall/fsmount.rs
bet4it: The exclusivity check here follows the Linux kernel's `sys_fsmount` implementation. In the kernel source ([`fs/namespace.c:4499-4510`](https://elixir.bootlin.com/linux/v6.15/source/fs/namespace.c#L4499)), `sys_fsmount` validates atime flags via a `switch` statement:

```c
switch (attr_flags & MOUNT_ATTR__ATIME) {
case MOUNT_ATTR_STRICTATIME:
    break;
case MOUNT_ATTR_NOATIME:
    mnt_flags |= MNT_NOATIME;
    break;
case MOUNT_ATTR_RELATIME:
    mnt_flags |= MNT_RELATIME;
    break;
default:
    return -EINVAL;
}
```

The `default` branch returns `-EINVAL` whenever more than one atime bit is set (since the combined value won't match any single `case`). The same pattern also appears in [`build_mount_kattr`](https://elixir.bootlin.com/linux/v6.15/source/fs/namespace.c#L5163) for `mount_setattr`.

While the legacy `mount(2)` accepts combinations of `MS_*` flags and silently resolves them by kernel priority order, the new mount API (`fsmount`/`mount_setattr`) requires exactly one atime policy to be specified. So the exclusivity check here is correct and consistent with Linux.

kernel/src/fs/vfs/path/mod.rs
bet4it: I have refactored the mount path to eliminate the duplication. The legacy `do_mount` helper has been removed. `Path::mount` is now implemented by creating a detached mount first (`Mount::new_detached`) and then calling `attach_detached_mount`, so all the topology validations (directory type check, namespace ownership, loop detection, etc.) are consolidated in a single code path shared by both old and new mount syscalls.

The source type check (directory vs non-directory) is necessary — Linux's `move_mount` requires the source and destination to be both directories or both non-directories, returning `ENOTDIR` otherwise. This is the same validation that the old `mount` code had, just now unified here.

test/initramfs/Makefile
bet4it: LTP's `tst_device` needs an unmounted block device to format and run tests against. The existing `/dev/vda` (ext2) and `/dev/vdb` (exfat) are already mounted by the init script and used by other tests, so they can't be repurposed. The NVMe device (`nvme0n1.img`) only exists in the q35 configuration — microvm and riscv environments don't have it. A dedicated `ltp_dev.img` ensures LTP has a consistent free block device across all architectures

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -582,6 +635,15 @@ impl Path {
     }
 }
 
+fn is_detached_submount(mount: &Arc<Mount>, mnt_ns: &Arc<MountNamespace>) -> bool {
+    let mut root = mount.clone();
+    while let Some(parent) = root.parent().and_then(|parent| parent.upgrade()) {
+        root = parent;
+    }
+
+    !Arc::ptr_eq(&root, mnt_ns.root()) && !Arc::ptr_eq(mount, &root)
```
cchanging: Can you provide some reference for this logic? I have not found related notes in the manual.

Besides, we can use `mount.id()` to do the comparison to avoid atomic operations.

kernel/src/fs/vfs/path/mod.rs
cchanging: I just found that we can directly use `move_mount_to` and don't need `attach_detached_mount`...

```rust
MoveMountSource::Detached(detached_mount) => {
    target_path.attach_detached_mount(&detached_mount, ctx)?;
}
```
We can write:
```rust
MoveMountSource::Detached(detached_mount) => {
    let detached_root = Path::new_fs_root(detached_mount);
    detached_root.move_mount_to(&target_path, ctx);
}
```

To achieve this, we need to:
1. Add the `source_is_dir != target_is_dir` checks in `move_mount_to`, apparently we missed this check before.
2. Replace the "the root mount can not be moved" check in `move_mount_to` with checking if the current mount is the namespace root.

Does this work? I noticed this issue because when this check (`source_is_dir != target_is_dir`) fails, the `move_mount` syscall and `mount` with `MS_MOVE` return `EINVAL`, whereas a regular `mount` operation returns `ENOTDIR`. So it seems the code reuse relationship is actually fairly clear.

test/initramfs/src/conformance/ltp/run_ltp_test.sh
```diff
@@ -10,6 +10,7 @@ RESULT=0
 # Some test cases require a block device. Select the first one by default.
 export LTP_DEV=${LTP_DEV:-/dev/vda}
```
cchanging: So we can remove this line? 

@cqs21 Could you help check whether the test-related changes are reasonable?

book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsconfig.scml
```diff
@@ -0,0 +1,7 @@
+// Configure a filesystem context
+fsconfig(
+    fs_fd,
+    cmd = FSCONFIG_SET_FLAG | FSCONFIG_SET_STRING |
+          FSCONFIG_CMD_CREATE | FSCONFIG_CMD_RECONFIGURE,
+    key, value, aux
+);
```
cchanging, line 7: @taosue Could you help check whether the changes to SCML files are reasonable?

kernel/src/fs/vfs/path/mod.rs
bet4it: This condition detects a non-root mount inside a detached mount tree: `root != mnt_ns.root()` means the source tree is not the current namespace's main tree, and `mount != root` means the source is not that detached tree's root.

I did not find this in the manual either; it comes from Linux's implementation. `fsmount()` puts the detached mount in an anonymous namespace and sets it as `ns->root`:
https://elixir.bootlin.com/linux/v6.15/source/fs/namespace.c#L4558-L4567

`do_move_mount()` then accepts either an attached mount that passes `check_mnt(old)`, or a parentless mount from an anonymous namespace:
https://elixir.bootlin.com/linux/v6.15/source/fs/namespace.c#L3615-L3649

A non-root mount inside a detached tree has a parent, so it goes through the attached branch and is rejected. This helper models that case.

book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
```diff
@@ -60,6 +60,74 @@ Silently-ignored flags:
...
+
+```c
+{{#include fsmount.scml}}
+```
+
+Partially supported mount attributes:
+* `MOUNT_ATTR_NOATIME` can be set but have no actual effect.
+* `MOUNT_ATTR_NODIRATIME` can be set but have no actual effect.
+* `MOUNT_ATTR_RELATIME` can be set but have no actual effect.
+* `MOUNT_ATTR_STRICTATIME` can be set but have no actual effect.
```
taosue: If these attributes are accepted but have no effect, call them "Silently-ignored mount attributes" rather than "partially supported" — "partially supported" implies some of the functionality works.

book/src/kernel/linux-compatibility/README.md
```diff
@@ -344,6 +344,10 @@ which are summarized in the table below.
 | 328     | pwritev2               | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#preadv2-and-pwritev2) |
 | 332     | statx                  | ✅             | [⚠️](syscall-flag-coverage/file-and-directory-operations/#statx) |
 | 424     | pidfd_send_signal      | ✅             | 💯 |
+| 429     | move_mount             | ✅             | [⚠️](syscall-flag-coverage/file-systems-and-mount-control/#move_mount) |
+| 430     | fsopen                 | ✅             | 💯 |
```
taosue, line 348: If `fsopen` is fully supported, its SCML should go in `fully_covered.scml`.

book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/fsmount.scml
```diff
@@ -0,0 +1,9 @@
+// Create a mount object from a filesystem context
+mount_attrs = MOUNT_ATTR_RDONLY | MOUNT_ATTR_NOSUID | MOUNT_ATTR_NODEV |
+              MOUNT_ATTR_NOEXEC | MOUNT_ATTR_RELATIME | MOUNT_ATTR_NOATIME |
+              MOUNT_ATTR_STRICTATIME | MOUNT_ATTR_NODIRATIME |
+              MOUNT_ATTR_NOSYMFOLLOW;
+
+fsmount(fs_fd, flags = 0, mount_attrs = 0);
+fsmount(fs_fd, flags = FSMOUNT_CLOEXEC, mount_attrs = 0);
+fsmount(fs_fd, flags = 0, mount_attrs = <mount_attrs>);
```
taosue: Each syscall pattern should have a comment describing the functionality it supports, per the SCML convention. Bitflag variables don't need comment unless they carry a special case.

Also, pattern 1 (`fsmount(fs_fd, flags = 0, mount_attrs = 0)`) is a subset of pattern 2 or 3 — unless it represents a common-enough case worth calling out explicitly, I'd suggest dropping it to avoid redundancy.

Ditto for the `fsopen` and `move_mount` SCML files.

kernel/src/syscall/fsmount.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        let mut flags = PerMountFlags::from_bits_truncate(
+            (attrs
+                & (MountAttrs::RDONLY
+                    | MountAttrs::NOSUID
+                    | MountAttrs::NODEV
+                    | MountAttrs::NOEXEC
+                    | MountAttrs::NOSYMFOLLOW))
+                .bits(),
```
cchanging: I think `from_bits_truncate(attrs)` is enough? The bits filtered out by this expression do not have corresponding bits in `PerMountFlags`.

taosue (commented):
Thanks for supporting these syscalls. Here are some comments about SCML.

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -582,6 +598,15 @@ impl Path {
     }
 }
 
+fn is_detached_submount(mount: &Arc<Mount>, mnt_ns: &Arc<MountNamespace>) -> bool {
```
cchanging: Thanks for the explanation. I now realize there is a major issue at the moment: our current mount namespace mechanism for detached mount is not correct.

In the current implementation, because the mount namespace field in `Mount` is immutable, a detached mount is created with the mount namespace of the process that calls `fsmount`. However, this does not seem consistent with Linux behavior. According to the reference you provided, a detached mount should only have its mount namespace updated to that of the target mount when it is actually attached.

If we want to implement this correctly, we would probably need to introduce the semantics of an anonymous mount namespace, as well as make the namespace field in `Mount` mutable. That does add a fair amount of complexity and would further expand the scope of this PR.

So at this point, I think it would be reasonable for this PR to retain some limitations. My view on such limitations is that it is acceptable to temporarily not support operations that are valid and feasible, but we should try not to allow operations that are invalid or infeasible and introduce complicated workaround.

Given that, my current suggestion is:

1. Keep the `new_detached` logic unchanged for now, but add a `FIXME` comment indicating that we should later introduce anonymous mount namespaces and mutability.
2. Do not support moving a mount onto a detached mount tree for now. `move_mount` operation seems to involve a lot of judgment logic—for example, an attached mount should not be allowed to move onto a detached mount tree—and without support for anonymous mount namespaces, these checks would only become more complicated.
3. Based on point 2, we should be able to remove the `is_detached_submount` check for now and simplify the logic, since there will be no detached submounts in this model. After anonymous mount namespaces are introduced, this kind of relatively complicated handling should probably no longer be necessary either.
4. Keep the previous logic in `Path::mount` and decouple it from `move_mount`. Right now, reusing `move_mount` for `mount` introduces many redundant checks, and in the future the `mount` operation should not need to perform mount namespace checks between the source mount and destination mount. Sorry for the noise I caused earlier.

With this approach, the operations we do support should all correspond to behavior that is also correctly supported on Linux, while avoiding both complicated workarounds and support for operations that would actually be invalid on Linux.

kernel/src/fs/file/mod.rs
```diff
@@ -16,6 +17,7 @@ pub use file_attr::{
     status_flags::{AtomicStatusFlags, StatusFlags},
 };
 pub use file_handle::{FileLike, Mappable};
+pub(crate) use fs_config_file::{DetachedMountFile, FsConfigFile};
```
cchanging: -> `pub use`

`pub(crate) use inode_attr::mode::xxx` is special since it exports the macros.

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -262,6 +259,16 @@ impl Mount {
         Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new(), None)
     }
 
+    /// Creates a mount node that is not attached to the mount tree.
+    pub(crate) fn new_detached(
```
cchanging: -> pub fn

tools/qemu_args.sh
```diff
@@ -85,8 +85,10 @@ if [ "$1" = "riscv" ]; then
         -chardev stdio,id=mux,mux=on,signal=off,logfile=qemu.log \
         -drive if=none,format=raw,id=x0,file=./test/initramfs/build/ext2.img \
         -drive if=none,format=raw,id=x1,file=./test/initramfs/build/exfat.img \
+        -drive if=none,format=raw,id=x2,file=./test/initramfs/build/ltp_dev.img \
         -device virtio-blk-device,drive=x1 \
         -device virtio-blk-device,drive=x0 \
+        -device virtio-blk-device,drive=x2 \
```
cqs21: The ordering breaks riscv64 tests. Running e.g. `make run_kernel TARGET_ARCH=riscv64` then fails with `mounting /dev/vdb on /exfat failed: Invalid argument`.

On the riscv `virt` machine (virtio-mmio), the guest enumerates `virtio-blk-device` entries in the **reverse** of command-line order — which is why the pre-existing code deliberately lists `x1` (exfat) before `x0` (ext2) so that after reversal `x0`/ext2 becomes `/dev/vda` and `x1`/exfat becomes `/dev/vdb`. Appending the new `x2` (`ltp_dev.img`) at the end of the device list makes it enumerate *first*, shifting every disk by one: `x2`→`vda`, `x0`/ext2→`vdb`, `x1`/exfat→`vdc`.

kernel/src/syscall/fsmount.rs
bet4it: I don't think using `PerMountFlags::from_bits_truncate(attrs.bits())` directly is correct here.

  Only `MOUNT_ATTR_RDONLY`, `MOUNT_ATTR_NOSUID`, `MOUNT_ATTR_NODEV`, and `MOUNT_ATTR_NOEXEC` have bit values that match `PerMountFlags`. The other mount attributes need explicit translation.
  In particular, `MOUNT_ATTR_NOSYMFOLLOW` is `0x00200000`, which would be interpreted as `PerMountFlags::RELATIME` in Asterinas if we passed all attrs to `from_bits_truncate` directly.
  `NOATIME`, `STRICTATIME`, and `NODIRATIME` also use different bit positions from `PerMountFlags`.

  So I kept `from_bits_truncate` only for the bit-compatible attrs and mapped the remaining attrs explicitly.

book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/move_mount.scml
```diff
@@ -0,0 +1,5 @@
+// Move a mount referenced by a nonempty source path
+move_mount(from_dfd, from_path, to_dfd, to_path, flags = 0);
+
+// Move a mount while allowing an empty source path
+move_mount(from_dfd, from_path, to_dfd, to_path, flags = MOVE_MOUNT_F_EMPTY_PATH);
```
taosue: Splitting on `MOVE_MOUNT_F_EMPTY_PATH` seems unnecessary. Allowing an empty source path is just a flag variation, not a distinct sub-function. I'd keep a single pattern:
```
// Move or attach mount object to filesystem
move_mount(from_dfd, from_path, to_dfd, to_path, flags = MOVE_MOUNT_F_EMPTY_PATH);
```

book/src/kernel/linux-compatibility/syscall-flag-coverage/file-systems-and-mount-control/README.md
```diff
@@ -60,6 +60,63 @@ Silently-ignored flags:
...
+
+```c
+{{#include fsmount.scml}}
+```
+
+Silently-ignored mount attributes:
+* `MOUNT_ATTR_NOATIME` can be set but has no effect.
+* `MOUNT_ATTR_NODIRATIME` can be set but has no effect.
+* `MOUNT_ATTR_RELATIME` can be set but has no effect.
+* `MOUNT_ATTR_STRICTATIME` can be set but has no effect.
```
taosue: "Silently-ignored" already implies the flag can be set but has no effect — no need to repeat it per entry.

taosue (commented):
Thanks for the update.

kernel/src/syscall/fsmount.rs
cchanging: Thanks for the explanation! I missed `MOUNT_ATTR_NOSYMFOLLOW` before. Maybe you could define a const variable here to make the semantic clearer:

```rust
const SHARED_BITS: MountAttrs = MountAttrs::RDONLY | MountAttrs::NOSUID | MountAttrs::NODEV | MountAttrs::NOEXEC;

...attrs & SHARED_BITS...
```

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,276 @@
...
+    fs_type: &'static dyn FsType,
+    fs_flags: FsFlags,
+    source: Option<String>,
+    mode: Option<InodeMode>,
+    fs: Option<Arc<dyn FileSystem>>,
+    mounted: bool,
+    /// Accumulates filesystem-specific mount options as a comma-separated
+    /// string, so they can be forwarded to `FsCreationCtx`.
+    extra_options: String,
+}
```
cchanging, line 44: Rust tends to make the semantics clear directly through definitions. I wonder if we can use the following definitions to replace `FsConfigFile` and `FsConfigContext`:

```rust
pub struct FsConfigFile {
    fs_type: &'static dyn FsType,
    context: Mutex<FsConfigState>,
    pseudo_path: Path,
}

enum FsConfigState {
    Configuring(FsCreationConfig),
    Created(Option<Arc<dyn FileSystem>>),
}

struct FsCreationConfig {
    flags: FsFlags,
    source: Option<String>,
    mode: Option<InodeMode>,
    extra_options: String,
}

```

kernel/src/fs/vfs/path/mount.rs
```diff
@@ -262,6 +262,22 @@ impl Mount {
         Self::new(fs, PerMountFlags::KERNMOUNT, None, Weak::new(), None)
     }
 
+    /// Creates a mount node that is not attached to the mount tree.
+    ///
+    /// FIXME: Linux creates detached mounts in an anonymous mount namespace
+    /// and moves them into the target namespace when they are attached. Asterinas
+    /// currently records the caller's namespace here because `Mount::mnt_ns` is
+    /// immutable. This should be changed once mount namespaces can be updated
+    /// during attach.
```
cchanging:
```suggestion
    /// Creates a mount node that is not attached to the mount tree.
    //
    // FIXME: Linux creates detached mounts in an anonymous mount namespace
    // and moves them into the target namespace when they are attached. Asterinas
    // currently records the caller's namespace here because `Mount::mnt_ns` is
    // immutable. This should be changed once mount namespaces can be updated
    // during attach.
```

cchanging (commented):
Thanks for the revision! I only have a few suggestions regarding the definition of `FsConfigFile` now.

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,304 @@
...
+                flags: FsFlags::empty(),
+                source: None,
+                mode: None,
+                extra_options: String::new(),
+            })),
+            pseudo_path: AnonInodeFs::new_path(|_| "anon_inode:[fscontext]".to_string()),
+        }
+    }
+
+    /// Sets a boolean filesystem configuration option.
```
cchanging: "a boolean"?

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,304 @@
...
+            }
+            _ => {
+                append_option(&mut config.extra_options, key, Some(value));
+                Ok(())
+            }
+        }
+    }
+
+    /// Creates the configured filesystem.
+    pub fn create_fs(&self, ctx: &Context) -> Result<()> {
```
cchanging, line 124: I think the documentation for this file’s public API should be expanded with more details about the invariants—for example, clarifying that it will fail if the file has already been created. That way, through the public API docs, users can better understand how to use the interface and how it relates to state transitions.

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,304 @@
...
+            FsConfigState::Configuring(_) => {
+                return_errno_with_message!(Errno::EINVAL, "the file system is not created");
+            }
+        };
+
+        fs.set_fs_flags(flags, None, ctx)
+    }
+
+    /// Takes the created filesystem for `fsmount`.
+    pub fn take_created_fs(&self) -> Result<(Arc<dyn FileSystem>, FsFlags, Option<String>)> {
```
cchanging: Could we change this API to `create_detached_mount`? That way, the mount can be created in one step, and we wouldn’t need to expose a rollback API like `restore_created_fs` to users.

BTW, `pub(crate) fn restore_created_fs` -> `pub fn restore_created_fs`

kernel/src/fs/vfs/path/mod.rs
```diff
@@ -538,6 +539,43 @@ impl Path {
                 "the destination path is not in this mount namespace"
             );
         }
+        let source_tree_root = mount_tree_root(self.mount_node());
+        let target_tree_root = mount_tree_root(dst_path.mount_node());
```
cchanging, line 545: You could add a `TODO` explaining that once we have an anonymous mount namespace, we won’t need to determine whether it is detached in this way anymore.

Also, without protection from the global lock here, there may be a race condition: it may start out in a detached state, but after obtaining a result like `source_is_detached`, a concurrent operation could `move_mount` it into an attached state. In that case, the subsequent checks could become incorrect.

You may want to consider acquiring the global lock earlier.

cchanging (approved):
Some minor comments, otherwise LGTM!

test/initramfs/src/regression/fs/run_test.sh
```diff
@@ -123,6 +123,7 @@ echo "All mount bind file test passed."
 ./isolation/pivot_root
 
 ./mount/mount_move
+./mount/mount_api
```
lrh2000:
```suggestion
./mount/mount_api
./mount/mount_move
```

test/initramfs/src/regression/fs/mount/mount_api.c
```diff
@@ -0,0 +1,164 @@
...
+#endif
+#ifndef FSCONFIG_SET_STRING
+#define FSCONFIG_SET_STRING 1
+#endif
+#ifndef FSCONFIG_CMD_CREATE
+#define FSCONFIG_CMD_CREATE 6
+#endif
+#ifndef MOVE_MOUNT_F_EMPTY_PATH
+#define MOVE_MOUNT_F_EMPTY_PATH 0x00000004
+#endif
```
lrh2000:
```suggestion
```


For `SYS_*` macros, you should use them from the standard library headers (e.g., `SYS_fsconfig`). I believe that they're available.

For other macros, they're also defined somewhere:
```
> grep -R "FSCONFIG_CMD_CREATE" /usr/include
/usr/include/linux/mount.h:	FSCONFIG_CMD_CREATE	= 6,	/* Create new or reuse existing superblock */
/usr/include/linux/mount.h:	FSCONFIG_CMD_CREATE_EXCL = 8,	/* Create new superblock, fail if reusing existing superblock */
/usr/include/sys/mount.h:  FSCONFIG_CMD_CREATE     = 6,    /* Invoke superblock creation */
/usr/include/sys/mount.h:# define FSCONFIG_CMD_CREATE FSCONFIG_CMD_CREATE
/usr/include/sys/mount.h:  FSCONFIG_CMD_CREATE_EXCL = 8,    /* Create new superblock, fail if reusing existing superblock */
/usr/include/sys/mount.h:# define FSCONFIG_CMD_CREATE_EXCL FSCONFIG_CMD_CREATE_EXCL
```

If posible, use them. If not, give a reason, e.g.,
https://github.com/asterinas/asterinas/blob/bf8e15eff0877954deba68c4bf188f700d501572/test/initramfs/src/regression/process/signal/sigaltstack.c#L16-L18

kernel/src/syscall/fsopen.rs
```diff
@@ -0,0 +1,76 @@
...
+        vfs::registry::look_up,
+    },
+    prelude::*,
+};
+
+pub fn sys_fsopen(fs_name_addr: Vaddr, flags: u32, ctx: &Context) -> Result<SyscallReturn> {
+    if fs_name_addr == 0 {
+        return_errno_with_message!(Errno::EFAULT, "fsopen filesystem name is NULL");
+    }
+
```
lrh2000:
```suggestion
```

The zero address may be a valid address if `mmap_min_addr == 0`. [Linux does not seem to have this check.](https://elixir.bootlin.com/linux/v7.1.2/source/fs/fsopen.c#L120)

kernel/src/syscall/fsconfig.rs
```diff
@@ -0,0 +1,108 @@
...
+            return_errno_with_message!(Errno::EOPNOTSUPP, "fsconfig command not supported");
+        }
+        FsConfigOps::Create => {
+            ensure_valid_args(key_addr == 0 && value_addr == 0 && aux == 0)?;
+            super::fsopen::check_mount_api_capability(ctx)?;
+            fs_config.create_fs(ctx)?;
+        }
+        FsConfigOps::CreateExcl => {
+            ensure_valid_args(key_addr == 0 && value_addr == 0 && aux == 0)?;
+            return_errno_with_message!(Errno::EOPNOTSUPP, "fsconfig command not supported");
```
lrh2000: `fsconfig command not supported` -> `unsupported fsconfig command`

For consistency with your own code:
```rust
        .map_err(|_| Error::with_message(Errno::EOPNOTSUPP, "unsupported fsconfig command"))?;
```

kernel/src/syscall/fsconfig.rs
```diff
@@ -0,0 +1,108 @@
...
+    fs_fd: RawFileDesc,
+    cmd: u32,
+    key_addr: Vaddr,
+    value_addr: Vaddr,
+    aux: i32,
+    ctx: &Context,
+) -> Result<SyscallReturn> {
+    if fs_fd < 0 {
+        return_errno_with_message!(Errno::EINVAL, "invalid fsconfig fd");
+    }
```
lrh2000:
```suggestion
```

This is checked below in `fs_fd.try_into()?`.

**EDIT:** Oh sorry. It seems that you want `EINVAL` instead of `EBADF`. Then you should use `FileDesc::try_from().map_err()` and remove the following `fs_fd.try_info()`. Anyway, there is no need to check the same thing twice.

kernel/src/syscall/fsopen.rs
```diff
@@ -0,0 +1,76 @@
...
+    let flags = FsOpenFlags::try_from(flags)?;
+    check_mount_api_capability(ctx)?;
+
+    let fs_name = ctx
+        .user_space()
+        .read_cstring(fs_name_addr, MAX_FILENAME_LEN)?;
+    let fs_name = fs_name
+        .to_str()
+        .map_err(|_| Error::with_message(Errno::EINVAL, "invalid file system name"))?;
```
lrh2000:
```suggestion
        .map_err(|_| Error::with_message(Errno::EINVAL, "invalid filesystem name"))?;
```

kernel/src/syscall/fsconfig.rs
```diff
@@ -0,0 +1,108 @@
...
+fn ensure_valid_args(valid: bool) -> Result<()> {
+    if !valid {
+        return_errno_with_message!(Errno::EINVAL, "invalid fsconfig arguments");
+    }
+    Ok(())
+}
+
+/// The configuration operations supported by `fsconfig`.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromInt)]
+#[repr(u32)]
```
lrh2000:
```suggestion
#[repr(u32)]
#[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromInt)]
```

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,320 @@
...
+        if self.fd_flags.contains(FdFlags::CLOEXEC) {
+            flags |= CreationFlags::O_CLOEXEC.bits();
+        }
+
+        writeln!(f, "pos:\t{}", 0)?;
+        writeln!(f, "flags:\t0{:o}", flags)?;
+        writeln!(f, "mnt_id:\t{}", AnonInodeFs::mount_node().id())?;
+        writeln!(f, "ino:\t{}", AnonInodeFs::shared_inode().ino())
+    }
+}
```
lrh2000: All other `FdInfo` objects are defined within the `dump_proc_fdinfo` method. Do the same.
```
> grep -R "FdInfo" kernel
kernel/src/events/epoll/file.rs:        struct FdInfo {
kernel/src/events/epoll/file.rs:        impl Display for FdInfo {
kernel/src/events/epoll/file.rs:        Box::new(FdInfo {
kernel/src/fs/file/inode_handle.rs:        struct FdInfo {
kernel/src/fs/file/inode_handle.rs:        impl Display for FdInfo {
kernel/src/fs/file/inode_handle.rs:        Box::new(FdInfo {
kernel/src/fs/vfs/notify/inotify.rs:        struct FdInfo {
kernel/src/fs/vfs/notify/inotify.rs:        impl Display for FdInfo {
kernel/src/fs/vfs/notify/inotify.rs:        Box::new(FdInfo {
kernel/src/net/socket/mod.rs:        struct FdInfo {
kernel/src/net/socket/mod.rs:        impl Display for FdInfo {
kernel/src/net/socket/mod.rs:        Box::new(FdInfo {
kernel/src/process/pid_file.rs:        struct FdInfo {
kernel/src/process/pid_file.rs:        impl Display for FdInfo {
kernel/src/process/pid_file.rs:        Box::new(FdInfo { flags, pid })
kernel/src/syscall/eventfd.rs:        struct FdInfo {
kernel/src/syscall/eventfd.rs:        impl Display for FdInfo {
kernel/src/syscall/eventfd.rs:        Box::new(FdInfo {
kernel/src/syscall/signalfd.rs:        struct FdInfo {
kernel/src/syscall/signalfd.rs:        impl Display for FdInfo {
kernel/src/syscall/signalfd.rs:        Box::new(FdInfo {
kernel/src/time/timerfd.rs:        struct FdInfo {
kernel/src/time/timerfd.rs:        impl Display for FdInfo {
kernel/src/time/timerfd.rs:        Box::new(FdInfo {
```

kernel/src/syscall/fsopen.rs
```diff
@@ -0,0 +1,76 @@
...
+    let fd = ctx
+        .thread_local
+        .borrow_file_table()
+        .unwrap()
+        .write()
+        .insert(file, FdFlags::from(flags));
+    Ok(SyscallReturn::Return(fd.into()))
+}
+
+pub(crate) fn check_mount_api_capability(ctx: &Context) -> Result<()> {
```
lrh2000:
```suggestion
pub(super) fn check_mount_api_capability(ctx: &Context) -> Result<()> {
```

lrh2000 (commented):
Obviously, I don't have time to check the details and I also don't plan to check them.

So the following comments are just some obvious (and minor) problems...

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,339 @@
...
+
+impl Pollable for FsConfigFile {
+    fn poll(&self, _mask: IoEvents, _poller: Option<&mut PollHandle>) -> IoEvents {
+        IoEvents::empty()
+    }
+}
+
+impl FileLike for FsConfigFile {
+    fn access_mode(&self) -> AccessMode {
+        AccessMode::O_RDWR
```
lrh2000, line 225:
```suggestion
        // Reference: <https://elixir.bootlin.com/linux/v7.1.2/source/fs/fsopen.c#L97>.
        AccessMode::O_RDWR
```

It seems that we should have a reference link here:
```
> grep -R "O_RDONLY" kernel -B1 | grep 'Ref'
kernel/src/fs/vfs/notify/inotify.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/notify/inotify/inotify_user.c#L711>.
> grep -R "O_RDWR" kernel -B1 | grep 'Ref'
kernel/src/events/epoll/file.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/eventpoll.c#L2191>.
kernel/src/net/socket/mod.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/net/socket.c#L483>.
kernel/src/process/pid_file.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L1898>.
kernel/src/syscall/eventfd.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/eventfd.c#L401>.
kernel/src/syscall/signalfd.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/signalfd.c#L276>.
kernel/src/time/timerfd.rs-        // Reference: <https://elixir.bootlin.com/linux/v7.0/source/fs/timerfd.c#L436>.
```

kernel/src/fs/file/fs_config_file.rs
```diff
@@ -0,0 +1,339 @@
...
+
+impl Pollable for DetachedMountFile {
+    fn poll(&self, _mask: IoEvents, _poller: Option<&mut PollHandle>) -> IoEvents {
+        IoEvents::empty()
+    }
+}
+
+impl FileLike for DetachedMountFile {
+    fn access_mode(&self) -> AccessMode {
+        AccessMode::O_RDONLY
```
lrh2000, line 287: Same here?
```suggestion
        // Reference: <https://elixir.bootlin.com/linux/v7.1.2/source/fs/namespace.c#L4530>.
        AccessMode::O_RDONLY
````

StevenJiang1110 (approved):

taosue (approved):

cqs21 (approved):

junyang-zh (approved):
LGTM for my part. Thanks for contribution.

PR #2320, opened 2025-08-08
Add RISC-V FPU support

jellllly420:
This PR adds RISC-V FPU support.

RISC-V's floating-point registers and instructions are gated by different extensions (F/D/Q). Different extensions provides FP registers of different sizes and different FP instructions. We implement the final `FpuContext` which we exposed to the users as a enum and choose variants based on runtime extension detection.
```rust
enum FpuContext {
    F(FFpuContext),
    D(DFpuContext),
    Q(QFpuContext),
}

struct FFpuContext { ... }

struct DFpuContext { ... }

struct QFpuContext { ... }
```

This PR respects `aster_nix::process::posix_thread::thread_local::ThreadFpu`'s FPU management strategy though RISC-V actually provides opportunities for a lazy strategy with hardware's help. I'd like to leave this as future work.

PR #3493, opened 2026-06-30
Add RISC-V FPU regression test

le-monde-bleu:
## Summary

- Add a RISC-V FPU regression test for user FP state preservation across syscall, signal delivery, and fork-like `clone`.
- Build and run the test only on `riscv64` regression environments.
- This adds regression coverage for the RISC-V FPU context work related to #2320.

## Background

Asterinas now saves and restores RISC-V user FPU context through the supplementary user context path. The most important runtime paths to cover are syscall/trap return, signal delivery with `rt_sigreturn`, and process creation. Without a regression test, these paths can accidentally lose, leak, or fail to inherit FP registers and `fcsr`.

## Changes

- Add `test/initramfs/src/regression/process/signal/riscv_fpu.c`.
  - Directly reads and writes RISC-V D-extension FP registers and `fcsr`.
  - Verifies that a raw syscall preserves the interrupted FP state.
  - Verifies that a signal handler starts with reset FP state and that `rt_sigreturn` restores the interrupted FP state.
  - Verifies that a fork-like `clone` child inherits the expected FP state and that the parent FP state remains unchanged.
- Update `test/initramfs/src/regression/process/signal/Makefile` to build `riscv_fpu` only for `riscv64-linux`.
- Update `test/initramfs/src/regression/process/run_test.sh` to run the new test only when `uname -m` is `riscv64`.

## Implementation Notes

The test uses small assembly helpers instead of C floating-point expressions so that it can control the exact FP registers being tested. It also uses raw `ecall` syscall wrappers to avoid libc wrappers clobbering caller-saved FP registers while the test is checking syscall-boundary behavior.

### DIFF
--- a/test/initramfs/src/regression/process/signal/Makefile
+++ b/test/initramfs/src/regression/process/signal/Makefile
@@ -10,4 +10,9 @@ C_OBJS_FILTER += \
 	sigtrap
 endif
 
+ifneq ($(HOST_PLATFORM), riscv64-linux)
+C_OBJS_FILTER += \
+	riscv_fpu
+endif
+
 include ../../common/Makefile
--- a/test/initramfs/src/regression/process/signal/riscv_fpu.c
+++ b/test/initramfs/src/regression/process/signal/riscv_fpu.c
@@ -1,0 +1,293 @@
+// SPDX-License-Identifier: MPL-2.0
+
+#include <signal.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/syscall.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#if defined(__riscv) && __riscv_xlen == 64 && defined(__riscv_flen) && \
+	__riscv_flen >= 64
+
+#define FP_REG_NR 20
+#define FCSR_MASK 0xffU
+
+// Use caller-saved FP registers so the asm helpers can freely modify them.
+static const uint64_t initial_regs[FP_REG_NR] = {
+	0x0102030405060708ULL, 0x1112131415161718ULL, 0x2122232425262728ULL,
+	0x3132333435363738ULL, 0x4142434445464748ULL, 0x5152535455565758ULL,
+	0x6162636465666768ULL, 0x7172737475767778ULL, 0x8182838485868788ULL,
+	0x9192939495969798ULL, 0xa1a2a3a4a5a6a7a8ULL, 0xb1b2b3b4b5b6b7b8ULL,
+	0xc1c2c3c4c5c6c7c8ULL, 0xd1d2d3d4d5d6d7d8ULL, 0xe1e2e3e4e5e6e7e8ULL,
+	0xf1f2f3f4f5f6f7f8ULL, 0x0badf00d12345678ULL, 0x13579bdf2468ace0ULL,
+	0xfeedfacecafebeefULL, 0xdeadbeef0badc0deULL,
+};
+
+static const uint64_t handler_regs[FP_REG_NR] = {
+	0x8877665544332211ULL, 0x7867564534231201ULL, 0x1827364554637281ULL,
+	0x2837465564738291ULL, 0x38475665748392a1ULL, 0x485766758493a2b1ULL,
+	0x5867768594a3b2c1ULL, 0x68778695a4b3c2d1ULL, 0x788796a5b4c3d2e1ULL,
+	0x8897a6b5c4d3e2f1ULL, 0x98a7b6c5d4e3f201ULL, 0xa8b7c6d5e4f30211ULL,
+	0xb8c7d6e5f4031221ULL, 0xc8d7e6f504132231ULL, 0xd8e7f60514233241ULL,
+	0xe8f7061524334251ULL, 0x123456789abcdef0ULL, 0x0fedcba987654321ULL,
+	0x55aa55aa55aa55aaULL, 0xaa55aa55aa55aa55ULL,
+};
+
+static const uint64_t zero_regs[FP_REG_NR] = {};
+
+static volatile sig_atomic_t handler_ok;
+
+extern void load_fpu_state(const uint64_t *regs, unsigned long fcsr);
+extern void read_fpu_state(uint64_t *regs, uint32_t *fcsr);
+
+// Read and write FP registers and fcsr directly.
+__asm__(".text\n"
+	".option push\n"
+	".option arch, +d\n"
+	".globl load_fpu_state\n"
+	".type load_fpu_state, @function\n"
+	"load_fpu_state:\n"
+	"	fld ft0, 0(a0)\n"
+	"	fld ft1, 8(a0)\n"
+	"	fld ft2, 16(a0)\n"
+	"	fld ft3, 24(a0)\n"
+	"	fld ft4, 32(a0)\n"
+	"	fld ft5, 40(a0)\n"
+	"	fld ft6, 48(a0)\n"
+	"	fld ft7, 56(a0)\n"
+	"	fld fa0, 64(a0)\n"
+	"	fld fa1, 72(a0)\n"
+	"	fld fa2, 80(a0)\n"
+	"	fld fa3, 88(a0)\n"
+	"	fld fa4, 96(a0)\n"
+	"	fld fa5, 104(a0)\n"
+	"	fld fa6, 112(a0)\n"
+	"	fld fa7, 120(a0)\n"
+	"	fld ft8, 128(a0)\n"
+	"	fld ft9, 136(a0)\n"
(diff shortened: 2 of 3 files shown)


le-monde-bleu:
@cqs21 Could you please review this PR? Thanks!

test/initramfs/src/regression/process/signal/riscv_fpu.c
```diff
@@ -0,0 +1,293 @@
...
+
+static const uint64_t zero_regs[FP_REG_NR] = {};
+
+static volatile sig_atomic_t handler_ok;
+
+extern void load_fpu_state(const uint64_t *regs, unsigned long fcsr);
+extern void read_fpu_state(uint64_t *regs, uint32_t *fcsr);
+
+// Read and write FP registers and fcsr directly.
+__asm__(".text\n"
```
tatetian, line 46: Assembly code of this amount should be kept in a `.S` file for readability and maintainability.

test/initramfs/src/regression/process/signal/riscv_fpu.c
```diff
@@ -0,0 +1,293 @@
...
+	"	frcsr t0\n"
+	"	sw t0, 0(a1)\n"
+	"	ret\n"
+	".size read_fpu_state, .-read_fpu_state\n"
+	".option pop\n");
+
+extern long raw_syscall0(long n);
+extern long raw_syscall1(long n, long arg0);
+extern long raw_syscall2(long n, long arg0, long arg1);
+extern long raw_syscall5(long n, long arg0, long arg1, long arg2, long arg3,
```
tatetian, line 107: I believe that libc has raw syscall wrappers. Let's avoid reinventing the wheel.

test/initramfs/src/regression/process/signal/riscv_fpu.c
```diff
@@ -0,0 +1,293 @@
...
+			"child did not inherit the expected FPU state\n");
+		exit(EXIT_FAILURE);
+	}
+}
+
+int main(void)
+{
+	test_raw_syscall_preserves_fpu();
+	test_signal_preserves_interrupted_fpu();
+	test_clone_inherits_fpu();
```
tatetian, line 279: Please follow the convention of other regression tests, which use our in-house test framework.

tatetian (commented):
Thanks for adding the new test file. This PR should be rebase to the latest main branch.

test/initramfs/src/regression/process/signal/riscv_fpu.c
le-monde-bleu: Thank you for your revision! And sorry for the late reply.
I will move the assembly code into an isolate `riscv_fpu_asm.S` file.

test/initramfs/src/regression/process/signal/riscv_fpu.c
le-monde-bleu: Thank you for your revision! And sorry for the late reply.
I will use libc `syscall()` instead.

test/initramfs/src/regression/process/signal/riscv_fpu.c
le-monde-bleu: Thank you for your revision! And sorry for the late reply.
I will use the in-house test framework.

test/initramfs/src/regression/process/signal/Makefile
```diff
@@ -10,4 +10,19 @@ C_OBJS_FILTER += \
...
+	riscv_fpu_asm
+
 include ../../common/Makefile
+
+ifeq ($(HOST_PLATFORM), riscv64-linux)
+$(OBJ_OUTPUT_DIR)/riscv_fpu: riscv_fpu.c riscv_fpu_asm.S | $(OBJ_OUTPUT_DIR) $(DEP_OUTPUT_DIR)
+	@$(CC) $(C_FLAGS) $^ -o $@ $(EXTRA_C_FLAGS) \
+		-MMD -MF $(DEP_OUTPUT_DIR)/riscv_fpu.d
+	@echo "CC <= $@"
+endif
```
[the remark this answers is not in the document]
le-monde-bleu: This seems to be a false positive report. The current `Makefile` goes like:
```make
ifneq ($(HOST_PLATFORM), riscv64-linux)
C_OBJS_FILTER += \
	riscv_fpu
endif
```
which means, `riscv_fpu` is filtered only when `HOST_PLATFORM != riscv64-linux`. In RISC-V scheme build process, `riscv_fpu` is not filtered. And `run_test.sh` runs `./signal/riscv_fpu` only when `uname -m = riscv64`.

test/initramfs/src/regression/process/signal/riscv_fpu.c
```diff
@@ -0,0 +1,157 @@
...
+	CHECK(sigemptyset(&action.sa_mask));
+	CHECK(sigaction(SIGUSR1, &action, NULL));
+}
+END_SETUP()
+
+static bool syscall_preserves_fpu_state(void)
+{
+	load_fpu_state(initial_regs, INTERRUPTED_FCSR);
+
+	if (syscall(SYS_getpid) < 0) {
```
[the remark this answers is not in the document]
le-monde-bleu: This can be theoretically correct. But according to https://github.com/asterinas/asterinas/pull/3493#discussion_r3744125909, invoking a raw `ecall` wrapper might not be the best option. I will test the ABI callee-saved FP registers `fs0` - `fs11` rather than the caller-saved FP registers.

le-monde-bleu:
@tatetian 

I have updated the implementation:

- Moved the RISC-V assembly helpers out of `riscv_fpu.c` into a dedicated `riscv_fpu_asm.S` file.
- Replaced the custom raw syscall assembly wrappers with libc `syscall()` calls.
- Converted the test to use Asterinas's in-house regression framework with `FN_SETUP`, `FN_TEST`, and `TEST_RES`.
- Added an explicit Makefile rule to link `riscv_fpu.c` with `riscv_fpu_asm.S`, while preventing the `.S` file from being built as a standalone test.
- Switched the live FP test state from caller-saved FP registers to ABI callee-saved `fs0`-`fs11`, so libc calls cannot legally clobber the tested values.

test/initramfs/src/regression/process/signal/Makefile
```diff
@@ -10,4 +10,19 @@ C_OBJS_FILTER += \
...
+
+ASM_OBJS_FILTER += \
+	riscv_fpu_asm
+
 include ../../common/Makefile
+
+ifeq ($(HOST_PLATFORM), riscv64-linux)
+$(OBJ_OUTPUT_DIR)/riscv_fpu: riscv_fpu.c riscv_fpu_asm.S | $(OBJ_OUTPUT_DIR) $(DEP_OUTPUT_DIR)
+	@$(CC) $(C_FLAGS) $^ -o $@ $(EXTRA_C_FLAGS) \
```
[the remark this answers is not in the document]
le-monde-bleu: Fixed. Now compile `riscv_fpu_asm.S` separately into `$(DEP_OUTPUT_DIR)/riscv_fpu_asm.o` and link that object with `riscv_fpu.c`. As a result, `-MMD -MF` applies only to the C source, so its header dependencies can no longer be overwritten by the assembly input.

test/initramfs/src/regression/process/signal/riscv_fpu_asm.S
```diff
@@ -0,0 +1,47 @@
+# SPDX-License-Identifier: MPL-2.0
+
+.text
+
+.option push
+.option arch, +d
+
+.global load_fpu_state
+.type load_fpu_state, @function
+load_fpu_state:
```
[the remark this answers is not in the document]
le-monde-bleu: Fixed. I replaced `load_fpu_state` with an ABI-compliant assembly bridge that saves the caller's original `fs0`-`fs11` and `fcsr`, loads the test state, invokes a callback using libc `syscall()`, captures the resulting FP state, and restores the original state before returning to C. The signal handler now modifies only `fcsr` through a separate ABI-safe helper.

test/initramfs/src/regression/process/signal/riscv_fpu_asm.S
```diff
@@ -0,0 +1,120 @@
+# SPDX-License-Identifier: MPL-2.0
+
+.text
+
+.option push
+.option arch, +d
+
+.global run_with_fpu_state
+.type run_with_fpu_state, @function
+# Keep the test state inside this frame so the C caller's callee-saved FP
```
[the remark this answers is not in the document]
le-monde-bleu: Fixed. I moved the explanatory comment above the `.global`/`.type` block, so the function attributes are now directly adjacent to the `run_with_fpu_state` label as required by the assembly convention.

